PR #4246, opened 2024-08-23
Track enabled alpha features in telemetry

wbreza:
Tracks the enabled alpha features are part of telemetry requests.

### DIFF
--- a/cli/azd/cmd/middleware/telemetry.go
+++ b/cli/azd/cmd/middleware/telemetry.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"log"
+	"strings"
 
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
 	"github.com/azure/azure-dev/cli/azd/cmd/actions"
@@ -12,6 +13,7 @@ import (
 	"github.com/azure/azure-dev/cli/azd/internal/tracing"
 	"github.com/azure/azure-dev/cli/azd/internal/tracing/events"
 	"github.com/azure/azure-dev/cli/azd/internal/tracing/fields"
+	"github.com/azure/azure-dev/cli/azd/pkg/alpha"
 	"github.com/azure/azure-dev/cli/azd/pkg/azapi"
 	"github.com/azure/azure-dev/cli/azd/pkg/exec"
 	"github.com/azure/azure-dev/cli/azd/pkg/lazy"
@@ -21,15 +23,20 @@ import (
 
 // Telemetry middleware tracks telemetry for the given action
 type TelemetryMiddleware struct {
-	options            *Options
-	lazyPlatformConfig *lazy.Lazy[*platform.Config]
+	options             *Options
+	lazyPlatformConfig  *lazy.Lazy[*platform.Config]
+	alphaFeatureManager *alpha.FeatureManager
 }
 
 // Creates a new Telemetry middleware instance
-func NewTelemetryMiddleware(options *Options, lazyPlatformConfig *lazy.Lazy[*platform.Config]) Middleware {
+func NewTelemetryMiddleware(
+	options *Options,
+	lazyPlatformConfig *lazy.Lazy[*platform.Config],
+	alphaFeatureManager *alpha.FeatureManager) Middleware {
 	return &TelemetryMiddleware{
-		options:            options,
-		lazyPlatformConfig: lazyPlatformConfig,
+		options:             options,
+		lazyPlatformConfig:  lazyPlatformConfig,
+		alphaFeatureManager: alphaFeatureManager,
 	}
 }
 
@@ -61,6 +68,18 @@ func (m *TelemetryMiddleware) Run(ctx context.Context, next NextFn) (*actions.Ac
 	}
 
 	span.SetAttributes(fields.CmdArgsCount.Int(len(m.options.Args)))
+
+	// Track all alpha features that are enabled for the current command
+	if allFeatures, err := m.alphaFeatureManager.ListFeatures(); err == nil && len(allFeatures) > 0 {
+		enabledFeatures := []string{}
+		for _, feature := range allFeatures {
+			if strings.ToLower(feature.Status) == "on" {
+				enabledFeatures = append(enabledFeatures, feature.Id)
+			}
+		}
+
+		span.SetAttributes(fields.AlphaFeaturesKey.StringSlice(enabledFeatures))
+	}
 
 	// Set the platform type when available
 	// Valid platform types are validating in the platform config resolver and will error here if not known & valid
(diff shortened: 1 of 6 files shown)


ellismg (approved):
LGTM, FWIW, but I know @weikanglim is the expert here (and might have feedback on if we should be setting an attribute per enabled alpha feature instead of a single attribute with a slice value, I'm not sure what ends up being better for storage and querying.

Thanks for the additional validation in the end to end test :-)

cli/azd/cmd/middleware/telemetry.go
```diff
@@ -62,6 +69,18 @@ func (m *TelemetryMiddleware) Run(ctx context.Context, next NextFn) (*actions.Ac
 
 	span.SetAttributes(fields.CmdArgsCount.Int(len(m.options.Args)))
 
+	// Track all alpha features that are enabled for the current command
+	if allFeatures, err := m.alphaFeatureManager.ListFeatures(); err == nil && len(allFeatures) > 0 {
```
weikanglim, line 73: With the current approach, we would always do upfront calculation for feature flags for telemetry purposes. This does introduce a consistent performance overhead on all commands.

One thing that's stuck with me when working on telemetry, and a goal I have, is to try to make it minimally costly for users as well. For example, in [project.go](https://github.com/Azure/azure-dev/blob/a17935e47211ed01780281974c6d63b467de5b14/cli/azd/pkg/project/project.go#L136), we only emit telemetry for project config since we needed to load it anyways. The in-memory computations are extremely cheap compared to forcing on-disk load.

I was thinking if we would want to do something similar here, where we would do `tracing.SetUsageAttribute` on the result of [IsEnabled](https://github.com/Azure/azure-dev/blob/a17935e47211ed01780281974c6d63b467de5b14/cli/azd/pkg/alpha/alpha_feature_manager.go#L76). 

I'm thinking we'd want to track the unique set of feature flags, and I can make some changes in our attribute collection to make that easier if all this sounds good to you.

cli/azd/pkg/alpha/alpha_feature_manager.go
```diff
@@ -77,32 +79,32 @@ func (m *FeatureManager) IsEnabled(featureId FeatureId) bool {
 	// guard from using the alphaFeatureManager from multiple routines. Only the first one will create the cache.
 	m.withSync.Do(m.initConfigCache)
 
+	enabled := false
+
 	// For testing, and in CI, allow enabling alpha features via the environment.
-	envName := fmt.Sprintf("AZD_ALPHA_ENABLE_%s", strings.ToUpper(string(featureId)))
+	envName := fmt.Sprintf("AZD_ALPHA_ENABLE_%s", strings.ReplaceAll(strings.ToUpper(string(featureId)), ".", "_"))
```
vhvb1989, line 85: Is this a breaking change? 
If we want to support `_`, can we do it w/o removing the `.` ?

cli/azd/pkg/alpha/alpha_feature_manager.go
```diff
@@ -77,32 +79,32 @@ func (m *FeatureManager) IsEnabled(featureId FeatureId) bool {
...
-	// check if the feature is ON
-	if featureOn := isEnabled(m.userConfigCache, featureId); featureOn {
-		return true
-	}
-
-	// check if the feature has been set with a default value internally
-	if val, ok := defaultEnablement[strings.ToLower(string(featureId))]; ok {
-		return val
+	if enabled {
+		tracing.AppendUsageAttribute(fields.AlphaFeaturesKey.StringSlice([]string{string(featureId)}))
```
vhvb1989, line 104: Should we check if the attribute is already present in the trace?  or would that happen automatically?
This method could be called multiple times during a command execution

cli/azd/pkg/alpha/alpha_feature_manager.go
wbreza: I feel like it is a better experience when using environment variable override.  Example: Instead of `AZD_ALPHA_ENABLE_AKS.HELM` it is `AZD_ALPHA_ENABLE_AKS_HELM`

vhvb1989: agree. Can we support both so we don't break anyone currently relaying on the `FOO.BAR` ?

cli/azd/pkg/alpha/alpha_feature_manager.go
wbreza: Looks like @weikanglim has a PR #4250 out to track unique values.  I will migrate to that once his PR has been merged.

cli/azd/pkg/alpha/alpha_feature_manager.go
weikanglim: `.` isn't a valid variable ID character in many shells, and I don't see any existing [code references](https://github.com/search?q=AZD_ALPHA_ENABLE&type=code) in public repos.

```bash
bash-3.2$ export ASD.WHAT='true'
bash: export: `ASD.WHAT=true': not a valid identifier
```

I'd prefer if we break it, the uptake isn't too bad.

cli/azd/pkg/alpha/alpha_feature_manager.go
```diff
@@ -77,32 +79,32 @@ func (m *FeatureManager) IsEnabled(featureId FeatureId) bool {
...
 	if v, has := os.LookupEnv(envName); has {
 		if b, err := strconv.ParseBool(v); err == nil {
-			return b
+			enabled = b
 		} else {
 			log.Printf("could not parse %s as a bool when considering %s", v, envName)
 		}
+	} else if allOn := isEnabled(m.userConfigCache, AllId); allOn {
+		//check if all features is ON
+		enabled = true
```
weikanglim, line 94: One thing that could be interesting to know about is if `azd` users are using the `All` flag. If we think that's valuable, we could consider also adding `all` to the `config.features` attribute if this is present.

cli/azd/test/functional/telemetry_test.go
```diff
@@ -350,6 +352,44 @@ func Test_CLI_Telemetry_NestedCommands(t *testing.T) {
 	require.True(t, upCmdFound, "cmd.up not found")
 }
 
+func Test_Telemetry_AlphaFeatures_Enabled(t *testing.T) {
+	mockContext := mocks.NewMockContext(context.Background())
+
+	os.Setenv("AZD_ALPHA_ENABLE_INFRASYNTH", "true")
```
weikanglim, line 358: Maybe `t.Setenv` to avoid persisting the env var across tests.

cli/azd/pkg/alpha/alpha_feature_manager.go
vhvb1989: I think it is interesting.  The `All` flag is a way to turn azd into full experimental :) and receive the latest alpha feature always (no need to manual turning on).   Looking at most of folks turning this on will help us to probably offer one only switch for all alpha features, or if no-one uses it, maybe we can remove the `all` option (less to support)

vhvb1989 (approved):

weikanglim (approved):
Thanks for doing this!

PR #4250, opened 2024-08-26
tracing: Add the ability to append attributes with unique elements

weikanglim:
Sometimes, we need to track "a unique set of strings". This helps with that.

### DIFF
--- a/cli/azd/internal/tracing/attributes.go
+++ b/cli/azd/internal/tracing/attributes.go
@@ -78,6 +78,55 @@ func appendTo(v *valSynced, attr attribute.KeyValue) {
 	v.val.Store(newBaggage)
 }
 
+func appendToUnique(v *valSynced, attr attribute.KeyValue) {
+	v.mu.Lock()
+	defer v.mu.Unlock()
+
+	baggage := v.val.Load().(baggage.Baggage)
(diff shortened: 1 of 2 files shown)


cli/azd/internal/tracing/attributes.go
```diff
@@ -78,6 +78,55 @@ func appendTo(v *valSynced, attr attribute.KeyValue) {
...
+
+			attr = attr.Key.StringSlice(append(val.AsStringSlice(), attrValue))
+		case attribute.STRINGSLICE:
+			attrValues := attr.Value.AsStringSlice()
+			adds := make([]string, 0, len(attrValues))
+
+			for _, a := range attrValues {
+				found := false
+				for _, v := range val.AsStringSlice() {
+					if v == a {
```
vhvb1989, line 105: nit. Move the entire switch to a new function to reduce a little indentation

vhvb1989 (approved):
Not a fan of more than 3 indentation level. If possible, update the code with less levels

wbreza (approved):

cli/azd/internal/tracing/attributes.go
ellismg: `slices.Contains` might be helpful here.

Issue #4124, opened 2024-07-15
[Feature] `azure.yml` `image` to support environment variable substitution


johnnyreilly:
- [x] Make sure you've installed the latest version using [instructions in the wiki](../wiki/install)

**Output from `azd version`**
N/A

**Describe the bug**
Not a bug, a missing feature! 

Consider the following: `azure.yml`:

```yaml
# yaml-language-server: $schema=https://raw.githubusercontent.com/Azure/azure-dev/main/schemas/v1.0/azure.yaml.json

name: my-container-app
metadata:
  template: azd-init@1.9.4
services:
  web:
    image: myregistry.azurecr.io/my-project/my-container-app:${WEB_VERSION_TAG}
    host: containerapp
```

The `image` property under `services` **does not** support environment substitution according to: https://raw.githubusercontent.com/Azure/azure-dev/main/schemas/v1.0/azure.yaml.json

However, the same `image` property under `docker` **does** support environment substitution according to: https://raw.githubusercontent.com/Azure/azure-dev/main/schemas/v1.0/azure.yaml.json

**To Reproduce**
N/A

**Expected behavior**
Wouldn't it be amazing if `services.image` **did** support environment substitution in the same way as `docker.image`?

**Environment**
N/A

**Additional context**
This came out the work done in this post: https://johnnyreilly.com/using-azd-for-faster-incremental-azure-container-app-deployments-in-azure-devops

rajeshkamal5050:
@wbreza can we fix this?

PR #4254, opened 2024-08-27
Adds support for environment variable substitution for source container image

wbreza:
Fixes: #4124

### DIFF
--- a/cli/azd/pkg/project/container_helper_test.go
+++ b/cli/azd/pkg/project/container_helper_test.go
@@ -352,7 +352,7 @@ func Test_ContainerHelper_Deploy(t *testing.T) {
 			)
 			serviceConfig := createTestServiceConfig("./src/api", ContainerAppTarget, ServiceLanguageTypeScript)
 
-			serviceConfig.Image = tt.image
+			serviceConfig.Image = osutil.NewExpandableString(tt.image)
 			serviceConfig.RelativePath = tt.project
 			serviceConfig.Docker.Registry = tt.registry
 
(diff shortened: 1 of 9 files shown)


cli/azd/pkg/project/container_helper_test.go
```diff
@@ -352,7 +352,7 @@ func Test_ContainerHelper_Deploy(t *testing.T) {
 			)
 			serviceConfig := createTestServiceConfig("./src/api", ContainerAppTarget, ServiceLanguageTypeScript)
 
-			serviceConfig.Image = tt.image
+			serviceConfig.Image = osutil.NewExpandableString(tt.image)
```
vhvb1989, line 355: Happy to see this, but, have you thought about something on the long term to support env-substitutions at a lower support-maintain cost?  
I think it might be a continuous effort to keep moving one string at a time, based on customer's issues.

What if, for example, we just consider the entire azure.yaml as an `expandableString` and allow substitutions everywhere?  Do you think that might be simpler, cheaper and easy on the long future compared to individual field integration? (considering that, for each expandable field, we need to:
  - use the Expandable type
  - call `Envsubst(...)` before using the field
  - update the schema
)

vhvb1989 (approved):
Happy to see this, but worry about a possible accumulation of individual expandable fields... Added a comment about it.

cli/azd/pkg/project/container_helper_test.go
jongio: That was how my original POC did it. Just ran envsub on whole file.

ellismg: I am very nervous about the idea of just expanding the whole file.  One place we know that would break things is in embedded hooks which want to do shell like references (we had this problem in the past, either in azure.yaml or another file we pre-processed).

I'm also nervous about the expansion leading to invalid YAML and other stuff - I think keeping it scoped to fields and not the whole document (and specific fields, where we want this behavior) makes more sense.

ellismg (approved):

Issue #1039, opened 2022-10-31
[Issue] Deployed zip does not obey .funcignore


pamelafox:
- [X] Make sure you've installed the latest version using [instructions in the wiki](../wiki/install)

**Output from `azd version`**

azd version 0.3.0-beta.4-pr.1945868 (commit 3aadf6dc4c842ba6813e423e4e9d6788e6dff200)

**Output from `az version`** 

```
{
  "azure-cli": "2.40.0",
  "azure-cli-core": "2.40.0",
  "azure-cli-telemetry": "1.0.8",
  "extensions": {
    "containerapp": "0.3.11",
    "containerapp-compose": "0.2.2"
  }
}
```

**Describe the bug**

I'm fairly certain that azd is zipping up everything in the folder and not ignoring files/folders in funcignore which I would expect to be ignored. 

I suspect that because 1) zipping my little function folder takes a lot longer than I'd expect and 2) I see no evidence of funcignore parsing in codebase, and I see this TODO here about ignoring files:
https://github.com/Azure/azure-dev/blob/a74e800f053779cb97e365068b80c879c328c2a9/cli/azd/pkg/project/internal/project_utils.go#L16

**To Reproduce**

1. git clone https://github.com/pamelafox/regression-model-azure-demo at 6d8c7e76981186663f888a1f7cd55308b2ce0f3f
2. Run azd up
3. Browse "App Files" in Azure Portal, still see .DS_Store

**Expected behavior**

I expect it to not zip up files/folders specified in .funcignore

**Environment**

Mac OS X Ventura

**Additional context**

I can reorganize my project structure to avoid such a large deploy, but I think it'd be good to support .funcignore for consistency with other ways to deploy function apps.

ashikns:
+1 to this issue. I've had to restructure a project to avoid the frontend node_modules folder getting published. Another problem is that local.settings.json also gets published, which can contain sensitive info like connection strings.

Is there any workaround to this issue?

jongio:
@pamelafox and @ashikns - I have a PR out that should resolve this issue.  Can you please try that PR and confirm is resolves this?  

https://github.com/Azure/azure-dev/pull/4258

You will need to add a .funcignore file to the root of the service folder you want azd to honor.  

/src/service1/.funcignore.

Thanks
Jon

weikanglim:
Released with [1.18.0](https://github.com/Azure/azure-dev/releases/tag/azure-dev-cli_1.18.0)

Issue #4215, opened 2024-08-14
[Issue] azd deploy excludes node modules by default for zip deployment of node.js projects


tvriesde:
- [x] Make sure you've installed the latest version using [instructions in the wiki](../wiki/install)

In framework_service_npm.go node_modules seem to be excluded as part of the zip deployment. 

This requires App Service deployments to have SCM_DO_BUILD_DURING_DEPLOYMENT=true for app services.

It would be helpful if this is mentioned in the azd documentation.  (maybe i missed it)

and/or Node_module exclusion can be set as an option for azd deploy. So that developers can choose if they want to include node_modules in zip deployment or not.

rajeshkamal5050:
@weikanglim can you take a look and clarify?

weikanglim:
@tvriesde Thanks a lot for mentioning this. You are indeed correct that `SCM_DO_BUILD_DURING_DEPLOYMENT` is needed in `azd` today for nom-based apps so that app service would run a full build remotely (`npm install` and `npm build` in this case) while being deployed.

Do you know what you ended up deciding on? Did setting `SCM_DO_BUILD_DURING_DEPLOYMENT=true` satisfy your needs, or would you have preferred to upload a fully built application?

At first glance, a nice thing to do here is for `azd` to probe the app service settings and act accordingly, since specifying `SCM_DO_BUILD_DURING_DEPLOYMENT=false` and not including `node_modules` wouldn't quite work. Alternatively, we could consider ignore options for deployment targets.

tvriesde:
I chose to set SCM_DO_BUILD_DURING_DEPLOYMENT=true. 

In the end this works for me. But I ran into this because initially the app did not work due to the modules not being present at the server. 

When following the guidance from Microsoft. https://learn.microsoft.com/en-us/azure/app-service/configure-language-nodejs?pivots=platform-linux There is no mention of SCM_DO_BUILD_DURING_DEPLOYMENT. So AZD not including the node_modules, caused my app deployment to fail. 

Even though I follow the article https://learn.microsoft.com/en-us/azure/app-service/configure-language-nodejs?pivots=platform-linux.

So there is a bit of friction, that can be improved by:
Appending the documentation, or adapting the tool a bit, like you mention @weikanglim 

-------------------------------------
When it comes down to the solution you present. This would have helped me indeed. Or presenting it as an option, so that I make an "ïnfromed" decision. Because zipping the node_modules might not always be the preferred scenario.

[...]

PR #4146, opened 2024-07-24
Add support for .azdignore

jongio:
Fixes #4142

### Summary

This pull request introduces support for `.azdignore` files in the `Initializer` component, providing users with the ability to specify files and directories to be excluded during initialization. This functionality mirrors the behavior of `.gitignore` files, enhancing the flexibility and control over which files are included in the project setup. Additionally, the ability to load a template from a local folder has been added to facilitate ease of end-to-end testing with the init function.

### Scenario

A template author needs file in the repo to validate the repo, but doesn't want a consumer of the template to have those files copied when they call `azd init`.

With the addition of `.azdignore` support, users can now define ignore patterns upfront, simplifying the initialization process.

### New Features

1. **Support for `.azdignore` Files**:
   - Users can create a `.azdignore` file in the root of the project directory.
   - The `.azdignore` file uses the same syntax as `.gitignore`, allowing users to specify patterns for files and directories to be excluded.

2. **Automatic Exclusion of Specified Files**:
   - During initialization, the `Initializer` component will read the `.azdignore` file and automatically exclude the specified files and directories from being copied to the project directory.

3. **Loading Templates from Local Folders**:
   - Users can now specify a local folder path for templates, facilitating easier end-to-end testing and development.

### How to Use

1. **Create a `.azdignore` File**:
   - In the root of your project directory, create a file named `.azdignore`.
   - Add patterns to this file to specify which files and directories should be ignored.

   Example `.azdignore` file:

   ```plaintext
   # Ignore all log files
   *.log

   # Ignore directories
   tmp/
   build/

   # Ignore specific files
   secret.yaml
   ```

2. **Initialize the Project**:
   - Run the initialization process as usual.
   - The files and directories matching the patterns in the `.azdignore` file will be excluded automatically.

3. **Use Local Templates**:
   - Specify a local folder path when initializing with a template to use local template files.
   - Example: `azd init --template ./path/to/local/template`

### To-Do

- **Update Documentation**:
  - Update the README.md to include instructions and examples for using `.azdignore` files.
  - Include information on using local folder paths for templates.

Issue #3520, opened 2024-03-12
Run `azd pipeline config --provider azdo` pipeline missing Java SDK 


v-jiaodi:
**Describe the issue:**
Test `todo-java-mongo` template. Run `azd pipeline config --provider azdo`, get error as follow：
![image](https://github.com/Azure/azure-dev/assets/80496810/ab1fcccf-5d41-4a80-b5f3-4825c9ac707e)

**Repro Steps:**

1. Login in with command: `azd auth login`.
2. Execute command: `azd init -t todo-java-mongo -b staging`.
3. Run `azd pipeline config --provider azdo`.

**Environment:**
- Template: todo-java-mongo.
- OS: Windows, Mac, Codespace, Linux, DevContainer
- Azd version : 1.7.0-beta.1-daily.3582348  (commit https://github.com/Azure/azure-dev/commit/9a0cef9c79d83a5f3fe8cb8bc407a200139da62b)

**Expected behavior:**
Run `azd pipeline config --provider azdo`, pipeline log can pass.

@rajeshkamal5050  for notification.

rajeshkamal5050:
@vhvb1989 can you triage this?

vhvb1989:
the pipeline definition uses:

```yaml
  - task: JavaToolInstaller@0
    inputs:
      versionSpec: '17'
      jdkArchitectureOption: 'x64'
      jdkSourceOption: 'PreInstalled'
```

By using `jdkSourceOption: 'PreInstalled'`, it relies on JAVA SDK to be already pre-installed on the agent, which is the case for `Microsoft-hosted agents` (https://learn.microsoft.com/azure/devops/pipelines/tasks/reference/java-tool-installer-v0?view=azure-pipelines#inputs)

It looks like the agent, in this case, doesn't have a pre-install JAVA sdk.

@v-jiaodi , can you show the output from JavaToolInstaller?:

![image](https://github.com/Azure/azure-dev/assets/24213737/e60f18fd-7081-4c47-a36e-71abc4213e00)

Note:
This is not azd or template's issue.  This looks like an issue with the Azure DevOps instace configuration.

v-jiaodi:
@vhvb1989 Output from JavaToolInstaller:
![image](https://github.com/Azure/azure-dev/assets/80496810/c90ca201-5a3a-4acb-8c11-9dfb79c7b345)

zhangjiale-64:
We also encountered the same problem in this manual test.

**Environment:**


[...]

PR #4258, opened 2024-08-27
Add `.webappignore` and `.funcignore` Support for Service Packaging

jongio:
Fixes: https://github.com/Azure/azure-dev/issues/1039 https://github.com/Azure/azure-dev/issues/4215

### Description:

This PR introduces support for `.webappignore` and `.funcignore` files to the service packaging process within the Azure Developer CLI (azd). These files allow users to specify files and directories that should be excluded from the deployment package for App Service and Function Apps, providing fine-grained control at the service level.

#### Key Updates:

1. **`.webappignore` and `.funcignore` File Support:**
   - Implemented the ability to define `.webappignore` for App Services and `.funcignore` for Azure Functions within individual service directories.
   - If a `.webappignore` or `.funcignore` file exists in a service directory, the packaging process will follow its rules, allowing granular control over which files are included or excluded in the service's zip archive.
   - These ignore files ensure that users can customize file exclusions while packaging for App Services and Azure Functions independently.

2. **Updated Exclusion Logic:**
   - By default, the packaging process excludes directories like `__pycache__`, `.venv`, and `node_modules` unless a `.webappignore` or `.funcignore` file is present, which can override these default exclusions.
   - The packaging process respects these ignore files:
     - If a `.webappignore` file is present in an App Service, directories such as `node_modules` or `__pycache__` can be included or excluded as specified.
     - For Azure Functions, `.funcignore` can similarly be used to specify inclusion/exclusion rules.
   - If no `.webappignore` or `.funcignore` file is present, default exclusions apply for Python (`__pycache__`, `.venv`) and Node.js (`node_modules`).

3. **Test Coverage:**
   - Extensive tests have been added to validate the behavior under different scenarios:
     - Packaging with and without `.webappignore` and `.funcignore` files.
     - Ensuring correct exclusions of `__pycache__`, `.venv`, and `node_modules` based on the presence or absence of the ignore files.
     - Ensuring service-specific `.webappignore` and `.funcignore` behavior is correctly respected.
     - Verifying that specific files such as `logs/log.txt` can be excluded based on the ignore file configuration.

4. **Sample Project Structure:**
   - Updated the sample project structure used in tests to include services representing both App Services and Azure Functions (Python and Node.js examples).
   - Added scenarios where `.webappignore` files explicitly include or exclude `node_modules` or `__pycache__` based on real-world usage patterns.

5. **Documentation:**
   - **TODO:** Update documentation to include guidance on how to use `.webappignore` and `.funcignore` files with examples for both Python and Node.js projects.

### Testing:

All test cases have been successfully executed, covering a broad range of scenarios to ensure expected packaging behavior. The tests ensure that both `.webappignore` and `.funcignore` files are respected and that default exclusions are correctly applied when no ignore file is present.

### Additional Notes:

- The change is backward-compatible and only affects users who introduce `.webappignore` or `.funcignore` files into their services.
- This feature gives users more control over the contents of their deployment packages, allowing for more streamlined and efficient deployments.

### How to Review:

- Review the changes to the service packaging logic, focusing on how `.webappignore` and `.funcignore` files influence the exclusion logic.
- Ensure that test cases provide full coverage for both App Service and Azure Function scenarios.
- **TODO:** Ensure documentation updates reflect the new functionality for users.


cli/azd/pkg/project/framework_service_npm.go
```diff
@@ -130,11 +130,7 @@ func (np *npmProject) Package(
 	if err := buildForZip(
 		packageSource,
 		packageDest,
-		buildForZipOptions{
-			excludeConditions: []excludeDirEntryCondition{
-				excludeNodeModules,
-			},
-		}); err != nil {
+		buildForZipOptions{}); err != nil {
```
jongio, line 133: These exclusions where moved to project_utils, because they are now dependent on the existance of .zipignore

cli/azd/pkg/project/framework_service_python.go
```diff
@@ -130,12 +130,7 @@ func (pp *pythonProject) Package(
 	if err := buildForZip(
 		packageSource,
 		packageDest,
-		buildForZipOptions{
-			excludeConditions: []excludeDirEntryCondition{
-				excludeVirtualEnv,
-				excludePyCache,
-			},
-		}); err != nil {
+		buildForZipOptions{}); err != nil {
```
jongio, line 133: These exclusions where moved to project_utils, because they are now dependent on the existance of .zipignore

pamelafox:
This seems helpful, per discussions on the other PR! My main design question would be why a separate file versus an --ignore in the azure.yaml. Is it easier to define globs that way? Does YAML have annoying syntax restrictions?

vhvb1989:
> This seems helpful, per discussions on the other PR! My main design question would be why a separate file versus an --ignore in the azure.yaml. Is it easier to define globs that way? Does YAML have annoying syntax restrictions?

One of the reasons was for making this new ignore-file recognizable beyond azd; so you could use it when using web-app-deploy action (https://github.com/Azure/webapps-deploy) or your own tools.  @jongio was going to check with App Service if there was something already for it and if not, it might be worth it to have it beyond azd; similar to .dockerignore, .npmignore, etc.  Then azd would just honor this file like any other tool would.

jongio:
> > This seems helpful, per discussions on the other PR! My main design question would be why a separate file versus an --ignore in the azure.yaml. Is it easier to define globs that way? Does YAML have annoying syntax restrictions?
> 
> One of the reasons was for making this new ignore-file recognizable beyond azd; so you could use it when using web-app-deploy action (https://github.com/Azure/webapps-deploy) or your own tools. @jongio was going to check with App Service if there was something already for it and if not, it might be worth it to have it beyond azd; similar to .dockerignore, .npmignore, etc. Then azd would just honor this file like any other tool would.

Yes, I'm in talks with them on this.

pamelafox:
Ah, cool, that design makes sense then!

cli/azd/pkg/project/framework_service_python.go
vhvb1989, line 133: each `framework_service_language` should continue to provide a default `excludeConditions` for when no .zipignore is defined in the project.

Maybe you can call a method here from the `dotignore` package and pass the default excluding files.  The method will check if .zipignore exists. If no, it uses the default rules passed by the `framework_service_language`. If yes, it uses the rules from the .zipignore

cli/azd/pkg/project/project_utils.go
```diff
@@ -48,12 +57,46 @@ type buildForZipOptions struct {
...
+	if err != nil && !os.IsNotExist(err) {
+		return fmt.Errorf("reading .zipignore files: %w", err)
+	}
+
+	// Determine if the .zipignore file exists at the root of the service
+	zipIgnoreExists := len(ignoreMatchers) > 0
+
+	// Conditionally exclude virtual environments, __pycache__, and node_modules only if .zipignore doesn't exist
+	if !zipIgnoreExists {
+		options.excludeConditions = append(options.excludeConditions, excludeVirtualEnv, excludePyCache, excludeNodeModules)
```
vhvb1989, line 75: the list of things to exclude here should come from the `framework_service_language`.  This util method should not be the source of truth for that.

vhvb1989 (commented):
few comments

jongio:
Comment from @Petermarcu 

https://github.com/Azure/azure-dev/pull/4146#issuecomment-2319020209

Is zip too specific and this is more of a ignore on deploy in azd terms? deployignore? Will we use a similar concept if the output is not in the form of a zip file in the future? Even just a lose folder of files that is ready to deploy?


@ellismg - You mentioned first mentioned .zipdeployignore, we landed on .zipignore because it's not necessarily part of the deploy action, instead it is part of packaging.

We want the file name to be generic enough to be used by other app service related tooling.  In the same vein as .funcignore, .dockerignore, should we name this .webappignore?

Petermarcu:
Is this only going to be respected for app service services?

jongio:
> Is this only going to be respected for app service services?

Yes, most of the other services already have .*ignore files, .funcignore, .dockerignore

jongio:
Related: https://github.com/Azure/azure-dev/issues/4215

jongio:
> We want the file name to be generic enough to be used by other app service related tooling. In the same vein as .funcignore, .dockerignore, should we name this .webappignore?

@ellismg - I've confirmed with the web app and functions teams that they would prefer `.webappignore`.  If no objections, I think we should move forward with that.  Agree?

ellismg:
> @ellismg - I've confirmed with the web app and functions teams that they would prefer .webappignore. If no objections, I think we should move forward with that. Agree?

Sounds great to me, @jongio!

cli/azd/pkg/project/service_target.go
```diff
@@ -94,6 +94,19 @@ type ServiceTarget interface {
...
+func GetIgnoreFileNameByKind(kind ServiceTargetKind) string {
+	switch kind {
+	case AzureFunctionTarget:
+		return ".funcignore"
+	case AppServiceTarget:
+		return ".webappignore"
+	default:
+		return ""
+	}
+}
```
jongio, line 108: I debated on whether or not we should have a helper method like this, or add this to the interface, but since only a few of the service targets will have this capability I figured we should add it as a helper method to not mess with the interface of the service target. But I could be convinced the other direction, depending on how you want to design this.

jongio:
@wbreza @vhvb1989 @ellismg - Could you have a look at the updated code and PR description with all the changes?  I'm in discussion with the appservice team and the approach with .webappignore and .funcignore are aligned with them.

@pamelafox - Once we get a build could you verify this works as expected in your projects with and without .webappignore files?  (You'll obviously need to create the .webappignore file in the project this impacts)

@v-xuto - Can you please run manual regression / smoke tests on all the azdev ToDo templates and ensure that everything passes?

Thanks!

kristenwomack:
> @wbreza @vhvb1989 @ellismg - Could you have a look at the updated code and PR description with all the changes? I'm in discussion with the appservice team and the approach with .webappignore and .funcignore are aligned with them.
> 
> @pamelafox - Once we get a build could you verify this works as expected in your projects with and without .webappignore files? (You'll obviously need to create the .webappignore file in the project this impacts)
> 
> @v-xuto - Can you please run manual regression / smoke tests on all the azdev ToDo templates and ensure that everything passes?
> 
> Thanks!

Do you also want to run the azdev ToDo templates through the validation pipeline with this change? cc @anfibiacreativa

v-xuto:
> @v-xuto - Can you please run manual regression / smoke tests on all the azdev ToDo templates and ensure that everything passes?
> 

Got it. Testing.

cli/azd/pkg/project/service_target.go
weikanglim: I like flexibility -- simple functions are great for that purpose. I'm okay here.

cli/azd/pkg/project/project_utils.go
```diff
@@ -48,17 +50,43 @@ type buildForZipOptions struct {
...
-// details for each language which should not be ever copied.
-func buildForZip(src, dst string, options buildForZipOptions) error {
+// buildForZip is used by projects to prepare a directory for
+// zipping, excluding files based on the ignore file and other conditions.
+func buildForZip(src, dst string, options buildForZipOptions, serviceConfig *ServiceConfig) error {
+	// Lookup the appropriate ignore file name based on the service kind (Host)
+	ignoreFileName := GetIgnoreFileNameByKind(serviceConfig.Host)
+
+	// Read and honor the specified ignore file if it exists
+	ignoreMatcher, err := dotignore.ReadDotIgnoreFile(src, ignoreFileName)
```
weikanglim, line 60: Are we also thinking about supporting `*.ignore` files in any recursive directory of the service?

```
- funcapp/
   - .funcignore
   - src/
     - otherThing/
       - .funcignore # define ignore files relative to this path
```

I think we would honor both ignore files.

If we're doing this, I think we should *consider* implementing the correct recursion logic in `project` before generalizing it as a `dotignore` package -- I would like to see the design matured out before we push it up to an app library package in `azd`.

weikanglim (commented):
Don't have strong feelings yet on the PR, have an overall question about whether we're thinking about recursive directories.

weikanglim (commented):
> Are we also thinking about supporting `*.ignore` files in any recursive directory of the service?

> ```
- funcapp/
   - .funcignore
   - src/
     - otherThing/
       - .funcignore # define ignore files relative to this path
```

> I think we would honor both ignore files.

> If we're doing this, I think we should *consider* implementing the correct recursion logic in `project` before generalizing it as a `dotignore` package -- I would like to see the design matured out before we push it up to an app library package in `azd`.

File: cli/azd/pkg/project/project_utils.go (hunk 5 above)
Comment (weikanglim) [reply cli/azd/pkg/project/project_utils.go:60]: @ellismg had implemented `.dockerignore` recently -- wonder if he has thoughts here.

File: cli/azd/pkg/project/project_utils.go [hunk 6]
```diff
@@ -9,34 +9,36 @@ import (
...
 
+	// Close the zip file and return the path
 	if err := zipFile.Close(); err != nil {
-		// may fail but, again, we'll do our best to cleanup here.
 		os.Remove(zipFile.Name())
 		return "", err
 	}
 
-	return zipFile.Name(), nil
+	return filePath, nil
```
wbreza, line 41: I'm assuming `filePath` gives us the full absolute path while `zipFile.Name()` was only giving us the filename?  I would just double check that this is honored downstream of the caller.

wbreza (approved):
Added a couple minor comments & questions but otherwise looks good.

v-xuto:
> @v-xuto - Can you please run manual regression / smoke tests on all the azdev ToDo templates and ensure that everything passes?
> 
@jongio We have finished the manual regression tests. We filed four new issues: https://github.com/Azure/azure-dev/issues/4346 ,https://github.com/Azure/azure-dev/issues/4347, https://github.com/Azure/azure-dev/issues/4348, https://github.com/Azure/azure-dev/issues/4349, and there is an old issues https://github.com/Azure/azure-dev/issues/3520.

cli/azd/pkg/project/project_utils.go
jongio: > @ellismg had implemented `.dockerignore` recently

@ellismg - Are you thinking we migrate the pack_source.go implementation to use the new dotignore support in this PR?

weikanglim: > .funcignore only honors files in the root and in discussion with them and the app service team we decided to not honor recursive .ignore files.

That is quite an interesting decision. From a dev perspective, one main reason why `*.ignore` files are used today is precisely because they can always be placed relative to a directory and provide a more local ignore scope. If that is not being done, a simpler package manifest at the root of the package directory would be a more ideal format IMO.

I took a closer look. It also turns out that `.funcignore` does not conform to the `.gitignore` spec entirely. @ellismg @jongio 

- The issue on [functools](https://github.com/Azure/azure-functions-core-tools/issues/2701)
- Based on the previous issue, I gather that the library they're using is [gitignore-parser](https://github.com/codemix/gitignore-parser/issues) with its set of issues.

If we're already engaged in conversations with both app service and function teams, can we consider:

1. Evaluating the impact of having two different `.funcignore` interpretations (is there user pain we would worry about?)
2. Based on those conversations, decide if we would like to rally all teams to adopt the conformant spec, and think about potentially support recursive directories. We could also just support recursive directories from `azd` side, it would be perhaps 20 lines of code.

jongio: Clarifying that I meant to say..."service root", not the root of the repo.

This implementation honors .ignore files in the root of the service.  

I think we are on the same page.

jongio: @weikanglim - Any objections to moving forward with the current implementation as it matches what .funcignore does today and what they would like for .webappignore.  

This means a service zipping only recognizes the .ignore file that is in the root of that service folder and not any parent .ignore files.

cli/azd/pkg/project/project_utils.go
jongio: I have confirmed that both filePath and zipFile.Name() return the same value:

image

cli/azd/pkg/project/project_utils.go
weikanglim: @jongio Just checking -- where are we in resolving the difference between of behavior in interpretation for `.funcignore` by func-tools vs. azd, as highlighted [previously](https://github.com/Azure/azure-dev/pull/4258/files#r1767506732)? I haven't seen it mentioned.

Aside from that:
1. I'm a little unhappy about `dotignore` being a package and the implementation of the functions being tied to working "within a directory". I suspect in the future this may end up needing slight rewriting to support recursive directories. I'm more than happy to work with you to push some changes here to clean up the implementation.
2. I'm looking at the PR again, and it seems like there are significant changes to `rzip.go` that I'm unsure how or if they're related. They should be separated out if it's a bug fix.
3. I'm also a little nervous about the e2e tests that we're doing inspection of the zip files -- `package` likely requires changes in the future that would require us to rethink these test cases. The heavy addition of many test data files to set up "scenario tests" of all the projects also seem rather heavyweight and will be slightly harder to maintain in the long term. It feels like simply testing the "exclusion conditions" would suffice for testing the behavior with/without `*.ignore` files. 
4. We should double check again what the default behavior of without `.funcignore` should be -- I believe it should minimally ignore `local.settings.json` to be inline with function tools, see docs [here](https://learn.microsoft.com/en-us/azure/azure-functions/functions-run-local?tabs=macos%2Cisolated-process%2Cnode-v4%2Cpython-v2%2Chttp-trigger%2Ccontainer-apps&pivots=programming-language-java#upload-local-settings-to-azure).

Overall, I really do love the direction here. But I would love if we can have a simple clean landing of a minimal change -- there are currently 37 file changes in this PR alone. I'm more than happy to work with you to land a minimal change if we can document all the necessary requirements in a new issue referencing #1039 so we're all on the same page, and also to serve as official documentation for the feature.

lcephas:
Hi folks. I came here from https://github.com/Azure/azure-dev/issues/4215, because I ran into the issue myself. My question is regarding the defaults. What's the reason not to include node_modules by default and make .webappignore the custom if you want to use the SCM build instead? IMO, this would make it more inline with other languages supported by AZD and is the "expected" behavior.

jongio:
> Hi folks. I came here from #4215, because I ran into the issue myself. My question is regarding the defaults. What's the reason not to include node_modules by default and make .webappignore the custom if you want to use the SCM build instead? IMO, this would make it more inline with other languages supported by AZD and is the "expected" behavior.

@lcephas - Right now the default is to not include them, so instead of introducing unexpected behavior of including them by default and breaking all azd deployments with appservice or functions, we decided the best route is to maintain the existing behavior by default and allow a developer to break out of that behavior by using a custom .*ignore file.

cli/azd/pkg/project/project_utils.go
jongio: **On dotignore and recursive directories:**

Thanks for sharing your thoughts on dotignore. The current non-recursive behavior is consistent with how .funcignore works today, and I believe it's important to avoid introducing more differences. I’d prefer to maintain this alignment to keep things predictable and straightforward for developers.

**On changes to rzip.go:**

The changes in rzip.go are required to handle symbolic links in node_modules or virtual environments. Since these components could now be included, special code is necessary to ensure correct behavior. Keeping these updates within this PR ensures everything works seamlessly together.

**On e2e tests and test data files:**

Our current e2e tests are designed to cover any issues arising from the zip changes, particularly around symbolic links. The test data may be heavier, but it provides comprehensive coverage that ensures we don’t miss any potential edge cases.

**On default behavior for .funcignore:**

Regarding the behavior of .funcignore, I’m in touch with the functions team to verify if there are additional files that should be ignored, with or without the existence of a .funcignore file. It's also important to note that any issues where .funcignore doesn’t behave like .gitignore should be resolved within the func implementation. I don't think we should introduce behavior here that diverges from the expected .gitignore syntax.


Thank you for the offer, but I’m confident the current implementation is in line with our goals, and I don't feel any additional simplification is needed at this time. I appreciate your willingness to assist and your thoughtful feedback throughout this process

weikanglim: I've left a comment on [rzip.go](https://github.com/Azure/azure-dev/pull/4258/files#r1775899423). It isn't clear what motivates all the changes here. I'd still push towards creating a separate issue for `rzip.go` change so that we can understand and track the changes separately.

> I’m in touch with the functions team to verify if there are additional files that should be ignored

Would we consider transferring all the existing documentation on the PR into an issue so that we can have the requirements discussion? I'm also more than happy to help here if needed.

Overall, I do really want this to see these changes landed -- I'm just trying to make sure we're covering enough our bases from enough angles to make this a good user experience for our users.

weikanglim (changes requested):
Overall, I do really love the direction here and want to see these changes landed.

The small things I'm waiting on: I would love to see if we can land `rzip.go` separately, and also have the great documentation we have on this PR pushed upwards into an issue. Just let me know if you'd like any help with that.

cli/azd/pkg/project/project_utils.go
jongio: Here are the list of default excludes for function service type:
https://github.com/Azure/azure-functions-core-tools/blob/37b826f143b99cd62308b88320e71aa806f6197f/src/Azure.Functions.Cli/Common/FileSystemHelpers.cs#L144

We may also want to confirm with app service team if there are default excludes like appsettings.development.json

cli/azd/pkg/rzip/rzip.go
```diff
@@ -12,47 +12,77 @@ import (
...
 
-		header := &zip.FileHeader{
-			Name: strings.Replace(
-				strings.TrimPrefix(
-					strings.TrimPrefix(path, source),
-					string(filepath.Separator)), "\\", "/", -1),
-			Modified: fileInfo.ModTime(),
-			Method:   zip.Deflate,
+		// Skip symbolic links
+		if fileInfo.Mode()&os.ModeSymlink != 0 {
```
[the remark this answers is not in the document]
jongio: I'm actually thinking that maybe we want to keep venvs excluded by default.  I don't think we have a good use case for including venvs in a zip folder.

@pamelafox - Can you confirm this?  And also do you think that we should include __pycache__ by default if they have a .ignore file?  I'm thinking it will be large and might not be needed in any scenario.

PR #4259, opened 2024-08-28
chore: fix new `gosec` reported issues

weikanglim:
Fixing new reported `gosec` issues. Issues are largely false-positive in nature that don't matter in practice.


cli/azd/pkg/input/console.go
```diff
@@ -925,13 +925,13 @@ func (c *AskerConsole) Handles() ConsoleHandles {
 }
 
 // consoleWidth the number of columns in the active console window
-func consoleWidth() int {
+func consoleWidth() int64 {
 	width, _ := consolesize.GetConsoleSize()
```
weikanglim, line 929: I just threw more memory at the problem here -- there wasn't an actual overflow risk.

The underlying pixel counts are `u16`, being returned as `int`.

Happy to take suggestions.

cli/azd/pkg/output/ux/list_as_text.go
```diff
@@ -17,14 +17,15 @@ import (
...
-		log.Panic("calling itemsCountAsText() with empty list.")
+		log.Panic("calling ListAsText() with empty list.")
 	}
 
 	if count == 1 {
 		return items[0]
 	}
 
 	if count == 2 {
+		//nolint:gosec // G602: slice index out of range - false positive, we know the slice has at least 2 elements
```
weikanglim, line 28: could not think of a cleaner way to rewrite the function, `gosec` isn't correctly inferring `count` refers to `len(items)` in this usage.

cli/azd/pkg/tools/github/github.go
```diff
@@ -517,7 +517,7 @@ func extractFromTar(src, dst string) (string, error) {
 		// cspell: disable-next-line `Typeflag` is comming fron *tar.Header
 		if fileHeader.Typeflag == tar.TypeReg && fileName == "gh" {
 			filePath := filepath.Join(dst, fileName)
-			ghCliFile, err := os.OpenFile(filePath, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, os.FileMode(fileHeader.Mode))
+			ghCliFile, err := os.OpenFile(filePath, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, fileHeader.FileInfo().Mode())
```
weikanglim, line 520: This is shifting the problem to the library without actually changing behavior.. Under the covers, the same [conversion](https://cs.opensource.google/go/go/+/refs/tags/go1.23.0:src/archive/tar/common.go;drc=d838e4dcdf89124ed051e1c53e8472e900664a6b;l=563) happens?

cli/azd/pkg/output/ux/list_as_text.go
ellismg: I wonder if a `switch` on `len(items)` would help the analysis here?  🤷.

cli/azd/pkg/tools/github/github.go
ellismg: Can't think of anything better here either.

ellismg (approved):
LGTM - Thanks for the cleanups.

cli/azd/pkg/input/console.go
weikanglim: I ended up thinking about this more -- I don't see a future with column widths exceeding  `math.MaxInt32`. Ended up doing the conversion in this function and suppress the false-positive overflow

cli/azd/pkg/output/ux/list_as_text.go
weikanglim: The `switch` didn't work -- looks like `gosec` isn't following the temp variable assignment. The one way I did find to fix this is to have `if len(items) ==2`. Supposedly, with compiler optimization, repeated `len(x)` calls should incur the same cost, but I'm just going to leave it as-is for now.

PR #4263, opened 2024-08-28
project: Use `/` as path seperator in `azure.yaml`

### LINKED ISSUE #3891: [Issue] [aspire] `azd infra synth` ignored by `azd deploy` in azdo ci/cd pipeline

**Output from `azd version`**
`azd version 1.9.0 (commit 651394c3ddcfadff194d177f8b0ddf06fe3752bf)`

**Describe the bug**
`azd infra synth` has been executed, and the generated templates in the AppHost project are committed and pushed to the build branch, but `azd deploy` generates the `infra` itself, ignoring any manual changes.

**To Reproduce**
- Create a new Aspire Starter App.
- Run `azd init`
- Add `state: remote: backend` for remote configuration.
- Add .azdo/pipelines/azure-dev.yaml
- Run `azd pipeline config`
- Permission created `azconnection` service principal with `Storage Blob Contributor` on your remote configuration blob storage.
- Update project to use `AddConnectionString` in `AppHost`.
- Use `AddReference` on the `ApiService` resource.
- Add `Aspire.Azure.Storage.Blobs` to `ApiService`.
- Run `azd env set AZURE_BLOB_STORAGE_CONNECTION ""` to add string to env.
- Run `azd infra synth`
- Edit `apiservice.tmpl.yaml` to replace `'{{ securedParameter "BlobStorageConnection" }}'` with `{{ .Env.AZURE_BLOB_STORAGE_CONNECTION }}`
- Edit `main.bicep` to add output AZURE_BLOB_STORAGE_CONNECTION string = BlobStorageConnection
- Push to trigger build.

Build will fail at `azd deploy` with message:
`ERROR: failed deploying service 'apiservice': failing invoking action 'deploy', failed executing template file: template: containerApp.tmpl.yaml:21:19: executing "containerApp.tmpl.yaml" at <securedParameter "BlobStorageConnection">: error calling securedParameter: parameter BlobStorageConnection not found`

This indicates that `securedParameter "BlobStorageConnection"` is still in the template.  Note that `containerApp.tmpl.yaml` is the file name, which was the file name from `azd 1.8.2`.

Log message indicates that new template was generated, not the one created by `azd infra synth`:
`2024/05/09 23:45:47 service_target_dotnet_containerapp.go:195: generating container app manifest from /home/vsts/work/1/s/.\AspireInfraExample.AppHost\AspireInfraExample.AppHost.csproj for project apiservice`

**Expected behavior**
Infra created by `azd infra synth` should be used.

**Environment**
    - dotnet 8.0.300-preview.24203.14
    - Version 17.10.0 Preview 7.0
    - azd version 1.9.0 (commit 651394c3ddcfadff194d177f8b0ddf06fe3752bf)

**Additional context**
Complete `azure.yaml`:
```

name: AspireInfraExample
services:  
  app:
    language: dotnet
    project: .\AspireInfraExample.AppHost\AspireInfraExample.AppHost.csproj
    host: containerapp
state:
  remote:
    backend: AzureBlobStorage
    config:
      accountName: azdconfiguration
      containerName: aspireinfraexample
```

Complete `azure-dev.yaml`:
```
trigger:
- master
- staging

pool:
  vmImage: ubuntu-latest

steps:
  - task: DotNetCoreCLI@2
    displayName: Install Aspire Workload with Preview
    inputs:
      command: custom
      custom: 'workload'
      arguments: 'install aspire --include-previews --source https://api.nuget.org/v3/index.json'

  - task: setup-azd@0 
    displayName: Install azd

  - pwsh: |
      azd config set auth.useAzCliAuth "true"
    displayName: Configure AZD to Use AZ CLI Authentication.

  - task: AzureCLI@2
    displayName: Provision Infrastructure
    inputs:
      azureSubscription: azconnection
      scriptType: bash
      scriptLocation: inlineScript
      inlineScript: |
        azd provision --no-prompt
    env:
      AZURE_SUBSCRIPTION_ID: $(AZURE_SUBSCRIPTION_ID)
      AZURE_ENV_NAME: $(AZURE_ENV_NAME)
      AZURE_LOCATION: $(AZURE_LOCATION)

  - task: AzureCLI@2
    displayName: Deploy Application
    inputs:
      azureSubscription: azconnection
      scriptType: bash
      scriptLocation: inlineScript
      inlineScript: |
        azd deploy --no-prompt --debug
    env:
      AZURE_SUBSCRIPTION_ID: $(AZURE_SUBSCRIPTION_ID)
      AZURE_ENV_NAME: $(AZURE_ENV_NAME)
      AZURE_LOCATION: $(AZURE_LOCATION)
```

ellismg:
Since we expect an `azd` project to be used across operating systems which use `/` and `\` as path seperators, we need a common way of writing paths in `azure.yaml` that works across both. In the path we supported using both `\` and `/` in `azure.yaml` (and went out ouf or way in the Aspire case to write the platform specific version during `init`) but in some cases this would lead to weird behavior when you then puth the project into GitHub actions and *nix runners would look for files that may not exist, due to the path seperator.

To address this, we update our `ProjectConfig` yaml parsing and saving to assume forward slash seperated paths, and then use `filepath.FromSlash` and `filepath.ToSlash` on the boundaries. This means the data stored at runtime inside `ProjectConfig` is using the correct platform seperator, but when being saved to disk, we always use `/`.

To support backwards compatability, for paths which have only `\` and no `/`, we treat as if they were written with `/` instead of `\` (and we will fix them up if we end up rewrting azure.yaml).

Fixes #3891

### DIFF
--- a/cli/azd/pkg/apphost/generate.go
+++ b/cli/azd/pkg/apphost/generate.go
@@ -383,7 +383,7 @@ func GenerateProjectArtifacts(
 	projectFileContext := genProjectFileContext{
 		Name: projectName,
 		Services: map[string]string{
-			"app": fmt.Sprintf(".%s%s", string(filepath.Separator), appHostRel),
+			"app": fmt.Sprintf("./%s", filepath.ToSlash(appHostRel)),
 		},
 	}
 
--- a/cli/azd/pkg/project/project.go
+++ b/cli/azd/pkg/project/project.go
@@ -82,6 +82,12 @@ func Parse(ctx context.Context, yamlContent string) (*ProjectConfig, error) {
 		projectConfig.Infra.Path = "infra"
 	}
 
+	if strings.Contains(projectConfig.Infra.Path, "\\") && !strings.Contains(projectConfig.Infra.Path, "/") {
+		projectConfig.Infra.Path = strings.ReplaceAll(projectConfig.Infra.Path, "\\", "/")
+	}
+
+	projectConfig.Infra.Path = filepath.FromSlash(projectConfig.Infra.Path)
+
 	for key, svc := range projectConfig.Services {
 		svc.Name = key
 		svc.Project = &projectConfig
@@ -103,12 +109,30 @@ func Parse(ctx context.Context, yamlContent string) (*ProjectConfig, error) {
 			return nil, fmt.Errorf("parsing service %s: %w", svc.Name, err)
 		}
 
+		if strings.Contains(svc.Infra.Path, "\\") && !strings.Contains(svc.Infra.Path, "/") {
+			svc.Infra.Path = strings.ReplaceAll(svc.Infra.Path, "\\", "/")
+		}
+
+		svc.Infra.Path = filepath.FromSlash(svc.Infra.Path)
+
 		// TODO: Move parsing/validation requirements for service targets into their respective components.
 		// When working within container based applications users may be using external/pre-built images instead of source
 		// In this case it is valid to have not specified a language but would be required to specify a source image
 		if svc.Host == ContainerAppTarget && svc.Language == ServiceLanguageNone && svc.Image.Empty() {
 			return nil, fmt.Errorf("parsing service %s: must specify language or image", svc.Name)
 		}
+
+		if strings.ContainsRune(svc.RelativePath, '\\') && !strings.ContainsRune(svc.RelativePath, '/') {
+			svc.RelativePath = strings.ReplaceAll(svc.RelativePath, "\\", "/")
+		}
+
+		svc.RelativePath = filepath.FromSlash(svc.RelativePath)
+
+		if strings.ContainsRune(svc.OutputPath, '\\') && !strings.ContainsRune(svc.OutputPath, '/') {
+			svc.OutputPath = strings.ReplaceAll(svc.OutputPath, "\\", "/")
+		}
+
+		svc.OutputPath = filepath.FromSlash(svc.OutputPath)
 	}
 
 	return &projectConfig, nil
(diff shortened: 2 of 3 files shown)


cli/azd/pkg/apphost/generate.go
```diff
@@ -383,7 +383,7 @@ func GenerateProjectArtifacts(
 	projectFileContext := genProjectFileContext{
 		Name: projectName,
 		Services: map[string]string{
-			"app": fmt.Sprintf(".%s%s", string(filepath.Separator), appHostRel),
+			"app": fmt.Sprintf("./%s", filepath.ToSlash(appHostRel)),
```
jongio, line 386: Could you include tests that include both azure.yaml paths in both ways to ensure we don't break compat?

cli/azd/pkg/project/project.go
```diff
@@ -205,7 +229,17 @@ func SaveConfig(ctx context.Context, config config.Config, projectFilePath strin
 
 // Saves the current instance back to the azure.yaml file
 func Save(ctx context.Context, projectConfig *ProjectConfig, projectFilePath string) error {
-	projectBytes, err := yaml.Marshal(projectConfig)
+	copy := *projectConfig
```
ellismg, line 232: This doesn't do a deep copy so the `Services` end up having their paths wacked, which was unintended - need to fix this.

cli/azd/pkg/project/project.go
```diff
@@ -82,6 +82,12 @@ func Parse(ctx context.Context, yamlContent string) (*ProjectConfig, error) {
 		projectConfig.Infra.Path = "infra"
 	}
 
+	if strings.Contains(projectConfig.Infra.Path, "\\") && !strings.Contains(projectConfig.Infra.Path, "/") {
```
ellismg, line 85: In an earlier iteration I had tried creating a new type called `osutil.PathString` which had a `String()` method and then used that on `ProjectConfig` and friends - but it started to feel pretty tedious.

It might be better to handle this with custom unmarshalling logic at the ProjectConfig and nested type layers - where they can apply the normalization, not sure.  It is nice that everything is shunted through these save and load methods.

ellismg:
@vhvb1989 I didn't yet validate that this addresses the issue in GitHub, but based on what I understood from your investigation, it seemed like this is what we might want.  Are you able to take the privates here and see if this addresses the issue as you understood it?

vhvb1989:
Worked on my end.
My test:
   From Windows, using daily build, run:
- dotnet new aspire-starter
- azd init  (azure.yaml is created with windows style path `C:\aa\aa`)
- re-open folder in container WSL
- install azd daily version and run azd infra synth, got error:
image

- install azd from this PR and run azd infra synth. It worked as expected.  Azd is able to handle windows style path

Thank you @ellismg

vhvb1989 (approved):

cli/azd/pkg/apphost/generate.go
ellismg: Wrote some directed tests.

weikanglim (approved):

wbreza (approved):

cli/azd/pkg/project/project.go
```diff
@@ -205,7 +229,25 @@ func SaveConfig(ctx context.Context, config config.Config, projectFilePath strin
 
 // Saves the current instance back to the azure.yaml file
 func Save(ctx context.Context, projectConfig *ProjectConfig, projectFilePath string) error {
```
wbreza, line 231: If we know we are going to be making mutations on the object and don't want these to be available in the calling instance we may as well just change the func to pass the `ProjectConfig` by value here.  Then you would only need to worry about copying/cloning the services.

PR #4268, opened 2024-08-29
replace FormRecognizer with DocumentIntelligence

### LINKED ISSUE #4148: [Issue] AZD cli prints form recognizer instead of document intelligence

**Output from `azd version`**
azd version 1.9.5 (commit cd2b7af9995d358aab33c782614f801ac1997dde)

**Describe the bug**
image

**Additional context**
I am running azd within the context of this repo https://github.com/Azure-Samples/azure-search-openai-demo

hemarina:
fix #4148 

- Form Recognizer is named to Document Intelligence, but its kind is still defined as FormRecognizer. Replacing `FormRecognizer` with `DocumentIntelligence` until the kind is renamed. 
- Keep the deletion flow logic instead of having the resource type to display detailed name mappings. Current deletion flow didn't list resources one by one like provision because we delete the whole resource group and I think it is good to keep it.

image


cli/azd/pkg/infra/provisioning/bicep/bicep_provider.go
```diff
@@ -903,6 +903,10 @@ func cognitiveAccountsByKind(
 	for resourceGroup, cogAccounts := range accountsByResourceGroup {
 		for _, cogAccount := range cogAccounts {
 			kindName := *cogAccount.Kind
+			// Replace "FormRecognizer" with "DocumentIntelligence"
+			if kindName == "FormRecognizer" {
+				kindName = "DocumentIntelligence"
```
ellismg, line 908:
```suggestion
				kindName = "Document Intelligence"
```

Maybe having a space here will look a little better?

ellismg (approved):
LGTM

Issue #4264, opened 2024-08-28
KEDA scale rules reset after `azd deploy`


wbreza:
When using API version `2024-02-02-preview` for Azure container apps and configuring KEDA scale rules

Example:
```bicep
"rules": [
          {
            "name": "service-bus-queue-length-rule",
            "custom": {
              "type": "azure-servicebus",
              "metadata": {
                "messageCount": "10",
                "namespace": "sb-tl47kxvg5xjgo",
                "queueName": "ticket-render-requests"
              },
              "identity": "/subscriptions/{subscription}/resourceGroups/{resourceGroup}/providers/Microsoft.ManagedIdentity/userAssignedIdentities/id-app-xxxxxxx"
            }
          }
        ]
```

## Steps to reproduce

1. Run `azd provision`  _Scale rules created successfully_
2. Run `azd deploy` _Scale rules identity configuration reset_

### Actual behavior

KEDA scale rules identity are reset after running `azd deploy`

### Expected behavior

KEDA scale rules remain intact and work correctly.

wbreza:
After initial investigation the following is happening.

`azd` is using Azure SDK for go that is generated from API version `2023-11-02-preview`

During `azd deploy` the container app metadata is retrieved, a few modifications are made to the metadata to support updating the revisions and referenced container image and the updated metadata is PUT back to the ARM control plan.

Since the new scale rules property is from a newer API version it is not unmarshalled into the existing SDK struct and when serialized back during the PUT operation the scale rules are essentially reset causing the current behavior.

`azd` should respect the API version of the configured container app without unintentionally truncating the configuration.

PR #4272, opened 2024-08-29
Support configurable `api-version` for container app deployments

wbreza:
## Features

- Adds support to specify an explicit `api-version` to use for Container apps that is leveraged during container app deployment to avoid getting and unmarshalling to incompatible api-versions and losing configuration data set in bicep.

- Updates deployment scenarios to be more loosely coupled to avoid loosing custom configuration data during JSON marshalling to hard coded SDK versions.

If an `apiVersion` is not set in the service config the SDK will fall back to use the SDK `api-version` it was originally from.

Fixes: #4264

## Usage Example

```yaml
name: todo-nodejs-mongo-aca
services:
  web:
    project: ./src/web
    language: js
    host: containerapp
    apiVersion: 2024-02-02-preview # Force `azd` to use this API version for GET/PATCH operations
  api:
    project: ./src/api
    language: js
    host: containerapp
    apiVersion: 2024-02-02-preview # Force `azd` to use this API version for GET/PATCH operations
```

App can be validated with the following fork of todo-aca sample
```bash
azd init -t https://github.com/wbreza/todo-nodejs-mongo-aca -b api-version
```

### DIFF
--- a/cli/azd/pkg/containerapps/container_app.go
+++ b/cli/azd/pkg/containerapps/container_app.go
@@ -242,10 +242,6 @@ func (cas *containerAppService) DeployYaml(
 			return fmt.Errorf("applying manifest: %w", err)
 		}
 		poller = p
-
-		// Now that we've sent the request, clear the body so it is not injected on any subsequent requests (e.g. ones made
-		// by the poller when we poll).
-		customPolicy.body = nil
 	} else {
 		// When the apiVersion field is unset in the YAML, we can use the standard SDK to build the request and send it
 		// like normal.
@@ -534,10 +530,6 @@ func (cas *containerAppService) updateContainerApp(
 		return fmt.Errorf("begin updating ingress traffic: %w", err)
 	}
 
-	if apiVersionPolicy != nil {
-		apiVersionPolicy.body = nil
-	}
-
 	_, err = poller.PollUntilDone(ctx, nil)
 	if err != nil {
 		return fmt.Errorf("polling for container app update completion: %w", err)
@@ -616,6 +608,9 @@ func (p *containerAppCustomApiVersionAndBodyPolicy) Do(req *policy.Request) (*ht
 		if err := req.SetBody(streaming.NopCloser(bytes.NewReader(*p.body)), "application/json"); err != nil {
 			return nil, fmt.Errorf("updating request body: %w", err)
 		}
+
+		// Reset the body on the policy so it doesn't get reused on the next request
+		p.body = nil
 	}
 
 	return req.Next()
--- a/cli/azd/pkg/containerapps/container_app.go
+++ b/cli/azd/pkg/containerapps/container_app.go
@@ -18,6 +18,7 @@ import (
 	"github.com/azure/azure-dev/cli/azd/pkg/account"
 	"github.com/azure/azure-dev/cli/azd/pkg/alpha"
 	"github.com/azure/azure-dev/cli/azd/pkg/config"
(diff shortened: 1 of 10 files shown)


cli/azd/pkg/convert/util.go
```diff
@@ -54,17 +56,51 @@ func ToMap(value any) (map[string]any, error) {
...
+	var mapValue map[string]any
 	if err := json.Unmarshal(jsonValue, &mapValue); err != nil {
 		return nil, fmt.Errorf("failed to convert value to map: %w", err)
 	}
 
 	return mapValue, nil
 }
 
+// ToJsonArray converts the specified value (slice) to a json array
+func ToJsonArray(value any) ([]any, error) {
```
ellismg, line 68: Just wondering - do we have to do a round trip via JSON here?  Couldn't we just do a cast to `[]any`?

cli/azd/pkg/containerapps/container_app.go
```diff
@@ -118,44 +134,49 @@ func (cas *containerAppService) persistSettings(
...
 		log.Printf("failed getting current aca settings: %v. No settings will be persisted.", err)
 	}
 
-	if aca == nil ||
-		aca.Properties == nil ||
-		aca.Properties.Configuration == nil ||
-		aca.Properties.Configuration.Ingress == nil {
-		// no settings to persist
-		return obj, nil
+	var ingress *armappcontainers.Ingress
```
ellismg, line 152: Converting to a model type here means we could drop fields if a newer API version introduced new fields, I think?  Is it safe to do this?

cli/azd/pkg/containerapps/container_app.go
```diff
@@ -30,13 +45,15 @@ type ContainerAppService interface {
 		subscriptionId,
 		resourceGroup,
 		appName string,
+		options *ContainerAppOptions,
 	) (*ContainerAppIngressConfiguration, error)
```
ellismg, line 49: Do we need to update the return type to be something "weakly typed" so we don't drop things when options points to a newer API version?

ellismg (commented):
If we end up moving to a world where we do arm deployments for container apps I'm not sure what this `api-version` value on `azure.yaml` is going to mean.  But maybe adding it for now and then removing it later (as part of moving container apps to GA) makes sense.  I can't think of a better place to capture this information.

Things look reasonable to me here (and cool use of `config.Config`) but I did notice a few places where we still using the model types which may be unsafe.

cli/azd/pkg/convert/util.go
wbreza: We also want to make sure proper JSON marshalling happens to go from `PascalCase` to `camelCase` expected by the service.

cli/azd/pkg/containerapps/container_app.go
wbreza: We can make this loosely typed everywhere but wanted to get some feedback before making all these changes.

ellismg (approved):
LGTM - Really nicely integrated into the system here - cool to see `config` leveraged in this new way.

wbreza:
Adds support to specify an explicit api-version to use for Container apps that is leveraged during container app deployment to avoid getting and unmarshalling to incompatible api-versions and losing configuration data set in bicep.

Updates deployment scenarios to be more loosely coupled to avoid loosing custom configuration data during JSON marshalling to hard coded SDK versions.

If an apiVersion is not set in the service config the SDK will fall back to use the SDK api-version it was originally from.

PR #4286, opened 2024-09-03
Aspire: Use bicep for container app deployment

ellismg:
The plan for Aspire 9 is to support customization of the container app for services by leveraging the CDK to generate the bicep IaC for a given container app.

To prepare for that new world, we'd like `azd infra synth` to start writing bicep and to modify our deployment strategy for container apps to be based on ARM deployments instead of directly PUTing a copy of the resource to the RP.

This change starts to work through that.

This is presently guarded behind an alpha flag, so you'll need to run:

```
azd config set alpha.aspire.useBicepForContainerApps on
```

In order to enable this work.  When that is turned on, `azd` will generate bicep templates instead of yaml durning `infra synth` and also look for bicep files instead of yaml files when doing a deployment.

Note - this is a work in progress and complex Aspire features may not be supported yet - but this does work with aspire-starter template.  The secure outputs/secrets work is still pretty rough and likely doesn't work, but I'm exploring that next.

fix: [#4396](https://github.com/Azure/azure-dev/issues/4396)


davidfowl:
The main change on the aspire side is in with experimental public APIs to customize this and turn it on. It changes all container and project resources to project.v1 and container.v1 to support this new nested deployment field.

cli/azd/pkg/apphost/testdata/TestAspireContainerGeneration-main.bicep.snap
```diff
@@ -81,8 +81,8 @@ output AZURE_LOG_ANALYTICS_WORKSPACE_NAME string = resources.outputs.AZURE_LOG_A
 output AZURE_CONTAINER_APPS_ENVIRONMENT_NAME string = resources.outputs.AZURE_CONTAINER_APPS_ENVIRONMENT_NAME
 output AZURE_CONTAINER_APPS_ENVIRONMENT_ID string = resources.outputs.AZURE_CONTAINER_APPS_ENVIRONMENT_ID
 output AZURE_CONTAINER_APPS_ENVIRONMENT_DEFAULT_DOMAIN string = resources.outputs.AZURE_CONTAINER_APPS_ENVIRONMENT_DEFAULT_DOMAIN
-output SERVICE_MY_SQL_ABSTRACT_VOLUME_PROJECTAPPHOST_VOLUME_TWO_NAME string = resources.outputs.SERVICE_MY_SQL_ABSTRACT_VOLUME_PROJECTAPPHOST_VOLUME_TWO_NAME
-output SERVICE_MYSQLABSTRACT_VOLUME_PROJECTAPPHOST_VOLUMEONE_NAME string = resources.outputs.SERVICE_MYSQLABSTRACT_VOLUME_PROJECTAPPHOST_VOLUMEONE_NAME
+output SERVICE_MY_SQL_ABSTRACT_VOLUME_PROJECTAPPHOSTVOLUMETWO_NAME string = resources.outputs.SERVICE_MY_SQL_ABSTRACT_VOLUME_PROJECTAPPHOSTVOLUMETWO_NAME
```
weikanglim, line 84: Is this the expected naming syntax? It looks a little off.. Would we prefer `_` separators at all?

cli/azd/pkg/apphost/generate.go
```diff
@@ -540,11 +583,19 @@ func (b *infraGenerator) LoadManifest(m *Manifest) error {
...
+		var deploymentParams map[string]any
+		var deploymentSource string
+		if comp.Deployment != nil {
+			deploymentParams = comp.Deployment.Params
+			deploymentSource = filepath.Base(*comp.Deployment.Path)
+		}
 		switch comp.Type {
-		case "project.v0":
-			b.addProject(name, *comp.Path, comp.Env, comp.Bindings, comp.Args)
+		case "project.v0", "project.v1":
```
ellismg, line 593: Should we split these cases so that `project.v0` explicitly passes the `'nil` and `""` for deploymentParams and deploymentSource?

cli/azd/resources/apphost/templates/containerApp.tmpl.bicepparamt
```diff
@@ -0,0 +1,6 @@
+{{define "containerApp.tmpl.bicepparam" -}}
```
ellismg, line 1: This in combination with `buildDeployBlock` was a very clever idea, @vhvb1989.

cli/azd/resources/apphost/templates/resources.bicept
```diff
@@ -86,7 +86,7 @@ resource volumesAccountRoleAssignment 'Microsoft.Authorization/roleAssignments@2
 {{- range $volume := $value.Volumes}}
 resource {{mergeBicepName $name $volume.Name}}FileShare 'Microsoft.Storage/storageAccounts/fileServices/shares@2022-05-01' = {
   parent: storageVolumeFileService
-  name: take('${toLower('{{$name}}')}-${toLower('{{removeDot $volume.Name}}')}', 32)
+  name: take('${toLower('{{$name}}')}-${toLower('{{removeDot $volume.Name}}')}', 60)
```
ellismg, line 89: I'm not sure if we are worried about this or not but the physical resource name will change because of this when someone updates their copy of `azd` (if they had not run `infra synth`).  Are we okay with that?

ellismg (commented):
Thanks for the updates while I was gone, @vhvb1989!  I took a look over the changes and went ahead and cleaned some stuff up (mostly removing some of the stuff related to this being behind an alpha flag when I first added it and then no longer needed bicep templates).

Had a few questions about some of the changes but nothing blocking.  The one thing I do wonder is if some of the bind mount changes end up changing physical resource names and if so if we want to disallow them, but this looks good.

The `.bicepparms` file generation is a really smart idea.  I wish I had thought of that instead of that weird `@metadata` thing I was thinking.

cli/azd/pkg/apphost/generate.go
```diff
@@ -86,9 +86,7 @@ func init() {
 				"bicepParameterName": func(src string) string {
 					return strings.ReplaceAll(src, "-", "_")
 				},
-				"removeDot": func(src string) string {
-					return strings.ReplaceAll(src, ".", "")
-				},
+				"removeDot": scaffold.RemoveDotAndDash,
```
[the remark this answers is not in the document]
vhvb1989: not related to bindMounts directly, but to volumes in general.
When using `-` for the name, we were vulnerable to procuce a name with a `-` at the end when we use this bicep:

```bicep
name: take('${toLower('{{$name}}')}-${toLower('{{removeDot $volume.Name}}')}', 32)
```

Customers reported an issue on Aspire about deployiment failing b/c the name was 32 len but ending with `-`.  Removing the `-` makes the names uggly!  (I know), but it helped us to keep the `take(...)` function in the bicep.  

We should be able to change it all in the future, but remember this is all slowly moving to Aspire, so I though it was OK to have this tactical fix while we do the migration.

cli/azd/pkg/apphost/testdata/TestAspireContainerGeneration-main.bicep.snap
vhvb1989: it is the expected yes. 
It is a side effect of : https://github.com/Azure/azure-dev/pull/4286/files/4be3d0c087103452bff9c270a6bd21845cdf629b#r1785526924

cli/azd/resources/apphost/templates/resources.bicept
vhvb1989: I think it is fine.  Volumes are not very popular yet (b/c file system issues) and bindMounts are alpha.

davidfowl: We made some changes to aspire 9 that result in simmilar problems. We gave people a way to opt back into v8 naming schemes though. Not sure if it's worth it for volumes.

vhvb1989 (approved):

Issue #4275, opened 2024-08-30
Fail to run `azd pipeline config` on Codespaces


Menghua1:
**Describe the issue:**
When running the command `azd pipeline config`, it fails with the following error:
![image](https://github.com/user-attachments/assets/6022f512-dfe2-44ef-a09d-199212ea1edf)
According to the error prompt, run the command `gh auth login`, and encounter the following error:
![image](https://github.com/user-attachments/assets/1a11a787-3081-4f2c-9f72-89852f8bea69)

**Repro Steps:**

1. Open the Codespaces environment of the `todo-java-mongo` staging branch.
2. Install the daily version of azd.
3. Run `azd auth login`.
4. Run `azd pipeline conifg`.

**Environment:**

- Template: `todo-python-mongo-aca`, `todo-nodejs-mongo-swa-func`, `todo-java-mongo`.
- Azd version: `azd version 1.10.0-beta.1-daily.4086794 (commit 9445b193c91acfa2d4bcae129cb7471ecc101e11)`.
- OS: Codespaces.
- Branch: Staging.

**Expected behavior:**
`azd pipeline config` command can run successfully.

@rajeshkamal5050 for notification.

jongio:
Was GH removed from the template devcontainer?

rajeshkamal5050:
@vhvb1989 can you help triage this one?

vhvb1989:
@jongio , azd handles/manages `gh-cli`, It is automatically installed and used by azd.

This looks like a breaking change in the `not logged in` message on the gh-cli.  We recently updated the version azd uses, which could introduce this different error and broke azd :).  I'll take a look

vhvb1989:
Fix created

PR #4294, opened 2024-09-05
fix gh cli breaking change parsing err msg

vhvb1989:
Updating the expected message for gh-cli not-logged-in error after a regression on newer gh-cli version.
The wording was changed for newer version making azd to fail to recognize the error message. The change here make make azd
to know when this error happens again.

fix: https://github.com/Azure/azure-dev/issues/4275

### DIFF
--- a/cli/azd/pkg/tools/github/github_test.go
+++ b/cli/azd/pkg/tools/github/github_test.go
@@ -142,6 +142,45 @@ func TestNewGitHubCli(t *testing.T) {
 	require.Equal(t, Version.String(), ver)
 }
 
+func TestGetAuthStatus(t *testing.T) {
+	mockContext := mocks.NewMockContext(context.Background())
+
+	mockContext.CommandRunner.When(func(args exec.RunArgs, command string) bool {
+		return strings.Contains(args.Cmd, "gh") && len(args.Args) == 1 && args.Args[0] == "--version"
+	}).Respond(exec.NewRunResult(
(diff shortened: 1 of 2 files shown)


weikanglim (approved):

cli/azd/pkg/tools/github/github_test.go
```diff
@@ -142,6 +142,45 @@ func TestNewGitHubCli(t *testing.T) {
 	require.Equal(t, Version.String(), ver)
 }
 
+func TestGetAuthStatus(t *testing.T) {
```
weikanglim, line 145: If we feel that manually testing all the conditions is too tedious when we upgrade `gh`, we could create a set of contracts tests that tests against the "real live" gh and run againast all the known contracts as inputs and assert the expected azd behaviors.

Just want to highlight that this test will only catch that azd is working optimally in this code flow, but will not catch contractual changes if gh changes their outputs again.

vhvb1989: I was thinking how to create such contracts, but running a real live gh cli will have some challenges to control its state when running on CI/CD.  We might not want to logout when gh cli is logged in automatically, and things like that.

So, having at least a little coverage for now was my starting point.

Since we use the stdout/stderr from many tools as contracts, we might want to think on how to protect ourself from breaking changes like this one, but without manipulating the state around the tool too much

PR #4298, opened 2024-09-05
vs fix resource manager

vhvb1989:
Proposal ...


cli/azd/internal/vsrpc/environment_service_refresh.go
```diff
@@ -114,7 +114,8 @@ func (s *environmentService) refreshEnvironmentAsync(
 
 	_ = observer.OnNext(ctx, newInfoProgressMessage("Loading server resources"))
 
-	rgName, err := c.azureResourceManager.FindResourceGroupForEnvironment(ctx, subId, envName)
+	azureResourceManager := infraP.NewAzureResourceManager(c.resourceService, c.deploymentService)
```
wbreza, line 117: I'd prefer if we standardized and injected this from the container.

PR #4303, opened 2024-09-06
chore: Update comments in `launch.json`

ellismg:
I was talking with Marina today about my workflow for debugging changes in `azd`. I use the "Debug azd cli" launch configuration for most of my debugging and end up modifying `launch.json` to set the `cwd` and other parameters for the profile for the specific end to end I want to debug.  There are some `git` tricks that you can play to ignore modifications to this file (so you don't have to worry about checking them in).  I had documented them in one place but not the other (since I typically run VS Code from within the `cli/azd` sub folder, to have a smaller workspace) and didn't document how you undo this behavior.

This patch updates the comments to show how to turn this off, and adds them to our root `launch.json` as well.


weikanglim (approved):
Thanks @ellismg and @hemarina !

This really got me thinking about what I would want from an ideal DX experience, and I think I would want:

- Remember the last input I used
- Source the config from a on-disk file

I was about to ask @bwateratmsft  in case he knew if VSCode has support natively. Turns out that VSCode is implementing the prompt via [variable references](https://code.visualstudio.com/docs/editor/variables-reference), so I believe it isn't supported natively (based on the generic implementation).

The good news is that there is a VSCode extension, aptly named, [command variable](https://marketplace.visualstudio.com/items?itemName=rioj7.command-variable) that supports this! If we used `extension.commandvariable.promptStringRemember`, it would remember the last prompt.

The extension also supports a whole lot of options, including `extension.commandvariable.file.content`, JSONPath, JS expressions. So I imagine we could do a load from a git-ignored file equally as well.

I vaguely remember @vhvb1989 saying at one point he would love a more interactive prompt (perhaps picking from a command tree). I suspect that this would be possible as well if we `go generate` the command layout and do something clever here.

I do think this is a huge productivity boost that may be worth adding it as an extension pre-req for debugging. According to the docs, it would work as well on `web` (which I understood as VSCode remote).

vhvb1989:
My `launch.json` looks like this:

image

XD

I would be nice to remember you last prompt, but keeping all the history to pick from would be my ideal scenario @weikanglim :)

By keeping all the commands I have debugged in the past commented, I can easily re-use them when needed.

I am very used to have the 2 files in the changes tab
image

Since the 2 files are always at the top, I've learnt to remove them from the stage befome commit. I even feel like something's missing if I don't see the files there LOL

PR #4314, opened 2024-09-13
WIP: provisioning: Consider `resourceGroup` from `azure.yaml`

### LINKED ISSUE #4287: resourceGroup in azure.yaml not honored for rg-scoped deployment
1. `azd env new` to create a new environment
1. Set `resourceGroup: <some group>` in `azure.yaml`.
2. Run `azd provision`

Expected:
The `resourceGroup` property is used.

Actual:
I get prompted for a resource group to be selected.

ellismg:
We allow specifying the resource group which is used for discovery of resources (instead of probing for a resource group taged with `azd-env-name` inside `azure.yaml`. When doing a resource group scoped deployment, we should consider this value if set and use it instead of prompting the user to select a resource group or create a new one.

Fixes #4287

### DIFF
--- a/cli/azd/pkg/infra/provisioning/bicep/bicep_provider.go
+++ b/cli/azd/pkg/infra/provisioning/bicep/bicep_provider.go
@@ -36,6 +36,7 @@ import (
 	"github.com/azure/azure-dev/cli/azd/pkg/infra/provisioning"
 	"github.com/azure/azure-dev/cli/azd/pkg/input"
 	"github.com/azure/azure-dev/cli/azd/pkg/keyvault"
+	"github.com/azure/azure-dev/cli/azd/pkg/osutil"
 	"github.com/azure/azure-dev/cli/azd/pkg/output"
 	"github.com/azure/azure-dev/cli/azd/pkg/output/ux"
 	"github.com/azure/azure-dev/cli/azd/pkg/password"
@@ -91,7 +92,12 @@ func (p *BicepProvider) RequiredExternalTools() []tools.ExternalTool {
 
 // Initialize initializes provider state from the options.
 // It also calls EnsureEnv, which ensures the client-side state is ready for provisioning.
-func (p *BicepProvider) Initialize(ctx context.Context, projectPath string, options provisioning.Options) error {
+func (p *BicepProvider) Initialize(
+	ctx context.Context,
+	projectPath string,
+	options provisioning.Options,
+	rg osutil.ExpandableString,
+) error {
 	p.projectPath = projectPath
 	p.options = options
 	if p.options.Module == "" {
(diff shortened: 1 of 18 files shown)


cli/azd/pkg/infra/provisioning/provider.go
```diff
@@ -52,10 +54,9 @@ type StateResult struct {
 
 type Provider interface {
 	Name() string
-	Initialize(ctx context.Context, projectPath string, options Options) error
+	Initialize(ctx context.Context, projectPath string, options Options, rg osutil.ExpandableString) error
```
ellismg, line 57: I initially considered trying to have `Initialize` take the entire `project.ProjectConfig` object so it could grab this property, but that doesn't work because it creates an import cycle between the `project` and `provisioning` packages.

Sort of a bummer that the `resourceGroup` property doesn't hang off of the `provisioning.Options` type.  It probably would have made more sense to nest that under `infra` vs it being a top level property.

cli/azd/pkg/infra/provisioning/terraform/terraform_provider.go
```diff
@@ -76,7 +76,12 @@ func NewTerraformProvider(
 	return provider
 }
 
-func (t *TerraformProvider) Initialize(ctx context.Context, projectPath string, options provisioning.Options) error {
+func (t *TerraformProvider) Initialize(
+	ctx context.Context,
+	projectPath string,
+	options provisioning.Options,
+	_ osutil.ExpandableString,
```
ellismg, line 83: Wondering what if anything we should do here for Terraform, perhaps we should be adding this to `envVars` with something like `AZURE_RESOURCE_GROUP` so `tf` could observe it?

cli/azd/pkg/infra/provisioning/bicep/bicep_provider.go
```diff
@@ -163,6 +169,25 @@ func (p *BicepProvider) EnsureEnv(ctx context.Context) error {
...
+
+		if rgEval != "" {
+			// TODO(ellismg): We set this here because the rest of the system reads from `.env` when it needs
+			// AZURE_RESOURCE_GROUP. But I'm not sure this is exactly what we want to be doing here - it feels wrong
+			// on some level that changing the value in `azure.yaml` ends up replacing what was in the `.env` file. It seems
+			// like if it is explicitly set in `azure.yaml` than we do not need to store it in the `.env` file at all.
+			//
+			// We also need to understand the policy of what happens if both are set - which one wins?  I think that perhaps
+			// it should be `AZURE_RESOURCE_GROUP` since that also allows overriding via `AZURE_RESOURCE_GROUP=foo azd up`.
+			p.env.DotenvSet(environment.ResourceGroupEnvVarName, rgEval)
```
ellismg, line 185: This is pretty gross, IMHO, but the rest of the system right now reads `environment.ResourceGroupEnvVarName` from the `.env` to find the resource group to use.  I suspect that I need to do a deeper refactoring such that we can consider both what is in the `azure.yaml` file and what is in the `.env` but right now doing things this way lets the rest of the system work.

But there are open questions, hence the big TODO and the draft nature of this PR...

Issue #4311, opened 2024-09-12
Can not display option and select option when run `azd init` by GoLand


rujche:
**Describe the problem**
Can not display option and select option when run by [GoLand](https://www.jetbrains.com/go/)

**To Reproduce**
1. Use GoLand to open current project.
2. Run `cli/azd/main.go` by GoLand.
3. It outputs `Initializing an app to run on Azure (azd init)  How do you want to initialize your app? `, but no option displayed.

> ![image](https://github.com/user-attachments/assets/871425a1-083d-4361-8615-476cbc5ea902)

**Expected behavior**
Options displayed and can be selected.

**Environment**
Information on your environment:
 - **Language name and version**: go version go1.23.1 windows/amd64
 - **IDE and version**: GoLand 2024.2.1.1
 - **Device information**:
  > ![image](https://github.com/user-attachments/assets/5ce26105-13a1-47a0-9d55-40b4c1e0daeb)
 - **Go Settings in GoLand**:
  > ![image](https://github.com/user-attachments/assets/baba1c93-7c95-489b-b1be-4510c227501d)
  > ![image](https://github.com/user-attachments/assets/9bfd3127-f75f-4684-8ce1-8543bbd429d7)
  > ![image](https://github.com/user-attachments/assets/e030cc32-7949-4cf5-9802-ec2376ae8813)
  > ![image](https://github.com/user-attachments/assets/1a6f9802-482f-47ba-9a6e-a6363e6d54d4)

weikanglim:
@rujche I haven't looked into how the `goland` terminal is implemented, but the auto-TTY detection behavior isn't working as expected. 

With #4318, we'll give you the ability to specify `AZD_FORCE_TTY=true` to ignore the terminal detection logic. The private build should be available for you to test soon. I'll reply back when that's available. 

If you have spare time and would love to help us, you can run the code snippet from the library we're using to investigate why the terminal detection logic isn't working as expected -- and what we may want to contribute upstream (issue or code) to fix this behavior.

```golang
package main

import (
	"fmt"
	"github.com/mattn/go-isatty"
	"os"
)

func main() {
	if isatty.IsTerminal(os.Stdout.Fd()) {
		fmt.Println("Is Terminal")
	} else if isatty.IsCygwinTerminal(os.Stdout.Fd()) {
		fmt.Println("Is Cygwin/MSYS2 Terminal")
	} else {
		fmt.Println("Is Not Terminal")
	}
}
```

[...]

PR #4318, opened 2024-09-16
Make TTY output a forceable thing

weikanglim:
In this change, we introduce `AZD_FORCE_TTY` as a mechanism to force TTY output behavior.

Helps with #4311

### DIFF
--- a/cli/azd/pkg/input/console.go
+++ b/cli/azd/pkg/input/console.go
@@ -141,8 +141,15 @@ type AskerConsole struct {
 	// the writer the console was constructed with, and what we reset to when SetWriter(nil) is called.
 	defaultWriter io.Writer
 	// the writer which output is written to.
-	writer     io.Writer
-	formatter  output.Formatter
+	writer    io.Writer
+	formatter output.Formatter
+
+	// isTerminal controls whether terminal-style input/output will be used.
+	//
+	// When isTerminal is false, the following notable behaviors apply:
+	//   - Spinner progress will be written as standard newline messages.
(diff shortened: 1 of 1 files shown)


cli/azd/pkg/input/console.go
```diff
@@ -1056,9 +1056,9 @@ func NewConsole(
 // IsTerminal returns true if the given file descriptors are attached to a terminal,
 // taking into account of environment variables that force TTY behavior.
 func IsTerminal(stdoutFd uintptr, stdinFd uintptr) bool {
-	// User override to force non-TTY behavior
-	if ok, _ := strconv.ParseBool(os.Getenv("AZD_DEBUG_FORCE_NO_TTY")); ok {
-		return false
+	// User override to force TTY behavior
+	if forceTty, err := strconv.ParseBool(os.Getenv("AZD_FORCE_TTY")); err == nil {
```
weikanglim, line 1060: Perhaps it'd be nice to also make the TTY column width specifiable, here's what our friends at `gh` did for inspiration: https://github.com/cli/cli/pull/4146/files

ellismg (approved):
LGTM

wbreza (approved):

cli/azd/pkg/input/console.go
weikanglim: We don't have any output that is width-dependent currently (aside from the spinner). The spinner itself writes new-line messages when TTY-behavior is off. So there isn't a need right now.

In the future, perhaps we could still have this as an integer, but `1` or `0` is treated as boolean but other numbers are treated as width. Or we could introduce units: `1000w` and `200%`.

PR #4325, opened 2024-09-16
Allow persisting CORS for ACA deployments (applies to .NET Aspire)

vhvb1989:
fix: https://github.com/dotnet/aspire/issues/5029

Turn featue on with
`azd config set alpha.aca.persistCors on`

fix [#4395](https://github.com/Azure/azure-dev/issues/4395)


cli/azd/pkg/containerapps/container_app.go
```diff
@@ -127,6 +128,7 @@ const apiVersionKey = "api-version"
 
 var persistCustomDomainsFeature = alpha.MustFeatureKey("aca.persistDomains")
 var persistIngressSessionAffinity = alpha.MustFeatureKey("aca.persistIngressSessionAffinity")
+var persistCors = alpha.MustFeatureKey("aca.persistCors")
```
davidfowl, line 131: We're going to end up with like 50 of these 😄

vhvb1989: Hopefully we will have the AppHost producing the `ACA definition` soon and get rid of all of these.

davidfowl:
I think we should look at preserving all of the data from the remote but only in the case where infra synth isn't run.

ellismg:
> I think we should look at preserving all of the data from the remote but only in the case where infra synth isn't run.

@davidfowl Have you thought about how this is going to interact with the work you've started on for compute customization?  It feels like if we do this it is going to force you to fetch the current state of each container app during manifest publishing and then merge it with the desired state from the user.  Is that something you want?

Another thing to note, the shape of the object in azure is dependent on the API version you use when you send your `GET`.  Certain properties may not be present depending on the API version you use.

My gut says you're walking into a quagmire trying to do a diff and merge here...

cli/azd/pkg/containerapps/container_app.go
ellismg: Maybe the alpha feature should be something like: "give us the paths of keys that you want preserved" instead of adding a switch per property?  That would be a little more generalized and would at least force folks to think about the underlying container app object we are building.

wbreza: I'm not as familiar with many of the Aspire use cases - but why do we need to explicitly call out parts of the configuration vs taking the snapshot of the current version, then layer the yaml manifest config to form the final container app spec to update?

wbreza: When I mentioned snapshot I mean the call to getContainerApp that is already happening during deployment.  Then you additional have the yaml manifest that needs to get merged in.

vhvb1989: getContainerApp is not part of deployment.  We call it only if we want to persist something:

image

wbreza: I guess I'm not fully understanding the use case.  Why wouldn't we persist all of the configuration data that was set during the app provisioning?

weikanglim: What we're doing here is ignoring IaC as the source-of-truth to allow users to overcome a current bottleneck in the design (i.e., ACA resource definition not being full representable at the AppHost layer):
- Bicep/IaC is usually the source-of-truth of the Azure resource definition.
- But here, we are saying "let's allow the user to tweak this value in Azure, and we will just ignore Bicep when we reprovision"

> "give us the paths of keys that you want preserved" instead of adding a switch per property

I think this would be most flexible. Can we make a `config.json` property for a `.JSONPath` to ignore?

```json
"aspire.containerapp.unmanaged.properties": [
   "properties.configuration.ingress.corsPolicy",
   "properties.configuration.ingress.stickySessions"
]
```

We would just need better docs to show how to do all these things. We could still make the alpha feature trigger such a thing underneath, but it'd perhaps make the code churn easier?

wbreza: Thanks for providing more context here @weikanglim.  I'm just not convinced that ignoring IaC as the source of truth is the right choice here.  

I'm all for users experimenting with configuration in the Azure Portal, but if the user wants to keep those changes they should go back and make the appropriate tweaks in their IaC  / or manifest, etc.

Is the root cause here that the current AppHost layer does not support all the knobs required to set/persist these types of settings?

weikanglim: In case it wasn't clear from the previous message -- yes, this is a point-in-time thing. Aspire will eventually surface these knobs, but it requires modeling it at the AppHost layer, which I could imagine takes time.

wbreza (approved):

vhvb1989:
Closing this as not needed anymore. The path to customizing ACA is set for Aspire9 by using the AppHost + Azure Provisioning libraries

PR #4329, opened 2024-09-17
add Preflight Validate API support

### LINKED ISSUE #468: [Issue] Help developers recover from capacity errors on free accounts
image

Steps to repro:
1. Create a free account 
2. Pick a template (I believe Mandy used a node template)
3. Choose East US 2 as region (apparently Central US region worked)

hemarina:
Adds support for Preflight Validate API.

- [x] Fix #468 
- [ ] Related bug: https://github.com/Azure/azure-sdk-for-go/issues/23350
- [x] UX enhancement will be follow up in https://github.com/Azure/azure-dev/issues/4365

### What is Preflight?
Preflight API validates whether the specified template is syntactically correct and will be accepted by Azure Resource Manager. More information on [preflight wiki](https://armwiki.azurewebsites.net/api_contracts/preflight.html). 

### Why do we implement Validate API?
There’re two preflight API: Validate API and Deployment API. We implemented [deployment API](https://github.com/Azure/azure-dev/blob/main/cli/azd/pkg/azapi/standard_deployments.go#L253). However, deployment API creates a scenario where specific resource fails at preflight API and stops its deployment where some resources are already created. For example, resources are created at an unsupported location. This will cause more dev-test loop for cleaning resources and re-deploy. Adding Validate API support before provisioning to azure solves this situation. 

### What are Preflight Validate API limitations?
1. Around 60% applicable resource types are enabled with preflight. Check [preflight coverage dashboard](https://aka.ms/preflightreport). 
2. Preflight is not able to [validate nested resources](https://armwiki.azurewebsites.net/api_contracts/preflight.html#preflight-and-nested-templates) with parameters referencing runtime functions.

### Standard deployments
image

### Deployment stack
image


cli/azd/pkg/infra/provisioning/bicep/bicep_provider.go
```diff
@@ -561,6 +561,24 @@ func (p *BicepProvider) Deploy(ctx context.Context) (*provisioning.DeployResult,
...
+		deploymentTags[azure.TagKeyAzdDeploymentStateParamHashName] = to.Ptr(currentParamsHash)
+	}
+
+	err = p.validatePreflight(
+		ctx,
+		bicepDeploymentData.Target,
+		bicepDeploymentData.CompiledBicep.RawArmTemplate,
+		bicepDeploymentData.CompiledBicep.Parameters,
+		deploymentTags,
+	)
```
wbreza, line 577: I think it is great that the validate API exists but I am concerned that this may cause considerable time delay during deployments.

Do we have an idea of the amount of extra time the validations will take especially against more complex & nested deployments?

weikanglim: @hemarina and I had spoke awhile back about this while she was implementing -- I think this is pretty lightweight (<3s) from what she showed me, but it'd be great to test this out and gather the numbers with a few different templates.

cli/azd/pkg/azapi/stack_deployments.go
```diff
@@ -634,3 +634,124 @@ func convertFromStacksProvisioningState(
...
+	clonedTags[azure.TagKeyAzdDeploymentTemplateHashName] = &templateHash
+
+	stackParams := map[string]*armdeploymentstacks.DeploymentParameter{}
+	for k, v := range parameters {
+		stackParams[k] = &armdeploymentstacks.DeploymentParameter{
+			Value: v.Value,
+		}
+	}
+
+	deleteBehavior := armdeploymentstacks.DeploymentStacksDeleteDetachEnumDelete
```
weikanglim, line 667: FYI @hemarina , @wbreza had merged some changes here. I think we can reuse the `parseDeploymentStackOptions` function for getting the deployment stack options

wbreza (commented):
What is the expectation for templates with nested deployments?  For example, I used our Todo AKS template, specified an invalid K8s version parameter and expected that this would fail at our own preflight check call. 

Instead the preflight from `azd` passes, but then we get a preflight failure during ARM deployment.

```
Preflight validation check for resource(s) for container service aks-oc3hr2l7nqlee in resource group rg-wabrez-todo-aks-swed failed. Message: Web application routing requires at least Kubernetes 1.22. Details: 

 (Code: WebAppRoutingUnsupportedK8SVersion)
```

[Failing Deployment Example](https://ms.portal.azure.com/#view/HubsExtension/DeploymentDetailsBlade/~/overview/id/%2Fsubscriptions%2Ffaa080af-c1d8-40ad-9cce-e1a450ca5b57%2FresourceGroups%2Frg-wabrez-todo-aks-swed%2Fproviders%2FMicrosoft.Resources%2Fdeployments%2Faks)

If it is not going to catch preflight errors on nested deployments I question whether it is valuable to add at this point.

cli/azd/pkg/azapi/standard_deployments.go
```diff
@@ -688,3 +691,129 @@ func convertFromStandardProvisioningState(state armresources.ProvisioningState)
...
+	tags map[string]*string,
+	options map[string]any,
+) error {
+	deploymentClient, err := ds.createDeploymentsClient(ctx, subscriptionId)
+	if err != nil {
+		return fmt.Errorf("creating deployments client: %w", err)
+	}
+
+	var rawResponse *http.Response
+	ctxWithResp := runtime.WithCaptureResponse(ctx, &rawResponse)
```
wbreza, line 711: Do we need to do this?  If we get an HttpError we should be able to case the error to ARM http error to parse the error message.

cli/azd/pkg/azapi/standard_deployments.go
```diff
@@ -688,3 +691,129 @@ func convertFromStandardProvisioningState(state armresources.ProvisioningState)
...
+			},
+			Tags: tags,
+		}, nil)
+	if err != nil {
+		return validatePreflightError(rawResponse, err, "resource group")
+	}
+
+	_, err = validate.PollUntilDone(ctx, nil)
+	if err != nil {
+		deploymentError := createDeploymentError(err)
```
wbreza, line 811: We should use this same `createDeploymentError()` func in the deployment stack use cases as well.

cli/azd/pkg/azapi/standard_deployments.go
```diff
@@ -688,3 +691,129 @@ func convertFromStandardProvisioningState(state armresources.ProvisioningState)
...
+type PreflightErrorResponse struct {
+	Error struct {
+		Code    string `json:"code"`
+		Message string `json:"message"`
+		Details []struct {
+			Code    string `json:"code"`
+			Message string `json:"message"`
+		} `json:"details"`
+	} `json:"error"`
+}
```
wbreza, line 749: It looks like this basically captures the same ARM errors that we use in the `AzureDeploymentError` object that we already have.  Can we reuse that one instead?

cli/azd/pkg/azapi/stack_deployments.go
```diff
@@ -242,42 +244,11 @@ func (d *StackDeployments) DeployToSubscription(
...
-
-	stackParams := map[string]*armdeploymentstacks.DeploymentParameter{}
-	for k, v := range parameters {
-		stackParams[k] = &armdeploymentstacks.DeploymentParameter{
-			Value: v.Value,
-		}
-	}
-
-	deploymentStackOptions, err := parseDeploymentStackOptions(options)
+	client, stack, err := d.validateOrDeployToSubscriptionHelper(ctx, subscriptionId, location, armTemplate, parameters, tags, options)
```
weikanglim, line 247: I'm wondering if you may think it's a good idea to separate it along the lines of:

```golang
client, err := d.createClient(ctx, subscriptionId)
if err != nil {
  return nil, err
}

stack, err := d.stackFromArm(ctx, subscriptionId, location, armTemplate, parameters, tags, options)
if err != nil {
  return nil, err
}
```

cli/azd/pkg/azapi/standard_deployments.go
```diff
@@ -688,3 +691,106 @@ func convertFromStandardProvisioningState(state armresources.ProvisioningState)
...
+) error {
+	deploymentClient, err := ds.createDeploymentsClient(ctx, subscriptionId)
+	if err != nil {
+		return fmt.Errorf("creating deployments client: %w", err)
+	}
+
+	var rawResponse *http.Response
+	ctxWithResp := runtime.WithCaptureResponse(ctx, &rawResponse)
+
+	_, err = deploymentClient.BeginValidateAtSubscriptionScope(
```
weikanglim, line 713: Just curious, do we need to poll here for the Validate completion?

cli/azd/pkg/azapi/standard_deployments.go
```diff
@@ -691,3 +694,111 @@ func convertFromStandardProvisioningState(state armresources.ProvisioningState)
...
+	_, validateError := validateResult.PollUntilDone(ctx, nil)
+
+	if validateError != nil || err != nil {
+		return validatePreflightError(rawResponse, validateError, err, "subscription")
+	}
+
+	return nil
+}
+
+type PreflightErrorResponse struct {
```
weikanglim, line 736: I think this can be removed if it's unused

cli/azd/pkg/azapi/standard_deployments.go
```diff
@@ -691,3 +694,108 @@ func convertFromStandardProvisioningState(state armresources.ProvisioningState)
...
+	rawResponse *http.Response,
+	err error,
+	typeMessage string,
+) error {
+	var respErr *azcore.ResponseError
+	if errors.As(err, &respErr) {
+		err = createDeploymentError(err)
+	} else if err != nil {
+		// Error returned from azure sdk go bug
+		// https://github.com/Azure/azure-sdk-for-go/issues/23350
```
weikanglim, line 748: Would some additional context here help a person new to the codebase here?
```suggestion
		// Error returned from azure sdk go bug: we receive a 400 Bad Request from the API, but the client-handling in azure sdk fails internally with a different error 
		// This special-cased handling and rawResponse capture can be removed once https://github.com/Azure/azure-sdk-for-go/issues/23350 is fixed
```

cli/azd/pkg/azapi/standard_deployments.go
```diff
@@ -691,3 +694,108 @@ func convertFromStandardProvisioningState(state armresources.ProvisioningState)
...
+		if rawResponse == nil || rawResponse.StatusCode != 400 {
```
weikanglim, line 749: I suspect that this if-condition may need to be tweaked slightly from the previous iteration -- here we're trying to execute the if-block only if we detect that a 400 response is returned

```suggestion
		if rawResponse != nil && rawResponse.StatusCode == 400 {
```

cli/azd/pkg/azapi/stack_deployments.go
```diff
@@ -767,3 +765,111 @@ func convertToStackParams(parameters azure.ArmParameters) map[string]*armdeploym
...
+		return err
+	}
+
+	stack, err := d.stackFromArmForResourceGroup(ctx, subscriptionId, armTemplate, parameters, tags, options)
+	if err != nil {
+		return err
+	}
+
+	var rawResponse *http.Response
+	ctxWithResp := runtime.WithCaptureResponse(ctx, &rawResponse)
```
JeffreyCA, line 790: I'm seeing that this method is deprecated, does `policy.WithCaptureResponse` work here instead? I see we use it in `standard_deployments.go`.

> WithCaptureResponse applies the HTTP response retrieval annotation to the parent context. The resp parameter will contain the HTTP response after the request has completed. Deprecated: use [policy.WithCaptureResponse] instead.

cli/azd/pkg/azapi/stack_deployments.go
```diff
@@ -767,3 +765,111 @@ func convertToStackParams(parameters azure.ArmParameters) map[string]*armdeploym
...
+
+	validateResult, err := client.BeginValidateStackAtResourceGroup(ctxWithResp, resourceGroup, deploymentName, stack, nil)
+	if err != nil {
+		return validatePreflightError(rawResponse, err, "resource group")
+	}
+	_, err = validateResult.PollUntilDone(ctx, nil)
```
JeffreyCA, line 796: Not very familiar with contexts in go - how come we're using `ctx` here instead of `ctxWithResp`?

weikanglim: 🙂 Great inquiry @JeffreyCA ! I suspect we may need to use `ctxWithResp` for this to be handled correctly.

go [context](https://pkg.go.dev/context) are immutable data structures (for concurrency reasons -- the main use is to signal cancellation and deadlines/timeouts across goroutines), so with each `WithContext`, you get back a new context with added values in the context property bag.

I think it's perfectly okay for us to shadow the `ctx` on line 790 here, i.e. `ctx = policy.WithCaptureReseponse(ctx, &rawResponse)` works perfectly fine in my mind.

JeffreyCA: I see, thank you both for clarifying! 🙂

JeffreyCA (approved):
Thanks for landing these changes - this will bring significant improvements especially to AI templates as the preflight changes roll out everywhere

cli/azd/pkg/azapi/standard_deployments.go
```diff
@@ -691,3 +694,110 @@ func convertFromStandardProvisioningState(state armresources.ProvisioningState)
...
+			if errOnRawResponse != nil {
+				return fmt.Errorf("failed to read response error body from preflight validation api to %s: %w",
+					typeMessage, errOnRawResponse)
+			}
+
+			err = NewAzureDeploymentError(string(body))
+		}
+	}
+	return fmt.Errorf(
+		"validating preflight to %s:\n\nPreflight Error Details:\n%w",
```
weikanglim, line 763: Would something like this look any better?

"validating deployment:\n\nValidation errors:\n"

Issue #4307, opened 2024-09-10
Deployment Stacks: Support Customization of Deny and Detach settings


ellismg:
Today we don't allow customization of the deny or detach settings for the deployment stack that is created or used when you deploy using deployment stacks.

We should allow customization of these settings (probably via `azure.yaml`) before we advance the feature from alpha to beta.
```[tasklist]
### Tasks
- [x] Add optional `deploymentStacks` configuration section in infra provider
- [x] Unit tests
- [x] Update azure.yaml schema
```

PR #4331, opened 2024-09-17
Deployment Stacks: Expose configurable settings for `actionOnUnmanage` and `denySettings`

wbreza:
Add new **optional** `deploymentStacks` section for the infra provider to pass settings for deployment stacks.

Since `azd` supports pluggable infrastructure providers the `deploymentStacks` property within the `infra` section of the `azure.yaml` is lazily validated during provision command.

## Example Usage:

The following is an example showing custom configuration options for deployment stacks.

```yaml
# azure.yaml

name: storage
metadata: 
  template: azd-test/storagetest@v1
infra:
  provider: bicep
  deploymentStacks:
    actionOnUnmanage: 
      resourceGroups: delete      # delete or detach
      resources: delete           # delete or detach
    denySettings:
      mode: denyDelete            # denyDelete, denyWriteAndDelete, none
      applyToChildScopes: false   # true | false
      excludedActions:            # List of actions to exclude
        - action1
        - action2
      excludedPrincipals:         # List of principals to exclude
        - principal1
        - principal2
```

## Default Configuration

The following is the internal defaults for deployment stack options when not overridden within the `azure.yaml` file.

**bypassStackOutOfSyncError** : false
**actionOnUnmanage**: `delete` for all types
**denySettings**: `nil` (No deny settings by default)

## Overrides

Users can override `bypassStackOutOfSyncError` by setting OS env var `DEPLOYMENT_STACKS_BYPASS_STACK_OUT_OF_SYNC_ERROR` to a truthy value.

## Resources

See [official deployment stacks docs](https://learn.microsoft.com/en-us/azure/azure-resource-manager/bicep/deployment-stacks) for full list of available configuration options.

[More Docs](https://learn.microsoft.com/en-us/azure/templates/microsoft.resources/deploymentstacks?pivots=deployment-language-bicep)

### DIFF
--- a/cli/azd/pkg/azapi/stack_deployments.go
+++ b/cli/azd/pkg/azapi/stack_deployments.go
@@ -524,17 +524,17 @@ func parseDeploymentStackOptions(options map[string]any) (*deploymentStackOption
 
 	var deploymentStackOptions *deploymentStackOptions
 	has, err := optionsConfig.GetSection(deploymentStacksConfigKey, &deploymentStackOptions)
-	if !has {
-		return defaultDeploymentStackOptions, nil
-	}
-
 	if err != nil {
 		suggestion := &internal.ErrorWithSuggestion{
 			Err:        fmt.Errorf("failed parsing deployment stack options: %w", err),
-			Suggestion: "Review the 'infra.config.deploymentStacks' configuration section in the 'azure.yaml' file.",
+			Suggestion: "Review the 'infra.deploymentStacks' configuration section in the 'azure.yaml' file.",
 		}
 
 		return nil, suggestion
+	}
+
+	if !has || deploymentStackOptions == nil {
+		return defaultDeploymentStackOptions, nil
 	}
 
 	if deploymentStackOptions.BypassStackOutOfSyncError == nil {
--- a/cli/azd/pkg/azapi/stack_deployments.go
+++ b/cli/azd/pkg/azapi/stack_deployments.go
@@ -18,6 +18,7 @@ import (
 	"github.com/azure/azure-dev/cli/azd/pkg/async"
 	"github.com/azure/azure-dev/cli/azd/pkg/azure"
 	"github.com/azure/azure-dev/cli/azd/pkg/cloud"
+	"github.com/azure/azure-dev/cli/azd/pkg/config"
 	"github.com/azure/azure-dev/cli/azd/pkg/convert"
 	"github.com/azure/azure-dev/cli/azd/pkg/output"
 	"github.com/benbjohnson/clock"
@@ -27,14 +28,31 @@ import (
 var FeatureDeploymentStacks = alpha.MustFeatureKey("deployment.stacks")
 
 const (
-	stacksPortalUrlFragment = "#@microsoft.onmicrosoft.com/resource"
+	deploymentStacksConfigKey = "deploymentStacks"
+	stacksPortalUrlFragment   = "#@microsoft.onmicrosoft.com/resource"
 )
+
+var defaultDeploymentStackOptions = &deploymentStackOptions{
+	BypassStackOutOfSyncError: to.Ptr(false),
+	ActionOnUnmanage: &armdeploymentstacks.ActionOnUnmanage{
+		ManagementGroups: to.Ptr(armdeploymentstacks.DeploymentStacksDeleteDetachEnumDelete),
+		ResourceGroups:   to.Ptr(armdeploymentstacks.DeploymentStacksDeleteDetachEnumDelete),
+		Resources:        to.Ptr(armdeploymentstacks.DeploymentStacksDeleteDetachEnumDelete),
+	},
+	DenySettings: nil,
+}
 
 type StackDeployments struct {
 	credentialProvider  account.SubscriptionCredentialProvider
 	armClientOptions    *arm.ClientOptions
 	standardDeployments *StandardDeployments
 	cloud               *cloud.Cloud
+}
+
+type deploymentStackOptions struct {
+	BypassStackOutOfSyncError *bool                                 `yaml:"bypassStackOutOfSyncError,omitempty"`
+	ActionOnUnmanage          *armdeploymentstacks.ActionOnUnmanage `yaml:"actionOnUnmanage,omitempty"`
+	DenySettings              *armdeploymentstacks.DenySettings     `yaml:"denySettings,omitempty"`
 }
 
 func NewStackDeployments(
@@ -215,6 +233,7 @@ func (d *StackDeployments) DeployToSubscription(
 	armTemplate azure.RawArmTemplate,
 	parameters azure.ArmParameters,
 	tags map[string]*string,
+	options map[string]any,
 ) (*ResourceDeployment, error) {
 	client, err := d.createClient(ctx, subscriptionId)
 	if err != nil {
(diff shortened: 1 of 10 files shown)


cli/azd/pkg/infra/provisioning/provider.go
```diff
@@ -19,9 +19,10 @@ const (
 )
 
 type Options struct {
-	Provider ProviderKind `yaml:"provider,omitempty"`
-	Path     string       `yaml:"path,omitempty"`
-	Module   string       `yaml:"module,omitempty"`
+	Provider ProviderKind   `yaml:"provider,omitempty"`
+	Path     string         `yaml:"path,omitempty"`
+	Module   string         `yaml:"module,omitempty"`
+	Config   map[string]any `yaml:"config,omitempty"`
```
weikanglim, line 25: There are some tricks we can play in both JSON Schema for the YAML LSP, and go marshal/unmarshalling to avoid an extra `config` node in YAML for known union types. I can take a half hour to pull it out of my feature implementation [PR](https://github.com/Azure/azure-dev-pr/pull/1673/files#diff-b53ec5cd2fa533b7cc3ca2ff6715111a330bac7ba080efa8ec6cab3382ef4ea7), but just wanting to see a pulse if there's any interests here from a DX perspective.

So instead of:
```yaml
infra:
  provider: bicep
  config:
    deploymentStacks:
```

We would have:
```yaml
infra:
  provider: bicep
  deploymentStacks:
```

@ellismg @wbreza @vhvb1989

weikanglim: But I see that we're imagining these are "dynamic passthrough information", so perhaps what we have here suffices.

wbreza: I do actually feel like the `config` node works well for these provider based configurations but since the `infra` section existed before adopted this model I am happy to not introduce the `config` node here.

I was able to remove the extra layer without anything to fancy.

cli/azd/pkg/infra/provisioning/provider.go
```diff
@@ -19,9 +19,10 @@ const (
 )
 
 type Options struct {
-	Provider ProviderKind `yaml:"provider,omitempty"`
-	Path     string       `yaml:"path,omitempty"`
-	Module   string       `yaml:"module,omitempty"`
+	Provider         ProviderKind   `yaml:"provider,omitempty"`
+	Path             string         `yaml:"path,omitempty"`
+	Module           string         `yaml:"module,omitempty"`
+	DeploymentStacks map[string]any `yaml:"deploymentStacks,omitempty"`
```
weikanglim, line 25: Given that we parse it into a concrete struct `azapi.deploymentStackOptions` at a later point in time of execution, would we prefer to simply do the parsing into a structured object here?

I would prefer to avoid `options map[string]any` in `bicep_provider.go` -- Bicep provider clearly knows what the options *should* look like, perhaps we would not want this type of error to be caught so late?

weikanglim, line 25: Would `terraform` know about `deploymentStacks` here in provisioning_provider.go? I think we would want a generic-props that is translated into a structured-props. See example of `RawProps` vs `Props` an implementation [here](https://github.com/Azure/azure-dev-pr/blob/b5f96606635e53819ee4fdb1a21337c9642bd622/cli/azd/pkg/project/resources.go#L84).

cli/azd/pkg/azapi/stack_deployments.go
```diff
@@ -27,16 +29,35 @@ import (
...
 type StackDeployments struct {
 	credentialProvider  account.SubscriptionCredentialProvider
 	armClientOptions    *arm.ClientOptions
 	standardDeployments *StandardDeployments
 	cloud               *cloud.Cloud
 }
 
+type deploymentStackOptions struct {
+	BypassStackOutOfSyncError *bool                                 `yaml:"bypassStackOutOfSyncError,omitempty"`
+	ActionOnUnmanage          *armdeploymentstacks.ActionOnUnmanage `yaml:"actionOnUnmanage,omitempty"`
```
weikanglim, line 57: I'm a little worried about the approach of adding all the ARM properties to `azure.yaml` and the pass-through idea here.

Looking at what we have:

```yaml
infra:
  provider: bicep
  deploymentStacks:
    bypassStackOutOfSyncError: false
    actionOnUnmanage: 
      managementGroups: delete    # delete or detach
      resourceGroups: delete      # delete or detach
      resources: delete           # delete or detach
    denySettings:
      mode: denyDelete            # denyDelete, denyWriteAndDelete, none
      applyToChildScopes: false   # true | false
      excludedActions:            # List of actions to exclude
        - action1
        - action2
      excludedPrincipals:         # List of principals to exclude
        - principal1
        - principal2
```

I would have some questions like:

- Perhaps `bypassStackOutOfSyncError` is a one-time activity (would you really want the same behavior each time?) that would be covered by a `--force` gesture to avoid an interactive prompt?

- `actionOnUnmanage.managementGroups`. I would some time to look into the docs more, but would this only apply to management group deployments which `azd` doesn't support currently?

From an overall UX standpoint, I think we're taking the stance that "if you're here, you're an advanced user and we're just going to expose all the details of ARM". Another approach is "let's make these advanced features easier to use, or easier to understand -- what are features that a user would care about, with no Bicep or ARM specifics referenced here?"

cli/azd/pkg/infra/provisioning/provider.go
wbreza: No, terraform would not know or care about deployment stacks since it does not leverage ARM based deployments.

wbreza: I agree with you but we have a couple of constraints due to the package dependency graph that exist today that may cause circular dependencies.  Let me see if there are any other options available.

cli/azd/pkg/azapi/stack_deployments.go
wbreza: Yes, deployment stacks are technically an Azure resource. Currently, they’re created using the Go SDK, which exposes all the necessary configuration, although these options are mostly utilized by advanced users.

I’d like to begin exposing this in the `azure.yaml`. As we move forward with the alpha phase, we may receive additional feedback from users who might want to configure some of these settings at a higher level. Let’s hold off on drawing any conclusions until we have more data.

cli/azd/pkg/infra/provisioning/provider.go
wbreza: This is good feedback but since this feature is only in `alpha` phase right now I'd prefer to push out as is till we collect more feedback from users.

Once we are ready to move into beta or GA we can update the config to reference a strongly typed configuration.

cli/azd/pkg/azapi/stack_deployments.go
weikanglim: My ask is that we think a little more about the scenarios and motivation behind settings in #4307. I believe this would be a fair ask, given that we would also need to document them for our public docs.

I understand that we're thinking about sourcing feedback from users. That is indeed a useful tool, but we should at least be in the 80% ballpark of correctness with 20% refinement. Here we're taking the approach that this will be a black-box 3rd-party extensible integration that azd knows little about. This hasn't been the approach with Bicep. 

Here are the concrete examples listed above with more context that we could think about:

- `bypassStackOutOfSyncError` - In the [docs](https://learn.microsoft.com/en-us/azure/azure-resource-manager/bicep/deployment-stacks?tabs=azure-powershell#handle-the-stack-out-of-sync-error), it calls out that as: "This switch should only be used after thoroughly review the list of resources in the stack before rerunning the command. This switch should never be used by default." I wonder if this really fits the profile of a project-level property?
- `actionOnUnmanage.managementGroups` - azd doesn't support management groups provisioning currently -- perhaps it'd be wise for us to not expose this?

I think @vhvb1989 also brought up a bigger picture question of whether these configuration are environment-specific behavior -- does `prod` require the same level of protection as `dev`? If I think from that perspective, it is perhaps true that adding the settings here into `azure.yaml` would not help as much as we'd think it would.

If we examine the established pattern from `az` is that these properties are specific to a given deployment, and environments will have different deployment stacks, we should at least consider leaning this way instead of against it. My educated but unvalidated hypothesis is that users are likely choose different settings for different environments. The proposed design here locks us into project-level configuration that we'd need to think about again when environments are brought back into the picture.

wbreza: Thanks for the detailed thoughts on this! I agree that understanding the full scope is important, but for now, I think this implementation is sufficient, especially given that we're in the early stages with this alpha feature. We want to avoid over-engineering or spending too much time refining every detail upfront when we're still in the phase of gathering user feedback.

While we may not have everything figured out from the start, this approach gives us flexibility to evolve. It doesn't lock us into any specific configuration long-term, and we can still layer in environment-level config for `azd` if it becomes necessary down the line. Similarly, we retain the option to remove certain properties later if we find they're not useful or misaligned with the project-level scope.

At this point, it's about moving forward efficiently while keeping our options open for future adjustments based on what we learn from usage and feedback.

vhvb1989 (dismissed):
Let's gather more information around using azure.yaml before merging this.
I think a little more conversation would be beneficial for similar future changes.

cli/azd/pkg/infra/provisioning/provider.go
wbreza: Created #4399 to track this for upcoming sprint.

PR #4336, opened 2024-09-18
bicep: add spinner to `azd down`

weikanglim:
Small change to add a spinner message to `azd down`.  Remove remaining to-dos.

Before:
output

After:
after

### DIFF
--- a/cli/azd/pkg/infra/provisioning/bicep/bicep_provider.go
+++ b/cli/azd/pkg/infra/provisioning/bicep/bicep_provider.go
@@ -201,7 +201,6 @@ func (p *BicepProvider) State(ctx context.Context, options *provisioning.StateOp
 
 	var err error
 	spinnerMessage := "Loading Bicep template"
-	// TODO: Report progress, "Loading Bicep template"
 	p.console.ShowSpinner(ctx, spinnerMessage, input.Step)
 	defer func() {
 		// Make sure we stop the spinner if an error occurs with the last message.
@@ -238,7 +237,6 @@ func (p *BicepProvider) State(ctx context.Context, options *provisioning.StateOp
 		outputs = azure.ArmTemplateOutputs{}
 	}
 
-	// TODO: Report progress, "Retrieving Azure deployment"
 	spinnerMessage = "Retrieving Azure deployment"
 	p.console.ShowSpinner(ctx, spinnerMessage, input.Step)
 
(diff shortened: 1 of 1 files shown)


cli/azd/pkg/infra/provisioning/bicep/bicep_provider.go
```diff
@@ -730,7 +727,8 @@ func (p *BicepProvider) Destroy(
 	options provisioning.DestroyOptions,
 ) (*provisioning.DestroyResult, error) {
 	modulePath := p.modulePath()
-	// TODO: Report progress, "Compiling Bicep template"
+	p.console.ShowSpinner(ctx, "Discovering resources to delete...", input.Step)
+	defer p.console.StopSpinner(ctx, "", input.StepDone)
```
ellismg, line 731: Just checking - in the common case we'll call `StopSpinner` twice now, once on L793 and then once when this function returns.  That's okay right?

ellismg (approved):
It's the little things that make the tool feel more polished.  Thanks for taking the time to focus on them, @weikanglim :heart:.

wbreza (approved):

cli/azd/pkg/infra/provisioning/bicep/bicep_provider.go
weikanglim: The pattern to call `StopSpinner` multiple times is generally safe, per the docs:

```
	// Stop the current spinner from the console and change the spinner bar for the lastMessage
	// Set lastMessage to empty string to clear the spinner message instead of a displaying a last message
	// If there is no spinner running, this is a no-op function
```

So calling `StopSpinner` multiple times is ok.

But the order of operations here may be interesting -- I'm curious if @vhvb1989 views the setup here as idiomatic, i.e. 

In the happy path:
1. In the start of `Destroy()`, we create a spinner that 
2. The happy path spinner stops on L793
3. In the end of `Destroy()` we stop a spinner again

Supposedly, if we create another spinner in-between L793 and the end of the function that perhaps creates another spinner this may slightly quirky -- but if we did add another StopSpinner to the defer stack, I think we do observe the overall correct behavior. So from a correctness standpoint, I think we're solid, however I'm not sure if we like it idiomatically.

vhvb1989: seems fine to me.

vhvb1989 (approved):

PR #4340, opened 2024-09-18
ci: Use federated auth in CI

ellismg:
Move away from a long lived client secret for auth during CI testing
in favor of federated credentials via a service connection.

(To unblock the builds I am seeing how far we can get with our old
trick of delegating stuff to `az` before we teach `azd` how to use
`AzurePipelinesCredential`.)

Contributes To #4341

### DIFF
--- a/cli/azd/test/functional/cli_test.go
+++ b/cli/azd/test/functional/cli_test.go
@@ -65,8 +65,6 @@ type cliConfig struct {
 
 	// The client ID to use for live Azure tests.
 	ClientID string
-	// The client secret to use for live Azure tests.
-	ClientSecret string
 	// The tenant ID to use for live Azure tests.
 	TenantID string
 	// The Azure subscription ID to use for live Azure tests.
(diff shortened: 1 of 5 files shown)


cli/azd/test/functional/cli_test.go
```diff
@@ -81,9 +81,9 @@ type cliConfig struct {
 
 func (c *cliConfig) init() {
 	c.CI = os.Getenv("CI") != ""
-	c.ClientID = os.Getenv("AZD_TEST_CLIENT_ID")
-	c.ClientSecret = os.Getenv("AZD_TEST_CLIENT_SECRET")
-	c.TenantID = os.Getenv("AZD_TEST_TENANT_ID")
+	// c.ClientID = os.Getenv("AZD_TEST_CLIENT_ID")
+	// c.ClientSecret = os.Getenv("AZD_TEST_CLIENT_SECRET")
+	// c.TenantID = os.Getenv("AZD_TEST_TENANT_ID")
```
vhvb1989, line 86: I think the main purpose of this test is to check azd work with client-secret.  If we change this, we would be loosing coverage for it.
Can we remove/deprecate support for client-secret?   Otherwise we might want coverage

ellismg: Trying to get the builds off the floor first (with minimal damage) and then will revisit the coverage we want here

It's possible we'll just lose coverage in CI and have to move this to be a manual test if we get to a place where we can't be using client secrets in CI.

The place this was ready was `t.Skip`'d for now.

vhvb1989 (approved):
LGTM

Issue #4341, opened 2024-09-18
ci: Move away from client-secret in favor of service connections


ellismg:
Historically we used a client id and client secret pair to authenticate (for both `azd` itself and other tools it calls, like `terraform`) in CI.  This worked, but the downside of it is that there was a long-lived secret (the client secret) that we had to protect and update.

To mitigate this, our central engineering team is trying to adopt service connections for all our CI jobs.  This means that we need to move away from our client-secret based authentication and towards something that uses service connections.  For `azd` itself, it means teaching it how to use something like [`AzurePipelinesCredential`](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/azidentity#AzurePipelinesCredential) and then updating our jobs to use it.

I'm not sure what the exact answer is for terraform yet.  For places we use `az` we already have an answer via the `AzureCLI@2` task which works with service connections.

Since our builds are presently on the floor due to an expired client secret (which we don't want to renew) we'll do this in two parts:

1. Move away from client-secret by delegating auth where we can to the `az` CLI via the (`auth.useAzCliAuth` config flag we have) and disabling whatever tests we can't make work quickly with this.  This gets our builds green again.

2. Do the work to teach `azd` how to natively authenticate using a service connection (and figure out how to configure any other tools like `terraform` to work in this world) and re-enable any tests disabled as part of (1).

ellismg:
For terraform, I thought that I could be clever and build a `ARM_OIDC_REQUEST_URL` as outlined in https://registry.terraform.io/providers/hashicorp/azuread/latest/docs/guides/service_principal_oidc#configuring-terraform-to-use-oidc and have that just work for OIDC auth based on what AzurePipelinesCredential, but it did not seem to.  I got this error:

> cli.go:235: 14.306s [stderr] │ Error: building account: could not acquire access token to parse claims: githubAssertion: received HTTP status 405 with response: {"count":1,"value":{"Message":"The requested resource does not support http method 'GET'."}}

Could be I screwed something up with the URL building however, so I need to dig in more.

PR #4343, opened 2024-09-18
auth: Support `azure-pipelines` for `--federated-credential-provider`

ellismg:
This change adds `azure-pipelines` as a supported value for the `--federated-credential-provider` switch of `azd auth login`. This provider can be used to do OIDC based login within the context of a job running in Azure Pipelines.

When using this provider, `azd` uses the `AzurePipelinesCredential` type to manage the OIDC dance. In addition to the client and tenant id of the service principal you plan to authenticate with, we also need the "service connection id" (this is the ID of the object created in Azure Pipelines that contains the connection information) and the system access token which is the security token for the running build (it's what is used to authenticate to the OIDC endpoint that is being run inside the pipeline which we use to fetch the federated credential).  The client and tenant ids can be provided on the command line, via the existing `--client-id` and `--tenant-id` switches, or, when `azure-pipelines` is the `federated-credential-provider`, they can be read from the `AZURESUBSCRIPTION_CLIENT_ID` and `AZURESUBSCRIPTION_TEANT_ID` system environment varaibles.  The service connection ID and acceess token are read from the `AZURESUBSCRIPTION_SERVICE_CONNECTION_ID` and `SYSTEM_ACCESSTOKEN` environment variaibles.

These system environment variables correspond to what the `AzureCLI@2` sets when passing the `azureSubscription` parameter (after translating the service connection name to a service connection ID, which is handy) allowing something like this to work:

```yaml
- task: AzureCLI@2
  inputs:
    azureSubscription: azconnection
    scriptType: bash
    scriptLocation: inlineScript
    inlineScript: |
      azd auth login --federated-credential-provider "azure-pipelines"
      azd up
```

Note that while the AzureCLI task is used above, it's only used for setting these environment variables and doing the service connection name to id translation.  You could do something like the following instead:

```yaml
- bash: |
    azd auth login --federated-credential-provider "azure-pipelines"
    azd up
  env:
    AZURE_SUBSCRIPTION_CLIENT_ID: "Your Client ID"
    AZURE_SUBSCRIPTION_TENANT_ID: "Your Tenant ID"
    AZURE_SUBSCRIPTION_SERVICE_CONNECTION_ID: "Your Service Connection ID"
    SYSTEM_ACCESSTOKEN: $(System.AccessToken)
```

And things will work as well. This can be useful in cases where you can use the `AzureCLI@2` task directly.

We can consider writing our own task as some point as well.

Contributes To #4341


cli/azd/cmd/auth_login.go
```diff
@@ -139,6 +141,16 @@ func (lf *loginFlags) Bind(local *pflag.FlagSet, global *internal.GlobalCommandO
...
 		"The path to the client certificate for the service principal to authenticate with.")
+	local.StringVar(
+		&lf.serviceConnectionID,
+		"service-connection-id",
+		"",
+		"The service connection id to authenticate with.")
+	local.StringVar(
+		&lf.systemAccessTokenEnvVar,
+		"system-access-token-environment-variable-name",
+		"SYSTEM_ACCESSTOKEN",
```
ellismg, line 152: This is the sample name used in the example on how to get this (liked to via the identity SDK): https://learn.microsoft.com/en-us/azure/devops/pipelines/build/variables?view=azure-devops&tabs=yaml#systemaccesstoken, so I figured I'd use that as a default.

cli/azd/pkg/auth/manager.go
```diff
@@ -1033,11 +1115,13 @@ type federatedAuth struct {
...
-	FromOneAuth     bool    `json:"fromOneAuth,omitempty"`
-	ClientID        *string `json:"clientId,omitempty"`
-	TenantID        *string `json:"tenantId,omitempty"`
+	ManagedIdentity       bool    `json:"managedIdentity,omitempty"`
+	HomeAccountID         *string `json:"homeAccountId,omitempty"`
+	FromOneAuth           bool    `json:"fromOneAuth,omitempty"`
+	ClientID              *string `json:"clientId,omitempty"`
+	TenantID              *string `json:"tenantId,omitempty"`
+	ServiceConnectionID   *string `json:"serviceConnectionId,omitempty"`
+	SystemAccessTokenName *string `json:"systemAccessTokenName,omitempty"`
```
ellismg, line 1124: Instead of storing the value (which is both secret and per build) I figured we'd store the name of the env-var and then look up the value when we needed it.  That means we don't need to add anything to the persisted secret struct.

This is used as a bearer token when doing the OIDC dance, internally there's an injected env-var that has a endpoint that the credential make requests to and this token secures these requests.

eng/pipelines/templates/steps/azd-login.yml
```diff
@@ -1,6 +1,7 @@
 parameters:
   SubscriptionConfiguration: $(sub-config-azure-cloud-test-resources)
   AzdDirectory: ""
+  ServiceConnectionId: "3d79cc98-46f2-428c-bdd5-861414f85602"
```
ellismg, line 4: @danieljurek I assume the right thing to do here is to pull this off of `SubscriptionConfiguration` but I don't know enough about what that structure looks like to know if this value there or not.  These needs to be the ID of the service connection that we want to use.

The AzureCLI@2 task seems to do name to id translation (I am guessing) but I haven't looked into how that works yet and if we could support it.  I found this GUID by looking around at some of our internal configuration files.  I won't check this in until you tell me the right way to do it, but trying to get stuff off the ground and taking some shortcuts.

weikanglim: This may or may not help:

- [AzureCli task](https://github.com/microsoft/azure-pipelines-tasks/blob/9a01bdd13f107406603afc20bf088ac68a7223c4/Tasks/AzureCLIV2/azureclitask.ts#L198)
- [vsts-task-lib](https://github.com/microsoft/azure-pipelines-task-lib/blob/a56dfecac46753540e79da0d2c93fb41666c0c3e/node/task.ts#L469)

I think environment variables are being seeded -- I'm not sure if that means you definitely an `azdo` task to get everything working e2e.

cli/azd/cmd/auth_login.go
```diff
@@ -139,6 +141,16 @@ func (lf *loginFlags) Bind(local *pflag.FlagSet, global *internal.GlobalCommandO
 		cClientCertificateFlagName,
 		"",
 		"The path to the client certificate for the service principal to authenticate with.")
+	local.StringVar(
+		&lf.serviceConnectionID,
+		"service-connection-id",
```
vhvb1989, line 146: Ideally, I would like to hide all the implementation details from the connection because it is 100% unique to Azure Devops.
Azd currently has the flag `--federated-credential-provider` to hide all the complexity and implementation of how to do OICD in github actions. That means, when you want to log in azd in in GitHub actions, you do:

```
run: |
          azd auth login `
            --client-id "$Env:AZURE_CLIENT_ID" `
            --federated-credential-provider "github" `
            --tenant-id "$Env:AZURE_TENANT_ID"
```

The implementation about how to get the system token and any other tokens is inside azd's implementation.

So, I would like to just have a new provider `azdo`, which I can use in a Azdo task like:

```
run: |
          azd auth login `
            --client-id "$Env:AZURE_CLIENT_ID" `
            --federated-credential-provider "azdo" `
            --tenant-id "$Env:AZURE_TENANT_ID"
```

Azd would internally know what ENV VARS to looks for to get the system access token and the connection-id.
Azd can adapte itself to multiple scenarios. For example, running `azd login` from the AzureCLI task.

ellismg: I can look into this but I think the problem is that without the connection id or service connection name we are not going to be able to make this work.  We could consider something like `--federated-credential-provider azdo:<service-connection-name>` so it would be something like `--federated-credential-provider azdo:azure-sdk-tests` for us.  Maybe we can make that work?

Like you, I'd love to be able to do this.

weikanglim: Once we're done with investigation, I'd love to be able to make sense of whether `azd` should just be an AzDo task to support this scenario better (my hunch was always that it is indeed necessary, without moving mountains in AzDo).

If we are able to deliver an up-to-par experience, I'd love for that to be captured in #4341

eng/pipelines/templates/steps/azd-login.yml
ellismg: > I think environment variables are being seeded

Yes, it looks like this happens when you have a [task with an input of type `connectedService:AzureRM`](https://devblogs.microsoft.com/devops/public-preview-of-workload-identity-federation-for-azure-pipelines/#custom-azure-tasks-and-extensions).  So I guess we could make this "just work" in the context of an `AzureCLI@2` task (or create our own).

weshaggard: Yes see https://github.com/Azure/azure-sdk-for-net/blob/e7d65a13f3f83119ed54e927654e0512d35f087b/sdk/identity/Azure.Identity/samples/OtherCredentialSamples.md#example-of-using-an-azure-pipelines-task 

The AzurePipelineCredential pulls the needed variables that get set in the AzureCLi task.

cli/azd/cmd/auth_login.go
ellismg: Made some progress here, we now support using `azure-pipelines` as an argument to `--federated-credential-provider`.  We now sniff the environment variables that the AzCli and AzurePowerShell tasks set, or if you want to use a different task, you can explicitly set `AZURESUBSCRIPTION_SERVICE_CONNECTION_ID`.

We may want to end up writing our own wrapper task at some point.

vhvb1989: I started an azd task for azdo some time ago. Basically copying what the AzureCLI task does here: https://github.com/microsoft/azure-pipelines-tasks/tree/master/Tasks/AzureCLIV2

I stopped myself and started wondering if we want to update `AzureCLIV2` to check if azd is installed and login azd as well as az. It would be one less task to install for customers.

cli/azd/pkg/auth/manager.go
```diff
@@ -53,6 +53,31 @@ const useAzCliAuthKey = "auth.useAzCliAuth"
 // auth related configuration information (e.g. the home account id of the current user). This information is not secret.
 const authConfigFileName = "auth.json"
 
+// AzurePipelinesClientIDEnvVarName is the name of the environment variable that contains the client ID for the principal
+// to use when authenticating with Azure Pipelines via OIDC. It is set by both the AzureCLI@2 and AzurePowerShell@5 tasks
```
danieljurek, line 57: Of note but not blocking: We encountered issues where these values are not set in `AzurePowerShell@5` in Windows https://github.com/Azure/azure-sdk-for-net/pull/44684 

This can be mitigated by following the steps in that PR. No change needed here.

cli/azd/cmd/auth_login.go
```diff
@@ -389,6 +389,18 @@ func runningOnCodespacesBrowser(ctx context.Context, commandRunner exec.CommandR
...
 func (la *loginAction) login(ctx context.Context) error {
+	if la.flags.federatedTokenProvider == string(auth.AzurePipelinesFederatedTokenProvider) {
+		if la.flags.clientID == "" {
+			log.Printf("setting client id from environment variable %s", auth.AzurePipelinesClientIDEnvVarName)
+			la.flags.clientID = os.Getenv(auth.AzurePipelinesClientIDEnvVarName)
+		}
+
+		if la.flags.tenantID == "" {
+			log.Printf("setting tenant id from environment variable %s", auth.AzurePipelinesClientIDEnvVarName)
+			la.flags.clientID = os.Getenv(auth.AzurePipelinesTenantIDEnvVarName)
```
vhvb1989, line 400: typo, maybe?

```suggestion
			la.flags.tenantId = os.Getenv(auth.AzurePipelinesTenantIDEnvVarName)
```

Should we fail if env vars are not set?  or is it ok to wait to the log-in error at later ?

cli/azd/pkg/auth/manager.go
```diff
@@ -53,6 +53,31 @@ const useAzCliAuthKey = "auth.useAzCliAuth"
...
+
+// AzurePipelinesServiceConnectionNameEnvVarName is the name of the environment variable that contains the name of the
+// service connection to use when authenticating with Azure Pipelines via OIDC. It is set by both the AzureCLI@2 and
+// AzurePowerShell@5 tasks when using a service connection or can be set manually when not using these tasks.
+const AzurePipelinesServiceConnectionIDEnvVarName = "AZURESUBSCRIPTION_SERVICE_CONNECTION_ID"
+
+// AzurePipelinesSystemAccessTokenEnvVarName is the name of the environment variable that contains the system access token
+// used to auth against the ODIC endpoint for Azure Pipelines. It needs to be set by the task that runs the azd command by
+// adding `SYSTEM_ACCESSTOKEN: $(System.AccessToken)` to the `env` section of the task configuration.
+const AzurePipelinesSystemAccessTokenEnvVarName = "SYSTEM_ACCESSTOKEN"
```
vhvb1989, line 74: do we need to export these for other packages?

cli/azd/pkg/auth/manager.go
```diff
@@ -1017,16 +1104,20 @@ type persistedSecret struct {
...
 // federated auth token providers
 var (
-	gitHubFederatedAuth federatedTokenProvider = "github"
+	GitHubFederatedTokenProvider         FederatedTokenProvider = "github"
+	AzurePipelinesFederatedTokenProvider FederatedTokenProvider = "azure-pipelines"
 )
 
 // token provider for federated auth
-type federatedTokenProvider string
+type FederatedTokenProvider string
```
vhvb1989, line 1112: Why did it become exported?

vhvb1989 (approved):

cli/azd/cmd/auth_login.go
ellismg: > typo, maybe?

Yup, fixed!

> Should we fail if env vars are not set? or is it ok to wait to the log-in error at later ?

We could - that's how things work on GitHub (where you must pass both `--client-id` and `--tenant-id` explicitly.  I figured since we could source these from environment variables (since the task sets them) it would be nice to support allowing the user to not specify them when `azure-pipelines` is the `federated-credential-provider` but I'm not wed to that decision.   I do think the ergonomics are a little nicer.

cli/azd/pkg/auth/manager.go
ellismg: I'll undo this - I think I had this exported earlier iteration when it was a parameter to a `Login` method.

cli/azd/pkg/auth/manager.go
ellismg: The first two are only consumed by the `cmd` package (in `auth_login.go`) and the other two are only used here.  I'll move the first two over to the `cmd` and then unexport all of these.

PR #4161, opened 2024-08-01
project: Support Remote Builds

ellismg:
This change sketches out support for using the ACR remote build
feature and uses it in the `containerapp` service target.

To use it, add `remoteBuild: true` to the `docker` member of a
`service` in `azure.yaml`

addresses #509

PR #4357, opened 2024-09-19
core: Update `azure.yaml.json` for `remoteBuild`

### LINKED ISSUE #4353: Update `azure.yaml.json` schema for remoteBuild

ellismg:
This adds the `remoteBuild` configuration knob which was added in azure/azure-dev#4161 to our schema for `azure.yaml`.

Fixes #4353

### DIFF
--- a/schemas/alpha/azure.yaml.json
+++ b/schemas/alpha/azure.yaml.json
@@ -678,6 +678,11 @@
                     "items": {
                         "type": "string"
                     }
+                },
+                "remoteBuild": {
+                    "type": "boolean",
+                    "title": "Optional. Whether to build the image remotely",
+                    "description": "If set to true, the image will be built remotely using the Azure Container Registry remote build feature. If set to false, the image will be built locally using Docker."
                 }
             }
         },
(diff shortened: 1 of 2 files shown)


schemas/alpha/azure.yaml.json
```diff
@@ -678,6 +678,11 @@
                     "items": {
                         "type": "string"
                     }
+                },
+                "remoteBuild": {
+                    "type": "boolean",
+                    "title": "Optional. Whether to build the image remotely",
+                    "description": "If set to true, the image will be built remotely using the Azure Container Registry remote build feature. If set to false, the image will be built locally using Docker."
                 }
```
wbreza, line 686: Should we consider adding that this can only be used when using an ACR or do you feel it is implied?

wbreza (approved):

schemas/alpha/azure.yaml.json
ellismg: I feel like it is implied today by saying "using the Azure Container Registry remote build feature", but if we end up getting a lot of feedback that remote build doesn't work when using non ACR registries (and I expected it to) I will look at updating the prose here.

PR #4358, opened 2024-09-19
Updates `azure.yaml` schema for deployment stacks configuration

wbreza:
> [!IMPORTANT]
> Do not merge till after release

Updates the alpha schema for `azure.yaml` to include the optional deployment stacks configuration.

The `deploymentStacks` property will light up when `provider: bicep`. When specifying the `deploymentStacks` property the user must configure at least one of the properties otherwise the whole property can be omitted and use the internal defaults.

### DIFF
--- a/schemas/alpha/azure.yaml.json
+++ b/schemas/alpha/azure.yaml.json
@@ -1079,11 +1079,6 @@
             "oneOf": [
                 {
                     "required": [
-                        "bypassStackOutOfSyncError"
-                    ]
-                },
-                {
-                    "required": [
                         "actionOnUnmanage"
                     ]
                 },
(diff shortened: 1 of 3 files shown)


schemas/alpha/azure.yaml.json
```diff
@@ -1047,6 +1071,111 @@
             "required": [
                 "deployment"
             ]
+        },
+        "deploymentStacksConfig": {
+            "type": "object",
+            "title": "The deployment stack configuration used for the project.",
+            "additionalProperties": false,
+            "oneOf": [
```
ellismg, line 1079: What does this end up meaning?  Does it mean if you're going to have a `deploymentStacksConfig` object then it must have at least one property (i.e. disallowing the empty object as the value)?

schemas/alpha/azure.yaml.json
```diff
@@ -1047,6 +1071,111 @@
...
+                    ]
+                },
+                {
+                    "required": [
+                        "denySettings"
+                    ]
+                }
+            ],
+            "properties": {
+                "bypassStackOutOfSyncError": {
```
ellismg, line 1097: As a follow up to this - it would be good to know what the error experience is today in `azd` when this is set to false and we get a stack out of sync error - I suspect our error behavior today is not as good as it could be and we can polish the error experience here.

ellismg (approved):
Every time I see one of these PRs I feel like I learn a little more about JSON Schema.  Thanks for the updates here, @wbreza and for being a wizard at this stuff!  LGTM.

schemas/alpha/azure.yaml.json
wbreza: Yes, if the user bothers to add the `deploymentStacks` property, then enforce that they set at least one of the top level sections otherwise they can just omit the whole thing and use the internal defaults.

schemas/alpha/azure.yaml.json
wbreza: I'm not exactly sure how to get into this state but it would present as an ARM deployment error.

Issue #4310, opened 2024-09-12
SERVICE_XXX_RESOURCE_EXISTS environment variables are always `false` when passed as ARM Deployment params in CI environment


MarcelMichau:
- [X] Make sure you've installed the latest version using [instructions in the wiki](../wiki/install)

**Output from `azd version`**
azd version 1.10.1 (commit 31409a33266fb4a5fdbb644bc83988e725d6c7c9)

**Describe the bug**
In a CI environment where the `SERVICE_XXX_RESOURCE_EXISTS` environment variables do not pre-exist in the `.azure/<environmentName>/.env` file, when running `azd provision`, the `SERVICE_XXX_RESOURCE_EXISTS` environment variables are always `false` when passed as Bicep params.

**To Reproduce**

1. Run `azd init --template Azure-Samples/todo-python-mongo-aca`
2. Run `azd provision` & wait for resources to finish provisioning
3. Open the `.azure/<environmentName>/.env` file & delete all environment variables except for `AZURE_ENV_NAME`, `AZURE_LOCATION` & `AZURE_SUBSCRIPTION_ID` to mimic the [environment variables passed to the CI task](https://github.com/Azure-Samples/todo-python-mongo-aca/blob/f8a0665acfd8bc4f16c08cf21b5cf93ee2d39367/.azdo/pipelines/azure-dev.yml#L41).
4. Run `azd provision --no-state` to force an infra deployment
5. When the provision operation starts, observe the values for `SERVICE_API_RESOURCE_EXISTS` & `SERVICE_WEB_RESOURCE_EXISTS` in the `.azure/<environmentName>/.env` file will update to `true`
6. Inspect the Deployment Inputs in the Azure Portal - the values for the `apiAppExists` & `webAppExists` parameters are still `false`:

![image](https://github.com/user-attachments/assets/2c2baea6-902c-475d-ba8c-9d25c415451e)

7. Subsequently running `azd provision --no-state` when the `SERVICE_XXX_RESOURCE_EXISTS` environment variables are present & set to `true` in the `.azure/<environmentName>/.env` file will correctly set the Bicep params to `true`:

![image](https://github.com/user-attachments/assets/dc795a7d-db5f-4976-b713-e9db5e3f24cd)

**Expected behavior**
[...]

PR #3155, opened 2023-12-21
pipeline additional vars secrs

vhvb1989:
This change adds support for projects to define additional variables or secrets to be set as part of the gh actions workflow or azdo pipelines.

fix: https://github.com/Azure/azure-dev/issues/1597

## Strategy

Adding 2 new fields to `azure.yaml` for pipeline:
```yaml
pipeline:
  variables:
    - var_foo
    - var_bar
   secrets:
     - sec_foo
     - sec_bar
```

With this configuration, azd will find `variables` within the azd-environment and set them as variables for the pipeline during `azd pipeline config`.
The same will happen with `secrets`, but the values from azd-environment are set as secrets on the pipeline.

## other required changes

- Call `ensureParameters()` during bicep_provider initialization. This change enables `azd pipeline config` to prompt for required parameters and gives the option to save the answer.  When the answer is saved, it is used during CI/CD.  
This is a change that covers a similar scenario as the one we have for .NET Aspire, which prompts user to select what services to expose. But, instead of services to expose, azd ask the user for required bicep parameters w/o deploying.

- Create memory cache for compile bicep and ensure parameters. Since we are compiling bicep during init and also checking parameters, we want to skip doing this again when a plan is created (within the same azd execution).

PR #4363, opened 2024-09-20
bicep: remove `main.parameters.json` caching

### LINKED ISSUE #3973: `azd env set` in a hook script doesn't have the variables available in future steps

**Output from `azd version`**
> azd version 1.9.3 (commit e1624330dcc7dde440ecc1eda06aac40e68aa0a3)

**Describe the bug**

If you use a hook, such as preprovision, to add environment variables to azd using `azd env set`, those variables aren't available until the azd session is restarted (meaning you can't use `azd up`).

We come across this in https://github.com/microsoft/azure-openai-service-proxy/ in which we have to create an Entra ID app registration and then want to push the client id/tenant id to services that azd will deploy.

[Here is the preprovision hook definition](https://github.com/microsoft/azure-openai-service-proxy/blob/main/azure.yaml#L31-L35) and then [here is the script](https://github.com/microsoft/azure-openai-service-proxy/blob/main/infra/scripts/setup-app-registration.sh).

The reason that we rely on this is there is no bicep support for deploying app registrations presently, so we use the Azure CLI to handle it.

Ideally, each step of the azd pipeline would load the `.env` file, even if it's being run through a group command like `azd up`.

weikanglim:
With this change, we remove the in-memory `main.parameters.json` cache.

Backstory: In the old days of `azd`, we introduced in-memory caching for `main.parameters.json` with #3155 to avoid a re-prompt when the user declines the prompt "Save the value in the environment for future use?". We added this as a way to communicate within `azd` that "this value is configured for the current azd process lifetime".

After #3479, we ended up always saving the values to `config.json` and remove the "Save the value in the environment for future use?" prompt. Since then, the in-memory caching just reduces an extra load from disk, but behaviorally act the same.

Fixes #3973, #4310, 

## Verification

Since #4310 is implemented as an internal `preprovision` hook that sets `SERVICE_X_RESOURCE_EXISTS` and shares the same root cause, they would be fixed the same.

I modified our [storage](https://github.com/Azure/azure-dev/blob/3b48ae58706ea4c0feb1943d92caef9b9a24a12e/cli/azd/test/functional/testdata/samples/storage/azure.yaml) sample:
```yaml
name: storage
metadata: 
  template: azd-test/storagetest@v1
hooks:
  preprovision: 
   shell: sh
   run: azd env set INT_TAG_VALUE 500
```

And observed both in Portal and the output that `INT_TAG_VALUE` was successfully set to `500`.

### DIFF
--- a/cli/azd/pkg/infra/provisioning/bicep/bicep_provider.go
+++ b/cli/azd/pkg/infra/provisioning/bicep/bicep_provider.go
@@ -74,10 +74,8 @@ type BicepProvider struct {
 	ignoreDeploymentState bool
 	// compileBicepResult is cached to avoid recompiling the same bicep file multiple times in the same azd run.
 	compileBicepMemoryCache *compileBicepResult
-	// prevent resolving parameters multiple times in the same azd run.
-	ensureParamsInMemoryCache azure.ArmParameters
-	keyvaultService           keyvault.KeyVaultService
-	portalUrlBase             string
+	keyvaultService         keyvault.KeyVaultService
+	portalUrlBase           string
 }
 
 // Name gets the name of the infra provider
@@ -1793,10 +1791,6 @@ func (p *BicepProvider) ensureParameters(
 	ctx context.Context,
 	template azure.ArmTemplate,
 ) (azure.ArmParameters, error) {
-	if p.ensureParamsInMemoryCache != nil {
-		return maps.Clone(p.ensureParamsInMemoryCache), nil
-	}
-
 	parameters, err := p.loadParameters(ctx)
 	if err != nil {
 		return nil, fmt.Errorf("resolving bicep parameters file: %w", err)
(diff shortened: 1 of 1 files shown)


cli/azd/pkg/infra/provisioning/bicep/bicep_provider.go
```diff
@@ -1922,19 +1916,15 @@ func (p *BicepProvider) ensureParameters(
...
-				configuredParameters[prompt.key] = azure.ArmParameterValue{
-					Value: value,
-				}
 			}
 		}
 	}
 
 	if configModified {
 		if err := p.envManager.Save(ctx, p.env); err != nil {
-			p.console.Message(ctx, fmt.Sprintf("warning: failed to save configured values: %v", err))
```
weikanglim, line 1934: In case it isn't clear why ignoring this error may not be ideal:

Suppose that `Save()` failed due to flakiness once -- this variable ends up getting re-prompted which is confusing behavior to a user. Straight up failing is better here.

A code [demonstration](https://github.com/Azure/azure-dev/compare/main...weikanglim:weilim/rm-param-cache-failsave-demo?expand=1).

Here's it live:
Screenshot 2024-09-20 at 1 22 38 PM

wbreza (approved):

ellismg (approved):

vhvb1989 (approved):

PR #4369, opened 2024-09-23
Updates docker `buildargs` to expandable strings

### LINKED ISSUE #4062: [Issue] Docker buildargs does not support interpolation
- [ X ] Make sure you've installed the latest version using [instructions in the wiki](../wiki/install)

**Describe the bug**
If you have a Static App (like React) and a backend api in ACA, it would be nice if we could take the output of the infra deployment (the url) and pass it as `buildArgs` in the dockerfile:

For example: 

```biceps
# main.biceps
...
output REACT_APP_BACKEND_URL string = backend.outputs.uri
```

```yaml
# azure.yaml - Notice the environment variables!
...
    frontend:
        project: "src/frontend"
        host: containerapp
        language: ts
        dist: build
        docker:
            path: Dockerfile
            buildArgs:
                - REACT_APP_BACKEND_URL=${REACT_APP_BACKEND_URL}
```

```powershell
# deploy infra and populate outputs as environment variables
azd provision --environment dev
# utilize environment variables to interpolate the azure.yaml
azd deploy 
```

wbreza:
Resolves #4062 

Updates docker `buildargs` in `azure.yaml` to support environment variable substitutions.

### DIFF
--- a/cli/azd/pkg/project/dotnet_importer.go
+++ b/cli/azd/pkg/project/dotnet_importer.go
@@ -177,6 +177,23 @@ func mapToStringSlice(m map[string]string, separator string) []string {
 	return result
 }
 
+// mapToExpandableStringSlice converts a map of strings to a slice of expandable strings.
+// Each key-value pair in the map is converted to a string in the format "key:value",
+// where the separator is specified by the `separator` parameter.
+// If the value is an empty string, only the key is included in the resulting slice.
+// The resulting slice is returned without any string interpolation performed.
+func mapToExpandableStringSlice(m map[string]string, separator string) []osutil.ExpandableString {
+	var result []osutil.ExpandableString
+	for key, value := range m {
+		if value == "" {
+			result = append(result, osutil.NewExpandableString(key))
+		} else {
+			result = append(result, osutil.NewExpandableString(key+separator+value))
+		}
+	}
+	return result
+}
+
 func (ai *DotNetImporter) Services(
 	ctx context.Context, p *ProjectConfig, svcConfig *ServiceConfig,
 ) (map[string]*ServiceConfig, error) {
@@ -234,7 +251,7 @@ func (ai *DotNetImporter) Services(
 			Docker: DockerProjectOptions{
 				Path:      dockerfile.Path,
 				Context:   dockerfile.Context,
-				BuildArgs: mapToStringSlice(dockerfile.BuildArgs, "="),
+				BuildArgs: mapToExpandableStringSlice(dockerfile.BuildArgs, "="),
 			},
 		}
 
@@ -314,7 +331,7 @@ func (ai *DotNetImporter) Services(
 			dOptions = DockerProjectOptions{
 				Path:         bContainer.Build.Dockerfile,
 				Context:      bContainer.Build.Context,
-				BuildArgs:    mapToStringSlice(bArgs, "="),
+				BuildArgs:    mapToExpandableStringSlice(bArgs, "="),
 				BuildSecrets: bArgsArray,
 				BuildEnv:     reqEnv,
 			}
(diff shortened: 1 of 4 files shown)


cli/azd/pkg/project/dotnet_importer.go
```diff
@@ -177,6 +177,23 @@ func mapToStringSlice(m map[string]string, separator string) []string {
 	return result
 }
 
+// mapToExpandableStringSlice converts a map of strings to a slice of expandable strings.
+// Each key-value pair in the map is converted to a string in the format "key:value",
+// where the separator is specified by the `separator` parameter.
+// If the value is an empty string, only the key is included in the resulting slice.
+// The resulting slice is returned without any string interpolation performed.
```
wbreza, line 184:
```suggestion
// The resulting slice is returned without any string interpolation performed. 
```

cli/azd/pkg/project/dotnet_importer.go
```diff
@@ -177,6 +177,23 @@ func mapToStringSlice(m map[string]string, separator string) []string {
...
```
wbreza, line 184:
```suggestion
// The resulting slice is returned without any string interpolation performed.
```

weikanglim (approved):
LGTM. Thanks for doing this!

cli/azd/pkg/project/framework_service_docker.go
```diff
@@ -34,15 +34,15 @@ import (
...
-	BuildArgs   []string                `yaml:"buildArgs,omitempty"   json:"buildArgs,omitempty"`
+	Path        string                    `yaml:"path,omitempty"        json:"path,omitempty"`
+	Context     string                    `yaml:"context,omitempty"     json:"context,omitempty"`
+	Platform    string                    `yaml:"platform,omitempty"    json:"platform,omitempty"`
+	Target      string                    `yaml:"target,omitempty"      json:"target,omitempty"`
+	Registry    osutil.ExpandableString   `yaml:"registry,omitempty"    json:"registry,omitempty"`
+	Image       osutil.ExpandableString   `yaml:"image,omitempty"       json:"image,omitempty"`
+	Tag         osutil.ExpandableString   `yaml:"tag,omitempty"         json:"tag,omitempty"`
+	RemoteBuild bool                      `yaml:"remoteBuild,omitempty" json:"remoteBuild,omitempty"`
+	BuildArgs   []osutil.ExpandableString `yaml:"buildArgs,omitempty"   json:"buildArgs,omitempty"`
```
vhvb1989, line 45: Docker users might want to map the build arguments to env vars using docker directly (see: https://docs.docker.com/reference/cli/docker/buildx/build/#build-arg) 

When that's te case, azd should either use the full system environment when calling docker, or check if there are any buildArgs like `FOO_VAR` (w/o using the `=` symbol) and grab `FOO_VAR` from system env and use it when calling `docker`.

Having expandable string support is a plus, but it might not be idiomatic/expected to docker users.

cli/azd/pkg/project/project_config_test.go
```diff
@@ -126,7 +127,38 @@ services:
...
+	})
+
+	serviceConfig := &ServiceConfig{
+		Docker: DockerProjectOptions{
+			Registry: osutil.NewExpandableString("${REGISTRY}"),
+			Image:    osutil.NewExpandableString("${IMAGE}"),
+			Tag:      osutil.NewExpandableString("${TAG}"),
+			BuildArgs: []osutil.ExpandableString{
+				osutil.NewExpandableString("key1=${KEY1}"),
+				osutil.NewExpandableString("key2=${KEY2}"),
```
vhvb1989, line 152: nits
- What happen if the `${KEY}` is equal to something like `ff=23`?  should azd return an invalid arg error?  or would we just use the value as is.
- What happen if the separator `=` is missing?

wbreza: if the key returns a value like looks like `key=value`, then this would be interpolated before calling `docker build` and should be ok.  

If the separator is missing then it would fall back to the default behavior of `docker build` command and lookup env var matching that name.

vhvb1989 (approved):
LGTM

Issue #4359, opened 2024-09-20
Support not prompting for parameters that have a default value


davidfowl:
See https://github.com/dotnet/aspire/pull/5586

We can now express parameters that have a [default value](https://github.com/dotnet/aspire/blob/4d80745955ccc0c9b31b1a8fe94e8d5b5b8a1d77/src/Schema/aspire-8.0.json#L280-L286).

These values shouldn't be prompted for by default, but there could be a way to say (prompt for everything).

PR #4371, opened 2024-09-23
[Aspire] Add support for default.value for parameter.v0

vhvb1989:
fix: https://github.com/Azure/azure-dev/issues/4359

Parameter.v0 can have either a `generate` or `value` for the `default` settings. This PR adds support for `azd` to handle
the `value` settings from the manifest.

When `valued` is defined like:

```json
"param-with-value": {
      "type": "parameter.v0",
      "value": "{param-with-value.inputs.value}",
      "inputs": {
        "value": {
          "type": "string",
          "secret": false,
          "default":{
            "value": "default value for param"
          }
        }
      }
    },
```

azd generates a bicep input param like:

```bicep
param string paramName = 'default value for param'
```

and a mapping to system env var `AZURE_PARAM_NAME` to override the default value.

Notes:
- If `secret` is `true` and `default.value` is defined, azd returns an error because that would expose a secured() value.
- Mapping to system env vars works for both, provision and deploy.  For `deploy`, the YAML definition references the name of the parameter and the default value to support mapping to env var during deploy time.


wbreza (approved):

cli/azd/pkg/project/service_target_dotnet_containerapp.go
```diff
@@ -364,7 +366,16 @@ func (_ *containerAppTemplateManifestFuncs) UrlHost(s string) (string, error) {
 
 const infraParametersKey = "infra.parameters."
 
+// Parameter resolves the name of a parameter defined in the ACA yaml definition. The parameter can be mapped to a system
+// environment variable or persisted in the azd environment configuration.
 func (fns *containerAppTemplateManifestFuncs) Parameter(name string) (string, error) {
+	envVarMapping := scaffold.EnvFormat(name)
```
weikanglim, line 372: @vhvb1989 just a note, not a review comment. I really like this strategy, I think we can think of ways to make this a first-class concept: #4404 in case you have thoughts there.

cli/azd/pkg/apphost/generate.go
```diff
@@ -90,6 +90,12 @@ func init() {
 					return strings.ReplaceAll(src, ".", "")
 				},
 				"envFormat": scaffold.EnvFormat,
+				"bicepParameterValue": func(value *string) string {
```
weikanglim, line 93: I like how clean this looks. Thank you!

cli/azd/pkg/apphost/generate.go
```diff
@@ -90,6 +90,12 @@ func init() {
...
+					if value == nil {
+						return ""
+					}
+					return fmt.Sprintf(" = '%s'", *value)
```
weikanglim, line 97: Just double-checking here: we're only supporting strings for now? I think the single quotes need omitted for things like bool, int, object and array.

cli/azd/pkg/apphost/generate.go
```diff
@@ -274,14 +281,22 @@ func BicepTemplate(name string, manifest *Manifest, options AppHostOptions) (*me
 	for _, key := range genParametersKeys {
 		parameter := generator.bicepContext.InputParameters[key]
 		parameterMetadata := ""
-		if parameter.Default != nil && parameter.Default.Generate != nil {
-			pMetadata, err := inputMetadata(*parameter.Default.Generate)
-			if err != nil {
-				return nil, fmt.Errorf("generating input metadata for %s: %w", key, err)
+		var parameterDefaultValue *string
```
weikanglim, line 284: I wonder if it's more readable if we can build the structs incrementally, which avoids local variables.

Here's what I mean:
```golang
input := genInput{Name: key, Secret: parameter.Secret, Type: parameter.Type}
if parameter.Default != nil {
	input.Value = parameter.Default.Value
}

autoGen := autoGenInput{genInput: input, MetadataType: azure.AzdMetadataTypeGenerate}
if parameter.Default != nil && parameter.Default.Generate != nil {
	pMetadata, err := inputMetadata(*parameter.Default.Generate)
	if err != nil {
		return nil, fmt.Errorf("generating input metadata for %s: %w", key, err)
	}
	autoGen.MetadataConfig = pMetadata
	// Note: azd is not checking or validating that Default.Generate and Default.Value are not both set.
	// The AppHost prevents this from happening by not allowing both to be set at the same time.
}

parameters = append(parameters, autoGen)
```

I think there's less indentations here. It also *subjectively* slightly feels more idiomatic go for me.

cli/azd/pkg/project/service_target_dotnet_containerapp.go
```diff
@@ -377,6 +388,18 @@ func (fns *containerAppTemplateManifestFuncs) Parameter(name string) (string, er
 	return valString, nil
 }
 
+// ParameterWithDefault resolves the name of a parameter defined in the ACA yaml definition.
+// The parameter can be mapped to a system environment variable or be default to a value directly.
+func (fns *containerAppTemplateManifestFuncs) ParameterWithDefault(name string, defaultValue string) (string, error) {
+	envVarMapping := scaffold.EnvFormat(name)
+	// map only to system environment variables. Not adding support for mapping to azd environment by design (b/c
+	// parameters could be secured)
```
weikanglim, line 396: Let's track this somehow if that makes sense to you. I think either adding the issue here directly, or updating the issue to reference this file works.

weikanglim (approved):
Neat! Thanks for updating the test cases to show the behavior working out in practice.

cli/azd/pkg/apphost/generate.go
weikanglim: BTW, one thing that kinda falls out by doing this -- it feels weird that `autoGenInput` is a struct that wraps `genInput`. This may be a result of the text/template context usage, but in my mind, an input has optional auto-gen properties.

ellismg (approved):

PR #4386, opened 2024-09-30
Expand init description

alexwolfmsft:


weikanglim:
@alexwolfmsft Thanks for the submission here! Would it be possible to add an issue referencing the motivation behind the change? There's been some recent conversations around naming of things at azd  -- at first glance I'm noticing we're proposing to call the final output of `azd init` a "template"? This may be something that may be worthy of discussion.

alexwolfmsft:
> @alexwolfmsft Thanks for the submission here! Would it be possible to add an issue referencing the motivation behind the change? There's been some recent conversations around naming of things at azd -- at first glance I'm noticing we're proposing to call the final output of `azd init` a "template"? This may be something that may be worthy of discussion.

Hey @weikanglim - sure thing. The details about the user feedback item can be found below - this item was generated from our docs feedback tool. I know there's been some conversations around naming for our docs as well - for example, project vs app vs template, etc. In my opinion "initializes an app" makes less sense than "initializes a template" because azd is template-oriented. The app might already exist before azd is added to convert it to a template. Did you have other thoughts on this? Thanks!

https://dev.azure.com/msft-skilling/Content/_workitems/edit/280674

weikanglim:
@alexwolfmsft Thanks for sharing the item! That was really helpful! 

Based on some recent naming discussions with a pending outcome, "initializing an [azd] project" or "initializing an Azure project" comes to mind. 

While azd as-of-today is indeed template-oriented, having a more generic naming outside of "templates" may be a good idea for long-term runway. In my mind, azd users that aren't using the template ecosystem are a core part of user base (although those numbers are perhaps small today).

cli/azd/cmd/init.go
```diff
@@ -39,10 +39,11 @@ func newInitFlags(cmd *cobra.Command, global *internal.GlobalCommandOptions) *in
 	return flags
 }
 
+//nolint:lll
 func newInitCmd() *cobra.Command {
 	return &cobra.Command{
 		Use:   "init",
-		Short: "Initializes an existing template or creates a new template based on app code in the current directory.",
+		Short: "Initializes an existing template or creates a new template based on app code in the current directory. The initialization workflow retrieves or generates the template, prompts you for essential environment settings, and creates a `.azure` folder to store those settings.",
```
wbreza, line 46: Do we want this much detail in the "short" description? It might be better in the "long" description?

weikanglim:
hey 👋 @alexwolfmsft , I am slightly on the fence with the wording of "initializing a template". At the same time, I have some suggestions listed above that I don't feel comfortable either with moving forward. Would it be okay to move forward here with a terminology discussion on #4398?

I do agree with the ask at hand here and we will get to this -- it is very valid feedback -- but I feel more comfortable when we have fully landed on standard terminology. I do💯 appreciate your efforts here to improve the documentation here to address our users' pain points, we will work with you to address the concerns at hand.

alexwolfmsft:
> hey 👋 @alexwolfmsft , I am slightly on the fence with the wording of "initializing a template". At the same time, I have some suggestions listed above that I don't feel comfortable either with moving forward. Would it be okay to move forward here with a terminology discussion on #4398?
> 
> I do agree with the ask at hand here and we will get to this -- it is very valid feedback -- but I feel more comfortable when we have fully landed on standard terminology. I do💯 appreciate your efforts here to improve the documentation here to address our users' pain points, we will work with you to address the concerns at hand.

Hey @weikanglim thanks for the consideration here. Moving the discussion to the other issue works for me, if you'd like we can just close out this PR for now.

PR #4405, opened 2024-10-03
fix java devcontainer on latest feature issue

vhvb1989:
use feature version 1.6 because 1.6.1

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -14,7 +14,7 @@
         },
         "ghcr.io/devcontainers/features/azure-cli:1": {},
         "ghcr.io/devcontainers/features/dotnet:2": {},
-        "ghcr.io/devcontainers/features/java:1": {},
+        "ghcr.io/devcontainers/features/java:1.6.0": {},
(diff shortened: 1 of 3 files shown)


eng/pipelines/repoman.yml
```diff
@@ -96,19 +96,19 @@ extends:
...
-                  targetType: filePath
-                  filePath: ./eng/scripts/Update-PRComment.ps1
-                  arguments: >-
-                    -Repo 'azure/azure-dev'
-                    -PRNumber '$(PRNumber)'
-                    -BodyFile '$(CommentBodyFile)'
-                    -Tag '<!-- #comment-repoman-generate -->'
-                env:
-                  GH_TOKEN: $(azuresdk-github-pat)
+              # - task: PowerShell@2
```
weikanglim, line 99: Just double checking here: we'll want to re-enable this before the final merge?

vhvb1989: Just did it now.  Will see if it works now.  If not, I might just override the checks based on the previous succesful run

rajeshkamal5050 (approved):

PR #4409, opened 2024-10-04
Extensions Management Commands

wbreza:
This PR adds the `extension` command group into `azd` and enables the execution of installed extensions.

> [!IMPORTANT]
> The extensions feature is behind a feature flag and must be enabled by running:
> `azd config set alpha.extensions on`

## Commands 

- `azd extension list` - List available extensions in the registry
- `azd extension show <name>` - Show details of a specified extension
- `azd extension install <name>` - Install an extension
- `azd extension uninstall <name>` - Uninstall an extension
- `azd extension upgrade <name>` - Upgrades an extension

## Registry

The registry is loaded from an aka URL @ https://aka.ms/azd/extensions/registry

- The registry is protected by a digital signature verification. Only extensions from a verified registry can be installed.
- Binaries also require checksum validation before extensions can be installed.

### `azd extension list`
image

### `azd extension show ai`
image

### `azd extension install ai`
image

### `azd -h`
Extensions show up in standard azd help commands.
image

### `azd test -h`
Extension help shells into extension executable
image

### `azd test --unit`
Invoke extensions through azd command interface
image

### `azd extension uninstall ai`
image


weikanglim:
Thanks for sharing the work here, @wbreza! Just sharing some high-level UX observations and wonders for now while I find time to parse through more details.

Sharing some observations:

- The UX mockups for `azd extension` are great -- looks very presentable and usable.
- Verbs for `show` and `list` is great, we're adopting consistent verbiage that azd uses.
- The `install` command shows usage after installation, which is awesome. This is a perfect example of progressive disclosure models where the CLI leads the user to "next steps" that I really like.

Sharing some wonders I have:

- `azd test` extension example: I realize that `azd test` is the example of the extension being added here. Perhaps we could make it slightly clearer in documentation (separate section for Example extension). Once we do start thinking along these scenarios, it'd be good to imagine what running the extension command feels like. The question I have in the back of my mind is if the extension will look and feel as-if they're azd (or perhaps they don't have to).
- Extension names:  I'm wondering how we would think about extension names potentially squatting on command flags that `azd` uses. For example, if we shipped an extension `ai`, and at some point azd has decided to adopt `azd ai` as the first-party command -- how does that journey look like? Do we also want to think about reserving keywords upfront?
- Listing extensions with `azd -h`: With `azd -h` listing all extensions present, would this end up looking too lengthy if the user installs many extensions? We could choose to summarize this as a header: "Displaying (5 out of n extensions). To see help text for installed extensions, run `azd extension -h`" -- or something along those lines.

Overall, I think this is a really great start that will help us guide us towards productive conversations as we chat more about this today.

jongio:
Do you envision different types of extensions, such as root-level command extensions, command groups with multiple commands, or extensions added to existing commands (e.g., adding an extension to 'azd up')? Additionally, can I add an extension to extend infrastructure providers?

wbreza:
> Do you envision different types of extensions, such as root-level command extensions, command groups with multiple commands, or extensions added to existing commands (e.g., adding an extension to 'azd up')? Additionally, can I add an extension to extend infrastructure providers?

Yes, ultimately I envision different types of extensions.  The first set of extensions would allow extension authors to create new commands that can be executed through `azd`.  These could be single commands or command groups.

Over time I can also see extensions providing additional functionality to customize experience within `azd`, add additional providers to pluggable components, etc.

Regarding `azd up` - since the current extension model will expose new commands - users can take advantage of customizing their `up` workflow to call into these extensions.  I would also expect that hooks can also be leveraged for users to register pre/post hooks against custom extension commands.

cli/azd/pkg/extensions/manager.go
```diff
@@ -0,0 +1,533 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to copy file: %w", err)
+	}
+
+	return nil
+}
+
+// Verify verifies the given data and its Base64-encoded signature
+func verifySignature(data []byte, signature string) error {
+	publicKey, err := loadPublicKey(resources.PublicKey)
```
ellismg, line 492: If we end up keeping this public key stuff around to do signature verification - I think we should arrange things such that we do as much of this loading and praising at package `init` time and just `panic` on errors, then so that `verifySignature` can never fail due to an error reading the key we're going to use for verification.  In theory if the call to `loadPublicKey(resources.PublicKey)` fails, it means there's a problem with our embedded key and I'd prefer if that failure was loud and prompt.

cli/azd/pkg/cache/file_cache.go
```diff
@@ -0,0 +1,98 @@
...
+	"strconv"
+	"time"
+
+	"github.com/azure/azure-dev/cli/azd/pkg/osutil"
+)
+
+// CacheResolver is a function that resolves the cache value.
+type CacheResolver[T any] func(ctx context.Context) (*T, error)
+
+// FileCache is a cache that stores the value in a file otherwise resolves it.
```
ellismg, line 17: Nice abstraction - I'll try to move the stuff that caches the latest version for our update check over to it if possible.

You might note that it's the JSON encoding of `*T` that is stored in the cache here, since that matters a bit for consumers.

cli/azd/cmd/testdata/TestUsage-azd-test.snap
```diff
@@ -0,0 +1,18 @@
+
```
ellismg, line 1: Is this still intended to be part of this PR?

cli/azd/cmd/extensions.go
```diff
@@ -0,0 +1,174 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to get extension %s: %w", extensionName, err)
+	}
+
+	allEnv := []string{}
+	allEnv = append(allEnv, os.Environ()...)
+
+	forceColor := !color.NoColor
+	if forceColor {
+		allEnv = append(allEnv, "FORCE_COLOR=1")
```
ellismg, line 135: Is this some new convention we're building?

cli/azd/cmd/extensions.go
```diff
@@ -0,0 +1,174 @@
...
+	}
+
+	homeDir, err := os.UserHomeDir()
+	if err != nil {
+		return nil, fmt.Errorf("failed to get user home directory: %w", err)
+	}
+
+	extensionPath := filepath.Join(homeDir, extension.Path)
+
+	_, err = os.Stat(extensionPath)
```
ellismg, line 155: Would it be possible to move this logic in to `GetInstalled`?  Could the `extension` returned have a full path (including the home directory) and do the `Stat` to ensure the binary is there?

ellismg (commented):
Partial review - the public key stuff scares me the most - I think if we want to do this we need to engage with the security team and talk with them about how we do this - I think they may want changes to how we transport the signature so we don't, for example, have to change the data we download in order to validate the signature.

Would love to better understand the treat model for that stuff (I'm onboard with doing the SHA validation of downloads, as we do for other tools).

Like how clean and small this is - great first step on the new path forward.

cli/azd/pkg/extensions/manager.go
```diff
@@ -0,0 +1,533 @@
...
+	if err := json.Unmarshal(body, &registry); err != nil {
+		return nil, fmt.Errorf("failed to unmarshal raw registry JSON: %w", err)
+	}
+
+	return registry, nil
+}
+
+func validateRegistry(registry *Registry) error {
+	// Extract the signature
+	signature := registry.Signature
```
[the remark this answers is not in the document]
wbreza: IMHO - from a security point of view we need to protect the integrity of the registry file so unapproved extensions cannot be installed.

Without this integrity either the content of the request and/or file could be spoofed allowing installations of non-approved extensions that could be a security risk.

cli/azd/pkg/cache/file_cache.go
```diff
@@ -0,0 +1,98 @@
...
+		}
+		return c.value, nil
+	}
+
+	value, err := c.resolver(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("failed to resolve data: %w", err)
+	}
+
+	if err := c.Set(value); err != nil {
```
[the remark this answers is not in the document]
wbreza: I have no strong opinion if we should just exit early here or not if the flag is set.The `AZD_NO_CACHE` flag was primarily added to force get on the underlying data store vs whether the file should be written or not since the primary purpose of this component is to cache outputs.

cli/azd/pkg/cache/file_cache.go
```diff
@@ -0,0 +1,98 @@
...
+	val, has := os.LookupEnv("AZD_NO_CACHE")
+	if has {
+		noCache, err := strconv.ParseBool(val)
+		if err == nil && noCache {
+			return false
+		}
+	}
+
+	info, err := os.Stat(c.filePath)
+	if os.IsNotExist(err) {
```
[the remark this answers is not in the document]
wbreza: Added condition for permissions issues as well.

cli/azd/cmd/extension.go
```diff
@@ -0,0 +1,588 @@
...
+		if err != nil {
+			return err
+		}
+	}
+
+	return tabs.Flush()
+}
+
+func (a *extensionShowAction) Run(ctx context.Context) (*actions.ActionResult, error) {
+	extensionName := a.args[0]
```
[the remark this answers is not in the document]
wbreza: which commands are you referring to? This should be pretty consistent to most of the other commands that we support today.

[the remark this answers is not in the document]
wbreza: The distinction here is that `<name> param is "required" for the `azd ext show <name>` command however it is "optional" for the others and can also be used with the `--all` flag.  This behavior is consistent with other `azd` command that have required positions arguments like `azd env set <key> <value>`.

I'd prefer we keep a consistent experience and then apply a fix for all similar required named params in the future.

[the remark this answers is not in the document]
JeffreyCA: I see, thanks for clarifying

cli/azd/cmd/extension.go
```diff
@@ -0,0 +1,832 @@
+package cmd
```
vhvb1989, line 1:
```suggestion
// Copyright (c) Microsoft Corporation. All rights reserved.
// Licensed under the MIT License.

package cmd
```

cli/azd/pkg/cache/file_cache.go
```diff
@@ -0,0 +1,114 @@
...
+// FileCache is a cache that stores the value in a file otherwise resolves it.
+type FileCache[T any] struct {
+	filePath      string
+	resolver      CacheResolver[T]
+	cacheDuration time.Duration
+	value         *T
+}
+
+// NewFileCache creates a new file cache.
+func NewFileCache[T any](cacheFilePath string, cacheDuration time.Duration, resolver CacheResolver[T]) *FileCache[T] {
```
vhvb1989, line 28: How is this FileCache used for the extension commands?   Should it be a different PR?  What's the relation?  I only see it used from tests

vhvb1989 (approved):

wbreza:
Closing and using #4766 instead.

Issue #4410, opened 2024-10-04
Experimentation middleware is slowing azd, even making azd credential to fail getting tokens


vhvb1989:
Review if the experimentation middleware and entire feature in azd is working or if it should be removed.

Evidence log of Azure Developer CLI credentials failing:

```
2024/10/04 15:48:03 main.go:56: azd version: 1.10.1 (commit 31409a33266fb4a5fdbb644bc83988e725d6c7c9)
2024/10/04 15:48:03 main.go:214: using cached latest version: 1.10.1 (expires on: 2024-10-05T17:19:02Z)
2024/10/04 15:48:03 project.go:144: Reading project from file '/Users/pamelafox/aisearch-openai-rag-audio/azure.yaml'
2024/10/04 15:48:03 middleware.go:108: running middleware 'debug'
2024/10/04 15:48:03 middleware.go:108: running middleware 'ux'
2024/10/04 15:48:03 middleware.go:108: running middleware 'experimentation'
2024/10/04 15:48:03 assignment.go:88: could not read assignment from cache: cache expired
2024/10/04 15:48:03 main.go:50: Retry: =====> Try=1
2024/10/04 15:48:03 main.go:50: Request: ==> OUTGOING REQUEST (Try=1)
   GET https://default.exp-tas.com/exptas49/b80dfe81-554e-48ec-a7bc-1dd773cd6a54-azdexpws/api/v1/tas
   Accept: application/json
   User-Agent: azsdk-go-variantassignment/0.0.1 (go1.23.0 X:loopvar; darwin)
   X-ExP-Parameters: REDACTED
   X-Exp-Sdk-Version: REDACTED

2024/10/04 15:48:03 main.go:50: Response: ==> REQUEST/RESPONSE (Try=1/15.842958ms, OpTime=15.861208ms) -- REQUEST ERROR
   GET https://default.exp-tas.com/exptas49/b80dfe81-554e-48ec-a7bc-1dd773cd6a54-azdexpws/api/v1/tas
   Accept: application/json
   User-Agent: azsdk-go-variantassignment/0.0.1 (go1.23.0 X:loopvar; darwin)
   X-ExP-Parameters: REDACTED
   X-Exp-Sdk-Version: REDACTED
   --------------------------------------------------------------------------------
   ERROR:
[...]

PR #4416, opened 2024-10-08
Remove experimentation middleware from global middleware list

wbreza:
While we troubleshoot blocking issues when communicating with our backend experimentation we will remove the experimentation middleware from the global middleware list since we are not running any active experiments.

fix: https://github.com/Azure/azure-dev/issues/4410


weikanglim (approved):
LGTM. Since experimentation only sets an in-memory alpha feature flag, and we don't have any experimentation on-going right now, this change is very safe and shouldn't cause any unexpected things.

vhvb1989 (approved):

PR #4418, opened 2024-10-09
WIP: Prompt user to select or create new resource

wbreza:
Authoring bicep templates that work well across a wide range of scenarios is hard.  One difficult scenario is supporting a bicep module where the resource is conditionally created based on whether the consumer wants to use a new resource or leverage an existing resource.

We are seeing this scenario multiple times in various AI use cases where customers may already have AI studio hubs, projects already created or want to leverage existing services like an Azure Open AI Service, etc.

## Idea
Leverage custom `azd` metadata within bicep to enable authors to mark up parameters that map to a new or existing resource using bicep user defined types.

### Example of user defined type

```bicep
type OptionalResource = {
  name: string
  subscriptionId: string?
  resourceGroup: string?
  exists: bool
}
```

Then when specifying a bicep parameter additional `azd` metadata can be added to light up new experiences in `azd`

### Example of using parameters with custom metadata

```bicep
@metadata({
  azd: {
    type: 'resource'
    resource: {
      displayName: 'Key Vault'
      description: 'The key vault to use for storing secrets'
      type: 'Microsoft.KeyVault/vaults'
    }
  }
})
param keyVault OptionalResource
```

## `azd` Experience

When calling `azd provision` users will automatically be prompted to select a resource if this is the first time they are deploying the project.

image

The data for the custom input is serialized and sent in as bicep input parameters

image

The provisioning continues as normal and is up to the author of the bicep to ensure they reuse or create the resource based upon the input value.

image

### DIFF
--- a/cli/azd/pkg/azapi/resource_service.go
+++ b/cli/azd/pkg/azapi/resource_service.go
@@ -80,6 +80,44 @@ func (rs *ResourceService) GetResource(
 	}, nil
 }
 
+func (rs *ResourceService) ListSubscriptionResources(
+	ctx context.Context,
+	subscriptionId string,
+	listOptions *armresources.ClientListOptions,
+) ([]*Resource, error) {
+	client, err := rs.createResourcesClient(ctx, subscriptionId)
+	if err != nil {
+		return nil, err
+	}
+
+	// Filter expression on the underlying REST API are different from --query param in az cli.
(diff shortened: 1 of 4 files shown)


heaths:
Why a new bicep type i.e., `OptionalResource`? Seems this makes the bicep (or ARM) templates less portable. Callers shouldn't *need* `azd` for deployment. This prevents creation through other, existing means like the "Deploy to Azure" buttons (though, only ARM templates currently).

PR #4426, opened 2024-10-10
report inner error description

vhvb1989:
fix: https://github.com/Azure/azure-dev/issues/4425

### DIFF
--- a/cli/azd/pkg/azapi/stack_deployments.go
+++ b/cli/azd/pkg/azapi/stack_deployments.go
@@ -285,7 +285,11 @@ func (d *StackDeployments) DeployToSubscription(
 
 	_, err = poller.PollUntilDone(ctx, nil)
 	if err != nil {
-		return nil, err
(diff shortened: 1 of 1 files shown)


cli/azd/pkg/azapi/stack_deployments.go
```diff
@@ -285,7 +285,11 @@ func (d *StackDeployments) DeployToSubscription(
 
 	_, err = poller.PollUntilDone(ctx, nil)
 	if err != nil {
-		return nil, err
+		deploymentError := createDeploymentError(err)
```
weikanglim, line 288: I'm wondering if we would want to do something similar for DeployToResourceGroup.

vhvb1989: yes!
added it

wbreza (approved):

weikanglim (approved):
Thanks for the quick fix!

PR #4434, opened 2024-10-11
easy-init: allow for longer container app names

### LINKED ISSUE #4433: easyInit: service names with too long names results in image name collision
I ran `azd init` on two service directories:
- azure-ai-router-api
- azure-ai-router-web

In the bicep file generated, I notice the truncation:

```bicep
module azureAiRouterApi './app/azure-ai-router-api.bicep' = {
  name: 'azure-ai-router-api'
  params: {
    name: '${abbrs.appContainerApps}azure-ai-rou-${resourceToken}'
```

```bicep
module azureAiRouterWeb './app/azure-ai-router-web.bicep' = {
  name: 'azure-ai-router-web'
  params: {
    name: '${abbrs.appContainerApps}azure-ai-rou-${resourceToken}'
```

This results in an error when running `azd up`:

Cannot modify a container app '<name>'  because there is an active provisioning operation in progress.

weikanglim:
In the generated bicep for easy-init, remove resource token and `ca-` from the container app name.

This makes it inline with Aspire's generation experience. Overall, the new naming scheme improves DX: developers have matching service names with corresponding app/image name.

Fixes  #4433

### DIFF
--- a/cli/azd/internal/repository/detect_confirm.go
+++ b/cli/azd/internal/repository/detect_confirm.go
@@ -263,6 +263,7 @@ func (d *detectConfirm) remove(ctx context.Context) error {
 		confirm, err := d.console.Confirm(ctx, input.ConsoleOptions{
 			Message: fmt.Sprintf(
 				"Remove %s in %s?", projectDisplayName(svc), relSafe(d.root, svc.Path)),
+			DefaultValue: true,
 		})
 		if err != nil {
 			return err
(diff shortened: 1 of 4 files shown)


ellismg (approved):
Works for me!  I think eliding the `ca-` prefix here is also a good idea because it means the hostname for the container doesn't have the `ca-` prefix.

vhvb1989 (approved):

Issue #4431, opened 2024-10-11
easy-init: support mysql


weikanglim:
Background: @saragluna had started implementation on some Java enhancements in #4422. We will split this into smaller changes to land incrementally.

Work: Add MySQL database support. We can enable auto-detection for Java for now, and have separate follow-ups with other languages.

weikanglim:
https://github.com/Azure/azure-dev/pull/4651

rajeshkamal5050:
@weikanglim @saragluna anything pending here? 

Can we use this one to track https://github.com/Azure/azure-dev/issues/4671 ?

weikanglim:
This was done as part of #4651

Issue #4430, opened 2024-10-11
appdetect: analyze pom.xml for database dependencies


weikanglim:
Background: @saragluna had started implementation on some Java enhancements in #4422. We will split this into smaller changes to land incrementally. This is the first of few.

Work: Enhance `appdetect` to analyze root `pom.xml` and `pom.xml` submodules (different library/app projects present) to app detect, to auto-detect:

- MongoDB
- Redis
- PostgreSQL

rajeshkamal5050:
Closing this issue as not planned since we are moving towards a more streamlined, AI-driven initialization experience through GitHub Copilot for Azure and the Azure MCP Server, with app initialization, IaC provisioning and application deployments powered by the AZD MCP tools underneath. More AZD-native, AI-driven/agentic CLI experiences are on the way, your feedback is always welcome! @kristenwomack @puicchan @tg-msft 

https://learn.microsoft.com/en-us/azure/developer/azure-mcp-server/tools/azure-deploy

PR #4437, opened 2024-10-12
Add more java related analysis code

saragluna:
Address https://github.com/Azure/azure-dev/issues/4431 and part of https://github.com/Azure/azure-dev/issues/4430. 

This PR will:
- Enhance the Postgres and MySql detection for Java applications.
- Add support for multi-module Java projects.


cli/azd/internal/appdetect/java.go
```diff
@@ -16,13 +22,157 @@ func (jd *javaDetector) Language() Language {
...
+					return nil, fmt.Errorf("error applying rules: %w", err)
+				}
+				result = append(result, *project)
+
+			}
+		}
+	}
+
+	// todo (xiada) we should support multiple modules
+	return &result[0], nil
```
weikanglim, line 77: Some suggestions here on how we could implement multi-module support:

A Detector instance is kept-alive while recurring down directories. We can use this to our advantage.

1. When we hit the first root pom.xml, and we detect that it's a root project with `modules`, we can set a `rootProject` variable in the detector. We return `nil` here, and the recursion will continue.
2. When we hit the second `pom.xml`, we can check if `rootProject` if set, or use the metadata we set in `rootProject`. Either case, we know we are detecting a child module now, and we can decide to return the project.

weikanglim: Some explanation here on why want to keep the method signature `DetectProject(ctx context.Context, path string, entries []fs.DirEntry) (*Project, error)`, and not `[]Project, error`:

We want to assign each directory to a Project, or none at all. It helps with `appdetect.DetectDirectory` when we want to just examine a single directory without recursion. Maybe this isn't a good idea -- but at least we do have a way around this right now.

cli/azd/internal/appdetect/java.go
```diff
@@ -16,13 +22,157 @@ func (jd *javaDetector) Language() Language {
...
+	}
+
+	project.path = filepath.Dir(filePath)
+
+	return &project, nil
+}
+
+func detectDependencies(javaProject *javaProject, project *Project) (*Project, error) {
+	databaseDepMap := map[DatabaseDep]struct{}{}
+	if javaProject.mavenProject.Dependencies != nil {
```
weikanglim, line 157: this nil-check is unnecessary. a nice property of `go`: you can safely iterate over `nil`:

Here's a playground [example](https://go.dev/play/p/0S6M0idJDHC)

weikanglim (commented):
@saragluna Thanks for getting this far! I really love what we're doing to support the java dependencies and multi-module repositories.

I left some suggestions on multi-module implementation if it helps at all. Let's continue down this path!

cli/azd/internal/appdetect/java.go
```diff
@@ -16,13 +22,157 @@ func (jd *javaDetector) Language() Language {
...
+			return project, nil
 		}
 	}
 
 	return nil, nil
 }
+
+type javaProject struct {
+	mavenProject mavenProject
+}
```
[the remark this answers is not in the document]
saragluna: Ok, will remove it now.

cli/azd/internal/appdetect/java.go
```diff
@@ -16,13 +22,157 @@ func (jd *javaDetector) Language() Language {
...
+	var result []Project
+
+	entries, err := os.ReadDir(path)
+
+	if err != nil {
+		return nil, fmt.Errorf("error reading directory: %w", err)
+	}
+
+	for _, entry := range entries {
+		if "pom.xml" == entry.Name() {
```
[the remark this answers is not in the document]
saragluna: Yes, you are right.

cli/azd/internal/appdetect/java.go
```diff
@@ -16,13 +22,157 @@ func (jd *javaDetector) Language() Language {
...
+
+// Plugin represents a build plugin.
+type plugin struct {
+	GroupId    string `xml:"groupId"`
+	ArtifactId string `xml:"artifactId"`
+	Version    string `xml:"version"`
+}
+
+func analyzeMavenProject(projectPath string) ([]mavenProject, error) {
+	rootProject, _ := readMavenProject(filepath.Join(projectPath, "pom.xml"))
```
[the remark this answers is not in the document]
saragluna: Will do

cli/azd/internal/appdetect/testdata/java-multimodules/application/build.gradle
```diff
@@ -0,0 +1,23 @@
+plugins {
```
weikanglim, line 1: I'm wondering for making sure the test setup reflects the current feature set, would you agree with removing the `*.gradle` files for now?

We'll add it back when we support gradle -- but this requires a core azd change so it'll take some time before we get there.

cli/azd/internal/appdetect/appdetect_test.go
```diff
@@ -41,6 +41,16 @@ func TestDetect(t *testing.T) {
 					Path:          "java",
```
weikanglim, line 41: I think we can consider updating the pom.xml for `java` test cases to include the new dependencies. Then, we can update the test expectation here to include the expected `DatabaseDeps` here. 

The background here is that these are acceptance test cases. The way I like to think about is that `appdetect_test.go` contains "the happiest path scenarios", and `java_test.go` to contain "slightly unhappy path scenarios"

cli/azd/internal/appdetect/java.go
```diff
@@ -16,13 +23,121 @@ func (jd *javaDetector) Language() Language {
...
+type plugin struct {
+	GroupId    string `xml:"groupId"`
+	ArtifactId string `xml:"artifactId"`
+	Version    string `xml:"version"`
+}
+
+func readMavenProject(filePath string) (*mavenProject, error) {
+	bytes, err := os.ReadFile(filePath)
+	if err != nil {
+		return nil, fmt.Errorf("error reading file: %w", err)
```
weikanglim, line 110: Let's make the error message more concise here and just do `return nil, err`.

This makes more sense because in `DetectProject`, we have `error reading pom.xml:`, here we have `error reading file:`. So the final error message looks like: `error reading pom.xml: error reading file: <actual error>`.

I think we should just keep it `error reading pom.xml: <actual error>`.

cli/azd/internal/appdetect/java.go
```diff
@@ -16,13 +23,121 @@ func (jd *javaDetector) Language() Language {
...
+	}
+
+	var project mavenProject
+	if err := xml.Unmarshal(bytes, &project); err != nil {
+		return nil, fmt.Errorf("error parsing XML: %w", err)
```
weikanglim, line 115: Slight improvement here:

```suggestion
		return nil, fmt.Errorf("parsing xml: %w", err)
```

cli/azd/internal/appdetect/java_test.go
```diff
@@ -0,0 +1,154 @@
...
+	project, err = jd.DetectProject(context.Background(), filepath.Join(tempDir, "submodule"), entries)
+	if err != nil {
+		t.Fatalf("unexpected error: %v", err)
+	}
+	if project == nil {
+		t.Fatalf("expected project to be detected, got nil")
+	}
+}
+
+func TestDetectDependencies_WithDatabaseDeps(t *testing.T) {
```
weikanglim, line 113: If it makes sense to you: once we update `appdetect_test.go` to include the dependencies, this will already be covered by the existing acceptance test coverage, so we could consider removing this test.

cli/azd/internal/appdetect/java_test.go
```diff
@@ -0,0 +1,154 @@
...
+)
+
+func TestJavaDetector_Language(t *testing.T) {
+	jd := &javaDetector{}
+	if jd.Language() != Java {
+		t.Errorf("expected language to be Java, got %v", jd.Language())
+	}
+}
+
+func TestJavaDetector_DetectProject_WithPomXml(t *testing.T) {
```
weikanglim, line 18: If it make sense to you: this is covered by the acceptance tests in `appdetect_test.go`, so we could consider removing it.

cli/azd/internal/appdetect/java_test.go
```diff
@@ -0,0 +1,154 @@
...
+	project, err := jd.DetectProject(context.Background(), tempDir, entries)
+	if err != nil {
+		t.Fatalf("unexpected error: %v", err)
+	}
+	if project == nil {
+		t.Fatal("expected project to be detected, got nil")
+	}
+}
+
+func TestJavaDetector_DetectProject_WithoutPomXml(t *testing.T) {
```
weikanglim, line 40: If it make sense to you: this code path is already covered by the acceptance tests in `appdetect_test.go` by the fact that we'd call a similar codepath on each sub directory.

weikanglim (approved):
This looks good to me, thanks @saragluna for all your hard work here.

Let's update the PR description to say something along the lines of: "Add Postgres and MySql detection", and "Add support for multi-module Java repositories here".

cli/azd/internal/appdetect/testdata/java-multimodules/application/build.gradle
saragluna: Yes, will do

cli/azd/internal/appdetect/java.go
saragluna: Got it, so we only need to return a formatted error with message in the uppermost api

PR #4449, opened 2024-10-16
func: add mitigation for deployments that take awhile to "come alive"

### LINKED ISSUE #4439: [Issue] AZD Publishing of Azure Functions should poll the publish for status and display the results
This is a feature request for AZD when publishing to Azure Functions to show the results of the publishing/deployment logs (especially in the Flex Consumption hosting plan with the new OneDeploy `/api/publish` endpoint, but it also applies to others in the Zip Deploy endpoint). 

For example, the AZD publish for one of my recent Flex Consumption apps that failed to publish just showed the following: 

ERROR: error executing step command 'deploy --all': failed deploying service 'processor': publishing zip file: GET https://pbatum-flexbad-wus2-net-no-sbconnstring.scm.azurewebsites.net/api/deployments/8ad0b7ed-7bd8-4fa0-8677-e38dcfb79a1a
RESPONSE 404: 404 Not Found
ERROR CODE UNAVAILABLE
Deployment 8ad0b7ed-7bd8-4fa0-8677-e38dcfb79a1a not found.
TraceID: f0a83e4a6b0967eb175a0e66485191df

But if I tried to publish using Azure Functions Core Tools (`func azure functionapp publish` command) I can see the actual exception in the logs it shows: "Failed to upload because Azure Storage responded with Forbidden : Response status code does not indicate success: 403 (This request is not authorized to perform this operation.)". With that info I was able to fix what was wrong with the app configuration and deploy properly.

Basically you can poll for the latest deployment status /api/deployments/latest and it will return an HTTP 202 while it’s still deploying and eventually an HTTP 200 when it’s successful, or an error if not. 

Core Tools
* Here it loops through to get the status once every interval: https://github.com/Azure/azure-functions-core-tools/blob/b02b386e02848dc9b165c08609105920b5fcc033/src/Azure.Functions.Cli/Helpers/KuduLiteDeploymentHelpers.cs#L54
( Then it uses the deployment ID to get the full logs (/api/deployments/{id}/log for Flex, and /deployments/{id}/log" for other) after the deployment completes to display the log:  https://github.com/Azure/azure-functions-core-tools/blob/b02b386e02848dc9b165c08609105920b5fcc033/src/Azure.Functions.Cli/Helpers/KuduLiteDeploymentHelpers.cs#L149

AZ CLI
* I think here's where it does the checking for updates, seems to do a little differently: https://github.com/Azure/azure-cli/blob/a9bec7bc52ed4e90c9f92451944b3640e45c3446/src/azure-cli/azure/cli/command_modules/appservice/custom.py#L5927

The Functions deployment team can help with more specifics and how this would work for non Flex Consumption as well.

weikanglim:
Add polling delay and retries for deployments that take awhile to "come alive"

Fixes #4439

### DIFF
--- a/cli/azd/pkg/azsdk/funcapp_host_client.go
+++ b/cli/azd/pkg/azsdk/funcapp_host_client.go
@@ -117,8 +117,11 @@ func (c *FuncAppHostClient) Publish(
 
 func (c *FuncAppHostClient) waitForDeployment(ctx context.Context, location string) (PublishResponse, error) {
 	// This frequency is recommended by the service team.
-	polLDelay := 1 * time.Second
+	pollDelay := 1 * time.Second
 	var lastResponse *PublishResponse
+
+	// mitigation for deployments that take awhile to "come alive"
+	deploymentNotFoundAttempts := 0
 
 	for {
 		req, err := runtime.NewRequest(ctx, http.MethodGet, location)
@@ -129,6 +132,12 @@ func (c *FuncAppHostClient) waitForDeployment(ctx context.Context, location stri
 		response, err := c.pipeline.Do(req)
 		if err != nil {
 			return PublishResponse{}, err
+		}
+
+		if deploymentNotFoundAttempts <= 3 && response.StatusCode == http.StatusNotFound {
+			deploymentNotFoundAttempts++
+			time.Sleep(pollDelay)
+			continue
 		}
 
 		// It's possible to observe a 404 response after the deployment is complete.
@@ -167,9 +176,9 @@ func (c *FuncAppHostClient) waitForDeployment(ctx context.Context, location stri
 		// Record the latest response
 		lastResponse = &resp
 
-		delay := polLDelay
+		delay := pollDelay
 		if retryAfter := httputil.RetryAfter(response); retryAfter > 0 {
-			delay = polLDelay
+			delay = pollDelay
 		}
 
 		select {


cli/azd/pkg/azsdk/funcapp_host_client.go
```diff
@@ -117,9 +117,12 @@ func (c *FuncAppHostClient) Publish(
 
 func (c *FuncAppHostClient) waitForDeployment(ctx context.Context, location string) (PublishResponse, error) {
 	// This frequency is recommended by the service team.
-	polLDelay := 1 * time.Second
+	pollDelay := 1 * time.Second
```
wbreza, line 120: Have we considered using the `go-retry` package that we have used in the past? It seems to already support the polling logic with max retries, etc.

wbreza (approved):
Looks fine - just added one comment.

patelchandni (approved):
Looks good.

cli/azd/pkg/azsdk/funcapp_host_client.go
weikanglim: My preference would be:

1. Consider azure-sdk-for-go retry policies
2. If implementing HTTP directly, consider handling error conditions directly.

My main hesitancies against `go-retry` is that 1. It abstracts away for-loops, which "hides" the true performance cost from the programmer, and 2. It makes it too easy to add inner retries and outer retries that lead to exponential retry loops.

This is primarily speaking from my own experience on previous teams, where a client tool ended up retrying for an hour, because we added multiple retries at different layers. They all looked fine going through code review -- but it's only when you stare at a for-loop long enough that we understand the cost of execution that we are imposing on the program.

Issue #4440, opened 2024-10-16
panic: runtime error: invalid memory address or nil pointer dereference


mitchdenny:
- [x] Make sure you've installed the latest version using [instructions in the wiki](../wiki/install)

**Output from `azd version`**
Run `azd version` and copy and paste the output here:

```
azd version 1.10.2 (commit 84ea22048b9a1c167bc4ae2095cc7f91798ef11e)
```

**Describe the bug**
AZD is crashing when I am trying to deploy a .NET Aspire 8.2.1 project. Here is the error I get:

```
PS C:\Code\postgresbug\PostgresBug.AppHost> azd up

Packaging services (azd package)


Provisioning Azure resources (azd provision)
Provisioning Azure resources can take some time.

Subscription: midenn (9ab12ed0-87cd-4402-9e18-9b67142c6c30)
Location: East US 2

  (-) Skipped: Didn't find new changes.

Deploying services (azd deploy)

  |=======| Deploying service apiservice (Updating container app)panic: runtime error: invalid memory address or nil pointer dereference
[signal 0xc0000005 code=0x0 addr=0x30 pc=0xc6619e]

goroutine 1 [running]:
github.com/azure/azure-dev/cli/azd/pkg/containerapps.(*containerAppService).persistSettings(0xc0001f77d0, {0x155c780, 0xc0006376e0}, {0xc0001e6720, 0x24}, {0xc000512170, 0xf}, {0xc000496d80, 0xa}, 0xc000608ff0, ...)
        github.com/azure/azure-dev/cli/azd/pkg/containerapps/container_app.go:154 +0x23e
github.com/azure/azure-dev/cli/azd/pkg/containerapps.(*containerAppService).DeployYaml(0xc0001f77d0, {0x155c780, 0xc0006376e0}, {0xc0001e6720, 0x24}, {0xc000512170, 0xf}, {0xc000496d80, 0xa}, {0xc000689000, ...}, ...)
        github.com/azure/azure-dev/cli/azd/pkg/containerapps/container_app.go:187 +0x15c
github.com/azure/azure-dev/cli/azd/pkg/project.(*dotnetContainerAppTarget).Deploy(0xc000492a20, {0x155c780, 0xc0006376e0}, 0xc00063e488, 0xc000399590, 0xc000055540, 0xc00010c860)       
        github.com/azure/azure-dev/cli/azd/pkg/project/service_target_dotnet_containerapp.go:401 +0x19a9
[...]

PR #4450, opened 2024-10-16
fix persistSettings regression

vhvb1989:
fix: https://github.com/Azure/azure-dev/issues/4440

Regression analysis.
For azd version 1.10.2, we update the implementation details or deploying container apps (see: https://github.com/Azure/azure-dev/pull/4272).
As part of the change, the logic for persisting existing settings in Aspire deployment changed.
In the past, the logic was validating that an existing aca was not nil, as well as the properties to persist.
This validation was removed as part of the change. See: https://github.com/Azure/azure-dev/pull/4272/files#diff-d6903c0cec59e731280be935e1056d2c8354ade3d82eaa6bfb3652ea31b748afL134-L140

This PR adds a quick exit from persisting settings when the aca is not found or there's an error trying to get it.


cli/azd/pkg/containerapps/container_app.go
```diff
@@ -146,6 +146,7 @@ func (cas *containerAppService) persistSettings(
 	aca, err := cas.getContainerApp(ctx, subscriptionId, resourceGroupName, appName, options)
 	if err != nil {
 		log.Printf("failed getting current aca settings: %v. No settings will be persisted.", err)
+		return obj, nil
```
weikanglim, line 149: This is slightly better and probably fixes the regression... But I'm wondering if this also opens us up to weird future problems that we want to avoid entirely?

Imagine if we issued a `GET` request that results in 503 Server Unavailable, we end up not persisting the settings here, whereas I could imagine the right behavior is to fail, since we can't evaluate the diff?

I'm unsure about the background here behind the "soft failure" here -- are we trying to express that a 404 should be "no changes necessary here?"

ellismg: I'm in agreement with @weikanglim - my understanding of this function is that it's trying to merge a subset of properties from the existing container app (if it exists) with the desired state expressed by the YAML because we want to allow certain out of band edits to be applied over the generated desired state (to save folks the trouble of `infra synth`'ing and then editing the yaml themselves).  In Aspire 9 this is no longer needed because the customer can edit these properties directly in the app host (and if they wanted do similar logic to what we have in in C# and decide how they want to handle each failure).

For our case I feel like we should treat 404 as "we're creating this app so there's nothing to merge into the desired state" and other errors as "we can't be sure what's going on so we should conservatively fail the update with an error message".

I think that `persistSettings` is sort of a bad name at this point, maybe `mergePersistedSettingsFromExistingApp` or something to it's a little clearer from the function name what exactly this is intended to do?

vhvb1989: I thought about creating a unique error for `Not Found` . But this is an alpha feature which is planned to be retired/deprecated soon. 
Aspire already provides in version 9 a way to model the ACA settings directly in the AppHost and there is no need to ask azd to persist previous settings.
Actually, when moving to Aspire9 + `PublishAsAzureContainerApp()` like here: https://github.com/dotnet/aspire/blob/main/playground/AzureContainerApps/AzureContainerApps.AppHost/Program.cs#L29C9-L29C35 the `persisting feature` from azd becomes ignored/no-op.

We want anyone currently depending on persist settings with azd to use Aspire9.

That made me not wanting to invest more here into a better implementation for this.

ellismg: If we do go down this line then I could image a helpful comment above the `return obj, nil` along the lines of "if the container app doesn't exist, there's nothing for us to update in the desired state, so we can just return the existing state as is."

ellismg (approved):
This seems like a good quick fix to unblock folks but I do think we should revisit (in another PR later) what the error behavior is for non 404 errors.

vhvb1989:
> This seems like a good quick fix to unblock folks but I do think we should revisit (in another PR later) what the error behavior is for non 404 errors.

When azd tries to get the ACA, we use the retry policy from azure-core for non-404 errors, like the 5xx..  If the user gets more non-404 errors for more than 3 times, there would be something else going on beyond this only feature and more things are expected to fail in cascade

PR #4454, opened 2024-10-17
Auto detect port from Dockerfile

### LINKED ISSUE #4443: easy-init: azd should auto detect port from Dockerfile
After discussion with @puicchan:

1. Look in Dockerfile
2. If EXPOSE with single port is found, use it and don't ask
3. If EXPOSE is not found, ask the customer.
4. If more than one port is exposed, prompt for a selection (`console.Select`) with the multiple ports available, and an "Other" selection to provide the port number.
   
Multiple ports example:
```bash
Select a port number:
- 5050 
- 8080
- Other
```

rujche:
Resolve #4443
Auto detect port from Dockerfile


cli/azd/internal/repository/infra_confirm.go
```diff
@@ -107,6 +108,11 @@ func (i *Initializer) infraSpecFromDetect(
 			if svc.Language == appdetect.Java {
 				serviceSpec.Port = 8080
 			}
+		} else {
+			ports := i.detectPortInDockerfile(svc.Docker.Path)
+			if len(ports) == 1 {
+				serviceSpec.Port = ports[0]
+			}
```
rujche, line 115: @weikanglim 

In the issue description of #4443, you mentioned that

> 3. If EXPOSE is not found, default the port for customer.

I don't know what the `default the port` is, so I just keep original behavior: Not set the port, `ask the customer`.

rujche:
Hi, @saragluna , @weikanglim , @puicchan.
Please help to review this PR when you have time.

cli/azd/internal/repository/infra_confirm.go
```diff
@@ -1,8 +1,10 @@
 package repository
```
jongio, line 1: Can we also inspect any docker-compose.y*ml files to see if there's a port mapping there as well?

cli/azd/internal/repository/infra_confirm.go
```diff
@@ -194,3 +200,26 @@ func (i *Initializer) infraSpecFromDetect(
...
+		line := scanner.Text()
+		if strings.HasPrefix(line, "EXPOSE") {
+			var port int
+			_, err := fmt.Sscanf(line, "EXPOSE %d", &port)
+			if err == nil {
+				result = append(result, port)
+			}
+		}
+	}
+	return result
```
wbreza, line 224: Instead of us manually reading the dockerfile to discover the exposed port I would suggest we look into consuming existing OSS Go modules that already provide this functionality.

A few to consider are the following:

**moby/buildkit:**
Part of the Moby project (the same project that develops Docker itself), the BuildKit library can be used to parse Dockerfiles. It provides low-level access to the Dockerfile AST (Abstract Syntax Tree) and is widely used in Docker tooling.

**lucor/dockerfile-ast:**
A lighter and simpler AST parser for Dockerfiles with an easy-to-use API to extract metadata from Dockerfiles.

### Example
```go
import (
    "bytes"
    "fmt"
    "github.com/moby/buildkit/frontend/dockerfile/parser"
)

func getExposedPorts(dockerfileContent []byte) ([]string, error) {
    reader := bytes.NewReader(dockerfileContent)
    result, err := parser.Parse(reader)
    if err != nil {
        return nil, err
    }

    var ports []string
    // Loop through the parsed Dockerfile instructions
    for _, child := range result.AST.Children {
        if child.Value == "expose" {
            for _, next := range child.Next {
                ports = append(ports, next.Value) // Get the port value
            }
        }
    }
    return ports, nil
}
```

cli/azd/internal/repository/infra_confirm.go
weikanglim: @jongio I'm wondering if my understanding is correct: would docker-compose describe the application workload, and not a given service? If docker-compose is referring to the app workload, I think we may need to address this separately as "init from docker-compose.yml".

cli/azd/internal/repository/infra_confirm.go
weikanglim: Just offering a different perspective to see how we feel. As the maintainer of `appdetect` package, I think I'd be okay with the current approach, given how lightweight and simple the [EXPOSE](https://docs.docker.com/reference/dockerfile/#expose) syntax is.

The one thing I'd be slightly concerned about introducing AST analysis at the current stage is that `azd` would potentially need to either handle or throw parse errors for `Dockerfile` that it doesn't understand. When these errors occur, it gives a somewhat false impression to the user that `azd` understands Dockerfile at the deeper level than it needs to.

One guiding principle in `appdetect` is that it's fast, and it is minimal -- it will only grep metadata that it understands and ignores the rest. If you have an invalid Dockerfile, that's okay -- we'll grep the file, but not analyze the file for semantic or syntactic correctness. We're not trying to build a Dockerfile, we're just trying to grep to get the few things we needed.

In certain areas in the future, we may go deeper if we think the technology is interesting, but I would recommend staying a little light for Dockerfile for now.

cli/azd/internal/repository/infra_confirm.go
```diff
@@ -107,6 +108,11 @@ func (i *Initializer) infraSpecFromDetect(
 			if svc.Language == appdetect.Java {
 				serviceSpec.Port = 8080
 			}
+		} else {
+			ports := i.detectPortInDockerfile(svc.Docker.Path)
```
weikanglim, line 112: @rujche Let's move this logic into `appdetect` package. The concrete place I'm thinking is insider `docker.go`. We can update `Docker` to include the "ExposedPorts" property as part of the analysis.

cli/azd/internal/repository/infra_confirm.go
```diff
@@ -194,3 +200,26 @@ func (i *Initializer) infraSpecFromDetect(
...
+	}
+	defer file.Close()
+
+	var result []int
+	scanner := bufio.NewScanner(file)
+	for scanner.Scan() {
+		line := scanner.Text()
+		if strings.HasPrefix(line, "EXPOSE") {
+			var port int
+			_, err := fmt.Sscanf(line, "EXPOSE %d", &port)
```
weikanglim, line 218: This looks great!

For completeness, I'd make sure to handle:
- Multiple ports
- Transmission protocol (udp/tcp) being present

See [EXPOSE](https://docs.docker.com/reference/dockerfile/#expose) for reference.

weikanglim (commented):
@rujche Thanks for getting this started! I left two blocking comments:

 - Move the logic into `appdetect` and surface metadata that way
 - Make sure to handle the full EXPOSE syntax

cli/azd/internal/repository/infra_confirm.go
weikanglim: Asking the port is indeed the right behavior! I updated the issue

weikanglim: @rujche I also updated the multiple ports requirement:

4. If more than one port is exposed, prompt for a selection (`console.Select`) with the multiple ports available, and an "Other" selection to provide the port number.
   
Multiple ports example:
```bash
Select a port number:
- 5050 
- 8080
- Other
```

cli/azd/internal/repository/infra_confirm.go
jongio: I was referring to checking the services in docker-compose.yml to see if any have port mappings we need to address.  They are usually the same but sometimes different.  It's fine to consider this later.

cli/azd/internal/appdetect/docker.go
```diff
@@ -1,19 +1,69 @@
...
+				Path:         dockerFilePath,
+				ExposedPorts: exposedPorts,
 			}, nil
 		}
 	}
 
 	return nil, nil
 }
+
+func getExposedPorts(dockerfilePath string) (map[int]string, error) {
```
weikanglim, line 28: Let's think about breaking down this functionality differently. `getExposedPorts` right now takes the entire file, I think we could step down into a helper function that takes a line instead. I like the idea of `parsePorts(s string) ([]port, error)`.

The way I'm thinking about this: If we had to parse a different line of syntax, let's say, `RUN ...`, it wouldn't be easy to implement this additional feature without introducing additional changes. Breaking it down in the right way will also help us provide unit test coverage over the algorithm we care about: port parsing, without creating heavyweight tests that exercise the full functionality. Each test case is just a different input string, instead of a different input file.

My recommendations would be something along the following lines:

1. In `detectDocker`, let's open the file and close it for reading.
2. We read line by line using the scanner
3. If `line` has prefix `EXPOSE` -- we know it's an expose directive. We invoke `parsePorts(line[len("EXPOSE"):)` to parse the ports in the remainder of the line.

Note that `len("EXPOSE")` will be optimized by the compiler into a constant, but if we feel that the code is too verbose, we could also parse the first WORD / Directive from the line -- but I would recommend to keep the program simple at this point.

cli/azd/internal/appdetect/docker.go
```diff
@@ -1,19 +1,69 @@
...
+	if err := scanner.Err(); err != nil {
+		return nil, err
+	}
+
+	return ports, nil
+}
+
+type Docker struct {
+	Path         string
+	ExposedPorts map[int]string
```
weikanglim, line 68: I'm okay with introducing a struct here for `type Port struct { Number int Protocol string }`

cli/azd/internal/appdetect/docker.go
```diff
@@ -1,19 +1,69 @@
...
+	"regexp"
+	"strconv"
 	"strings"
 )
 
 func detectDocker(path string, entries []fs.DirEntry) (*Docker, error) {
 	for _, entry := range entries {
 		if strings.ToLower(entry.Name()) == "dockerfile" {
+			dockerFilePath := filepath.Join(path, entry.Name())
+			exposedPorts, _ := getExposedPorts(dockerFilePath)
```
weikanglim, line 17: Let's do propagate the `err` returned here. In general, if we are trying to suppress errors, we can consider adding a comment on why this suppression is necessary.

Imagine if the filesystem had a temporary disk error reading the file temporarily. We want the detection to halt because of that, instead of saying "I couldn't find ports" -- which on the next re-run, would produce a different set of results.

cli/azd/internal/repository/infra_confirm.go
```diff
@@ -103,10 +102,16 @@ func (i *Initializer) infraSpecFromDetect(
 		if svc.Docker == nil || svc.Docker.Path == "" {
 			// default builder always specifies port 80
 			serviceSpec.Port = 80
-
 			if svc.Language == appdetect.Java {
 				serviceSpec.Port = 8080
 			}
+		} else {
+			ports := svc.Docker.ExposedPorts
```
weikanglim, line 109: Just leaving this as a note, when you get to this requirement:

If more than one port is exposed, prompt for a selection (console.Select) with the multiple ports available, and an "Other" selection to provide the port number.

Feel free to ping me directly on Teams if you have any questions here.

weikanglim (commented):
Thanks @rujche ! This is looking better, thanks for adopting the suggestion to move the detection logic into `appdetect` package.

I left some comments for the next review iteration.

cli/azd/internal/appdetect/docker_test.go
```diff
@@ -0,0 +1,58 @@
+package appdetect
+
+import (
+	"github.com/stretchr/testify/require"
+	"path/filepath"
+	"reflect"
+	"testing"
+)
+
+func TestDetectPortInDockerfile(t *testing.T) {
```
weikanglim, line 10: Let's revisit how this test should look like, once we think about testing parsing a string instead of an entire file.

I'd recommend taking full advantage of go's table-driven tests to concisely represent the different test cases. Here's an example taking from `scaffold/funcs_test.go`, I'd imagine that we would be able to structure it similarly.

```golang
func Test_containerAppName(t *testing.T) {
	tests := []struct {
		name string
		in   string
		want string
	}{
		{"allowed characters", "MyApp_!#%^", "myapp"},
		{"dash at front or end", "-my-app-", "my-app"},
		{"multiple dashes", "my----app", "my-app"},
		{"at length", "123456789app", "123456789app"},
		{"over length", "123456789my-app", "123456789my"},
	}
	for _, tt := range tests {
		t.Run(tt.name, func(t *testing.T) {
			actual := containerAppName(tt.in, 12)
			assert.Equal(t, tt.want, actual)
		})
	}
}
```

cli/azd/internal/appdetect/docker.go
```diff
@@ -1,19 +1,69 @@
...
+func getExposedPorts(dockerfilePath string) (map[int]string, error) {
+	file, err := os.Open(dockerfilePath)
+	if err != nil {
+		return nil, err
+	}
+	defer file.Close()
+
+	ports := make(map[int]string)
+	scanner := bufio.NewScanner(file)
+	exposeRegex := regexp.MustCompile(`^EXPOSE\s+(.+)$`)
```
weikanglim, line 37: Let's avoid using regex here if we can. Some ideas from the other [comment](https://github.com/Azure/azure-dev/pull/4454/files#r1806657892).

cli/azd/internal/appdetect/docker.go
```diff
@@ -1,19 +1,69 @@
...
+		return nil, err
+	}
+
+	return ports, nil
+}
+
+type Docker struct {
+	Path         string
+	ExposedPorts map[int]string
+}
```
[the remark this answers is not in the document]
rujche: Updated.

cli/azd/internal/appdetect/docker.go
rujche: Updated.

cli/azd/internal/appdetect/docker.go
rujche: Updated.

cli/azd/internal/appdetect/docker.go
rujche: Updated.

cli/azd/internal/repository/infra_confirm.go
rujche: Updated.

cli/azd/internal/appdetect/docker_test.go
rujche: Udpated.

cli/azd/internal/appdetect/docker.go
rujche: Updated.

rujche:
Hi, @weikanglim .
I add some new commits, please help to review again.

cli/azd/internal/appdetect/docker.go
```diff
@@ -1,19 +1,67 @@
...
 		}
 	}
 
 	return nil, nil
 }
+
+func parsePorts(s string) ([]Port, error) {
+	s = strings.TrimSpace(s)
+	var ports []Port
+	portSpecs := strings.Split(s, " ")
```
weikanglim, line 48: I think we have to handle handle other whitespace characters, i.e. `\t`, `\r`. Luckily, there's `strings.Fields` to parse any whitespace delimited text. This also removes the need to perform `strings.TrimSpace` on line 46.

cli/azd/internal/appdetect/docker.go
```diff
@@ -1,19 +1,67 @@
...
+	for _, portSpec := range portSpecs {
+		if strings.Contains(portSpec, "/") {
+			parts := strings.Split(portSpec, "/")
+			portNumber, err := strconv.Atoi(parts[0])
+			if err != nil {
+				return nil, err
```
weikanglim, line 54: Add some context to this error:
```suggestion
				return nil, fmt.Errorf("parsing port number: %w", err)
```

cli/azd/internal/appdetect/docker.go
```diff
@@ -1,19 +1,67 @@
...
+			}
+			protocol := parts[1]
+			ports = append(ports, Port{portNumber, protocol})
+		} else {
+			portNumber, err := strconv.Atoi(portSpec)
+			if err != nil {
+				return nil, err
```
weikanglim, line 61: Add some context to this error:
```suggestion
				return nil, fmt.Errorf("parsing port number: %w", err)
```

cli/azd/internal/appdetect/docker.go
```diff
@@ -1,19 +1,67 @@
...
+			defer file.Close()
+			scanner := bufio.NewScanner(file)
+
+			var ports []Port
+			for scanner.Scan() {
+				line := scanner.Text()
+				if strings.HasPrefix(line, "EXPOSE") {
+					parsedPorts, err := parsePorts(line[len("EXPOSE"):])
+					if err != nil {
+						return nil, err
```
weikanglim, line 29: Looking at this once more, I now realize that your original implementation of not failing when we fail to parse `EXPOSE` is much better. We end up falling back to prompting the user.

My suggestion would be to log the error and move on:
```suggestion
				log.Errorf("parsing Dockerfile at %s: %v", dockerFilePath, err)
```

cli/azd/internal/appdetect/docker.go
```diff
@@ -1,19 +1,67 @@
...
 	"strings"
 )
 
 func detectDocker(path string, entries []fs.DirEntry) (*Docker, error) {
 	for _, entry := range entries {
 		if strings.ToLower(entry.Name()) == "dockerfile" {
+			dockerFilePath := filepath.Join(path, entry.Name())
+			file, err := os.Open(dockerFilePath)
+			if err != nil {
+				return nil, err
```
weikanglim, line 18: let's add context to the error here.
```suggestion
				return nil, fmt.Errorf("reading Dockerfile at %s: %w", dockerFilePath, err)
```

cli/azd/internal/appdetect/docker_test.go
```diff
@@ -0,0 +1,25 @@
...
+	}{
+		{"", nil},
+		{"80", []Port{{80, "tcp"}}},
+		{"80 3100", []Port{{80, "tcp"}, {3100, "tcp"}}},
+		{"80 3100/udp", []Port{{80, "tcp"}, {3100, "udp"}}},
+		{" 80/tcp 3100/udp ", []Port{{80, "tcp"}, {3100, "udp"}}},
+	}
+	for _, tt := range tests {
+		t.Run(tt.portString, func(t *testing.T) {
+			actual, _ := parsePorts(tt.portString)
```
weikanglim, line 21: We can also assert that

`assert.NoError(t, err)`

just for extra validation.

cli/azd/internal/appdetect/appdetect_test.go
```diff
@@ -193,7 +193,8 @@ func TestDetectDocker(t *testing.T) {
 		Path:          filepath.Join(dir, "dotnet"),
 		DetectionRule: "Inferred by presence of: dotnettestapp.csproj, Program.cs",
 		Docker: &Docker{
-			Path: filepath.Join(dir, "dotnet", "Dockerfile"),
+			Path:  filepath.Join(dir, "dotnet", "Dockerfile"),
+			Ports: nil,
```
weikanglim, line 197: Let's also update this test case to write different port handling conditions `Dockerfile`:

I recommend creating a new subtest within `TestDetectDocker` with:

```golang
t.Run("exposed ports", func(t *testing.T) {
})
```

I would think about using the same type of table-driven tests you had written for `docker_test.go`. Each test case specifies the `exposeLine` to write to the Dockerfile to set up the test case:

- Single port
- Multiple ports on single line
- Multiple ports on multiple lines
- No ports

cli/azd/internal/repository/infra_confirm.go
```diff
@@ -103,10 +102,45 @@ func (i *Initializer) infraSpecFromDetect(
 		if svc.Docker == nil || svc.Docker.Path == "" {
 			// default builder always specifies port 80
 			serviceSpec.Port = 80
-
 			if svc.Language == appdetect.Java {
 				serviceSpec.Port = 8080
 			}
+		} else {
+			ports := svc.Docker.Ports
```
weikanglim, line 109: Since we're also doing the prompting here -- let's think about moving the logic from L179-L203 here.

weikanglim: I think we can apply consistency to the messages and make it simpler.

Currently we have:
image

This should look more like: 

```bash
? What port does 'dotnet' listen on?
3100
3233
Other
```

I think it's okay to omit how we did the detection for now. Perhaps in the future if we do have user asks, we can decide how to best present this information.

weikanglim (commented):
@rujche Love the great progress here! I left a few more comments based on what I noticed, but I think we're really close.

I'm really happy with how much we have simplified the implementation -- it looks a lot cleaner!

cli/azd/internal/appdetect/docker.go
rujche: Updated.

cli/azd/internal/appdetect/docker.go
rujche: Updated.

cli/azd/internal/appdetect/docker.go
rujche: Updated.

cli/azd/internal/appdetect/docker.go
rujche: Updated.

cli/azd/internal/appdetect/docker.go
```diff
@@ -1,19 +1,67 @@
...
+			}
+			defer file.Close()
+			scanner := bufio.NewScanner(file)
+
+			var ports []Port
+			for scanner.Scan() {
+				line := scanner.Text()
+				if strings.HasPrefix(line, "EXPOSE") {
```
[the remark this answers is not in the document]
rujche: Updated.

cli/azd/internal/appdetect/docker.go
rujche: Updated.

cli/azd/internal/appdetect/docker_test.go
rujche: Updated.

cli/azd/internal/appdetect/appdetect_test.go
rujche: Added `TestParsePortsInFile` in `docker_test.go`

cli/azd/internal/repository/infra_confirm.go
rujche: Updated.

rujche:
Hi, @weikanglim 
Thank you for your great comments! 👍 
I added some new commits according to your comments, please help to review again.

cli/azd/internal/repository/infra_confirm.go
```diff
@@ -103,10 +102,43 @@ func (i *Initializer) infraSpecFromDetect(
...
+			if len(ports) == 0 {
+				port, err := i.getPortByPrompt(ctx, serviceSpec.Name)
```
weikanglim, line 111: Let's keep the experience the same for this initial prompt, i.e. the prompt message should be:

"What port does '" + serviceSpec.Name + "' listen on?

cli/azd/internal/appdetect/docker_test.go
```diff
@@ -0,0 +1,49 @@
...
+		})
+	}
+}
+
+func TestParsePortsInFile(t *testing.T) {
+	tests := []struct {
+		dockerFileName string
+		expectedPorts  []Port
+	}{
+		{"DockerfileNoPort", nil},
```
weikanglim, line 34: This looks good! One final comment here: instead of using files to capture the different test conditions, can we consider writing the EXPOSE lines here? 

```golang
tests := []struct {
		dockerFile string
		expectedPorts  []Port
	}{
		{"EXPOSE 80/tcp", []Port{{80, "tcp"}}},
		{"EXPOSE 80/tcp 3100/tcp", []Port{{80, "tcp"}, {3100, "tcp"}}},
		{"EXPOSE 80/tcp\n3100/tcp", []Port{{80, "tcp"}, {3100, "tcp"}}},
		{". EXPOSE 80/tcp\n3100/udp", []Port{{80, "tcp"}, {3100, "udp"}}},
}
```

Then we generate the file into a temp directory, and call `parsePortsInFile`. Keep in mind that these are actually perfectly valid (syntax-wise) format Dockerfiles, so there's some power in writing in this way. We could for example, at some point, decide to create variables to read from elsewhere, or just to keep it within the file -- any way works.

In the current implementation, I think we'd end up with *too many* files if we were to write test asset files for each test case. My proposal is trying to make it more concise to read and to maintain.

cli/azd/internal/appdetect/docker.go
```diff
@@ -1,19 +1,75 @@
...
+				Path:  dockerFilePath,
+				Ports: ports,
 			}, nil
 		}
 	}
 
 	return nil, nil
 }
+
+func parsePortsInFile(dockerFilePath string) ([]Port, error) {
```
weikanglim, line 32: I think I liked the previous iteration, where we didn't separate out `parsePortsInFile` as a function.

The main reason why I'd prefer the old version is that in this new version, we would introduce code churn (additional background changes) when we say, for example, parse `RUN` from the Dockerfile.

If we move the logic back in `detectDocker`, I'd feel much better about it.

cli/azd/internal/appdetect/docker.go
```diff
@@ -1,19 +1,75 @@
...
 	"strings"
 )
 
 func detectDocker(path string, entries []fs.DirEntry) (*Docker, error) {
 	for _, entry := range entries {
 		if strings.ToLower(entry.Name()) == "dockerfile" {
+			dockerFilePath := filepath.Join(path, entry.Name())
+			ports, err := parsePortsInFile(dockerFilePath)
+			if err != nil {
+				return nil, fmt.Errorf("parsing ports in file %s: %w", dockerFilePath, err)
```
weikanglim, line 20: We should just return without error formatting here:

```suggestion
				return nil, err
```

In the current implementation, this looks like:

```bash
parsing ports in file testdata/Dockerfile/DockerfileSinglePort: reading Dockerfile at testdata/Dockerfile/DockerfileSinglePort
```

which isn't quite right.

weikanglim (commented):
Overall looks really good! Narrowed it down a few more comments to address. We are getting real close!

cli/azd/internal/repository/infra_confirm.go
weikanglim: For the other flow, where we did detect multiple ports, what you have looks good.
In that flow, it looks like:

> $ What port does 'api' listen on? Other
> $ Provide the port number for 'api': 80

which looks perfectly fine!

rujche: Updated.

cli/azd/internal/appdetect/docker_test.go
rujche: Updated.

cli/azd/internal/appdetect/docker.go
rujche: Updated.

cli/azd/internal/appdetect/docker.go
rujche: Updated.

rujche:
Hi, @weikanglim 
Thanks for the quick response.
I added some new commits according to your comments. Please help to review again.

weikanglim (approved):
This looks good, @rujche! Thank you for taking the time to address all the comments.

I do really like how the final change looks like. It's very clean and feature rich. I think our users are going to love this feature addition!

Thank you!

rujche:
@weikanglim 
Thank you for your review and your comments, I learnt a lot from your comments. ❤️

PR #4457, opened 2024-10-17
aspire: Write container registry name as an output

ellismg:
It can be helpful to have just the bare name of the container registry (without the suffix (e.g. `azurecr.io`)) when orchestrating things together.

This change adds `AZURE_CONTAINER_REGISTRY_NAME` as an output when `azd` creates a container registry.


weikanglim (approved):

ellismg:
@microsoft-github-policy-service rerun

PR #4469, opened 2024-10-21
WIP: container: Trim spaces in computed image name

ellismg:


cli/azd/pkg/project/container_helper.go
```diff
@@ -62,9 +62,9 @@ func NewContainerHelper(
 // DefaultImageName returns a default image name generated from the service name and environment name.
 func (ch *ContainerHelper) DefaultImageName(serviceConfig *ServiceConfig) string {
 	return fmt.Sprintf("%s/%s-%s",
-		strings.ToLower(serviceConfig.Project.Name),
-		strings.ToLower(serviceConfig.Name),
-		strings.ToLower(ch.env.Name()))
+		strings.TrimSpace(strings.ToLower(serviceConfig.Project.Name)),
+		strings.TrimSpace(strings.ToLower(serviceConfig.Name)),
+		strings.TrimSpace(strings.ToLower(ch.env.Name())))
```
vhvb1989, line 67: I'd prefer to make this function panic or return an error if the project name, env name or config name are using invalid characters to create a valid image name.

Issue #4470, opened 2024-10-21
[Aspire] Fail deployment when using a folder with spaces in its name


vhvb1989:
We might have broken Aspire projects after updating the name generation here: https://github.com/Azure/azure-dev/pull/4403

Haven't tested if before this change we were able to use a folder with spaces in its name, but the change is not handling spaces, which ends up in Aspire generating invalid image names (with spaces)

Issue #4400, opened 2024-10-02
define: naming for `azure.yaml` project and services


weikanglim:
In future azd features, azd will start using the names of services (or elements in azure.yaml) as command targets.

It's important for us to think about what allowed naming characters we should allow. For example, if we allow `.` characters (which currently breaks parts of azd today -- such as in #3183), we'd have to think about if we ever use `.` as a property accessor. Example: `myapp.api.env` would end up being ambiguous for interpretation without escaping.

## Early design ideas

We have some prior-art for inspiration from DNS systems, which also influences K8s object names:

DNS Subdomain Names [RFC 1123](https://tools.ietf.org/html/rfc1123)
1. contain no more than 253 characters
2. contain only lowercase alphanumeric characters, '-' or '.'
3. start with an alphanumeric character
4. end with an alphanumeric character

RFC 1123 Label Names - [RFC 1123](https://tools.ietf.org/html/rfc1123).

1. contain at most 63 characters
2. contain only lowercase alphanumeric characters or '-'
3. start with an alphanumeric character
4. end with an alphanumeric character

RFC 1035 Label Names -  [RFC 1035](https://tools.ietf.org/html/rfc1035).

1. contain at most 63 characters
2. contain only lowercase alphanumeric characters or '-'
3. start with an alphabetic character
4. end with an alphanumeric character

From how azd is commonly used today, we seem to lean towards RFC 1123 label names -- lowercase, dash-separated, alphanumeric characters.

## Use cases to consider

The main question to consider is that whether all language developers will overall be comfortable with whatever naming scheme we choose. 

I'm opening up this issue for discussion, hoping we can arrive at some level of consensus. The next phase of implementation is likely to implement the naming scheme as a recommendation in the short/near term, and eventually as a restriction in the longer term.

weikanglim:
With #4403, I'm adding logic for default name generation in azd-init to match RFC 1123 label names as a point of experimentation and gathering user feedback.

weikanglim:
> It would be great if they were _not_ case-sensitive (unless there's some docker reason why they need be?) 

 _Originally posted by @brandonh-msft in [#3183](https://github.com/Azure/azure-dev/issues/3183#issuecomment-2407800882)_

rajeshkamal5050:
This issue has been open for over 1 year without substantive activity — no comments, no linked PRs, and no description updates.

To keep our backlog healthy and actionable, we're closing it as not planned.

[...]

PR #4472, opened 2024-10-21
fix projects with empty spaces aspire

vhvb1989:
Re-use the same logic to name a project from easy-init for the AppHost's project name generation
fix: https://github.com/Azure/azure-dev/issues/4470


cli/azd/pkg/apphost/generate.go
```diff
@@ -439,7 +440,7 @@ func GenerateProjectArtifacts(
 	generatedFS := memfs.New()
 
 	projectFileContext := genProjectFileContext{
-		Name: projectName,
+		Name: label.LabelName(projectName),
```
weikanglim, line 443: I think I'd prefer to do this at the layer where we pass in the name (see list of files below). My inclination here is that name sanitization/restriction enforcement is really important to be done at closest to the user interaction layer as possible.

- [environment_service_create.go](https://github.com/Azure/azure-dev/blob/4182b4138151b8fac4e9d5e6f6220b32dddace80/cli/azd/internal/vsrpc/environment_service_create.go#L73)
- [app_init.go](https://github.com/Azure/azure-dev/blob/4182b4138151b8fac4e9d5e6f6220b32dddace80/cli/azd/internal/repository/app_init.go#L176).

vhvb1989: moved the logic to `GetDefaultProjectName()`

cli/azd/internal/vsrpc/environment_service_create.go
```diff
@@ -67,7 +67,7 @@ func (s *environmentService) CreateEnvironmentAsync(
 			return false, fmt.Errorf("reading app host manifest: %w", err)
 		}
 
-		projectName := strings.TrimSuffix(filepath.Base(c.azdContext.ProjectDirectory()), ".AppHost")
+		projectName := strings.TrimSuffix(c.azdContext.GetDefaultProjectName(), ".AppHost")
```
weikanglim, line 70: Just wondering, would this code still make sense the refactor to `GetDefaultProjectName`? `GetDefaultProjectName` is now an implementation detail that we then use to trim the`.AppHost` suffix?

cli/azd/pkg/apphost/generate.go
```diff
@@ -431,6 +431,13 @@ func GenerateProjectArtifacts(
 	manifest *Manifest,
 	appHostProject string,
 ) (map[string]ContentsAndMode, error) {
+	// panic if projectName contains spaces. This is because the project name is used to generate the name of containers
+	// and other resources, and spaces are not allowed in those names.
+	if strings.Contains(projectName, " ") {
```
weikanglim, line 436: Just wondering, how would we maintain this code over time?

There are many, many other disallowed characters. `apphost.generate` knows not a lot about project name validation.

I'm also fine if we want to track the removal with the linked naming issue #4400. I would lean towards just removing this -- we can catch this via code review right now.

cli/azd/internal/vsrpc/environment_service_create.go
weikanglim: The current change actually regresses functionality -- we would have replaced `.` with `-` and `TrimSuffix("my-app-host", ".AppHost")` would not trim the actual suffix.

weikanglim: My coded up suggestions here:

https://github.com/vhvb1989/azure-dev/compare/fix-projects-with-empty-spaces-aspire...weikanglim:weilim/aspire-naming?expand=1

weikanglim (approved):
LGTM

PR #4474, opened 2024-10-22
Fix autogen docs issues.

alexwolfmsft:
The docs  autogen process is causing a few issues currently. I created manual PRs into the docs to fix them, but we should try to fix them at the source. There are two issues:

- HTTP links instead of HTTPS for Awesome AZD
- The "See also" links are including unwanted description data in them, as seen below:

image

Could someone verify these changes will have the desired outcome?

Contributes to #4480


weikanglim:
@alexwolfmsft I just pushed an update. think I was able to figure out how to resolve these issues cleanly based on the initial changes you had.

alexwolfmsft:
> @alexwolfmsft I just pushed an update. think I was able to figure out how to resolve these issues cleanly based on the initial changes you had.

Thank you! Looks like it fixed the build - now we just have to confirm if these are the right changes. I saw you opened the related issue.

weikanglim (approved):

vhvb1989 (approved):

wbreza (approved):

Issue #4397, opened 2024-10-02
compose: alpha / front-end v0


wbreza:
Items:
- [x] Works with existing easy init flow (**alpha**)
  - [x] UX changes
  - [x] Host support: Container app (backend)
  - [x] Host support: Container app (frontend)
  - [x] Database support (single): MongoDB, Postgres, Redis
- [x] Allow setting app environment variables  (**alpha**)
- [x] OpenAI support (**alpha**)
  - [x] Managed identity

Additional details captured in https://github.com/Azure/azure-dev-pr/issues/1682

PR #4475, opened 2024-10-22
compose: enable infra generation

weikanglim:
Enables infra generation, i.e. temporary infrastructure and synthesized infrastructure from resources.


cli/azd/internal/scaffold/scaffold.go
```diff
@@ -115,35 +116,94 @@ func ExecInfra(
...
+// are written to the target directory.
+func ExecInfraFs(
+	t *template.Template,
+	spec InfraSpec) (*memfs.FS, error) {
+	fs := memfs.New()
+
+	// Pre-execution expansion. Additional parameters are added, derived from the initial spec.
+	preExecExpand(&spec)
+
+	err := copyFsToMemFs(resources.ScaffoldBase, fs, baseRoot, ".")
```
vhvb1989, line 147: image

ScaffoldBase is already a memory file system. Can you help me understand why do we need to copy it to a new memfs?

cli/azd/pkg/project/importer.go
```diff
@@ -191,6 +198,10 @@ func (im *ImportManager) SynthAllInfrastructure(ctx context.Context, projectConf
 		}
 	}
 
+	if im.dotNetImporter.alphaFeatureManager.IsEnabled(alpha.FeatureId(alpha.Compose)) && len(projectConfig.Resources) > 0 {
+		return infraFsForProject(ctx, projectConfig)
+	}
+
 	return nil, fmt.Errorf("this project does not contain any infrastructure to synthesize")
```
vhvb1989, line 205: If the alpha-feature is not enabled, are we printing this error message about there's not infra to synth?

cli/azd/pkg/project/resources.go
```diff
@@ -0,0 +1,147 @@
...
+	case ResourceTypeDbMongo:
+		return "MongoDB"
+	case ResourceTypeHostContainerApp:
+		return "Container App"
+	case ResourceTypeOpenAiModel:
+		return "Open AI Model"
+	}
+
+	return ""
+}
```
vhvb1989, line 37: I'd use a `Resource Interface` here and create one file for each resource type.  Then make each type to implement `String() string`.
Then I'd create a Factory to define what Resource to create based on the name coming from azure.yaml.

I see a design like that easier to grow/evolve (even driven directly by the community) versus the current model where the more resources, the longer the `switch/case`, the longer this file.

Also for expanding what a Resource can do, It's easier to ask the editor to implement a new method in the interface in for all the implementations than asking to produce a new switch case for all cases in a new Method for `Resource`

vhvb1989 (approved):
Added a few comments. Nothing really blocking

cli/azd/pkg/project/resources.go
weikanglim: I don't disagree with the observation here that an interface is more loosely coupled than a strong union-type discriminator. It offers the flexibility at runtime that can be very useful for scenarios.

The trade-off we lose is compile-time safety. I'd like us to deeply think about the design here before committing to it.

The idea that "let's just use interfaces here" and "let's start with `Resource` interface", is indeed an artifact of how OOP philosophy is deeply ingrained in our current programming paradigm.

Adopting a go philosophy, I would start with thinking about the smallest interface we should expose.

Here's how K8s separates these ideas very cleanly for inspiration:

```golang
// Core interface that all Kubernetes runtime objects must implement
type Object interface {
    GetObjectKind() schema.ObjectKind
    DeepCopyObject() Object
}

// Interface for objects that can be stored in etcd
type ObjectMeta interface {
    GetName() string
    SetName(string)
    GetNamespace() string
    SetNamespace(string)
    GetLabels() map[string]string
    SetLabels(map[string]string)
    GetAnnotations() map[string]string
    SetAnnotations(map[string]string)
    GetResourceVersion() string
    SetResourceVersion(string)
}

// Interface for type information
type ObjectKind interface {
    SetGroupVersionKind(kind schema.GroupVersionKind)
    GroupVersionKind() schema.GroupVersionKind
}

// Interface for list types
type ListInterface interface {
    GetListMeta() ListMeta
    SetListMeta(ListMeta)
}

// Example of how a concrete type implements these interfaces
type Pod struct {
    metav1.TypeMeta   `json:",inline"`
    metav1.ObjectMeta `json:"metadata,omitempty"`
    Spec   PodSpec    `json:"spec,omitempty"`
    Status PodStatus  `json:"status,omitempty"`
}

// Implementation of Object interface
func (in *Pod) DeepCopyObject() runtime.Object {
    return in.DeepCopy()
}

// Client interface for pod operations
type PodInterface interface {
    Create(ctx context.Context, pod *v1.Pod, opts metav1.CreateOptions) (*v1.Pod, error)
    Update(ctx context.Context, pod *v1.Pod, opts metav1.UpdateOptions) (*v1.Pod, error)
    Delete(ctx context.Context, name string, opts metav1.DeleteOptions) error
    Get(ctx context.Context, name string, opts metav1.GetOptions) (*v1.Pod, error)
    List(ctx context.Context, opts metav1.ListOptions) (*v1.PodList, error)
    Watch(ctx context.Context, opts metav1.ListOptions) (watch.Interface, error)
}
```

Over time, we can evolve it in a way that is more elegant (in my mind) than `Resource interface`

cli/azd/internal/scaffold/scaffold.go
weikanglim: `ScaffoldBase` is a filesystem containing assets to generate from.

Here, we are generating an in-memory filesystem contains the actual `infra` Bicep files that would be used for `azd provision`. This is how Aspire works today as well.

cli/azd/pkg/project/importer.go
weikanglim: We can definitely consider a better error message for enablement.

cli/azd/internal/scaffold/scaffold.go
vhvb1989: I'm confused about the copy from fs to a new fs.
We create the template directly from the fs with `ParseFS()` https://github.com/Azure/azure-dev/blob/main/cli/azd/pkg/apphost/generate.go#L99

Then executeToFS() uses an existing fs.   Is this `existing fs` the equivalent to the new fs you are creating?

That's why I wonder if you can just re-use the same fs, maybe just to use a root folder like `out` ?

cli/azd/pkg/project/scaffold_gen.go
```diff
@@ -0,0 +1,241 @@
...
+		if isSecret {
+			// TODO: handle secrets
+			continue
+		}
+
+		svcSpec.Env[envVar.Name] = value
+	}
+
+	port := props.Port
+	if port < 1 || port > 65535 {
```
wbreza, line 203: nit: consider moving 65535 out to a constant.

wbreza (approved):

cli/azd/pkg/project/scaffold_gen.go
weikanglim: I will leave it as-is for now -- we could think about creating some sort of a networking package that validates port numbers amongst other things. But creating a constant right now would end up adding to global variables in `project` that wouldn't quite be consumable either, so I'm pretty happy with how it looks right now.

PR #4477, opened 2024-10-22
compose: add experimental envsubt 

weikanglim:
Add primitive support for envsubst expressions configuration.

### DIFF
--- a/cli/azd/pkg/project/scaffold_gen.go
+++ b/cli/azd/pkg/project/scaffold_gen.go
@@ -154,7 +154,7 @@ func infraSpec(projectConfig *ProjectConfig) (*scaffold.InfraSpec, error) {
 				Port: -1,
 			}
 
-			err := mapContainerApp(res, &svcSpec)
+			err := mapContainerApp(res, &svcSpec, &infraSpec)
 			if err != nil {
 				return nil, err
 			}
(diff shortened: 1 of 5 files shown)


cli/azd/pkg/project/scaffold_gen_util.go
```diff
@@ -0,0 +1,59 @@
...
+	stop  int
+}
+
+// parseEnvSubstVariables parses the envsubst expression(s) present in a string.
+// substitutions, returning the locations of the expressions and the names of the variables.
+//
+// It works with both:
+//   - ${var} and
+//   - ${var:=default} syntaxes
+func parseEnvSubstVariables(s string) (names []string, expressions []location) {
```
ellismg, line 22: We may at some point want a way to escape `${` in these strings.

cli/azd/pkg/project/scaffold_gen.go
```diff
@@ -186,17 +186,36 @@ func infraSpec(projectConfig *ProjectConfig) (*scaffold.InfraSpec, error) {
...
 		isSecret := len(envVar.Secret) > 0
 		value := envVar.Value
 		if isSecret {
-			// TODO: handle secrets
-			continue
+			value = envVar.Secret
 		}
 
-		svcSpec.Env[envVar.Name] = value
+		// Notice that we derive isSecret from it's usage.
```
ellismg, line 212:
```suggestion
		// Notice that we derive isSecret from its usage.
```

cli/azd/pkg/project/scaffold_gen.go
```diff
@@ -186,17 +186,36 @@ func infraSpec(projectConfig *ProjectConfig) (*scaffold.InfraSpec, error) {
...
+		// This is generally correct, except for the case where:
+		// - CONNECTION_STRING: ${DB_HOST}:${DB_SECRET}
+		// Here, DB_HOST is not a secret, but DB_SECRET is. And yet, DB_HOST will be marked as a secrete.
```
ellismg, line 215:
```suggestion
		// Here, DB_HOST is not a secret, but DB_SECRET is. And yet, DB_HOST will be marked as a secret.
```

cli/azd/pkg/project/scaffold_gen.go
```diff
@@ -239,3 +258,70 @@ func mapHostUses(
...
+// evalEnvSubst evaluates a value that may contain multiple envsubst expressions, returning the evaluated value.
+//
+// The value may contain multiple expressions, e.g.:
+//   - "Hello, ${world:=okay}!"
+//   - "${CAT} and ${DOG}"
+//   - "${DB_HOST:='local'}:${DB_USERNAME:='okay'}"
+//
+// Each expression is evaluated and replaced with a reference to a parameter in the infra spec.
+// generateAsSecret is used to determine if the parameter generated should be marked as a secret.
+func evalEnvSubst(value string, generateAsSecret bool, infraSpec *scaffold.InfraSpec) string {
```
ellismg, line 298: FWIW - It took me a while to fully understand the semantics on how the output string is formed here.  From looking at the implementation sometimes the value is surrounded by single quotes, sometimes it isn't and sometimes the `${}`'s end up sticking around.

I think that this is done because we are trying to generate text which is a bicep string value, and this method is trying to generate an idomatic bicep string value (which explains why in the single reference case we just emit a variable reference (e.g. `foo`) instead of a something like `'${foo}'`, since the linters yell at you for that.

Maybe adding something in the name of the function or the comment about how it is trying to write the value as an idomatic bicep string literal?  That might also make it clearer was to why we call `scaffold.BicepName` on the value.

For a long while I was just thinking this was actually going to evaluate the string to a ground value instead of rewriting it as an interperolated string in a different format.

ellismg (approved):
Some small comments around naming, it took be a bit to understand exactly what was going on here.

cli/azd/pkg/project/scaffold_gen_util.go
weikanglim: Indeed. I suspect that we'd perhaps also move away from `${VARIABLE}` into something more powerful for different kinds of use-cases.

For alpha, the current thinking is to expose the basic functionality enough for users to find the value in it.

cli/azd/pkg/project/scaffold_gen.go
weikanglim: Agreed 💯 on the confusing names and documents here.

I updated the name to `genBicepParamsFromEnvSubst` to better reflect what it's doing. Updated the comments to describe the Bicep outputs.

Issue #4489, opened 2024-10-25
Customizing ContainerApp name with dotnet aspire and alpha.aspire.useBicepForContainerApps


vhvb1989:
from: https://github.com/Azure/azure-dev/discussions/4488

vhvb1989:
closing as not to be fixed based on: https://github.com/Azure/azure-dev/pull/4490#issuecomment-2475039890

PR #4490, opened 2024-10-25
Fix finding the ACA from Aspire9 deployment with custom ACA name

vhvb1989:
This PR makes azd to get the ACA name when setting a custom name in the AppHost.

The name is required only to display the endpoint.

fix: https://github.com/Azure/azure-dev/issues/4489

### DIFF
--- a/cli/azd/pkg/project/service_target_dotnet_containerapp.go
+++ b/cli/azd/pkg/project/service_target_dotnet_containerapp.go
@@ -300,6 +300,9 @@ func (at *dotnetContainerAppTarget) Deploy(
 		return nil, fmt.Errorf("failed executing template file: %w", err)
 	}
 
+	// default name for the container app is the service name. This is how this work for using YAML definition where
+	// azd owns the ACA definition. When using bicep, the name is defined in the bicep file and is set after deployment.
+	acaName := serviceConfig.Name
 	if useBicepForContainerApps {
 		// Compile the bicep template
 		compiled, params, err := func() (azure.RawArmTemplate, azure.ArmParameters, error) {
@@ -382,7 +385,7 @@ func (at *dotnetContainerAppTarget) Deploy(
 		armTemplate = &compiled
 		armParams = params
 
-		_, err = at.deploymentService.DeployToResourceGroup(
+		deploymentResult, err := at.deploymentService.DeployToResourceGroup(
 			ctx,
 			at.env.GetSubscriptionId(),
 			targetResource.ResourceGroupName(),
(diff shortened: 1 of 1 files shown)


cli/azd/pkg/project/service_target_dotnet_containerapp.go
```diff
@@ -393,6 +396,26 @@ func (at *dotnetContainerAppTarget) Deploy(
...
+		if len(deploymentResult.Resources) == 0 {
+			return nil, fmt.Errorf("no resources found after deployment. Expected at least one Azure Container App")
+		}
+		acaCount := 0
+		for _, res := range deploymentResult.Resources {
+			if res.ID == nil {
+				continue
+			}
+			resourceId := *res.ID
+			if strings.Contains(resourceId, "providers/Microsoft.App/containerApps") {
```
weikanglim, line 409: I'm wondering if this ends up matching a nested resource of `providers/Microsoft.App/containerApps/<name>/<child resource>`?

Using `arm.ParseResourceID` could help here. From the resulting `resId`, you could check `resId.ResourceType.Type ==  Microsoft.App/containerApps &&  resId.ResourceType.Types == nil`.

cli/azd/pkg/project/service_target_dotnet_containerapp.go
```diff
@@ -393,6 +396,26 @@ func (at *dotnetContainerAppTarget) Deploy(
 		if err != nil {
 			return nil, fmt.Errorf("deploying bicep template: %w", err)
 		}
+		// set the name of the container app to the name of the deployment
+		if len(deploymentResult.Resources) == 0 {
```
weikanglim, line 400: I think this is probably okay for now. One thing to note is that this doesn't handle nested Bicep modules / ARM deployments IIRC.

weikanglim: Another alternative if we want to make it similar to what `azd` is doing elsewhere today:

Honor the conventional `AZURE_CONTAINER_APP_ID` output variable -- this would solve it recursively, but will require user manual input.

vhvb1989: The definition for the ACA that Aspire provides is unlikely to bring more than one resource.  It currently uses a resource-group deployment and just defines the ACA creation.

I wanted to add just a little heads-up here in case this ever changes and/or there's a bug in Aspire.

Now, folks can still run azd infra synth and take control over the bicep that creates the ACA to bring more service's infra here, but that's a super edge case right now. So we don't care about this for now. :)

cli/azd/pkg/project/service_target_dotnet_containerapp.go
vhvb1989: It's fine to start here with a simple validation.  Most likely we will not need more than that.

vhvb1989:
> It feels like this is handling a case where the physical resource name differs from the logical resource name in the apphost model (since that is what `serviceConfig.Name` is built from).
> 
> In that case, would we also need to use the updated physical resource name when using it in a binding expression like `.url`?
> 
> I don't doubt that this fixes the issue, I'm just wondering if there are more dragons lurking here...

I think that's a conversation about that here: https://github.com/Azure/azure-dev/discussions/4488#discussioncomment-11064544

What this PR is doing is not to fail when trying to fetch the name of the just-deployed ACA, which seems like a bug.

Maybe the AppHost model should block setting the ACA name field or fail when you change it. But that's on the Aspire side @mitchdenny @davidfowl

ellismg:
I don't think we should do this - in practice changing the name of the container app is something we don't support today and even with this change we still can't feel good about it, since things like the `.url` property which could be used in a binding expression don't take this into account.

I think if this is something we want to support - we need some better way (maybe in the manifest) to communicate what the physical name of the container app is - so `azd` can flow it as needed.

I think if we want to support this we need to design the full end to end feature first - doing this only helps somewhat.

PR #3976, opened 2024-06-04
Update the infra/core modules to AVM modules

Menghua1:
Fixes: https://github.com/Azure/azure-dev/issues/3327

@jongio for notification.

PR #4491, opened 2024-10-25
Update appservice.bicep with App Insights env var

pamelafox:
Per https://learn.microsoft.com/en-us/azure/azure-monitor/app/codeless-app-service?tabs=python#automate-monitoring , an additional environment variable is required for application monitoring in App Service.

I'm adding the same var to azure-search-openai-demo in this PR:
https://github.com/Azure-Samples/azure-search-openai-demo/pull/2083/files
And testing the deploy now.

### DIFF
--- a/templates/common/infra/bicep/core/host/appservice.bicep
+++ b/templates/common/infra/bicep/core/host/appservice.bicep
@@ -92,6 +92,7 @@ module configAppSettings 'appservice-appsettings.bicep' = {
       {
         SCM_DO_BUILD_DURING_DEPLOYMENT: string(scmDoBuildDuringDeployment)
         ENABLE_ORYX_BUILD: string(enableOryxBuild)
+        ApplicationInsightsAgent_EXTENSION_VERSION: contains(kind, 'linux') ? '~3' : '~2'
       },
       runtimeName == 'python' && appCommandLine == '' ? { PYTHON_ENABLE_GUNICORN_MULTIWORKERS: 'true'} : {},
       !empty(applicationInsightsName) ? { APPLICATIONINSIGHTS_CONNECTION_STRING: applicationInsights.properties.ConnectionString } : {},


vhvb1989:
Thanks @pamelafox 
I think we are not taking any more changes to bicep core modules as we are almost there with AVMs here: https://github.com/Azure/azure-dev/pull/3976

common/infra/bicep/app/web-appservice.bicep is becoming: /web-appservice-avm.bicep here: https://github.com/Azure/azure-dev/pull/3976/files#diff-c1aa701cf6c1c9590e75f3b45869a2b31013a8435ceac00fe13da4778bc37bb8R1

pamelafox:
Okay I'll comment on that PR. I did a quick look through avm source code for the string and didnt see it, but maybe the modules are in multiple repos.

templates/common/infra/bicep/core/host/appservice.bicep
```diff
@@ -92,6 +92,7 @@ module configAppSettings 'appservice-appsettings.bicep' = {
       {
         SCM_DO_BUILD_DURING_DEPLOYMENT: string(scmDoBuildDuringDeployment)
         ENABLE_ORYX_BUILD: string(enableOryxBuild)
+        ApplicationInsightsAgent_EXTENSION_VERSION: contains(kind, 'linux') ? '~3' : '~2'
```
jongio, line 95: Seems like this should be screaming caps to match env var naming convention.

pamelafox: That's not what their docs show:
https://learn.microsoft.com/en-us/azure/azure-monitor/app/codeless-app-service?tabs=python#automate-monitoring

pamelafox:
FYI, I haven't gotten this to successfully deploy yet, and apparently its not 3.12 compatible yet, only 3.11, so maybe its not ready for primetime. I'm working with the eng team.

pamelafox:
Update: I discussed this with the App Insights team, and they said that the variable is only relevant if you're using auto-instrumentation, as it will turn on the auto-instrumentation. Many of our azd templates are using manual instrumentation, so it's not relevant (and in fact, does not work as it produces a conflict). I've asked them to perhaps add a note in that "App Insights" blade that App Insights might still be enabled, even if it hasn't been auto-enabled.
Closing this PR

PR #4500, opened 2024-10-29
compose: refactor to add alpha switch

weikanglim:
A small refactor to integrate the alpha switch change.

In the easy init flow, we would want `azd init` without composability alpha feature enabled to continue materializing infrastructure as-is. If composability is enabled, we would want to hold the infrastructure in-memory.


ellismg (approved):
LGTM

PR #4508, opened 2024-10-31
chore: refactor port prompt

weikanglim:
Refactoring port prompting. These changes are being teased out to reshare implementation across composability-enabled and regular pathways.


vhvb1989 (commented):
I'll try to improve the input.Console and the input.ConsoleOptions instead of building more prompting functionality on top of it. 
The input.Console already handles things like `isPassword` , allowing all components on top the input package to use it.

ellismg (approved):
LGTM

PR #1715, opened 2023-03-11
Adding `Alpha Features`

vhvb1989:
Adding `feature design` for @ellismg approval :) and for team's comments.
I will start working on the implementation-design parallelly, while the functional spec is acknowledge.

fix: https://github.com/Azure/azure-dev/issues/1633

PR #4509, opened 2024-10-31
easy-init: generate resources in `azure.yaml`

weikanglim:
Generate resources in `azure.yaml` and delay infrastructure generation if `alpha.compose` is enabled

### DIFF
--- a/cli/azd/internal/repository/app_init.go
+++ b/cli/azd/internal/repository/app_init.go
@@ -3,8 +3,11 @@ package repository
 import (
 	"context"
 	"fmt"
+	"maps"
 	"os"
 	"path/filepath"
+	"slices"
+	"strings"
 	"time"
 
 	"github.com/azure/azure-dev/cli/azd/internal"
@@ -38,6 +41,8 @@ var dbMap = map[appdetect.DatabaseDep]struct{}{
 	appdetect.DbPostgres: {},
 	appdetect.DbRedis:    {},
 }
+
+var featureCompose = alpha.MustFeatureKey("compose")
 
 // InitFromApp initializes the infra directory and project file from the current existing app.
 func (i *Initializer) InitFromApp(
@@ -244,30 +249,31 @@ func (i *Initializer) InitFromApp(
 
 	// Create the infra spec
 	var infraSpec *scaffold.InfraSpec
-	if !i.features.IsEnabled(alpha.Compose) { // backwards compatibility
+	composeEnabled := i.features.IsEnabled(featureCompose)
+	if !composeEnabled { // backwards compatibility
 		spec, err := i.infraSpecFromDetect(ctx, detect)
 		if err != nil {
 			return err
 		}
 		infraSpec = &spec
-	}
 
-	// Prompt for environment before proceeding with generation
-	_, err = initializeEnv()
-	if err != nil {
-		return err
+		// Prompt for environment before proceeding with generation
+		_, err = initializeEnv()
+		if err != nil {
+			return err
+		}
 	}
 
 	tracing.SetUsageAttributes(fields.AppInitLastStep.String("generate"))
 
-	i.console.Message(ctx, "\n"+output.WithBold("Generating files to run your app on Azure:")+"\n")
 	title = "Generating " + output.WithHighLightFormat("./"+azdcontext.ProjectFileName)
 	i.console.ShowSpinner(ctx, title, input.Step)
-	err = i.genProjectFile(ctx, azdCtx, detect)
+	err = i.genProjectFile(ctx, azdCtx, detect, composeEnabled)
 	if err != nil {
 		i.console.StopSpinner(ctx, title, input.GetStepResultFormat(err))
 		return err
 	}
+	i.console.Message(ctx, "\n"+output.WithBold("Generating files to run your app on Azure:")+"\n")
 	i.console.StopSpinner(ctx, title, input.StepDone)
 
 	if infraSpec != nil {
@@ -279,6 +285,20 @@ func (i *Initializer) InitFromApp(
 			return err
 		}
 		i.console.StopSpinner(ctx, title, input.StepDone)
+	} else {
+		t, err := scaffold.Load()
+		if err != nil {
+			return fmt.Errorf("loading scaffold templates: %w", err)
+		}
+
+		err = scaffold.Execute(t, "next-steps-alpha.md", nil, filepath.Join(azdCtx.ProjectDirectory(), "next-steps.md"))
+		if err != nil {
+			return err
+		}
+
+		i.console.MessageUxItem(ctx, &ux.DoneMessage{
+			Message: "Generating " + output.WithHighLightFormat("./next-steps.md"),
+		})
 	}
 
 	return nil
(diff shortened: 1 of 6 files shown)


cli/azd/internal/repository/infra_confirm.go
```diff
@@ -33,7 +33,7 @@ func (i *Initializer) infraSpecFromDetect(
 
 	dbPrompt:
 		for {
-			dbName, err := i.promptDbName(ctx, database)
+			dbName, err := promptDbName(i.console, ctx, database)
```
weikanglim, line 36: This change builds on top of #4508, `infra_confirm.go` is unchanged between these two PRs, so the three files that need reviewed are:

- app_init.go
- app_init_test.go
- next-steps-alpha.mdt

cli/azd/internal/repository/app_init.go
```diff
@@ -244,30 +244,31 @@ func (i *Initializer) InitFromApp(
 
 	// Create the infra spec
 	var infraSpec *scaffold.InfraSpec
-	if !i.features.IsEnabled(alpha.Compose) { // backwards compatibility
+	composeEnabled := i.features.IsEnabled(alpha.Compose)
```
vhvb1989, line 247: Can you move the `alpha.Compose` FeatureId to the package that exposes the feature?
The alpha package should not be aware of what alpha features exists in the product, it is a package that provides the functionality to define and use the features, but features should be defined where they belong.

cli/azd/internal/repository/app_init.go
```diff
@@ -429,7 +449,103 @@ func prjConfigFromDetect(
...
+				}
+
+				if dbName == "" {
+					i.console.Message(ctx, "Database name is required.")
+					continue
+				}
+
+				db.Name = dbName
+				break
+			}
```
vhvb1989, line 498: please move this to a new method to help with indentation and readability.  Since the context is already a for loop, it is easy to confuse a reader about the existing of a new `for`, taking extra time to recognize the effects of `continue` and `break`.

cli/azd/internal/repository/app_init.go
```diff
@@ -429,7 +449,103 @@ func prjConfigFromDetect(
...
+			for {
+				dbName, err := promptDbName(i.console, ctx, database)
+				if err != nil {
+					return config, err
+				}
+
+				if dbName == "" {
+					i.console.Message(ctx, "Database name is required.")
+					continue
+				}
```
vhvb1989, line 494: nit. Consider a `Cancel` option here.  Or if `promptDbName` already supports user-cancellation with an error, please add a note about cancelling is handled with an error.

cli/azd/internal/repository/app_init.go
```diff
@@ -359,16 +375,20 @@ func (i *Initializer) genProjectFile(
 
 const InitGenTemplateId = "azd-init"
 
-func prjConfigFromDetect(
+func (i *Initializer) prjConfigFromDetect(
```
vhvb1989, line 378: nit. I'm not a fan of the super long methods.  Is this almost 200?

cli/azd/internal/repository/infra_confirm.go
```diff
@@ -167,10 +131,10 @@ func (i *Initializer) infraSpecFromDetect(
 	return spec, nil
 }
 
-func (i *Initializer) getPortByPrompt(ctx context.Context, promptMessage string) (int, error) {
+func promptPortNumber(console input.Console, ctx context.Context, promptMessage string) (int, error) {
```
vhvb1989, line 134: consider moving this functionality to the input.Console by using the `input.ConsoleOptions` where you could define the type Number and the range.  This would benefit the entire product and not only things around composibility

cli/azd/internal/repository/infra_confirm.go
```diff
@@ -207,10 +171,10 @@ func (i *Initializer) promptDbName(ctx context.Context, database appdetect.Datab
 		}
 
 		if strings.ContainsAny(dbName, " ") {
-			i.console.MessageUxItem(ctx, &ux.WarningMessage{
+			console.MessageUxItem(ctx, &ux.WarningMessage{
 				Description: "Database name contains whitespace. This might not be allowed by the database server.",
 			})
-			confirm, err := i.console.Confirm(ctx, input.ConsoleOptions{
+			confirm, err := console.Confirm(ctx, input.ConsoleOptions{
 				Message: fmt.Sprintf("Continue with name '%s'?", dbName),
```
vhvb1989, line 178: Looks like the `input-text + warning + confirm` pattern is repeated multiple times around here. Consider making it on general function or even better, supported by input.Console.

vhvb1989 (changes requested):
blocking due to the infra feature definition within the infra package.
Think about this like defining strings within the strings package.

cli/azd/internal/repository/app_init.go
weikanglim: Just to clarify, are we changing directions from our previous efforts to add alpha feature string constants into `alpha_features.go`, as in #1715, where we added [terraform](https://github.com/azure/azure-dev/blob/main/cli/azd/pkg/alpha/alpha_features.go), which it still remains to date?

Keep in mind that this is indeed a string constant, it is not any meaning functionality that would violate any directional dependencies.

cli/azd/internal/repository/infra_confirm.go
weikanglim: I could be wrong, but I think you're asking for features of a UI dialog prompt library, such as the default `survey.Questions`: https://pkg.go.dev/github.com/AlecAivazis/survey/v2#section-readme.

`azd` had decidedly avoid this multi-prompt dialog pattern to gain control over each interaction. I do somewhat see over time, as we're building more complex prompts, the need for this. But this isn't a suggestion that we can land within a feature change itself; let's take a holistic consideration on this over time.

cli/azd/internal/repository/infra_confirm.go
weikanglim: See the answer [here](https://github.com/Azure/azure-dev/pull/4509#discussion_r1828181015)

cli/azd/internal/repository/app_init.go
weikanglim: Let's also think about the cost of separation tradeoff here: would a bug be as easily caught in review when the function is broken into 5 separate functions, each with their own input/output contracts, and data is flowing through them?

cli/azd/internal/repository/app_init.go
vhvb1989: the terraform constant is dead code (not used anywhere). We should just remove it. 
alpha feature constants should not be defined within the alpha package.

cli/azd/internal/repository/app_init.go
weikanglim: It is written this way so that we can more easily merge the alpha vs. non-alpha code cleanly, once we remove the backcompat `alpha` feature. The code is slightly ugly here so that we can move quicker and cleaner in the future -- I hope that is acceptable in your mind.

vhvb1989: Help me understand how's the alpha switch related to having a for loop inside another for loop and having `break` and `continue` for both loops

weikanglim: Certainly.

The logic 

```golang
for {
				dbName, err := promptDbName(i.console, ctx, database)
				if err != nil {
					return config, err
				}

				if dbName == "" {
					i.console.Message(ctx, "Database name is required.")
					continue
				}

				db.Name = dbName
				break
			}
```

would look like

```golang
dbName, err := promptDbName(i.console, ctx, database)
if err != nil {
   return config, err
}
```

in the future, once we merge the name validation into the `promptDbName` prompt directly. The reason why it doesn't work today, is that easy init requires the `promptDbName` functionality to not validate for an empty name.

cli/azd/internal/repository/app_init.go
weikanglim: User cancellation is supported via Ctrl+C as the standard prompt interfacing; I don't think we will want to introduce additional cancellation options here today.

vhvb1989 (approved):

PR #4511, opened 2024-11-01
compose: generate openai in the backend

weikanglim:
Add backend infra generation for OpenAI models.

### DIFF
--- a/cli/azd/pkg/project/scaffold_gen.go
+++ b/cli/azd/pkg/project/scaffold_gen.go
@@ -165,6 +165,23 @@ func infraSpec(projectConfig *ProjectConfig) (*scaffold.InfraSpec, error) {
 			}
 
 			infraSpec.Services = append(infraSpec.Services, svcSpec)
+		case ResourceTypeOpenAiModel:
+			props := res.Props.(AIModelProps)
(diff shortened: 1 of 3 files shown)


ellismg (approved):

cli/azd/pkg/project/scaffold_gen.go
```diff
@@ -165,6 +165,23 @@ func infraSpec(projectConfig *ProjectConfig) (*scaffold.InfraSpec, error) {
 			}
 
 			infraSpec.Services = append(infraSpec.Services, svcSpec)
+		case ResourceTypeOpenAiModel:
```
vhvb1989, line 168: Feels like a deja-vu to Aspire origins.  There should be learnings about why the strategy of manually adding supported resources was not solid and made us pivot from it.

vhvb1989 (approved):
Leaving a quick note about the current strategy.
This feels like not going towards a scalable solution.  otherwise LGTM

cli/azd/pkg/project/scaffold_gen.go
weikanglim: We will be rewriting the backend -- it will scale; just not for alpha  👌

PR #4515, opened 2024-11-04
dotnet: Set `DOTNET_NOLOGO` when running external commands

### LINKED ISSUE #4513: [Aspire] First time running `dotnet msbuild --getProperty:IsAspireHost` fails due to telemetry notice
Usually, when running `dotnet msbuild --getProperty:IsAspireHost` , the response is either `false` or `true`. Azd uses this output to detect an Aspire project.

However, on a very fresh dotnet installation, the first response from `dotnet msbuild --getProperty:IsAspireHost` includes a telemetry notice like:

```
PS C:\Users\vivazqu\t2\eShop_dyigitbas\src\eShop.AppHost> dotnet msbuild --getProperty:IsAspireHost

This is .NET 9.0.
---------------------
SDK Version: 9.0.100-rc.2.24474.11

Telemetry
---------
The .NET tools collect usage data to help us improve your experience. Microsoft collects this data and shares it with the community. You can opt out of telemetry by setting the DOTNET_CLI_TELEMETRY_OPTOUT environment variable to "1" or "true" using your favorite shell.

Read more about .NET CLI tools telemetry: https://aka.ms/dotnet-cli-telemetry

----------------
Install an ASP.NET Core HTTPS development certificate.
To trust the certificate, run "dotnet dev-certs https --trust"
Learn about HTTPS: https://aka.ms/dotnet-https

----------------
Write your first application: https://aka.ms/dotnet-hello-world
Discover what's new: https://aka.ms/dotnet-whats-new
Explore the documentation: https://aka.ms/dotnet-docs
Report issues and find the source code on GitHub: https://github.com/dotnet/core
Use "dotnet --help" to see available commands or visit: https://aka.ms/dotnet-cli
--------------------------------------------------------------------------------------
true
```

This output makes `azd` to think the project is not Aspire.

I was able to reproduce by installing dotnet 9 on a fresh new Windows instance and running `dotnet msbuild --getProperty:IsAspireHost` inside an AppHost project.  Here is some pictures (the VM was in Spanish, but it should help to illustrate the issue)

Image

// Note how, from the second time ahead, the output is the expected.

ellismg:
Per https://learn.microsoft.com/dotnet/core/tools/dotnet-environment-variables#dotnet_nologo:

> Specifies whether .NET welcome and telemetry messages are displayed
  on the first run. Set to true to mute these messages (values `true`,
  `1`, or `yes` accepted) or set to false to allow them (values
  `false`, `0`, or `no` accepted). If not set, the default is `false`
  and the messages will be displayed on the first run.

Since there are cases where we interpret the output of the `dotnet` CLI (for example, when calling `dotnet msbuild
--getProperty:IsAspireHost`) we should set this value to ensure that this message is not printed, which would break our deserialization logic, which expects just JSON data to be printed to stdout.

Fixes #4513

### DIFF
--- a/cli/azd/.vscode/cspell-azd-dictionary.txt
+++ b/cli/azd/.vscode/cspell-azd-dictionary.txt
@@ -144,6 +144,7 @@ mysqldb
 nobanner
 nodeapp
 nolint
+nologo
 notrail
 omitempty
 oneauth
--- a/cli/azd/pkg/tools/dotnet/dotnet.go
+++ b/cli/azd/pkg/tools/dotnet/dotnet.go
@@ -316,6 +316,7 @@ func newDotNetRunArgs(args ...string) exec.RunArgs {
 
 	runArgs = runArgs.WithEnv([]string{
 		"DOTNET_CLI_WORKLOAD_UPDATE_NOTIFY_DISABLE=1",
+		"DOTNET_NOLOGO=1",
 	})
 
 	return runArgs


ellismg:
Wrote this based on your excellent writeup in #4513, @vhvb1989.  I'll get a fresh dev-box and exercise the scenario we suspected here.

vhvb1989:
I think what we need is `DOTNET_SKIP_FIRST_TIME_EXPERIENCE` ( see: https://learn.microsoft.com/dotnet/core/tools/dotnet-environment-variables#dotnet_skip_first_time_experience).

vhvb1989 (approved):

ellismg:
>  I'll get a fresh dev-box and exercise the scenario we suspected here.

Yup - I was able to reproduce this by setting up fresh dev boxes, installing RC2 of .NET 9.0 and then running `azd init --from-code` in a fresh clone of eShop.  With this change, the app is now correctly detected as an Aspire project.

weikanglim (approved):

Issue #4501, opened 2024-10-29
[EngSys] Remove usage of azd-login.yml


danieljurek:
Environment/auth values set in azd-login.yml should either be redundant or set in the AzureCli@2 task and mapped appropriately from there. The ultimate goal is to eliminate usage of the subscription config secret in azd-login.yml

danieljurek:
This was fixed by Matt's PR https://github.com/Azure/azure-dev/pull/4519/files#diff-db841e68541ac5106396b8d0b77ea8d1d37dbe82cf355ec089136d65140f1b62

PR #4516, opened 2024-11-04
Migrate live testing to TME

danieljurek:
Partial #4501

Migrate testing to using TME, uses azd auth, flows auth information via environment variables. Still uses legacy az CLI auth. 

When Matt's change (https://github.com/Azure/azure-dev/pull/4343) merges this can change to incorporate the new federated auth features.


danieljurek:
Very strange. There appear to be some testing failures which we didn't have last week. I'll investigate and then take this PR out of draft.

eng/pipelines/templates/steps/template-test-run-job.yml
```diff
@@ -12,8 +12,6 @@ steps:
       parameters:
           Version: $(AzdVersion)
 
-    - template: /eng/pipelines/templates/steps/azd-login.yml
```
ellismg, line 15: I have removed this as well in my branch where I am getting this job working on TME with OIDC.

ellismg:
This was fixed in #4519

Issue #3699, opened 2024-04-12
[Feature request] Secure secrets retrieval in .env file


sinedied:
- [x] Make sure you've installed the latest version using [instructions in the wiki](../wiki/install)

**Output from `azd version`**
azd version 1.8.0 (commit 8246323c2472148288be4b3cbc3c424bd046b985)

**Describe the issue**

`.env` files are often used during development to test/debug an app locally, while connecting to cloud resources. AZD generates such `.env` files in `.azure` folder, and allows to retrieve these values with `azd env get-values`.

The problem is that currently the only way to retrieve secrets such as connection strings or API keys in that `.env` file is to expose them as output in the Bicep templates, an unsecure practice as these secrets then gets exposed in Azure deployments.

A workaround to that issue is to use Key Vault to store the secrets, but in that case to get these secrets added into the `.env` for local testing, it complicates things a lot for the users:

- You have to install AZ CLI in addition to AZD (~1 Go)
- You have to log in to AZ CLI in addition to AZD
- You have to create a script that list the the secrets in KV, retrieve them 1 by 1 and add them to `.env` file. And double that if you need to support both Mac/Linux and Windows

Given that AZD aims to be the main CLI tool for developers, it would make sense to integrate a feature to automatically retrieve secrets from a KV and include them in the generated `.env` file.

> **Note**: I'm aware that Managed Identity is the best way to avoid handling secrets, but not all services and tools supports them. For example, Azure CosmosDB for MongoDB vCore only supports connection strings, and Azure Database Vs Code extension can only connect to DB using a connection strings.

weikanglim:
Inspirationally, I would love if we could tell a story like this for users that are used to environment variables:

```bash
export OPENAI_KEY='<secret>' 
azd provision
```

Or:

```bash
# this works just like before,
# the variable is just stored securely for repeatability and automated CI setup 
azd env set --secret OPENAI_KEY='<secret>'
azd provision
```

PR #4517, opened 2024-11-06
azd env set-secret

vhvb1989:
Adding `azd env set-secret`

### Description:

With this change, AZD allows customers to set Azure Key Vault Secrets (*Akvs*) within the azd environment. 

### Adding a Akvs

Folks run `azd env set-secret <name>` and they can select to either "create a new Key Vault secret" or "select an existing one".  For the `Create a new Akvs`, folks can select an existing Azure Key Vault Account or Create a new one. See the next table for the options for a user:

| Flow | Description |
| -------- | -------- |
| 1) Create akvs && Account && RG | Creating a new Resource Group with one Azure Key Vault Account and one akvs in it. User must select a Subscription, location and provide a name for the rg and kv account |
| 2) Create akvs && Account && Existing RG | Creating Azure Key Vault Account and one akvs in it on existing Resource Group. User must select a Subscription, location and provide a name for the kv account |
| 3) Create akvs && Existing Account | Creating one akvs in existing Key Vault Account. User must select a Subscription where the Key Vault Account is |
| 4) Existing akvs | Selecting existing akvs in existing Key Vault Account. User must select a Subscription where the Key Vault Account is. *If no key vault accounts are found in the subscription, the `create account` option is turned ON  |

At the end of the flow, the akvs is persisted within the .env file (the azd environment) using the `<name>` provided by the user for the secret and the reference to the akvs as the value, which looks like:

```
NAME="akvs://key-vault-account-name/akvs-name"
```

Azd uses the prefix `akvs://` as a way to identify a reference to an Azure key Vault Secret.

### Using secrets

AZD can get the akvs value during the next flows:

1. `azd provision + bicep + main.parameters.json`. 
      -  Adding input parameter like:  

```bicep
@secure
input parameterName type
```
 
> The type can be any supported type. The main requirement is to make the parameter `secure`.
> AZD will not get akvs values for non-secure parameters.

  -  Create a mapping in `main.parameters.json` for the parameter to the secret from the .env. For example, when calling `azd env set-secret OPEN_ID_KEY` the mapping would look like:

```json
"parameterName": {
  "value": "${OPEN_ID_KEY}"
},
```

  - When AZD notices the parameter is secured, it will check if the value in the .env is an akvs reference and if yes, it will try to pull the value.

2. `azd provision + bicep + .bicepparams`.  ->  WIP

3.  azd hooks:
    - When azd runs a hook, it uses its .env for the hook invocation. The hook implementation can access the values from azd .env as environment variables. However, if there are akvs in the .env, the values are NOT automatically converted to its value.
    - If folks want to get akvs values, they need to create a mapping in the hook definition, like:

```yaml
hooks:
  preprovision:
    posix:
      shell: sh
      run: ./infra/azd/hooks/postprovision.sh
      secrets:
        OPEN_ID_KEY_VALUE: OPEN_ID_KEY
```

    - On this example, azd will add one addional env var `OPEN_ID_KEY_VALUE` after pulling the value from the akvs reference.

Any other interaction with Azd environment will not resolve akvs into values. For example, running `azd env get-values` will display the akvs reference.   

fix: https://github.com/Azure/azure-dev/issues/3699
fix:


wbreza (commented):
This is looking good but just have some questions and concerns about when and where the secret reference to secret value translation occurs.

cli/azd/cmd/env.go
```diff
@@ -150,6 +163,265 @@ func (e *envSetAction) Run(ctx context.Context) (*actions.ActionResult, error) {
...
+			e.console.ShowSpinner(ctx, "Adding Administrator Role", input.Step)
+			principalId, err := azureutil.GetCurrentPrincipalId(ctx, e.userProfileService, tenantId)
+			e.console.StopSpinner(ctx, "", input.Step)
+			if err != nil {
+				return nil, fmt.Errorf("getting current principal ID: %w", err)
+			}
+			err = e.entraIdService.CreateRbac(ctx, subId, kvAccount.Id, keyvault.RoleIdKeyVaultAdministrator, principalId)
+			if err != nil {
+				return nil, fmt.Errorf("creating Key Vault RBAC: %w", err)
+			}
```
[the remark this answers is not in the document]
vhvb1989: The .env file is not committed to the repository.
The strategy for CI will be covered in a separate issue (https://github.com/Azure/azure-dev/issues/4699)

cli/azd/pkg/ext/hooks_runner.go
```diff
@@ -134,7 +139,31 @@ func (h *HooksRunner) execHook(ctx context.Context, hookConfig *HookConfig, opti
...
+						if err != nil {
+							return err
+						}
+						valueFromEnv = secretValue
+					}
+					setValue = valueFromEnv
+				}
+				hookEnv.DotenvSet(key, setValue)
+			}
+			return nil
```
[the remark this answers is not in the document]
vhvb1989: Right now there are only 2 components who care about doing `secret ref => secret value` 
- hooks
- bicep provider

each component has a unique way to enable and use the feature, so the component is handling it and making the azd internals (environment manager) unaware of what is or not a secret.

Let's wait some time and see if we ever need more components to use this feature and then decide if the environment manager should control the logic for all components or if we need a new component on top of the environment manager to handle it.

cli/azd/cmd/env.go
```diff
@@ -45,6 +50,17 @@ func envActions(root *actions.ActionDescriptor) *actions.ActionDescriptor {
 		ActionResolver: newEnvSetAction,
 	})
 
+	group.Add("set-secret", &actions.ActionDescriptorOptions{
```
weikanglim, line 53: Just curious about how you feel about this -- if the ultimate thing that happens is that we would set an env value, and we're avoiding introducing new azd types, would this command make more sense as `azd env set FOOBAR --secret`? 

I do very much see the perspectives you shared about avoiding adding additional concepts, just curious if it then makes more sense for us to expose it this way

vhvb1989: The problem of using `azd env set <name> <value> --secret` is that `env set` is currently fixed to take 2 arguments.

I can change it to take 1 or 2 but now it would introduce an invalid input state:
- azd env set FOO BAR  -> OK
- azd env set FOO BAR --secret ->  **Invalid!**  should azd fail?  or ignore the 2nd argument?  Is `BAR` a leaked secret in the terminal now?
- azd env set FOO --secret -> Valid
- azd env set FOO  ->  Invalid!!  Should azd just prompt for the value instead of failing?

This analysis is what made me go in favor of `azd env set-secret`.  I am still open to rename it to `azd env set-secret-reference` of `azd env set-akvs` .   I just think it has to:
- Be under `env` group because that's where it lands
- Has a unique name to explicitly be observed differently than `env set` and not just like a `env set` decoration

weikanglim: I very much agree that we're not in a great situation where we have to choose between a more dynamic contract for `azd env set` (which I agree very much with your proposed handling) or introduce a new command such as `azd env set-secret`.

With `azd env set-secret`, I don't find it to be self-documenting since a concept of a "secret" in `azd env` doesn't actually exist (which is a dealbreaker for me). From your suggestions, I do think `azd env set-secret-ref` sounds better since it does ultimately set a secret reference, but it isn't a user-friendly name and the concept itself is an implementation of "value expressions".

wbreza (approved):

cli/azd/cmd/env.go
```diff
@@ -150,6 +168,293 @@ func (e *envSetAction) Run(ctx context.Context) (*actions.ActionResult, error) {
...
+	entraIdService     entraid.EntraIdService
+	subResolver        account.SubscriptionTenantResolver
+	userProfileService *azapi.UserProfileService
+}
+
+func (e *envSetSecretAction) Run(ctx context.Context) (*actions.ActionResult, error) {
+
+	if len(e.args) < 1 {
+		return nil, fmt.Errorf(
+			"no <name> provided. Please provide a name as argument like: 'azd env set <name>'")
```
weikanglim, line 206:
```suggestion
			"no <name> provided. Please provide a name as argument like: 'azd env set-secret <name>'") 
```

cli/azd/cmd/env.go
```diff
@@ -150,6 +168,293 @@ func (e *envSetAction) Run(ctx context.Context) (*actions.ActionResult, error) {
...
+	selectExistingStrategy := "Select an existing Key Vault secret"
+	setSecretStrategies := []string{createNewStrategy, selectExistingStrategy}
+	selectedStrategyIndex, err := e.console.Select(
+		ctx,
+		input.ConsoleOptions{
+			Message:      "Select how you want to set " + secretName,
+			Options:      setSecretStrategies,
+			DefaultValue: createNewStrategy,
+			Help: "When creating a new Key Vault secret, you can either create a new Key Vault account or" +
+				" pick an existing one. A Key Vault secret belongs to a Key Vault account.",
```
weikanglim, line 220: Key Vault:
https://styleguides.azurewebsites.net/Styleguide/Read?id=2696&topicid=72112

```suggestion
			Help: "When creating a new Key Vault secret, you can either create a new Key Vault or" +
				" pick an existing one. A Key Vault secret belongs to a Key Vault.",
```

weikanglim: same elsewhere in the text

cli/azd/pkg/ext/models.go
```diff
@@ -70,6 +70,9 @@ type HookConfig struct {
 	Windows *HookConfig `yaml:"windows,omitempty"`
 	// When running on linux/macos use this override config
 	Posix *HookConfig `yaml:"posix,omitempty"`
+	// Environment variables in this list are added to the hook script and if the value is a akvs:// reference
+	// it will be resolved to the secret value
+	Secrets map[string]string `yaml:"secrets,omitempty"`
```
weikanglim, line 75: Just curious, do you feel that it is common for 

```yaml
secrets:
  - KV_SECRET:KV_SECRET
```

to be used; and if so, does:

```yaml
secrets:
  - KV_SECRET
```

feel better as syntactic sugar?

cli/azd/cmd/env.go
```diff
@@ -150,6 +168,293 @@ func (e *envSetAction) Run(ctx context.Context) (*actions.ActionResult, error) {
...
+			if kvAccountName == "" {
+				e.console.Message(ctx, "Key Vault account name cannot be empty")
+				continue
+			}
+			e.console.ShowSpinner(ctx, "Creating Key Vault account", input.Step)
+			vault, err := e.kvService.CreateVault(ctx, tenantId, subId, rg, location, kvAccountName)
+			e.console.StopSpinner(ctx, "", input.Step)
+			if err != nil {
+				e.console.Message(ctx, fmt.Sprintf("Error creating Key Vault account: %v", err))
+				continue
```
weikanglim, line 328: Does it make sense to separate out errors before re-prompting? I can imagine if the user doesn't have a permission to create KeyVault for example, this would end up re-prompting them.

cli/azd/cmd/env.go
```diff
@@ -150,6 +168,309 @@ func (e *envSetAction) Run(ctx context.Context) (*actions.ActionResult, error) {
...
+		}
+		e.console.StopSpinner(ctx, "", input.Step)
+	}
+
+	var kvSecretName string
+	if willCreateNewSecret {
+		for {
+			kvSecretName, err = e.console.Prompt(ctx, input.ConsoleOptions{
+				Message:      "Enter a name for the Key Vault secret",
+				DefaultValue: strings.ReplaceAll(secretName, "_", "-") + "-kv-secret",
```
weikanglim, line 365: I'm wondering how you feel about dropping the `-kv-secret` suffix here. I feel that after dropping the suffix, the suggested default would be closer to a value that the user will just press 'enter' and move on. Appending `-kv-secret` seemed a little less idiomatic in mind. I know we do suffix for Azure resources in general, but not for the properties of Azure resources (like secrets).

```suggestion
				DefaultValue: strings.ReplaceAll(secretName, "_", "-"),
```

cli/azd/cmd/env.go
```diff
@@ -150,6 +168,309 @@ func (e *envSetAction) Run(ctx context.Context) (*actions.ActionResult, error) {
...
+	if len(e.args) < 1 {
+		return nil, fmt.Errorf(
+			"no <name> provided. Please provide a name as argument like: 'azd env set-secret <name>'")
+	}
+	secretName := e.args[0]
+
+	// When AZD_ENV_SET_SECRET_LIST_WITHOUT_NUMBERS is defined in azd env, azd will not add numbers to the list when
+	// asking to select options. For example, instead of showing "1. Option 1", it will show "Option 1". This is useful
+	// when the user wants to prefill the selection in stdin before calling azd env set-secret (e.g. in a script).
+	_, listWithoutNumbers := e.env.LookupEnv("AZD_ENV_SET_SECRET_LIST_WITHOUT_NUMBERS")
```
weikanglim, line 213: Also, happy if we want to rely on `console.IsSpinnerInteractive()` (badly named -- it's simply a determination of `isTty`) to drop the numbers -- I know we do have precedence in other places of the codebase.

cli/azd/cmd/env.go
```diff
@@ -150,6 +168,309 @@ func (e *envSetAction) Run(ctx context.Context) (*actions.ActionResult, error) {
...
+			Help: "When creating a new Key Vault secret, you can either create a new Key Vault or" +
+				" pick an existing one. A Key Vault secret belongs to a Key Vault.",
+		})
+	if err != nil {
+		return nil, fmt.Errorf("selecting secret setting strategy: %w", err)
+	}
+
+	willCreateNewSecret := setSecretStrategies[selectedStrategyIndex] == createNewStrategy
+
+	subscriptionNote := "\nYou can set the Key Vault secret from any Azure Subscription where you have access to."
```
weikanglim, line 233: Based on the [style guide](https://styleguides.azurewebsites.net/Styleguide/Read?id=2696&topicid=72145), I'm guessing we would choose "Azure subscription"  

```suggestion
	subscriptionNote := "\nYou can set the Key Vault secret from any Azure subscription where you have access to."
```

cli/azd/cmd/env.go
```diff
@@ -150,6 +168,309 @@ func (e *envSetAction) Run(ctx context.Context) (*actions.ActionResult, error) {
...
+			}
+			kvAccountName = kvAccountNameInput
+			break
+		}
+
+		e.console.ShowSpinner(ctx, "Creating Key Vault", input.Step)
+		vault, err := e.kvService.CreateVault(ctx, tenantId, subId, rg, location, kvAccountName)
+		e.console.StopSpinner(ctx, "", input.Step)
+		if err != nil {
+			return nil, fmt.Errorf("Error creating Key Vault: %w", err)
```
weikanglim, line 342: if it helps with the linter rule...

```suggestion
			return nil, fmt.Errorf("error creating Key Vault: %w", err)
```

cli/azd/cmd/env.go
```diff
@@ -150,6 +168,309 @@ func (e *envSetAction) Run(ctx context.Context) (*actions.ActionResult, error) {
...
+		kvAccount = vaultsList[kvAccountSelectionIndex]
+	}
+
+	if willCreateNewKvAccount {
+		location, err := e.prompter.PromptLocation(ctx, subId, "Select the location to create the Key Vault", nil)
+		if err != nil {
+			return nil, fmt.Errorf("prompting for Key Vault location: %w", err)
+		}
+		rg, err := e.prompter.PromptResourceGroupFrom(ctx, subId, location, prompt.PromptResourceGroupFromOptions{
+			DefaultName: "rg-for-my-kv-account",
```
weikanglim, line 315: This reads closer to a placeholder value. I'm wondering if we could either suggest `rg-<project name>-kv`, or just avoid defaulting altogether.

weikanglim (approved):
LGTM! I expect that we will see users that are happy that they are able to use `azd` to easily store and retrieve their secrets in Key Vault.

cli/azd/cmd/env.go
```diff
@@ -45,6 +52,17 @@ func envActions(root *actions.ActionDescriptor) *actions.ActionDescriptor {
 		ActionResolver: newEnvSetAction,
 	})
 
+	group.Add("set-secret", &actions.ActionDescriptorOptions{
+		Command: &cobra.Command{
+			Use:   "set-secret <name>",
+			Short: "Set a <name> as a reference to a Key Vault secret in the environment.",
```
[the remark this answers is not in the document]
vhvb1989: https://github.com/Azure/azure-dev/issues/4734

cli/azd/cmd/env.go
vhvb1989: I didn't want to use the project name because it will be common to use one Key Vault for multiple projects.  This is a personal Key Vault that becomes handy cross-projects, if customer wants to.

We did remove the concept of a key vault account, so I do need to remove that part from here.

PR #4519, opened 2024-11-06
ci: Use TME via OIDC in build-cli

### LINKED ISSUE #4341: ci: Move away from client-secret in favor of service connections

### LINKED ISSUE #4501: [EngSys] Remove usage of azd-login.yml

ellismg:
Now that `azd` supports OIDC in Azure Pipelines via #4343, let's use it in the `build-cli` leg of CI. Since we now use OIDC, we can also migrate to the TME environment for the resources created during our tests, by using the new `azd-service-connection` service connection.

In addition to the build authoring updates, I needed to make some small changes to the integration testing framework and recording framework to get these changes to work when running in playback mode.

I added `AZD_DEBUG_LOGIN_FORCE_SUBSCRIPTION_REFRESH` which can be used to force `azd login` to load (and cache) subscriptions when you log in. This refresh does not usually happen for service principal based logins, but we want it to happen here so that the subscription list can be served from cache instead of hitting the `/subscriptions` endpoint of ARM to list subscriptions in each test (and record the result). This also ensures the environment during playback is a bit closer to the one like a developer will have when hacking on `azd` (since they too will have run `azd login` locally thus primed their subscription cache).

Moving to the TME user also exposed an issue with `azd` during playback if the list of subscriptions the user has does not include the subscription that was used for a recorded test. In this case, `azd` could end up trying to fetch information about this subscription, leading to test failures because this generates requests that don't have recorded interactions. To work around this issue, I've added a new `AZD_DEBUG_SYNTHETIC_SUBSCRIPTION` environment variable that can be set to a subscription ID. When set, the `SubscriptionManager` ensures the list of subscriptions returned by `GetSubscriptions` includes an entry for this subscription.

I added `AZD_DEBUG_LOGIN_FORCE_SUBSCRIPTION_REFRESH` which can be used to force `azd login` to load (and cache) subscriptions when you log in. This refresh does not usually happen for service principal based logins, but we want it to happen here so that the subscription list can be served from cache instead of hitting the `/subscriptions` endpoint of ARM to list subscriptions in each test (and record the result). This also ensures the environment during playback is a bit closer to the one like a developer will have when hacking on `azd` (since they too will have run `azd login` locally thus primed their subscription cache).

Moving to the TME user also exposed an issue with `azd` during playback if the list of subscriptions the user has does not include the subscription that was used for a recorded test. In this case, `azd` could end up trying to fetch information about this subscription, leading to test failures because this generates requests that don't have recorded interactions. To work around this issue, I've added a new `AZD_DEBUG_SYNTHETIC_SUBSCRIPTION` environment variable that can be set to a subscription ID. When set, the `SubscriptionManager` ensures the list of subscriptions returned by `GetSubscriptions` includes an entry for this subscription. The integration test framework arranges for this value to be set during playback. This allows the rest of the test to work (the fact that the principal running the test doesn't have access to the subscription used for the recording ends up not mattering, since we serve all requests from the recordings instead of against live Azure).

Finally, we needed to add the endpoint that is used inside Azure DevOps to fetch the ID token during the OIDC flow to the list of URLs that the recording infrastructure that should not be recorded.

With this change, the `azd-login` step is no longer used across any pipelines, so we can remove it.

Fixes #4341
Fixes #4501


eng/pipelines/templates/jobs/build-cli.yml
```diff
@@ -89,14 +89,37 @@ jobs:
...
+          scriptType: pscore
+          scriptLocation: inlineScript
+          inlineScript: |
+            $subscriptionId = (az account show -o json | ConvertFrom-Json -AsHashtable)['id']
+
+            Write-Host "##vso[task.setvariable variable=AzureSubscriptionClientId;issecret=false]$($env:AZURESUBSCRIPTION_CLIENT_ID)"
+            Write-Host "##vso[task.setvariable variable=AzureSubscriptionTenantId;issecret=false]$($env:AZURESUBSCRIPTION_TENANT_ID)"
+            Write-Host "##vso[task.setvariable variable=AzureSubscriptionServiceConnectionId;issecret=false]$($env:AZURESUBSCRIPTION_SERVICE_CONNECTION_ID)"
+            Write-Host "##vso[task.setvariable variable=SubscriptionId;issecret=false]$($subscriptionId)"
+        displayName: Configure OIDC Authentication
```
danieljurek, line 105: This should be in a template and put in https://github.com/Azure/azure-dev/blob/main/eng/pipelines/templates/steps/template-test-run-job.yml

eng/pipelines/templates/jobs/build-cli.yml
```diff
@@ -89,14 +89,37 @@ jobs:
...
+
+      - pwsh: |
+            ./azd auth login --federated-credential-provider "azure-pipelines"
+            ./azd auth token --output json
```
danieljurek, line 109: Looks like DevOps is redacting the output properly but we should not rely on that. Remove this line or extract the token and `Write-Host "##vso[task.setvariable variable=redactSecret;issecret=true]$($tokenOutput.token)"` before outputting it.

ellismg: I intend to remove it.

eng/pipelines/templates/jobs/build-cli.yml
```diff
@@ -116,14 +139,13 @@ jobs:
           # AZD live test setup variables
           CI: true
           AZD_TEST_CLI_VERSION: $(CLI_VERSION)
-          AZD_TEST_CLIENT_ID: $(arm-client-id)
-          AZD_TEST_TENANT_ID: $(arm-tenant-id)
+          AZD_TEST_CLIENT_ID: $(AzureSubscriptionClientId)
```
danieljurek, line 142: env map also needs `SYSTEM_ACCESSTOKEN: $(System.AccessToken)`

cli/azd/cmd/auth_login.go
```diff
@@ -338,7 +338,7 @@ func (la *loginAction) Run(ctx context.Context) (*actions.ActionResult, error) {
 		return nil, err
 	}
 
-	if la.flags.clientID == "" {
+	if v, err := strconv.ParseBool(os.Getenv("AZD_DEBUG_LOGIN_FORCE_SUBSCRIPTION_REFRESH")); (err == nil && v) || la.flags.clientID == "" {
```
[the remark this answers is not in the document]
ellismg: This seemed to help, but was not a complete solution - we still seem to be trying to pull subscription information (for the subscription id in the recording) without having a corresponding recorded interaction which breaks each end to end test.

eng/pipelines/templates/jobs/build-cli.yml
```diff
@@ -88,14 +88,25 @@ jobs:
...
+          AZURESUBSCRIPTION_SERVICE_CONNECTION_ID: $(AzureSubscriptionServiceConnectionId)
+          SYSTEM_ACCESSTOKEN: $(System.AccessToken)
+          AZD_DEBUG_LOGIN_FORCE_SUBSCRIPTION_REFRESH: true
+        displayName: AZD Login
 
       - task: AzureCLI@2
         condition: and(succeeded(), ne(variables['Skip.LiveTest'], 'true'))
         inputs:
-          azureSubscription: azure-sdk-tests
+          azureSubscription: azd-service-connection
```
danieljurek, line 109: Are terraform tests running in this context? If yes, no action required here. 

If no... 

`AzureCLI@2` should not be necessary here given that everything needed for auth is set in `configure-oidc-auth.yml`. Replacing this with a `PowerShell@2` (and no service connection) could take up to a minute off this task because we won't need to do any of the `az login` work that `AzureCLI@2` does (that's all handled by `configure-oidc-auth.yml`)

danieljurek (approved):
LGTM. If no terraform tests, you can save `az login` time during the test step. If not, no changes needed from the EngSys perspective.

eng/pipelines/templates/jobs/build-cli.yml
danieljurek: Looks like terraform tests are skipped: 

image

ellismg: I want to get these terraform tests working again shortly, so I will leave this as is for now.

PR #4522, opened 2024-11-06
ci: Use ODIC and TME for `test-templates`

ellismg:
This change reworks the `test-templates` test legs to use OIDC and move to our TME subscription.

To do so, we capture the relevent `AZURESUBSCRIPTION_` environment variables that are set by the AzureCLI task and then flow them into the dev container when we run tests. This allows us to do `azd auth login` inside the devcontainer but still use OIDC so we don't have a long lived secret.

The call to `az login` was removed because we don't yet have a way to have this work well with ODIC in the container. This means the terraform flavor of the tests won't work (since auth is broken) but these legs have been broken in both the bicep and terraform flavors for a while, so this moves us in a better direction (and gets us off of client secrets and onto TME for these tests).

Contributes To #4341

### DIFF
--- a/eng/pipelines/templates/steps/configure-oidc-auth.yml
+++ b/eng/pipelines/templates/steps/configure-oidc-auth.yml
@@ -1,0 +1,16 @@
+parameters:
+  ServiceConnectionName: "azd-service-connection"
+
+steps:
+  - task: AzureCLI@2
+    condition: and(succeeded(), ne(variables['Skip.LiveTest'], 'true'))
+    inputs:
+      azureSubscription: ${{ parameters.ServiceConnectionName }}
+      scriptType: pscore
+      scriptLocation: inlineScript
+      inlineScript: |
+        Write-Host "##vso[task.setvariable variable=AzureSubscriptionClientId;issecret=false]$($env:AZURESUBSCRIPTION_CLIENT_ID)"
+        Write-Host "##vso[task.setvariable variable=AzureSubscriptionTenantId;issecret=false]$($env:AZURESUBSCRIPTION_TENANT_ID)"
+        Write-Host "##vso[task.setvariable variable=AzureSubscriptionServiceConnectionId;issecret=false]$($env:AZURESUBSCRIPTION_SERVICE_CONNECTION_ID)"
+        Write-Host "##vso[task.setvariable variable=SubscriptionId;issecret=false]$((az account show -o json | ConvertFrom-Json -AsHashtable)['id'])"
+    displayName: Configure OIDC Authentication
--- a/eng/pipelines/templates/steps/template-test-run-job.yml
+++ b/eng/pipelines/templates/steps/template-test-run-job.yml
@@ -12,14 +12,10 @@ steps:
       parameters:
           Version: $(AzdVersion)
 
-    - template: /eng/pipelines/templates/steps/azd-login.yml
-
-    # Java pre-requisites that isn't met
-    # TODO: Use azd container as a form of validation that the container works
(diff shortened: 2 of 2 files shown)


ellismg:
Here's an internal run of the test-templates leg: https://dev.azure.com/azure-sdk/internal/_build/results?buildId=4307100&view=results

Much better than we were before - and for the legs that fail I don't think it's due to OIDC, we were always able to provision the required infrastructure.

eng/pipelines/templates/steps/template-test-run-job.yml
```diff
@@ -12,15 +12,22 @@ steps:
...
-
-    # Java pre-requisites that isn't met
-    # TODO: Use azd container as a form of validation that the container works
-    - template: /eng/pipelines/templates/steps/install-ms-openjdk.yml
-
     # Required to clone repos that are not yet public
     - template: /eng/pipelines/templates/steps/set-git-credentials.yml
 
+    # Use the AzureCLI@2 task to set the variables azd will need to use later to authenticate over OIDC.
+    - task: AzureCLI@2
```
ellismg, line 19: I'll move this into its own step, so we can share it with what I'm doing for `build-cli`.

eng/pipelines/templates/steps/configure-oidc-auth.yml
```diff
@@ -0,0 +1,16 @@
+parameters:
+  ServiceConnectionName: ""
```
danieljurek, line 2: The default value should be `azd-service-connection` here so you don't have to duplicate that into the other templates

danieljurek:
Since it's difficult to set a regex which excludes the `-terraform` templates it might make sense to set a condition on the matrix jobs themselves: 

in [run-template-tests.yml](https://github.com/Azure/azure-dev/blob/main/eng/pipelines/templates/jobs/run-template-tests.yml)

```yaml
  - job: TemplateTest_RunTest
    condition: not(contains(variables['TemplateName'], 'terraform'))    # something like this could work
    timeoutInMinutes: 180
    dependsOn: TemplateTest_CreateMatrix
```


That way it's obvious that the terraform jobs would've been created but they were skipped... one could adjust the condition to re-enable them. 

Other efforts to exclude terraform could include: 
* Explicitly excluding it in matrix generation 
* Changes to `azd template list` to exclude terraform
* An exclusion regex that removes matches after the `TemplateListFilter` is processed

danieljurek:
> Since it's difficult to set a regex which excludes the `-terraform` templates it might make sense to set a condition on the matrix jobs themselves:
> 
> in [run-template-tests.yml](https://github.com/Azure/azure-dev/blob/main/eng/pipelines/templates/jobs/run-template-tests.yml)
> 
> ```yaml
>   - job: TemplateTest_RunTest
>     condition: not(contains(variables['TemplateName'], 'terraform'))    # something like this could work
>     timeoutInMinutes: 180
>     dependsOn: TemplateTest_CreateMatrix
> ```
> 
> That way it's obvious that the terraform jobs would've been created but they were skipped... one could adjust the condition to re-enable them.
> 
> Other efforts to exclude terraform could include:
> 
> * Explicitly excluding it in matrix generation
> * Changes to `azd template list` to exclude terraform
> * An exclusion regex that removes matches after the `TemplateListFilter` is processed

Further testing reveals the skipping approach does not work. Even in static matrices, matrix variables don't appear to be available to jobs... The best option for future users of CI to understand intent here is probably an exclusion option.

danieljurek (approved):

danieljurek:
Approving now as this is correct and moves us into a better place. Terraform jobs will still fail but more work will be needed to get this pipeline totally green and migrating to TME and eliminating the secret are higher priority.

PR #4526, opened 2024-11-07
Fix some errors found in scenario: frontend + backend + mongodb

rujche:
Fix some errors found in scenario: frontend + backend + mongodb

 - Check the comments below to see the detailed error.
 - Link to related PR: https://github.com/azure-javaee/azure-dev/pull/14
 - Link to sample project: https://github.com/rujche/todo-java-mongo-aca/tree/sample-project-used-todemo-detect-mongodb-in-azd

### DIFF
--- a/cli/azd/resources/scaffold/templates/resources.bicept
+++ b/cli/azd/resources/scaffold/templates/resources.bicept
@@ -209,6 +209,11 @@ var {{bicepName .Name}}Env = map(filter({{bicepName .Name}}AppSettingsArray, i =
 
 module {{bicepName .Name}} 'br/public:avm/res/app/container-app:0.8.0' = {
   name: '{{bicepName .Name}}'
+  dependsOn: [
+    {{- if .DbCosmosMongo}}
+    cosmos
+    {{- end}}
+  ]
   params: {
     name: '{{.Name}}'
     {{- if ne .Port 0}}
@@ -220,6 +225,12 @@ module {{bicepName .Name}} 'br/public:avm/res/app/container-app:0.8.0' = {
         {{- range .Backend.Frontends}}
         'https://{{.Name}}.${containerAppsEnvironment.outputs.defaultDomain}'
         {{- end}}
+      ]
+      allowedMethods: [
+        'GET'
+        'PUT'
+        'POST'
+        'DELETE'
       ]
     }
     {{- end}}
@@ -343,7 +354,7 @@ module {{bicepName .Name}} 'br/public:avm/res/app/container-app:0.8.0' = {
           {{- range $i, $e := .Frontend.Backends}}
           {
             name: '{{upper .Name}}_BASE_URL'
-            value: 'https://{{.Name}}.internal.${containerAppsEnvironment.outputs.defaultDomain}'
+            value: 'https://{{.Name}}.${containerAppsEnvironment.outputs.defaultDomain}'
           }
           {{- end}}
           {{- end}}
--- a/cli/azd/resources/scaffold/templates/resources.bicept
+++ b/cli/azd/resources/scaffold/templates/resources.bicept
@@ -66,9 +66,11 @@ module containerAppsEnvironment 'br/public:avm/res/app/managed-environment:0.4.5
 {{- end}}
 
 {{- if .DbCosmosMongo}}
-module cosmos 'br/public:avm/res/document-db/database-account:0.4.0' = {
+module cosmos 'br/public:avm/res/document-db/database-account:0.8.1' = {
   name: 'cosmos'
   params: {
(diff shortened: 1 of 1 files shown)


cli/azd/resources/scaffold/templates/resources.bicept
```diff
@@ -209,6 +209,11 @@ var {{bicepName .Name}}Env = map(filter({{bicepName .Name}}AppSettingsArray, i =
 
 module {{bicepName .Name}} 'br/public:avm/res/app/container-app:0.8.0' = {
   name: '{{bicepName .Name}}'
+  dependsOn: [
+    {{- if .DbCosmosMongo}}
+    cosmos
+    {{- end}}
+  ]
```
rujche, line 216: Without this, the cosmos may created after ACA, it will return error like "can not get mongodb-url using managed identity xxx". This error is not caused of authorization, it's caused by the secret not written to keyvault yet.

cli/azd/resources/scaffold/templates/resources.bicept
```diff
@@ -221,6 +226,12 @@ module {{bicepName .Name}} 'br/public:avm/res/app/container-app:0.8.0' = {
         'https://{{.Name}}.${containerAppsEnvironment.outputs.defaultDomain}'
         {{- end}}
       ]
+      allowedMethods: [
+        'GET'
+        'PUT'
+        'POST'
+        'DELETE'
+      ]
```
rujche, line 234: Without this, the frontend app can not do write / delete operation.

cli/azd/resources/scaffold/templates/resources.bicept
```diff
@@ -343,7 +354,7 @@ module {{bicepName .Name}} 'br/public:avm/res/app/container-app:0.8.0' = {
           {{- range $i, $e := .Frontend.Backends}}
           {
             name: '{{upper .Name}}_BASE_URL'
-            value: 'https://{{.Name}}.internal.${containerAppsEnvironment.outputs.defaultDomain}'
+            value: 'https://{{.Name}}.${containerAppsEnvironment.outputs.defaultDomain}'
```
rujche, line 357: The front end can not access the URL with `internal.` inside. Delete `internal.` will make it work well.

cli/azd/resources/scaffold/templates/resources.bicept
weikanglim: I see. Can you look into if there's a way we can leverage [exportedSecrets](https://github.com/Azure/bicep-registry-modules/blob/main/avm/res/document-db/database-account/README.md#Outputs)?

cli/azd/resources/scaffold/templates/resources.bicept
weikanglim: That is interesting. Can you check with `http://{{.Name}}` works as well? `internal` was in their previous documentation, it looks like it did end up changing based on these new [docs](https://learn.microsoft.com/en-us/azure/container-apps/connect-apps?tabs=bash).

I will have to play around with this a little to test this out, will get back to you in a day or two.

cli/azd/resources/scaffold/templates/resources.bicept
weikanglim: I see; it's likely that the allowedMethods is just `GET` `PUT` `POST` as the default. 

Does `*` work? I think based on these [docs](https://learn.microsoft.com/en-us/azure/container-apps/cors?tabs=arm&pivots=azure-portal#enable-and-configure-cors) it should.

Otherwise, we may have to worry about `HEAD` requests that seem to be common these days with browser pre-flight checks.

weikanglim:
Thanks for this contribution @rujche , very clean and targeted. I left some ideas we could explore further if it makes sense to you. I'll also need to do some testing on my end, and get back to you on some final conclusions.

cli/azd/resources/scaffold/templates/resources.bicept
rujche: Updated.

cli/azd/resources/scaffold/templates/resources.bicept
rujche: Updated.

cli/azd/resources/scaffold/templates/resources.bicept
rujche: I tried `http://{{.Name}}`, it doesn't work. It's not hard to understand: When the `API_BASE_URL` used in frontend application, it is used in js file in browser, it's not internal network.

rujche:
Hi, @weikanglim .
Thank you very much for your quick response and meaningful comments! 
I added a new commit according to your comments, please help to review again.

weikanglim (approved):
LGTM! Thanks for fixing the issue here!

PR #4527, opened 2024-11-07
compose: add command

weikanglim:
`add` command that adds resource configuration to `azure.yaml`.

In this experience, users are able to proceed with running `provision` or `up` after committing the `azure.yaml` change.


cli/azd/internal/cmd/add/add.go
```diff
@@ -0,0 +1,392 @@
...
+		panic(fmt.Sprintf("encoding yaml node: %v", err))
+	}
+
+	err = yamlnode.Set(&doc, fmt.Sprintf("resources?.%s", resourceToAdd.Name), resourceNode)
+	if err != nil {
+		return nil, fmt.Errorf("setting resource: %w", err)
+	}
+
+	for _, svc := range usedBy {
+		err = yamlnode.Append(&doc, fmt.Sprintf("resources.%s.uses[]?", svc), &yaml.Node{
```
weikanglim, line 176: verify `uses` beforehand for duplicates somehow

weikanglim: no duplicates possible; but I did end up exposing host bi-directional uses and hardened it to avoid cycles there

cli/azd/cmd/root.go
```diff
@@ -320,6 +321,11 @@ func NewRootCmd(
 			},
 		}).
 		UseMiddleware("hooks", middleware.NewHooksMiddleware)
+	root.
+		Add("add", &actions.ActionDescriptorOptions{
+			Command:        add.NewAddCmd(),
+			ActionResolver: add.NewAddAction,
+		})
```
vhvb1989, line 328: Can you add `HelpOptions` and `GroupingOptions` ?

Like
```
HelpOptions: actions.ActionHelpOptions{
			Description: getCmdEnvHelpDescription,
		},
		GroupingOptions: actions.CommandGroupOptions{
			RootLevelHelp: actions.CmdGroupManage,
		},
```

See the examples from other commands.  It should improve the docs for the help and you can add a few examples on the bottom

cli/azd/internal/appdetect/docker.go
```diff
@@ -11,18 +11,19 @@ import (
...
 
 	return nil, nil
 }
 
-func detectDockerFromFile(dockerFilePath string) (*Docker, error) {
+// DetectDocker detects Docker from a Dockerfile path.
+func DetectDocker(dockerFilePath string) (*Docker, error) {
 	file, err := os.Open(dockerFilePath)
 	if err != nil {
 		return nil, fmt.Errorf("reading Dockerfile at %s: %w", dockerFilePath, err)
```
vhvb1989, line 29: How you feel about having a non-exported method calling an exported method?  I usually advocate against it.   If you share the feeling, can you update this to make `DetectDocker` to call `detectDockerInDirectory` ?  (moving the implementation of `DetectDocker` to `detectDockerDirectory`

cli/azd/cmd/root.go
```diff
@@ -320,6 +321,11 @@ func NewRootCmd(
 			},
 		}).
 		UseMiddleware("hooks", middleware.NewHooksMiddleware)
+	root.
+		Add("add", &actions.ActionDescriptorOptions{
```
vhvb1989, line 325: Should we condition adding this command based on enabling the alpha feature?   Like, show the command in the list only when the feature is enabled, otherwise we don't show it?

cli/azd/internal/cmd/add/add.go
```diff
@@ -0,0 +1,379 @@
...
+
+	prjConfig, err := project.Load(ctx, a.azdCtx.ProjectPath())
+	if err != nil {
+		return nil, err
+	}
+
+	selectMenu := a.selectMenu()
+	slices.SortFunc(selectMenu, func(a, b Menu) int {
+		return strings.Compare(a.Label, b.Label)
+	})
```
vhvb1989, line 77: nit.  Should `selectMenu()` return a sorted-by-label slice?

cli/azd/internal/cmd/add/add_configure_host.go
```diff
@@ -0,0 +1,256 @@
...
+	"github.com/azure/azure-dev/cli/azd/pkg/input"
+	"github.com/azure/azure-dev/cli/azd/pkg/output"
+	"github.com/azure/azure-dev/cli/azd/pkg/output/ux"
+	"github.com/azure/azure-dev/cli/azd/pkg/project"
+	"github.com/fatih/color"
+)
+
+func (a *AddAction) configureHost(
+	console input.Console,
+	ctx promptCtx) (*project.ServiceConfig, *project.ResourceConfig, error) {
```
vhvb1989, line 26: move the context to be the first arg

vhvb1989: Seems like you added it to all methods at the end.  Curious about why?

cli/azd/internal/cmd/add/add_configure.go
```diff
@@ -0,0 +1,209 @@
...
+func fillUses(
+	r *project.ResourceConfig,
+	console input.Console,
+	ctx promptCtx) (*project.ResourceConfig, error) {
+	type resourceDisplay struct {
+		Resource *project.ResourceConfig
+		Display  string
+	}
+	res := make([]resourceDisplay, 0, len(ctx.prj.Resources))
+	for _, r := range ctx.prj.Resources {
```
vhvb1989, line 128: I'm concern about using `ctx.prj`  the way it currently is.  Since it is a pointer, there is a big assumption about the caller will aways make sure to set this pointer, but if that doesn't happen, there will be a `panic` on runtime.

Are you creating a context type instead of using `context.WithValue(parent, key, value)` ??  I think that might be a more idiomatic way to work with just one context type.

vhvb1989 (commented):
Partial review.

cli/azd/internal/cmd/add/add_configure.go
```diff
@@ -0,0 +1,209 @@
...
+	switch r.Type {
+	case project.ResourceTypeHostContainerApp:
+		return fillUses(r, console, ctx)
+	case project.ResourceTypeOpenAiModel:
+		return fillAiModelName(r, console, ctx)
+	case project.ResourceTypeDbPostgres,
+		project.ResourceTypeDbMongo:
+		return fillDatabaseName(r, console, ctx)
+	case project.ResourceTypeDbRedis:
+		r.Name = "redis"
```
ellismg, line 35: Sort of surprised not to see a `fillRedisName`?  Do we not need to prompt the user for the name of this?  Is the assumption that the user would not want to control the name for some reason?

cli/azd/internal/cmd/add/add_configure.go
```diff
@@ -0,0 +1,209 @@
...
+				defaultName = fmt.Sprintf("%s-%d", defaultName, i)
+			} else {
+				break
+			}
+		}
+	}
+
+	for {
+		modelName, err := console.Prompt(ctx, input.ConsoleOptions{
+			Message:      "What is the name of this model?",
```
ellismg, line 100: Just wondering (maybe we'll discover this from user studies) - I assume this is the resource name of the model, not something like "gpt-3.5" right?  Do users understand what "name of this model" means here?

cli/azd/internal/cmd/add/add_preview.go
```diff
@@ -0,0 +1,197 @@
...
+	case project.ResourceTypeDbMongo:
+		res.AzureResourceType = "Microsoft.DocumentDB/databaseAccounts/mongodbDatabases"
+		res.UseEnvVars = []string{
+			"MONGODB_URL",
+		}
+	case project.ResourceTypeOpenAiModel:
+		res.AzureResourceType = "Microsoft.CognitiveAccounts/accounts/deployments"
+		res.UseEnvVars = []string{
+			"AZURE_OPENAI_ENDPOINT",
+			"Keyless (Microsoft Entra ID)",
```
ellismg, line 66: There's an environment variable floating around with this name?  Interesting!

cli/azd/internal/cmd/add/add_select_ai.go
```diff
@@ -0,0 +1,192 @@
...
+		if err != nil {
+			return nil, err
+		}
+
+		cred, err := a.creds.CredentialForSubscription(ctx, a.env.GetSubscriptionId())
+		if err != nil {
+			return nil, fmt.Errorf("getting credentials: %w", err)
+		}
+
+		pipeline, err := armruntime.NewPipeline(
```
ellismg, line 47: Guessing there's no wrapper in `azure-sdk-for-go` for this part of the management plane and so we had to do our own thing?

cli/azd/internal/cmd/add/add_select_ai.go
```diff
@@ -0,0 +1,192 @@
...
+
+	displayModels := make([]string, 0, len(allModels))
+	models := make([]Model, 0, len(allModels))
+	for _, model := range allModels {
+		models = append(models, model.Model)
+		displayModels = append(displayModels, fmt.Sprintf("%s\t%s", model.Model.Name, model.Model.Version))
+	}
+
+	if console.IsSpinnerInteractive() {
+		displayModels, err = output.TabAlign(displayModels, 5)
```
ellismg, line 146: Just wondering - in the other two calls to `output.TabAlign` we passed `3` not `5`.  Was it intended that this is more than the others?

cli/azd/internal/cmd/add/util.go
```diff
@@ -0,0 +1,170 @@
...
+		path, err = filepath.Abs(path)
+		if err != nil {
+			return "", err
+		}
+
+		return path, err
+	}
+}
+
+// pathHasInfraModule returns true if there is a file named "<module>" or "<module.bicep>" in path.
```
ellismg, line 131:
```suggestion
// pathHasInfraModule returns true if there is a file named "<module>" or "<module.ext>" in path.
```

It seems like this doesn't not require the `.bicep` extension, since we end up just trimming the result of `filepath.Ext`?

cli/azd/internal/repository/app_init.go
```diff
@@ -560,3 +504,77 @@ func (i *Initializer) prjConfigFromDetect(
...
+		}
+	}
+
+	if prj.HasWebUIFramework() {
+		// By default, use 'dist'. This is common for frameworks such as:
+		// - TypeScript
+		// - Vite
+		svc.OutputPath = "dist"
+
+	loop:
```
ellismg, line 557: NIT: I think that with the refactoring to extract this in to it's own function, you can now remove the label and just call `break`?

ellismg (commented):
Very exciting to see this!  Thanks for all the work over the past few weeks with the smaller PRs - this was pretty focused and easy for me to digest.

I think the thing I care most about this the `promptCtx` type.  Given how stuff shook out I wonder if we really want to be tying the `context.Context` and `*project.ProjectConfig` together like this.  It doesn't seem like it makes the parameter list of methods that use it that much nicer, and I'll admit that I'm sort of always wary about combining the `context.Context` with other stuff.

Added some comments about things that jumped out at me and tried to respond to Victor's.

I'll admit I didn't examine the diff generation logic in great detail yet.

Overall though this feels really good, great job here, @weikanglim!

cli/azd/internal/cmd/add/add_configure.go
vhvb1989: > which not what we want here.

Why not?  I think what we want is to use a context as a backpack to carry data across the stack (calling one method after another) without making all methods to know about all those parameters.
For example, consider methods A, B, C and D.   A will call B, then B calls C, and finally C calls D.
Now, D requires param `foo`, but neither `C` or `B` uses `foo`.
By using a context backpack, A can call `withValue(ctx, fooKey, foo)` and call B(ctx).
When `D` is finally called, it will try to pull `foo` from the `ctx`.   It will panic or fail if `foo` is not in the ctx.

The composite approach from Wei comes with some type safety, but it would still let `B` and `C` to see and even update/mutate the current context... which I think shouldn't happen.   If B or C would like to change the scope for `D`, they should create a new ctx with a new `foo` :)

cli/azd/internal/cmd/add/add_configure.go
```diff
@@ -0,0 +1,209 @@
...
+	"strings"
+	"unicode"
+
+	"github.com/azure/azure-dev/cli/azd/pkg/input"
+	"github.com/azure/azure-dev/cli/azd/pkg/output"
+	"github.com/azure/azure-dev/cli/azd/pkg/project"
+	"github.com/fatih/color"
+)
+
+type promptCtx struct {
```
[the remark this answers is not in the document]
weikanglim: @vhvb1989 @ellismg 

I can see how at a glance this may be slightly confusing, and can understand where the concern is coming from. I'm overall happy to take this in any direction that both you and @vhvb1989 decides in, but I do want to set up the stage for the discussion on why this is implemented this way:

The design language I'm going for here is that there are common prompt options that are reshared for all general prompting functions within `add` itself, which itself has layers to it. It's important for all prompts to be able to leverage these options without explicitly passing at each layer. The azure-sdk-for-go analogy here is that these are `arm.clientOptions` for client functionality, i.e. `promptCtx` for prompt functionality.

Some ideas that I think could help alleviate the initial concerns:

1. We could name it `promptOptions` similar to `arm.clientOptions`. The upside here is that it's more consistent to the go design of "invoke a function" with "grouped optional options". The downside here for me comes down to how it doesn't handle prompt nesting well. Prompts are by nature, more nested, and I'd like the flexibility offered by having this general context member that can easily be flowed through multiple prompt callstacks.
2. We could avoid struct embedding here and have `promptOptions` with `ctx context.Context` as an explicit member.  The code does end up being slightly more verbose in everywhere this is being used, i.e. `console.Message(ctx, "")` becomes `console.Message(po.ctx, "")`. It also does makes it harder to lift-and-shift code from this particular package into other use-cases -- which could be useful to have while we solidify the UI patterns here.

Once again, happy to move in either way, but I just wanted to surface my deeper level thoughts here on how this makes sense in coding up an UI prompt domain.

cli/azd/internal/cmd/add/add_configure.go
weikanglim: I could see from an infrastructure management perspective and knowing how Azure works; it may be nice to have a name provided.

From an app developer perspective, I wouldn't expect to be prompted for a redis database name; generally, redis databases are unnamed and numbered: https://www.digitalocean.com/community/cheatsheets/how-to-manage-redis-databases-and-keys.

We don't support N-databases yet (it'd be important to work through what this actually looks like from an infra perspective), so the user doesn't lose any functionality with the current flow.

What do you think an experience for Redis naming would look like?

cli/azd/internal/appdetect/docker.go
weikanglim: The "gut feeling" that @vhvb1989 has is valid. The `appdetect` package is primarily focused on detection of filesystem projects with the input being the filesystem. Here, we're exposing a gut internal of the package, because we simply need it. 

The long-term plan is that if these patterns occur over time, we'd separate out the functionality into a different package for clean API design. For example, the packages `appdetect` and perhaps `appanalyze`; which `appdetect` uses `appanalyze` to perform file-level analysis.

For pragmatic reasons, I think we should keep this one-line of exposed functionality within the `appdetect` package for now. We're not shipping a public library that the entire world is dependent on; let's focus on solving azd user problems over the cleanest, purest library designs for now.

I'll rename it to `AnalyzeDocker` to call out the difference in API designs, which also set us up for a package separation if that need ever arises. But we do need to expose the gut internals right now for the best UX experience, so let's not compromise on that.

weikanglim: > I'll just use this answer from copilot to describe where my mental-model is coming from

Just a small thing I'd like recommend as a guideline for all of us in the interest of clarity of discussion: it'd be best for us to summarize our own positions from the AI's response. Doing so allows us to understand each other's position. The AI's response is neither our position, nor is it fact-checked -- we should think about parse through what the AI is saying before presenting it as evidence.

I'm also going to note briefly that "non-exposed method calling an exposed-method from the same struct." is not the situation we're describing here; we're discussing package/API/library design.

cli/azd/cmd/root.go
weikanglim: @vhvb1989

I believe based on `azd help` is currently working, which is based on the work you had previously done to separate out our help from `cobra.cmd`, our root help [menu](https://github.com/Azure/azure-dev/blob/main/cli/azd/cmd/testdata/TestUsage-azd.snap) doesn't automatically light up new commands. It actually requires explicit configuration within the codebase. You may notice that TestUsage-azd.snap is unchanged in this PR.

Correct if I'm mistaken, but the UX concern about "listing the command only when it is enabled" isn't applicable here, since it is never listed. I think this is acceptable for alpha.

On the overall UX thinking, I think it's favorable to have `azd <alpha-command> --help` still show up, and I think it's favorable to have `azd <alpha-command>` provide a follow-up on what to do next. This is much preferable to having "invisible commands". As always, I'm happy to take an issue or a design proposal for discussion separately.

weikanglim: Captured the line item for alpha advancement to surface in root menu in #4538

cli/azd/cmd/root.go
weikanglim: Tracked with #4538

cli/azd/internal/cmd/add/add.go
```diff
@@ -0,0 +1,379 @@
...
+
+	if serviceToAdd != nil {
+		serviceNode, err := yamlnode.Encode(serviceToAdd)
+		if err != nil {
+			panic(fmt.Sprintf("encoding yaml node: %v", err))
+		}
+
+		err = yamlnode.Set(&doc, fmt.Sprintf("services?.%s", serviceToAdd.Name), serviceNode)
+		if err != nil {
+			return nil, fmt.Errorf("finding services: %w", err)
```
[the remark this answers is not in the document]
weikanglim: ack, and agreed

cli/azd/internal/cmd/add/add.go
```diff
@@ -0,0 +1,379 @@
...
+	encoder := yaml.NewEncoder(file)
+	encoder.SetIndent(2)
+	// preserve multi-line blocks style
+	encoder.SetAssumeBlockAsLiteral(true)
+	err = encoder.Encode(&doc)
+	if err != nil {
+		return nil, fmt.Errorf("failed to encode: %w", err)
+	}
+
+	err = file.Close()
```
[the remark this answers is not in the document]
weikanglim: It felt right for "let's try to at least clean up in all cases" as I was writing `os.OpenFile`, preferring an extra cleanup if we ended up erroring out early.

cli/azd/internal/cmd/add/add.go
weikanglim: I do prefer it this way -- I'm setting up `Menu` as a data contract of items to display, but I'd prefer not making it a UI component for now. The UI component is actually within `add.go` that can decide how to render elements based on what is best for the presentation.

cli/azd/internal/cmd/add/add_select_ai.go
weikanglim: Indeed, it's a pretty undocumented API right now

cli/azd/internal/cmd/add/util.go
weikanglim: ack and agreed

cli/azd/internal/cmd/add/add_select_ai.go
weikanglim: It was an intentional change that does feel better here.

The other texts have a bracket label for one of the columns, i.e.:

`[Container App] api`

which feels good.

Here we have

`gpt-4o     2023-04-11` 

which feels better with more space

cli/azd/internal/cmd/add/add_configure.go
ellismg: > We don't support N-databases yet (it'd be important to work through what this actually looks like from an infra perspective), so the user doesn't lose any functionality with the current flow.
> 
> What do you think an experience for Redis naming would look like?

Not sure - the thing I sort of wondered about was at what level this name needs to be unique?  I was naively assuming that this would influence the physical name and that without the user being able to control it you'd only be able to create one of these per subscription?

If we aren't going to run into uniqueness issues on the name then I'm fine deferring for now - I agree that having multiple redis resources would be uncommon.

cli/azd/internal/cmd/add/add.go
```diff
@@ -0,0 +1,378 @@
...
+	resourceToAdd, err = configure(resourceToAdd, a.console, ctxPrompt)
+	if err != nil {
+		return nil, err
+	}
+
+	usedBy, err := promptUsedBy(resourceToAdd, a.console, ctxPrompt)
+	if err != nil {
+		return nil, err
+	}
+
```
weikanglim, line 123: add an invariant check for name

cli/azd/internal/cmd/add/add_preview.go
weikanglim: Removed this for now

cli/azd/cmd/testdata/TestUsage-azd-add.snap
```diff
@@ -0,0 +1,18 @@
+
+Add a component to your application. (Alpha)
```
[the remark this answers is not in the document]
weikanglim: Sticking to component for now

cli/azd/internal/cmd/add/add_configure.go
weikanglim: Separated out as options for now

cli/azd/internal/cmd/add/add_configure_host.go
weikanglim: Reordered parameters

ellismg (approved):
Thanks for the promptOptions/context decoupling!  Really excited to see what the larger world thinks of this!  Great job, @weikanglim!

vhvb1989 (approved):

cli/azd/internal/cmd/add/add_configure.go
weikanglim: Added validation for this, track the item with #4538

PR #4531, opened 2024-11-08
init: surface minimal option

weikanglim:
This change surfaces "Create a minimal project" as a main menu option in `init`.

**Background**: Before this change, initializing a minimal project would require selecting "Initialize a template -> Minimal". This was found to be confusing for users during user studies. There has also been general user feedback about discoverability with this flow.

**Functional changes**: When compose is enabled, only a project file is generated. When compose is not enabled, the previous functional behavior is retained -- no UX changes made.

UX (compose enabled):
image


ellismg (approved):
LGTM - Love the new UX!

PR #4535, opened 2024-11-09
Prompt for resources with optional resourceType

### LINKED ISSUE #4530: Support resource picker for parameters
`azd` supports a few "pickers" for parameters without default values today e.g.,

```bicep
@metadata({
    azd: {
        type: 'location'
    }
})
param frontendLocation string
```

Just like for any parameter named `location`, you'll prompt with a list of locations. This is also supported for [a couple other types](https://github.com/Azure/azure-dev/blob/f030c4c553ea089f16f43dbcb944d7834d5804aa/cli/azd/pkg/azure/arm_template.go#L100-L104) like `generate`.

It would be great to add support to query resources of a specific type e.g.,

```bicep
@metadata({
    azd: {
        type: 'resource'
        resourceType: 'Microsoft.OperationalInsights/workspaces'
    }
})
param frontendLocation string
```

This would list all the resources of that type discoverable to the user. Ideally, the display would put the location in parenthesis since they may want to limit those resources to just a location but you would have a hard time doing it above in any meaningful way because you'd either have to author it up front (not flexible), tie it to the `location` parameter (which you'd have to parse as an expression), or maybe just define a `useLocation: bool` parameter and use the `location` after you've already prompted/assigned it. Not sure how useful any of that would be, but listing names like "{name} ({location})" would be easy and useful, since ARM gives you back the locations as well.

E.g. `az` CLI command:

```bash
az resource list --resource-type Microsoft.OperationalInsights/workspaces
```

## Scenarios

The reason why this would be useful is to provide a friendly interface for selecting resources you may want to reuse like AppInsights workspaces (often aggregates apps; this is my impetus for filing this feature request), dashboards (again, often aggregates), or for more expensive resources with multi-project support like Cognitive Services that may also pull from different organizational resources anyway.

It's possible now by passing an environment variable and using that in your `.parameters.json` file, but it's more work up front and doesn't really fit the UX you already have: you prompt for subscription and location already. Those could also be passed as required parameters but you provide a friendly UX to list them. Getting the resource ID of an existing resources can be a rather daunting task, so it seems that could be made easier as well for those "Getting started" scenarios that some `azd` users are trying to provide for their downstream users e.g., if we used `azd` in our Azure SDKs' "Getting started" sections instead of `az` CLI commands (an idea I recently proposed for Azure SDK for Rust).

heaths:
Resolves #4530


heaths:
Alternatively, maybe instead of a `resourceType` we just have a `filter`. Then template authors can use whatever filter they want, such as `resourceType eq 'Microsoft.KeyVault/vaults' and location eq $location`. Though, to really make that work we'd need to support variable replacement, which also means resolving parameters sequentially, ideally doing location before all others because it'd be a likely filter.

heaths:
I decided to make `resourceType` required for use in templates. As I was thinking through the scenarios and considering how to approach documentation, I realized that any resource is useless in a template (but allow the underlying prompting functions to do it in case that's useful down the road). You can't declare an `existing` resources in Bicep without knowing the type. You can't use ARM or Bicep template functions `reference()` or `resourceId()` without knowing the type. You also can't conceptually reference just any resource type e.g., you need to calculate a connection string to AppInsights but someone selected a Key Vault resource. It just doesn't make sense *not* to require `resourceType` in this case.

That would be one problem with a general `filter` as I proposed above as well. So maybe we keep and require `resourceType`, but you could still have an optional `filter` that we'd combine e.g. if a `filter` were specified, we'd `and` it in parens: `resourceType eq "${resourceType}" and (${filter})`.

cli/azd/pkg/prompt/prompter.go
```diff
@@ -111,6 +112,47 @@ func (p *DefaultPrompter) PromptLocation(
...
+	choices := make([]string, len(resources))
+	for idx, resource := range resources {
+		// TODO: Get location display names from account manager instead?
+		choices[idx] = fmt.Sprintf("%d. %s (%s)", idx+1, resource.Name, resource.Location)
+	}
+
+	choice, err := p.console.Select(ctx, input.ConsoleOptions{
+		Message: msg,
+		Options: choices,
+	})
```
vhvb1989, line 148: I think we should consider/handle if `choices` is empty.  Would that mean:  The template I want to deploy requires me to have an existing resource of type `resourceType` but I don't have any in my Subscription.  Probably returning an error before the prompt might help. Like: `Template requires an existing resource of type <fooo> but there none was found in the subscription`.

heaths: Good point. I'll handle the requirement in the prompt instead of the prompter, leaving the prompter more generalized. The prompt error will have access to more information as well such as the parameter name (key).

wbreza:
Hey @heaths - i have a very similar PR out.  Looks like we both took a very similar route to resolved.
https://github.com/Azure/azure-dev/pull/4418

heaths:
@vhvb1989 you folks still interested in this one?

cli/azd/pkg/prompt/prompter.go
```diff
@@ -111,6 +112,51 @@ func (p *DefaultPrompter) PromptLocation(
...
+	options := azapi.ListResourcesOptions{
+		ResourceType: resourceType,
+	}
+	resources, err := p.resourceService.ListResources(ctx, p.env.GetSubscriptionId(), &options)
+	if err != nil {
+		return "", fmt.Errorf("listing resources: %w", err)
+	}
+	if len(resources) == 0 {
+		return "", nil
+	}
```
vhvb1989, line 136: I think this should return an error. Since the intention of the method is to prompt, the fact of an empty list is a blocker to do the prompting, so it should return an error to the caller and report that there's no resources to prompt for.

weikanglim: FWIW, I ended up leveraging what @heaths had here in #4943. 

While engaging more deeply with the code, I appreciated @heaths implementation here, since whether the resource type is optional is left up to the caller to decide. It allows for a shorter and user-friendly error message.

cli/azd/pkg/prompt/prompter.go
```diff
@@ -111,6 +112,51 @@ func (p *DefaultPrompter) PromptLocation(
...
+// if you know the `resourceType`.
+func (p *DefaultPrompter) PromptResource(
+	ctx context.Context,
+	msg string,
+	resourceType string,
+) (string, error) {
+	options := azapi.ListResourcesOptions{
+		ResourceType: resourceType,
+	}
+	resources, err := p.resourceService.ListResources(ctx, p.env.GetSubscriptionId(), &options)
```
weikanglim, line 130: As I was implementing #4943, I realized that this listing would list resources in the subscription, in which case returning just the resource `Name` isn't enough to fully qualify the resource ID. 

Returning resource ID is obviously more powerful, and still portable, but the Bicep referencing logic is rather messy. There isn't first-party functions that support this, so it ends up being a lot of string splitting with known indexing. Working code -- but extremely unappealing, see example [here](https://github.com/Azure/azure-dev/pull/4943/files#diff-d0d350d90c0df12cb1e6b807657c13203b90428ea24368b0ba0d0b9bcb58f03a).

As @JeffreyCA kindly noted, this may be simpler when Bicep supports dedicated function parsing for [resource IDs](https://github.com/Azure/bicep/issues/9838).

All this is to say that our hands are a little tied on making a nice little prompt experience here, and I'm honestly not sure which way we want to lean, but I did end up returning resource ID in #4943 because subscription-scope resources were a requirement there.

cli/azd/pkg/prompt/prompter.go
heaths: FWIW, my main thought with separating optional out was to also allow more permutations for other selector types. I've seen some APIs that have things like `foo`, `fooOptional`, `bar`, `barOptional`, etc., and the maintenance of and UX for all those is cumbersome. I like to separate "dials".

PR #4536, opened 2024-11-10
Add Rust language support

### LINKED ISSUE #423: Scenario "Rust dev": As a Rust dev, I want to dev Rust apps with azd

heaths:
Resolves #423


heaths:
This initial draft adds only language detection support but I wanted to push a "safety commit". We have some partners working on Rust projects so first-class support in `azd` would be nice.

weikanglim:
@heaths Really appreciate the PR here! I'm really happy with the direction here, so we'll continue to move forwards in this direction.

My Rust knowledge is circa 2022, and admittedly I didn't explore the build system, mostly language, so just some basic questions of the toolchain right now:

1. Does `cargo.toml` have a workspace correspondence?
2. For multiple binary crates, would each crate have a `cargo.toml`?

I'd also like take this opportunity to solicit any immediate feedback/concerns you may have around the current `appdetect` library in terms of Rust support -- just happy to hear from your perspective as well.

heaths:
Rust is case-sensitive, so `Cargo.toml` and, yes, you can have a workspace declare in one. See https://github.comAzure/azure-sdk-for-rust's root `Cargo.toml` for an example. You can also declare a crate and a workspace in a single one. See https://github.com/heaths/recorded-tests for an example.

Yes, each crate will have a `Cargo.toml`. As far as packaging goes, `cargo` - Rust's de facto toolchain - is akin to Node's `npm`. But as far as build and test go, `dotnet` is a better comparison. Thus, I'm sort of cribbing off JS's and .NET's code in `azd` as a blueprint.

PR #4541, opened 2024-11-12
compose: with existing infra

### LINKED ISSUE #4578: Compose: Iac Mode support for brownfield projects
Today compose supports template initialized using `azd init` with azd composability feature enabled. 

Customers should be able to run `azd add` for:

- [ ] any azd templates in AI App Gallary and Awesome azd
- [ ] azd template (infra) generated by CodeToCloud
- [ ] azd template with hand-coded infra

weikanglim:
This change allows for composability to work with an existing `infra` directory unmanaged by azd.

**Infra provisioning.** The provisioning engine has been updated to serially provision the root `./infra`, then `./infra/azd` if it exists. The following behavioral changes have been applied:

1. **Deployment order.** Sequential; the progress display shows the root deployment URL, then displays resources created as each deployment runs in the background.
2. **Outputs.** Outputs are written to the environment in the deployment order. This means in cases of output collisions, the last deployment wins.
3. **Preview.** Merge deployment resources into a single preview list, deduplicate resource ID + type.
4. **Destroy.** Merge deployment resources for display, and for purge. For destroy, each deployment is executed sequentially. This works well for stack deployments if resources are mutually exclusive. For standard deployments, this works since the second deletion on RG is a no-opt.
5. **State.** The deployments are fetched and applied to the environment in the same order as the deployment order. This shares the same last-write-win semantics.

**Composability synth.** The synthesized infra directory has a `.azd` file to serve as a sentinel. If `.azd` exists in a directory, i.e. `./infra/foo`, it means the directory `./infra/foo` is "azd managed". This used to achieve the following behavior when running `infra synth`:
- If `./infra/.azd` exists, we overwrite `./infra`.
- Otherwise, we write to `./infra/azd`.

## UX

Provision:
image

Destroy:
TBD

Fixes #4578

### DIFF
--- a/cli/azd/internal/vsrpc/environment_service_refresh.go
+++ b/cli/azd/internal/vsrpc/environment_service_refresh.go
@@ -88,7 +88,7 @@ func (s *environmentService) refreshEnvironmentAsync(
 
 	_ = observer.OnNext(ctx, newInfoProgressMessage("Loading latest deployment information"))
 
-	deployment, err := bicepProvider.LastDeployment(ctx)
+	deployment, err := bicepProvider.LastRootDeployment(ctx)
 	if err != nil {
 		log.Printf("failed to get latest deployment result: %v", err)
 	} else {
--- a/cli/azd/pkg/azapi/resource_service.go
+++ b/cli/azd/pkg/azapi/resource_service.go
@@ -2,8 +2,11 @@ package azapi
 
 import (
 	"context"
+	"errors"
 	"fmt"
+	"slices"
 
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/arm"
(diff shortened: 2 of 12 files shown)


ellismg:
> Composability synth. The synthesized infra directory has a .azure file to serve as a sentinel. If .azure exists in a directory, i.e. ./infra/foo, it means the directory ./infra/foo is "azd managed". This used to achieve the following behavior when running infra synth:

This is an interesting idea, @weikanglim.  I like having a sentinel file to help `azd` understand if it manages this directory or not.  I'm wondering if we'll want to pick a different file name than `.azure` (I'm just a little worried that it uses the same name as we use for our bookkeeping folder).

weikanglim:
@ellismg We can pick a different name, like `.azd` (what I had). I had made it gitignored to start with; I think actually it works better as a checked in source file.

weikanglim:
@ellismg Updated the UX, I do feel better about exposing the deployments here considering it's a backbone of the feature that is noteworthy. I also like where we're going here.

image

I will work with our UX friends for a final pass once everyone is back.

Issue #511, opened 2022-08-22
[Issue] azd env set adds a / character to value


greenie-msft:
- [x] Make sure you've installed the latest version using [instructions in the wiki](../wiki/install)

**Output from `azd version`**
Run `azd version` and copy and paste the output here:

azd version 0.1.0-beta.4 (commit fd96b3e9b283598bc4dc736e893f1b47080fbf7d)

**Output from `az version`** 
Run `az version` and copy and paste the output here (minimum required version is 2.38.0): 

azd version 0.1.0-beta.4 (commit fd96b3e9b283598bc4dc736e893f1b47080fbf7d)

**Describe the bug**
When running the CLI command azd env set, the value that's added to the .env file has a "\" character before a special character, such as a !

**To Reproduce**
Steps to reproduce the behavior...
Run the azd env set CLI command

**Expected behavior**
The azd env set command should take the input and add it to the .env as entered from the user.

jongio:
Wasn't it a trailing "\"?

greenie-msft:
@jongio actually when the env variable included a special character, for example !, it was adding a / before it.

savannahostrowski:
@greenie-msft Does this still repro?

rajeshkamal5050:
@v-xuto can you see if this is still reproducible?

Menghua1:
@rajeshkamal5050 This issue can be reproduced, and we started trying to fix it.

vhvb1989:
This is expected behavior.  Some special characters like `!` have specific meaning in terminals like `sh`
If the character is not escaped, it would cause issues when trying to parse it back or if users try to pipe it.

Closing this as expected bahavior.
Please re-open if this is blocking some specific use-case and share the use case.
I would prefer making azd to fail and reject the use of special chars like `!` as values for the .env (as that's probably a hint of a password stored as a value).

PR #4542, opened 2024-11-12
`azd env set` un-escapes special characters in values

Menghua1:
Fixes issue[#511](https://github.com/Azure/azure-dev/issues/511).

`azd env set` un-escapes special characters in the value, such as "!", "$", "`".

### DIFF
--- a/cli/azd/pkg/environment/environment.go
+++ b/cli/azd/pkg/environment/environment.go
@@ -20,7 +20,7 @@ import (
 	"github.com/azure/azure-dev/cli/azd/pkg/config"
 )
 
-// doubleQuoteSpecialChars defines the characters that need to be escaped.
+// DoubleQuoteSpecialChars defines the characters that need to be escaped.
 const doubleQuoteSpecialChars = "\\\n\r\""
 
 // EnvNameEnvVarName is the name of the key used to store the envname property in the environment.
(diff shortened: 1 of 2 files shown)


cli/azd/pkg/environment/environment.go
```diff
@@ -281,10 +286,42 @@ func fixupUnquotedDotenv(values map[string]string, dotenv string) string {
 // Instead of calling `godotenv.Write` directly, we need to save the file ourselves, so we can fixup any numeric values
 // that were incorrectly unquoted.
 func marshallDotEnv(env *Environment) (string, error) {
-	marshalled, err := godotenv.Marshal(env.dotenv)
+	marshalled, err := Marshal(env.dotenv)
```
wbreza, line 289: Please add unit tests to verify this functionality to check for well known uses cases as well as cases that fall into this special treatment.

Menghua1: Unit tests have been added, please check.

v-xuto: @wbreza We have updated by your comments, please re-review. Thanks.

v-xuto:
@wbreza We have updated by your comments, please re-review. Thanks.

rajeshkamal5050:
@vhvb1989 @hemarina can you review?

cli/azd/pkg/environment/environment.go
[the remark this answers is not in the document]
Menghua1: @hemarina The original `godotenv.Marshal()` calls [doubleQuoteEscape()](https://github.com/joho/godotenv/blob/v1.5.1/godotenv.go#L216-L228) to escape special characters in `godotenv.go`, including "!", "$", "`". And they are in the [doubleQuoteSpecialChars](https://github.com/joho/godotenv/blob/v1.5.1/godotenv.go#L27) constant.

As of now, we haven't found a replace function that doesn't escape these special characters. So we overwrote the `godotenv.Marshal()` method, these special characters have been removed from [doubleQuoteSpecialChars](https://github.com/Azure/azure-dev/blob/745abc521bf371f3e5432ffc726f1c81b677f671/cli/azd/pkg/environment/environment.go#L24) and are no longer escaped.

vhvb1989:
Closed the original issue as expected behavior.

Issue #289, opened 2022-06-20
`azd env set` does not check for case sensitivity


markweitzel:

savannahostrowski:
Is this still reproduceable?

rajeshkamal5050:
@v-xuto can you try to repro?

v-xuto:
@rajeshkamal5050 This issue can be reproduced, and we will try to fix it.

vhvb1989:
@v-xuto , before you start a PR for this, can you share what is the fix you would be doing?  Since there are 3 options on the issue, we need to make sure about including @ellismg and @kristenwomack for thumbs up :)

NanaXiong00:
@vhvb1989 We will choose the "force upper case" method to fix this issue.

vhvb1989:
@NanaXiong00 , in general .env files are case sensitive.  Developers might be familiar with this concept as it is well known in all languages when using .env files.  Maybe forcing upper case could make azd to work differently than the universal expectation.

rajeshkamal5050:
@vhvb1989 is there anything which needs to be addressed here? should we close this as wont-fix?

vhvb1989:
@rajeshkamal5050 , I thing the approach of `warn if the same value is present but in mixed case` would be enough here

kristenwomack:
> [@rajeshkamal5050](https://github.com/rajeshkamal5050) , I thing the approach of `warn if the same value is present but in mixed case` would be enough here

I like this solution the most 👆

v-xuto:
> [@rajeshkamal5050](https://github.com/rajeshkamal5050) , I thing the approach of `warn if the same value is present but in mixed case` would be enough here

Okay, thank you for your suggestions. We will use this approach to fix.

v-xuto:
@rajeshkamal5050, @vhvb1989 PR https://github.com/Azure/azure-dev/pull/4547 is ready, please help review.

weikanglim:
Just double checking on our UX requirements here with @vhvb1989 and @kristenwomack :

Would we want to print a warning message, and then ask the user to confirm?

In the current PR being proposed, we would print the warning message then continue -- I think this does meet the requirement of informing the user that this is happening, but it may be too late as the value is already saved. We don't have a straightforward `undo` command today, so I wonder if an upfront confirmation is better.

vhvb1989:
[...]

PR #4547, opened 2024-11-13
`azd env set` add warning

NanaXiong00:
Fixes issue https://github.com/Azure/azure-dev/issues/289.

`azd env set` adds warning about case sensitivity: warn if the same value is present but in mixed case. 
As shown in the figure below:
image


cli/azd/cmd/env.go
```diff
@@ -150,6 +153,24 @@ func (e *envSetAction) Run(ctx context.Context) (*actions.ActionResult, error) {
...
+		if existingKey == key {
+			continue
+		}
+
+		if strings.ToLower(existingKey) == lowerKey {
+			fmt.Printf(
+				"WARN: The environment variable '%v' already exists in the .env file with a different case.\n",
+				existingKey)
+			return
+		}
```
wbreza, line 170: Please ensure warning messages are consistent with other warnings in the system.  Messages should read "WARNING: [Message]" in Yellow by using our `output.WithWarningFormat(...)` helper function.

wbreza (approved):
Looks good - but please update warning style to be consistent with UX guidelines before merging.

cli/azd/cmd/env.go
NanaXiong00: @wbreza I have updated based on suggestions, please re-review.

cli/azd/cmd/env.go
```diff
@@ -150,6 +153,30 @@ func (e *envSetAction) Run(ctx context.Context) (*actions.ActionResult, error) {
...
+		if strings.ToLower(existingKey) == lowerKey {
+			conflictKeys = append(conflictKeys, fmt.Sprintf(`"%s"`, existingKey))
+		}
+	}
+
+	if len(conflictKeys) > 0 {
+		conflictKeysStr := strings.Join(conflictKeys, " and ")
+		fmt.Print(
+			output.WithWarningFormat(
+				"WARNING: The environment variable %s already exists in the .env file with a different case.\n",
```
weikanglim, line 175: I'd like to continue with the requirements discussion [here](https://github.com/Azure/azure-dev/issues/289#issuecomment-2477152902) before we finalize the changes.

vhvb1989: @weikanglim ping.

weikanglim: @vhvb1989 pushed a change, does this look good to you?

vhvb1989 (approved):

cli/azd/cmd/env.go
vhvb1989: merge merge merge

JeffreyCA: Maybe just me but the word "instead" is throwing me off a bit since it's printing the same name used in the original command. I think it makes more sense to either omit "instead" or change the phrasing to "Did you mean to set 'foo' instead":

image

Any thoughts?

weikanglim: It definitely should be reworded as `did you mean to set 'foo' instead?`

@vhvb1989 The multiple insensitive cases is still a weird situation :D it almost clearly signals user intention of wanting to set multiple different casings. Still feels as-if the confirmation would be a better approach

or we can just stop printing warning messages for n>1

v-xuto:
@vhvb1989 Does this PR require review from others? If not, could you please help to merge it? Thanks

Issue #4540, opened 2024-11-12
Clean up subscription level deployments


danieljurek:
`azd`'s CI tests can leave behind subscription-wide deployments (not associated with any resource groups). Something should be cleaning these up. 

Options: 
* azd cleans them up itself after `azd down` ... This prevents customers from running into the same issue that is created here
* engsys cleans up these deployments itself (ideally the deployment name is available somewhere so that it can be cleaned up directly, otherwise there's a risk that deployments could fall through) 

This issue is currently mitigated by the cleanup script now cleaning up subscription-wide deployments but we should not rely on that.

PR #4558, opened 2024-11-14
tests: Cleanup subscription deployments as part of tests

ellismg:


cli/azd/test/functional/aspire_test.go
```diff
@@ -240,6 +244,47 @@ func Test_CLI_Aspire_Deploy(t *testing.T) {
 	require.NoError(t, err)
 }
 
+// cleanupDeployments deletes all subscription level deployments tagged with `azd-env-name` equal to envName. If the session
+// indcates we are in playback mode, this function is a no-op.
+func cleanupDeployments(ctx context.Context, t *testing.T, azCLI *azdcli.CLI, session *recording.Session, envName string) {
```
ellismg, line 249: My strategy to figure out what tests needed to call this was to look for calls to `NewCLI` where we pass in a `*Session` and arrange for these tests to call `cleanupDeployments`.

ellismg:
Here's an internal run: https://dev.azure.com/azure-sdk/internal/_build/results?buildId=4335287&view=results which should exercise the delete codepath (we don't get coverage from our regular PR CI runs of this, since we use playback mode).

weikanglim:
@ellismg Just wondering about the direction here: is there value embracing the idea that automation tests try their best to simulate production behavior and avoid extra test-only deployment cleanup? For example, perhaps in the future, we may want the deployment being around as part of a production feature set.

I'm guessing that there is some ARM behavior here that is causing us to add the test-specific cleanup. My guess is that ARM is failing to keep up with deployment history purge due to the volume of deployments?

cli/azd/test/functional/aspire_test.go
danieljurek: The engsys cleanup script now also deletes sub-wide deployments. Thanks for looking for these. I couldn't find any others in my cursory searches... but my knowledge of the test suite is not as exhaustive as it was previously.

danieljurek:
> @ellismg Thanks for doing this! Just wondering about the direction here before approving: is there value embracing the idea that automation tests try their best to simulate production behavior and avoid extra test-only deployment cleanup? For example, perhaps in the future, we may want the deployment being around as part of a production feature set.
> 
> I'm guessing that there is some ARM behavior here that is causing us to add the test-specific cleanup. My guess is that ARM is failing to keep up with deployment history purge due to the volume of deployments?

Indeed, this change addresses an issue where ARM can fall behind on the volume of subscription-level deployments being created. Later sub-wide deployments fail in that scenario. This could conceivably happen in a customer's environment (so worth examining whether it should be done long term) but has most recently been observed happening in a test resources environment which sees a high volume of such deployments.

danieljurek:
Fixes https://github.com/Azure/azure-dev/issues/4540

danieljurek (approved):
LGTM.

danieljurek (changes requested):
The code looks good but I'm seeing a failing test in the live test run: https://dev.azure.com/azure-sdk/internal/_build/results?buildId=4335287&view=logs&j=3748a3ab-7269-5e9a-9cea-5ecb3b6f562c&t=9f3343b3-b8c5-51bf-d443-9aa044c5c18d&l=296 ... Perhaps the cleanup should be more resilient to deployments that aren't found? Or the deployment in `provisionRemoteStateStorage` is resource group specific?

danieljurek (approved):
Approved. Tests are passing.

Issue #4411, opened 2024-10-05
[Issue] `azd up -e <env>` does not run with the environment set explicitly


paulbatum:
- [x] Make sure you've installed the latest version using [instructions in the wiki](../wiki/install)

**Output from `azd version`**
azd version 1.10.1 (commit 31409a33266fb4a5fdbb644bc83988e725d6c7c9)

**Describe the bug**
If config.json refers to a default environment that no longer exists (lets call it foo), running `azd up --environment bar` will ignore the specified env and instead prompt you `Environment 'foo' does not exist, would you like to create it? (Y/n)`

**To Reproduce**
mkdir azdtest
cd azdtest
azd init -t hello-azd -e bar
azd env new foo
rd -r .\.azure\foo
azd up -e bar

**Expected behavior**
It runs on azd up on environment bar

**Actual behavior**
It displays the following prompt:
`Environment 'foo' does not exist, would you like to create it? (Y/n)`

rajeshkamal5050:
@v-xuto can you help triage this?

Menghua1:
@rajeshkamal5050 We were able to reproduce this issue and have created a PR https://github.com/Azure/azure-dev/pull/4561 to fix it, please review.

Menghua1:
@vhvb1989 For this issue, we have the following two points to discuss with you:

1. The reproduction steps for this issue are incorrect:

    Firstly, we think the current operation is incorrect. To create and deploy resources in the `bar` environment, the command `azd env select bar` should be used to switch the environment to `bar`, and then `azd up` should be run to create and deploy resources in that environment.

2. Design issues with `azd up -e``:

    We noticed that even when specifying `-e bar` with the command `azd up -e bar`, the resources are created and deployed in the default environment `foo` instead of the specified `bar` environment.
    
    Is this behavior expected?
    
    If not, when `-e bar` is specified, should it create and deploy resources in the `bar` environment, and should it also change the default environment to `bar`?

Which one do you prefer?

vhvb1989:
> We noticed that even when specifying -e bar with the command azd up -e bar, the resources are created and deployed in the default environment foo instead of the specified bar environment.

[...]

PR #4561, opened 2024-11-15
Fix: Ensure `azd up -e` deploys resources to the specified environment, ignoring the default environment

Menghua1:
Fix issue[#4411](https://github.com/Azure/azure-dev/issues/4411).


cli/azd/cmd/up.go
```diff
@@ -130,6 +130,14 @@ func (u *upAction) Run(ctx context.Context) (*actions.ActionResult, error) {
 		u.console.Message(ctx, output.WithGrayFormat("Note: Running custom 'up' workflow from azure.yaml"))
 	}
 
+	envValue := u.flags.EnvFlag.EnvironmentName
+	for _, step := range upWorkflow.Steps {
+		if envValue != "" {
+			stepArgs := append(step.AzdCommand.Args, "-e", envValue)
+			step.AzdCommand.Args = stepArgs
+		}
+	}
```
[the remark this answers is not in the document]
Menghua1: @vhvb1989 Thanks for your feedback!

When debugging the `azd up -e <environment>` command, we found that the actual executed steps (such as `azd package`, `azd provision`, and `azd deploy`) did not include the `-e` flag, which caused these commands to use the default environment instead of the environment specified by the user through `-e`. To ensure that these subcommands can correctly use the environment specified by `-e`, I manually set it in the code.

If you think there is a better solution, you are welcome to discuss further in the issue https://github.com/Azure/azure-dev/issues/4411#issuecomment-2482358893!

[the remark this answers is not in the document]
weikanglim: The current solution wouldn't work if the user added the workflow step to reference an azd command not accepting `-e`. An immediate solution doesn't come to mind, but will require some additional team discussion.

cli/azd/pkg/environment/manager.go
```diff
@@ -177,6 +177,17 @@ func (m *manager) LoadOrInitInteractive(ctx context.Context, environmentName str
 		}
 	}
 
+	defaultEnvironmentName, err := m.azdContext.GetDefaultEnvironmentName()
+	if err != nil {
+		return nil, fmt.Errorf("getting default environment: %w", err)
+	}
+
+	if env.Name() != defaultEnvironmentName {
+		if err := m.azdContext.SetProjectState(azdcontext.ProjectState{DefaultEnvironment: env.Name()}); err != nil {
```
vhvb1989, line 186: I actually like this approach. The side effect is that running one command with the `-e foo` flag makes the `foo` environment the default one, which also means that the last environment you use is always now the default environment.

That's probably a breaking change for someone running something like
azd env new foo
azd env ser a 1
azd provision -e bar
azd provision // expecting to use environment foo but `bar` will be used instead as the default

Menghua1: > That's probably a breaking change for someone running something like
azd env new foo
azd env ser a 1
azd provision -e bar
azd provision // expecting to use environment foo but `bar` will be used instead as the default

@vhvb1989 According to the azd logic, when executing the `azd provision -e bar` command,

1. If the `bar` environment does not exist, azd sets it as the default environment, and **the default environment is `bar`**.

2. If the `bar` environment already exists and is not the default environment, azd will not switch the default environment, and **the default environment is `foo`**.

The processing logic in these two cases is inconsistent, which may cause confusion to users.

Therefore, I think we need to adjust the logic of azd to ensure the consistency of its behavior. Regardless of whether the environment specified by the user already exists, as long as the user executes a command with the `-e` flag, azd should follow a consistent behavior: either do not change the default environment, or set the specified environment as the default environment in any case.

vhvb1989: You are right. Whenever azd creates a new environment, either with `azd env new` or from `azd provision -e new-env`, that environment is set as the default environment. Create new environment is expected to have this effect.

In all other cases, the default environment should not be changed.

It's definitely somehow confusing and probably azd should ask/confirm before setting a recent created environment as the new default. https://github.com/Azure/azure-dev/issues/4800

Menghua1: @vhvb1989 I have updated the PR code  and submitted a new solution, as follows:

Added a parameter `ParentEnvironmentName` to `envManager.azdContext`, when running `azd up -e bar`, the value of ParentEnvironmentName is set to `bar`. Then, when the environment is loaded, check if `parentEnvironmentName` is not empty, and if not, use it instead of the default environment. In this way, even if the subcommand does not explicitly specify the `-e` parameter, it will be executed in the environment specified by the parent command.

vhvb1989 (approved):
LGTM

cli/azd/pkg/environment/manager.go
vhvb1989: I like it. Thank you.

Menghua1:
>  I have updated the PR code and submitted a new solution, as follows:
>
> Added a parameter `ParentEnvironmentName` to `envManager.azdContext`, when running `azd up -e bar`, the value of ParentEnvironmentName is set to `bar`. Then, when the environment is loaded, check if `parentEnvironmentName` is not empty, and if not, use it instead of the default environment. In this way, even if the subcommand does not explicitly specify the `-e` parameter, it will be executed in the environment specified by the parent command.


@weikanglim and @wbreza Can you help review this PR?

PR #4565, opened 2024-11-15
 ci: Update bicep in lint-bicep to v0.31.92

ellismg:
The current version that is installed on the runners (0.31.34) has some bugs around linting and causes lint warnings during our legs. These errors are not present on the latest Bicep CLI.
    
Force an upgrade of the binary that is installed as part of the test runner to latest version. We should be able to remove this step once new image is built with an upgraded bicep (the `README.md` at https://github.com/actions/runner-images/tree/main/images/ubuntu has information about what is on the runner.)

### DIFF
--- a/.github/workflows/lint-bicep.yml
+++ b/.github/workflows/lint-bicep.yml
@@ -14,6 +14,11 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
+      - name: Upgrade bicep
+        run: |
+          which bicep
+          sudo curl -o $(which bicep) -L https://github.com/Azure/bicep/releases/download/v0.31.92/bicep-linux-x64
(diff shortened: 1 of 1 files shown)


templates/todo/projects/java-mongo-aca/.repo/bicep/infra/main.bicep
```diff
@@ -40,9 +40,6 @@ param apimSku string = 'Consumption'
 @description('Id of the user or app to assign application roles')
 param principalId string = ''
 
-@description('The base URL used by the web service for sending API requests')
-param webApiBaseUrl string = ''
```
ellismg, line 44: The linter was complaining that this parameter was unused, which is true.  So I removed it from here and from the associated `main.parameters.json` file.

weikanglim (approved):

Issue #393, opened 2022-05-09
[Issue] Test if docker is running before calling docker commands


jongio:
`azd up`

Got this:

```
Error: Command Failed: deploying service jongtodo100api: packaging service jongtodo100api: building container image: jongtodo100api at src/api: building container image: exit status 1 : exit code: 1, stdout: , stderr: error during connect: In the default daemon configuration on Windows, the docker client must be run with elevated privileges to connect.: Post "http://%2F%2F.%2Fpipe%2Fdocker_engine/v1.24/build?buildargs=%7B%7D&cachefrom=%5B%5D&cgroupparent=&cpuperiod=0&cpuquota=0&cpusetcpus=&cpusetmems=&cpushares=0&dockerfile=Dockerfile&labels=%7B%7D&memory=0&memswap=0&networkmode=default&platform=amd64&rm=1&shmsize=0&t=jongtodo100api&target=&ulimits=null&version=1": open //./pipe/docker_engine: The system cannot find the file specified.
```

Turns out docker wasn't running

I would like azd to tell me it isn't running before it tries to execute the command.

savannahostrowski:
This came up a bunch in our UXR for simplified init. We should investigate this as polish during Germanium

isaacrlevin:
I am learning Go and wanted to take a look at a fully fledged app, so I came across this help-wanted issue. Looking at this it seems like the `EnsureInstalled()` function needs to be updated. The function checks to see if a tool is installed, and if it is, it passes the check. This won't work for Docker as it can be installed but not running. Here is the line in question that needs to be updated.

[for _, tool := range tools {](https://github.com/Azure/azure-dev/blob/7990e2111872c524d12f3445a65ff83107fa259a/cli/azd/pkg/tools/ensure.go#L44-L49)
```go
for _, tool := range tools {
    _, ok := confirmedTools[tool.Name()]
    if ok {
	log.Printf("Skipping install check for '%s'. It was previously confirmed.", tool.Name())
	continue
}
```

Ideally there needs to be a check in this area to check if docker is running (maybe docker info). One thing that should be added to the `ExternalTool` interface to flag tools that need to be checked and that calls an alternate workflow (possibly an `EnsureRunning()` function)

isaacrlevin:
@jongio what sample did you get this error with? I am not able to replicate using Dev Containers. Not sure if that is expected however

rajeshkamal5050:
@v-xuto can you see if this is reproducible?

[...]

PR #4567, opened 2024-11-19
Fix azd issue #393

v-weiyding:
Fix [#393](https://github.com/Azure/azure-dev/issues/393).

Add code to the `CheckInstalled` function in the `docker.go` file to check if Docker is running.
The latest error information is shown below:
image


cli/azd/pkg/tools/docker/docker.go
```diff
@@ -263,6 +263,10 @@ func (d *Cli) CheckInstalled(ctx context.Context) error {
 	if !supported {
 		return &tools.ErrSemver{ToolName: d.Name(), VersionInfo: d.versionInfo()}
 	}
+	// Check if docker is running
+	if _, err := tools.ExecuteCommand(ctx, d.commandRunner, "docker", "ps"); err != nil {
+		return err
+	}
```
wbreza, line 269: We should return a better error message here.  The one you get does not tell them to check if their docker daemon is running.

v-weiyding: @wbreza, I have updated by your comments, please re-review.

cli/azd/pkg/tools/docker/docker.go
```diff
@@ -263,6 +263,10 @@ func (d *Cli) CheckInstalled(ctx context.Context) error {
 	if !supported {
 		return &tools.ErrSemver{ToolName: d.Name(), VersionInfo: d.versionInfo()}
 	}
+	// Check if docker daemon is running
+	if _, err := tools.ExecuteCommand(ctx, d.commandRunner, "docker", "ps"); err != nil {
+		return fmt.Errorf("the Docker daemon is not running, please start the Docker service: %w", err)
```
vhvb1989, line 268: From line L256, it looks like we use `d.Name()` instead of using `Docker` label directly.

This is an easy way to keep the label defined in just one place.  
Please define a local variable before line 256 like `toolName := d.Name()` and then use `toolName` in this error.

vhvb1989 (approved):
small fix requested. otherwise LGTM

cli/azd/pkg/tools/docker/docker.go
v-weiyding: @vhvb1989 , I have updated by your comments, please re-review.

vhvb1989 (approved):

PR #4573, opened 2024-11-21
dotnet: Support `dotnet publish` to produce container image

### LINKED ISSUE #2632: Support .NET Native Container Image build
As of .NET 7, .NET developers can [build their apps into containers natively](https://devblogs.microsoft.com/dotnet/announcing-builtin-container-support-for-the-dotnet-sdk/). It would be great if, when using AZD in the absence of Docker, that an image could be built natively. 

I'm not sure an ACR push could be implemented without Docker on the machine, but, if there's a way AZD could support .NET's native container features, it'd be great.

ellismg:
The `dotnet` tool has native support for producing (and pushing) a container image. This does not require a local docker daemon and in general "does the right thing" and is the prefered way for .NET customers to produce container
images. https://learn.microsoft.com/dotnet/core/docker/publish-as-container gives a good overview of how this support works and we've been using it to build and push container images for Aspire apps.

This change updates things such that we can use this support when building a non Aspire based .NET app. If a `Dockerfile` exists, we respect it, but otherwise instead of trying to use Oryx to produce a container image, we will use `dotnet publish`.

A new recorded test was added - it builds on top of the existing `containerapp` sample, and the test aranges to remove the `Dockerfile` file from the template before running `up`.

Fixes #2632

### DIFF
--- a/cli/azd/pkg/project/container_helper.go
+++ b/cli/azd/pkg/project/container_helper.go
@@ -22,6 +22,7 @@ import (
 	"github.com/azure/azure-dev/cli/azd/pkg/tools"
 	"github.com/azure/azure-dev/cli/azd/pkg/tools/azcli"
 	"github.com/azure/azure-dev/cli/azd/pkg/tools/docker"
+	"github.com/azure/azure-dev/cli/azd/pkg/tools/dotnet"
 	"github.com/benbjohnson/clock"
 	"github.com/sethvargo/go-retry"
 )
@@ -32,6 +33,7 @@ type ContainerHelper struct {
 	remoteBuildManager       *containerregistry.RemoteBuildManager
 	containerRegistryService azcli.ContainerRegistryService
 	docker                   *docker.Cli
+	dotNetCli                *dotnet.Cli
 	clock                    clock.Clock
 	console                  input.Console
 	cloud                    *cloud.Cloud
@@ -44,6 +46,7 @@ func NewContainerHelper(
 	containerRegistryService azcli.ContainerRegistryService,
 	remoteBuildManager *containerregistry.RemoteBuildManager,
 	docker *docker.Cli,
+	dotNetCli *dotnet.Cli,
 	console input.Console,
 	cloud *cloud.Cloud,
 ) *ContainerHelper {
@@ -53,6 +56,7 @@ func NewContainerHelper(
 		remoteBuildManager:       remoteBuildManager,
 		containerRegistryService: containerRegistryService,
 		docker:                   docker,
+		dotNetCli:                dotNetCli,
 		clock:                    clock,
 		console:                  console,
 		cloud:                    cloud,
@@ -194,6 +198,10 @@ func (ch *ContainerHelper) RequiredExternalTools(ctx context.Context, serviceCon
 		return []tools.ExternalTool{}
 	}
 
+	if useDotnetPublishForDockerBuild(serviceConfig) {
+		return []tools.ExternalTool{ch.dotNetCli}
+	}
+
 	return []tools.ExternalTool{ch.docker}
 }
 
@@ -270,6 +278,8 @@ func (ch *ContainerHelper) Deploy(
 
 	if serviceConfig.Docker.RemoteBuild {
 		remoteImage, err = ch.runRemoteBuild(ctx, serviceConfig, targetResource, progress)
+	} else if useDotnetPublishForDockerBuild(serviceConfig) {
+		remoteImage, err = ch.runDotnetPublish(ctx, serviceConfig, targetResource, progress)
 	} else {
 		remoteImage, err = ch.runLocalBuild(ctx, serviceConfig, packageOutput, progress)
 	}
@@ -382,7 +392,8 @@ func (ch *ContainerHelper) runLocalBuild(
 	return remoteImage, nil
 }
 
-// runLocalBuild builds the image using a remote azure container registry and tags it. It returns the full remote image name.
+// runRemoteBuild builds the image using a remote azure container registry and tags it.
+// It returns the full remote image name.
 func (ch *ContainerHelper) runRemoteBuild(
 	ctx context.Context,
 	serviceConfig *ServiceConfig,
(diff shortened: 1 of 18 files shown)


cli/azd/test/recording/sanitize.go
```diff
@@ -56,8 +56,8 @@ func sanitizeContainerAppListSecrets(i *cassette.Interaction) error {
 					// Redis requirepass. Sanitize the password, remove other config.
 					if strings.Contains(val, "requirepass ") {
 						body.Value[i].Value = to.Ptr("requirepass SANITIZED")
+						continue
```
ellismg, line 59: @weikanglim - I think this was subtly wrong before and the `continue` to was preventing us from sanitizing things unrelated to the `requirepass` entry. I think the code now matches the intention of the comment but I wanted to bring this to your attention.

cli/azd/test/functional/testdata/samples/containerapp/src/dotnet/Dockerfile
```diff
@@ -12,7 +12,7 @@ RUN dotnet publish -c Release -o out
 FROM mcr.microsoft.com/dotnet/aspnet:8.0
 WORKDIR /app
 COPY --from=build-env /app/out .
-EXPOSE 3100
-ENV ASPNETCORE_URLS=http://+:3100
+EXPOSE 8080
+ENV ASPNETCORE_URLS=http://+:8080
```
ellismg, line 16: Adjusting things in the Dockerfile so that the external port is the same as the one used by `dotnet publish` which allows us to re-use the IaC files between the `docker build` and `dotnet publish` flavors of the test.

cli/azd/test/recording/sanitize.go
ellismg: I had to make this change because the end to end tests ends up fetching the container secrets as we try to do our PATCH style update of the container apps image and the list of secrets included the registry password.

cli/azd/pkg/project/framework_service_docker.go
```diff
@@ -341,13 +338,40 @@ func (p *dockerProject) Build(
 	}, nil
 }
 
+func useDotnetPublishForDockerBuild(serviceConfig *ServiceConfig) bool {
+	if serviceConfig.useDotNetPublishForDockerBuild != nil {
```
ellismg, line 342: It felt like we were calling this a lot and caching the value (to save the stat calls) was worthwhile - but I could be convinced otherwise.  I didn't have anything scientific here - I just had a breakpoint in this method while I was developing things and felt like it was called many times during the `up` flow.

cli/azd/test/recording/sanitize.go
weikanglim: This is right; there was a change in a previous PR for this fix that was reverted. Just for reference: https://github.com/Azure/azure-dev/commit/e4a70d0e5141de8c2823ae88bc053a311ef3d7c1#diff-4e16e271f78e59f4ae2cebd907f7c53aeb186b6a2f3a0adf8fec9de1664cd62a

wbreza (approved):
Looks good - just a couple questions to consider before merging.

cli/azd/pkg/project/framework_service_docker.go
```diff
@@ -341,13 +338,40 @@ func (p *dockerProject) Build(
...
+		projectPath := serviceConfig.Path()
+
+		var dockerFilePath string
+
+		s, err := os.Stat(projectPath)
+		if err == nil && s.IsDir() {
+			dockerFilePath = filepath.Join(projectPath, "Dockerfile")
+		} else {
+			dockerFilePath = filepath.Join(filepath.Dir(projectPath), "Dockerfile")
+		}
```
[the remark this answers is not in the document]
ellismg: Good idea! - I'll add that and make a note about how you could use that to control what `Dockerfile` was used if you wanted to use docker but with a non-default location.

cli/azd/pkg/project/service_config.go
```diff
@@ -45,6 +45,9 @@ type ServiceConfig struct {
 	DotNetContainerApp *DotNetContainerAppOptions `yaml:"-,omitempty"`
 	// Custom configuration for the service target
 	Config map[string]any `yaml:"config,omitempty"`
+	// Computed lazily by useDotnetPublishForDockerBuild and cached. This is true when the project
+	// is a dotnet project and there is not an explicit Dockerfile in the project directory.
+	useDotNetPublishForDockerBuild *bool
```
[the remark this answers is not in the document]
ellismg: The function `useDotnetPublishForDockerBuild` (which is just a package level function in the `project` package) sets it.

Looking at the code and the callers of it - I think I might be able to arrange things such instead of storing the value here we store it in some cache managed by `ContainerHelper` (likely using the `Path` or something as the cache key).  Do you think that's a better strategy, @wbreza?  FWIW, I did try to call out what was going on here in the comment, but perhaps I should have put `useDotNetPublishForDockerBuild` in square brackets like `[useDotNetPublishForDockerBuild]` so it stood out in the comment a bit more.

cli/azd/pkg/project/container_helper.go
```diff
@@ -473,6 +484,41 @@ func (ch *ContainerHelper) runRemoteBuild(
...
+		return "", fmt.Errorf("logging in to registry: %w", err)
+	}
+
+	progress.SetProgress(NewServiceProgress("Publishing container image"))
+
+	imageName := fmt.Sprintf("%s:%s",
+		ch.DefaultImageName(serviceConfig),
+		ch.DefaultImageTag())
+
+	_, err = ch.dotNetCli.PublishContainer(
```
weikanglim, line 507: It's really difficult to see the forest for the trees right now given the multiple, multiple layers we're stacking on top of here, and our azd product stretch from compile-time to runtime --

But I think this does end up changing the default listening port from 80 to 8080 in the final image. Maybe we could consider updating the generated infra assets from easyInit/compose with this idea in mind for .NET languages:

- add_configure_host.go, L240
- infra_confirm.go, L215

I was reminded of this by a very timely community issue feedback #4583

cli/azd/test/functional/testdata/samples/containerapp/src/dotnet/Dockerfile
weikanglim: Nice! I think this also helped me orient myself around the default PORT change

weikanglim (approved):
Love it! It makes the dockerfile-less builds experience for dotnet developers much better than the current status quo.

Just left a suggestion on if we wanted to update the default ports for generated infra-assets based on this change.

cli/azd/pkg/project/service_config.go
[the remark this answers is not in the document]
weikanglim: this looks good to me

ellismg:
> Just left a suggestion on if we wanted to update the default ports for generated infra-assets based on this change.

Good call out - I really hadn't considered the end-to-end flow for a non-aspire based .NET app via our easy-init stuff.

I'll make the relevant changes and then play around with the `azd init --from-code` experience on my sample application from the tests to make sure it feels good.

cli/azd/pkg/project/container_helper.go
ellismg: I did this and verified that the in the easy-init flow things worked as expected.  I assume they will also work for the compose flow where `infra` is not generated, but I didn't explicitly test it.

It made me smile to see the whole thing work end to end, and it was nice how little I had to change to make it all work!

cli/azd/pkg/project/framework_service_docker.go
```diff
@@ -199,7 +196,7 @@ func (p *dockerProject) Build(
 	restoreOutput *ServiceRestoreResult,
 	progress *async.Progress[ServiceProgress],
 ) (*ServiceBuildResult, error) {
-	if serviceConfig.Docker.RemoteBuild {
+	if serviceConfig.Docker.RemoteBuild || useDotnetPublishForDockerBuild(serviceConfig) {
```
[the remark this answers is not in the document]
ellismg: I think so as well (we have the same deficiency in the Aspire case as well now).  I'm not sure exactly what we want `azd package` to do long term for containers.  Today it is something like: "arrange for a local image to be built and tagged with the remote image name" but the "artifact" is just the remote image tag - it relies on the content being built and cached and tagged locally.

We could have `dotnet publish` build and tag a local image (it would require the customer have Docker installed) or we could exploit the fact that newer SDKs can now produce a tgz (i.e what you get from `docker save`) but then the problem becomes pushing it to the remote registry now needs the docker toolchain.

My long term thought is that we will want to end up in a place where `azd package` produces the `tgz` as `docker save` would and then we have a way for `deploy --from-package` to push that image, ideally without requiring `docker` (we should be able to pull in some of the OCI SDK and use that to push to a registry ourselves).

I'll file something for the backlog about a reasonable semantics for `azd package` for container based apps.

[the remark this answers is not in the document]
ellismg: https://github.com/Azure/azure-dev/issues/4584

PR #4585, opened 2024-11-22
Show full Entra error description on auth failure

### LINKED ISSUE #4559: Improve `azd pipeline config` error message when it fails in a Codespace due to Conditional Access Policy

**Output from `azd version`**
azd version 1.11.0 (commit 5b92e0687e1fa96dfc8292f4b900c0c58610b6a5)

**Describe the bug**
When running `azd pipeline config` on an `azd` template inside of a GitHub Codespace, if I originally authenticated using `azd auth login`, it can fail while trying to create the service principal:

Image

The error message says to run `azd auth login --scope https://graph.microsoft.com/.default` to reauthenticate, but that fails with:
Image

The solution to this issue was to follow the steps in https://learn.microsoft.com/en-us/azure/developer/azure-developer-cli/troubleshoot?tabs=Browser#azd-pipeline-config-failure-due-to-conditional-access-policy to instead authenticate using `azd auth login --use-device-code=false`, which isn't clear from the original error message.

Running `azd pipeline config --debug 2> debug.log` in debug mode reveals the following error:

```
2024/11/14 23:26:54 main.go:50: Retry: =====> Try=1 for GET https://graph.microsoft.com/v1.0/applications(appId='az-dev-11-14-2024-23-26-54')
2024/11/14 23:26:54 azd_credential.go:37: POST https://login.microsoftonline.com/72f988bf-86f1-41af-91ab-2d7cd011db47/oauth2/v2.0/token
--------------------------------------------------------------------------------
RESPONSE 400: 400 Bad Request
--------------------------------------------------------------------------------
{
  "error": "invalid_grant",
  "error_description": "AADSTS50005: User tried to log in to a device from a platform (Unknown) that's currently not supported through Conditional Access policy. Supported device platforms are: iOS, Android, Mac, and Windows flavors. Trace ID: cba361ee-c55a-4a28-a4ed-84f5d5143600 Correlation ID: 6af12d70-5003-4f3a-b6ac-8237166bea1c Timestamp: 2024-11-14 23:26:54Z",
  "error_codes": [
    50005
  ],
  "timestamp": "2024-11-14 23:26:54Z",
  "trace_id": "cba361ee-c55a-4a28-a4ed-84f5d5143600",
  "correlation_id": "6af12d70-5003-4f3a-b6ac-8237166bea1c"
}
--------------------------------------------------------------------------------
```

**To Reproduce**
1. Open a template in a GitHub Codespace (e.g. https://codespaces.new/Azure-Samples/openai-chat-vision-quickstart)
2. Run `azd auth login`
3. Run `azd pipeline config`

**Expected behavior**
It may be helpful to surface the underlying error message (`AADSTS50005: User...`) to the output without needing `--debug`.

The current error message could also be improved to maybe suggest doing `azd auth login --use-device-code=false` or point to that troubleshooting article https://learn.microsoft.com/en-us/azure/developer/azure-developer-cli/troubleshoot?tabs=Browser#azd-pipeline-config-failure-due-to-conditional-access-policy

**Environment**
Information on your environment: GitHub Codespace

**Additional context**
@weikanglim, feel free to add your thoughts 🙂

Related: #2184

JeffreyCA:
Fixes #4559 so that the full Entra error description is shown to the user when there's an auth failure. In addition, the instruction text to reauthenticate is now wrapped in a suggestion.

**Existing behaviour:**
image

**New behaviour:**
image

### DIFF
--- a/cli/azd/pkg/auth/errors.go
+++ b/cli/azd/pkg/auth/errors.go
@@ -77,8 +77,8 @@ func (e *ReLoginRequiredError) Error() string {
 }
 
 func (e *ReLoginRequiredError) ErrorWithSuggestion() internal.ErrorWithSuggestion {
-	return internal.ErrorWithSuggestion {
-		Err: e,
+	return internal.ErrorWithSuggestion{
+		Err:        e,
 		Suggestion: fmt.Sprintf("Suggestion: %s, run `%s` to acquire a new token", e.scenario, e.loginCmd),
 	}
 }
--- a/cli/azd/pkg/auth/errors.go
+++ b/cli/azd/pkg/auth/errors.go
@@ -29,7 +29,7 @@ type ReLoginRequiredError struct {
 	// The scenario in which the login is required
 	scenario string
 
-	description string
+	errText string
 
 	helpLink string
 }
@@ -66,7 +66,7 @@ func newReLoginRequiredError(
 }
 
 func (e *ReLoginRequiredError) init(response *AadErrorResponse, scopes []string, cloud *cloud.Cloud) {
-	e.description = response.ErrorDescription
+	e.errText = response.ErrorDescription
 	e.scenario = "reauthentication required"
 	e.loginCmd = "azd auth login"
 	if !matchesLoginScopes(scopes, cloud) { // if matching default login scopes, no scopes need to be specified
@@ -81,7 +81,7 @@ func (e *ReLoginRequiredError) init(response *AadErrorResponse, scopes []string,
 	}
 
 	// User tried to sign in to a device from a platform not currently supported through Conditional Access policy
-	if slices.Contains((response.ErrorCodes), 50005) {
+	if slices.Contains(response.ErrorCodes, 50005) {
 		e.loginCmd += " --use-device-code=false"
 		// TODO: Use aka.ms short link
 		//nolint:lll
@@ -90,7 +90,7 @@ func (e *ReLoginRequiredError) init(response *AadErrorResponse, scopes []string,
 }
 
 func (e *ReLoginRequiredError) Error() string {
-	return e.description
+	return e.errText
 }
 
 // matchesLoginScopes checks if the elements contained in the slice match the scopes acquired during login.
--- a/cli/azd/pkg/auth/errors.go
+++ b/cli/azd/pkg/auth/errors.go
@@ -80,12 +80,11 @@ func (e *ReLoginRequiredError) init(response *AadErrorResponse, scopes []string,
 		e.scenario = "login expired"
 	}
 
-	// User tried to sign in to a device from a platform not currently supported through Conditional Access policy
+	// In a Codespaces environment, `azd auth login` defaults to device code flow, which can cause issues getting tokens if the
+	// the Entra tenant has Conditional Access Policies set.
 	if slices.Contains(response.ErrorCodes, 50005) {
 		e.loginCmd += " --use-device-code=false"
-		// TODO: Use aka.ms short link
-		//nolint:lll
-		e.helpLink = "https://learn.microsoft.com/azure/developer/azure-developer-cli/troubleshoot#azd-pipeline-config-failure-due-to-conditional-access-policy"
+		e.helpLink = "https://aka.ms/azd/troubleshoot/conditional-access-policy"
 	}
 }
 
--- a/cli/azd/pkg/auth/errors.go
+++ b/cli/azd/pkg/auth/errors.go
@@ -30,6 +30,8 @@ type ReLoginRequiredError struct {
 	scenario string
 
 	description string
+
+	helpLink string
 }
 
 // newReLoginRequiredError returns an error if the response indicates that the user needs to reauthenticate.
@@ -50,9 +52,13 @@ func newReLoginRequiredError(
 		"interaction_required":
 		err := ReLoginRequiredError{}
 		err.init(response, scopes, cloud)
+		suggestion := fmt.Sprintf("Suggestion: %s, run `%s` to acquire a new token.", err.scenario, err.loginCmd)
+		if err.helpLink != "" {
+			suggestion += fmt.Sprintf(" See %s for more info.", err.helpLink)
+		}
 		return &internal.ErrorWithSuggestion{
 			Err:        &err,
-			Suggestion: fmt.Sprintf("Suggestion: %s, run `%s` to acquire a new token", err.scenario, err.loginCmd),
+			Suggestion: suggestion,
 		}, true
 	}
 
@@ -69,8 +75,16 @@ func (e *ReLoginRequiredError) init(response *AadErrorResponse, scopes []string,
 		}
 	}
 
+	// BadTokenDueToSignInFrequency - The refresh token has expired or is invalid due to sign-in frequency checks by Conditional Access.
 	if slices.Contains(response.ErrorCodes, 70043) {
 		e.scenario = "login expired"
+	}
+
+	// DevicePolicyError - User tried to sign in to a device from a platform not currently supported through Conditional Access policy
+	if slices.Contains((response.ErrorCodes), 50005) {
+		e.loginCmd += " --use-device-code=false"
+		// TODO: Use aka.ms short link
+		e.helpLink = "https://learn.microsoft.com/azure/developer/azure-developer-cli/troubleshoot#azd-pipeline-config-failure-due-to-conditional-access-policy"
 	}
 }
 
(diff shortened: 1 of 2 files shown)


cli/azd/pkg/auth/errors.go
```diff
@@ -47,13 +50,15 @@ func newReLoginRequiredError(
 		"interaction_required":
 		err := ReLoginRequiredError{}
 		err.init(response, scopes, cloud)
-		return &err, true
+		errWithSuggestion := err.ErrorWithSuggestion()
```
vhvb1989, line 53: Can we create `error with suggestion` w/o depending on creating a `ReLoginRequiredError` first?

Or, can we just make the `Error()` default message from `ReLoginRequiredError` to contain the suggestion?  Maybe we don't need to keep the `Error()` without a suggestion, right?

ellismg: I think I like separating the message from the server (i.e the `response.ErrorDescription` becomes the string returned by `Error()`) from our suggestion as @JeffreyCA has done here.

But I also agree that creating the `*internal.ErrorWithSuggestion` directly makes more sense.  Since `newReLoginRequiredError` is typed as returning `error` and not something like `*ReLoginRequiredError`, I think what I would have done is something like this:

```go
err := ReLoginRequiredError{}
err.init(response, scopes, cloud)
	Err: err,
	Suggestion: fmt.Sprintf("Suggestion: %s, run `%s` to acquire a new token", err.scenario, err.loginCmd)
}, nil
```

And not added `ErrorWithSuggestion()` to `*ReLoginRequiredError`

FWIW (and this isn't feedback that needs to be addressed with this PR), with a fresh set of eyes, I'm feeling a little strange about the design of `newReLoginRequiredError` - it's sort of surprising to me that we have this `init` pattern (which is only called from `newReLoginRequiredError`) and it's also a little strange that `newReLoginRequiredError` works the way it does (it can return nil, false) when the `*AadErrorResponse` does indicate an error due to needing to login, and the caller needs to handle that.

ellismg:
Thanks for sharing the screen shots, @JeffreyCA!

One of the things that jumped out at me with them (which I don't think is a problem introduced by this PR) is that the suggestion text is sort of hard to find, especially with the large amount of red text before.

In a future PR we might want to introduce some whitespace (and maybe some color?) to help it pop a little more - just my two cents.

cli/azd/pkg/auth/errors.go
JeffreyCA: Thanks for the comments - I pushed a change to inline it instead of the separate `ErrorWithSuggestion` function.
From what I understand, there seems to be a fallback to handle the error as a `AuthFailedError` if `nil, false` is returned:

https://github.com/Azure/azure-dev/blob/84b605c64f875ce444913a5d965c2d6e6d6ccbb1/cli/azd/pkg/auth/azd_credential.go#L33-L45

I'm not sure if this addresses your comment @vhvb1989 - if not, do you mind providing a snippet as an example?

vhvb1989: Looks much better to me. Now you can create the ErrorWithSuggestion for any error :).

As a nit, I would consider using the same pattern as `fmt.Errorf()`, like:
image

And create a `internal.ErrorWithSuggest(string, any)`, where you can do the error wrapping like 

```go
err := ReLoginRequiredError{}
err.init(response, scopes, cloud)
return internal.ErrorWithSuggestion(
  fmt.Sprintf("%s, run `%s` to acquire a new token", err.scenario, err.loginCmd),
  err,
), nil
```

The function would automatically add the `Suggestion: ` label, so we don't need to set it everywhere and we can keep the same look and feel (probably making the label to use some color).

The function is just sugar-syntax for creating the Error. So, consider this a super NIT

vhvb1989 (approved):

cli/azd/pkg/auth/errors.go
JeffreyCA: Thanks, planning to address these in a follow-up PR

cli/azd/pkg/auth/errors.go
```diff
@@ -62,13 +75,22 @@ func (e *ReLoginRequiredError) init(response *AadErrorResponse, scopes []string,
...
+		e.loginCmd += " --use-device-code=false"
+		// TODO: Use aka.ms short link
+		//nolint:lll
+		e.helpLink = "https://learn.microsoft.com/azure/developer/azure-developer-cli/troubleshoot#azd-pipeline-config-failure-due-to-conditional-access-policy"
+	}
 }
 
 func (e *ReLoginRequiredError) Error() string {
-	return fmt.Sprintf("%s, run `%s` to log in", e.scenario, e.loginCmd)
+	return e.description
```
weikanglim, line 93: What do you think about naming this based on its intended usage, i.e. `errText` or `errString`?

cli/azd/pkg/auth/errors.go
```diff
@@ -69,9 +75,17 @@ func (e *ReLoginRequiredError) init(response *AadErrorResponse, scopes []string,
...
+	// BadTokenDueToSignInFrequency - The refresh token has expired or is invalid due to sign-in frequency checks by Conditional Access.
 	if slices.Contains(response.ErrorCodes, 70043) {
 		e.scenario = "login expired"
 	}
+
+	// DevicePolicyError - User tried to sign in to a device from a platform not currently supported through Conditional Access policy
+	if slices.Contains((response.ErrorCodes), 50005) {
+		e.loginCmd += " --use-device-code=false"
+		// TODO: Use aka.ms short link
+		e.helpLink = "https://learn.microsoft.com/azure/developer/azure-developer-cli/troubleshoot#azd-pipeline-config-failure-due-to-conditional-access-policy"
```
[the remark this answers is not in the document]
weikanglim: I'm wondering if the user may not have ran `pipeline config` here at all; while the documentation we provide here would be specific to pipeline config.

cli/azd/pkg/auth/errors.go
```diff
@@ -62,13 +75,22 @@ func (e *ReLoginRequiredError) init(response *AadErrorResponse, scopes []string,
 		}
 	}
 
+	// The refresh token has expired or is invalid due to sign-in frequency checks by Conditional Access.
 	if slices.Contains(response.ErrorCodes, 70043) {
 		e.scenario = "login expired"
 	}
+
+	// User tried to sign in to a device from a platform not currently supported through Conditional Access policy
+	if slices.Contains((response.ErrorCodes), 50005) {
```
weikanglim, line 84: Interesting to see the extra surrounding `(` `)` characters here. Usually this would force order of operations, but here I'm wondering if this is a literal that is unnecessary.

cli/azd/pkg/auth/errors_test.go
```diff
@@ -100,3 +100,35 @@ func TestReLoginRequired(t *testing.T) {
...
+	tests := []struct {
+		name string
+		resp *AadErrorResponse
+		want string
+	}{
+		{
+			"invalid_grant",
+			&AadErrorResponse{
+				Error:            "invalid_grant",
+				ErrorDescription: "AADSTS50005",
```
weikanglim, line 114: I will admit that it took a while for me to fully comprehend what is being tested here. I'm wondering if we would consider small improvements here:

- **Naming**. `TestReLoginRequiredError` instead of `TestReLoginRequiredDescription` -- We are testing the `.Error()` behavior; the description is an implementation detail member.
- **Values used in ErrorDescription**. I'm wondering if we would use values that are more descriptive of its' real-world usage. I could imagine that even a value like "description 1" makes it clearer that this is a text field; unlike "AADSTS50005" would suggest an error code field.

weikanglim (commented):
Thanks for starting the PR! Left some minor comments. Happy to discuss more on the documentation we provide for pipeline config / conditional access policy handling.

cli/azd/pkg/auth/errors_test.go
JeffreyCA: Thanks for the suggestions, updated

cli/azd/pkg/auth/errors.go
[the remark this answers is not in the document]
weikanglim: This is okay -- since pipeline config is the current only path; and the documentation would apply regardless.

weikanglim (approved):
LGTM! Thanks for making the improvement!

cli/azd/pkg/auth/errors.go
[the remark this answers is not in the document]
JeffreyCA: Updated with aka.ms link: https://aka.ms/azd/troubleshoot/conditional-access-policy

weikanglim (approved):

Issue #4592, opened 2024-11-27
[Issue] Powershell scripts should be executed with the -NoProfile switch


markwragg:
- [ x ] Make sure you've installed the latest version using [instructions in the wiki](../wiki/install)

**Output from `azd version`**

azd version 1.11.0 (commit 5b92e0687e1fa96dfc8292f4b900c0c58610b6a5)

**Describe the bug**

When PowerShell hook scripts are executed, if the user has a profile.ps1 then the profile script is executed. It would be better to call pwsh.exe with the `-noprofile` switch to ensure profile scripts are not loaded.

**To Reproduce**

Execute azd with a PowerShell hook script configured on a machine where a profile.ps1 exists in one of the standard directories: https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_profiles?view=powershell-7.4

**Expected behavior**

The profile script should not be executed.

**Environment**
Information on your environment:
    * PowerShell 7.4.6, Windows 11
    * IDE and version : Visual Studio Code

**Additional context**

The -noprofile switch for pwsh.exe is documented here: https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_pwsh?view=powershell-7.4#-noprofile---nop

Yionse:
We can reproduce the issue and are currently attempting to fix it.
 
The currently planned solution is to check if the user explicitly passed the `-NoProfile` parameter before executing the PowerShell hook command. If the user passed the `-NoProfile` parameter, then manually pass the `-NoProfile` parameter when calling pwsh.exe to execute the hook command. If the user did not pass the `-NoProfile` parameter, pwsh.exe will not include it when executing and will load the `profile.ps1` configuration file.

Yionse:
@rajeshkamal5050, Another solution is to add a configuration item for users to choose. As shown below:
![Image](https://github.com/user-attachments/assets/78e8a2f3-f3b3-4589-a003-5ac4ffd0e5ce)

kristenwomack:
https://github.com/Azure/azure-dev/pull/4595 closes this issue

PR #4595, opened 2024-12-03
Support Custom Command Format for `pwsh` in the shell Field

Yionse:
Fix issue #4592. This PR supports users to customize the command format of pwsh in the`shel` field, for example: `pwsh [more_options]`. Azd will directly launch the command window with this customized value.
 
For example, to avoid loading the `profile.ps1` file, users can set the `shell: pwsh -NoProfile` to start in no-profile mode, as shown:
image


v-xuto:
@rajeshkamal5050  Who should we reach out to for reviewing this PR?

cli/azd/pkg/tools/powershell/powershell.go
```diff
@@ -25,7 +25,15 @@ type powershellScript struct {
 // Executes the specified powershell script
 // When interactive is true will attach to stdin, stdout & stderr
 func (bs *powershellScript) Execute(ctx context.Context, path string, options tools.ExecOptions) (exec.RunResult, error) {
-	runArgs := exec.NewRunArgs("pwsh", path).
+
+	cmdConfig := " -NoProfile"
+	normalCmd := "pwsh"
+
+	if !options.IsRunProfile {
```
ellismg, line 32: Sort of wondering if instead of adding `IsRunProfile` and setting it in `hooks.go`, I wonder if we should be doing it here, so all invocations of `pwsh` get it out of the box?  I'm sort of wondering if we ever want to not do this for pwsh.

Yionse: @ellismg In the current version of azd, pwsh does not support the `-NoProfile` parameter because in the `exec.NewRunArgs("pwsh", path)` method in the [powershell.go](https://github.com/Azure/azure-dev/blob/main/cli/azd/pkg/tools/powershell/powershell.go#L28) file, the default is not have the -`NoProfile` parameter, and the hard-coded value is `pwsh`, as follow:
image
 
So if we want to support the `-NoProfile` parameter, we can only add `IsRunProfile` here.

rajeshkamal5050:
@hemarina @vhvb1989 can you help review?

cli/azd/cmd/hooks.go
```diff
@@ -235,7 +236,14 @@ func (hra *hooksRunAction) execHook(
 	})
 	defer hra.console.StopPreviewer(ctx, false)
 
-	runOptions := &tools.ExecOptions{StdOut: previewer}
+	isRunProfile := true
+	if hook.Windows != nil {
+		if hook.Windows.Run != "" {
+			isRunProfile = !strings.Contains(strings.ToLower(hook.Windows.Run), "-noprofile")
+		}
+	}
```
vhvb1989, line 244: Let's switch to use default starting values `false` for bools

```suggestion
	isRunWithNoProfile := false
	if hook.Windows != nil {
		if hook.Windows.Run != "" {
			isRunWithNoProfile = strings.Contains(strings.ToLower(hook.Windows.Run), "-noprofile")
		}
	}
```

Yionses: @vhvb1989, According to your suggestion, the initial value of `isRunWithNoProfile` is `false`, and the profile file will not be loaded by default. This means that if the user wants to load the Profile file, `pwsh -NoProfile` must be added to the command line, which actually contradicts the design of pwsh, because the function of `-NoProfile` is not to load the Profile file. Reference document: [-NoProfile](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_pwsh?view=powershell-7.4#-noprofile---nop).
 
I understand that you want to not execute the Profile file by default. We can also consider changing the parameter to `pwsh -profile`, and only execute the `Profile` file when this parameter exists, and not execute it by default. However, pwsh itself does not have the `-profile` parameter, which is inconsistent with the existing design of pwsh and may cause misunderstandings to users, as shown in the figure:

image
 
Therefore, we recommend following the existing design of pwsh: load the Profile file by default, and only not load it when `-NoProfile` is used.

vhvb1989: yes, I am hoping to only change the implementation logic to initialize a bool with `false`.  
The default functionality can stay as you have it.  Does it make sense?

Yionse: @vhvb1989 Got it.  I've submitted new commits based on your comment. Please review it, thanks.

cli/azd/pkg/ext/hooks_runner.go
```diff
@@ -159,6 +159,14 @@ func (h *HooksRunner) execHook(ctx context.Context, hookConfig *HookConfig, opti
 		defer h.console.StopPreviewer(ctx, false)
 	}
 
+	isRunWithNoProfile := false
```
vhvb1989, line 162: nit 

```suggestion
	var isRunWithNoProfile bool
```

vhvb1989 (approved):
LGTM

cli/azd/pkg/ext/hooks_runner.go
Yionse: @vhvb1989 Got it. I've submitted new commits based on your comment. Please review it, thanks.

cli/azd/pkg/ext/hooks_runner.go
```diff
@@ -159,6 +159,14 @@ func (h *HooksRunner) execHook(ctx context.Context, hookConfig *HookConfig, opti
 		defer h.console.StopPreviewer(ctx, false)
 	}
 
+	var isRunWithNoProfile bool
+	if hookConfig.Shell == "pwsh" {
+		if hookConfig.Run != "" {
+			isRunWithNoProfile = strings.Contains(strings.ToLower(hookConfig.Run), "-noprofile")
```
weikanglim, line 165: I'm wondering if this is the best approach here, it could match something that isn't an invocation flag argument, like: `myscript-noprofile.ps1`

weikanglim:
@Yionse @v-xuto I'm wondering when we're doing:

```pwsh
pwsh -NoProfile myScript.ps1
```

This would actually:

1. Launch a process `pwsh`
2. The `pwsh` process launches a child process `pwsh -NoProfile myScript.ps1`

which is less ideal.

Something more comprehensive is probably allowing the configuration of the `shell` command: https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#custom-shell

cli/azd/pkg/ext/hooks_runner.go
Yionse: @weikanglim Got it. I've submitted new commits based on your comment. Please review it, thanks.

Yionse:
@weikanglim, According to your suggestion, I have resubmitted. Now can configure the `shell`.
@hemarina, @vhvb1989, @ellismg Please review it, thanks.

Issue #4564, opened 2024-11-15
ci: Re-enable Terraform Tests


ellismg:
Our terraform tests in CI are currently disabled (in the case of the inner-loop `build-cli` leg) or failing (in the case of template tests) since moving to the TME environment since we need to use OIDC to connect.

This bug tracks re-enabling these disabled tests.

ellismg:
Trying to re-enable the remote state terraform tests failed with this error coming back from TF:

```
=== FAIL: cli/azd/test/functional Test_CLI_InfraCreateAndDeleteResourceTerraformRemote (113.42s)
    cli_test.go:833: DIR: /tmp/Test_CLI_InfraCreateAndDeleteResourceTerraformRemote560961891/001
    cli_test.go:841: AZURE_ENV_NAME: azdtest-lb61dc9
2024/11/15 19:05:09 Run exec: 'az group create --name rs-azdtest-lb61dc9 --location eastus2' , exit code: 0
2024/11/15 19:05:34 Run exec: 'az storage account create --resource-group rs-azdtest-lb61dc9 --name azdtestlb61dc9 --sku Standard_LRS --encryption-services blob' , exit code: 0
2024/11/15 19:05:36 Run exec: 'az storage account keys list --resource-group rs-azdtest-lb61dc9 --account-name azdtestlb61dc9 --query [0].value -o tsv' , exit code: 0
2024/11/15 19:05:37 Run exec: 'az storage container create --name tfstate --account-name azdtestlb61dc9 --account-key <redacted>' , exit code: 0
    cli.go:241: 17ms [stdout] 
    cli.go:241: 19ms [stdout] Initializing an app to run on Azure (azd init)
    cli.go:241: 19ms [stdout] 
    cli.go:241: 37ms [stdout] Enter a new environment name: 
    cli.go:241: 42ms [stdout] SUCCESS: Initialized environment azdtest-lb61dc9.
    cli_test.go:903: Starting infra create
    cli.go:241: 20ms [stdout] 
    cli.go:241: 24ms [stdout] Provisioning Azure resources (azd provision)
    cli.go:241: 27ms [stdout] Provisioning Azure resources can take some time.
    cli.go:241: 27ms [stdout] 
    cli.go:241: 1.489s [stdout] Select an Azure Subscription to use (or hit enter to use the default  1. Azure SDK Test Resources - TME (4d042dc6-fe17-4698-a23f-ec6a8d1e98f4)): Reading subscription and location from environment...
[...]

PR #4601, opened 2024-12-03
tests: Use OIDC for Terraform in template-tests

ellismg:
The terraform legs in our template test pipelines have been broken since we moved to ODIC and this change gets them working again. Because we run the template tests inside a dev container, we can't leverage the AzureCLI task to configure OIDC for `az`, so we have to do things the hard way.

The AzureRM Terraform Provider does support OIDC auth an even has knobs to configure the URL and token used when fetching the federated token used to do the ODIC dance, but it expects a slightly different protocol than what Azure Pipelines provides, so we now have a small adapter program `tfoidc` which exposes the endpoint in the format that the provider expects (it's the format spoken by GitHub Actions) which proxies to the Azure Pipelines version.

In addition, to support local-exec scenarios where `az` is invoked, `tfoidc` can manage calling `az login` and `az account set` as the AzureCLI task would (refreshing as needed).

The template-test pipelines have been updated to build this binary and place it in the root of the folder that is mounted into the dev container when the template-test script is run and to launch `tfoidc` and configure things as needed before running the `template-test.sh` script.

Contributes To: #4564

### DIFF
--- a/cli/azd/.vscode/cspell.yaml
+++ b/cli/azd/.vscode/cspell.yaml
@@ -116,6 +116,11 @@ overrides:
   - filename: internal/vsrpc/handler.go
     words:
       - arity
+  - filename: test/internal/tfoidc/main.go
+    words:
+      - tfoidc
+      - OIDCREQUESTURI
+      - oidctoken
 ignorePaths:
   - "**/*_test.go"
   - "**/mock*.go"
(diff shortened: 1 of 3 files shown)


weikanglim (approved):
Neat! Love the small program to make long-running tests possible.

PR #4616, opened 2024-12-09
publish: Stop publishing "All In One" container

ellismg:
As part of our release process, we build and push a container image that contains `azd` plus a large collection of language runtimes and other tools.

In practice, the resulting image is giant and anyone who wants to use `azd` inside of a container is better served by building their own and including `azd`. We also have not been updating this container over time (and a while back we removed it from the list of artifacts we list per PR).

To make our lives simpler, let's just stop building and publishing this container image. If folks were depedent on it and need updates, let's move them over to using one of our supported install methods to get them to add `azd` to their container image as the build it instead of using ours.


weikanglim (approved):

danieljurek (approved):

JeffreyCA:
Are we going to remove it from the README at some point as well?

image

danieljurek:
There's a small amount of engsys cleanup work that's needed here. https://github.com/Azure/azure-dev/issues/4619

Also, remove the line from `README`.md` that @JeffreyCA mentioned: https://github.com/Azure/azure-dev/blob/main/README.md?plain=1#L9

ellismg:
> Are we going to remove it from the README at some point as well?

Thanks, @JeffreyCA - I updated the PR to remove it.

> There's a small amount of engsys cleanup work that's needed here. https://github.com/Azure/azure-dev/issues/4619

@danieljurek - I looked that over to see if there was anything there I could do myself but I don't think I can.  I'll land this is as is (with the change Jeff pointed out) and if you need help with other stuff in #4619, I am happy to help.

JeffreyCA (approved):

PR #4624, opened 2024-12-11
refactor: Move remaining code in `pkg/tools/azcli/` to `pkg/azapi`

### LINKED ISSUE #854: Deprecate and split up `AzCli` interface

ellismg:
When we started `azd` instead of making REST calls against the Azure using the SDKs, we had an `AzCli` type which shelled out to `az` to preform work. Over time we refactored this code from using the `exec` package to run `az` to use the proper go SDKs from azure-sdk-for-go and as we did this we ended up pulling off related operations into their own services and packages.

Over time we got close to completely removing the `AzCli` type, but a few operations still lingered.

To finish the job while minimizing the destruction to the rest of the codebase, I've taken the `AzCli` type in `pkg/tools/azcli/`, moved all the code into `pkg/azapi` and then renamed the type from `AzCli` to `AzureClient` (following the azure-sdk-for-go convention of `Client` suffixes). While doing so, I removed the intermediate wrapper interface and now other components just take a `*azapi.AzureClient` instead of a `azcli.AzCli`.

Fixes #854


ellismg:
There were a handful of methods left of the old `AzCli` interface, and I considered breaking things down into a set of perhaps finer grained types as part of moving this code (instead of a more generic `AzureClient` as I did here) but I didn't see much of the benefit of doing so at this time.  I also keep the `AzCli` prefix on some of the exported types we have (even though the values don't come from the CLI anymore) again to just reduce the size of the diff.  I could imagine us removing these wrapper types at some point - they mostly existed so we could `json.Unmarshal` the output of `az`, which we don't do anymore - we could consider just exposing the types from the SDK directly.

weikanglim (approved):
I love the minimally invasive change we're making here, and have no concerns -- it's approved for me.

I really like that we're keeping the shape of the type from `AzCli` -> `AzureClient` while we figure out what this all means long term.

On the long term direction, just dropping a note here while things are top of mind. I feel like minimally, the things we want:

1. Hide the "glue-code" that we often write for setting up azure clients with the correct credentials
2. Have the ability to define common library functions that capture the repeated, higher-level interactions with the REST API

I feel that if we think about these problems separately; and also think about solving them separately -- we can avoid the additional layer worth of type abstractions we have been promoting.

PR #4629, opened 2024-12-13
Migrate to new CDN endpoint 

### LINKED ISSUE #4621: CDN migration
Migrate release CDNs:

* [ ] Installers
    * [ ] Install scripts (.ps1, .sh) 
    * [ ] DevOps extension
    * [ ] GitHub Action 

* Some test code is impacted but mitigation is simple
* Choco icon image might be impacted but will work going forward (they use their own copy too, so impact is unlikely) 
* Kitchen sink logo image is also impacted but that is also served from the app, not from the location we gave them (they use their own copy) 

Not impacted:
* brew ([uses GitHub releases](https://github.com/Azure/homebrew-azd/blob/main/Formula/azd.rb#L6))
* winget ([uses GitHub releases](https://github.com/microsoft/winget-pkgs/blob/master/manifests/m/Microsoft/Azd/1.11.100/Microsoft.Azd.installer.yaml#L13))
* choco ([uses choco distribution](https://community.chocolatey.org/packages/azd#files))

danieljurek:
Fixes #4621

Impacted areas: 
* Install Scripts 
* Azure DevOps 
* GitHub Actions
* Some testing
* Choco image location (choco makes local copies of these so the image shouldn't go away and if it does, later releases will show the correct image)

### DIFF
--- a/cli/azd/.vscode/cspell.yaml
+++ b/cli/azd/.vscode/cspell.yaml
@@ -1,18 +1,20 @@
 import: ../../../.vscode/cspell.global.yaml
 words:
+  - azcloud
   - backoff
   - Canonicalize
+  - chinacloudapi
   - Codespace
   - Codespaces
   - devcontainers
+  # CDN host name
+  - gfgac2cmf7b8cuay
   - goversioninfo
   - nosec
   - Retryable
   - runcontext
-  - azcloud
-  - usgovcloudapi
-  - chinacloudapi
   - unmarshals
+  - usgovcloudapi
 languageSettings:
   - languageId: go
     ignoreRegExpList:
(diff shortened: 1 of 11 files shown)


vhvb1989 (approved):

danieljurek:
GH Actions tracked in https://github.com/Azure/setup-azd/pull/213

danieljurek:
Validated: 

* PR changes in install scripts
    * Windows
    * Linux
    * macos
* DevOps extension on windows-latest, ubuntu-latest, macos-latest
* GitHub Actions on windows-latest, ubuntu-latest, macos-latest

All succeeded as expected and logged URL for new CDN.

PR #4630, opened 2024-12-13
rpc: Support handlers of any arity

ellismg:
Previously, we had of hand written adapters that allowed exposing go functions over our JSON-RPC boundary.  Changes to the number of arguments or return values would require changes to the adapter used (or writing a new adapter by copy/pasting an existing one).

Now, we have a single `NewHandler` which allows exposing arbitrary go functions over the RPC boundary (as long as the follow some rules, like taking a `context.Context` as the first paremeter, and returning a value and `error` or just an `error`.

As part of this change, the `IObserver` type is now called `Observer` and it is passed by reference instead of value in signatures.

### DIFF
--- a/cli/azd/internal/vsrpc/aspire_service.go
+++ b/cli/azd/internal/vsrpc/aspire_service.go
@@ -28,7 +28,7 @@ func newAspireService(server *Server) *aspireService {
 // GetAspireHostAsync is the server implementation of:
 // ValueTask<AspireHost> GetAspireHostAsync(Session session, string aspireEnv, CancellationToken cancellationToken).
 func (s *aspireService) GetAspireHostAsync(
-	ctx context.Context, rc RequestContext, aspireEnv string, observer IObserver[ProgressMessage],
+	ctx context.Context, rc RequestContext, aspireEnv string, observer *Observer[ProgressMessage],
 ) (*AspireHost, error) {
 	session, err := s.server.validateSession(rc.Session)
 	if err != nil {
@@ -65,7 +65,7 @@ func (s *aspireService) GetAspireHostAsync(
 // RenameAspireHostAsync is the server implementation of:
 // ValueTask RenameAspireHostAsync(Session session, string newPath, CancellationToken cancellationToken).
 func (s *aspireService) RenameAspireHostAsync(
-	ctx context.Context, rc RequestContext, newPath string, observer IObserver[ProgressMessage],
+	ctx context.Context, rc RequestContext, newPath string, observer *Observer[ProgressMessage],
 ) error {
 	_, err := s.server.validateSession(rc.Session)
 	if err != nil {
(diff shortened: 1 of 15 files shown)


cli/azd/internal/vsrpc/aspire_service.go
```diff
@@ -28,7 +28,7 @@ func newAspireService(server *Server) *aspireService {
 // GetAspireHostAsync is the server implementation of:
 // ValueTask<AspireHost> GetAspireHostAsync(Session session, string aspireEnv, CancellationToken cancellationToken).
 func (s *aspireService) GetAspireHostAsync(
-	ctx context.Context, rc RequestContext, aspireEnv string, observer IObserver[ProgressMessage],
+	ctx context.Context, rc RequestContext, aspireEnv string, observer *IObserver[ProgressMessage],
```
ellismg, line 31: It's non-obvious from the diff why I went and did this.  The issues that IObserver needs some special attention during unmarshalling.  The way that works is we have an internal interface:

```go
// connectionObserver is an interface that can be implemented by types that want to be notified when they are deserialized
// during a JSON RPC call.
type connectionObserver interface {
	attachConnection(c jsonrpc2.Conn)
}
```

When we unmarshall arguments, we check to see if the type we unmarshalled to implements this interface and if so, we call it.  IObserver's implementation of the method stores the connection so it can send messages to it later, as items are emitted.

The interface is implemented on `*IObserver[T]` not `IObserver[T]`.

The old version of `unmarshalArg` handled this by adding a level of indirection on thet ype during interface checking:

```go
	if v, ok := (any(&arg)).(connectionObserver); ok {
		v.attachConnection(conn)
	}
```

But it felt cleaner to not do this extra level of indirection in the new version and instead ensure the argument was just `*IObserver[T]`.

ellismg: I can image that we could also do both - check to see if T and *T implement the interface, (preferring one if both are there) and calling the function.

weikanglim: The current change looks more idiomatic to me; the old indirection does look a little weird.

weikanglim (approved):
LGTM!

cli/azd/internal/vsrpc/aspire_service.go
wbreza: From looking at the code I see that `IObserver[T]` is actually a struct and not an interface which I would have expected giving the name.  I would consider renaming this to just `Observer[T]` to avoid this initial confusion.  I was further confused by seeing the pointer reference to the assumed interface making me think of pointer to a pointer initially.

wbreza (approved):
Looks good! The only recommendation I have to rename the `IObserver` to `Observer` so that people don't get confused and think it is an interface when it actually is a struct.

Issue #4570, opened 2024-11-20
Support muti-level logger in azd


rujche:
For current azd, the logger only has one level: output or not output.
1. Without "--debug" program parameter, there is no output. Customer will lack necessary information.
2. WIth "--debug" program parameter, there will be too much (hundreds or even thousands of lines) logs.

It's a reasonable request to support multi-level log in azd. Here are common used levels: info, warn, error. 

And I found loggers for go lang:
1. https://github.com/sirupsen/logrus
2. https://github.com/uber-go/zap

ellismg:
FYI, In https://github.com/Azure/azure-dev/pull/4635 I am adopting `log/slog` which is leveled.  I used `Info` as the level everywhere during my conversion, but we could use other levels as well.

I would strongly prefer we use `slog` instead of taking a dependency on a third-party package.

rujche:
Hi, @ellismg 

Thanks for telling me this. I plan to adopt this to our feature branch: https://github.com/Azure/azure-dev/pull/4473

weikanglim:
> Without "--debug" program parameter, there is no output. Customer will lack necessary information.

@rujche This is quite interesting. I'm wondering what information that users aren't able to access currently. Or perhaps there was something that pushed us in the direction of needing this.

In my mind, all error and warning logs should be printed out to the interactive console. I do agree with @ellismg on perhaps having one extra verbosity level for the debug log in some cases, but I don't feel we benefit a whole lot from it today.

The one downside I could think of when having multiple log levels is in the nature of complexity, i.e. with having multiple things to deal with, possibility of human errors arises. A programmer may mistakenly intend to log something to the wrong log levels; similarly, a user may provide the log file of a different verbosity. We should also be clear about the usage of logs here -- it's meant for debugging, not a place for error and warning messages to hide as I've seen and experienced in other tooling.

rujche:
Hi, @weikanglim 

Sorry for my late response.

> I'm wondering what information that users aren't able to access currently. Or perhaps there was something that pushed us in the direction of needing this.

I have this requirement because I want to print the `uses` related environment variables in this PR: https://github.com/azure-javaee/azure-dev/pull/46/files#diff-565dec703d1ca826ff53acb64c2a4f4bedf04533661c163fa11bcad3b8dbc2cd

> ![Image](https://github.com/user-attachments/assets/d9ac6d7f-5846-4092-870f-fe76982f6495)

[...]

PR #4635, opened 2024-12-17
Use `log/slog` instead of `log`

ellismg:
Go added [`log/slog`](https://pkg.go.dev/log/slog) to the standard library in Go 1.21.  It provides a leveled logging API that allows for structured logging. If this package had existed in the standard library when we started `azd`, we would have used it over `log`.

This change moves us from `log` to `log/slog`.

To do so, I've replaced all the calls to `log.Print`, `log.Printf` and `log.Println` with calls to `slog.InfoContext`.  If a `context.Context` object was available in the method, I used it, otherwise I used `context.TODO()` so we can go back and thread the context parameter.  Passing the context object like this should allow us to tie these log messages to tracing spans at some point.  In practice, `slog.InfoContext(context.TODO, ....)` behaves `slog.Info(...)` (since `context.TODO()` is like `context.Background()` and `slog.Info` calls `slog.InfoContext` with `context.Background()`.

While `slog` does have levels, for the first pass, I chose to always use `slog.InfoContext` even if `slog.ErrorContext` or `slog.WarningContext` may have made more sense based on the log message.  We can evolve this over time.
 
Since `slog` does structured logging, it is expected that insead of intermixing the logging message with the data to be logged, that you pass these as separate values, I did light editing of the logging messages when it made sense.
 
Calls to `log.Panic` and `log.Panicf` were replaced with direct calls to `panic` (perhaps with a call to `fmt.Sprintf` to produce the message to pass to panic) since `slog` doesn't have helpers like `Panic` or `Fatal`.  This does mean we no longer log a message in these cases, but from inspection it felt like this would be OK.

The `--debug` output of `azd` is slightly different after this change, since we now have a level printed for the log message and the data is printed after the message:

Before:

```
matell@Matts-Work-MacBook azd % azd version --debug 
2024/12/16 20:20:55 main.go:56: azd version: 0.0.0-dev.0 (commit 0000000000000000000000000000000000000000)
2024/12/16 20:20:55 main.go:214: using cached latest version: 1.11.0 (expires on: 2024-12-18T02:28:37Z)
2024/12/16 20:20:55 middleware.go:108: running middleware 'debug'
2024/12/16 20:20:55 middleware.go:108: running middleware 'ux'
azd version 0.0.0-dev.0 (commit 0000000000000000000000000000000000000000)
2024/12/16 20:20:55 main.go:88: eliding update message for dev build
```

After:

```
matell@Matts-Work-MacBook azd % azd version --debug              
2024/12/16 20:21:24 main.go:57: INFO azd version version="0.0.0-dev.0 (commit 0000000000000000000000000000000000000000)"
2024/12/16 20:21:24 main.go:215: INFO using cached latest version version=1.11.0 expiresOn=2024-12-18T02:28:37Z
2024/12/16 20:21:24 middleware.go:108: INFO running middleware name=debug
2024/12/16 20:21:24 middleware.go:108: INFO running middleware name=ux
azd version 0.0.0-dev.0 (commit 0000000000000000000000000000000000000000)
2024/12/16 20:21:24 main.go:89: INFO eliding update message for dev build
```

### DIFF
--- a/cli/azd/cmd/cobra_builder.go
+++ b/cli/azd/cmd/cobra_builder.go
@@ -2,7 +2,6 @@ package cmd
 
 import (
 	"fmt"
-	"log"
 	"slices"
 	"strconv"
 	"strings"
@@ -175,7 +174,7 @@ func (df *docsFlag) Set(value string) error {
 	// Setting help to true will make cobra to stop and call the HelpFunc
 	if err = df.command.Flag("help").Value.Set("true"); err != nil {
 		// dev-issue: help flag should be already been added when
-		log.Panic("tried to set help after docs parameter: %w", err)
+		panic(fmt.Sprintf("tried to set help after docs parameter: %v", err))
 	}
 
 	// keeping the default help function allows to set --help with higher priority and use it
@@ -213,7 +212,7 @@ func (cb *CobraBuilder) bindCommand(cmd *cobra.Command, descriptor *actions.Acti
 		consoleFn: func() input.Console {
 			var console input.Console
 			if err := cb.container.Resolve(&console); err != nil {
-				log.Panic("creating docs flag: %w", err)
+				panic(fmt.Sprintf("creating docs flag: %v", err))
 			}
 			return console
 		},
--- a/cli/azd/cmd/auth_login.go
+++ b/cli/azd/cmd/auth_login.go
@@ -8,7 +8,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
-	"log"
+	"log/slog"
 	"os"
 	"strconv"
 	"strings"
@@ -350,7 +350,7 @@ func (la *loginAction) Run(ctx context.Context) (*actions.ActionResult, error) {
 		if err := la.accountSubManager.RefreshSubscriptions(ctx); err != nil {
 			// If this fails, the subscriptions will still be loaded on-demand.
 			// erroring out when the user interacts with subscriptions is much more user-friendly.
-			log.Printf("failed retrieving subscriptions: %v", err)
+			slog.InfoContext(ctx, "failed retrieving subscriptions", "err", err)
 		}
 	}
 
@@ -405,7 +405,7 @@ func runningOnCodespacesBrowser(ctx context.Context, commandRunner exec.CommandR
 	if err != nil {
 		// An error here means VSCode is not installed or found, or something else.
 		// At any case, we know VSCode is not within a webBrowser
-		log.Printf("error running code --status: %s", err.Error())
+		slog.InfoContext(ctx, "error running code --status", "err", err)
 		return false
 	}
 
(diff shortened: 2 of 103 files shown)


ellismg:
Happy to break this up into chunks (maybe across package boundaries) if that's helpful.  In general the two logging packages can coexist.

cli/azd/cmd/cobra_builder.go
```diff
@@ -213,7 +212,7 @@ func (cb *CobraBuilder) bindCommand(cmd *cobra.Command, descriptor *actions.Acti
 		consoleFn: func() input.Console {
 			var console input.Console
 			if err := cb.container.Resolve(&console); err != nil {
-				log.Panic("creating docs flag: %w", err)
+				panic(fmt.Sprintf("creating docs flag: %v", err))
```
weikanglim, line 215: @vhvb1989 had mentioned something about `log.Panic` being preferable before, so curious what he thinks here.

weikanglim (commented):
Overall, I really like that `slog` exists and could really see its usefulness if we were to take advantage of the structured metadata. Currently, I'm undecided if I like `slog` better, as I still prefer simplicity of `log` and the only included `--debug` level. If we were to have plans to take full advantage of this, I could be convinced otherwise.

One thing that comes to mind is that our debug logs are intended to be a "somewhat dev-friendly message for debugging purposes" and it feels that the current handwritten `Printf` format strings do a slightly better job at that. With the structured formatting, the messages look "semi-structured" which right now doesn't look that great. Perhaps with more looks at these logs over time, I'd be more in favor of it.

It also feels that `slog.InfoContext` and the likes are very specific to scenarios where `ctx` matters; and yet the codebase serves a simple CLI that is unlikely? to matter.

Just curious about your overall thoughts on this @ellismg. The simpler logging code of today is still pushing me towards keeping the simple `Printf`, but I wonder if you have some future ideas.

ellismg:
> Just curious about your overall thoughts on this @ellismg. The simpler logging code of today is still pushing me towards keeping the simple `Printf`, but I wonder if you have some future ideas.

It's a reasonable point - There are a few things that motivated me to do this pass:

1. Adopting `slog` would make it easier to do leveled logging, which may be useful for a dev logging scenario.  In #4570 there's discussion about how having levels might make things easier, and I will admit there were times I wish we had a "verbose" level and previous Matt had just logged more data that I could get back with a `--debug` (and maybe a new `--debug-level verbose` or something).
2. The idea that structured logging may be useful since it makes writing tools that consume the log data more useful (no hard evidence we'd write such tools, except that we did sort of exploit a variant of this idea in some of our telemetry tests where we capture otel events to a file and then parse that file in a structured way, and it sort of plays into...)
3. The idea that at some point we might unify the logging and tracing systems and start producing OTel compatible logs tied to our OTel spans.  I've always in the back of my mind thought it would be "cool" to be able to capture an OTel trace of arbitrary azd commands to a file and then be able to view them later.  That's what pushed me towards the `Context` versions of the the logging overloads, so we could recover the span information (it felt like something like https://pkg.go.dev/go.opentelemetry.io/contrib/bridges/otelslog would let us get to a place where we could emit spans + logs that could be viewed either via Jager or the Aspire Dashboard (which you can also run locally as just an OTel collection endpoint) and that would be a useful observability and diagnostic primitive.

And, to be honest, things were quiet over the holidays and so it felt like if we wanted to make a run at adopting `slog` this was a good time to do the pass over the codebase and see what it felt like.

Issue #4606, opened 2024-12-04
[Issue] `azd init` should inspect dockerfile ARGS


jongio:
- [ ] Make sure you've installed the latest version using [instructions in the wiki](../wiki/install)

**Output from `azd version`**
1.11

**Describe the bug**

My Docker files require some arguments.

The Azure.Yaml Docker file buildArgs is nice once you get it, but isn't very intuitive on how to set up the build args.

It would be great if the AZD init process determined what Args are in the Docker file. And then. I'll automatically added them to the azure yeml file.

Final azure.yaml I created:
![Image](https://github.com/user-attachments/assets/aea60dc5-b6e3-42b7-80fe-f916bff25b55)

My Dockerfile with the ARGS
![Image](https://github.com/user-attachments/assets/d774ed60-275e-4e93-a205-37eb37763550)

v-xuto:
We are following up on this issue. cc @rajeshkamal5050

Menghua1:
@weikanglim My current approach is ​​to add a function that extracts the parameters following `ARG` in the `Dockerfile` under the service directory, and then write it as `buildArgs` in the `azure.yaml` file. 

If there are no problems, please review this PR https://github.com/Azure/azure-dev/pull/4639.

weikanglim:
I feel that the UX we're delivering here is extremely unintuitive.

Docker build args are primarily used as a passed-in CLI flag, in the [docs](https://docs.docker.com/build/building/variables/#env-usage-example):

`docker build --build-arg NODE_ENV=development `

I feel it's extremely weird if `azd` were to now treat this as an environment variable that is passed in. Would it be more idiomatic to allow for:

`azd package -- --build-arg NODE_ENV=development` as a flag passthrough?

That way we avoid having this needing configuration upfront.

Menghua1:
@weikanglim Thanks for your feedback! Based on the analysis, we recommend supporting both **Approach 1 and Approach 2**. This design retains the flexibility of Approach 1, which aligns with the habits of Docker users, while also allowing users to centrally manage ARG parameters through Approach 2. In particular, Approach 2 offers better maintainability and consistency in multi-environment scenarios.

1. Using `azd package -- --build-arg NODE_ENV=development`.
  
2. Using the `buildArgs` field in the `azure.yaml` configuration. 


**Detailed Design:**
[...]

PR #4639, opened 2024-12-18
Enhance `azd init` to auto-detect Dockerfile ARGS and update azure.yaml

Menghua1:
Fix issue https://github.com/Azure/azure-dev/issues/4606, this PR adds automatic detection of ARG parameters in Dockerfiles during `azd init`.

When a user runs `azd init` to initialize an existing application as an azd project, the command will now scan the Dockerfile, identify all defined ARG parameters, and automatically populate them into the `buildArgs` section of the generated `azure.yaml` file.

### DIFF
--- a/cli/azd/internal/repository/app_init.go
+++ b/cli/azd/internal/repository/app_init.go
@@ -1,6 +1,7 @@
 package repository
 
 import (
+	"bufio"
 	"context"
 	"fmt"
 	"maps"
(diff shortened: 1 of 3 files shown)


cli/azd/internal/repository/app_init.go
```diff
@@ -44,6 +45,34 @@ var dbMap = map[appdetect.DatabaseDep]struct{}{
...
+	}
+	defer file.Close()
+
+	scanner := bufio.NewScanner(file)
+	for scanner.Scan() {
+		line := strings.TrimSpace(scanner.Text())
+		if strings.HasPrefix(line, "ARG ") {
+			argLine := strings.TrimPrefix(line, "ARG ")
+			if len(argLine) > 0 {
+				buildArgs = append(buildArgs, osutil.NewExpandableString(argLine))
```
ellismg, line 64: I wonder if creating an `osutil.ExpandableString` here is the right thing?  I think that we probably just want to use `string` (or maybe a structured type that contains the arg name and optional default value).  I don't think you'd ever want to call `Envsubst` or `MustEnvsubst` on this value since the docker language doesn't think about doing env-var substitution for `ARG`s.

For the docs, it looks like the `ARG` directive in a dockerfile could look like this:

```dockerfile
ARG foo=${BAR}
````

Which from my understanding would mean an argument called "foo" with the default value of `${BAR}`.

I'm a little worried that when faced with a file like this, we'd end up doing the wrong thing end to end (but I will admit that from the linked issue, I also don't have a great idea of what the expectation is here either).

Issue #4075, opened 2024-07-05
[Issue] -o is available, but not in help.


jongio:
Many commands support the -o parameter to allow the user to select an output format.  But when you run -h on commands you don't see -o as an option:

![image](https://github.com/Azure/azure-dev/assets/2163001/2752612f-d921-4c3b-9ed5-88a0429a4df0)

weikanglim:
From https://github.com/Azure/azure-dev-pr/pull/1491 [rich view](https://github.com/Azure/azure-dev-pr/blob/496c456ca0e37a3fdb3097851b050c745aecd7c9/docs/specs/3-13-2023-commands-fit-n-finsh.md), there were some unresolved follow-ups with regards to `--output` flag before the product team was comfortable making this available.

Back then, I had some initial reservations about `--output` being applied as a general pattern here, and would prefer `--json` for JSON output. Generally speaking, most commands do not actually produce meaningful output for formatting purposes. I do also find the naming of `none` output to perhaps be less self-documenting in nature.

If we were to move in any direction here, I'd recommend having a product member to do a quick review of the current usage and take in community input to make a quick proposal as short feature work.

spboyer:
The `-o/--output` flag is intentionally hidden from help text via `MarkHidden` in `cli/azd/pkg/output/parameter.go:30`. The flag works for commands that support it (e.g., `azd env list -o json`), but is hidden from general help to reduce noise — not all commands support structured output.

Closing as by-design.

PR #4656, opened 2025-01-02
Do not hide `--output` Flag in help

Menghua1:
Fix issue https://github.com/Azure/azure-dev/issues/4075.

Commented out the `_ = f.MarkHidden(outputFlagName)` line in the `AddOutputFlag` function to ensure the `-o` flag is no longer hidden.
The `-o` flag is now correctly listed in the command help, making it easier for users to discover and use it.

### DIFF
--- a/cli/azd/pkg/output/parameter.go
+++ b/cli/azd/pkg/output/parameter.go
@@ -25,7 +25,7 @@ func AddOutputFlag(f *pflag.FlagSet, s *string, supportedFormats []Format, defau
 	description := fmt.Sprintf("The output format (the supported formats are %s).", strings.Join(formatNames, ", "))
 	f.StringVarP(s, outputFlagName, "o", string(defaultFormat), description)
 	//preview:flag hide --output
-	_ = f.MarkHidden(outputFlagName)
+	// _ = f.MarkHidden(outputFlagName)
 
 	// Only error that can occur is "flag not found", which is not possible given we just added the flag on the previous line
 	_ = f.SetAnnotation(outputFlagName, supportedFormatterAnnotation, formatNames)


wbreza:
I believe that one of the reasons we decided to hide the `--output / -o` flag was because we were hesitant to lock-in on the format/schema of responses when using `json` output option.

@rajeshkamal5050 / @weikanglim  to add any more context. here.

cli/azd/cmd/testdata/TestUsage-azd-infra-synth.snap
```diff
@@ -7,6 +7,7 @@ Usage
 Flags
     -e, --environment string 	: The name of the environment to use.
         --force              	: Overwrite any existing files without prompting
+    -o, --output string      	: The output format (the supported formats are none).
```
weikanglim, line 10: This is a negative UX impact -- we're offering a flag that doesn't change the behavior of the command.

weikanglim:
@wbreza I believe you are indeed quite right, and I would encourage some discussion on the issue before proceeding forward here.

ellismg:
A plus one to the history here - another thing that we should be mindful of if/when we land this is the fact that we "none" here to mean "normal human readable output" but I expect almost anyone would think that `-o none` would be something like passing `> /dev/null` to the command (i.e. silence the output).

The formatting stuff is one of the things that we built early on thinking we might use it in a more uniform and general way (similar to how `az` does stuff (that's also something that pushed us towards the table formatting)) but since then we've focused more on the interactive UX and less as azd as a general purpose tool for scripting.  A few commands like `azd show` we really think about the `--output json` format because we expect them to be in tooling scenarios, but for most commands we really only think about the interactive experience.

It may make sense over time to shift the mindset from "every command results in an object which can be rendered in a bunch of different formats (like `az`) and towards a most commands write output for humans, not tools, and the few that have tool driven counterparts have `--json` or something to control JSON formatting vs trying to model it as a general purpose output stream formatting problem that hits every command.

vhvb1989:
What if we make the `--output` flag an alpha thing? Then folks can expect the output to change at some time.  
I think that would be a little better than hiding the flag.

Also, let's make azd to add the flag only when there is at least 2 formats supported and not when format `none` is the only one.
The help string should also handle one format v/s more than one like:
- "Supports JSON output format"  for when only supporting none and JSON
- "Supports JSON and table formats" for when supporting none, JSON and table.

Issue #4503, opened 2024-10-30
aspire: Support Shared Infrastructure Generated By Apphost


ellismg:
In Aspire 8, `azd` was responsible for all infrastructure generation for an Aspire Application.  With Aspire 9 and the introduction of the Azure Provisioning set of packages, it is now possible for the App Host to generate the IaC for the container app for a given project and provide that to `azd` for deployment.  This allows Aspire developers to customize features of the container app in C# code instead of directly via bicep/yaml via `infra synth`.  This support was delivered as the `project.v1` resource type.

We want to support this same experience for the shared infrastructure (the container app environment, the container app registry, the managed identity that is allocated for the application) post Aspire 9.  As part of doing this, we need a way for `azd` to stop generating these parts of the shared infrastructure as they move over to be generated by Aspire (and `azd` should just "consume" the generated versions instead).

This issue tracks supporting designing and supporting what we need to support something like this.

The current thinking is that our "dot imports" in binding expressions which today Aspire uses in the manifest to wire up logical "outputs" of the shared infrastructure to "inputs" on the bicep generated for each container app is a way we can detect if the Aspire app "needs" a resource from `azd` and if there is no reference do not generate the infrastructure.  That should be a useful heuristic to employ.  The `Image` input of a container app might be interesting, since that references the container registry and `azd` needs to understand if it's expected to create that or not.

We also need to understand how Aspire communicates back to `azd` information about things like the container registry for a service or other cases where `azd` needs to get information or reference a resource generated by the AppHost.

It would be nice if we could do this incrementally instead of having one big bang, but supporting that may be too hard of a task unless we have a decent way of doing cross tool references.

vhvb1989:
fixed by https://github.com/Azure/azure-dev/pull/4884

We have introduced an alpha feature `alpha.apphost.infra.migration`. When this switch is on, azd does not generate the based compute infrastructure for Aspire (resources.bicep).

With this strategy, we don't need the Aspire Manifest to model a way to ask AZD for skipping the base infra generation. Instead, it allows .NET folks to immediately turn it off and work on making the AppHost to generate the base models.

In the future, when the AppHost can produce all it needs, we will remove the alpha feature and skip the base infrastructure generation by default.  This is way the switch was named `migration`

PR #4678, opened 2025-01-08
[Feature] Allow overriding of built in app host templates

FullStackChef:
### Describe the feature or problem you’d like to solve
When using AZD with Aspire the bicep templates make assumptions about the infrastructure and specifically the naming conventions / existence of resources 

AZD uses embedded file system that contains templates for creating base bicep files for use with app host - these are good for a large number of cases however create issues when attempting to migrate existing projects to use Aspire & AZD as well as situations where organisations have pre-defined naming conventions

### Proposed solution
Introduce an overlay file system for app host based generation that allows developers to override one or more of the existing files by creating a mirrored directory that contains the templates they want to override

How will it benefit CLI and its users?
For most users there will be no difference in the operation of the cli - however where users may be prevented from using the cli due to limitations of these templates, it will allow customisation of these templates without having to synth infra and then make changes which causes issues when there are changes

### Additional context

Add any other context like screenshots or mockups are helpful, if applicable.
Image


FullStackChef:

ellismg:
The other angle of attack here is work that will happen on the Aspire side, which involves moving the responsibility of generation of the shared infrastructure into the Aspire app host itself (similar how in Aspire 9 they delivered the ability to use `Azure.Provisioning` to configure aspects of the generated container app and other Azure resources).

My concern about allowing arbitrary template replacement here is just one of supportability (since there's now an API between `azd` and the authors of the template with respect to the shape of the context object for the template as well as functions that can be invoked) and we never really thought about APIs at that layer.  Given that the plan was long term to move all this responsibility into the Aspire program itself, I'd worry a little about adding this customization support.

The OverlayFS stuff is cool - and it's neat to see how building on the io.FS abstraction allowed you a nice way to plug in here.

If we do land this (i.e. we are comfortable with the extension point) I think we should at least consider reworking the code so we don't panic in `init()` if we can't compute the CWD.  I might consider moving that logic (and the construction of the overlay FS, perhaps based on a stat or something) elsewhere where if there was an `error` we could return it and have it reported by the standard flow - but I will admit I didn't think too much about the diff because I'm not certain we want to add the extension point.

ellismg:
https://github.com/Azure/azure-dev/issues/4503 is the issue we had that tracks the fact that we want to move this responsibility into the app host (and hence there would need to be a way to have `azd` detect that the app host was now generating this shared infrastructure and we should not).

ellismg:
Thanks for opening this and starting the discussion, @FullStackChef - I kicked off the PR builds so you should get privates generated with these changes if that's helpful for you.

Jotted some quick thoughts in the issue based on my understand on where Aspire is moving.

FullStackChef:
Thanks @ellismg - Honestly I see this as a temporary stop gap until the Apphost takes responsibility for the shared infrastructure. 
(this is why I added the overlay FS directly to the app host package) it was my expectation that once the app host was taking care of shared infrastructure the generate.go / apphost/templates etc. would likely change significantly and this could be removed

Ultimately - the change has allowed me to build a local version of AZD that we can chuck in our repos for now so if it's not suitable for release I'm not going to be upset, I am aware from conversations in the Aspire discord channel that there are other people that are struggling with some of the restrictions of the built-in templates (and it was going to delay our adoption of the benefits of azd + aspire)

ellismg:
FWIW, @FullStackChef, I had our bots create privates with your fix incase that's helpful for you as well :-), even though it sounds like from your message you have all the machinery on your end to produce privates.

FullStackChef:
> FWIW, @FullStackChef, I had our bots create privates with your fix incase that's helpful for you as well :-), even though it sounds like from your message you have all the machinery on your end to produce privates.

It's actually incredibly helpful thank you @ellismg - my "build" was to build the exe locally and c/p 😬  this is much more useful

FullStackChef:
@ellismg I've also addressed your feedback around the panic in init - I've adjusted this so it instead gracefully falls back to the embedded fs

vhvb1989:
> > FWIW, @FullStackChef, I had our bots create privates with your fix incase that's helpful for you as well :-), even though it sounds like from your message you have all the machinery on your end to produce privates.
> 
> It's actually incredibly helpful thank you @ellismg - my "build" was to build the exe locally and c/p 😬 this is much more useful

The private binaries are eventually removed after some time.  Let us know if that happen and we can re-generate new ones.

PR #4684, opened 2025-01-09
bicep: Support `@sealed()` decorator on user defined types

### LINKED ISSUE #4659: [Issue] Error when trying to  provision a bicep file using custom types with a `sealed()` decorator

**Output from `azd version`**

azd version 1.11.0 (commit 5b92e0687e1fa96dfc8292f4b900c0c58610b6a5)

**Describe the bug**

When trying to deploy a bicep file that makes use of custom types and the decorator `sealed()` to prevent additional properties, the `azd provision` commands fail with the following error:

`ERROR: initializing provisioning manager: failed unmarshalling arm template from json: json: cannot unmarshal bool into Go struct field ArmTemplateParameterDefinition.definitions.additionalProperties of type azure.ArmTemplateParameterAdditionalProperties`

**To Reproduce**

Run azd provision with a main.bicep file containing the following:

```bicep
@sealed()
type sealedType = {
  test: string
}

output sealedTypeOutput sealedType = {
  test: 'test'
}

```

If you remove the `sealed` decorator, the command will proceed as expected

**Expected behavior**
The ARM JSON `"additionalProperties": false` is supported so that users can make use of the `sealed()` decorator

**Environment**
    * Bicep v0.32
    * IDE and version : VS code 1.96.2

**Additional context**

ellismg:
When the `@sealed()` decorator is applied on user defined types, it causes an `additionalProperties: false` property to be added to the type defintition in the generated arm deployment. Our unmarshalling logic did not understand this pattern and so it would cause an error when trying to unmarshall this, since it expected if `additionalProperties` was set, it would be to set to an object that described information about the additional properties.

I've updated the marshalling code to support `false` as a valid value for the additional properties node when parsing an ARM template. Since Go doesn't have great support for modeling descrimited unions, the new `ArmTemplateParameterAdditionalPropertiesValue` type is used to provide the logcal descrimited union of `false` or a object like value.

Fixes #4659

### DIFF
--- a/cli/azd/pkg/infra/provisioning/bicep/bicep_provider_test.go
+++ b/cli/azd/pkg/infra/provisioning/bicep/bicep_provider_test.go
@@ -1032,9 +1032,10 @@ func TestUserDefinedTypes(t *testing.T) {
 		},
 		objectParam.Properties)
 	require.NotNil(t, objectParam.AdditionalProperties)
+	require.True(t, objectParam.AdditionalProperties.HasAdditionalProperties())
 	require.Equal(
 		t,
-		azure.ArmTemplateParameterAdditionalProperties{
+		azure.ArmTemplateParameterAdditionalPropertiesProperties{
 			Type:      "string",
 			MinLength: to.Ptr(10),
 			Metadata: map[string]json.RawMessage{
@@ -1042,7 +1043,7 @@ func TestUserDefinedTypes(t *testing.T) {
 				"fromDefinitionBar": []byte(`"bar"`),
 			},
 		},
-		objectParam.AdditionalProperties)
+		objectParam.AdditionalProperties.Properties())
 	require.NotNil(t, objectParam.Metadata)
 	require.Equal(
 		t,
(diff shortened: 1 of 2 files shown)


cli/azd/pkg/infra/provisioning/bicep/bicep_provider_test.go
```diff
@@ -1211,6 +1227,18 @@ const userDefinedParamsSample = `{
 			"fromDefinitionFoo": "foo",
 			"fromDefinitionBar": "bar"
 		}
+	  },
```
ellismg, line 1230: I just edited the test template here by hand, @vhvb1989, but if there's some corresponding bicep file that we had that you also want me to go update, just let me know.

cli/azd/pkg/infra/provisioning/bicep/bicep_provider_test.go
```diff
@@ -1032,17 +1032,18 @@ func TestUserDefinedTypes(t *testing.T) {
 		},
 		objectParam.Properties)
 	require.NotNil(t, objectParam.AdditionalProperties)
```
ellismg, line 1034: Just to note, since previously AdditionalProperties was not typed as a `*` value, I think this was always returning true, since the default value would be the zero value which was non nil.

In my change, I made this a pointer type.

ellismg (commented):
You have more context in this part of the code than I do, @vhvb1989, so you'll have a better sense of if I did the right thing here or not.

I'm not sure if there's a case where they can emit `additionalProperties` as `true` (I think this is allowed by JSON Schema, but I'm not sure if bicep ever emits this), but it wasn't clear exactly to me what we should do when `AdditionalProperties` was `true`.  I think in practice we don't have to analyze this part of the document, just make sure as we transform it we keep the text the user provided.  I could imagine changing `ArmTemplateParameterAdditionalPropertiesValue` to also have a `*bool` field, with the idea that either it or the `*ArmTemplateParameterAdditionalPropertiesProperties` is always set (the bool is set to true or false if it was in the document, otherwise the value was an object and the `props` field is set), and we also modify `MarshalJSON` to emit `true` when needed.

vhvb1989 (approved):
LGTM

PR #4690, opened 2025-01-10
Use NET SDK with no Aspire workload in auto-gen pipelines

vhvb1989:
fix: https://github.com/Azure/azure-dev/issues/4691

### DIFF
--- a/cli/azd/resources/pipeline/.azdo/azure-dev.ymlt
+++ b/cli/azd/resources/pipeline/.azdo/azure-dev.ymlt
@@ -25,13 +25,15 @@ steps:
   - pwsh: |
       azd config set auth.useAzCliAuth "true"
     displayName: Configure AZD to Use AZ CLI Authentication.  
-{{- if .InstallDotNetAspire}}
-  - task: Bash@3
-    displayName: Install .NET Aspire workload
+{{- if .InstallDotNetForAspire}}
+  - task: UseDotNet@2
     inputs:
-      targetType: 'inline'
-      script: |
-        dotnet workload install aspire
+      version: '8.x'
+    displayName: Set up .NET 8
+  - task: UseDotNet@2
+    inputs:
+      version: '9.x'
+    displayName: Set up .NET 9
 {{ end }}
   - task: AzureCLI@2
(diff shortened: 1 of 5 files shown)


cli/azd/resources/pipeline/.azdo/azure-dev.ymlt
```diff
@@ -25,13 +25,15 @@ steps:
...
     displayName: Configure AZD to Use AZ CLI Authentication.  
-{{- if .InstallDotNetAspire}}
-  - task: Bash@3
-    displayName: Install .NET Aspire workload
+{{- if .InstallDotNetForAspire}}
+  - task: UseDotNet@2
     inputs:
-      targetType: 'inline'
-      script: |
-        dotnet workload install aspire
```
ellismg, line 34: I know the workload was removed for Aspire 9, but for Aspire 8, it's still present.  Do we need to do with `dotnet workload install` still?  Given that we still install .NET 8, I suspect we are also concerned about support Aspire 8 here (which makes sense, for projects that haven't move to nine)

ellismg (commented):
I think we might still want to install the workload?

cli/azd/resources/pipeline/.azdo/azure-dev.ymlt
vhvb1989: My understanding is that folks are asked to move to Aspire9 instead of bringing fixes for Aspire8.  I don't think we should bring tests for both in azd.
If customers open a bug for latest azd + Aspire8, I think we should ask them to move to Aspire9.

ellismg: > My understanding is that folks are asked to move to Aspire9 instead of bringing fixes for Aspire8. I don't think we should bring tests for both in azd.

Makes sense to me - I suppose if a customer wants to stay on Aspire 8, they just need to edit the generated `azure-dev.yaml` pipeline to install the workload, which is something they can do since it is under their control.

ellismg (approved):
LGTM - The baselines need to be updated and I think that will address the test failures.

PR #4696, opened 2025-01-14
Extensions Framework: Adds v2 UX Components

wbreza:
Adds v2 UX components used within extension framework experiences.  
Components are built upon a canvas rendering strategy that support inline re-rendering of components as needed.

### Prompt

Prompts a user for a single piece of data.  Includes formatting and validation and automatic user feedback.
image

### Confirm

Prompts a user to confirm (yes/no) some operation

### Select

Prompt a user to select a value from a list. Includes filtering and paging of long data.

### Spinner

Lightweight spinner component

### Task List

Shows a series of tasks that are in progress with automatic reporting of status.  Support multiple async operations

### DIFF
--- a/cli/azd/pkg/common/detailed_error.go
+++ b/cli/azd/pkg/common/detailed_error.go
@@ -1,3 +1,6 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
 package common
 
 import "fmt"
--- a/cli/azd/pkg/common/detailed_error.go
+++ b/cli/azd/pkg/common/detailed_error.go
@@ -1,0 +1,28 @@
+package common
+
+import "fmt"
+
+type DetailedError struct {
+	description string
+	err         error
+}
+
+func (e *DetailedError) Error() string {
+	return fmt.Sprintf("%s\n\nDetails:\n%s", e.description, e.err.Error())
+}
+
+func (e *DetailedError) Unwrap() error {
+	return e.err
+}
+
+func (e *DetailedError) Description() string {
+	return e.description
+}
+
+// Factory function to create a new DetailedError
+func NewDetailedError(description string, err error) *DetailedError {
+	return &DetailedError{
+		description: description,
+		err:         err,
+	}
+}
--- a/cli/azd/pkg/ux/canvas.go
+++ b/cli/azd/pkg/ux/canvas.go
@@ -1,0 +1,94 @@
+package ux
+
+import (
+	"io"
+	"sync"
+)
+
+// Canvas is a base component for UX components that require a canvas for rendering.
+type canvas struct {
+	visuals    []Visual
+	printer    Printer
+	writer     io.Writer
+	renderMap  map[Visual]*VisualContext
+	updateLock sync.Mutex
+}
+
+type Canvas interface {
+	Run() error
+	Update() error
+	WithWriter(writer io.Writer) Canvas
+}
+
+// NewCanvas creates a new Canvas instance.
+func NewCanvas(visuals ...Visual) Canvas {
+	canvas := &canvas{
+		visuals:   visuals,
+		renderMap: make(map[Visual]*VisualContext),
+	}
+
+	return canvas
+}
+
+// WithWriter sets the writer for the canvas.
+func (c *canvas) WithWriter(writer io.Writer) Canvas {
+	c.writer = writer
+	return c
+}
+
+// Run starts the canvas.
+func (c *canvas) Run() error {
+	c.printer = NewPrinter(c.writer)
+	return c.Update()
+}
+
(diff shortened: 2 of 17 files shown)


cli/azd/pkg/common/detailed_error.go
```diff
@@ -0,0 +1,28 @@
+package common
```
vhvb1989, line 1:
```suggestion
// Copyright (c) Microsoft Corporation. All rights reserved.
// Licensed under the MIT License.

package common
```

cli/azd/pkg/ux/select.go
```diff
@@ -0,0 +1,424 @@
...
+	if v2 {
+		p.renderMessage2(printer)
+		indent = "  "
+	} else {
+		p.renderMessage1(printer)
+	}
+
+	if p.complete || p.cancelled {
+		return nil
+	}
```
vhvb1989, line 399: is v2 ever `false` ?  Should you remove `renderMessage1()` ?

cli/azd/pkg/ux/confirm.go
```diff
@@ -0,0 +1,259 @@
...
+// NewConfirm creates a new Confirm instance.
+func NewConfirm(options *ConfirmOptions) *Confirm {
+	mergedOptions := ConfirmOptions{}
+	if err := mergo.Merge(&mergedOptions, options, mergo.WithoutDereference); err != nil {
+		panic(err)
+	}
+
+	if err := mergo.Merge(&mergedOptions, DefaultConfirmOptions, mergo.WithoutDereference); err != nil {
+		panic(err)
+	}
```
vhvb1989, line 65: I don't think we need to bring the mergo library for what we need here. 
Since the options arg is a pointer, we just need to either use it (when different than nil) or not (when nil we create and use the default options).

cli/azd/pkg/common/detailed_error.go
weikanglim, line 1: I realized that we did not close on this, but during the discussion in https://github.com/Azure/azure-dev-docs/issues/6, I felt that we were leaning towards avoiding introducing `core` or `common` packages (.NET idiom), but if it's truly a "core helper" to just include it under `internal/` package (go idiom).

cli/azd/pkg/ux/canvas.go
```diff
@@ -0,0 +1,94 @@
...
+	canvas := &canvas{
+		visuals:   visuals,
+		renderMap: make(map[Visual]*VisualContext),
+	}
+
+	return canvas
+}
+
+// WithWriter sets the writer for the canvas.
+func (c *canvas) WithWriter(writer io.Writer) Canvas {
```
weikanglim, line 34: Currently `WithWriter` is used to set `c.writer`, but `c.writer` is only used when `Run()` is called. The effect is that calling `WithWriter` after `Update()` doesn't actually change any behavior -- 

does it make more sense to:
1.  Have `Writer` be part of the constructor, or
2. Document this ineffectual behavior?

cli/azd/pkg/ux/canvas.go
```diff
@@ -0,0 +1,94 @@
...
+	c.printer = NewPrinter(c.writer)
+	return c.Update()
+}
+
+// Update updates the canvas to force a re-render.
+func (c *canvas) Update() error {
+	c.updateLock.Lock()
+	defer c.updateLock.Unlock()
+
+	if c.printer == nil {
```
weikanglim, line 50: Just wondering, what does the `updateLock` protect here?

At a glance, it's not immediately clear what the actual critical resource we're trying to protect here, and if this lock is necessary.

FWIW, it's really hard to reason about critical resources here since we have separated components based on abstract responsibilities rather than control flow. I do prefer when the underlying implementation is immediately clear, such as in bubbletea's [renderer](https://github.com/charmbracelet/bubbletea/blob/main/standard_renderer.go).

wbreza: The `updateLock` prevents multiple re-renders of a component at the same time.

cli/azd/pkg/ux/canvas.go
wbreza: The canvas component and the `Run` function are used internally by each UX component.  
If the component is not rendering inside a larger multi-component canvas it creates its own internal canvas for the main rendering loop out to the default writer.

cli/azd/pkg/ux/confirm.go
wbreza: The goal here was to support the user on only setting the options that they really need and allowing the remaining options to be set using the component defaults vs them having to manually constructing a fully valid options object.

I've taken the approach of making any optional options accept `nil` values to help support this concept.

vhvb1989 (approved):

PR #4698, opened 2025-01-15
chore: `init` to depend on `add.Configure`

weikanglim:
This change reshares UI prompting logic between `init` and `add`.

**Context**
`init` logically depends on `add` since it is essentially a quick wizard that ends up prompting attributes for each resource. In the limit, `init` can be thought of as a series of `add` calls, though this is not reflected in practice due to difference in UX design considerations.

This is a pure refactor and does not introduce any functional changes. The UI prompting is covered by existing tests.

### DIFF
--- a/cli/azd/internal/cmd/add/add.go
+++ b/cli/azd/internal/cmd/add/add.go
@@ -15,7 +15,6 @@ import (
 
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/arm"
 	"github.com/azure/azure-dev/cli/azd/cmd/actions"
-	"github.com/azure/azure-dev/cli/azd/internal/repository"
 	"github.com/azure/azure-dev/cli/azd/pkg/account"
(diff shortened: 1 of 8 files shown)


rajeshkamal5050:
I see, so previously `azd init` was detecting and doing infra/app building. Now with `azd add` we are making `azd init` use the new layer?

rajeshkamal5050 (approved):

ellismg (approved):

weikanglim:
@rajeshkamal5050 Thanks for asking! The code sharing is mostly in the prompting logic. In my mind, both packages are app-level packages, and I'm avoiding introducing additional layers or packages for "resharing prompting logic" right now, we'll break it down later if needed.

We did reshare library components between them (such as the `appdetect` library), but the main thing I do want shared across is this idea of a "Configure" phase for each resource/service, which is asking for user input for a given resource.

vhvb1989 (approved):

PR #4703, opened 2025-01-17
apply install script in azdo setup-azd

### LINKED ISSUE #4631: Use install scripts in devops and github actions extensions
These extensions currently have hardcoded hostnames for downloading azd. These should instead use install scripts instead of duplicating the behavior of the install scripts.

hemarina:
- fix #4631 
- tested in https://dev.azure.com/hemarina0907/hemarina/_build/results?buildId=54&view=results

### DIFF
--- a/ext/azuredevops/CONTRIBUTING.md
+++ b/ext/azuredevops/CONTRIBUTING.md
@@ -31,6 +31,14 @@ behalf.
 
 From `setupAzd` folder, run `npm test`
 
+### Testing in Azure Devops using private extension
+
+- Update `publisher` to you test use case and change `galleryFlags` to `private` in `vss-extension.json`.
+- Use a new `id` in `setupAzd/task.json`.
+- Run the `build` steps to produce the `vsix` release artifact.
+- Publish it to the Marketplace in private mode and share it with your test organization. 
+- Go to you test organization and install the shared extension. Then, you could run the private extension for testing. 
+
 ## Release
 
 - Update `setupAzd/task.json` with the `version` number.
--- a/ext/azuredevops/CONTRIBUTING.md
+++ b/ext/azuredevops/CONTRIBUTING.md
@@ -33,11 +33,11 @@ From `setupAzd` folder, run `npm test`
 
 ### Testing in Azure Devops using private extension
 
-- Update `publisher` to you test use case and change `galleryFlags` to `private` in `vss-extension.json`.
+- Update `publisher` to your test use case and change `galleryFlags` to `private` in `vss-extension.json`.
 - Use a new `id` in `setupAzd/task.json`.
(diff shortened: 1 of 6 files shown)


ext/azuredevops/CONTRIBUTING.md
```diff
@@ -31,6 +31,14 @@ behalf.
 
 From `setupAzd` folder, run `npm test`
 
+### Testing in Azure Devops using private extension
+
+- Update `publisher` to you test use case and change `galleryFlags` to `private` in `vss-extension.json`.
```
JeffreyCA, line 36:
```suggestion
- Update `publisher` to your test use case and change `galleryFlags` to `private` in `vss-extension.json`.
```

ext/azuredevops/CONTRIBUTING.md
```diff
@@ -31,6 +31,14 @@ behalf.
...
+- Use a new `id` in `setupAzd/task.json`.
+- Run the `build` steps to produce the `vsix` release artifact.
+- Publish it to the Marketplace in private mode and share it with your test organization. 
+- Go to you test organization and install the shared extension. Then, you could run the private extension for testing.
```
JeffreyCA, line 40:
```suggestion
- Go to your test organization and install the shared extension. Then, you could run the private extension for testing. 
```

ext/azuredevops/setupAzd/index.ts
JeffreyCA, line 1: Some lines end with semicolons and some don't - is there a particular convention we use? Could we make that consistent throughout?

vhvb1989 (approved):

ext/azuredevops/setupAzd/tests/InstallSuccess.ts
```diff
@@ -1,7 +1,7 @@
 import * as path from 'path';
 import * as assert from 'assert';
 import * as ttm from 'azure-pipelines-task-lib/mock-test';
-import * as fs from 'fs'
+import * as fs from 'fs';
 
 describe('setup azd tests', function () {
     setTimeout(() => { }, 10000);
```
JeffreyCA, line 7: Is this supposed to control the test timeout? If so we may need to use `this.timeout(...)`: https://mochajs.org/#timeouts

JeffreyCA: Usually `setTimeout` is used to delay the execution of the function (first argument), but here it's empty: `() => { }` which I don't quite understand - in the documentation I think `setTimeout` is used to simulate a function that takes `x` milliseconds to complete

JeffreyCA: Glad that worked!

Issue #4653, opened 2024-12-26
[Issue] Failed to deploy app when app name too long or different app name has same long prefix


rujche:
Link to sample project: https://github.com/microsoft/SJAD

## Reproduce step:
1. Run `azd init`, then manually change the app name in `azure.yaml`. (It can be achieved by another way: Rename the folder name then run `azd init`)
3. Run `azd up`

### 1.1. Example `azure.yaml`:
```yaml
# yaml-language-server: $schema=https://raw.githubusercontent.com/azure-javaee/azure-dev/feature/sjad/schemas/alpha/azure.yaml.json

name: sjad
metadata:
    template: azd-init@0.0.0-dev.0
services:
    long-long-long-long-long-long-long-long-long-common-prefix-api:
        parentPath: C:\Users\rujche\Work\git-workspace\main\SJAD
        project: api
        host: containerapp
        language: java
        docker:
            path: Dockerfile
    long-long-long-long-long-long-long-long-long-common-prefix-email-service:
        parentPath: C:\Users\rujche\Work\git-workspace\main\SJAD
        project: email-service
        host: containerapp
        language: java
        docker:
            path: Dockerfile
    long-long-long-long-long-long-long-long-long-common-prefix-web:
        parentPath: C:\Users\rujche\Work\git-workspace\main\SJAD
        project: web
        host: containerapp
        language: java
        docker:
            path: Dockerfile
hooks:
    prepackage:
        windows:
            shell: pwsh
            run: .\mvnw.cmd clean package -DskipTests
        posix:
            shell: sh
            run: ./mvnw clean package -DskipTests
resources:
    long-long-long-long-long-long-long-long-long-common-prefix-api:
        type: host.containerapp
        uses:
            - mongo
            - servicebus
        port: 3100
    long-long-long-long-long-long-long-long-long-common-prefix-email-service:
        type: host.containerapp
        uses:
            - servicebus
        port: 8081
    long-long-long-long-long-long-long-long-long-common-prefix-web:
        type: host.containerapp
        uses:
            - long-long-long-long-long-long-long-long-long-common-prefix-api
            - long-long-long-long-long-long-long-long-long-common-prefix-email-service
        port: 8080
    mongo:
        type: db.mongo
        databaseName: todo
    servicebus:
        type: messaging.servicebus
        authType: userAssignedManagedIdentity
        isJms: true
```

### 1.2. Error message:
```shell
[...]

PR #4455, opened 2024-10-17
easy-init: Switch to avm (Azure Verified Modules)

weikanglim:
With this change, we switch easy-init's generated infrastructure to use avm (Azure Verified Modules).

Since AVM allows us to express resource definitions in a more condensed form, we have a single resources.bicep that uses all the different modules available, similar to Aspire's generated Bicep.

Notable behavioral changes with this PR:

**Postgres**
- Server version upgraded from version 13 to version 15.
- Storage size reduced from 128GB to 32GB (AVM default).

**MongoDB**
 - Server version upgraded from 4.0 to 4.2.

**Redis**
- Redis is now implemented as an Azure Cache for Redis instance instead of previously as a container app image.

Standardized environment variables:
- POSTGRES_URL, MONGODB_URL, REDIS_URL all serves as the full connection string URI to the respective instances.

Contributes to: https://github.com/Azure/azure-dev-pr/issues/1682

PR #4707, opened 2025-01-17
New feature: Update spring boot properties file for postgresql

rujche:
## 1. Description
New feature: Update spring boot properties file for postgresql.

## 2. Try this feature

1. Link to sample project: https://github.com/azure-javaee/azure-spring-boot-samples/blob/788cc9a14f3fa2bfa826b0e875ba6dd729726dfa/postgresql/spring-cloud-azure-starter-jdbc-postgresql/spring-cloud-azure-postgresql-sample/pom.xml#L39

2. Screenshot before running `azd init`:
>image

3. Screenshot after running `azd init`:
> image

4. Screenshot that confirm that sample project runs well on cloud:
> image

5. About the `todo`s in current PR, my current plan is do them after current PR merged.


rujche:
Hi, @saragluna , @weikanglim 
Please help to review this PR.

cli/azd/resources/scaffold/templates/resources.bicept
```diff
@@ -217,7 +217,7 @@ var {{bicepName .Name}}Env = map(filter({{bicepName .Name}}AppSettingsArray, i =
 module {{bicepName .Name}} 'br/public:avm/res/app/container-app:0.8.0' = {
   name: '{{bicepName .Name}}'
   params: {
-    name: '{{.Name}}'
+    name: '{{containerAppName .Name}}'
```
weikanglim, line 220: Curious about the change here -- we did intentionally move away from `containerAppName .Name` to `.Name` in #4434

rujche: > we did intentionally move away from containerAppName .Name to .Name in https://github.com/Azure/azure-dev/pull/4434

1. Do you mean change from `.Name` to `containerAppName  .Name` in https://github.com/Azure/azure-dev/pull/4434 ?

2. In think the mistake (missing `containerAppName`) is added in this PR: https://github.com/Azure/azure-dev/pull/4455/files

> image

> image

weikanglim: I see. I think when #4455 was made, we also started adding upfront name validation in `azd add` (UI layer, not at the schema layer).

Keeping it as `.Name` instead of `containerAppName .Name` (which does the truncation) transformation may help with #4653, so I think `.Name` still looks better directionally. What do you think?

rujche: How about doing like this:

1. In current PR, keep using `containerAppName .Name` because it can solve the problem appeared in the sample project for current PR: https://github.com/azure-javaee/azure-spring-boot-samples/blob/788cc9a14f3fa2bfa826b0e875ba6dd729726dfa/postgresql/spring-cloud-azure-starter-jdbc-postgresql/spring-cloud-azure-postgresql-sample/pom.xml#L39
2. For https://github.com/Azure/azure-dev/issues/4653, we can fix it in another PR. In that PR, the same name prefix should be considered,

cli/azd/internal/appdetect/spring_boot_properties_file.go
```diff
@@ -0,0 +1,193 @@
...
+	"strings"
+)
+
+type property struct {
+	key   string
+	value string
+}
+type propertyMergeFunc func(string, string) string
+
+const azureProfileName = "azure"
```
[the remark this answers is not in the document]
rujche: Good point. 
In this case, original value of `spring.profiles.active` will not be changed. See these test cases:
> image

And here is related source code:
> image

cli/azd/internal/appdetect/spring_boot_properties_file.go
```diff
@@ -0,0 +1,193 @@
...
+
+// todo: manage following placeholders and together with resources.bicept.
+
+const placeholderPostgresHost = "${POSTGRES_HOST}"
+const placeholderPostgresPort = "${POSTGRES_PORT}"
+const placeholderPostgresDatabase = "${POSTGRES_DATABASE}"
+const placeholderPostgresUsername = "${POSTGRES_USERNAME}"
```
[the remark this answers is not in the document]
rujche: Good point. Updated in this commit: https://github.com/Azure/azure-dev/pull/4707/commits/cdc9acddd069570d8841a7fb4e2fb1046e34baa9

rujche:
Hi, @saragluna , @weikanglim 
PR updated. Please help to review again.

rujche:
Hi, @saragluna , @weikanglim .
PR updated. Please help to review again.

cli/azd/internal/appdetect/appdetect.go
```diff
@@ -103,6 +103,21 @@ func (f Dependency) IsWebUIFramework() bool {
 	return false
 }
 
+type RawDependency struct {
```
weikanglim, line 106: If it's easier, we can just return the RawProject pom.xml representation and let the "app generation" part decide what to do with that information. I think we want to avoid having additional schema translations

cli/azd/internal/repository/app_init.go
```diff
@@ -467,3 +495,179 @@ func (i *Initializer) prjConfigFromDetect(
...
+	for _, n := range a {
+		if x == n {
+			return true
+		}
+	}
+	return false
+}
+
+func updatePropertyFile(filePath string, newProperties []property, function propertyMergeFunc) error {
+	err := createFileIfNotExist(filePath)
```
weikanglim, line 575: Would it be simpler to just have a single function that opens the file, and just does a one-pass update? 

I know that we've created a lot of functions here to attempt to break down the complexity. In doing so, they all "look right" individually, but if we zoom out to see the whole algorithm: 

- If `createFileNotExists` succeeds, but we fail in `readProperties`, we create an empty file
- We also create the file again in `writeProperties`

FWIW, this is related to what we had talked about creating too many functions.

cli/azd/internal/repository/app_init.go
```diff
@@ -290,6 +291,35 @@ func (i *Initializer) InitFromApp(
...
+			}
+		}
+		if !isSpringBootProject {
+			continue
+		}
+		for _, dep := range mavenProject.Dependencies {
+			name := dep.GroupId + ":" + dep.ArtifactId
+			switch name {
+			case "org.postgresql:postgresql",
+				"com.azure.spring:spring-cloud-azure-starter-jdbc-postgresql":
```
weikanglim, line 314: Is it possible for both these dependencies to exist at the same time, and if so -- do we only want to generate once?

rujche:
Closing this PR because now SJAD team plan to create azure.yaml by AI-based method. Similarity, application.properties will be updated by AI-based method, too.

PR #4710, opened 2025-01-20
Use effective pom to analyze pom.xml

rujche:
Try to use effective pom if possible (both mvn and java command exists)


cli/azd/internal/appdetect/pom.go
```diff
@@ -0,0 +1,94 @@
...
+	result, err := toEffectivePom(ctx, mvnCli, pomFilePath)
+	if err == nil {
+		result.path = filepath.Dir(pomFilePath)
+		return result, nil
+	}
+
+	result, err = unmarshalPomFile(pomFilePath)
+	if err == nil {
+		result.path = filepath.Dir(pomFilePath)
+		// todo: handle pom, for example: <version>${project.version}<version>
```
rujche, line 23: Hi, @weikanglim 

1. My current plan is do this after current PR merged.
2. Anyway, analyze the unmarshalled pom will be done. If you think `toSimulatedEffectivePom` is not accurate, I can choose another func name, for example: `pom := analyzeUnmarshalledPom()`

rujche:
Hi, @saragluna , @haoozhang, @weikanglim 
Please help to review this PR.

BTW, I can't set @haoozhang as reviewer. @weikanglim , could you please help to set this?

cli/azd/internal/appdetect/pom.go
weikanglim: Previous comment here: https://github.com/Azure/azure-dev/pull/4473#discussion_r1921708448

I still strongly suggest we do not do this. A user is asking `azd` to scan a Java project that has `pom.xml`. The `pom.xml` clearly requires maven tooling installed to run. User also needs maven tooling installed to run `azd package` or `azd up`. The overall scenario requires the user to install maven.

I don't like the idea of trying to replicate maven evaluation because evaluation rules can be tricky to replicate; and I prefer tools that just work exactly, instead of tools that are *subtly wrong*. This can lead to downstream impacts that are harder to debug.

On the flip side, if we are confident in replicating [maven evaluation rules](https://maven.apache.org/guides/introduction/introduction-to-the-pom.html#Available_Variables), I would prefer removing the current effective-POM evaluation and have the native-go implementation take over -- it's cool work to have a 100% compatible maven evaluator in `go`, but I'm not confident that we can always replicate it since it's not a fully documented standard -- and it's one that changes over time.

cli/azd/pkg/tools/maven/maven.go
```diff
@@ -237,6 +238,48 @@ func (cli *Cli) GetProperty(ctx context.Context, propertyPath string, projectPat
...
+func (cli *Cli) EffectivePom(ctx context.Context, pomPath string) (string, error) {
+	mvnCmd, err := cli.mvnCmd()
+	if err != nil {
+		return "", err
+	}
+	pomDir := filepath.Dir(pomPath)
+	runArgs := exec.NewRunArgs(mvnCmd, "help:effective-pom", "-f", pomPath).WithCwd(pomDir)
+	result, err := cli.commandRunner.Run(ctx, runArgs)
+	if err != nil {
+		return "", fmt.Errorf("failed to run mvn help:effective-pom for pom file: %s. error = %w", pomPath, err)
```
weikanglim, line 250:
```suggestion
		return "", fmt.Errorf("mvn help:effective-pom on project '%s' failed: %w", pomPath, err)
```

cli/azd/pkg/tools/maven/maven.go
```diff
@@ -237,6 +238,48 @@ func (cli *Cli) GetProperty(ctx context.Context, propertyPath string, projectPat
...
+			break
+		}
+		if inProject {
+			if strings.TrimSpace(line) != "" {
+				builder.WriteString(line)
+			}
+		}
+	}
+	if err := scanner.Err(); err != nil {
+		return "", fmt.Errorf("failed to scan console output. %w", err)
```
weikanglim, line 274:
```suggestion
		return "", fmt.Errorf("failed to scan console output: %w", err)
```

cli/azd/pkg/tools/maven/maven.go
```diff
@@ -237,6 +238,48 @@ func (cli *Cli) GetProperty(ctx context.Context, propertyPath string, projectPat
...
+		} else if strings.HasPrefix(strings.TrimSpace(line), "</project>") {
+			builder.WriteString(line)
+			break
+		}
+		if inProject {
+			if strings.TrimSpace(line) != "" {
+				builder.WriteString(line)
+			}
+		}
+	}
```
weikanglim, line 272: a small suggestion here to avoid string allocations and streamline the code

```suggestion
	inProject := false
	projectStart := regexp.MustCompile(`\s+<project`)
	projectEnd := regexp.MustCompile(`\s+</project>`)

	for scanner.Scan() {
		line := scanner.Text()
		if projectStart.MatchString(line) {
			inProject = true
		} else if projectEnd.MatchString(line) {
			builder.WriteString(line)
			break
		}
		if inProject {
		  builder.WriteString(line)
		}
	}
```

cli/azd/internal/appdetect/pom.go
```diff
@@ -0,0 +1,94 @@
+package appdetect
```
weikanglim, line 1: As a general `go` programming convention here, I would like to avoid adding more files like `maven_project.go` or 
`pom.go` here.  We can fit all under `java.go`. In the future, we could rename this to `java_maven.go` when gradle support is added.

cli/azd/internal/appdetect/pom.go
weikanglim: If we end up deciding removing the maven evaluation, perhaps `pom_test.go` can be removed along those lines as well.

vhvb1989 (approved):

weikanglim (commented):
Finished reviewing. Overall looks good. Left some comments.

I do feel it'd be nice for us to avoid introducing additional types like `pom` instead of `mavenProject`, and keep it simple for now by having the shared logic in `java.go` to start.

cli/azd/internal/appdetect/pom.go
rujche: > If we end up deciding removing the maven evaluation, perhaps pom_test.go can be removed along those lines as well.

Agree. Now effective pom is generated by `mvn help:effective-pom`, it's not necessary to test it. I deleted `pom_test.go`.

weikanglim: > Do you think it's necessary to exit analyze and ask customer to install mvn?

Yes, I think it's better than doing a partial analysis

cli/azd/pkg/tools/maven/maven.go
rujche: FYI:
1. I'll accept the changes by clicking `commit suggestion`.
2. I'll fix the indent problem in line 271 in another commit.
3. To test the changes, make sure it won't bring new bug, I'll recover `pom_test.go`.

rujche: FYI:

> To test the changes, make sure it won't bring new bug, I'll recover pom_test.go.

Test failed.

Here is the screenshot when I debug:

> image

1. I'll fix the bug in another commit.
2. It shows that `pom_test.go` is useful, I'll keep it.

rujche: > I'll fix the bug in another commit.

Done in this commit: https://github.com/Azure/azure-dev/pull/4710/commits/90e806f32ec9c4a74864b29eae8bedb70763b4b9

cli/azd/internal/appdetect/maven_project.go
```diff
@@ -0,0 +1,44 @@
...
+	"maps"
+	"slices"
+	"strings"
+
+	"github.com/azure/azure-dev/cli/azd/pkg/tools/maven"
+)
+
+type mavenProject struct {
+	pom pom
+}
```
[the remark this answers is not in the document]
rujche: I'm OK to delete mavenProject and use pom directly. 

Hi, @saragluna , This option is required by you before. Now if you agree, I'll delete mavenProject and use pom directly.

cli/azd/internal/appdetect/pom.go
rujche: > As a general go programming convention here

1. Thanks for telling me that. I didn't know this convention before. BTW, could you please share related link about this convention? I guess there should be some official doc like this one: https://go.dev/doc/effective_go
2. I use different files because:
  2.1. `java` and `pom` are different things in concept.
  2.2. Separate codes into multiple file can avoid one file too long. Too long file will make it hard to maintain.

rujche:
> Finished reviewing. Overall looks good. Left some comments.
> 
> I do feel it'd be nice for us to avoid introducing additional types like `pom` instead of `mavenProject`, and keep it simple for now by having the shared logic in `java.go` to start.

@weikanglim 
I replied here: https://github.com/Azure/azure-dev/pull/4710#discussion_r1926278884

cli/azd/internal/appdetect/pom.go
rujche: Updated in this commit: https://github.com/Azure/azure-dev/pull/4710/commits/bfab5a31bb5765863d2185d5c423822b80986767

cli/azd/pkg/tools/maven/maven.go
rujche: @weikanglim , @saragluna 

FYI: Here is the current output when `mvn` command not exist:

> image

cli/azd/internal/appdetect/maven_project.go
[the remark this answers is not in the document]
saragluna: Yes, you could go ahead please. But what I meant (insisted) is we should call this concept a maven project, instead of just a pom, which conceptually should be under the maven project.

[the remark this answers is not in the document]
rujche: To make the PR been merged as soon as possible, I'll revert all refactor-related changes, just keep necessary changes.

rujche:
Closing current PR. Let's use this one instead: https://github.com/Azure/azure-dev/pull/4729

PR #4719, opened 2025-01-22
Use Redis AVM's `secretsExportConfiguration` param to store secrets in Key Vault

### LINKED ISSUE #4681: compose: investigate moving `set-redis-conn.bicep` logic to upstream AVM
We currently rely on our own custom Bicep module [`set-redis-conn.bicep`](https://github.com/Azure/azure-dev/blob/main/cli/azd/resources/scaffold/base/modules/set-redis-conn.bicep) to write the Redis access key and connection strings to AKV after the cache has been provisioned.

We may want to move this logic upstream if possible, by having the Redis AVM support [Secrets export](https://azure.github.io/Azure-Verified-Modules/specs/bcp/res/interfaces/#secrets-export), which would expose a standardized input param called `secretsExportConfiguration`.

We currently leverage this param for Cosmos DB:
https://github.com/Azure/azure-dev/blob/ccbf5e7fd4f20e7ffdcacfa4ef2f9d34ff8648eb/cli/azd/resources/scaffold/templates/resources.bicept#L94-L97

That way, as we add support for more services that involve access keys/secrets, they can all use the underlying AVM's `secretsExportConfiguration` to write to AKV.

Other Azure services whose AVM already support secrets export include:

- Cosmos DB
- Storage account
- SQL Server
- Cognitive services

JeffreyCA:
Closes #4681 

[v0.9.0 of the Redis AVM](https://github.com/Azure/bicep-registry-modules/pull/4138) adds support for `secretsExportConfiguration`, a standardized input param used by several other AVMs to export access keys and other secrets to Azure Key Vault after the resource has been provisioned. ([Reference](https://github.com/Azure/bicep-registry-modules/tree/main/avm/res/cache/redis#example-3-deploying-with-a-key-vault-reference-to-save-secrets))

This PR updates compose's Bicep generation to use the new module version and  `secretsExportConfiguration` to write the Redis primary access key and connection string to Key Vault instead of our custom `set-redis-conn.bicep` Bicep module.

### Validation

Deployed a minimal Python Flask web app with Redis:

**azure.yaml:**

```yml
name: new
resources:
  redis:
    type: db.redis
  pyservice:
    type: host.containerapp
    uses:
      - redis
    port: 5000
services:
  pyservice:
    project: .
    host: containerapp
    language: python
```

**resources.bicep:**

```bicep
...
module redis 'br/public:avm/res/cache/redis:0.9.0' = {
  name: 'redisDeployment'
  params: {
    // Required parameters
    name: '${abbrs.cacheRedis}${resourceToken}'
    // Non-required parameters
    location: location
    skuName: 'Basic'
    secretsExportConfiguration: {
      keyVaultResourceId: keyVault.outputs.resourceId
      primaryAccessKeyName: 'REDIS-PASSWORD'
      primaryConnectionStringName: 'REDIS-URL'
    }
  }
}
...
```

**Python app code:**

```py
import os
...
redis_host = os.environ.get('REDIS_HOST')
redis_port = os.environ.get('REDIS_PORT')
redis_url = os.environ.get('REDIS_URL')
redis_password = os.environ.get('REDIS_PASSWORD')

# Using REDIS_URL
r = redis.Redis.from_url(redis_url)
# Using REDIS_PASSWORD
r = redis.Redis(host=redis_host, port=int(redis_port), db=0, password=redis_password, ssl=True)

result = r.get('abc')
return result
```

**Key vault:**
image

image


cli/azd/resources/scaffold/templates/resources.bicept
```diff
@@ -389,28 +389,23 @@ module {{bicepName .Name}} 'br/public:avm/res/app/container-app:0.8.0' = {
 {{- end}}
 
 {{- if .DbRedis}}
-module redis 'br/public:avm/res/cache/redis:0.3.2' = {
+module redis 'br/public:avm/res/cache/redis:0.9.0' = {
```
wbreza, line 392: Do we have a process yet for when/how we upgrade to new versions of AVM modules?

cli/azd/resources/scaffold/templates/resources.bicept
```diff
@@ -389,28 +389,23 @@ module {{bicepName .Name}} 'br/public:avm/res/app/container-app:0.8.0' = {
...
     // Required parameters
     name: '${abbrs.cacheRedis}${resourceToken}'
     // Non-required parameters
     location: location
     skuName: 'Basic'
+    {{- if .Services}}
+    secretsExportConfiguration: {
+      keyVaultResourceId: keyVault.outputs.resourceId
+      primaryAccessKeyName: 'REDIS-PASSWORD'
+      primaryConnectionStringName: 'REDIS-URL'
```
wbreza, line 404: Are these `REDIS-PASSWORD` AND `REDIS-URL` param expected to be environment variables? Or just placeholders for end users to update?

wbreza (commented):
Looks good - just a couple questions.

cli/azd/resources/scaffold/templates/resources.bicept
JeffreyCA: My understanding is `REDIS-PASSWORD` and `REDIS-URL` are the Key Vault secret **names**.

And as part of the generated Bicep we create 2 container app secrets **redis-pass** and **redis-url** that refer to these Key Vault secrets:

image

Which are exposed through the environment variables `REDIS_URL` and `REDIS_PASSWORD`:

image

vhvb1989 (approved):

cli/azd/resources/scaffold/templates/resources.bicept
weikanglim: That's a very good question. Do you have any suggestions? And if there's a need for something solid to be in-place today?

In general, it feels not too different to me than when we update `api-version` fħr any Azure resources, which boils down to when a version upgrade is required for a specific feature. Currently, this is mostly hidden away from the user so it wouldn't matter too much.

weikanglim (approved):
LGTM, thanks for the clean change.

FYI @JeffreyCA, I'm not sure if it's helpful but the tests in `scaffold_test.go` does validate the Bicep syntax of the resulting generation. I also logged #4720 for us to move this up a layer where we can have `azure.yaml` test cases as input, and add scenario tests that way.

cli/azd/resources/scaffold/templates/resources.bicept
JeffreyCA: I'm not sure about this, @weikanglim do you know?

JeffreyCA (commented):
> LGTM, thanks for the clean change.
> 
> FYI @JeffreyCA, I'm not sure if it's helpful but the tests in `scaffold_test.go` does validate the Bicep syntax of the resulting generation. I also logged #4720 for us to move this up a layer where we can have `azure.yaml` test cases as input, and add scenario tests that way. 

Thanks, that's good to know

cli/azd/resources/scaffold/templates/resources.bicept
JeffreyCA: Since this is out of scope of this PR I created #4728 to continue this discussion further

PR #4725, opened 2025-01-23
Migrate storage location

danieljurek:
The old `azuresdkreleasepreview` storage location will be deprecated. Use the new `azuresdkartifacts` with static web hosting.

### DIFF
--- a/eng/pipelines/templates/variables/globals.yml
+++ b/eng/pipelines/templates/variables/globals.yml
@@ -8,6 +8,10 @@ variables:
   ${{ if eq(variables['Build.Repository.Name'], 'Azure/azure-dev') }}:
     publish-storage-location: 'https://azuresdkartifacts.blob.core.windows.net'
     publish-storage-static-host: 'https://azuresdkartifacts.z5.web.core.windows.net/'
+
+    # TODO: Remove these before merge
+    Skip.LiveTest: true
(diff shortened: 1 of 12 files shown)


eng/pipelines/templates/variables/globals.yml
```diff
@@ -6,6 +6,9 @@ variables:
   Codeql.SkipTaskAutoInjection: true
 
   ${{ if eq(variables['Build.Repository.Name'], 'Azure/azure-dev') }}:
-    publish-storage-location: 'https://azuresdkreleasepreview.blob.core.windows.net'
+    publish-storage-location: 'https://azuresdkartifacts.blob.core.windows.net'
+    publish-storage-static-host: 'https://azuresdkartifacts.z5.web.core.windows.net'
+
   ${{ else }}:
-    publish-storage-location: 'https://azdprivatebuilds7bc140ae.blob.core.windows.net'
+    publish-storage-location: 'NOT-SUPPORTED'
```
weshaggard, line 13: Are we sure we don't want to support here?

danieljurek: Container settings preclude this storage account from working in the PR scenario. I think a longer-term solution should go here instead. Likely links to just the executable binaries as hosted in pipeline artifacts. It won't be as smooth as the current PR workflow.

PR #4729, opened 2025-01-24
Use effective to construct mavenProject

rujche:
Use effective to construct mavenProject.


cli/azd/internal/appdetect/testdata/java/pom.xml
```diff
@@ -2,7 +2,6 @@
   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
   xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
   <modelVersion>4.0.0</modelVersion>
-  <parent></parent>
```
rujche, line 5: FYI: Empty parent will cause `mvn effective:pom` failed.

PR #4738, opened 2025-01-28
Extensions Framework: Adds gRPC server implementations

wbreza:
Adds implementations for the server-side implementations.

> [!NOTE]
> Also adds some minor changes to return `ResourceExtended` struct instead of standard `Resource` struct so we include the `kind` property which is used by many Azure resources.

### DIFF
--- a/cli/azd/.vscode/cspell.yaml
+++ b/cli/azd/.vscode/cspell.yaml
@@ -2,6 +2,7 @@ import: ../../../.vscode/cspell.global.yaml
 words:
   - azcloud
   - azdext
+  - azdgrpc
   - azurefd
   - backoff
   - Canonicalize
--- a/cli/azd/.vscode/cspell.yaml
+++ b/cli/azd/.vscode/cspell.yaml
@@ -2,7 +2,6 @@ import: ../../../.vscode/cspell.global.yaml
 words:
   - azcloud
   - azdext
-  - azdgrpc
   - azurefd
   - backoff
   - Canonicalize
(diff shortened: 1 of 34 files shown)


cli/azd/internal/azdgrpc/project_service.go
```diff
@@ -0,0 +1,121 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package azdgrpc
```
vhvb1989, line 4: How do you feel about this package name?  

I wonder is something like `grpcserver` might describe what this package is for.  I think that adding the prefix `azd` to an internal azd package seems redundant.

cli/azd/.vscode/cspell.yaml
```diff
@@ -2,6 +2,7 @@ import: ../../../.vscode/cspell.global.yaml
 words:
   - azcloud
   - azdext
+  - azdgrpc
```
vhvb1989, line 5: do we still need this?

cli/azd/pkg/prompt/prompt_service_test.go
```diff
@@ -0,0 +1,60 @@
...
+				DisplayName:    to.Ptr("Subscription 2"),
+			},
+		}, nil)
+
+	promptService := NewPromptService(authManager, userConfigManager, subscriptionService, resourceService)
+	require.NotNil(t, promptService)
+
+	// selected, err := promptService.PromptSubscription(*mockContext.Context, nil)
+	// require.NoError(t, err)
+	// require.NotNil(t, selected)
```
vhvb1989, line 59: delete?

vhvb1989 (approved):
LGTM

Issue #4742, opened 2025-01-31
compose: messaging - event hubs and service bus


weikanglim:
This issue tracks enablement of initial set of messaging services: Event Hubs and Service Bus.

# Schema

Managed identity is enabled by default.

## Event Hubs

Definition:
```yaml
resources:
  event-hubs
    type: messaging.eventhubs
    hubs:
    - events
```

Binding environment variables:

- `AZURE_EVENT_HUBS_NAME` - Name of the Event Hubs namespace (Java clients)
- `AZURE_EVENT_HUBS_HOST` - Hostname of the Event Hubs namespace (most other languages)

## Service Bus

```yaml
resources:
  service-bus
    type: messaging.servicebus
    queues:
    - myQueue
    topics:
    - myTopic
```

Binding environment variables:

- `AZURE_SERVICE_BUS_NAME` - Name of the Service Hub namespace (Java clients)
- `AZURE_SERVICE_BUS_HOST` - Hostname of the Service Hub namespace (most other languages)

## UI

`azd add`

https://github.com/user-attachments/assets/d7a8a9fb-9acb-4789-953c-810db68a6178

weikanglim:
Completed with #4743

PR #4743, opened 2025-01-31
compose: service bus and event hubs

weikanglim:
Add support for creating Service Bus and Event Hubs.

Addresses #4742


vhvb1989 (approved):
LGTM

schemas/alpha/azure.yaml.json
```diff
@@ -1298,6 +1302,51 @@
...
+                    "items": {
+                        "type": "string"
+                    },
+                    "uniqueItems": true
+                }
+            }
+        },
+        "serviceBusResource": {
+            "type": "object",
+            "description": "An Azure Service Hub namespace.",
```
rajeshkamal5050, line 1326: Service Bus?

rajeshkamal5050 (approved):

wbreza (approved):
Looks good.  Have we thought about ways to make these types of resources available in a more dynamic nature vs the expanding switch statements?

SophCarp:
UI Review for Event Hub: 
image
 - If an input request doesn't end with a `?` it should end in a `:`
 
image
- Line duplication bug

image
- I'm realizing when we list the new environment variables, we tab the `+`, whereas everywhere else the `+` is untabbed. I think it would be best to update to be inline with all the other `+`s

SophCarp:
Ui Comments for Service Bus experience: 

image
- line duplication

image
- add `:` to end of each line (gives distinction between the request and the user input)

puicchan:
1. We should always provide default value for name of event hub and service bus in ( xxx ) in gray.
2. Consider including naming convention in [? for help]
image
3. Error handling when the name I provided is too short is not consistent with error handling for storage account. 
Service bus error immediately while storage account states the issue and asks for better name.

cli/azd/pkg/project/scaffold_gen.go
```diff
@@ -273,6 +291,10 @@ func mapHostUses(
 			backendMapping[use] = res.Name // record the backend -> frontend mapping
 		case ResourceTypeOpenAiModel:
 			svcSpec.AIModels = append(svcSpec.AIModels, scaffold.AIModelReference{Name: use})
+		case ResourceTypeMessagingEventHubs:
+			svcSpec.EventHubs = &scaffold.EventHubs{}
```
saragluna, line 295: This means one project can only configure one Event Hubs or Service Bus.

cli/azd/resources/scaffold/templates/resources.bicept
```diff
@@ -343,6 +414,26 @@ module {{bicepName .Name}} 'br/public:avm/res/app/container-app:0.8.0' = {
...
+          {{- end}}
+          {{- if .ServiceBus}}
+          {
+            name: 'AZURE_SERVICE_BUS_NAME'
+            value: serviceBusNamespace.outputs.name
+          }
+          {
+            name: 'AZURE_SERVICE_BUS_HOST'
+            value: '${serviceBusNamespace.outputs.name}.servicebus.windows.net'
+          }
```
saragluna, line 435: Do we need to expose the resource id at the end of this bicep file?

weikanglim: Yes indeed -- updated

PR #4744, opened 2025-01-31
Support Key Vault references in Bicep and JSON parameter files

### LINKED ISSUE #4706: [Issue] Keyvault references in ARM parameter files are not supported

**Output from `azd version`**

azd version 1.11.0 (commit https://github.com/Azure/azure-dev/commit/5b92e0687e1fa96dfc8292f4b900c0c58610b6a5)

**Describe the bug**
When deploying the infrastructure with a  parameter file making use of [keyvault references](https://learn.microsoft.com/en-us/azure/azure-resource-manager/templates/key-vault-parameter?tabs=azure-cli), azd prompts the user to input a value for the parameter and overrides the value instead of fetching the value from the referenced keyvault + secret name combination.

**To Reproduce**
1. Use azd init with the minimal solution option
2. Add the following lines to your bicep file
```bicep
@secure()
param test_kv string 


output test string = test_kv
```
3. Create a keyvault on Azure and a new secret named "test-kv" with any value inside that keyvault. Also store the id of the keyvault somewhere
4. Modify the parameter file as such
```json
{
    "$schema": "https://schema.management.azure.com/schemas/2019-04-01/deploymentParameters.json#",
    "contentVersion": "1.0.0.0",
    "parameters": {
      "environmentName": {
        "value": "${AZURE_ENV_NAME}"
      },
      "location": {
        "value": "${AZURE_LOCATION}"
      },
      "test_kv": {
        "reference": {
          "keyVault": {
            "id": "keyvault id your stored at the previous step"
          },
          "secretName": "test-kv"
        }
      }
    }
}
```
5. Run azd provision
6. You will be prompted for a value for the `test_kv` parameter. Enter a value different from what is in the keyvault
7. After the deployment, check the deployment output in the portal, the output value `test` reuses the value provided at step 6 instead of the value of the keyvault set at step 3

**Expected behavior**
Keyvault references in JSON parameter files are supported and treated the same way as parameters defined with the `value` property, instead of being hijacked by the AZD cli

**Environment**
* Bicep v0.32
* IDE and version : VS code 1.96.2

**Additional context**
The error caused when using the `--no-prompt` switch is also unclear but covered by the issue https://github.com/Azure/azure-dev/issues/3795

JeffreyCA:
Fixes #4706

This PR adds support for [Key Vault references in parameter files](https://learn.microsoft.com/en-us/azure/azure-resource-manager/bicep/key-vault-parameter?tabs=azure-cli#retrieve-secrets-in-parameters-file), so you can now define a `param` in `main.bicep` that has a key vault reference (instead of a simple value) in `main.bicepparam` or `main.parameters.json`.

The issue was that when deserializing the `main.parameters.json` file, the model type used did not have the necessary fields for key vault references. As a result, the `Value` field would be `nil` and cause `azd` to prompt the user for a value, ignoring the value in the Key Vault secret.

### Example

**main.bicep:**
```bicep
@secure()
param kvSecret string
```

**main.bicepparam:**
```bicep
param kvSecret = getSecret(subscriptionId, resourceGroupName, keyVaultName, 'testsecret')
```

**main.parameters.json:**
```json
{
  "$schema": "https://schema.management.azure.com/schemas/2019-04-01/deploymentParameters.json#",
  "contentVersion": "1.0.0.0",
  "parameters": {
    "kvSecret": {
      "reference": {
        "keyVault": {
          "id": "/subscriptions/<subscription>/resourceGroups/<resource group>/providers/Microsoft.KeyVault/vaults/<vault>"
        },
        "secretName": "testsecret"
      }
    }
  }
}
```

### Validation

Tested `azd up` on [sample code](https://github.com/JeffreyCA/azd-4706-repro) that uses Bicep and JSON parameter files with KV references and confirmed that the secret value flowed through to the deployment output.

Also tested with `alpha.deployment.stacks` set to `on` and `off`.

image


cli/azd/pkg/infra/provisioning/bicep/bicep_provider_test.go
```diff
@@ -63,6 +63,24 @@ func TestBicepPlan(t *testing.T) {
...
+	infraProvider := createBicepProvider(t, mockContext)
+
+	deploymentPlan, err := infraProvider.plan(*mockContext.Context)
+
+	require.Nil(t, err)
+
+	require.IsType(t, &deploymentDetails{}, deploymentPlan)
+	configuredParameters := deploymentPlan.CompiledBicep.Parameters
+
+	require.NotEmpty(t, configuredParameters["kvSecret"])
```
weikanglim, line 78: Just noting this for anyone else who finds this slightly confusing: it took me awhile to figure out that these tests, as part of `createBicepProvider`, references the `test/functional/testdata/samples/webapp` samples, which is where `kvSecret` is getting the keyvault reference set.

In these tests, we also mock out bicep build implementation, hence why the `kvSecret` needs to be updated here. I could see why we do this to avoid invoking real `bicep build`, but I'm also not entirely sure if the mocks actually replicate real-world behavior in Bicep, but at least the tests seem to pass for now.

cli/azd/pkg/infra/provisioning/bicep/bicep_provider.go
```diff
@@ -1844,14 +1844,19 @@ func (p *BicepProvider) ensureParameters(
...
-				configuredParameters[key] = azure.ArmParameterValue{
+				configuredParameters[key] = azure.ArmParameter{
 					Value: paramValue,
 				}
 				if needForDeployParameter {
 					mustSetParamAsConfig(key, paramValue, p.env.Config, param.Secure())
 					configModified = true
 				}
 				continue
+			} else if v.KeyVaultReference != nil {
```
weikanglim, line 1855: Just a small detail here: I'm wondering how we think about given evaluation priority to `KeyVaultReference` (I think this ends up mirroring the stack_deployments logic):

```golang
if v.KeyVaultReference != nil {
   ...  // set reference
   continue
}

// check if v.Value is set and is assignable to the declared type
paramValue := armParameterFileValue(parameterType, v.Value, param.DefaultValue)
if paramValue != nil {
  ...
  continue
}
```

weikanglim (approved):
LGTM. I really appreciate how clean these changes were. It really helped highlight to me the idea of how much of `azd` interpreting bicepparam / ARM files are really "passthrough changes to the ARM control plane".

I also really like that the existing code allowed us to add this change simply without affecting how param prompting works in azd.

cli/azd/pkg/infra/provisioning/bicep/bicep_provider.go
JeffreyCA: Makes sense, I've updated this

cli/azd/pkg/azure/arm_parameters.go
```diff
@@ -14,7 +14,20 @@ type ArmParameterFile struct {
...
+// KeyVaultParameterReference is the model type for a Key Vault parameter reference.
+type KeyVaultParameterReference struct {
+	KeyVault      KeyVaultReference `json:"keyVault"`
+	SecretName    string            `json:"secretName"`
+	SecretVersion string            `json:"secretVersion"`
+}
+
+// KeyVaultReference represents the Key Vault resource ID.
+type KeyVaultReference struct {
+	ID string `json:"id"`
```
vhvb1989, line 32: Wondering if we might want to use `any` for the KeyVaultReference instead of describing its shape.  I don't think we are ever reading the values from the Key Vault reference, are we?

vhvb1989 (approved):
LGTM.
Just adding a question about if we should be parsing the key vault reference or not.

cli/azd/pkg/azure/arm_parameters.go
JeffreyCA: In [stack_deployments.go](https://github.com/Azure/azure-dev/pull/4744/files#diff-9880f2649334e144c8c2fa150eff704aa035c77da8d945e2b634a45c5b45636c) we reference the fields to convert parameters into [`armdeploymentstacks.DeploymentParameter` (azure-sdk-for-go)](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/resources/armdeploymentstacks#DeploymentParameter), which has this shape:

```go
// DeploymentParameter - Deployment parameter for the template.
type DeploymentParameter struct {
	// Azure Key Vault parameter reference.
	Reference *KeyVaultParameterReference

	// Type of the value.
	Type *string

	// Input value to the parameter.
	Value any
}
```

I considered using `armdeploymentstacks.DeploymentParameter` directly instead of redefining the types in `arm_parameter.go`, but it didn't feel quite right since it's part of the `armdeploymentstacks` module. I also couldn't find an equivalent type in azure-sdk-for-go's "non-deployment stack" module.

Do you have ideas to make the changes work with `any`?

vhvb1989: btw, if you think this makes it more complicated, I am OK with having the type definition and doing the manual type creating by mapping each field to field, just like how you have it.  I don't have strong opinion about it.  So, feel free to pick what you prefer

JeffreyCA: Thank you for the example! I understand there are pros/cons to both approaches but am slightly favouring the existing approach for the type safety and it being a bit easier to understand. I'm not too familiar with JSON in Go, and the Marshal-Unmarshal pattern seems pretty useful to know so thanks again for sharing that.

PR #4750, opened 2025-02-01
support scope in aspire manifest

vhvb1989:
Adding support for: https://github.com/dotnet/aspire/blob/f192245405d991fd3c9d62980a6f25fd2929f321/tests/Aspire.Hosting.Azure.Tests/ExistingAzureResourceTests.cs#L225


cli/azd/pkg/apphost/generate_types.go
```diff
@@ -128,6 +128,7 @@ type genOutputParameter struct {
 type genBicepModules struct {
 	Path   string
 	Params map[string]string
+	Scope  *string
```
captainsafia, line 131:
```suggestion
	Scope  map[string]string
```

cli/azd/pkg/apphost/manifest.go
```diff
@@ -100,6 +100,9 @@ type Resource struct {
 
 	// project.v1 and container.v1 uses deployment when the AppHost owns the ACA bicep definitions.
 	Deployment *DeploymentMetadata `json:"deployment,omitempty"`
+
+	// Reference is present on bicep modules to control the scope of the module.
+	Scope *string `json:"scope,omitempty"`
```
captainsafia, line 105:
```suggestion
	Scope map[string]any `json:"params,omitempty"
```

rajeshkamal5050 (approved):

Issue #4749, opened 2025-02-01
[Issue] The location picker ignores the allowed list if metadata is also present


pamelafox:
- [X] Make sure you've installed the latest version using [instructions in the wiki](../wiki/install)

**Output from `azd version`**
1.11.1

**Describe the bug**

https://github.com/Azure-Samples/azure-openai-keyless-python/blob/main/infra/main.bicep

This does not currently restrict the main location picker, but it should!

The fix is to remove the metadata.

This seems like a regression however.

@vhvb1989

PR #4752, opened 2025-02-01
support for default location

vhvb1989:
This PR adds the ability to set a default location, as part of the bicep definition, to be selected by azd during user prompt.

This feature works with priority levels. The default location setting priority is as follows, from highest to lowest:
- 1. Set AZURE_LOCATION in system environment for the main project `location`.  Azd skips the location prompt for main location and uses the value from the system environment when set.
- 2. Use azd metadata for bicep parameter like:

```bicep
@metadata({azd: { 
  type: 'location'
  defaultLocation: 'southcentralus'
}})
param foo string
```

   When using metadata type `location` plus `defaultLocation`, azd will list all locations and select the `defaultLocation` from the list.
    
- Use the `@allowed` tag in the bicep parameter like

```bicep
@allowed([
  'westus3'
  'southcentralus'
  'eastus'
  'eastus2'
  'westus2'
  'centralus'
])
param foo string
```

     Azd will use the fist location in the list as the default location.  Note that azd will not display the values with `location` style but only as a simple list.  Adding the metadata `@location` makes azd to display the location style. Like:

```bicep
@metadata({azd: { 
  type: 'location'
}})
@allowed([
  'westus3'
  'southcentralus'
  'eastus'
  'eastus2'
  'westus2'
  'centralus'
])
param foo string
```

   The `default location` in this case is the first location from the allowed array.  You can also add the `defaultLocation` to the metadata to set a specific location as the default, like:

```bicep
@metadata({azd: { 
  type: 'location'
  defaultLocation: 'southcentralus'
}})
@allowed([
  'westus3'
  'southcentralus'
  'eastus'
  'eastus2'
  'westus2'
  'centralus'
])
param foo string
```

   The metadata `defaultLocation` takes higher priority then the allowed array and is used.

- 3. When neither `defaultLocation`  (metadata) or allowed array is set, azd check if there is a default location saved in the azd local config for the user (~/.azd/config.json) and use that value as the default.  This is the lowest priority and the current behavior before this PR.

fix: https://github.com/Azure/azure-dev/issues/4749

### DIFF
--- a/cli/azd/pkg/azure/arm_template.go
+++ b/cli/azd/pkg/azure/arm_template.go
@@ -143,7 +143,7 @@ type AzdMetadata struct {
 	Type               *AzdMetadataType `json:"type,omitempty"`
 	AutoGenerateConfig *AutoGenInput    `json:"config,omitempty"`
 	DefaultValueExpr   *string          `json:"defaultValueExpr,omitempty"`
-	DefaultLocation    *string          `json:"defaultLocation,omitempty"`
+	Default            *string          `json:"default,omitempty"`
 }
 
 // Description returns the value of the "Description" string metadata for this parameter or empty if it can not be found.
--- a/cli/azd/pkg/azure/arm_template.go
+++ b/cli/azd/pkg/azure/arm_template.go
@@ -143,6 +143,7 @@ type AzdMetadata struct {
 	Type               *AzdMetadataType `json:"type,omitempty"`
 	AutoGenerateConfig *AutoGenInput    `json:"config,omitempty"`
 	DefaultValueExpr   *string          `json:"defaultValueExpr,omitempty"`
+	DefaultLocation    *string          `json:"defaultLocation,omitempty"`
 }
 
 // Description returns the value of the "Description" string metadata for this parameter or empty if it can not be found.
--- a/cli/azd/pkg/infra/provisioning/bicep/bicep_provider.go
+++ b/cli/azd/pkg/infra/provisioning/bicep/bicep_provider.go
@@ -120,7 +120,8 @@ func (p *BicepProvider) EnsureEnv(ctx context.Context) error {
 	// for .bicepparam, we first prompt for environment values before calling compiling bicepparam file
 	// which can reference these values
 	if isBicepParamFile(modulePath) {
-		if err := provisioning.EnsureSubscriptionAndLocation(ctx, p.envManager, p.env, p.prompters, nil); err != nil {
+		if err := provisioning.EnsureSubscriptionAndLocation(
+			ctx, p.envManager, p.env, p.prompters, provisioning.EnsureSubscriptionAndLocationOptions{}); err != nil {
 			return err
 		}
 	}
@@ -130,10 +131,12 @@ func (p *BicepProvider) EnsureEnv(ctx context.Context) error {
 		return fmt.Errorf("%w%w", ErrEnsureEnvPreReqBicepCompileFailed, compileErr)
 	}
 
-	// for .bicep, azd must load a parameters.json file and create the ArmParameters
+	// for .bicep, azd must load a parameters.json file and create the ArmParameters so we know if the are filters
+	// to apply for location (using the allowedValues or the location azd metadata)
 	if isBicepFile(modulePath) {
+		locationParam, locationParamDefined := compileResult.Template.Parameters["location"]
 		var filterLocation = func(loc account.Location) bool {
-			if locationParam, defined := compileResult.Template.Parameters["location"]; defined {
+			if locationParamDefined {
 				if locationParam.AllowedValues != nil {
 					return slices.IndexFunc(*locationParam.AllowedValues, func(allowedValue any) bool {
 						allowedValueString, goodCast := allowedValue.(string)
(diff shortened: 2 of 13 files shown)


cli/azd/pkg/azure/arm_template.go
```diff
@@ -143,6 +143,7 @@ type AzdMetadata struct {
 	Type               *AzdMetadataType `json:"type,omitempty"`
 	AutoGenerateConfig *AutoGenInput    `json:"config,omitempty"`
 	DefaultValueExpr   *string          `json:"defaultValueExpr,omitempty"`
+	DefaultLocation    *string          `json:"defaultLocation,omitempty"`
```
jongio, line 146: Would it be possible to name this default instead of default location?

vhvb1989: yes, that makes sense

cli/azd/pkg/infra/provisioning/bicep/bicep_provider.go
```diff
@@ -176,6 +179,41 @@ func (p *BicepProvider) EnsureEnv(ctx context.Context) error {
...
+		return ok && location.Name == s
+	}) != -1
+}
+
+// defaultLocationToSelectFn resolves if there is an intention from a location parameter to use a default location.
+// If the parameter has the @allowedValues set, it will return the first value from the array as the default.
+// If the parameter has the location azd metadata, it will check if there is configuration to set a default location.
+// If both @allowedValues and azd location metadata config are set, the metadata config will take precedence.
+// If nothing is set, it will return nil.
+func defaultLocationToSelectFn(locationParam azure.ArmTemplateParameterDefinition) *string {
```
weikanglim, line 198: Curious about the intention here, I'm not immediately familiar with the suffix "fn" used here.

cli/azd/pkg/infra/provisioning/bicep/bicep_provider.go
```diff
@@ -176,6 +179,41 @@ func (p *BicepProvider) EnsureEnv(ctx context.Context) error {
...
+	}
+
+	return slices.IndexFunc(*param.AllowedValues, func(v any) bool {
+		s, ok := v.(string)
+		return ok && location.Name == s
+	}) != -1
+}
+
+// defaultLocationToSelectFn resolves if there is an intention from a location parameter to use a default location.
+// If the parameter has the @allowedValues set, it will return the first value from the array as the default.
```
weikanglim, line 194: Extra blank line here to make the comments pretty? 

```suggestion
// defaultLocationToSelectFn resolves if there is an intention from a location parameter to use a default location.
//
// If the parameter has the @allowedValues set, it will return the first value from the array as the default.
```

cli/azd/pkg/infra/provisioning/bicep/bicep_provider.go
```diff
@@ -176,6 +179,41 @@ func (p *BicepProvider) EnsureEnv(ctx context.Context) error {
...
+// If the parameter has the location azd metadata, it will check if there is configuration to set a default location.
+// If both @allowedValues and azd location metadata config are set, the metadata config will take precedence.
```
weikanglim, line 196: FWIW, the comments felt really verbose here. I don't know if you may prefer just explaining the comments in order of precedence.

This was my attempt at shortening here just for reference.

```
// If the parameter has AzdMetadataTypeLocation, with a default location set, the default location is returned.
// If the parameter has AllowedValues, the first option value is returned.
// Otherwise, nil is returned to indicate no user-provided default value.
```

cli/azd/pkg/azure/arm_template.go
```diff
@@ -143,6 +143,7 @@ type AzdMetadata struct {
 	Type               *AzdMetadataType `json:"type,omitempty"`
 	AutoGenerateConfig *AutoGenInput    `json:"config,omitempty"`
 	DefaultValueExpr   *string          `json:"defaultValueExpr,omitempty"`
+	Default            *string          `json:"default,omitempty"`
```
weikanglim, line 146: I'm wondering maybe this would be a good place to document the intended usage of the API contract

cli/azd/pkg/infra/provisioning/bicep/bicep_provider.go
weikanglim: Does it look any different set `defaultPromptValue`

weikanglim (approved):
LGTM. This is a solid value add to our azd picklist customization strategy in Bicep. Only left some very minor comments.

rajeshkamal5050:
@vhvb1989 any impact for VS integration?

cli/azd/pkg/infra/provisioning/bicep/bicep_provider.go
vhvb1989: I think I used copilot to document the function.  I also used it to generate the test cases.
I have a very low bar for naming and documenting.  My bar is just to be enough understandable. So, happy to get any improvements here.  I've taken your suggestion for the name and docs. Thank you.

cli/azd/pkg/infra/provisioning/bicep/bicep_provider.go
vhvb1989: taken

vhvb1989:
> @vhvb1989 any impact for VS integration?

No. VS does neither understand or use azd metadata. VS owns the prompting logic and the UI for it.  IIRC, they use combobox controls to pick sub and location and a simple inputbox for other string parameters. Most likely they ignore annotations like `@allow` (not too bad b/c that annotation is not used for Aspire when generating code).

PR #4760, opened 2025-02-05
fix main with mocked samples

vhvb1989:
Introducing mocked configuration for the samples that we run live in CI (here: https://github.com/Azure/azure-dev/pull/4744/files#diff-b44f84f811e2bf53b8ce7f56d2dc304be0301aaf01927d899d548304032f2944R7) broke the daily build.

The change was able to pass PR checks b/c tests runs with recordings.

When the changes for the samples ran on live mode, deployment fail after trying to fetch a mocked kv.

This PR is adding a new folder to test data to have mocked samples that we can use for unit test only.  This could be a temporal patch until we evaluate with the team if there's a better option to avoid duplicating samples.


JeffreyCA (approved):
Thanks for fixing this!

rajeshkamal5050 (approved):

Issue #4746, opened 2025-01-31
compose: resource conventions


weikanglim:
!! WIP: Work in progress. This is a draft and does not represent final work.

## General Conventions

1. **Type Naming:**
   - Resource types are written without hyphens.
   - Example: `messaging.eventhubs` instead of `messaging-eventhubs`.

2. **Default Resource Naming (simplified scaffolding scenarios):**
   - When adding a resource using `azd add`, a single resource is created with a **default name**. 
   - Example:
     - `event-hubs` for `messaging.eventhubs`.
     - The default name follows a hyphenated format.

3. **Environment Variables:**
   - Environment variables follow the format: `AZURE_<TYPE>_<PROPERTY>`. 
   - With `messaging.eventhubs`, `AZURE_EVENT_HUBS_NAME` represents the Event Hubs namespace.
   - When there are more than one types, an explicit environment variable mapping must be performed (which `azd add` can provide a default configuration modification that is used).

## Common Properties

| Suffix        | Description                              |
|--------------|----------------------------------|
| `_HOST`     | Hostname, i.e. `myaccount.blob.core.windows.net`                         |
| `_URL`      | URI reference , i.e. `redis://`                   |
| `_ENDPOINT` | HTTP URL , i.e. `https://<vault-name>.vault.azure.net/`                         |
| `_PORT`     | Port number, i.e. `333`                       |

#### YAML Definition
```yaml
resources:
  event-hubs:
    type: messaging.eventhubs
    hubs:
      - events
```

#### Binding Environment Variables
| Variable                  | Description                                      |
|---------------------------|--------------------------------------------------|
| `AZURE_EVENT_HUBS_NAME`   | Name of the Event Hubs namespace |
| `AZURE_EVENT_HUBS_HOST`   | Hostname of the Event Hubs namespace  |

---

#### YAML Definition
```yaml
resources:
  service-bus:
    type: messaging.servicebus
    queues:
      - myQueue
    topics:
      - myTopic
```

#### Binding Environment Variables
| Variable                  | Description                                      |
|---------------------------|--------------------------------------------------|
| `AZURE_SERVICE_BUS_NAME`  | Name of the Service Bus namespace |
| `AZURE_SERVICE_BUS_HOST`  | Hostname of the Service Bus namespace |

---

JeffreyCA:
> Environment variables follow the format: `AZURE_<TYPE>_<PROPERTY>`.

[...]

Issue #4727, opened 2025-01-23
compose: discover and use env variables without provisioning host service


weikanglim:
Can we display environment variables added before provisioning? Suggested by @JeffreyCA

JeffreyCA:
It looks like we kind of already do this when you answer "Yes (preview changes)" to the prompt, "Do you want to provision and deploy these changes?", when there's a host or a host is being added:

https://github.com/Azure/azure-dev/blob/bacd7a2ee8d67f88a6dcff820ca332315a4c88d1/cli/azd/internal/cmd/add/add_preview.go#L30-L49

weikanglim:
@JeffreyCA , indeed. I think this enhancement would target when the user says no to preview provisioning.

One idea I have currently is to have `azd show api` show the environment variables, and for values that can't be resolved until provisioning, we can display a placeholder value `<available after provisioning>`.

weikanglim:
I think we should expand on what we already do today with `azd show <AI model name>`

And it should support `azd show <name>.env` to print out only the standard environment variable names

For example, `azd show service-bus`

```
AZURE_SERVICE_BUS_NAME='<bus name>'
AZURE_SERVICE_BUS_HOST='<bus name>.servicebus.windows.net'
```

I'm going take a stab at implementing this to decide if whether a generic/dynamic fetch works better than the concrete Azure SDK types -- I did find a little nervous when I was experimenting with the `show <model>` approach because we end up taking dependencies on the entire Azure SDK libraries for a little functionality we currently need.

weikanglim:
We also want to update the preview experience to also show the variables that will be made available.

![Image](https://github.com/user-attachments/assets/b0830093-5ab7-4b11-8a55-080ffd2d7646)

PR #4765, opened 2025-02-05
compose: Add `azd add` support for storage accounts (blob service)

### LINKED ISSUE #4716: compose: `azd add` support for Storage Accounts
Java support is already covered in #4641 by the SJAD team

- [x] `azd add` prompts
- [x] azure.yaml integration
- [x] Bicep generation
- [x] AKV secrets export

JeffreyCA:
Closes #4716

This PR adds `azd add` support for storage accounts (blob service). These changes allow only one storage account resource to be added. The app is set up to connect to the storage account using keyless authentication. The "Storage Blob Data Contributor" role assignment is given to the local user and the user-assigned identity.

`azure.yml` supports multiple blob container names and `azd provision` will create them, but the `azd add` experience will only add one blob container for now.

## TODO

- [x] Finalize on auth type support: will use managed identity by default
- [ ] Finalize UX/wording

## azure.yaml schema

```yaml
resources:
  storage
    type: storage
    containers:
      - <user-provided name>
```

## Binding variables

`AZURE_STORAGE_ACCOUNT_NAME` - Storage account name
`AZURE_STORAGE_BLOB_ENDPOINT` - Storage account blob endpoint

## Demo

https://github.com/user-attachments/assets/ddd731e4-b1f9-4f33-8997-863f67f535b1

### DIFF
--- a/cli/azd/internal/cmd/add/add_configure_storage.go
+++ b/cli/azd/internal/cmd/add/add_configure_storage.go
@@ -46,19 +46,6 @@ func fillStorageDetails(
 		return nil, err
 	}
 
-	selection, err := console.Select(ctx, input.ConsoleOptions{
-		Message: "Use keyless authentication?",
-		Options: []string{
-			"Yes (recommended)",
-			"No",
-		},
-	})
-	if err != nil {
-		return nil, err
-	}
-
-	modelProps.KeylessAuth = selection == 0
-
 	for _, option := range selectedDataTypes {
 		switch option {
 		case StorageDataTypeBlob:
--- a/cli/azd/internal/cmd/add/add_configure_storage.go
+++ b/cli/azd/internal/cmd/add/add_configure_storage.go
@@ -1,0 +1,142 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package add
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"strings"
+
+	"github.com/azure/azure-dev/cli/azd/internal/names"
+	"github.com/azure/azure-dev/cli/azd/pkg/input"
+	"github.com/azure/azure-dev/cli/azd/pkg/output"
+	"github.com/azure/azure-dev/cli/azd/pkg/project"
+)
+
+const (
+	StorageDataTypeBlob = "Blobs"
+)
+
+func allStorageDataTypes() []string {
+	return []string{StorageDataTypeBlob}
+}
+
+func fillStorageDetails(
+	ctx context.Context,
+	r *project.ResourceConfig,
+	console input.Console,
+	p PromptOptions) (*project.ResourceConfig, error) {
+	if _, exists := p.PrjConfig.Resources["storage"]; exists {
+		return nil, fmt.Errorf("only one Storage resource is allowed at this time")
+	}
+
+	if r.Name == "" {
+		r.Name = "storage"
+	}
+
+	modelProps, ok := r.Props.(project.StorageProps)
+	if !ok {
+		return nil, fmt.Errorf("invalid resource properties")
+	}
+
+	selectedDataTypes, err := selectStorageDataTypes(ctx, console)
+	if err != nil {
+		return nil, err
+	}
+
+	selection, err := console.Select(ctx, input.ConsoleOptions{
+		Message: "Use keyless authentication?",
+		Options: []string{
+			"Yes (recommended)",
+			"No",
+		},
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	modelProps.KeylessAuth = selection == 0
+
+	for _, option := range selectedDataTypes {
+		switch option {
+		case StorageDataTypeBlob:
+			if err := fillBlobDetails(ctx, console, &modelProps); err != nil {
+				return nil, err
+			}
+		default:
+			return nil, fmt.Errorf("unsupported data type: %s", option)
+		}
+	}
+
+	r.Props = modelProps
+	return r, nil
+}
+
+func selectStorageDataTypes(ctx context.Context, console input.Console) ([]string, error) {
+	var selectedDataOptions []string
+	for {
+		var err error
+		selectedDataOptions, err = console.MultiSelect(ctx, input.ConsoleOptions{
+			Message:      "What type of data do you want to store?",
+			Options:      allStorageDataTypes(),
+			DefaultValue: []string{StorageDataTypeBlob},
+		})
+		if err != nil {
+			return nil, err
+		}
+
+		if len(selectedDataOptions) == 0 {
+			console.Message(ctx, output.WithErrorFormat("At least one data type must be selected"))
+			continue
+		}
+		break
+	}
+	return selectedDataOptions, nil
+}
+
+func fillBlobDetails(ctx context.Context, console input.Console, modelProps *project.StorageProps) error {
+	for {
+		containerName, err := console.Prompt(ctx, input.ConsoleOptions{
+			Message: "Input a blob container name to be created:",
+			Help: "Blob container name\n\n" +
+				"A blob container organizes a set of blobs, similar to a directory in a file system.",
+		})
+		if err != nil {
+			return err
+		}
+
+		if err := validateContainerName(containerName); err != nil {
+			console.Message(ctx, err.Error())
+			continue
+		}
+		modelProps.Containers = append(modelProps.Containers, containerName)
+		break
+	}
+	return nil
+}
+
+// validateContainerName validates storage account container names.
+// Reference:
+// https://learn.microsoft.com/rest/api/storageservices/naming-and-referencing-containers--blobs--and-metadata
+func validateContainerName(name string) error {
+	if len(name) < 3 {
+		return errors.New("name must be 3 characters or more")
+	}
+
+	if strings.Contains(name, "--") {
+		return errors.New("name cannot contain consecutive hyphens")
+	}
+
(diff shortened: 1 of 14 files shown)


cli/azd/internal/cmd/add/add_configure_storage.go
```diff
@@ -0,0 +1,142 @@
...
+
+	if strings.Contains(name, "--") {
+		return errors.New("name cannot contain consecutive hyphens")
+	}
+
+	if strings.ToLower(name) != name {
+		return errors.New("name must be all lower case")
+	}
+
+	err := names.ValidateLabelName(name)
```
JeffreyCA, line 136: Not directly related but I saw `ValidateLabelName` uses a regex that checks for a minimum string length of 2 - is that intentional? I couldn't find in the RFC 1123 spec that it had to have a min length of 2.

```go
var rfc1123LabelRegex = regexp.MustCompile(`^[a-zA-Z0-9][a-zA-Z0-9-]{0,61}[a-zA-Z0-9]$`)
...
if !rfc1123LabelRegex.MatchString(name) {
    return errors.New("name must contain only lower-cased alphanumeric characters or '-', i.e. a-z, 0-9, or '-'")
}
```

cli/azd/pkg/project/resources.go
```diff
@@ -128,6 +137,16 @@ func (r *ResourceConfig) UnmarshalYAML(value *yaml.Node) error {
...
 		raw.Props = cap
+	case ResourceTypeStorage:
+		sp := StorageProps{}
+		if err := unmarshalProps(&sp); err != nil {
+			return err
+		}
+		// Default KeylessAuth to true if omitted
+		if _, ok := raw.RawProps["keyless"]; !ok {
+			sp.KeylessAuth = true
+		}
```
weikanglim, line 148: This should work:

```suggestion
		sp := StorageProps{
		   Keyless: true,
		}
		if err := unmarshalProps(&sp); err != nil {
			return err
		}
```

cli/azd/internal/cmd/add/add_configure_storage.go
weikanglim: a 1 letter name should be allowed, feel free to fix the regex if you'd like!

cli/azd/internal/cmd/add/add_preview.go
```diff
@@ -67,6 +67,17 @@ func Metadata(r *project.ResourceConfig) resourceMeta {
...
+	case project.ResourceTypeStorage:
+		res.AzureResourceType = "Microsoft.Storage/storageAccounts"
+		res.UseEnvVars = []string{
+			"AZURE_STORAGE_ACCOUNT_NAME",
+			"AZURE_STORAGE_BLOB_ENDPOINT",
+		}
+
+		if modelProps, ok := r.Props.(project.StorageProps); ok && !modelProps.KeylessAuth {
+			res.UseEnvVars = append(res.UseEnvVars, "AZURE_STORAGE_ACCOUNT_KEY")
+			res.UseEnvVars = append(res.UseEnvVars, "AZURE_STORAGE_CONNECTION_STRING")
```
weikanglim, line 79: there is a variadic append if that looks any better to you:

```golang
res.UseEnvVars = append(res.UseEnvVars, 
     "AZURE_STORAGE_ACCOUNT_KEY",
     "AZURE_STORAGE_CONNECTION_STRING",
)
```

cli/azd/resources/scaffold/templates/resources.bicept
```diff
@@ -134,6 +134,51 @@ module postgreServer 'br/public:avm/res/db-for-postgre-sql/flexible-server:0.1.4
...
+    roleAssignments: [
+      {{- if .StorageAccount.KeylessAuth }}
+      {{- range .Services}}
+      {
+        principalId: {{bicepName .Name}}Identity.outputs.principalId
+        principalType: 'ServicePrincipal'
+        roleDefinitionIdOrName: 'Storage Blob Data Owner'
+      }
+      {{- end}}
+      {{- end}}
```
weikanglim, line 165: add the local user too (our current local dev story)

```bicep
{
        principalId: principalIdd
        principalType: 'User'
        roleDefinitionIdOrName: 'Storage Blob Data Contributor'
}
```

cli/azd/resources/scaffold/templates/resources.bicept
```diff
@@ -134,6 +134,51 @@ module postgreServer 'br/public:avm/res/db-for-postgre-sql/flexible-server:0.1.4
...
+      {{- end }}
+    }
+    location: location
+    roleAssignments: [
+      {{- if .StorageAccount.KeylessAuth }}
+      {{- range .Services}}
+      {
+        principalId: {{bicepName .Name}}Identity.outputs.principalId
+        principalType: 'ServicePrincipal'
+        roleDefinitionIdOrName: 'Storage Blob Data Owner'
```
weikanglim, line 162: Does it make more sense as `Storage Blob Data Contributor`?

cli/azd/internal/scaffold/spec.go
```diff
@@ -54,6 +56,11 @@ type AIModelModel struct {
 	Version string
 }
 
+type StorageAccount struct {
+	Containers  []string
+	KeylessAuth bool
```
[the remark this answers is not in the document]
weikanglim: The terminology in this area is slowly being turned around (and arguably, from my perspective, for good reasons).

Most starting out docs have "passwordless connections"
- [Storage blob quickstart](https://learn.microsoft.com/en-us/azure/storage/blobs/storage-quickstart-blobs-python?tabs=managed-identity%2Croles-azure-portal%2Csign-in-azure-cli&pivots=blob-storage-quickstart-scratch)
- [Passwordless overview[(https://learn.microsoft.com/en-us/azure/developer/intro/passwordless-overview)
- Keyless is more teminology correct in my mind based on the idead of "pre-negiotated key" [Keyless connections](https://learn.microsoft.com/en-us/azure/developer/ai/keyless-connections?tabs=csharp%2Cazure-cli)

cli/azd/internal/cmd/add/add_preview.go
JeffreyCA: Good to know, thanks 🙂

cli/azd/internal/scaffold/spec.go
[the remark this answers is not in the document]
JeffreyCA: Thanks for chiming in - I removed this as we've finalized sticking with Entra/managed identity for now and not prompt for or expose the auth type in the schema. The SJAD team is onboard with this as well.

SophCarp:
image
Shouldn't there be environment variables in the highlighted section?

SophCarp:
After I provisioned the storage account: 
image

We should make sure that `azd show` will support this resource type

SophCarp:
Ui bug: 
image
Should ensure that `Provision these changes to Azure` doesn't duplicate lines when you enter an input

weikanglim:
> We should make sure that azd show will support this resource type

@SophCarp I'll take a pass to enable `show` for all resources as part of #4727 . The idea is that it will display all variables.
We may also need to adjust the "Environment variables" section to display the "standard variables" that the user can use after provisioning. This is now possible with the design direction in #4746.

JeffreyCA:
Thanks for the feedback!

@SophCarp @weikanglim 

Do we want `show` to display other details relevant to the resource besides the variable names and values? I was able to experiment with displaying some basic details about the resource for `azd show storage [--show-secrets]`, but it does add a new dependency on `azure-sdk-for-go/sdk/resourcemanager/storage/armstorage`.

image

weikanglim:
@JeffreyCA  

> but it does add a new dependency on azure-sdk-for-go/sdk/resourcemanager/storage/armstorage.

You are indeed right, and part of my goal is to make this a generic HTTP rest call that skips the SDK dependency to scale (and why we haven't added support for other resources) -- we want the Bicep authoring to be the central focus of adding a resource right now. I'm starting work on this pretty soon, but feel free to share the private branch with the work if you'd like.

JeffreyCA:
> @JeffreyCA
> 
> > but it does add a new dependency on azure-sdk-for-go/sdk/resourcemanager/storage/armstorage.
> 
> You are indeed right, and part of my goal is to make this a generic HTTP rest call that skips the SDK dependency to scale (and why we haven't added support for other resources) -- we want the Bicep authoring to be the central focus of adding a resource right now. I'm starting work on this pretty soon, but feel free to share the private branch with the work if you'd like.

I see, I'll leave the `show` changes out of this PR

puicchan:
1. Should we provide a default container name like we did for AI model? 
image

3. If I type ? for "Input a blob container name to be created" - should we add to below this info? 
      The name must be 3 characters or more.
      The name cannot contain consecutive hyphens.
      The name must be all lowercase.
image

cli/azd/internal/names/label.go
```diff
@@ -9,7 +9,7 @@ import (
 	"strings"
 )
 
-var rfc1123LabelRegex = regexp.MustCompile(`^[a-zA-Z0-9][a-zA-Z0-9-]{0,61}[a-zA-Z0-9]$`)
+var rfc1123LabelRegex = regexp.MustCompile(`^(?:[A-Za-z0-9]|[A-Za-z0-9][A-Za-z0-9-]{0,61}[A-Za-z0-9])$`)
```
weikanglim, line 12: Thanks for doing this!

weikanglim (approved):
LGTM! Thanks for also taking the time to fix the name regex!

JeffreyCA:
Merging this as-is and will address remaining feedback in a follow up PR

PR #4766, opened 2025-02-05
Extensions: Adds `extensions` command group

wbreza:
This PR adds the `extension` command group into `azd` and enables the execution of installed extensions.

> [!IMPORTANT]
> The extensions feature is behind a feature flag and must be enabled by running:
> `azd config set alpha.extensions on`

## Extension Commands 

- `azd extension list` - List available extensions in the registry
- `azd extension show <name>` - Show details of a specified extension
- `azd extension install <name>` - Install an extension
- `azd extension uninstall <name>` - Uninstall an extension
- `azd extension upgrade <name>` - Upgrades an extension

> [!NOTE]
> The extension command group is also available as a shorter alias `ext`:

### Source Commands

Extension sources are used to reference different extension registries.  

- `azd extension source list` - List available extension sources
- `azd extension source add` - Add a new extension source
- `azd extension source remove` - Remove an extension source

## Registry

The registry is loaded from an aka URL @ https://aka.ms/azd/extensions/registry

- The registry is protected by a digital signature verification. Only extensions from a verified registry can be installed.
- Binaries also require checksum validation before extensions can be installed.

### `azd extension list`
image

### `azd extension show ai`
image

### `azd extension install ai`
image

### `azd -h`
Extensions show up in standard azd help commands.
image

### `azd test -h`
Extension help shells into extension executable
image

### `azd test --unit`
Invoke extensions through azd command interface
image

### `azd extension uninstall ai`
image


cli/azd/extensions/registry.schema.json
vhvb1989, line 1: Should this schema be at `azure-dev/schemas/` ?

cli/azd/cmd/actions/action_descriptor.go
```diff
@@ -133,16 +133,17 @@ type ActionHelpOptions struct {
...
-	CmdGroupConfig  RootLevelHelpOption = "Configure and develop your app"
-	CmdGroupManage  RootLevelHelpOption = "Manage Azure resources and app deployments"
-	CmdGroupMonitor RootLevelHelpOption = "Monitor, test and release your app"
-	CmdGroupAbout   RootLevelHelpOption = "About, help and upgrade"
+	CmdGroupNone       RootLevelHelpOption = ""
+	CmdGroupConfig     RootLevelHelpOption = "Configure and develop your app"
+	CmdGroupManage     RootLevelHelpOption = "Manage Azure resources and app deployments"
+	CmdGroupMonitor    RootLevelHelpOption = "Monitor, test and release your app"
+	CmdGroupAbout      RootLevelHelpOption = "About, help and upgrade"
+	CmdGroupExtensions RootLevelHelpOption = "Installed Extensions"
```
vhvb1989, line 141: I feel like adding and showing the installed extensions can be a good and bad thing.  

image

The more installed extensions, the longer the output for the main help. The longer the main help is, the more chances for the user not having enough room to see it all and end up looking just the last part of the help description.

Since we are providing a command to list and see the extensions (azd extensions list), I wonder if having the extensions also listed by default in the main help is beneficial?

It might also bring issues for testing the Test-usage snaps, as it would depend on what extensions you have when you generate the snap.

As an alternative, we could show a hardcoded message for the Installed Extensions group like:
```
  Installed Extensions
    There are installed extensions: run azd extensions list to discover it
```

We would only show the message after detecting at least one extension installed?

We can also maybe use some input/feedback from @SophCarp and @ccbarragan ?

vhvb1989 (approved):
LGTM.
Just wondering if we could have some UX conversation about displaying all installed extensions for main help

cli/azd/cmd/actions/action_descriptor.go
wbreza: Thanks for the feedback @vhvb1989.  We'll start out as is and see what feedback we receive and change as needed.

cli/azd/extensions/registry.schema.json
wbreza: i thought it would be useful to be close to the file itself.  Open for discussions.

PR #4768, opened 2025-02-06
package: copy symlinks in buildForZip

### LINKED ISSUE #4715: [Issue] Symlinks ignored when zipping for app service deployment
- [/] Make sure you've installed the latest version using [instructions in the wiki](../wiki/install)

**Output from `azd version`**
`azd version 1.11.1 (commit ae08ceba17c7078c547df9ee5f97e9e7ceb2fe53)`

**Describe the bug**
We have the following `azure.yaml`:

```yaml

name: ai-document-review
metadata:
  template: ai-document-review@1.0.0
infra:
  provider: terraform
services:
  app:
    project: app/api
    host: appservice
    language: python
    hooks:
      # Build the UI code into the API's www directory.
      prepackage:
        shell: sh
        run: 'task app-build-ui'
  flows:
    project: flows
    host: appservice
    language: python
```

The `app` service is a FastAPI Python application, and it consumes a set of common utils in our monorepo from the top-level directory (in a folder called `common`). To reference these we use a symlink within the `app/api` directory:

Image

When trying to deploy the service with `azd up`, we run into the following challenge:

```
ERROR: error executing step command 'package --all': failed packaging service 'app': open /var/folders/kl/dwz9d6xx5jx9s3ygwtqc5q_c0000gn/T/azd2971955672/common: no such file or directory
```

Which I've tracked down to the copy implementation in the cli:

https://github.com/Azure/azure-dev/blob/88ad9bc2cf8e63ef6e0e6b4577770127a162fc3d/cli/azd/pkg/project/project_utils.go#L54 

It seems it uses [this library](https://github.com/otiai10/copy) and doesn't pass any options for:

```go
OnSymlink func(src string) SymlinkAction
```

Which defaults to ignoring them: https://github.com/otiai10/copy/blob/5ef5923d6ff098ea7a51792c1ec963d04fad9794/copy.go#L307

**To Reproduce**
1. Create new Python App Service `azd` template with an appservice service in the `azure.yaml` config
2. Add `common` folder to root and create dummy util
3. Add symlink to common within `src/api` directory
4. Call the util from the api `main.py`
5. `azd up`

**Expected behavior**
Symlinks should be respected and `Deep` copied, or failing that a new option should be introduced to enable this

**Environment**
- Mac M3, Python 3.12.8

jjgriff93:
Fixes #4715

Resolves issues resulting from ignored symlinks during directory copy in packaging operations, by adding a `copy.Deep` option to the directory copy handler.

### Improvements to functionality:

* Added handling for symlinks during the copy process. [[1]](diffhunk://#diff-3e9eda9e4fde242c314ba30a95e5225b8d46ea40f4de75520ccbe54c137ab48aR59-R63) [[2]](diffhunk://#diff-3e9eda9e4fde242c314ba30a95e5225b8d46ea40f4de75520ccbe54c137ab48aR73)

### New tests:

* Added a new test `Test_CLI_Package_FromServiceDirectorySymlinked` to verify packaging functionality when the source directory contains symlinks.


cli/azd/test/functional/package_test.go
```diff
@@ -69,6 +69,48 @@ func Test_CLI_Package_FromServiceDirectory(t *testing.T) {
 	require.Contains(t, result.Stdout, "Packaging service web")
 }
 
+func Test_CLI_Package_FromServiceDirectorySymlinked(t *testing.T) {
```
weikanglim, line 72: This is really useful! I'm going to have to take a second look here on the proposed fix. From what I gander, the proposed fix *does* solve the particular case of python packaging... but there are still edge cases with how `azd` does zip deploy for other languages where a `copy.Copy` isn't performed.

I'm seeing if I can use this test to verify a fix in `rzip.CreateFromDirectory` to do a deep copy.. Supposedly for the python code path, we'd end up doing a shallow copy in our current behavior, but a deep copy when it's time to zip for deployment.

weikanglim: I did end up submitting #4773 which I confirmed locally passes this test. What happens underneath the covers is that `azd` will do a shallow copy with the symlinks, but during zip-time, it would resolve the symbolically linked directories -- it also happens to be the source of where the reported error come from.

@jjgriff93 feel free to see if the private bits there (see install instructions [here](https://github.com/Azure/azure-dev/pull/4773#issuecomment-2644072908)) also happen to addresss your scenario. I do appreciate the effort spent here to document and propose the changes -- hope it helps!

jjgriff93: Just tested and working, thanks @weikanglim! Shall close as fixed now that https://github.com/Azure/azure-dev/pull/4773 is merged

jjgriff93:
Fixed in https://github.com/Azure/azure-dev/pull/4773

Issue #4699, opened 2025-01-16
[env with secrets] Support for pipeline config


vhvb1989:
When mapping an ENV key to a secret for the pipeline configuration in azure.yaml like:

```yaml
pipeline:
  secrets:
    - ENV-FOO 
  variables:
    - ENV-FOO
```

azd should check if the value of the env is an Azure Key Vault secret and if yes, pull the value and save it as secret for the pipeline.

But, if the env key is defined as a variable, azd should grant read-access to the key-vault account for the service-principal used for the pipeline.

PR #4770, opened 2025-02-07
Integrating akvs to pipeline config

vhvb1989:
fix: https://github.com/Azure/azure-dev/issues/4699

## Pipeline config

The Azure Developer CLI simplifies the process of setting up continuous integration (CI) for your application. Whether you are using GitHub or Azure DevOps, you can run the command `azd pipeline config` and follow the guided steps provided by AZD to configure CI/CD.

As part of the automatic configuration, AZD creates secrets and variables for your CI/CD deployment workflow. For example, the Azure Subscription ID and location are set as variables. Additionally, you can define a list of variables and secrets by using the `pipeline` configuration in the `azure.yaml` file within your project. The list of variables or secrets you define corresponds to the names of the keys in your AZD environment (.env). If the name of the key holds a secret reference (akvs), AZD will apply the following rules to set the value in your CI/CD settings.

### Variables

If the secret is added to the `variables` section of the pipeline configuration, the Azure Developer CLI (AZD) will use the value from the environment without retrieving the actual secret value. This approach is beneficial when you prefer to maintain Azure Key Vault references within your CI/CD settings. By doing so, you can rotate your secrets in the Key Vault, ensuring that your CI/CD pipeline uses the latest secret values without the need to update your workflow variables or secrets.

#### Example

- From an initialized AZD project, you have already set a secret. For example, you ran:

```sh
azd env set-secret SECURE_KEY
```

If you run `azd env get-values`, you would see an output like:

```
AZURE_ENV_NAME="your-env-name"
AZURE_LOCATION="location"
AZURE_SUBSCRIPTION_ID="faa080af-c1d8-40ad-9cce-000000000000"
SECURE_KEY="akvs://faa080af-c1d8-40ad-9cce-000000000000/vivazqu-kv/SECURE-KEY-kv-secret"
```

- Then, you can set `SECURE_KEY` as a `variable` for your CI/CD pipeline by adding the `pipeline` field to the `azure.yaml` like:

```yaml
name: your-project-name
pipeline:
  variables:
    - SECURE_KEY
```
When you run `azd pipeline config`, the `SECURE_KEY` will be set as a variable in your CI/CD workflow, and its value will be the Azure Key Vault reference. For the example above, it would be: `akvs://faa080af-c1d8-40ad-9cce-000000000000/vivazqu-kv/SECURE-KEY-kv-secret`.

> Note: AZD will attempt to assign a read-access role to the service principal used for logging into AZD from the CI/CD workflow. If you do not have sufficient permissions to assign the read role for the Key Vault, the operation will fail.

You can reference the `SECURE_KEY` variable as a system environment variable. If `SECURE_KEY` is mapped to a Bicep input parameter or if it is mapped from a hook definition (refer to the earlier sections of this document for more details), AZD will automatically retrieve the value for the secret.

### Secrets

If the secret is added to the `secrets` section of the pipeline configuration, the Azure Developer CLI (AZD) will retrieve the actual value of the secret from the Azure Key Vault. This method is useful when it is not possible to assign a read-access role to the service principal used in your CI/CD workflow. However, it is important to note that you will need to update the secret value manually each time the secret is rotated. To re-apply the updated value, you can run the `azd pipeline config` command again.

#### Example

- From an initialized AZD project, you have already set a secret. For example, you ran:

```sh
azd env set-secret SECURE_KEY
```

If you run `azd env get-values`, you would see an output like:

```
AZURE_ENV_NAME="your-env-name"
AZURE_LOCATION="location"
AZURE_SUBSCRIPTION_ID="faa080af-c1d8-40ad-9cce-000000000000"
SECURE_KEY="akvs://faa080af-c1d8-40ad-9cce-000000000000/vivazqu-kv/SECURE-KEY-kv-secret"
```

- Then, you can set `SECURE_KEY` as a `secret` for your CI/CD pipeline by adding the `pipeline` field to the `azure.yaml` like:

```yaml
name: your-project-name
pipeline:
  secrets:
    - SECURE_KEY
```

- When you run `azd pipeline config`, the `SECURE_KEY` will be set as a secret in your CI/CD workflow and its value will be the Azure Key Vault value.

### DIFF
--- a/cli/azd/pkg/keyvault/keyvault.go
+++ b/cli/azd/pkg/keyvault/keyvault.go
@@ -356,20 +356,11 @@ func NewAkvs(subId, vaultId, secretName string) string {
 }
 
 func (kvs *keyVaultService) SecretFromAkvs(ctx context.Context, akvs string) (string, error) {
-	if !IsAkvs(akvs) {
-		return "", fmt.Errorf("invalid Azure Key Vault Secret reference: %s", akvs)
+	parseAkvs, err := ParseAkvs(akvs)
+	if err != nil {
+		return "", err
 	}
-
-	noSchema := strings.TrimPrefix(akvs, vaultSchemaAkvs)
-	vaultParts := strings.Split(noSchema, "/")
-	if len(vaultParts) != 3 {
-		return "", fmt.Errorf(
-			"invalid Azure Key Vault Secret reference: %s. Expected format: %s",
-			akvs,
-			vaultSchemaAkvs+"<subscription-id>/<vault-name>/<secret-name>",
-		)
-	}
-	subscriptionId, vaultName, secretName := vaultParts[0], vaultParts[1], vaultParts[2]
+	subscriptionId, vaultName, secretName := parseAkvs.SubscriptionId, parseAkvs.VaultName, parseAkvs.SecretName
 	// subscriptionId is required by the Key Vault service to figure the TenantId for the
 	// tokenCredential. The assumption here is that the user has access to the Tenant
 	// used to deploy the app and to whatever Tenant the Key Vault is in. And the tokenCredential
@@ -379,4 +370,31 @@ func (kvs *keyVaultService) SecretFromAkvs(ctx context.Context, akvs string) (st
 		return "", fmt.Errorf("fetching secret value from key vault: %w", err)
 	}
 	return secretValue.Value, nil
+}
+
+type Akvs struct {
+	SubscriptionId string
+	VaultName      string
+	SecretName     string
+}
+
+func ParseAkvs(akvs string) (Akvs, error) {
+	if !IsAkvs(akvs) {
+		return Akvs{}, fmt.Errorf("invalid Azure Key Vault Secret reference: %s", akvs)
+	}
+
+	noSchema := strings.TrimPrefix(akvs, vaultSchemaAkvs)
+	vaultParts := strings.Split(noSchema, "/")
+	if len(vaultParts) != 3 {
+		return Akvs{}, fmt.Errorf(
+			"invalid Azure Key Vault Secret reference: %s. Expected format: %s",
+			akvs,
+			vaultSchemaAkvs+"<subscription-id>/<vault-name>/<secret-name>",
+		)
+	}
+	return Akvs{
+		SubscriptionId: vaultParts[0],
+		VaultName:      vaultParts[1],
+		SecretName:     vaultParts[2],
+	}, nil
 }
--- a/cli/azd/pkg/keyvault/keyvault.go
+++ b/cli/azd/pkg/keyvault/keyvault.go
@@ -341,7 +341,7 @@ const (
 	RoleIdKeyVaultSecretsUser   string = resourceIdPathPrefix + "4633458b-17de-408a-b874-0445c86b69e6"
 )
 
-func IsAkvs(id string) bool {
+func IsAzureKeyVaultSecret(id string) bool {
 	return strings.HasPrefix(id, vaultSchemaAkvs)
 }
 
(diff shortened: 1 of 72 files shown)


cli/azd/pkg/pipeline/testdata/Test_promptForCiFiles-no_files_-_github_selected_-_App_host_-_fed_Cred.snap
```diff
@@ -22,7 +22,7 @@ jobs:
       AZURE_TENANT_ID: ${{ vars.AZURE_TENANT_ID }}
       AZURE_SUBSCRIPTION_ID: ${{ vars.AZURE_SUBSCRIPTION_ID }}
       AZURE_ENV_NAME: ${{ vars.AZURE_ENV_NAME }}
-      AZURE_LOCATION: ${{ vars.AZURE_LOCATION }}
+      AZURE_LOCATION: ${{ vars.AZURE_LOCATION }}
```
wbreza, line 25: nit: Remove whitespace.
```suggestion
      AZURE_LOCATION: ${{ vars.AZURE_LOCATION }}
```

cli/azd/pkg/pipeline/pipeline_manager.go
```diff
@@ -408,6 +412,70 @@ func (pm *PipelineManager) Configure(ctx context.Context, projectName string) (r
...
+				"assigning read access role for Key Vault to service principal: "+
+					"key vault '%s' not found in subscription '%s'", akvs.VaultName, akvs.SubscriptionId)
+		}
+
+		err = pm.entraIdService.CreateRbac(
+			ctx, akvs.SubscriptionId, vaultResourceId, keyvault.RoleIdKeyVaultSecretsUser, *servicePrincipal.Id)
+		if err != nil {
+			return result, fmt.Errorf(
+				"assigning read access role for Key Vault to service principal: %w", err)
+		}
```
wbreza, line 473: What happens when the user runs `azd pipeline config` multiple times? It doesn't look like we are checking to see if an RBAC role assignment was previously made and thinking that this could fail or potentially create duplicate assignments.

cli/azd/pkg/keyvault/keyvault.go
```diff
@@ -356,27 +356,45 @@ func NewAkvs(subId, vaultId, secretName string) string {
...
+	return secretValue.Value, nil
+}
+
+type Akvs struct {
+	SubscriptionId string
+	VaultName      string
+	SecretName     string
+}
+
+func ParseAkvs(akvs string) (Akvs, error) {
```
wbreza, line 381: I'd recommend spelling out `Akvs` since users may not be immediately aware that it refers to an Azure KeyVault Secret reference.

cli/azd/pkg/keyvault/keyvault.go
```diff
@@ -356,27 +356,45 @@ func NewAkvs(subId, vaultId, secretName string) string {
 }
 
 func (kvs *keyVaultService) SecretFromAkvs(ctx context.Context, akvs string) (string, error) {
+	parseAkvs, err := ParseAkvs(akvs)
+	if err != nil {
+		return "", err
+	}
+	subscriptionId, vaultName, secretName := parseAkvs.SubscriptionId, parseAkvs.VaultName, parseAkvs.SecretName
```
wbreza, line 363: This is a cool little variable assignment notation that I wasn't aware of.  Is there a reason to create these new variables vs using the parseAkvs.* members in the next call to get the keyvault secret?

wbreza (approved):
Looks good overall!  I added a couple comments and questions.

cli/azd/pkg/pipeline/testdata/Test_promptForCiFiles-no_files_-_github_selected_-_App_host_-_fed_Cred.snap
vhvb1989: This is coming from a go template. I will update the template and re-run the test.
Good catch!

vhvb1989 (commented):
ty!

cli/azd/pkg/pipeline/pipeline_manager.go
wbreza: Good to know - i was curious since our implementation uses `roleAssignmentId := uuid.New().String()` to generate a unique assignment id on each call.

weikanglim:
@vhvb1989

One quick question on the PR description... If I'm reading the description right, my understanding is as follows:

```yaml
name: your-project-name
pipeline:
  variables:
    - SECURE_KEY # Set as a variable reference; the value is resolved at deployment time
   secrets:
     - SECURE_KEY # Set as a value secret; the value is resolved during `pipeline config` and set as a literal value in CI
```

Does that sound right? In some sense:
- `variables.SECURE_KEY` means "variable reference", but
- `secrets.SECURE_KEY` means "literal value".

vhvb1989:
> @vhvb1989
> 
> One quick question on the PR description... If I'm reading the description right, my understanding is as follows:
> 
> ```yaml
> # yaml-language-server: $schema=https://raw.githubusercontent.com/Azure/azure-dev/main/schemas/v1.0/azure.yaml.json
> name: your-project-name
> pipeline:
>   variables:
>     - SECURE_KEY # Set as a variable reference; the value is resolved at deployment time
>    secrets:
>      - SECURE_KEY # Set as a value secret; the value is resolved during `pipeline config` and set as a literal value in CI
> ```
> 
> Does that sound right? In some sense:
> 
> * `variables.SECURE_KEY` means "set as variable reference in the CI pipeline", but
> * `secrets.SECURE_KEY` means "set as literal value in the CI pipeline".

That´s correct.
There is a longer explanation and some suggestions about when you might want to use one or the other in the doc file: https://github.com/Azure/azure-dev/pull/4770/files#diff-e092bdc598ea8a281090500ff50bddfe2d152e7af4364fea2af7eec9da2228a8R92

The main thing to decide is if you want/can to grant Read Access to your Service Principal for the Key Vault to get the benefit of azd auto-secret resolution on CI and the ability to rotate secrets and don't worry about the pipelines.
Variables are easier to work with because it is fine to see its value. It actually helps you to see what akvs you are using during the logs.

PR #4771, opened 2025-02-07
Extensions: Auto install extensions during `azd init` command

wbreza:
# New features:
- Extensions can now be referenced directly within `azure.yaml` under the `requiredVersions.extensions` configuration element.
- Referenced extensions are automatically installed during `azd init` command when the extensions alpha feature is enabled.
- Extension dependencies now support semver constraints
- Minor updates on `azd extension` commands

## Why?
When a users project is dependent on `azd` extensions and another team member onboards or a new machine/workstation is used users will need to manually install the required extensions otherwise downstream failure will occur when those extensions are expected to be invoked.  With this feature, any time a template is initialized `azd` will ensure that any referenced extensions are installed with based on the required version constraints.

## Example 
```yaml
#azure.yaml

name: todo-nodejs-mongo-aca
metadata:
  template: todo-nodejs-mongo-aca@0.0.1-beta
requiredVersions:
  extensions:
    microsoft.azd.ai: ">=0.2.0" # Supports semver constraints
services:
  web:
    project: ./src/web
    language: js
    host: containerapp
  api:
    project: ./src/api
    language: js
    host: containerapp
 ```
 
 ### Usage: Runing `azd init`
 When running `azd init` the required extensions will automatically be inspected and install the latest version matching the project constraint.
 
image

## Potential Issues

A potential complication arises when different azd projects require conflicting extension versions. Since azd currently supports only global extension installations rather than per-project installations, a version mismatch could occur.

### DIFF
--- a/cli/azd/cmd/extension.go
+++ b/cli/azd/cmd/extension.go
@@ -146,6 +146,7 @@ func newExtensionListFlags(cmd *cobra.Command) *extensionListFlags {
 type extensionListAction struct {
 	flags            *extensionListFlags
 	formatter        output.Formatter
+	console          input.Console
 	writer           io.Writer
 	sourceManager    *extensions.SourceManager
 	extensionManager *extensions.Manager
@@ -154,6 +155,7 @@ type extensionListAction struct {
 func newExtensionListAction(
 	flags *extensionListFlags,
 	formatter output.Formatter,
+	console input.Console,
 	writer io.Writer,
 	sourceManager *extensions.SourceManager,
 	extensionManager *extensions.Manager,
@@ -161,6 +163,7 @@ func newExtensionListAction(
 	return &extensionListAction{
 		flags:            flags,
 		formatter:        formatter,
+		console:          console,
 		writer:           writer,
 		sourceManager:    sourceManager,
 		extensionManager: extensionManager,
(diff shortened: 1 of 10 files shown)

### CODE REVIEW

File: cli/azd/cmd/extension.go [hunk 1]
```diff
@@ -217,10 +218,29 @@ func (a *extensionListAction) Run(ctx context.Context) (*actions.ActionResult, e
...
 			Namespace: extension.Namespace,
 			Version:   version,
+			Source:    extension.Source,
 			Installed: installedExtensions[extension.Id] != nil,
 		})
 	}
 
+	if len(extensionRows) == 0 {
+		if a.flags.installed {
+			fmt.Println(output.WithWarningFormat("WARNING: No extensions installed.\n"))
```
vhvb1989, line 228: Curious about why you are using `fmt.Println`, I think we don't use. IIRC, we decided to use `fmt.Fprintln()` so we could set the io.Writer to control the `--output json` or things like that.

Can you use the `Console.Message()` here?

schemas/alpha/azure.yaml.json
```diff
@@ -504,6 +505,20 @@
                     "examples": [
                         ">= 0.6.0-beta.3"
                     ]
+                },
+                "extensions": {
+                    "type": "object",
+                    "title": "A range of supported versions of extensions for this project",
```
vhvb1989, line 511: Wondering if we should mention that whatever extension is listed here is a `must have` for init and use the template.  

For example, failing to install the extension blocks azd from initializing the template.  The extension is not really optional when it is listed here because the functionality of the template depends on the extension's existence.

Or, should we make `extensions` an object array where, for each extension you can define more things for the extension?  Like the version, if it is mandatory for the template or optional, and any other future setting?

wbreza: Agreed that if you specify it here that it should be considered required otherwise you're not going to be able to get the full benefit of the template without it.

cli/azd/cmd/extension.go
wbreza: Will update to use console.message.  Thanks for the reminder 🙏

vhvb1989 (approved):
LGTM

Issue #4381, opened 2024-09-26
package: support ignore files


weikanglim:
As authored by @jongio based on #4258. Related to #1039 , #4215

## Background

We currently support `.dockerignore` files when uploading source for remote docker builds. This feature enhancement tracks addition of ignore files for other packaging protocols.

## Work In-scope

@jongio had gotten agreements from service teams to add support for `.webappignore` and `.funcignore` files. These files are interpreted as standard .gitignore syntax. When present, these ignore files should be read as configuration to exclude files. The exclusion files apply relatively to the directory which the ignore file is present. Additional details of the implementation are specified below:

### Default behavior when ignore files not present

With `SCM_DO_BUILD_DURING_DEPLOYMENT=true` (for app service), `remoteBuild=true` (for func-flex-consumption), users would not expect the following directories to be packaged:
- `node_modules` (js)
- `__py_cache__`, `.venv` (python)

Thus, we should continue applying these default exclusion behaviors when ignore files are not present.

For function apps, `local.settings.json` is a local configuration file that should not be expected to be published by default. We should take this into account as well. This isn't backwards compatible -- but is the right thing to do since it has been a missing user-requested feature of azd.

## Explorations

1. Should the ignore logic differ based on `SCM_DO_BUILD_DURING_DEPLOYMENT`? See comment [here](https://github.com/Azure/azure-dev/issues/4215#issuecomment-2290825312)
2. What is the function tools team plan to address the non-standard interpretation of `.funcignore`? https://github.com/Azure/azure-functions-core-tools/issues/2701
3. Should we support ignore files in subdirectories? What are good use cases for this?

pamelafox:
## One use case for .webappignore:

We would like to support developers using NSGs in their VNet with Azure App Service. That's not currently possible since App Service build uses Oryx which calls out to a Oryx server that doesn't have a static IP address. If we could do the build locally, then we could disable the Oryx build, and be able to use NSGs. If we try that now, our venv won't be uploaded, since your default behavior is to ignore any folders that resemble Python virtual environments.
So we need webappignore so we can get azd to *not* ignore folders.

## Another use case:

We are supporting both ACA and App Service in our repo now. If we had a .webapp ignore, we could avoid uploading the Dockerfile to App Service. Not a huge deal since that's a tiny file. 

## One more use case:

Many devs write their tests/ folder inside the src/ folder, due to the difficulties with having tests/ outside of src/. Ideally those tests would not get uploaded to App Service.

[...]

PR #4773, opened 2025-02-07
rzip: handle symbolic linked directories

### LINKED ISSUE #4715: [Issue] Symlinks ignored when zipping for app service deployment

weikanglim:
In our previous implementation of `CreateFromDirectory`, symbolic linked directories were not handled correctly and would result in an `os.Stat` error as describe in #4715. Symbolic linked files were handled with `os.Stat` that follows the target.

This change adds support for symbolic linked directory traversal, and accompanying test suite.

Fixes  #4715, Contributes to #4381


cli/azd/pkg/rzip/rzip.go
```diff
@@ -6,55 +6,129 @@ package rzip
...
-		}
+	return w.Close()
+}
 
-		f, err := w.CreateHeader(header)
+func addDirRoot(
+	w *zip.Writer,
+	src string) error {
+	return addDir(w, "", src, 0)
+}
```
wbreza, line 34: nit: I don't think this would violate our line length rule - any reason for wrapping this function signature?

cli/azd/pkg/rzip/rzip.go
```diff
@@ -6,55 +6,129 @@ package rzip
...
+func addDir(
+	w *zip.Writer,
+	destRoot,
+	src string,
+	symlinkDepth int) error {
+	if symlinkDepth > 40 {
+		// too deep, bail out similarly to the 'zip' tool
+		log.Println("skipping", src, "too many levels of symbolic links")
+		return nil
+	}
```
wbreza, line 45: Just curious - Is `40` some well-known number to bail out?

cli/azd/pkg/rzip/rzip.go
```diff
@@ -6,55 +6,129 @@ package rzip
...
 	"os"
 	"path/filepath"
 	"strings"
 )
 
+// CreateFromDirectory creates a zip archive from the given directory recursively,
+// that is suitable for transporting across machines.
+//
+// It resolves any symlinks it encounters.
 func CreateFromDirectory(source string, buf *os.File) error {
```
wbreza, line 19: Is this common practice of Java or similar where symlinked folders would exist in some packaged artifact?

wbreza (approved):
A couple questions but otherwise looks good.

cli/azd/pkg/rzip/rzip.go
weikanglim: nothing in particular, just keeping it similar in style with other functions here

cli/azd/pkg/rzip/rzip.go
weikanglim: yeah, it's `MAXSYMLINKS` in the linux kernel

cli/azd/pkg/rzip/rzip.go
weikanglim: It's most common for python, but generally it can be used for monorepo / referencing source from other repos, #4715 was one example

vhvb1989 (approved):

PR #4775, opened 2025-02-08
Extensions: Adds initial `azd` demo extension

wbreza:
## Features
- Adds new demo extension to showcase the available integration capabilities.
- Fixes a couple minor bugs around subscription and resource selection

### `azd demo context`
image

### `azd demo prompt`
image


weikanglim (approved):

SophCarp:
Trying out the build and hit something - I'm not sure if this happened because of the PR or because of something else. Pretty much, I added the extension, then went and made an empty folder. I ran azd init to create a minimal template, and it responded with "panic". Not sure what the issue is, and it seems like an azure.yaml file was made anyway. 

Could this be because I'm using both composability and extensions?

image

SophCarp:
How do I add an extension? 

I've 
- enabled extensions as an alpha feature. 
- run `azd extensions list` and returned nothing
- run azd extensions source list` and just got azd as a source. 

image

Do I need to run azd extension source add? if so what details do I need?

PR #4781, opened 2025-02-11
Extensions: Adds registry generator to automate updates to registry.json

wbreza:
Adds registry generator tool.  The tool is used to build, package and update an `azd` extension registry with the updated extension metadata.

Additional the tool has been made into an extension itself.

## Example
- Install extension
 ```bash
 azd ext install azd.internal.registry
 ```

- Run extension
```bash
azd registry package -p <extension-path>
  -r <registry-path>
  -o <artifacts-output-path>
  --base-url <base-url-for-artifacts>
```

image


cli/azd/extensions/registry.json
```diff
@@ -1,4 +1,78 @@
...
-    "extensions": []
+  "extensions": [
+    {
+      "id": "microsoft.azd.demo",
+      "namespace": "demo",
+      "displayName": "Demo Extension",
+      "description": "This extension provides examples of the AZD extension framework.",
+      "versions": [
+        {
+          "version": "0.0.2",
```
weikanglim, line 10: Just wondering here if you have ideas around when a version could be removed, since it may be a maintenance burden for us.

I am familiar with krew for kubectl, which only registers the latest version since extensions are meant to be slightly more volatile, and the strategy they advise users is "always use latest". Example [here](https://github.com/kubernetes-sigs/krew-index/blob/master/plugins/browse-pvc.yaml)

cli/azd/extensions/azd.internal.registry/internal/cmd/root.go
```diff
@@ -0,0 +1,27 @@
...
+
+package cmd
+
+import (
+	"github.com/spf13/cobra"
+)
+
+func NewRootCommand() *cobra.Command {
+	rootCmd := &cobra.Command{
+		Use:           "azd registry <command> [options]",
```
weikanglim, line 12: Just wondering if the `azd` prefix is expected here. Feel free to address this separately.

I could imagine that the extension framework automatically appends "azd" to the `Use` here. For examples, it makes sense that the extension needs to include the full invocation path.

weikanglim (approved):
LGTM.

Just one small question for you @wbreza:

image

The usage printed out here is `azd package` -- I could imagine this is an artifact of using a debug build, but just thought I'd mention it.

cli/azd/extensions/azd.internal.registry/internal/cmd/root.go
wbreza: Most of this comes down to the correct usage of cobra or other CLI toolkits for constructing the correct help experience for the custom extension commands.  The goal is for the help experience to still show `azd <extension-namespace> <command> [options]` usage patterns for commands but I just haven't figured out the correct configuration here.

This is a great question that we will figure out over time and maybe the extension framework can help compose this together better.

cli/azd/extensions/registry.json
wbreza: Another great question.  We'll need to have some ability to deprecate old versions that we don't want users to install.  Lets discuss and we can figure something out later.

PR #4797, opened 2025-02-14
useragent: onboard azure agent

weikanglim:
Add environment detection for `@azure` agent

### DIFF
--- a/cli/azd/internal/tracing/fields/fields.go
+++ b/cli/azd/internal/tracing/fields/fields.go
@@ -136,9 +136,10 @@ const (
 	// Environments that are wrapped by an intermediate calling program, and are significant enough to warrant
 	// being an environment and not an environment modifier.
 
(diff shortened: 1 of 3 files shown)


rajeshkamal5050 (approved):

vhvb1989 (approved):
Curious. Why are we adding this detection for?

kristenwomack (approved):
LGTM cc: @AngelosP

kristenwomack:
> Curious. Why are we adding this detection for?

For our collaborators working on the Azure dev agent to see usage telemetry.

AngelosP:
Confused

rajeshkamal5050:
@AngelosP clarified this internally. Just adding details here too.

I believe we should be okay to inject upstream tools calling azd and passing on caller user agent string. Since azd internally concatenates to azdev/{version} {caller user agent }

@weikanglim @kristenwomack

PR #4806, opened 2025-02-19
Fix error: failed to get effective pom for multiple module pom

rujche:
### Fix error: failed to get effective pom for multiple module pom

For a pom which contains multiple modules:
1. Before current PR, the `mvn help:effective-pom -f xxx/pom.xml` will output effective pom for both root pom and child poms. The `getEffectivePomFromConsoleOutput` func will always return the last one. But the actually, the effective pom of root pom is not always appear at last.
2. After current PR, the `mvn help:effective0pom -f xxx/pom.xml -pl pom.xml` will only output effective pom for root pom. Link to "-pl" related doc: https://maven.apache.org/ref/3.1.0/maven-embedder/cli.html

### DIFF
--- a/cli/azd/pkg/tools/maven/maven.go
+++ b/cli/azd/pkg/tools/maven/maven.go
@@ -247,33 +247,34 @@ func (cli *Cli) EffectivePom(ctx context.Context, pomPath string) (string, error
 		return "", err
 	}
 	pomDir := filepath.Dir(pomPath)
-	runArgs := exec.NewRunArgs(mvnCmd, "help:effective-pom", "-f", pomPath).WithCwd(pomDir)
+	runArgs := exec.NewRunArgs(mvnCmd, "help:effective-pom", "-f", pomPath, "-pl", filepath.Base(pomPath)).WithCwd(pomDir)
 	result, err := cli.commandRunner.Run(ctx, runArgs)
 	if err != nil {
 		return "", fmt.Errorf("mvn help:effective-pom on project '%s' failed: %w", pomPath, err)
 	}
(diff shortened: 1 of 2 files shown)


cli/azd/pkg/tools/maven/maven.go
```diff
@@ -247,34 +247,35 @@ func (cli *Cli) EffectivePom(ctx context.Context, pomPath string) (string, error
 		return "", err
 	}
 	pomDir := filepath.Dir(pomPath)
-	runArgs := exec.NewRunArgs(mvnCmd, "help:effective-pom", "-f", pomPath).WithCwd(pomDir)
+	runArgs := exec.NewRunArgs(mvnCmd, "help:effective-pom", "-f", pomPath, "-pl", filepath.Base(pomPath)).WithCwd(pomDir)
```
weikanglim, line 250: Just curious, what does `-pl` do here? I haven't found the documentation for it in the official docs: https://maven.apache.org/plugins/maven-help-plugin/effective-pom-mojo.html

weikanglim: I think this is the required flag to isolate the pom.xml for multimodule builds? If so, feel free to drop a link to documentation or souce -- just want to get a better understanding of how this is used.

rujche: Oh... Sorry that I didn't make this clear before. Here is the link to "-pl" related doc: https://maven.apache.org/ref/3.1.0/maven-embedder/cli.html

And I added this link to the PR description and comment of source code.

PR #4842, opened 2025-02-25
compose: `azd add` support for Key Vault

### LINKED ISSUE #4682: compose: `azd add` support for Key Vaults
------------
Edge cases:

- Empty -> add MongoDB or Redis
  -  Preview provision should show KV
- KV exists -> add MongoDB or Redis
  -  Preview provision should not show KV?
- Empty -> add Postgres or other

JeffreyCA:
Closes #4682 

This PR adds `azd add` support for Azure Key Vault. Prior to these changes, the generated Bicep code was already provisioning a Key Vault behind the scenes to store database secrets. These changes make Key Vault an explicit resource type in `azure.yaml` that can be directly added using `azd add`.

When a database resource like MongoDB is added through `azd add`, a Key Vault resource named `vault` is also added to `azure.yaml` because the database resource is dependent on Key Vault. A resource type has a dependency on Key Vault if secrets belonging to that resource are stored in it. The following resource types currently have this dependency:

| Resource type | Key Vault secret name |
|--------------|---------------------|
| `db.postgres` | `postgres-password` |
| `db.mysql` | `mysql-password` |
| `db.redis` | `redis-password`, `redis-url` |
| `db.mongo` | `mongodb-url` |

## azure.yaml schema

```yaml
resources:
  vault
    type: keyvault
```

## Binding variables

`AZURE_KEY_VAULT_ENDPOINT` - Key Vault endpoint
`AZURE_KEY_VAULT_NAME` - Key Vault name

The above variables are available to consume from the host container app, regardless of whether the host resource has `vault` listed under **uses:** in `azure.yaml`. This was primarily done to not break existing scenarios.

### Host + KV only

<table>
<tr>
<td> Variant 1 </td> <td> Variant 2 </td>
</tr>
<tr>
<td>

```yaml
resources:
  app:
    type: host.containerapp
    port: 80
  vault:
    type: keyvault
```

</td>
<td>
    
```yaml
resources:
  app:
    type: host.containerapp
    port: 80
    uses:
      - vault
  vault:
    type: keyvault
```

</td>
</tr>
</table>

Each of the above `azure.yaml` snippets produces the same infra Bicep:

- Key Vault resource is created
- Container app environment variables include `AZURE_KEY_VAULT_ENDPOINT` and `AZURE_KEY_VAULT_NAME`

### Host uses KV-dependent resource

<table>
<tr>
<td> Variant 1 </td> <td> Variant 2 </td> <td> Variant 3 </td>
</tr>
<tr>
<td>
    
```yaml
resources:
  app:
    type: host.containerapp
    port: 80
    uses:
      - mdb
      - vault
  mdb:
    type: db.mongo
  vault:
    type: keyvault
```

</td>
<td>

```yaml
resources:
  app:
    type: host.containerapp
    port: 80
    uses:
      - mdb
  mdb:
    type: db.mongo
```

</td>
<td>
    
```yaml
resources:
  app:
    type: host.containerapp
    port: 80
    uses:
      - mdb
  mdb:
    type: db.mongo
  vault:
    type: keyvault
```

</td>
</tr>
</table>

> [!NOTE]
> **Variant 1** is what's generated by `azd add` to add a MongoDB resource that is **used** by the container app host.

Each of the above `azure.yaml` snippets also produces the same infra Bicep:

- Key Vault resource is created
- Key Vault secret is created with the MongoDB connection string
- Container app secret is created with a Key Vault reference to the above secret, referenced through `MONGODB_URL` environment variable
- Container app environment variables also include `AZURE_KEY_VAULT_ENDPOINT` and `AZURE_KEY_VAULT_NAME`

### KV-dependent resource defined but not used by host

<table>
<tr>
<td> Variant 1 </td> <td> Variant 2 </td> <td> Variant 3 </td>
</tr>
<tr>
<td>
    
```yaml
resources:
  app:
    type: host.containerapp
    port: 80
  mdb:
    type: db.mongo
  vault:
    type: keyvault
```

</td>
<td>
    
```yaml
resources:
  app:
    type: host.containerapp
    port: 80
    uses:
      - vault
  mdb:
    type: db.mongo
  vault:
    type: keyvault
```

</td>
<td>

```yaml
resources:
  app:
    type: host.containerapp
    port: 80
  mdb:
    type: db.mongo
```

</td>
</tr>
</table>

> [!NOTE]
> **Variant 1** is what's generated by `azd add` to add a MongoDB resource that **isn't used** by the container app host.

Each of the above `azure.yaml` snippets also produces the same infra Bicep:

- Key Vault resource is created
- Key Vault secret is created with the MongoDB connection string
- Container app secret **not** created with a Key Vault reference to the above secret
    - Environment variable `MONGODB_URL` **not** created either
- Container app environment variables include `AZURE_KEY_VAULT_ENDPOINT` and `AZURE_KEY_VAULT_NAME`

## Demo
https://github.com/user-attachments/assets/d4d43008-94c8-4962-8b3b-9210b97ba6ea


rajeshkamal5050 (approved):

rajeshkamal5050:
@JeffreyCA does this also cover the all language simplified-init flow and also the SJAD flow for adding key-vault as first-class resources in azure.yaml?

cc @weikanglim @saragluna

cli/azd/internal/cmd/add/add.go
```diff
@@ -300,21 +367,23 @@ func (a *AddAction) Run(ctx context.Context) (*actions.ActionResult, error) {
...
 
-		return &actions.ActionResult{
-			Message: &actions.ResultMessage{
-				FollowUp: "Run '" +
-					color.BlueString(fmt.Sprintf("azd show %s", resourceToAdd.Name)) +
-					"' to show details about the newly provisioned resource.",
-			},
-		}, nil
+		followUpMessage = "Run '" +
+			color.BlueString(fmt.Sprintf("azd show %s", resourceToAdd.Name)) +
```
wbreza, line 371: Nit: We have predefined format for common colors used in the CLI.  This should likely use `output.WithHighlightFormat(...)`

https://github.com/Azure/azure-dev/blob/main/cli/azd/pkg/output/colors.go#L18

wbreza (approved):
Looks good!

cli/azd/pkg/project/scaffold_gen.go
```diff
@@ -385,3 +392,40 @@ func genBicepParamsFromEnvSubst(
...
+	switch resource.Type {
+	case ResourceTypeDbMongo, ResourceTypeDbMySql, ResourceTypeDbPostgres, ResourceTypeDbRedis:
+		return []*ResourceConfig{{Name: "key-vault", Type: ResourceTypeKeyVault}}
+	default:
+		return nil
+	}
+}
+
+// WithResolvedDependencies returns a map of resource configs with all dependencies resolved.
+func WithResolvedDependencies(resources map[string]*ResourceConfig) map[string]*ResourceConfig {
```
weikanglim, line 409: I don't feel particularly good about this function living here as part of the API. I'm wondering if we can inline this function inside `infraSpec` to start with.

Where I'm coming from: if we were to do extra pre-processing that weren't just "dependent resources", we'd invent another function that does another pass through the resources list to do yet another thing. With that in mind, this is probably closer to a "pre-processor" path of a compiler, and I'd like that more.

For now, I just recommend adding an extra for loop inside `infraSpec`. I would recommend using `maps.Clone` to do a deep copy to create the "virtual resources list" that we're generating.

I don't know if the below feels right to you, please feel free to let me know in any shape or form.

```golang
func infraSpec(projectConfig *ProjectConfig) (*scaffold.InfraSpec, error) {
        ...
        // Create a "virtual" copy since we're adding any implicitly dependent resources that is unrepresented by the current user-provided schema
	resources := maps.Clone(projectConfig.Resources)

        // Add any implicit dependencies
	for _, res := range resources {
		dependencies := DependentResourcesOf(res)
		for _, dep := range dependencies {
			if _, exists := resources[dep.Name]; !exists {
				resources[dep.Name] = dep
			}
		}
	}

        // Map to infra spec
	for _, res := range resources {
```

cli/azd/internal/cmd/add/add_preview.go
```diff
@@ -103,6 +109,7 @@ func (a *AddAction) previewProvision(
 	ctx context.Context,
 	prjConfig *project.ProjectConfig,
 	resourceToAdd *project.ResourceConfig,
+	depsToAdd []*project.ResourceConfig,
```
weikanglim, line 112: I could see that we can merge the two here and just make `resourcesToAdd`

cli/azd/internal/cmd/add/add.go
```diff
@@ -174,6 +204,21 @@ func (a *AddAction) Run(ctx context.Context) (*actions.ActionResult, error) {
...
 			return nil, fmt.Errorf("appending resource: %w", err)
 		}
+
+		// Add dependent resources to 'uses'
+		for _, dep := range dependentResources {
+			// Skip if already in 'uses'
+			if slices.Contains(prjConfig.Resources[svc].Uses, dep.Name) {
+				continue
+			}
+			err = yamlnode.Append(&doc, fmt.Sprintf("resources.%s.uses[]?", svc), &yaml.Node{
```
weikanglim, line 214: I feel that a discussion is warranted on whether `uses` would mean the keyvault relationship here. Perhaps we should restrict `uses` in the schema for now to just "host" applications to mean "connects to resources"

cli/azd/internal/cmd/add/add.go
```diff
@@ -166,6 +167,35 @@ func (a *AddAction) Run(ctx context.Context) (*actions.ActionResult, error) {
 		return nil, fmt.Errorf("setting resource: %w", err)
 	}
 
+	// Dependent resources (both existing and to be added)
+	dependentResources := project.GetRequiredDependencies(resourceToAdd)
+	// New dependent resources to be added
+	dependentResourcesToAdd := make([]*project.ResourceConfig, 0)
```
weikanglim, line 173: I suspect the code actually looks "simpler" if we just treat `resourcesToAdd` as an array that gets filled with dependencies.

And to track the dependency relationship (used various places for printing), we can use a `map[*project.ResourceConfig]*project.ResourceConfig`

weikanglim (commented):
Overall looks good! Did an initial review, leaving some comments to discuss on:

1. If we want `uses` to represent cosmos -> key-vault.
2. If the Bicep templating logic looks right based on the generation layer

And some minor code suggestions

JeffreyCA:
With these changes, if you add a KV-dependent resource (any database) and provision it, we print out a message like this at the end:

```
Run 'azd show mysqldb' to show details about the newly provisioned resource.
Run 'azd env set-secret <name>' to add a secret to the key vault.
```

Should we also include 'azd show vault' there (or any newly added dependent resource)? And if so, how should the message look like?

cli/azd/internal/cmd/add/diff.go
```diff
@@ -89,7 +89,7 @@ func DiffBlocks(old map[string]*project.ResourceConfig, new map[string]*project.
 		sb.WriteString("\n")
 	}
 
-	return sb.String(), nil
+	return strings.TrimSuffix(sb.String(), "\n"), nil
```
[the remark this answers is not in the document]
JeffreyCA: The trailing `\n` results in an extra line between the diff and subsequent content:

image

cli/azd/pkg/project/resources.go
```diff
@@ -35,6 +36,7 @@ const (
 	ResourceTypeMessagingEventHubs  ResourceType = "messaging.eventhubs"
 	ResourceTypeMessagingServiceBus ResourceType = "messaging.servicebus"
 	ResourceTypeStorage             ResourceType = "storage"
+	ResourceTypeKeyVault            ResourceType = "key-vault"
```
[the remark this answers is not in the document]
JeffreyCA: Yes I'm fine with that

cli/azd/resources/scaffold/templates/resources.bicept
```diff
@@ -604,7 +617,7 @@ module redis 'br/public:avm/res/cache/redis:0.9.0' = {
     // Non-required parameters
     location: location
     skuName: 'Basic'
-    {{- if .Services}}
+    {{- if .KeyVault}}
```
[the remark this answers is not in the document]
JeffreyCA: Yes that makes sense. Thanks for pointing that out!

cli/azd/pkg/project/scaffold_gen.go
JeffreyCA: Yes I think that makes sense.

> I would recommend using maps.Clone to do a deep copy to create the "virtual resources list" that we're generating.

I'm reading that `maps.Clone` does the opposite (i.e. a shallow clone) - could you clarify what you meant by doing a deep copy?

cli/azd/internal/cmd/add/add.go
JeffreyCA: Good point! Yes it did simplify things quite a bit, and I was able to forgo the dependency map

JeffreyCA:
> @JeffreyCA does this also cover all language simplified-init flow and also the SJAD flow for adding key-vault as first-class resources in azure.yaml? Not blocking, but may be a follow-up change, if need be.
> 
> cc @weikanglim @saragluna

No, this does not handle auto-detection of KV libraries during simplified init. What's the SJAD flow you're referring to?

cli/azd/pkg/project/scaffold_gen.go
weikanglim: Ah, I do mean just having a separate "map copy" so we can add an extra entry without mutating the original. `maps.Clone` would be fine for this purpose right now.

cli/azd/resources/scaffold/templates/resources.bicept
```diff
@@ -550,6 +549,16 @@ module {{bicepName .Name}} 'br/public:avm/res/app/container-app:0.8.0' = {
             value: storageAccount.outputs.serviceEndpoints.blob
           }
           {{- end}}
+          {{- if $infra.KeyVault}}
```
weikanglim, line 552: What are your thoughts on if the service should always be allowed access to key vault? Or is this a case of `uses`? I'm debating this myself

cli/azd/pkg/project/scaffold_gen_test.go
```diff
@@ -74,3 +74,70 @@ func Test_genBicepParamsFromEnvSubst(t *testing.T) {
 		})
 	}
 }
+
+func Test_DependentResourcesOf(t *testing.T) {
```
weikanglim, line 78: Out of curiosity, do you still find value in this test after the recent changes? I understand this as an expectation test, and could easily lean towards removing it altogether given the implementation being generally *trivial*

cli/azd/internal/cmd/add/diff.go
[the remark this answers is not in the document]
weikanglim: I see. Looking into it more, I could see this as an artifact of `console.Message` always appending an extra newline.

Supposedly from a library perspective, I could imagine that the newline being an expected piece of the diff output, but I think this makes sense overall given our constraints.

cli/azd/resources/scaffold/templates/resources.bicept
JeffreyCA: With these proposed changes, Key Vault is kind of being treated as a special case where `uses: vault` doesn't really do anything differently.

If `uses` determines whether `AZURE_KEY_VAULT_NAME`, `AZURE_KEY_VAULT_ENDPOINT` are set in the container app, then there'd be more cases to handle like what if `vault` isn't explicitly defined in `azure.yaml` a database resource is - would that mean `AZURE_KEY_VAULT_NAME` isn't set?

I could also see making key vault an implicit dependency of container app, then this `{{- if $infra.KeyVault}}` wouldn't be needed.

cli/azd/pkg/project/scaffold_gen_test.go
JeffreyCA: Yeah I had the same thought - I can remove it

cli/azd/internal/cmd/add/add_configure.go
```diff
@@ -58,6 +58,16 @@ func Configure(
 		return r, nil
 	case project.ResourceTypeStorage:
 		return fillStorageDetails(ctx, r, console, p)
+	case project.ResourceTypeKeyVault:
+		if _, exists := p.PrjConfig.Resources["vault"]; exists {
+			return nil, fmt.Errorf(
+				"you already have a project key vault named 'vault'. " +
+					"To add a secret to it, run 'azd env set-secret <name>'",
+			)
+		}
```
vhvb1989, line 67: `project key vault` ?  Why do we call it a project?

If we are trying to limit having just one only key vault, should we check for the type?  I think folks can manually add a key vault resource with a name like `kv-foo` and this condition wouldn't catch it

vhvb1989 (approved):

cli/azd/internal/cmd/add/add_configure.go
JeffreyCA: > `project key vault` ? Why do we call it a project?

That's what we landed on in our initial designs, but open to calling it something else

> If we are trying to limit having just one only key vault, should we check for the type? I think folks can manually add a key vault resource with a name like kv-foo and this condition wouldn't catch it

Yes that's a limitation of resource types in composability that support at most 1 resource today. Similar checks exist for other resource types:

image

Want to add anything here @weikanglim?

cli/azd/internal/cmd/add/add.go
JeffreyCA: Thanks, I'll address this in https://github.com/Azure/azure-dev/pull/4847

cli/azd/internal/cmd/add/add_configure.go
weikanglim: FWIW, @JeffreyCA, we do have an extra safeguard in `scaffold_gen.go` for `infra synth` to generate an error for multiple resources of a particular type. But here, the idea is presented along the lines of "there is one primary azd resource called `vault`" and that is special.

I'm not sure if it's helpful to orient ourselves along the lines that we're trying to simplify a building experience for a new user to Azure.

In these scenarios, the user doesn't really need more than one of anything -- they really just see Azure as a cloud provider that allows them to use different types of functionality.

cli/azd/resources/scaffold/templates/resources.bicept
weikanglim: That's fine either way, I think we can leave it as-is for now.

cli/azd/internal/cmd/add/add.go
weikanglim: I think it would be better to avoid automatically adding `uses`.. For vault in particular, let's just keep it as an implicit "uses".

Generally speaking, the dependency relationship ought not to always be structured that way, so I have a slight preference here to remove this logic.

weikanglim (approved):
LGTM! Thanks for getting this through!

cli/azd/internal/cmd/add/add.go
JeffreyCA: Updated, thanks. Also realized with the current `resources.bicept` this isn't really doing anything either.

PR #4847, opened 2025-02-25
Align use of colours with design pattern guide

### LINKED ISSUE #4846: UX: Prompt navigation hints and responses should be blue instead of cyan
Elements shown below should be blue:
Image

JeffreyCA:
Fixes #4846 to align with our design pattern guide:

image

This PR also replaces several instances of `color.XYZString` with the appropriate wrapper functions from [pkg/output/colors.go](https://github.com/Azure/azure-dev/blob/main/cli/azd/pkg/output/colors.go):

https://github.com/Azure/azure-dev/blob/fbe4819b0c602b8da0c9bc04b219d051cf59130f/cli/azd/pkg/output/colors.go#L12-L24

### `azd help`

image
image

### `azd init`
image
image

### `azd add`
image

image

### `azd show`
Difference is minor - we were using Bright Blue instead of Blue.

image


JeffreyCA:
We have a few usages of magenta in a few places, which I didn't touch in this PR and am open to further discussion on:
image

image

I also left all the "UX v2" components introduced by https://github.com/Azure/azure-dev/pull/4696 under `cli/azd/pkg/ux` and the demo extension under `/cli/azd/extensions/microsoft.azd.demo/internal/cmd`. They're currently directly using the `github.com/fatih/color` package instead of the wrapper functions in https://github.com/Azure/azure-dev/blob/main/cli/azd/pkg/output/colors.go.

@wbreza / @SophCarp do you know if there are any plans to do so? Are they already consistent with the design guide? Or will there be a new colour scheme?

cli/azd/pkg/input/asker.go
```diff
@@ -268,10 +268,20 @@ func init() {
...
+
+	// Replace cyan with blue in other other templates
+	otherTemplates := []*string{
+		&survey.ConfirmQuestionTemplate,
+		&survey.SelectQuestionTemplate,
+		&survey.MultiSelectQuestionTemplate,
+	}
+	for _, template := range otherTemplates {
+		*template = strings.ReplaceAll(*template, `"cyan"`, `"blue"`)
+	}
```
weikanglim, line 286: I honestly wouldn't have thought about going this far.. but thos is actually pretty amazing and does work!!

weikanglim (approved):
LGTM! If we want to separate out the magenta changes as a separate change, I'm all for it

wbreza (approved):
Looks good, i would just verify the colors show as intended.  Blue tends to render much darker than the intended color from the screen shot in my opinion.

To me Cyan is better match for Blue and HiCyan a better match for Cyan.

SophCarp:
This is great, thank you Jeffrey! If we want a conversation about using Cyan over blue, we can. Currently the design we have states that Cyan is only for hyperlinks, and blue is for everything else. inconsistently using the colors is the main thing. I approve this work, and then we can take an action item to discuss the updating the color scheme designs

SophCarp (approved):

SophCarp:
I also don't know about the magenta. Will take a look

vhvb1989 (approved):

JeffreyCA:
Thanks for pointing that out, @wbreza! I can also confirm that the new blue is a bit darker in Windows Terminal and may be harder to read.

I went back and gathered screenshots for different combinations for the highlight and hyperlink colours in VS Code and Windows Terminal. Here's an comparison of the before/after along with other combinations we could use:
Theme Comparison

Personally, **Bright Blue** for highlights (commands, parameters, system inputs) and **Bright Cyan** for hyperlinks looks like a better combination that resembles the design guide the closest on both Windows Terminal and VS Code.

@SophCarp any thoughts? I'm also fine if we want to merge this with the current (non-bright) Blue / Cyan mapping.

SophCarp:
Love the comparison chart! Out of all of these, I think Bright Blue/Bright Cyan gives the best combo or readability and distinction. Can I get the updated PR to try out on my computer?

SophCarp:
Solarized light:
image

Solarized Dark:
image

Tango Dark: 
image

Ubuntu: 
image

Vintage: 
image

My main issue is that solarized themes grey out any high color. Also, apparently there is DevDiv design guidance for terminal coloring but the link is dead. I'm asking the design team where the resource got moved, and will get back to you!

SophCarp:
I'm not sure how common solarized themes are - I personally prefer them and use them, but I don't know how to check its popularity

JeffreyCA:
Thanks for your help looking into this @SophCarp! Pushed a fix so the prompt hints and responses are also bright blue:

image
image

Issue #3093, opened 2023-12-08
[ADO Pipelines] Create ado pipeline step to run azd with a service-connection, just like `AzureCLI@2`


vhvb1989:
based from: https://github.com/Azure/azure-dev/issues/3088

Create an Azure DevOps pipeline - task like `AzureCLI@2` for azd.
This task would allow folks to run azd commands using a service-connection like:

```
- task: AzureDeveloperCLI
    displayName: Provision Infrastructure
    inputs:
      azureSubscription: azconnection
      scriptType: bash
      scriptLocation: inlineScript
      inlineScript: |
        azd provision --no-prompt
    env:
      AZURE_SUBSCRIPTION_ID: $(AZURE_SUBSCRIPTION_ID)
      AZURE_ENV_NAME: $(AZURE_ENV_NAME)
      AZURE_LOCATION: $(AZURE_LOCATION)
```

The task should use the `azconnection` to run `azd auth login ...` before running the `inlineScript` and it should call `azd auth logout` at the end

KSemenenko:
Is there a way to do this right now? or we have to wait for this issue?

vhvb1989:
@KSemenenko ,  it is currently possible to use a service-connection for azd, but it requires setting azd to use `az` as its auth-provider. Here's an example:

```yaml
trigger:
  - main
  - master

# Azure Pipelines workflow to deploy to Azure using azd
# To configure required secrets for connecting to Azure, simply run `azd pipeline config --provider azdo`

pool:
  vmImage: ubuntu-latest

steps:
  - task: setup-azd@0 
    displayName: Install azd

  - pwsh: |
      azd config set auth.useAzCliAuth "true"
    displayName: Configure AZD to Use AZ CLI Authentication.

  - task: AzureCLI@2
    displayName: Provision Infrastructure
    inputs:
      # *****************THIS IS THE Service Connection here ***************
      azureSubscription: azconnection
      scriptType: bash
      scriptLocation: inlineScript
      inlineScript: |
        azd provision --no-prompt
    env:
      AZURE_SUBSCRIPTION_ID: $(AZURE_SUBSCRIPTION_ID)
      AZURE_ENV_NAME: $(AZURE_ENV_NAME)
      AZURE_LOCATION: $(AZURE_LOCATION)
      
```

Notes:
 - you don't need to run `azd auth login` in this mode. azd will ask `az` for auth-tokens, and `az` will use the service connection to authenticate to Azure.


[...]

PR #4850, opened 2025-02-26
set default for apphost confirmation

vhvb1989:
fix: https://github.com/Azure/azure-dev/issues/3093

Allow running `azd init --from-code --no-prompt` in CI to initialize and deploy Aspire Applications.

This should remove the need of running `azd init` like: 

```sh
yes "Confirm and continue initializing my app" | azd init --from-code
```


wbreza (approved):

weikanglim (approved):
LGTM

PR #4851, opened 2025-02-26
Enable run azd init from a Maven project's pom file on right click

saragluna:
This PR is to enable running `azd init` on a Java Maven project, when right click on the pom file. The expected output is the `azure.yaml` after analyzing the project. This PR changes the title of running `azd init` from `Initialize App` to `Generate Azure Deployment Script`.

Below are the screenshots:

1. Initialize from a pom file

image

2. Right click on `azure.yaml` (nothing changed)

image

### DIFF
--- a/ext/vscode/package.json
+++ b/ext/vscode/package.json
@@ -185,11 +185,16 @@
             "explorer/context": [
                 {
                     "submenu": "azure-dev.explorer.submenu",
-                    "when": "resourceFilename =~ /azure.yaml/i",
+                    "when": "resourceFilename =~ /(azure.yaml|pom.xml)/i",
                     "group": "azure-dev"
                 }
             ],
             "azure-dev.explorer.submenu": [
+                {
+                    "when": "resourceFilename =~ /pom.xml/i",
+                    "command": "azure-dev.commands.cli.init",
+                    "group": "10provision@10"
+                },
                 {
                     "when": "resourceFilename =~ /azure.yaml/i",
                     "command": "azure-dev.commands.cli.provision",
--- a/ext/vscode/package.json
+++ b/ext/vscode/package.json
@@ -171,6 +171,10 @@
         "menus": {
             "commandPalette": [
                 {
+                    "command": "azure-dev.commands.cli.initFromPom",
+                    "when": "false"
+                },
+                {
                     "command": "azure-dev.commands.getDotEnvFilePath",
                     "when": "false"
                 },
--- a/ext/vscode/package.json
+++ b/ext/vscode/package.json
@@ -40,6 +40,11 @@
             },
             {
                 "category": "%azure-dev.commands_category%",
+                "command": "azure-dev.commands.cli.initFromPom",
+                "title": "%azure-dev.commands.cli.initFromPom.title%"
+            },
+            {
+                "category": "%azure-dev.commands_category%",
                 "command": "azure-dev.commands.cli.provision",
                 "title": "%azure-dev.commands.cli.provision.title%"
             },
@@ -192,7 +197,7 @@
             "azure-dev.explorer.submenu": [
                 {
                     "when": "resourceFilename =~ /pom.xml/i",
-                    "command": "azure-dev.commands.cli.init",
+                    "command": "azure-dev.commands.cli.initFromPom",
                     "group": "10provision@10"
                 },
                 {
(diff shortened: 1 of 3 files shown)


saragluna:
Involve @taoxu0903.

bwateratmsft (commented):
@saragluna can you include a screenshot of the menu in action?

bwateratmsft (commented):
@kristenwomack I don't see a problem with this change but I want to get some PM input from you.

ext/vscode/package.nls.json
```diff
@@ -1,7 +1,7 @@
 {
     "azure-dev.commands_category": "Azure Developer CLI (azd)",
 
-    "azure-dev.commands.cli.init.title": "Initialize App (init)",
+    "azure-dev.commands.cli.init.title": "Generate Azure Deployment Script (init)",
```
bwateratmsft, line 4: I am a little concerned that this will cause confusion because it changes the command title in the command palette too.

weikanglim: @bwateratmsft Do you have recommendations that would match naming conventions here?

I could see something like `azure-dev.commands.cli.initFromPom` to describe the scenario better

bwateratmsft: That sounds fine

weikanglim, line 4: From azd's perspective, maybe something like "Intialize App for Azure"? or "App" is not quite the right word here for Java context, how about "Initialize for Azure"?

saragluna: @taoxu0903, what's your opinion?

saragluna:
> @saragluna can you include a screenshot of the menu in action?

Updated the description of this issue.

ext/vscode/package.nls.json
saragluna: Changed. I have another question about the telemetry, can we add another entry for `initFromPomCli` in this file https://github.com/Azure/azure-dev/blob/22222bd5f3fd5342f39e3d6dd55111e34c5d1c84/ext/vscode/src/telemetry/telemetryId.ts#L53, to track the usage of `initFromPom` in VSCode?

taoxu0903: @saragluna LGTM, "Initialize for Azure" is more consistent naming with other names in AZD extension.

bwateratmsft: The `registerActivityCommand` call wraps the command with telemetry tracking, so that's already done. It'll show up in telemetry under the event `ms-azuretools.azure-dev/azure-dev.commands.cli.initFromPom`.

kristenwomack: `initFromPom` is clearer and more descriptive, nice suggestion @weikanglim

kristenwomack:
> @kristenwomack I don't see a problem with this change but I want to get some PM input from you.

with the changes, it looks good. I'm onboard with this.

ext/vscode/package.json
```diff
@@ -38,6 +38,11 @@
                 "command": "azure-dev.commands.cli.init",
                 "title": "%azure-dev.commands.cli.init.title%"
             },
+            {
```
bwateratmsft, line 41: We should probably hide this command from the command palette, so that it doesn't cause confusion by seeing two entries for init. You can do that by adding an entry in `contributes.menus.commandPalette`. See `azure-dev.commands.getDotEnvFilePath` for an example.

vhvb1989 (approved):

kristenwomack (approved):
LGTM

ext/vscode/package.json
saragluna: Thanks @bwateratmsft! Without hiding it I can see:

image

But after adding the following to `contributes.menus.commandPalette`, there's only one entry now:

```json
 {
   "command": "azure-dev.commands.cli.initFromPom",
   "when": "false"
 },
```

image

bwateratmsft (approved):

bwateratmsft:
@saragluna when does this need to be released?

saragluna:
@bwateratmsft when will the next scheduled release be? Will there be a release note, we want to hold this change to be mentioned later when our public preview is ready, which will be in May.

bwateratmsft:
@saragluna there are not any releases scheduled but we can release pretty much any time we want to. We don't have to mention anything in the release notes if we don't want to.

saragluna:
@bwateratmsft thanks, I will discuss with @taoxu0903 and update you later.

PR #4862, opened 2025-02-27
UX Multi Select Component (v2)

wbreza:
## What's Included

- New Multi Select Component

## Example
**Cambell Theme**

image

**Dark Theme**

https://github.com/user-attachments/assets/d6b73c80-7444-4390-8ac1-85ac05fec54d

## Usage

Example `go` code to use the component.

```go
choices := []*ux.MultiSelectChoice{}
for i := 1; i <= 10; i++ {
  choices = append(choices, &ux.MultiSelectChoice{
	Key:   fmt.Sprintf("options-%d", i),
	Value: fmt.Sprintf("Option %d", i),
  })
}

choices[0].Selected = true

options := &ux.MultiSelectOptions{
  Message:        "Select an option",
  Allowed:        choices,
  HelpMessage:    "Help message",
  DisplayNumbers: ux.Ptr(true),
}

multiSelect := ux.NewMultiSelect(options)
selectedChoices, err := multiSelect.Ask()
```


SophCarp:
UX Notes: 

image
- Is the line between the list and the help text a common UI point in this new style? I haven't seen it anywhere else in azd, so probably would avoid it. 

- If the list is less than 10, let's avoid the "..." and keep the list static. 

image
- I like adding each item that's selected to the list! Should that text be gray (that's usually the customer input text) rather than blue? Usually if I'm asked for an input, I write it in gray and only once I've submitted the input the text turns blue. 

- Wording-wise, I think the question should be phrased in a way its clear you can select multiple options. Maybe "select one or more options".


All in all, I like this!

cli/azd/pkg/ux/multi_select.go
```diff
@@ -0,0 +1,466 @@
...
+
+func (p *MultiSelect) renderFooter(printer Printer) {
+	if p.cancelled || p.complete {
+		return
+	}
+
+	printer.Fprintln()
+	printer.Fprintln(color.HiBlackString("───────────────────────────────────"))
+	printer.Fprintln(color.HiBlackString("Use ↑/↓ to move, <space> to select"))
+	printer.Fprintln(color.HiBlackString("Use <enter> to submit, <?> for help"))
```
ccbarragan, line 465: @wbreza @SophCarp For the instructions text, suggest replacing symbols with words where we can to make it screen reader accessible. I believe this was one of the learnings from the [accessibility user study](https://microsoft-my.sharepoint.com/:p:/p/austinauth/EcEzJhCzWbJJsuNKrTtVZRwBdr3LirMZ3RDZSoHAA2qGcA?e=uw6Tej) from last year.
```suggestion
	printer.Fprintln(color.HiBlackString("Use arrow keys to move, use space to select"))
	printer.Fprintln(color.HiBlackString("Use enter to submit, type ? for help"))
```

cli/azd/pkg/ux/multi_select.go
```diff
@@ -0,0 +1,466 @@
...
+}
+
+func (p *MultiSelect) renderFooter(printer Printer) {
+	if p.cancelled || p.complete {
+		return
+	}
+
+	printer.Fprintln()
+	printer.Fprintln(color.HiBlackString("───────────────────────────────────"))
+	printer.Fprintln(color.HiBlackString("Use ↑/↓ to move, <space> to select"))
```
JeffreyCA, line 464: Have we considered including gestures for "Select All" and "Deselect All"?

I see the existing Multi Select component has them, which are mapped to the right and left arrow keys:
image

cli/azd/pkg/ux/multi_select.go
ccbarragan: Also worth noting from the accessibility findings:

>Our wording of the instructions add confusion to the interaction (specifically "press enter to confirm"). Suggested fix: "Remove 'press enter to select' since that is just default terminal behavior"


https://microsoft-my.sharepoint.com/:p:/p/austinauth/EcEzJhCzWbJJsuNKrTtVZRwBdr3LirMZ3RDZSoHAA2qGcA?e=A0vvfM

wbreza:
> Is the line between the list and the help text a common UI point in this new style? I haven't seen it anywhere else in azd, so probably would avoid it.

It was in a design that @Austinauth did at one point and it looked nice.  Open to suggestions

> If the list is less than 10, let's avoid the "..." and keep the list static.

This is configurable by the developer, defaults to 6 at the moment.

> I like adding each item that's selected to the list! Should that text be gray (that's usually the customer input text) rather than blue? Usually if I'm asked for an input, I write it in gray and only once I've submitted the input the text turns blue.

Happy to make the change - lets get consensus from you all and happy to make that happen.

> Wording-wise, I think the question should be phrased in a way its clear you can select multiple options. Maybe "select one or more options".

The input message is configurable by the developer.

Austinauth:
**Re: Horizonal line break**
Feel free to remove. I'd opt to go with whatever is most consistent with the pattern refresh that's being worked on. Essentially, use it for all or use it for none.

SophCarp:
Added a commit with some changes.

SophCarp:
Screenshot 2025-02-28 130433
Bug: When I filter for something, I can't navigate up and down the list. For example, I filtered for "1", but then if I tried to move down to "10" I couldn't select it

SophCarp:
Thoughts on the color of the selected choices: I tested it out with them gray, and I think I prefer them blue. Input that the customer types can be gray then turn blue when submitted. The list of selected options isn't the same as text the customer writes, and it makes sense to have it be highlighted in blue for the customer to easily see.

SophCarp:
I also tried out changing the indentation (which is why there are some commented out lines). After trying some variations, I decided the indentation you set looks the best.

SophCarp:
I removed the filter ghost text and added it to the bottom instructions. Ghost text has some accessibility issues, and this way all the instructions are in one space.

vhvb1989 (approved):

kristenwomack (approved):
LGTM, and we can iterate on changes from here

cli/azd/pkg/input/console.go
```diff
@@ -407,6 +408,9 @@ func (c *AskerConsole) ShowSpinner(ctx context.Context, title string, format Spi
 		// calling Start may result in an additional line of output being written in non-tty scenarios
 		_ = c.spinner.Start()
 	}
+
+	watchTerminalInterrupt(c)
```
weikanglim, line 412: I suspect this ends up leaking a goroutine each time the spinner starts. I'd like to propose a quick change here to avoid this behavior -- when I initially wrote some of this, it was intentionally coded to only have a single signal interception, we may need to change this behavior slighty to match

weikanglim (approved):

PR #4872, opened 2025-03-03
improve error message and check if powershell is installed

### LINKED ISSUE #4740: Fail fast for projects with hooks using pwsh
When a template is using hooks with pwsh, azd does not currently validate if pwsh is installed at the beginning of a command (like up or provision)
This results in a bad UX b/c azd could fail in the middle of a long process.

For this issue, implement a fail fast approach to check the hooks definitions and fail if pwsh is missing. Provide a clear message about how to install Powershell 7.

hemarina:
fix #4740 
If pwsh is not installed, it will return 
```
'pwsh' is not recognized as an internal or external command,
operable program or batch file.

ERROR: failed running pre hooks: 'preprovision' hook failed with exit code: '1', Path: '.\scripts\auth_init.ps1'. : exit code: 1
```
Improve the error message to be clear that pwsh is not installed:
image


cli/azd/pkg/tools/powershell/powershell.go
```diff
@@ -25,9 +26,21 @@ type powershellScript struct {
...
+}
+
 // Executes the specified powershell script
 // When interactive is true will attach to stdin, stdout & stderr
 func (bs *powershellScript) Execute(ctx context.Context, path string, options tools.ExecOptions) (exec.RunResult, error) {
+	if err := checkInstalled(options); err != nil {
+		return exec.RunResult{}, &internal.ErrorWithSuggestion{
+			Err: err,
+			Suggestion: "PowerShell is not installed or not in the path. To install PowerShell, " +
+				"visit https://learn.microsoft.com/powershell/scripting/install/installing-powershell",
```
JeffreyCA, line 40: Can we use `WithLinkFormat` to format the link?

https://github.com/Azure/azure-dev/blob/e63055e132c80705ba531c8190daf4188619c1d2/cli/azd/pkg/output/colors.go#L12-L15

cli/azd/pkg/tools/powershell/powershell.go
```diff
@@ -25,9 +26,21 @@ type powershellScript struct {
...
+	return tools.ToolInPath(options.UserPwsh)
+}
+
 // Executes the specified powershell script
 // When interactive is true will attach to stdin, stdout & stderr
 func (bs *powershellScript) Execute(ctx context.Context, path string, options tools.ExecOptions) (exec.RunResult, error) {
+	if err := checkInstalled(options); err != nil {
+		return exec.RunResult{}, &internal.ErrorWithSuggestion{
+			Err: err,
+			Suggestion: "PowerShell is not installed or not in the path. To install PowerShell, " +
```
JeffreyCA, line 39: Any thoughts if we should mention this is PowerShell **7** specifically?

wbreza: Agreed - we should add that it is Powershell 7.

wbreza (approved):

JeffreyCA (approved):

cli/azd/pkg/tools/powershell/powershell.go
```diff
@@ -25,9 +28,21 @@ type powershellScript struct {
 	envVars       []string
 }
 
+func checkInstalled(options tools.ExecOptions) (err error) {
+	return tools.ToolInPath(options.UserPwsh)
```
weikanglim, line 32: With the recent changes in #4595, I think you may have to do `string.Split(options.UserPwsh, " ")[0]` to handle `pwsh -NoProfile` for example.

cli/azd/pkg/tools/powershell/powershell.go
```diff
@@ -25,9 +28,21 @@ type powershellScript struct {
...
+}
+
 // Executes the specified powershell script
 // When interactive is true will attach to stdin, stdout & stderr
 func (bs *powershellScript) Execute(ctx context.Context, path string, options tools.ExecOptions) (exec.RunResult, error) {
+	if err := checkInstalled(options); err != nil {
+		return exec.RunResult{}, &internal.ErrorWithSuggestion{
+			Err: err,
+			Suggestion: fmt.Sprintf("PowerShell 7 is not installed or not in the path. To install PowerShell 7, visit %s",
```
weikanglim, line 41: Sometimes including the developer name may also be handy -- feel free to decide if this makes it better or wose
```suggestion
			Suggestion: fmt.Sprintf("PowerShell 7 (pwsh.exe) is not installed or not in the path. To install PowerShell 7, visit %s",
```

weikanglim (approved):
LGTM!

cli/azd/pkg/tools/powershell/powershell.go
weikanglim: In linux or mac, it's `pwsh` over `pwsh.exe`

PR #4874, opened 2025-03-04
set-secrets: Add shortcuts for using the Key Vault created by `azd add`

### LINKED ISSUE #4838: compose: `azd env set-secret` Key Vault integration
When a Key Vault already exists in a compose-enabled project, users should be able to bypass the current `azd env set-secret` flow and directly pick the compose-added Key Vault. Users should still be able to reference a different Key Vault if they wish.

JeffreyCA:
Closes #4838

This PR adds "shortcuts" to `azd env set-secrets <name>` so that if a Key Vault added by `azd add` exists, the user can directly pick that one to use instead of having to go through the typical **Subscription > Resource Group > Key Vault** flow.

If a key vault resource is defined in `azure.yaml`, but hasn't been provisioned yet, the following warning message is displayed, and the user is asked if they want to pick another KV or cancel the operation.

image

These changes do not touch the **Create a new Key Vault** flow, but in the future we could consider adding logic to detect if it's a compose project and ask if the user wants to add one to `azure.yaml` and provision it.

### Other changes

1. Update generated Bicep code for Key Vault to set the "set secret" permission on `principalId` so `azd env set-secret` can create new secrets in the project KV.
2. Add `secrets` to azure.yaml alpha schema and fix `examples` property so it auto-fills and shows up when you hover it in VS Code:
  image
3. The option to create a key vault or resource group in selection prompts are now numbered based on UX feedback:
  image

### Demo
https://github.com/user-attachments/assets/a047dee7-ee85-4cbb-8e4e-1d2776fec7dc

### DIFF
--- a/cli/azd/cmd/env.go
+++ b/cli/azd/cmd/env.go
@@ -12,9 +12,11 @@ import (
 	"strings"
 	"time"
 
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/arm"
 	"github.com/azure/azure-dev/cli/azd/cmd/actions"
 	"github.com/azure/azure-dev/cli/azd/internal"
 	"github.com/azure/azure-dev/cli/azd/pkg/account"
+	"github.com/azure/azure-dev/cli/azd/pkg/alpha"
 	"github.com/azure/azure-dev/cli/azd/pkg/azapi"
 	"github.com/azure/azure-dev/cli/azd/pkg/azureutil"
 	"github.com/azure/azure-dev/cli/azd/pkg/entraid"
@@ -32,6 +34,8 @@ import (
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
 )
+
+var featureCompose = alpha.MustFeatureKey("compose")
 
 func envActions(root *actions.ActionDescriptor) *actions.ActionDescriptor {
 	group := root.Add("env", &actions.ActionDescriptorOptions{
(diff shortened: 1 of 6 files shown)


SophCarp:
image

"Create a new Keyvault" should be 1 in the list. That way readers see it as a list option, rather than thinking it's a list title an
d skipping over it

SophCarp:
image
I'm not sure if this issue is my misunderstanding of set-secret, if I have some weird setup, or if there's a bug. @vhvb1989 for awareness

weikanglim:
@SophCarp Did it work after running `azd auth login` again?

SophCarp:
> @SophCarp Did it work after running `azd auth login` again?

Nope! It looks like it made the keyvaults in azure, but whether I tried to use an existing key vault or a new keyvault, I couldn't add a new secret to it. 

I added a secret in the portal, and I could connect the secret in azd just fine

SophCarp:
> Just curious, would it be indented or called out differently? I think it's rendered the way it is to separate out "create action" vs "noun (select one)". I guess "1. Create a new Key Vault" could be confused as both a noun and a verb. All this is done to optimize the workflow -- avoid an additional prompt (at the cost of UX currently).

Personally, I think the names of the key vaults are distinct enough from the action "Create a new keyvault" that I think it would be fine. Key vault names don't have spaces, whereas this is a sentence. It feels weird to me for items in a list to have different tabbing/numbering schemes. It makes the top item look more like a header than an item.

JeffreyCA:
> > @SophCarp Did it work after running `azd auth login` again?
> 
> Nope! It looks like it made the keyvaults in azure, but whether I tried to use an existing key vault or a new keyvault, I couldn't add a new secret to it.
> 
> I added a secret in the portal, and I could connect the secret in azd just fine

Sorry about that! This was a bug I inadvertently introduced - it should be fixed now

SophCarp:
image


I think this would be a bit more clear. I don't think it needs any coloring. 
```
Key vault detected in this project. Use this key vault?
- Yes
- No, use different key vault
```

SophCarp (approved):

JeffreyCA:
> > Just curious, would it be indented or called out differently? I think it's rendered the way it is to separate out "create action" vs "noun (select one)". I guess "1. Create a new Key Vault" could be confused as both a noun and a verb. All this is done to optimize the workflow -- avoid an additional prompt (at the cost of UX currently).
> 
> Personally, I think the names of the key vaults are distinct enough from the action "Create a new keyvault" that I think it would be fine. Key vault names don't have spaces, whereas this is a sentence. It feels weird to me for items in a list to have different tabbing/numbering schemes. It makes the top item look more like a header than an item.

Do you think we should also update the Resource Group picker?

image

SophCarp:
> Do you think we should also update the Resource Group picker?
> 
> image

How about `select/create a resource group to use:` ?

JeffreyCA:
> > Do you think we should also update the Resource Group picker?
> > image
> 
> How about `select/create a resource group to use:` ?

Oh I meant the numbering - I pushed a change so "Create a new resource group" is numbered as 1. 

image

Let me know what you think

SophCarp:
> Oh I meant the numbering - I pushed a change so "Create a new resource group" is numbered as 1.

Oooh, that's what you meant. Yes, that's great, thanks!

weikanglim (approved):

vhvb1989 (approved):

PR #4885, opened 2025-03-05
Pack build multi modules project

haoozhang:
This PR is to enhance the `pack build` to support multi-module projects. 

When `azd package` for a multi-module project before, it failed to package into image because of no context of parent project. Now we detect the parent project when `azd init` and use it as the context for `pack build` during `azd package`. This property is set as `docker.context`.

The code change is verified by the updated unit tests and following test projects:
+ [spring-petclinic-mysql](https://github.com/azure-javaee/azure-spring-boot-samples/tree/main/mysql/spring-cloud-azure-starter-jdbc-mysql/spring-petclinic) (single-module project)
+ [spring-petclinic-microservices](https://github.com/azure-javaee/spring-petclinic-microservices/tree/sjad?tab=readme-ov-file#starting-services-with-azd) (multi-module project)

## Example configuration

With the following filesystem setup:

```bash
├── azure.yaml
├── pom.xml # multimodule project
├── app
│   ├── pom.xml # referenced by ../pom.xml
```

The following azure.yaml configuration can be used to build `app` from source:

```yaml
services:
  app:
      project: app
      host: containerapp
      language: java
      docker:
          context: ..
```


haoozhang:
@saragluna @rujche could you please help to review this PR, thanks!

cli/azd/internal/appdetect/appdetect.go
```diff
@@ -142,6 +142,9 @@ type Project struct {
 	// Experimental: Database dependencies inferred through heuristics while scanning dependencies in the project.
 	DatabaseDeps []DatabaseDep
 
+	// The path to the parent project directory.
+	ParentPath string
+
```
saragluna, line 147: This `ParentPath` is more like a Java only property, maybe we should think about how to make it more general

cli/azd/pkg/project/framework_service_docker.go
```diff
@@ -458,6 +460,16 @@ func (p *dockerProject) packBuild(
...
 
+		// Specify parent context and build module for multi-module project when pack build
+		if svc.ParentPath != "" {
+			buildContext = svc.ParentPath
+			svcRelPath, err := filepath.Rel(buildContext, svc.Path())
+			if err != nil {
+				return nil, err
+			}
+			environ = append(environ, fmt.Sprintf("BP_MAVEN_BUILT_MODULE=%s", filepath.ToSlash(svcRelPath)))
+		}
```
saragluna, line 471: Should we enable this only when it's a Java & Maven project?

schemas/alpha/azure.yaml.json
```diff
@@ -107,6 +107,10 @@
                         "type": "string",
                         "title": "Path to the service source code directory"
                     },
+                    "parentPath": {
+                        "type": "string",
+                        "title": "Path to the parent directory of the service"
+                    },
```
saragluna, line 113: Why should this be defined in the `azure.yaml.json`? Is it a user facing option?

haoozhang: We need to detect this info during `azd init` and pass it into the `azd package` phrase, in which it performs the `pack build` and uses this info. So here I write this info into `azure.yaml`.

cli/azd/internal/appdetect/appdetect.go
haoozhang: We currently use it for Java multi-module projects, not sure if this property can apply other languages. maybe other language can reuse it?

cli/azd/internal/appdetect/java.go
```diff
@@ -17,8 +17,9 @@ import (
 )
 
 type javaDetector struct {
-	mvnCli       *maven.Cli
-	rootProjects []mavenProject
+	mvnCli         *maven.Cli
+	rootProjects   []mavenProject
+	moduleProjects map[string]mavenProject
```
rujche, line 22: Can we use a `rootPomFilePathMap map[string]string` to replace above 2 fields? In the `rootPomFilePath`, the key is module pom file path, and the value is root pom file path. (Note: it's root pom file path, not parent file path, which means it when fill the map, it should be responsible to find the root. )

Another proposal is use `parentPomFilePathMap`, when use the map, it should be responsible to find the root. 

I prefer the second way, because it can keep more information, it's possible to be used in the future. And we can provide a func `getRootPomFilePath(pomFilePath string, parentPomFilePathMap map[string]string) string` when retrieve root pom file path.

cli/azd/internal/appdetect/java.go
```diff
@@ -35,21 +36,24 @@ func (jd *javaDetector) DetectProject(ctx context.Context, path string, entries
...
 				return nil, nil
 			}
 
 			var currentRoot *mavenProject
 			for _, rootProject := range jd.rootProjects {
-				// we can say that the project is in the root project if the path is under the project
-				if inRoot := strings.HasPrefix(pomFile, rootProject.path); inRoot {
+				// We can say that the project is in the root project if
+				// 1) the project path is under the root project
+				// 2) the project is the module of root project
```
rujche, line 48: Is the first case valid?

cli/azd/internal/appdetect/java.go
```diff
@@ -58,6 +62,9 @@ func (jd *javaDetector) DetectProject(ctx context.Context, path string, entries
 			if err != nil {
 				return nil, fmt.Errorf("detecting dependencies: %w", err)
 			}
+			if currentRoot != nil {
+				result.ParentPath = currentRoot.path
```
rujche, line 66: In current implementation, `RootPath` is a better name than `ParentPath`.

cli/azd/internal/appdetect/java.go
rujche: No matter which way we choose, the `rootProjects` field can be deleted.

cli/azd/pkg/project/framework_service_docker.go
haoozhang: Make sense, I move it into Java block temporarily.

cli/azd/internal/appdetect/java.go
haoozhang: Maybe the naming `moduleProjects` is confusing. I have updated the naming. Actually, the current implementation is to store the mapping between module path and actual root project (not the direct parent path), i.e., the first thought you proposed.

Currently we don't foresee the use of direct parent path. Let's store the actual root directly for use, and we can refactor this method if necessary.

cli/azd/internal/appdetect/java.go
haoozhang: If a project is considered as a submodule of a root project, the project's path should be under the root project's path.
This is the initial version of parent detection before.

cli/azd/pkg/project/framework_service_docker.go
haoozhang: I think we should support two kinds of project: Maven based and Gradle based. The corresponding Envs are `BP_MAVEN_BUILT_MODULE` and `BP_GRADLE_BUILT_MODULE` respectively. We should consider how to differ from the build type in this code point later.

cli/azd/internal/appdetect/java.go
haoozhang: Updated.

cli/azd/internal/appdetect/java.go
rujche: I mean, we can delete `jd.rootProjects` and remove the check of `1) the project path is under the root project`, just keep the second one is enough: `2) the project is the module of root project`.

schemas/alpha/azure.yaml.json
weikanglim: Proposing to leverage `docker.context` and hide the concept for now: https://github.com/azure-javaee/azure-dev/pull/108

The root project property isn't cross-language specific, but does make sense in the context of docker/pack

cli/azd/internal/appdetect/java.go
weikanglim: I think @rujche is right here that we can consider picking one or the other. I chose one approach in https://github.com/azure-javaee/azure-dev/pull/108

weikanglim:
@haoozhang I made some changes in https://github.com/azure-javaee/azure-dev/pull/108 to avoid introducing a new property in `azure.yaml`, and kept the implementation in java.go a little more lean. Feel free to merge this if it looks good to you.

weikanglim (approved):

PR #4886, opened 2025-03-05
Enhance add databaseDepMap by maven dependency

rujche:
1. Enhance add databaseDepMap by maven dependency. 
2. Fix some lacking codes about `appdetect.DbMySql`.
3. `DbCosmos` is not added now, my current plan is add it in next PR, after current PR merged.

### DIFF
--- a/cli/azd/internal/repository/detect_confirm.go
+++ b/cli/azd/internal/repository/detect_confirm.go
@@ -15,8 +15,6 @@ import (
 	"strings"
 	"time"
 
-	"go.opentelemetry.io/otel/attribute"
-
 	"github.com/AlecAivazis/survey/v2/terminal"
 	"github.com/azure/azure-dev/cli/azd/internal/appdetect"
 	"github.com/azure/azure-dev/cli/azd/internal/cmd/add"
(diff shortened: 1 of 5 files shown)


cli/azd/internal/repository/detect_confirm.go
```diff
@@ -15,6 +15,8 @@ import (
 	"strings"
 	"time"
 
+	"go.opentelemetry.io/otel/attribute"
+
```
saragluna, line 19: Is this due to some auto-formatting rule?

cli/azd/internal/repository/infra_confirm.go
```diff
@@ -47,7 +47,7 @@ func (i *Initializer) infraSpecFromDetect(
 					DatabaseName: dbName,
 				}
 				break dbPrompt
-			case appdetect.DbPostgres:
+			case appdetect.DbPostgres, appdetect.DbMySql:
```
weikanglim, line 50: Hey @rujche ! Thanks for making this change!

`infra_confirm.go` is the legacy codepath when `alpha.compose` is not enabled. Will SJAD team want to support customers in this path? For our own maintenance reasons, I think we can avoid backporting any feature additions here.

weikanglim: If you are thinking about making this change, I think you'll want a different switch case to set `spec.DbMySql`

weikanglim (commented):
@rujche This looks good to me. I recommend removing the `infra_confirm.go` change as it's the legacy codepath with compose not enabled. Otherwise, I'm ready to merge this on the next iteration.

rujche:
@weikanglim 

Thanks for the quick response!

> I recommend removing the infra_confirm.go change as it's the legacy codepath with compose not enabled.

Done. Please help to review again.

weikanglim (approved):
LGTM!

PR #4900, opened 2025-03-05
Extension Framework: Adds support for multi-select prompt

wbreza:
Adds support for multi-select prompt to the Extension Framework.

> [!NOTE]
> Also includes regenerated proto/grpc files

## UX Example
image

## Code Usage

```go
// Create a new AZD client
azdClient, err := azdext.NewAzdClient()
if err != nil {
	return fmt.Errorf("failed to create azd client: %w", err)
}

defer azdClient.Close()

selectedValues, err = azdClient.Prompt().MultiSelect(ctx, &azdext.MultiSelectRequest{
	Options: &azdext.MultiSelectOptions{
		Message: "Which Azure services do you use most with AZD?",
		Choices: []*azdext.MultiSelectChoice{
			{Label: "Container Apps", Value: "container-apps"},
			{Label: "Functions", Value: "functions"},
			{Label: "Static Web Apps", Value: "static-web-apps"},
			{Label: "App Service", Value: "app-service"},
			{Label: "Cosmos DB", Value: "cosmos-db"},
			{Label: "SQL Database", Value: "sql-db"},
			{Label: "Storage", Value: "storage"},
			{Label: "Key Vault", Value: "key-vault"},
			{Label: "Kubernetes Service", Value: "kubernetes-service"},
		},
	},
})
if err != nil {
	return nil
}
```


cli/azd/extensions/microsoft.azd.demo/internal/cmd/prompt.go
```diff
@@ -31,6 +31,26 @@ func newPromptCommand() *cobra.Command {
...
+						{Label: "Cosmos DB", Value: "cosmos-db"},
+						{Label: "SQL Database", Value: "sql-db"},
+						{Label: "Storage", Value: "storage"},
+						{Label: "Key Vault", Value: "key-vault"},
+						{Label: "Kubernetes Service", Value: "kubernetes-service"},
+					},
+				},
+			})
+			if err != nil {
+				return nil
```
JeffreyCA, line 51: How come we're returning `nil` here instead of `err`, like we do below?

JeffreyCA (approved):

cli/azd/extensions/microsoft.azd.demo/internal/cmd/prompt.go
wbreza: Good catch! Will update.

PR #2005, opened 2023-04-21
Refactor `ci-build.ps1` to allow configurable flags

weikanglim:
In upcoming work for #1996, we need to make the flags in `ci-build.ps1` dynamically configurable. Thus, this precursor work refactors existing flags to support dynamic options.

This is a strict refactor. The actual build flags used are unchanged.

The observed build flags after this change (copied from the pipeline run) can be diffed against existing, which shows the exact build flags being preserved.

Windows:
```
go build `
  -gcflags="-trimpath" `
  -asmflags="-trimpath" `
  -buildmode="exe" `
  -trimpath `
  -tags="cfi,cfg,osusergo" `
  -ldflags="-s -w -X 'github.com/azure/azure-dev/cli/azd/internal.Version=0.9.0-beta.2-pr.2782337 (commit 7b64d5b55fa5c6c7b31c8f83fdd9f1fe5cb5288a)' -linkmode=auto -extldflags=-Wl,--high-entropy-va"
```

Linux:
```
go build `
  -gcflags="-trimpath" `
  -asmflags="-trimpath" `
  -buildmode="pie" `
  -tags="cfi,cfg,osusergo,cfgo" `
  -ldflags="-s -w -X 'github.com/azure/azure-dev/cli/azd/internal.Version=0.9.0-beta.2-pr.2782337 (commit 7b64d5b55fa5c6c7b31c8f83fdd9f1fe5cb5288a)' -extldflags=-Wl,--high-entropy-va"
```

MacOs:
```
go build `
  -gcflags="-trimpath" `
  -asmflags="-trimpath" `
  -buildmode="pie" `
  -tags="cfi,cfg,osusergo,cfgo" `
  -ldflags="-s -w -X 'github.com/azure/azure-dev/cli/azd/internal.Version=0.9.0-beta.2-pr.2782337 (commit 7b64d5b55fa5c6c7b31c8f83fdd9f1fe5cb5288a)' -linkmode=auto"
```

PR #4906, opened 2025-03-06
ci: set `buildmode=pie`

weikanglim:
Set `buildmode=pie` for Windows. Clean up ineffectual build flags.

Before this change, the flag `--high-entropy-va` was set explicitly. However, this turned out to be ineffectual since the build mode was set explicitly to `exe`. [ASLR](https://en.wikipedia.org/wiki/Address_space_layout_randomization) security techniques requires [PIE](https://en.wikipedia.org/wiki/Position-independent_code#PIE).

Enabling `buildmode=pie` automatically sets the additional flags on Windows to fully enable ASLR. With these changes, the PE header using DUMPBIN now reads:
```
 DLL characteristics
     High Entropy Virtual Addresses
     Dynamic base
     NX compatible
     Terminal Server Aware
```

### DIFF
--- a/cli/azd/ci-build.ps1
+++ b/cli/azd/ci-build.ps1
@@ -184,9 +184,9 @@ $oldGOEXPERIMENT = $env:GOEXPERIMENT
 $env:GOEXPERIMENT="loopvar"
 
 try {
-    Write-Host "Running: go build ``"
-    PrintFlags -flags $buildFlags
     if ($OneAuth) {
+        Write-Host "Running: go build (oneauth)``"
+        PrintFlags -flags $buildFlags
         # write the go build command line to a script because that's simpler than trying
         # to escape the build flags, which contain commas and both kinds of quotes
         Set-Content -Path build.sh -Value "go build $($buildFlags)"
@@ -217,6 +217,8 @@ try {
         go build @buildFlags
     } 
     else {
+        Write-Host "Running: go build ``"
+        PrintFlags -flags $buildFlags
         go build @buildFlags
     }
     
--- a/cli/azd/ci-build.ps1
+++ b/cli/azd/ci-build.ps1
@@ -91,10 +91,18 @@ if ($IsWindows) {
 
 # Run `go help build` to obtain detailed information about `go build` flags.
 $buildFlags = @(
-    # Remove file system paths from the compiled binary
-    "-gcflags=-trimpath",
-    # Remove file system paths from the assembled code
-    "-asmflags=-trimpath"
+    # remove all file system paths from the resulting executable.
+    # Instead of absolute file system paths, the recorded file names
+    # will begin either a module path@version (when using modules),
+    # or a plain import path (when using the standard library, or GOPATH).
+    "-trimpath",
+    
+    # Use buildmode=pie (Position Independent Executable) for enhanced security across platforms
+    # against memory corruption exploits across all major platforms.
+    #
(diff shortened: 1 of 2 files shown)


cli/azd/ci-build.ps1
```diff
@@ -120,37 +128,20 @@ if ($IsWindows) {
...
-        $tagsFlag,
-        # -extldflags=-Wl,--high-entropy-va: Pass the high-entropy VA flag to the linker to enable high entropy virtual addresses
-        ($ldFlag + "-linkmode=auto -extldflags=-Wl,--high-entropy-va")
-    )
 }
 elseif ($IsLinux) {
     Write-Host "Building for linux"
-    $buildFlags += @(
-        "-buildmode=pie",
-        ($tagsFlag + ",cfgo"),
```
danieljurek, line 139: Did we eliminate the need for `cfgo`?

cli/azd/ci-build.ps1
```diff
@@ -120,37 +128,20 @@ if ($IsWindows) {
...
-        ($ldFlag + "-extldflags=-Wl,--high-entropy-va")
-    )
 }
 elseif ($IsMacOS) {
     Write-Host "Building for macOS"
-    $buildFlags += @(
-        "-buildmode=pie",
-        ($tagsFlag + ",cfgo"),
-        # -linkmode=auto: Link Go object files and C object files together
-        ($ldFlag + "-linkmode=auto")
```
danieljurek, line 150: We have been using cgo in the past... I don't think this is a problem because I believe (can't find much documentation evidence) that `-linkmode=auto` is default so you probably won't have problems.

danieljurek (approved):

cli/azd/ci-build.ps1
danieljurek: Some documentation: https://github.com/golang/go/blob/master/src/cmd/cgo/doc.go#L1085-L1089

> By default, cmd/link will decide the linking mode as follows: if the only
packages using cgo are those on a list of known standard library
packages (net, os/user, runtime/cgo), cmd/link will use internal linking
mode. Otherwise, there are non-standard cgo packages involved, and cmd/link
will use external linking mode... 

Probably a good enough heuristic.

danieljurek:
@weikanglim -- Please also delete this entry in the `.gdnbaselines` file as this PR fixes the associated issues: https://github.com/Azure/azure-dev/blob/main/.config/guardian/.gdnbaselines#L28-L51

vhvb1989 (approved):

cli/azd/ci-build.ps1
weikanglim: Yeah, we never got around to removing it: [context](https://github.com/Azure/azure-dev/pull/2005#discussion_r1196929309)

cli/azd/ci-build.ps1
weikanglim: I spelunked into linking and it seems that the default value is [auto](https://github.com/golang/go/blob/bc5f4a555e933e6861d12edba4c2d87ef6caf8e6/src/cmd/link/internal/ld/config.go#L85)

weikanglim:
builds are successful:
https://dev.azure.com/azure-sdk/internal/_build/results?buildId=4636124&view=results
https://dev.azure.com/azure-sdk/internal/_build/results?buildId=4636126&view=results

i cancelled my manually triggered build to test BinSkim issue, it looks like check enforcer is picking up this run instead of the one automatically triggered. will override as a result

Issue #4747, opened 2025-01-31
compose: explicit mapping


weikanglim:
**Prelude**
1. This issue is under proposal, and has not yet been committed into plans.
2. This issue may require additional design around the "object model" of resources since it introduces concepts like `${event-hubs-prod.name}`. The design around the object model would be a separate scoped item. This object model may also be required as AVM modules also expose intermediate types that are not the underlying ARM types.

# Explicit mapping

## Overview

Explicit mapping is an intermediate/advanced feature that extends the current `uses` functionality. In contrast, to `uses`, which automatically injects environment variables into host resources based on standard naming conventions, explicit mapping does this explicitly.

Explicit mapping allows explicit association of resource properties (a higher-level object model) with environment variables, which is helpful in multi-dependency scenarios. The `uses` relationship is automatically derived from this mapping.

### Example Mapping

```yaml
resources:
  event-hubs-prod:
    type: messaging.eventhubs
  my-app:
    type: host.containerapp
    env:
      PRIMARY_EVENT_HUBS_NAME: ${event-hubs-prod.name}
```

This mapping explicitly assigns `PRIMARY_EVENT_HUBS_NAME` to the `name` of `event-hubs-prod`.

## Scenario: Handling Multiple Dependencies

When a resource has multiple dependencies of the same type, automatic environment variable injection can cause ambiguity.

### Problem
```yaml
resources:
  event-hubs-prod:
    type: messaging.eventhubs
  event-hubs-dev:
    type: messaging.eventhubs
  my-app:
    type: compute.containerapp
    uses:
      - event-hubs-prod
      - event-hubs-dev
    # This would auto-inject environment variables but become ambiguous
    # Which event hub does AZURE_EVENT_HUBS_NAME refer to?
```

### Proposed Solution
To remove ambiguity, explicit environment variable mapping is used:

```yaml
resources:
  event-hubs-prod:
    type: messaging.eventhubs
  event-hubs-dev:
    type: messaging.eventhubs
  my-app:
    type: compute.containerapp
    env:
      PRIMARY_EVENT_HUBS_NAME: ${event-hubs-prod.name}
      SECONDARY_EVENT_HUBS_NAME: ${event-hubs-dev.name}
```

## Proposed Implementation

The `add` command will generate default environment variable mappings when adding a second dependency of the same type. Users can override these mappings explicitly to provide meaningful names.

JeffreyCA:
[...]

Issue #4577, opened 2024-11-21
compose: existing resource support


weikanglim:
## Background

This issue tracks the ability to reference existing resources in `azure.yaml`.

For example, the schema could look something like: 
```yaml
resources:
  existing-db:
    existing: true
    resourceId: <resource ID> # Virtual -- stored in `.env` for cross-env purposes
```

This enhancement builds on top of the idea that `azure.yaml` is representing the app-centric *view* of the infrastructure resources. Similar to a SQL view, resources can either be created (materialized in SQL terms) or existing (queried from in SQL terms).

## Scenarios

When this type of reference metadata is present, scenarios we would want to enable for overall consistency of the system:

1. `use` binding relationships works with existing resources. We could imagine that environment variables are established in the same way a created resource does. This would require updating the infra synthesis to account for existing declarations. 
2. Command gestures like `show` works with existing resources. This change would be a minimal addition of the currently established design, that each resource maps to an `AZURE_RESOURCE_xx_ID` upon provisioning. See #4534 for details.

weikanglim:
100% agree that multiple environments scenarios should always be supported.

There are many ways we can do that, currently:

### Current established paradigms:

1. The current, already supported approach. Define in azure.yaml, reference the environment variable:
```yaml
resources:
  existing-db:
    id: ${ACTUAL_ID_KEY}
```
2. Define environment variable directly: `${AZURE_RESOURCE_xx_ID}` -- This could serve as an override.

### Future Improvements

Some alternative improvements over the long run:

1. More explicit parameterization with `${env.someId}`
1. Some other wild ideas: Explicit configuration files that are merged together. Perhaps you could think of:
    - `azure.yaml`
    - `azure.prod.yaml` - With prod specific configuration

I personally see that 1 and 2 become increasing cumbersome over the long run, and perhaps at some breaking point, separate configuration files are more maintainable.

I do very much see the need to have a system of this essence very clearly in-place and well-documented as a pattern.

puicchan:
For host, my vote is to create an issue to track and keep in backlog.
For DB, keep to asking for secret in KV. We can assume dev has full access to KV to start.

PR #4914, opened 2025-03-08
compose: show resources

### LINKED ISSUE #4727: compose: discover and use env variables without provisioning host service

weikanglim:
Add ability to `azd show <name>` for all existing supported resources.

With this change, we add the ability to display fully-evaluated connection variables for resources without provisioning a `host` resource.

To do this, we create a generic mapping between ARM and the app representation. This mapping, which is comprised of simple templating-like expressions, are evaluated against live Azure in the `show` command. 

Example output:
image

## Forward-looking ideas

1. This contributes to #4577, in which `azd show <full resource ID>` could be used to display, or as a base layer, for obtaining connection information to an existing resource. The schema does not track managed identity vs key-based approaches, but I do see a somewhat straightforward pathway there.

2. This also contributes to #4747, as we are laying down a schema mapping that could be used to *emit* Bicep representation during generation. This may end up unifying the story between generation and evaluation for roundtripping.

3. This also contributes to #4747. we expect to use parts of the parsing logic in to support a similar expression-model at the app layer.

Closes #4727
Contributes to #4577, #4747, #4887

### DIFF
--- a/.vscode/cspell-github-user-aliases.txt
+++ b/.vscode/cspell-github-user-aliases.txt
@@ -33,6 +33,7 @@ sebastianmattar
 sergi
 sethvargo
 stretchr
+tidwall
 theckman
 TheEskhaton
 tonybaloney
--- a/cli/azd/.vscode/cspell-azd-dictionary.txt
+++ b/cli/azd/.vscode/cspell-azd-dictionary.txt
@@ -31,6 +31,7 @@ aspnet
 aspnetcore
 asyncmy
 asyncpg
+avmres
 azapi
 azblob
 AZCLI
@@ -57,6 +58,7 @@ azuretools
 azureutil
 azureyaml
 Backticks
+bicept
 bicepparam
 blockblob
 BOOLSLICE
@@ -111,6 +113,7 @@ Frontends
 funcapp
 functestapp
 functionapp
+gjson
 go-imath
 GOARCH
 GOCOVERDIR
--- a/cli/azd/internal/cmd/add/add_preview.go
+++ b/cli/azd/internal/cmd/add/add_preview.go
@@ -9,9 +9,12 @@ import (
 	"fmt"
 	"io"
 	"log"
+	"maps"
+	"slices"
 	"strings"
 	"text/tabwriter"
 
+	"github.com/azure/azure-dev/cli/azd/internal/scaffold"
 	"github.com/azure/azure-dev/cli/azd/pkg/infra/provisioning"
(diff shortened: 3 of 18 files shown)


SophCarp:
UX Review: 
azd show -h: 
image
- Shouldn't it be `azd show <resource> [flags]` ?  I think aditionally some examples could be useful: 

```
Display information about your app and its resources. no resource is selected, shows summary of project. select resources by it's name in azure.yaml

example: 
  azd show api //shows a resource named api
  azd show vault -e test //shows a resource named vault in the test evironment

Usage
  azd show <resource> [flags]
```
`azd show`
image
- I would change the order so that environments is first, then Services, then "View in Azure". 
- Is there a reason why only services are shared in the overview and not all resources? Maybe having a flag `--all` or something to list both services and resources in one list.

`azd show api`
image
- The big chunk of blue doesn't lead to clear readability. Could just the env variable names be blue, and everything after the `=` be primary (or secondary) text color?
- I'd also only have the name be the magenta color. The type can by in secondary text color.

SophCarp:
Thinking through the experience: 

`azd show` provides: 
- list of environments (with current environment marked)
- List of services and their endpoint links
- Link to the resource groups azure portal. 

`azd show <resource>` provides: 
- Resource name
- Resource type
- Endpoint (if it's a service)
- List of environment variables

These two options very different results despite being the same command. It also feels like it's stretching the definition of "show"

usually, `show` is the verb to share one concrete detail or item. It's often paired with `list` which provides a list of items in the same type. `show` usually works best when given a clear definition of what it shows in the subgroups: 
- `az containerapp revision show <revision>` shows the detail of one revision
-  `az containerapp env show <env> shows the detail of one container app environment
- `az containerapp show <containerapp> shows the details of the container app

These commands come with a matching list command that lists the available items that can be specified in show:
- `az containerapp revision list`
-  `az containerapp env list`
-  `az containerapp list`

That way you can run 'az containerapp list' to see all the options then run `az containerapp show <containerapp>` for the specific item. You can't run `az containerapp show` without specifying what individual container app you want to see. 

`azd show` doesn't act in the same way. Maybe this is because azd ends up being more "verb-first" than "Noun-first" when it comes to commands? https://clig.dev/?trk=public_post-text#subcommands Or maybe this command is larger than `show`. Not sure what next step to take on this, but wanted to share the thoughts.

weikanglim:
@SophCarp I will doing be a one-pass working with you on addressing  the items in #4933 shortly

Issue #4879, opened 2025-03-04
[AI Compose] Migrate connections to use EntraID instead of keys


vhvb1989:
Consider moving dependencies from a hub, like storage and key vault to use EntraID instead of Keys.
Also check how and if we can use EntraID to create the connections for the Hub to AI Services

PR #4931, opened 2025-03-12
compose: `azd add` support for Azure AI Search

### LINKED ISSUE #4889: compose: `azd add` support for Azure AI Search
Adds support for adding AI Search service via composability.

- Adds support for top level resource type
- Identify if we can easily support data plane resources (this may be a better fit for the AI extension)
  - Indexes
  - Index aliases
  - Indexers (requires an index + data source, skillset is optional)
  - Data sources
  - Skillsets
  - Debug sessions

JeffreyCA:
Closes #4889 

This PR adds `azd add` support for Azure AI Search (aka Azure Cognitive Search). **Azure AI Search** now appears as an option under the **AI** group (previously **AI models**) in the `azd add` menu:

image

Support for dataplane resources (indexes, index aliases, indexers, data sources, skillsets, debug sessions) are out of scope of this PR.

### Auth
The AI Search resource is configured with a "hybrid" authentication configuration, allowing access via API keys and Entra/RBAC. The [**Search Service Contributor** and  **Search Index Data Contributor** roles](https://learn.microsoft.com/en-us/azure/search/search-security-rbac#built-in-roles-used-in-search) are assigned to the user and host identity (if host **uses** `search`). 

image

If both an Azure AI Search resource and a Foundry model are present in `azure.yaml`, the generated Bicep azd additionally creates an API key-based connection between the AI hub and AI Search as a "Connected resource". This connection currently uses API key authentication due to limited support for key-less/Entra auth. Issue #4879 tracks moving to Entra-based authentication.

image

### Connecting to resource

Users can connect to the resource directly using the `AZURE_AI_SEARCH_ENDPOINT` environment variable with an AI Search client library.

If Foundry resources are present, they can also connect to the resource through the Foundry SDK using the AI project connection string: [example](https://learn.microsoft.com/en-us/azure/ai-foundry/how-to/develop/sdk-overview?tabs=sync&pivots=programming-language-python#azure-ai-search).

## azure.yaml schema

```yaml
resources:
  search:
    type: ai.search
```

## Binding variables

`AZURE_AI_SEARCH_ENDPOINT` - Azure AI Search endpoint

## Testing
Validated using sample app https://github.com/JeffreyCA/azd-compose-search-demo

### DIFF
--- a/cli/azd/resources/scaffold/templates/main.bicept
+++ b/cli/azd/resources/scaffold/templates/main.bicept
@@ -132,6 +132,6 @@ output AZURE_AIPROJECT_CONNECTION_STRING string = aiModelsDeploy.outputs.aiFound
 output AZURE_RESOURCE_AI_PROJECT_ID string = aiModelsDeploy.outputs.projectId
 {{- end}}
 {{- if .AISearch }}
-output AZURE_RESOURCE_SEARCH_ID string = resources.outputs.aiSearchName
+output AZURE_RESOURCE_SEARCH_ID string = resources.outputs.AZURE_RESOURCE_SEARCH_ID
 {{- end}}
 {{ end}}
--- a/cli/azd/resources/scaffold/templates/main.bicept
+++ b/cli/azd/resources/scaffold/templates/main.bicept
@@ -132,6 +132,7 @@ output AZURE_AIPROJECT_CONNECTION_STRING string = aiModelsDeploy.outputs.aiFound
 output AZURE_RESOURCE_AI_PROJECT_ID string = aiModelsDeploy.outputs.projectId
 {{- end}}
 {{- if .AISearch }}
+output AZURE_AI_SEARCH_ENDPOINT string = resources.outputs.AZURE_AI_SEARCH_ENDPOINT
 output AZURE_RESOURCE_SEARCH_ID string = resources.outputs.AZURE_RESOURCE_SEARCH_ID
 {{- end}}
 {{ end}}
--- a/cli/azd/resources/scaffold/templates/main.bicept
+++ b/cli/azd/resources/scaffold/templates/main.bicept
@@ -76,6 +76,17 @@ module aiModelsDeploy 'ai-project.bicep' = {
     envName: environmentName
   }
 }
+
+{{- if .AISearch }}
+module aiSearchConnection 'ai-search-conn.bicep' = {
+  scope: rg
+  name: 'ai-search-connection'
+  params: {
+    aiHubName: aiModelsDeploy.outputs.hubName
+    aiSearchName: resources.outputs.aiSearchName
+  }
+}
+{{- end}}
 {{- end}}
 
 {{- if .Services}}
@@ -119,5 +130,8 @@ output AZURE_RESOURCE_SERVICE_BUS_ID string = resources.outputs.AZURE_RESOURCE_S
 {{- if .AiFoundryProject }}
 output AZURE_AIPROJECT_CONNECTION_STRING string = aiModelsDeploy.outputs.aiFoundryProjectConnectionString
 output AZURE_RESOURCE_AI_PROJECT_ID string = aiModelsDeploy.outputs.projectId
+{{- end}}
+{{- if .AISearch }}
+output AZURE_RESOURCE_SEARCH_ID string = resources.outputs.aiSearchName
 {{- end}}
 {{ end}}
--- a/cli/azd/resources/scaffold/templates/main.bicept
+++ b/cli/azd/resources/scaffold/templates/main.bicept
@@ -78,7 +78,7 @@ module aiModelsDeploy 'ai-project.bicep' = {
 }
 
 {{- if .AISearch }}
-module aiSearchConnection 'ai-search-conn.bicep' = {
+module aiSearchConnection 'modules/ai-search-conn.bicep' = {
   scope: rg
   name: 'ai-search-connection'
   params: {
--- a/cli/azd/resources/scaffold/templates/resources.bicept
+++ b/cli/azd/resources/scaffold/templates/resources.bicept
@@ -299,11 +299,12 @@ resource localUserOpenAIIdentity 'Microsoft.Authorization/roleAssignments@2022-0
 {{- end}}
 
 {{- if .AISearch}}
-module searchService 'br/public:avm/res/search/search-service:0.9.1' = {
+module search 'br/public:avm/res/search/search-service:0.9.1' = {
   name: 'aiSearch'
   params: {
     name: '${abbrs.storageStorageAccounts}${resourceToken}'
     location: location
+    tags: tags
     sku: 'basic'
     replicaCount: 1
     managedIdentities: {
@@ -333,7 +334,12 @@ module searchService 'br/public:avm/res/search/search-service:0.9.1' = {
       }
       {{- end}}
     ]
-    disableLocalAuth: true
+    disableLocalAuth: false
+    authOptions: {
+      aadOrApiKey: {
+        aadAuthFailureMode: 'http401WithBearerChallenge'
+      }
+    }
     publicNetworkAccess: 'Enabled'
   }
 }
@@ -668,6 +674,12 @@ module {{bicepName .Name}} 'br/public:avm/res/app/container-app:0.8.0' = {
             value: account.outputs.endpoint
           }
           {{- end}}
+          {{- if .AISearch}}
+          {
+            name: 'AZURE_SEARCH_ENDPOINT'
+            value: 'https://${search.outputs.name}.search.windows.net'
+          }
+          {{- end}}
           {{- if .HasAiFoundryProject }}
           {
             name: 'AZURE_AIPROJECT_CONNECTION_STRING'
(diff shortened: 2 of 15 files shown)


wbreza (approved):
Looks great!

cli/azd/resources/scaffold/templates/main.bicept
```diff
@@ -120,4 +131,7 @@ output AZURE_RESOURCE_SERVICE_BUS_ID string = resources.outputs.AZURE_RESOURCE_S
 output AZURE_AIPROJECT_CONNECTION_STRING string = aiModelsDeploy.outputs.aiFoundryProjectConnectionString
 output AZURE_RESOURCE_AI_PROJECT_ID string = aiModelsDeploy.outputs.projectId
 {{- end}}
+{{- if .AISearch }}
+output AZURE_RESOURCE_SEARCH_ID string = resources.outputs.AZURE_RESOURCE_SEARCH_ID
```
vhvb1989, line 135: We need to output AZURE_AI_SEARCH_ENDPOINT here as well

JeffreyCA: Just trying to understand, how is that used?

rajeshkamal5050: May be like this?? - https://github.com/Azure-Samples/get-started-with-ai-agents/blob/main/src/gunicorn.conf.py#L51

vhvb1989: We need to set it for the ACA environment (you are already doing that) for when the app runs on Azure.
And we also need to output it, so it gets added to the azd .env. For when you run the app locally.

All ENV VARS you automatically set for the ACA environment must be also set locally to the .env file to ensure you can launch the application locally (without manually exporting the ENV VARS to the system env)

JeffreyCA: I see, so for local dev. Thanks 👍

weikanglim: With the current direction of composability, we are only outputting resourceId, with the idea being that we can recover all the environment variables as an implementation detail for `azd show` (current) or `azd run` (future). Are we not liking this idea here? Today, `azd show api` for example will display all environment variables set in the container app. `azd run api` could simply use this as a straightforward path to set up a local environment.

I don't have strong qualms if the idea of "outputting all application environment variables" across all services as environment state is something we want to do. I will admit that I'm not sure how this works yet when there are two containerapp frontend and backends, or when there is only a resource created but no container app. In these cases, what would we choose to output? If there's a strong proposal on how this all should work, I'm happy to accept the directional shift.

vhvb1989: You can make `azd show` and `azd run` the way to go (expected way of doing things), for sure.
But what happen to other scenarios, like
- Running post-provision hooks which communicate to the service, like uploading files and creating an index for the AI Search service. Will folks need to run `azd show` inside the script and parse the output ?  Or will you make azd to run `azd show` before calling a hook to figure what env vars to inject?  (for every hook?)

There are many python-templates currently calling `azd env get-values` to output/generate an .env file to run the application locally. They can do this because they have control of what to output from IaC.  Are we not only removing the control of that to output from IaC (with compossibility) but also asking folks to change the way they are currently working (running their apps locally) to something different, if they want to use composability?

vhvb1989: > I'm not sure how this works yet when there are two containerapp frontend and backends

Whenever you support more than one only instance of a resource type, you have to use the name of the resource as part of the output:
   AZURE_SERVICE_<resource-name>_ENDPOINT

JeffreyCA: With the logic we have now for generic `azd show`, would it be any better if we somehow used that to set the environment variables for each resource in `.env`, instead of setting output params in the generated Bicep like `AZURE_AIPROJECT_CONNECTION_STRING` and `AZURE_AI_SEARCH_ENDPOINT`?

It looks like we're only outputting endpoints/connection strings in Bicep for container registry, key vault, and AI project connection string today.

> I don't have strong qualms if the idea of "outputting all application environment variables" across all services as environment state is something we want to do.

I recall some of the initial ideas we had for `azd run` was to be able to run any arbitrary command with all of the variables from `.env` loaded (e.g. `azd run python app.py`) and this could tie together nicely. In that world, I'm not sure how whether a container host exists or there's a 'uses' relationship between resources would affect things. It's almost as if the local python instance in this example *becomes* the host

vhvb1989: We only set the outputs for main.bicep to support the `infra synth` scenario, where everything still works the same after you infra synth to own the infrastructure.

We do the same for Aspire, btw.  
For supporting bindMounts, we output the Storage Endpoint for where the file-share was created. An operation after provision takes care of uploading files to that file-share before deploying the ACA.

At some point, for composability, we might want to also move the ACA creation to `azd deploy`, instead of creating the ACA with a placeholder img during provision and deploying the service img during deploy.  If we ever go to that direction, the .env is what we use to feed `deploy` for the creating of the ACA.

cli/azd/resources/scaffold/templates/resources.bicept
```diff
@@ -298,6 +298,53 @@ resource localUserOpenAIIdentity 'Microsoft.Authorization/roleAssignments@2022-0
 }
 {{- end}}
 
+{{- if .AISearch}}
+module search 'br/public:avm/res/search/search-service:0.9.1' = {
+  name: 'ai-search'
+  params: {
+    name: '${abbrs.storageStorageAccounts}${resourceToken}'
```
weikanglim, line 305: Do we need a new abbreviation for search here? I usually look up abbreviations here: https://learn.microsoft.com/en-us/azure/cloud-adoption-framework/ready/azure-best-practices/resource-abbreviations

I had a tool available at one point available in a private [branch](https://github.com/Azure/azure-dev-pr/pull/1720/files#diff-a0d34c43dfde1ff99f6c4760e6ebd53380f0f184df59a2e1906fcd8df6ab9e4c) but haven't had time to find a way to get this integrated into mainline.

cli/azd/resources/scaffold/templates/main.bicept
weikanglim: Thanks all for continuing the great discussion here.

@vhvb1989 I may have misunderstood the nature of your original suggestion. From what you're describing, you're recommending that common infrastructure dataplane endpoints to be outputted, as opposed to "All ENV VARS you automatically set for the ACA environment to ensure you can launch the application locally".

For example, would you recommend outputting `REDIS_URL` and `REDIS_PASSWORD` as akvs references here so that local dev is possible?

I don't know if you feel it's valuable, but the primitive we currently have is:

Given `AZURE_RESOURCE_<xxx>_ID` stored in `.env`, azd has the ability to expand this into environment variables suitable for any running program.

I could imagine that in `postprovision` hooks, the hook could say "run as the entire Azure environment", just "run connected to service bus", or "as an api", and `azd` would compute the dataplane accessible environment similar to how `azd run` would work. Personally speaking, in the long tail future, having postprovision hooks simply be azd commands/extensions that automatically do this (they target a resource construct, azd provides environment variables) would be a great story to tell.

vhvb1989: It's hard to set a general approach, I think.
The Redis example is a good way to see how/why we wouldn't like to just set, either the values or akvs to the azd .env
For things like Redis, storage or MongoDB, it is not too crazy to think/expect to launch local emulators.
But for AI search or AI Project, I think we must connect folks to cloud-resouces directly.

What I think we can 100% follow as a general guideline is to avoid mandating a unique way for users to consume resources thru AZD. This means allowing AZD to adapt to different scenarios and never forcing folks into some AZD way. If we follow such guideline, we can output the resourceID for the experience you want to offer by using AZD directly (with azd show + azd run), but we would be also including outputs for other users who would currently be expecting such endpoints to be available.

This is similar to how we are offering AI models thru OpenAI service endpoint or thru AI Project

rajeshkamal5050: I think for folks to incrementally adopt into azd features we should work well with existing flows and tools while we continue to build them.
- For local, definitely azd show and azd run is the way to go, making local experiences better.
- For bicep output params, adding application runtime properties along with resource ids into .env will help integration through hooks or even other external tools. Also this will keep the synth'ed bicep usable as-is.

cli/azd/resources/scaffold/templates/resources.bicept
JeffreyCA: Good catch! The abbreviation (`srch-`) is already in `abbreviations.json` so I just updated this.

cli/azd/resources/scaffold/templates/main.bicept
wbreza: As an app dev using composability I don't want to call `azd show <resource>` multiple times to discover some environment variables that I need to connect the service.

I would rather see a naming convention something like `AZD_RESOURCE_<NAME>_<VAR>` where `<NAME>` is the alias of the resource defined in the `azure.yaml` and `<VAR>` is whatever variable the resource would expose.  This enables support for multiple instances of the same resource type and gets us back to a point where the `azd` env is directly usable by the consuming applications similar to traditional templates.

JeffreyCA: Thanks all for your inputs, I'll resolve this for now and create a follow up issue where we can continue the discussion since the scope of the discussion is much larger.

PR #4939, opened 2025-03-13
azd init with env file

vhvb1989:
Fix: [Issue #4935](https://github.com/Azure/azure-dev/issues/4935)

This PR enhances the `azd init` command to utilize an initial `.env` file, allowing it to influence the initial state of the environment created during initialization. 

With this feature, users can run `azd init -t some-template` from a directory containing a pre-configured `.env` file with the necessary values for the environment's starting state. This is particularly useful for connecting to existing resources when initializing a template.

Additionally, users can set the key `AZD_ALLOW_NON_EMPTY_FOLDER=` in the `.env` file to bypass the warning about initializing a template in a non-empty directory. This assumes that the `.env` file and the current directory are pre-configured and ready for `azd init -t template-name`.

After this PR, users can define the name of the environment with any of the next alternatives:
- 1. Setting AZURE_ENV_VAR as system environment. AZD takes the value (if not empty) w/o asking or confirming the name to create the environment as part of `azd init`
- 2. Using argument `-e name`.  AZD overrides any value from the system environment and takes the value defined as argument during the call to `azd init -e name`.
- 3. Using `.env` file.  Set `AZURE_ENV_VAR=name` inside the .env file and run `azd init` in the folder where the .env file is. AZD overrides the System Environment with the values from the .env file. However, arguments are not overridden and are the highest priority

### DIFF
--- a/cli/azd/cmd/init.go
+++ b/cli/azd/cmd/init.go
@@ -30,6 +30,7 @@ import (
 	"github.com/azure/azure-dev/cli/azd/pkg/tools"
 	"github.com/azure/azure-dev/cli/azd/pkg/tools/git"
 	"github.com/fatih/color"
+	"github.com/joho/godotenv"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
 )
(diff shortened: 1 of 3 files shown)


cli/azd/cmd/init.go
```diff
@@ -165,6 +166,24 @@ func (i *initAction) Run(ctx context.Context) (*actions.ActionResult, error) {
 		Title: "Initializing an app to run on Azure (azd init)",
 	})
 
+	// AZD supports having .env at the root of the project directory as the initial environment file.
```
rajeshkamal5050, line 169: Any recommendations for the .env post initializing a template with it? Since it might be lingering? its also excluded in .gitignore?

rajeshkamal5050 (approved):

weikanglim:
@vhvb1989 @rajeshkamal5050  I'm wondering if `.env` in the root folder being checked in is idiomatic to the dev ecosystem. Normally, `.env` is git ignored, and is reserved for application development. Here we're using it explicit for an azd feature.

- For template starting state: it feels that the metadata would be stored inside `azure.yaml` under a `template` node for template customization?
- For general, non-template inntialization: it feels that an env var could be at the system/container level that influences azd?

vhvb1989:
> @vhvb1989 @rajeshkamal5050 I'm wondering if `.env` in the root folder being checked in is idiomatic to the dev ecosystem. Normally, `.env` is git ignored, and is reserved for application development. Here we're using it explicit for an azd feature.
> 
> * For template starting state: I'm wondering how we feel about metadata stored inside `azure.yaml` under a `template` node for template customization.
> * For general, non-template inntialization: I'm wondering if there are system/container env variables that could be used to influence azd

the `.env` is not expected to be committed to the repo.  It is a way to enable/set an initial state for a template azd will init.

The general use-case will look like:
- You are given an initial set of files in a folder and an .env file.
- If those files are python files, you can run them directly.  Those files can use the values from the .env in that directory.
   ** NOTE: **  At to this point, there is nothing about azd or azd templates
- Within the files, you can include a README file where you can say something like:  "If you want to see an entire project sample, run `azd init -t sample/cool-example` from the root and then `azd up` to deploy it to Azure".
  - Note that the experience is not starting from an azd-template, but making it an optional follow-up.
  - If folks want to explore the azd-template alternative/option, the expectation is to connect the template to the same existing services described in the `.env` file (instead of running `azd env set SERVICE url` for each service after azd init) as part of the `azd init` flow.

There is a little room for misunderstanding for customers after running `azd up`, as they might think they can update the `.env` file from the root and re-deploy the app with `azd up`.  If this is ever observed, we can include a prompt at the end of `azd init` to ask if you want to delete the `.env` file.

So, think about this change as a way to enable/empower someone who wants to give a step-by-step class to an audience. They want to start by having Azure resources created and focus on how to connect to them (teaching about client libraries to talk to resources).  They show an initial super simple script with just the basic connection code and run it to demonstrate how easy it is. Then, as a second step, they run `azd init -t full-sample` to show a real and complete example in action. But only after they have focused on the very basics.

Issue #4941, opened 2025-03-13
Feature: support setting multiple environment key-value pairs or from file


JeffreyCA:
It would be useful to be able to set multiple environment key-value pairs with a single command, something like:

```
azd env set KEY1 val1 KEY2 val2 ...
```

Or additionally, loading environment values from an external file (related to #4935 but outside of `azd init`), like:

```
azd env set --file .env
```

weikanglim:
Just curious to learn from @JeffreyCA the scenario you ran into that pushed towards the consideration of `azd env set <key> <value> <key> <value>`.

I could see `azd env set --file .env` or something similar for programmatic purposes, but it feels that for a general terminal experience `azd env set <key> <value>` is more readable than the proposed shorthand.

I notably find whitespace delimiter being used for `key|value` and `key-value|key-value| is giving me a little pause.

JeffreyCA:
The scenario is being able to set multiple environment key-value pairs in one command. Alternating `key/value` was the first approach that came to my mind initially. Now that I think about it, that may have been influenced by my experience with the Redis API, which uses a similar syntax in some commands like [MSET](https://redis.io/docs/latest/commands/mset/). 

It looks like many other CLIs (az cli, kubectl) adopt a different syntax like `key=value key2=value2`, which I think is worth exploring further and might feel better as well.

What do you think about something like:

| Command                                | Description                         |
|----------------------------------------|-------------------------------------|
| `azd env set <key> <value>`              | Set a single environment variable   |
| `azd env set <key>=<value> <key2>=<value2> ... <keyN>=<valueN>` | Set 1 or more environment variables |
| `azd env set --file <filepath>`          | From file, maybe `-f` as shorthand |

PR #4942, opened 2025-03-13
Update `env set` to accept multiple key-value pairs

JeffreyCA:
Contributes to #4941

This PR updates the `env set` command so you can now set multiple environment key-value pairs, like:

```
azd set key1 val1 key2 val2
```

https://github.com/user-attachments/assets/604ef21c-66fc-408a-be1c-d9d1c68bd4da

### DIFF
--- a/cli/azd/cmd/env.go
+++ b/cli/azd/cmd/env.go
@@ -116,9 +116,18 @@ func newEnvSetFlags(cmd *cobra.Command, global *internal.GlobalCommandOptions) *
 
 func newEnvSetCmd() *cobra.Command {
 	return &cobra.Command{
-		Use:   "set <key> <value>",
-		Short: "Manage your environment settings.",
-		Args:  cobra.ExactArgs(2),
+		Use:   "set <key> <value> [<key2> <value2> ...]",
+		Short: "Set one or more environment values.",
+		Long:  "Set one or more key-value pairs in the environment. Each key must be followed by its value.",
+		Args: func(cmd *cobra.Command, args []string) error {
+			if len(args) < 2 {
+				return fmt.Errorf("requires at least one key-value pair")
+			}
+			if len(args)%2 != 0 {
+				return fmt.Errorf("each key must be followed by a value")
+			}
+			return nil
+		},
 	}
 }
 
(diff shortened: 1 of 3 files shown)


cli/azd/cmd/env.go
```diff
@@ -116,9 +116,22 @@ func newEnvSetFlags(cmd *cobra.Command, global *internal.GlobalCommandOptions) *
 
 func newEnvSetCmd() *cobra.Command {
 	return &cobra.Command{
-		Use:   "set <key> <value>",
-		Short: "Manage your environment settings.",
-		Args:  cobra.ExactArgs(2),
+		Use:   "set <key> <value> [<key2> <value2> ...]",
```
vhvb1989, line 119: In the current implementation, if someone forgets to add quotes for a value with spaces, you get an error. For example:

azd env set key this is my value foo   (fails with error:  exact 2 expected)

Folks can then update to

azd env set key "this is my value foo"  

I think this is a common copy/paste issue (I've been there myself).  I am a little worried that, with this change, we will remove the error and just add keys: [key, is, value] .

Should we introduce a confirmation prompt (when setting more than just one var) before updating the environment?  Like:

azd env set key this is my value foo 
The key [key, is and value] will be set now. Confirm [Y/n]:  ?

Folks who are super sure about what they are doing can use `--no-prompt` to skip the confirmation. But other folks who might be on the copy/paste danger zone will be able to prevent an unfortunate event XD

JeffreyCA: Good callout, I like that idea

Not urgent but would love to get your thoughts here too @SophCarp. Should there also be any confirmation message that's displayed after they answer 'yes' or 'no'?

image

vhvb1989 (approved):

cli/azd/cmd/env.go
weikanglim: LIke @vhvb1989, I feel slightly torn about the overall UX presented here, but I feel the confirmation prompt is a band-aid that I'm not entirely in favor of. 

If it was `azd env set KEY=VALUE KEY=VALUE`, or `azd env set --KEY VALUE --KEY VALUE` I'd be super happy, but the current whitespace delineation makes the UX a little messy.

weikanglim: Don't have any immediate suggestions, but was curious how we can find a way to satisfy the broader scenario. Left a comment in the [issue](https://github.com/Azure/azure-dev/issues/4941#issuecomment-2731182898)

JeffreyCA: I pushed some changes to use the `azd env set KEY=VALUE KEY=VALUE` syntax and a working `azd env set --file <filename>`. Let me know what you think

There's definitely room for more hardening against special characters like whitespaces and quotation marks, but for now I've kept most of the same behaviour as the existing `azd env set`.

cli/azd/cmd/env.go
```diff
@@ -164,13 +173,57 @@ func newEnvSetAction(
...
+	} else {
+		// Handle key=value format: azd env set key=value [key2=value2 ...]
+		for _, arg := range e.args {
+			key, value, err := parseKeyValue(arg)
+			if err != nil {
+				return nil, err
+			}
+			keyValues[key] = value
+		}
+	}
```
wbreza, line 213: We would also support multiple <key> <value> syntax like `azd env set key1 value1 key2 value2` or would we require `=` for setting multiple values like `azd env set key1=value1 key2=value2` ?

vhvb1989 (approved):

cli/azd/cmd/env.go
JeffreyCA: Just the `azd env set key1=value1 key2=value2` syntax for setting multiple values

PR #4943, opened 2025-03-13
compose: `existing` support

weikanglim:
This change adds support for referencing `existing` resources.

## User-facing changes

`azd add` now supports adding existing resources by listing resources interactively. In the schema, a new `existing` keyword is added.

An example resulting schema looks like:

```yaml
  shared:
    type: storage
    existing: true
```

The resource ID, in the format of, `AZURE_RESOURCE_SHARED_STORAGE_ID` is stored in azd's `.env` file to support cross environment.

Existing resources support `uses` linking, just as before:

```yaml
  stg-exist:
    type: host.containerapp
    port: 80
    uses:
      - shared # Adds env variables `AZURE_STORAGE_SHARED_ACCOUNT_NAME`,  `AZURE_STORAGE_SHARED_ACCOUNT_BLOB_ENDPOINT`
```

`azd show <resource>`, i.e. `azd show shared` also works to display the environment variables for existing resources.

### Backend Bicep implementation

1. The resource is added as an `existing` resource to fetch properties.
2. The resource is assumed to be using managed identity, role permissions are assigned to the appropriate service identity when `uses` is added.

## Demo: 

https://github.com/user-attachments/assets/3254fbd2-9938-498e-8bc2-78084a9dde14

Completes #4577

### DIFF
--- a/cli/azd/internal/cmd/add/add_select.go
+++ b/cli/azd/internal/cmd/add/add_select.go
@@ -159,7 +159,6 @@ func (a *AddAction) selectExistingResource(
 			}
 
 			if menu.Namespace == "host" || // host resources are not yet supported
-				menu.Namespace == "keyvault" || // keyvault (RBAC) is not yet supported
 				menu.Namespace == "db" { // db resources are not yet supported
 				continue
 			}
--- a/cli/azd/internal/cmd/add/add_select.go
+++ b/cli/azd/internal/cmd/add/add_select.go
@@ -11,6 +11,9 @@ import (
 	"strings"
 
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/arm"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/to"
+	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/resources/armresources"
+	"github.com/azure/azure-dev/cli/azd/internal/scaffold"
 	"github.com/azure/azure-dev/cli/azd/pkg/azapi"
 	"github.com/azure/azure-dev/cli/azd/pkg/input"
 	"github.com/azure/azure-dev/cli/azd/pkg/project"
@@ -154,6 +157,7 @@ func (a *AddAction) selectExistingResource(
 				selectMenu = append(selectMenu, menu)
 			}
 		}
+
 		slices.SortFunc(selectMenu, func(a, b Menu) int {
 			return strings.Compare(a.Label, b.Label)
 		})
@@ -178,7 +182,17 @@ func (a *AddAction) selectExistingResource(
 		}
 
 		azureResourceType := r.Type.AzureResourceType()
-		resourceId, err := a.promptResource(ctx, "Which resource?", azureResourceType)
+		resourceMeta, ok := scaffold.ResourceMetaFromType(azureResourceType)
+		if ok && resourceMeta.ParentForEval != "" {
+			azureResourceType = resourceMeta.ParentForEval
+		}
+
+		displayName := azureResourceType
+		if friendlyName := azapi.GetResourceTypeDisplayName(azapi.AzureResourceType(azureResourceType)); friendlyName != "" {
+			displayName = friendlyName
+		}
+
+		resourceId, err := a.promptResource(ctx, fmt.Sprintf("Which %s resource?", displayName), azureResourceType)
 		if err != nil {
 			return nil, fmt.Errorf("prompting for resource: %w", err)
 		}
@@ -213,12 +227,12 @@ func (a *AddAction) promptResource(
 	msg string,
 	resourceType string,
 ) (string, error) {
-	options := azapi.ListResourcesOptions{
-		ResourceType: resourceType,
+	options := armresources.ClientListOptions{
+		Filter: to.Ptr(fmt.Sprintf("resourceType eq '%s'", resourceType)),
 	}
 
 	a.console.ShowSpinner(ctx, "Listing resources...", input.Step)
-	resources, err := a.resourceService.ListResources(ctx, a.env.GetSubscriptionId(), &options)
+	resources, err := a.resourceService.ListSubscriptionResources(ctx, a.env.GetSubscriptionId(), &options)
 	if err != nil {
 		return "", fmt.Errorf("listing resources: %w", err)
 	}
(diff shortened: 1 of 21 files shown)


cli/azd/internal/scaffold/resource_meta.go
```diff
@@ -25,6 +25,39 @@ type ResourceMeta struct {
...
 	// To evaluate the actual values, see [Eval].
 	Variables map[string]string
+
+	// RoleAssignments are related role assignments the resource.
+	RoleAssignments RoleAssignments
+}
+
+type RoleAssignments struct {
+	Read  []RoleAssignment
+	Write []RoleAssignment
```
JeffreyCA, line 35: How is the Read/Write distinction going to be used?

cli/azd/internal/scaffold/spec.go
```diff
@@ -176,6 +185,19 @@ type StorageReference struct {
...
 
+type ExistingResource struct {
+	// The unique logical name of the existing resource in the infra scope.
+	Name string
+	// The resource ID of the resource.
+	ResourceIdEnvVar string
+	// The resource type of the resource. This should match the type contained in the resource ID.
+	ResourceType string
+	// The API version of the resource to look up values.
+	ApiVersion string
```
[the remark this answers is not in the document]
weikanglim: We do need this today. The Bicep `existing` keyword requires an API version, and we need `existing` to grab the resource attributes from the API.

cli/azd/pkg/azapi/resource_service.go
```diff
@@ -251,6 +255,45 @@ func (rs *ResourceService) ListSubscriptionResources(
 	return resources, nil
 }
 
+// ListResources returns a slice of resources - optionally filtered on fields in `ListResourcesOptions` - including the
+// ID, Name, Type, and Location of each resource.
+func (rs *ResourceService) ListResources(
```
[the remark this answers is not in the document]
weikanglim: I can leverage `ListSubscriptionResources`

cli/azd/internal/scaffold/resource_meta.go
weikanglim: It's unused today and just serves as metadata.

I suspect at some point, we may want the following:

```yaml
resources:
  api:
    uses:
    - name: storage
      permissions: [read | write]
```

To configure the level of permissioning required. I don't know what this means for the local user access, but this all came from @wbreza 's recent astute observation about blob storage access, that it's likely in some scenarios, the api may only want to serve files and not write content.

cli/azd/internal/scaffold/resource_meta.go
```diff
@@ -160,10 +236,20 @@ var Resources = []ResourceMeta{
 		ResourceType:      "Microsoft.MachineLearningServices/workspaces",
 		ResourceKind:      "Project",
 		ApiVersion:        "2024-10-01",
-		StandardVarPrefix: "AZURE_AI_PROJECT",
+		StandardVarPrefix: "AZURE_AIPROJECT",
```
weikanglim, line 239: `AZURE_AI_PROJECT` vs `AZURE_AIPROJECT`?

weikanglim: Do this separately

cli/azd/internal/cmd/add/add_select.go
```diff
@@ -126,3 +130,111 @@ func selectKeyVault(console input.Console, ctx context.Context, p PromptOptions)
...
+		}
+
+		res.Type = resourceType
+		res.ResourceId = resourceId.String()
+	}
+
+	return res, nil
+}
+
+func (a *AddAction) promptResource(
```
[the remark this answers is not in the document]
weikanglim: I did end up making the prompting logic a little more complicated in handling excluded resource IDs. The current implementation also lists location instead of RG. I think it would be a good idea to have both versions out and get some feedback on how/if a general resource prompt makes sense.

cli/azd/internal/scaffold/resource_meta.go
JeffreyCA, line 1: What are your thoughts about assigning roles for other sub resources like keys and certs for AKV and queues, tables, etc. for storage?

And how do you see this playing out in relation to how we currently define role assignments for new resources through AVM in `resources.bicept`?

rajeshkamal5050:
Nice work Wei !!

Good to see how this has shaped up with simple `existing` param in azure.yaml and the associated selecting existing resource experiences via `azd add`.

For later, are we thinking about pushing these `selecting existing resource experiences` also to azd provision. Since for composability enabled projects which have an expressed azure.yaml like below,
```
  shared-ai-project:
    type: ai.project
    existing: true
  shared-keyvault:
    type: keyvault
    existing: true
```

Running azd provision goes down the older bicep param prompting route and stores state in env's config.json

```
{
  "infra": {
    "parameters": {
      "existingSharedAiProjectId": "shared-ai-project-id",
      "existingSharedKeyvaultId": "shared-keyvault-id"
    }
  }
}
```

and subsequently azd show experiences might not work.

rajeshkamal5050 (approved):

cli/azd/internal/scaffold/resource_meta.go
```diff
@@ -133,6 +175,15 @@ var Resources = []ResourceMeta{
...
 		},
+		RoleAssignments: RoleAssignments{
+			Read: []RoleAssignment{
+				{
+					Name:               "Reader",
+					RoleDefinitionName: "Key Vault Secrets User",
+					RoleDefinitionId:   "4633458b-17de-408a-b874-0445c86b69e6",
+				},
+			},
+		},
```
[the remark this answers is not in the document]
weikanglim: I have a note down to create an issue after this lands -- we won't modify the IaC, but we'll prompt the user to modify it themselves

wbreza (commented):
Since the resource ID's are stored in the .env file - what is the story to support this for CI/CD?
Is there an env variable convention that can be used when resources are added outside the add flow (like via extensions)?

vhvb1989 (approved):

cli/azd/internal/scaffold/resource_meta.go
weikanglim: > What are your thoughts about assigning roles for other sub resources like keys and certs for AKV and queues, tables, etc. for storage?

This should be possible for the backend. From the frontend modeling/design perspective in compose, I suspect we'd need more granular types like "keyvault.secret". 

> And how do you see this playing out in relation to how we currently define role assignments for new resources through AVM in resources.bicept?

Eventually, as we scale up generation, we'll move this model to be slightly more dynamic and granular in generation.

PR #4944, opened 2025-03-14
WIP: AI Builder Extension

wbreza:
image
image
image


ccbarragan:
@wbreza @SophCarp Thank you for sharing your work in progress! Some ideas around folding in a bit more progressive disclosure, especially since there are many steps / sub-steps:

- after running `azd ai start`, a blurb or bullet list that tells me at a high-level what to expect (describe value prop and outcome). Something similar to:
  ```
  This tool helps you to build an AI app with Azure services. You'll be able to:

  - Start from an AI scenario, for example 'build a RAG app with custom data'
  - Configure resources to build your app, such as storage, database, and AI models 
  - Provision and deploy to Azure
  ```
- a progress indicator that shows where I'm at / how much further I have to go

wbreza:
@ccbarragan 
> * a progress indicator that shows where I'm at / how much further I have to go

Can you provide more context to the progress indicator idea? Since the flow is dynamic and you can have different paths depending on your selections it would be difficult to display a progress indicator since it could fluctuate? I'm also not clear on what or where the UX would look like for this and if we have any existing work we could pull from.

ccbarragan:
@wbreza 
> Since the flow is dynamic and you can have different paths depending on your selections it would be difficult to display a progress indicator since it could fluctuate? 

Good point! If the flow is dynamic, then a granular indicator of progress won't be helpful. This POC already does a good job of chunking the major steps + signposting each section for sighted users (using the magenta color as indicator of a major step).

SophCarp:
There are a couple style notes I'd update to be in line with the azd experience:
- When an answer is being input but hasn't been submitted yet, it should be white. Only after the answer is submitted does it turn bright blue. 
- I'd change the magenta section titles to white and the descriptions underneath to grey (secondary text) - that way we can reserve magenta for really important call outs. 
- All questions for the user should be bold white, so it's easy to ID for the reader
- I like the gap between each section, but within a section I think we don't need any line gaps. Especially if the descriptions are in secondary text, it can be one chunk. 

We discussed the idea for a quick option and for an undo option. I think neither are blocking, but we can set up some time to discuss!

wbreza:
@SophCarp 
Without the color the white headings, etc just fade into the background and makes it difficult to tell that you have transitioned to a new section.  Maybe magenta is not the right color - but if feels like it needs something with such a text heavy set of interctions.

**example**
image

wbreza:
Closed - refer to new PR #5085

PR #4946, opened 2025-03-15
Extensions: Adds execute permissions to extension binary artifact during installation

### LINKED ISSUE #4901: Extension commands not working in Linux environments
After installing `microsoft.azd.demo` in a Linux Dev Container, none of the `azd demo` commands are working:

Image

Upon further inspection inside `~/.azd/extensions/microsoft.azd.demo`, the binary appears to be missing execute permissions. Running `chmod +x` on the binary appears to fix the issue:

Image

wbreza:
Resolves #4901

### DIFF
--- a/cli/azd/pkg/extensions/manager.go
+++ b/cli/azd/pkg/extensions/manager.go
@@ -428,6 +428,7 @@ func (m *Manager) Install(ctx context.Context, id string, options *FilterOptions
 		targetPath := filepath.Join(targetDir, entryPoint)
 
 		// Need to set the executable permission for the binary
+		// This change is specifically required for Linux but will apply consistently across all platforms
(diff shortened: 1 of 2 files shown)


cli/azd/pkg/extensions/manager.go
```diff
@@ -426,6 +427,11 @@ func (m *Manager) Install(ctx context.Context, id string, options *FilterOptions
 
 		targetPath := filepath.Join(targetDir, entryPoint)
 
+		// Need to set the executable permission for the binary
+		if err := os.Chmod(targetPath, osutil.PermissionExecutableFile); err != nil {
```
vhvb1989, line 431: should we add a note about this only applies to Linux?

vhvb1989 (approved):

JeffreyCA (approved):
Thanks for fixing this!

PR #4947, opened 2025-03-15
allow user to disable provision preflight

### LINKED ISSUE #4932: [Preflight] allow option out feature
We need to provide a way to option out preflight validation.

User reported an issue where the preflight validation failed because he is trying to provision to azure, but he doesn't have enough access for role assignment. Preflight validation API is not checking user's permission on specific resource group, so this will not serve for the needs where admin want to only grant specific access to a user on individual resource group. 

Template used: [Azure-Samples/get-started-with-ai-chat: Basic sample for deploying chat web apps with Azure AI Foundry and SDKs](https://github.com/Azure-Samples/get-started-with-ai-chat) 

```
ERROR: error executing step command 'provision': deployment failed: error deploying infrastructure: validating deployment to subscription:
Validation Error Details:
InvalidTemplateDeployment: 
Deployment failed with multiple errors: 'Authorization failed for template resource '...' of type 'Microsoft.Authorization/roleAssignments'. 
The client '...' with object id '...' does not have permission to perform action 'Microsoft.Authorization/roleAssignments/write' at scope '/subscriptions/.../resourceGroups/rg-.../providers/Microsoft.Authorization/roleAssignments/...'.:
Authorization failed for template resource '...' of type 'Microsoft.Authorization/roleAssignments'. 
The client '...' with object id '...' does not have permission to perform action 'Microsoft.Authorization/roleAssignments/write' at scope '/subscriptions/.../resourceGroups/rg-.../providers/Microsoft.Authorization/roleAssignments/f...'.:
Authorization failed for template resource '...' of type 'Microsoft.Authorization/roleAssignments'....
```

hemarina:
- fix #4932 
- Added opt out support for skipping preflight validation

Current method (set .env):
During the discussion on Monday, we went through three methods including set config, set .env, or add flag. It is clear to me that set config is not the way we want as this lets users to completely disable Preflight in all scenarios. We want the disabled functionality to be more project specific. I thought about the trades off between set .env vs add flag and discussed with @weikanglim. Considering our use case, we prefer to be environment/project-specific and speaks the same language that template authors want to speak to users. I believe set .env method is what we are looking for. 
If disable: 
image

allow user to disable it when an error is blocking:
image

Previous method (set config): 
I used the similar logic of alpha feature and added option out feature so that we could use it to option out other features in future. If folks think it is duplication of alpha feature, happy to migrate it under alpha feature support or just directly support config. 

After gathering folks feedback, update the method to directly support config:
Migrate to use `azd config set provision.disableValidation on`. Use `validation` since preflight is an internal word.
If disable, run provision directly:
image

Allow user to disable it when an error is blocking:
image

### DIFF
--- a/cli/azd/pkg/environment/environment.go
+++ b/cli/azd/pkg/environment/environment.go
@@ -46,6 +46,9 @@ const ResourceGroupEnvVarName = "AZURE_RESOURCE_GROUP"
 // PlatformTypeEnvVarName is the name of the key used to store the current azd platform type
 const PlatformTypeEnvVarName = "AZD_PLATFORM_TYPE"
 
+// DisablePreflightName is used to allow users to skip the preflight validation check
+const DisablePreflightName = "AZURE_PROVISION_SKIP_VALIDATE"
+
 // The zero value of an Environment is not valid. Use [New] to create one. When writing tests,
 // [Ephemeral] and [EphemeralWithValues] are useful to create environments which are not persisted to disk.
 type Environment struct {
--- a/cli/azd/pkg/infra/provisioning/bicep/bicep_provider.go
+++ b/cli/azd/pkg/infra/provisioning/bicep/bicep_provider.go
@@ -23,6 +23,7 @@ import (
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/arm"
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/to"
 	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/cognitiveservices/armcognitiveservices"
+	"github.com/azure/azure-dev/cli/azd/internal"
 	"github.com/azure/azure-dev/cli/azd/pkg/account"
 	"github.com/azure/azure-dev/cli/azd/pkg/async"
 	"github.com/azure/azure-dev/cli/azd/pkg/azapi"
@@ -77,6 +78,7 @@ type BicepProvider struct {
 	portalUrlBase           string
 	subscriptionManager     *account.SubscriptionsManager
 	azureClient             *azapi.AzureClient
+	userConfigManager       config.UserConfigManager
 }
 
 // Name gets the name of the infra provider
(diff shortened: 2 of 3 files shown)


JeffreyCA (commented):
Small note on wording, I believe the more widely used term is "opt/opted/opting out" as a verb or "opt-out" as an adjective or noun

cli/azd/pkg/infra/provisioning/bicep/bicep_provider.go
```diff
@@ -611,16 +613,34 @@ func (p *BicepProvider) Deploy(ctx context.Context) (*provisioning.DeployResult,
...
+		if err != nil {
+			return nil, &internal.ErrorWithSuggestion{
+				Err: err,
+				Suggestion: fmt.Sprintf("To disable provision validation, please run %s.",
+					output.WithHighLightFormat("`azd config set %s on`", preflightDisableVar)),
+			}
+		}
+	} else {
+		warningMessage := fmt.Sprintf("WARNING: Provision validation is disabled. "+
+			"To enable it, please run `azd config set %s off`.\n",
```
vhvb1989, line 641:
```suggestion
			"To enable it, please run `azd config unset %s`.\n",
```

cli/azd/pkg/infra/provisioning/bicep/bicep_provider.go
```diff
@@ -611,16 +613,34 @@ func (p *BicepProvider) Deploy(ctx context.Context) (*provisioning.DeployResult,
...
-	)
+	userConfig, err := p.userConfigManager.Load()
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("loading user configuration: %w", err)
+	}
+
+	preflightDisableVar := "provision.disableValidation"
+	preflightDisable, exists := userConfig.GetString(preflightDisableVar)
+	if !(preflightDisable == "on" && exists) {
```
vhvb1989, line 623: Check `exists` first in the condition. If it `false` the second condition after && is not even evaluated.
Then, support any value for `preflightDisable` which is different then `off`, like `true`, `1`, `si`, `Y`, etc.
Optionally, re-use or copy this function: https://github.com/vhvb1989/azure-dev/blob/main/cli/azd/pkg/alpha/alpha_feature_manager.go#L139 to check if the setting is ON or OFF. 


```suggestion
	if exists && strings.ToLower(preflightDisable) != "off"  {
```

cli/azd/pkg/infra/provisioning/bicep/bicep_provider.go
```diff
@@ -611,16 +613,34 @@ func (p *BicepProvider) Deploy(ctx context.Context) (*provisioning.DeployResult,
...
-		optionsMap,
-	)
+	userConfig, err := p.userConfigManager.Load()
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("loading user configuration: %w", err)
+	}
+
+	preflightDisableVar := "provision.disableValidation"
+	preflightDisable, exists := userConfig.GetString(preflightDisableVar)
```
vhvb1989, line 622: I think `userConfig.GetString()` can return `false` if the key exists but it is not a String.
https://github.com/vhvb1989/azure-dev/blob/main/cli/azd/pkg/config/config.go#L273

cli/azd/pkg/infra/provisioning/bicep/bicep_provider.go
```diff
@@ -611,16 +612,27 @@ func (p *BicepProvider) Deploy(ctx context.Context) (*provisioning.DeployResult,
...
-		ctx,
-		bicepDeploymentData.Target,
-		bicepDeploymentData.CompiledBicep.RawArmTemplate,
-		bicepDeploymentData.CompiledBicep.Parameters,
-		deploymentTags,
-		optionsMap,
-	)
-	if err != nil {
-		return nil, err
+	if strings.ToLower(p.env.Getenv(environment.DisablePreflightName)) != "true" {
```
vhvb1989, line 615: `.Getenv()` falls back to check the system environment (not only the values from azd's .env file)

Do we want to make it only for azd's .env file?

In the `warningMessage` below we are telling folks they can run `azd env set XXXXX false` or remove the value from the .env;  If the env var is set on system environment, should we re-word the error?

cli/azd/pkg/environment/environment.go
```diff
@@ -46,6 +46,9 @@ const ResourceGroupEnvVarName = "AZURE_RESOURCE_GROUP"
 // PlatformTypeEnvVarName is the name of the key used to store the current azd platform type
 const PlatformTypeEnvVarName = "AZD_PLATFORM_TYPE"
 
+// DisablePreflightName is used to allow users to skip the preflight validation check
+const DisablePreflightName = "AZURE_PROVISION_SKIP_VALIDATE"
```
vhvb1989, line 50: Consider adding this env var to: https://github.com/Azure/azure-dev/blob/main/cli/azd/docs/environment-variables.md

hemarina:
Closing this PR for now as no users are asking about preflight option out for a while. Will reopen this PR if needed.

Issue #4700, opened 2025-01-16
[Issue] Azure/setup-azd gives 'Unpinned tag' alert


RicardoNiepel:
Using `Azure/setup-azd@v2` gives "Unpinned tag for a non-immutable Action in workflow" as "Code scanning alert" inside GitHub.

What's the recommended practice here?

**CodeQL Rule:**
https://github.com/github/codeql/blob/main/actions/ql/src/Security/CWE-829/UnpinnedActionsTag.ql

**CWE:**
https://cwe.mitre.org/data/definitions/829.html

**Details:**
![Image](https://github.com/user-attachments/assets/cd75b12b-c679-44c8-a307-419ee5f6d56d)

rajeshkamal5050:
@RicardoNiepel thanks for reporting.

@hemarina can you triage this?

aegilops:
Pinning to a specific release such as `v2.1.0` is not sufficient to protect against supply chain attacks.

Pinning to a commit is the only current way to ensure a particular unchangeable version of the Action is used.

PR #4951, opened 2025-03-17
update templates to use setup-azd with a specific commit

hemarina:
https://github.com/Azure/azure-dev/issues/4700

### DIFF
--- a/cli/azd/pkg/pipeline/testdata/Test_promptForCiFiles-no_files_-_github_selected_-_App_host_-_fed_Cred.snap
+++ b/cli/azd/pkg/pipeline/testdata/Test_promptForCiFiles-no_files_-_github_selected_-_App_host_-_fed_Cred.snap
@@ -27,7 +27,7 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v4
       - name: Install azd
-        uses: Azure/setup-azd@v2
+        uses: Azure/setup-azd@ae0f8b5482eeac61e940f447327d84c73beb8b1e
       - name: Setup .NET
         uses: actions/setup-dotnet@v4
         with:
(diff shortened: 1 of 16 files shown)


cli/azd/pkg/pipeline/testdata/Test_promptForCiFiles-no_files_-_github_selected_-_App_host_-_fed_Cred.snap
```diff
@@ -27,7 +27,7 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v4
       - name: Install azd
-        uses: Azure/setup-azd@v2
+        uses: Azure/setup-azd@ae0f8b5482eeac61e940f447327d84c73beb8b1e
```
weikanglim, line 30: @danieljurek Is this a recommended change from what we discussed?

I may have misunderstood from our conversation earlier today, as I was under the impression that we were only updating our recommendation in the linked issue but not changing anything released by azd.

The recent incident unrelated to us exposed a vulnerability, but the vulnerability itself is presented in upstream. Version pinning is indeed a security hardening effort that should be recommended if they feel this is a strong vector of attack, and I would whole-heartedly recommend commit hash pinning in highly secure environments, but as a default for OSS this seems way overblown. I may be completely wrong in positioning here.

One line of inquiry that may be worth considering here: do we equally trust any of the language setup tools like `setup-dotnet@v4` that are unpinned?

rajeshkamal5050: +1 don't think we need to change our templates.

PR #2875, opened 2023-10-17
Allow app settings to be configured in `main.parameters.json` for simplified init apps

weikanglim:
Allow app settings to be configured in `main.parameters.json` for simplified init apps.

To this end, the bicep files in simplified init apps has been updated to allow flowing of any required app settings to the container app host, which enables configuration to be done simply as the following:
**main.parameters.json**
```json
    "apiDefinition": {
      "value": {
        "settings": [
          {
            "name": "AZURE_ENV_NAME",
            "value": "${AZURE_ENV_NAME}",
          },
          {
            "name": "OPENAI_API_KEY",
            "value": "${OPENAI_API_KEY}",
            "secret": true
          }
        ]
      }
  }
```

To ease discovery, the generated `main.parameters.json` includes placeholder values and documentation. `next-steps.md` has also been updated to include the information.

image

Fixes #2739

PR #4960, opened 2025-03-18
compose: update service-bus sku, remove api passing

weikanglim:
This change updates the SKU for service bus to Standard, as opposed to the default value in AVM which is Premium. Premium sku was accounting for a cost total of 200+ per month.

It also includes the following the changes:

1. **Remove `main.parameters.json` setting passing**: Remove an old feature of passing host settings through main.parameters.json. This was previously introduced in #2875 for simplified init (`azd init` with `azd config set alpha.compose off`), that hasn't been widely adopted in real world usage. It is also unused when `alpha.compose` is enabled.

2. **Fix simplified init and tests**: Fixes simplified init not automatically setting a KeyVault and failing to provision in the process. This was caught by `TestExecInfra` in scaffold_test.go, but this test was not running in CI. This has also been fixed to be running with a change in `ci-test.ps1`.

### DIFF
--- a/cli/azd/internal/repository/infra_confirm.go
+++ b/cli/azd/internal/repository/infra_confirm.go
@@ -27,7 +27,10 @@ func (i *Initializer) infraSpecFromDetect(
 	ctx context.Context,
 	detect detectConfirm) (scaffold.InfraSpec, error) {
 	spec := scaffold.InfraSpec{}
+
+	hasDb := false
 	for database := range detect.Databases {
+		hasDb = true
 		if database == appdetect.DbRedis {
 			spec.DbRedis = &scaffold.DatabaseRedis{}
 			// no further configuration needed for redis
@@ -59,6 +62,10 @@ func (i *Initializer) infraSpecFromDetect(
 			}
 			break dbPrompt
 		}
+	}
+
+	if hasDb {
+		spec.KeyVault = &scaffold.KeyVault{}
 	}
 
 	for _, svc := range detect.Services {
--- a/cli/azd/internal/scaffold/scaffold_test.go
+++ b/cli/azd/internal/scaffold/scaffold_test.go
@@ -14,6 +14,7 @@ import (
 	"github.com/azure/azure-dev/cli/azd/pkg/tools/bicep"
 	"github.com/azure/azure-dev/cli/azd/test/mocks/mockinput"
 	"github.com/otiai10/copy"
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
 
@@ -117,6 +118,7 @@ func TestExecInfra(t *testing.T) {
 				EventHubs:      &EventHubs{},
 				StorageAccount: &StorageAccount{},
 				KeyVault:       &KeyVault{},
+				AISearch:       &AISearch{},
 				Services: []ServiceSpec{
 					{
 						Name: "api",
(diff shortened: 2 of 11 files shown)


cli/azd/internal/scaffold/scaffold_test.go
```diff
@@ -180,70 +184,6 @@ func TestExecInfra(t *testing.T) {
 				},
 			},
 		},
-		{
-			"API with MySQL",
```
weikanglim, line 184: Kept one database test around, otherwise I think we're okay with the encompassing All at the moment.

cli/azd/tools/avmres/main.go
```diff
@@ -13,8 +13,6 @@ import (
 	"slices"
 	"strings"
 
-	//nolint:ST1001
```
weikanglim, line 16: Was added mistakenly in #4914, removing it now.

cli/azd/internal/repository/infra_confirm.go
```diff
@@ -61,6 +64,10 @@ func (i *Initializer) infraSpecFromDetect(
 		}
 	}
 
+	if hasDb {
+		spec.KeyVault = &scaffold.KeyVault{}
```
rajeshkamal5050, line 68: This is to bring simplified-init parity with native compose db+key-vault explicit resources in azure.yaml?

rajeshkamal5050 (approved):

vhvb1989 (approved):

cli/azd/internal/repository/infra_confirm.go
weikanglim: It's to fix a small regression for `azd init` with `azd config set alpha.compose off` I discovered.

The databases always required a keyvault, but when we shifted the dependency ordering for composability -- the simplified init path which depends on infra generation layer directly broke. This is the minimal fix for this and hoping this part is just left unchanged until we remove the legacy codepath sometime in the future.

Noting that the regression was caught by the tests, but the tests weren't running in CI which was unfortunate.

JeffreyCA (approved):

cli/azd/internal/repository/infra_confirm.go
weikanglim: `cd cli/azd/internal/scaffold && go test` should reproduce the failures.

The tests will now run in CI, so at least we'll get coverage...

karpikpl:
Hey @weikanglim 
I'm trying to understand this change `Remove main.parameters.json setting passing: `

I have an azd app with
**main.parameters.json**:
```json
      "elasticSearchDefinition": {
        "value": {
          "cpu": "1.0",
          "memory": "2.0Gi",
          "settings": [ 
          ]
        }
      },
```

and **main.bicep**:
```bicep
param elasticSearchDefinition object
```

that started failing after upgrade to azd 1.15

does it mean objects are no longer passed from `main.parameters.json` to bicep?

JeffreyCA:
> Hey @weikanglim I'm trying to understand this change `Remove main.parameters.json setting passing: `
> 
> I have an azd app with **main.parameters.json**:
> 
> ```json
>       "elasticSearchDefinition": {
>         "value": {
>           "cpu": "1.0",
>           "memory": "2.0Gi",
>           "settings": [ 
>           ]
>         }
>       },
> ```
> 
> and **main.bicep**:
> 
> ```bicep
> param elasticSearchDefinition object
> ```
> 
> that started failing after upgrade to azd 1.15
> 
> does it mean objects are no longer passed from `main.parameters.json` to bicep?

Hi, this is likely due to https://github.com/Azure/azure-dev/issues/5151 - we're releasing a new update that fixes this soon.

Issue #4933, opened 2025-03-12
compose: show UX enhancements


weikanglim:
Captured from [UX Review](https://github.com/Azure/azure-dev/pull/4914#issuecomment-2711971793)

Immediate items:
- [ ] Color scheme for environment variables
- [ ] Color for show resource heading -- type and name
- [ ] Display logical name / where does resource name go?

Discussion items:
- [ ] Can help text for `azd show` be updated to include resources? -- Dev note: Currently this is in alpha, have to investigate if dynamic help text is possible 
- [ ] Should `azd show` include resources?
- [ ] Environment variables vs. variables.
    - Dev note: There's a slight nuance between variables displayed for a resource that isn't "attached to an environment"

weikanglim:
Probably fix `AZURE_OPENAI_ENDPOINT` missing for preview

rajeshkamal5050:
Closing this issue as not planned since we are exploring on a simpler, more flexible and scalable(with AI/Agentic) compose v2. Stay tuned for updates, your feedback is always welcome! @kristenwomack @puicchan @tg-msft

PR #4966, opened 2025-03-19
compose: openai endpoint

weikanglim:
Fix `AZURE_OPENAI_ENDPOINT` not showing in add preview.

In #4914, we shifted the preview to leverage the newly added resource metadata. Azure OpenAI models had a resource specific show implementation built that wasn't immediately refactored in the change. This change shifts the Azure OpenAI implementation and fixes the add preview

Contributes to #4933

### DIFF
--- a/cli/azd/internal/cmd/show/show_resource.go
+++ b/cli/azd/internal/cmd/show/show_resource.go
@@ -99,6 +99,10 @@ func (s *showResource) showResourceGeneric(
 		Variables:   envValues,
 	}
 
+	if opts.resourceSpec != nil {
+		showRes.Name = opts.resourceSpec.Name
+	}
+
 	return &showRes, nil
 }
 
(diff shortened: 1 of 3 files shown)


cli/azd/internal/cmd/show/show_resource.go
JeffreyCA, line 1: I tried `azd show gpt-4o` on a gpt-4o OpenAI model and it seems to hang indefinitely on "Gathering information about your app and its resources..."

I tried debugging this and it looks like it might be infinite-looping in `getResourceMeta` as `parentId.ResourceType.Type` doesn't include the namespace (`Microsoft.CognitiveServices`) but `res.ParentForEval` does:

```go
for {
	if parentId.Parent != nil {
		parentId = parentId.Parent
	}

	if parentId.ResourceType.Type == res.ParentForEval {
		break
	}
}
```

vhvb1989 (approved):

cli/azd/internal/cmd/show/show_resource.go
weikanglim: My bad. It is meant to be `parentId.ResourceType.String()`. Fixing the bug and adding a panic to avoid the infite loop.

JeffreyCA (approved):

PR #4969, opened 2025-03-19
prompt: case-insensitive sorting for subscriptions

### LINKED ISSUE #4961: Case insensitive sorting for subscriptions list
It looks like the subscriptions list are ASCII-sorted, but then some developers dont realize their subscriptions are actually lowercased. Could it do some normalization of the casing before it sorts the subscriptions?

weikanglim:
Apply case-insensitive sorting at the subscription prompt display

Fixes #4961


JeffreyCA (dismissed):

vhvb1989 (dismissed):

rajeshkamal5050 (approved):

cli/azd/pkg/prompt/prompter.go
```diff
@@ -218,6 +219,10 @@ func (p *DefaultPrompter) getSubscriptionOptions(ctx context.Context) ([]string,
 		return nil, nil, nil, fmt.Errorf("listing accounts: %w", err)
 	}
 
+	sort.Slice(subscriptionInfos, func(i, j int) bool {
+		return strings.ToLower(subscriptionInfos[i].Name) < strings.ToLower(subscriptionInfos[j].Name)
```
weikanglim, line 223: The calls to `strings.ToLower`  here ends up allocating 2x string copies for each element.

A way to avoid string allocations is to leverage slices and lowercasing each character rune. This kind of string comparison doesn't work for all languages -- but it would work for ASCII usage.

cli/azd/pkg/prompt/prompter.go
```diff
@@ -157,7 +157,7 @@ func (p *DefaultPrompter) PromptResourceGroupFrom(
 	}
 
 	slices.SortFunc(groups, func(a, b *azapi.Resource) int {
-		return strings.Compare(a.Name, b.Name)
+		return stringutil.CompareLower(a.Name, b.Name)
```
weikanglim, line 160: @vhvb1989 , @JeffreyCA made a really good point on the linked issue that resource groups should behave similarly. 

I suspect it goes beyond just these resources, so I added a new comparison function that avoids string allocations (previous implementation used `strings.ToLower` which creates a new string copy per each element iteration), if we wanted to do this more generally.

However, I'm truly wondering if we're down the right track here. There are many different Azure resources that will likely encounter the same problem, and I'm not sure if we'll want to apply this rule consistently.

Does the case-insensitive comparison truly feel any better? Presentation-wise it can get look a little messy at times with mixed capitalization. Curious what everyone thinks here -- My position shifted to slightly against this after looking at it more.

vhvb1989: As long as the output is deterministic, I don't have any strong opinion about the sorting details.

Similar prompts are: Key vault Accounts and Secrets (or maybe not due to naming-rules for key vault), AI Models listing, AI Projects listing, etc.

wbreza: I wouldn't have any objections with just an easy normalized to lower based sorting instead of something more complex.

rajeshkamal5050 (approved):

PR #4973, opened 2025-03-21
Remote environment: Support infrastructure files for Terraform.

### LINKED ISSUE #4756: AZD Remote environment with Terraform local backend
When AZD is configured to use Remote environment, it stores the environment related files in the storage account. 

Using Terraform as IaC deployment tool, there are env specific additional files withing (.terraform) folder that gets created inside the .azure/ENVNAME folder like state files, and plan files used for the deployment folder. These files need to be stored to the env files in the storage account as well to be able to have the same consistent deployment details and state file available when using different vms as the files now stored globally and not locally on the machine.

Repro step:

- Clone terrasform template like https://github.com/Azure-Samples/todo-nodejs-mongo-terraform.git
- Update azure.yaml to enable azd remote state 
 ``` yaml
name: azd-project-name
state:
  remote:
    backend: AzureBlobStorage
    config:
      accountName: azdremotestate
      containerName: myproject # Defaults to project name if not specified

```
- Run azd up cmd to deploy the app and infrastructure.
- Check the storage account for the environment. the state files are not there

### PR DESCRIPTION
**Pull request by HadwaAbdelhalem**
Using Terraform as IaC deployment tool, there are env specific additional files that gets created inside the .azure/ENVNAME folder like state files, and plan files used for the deployment folder. 

This PR updates the remote environment feature to manage the infrastructure files for terraform deployments in remote data store (azure blob) when azd is configured to use remote environment offering consistent deployment details.

Changes in behavior 
``` bash
azd provision 

// now stores the additional .azure/envName/infra directory files in the remote environment storage blob after the deployment completes
```

```bash
azd env select <EnvName>

// now restores the envName/infra directory files in the remote environment storage blob to the local azure/envName/infra folder 
```

fixes #4756 
closes #4756

### DIFF
--- a/cli/azd/pkg/environment/storage_blob_data_store.go
+++ b/cli/azd/pkg/environment/storage_blob_data_store.go
@@ -36,7 +36,8 @@ type StorageBlobDataStore struct {
 	azdContext    *azdcontext.AzdContext
 }
 
-func NewStorageBlobDataStore(configManager config.Manager, blobClient storage.BlobClient, azdContext *azdcontext.AzdContext) RemoteDataStore {
+func NewStorageBlobDataStore(configManager config.Manager, blobClient storage.BlobClient,
+	azdContext *azdcontext.AzdContext) RemoteDataStore {
 	return &StorageBlobDataStore{
 		configManager: configManager,
 		blobClient:    blobClient,


cli/azd/pkg/environment/storage_blob_data_store.go
```diff
@@ -143,6 +153,22 @@ func (sbd *StorageBlobDataStore) Save(ctx context.Context, env *Environment, opt
...
+		fileBuffer, err := os.Open(file)
+		if err != nil {
+			return fmt.Errorf("failed to open file for upload: %w", err)
+		}
+		defer fileBuffer.Close()
+		err = sbd.blobClient.Upload(ctx, fmt.Sprintf("%s/%s/%s", env.name, InfraFilesDir, fileName), fileBuffer)
+		if err != nil {
+			return fmt.Errorf("uploading infra file: %w", err)
+		}
+	}
```
wbreza, line 170: Instead of being selective here - what if we just upload any files that are in the environment folder? That would solve this specific use case as well as any other similar issues.

HadwaAbdelhalem: @wbreza I wanted to be more specific in case users have local tmp, or test files in the local environment folder that was not meant to be persisted and to follow the pattern of the environment items being defined ( configfile, env file) by adding infrafiles

cli/azd/pkg/environment/storage_blob_data_store.go
```diff
@@ -143,6 +153,22 @@ func (sbd *StorageBlobDataStore) Save(ctx context.Context, env *Environment, opt
...
+	filesToUpload := sbd.azdContext.GetEnvironmentInfraFiles(env.name)
+
+	for _, file := range filesToUpload {
+		fileName := filepath.Base(file)
+		fileBuffer, err := os.Open(file)
+		if err != nil {
+			return fmt.Errorf("failed to open file for upload: %w", err)
+		}
+		defer fileBuffer.Close()
+		err = sbd.blobClient.Upload(ctx, fmt.Sprintf("%s/%s/%s", env.name, InfraFilesDir, fileName), fileBuffer)
```
mattdot, line 166: How are secrets protected in this approach? Seems like a user could upload a lot of sensitive information to a storage account that I don't see any documentation on how to secure correctly.  There's no mention of turning on blob encryption, turning off public access, restricting network access, or any other best practices for putting tfstate in a storage account https://learn.microsoft.com/en-us/azure/developer/azure-developer-cli/remote-environments-support

If I'm working across two computers, how is the other computer notified of the change so I have the latest state?  Do I have to manually re-run `azd env refresh`?

mattdot (commented):
I think the better approach here is to just recommend that people use Terraform Remote State when they're using azd Remote Environments (perhaps with a warning when they are using local state files and azd remote env at the same time).  If we want to make it automagic then azd could dynamically generate a backend config for Terraform (pointing to the same Azure Storage container, perhaps under the hood using the environment’s storage account) and run terraform init with it.  There also seems to be some overlap with [terraform workspaces](https://developer.hashicorp.com/terraform/cli/workspaces) which could probably be better integrated here as well.

Even if we continue to implement this PRs approach, I think using terraform remote state should be a **strong recommendation** in the docs (over synchronizing local files) which we could also do a docs PR for.

Issue #4940, opened 2025-03-13
Support Deploy Prompt in Azd init


LianwMS:
Add the prompt to ask customer if want deploy template after current init complete.

Have synced with @rajeshkamal5050 about the requirement.

vhvb1989:
implemented by [Prompt Deploy after Init #4977](https://github.com/Azure/azure-dev/pull/4977)

Closing this now. @LianwMS , LMK if this needs to be re-open

PR #4977, opened 2025-03-24
Prompt Deploy after Init

LianwMS:
#4940


vhvb1989 (approved):

vhvb1989:
Tested locally with a few combinations of azd init, like using `-C`, using `-t xxx` , selecting a template.

NOTE:
** A breaking  change is expected for anyone currently running `azd init -t xxx --no-prompt` from Ci/Cd, as they would not be expecting the automatic-deployment to start happening.

PR #4978, opened 2025-03-24
[1ES] CLI Deployment migration

danieljurek:
CLI only
* Use only artifacts (where possible)
* Deployment jobs (where possible)

* Breaks release tasks into specific jobs with a dependency graph. This makes it so that a failure in one part of the sequence doesn't affect other release tasks.


rajeshkamal5050 (approved):

eng/pipelines/templates/stages/publish.yml
```diff
@@ -198,98 +232,98 @@ stages:
         - template: /eng/pipelines/templates/variables/globals.yml
 
       jobs:
-        - deployment: Upload_Version_Txt
+        - deployment: ReleaseGate
```
weshaggard, line 235: Why is this needed given you already have other deployment jobs?

eng/pipelines/templates/stages/publish.yml
```diff
@@ -80,102 +77,139 @@ stages:
...
           dependsOn: Publish_Release
           condition: >-
             and(
               succeeded(),
               ne('true', variables['Skip.Publish'])
             )
-          environment: azure-dev
 
           pool:
             name: azsdk-pool-mms-win-2022-general
```
weshaggard, line 115: Jobs should use images.yml to reduce the number of hard-coded images we have around.

eng/pipelines/templates/stages/publish.yml
danieljurek: The delay between a release publishing and Choco/WinGet having the latest version can be "long". Adding an approval allows azd to launch the version.txt update once the release is live everywhere.

weshaggard (approved):

PR #4979, opened 2025-03-25
aspire support webapp

vhvb1989:
Update deployment for Aspire services to fetch endpoints from WebApp.

This only works when Aspire generates bicep files for deployment.

After running deployment, azd checks the list of resources from the resource-group deployment result.
If a webApp resource is found, azd takes the name of the webapp from the resourceId and change the default implementation to fetch Endpoints from WebApp instead of ACA


cli/azd/pkg/project/service_target_dotnet_containerapp.go
```diff
@@ -393,6 +400,17 @@ func (at *dotnetContainerAppTarget) Deploy(
 		if err != nil {
 			return nil, fmt.Errorf("deploying bicep template: %w", err)
 		}
+		if slices.ContainsFunc(deploymentResult.Resources,
+			func(resource *armresources.ResourceReference) bool {
+				webAppType := string(azapi.AzureResourceTypeWebSite)
+				if foundWebApp := strings.Contains(*resource.ID, webAppType); foundWebApp {
+					resourceName = strings.Split(*resource.ID, webAppType+"/")[1]
```
wbreza, line 407: I'd recommend using `arm.ParseResourceID(resourceId)` method instead of the manual parsing here.

cli/azd/pkg/project/service_target_dotnet_containerapp.go
```diff
@@ -393,6 +400,17 @@ func (at *dotnetContainerAppTarget) Deploy(
 		if err != nil {
 			return nil, fmt.Errorf("deploying bicep template: %w", err)
 		}
+		if slices.ContainsFunc(deploymentResult.Resources,
+			func(resource *armresources.ResourceReference) bool {
```
wbreza, line 404: Nit: Consider declaring the comparison function outside of the `ContainsFunc()` call to making this a bit easier to read.

wbreza (approved):
A couple minor comments but otherwise looks good.

cli/azd/pkg/project/service_target_dotnet_containerapp.go
```diff
@@ -445,23 +463,47 @@ func (at *dotnetContainerAppTarget) Endpoints(
...
+			}
+
+			return endpoints, nil
+		}
+	}
 
-	if ingressConfig, err := at.containerAppService.GetIngressConfiguration(
+	// Not ACA, had to be WebApp
+	if resourceType != string(azapi.AzureResourceTypeWebSite) {
+		return nil, fmt.Errorf("unsupported resource type")
```
JeffreyCA, line 488: Could we log `resourceType` here for better debugging?

JeffreyCA (approved):

PR #4985, opened 2025-03-25
Ext Framework: Adds composability and workflow services

wbreza:
**Adds the following:**
- Compose Service: Manages composability resources in an `azd` project
- Workflow Service: Adds support to run `azd` workflow from within an extension


cli/azd/internal/grpcserver/compose_service.go
```diff
@@ -0,0 +1,234 @@
...
+	case project.ResourceTypeMessagingServiceBus:
+		props := project.ServiceBusProps{}
+		if len(config) == 0 {
+			return props, nil
+		}
+		if err := json.Unmarshal(config, &props); err != nil {
+			return nil, err
+		}
+		return props, nil
+	default:
```
JeffreyCA, line 231: Should this always match what's in `UnmarshalYAML` in `cli/azd/pkg/project/resources.go`? If so could we document this somehow?

wbreza: Yes, in the short term we'll need to keep this in sync.  I'll add a note in here.

vhvb1989 (approved):

PR #4993, opened 2025-03-26
[1ES] Vscode Deployment migration

danieljurek:


bwateratmsft (approved):

eng/pipelines/templates/steps/publish-vscode.yml
```diff
@@ -1,28 +1,29 @@
...
-      artifact: signed
-      path: signed
-
   - ${{ if eq('true', parameters.TagRepository) }}:
     # Perform tag verification before publishing anything
-    - task: PowerShell@2
-      inputs:
-        pwsh: true
-        targetType: filePath
-        filePath: ./eng/scripts/Set-GitHubReleaseTag.ps1
```
weshaggard, line 20: I assume you did this because it is happening in the release deployment job. We could still do the tagging in a normal job as well which is what we did in the sdk repos.

weshaggard: If we do stick with inline over this script we should probably delete the script.

eng/pipelines/templates/stages/vscode-build-and-test.yml
```diff
@@ -43,3 +43,68 @@ stages:
...
+
+        templateContext:
+          outputs: 
+            - output: pipelineArtifact
+              path: release-metadata
+              condition: succeeded()
+              artifact: release-metadata
+              displayName: Upload release metadata
+
+            - output: pipelineArtifact
```
weshaggard, line 108: Why is the changelog going as an independent artifact?

eng/pipelines/templates/steps/publish-vscode.yml
danieljurek: azd uploads signed artifacts when it creates a release. We could upload the artifacts in a separate step but this approach tags, creates the release, and uploads the artifacts in a single job.

eng/pipelines/templates/stages/vscode-build-and-test.yml
danieljurek: Changelog is used to create the GitHub release. The GitHub release also includes signed artifacts. Rather than break up the creation of the GitHub Release and the uploading of artifacts, I just made the release creation and artifact upload a deployment job.

weshaggard (approved):

danieljurek:
This PR has been tested. Latest version doesn't include changes which trigger CI so no pipelines run against that last commit. 

Overriding check-enforcer

PR #4995, opened 2025-03-27
Add Up Flag for Init from Template

LianwMS:


cli/azd/cmd/init.go
```diff
@@ -242,25 +250,27 @@ func (i *initAction) Run(ctx context.Context) (*actions.ActionResult, error) {
...
-			i.azd.SetArgs([]string{"up", "--cwd", azdCtx.ProjectDirectory()})
-			err := i.azd.ExecuteContext(ctx)
-			header = "New project initialized! Provision and deploy to Azure is completed in " +
-				ux.DurationAsText(since(startTime)) + "."
+		if i.flags.up {
+			// Prompt to deploy to Azure
+			deploy, err := i.console.Confirm(ctx, input.ConsoleOptions{
+				Message:      "Deploy to Azure now?",
+				DefaultValue: true,
+			})
```
JeffreyCA, line 258: Any thoughts on if we should omit the prompt, since the user has to explicitly pass in `--up` now?

cli/azd/cmd/init.go
```diff
@@ -98,6 +99,13 @@ func (i *initFlags) Bind(local *pflag.FlagSet, global *internal.GlobalCommandOpt
 		false,
 		"Initializes a new application from your existing code.",
 	)
+	local.BoolVarP(
+		&i.up,
+		"up",
+		"",
+		false,
+		"Provision and deploy to Azure after initializing the project from a template.",
+	)
```
JeffreyCA, line 108: We may need to also update the [`TestUsage-azd-init.snap`](https://github.com/Azure/azure-dev/blob/main/cli/azd/cmd/testdata/TestUsage-azd-init.snap) file - some instructions below:

https://github.com/Azure/azure-dev/blob/0340065907dee7ca71cd68ed132033cabb38cbce/cli/azd/test/snapshot/snapshot.go#L68-L70

cli/azd/cmd/init.go
vhvb1989: I think people would use `--no-prompt` for accepting the prompt right away (like in a Ci/Cd )

Only if the default option for the prompt is not Yes by default, then you would want to use the flag as the confirmation.

For example `azd down` supports  force and purge flags:
image

Only because the default for those questions is `no`; meaning that if you run `azd down --no-prompt` you would get an error when the confirmations are displayed.

For the case of `up`, the default is Yes, so the flag is actually to request the prompt and setting the default as yes.

So, probably we should rename the flag to `--prompt-up` ??

LianwMS: “If --up is not passed, prompt with default being No. If --no-prompt is passed, init but do not provision/deploy."
--This maybe a breaking change for current `init -t` user. They did not expect the option hang on prompt.

If --up is passed, do not prompt and always provision + deploy.
--I thought this before. But, currently, up is a heavy operation. It might be beneficial for users to be aware of this process once init is complete. Adding a prompt could help achieve this awareness, ensuring users understand that the next step involves provisioning and deployment.

cli/azd/cmd/init.go
LianwMS: Updated

cli/azd/cmd/init.go
```diff
@@ -242,25 +250,27 @@ func (i *initAction) Run(ctx context.Context) (*actions.ActionResult, error) {
...
 			if err != nil {
 				return nil, err
 			}
+
+			if deploy {
+				// Call azd up
+				startTime := time.Now()
+				i.azd.SetArgs([]string{"up", "--cwd", azdCtx.ProjectDirectory()})
+				err := i.azd.ExecuteContext(ctx)
+				header = "New project initialized! Provision and deploy to Azure is completed in " +
```
vhvb1989, line 268: nit
```suggestion
				header = "New project initialized! Provision and deploy to Azure was completed in " +
```

cli/azd/cmd/init.go
```diff
@@ -242,25 +250,27 @@ func (i *initAction) Run(ctx context.Context) (*actions.ActionResult, error) {
...
-			// Call azd up
-			startTime := time.Now()
-			i.azd.SetArgs([]string{"up", "--cwd", azdCtx.ProjectDirectory()})
-			err := i.azd.ExecuteContext(ctx)
-			header = "New project initialized! Provision and deploy to Azure is completed in " +
-				ux.DurationAsText(since(startTime)) + "."
+		if i.flags.up {
+			// Prompt to deploy to Azure
+			deploy, err := i.console.Confirm(ctx, input.ConsoleOptions{
+				Message:      "Deploy to Azure now?",
```
vhvb1989, line 256: This might sound a little cold/direct as a question.... compared to other prompts where we use the style of `Do you want to ....  ?`

I wonder if we should provide a little bit of an explanation/context.  Something like:

```
Template files have been initialized in your directory. At this point, everything is ready for running `azd up` to deploy your Azure resources and services. Do you want to do that now?
```

We can also add a `hint` for the prompt (is a property for the `input.ConsoleOptions` ) which is displayed when the user pressed the question mark.  We can use the hint to mention that the operation might take several minutes to complete.

++ @SophCarp (UX Design expert )  -  Wondering if you could advise on the wording for this prompt and the one below (the title at the end of provision)

SophCarp, line 256:
```suggestion
				Message:      "Do you want to run `azd up` now?",
```
I'd also add a help text with the content: 
"Template files have been initialized in your local directory. If you want to provision and deploy now without making changes, select Y. If not, select N." 

@vhvb1989 thoughts?

SophCarp:
image
Not sure if this is the PR or my set-up, but my project deployment is freezing after I select the Azure Location. It's been like this for at least 10 minutes.

JeffreyCA:
> image Not sure if this is the PR or my set-up, but my project deployment is freezing after I select the Azure Location. It's been like this for at least 10 minutes.

Do you have the demo extension installed? I've found it sometimes interferes with `azd up` and I have to uninstall it to unblock the operation.

vhvb1989 (approved):
LGTM

SophCarp:
> Do you have the demo extension installed? I've found it sometimes interferes with `azd up` and I have to uninstall it to unblock the operation.

Ahhh, yep I do, thanks!

cli/azd/cmd/init.go
```diff
@@ -242,25 +250,28 @@ func (i *initAction) Run(ctx context.Context) (*actions.ActionResult, error) {
...
-			startTime := time.Now()
-			i.azd.SetArgs([]string{"up", "--cwd", azdCtx.ProjectDirectory()})
-			err := i.azd.ExecuteContext(ctx)
-			header = "New project initialized! Provision and deploy to Azure is completed in " +
-				ux.DurationAsText(since(startTime)) + "."
+		if i.flags.up {
+			// Prompt to deploy to Azure
+			deploy, err := i.console.Confirm(ctx, input.ConsoleOptions{
+				Message: "Template files have been initialized in your local directory. " +
+					"If you want to provision and deploy now without making changes, select Y. If not, select N.",
```
JeffreyCA, line 257:
```suggestion
                Message: "Do you want to run " + output.WithHighLightFormat("azd up") + " now?",
				Help: "Template files have been initialized in your local directory. " +
					"If you want to provision and deploy now without making changes, select Y. If not, select N.",
```

I think the original suggestion from https://github.com/Azure/azure-dev/pull/4995#discussion_r2021759722 was to have both the question and help text - would something like this work?

image

After entering `?`:
image

JeffreyCA (approved):

cli/azd/cmd/init.go
vhvb1989: ok

vhvb1989 (approved):

rajeshkamal5050 (approved):

PR #5010, opened 2025-04-01
support usagename for main location

vhvb1989:
When adding the metadata description for the `location` input parameter in bicep, it was not working because how AZD handles the prompt for location and subscriptionID.

Regardless of what input parameters are defined in bicep, AZD always starts by asking for subscription Id and Location, expecting that the template author
will define a `location` parameter and make it be mapped to the AZURE_LOCATION env var.

This special behavior blocked AZD from looking into the `usageName` for the location parameter to handle AI-models filtering because the special prompt for `location` has no support for `usageName`, it only supports `allowedLocation` filtering.

The changes in this PR removes the special case for location, making AZD to prompt for location only when it is defined and as any other parameter.
During the parameter prompt, the bicep-provider is aware of the name of the parameter and still will handle `location` with the same rules to persist it 
in the .env file (instead of config), this is mainly to make it compatible for how everything is expected to work. 

This PR is also updating the logic for filtering AI Models because the current logic to check for locations where it is possible to create an Azure AI Services resource was not working as expected. While we found a new Endpoint or way to get the list of locations which supports Azure AI Services, we are adding the locations that the Azure Portal is providing currently for Azure AI Services creation. This will help to remove the cases where the user find a location where it is possible to deploy AI models (quota OK) but there is an error because the Azure AI Services resource can't be created at that location.


cli/azd/pkg/infra/provisioning/bicep/prompt.go
```diff
@@ -81,6 +81,37 @@ func autoGenerate(parameter string, azdMetadata azure.AzdMetadata) (string, erro
...
+	"southindia",
+	"southeastasia",
+	"swedencentral",
+	"switzerlandnorth",
+	"uaenorth",
+	"uksouth",
+	"westeurope",
+	"westus",
+	"westus3",
+}
```
JeffreyCA, line 113: From the network logs, the AI Services Portal blade seems to be getting these locations from the [Resource Skus - List endpoint](https://learn.microsoft.com/en-us/rest/api/aiservices/accountmanagement/resource-skus/list?view=rest-aiservices-accountmanagement-2024-10-01&tabs=HTTP). Could we use that instead of hardcoding locations here? When testing on my sub, I also get 26 entries/locations where `kind=AIServices` that match the locations here.

I'm not sure if this was what you were referring to in the description - I didn't quite understand this part:

> While we found a new Endpoint or way to get the list of locations which supports Azure AI Services, we are adding the locations that the Azure Portal is providing currently for Azure AI Services creation.

vhvb1989: updated

cli/azd/pkg/azapi/cognitive_service.go
```diff
@@ -153,3 +155,80 @@ func (cli *AzureClient) createUsagesClient(
...
+				}
+			}
+		}
+	}
+	if len(locationsUniqueName) == 0 {
+		return nil, fmt.Errorf("no locations found for sku %s", sku)
+	}
+	locations := make([]string, 0, len(locationsUniqueName))
+	for location := range locationsUniqueName {
+		println(location)
```
JeffreyCA, line 213: Should this be removed or just be a debug log?

cli/azd/pkg/infra/provisioning/bicep/prompt.go
```diff
@@ -222,174 +225,196 @@ func (p *BicepProvider) promptForParameter(
...
+				if allowedLocations == nil {
+					allLocations, err := p.subscriptionManager.ListLocations(ctx, p.env.GetSubscriptionId())
+					if err != nil {
+						return nil, fmt.Errorf("listing locations: %w", err)
+					}
+					allowedLocations = make([]string, len(allLocations))
+					for i, location := range allLocations {
+						allowedLocations[i] = location.Name
+					}
+				}
```
JeffreyCA, line 273: Could we skip this altogether, since if `allowedLocations == nil`, `locationsWithQuotaFor` would effectively just use the locations from `GetResourceSkuLocations`? (assuming `p.subscriptionManager.ListLocations` is a superset of `GetResourceSkuLocations`)

cli/azd/pkg/infra/provisioning/bicep/prompt.go
```diff
@@ -122,14 +133,6 @@ func (a *BicepProvider) locationsWithQuotaFor(
...
 	sharedResults.Range(func(location, quotaDetails any) bool {
 		usages := quotaDetails.([]*armcognitiveservices.Usage)
-		hasS0SkuQuota := slices.ContainsFunc(usages, func(q *armcognitiveservices.Usage) bool {
-			// The minimum quota for the S0 SKU in Microsoft.CognitiveServices/accounts is 2 capacity units
-			return *q.Name.Value == "OpenAI.S0.AccountCount" && (*q.Limit-*q.CurrentValue) >= 2
-		})
-		if !hasS0SkuQuota {
-			// If the S0 SKU quota is not available, skip this location
-			return true
-		}
```
JeffreyCA, line 132: How come we're removing this quota check?

vhvb1989: I thought it was not required any longer... but on second thought, we still need it

cli/azd/pkg/infra/provisioning/bicep/prompt.go
vhvb1989: yes, that is a very good catch.

JeffreyCA: I think we still need to add something like this in `locationsWithQuotaFor`:

```go
	if locations == nil {
		locations = azureAiServicesLocations
	}
```

cli/azd/pkg/infra/provisioning/bicep/prompt.go
```diff
@@ -103,7 +103,23 @@ func (a *BicepProvider) locationsWithQuotaFor(
...
+	}
+
+	if locations == nil {
+		// If no locations are provided, use the Azure AI Services locations
+		locations = azureAiServicesLocations
+	}
+
 	for _, location := range locations {
+		if !slices.Contains(azureAiServicesLocations, location) {
+			// Skip locations that are not in the list of Azure AI Services locations
```
weikanglim, line 120: Is `azureAiServicesLocations` expected to be a reliable source of truth here, when `locations` is provided? Just wondering how the "S0" "Standard" sku factors in here

cli/azd/pkg/infra/provisioning/bicep/prompt.go
```diff
@@ -222,12 +238,25 @@ func (p *BicepProvider) promptForParameter(
 	help, _ := param.Description()
 	azdMetadata, _ := param.AzdMetadata()
 	paramType := p.mapBicepTypeToInterfaceType(param.Type)
+	// parameter `location` is a special case for azd, as it can be set in the environment
+	// and its value must be persisted in the .env file as AZURE_LOCATION
+	isLocationParam := key == "location"
```
weikanglim, line 243: Supposedly, if `location` isn't mapped to `AZURE_LOCATION`, would the `location` variable name be the way we'd go about it?

Just wondering out loud since this may not be immediately possible today and we'd scope it out, but seems more likely to lead to a path of success.

cli/azd/pkg/infra/provisioning/bicep/bicep_provider.go
```diff
@@ -136,34 +136,39 @@ func (p *BicepProvider) EnsureEnv(ctx context.Context) error {
...
 		}
 
-		if _, err := p.ensureParameters(ctx, compileResult.Template); err != nil {
-			return err
+		// Check if location is a parameter and if it is not set in the AZD env
+		// If it is not set, we need to prompt for it.
+		if locParam, hasLocationParam := deploymentParams[environment.LocationEnvVarName]; hasLocationParam {
+			if _, locationInAzdEnv := p.env.Dotenv()[environment.LocationEnvVarName]; !locationInAzdEnv {
+				p.env.SetLocation(locParam.Value.(string))
+				p.envManager.Save(ctx, p.env)
```
weikanglim, line 170: Just curious if we care if `Save()` errors out here

cli/azd/pkg/infra/provisioning/bicep/prompt.go
vhvb1989: I think we can make this better by finding the parameter mapped to `AZURE_LOCATION`
We can do that as a follow up PR

In that world, AZD would handle parameters prompt as locations when they are decorated by location-metadata or when they are mapped to AZURE_LOCATION, regardless of the name.

I don't think there is any case right now, out there, from someone using `location` as parameter w/o mapping it to AZURE_LOCATION

vhvb1989: https://github.com/Azure/azure-dev/issues/5031

rajeshkamal5050 (approved):

weikanglim (approved):

cli/azd/pkg/infra/provisioning/bicep/bicep_provider.go
```diff
@@ -186,7 +188,7 @@ func (p *BicepProvider) EnsureEnv(ctx context.Context) error {
 
 			p.env.DotenvSet(environment.ResourceGroupEnvVarName, rgName)
 			if err := p.envManager.Save(ctx, p.env); err != nil {
-				return fmt.Errorf("saving resource group name: %w", err)
+				return fmt.Errorf("saving location to env: %w", err)
```
JeffreyCA, line 191: Should this be reverted? Seems to be setting the resource group not location

JeffreyCA (approved):

cli/azd/pkg/infra/provisioning/bicep/prompt.go
```diff
@@ -103,7 +103,23 @@ func (a *BicepProvider) locationsWithQuotaFor(
 	ctx context.Context, subId string, locations []string, quotaFor []string) ([]string, error) {
 	var sharedResults sync.Map
 	var wg sync.WaitGroup
+
+	azureAiServicesLocations, err := a.azureClient.GetResourceSkuLocations(
+		ctx, subId, "AIServices", "S0", "Standard", "accounts")
```
wbreza, line 108: Can we add some comments here about why we are hard coding the SKU, tier, etc.  What happens if the user wants "GlobalStandard" instead? Are we supporting that?

cli/azd/pkg/infra/provisioning/bicep/bicep_provider.go
vhvb1989: https://github.com/Azure/azure-dev/pull/5032

wbreza (commented):
Are there any test cases we can add here to verify we're not breaking any existing scenarios?

PR #5067, opened 2025-04-09
Update `armresources` package

JeffreyCA:


rajeshkamal5050 (approved):

cli/azd/test/functional/testdata/recordings/Test_StorageBlobClient.yaml
```diff
@@ -22,9 +22,9 @@ interactions:
             Authorization:
                 - SANITIZED
             User-Agent:
-                - azsdk-go-armsubscriptions/v1.0.0 (go1.24.2; Windows_NT),azdev/0.0.0-dev.0 (Go go1.24.2; windows/amd64)
```
weikanglim, line 25: This all looks good to me. But being slightly curious here, did any request/response change from the library upgrade that required re-recording?

My initial assumption is that with a library upgrade, the request/response should stay the same, and so the recording tests would pass without any recording changes.

JeffreyCA: Some of the functional tests began to fail after the upgrade: https://dev.azure.com/azure-sdk/internal/_build/results?buildId=4742114&view=logs&j=c8fbea3f-f5ea-553e-32cf-eaf013f8156e&t=0406ef3f-de29-5cc3-ca4b-05f62e17b6ba

I couldn't easily tell *what* exactly changed, but the recording diff shows the user agent and the order of some of the query parameters changed:

```diff
- body: '{"nextLink":"https://management.azure.com/subscriptions/faa080af-c1d8-40ad-9cce-e1a450ca5b57/providers/Microsoft.Resources/deployments/?api-version=2021-04-01\u0026%24skiptoken=q1ZKzs8rycwrTSzJzM8Lyc9OzStWsqpWcnUMDgkNNlKyyivNydGBcmG8cFeQJIwX6h3sHxriAeMG%2bPs4%2brk4u%2fqFBDn6wAQhOuDGQbjGENnaWgA%3d","value":[]}'
+ body: '{"nextLink":"https://management.azure.com/subscriptions/faa080af-c1d8-40ad-9cce-e1a450ca5b57/providers/Microsoft.Resources/deployments/?%24skiptoken=XY9Na4NAFEV%2fi7M2oPkoxV3Me0Jo5lnHN1q7C6kpRhmhNYwm%2bN9bG7JIV5fDPVy4V3FoTVeZ876rWsNtXZpvEVwFrlPWqQjMuWlckeNE80e80039Vy6mEVP23ev%2bq6um7ZdyEIHwnWeHuOjlpZgJ989Qrb13%2fnzpqDoKJXzaRL%2bD1MmSIAwVNJB4WSQ1esQhJJxtiD%2bOyqd4l3o2Bv3r6RWd0BJvrTzJPmZp5eC%2fKd3IVHtPslac4WHKKMeGM2gHBbggwFXMxYUgmYnRFWmOgLRBYrXe3T6O4w8%3d\u0026api-version=2021-04-01","value":[]}'
```

There was some discussion in the internal Engineering Chat as well if you want to catch up on it

weikanglim: Thanks for the explanation! I suspect that we're observing some very interesting but harmless details about the API behaving differently -- the API is likely returning more results in a single page than before:

```diff
-        content_length: 855807
+        content_length: 1823541
```

This then results in a change in the `nextLink` generated which **does** affect our recordings.

There isn't a change in the body or headers of the client request, so I suspect this may be acting entirely based on the user agent header change... This is a weird world we're living in, but I'm perfectly okay with it.

weikanglim (approved):

PR #5071, opened 2025-04-10
Migrate to ubuntu 22

danieljurek:
Removed the `CGO_ENABLED: 0` which was previously discussed ... If it's needed it can be added later. 

The result of not setting `CGO_ENABLED: 0` in builds is that Linux binaries won't be compatible with older Linux versions. For example, future releases will not run properly on Ubuntu 20.04 (which has reached the end of LTS) 

Change also migrates a lot of the installer testing infrastructure.


rajeshkamal5050 (approved):

danieljurek:
Overriding check-enforcer. The stuck check is a release build that was not approved.

Issue #5050, opened 2025-04-07
compose: App Service - containerless/runtime-specific support


JeffreyCA:
### Notes

- App Service has [Language-specific settings](https://learn.microsoft.com/azure/app-service/reference-app-settings?tabs=kudu%2Cdotnet#language-specific-settings)
    - Also see: https://github.com/microsoft/Oryx/tree/main/doc/runtimes
    - Python has a special environment variable for [Gunicorn multiple workers support](https://github.com/microsoft/Oryx/blob/main/doc/runtimes/python.md): `PYTHON_ENABLE_GUNICORN_MULTIWORKERS=true`
- May need to expose a way to configure the startup command: https://azureossd.github.io/2022/02/22/Using-PM2-on-App-Service-Linux/
    - Also see: https://learn.microsoft.com/troubleshoot/azure/app-service/faqs-app-service-linux#what-are-the-expected-values-for-the-startup-file-section-when-i-configure-the-runtime-stack- 

### App settings/connection strings
From https://learn.microsoft.com/azure/app-service/configure-common?tabs=portal#configure-connection-strings:

> For ASP.NET and ASP.NET Core developers, setting connection strings in App Service is like setting them in <connectionStrings> in Web.config. The values that you set in App Service override the ones in Web.config. You can keep development settings, such as a database file, in Web.config. You can keep production secrets, such as SQL database credentials, safely in App Service. The same code uses your development settings when you debug locally. It uses your production secrets when you deploy it to Azure.
>
> For other language stacks, it's better to use [app settings](https://learn.microsoft.com/en-us/azure/app-service/configure-common?tabs=portal#configure-app-settings) instead. Connection strings require special formatting in the variable keys to access the values.
>
> There's one case where you might want to use connection strings instead of app settings for non-.NET languages. Certain Azure database types are backed up along with the app only if you configure a connection string for the database in your App Service app. For more information, see [Create a custom backup](https://learn.microsoft.com/en-us/azure/app-service/manage-backup#create-a-custom-backup). If you don't need this automated backup, use app settings.

May be relevant if configuring connection strings: https://github.com/dotnet/runtime/issues/36123

rajeshkamal5050:
Closing this issue as not planned since we are exploring on a simpler, more flexible and scalable(with AI/Agentic) compose v2. Stay tuned for updates, your feedback is always welcome! @kristenwomack @puicchan @tg-msft

PR #5077, opened 2025-04-11
compose: `azd add` support for Azure App Service (Node and Python)

### LINKED ISSUE #5051: compose: App Service support - Python

### LINKED ISSUE #5052: compose: App Service support - Node.js

JeffreyCA:
Closes #5051
Closes #5052
Contributes to #5050

This PR adds `azd add` support for Azure App Service (Linux) as a service host for Node and Python runtimes. **App Service** now appears as an option under the **Host service** group in the `azd add` menu:

image

## azure.yaml schema

```yaml
resources:
  host:
    type: host.appservice
    port: <port>
    runtime:
      stack: <node|python>
      version: <version>
    startupCommand: <custom command>
    uses:
      - <resource>
```

## Runtime stacks

`runtime` is a resource-level property in `azure.yaml`, and is inferred from the language azd detected with `appdetect`. Currently the runtime stack version is hardcoded, but in the future we may want to make it a prompt and dynamically pull available runtimes from the [Get Web App Stacks API](https://learn.microsoft.com/rest/api/appservice/provider/get-web-app-stacks?view=rest-appservice-2024-04-01&tabs=HTTP).

| Detected Language | Runtime stack |
|-------------------|---------------|
| Python            | Python 3.13   |
| JavaScript        | Node 22-lts   |
| TypeScript        | Node 22-lts   |

The following environment variables are set by default in the App Service's app settings:
| Environment Variable | Value |
|-- | -- |
| [`ENABLE_ORYX_BUILD`](https://learn.microsoft.com/en-us/azure/azure-functions/functions-app-settings#enable_oryx_build) | true |
| [`SCM_DO_BUILD_DURING_DEPLOYMENT`](https://learn.microsoft.com/en-us/azure/azure-functions/functions-app-settings#scm_do_build_during_deployment) | true |

For Python only:
|Environment Variable | Value |
|-- | -- |
| [`PYTHON_ENABLE_GUNICORN_MULTIWORKERS`](https://github.com/microsoft/Oryx/blob/main/doc/runtimes/python.md#gunicorn-multiple-workers-support) | true |

These can be overridden in `azure.yaml`:

```yaml
  web2:
    type: host.appservice
    port: 80
    runtime:
      stack: node
      version: 22-lts
    env:
      - name: ENABLE_ORYX_BUILD
        value: 'false'
```

## Database secrets

App Service does not have the same first-party concept of secrets like App Container does, so for databases like MySQL and Postgres, their secrets are currently stored as app settings. In the future we should probably use Key Vault secret references instead.

Key Vault secret references (`@Microsoft.KeyVault(SecretUri=<uri>)`) are used for MongoDB and Redis.

## Unsupported scenarios
1. Simplified init - currently it still defaults to Container App
1. User tries to `azd add` an App Service with a `Dockerfile` present:
    image
1. Static Web Apps (user tries to `azd add` an App Service but azd detects a Web UI framework with static build output):
    image
1. User tries to `azd add` an App Service but azd detects an unsupported language (not Python or JS/TS):
    image
1. User tries to link App Service host to an ACA host (the uses prompts only hides resources of other host types):
    image

## Testing
Deployed sample app [msdocs-python-flask-webapp-quickstart](https://github.com/Azure-Samples/msdocs-python-flask-webapp-quickstart) (removed `infra` and `azure.yaml` first)


cli/azd/pkg/project/scaffold_gen.go
```diff
@@ -370,6 +398,48 @@ func mapContainerApp(res *ResourceConfig, svcSpec *scaffold.ServiceSpec, infraSp
...
+
+	// Set common environment variables for App Service
+	svcSpec.Env["SCM_DO_BUILD_DURING_DEPLOYMENT"] = "true"
+	svcSpec.Env["ENABLE_ORYX_BUILD"] = "true"
+
+	// Set language-specific environment variables
+	switch svcConfig.Language {
+	case ServiceLanguagePython:
+		svcSpec.Env["PYTHON_ENABLE_GUNICORN_MULTIWORKERS"] = "true"
+	}
```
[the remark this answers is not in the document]
rajeshkamal5050: - Always good to have such opinionated properties configurable. Need not be defaulted in the scaffolded azure.yaml. But can be captured in the azure.yaml schema for AI/models to understand.
- Making them configurable should also make sense from AZD's workflow POV. ex: SCM_DO_BUILD_DURING_DEPLOYMENT=false means azd build/package should generate a self-contained bits for app service to run.

puicchan:
If runtime is inferred from the language azd detected with appdetect, I wonder if we can ask less questions. 

Now:
? What would you like to add? Host service
? Which type of host? App Service
? Where is your app code project located? ./   [current directory]
? Enter a name for this service: (msdocs-python-flask-webapp-quickstart) web

the final confirmation message before provision starts: 

For Python: "azd will use python 3.13 to host this project on Azure App Service. Continue?" 
For TypeScript/JavaScript: "azd will use node 22-lts to host this project on Azure App Service. Continue?" 

for `azd init` flow, it's e.g., 

Detected services:

  Python
  Detected in: .

azd will generate the files necessary to host your app on Azure using Azure Container Apps.

Should we do something similar for `azd add` (host) if we are able to detect the lang and source code location? e.g., 

? What would you like to add? Host service
? Which type of host? App Service

Detected:
  Python
  Detected in: . 

azd will use runtime Python 3.13 to host this project on Azure App Service. Continue?

cli/azd/pkg/project/scaffold_gen.go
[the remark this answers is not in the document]
puicchan: Are these needed to run the app locally? If not, my vote is not to exposed under `env:`. These are smart defaults configured in Bicep and I can see them in resources.bicep after running `azd infra synth`.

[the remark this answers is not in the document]
JeffreyCA: I see, makes sense. I pushed a change so users can override these in `azure.yaml` like so, but by default they won't be there:

```yaml
resources:
  host:
    type: host.appservice
    port: 3000
    runtime:
      stack: node
      version: 22-lts
    env:
      - name: ENABLE_ORYX_BUILD
        value: 'false'
```

> Are these needed to run the app locally?
No, only for deployment

rajeshkamal5050 (approved):
LGTM.

schemas/alpha/azure.yaml.json
```diff
@@ -1523,6 +1526,79 @@
...
+                "type": true,
+                "uses": true,
+                "port": {
+                    "type": "integer",
+                    "title": "Port that the web app listens on",
+                    "description": "Optional. The port that the web app listens on. (Default: 80)"
+                },
+                "env": {
+                    "type": "array",
+                    "title": "Environment variables to set for the container app",
```
weikanglim, line 1548:
```suggestion
                    "title": "Environment variables to set for the web app",
```

cli/azd/internal/cmd/add/add_configure_host.go
```diff
@@ -231,50 +264,70 @@ func addServiceAsResource(
...
 		if _, err := os.Stat(filepath.Join(svc.RelativePath, "Dockerfile")); errors.Is(err, os.ErrNotExist) {
 			// default builder always specifies port 80
-			props.Port = 80
+			port = 80
 			if svc.Language == project.ServiceLanguageJava || svc.Language.IsDotNet() {
-				props.Port = 8080
+				port = 8080
 			}
 		}
 	}
```
JeffreyCA, line 287: @weikanglim @puicchan any thoughts on whether we should prompt for the port number in app service case? Right now it just defaults to 80, but can be changed in `azure.yaml`

weikanglim: Defaulting and exposing it is fine.

cli/azd/resources/scaffold/templates/resources.bicept
```diff
@@ -738,6 +755,131 @@ module {{bicepName .Name}} 'br/public:avm/res/app/container-app:0.8.0' = {
...
+          {{- end}}
+        ]
+      }
+    }
+    clientAffinityEnabled: false
+    httpsOnly: true
+    appSettingsKeyValuePairs: {
+      AZURE_CLIENT_ID: {{bicepName .Name}}Identity.outputs.clientId
+      {{- if .DbCosmosMongo}}
+      'MONGODB_URL': '@Microsoft.KeyVault(SecretUri=${cosmosMongo.outputs.exportedSecrets['mongodb-url'].secretUri})'
```
weikanglim, line 794:
```suggestion
      MONGODB_URL: '@Microsoft.KeyVault(SecretUri=${cosmosMongo.outputs.exportedSecrets['mongodb-url'].secretUri})'
```

Does this make any difference?

weikanglim (approved):
@JeffreyCA Great work here!  This looks really good! Thanks for taking the extra care to landing this cleanly!

cli/azd/resources/scaffold/templates/resources.bicept
JeffreyCA: Good catch, thanks

weikanglim (approved):

Issue #4755, opened 2025-02-03
Feature request: Support multiple infrastructure deployments in AZD provision cmd.


HadwaAbdelhalem:
AZD provision cmd deploys infrastructure resources in a single deployment. For some use cases, the resources can be created in separate multiple deployments for various reasons, like different resource lifecycles, different personas have access/ownership to certain resource type deployments.

Proposal:
update the azd provision cmd to specify deployment.

```
# Folder Structure

- Infra
 ├──network
    - main.tf OR main Bicep
 ├── database
   - main.tf OR main Bicep
 ├── Apps
   - main.tf OR main Bicep

```
# azure.yaml
proposal adds an optional deployment property to the infra section of the azure.yaml schema.
 -  Deployments is list of objects to list [deployment name, relative path to the deployment folder from ./infra , and order for the deployment]. 
 - Deployments property cannot be provided if module property is provided.
 - Deployment property aims to control/support the multi deployment approach, while module property supports the single deployment style.

```yaml
infra:
  provider: terraform
  module: ./infra
  path: ./infra
  deployments: [
    {
      name: "network",
      path: "./network",
      order: 1
    },
    {
      name: "database",
      path: "./database",
      order: 2
    },
    {
      name: "applications",
      path: "./app",
      order: 3
    }
  ]
```
# azd cmds
```cmd
azd provision --deployment <deploymentName> 

Deploy specific group of resources in a single deployment.

azd provision --all

Deploy all resources in all deployments/folders in sequence based on the order field within the deployments array.
```

ljtill:
It would also be useful to have an option to specify whether the deployment should use Deployment Stacks (DS) or not. For example, we may want to disable DS when deploying infra/components that depend on Graph resources in Bicep, since DS currently does not support Graph resources.

wbreza:
A couple questions and recommendations.

**Q:**: Do we need the `order` property since the order should already be preserved within the array during yaml unmarshalling.

**Q:**: Can we follow a similar pattern to `azd deploy [service-name]` I would propose we opt for ordered argument instead of `--deployment` flag. Ex) `azd provision network` instead of `azd provision --deployment network`.

**Q:** I would also expect `--all` to be implicit when a deployment name is not specified.

**Q:** What is the plan for input parameters? Would each module be expected to have a corresponding `main.parameters.json` or similar that would be used?

[...]

PR #5088, opened 2025-04-16
compose: with existing infra

### LINKED ISSUE #4578: Compose: Iac Mode support for brownfield projects

weikanglim:
This change allows for composability to work with an existing `infra` directory unmanaged by azd.

**Infra provisioning.** The provisioning engine has been updated to serially provision the root `./infra`, then `./infra/app` if it exists. The following behavioral changes have been applied:

1. **Deployment order.** Sequential; the progress display shows the root deployment URL, then displays resources created as each deployment runs in the background.
2. **Outputs.** Outputs are written to the environment in the deployment order. This means in cases of output collisions, the last deployment wins.
3. **Preview.** Merge deployment resources into a single preview list, deduplicate resource ID + type.
4. **Destroy.** Merge deployment resources for display, and for purge. For destroy, each deployment is executed sequentially. This works for both standard and stack deployments.
5. **State.** The deployments are fetched and applied to the environment in the same order as the deployment order. This shares the same last-write-win semantics.

**Composability synth.** The synthesized infra directory has a `.azd` file to serve as a sentinel. If `.azd` exists in a directory, i.e. `./infra/foo`, it means the directory `./infra/foo` is "azd managed". This used to achieve the following behavior when running `infra synth`:
- If `./infra/.azd` exists, we overwrite `./infra`.
- Otherwise, we write to `./infra/app`.

## UX

Provision:
image

Fixes #4578, Contributes slightly to #4755


cli/azd/pkg/infra/provisioning/bicep/bicep_provider.go
```diff
@@ -204,8 +224,8 @@ func defaultPromptValue(locationParam azure.ArmTemplateParameterDefinition) *str
 	return nil
 }
 
-func (p *BicepProvider) LastDeployment(ctx context.Context) (*azapi.ResourceDeployment, error) {
-	modulePath := p.modulePath()
+func (p *BicepProvider) LastRootDeployment(ctx context.Context) (*azapi.ResourceDeployment, error) {
```
wbreza, line 227: I feel like this bicep provider implementation has been getting unwieldy for a while.  So many different cases and edge cases happening in here now and might be time for a general refactor.

Some idea:
- Have a 1:1 between calls to the bicep provider and bicep deployment
- Perhaps composability is a new infra provider that then consumes the bicep provider after the bicep is generated.

weikanglim: Agreed that  a refactor for bicepModule is long overdue. I plan to tackle this after landing the changes, since it will help inform the shape of the contracts between bicepProvider and bicepModule.

From the changes that are being applied here, there isn't behavior specific to composability other than the new `appModules()` method. Composability plays within the rules of Bicep provider, and not the other way around.

weikanglim: I suspect that for this particular change of `LastDeployment` -> `LastRootDeployment`,  we expect the "DeploymentService" (which also includes VS) to be passing a modulePath as the parameter -- which is an API breaking change, but it's kept as `LastRootDeployment` to maintain backwards compatibility and limit scope. I can write a follow-up issue to delegate that responsibility downwards.

rajeshkamal5050 (approved):

PR #5110, opened 2025-04-22
[extension feature branch] Get build working for direct compile scenarios

danieljurek:
Example build: https://dev.azure.com/azure-sdk/internal/_build/results?buildId=4785825&view=results


eng/pipelines/templates/jobs/build-azd-extension.yml
```diff
@@ -34,87 +34,36 @@ jobs:
...
+          Write-Host "##vso[task.setvariable variable=EXT_VERSION;]$extVersion"
+        workingDirectory: ${{ parameters.AzdExtensionDirectory }}
+        displayName: Set EXT_VERSION
 
       - task: PowerShell@2
         inputs:
           pwsh: true
           targetType: filePath
-          filePath: cli/azd/extensions/scripts/ci-build.ps1
+          filePath: ${{ parameters.AzdExtensionDirectory }}/ci-build.ps1
```
wbreza, line 53: My original intent was to have a shared `ci-build.ps1` that all extensions could use instead of each extension having a copy of the build script, thoughts?

Ultimately, we may have to have different flavors of this per programming language but wanted to reduce the cost of ownership of extension developers.

wbreza (approved):

eng/pipelines/templates/jobs/build-azd-extension.yml
danieljurek: Chatted offline: If there is shared code it should be in the `scripts/` folder and referenced from the `<extension>/ci-build.ps1` script. This makes it so pipelines can operate entirely in the extension's folder. 

However, things should be simple enough today that `ci-build.ps1` would be trivially simple for all scenarios.

PR #5115, opened 2025-04-23
CGO_ENABLED: 0 for Linux x64 build

danieljurek:


rajeshkamal5050 (approved):

weikanglim:
LGTM. This is perfectly fine and helps azd being compatible with a large variety of *nix distributions without necessarily having to mirror the distributions' glibc setup in our builds.

Users are unlikely going to see a difference in behavior. Notably, popular golang-based CLIs such as Terraform also set `CGO_ENABLED=0` by default.

PR #5117, opened 2025-04-23
fix tags bug after breaking change in awesome-azd

### LINKED ISSUE #5107: Fix template list with filter after Awesome AZD source breaking change
For awesome-azd, tags are now split into 3 fields: tags, language and azure_service.
AZD must re-construct a `tags` field with the union of the 3 fields to work as it used to.

hemarina:
fix #5107

image
image

### DIFF
--- a/cli/azd/pkg/templates/awesome_source.go
+++ b/cli/azd/pkg/templates/awesome_source.go
@@ -17,10 +17,12 @@ import (
 )
 
 type awesomeAzdTemplate struct {
-	Title       string   `json:"title"`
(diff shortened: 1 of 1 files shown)


wbreza (approved):

rajeshkamal5050:
Thanks @hemarina for the fix. 

We had the generic tags notion. Now its broken down into generic and some specific ones. Do you think general `tags` notion might get broken down further in future?

rajeshkamal5050 (approved):

PR #5121, opened 2025-04-24
Support scaffolding for JavaScript based extensions.

zedy-wj:
1. Fix https://github.com/Azure/azure-dev/issues/5092. This PR implements the `azd x init` command based on JavaScript extension.

The scaffolding initialized by `azd x init` already includes support for the `azd x build` and `azd x package` commands, and also supports the `azd x watch` command.

>Note: For the core command functions in the JavaScript extension (such as context, prompt, listen.), it is still under development.

2. Move the `proto` folder out of `dotnet` and place it at the same level as the `languages` directory, making it a shared resource.

## Create a JavaScript extension
Run `azd x init`

This will create a new JavaScript based extension from a template, create a local `azd` extension source, build, package and install the extension for use in local development.
image

## Build your extension
Run `azd x build` from within your extension folder.
image

## Package your extension
Run `azd x package` from within your extension folder
image

This will package a new version and publish it to your local extensions source.

## Iterate on your extension
Run `azd x watch` to start watching for local changes. Changes are automatically detected and a new version of your extension will be rebuilt and installed locally.
image

@rajeshkamal5050 and @wbreza for notification.


zedy-wj:
Supplement about `azd X context`, `listen`, `prompt` commands:

## azd X context
image

## azd X listen
image

**Notes:** Currently, this function is limited to sending simulated events for testing purposes.
Upon investigation, we observed that in the .NET environment, lifecycle events (such as preprovision) are triggered by the azd CLI main process via gRPC and directly propagated to the extension’s existing event stream.

However, in the JavaScript environment, the azd CLI does not handle event propagation in the same manner.

We attempted to manually emit lifecycle events using a lifecycleRunner utility, but this approach required establishing a new gRPC connection, which is isolated from the listener’s active event stream.

Consequently, even though the event was successfully pushed, the listener was unable to receive the corresponding Invoke message due to the separation between the gRPC streams.

## azd X prompt
image

@rajeshkamal5050 and @wbreza please review this PR, thanks.

v-xuto:
@wbreza The PR for scaffolding regarding JavaScript is ready, please review.

wbreza:
> @wbreza The PR for scaffolding regarding JavaScript is ready, please review. Thanks.

I'm getting the following error when attempt to run either the `context` or `prompt` demo commands.

```
Error: ENOENT: no such file or directory, open 'D:/dev/azure-samples/todo-nodejs-mongo-aca/proto/compose.proto'
```

From looking at the code it looks like its using `process.cwd()` to get the path to the proto instead of using the embedded `pkg` assets?

Regarding the lifecycle events - not sure i understand the issue.  It's all gRPC.  

1. azd cli calls `listen` command on extension
2. Extension should register callbacks (and keep the stream open, don't exit function)
3. azd then sends gRPC events back to client who then invokes the event handler.

cli/azd/extensions/microsoft.azd.extensions/internal/resources/languages/javascript/azdClient.js
```diff
@@ -0,0 +1,57 @@
...
+
+    // Load each proto service
+    const composeProto = loadService('compose');
+    const deploymentProto = loadService('deployment');
+    const environmentProto = loadService('environment');
+    const eventProto = loadService('event');
+    const projectProto = loadService('project');
+    const promptProto = loadService('prompt');
+    const userConfigProto = loadService('user_config');
+    const workflowProto = loadService('workflow');
```
wbreza, line 44: I was playing around with some of the issue here with they dynamic gRPC client generation from the proto files and the way assets are packaged with the `exe`.  

Are there other libraries that can be used that brings us closer to statically generated clients similar to the other languages? Maybe that would be a better/easier experience?

SophCarp:
Trying out this PR: 
image
I ran `azd extension lit` rather than `azd extension list`, and didn't get an error. I just got summary of `azd extension`. Is that standard? I thought I'd get an error that `lit` isn't a known command. 

Secondly, what is the source for the "x" extension? I think I need to sun `azd extension source add [source]` but I don't see what that source should be.

wbreza:
> Trying out this PR: I ran `azd extension lit` rather than `azd extension list`, and didn't get an error. I just got summary of `azd extension`. Is that standard? I thought I'd get an error that `lit` isn't a known command.
> 
> Secondly, what is the source for the "x" extension? I think I need to sun `azd extension source add [source]` but I don't see what that source should be.

Since the `extensions` is a command group, if you misspell a command you get the `help` for the command group similar to running `azd extension -h`

[Docs](https://github.com/Azure/azure-dev/blob/main/cli/azd/docs/extension-framework.md) have recently been updated. The `x` extension is available in our `dev` extension source.

```bash
# Add a new extension source name 'dev' to your `azd` configuration.
azd extension source add -n dev -t url -l "https://aka.ms/azd/extensions/registry/dev"
```

cli/azd/extensions/microsoft.azd.extensions/internal/resources/languages/javascript/azdClient.js
zedy-wj: We are currently revising this section, but it will take some time. If this PR is urgent, feel free to merge it first, and we will follow up with another PR to track the changes later.

wbreza (approved):
Looks good - there are some issue with the event manager implementation but I can fix them after merged.

PR #5132, opened 2025-04-29
Extensions: Pass console width to extensions when invoked

wbreza:
When a command is invoked through a sub process/shell, the called extension is unable to determine the correct console width using the `consolesize.GetConsoleSize()` that is used within `azd`.

To work around this issue, the currently detected console width will be set in a `COLUMNS` environment variable and then read by the extensions if unable to detect the size using standard libraries.

The `COLUMNS` environment variable is a semi-standard used in some Mac/unix based terminals.  If a console width is still unable to be detected we fallback to a default of `120`.

### DIFF
--- a/cli/azd/cmd/extensions.go
+++ b/cli/azd/cmd/extensions.go
@@ -18,6 +18,7 @@ import (
 	"github.com/azure/azure-dev/cli/azd/pkg/input"
 	"github.com/azure/azure-dev/cli/azd/pkg/ioc"
 	"github.com/azure/azure-dev/cli/azd/pkg/lazy"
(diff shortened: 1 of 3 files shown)


rajeshkamal5050 (approved):

weikanglim:
Just wondering, does it matter at all that the console width could change if the user were to resize the window?

wbreza:
> Just wondering, does it matter at all that the console width could change if the user were to resize the window?

Ideally we would want to handle that but we don't have any mechanism in place to detect and update that values from azd core and the extension at the moment.  This is a good first step in my opinion that we can improve over time.

vhvb1989 (approved):

PR #5143, opened 2025-05-01
Update pipeline config - provider parameters

vhvb1989:
This change introduces a feature for pipeline config called `provider parameters`.
This feature allows infrastructure providers like bicep, to generate a list of parameters required for provisioning the infrastructure definition.
Then, `pipeline config` gets this information and uses it to generate the pipeline definition and to automatically know what variables and secrets must be set for running provision on CI/CD. 
This feature enhance the current approach that requires people to manually define the required variables and secrets for the pipeline in azure.yaml. If the infrastructure provider can infer this information from the IaC, the user will no longer need to manually worry about creating the configuration.

For the bicep-provider, for example, the required variables and secrets for CI are inferred from the parameters file.

By using this strategy, AZD no longer requires the `AZD_INITIAL_ENVIRONMENT_CONFIG` to smuggle local prompts to CI. As long as there is one ENV VAR mapping to a parameter, AZD will know what variable or secret to create in CI and the value for it.

This change also improves the experience for Aspire, where the IaC is automatically generated and a mapping to ENV VARS is always generated.

Note:  All previous configuration will still be supported, but after this change, AZD will use the new strategy.


cli/azd/pkg/infra/provisioning/bicep/bicep_provider.go
```diff
@@ -2293,3 +2291,48 @@ func NewBicepProvider(
...
+			Name:          key,
+			Secret:        param.Secure(),
+			Value:         resolvedParams[key].Value,
+			EnvVarMapping: parametersInfo.envMapping[key],
+			// No env var mapping and param is persisted in env config infra.parameters means local prompt only
+			// If user set an env var mapping after a local prompt, the env var overrides the value persisted in config
+			// which turns local prompt false
+			LocalPrompt: isPrompt && len(parametersInfo.envMapping[key]) == 0,
+		})
+	}
```
wbreza, line 2335: Looks like the length of the slice may not be 100% deterministic based on the first useParam short circuit.  Consider just creating an unbounded slice and append.

cli/azd/pkg/pipeline/pipeline.go
```diff
@@ -136,13 +142,74 @@ type CiProvider interface {
...
-// The initial values reference azd known values, which are merged with the ones defined on azure.yaml by the user.
+// The initial values reference azd known values, which are merged with the ones defined on azure.yaml by the user and the
+// provider parameters.
 func mergeProjectVariablesAndSecrets(
 	projectVariables, projectSecrets []string,
-	initialVariables, initialSecrets, env map[string]string) (variables, secrets map[string]string) {
+	initialVariables, initialSecrets map[string]string,
+	providerParameters []provisioning.Parameter,
+	env map[string]string) (variables, secrets map[string]string, err error) {
+
```
wbreza, line 152: Consider adding some unit tests for this.  Lots of logic branches here.

cli/azd/pkg/pipeline/pipeline_manager.go
```diff
@@ -877,11 +826,11 @@ func (pm *PipelineManager) initialize(ctx context.Context, override string) erro
...
-	pm.configOptions = &configurePipelineOptions{
-		projectVariables:     slices.Clone(prjConfig.Pipeline.Variables),
-		projectSecrets:       slices.Clone(prjConfig.Pipeline.Secrets),
-		provisioningProvider: &pm.infra.Options,
-	}
+	// pm.configOptions = &configurePipelineOptions{
+	// 	projectVariables:     slices.Clone(prjConfig.Pipeline.Variables),
+	// 	projectSecrets:       slices.Clone(prjConfig.Pipeline.Secrets),
+	// 	provisioningProvider: &pm.infra.Options,
+	// }
```
wbreza, line 833: Nit: remove if no longer needed.

cli/azd/pkg/pipeline/pipeline_manager_test.go
```diff
@@ -339,6 +394,13 @@ func Test_PipelineManager_Initialize(t *testing.T) {
 		assert.NotNil(t, manager)
 		assert.NoError(t, err)
 
+		manager.infra = &project.Infra{
+			Options: provisioning.Options{
+				Provider: provisioning.Bicep,
+			},
+			IsCompose: false}
+		manager.configOptions = &configurePipelineOptions{}
```
wbreza, line 402: nit: Can we omit setting these empty options and have the manager handle this internally setting to empty options?

cli/azd/pkg/infra/provisioning/bicep/bicep_provider.go
```diff
@@ -2027,24 +2040,9 @@ func (p *BicepProvider) ensureParameters(
...
-					}
-				}
 				configuredParameters[key] = azure.ArmParameter{
 					Value: paramValue,
 				}
-				if needForDeployParameter {
-					mustSetParamAsConfig(key, paramValue, p.env.Config, param.Secure())
-					configModified = true
-				}
 				continue
```
wbreza, line 2046: OK to remove this and still have backwards compatibility?

wbreza (approved):
Looks good - but i would recommend adding some unit tests to validate some of the new behavior.  I see we have the e2e snapshot tests but a few more unit tests might enable us to more quickly iterate on future changes.

davidfowl:
Once this is done, we need to manage state in the CI pipeline (github, ado) and we need to make it such that when you add new parameters that they don't overwrite old ones.

rajeshkamal5050 (approved):

vhvb1989:
overriding checks for todo-templates b/c it has a different issue. This PR does not affect todo-templates direct functionality and its pipeline functionality is not tested in workflows

Kumima:
@vhvb1989 , what's the env var name pattern?
builder.AddParameter("api-secret", true). What's the env var name for this secret parameter?

vhvb1989:
> api-secret

That would be `AZURE_API_SECRET`
The pattern is `AZURE_{ TO_SNAKE_UPPER_CASE( parameter_name ) }`

api-secret ->  to snake upper case -> API_SECRET

Yes. You would get the same by having `apiSecret` as the parameter name. (name collision).

Fazer01:
@vhvb1989 

Where is this documented? Also, should we use the `azd env set` method to set these 'ENV VARS' before the azd provision command in the pipeline or also use the `env:` in the task to define the parameter? 

```yaml
azd env set AZURE_SOME_KEY $(VALUE_FROM_PIPELINE_LIBRARY)
azd env get-values   
azd provision --no-prompt
```
```yaml
 - task: AzureCLI@2
    displayName: Provision Infrastructure
    inputs:
      azureSubscription: ${{ variables.AzureServiceConnection }}
      scriptType: bash
      scriptLocation: inlineScript
      inlineScript: |
        dotnet ef --version
        cd $(Build.SourcesDirectory)/src/RegimeService.AppHost 
        azd env new $(AZURE_ENV_NAME)
        azd env set AZURE_SOME_KEY $(VALUE_FROM_PIPELINE_LIBRARY)      
        azd env get-values   
        azd provision --no-prompt
    env:
      AZURE_SUBSCRIPTION_ID: $(AZURE_SUBSCRIPTION_ID)
      AZURE_ENV_NAME: $(AZURE_ENV_NAME)
      AZURE_LOCATION: 'westeurope'     
      AZURE_SOME_KEY: $(VALUE_FROM_PIPELINE_LIBRARY)         
   
```

I'm awaiting your reply. 
Thanks in advance. 

Tom

Fazer01:
Alright! Thanks for the extensive response @vhvb1989! We haven't upgraded to .NET Aspire 9.3 yet, so that's what I'm going to do first. We have created our own pipeline definition, not using `azd pipeline config`.
We use a pipeline library (variable group) to set these variables, does that perhaps make a difference?

PR #5153, opened 2025-05-02
Developer Extension Refactors usage of `gh` cli and adds better error messages

wbreza:
- Fixes issues when specifying `--cwd` flag on commands
- Improved usage of `gh` cli
- Better error messages on `gh` errors

### DIFF
--- a/cli/azd/extensions/microsoft.azd.extensions/internal/github/github.go
+++ b/cli/azd/extensions/microsoft.azd.extensions/internal/github/github.go
@@ -1,0 +1,246 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package github
+
+import (
+	"encoding/json"
+	"fmt"
+	"os/exec"
+	"runtime"
+	"strings"
+
+	"github.com/azure/azure-dev/cli/azd/extensions/microsoft.azd.extensions/internal"
+	"github.com/azure/azure-dev/cli/azd/pkg/common"
+)
+
+var (
+	ErrReleaseAlreadyExists = fmt.Errorf("release already exists")
+	ErrReleaseNotFound      = fmt.Errorf("release not found")
+)
+
+// GitHubCli provides access to GitHub CLI functionality
+type GitHubCli struct {
+	// ExecutablePath is the path to the GitHub CLI executable
+	ExecutablePath string
+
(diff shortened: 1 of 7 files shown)


cli/azd/extensions/microsoft.azd.extensions/internal/github/github.go
```diff
@@ -0,0 +1,246 @@
...
+
+import (
+	"encoding/json"
+	"fmt"
+	"os/exec"
+	"runtime"
+	"strings"
+
+	"github.com/azure/azure-dev/cli/azd/extensions/microsoft.azd.extensions/internal"
+	"github.com/azure/azure-dev/cli/azd/pkg/common"
```
vhvb1989, line 14: Is it expected that extensions can import from `/azd/pkg` ?  (out of `/azd/extensions`) ??   Seems like a shortcut to bypass AZD cli to get AZD's functionality. Something possible when using GO, but maybe something we might want to avoid?

cli/azd/extensions/microsoft.azd.extensions/internal/github/github.go
```diff
@@ -0,0 +1,246 @@
...
+)
+
+var (
+	ErrReleaseAlreadyExists = fmt.Errorf("release already exists")
+	ErrReleaseNotFound      = fmt.Errorf("release not found")
+)
+
+// GitHubCli provides access to GitHub CLI functionality
+type GitHubCli struct {
```
vhvb1989, line 23: Why are we creating a copy of this tool instead of leveraging how AZD can handle this gh cli?  
AZD knows how to auto-install, login, and more operations

cli/azd/extensions/microsoft.azd.extensions/internal/github/github.go
wbreza: Yeah - i was torn about this.  Yes, go extensions can use any packages in `azd` if we want - I was just considering whether this was good or bad practice.  I had created the `azdext` as a start of exposing stuff for extensions.

Happy to just add the new commands if we are OK with this approach.

cli/azd/extensions/microsoft.azd.extensions/internal/github/github.go
wbreza: See comments above.

vhvb1989 (approved):

PR #5158, opened 2025-05-06
compose - configure Foundry Hub to use identity-based storage access

JeffreyCA:
Contributes to #4879.

In certain Azure tenants with stricter policies, deploying an AI project through composability can fail because the storage account created for the Foundry Hub has anonymous blob access and account key access enabled.

This PR disables anonymous blob access and account key access, and configures the Foundry Hub to use Identity-based access to access the storage account.

It also sets the appropriate role assignments for the user, **Storage Blob Data Contributor** and **Storage File Data Privileged Contributor** ([documentation](https://aka.ms/ai-hub-storage-account-role-assignments)).

### DIFF
--- a/cli/azd/resources/scaffold/templates/ai-project.bicept
+++ b/cli/azd/resources/scaffold/templates/ai-project.bicept
@@ -17,6 +17,9 @@ var resourceToken = uniqueString(subscription().id, resourceGroup().id, location
 @description('The name of the environment')
 param envName string
 
+@description('Id of the user or app to assign application roles')
+param principalId string
+
 module keyVault 'br/public:avm/res/key-vault/vault:0.6.1' = {
   name: 'keyvaultForHub'
   params: {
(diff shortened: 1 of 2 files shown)


cli/azd/resources/scaffold/templates/ai-project.bicept
```diff
@@ -125,6 +136,7 @@ resource hub 'Microsoft.MachineLearningServices/workspaces@2024-10-01' = {
     }
     v1LegacyMode: false
     publicNetworkAccess: 'Enabled'
+    systemDatastoresAuthMode: 'Identity'
```
JeffreyCA, line 139: By setting this, the Storage Blob Data Contributor and Storage File Data Privileged Contributor role assignments for the hub's system-assigned identity are automatically created during deployment:

image

cli/azd/resources/scaffold/templates/ai-project.bicept
```diff
@@ -83,11 +98,7 @@ resource {{bicepName .Name}}{{bicepName .Version}}Deploy 'Microsoft.CognitiveSer
     name: 'S0'
   }
   kind: 'AIServices'
-  properties: {
-    apiProperties: {
-      statisticsEnabled: false
-    }
-  }
+  properties: {}
```
JeffreyCA, line 101: `statisticsEnabled` is only used for Bing Search and it seems the Cognitive Services RP now has stricter validation on this, causing create requests to fail when this is set.

image

Issue #5151, opened 2025-05-02
Deployment with bicep templates fails with array as a parameter in main.parameters.json


KseniyaMikhailiuk:
- [x] Make sure you've installed the latest version using [instructions in the wiki](../wiki/install)

**Output from `azd version`**
1.15.0 (2025-05-01)

**Describe the bug**
`azd provision --no-prompt`
Provisioning with bicep template + .parameters.json fails since 01.05.2025 while it worked the day before.
The error in logs:
`ERROR: initializing provisioning manager: prompting for value: no default response for prompt 'Enter a value for the '{{paramName}}' infrastructure parameter:'`
Parameter is an array, string parameter does not produce an error

**To Reproduce**
Run `azd provision --no-prompt` with any bicep template which expects array as a parameter from  .parameter.json file.

**Expected behavior**
Array parameter is populated successfully, no exception is thrown.

**Additional context**
[Potential breaking change](https://github.com/Azure/azure-dev/pull/5061/files#diff-2c22edb40a26b0cfa31df98cdaf6a05bb96719d424366e53c088792e50714754R1620-R1631)
Can it be that on line 1621 we lost all parameters which are not string: never saved them in resolvedParams (line 1631)?

Mstiekema:
I am encountering the same issue as of today.

albert7205:
Same issue observed over here, a fix would be welcome.

sethsteenken:
Also having same issue on 1.15.0 with main.parameters.json. I reverted to 1.14.0 and parameters worked as expected.

Mstiekema:
> Additional context
[Potential breaking change](https://github.com/Azure/azure-dev/pull/5061/files#diff-2c22edb40a26b0cfa31df98cdaf6a05bb96719d424366e53c088792e50714754R1620-R1631)
Can it be that on line 1621 we lost all parameters which are not string: never saved them in resolvedParams (line 1631)?

I have checked this by building `azd` locally and have confirmed that this is the issue. I am working on a fix and will submit it as a PR later.

rajeshkamal5050:
cc @vhvb1989 for reviewing.

PR #5162, opened 2025-05-06
Properly reading objects and arrays from parameter file

Mstiekema:
This PR aims to solve #5151.

I have implemented additional checks for the type of a parameter that is read from the `main.parameters.json` file. I have defined the following restrictions. Please let me know if you all agree with these or whether I should change it up!

- `string`  is valid when it's not `null` (this was already implemented).
- `array`  is always valid. So this means that `[]` and `[...]` is supported.
- `object` is only valid when at least one element is defined. This means that `{}` is invalid and any form of `{key:value}` will be accepted.

Looking forward to any reviews/comments on my first contribution to this repository.

### DIFF
--- a/cli/azd/pkg/infra/provisioning/bicep/bicep_provider.go
+++ b/cli/azd/pkg/infra/provisioning/bicep/bicep_provider.go
@@ -1617,8 +1617,17 @@ func (p *BicepProvider) loadParameters(ctx context.Context) (loadParametersResul
 			resolvedParams[paramName] = resolvedParam
 			continue
 		}
+
+		// Check if the parameter has a valid type
 		stringValue, isString := resolvedParam.Value.(string)
-		if !isString {
+		mapValue, isMap := resolvedParam.Value.(map[string]any)
+		_, isArray := resolvedParam.Value.([]any)
+		if !isString && !isMap && !isArray {
+			continue
+		}
+
+		// Ensure that an object is not empty
+		if isMap && len(mapValue) == 0 {
 			continue
 		}
 
--- a/cli/azd/pkg/infra/provisioning/bicep/bicep_provider.go
+++ b/cli/azd/pkg/infra/provisioning/bicep/bicep_provider.go
@@ -1618,24 +1618,15 @@ func (p *BicepProvider) loadParameters(ctx context.Context) (loadParametersResul
 			continue
 		}
 
-		// Check if the parameter has a valid type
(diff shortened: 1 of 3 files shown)


Mstiekema:

> @Mstiekema please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> @microsoft-github-policy-service agree [company="{your company}"]
> ```
> 
> > Options:
> > 
> > * (default - no company specified) I have sole ownership of intellectual property rights to my Submissions and I am not making Submissions in the course of work for my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > * (when company given) I am making Submissions in the course of work for my employer (or my employer has intellectual property rights in my Submissions by contract or applicable law). I have permission from my employer to make Submissions and enter into this Agreement on behalf of my employer. By signing below, the defined term “You” includes me and my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree company="Microsoft"
> > ```
> 
> Contributor License Agreement

cli/azd/pkg/infra/provisioning/bicep/bicep_provider.go
```diff
@@ -1617,8 +1617,17 @@ func (p *BicepProvider) loadParameters(ctx context.Context) (loadParametersResul
 			resolvedParams[paramName] = resolvedParam
 			continue
 		}
+
+		// Check if the parameter has a valid type
 		stringValue, isString := resolvedParam.Value.(string)
-		if !isString {
+		mapValue, isMap := resolvedParam.Value.(map[string]any)
+		_, isArray := resolvedParam.Value.([]any)
+		if !isString && !isMap && !isArray {
```
vhvb1989, line 1625: thank you for your suggestion @Mstiekema.  It is a good one, but we need something different.

The intention of checking if the parameter's value type is string is to handle a very unique case for string parameters where we don't know if the value is empty b/c it s mapped to an ENV VAR that is not defined or b/c it is mapped to an ENV VAR that is defined with an empty string.

We are trying just to ignore the first case when we were setting an empty string after a missing ENV VAR.

I made a mistake by ignoring any non-string parameter here. We only want to ignore string parameters with empty value, non-nil and hasUnsetEnvVar = true.

vhvb1989 (approved):

rajeshkamal5050 (approved):

PR #5176, opened 2025-05-08
compose - `azd add` fit and finish

### LINKED ISSUE #4877: [AI compose] enhance spinner while checking for quota
There is UX improvements to be done for the check-quota and list AI models.
The change might require touching the list-locations implementation

### LINKED ISSUE #5184: UX Review azd add service app
Review the UX of adding Service App

Image
- Container app is still what we mostly work with, so it should be at the top. I think it should have an indicator that it's the default. 
- My proposal: have container app at the top of the list, and label it `container app (default)`

Image
- What are the supported languages for App service so far?

I went to look up the limitations and found: 
Image
- https://learn.microsoft.com/en-us/azure/developer/azure-developer-cli/azd-compose
- We should use the same groupings in the add flow as here. Also, if there are limitations, we should document them, either here or somewhere else. 

Image
- `azure app services` doesn't need to be magenta, it should be the blue we use to highlight. 

Otherwise, this is good!

JeffreyCA:
Resolves #4877
Resolves #5184

This PR contains some fixes and enhancements in the `azd add` flow:

### 1. Prevent same resource being added under 'uses' multiple times
When adding multiple AI Services/Foundry models, if a service host already 'uses' the AI project, azd will not show that service when prompting for 'uses'.

image

### 2. Fix `azd add` preview for MySQL and Postgres
Previously, the `azd add` preview was not showing the resource type or binding variables.

image

### 3. Show spinner while loading available Foundry models
After the "Retrieving locations..." spinner:

https://github.com/user-attachments/assets/52675ea7-e93d-4373-bd68-67c39647af74

### 4. App Service UX review changes

Host types are displayed in a custom sorted order, with **Container App (default)** as the first option.
image

The text **Azure App Service** is now highlighted in the standard bright blue colour.
image

### DIFF
--- a/cli/azd/internal/cmd/add/add.go
+++ b/cli/azd/internal/cmd/add/add.go
@@ -197,6 +197,9 @@ func (a *AddAction) Run(ctx context.Context) (*actions.ActionResult, error) {
 	}
 
 	for _, svc := range usedBy {
+		if slices.Contains(prjConfig.Resources[svc].Uses, resourceToAdd.Name) {
+			continue
+		}
 		err = yamlnode.Append(&doc, fmt.Sprintf("resources.%s.uses[]?", svc), &yaml.Node{
 			Kind:  yaml.ScalarNode,
 			Value: resourceToAdd.Name,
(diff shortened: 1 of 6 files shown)


weikanglim (approved):

SophCarp (approved):
Thanks Jeffrey!

Issue #5167, opened 2025-05-07
Failed to initialize `Javascript extension` using `Azd Developer Extension`


Menghua1:
**Description:**
Initializing the `Javascript extension` using `Azd Developer Extension` failed. The error messages on different machines are as follows:
- Linux and Mac:
![Image](https://github.com/user-attachments/assets/123130f1-8cc7-463d-b635-0f76af913714)

   After entering the Javascript extension directory, manually run `azd x build`, and more error messages are as follows:
![Image](https://github.com/user-attachments/assets/9df18966-4123-4ac4-809d-c7c6c8ff3a2c)

- Windows:
   Although `azd x build` ran successfully for the `JavaScript extension`, it did not generate a .zip file in the bin directory, which caused `azd x publish` to fail.
![Image](https://github.com/user-attachments/assets/f47c6ddc-3915-4c84-82c1-1ef181b36696)

   After entering the Javascript extension directory, manually run `azd x publish`, and more error messages are as follows:
![Image](https://github.com/user-attachments/assets/ee32f821-ec78-48d1-baa8-3109a17fde07)

**Repo Steps:**
1. Enable extensions alpha feature
```bash
azd config set alpha.extensions on
```

2. Install `dev` extension source

```bash
# Add a new extension source name 'dev' to your `azd` configuration.
azd extension source add -n dev -t url -l "https://aka.ms/azd/extensions/registry/dev"
```

3. Run `azd extension install microsoft.azd.extensions`.
4. Run `azd x init` and select **Javascript** language.

**Expected Results:**
The `Javascript extension` can be successfully initialized.

**Environment:**
- OS: Windows, Linux and Mac.
- Azd Version: azd version 1.15.1-daily.4836473 (commit b8385e796d2a4b2e5b431ddf3fcc2069034d7ce3)

@rajeshkamal5050 and @wbreza for notification.

PR #5178, opened 2025-05-09
Extension Templates: add missing Javascript packages to support `azd x build`

Menghua1:
Fixes https://github.com/Azure/azure-dev/issues/5167.
- Added `pkg` package.

@rajeshkamal5050 and @wbreza for notification.

### DIFF
--- a/cli/azd/extensions/microsoft.azd.extensions/internal/resources/languages/javascript/build.ps1
+++ b/cli/azd/extensions/microsoft.azd.extensions/internal/resources/languages/javascript/build.ps1
@@ -61,6 +61,7 @@ foreach ($PLATFORM in $PLATFORMS) {
     }
 
     Write-Host "Installing dependencies..."
+    npm install -g pkg
     npm install
 
     Write-Host "Building JavaScript extension for $OS/$ARCH..."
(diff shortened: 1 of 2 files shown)


cli/azd/extensions/microsoft.azd.extensions/internal/resources/languages/javascript/build.ps1
```diff
@@ -61,6 +61,7 @@ foreach ($PLATFORM in $PLATFORMS) {
     }
 
     Write-Host "Installing dependencies..."
+    npm install -g pkg
```
vhvb1989, line 64: Probably we should not install global packages on user's system w/o asking for confirmation.

Can we use npx instead? Or should we fail fast and ask user to install pkg first?  What happen, for example, if Node.js is not installed?

Menghua1: @vhvb1989 Thank you for the suggestion! I've updated the script to use `npx` to install pkg instead of installing it globally. The script also now checks whether `Node.js` and `npm` are installed, and it will exit with a message if they are missing.

vhvb1989 (approved):
LGTM

PR #5182, opened 2025-05-11
Statically generate gRPC client for JavaScript based extensions.

zedy-wj:
This PR modifies the dynamically generated gRPC clients in JavaScript-based extensions to use statically generated clients, based on [this comment](https://github.com/Azure/azure-dev/pull/5121#discussion_r2064174958).

**Prerequisite:**
When we compile the `.proto` file into a `js` file, you need to make the following preparations:
1. Install the `protoc` tool: https://github.com/protocolbuffers/protobuf/releases
2. Install the `protoc-gen-grpc-js` plugin under your project: `npm install  protoc-gen-grpc-js`.
image

**Steps:**
1. Prepare your `.proto` files.
image
2. Prepare a target path for generated files.
image
3. Run command: 
```
protoc --proto_path=<proto_path>  --js_out=import_style=commonjs,binary:<generated_path>  --grpc_out=grpc_js:<generated_path>  --plugin=protoc-gen-grpc=<protoc-gen-grpc-js.cmd_file_path>  proto/*.proto
```
**Notes:** Please replace the placeholder with your actual path. The `<protoc-gen-grpc-js.cmd_file_path>` must be an absolute path. Using a relative path sometimes does not work.

Example:
```
protoc --proto_path=./proto --js_out=import_style=commonjs,binary:./src/generated --grpc_out=grpc_js:./src/generated --plugin=protoc-gen-grpc=D:/data/repo/azd-test-extention/microsoft.azd.extensions.test02/node_modules/.bin/protoc-gen-grpc-js.cmd proto/*.proto
```
4. `js` file will be generated:
image

@rajeshkamal5050 , @wbreza for notification.


wbreza (commented):
Can you include some details in the PR description on the tools used to generate the proto implementations?

- What tools are required
- What is the full command required to generate the files

I'd also consider still including the "proto" files as well as the information requested above into a `readme.md` or similar so that the user can easily upgrade their project to support new proto versions in the future.

zedy-wj:
> Can you include some details in the PR description on the tools used to generate the proto implementations?
> 
> * What tools are required
> * What is the full command required to generate the files
> 
> I'd also consider still including the "proto" files as well as the information requested above into a `readme.md` or similar so that the user can easily upgrade their project to support new proto versions in the future.

@wbreza  We have updated description according to your comment, please re-review it, thanks!

cli/azd/extensions/microsoft.azd.extensions/internal/cmd/init.go
```diff
@@ -598,7 +598,7 @@ func createExtensionDirectory(
 		return fmt.Errorf("failed to copy and process templates: %w", err)
 	}
 
-	if extensionMetadata.Language == "dotnet" || extensionMetadata.Language == "javascript" {
```
wbreza, line 601: Lets keep the proto files for javascript and also add python so that users have a better path for regenerating new client libraries as the proto files evolve.

wbreza (approved):
Looks good - Let's just keep the proto files in the project folders for all languages except Go since Go is directly leverages the client libraries from `azd`.

cli/azd/extensions/microsoft.azd.extensions/internal/cmd/init.go
zedy-wj: We changed it to: `if extensionMetadata.Language != "go" {`.

zedy-wj:
@wbreza - Updated according to your comments, please re-review it, thanks!

wbreza (approved):

PR #5183, opened 2025-05-12
Pipeline-config - auto-params

vhvb1989:
Fix the way parameters are converted to string variable/secret values.
Add logic to only auto-set the parameters with existing mappings or with local-prompt


rajeshkamal5050 (approved):

vhvb1989:
ignoring CI errors from Windows and Mac as not related

PR #5187, opened 2025-05-12
UX: azd init confirmation declined message

### LINKED ISSUE #4836: [UX improvement] confirmation vs error
The UX here is a confusing. It gives an error message, but the action indicated was successful.

Scenario:
- `azd init -t <template-name>`
- notification that you're not in an empty directory and asks _"continue initializing app here?"_
- when selecting `no` the response is `ERROR: confirmation declined` which feels indicated that the action I took (answering `no`) just failed when in fact the intention succeeded

Image

Potential solution:
Provide a positive confirmation that the action of selecting `no` was successful and you can `cd` to your intended directory or `mkdir` for the new app you're initializing. 

Brainstorm of potential messages with @SophCarp:
- `SUCCESS: confirmation declined; app was not initialized`
- `SUCCESS: current directory is not initialized`
- `SUCCESS: confirmation declined`
- `SUCCESS: app was not initialized`

cc: @rajeshkamal5050

hemarina:
fix #4836 
image

### DIFF
--- a/cli/azd/internal/repository/initializer_test.go
+++ b/cli/azd/internal/repository/initializer_test.go
@@ -574,10 +574,10 @@ func TestInitializer_PromptIfNonEmpty(t *testing.T) {
 		files []string
 	}
 	tests := []struct {
-		name        string
-		dir         dirSetup
-		userConfirm bool
-		expectedErr string
+		name           string
+		dir            dirSetup
+		userConfirm    bool
+		declinedOutput string
 	}{
 		{
 			"EmptyDir",
@@ -595,7 +595,7 @@ func TestInitializer_PromptIfNonEmpty(t *testing.T) {
 			"NonEmptyDir_Declined",
 			dirSetup{false, []string{"a.txt"}},
 			false,
-			"confirmation declined",
+			"confirmation declined; app was not initialized",
 		},
 		{
 			"NonEmptyGitDir",
@@ -607,7 +607,7 @@ func TestInitializer_PromptIfNonEmpty(t *testing.T) {
 			"NonEmptyGitDir_Declined",
 			dirSetup{true, []string{"a.txt"}},
 			false,
-			"confirmation declined",
+			"confirmation declined; app was not initialized",
 		},
 	}
 	for _, tt := range tests {
@@ -642,12 +642,12 @@ func TestInitializer_PromptIfNonEmpty(t *testing.T) {
 				gitCli:  gitCli,
 			}
 			azdCtx := azdcontext.NewAzdContextWithDirectory(dir)
-			err := i.PromptIfNonEmpty(context.Background(), azdCtx)
 
-			if tt.expectedErr != "" {
-				require.Error(t, err)
-				assert.Contains(t, err.Error(), tt.expectedErr)
-			} else {
+			// we only test if declinedOutput is empty
+			// if confirmation is declined and app is not initialized
+			// we skip the test as it will exit with code 1
+			if tt.declinedOutput == "" {
+				err := i.PromptIfNonEmpty(context.Background(), azdCtx)
 				require.NoError(t, err)
 			}
 		})


wbreza (commented):
I don't think we should show anything "Successful" about cancelling the prompt.  The user declined the message and therefore the initialization failed.

We should come up with a better error message but IMO should stay as a red ERROR message.

vhvb1989:
> I don't think we should show anything "Successful" about cancelling the prompt. The user declined the message and therefore the initialization failed.
> 
> We should come up with a better error message but IMO should stay as a red ERROR message.

People might think that there was an error while trying to decline the overriding. Maybe we need a more neutral output for a no-op message.  The green color and SUCCESS message implies AZD did something successfully.  The red color and error means AZD encounter an error while trying to do something (either requested by the user or part of a process).

What we need here is a way to communicate:  "Ok, nothing happened. Nothing's changed".

Some other program would just not-output anything else and terminate w/o any specific output. There would be just a verbose log in stderr like "user cancel operation. Program exit without changes. No revert or rollover required".  The message would only be visible with `--debug`. Otherwise azd just stops with exit-code equals 0 (no error)

JeffreyCA:
> > I don't think we should show anything "Successful" about cancelling the prompt. The user declined the message and therefore the initialization failed.
> > We should come up with a better error message but IMO should stay as a red ERROR message.
> 
> People might think that there was an error while trying to decline the overriding. Maybe we need a more neutral output for a no-op message. The green color and SUCCESS message implies AZD did something successfully. The red color and error means AZD encounter an error while trying to do something (either requested by the user or part of a process).
> 
> What we need here is a way to communicate: "Ok, nothing happened. Nothing's changed".
> 
> Some other program would just not-output anything else and terminate w/o any specific output. There would be just a verbose log in stderr like "user cancel operation. Program exit without changes. No revert or rollover required". The message would only be visible with `--debug`. Otherwise azd just stops with exit-code equals 0 (no error)

This makes sense to me as well, I think I'd prefer a neutral message (and neutral colour), without the "SUCCESS:"/"ERROR:" prefix, or not print anything at all

wbreza:
> This makes sense to me as well, I think I'd prefer a neutral message (and neutral colour) without the "SUCCESS:"/"ERROR:" prefix, or not print anything at all

Maybe we display in our warning (yellow) color and then just return an appropriate exit code.  Based on my quick research a common exit code for use cancellation is `130` for Script terminated by Ctrl+C (SIGINT)

weikanglim:
Just my $0.02: Taking inspiration from what everyone has already said and being slightly inspired from other tooling, I wonder if using our `ErrorWithSuggestion` patterns to soften the user-perceived impact of the cancellation error is enough here:

image



It might achieve the result of softening the UX impact of the cancellation error, but retain the behavior of representing the user-cancellation request as a non-success state. @SophCarp for final thoughts here; there's a lot of good commentary on the PR comments above.

vhvb1989:
If we are collecting telemetry after errors (AZD returning non-zero exit code), could this be misleading to see as an error in the reports?  After all, AZD is just doing what is been asked...  

So, I think we have 2 things to consider:
- User experience - How clear is it that this is an AZD Error or not? 
- Telemetry collected by AZD and what would it means in one report to see the % of errors from AZD

SophCarp:
Did some quick mockups of everyone's suggestions as best I understood them

1: Error, with longer description (@wbreza)
Terminal

2: Error with suggestion @weikanglim 
Terminal-1

3:Success @SophCarp and @kristenwomack 
Terminal-2

4: Warning @wbreza 
Terminal-3

5: Warning with Exit Code @wbreza 
Terminal-4

6: Neutral text @vhvb1989 and @JeffreyCA 
Terminal-5

Personally, I think neutral is best. I wasn't sure if this is how the exit code is formatted, but I do like having an exit code, so that's my second-place item. I think errors make a user think they made a mistake (and it's incorrect to give an error when azd is following the users instructions).

wbreza:
> Personally, I think neutral is best. I wasn't sure if this is how the exit code is formatted, but I do like having an exit code, so that's my second-place item. I think errors make a user think they made a mistake (and it's incorrect to give an error when azd is following the users instructions).

image

image

I prefer either error or warning text followed by a next steps description.  It is not required to print anything about the exit code to the user - that is purely for the calling application.  If the command were called inside a script the script would then be able to react to the non-zero exit code vs assume it completed successfully.

kristenwomack:
This is the solution I'd like to move forward with because I think it clearly let's a person know that the project was _not_ initialized (still need to take an action, like `cd`) and, if I understand your point correctly @wbreza, the warning will also serve the script use case.

image

(Here's where we had discussion on this before https://github.com/Azure/azure-dev/issues/4836 where there is less conversation than I remembered.)

rajeshkamal5050 (approved):

JeffreyCA (approved):

PR #5200, opened 2025-05-15
fix support default selection for prompt param

vhvb1989:
This change allows users to define a default option (or value) for input parameters and use it when AZD prompt for it.

The use case is to force users to input/select values for parameters, making them aware of the selection, but controlling what AZD should highlight as the default option during the prompt. This is different from setting a default value directly for the parameter in bicep, which would skip the prompt.

This change also enables running with `--no-prompt` to use the default values.

### DIFF
--- a/cli/azd/pkg/infra/provisioning/bicep/prompt.go
+++ b/cli/azd/pkg/infra/provisioning/bicep/prompt.go
@@ -350,23 +350,40 @@ func (p *BicepProvider) promptForParameter(
 			return nil, fmt.Errorf("parameter '%s' has no allowed values defined", key)
 		}
 
+		// defaultOption enables running with --no-prompt, taking the default value. We use the first option as default.
+		defaultOption := options[0]
+		// user can override the default value with azd metadata
+		if azdMetadata.Default != nil {
+			if !slices.Contains(options, *azdMetadata.Default) {
+				return nil, fmt.Errorf(
+					"default value '%s' is not in the allowed values for parameter '%s'", *azdMetadata.Default, key)
+			}
+			defaultOption = *azdMetadata.Default
+		}
+
 		choice, err := p.console.Select(ctx, input.ConsoleOptions{
-			Message: msg,
-			Help:    help,
-			Options: options,
+			Message:      msg,
+			Help:         help,
+			Options:      options,
+			DefaultValue: defaultOption,
 		})
 		if err != nil {
 			return nil, err
 		}
 		value = (*param.AllowedValues)[choice]
 	} else {
+		var defaultValueForPrompt *string
+		if azdMetadata.Default != nil {
+			defaultValueForPrompt = azdMetadata.Default
+		}
 		switch paramType {
 		case provisioning.ParameterTypeBoolean:
 			options := []string{"False", "True"}
 			choice, err := p.console.Select(ctx, input.ConsoleOptions{
-				Message: msg,
-				Help:    help,
-				Options: options,
+				Message:      msg,
+				Help:         help,
+				Options:      options,
+				DefaultValue: defaultValueForPrompt,
 			})
 			if err != nil {
 				return nil, err
@@ -374,8 +391,9 @@ func (p *BicepProvider) promptForParameter(
 			value = (options[choice] == "True")
 		case provisioning.ParameterTypeNumber:
 			userValue, err := promptWithValidation(ctx, p.console, input.ConsoleOptions{
-				Message: msg,
-				Help:    help,
+				Message:      msg,
+				Help:         help,
+				DefaultValue: defaultValueForPrompt,
 			}, convertInt, validateValueRange(key, param.MinValue, param.MaxValue))
 			if err != nil {
 				return nil, err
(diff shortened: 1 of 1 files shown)


cli/azd/pkg/infra/provisioning/bicep/prompt.go
```diff
@@ -350,42 +350,61 @@ func (p *BicepProvider) promptForParameter(
 			return nil, fmt.Errorf("parameter '%s' has no allowed values defined", key)
 		}
 
+		// defaultOption enables running with --no-prompt, taking the default value. We use the first option as default.
+		defaultOption := options[0]
```
weikanglim, line 354: This isn't explicitly called out in the PR description, but I'm guessing we're making an intentional change to have `--no-prompt` always select the "first value". Is the "first value" is guaranteed to be stable across runs?

I would also admit that as a first-time user, this may not be extremely intuitive, but perhaps there are benefits that we are imagining from this change.

cli/azd/pkg/infra/provisioning/bicep/prompt.go
```diff
@@ -350,42 +350,61 @@ func (p *BicepProvider) promptForParameter(
...
+		// user can override the default value with azd metadata
+		if azdMetadata.Default != nil {
```
weikanglim, line 356: I understood from the PR description that:

```bicep
@metadata({azd: {
  default: 'foo'
}})
param fooParam string
```

is different than

```bicep
param fooParam string = 'foo'
```

Do you happen to know the context behind the intentional design separation? In an ideal world, I could wish for the Bicep defaulting case to be the de-facto implementation, but it seems that forces have pushed us down the first path. 

FWIW, I could see `defaultSelection` being slightly more descriptive than `default`, but I can see from the code diff that `default` was cemented in as part of the API previously.

cli/azd/pkg/infra/provisioning/bicep/prompt.go
```diff
@@ -350,42 +350,61 @@ func (p *BicepProvider) promptForParameter(
...
 				return nil, err
 			}
 			value = (options[choice] == "True")
 		case provisioning.ParameterTypeNumber:
 			userValue, err := promptWithValidation(ctx, p.console, input.ConsoleOptions{
-				Message: msg,
-				Help:    help,
+				Message:      msg,
+				Help:         help,
+				DefaultValue: defaultValueForPrompt,
```
weikanglim, line 396: Do you happen to have examples to share for number and booleans? I'm curious how the type conversion ends up happening and what the user provided values end up looking like.

weikanglim:
Nothing blocking, just left a few minor comments

cli/azd/pkg/infra/provisioning/bicep/prompt.go
vhvb1989: yeah, this change is trying to enable `--no-prompt` . The current user's feedback is: "why isn't AZD choosing the highlighted value I see during the prompt like in other prompts, like the location parameter?"  

User is aware of how --no-prompt can automate the process by taking whatever selection is displayed in a list. Wondering why that's not the case when not using location parameters

cli/azd/pkg/infra/provisioning/bicep/prompt.go
vhvb1989: yeah, I initially named it `defaultSelection` but renamed it during PR review.  

The difference for one and the other is if the parameter selection should be presented to the user for awareness or not.
Bicep default prevents from any AZD-prompting, making users to possibly ignore available options, of forcing them to set ENV VARS or manually updating the bicep file to change things.

The defaultSelection helps templates to force a user interaction to at least accept the default proposed input for the template. 

We could make AZD to always prompt to confirm even default values... but then we could see feedback about too much prompting.

cli/azd/pkg/infra/provisioning/bicep/prompt.go
vhvb1989: The `default` field for the azd metadata is `*string` .  This means that if you want to set a default, you must quote it as string. If you don't quote it, it is ignored and the value becomes nil.

I didn't want to update the *string to any as part of this change.

In the future, if someone ask for it, we can improve this by supporting any type and calling `sprintf(%v, value)` to get its string representation.  But I dont think we need it now

cli/azd/pkg/infra/provisioning/bicep/prompt.go
weikanglim: Thanks for the explanation! I guess this does preserve Bicep semantics where default parameters would be accepted by default. And if I read `azd:{}` as "azd prompt metadata", this does all make a whole lot of sense.

cli/azd/pkg/infra/provisioning/bicep/prompt.go
weikanglim: I see. Thanks for sharing that perspective for `--no-prompt` to skip most prompts for convenience.

Just to double check on the initial question: is `options[0]` chosen here corresponding to the first value declared in the allowed list?

For example, in `@allowed(['foo', 'bar'])`, is `foo` always chosen? I guess if we do document it as such, it is understandable behavior, and makes `--no-prompt` useful for this specific purpose, so I'm in favor of doing it assuming the order is indeed preserved.

cli/azd/pkg/infra/provisioning/bicep/prompt.go
weikanglim: Thanks for clarifying the behavior!

If I understand the explanation correctly, this is what we handle currently:

```bicep
@metadata({
  azd: {
    default: 'True'
  }
})
param somethingElse bool

@metadata({
  azd: {
    default: '12345'
  }
})
param digit int
```

While a more realistic version that for a Bicep user may be more along the lines of:

```bicep
@metadata({
  azd: {
    default: true
  }
})
param somethingElse bool

@metadata({
  azd: {
    default: 12345
  }
})
param digit int
```

From the above, I'm wondering if we could consider defer adding support for these types, or perform the conversion to match the expected types when writing Bicep -- the bool case of 'True'/'False' isn't particularly intuitive...

The main thing is that I'd prefer if we wouldn't add or maintain support for the stringified version for longer term.

PR #5128, opened 2025-04-28
[ext builds] Cleanup

danieljurek:
Cleanup some lingering TODOs

PR #5202, opened 2025-05-15
Extension EngSys

danieljurek:
* Builds
* Signs
* Releases


danieljurek:
Waiting on https://github.com/Azure/azure-dev/pull/5128 to merge and the pending comment

eng/pipelines/release-ext-demo.yml
```diff
@@ -0,0 +1,33 @@
+# Continuous deployment trigger
```
wbreza, line 1: Nit: We may want to come up with naming strategy for these pipelines based off the "id" of the extension.  For example, the full id of this extension is `microsoft.azd.demo` so maybe the pipeline should be `release-ext-microsoft-azd-demo.yml` ?

wbreza (approved):

PR #5204, opened 2025-05-15
Login guard middleware with automatic login

wbreza:
## Adds Login Guard middleware to detect if a user is logged in
- Action descriptors can specify whether or not they require login
- Middleware runs when action or parent action requires login

Adds login guard for the following:
- `azd pipeline *`
- `azd provision`
- `azd deploy`
- `azd up`
- `azd down`

_Let me know if i am missing any commands_

## Updated to prompt for auto-login

One of the cool things we can do with the middleware is to prompt the user to login right now - then after login the action they wanted to run will just kick off.

### User is not logged in yet
image

### User choose to continue with login
After login - original action continues without interrupting the user or throwing an error.
image


rajeshkamal5050 (approved):

cli/azd/cmd/container.go
JeffreyCA, line 1: What does this affect from a user perspective? I noticed without these changes we already show an error:

```
vscode ➜ /workspaces/ai $ azd up

ERROR: not logged in, run `azd auth login` to login
```

JeffreyCA (approved):

cli/azd/cmd/container.go
wbreza: Right now it is a transparent change to the user.  IMO - it's just a cleaner and consistent way to support verify a user is logged in before running a command.

One additional thing we could enable here in this scenario is automatically run the `login` action as part of the middleware.  After successful login the command could just continue as normal.

wbreza:
@SophCarp  - Thoughts on this auto-login feature? Why throw an error when we can just help the user get logged in now.

SophCarp:
Thats great, I love it!

SophCarp (approved):

PR #5242, opened 2025-05-30
compose beta: Remove `alpha.compose` flag and enable compose codepaths by default

### LINKED ISSUE #5231: compose beta: Remove `alpha.compose` and enable compose codepaths by default
Implications of removing `alpha.compose` feature flag (and enabling it by default):

### `azd init`

For users that did not have `alpha.compose` turned on, the main behavioural changes after upgrading to the new version of azd would be:

1. **`infra/` folder no longer gets generated after `azd init`** - this is because the Bicep infra will be generated on the fly, in-memory during `azd up`. Unless the user explicitly runs `azd infra synth` to output the Bicep to disk.

I wonder if we should add a prompt to `azd init` asking if azd should generate the infra files on disk (like how azd behaved previously)?

And was also thinking if we should mention `azd add` somewhere here:

Image

2. New projects initialized with `azd init` will have a new `resources` node in `azure.yaml`:

```diff
name: src
metadata:
    template: azd-init@0.0.0-dev.0
services:
    src:
        project: .
        host: containerapp
        language: python
        docker:
            path: Dockerfile
+ resources:
+     src:
+         type: host.containerapp
+         port: 5000
```

Users might wonder about the new `resources:` node, but probably a minor thing.

3. If a user has an existing project that was initialized with simplified init and has a service host like ACA, `azd add` does not work very well:
    **Reason 1**: non-compose projects don't have a `resources` node in `azure.yaml`, so in the `azd add` flow, azd won't prompt for the 'uses' relationship when adding a resource like a DB or AI resource, even though there's a service defined in `azure.yaml`.

    **Reason 2**: If the user runs `azd add` to add AI Search for example, they would need to run `azd infra synth` in order to be able to provision it. But that leads to another problem, see Reason 3.

    **Reason 3**: `azd infra synth` can be dangerous as it may erase all existing Bicep in the `infra` folder, like the Bicep code for the ACA host. This is related to Reason 1 as there's no `containerapp` resource in a non-compose `azure.yaml`.

@kristenwomack / @puicchan / @SophCarp / @rajeshkamal5050 Any thoughts on these implications and whether we should change any of the UX here? And please call out anything I may have missed.

### `azd show`

Should not be disruptive, as it's adds a new ability to run `azd show <resource-name>`.

### `azd env set-secret`

There are some shortcuts to ask the user up front if they wish to use the compose Key Vault, but they can always fall back to manually picking a key vault, so this shouldn't be disruptive either.

JeffreyCA:
Closes #5231

`azd infra synth` changes will be handled in a separate PR.

In general, references to `alpha.compose` were removed and codepaths that had something like:

```go

if e.alphaFeatureManager.IsEnabled(featureCompose) {
   (composeCode)
} else {
   ...
}
```

were changed to something like this with the non-compose codepaths removed:

```
(composeCode)
```

### Help text
image

image

### azd init

- `infra/` folder no longer gets generated during `azd init` when initializing from app code or minimal project.
- When initializing from app code or minimal project, env initialization is deferred to `azd up`, **unless** the user passes `--environment <env name>`/`-e <env name>` to the `azd init` command, or the `AZURE_ENV_NAME` env var is set.
- Reworded the success message to mention `azd add` and and `azd infra synth`:

  image

### azd add
- Command is blocked for incompatible projects and templates.
- A project is considered **incompatible** with `azd add` if it has an `infra` folder **AND** there is no `resources` node in azure.yaml. This should cover most existing templates out there.
- Error message:

  image

### azd show
- New optional argument for `azd show`- users can pass in a resource name or full Azure resource ID: `azd show <resource name>` and `azd show <resource ID>`.

### azd env set-secret
- azd will prompt whether to use the project/compose AKV, if one exists. Users can always fallback to manually selecting a key vault.


cli/azd/cmd/env.go
JeffreyCA, line 1: These changes just remove the if statement on `featureCompose` and adjusts the indentation.

cli/azd/cmd/init.go
```diff
@@ -314,54 +316,13 @@ func (i *initAction) Run(ctx context.Context) (*actions.ActionResult, error) {
...
-				prjConfig.MetaSchemaVersion = "alpha"
-			}
-
-			err = project.Save(ctx, &prjConfig, azdCtx.ProjectPath())
-			if err != nil {
-				return nil, fmt.Errorf("saving project config: %w", err)
-			}
-
-			followUp = "Run " + output.WithHighLightFormat("azd add") + " to add new Azure components to your project."
+		err = i.repoInitializer.InitializeMinimal(ctx, azdCtx)
```
JeffreyCA, line 319: The previous 'else' block (when composeAlphaEnabled was true) has been moved into `InitializeMinimal(ctx, azdCtx)`

cli/azd/cmd/init.go
```diff
@@ -314,54 +316,13 @@ func (i *initAction) Run(ctx context.Context) (*actions.ActionResult, error) {
...
-			}
-
-			if composeAlphaEnabled {
-				prjConfig.MetaSchemaVersion = "alpha"
-			}
-
-			err = project.Save(ctx, &prjConfig, azdCtx.ProjectPath())
-			if err != nil {
-				return nil, fmt.Errorf("saving project config: %w", err)
-			}
```
JeffreyCA, line 359: This 'else' block logic (when composeAlphaEnabled was true) has been moved into `InitializeMinimal(ctx, azdCtx)`

cli/azd/internal/repository/app_init.go
```diff
@@ -235,107 +232,25 @@ func (i *Initializer) InitFromApp(
...
-
-		err = scaffold.Execute(t, "next-steps-alpha.md", nil, filepath.Join(azdCtx.ProjectDirectory(), "next-steps.md"))
-		if err != nil {
-			return err
-		}
-
-		i.console.MessageUxItem(ctx, &ux.DoneMessage{
-			Message: "Generating " + output.WithHighLightFormat("./next-steps.md"),
-		})
-	}
```
JeffreyCA, line 291: Keep only the 'else' block contents (`infraSpec` is nil when compose is enabled)

rajeshkamal5050 (approved):

cli/azd/cmd/init.go
```diff
@@ -279,9 +279,11 @@ func (i *initAction) Run(ctx context.Context) (*actions.ActionResult, error) {
 		tracing.SetUsageAttributes(fields.InitMethod.String("app"))
 
 		header = "Your app is ready for the cloud!"
-		followUp = "You can provision and deploy your app to Azure by running the " + output.WithHighLightFormat("azd up") +
-			" command in this directory. For more information on configuring your app, see " +
-			output.WithHighLightFormat("./next-steps.md")
+		followUp = "Run " + output.WithHighLightFormat("azd up") + " to provision and deploy your app to Azure.\n" +
+			"Run " + output.WithHighLightFormat("azd infra synth") + " to write infrastructure as code files to disk " +
```
SophCarp, line 283:
```suggestion
			"Run " + output.WithHighLightFormat("azd infra gen") + " to generate infrastructure as code files to disk " +
```
I know you've got another infra gen PR update, want to make sure this is captured in one of them.

cli/azd/internal/cmd/add/add.go
```diff
@@ -390,6 +386,26 @@ func (a *AddAction) Run(ctx context.Context) (*actions.ActionResult, error) {
...
+func ensureCompatibleProject(
+	prjConfig *project.ProjectConfig,
+) error {
+	infraRoot := prjConfig.Infra.Path
+	if !filepath.IsAbs(infraRoot) {
+		infraRoot = filepath.Join(prjConfig.Path, infraRoot)
+	}
+
+	hasResources := len(prjConfig.Resources) > 0
+	hasInfra, _ := pathHasInfraModule(infraRoot, prjConfig.Infra.Module)
```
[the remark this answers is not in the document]
JeffreyCA: Thanks, updated

cli/azd/internal/repository/initializer.go
```diff
@@ -452,7 +425,15 @@ func (i *Initializer) writeCoreAssets(ctx context.Context, azdCtx *azdcontext.Az
...
 
-	//create .gitignore or open existing .gitignore file, and contains .azure
+	err = i.writeGitignore(ctx, azdCtx)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+// writeGitignore creates a project .gitignore or opens the existing .gitignore file, and ensures it contains .azure.
```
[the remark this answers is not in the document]
JeffreyCA: This is currently how it behaves already

cli/azd/internal/cmd/add/add.go
```diff
@@ -390,6 +387,34 @@ func (a *AddAction) Run(ctx context.Context) (*actions.ActionResult, error) {
...
+	if err != nil {
+		if errors.Is(err, os.ErrNotExist) {
+			hasInfra = false
+		} else {
+			return err
+		}
+	}
+
+	if hasInfra && !hasResources {
+		return fmt.Errorf("incompatible project: found infra directory with no resources defined in azure.yaml")
```
SophCarp, line 412: Might want to change this, not sure. Can't think of a better way to phrase it yet.

SophCarp: Or be sure we have good docs related to this error code exactly. Maybe link a doc to this line?

SophCarp (approved):

cli/azd/cmd/init.go
JeffreyCA: Thanks, yes will update this in the infra gen PR

cli/azd/internal/cmd/add/add.go
JeffreyCA: Yes, I think we can include an aka.ms shortlink to a troubleshooting section on https://learn.microsoft.com/en-us/azure/developer/azure-developer-cli/troubleshoot. Happy to also update the error text in a follow up PR if you think of a better message.

PR #5246, opened 2025-06-02
Remove ACA Environment Contributor role assignment

eerhardt:
This role assignment was necessary as an early workaround for getting access to the Aspire Dashboard hosted in ACA.

The explicitly role assignment is no longer necessary because the user doing the deployment will automatically inherit the permission because they are the user who created the ACA environment.

cc @vhvb1989


vhvb1989 (approved):
thank you

cli/azd/pkg/apphost/testdata/TestAspireBicepGeneration-resources.bicep.snap
```diff
@@ -72,14 +72,6 @@ resource containerAppEnvironment 'Microsoft.App/managedEnvironments@2024-02-02-p
 
 }
```
eerhardt, line 74: I'm going to change it to not have this double blank line.

Issue #4174, opened 2024-08-02
Support MI for azd pipeline config + rg deployment


vhvb1989:
When using resource group deployments, azd can provide a way to use an existing MI (or create one) instead of using an app registration.
azd would create a fed-credential for the MI and set it up for the CI/CD pipeline

seesharprun:
MI support would be amazing. Even OIDC would be better than the current "SDK-style auth": https://github.com/azure/login?tab=readme-ov-file#login-with-openid-connect-oidc-recommended

I think `azure/login` even shows a warning for using SDK-style auth and the Azure CLI shows deprecation warnings from trying to get your SP's credentials using this older style of auth.

seesharprun:
Here's where they talk about `--sdk-auth` being deprecated (eventually):

- https://github.com/Azure/azure-cli/issues/22297
- https://github.com/Azure/login/issues/223

kewalaka:
I wrote some notes for this, would love some feedback on the approach below.  One advantage of an SP is you don't need a location to host it, whereas with a UMI you need an RG to put it in.

---

## Goals

- Allow users to opt for a Managed Identity (UAMI) instead of a Service Principal for pipeline authentication.
- Prefer a project-scoped MI by default (one MI per project), with the option to use an existing user-specified MI.
- Automatically create or select a resource group for the MI if not specified.
- Auto-generate both RG and MI names if not provided, e.g., `rg-<random>` and `mi-<random>`.
- Add federated credentials to the MI for GitHub Actions/Azure Pipelines OIDC trust.
- Output MI details for use in pipeline secrets/variables.
- Ensure `azd down` removes auto-generated RG and MI artifacts (unless user-specified).
- Maintain backward compatibility with existing SP-based flows.

---

## High-Level User Flow

1. **New MI flow**:  
    ```
    azd pipeline config --use-managed-identity [--mi-resource-group <name>] [--mi-name <name>] [other options]
    ```
2. azd:
    - If `--mi-resource-group` is specified, use it; otherwise, auto-generate (`rg-<random>`).
    - If `--mi-name` is specified, use it; otherwise, auto-generate (`mi-<random>`).
    - Creates (or uses) the RG and MI.
    - Adds a federated credential to the MI for the pipeline’s OIDC issuer/subject.
    - Assigns required roles to the MI (Contributor or custom).
    - Outputs MI client ID/tenant ID for pipeline configuration.
    - (Optionally) automates GitHub Environment and secret setup.

[...]

PR #5248, opened 2025-06-02
Support MSI auth set-up for pipeline config

vhvb1989:
This PR allows folks to configure CI/CD using MSI.
AZD can use an existing MSI or create a new one.
As part of the changes, AZD can also allow skipping the auth set-up (for people doing manual-custom auth)

fix: https://github.com/Azure/azure-dev/issues/4174


cli/azd/pkg/armmsi/armmsi.go
```diff
@@ -0,0 +1,259 @@
...
+//   - credentialProvider: Provides credentials for the subscription.
+//   - armClientOptions: Optional configuration options for the ARM client.
+//
+// Returns:
+//
+//	An initialized ArmMsiService instance.
+//	- error: An error object if the operation fails, otherwise nil.
+func (s *ArmMsiService) CreateUserIdentity(
+	ctx context.Context,
+	subscriptionId, resourceGroup, location, name string) (armmsi.Identity, error) {
```
wbreza, line 49: I feel like we should continue with our model to return (pointer, err) on these calls.  On successful calls we can return a pointer to the identity, on error scenarios return nil pointer instead of empty struct.

I know we have had this same conversation multiple times.

cli/azd/pkg/armmsi/armmsi.go
```diff
@@ -0,0 +1,259 @@
...
+			Location: to.Ptr(location),
+		}, nil)
+	if err != nil {
+		return armmsi.Identity{}, err
+	}
+
+	return msi.Identity, nil
+}
+
+// ListUserIdentities retrieves a list of user-assigned managed identities within a specified resource group.
```
wbreza: Nit: Looks like this returns all user assigned identities for the subscription, not specific to a resource group.

wbreza (commented):
Looks good.  The only feedback I have is around the standard for when we return pointers vs empty structs.

In this example - if we followed the model from the Entra service - which this is very similar that one returns pointer/error on most calls.

cli/azd/pkg/armmsi/armmsi.go
vhvb1989: tks

cli/azd/pkg/armmsi/armmsi.go
vhvb1989: I'm keeping this to make the CRUD operations for this ArmMsiService consistent. By consistent I mean that the service returns values or list of values.

API:
```go
ApplyFederatedCredentials(ctx context.Context, subscriptionId string, msiResourceId string, federatedCredentials []armmsi.FederatedIdentityCredential) ([]armmsi.FederatedIdentityCredential, error)
CreateFederatedCredential(ctx context.Context, subscriptionId string, resourceGroup string, msiName string, name string, subject string, issuer string, audiences []string) (armmsi.FederatedIdentityCredential, error)
CreateUserIdentity(ctx context.Context, subscriptionId string, resourceGroup string, location string, name string) (armmsi.Identity, error)
GetUserIdentity(ctx context.Context, resourceId string) (armmsi.Identity, error)
ListUserIdentities(ctx context.Context, subscriptionId string) ([]armmsi.Identity, error)
```

Moving to return `(*armmsi.Identity, error)` would be inconsistent with `ListUserIdentities()` returning `([]armmsi.Identity, error)`.  I would need to update it to return `[]*armmsi.Identity`, which is a double reference that adds extra complexity to this simple CRUD. Using slices of pointers is usually a last resort for me.

vhvb1989 (commented):
responding to comments

vhvb1989 (commented):
responding to comments

rajeshkamal5050 (approved):

Issue #5125, opened 2025-04-25
.NET Aspire single-service deployment


kristenwomack:
Request: https://devdiv.visualstudio.com/DevDiv/_workitems/edit/2332938

https://aka.ms/dev18-azure-singleservicedeployment

PR #5249, opened 2025-06-03
vs aspire integration service name support

vhvb1989:
Adding `DeployServiceAsync` to VS-Server to allow deploying one service instead of all services.
fix: https://github.com/Azure/azure-dev/issues/5125


bradygaster:
very exciting!

PR #5270, opened 2025-06-05
terraform pipelineconfig oidc

vhvb1989:
- Remove the blocking paths for Terraform+OIDC and the warnings and code to force client-secret for Terraform.
- Update pipeline definition to install Terraform and set use oidc flag when using OIDC


spboyer:
@vhvb1989 can we implement OIDC for all pipeline config? Service Principal is blocked by policy for some/most. Is this code reusable there?

@paulyuk

vhvb1989:
> @vhvb1989 can we implement OIDC for all pipeline config? Service Principal is blocked by policy for some/most. Is this code reusable there?
> 
> @paulyuk

OIDC is already enabled for `pipeline config`.  The only missing part was OIDC for Terraform (hence this PR).

cli/azd/pkg/pipeline/pipeline_manager.go
```diff
@@ -1270,6 +1276,16 @@ func generatePipelineDefinition(path string, props projectProperties) error {
 		}
 	}
 
+	if props.InfraProvider == infraProviderTerraform {
+		// terraform provider does not resolve this variables automatically, AZD needs to define them
+		tmplContext.Variables = append(tmplContext.Variables, "AZURE_LOCATION")
+		tmplContext.Variables = append(tmplContext.Variables, "AZURE_ENV_NAME")
```
hemarina, line 1282: Just curious, if location and env name needs to be define, do we need to define other variables like subscription?

vhvb1989: Good question.  Subscription is automatically set by the pipeline manager.  Location and Env used to also be handled by the pipeline manager, but we changed those two to be provided by the provisioning provider (bicep or terraform).

PR #5280, opened 2025-06-10
`azd init` - Combine "Create a minimal project" with "Scan current directory" option and add `--minimal` flag

### LINKED ISSUE #5258: azd init: combine "Use code in the current directory" and "Create a minimal project" options
With composability in beta, the "Create a minimal project" init option doesn't make as much sense as its own init option since it only creates an empty `azure.yaml` file

The team agreed on combining the "Create a minimal project" logic into the "Use code in the current directory" option when azd doesn't detect any app code and introduce a new flag like `azd init --minimal` that does what the "Create a minimal project" option does today

JeffreyCA:
Closes #5258

`azd init` now shows two options, with the "Use code in the current directory" option renamed to "Scan current directory":
image

To directly initialize a minimal project, users can pass `--minimal` or `-m` to `azd init`:

https://github.com/user-attachments/assets/f07f7180-3cf4-4594-b5f7-94b0b13e8ebc

To initialize a minimal project using the interactive `azd init` experience, there are two ways:

Case 1: azd detects services/databases

1. Run `azd init` and select **Use code in the current directory**.
2. When prompted, select **Remove a detected service** to remove all detected services and databases.
3. Select **Confirm and continue initializing my app**.

https://github.com/user-attachments/assets/a90c9200-8e89-46f6-b77b-43041b06be56

Case 2: azd does not detect any services/databases

1. Run `azd init` and select **Use code in the current directory**.
2. When prompted, select **Confirm and continue initializing my app** to initialize project without any services/databases.

https://github.com/user-attachments/assets/4c77eab8-a08c-4361-b174-55adf7115028

If multiple incompatible flags are passed, `azd init` fails with an error:

image


cli/azd/cmd/init.go
```diff
@@ -358,17 +366,15 @@ const (
...
-	initProject
 	initEnvironment
 )
 
 func promptInitType(console input.Console, ctx context.Context) (initType, error) {
 	selection, err := console.Select(ctx, input.ConsoleOptions{
 		Message: "How do you want to initialize your app?",
 		Options: []string{
-			"Use code in the current directory",
+			"Use code in the current directory", // This now covers minimal project creation too
```
SophCarp:
```suggestion
			"Scan current directory", // This now covers minimal project creation too
```

This was one change we briefly talked about. "Use code in current directory" sounds like it doesn't work if you don't have code in the file. "Scan current directory" is accurate, and hopefully feels like your directory can be scanned even if it's empty.

SophCarp (approved):
Looks good! I made one suggestion but that's it.

cli/azd/cmd/init.go
JeffreyCA: Thank you! I forgot about that one

vhvb1989 (approved):
LGTM

Issue #4638, opened 2024-12-17
Slash '/' in 'name:' of azure.yaml will break service packaging


crpietschmann:
- [x] Make sure you've installed the latest version using [instructions in the wiki](../wiki/install)

**Output from `azd version`**
Run `azd version` and copy and paste the output here:

```
azd version 1.11.0 (commit 5b92e0687e1fa96dfc8292f4b900c0c58610b6a5)
```

**Describe the bug**
When the `name:` within `azure.yaml` contains a slash `/` character, it breaks the packaging of services.

error that ocurrs:

```
root@MacBookPro azd-react-bootstrap-dashboard # azd package api

Packaging services (azd package)

  (x) Failed: Packaging service api

ERROR: failed packaging service 'api': failed when creating zip package to deploy api: open /tmp/build5nines/azd-react-bootstrap-dashboard-api-azddeploy-1734478898.zip: no such file or directory
```

**To Reproduce**
1. change the `name:` of the template in the `azure.yaml` file to have a slash `/` character in it; such as `build5nines/azd-react-bootstrap-dashboard'
2. add a service for a web project, such as a node.js app

**Expected behavior**
It's expected that the service packaging complete successfully, or gives a warning that the `name` contains an invalid character.

```
root@MacBookPro azd-react-bootstrap-dashboard # azd package api     

Packaging services (azd package)

  (✓) Done: Packaging service api
  - Package Output: /tmp/build5nines-azd-react-bootstrap-dashboard-api-azddeploy-1734479080.zip

  (-) Skipped: Packaging service dashboard
```

**Environment**
Information on your environment:
    * Language name and version: Node.js v23.3.0
    * IDE and version : Latest VS Code
    * OS: macOS

**Additional context**

NanaXiong00:
This issue can be reproduced, and we will fix it in the following approach:

If there is a slash `/` in the "name:" of `azure.yaml`, change the `/` to `-` when packaging without any prompts.

CC: @rajeshkamal5050

Menghua1:
> This issue can be reproduced, and we will fix it in the following approach:
> 
> If there is a slash `/` in the "name:" of `azure.yaml`, change the `/` to `-` when packaging without any prompts.

@vhvb1989 What do you think about this approach? Should we silently replace `/` with `-`, or would it be better to warn the user that the name contains an invalid character and prompt them to fix it?

vhvb1989:
Thanks for taking thins @Menghua1 

[...]

PR #5380, opened 2025-06-19
Add schema validation for `name` in azure.yaml and add name validity check in AZD

Menghua1:
Fix https://github.com/Azure/azure-dev/issues/4638. This PR includes the following changes:
 
1. Added a regex pattern to the `name` field in `azure.yaml.json` to ensure it only contains letters, numbers, and "-". and the name must start and end with a letter or number. When editing, any invalid characters will be indicated with a red wavy underline as shown below:
image
 
2. Added validation of project names during the `azd init` command: **only lowercase letters, numbers, and "-" are allowed, and the name must start and end with a letter or number**. If the name is invalid, the operation will exit early, as shown below:
image


cli/azd/pkg/project/service_manager.go
```diff
@@ -278,6 +279,11 @@ func (sm *serviceManager) Package(
 		options = &PackageOptions{}
 	}
 
+	// Check if the application name is valid.
+	if err := validateAppName(serviceConfig.Project.Name); err != nil {
+		return nil, err
+	}
+
```
vhvb1989: This might be too late as a validation.

We need to make sure that AZD is not allowing folks to name the project with something invalid when they run `azd init` hoping to azd-fy an existing project.

See: https://github.com/azure/azure-dev/blob/main/cli/azd/internal/repository/initializer.go#L338

AZD should not allow giving a name for the project like `uno/dos` only to later fail during package about having an invalid name.

Menghua1: @vhvb1989 The code has been updated as suggested. Please review again, thanks!

vhvb1989 (approved):
LGTM - Thank you.

cli/azd/internal/repository/initializer.go
```diff
@@ -343,6 +344,11 @@ func (i *Initializer) InitializeMinimal(ctx context.Context, azdCtx *azdcontext.
 		return err
 	}
 
+	// Check if the project name is valid.
+	if err := names.ValidateProjectName(name); err != nil {
+		return err
+	}
+
```
weikanglim: FYI, This is actually dead code here -- `ProjectName()` already cleans the auto-generated name.

@vhvb1989 Just checking from your side, was your recommendation to validate azure.yaml on Load()?

vhvb1989: on `Load()` might be too late, @weikanglim .  AZD needs to check for valid name when it prompts the user for a project name.  It would be a bad experience to prompt and allow folks to input `foo/bar` just to fail on the next call to azd while trying to load the project again.  

Checking during `load()` could be good for anyone manually creating azure.yaml.  That could be either part of this same PR or as a new PR.  I think the rules about the valid names are already defined in the YAML schema, so I would expect any manual writers to see a red line in the name if they do something like `foo/bar`.  So I am a little less worried about those cases and only currently laser focus on blocking AZD prompt from allowing `foo/bar` as the name of the project

JeffreyCA: From testing, this is still used when the user specifies a custom name with invalid characters. Also, I think we should put this and the `Prompt` call in a `for` block, so it re-prompts the user for a valid name, instead of aborting the entire init flow.

Also wondering if `ProjectName()` should also strip away leading/trailing hyphens? When the working directory is named something like `_app`, `ProjectName()` turns it into `-app` which is invalid

weikanglim: Thanks for the explanations!

I did miss the part where we do actually prompt the user for a name here, and softly validating in the yaml schema without a hard break sounds reasonable.

@Menghua1 I agree with @JeffreyCA that a re-prompting should happen -- not an immediate failure. It's an important piece of the first-time experience. Would you be able to address this?

weikanglim: > Also wondering if ProjectName() should also strip away leading/trailing hyphens? When the working directory is named something like _app, ProjectName() turns it into -app which is invalid

That's a little odd.. We do have a test case for this: 

```golang
{"StartEndHyphens", "-abc-", "abc"},
```

And it did work for me here:

Image

JeffreyCA: Ah I had consecutive underscores: `__app` -> `-app` 😅

Menghua1: @weikanglim @JeffreyCA Thanks for the suggestions!
I've updated the initialization process to re-prompt for a valid project name instead of aborting on invalid input, as shown below:
image

Additionally, `ProjectName()` now removes leading and trailing delimiters.
Please review again, thanks!

JeffreyCA (approved):

cli/azd/internal/repository/initializer.go
JeffreyCA: Looks good, thanks!

PR #5384, opened 2025-06-20
Update the `default` azd metadata field's type to `any`

Menghua1:
This PR updates the `AzdMetadata.Default` field type to `any` to support default values of type `bool` and `int`.

- For `bool` parameters: The `AzdMetadata.Default` field must be of type `bool`. 
image

  Other types are not supported and will result in an error as follow:
image

- For `int` parameters: The `AzdMetadata.Default` field must be of type `int`.
image

  Other types are not supported and will result in an error as follow:
image

### DIFF
--- a/cli/azd/pkg/infra/provisioning/bicep/prompt_test.go
+++ b/cli/azd/pkg/infra/provisioning/bicep/prompt_test.go
@@ -329,3 +329,168 @@ func (m *mockCurrentPrincipal) CurrentPrincipalId(_ context.Context) (string, er
 func (m *mockCurrentPrincipal) CurrentPrincipalType(_ context.Context) (provisioning.PrincipalType, error) {
 	return provisioning.UserType, nil
 }
+
+func TestPromptForParameterOverrideDefault(t *testing.T) {
+	t.Parallel()
+
+	mockContext := mocks.NewMockContext(context.Background())
+
+	prepareBicepMocks(mockContext)
+
+	p := createBicepProvider(t, mockContext)
+
+	mockContext.Console.WhenSelect(func(options input.ConsoleOptions) bool {
+		return strings.Contains(options.Message, "for the 'testParam' infrastructure parameter")
+	}).RespondFn(func(options input.ConsoleOptions) (any, error) {
+		require.Equal(t, 3, len(options.Options))
+		require.Equal(t, "good", options.DefaultValue)
+		return 1, nil
+	})
+
+	value, err := p.promptForParameter(*mockContext.Context, "testParam", azure.ArmTemplateParameterDefinition{
+		Type:          "string",
+		AllowedValues: to.Ptr([]any{"three", "good", "choices"}),
+		Metadata: map[string]json.RawMessage{
+			"azd": json.RawMessage(`{"default": "good"}`),
+		},
+	}, nil)
+
+	require.NoError(t, err)
+	require.Equal(t, "good", value)
+}
+
+func TestPromptForParameterOverrideDefaultError(t *testing.T) {
+	t.Parallel()
+
+	mockContext := mocks.NewMockContext(context.Background())
+
+	prepareBicepMocks(mockContext)
+
+	p := createBicepProvider(t, mockContext)
+
+	_, err := p.promptForParameter(*mockContext.Context, "testParam", azure.ArmTemplateParameterDefinition{
+		Type:          "string",
+		AllowedValues: to.Ptr([]any{"three", "good", "choices"}),
+		Metadata: map[string]json.RawMessage{
+			"azd": json.RawMessage(`{"default": "other"}`),
+		},
+	}, nil)
+
+	require.Error(t, err)
+}
+
+func TestPromptForParameterEmptyAllowedValuesError(t *testing.T) {
+	t.Parallel()
+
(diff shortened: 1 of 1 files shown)


kristenwomack:
FYI @alexwolfmsft for updating the docs

cli/azd/pkg/infra/provisioning/bicep/bicep_provider.go
```diff
@@ -191,7 +191,9 @@ func defaultPromptValue(locationParam azure.ArmTemplateParameterDefinition) *str
 		azdMetadata.Type != nil && *azdMetadata.Type == azure.AzdMetadataTypeLocation &&
 		azdMetadata.Default != nil {
 		// Metadata using location type and a default location. This is the highest priority.
-		return azdMetadata.Default
+		defaultStr := fmt.Sprintf("%v", azdMetadata.Default)
+		tmp := defaultStr
+		return &tmp
```
JeffreyCA:
```suggestion
		defaultStr := fmt.Sprintf("%v", azdMetadata.Default)
		return &defaultStr
```

cli/azd/pkg/infra/provisioning/bicep/prompt.go
```diff
@@ -372,23 +373,38 @@ func (p *BicepProvider) promptForParameter(
...
+				}
+			}
+			msg := fmt.Sprintf("Would you like to set the '%s' infrastructure %s to %t?", key, securedParam, defaultBool)
+			confirmValue, err := p.console.Confirm(ctx, input.ConsoleOptions{
 				Message:      msg,
 				Help:         help,
-				Options:      options,
-				DefaultValue: defaultValueForPrompt,
+				DefaultValue: true,
 			})
```
JeffreyCA: I feel like the new prompt change could get a bit confusing, especially if the Bicep param has a default of `false` because of the double negative.

Bicep:
```bicep
@metadata({azd: {
  default: false
}})
param param1 bool
```

New prompt: if the user inputs 'n', then the value becomes `true` in this case:

image

Old prompt, a bit more explicit:
image

@SophCarp would love to get your thoughts here as well 🙂

hemarina: I prefer the way of selection (y/n) in the new prompt, but it does feel confusing due to double negative. Could we ask if users would like to set 'paraml' to true (y/N)?

SophCarp: hmmm. By default, param1 would be false. y/n isn't the same as true/false, but rather "confirm/change". "default" should be included in the question, because otherwise a customer might be confused why they're being asked a double negative question.

Maybe :
```bash
? 'param1's default value is false. Keep default value? y/n
```
or:
 
```bash
? Infrastructure parameter 'param1' has a default value of 'false'. Keep default value?
> Keep (param1=false)
Change (param1=true) 
```

or:

```bash
? Confirm infrastructure parameter 'param1' = false:
> Yes (param1=false)
   No (param1=true)
```

or:
```
? Confirm infrastructure parameter 'param1' = false: y/n
```
Is there more direct wording than "infrastructure parameter"? This is different from "environment variable", right?

vhvb1989:
used https://github.com/Azure/azure-dev/pull/5386 to fix the prompt regressions on bool/int with no defaults.

@Menghua1 , let's use this PR for only adding support for types in the default azd metadata field.  
Please rebase on top of main and use this PR to allow something like

image


Instead of the current:

image


Similar for `int` params, we should support numbers instead of asking for string only.  It's not a high priority, so no need to rush.

Menghua1:
@vhvb1989 Got it. The code has been updated to only add support for types in the default azd metadata field.
The PR description has also been updated, Please review again, thank you!

cli/azd/pkg/infra/provisioning/bicep/prompt.go
```diff
@@ -374,11 +375,23 @@ func (p *BicepProvider) promptForParameter(
...
+				case bool:
+					if v {
+						defaultValueForPrompt = "True"
+					} else {
+						defaultValueForPrompt = "False"
+					}
+				default:
+					return nil, fmt.Errorf("unsupported default value type %T for bool parameter: %v", v, key)
+				}
+			}
```
vhvb1989: instead of trying to parse to `bool`, just start by assigning `defaultValueForPrompt = "True"` , then check if  `strings.ToLower(defaultValueForPrompt) != "true"`  and if yes, change `defaultValueForPrompt  = "False"`.

That will allow people to set `default: true` or `default: "TRUE"` or `default: 0` and AZD will work with no error. Basically anything that is not converted to "true" will set the default to false.

Menghua1: @vhvb1989 The code has been updated as suggested. Please review again, thanks!

JeffreyCA (approved):
Are there any tests that can be added or updated here?

vhvb1989 (approved):

vhvb1989:
> Are there any tests that can be added or updated here?

Added coverage for the added code in prompt_test.go @JeffreyCA .  It will require your approval again

JeffreyCA (approved):

PR #5405, opened 2025-06-25
[EngSys] Migrate to azsdk-pool*

danieljurek:
Migrate from old pools to new `azure-sdk*` pools.


eng/pipelines/templates/stages/publish.yml
```diff
@@ -80,8 +80,8 @@ stages:
           environment: package-publish
 
           pool:
-            name: azsdk-pool-mms-win-2022-general
-            image: azsdk-pool-mms-win-2022-1espt
+            name: azsdk-pool
+            image: windows-2022
```
vhvb1989, line 84: When do we use `demands: ImageOverride -equals` ?  Example: https://github.com/Azure/ai-app-templates/pull/560/files#diff-d0115ae8ee36665b7c95c552d5364b0623c7db2a4d27ec61436c5caa933cee8dR17

Wondering about why it is different ???   @mikeharder

mikeharder: If you are using a 1ES template, then you specify `pool.name` and `pool.image`, and the 1ES template handles the rest.

If you are directly requesting an agent from our 1ES pool, you need to use `demands`.

Many repos use some of both:

https://github.com/search?q=repo%3AAzure%2Fazure-rest-api-specs%20ubuntu-24.04&type=code

danieljurek: When to use `demands` vs `image` 
  * When using `Azure Pipelines` public hosted pool (mac images) use `image` 
  * When using 1ES Hosted pools (non-mac images) use `demands` 

OSVmImage is a misnomer that came with us all the way from when we first migrated off of the `Azure Pipelines` pool. In those days we used `pool` and `image` to configure the image. Eventually we switched to using pools with one image per-pool so the pool was something like `azsdk-pool-mms-ubuntu-2004` and the `image` property was always ignored in those cases. 

Fast forward to the present: We have one pool per-architecture (x86_64 vs arm64) and each pool has multiple images. The correct way to specify images is to use `demands`


This is also further complicated by our 1ES redirect which uses its own templates to properly specify image overrides. So you may see something like this in the context of a 1ES template:

```yaml
pool: 
  name: azsdk-pool
  image: ubuntu-22.04
  os: linux
```

In that case, the 1ES template is setting up the demands itself.

danieljurek:
This can merge. Overriding check enforcer

PR #5411, opened 2025-06-26
Adding llm package - first step for ai functionality

vhvb1989:
The llm package provides the core logic for AZD to auto-detect LLM configuration.

This PR introduces a new manager under an alpha feature switch which can be injected to any command.
When injected, a command can use it to get LLM config automatically and create an LLM client.


cli/azd/pkg/llm/ollama.go
```diff
@@ -0,0 +1,33 @@
+package llm
+
+import (
+	"log"
+	"os"
+
+	"github.com/tmc/langchaingo/llms/ollama"
+)
+
+func loadOllama() (InfoResponse, error) {
```
JeffreyCA, line 13: When I ran this without ollama installed I think it still silently succeeded. Any additional checks or error handling we can add here?

cli/azd/cmd/version.go
```diff
@@ -33,30 +35,39 @@ func newVersionFlags(cmd *cobra.Command, global *internal.GlobalCommandOptions)
...
 func (v *versionAction) Run(ctx context.Context) (*actions.ActionResult, error) {
 	switch v.formatter.Kind() {
 	case output.NoneFormat:
 		fmt.Fprintf(v.console.Handles().Stdout, "azd version %s\n", internal.Version)
+		time.Sleep(500 * time.Millisecond)
+		_, err := v.llmManager.Info(v.console.Handles().Stdout)
+		if err != nil {
+			return nil, fmt.Errorf("failed to get LLM info: %w", err)
+		}
+		fmt.Fprintf(v.console.Handles().Stdout, "\n")
```
JeffreyCA: What does the sleep do here? Also noticed `azd version` now shows this error if I don't have the alpha flag set:
image

cli/azd/pkg/llm/azure_openai.go
```diff
@@ -0,0 +1,75 @@
...
+
+	"github.com/azure/azure-dev/cli/azd/pkg/output/ux"
+	"github.com/tmc/langchaingo/llms/openai"
+)
+
+const (
+	modelEnvVar   = "AZD_AZURE_OPENAI_MODEL"
+	versionEnvVar = "AZD_AZURE_OPENAI_VERSION"
+	urlEnvVar     = "AZD_AZURE_OPENAI_URL"
+	keyEnvVar     = "OPENAI_API_KEY"
```
JeffreyCA, line 19: Should this also be namespaced under `AZD_OPENAI_API_KEY`?

cli/azd/cmd/version.go
vhvb1989: oh, I shouldn't have added the changes to azd version XD.
That's just how I was testing. I will remove it

cli/azd/pkg/llm/azure_openai.go
vhvb1989: Since it is an already popular and known ENV VAR people use, I just use it like that.

We should be able to rename and fine-tune everything before switching from Alpha to Beta

cli/azd/pkg/llm/ollama.go
vhvb1989: Not that I know.  
The error would not happen until you try to use the LLM client.

It's not the best UX right now, but I didn't find some sort of ping or handshake to validate the config.  Hopefully this can be improved in the future.

cli/azd/pkg/llm/azure_openai.go
```diff
@@ -0,0 +1,78 @@
...
+	modelEnvVar:   {name: modelEnvVar},
+	versionEnvVar: {name: versionEnvVar},
+	urlEnvVar:     {name: urlEnvVar},
+	keyEnvVar:     {name: keyEnvVar},
+}
+
+func loadAzureOpenAi() (InfoResponse, error) {
+
+	envVars := maps.Clone(requiredEnvVars)
+	hasMissing := false
```
hemarina: Just curious, why do we use `hasMissing` instead of ` missingEnvVars`? For example,
```
    missingEnvVars := []string{}

    for name, envVar := range envVars {
        if value, isDefined := os.LookupEnv(envVar.name); isDefined {
            envVar.value = value
            envVar.isDefined = true
            envVars[name] = envVar
        } else {
            missingEnvVars = append(missingEnvVars, envVar.name)
        }
    }

    if len(missingEnvVars) > 0 {
        return InfoResponse{}, fmt.Errorf(
			"missing required environment variable(s): %s", ux.ListAsText(missingEnvVars))
    }
```

vhvb1989: Good catch. :)

JeffreyCA (approved):

PR #5434, opened 2025-07-03
Handle unrecognized expressions in Aspire Manifest

vhvb1989:
This change lets an `evalExpr` function to return a new error `UnrecognizedExpressionError` as a way to skip trying to evaluate such expression.

Unrecognized expressions are equivalent to skipped expressions, like

{{ skippedExpression }} == { unrecognizedExpression }

Using unrecognized expressions allows using non-escaped expressions in the manifest.


cli/azd/pkg/apphost/eval.go
```diff
@@ -30,6 +43,12 @@ func EvalString(src string, evalExpr func(string) (string, error)) (string, erro
 			for j := i + 1; j < len(src); j++ {
 				if src[j] == '}' {
 					v, err := evalExpr(src[i+1 : j])
+					if errors.Is(err, UnrecognizedExpressionError{}) {
```
hemarina, line 46: Could we add a test for `UnrecognizedExpressionError` case in eval_test.go?

eerhardt:
I tested this change with the Yarp playground in Aspire - https://github.com/dotnet/aspire/tree/main/playground/yarp and the catch-all routes work correctly for me.

cli/azd/pkg/apphost/eval.go
```diff
@@ -4,16 +4,29 @@
...
 // in the string.  It supports strings that contain expressions of the form "{expression}" where "expression" is any string
 // that does not contain a '}' character.  The evalExpr function is called with the expression (without the enclosing '{'
 // and '}' characters) and should return the value to be substituted into the string.  If the evalExpr function returns
 // an error, evalString will return that error. The '{' and '}' characters can be escaped by doubling them, e.g.
 // "{{" and "}}". If a string is malformed (e.g. an unmatched '{' or '}' character), evalString will return an error.
+// evalExpr can return error: UnrecognizedExpressionError to indicate that the expression could not be recognized. This
+// means there was not an error evaluating the expression, but rather that the evalExpr function has no idea how to handle
+// such expression. In this case, evalString will consider the expression as if it was skipped with '{{' and '}}' characters,
+// returning the full expression as is. For example, finding "{unknown}" will return "{unknown}" as part of the result.
```
weikanglim, line 29: I suspect that the ideal fix is to escape user provided input up higher in the stack, maybe that was the proposed alternative in https://github.com/dotnet/aspire/pull/10251.

Will "silent failure" as a way to handle unescaped user input end up being worse for us later? I'm happy with whatever you decide here -- you have a lot more context here.

PR #5446, opened 2025-07-08
bicep: cleanup + fix bicepparam evaluation

### LINKED ISSUE #4402: bicep: main.bicepparam environment variables being evaluated too early
With #4363, we fixed an issue with the main.parameters.json being prompted early in the flow and being cached. This caused environment variables that typically would be evaluated at deployment time, to be evaluated and its results cached early in the process.

The fix did not address the same bug for main.bicepparam. This is due to how main.bicepparam is currently designed -- it is  currently cached as part of the result of `compileBicepMemoryCache` in `bicep_provider.go`.

We should fix this. Likely, we would need to end up splitting out the compile bicep logic from compile parameters logic. This would introduce a perf hit of a few seconds since we need to invoke `bicep build` twice -- but still worth separating for correctness and completeness.

weikanglim:
Light refactoring to replace functions for retrieving module path and bicep module with upfront initialization.

Force a full `bicep build-params` to happen right before deploy. This ends up being the "right behavior" that matches closer to azd's usage of environment variables to be injected right before deployment occurs.

Fixes #4402

### DIFF
--- a/cli/azd/pkg/infra/provisioning/bicep/bicep_provider.go
+++ b/cli/azd/pkg/infra/provisioning/bicep/bicep_provider.go
@@ -40,7 +40,6 @@ import (
 	"github.com/azure/azure-dev/cli/azd/pkg/output/ux"
 	"github.com/azure/azure-dev/cli/azd/pkg/password"
 	"github.com/azure/azure-dev/cli/azd/pkg/prompt"
-	"github.com/azure/azure-dev/cli/azd/pkg/tools"
 	"github.com/azure/azure-dev/cli/azd/pkg/tools/bicep"
 	"github.com/drone/envsubst"
 )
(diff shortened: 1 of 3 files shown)


cli/azd/pkg/infra/provisioning/bicep/bicep_provider.go
```diff
@@ -50,61 +49,75 @@ const (
...
-	console               input.Console
-	bicepCli              *bicep.Cli
-	azapi                 *azapi.AzureClient
-	resourceService       *azapi.ResourceService
-	deploymentManager     *infra.DeploymentManager
-	prompters             prompt.Prompter
-	curPrincipal          provisioning.CurrentPrincipalIdProvider
+	projectPath           string
+	path                  string
+	mode                  bicepFileMode
```
vhvb1989, line 65: IIRC, we would try our best and avoid bringing state to components, usually preferring to run and re-calculate things multiple times. Are we leaving that as a Matt E era's memory? XD

weikanglim: In my mind, they aren't state, but are input parameters that are unchanged throughout the struct lifetime.

What is your preference?

I do prefer cementing the idea that `BicepProvider` operates on a given "module path" that remain `const` throughout the lifetime. After construction, the component wouldn't need to recalculate `modulePath()` or `isBicepFile(modulePath())` repeatedly, as these values are truly `const`. And in fact, if these values did change -- it would signal a bug in the underlying code.

weikanglim: The other part I liked about the changes is that it also cements more clearly the idea that bicepProvider indeed handles both bicep and bicepparam (with 90% shared codepaths), and they may warrant different handling due to differences in the underlying system. It didn't feel right for me to add another inline `isBicepParam` check.

vhvb1989: I think the change is good and better than what we have :) 

If you want to make read-only input parameters, you would need to inject the mode using the constructor. And I would still think about it as a component with state. XD

That been said, I do like stateful components when it makes sense, and I am happy to see this change :)

PR #5475, opened 2025-07-11
test the openai template for azure app services

jannette-wang:


vhvb1989:
Hi @jannette-wang 
Can you provide some details about what you are trying to do here?  Or close this PR if you don't need it anymore.
Thank you

PR #5476, opened 2025-07-11
[DISCUSS] Feat/new sdk provider for azure sdk

anfibiacreativa:
Opening this draft PR for discussion


weikanglim:
@anfibiacreativa Thanks for submitting this working draft to start the discussion! Some honest thoughts of mine below:

I am very much in the favor of moving infrastructure provisioning logic into the programming language that the user's application is coded in. In my opinion, it's a better DX for everyone to be able to write infrastructure with the language they already know of; and it also creates a highly composable and flexible workflow.

Some current examples to back these claims: In the IaC world, we see this emerging trend with Pulumi, Aspire. In the programming languages world, we see newer languages like Rust and Zig that allow build scripts defined in the same language. If I needed to take a big bet on the next emerging technology around infra provisioning for simplification, this would be the direction I'd push towards. 

Some technical details to consider for overall completenes:
1. The resources created should be tracked in some form, to enable a more native `azd down` experience. I understand that in the current change, we support a basic `azd down` experience by deleting the resource group, similar to how native azd does it before enabling `deploymentStacks`, but speaking from experience, this needs to be more expansive for a longer runway. 
2. Relatedly, a common feature that users will want is some form of `--preview` experience. More importantly, it needs to support a GitOps workflow, where the changes are projected onto filesystem files, tracked in source control, before being applied by a separate operator.

On the project side of things: pushing for something in this direction would be a sizeable undertaking to say the least. The biggest challenge is momentum and volume in user adoption. To that end, I would say that building a language provider on top of current-day azd constructs may limit the opportunity in some ways. To shoot for the moon, one needs to break the ground significantly to say the least.

PR #5492, opened 2025-07-17
provisioning: allow for provisioning stages

weikanglim:
This change allows for additional provisioning stages to be defined in azure.yaml as `infra.stages`.

The following commands have been updated to reflect this change:

- `provision [stage]` -- Provisioning now accepts a stage name
- `env refresh` -- Refresh state from all stages by default
- `env refresh --stage [stage]` -- Refresh from a stage
- `down` -- Delete resources from all stages by default
- `down --stage [stage]` -- Delete resources from a stage 

Note: Behavior of`pipeline config` is left unchanged; and will require separate treatment in a follow-up change.


cli/azd/internal/cmd/provision.go
```diff
@@ -92,13 +92,17 @@ func NewProvisionFlagsFromEnvAndOptions(envFlag *internal.EnvFlag, global *inter
 }
 
 func NewProvisionCmd() *cobra.Command {
-	return &cobra.Command{
-		Use:   "provision",
+	cmd := &cobra.Command{
+		Use:   "provision [stage]",
```
JeffreyCA: What are your thoughts on having an `--all` flag to be able to provision all stages, similar to how `azd deploy --all` deploys all services?

cli/azd/pkg/infra/provisioning/bicep/bicep_provider.go
```diff
@@ -598,7 +605,8 @@ func (p *BicepProvider) Deploy(ctx context.Context) (*provisioning.DeployResult,
 	}
 
 	deploymentTags := map[string]*string{
-		azure.TagKeyAzdEnvName: to.Ptr(p.env.Name()),
+		azure.TagKeyAzdEnvName:   to.Ptr(p.env.Name()),
+		azure.TagKeyAzdStageName: &p.stage,
```
JeffreyCA: What value are we setting the `azd-stage-name` tag to for the "default" stage (the one provisioned by running `azd provision`)?

Just trying to understand how this works with the changes in `cli/azd/pkg/infra/deployment_manager.go`

cli/azd/internal/cmd/provision.go
wbreza: For this concept I believe stage is referring to things like `dev`, `test`, `prod`, etc.  It would be weird for an `azd provision` command to provision resources for all of these stages in a given command.

cli/azd/cmd/env.go
```diff
@@ -929,12 +929,14 @@ func (en *envNewAction) Run(ctx context.Context) (*actions.ActionResult, error)
...
 type envRefreshFlags struct {
 	hint   string
+	stage  string
 	global *internal.GlobalCommandOptions
 	internal.EnvFlag
 }
 
 func (er *envRefreshFlags) Bind(local *pflag.FlagSet, global *internal.GlobalCommandOptions) {
 	local.StringVarP(&er.hint, "hint", "", "", "Hint to help identify the environment to refresh")
+	local.StringVarP(&er.stage, "stage", "", "", "Stage to refresh the environment from.")
```
wbreza: How is an environment associated with a stage? Is it just the usage of the `--stage` attribute during provision commands right now?

wbreza (commented):
This is a good start but have a couple questions:

In your example you show `app` as a stage.  The term `app` feels like a layer to me and tags like dev, test & prod seem more appropriate as a stage.

Are we mixing the 2 constructs? Let's discuss.

puicchan:
1. Stage should be renamed to layer. Based on GHCP, phase is more idiomatic and common in  Azure docs and Azure ecosystem.
2. UX: for "Stage: infra" consider calling "infra" as base or core? Can be confusing as Bicep for my 2nd layer `eventgrid` is under /infra.
3. Should support recordgroup targetscope? My eventgrind.bicep original scope is recordgroup and had to rewrite to use subscription scope. Less rewrite is better.

cli/azd/internal/cmd/provision.go
weikanglim: Renamed to layers.

To address the original question, `--all` does make sense, but I'd like to hold until we receive a request.

`--all` may differ than the expected `up` workflow definition where additional steps are interleaved, thus great for an ad-hoc test, but perhaps not the highest utility.

cli/azd/pkg/infra/provisioning/bicep/bicep_provider.go
weikanglim: It would be set as "default", whatever the default value is.

On matching, the new deployment would match on tags. However, for older deployments already out there, we'd match based on env name alone.

weikanglim:
@puicchan @wbreza 

Latest changes are renamed to `layers`. A pretty good AI summary attached below on trade-offs, which I do see it as a balanced take.

---

# Stages vs. Layers

## 🔁 **Stages** – Imply sequencing

- **Connotation**: Ordered steps in a process (e.g., "provision storage, then roles").
- **Good for**: Progressive deployments, CI/CD pipelines, partial updates.
- **Typical user question**: *“What’s the next stage I need to deploy?”*

### ✅ Pros
- Familiar term in DevOps workflows (e.g., *build*, *test*, *deploy*).
- Encourages clean separation and partial provisioning.

### ⚠️ Cons
- Implies a linear dependency, even if stages are independent.
- Can confuse users if stages are actually parallel or optional.

---

## 🧱 **Layers** – Imply structure

- **Connotation**: Stackable, composable units (e.g., networking layer, security layer).
- **Good for**: Modular infra design, shared components, composability.
- **Typical user question**: *“What layers make up my infrastructure?”*

### ✅ Pros
- Better metaphor for independent or reusable infra modules.
- Neutral to order — avoids assumptions about execution flow.

### ⚠️ Cons
- Less common in deployment tooling vocabulary.
- May sound like a lower-level abstraction to non-infra folks.

---

## 🧠 Which to choose?

| Use Case                          | Recommended Term |
|----------------------------------|------------------|
| CI/CD pipelines                  | `stages`         |
| Modular, reusable infra blocks   | `layers`         |
| Explicit control over order      | `stages`         |
| Parallel or optional components  | `layers`         |

weikanglim:
@puicchan 

> UX: for "Stage: infra" consider calling "infra" as base or core? Can be confusing as Bicep for my 2nd layer eventgrid is under /infra.

Switched it to "base". I personally find "infra" to be the least confusing of all possible options.

> Should support recordgroup targetscope? My eventgrind.bicep original scope is recordgroup and had to rewrite to use subscription scope. Less rewrite is better.

We'll want to do this -- but as a separate scoped item to address multiple resource groups; currently, this requires breaking through specific azd built walls in our codebase.

cli/azd/internal/cmd/provision.go
```diff
@@ -199,7 +205,26 @@ func (p *ProvisionAction) Run(ctx context.Context) (*actions.ActionResult, error
...
+
 	infraOptions := infra.Options
+	if layer != "" {
+		layerOption, err := infraOptions.GetLayer(layer)
+		if err != nil {
+			return nil, err
+		}
+
+		infraOptions = layerOption
+	}
```
JeffreyCA: If I have my infra set up like this, I currently get this error:

> ERROR: error executing step command 'provision layer1': layer 'layer1' not found in azure.yaml. available layers:

Seems like a base layer is required?

```
infra
├── abbreviations.json
├── layer-1
│   ├── main.bicep
│   └── main.parameters.json
├── layer-2
│   ├── main.bicep
│   └── main.parameters.json
├── layer-3
│   ├── main.bicep
│   └── main.parameters.json
└── resources
```

azure.yaml:

```yaml
infra:
  layers:
  - name: layer1
    path: ./infra/layer-1
  - name: layer2
    path: ./infra/layer-2
  - name: layer3
    path: ./infra/layer-3
workflows:
  up:
   steps:
    - azd: provision layer1
    - azd: provision layer2
    - azd: provision layer3
```

JeffreyCA: @weikanglim can you try to `azd up` [this sample](https://github.com/JeffreyCA/azd-dev-prod-aca/blob/layer-err/azure.yaml)?

I get this error:

```
$ azd up
? Select an Azure Subscription to use: 25. XXX
? Select an Azure location to use: 50. (US) East US 2 (eastus2)
Note: Running custom 'up' workflow from azure.yaml

ERROR: error executing step command 'provision base-layer': layer 'base-layer' not found in azure.yaml. available layers:
```

weikanglim: I see. It's happening in this case where `infra/main.bicep` isn't present. The [ProjectImporter](https://github.com/Azure/azure-dev/blob/7c1e561f34252b39d8acd455cc5b48c3ba29503b/cli/azd/pkg/project/importer.go#L169) was returning a completely empty config in this case. This isn't quite as expected, and I've fixed the logic here.

JeffreyCA: Thanks! With the new changes it detects the layers now

vhvb1989 (approved):

wbreza (approved):

rajeshkamal5050:
@weikanglim can we merge this?

ronaldbosma:
I’m curious if it’s possible to define hooks between layers. My use case is:
- Deploy a service principal and Key Vault in the first layer
- Run a hook to generate a client secret/certificate and store it in Key Vault
- Deploy a second layer that uses the secret from Key Vault

Is something like this supported?

Issue #5491, opened 2025-07-16
azd infra gen spins forever w/ .NET Web API project


cwiederspan:
- [X] Make sure you've installed the latest version using [instructions](https://github.com/Azure/azure-dev?tab=readme-ov-file#installupgrade-azure-developer-cli)

**Output from `azd version`**
Run `azd version` and copy and paste the output here: azd version 1.17.2 (commit c241503424c35882c04d9a909ff9bb1469cd68c8)

**Describe the bug**
`azd infra gen` does not seem to run from within a GitHub Codespace and simply spins forever on the "Analyzing Aspire Application" step.

**To Reproduce**
- Create an empty GitHub Codespace
- Install Azure CLI
- mkdir src
- cd src
- dotnet new webapi -n Something
- cd ..
- azd init
- azd infra gen

At this point, it pops open the forwarded ports popup with the green "Open in Browser" button, and the console shows "Analyzing Aspire Application", but it just sits there and spins forever.

**Expected behavior**
This should generate the ./infra folder with the bicep files.

**Environment**
Information on your environment:
    * Language name and version
    * IDE and version : Running Linux Codespace in VS Code (latest version)

**Additional context**
Filed by MS FTE - Alias CHWIEDER

weikanglim:
@cwiederspan As a preliminary step for debugging, would you mind trying to run the following command and sharing the results?

```bash
azd infra gen --debug 2>debug.log
```

This will generate a debug log that might help us pinpoint where the issue is occurring. Based on the behavior you're seeing—particularly the "Analyzing Aspire Application..." message—I suspect that azd may be getting stuck while invoking a `dotnet publish` process under the hood.

cwiederspan:
I ran that and there's not much in there. Here's what the Debug.log file looks like...

2025/07/17 12:18:01 main.go:56: azd version: 1.17.2 (commit c241503424c35882c04d9a909ff9bb1469cd68c8)
2025/07/17 12:18:01 main.go:214: using cached latest version: 1.17.2 (expires on: 2025-07-17T21:42:08Z)
2025/07/17 12:18:01 project.go:151: Reading project from file '/workspaces/junk_20250716/azure.yaml'
2025/07/17 12:18:01 middleware.go:111: running middleware 'debug'
2025/07/17 12:18:01 middleware.go:111: running middleware 'ux'
2025/07/17 12:18:01 middleware.go:111: running middleware 'telemetry'
[...]

PR #5501, opened 2025-07-17
fix: `infra gen` when dotnet project is present

weikanglim:
This change adds missing logic to check if a dotnet project is an Aspire project before attempting to run the `dotnet run` command.

Before this change, when users attempted to run `infra gen` with a dotnet project present; the spawned `dotnet run` command would end up running the application itself, which results in unexpected behavior such as azd hanging while waiting for the application to return.

Fixes: #5491

### DIFF
--- a/cli/azd/pkg/project/importer.go
+++ b/cli/azd/pkg/project/importer.go
@@ -189,11 +189,20 @@ func pathHasModule(path, module string) (bool, error) {
 func (im *ImportManager) GenerateAllInfrastructure(ctx context.Context, projectConfig *ProjectConfig) (fs.FS, error) {
 	for _, svcConfig := range projectConfig.Services {
 		if svcConfig.Language == ServiceLanguageDotNet {
-			if len(projectConfig.Services) != 1 {
-				return nil, errNoMultipleServicesWithAppHost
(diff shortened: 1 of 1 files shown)


JeffreyCA (commented):
How come in the issue the output of the `dotnet run` command wasn't being logged when running `azd infra gen` with `--debug`? Would that be easy to add for better debuggability going forward?

weikanglim:
@JeffreyCA Good question.

Right now our command runner only logs the full output of the command after execution:

> 2025/07/17 19:22:52 command_runner.go:310: Run exec: 'git --version' , exit code: 0
-------------------------------------stdout-------------------------------------------
git version 2.45.3

A possible enhancement is to split this out into two messages:

> 2025/07/17 19:22:52 command_runner.go:310: Start: 'git --version'
> 2025/07/17 19:22:52 command_runner.go:310: Stop: 'git --version' , exit code: 0
-------------------------------------stdout-------------------------------------------
git version 2.45.3

which would get us the logs we wanted.

Happy to take an issue on it.

JeffreyCA (approved):
Thanks for the explanation! Created #5502

rajeshkamal5050 (approved):

PR #5513, opened 2025-07-22
Adding MCP - test server

vhvb1989:
Adding a simple MCP server as part of the azd-llm alpha feature.

This server becomes useful to expose azd mcp tools and test mcp features like mcp-sampling with other clients like VSCode gh copilot.


cli/azd/cmd/mcp.go
```diff
@@ -0,0 +1,96 @@
...
+	"github.com/azure/azure-dev/cli/azd/pkg/alpha"
+	"github.com/azure/azure-dev/cli/azd/pkg/exec"
+	"github.com/azure/azure-dev/cli/azd/pkg/input"
+	"github.com/azure/azure-dev/cli/azd/pkg/llm/tools"
+	"github.com/mark3labs/mcp-go/server"
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+)
+
+func newMcpFlags(cmd *cobra.Command, global *internal.GlobalCommandOptions) *mcpFlags {
```
wbreza, line 21: Can we make the command `azd mcp start` ? Gives us the opportunity to add other commands under `mcp` if needed.

We also should consider making the command `hidden` since it is not really a core `azd` command and would only be used through AI scenarios.

cli/azd/pkg/llm/manager.go
```diff
@@ -17,6 +17,17 @@ import (
 
 var featureLlm = alpha.MustFeatureKey("llm")
 
+func IsLlmFeatureEnabled(alphaManager *alpha.FeatureManager) error {
```
wbreza, line 20: Can we call it the `ai` feature instead of LLM?

wbreza (approved):
Looks great! Just a couple suggestions.

cli/azd/cmd/mcp.go
vhvb1989: Can we leave it as "azd mcp" for now?   Having "azd mcp start" makes me think there should be one "azd mcp stop".
This is still under an alpha feature.

It should not be hidden. It should be documented for any host to use it.

cli/azd/pkg/llm/manager.go
vhvb1989: I've been thinking about "agentic" as the top concept here. I intentionally avoided "ai" because it is kind of overloaded already (extension, composability, azure service...)

cli/azd/pkg/llm/tools/hello.go
```diff
@@ -0,0 +1,98 @@
...
+}
+
+func helloHandler(ctx context.Context, request mcp.CallToolRequest) (*mcp.CallToolResult, error) {
+	name, err := request.RequireString("name")
+	if err != nil {
+		return mcp.NewToolResultError(err.Error()), nil
+	}
+
+	// Get the client session from context
+	session := server.ClientSessionFromContext(ctx)
```
hemarina, line 40: If we build more tools, would this part of the code be being repetitive like we may create a function for this logic?

vhvb1989: We'll do it for sure if we observe that while adding more tools

vhvb1989:
> When I run `azd mcp`, it didn't respond and hang there forever. Is that expected? I thought it will use helloHandler to print out greetings.

Yes, that's expected. It is not hang, it is waiting to get a request from a client.

cli/azd/cmd/mcp.go
```diff
@@ -0,0 +1,96 @@
...
+
+type mcpFlags struct {
+	global *internal.GlobalCommandOptions
+	internal.EnvFlag
+}
+
+func (i *mcpFlags) Bind(local *pflag.FlagSet, global *internal.GlobalCommandOptions) {
+	i.EnvFlag.Bind(local, global)
+	i.global = global
+}
```
JeffreyCA, line 43: Just wondering, how is the MCP command going to use the environment flag?

cli/azd/cmd/mcp.go
JeffreyCA, line 1: I've configured my `mcp.json` like this:

```json
{
	"servers": {
		"azd-mcp": {
			"type": "stdio",
			"command": "azd",
			"args": [
				"mcp"
			]
		}
	},
	"inputs": []
}
```

When using **Gemini 2.5 Pro (Preview)** for some reason, it gets stuck in a never-ending loop:
Message: "Say Hello to Copilot"

image


Is that an issue with the MCP server? Or something else?

vhvb1989: Seems like an issue with Github copilot as it keeps calling the tool again and again XD.

cli/azd/cmd/mcp.go
vhvb1989: We haven't decide if:
- You need to start the mcp server with --cwd and optionally a env name so the server is scoped to an specific project
or
- Each tool needs to get the path and env name from the mcp-client (agent must get it and send it all the time)
or
- Tools should never depend on azd specific paths and should do and return work which the agent can apply to the current project.

Depending on that, we would either remove this env or do something else.
Right now I just put it there as I was copy/pasting from another command XD

SophCarp:
I'm a little confused what the command `azd mcp` does here. When I run the command, it doesn't look like the MCP server starts up. Does the user ever need to run `azd mcp`?

If `azd mcp` does start the server running, I think the command should be `azd mcp start` or `azd mcp run`. "mcp" is a noun, and we usually use them as subcommands rather than commands. IT also gives room for future commands within mcp.

cli/azd/cmd/mcp.go
SophCarp: Should there be an `azd mcp stop` command?

vhvb1989:
> I'm a little confused what the command `azd mcp` does here. When I run the command, it doesn't look like the MCP server starts up. Does the user ever need to run `azd mcp`?
> 
> If `azd mcp` does start the server running, I think the command should be `azd mcp start` or `azd mcp run`. "mcp" is a noun, and we usually use them as subcommands rather than commands. IT also gives room for future commands within mcp.

@SophCarp  I was really trying to avoid "start" or "run" because there will never be "stop".  I think something like "serve" would work if we really want to force the pattern "azd <grouper> <verb>" (which was broken for 'azd provision/up/down/deploy` anyway)
I don't see any other "verb" comming in a near future to "azd mcp", and even if it comes, we can make it work. We can support a command to be a grouper and a command at the same time :)

I'm speaking here with some context for how "azd pipeline config" was created expecting other verbs in the future. It's been years already since then and there's nothing even planned for "azd pipeline <other-verb>" 😊.  So, I wish, for the number of times I've written "azd pipeline config", that it could've just been "azd pipeline"

cli/azd/cmd/testdata/TestUsage-azd-mcp.snap
JeffreyCA, line 1: This doesn't seem to match the actual output. Do we need to update this?

```
Starts the azd MCP server.

Usage
  azd mcp [flags]

Flags
    -e, --environment string    : The name of the environment to use.

Global Flags
    -C, --cwd string    : Sets the current working directory.
        --debug         : Enables debugging and diagnostics logging.
        --docs          : Opens the documentation for azd mcp in your web browser.
    -h, --help          : Gets help for mcp.
        --no-prompt     : Accepts the default value instead of prompting, or it fails if there is no default.
```

cli/azd/cmd/root.go
```diff
@@ -414,6 +415,35 @@ func NewRootCmd(
...
+			})
+		} else {
+			root.Add("mcp", &actions.ActionDescriptorOptions{
+				Command:        newMcpCmd(),
+				FlagsResolver:  newMcpFlags,
+				ActionResolver: newMcpAction,
+				HelpOptions: actions.ActionHelpOptions{
+					Description: getCmdMcpHelpDescription,
+					Footer:      getCmdMcpHelpFooter,
+				},
```
JeffreyCA, line 436: I think we may not need this and can remove `getCmdMcpHelpDescription` and `getCmdMcpHelpFooter` since we don't have extra description or footer content, like "add" command for example

JeffreyCA (approved):

Issue #5277, opened 2025-06-08
Display a deprecate notice for Aspire projects using project.v1 w/o deployment field


vhvb1989:
When deployment field is not provided, AZD generate YAML files for deploying services.  This mode is being deprecated.
For this issue, if the deployment field is not defined in the manifest, AZD should display a warning about deprecation and
some notes (probably a URL) for how to migrate/update their AppHost.

davidfowl:
https://github.com/dotnet/docs-aspire/issues/3669

PR #5518, opened 2025-07-23
Add suggestion text for Container App unmarshal errors during `azd deploy`

### LINKED ISSUE #5109: azd deploy do not use the same API version than defined in the BICEP file for Microsoft.App/containerApps
Hello,

i'm using a preview, fresh release of the resource containerApp 'Microsoft.App/containerApps@2024-10-02-preview' that have a new value for the BindingType information (for customDomains)

but it failed to deploy my application with this error : 

ERROR: failed deploying service 'chat-app': updating container app service: getting container app: getting container app: unmarshalling type *armappcontainers.ContainerApp: unmarshalling type *armappcontainers.ContainerApp: struct field Properties: unmarshalling type *armappcontainers.ContainerAppProperties: struct field Configuration: unmarshalling type *armappcontainers.Configuration: struct field Ingress: unmarshalling type *armappcontainers.Ingress: struct field CustomDomains: unmarshalling type *armappcontainers.CustomDomain: struct field BindingType: json: **cannot unmarshal number** into Go value of type armappcontainers.BindingType

For some reason the BindingType is a number :-(

When i use azd deploy --debug i can see that it GET the container information/data from a very old version : 
 --> providers/Microsoft.App/containerApps/my-app?api-version=2023-11-02-preview

and of course de bindingType value is not good (i can see value number 2)

When i use CURL to manually get the information of the containerApp with the correct API version BindingType is visible as a string with value "Auto"

My question : should AZD use what is defined in the BICEP to get information of containerApp API version instead of some kind of "hard coded" version ? And today how can i force azd deploy to use the latest API version for my deployment ?

thanks a lot in advance,

PS : the new API version provide a new BindingType value to manage certificate and customDomains --> https://github.com/microsoft/azure-container-apps/issues/796

JeffreyCA:
Resolves #5109

This PR adds actionable suggestion text for unmarshalling errors encountered when deploying to Container Apps:
image

This is a known limitation that we addressed in #4272 with the `apiVersion` service config property.

These changes also handle the Aspire (YAML mode) case, where azd GETs the ACA resource to persist custom domain and session affinity configuration.

### DIFF
--- a/cli/azd/pkg/containerapps/container_app.go
+++ b/cli/azd/pkg/containerapps/container_app.go
@@ -7,10 +7,12 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"log"
 	"net/http"
 	"slices"
+	"strings"
 
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/arm"
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/policy"
@@ -18,6 +20,7 @@ import (
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/streaming"
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/to"
 	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/appcontainers/armappcontainers/v3"
+	"github.com/azure/azure-dev/cli/azd/internal"
 	"github.com/azure/azure-dev/cli/azd/pkg/account"
 	"github.com/azure/azure-dev/cli/azd/pkg/alpha"
 	"github.com/azure/azure-dev/cli/azd/pkg/config"
@@ -148,6 +151,10 @@ func (cas *containerAppService) persistSettings(
 
 	aca, err := cas.getContainerApp(ctx, subscriptionId, resourceGroupName, appName, options)
 	if err != nil {
+		var errWithSuggestion *internal.ErrorWithSuggestion
+		if errors.As(err, &errWithSuggestion) {
+			return nil, err
+		}
 		log.Printf("failed getting current aca settings: %v. No settings will be persisted.", err)
 		// if the container app doesn't exist, there's nothing for us to update in the desired state,
 		// so we can just return the existing state as is.
@@ -249,7 +256,7 @@ func (cas *containerAppService) DeployYaml(
 
 		var containerApp armappcontainers.ContainerApp
 		if err := json.Unmarshal(containerAppJson, &containerApp); err != nil {
-			return fmt.Errorf("converting to container app type: %w", err)
+			return withApiVersionSuggestion(fmt.Errorf("converting to container app type: %w", err))
 		}
 
 		p, err := appClient.BeginCreateOrUpdate(ctx, resourceGroupName, appName, containerApp, nil)
@@ -458,7 +465,11 @@ func (cas *containerAppService) getContainerApp(
 
 	_, err = appClient.Get(ctx, resourceGroupName, appName, nil)
 	if err != nil {
-		return nil, fmt.Errorf("getting container app: %w", err)
+		err = fmt.Errorf("getting container app: %w", err)
+		if strings.Contains(err.Error(), "unmarshalling type") {
+			err = withApiVersionSuggestion(err)
+		}
+		return nil, err
 	}
 
 	var containAppMap map[string]any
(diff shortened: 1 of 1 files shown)


cli/azd/pkg/containerapps/container_app.go
```diff
@@ -148,6 +151,10 @@ func (cas *containerAppService) persistSettings(
 
 	aca, err := cas.getContainerApp(ctx, subscriptionId, resourceGroupName, appName, options)
 	if err != nil {
+		var errWithSuggestion *internal.ErrorWithSuggestion
+		if errors.As(err, &errWithSuggestion) {
+			return nil, err
+		}
 		log.Printf("failed getting current aca settings: %v. No settings will be persisted.", err)
```
weikanglim: I noticed that @JeffreyCA had to defensively code around this.

The underlying issue is that any errors happening in `getContainerApp` is being treated at this callsite as "the container app does not exist", which isn't quite the right. The errors happening in `getContainerApp` could include: auth failure, unrelated service errors, and in this case, json unmarshalling error.

As a general quality-of-life improvement, I'm thinking that `getContainerApp` should check for a 404 response explicitly, and return a specific error condition for `ErrContainerAppNotFound`. We would test specifically against this error here.

An example I picked up from somewhere in the codebase that we could apply to L468:

```golang
		var errDetails *azcore.ResponseError
		if errors.As(err, &errDetails) && errDetails.StatusCode == 404 {
			return nil, ErrDeploymentNotFound
		}
		return nil, fmt.Errorf("getting deployment from resource group: %w", err)

```

@vhvb1989 just checking if this is a good idea, or if we should keep the way it's written for other reasons.

vhvb1989: This is a good improvement, but I am afraid it will most likely never occur.

This function `persistSettings()` is only called by `DeployYaml()` which is only called by `service_target_dotnet_containerapp.go`, which is the implementation for deploying Aspire only projects.

When folks are working with Aspire, there is zero expectation about users manually changing/updating azure.yaml.  Actually, for Aspire, azure.yaml describes only the AppHost and not the list of each dotnet service.

So, having azd returning a recommendation for adding `apiVersion` to azure.yaml is actually not accurate for Aspire projects.

I recommend leaving the Aspire-deployment flow as it is because deploying YAML with Aspire is currently in deprecation mode and should be seen less and less as customers migrate to the latest API from Aspire.  We have this issue https://github.com/Azure/azure-dev/issues/5277 to display a warning for users who are still using YAML deployment in Aspire to ask them to migrate.
So, at some point, we will just remove the persistSettings form Aspire.

cli/azd/pkg/containerapps/container_app.go
```diff
@@ -603,3 +614,16 @@ func createApiVersionPolicy(options *ContainerAppOptions) *containerAppCustomApi
 		apiVersion: options.ApiVersion,
 	}
 }
+
+func withApiVersionSuggestion(err error) error {
+	suggestion := "Suggestion: set 'apiVersion' on your service in azure.yaml to match the API version " +
+		"in your IaC:\n\n" +
+		"services:\n" +
+		"  your-service:\n" +
+		"    apiVersion: 2025-02-02-preview"
```
weikanglim: Does it look better to make this a different color? Like green.. Same for the resource group

cli/azd/pkg/containerapps/container_app.go
```diff
@@ -249,7 +256,7 @@ func (cas *containerAppService) DeployYaml(
 
 		var containerApp armappcontainers.ContainerApp
 		if err := json.Unmarshal(containerAppJson, &containerApp); err != nil {
-			return fmt.Errorf("converting to container app type: %w", err)
+			return withApiVersionSuggestion(fmt.Errorf("converting to container app type: %w", err))
```
vhvb1989: we don't need this as DeployYaml is only for Aspire

cli/azd/pkg/containerapps/container_app.go
JeffreyCA: image
image
image

I can see grey potentially working, though green seems a little out of place to me.

@SophCarp do you have thoughts on the formatting or wording of the error and suggestion here? (I think @weikanglim is referring to using a different colour for the yaml snippet)

weikanglim: For just the highlighted line of change, i.e. `resourceGroup`. It helps make it stand out as the call to action.

vhvb1989 (approved):

cli/azd/pkg/containerapps/container_app.go
JeffreyCA: Oh sorry I misunderstood. That makes more sense:

image

weikanglim (approved):

vhvb1989 (approved):

PR #5521, opened 2025-07-23
agentic error handling workflow

### LINKED ISSUE #5465: Deliver AI-first(VSCode/MCP) error handling experiences
Deliver error handling/messaging improvements in AI-first way: 

- [x] Resolve and/or make errors actionable using MCP sampling in AZD for experiences delivered via VSCode/MCP
- [x] Explore and gracefully handle for native CLI errors i.e., non-MCP/LLM-less

hemarina:
fix #5465 #5597 

- [x] This feature is in alpha feature `llm`
- [x] Create `azd_error_troubleshooting` mcp tool to identify, explain and diagnose errors when running azd commands
- [x] Implement `error middleware`
- [x] Provide prompt to ask user to user agent model to get a troubleshooting guide or let agent to fix the error
- [x] Add prompt consent
- [x] azd command will re-run after error is fix
- [x] If same error happened more than 3 times, this means AI is not able to fix it. The error handling workflow will end by return the error with a message to ask user to fix the error manually
- [x] Improve UX with designer. 
- [x] Print file changes

For UX, I updated most of them. Some UX related to global change (i.e init workflow) will be discussed/revisited with designer and included in follow up PR. 

Example workflow:
image
image

Consent for prompt:
image


cli/azd/pkg/agentRunner/agentRunner.go
```diff
@@ -0,0 +1,332 @@
...
+	if finish.ReturnValues != nil {
+		log.Printf("   Final output: %v\n", finish.ReturnValues)
+	}
+}
+
+func (h *agentLogHandler) HandleLLMGenerateContentEnd(ctx context.Context, response *llms.ContentResponse) {
+	log.Println("✨ LLM content generation completed")
+}
+
+func promptingWithDifferentErrors(err error) string {
```
wbreza: Instead of dynamically changing the prompt here one suggestion would be to create an MCP tool called something like `azd_error_troubleshooting` or something.  The response of this error can be a straight forward makrdown document that describes well know types of errors.  Each error can provide some troubleshooting advise that the LLM can then use to fix the issue.

The initial prompt can also mention to use this tool to get troubleshooting feedback.  Let me know if you have questions.

cli/azd/internal/mcp/tools/prompts/azd_common_error.md
```diff
@@ -0,0 +1,14 @@
+# AZD Provision Common Error Resolution Instructions
```
hemarina: This might be removed and put into a follow-up PR.

cli/azd/internal/agent/conversational_agent.go
```diff
@@ -178,3 +206,95 @@ func (aai *ConversationalAzdAiAgent) renderThoughts(ctx context.Context) (func()
...
+
+func startWatcher(ctx context.Context, fileChanges *FileChanges, mu *sync.Mutex) (*fsnotify.Watcher, chan bool, error) {
+	watcher, err := fsnotify.NewWatcher()
+	if err != nil {
+		return nil, nil, fmt.Errorf("failed to create watcher: %w", err)
+	}
+
+	done := make(chan bool)
+
+	go func() {
```
hemarina: Please hold run test with this PR. The go routine is having issues about hanging at the background. I'm still debugging it.

cli/azd/cmd/middleware/error.go
```diff
@@ -0,0 +1,339 @@
...
+		if err != nil {
+			return nil, err
+		}
+
+		defer azdAgent.Stop()
+
+		for {
+			if originalError == nil {
+				break
+			}
```
wbreza: This check can happen higher up outside the loop and before agent creation.

cli/azd/cmd/middleware/error.go
```diff
@@ -0,0 +1,339 @@
...
+
+			for _, s := range skipAnalyzingErrors {
+				if strings.Contains(originalError.Error(), s) {
+					return actionResult, originalError
+				}
+			}
+
+			if previousError != nil && errors.Is(originalError, previousError) {
+				attempt++
+				if attempt > 3 {
```
wbreza: Nit: I would do `>= 3`

cli/azd/cmd/middleware/ux.go
```diff
@@ -37,7 +41,7 @@ func (m *UxMiddleware) Run(ctx context.Context, next NextFn) (*actions.ActionRes
 	// Stop the spinner always to un-hide cursor
 	m.console.StopSpinner(ctx, "", input.Step)
 
-	if err != nil {
+	if err != nil && !m.featuresManager.IsEnabled(llm.FeatureLlm) {
```
wbreza, line 44: Why are we adding LLM feature check here?

cli/azd/internal/agent/conversational_agent.go
```diff
@@ -178,3 +206,95 @@ func (aai *ConversationalAzdAiAgent) renderThoughts(ctx context.Context) (func()
 
 	return cleanup, canvas.Run()
 }
+
+func startWatcher(ctx context.Context, fileChanges *FileChanges, mu *sync.Mutex) (*fsnotify.Watcher, chan bool, error) {
```
wbreza: We should move the file watching semantics outside of the agent code then invoke it from here so this can be tested/used in isolation or in other areas.

We should also consider splitting out the rendering of the file changes from the `SendMessage(...)` call

cli/azd/cmd/middleware/error.go
wbreza: This could be the cause of the multiple `azd` processes.

cli/azd/cmd/middleware/ux.go
hemarina: If we don't check it, the error message will be printed out twice. One is in error middleware (before AI fix) and one in UX middleware (the error before the fix will be printed out after re-run command succeed).

cli/azd/internal/agent/conversational_agent.go
```diff
@@ -88,17 +91,36 @@ func NewConversationalAzdAiAgent(llm llms.Model, opts ...AgentCreateOption) (Age
 }
 
 // SendMessage processes a single message through the agent and returns the response
-func (aai *ConversationalAzdAiAgent) SendMessage(ctx context.Context, args ...string) (string, error) {
+func (aai *ConversationalAzdAiAgent) SendMessage(ctx context.Context, useWatch bool, args ...string) (string, error) {
```
wbreza: To enable optional features on the agent we need to follow the options model when creating the agent with the agent factory instead of using a boolean flag here.

cli/azd/pkg/watch/watch.go
```diff
@@ -0,0 +1,133 @@
...
+	"github.com/fsnotify/fsnotify"
+)
+
+type FileChanges struct {
+	Created  map[string]bool
+	Modified map[string]bool
+	Deleted  map[string]bool
+}
+
+func StartWatcher(ctx context.Context, mu *sync.Mutex) (*fsnotify.Watcher, chan bool, *FileChanges, error) {
```
wbreza: can we keep the implementation details hidden? Just start the watcher and have function to print changes... Consumer doesn't really need to know the internal details that its using recursive file watching, channels, etc...

Handle done/cleanup via context

wbreza (commented):
Looking great overall - just some feedback around the usage patterns that we should clean up before merging.

cli/azd/cmd/middleware/ux.go
wbreza: if the AI is able to fix the issue then there wouldn't be any error to show, right?

cli/azd/pkg/watch/watch.go
```diff
@@ -0,0 +1,142 @@
...
+	mu          sync.Mutex
+}
+
+type fileChanges struct {
+	Created  map[string]bool
+	Modified map[string]bool
+	Deleted  map[string]bool
+}
+
+func StartWatcher(ctx context.Context) (Watcher, error) {
```
wbreza: Do be more consistent with Go we should either standardize this to be a constructor function named `NewWatcher(...)` since it does create and return the new `fileWatcher`.  At that point we should also consider whether or not it automatically starts or if we need  Start/Stop functions.

For the short term i would just recommend we name it `NewWatcher()` and it automatically starts.

cli/azd/pkg/watch/watch.go
```diff
@@ -0,0 +1,142 @@
...
+type fileWatcher struct {
+	fileChanges *fileChanges
+	mu          sync.Mutex
+}
+
+type fileChanges struct {
+	Created  map[string]bool
+	Modified map[string]bool
+	Deleted  map[string]bool
+}
```
wbreza, line 32: A couple things to consider here as a fast follow... right now what happens if a file is created, then modified.  Right now I think it would show up in both lists... If it is created we likely shouldn't also show it in modified.  we may consider just keying off of the file name and then have some state associated with each file whether it is created, modified or deleted.

wbreza: Actually it looks like you are resolving some of that already :)

cli/azd/internal/agent/tools/dev/command_executor.go
```diff
@@ -120,6 +120,16 @@ func (t CommandExecutorTool) Call(ctx context.Context, input string) (string, er
...
 
+	if req.Command == "azd" {
+		errorResponse := common.ErrorResponse{
+			Error:   true,
+			Message: "azd command is not supported",
+		}
+
+		jsonData, _ := json.MarshalIndent(errorResponse, "", "  ")
+		return string(jsonData), nil
+	}
```
wbreza: This will break some of the agentic azd init flows... right now it does invoke azd for provision preview, azd package verification and environment creation.

I know we briefly discussed removing `azd` from being allowed to be executed but we'll need to figure something else out... Maybe we don't allow it to re-invoke the same command that is running to prevent it from starting its own inner loop?

wbreza (commented):
A few more comments. Let me know if you have any questions.

cli/azd/cmd/middleware/error.go
```diff
@@ -0,0 +1,353 @@
...
+		featuresManager:   featuresManager,
+		userConfigManager: userConfigManager,
+	}
+}
+
+func (e *ErrorMiddleware) Run(ctx context.Context, next NextFn) (*actions.ActionResult, error) {
+	var actionResult *actions.ActionResult
+	var err error
+
+	if e.featuresManager.IsEnabled(llm.FeatureLlm) {
```
vhvb1989: Can you re-write this in a way that the function exit/continue if the feature is not enabled?  That would save us from one layer of indentation.

cli/azd/cmd/middleware/error.go
```diff
@@ -0,0 +1,353 @@
...
+		if e.options.IsChildAction(ctx) {
+			actionResult, err = next(ctx)
+			if err != nil {
+				e.console.StopSpinner(ctx, "", input.Step)
+			}
+			return actionResult, err
+		}
```
vhvb1989: Can you add a comment about why errors in child actions are ignored?

vhvb1989: And why do we need to call stopSpinner for child actions?

cli/azd/cmd/middleware/error.go
```diff
@@ -0,0 +1,353 @@
...
+
+		actionResult, err = next(ctx)
+		if err == nil {
+			return actionResult, err
+		}
+
+		// Stop the spinner always to un-hide cursor
+		e.console.StopSpinner(ctx, "", input.Step)
```
vhvb1989: Looks like regardless of `isChildAction()`, we are doing:

```go
actionResult, err = next(ctx)
		if err == nil {
			return actionResult, err
		}

		// Stop the spinner always to un-hide cursor
		e.console.StopSpinner(ctx, "", input.Step)
```

Maybe run that and only check for isChildAction after?

cli/azd/cmd/middleware/error.go
```diff
@@ -0,0 +1,353 @@
...
+		suggestion := ""
+		var previousError error
+		var suggestionErr *internal.ErrorWithSuggestion
+		var errorWithTraceId *internal.ErrorWithTraceId
+		skipAnalyzingErrors := []string{
+			"environment already initialized",
+			"interrupt",
+			"no project exists",
+		}
+		AIDisclaimer := output.WithGrayFormat("The following content is AI-generated. AI responses may be incorrect.")
```
vhvb1989: Do we really need this?  Even after having this as an Alpha feature (which, by definition is an experiment and not 100% ready feature) ??  Feels redundant to use this for alpha features

cli/azd/pkg/watch/watch.go
hemarina: Yeah, current logic is that if it is created, it won't show in modified.

cli/azd/cmd/middleware/ux.go
hemarina: Yes. And if there's error, it will be printed out in error middleware instead of in UX.

cli/azd/cmd/middleware/error.go
```diff
@@ -0,0 +1,353 @@
...
+			e.console.Message(ctx, output.WithErrorFormat("\nERROR: %s", originalError.Error()))
+
+			if errors.As(originalError, &errorWithTraceId) {
+				e.console.Message(ctx, output.WithErrorFormat("TraceID: %s", errorWithTraceId.TraceId))
+			}
+			if errors.As(originalError, &suggestionErr) {
+				suggestion = suggestionErr.Suggestion
+				e.console.Message(ctx, suggestion)
+				return actionResult, originalError
+			}
```
vhvb1989, line 115: This error handling is not using the AI agent.
Probably move this logic out of the loop (before the loop).

This is only checking if there is a suggestion from the error and returning that.  Right?  
If yes, this can happen before creating the AI agent. And it might be automatically filtering all errors that returns suggestions

cli/azd/cmd/middleware/error.go
```diff
@@ -0,0 +1,353 @@
...
+				if err != nil {
+					if agentOutput != "" {
+						e.console.Message(ctx, AIDisclaimer)
+						e.console.Message(ctx, output.WithMarkdown(agentOutput))
+					}
+
+					return nil, err
+				}
+
+				e.console.Message(ctx, AIDisclaimer)
```
vhvb1989: looks like we are really making sure this AIDisclaimer is displayed as many times as possible XD.  Right?

cli/azd/cmd/middleware/error.go
hemarina: This is the disclaimer for AI generated content and discussed during UX office hour. @ccbarragan mentioned that this is needed for some disclaimer policy. Could you help me explain the policy?

cli/azd/cmd/middleware/error.go
hemarina: Yes, every time if we print out agentOutput. We will print out AIDisclaimer. If we confirmed this behavior will happen in other places like azd init Agent flow (need discussion with designers). I'll create a logic to add it into agent.go or a ux function for agent.go in follow up PR.

cli/azd/internal/agent/tools/dev/command_executor.go
hemarina: Discussed offline. We will use a block list strategy for now and redesign it in follow up PR.

cli/azd/cmd/middleware/ux.go
wbreza: hmm... if we weren't able to fix the error and there is still an error i would expect to call `next(...)` with the error to allow the ux middleware to handle it consistently??

wbreza (approved):
Looks good enough for now! Lets get some usage of this and make sure it doesn't cause any regressions.

cli/azd/cmd/middleware/ux.go
hemarina: If we are not able to fix it, in error.go, the behavior is we return this error message and a warning that AI is not able to fix it and go back to ux middleware. Its error is already returned. It is the same format of how ux middleware handle error message.
```
if previousError != nil && errors.Is(originalError, previousError) {
			attempt++
			if attempt >= 3 {
				e.console.Message(ctx, fmt.Sprintf("Please review the error and fix it manually, "+
					"%s was unable to resolve the error after multiple attempts.", agentName))
				return actionResult, originalError
			}
		}
```.

rajeshkamal5050 (approved):

cli/azd/cmd/middleware/error.go
ccbarragan: sure @hemarina @vhvb1989 - In product UIs, AI disclaimers are required by CELA for AI transparency and trust. I don't think there's a hard-and-fast rule for CLIs that I've seen but I would recommend including a transparency note to follow the guidance from CELA.

- https://microsoft.sharepoint.com/sites/CELAWeb-Marketing/SitePages/AI-Messaging-Guidance.aspx?web=1
- https://learn.microsoft.com/en-us/product-style-guide-msft-internal/responsible-ai-style-guide/transparency/communication-talking-about-our-ai-services-and-products-with-transparency

PR #5533, opened 2025-07-25
fix html elements

alexwolfmsft:


cli/azd/cmd/testdata/TestUsage-azd-env.snap
```diff
@@ -17,7 +17,7 @@ Available Commands
   refresh   	: Refresh environment settings by using information from a previous infrastructure provision.
   select    	: Set the default environment.
   set       	: Set one or more environment values.
-  set-secret	: Set a <name> as a reference to a Key Vault secret in the environment.
+  set-secret	: Set a `<name>` as a reference to a Key Vault secret in the environment.
```
JeffreyCA: Do you mind sharing some more context around this change? I noticed some unescaped `<` and `>` characters above (`<environment-name>`) - does that need to be changed too?

alexwolfmsft: > Do you mind sharing some more context around this change? I noticed some unescaped `<` and `>` characters above (`<environment-name>`) - does that need to be changed too?

These particular elements are rendering as HTML elements and causing warnings on all auto-generated doc updates for the command reference page. We're not able to merge PRs with warnings so it requires manual fixing each time - this is to escape them so they don't render as html.

JeffreyCA: I see, so this piece of text?

image

We could also just drop the angle brackets in the description and have it say "Set a name as a reference..." - I don't think it loses any clarity

cli/azd/cmd/env.go
```diff
@@ -59,8 +59,8 @@ func envActions(root *actions.ActionDescriptor) *actions.ActionDescriptor {
 
 	group.Add("set-secret", &actions.ActionDescriptorOptions{
 		Command: &cobra.Command{
-			Use:   "set-secret <name>",
-			Short: "Set a <name> as a reference to a Key Vault secret in the environment.",
+			Use:   "set-secret " + output.WithBackticks("<name>") + "",
```
JeffreyCA: Is this `Use` string causing issues as well? We don't seem to escape these on other commands

JeffreyCA (approved):

JeffreyCA (approved):

danieljurek (approved):
LGTM 

image


image

PR #5536, opened 2025-07-25
vsrpc: support non-aspire projects

### LINKED ISSUE #5516: azd vs-server should support not-just-Aspire scenarios
We'd like to investigate being able to call azd from VS in ways unrelated to Aspire but atm, the integration relies on the presence of an App Host. We wouldn't want any new magic invented; the implication is, we'd like to enable folks to be able to use azd to deploy from within VS like they can for Aspire apps, if their projects are already dialed-up with azd.

weikanglim:
This change adds underlying support for non-Aspire (i.e. native Blazor projects) for the existing vsrpc server that works with Visual Studio connected services.

This change does not light up the full functionality, which requires changes further upstream changes for VS, but has been tested to work end-to-end:
image

Closes #5516


weikanglim:
This is still active work

Issue #4079, opened 2024-07-06
"azd completion bash --help" does not show dedicated help


cforce:
According to doc  https://github.com/Azure/azure-dev#set-up-shell-completion

"azd completion bash --help" does show dedicated help - but does not - juts genersl cmd line help
"azd completion bash" onloy outrputs completion script without any help as well.

Manual or cmd help is bugg,y either ir

Issue #2846, opened 2023-10-09
[Issue] Running `azd completion <shell> --help` does not print help related to installing shell completions


agpt8:
- [  ] Make sure you've installed the latest version using [instructions in the wiki](../wiki/install)

**Output from `azd version`**
azd version 1.4.1 (commit 1cbed85cf47ed235b5ed31868eda79619525bfcc)

**Describe the bug**
Under the shell completion section of the [README](https://github.com/Azure/azure-dev#set-up-shell-completion), it says to get instructions for specific shell, run `azd completion <shell> --help` (substituting actual name here as the output is same for each shell).

However, instead of printing specific instructions for installing completions for each shell, the generic help (which can also be viewed with `azd --help` is printed.

**To Reproduce**
1. Open terminal with a shell of your choice
2. Ensure azd is installed.
3. Run `azd --help` and see the output.
4. Now run `azd completion <your shell name> --help` and see the output.

**Expected behavior**
The output of the command `azd completion <your shell name> --help` should provide specific instruction to install completion for that shell and not print the generic help which is the same output as `azd --help`.

**Environment**
Information on your environment:
    * macOS Sonoma with regular terminal
    * zsh version (this came built in): zsh 5.9 (x86_64-apple-darwin23.0)
    * bash version (this came built in): GNU bash, version 3.2.57(1)-release (arm64-apple-darwin23)
    * fish version: fish, version 3.6.1

**Additional context**
This issue is encountered on a fresh install of azd and the system itself. Login process has not been initiated yet, though I don't think it should affect it in anyway.

agpt8:
Any updates on this? I can still reproduce this issue on azd version 1.5.0 (commit 012ae734904e0c376ce5074605a6d0d3f05789ee)

vhvb1989:
Hello @agpt8, thank you for reporting this issue.  We'll plan for a fix soon (latest features for 1.5.0 have kept us busy :)

Have you used azd completions before? I'd like to understand what's the common use-case. Would you expect to generate completion after each azd-release to update your shell?

agpt8:
I have used azd completions for powershell and bash. While I knew how to add those completions to my powershell profiles file,  wanted to know if azd recommends some other way to do that or if it had any other mechanims to provide the completions for that shell. Which is why I used the `azd completion <shell> --help` hoping that it would provide instructions on how to use those completions or atleast some helpful info.

[...]

PR #5538, opened 2025-07-26
Add CLI Command Shortcuts Feature

vhvb1989:
# Add CLI Command Shortcuts Feature

## Summary

This PR introduces a comprehensive command shortcuts feature for the Azure Developer CLI (azd), allowing users to use abbreviated command prefixes to speed up common operations.

### 🚀 **Intelligent Prefix Matching**
- Use shortest unique prefixes: `azd b` → `azd build`, `azd u` → `azd up`
- Multi-level shortcuts: `azd pi c` → `azd pipeline config`, `azd au logi` → `azd auth login`
- Smart ambiguity detection with helpful suggestions

### 🎯 **User-Friendly Configuration**
- **Disabled by default** - opt-in feature via `azd config set command.shortcuts on`
- Integrates with azd's existing configuration system
- Persistent settings stored in user config (`~/.azd/config.json`)
- Easy enable/disable: `azd config set command.shortcuts on/off`

### 🛡️ **Robust Error Handling**
- Clear ambiguity messages with minimum prefix suggestions
- Fallback to normal Cobra behavior for unknown commands
- Maintains full backwards compatibility

## Examples

```bash
# Enable shortcuts
azd config set command.shortcuts on

# Basic shortcuts
azd b           # → azd build
azd u           # → azd up  
azd m           # → azd monitor

# Multi-level shortcuts
azd pi c        # → azd pipeline config
azd au logi     # → azd auth login
azd inf d       # → azd infra delete

# Ambiguity handling
azd p           # → Error with suggestions: package, pipeline, provision
```

### Architecture
- **Dynamic Command Discovery**: Automatically builds shortcut registry from Cobra command tree
- **Minimum Prefix Calculation**: Ensures shortest unique prefixes to avoid ambiguity
- **Preprocessing Pipeline**: Arguments are expanded before Cobra handles them
- **Configuration Integration**: Uses azd's existing config system for persistence

### Files Added/Modified
- `cmd/shortcuts.go` - Core shortcut matching logic (209 lines)
- `cmd/shortcuts_config.go` - Configuration management (61 lines)
- `cmd/shortcuts_utils.go` - Utility functions and help (123 lines)
- `cmd/shortcuts_test.go` - Comprehensive test suite (415 lines)
- `cmd/root.go` - Integration with main CLI execution flow (+75 lines)
- `main.go` - Updated entry point (+4 lines)
- `docs/SHORTCUTS.md` - Complete feature documentation (199 lines)

## Breaking Changes

**None** - This is a purely additive feature that:
- Is disabled by default
- Maintains full backwards compatibility
- Doesn't modify existing command behavior
- Falls back gracefully for unknown commands

## Testing

- ✅ Comprehensive test suite (415 lines)
- ✅ All shortcut combinations tested
- ✅ Configuration persistence verified
- ✅ Performance impact minimal

This feature significantly improves the developer experience by reducing typing and enabling faster command execution.


weikanglim:
@vhvb1989 I do appreciate the work and thought put into this. I'm wondering if the shape of the API may change overtime as more commands are added, due to the nature of dynamic generation. My primary concern as a user is that I wouldn't be too happy if I needed to re-evaluate what keys to press as new versions of azd are released.

I do like the idea of "shortcuts" overall. In my own workflow, this is an autocompletion feature, followed by shell aliases for very commonly repeated cvommands.

I'm curious to learn more from your workflow if you have azd autocompletion set up in your current shell.

With my current autocompletion setup, I tend to just type `azd pi` and tab (autocomplete) to get `pipeline`.
If there's ambiguity, I like how suggestions are presented as an overlay to me in my shell:
image

vhvb1989:
@weikanglim , I like autocompletion and aliases features, but I don't like that you need to manage the setup per device.

Sure, you can find or create tools to sync files across devices, like your `.bashrc`.  But it adds even more steps (and auth) for a very simple and fundamental request:  "be lazy to type the less to get the most".  Based in this request, you are hoping to get the functionality from out-of-the-box instead of making it something that depends on the user to set up their device for it.

Another difference I see between shortcuts and autocompletion/aliases is that shortcuts is a 100% feature for experienced users who work with the tools every day, run some commands over and over. They can create an alias, sure, but it is not only that aliases are per-device, it is also the fact that, as a user, you can't share a video or do a live demo using your aliases w/o having to explain to the audience what is that you are typing 🎭.  Shortcuts bring a feature that works for all.

Here's one of my main ideas enhanced by copilot-writting:

Some features in a product go far beyond their functional utility—they become reflections of the culture and identity of the team behind it. These details speak volumes about how a team thinks: whether they embrace creativity and bold innovation, or lean toward a more conservative approach by echoing what others have already done. It's in the nuanced choices—the unexpected delight, the thoughtful friction, the elegant solution to a subtle problem—that a product begins to tell a story. These features shape not just how a tool works, but how it feels. They transform mere usage into a meaningful experience, forging a connection between the user and the product that’s rooted in personality, intention, and care.

-----

So, to me, a feature like this is more about telling a user they don't need to set up autocompletion and/or aliases if they don't want to, that we got something they can try because we care 😊

weikanglim:
@vhvb1989 Thanks for engaging in the discussion! I understand that you're looking for the most straightforward path to enable a great experience for all our users, and your heart is definitely in the right place. 

I would love to support your efforts here and try to see this as an overall improvement, but I'm hesistant based on DX painpoints related to day-to-day usage experience:

1. `azd b` works today, but tomorrow it may stop working entirely.
2. `azd b` is not a great demo experience -- it is a very personal alias workflow, that may not be suited for the best demo, and the concern from 1. applies, where I could imagine that a demo recording of `azd b` could easily be made outdated by unrelated changes to azd command structure.

I still see autocompletion as a general better usage experience for most users -- FWIW, I would love to see #4079, #2846 being addressed to make autocompletion setup as frictionless as possible.

vhvb1989:
I understand the concern @weikanglim . Thanks for sharing your POV.

I think the main difference in our lines of thinking is here:

> azd b is not a great demo experience -- it is a very personal alias workflow, that may not be suited for the best demo

I see this the other way around 😆.  It is OK because at this point we are no longer talking about the actual functionality, but 100% about the impression a product can give you by going out of the common.  I do not worry, for example, about future commands because we have enough evidence from the last years about the adding-new-commands rhythm (and the expectation about having more extensions than new commands).  
And I am even making this feature opt-in. People would basically need to enable it to try it, and have an easy way to disable it if doesn't fit their needs.

I usually support whenever there are these types of things like eastern eggs or special functionality you can unblock and try from products. It makes them fun IMO. And I also think that creativity and disruptive in products is something teams needs to harvest with care, as innovation usually is an output of iterating within the unexpected

cli/azd/main.go
```diff
@@ -62,7 +62,9 @@ func main() {
 
 	rootContainer := ioc.NewNestedContainer(nil)
 	ioc.RegisterInstance(rootContainer, ctx)
-	cmdErr := cmd.NewRootCmd(false, nil, rootContainer).ExecuteContext(ctx)
+
+	// Execute with shortcut preprocessing support
+	cmdErr := cmd.ExecuteRootCmdWithShortcuts(ctx, false, nil, rootContainer)
```
wbreza, line 67: If we want to move forward with this feature, I think building a `shortcut` middleware would be a more maintainable approach

Issue #5506, opened 2025-07-18
[Issue] Problem deploying a simple static HTML app to SWA using azd


davrous:
- [ ] Make sure you've installed the latest version using [instructions](https://github.com/Azure/azure-dev?tab=readme-ov-file#installupgrade-azure-developer-cli)

**Output from `azd version`**
azd version 1.18.0 (commit 2c3bc5a3f885fed2e009195f466503f2fe2c22a1)

**Describe the bug**
I wanted to do a fun experiment where I would create a simple space invaders game in HTML5, in vibe coding with GHC + Claude Sonnet 4, and then use Azure MCP servers to deploy it to Azure.

I tried 3 times and every time, the game was perfectly created but the deployment, using SWA as recommendation from AI, failed. It was able to create the Azure RG and create the SWA inside it but every time, the SWA was empty. The agentic DevEx wasn’t able to package the local static files to deploy them to Azure for whatever reason.

GH Copilot tried sometimes to fall back to the SWA CLI but failed also. 

At the end, Copilot shared the link to the public web app but it only display the default page of an empty static web app.

**To Reproduce**
•	Inside GH Copilot, with Agent mode and Claude Sonnet 4
•	First prompt "create a space invaders game using HTML5"
•	Second part: "can you please deploy the web app to Azure?"
•	GHC uses our MCP server to create the yaml, bicep files and uses azd up to deploy a SWA to Azure.

**Expected behavior**
The static web files should be deployed as expected

**Environment**
Information on your environment:
-	VS Code stable release v1.102.1 
-	GitHub Copilot pre-release 1.345.1685
-	GitHub Copilot for Azure 1.0.42
-	Azure Tools 1.6.0
-	AI Toolkit pre-release
-	Azure AI Foundry release 0.7.0
-	Azure App Service preview 0.26.3
-	Azure CLI Tools 0.6.0
-	Etc (complete list below)

ecmel.vscode-html-css
github.copilot
github.copilot-chat
ms-azure-load-testing.microsoft-testing
ms-azuretools.azure-dev
ms-azuretools.vscode-azure-github-copilot
ms-azuretools.vscode-azureappservice
ms-azuretools.vscode-azurecontainerapps
ms-azuretools.vscode-azurefunctions
ms-azuretools.vscode-azureresourcegroups
ms-azuretools.vscode-azurestaticwebapps
ms-azuretools.vscode-azurestorage
ms-azuretools.vscode-azurevirtualmachines
[...]

PR #5539, opened 2025-07-28
Fail fast when deploying SWA from the root with no output folder

vhvb1989:
fix: https://github.com/Azure/azure-dev/issues/5506

This change makes the SWA deployment to validate if users are trying to deploy a service from the root of the project (next to azure.yaml) without using an output folder for the service and fail with a details error and suggestion about it.

Azure SWA has a limitation (Azure/static-web-apps#1672) about deploying applications which source code is the output at the same time. For example, trying to deploy a single HTML file to say hello world.

The returned suggestion helps AI agents to make the required changes to the project to make deployment work.

```
ERROR: error executing step command 'deploy --all': failed deploying service 'name': Application folder path and Application Output folder path cannot be the same
If your service is at the root of your project, next to azure.yaml, move your service to a subfolder.
Azure Static Web Apps does not support deploying from a folder that is for both the service source and the output folder.
Update the path of the service in azure.yaml to point to the subfolder and try deploying again.
```

### DIFF
--- a/cli/azd/pkg/project/service_target_staticwebapp.go
+++ b/cli/azd/pkg/project/service_target_staticwebapp.go
@@ -15,7 +15,6 @@ import (
 	"github.com/azure/azure-dev/cli/azd/pkg/azapi"
 	"github.com/azure/azure-dev/cli/azd/pkg/azure"
 	"github.com/azure/azure-dev/cli/azd/pkg/environment"
-	"github.com/azure/azure-dev/cli/azd/pkg/output/ux"
 	"github.com/azure/azure-dev/cli/azd/pkg/tools"
 	"github.com/azure/azure-dev/cli/azd/pkg/tools/swa"
 )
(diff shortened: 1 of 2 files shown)


cli/azd/pkg/project/service_target_staticwebapp.go
```diff
@@ -117,6 +119,20 @@ func (at *staticWebAppTarget) Deploy(
...
+		if (packageOutput.PackagePath == "" || cwd == packageOutput.PackagePath) &&
+			cwd == serviceConfig.Project.Path {
+			return nil, &internal.ErrorWithSuggestion{
+				Err: fmt.Errorf("service source and output folder cannot be at the root: %s", serviceConfig.RelativePath),
+				Suggestion: (&ux.MultilineMessage{Lines: []string{
+					"If your service is at the root of your project, next to azure.yaml, move your service to a subfolder.",
+					"Azure Static Web Apps does not support deploying from a folder that is for both the service" +
+						" source and the output folder.",
+					"Update the path of the service in azure.yaml to point to the subfolder and try deploying again.",
+				}}).ToString(""),
```
JeffreyCA: Would `strings.Join` be more appropriate here? I see we usually use `ux.*` types with `Console.MessageUxItem` and it feels a little unusual seeing it used outside of that context.

vhvb1989: makes sense, yes

Issue #5425, opened 2025-07-01
MFA impact and changes for AZD


rajeshkamal5050:
Impact: Starting Sept 1st, ARM will require MFA for all create/update/delete operations on public applications. If a non-MFA token is used, Azure will respond with claims challenges. Several scenarios will be affected (assuming MFA is not enforced at the tenant level)

Ref: https://aka.ms/mfaforazure

Since AZD uses MSAL for Go Client libraries, determine impact and address changes.
https://learn.microsoft.com/en-us/entra/identity/authentication/concept-mandatory-multifactor-authentication?tabs=go#client-libraries

Tasks:
- [x] azd: Handle claims challenges and user re-authentication - https://github.com/Azure/azure-dev/pull/5545
- [x] ~~vsserver: Handle auth flow~~ - VS team committed to showing error messages for now
- [x] ~~vscode: Handle auth flow~~ Logged: https://github.com/Azure/azure-dev/issues/5553
- [x] ~~azd-credential: Handle claim challenges~~ - Azure SDK team is driving this

PR #5545, opened 2025-07-30
auth: handle claims challenges due to Conditional Access Evaluation

weikanglim:
Adds support for claims challenges that can occur as part of [Conditional Access Evaluation](https://learn.microsoft.com/en-us/entra/identity/conditional-access/concept-continuous-access-evaluation).

When a request is rejected by ARM, the following will occur:
1. ARM returns 401 Unauthorized with `claims` populated in `Www-Authenticate` header.
2. [BearerTokenPolicy](https://github.com/Azure/azure-sdk-for-go/blob/4975987a031547cc54fff1a162f55144498bfad4/sdk/azcore/runtime/policy_bearer_token.go#L152) handles parsing of the claims, and will call the credential's `GetToken` implementation with `TokenRequestOptions.Claims` populated.
3. In our credential implementation (azd_credential.go), we now pass the claims through to `azidentity`'s `AcquireTokenSilent` implementation. In almost all cases, this call will fail, meaning the claims aren't satisifable by the current auth token. We save the claims and return a `newReLoginRequiredError` informing the user to log in.

Additionally, two other small changes included to fix the error presented to the user:

1. Refactoring the deployment error to avoid presenting to the user the login error as a "Validation Error" https://github.com/Azure/azure-dev/commit/baeb3453c17b12c22bdeadaefb0da92cff6a6d06
2. Fixing the error message to omit default scope for ARM, i.e. show `azd auth login` instead of `azd auth login --scope https://management.core.windows.net//.default` https://github.com/Azure/azure-dev/commit/f7536a80e320d57cfc4516fddb3d2544af8d5372

Error message displayed:
image

Contributes to #5425

### DIFF
--- a/cli/azd/pkg/auth/manager.go
+++ b/cli/azd/pkg/auth/manager.go
@@ -697,6 +697,25 @@ func (m *Manager) LoginInteractive(
 		acquireTokenOptions = append(acquireTokenOptions, public.WithOpenURL(options.WithOpenUrl))
 	}
 
+	claimsFile, err := claimsFilePath()
+	if err != nil {
+		return nil, err
+	}
+
+	bytes, err := os.ReadFile(claimsFile)
+	if errors.Is(err, os.ErrNotExist) {
+		// do nothing, no claims to add
+	} else if err != nil {
+		return nil, fmt.Errorf("reading claims file: %w", err)
+	} else {
+		var validJson map[string]any
+		if err := json.Unmarshal(bytes, &validJson); err == nil {
+			acquireTokenOptions = append(acquireTokenOptions, public.WithClaims(string(bytes)))
+		} else if err := os.Remove(claimsFile); err != nil { // remove file immediately if it's not valid json
+			return nil, fmt.Errorf("removing claims file '%s': %w. Remove this file manually to recover", claimsFile, err)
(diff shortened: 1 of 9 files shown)


cli/azd/pkg/auth/manager.go
```diff
@@ -1230,6 +1253,15 @@ func readUserProperties(cfg config.Config) (*userProperties, error) {
 	return &user, nil
 }
 
+func claimsFilePath() (string, error) {
+	configDir, err := config.GetUserConfigDir()
+	if err != nil {
+		return "", fmt.Errorf("failed to get config dir: %w", err)
+	}
+
+	return filepath.Join(configDir, "auth.claims"), nil
```
vhvb1989, line 1262: Are we creating a folder "auth.claims" ?  or a folder "claims" inside the "auth" folder?

weikanglim: `auth.claims` as a file, i.e. `~/.azd/auth.claims`. Similar to `~/.azd/auth.json`.

The `~/.azd/auth/` directory is largely reserved for `azidentity` storage currently.

vhvb1989 (approved):
LGTM

PR #5565, opened 2025-08-06
Add fallback to PowerShell 5 when PowerShell 7 is unavailable

spboyer:
Implement a feature that allows the script runner to fall back to PowerShell 5 if PowerShell 7 is not available, enhancing compatibility and user experience. Include configuration options to enable or disable this feature.


vhvb1989:
@spboyer , I am closing this in favor of https://github.com/Azure/azure-dev/pull/5570

The strategy is similar but a little bit more general, thinking that other shells might use the idea of `alternative shells` at any point in the future and not only powershell.

And I didn't want to use an alpha feature a switch. Instead, this is becomes opt-out as part of the hook settings defintion.

Issue #5507, opened 2025-07-20
Broken .NET Aspire Azure CI/CD pipeline: error unmarshalling Bicep template parameters


Dayonel:
- [x] Make sure you've installed the latest version using [instructions](https://github.com/Azure/azure-dev?tab=readme-ov-file#installupgrade-azure-developer-cli)

**Output from `azd version`**
Run `azd version` and copy and paste the output here:

azd version 1.18.0 (commit 2c3bc5a3f885fed2e009195f466503f2fe2c22a1)

**Describe the bug**

I encountered this issue after upgrading my .NET Aspire project to 9.3.1 in the ci/cd pipeline (Azure DevOps):

`ERROR: initializing provisioning manager: resolving bicep parameters file: error unmarshalling Bicep template parameters: invalid character 'a' after object key:value pair`

Unfortunately I tried many things but can't pass this issue.

Last working pipeline used `AZD_INITIAL_ENVIRONMENT_CONFIG` that now is deprecated.
Locally `azd provision` and `azd deploy` works.

**To Reproduce**
Upgrade an Aspire project from 9.1.0 to 9.3.0.
Create a CI/CD pipeline using this command `azd pipeline config --provider azdo`
Have this `main.parameters.json`:
```
{
    "$schema": "https://schema.management.azure.com/schemas/2019-04-01/deploymentParameters.json#",
    "contentVersion": "1.0.0.0",
    "parameters": {
      "principalId": {
        "value": "${AZURE_PRINCIPAL_ID}"
      },
      "Authorities_AzureAd": {
        "value": "${AZURE_AUTHORITIES_AZURE_AD}"
      },
      "Authorities_EntraExternalId": {
        "value": "${AZURE_AUTHORITIES_ENTRA_EXTERNAL_ID}"
      },
      "AzureAd_ClientId": {
        "value": "${AZURE_AZURE_AD_CLIENT_ID}"
      },
      "AzureAd_Instance": {
        "value": "${AZURE_AZURE_AD_INSTANCE}"
      },
      "AzureAd_RedirectUri": {
        "value": "${AZURE_AZURE_AD_REDIRECT_URI}"
      },
      "AzureAd_TenantId": {
        "value": "${AZURE_AZURE_AD_TENANT_ID}"
      },
      "AzureAd_TokenValidationParameters_ValidAudiences": {
        "value": "${AZURE_AZURE_AD_TOKEN_VALIDATION_PARAMETERS_VALID_AUDIENCES}"
      },
[...]

PR #5568, opened 2025-08-07
Fix: Add JSON escape to variables before syncing to remote pipeline 

Menghua1:
Fixes issue https://github.com/Azure/azure-dev/issues/5507.

This PR adds a step to JSON-escape the parameters from the local `.azure/envName/config.json` before syncing them to the remote pipeline, ensuring that environment parameter values remain consistent between local and remote.

## **Issue Example**
In the local `.azure/envName/config.json` file, there is a parameter defined as a JSON string:
```json
{
"AzureAd_TokenValidationParameters_ValidAudiences": "[\"api://e935a748-8b59-4c26-a59c-9bcc83f5ab57\"]"
}
```

When running `azd pipeline config` and syncing the variables in the `.azure/envName/config.json` file to the remote pipeline, the value becomes unescaped:
```
["api://e935a748-8b59-4c26-a59c-9bcc83f5ab57"]
```

This causes azd to treat the value as an array type instead of a string in the remote pipeline, resulting in Bicep parameter parsing failures `ERROR: initializing provisioning manager: resolving bicep parameters file: error unmarshalling Bicep template parameters: invalid character 'a' after object key:value pair`. 

Notably, running `azd up` locally works as expected because the value is preserved as a JSON string in the `.azure/envName/config.json` file.

This PR ensures parameter values are JSON-escaped before syncing to the remote pipeline, keeping them consistent with the local environment.

### DIFF
--- a/cli/azd/pkg/azdo/pipeline.go
+++ b/cli/azd/pkg/azdo/pipeline.go
@@ -5,7 +5,6 @@ package azdo
 
 import (
 	"context"
-	"encoding/json"
 	"fmt"
 	"strings"
 
@@ -117,7 +116,6 @@ func CreatePipeline(
 		if err != nil {
 			return nil, err
 		}
-		EscapeBuildDefinitionVariables(buildDefinitionVariables)
 		definition.Variables = buildDefinitionVariables
 		definition, err := client.UpdateDefinition(ctx, build.UpdateDefinitionArgs{
 			Definition:   definition,
@@ -142,28 +140,12 @@ func CreatePipeline(
 		return nil, err
 	}
 
-	EscapeBuildDefinitionVariables(createDefinitionArgs.Definition.Variables)
 	newBuildDefinition, err := client.CreateDefinition(ctx, *createDefinitionArgs)
 	if err != nil {
 		return nil, err
 	}
 
 	return newBuildDefinition, nil
-}
-
-func EscapeBuildDefinitionVariables(vars *map[string]build.BuildDefinitionVariable) {
-	for key, variable := range *vars {
-		if variable.Value != nil {
-			original := *variable.Value
-			b, _ := json.Marshal(original)
-			s := string(b)
-			if len(s) >= 2 && s[0] == '"' && s[len(s)-1] == '"' {
-				s = s[1 : len(s)-1]
-			}
-			variable.Value = &s
-			(*vars)[key] = variable
-		}
-	}
 }
 
 func getDefinitionVariables(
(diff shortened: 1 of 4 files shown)


Menghua1:
> Please provide some examples of what the PR solves. I don't really understand from the current description.

@wbreza I’ve updated the PR description with a specific issue example. Please have a look!

rajeshkamal5050:
@vhvb1989 can you review this?

weikanglim:
@vhvb1989 @rajeshkamal5050 FWIW, I suspect that this happens only in Aspire use-cases when a user provides a JSON string as a parameter value.

It wouldn't happen in normal azd when the parameter is marked as an `object`, because we would persist a structured object into `config.json`. 

I wasn't entirely sure if moving forward wwith this fix would bring about other edge cases to consider. From my quick eyes, arbitrarily assuming one way or another didn't seem like the right approach, but I haven't dug into any deeper.

Menghua1:
This issue also occurs in non-Aspire azd projects when user-entered parameter values ​​contain quotes, backslashes, or special characters that require escaping.

For example, if the input is `foo"bar`, it is saved as `foo\"bar` in `config.json`. However, when syncing to a remote pipeline, these parameters are unescaped, resulting in the remote variable value becoming the original unescaped string `foo"bar`, which is inconsistent with the local `config.json` file. Consequently, special characters in the parameter cause Bicep parameter compilation failures.

This PR escapes the parameters before syncing, ensuring that the remote variable value is consistent with the local `config.json`, thus avoiding compilation issues. @vhvb1989 What are your thoughts on this PR?

vhvb1989:
@Menghua1 , I don't love the idea of having each provider (gh and azdo) to manually handle the values within their implementation. I wonder if this is something that the pipeline manager can handle before calling the provider so the provider doesn't need to care.

Ideally, I would love to see a change where the values are scaped right away when they are read from the environment config, with a note about it. 
Or even better, I would check if we can prevent writing the value incorrectly and prevent having a value in the config file which needs to be handled as special case.

vhvb1989:
@Menghua1 , can you see my last comment and update the fix so we don't do the escaping on each provider?

Menghua1:
> @Menghua1 , can you see my last comment and update the fix so we don't do the escaping on each provider?

@vhvb1989 I have updated the code and now handle parameter escaping in the [mergeProjectVariablesAndSecrets()](https://github.com/Azure/azure-dev/blob/main/cli/azd/pkg/pipeline/pipeline_manager.go#L679) method in the Pipeline Manager. Please review these changes again, thank you!

PR #5570, opened 2025-08-07
Alternative shell support

vhvb1989:
Adding alternative shells support for running hooks. This allows pwsh to try running in pwsh5 when pwsh7 is not found in windows. User can set templates to be ShellStrict and fail if pwsh7 is not found

### DIFF
--- a/cli/azd/pkg/ext/models.go
+++ b/cli/azd/pkg/ext/models.go
@@ -63,6 +63,9 @@ type HookConfig struct {
 	Name string `yaml:",omitempty"`
 	// The type of script hook (bash or powershell)
 	Shell ShellType `yaml:"shell,omitempty"`
+	// If true, azd won't try to use alternative shell to execute the command. For example, pwsh won't be tried with
+	// powershell (pwsh5) in Windows.
+	StrictShell bool `yaml:"strictShell,omitempty"`
 	// The inline script to execute or path to existing file
 	Run string `yaml:"run,omitempty"`
 	// When set to true will not halt command execution even when a script error occurs.
--- a/cli/azd/pkg/tools/powershell/powershell.go
+++ b/cli/azd/pkg/tools/powershell/powershell.go
@@ -6,6 +6,7 @@ package powershell
 import (
 	"context"
 	"fmt"
+	"runtime"
 	"strings"
 
 	"github.com/azure/azure-dev/cli/azd/internal"
@@ -54,11 +55,26 @@ func checkPath(options tools.ExecOptions) (err error) {
 // Executes the specified powershell script
 // When interactive is true will attach to stdin, stdout & stderr
 func (bs *powershellScript) Execute(ctx context.Context, path string, options tools.ExecOptions) (exec.RunResult, error) {
-	if err := bs.checkInstalled(options); err != nil {
+	// block alternative shells for non-windows.
+	// This is because the only alternative shell supported is powershell5 which is not available on non-windows platforms.
+	if runtime.GOOS != "windows" {
+		options.StrictShell = true
+	}
+
+	if err := bs.checkInstalled(options); err != nil && options.StrictShell {
 		return exec.RunResult{}, &internal.ErrorWithSuggestion{
 			Err: err,
 			Suggestion: fmt.Sprintf("PowerShell 7 is not installed or not in the path. To install PowerShell 7, visit %s",
 				output.WithLinkFormat("https://learn.microsoft.com/powershell/scripting/install/installing-powershell")),
+		}
+	}
+
+	// non-strict mode, check for alternative shell powershell 5
+	options.UserPwsh = "powershell"
+	if err := bs.checkInstalled(options); err != nil {
+		return exec.RunResult{}, &internal.ErrorWithSuggestion{
+			Err:        err,
+			Suggestion: "Make sure Powershell is installed in your system.",
 		}
 	}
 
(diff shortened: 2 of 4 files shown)


cli/azd/pkg/tools/powershell/powershell.go
```diff
@@ -54,11 +55,29 @@ func checkPath(options tools.ExecOptions) (err error) {
...
+					output.WithLinkFormat("https://learn.microsoft.com/powershell/scripting/install/installing-powershell")),
+			}
+		}
+
+		// non-strict mode, check for alternative shell powershell 5
+		options.UserPwsh = "powershell"
+		if err := bs.checkInstalled(options); err != nil {
+			return exec.RunResult{}, &internal.ErrorWithSuggestion{
+				Err:        err,
+				Suggestion: "Make sure Powershell is installed in your system.",
```
JeffreyCA:
```suggestion
				Suggestion: "Make sure PowerShell is installed on your system.",
```

schemas/v1.0/azure.yaml.json
JeffreyCA, line 1: Can we also update the alpha schema?

cli/azd/pkg/ext/models.go
```diff
@@ -63,6 +63,9 @@ type HookConfig struct {
 	Name string `yaml:",omitempty"`
 	// The type of script hook (bash or powershell)
 	Shell ShellType `yaml:"shell,omitempty"`
+	// If true, azd won't try to use alternative shell to execute the command. For example, pwsh won't be tried with
+	// powershell (pwsh5) in Windows.
+	StrictShell bool `yaml:"strictShell,omitempty"`
```
wbreza: This feels like a weird feature.  Instead of introducing a new config like "strictShell" and alternative design would be to have a scripts and which shell to execute them in.

Then in our implementation we could check if that shell is available.  If the shell is available invoke the script.  If not proceed down the list to the next shell and repeat.

This would actually make it easier so the configuration can be less OS specific especially for the cases when cross platform users have multiple shells available.

wbreza (commented):
This would be a good item to review in our next team architecture sync.

cli/azd/pkg/ext/models.go
vhvb1989: > Then in our implementation we could check if that shell is available. If the shell is available invoke the script. If not proceed down the list to the next shell and repeat.

Do you mean, adding the list of shells to the hook configuration?  In other words, supporting an array-list for the Shell field?

I didn't want to go that route b/c we are trying to fix all current azd templates w/o asking all of them to update to a new Shell configuration for the hooks.

That's why the change from this PR makes powershell5 an alternative for pwsh7 automatically (no changes required in azure.yaml).

The strictShell option is a way to opt-out from the shell alternatives, resulting in the actual behavior.

cli/azd/pkg/tools/powershell/powershell.go
```diff
@@ -54,11 +55,27 @@ func checkPath(options tools.ExecOptions) (err error) {
...
+		if runtime.GOOS != "windows" {
+			return exec.RunResult{}, &internal.ErrorWithSuggestion{
+				Err: noPwshError,
+				Suggestion: fmt.Sprintf(
+					"PowerShell 7 is not installed or not in the path. To install PowerShell 7, visit %s",
+					output.WithLinkFormat("https://learn.microsoft.com/powershell/scripting/install/installing-powershell")),
+			}
+		}
+
+		// non-strict mode, check for alternative shell powershell 5
```
wbreza: Nit - remove strict mode comment.

cli/azd/pkg/tools/powershell/powershell.go
```diff
@@ -54,11 +55,26 @@ func checkPath(options tools.ExecOptions) (err error) {
...
+					output.WithLinkFormat("https://learn.microsoft.com/powershell/scripting/install/installing-powershell")),
+			}
+		}
+
+		options.UserPwsh = "powershell"
+		if err := bs.checkInstalled(options); err != nil {
+			return exec.RunResult{}, &internal.ErrorWithSuggestion{
+				Err: err,
+				Suggestion: fmt.Sprintf(
+					"Make sure pwsh (Powershell 7) or powershell (Powershell 5) is installed on your system, visit %s",
```
JeffreyCA:
```suggestion
					"Make sure pwsh (PowerShell 7) or powershell (PowerShell 5) is installed on your system, visit %s",
```

cli/azd/pkg/tools/powershell/powershell.go
```diff
@@ -75,5 +91,17 @@ func (bs *powershellScript) Execute(ctx context.Context, path string, options to
...
 
-	return bs.commandRunner.Run(ctx, runArgs)
+	result, err := bs.commandRunner.Run(ctx, runArgs)
+	if err != nil {
+		if noPwshError != nil {
+			err = &internal.ErrorWithSuggestion{
+				Err: err,
+				Suggestion: "pwsh (Powershell 7) was not found and powershell (Powershell 5) was automatically used " +
+					"instead. You can try installing pwsh and trying again in case this script is not compatible with " +
+					"Powershell 5.",
```
JeffreyCA:
```suggestion
				Suggestion: "pwsh (PowerShell 7) was not found and powershell (PowerShell 5) was automatically used " +
					"instead. You can try installing pwsh and trying again in case this script is not compatible with " +
					"PowerShell 5.",
```

JeffreyCA: I think we should mention the Learn link here as well, or the command that @weikanglim suggested, any thoughts?

JeffreyCA (commented):
Looks good, there are some failing tests though

PR #5575, opened 2025-08-11
EngSys: Disable SBOM generation in non-release builds

danieljurek:


danieljurek:
Need to use fork. Closing.

PR #5585, opened 2025-08-14
Adds warnings when 'pwsh' is not available

wbreza:


cli/azd/pkg/ext/hooks_manager.go
```diff
@@ -114,3 +123,68 @@ func (h *HooksManager) filterConfigs(
...
+			}
+
+			// Append install instructions link
+			warningMessage = fmt.Sprintf(
+				"%s (%s)",
+				warningMessage,
+				output.WithHyperlink(
+					//nolint:lll
+					"https://learn.microsoft.com/en-us/powershell/scripting/install/installing-powershell?view=powershell-7.4",
+					"Install Instructions",
```
JeffreyCA, line 179:
```suggestion
					"https://learn.microsoft.com/powershell/scripting/install/installing-powershell",
					"Install Instructions",
```

cli/azd/pkg/ext/hooks_manager.go
```diff
@@ -114,3 +123,68 @@ func (h *HooksManager) filterConfigs(
...
+		if err := h.commandRunner.ToolInPath("pwsh"); errors.Is(err, osexec.ErrNotFound) {
+			var warningMessage string
+
+			// Check if legacy powershell is available
+			if powershellErr := h.commandRunner.ToolInPath("powershell"); !errors.Is(powershellErr, osexec.ErrNotFound) {
+				//nolint:lll
+				warningMessage = "PowerShell 7 (`pwsh`) commands found in project. Your computer only has PowerShell 5.1 (`powershell`) installed. azd will use `powershell` but errors may occur.\n\nTo resolve warning, install `pwsh`"
+			} else {
+				//nolint:lll
+				warningMessage = "PowerShell 7 (`pwsh`) commands found in project. No PowerShell installation detected. Powershell scripts will fail. \n\nTo resolve warning, install `pwsh`"
```
JeffreyCA, line 169:
```suggestion
				warningMessage = "PowerShell 7 (`pwsh`) commands found in project. No PowerShell installation detected. PowerShell scripts will fail. \n\nTo resolve warning, install `pwsh`"
```

cli/azd/cmd/middleware/hooks.go
```diff
@@ -66,6 +67,13 @@ func (m *HooksMiddleware) Run(ctx context.Context, next NextFn) (*actions.Action
 		return next(ctx)
 	}
 
+	// Validate hooks and display any warnings
+	if !m.options.IsChildAction(ctx) {
```
JeffreyCA, line 71: What does "child action" mean here?

wbreza: Only runs from the root most command... example - user runs `azd up` which calls a workflow... we only want to show the error during the initial `azd up` command and not for any other command referenced in the workflow.

weikanglim (approved):

cli/azd/pkg/ext/hooks_manager.go
JeffreyCA, line 179: Looks like some of these changes got reverted as part of the merge - I'm fine with restoring these as a follow up

JeffreyCA (approved):

cli/azd/pkg/ext/hooks_manager.go
vhvb1989, line 169: Would it be better to fail and stop here?  The warning seems like a heads up about hook will fail. Why don't we stop and do early-exit here?   

This scenario feels like pwsh not present in Linux and user didn't use the posix and Windows configuration, requesting `pwsh` as the only hook.  I think that's an opportunity here to fail

cli/azd/pkg/ext/hooks_manager.go
```diff
@@ -114,3 +123,68 @@ func (h *HooksManager) filterConfigs(
...
+		}
+	}
+
+	// If we found PowerShell hooks, check if pwsh is available
+	if hasPowerShellHooks {
+		if err := h.commandRunner.ToolInPath("pwsh"); errors.Is(err, osexec.ErrNotFound) {
+			var warningMessage string
+
+			// Check if legacy powershell is available
+			if powershellErr := h.commandRunner.ToolInPath("powershell"); !errors.Is(powershellErr, osexec.ErrNotFound) {
```
vhvb1989, line 164: Should we only check for poweshell when runtime.os == Windows?  

I think there's no point on checking for powershell when azd is not running in Windows

cli/azd/pkg/ext/hooks_manager.go
wbreza: The user may only be running a command that doesn't have a hook.  For example, maybe the project has a `postdeploy` hook or something.  Running `azd provision` would still work.  They have the opportunity to `ctrl+c` to quit and fix the issue.

cli/azd/pkg/ext/hooks_manager.go
wbreza: It also doesn't hurt anybody :)

cli/azd/pkg/ext/hooks_manager.go
vhvb1989: Mmm
So are we showing the warning ⚠️ even when the hook won't be triggered? 
That doesn't seem good.

For example running azd env set foo bar would display the warning if there is a hook in the project and there's no pwsh? 

Do we want that?

PR #5592, opened 2025-08-15
Update azd to golang 1.25.0 🚀

vhvb1989:
Moving azd from go 1.24.0 to 1.25.0

https://go.dev/doc/go1.25

fix: https://github.com/Azure/azure-dev/issues/5593


JeffreyCA (approved):

JeffreyCA:
The **CodeCoverage_Upload** task seems to be failing on this line:

https://github.com/Azure/azure-dev/blob/a7b2bdc861f0616438327854f11169a08fdfec96/eng/pipelines/templates/stages/code-coverage-upload.yml#L53

image

Fails for me locally too, but looks like that package is archived 😢: https://github.com/axw/gocov

According to [this related change in Go source](https://go-review.googlesource.com/c/example/+/679037) we may need to just update `golang.org/x/tools`.

Found a couple of forks that already updated it:
https://github.com/Zacama/gocov
https://github.com/afunix/gocov

Or should we create our own?

eng/scripts/Convert-GoCoverageToCobertura.ps1
JeffreyCA, line 1: How come there's such a big discrepancy compared to before? See `hooks_manager.go` for example:

Image

Also, all of the green lines in the new reports show either `6` or `12`. Should that be fixed? Seems consistent across all files

https://dev.azure.com/azure-sdk/internal/_build/results?buildId=5225019&view=codecoverage-tab
https://dev.azure.com/azure-sdk/internal/_build/results?buildId=5228305&_a=summary&tab=ms.vss-codecoverage-web.code-coverage-details&view=codecoverage-tab

vhvb1989: Yeah, I noticed the report went down on % and many differences in how the coverage is calculated.

I'm planning to keep working on Coverage report in the next iterations: https://github.com/Azure/azure-dev/issues/5599

For now, it is OK to be working with go 1.25.

vhvb1989: I want to also check with the team how much we want to care about Coverga reports... as it's been a while we had the reports but we have never really acted on them.  So, before jumping to fix this propertly, we need to talk about it.

eng/pipelines/templates/stages/code-coverage-upload.yml
```diff
@@ -27,36 +27,74 @@ stages:
...
+              Write-Host "Found integration coverage file: $($_.FullName)"
+          }
+          
+          # Merge unit test coverage files
+          if ($unitCoverageFiles.Count -gt 0) {
+              Write-Host "Merging $($unitCoverageFiles.Count) unit test coverage files..."
+              $unitFilesParam = $unitCoverageFiles -join ","
+              pwsh -File "$(Build.SourcesDirectory)/eng/scripts/Merge-GoCoverageProfiles.ps1" -InputFiles $unitFilesParam -OutputFile "merged-coverage/unit-coverage.out" -NormalizePaths
+              if ($LASTEXITCODE -ne 0) { 
+                  Write-Error "Failed to merge unit test coverage files"
```
weikanglim: What was the motivating factor to remove the `go tool covdata` invocations? From a quick look, it seems that [covdata](https://pkg.go.dev/cmd/covdata) is still supported today.

eng/scripts/Convert-GoCoverageToCobertura.ps1
weikanglim: I'd vote to switch to the fork and minimize churn, let's pin the version so we keep ourselves safe.

eng/pipelines/templates/stages/code-coverage-upload.yml
vhvb1989: My mistake. As I was reading and understanding why we use `go tool covdata`, I though that in go 1.21, we could have migrated to use coverageprofile instead. But I am reading again and it looks like that's the legacy and gocoverdir is better with binary files and merging with covdata.

I'll revert those changes.

weikanglim (approved):
LGTM. Thanks for taking the time to address feedbacl.

rajeshkamal5050 (approved):

Issue #5221, opened 2025-05-27
Improve the experience for `pipeline config` for Tenants where applicationServiceManagementReference is mandatory


vhvb1989:
Tenants can make `applicationServiceManagementReference` to be mandatory for creating a Service Principal.
If that's the case, AZD is currently just returning an error (400 bad request) with the notes from the Tenant about the missing applicationServiceManagementReference field.

AZD can improve the user experience here. For example:

1) Make AZD to identify the `400-bad request` response due to the missing field and prompt for a value. Tell the user their tenant has requested to provide a value for that field and wait for the value. Retry the request and continue the process after customer's input.

2) Re-write the error for the user and add a suggestion at the end about how to set the value for applicationServiceManagementReference, either by using the AZD supported argument or by making this an AZD global setting.

vhvb1989:
related to: https://github.com/Azure/azure-dev/pull/4049

PR #5598, opened 2025-08-18
Prompt for App Service Management Ref if required during SP creation

vhvb1989:
When running `azd pipeline config`, if the request to create a SP return an error code: `ServiceTreeNullValueError` or `ServiceTreeInvalid`, instead of returning an error in the terminal, azd will prompt user to input a Service Tree id, informing that the Tenant requires a valid Service Tree for creating a SP.

fixes: https://github.com/Azure/azure-dev/issues/5221


.gitattributes
```diff
@@ -1,4 +1,7 @@
 *.go text eol=lf
+**/testdata/**/*.snap text eol=lf
+**/testdata/**/*.md text eol=lf
+*.txt text eol=lf
```
weikanglim: Do we still need this? I notice that `*.snap` and `*.txt` is noted below

vhvb1989: My bad. Somehow I keep bringing this change with all my PRs LoL.  I need to fix this.

cli/azd/pkg/pipeline/pipeline_manager.go
```diff
@@ -369,15 +369,54 @@ func (pm *PipelineManager) Configure(
...
+					if invalidInput {
+						invalidInputNotes = serviceTreeInvalidError.Error()
+					}
+
+					// Prompt user for Service Tree ID (OID)
+					serviceTreeId, promptErr := pm.promptForServiceTreeId(ctx, promptForServiceTreeIdOptions{
+						PreviousWasInvalid: invalidInputNotes,
+					})
+					if promptErr != nil {
+						pm.console.Message(ctx, output.WithErrorFormat("Failed to get Service Tree ID: %s", promptErr))
```
weikanglim: Consider returning `promptErr` here? 

I believe with the current changes, a user pressing Ctrl+C ends up triggering this message and reprompting.

cli/azd/pkg/entraid/entraid.go
```diff
@@ -36,6 +36,34 @@ type AzureCredentials struct {
 	TenantId       string `json:"tenantId"`
 }
 
+// ServiceTreeNullValueError represents an error that occurs when a null value is provided for ServiceManagementReference
+type ServiceTreeNullValueError struct {
```
weikanglim, line 40: nit: would it feel better and less code to maintain with a single `ServiceTreeValidationError` with different error codes?

Currently, the only consumer of the error treats both error conditions the same. Having a single error type would leave flexibility for handling potential additional error codes. 

I feel like something like:
```golang
const (
  ServiceTreeNullValueErrorCode = 'ServiceTreeNullValueProvided'
  ServiceTreeInvalidErrorCode = 'ServiceTreeInvalidError'
)

// A validation error from service tree
type ServiceTreeValidateErr {
  ApplicationName string
  ErrorCode string
  Inner error
}
```

vhvb1989: I felt like it could not feel like writing go, so I asked copilot:

```
🔍 Short answer: Using error codes inside custom errors is not idiomatic Go, but it’s not frowned upon either — it’s just less common because Go favors simplicity and clarity over structured error taxonomies.

🧠 Why It’s Not Idiomatic
Go’s standard library and community conventions lean toward:
- Simple error values (error interface)
- Wrapping with context using fmt.Errorf("...: %w", err)
- Type assertions for custom error types when needed
You won’t find error codes in os, net/http, or database/sql errors. Instead, Go encourages checking error types or using sentinel errors like io.EOF.

```

I'll use the two error just for simplicity

weikanglim: Yeah, no worries either way. I do agree with copilot's assessment, but only in context of "use error code" for control flow.

My 2cents is error types relates to different error conditions, i.e. if a caller is going to branch on FooError vs. BarError to be handled differently from an error handling perspective, I would definitely use different error types. But if the error represents a single error condition, IMO it's a solid use case for not breaking it apart.

Similar to `azcore.ResponseError` which does have error codes in them, but to the caller, they just want to handle "a server response error".

weikanglim (approved):
Looks good!

Issue #5653, opened 2025-08-25
azd fails to deploy projects with both *.sln and *.csproj in the same directory


weikanglim:
> If I'm reading the repro right, there's probably both file(s) existing:
> 
> - `C:\Users\anvillan\source\repos\AzdAppService\BasicNewAppRazorApp.sln`
> - `C:\Users\anvillan\source\repos\AzdAppService\BasicNewAppRazorApp.csproj`
> 
> In this case, the `msbuild` command would fail. 
> 
> In `azure.yaml` (whereever it may be located), if we change:
> 
> ```yaml
> services:
>   app:
>     project: ./AzdAppService
> ```
> 
> into
> 
> ```yaml
> services:
>   app:
>     project: ./AzdAppService/BasicNewAppRazorApp.csproj
> ```
> 
> It might work.
> 
> I have to think a little more about how we should handle this in azd. This has been the existing behavior, and in previous cases, azd would consider the project as "non-App host" -- but this switch is critical for us here when we're trying to differentiate an AppHost project from a standard dotnet project. 

 _Originally posted by @weikanglim in [#5516](https://github.com/Azure/azure-dev/issues/5516#issuecomment-3166188687)_

weikanglim:
Logs:

```
2025/08/25 15:30:57 command_runner.go:310: Run exec: 'dotnet msbuild /Users/weilim/repos/playground-azd/dotnet/src/App.Project --getProperty:IsAspireHost --getItem:ProjectCapability' , exit code: 1
Additional env:
   DOTNET_CLI_WORKLOAD_UPDATE_NOTIFY_DISABLE=<redacted>
   DOTNET_NOLOGO=<redacted>
-------------------------------------stdout-------------------------------------------
MSBUILD : error MSB1063: Cannot access properties or items when building solution files or solution filter files. This feature is only available when building individual projects.
Switch: getProperty

For switch syntax, type "MSBuild -help"
2025/08/25 15:30:57 importer.go:82: error checking if /Users/weilim/repos/playground-azd/dotnet/src/App.Project is an app host project: running dotnet msbuild on project '/Users/weilim/repos/playground-azd/dotnet/src/App.Project': exit code: 1, stdout: MSBUILD : error MSB1063: Cannot a
ccess properties or items when building solution files or solution filter files. This feature is only available when building individual projects.
Switch: getProperty
```

PR #5654, opened 2025-08-25
fix msbuild/sln confusion

weikanglim:
Fixes: https://github.com/Azure/azure-dev/issues/5653


cli/azd/pkg/project/service_target_containerapp.go
```diff
@@ -86,26 +95,74 @@ func (at *containerAppTarget) Deploy(
 		return nil, err
 	}
 
-	containerAppOptions := containerapps.ContainerAppOptions{
-		ApiVersion: serviceConfig.ApiVersion,
+	// Default resource name and type
+	resourceName := targetResource.ResourceName()
+	aspireDeploymentType := azapi.AzureResourceTypeContainerApp
```
vhvb1989: this service_target was not aware of Aspire.  Are we duplicating Aspire deploying logic here and in service_target_dotnet_containerapp ?

weikanglim: I switche branches while pushing, need to roll this back

weikanglim: This is a preview of a separate feature I'm currently pushing towwards :)

vhvb1989 (approved):

PR #5655, opened 2025-08-26
Fix `azd completion` help text

### LINKED ISSUE #2846: [Issue] Running `azd completion <shell> --help` does not print help related to installing shell completions

### LINKED ISSUE #4079: "azd completion bash --help" does not show dedicated help

JeffreyCA:
Fixes #2846
Fixes #4079

### `azd --help`
image

### `azd completion --help`
image

### `azd completion bash --help`
image

### `azd completion fish --help`
image

### `azd completion zsh --help`
image

### `azd completion powershell --help`
image

### DIFF
--- a/cli/azd/cmd/completion.go
+++ b/cli/azd/cmd/completion.go
@@ -186,7 +186,8 @@ func getCmdCompletionPowerShellHelpDescription(cmd *cobra.Command) string {
 
 func getCmdCompletionPowerShellHelpFooter(cmd *cobra.Command) string {
 	return generateCmdHelpSamplesBlock(map[string]string{
-		"Load completions in current session":  "azd completion powershell | Out-String | Invoke-Expression",
-		"Install completions for all sessions": "azd completion powershell >> $PROFILE",
+		"Load completions in current session": "azd completion powershell | Out-String | Invoke-Expression",
+		"Install completions for all sessions": "azd completion powershell > $HOME\\.azd\\completion.ps1" +
+			"\n    Add-Content $PROFILE \". $HOME\\.azd\\completion.ps1\"",
 	})
 }
--- a/cli/azd/cmd/testdata/TestUsage-azd-completion-powershell.snap
+++ b/cli/azd/cmd/testdata/TestUsage-azd-completion-powershell.snap
@@ -13,7 +13,8 @@ Global Flags
 
 Examples
   Install completions for all sessions
-    azd completion powershell >> $PROFILE
+    azd completion powershell > $HOME\.azd\completion.ps1
+    Add-Content $PROFILE ". $HOME\.azd\completion.ps1"
 
   Load completions in current session
     azd completion powershell | Out-String | Invoke-Expression


cli/azd/cmd/completion.go
```diff
@@ -0,0 +1,188 @@
...
+	var err error
+	switch a.shell {
+	case "bash":
+		err = rootCmd.GenBashCompletion(a.cmd.OutOrStdout())
+	case "zsh":
+		err = rootCmd.GenZshCompletion(a.cmd.OutOrStdout())
+	case "fish":
+		err = rootCmd.GenFishCompletion(a.cmd.OutOrStdout(), true)
+	case "powershell":
+		err = rootCmd.GenPowerShellCompletion(a.cmd.OutOrStdout())
```
vhvb1989: nit. Make constants from the strings and use them when you add the subcommand and here

cli/azd/cmd/root.go
```diff
@@ -90,8 +90,6 @@ func NewRootCmd(
 		DisableAutoGenTag: true,
 	}
 
-	rootCmd.CompletionOptions.HiddenDefaultCmd = true
```
vhvb1989, line 93: Wondering why did we do this in the past....  Would the default command fail if there is not a definition for the command like the one you are adding?

JeffreyCA: > Would the default command fail if there is not a definition for the command like the one you are adding?

There's a separate option `DisableDefaultCmd` for that:

image

I'm not sure why even after removing `HiddenDefaultCmd  = true`, it doesn't show in `azd --help`. Maybe because of how we override the default `--help` text in https://github.com/Azure/azure-dev/blob/69d00310ac3dcf9dc47d71f33aa3f87649a6d02a/cli/azd/cmd/root.go#L471-L473

In any case, the docs suggest this approach to provide a custom implementation: https://github.com/spf13/cobra/blob/main/site/content/completions/_index.md#creating-your-own-completion-command

cli/azd/cmd/testdata/TestUsage-azd-completion-powershell.snap
```diff
@@ -0,0 +1,21 @@
...
+Global Flags
+    -C, --cwd string 	: Sets the current working directory.
+        --debug      	: Enables debugging and diagnostics logging.
+        --docs       	: Opens the documentation for azd completion powershell in your web browser.
+    -h, --help       	: Gets help for powershell.
+        --no-prompt  	: Accepts the default value instead of prompting, or it fails if there is no default.
+
+Examples
+  Install completions for all sessions
+    azd completion powershell >> $PROFILE
```
weikanglim: This ends up adding a lot of auto-generated stuff into Profile.ps1. 

```bash
$ azd completion powershell | wc -l          
270
```

Does it make more sense to create a separate script and append the loading instead?

```pwsh
 $ azd completion powershell > $HOME\.azd\completion.ps1
 $ Add-Content $PROFILE "$HOME\.azd\completion.ps1"
```

I saved it to `.azd\completion.ps1" as it seemed like the best place for this type of thing.

weikanglim (approved):
Love it!

vhvb1989 (approved):

cli/azd/cmd/testdata/TestUsage-azd-completion-powershell.snap
JeffreyCA: Updated, thanks for the suggestion - just needed to add a `.` to it:

image

PR #5661, opened 2025-08-27
Agent cleanup

### LINKED ISSUE #5659: Agent open issues
- [ ] Distinction between built-in vs MCP tools
- [ ] Configuration of max iterations
- [ ] Configuration of tool response body limits
- [ ] Need `azd mcp tool list` or similar to discover available tools.
- [ ] Handle integration with Azure MCP with `npx` and node aren't available.
- [ ] UX review on `azd mcp` commands, descriptions, etc.
- [ ] Better error handling/retry if agent message fails

wbreza:
Partially resolves #5659


cli/azd/internal/agent/tools/common/utils.go
```diff
@@ -12,10 +17,31 @@ func ToPtr[T any](value T) *T {
...
+	for i, tool := range annotatedTools {
+		rawTools[i] = tool
 	}
-
 	return rawTools
 }
+
+// CreateErrorResponse creates a JSON error response with consistent formatting
+// Used by all IO tools to maintain consistent error response structure
+func CreateErrorResponse(err error, message string) (string, error) {
```
vhvb1989, line 29: nit. If we want to enforce consistent error across IO tools, use the tool interface definition. For example, mandating that `Call()` should return a JSON object.

The current approach is expecting consistency based on:
- Tools must return a valid JSON string and nil for `Call(ctx context.Context, input string) (string, error)`
- Tools should never use the `error`. Instead, they should call this `CreateErrorResponse` to transform an error into a JSON response model for the error.

Anyone creating a new tool who is not aware of this might find inconsistency errors?

vhvb1989 (approved):

cli/azd/internal/agent/tools/common/utils.go
wbreza: Right now this is based on the langhain `Tool` interface which only uses raw string input and `(string, error)` output.  Additionally, we can create our own more robust tool interface that wraps the langchain tool to provide more input/output option that use JSON objects.  I'll take that as a follow-up since this can provide a nicer interface for developers to use.

PR #5663, opened 2025-08-27
Separate publishing to ACR from `azd deploy`

JeffreyCA:
This PR introduces a separate `azd publish` command.

`azd publish` supports these flags:

- `--all` (similar to `azd deploy`)
- `--from-package` (similar to `azd deploy`)
- `--image` - override default container image name
    - Takes precedence over `image` property in azure.yaml:
    ```yml
	name: dev-prod-aca
	services:
	  svc:
	    ...
	    docker:
	      image: 'dev-prod-aca/svc' # This overrides azd default, '{appName}/{serviceName}-{environmentName}'
	```
- `--tag` - override default tag name
    - Takes precedence over `tag` property in azure.yaml:
    ```yml
	name: dev-prod-aca
	services:
	  svc:
	    ...
	    docker:
	      tag: 'latest' # This overrides azd default, 'azd-deploy-{unix time (seconds)}'
	```

------

`azd publish <svc>`

1. Package/build image (skipped if `--from-package` provided)
2. Push to ACR

`azd deploy <svc>`

1. Package/build image  (skipped if `--from-package` provided)
2. Push to ACR (skipped if image+tag already exists in ACR)
3. Create ACA revision

------

## Open questions

1. How should `azd publish` and `azd deploy` behave if the image being pushed already exists in ACR?
    - **Current PR behaviour:** `azd publish` always pushes and will overwrite existing tag if it exists. `azd deploy` skips pushing to ACR if existing tag exists`.
    - Should we add a new flag for `azd deploy` that always pushes to ACR?
    - Or should `azd deploy` default to always overwrite existing tag (old behaviour)? If so, do we need a flag to skip pushing to ACR, like `--skip-publish`?
2. Do the terminologies make sense (tag, image, etc.)? E.g. ACR uses "repository": 
		image


cli/azd/internal/cmd/deploy.go
```diff
@@ -70,6 +71,14 @@ func (d *DeployFlags) bindCommon(local *pflag.FlagSet, global *internal.GlobalCo
 		false,
 		"Deploys all services that are listed in "+azdcontext.ProjectFileName,
 	)
+	local.BoolVar(
+		&d.PublishOnly,
+		"publish-only",
+		false,
+		"Publishes the container image to the registry without deploying the application.",
+	)
```
weikanglim: Would this end up being the same thing as `azd publish`?

cli/azd/internal/cmd/publish.go
```diff
@@ -0,0 +1,196 @@
...
+	if targetServiceName == "" && (p.flags.Image != "" || p.flags.Tag != "") {
+		return nil, errors.New(
+			//nolint:lll
+			"'--tag' and '--image' cannot be specified when publishing all services. Specify a specific service by passing a <service>",
+		)
+	}
+
+	ctx = project.WithImageName(ctx, p.flags.Image)
+	ctx = project.WithImageTag(ctx, p.flags.Tag)
+	return p.deployAction.Run(ctx)
```
weikanglim: This seems a little scary. I'm curious to learn what pushed our implementation down this path.

Conceptually, I could see that deploy is composed of two steps: 1. publish to ACR, and 2. update revision, and I'm leaning towards inverting the composition here.

cli/azd/pkg/project/service_context.go
```diff
@@ -0,0 +1,61 @@
...
+
+const (
+	publishOnlyContextKey contextKey = "publishOnly"
+	imageNameContextKey   contextKey = "imageName"
+	imageTagContextKey    contextKey = "imageTag"
+)
+
+// WithPublishing sets a flag on the context to indicate that the caller is only interested in building and
+// publishing artifacts, not deploying them. This is used by `azd publish`.
+func WithPublishOnly(ctx context.Context, publishOnly bool) context.Context {
```
weikanglim: I wouldn't use context keys here unless if I had to -- I suspect this may be an artifact of the decision of high-level wrapping `publish` around `deploy`.

cli/azd/internal/cmd/publish.go
weikanglim: One immediate fallout that I suspect we'll deal with is that we've traditionally and somewhat intentionally skipped what `package` means for Aspire / .NET publishing [here](https://github.com/Azure/azure-dev/blob/a9a799fcecb24305bdf57298453574e3b0ac9622/cli/azd/pkg/project/service_target_dotnet_containerapp.go#L99), and we would have to reconcile the difference. We can capture this a separate scoped task.

cli/azd/cmd/testdata/TestUsage-azd-publish.snap
```diff
@@ -0,0 +1,36 @@
+
+Publish your project to Azure Container Registry.
+
+  • Only works with Container App services.
```
weikanglim: To make this "dev-friendly", I suspect we'll want to call out `AZURE_CONTAINER_REGISTRY_NAME` as the env dependency in some shape of form.

SESA444562:
Hello,
Good feature.

I don't know if it can be included in this PR or not but It could be great to be able to publish to another docker registry than ACR.
In my enterprise we use a private docker registry and we would like to publish in it.

Thanks for your work

JeffreyCA:
> Hello, Good feature.
> 
> I don't know if it can be included in this PR or not but It could be great to be able to publish to another docker registry than ACR. In my enterprise we use a private docker registry and we would like to publish in it.
> 
> Thanks for your work

Thank you for the input!

I'm testing this scenario out and it looks like it shouldn't be too hard to support in this PR given that [azd deploy supports external registries already today](https://learn.microsoft.com/azure/developer/azure-developer-cli/use-external-registry).

wbreza (commented):
A couple questions and issues to consider.

1. The `ServiceTarget` interface exposes `Package` and `Deploy` functions.  Given this change we should consider adding `Publish` to this interface and implementing it on supported service targets like container app

2. Given a new top level command we should also consider exposing new `publish` project & service hooks so users can more easily integrate with this new capability.

3. Given the separation of publishing from deploy we should consider whether our default `up` workflow should still only call `deploy` or also include `publish` as well.

4. Consider adding `Publish` support for AKS service target (future PR)

cli/azd/internal/cmd/deploy.go
```diff
@@ -70,6 +71,14 @@ func (d *DeployFlags) bindCommon(local *pflag.FlagSet, global *internal.GlobalCo
...
+	// `azd publish` is an alias for this
+	_ = local.MarkHidden("publish-only")
```
wbreza: If we are also adding `azd publish` what is the scenario where we need to add a flag on deploy to only publish?

cli/azd/internal/cmd/publish.go
```diff
@@ -0,0 +1,196 @@
...
```
wbreza: I would consider a full implementation within the `ServiceManager` and expose a `Publish` function like we do for other service target functions.

This would enable the following:
- Support for other service targets to easily opt-in to this capability
- Support for exposing new pre/post `publish` hooks
- Be more consistent and standardized within `azd` service target specific commands.

JeffreyCA: Thanks for the feedback! Yes I'm planning to change this to a "full" implementation instead of the current "publish calling deploy" approach. I initially wanted to get a working PoC without changing too much of the existing code

cli/azd/pkg/project/service_context.go
JeffreyCA: Thanks for the feedback! I'll work on changing this to a "full" implementation instead of the current "publish calling deploy" approach. I initially wanted to get a working PoC without changing too much of the existing code

vhvb1989: Maybe consider creating an extension instead?  It might save you from the full implementation?
up to you.

JeffreyCA: Currently the publish logic is already part of the core `azd deploy` so it's more of a refactoring that needs to be done. I'm not sure it would make the most sense to pull that out into a separate extension or keep it in 2 places.

I'm thinking for other hosts we could explore the extension route

cli/azd/cmd/testdata/TestUsage-azd-publish.snap
JeffreyCA: Updated to mention `AZURE_CONTAINER_REGISTRY_ENDPOINT`

cli/azd/internal/cmd/publish.go
JeffreyCA: Created https://github.com/Azure/azure-dev/issues/5695

cli/azd/pkg/project/container_helper.go
```diff
@@ -168,12 +178,19 @@ func (ch *ContainerHelper) RemoteImageTag(
 	ctx context.Context,
 	serviceConfig *ServiceConfig,
 	localImageTag string,
+	options *PublishOptions,
```
weikanglim: Does this feel unnatural? I was a little surprised to see `RemoteTag()` take in `PublishOptions`.

One observation I can offer: docker supports both local and remote contexts in the image [tag](https://docs.docker.com/reference/cli/docker/image/tag/). 

I wonder if there's a line of thought we could apply here: What if `packagePath` had local vs. remote segments in the path itself?

cli/azd/pkg/project/service_manager.go
```diff
@@ -85,6 +88,17 @@ type ServiceManager interface {
 		options *PackageOptions,
 	) (*ServicePackageResult, error)
 
+	// Publishes the service to make it available to other services
+	// Common examples would be pushing a message to a service bus or
+	// registering the service in a service registry.
```
weikanglim: These examples weren't that useful for me, but curious to hear your thoughts on this.

cli/azd/pkg/project/container_helper.go
```diff
@@ -440,20 +496,48 @@ func (ch *ContainerHelper) runRemoteBuild(
...
-		ctx, target.SubscriptionId(), target.ResourceGroupName(), registryResourceName, contextPath)
+	imageName, err := ch.RemoteImageTag(ctx, serviceConfig, localImageTag, options)
 	if err != nil {
 		return "", err
 	}
 
-	localImageTag, err := ch.LocalImageTag(ctx, serviceConfig)
+	// `azd deploy` skips building if the image already exists in the registry
+	// `azd publish` always overwrites if the image already exists in the registry
+	if options == nil || !options.Overwrite {
```
weikanglim: Can you imagine a way forward to separate the logic out such that this logic wouldn't be "dual-purposed"?

cli/azd/internal/cmd/publish.go
```diff
@@ -0,0 +1,373 @@
...
+}
+
+func GetCmdPublishHelpFooter(*cobra.Command) string {
+	return generateCmdHelpSamplesBlock(map[string]string{
+		"Publish all services in the current project.": output.WithHighLightFormat(
+			"azd publish --all",
+		),
+		"Publish the service named 'api'.": output.WithHighLightFormat(
+			"azd publish api",
+		),
```
weikanglim, line 361: I am thinking out loud and sitting on this, wondering how natural this feels. I feel like it would be nice to have separate atomic commands, that `azd publish api` wouldn't "over-emcompass" the idea of packaging... However, what you have is indeed consistent with our current model.

cli/azd/pkg/project/service_manager.go
```diff
@@ -402,13 +417,63 @@ func (sm *serviceManager) Package(
...
+}
+
 // Deploys the generated artifacts to the Azure resource that will host the service application
 // Common examples would be uploading zip archive using ZipDeploy deployment or
 // pushing container images to a container registry.
 func (sm *serviceManager) Deploy(
 	ctx context.Context,
 	serviceConfig *ServiceConfig,
 	packageResult *ServicePackageResult,
+	publishResult *ServicePublishResult,
```
weikanglim, line 475: I'm personally a little nervous to consider "publish" a lifecycle with its own set of components, and having stack another layer here.

I could imagine a natural evolution here is to create proper data structure around `PackagedService` which perhaps looks like `*ServicePackageResult`, but either just by the path around, or some attribute, indicating whether it's remote. An interesting lead for me is to consider how docker image tags are already shaped today, in this other comment [here](https://github.com/Azure/azure-dev/pull/5663/files#r2335123186)

weikanglim (commented):
I can see how the current implementation is sitting nicely. 

I'm pondering about and happy to discuss with you and the rest of the team some overall design thoughts in the comments that I left.

cli/azd/pkg/project/container_helper.go
JeffreyCA: Do you mind clarifying? Currently the `Overwrite` option is set in `deploy.go` and `publish.go`

cli/azd/internal/cmd/publish.go
JeffreyCA: In that case would `publish` always require `--from-package` to be specified? Or have it depend on the `IMAGE_NAME` service environment variable value?

cli/azd/pkg/project/container_helper.go
JeffreyCA: > What if packagePath had local vs. remote segments in the path itself?
> docker supports both local and remote contexts in the image tag.

Could you provide an example?

Are you suggesting ServicePackageResult.PackagePath be something like:

```go
type PackagePath struct {
    Local  string
    Remote string
}
```

and have `RemoteImageTag` use that instead of generating the remote image tag from `localImageTag` and `PublishOptions`?

weikanglim: I'm going to continue the discussion for the sake of idealization and education. I'm still working through a set of refactorings I could imagine taking place that can help shape these longer term payoff ideas.

In the most simplistic idealistic world, I would start with a URI.

The frontend (whether it goes through publishing, or via a remote `image` reference) would just produce a `PackageService` with `Path`: `example.com:5000/team/my-app:2.0` OR `myimage` OR `imageId`.

The backend is agnostic to how this was done, the concept of "publishing options" wouldn't materialize at this layer.

cli/azd/pkg/project/container_helper.go
weikanglim: With the current implementation, `Publish()` says pushes and image to a remote server, *only if the remote image doesn't already exist*. Technically, the API Is `PublishIfNotAlreadyExists()`.

Another thing that adds to the API behavioral complexity is that the logic is actually embedded deep in the bowels of `runLocalBuild` and `runRemoteBuild`. Sometimes, there are good reasons to do this -- performance is usually a good one. But here, I suspect we're simply following the path of least resistance. 

I am still getting a deeper look into a better overall recommendation for a refactor.

cli/azd/internal/cmd/publish.go
weikanglim: Without diving further into implementation detail, I do believe as I'm staring at my screen at my terminal, my fingers would naturally reach for:
- `azd package api` -- Yay it builds
- `azd publish api` -- Cool, I got it to registry

And right now the behavior is that `azd publish api` actually does a rebuild. 

How does one make this possible?

In the simplest form, I'd just expect some sort of a manifest stored on the filesystem that records the "latest package" that I can just reference. Using environment state is potentially a second option here for me, but I'd have to dive further.

cli/azd/cmd/testdata/TestUsage-azd-deploy.snap
```diff
@@ -11,6 +11,7 @@ Usage
 Flags
         --all                 	: Deploys all services that are listed in azure.yaml
     -e, --environment string  	: The name of the environment to use.
+        --force-publish       	: Forcibly publishes service, overwriting any existing published artifacts.
```
wbreza: I'm a bit confused by this flag.  What is the use case.  Prior to this `azd deploy` always publishes a new container to the registry.  So is this the same behavior with a default value of `true` ? Or am I missing a new scenario?

cli/azd/internal/cmd/publish.go
```diff
@@ -0,0 +1,373 @@
...
+		"Publish the service named 'api' with image name 'app/api' and tag 'prod'.": output.WithHighLightFormat(
+			"azd publish api --image app/api --tag prod",
+		),
+		"Publish the service named 'api' from a previously generated package.": output.WithHighLightFormat(
+			"azd publish api --from-package <image-tag>",
+		),
+	})
```
wbreza: I wonder if we can leverage this new `azd publish` behavior to simplify some container based scenarios where it is now closer to gitops type flow where we can just publish a new "latest" tag for the same image and then the host/service target is responsible for the heavy lifting of fetching the updated container, restarting the instance... Only need a full deploy flow if your spec changes or has additional configuration/dependencies.

cli/azd/pkg/project/container_helper.go
wbreza: We do already have some prior art here around remote vs local tag in the container helper.  Looks like we're reusing that here with `docker.ContainerImage` struct.

cli/azd/pkg/project/container_helper.go
wbreza: It will be interesting to consider if overrwrite should be the default behavior or not.  In many container based tagging strategies you may create publish a new container with multiple tags... version x.x.x and latest.  Versioned tag is not likely to already exist but the latest vesion will exist and user just want to apply this tag to the container.

cli/azd/pkg/project/service_manager.go
wbreza: We are really just teasing out work from deploy since it likely does too much today.  I like the idea of have smaller incremental lifecycle events like build, package, publish, deploy that each have single responsibility per host/service target.  It will also enable developers to provide more precise integration points from hooks/extensions to do what they want to do.

cli/azd/pkg/project/service_manager.go
```diff
@@ -658,6 +684,59 @@ func runCommand[T any](
...
 }
 
+// getTargetResourceForService resolves the target resource for a service configuration.
+// For DotNetContainerAppTarget, it handles container app environment resolution.
+// For other service types, it delegates to the resource manager.
+func (sm *serviceManager) getTargetResourceForService(
+	ctx context.Context,
+	serviceConfig *ServiceConfig,
+) (*environment.TargetResource, error) {
+	if serviceConfig.Host == DotNetContainerAppTarget {
```
wbreza, line 694: Would love to move aspire to its own extension (someday) with its own host/service target so it can just do things it cares about instead of all the special handling throughout `azd`.

wbreza (commented):
Overall this looks amazing! I only have a couple open questions around the following:
- Default overrwrite behavior? Should be true/false if the image/tag already exists?
- Just need to ensure all our e2e tests cases still work to ensure we don't have any regressions.

cli/azd/cmd/testdata/TestUsage-azd-deploy.snap
JeffreyCA: Yeah pretty much, say the user has a hardcoded image and tag in `azure.yaml`, which already exists in the remote container registry.

```yaml
services:
  svc:
    ...
    docker:
      image: 'web'
      tag: latest
```

If the user runs the following sequence of commands, they'd need to set `--force-publish` in order to update the current image in the registry with the newly built one:

1. `azd package web` - produces `web:latest`
2. `azd deploy web --from-package web:latest --force-publish`

Or if using `publish`:

1. `azd package web` - produces `web:latest`
2. `azd publish web --from-package web:latest`
3. `azd deploy web --from-package web:latest`

**If we instead inverted the overwrite behaviour (`deploy` always publishes), then it might look like this instead:**

1. `azd package web` - produces `web:latest`
2. `azd deploy web --from-package web:latest`

Or if using `publish`:

1. `azd package web` - produces `web:latest`
2. `azd publish web --from-package web:latest`
3. `azd deploy web --from-package web:latest --skip-publish`

cli/azd/internal/cmd/publish.go
hemarina: +1. It makes sense that `azd publish` wouldn't rebuild by default. For the "latest package", I'm not sure if I understand. Could you elaborate on when you want to reference it?

cli/azd/pkg/project/service_manager.go
weikanglim: Rephrasing question for clarity: Should `Deploy()` take in both `ServicePackageResult` and `ServicePublishResult`? What does it mean to have both set?

cli/azd/cmd/testdata/TestUsage-azd-deploy.snap
weikanglim: If I needed to vote:
1. `web:latest` is antipattern, see https://learn.microsoft.com/en-us/azure/container-registry/container-registry-image-tag-version#unique-tags. This leans me to not introduce this behavior or flag until we need it later.
2. If we needed to address this as a user feature, do the smart thing -- not a configuration, not a flag. Is there a way to do a delta diff and publish the "most latest" image?

cli/azd/internal/cmd/publish.go
weikanglim: That is interesting. Do you happen to have working scenarios to share?

At a glance, I suspect this may not be easy to do -- the secrets (for example, keyvault secrets) aren't stamped directly into the image.

cli/azd/internal/cmd/publish.go
weikanglim: Based on recent discussions, we're moving forward with rebuilding. I'm happy with it.

cli/azd/cmd/testdata/TestUsage-azd-deploy.snap
JeffreyCA: From this week's earlier discussion, we agreed to change this so that: if the user runs `azd deploy web --from-package` with a fully qualified image reference like `myacr.azurecr.io/web:20250918`, that would instruct `azd deploy` to skip publishing (and packaging as well). That way we don't need to support any additional `--skip-publish` or `--force-publish` deploy flags.

Another suggestion that came up was for `azd publish` to have a single flag to be able to set the destination registry, image, and tag instead of separate `--image` and `--tag`. Something like:

```sh
# Override registry, repository, and tag
azd publish web --from-package web:20250918 --to myacr.azurecr.io/aca/web:latest

# Override repository and tag
azd publish web --from-package web:20250918 --to web:latest

# Override registry and repository
azd publish web --from-package web:20250918 --to myacr.azurecr.io/aca/web

# Override repository only
azd publish web --from-package web:20250918 --to aca/web
```

However, it's a bit unclear how this would support overriding only the tag, or other combinations that separate `--tag` and `--image` supports.

**Update:** I've updated the PR to use this single `--to` flag, only supporting the above 4 combinations.

Any thoughts?

cli/azd/cmd/testdata/TestUsage-azd-publish.snap
```diff
@@ -0,0 +1,37 @@
+
+Publish a service to a container registry.
+
+  • Supports Container App services only.
+  • Target registry set by AZURE_CONTAINER_REGISTRY_ENDPOINT environment variable or docker.registry in azure.yaml.
```
weikanglim: Is `--to [registry]/image` a valid use case with the latest changes?

cli/azd/pkg/project/container_helper.go
weikanglim: Thanks for addressing this!

cli/azd/pkg/project/service_manager.go
weikanglim: I do still believe it's much cleaner to have an idea of a "package" with a locator that is globally addressable, such that `deploy` would just reference a "package" that can be a local image, or an image uploaded to a registry, and this would also hold for things like storage account for files.

However, I'm not blocking the PR for this change, so this looks good to me.

I suspect that over time, we'll notice that the "giant square blocks" of the way we separate by "host providers", rather than low-level primitives to slow us down in implementation and flexibility.

For example, we'll end up having "function app" depend on containerHelper in the future to support run from images. In reality, if we squint hard enough, there are low-level primitives "container-based package" and "file-based package", and these are agnostic across service providers. In other words, build libraries and primitives that compose well together.

cli/azd/cmd/testdata/TestUsage-azd-deploy.snap
weikanglim: The `--to` flag is clever and is what I would vote for today. I think it helps us retain the design of "globally addressable stores", i.e. registry with image/tag OR storage account with blob paths, and is consistent with the `--from-package` design.

weikanglim (approved):
LGTM. Thank you @JeffreyCA for taking the time to address the comments to land this cleanly.

cli/azd/cmd/testdata/TestUsage-azd-publish.snap
JeffreyCA: Yes, updated the bullet point - thanks.

cli/azd/internal/cmd/publish.go
```diff
@@ -0,0 +1,369 @@
...
+func (f *PublishFlags) Bind(local *pflag.FlagSet, global *internal.GlobalCommandOptions) {
+	f.EnvFlag = &internal.EnvFlag{}
+	f.EnvFlag.Bind(local, global)
+	f.global = global
+
+	local.BoolVar(
+		&f.All,
+		"all",
+		false,
+		"Deploys all services that are listed in "+azdcontext.ProjectFileName,
```
vhvb1989: Should this be `Publishes` ? 

```suggestion
		"Publishes all services that are listed in "+azdcontext.ProjectFileName,
```

cli/azd/pkg/azapi/container_registry.go
```diff
@@ -50,6 +51,8 @@ type ContainerRegistryService interface {
 	Credentials(ctx context.Context, subscriptionId string, loginServer string) (*DockerCredentials, error)
 	// Gets a list of container registries for the specified subscription
 	GetContainerRegistries(ctx context.Context, subscriptionId string) ([]*armcontainerregistry.Registry, error)
+	// Checks if a specific image tag exists in the container registry
+	TagExists(ctx context.Context, subscriptionId string, loginServer string, repository string, tag string) (bool, error)
```
vhvb1989: Is this used somewhere? Seems like maybe you needed at some point but then you stopped using it maybe?  (but stayed here)?  Should we remove it?

cli/azd/pkg/project/container_helper.go
```diff
@@ -395,13 +414,41 @@ func (ch *ContainerHelper) runLocalBuild(
 	return remoteImage, nil
 }
 
+type ImageOverride docker.ContainerImage
```
vhvb1989: Looks like we should not be exporting this type?   Should we rename it to `imageOverride` ?

cli/azd/pkg/project/service_models.go
```diff
@@ -96,9 +96,36 @@ func (spr *ServicePackageResult) MarshalJSON() ([]byte, error) {
...
+	RemoteImage string `json:"remoteImage"`
+}
+
+// Supports rendering messages for UX items
+func (spr *ServicePublishResult) ToString(currentIndentation string) string {
+	if containerDetails, ok := spr.Details.(*ContainerPublishDetails); ok && containerDetails.RemoteImage != "" {
+		return fmt.Sprintf("%s- Remote Image: %s\n", currentIndentation, output.WithLinkFormat(containerDetails.RemoteImage))
+	}
+
+	return ""
```
vhvb1989, line 119: is it expected to print an empty string if the remote image is empty?  or should that be a different output?  
Maybe add a comment about why the empty string makes sense there and is expected?

cli/azd/pkg/project/service_models.go
```diff
@@ -96,9 +96,36 @@ func (spr *ServicePackageResult) MarshalJSON() ([]byte, error) {
...
+}
+
+func (spr *ServicePublishResult) MarshalJSON() ([]byte, error) {
+	return json.Marshal(*spr)
```
vhvb1989, line 123: We use `output.EventForMessage()` for mostly all UX components to remove the colors and ascii formats. See examples in the `pkg/output/ux` folder

cli/azd/pkg/project/service_target.go
```diff
@@ -77,11 +77,22 @@ type ServiceTarget interface {
...
 
+	// Publish pushes the prepared artifacts without performing deployment.
+	Publish(
+		ctx context.Context,
+		serviceConfig *ServiceConfig,
+		frameworkPackageOutput *ServicePackageResult,
+		targetResource *environment.TargetResource,
+		progress *async.Progress[ServiceProgress],
+		publishOptions *PublishOptions,
+	) (*ServicePublishResult, error)
```
vhvb1989, line 88: I don't love how we keep this interface growing 🥲
Only a few serviceTargets would actually implement this contract while the others will ignore.

(No action for this comment... just sharing a feeling here LOL)

cli/azd/internal/cmd/publish.go
JeffreyCA: Yes, thanks for catching this!

cli/azd/pkg/project/service_models.go
JeffreyCA: Thanks, marshalling the entire struct is consistent with the other service result structs defined in this file, like `ServicePackageResult` and `ServiceDeployResult` so I'm inclined to keep it this way.

image

cli/azd/pkg/azapi/container_registry.go
JeffreyCA: Thanks, removed it

Issue #5473, opened 2025-07-10
azd login with federated credential fails because the subject is scoped to the GitHub environment name


puicchan:
GitHub environments allows for gated approval stages.

Now `azd login` with federated credential fails because the subject is scoped to the GitHub environment name:

AADSTS700213: No matching federated identity record found for presented assertion subject 'repo:JeffreyCA/azd-dev-prod:environment:dev'. Check your federated identity credential Subject, Audience and Issuer against the presented assertion.

weikanglim:
Reading through the GitHub [OpenID Connect docs](https://docs.github.com/en/actions/reference/openid-connect-reference), I'm not immediately seeing how azd would influence the subject used here.

I believe the [subject claims ](https://docs.github.com/en/actions/reference/openid-connect-reference#example-subject-claims)are automatically populated based on the runtime context of the GitHub workflow.

As a first step to move towards resolution, I'd recommend using the [oidc-debugger](https://github.com/github/actions-oidc-debugger) to see what the auto-generated subject looks like. My hunch is that the workflow is running in a non-environment context that wouldn't match the Azure configuration.

JeffreyCA:
With oidc-debugger, the value of `sub` when an environment is not specified looks like: `repo:JeffreyCA/dev-prod-single:ref:refs/heads/main` (triggered from push to main branch)

When an environment is specified (e.g. `prod`), then `sub` looks like: `repo:JeffreyCA/dev-prod-single:environment:prod`.

```yaml
  oidc_debug_test:
    runs-on: ubuntu-latest
    environment:
      name: prod
    permissions:
      contents: read
      id-token: write
    name: A test of the oidc debugger
    steps:
      - name: Debug OIDC Claims
        uses: github/actions-oidc-debugger@main
        with:
          audience: '${{ github.server_url }}/${{ github.repository_owner }}'
```

I was exploring the possibility of `azd pipeline config` generating an `azure-dev.yml` GitHub Action workflow that sets `environment` to `${{ vars.AZURE_ENV_NAME }}` and as part of the "**Federated User Managed Identity (MSI + OIDC)**" setup flow, create an additional federated credential with the subject identifier `repo:JeffreyCA/dev-prod-single:environment:<env-name>`.

If azd doesn't do this for the user, then I think they have to manually create the federated credential in Azure for `azd login` to work if they're using GitHub Environments. Not sure if that clarifies things?

weikanglim:
[...]

Issue #5329, opened 2025-06-16
Environment-Specific CI/CD Integration


wbreza:
**Summary**: Align environment types with CI/CD platforms to enable environment promotion, scoped access, and approval workflows.

**Acceptance Criteria**:

- [ ] Integration with GitHub Environments
- [ ] Integration with Azure DevOps Environments
- [ ] Environment promotion workflows (dev → test → prod)
- [ ] Scoped access control per environment type
- [ ] Approval gates for production environments
- [ ] Environment-specific deployment strategies
- [ ] Audit trails for environment changes

ronaldbosma:
@wbreza Would it also make sense to create the `AZURE_CLIENT_ID`, `AZURE_TENANT_ID` and `AZURE_SUBSCRIPTION_ID` settings as secrets instead of variables when running `az pipeline config`? This seems to align better with the approach described in the official documentation: [Use the Azure Login action with OpenID Connect - Create GitHub Secrets](https://learn.microsoft.com/en-us/azure/developer/github/connect-from-azure-openid-connect#create-github-secrets).

Issue #2373, opened 2023-06-07
Workflow for multi-environment management + promoting releases between environments


savannahostrowski:
Developers may want different IaC for each environment (dev, test, prod). 
- What is our guidance for managing that? 
- How do we enable a developer to advance their app from one environment to another (e.g. dev -> prod)?

Note that using branches for management is sort of an anti-pattern/dated approach and there are developers who do not view this as an option (e.g. https://codefresh.io/blog/stop-using-branches-deploying-different-gitops-environments/)

weikanglim:
I would lean heavily into our CI providers: [GitHub](https://docs.github.com/en/actions/deployment/targeting-different-environments/using-environments-for-deployment) and [Azure Pipelines](https://learn.microsoft.com/en-us/azure/devops/pipelines/process/environments?view=azure-devops), and provide a mechanism that allows `azd` environments to easily be reused in those scenarios.

wbreza:
@savannahostrowski @ellismg This likely should be required as part of our ADE integration

ellismg:
> What is our guidance for managing that?
> How do we enable a developer to advance their app from one environment to another (e.g. dev -> prod)?

Our current strategy allows you to set per environment configuration for your IAC, so at a minimum you could imagine adding a new parameter to your infrastructure on what sort of environment type (dev, prod, etc) you were deploying to and your infrastructure template could conditionalize resources on that.

Or, you could just examine `AZURE_ENV_NAME` in your bicep and based on it's structure do things (perhaps your organization has a way to name environments that encodes some semantics).

I think that can get us very far today.  To @weikanglim's point, you can define these values in your CI system (perhaps tied to branches, perhaps now).

It would be good to know where this falls over or where the rough edges are and we can start to sand them down.

> @savannahostrowski @ellismg This likely should be required as part of our ADE integration

This sort of surprised me - given that some of the initial design work in the ADE stream led me to believe that in ADE mode we'd only allow a single environment per project.  What do you think is going to push us to figure this out for ADE, @wbreza?

wbreza:
> This sort of surprised me - given that some of the initial design work in the ADE stream led me to believe that in ADE mode we'd only allow a single environment per project. What do you think is going to push us to figure this out for ADE, @wbreza?

ADE is all about separating out the governance and creation of infrastructure typically required in a more structured organization or enterprise.  Devs won't have access to provision resources, but if the enterprise wants to leverage ADE + AZD for integration scenarios we should have a good story here.

[...]

PR #5664, opened 2025-08-28
GitHub Environments & Promotion Pipeline

73dward5:
### Overview
This adds optional flags to `azd pipeline config` to allow pushing environment variables/secrets to the GitHub environment instead of the repo level variables/secrets and federating the identity with that environment instead of the branch or PR.

Adds:
- `--github-use-environments` to generate a GitHub Actions workflow using GitHub Environments (single or matrix when multiple azd envs).
- `--force-regenerate` to overwrite the existing workflow unconditionally.

Also: migration heuristics, federated credential pruning, repo variable cleanup, updated snapshots, lint/spelling fixes.

### Key Changes
- Flags: `--github-use-environments`, `--force-regenerate`
- Workflow template: conditional matrix, `max-parallel: 1`, secret/variable validation
- Auto-regeneration when env-mode mismatch detected
- Prunes legacy federated credentials (branch / PR subjects) for SP & MSI
- Removes redundant repo-level variables after migration
- Manager helpers: `EnableGitHubEnvironments`, `ForceRegenerateWorkflow`
- Snapshot & usage/help updates
- Typo fix: `progess` → `progress`
- All `golangci-lint` issues resolved

### Behavior & Migration
| Scenario | Result |
|----------|--------|
| Legacy workflow + opt-in | Regenerates with env matrix (if >1 env) |
| Env workflow + opt-out | Regenerates to legacy single |
| Any state + `--force-regenerate` | Always rewrites |

### Pruning Scope
Only deletes federated credentials with `:pull_request` or `:ref:refs/heads/` (keeps `:environment:` subjects).

### Testing
- Short tests pass
- Updated snapshots pass
- Functional failures unrelated (local env missing .NET 8, expired Azure auth token)
- Lint: clean

### Risks & Mitigations
| Risk | Mitigation |
|------|------------|
| Unexpected overwrite | Requires explicit flag or mismatch |
| Over-pruning | Filters to known legacy subjects |
| Env var API limitations | Heuristic delete for known migrated names |

### Follow-Ups
- Docs for env vs legacy mode
- Release notes entry
- (Optional) cspell dictionary tuning

### Checklist
- [x] Flags
- [x] Template logic
- [x] Migration + pruning
- [x] Tests & snapshots
- [x] Lint/spelling
- [ ] Docs
- [ ] Release notes

### Release Notes Blurb
Adds opt-in GitHub Environments workflow generation (`--github-use-environments`) and `--force-regenerate` to forcibly recreate workflows, with automatic migration and cleanup of


73dward5:
* [[5664]](https://github.com/Azure/azure-dev/pull/5664) GitHub pipeline config: supports setting secrets & variables in a GitHub Environment named after the current azd environment (automatically created if needed) and adds a matching OIDC federated credential subject (`repo:<owner>/<repo>:environment:<AZURE_ENV_NAME>`). (Previously configurable via `AZD_GITHUB_ENV`; that variable is now ignored.) Fixes [#5473](https://github.com/Azure/azure-dev/issues/5473).
* [[5664]](https://github.com/Azure/azure-dev/pull/5664) Added optional `--github-use-environments` flag to `azd pipeline config` to emit GitHub Environment (and matrix when multiple) blocks. Disabled by default; existing workflows remain unchanged unless flag is provided. Re-running with the flag toggled will migrate the existing workflow by adding or removing the environment/matrix section. Addresses matrix & multi‑environment ask in [#2373](https://github.com/Azure/azure-dev/issues/2373) and community feedback in [discussion #3585](https://github.com/Azure/azure-dev/discussions/3585).
* [[5664]](https://github.com/Azure/azure-dev/pull/5664) Environment mode now migrates standard AZD pipeline variables to the GitHub Environment scope (removing repo-level duplicates) and generates only the environment federated credential subject. Related to [#5473](https://github.com/Azure/azure-dev/issues/5473).
* [[5664]](https://github.com/Azure/azure-dev/pull/5664) Automatic pruning of legacy branch & pull_request federated identity credentials when switching to environment-only mode (service principal & MSI identities). Part of cleanup for [#5473](https://github.com/Azure/azure-dev/issues/5473).

73dward5:

73dward5:
I am fine with this PR being rejected, as long as you can derive some benefit from the code to implement GitHub environments.

weikanglim:
Thanks @73dward5 for taking the time to submit this PR!

I do like inspirationally where this is going, and this does help illuminate certain design details upfront for some work we have in the backlog with #5329. 

I'm curious to learn from your experience:

1. How was your experience with `azd pipeline config` looping through all environments? `azd pipeline config` traditionally targets a single environment. I could see a different implementation where `azd pipeline config` makes a YAML modification to the environment matrix, instead of regenerating it from scratch. But this may also introduce other potential drawbacks, cc: @vhvb1989 .
2. Was the generated matrix (run environment deployments in parallel) "good enough" for your use-case, or were you expecting some idea of promotion between environments, with perhaps a build stage that spawns deployment jobs?
3. Was it a good experience when using `AZURE_ENV_NAME` as the github environment name? This has come up in discussions in the past, because azd currently expects `AZURE_ENV_NAME` to be unique per subscription, so you may end up having something closer like `myproject-dev` rather than `dev`.

Thanks for taking the time to get this PR out and sharing your experience!

73dward5:
1. How was your experience with azd pipeline config looping through all environments? azd pipeline config traditionally targets a single environment. I could see a different implementation where azd pipeline config makes a YAML modification to the environment matrix, instead of regenerating it from scratch. But this may also introduce other potential drawbacks, cc: @vhvb1989 .

- Having one pipeline for all environments does a couple of things. One, it helps keep the pipeline code DRY and free from duplication or misconfiguration between environments. Secondly, it allows for the environments to be promotable without having to orchestrate the environment promotions with separate pipelines. Or a branching strategy that ends up being closer to your deployment strategy, rather than staying true to your branching and contribution model.

3. Was the generated matrix (run environment deployments in parallel) "good enough" for your use-case, or were you expecting some idea of promotion between environments, with perhaps a build stage that spawns deployment jobs?

- The generated matrix works well as long as the `max-parallel` is `1`, ensuring the environments do in fact stay promotable (e.g. `Dev` finishes, then `Test`, then `Prod`) instead of trying to run them all at the same time. In theory, the build should only need to run once before the deployment. However, it can be more challenging for some languages/tools to separate the environment variables from the build artifacts. For example, Terraform would need the tfvars for the plan, or a Node.js SPA might need the environment variables to be in the js. Other languages, like .NET, can be built, and then environment variables can be injected on the server (during the deployment) after the artifact is created.

5. Was it a good experience when using AZURE_ENV_NAME as the github environment name? This has come up in discussions in the past, because azd currently expects AZURE_ENV_NAME to be unique per subscription, so you may end up having something closer like myproject-dev rather than dev.

- I think `AZURE_ENV_NAME` as the github environment name works great to provide your current environment working context locally, but if you are building a promotable pipeline, you may want to know about all of your environments. In my opinion, if you are publishing a pipeline, you are essentially moving the build/deploy responsibility to CI/CD and no longer managing that process from your local machine. I do think that allowing custom environment names like myproject-dev rather than dev does contribute to developer flexibility and autonomy. With that flexibility, you do sort of run into some complexities in programmatically determining the promotion order, though.

PR #5666, opened 2025-08-28
Adds protection to agents built-in file system tools

wbreza:
Adds protection to agents built-in file system tools to only allow actions against the current workspace / working directory.

### DIFF
--- a/cli/azd/cmd/container.go
+++ b/cli/azd/cmd/container.go
@@ -23,6 +23,7 @@ import (
 	"github.com/azure/azure-dev/cli/azd/internal"
 	"github.com/azure/azure-dev/cli/azd/internal/agent"
 	"github.com/azure/azure-dev/cli/azd/internal/agent/consent"
+	"github.com/azure/azure-dev/cli/azd/internal/agent/security"
 	"github.com/azure/azure-dev/cli/azd/internal/cmd"
 	"github.com/azure/azure-dev/cli/azd/internal/grpcserver"
 	"github.com/azure/azure-dev/cli/azd/internal/repository"
@@ -556,6 +557,16 @@ func registerCommonDependencies(container *ioc.NestedContainer) {
 	container.MustRegisterScoped(consent.NewConsentManager)
 	container.MustRegisterNamedSingleton("ollama", llm.NewOllamaModelProvider)
 	container.MustRegisterNamedSingleton("azure", llm.NewAzureOpenAiModelProvider)
+
+	// Agent security manager
+	container.MustRegisterSingleton(func() (*security.Manager, error) {
+		// Use current working directory as security root
+		cwd, err := os.Getwd()
+		if err != nil {
+			return nil, fmt.Errorf("failed to get current working directory: %w", err)
+		}
+		return security.NewManager(cwd)
+	})
 
 	container.MustRegisterSingleton(repository.NewInitializer)
 	container.MustRegisterSingleton(alpha.NewFeaturesManager)
--- a/cli/azd/internal/agent/security/manager.go
+++ b/cli/azd/internal/agent/security/manager.go
@@ -1,0 +1,109 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package security
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+	"sync"
+)
+
+// Manager handles path validation and access control for agent tools
+type Manager struct {
+	securityRoot string
+	mu           sync.RWMutex
+}
+
+// NewManager creates a new security manager with the specified root directory
+func NewManager(rootPath string) (*Manager, error) {
+	// Convert to absolute path and clean it
+	absRoot, err := filepath.Abs(rootPath)
+	if err != nil {
+		return nil, fmt.Errorf("failed to resolve security root path %s: %w", rootPath, err)
+	}
+
+	// Clean the path to normalize it
+	cleanRoot := filepath.Clean(absRoot)
+
+	// Verify the directory exists
+	info, err := os.Stat(cleanRoot)
+	if err != nil {
+		return nil, fmt.Errorf("security root directory does not exist: %s: %w", cleanRoot, err)
+	}
+
+	if !info.IsDir() {
+		return nil, fmt.Errorf("security root path is not a directory: %s", cleanRoot)
+	}
+
+	return &Manager{
+		securityRoot: cleanRoot,
+	}, nil
+}
+
+// ValidatePath validates that the given path is within the security boundary
+func (sm *Manager) ValidatePath(inputPath string) (string, error) {
+	sm.mu.RLock()
+	defer sm.mu.RUnlock()
+
+	// Convert to absolute path
+	absPath, err := filepath.Abs(inputPath)
+	if err != nil {
+		return "", fmt.Errorf("failed to resolve path: %w", err)
+	}
+
+	// Clean the path to resolve any . or .. components
+	cleanPath := filepath.Clean(absPath)
+
+	// Check if the path is within the security root
+	if !sm.isWithinSecurityRoot(cleanPath) {
+		return "", fmt.Errorf("access denied: path outside allowed directory")
+	}
+
+	return cleanPath, nil
+}
+
+// ValidateDirectoryChange validates a directory change operation
+// This allows navigation within the security root boundary
+func (sm *Manager) ValidateDirectoryChange(inputPath string) (string, error) {
+	sm.mu.RLock()
+	defer sm.mu.RUnlock()
+
+	// Convert input to absolute path
+	absPath, err := filepath.Abs(inputPath)
+	if err != nil {
+		return "", fmt.Errorf("failed to resolve path: %w", err)
+	}
+
+	// Clean the path
+	cleanPath := filepath.Clean(absPath)
+
+	// Check: must be within security root
+	if !sm.isWithinSecurityRoot(cleanPath) {
+		return "", fmt.Errorf("access denied: cannot navigate outside project directory")
+	}
+
+	// Allow navigation within the security root
+	return cleanPath, nil
+}
+
+// GetSecurityRoot returns the current security root directory
+func (sm *Manager) GetSecurityRoot() string {
+	sm.mu.RLock()
+	defer sm.mu.RUnlock()
+	return sm.securityRoot
+}
+
+// isWithinSecurityRoot checks if the given path is within the security root
+func (sm *Manager) isWithinSecurityRoot(path string) bool {
+	// Normalize both paths for comparison
+	relPath, err := filepath.Rel(sm.securityRoot, path)
+	if err != nil {
+		return false
+	}
+
+	// If the relative path starts with "..", it's outside the security root
+	return !strings.HasPrefix(relPath, "..") && relPath != ".."
+}
--- a/cli/azd/internal/agent/security/manager_test.go
+++ b/cli/azd/internal/agent/security/manager_test.go
@@ -1,0 +1,107 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package security
+
+import (
+	"os"
+	"path/filepath"
+	"testing"
+)
+
+func TestSecurityManager_ValidatePath(t *testing.T) {
+	// Create a temporary directory for testing
+	tempDir, err := os.MkdirTemp("", "security_test")
+	if err != nil {
+		t.Fatalf("Failed to create temp dir: %v", err)
+	}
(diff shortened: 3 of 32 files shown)


cli/azd/cmd/container.go
```diff
@@ -557,6 +558,16 @@ func registerCommonDependencies(container *ioc.NestedContainer) {
 	container.MustRegisterNamedSingleton("ollama", llm.NewOllamaModelProvider)
 	container.MustRegisterNamedSingleton("azure", llm.NewAzureOpenAiModelProvider)
 
+	// Agent security manager
+	container.MustRegisterSingleton(func() (*security.Manager, error) {
+		// Use current working directory as security root
+		cwd, err := os.Getwd()
```
vhvb1989, line 564: should we care about the `--cwd` flag?  I think it's a global flag

cli/azd/internal/agent/security/manager.go
```diff
@@ -0,0 +1,147 @@
...
+func (sm *Manager) safeJoin(relativePath string) (string, error) {
+	// Join with the security root
+	joined := filepath.Join(sm.securityRoot, relativePath)
+
+	// Resolve to canonical form
+	return resolvePath(joined)
+}
+
+// resolvePath resolves a path to its canonical form, handling symlinks consistently
+func resolvePath(inputPath string) (string, error) {
```
vhvb1989, line 86: Something to consider:  Azd supports running within a project's subfolder. For example, running `azd deploy` inside a service folder is supported and is the same as running `azd deploy serviceName` from the project root (next to azure.yaml).

So I wonder if those cases would should be considered by this security manager.

cli/azd/internal/agent/security/manager.go
```diff
@@ -0,0 +1,137 @@
...
+
+	return &Manager{
+		securityRoot: resolvedRoot,
+	}, nil
+}
+
+// GetSecurityRoot returns the current security root directory
+func (sm *Manager) GetSecurityRoot() string {
+	sm.mu.RLock()
+	defer sm.mu.RUnlock()
```
vhvb1989: From copilot:

-----------------
Based on my analysis of the code, there is currently no risk that justifies the use of the RLock for the securityRoot string. Here's what I found:

Current State Analysis:
No Setter Methods: The Manager struct has no methods that modify the securityRoot field after initialization. The only methods are:

NewManager() - Sets the field during construction
GetSecurityRoot() - Reads the field
ValidatePath() - Reads the field
safeJoin() - Reads the field
Field is Set Once: The securityRoot field is only set during struct initialization in NewManager() and never modified afterward.

No External Modification: Looking at all usages of the Manager type across the codebase, there are no external functions or methods that modify the securityRoot field.

Strings are Immutable: In Go, strings are immutable, so even if multiple goroutines read the field simultaneously, there's no risk of data corruption.

Conclusion:
The sync.RWMutex and RLock() calls are unnecessary in the current implementation. Since:

The securityRoot field is never modified after initialization
Strings in Go are immutable and safe for concurrent reads
There are no setter methods or external modifications

------

cli/azd/internal/agent/security/manager.go
```diff
@@ -0,0 +1,137 @@
...
+		}
+	} else {
+		// Handle relative paths - use safeJoin approach
+		resolvedPath, err = sm.safeJoin(inputPath)
+		if err != nil {
+			return "", err
+		}
+	}
+
+	// Verify prefix relationship - the path should start with the security root
```
vhvb1989, line 65: nit.  We don't usually log to `os.Stderr` like this. I think we would either use `log.debug` or `fmt.Fprintf()` taking the IOWriter from the input.Console or as an injected parameter for your component, allowing to change the IOWriter for the component.

cli/azd/internal/agent/security/manager_test.go
```diff
@@ -0,0 +1,346 @@
...
+
+	// Test case 1: Multiple non-existent levels
+	// tempDir/existing/nonexistent1/nonexistent2/file.txt
+	deepPath := filepath.Join(existingDir, "nonexistent1", "nonexistent2", "file.txt")
+	resolved := resolvePartialPath(deepPath)
+
+	// Should resolve existingDir and append the non-existent parts
+	resolvedExisting, err := filepath.EvalSymlinks(existingDir)
+	if err != nil {
+		t.Fatalf("Failed to resolve existing dir: %v", err)
```
vhvb1989, line 286: nit. Should we use `request.noError(t, err)` instead of this?

cli/azd/internal/agent/security/manager.go
```diff
@@ -0,0 +1,137 @@
...
+	// Split the path into components to work backwards
+	var pathComponents []string
+	currentPath := fullPath
+
+	// Collect all path components from the full path back to an existing directory
+	for currentPath != filepath.Dir(currentPath) { // Stop when we reach root
+		pathComponents = append([]string{filepath.Base(currentPath)}, pathComponents...)
+		currentPath = filepath.Dir(currentPath)
+
+		// Try to resolve this level
```
vhvb1989, line 114: Took me awhile to get this function XD.  Finally got it with some AI help.

----
Recommendation:
The resolvePartialPath function is likely unnecessary for the security use case. The standard library's filepath.Abs() + filepath.Clean() should be sufficient because:

Security validation doesn't require symlink resolution of non-existent files
filepath.Abs() already handles non-existent paths correctly
The complex symlink resolution logic introduces bugs and complexity
The current implementation seems over-engineered for what's actually needed in a security context.

-----

suggested:
```go
func resolvePath(inputPath string) (string, error) {
    // This handles non-existent paths just fine
    absPath, err := filepath.Abs(inputPath)
    if err != nil {
        return "", fmt.Errorf("failed to resolve path: %w", err)
    }
    
    // Clean the path to handle . and .. components
    cleanPath := filepath.Clean(absPath)
    
    // Try to resolve symlinks for existing parts only
    if resolvedPath, err := filepath.EvalSymlinks(cleanPath); err == nil {
        return resolvedPath, nil
    }
    
    // If it doesn't exist, just return the clean absolute path
    return cleanPath, nil
}
```

cli/azd/internal/agent/security/manager.go
```diff
@@ -0,0 +1,137 @@
...
+		resolvedPath = resolvePartialPath(absPath)
+	}
+
+	return resolvedPath, nil
+}
+
+// resolvePartialPath recursively finds the deepest existing part of a path and resolves it
+func resolvePartialPath(fullPath string) string {
+	// Split the path into components to work backwards
+	var pathComponents []string
```
vhvb1989, line 106: this might not be required...  See the comment in this method about it.
It might be possible to just bring `filepath.Clean()` to this function before calling `EvalSymlinks()`

cli/azd/internal/agent/security/manager_test.go
```diff
@@ -0,0 +1,346 @@
...
+	}
+
+	// Should resolve up to level2 and append the rest
+	expected = filepath.Join(resolvedLevel2, "level3", "level4", "nonexistent.txt")
+	if resolved != expected {
+		t.Errorf("Expected %s, got %s", expected, resolved)
+	}
+}
+
+func TestResolvePartialPath_Recursive(t *testing.T) {
```
vhvb1989, line 268: Is this test really covering the purpose of the path with symlinks resolution even without using symlinks?  

Should we try to add symlinks here? Like making it handle things with `/path/../path/foo/../foo` ?

cli/azd/cmd/container.go
wbreza: When `--cwd` is set it basically changes the working directory in the persistent prefun cobra func.  I will validate this works as expected.

cli/azd/internal/agent/security/manager.go
wbreza: True - but it also run in folders where an `azure.yaml` doesn't already exist.  We'll have to identify the balance that we want to have.

cli/azd/internal/agent/security/manager.go
wbreza: Goot catch - early on in my feature there was more happening.  Now this shouldn't be required.

cli/azd/internal/agent/security/manager.go
wbreza: You may have been reviewing when I was debugging some weird macos issues in CI.  I have removed the debug logging.

cli/azd/internal/agent/security/manager_test.go
wbreza: Same as above - logging has been removed or uses our standard logging.

cli/azd/internal/agent/security/manager.go
wbreza: There were some very tricky things happening, especially in macos.  symlinks are used a lot more in macos than other OSes.

I've observed that `filepath.Abs(...)` internally resolves symlinks.  In some cases the `inputPath` being validate does not yet exist... like creating a file in a new folder tree for example.  So we attempt to recurse up the folder path till we have a part of the path that does exist and apply the consistent resolution strategy.

I've spent too many cycles on this and now have this working implementation that works across all our target OSes.

cli/azd/internal/agent/security/manager_test.go
wbreza: In macos `/tmp` is actually a symlink to `/private/tmp` so whether or not the test is actually creating symlinks in MacOS the symlinks are being created during the tests temp folder creation.

This validates consistency in our handling across OSes.

cli/azd/internal/agent/security/manager.go
wbreza: I've spent many cycles debugging cross OS issues related to the resolution of the paths.  The `resolvePartialPath` comes into play when the path being validated does not fully exist.  Examples - creating new folder/file in a folder.  We still want to walk up the tree and resolve the closes existing path and then append the new path components.  This ensures all symlinks are validated propertly.

cli/azd/internal/agent/tools/io/security.go
```diff
@@ -0,0 +1,189 @@
...
+
+	return initErr
+}
+
+// GetSecurityManager returns the global security manager instance
+func GetSecurityManager() *SecurityManager {
+	if globalSecurityManager == nil {
+		// If not explicitly initialized, try to initialize with current working directory
+		if cwd, err := os.Getwd(); err == nil {
+			_ = InitializeSecurityManager(cwd)
```
hemarina: May I ask why we skip the error for InitializeSecurityManager?

wbreza: This is actually an old version! I moved it out into a `security` package.  I will go ahead and remove this file.

cli/azd/cmd/container.go
```diff
@@ -557,6 +558,16 @@ func registerCommonDependencies(container *ioc.NestedContainer) {
 	container.MustRegisterNamedSingleton("ollama", llm.NewOllamaModelProvider)
 	container.MustRegisterNamedSingleton("azure", llm.NewAzureOpenAiModelProvider)
 
+	// Agent security manager
+	container.MustRegisterSingleton(func() (*security.Manager, error) {
```
weikanglim, line 562: nit: This feels like an implementation detail of the agent factory, and could be moved there. 

`agent_factory.go`:

```golang
func (f *AgentFactory) Create(opts ...AgentCreateOption) (Agent, error) {
   securityManager := security.NewManager(cwd)
   ...
   localtools.NewLocalToolsLoader(securityManager),
}
```

cli/azd/internal/agent/security/manager.go
weikanglim: In my mind, to maximize correctness and provide a simple provable working solution, there's both:
- `ValidatePath`: validate a static path is within root, and
- `ValidateFsPathBeforeOpen`: validate a filesystem path is within root before calling `Open()` -- this would additionally resolve the symlink and fail if the symlink points to a non-existing file/directory

I would still be a little nervous on my end to implement this *correctly*, and I do value correctness a lot in a security feature...

Have we considered using [os.Root](https://go.dev/blog/osroot) (new to 1.24, enhanced in 1.25) to implement this security feature?

weikanglim (commented):
If @wbreza thinks what we have is good enough, I'm okay moving forward with an alhpa feature.

I would recommend creating an issue for aligning with other agents on implementation at some point when the features gets closer to non-alpha state.

cli/azd/cmd/container.go
wbreza: Technically we could move it there but I'd rather keep it where it is so that we continue to follow the same initialization strategy and use the `cwd` at time of `azd` initialization.

cli/azd/internal/agent/security/manager.go
wbreza: `os.Root` looks like a great way to add some additional security here. Added #5689 to track this.

PR #5688, opened 2025-09-04
azd metadata docs

vhvb1989:
Documenting supported azd metadata fields for bicep input parameters


vhvb1989:
This is now part of the official docs: https://learn.microsoft.com/en-us/azure/developer/azure-developer-cli/metadata

Issue #3116, opened 2023-12-13
[Spike] improve azure container app deployments to avoid error-prone and confusing multiple revisions


vicondoa:
- [X] Make sure you've installed the latest version using [instructions in the wiki](../wiki/install)

**Output from `azd version`**
Run `azd version` and copy and paste the output here:
azd version 1.5.0 (commit 012ae734904e0c376ce5074605a6d0d3f05789ee)

**Describe the bug**
Azure Container Apps with Azure Developer CLI currently have a major limitation in how it handles deployments. Deployments of updated revisions occur in two phases: provision where the spec for the container app is changed and deployment where the image is updated. This creates a problem when developers revise their service and bicep configs, as it often has dependencies between its code in its Docker container, environment variables, resources, mounts, health probes, load balancers, and more. If these dependencies are not deployed in the right order, deployments can fail, and this can possibly create an outage because the container app can't start up. 

`azd deploy` will only replace `containers[0].image` with the new image. This undocumented behavior is a huge limitation in the case where someone would like to run the same image in a sidecar or init container. A developer shouldn't need to have a code change made to `azd` anytime they come up with a new way an image should be included in a a container app spec.

The current setup is confusing. It's unclear what the special `exists` behavior is for in the templates. In a team of 10 people, most of us were really surprised to find out that the bicep configs for a container app don't actually determine what image should be used.

We mitigate this issue by adding a post-deploy hook to rollout a third revision to copy the image on `containers[0]` to all other containers where it's needed. A container app should not need three revisions per deployment. Each revision is an opportunity of incompatibility between bicep spec, revision 1, and revision 2 that is very error prone. We have burnt many hours of engineer time working around this.

**To Reproduce**
1. create a container app that has two containers, where you would like it to use the built image on both
2. run `azd up`
3. notice how only the first container has changed in the container app

**Expected behavior**

`azd up` should only produce one revision per service. developer should be able to specify container app spec and how the images are used all in one place rather than the undocumented behavior and `exists` hack in the template. Consider how kubernetes deployments work with kubectl apply.

**Environment**
Information on your environment:
    * Azure container apps
    * python

Petermarcu:
@ellismg is some of this related to the getting started scenario where some people don't have ACR or containers already in a registry to point to at the time of provisioning the ACA infra?

vhvb1989:
[...]

Issue #5352, opened 2025-06-17
Container Apps Revision Strategy


wbreza:
**Summary**: Implement revision-based deployment strategy for Azure Container Apps with traffic splitting and gradual rollout capabilities.

**Acceptance Criteria**:

- [ ] Deploy new version as Container Apps revision
- [ ] Gradual traffic routing to new revision
- [ ] Configurable traffic splitting percentages
- [ ] Health check integration before traffic shift
- [ ] Revision naming and labeling strategy
- [ ] Rollback to previous revision capability
- [ ] Integration with Container Apps revisions API

weikanglim:
My current thinking is that we will want to support these in a way that feels natural and doesn't deviate too much from what is currently documented as best practices for container app, i.e. [blue-green deployments](https://learn.microsoft.com/en-us/azure/container-apps/blue-green-deployment?pivots=bicep#create-a-container-app-with-multiple-active-revisions-enabled). 

Right now, azd is promoting an "[image upsert](https://learn.microsoft.com/en-us/azure/developer/azure-developer-cli/container-apps-workflows#container-app-upsert-strategy)" strategy, where a dummy image alongside with desired container app properties is created in `azd provision` the first time, and `azd deploy` is done separately to mimic a smallest possible image update operation. This works great in simple deployment scenarios.

Looking forward to production scenarios, if traffic splitting, multiple active revisions were supposedly supported by `azd deploy`, `azd provision` attempts would end up resetting the state of the container app back to a single revision mode.

This would happen since the source of truth is actually split across the Bicep files, and *possibly* this new [proposed](https://github.com/Azure/azure-dev/issues/5338) YAML definition we are thinking about.

If we are also thinking about production scenarios, being able to perform both image + app configuration (environment variable) updates in a single revision would be quite important.

I have put out #5694 as an inspirational work on what this could look like moving forward.

Issue #4232, opened 2024-08-21
Split container app deployment out from provisioning


weikanglim:
Currently, `azd` deploys container apps (non-Aspire) by updating the container app properties in the main provisioning flow, then patching the container app image during  `azd deploy`. 

The primary pain point is that users do not have the ability to patch the container app in a single revision with the provision + deploy flow (#3116). A secondary pain point is that if the user doesn't store the container app existence state in CI/CD scenarios, the container app image may be reset during CI/CD provisioning. The UX is also slightly more challenging since multiple parts in the application have to be set up and coordinate well together to build up the overall experience.

There are two ways we can consider splitting out container app deployment from provisioning:

1. Container app deployed using Bicep 
2. Container app deployed using YAML

## Container app deployed using Bicep

This is familiar for a person that is using and knowing Bicep. It is also quite common to have multiple Bicep modules that are deployed and managed independently. Bicep allows the user to reference any existing provisioned resources and link it to fill application environment settings.

The main drawback from azd's perspective is that azd interacts with Bicep opaquely. azd doesn't know if the user had provisioned other resources within the Bicep module. azd can only communicate with the Bicep module via input/outputs just like other provisioning modules.

Historically, this deployment mechanism was previously used by azd. It was noted to be slow, however this may have changed recently.

## Container app deployed using YAML

This is familiar for a person that has prior K8s experience. A manifest style approach for an app operator that doesn't need to touch infrastructure. This provides the right amount of separation for the app developer without having to learn IaC.

The expectation would then be that `azd` needs to provide and document templating functions to "tie-in" the container app to the rest of the application. At the very least, I could imagine:

- The context environment variables (especially a managed identity, if being used)
- The container app environment ID
- A way to reference KeyVault secrets
- A way to reference properties (perhaps `$.jsonpath`) of arbitrary ARM resources by resource ID

## Discussion

Should container app deployment be split from provisioning?

Do you prefer deploying container app using Bicep or YAML?

jongio:
I would be hesitant to complicate the ACA deployment scenario with another YAML spec.  

Can you provide an example of:

1. Customer complaints/concerns on this
2. Demo of existing issues
3. Demo of proposed solution

Thanks

wbreza:
Agree on the pain points listed above.  It is a tricky situation since the Bicep (CAE + ACA) today is a full provision+deploy that can contain references to the container images being used by the service. 

[...]

PR #5694, opened 2025-09-09
support bicep containerapp revisions

weikanglim:
This change allows for revisions to be defined and managed centrally in Bicep separately from `azd provision`, i.e. Option 1 as described in #4232. 

The bicep deployment behavior is that environment variables are substituted, output variables are saved. No prompting is allowed.

Opens up scenarios for #5352, i.e. blue green scenarios are now possible and similar to what is [documented](https://learn.microsoft.com/en-us/azure/container-apps/blue-green-deployment?pivots=bicep#create-a-container-app-with-multiple-active-revisions-enabled) currently for container apps.

Demo [changes](https://github.com/Azure-Samples/todo-nodejs-mongo-aca/pull/8/files#diff-834bc11e3cb55730076a7d70767b75f794dc9a37305f27d99926e0fd24363c13) for todo-nodejs-mongo-aca

Fixes: #3116


cli/azd/pkg/project/service_target_containerapp.go
```diff
@@ -22,6 +30,11 @@ type containerAppTarget struct {
 	containerHelper     *ContainerHelper
 	containerAppService containerapps.ContainerAppService
 	resourceManager     ResourceManager
+	armDeployments      *azapi.StandardDeployments
+	console             input.Console
+	commandRunner       exec.CommandRunner
+
+	bicepCli func() (*bicep.Cli, error)
```
weikanglim, line 38: I made this change intentionally as it didn't feel right to have `bicepCli` injected into the constructor, which leads to fetching of Bicep CLI **unconditionally** in container app services.

wbreza (commented):
We need a solution that does not invoke another round of bicep deployments.  This is basically bringing us back to our v1 implementation.  This would not be supported in other providers like Terraform.

vhvb1989 (approved):
LGTM

PR #5748, opened 2025-09-18
Adding functions to the GitHub CLI wrapper to create environments and add variables

### LINKED ISSUE #5729: Create/update the `copilot` environment in Github

richardpark-msft:
Adding in some additional wrapper functions to work with GitHub deployment environments, including creating/deleting environments and getting/updating environment variables.

Fixes #5729


cli/azd/pkg/tools/github/github.go
```diff
@@ -257,6 +265,104 @@ func (cli *ghCli) GitHubActionsExists(ctx context.Context, repoSlug string) (boo
...
+
+	return nil
+}
+
+func (cli *ghCli) GetEnvironmentVariable(ctx context.Context, repoName string, envName string, variableName string) (string, error) {
+	// Doc: https://docs.github.com/en/rest/actions/variables?apiVersion=2022-11-28#get-an-environment-variable
+	runArgs := cli.newRunArgs("api",
+		"-X", "GET",
+		fmt.Sprintf("/repos/%s/environments/%s/variables/%s", repoName, envName, variableName),
+	).WithEnrichError(true)
```
vhvb1989: What is this?

There is also `runArgs.EnrichError = true` in line 298.

cli/azd/pkg/tools/github/github.go
```diff
@@ -257,6 +265,104 @@ func (cli *ghCli) GitHubActionsExists(ctx context.Context, repoSlug string) (boo
...
+	}
+
+	if err := json.Unmarshal([]byte(res.Stdout), &v); err != nil {
+		return "", err
+	}
+
+	return v.Value, nil
+}
+
+func (cli *ghCli) CreateOrUpdateEnvironmentVariable(ctx context.Context, repoName string, envName string, variableName string, variableValue string) (created bool, err error) {
```
vhvb1989: What are you thoughts about using the existing

```go
func (cli *Cli) SetSecret(ctx context.Context, repoSlug string, name string, value string) error
func (cli *Cli) SetVariable(ctx context.Context, repoSlug string, name string, value string) error
```

I am wondering if we could just add an options struct in the signature, like

```go
func (cli *Cli) SetSecret(ctx context.Context, repoSlug string, name string, value string, options setSecretOptions) error
func (cli *Cli) SetVariable(ctx context.Context, repoSlug string, name string, value string, options, setVariableOptions) error
```

Then, we make the options struct to support a string to set an environment?

Also, the `gh cli` supports variables and secrets directly (no need to use the api).  You can do

```sh
gh variable set MY_VARIABLE --body "my-value" --env production
```

So we can use the same methods we have, just by optionally adding the `--env` argument

richardpark-msft: Ah, that makes a lot of sense, especially if we're leveraging the same command. I didn't spot that, but I realized there was a bit of hierarchy there.

cli/azd/pkg/tools/github/github.go
richardpark-msft: I must have started from an older version of the tree. The doc comment on it said it'd cause the error to also include stdout/stderr. I'll remove it, or look for the equivalent (or maybe that's default behavior now).

richardpark-msft: Okay, it's been removed.

cli/azd/pkg/tools/github/github.go
richardpark-msft: I updated SetVariable and ListVariables, but not others as I'm trying to stay within a small scope of what I'm actually testing at the moment.

cli/azd/pkg/tools/github/github_test.go
```diff
@@ -21,11 +22,118 @@ import (
 	"github.com/azure/azure-dev/cli/azd/pkg/input"
 	"github.com/azure/azure-dev/cli/azd/pkg/osutil"
 	"github.com/azure/azure-dev/cli/azd/test/mocks"
+	"github.com/azure/azure-dev/cli/azd/test/mocks/mockexec"
 	"github.com/azure/azure-dev/cli/azd/test/mocks/mockinput"
 	"github.com/blang/semver/v4"
 	"github.com/stretchr/testify/require"
 )
 
+func TestGithubCLIDeploymentEnvironments(t *testing.T) {
```
weikanglim, line 29: Speaking from experience, the tradeoff between increased test coverage vs maintenance may not be worth it here, but I'm curious to learn from your thoughts.

From what I'm currently seeing, the actual code being asserted is that `CreateEnvironmentIfNotExist` does indeed call `gh variable set <var> --env <env>` correctly, which is quickly easy to review and spot from the code changes themselves.

I'm not currently in favor promote the pattern of adding mocked tests for each wrapped cli call as it does seem a little heavyweight.

I could see us adding some simple snaphots/recording tests to exercise a full e2e of a "environment creation and setup" that would also live test against `gh` without interactions mocked.

cli/azd/pkg/tools/github/github.go
```diff
@@ -424,6 +444,46 @@ func (cli *Cli) GitHubActionsExists(ctx context.Context, repoSlug string) (bool,
...
+	)
+
+	res, err := cli.commandRunner.Run(ctx, runArgs)
+	if err != nil {
+		return err
+	}
+
+	if res.ExitCode != 0 {
+		return err
+	}
```
weikanglim: minor: 1. `if res.ExitCode != 0 { return err }` seems unnecessary, and 2. it may be better to call `cli.run()`, notes below.

```suggestion
	res, err := cli.run(ctx, runArgs)
	if err != nil {
		return fmt.Errorf("creating new gh environment: %w", err)
	}
```


Notes:
1. On `if res.ExitCode != 0 { return err }`: a non-nil err is always returned if `ExitCode != 0`. Supposedly, if the err was nil, `if res.ExitCode != 0 { return err }` would end up returning `nil` anyways.
2. Using `cli.run()` seems to be the "right option" -- it wraps the error return for not logged in..

cli/azd/pkg/tools/github/github_test.go
richardpark-msft: I'd much rather just use live testing, especially for things like this. Is that possible with today's testing - I only need a GitHub repo that's okay to test with.

weikanglim: I don't think we have prior art yet.. @vhvb1989 @danieljurek ?

richardpark-msft: For expediency, I'll remove the mock tests, leave the live test in, but keep it skipped so we can hopefully activate it some time in the future.

cli/azd/pkg/tools/github/github.go
richardpark-msft: I'll swap it out - using commandRunner directly wasn't really a choice on my part :)

richardpark-msft: Done. I also wasn't using 'res', so I can just return the error directly-ish.

cli/azd/pkg/tools/github/github_test.go
richardpark-msft: Done!

vhvb1989: Our azd-test-recorder supports tool-recording.
I don't remember if it already knows how to record responses from gh cli, but that could be an option...  recording functional tests only.

For live, you would need to handle creating a new repo, running and deleting the repo at the end :)

I think skipping is just fine for now. I don't think we have any tests for this right now XD

richardpark-msft:
@weikanglim, did another change to fix some long lines, so I'll need another approval.

weikanglim (approved):

PR #5753, opened 2025-09-18
feat: implement auto-installation workflow for extensions

### LINKED ISSUE #5752: Improve `azd` extension experience with a first-run install + run workflow
Currently using an `azd` extension is a multi-step process i.e., find the extension, install it, then run the command. This makes extensions hard to discover and use, which could push us towards bundling them by default which is something we want to avoid.

Instead we can optimize it. If a user runs a command for an uninstalled extension (like `azd ai agents init `), `azd` should ask them if they want to install it. If they say yes, it installs the extension and immediately runs the command.

**Current flow:**
1.  `azd extension list`
2.  `azd extension install microsoft.ai.agents`
3.  `azd ai agents init`

**Proposed new flow:**
1.  `azd ai agents init`

This approach simplifies the user experience keeping installations intentional and avoids a bloating users environment.

vhvb1989:
## Overview

This PR implements auto-installation of extensions when users attempt to run commands that match available extension namespaces but are not currently installed.

## Problem

Users who try to run extension commands without having the extension installed receive unhelpful error messages like:
```
Error: unknown command "demo" for "azd"
Run 'azd --help' for usage.
```

This creates friction in the user experience as users need to manually discover and install extensions.

## Solution

**New Approach: Proactive Extension Detection**

Instead of running the command and capturing errors, the implementation now:

1. **Parses command arguments** before execution to identify the primary command
2. **Checks extension registry** proactively to see if the command matches any available extension namespace
3. **Prompts for auto-installation** if a match is found and the extension isn't already installed
4. **Proceeds with normal execution** after installation or if no match is found

### Key Implementation Details

- **`findFirstNonFlagArg()`**: Extracts the actual command from arguments, ignoring flags
- **`checkForMatchingExtension()`**: Queries the extension registry to find namespace matches
- **`tryAutoInstallExtension()`**: Handles user prompting and extension installation
- **`ExecuteWithAutoInstall()`**: Main orchestration function that integrates with existing command execution

### User Experience Improvements

**Before:**
```
$ azd demo
Error: unknown command "demo" for "azd"
Run 'azd --help' for usage.
```

**After:**
```
$ azd demo
Command 'demo' not found, but there's an available extension that provides it.
Extension: AZD Demo Extension (Demonstrates core azd functionality)
Would you like to install it? (Y/n): y
Installing extension 'microsoft.azd.demo'...
Extension 'microsoft.azd.demo' installed successfully!
[demo command executes normally]
```

## Changes Made

- **`cmd/auto_install.go`**: New file containing auto-installation logic
- **`main.go`**: Modified to use `ExecuteWithAutoInstall` instead of direct command execution
- **Extension Integration**: Leverages existing extension registry and manager infrastructure

## Testing

- ✅ Auto-installation works for extension commands (e.g., `azd demo`, `azd x`)
- ✅ Normal error behavior preserved for non-extension commands
- ✅ User can accept or decline installation
- ✅ Already installed extensions are not prompted for reinstallation
- ✅ Proper fallback handling for console resolution issues

Fixes #5752


cli/azd/cmd/auto_install.go
```diff
@@ -0,0 +1,118 @@
...
+
+	// Check if the unknown command matches any available extension namespace
+	options := &extensions.ListOptions{}
+	registryExtensions, err := extensionManager.ListFromRegistry(ctx, options)
+	if err != nil {
+		// If we can't list registry extensions, we can't auto-install
+		return false, nil
+	}
+
+	var matchingExtension *extensions.ExtensionMetadata
```
wbreza: A couple things to validate:

User runs `azd ai agent init`. The namespace for this extension could be `ai`, `ai.agent` or `ai.agent.config`.  Some agents may have a default entrypoint/command we should likely be recursive - start with the full match that you have and work back to the root to checking for matching extensions.

cli/azd/cmd/auto_install.go
```diff
@@ -0,0 +1,118 @@
...
+	fmt.Printf("Extension: %s (%s)\n", matchingExtension.DisplayName, matchingExtension.Description)
+	fmt.Printf("Would you like to install it? (y/N): ")
+
+	var response string
+	fmt.Scanln(&response)
+	response = strings.ToLower(strings.TrimSpace(response))
+
+	if response != "y" && response != "yes" {
+		return false, nil
+	}
```
wbreza: We should use our standard prompting libraries here.  You likely don't have easy access since you are executing this flow from `main.go`.  Alternatively we could move this into the error middleware and use IoC/DI to get access to our standard components.

cli/azd/cmd/auto_install.go
```diff
@@ -0,0 +1,118 @@
...
+
+	// Install the extension
+	fmt.Printf("Installing extension '%s'...\n", matchingExtension.Id)
+	filterOptions := &extensions.FilterOptions{}
+	_, err = extensionManager.Install(ctx, matchingExtension.Id, filterOptions)
+	if err != nil {
+		return false, fmt.Errorf("failed to install extension: %w", err)
+	}
+
+	fmt.Printf("Extension '%s' installed successfully!\n", matchingExtension.Id)
```
wbreza: At the end of the flow it would be great if we can automatically retry the flow without the user having to execute the command again.  This is another place where a middleware may be helpful.  The new agentic error handling uses a similar concept where it retries `next()` multiple times within the same middleware after applying fixes.

wbreza (approved):
This is a great start! Added some feedback.  Reach out if you have any questions.

cli/azd/cmd/auto_install.go
vhvb1989: https://github.com/Azure/azure-dev/issues/5755

cli/azd/cmd/auto_install.go
vhvb1989: Updated to use the input.Console :)

cli/azd/cmd/auto_install.go
vhvb1989: This is currently happening. Once the installation is completed, the cobraCommand is re-constructed (now including the installed extension commands) and the command is then executed :)

cli/azd/cmd/auto_install_test.go
```diff
@@ -0,0 +1,166 @@
...
+		{
+			name:     "single character flags",
+			args:     []string{"-v", "-h", "up", "--debug"},
+			expected: "up",
+		},
+		{
+			name:     "command with flag value",
+			args:     []string{"--output", "json", "demo", "subcommand"},
+			expected: "json",
+		},
```
JeffreyCA, line 86: Should the expected value be `demo` in this case, since `--output json` belong together? Not sure how hard it would be to handle this, since some flags require additional values to be passed and some don't, like `--help`.

cli/azd/main.go
```diff
@@ -62,7 +62,9 @@ func main() {
 
 	rootContainer := ioc.NewNestedContainer(nil)
 	ioc.RegisterInstance(rootContainer, ctx)
-	cmdErr := cmd.NewRootCmd(false, nil, rootContainer).ExecuteContext(ctx)
+
+	// Execute command with auto-installation support for extensions
+	cmdErr := cmd.ExecuteWithAutoInstall(ctx, rootContainer)
```
JeffreyCA, line 67: How does this feature work with the `alpha.extensions` flag (i.e. when set to `on` vs `off`)?

vhvb1989: This is currently ignoring the alpha flags!!! XD  

I wonder at what point we will remove the alpha flag from extensions..  and if it is worth it to still keep it as alpha

cli/azd/cmd/auto_install.go
```diff
@@ -0,0 +1,281 @@
...
+
+		flagsList := strings.Join(unknownFlags, ", ")
+		errorMsg := fmt.Sprintf(
+			"Unknown flags detected before command '%s': %s\n\n"+
+				"If you're trying to run an extension command, the extension name must come BEFORE any flags.\n"+
+				"This is because extension-specific flags are not known until the extension is installed.\n\n"+
+				"Correct usage:\n"+
+				"  azd %s %s    # Extension name first, then flags\n"+
+				"  azd %s --help          # Get help for the extension\n\n"+
+				"If this is not an extension command, please check the flag names for typos.",
```
JeffreyCA: Is this too much of a behavioural change? If the user runs `azd --debug up`, they'll run into this error message where as before it would've worked normally.

vhvb1989: Updated it.  Nice catch.
It should now move on if the command is detected to be a built-in command :)

cli/azd/cmd/auto_install.go
```diff
@@ -0,0 +1,390 @@
...
+		if flag.Value.Type() != "bool" {
+			flagsWithValues["--"+flag.Name] = true
+			if flag.Shorthand != "" {
+				flagsWithValues["-"+flag.Shorthand] = true
+			}
+		}
+	})
+
+	// Also check persistent flags (global flags)
+	cmd.PersistentFlags().VisitAll(func(flag *pflag.Flag) {
```
weikanglim, line 52: redundant

```golang
// Flags returns the complete FlagSet that applies
// to this command (local and persistent declared here and by all parents).
func (c *Command) Flags() *flag.FlagSet {
```

cli/azd/cmd/auto_install.go
```diff
@@ -0,0 +1,390 @@
...
+		if i+1 < len(args) && i+2 < len(args) {
+			nextArg := args[i+1]
+			argAfterNext := args[i+2]
+			if !strings.HasPrefix(nextArg, "-") && !strings.HasPrefix(argAfterNext, "-") {
+				// Pattern: --unknown value command
+				// Skip the value, let command be found next
+				skipNext = true
+			}
+		}
+	}
```
weikanglim, line 118: I am curious about what pushed us dowwn this path. I would be concerned about the custom flag parsing logic here not matching cobra.. Is this something you may consider as well?

weikanglim: I haven't looked that far, but my gut is telling me that we could simply invoke the command, and check for "unknown command" which is a public-facing documented thing [here](https://github.com/spf13/cobra/blob/7da941c3547e93b8c9f70bbd3befca79c6335388/site/content/user_guide.md?plain=1#L709).

cli/azd/cmd/auto_install.go
```diff
@@ -0,0 +1,390 @@
...
+// extractFlagsWithValues extracts flags that take values from a cobra command.
+// This ensures we have a single source of truth for flag definitions by
+// dynamically inspecting the command's flag definitions instead of
+// maintaining a separate hardcoded list.
+//
+// The function inspects both regular flags and persistent flags, checking
+// the flag's value type to determine if it takes an argument:
+// - Bool flags don't take values
+// - String, Int, StringSlice, etc. flags do take values
+func extractFlagsWithValues(cmd *cobra.Command) map[string]bool {
```
wbreza, line 31: Why do we need custom flag parsing? `azd` flags should be handled by standard `azd` core.  Extension flags should be handled by extension itself...

cli/azd/cmd/auto_install.go
```diff
@@ -0,0 +1,390 @@
...
+	})
+
+	return flagsWithValues
+}
+
+// findFirstNonFlagArg finds the first argument that doesn't start with '-' and isn't a flag value.
+// This function properly handles flags that take values (like --output json) to avoid
+// incorrectly identifying flag values as commands.
+// Returns the command and any unknown flags encountered before the command.
+func findFirstNonFlagArg(args []string, flagsWithValues map[string]bool) (command string, unknownFlags []string) {
```
wbreza, line 68: Still not sure we need any of this??

cli/azd/cmd/auto_install.go
```diff
@@ -0,0 +1,390 @@
...
+	if err != nil {
+		return false, fmt.Errorf("failed to install extension: %w", err)
+	}
+
+	console.Message(ctx, fmt.Sprintf("Extension '%s' installed successfully!\n", extension.Id))
+	return true, nil
+}
+
+// ExecuteWithAutoInstall executes the command and handles auto-installation of extensions for unknown commands.
+func ExecuteWithAutoInstall(ctx context.Context, rootContainer *ioc.NestedContainer) error {
```
wbreza, line 272: I think we are doing to much custom work for flag/arg passing here... I would have expected to see the following workflow.

- Invoke rootCmd
- Check error
- If command not found then
  - Find matching extension
  - Install matching extension
  - Bind extension
- Re-execute root cmd (show pass now)

cli/azd/cmd/auto_install.go
vhvb1989: This is to handle running something like `azd --cwd path demo` where the `demo` is the extension.
We don't want azd to think `--cwd` or `path` is the command we are trying to run XD

cli/azd/cmd/auto_install.go
vhvb1989: Not for the happy path, but this is mostly for starting with flags

cli/azd/cmd/auto_install.go
vhvb1989: Well, this is expected to be different than what Cobra does, and at this point we are looking at flags before the cobra command is executed.

cli/azd/cmd/auto_install.go
vhvb1989: That flow was the initial one... It is good for the happy path `azd extension` but it was not enough as a general solution.
What I didn't like (the most) of trying to run first was that azd would start by printing the error "command not found". It didn't look or feel right.
So it evolved into:

- Create rootCmd
- Check if extensions is enabled - skip and run w/o auto-install if not
- Try to get the command invocation after removing flags - fail if there is ambiguity due to flags ordering
- Check if the command invocation is built-in or known already - skip and run w/o auto-install if so
- Find the extension(s) that matches and install - skip and run w/o auto-install if no matches
- Rebuild rootCmd and execute 🤝

cli/azd/cmd/auto_install.go
vhvb1989: > we could simply invoke the command, and check for "unknown command"

That was my first implementation. But before running auto-install, folks will always see an error like:

```
vivazqu@vib33:~/workspace/build/debug/custom-lang$ azd myExtension
Error: unknown command "myExtension" for "azd"

Did you mean this?
        extension

Run 'azd --help' for usage.
```

That was really annoying to see and forced me to re-think in a better UX where we make it cooler 🎼

weikanglim: Does this work? Override Stdout and Stderr for cobra, buffer it and decide to print it out after the error.

@vhvb1989 you rock for trying that out as the first implementation :)

vhvb1989: > Override Stdout and Stderr for cobra, buffer it and decide to print it out after the error.

That means that if you run `azd up` , you won't see any output until the end of executing cobra. We definitely don't want that 🏃‍♀️‍➡️

weikanglim: Ok, thanks for keeping me honest.. I'll come back to this real quick..

weikanglim: Second quick scan: Does calling `command.Find(args)` work?

```golang
// Find the target command given the args and command tree
// Meant to be run on the highest node. Only searches down.
func (c *Command) Find(args []string) (*Command, []string, error) {
```

From a quick look, this is used in cobra's path for figuring out the command invocation before printing out error and executing.

cli/azd/cmd/auto_install.go
wbreza: Shouldn't we be able to get the error and choose whether it gets sent to the console or not? or is it too late in the process?

cli/azd/cmd/auto_install.go
wbreza: I'd recommend we start with the happy path for now and not handle all the random ways people can call the commands... Most people will run `azd <extension-path> <command> [flags]`

cli/azd/cmd/auto_install.go
vhvb1989: see: https://github.com/Azure/azure-dev/pull/5753#discussion_r2376648125

cli/azd/cmd/auto_install.go
vhvb1989: I understand the desire to simplify, but I’d prefer to keep the feature as-is because it’s already covering edge cases that would otherwise surface as bugs or user friction. Shipping only the happy path might look cleaner short-term, but it risks making the system fragile and harder to trust. Robustness here isn’t just a nice-to-have—it’s what makes the feature reliable in real-world use.

weikanglim (approved):

PR #5770, opened 2025-09-23
update gh cli to latest 2.80

vhvb1989:
fix: https://github.com/Azure/azure-dev/issues/5769


cli/azd/pkg/tools/github/github.go
```diff
@@ -257,19 +257,8 @@ func (cli *Cli) ListSecrets(ctx context.Context, repoSlug string) ([]string, err
...
-//nolint:lll
-func (cli *Cli) ListVariables(ctx context.Context, repoSlug string, options *ListVariablesOptions) (map[string]string, error) {
-	args := []string{"-R", repoSlug, "variable", "list"}
-
-	if options != nil && options.Environment != "" {
-		args = append(args, "--env", options.Environment)
-	}
-
-	runArgs := cli.newRunArgs(args...)
+func (cli *Cli) ListVariables(ctx context.Context, repoSlug string) (map[string]string, error) {
```
wbreza: We are changing the implementation/signature of the function - I would have expected to see some updates to callers of these APIs ??

PR #5771, opened 2025-09-23
Support language custom

vhvb1989:
fix: https://github.com/Azure/azure-dev/issues/5810


heaths:
Because `azd` doesn't support deploying a Functions app using a container. This is for Rust support - or any language `azd` doesn't currently support - where all I need to do is hook `build` and `package` and the existing functionality should otherwise work.

See https://github.com/heaths/functions-quickstart-rust-azd/pull/2

We have an internal discussion in Teams on this. Please see that for more information. Rust and Go are other popular (and growing in popularity) languages, and if `azd` won't support them directly nor support them via extensions (language support is mentioned in your roadmap as merely "planned"), then having a "custom" language would allow other languages to hook their way to building and packaging. That's really all I need here. `azd` already does most of what I need. Just let me tell it how to build and package.

@vhvb1989 unfortunately, this doesn't quite work but may be close. My `prebuild` and `prepackage` hooks are working, but I'm still getting:

> ERROR: failed packaging service 'api': creating deployable zip: open : no such file or director

I would expect that `language: custom` wouldn't try to run any commands itself, but would still respect build -> package -> deploy (or similar). I had a previous iteration that everything was done in `prepackage` and even `predepoy`. Like this, `azd` also doesn't seem to build prior to package. I ran with `--debug` and don't see any signs of a build step.

heaths:
@vhvb1989 I updated this PR with something closer to what I was thinking. It'd be great if we supported all hooks for a `custom` language, but off hand I wasn't sure about the ramifications of such a change so I figured this suffices as an MVP. I'm working on my PR linked above. It's mostly working, but I needed to make a few changes to the consumption plan.

cli/azd/internal/cmd/show/show.go
```diff
@@ -539,6 +539,8 @@ func showTypeFromLanguage(language project.ServiceLanguageKind) contracts.ShowTy
 		return contracts.ShowTypeNode
 	case project.ServiceLanguageJava:
 		return contracts.ShowTypeJava
+	case project.ServiceLanguageCustom:
+		return contracts.ShowTypeCustom
```
tg-msft, line 543: Can we (cheaply) wrap this in a feature flag?  I think we might want to revisit this in a couple of months when we're further a long with our extension story.

heaths: IMO that would be fine, but I think it comes down to whether ya'll want to take this PR at all, right? And if so, I would like to discuss the exact details a little. I did a minimum change to @vhvb1989's original code but I think slightly more along those same lines might warrant discussion. If we will get custom language support via extensions, I certainly don't want to over-engineer this now.

wbreza: I'd recommend we hold off on this till we enable Language support via our Extension framework.

heaths: When can we expect that? Azure SDK for Rust is GA'ing soon and I'd like to have samples ready for customers.

vhvb1989: feature flag added

rajeshkamal5050 (approved):

rajeshkamal5050 (approved):

Issue #5767, opened 2025-09-23
[EPIC] Extension enhancements


tg-msft:
Creating a tracking issue for extension enhancements we discussed during 9/23/05's AZD Engineering Chat meeting.

richardpark-msft:
@tg-msft, one thing about the 'experimental' feed is whether users should be able to see those by default, with stable feeds overriding the experimental options, if they exist.

Otherwise I think uptake on experimental commands will be low - nobody will even know they exist (similar to SDK betas). So the experimental feed would be installed, by default, but come with some extra disclaimers when the command runs, etc..

PR #5773, opened 2025-09-23
add guiding principles for extending azd

kristenwomack:
Adding a new guiding principles outline for adding azd commands, arguments, flags, etc. This is especially important guidance for contribution and the azd extension framework.

cc: @SophCarp @ccbarragan  - extending from the figma file to a descriptive format

### DIFF
--- a/cli/azd/docs/guiding-principles.md
+++ b/cli/azd/docs/guiding-principles.md
@@ -1,0 +1,109 @@
+# Command Structure Design Principles for `azd`
+
+## Overview
+
+The Azure Developer CLI (`azd`) follows a systematic approach to command structure design that ensures discoverability, consistency, and extensibility across all features and future enhancements. These principles provide a foundation for building any new functionality while maintaining user familiarity and operational efficiency.
+
+## Core Design Principles
+
+### 1. **Verb-First Command Structure**
+
+- Keep verb selection minimal to avoid user confusion: `up`, `add`, `deploy`
+- Use simple, common words that users already understand
+- Each command group should deal with one type of object to maintain clarity
+
+### 2. **Build on Existing Foundations**
+
+- Leverage established `azd verbs` rather than creating parallel command structures
+- Extend existing workflows to accommodate new functionality
+- Maintain backward compatibility with current usage patterns
+- Avoid verb duplication; build on core verbs instead of creating new ones
+
+### 3. **Progressive Disclosure**
+
+- Basic commands work simply, advanced features are available when needed
+- Users should naturally discover advanced capabilities within familiar command flows
+- Maintain graceful degradation when advanced features are unavailable
+
+### 4. **Consistency Across All Operations**
+
+- Maintain consistent parameter patterns across similar commands
+- Use established naming conventions and flag structures
+- Ensure similar operations follow similar command patterns
+
+## Command Structure Framework
+
+### Primary Command Categories
+
+#### **View** (Information & Status)
+
+- Purpose: Display current state, configuration, and system information
+- Examples: `list`, `show`, `version`
+- Principle: Read-only operations that help users understand current state
+
+#### **Edit** (Configuration & Management)
+
+- Purpose: Modify configuration, manage resources, and setup operations
+- Examples: `set`, `reset`, `unset`, `config`, `refresh`, `select`, `add`, `remove`, `install`, `uninstall`, `new`
+- Principle: Operations that change system state or configuration
+
+#### **Run** (Actions & Operations)
+
+- Purpose: Execute workflows, deploy resources, and perform active operations
+- Examples: `up`, `down`, `init`, `package`, `provision`, `deploy`, `run`
+- Principle: Operations that perform work or execute processes
+
+#### **Other** (Specialized Operations)
+
+- Purpose: Specialized tasks that don't fit standard categories
+- Examples: `restore`, `monitor`, `generate`, `login`, `logout`
+- Principle: Unique operations specific to particular workflows
+
+## Design Guidelines for Extensions
+
+### 1. **Command Integration Strategy**
+
+- New functionality should extend existing command categories
+- Use subcommands and parameters rather than new top-level verbs
+- Example: `azd add <new-resource-type>` instead of `azd <new-resource-type> add`
+
+### 2. **Parameter and Flag Consistency**
+
+- Reuse established parameter patterns across new commands
+- Maintain consistent naming conventions (e.g., `--subscription`, `--name`, `--type`)
+- Provide sensible defaults to reduce cognitive load
+
+### 3. **Help and Discoverability**
+
+- Integrate new functionality into existing `azd help` structure
+- Provide contextual guidance within established command flows
+- Maintain documentation consistency across core and extended features
+
+### 4. **Template and Resource Integration**
+
+- Leverage existing template system for new resource types
+- Follow established template discovery and management patterns
+- Integrate new resources into azd's resource lifecycle management
+
+## Implementation Benefits
+
+- **User Familiarity**: Builds on known command patterns and reduces learning curve
+- **Discoverability**: New capabilities are found through existing workflows
+- **Consistency**: Predictable behavior across all azd commands and extensions
+- **Maintainability**: Systematic approach reduces complexity and technical debt
+- **Extensibility**: Clear framework for adding capabilities without breaking existing patterns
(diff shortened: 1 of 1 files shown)


cli/azd/docs/guiding-principles.md
```diff
@@ -0,0 +1,109 @@
...
+
+### 1. **Verb-First Command Structure**
+
+- Keep verb selection minimal to avoid user confusion: `up`, `add`, `deploy`
+- Use simple, common words that users already understand
+- Each command group should deal with one type of object to maintain clarity
+
+### 2. **Build on Existing Foundations**
+
+- Leverage established `azd verbs` rather than creating parallel command structures
```
JeffreyCA: Could we include an example of what's preferred and what's not? It wasn't clear to me what "parallel command structures" refers to

cli/azd/docs/guiding-principles.md
```diff
@@ -0,0 +1,109 @@
...
+- Purpose: Specialized tasks that don't fit standard categories
+- Examples: `restore`, `monitor`, `generate`, `login`, `logout`
+- Principle: Unique operations specific to particular workflows
+
+## Design Guidelines for Extensions
+
+### 1. **Command Integration Strategy**
+
+- New functionality should extend existing command categories
+- Use subcommands and parameters rather than new top-level verbs
```
JeffreyCA:
```suggestion
- Use a verb-first structure, where the primary action (e.g., add, create, delete, list) is the top-level command, and the target entity or context follows as an argument or subcommand.
```

This made a bit more sense to me with the given example

cli/azd/docs/guiding-principles.md
```diff
@@ -0,0 +1,109 @@
...
+- Keep verb selection minimal to avoid user confusion: `up`, `add`, `deploy`
+- Use simple, common words that users already understand
+- Each command group should deal with one type of object to maintain clarity
+
+### 2. **Build on Existing Foundations**
+
+- Leverage established `azd verbs` rather than creating parallel command structures
+- Extend existing workflows to accommodate new functionality
+- Maintain backward compatibility with current usage patterns
+- Avoid verb duplication; build on core verbs instead of creating new ones
```
JeffreyCA: How about:

```suggestion
- Avoid ambiguous or similarly-named verbs; build on core verbs instead of creating new ones
```

I got this from [clig.dev#subcommands:](https://clig.dev/#subcommands:)

> Don’t have ambiguous or similarly-named commands. For example, having two subcommands called “update” and “upgrade” is quite confusing. You might want to use different words, or disambiguate with extra words.

Some examples could be:

- update and upgrade
- show and view

cli/azd/docs/guiding-principles.md
```diff
@@ -0,0 +1,109 @@
+# Command Structure Design Principles for `azd`
+
+## Overview
```
tg-msft, line 3: Nit - it'd be great to link this from somewhere in https://github.com/Azure/azure-dev/blob/main/cli/azd/CONTRIBUTING.md for good measure

cli/azd/docs/guiding-principles.md
```diff
@@ -0,0 +1,109 @@
...
+### 1. **Command Integration Strategy**
+
+- New functionality should extend existing command categories
+- Use a verb-first structure, where the primary action (e.g., add, create, delete, list) is the top-level command, and the target entity or context follows as an argument or subcommand.
+- Example: `azd add <new-resource-type>` instead of `azd <new-resource-type> add`
+
+### 2. **Parameter and Flag Consistency**
+
+- Reuse established parameter patterns across new commands
+- Maintain consistent naming conventions (e.g., `--subscription`, `--name`, `--type`)
```
tg-msft, line 73: Future idea - it'd be great to codify things like this with linting rules that yell about easy mistakes like `--subscriptionId` instead of `--subscription`.  Do folks think that'd make more sense in our repo's linting or something shipped separately so people could use it from any repo?  I filed https://github.com/Azure/azure-dev/issues/5826 to track.

cli/azd/docs/guiding-principles.md
```diff
@@ -0,0 +1,109 @@
...
+- Provide sensible defaults to reduce cognitive load
+
+### 3. **Help and Discoverability**
+
+- Integrate new functionality into existing `azd help` structure
+- Provide contextual guidance within established command flows
+- Maintain documentation consistency across core and extended features
```
tg-msft, line 80: Nit - how are we thinking about docs for extensions?  Might be worth a separate conversation in the near future.

cli/azd/docs/guiding-principles.md
```diff
@@ -0,0 +1,109 @@
...
+
+## Future Considerations
+
+This framework enables:
+
+- Advanced workflow automation
+- Enhanced developer productivity features
+- Consistent user experience across all `azd` functionality
+- Integration of new Azure services and capabilities
+- Third-party extension development
```
tg-msft, line 110: I think in a month or two we should update these to add a section on "Design as an Extension first" to push teams to use the extension framework rather than trying to build in azd core.  I'd like to make a little more progress knocking down sharp edges first though.  Added to https://github.com/Azure/azure-dev/issues/5767 to track.

kristenwomack:
> Can we work in some guiding principles for CICD and infra tooling support as well? e.g.,
> 
> * CICD related features: must have for both GH Action and azdo.
> * Infra: Bicep first. Terraform can be later but must be considered.

I just added this, great suggestion. On the IaC, I did leave it more open-ended for scenarios where an IaC provider might have an extension that focuses on their IaC flavor.

cli/azd/docs/guiding-principles.md
kristenwomack: excellent call out, ty for logging the issue

cli/azd/docs/guiding-principles.md
kristenwomack: Yes, extensions on the public registry must have documentation. I'll start drafting those guidelines.

cli/azd/docs/guiding-principles.md
kristenwomack: Excellent, yes! ty for logging this

cli/azd/docs/guiding-principles.md
kristenwomack: Yes, 100% that is on my list!

cli/azd/docs/guiding-principles.md
```diff
@@ -0,0 +1,114 @@
...
+
+- Purpose: Specialized tasks that don't fit standard categories
+- Examples: `restore`, `monitor`, `generate`, `login`, `logout`
+- Principle: Unique operations specific to particular workflows
+
+## Design Guidelines for Extensions
+
+### 1. **Command Integration Strategy**
+
+- New functionality should extend existing command categories
```
rajeshkamal5050, line 66: This one could be tricky in general.

- If users are adding functionality/extensions which is overriding the core commands then this would work i.e., I want to extend more resources for `azd add` or more deployment targets `azd deploy` or more provisioning provider types ex: pulumi etc for `azd add`
- If users are adding functionality/extensions which is net new and has no verbs/commands to extend or leverage from core. Then we might have to use a azd <resource/noun> <verb> semantics. Since such verbs can overlap and might need a separation above via resource/noun.

Non-blocking comment for now :-)

@tg-msft @wbreza thoughts?

rajeshkamal5050 (approved):

rajeshkamal5050:
```
## Design Guidelines for Extensions

### 1. **Command Integration Strategy**
- New functionality should extend existing command categories
- Use a verb-first structure, where the primary action (e.g., add, create, delete, list) is the top-level command, and the target entity or context follows as an argument or subcommand.
- Example: `azd add <new-resource-type>` instead of `azd <new-resource-type> add`
```
This one might be tricky I guess, Non-blocking comment for now :-)

- If users are adding functionality/extensions which is overriding the core commands then this would work i.e., I want to extend more resources for `azd add` or more deployment targets `azd deploy` or more provisioning provider types ex: pulumi etc for `azd add`
- If users are adding functionality/extensions which is net new and has no verbs/commands to extend or leverage from core. Then we might have to use a `azd resource/noun verb` semantics. Since such verbs can overlap and might need a separation above via resource/noun.

@tg-msft @wbreza thoughts?

rajeshkamal5050 (approved):

PR #5777, opened 2025-09-24
Removes feature flag block for MCP server

wbreza:
Removes the `llm` feature flag from blocking the usage of the `mcp` command group.

### DIFF
--- a/cli/azd/cmd/root.go
+++ b/cli/azd/cmd/root.go
@@ -19,7 +19,6 @@ import (
 	"github.com/azure/azure-dev/cli/azd/pkg/azd"
 	"github.com/azure/azure-dev/cli/azd/pkg/extensions"
 	"github.com/azure/azure-dev/cli/azd/pkg/ioc"
-	"github.com/azure/azure-dev/cli/azd/pkg/llm"
 	"github.com/azure/azure-dev/cli/azd/pkg/platform"
 
 	"github.com/azure/azure-dev/cli/azd/internal"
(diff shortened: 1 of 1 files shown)


cli/azd/cmd/testdata/TestUsage-azd.snap
```diff
@@ -33,6 +33,13 @@ Commands
     restore   	: Restores the project's dependencies.
     template  	: Find and view template details.
 
+  Enabled alpha commands
+    extension 	: Manage azd extensions.
+    mcp       	: Manage Model Context Protocol (MCP) server.
+
+  Enabled extensions commands 
+    x         	: This extension provides a set of tools for AZD extension developers to test and debug their extensions.
+
```
JeffreyCA: Since `mcp` isn't tied to an alpha feature, would it make more sense to have it under a different group in the help text?

Also we may need to remove `extension` and `x` since they're not installed when the pipeline runs the tests.

cli/azd/cmd/testdata/TestUsage-azd-mcp.snap
```diff
@@ -1,8 +1,12 @@
 
-
+Manage Model Context Protocol (MCP) server.
```
JeffreyCA: If `azd mcp` commands are alpha should we include "(Alpha)" in the descriptions? I believe we did something similar for `azd add` when composability was in alpha.

cli/azd/cmd/testdata/TestUsage-azd.snap
wbreza: Lets just keep it under "alpha" or not but not hide it under a flag.

PR #5804, opened 2025-09-26
Remove cgo requirement for osversion on Darwin

parkplusplus:
If you're using 'azd' as a library (for instance, writing an extension in Go), obversion will require you to have CGO_ENABLED _and_ you have to have the proper cross compilers for it to work. 

This call appears to only be used once (when initializing telemetry) and it's fast so I think this should still fit in well with everything else in azd.


parkplusplus:

Issue #5793, opened 2025-09-25
UX: add summary at the end for azd init


hemarina:
<img width="1095" height="528" alt="Image" src="https://github.com/user-attachments/assets/43d80d3b-6028-4cb0-a5b3-67f2baef5f9a" />

Issue #5792, opened 2025-09-25
UX: update last feedback


hemarina:
<img width="1058" height="88" alt="Image" src="https://github.com/user-attachments/assets/ff41737e-0d14-4b37-b05a-e86ea699c83d" />

Issue #5791, opened 2025-09-25
UX: create a function for azd init and error handling workflow for feedback


hemarina:
Create feedback function that could leverage by azd init and error handling workflow.
Update UX for feedback:
<img width="769" height="97" alt="Image" src="https://github.com/user-attachments/assets/5d3b737e-ab04-4590-81ef-3b53dea4c076" />

Issue #5790, opened 2025-09-25
UX: enable file watch for azd init


hemarina:
<img width="1123" height="373" alt="Image" src="https://github.com/user-attachments/assets/5cb42593-fe38-443a-94d0-a27b8c53dd67" />

PR #5808, opened 2025-09-27
improve ux for agentic azd init

### LINKED ISSUE #5775: UX: azd init agent initial selection
Current UX: 
Image

Expected UX:
Image

hemarina:
fix #5775, #5783, #5784, #5785, #5790, #5791, #5792, #5793, #5789, #5788, #5774

image
image

...

image

### DIFF
--- a/cli/azd/cmd/init.go
+++ b/cli/azd/cmd/init.go
@@ -262,9 +262,11 @@ func (i *initAction) Run(ctx context.Context) (*actions.ActionResult, error) {
 	header := "New project initialized!"
 	followUp := heredoc.Docf(`
 	You can view the template code in your directory: %s
-	Learn more about running 3rd party code on our DevHub: %s`,
+	Learn more about running 3rd party code on our DevHub: %s
+	%s Run azd up to deploy project to the cloud.`,
 		output.WithLinkFormat("%s", wd),
-		output.WithLinkFormat("%s", "https://aka.ms/azd-third-party-code-notice"))
+		output.WithLinkFormat("%s", "https://aka.ms/azd-third-party-code-notice"),
+		color.HiMagentaString("Next steps:"))
 
 	switch initTypeSelect {
 	case initAppTemplate:
@@ -404,20 +406,16 @@ func (i *initAction) initAppWithAgent(ctx context.Context) error {
 	}
 
 	if !readOnlyRule.Allowed {
-		if err := i.console.DoInteraction(func() error {
-			consentChecker := consent.NewConsentChecker(i.consentManager, "")
-			promptErr := consentChecker.PromptAndGrantReadOnlyToolConsent(ctx)
-			i.console.Message(ctx, "")
-
-			return promptErr
-		}); err != nil {
+		consentChecker := consent.NewConsentChecker(i.consentManager, "")
+		err = consentChecker.PromptAndGrantReadOnlyToolConsent(ctx)
+		if err != nil {
 			return err
 		}
+		i.console.Message(ctx, "")
 	}
 
 	azdAgent, err := i.agentFactory.Create(
 		agent.WithDebug(i.flags.global.EnableDebugLogging),
-		agent.WithFileWatching(true),
 	)
 	if err != nil {
 		return err
@@ -426,8 +424,9 @@ func (i *initAction) initAppWithAgent(ctx context.Context) error {
 	defer azdAgent.Stop()
 
 	type initStep struct {
-		Name        string
-		Description string
+		Name         string
+		Description  string
+		SummaryTitle string
 	}
 
 	taskInput := `Your task: %s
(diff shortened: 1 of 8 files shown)


cli/azd/cmd/middleware/error.go
```diff
@@ -217,41 +217,15 @@ func (e *ErrorMiddleware) collectAndApplyFeedback(
...
 	AIDisclaimer string,
 ) error {
-	userInputPrompt := uxlib.NewPrompt(&uxlib.PromptOptions{
-		Message:  "Any changes you'd like to make?",
-		Hint:     "Describe your changes or press enter to skip.",
-		Required: false,
+	collector := feedback.NewFeedbackCollector(e.console, feedback.FeedbackCollectorOptions{
+		EnableLoop:      false,
+		FeedbackPrompt:  "Any changes you'd like to make?",
+		FeedbackHint:    "Describe your changes or press enter to skip.",
```
kristenwomack, line 223: We still need to refine this section, but can do that after this PR. Not a blocker.

kristenwomack (approved):

cli/azd/cmd/init.go
```diff
@@ -374,10 +380,44 @@ func (i *initAction) Run(ctx context.Context) (*actions.ActionResult, error) {
...
+	}
+
+	if !readOnlyRule.Allowed {
+		if err := i.console.DoInteraction(func() error {
+			consentChecker := consent.NewConsentChecker(i.consentManager, "")
+			promptErr := consentChecker.PromptAndGrantReadOnlyToolConsent(ctx)
+			i.console.Message(ctx, "")
+
+			return promptErr
+		}); err != nil {
```
wbreza: Do we need the `DoInteraction` block here?

cli/azd/cmd/init.go
```diff
@@ -374,10 +380,44 @@ func (i *initAction) Run(ctx context.Context) (*actions.ActionResult, error) {
...
+			return err
+		}
+	}
 
 	azdAgent, err := i.agentFactory.Create(
 		agent.WithDebug(i.flags.global.EnableDebugLogging),
+		agent.WithFileWatching(true),
```
wbreza: Do we have a case where we don't want the file watching behavior? If not lets make it the default so that each caller doesn't need to set it.  Only need to set it when we want to turn it off.

cli/azd/internal/agent/consent/checker.go
```diff
@@ -190,66 +208,110 @@ func (cc *ConsentChecker) promptForToolConsent(
...
+	// if isReadOnlyTool {
+	// 	choices = append(choices, &ux.SelectChoice{
+	// 		Value: "readonly_server",
+	// 		Label: "Allow all read-only tools from this server",
+	// 	})
+
+	// choices = append(choices, &ux.SelectChoice{
+	// 	Value: "global",
+	// 	Label: "Allow all tools from any server",
+	// })
```
wbreza: I would delete all the comments... we can always get old code from git history.

cli/azd/pkg/ux/prompt.go
```diff
@@ -44,6 +44,8 @@ type PromptOptions struct {
 	ClearOnCompletion bool
 	// Whether or not to capture hint keys (default: true)
 	IgnoreHintKeys bool
+	// The optional second line message to display (default: "")
+	SecondLineMessage string
```
wbreza: I feel like we should be more descriptive about what the intent of this field is... it is basically help text, right? 
We should have a single way of representing help text... we either show it always or on "?" hint key press.

cli/azd/internal/agent/conversational_agent.go
```diff
@@ -173,12 +203,17 @@ func (aai *ConversationalAzdAiAgent) renderThoughts(ctx context.Context) (func()
...
+				elapsedSeconds := int(time.Since(toolStartTime).Seconds())
+
+				spinnerText = fmt.Sprintf("Running %s tool", color.MagentaString(latestAction))
 				if latestActionInput != "" {
-					spinnerText += " with " + color.BlueString(latestActionInput)
+					spinnerText += " with " + color.HiBlackString(latestActionInput)
 				}
 
 				spinnerText += "..."
+				spinnerText += color.HiBlackString(fmt.Sprintf("\n(%ds, esc exit agentic mode)", elapsedSeconds))
```
wbreza: I don't think we implemented "esc to exit" yet... if not we should remove this.

cli/azd/internal/agent/conversational_agent.go
```diff
@@ -124,8 +127,9 @@ func (aai *ConversationalAzdAiAgent) SendMessage(ctx context.Context, args ...st
 	return output, nil
 }
 
-func (aai *ConversationalAzdAiAgent) renderThoughts(ctx context.Context) (func(), error) {
+func (aai *ConversationalAzdAiAgent) renderThoughts(ctx context.Context) (func(), chan struct{}, error) {
 	var latestThought string
+	completionChan := make(chan struct{})
```
wbreza: I don't think we should need another channel and have to manage closing this outside of the function.  In Go the channel should be closed by the creator. In this case the channel is created in the function and expected to be close outside.

In our case we have the incoming context that is "cancelled" when we are done.  We likely can use this to handle the last action printing vs needing another special channel.

cli/azd/cmd/init.go
```diff
@@ -424,14 +464,32 @@ Do not stop until all tasks are complete and fully resolved.
...
 	}
 
+	stepsSummaryLabels := []string{
+		"Step 1 (discovery & analysis)",
+		"Step 2 (architecture plan)",
+		"Step 3 (dockerfile generation)",
+		"Step 4 (infrastructure generation)",
+		"Step 5 (azure.yaml generation)",
+		"Step 6 (project validation)",
+	}
```
wbreza: How are these labels different from the `initStep` names above? Are we using both? If we need them both lets consolidate them into that same object.

wbreza (commented):
Looks great! Added some feedback that I'd like to see addressed before we merge.

wbreza:
image 

> Looks like a huge wall of white text and hard to read.

cli/azd/internal/agent/conversational_agent.go
hemarina: Chatted with Pearce. We will update it to Ctrl+C to exit.

cli/azd/pkg/ux/prompt.go
hemarina: I updated the variable to be more descriptive, but we have two kinds of help message. 
@SophCarp Do we want to choose a way for help message between user type "?" or show the message on second line?

cli/azd/internal/agent/conversational_agent.go
hemarina: Due to the UX order of "Ran..." message, I'll need to cancelCtx() first and then run cleanup(). I'm a little concerned if this order will cause any potential problems like 
- Race conditions between goroutine and main thread accessing canvas
- Possible double-clearing of canvas resources
- Undefined behavior if canvas operations aren't thread-safe

So, I have kept the waiting `time.Sleep(100 * time.Millisecond)` and that should prevent these potential problems. May I double check with you to get your insights?

hemarina:
> image
> 
> > Looks like a huge wall of white text and hard to read.

I have adjust the prompt to list around 6 bulletpoints so that it is not very long and hard to read.

rajeshkamal5050 (approved):

PR #5809, opened 2025-09-29
fix: Make hook shell optional with OS-based defaults (#3613)

### LINKED ISSUE #3613: [Issue/enhancement] Cross-platform shell commands behavior

**Output from `azd version`**
Run `azd version` and copy and paste the output here: azd version 1.7.0 (commit 49d6adc2efb178083f61822e6b4715258560803d)

**Describe the bug**
When defining scripts, we currently use yaml like this:

```yaml
hooks:
  predeploy:
    windows:
      shell: pwsh
      run: Export-ModuleMember -Variable API_URL && npm run build
    posix:
      shell: sh
      run: export API_URL && npm run build
```

This separate the commands to run into two categories: execution for posix-systems (Linux/Mac), and Windows.

The issue is that the distinction is not clear at all, and even misleading:
- If I'm on Windows and using WSL, which version will run?
- If I'm on Windows and using Git Bash, which version will run?
- `sh` will actually run `bash` and not `sh`, which is misleading (bash has a different feature set and by default is not POSIX-compliant)
- If I'm on Windows without Powershell install, how can I use the default cmd.exe? (for simpler commands)
- If I'm on Mac or Linux but with Powershell installed, which version will run?

These issues could be cleared by using a simpler script definition based on the shell name, similar to how it works in GitHub Actions:

```yaml
hooks:
  predeploy:
    pwsh:
      run: Export-ModuleMember -Variable API_URL && npm run build
    bash:
      run: export API_URL && npm run build
```

Instead of discriminating the system (which makes no sense when using WSL or Git Bash), we could just specify which shell to use. The logic could be:

1. Try to run the first specified shell in the list (in this example, pwsh). Test could be simply running `<shell> --version` and see if it fails.
2. If the shell is available run the script, otherwise continue testing with the next shell option.

You could even make the shell specification optional, and use whatever default shell is available on the system for simple commands, for example:

```yaml
hooks:
  predeploy:
    run: npm run build
```

It would even be possible to support the new syntax and keep the old deprecated syntax for backward compatibility with existing azure.yaml configurations.

What do you think about this proposal?

vhvb1989:
## Overview

This PR addresses issue #3613 by making the shell configuration optional for hooks and implementing automatic OS-based shell detection.

### Core Functionality
- **Auto-detect shell based on OS**: When no shell is specified, automatically use `pwsh` on Windows and `sh` on Linux/macOS
- **Comprehensive warning system**: Display user-friendly warnings when default shells are used
- **Backward compatibility**: Existing hook configurations with explicit shell settings continue to work unchanged

### Key Files Modified
- `pkg/ext/models.go`: Added OS-based shell detection and default shell tracking
- `pkg/ext/hooks_manager.go`: Enhanced validation to detect and warn about default shell usage
- `cmd/hooks.go`: Added validation and warning display to `azd hooks run` command
- Updated tests to reflect new behavior where missing shell is no longer an error

### Warning Examples

**Before (Error):**
```
ERROR: unable to determine script type. Ensure 'Shell' parameter is set in configuration options
```

**After (Working with Warning):**
```
WARNING: Hook configurations found without explicit shell specification. Using OS default shell 'sh'. 
For better reliability, consider specifying the shell explicitly in your hook configuration.
Add 'shell: sh' to your hook configurations to remove this warning.
```

### Example Usage

```yaml
# This now works (was previously an error)
hooks:
  prebuild:
    - run: "echo 'Building...'"  # Uses pwsh on Windows, sh on Linux/macOS
      
# Explicit shell (recommended, no warning)
hooks:
  prebuild:
    - shell: sh
      run: "echo 'Building...'"
```

## Testing

- All existing tests pass
- New tests added for default shell behavior and warning system
- Integration tested with `azd hooks run` command
- No regressions in middleware or core hook functionality

## Fixes

Closes #3613

## Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [x] Improvement (non-breaking change which improves existing functionality)

### DIFF
--- a/cli/azd/pkg/ext/hooks_manager.go
+++ b/cli/azd/pkg/ext/hooks_manager.go
@@ -10,6 +10,7 @@ import (
 	"log"
 	"os"
 	osexec "os/exec"
+	"path/filepath"
 	"runtime"
 	"strings"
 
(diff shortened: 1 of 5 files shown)


kristenwomack:
This is fantastic, @vhvb1989 -- thank you! What do you think about adding a short aka link to the error message? [aka.ms/azd-hooks ](https://aka.ms/azd-hooks)

kristenwomack (approved):
LGTM

cli/azd/pkg/ext/hooks_manager.go
```diff
@@ -142,17 +143,51 @@ func (h *HooksManager) ValidateHooks(ctx context.Context, allHooks map[string][]
...
+					} else {
+						hookConfig.Shell = ShellTypeBash
+					}
+					hookConfig.usingDefaultShell = true
+				}
+			}
+		}
+	}
+
+	// Now check all hooks for warning conditions
```
wbreza: We're doing 2 full passes on all the hooks now - could we just perform this same validation in the same pass?

vhvb1989: yeah, I'll add a note about why we need a 2-step validation.
Hopefully not that bad b/c the number of hooks it likely never that big to hit performance

vhvb1989 (commented):
Updated

sinedied:
Thanks, this will simplify many of our config files!

PR #5812, opened 2025-09-29
fix: Aspire owning compute env can't handle AddBicepModule api

vhvb1989:
Updating the hardcoded string `resources.` to handle when the AppHost provides the compute environment. In this case, the name of the resource providing the compute env should used.


davidfowl:
Tested on customer machine, works well!

PR #5827, opened 2025-10-01
Bypass agentic error handling flow in non-interactive scenarios

wbreza:
## Overview

Enhances the error middleware and MCP command executor to properly handle non-interactive scenarios by preventing AI-powered error handling when user interaction is not possible or desired.

We want to avoid `azd` getting stuck in an agentic error handling loop when it may choose to call another `azd` command from within itself especially during an error handling flow.

### Enhanced Error Middleware Short-Circuit Logic
- Added CI detection capability to bypass AI error handling in automated environments
- Extended short-circuit conditions to include:
  - LLM feature disabled
  - `--no-prompt` flag specified
  - Running in CI/CD environment
- Prevents AI error handling from attempting user interaction in automated pipelines

### Simplified MCP Command Executor for `azd` Commands
- Removed overly restrictive command blocking for `up`, `provision`, `deploy`, `down` commands
- Automatically appends `--no-prompt` flag to `azd` commands when not already present
- Allows AI agents to execute commands while ensuring non-interactive mode

## Benefits

- **CI/CD Safety**: Prevents hanging in automated environments where user interaction is impossible
- **Performance**: Eliminates unnecessary AI processing in non-interactive scenarios
- **Better UX**: AI error handling only runs when actually useful (interactive sessions)
- **Reliability**: Uses existing, well-tested CI detection supporting 15+ platforms

### DIFF
--- a/cli/azd/internal/agent/tools/dev/command_executor.go
+++ b/cli/azd/internal/agent/tools/dev/command_executor.go
@@ -121,18 +121,17 @@ func (t CommandExecutorTool) Call(ctx context.Context, input string) (string, er
 	}
 
 	if req.Command == "azd" {
-		blockedCommands := []string{"up", "provision", "deploy", "down"}
-
-		for _, blocked := range blockedCommands {
-			if req.Args[0] == blocked && (len(req.Args) < 2 || req.Args[1] != "--preview") {
-				errorResponse := common.ErrorResponse{
-					Error:   true,
-					Message: "azd command is not supported",
-				}
-
-				jsonData, _ := json.MarshalIndent(errorResponse, "", "  ")
-				return string(jsonData), nil
+		// Ensure --no-prompt is included in args
+		hasNoPrompt := false
+		for _, arg := range req.Args {
+			if arg == "--no-prompt" {
+				hasNoPrompt = true
+				break
 			}
+		}
+
+		if !hasNoPrompt {
+			req.Args = append(req.Args, "--no-prompt")
 		}
 	}
 
--- a/cli/azd/internal/agent/tools/dev/command_executor.go
+++ b/cli/azd/internal/agent/tools/dev/command_executor.go
@@ -11,6 +11,7 @@ import (
 	"os"
 	"os/exec"
 	"runtime"
+	"slices"
 	"strings"
 
 	"github.com/azure/azure-dev/cli/azd/internal/agent/tools/common"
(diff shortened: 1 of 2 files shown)


cli/azd/internal/agent/tools/dev/command_executor.go
```diff
@@ -121,19 +121,18 @@ func (t CommandExecutorTool) Call(ctx context.Context, input string) (string, er
...
+		for _, arg := range req.Args {
+			if arg == "--no-prompt" {
+				hasNoPrompt = true
+				break
 			}
 		}
+
+		if !hasNoPrompt {
+			req.Args = append(req.Args, "--no-prompt")
+		}
```
vhvb1989: Can you help me understand this?  Are we forcing `--no-prompt` ?

JeffreyCA: And could we also simplify this to `if !slices.Contains(req.Args, "--no-prompt")`?

cli/azd/internal/agent/tools/dev/command_executor.go
JeffreyCA, line 1: Could you clarify why we're removing the blocked command logic?

cli/azd/internal/agent/tools/dev/command_executor.go
wbreza: Yes, if our own agent wants to call back into `azd` we are forcing it to always pass it to run with `--no-prompt` so it doesn't get stuck in a prompt flow that it cannot answer.

cli/azd/internal/agent/tools/dev/command_executor.go
wbreza: It was a naive solution to the same issue... we don't want to get stuck in an agentic loop where `azd` calls itself, errors for some reason, then starts a nested agent error resolution flow.

cli/azd/internal/agent/tools/dev/command_executor.go
vhvb1989: Does it mean that, if user runs `azd foo` and this fails and starts the AI-troubleshooting agent trying to make some fixes, would eventually try to call `azd foo --no-prompt` to try again?   (preventing a new instance of the AI error troubleshooting and also interactive?)   But at the same time, would it change the user's expectations for the command?

JeffreyCA (approved):

cli/azd/internal/agent/tools/dev/command_executor.go
wbreza: No - this should only be in the case where the agent is making the call (via MCP tool)... The retry of the original function is re-executed by calling `next(ctx)` from within the middleware function and will run with the original user parameters.

cli/azd/internal/agent/tools/dev/command_executor.go
```diff
@@ -121,18 +122,12 @@ func (t CommandExecutorTool) Call(ctx context.Context, input string) (string, er
...
-				}
-
-				jsonData, _ := json.MarshalIndent(errorResponse, "", "  ")
-				return string(jsonData), nil
-			}
+		// Ensure --no-prompt is included in args to prevent interactive prompts
+		// that would block execution in agent/automation scenarios. The azd CLI
+		// may prompt for user input (confirmations, selections, etc.) which cannot
+		// be handled in a non-interactive agent context, so we force non-interactive mode.
+		if !slices.Contains(req.Args, "--no-prompt") {
```
vhvb1989, line 129: Should a dev/tool be doing this logic?  

Zooming out, I noticed this CommandExecutor is a tool for the Agent and we are making the tool aware of specific cases like how to call azd.

I wonder if this could be harmful or unexpected.

hemarina:
@wbreza During recent testing of the error-handling flow, it creates a long waiting time. Can we revisit whether this is expected behavior? If we decide to keep it, I recommend updating the UX to display the full azd command being executed. This would give users more visibility into what’s happening and help them understand the reason for the delay.
image

PR #5835, opened 2025-10-02
Agentic UX output tweaks

wbreza:
## Changes

- Moves `SendMessageWithRetry` to `agent`

### **UX Improvements for Agent Output** 

Fixes handling of repetitive tool calls to show output of each.
image

Updates file changes handling to detect new files from new folders
image

### 📝 **Line Endings Fix**

- Corrected line endings in Azure coding agent extension
- **Files modified:**
  - `cli/azd/extensions/azure.coding-agent/version.txt`

### DIFF
--- a/cli/azd/pkg/watch/watch.go
+++ b/cli/azd/pkg/watch/watch.go
@@ -12,6 +12,7 @@ import (
 	"sync"
 
 	"github.com/azure/azure-dev/cli/azd/pkg/output"
+	"github.com/bmatcuk/doublestar/v4"
 	"github.com/fatih/color"
 	"github.com/fsnotify/fsnotify"
 )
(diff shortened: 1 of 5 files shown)


cli/azd/pkg/watch/watch.go
```diff
@@ -84,19 +134,24 @@ func NewWatcher(ctx context.Context) (Watcher, error) {
...
-	if err := watchRecursive(cwd, watcher); err != nil {
+	if err := fw.watchRecursive(cwd, watcher); err != nil {
 		return nil, fmt.Errorf("watcher failed: %w", err)
 	}
 
 	return fw, nil
 }
 
-func watchRecursive(root string, watcher *fsnotify.Watcher) error {
+func (fw *fileWatcher) watchRecursive(root string, watcher *fsnotify.Watcher) error {
```
vhvb1989, line 144: I am against file watched because it will easily bring mistakes for environments where there could be more things happening at the same time as your agent is running. 
Or in a world where you want to run more than one agent at a time.

I would expect any tool that do I/O to return the list of modified changes as part of its contract.

Not a blocker for now.

vhvb1989 (approved):
adding one consern, but not blocking

cli/azd/pkg/watch/watch.go
wbreza: You are 100% correct that the current implementation could catch any IO changes made during the invocation of the agent... even direct user changes.  Long term we could build some of this into the local file tools if needed.  Will save that for another day :)

PR #5845, opened 2025-10-02
avoid saving IMAGE_NAME in env for containerapp target

weikanglim:
Avoid saving IMAGE_NAME in env for containerapp target. Instead, apply it in context when needed.

Makes it easier to reuse publishing for init/sidecar containers without stepping on toes.

### DIFF
--- a/cli/azd/pkg/project/service_target_containerapp.go
+++ b/cli/azd/pkg/project/service_target_containerapp.go
@@ -126,19 +126,6 @@ func (at *containerAppTarget) Publish(
 		}
 	}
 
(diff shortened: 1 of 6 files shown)


wbreza:
What is the backwards compatibility story for this?

vhvb1989 (commented):
LGTM but I wonder if we could have a little bit more of context?  Can we have an issue to describe what this is helping to?

weikanglim:
@vhvb1989 Logged #5865 but I won't be pushing for implementation for now.

PR #5856, opened 2025-10-03
Services dependencies

wbreza:
## New Feature: Service Dependencies

**What it does:** Services can now declare dependencies on other services/resources using a `uses` property. Services are automatically deployed in dependency order.

**Key capabilities:**
- Dependency ordering during deployment
- Circular dependency detection and prevention
- Backward compatible (existing configs work unchanged)

### Example YAML

```yaml
name: my-app

services:
  backend:
    project: ./api
    host: containerapp
  frontend:
    project: ./frontend
    host: staticwebapp
    uses:
      - backend
```

**Deployment order:** backend → frontend

**Error protection:** Prevents circular dependencies (e.g., if `database` tried to use `frontend`)

That's it - services deploy in the right order automatically.

### DIFF
--- a/cli/azd/pkg/project/importer_test.go
+++ b/cli/azd/pkg/project/importer_test.go
@@ -562,6 +562,14 @@ func TestImportManagerServiceStableWithDependencyOrdering(t *testing.T) {
 			errorMsg:    "circular dependency detected",
 		},
 		{
+			name: "self dependency",
+			services: map[string]*ServiceConfig{
+				"api": {Name: "api", Uses: []string{"api"}},
+			},
+			shouldError: true,
+			errorMsg:    "circular dependency detected",
+		},
+		{
 			name: "missing dependency",
 			services: map[string]*ServiceConfig{
 				"api": {Name: "api", Uses: []string{"nonexistent"}},


cli/azd/pkg/project/importer.go
```diff
@@ -92,11 +92,122 @@ func (im *ImportManager) ServiceStable(ctx context.Context, projectConfig *Proje
...
+func (im *ImportManager) validateServiceDependencies(services []*ServiceConfig, projectConfig *ProjectConfig) error {
+	serviceNames := make(map[string]bool)
+	for _, svc := range services {
+		serviceNames[svc.Name] = true
+	}
+
+	for _, svc := range services {
+		for _, dependency := range svc.Uses {
+			// Check if dependency exists as a service or resource
+			if !serviceNames[dependency] && (projectConfig.Resources == nil || projectConfig.Resources[dependency] == nil) {
```
vhvb1989, line 200: should there be a condition to catch if `svc.Name == dependency`? and error as service can't use itself?

vhvb1989 (approved):
LGTM

cli/azd/pkg/project/service_config.go
```diff
@@ -45,6 +45,8 @@ type ServiceConfig struct {
 	Infra provisioning.Options `yaml:"infra,omitempty"`
 	// Hook configuration for service
 	Hooks HooksConfig `yaml:"hooks,omitempty"`
+	// Dependencies on other services and resources
+	Uses []string `yaml:"uses,omitempty"`
```
weikanglim, line 49: `dependsOn` may be a better name if we're strictly using this to order deployment order. Unsure about the `resources` use case; that seems experimental right noww

cli/azd/pkg/project/importer.go
wbreza: The circurlar dependency check will catch self dependencies. I've added a test case to validate.

cli/azd/pkg/project/service_config.go
wbreza: I'm thinking longer term this will be a good story for integration into composability so the app code has more insights into what is available.  Might help support a local `azd run` scenario as well.

PR #5858, opened 2025-10-03
Promote extensions from alpha to beta

wbreza:
Promotes extension command from alpha to beta.


vhvb1989:
We will also need updates for the Docs @kristenwomack @alexwolfmsft

rajeshkamal5050 (approved):

PR #5861, opened 2025-10-04
Add experimental feed with 'azd coding-agent' in it.

richardpark-msft:


cli/azd/extensions/registry-experimental.json
vhvb1989, line 1: I've been using local extensions with the same name as non-local and that becomes really uncomfortable.

Before adding an extra registry, we might want to design the experience for an extension-owner going around between their local-dev, prod and now experimental versions.

@wbreza  @tg-msft

wbreza: #5858 has capabilities to provide a better experience when extensions exist in multiple sources.

Issue #5830, opened 2025-10-01
Extension auto-installation when deploying to a custom service target


JeffreyCA:
Related to #5752

Similar to how we have a flow to auto-install an extension (if not installed) when the user runs a custom extension command, we could extend that to handle azd packaging/publishing/deploying to a [custom service target](https://github.com/Azure/azure-dev/pull/5815).

Currently, users would see something like:

<img width="1519" height="110" alt="Image" src="https://github.com/user-attachments/assets/7327d46b-d00b-4f8d-985b-02aa95fe9a61" />

To support this, we may need to store metadata about what custom host types an extension supports, probably in the manifest or registry?

_Originally posted by @JeffreyCA in https://github.com/Azure/azure-dev/pull/5815#discussion_r2395966556_

PR #5863, opened 2025-10-04
auto-install - support for service target

vhvb1989:
fix: https://github.com/Azure/azure-dev/issues/5830


rajeshkamal5050:
@vhvb1989 auto-install when running `azd deploy` will be executed only when a non-core service target is encountered? so there isn't any overhead for the existing core service targets?

wbreza (commented):
This is too much work for the `azd` client to install/acquire all extensions that have service target capability support in order to support auto install.  This might be fast now but wouldn't be a maintainable solution as the registry grows.

Instead, I would prefer to explore adding additional metadata into the extension manifest within the registry so that we can quickly identify what service targes and other provider names are available.

vhvb1989:
> @vhvb1989 auto-install when running `azd deploy` will be executed only when a non-core service target is encountered? so there isn't any overhead for the existing core service targets?

correct. And like @wbreza mentioned, this is currently fast b/c there are only a few extensions.
I implemented this to pull all non-installed extensions which provides target-service and get the target-service list from each in parallel.  
As the number of extensions with service-targets grow, the time for looking should not increase but the CPU-demand will.

My current expectation is that the number of target-service is not going to grow crazy (to the order of hundreds or more!!).
I am guessing that there would be a top most of around 20 to 30 (exaggerating) service-target offered from extensions.
And there will also a constant review for any service-target getting very popular to be promoted into the core built-in.

That's the main motivation of creating this POC for dynamic discovery and installation.

Also, in the future, if the process for searching an extension for the service-targer becomes a problem, we can ask user before if they want azd to start the search, or open the extensions's marketplace (if it exists someday), or maybe use AI to search from extension docs.

This is currently a working approach which will not require updating all extensions or creates new rules (like adding the service-target or language to a manifest) for extensions. 

Note: The feature does NOT depend on the implementation. So if we decide to ask extension to define the service-targers in the manifest or do something different, we can change/update the auto-install implementation at any moment in the future w/o impacting how the feature behaves 😀

wbreza:
@vhvb1989 My PR #5858 - [Diff](https://github.com/Azure/azure-dev/pull/5858/files#diff-667a818b4f10ed6efc941fbba540559a62d5ab1e33c974ae663959723e049be3) adds new metadata to the registry/extension manifest to better support these scenarios.  

Once we get my PR in we can revisit this and implement a lower cost implementation that does not require downloading & installing all extensions with the service-target capability to determine whether or not an extension supports the required host type.

Once landed this will be a great improvement as extension adoption increases!

rajeshkamal5050:
@vhvb1989  @wbreza Let us get the registry enhanced. That will be better since we don't have any existing service-target/host extensions yet.

@wbreza extensions supporting service target override capability will contain something like this as a capability?
```
- name: foundry.hostedagent
- type: service-target/host
```

vhvb1989:
@rajeshkamal5050  I talked to @wbreza about this.

Ideally, I would really like to push for a "poka-yoke" design (mistake-proof) as the foundation for the extension platform. 

There should be already learned lessons from the templates world about what happen when there is no way to enforce controls as part of the design 😀.  You will find all sort of extension flavors and issues coming out of them, and try to implement controls on top of a poor design to guarantee a good experience for a curated list of extensions. Just like awesome-azd v/s ai-templates.  

That been said, I am closing this PR with just a left behind candy here 🍬

PR #5867, opened 2025-10-06
Makes language optional in the schema

wbreza:
To support hosts that support more dynamic languages or hosts that do not require any language such as declarative foundry agents we will turn off strict schema validation.

For example - the upcoming `foundry.hostedagent` can be used without any language when building declarative agents.  Since this service target comes from an extension we also do not have enough metadata to know whether or not language is required.

### DIFF
--- a/schemas/v1.0/azure.yaml.json
+++ b/schemas/v1.0/azure.yaml.json
@@ -112,7 +112,7 @@
                     "language": {
                         "type": "string",
                         "title": "Service implementation language",
-                        "enum": [
+                        "examples": [
                             "dotnet",
                             "csharp",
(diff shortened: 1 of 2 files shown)


schemas/v1.0/azure.yaml.json
```diff
@@ -261,8 +258,7 @@
                         },
                         "then": {
                             "required": [
-                                "project",
-                                "language"
+                                "project"
                             ],
                             "properties": {
                                 "docker": false
```
rajeshkamal5050, line 264: Will lang: docker fit for declarative agents? Also for code based it can be our usual lang set?

PR #5878, opened 2025-10-07
tracing: add mcp server instrumentation

### LINKED ISSUE #5747: Add telemetry tracking for all MCP tool usage
Need the ability to track all MCP tools usage:

Acceptance Criteria:
- Track calling MCP host, version (azd, VS Code Copilot, etc)
- Track tool calls (Include any Non-PII data linking to templates, subscriptions, etc when available)
- Can we track tool calls from withing the same user session?
- Track sampling usage (When a tool is using sampling through `azd`)

weikanglim:
Enable tracing for MCP server.

1. Add tracing spans for each MCP tool usage.
2. Enable background telemetry emission for MCP server.
3. Record client information.

Fixes #5747

### DIFF
--- a/cli/azd/.vscode/cspell-azd-dictionary.txt
+++ b/cli/azd/.vscode/cspell-azd-dictionary.txt
@@ -157,6 +157,7 @@ mgmt
 mgutz
 microsoftonline
 missingkey
+mmcp
 moby
 mockarmresources
 mockazcli


cli/azd/cmd/mcp.go
```diff
@@ -175,11 +180,48 @@ func (a *mcpStartAction) Run(ctx context.Context) (*actions.ActionResult, error)
...
+
+				select {
+				case <-ctx.Done():
+					return
+				case <-ticker.C:
+				}
+			}
+		}
+		go uploadTelemetry()
+	}
```
wbreza, line 201: Could this block move into the telemetry middleware? It doesn't feel like it should be a concern of the MCP server function

cli/azd/cmd/mcp.go
```diff
@@ -175,11 +180,48 @@ func (a *mcpStartAction) Run(ctx context.Context) (*actions.ActionResult, error)
...
+		server.WithToolHandlerMiddleware(func(next server.ToolHandlerFunc) server.ToolHandlerFunc {
+			return func(ctx context.Context, request mmcp.CallToolRequest) (result *mmcp.CallToolResult, err error) {
+				ctx, span := tracing.Start(ctx, "mcp."+request.Params.Name)
+				if session := server.ClientSessionFromContext(ctx); session != nil {
+					if sessionWithClientInfo, ok := session.(server.SessionWithClientInfo); ok {
+						clientInfo := sessionWithClientInfo.GetClientInfo()
+						span.SetAttributes(fields.McpClientName.String(clientInfo.Name))
+						span.SetAttributes(fields.McpClientVersion.String(clientInfo.Version))
+					}
+				}
```
wbreza, line 217: From within our individual tools does this still give us the capability to find this span from the context and add additional attributes if a tool wanted to capture something more specific?

cli/azd/cmd/mcp.go
weikanglim: We only do this for azd in server mode, where we want to have multiple long-running goroutines for execution. In proper azd (cli), this is kicked off automatically as a background async process.

cli/azd/cmd/mcp.go
weikanglim: Indeed, you can grab the span from context, set attributes on the span, or start a new child span. All are possible as part of opentelemetry go (related package [doc](https://pkg.go.dev/go.opentelemetry.io/otel/trace)).

cli/azd/cmd/mcp.go
wbreza: Not sure i'm following... since this is still spawned the same was as any other command can we just take advantage of that same process instead of having it here?

weikanglim: In normal CLI commands, the process runs once and exits when the command completes. Before it exits, we launch a background process to handle telemetry upload. This ensures telemetry is collected without delaying the user-facing command.

In this case, and the same case for vsrpc however, the server process is long-running and does not exit after a single operation. To make telemetry collection more efficient, we take advantage of this long-running nature. Instead of waiting for the process to end, we start a separate goroutine that periodically uploads telemetry while the server is running. This allows telemetry data to be sent incrementally and reduces the amount of work needed at shutdown.

weikanglim: In terms of organization, this function is responsible for orchestrating multiple long-running processes — the gRPC server, the MCP server, and others. Placing the telemetry collection logic here makes sense, since it aligns with the lifecycle management of those components. This allows telemetry to start and stop together with the rest of the system, keeping the control flow centralized and consistent.

wbreza: It just doesn't feel centralized and consistent when we have to add it into multiple areas within the app.  Instead of sprinkling the same code block in those mentioned areas it could just be centrally handled in the telemetry middleware for root commands/actions.

weikanglim: I understand the push for code organization. Currently, we do not have a centralized concept of a server higher up in the stack. Hence why we're deciding to add the code here to the mcp server implementation.

Moving it in into the middleware isn't quite what we want. IMO, commands like `azd extension list`, would end up spawning the goroutine and immediately terminating it. I'd vote for purposeful separation today, but we can always refactor later if this is a pressing concern.

PR #5882, opened 2025-10-08
[foundry extension] - service target support for deploy agent

vhvb1989:
fix: https://github.com/Azure/azure-dev-pr/issues/1743


cli/azd/extensions/azure.foundry.ai.agents/internal/cmd/listen.go
```diff
@@ -28,7 +28,7 @@ func newListenCommand() *cobra.Command {
 
 			provider := project.NewAgentServiceTargetProvider(azdClient)
 			host := azdext.NewExtensionHost(azdClient).
-				WithServiceTarget("foundry.hostedagent", provider)
+				WithServiceTarget("foundry.containeragent", provider)
```
wbreza, line 31: Is this the final name that was chosen?

cli/azd/extensions/azure.foundry.ai.agents/internal/project/remote_build.go
```diff
@@ -0,0 +1,544 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package project
```
wbreza, line 4: We should leverage the existing capability from `azd` directly instead of having to reinvent that here.

cli/azd/extensions/azure.foundry.ai.agents/internal/project/service_target_agent.go
```diff
@@ -30,14 +37,59 @@ func NewAgentServiceTargetProvider(azdClient *azdext.AzdClient) azdext.ServiceTa
...
+	if err != nil {
+		return fmt.Errorf("failed to get project: %w", err)
+	}
+	servicePath := serviceConfig.RelativePath
+	fullPath := filepath.Join(proj.Project.Path, servicePath)
+
+	log.Println(proj.Project.Path, fullPath)
+
+	// Check if user has specified agent definition path via environment variable
+	if envPath := os.Getenv("FOUNDRY_AGENT_DEFINITION_PATH"); envPath != "" {
```
wbreza, line 59: Why do we need an environment variable? The `agent.yaml` file should just be within the root of the service project...?

cli/azd/extensions/azure.foundry.ai.agents/internal/project/service_target_agent.go
```diff
@@ -164,62 +169,262 @@ func (p *AgentServiceTargetProvider) Deploy(
...
+			os.Stderr,
+			"Agent container '%s' (version: %s) start operation initiated (ID: %s).\n",
+			agentResponse.Name, agentResponse.Version, operation.Body.ID)
+	}
+
+	return &azdext.ServiceDeployResult{
+		TargetResourceId: "",
 		Kind:             "agent",
-		Endpoints:        endpoints,
+		Endpoints:        nil,
```
wbreza, line 425: Will we not have endpoints? I would expect the foundry agent service would provide the endpoints to the agent?

cli/azd/extensions/azure.foundry.ai.agents/internal/cmd/listen.go
vhvb1989: That's the current one used in the sample. We can change it later I guess.

cli/azd/extensions/azure.foundry.ai.agents/internal/project/service_target_agent.go
vhvb1989: You don't need it. It is a way to define the path if you don't have it in the root and with that name.

cli/azd/extensions/azure.foundry.ai.agents/internal/project/service_target_agent.go
vhvb1989: There are no endpoints or anything in the current sample.  
My understanding is that you would use the foundry project url plus the name and version of the agent to use it.  Idlk

cli/azd/extensions/azure.foundry.ai.agents/internal/project/remote_build.go
vhvb1989: for sure. But we'll add it to the list of things we want to expose.
Since it was already implemented here, I don't need to block the extension until it can leverage azd's remote deploy features

PR #5905, opened 2025-10-09
Updating the azd CI to not run for extensions.

richardpark-msft:


.github/workflows/cli-ci.yml
```diff
@@ -4,6 +4,7 @@ on:
   pull_request:
     paths:
       - "cli/**"
+      - "!cli/azd/extensions/**"
```
vhvb1989: Let's not remove this.
We do `golangci-lint` in this one, which should include extensions

richardpark-msft: Done!

PR #5906, opened 2025-10-09
Releasing the v0.2.0 of the `coding-agent` command in the official feed

richardpark-msft:


cli/azd/extensions/registry.json
```diff
@@ -423,6 +423,72 @@
...
+              "entryPoint": "azure-coding-agent-windows-arm64.exe",
+              "url": "https://github.com/Azure/azure-dev/releases/download/azd-ext-azure-coding-agent_0.2.0/azure-coding-agent-windows-arm64.zip"
+            }
+          }
+        }
+      ]
     }
   ]
-}
+}
```
RickWinter, line 494:
```suggestion
}

```

richardpark-msft: (I will fix this one afterwards, otherwise I'll lose my approval!)

richardpark-msft: (actually, I shouldn't fix this, that file is not "human" edited, so that's how it was produced)

Issue #5911, opened 2025-10-11
`azd up`/`azd deploy` of Aspire app with Dockerfile build fails when Dockerfile is in same directory as generated manifest.


mitchdenny:
- [x] Make sure you've installed the latest version using [instructions](https://github.com/Azure/azure-dev?tab=readme-ov-file#installupgrade-azure-developer-cli)

**Output from `azd version`**
Run `azd version` and copy and paste the output here:

`azd version 1.20.0 (commit 0031e30bc2c687abafc54bf589e07d3f1d40e954)`

**Describe the bug**
In Aspire 13.0 we are adding a feature that allows developers to dynamically generate a Dockerfile as part of the publishing process. In the case of Aspire publish triggered by `azd` this means that we will drop a docker file names `[resourcename].Dockerfile` in the same directory as the `aspire-manifest.json` output file.

Unfortunately, this doesn't appear to be working with `azd`. We are seeing that after the Aspire apphost has executed and it has placed the files on disk that `azd` is deleting the directory containing the aspire manifest and the generated dockerfiles.

This might be because `azd` doesn't expect any useful files to remain in this directory after the other files are proceeded - the new feature breaks this assumption.

We've documented the issue over on this PR which is fixing up another bug in our generation logic:

https://github.com/dotnet/aspire/pull/11958

**To Reproduce**

1. Acquire the latest (daily build) Aspire CLI: `iex "& { $(irm https://aspire.dev/install.ps1) } -quality dev"`
2. Create a new starter project: `aspire new` (select starter template)
3. Add the redis integration: `aspire add redis`
4. Add the Azure app containers integration: `aspire add azure-appcontainers`
5. Update `AppHost.cs`

```csharp
var builder = DistributedApplication.CreateBuilder(args);

builder.AddAzureContainerAppEnvironment("env");

var cache = builder.AddRedis("cache")
    .WithDockerfile(".", context =>
    {
        if (!context.Resource.TryGetContainerImageName(useBuiltImage: false, out var imageName))
        {
            throw new InvalidOperationException("Could not determine base image name.");
        }

        return
            $$"""
              FROM {{imageName}}
              """;
    });

var apiService = builder.AddProject<Projects.Hello11958_ApiService>("apiservice")
    .WithHttpHealthCheck("/health");

builder.AddProject<Projects.Hello11958_Web>("webfrontend")
    .WithExternalHttpEndpoints()
    .WithHttpHealthCheck("/health")
    .WithReference(cache)
[...]

PR #5916, opened 2025-10-14
update to support docker-in-mem generation

vhvb1989:
fix: https://github.com/Azure/azure-dev/issues/5911

Notes:
 - Implementation to handle `container.v1` updated to always load dockerfile content from `build.dockerfile` in-memory filesystem. Parsing manifest will fail if the dockerfile referenced in manifest does not exists. 
 - dotnet-importer checks if the dockerfile is saved in the expected path. If yes, it just uses it from there.
 If not, it uses the Dockerfile content from the in-memory file system.
   - This enabled running `azd infra gen` to write and use a Dockerfile from disk, but fallback to use the generated Dockerfile from the Apphost if the file is missing. Following the same pattern as bicep files.


cli/azd/pkg/project/framework_service_docker.go
```diff
@@ -310,10 +315,36 @@ func (p *dockerProject) Build(
...
+				log.Printf("removing temp dockerfile dir %s: %v", tempDir, err)
+			}
+		}()
+	}
+
 	imageId, err := p.docker.Build(
 		ctx,
 		serviceConfig.Path(),
-		dockerOptions.Path,
+		defaultPathForDockerfile,
```
weikanglim: minor nit: I'm finding `defaultPathForDockerfile` to be a confusing name when reading the overall scenario, consider `dockerFilePath`.

cli/azd/pkg/project/framework_service_docker.go
```diff
@@ -49,6 +49,11 @@ type DockerProjectOptions struct {
 	// Aspire would pass the secret keys, which are env vars that azd will set just to run docker build.
 	BuildSecrets []string `yaml:"-"                     json:"-"`
 	BuildEnv     []string `yaml:"-"                     json:"-"`
+	//InMemDockerfile allow projects to specify a dockerfile contents directly instead of a path on disk.
+	// This is not supported from azure.yaml.
+	// This is used by projects like Aspire that can generate a dockerfile on the fly and don't want to write it to disk.
+	// When this is set, whatever value in Path is ignored and the dockerfile contents in this property is used instead.
```
weikanglim, line 55: > whatever value in Path is ignored and the dockerfile contents in this property is used instead

I suspect this isn't happening right now, there's an earlier short-circuit for non-Dockerfile build. 
https://github.com/vhvb1989/azure-dev/blob/bb9a5759ab54797331defd9e1e5d6e6e89be404f/cli/azd/pkg/project/framework_service_docker.go#L287

Would you mind checking on this?

cli/azd/pkg/project/dotnet_importer.go
```diff
@@ -323,7 +323,26 @@ func (ai *DotNetImporter) Services(
...
-			relPath, err := filepath.Rel(p.Path, filepath.Dir(bContainer.Build.Dockerfile))
+			// dockerfiles are copied to the infra folder like bicep files to ensure
+			// provision and deploy works after infra gen.
+			bContainer.Build.Dockerfile = filepath.Join(
+				svcConfig.Project.Path, svcConfig.Project.Infra.Path, name, filepath.Base(bContainer.Build.Dockerfile))
+
+			// If the dockerfile is not in disk, it could have been manually deleted (after infra gen) or
+			// infra gen was never run. In any case, use the in-memory generated dockerfile to build the container.
+			var inMemDockerfile []byte
+			if _, err := os.Stat(bContainer.Build.Dockerfile); os.IsNotExist(err) {
```
weikanglim: nit: avoid deprecated code, use `errors.Is(err, os.ErrNotExist)`

Image

cli/azd/pkg/project/dotnet_importer.go
```diff
@@ -323,7 +323,26 @@ func (ai *DotNetImporter) Services(
 
 		if bContainer.Build != nil {
 			defaultLanguage = ServiceLanguageDocker
-			relPath, err := filepath.Rel(p.Path, filepath.Dir(bContainer.Build.Dockerfile))
+			// dockerfiles are copied to the infra folder like bicep files to ensure
+			// provision and deploy works after infra gen.
+			bContainer.Build.Dockerfile = filepath.Join(
+				svcConfig.Project.Path, svcConfig.Project.Infra.Path, name, filepath.Base(bContainer.Build.Dockerfile))
```
weikanglim, line 329: Requirements question: Is the desired behavior for `azd infra gen` to spit out the Dockerfile into infra directory, shifting the source of truth to the generated Dockerfile?

From reading @mitchdenny 's description in #5911, the file would be generated alongside `aspire-manifest.json`, which gave me a *slight* impression that this would be sourced from apphost each time the build runs.

weikanglim: Slightly related: in #5911, we call out that Aspire is generating this file in the publish apphost directory, with a comment:

> azd doesn't expect any useful files to remain in this directory after the other files are proceeded - the new feature breaks this assumption

Wondering out loud here: should we consider keeping the apphost directory around until the CLI command exits? This may be useful for "future features" that depend on this behavior.

vhvb1989: Yeah, I made this behavior by design. 
`azd infra gen` is the way for moving the source of truth from AppHost to local files (representing deployable artifacts). 

I initially implemented this by always just using the Dockerfile from memory. But I am pretty sure people would ask and expect to have a way to see how that Dockerfile looks like from `infra gen`.  And I was feeling all mixed-up to have bicep files combined with in-mem-only files.

The current implemented design follows the same approach we have for all services. If you remove the service's generated files (bicep and Dockerfile), you automatically switch to use in-mem generated files for that service only.

cli/azd/pkg/project/framework_service_docker.go
vhvb1989: That short-circuit is for when there is no `path` provided.
This new option is to ignore the value from path.

PR #5922, opened 2025-10-14
auto install service targ

vhvb1989:
Supporting auto-install for service-targets
fix: https://github.com/Azure/azure-dev/issues/5830

### DIFF
--- a/cli/azd/pkg/project/mapper_registry.go
+++ b/cli/azd/pkg/project/mapper_registry.go
@@ -188,48 +188,6 @@ func registerProjectMappings() {
 		}, nil
 	})
 
-	// ServiceRestoreResult -> proto ServiceRestoreResult conversion
-	mapper.MustRegister(func(ctx context.Context, src *ServiceRestoreResult) (*azdext.ServiceRestoreResult, error) {
-		if src == nil {
-			return nil, nil
-		}
-
-		protoResult := &azdext.ServiceRestoreResult{}
-
-		details := detailsInterfaceToStringMap(src.Details)
-		if len(details) > 0 {
-			protoResult.Details = details
-		}
-
-		return protoResult, nil
-	})
-
-	// ServiceBuildResult -> proto ServiceBuildResult conversion
-	mapper.MustRegister(func(ctx context.Context, src *ServiceBuildResult) (*azdext.ServiceBuildResult, error) {
-		if src == nil {
-			return nil, nil
-		}
-
-		protoResult := &azdext.ServiceBuildResult{}
-
-		// Convert nested Restore result if present
-		if src.Restore != nil {
-			var restoreResult *azdext.ServiceRestoreResult
-			err := mapper.Convert(src.Restore, &restoreResult)
-			if err != nil {
-				return nil, fmt.Errorf("converting restore result: %w", err)
-			}
-			protoResult.Restore = restoreResult
-		}
-
-		details := detailsInterfaceToStringMap(src.Details)
-		if len(details) > 0 {
-			protoResult.Details = details
-		}
-
-		return protoResult, nil
-	})
-
 	// ServicePublishResult -> proto ServicePublishResult conversion
 	mapper.MustRegister(func(ctx context.Context, src ServicePublishResult) (*azdext.ServicePublishResult, error) {
 		var artifacts []*azdext.Artifact


cli/azd/cmd/auto_install.go
```diff
@@ -383,3 +442,100 @@ func ExecuteWithAutoInstall(ctx context.Context, rootContainer *ioc.NestedContai
...
+	}
+
+	if len(nonInstalledExtensions) == 0 {
+		return nil, nil
+	}
+
+	resultChan := make(chan discoveryResult, len(nonInstalledExtensions))
+	var wg sync.WaitGroup
+
+	// Launch a goroutine for each non-installed extension
```
weikanglim: what was the perf bottleneck that led us down spinning up goroutines per each entry here?

cli/azd/cmd/auto_install.go
```diff
@@ -383,3 +442,100 @@ func ExecuteWithAutoInstall(ctx context.Context, rootContainer *ioc.NestedContai
...
+		return nil, fmt.Errorf("failed to list extensions from registry: %w", err)
+	}
+
+	// Filter out already installed extensions
+	var nonInstalledExtensions []*extensions.ExtensionMetadata
+	for _, ext := range registryExtensions {
+		_, err := extensionManager.GetInstalled(extensions.FilterOptions{
+			Id: ext.Id,
+		})
+		if err != nil {
```
weikanglim: consider checking for a specific error

cli/azd/cmd/middleware/ux.go
```diff
@@ -43,15 +45,30 @@ func (m *UxMiddleware) Run(ctx context.Context, next NextFn) (*actions.ActionRes
...
 		if errors.As(err, &suggestionErr) {
-			m.console.Message(ctx, suggestionErr.Suggestion)
+			errorMessage.WriteString("\n" + suggestionErr.Suggestion)
 		}
+
+		// UnsupportedServiceHostError is a special error which needs to float up without printing output here yet.
+		// The error is bubble up for the caller to decide to show it or not
+		var unsupportedErr *project.UnsupportedServiceHostError
+		errMessage := errorMessage.String()
+		if errors.As(err, &unsupportedErr) {
```
weikanglim, line 64: consider putting this logic in a better place to surface a library error like `project.UnsupportedServiceHostError`

cli/azd/cmd/auto_install.go
```diff
@@ -271,13 +273,76 @@ func ExecuteWithAutoInstall(ctx context.Context, rootContainer *ioc.NestedContai
...
 	rootCmd := NewRootCmd(false, nil, rootContainer)
 
+	var extensionManager *extensions.Manager
+	if err := rootContainer.Resolve(&extensionManager); err != nil {
+		log.Panic("failed to resolve extension manager for auto-install:", err)
+	}
+	var console input.Console
+	if err := rootContainer.Resolve(&console); err != nil {
+		log.Panic("failed to resolve console for unknown flags error:", err)
+	}
```
weikanglim: for general hygiene, consider placing this logic outside the happy path where it's not needed

cli/azd/cmd/auto_install.go
```diff
@@ -271,13 +273,76 @@ func ExecuteWithAutoInstall(ctx context.Context, rootContainer *ioc.NestedContai
...
+
 	// rootCmd.Find() returns error if the command is not identified. Cobra checks all the registered commands
 	// and returns error if the input command is not registered.
 	// This allows us to determine if a subcommand was provided or not or if the command is unknown.
 	_, originalArgs, err := rootCmd.Find(os.Args[1:])
 	if err == nil {
 		// Known command, no need to auto-install
-		return rootCmd.ExecuteContext(ctx)
+		err := rootCmd.ExecuteContext(ctx)
```
weikanglim, line 283: readability: consider making the happy path exit more explicit and attach the comment there

cli/azd/cmd/auto_install.go
```diff
@@ -271,13 +273,76 @@ func ExecuteWithAutoInstall(ctx context.Context, rootContainer *ioc.NestedContai
...
+				if err != nil {
+					console.Message(ctx, fmt.Sprintf("Error selecting extension: %v", err))
+					return err
+				}
+				extensionIdToInstall = *chosenExtension
+			}
+
+			installed, installErr := tryAutoInstallExtension(ctx, console, extensionManager, extensionIdToInstall)
+			if installErr != nil {
+				// Error needs to be printed here or else it will be hidden b/c the error printing is handled inside runtime
```
weikanglim: Is this a TODO? We have many places here in the code where we explicitl print out the error message here, rather than following general go conventions of bubbling error, which is puzzling

cli/azd/cmd/middleware/ux.go
weikanglim: I see this is used as a suppression -- this is probably fine

cli/azd/cmd/auto_install.go
```diff
@@ -271,13 +273,76 @@ func ExecuteWithAutoInstall(ctx context.Context, rootContainer *ioc.NestedContai
...
+			}
+			filterMatches := []*extensions.ExtensionMetadata{}
+			for _, result := range discoveryResults {
+				if slices.Contains(result.serviceTargets, unsupportedErr.Host) {
+					filterMatches = append(filterMatches, result.extensionMetadata)
+				}
+			}
+			if len(filterMatches) == 0 {
+				// did not find an extension with the capability, just print the original error message
+				console.Message(ctx, unsupportedErr.ErrorMessage)
```
weikanglim: is this a bug where we don't return?

cli/azd/pkg/project/framework_service_external.go
```diff
@@ -464,8 +464,14 @@ func (efs *ExternalFrameworkService) toProtoServiceConfig(serviceConfig *Service
 		return nil, nil
 	}
 
+	// Use an empty resolver since ExternalFrameworkService doesn't have access to environment
+	// The extension is responsible for handling environment variable substitution
+	emptyResolver := func(key string) string {
+		return ""
+	}
```
wbreza, line 448: If we need an environment we can inject it into the framework service otherwise this is fine.

cli/azd/cmd/auto_install.go
```diff
@@ -383,3 +442,100 @@ func ExecuteWithAutoInstall(ctx context.Context, rootContainer *ioc.NestedContai
...
+	err               error
+}
+
+// DiscoverServiceTargetCapabilities discovers service target capabilities from all non-installed extensions
+// by temporarily pulling their binaries and checking what service targets they provide.
+func DiscoverServiceTargetCapabilities(
+	ctx context.Context, extensionManager *extensions.Manager) ([]discoveryResult, error) {
+	// Get all extensions from registry
+	options := &extensions.FilterOptions{}
+	registryExtensions, err := extensionManager.FindExtensions(ctx, options)
```
wbreza: The `FilterOptions` should now include the capabilities as well as provider if we know it in advance so we should be able to easily find the extension that supports the given provider name + capability, then just check if it is already installed or not.

cli/azd/cmd/auto_install.go
vhvb1989: Not a TODO.
It explains why we are calling `console.Message(ctx, installErr.Error())` instead of returning the error and expecting the caller to print the error.

We have disabled the Cobra built-in feature to display errors. Instead, we use a middleware in azd to make our errors looks pretty.
But, auto-install is out of the middlewares pipeline. 
So the print-error middleware now ignores the `project.UnsupportedServiceHostError` error, which is used to signal auto-install to look for an extension.

That's why, if auto-install fails, it needs to print the original error (including the formatting generated by the error-framework) 🎃

yes, i tis puzzling !!  Hopefully we can refactor/improve this in the future

cli/azd/cmd/auto_install.go
```diff
@@ -231,21 +229,23 @@ func tryAutoInstallExtension(
...
-				"Command '%s' not found, but there's an available extension that provides it.\n"+
-					"However, auto-installation is not supported in CI/CD environments.\n"+
+				"Auto-installation is not supported in CI/CD environments.\n"+
 					"Run '%s' to install it manually.",
-				extension.Namespace,
 				fmt.Sprintf("azd extension install %s", extension.Id))
 	}
 
+	console.MessageUxItem(ctx, &ux.WarningMessage{
+		Description: "You are about to install an extension !!",
```
JeffreyCA:
```suggestion
		Description: "You are about to install an extension!",
```

JeffreyCA (approved):

rajeshkamal5050 (approved):

Issue #5949, opened 2025-10-17
[Bug] When auth is delegated to az, running azd auth login should exit fast and mentioned auth is in delegated to az mode


vhvb1989:

weikanglim:
I'm wondering if we had a concrete use ask here. I feel that this is quite an investment we're making.

The goal I'd love to have in the back of our minds is push towards using azd, not expand support for az.

vhvb1989:
> I'm wondering if we had a concrete use ask here. I feel that this is quite an investment we're making.
> 
> The goal I'd love to have in the back of our minds is push towards using azd, not expand support for az.

Yeah, I know.
I was talking to @chlowell and apparently it is simpler to recommend and use az-delegated auth if folks are looking to leverage one-auth (since it is supported in az).  Compared with the amount of work we would need to make one-auth work in azd

chlowell:
I do see `useAzcliAuth` as useful primarily because it enables indirect broker access but in my mind this issue is more about making the feature consistent and intuitive.

When I'm not logged in to az, I see an error like "Please run 'az login' to setup account" when I run an azd command that requires authN. That makes sense to me. But if I run `azd auth login`, it launches the browser login flow. This is surprising because I thought I configured azd to delegate authN to az. Was it az or azd that opened the browser? Why didn't it open the broker's login window?

The result can be confusing as well. It depends on whether I'm logged in to az:
1. if I'm not logged in to az, `azd auth login` fails, telling me to run `az login` 
2. if I'm logged in to az, `azd auth login` succeeds

I suppose I know what happened with result 1, but how do I interpret result 2? What happened in azd and was there any effect on az?

weikanglim:
@chlowell Thanks for clarifying! I do agree that `azd auth login` silently prompting the browser in this case is unexpected.

I'm definitely onboard with erroring out and let the user know to use `az login` when `az. useAzcliAuth` set.

PR #5954, opened 2025-10-20
Better handle auth when not using built-in auth

vhvb1989:
fix: https://github.com/Azure/azure-dev/issues/5949


cli/azd/cmd/auth_login.go
```diff
@@ -288,6 +288,32 @@ func newLoginAction(
...
+		if err != nil {
+			return nil, err
+		}
+		if !response {
+			return nil, fmt.Errorf("log in is not supported on current mode: %s", loginMode)
+		}
+		if err := la.authManager.SetMode(ctx, auth.AzdBuiltIn); err != nil {
+			return nil, fmt.Errorf("setting auth mode: %w", err)
+		}
+	}
```
wbreza, line 283: What is the motivation for this? Are we seeing people using az legacy mode when they shouldn't be? Or didn't know there were and we want to reset for them?

vhvb1989: I was talking to @chlowell the other day (at the office!!!!  😎) .  He's done some work for one-auth, which works for az cli but it's not 100% completed for azd.  If folks want to use one-auth, they need to delegate auth to az.
But the current behavior having azd still doing the `auth login` flow after setting it to use az was confusing.

So I started this fix to make it simpler and let folks to use `azd auth login` as a way to switch back to build-in auth if they want (as a plus)

cli/azd/cmd/auth_login.go
```diff
@@ -288,6 +288,34 @@ func newLoginAction(
...
+	loginMode, err := la.authManager.Mode()
+	if err != nil {
+		return nil, err
+	}
+	if loginMode != auth.AzdBuiltIn {
+		la.console.MessageUxItem(ctx, &ux.WarningAltMessage{
+			Message: fmt.Sprintf(
+				"Azd is not using the built-in authentication mode, but rather '%s'", loginMode),
+		})
+		la.console.Message(ctx, "If you want to use 'azd auth login', you need to disable the current auth mode.")
```
weikanglim, line 265: This removes the use for `azd auth login --check-status` to verify azd is logged in using az.

vhvb1989: added `!la.flags.onlyCheckStatus` to the condition here. That should

rajeshkamal5050:
@vhvb1989 can we move this PR along?

vhvb1989:
@copilot open a new pull request to apply changes based on the comments in [this thread](https://github.com/Azure/azure-dev/pull/5954#pullrequestreview-3585813822)

cli/azd/cmd/auth_login.go
```diff
@@ -288,6 +288,34 @@ func newLoginAction(
...
 
 func (la *loginAction) Run(ctx context.Context) (*actions.ActionResult, error) {
+	loginMode, err := la.authManager.Mode()
+	if err != nil {
+		return nil, err
+	}
+	if loginMode != auth.AzdBuiltIn && !la.flags.onlyCheckStatus {
+		la.console.MessageUxItem(ctx, &ux.WarningAltMessage{
+			Message: fmt.Sprintf(
+				"Azd is not using the built-in authentication mode, but rather '%s'", loginMode),
```
weikanglim: Consider renaming Azd -> azd consistently

cli/azd/pkg/auth/manager.go
```diff
@@ -1409,3 +1409,67 @@ func (m *Manager) LogInDetails(ctx context.Context) (*LogInDetails, error) {
...
 	return nil, ErrNoCurrentUser
 }
+
+type AuthMode string
+
+const (
+	AzdBuiltIn      AuthMode = "azd built in"
+	AzDelegated     AuthMode = "delegated to az cli"
+	ExternalRequest AuthMode = "external token request"
+)
```
weikanglim: It might be worth standardizing on what formal names we want for these options.

In the user message, we currently say "Azure CLI authentication" and "External request".

2c suggestion is to treat each auth as an "AuthSource", and we would say "az cli", "external endpoint"

cli/azd/pkg/auth/manager.go
```diff
@@ -1409,3 +1409,67 @@ func (m *Manager) LogInDetails(ctx context.Context) (*LogInDetails, error) {
...
+	if err != nil {
+		return fmt.Errorf("fetching current auth mode: %w", err)
+	}
+	if currentMode == AzdBuiltIn {
+		return nil
+	}
+
+	if currentMode == ExternalRequest {
+		return fmt.Errorf("cannot change auth mode when external token mode is set. See %s",
+			"https://github.com/Azure/azure-dev/blob/main/cli/azd/docs/external-authentication.md")
```
weikanglim: I like the use of github docs here! Does it make sense to set up an `aka.ms` link just in case this gets reorganized in the future? It's fairly cheap and I recently set one up.

weikanglim (approved):
LGTM. Thanks for handling the `check-status` flag.

weikanglim (approved):
Looks good @vhvb1989 ! Thanks for doing this!

PR #5955, opened 2025-10-20
Add command to generate Fig spec for VS Code terminal IntelliSense

### LINKED ISSUE #5656: VS Code integrated terminal Intellisense support for `azd`
There's a new [VS Code preview feature](https://code.visualstudio.com/docs/terminal/shell-integration#_intellisense-preview) that adds Intellisense to the integrated terminal, powered by [Fig autocomplete specs](https://github.com/withfig/autocomplete):

Image

The default ones bundled with VS Code are located here (looks like some are pulled from upstream [withfig/autocomplete](https://github.com/withfig/autocomplete)): https://github.com/microsoft/vscode/tree/main/extensions/terminal-suggest/src/completions.

Maybe we could contribute a spec for `azd` upstream?

Relevant:
https://code.visualstudio.com/docs/terminal/shell-integration#_intellisense-preview
https://fig.io/docs/guides/integrating/integrations/cobra

JeffreyCA:
Resolves #5656
Related: https://github.com/microsoft/vscode/pull/272348

This PR adds a new command `azd completion fig` to automatically generate a [Fig autocomplete spec](https://fig.io/docs/concepts/cli-skeleton) for integration with VS Code's terminal IntelliSense. The Fig spec for azd used by VS Code IntelliSense lives in the vscode repo at [`extensions/terminal-suggest/src/completions/azd.ts`](https://github.com/microsoft/vscode/blob/main/extensions/terminal-suggest/src/completions/azd.ts).

image

image

The command works by traversing the Cobra command tree to extract all commands, flags, and arguments and converts this information to TypeScript, following the [spec format](https://github.com/microsoft/vscode/blob/main/extensions/terminal-suggest/src/completions/index.d.ts). Hidden commands and flags are omitted from the spec, but there is a `--include-hidden` flag (also hidden) that can be passed.

### Dynamic suggestions

The azd spec leverages [Fig generators](https://fig.io/docs/reference/generator) to dynamically provide context-aware suggestions for certain commands.

For example, when the user types `azd init -t`, VS Code in the background executes and parses `azd template list --output json` to provide dynamic suggestions:

image

Other commands with dynamic suggestions include:

- `azd env select`
- `azd env get-value`
- `azd extension install`

Open to thoughts but I think this is a sufficient enough initial spec - there's plenty of opportunity to extend the spec further and we can track them as separate issues.

### Implementation/architecture
See documentation in `cli/azd/docs/fig-spec.md`

### Testing

This PR adds a new `figspec_test.go` snapshot test. The test creates a snapshot of the Fig spec at `cli/azd/cmd/testdata/TestFigSpec.ts`. Similar to the `TestUsage` snapshot tests, whenever the command structure changes (such as when adding a new command or flag), the Fig spec snapshot needs to be updated as well or the test would fail:

```bash
# Update snapshot when commands change
UPDATE_SNAPSHOTS=true go test ./cmd -run TestFigSpec
```

### DIFF
--- a/cli/azd/.vscode/cspell.yaml
+++ b/cli/azd/.vscode/cspell.yaml
@@ -210,6 +210,9 @@ overrides:
       - azdcontext
       - azapi
       - eastus
+  - filename: cmd/cmd_help.go
+    words:
+      - cmdinternal
   - filename: .github/copilot-instructions.md
     words:
       - golines


cli/azd/internal/figspec/fig_generators.go
```diff
@@ -0,0 +1,213 @@
...
+	FigGenListExtensions = "azdGenerators.listExtensions"
+
+	// FigGenListInstalledExtensions generates suggestions from installed extensions only
+	FigGenListInstalledExtensions = "azdGenerators.listInstalledExtensions"
+)
+
+// figGeneratorDefinitions returns the TypeScript code defining all Fig generators and interfaces used by them.
+// The generator names in this code must match the constants defined above.
+func figGeneratorDefinitions() string {
+	return `interface AzdEnvListItem {
```
wbreza: nit: Instead of inlining this TypeScript stuff could we save this as an embedded resource and return it from there? It would be easier to catch TS syntax issues/errors can get better IntelliSense during updates?

wbreza (approved):
This is awesome!

cli/azd/internal/figspec/spec_builder.go
```diff
@@ -0,0 +1,256 @@
...
+		names := []string{"--" + flag.Name}
+		if flag.Shorthand != "" {
+			names = append(names, "-"+flag.Shorthand)
+		}
+
+		isRepeatable := strings.Contains(flag.Value.Type(), "Slice") ||
+			strings.Contains(flag.Value.Type(), "Array")
+
+		isRequired := false
+		if annotations := flag.Annotations["cobra_annotation_bash_completion_one_required_flag"]; len(annotations) > 0 {
```
weikanglim: nit: would be nice to reuse `cobra.BashCompOneRequiredFlag`, and explain this more nicely in a comment:

```golang
// Handle flags that are marked as `cmd.MarkFlagRequired`
```

cli/azd/internal/figspec/customizations.go
```diff
@@ -0,0 +1,199 @@
...
+var serviceCommandPaths = []string{
+	"azd build",
+	"azd deploy",
+	"azd package",
+	"azd publish",
+	"azd restore",
+}
+
+// Flags that should only appear at root level, not duplicated on subcommands
+var persistentOnlyFlags = []string{
```
weikanglim: Let's remove this duplication that is currently maintained in both azd and figspec, and reuse `root.PersistentFlags()`.

If we simply reused `root.PersistentFlags()`, I believe this will still lack two flags that are special to `azd`: `--help` and `--docs`, which are under-the-covers generated per-command. I think we should make a quick refactor for this. 

Inside `cmd/cmd_help.go`, L97, we have these constants already defined:

```golang
	// force the following flags as global flags for display purposes when displaying help.
	forceGlobalFlagNames := map[string]struct{}{
		"help": {},
		"docs": {},
	}
```

It makes sense to refactor this out as a constant value `NonPersistentGlobalFlags`. Combining the names of `root.PersistentFlags()` and `NonPersistentGlobalFlags` should get us back to this list.

I think this is easy enough for us to clean up, and would feel strongly about removing a source of duplication.

cli/azd/internal/figspec/resources/generators.ts
```diff
@@ -0,0 +1,179 @@
...
+			} catch {
+				return [];
+			}
+		},
+		cache: {
+			strategy: 'stale-while-revalidate',
+		}
+	},
+	listInstalledExtensions: {
+		script: ['azd', 'ext', 'list', '--installed', '--output', 'json'],
```
weikanglim, line 158: as a V2 improvement, I would love if we could move the `script` definition here back into core azd completions. If we did this, we would keep the `postProcess` around because that is still useful

cli/azd/internal/figspec/customizations.go
```diff
@@ -0,0 +1,199 @@
...
+
+// GetSuggestions returns static suggestion values for flags that accept a fixed set of options
+func (c *Customizations) GetSuggestions(ctx *FlagContext) []string {
+	path := ctx.CommandPath
+	flagName := ctx.Flag.Name
+
+	switch path {
+	case "azd auth login":
+		if flagName == "federated-credential-provider" {
+			return []string{"github", "azure-pipelines", "oidc"}
```
weikanglim, line 46: It may be worth creating an issue to move these completions into core azd, and avoid the ownership here in figspec. What are your thoughts?

The most native way is when to pull out the flag completion funcs defined in cobra as below:

```golang
cmd.RegisterFlagCompletionFunc("mode", func(
    cmd *cobra.Command, args []string, toComplete string,
) ([]string, cobra.ShellCompDirective) {
    return []string{"dev", "staging", "prod"}, cobra.ShellCompDirectiveNoFileComp
})
```

cli/azd/internal/figspec/customizations.go
```diff
@@ -0,0 +1,199 @@
...
+		return []Arg{
+			{Name: "resource-name|resource-id", IsOptional: true},
+		}
+	case "azd hooks run":
+		return []Arg{
+			{Name: "name", Suggestions: hookNameValues},
+		}
+	case "azd extension install":
+		return []Arg{
+			{Name: "extension-id"},
```
weikanglim: Is the "right fix" here to rename the current arg `extension-name` in `azd extension install <extension-name> ` to `extension-id`?

cli/azd/internal/figspec/customizations.go
```diff
@@ -0,0 +1,199 @@
...
+func (c *Customizations) GetCommandArgs(ctx *CommandContext) []Arg {
+	switch ctx.CommandPath {
+	case "azd env set":
+		return []Arg{
+			{Name: "key", IsOptional: true},
+			{Name: "value", IsOptional: true},
+		}
+	case "azd show":
+		return []Arg{
+			{Name: "resource-name|resource-id", IsOptional: true},
```
weikanglim: I don't mind if we want to update `azd show` to `azd show [resource-name|resource-id]`

cli/azd/internal/figspec/spec_builder.go
```diff
@@ -0,0 +1,256 @@
...
+		Description: "Help about any command",
+		Subcommands: sb.generateHelpSubcommands(root),
+	}
+}
+
+func (sb *SpecBuilder) generateHelpSubcommands(cmd *cobra.Command) []Subcommand {
+	var subcommands []Subcommand
+
+	for _, sub := range cmd.Commands() {
+		if sub.Hidden || sub.Name() == "help" {
```
weikanglim: nit: this doesn't respect `sb.includeHidden` as well, perhaps it should

cli/azd/docs/fig-spec.md
```diff
@@ -0,0 +1,182 @@
...
+
+A Fig spec is a TypeScript object that declaratively describes a CLI tool's interface, including commands, subcommands, flags/options, positional arguments, and dynamic generators for context-aware completions. The type definition used by VS Code is available at [`index.d.ts`](https://github.com/microsoft/vscode/blob/main/extensions/terminal-suggest/src/completions/index.d.ts). The original reference documentation can be found on [Fig's official docs](https://fig.io/docs/reference/subcommand).
+
+The Fig spec used for IntelliSense lives in the VS Code repo at [`extensions/terminal-suggest/src/completions/azd.ts`](https://github.com/microsoft/vscode/blob/main/extensions/terminal-suggest/src/completions/azd.ts).
+
+A copy of it exists in this repo under `cli/azd/cmd/testdata/TestFigSpec.ts` for snapshot testing purposes.
+
+## Usage
+
+### Generating the spec
```
weikanglim, line 17: This isn't covering a very important question around dev setup that I have in the back of my mind:

1. alpha features: Should we enable all, or disable all `alpha` features while generating? This impacts the command structure generator. I'd vote for enabling. I wouldn't be too concerned about these commands showing up.
1. extensions: Should we remove all extensions while generating? Or perhaps should the generation be excluding extension commands?

cli/azd/internal/figspec/customizations.go
JeffreyCA: Agreed, thanks!

cli/azd/internal/figspec/resources/generators.ts
JeffreyCA: Created #6114

cli/azd/internal/figspec/customizations.go
JeffreyCA: Tracking in https://github.com/Azure/azure-dev/issues/6114

cli/azd/docs/fig-spec.md
JeffreyCA: Thanks for bringing this up - currently it's not much different from `usage_test` where it expects no extensions installed because those tests run in our pipelines without extensions installed as well.

I think a good next enhancement would be to include extension commands since it'll complement the [auto install feature](https://github.com/Azure/azure-dev/pull/5753) quite nicely.

JeffreyCA: https://github.com/Azure/azure-dev/issues/6121

PR #5959, opened 2025-10-21
support cobo with preprovision and postdeploy

vhvb1989:
fix: https://github.com/Azure/azure-dev/issues/5927

### DIFF
--- a/cli/azd/extensions/azure.foundry.ai.agents/internal/cmd/listen.go
+++ b/cli/azd/extensions/azure.foundry.ai.agents/internal/cmd/listen.go
@@ -29,8 +29,9 @@ func newListenCommand() *cobra.Command {
 
 			provider := project.NewAgentServiceTargetProvider(azdClient)
 			projectParser := &project.FoundryParser{AzdClient: azdClient}
+			// IMPORTANT: service target name here must match the name used in the extension manifest.
 			host := azdext.NewExtensionHost(azdClient).
-				WithServiceTarget("foundry.agent", provider).
+				WithServiceTarget("foundry.containeragent", provider).
 				WithProjectEventHandler("preprovision", projectParser.SetIdentity).
 				WithProjectEventHandler("postdeploy", projectParser.CoboPostDeploy)
 
--- a/cli/azd/cmd/auto_install.go
+++ b/cli/azd/cmd/auto_install.go
@@ -5,6 +5,7 @@ package cmd
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"log"
 	"os"
@@ -14,6 +15,8 @@ import (
 	"github.com/azure/azure-dev/cli/azd/pkg/extensions"
 	"github.com/azure/azure-dev/cli/azd/pkg/input"
 	"github.com/azure/azure-dev/cli/azd/pkg/ioc"
+	"github.com/azure/azure-dev/cli/azd/pkg/output/ux"
+	"github.com/azure/azure-dev/cli/azd/pkg/project"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
 )
@@ -159,22 +162,17 @@ func promptForExtensionChoice(
 	extensions []*extensions.ExtensionMetadata) (*extensions.ExtensionMetadata, error) {
 
 	if len(extensions) == 0 {
-		return nil, nil
+		return nil, fmt.Errorf("no extensions to choose from")
 	}
 
 	if len(extensions) == 1 {
 		return extensions[0], nil
 	}
 
-	console.Message(ctx, "Multiple extensions found that match your command:")
-	console.Message(ctx, "")
-
 	options := make([]string, len(extensions))
 	for i, ext := range extensions {
-		options[i] = fmt.Sprintf("%s (%s) - %s", ext.Namespace, ext.DisplayName, ext.Description)
-		console.Message(ctx, fmt.Sprintf("  %d. %s", i+1, options[i]))
+		options[i] = fmt.Sprintf("%s (%s) - %s", ext.DisplayName, ext.Source, ext.Description)
 	}
-	console.Message(ctx, "")
 
 	choice, err := console.Select(ctx, input.ConsoleOptions{
 		Message: "Which extension would you like to install?",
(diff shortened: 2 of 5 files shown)


cli/azd/extensions/azure.foundry.ai.agents/internal/project/parser.go
```diff
@@ -0,0 +1,1310 @@
...
+		if part == "subscriptions" && i+1 < len(parts) {
+			subscription = parts[i+1]
+		}
+		if part == "resourceGroups" && i+1 < len(parts) {
+			resourceGroup = parts[i+1]
+		}
+		if part == "containerApps" && i+1 < len(parts) {
+			appName = parts[i+1]
+		}
+	}
```
JeffreyCA: May be better to use `arm.ParseResourceID` to parse resource IDs?

cli/azd/extensions/azure.foundry.ai.agents/internal/project/parser.go
```diff
@@ -0,0 +1,1291 @@
...
+
+	if err := validateRequired("SERVICE_API_RESOURCE_ID", resourceID); err != nil {
+		return err
+	}
+
+	// Deactivate hello-world revision
+	if err := deactivateHelloWorldRevision(ctx, resourceID); err != nil {
+		fmt.Fprintf(os.Stderr, "Warning: Failed to deactivate hello-world revision: %v\n", err)
+		// Don't return error, just warn - this is not critical for the deployment
+	}
```
wbreza, line 470: Wouldn't the standard ACA deployment already set the new revision? Unclear why we would need to deactivate some revision manually.

cli/azd/extensions/azure.foundry.ai.agents/internal/project/parser.go
```diff
@@ -0,0 +1,1291 @@
...
+func deactivateHelloWorldRevision(ctx context.Context, resourceID string) error {
+	fmt.Println()
+	fmt.Println("======================================")
+	fmt.Println("Deactivating Hello-World Revision")
+	fmt.Println("======================================")
+	fmt.Println("ℹ️  Azure Container Apps requires an image during provision, but with remote Docker")
+	fmt.Println("   build, the app image doesn't exist yet. A hello-world placeholder image is used")
+	fmt.Println("   during 'azd provision', then replaced with your app image during 'azd deploy'.")
+	fmt.Println("   Now that your app is deployed, we'll deactivate the placeholder revision.")
+	fmt.Println()
```
wbreza, line 723: Still unclear why we need this... We don't do anything like this for regular ACA that follows the same flow.

cli/azd/extensions/azure.foundry.ai.agents/internal/project/parser.go
```diff
@@ -0,0 +1,1291 @@
...
+	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/resources/armresources"
+	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/resources/armsubscriptions"
+	"github.com/azure/azure-dev/cli/azd/pkg/azdext"
+	"github.com/azure/azure-dev/cli/azd/pkg/graphsdk"
+	"github.com/google/uuid"
+)
+
+type FoundryParser struct {
+	AzdClient *azdext.AzdClient
+}
```
wbreza, line 35: Nit: This file seems a bit overloaded and isn't just a parser.

cli/azd/extensions/azure.foundry.ai.agents/internal/cmd/listen.go
```diff
@@ -28,8 +28,11 @@ func newListenCommand() *cobra.Command {
 			defer azdClient.Close()
 
 			provider := project.NewAgentServiceTargetProvider(azdClient)
+			projectParser := &project.FoundryParser{AzdClient: azdClient}
 			host := azdext.NewExtensionHost(azdClient).
-				WithServiceTarget("foundry.agent", provider)
+				WithServiceTarget("foundry.agent", provider).
+				WithProjectEventHandler("preprovision", projectParser.SetIdentity).
+				WithProjectEventHandler("postdeploy", projectParser.CoboPostDeploy)
```
wbreza, line 36: I'd consider adding these as a service level event so we can handle these incrementally during deployment so that a transient failure on some other service does not stop the overall postdeploy from being invoked.

cli/azd/extensions/azure.foundry.ai.agents/internal/project/parser.go
```diff
@@ -0,0 +1,1291 @@
...
+	projectClient := p.AzdClient.Project()
+	projectResponse, err := projectClient.Get(ctx, &azdext.EmptyRequest{})
+	if err != nil {
+		return false, fmt.Errorf("failed to get project: %w", err)
+	}
+	project := projectResponse.Project
+
+	projectPath := project.Path
+	for _, service := range project.Services {
+		if service.Host == "containerapp" {
```
wbreza, line 41: The service level registration would also avoid this check since you can filter to only bind to specific host types.

cli/azd/extensions/azure.foundry.ai.agents/internal/project/parser.go
```diff
@@ -0,0 +1,1291 @@
...
+				return agent.Agent.Kind == agent_yaml.AgentKindYamlContainerApp, nil
+			}
+		}
+	}
+
+	return false, nil
+}
+
+func (p *FoundryParser) SetIdentity(ctx context.Context, args *azdext.ProjectEventArgs) error {
+	shouldRun, err := p.shouldAttach(ctx)
```
wbreza: The project event args already contains the project config... you shouldn't need to call a service again to get the project configuration.

cli/azd/extensions/azure.foundry.ai.agents/internal/project/parser.go
vhvb1989: I am not defining the file structure myself...  I am mostly focusing on making things work by migrating the current hooks into go-code.
After that, extension-owner is moving things and giving the expected file structure they want.

cli/azd/extensions/azure.foundry.ai.agents/internal/cmd/listen.go
vhvb1989: I am using the original hook which was a project level hook for postdeploy.  It should be easy to switch in the future if needed. multi-service is actually not part of the spec

cli/azd/extensions/azure.foundry.ai.agents/internal/project/parser.go
vhvb1989: Yeah, but it is trigger once per service, which is not expected for this Event to be the case. 
Even if there are more containerapp services, the event is currently designed to run just once

cli/azd/extensions/azure.foundry.ai.agents/internal/project/parser.go
vhvb1989: That's how I had it initially and then switched to this trying to use `shouldAttach()` from the listen cmd to decide when actually adding the event handler or not.  I'll revert it to how it was before :)

cli/azd/extensions/azure.foundry.ai.agents/internal/project/parser.go
guming-learning: @wbreza are you using the remote docker build for "regular ACA", since only in that way you can get an image before provision phase.

cli/azd/extensions/azure.foundry.ai.agents/internal/project/parser.go
guming-learning: Yes it sets new revision. But old ones won't get deactivated automatically with it, and they shouldn't since the design is to keep multiple revisions active and reached by specifying different suffix corresponding to agent versions.

PR #5981, opened 2025-10-22
allow folks to set env var to write progress logs to file

vhvb1989:
Successfully implemented debug logging support for the progressLog component. Here's what was added:

Key Changes:
Environment Variable Support: When `AZD_DEBUG_PROGRESS_LOGS` is set, the component now writes all logs to a file.

New Struct Fields:

```
debugLogFile *os.File: File handle for debug logging
debugLogBuffer strings.Builder: Buffer to accumulate all log content
File Naming Strategy:
```

Files are created with a sanitized version of the title
Timestamp format: YYYYMMDD-HHMMSS
Example filename: Deployment_Status-20251022-214104.log
Files are created in the current working directory
Updated Methods:

Start(): Checks for `AZD_DEBUG_PROGRESS_LOGS` env var and initializes debug logging, writes header and title to the buffer
Write(): Writes log data to both the screen and the debug buffer
Header(): Updates header in both display and debug buffer
Stop(): Writes the complete buffer to the file and closes it
Helper Functions:

initDebugLogFile(): Creates the debug log file with unique timestamp
sanitizeFilename(): Cleans the title to create a safe filename (removes invalid characters, limits length, replaces spaces with underscores)
Test Coverage: Added Test_progressLogDebugFile to verify:

File is created when env var is set
Filename contains sanitized title and timestamp
File content includes all logs, header, and title

Example from foundry extension writing logs to file for debugging purposes:
image

fix: https://github.com/Azure/azure-dev/issues/5982

### DIFF
--- a/cli/azd/pkg/input/progress_log.go
+++ b/cli/azd/pkg/input/progress_log.go
@@ -5,9 +5,14 @@ package input
 
 import (
 	"bufio"
+	"fmt"
 	"log"
+	"os"
+	"path/filepath"
+	"regexp"
 	"strings"
 	"sync"
+	"time"
 
 	"github.com/adam-lavrik/go-imath/ix"
 	tm "github.com/buger/goterm"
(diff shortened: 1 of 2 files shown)


cli/azd/pkg/input/progress_log.go
```diff
@@ -106,13 +123,25 @@ func (p *progressLog) Start() {
 		tm.Println(p.header)
 		// margin after title
 		tm.Println("")
+
+		// Write header to debug log file
+		if p.debugLogFile != nil {
+			p.debugLogBuffer.WriteString(p.header)
+			p.debugLogBuffer.WriteString("\n\n")
```
hemarina, line 130: Just curious why adding "\n\n" in the next line rather than in L129?

vhvb1989:
I'll keep this for me only.
Talked with the team and we might want a more general logging solution than implementing this by component.

Issue #5972, opened 2025-10-22
[Aspire] Add support for `annotated.string`


vhvb1989:
Related to: https://github.com/dotnet/aspire/pull/12216

PR #5990, opened 2025-10-23
[Aspire] handle annotated strings with uri format

vhvb1989:
Annotated strings are used within the manifest to represent the usage intention for a parameter.
When using uri filter, azd must encode the value as URI.

For example, for a value like `pFbl=TJGk@L9S@qh918zOk`  annotated with filter=uri, azd applies `url.QueryEscape` to the value, generating something like `pFbl%3DTJGk%40L9S%40qh918zOk`

fix: https://github.com/Azure/azure-dev/issues/5972

### DIFF
--- a/cli/azd/pkg/apphost/generate.go
+++ b/cli/azd/pkg/apphost/generate.go
@@ -510,13 +510,19 @@ func GenerateProjectArtifacts(
 	return files, nil
 }
 
+type annotatedString struct {
+	Filter string
+	Value  string
+}
+
 type infraGenerator struct {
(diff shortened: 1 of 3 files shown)


sebastienros:
It looks good. I don't know how your eng process works but I would suggest a test that uses a template where both the standard and filtered values are used in outputs. The redis template should already do that: `PASSWORD` is not encoded and `URI` is partially.

sebastienros (approved):

weikanglim (approved):

Issue #6000, opened 2025-10-24
Bug: progressLog.Write() panic with 'index out of range [-1]'


vhvb1989:
## Description
The `progressLog.Write()` method in `pkg/input/progress_log.go` can panic with a runtime error `index out of range [-1]` when concurrent access or edge cases result in an empty output buffer.

## Stack Trace
```
panic: runtime error: index out of range [-1]

goroutine 33 [running]:
github.com/azure/azure-dev/cli/azd/pkg/input.(*progressLog).Write(0xc000c12b00, {0xc0007be000, 0x15, 0xc000c87d48?})
        github.com/azure/azure-dev/cli/azd/pkg/input/progress_log.go:179 +0x82d
github.com/azure/azure-dev/cli/azd/pkg/input.(*consolePreviewerWriter).Write(0x17ff77be58920?, {0xc0007be000?, 0x15?, 0x0?})
        github.com/azure/azure-dev/cli/azd/pkg/input/console_previewer_writer.go:23 +0x75
```

## Root Causes
1. **Race Condition**: The original code checks `if p.output == nil` before acquiring the mutex, allowing other goroutines to modify `p.output` between the check and lock acquisition
2. **Empty Slice Access**: When `len(p.output) == 0`, accessing `p.output[len(p.output)-1]` tries to access index -1
3. **Zero Lines Edge Case**: When `p.lines == 0`, `make([]string, p.lines)` creates an empty slice
4. **Unsafe Slice Operations**: Operations like `p.output = append(p.output[1:], p.prefix)` can create empty slices

## Impact
- Application crashes during log output operations
- Affects concurrent usage scenarios
- Edge cases with zero line configurations fail

## Expected Behavior
The `progressLog.Write()` method should handle all edge cases gracefully without panicking, even under concurrent access or with zero/minimal line configurations.

PR #6001, opened 2025-10-24
fix: prevent index out of range panic in progressLog.Write()

vhvb1989:
## Description
This PR fixes a critical runtime panic in `progressLog.Write()` that occurs when concurrent access or edge cases result in an empty output buffer.

Fixes Azure/azure-dev#6000

## Root Cause Analysis
The panic `runtime error: index out of range [-1]` was caused by:

1. **Race Condition**: Checking `if p.output == nil` before acquiring mutex allowed state changes between check and lock
2. **Empty Slice Access**: When `len(p.output) == 0`, accessing `p.output[len(p.output)-1]` tries to access index -1
3. **Zero Lines Edge Case**: When `p.lines == 0`, `make([]string, p.lines)` creates empty slice
4. **Unsafe Slice Operations**: Operations like `p.output = append(p.output[1:], p.prefix)` could create empty slices

### Thread Safety Improvements
- Moved mutex acquisition to beginning of `Write()` method before all state checks
- All output buffer access now happens under mutex protection

### Safety Checks Added
- Multiple safety checks ensure output buffer always has at least one element
- Handle edge case where `p.lines == 0` by ensuring minimum 1 element in buffer
- Added bounds checking before all array access operations
- Protected slice operations to prevent empty slice creation

### Key Code Changes
```go
// Before: Race condition possible
if p.output == nil {
    return len(logBytes), nil
}
// ... other code ...
p.outputMutex.Lock()

// After: Thread-safe
p.outputMutex.Lock()
defer p.outputMutex.Unlock()
if p.output == nil {
    return len(logBytes), nil
}
```

```go
// Before: Could panic with index -1
lastIndex := len(p.output) - 1
if p.output[lastIndex] == "" {

// After: Bounds checking
lastIndex := len(p.output) - 1
if lastIndex >= 0 && p.output[lastIndex] == "" {
```

## Testing
- Added `Test_progressLogConcurrentWriteProtection`: Tests concurrent writes and stop/start cycles
- Added `Test_progressLogEmptySliceProtection`: Tests zero lines config and forced empty slices
- All existing tests pass
- No regressions introduced

## Validation
✅ All tests pass  
✅ Code builds successfully  
✅ No linting issues (0 golangci-lint issues)  
✅ Thread-safe concurrent access  
✅ Edge cases with zero/minimal configurations handled  

## Impact
- Prevents application crashes during log operations
- Ensures reliable operation under concurrent access
- Handles all edge cases gracefully without panicking

### DIFF
--- a/cli/azd/pkg/input/progress_log.go
+++ b/cli/azd/pkg/input/progress_log.go
@@ -158,14 +158,7 @@ func (p *progressLog) Write(logBytes []byte) (int, error) {
 	}
 
 	// Safety check: ensure output buffer has at least one element
-	if len(p.output) == 0 {
-		// Ensure we have at least 1 line even if p.lines is 0
-		lineCount := p.lines
-		if lineCount == 0 {
-			lineCount = 1
-		}
-		p.output = make([]string, lineCount)
-	}
+	p.ensureOutputBuffer()
 
 	maxWidth := p.terminalWidthFn()
 	if maxWidth <= 0 {
@@ -190,14 +183,7 @@ func (p *progressLog) Write(logBytes []byte) (int, error) {
 		}
 
 		// Safety check: ensure we have at least one element after slice operations
-		if len(p.output) == 0 {
-			// Ensure we have at least 1 line even if p.lines is 0
-			lineCount := p.lines
-			if lineCount == 0 {
-				lineCount = 1
-			}
-			p.output = make([]string, lineCount)
-		}
+		p.ensureOutputBuffer()
 
 		fullLog := log
 		lastIndex := len(p.output) - 1
@@ -208,14 +194,7 @@ func (p *progressLog) Write(logBytes []byte) (int, error) {
 
 		for fullLogLen > 0 {
 			// Safety check before accessing last element
-			if len(p.output) == 0 {
-				// Ensure we have at least 1 line even if p.lines is 0
-				lineCount := p.lines
-				if lineCount == 0 {
-					lineCount = 1
-				}
-				p.output = make([]string, lineCount)
-			}
+			p.ensureOutputBuffer()
 			lastIndex = len(p.output) - 1
 
 			// Get whatever is the empty space on current line
@@ -279,6 +258,19 @@ func (p *progressLog) Header(header string) {
 }
 
 /****************** Not exported method ****************/
+
+// ensureOutputBuffer ensures the output buffer has at least one element.
+// This prevents index out of range panics when accessing p.output[len(p.output)-1].
+func (p *progressLog) ensureOutputBuffer() {
+	if len(p.output) == 0 {
+		// Ensure we have at least 1 line even if p.lines is 0
+		lineCount := p.lines
+		if lineCount == 0 {
+			lineCount = 1
+		}
+		p.output = make([]string, lineCount)
+	}
+}
 
 // clearLine override text with empty spaces.
 func clearLine() {


cli/azd/pkg/input/progress_log.go
```diff
@@ -175,25 +189,48 @@ func (p *progressLog) Write(logBytes []byte) (int, error) {
...
 
+		// Safety check: ensure we have at least one element after slice operations
+		if len(p.output) == 0 {
+			// Ensure we have at least 1 line even if p.lines is 0
+			lineCount := p.lines
+			if lineCount == 0 {
+				lineCount = 1
+			}
+			p.output = make([]string, lineCount)
+		}
```
wbreza: Can we DRY this up a bit? Looks like we're duplicating this block a few times...

cli/azd/pkg/input/progress_log.go
```diff
@@ -148,9 +148,25 @@ func (p *progressLog) Stop(keepLogs bool) {
 // Write implements oi.Writer and updates the internal buffer before flushing it into the screen.
 // Calling Write() before Start() or after Stop() is a no-op
 func (p *progressLog) Write(logBytes []byte) (int, error) {
+	// Acquire mutex first to prevent race conditions
+	p.outputMutex.Lock()
+	defer p.outputMutex.Unlock()
```
weikanglim, line 153: Could you please help me understand how this is happening? I'm concerned about there being concurrent access of progress log in the first place. I don't understand what this means for the actual UI presented to the user.

cli/azd/pkg/input/progress_log_test.go
```diff
@@ -330,3 +330,77 @@ func Test_progressChangeHeader(t *testing.T) {
...
+	<-done
+	pg.Stop(false)
+}
+
+func Test_progressLogEmptySliceProtection(t *testing.T) {
+	sizeFn := func() int {
+		return 40
+	}
+	// Create with 0 lines to test edge case
+	pg := newProgressLogWithWidthFn(0, prefix, title, header, sizeFn)
```
weikanglim, line 382: Should this be a programmatic error to create a progress log with 0 lines?

rajeshkamal5050 (approved):

cli/azd/pkg/input/progress_log.go
vhvb1989: The Extension Event Flow
1. Extension Registration and Event Handling
When an azd extension registers a postdeploy event, here's what happens:

```go
// Extension registers postdeploy hook
extension.RegisterEvent("postdeploy", extensionCommand)

// During azd up/deploy, after main deployment:
eventService.Emit("postdeploy", eventData)
```
2. Event Service Execution
The event service does:

```go
// Executes extension command in separate goroutine
go func() {
    cmd := exec.Command(extensionPath, args...)
    
    // Sets up IO pipes - THIS IS KEY
    cmd.Stdout = multiWriter  // Multiple destinations including progressLog
    cmd.Stderr = multiWriter
    
    // Starts streaming immediately
    cmd.Start()
}()
```

3. The Critical Multi-Writer Setup
The event service creates an io.MultiWriter that sends extension output to:

progressLog (for the UI)
File logger (for debugging)
Event stream (for gRPC clients)
Possibly metrics collection

```
multiWriter := io.MultiWriter(
    progressLog,           // UI display
    fileLogger,           // Debug logs
    eventStreamWriter,    // gRPC stream
)
```

# How the Panic Occurred During Extension Streaming

The Exact Sequence
Extension Starts Outputting

```
# Extension command runs and outputs:
echo "Starting post-deployment validation..."
echo "Checking service health..."
echo "Running integration tests..."
```
- Concurrent Streaming Begins

```go
// Multiple goroutines are created by io.Copy:

// Goroutine 1: stdout → multiWriter
go io.Copy(multiWriter, cmd.Stdout)

// Goroutine 2: stderr → multiWriter  
go io.Copy(multiWriter, cmd.Stderr)

// Goroutine 3: Event service cleanup
go func() {
    cmd.Wait()
    // Clean up progress log when done
    progressLog.Stop(false)
}()
```
The Race Condition Window

```go
// Goroutine 1: Extension outputs text
progressLog.Write([]byte("Checking service health...\n"))
// ↑ Gets past nil check, but before acquiring mutex

// Goroutine 3: Extension finishes quickly
progressLog.Stop(false)  // Sets p.output = nil

// Goroutine 1: Continues with stale reference
// ↓ PANIC: tries to access p.output[len(p.output)-1] where len() = -1
```

Why Extensions Made This Worse
Rapid Output Bursts: Extensions often output multiple lines quickly
Short-lived Processes: Extension commands finish faster than the main azd operations
Multiple Cleanup Points: Both the main deployment AND the extension can trigger Stop()
Buffered IO: Extension output might be buffered and flushed in chunks
The Specific Stack Trace Mapping
Looking at the original stack trace in the context of extension execution:

```
goroutine 33 [running]:
github.com/azure/azure-dev/cli/azd/pkg/input.(*progressLog).Write()
    ↑ Extension output being written to progress log

github.com/azure/azure-dev/cli/azd/pkg/output.(*DynamicMultiWriter).Write()
    ↑ Event service's multi-writer distributing extension output

io.(*multiWriter).Write() / io.copyBuffer() 
    ↑ Go's IO system copying from extension's stdout/stderr

os/exec.(*Cmd).writerDescriptor.func1()
os/exec.(*Cmd).Start.func2()
    ↑ Extension command execution goroutines
```

Extension-Specific Edge Cases
1. Fast Extension Completion

```
// Extension that exits immediately:
#!/bin/bash
echo "Post-deploy complete"
exit 0
# Duration: ~10ms, but output still being processed
```

2. Buffered Output Flush

```
// Extension outputs multiple lines:
fmt.Println("Line 1")
fmt.Println("Line 2") 
fmt.Println("Line 3")
// OS buffers these, then flushes as one Write() call
// But extension process already exited
```

3. Multiple Event Handlers
```
// Multiple extensions for same event:
extension1.RegisterEvent("postdeploy", cmd1)  // Starts goroutine 1
extension2.RegisterEvent("postdeploy", cmd2)  // Starts goroutine 2
// Both writing to same progressLog concurrently
```

How the Fix Solves Extension Issues
Before Fix (Panic-Prone)
```go
func (p *progressLog) Write(logBytes []byte) (int, error) {
    if p.output == nil {  // ← Extension goroutine checks
        return len(logBytes), nil
    }
    // Extension finishes, main thread calls Stop()
    // p.output becomes nil here ↓
    
    p.outputMutex.Lock()  // ← Too late, stale reference
    // PANIC: p.output[len(p.output)-1] with len = -1
}
```

After Fix (Safe)

```go
func (p *progressLog) Write(logBytes []byte) (int, error) {
    p.outputMutex.Lock()         // ← Acquire lock FIRST
    defer p.outputMutex.Unlock()
    
    if p.output == nil {         // ← Check state under lock
        return len(logBytes), nil  // Safe return
    }
    
    p.ensureOutputBuffer()       // ← Guarantee valid state
    // All access now safe, even with rapid extension completion
}
```

The fix ensures that extension output streams safely to the progress log even when extensions complete rapidly or multiple extensions run concurrently, providing users with complete visibility into the deployment process without crashes.

Issue #5872, opened 2025-10-07
Extensions dashboard


kristenwomack:
[Spec details](https://microsoft.sharepoint.com/:fl:/s/1e32fb7a-ff2a-4c7d-ad2a-b046eef1247d/ESVKwSz5MV9OkpVQ_b_zzeUBw79JfuxIYb1L5CzPYs9JGA?e=3Ehmvw&nav=cz0lMkZzaXRlcyUyRjFlMzJmYjdhLWZmMmEtNGM3ZC1hZDJhLWIwNDZlZWYxMjQ3ZCZkPWIlMjFJSHBrNVhrem9rNk5CMGVUdFJmZ28tbG5DcGc2WXlsTGs5VjlQWWZ4RnliRmxLWG02dWpGU0pva3Rqak4yWmQ2JmY9MDFUTUlNWjNSRkpMQVNaNkpSTDVISkZGS1E3Vzc3SFRQRiZjPSUyRiZhPUxvb3BBcHAmcD0lNDBmbHVpZHglMkZsb29wLXBhZ2UtY29udGFpbmVyJng9JTdCJTIydyUyMiUzQSUyMlQwUlRVSHh0YVdOeWIzTnZablF1YzJoaGNtVndiMmx1ZEM1amIyMThZaUZKU0hCck5WaHJlbTlyTms1Q01HVlVkRkptWjI4dGJHNURjR2MyV1hsc1RHczVWamxRV1daNFJubGlSbXhMV0cwMmRXcEdVMHB2YTNScWFrNHlXbVEyZkRBeFZFMUpUVm96VVV3MVdFRk1VekpDTWtOR1FrdExSRFV6VjBsUldWaERNa1klM0QlMjIlMkMlMjJpJTIyJTNBJTIyMzM4Mjg1MzYtMTVkMi00NWU5LWFjNDktZTg2ZmY5NzkxNTM1JTIyJTdE)

rajeshkamal5050:
@kristenwomack @AngelosP can we close this one? Anything else pending on this?

rajeshkamal5050:
Closing. This has been open for 6 months with no implementation work. The spec is in SharePoint if this needs to be revisited.

If the dashboard work is being planned, please reopen with updated scope.

kristenwomack:
This is exactly what I was asking about a few days ago in stand up. We still need telemetry on aspects of this. I'll clarify in the spec.

PR #6010, opened 2025-10-27
tracing: add usage attributes for extension installs and runs

### LINKED ISSUE #5937: extensions: telemetry for failing cases may not be reported
I'm not seeing traces for failed extension commands show up in our logs.

weikanglim:
Add usage events and attributes for extension installs (`ext.install`) and runs (`ext.run`).

Expose both `Annotations` and `DisableTroubleshooting` in the action descriptor/options since that was needed.

Fixes #5937 
Contributes towards #5872

### DIFF
--- a/cli/azd/cmd/root.go
+++ b/cli/azd/cmd/root.go
@@ -379,7 +379,9 @@ func NewRootCmd(
 	root.
 		UseMiddleware("debug", middleware.NewDebugMiddleware).
 		UseMiddleware("ux", middleware.NewUxMiddleware).
-		UseMiddleware("error", middleware.NewErrorMiddleware).
+		UseMiddlewareWhen("error", middleware.NewErrorMiddleware, func(descriptor *actions.ActionDescriptor) bool {
+			return !descriptor.Options.DisableTroubleshooting
+		}).
 		UseMiddlewareWhen("telemetry", middleware.NewTelemetryMiddleware, func(descriptor *actions.ActionDescriptor) bool {
 			return !descriptor.Options.DisableTelemetry
 		}).
--- a/cli/azd/cmd/actions/action_descriptor.go
+++ b/cli/azd/cmd/actions/action_descriptor.go
@@ -187,6 +187,8 @@ type ActionDescriptorOptions struct {
 	DefaultFormat output.Format
 	// Whether or not telemetry should be disabled for the current action
 	DisableTelemetry bool
+	// Whether or not troubleshooting should be disabled for the current action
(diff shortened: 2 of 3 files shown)


rajeshkamal5050 (approved):

cli/azd/cmd/root.go
```diff
@@ -379,7 +379,9 @@ func NewRootCmd(
 	root.
 		UseMiddleware("debug", middleware.NewDebugMiddleware).
 		UseMiddleware("ux", middleware.NewUxMiddleware).
-		UseMiddleware("error", middleware.NewErrorMiddleware).
+		UseMiddlewareWhen("error", middleware.NewErrorMiddleware, func(descriptor *actions.ActionDescriptor) bool {
+			return !descriptor.Options.DisableTroubleshooting
```
wbreza, line 383: I really like this because sometimes the error troubleshooting just doesn't make sense for "all" errors :)

cli/azd/pkg/extensions/manager.go
```diff
@@ -489,6 +497,10 @@ func (m *Manager) Install(
 		return nil, fmt.Errorf("failed to save user config: %w", err)
 	}
 
+	span.SetAttributes(
+		fields.ExtensionId.String(extension.Id),
+		fields.ExtensionVersion.String(selectedVersion.Version))
```
wbreza, line 502: Should we consider capturing the extension source name?

vhvb1989 (approved):

cli/azd/cmd/root.go
weikanglim: I get that sentiment. I kept it consistent with what we already had for telemetry.

However, briefly noting that this isn't my personal preference. If you squint closely at how the error middleware is constructed, we already have a list of errors that we're filtering out. I'd personally would be much in favor of using the command path/metadata as a filtering, with a single source of truth.

cli/azd/pkg/extensions/manager.go
weikanglim: Thanks for the suggestion. I looked into this briefly, but left it as-is for now. Feel free to log an issue if we'd like to pursue further.

I did note the following that left me undecided:
- If we were to do this, we would have to hash the extension source name, since we wouldn't a private registry like `mySecretCompanyProject` to show up in our telemetry.
- Extension source name is a great approximation, however it isn't always accurate? A user could have easily renamed `azd2` to point to the same official registry. It seems that using a more natural key like `Location` would be ideal, but I didn't quite find where to do this.

PR #6013, opened 2025-10-28
Adds support for custom ServiceConfig settings for extensions

wbreza:
### Overview
This PR adds support for the new `Config` field in `azdext.ServiceConfig`, enabling service targets to use custom configuration while maintaining backward compatibility.

### Benefits
- Enables custom configuration for service targets through extensions
- Maintains backward compatibility with existing configurations
- Supports complex nested configuration structures

This enhancement supports the broader extensibility framework allowing `azd` extensions to provide custom service targets with their own configuration requirements.

## Example `azure.yaml`
```yaml
name: ai-foundry-starter-basic
requiredVersions:
    extensions:
        azure.foundry.ai.agents: ">=0.0.1"
services:
    echo-agent:
        project: src/echo-agent
        host: foundry.containeragent
        language: python
        config:
            projectEndpoint: "https://ai.azure.com/foundry/project"
            env:
                foo: bar
                baz: ${QUX}
            scale:
                minReplicas: 1
                maxReplicas: 3

```

### DIFF
--- a/cli/azd/extensions/azure.ai.agents/internal/project/service_target_agent.go
+++ b/cli/azd/extensions/azure.ai.agents/internal/project/service_target_agent.go
@@ -578,47 +578,9 @@ func (p *AgentServiceTargetProvider) startAgentContainer(
 	const maxWaitTime = 10 * time.Minute
 	const apiVersion = "2025-05-15-preview"
 
-	// Extract replica configuration from agent manifest
-	minReplicas := int32(1) // Default values
-	maxReplicas := int32(1)
-
-	// Check if the agent definition has scale configuration
-	if containerAgent, ok := interface{}(agentManifest.Template).(agent_yaml.ContainerAgent); ok {
-		// For ContainerAgent, check if Options contains scale information
-		if containerAgent.Options != nil {
-			if options, exists := (*containerAgent.Options)["scale"]; exists {
-				if scaleMap, ok := options.(map[string]interface{}); ok {
-					if minReplicasFloat, exists := scaleMap["minReplicas"]; exists {
-						if minReplicasVal, ok := minReplicasFloat.(float64); ok {
-							minReplicas = int32(minReplicasVal)
-						}
-					}
-					if maxReplicasFloat, exists := scaleMap["maxReplicas"]; exists {
-						if maxReplicasVal, ok := maxReplicasFloat.(float64); ok {
-							maxReplicas = int32(maxReplicasVal)
-						}
-					}
-				}
-			}
-		}
-	}
-
-	// Validate replica counts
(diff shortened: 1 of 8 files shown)


cli/azd/extensions/azure.ai.agents/internal/project/service_target_agent.go
```diff
@@ -232,6 +232,13 @@ func (p *AgentServiceTargetProvider) Deploy(
 		azdEnv[kval.Key] = kval.Value
 	}
 
+	var serviceTargetConfig *FoundryAgentConfig
+	if err := UnmarshalStruct(serviceConfig.Config, &serviceTargetConfig); err != nil {
```
wbreza, line 237: This is an example of the usage pattern to unmarshall custom configuration during a deploy operation.

cli/azd/extensions/azure.ai.agents/internal/project/config.go
```diff
@@ -0,0 +1,36 @@
...
+import (
+	"encoding/json"
+
+	"google.golang.org/protobuf/encoding/protojson"
+	"google.golang.org/protobuf/types/known/structpb"
+)
+
+// FoundryAgentConfig provides custom configuration for the Foundry AI Service target
+type FoundryAgentConfig struct {
+	FoundryProjectEndpoint string            `json:"projectEndpoint,omitempty"`
```
jfomhover: can we remove this from the config?

vhvb1989 (approved):
I am worried about who's responsible for writing the service-config....

Is it the extension `init` command?
Is it a template-owner who knows about this config?
Is it the end-user who needs to know where to find and update this config?

Part of me feels like extensions should not have write access to azure.yaml (only read)... and if azd should protect the project somehow.

cli/azd/extensions/azure.ai.agents/internal/project/config.go
wbreza: Whoever is consuming this can remove it after this PR is merged.

wbreza:
> I am worried about who's responsible for writing the service-config....
> 
> Is it the extension `init` command? Is it a template-owner who knows about this config? Is it the end-user who needs to know where to find and update this config?

The extension author will be required to maintain and support the `config` for their custom service targets. From the `azd` perspective this is just a generic property bag of key/value pairs.  The extension will be responsible for defining the shape of the configuration, handle versioning of the config and future updates.

Right now part of the `azd ai agent init` flow will write back to the `azure.yaml` to add new services complete with any configuration data.

> 
> Part of me feels like extensions should not have write access to azure.yaml (only read)... and if azd should protect the project somehow.

Would love to know more about this and why we should limit write access?

PR #6021, opened 2025-10-28
[Agent Extension] Add support for getting model details during init

trangevi:


cli/azd/extensions/azure.ai.agents/internal/cmd/init.go
```diff
@@ -1303,16 +1322,49 @@ func (a *InitAction) updateEnvironment(ctx context.Context, agentManifest *agent
...
+			}
+			deploymentDetails = append(deploymentDetails, *modelDeployment)
+		}
+	}
+
+	deploymentsJson, err := json.Marshal(deploymentDetails)
+	if err != nil {
+		return fmt.Errorf("failed to marshal deployment details to JSON: %w", err)
+	}
+	if err := a.setEnvVar(ctx, envName, "AI_PROJECT_DEPLOYMENTS", string(deploymentsJson)); err != nil {
```
JeffreyCA, line 1363: What's the difference between `AI_PROJECT_DEPLOYMENTS` and `AZURE_AI_FOUNDRY_MODEL_NAME`? Are both needed?

jfomhover: AZURE_AI_FOUNDRY_MODEL_NAME should not be needed anymore and we should deprecate it, since it works only for a single-model scenario, and is not used in the template anyway

vhvb1989:
/
/check-enforcer override

PR #6028, opened 2025-10-29
Fixes lifetime issues with container helper from extension container service

### LINKED ISSUE #6019: Custom service targets - `azd up` can fail during deploy when using ContainerService's Publish
`azd up` can fail during deploy (publish step) due to `AZURE_CONTAINER_REGISTRY_ENDPOINT` supposedly not being set, despite it being set in .env after provisioning:

Image

Through the debugger, you can see how the env only has 3 entries at that point:
Image

wbreza:
Resolves #6019 - This change updates how the `azd` environment is resolved from within the container helper.  Instead of relying on the `environment` field of the `ContainerHelper` we refactor to take a dependency of the environment manager and then fetch the environment before usage.

The original issue was caused by the container helper being loaded within a singleton container service for the extension framework. Since this service is long lived throughout the full `azd` invocation this was causing issues in workflow based scenarios where a stale or cached `azd` environment was being referenced instead of the correct scoped instance.

This ensures that the `azd` environment we are working against is always an up to date version that includes all variables and configuration.

### DIFF
--- a/cli/azd/go.work
+++ b/cli/azd/go.work
@@ -2,6 +2,6 @@ go 1.25.0
 
 use (
 	../../
-	./extensions/azure.coding-agent
 	./extensions/azure.ai.agents
+	./extensions/azure.coding-agent
 )


cli/azd/internal/cmd/deploy.go
```diff
@@ -288,14 +288,9 @@ func (da *DeployAction) Run(ctx context.Context) (*actions.ActionResult, error)
 					da.console.StopSpinner(ctx, stepMessage, input.StepFailed)
 					return err
 				}
-
-				// Append package artifacts
-				if err := serviceContext.Package.Add(packageResult.Artifacts...); err != nil {
-					return err
-				}
```
wbreza, line 295: The `ServiceManager` is now fully responsible for appending any newly generated artifacts from the service level operations.

cli/azd/pkg/project/container_helper.go
```diff
@@ -86,12 +87,36 @@ func NewContainerHelper(
 	}
 }
 
+// getCurrentEnvironment gets the current environment using the standard pattern
+func (ch *ContainerHelper) getCurrentEnvironment(ctx context.Context) (*environment.Environment, error) {
+	defaultEnvironment, err := ch.azdContext.GetDefaultEnvironmentName()
```
JeffreyCA, line 92: Do we need to handle cases where for example `--environment` flag is passed to override the default env?

vhvb1989 (approved):

rajeshkamal5050 (approved):

hemarina (dismissed):

cli/azd/pkg/project/container_helper.go
wbreza: Let me review and we can take that as a follow-up fix if needed.

vhvb1989 (approved):

PR #6029, opened 2025-10-29
[AI Agent Extension] Add host flag

trangevi:


cli/azd/extensions/azure.ai.agents/internal/cmd/init.go
```diff
@@ -144,6 +145,9 @@ func newInitCommand() *cobra.Command {
 	cmd.Flags().StringVarP(&flags.src, "src", "s", "",
 		"[Optional] Directory to download the agent yaml to (defaults to 'src/<agent-id>')")
 
+	cmd.Flags().StringVarP(&flags.host, "host", "", "",
```
wbreza, line 150: Do you want to include the default value here? or will you apply it later when you find this is empty?

cli/azd/extensions/azure.ai.agents/internal/cmd/init.go
```diff
@@ -769,22 +774,19 @@ func (a *InitAction) addToProject(ctx context.Context, targetDir string, agentMa
...
-		host = "azure.ai.agents"
+	var serviceHost string
+
+	switch host {
+	case "containerapp":
+		serviceHost = "containerapp"
 	default:
-		// except here
-		return fmt.Errorf("unsupported agent kind: %s", agentDef.Kind)
+		serviceHost = "azure.ai.agents"
```
wbreza: Can we make this a constant? Feels like we may be referencing this string in multiple places.  I also though we were going to use the singular form here since all other `azd` host types are singular form.  I'd recommend `azure.ai.agent`.

cli/azd/extensions/azure.ai.agents/internal/pkg/agents/agent_yaml/yaml.go
```diff
@@ -7,18 +7,15 @@ package agent_yaml
...
+	AgentKindPrompt   AgentKind = "prompt"
+	AgentKindHosted   AgentKind = "hosted"
+	AgentKindWorkflow AgentKind = "workflow"
 )
 
 // IsValidAgentKind checks if the provided AgentKind is valid
 func IsValidAgentKind(kind AgentKind) bool {
 	switch kind {
-	case AgentKindPrompt, AgentKindHosted, AgentKindContainerApp, AgentKindWorkflow, AgentKindYamlContainerApp:
+	case AgentKindPrompt, AgentKindHosted, AgentKindWorkflow:
```
wbreza: nit: Since you already have `ValidAgentKinds()` that retuns the slice of valid kinds you may just want to resuse that here using a `slices.Contains(...)` call.

wbreza (approved):
Looks good - just left a couple minor comments.

cli/azd/extensions/azure.ai.agents/internal/cmd/init.go
trangevi: I currently apply it later when I find it empty, partially because "azure.ai.agent" is a bit of a weird host in this context I feel like. If we want to agree on a default string that maybe means a bit more to a user, I'm up for that, but I think for now it's easier to just default to empty.

PR #6042, opened 2025-10-30
[AI Agents Extension] Add models to service config

trangevi:


cli/azd/extensions/azure.ai.agents/internal/cmd/listen.go
```diff
@@ -77,3 +63,29 @@ func newListenCommand() *cobra.Command {
...
+	deploymentsJson, err := json.Marshal(foundryAgentConfig.Deployments)
+	if err != nil {
+		return fmt.Errorf("failed to marshal deployment details to JSON: %w", err)
+	}
+
+	azdClient.Environment().SetValue(ctx, &azdext.SetEnvRequest{
+		EnvName: currentEnvResponse.Environment.Name,
+		Key:     "AI_PROJECT_DEPLOYMENTS",
+		Value:   string(deploymentsJson),
+	})
```
JeffreyCA: Can we make sure `ENABLE_HOSTED_AGENTS`/`ENABLE_CONTAINER_AGENTS` is set accordingly too during preprovision? Were those are supposed to go under the service config as well?

trangevi: Those are currently being set still during init. I'm doing things in pieces anyway, will be adding more to this in future PRs, so I think even if those need to be moved then we can do it in a subsequent PR.

Or do you mean just checking to make sure they are set correctly during preprovision?

JeffreyCA: Just wanted to make sure this flow works as expected:

1. `azd init -t <template>`
2. `azd ai agent init -m <...>`
3. `azd up`
4. `azd env new env-2`
5. `azd up`

With these current preprovision changes, only `AI_PROJECT_DEPLOYMENTS` would get set when running `azd up` the second time (step 5), but we'll need to also set `ENABLE_HOSTED_AGENTS` to `true` for hosted agents or `ENABLE_CONTAINER_AGENTS` to `true` for container agents.

If you were planning to do this in a subsequent PR then that's fine, or I can make those changes too.

cli/azd/extensions/azure.ai.agents/internal/cmd/listen.go
wbreza: Let's make sure we update this to whatever is the new host/service target constant string we are using.

cli/azd/extensions/azure.ai.agents/internal/project/config.go
```diff
@@ -14,6 +15,7 @@ import (
 type ServiceTargetAgentConfig struct {
 	Environment map[string]string `json:"env,omitempty"`
 	Scale       *ScaleSettings    `json:"scale,omitempty"`
+	Deployments []Deployment      `json:"deployments,omitempty"`
```
wbreza, line 18: You should also try to keep that JSON schema up to date with any config schema changes so that we can provide a good IntelliSense experience for users they may be manually updating the `azure.yaml` file or for agents that run validation against our schemas.

wbreza (approved):
Looks good - add a couple unblocking suggestions that we should get in before release.

PR #6048, opened 2025-10-31
`azure.ai.agents` - Show agent endpoint on successful deploy and improve error handling

### LINKED ISSUE #6044: [AI Agent Extension] azd deploy - show endpoint info post successful agent deployment
After a successful azd deploy of an agent, the output should include key endpoint details for integration.

@jfomhover please add more details.

JeffreyCA:
Closes #6044

This PR improves the deployment and error handling logic for hosted agents in the `azure.ai.agents` extension.

### Deploy artifact

`Deploy()` now returns the agent details as an endpoint artifact to show up in the output on a successful deploy:

image

### Error handling and diagnostics

Based on guidance from the service team, when starting an agent container fails, the extension now attempts to get the failure error message from the agent container API. Sometimes the API may return an empty error message, in which case the following is shown:

image

In addition, the operation ID is now included to help with debugging and investigations.

### DIFF
--- a/cli/azd/extensions/azure.ai.agents/internal/project/service_target_agent.go
+++ b/cli/azd/extensions/azure.ai.agents/internal/project/service_target_agent.go
@@ -630,11 +630,25 @@ func (p *AgentServiceTargetProvider) startAgentContainer(
 				}
 
 				// Check if operation is complete
-				if completedOperation.Status == "Succeeded" || completedOperation.Status == "Failed" {
-					if completedOperation.Status == "Failed" {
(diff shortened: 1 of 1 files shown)


JeffreyCA:
> Nice.
> 
> Not blocking. But, what does it mean for a COBO agent? Would we get similar URL after a post-deploy hook?

Need to confirm but it should at least print out the container app endpoint. Having trouble testing COBO due to panic issue: https://github.com/Azure/azure-dev/issues/6050

PR #6052, opened 2025-10-31
`azure.ai.agents` - Update changelog and version to 0.0.2

JeffreyCA:


cli/azd/extensions/azure.ai.agents/extension.yaml
```diff
@@ -5,15 +5,15 @@ displayName: AI Foundry Agents
...
+version: 0.0.2
 language: go
 capabilities:
   - custom-commands
   - lifecycle-events
   - mcp-server
   - service-target-provider
 providers:
-  - name: azure.ai.agents
+  - name: azure.ai.agent
```
vhvb1989, line 16: You have to change this in the listen.go file

JeffreyCA: It was already updated in the extension but not the manifest

PR #6073, opened 2025-11-04
Enhance `no-prompt` support in extensions & prompt service

wbreza:
## Summary

Adds support for `azd ai agent init -m <manifest-uri> -e <environment-name> --no-prompt`
Will initialize an `azd` environment with the specified name and use intelligent defaults for prompts when available.

### Global Flag Parsing Improvements
- **Early flag parsing**: Parse global flags (`--no-prompt`, `--debug`, `--cwd`) before Cobra command tree construction to enable pre-execution logic and better support during extension callbacks
- **Singleton pattern**: Register `GlobalCommandOptions` as singleton in IoC container for consistent access across all components
- **Extension support**: Use `ParseErrorsWhitelist` to gracefully ignore unknown extension-specific flags during global flag parsing

### `--environment, -e` Support

- Adds optional `--environment` flag to the `azd ai agent init` command to support creating new `azd` environment inline with support for `--no-prompt` mode.

### `--no-prompt` Mode Support
- **Prompt service layer**: Added `--no-prompt` handling to `PromptSubscription`, `PromptLocation`, and `PromptResourceGroup` functions with config-based defaults
- **gRPC layer**: Added `--no-prompt` support for `Confirm`, `Select`, `MultiSelect`, and `Prompt` operations with default value fallback
- **Clean architecture**: Resource selection prompts (`PromptSubscriptionResource`, `PromptResourceGroupResource`) fail fast with clear error messages in `--no-prompt` mode

### DIFF
--- a/cli/azd/extensions/azure.ai.agents/internal/cmd/init.go
+++ b/cli/azd/extensions/azure.ai.agents/internal/cmd/init.go
@@ -58,6 +58,7 @@ type InitAction struct {
 	modelCatalogService *ai.ModelCatalogService
 	projectConfig       *azdext.ProjectConfig
 	environment         *azdext.Environment
+	flags               *initFlags
 }
 
 // GitHubUrlInfo holds parsed information from a GitHub URL
@@ -87,14 +88,9 @@ func newInitCommand() *cobra.Command {
 			}
 			defer azdClient.Close()
 
-			azureContext, projectConfig, err := ensureAzureContext(ctx, flags, azdClient)
+			azureContext, projectConfig, environment, err := ensureAzureContext(ctx, flags, azdClient)
 			if err != nil {
 				return fmt.Errorf("failed to ground into a project context: %w", err)
-			}
-
-			environment, err := ensureEnvironment(ctx, flags, azdClient)
-			if err != nil {
-				return fmt.Errorf("failed to ground into an environment: %w", err)
 			}
 
 			// getComposedResourcesResponse, err := azdClient.Compose().ListResources(ctx, &azdext.EmptyRequest{})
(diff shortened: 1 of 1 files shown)


trangevi (approved):

cli/azd/extensions/azure.ai.agents/internal/cmd/init.go
```diff
@@ -258,17 +266,22 @@ func ensureProject(ctx context.Context, azdClient *azdext.AzdClient) (*azdext.Pr
...
+func ensureEnvironment(ctx context.Context, flags *initFlags, azdClient *azdext.AzdClient) (*azdext.Environment, error) {
 	envResponse, err := azdClient.Environment().GetCurrent(ctx, &azdext.EmptyRequest{})
 	if err != nil {
 		fmt.Println("Lets create a new default environment for your project.")
 
+		envArgs := []string{"env", "new"}
+		if flags.env != "" {
+			envArgs = append(envArgs, flags.env)
+		}
+
```
JeffreyCA: Is the intended behaviour that if you already have an environment, the `flags.env` value is ignored? I was expecting `init` to create `flags.env` regardless

wbreza: Yeah - this would likely be the intended behavior.  I will update this. Nice catch!

cli/azd/cmd/auto_install.go
```diff
@@ -453,3 +468,67 @@ func ExecuteWithAutoInstall(ctx context.Context, rootContainer *ioc.NestedContai
...
+
+// CreateGlobalFlagSet creates a new flag set with all global flags defined.
+// This is the single source of truth for global flag definitions.
+func CreateGlobalFlagSet() *pflag.FlagSet {
+	globalFlags := pflag.NewFlagSet("global", pflag.ContinueOnError)
+
+	globalFlags.StringP("cwd", "C", "", "Sets the current working directory.")
+	globalFlags.Bool("debug", false, "Enables debugging and diagnostics logging.")
+	globalFlags.Bool(
+		"no-prompt",
```
vhvb1989, line 480: Can we avoid the duplication here?  We should not re-define commands or flags as part of auto-install..  All flags are already registered in the root cobra command

wbreza: This is now the same function used in the root cobra command... The issue is that the flags don't end up getting parsed in the case of the extension flow since it happens too soon.

JeffreyCA (approved):

trangevi (approved):

PR #6076, opened 2025-11-05
Environment Instance Caching

wbreza:
## Summary

Implements instance-level caching in the environment manager to ensure the same `*Environment` pointer is returned for the same environment name across different IoC scopes.

> [!NOTE]
> Caching isn't likely the ideal fix but this does address the issues without a large change.

### Problem
- gRPC services registered at root scope couldn't access environment instances created in command scopes
- Each IoC scope resolved `lazy.Lazy[*Environment]` to different instances
- Extensions modifying environment in one scope weren't visible in other scopes

### Solution
- Added instance cache (`map[string]*Environment`) in environment manager
- Cache keyed by environment name ensures consistent instance identity
- Cached instances are automatically reloaded on each `Get()` to sync with disk state
- Cache invalidated on `Delete()` operations

### Changes
- **manager.go**
  - Added `cacheMu` and `envCache` fields with initialization
  - Modified `Get()` to check cache before loading from disk
  - Added `getFromCache()` helper with automatic reload
  - Updated `Delete()` to invalidate cache entries
  
- **manager_test.go**
  - Added comprehensive test suite (`Test_EnvManager_InstanceCaching`)
  - Tests cover: instance identity, reload behavior, error handling, cache invalidation, multiple environments

### Benefits
✅ gRPC services and commands share same environment instance  
✅ Extension modifications visible across all scopes  
✅ Thread-safe with `sync.RWMutex` protection  
✅ Automatic reload ensures fresh data from disk  
✅ No breaking changes to existing API

### DIFF
--- a/cli/azd/internal/grpcserver/event_service_test.go
+++ b/cli/azd/internal/grpcserver/event_service_test.go
@@ -90,6 +90,11 @@ func createTestEventService() (*eventService, *MockEventStreamingServer) {
 	mockStream := &MockEventStreamingServer{}
 	extensionManager := &extensions.Manager{}
 
+	// Create lazy environment manager (mock)
+	lazyEnvManager := lazy.NewLazy(func() (environment.Manager, error) {
+		return nil, nil // Tests don't need actual environment manager
+	})
+
 	// Create lazy project with simple test config
 	lazyProject := lazy.NewLazy(func() (*project.ProjectConfig, error) {
 		projectConfig := &project.ProjectConfig{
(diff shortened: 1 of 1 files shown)


cli/azd/pkg/environment/manager.go
```diff
@@ -420,9 +437,34 @@ func (m *manager) Get(ctx context.Context, name string) (*Environment, error) {
...
+	m.cacheMu.RUnlock()
+
+	if !ok {
+		return nil, nil
+	}
+
+	// Reload cached instance to ensure it has latest data from disk
+	if err := m.Reload(ctx, cached); err != nil {
+		return nil, err
+	}
```
vhvb1989, line 463: Is this a cache if we are always reloading?

wbreza: Yeah, the idea was to just trigger a reload if returning from cache to ensure it has the latest data.

rajeshkamal5050 (approved):

PR #6083, opened 2025-11-05
Moving hooks warning from console to logs

vhvb1989:
As suggested in : https://github.com/Azure/azure-dev/issues/3613#issuecomment-3471755509

We will only show the note in logs (when running with debug) but should not use the main output by default


cli/azd/pkg/ext/hooks_manager.go
weikanglim, line 238: remove formatting before logging if it makes sense to

PR #6096, opened 2025-11-06
Extension Framework: gRPC Streaming updates

wbreza:
### Summary
Introduced a generic `MessageBroker[T]` pattern used in both client and server to eliminate duplicate streaming logic across extension services, reducing complexity while fixing critical concurrency issues.

### Key Changes
- **New**: Generic message broker (message_broker.go) with envelope pattern for correlation-based message routing
- **Refactored**: Event, framework, and service target managers now use shared broker (~600 lines removed)
- **Improved**: EOF detection, error handling, and test reliability

### API Changes
- `Send()` - fire-and-forget (renamed from `SendFireAndForget`)
- `SendAndWait()` - request/response with correlation (renamed from `Send`)
- `SendAndWaitWithProgress()` - with progress callbacks

### DIFF
--- a/cli/azd/pkg/azdext/extension_host.go
+++ b/cli/azd/pkg/azdext/extension_host.go
@@ -184,8 +184,8 @@ func (er *ExtensionHost) Run(ctx context.Context) error {
 
 	// Register all registrations in parallel - service targets, framework services, and event handlers
 	var allRegistrationsWg sync.WaitGroup
-	registrationCount := len(er.serviceTargets) + len(er.frameworkServices) + len(er.projectHandlers) + len(er.serviceHandlers)
-	registrationErrChan := make(chan error, registrationCount)
+	totalCount := len(er.serviceTargets) + len(er.frameworkServices) + len(er.projectHandlers) + len(er.serviceHandlers)
+	registrationErrChan := make(chan error, totalCount)
 
 	// Register service targets in parallel
 	for _, reg := range er.serviceTargets {


cli/azd/extensions/azure.ai.agents/internal/cmd/listen.go
```diff
@@ -23,6 +25,9 @@ func newListenCommand() *cobra.Command {
 		Use:   "listen",
 		Short: "Starts the extension and listens for events.",
 		RunE: func(cmd *cobra.Command, args []string) error {
+			// Suppress all log output during listen mode
+			log.SetOutput(io.Discard)
+
```
JeffreyCA: With this set, I'm not getting any output when I run `azd up`, and if I press `Ctrl+C` to stop, it seems to leave an orphaned `azure-ai-agents-linux-amd64` process as well

I had to run `pkill -f azure-ai-agents-linux-amd64` to kill it

wbreza: werid - it should only effect the extension itself... I'll look into it.

cli/azd/extensions/azure.ai.agents/internal/cmd/listen.go
```diff
@@ -33,6 +38,13 @@ func newListenCommand() *cobra.Command {
 			}
 			defer azdClient.Close()
 
+			// azdClient.Prompt().Confirm(ctx, &azdext.ConfirmRequest{
+			// 	Options: &azdext.ConfirmOptions{
+			// 		Message:      "Debug Extension",
+			// 		DefaultValue: ux.Ptr(true),
+			// 	},
+			// })
+
```
vhvb1989: remove?

cli/azd/extensions/azure.ai.agents/internal/cmd/listen.go
wbreza: Let me know if you are still able to reproduce this.

JeffreyCA: I see some output, but others like Docker logs are missing:

image

And the stale extension process issue still happens for me:

image

PR #6107, opened 2025-11-08
Adds `azure.ai.agent` support to schema

### LINKED ISSUE #6056: [ai agents extension] Schema Validation Error for `remoteBuild` under extension service target

**Output from `azd version`**

```text
azd version 1.21.0-beta.1-daily.5517383 (commit 922d09225ceadcfcf14a2851e62d030ed3c95621)
```

**Describe the bug**
The `remoteBuild` field is not recognized in the azure.yaml schema v1.0, causing validation warnings in the editor when trying to use the workaround for Docker connectivity issues.

**To Reproduce**

1. Open azure.yaml file in VS Code
2. Add the following configuration to work around Docker issues:
   ```yaml
   services:
       echo-agent:
           project: src/echo-agent
           host: azure.ai.agent
           language: python
           docker:
               remoteBuild: true
   ```
3. Observe schema validation warnings in the editor

**Expected behavior**
The `docker.remoteBuild` property should be recognized as a valid configuration option in the azure.yaml schema without causing validation warnings.

**Environment**

- OS: Windows
- Shell: PowerShell
- Language name and version: YAML schema validation
- IDE and version: VS Code with YAML language server

**Additional context**
The schema validation error occurs when using the yaml-language-server with the schema:
`https://raw.githubusercontent.com/Azure/azure-dev/main/schemas/v1.0/azure.yaml.json`

The `remoteBuild` option appears to be a valid workaround for Docker connectivity issues but is not documented in the schema, leading to editor warnings that may confuse users about whether this is a supported configuration option.

wbreza:
Fixes #6056

Adds support to the `azure.yaml` schema where `host: azure.ai.agent`

- Enables docker options
- Enables custom configuration

> [!NOTE]
> Also reorganizes conditionals improve readability.

### DIFF
--- a/schemas/alpha/azure.yaml.json
+++ b/schemas/alpha/azure.yaml.json
@@ -376,7 +376,7 @@
                             "required": ["project"],
                             "properties": {
                                 "config": {
-                                    "$ref": "https://raw.githubusercontent.com/wbreza/azure-dev/refs/heads/ext-agent-schema-int/cli/azd/extensions/azure.ai.agents/schemas/azure.ai.agent.json",
+                                    "$ref": "https://raw.githubusercontent.com/Azure/azure-dev/refs/heads/main/cli/azd/extensions/azure.ai.agents/schemas/azure.ai.agent.json",
                                     "title": "The Azure AI Agent configuration.",
                                     "description": "Optional. Provides additional configuration for Azure AI Agent deployment."
                                 },


schemas/alpha/azure.yaml.json
```diff
@@ -306,156 +307,101 @@
...
                         },
                         "then": {
+                            "required": ["project"],
                             "properties": {
-                                "dist": {
-                                    "type": "string",
-                                    "description": "Optional. The path to the directory containing a single Java archive file (.jar/.ear/.war), or the path to the specific Java archive file to be included in the deployment artifact. If omitted, the CLI will detect the output directory based on the build system in-use. For maven, the default output directory 'target' is assumed."
-                                }
+                                "config": {
+                                    "$ref": "https://raw.githubusercontent.com/wbreza/azure-dev/refs/heads/ext-agent-schema-int/cli/azd/extensions/azure.ai.agents/schemas/azure.ai.agent.json",
```
JeffreyCA: Should this point to `Azure/azure-dev`?

weikanglim (approved):

wbreza:
> It doesn't feel OK to me. The schema would be there even w/o installing the extension What if another extension wants to do the same?

I'm open to other ideas but JSON schema is not dynamic.  Since we have auto install - if a user followed the schema, chose a host at runtime the extension would get installed and all would work.  Perhaps this is the way for now for 1st party extensions till we figure out a better way to support this in the future.

rajeshkamal5050 (approved):

vhvb1989:
> > It doesn't feel OK to me. The schema would be there even w/o installing the extension What if another extension wants to do the same?
> 
> I'm open to other ideas but JSON schema is not dynamic. Since we have auto install - if a user followed the schema, chose a host at runtime the extension would get installed and all would work. Perhaps this is the way for now for 1st party extensions till we figure out a better way to support this in the future.

My recommendation is to avoid adding anything to `azure.yaml` that is only understood by some extension. 
Keep azure.yaml's schema clean and easy - it's overcomplicated enough already with azd's conditionals 🤓
Avoid to user azd as a midle-man between some extension and the user.
When it comes to extensions, it is hard for me to see the benefit(s) of using azure.yaml for something that is not only opt-in (extending azd functionality).

PR #6124, opened 2025-11-11
[AI Agents Extension] Saved Template Update

trangevi:
Refactor so we only expect to read a manifest one time. The rest of the time we're reading an agent definition

### DIFF
--- a/cli/azd/extensions/azure.ai.agents/internal/project/service_target_agent.go
+++ b/cli/azd/extensions/azure.ai.agents/internal/project/service_target_agent.go
@@ -361,6 +361,11 @@ func (p *AgentServiceTargetProvider) Deploy(
 		return nil, fmt.Errorf("failed to read YAML file: %w", err)
 	}
 
+	err = agent_yaml.ValidateAgentDefinition(data)
+	if err != nil {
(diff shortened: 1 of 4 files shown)


cli/azd/extensions/azure.ai.agents/internal/project/service_target_agent.go
```diff
@@ -407,81 +396,38 @@ func (p *AgentServiceTargetProvider) isContainerAgent() bool {
...
 
-	agentManifest, err := agent_yaml.LoadAndValidateAgentManifest(data)
-	if err != nil {
-		return false
-	}
-
-	// Convert the template to bytes
-	templateBytes, err := json.Marshal(agentManifest.Template)
-	if err != nil {
+	var genericTemplate map[string]interface{}
```
jfomhover, line 409: can we still using something like LoadAndValidateAgentManifest to ensure whatever we're loading is actually correct?

trangevi: I'm repurposing the validation logic we already had, since it really wasn't validating the manifest. I have a couple ideas for how to clean this up more generally, but that can wait for now.

jfomhover (approved):

PR #6129, opened 2025-11-11
[AI Agents Extension] More --project-id support

trangevi:


cli/azd/extensions/azure.ai.agents/internal/cmd/init.go
```diff
@@ -317,6 +363,33 @@ func ensureEnvironment(ctx context.Context, flags *initFlags, azdClient *azdext.
...
 			return nil, fmt.Errorf("environment not found")
 		}
+	} else if flags.projectResourceId != "" {
+		currentSubscription, err := azdClient.Environment().GetValue(ctx, &azdext.GetEnvRequest{
+			EnvName: existingEnv.Name,
+			Key:     "AZURE_SUBSCRIPTION_ID",
+		})
+		if err != nil {
+			return nil, fmt.Errorf("failed to get current subscription ID from environment: %w", err)
+		}
```
JeffreyCA: We may want to handle if `AZURE_SUBSCRIPTION_ID` or `AZURE_LOCATION` doesn't exist

trangevi: By this point we've either gotten an existing environment or created a new one, is it possible to have either of those flows succeed and not have a subscription and/or location set? Assuming not, then it seems valid to fail out here like we're doing

JeffreyCA: By default `azd env new <envname>` doesn't prompt for location or subscription, and neither does `azd init`

cli/azd/extensions/azure.ai.agents/internal/cmd/init.go
```diff
@@ -369,7 +369,15 @@ func ensureEnvironment(ctx context.Context, flags *initFlags, azdClient *azdext.
...
-			return nil, fmt.Errorf("failed to get current subscription ID from environment: %w", err)
+			// Set the subscription ID in the environment
+			_, err = azdClient.Environment().SetValue(ctx, &azdext.SetEnvRequest{
+				EnvName: existingEnv.Name,
+				Key:     "AZURE_SUBSCRIPTION_ID",
+				Value:   foundryProject.SubscriptionId,
+			})
+			if err != nil {
+				return nil, fmt.Errorf("failed to set tenant ID in environment: %w", err)
+			}
```
JeffreyCA, line 384: If a value doesn't exist, I believe `currentSubscription, err` would be `"", nil` so this should go outside the `err != nil` block?

trangevi: I see. I wonder if we have any other places where we expect that a "does not exist" will result in an exception... Let me update this here to start

JeffreyCA (approved):

PR #6137, opened 2025-11-11
Updates http request processing for AgentClient

wbreza:
Updates the `AgentClient` to use the Azure GO SDK pipeline & policies for request processing.


rajeshkamal5050:
@JeffreyCA @wbreza Can we also update the AgentClient to also use the userAgent and correlation policies set in our core too. That way the user-agent/correlation aspects flows through to the service teams.

Will be useful for telemetry(correlating) and also triaging issues requiring id's??

Issue #6130, opened 2025-11-11
[Aspire] Support for AppService - WithDeploymentSlot()


vhvb1989:
Expected for Aspire 13.1

PR #6158, opened 2025-11-13
[Aspire] Add support for webapp deploy to slots

vhvb1989:
fix: https://github.com/Azure/azure-dev/issues/6130

### DIFF
--- a/cli/azd/pkg/project/service_target_dotnet_containerapp.go
+++ b/cli/azd/pkg/project/service_target_dotnet_containerapp.go
@@ -305,6 +305,7 @@ func (at *dotnetContainerAppTarget) Deploy(
 
 	aspireDeploymentType := azapi.AzureResourceTypeContainerApp
 	resourceName := serviceConfig.Name
+	resourceMetadata := map[string]string{}
 	if useBicepForContainerApps {
 		// Compile the bicep template
 		deployment, err := func() (armDeployment, error) {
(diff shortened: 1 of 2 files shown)


cli/azd/pkg/project/service_target_dotnet_containerapp.go
```diff
@@ -890,6 +907,7 @@ func (at *dotnetContainerAppTarget) Endpoints(
 	resourceType := azapi.AzureResourceType(targetResource.ResourceType())
 	// Currently supports ACA, ACA Jobs, and WebApp for Aspire (on reading Endpoints)
 	if resourceType != azapi.AzureResourceTypeWebSite &&
+		resourceType != azapi.AzureResourceTypeWebSiteSlot &&
```
wbreza, line 910: Kind of odd that this aspire containerapp service target also supports deploying to webapps... We should consider renaming the service target to something more generic for aspire if this is the path forward.

weikanglim (approved):

cli/azd/pkg/project/service_target_dotnet_containerapp.go
vhvb1989: Yeah, the name is not accurate anymore. 
There was an idea about using `Aspire` as the host name in Azure.yaml, but that's not really a service in Azure.
So it is currently containerapp, even if you are deploying to WebApp...

Issue #6161, opened 2025-11-13
Layered provisioning to Beta


rajeshkamal5050:
Remaining work for Beta,
- Remove feature flag for the layered provisioning flow @weikanglim 
- Add Layered Provisioning – Beta to feature-stages.md @weikanglim 
- Include this update in the blog post @puicchan 
- Update provisioning docs, if any wasn't sure @alexwolfmsft @weikanglim

spboyer:
Close issue @rajeshkamal5050 from #6169

Issue #4289, opened 2024-09-04
ACA improvements for Stable/GA


rajeshkamal5050:

rajeshkamal5050:
Remaining work to move to stable/GA
- No new core changes needed. Related features are already merged as part of the dev-to-prod work i.e., azd publish/deploy decoupling, bicep based revisions, layered provisioning etc.
- Update feature-stages.md to mark ACA - Stable @weikanglim 
- Include this update in the blog post @puicchan 
- Update public ACA docs @alexwolfmsft @weikanglim

rajeshkamal5050:
https://github.com/Azure/azure-dev/blob/main/cli/azd/docs/feature-stages.md ACA is now stable.

PR #6169, opened 2025-11-13
Promote layered provisioning to beta

weikanglim:
Promote layered provisioning to beta, which includes these changes:

- Remove warning messages and config associated with alpha
- Add documentation for flags and args
- Update feature-stages.md 

Addresses #6161 #4289


rajeshkamal5050 (approved):
Thanks Wei 🚀

PR #6193, opened 2025-11-14
[Refactor] moving go.mod from root

vhvb1989:
This makes the repo structure idiomatic to golang multi-modules at monorepo projects

### DIFF
--- a/cli/azd/ci-build.ps1
+++ b/cli/azd/ci-build.ps1
@@ -146,13 +146,13 @@ $tagsFlag = "-tags=$($tags -join ',')"
 $ldFlags = @(
     "-s",
     "-w",
-    "-X 'github.com/azure/azure-dev/cli/azd/internal.Version=$Version (commit $SourceVersion)'"
+    "-X 'github.com/azure/azure-dev/internal.Version=$Version (commit $SourceVersion)'"
 )
 
 # Add GitHub Copilot linker flags if enabled
 if ($GitHubCopilotEnabled) {
-    $ldFlags += "-X 'github.com/azure/azure-dev/cli/azd/pkg/llm.clientID=$GitHubCopilotClientId'"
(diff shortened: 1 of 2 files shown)


hemarina (approved):
LGTM

cli/azd/go.mod
wbreza, line 1: If we are going to reorganize the modules we should split out to work with the following:

1. Go module for CLI
2. Go module for core package
3. Separate extensions directory

Right now the `go.mod` defines the following
```
module github.com/azure/azure-dev
```

But all the packages still existing in `cli/azd` so all the using references are out of sync.

vhvb1989: Before renaming, the go.mod needs to exists for current extensions to consume it.

see this error:

```
The issue is that the extensions are trying to fetch the module from GitHub, but it can't find go.mod at that path in the remote repository yet (since we haven't pushed)
```

This needs to be broken into multiple steps. I'll start with this one

eng/pipelines/templates/steps/template-test-run-job.yml
```diff
@@ -63,7 +63,7 @@ steps:
       displayName: Copy test-templates.sh
 
     - bash: |
-        go build -o ./temp/tfoidc ./cli/azd/test/internal/tfoidc 
+        cd cli/azd && go build -o ../../temp/tfoidc ./test/internal/tfoidc
```
danieljurek, line 66: This could also use a `workingDirectory` of `cli/azd` then you don't have to cd in your command.

danieljurek (approved):

eng/pipelines/templates/steps/template-test-run-job.yml
vhvb1989: This is the first PR, after this we will rename the module from azure-dev to azure-dev/cli/azd to match the path.
But the go.mod must exists first   I'll make the change you suggested in the next PR

PR #6212, opened 2025-11-17
[AI Agents Extension] Properly pass container resources

trangevi:


cli/azd/extensions/azure.ai.agents/internal/pkg/agents/agent_api/operations.go
```diff
@@ -549,10 +549,16 @@ func (c *AgentClient) DeleteAgentEventHandler(ctx context.Context, agentName, ev
 // Container Operations
 
 // StartAgentContainer starts a container for a specific version of an agent
-func (c *AgentClient) StartAgentContainer(ctx context.Context, agentName, agentVersion string, apiVersion string) (*AcceptedAgentContainerOperation, error) {
+func (c *AgentClient) StartAgentContainer(ctx context.Context, agentName, agentVersion string, minReplicas *int32, maxReplicas *int32, apiVersion string) (*AcceptedAgentContainerOperation, error) {
```
vhvb1989: nit. Move all the optional args into a `StartAgentContainerOptions` struct which can continue to grow w/o you adding more and more args to the function signature.

cli/azd/extensions/azure.ai.agents/internal/project/service_target_agent.go
```diff
@@ -693,9 +707,17 @@ func (p *AgentServiceTargetProvider) startAgentContainer(
...
+		minReplicasInt32 := int32(foundryAgentConfig.Container.Scale.MinReplicas)
+		maxReplicasInt32 := int32(foundryAgentConfig.Container.Scale.MaxReplicas)
+		minReplicas = &minReplicasInt32
+		maxReplicas = &maxReplicasInt32
+	}
+
+	// Start agent container
 	operation, err := agentClient.StartAgentContainer(
-		ctx, agentVersionResponse.Name, agentVersionResponse.Version, apiVersion)
+		ctx, agentVersionResponse.Name, agentVersionResponse.Version, minReplicas, maxReplicas, apiVersion)
```
wbreza: Nit: Consider applying the option based approach from the cpu/mem model to build up the desired request body instead of just adding additional arguments to the function.  May be easier to refactor/adjust over time.

wbreza: Do you want to pass any defaults? or just let the service handle defaults if not specified?

cli/azd/extensions/azure.ai.agents/internal/project/service_target_agent.go
```diff
@@ -516,10 +516,23 @@ func (p *AgentServiceTargetProvider) deployHostedAgent(
...
+		cpu = foundryAgentConfig.Container.Resources.Cpu
+		memory = foundryAgentConfig.Container.Resources.Memory
+	}
+
 	request, err := agent_yaml.CreateAgentAPIRequestFromDefinition(
 		agentDef,
 		agent_yaml.WithImageURL(fullImageURL),
 		agent_yaml.WithEnvironmentVariables(resolvedEnvVars),
+		agent_yaml.WithCPU(cpu),
+		agent_yaml.WithMemory(memory),
```
vhvb1989: When `cpu` and `memory` are not defined in `azure.yaml` you would see empty strings in here.
Make sure that empty strings are supported by the request you send to the service.

It is different to do-not  include a field `cpu` than sending a cpu with an empty string - as a way to declare that the user did not set any override for it.

cli/azd/extensions/azure.ai.agents/internal/project/service_target_agent.go
```diff
@@ -693,9 +707,17 @@ func (p *AgentServiceTargetProvider) startAgentContainer(
 	// Create agent client
 	agentClient := agent_api.NewAgentClient(azdEnv["AZURE_AI_PROJECT_ENDPOINT"], p.credential)
 
-	// Start agent container (minReplicas and maxReplicas are already int32)
+	var minReplicas, maxReplicas *int32
+	if foundryAgentConfig.Container != nil && foundryAgentConfig.Container.Scale != nil {
+		minReplicasInt32 := int32(foundryAgentConfig.Container.Scale.MinReplicas)
+		maxReplicasInt32 := int32(foundryAgentConfig.Container.Scale.MaxReplicas)
```
vhvb1989: Make sure that when the values are not defined in azure.yaml, we don't set them as `0` here (they default value).
nit. Validate the cast to int32 in case people add crazy big numbers or negative number or other symbols in the azure.yaml config to prevent panics

cli/azd/extensions/azure.ai.agents/internal/project/service_target_agent.go
trangevi: We'll let the service handle defaults if it gets that far with nothing set.

cli/azd/extensions/azure.ai.agents/internal/project/service_target_agent.go
trangevi: Since the config objects expect an int, we should fail sooner than this for any "other symbols"

PR #6255, opened 2025-11-19
Added helper func to get infra config with defaults applied

### LINKED ISSUE #6254: [Issue] Revision-based ACA deployment is not working in latest release
- [ ] Make sure you've installed the latest version using [instructions](https://github.com/Azure/azure-dev?tab=readme-ov-file#installupgrade-azure-developer-cli)

**Output from `azd version`**
azd version 1.21.3 (commit b01891fc63b1faf9a9633cdc630deab0b4d569c4)

**Describe the bug**
Working on blue/green deployment blog. Everything is working fine with version 1.21.0 but hit below error after upgrading to latest release 1.21.3

```

  (x) Failed: Deploying service web

ERROR: error executing step command 'deploy --all': failed deploying service 'web': updating container app service: getting container app: getting container app: parameter containerAppName cannot be empty
```

When using 1.21.0

Image

wbreza:
Fixes #6254

### DIFF
--- a/cli/azd/internal/cmd/add/add.go
+++ b/cli/azd/internal/cmd/add/add.go
@@ -286,7 +286,7 @@ func (a *AddAction) Run(ctx context.Context) (*actions.ActionResult, error) {
 
 	infraOptions, err := prjConfig.Infra.GetWithDefaults()
 	if err != nil {
-		return nil, fmt.Errorf("getting infra options: %w", err)
+		return nil, err
 	}
 
 	infraRoot := infraOptions.Path
--- a/cli/azd/pkg/infra/provisioning/bicep/bicep_provider.go
+++ b/cli/azd/pkg/infra/provisioning/bicep/bicep_provider.go
@@ -46,13 +46,6 @@ import (
 
 type bicepFileMode int
 
-var (
-	defaultOptions = provisioning.Options{
-		Module: "main",
-		Path:   "infra",
-	}
-)
-
 const (
 	bicepMode bicepFileMode = iota
 	bicepparamMode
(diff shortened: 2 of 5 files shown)


cli/azd/pkg/infra/provisioning/provider.go
```diff
@@ -34,6 +36,30 @@ type Options struct {
 	Layers []Options `yaml:"layers,omitempty"`
 }
 
+// GetWithDefaults merges the provided infra options with the default provisioning options
```
JeffreyCA, line 39: What was the underlying issue? The order of the merges?

wbreza: The actual issue was just a missed check/merge in the containerapp service target.

cli/azd/pkg/infra/provisioning/provider.go
```diff
@@ -34,6 +36,30 @@ type Options struct {
...
+func (o Options) GetWithDefaults(other ...Options) (Options, error) {
```
vhvb1989, line 40: We don't usually add the `Get` prefix to methods. And we typically reserve `With.....()` for option builder pattern.

This method is composing 3 layers or the same options into one, right ?  
1- the current state in `o Options`
2- the provided state(s) from `other ...Options`
3- default for whatever is still not set.

I wonder why do we all of these. Are we making this too much complicated for some special use-case?

I wish we could just always start from `defaultOptions` and directly override values as we need.

rajeshkamal5050: Created a issue for it - https://github.com/Azure/azure-dev/issues/6284

PR #6257, opened 2025-11-19
[Refeactoring] Promote azdext as a new go mod

vhvb1989:
This PR separates the azdext package as a lib module to define the protocol between azd and extensions.
Then make azd to use it to implement the functionality
And make extensions to consume the azdext package as an SDK to get azd's functionality.

In this PR we are also removing the replace settings in extensions modules so they use the azdext module


cli/azd/pkg/azdext/go.mod
```diff
@@ -0,0 +1,22 @@
+module github.com/azure/azure-dev/cli/azd/pkg/azdext
```
wbreza, line 1: We need to get more organized on the overall strategy before we start introducing these changes... the module paths just aren't as clear as they can be... For example, azdext should not be under `cli`. The CLI should be its own module, shared libs/packages should be outside of this.

vhvb1989: The path doesn't really matter. You can have azdext nested inside /cli/azd and still be an standalone go module.

What it matters is how independent as a package it is. When I was doing this I discovered that azdext depends on azd/package modules while azd also depends on azdext.  That would still happen even if you move the modules to be sibling folders.

PR #6258, opened 2025-11-19
Fix for demo extension + funct tests

vhvb1989:
- **Fix missing mapper and add functional sample for demo ext to build and publish container**
- **functional tests using demo extension**

### DIFF
--- a/cli/azd/test/functional/testdata/samples/storage-rg/infra/main.parameters.json
+++ b/cli/azd/test/functional/testdata/samples/storage-rg/infra/main.parameters.json
@@ -1,0 +1,12 @@
+{
+  "$schema": "https://schema.management.azure.com/schemas/2019-04-01/deploymentParameters.json#",
+  "contentVersion": "1.0.0.0",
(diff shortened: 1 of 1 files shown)


hemarina (approved):
LGTM

JeffreyCA:
Looks like tests are timing out:

> Enter a value for the 'environmentName' infrastructure parameter: Error: value for 'environmentName' must be at least '1' in length.

vhvb1989:
> Looks like tests are timing out:
> 
> > Enter a value for the 'environmentName' infrastructure parameter: Error: value for 'environmentName' must be at least '1' in length.

Yeah, I thought recordings would work but is doesn't - For now I changed the test to live only.

Something tells me that all I need is to use the azd binary which has the recorder for the instance of azd that extensions runs... but I might be wrong. Maybe what we actually need is to move the recorder out-of-proc so it can be used by multiple azd binaries.

JeffreyCA (approved):

Issue #6175, opened 2025-11-13
`azd template list` and `azd init` template option update


SophCarp:
This is an initial change just to clean up the commands. A second issue will be made to create a more higher quality experience. 

In both `azd template list` and `azd init` (template) option, replace curret experience with the following text then exit the interaction: 

> For Microsoft Ai Gallery visit https://azure.github.io/ai-app-templates/
> For azd community gallery (awesome azd) visit https://azure.github.io/awesome-azd/ 
> 
> Select a template from the gallery, then run azd init -t [template name]

Then exit the interaction. 

<img width="625" height="189" alt="Image" src="https://github.com/user-attachments/assets/065aa8ef-913f-4e91-9dd6-23ebbcb413df" />

<img width="1250" height="414" alt="Image" src="https://github.com/user-attachments/assets/713906f7-1b1a-48a7-a603-a0aed2724e29" />

Giving the links to the galleries gives a more useful interaction than the current long list of templates.

SophCarp:
This starts to address #66 #3964 and #5027

SophCarp:
Assigning to @rajeshkamal5050, @kristenwomack for awareness.

kristenwomack:
Very excited about this first step!

weikanglim:
I'm not entirely opposed to the suggestion, although I do see great value in having a CLI experience that is streamlined and interactive.

I do believe we have rather huge usability concerns with the current listing. I'm wondering if we would consider taking a stab at making it better. I put forward a PR in #6264 just to see what changes I would suggest.

Repasting the comparison screenshots below.

Before:
<img width="2073" height="724" alt="image" src="https://github.com/user-attachments/assets/e66d329f-2e99-4120-8d5b-a0633e90782c" />

After:
<img width="2095" height="388" alt="image" src="https://github.com/user-attachments/assets/3ff18ac2-f093-43c8-ab0f-afb720b3df3b" />

weikanglim:
I do like the gallery URLs being printed out. I could see this as extremely helpful for discovery.

I would suggest printing out the gallery URLs when the user cancels out of the selection menu.

> To browse all available templates in a web gallery, visit aka.ms/awesome-azd or aka.ms/ai-apps

vhvb1989:
Maybe the problem is that this doesn't play well in all terminal sizes, @weikanglim 
[...]

PR #6264, opened 2025-11-20
rework template list in 'azd init' for usability

weikanglim:
Changes include:
1. Sort lines alphabetically to reduce visual noise.
2. DIsplay selection entries individually on single lines by default to avoid visual scanning required.
3. Add language information since that is maximally useful.

Before:
image

After:
image

Contributes to #6175

### DIFF
--- a/cli/azd/pkg/templates/template.go
+++ b/cli/azd/pkg/templates/template.go
@@ -36,11 +36,6 @@ type Template struct {
 	// A list of tags associated with the template
 	Tags []string `json:"tags"`
 
-	// A list of languages supported by the template
-	//
-	// As of November 2025, known values include: bicep, php, javascript, dotnetCsharp, typescript, python, nodejs, java
-	Languages []string `json:"languages,omitempty"`
-
 	// Additional metadata about the template
 	Metadata Metadata `json:"metadata,omitempty"`
 }
--- a/cli/azd/pkg/templates/awesome_source.go
+++ b/cli/azd/pkg/templates/awesome_source.go
@@ -22,7 +22,11 @@ type awesomeAzdTemplate struct {
 	Source           string   `json:"source"`
 	Tags             []string `json:"tags"`
 	AzureServiceTags []string `json:"azureServices"`
-	LanguageTags     []string `json:"languages"`
+
+	// A list of languages supported by the template
+	//
+	// As of November 2025, known values include: bicep, php, javascript, dotnetCsharp, typescript, python, nodejs, java
+	LanguageTags []string `json:"languages"`
 }
 
 // newAwesomeAzdTemplateSource creates a new template source from the awesome-azd templates json file.


vhvb1989:
@weikanglim , isn't it supposed to exit the iteration with a message?  instead of showing a list?

> In both azd template list and azd init (template) option, replace curret experience with the following text then exit the interaction

According to the referenced issue

kristenwomack:
If we are not going to exit to the galleries and keep the interactive experience, it needs to be revised. @SophCarp already has several designs in the referenced issues. The one thing I don't think is helpful is alphabetical order. I would much prefer that we order by template success. Imagining that someone is starting from the list, we should set them up to succeed. If they are truly browsing and comparing, the gallery is the right place for that to see more attributes.

cli/azd/pkg/templates/template_manager.go
```diff
@@ -196,47 +197,41 @@ func PromptTemplate(
...
+		path := template.CanonicalPath()
 
-		if slices.Contains(templateNames, templateChoice) {
-			duplicateNames = append(duplicateNames, templateChoice)
-		}
+		choices = append(choices, fmt.Sprintf("%s\t%s\t[%s]",
+			choice,
+			path,
+			strings.Join(languages, ",")))
+	}
```
wbreza, line 225: I feel like this design is going to cause lots of line wrapping due to template names + URLs + languages unless user is running in a very wide terminal window.

cli/azd/pkg/templates/template.go
```diff
@@ -35,6 +36,11 @@ type Template struct {
 	// A list of tags associated with the template
 	Tags []string `json:"tags"`
 
+	// A list of languages supported by the template
+	//
+	// As of November 2025, known values include: bicep, php, javascript, dotnetCsharp, typescript, python, nodejs, java
+	Languages []string `json:"languages,omitempty"`
```
wbreza: Where is languages initialized and set on the Template struct? I don't see that change here.

cli/azd/pkg/templates/template.go
```diff
@@ -73,3 +79,42 @@ func (t *Template) Display(writer io.Writer) error {
...
+			if !slices.Contains(t.Tags, "nodejs") {
+				languages = append(languages, "ts")
+			}
+		case "python", "java":
+			languages = append(languages, lang)
+		}
+	}
+
+	return languages
+}
```
wbreza, line 101: Feels like the "Languages" property of the template should be set/initialized when the template list is retrieved & parsed like other members on the struct.

cli/azd/pkg/templates/template_manager.go
weikanglim: It does wrap in an aligned fashion: 
image

cli/azd/pkg/templates/template.go
weikanglim: Removed this.  We have a custom github template source that appends the `languages` property we store in awesome-azd to `tags`

cli/azd/pkg/templates/template.go
weikanglim: Removed this -- the generic template source only knows about `Tags` and not `Languages` (awesome-azd specific).

PR #4011, opened 2024-06-17
hooks: require environment and project always 

weikanglim:
Since hooks was introduced, it has always taken a soft dependency on project and environment being available. This change introduces a hard dependency and validates environment and project being available before deciding to register or run hooks.

PR #6277, opened 2025-11-22
Allow hooks to run without initial `.azure` directory

### LINKED ISSUE #3920: Project hooks don't work when there's not environment, like in CI/CD
Consider the next azd project sample:

```yaml
name: ehooks
hooks:
  preprovision:
    shell: sh
    continueOnError: false
    interactive: true
    run: echo "Preprovisioning..."
```

If we run `azd provision --no-prompt` and:
 - there's is not an `.azure` folder:  The hook is not triggered b/c there is no environment.
 > hooks.go:52: azd environment is not available, skipping all hook registrations

See: https://github.com/Azure-Samples/azure-search-openai-demo/issues/1603

This is an issue when azd runs in CI, because the `.azure` folder is not in the repo.  Instead, azd uses `AZURE_ENV_NAME` to create the .azure folder with a new env using the name from that env var.
However, the hooks registration happens before the environment is created, so the hooks are ignored.

Workaround:
For CI, add one step to create the environment before calling azd provision, like:

```yaml
      - name: Create azd env
        run: azd env new $AZURE_ENV_NAME
        env:
          AZURE_ENV_NAME: ${{ vars.AZURE_ENV_NAME }}
```

The `--no-prompt` flag is not supported for `azd env new`, so, we need to use the env var from github as the argument for creating the environment.

JeffreyCA:
Fixes #3920

This PR revives the main changes from #4011 and addresses the issue where project-level hooks (for example, `preprovision`) would not run when there was no existing `.azure` environment folder. This commonly occurs in CI/CD scenarios:

- The repo does **not** contain `.azure/…`.
- `azd provision --no-prompt` is run with `AZURE_ENV_NAME` set.

Previously, the hooks middleware relied on lazily-resolved environment and project configuration. As a result, hooks that should have run on the very first `provision` (such as pre-provision validation or setup) were ignored.

### What this change does

- Refactors the hooks middleware to take concrete dependencies (`environment.Manager`, `*environment.Environment`, `*project.ProjectConfig`) instead of `lazy.Lazy` wrappers. This removes the extra lazy lookup layer that used `environment.Manager.Get` (which only works when `.azure/<env>` already exists) and aligns the middleware with the same `*environment.Environment` instance that the rest of the command pipeline initializes via `LoadOrInitInteractive`, so hooks now see newly created environments (including the first CI run with `AZURE_ENV_NAME`) instead of only environments that already existed on disk.
- Adds explicit, defensive checks in the middleware `Run` method (addressing the review feedback from [PR #4011](https://github.com/Azure/azure-dev/pull/4011)) so that if the environment or project configuration is not yet available, the middleware logs a clear message and safely skips hook registration without panicking.

### `azd provision` without `AZURE_ENV_NAME` set

image

### `azd provision` with `AZURE_ENV_NAME` set

image

### GitHub Actions - `azd provision` with `AZURE_ENV_NAME` set

image

### DIFF
--- a/cli/azd/cmd/middleware/hooks.go
+++ b/cli/azd/cmd/middleware/hooks.go
@@ -54,16 +54,6 @@ func NewHooksMiddleware(
 
 // Runs the Hooks middleware
 func (m *HooksMiddleware) Run(ctx context.Context, next NextFn) (*actions.ActionResult, error) {
-	if m.env == nil {
-		log.Println("azd environment is not available, skipping all hook registrations.")
-		return next(ctx)
-	}
-
-	if m.projectConfig == nil {
-		log.Println("azd project is not available, skipping all hook registrations.")
-		return next(ctx)
-	}
-
 	// Validate hooks and display any warnings
 	if !m.options.IsChildAction(ctx) {
 		if err := m.validateHooks(ctx, m.projectConfig); err != nil {


cli/azd/cmd/middleware/hooks.go
```diff
@@ -14,102 +14,92 @@ import (
...
 		log.Println("azd environment is not available, skipping all hook registrations.")
 		return next(ctx)
 	}
 
-	projectConfig, err := m.lazyProjectConfig.GetValue()
-	if err != nil || projectConfig == nil {
+	if m.projectConfig == nil {
 		log.Println("azd project is not available, skipping all hook registrations.")
 		return next(ctx)
 	}
```
weikanglim: Consider removing the nil check since neither `env` nor `projectConfig` could be nil here by construction. It isn't something we check for in other usages in azd.  

It does mean that the registration of hooks itself depends on environment being present always, which is fine given the current usage. This may be a limitation that shows up over time. However, we can cross that bridge when we need to.

cli/azd/cmd/middleware/hooks.go
```diff
@@ -14,102 +14,92 @@ import (
...
 func (m *HooksMiddleware) registerCommandHooks(
 	ctx context.Context,
-	env *environment.Environment,
-	projectConfig *project.ProjectConfig,
 	next NextFn,
 ) (*actions.ActionResult, error) {
-	if len(projectConfig.Hooks) == 0 {
+	if len(m.projectConfig.Hooks) == 0 {
 		log.Println(
 			"azd project is not available or does not contain any command hooks, skipping command hook registrations.",
```
weikanglim:
```suggestion
			"azd project does not contain any command hooks, skipping command hook registrations.",
```

Issue #6207, opened 2025-11-17
'azd down' does not remove deployment state if resources are already gone


irinasmoke:
**Output from `azd version`**
Run `azd version` and copy and paste the output here: azd version 1.21.3

**Describe the bug**
When running `azd down`, if resources have previously been manually deleted, the deployment state is not removed. This leads to subsequent `azd provision` calls not working properly.

**To Reproduce**
1. `azd up` a project
2. Manually delete resource group
3. Run `azd down`
4. Run `azd up` 
5. Deployment fails because provisioning fails

**Expected behavior**
`azd down` should remove deployment state, giving the user a clear path to re-provision, even if resources have already been removed by other means.

**Additional context**
Actually, the ideal experience would be to have `azd provision` be able to detect that resources have been manually deleted and be able to re-provision anyway without users needing to know about the --no-state flag, but I have been told this is not so straightforward, so this is my next best proposed solution - at least make `azd down` behave as expected in this scenario.

weikanglim:
> the ideal experience would be to have azd provision be able to detect that resources have been manually deleted and be able to re-provision anyway without users needing to know about the --no-state flag,

I do very much prefer this behavior of azd checking if resource groups have been delted before proceeding. Proposed a change in #6281 based on this.

kristenwomack:
Does #6281 also fix and close this issue?

rajeshkamal5050:
Updated behavior post fixes around `azd down` and `azd provision`
- AZD based provisioned and down. No changes.
- AZD provisioned, Portal deleted resource group, subsequent AZD provision - Reprovisions detecting rg was deleted outside of azd.
- AZD provisioned, Portal deleted resource group. AZD down - No Azure resources were found.

Note: Deleting deployment history is not a good practice, since those are primarily kept for book-keeping purposes i.e., auditing, compliance etc.

Issue #5667, opened 2025-08-29
Improve error message for "resource not found" when resource group has been manually deleted (suggest --no-state workaround)


Be3751:
- [x] Make sure you've installed the latest version using [instructions](https://github.com/Azure/azure-dev?tab=readme-ov-file#installupgrade-azure-developer-cli)

**Describe the bug**
When running `azd up`, if the target resource group has been manually deleted by the user (e.g., via Azure Portal or CLI), the command fails with an error like:
```
  (-) Skipped: Didn't find new changes.

Deploying services (azd deploy)

  (x) Failed: Deploying service checkout-worker

ERROR: error executing step command 'deploy --all': getting target resource: getting default resource groups for environment: aca-otel-tracing: resource not found: 0 resource groups with prefix or suffix with value: 'aca-otel-tracing'
```

Currently, the error message does not provide guidance on how to resolve this situation. 

In fact, running `azd provision --no-state` allows the user to ignore the cached state and successfully re-provision the resources. However, this workaround is not obvious to most users.

**To Reproduce**
1. Run `azd up` to create an environment.
2. Delete the resource group manually in the Azure Portal or with `az group delete`.
3. Run `azd up` again.
4. Observe the `resource not found` error without clear recovery instructions.

**Expected behavior**
The error message should include a hint such as:  
> "If the resource group was manually deleted, try running `azd provision --no-state` to re-create the resources."

This would improve the user experience by making the workaround discoverable and reducing friction.

**Environment**
- Language: Python 3.11.10
- IDE: VS Code 1.103.1
- OS: Windows 11 / WSL2 (Ubuntu 22.04.3)

**Additional context**
This issue is related to #3559 and #3771, where the inability to detect deleted resources has been discussed.  
This proposal does not aim to solve the detection logic itself, but rather to **improve the user experience by surfacing the existing workaround (`--no-state`) in the error message**.  

I am considering submitting a PR for this change, but I wanted to confirm the team’s direction first.

weikanglim:
> The error message should include a hint such as:
> "If the resource group was manually deleted, try running azd provision --no-state to re-create the resources."

This seems like a perfectly fine improvement. Happy to take a PR.

weikanglim:
Superseded by #6281 azd now checks for resource groups before skipping the deployment

PR #6281, opened 2025-11-25
invalidate deployment state based on resource group existence checks

### LINKED ISSUE #3771: [Issue] azd up should consider whether the resource group exists before opting not to create it
- [X] Make sure you've installed the latest version using [instructions in the wiki](../wiki/install)

**Output from `azd version`**

1fastapi-azure-function-apim % azd version
azd version 1.8.0 (commit 8246323c2472148288be4b3cbc3c424bd046b985)

**Describe the bug**
I ran `azd up` on a resource group that I probably deleted at some point. I got this error:

```
fastapi-azure-function-apim % azd env select fastapi-azf-apim
fastapi-azure-function-apim % azd up

Packaging services (azd package)

  (✓) Done: Packaging service api
  - Package Output: /var/folders/nq/2sq7jzp17tl97sgxzbptr2d80000gn/T/fastapi-azure-function-apim-api-azddeploy-1713804018.zip

SUCCESS: Your application was packaged for Azure in 5 seconds.

Subscription: ca-pamelafox-demo-test (32ea8a26-5b40-4838-b6cb-be5c89a57c16)
Location: East US

  (-) Skipped: Didn't find new changes.

SUCCESS: There are no changes to provision for your application.

  (x) Failed: Deploying service api

ERROR: getting target resource: getting default resource groups for environment: fastapi-azf-apim: resource not found: 0 resource groups with prefix or suffix with value: 'fastapi-azf-apim'

ERROR: error executing step command 'deploy --all': getting target resource: getting default resource groups for environment: fastapi-azf-apim: resource not found: 0 resource groups with prefix or suffix with value: 'fastapi-azf-apim'
```

**To Reproduce**

You probably need to run azd up, then delete the RG in Portal, then run again (with no Bicep change)

**Expected behavior**

It should have provisioned.

**Environment**
Information on your environment: Mac OSX M1 Terminal

**Additional context**

To workaround, I will call `azd provision --no-state`

weikanglim:
This change factors in quick resource-group existence checks as heuristics to the deployment state calculation prior to skipping the deployment. 

With this change, when resource groups are deleted outside of azd, `azd provision` will detect this and continue with provisioning without skipping.

The resource group check leverages the most lightweight possible  HTTP `HEAD` call to the relevant ARM endpoint, and is measured to add ~100ms for a single resource group. Given that most deployments likely have 1-2 resource groups, it is then expected that the overall latency increase is <0.2 seconds overall.

Note that only resource groups **created** by the deployment are checked for. Existing resource groups associated to resources deployed into are not factored in.

Fixes #3771
Would supersede #5667

### DIFF
--- a/cli/azd/test/recording/recording.go
+++ b/cli/azd/test/recording/recording.go
@@ -344,7 +344,7 @@ func Start(t *testing.T, opts ...Options) *Session {
 					return
 				}
 
-				t.Fatal("recorderProxy: " + msg)
+				t.Log("recorderProxy: " + msg)
 			},
 			Recorder: vcr,
 		},


JeffreyCA:
Would this also fix https://github.com/Azure/azure-dev/issues/6207? There's a related PR https://github.com/Azure/azure-dev/pull/6267 that I wasn't sure about either...

cli/azd/pkg/infra/provisioning/bicep/bicep_provider.go
```diff
@@ -588,8 +588,29 @@ func (p *BicepProvider) Deploy(ctx context.Context) (*provisioning.DeployResult,
...
-		deploymentState, err := p.deploymentState(ctx, planned, deployment, currentParamsHash)
-		if err == nil {
+		deploymentState, stateErr := p.deploymentState(ctx, planned, deployment, currentParamsHash)
+		if stateErr == nil {
+			// As a heuristic, we also check the existence of all resource groups
+			// created by the deployment to validate the deployment state.
+			// This handles the scenario of resource group(s) being deleted outside of azd,
+			// which is quite common.
+			// This check adds ~100ms per resource group to the deployment time.
+			for _, res := range deploymentState.Resources {
```
vhvb1989, line 598: Should this be recursive?  In case `Resources` contains a resource type deployment which contains rg?

weikanglim: Great question! Recursion isn't necessary here. The `outputResources` property from ARM already contains all resource(s) created from the current and nested deployments.

rajeshkamal5050:
@weikanglim will this be the new behavior?

When rg is manually deleted, `azd provision` will error `resource group foo no longer exists, invalidating deployment state`? are we also `invalidating deployment state`? did not see any code for that. If yes, can we also handle the same for when doing `azd down` question which @JeffreyCA had too?

weikanglim:
@rajeshkamal5050 The new behavior is as follows:

When resource groups are deleted outside of azd, `azd provision` will detect this and continue with provisioning **without skipping**.

@JeffreyCA I left a comment on #6207 to fix the underlying behavior without changing `azd down`.

JeffreyCA (approved):

PR #6288, opened 2025-11-26
add source classification for existing usage metadata

weikanglim:
Wrap existing attribute keys with additional metadata for classification.

Contributes to https://github.com/Azure/azure-dev-pr/issues/1744

### DIFF
--- a/cli/azd/.vscode/cspell-azd-dictionary.txt
+++ b/cli/azd/.vscode/cspell-azd-dictionary.txt
@@ -73,6 +73,7 @@ BUILDID
 BUILDNUMBER
 buildpacks
 byoi
+callstack
 cflags
 charmbracelet
 circleci
@@ -205,6 +206,7 @@ preinit
 protogen
 proxying
 psanford
+pseudonymized
 psycopg
 psycopgbinary
 pulumi


cli/azd/internal/tracing/fields/fields.go
```diff
@@ -10,118 +10,253 @@ import (
...
+		Classification: SystemMetadata,
+		Purpose:        FeatureInsight,
+	}
 )
 
 // Context level fields. Availability depends on the command running.
-const (
+var (
 	// Object ID of the principal.
 	ObjectIdKey = attribute.Key(contracts.UserAuthUserId) // user_AuthenticatedId
```
JeffreyCA, line 139: Should `ObjectIdKey` also be converted?

JeffreyCA (approved):

rajeshkamal5050 (approved):

wbreza (approved):

cli/azd/internal/tracing/fields/fields.go
weikanglim: It isn't necessary for this change -- I didn't include it since it isn't used currently.

PR #6289, opened 2025-11-26
Add operations-per-run limit to stale issues workflow

spboyer:
Currently we are running everyday at 3am, it will grab 30 at a time (it does auto paginate), picks up where it left off. However, at this pace will take ~31 days to get through our issues. However when new issues are added, it starts over.


kristenwomack:
@rajeshkamal5050 - will you review and push this change? Looks like it needs one code owner.

rajeshkamal5050 (approved):

PR #6313, opened 2025-12-03
implement three solutions prompt on error handling flow

### LINKED ISSUE #6302: error handling improvements
- Remove the open-ended prompt (requesting user feedback) from the error handling workflow
- Test and enable three options/continue/cancel for users to select
- Clean up and prepare for testing and rollout

hemarina:
* upgrade to tmc/langchaingo 0.1.14 (contains bug fix)
* add three solutions prompt
* remove open-ended prompt in error handling flow
* had a UX quick review with designers
* fix https://github.com/Azure/azure-dev/issues/6302
image
image
image
image
image


hemarina:
Could we also discuss this https://github.com/Azure/azure-dev/pull/5827#issuecomment-3609172808? My workaround is to make it clear that do not run azd commands like up, provision, deploy, down so it only continues with provision after the fix. 
> @wbreza During recent testing of the error-handling flow, it creates a long waiting time. Can we revisit whether this is expected behavior? If we decide to keep it, I recommend updating the UX to display the full azd command being executed besides log. This would give users more visibility into what’s happening and help them understand the reason for the delay.
> image

vhvb1989 (approved):

cli/azd/cmd/middleware/error.go
```diff
@@ -340,3 +365,103 @@ func promptForErrorHandlingConsent(
...
+	for i, solution := range solutions {
+		choices[i] = &uxlib.SelectChoice{
+			Value: solution,
+			Label: "Yes. " + solution,
+		}
+	}
+
+	choices[len(solutions)] = &uxlib.SelectChoice{
+		Value: "continue",
+		Label: fmt.Sprintf("Yes, allowing %s to fix the error independently", agentName),
```
JeffreyCA:
```suggestion
		Label: fmt.Sprintf("Yes, let %s choose the best approach", agentName),
```
I think something like this sounds a bit better and clearer

cli/azd/cmd/middleware/error.go
```diff
@@ -340,3 +365,103 @@ func promptForErrorHandlingConsent(
...
+	}
+
+	choices[len(solutions)+1] = &uxlib.SelectChoice{
+		Value: "cancel",
+		Label: "No, cancel",
+	}
+
+	selector := uxlib.NewSelect(&uxlib.SelectOptions{
+		Message:         fmt.Sprintf("Allow %s to fix the error?", agentName),
+		HelpMessage:     "Select a solution to proceed, continue with AI fix, or cancel",
```
JeffreyCA: The help message could be a little confusing as the 3 solutions would be fixed with AI as well - thoughts on something like "Select a specific solution or let AI determine the best fix" or "Select a suggested fix, or let AI decide" or similar?

JeffreyCA (approved):

cli/azd/cmd/middleware/error.go
hemarina: @SophCarp I think Jeffrey's suggestion is great. Checking with you for your opinion as well

SophCarp: I'm happy with that!

SophCarp (approved):
Looks good to me! I agree with Jeffrey's comments

JeffreyCA (approved):

cli/azd/cmd/middleware/error.go
```diff
@@ -207,26 +209,67 @@ func (e *ErrorMiddleware) Run(ctx context.Context, next NextFn) (*actions.Action
 		previousError = originalError
 		agentOutput, err := azdAgent.SendMessage(ctx, fmt.Sprintf(
 			`Steps to follow:
-			1. Use available tool to identify, explain and diagnose this error when running azd command and its root cause.
-			2. Resolve the error by making the minimal, targeted change required to the code or configuration.
-			Avoid unnecessary modifications and focus only on what is essential to restore correct functionality.
-			3. Remove any changes that were created solely for validation and are not part of the actual error fix.
-			Error details: %s`, errorInput))
+            1. Use available tool to identify, explain and diagnose this error when running azd command and its root cause.
```
wbreza:
```suggestion
            1. Use available tools to identify, explain and diagnose this error when running azd command and its root cause.
```

cli/azd/cmd/middleware/error.go
```diff
@@ -207,26 +209,67 @@ func (e *ErrorMiddleware) Run(ctx context.Context, next NextFn) (*actions.Action
...
+			e.console.Message(ctx, "")
+			selectedSolution, continueWithFix, err := promptUserForSolution(ctx, solutions, agentName)
+			if err != nil {
+				return nil, fmt.Errorf("prompting for solution selection: %w", err)
+			}
+
+			if continueWithFix {
+				agentOutput, err := azdAgent.SendMessage(ctx, fmt.Sprintf(
+					`Steps to follow:
+            1. Use available tool to identify, explain and diagnose this error when running azd command and its root cause.
```
wbreza:
```suggestion
            1. Use available tools to identify, explain and diagnose this error when running azd command and its root cause.
```

cli/azd/cmd/middleware/error.go
```diff
@@ -340,3 +366,103 @@ func promptForErrorHandlingConsent(
 
 	return choices[*choiceIndex].Value, nil
 }
+
+// extractSuggestedSolutions extracts the three solutions from the "Brainstorm Solutions" section
+// in the LLM response. Returns a slice of solution strings.
+func extractSuggestedSolutions(llmResponse string) []string {
```
wbreza: Nit: Instead of this manual parsing of the response you could instead ask the agent to return a structured JSON response based on some schema that you specify.  They you can parse the response and marshal it into some types.

wbreza (approved):
Looks good - added a couple minor suggestions.

JeffreyCA (approved):

cli/azd/cmd/middleware/error.go
```diff
@@ -53,19 +53,35 @@ func NewErrorMiddleware(
...
 	// - Running in CI/CD environment where user interaction is not possible
 	if !e.featuresManager.IsEnabled(llm.FeatureLlm) || e.global.NoPrompt || resource.IsRunningOnCI() {
-		return actionResult, err
+		return next(ctx)
 	}
 
+	// Preserve a non-cancellable parent context BEFORE making the first attempt
+	// This ensures that if the context gets cancelled during the first attempt,
+	// retries can use a fresh context
+	parentCtx := context.WithoutCancel(ctx)
```
hemarina, line 78: @wbreza This is the context I'm talking about. The error shows that context is canceled which I believe it happened in HTTP Client Layer. The error returned in part of microsoft-auth-library while running function [AcquireTokenSilent](https://github.com/AzureAD/microsoft-authentication-library-for-go/blob/v1.4.2/apps/public/public.go#L337). Since original context was cancelled, retry loop will be failed on a canceled context. The fix created context from parent context which is not cancellable, and the context is valid for authentication.
```
ERROR: error executing step command 'provision': failed to authenticate: server response error:
Get "https://login.microsoftonline.com/common/discovery/instance?api-version=1.1&authorization_endpoint=https%3A%2F%2Flogin.microsoftonline.com%2Forganizations%2Foauth2%2Fv2.0%2Fauthorize%22: context canceled"
```

Issue #2020, opened 2023-04-25
azd down prompts init steps when running in a directory without a .azure file


Austinauth:
- [x] Make sure you've installed the latest version using [instructions in the wiki](../wiki/install)

**Output from `azd version`**
Run `azd version` and copy and paste the output here: azd version 0.8.0-beta.2 (commit 429b6ccf9a9d1616efa8988f2f8561deeb464c71)

**Describe the bug**
Running `azd down` in a directory with no `.azure` file prompts the user input an environment name, subscription and location before attempting to delete the resources from azure.

**To Reproduce**

Run `azd down` in a directory with no `.azure` file, but containing other template files.
<img width="690" alt="MicrosoftTeams-image (3)" src="https://user-images.githubusercontent.com/6994218/234326035-7d83a76b-7792-4739-bf4d-6d69f80c5dc3.png">

<img width="880" alt="MicrosoftTeams-image (4)" src="https://user-images.githubusercontent.com/6994218/234326402-ea9363a1-ced1-44dc-a160-7d6133f4e656.png">

**Expected behavior**
In this scenario we should probably show the user an error message in this scenario and not prompt them to enter the env, location, and sub. 

**Additional context**
- I ran `azd down` in a directory I had previously ran `azd init` in
- I did not complete the init originally, hence the reason there are some azd files in the directory, but not the `.azure` file
- I ran `azd down` in this directory because I wasn't sure if I had deployed to Azure and wanted to be safe and delete any resources if I had.

spboyer:
Investigated. The root cause is in the DI registration (cmd/container.go:239-263) where *environment.Environment is wired through LoadOrInitInteractive() for all commands. When azure.yaml exists but no .azure directory/default env, azd down falls through to interactive env create/select prompts instead of erroring.

The fix would be to change the down command path to resolve the environment via envManager.Get() (non-interactive) instead of LoadOrInitInteractive(). If no environment exists, return a clear error:
'No environment is configured. Run azd env new or azd init first.'

This requires changing the DI wiring specifically for the down action to avoid the interactive fallback, which is a targeted but careful change in container.go.

PR #6345, opened 2025-12-10
Improve deletion behavior for resource-group–scoped deployments

### LINKED ISSUE #6295: azd down with Layered Provisioning prompts for resource group creation after first layer deletes it
### Describe the bug
When using `azd down` with Layered Provisioning (Beta), if the first processed layer deletes the resource group, subsequent layers prompt the user to "Pick a resource group to use" with options including "Create a new resource group". This is unexpected behavior during a teardown operation.

### To Reproduce
1. Create a project with Layered Provisioning enabled (`layered: true` in azure.yaml)
2. Configure at least two layers where one layer creates the Azure resource group
3. Run `azd up` to deploy all layers
4. Run `azd down`
5. After the first layer completes deletion (which removes the resource group), observe that the next layer prompts for resource group selection

### Expected behavior
`azd down` should gracefully handle the case where the resource group has already been deleted by a previous layer, without prompting the user to create or select a new resource group.

### Screenshots or Videos
After the first layer deletion completed:
```
Deleted: <resources from first layer>

(✓) Done: Deleting Azure resources
===== Layer: core =====

? Pick a resource group to use:  [Use arrows to move, type to filter]
> 1. Create a new resource group
  2. ...
```

### Environment
- OS: Linux
- azd version: 1.21.3

### Additional context
- This occurs with Layered Provisioning (Beta) feature
- Workaround: Press Ctrl+C when the prompt appears, then manually verify resource deletion
- Layer order in azure.yaml matters - `azd down` processes layers in reverse order of `azd up`

weikanglim:
Resource-group–scoped projects have been hitting rough edges during `azd down`, mostly around unexpected prompts and inconsistent handling of missing resources (#6295, #2020). This change tightens up that workflow so deletion behaves predictably and doesn't get in the user’s way.

Severall small changes are made to accomplish the fix:

* **Keep `AZURE_RESOURCE_GROUP` intact during deletion.**  
  Multiple layers may rely on the same group, so removing it during `down` caused unnecessary prompts and confusion. We now keep the value and let `azd provision` decide how to handle it by checking whether the group actually exists before asking anything.

* **Minimize what we prompt for in `azd down`.**  
  Deletion doesn’t need full parameter resolution. We now validate only the minimal scope (`AZURE_SUBSCRIPTION`, `AZURE_LOCATION`, and/or `AZURE_RESOURCE_GROUP`) and skip the rest.

* **Handle 404s more gracefully when fetching resources.**  
  A missing resource group shouldn’t break the flow. We now treat it as a normal deletion path instead of an error that triggers prompts.

Also, fix a minor UX bug with resource group deletions not being displayed correctly.

Fixes #6295, #2020.

### DIFF
--- a/cli/azd/test/functional/testdata/recordings/Test_CLI_Up_ResourceGroupScope.yaml
+++ b/cli/azd/test/functional/testdata/recordings/Test_CLI_Up_ResourceGroupScope.yaml
@@ -12,7 +12,7 @@ interactions:
         host: management.azure.com
         remote_addr: ""
         request_uri: ""
-        body: '{"location":"eastus2","name":"rg-azdtest-we60cd1","tags":{"DeleteAfter":"2025-04-15T18:41:13Z"}}'
+        body: '{"location":"eastus2","name":"rg-azdtest-da130ca","tags":{"DeleteAfter":"2025-12-10T20:19:33Z"}}'
         form: {}
         headers:
             Accept:
@@ -26,8 +26,8 @@ interactions:
             Content-Type:
                 - application/json
             User-Agent:
-                - azsdk-go-armresources/v1.2.0 (go1.24.2; Windows_NT)
-        url: https://management.azure.com:443/subscriptions/faa080af-c1d8-40ad-9cce-e1a450ca5b57/resourcegroups/rg-azdtest-we60cd1?api-version=2021-04-01
+                - azsdk-go-armresources/v1.2.0 (go1.25.1; darwin)
+        url: https://management.azure.com:443/subscriptions/faa080af-c1d8-40ad-9cce-e1a450ca5b57/resourcegroups/rg-azdtest-da130ca?api-version=2021-04-01
         method: PUT
       response:
         proto: HTTP/2.0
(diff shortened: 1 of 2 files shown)


cli/azd/pkg/infra/provisioning/provider.go
```diff
@@ -22,18 +22,31 @@ const (
...
-
 	// Provisioning options for each individually defined layer.
 	Layers []Options `yaml:"layers,omitempty"`
+
+	// Runtime options
+
+	// IgnoreDeploymentState when true, skips the deployment state check.
+	IgnoreDeploymentState bool `yaml:"-"`
+	// The mode in which the deployment is being run.
+	Mode Mode `yaml:"-"`
```
vhvb1989, line 49: is this `Mode` something that is azd specific? I got confused for a moment if we were adding the ARM deployment mode (incremental / complete / etc). Should we reserve the `mode` for that?

vhvb1989 (approved):

cli/azd/pkg/infra/provisioning/provider.go
weikanglim: I did take a second look at this, and decided to leave it as-is for now.

In the grand scheme of things, I believe this should be eventually made in "`PlanMode`", which is something that is generalizable and is supported today by Terraform.

In Terraform, `plan --destroy` and `plan` (for apply) are two separate modes. This makes sense when we think about how `destroy` doesn't need all parameters to execute and isn't functionally equivaelent.

As for how it is implemented today, I tried separating out `Plan` from `Initialize`, but it ended up being a difficult refactor currently with bicep/bicepparam/ade codepaths all merged together and was too much scope for the fix. 

> if we were adding the ARM deployment mode (incremental / complete / etc)

I wouldn't worry too much about this given how the provider is non-bicep specific, but we can definitely rename all these and cross that bridge if we were to drop other provider support and make everything bicep-specific.

Issue #6332, opened 2025-12-08
[Issue] azd ai agent init -m truncates branch names containing “/”, causing file-not-found errors


iuiaoin:
- [x] Make sure you've installed the latest version using [instructions](https://github.com/Azure/azure-dev?tab=readme-ov-file#installupgrade-azure-developer-cli)

**Output from `azd version`**
Run `azd version` and copy and paste the output here:
azd version 1.22.0 (commit c9ab0f2dcc59298e258032e804b2d3c85338c56c)

**Describe the bug**
When running azd ai agent init -m, I encountered an issue where GitHub branch names containing a slash (/) are incorrectly truncated.

For example, our actual branch name is:
```shell
agentserver/first-release
```

However, during the initialization process, azd truncates it to:
```shell
agentserver
```

Since the truncated branch does not exist in the repository, the command fails with a file-not-found error.

**To Reproduce**
azd ai agent init -m https://github.com/Azure/azure-sdk-for-net/blob/agentserver/first-release-lixiaoli/sdk/agentserver/samples/AgentFramework/Workflow/BasicWorkflow/agent.yaml

**Expected behavior**
azd should correctly handle branch names containing slashes and use the full branch name without truncation.

**Environment**
Information on your environment:
    * macos
    * dotnet 9.0.306

**Additional context**
Branch names are truncated at the first /, resulting in invalid branch references and failed operations.
Please investigate and fix branch name parsing for the -m option.

vhvb1989:
Assigning @trangevi back.  I have added the fix to azd and exposed the api for the extension to use it.  @trangevi you can now replace the gh-url parsing in the extension to use azd.

PR #6351, opened 2025-12-10
Fix gh-url parse for branches with slashes

vhvb1989:
In this PR:
 - Improving the gh-url parsing to extract information from different url formats. Now we support urls that contains a branch with slashes in it. Like `feature/branch`. The parser uses gh cli to validate each combination until the longest valid branch is discovered.
 - Exposing this functionality to the extension framework.
 - Updating Demo extension to show how to use the functionality from extensions.  

 Example running the demo extension:
 ```
 azdd demo gh-url-parse https://github.com/Azure/azure-sdk-for-net/blob/agentserver/first-release-lixiaoli/sdk/agentserver/samples/AgentFramework/Workflow/BasicWorkflow/agent.yaml

GitHub URL Information
Hostname:    github.com
Repository:  Azure/azure-sdk-for-net
Branch:      agentserver/first-release-lixiaoli
File Path:   sdk/agentserver/samples/AgentFramework/Workflow/BasicWorkflow/agent.yaml

 ```

required for: https://github.com/Azure/azure-dev/issues/6332

### DIFF
--- a/cli/azd/internal/grpcserver/project_service_test.go
+++ b/cli/azd/internal/grpcserver/project_service_test.go
@@ -6,14 +6,17 @@ package grpcserver
 import (
 	"context"
 	"path/filepath"
+	"strings"
 	"testing"
 
 	"github.com/azure/azure-dev/cli/azd/pkg/azdext"
 	"github.com/azure/azure-dev/cli/azd/pkg/config"
 	"github.com/azure/azure-dev/cli/azd/pkg/environment"
 	"github.com/azure/azure-dev/cli/azd/pkg/environment/azdcontext"
+	"github.com/azure/azure-dev/cli/azd/pkg/exec"
 	"github.com/azure/azure-dev/cli/azd/pkg/lazy"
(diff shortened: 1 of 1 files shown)


cli/azd/grpc/proto/project.proto
```diff
@@ -15,6 +15,9 @@ service ProjectService {
 
   // AddService adds a new service to the project.
   rpc AddService(AddServiceRequest) returns (EmptyResponse);
+
+  // ParseGitHubUrl parses a GitHub URL and extracts repository information.
+  rpc ParseGitHubUrl(ParseGitHubUrlRequest) returns (ParseGitHubUrlResponse);
```
weikanglim, line 20: Curious here -- what was the motivating factor to choose api instead of library?

vhvb1989: You mean to have it accesible from the extension layer instead of creating a public library?

This parser uses `gh cli`, which is auto-installed and managed by azd.

rajeshkamal5050 (approved):

PR #6356, opened 2025-12-10
fix resource discovery logic from deployments

### LINKED ISSUE #6324: [Issue] `azd down <layer>` ignores layer argument and deletes all resource groups with same `azd-env-name` tag

**Output from `azd version`**

```
azd version 1.22.0 (commit 8e77dc23aa579dad562d70e5f895d3e78387c67f)
```

**Describe the bug**

When using `azd down <layer>` to delete resources for a specific layer in a multi-layer deployment, the command ignores the layer argument and deletes **all** resource groups that have the same `azd-env-name` tag, including resource groups from other layers.

The help text for `azd down` states:
```
When <layer> is specified, only deletes resources for the given layer.
```

However, this does not work as documented.

**To Reproduce**

1. Create a multi-layer deployment with `azure.yaml`:

```yaml
name: my-project
infra:
  layers:
    - name: shared
      path: ./infra/shared
    - name: environment
      path: ./infra/environment
services:
  myapp:
    host: containerapp
    project: ./src/myapp
```

2. Deploy all layers:

```bash
azd up
```

This creates two resource groups:
- `rg-my-project-shared`
- `rg-my-project-dev`

Both have the same `azd-env-name` tag value, but **neither has the `azd-layer-name` tag** that would distinguish them.

3. Attempt to delete only the `environment` layer:

```bash
azd down environment
```

4. **Result**: Both `rg-my-project-shared` AND `rg-my-project-dev` are deleted.

**Expected behavior**

`azd down environment` should only delete the `rg-my-project-dev` resource group (belonging to the `environment` layer), leaving the shared resource group intact.

**Environment**

- OS: Ubuntu 22.04 (WSL)
- Shell: zsh

**Additional context**

**Root Cause Analysis**

After tracing the code, the issue is in `ListSubscriptionDeploymentResources()` in [`cli/azd/pkg/azapi/standard_deployments.go` (lines 340-359)](https://github.com/Azure/azure-dev/blob/main/cli/azd/pkg/azapi/standard_deployments.go#L340-L359):

```go
func (ds *StandardDeployments) ListSubscriptionDeploymentResources(...) {
    // Get the environment name from the deployment tags
    envName, has := subscriptionDeployment.Tags[azure.TagKeyAzdEnvName]
    ...
    // Get all resource groups tagged with the azd-env-name tag
    resourceGroups, err := ds.resourceService.ListResourceGroup(ctx, subscriptionId, &ListResourceGroupOptions{
        TagFilter: &Filter{Key: azure.TagKeyAzdEnvName, Value: *envName},
    })
```

**The deletion flow:**

1. `downAction.Run()` in `cmd/down.go` correctly identifies the layer and calls `Destroy()` for the specified layer only
2. `BicepProvider.Destroy()` calls `CompletedDeployments(ctx, scope, p.env.Name(), p.layer, "")` - passing the layer parameter
3. `CompletedDeployments()` in `deployment_manager.go` correctly finds the deployment matching both `azd-env-name` AND `azd-layer-name` tags (confirmed: deployments DO have `azd-layer-name` tags)
4. `deploymentToDelete.Resources()` → calls `ListSubscriptionDeploymentResources()`
5. **BUG HERE**: `ListSubscriptionDeploymentResources()` queries resource groups using **only** `azd-env-name`, ignoring `azd-layer-name`

**Evidence:**

Subscription deployments have both tags correctly:
```
$ az deployment sub list --query "[?starts_with(name, 'dev')].{name:name, tags:tags}" -o json
[
  {"name": "dev-environment-...", "tags": {"azd-env-name": "dev", "azd-layer-name": "environment", ...}},
  {"name": "dev-shared-...", "tags": {"azd-env-name": "dev", "azd-layer-name": "shared", ...}}
]
```

However, resource groups only have `azd-env-name`:
```
$ az group show -n rg-aca-rev-mode-lab-dev --query tags
{"azd-env-name": "dev"}  # No azd-layer-name tag!
```

**Summary:**
- The `azd-layer-name` tag IS applied to deployments (set in `bicep_provider.go` Deploy function)
- The `azd-layer-name` tag is NOT applied to resource groups during provisioning
- `ListSubscriptionDeploymentResources()` filters resource groups by `azd-env-name` only, ignoring layer information

weikanglim:
In #6324, it was discovered that deletion for deployments in separate layers were deleting more resources than intended. Specifically, all resource groups associated with `azd-env-name` tag were deleted, rather than resource groups associated with the deployment.

To fix this, the logic in `ListSubscriptionDeploymentResources` has been updated to find resources associated to a deployment. This change effectively restores the behavior prior to https://github.com/Azure/azure-dev/commit/72bc4981adcb61ef014377dbfa1c2b7ab2aad7c0.

As part of this change, the existing logic in `bicep_provider.go` was migrated over. 

Fixes #6324

### DIFF
--- a/cli/azd/pkg/azapi/standard_deployments_test.go
+++ b/cli/azd/pkg/azapi/standard_deployments_test.go
@@ -91,44 +91,6 @@ func TestResourceGroupsFromDeployment(t *testing.T) {
 		require.Equal(t, []string{"matell-2508-rg"}, resourceGroupsFromDeployment(mockDeployment))
 	})
 
-	t.Run("references used when no output resources", func(t *testing.T) {
-		mockDeployment := &ResourceDeployment{
-			//nolint:lll
-			Id: "/subscriptions/faa080af-c1d8-40ad-9cce-e1a450ca5b57/providers/Microsoft.Resources/deployments/matell-2508-1689982746",
-			//nolint:lll
-			DeploymentId: "/subscriptions/faa080af-c1d8-40ad-9cce-e1a450ca5b57/providers/Microsoft.Resources/deployments/matell-2508-1689982746",
-			Name:         "matell-2508",
-			Type:         "Microsoft.Resources/deployments",
-			Tags: map[string]*string{
-				"azd-env-name": to.Ptr("matell-2508"),
-			},
-			ProvisioningState: DeploymentProvisioningStateFailed,
-			Timestamp:         time.Now(),
-			Dependencies: []*armresources.Dependency{
-				{
-					//nolint:lll
-					ID: to.Ptr(
-						"/subscriptions/faa080af-c1d8-40ad-9cce-e1a450ca5b57/resourceGroups/matell-2508-rg/providers/Microsoft.Resources/deployments/resources",
-					),
-					ResourceName: to.Ptr("resources"),
-					ResourceType: to.Ptr("Microsoft.Resources/deployments"),
-					DependsOn: []*armresources.BasicDependency{
-						{
-							//nolint:lll
-							ID: to.Ptr(
-								"/subscriptions/faa080af-c1d8-40ad-9cce-e1a450ca5b57/resourceGroups/matell-2508-rg",
-							),
-							ResourceName: to.Ptr("matell-2508-rg"),
-							ResourceType: to.Ptr("Microsoft.Resources/resourceGroups"),
-						},
-					},
-				},
-			},
-		}
-
-		require.Equal(t, []string{"matell-2508-rg"}, resourceGroupsFromDeployment(mockDeployment))
-	})
-
 	t.Run("duplicate resource groups ignored", func(t *testing.T) {
 
 		mockDeployment := ResourceDeployment{


cli/azd/pkg/azapi/standard_deployments_test.go
```diff
@@ -45,3 +49,117 @@ func Test_StandardDeployments_GenerateDeploymentName(t *testing.T) {
...
+							ResourceName: to.Ptr("matell-2508-rg"),
+							ResourceType: to.Ptr("Microsoft.Resources/resourceGroups"),
+						},
+					},
+				},
+			},
+		}
+
+		require.Equal(t, []string{"matell-2508-rg"}, resourceGroupsFromDeployment(mockDeployment))
+	})
```
JeffreyCA, line 92: This seems to be a dupe of the above test, should one be removed?

cli/azd/pkg/azapi/standard_deployments_test.go
```diff
@@ -45,3 +49,117 @@ func Test_StandardDeployments_GenerateDeploymentName(t *testing.T) {
...
+
+	t.Run("references used when no output resources", func(t *testing.T) {
```
weikanglim:
```suggestion
	t.Run("references used when no output resources", func(t *testing.T) {
```

cli/azd/pkg/azapi/standard_deployments_test.go
weikanglim: It does seem to be duplicated, removed.

rajeshkamal5050 (approved):

PR #6367, opened 2025-12-12
azd x support interactive

vhvb1989:
Adding this feature in the extensions-framework and to the `azd x` extension to allow extensions to opt-in to be executed as interactive-mode, allowing them to own stdout and stdin completely.
Without this option, extensions can only run writing to azd buffers. This blocks extensions to be fully TUI CLI programs.

This change is required to use `azd concurx` extension as TUI app. I will create a PR that uses this changes in concurX next.

### DIFF
--- a/cli/azd/cmd/extensions.go
+++ b/cli/azd/cmd/extensions.go
@@ -184,6 +184,8 @@ func (a *extensionAction) Run(ctx context.Context) (*actions.ActionResult, error
 	options := &extensions.InvokeOptions{
 		Args: a.args,
 		Env:  allEnv,
+		// cmd extensions are always interactive (connected to terminal)
+		Interactive: true,
 	}
 
 	_, err = a.extensionRunner.Invoke(ctx, extension, options)


cli/azd/pkg/extensions/runner.go
```diff
@@ -63,6 +63,13 @@ func (r *Runner) Invoke(ctx context.Context, extension *Extension, options *Invo
 		runArgs = runArgs.WithStdErr(options.StdErr)
 	}
 
+	// Enable interactive mode for extensions that opt-in via the "interactive" property
+	// in their extension.yaml. This ensures stdin/stdout/stderr are directly connected
+	// to the terminal, which is required for TUI applications like Bubble Tea.
+	if extension.Interactive {
+		runArgs = runArgs.WithInteractive(true)
+	}
+
```
wbreza, line 70: The extension runner is already invoked with options to set stdin/stdout/stderr - So what does Interactive provide us here? This seems redundant.

vhvb1989: The extensions middleware uses custom buffers (ext.StdIn(), ext.StdOut(), ext.StdErr()) - these are bytes.Buffer and DynamicMultiWriter for capturing output). 

The `interactive` option is a way to override this behavior and let the extension to use the terminal directly.

I will make a little change to make this cleaner

weikanglim (approved):

Issue #5971, opened 2025-10-22
Locked or Deprecated Dependencies Found


vhvb1989:
1. Azure Spring Apps SDK - DEPRECATED ⚠️
Package: github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/appplatform/armappplatform/v2 v2.0.1
Status: DEPRECATED (service retired March 31, 2025)
Used by: spring_app.go (direct dependency)
Message: "Please note, this package has been deprecated following the retirement of the Azure Spring Apps service on March 31, 2025. For more information, please refer to https://aka.ms/asaretirement."
Impact: High - this is a direct dependency actively used in the codebase

vhvb1989:
@weikanglim  should we remove all we have in azd for spring app service?

rajeshkamal5050:
Yes, @vhvb1989 @weikanglim let us remove support for ASA in AZD. Related issue https://github.com/Azure/azure-dev/issues/4504
- https://learn.microsoft.com/en-us/azure/spring-apps/basic-standard/retirement-announcement

vhvb1989:
> Yes, [@vhvb1989](https://github.com/vhvb1989) [@weikanglim](https://github.com/weikanglim) let us remove support for ASA in AZD. Related issue [#4504](https://github.com/Azure/azure-dev/issues/4504)
> 
> * https://learn.microsoft.com/en-us/azure/spring-apps/basic-standard/retirement-announcement

Created a PR for it.

Issue #4504, opened 2024-10-31
Fail to run `azd up` for `ASA-Samples-Event-Driven-Application`


Menghua1:
**Describe the issue:**
Running `azd up` fails with the following error:
![Image](https://github.com/user-attachments/assets/7680b8a8-9a66-4522-ba0f-e334b487c0b5)

**Repro Steps:**

1. Login in with command: `azd auth login`.
2. Run command `azd init -t ASA-Samples-Event-Driven-Application`.
3. Run command `azd up`.

**Environment:**

- Template: `ASA-Samples-Event-Driven-Application`.
- Branch: main
- OS: `Devcontainer`, `Windows`.
- Azd version: azd version 1.11.0-beta.1-daily.4247716 (commit https://github.com/Azure/azure-dev/commit/f4695c408eab9ea3cdedd1c3cb39aa01ad90123e).
 >Notes: We also have this issue when testing with the azd release version: azd version 1.10.3 (commit 0595f33fe948ee6df3da492567e3e7943cb9a733).

**Expected behavior:**
`azd up` can run successfully.

@rajeshkamal5050  for notification.

rajeshkamal5050:
@fangjian0423 @yiliuTo can you help cleanup and remove,
- Core ASA Support in AZD
- Doc changes/updates
- Templates removal

Menghua1:
@rajeshkamal5050 In the latest round of azd manual testing, this issue still exists.

- Template: ASA-Samples-Event-Driven-Application.
- Branch: staging.
- OS: Devcontainer.
- Azd version: azd version 1.12.0-beta.1-daily.4359062 (commit https://github.com/Azure/azure-dev/commit/6cd82ae9ff02e156a0056e994215191469442ab3).

Menghua1:
@rajeshkamal5050 In the latest round of azd manual testing, this issue still exists.

- Template: ASA-Samples-Event-Driven-Application.
- Branch: staging.
- OS: Devcontainer.
- Azd version: azd version 1.12.0-beta.1-daily.4442669 (commit 9b60cbc2a703200eb04d325988dc359886b2dfcb).

Menghua1:
@rajeshkamal5050 In the latest round of azd manual testing, this issue still exists.

- Template: ASA-Samples-Event-Driven-Application.
- Branch: staging.
- OS: Devcontainer.
[...]

PR #6369, opened 2025-12-12
Deprecating Azure Spring Apps support

vhvb1989:
fix: https://github.com/Azure/azure-dev/issues/5971
fix: https://github.com/Azure/azure-dev/issues/4504

After this PR, azd would return this error for any app/template using Azure Spring Apps:

image

### DIFF
--- a/cli/azd/pkg/project/service_target_springapp.go
+++ b/cli/azd/pkg/project/service_target_springapp.go
@@ -14,7 +14,7 @@ import (
 
 var errSpringAppDeprecated = errors.New(
 	"Azure Spring Apps is no longer supported. Recommend using Azure Container Apps." +
-		"For more information, please refer to https://aka.ms/asaretirement",
+		" For more information, please refer to https://aka.ms/asaretirement",
 )
 
 type springAppTarget struct {


spboyer:
@kristenwomack @haileyhuber8 - please remove any related templates from the galleries and add an issue to those repos for deprecation notice on the service.

vhvb1989:
> Azure spring is not fully turned off till 2028 - Removing this now will essentially break all users who have not migrated their applications. Have we reviewed our telemetry usage to understand the full impact of this change?

won't break their apps. Will only break a new version of azd if they want to deploy.
They will be able to use previous version of azd or some other way if they don't migrate yet.
@kristenwomack might know the numbers of users... but I think it's close to zero

cli/azd/pkg/project/service_target.go
```diff
@@ -30,12 +30,12 @@ const (
 
 // DotNetContainerAppTarget is intentionally omitted because it is only used internally when
 // containerizing .NET projects and is not a valid service host value in azure.yaml.
+// SpringAppTarget is intentionally omitted because support has been removed.
```
kristenwomack, line 33: thank you for adding this

cli/azd/pkg/project/service_target_springapp.go
```diff
@@ -6,65 +6,51 @@ package project
...
 	"github.com/azure/azure-dev/cli/azd/pkg/async"
-	"github.com/azure/azure-dev/cli/azd/pkg/azapi"
 	"github.com/azure/azure-dev/cli/azd/pkg/environment"
 	"github.com/azure/azure-dev/cli/azd/pkg/tools"
 )
 
-const (
-	defaultDeploymentName = "default"
+var errSpringAppDeprecated = errors.New(
+	"support for Azure Spring Apps has been removed from azd. " +
```
kristenwomack:
```suggestion
	"Azure Spring Apps is no longer supported. Recommend using Azure Container Apps." +
```

kristenwomack:
> > Azure spring is not fully turned off till 2028 - Removing this now will essentially break all users who have not migrated their applications. Have we reviewed our telemetry usage to understand the full impact of this change?
> 
> won't break their apps. Will only break a new version of azd if they want to deploy. They will be able to use previous version of azd or some other way if they don't migrate yet. @kristenwomack might know the numbers of users... but I think it's close to zero

yes, @rajeshkamal5050 investigated the numbers and it's very low

rajeshkamal5050 (approved):

PR #6389, opened 2025-12-16
map all errors

weikanglim:
Update error telemetry to capture actual error types instead of generic "UnknownError" classifications for a better error breakdown.

**Changes:**

- Replace generic UnknownError with specific error type classification using a new `errorType()` helper that unwraps errors to find the root cause
- Add `ErrType` telemetry attribute to explicitly track error types
- Introduce `DeploymentOperation` enum to distinguish between deployment, validation, and preview operations with granular error codes
- Update `MissingToolErrors` to include which tools are missing
- Add explicit handling for `ErrorWithSuggestion` and `LogInRequiredError`

### DIFF
--- a/cli/azd/internal/cmd/errors.go
+++ b/cli/azd/internal/cmd/errors.go
@@ -101,7 +101,13 @@ func MapError(err error, span tracing.Span) {
 			}
 		}
 
-		errCode = "service.arm.deployment.failed"
+		// Use operation-specific error code if available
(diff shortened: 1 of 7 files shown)


rajeshkamal5050:
Thanks @weikanglim this should help us more data for UnknownError cases in telemetry. @JeffreyCA lets rollout a patch release today, so we can start to see data coming in cc @kristenwomack @spboyer

PR #6426, opened 2025-12-23
Moving the code to main fork

achauhan-scc:


cli/azd/extensions/registry.json
```diff
@@ -1333,152 +1035,367 @@
...
+            }
+          ],
+          "artifacts": {
+            "darwin/amd64": {
+              "checksum": {
+                "algorithm": "sha256",
+                "value": "af11047544db004c13d0c68b781fb57edeb6358656b1cf3b4b1bf80e6dffd13b"
+              },
+              "entryPoint": "azure-ai-finetune-darwin-amd64",
+              "url": "https://github.com/achauhan-scc/azure-dev/releases/download/azd-ext-azure-ai-finetune_0.0.2-preview/azure-ai-finetune-darwin-amd64.zip"
```
vhvb1989: if the plan is to move the code to azure-dev repo, let's hold on updating the registry until the binaries are not provided from a fork

cli/azd/extensions/registry.json
```diff
@@ -595,184 +510,6 @@
...
-            },
-            {
-              "name": "publish",
-              "description": "Publish the AZD extension project to an extension source.",
-              "usage": "azd x publish"
-            },
-            {
-              "name": "release",
-              "description": "Create an new release of the AZD extension project to a Github repository.",
-              "usage": "azd x release"
```
vhvb1989, line 718: This is not right.
Maybe this is because you had an old registry or something.  

Let's remove the changes to the registry (all this file).  And make one single PR in the future to update the registry.
We need to create a release-pipeline to produce the signed binaries in azure-dev and then update the registry from there

PR #6446, opened 2026-01-05
Fix: Context Cancelled Error in Workflow Steps

### LINKED ISSUE #6431: [Issue] context canceled for azd up (but azd deploy works)
**Output from `azd version`**
azd version 1.22.5

**Describe the bug**
azd up fails with 'context canceled' on second deployment step

**To Reproduce**

I'm working with https://github.com/Azure-Samples/azure-ai-travel-agents

```
giulio@myhome:~/csharp/azd-multiagent$azd up
...
...
...
ERROR: error executing step command 'deploy api-llamaindex-ts': failed to authenticate: server response error:
 Get "https://login.microsoftonline.com/common/discovery/instance?api-version=1.1&authorization_endpoint=https%3A%2F%2Flogin.microsoftonline.com%2Forganizations%2Foauth2%2Fv2.0%2Fauthorize": context canceled
giulio@myhome:~/csharp/azd-multiagent$

```

**Expected behavior**

It should not fail indeed `azd deploy` does work fine!

**Environment**
    * ssh session
    * fedora 42
    * bash

**Additional context**

I've already tried clearing auth cache, 

```
rm -rf ~/.azd/auth
rm -rf ~/.azure/accessTokens.json ~/.azure/azureProfile.json

azd auth logout
azd auth login --use-device-code
```

disabling IPv6, and curl to the login endpoint works

My finding that azd deploy works correctly (deploying all services in one go), while azd up fails during the deployment phase, narrows down the problem significantly: the issue is specifically within the `azd up` command's orchestration, particularly in the handoff or state management between its `provision` and subsequent `deploy` stages. It's not a general azd deploy issue, nor a single-service deployment issue, nor a networking issue for individual calls.

wbreza:
Fixes #6431 

## Problem
Multi-step workflows (e.g., [azd up](vscode-file://vscode-app/c:/Users/wabrez/AppData/Local/Programs/Microsoft%20VS%20Code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) deploying multiple services) would fail on subsequent steps with [context canceled](vscode-file://vscode-app/c:/Users/wabrez/AppData/Local/Programs/Microsoft%20VS%20Code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) errors during authentication, even though the context was valid.

## Root Cause
When reusing the same cobra command instance across workflow steps, subcommands retained cancelled contexts from previous executions. Cobra's ExecuteContext() only updates the root command's context, not subcommands, causing middleware to receive stale cancelled contexts.

## Solution
Modified workflowCmdAdapter.ExecuteContext() to recursively set the context on all commands (root and subcommands) before execution, ensuring every command in the tree receives the fresh context.

### DIFF
--- a/cli/azd/cmd/container_test.go
+++ b/cli/azd/cmd/container_test.go
@@ -224,7 +224,7 @@ func Test_WorkflowCmdAdapter_ContextPropagation(t *testing.T) {
 		// not the cancelled context from the first execution
 		select {
 		case <-receivedContexts[1].Done():
-			t.Fatal("Second execution should have received a fresh valid context, not the cancelled one from first execution")
+			t.Fatal("2nd execution should have received a fresh valid context, not the cancelled one from 1st execution")
 		default:
 			// Expected: second context is valid
 		}


cli/azd/cmd/container.go
```diff
@@ -911,9 +911,29 @@ func (w *workflowCmdAdapter) SetArgs(args []string) {
 // ExecuteContext implements workflow.AzdCommandRunner
 func (w *workflowCmdAdapter) ExecuteContext(ctx context.Context) error {
 	childCtx := middleware.WithChildAction(ctx)
+
+	// CRITICAL FIX: Explicitly set the context on the command before calling ExecuteContext.
+	// When the same command instance is reused across workflow steps, cobra may retain
```
vhvb1989, line 916: Im curious about this. Should a command instance be re-used?   What does it mean?
Is it like running `azd deploy foo` and then `azd deploy bar` re-using `azd deploy` ?   Should it be a new instance of the command instead of the same?

vhvb1989: If it is expected to use the same instance, then should we not cancel the context for that command?

I am confused about what are we cleaning up by calling context -> cancel() and what we are keeping around to re-use by re-seting a context.

weikanglim (approved):
LGTM

cli/azd/cmd/container.go
wbreza: We need to cancel after each step... Some steps/command register event handlers that need to be cleaned up so they don't run in future steps.  Event handlers are registered with a Go context and when the context is canceled those handlers are automatically unregistered.

PR #6471, opened 2026-01-07
WIP: Extensions Example recorder test

wbreza:
This is an example test that uses the test recorder framework.  

**TLDR**: Test recorder not recommended for `azd` extensions right now

This test example will capture `azd` internal HTTP calls, but will not capture any custom HTTP calls that happen within the extension process.

Reasons:
- Test recorder requires integrated code changes via build tags within compiled binary to trace HTTP calls
- All extension would need manually implement the same requirements to participate

Recommended:
- Extension author their own manual integration tests that can be run in Azure DevOps / Github on a scheduled/ad-hoc basis.

### DIFF
--- a/cli/azd/test/functional/extension_microsoft_azd_demo_test.go
+++ b/cli/azd/test/functional/extension_microsoft_azd_demo_test.go
@@ -1,0 +1,134 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package cli_test
+
+import (
+	"os"
+	"testing"
+
+	"github.com/azure/azure-dev/cli/azd/test/azdcli"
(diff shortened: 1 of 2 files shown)


weikanglim:
> Test recorder requires integrated code changes via build tags within compiled binary to trace HTTP calls

This isn't necessary. Setting `HTTPS_PROXY` would work. However, one would have to disable TLS verification since the proxy is handling HTTPS traffic. It's possible to do this as a runtime setting or env var to remove the compile-time build flag requirement.

It's also entirely possible to use the recording library in-proc, and that may be worthwhile doing.

I'm unsure what we're aiming for yet, if we're trying to test azd + extension integration, or just the extension piece. Happy to provide more targeted options once I grok the scenario here.

PR #6474, opened 2026-01-08
Replace DAC due to - [SM05138] Detect unsafe use of DefaultAzureCredential

vhvb1989:
fix: https://github.com/Azure/azure-dev/issues/6475

### DIFF
--- a/cli/azd/extensions/microsoft.azd.extensions/CHANGELOG.md
+++ b/cli/azd/extensions/microsoft.azd.extensions/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## 0.8.0 (2026-01-08)
 
-- Remove Default Azure Credential from JS and Python scaffolding
+- [[#6474]](https://github.com/Azure/azure-dev/pull/6474) Remove Default Azure Credential from JS and Python scaffolding
 
 ## 0.7.1 (2025-12-17)
 


cli/azd/extensions/microsoft.azd.extensions/internal/resources/languages/javascript/commands/prompt.js
```diff
@@ -87,7 +87,7 @@ function createPromptCommand() {
     logger.info('Subscription selected', { subscription_id });
     logger.info('Tenant', { tenant_id });
 
-    const credential = new DefaultAzureCredential();
+    const credential = new AzureDeveloperCliCredential({ tenantId: tenant_id });
```
scottaddie, line 90: Is there ever a need to run this code in Azure? If so, using DAC with the `AZURE_TOKEN_CREDENTIALS` env var is what I'd recommend.

vhvb1989: We have one S365 alert about using DAC and we don't really need it. We are more than OK using azd's credential.
This is scaffolding code to create a new azd extension

hemarina (approved):

rajeshkamal5050 (approved):

PR #6491, opened 2026-01-10
Apply azdo ext best practices for testing and fix mocks

vhvb1989:
Improves Azure DevOps extension testing by applying best practices and fixing mocks to work on both Windows and Linux (previously only compatible with Linux).

### DIFF
--- a/ext/azuredevops/setupAzd/tests/_suite.ts
+++ b/ext/azuredevops/setupAzd/tests/_suite.ts
@@ -1,0 +1,64 @@
+import * as path from 'path';
+import * as assert from 'assert';
+import * as ttm from 'azure-pipelines-task-lib/mock-test';
+
+describe('Setup azd task tests', function () {
+
+    before(function() {
+        // Setup before tests
+    });
+
+    after(() => {
+        // Cleanup after tests
+    });
+
+    it('should succeed with default version (latest)', function(done: Mocha.Done) {
+        this.timeout(5000);
+
+        const tp: string = path.join(__dirname, 'success.js');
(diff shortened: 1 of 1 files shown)


ext/azuredevops/setupAzd/index.ts
```diff
@@ -16,25 +16,51 @@ export async function runMain(): Promise<void> {
...
+                task.setResult(task.TaskResult.Failed, `azd version check failed. Exit code: ${versionResult}`)
+                return
+            }
+        } else {
+            const bashPath = task.which('bash', true)
+            const bash: toolRunner.ToolRunner = task.tool(bashPath)
+            bash.arg('-c')
+            bash.arg(`curl -fsSL https://aka.ms/install-azd.sh | sudo bash -s -- --version ${version} --verbose`)
+            
+            const installResult = await bash.exec()
```
weikanglim, line 56: Is `exec()` deprecated in favor of `execAsync`? I pulled the source [docs](https://github.com/microsoft/azure-pipelines-task-lib/blob/f5cfa1678b153fbc2c2cc85929600eb5ba260dbd/node/toolrunner.ts#L1211)

```
    /**
     * Exec a tool.
     * Output will be streamed to the live console.
     * Returns promise with return code
     *
     * @deprecated Use the `execAsync` method that returns a native Javascript promise instead
     * @param     tool     path to tool to exec
     * @param     options  optional exec options.  See IExecOptions
     * @returns   number
     */
```

I do like the new streaming to live console -- that is a pretty neat improvement!

tg-msft (approved):

Issue #6097, opened 2025-11-06
tracing: capture command path for extensions


weikanglim:
When an extension command is invoked, our usage data does not currently capture the command tree within the extension, since azd doesn't understand if the extension binary has subcommands, or is a simple argument command, as well as general privacy concerns.
 
To make this happen, we could consider:
1. Hashing arguments that were passed on to the extension binary (which requires de-hashing later on).
2. Leverage metadata that we know about the extension and emit arguments that are "safe". For example, we could leverage examples.usages in the [schema](https://github.com/Azure/azure-dev/blob/a6dfa6df80205f8e416b5038f5818ca83dced28b/cli/azd/extensions/azure.ai.agents/extension.yaml#L22) that could work for this, or we could add a separate attribute for command paths.
3. Have the extension call back to azd to report its known command paths.

weikanglim:
Of the preferred options, **option 2** seems the simplest to me.  

**Option 3** is also feasible if we want extension authors to explicitly send the invoked command path back to `azd`’s gRPC server during execution.

While looking into option 2, I wasn’t entirely sure which source of truth to rely on. In `~/.azd/config.json`, we currently have the following metadata — which does not include `examples`:

```json
"installed": {
  "azure.coding-agent": {
    "capabilities": [
      "custom-commands"
    ],
    "description": "This extension configures GitHub Copilot Coding Agent access to Azure",
    "displayName": "Coding agent configuration extension",
    "id": "azure.coding-agent",
    "namespace": "coding-agent",
    "path": "extensions/azure.coding-agent/azure-coding-agent-darwin-arm64",
    "source": "azd",
    "usage": "azd coding-agent <command> [options]",
    "version": "0.5.1"
  }
}
```

The `examples.usages` field does appear in `~/.azd/extensions/azure.coding-agent/extension.yaml`, but I wasn’t sure if that’s something we can rely on.  

It might be cleaner to introduce a dedicated `usages` array listing possible command paths, rather than leveraging `examples.usages`. We could also consider deprecating the existing `usage` field, since commands with subcommands would naturally define those under the root of their command tree.

PR #6496, opened 2026-01-12
Adds extension metadata capability

wbreza:
### Summary
Adds metadata command capability to the extension framework, enabling extensions to expose their command structure, configuration schemas, and documentation in a machine-readable JSON format. This enables AI agents and tooling to discover and understand extension capabilities dynamically.

### Key Features

**Metadata Command**
- Extensions implement a `metadata` command that outputs structured JSON
- Automatic generation via `azdext.GenerateExtensionMetadata()` helper
- Captures commands, flags, arguments, configuration schemas, and documentation
- Cached locally after installation for fast access

**Architecture**
- `MetadataCapability` marks extensions supporting metadata retrieval
- Manager fetches metadata during installation (10s timeout, non-blocking)
- Uses lazy.Lazy[*Runner] pattern to resolve DI scoping (Singleton Manager → Scoped Runner)
- Graceful fallback if metadata command unavailable

**Metadata Schema**
```json
{
  "schemaVersion": "1.0",
  "id": "extension.id",
  "commands": [...],
  "configuration": {
    "global": {...},
    "project": {...},
    "service": {...}
  }
}
```

**Configuration Schemas**
- Extensions define typed config structures with JSON Schema tags
- `invopop/jsonschema` generates schemas from Go types
- Three levels: global, project, service
- Supports nested namespacing to avoid conflicts

### DIFF
--- a/cli/azd/pkg/extensions/metadata.go
+++ b/cli/azd/pkg/extensions/metadata.go
@@ -31,7 +31,7 @@ type Command struct {
 	Usage string `json:"usage,omitempty"`
 	// Examples contains example usages of the command
 	Examples []CommandExample `json:"examples,omitempty"`
-	// Args defines the positional arguments accepted by the command
+	// Args defines the positional arguments accepted by the command, in the order it is received
 	Args []Argument `json:"args,omitempty"`
 	// Flags defines the flags/options accepted by the command
 	Flags []Flag `json:"flags,omitempty"`


cli/azd/pkg/azdext/metadata_generator.go
JeffreyCA, line 1: The metadata generator seems to generate an empty `null` command:

Demo extension:
```json
{
  "schemaVersion": "1.0",
  "id": "microsoft.azd.demo",
  "commands": [
    {
      "name": null,
      "short": "",
      "usage": "azd ",
      "hidden": true
    },
```

Agents extension:
```json
{
  "schemaVersion": "1.0",
  "id": "microsoft.azd.demo",
  "commands": [
    {
      "name": null,
      "short": "",
      "usage": "agent ",
      "hidden": true
    },
```

cli/azd/pkg/extensions/manager.go
```diff
@@ -777,3 +791,146 @@ func copyFile(src, dst string) error {
...
+// Returns nil error if metadata was successfully fetched and cached, or if extension doesn't support metadata
+// Returns warning-level error if metadata fetch failed (installation should still succeed)
+func (m *Manager) fetchAndCacheMetadata(
+	ctx context.Context,
+	extension *Extension,
+) error {
+	// Check if extension has metadata capability
+	if !extension.HasCapability(MetadataCapability) {
+		return nil // Extension doesn't support metadata - this is fine
+	}
```
JeffreyCA: Should any of the other methods have capability checks too?

cli/azd/pkg/extensions/manager.go
```diff
@@ -777,3 +791,146 @@ func copyFile(src, dst string) error {
...
+
+const (
+	metadataFileName    = "metadata.json"
+	metadataCommandName = "metadata"
+	metadataTimeout     = 10 * time.Second
+)
+
+// fetchAndCacheMetadata fetches metadata from an extension and caches it to disk
+// Returns nil error if metadata was successfully fetched and cached, or if extension doesn't support metadata
+// Returns warning-level error if metadata fetch failed (installation should still succeed)
```
weikanglim: `// Returns warning-level error if metadata fetch failed (installation should still succeed)`

This comment seems a little off -- I believe we simply return an error if any operations failed

cli/azd/pkg/extensions/manager.go
weikanglim: It seems cleaner to have the caller of `fetchAndCacheMetadata` check for the capability i.e.

```golang
if extension.HasCapability(MetadataCapability) {
     if err := m.fetchAndCacheMetadata(ctx, installedExtension); err != nil {
		// Log warning but don't fail installation
		log.Printf("Warning: Failed to fetch extension metadata for '%s': %v\n", extension.Id, err)
	}
}
```

I like this version more, the responsibility separates more cleanly and the functions compose better

cli/azd/pkg/extensions/metadata.go
```diff
@@ -0,0 +1,143 @@
...
+	Name []string `json:"name"`
+	// Short is a brief one-line description of the command
+	Short string `json:"short"`
+	// Long is an optional detailed multi-line description (markdown supported)
+	Long string `json:"long,omitempty"`
+	// Usage is an optional usage template string
+	Usage string `json:"usage,omitempty"`
+	// Examples contains example usages of the command
+	Examples []CommandExample `json:"examples,omitempty"`
+	// Args defines the positional arguments accepted by the command
```
weikanglim:
```suggestion
	// Args defines the positional arguments accepted by the command, in the order it is received
```

(if I'm understanding this correctly)

cli/azd/pkg/extensions/metadata.go
```diff
@@ -0,0 +1,143 @@
...
+
+// Command represents a command or subcommand in the extension's command tree
+type Command struct {
+	// Name is the command path as an array of strings (e.g., ["demo", "context"])
+	Name []string `json:"name"`
+	// Short is a brief one-line description of the command
+	Short string `json:"short"`
+	// Long is an optional detailed multi-line description (markdown supported)
+	Long string `json:"long,omitempty"`
+	// Usage is an optional usage template string
```
weikanglim, line 30: `// Usage is an optional usage template string`
Is `Usage` a template string?

weikanglim (approved):
Overall looks good! I believe this will nicely help us towards resolving #6097

cli/azd/pkg/extensions/metadata.go
wbreza: This comes from the "use" line in cobra for example.

cli/azd/pkg/extensions/manager.go
wbreza: The other new funcs aren't in use but i would expect the caller to check the extension capability before calling to load metadata for example.

PR #6499, opened 2026-01-12
make azd init alpha feature more discoverable

### LINKED ISSUE #6488: Make agentic azd init experimental feature more discoverable
Make agentic azd init more discoverable so that I can discover and try experimental AI features without needing to know about hidden alpha flags.

Background
Currently, the LLM/AI agent features are hidden behind the alpha.llm configuration flag. To use the feature, users must:

- discover the flag exists in azd main menu
- run `azd config set alpha.llm on`
- then run azd init to use AI-assisted initialization

This creates a discoverability problem—users who would benefit from these features don't know they exist.

Proposed solution
Add option in the init flow to discover agentic init. When a customer runs `azd init` and the alpha feature isn't enabled, prompt them to enable it. This surfaces the feature and maintains opt-in behavior for alpha experimental features.

Acceptance Criteria

- [ ] azd init shows option for `Use agent mode (experimental)`
- [ ] When selected without alpha.llm enabled, display clear instructions: `azd config set alpha.llm on` _"To use this feature run `azd config set alpha.llm on` and then run `azd init` again"_
- [ ] When selected with alpha.llm already enabled, proceed directly to agent mode

hemarina:
fix #6488

image
image


cli/azd/cmd/init.go
JeffreyCA, line 1: The new error message looks great. I noticed if I only have `alpha.llm` set, but not `ai.agent.model.type`, I get this error:

Image

Could we make a similar improvement to the err message here and mention that `ai.agent.model.type` needs to be set?

hemarina: The second part of error message is asking user to set ai.agent.model.type with gh-copilot. Do you mean make it clearer like separate the sentence?

spboyer:
related https://github.com/Azure/azure-dev/pull/6390

cli/azd/cmd/init.go
rajeshkamal5050: I think @JeffreyCA is pointing to this one
> ERROR: Default model type has not been set

Should it be something like,
> To use this feature, run `azd config set ai.agent.model.type github-copilot` to set the agent model type.

JeffreyCA: Yes exactly

JeffreyCA (approved):

vhvb1989:
image

@hemarina , what if, instead of returning this error, we tell and ask the user if they want to turn the alpha feature ON to continue? That would save the user the copy/paste - manual invocation.  @kristenwomack @tg-msft

cli/azd/cmd/init.go
```diff
@@ -615,11 +611,14 @@ func promptInitType(console input.Console, ctx context.Context, featuresManager
...
 		return initAppTemplate, nil
 	case 2:
-		// Only return initWithCopilot if the LLM feature is enabled and we have 3 options
+		// Only return initWithAgent if the LLM feature is enabled and we have 3 options
 		if featuresManager.IsEnabled(llm.FeatureLlm) {
 			return initWithAgent, nil
+		} else {
+			return initUnknown, errors.New("To use this feature, run `azd config set alpha.llm on`, " +
+				"set the agent model type with `azd config set ai.agent.model.type github-copilot`, " +
+				"then run `azd init` again.")
```
wbreza: Lets narrow in on expectations based on discussion in standup.  My recommendation.

* If the alpha featue is not already enabled - show message to user "The azd agent features have been enabled to support this new experience. To turn off in the future run `azd config unset...`
* Enable the alpha feature
* Run the experience.

vhvb1989: You might need to set `gihub` as the llm provider in the config as well - that should make the gh-cp config flow to be invoked next. Otherwise the user would see an error right after about they need to set the llm config

vhvb1989:
@hemarina , about:

image


I think @wbreza mentioned that, instead of asking `do you want to enable...` , if the user select `agentic init`, we just assume the confirmation and display the message `llm feature has been turned on for this experice... ..  bla bla bla... run ________ if you want do turn it off`

vhvb1989 (approved):

cli/azd/cmd/init.go
```diff
@@ -615,11 +620,35 @@ func promptInitType(console input.Console, ctx context.Context, featuresManager
...
+				return initUnknown, fmt.Errorf("failed to save config: %w", err)
+			}
+
+			console.Message(ctx, "\nThe azd agent feature has been enabled to support this new experience."+
+				" To turn off in the future run `azd config unset alpha.llm`.")
+
+			err = azdConfig.Set("ai.agent.model.type", "github-copilot")
+			if err != nil {
+				return initUnknown, fmt.Errorf("failed to set ai.agent.model.type config: %w", err)
+			}
```
vhvb1989, line 645: should this be called before calling `configManager.Save(...)` ???   

I don't see a new call to save the config after this, so the setting might be applied only on memory but not persisted?

vhvb1989 (approved):

cli/azd/cmd/init.go
hemarina: Thanks. Nice catch.

cli/azd/cmd/init.go
```diff
@@ -615,11 +620,35 @@ func promptInitType(console input.Console, ctx context.Context, featuresManager
...
+
```
hemarina, line 646:
```suggestion

            err = configManager.Save(azdConfig)
			if err != nil {
				return initUnknown, fmt.Errorf("failed to save config: %w", err)
			}
			
```

vhvb1989 (approved):

PR #6511, opened 2026-01-14
`azd env remove` for removing local environment files

### LINKED ISSUE #937: [Feature] `azd env delete` command to delete an existing environment
While a `azd env new` command exists to create environments, there's no symmetric command for deleting them.  While users/tooling can remove the environment folder manually, there's related AZD internal state (such as the current default environment setting) that may be impacted and would be best for the `azd` CLI to manage.

weikanglim:
## `azd env remove`

New command to delete local environment config (`.azure/<env>/`).

```bash
azd env remove # remove current
azd env remove <environment>
azd env rm -e <environment>
```

`--force` to skip confirmation. Does not delete Azure resources. Run `azd down` first if needed.

Fixes #937

### DIFF
--- a/cli/azd/cmd/testdata/TestFigSpec.ts
+++ b/cli/azd/cmd/testdata/TestFigSpec.ts
@@ -188,7 +188,7 @@ const completionSpec: Fig.Spec = {
 		},
 		{
 			name: ['ai'],
-			description: 'Extension for Foundry Fine Tuning. (Preview)',
+			description: 'Extension for the Foundry Agent Service. (Preview)',
 			subcommands: [
 				{
 					name: ['agent'],
@@ -1620,7 +1620,7 @@ const completionSpec: Fig.Spec = {
 				},
 				{
 					name: ['ai'],
-					description: 'Extension for Foundry Fine Tuning. (Preview)',
+					description: 'Extension for the Foundry Agent Service. (Preview)',
 					subcommands: [
 						{
 							name: ['agent'],


cli/azd/cmd/env.go
```diff
@@ -1272,6 +1281,18 @@ func (eg *envGetValuesAction) Run(ctx context.Context) (*actions.ActionResult, e
...
+		// No environment specified, and default environment is not selected.
+		// Prompt to choose an environment and set as default.
+		loaded, err := eg.envManager.LoadOrInitInteractive(ctx, name)
+		if err != nil {
+			return nil, err
+		}
+
+		name = loaded.Name()
+	}
+
```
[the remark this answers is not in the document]
weikanglim: Required so that `azd env get-values` doesn't error out and makes it consistent with other places where environment is handled.

cli/azd/cmd/env_remove.go
```diff
@@ -0,0 +1,170 @@
...
+		return nil, err
+	}
+
+	if er.flags.EnvironmentName != "" {
+		name = er.flags.EnvironmentName
+	}
+
+	if name == "" {
+		return nil, fmt.Errorf("no environment specified")
+	}
```
vhvb1989, line 132: nit -> should this be a panic?  What should a user do after this error which should probably never happen

cli/azd/cmd/env_remove.go
```diff
@@ -0,0 +1,170 @@
...
+		return nil, err
+	}
+
+	idx := slices.IndexFunc(envs, func(env *environment.Description) bool {
+		return env.Name == name
+	})
+
+	if idx < 0 {
+		return nil, fmt.Errorf("environment '%s' does not exist", name)
+	}
```
vhvb1989, line 145: curious why you don't used `envManager.Get()` here?

cli/azd/cmd/env_remove.go
```diff
@@ -0,0 +1,170 @@
...
+
+	env := envs[idx]
+	if !er.flags.force {
+		confirm, err := er.console.Confirm(
+			ctx,
+			input.ConsoleOptions{
+				Message: fmt.Sprintf(
+					"Remove the environment '%s'?", env.Name),
```
vhvb1989, line 153: Consider enhancing this message to communicate that:
- environment can't be recovered/undeleted after this.
- Any resource in Azure linked to this env is not deleted

weikanglim: I get the suggestion. Does the message printout above feel sufficient? 

Image

My UX preference here is for prompt messages to be short and concise. 

Something like "Remove the environment 'dev'? This will not delete any Azure resources" may dilute the prompt confirmation, but I'm happy to hear about any examples you may have in mind.

cli/azd/cmd/env_remove.go
weikanglim: There isn’t much deep reasoning here, but conceptually I prefer using `List()` when calling `Remove()`. Using `Get()` would usually imply fetching the full item, which isn’t necessary.

A good analogy is deleting a file from a filesystem: you wouldn’t `Open` or `Read` the file first—you’d simply look up the directory entry and remove it.

In this case, the code doesn’t need to load the full environment, only the name, so `List()` feels like the more appropriate behavior.

cli/azd/cmd/env_remove.go
weikanglim: It does occur in practice currently -- when user calls `remove()` without previously creating any environments.

wbreza: I don't think it should panic - error is good here.  We should really "never" force a panic unless it is like a dev time requirement.

cli/azd/cmd/env_remove.go
wbreza: +1 for just using envManager to get the environment.  it does the lookup internally.

cli/azd/cmd/env_remove.go
wbreza: I would consider doing a check to see if this environment does contain any provisioned resources and if so warning the user.  I agree that this command shouldn't remove them, but should let the user know that there will be orphaned resources that will need to be manually cleaned up.

cli/azd/proposal.md
```diff
@@ -0,0 +1,236 @@
+# Feature Request: Revision-aware deployments and rollback support for Azure Container Apps
```
wbreza: Is this intended to be part of this PR?

cli/azd/cmd/env_remove.go
weikanglim: See my comment above.

cli/azd/cmd/env_remove.go
weikanglim: I agree that would be a nicer user experience as well. Immediately, it may be somewhat tricky to land.

The naive approach of calling `State()` takes a significant amount of time, since it involves compiling the infra, querying the deployment state, as I had [noted](https://github.com/Azure/azure-dev/issues/937#issuecomment-1942907860) in the issue. The intention is to indeed serve the notice, but the notice itself takes a long time to get around to, which in my mind degrades the perceived user experience.

Longer term, we may add the new confirmation prompt / warning. My vote is this would be a non-blocking future improvement. I logged #6522 as a placeholder issue to explore storing state to enable this scenario.

Note that deleting the resources after an "accident removal" using `azd` is still completely possible with `azd env new` with the same name, and choosing the subscription + location, before running `azd down`. This used to prompt the user for all parameters, but has been cleaned up with #6345, so I do like the overall story we are telling here today.

cli/azd/proposal.md
weikanglim: it snuck in  🦝

cli/azd/cmd/env_remove.go
vhvb1989: I would only add the `this action is irreversible` to the wording.

And I would consider a variation to what Wallace mentioned. Instead of checking if the env was provisioned before. I would only look at the number of keys defined in the .env and the keys in the config.json and mention that in the wording with a warning when the number of keys is greater than 3 (assuming location, sub and env-name the 3 based keys) , or if the number of keys after removing the known keys is greater than 0.
Then I would print: 

"This action will delete <key number> permanently [and <params-num-from-config> saved parameters. This will be irreversible"

Those numbers should be super fast to get

vhvb1989 (approved):

PR #6512, opened 2026-01-14
Extension Update Warning

wbreza:
## Extension Update Warning

When users run an extension command, azd now checks if a newer version is available and displays a warning after execution with upgrade instructions.

### Key Features
- **Zero latency impact**: Update check runs in parallel with extension execution via goroutine
- **Per-source caching**: Registry data cached at `~/.azd/cache/extensions/<source>.json` with 4hr TTL (configurable via `AZD_EXTENSION_CACHE_TTL`)
- **24hr warning cooldown**: Stored in `Extension.LastUpdateWarning` field - auto-cleaned on uninstall
- **Suppressed for JSON output**: No warning when `--output json` is used

### Example Output

image

### DIFF
--- a/cli/azd/cmd/extensions.go
+++ b/cli/azd/cmd/extensions.go
@@ -160,7 +160,22 @@ func (a *extensionAction) Run(ctx context.Context) (*actions.ActionResult, error
 	showUpdateWarning := !isJsonOutputFromArgs(os.Args)
 	if showUpdateWarning {
 		updateResultChan := make(chan *updateCheckOutcome, 1)
-		go a.checkForUpdateAsync(ctx, extension, updateResultChan)
+		// Create a shallow copy of extension data for the goroutine to avoid race condition.
+		// The goroutine mutates LastUpdateWarning, so we copy only the needed fields.
+		// Cannot copy the full Extension due to sync.Once (contains sync.noCopy).
+		extForCheck := &extensions.Extension{
+			Id:                extension.Id,
+			Namespace:         extension.Namespace,
+			DisplayName:       extension.DisplayName,
+			Description:       extension.Description,
+			Version:           extension.Version,
+			Source:            extension.Source,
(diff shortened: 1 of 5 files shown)


cli/azd/pkg/extensions/extension.go
JeffreyCA, line 1: We also discussed clarifying that the Version column in `azd ext list` means latest version - should we also update that here too?

Image

JeffreyCA: I think @tg-msft also suggested an idea around leaving Latest value empty when the user already has the latest installed

wbreza: Will take this as a different changeset... this is only for update warning right now.

cli/azd/cmd/extensions.go
```diff
@@ -18,11 +19,27 @@ import (
...
+		if arg == "--output" || arg == "-o" {
+			if i+1 < len(args) && args[i+1] == "json" {
+				return true
+			}
+		}
+		if arg == "--output=json" || arg == "-o=json" {
+			return true
+		}
+	}
+	return false
```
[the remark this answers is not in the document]
wbreza: Not right now.

cli/azd/cmd/extensions.go
```diff
@@ -135,6 +152,26 @@ func (a *extensionAction) Run(ctx context.Context) (*actions.ActionResult, error
...
+			select {
+			case result := <-updateResultChan:
+				if result != nil && result.shouldShow && result.warning != nil {
+					a.console.MessageUxItem(ctx, result.warning)
+					a.console.Message(ctx, "")
+				}
+			default:
+				// Check didn't complete in time, skip warning
+			}
+		}()
```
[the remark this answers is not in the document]
wbreza: That is fine - if the command executes faster than the update check we will defer showing the message.

PR #6525, opened 2026-01-14
Fix non-deterministic help text when multiple extensions share namespace

JeffreyCA:
This PR fixes non-deterministic help text when multiple extensions share a namespace prefix (e.g., `azure.ai.agents` and `azure.ai.finetune` both using `ai.*`). Previously, the intermediate `azd ai` command would randomly show "Extension for the Foundry Agent Service" or "Extension for Foundry Fine Tuning" if both `azure.ai.agents` and `azure.ai.finetune` extensions were installed.

### Solution
1. Use a generic description for intermediate namespace commands showing the full namespace path: `"Commands for the ai extension namespace."`
2. Sort extensions by ID before binding to ensure deterministic ordering

Extensions with simple namespaces (like `demo` or `x`) still show the actual extension description:

```
$ azd help
...
  Enabled extensions commands 
    ai          : Commands for the ai extension namespace.
    demo        : This extension provides examples of the AZD extension framework.
    x           : This extension provides a set of tools for AZD extension developers to test and debug their extensions.
```

Deeply nested namespaces should also work, so if we had 2 extensions that had namespaces like `ai.models.a` and `ai.models.b`:

```
$ azd ai help
...
  Available Commands
    models    : Commands for the ai.models extension namespace.
```

### DIFF
--- a/cli/azd/cmd/extensions.go
+++ b/cli/azd/cmd/extensions.go
@@ -50,9 +50,12 @@ func bindExtension(
 
 		// If not found, create a new command
 		if !found {
+			// Build the full namespace path up to this point for the description
+			namespacePath := strings.Join(namespaceParts[:i+1], ".")
+			description := fmt.Sprintf("Commands for the %s extension namespace.", namespacePath)
 			cmd := &cobra.Command{
 				Use:   part,
-				Short: extension.Description,
+				Short: description,
 			}
 
 			current = current.Add(part, &actions.ActionDescriptorOptions{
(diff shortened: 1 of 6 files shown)


weikanglim (approved):
LGTM! Both the deterministic sorting and the auto-generated description looks good to me.

I could imagine making namespace descriptions configurable as a future improvement, but I'm overall indifferent towards either dierction. I believe perhaps @wbreza has the best intuition here on what makes the most sense long term.

vhvb1989 (approved):

PR #6536, opened 2026-01-15
Fixes context cancellation issue

### LINKED ISSUE #6530: [Issue] `context canceled` error when attempting to download Bicep during fresh `azd up`

**Output from `azd version`**
azd version 1.23.0 (commit 9a02add5930c2c87b6b2c78443e4d8132587417d)

**Describe the bug**
On a newly initialised install of `azd`, running `azd up` produces a `context canceled` error when attempting to install Bicep immediately after logging in. This happens consistently.

**To Reproduce**
1) `rm -rf ~/.azd`
2) Create new azd project, e.g. `azd init --template todo-python-mongo`
3) `azd up`
4) After logging in, the following error is seen:

```console
ERROR: initializing infrastructure provider: failed resolving IaC provider 'bicep': downloading bicep: Get "https://downloads.bicep.azure.com/v0.39.26/bicep-linux-x64": context canceled
```

**Expected behavior**
After logging in, `bicep` is downloaded and the install begins.

**Environment**
Ubuntu 24.04.1 LTS (WSL2, Windows 11)

wbreza:
Fixes #6530

### Problem
Singletons (like CLI tool wrappers) were capturing step-scoped contexts that get cancelled after each command step completes, causing "context canceled" errors in subsequent operations (e.g., deploy failing after provision).

### Solution
Three-layer fix to prevent stale contexts from breaking singleton behavior:

1. **main.go**: Wrap root context with `context.WithoutCancel()` to protect the entire command tree from external cancellation signals

2. **EnsureInstalled pattern for CLI tools** (`bicep`, `github`, `pack`):
   - Constructor takes no context, returns `*Cli`
   - New `EnsureInstalled(ctx)` method does I/O operations with `sync.Once` for thread-safe one-time initialization
   - Callers must explicitly call `EnsureInstalled` before first use

3. **root.go**: Clear context from cobra command tree in `PersistentPostRunE` to prevent stale context references

### DIFF
--- a/cli/azd/pkg/infra/provisioning/bicep/bicep_provider_test.go
+++ b/cli/azd/pkg/infra/provisioning/bicep/bicep_provider_test.go
@@ -950,6 +950,7 @@ func TestUserDefinedTypes(t *testing.T) {
 
 	azCli := mockazapi.NewAzureClientFromMockContext(mockContext)
 	bicepCli := bicep.NewCli(mockContext.Console, mockContext.CommandRunner)
+	require.NoError(t, bicepCli.EnsureInstalled(*mockContext.Context))
 	env := environment.NewWithValues("test-env", map[string]string{})
 
 	mockContext.CommandRunner.When(func(args exec.RunArgs, command string) bool {


cli/azd/cmd/container.go
```diff
@@ -919,25 +919,27 @@ func (w *workflowCmdAdapter) SetArgs(args []string) {
...
-	// from previous executions when commands are reused.
-	cmd.SetContext(ctx)
+// clearContextRecursively clears the context on a command and all its subcommands recursively
+// to remove any stale cancelled contexts from previous workflow step executions.
+func clearContextRecursively(cmd *cobra.Command) {
+	// Set context to TODO() to clear any stale context. cobra.Command.Context() returns
+	// context.Background() if no context is set, so SetContext is safe to call on any command.
+	// The actual context will be set by ExecuteContext on the root command and propagated
+	// to the executed subcommand by cobra's internal execution flow.
+	cmd.SetContext(context.TODO())
```
vhvb1989: Sharing what this makes me think (with some AI help):

The solution relies on:

- Cobra's internal context propagation in ExecuteC)
- The fact that `context.TODO` is not nil (cobra checks for nil)
- Singletons being resolved during command execution (not after cancellation)

This is fragile because:

- It's not explicit about which context should be used where
- Future refactoring might break these assumptions
- Different parts of the code treat context differently (some as DI, some for cancellation)

Context Semantic Confusion

The codebase conflates two uses of context:

- Cancellation signal - for stopping operations
- Dependency injection - for passing context-based services

----------------

Here an interesting finding too:

Why are we using a step context to resolve singletons (which should be using the main root context):

Looking at the comment in
https://github.com/Azure/azure-dev/blob/d5572874ca1d6bf5476ac38431c0aba39117acd4/cli/azd/cmd/cobra_builder.go#L96-L99

The comment says "root go context" but the code doesn't actually ensure it's the root context—it just grabs whatever `cmd.Context()` returns.

This was probably written when:

- Commands were only executed once (no workflow reuse)
- The context from `main.go` flowed straight through to the command
- There was no `context.WithCancel` wrapping in the workflow runner

In that original scenario, `cmd.Context()` would return the long-lived context.Background() from main.go, and everything would work fine.

Summary
It appears to be an `architectural oversight` that became a bug when workflow support was added. The current fix (clearContextRecursively) is a workaround that happens to work, but doesn't address the root cause: singletons shouldn't capture construction-time contexts at all.

wbreza: Agree - this is a workaround to resolve this issue right now.  There is an opportunity to make a larger change to address this from an architectural point of view based on addition to the workflow runner++.

PR #6566, opened 2026-01-21
add purge support for log analytics

### LINKED ISSUE #5080: [Issue] Logging doesn't always appear in App Insights and Log Analytics after taking down an environment and redeploying it
- [x ] Make sure you've installed the latest version using [instructions in the wiki](../wiki/install)

**Output from `azd version`**
azd version 1.14.0 (commit c928795c47f27d1e997c217147dc649054ac05c8)

**Describe the bug**
Logging doesn't always appear in App Insights and Log Analytics after taking down an environment and redeploying it with the same resource names. In short:
- Requests and traces sometimes fail to show/get logged in App Insights & Log Analytics after redeployment.
- Metrics continue to appear in App Insights.
- Live Metrics still displays the data.
- Sampling is disabled, and I've verified that the connection strings are correct.

This might have something to do with the soft-delete feature of Log Analytics. So, I'm wondering if `azd` should permanently delete the Log Analytics Workspace when e.g. the `--purge` argument is specified in the `azd down` command?

_See additional context for more details_

**To Reproduce**
Because it doesn't always happen, this issue is difficult to reproduce. You can follow these steps to try and reproduce:

1. Deploy [Azure Integration Services Quickstart](https://github.com/ronaldbosma/azure-integration-services-quickstart) with `azd up`. Set all optional parameters to `true` on the first deployment.
2. Execute the [test scenario](https://github.com/ronaldbosma/azure-integration-services-quickstart#test).
3. Check that request and trace logging is visible in App Insights (this might take a few minutes).
4. Remove the environment with `azd down --purge`.
5. Redeploy the environment following the previous steps until the request and trace logging fails to appear in App Insights. 
 
Note that this might take a few redeployments before the issue occurs. When I tried to reproduce myself, I did this multiple times and logging kept working. When I redeployed after a few days, the issue occured.

**Expected behavior**
Request and trace logging should always appear in App Insights and Log Analytics after redeployment.

Although this seems to be more of an issue with App Insights / Log Analytics itself. Permanently removing the Log Analytics Workspace when executing `azd down` might prevent it from occurring.

**Environment**
Bicep CLI version 0.34.44 (1f661a23c4)

**Additional context**
I was working on my azd template [Azure Integration Services Quickstart](https://github.com/ronaldbosma/azure-integration-services-quickstart). The template deploys App Insights with a Log Analytics Workspace. A Function App, Logic App, and API Management service are connected to App Insights. I noticed that when I use the same resource names and redeploy the environment multiple times (using `azd up` and `azd down`), request and trace logging doesn't always work.

When I run my [test scenario](https://github.com/ronaldbosma/azure-integration-services-quickstart#test) there should be request and trace logging in App Insights from my Function App, Logic App, and API Management service. The first time I deploy a new environment, this works fine. However, after I redeploy it, I consistently see the metrics coming in, but sometimes the requests and traces do not show up in Application Insights & Log Analytics. When I use the Live Metrics feature, I do see that App Insights receives the requests and traces. It's almost like App Insights can't store them in the Log Analytics Workspace.

When logging failed, I manually removed the Log Analytics Workspace (not permanently!) and ran `azd provision` again. I executed the test scenario again and the logging worked.

hemarina:
Fix #5080 

This pull request adds support for purging Azure Log Analytics Workspaces during resource group deletion in the Bicep infrastructure provider. The main changes include introducing new methods for identifying and purging Log Analytics Workspaces, updating the destroy logic to handle these resources, and updating dependencies to support the new functionality.

**Log Analytics Workspace Purge Support:**

* Added `AzCliLogAnalyticsWorkspace` struct and related methods to `azapi/log_analytics.go` for retrieving and purging Log Analytics Workspaces using the Azure SDK.
* Implemented `getLogAnalyticsWorkspacesToPurge` and `purgeLogAnalyticsWorkspaces` methods in `bicep_provider.go` to identify and purge Log Analytics Workspaces before deleting the resource group. [[1]](diffhunk://#diff-2c22edb40a26b0cfa31df98cdaf6a05bb96719d424366e53c088792e50714754R1598-R1624) [[2]](diffhunk://#diff-2c22edb40a26b0cfa31df98cdaf6a05bb96719d424366e53c088792e50714754R1671-R1688)
* Updated `destroyDeployment` logic to handle Log Analytics Workspace purge when the purge option is enabled, ensuring these resources are purged before resource group deletion.
* Modified `Destroy` method to collect Log Analytics Workspaces to purge and pass them to the destroy logic. [[1]](diffhunk://#diff-2c22edb40a26b0cfa31df98cdaf6a05bb96719d424366e53c088792e50714754L974-R974) [[2]](diffhunk://#diff-2c22edb40a26b0cfa31df98cdaf6a05bb96719d424366e53c088792e50714754R1003-R1007) [[3]](diffhunk://#diff-2c22edb40a26b0cfa31df98cdaf6a05bb96719d424366e53c088792e50714754L1012-R1017)

**Dependency Updates:**

* Updated `go.mod` to add the Azure Operational Insights SDK and other dependencies required for Log Analytics Workspace management, as well as version bumps for several indirect dependencies. [[1]](diffhunk://#diff-ee9d4acfd4a8a1a9e8260d2991c50464a724c043c8a1b3afcf69b80cfc8f5904R22) [[2]](diffhunk://#diff-ee9d4acfd4a8a1a9e8260d2991c50464a724c043c8a1b3afcf69b80cfc8f5904R53) [[3]](diffhunk://#diff-ee9d4acfd4a8a1a9e8260d2991c50464a724c043c8a1b3afcf69b80cfc8f5904L81-R83) [[4]](diffhunk://#diff-ee9d4acfd4a8a1a9e8260d2991c50464a724c043c8a1b3afcf69b80cfc8f5904L116) [[5]](diffhunk://#diff-ee9d4acfd4a8a1a9e8260d2991c50464a724c043c8a1b3afcf69b80cfc8f5904L158-R164)

Checked in log analytics workspace recycle bin that the workspace is purged. 
image


cli/azd/pkg/azapi/log_analytics.go
```diff
@@ -0,0 +1,84 @@
...
+	workspacesClient, err := cli.createLogAnalyticsWorkspacesClient(ctx, subscriptionId)
+	if err != nil {
+		return err
+	}
+
+	deleteOpts := &armoperationalinsights.WorkspacesClientBeginDeleteOptions{
+		Force: to.Ptr(true),
+	}
+
+	poller, err := workspacesClient.BeginDelete(ctx, resourceGroupName, workspaceName, deleteOpts)
```
vhvb1989, line 58: Add documentation to this method to explain why delete+force is the way to purge deleted workspaces.
Maybe with a link to: https://learn.microsoft.com/rest/api/loganalytics/workspaces/delete?view=rest-loganalytics-2025-07-01&tabs=HTTP

cli/azd/pkg/infra/provisioning/bicep/bicep_provider.go
```diff
@@ -971,7 +971,7 @@ func (p *BicepProvider) Destroy(
 	if len(groupedResources) == 0 {
 		p.console.StopSpinner(ctx, "", input.StepDone)
 		// Call deployment.Delete to void the state even though there are no resources to delete
-		if err := p.destroyDeployment(ctx, deploymentToDelete); err != nil {
+		if err := p.destroyDeployment(ctx, options, deploymentToDelete, []*azapi.AzCliLogAnalyticsWorkspace{}); err != nil {
```
vhvb1989: Why are we using `destroyDeployment()` to force-delete the workplaces? 

I would consider renaming the function to purge workspaces to `forceDelete()` - make it unique for workspaces.

Then, after getting `groupedResources` and `resourcesToDelete`, check or find if `purge is true` and if yes, check/get any workspace and call forceDelete. This would make this step to run before deleting the resource-groups, directly to start with force-delete for workspaces.

Then let the current flow to continue and find things to purge and delete the rg - then purge.

Doing this we don't need to delete the workspaces twice when purge is true.
We can use this special way to purge workspaces with force delete since the beginning instead of doing soft-delete first and then force-delete to purge.

vhvb1989 (approved):

JeffreyCA (approved):

rajeshkamal5050 (approved):

Issue #6591, opened 2026-01-22
update vscode-extension to fix security alert


vhvb1989:
fix: https://github.com/Azure/azure-dev/security/dependabot/229

PR #6592, opened 2026-01-22
fix security issue with playwright/test 1.49.1

vhvb1989:
fix: https://github.com/Azure/azure-dev/issues/6591


bwateratmsft:
Yeah this wasn't supposed to be in the extension at all. Guess we forgot to `npm i` after removing it.

JeffreyCA (approved):

PR #6611, opened 2026-01-25
azure.ai.agents - Set `CGO_ENABLED=0` for x64 Linux build

JeffreyCA:
This statically links the binary built by the pipeline to increase backwards compatibility with older Linux versions like Debian 11.

image

### DIFF
--- a/cli/azd/extensions/azure.ai.agents/CHANGELOG.md
+++ b/cli/azd/extensions/azure.ai.agents/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Release History
 
+## 0.1.7-preview (2026-01-26)
+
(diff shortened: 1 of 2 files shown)


cli/azd/extensions/azure.ai.agents/ci-build.ps1
```diff
@@ -55,6 +55,13 @@ if ($IsWindows) {
 }
 elseif ($IsLinux) {
     Write-Host "Building for linux"
+    
+    # Disable cgo in the x64 Linux build. This will also statically
+    # link the resulting binary which increases backwards 
+    # compatibility with older versions of Linux.
+    if ($env:GOARCH -ne "arm64") {
+        $env:CGO_ENABLED = "0"
+    }
```
danieljurek, line 64: Not a blocker: I'd check explicitly for `amd64` and then set CGO_ENABLED=0 in that case.

PR #6626, opened 2026-01-28
Adds default 30 second timeout to all prompts

wbreza:
### What Changed
Added automatic 30-second timeout for all interactive prompts in `azd`, configurable via `AZD_PROMPT_TIMEOUT` environment variable.

### Business Justification

**Problem**: In CI/CD pipelines, automated scripts, and coding agent scenarios, `azd` commands can hang indefinitely waiting for user input that will never come - blocking pipelines, wasting compute resources, and causing deployment failures.

**Solution**: Prompts now automatically timeout after 30 seconds (configurable), failing fast with a clear error message and actionable suggestion to set `--no-prompt` or provide required inputs.

### Behavior

| `AZD_PROMPT_TIMEOUT` Value | Behavior |
|---------------------------|----------|
| Not set | **30 second timeout (default)** |
| `0` or negative | Timeout disabled (wait forever) |
| Positive integer | Custom timeout in seconds |
| Invalid (non-numeric) | 30 second default |

### User Experience

When a prompt times out, users see:
image

### Impact

- **CI/CD Reliability**: Pipelines fail fast instead of hanging
- **Coding Agents**: GitHub Copilot workspace and similar tools get clear feedback
- **Resource Efficiency**: No more zombie processes waiting for input
- **Backward Compatible**: Set `AZD_PROMPT_TIMEOUT=0` to restore old behavior

### DIFF
--- a/cli/azd/cmd/middleware/ux.go
+++ b/cli/azd/cmd/middleware/ux.go
@@ -60,7 +60,8 @@ func (m *UxMiddleware) Run(ctx context.Context, next NextFn) (*actions.ActionRes
 		var promptTimeoutErr *uxlib.ErrPromptTimeout
 		if errors.As(err, &promptTimeoutErr) {
 			suggestion := fmt.Sprintf(
-				"\nSuggestion: To run without prompts, use: %s --no-prompt",
+				"\nSuggestion: To run non-interactively without input prompts, use: %s --no-prompt\n"+
+					"To disable prompt timeouts, set the AZD_PROMPT_TIMEOUT environment variable to 0.",
 				m.options.CommandPath,
 			)
 			errorMessage.WriteString(suggestion)
--- a/cli/azd/internal/cmd/errors.go
+++ b/cli/azd/internal/cmd/errors.go
@@ -23,6 +23,7 @@ import (
 	"github.com/azure/azure-dev/cli/azd/pkg/exec"
 	"github.com/azure/azure-dev/cli/azd/pkg/extensions"
 	"github.com/azure/azure-dev/cli/azd/pkg/tools"
+	"github.com/azure/azure-dev/cli/azd/pkg/ux"
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/codes"
 )
@@ -46,8 +47,12 @@ func MapError(err error, span tracing.Span) {
 	// internal errors
 	var errWithSuggestion *internal.ErrorWithSuggestion
 	var loginErr *auth.ReLoginRequiredError
+	var promptTimeoutErr *ux.ErrPromptTimeout
 
-	if errors.As(err, &loginErr) {
+	if errors.As(err, &promptTimeoutErr) {
+		errCode = "user.prompt_timeout"
+		errDetails = append(errDetails, fields.PromptTimeoutSeconds.Int(int(promptTimeoutErr.Duration.Seconds())))
+	} else if errors.As(err, &loginErr) {
 		errCode = "auth.login_required"
 	} else if errors.As(err, &errWithSuggestion) {
 		errCode = "error.suggestion"
--- a/cli/azd/internal/cmd/errors_test.go
+++ b/cli/azd/internal/cmd/errors_test.go
@@ -8,6 +8,7 @@ import (
 	"errors"
 	"net/http"
 	"testing"
+	"time"
 
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
 	"github.com/azure/azure-dev/cli/azd/internal/tracing/fields"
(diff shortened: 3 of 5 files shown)


cli/azd/pkg/ux/multi_select.go
```diff
@@ -129,6 +131,14 @@ func (p *MultiSelect) WithCanvas(canvas Canvas) Visual {
 
 // Ask prompts the user to select an option from a list.
 func (p *MultiSelect) Ask(ctx context.Context) ([]*MultiSelectChoice, error) {
+	// Auto-apply prompt timeout if configured globally
+	var timeout time.Duration
+	if timeout = GetPromptTimeout(); timeout > 0 {
+		var cancel context.CancelFunc
+		ctx, cancel = context.WithTimeout(ctx, timeout)
```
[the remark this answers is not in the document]
wbreza: No thanks.

cli/azd/cmd/middleware/ux.go
```diff
@@ -54,7 +56,15 @@ func (m *UxMiddleware) Run(ctx context.Context, next NextFn) (*actions.ActionRes
 			errorMessage.WriteString(output.WithErrorFormat("\nTraceID: %s", errorWithTraceId.TraceId))
 		}
 
-		if errors.As(err, &suggestionErr) {
+		// Handle prompt timeout by adding a suggestion to use --no-prompt
+		var promptTimeoutErr *uxlib.ErrPromptTimeout
+		if errors.As(err, &promptTimeoutErr) {
+			suggestion := fmt.Sprintf(
+				"\nSuggestion: To run without prompts, use: %s --no-prompt",
```
vhvb1989: should the suggestion be `remove time-out set ENV VAR = -1 ` ?

cli/azd/pkg/ux/config.go
```diff
@@ -0,0 +1,50 @@
...
+// This typically occurs when AZD_PROMPT_TIMEOUT is configured and the user does not respond
+// within the specified duration.
+type ErrPromptTimeout struct {
+	Duration time.Duration
+}
+
+// Error returns the error message
+func (e *ErrPromptTimeout) Error() string {
+	return fmt.Sprintf("prompt timed out after %d seconds", int(e.Duration.Seconds()))
+}
```
JeffreyCA, line 50: Consider mapping this error type in `cli/azd/internal/cmd/errors.go` for better tracing, similar to `ErrorWithSuggestion` or `InterruptErr`

cli/azd/docs/environment-variables.md
```diff
@@ -12,6 +12,7 @@ Environment variables that can be used to configure `azd` behavior, usually set
 - `AZD_DEMO_MODE`: If true, enables demo mode. This hides personal output, such as subscription IDs, from being displayed in output.
 - `AZD_FORCE_TTY`: If true, forces `azd` to write terminal-style output.
 - `AZD_IN_CLOUDSHELL`: If true, `azd` runs with Azure Cloud Shell specific behavior.
+- `AZD_PROMPT_TIMEOUT`: Timeout in seconds for interactive prompts. When set, prompts will timeout after the specified duration. If a default value is available, it will be auto-selected; otherwise, the command fails with a suggestion to use `--no-prompt`. Useful for automated agents (e.g., GitHub Copilot, Claude Code) that cannot respond to interactive prompts. Set to `0` or leave empty to disable (default behavior for human users).
```
vhvb1989, line 15: I would make `AZD_DEMO_MODE` to disable the timeouts.  People might want to pause and explain someone what is happening while azd is waiting in the screen for an input.

cli/azd/pkg/input/prompt_timeout.go
JeffreyCA, line 1: Does the countdown "reset" on user key input? E.g. when the user presses up/down arrow keys on a multiple choice prompt.

cli/azd/cmd/middleware/ux.go
tg-msft: We want the current suggestion for agents but should add the env var workaround for humans.
```suggestion
				"\nSuggestion: To run non-interactively without input prompts, use: %s --no-prompt\nSuggestion: To disable input prompt timeouts, set ...",
```

cli/azd/pkg/input/prompt_timeout.go
tg-msft: We might want to bump the default higher than 30s if we're not able to do this now.

wbreza: Not right now. We could do this on our own `ux` package components, but many areas still use the OSS prompting library that doesn't support any of this.  Choice was to make a change that would apply consistently across all our prompt implementations.

cli/azd/pkg/ux/config.go
wbreza: Since the `ux` lib can be used directly by extension authors I opted to keep this in the `ux` lib so they are able to catch and react to this within an extension as needed.  If it was in our `internal` azd package they would not be able to check this without manually checking error message.

cli/azd/docs/environment-variables.md
wbreza: I have concerns tying this into demo mode... if they enable demo mode and are demoing an agentic scenario then this could cause some conflicts.  Opting to keeping this as a separate env var for now.

PR #6641, opened 2026-01-29
Fix load environment lazily to respect -e flag

lim-dy:
This PR fixes a bug where running `azd deploy -e <env-name>` would ignore the provided environment flag and instead deploy to the currently selected default environment.

**Root Cause:**
The `DeployAction` was requesting the `*environment.Environment` object via constructor injection. The IoC container resolved this dependency eagerly (during application startup) before the command-line flags (specifically `-e`) were fully processed. As a result, the container fell back to the default environment stored in the local context, permanently binding the wrong environment to the action.

**The Fix:**

* Refactored `NewDeployAction` to remove the eager injection of `*environment.Environment`.
* Updated `DeployAction` to hold a reference to `environment.Manager`.
* Moved the environment resolution logic into the `Run` method. The environment is now loaded lazily using `da.flags.EnvironmentName`, ensuring the CLI flag is respected.

#### Testing

**Reproduction Steps (Before Fix):**

1. Initialize a project with two environments: `dev` and `prod`.
2. Select `dev` as default: `azd env select dev`.
3. Run `azd deploy -e prod`.
4. **Result:** Deployment targets `dev` (Incorrect).

**Verification (After Fix):**

1. Initialize a project with two environments: `dev` and `prod`.
2. Select `dev` as default: `azd env select dev`.
3. Run `azd deploy -e prod`.
4. **Result:** Deployment targets `prod` (Correct).

#### Affected Commands

* `azd deploy`

### DIFF
--- a/cli/azd/internal/cmd/deploy.go
+++ b/cli/azd/internal/cmd/deploy.go
@@ -380,7 +380,7 @@ func (da *DeployAction) Run(ctx context.Context) (*actions.ActionResult, error) 
 	// Invalidate cache after successful deploy so azd show will refresh
 	if err := da.envManager.InvalidateEnvCache(ctx, env.Name()); err != nil {
 		log.Printf("warning: failed to invalidate state cache: %v", err)
-	} 
+	}
 
 	return &actions.ActionResult{
 		Message: &actions.ResultMessage{


weikanglim (commented):
Hey @lim-dy ! Thanks for proposing the changes.

Could you file an issue with the repro steps and the specific azd version you may be using? I've tried running the same repro steps with `azd deploy -e dev` as provided, and it seems to be working as intended.

This also agrees with my understanding of how `*environment.Environment` is injected -- it parses the envFlag first before determining which environment to load. For these reasons, I'm holding off on approval for now.

EDIT: @JeffreyCA mentioned a potential bug with container_helper.go that may explain what you're seeing. I submitted #6649 as a potential fix.

weikanglim:
Closing in favor of #6649

Issue #5850, opened 2025-10-03
[Issue] Unclear how to provide array as a parameter


wasker:
- [x] Make sure you've installed the latest version using [instructions](https://github.com/Azure/azure-dev?tab=readme-ov-file#installupgrade-azure-developer-cli)

**Output from `azd version`**
Run `azd version` and copy and paste the output here: azd version 1.19.0 (commit b3d68cea969b2bfbaa7b7fa289424428edb93e97)

**Describe the bug**

Bicep supports `array` as a type of parameter, however, it's unclear how to define that in the parameters file and `azd env set`.

**Expected behavior**

Documentation to show how array type is handled by azd.

Menghua1:
@wasker Bicep supports multiple array types. Here's how to define them:
| Type | main.bicep | main.parameters.json | azd env set |
|------|------------|---------------------|-------------|
| `string[]` | `param regions string[] = ["eastus", "westus"]` | `"regions": { "value": ["eastus", "westus"] }` | `azd env set REGIONS '["eastus","westus"]'` |
| `int[]` | `param ports int[] = [80, 443]` | `"ports": { "value": [80, 443] }` | `azd env set PORTS '[80,443]'` |
| `bool[]` | `param flags bool[] = [true, false]` | `"flags": { "value": [true, false] }` | `azd env set FLAGS '[true,false]'` |
| `object[]` | `param configs object[] = [{"name":"dev"}]` | `"configs": { "value": [{"name":"dev"}] }` | `azd env set CONFIGS '[{"name":"dev"}]'` |
| `array` | `param items array = ["a", 1, true]` | `"items": { "value": ["a", 1, true] }` | `azd env set ITEMS '["a",1,true]'` |

**Note:** `string[]`, `int[]`, `bool[]`, `object[]` enforce element types at compile time; `array` allows mixed types.

More documentation can be found at: https://learn.microsoft.com/en-us/azure/azure-resource-manager/bicep/data-types#arrays.

Cc: @rajeshkamal5050

wasker:
Did you verify the recommendation?

How would `azd env set REGIONS` translate into a change for main.parameters.json?

On the screenshot below you can see that e.g., `environmentName` is set to `"value": "${AZURE_ENV_NAME}"`. From your recommendation `REGIONS` is supposed to somehow replace `"regions"` value in the JSON file, but there's no variable placeholder in the suggested value. The documentation doesn't describe how the placeholder should look like in case of an array.

[...]

PR #6694, opened 2026-02-03
Handle JSON values for array/object params in Bicep

weikanglim:
## Problem

When using environment variables to set array or object parameter values in Bicep templates, envsubst would fail with invalid JSON errors.

**Example failure:**
```bash
azd env set MY_ARRAY '["val1", "val2"]'
```

With parameter definition: `"arrayValue": { "value": "${MY_ARRAY}" }`

Result: `invalid character 'v' after object key:value pair`

**Root cause:** The code was marshaling the entire parameter JSON before running envsubst, which caused double-escaping of JSON arrays/objects:
- Before: `{"value": "${MY_ARRAY}"}`
- After envsubst: `{"value": "["val1","val2"]"}` ← Invalid JSON (quotes inside string)

## Solution

Add a new handling for complex types with string values. In these cases, we run envsubst on **only the string value** and parse the substituted result as JSON to get the actual array/object. This also adds validation for the string value to be a valid JSON before continuing with further execution.

Updated end-to-end test `test/functional/cli_test.go::Test_CLI_InfraCreateAndDelete` to validate provisioning from array and object parameters from env vars.

## Example Usage

```bash
# Set array parameter
azd env set MY_ARRAY '["app1","app2","app3"]'

# Set object parameter  
azd env set MY_CONFIG '{"debug":true,"timeout":30,"tags":["prod","critical"]}'

# Provision with complex types
azd provision
```

Parameters are correctly parsed and passed to the ARM template deployment.

fixes: https://github.com/Azure/azure-dev/issues/5850


JeffreyCA:
We also have the newly added `azd env config` commands (#6348) - do we want similar support there as well?

weikanglim:
@JeffreyCA They should be supported already by the nature of how the prompt stores objects/array as native JSON elements. Although admittedly I haven't done any testing around it.

vhvb1989:
The issue with the env-config, @JeffreyCA, is that you can't map parameters to env-config if the parameters has a default value.
The current design is that env-config holds the value for parameters w/o a default-value after prompting. Or you can set the value in advance in the env-config, but if a default value is defined for the parameter in bicep, it becomes more relevant than the value in the env-config.

JeffreyCA:
> @JeffreyCA They should be supported already by the nature of how the prompt stores objects/array as native JSON elements. Although admittedly I haven't done any testing around it.



> The issue with the env-config, @JeffreyCA, is that you can't map parameters to env-config if the parameters has a default value. The current design is that env-config holds the value for parameters w/o a default-value after prompting. Or you can set the value in advance in the env-config, but if a default value is defined for the parameter in bicep, it becomes more relevant than the value in the env-config.

I meant being able to set types other than string. Currently, if I have in my bicep:

```bicep
@description('Test bool')
param testBool bool

@description('Test array')
param testStrings string[]
```

if I run `azd env config set infra.parameters.testBool`, it writes the following to config.json:
```json
{
  "infra": {
    "parameters": {
      "testBool": "true"
      }
  }
}
```

Similarly for `azd env config set infra.parameters.testStrings '["one", "two", "three"]'`, it writes:
```json
{
  "infra": {
    "parameters": {
      "testStrings": "[\"one\", \"two\", \"three\"]"
      }
  }
}
```

During `azd provision` I'm prompted to re-enter those because they're not the right type:

image

If I manually set it to:

```json
{
  "infra": {
    "parameters": {
      "testBool": true,
      "testStrings": ["one", "two", "three"]
      }
  }
}
```

Then `azd provision` skips the prompts.

---
Created https://github.com/Azure/azure-dev/issues/6701 to track this

PR #6700, opened 2026-02-03
Ability to easily provide custom error messages/suggestions/docs for errors.

wbreza:
## Summary

Adds a YAML-driven error handling pipeline that matches raw Azure errors against well-known patterns and wraps them with user-friendly messages, actionable suggestions, and reference links. The goal: anyone can improve the error experience by editing a single YAML file.

For errors that need runtime context (like querying Azure for available regions), the pipeline supports named handlers registered in the IoC container that compute suggestions dynamically while still pulling static data (links, etc.) from the YAML rule that matched.

This PR also removes `soft_delete_hint.go` (from #6810) since those scenarios are now expressed declaratively as YAML rules.

## How It Works

Error middleware flow:
1. Error occurs
2. Already has suggestion? Pass through
3. Run through ErrorHandlerPipeline - walk rules top-to-bottom (first match wins)
4. Match by: errorType + properties (reflection), text patterns, or both
5. For typed errors: DFS traversal via Unwrap() []error into deep ARM error trees
6. Produce: ErrorWithSuggestion (message + suggestion + links)
7. No match? Fall through to existing AI agent flow

## YAML Rule Format

Rules live in `resources/error_suggestions.yaml` (embedded at build time). Each rule can use text patterns, typed error matching, or both.

**Matching logic:**
- `patterns` - OR (any pattern matches); case-insensitive substring by default, regex when `regex: true`
- `properties` - AND (all must match); resolved via reflection on the matched error type
- Across groups - AND (errorType AND properties AND patterns)
- First match wins - most specific rules go first

**Response fields:**
- `message` - user-friendly explanation
- `suggestion` - actionable next steps
- `links` - list of `{url, title?}` rendered as hyperlinked bullets
- `handler` - name of a registered ErrorHandler for dynamic suggestions (receives the matching rule)

## Custom Handlers

When a rule specifies `handler`, the pipeline resolves the named ErrorHandler from the IoC container and invokes it with the error and the matching rule. The handler can use rule data (e.g., links) or ignore it.

**Built-in: ResourceNotAvailableHandler** handles LocationNotAvailableForResourceType:
1. Extracts the ARM resource type from the error message (prefers the quoted form to avoid matching URLs)
2. Reads AZURE_LOCATION and AZURE_SUBSCRIPTION_ID from the azd environment
3. Queries the ARM Providers API (ResourceTypeLocationService) for available regions
4. Builds a suggestion listing the available regions and how to switch
5. Merges links from the YAML rule into the response

## ARM Error Refactoring

DeploymentErrorLine now implements the Go error interface and Unwrap() []error, enabling:
- errors.As() to find specific error types deep in ARM error trees
- The pipeline DFS traversal to match error codes buried 3-4 levels deep
- Type+property matching that checks both type name AND properties during traversal

## Covered Scenarios

| Category | Errors |
|---|---|
| **Soft-delete conflicts** | FlagMustBeSetForRestore, ConflictError, Conflict/RequestConflict + soft-delete keywords |
| **Quota** | InsufficientQuota, SubscriptionIsOverQuotaForSku, general quota text patterns |
| **Region availability** | LocationNotAvailableForResourceType (both ResponseError and DeploymentErrorLine), via dynamic handler |
| **Auth** | AuthorizationFailed, AADSTS tokens, InvalidAuthenticationToken, ExpiredAuthenticationToken |
| **Templates** | InvalidTemplate, ValidationError, BCP Bicep errors |
| **Resources** | ResourceNotFound, generic Conflict |
| **PowerShell hooks** | Import-Module, missing Az module, UnauthorizedAccess, ErrorActionPreference, Connect-AzAccount, expired login |
| **Project config** | azure.yaml parsing errors |

## Files Changed

| Area | Files |
|---|---|
| **New: Error pipeline** | pkg/errorhandler/ - types, matcher (thread-safe), pipeline, reflect (DFS traversal), handler interface, errors |
| **New: Resource availability** | pkg/errorhandler/resource_availability_handler.go, pkg/azapi/resource_type_locations.go |
| **New: Error rules** | resources/error_suggestions.yaml (25 rules), resources/resources.go (embed) |
| **New: UX component** | pkg/output/ux/error_with_suggestion.go - renders message + suggestion + hyperlinked bullet links + grey raw error |
| **New: Documentation** | docs/error-suggestions.md |
| **Modified: ARM errors** | pkg/azapi/azure_deployment_error.go - Unwrap() []error on both AzureDeploymentError and DeploymentErrorLine |
| **Modified: Middleware** | cmd/middleware/error.go - pipeline integration, cmd/middleware/ux.go - display mapping |
| **Modified: IoC** | cmd/container.go - registers pipeline, handler, location service |
| **Modified: Error type** | internal/errors.go - type alias to errorhandler.ErrorWithSuggestion |
| **Removed** | cmd/middleware/soft_delete_hint.go + tests (superseded by YAML rules) |

## Related PRs

Subsumes error handling from:
- #6810 (soft-delete hints)
- #6801 (ARM root-cause error hints)
- #6804 (PowerShell hook error detection)

### DIFF
--- a/cli/azd/docs/error-suggestions.md
+++ b/cli/azd/docs/error-suggestions.md
@@ -113,7 +113,7 @@ For more complex matching, prefix the pattern with `regex:`. This enables full r
 | `.*` | Match any characters |
 | `\\d+` | Match one or more digits |
 | `\\d{3}` | Match exactly 3 digits |
-| `(foo\|bar)` | Match "foo" or "bar" |
+| `(foo|bar)` | Match "foo" or "bar" |
 | `\\s+` | Match whitespace |
 
 **Note:** In YAML, backslashes must be escaped as `\\`.


cli/azd/resources/error_suggestions.yaml
```diff
@@ -0,0 +1,279 @@
+# Error Suggestions Configuration
```
wbreza, line 1: These are just some generated examples that we can take or remove.

cli/azd/pkg/errorhandler/matcher_test.go
```diff
@@ -0,0 +1,220 @@
...
+				assert.Nil(t, result)
+			}
+		})
+	}
+}
+
+func TestErrorSuggestionService_FirstMatchWins(t *testing.T) {
+	service := NewErrorSuggestionService()
+
+	// An error that could match multiple patterns should return the first match
```
vhvb1989, line 214: How can we prevent that in the futrure, we don't create an entry for an error that overrule others and unintentionally changes all

cli/azd/internal/errors.go
JeffreyCA, line 1: Could we consider making `ErrorWithSuggestion` usable from extensions as well? Currently it's in `internal` which makes it hard to reuse outside of core. I saw the `azd x` extension uses its own [UserFriendlyError](https://github.com/Azure/azure-dev/blob/main/cli/azd/extensions/microsoft.azd.extensions/internal/user_friendly_error.go) but would be great to have something shared

We're facing similar challenges in the AI Agents extension (see https://github.com/Azure/azure-dev/pull/6683)

cli/azd/cmd/middleware/error_test.go
```diff
@@ -77,12 +82,13 @@ func Test_ErrorMiddleware_LLMAlphaFeatureDisabled(t *testing.T) {
 
 	result, err := middleware.Run(*mockContext.Context, nextFn)
 
+	// Should return error without AI intervention in no-prompt mode
```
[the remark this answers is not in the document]
kristenwomack: I don't think this change copilot is suggesting is correct. I think it's conflating no-prompt mode for agent interaction with the alpha agentic mode. @wbreza pls confirm

cli/azd/docs/error-suggestions.md
```diff
@@ -0,0 +1,220 @@
...
+
+**Common regex patterns:**
+
+| Pattern | Meaning |
+|---------|---------|
+| `(?i)` | Case-insensitive matching |
+| `.*` | Match any characters |
+| `\\d+` | Match one or more digits |
+| `\\d{3}` | Match exactly 3 digits |
+| `(foo\|bar)` | Match "foo" or "bar" |
```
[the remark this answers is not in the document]
kristenwomack: looks like a typo

cli/azd/internal/errors.go
kristenwomack: yes!

cli/azd/pkg/errorhandler/matcher_test.go
kristenwomack: Do you mean a new error pattern?

kristenwomack:
I'm very excited about this change

cli/azd/internal/errors.go
wbreza: So i think the ask is that we want extensions themselves to be able to participate in the error handling flow and maybe have their own associated file for looking for error patterns?

cli/azd/pkg/errorhandler/matcher_test.go
wbreza: I think we just need to be conscious and not add error patterns that are too generic.

vhvb1989: Yes, a new pattern.
The current implementation resolves multiple patterns with a match by taking the first match.
This means that the order will decide the importance of each pattern.  This could lead to a future scenario where it is hard to pick the right place for a new pattern if it could re-wire the existing pattern after it.

I guess we can cross that bridge when we get there 😜

vhvb1989 (approved):

cli/azd/internal/errors.go
kristenwomack: Yes, but depending on the complexity maybe we ship this and then evolve/add that.

JeffreyCA: The file-based error suggestion flow would be nice to have but I was thinking more straightforward/immediate: extensions being able to use the `ErrorWithSuggestion` type and the UX rendering

cli/azd/docs/error-suggestions.md
```diff
@@ -0,0 +1,220 @@
...
+| `(foo|bar)` | Match "foo" or "bar" |
+| `\\s+` | Match whitespace |
+
+**Note:** In YAML, backslashes must be escaped as `\\`.
+
+### Rule Evaluation
+
+- **First match wins**: Rules are evaluated in order from top to bottom
+- **Order matters**: Place more specific patterns before general ones
+- **Multiple patterns per rule**: If any pattern in a rule matches, that rule wins
```
tg-msft, line 125: Regular expressions are not rich enough to solve this problem. 
- We're completely skipping HTTP status codes and `x-ms-error-code`s which are the two most important details for any Azure error.
- We're completely ignoring the structure.  Being able to discern between the top-level error response and nested details matters a lot in some situations.
- Simple regexes aren't enough to distinguish context in many situations (i.e., you'd handle "quota exceeded" subscription limits very differently than you would AI model quota)
- This doesn't compose/scale well.  "First rule wins" text matching isn't going to let us easily customize the error experience for many services each loaded in their own extensions.
- I can't run arbitrary code to compute my suggestion (i.e., I can't determine regions where your active subscription still has quota).

I think we need to add a layer underneath what you have here.  We should create an interface for matching errors and computing suggested responses that mirrors an HTTP middleware pipeline.  We could implement this regex matcher with it, the agentic matcher with it, and eventually allow other extensions to register their own error handlers.  We could also explore allowing all matches to flow through and either let users flip between them or asking an agent which one makes the most sense.

hemarina (approved):
LGTM

cli/azd/pkg/errorhandler/matcher.go
```diff
@@ -0,0 +1,68 @@
...
+	"regexp"
+	"strings"
+)
+
+const regexPrefix = "regex:"
+
+// PatternMatcher handles matching error messages against patterns.
+type PatternMatcher struct {
+	// compiledPatterns caches compiled regex patterns for performance
+	compiledPatterns map[string]*regexp.Regexp
```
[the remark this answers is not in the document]
wbreza: Good catch — added `sync.RWMutex` to protect the compiled regex cache. Read lock for cache hits, write lock only when compiling a new pattern.

cli/azd/cmd/middleware/error_test.go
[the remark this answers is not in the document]
wbreza: Fixed — updated the comment to: *Should return error without AI intervention when LLM alpha feature is not enabled*

cli/azd/docs/error-suggestions.md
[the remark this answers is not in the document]
wbreza: Fixed — the escaped pipe was corrected in a prior commit.

cli/azd/resources/error_suggestions.yaml
```diff
@@ -0,0 +1,279 @@
...
+#       - "some error text"
+#       - "regex:pattern\\d+"
+#     message: "A brief, user-friendly explanation of the error."
+#     suggestion: "Clear instruction on how to fix the issue."
+#     docUrl: "https://learn.microsoft.com/..."
+
+rules:
+  - patterns:
+      - "parsing project file"
+    message: "Your azure.yaml file in invalid"
```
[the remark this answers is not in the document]
wbreza: Fixed in a prior commit.

cli/azd/internal/errors.go
wbreza: Agreed — making `ErrorWithSuggestion` and the UX rendering available to extensions is a great next step. For this PR we'll ship the core pipeline and iterate from there.

cli/azd/docs/error-suggestions.md
wbreza: Great feedback. This PR now includes typed error matching via reflection (`errorType` + `properties`), deep ARM error tree traversal via multi-unwrap, and named handlers (`ResourceNotAvailableHandler` queries ARM for available regions). The pipeline architecture with the `ErrorHandler` interface is designed to be extended — adding HTTP status code matching, extension-registered handlers, and richer composition are natural next steps on top of this foundation.

cli/azd/pkg/errorhandler/matcher_test.go
wbreza: The YAML file now has a header comment explaining the specificity-first ordering convention, and the docs cover it as well. For now the ordering is manageable — we can add validation tooling if the rule set grows significantly.

cli/azd/docs/error-suggestions.md
wbreza: Updated with a more detailed response now that the implementation has evolved significantly:

> We're completely skipping HTTP status codes and x-ms-error-code

We now match `azcore.ResponseError` by type via reflection, with `ErrorCode` property matching. For example, `LocationNotAvailableForResourceType` is caught both as a `ResponseError` (pre-polling validation failure) and as a `DeploymentErrorLine` (nested in deployment errors). Adding `StatusCode` matching would be straightforward since `resolvePropertyPath` already handles any exported field.

> We're completely ignoring the structure

The pipeline now does DFS traversal through `Unwrap() []error` on ARM error trees, matching `DeploymentErrorLine` nodes by their `Code` property. This finds error codes buried 3-4 levels deep (e.g., `DeploymentFailed > ResourceDeploymentFailure > FlagMustBeSetForRestore`). Type + property matching happens together during traversal so we match the right node, not just the first type match.

> Simple regexes aren't enough to distinguish context

Agreed — that's why we have typed error matching + properties as the primary strategy for structured errors. Text patterns are a fallback for unstructured errors. Rules can combine both: `errorType: "DeploymentErrorLine"` + `properties: {Code: "Conflict"}` + `patterns: ["(?i)soft.?delete"]` to distinguish soft-delete conflicts from generic ones.

> This doesn't compose/scale well

The `ErrorHandler` interface is the extension point here. Named handlers are resolved from the IoC container and receive the full matching rule. This is designed so extensions can register their own handlers. Evolving to let extensions also register their own rules is a natural next step.

> I can't run arbitrary code to compute my suggestion

Named handlers solve this. The built-in `ResourceNotAvailableHandler` queries the ARM Providers API for available regions, reads from the azd environment, and builds a dynamic suggestion. Any handler can run arbitrary code — it receives the context, error, and matching rule.

cli/azd/internal/errors.go
wbreza: Good news — `ErrorWithSuggestion` now lives in `pkg/errorhandler` (the `internal` package just has a type alias). Extensions can import and use it directly. The UX rendering component is in `pkg/output/ux/error_with_suggestion.go` which is also importable.

PR #6730, opened 2026-02-05
[Agents Extension] Improving handling around model selection

trangevi:


cli/azd/extensions/azure.ai.agents/internal/cmd/init.go
```diff
@@ -2207,6 +2240,219 @@ func (a *InitAction) getModelDetails(ctx context.Context, modelName string) (*ai
...
+		}
+
+		_, err = a.azdClient.Workflow().Run(ctx, &azdext.RunWorkflowRequest{
+			Workflow: workflow,
+		})
+		if err != nil {
+			return nil, "", fmt.Errorf("failed to update AZURE_LOCATION in azd environment: %w", err)
+		}
+
+		fmt.Printf("Updated AZURE_LOCATION to '%s' in your azd environment.\n", selectedLocation)
```
JeffreyCA: Any particular reason we're not using `azdClient.Environment()`?

trangevi: Because I forgot about it...

JeffreyCA: GetValue/SetValue might not work if you omit the env name, I'll see if we can support that easily

trangevi: The change I made just uses the current env, like I'm doing elsewhere

trangevi:
Need to validate sku bug before merge

rajeshkamal5050 (approved):

PR #6664, opened 2026-01-30
Add JMESPath query support for JSON output in CLI commands

scottaddie:
Addresses https://github.com/Azure/azure-dev/issues/3206, https://github.com/Azure/azure-dev/issues/4067#issuecomment-2215287564

Azure CLI allows you to run JMESPath queries on JSON output. This PR brings that same capability to azd commands that support JSON output. A dependency on [go-jmespath](https://jmespath.site/main/#libraries) is needed to accomplish this.

**Usage example from a local run:**
```
./azd.exe config show -o json --query "template.sources"
{
  "awesome-azd": {
    "key": "awesome-azd",
    "location": "https://aka.ms/awesome-azd/templates.json",
    "name": "Awesome AZD",
    "type": "awesome-azd"
  }
}
```

PR #6735, opened 2026-02-06
Extend JMESPath support to Message methods

scottaddie:
Enhance the JMESPath support added in https://github.com/Azure/azure-dev/pull/6664 by extending it to the Message methods. 

## Current behavior
When logged out, `azd auth token --output json --query data.message` returns:

```json
{"type":"consoleMessage","timestamp":"2026-02-06T15:48:49.8045685-06:00","data":{"message":"\nERROR: fetching token: AzureCLICredential: ERROR: Please run 'az login' to setup account.\r\n\n"}}
```

## New behavior
When logged out, `azd auth token --output json --query data.message` returns:

```
"\nERROR: fetching token: AzureCLICredential: ERROR: Please run 'az login' to setup account.\r\n\n"
```

This new support allows us to avoid writing code to manually parse the JSON. For example, in [this PR](https://github.com/Azure/azure-sdk-for-net/pull/55553).

### DIFF
--- a/cli/azd/pkg/input/console_test.go
+++ b/cli/azd/pkg/input/console_test.go
@@ -14,6 +14,7 @@ import (
 	"time"
 
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/to"
+	"github.com/azure/azure-dev/cli/azd/pkg/contracts"
 	"github.com/azure/azure-dev/cli/azd/pkg/output"
 	"github.com/stretchr/testify/require"
 )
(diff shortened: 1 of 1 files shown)


vhvb1989:
This would be a breaking change for AzureDeveloperCredential in all languages and old versions.

My recommendation would be not to do this @rajeshkamal5050 @kristenwomack 

FYI @chlowell  - You were asking the other day how stable was the response.

scottaddie:
I agree we don't want to introduce breaking changes. Is there a way something like this could be implemented in a non-breaking fashion? It's painful to see consumers writing code to manually parse the response, when it can easily be solved with JMESPath.

vhvb1989:
@scottaddie , What about supporting `azd auth token` (w/o `-o json`) -   The `json` flag should be returning valid JSON.
The `azd auth token` currently only supports this flag as mandatory, but maybe we can add support to either output the token only as string or return the error message string. What do you think?

JeffreyCA:
> This would be a breaking change for AzureDeveloperCredential in all languages and old versions.
> 
> My recommendation would be not to do this @rajeshkamal5050 @kristenwomack
> 
> FYI @chlowell - You were asking the other day how stable was the response.

My read of this PR was that `azd auth token --output json` **without** `--query` still returns the same JSON as before, and isn't affected so I'm not seeing how this would be a breaking change, especially since `--query` support (https://github.com/Azure/azure-dev/pull/6664) isn't officially released yet (not in 1.23.3). Do you mind clarifying?

vhvb1989:
> > This would be a breaking change for AzureDeveloperCredential in all languages and old versions.
> > My recommendation would be not to do this @rajeshkamal5050 @kristenwomack
> > FYI @chlowell - You were asking the other day how stable was the response.
> 
> My read of this PR was that `azd auth token --output json` **without** `--query` still returns the same JSON as before, and isn't affected so I'm not seeing how this would be a breaking change, especially since `--query` support (#6664) isn't officially released yet (not in 1.23.3). Do you mind clarifying?

I got confused, then. I thought we were changing the output even w/o the --query.
Based on what you said @JeffreyCA , this is a bug-fix then

puicchan:
Based on latest comment, @vhvb1989 & @JeffreyCA, are we good to merge this? I have a 1-min blog post created for https://github.com/Azure/azure-dev/pull/6664. I will update that draft to include this if yes.

PR #6739, opened 2026-02-09
Fix arrow keys in ux prompt library with Ghostty 

### LINKED ISSUE #5704: [Issue] Arrow keys not working in azd extension prompts on MacOS with Ghostty terminal

**Output from `azd version`**
```
azd version 1.18.2 (commit a7b2bdc861f0616438327854f11169a08fdfec96)
```

**Describe the bug**
When running azd extension prompts (e.g. azd dev init), the up/down arrow keys do not work in the selection menus. This prevents selecting Python (or other options) in the interactive prompt.

The issue seems terminal-specific:
- Not working in: [Ghostty](https://ghostty.org/)
- Working in: iTerm2

**To Reproduce**
- Open Ghostty terminal on macOS 15.6.1
- Run azd dev init with an extension that asks for language selection
- Try to navigate with the up/down arrows

**Expected behavior**
Arrow keys should allow selecting options (e.g. Python) in the interactive prompt.

**Environment**
- OS: macOS 15.6.1 (24G90)
- Terminal: Ghostty 1.1.2 Build 9416 Commit [ee963f62](https://github.com/ghostty-org/ghostty/commits/ee963f62)

**Additional context**
Issue does not reproduce in iTerm2.

weikanglim:
## Problem

Arrow keys displayed as escape sequences (`^[[A^[[B`) in Ghostty terminal when using `pkg/ux` components (Select, MultiSelect, Prompt, Confirm).

**Root Cause:** The `eiannone/keyboard` library loads terminfo before deciding how to interpret keyboard sequences using a hardcoded list of what type of terminals. Ghostty is an unrecognized terminal, and `eiannone/keyboard` is unmaintained. 

## Solution
Replaced `github.com/eiannone/keyboard` with `github.com/AlecAivazis/survey/v2/terminal`, which always assumes VT100 ANSII escape sequences and avoids terminfo sniffing. This replacement is extremely safe since we're already using `terminal` underneath  in our default prompts which has been battle tested.

Tested on macOS, Linux, and Windows.

Fixes #5704

### DIFF
--- a/cli/azd/pkg/ux/internal/input.go
+++ b/cli/azd/pkg/ux/internal/input.go
@@ -83,8 +83,6 @@ func (i *Input) ReadInput(ctx context.Context, config *InputConfig, handler KeyP
 		signal.Stop(signalChan)
 	}()
 
-	// Create a RuneReader from survey's terminal package.
-	// This uses hardcoded ANSI parsing instead of terminfo, which is more reliable across different terminals.
 	stdio := surveyterm.Stdio{In: os.Stdin, Out: os.Stdout, Err: os.Stderr}
 	rr := surveyterm.NewRuneReader(stdio)
 
@@ -93,6 +91,8 @@ func (i *Input) ReadInput(ctx context.Context, config *InputConfig, handler KeyP
 	// To ensure we can still handle Ctrl+C or context cancellations.
 	go func() {
 		// Set terminal to raw mode for reading
+		// This allows us to read all user inputs without the terminal processing them
+		// (e.g., handling backspace, Ctrl+C, etc.).
 		if err := rr.SetTermMode(); err != nil {
 			errChan <- err
 			return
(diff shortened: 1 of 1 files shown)


cli/azd/pkg/ux/internal/input.go
```diff
@@ -10,10 +10,9 @@ import (
 	"os"
 	"os/signal"
 	"syscall"
-	"time"
 	"unicode"
 
-	"github.com/eiannone/keyboard"
+	surveyterm "github.com/AlecAivazis/survey/v2/terminal"
```
vhvb1989, line 15: I wonder if we can avoid increasing the use of `AlecAivazis/survery` since it is now archived.
And `eiannone/keyboard` is not archived but it's last commit is 4 years old - stills runs on go 1.18

weikanglim: I think the usage is perfectly fine here. While unmaintained, the code we're importing here is very stable and unlikely to change. I did not find a different package that would offer similar functionality.

It is very achievable for us to vendor the relevant parts we needed if we were to decide to break away from `survey` completely. I wouldn't be too worried here.

JeffreyCA (approved):

Issue #6531, opened 2026-01-15
[Feature request] Support new properties in azure.yaml for Azure Function app deployment


lilyjma:
## Context
Azd does remote build by default when deploying a Node Function app. Azd also runs npm install and build automatically. These two actions lead to failed deployments if the user tries to deploy the app on a Windows machine. The reason is the following. The generated `node_modules/` directory from the install and build step gets included as part of the deployment payload. During deployment, remote build (running on Oryx) tries to run the typescript package inside the directory but that package doesn't work for Linux, so the user gets an Oryx error and deployment fails. 

### Workaround and problems
Node users need include `node_modules/` in `.funcignore` so that it's excluded during project packaging; remote build will generate a new `node_modules/`.  This workaround isn't ideal because users would have to know the internals of Function app deployment to know this. 

This workaround also doesn't address a bigger problem, which is other clients like VSCode **don't** turn on remote build by default, so users wouldn't have to include `node_modules/` in `.funcignore`. This means depending on how they're deploying the function app, they need different `.funcignore`s, making it difficult to switch between clients. 

## Ask

- Keep the default today, i.e. remote build is on, for backwards compat purposes but support specifying `remoteBuild` in azure.yaml for turning it off:
  ```yaml
  services:
    api:
      project: .
      language: js
      host: function
      remoteBuild: false | true
  ```

How this helps customers: Functions Node customers deploying with azd would specify `remoteBuild:false` in azure.yaml if they've already built the project locally (perhaps for local testing). If they later switch to another client like VSCode, they can keep the same `.funcignore` for successful deployment.  

- Support skipping npm install and run when remote build is on with `packageOnly`: 
  ```yaml
  services:
    api:
      project: .
      language: js
      host: function
      packageOnly: true | false
  ```

How this helps customers:  For customers that just clone the Function sample and immediately deploy, the`packageOnly: true` lets remote build in Oryx run npm install/build, ensuring that the right `node_modules/` is generated for deployment.

anthonychu:
Currently, the package step always runs npm install and build. Can it also look at this setting so it skips those steps when remote build is on?

vhvb1989:
remoteBuild does not currently apply for functionapp because we don't support deploying containers to functionapp (yet)
We shouldn't use `docker.remoteBuild` for a non-containers flow.

That being said, I think I understand the requirement as if, for your use case, you don't need to build the service locally because you need to deploy the full source-code to functionapp anyway?

[...]

PR #6748, opened 2026-02-10
add remoteBuild for flex consumption apps

weikanglim:
Adds support for `remoteBuild` option in `azure.yaml` to allow users to explicitly control remote build behavior for Azure Function services on `flexconsumption` plan.

Addresses #6531

### DIFF
--- a/cli/azd/pkg/project/service_config.go
+++ b/cli/azd/pkg/project/service_config.go
@@ -59,6 +59,9 @@ type ServiceConfig struct {
 	// Condition for deploying the service. When evaluated, the service is only deployed if the value
 	// is a truthy boolean (1, true, TRUE, True, yes). If not defined, the service is enabled by default.
 	Condition osutil.ExpandableString `yaml:"condition,omitempty"`
+	// Whether to build the service remotely. Only applicable to function app services.
+	// When set to nil (unset), the default behavior based on language is used.
+	RemoteBuild *bool `yaml:"remoteBuild,omitempty"`
 
 	// AdditionalProperties captures any unknown YAML fields for extension support
 	AdditionalProperties map[string]interface{} `yaml:",inline"`
--- a/cli/azd/pkg/project/service_target_functionapp.go
+++ b/cli/azd/pkg/project/service_target_functionapp.go
@@ -10,28 +10,33 @@ import (
 	"path/filepath"
 	"strings"
 
+	"github.com/azure/azure-dev/cli/azd/internal"
 	"github.com/azure/azure-dev/cli/azd/internal/mapper"
 	"github.com/azure/azure-dev/cli/azd/pkg/async"
 	"github.com/azure/azure-dev/cli/azd/pkg/azapi"
 	"github.com/azure/azure-dev/cli/azd/pkg/environment"
+	"github.com/azure/azure-dev/cli/azd/pkg/input"
 	"github.com/azure/azure-dev/cli/azd/pkg/tools"
 )
 
 // functionAppTarget specifies an Azure Function to deploy to.
 // Implements `project.ServiceTarget`
 type functionAppTarget struct {
-	env *environment.Environment
-	cli *azapi.AzureClient
+	env     *environment.Environment
+	cli     *azapi.AzureClient
+	console input.Console
 }
 
 // NewFunctionAppTarget creates a new instance of the Function App target
 func NewFunctionAppTarget(
 	env *environment.Environment,
 	azCli *azapi.AzureClient,
+	console input.Console,
 ) ServiceTarget {
 	return &functionAppTarget{
-		env: env,
-		cli: azCli,
+		env:     env,
+		cli:     azCli,
+		console: console,
 	}
 }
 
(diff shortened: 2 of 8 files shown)


cli/azd/pkg/project/service_target_functionapp.go
```diff
@@ -125,9 +125,14 @@ func (f *functionAppTarget) Deploy(
...
-		serviceConfig.Language == ServiceLanguageTypeScript ||
-		serviceConfig.Language == ServiceLanguagePython
+	var remoteBuild bool
+	if serviceConfig.RemoteBuild != nil {
+		remoteBuild = *serviceConfig.RemoteBuild
+	} else {
+		remoteBuild = serviceConfig.Language == ServiceLanguageJavaScript ||
+			serviceConfig.Language == ServiceLanguageTypeScript ||
+			serviceConfig.Language == ServiceLanguagePython
+	}
```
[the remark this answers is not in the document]
kristenwomack: This is a good catch. We should make this explicit for non-flex plans. @weikanglim

kristenwomack:
Can we set up a test/telemetry to measure if this change is helping customers navigate remote build in this scenario (by language)?

schemas/alpha/azure.yaml.json
```diff
@@ -228,6 +228,11 @@
                         "type": "string",
                         "title": "Relative path to service deployment artifacts"
                     },
+                    "remoteBuild": {
+                        "type": "boolean",
+                        "title": "Optional. Whether to build the service remotely",
+                        "description": "Optional. When set to true, the service will be built remotely using Azure's remote build service (Oryx). When set to false, the service will be deployed as-is from the local build output. Only applicable to 'function' host type on 'flexconsumption' SKU. If not specified, the default behavior based on the service language is used (true for JavaScript, TypeScript, and Python; false for others)."
+                    },
```
vhvb1989: Can we make azd to know when to use `remoteBuild` based on the appservice settings?
Like, during deploy, we get the settings and make azd to honor the configuration for it.

That way we prevent errors when the appservice's configuration defines build-on-deploy (remote build) but azd did not set the flag.

weikanglim: There's no appservice setting for flex consumption apps. See [docs](https://learn.microsoft.com/en-us/azure/azure-functions/functions-app-settings#flex-consumption-plan-deprecations)

vhvb1989: oh, but there is a default in the server when it is not provided during deployment, right?  Is it true by default?  
So azd during zip+deploy w/o setting the remoteBuild property goes to remoteBuild=true by default?

schemas/alpha/azure.yaml.json
```diff
@@ -228,6 +228,11 @@
                         "type": "string",
                         "title": "Relative path to service deployment artifacts"
                     },
+                    "remoteBuild": {
+                        "type": "boolean",
+                        "title": "Optional. Whether to build the service remotely",
+                        "description": "Optional. When set to true, the service will be built remotely using Azure's remote build service (Oryx). When set to false, the service will be deployed as-is from the local build output. Only applicable to 'function' host type on 'flexconsumption' SKU. If not specified, the default behavior based on the service language is used (true for JavaScript, TypeScript, and Python; false for others)."
```
vhvb1989: Should we speak in terms of runtime?  Like calling `node` instead of `Js/Ts` ?

azd can read the runtime from the appservice and it could be:

```
@allowed([
  'dotnet', 'dotnetcore', 'dotnet-isolated', 'node', 'python', 'java', 'powershell', 'custom'
])
```

cli/azd/pkg/project/service_target_functionapp.go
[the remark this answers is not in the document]
weikanglim: Updated

JeffreyCA (approved):

cli/azd/pkg/project/service_target_functionapp.go
```diff
@@ -124,19 +129,61 @@ func (f *functionAppTarget) Deploy(
...
+
+	progress.SetProgress(NewServiceProgress("Uploading deployment package"))
+	var remoteBuild bool
+	if serviceConfig.RemoteBuild != nil {
+		remoteBuild = *serviceConfig.RemoteBuild
+	} else {
+		remoteBuild = serviceConfig.Language == ServiceLanguageJavaScript ||
+			serviceConfig.Language == ServiceLanguageTypeScript ||
+			serviceConfig.Language == ServiceLanguagePython
```
rajeshkamal5050, line 165: In a subsequent change and a future release, we will change and default ts/js to local builds?

@spboyer @weikanglim

cli/azd/pkg/project/service_config.go
```diff
@@ -59,6 +59,9 @@ type ServiceConfig struct {
 	// Condition for deploying the service. When evaluated, the service is only deployed if the value
 	// is a truthy boolean (1, true, TRUE, True, yes). If not defined, the service is enabled by default.
 	Condition osutil.ExpandableString `yaml:"condition,omitempty"`
+	// Whether to build the service remotely. Only applicable to function app services.
+	// When set to nil (unset), the default behavior based on language is used.
+	RemoteBuild *bool `yaml:"remoteBuild,omitempty"`
```
wbreza, line 64: I'm concerned about having the same/similar property names at multiple levels... a service already as a "remote build" option in the docker options now we have this one? Can we think about how we may consolidate this to a single property?

rajeshkamal5050: Added an issue Wallace - https://github.com/Azure/azure-dev/issues/6756 since refactoring touches the existing docker config and adjusting to use the common one cc @weikanglim

PR #6777, opened 2026-02-18
Add --subscription and --location flags to provision and up commands

### LINKED ISSUE #6774: Add --subscription and --location inline flags to azd up and azd provision
## Problem

Every Azure CLI tool accepts `--subscription` as an inline flag. When users (human or AI agent) try `azd up --subscription <id>`, they get `Error: unknown flag: --subscription` and must discover the `azd env set AZURE_SUBSCRIPTION_ID` pattern. Same for `--location`.

This breaks the principle of least surprise for anyone coming from `az` CLI and is the #1 friction point reported by AI agents attempting to use azd non-interactively.

## Proposed Solution

Add `--subscription` and `--location` flags to `azd provision` (and by extension `azd up` which embeds `ProvisionFlags`). When provided, these flags set the corresponding environment variables (`AZURE_SUBSCRIPTION_ID`, `AZURE_LOCATION`) before `EnsureSubscriptionAndLocation` runs.

### Existing precedent
`azd env new` already accepts `--subscription` and `--location` flags (`cli/azd/cmd/env.go:888-904`). This change follows the same pattern.

### Implementation sketch
- Add `subscription` and `location` fields to `ProvisionFlags` in `cli/azd/internal/cmd/provision.go`
- In the provision action's `Run()`, apply flag values to the environment before provisioning begins
- `azd up` inherits these flags automatically via embedded `ProvisionFlags`
- Values are persisted to the environment (same as `azd env set`)

### Backward compatibility
Fully backward-compatible. Existing `azd env set` workflow continues to work. Inline flags are an additive convenience.

spboyer:
## Summary

Adds `--subscription` and `--location` inline flags to `azd provision` and `azd up` commands. This allows users to specify the Azure subscription and location directly on the command line, skipping the interactive prompts.

Fixes #6774

## Changes

- **`cli/azd/internal/cmd/provision.go`**: Added `subscription` and `location` fields to `ProvisionFlags`, registered them in `BindNonCommon()`, and added logic in `ProvisionAction.Run()` to apply flag values to the environment before provisioning begins.
- **`cli/azd/cmd/up.go`**: Added logic in `upAction.Run()` to apply subscription/location flag values to the environment before the provisioning manager initializes (and before `EnsureSubscriptionAndLocation` is called).
- **Snapshot files**: Updated `TestUsage-azd-provision.snap` and `TestUsage-azd-up.snap` to reflect the new flags in help output.

## Usage

```bash
# Provision with inline flags
azd provision --subscription "my-sub-id" --location eastus2

# Up with inline flags
azd up --subscription "my-sub-id" -l westus3

# Flags persist to the environment (same as azd env set)
```

## Design

Follows the existing pattern from `azd env new` which already supports these flags. Flag values are applied to the environment via `env.SetSubscriptionId()` / `env.SetLocation()` + `envManager.Save()` **before** the provisioning manager initializes, so `EnsureSubscriptionAndLocation` sees the values and skips prompting. Fully backward-compatible - existing workflows are unchanged.

### DIFF
--- a/cli/azd/cmd/up.go
+++ b/cli/azd/cmd/up.go
@@ -105,12 +105,24 @@ func newUpAction(
 func (u *upAction) Run(ctx context.Context) (*actions.ActionResult, error) {
 	// Apply --subscription and --location flags to the environment before provisioning
 	updatedEnv := false
-	if u.flags.ProvisionFlags.Subscription() != "" {
-		u.env.SetSubscriptionId(u.flags.ProvisionFlags.Subscription())
+	if flagSub := u.flags.ProvisionFlags.Subscription(); flagSub != "" {
+		if existing := u.env.GetSubscriptionId(); existing != "" && existing != flagSub {
+			return nil, fmt.Errorf(
+				"cannot change subscription for existing environment '%s' (current: %s, requested: %s). "+
+					"Create a new environment with 'azd env new' instead",
+				u.env.Name(), existing, flagSub)
+		}
+		u.env.SetSubscriptionId(flagSub)
 		updatedEnv = true
 	}
-	if u.flags.ProvisionFlags.Location() != "" {
-		u.env.SetLocation(u.flags.ProvisionFlags.Location())
+	if flagLoc := u.flags.ProvisionFlags.Location(); flagLoc != "" {
+		if existing := u.env.GetLocation(); existing != "" && existing != flagLoc {
+			return nil, fmt.Errorf(
+				"cannot change location for existing environment '%s' (current: %s, requested: %s). "+
+					"Create a new environment with 'azd env new' instead",
+				u.env.Name(), existing, flagLoc)
+		}
+		u.env.SetLocation(flagLoc)
 		updatedEnv = true
 	}
 	if updatedEnv {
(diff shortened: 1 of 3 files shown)


cli/azd/cmd/up.go
```diff
@@ -103,6 +103,22 @@ func newUpAction(
...
+	}
+	if u.flags.ProvisionFlags.Location() != "" {
+		u.env.SetLocation(u.flags.ProvisionFlags.Location())
+		updatedEnv = true
+	}
+	if updatedEnv {
+		if err := u.envManager.Save(ctx, u.env); err != nil {
+			return nil, fmt.Errorf("saving environment: %w", err)
+		}
+	}
```
vhvb1989, line 132: I understand the motivation - for someone coming from `az`.
But this has a side impact to the environment mechanics in azd. Having an environment as state context is probably the biggest difference between az and azd. It is design exactly so you don't need to set all arguments for each call you do - like sub, location, rg, etc. 

By allowing sub and loc arg in up and provision, we are allowing to easily change the sub and loc previously used for one env - For an already provisioned application, this would mean to use the same env to re-provision in another sub/loc, leaking the previous provision.

Maybe we should check if the sub and loc are already set in the env and return an error like:
"You are trying to use a different [sub/loc] than what is currently defined for the environment. Create a new environment instead"

spboyer:
Great catch! Added a guard that returns an error if you try to change the subscription or location for an environment that already has them set. The error message directs the user to create a new environment instead. Same value = no-op (allowed), empty value = first time setting (allowed), different value = error with clear guidance.

PR #6779, opened 2026-02-18
Fail fast in azd init when --environment missing non-interactively

### LINKED ISSUE #6775: azd init should fail fast when --environment is missing in non-interactive mode
## Problem

When running `azd init --template <url>` without `--environment` in a non-interactive context (`--no-prompt` or non-TTY), azd downloads the entire template repository **before** failing at the environment name prompt. This leaves a half-initialized directory that must be manually cleaned up before retrying.

Current execution order:
1. `initializeTemplate()` — downloads and copies template files ✅ (expensive, succeeds)
2. `initializeEnv()` — prompts for environment name ❌ (fails in non-interactive mode)

This is the worst outcome for automated workflows: do expensive work, then fail.

## Proposed Solution

Add an early validation check in `cli/azd/cmd/init.go` **before** `initializeTemplate()` is called:

- If running non-interactively (`--no-prompt` flag or non-TTY stdin) **and** `--environment` is not provided, return a clear error immediately:
  `--environment is required when running non-interactively. Use: azd init --template <url> --environment <name>`

### Scope
- Only affects non-interactive mode — interactive users still get the prompt as before
- No changes to template download logic
- Targeted guard, minimal code change

### Impact
Prevents the frustrating download-then-fail-then-cleanup cycle for agents and CI/CD pipelines.

spboyer:
Fixes #6775

When running `azd init --template <url> --no-prompt` without `--environment`, the template would be downloaded first (expensive operation), only to fail at the environment name prompt. This change adds an early validation check that returns a clear error message before any expensive operations.

## Changes
- Added fail-fast validation in `initAction.Run()` before the git check and template download
- Only triggers when `--no-prompt` + `--template` are set but `--environment` is missing
- Interactive mode and non-template flows are unaffected
- Added unit tests for the fail-fast behavior

### DIFF
--- a/cli/azd/test/functional/init_test.go
+++ b/cli/azd/test/functional/init_test.go
@@ -301,12 +301,12 @@ func Test_CLI_Init_WithCwdAutoCreate(t *testing.T) {
 		{
 			name:   "single level directory",
 			subDir: "new-project",
-			args:   []string{"init", "-t", "azure-samples/todo-nodejs-mongo", "--no-prompt"},
+			args:   []string{"init", "-t", "azure-samples/todo-nodejs-mongo", "--no-prompt", "-e", "test-env"},
 		},
 		{
 			name:   "nested directory",
 			subDir: "parent/child/project",
-			args:   []string{"init", "-t", "azure-samples/todo-nodejs-mongo", "--no-prompt"},
+			args:   []string{"init", "-t", "azure-samples/todo-nodejs-mongo", "--no-prompt", "-e", "test-env"},
 		},
 	}
 
@@ -328,8 +328,7 @@ func Test_CLI_Init_WithCwdAutoCreate(t *testing.T) {
 			require.NoDirExists(t, targetDir)
 
 			// Run the command
-			// Note: We expect an error because --no-prompt will fail on environment name prompt
-			// but the directory creation should succeed before that
+			// The -e flag provides an environment name to avoid fail-fast when using --no-prompt with --template
 			cli.RunCommand(ctx, args...)
 
 			// Verify the directory was created


cli/azd/cmd/init.go
```diff
@@ -194,6 +194,14 @@ func (i *initAction) Run(ctx context.Context) (*actions.ActionResult, error) {
 				"using branch argument (-b or --branch) requires a template argument (--template or -t) to be specified")
 	}
 
+	// Fail fast when running non-interactively with --template but without --environment
+	// to avoid downloading the template and then failing at the environment name prompt.
+	if i.flags.global.NoPrompt && i.flags.templatePath != "" && i.flags.EnvironmentName == "" {
+		return nil, errors.New(
+			"--environment is required when running in non-interactive mode (--no-prompt) with --template. " +
+				"Use: azd init --template <url> --environment <name> --no-prompt")
```
vhvb1989: Since this could be some automation or an Agent, what if we create a default env name?  Something like calling the env `unnamed`, just to complete the init step (the name does not make any difference at that point anyway).  

Then we can ask to re-name the env before provision runs... 

This is just an idea to make naming the env less important (not the very first thing to know).

cli/azd/cmd/init.go
```diff
@@ -194,6 +194,14 @@ func (i *initAction) Run(ctx context.Context) (*actions.ActionResult, error) {
 				"using branch argument (-b or --branch) requires a template argument (--template or -t) to be specified")
 	}
 
+	// Fail fast when running non-interactively with --template but without --environment
+	// to avoid downloading the template and then failing at the environment name prompt.
+	if i.flags.global.NoPrompt && i.flags.templatePath != "" && i.flags.EnvironmentName == "" {
```
vhvb1989: What happen if an environment already exists?  Like running `azd init` again after it completed? (like some people do to pull the latest changes for a template) ?   Would the existing env populate the i.flags.EnvironmentName and false this condition?  or would we return this error asking for -e arg ?

cli/azd/cmd/init_test.go
```diff
@@ -0,0 +1,74 @@
...
+		// Verify the condition: without template, no fail-fast error
+		flags := &initFlags{
+			templatePath: "",
+			global:       &internal.GlobalCommandOptions{NoPrompt: true},
+		}
+
+		shouldFailFast := flags.global.NoPrompt && flags.templatePath != "" && flags.EnvironmentName == ""
+		require.False(t, shouldFailFast)
+	})
+}
```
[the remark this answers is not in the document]
spboyer: This test case (DoesNotFailWhenEnvProvidedViaDotEnv) was added in commit b667ad15 as part of the fix that reordered the .env loading to happen before the fail-fast check. The test writes a .env file with AZURE_ENV_NAME=from-dotenv and verifies the fail-fast error does not trigger.

cli/azd/cmd/init.go
spboyer: Great question! When re-running \azd init\ in a directory that already has an environment, this works correctly:

1. The first \azd init\ run creates the environment and writes \AZURE_ENV_NAME\ into the \.env\ file (via \nvManager.Save\).
2. On subsequent runs, the \.env\ file is loaded via \godotenv.Overload()\ at [lines 210-215](https://github.com/Azure/azure-dev/pull/6779/files#diff-init-go-L210) **before** the fail-fast check at [line 228](https://github.com/Azure/azure-dev/pull/6779/files#diff-init-go-L228).
3. After loading \.env\, \AZURE_ENV_NAME\ is picked up at [line 222](https://github.com/Azure/azure-dev/pull/6779/files#diff-init-go-L222) and set into \lags.EnvironmentName\, so the fail-fast condition (\lags.EnvironmentName == \\\) is false.

The fix in commit b667ad15 specifically reordered these operations to ensure \.env\ loading happens first. We also have a test case (\DoesNotFailWhenEnvProvidedViaDotEnv\) that validates this exact scenario — it writes a \.env\ with \AZURE_ENV_NAME\ and confirms the fail-fast check does not trigger.

cli/azd/cmd/init.go
spboyer: Good suggestion! We considered this, but there are a few concerns with auto-generating a default name like `unnamed`:

1. **Collision risk** — If multiple `azd init` invocations (from different agents or automation runs) all create an environment named `unnamed`, they'd silently share/overwrite the same environment state, which could lead to confusing deployment issues.

2. **The fail-fast is specifically for the template download path** — The main goal is to avoid the expensive template clone + extraction only to fail at the environment prompt. Failing early with a clear message gives the caller (agent or CI script) immediate, actionable guidance on what to provide.

3. **Clear contract for automation** — Non-interactive callers benefit from explicit requirements rather than implicit defaults that may produce unexpected behavior.

That said, auto-generating a unique name (e.g. `unnamed-<short-hash>`) could be an interesting follow-up if the team prefers that direction. Happy to explore it as a separate PR if there's appetite for it.

JeffreyCA (approved):

PR #6790, opened 2026-02-19
Ensure custom user-agent flows to auth HTTP calls

### LINKED ISSUE #2723: Ensure our custom user agent is flow to auth HTTP calls
This isn't happening out of the box today - there are two things we need to do:

1. for places where we leverage existing the existing Azure Identity SDK clients, ensure we pass a ClientOptions that installs our UserAgent policy.
2. For places where we use msal directly, also ensure we are setting a user-agent on these requests.

spboyer:
## Summary

Fixes #2723

The custom azd user-agent string (`azdev/<version>`) was not flowing to authentication HTTP calls made through the Azure Identity SDK and MSAL. This made it impossible to identify azd-originated auth traffic in Azure telemetry.

## Changes

- **`cli/azd/pkg/auth/manager.go`**: Added `userAgent` field to `Manager` struct. New `authClientOptions()` helper returns `azcore.ClientOptions` with `Telemetry.ApplicationID` set to the azd user-agent. Applied to all credential factories: `ManagedIdentity`, `ClientSecret`, `ClientCertificate`, `FederatedToken` (GitHub, AzurePipelines, OIDC). Wrapped MSAL HTTP client with user-agent injection.
- **`cli/azd/pkg/auth/user_agent_client.go`** *(new)*: HTTP client wrapper that injects the user-agent header on all MSAL requests.
- **`cli/azd/cmd/container.go`**: Updated `auth.NewManager` registration to pass `internal.UserAgent()`.
- **Test files**: Updated two call sites to match new `NewManager` signature.

## Approach

Per [comment from @danieljurek](https://github.com/Azure/azure-dev/issues/2723#issuecomment-1981826344), using `ClientOptions.Telemetry.ApplicationID` is the recommended way to mark requests as originating from azd. This prepends the value to the SDK's built-in user-agent string.

For MSAL (which doesn't use `azcore.ClientOptions`), a lightweight HTTP client wrapper injects the header directly.

## Testing

- All existing `pkg/auth` tests pass
- Build succeeds
- No snapshot changes needed

### DIFF
--- a/cli/azd/cmd/container.go
+++ b/cli/azd/cmd/container.go
@@ -634,18 +634,10 @@ func registerCommonDependencies(container *ioc.NestedContainer) {
 			Key:         key,
 		}, nil
 	})
-	container.MustRegisterScoped(func(
-		configManager config.FileConfigManager,
-		userConfigManager config.UserConfigManager,
-		cloud *cloud.Cloud,
-		httpClient auth.HttpClient,
-		console input.Console,
-		externalAuthCfg auth.ExternalAuthConfiguration,
-		azCli az.AzCli,
-	) (*auth.Manager, error) {
-		return auth.NewManager(
-			configManager, userConfigManager, cloud, httpClient, console, externalAuthCfg, azCli, internal.UserAgent())
+	container.MustRegisterSingleton(func() auth.UserAgent {
+		return auth.UserAgent(internal.UserAgent())
 	})
+	container.MustRegisterScoped(auth.NewManager)
 	container.MustRegisterSingleton(azapi.NewUserProfileService)
 	container.MustRegisterScoped(func(authManager *auth.Manager) middleware.CurrentUserAuthManager {
 		return authManager
--- a/cli/azd/pkg/auth/manager.go
+++ b/cli/azd/pkg/auth/manager.go
@@ -101,6 +101,10 @@ type Manager struct {
 	userAgent           string
 }
 
+// UserAgent is a typed string for the application user-agent,
+// used for dependency injection.
+type UserAgent string
+
 type ExternalAuthConfiguration struct {
 	Endpoint    string
 	Key         string
@@ -115,7 +119,7 @@ func NewManager(
 	console input.Console,
 	externalAuthCfg ExternalAuthConfiguration,
 	azCli az.AzCli,
-	userAgent string,
+	userAgent UserAgent,
 ) (*Manager, error) {
 	cfgRoot, err := config.GetUserConfigDir()
 	if err != nil {
@@ -137,7 +141,7 @@ func NewManager(
 		return nil, fmt.Errorf("joining authority url: %w", err)
 	}
 
-	msalClient := newUserAgentClient(httpClient, userAgent)
+	msalClient := newUserAgentClient(httpClient, string(userAgent))
 
 	options := []public.Option{
 		public.WithCache(newCache(cacheRoot)),
@@ -161,7 +165,7 @@ func NewManager(
 		console:             console,
 		externalAuthCfg:     externalAuthCfg,
 		azCli:               azCli,
-		userAgent:           userAgent,
+		userAgent:           string(userAgent),
 	}, nil
 }
 


JeffreyCA (approved):

cli/azd/cmd/container.go
```diff
@@ -633,7 +633,18 @@ func registerCommonDependencies(container *ioc.NestedContainer) {
...
+		userConfigManager config.UserConfigManager,
+		cloud *cloud.Cloud,
+		httpClient auth.HttpClient,
+		console input.Console,
+		externalAuthCfg auth.ExternalAuthConfiguration,
+		azCli az.AzCli,
+	) (*auth.Manager, error) {
+		return auth.NewManager(
+			configManager, userConfigManager, cloud, httpClient, console, externalAuthCfg, azCli, internal.UserAgent())
+	})
```
wbreza: We should be able to just set the user agent to `internal.UserAgent()` within the existing constructor instead of having to override the IoC func. 

Or - create an typed alias for `UserAgentString` or similar and register it against `internal.UserAgent()`

cli/azd/pkg/auth/manager.go
```diff
@@ -135,10 +137,12 @@ func NewManager(
 		return nil, fmt.Errorf("joining authority url: %w", err)
 	}
 
+	msalClient := newUserAgentClient(httpClient, userAgent)
```
wbreza: I'd prefer to see if we can leverage the az sdk pipeline/policy types like we use in other areas instead of wrapping the http client. There was automatically use the existing user agent policy.

cli/azd/pkg/auth/user_agent_client.go
```diff
@@ -0,0 +1,36 @@
...
+		req.Header = make(http.Header)
+	}
+	existingUA := req.Header.Get("User-Agent")
+	if existingUA == "" {
+		req.Header.Set("User-Agent", c.userAgent)
+	} else {
+		req.Header.Set("User-Agent", existingUA+","+c.userAgent)
+	}
+	return c.inner.Do(req)
+}
```
wbreza, line 32: Are we able to leverage the azsdk policy / pipeline with existing user agent policy integration instead of adding this new wrapped HTTP client?

spboyer:
@wbreza Unfortunately not for the MSAL path -- MSAL's `public.New()` and `confidential.New()` only accept a raw `http.Client` via `WithHTTPClient()`. They don't support `azcore.ClientOptions` or the Azure SDK policy/pipeline abstraction.

For the `azcore`-based credential factories (`ManagedIdentityCredential`, `ClientSecretCredential`, etc.), the PR already uses `ClientOptions.Telemetry.ApplicationID` which flows through the SDK's built-in user-agent policy -- so that part is already using the pattern you're describing.

The HTTP client wrapper is only needed for MSAL traffic specifically because MSAL doesn't have an equivalent mechanism. The wrapper is minimal (just prepends the user-agent header on `Do()`).

Options I see:
1. **Keep the wrapper** -- it's the only way to inject user-agent into MSAL requests
2. **Drop MSAL user-agent** -- only tag azcore-based requests (loses visibility into token acquisition traffic)

I'd lean toward option 1 since the wrapper is ~15 lines and gives us full coverage. What do you think?

cli/azd/pkg/auth/user_agent_client.go
spboyer: Unfortunately not for the MSAL path -- MSAL's public.New() and confidential.New() only accept a raw http.Client via WithHTTPClient(). They don't support azcore.ClientOptions or the Azure SDK policy/pipeline abstraction.

For the azcore-based credential factories (ManagedIdentityCredential, ClientSecretCredential, etc.), the PR already uses ClientOptions.Telemetry.ApplicationID which flows through the SDK's built-in user-agent policy.

The HTTP client wrapper is only needed for MSAL traffic because MSAL doesn't have an equivalent mechanism. It's ~15 lines and gives us full coverage of auth requests.

PR #6808, opened 2026-02-19
Improve delegated auth experience

scottaddie:
## Summary

This PR fixes an auth UX issue where azd showed built-in login guidance even when a non-built-in mode (especially az cli/delegated auth) was active, which could mislead users and surface less-helpful errors when no Azure CLI session existed. With the change, unauthenticated messaging is mode-aware (az login for az-cli mode), login guard guidance is clearer, and azd auth status now offers an easy switch back to built-in auth.

### When not authenticated and `auth.useAzCliAuth` is set to `true`:

```
> azd auth status
fetching az cli account: exit code: 1, stdout: , stderr: ERROR: Please run 'az login' to setup account.

Not logged in, run `azd auth login` to login to Azure
```

### When authenticated via Azure CLI and `auth.useAzCliAuth` is set to `true`:

```
> azd auth status
Logged in to Azure as <email_address>
```

### When not authenticated and `auth.useAzCliAuth` is set to `true`:

```
> ./azd.exe auth status
Not logged in to Azure. Run `az login` to log in or authenticate using `azd`.

? Authenticate using azd (recommended)? (y/N)
```

### When authenticated via Azure CLI and `auth.useAzCliAuth` is set to `true`:

```
> ./azd.exe auth status
Logged in to Azure as <email_address>

? Authenticate using azd (recommended)? (y/N)
```

### DIFF
--- a/cli/azd/cmd/auth_status.go
+++ b/cli/azd/cmd/auth_status.go
@@ -124,7 +124,7 @@ func (a *authStatusAction) Run(ctx context.Context) (*actions.ActionResult, erro
 		if !a.flags.global.NoPrompt && !resource.IsRunningOnCI() {
 			a.console.Message(ctx, "")
 			response, err := a.console.Confirm(ctx, input.ConsoleOptions{
-				Message:      "Authenticate using azd (recommended)",
+				Message:      "Authenticate using azd (recommended)?",
 				DefaultValue: false,
 			})
 			if err != nil {


spboyer:
@ccbarragan - can you review here for design?

spboyer:
@scottaddie Nice UX improvement -- here is some feedback from my review:

**1. (High) ErrNoCurrentUser over-wrapping (manager.go:1369)**
The change `fmt.Errorf("...: %w: %w", err, ErrNoCurrentUser)` attaches `ErrNoCurrentUser` to every `az account show` failure, including CLI execution errors, JSON parse failures, etc. Callers that branch on `errors.Is(err, auth.ErrNoCurrentUser)` will now misclassify genuine failures as "not logged in." Should only attach the sentinel when the underlying error actually indicates unauthenticated state.

**2. (Medium) Switch prompt shown for ExternalRequest mode (auth_status.go:122-149)**
The "switch back to built-in?" prompt fires for any non-built-in mode, including `ExternalRequest`. But `SetBuiltInAuthMode()` rejects `ExternalRequest` with an error. So the user gets prompted for an action that will always fail. Consider gating the prompt to `AzDelegated` only, and showing a different message for `ExternalRequest` explaining that the mode is controlled externally.

**3. (Medium) Hard failure on Mode() error (auth_status.go:73-76)**
Previously `auth status` degraded gracefully on config read errors and still showed status. Now it fails outright if `Mode()` returns an error. Consider defaulting to `AzdBuiltIn` on error to maintain the informative behavior.

**4. (Medium) Interactive prompt breaks CI/scripting (auth_status.go:141)**
`auth status` now always emits an interactive Confirm prompt in non-built-in modes. This can break CI/non-interactive usage and `--no-prompt` runs. Consider gating the prompt behind `!NoPrompt && !IsRunningOnCI()` so `auth status` remains safely scriptable.

**5. (Low) Hard-coded "az cli" string (auth_status.go, auth_status_view.go)**
The `AuthStatusView` compares against the literal string `"az cli"` instead of `string(auth.AzDelegated)`. If the display string ever changes, this will silently break. Consider comparing against the typed constant.

**6. (Low) Login guard double-prompt (login_guard.go:75)**
In delegated mode, the middleware shows "run az login" guidance but then still proceeds to offer the `azd auth login` workflow, which is disabled in delegated mode. This creates a confusing double-prompt. Consider short-circuiting after the az login guidance in delegated mode.

cli/azd/cmd/auth_status.go
```diff
@@ -110,7 +115,42 @@ func (a *authStatusAction) Run(ctx context.Context) (*actions.ActionResult, erro
 		return nil, nil
 	}
 
-	a.console.MessageUxItem(ctx, &ux.AuthStatusView{Result: &res})
+	// Show warning and guidance when not using built-in auth
+	if loginMode != auth.AzdBuiltIn {
+		a.console.Message(ctx, "")
+		a.console.MessageUxItem(ctx, &ux.WarningAltMessage{
+			Message: fmt.Sprintf(
+				"azd is not using the built-in authentication mode, but rather '%s'", loginMode),
```
ccbarragan: >  (!) Warning: azd is not using the built-in authentication mode, but rather 'az cli'
>
>Not logged in, run `az login` to login to Azure
>
>To use azd to authenticate instead, select 'y' in the following prompt.
>
>? Do you want to switch back to azd built-in authentication? [? for help] (y/N)

@scottaddie Suggest simplifying and making this more succinct if we can. Is a warning truly needed here, or is it more about listing options for a user (use `az` or `azd` to log in) and providing a recommended default? Maybe something similar to:

>$ azd auth status
>
>Not logged in to Azure. Run `az login` to log in or authenticate using `azd`.
>
> ? Authenticate using `azd` (recommended): (y/N)

scottaddie: @ccbarragan Great feedback! Please see my original PR description for the updated output.

cli/azd/cmd/auth_status.go
```diff
@@ -110,7 +117,37 @@ func (a *authStatusAction) Run(ctx context.Context) (*actions.ActionResult, erro
...
 
-	a.console.MessageUxItem(ctx, &ux.AuthStatusView{Result: &res})
+	a.console.MessageUxItem(ctx, &ux.AuthStatusView{Result: &res, AuthMode: string(loginMode)})
+
+	// Offer to switch back to built-in auth (only for AzDelegated mode)
+	if loginMode == auth.AzDelegated {
+		if !a.flags.global.NoPrompt && !resource.IsRunningOnCI() {
+			a.console.Message(ctx, "")
+			response, err := a.console.Confirm(ctx, input.ConsoleOptions{
+				Message:      "Authenticate using azd (recommended)",
```
ccbarragan: @scottaddie Can we update this to make sure it follows our pattern (end with `:`). Thank you!!

>? Authenticate using azd (recommended):

ccbarragan (changes requested):

cli/azd/cmd/auth_status.go
scottaddie: I chatted with Crystal offline, and we decided to go with a question mark instead.

Issue #6820, opened 2026-02-20
Show actionable next-step hints when `azd ai agent init` fails


spboyer:
## Problem

When \zd ai agent init\ fails in non-interactive mode, error messages are raw and don't suggest fixes. Users must discover workarounds through trial and error.

Examples of unhelpful messages:
- \
o default response for prompt 'Enter a unique environment name:'\ -- should suggest \-e <name>\
- \subscription selection required but cannot prompt in --no-prompt mode\ -- should suggest \zd config set defaults.subscription <id>\
- \cannot prompt for deployment configuration in non-interactive mode\ -- should suggest the flag from #6816

## Proposed Solution

Apply the hint pattern established in core azd (PRs #6801-6804, #6810) to extension error messages:
- Detect specific error conditions
- Display actionable \Hint:\ messages with exact commands to run
- Example: \Hint: Set a default subscription with 'azd config set defaults.subscription <id>'\

## Related Issues

- #6816 - Model deployment flag
- #6817 - Environment name default  
- #6818 - Subscription fallback
- #6819 - Documentation

spboyer:
**Reviewed during issue triage — keeping open.**

Partial progress: the `--model-deployment` no-prompt error path already emits a hint ("use --model-deployment to specify a deployment name in non-interactive mode"). Core azd also now has the `ErrorWithSuggestion` framework from PR #6827 plus the YAML-driven `error_suggestions.yaml` pipeline.

TL;DR remaining work:
- Audit error sites in `cli/azd/extensions/azure.ai.agents/internal/cmd/` and wrap user-fixable failures in `ErrorWithSuggestion` (populate `Err`, `Suggestion`, `Message`, `Links` — see `cli/azd/AGENTS.md` for the completeness rule).
- Specific gaps to cover:
  - Missing environment name → suggest `-e <name>`.
  - Missing subscription in `--no-prompt` → suggest `azd config set defaults.subscription <id>`.
  - Generic Foundry resolution failures → include the `--project-endpoint` / `--subscription` hint (matches the fix in #8093).

therealjohn:
Related #7975

therealjohn:
Closing as this covered by https://github.com/Azure/azure-dev/issues/7975

Issue #6819, opened 2026-02-20
azd ai agent init: Document --no-prompt behavior and ensure CLI flag coverage


spboyer:
## Problem

\zd ai agent init\ silently activates \--no-prompt\ mode when no TTY is detected (CI/CD, IDE terminals, AI assistants). This behavior is:
1. Not documented
2. Causes confusing failures because not all required inputs have corresponding CLI flags

Users encounter a cascade of failures that are only discoverable one at a time:
1. Missing environment name (fixable with \-e\)
2. Missing subscription (fixable with \zd config set\)
3. Missing model deployment (no CLI flag exists -- see #6816)

## Proposed Solution

1. Document the non-TTY auto-\--no-prompt\ behavior in CLI help text and docs
2. Audit all interactive prompts in \zd ai agent init\ and ensure each has a corresponding CLI flag or config default
3. Add a table to docs showing the non-interactive equivalent for each interactive prompt

## Related Issues

- #6816 - Add \--model-deployment\ flag (the most critical gap)
- #6817 - Default environment name
- #6818 - Subscription fallback

spboyer:
**Reviewed during issue triage — keeping open.**

Partial progress: `--model-deployment` from #6816 is now implemented and emits a hint in non-interactive mode.

TL;DR remaining work:
1. **Document** the auto-`--no-prompt`-on-non-TTY behavior in `azd ai agent init --help` text and in the extension docs.
2. **Audit** every interactive prompt in `init.go` / `init_models.go` / `init_from_code.go` and confirm each has a flag, config default, or sensible no-prompt fallback.
3. **Publish a table** in the extension README mapping interactive prompt → CLI flag → `azd config` key → env var.

Linked work: #6816 (done), #6817 (env-name default), #6818 (subscription fallback).

trangevi:
We've made a number of changes to better support --no-prompt, going to close this and we can track individual issues which remain with individual items.

Issue #6816, opened 2026-02-20
azd ai agent init: Add --model-deployment flag for non-interactive usage


spboyer:
## Problem

\zd ai agent init\ is unusable in non-interactive (non-TTY) environments because the model deployment selection step has no CLI flag to bypass the interactive prompt. This blocks CI/CD pipelines, IDE-integrated terminals, AI coding assistants, and any scripted automation.

## Current Behavior

When running in non-TTY mode, \--no-prompt\ is silently activated. The command successfully downloads the manifest, validates the YAML, and loads the model catalog, then fails with:

\\\
failed to prompt for model deployment: rpc error: code = FailedPrecondition
desc = cannot prompt for deployment configuration in non-interactive mode
\\\

\
esolveModelDeploymentNoPrompt()\ in \init_models.go\ exists and attempts auto-selection based on SKU priority and capacity, but when it cannot find a candidate, there is no way to provide the value via CLI flag.

## Proposed Solution

Add flags to allow explicit model deployment specification:

- \--model-deployment <deployment-name>\ - Select an existing deployment by name
- Optionally: \--model-name <name>\ and \--model-version <version>\ for more granular control

When provided, these flags should skip both the interactive prompt and the auto-resolution logic, directly using the specified deployment.

## Impact

P1 blocker for non-interactive adoption. Any non-TTY environment cannot complete \zd ai agent init\ for manifests requiring model deployment configuration.

## Files

- \cli/azd/extensions/azure.ai.agents/internal/cmd/init_models.go\ - Model deployment selection logic
- \cli/azd/extensions/azure.ai.agents/internal/cmd/init.go\ - Flag registration

spboyer:
Reviewed during issue triage. Closing — the `--model-deployment` flag is implemented and shipped.

- Flag registered in `cli/azd/extensions/azure.ai.agents/internal/cmd/init.go` (`-d` / `--model-deployment`).
- Wired through `selectModelDeployment` in `init_from_code.go` and `resolveModelDeployment*` helpers in `init_foundry_resources_helpers.go`.
- Non-interactive mode now surfaces an actionable hint: _"use --model-deployment to specify a deployment name in non-interactive mode"_.

Follow-up gaps (docs audit, broader hint coverage) are tracked in #6819 and #6820.

PR #6823, opened 2026-02-20
Improve azd ai agent init for non-interactive environments

### LINKED ISSUE #6817: azd ai agent init: Default environment name to directory name
## Problem

Unlike \zd init\, which suggests environment names based on the current directory name (e.g., \myapp-dev\, \myapp-test\), \zd ai agent init\ has no default for environment name. In non-TTY environments where \--no-prompt\ is silently activated, this causes an immediate failure:

\\\
no default response for prompt 'Enter a unique environment name:'
\\\

## Proposed Solution

In \nsureEnvironment()\ (\init.go\), when no \-e\ flag is provided and no existing environment is found, default to \ilepath.Base(cwd)\ (matching \zd init\ behavior) instead of requiring an interactive prompt.

## Files

- \cli/azd/extensions/azure.ai.agents/internal/cmd/init.go\ - \nsureEnvironment()\ function

### LINKED ISSUE #6818: azd ai agent init: Auto-select subscription in non-interactive mode
## Problem

\zd ai agent init\ requires interactive subscription selection in non-TTY environments, even when only one subscription is available or a default is already configured.

## Current Behavior

In non-interactive mode, the command fails at subscription selection with no workaround other than manually running \zd config set defaults.subscription <id>\ beforehand.

## Proposed Solution

In \nsureAzureContext()\ (\init.go\), when running non-interactively:

1. Check if \defaults.subscription\ is already set via the account manager (\Manager.GetDefaultSubscriptionID()\) -- if so, use it
2. If not set, call \zdClient.Account().ListSubscriptions()\ to enumerate available subscriptions
3. If exactly one subscription is available, auto-select it
4. If multiple subscriptions exist, fail with a clear error message suggesting \zd config set defaults.subscription <id>\

This uses azd's native subscription management (Azure SDK via \rmsubscriptions\, with caching) rather than falling back to \z CLI\.

## Files

- \cli/azd/extensions/azure.ai.agents/internal/cmd/init.go\ - \nsureAzureContext()\ function
- \cli/azd/pkg/account/manager.go\ - \GetDefaultSubscriptionID()\, subscription management
- \cli/azd/internal/grpcserver/account_service.go\ - gRPC \ListSubscriptions()\ service

spboyer:
## Problem

`azd ai agent init` is unusable in non-interactive (non-TTY) environments because two required inputs (environment name and subscription) have no non-interactive defaults. This blocks CI/CD pipelines, IDE-integrated terminals, AI coding assistants, and scripted automation.

Root cause analysis identified 4 sequential blockers when running without a TTY. This PR addresses the first two:

### 1. Default environment name to directory basename (#6817)

When `--no-prompt` is active and no `-e` flag is provided, the environment name now defaults to the current directory's basename -- consistent with how `azd init` suggests `-dev` etc.

**Before:** `no default response for prompt 'Enter a unique environment name:'`
**After:** Automatically creates environment named after the working directory

### 2. Auto-select subscription in non-interactive mode (#6818)

When `--no-prompt` is active and no subscription is configured, the command now:
- Lists available subscriptions via `azdClient.Account().ListSubscriptions()`
- Auto-selects if exactly one subscription exists
- Returns a clear error with remediation hint if multiple subscriptions exist

**Before:** `failed to prompt for subscription: rpc error: ... cannot prompt in non-interactive mode`
**After:**
- 1 subscription: `Auto-selected subscription: My Sub (guid)`
- Multiple: `multiple subscriptions found but running in non-interactive mode. Set a default with 'azd config set defaults.subscription <id>'`

## Testing

All existing extension tests pass. Changes are in the initialization flow which requires end-to-end testing with Azure resources.

## Related Issues

- #6816 (P1, assigned to @JeffreyCA) -- `--model-deployment` flag (still blocking for full non-interactive flow)
- #6819 (P3, assigned to @alexwolfmsft) -- Document `--no-prompt` behavior
- #6820 (P3, assigned to @JeffreyCA) -- Remediation hints in error messages

Fixes #6817
Fixes #6818

### DIFF
--- a/cli/azd/extensions/azure.ai.agents/internal/cmd/init.go
+++ b/cli/azd/extensions/azure.ai.agents/internal/cmd/init.go
@@ -398,19 +398,9 @@ func ensureEnvironment(ctx context.Context, flags *initFlags, azdClient *azdext.
 						"for default environment name: %w. "+
 						"Pass '-e <name>' explicitly", cwdErr)
 			}
-			name := filepath.Base(cwd)
-			// Sanitize: keep only valid env name characters
-			sanitized := strings.Map(func(r rune) rune {
-				if (r >= 'a' && r <= 'z') ||
-					(r >= 'A' && r <= 'Z') ||
-					(r >= '0' && r <= '9') ||
-					r == '-' || r == '_' || r == '.' {
-					return r
-				}
-				return '-'
-			}, name)
-			if sanitized != "" && sanitized != "-" &&
-				sanitized != "." {
+			if sanitized := sanitizeEnvName(
+				filepath.Base(cwd),
+			); sanitized != "" {
 				envArgs = append(envArgs, sanitized)
 			}
 		}
@@ -529,44 +519,71 @@ func ensureAzureContext(
 
 	if azureContext.Scope.SubscriptionId == "" {
 		if flags.NoPrompt {
-			// In non-interactive mode, try auto-selecting subscription
-			subsResp, listErr := azdClient.Account().ListSubscriptions(
+			// First check for configured default subscription
+			cfgResp, cfgErr := azdClient.UserConfig().GetString(
 				ctx,
-				&azdext.ListSubscriptionsRequest{},
+				&azdext.GetUserConfigStringRequest{
+					Path: "defaults.subscription",
+				},
 			)
-			if listErr != nil {
-				return nil, nil, nil, fmt.Errorf(
-					"failed to list subscriptions: %w. "+
-						"Set a default with "+
-						"'azd config set defaults.subscription <id>'",
-					listErr,
-				)
+			if cfgErr == nil && cfgResp.GetFound() {
+				if v := strings.TrimSpace(
+					cfgResp.GetValue(),
+				); v != "" {
+					azureContext.Scope.SubscriptionId = v
+				}
 			}
 
-			subs := subsResp.GetSubscriptions()
-			switch len(subs) {
-			case 0:
-				return nil, nil, nil, fmt.Errorf(
-					"no subscriptions found. " +
-						"Run 'azd auth login' to authenticate",
+			// Fall back to auto-selecting when only one exists
+			if azureContext.Scope.SubscriptionId == "" {
+				subsResp, listErr := azdClient.Account().ListSubscriptions(
+					ctx,
+					&azdext.ListSubscriptionsRequest{},
 				)
-			case 1:
-				fmt.Printf(
-					"Auto-selected subscription: %s (%s)\n",
-					subs[0].Name, subs[0].Id,
-				)
-				azureContext.Scope.SubscriptionId = subs[0].Id
-				azureContext.Scope.TenantId = subs[0].TenantId
-			default:
-				return nil, nil, nil, fmt.Errorf(
-					"multiple subscriptions found but running " +
-						"in non-interactive mode. " +
-						"Set a default with " +
-						"'azd config set defaults.subscription" +
-						" <id>' or pass '-e <env>' to use an " +
-						"azd environment that already has " +
-						"AZURE_SUBSCRIPTION_ID set",
-				)
+				if listErr != nil {
+					return nil, nil, nil, fmt.Errorf(
+						"failed to list subscriptions: %w. "+
+							"Set a default with "+
+							"'azd config set "+
+							"defaults.subscription <id>'",
+						listErr,
+					)
+				}
+
+				subs := subsResp.GetSubscriptions()
+				switch len(subs) {
+				case 0:
+					return nil, nil, nil, fmt.Errorf(
+						"no subscriptions found. " +
+							"Run 'azd auth login' " +
+							"to authenticate",
+					)
+				case 1:
+					fmt.Printf(
+						"Auto-selected subscription:"+
+							" %s (%s)\n",
+						subs[0].Name, subs[0].Id,
+					)
+					azureContext.Scope.SubscriptionId =
+						subs[0].Id
+					azureContext.Scope.TenantId =
+						subs[0].TenantId
+				default:
+					return nil, nil, nil, fmt.Errorf(
+						"multiple subscriptions " +
+							"found but running in " +
+							"non-interactive mode. " +
+							"Set a default with " +
+							"'azd config set " +
+							"defaults.subscription " +
+							"<id>' or pass '-e " +
+							"<env>' to use an azd " +
+							"environment that " +
+							"already has " +
+							"AZURE_SUBSCRIPTION_ID" +
+							" set",
+					)
+				}
 			}
 		} else {
 			fmt.Print()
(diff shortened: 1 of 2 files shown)


cli/azd/extensions/azure.ai.agents/internal/cmd/init.go
```diff
@@ -480,16 +487,64 @@ func ensureAzureContext(
...
+				azureContext.Scope.TenantId = subs[0].TenantId
+			default:
+				return nil, nil, nil, fmt.Errorf(
+					"multiple subscriptions found but running " +
+						"in non-interactive mode. " +
+						"Set a default with " +
+						"'azd config set defaults.subscription <id>' " +
+						"or pass '-e' with a configured environment",
+				)
+			}
```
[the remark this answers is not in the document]
spboyer: Addressed: extracted sanitizeEnvName() helper and added 12 table-driven test cases covering valid names, special chars, unicode, edge cases (root, dot, dash, empty). The subscription/env integration paths require gRPC mocking which is out of scope for this PR.

cli/azd/extensions/azure.ai.agents/internal/cmd/init.go
```diff
@@ -365,6 +365,13 @@ func ensureEnvironment(ctx context.Context, flags *initFlags, azdClient *azdext.
 		envArgs := []string{"env", "new"}
 		if flags.env != "" {
 			envArgs = append(envArgs, flags.env)
+		} else if flags.NoPrompt {
+			// Default to directory name in non-interactive mode,
+			// consistent with azd init behavior.
+			cwd, err := os.Getwd()
+			if err == nil {
+				envArgs = append(envArgs, filepath.Base(cwd))
+			}
```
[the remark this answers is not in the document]
spboyer: Both addressed: Getwd failure now returns an error with a hint to pass '-e <name>' explicitly. The name is sanitized via sanitizeEnvName() which replaces invalid chars with '-' and returns empty for degenerate inputs (root, dot, dash).

cli/azd/extensions/azure.ai.agents/internal/cmd/init.go
```diff
@@ -480,16 +487,64 @@ func ensureAzureContext(
...
+				&azdext.ListSubscriptionsRequest{},
+			)
+			if listErr != nil {
+				return nil, nil, nil, fmt.Errorf(
+					"failed to list subscriptions: %w. "+
+						"Set a default with "+
+						"'azd config set defaults.subscription <id>'",
+					listErr,
+				)
+			}
```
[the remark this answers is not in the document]
spboyer: Addressed: now queries UserConfig().GetString() for defaults.subscription first. Only falls back to ListSubscriptions auto-select when no default is configured. The hint text also now clarifies the -e flag usage.

cli/azd/extensions/azure.ai.agents/internal/cmd/init.go
```diff
@@ -480,16 +495,93 @@ func ensureAzureContext(
...
+					Path: "defaults.subscription",
+				},
+			)
+			if cfgErr == nil && cfgResp.GetFound() {
+				if v := strings.TrimSpace(
+					cfgResp.GetValue(),
+				); v != "" {
+					azureContext.Scope.SubscriptionId = v
+				}
+			}
```
vhvb1989, line 535: Having a default value to set an environment name when using `--no-prompt` seems like something azd core should be deciding. 
Do we want to have this behavior to be exclusive for the ai.agents extension?

kristenwomack: Is the plan to test this in the extension and then apply to the core? @spboyer

cli/azd/extensions/azure.ai.agents/internal/cmd/init.go
```diff
@@ -365,6 +365,21 @@ func ensureEnvironment(ctx context.Context, flags *initFlags, azdClient *azdext.
 		envArgs := []string{"env", "new"}
 		if flags.env != "" {
 			envArgs = append(envArgs, flags.env)
+		} else if flags.NoPrompt {
```
trangevi, line 391: Would it be better to just pass --no-prompt to the `env new` call that we're making, and rely on however core wants to handle that flag, instead of having special logic in the extension? Seems better for consistency

cli/azd/extensions/azure.ai.agents/internal/cmd/init.go
trangevi: I guess this is the same question I was asking above. I believe we're calling azd handling throughout here, it seems like we should be relying on the azd core/extension framework way of handling the no-prompt flag, not adding our own logic

spboyer:
Good points from all three of you. To clarify the intent:

This was originally a **quick unblock** for AI coding assistants (Copilot CLI, etc.) that need to run `azd ai agent init --no-prompt` end-to-end without a TTY. The core `azd init` flow doesn't hit this code path — the extension has its own `InitAction` that independently prompts for environment name and subscription.

That said, I agree the **principle** should live in core:
- Default env name from directory basename in `--no-prompt` mode
- Subscription auto-select / `defaults.subscription` fallback in `--no-prompt` mode
- Passing `--no-prompt` through to sub-commands like `env new`

**Proposal**: I'll file a core issue to track making `--no-prompt` smarter across all azd commands, and we can reference that from here. For this PR, the question is whether to:

1. **Keep as-is** — extension-local workaround while core catches up
2. **Simplify** — pass `--no-prompt` to `env new` (trangevi's suggestion) and only keep the subscription auto-select logic
3. **Close** — wait for core support

@vhvb1989 @trangevi @kristenwomack what's your preference? Happy to go any direction.

vhvb1989:
> Good points from all three of you. To clarify the intent:
> 
> This was originally a **quick unblock** for AI coding assistants (Copilot CLI, etc.) that need to run `azd ai agent init --no-prompt` end-to-end without a TTY. The core `azd init` flow doesn't hit this code path — the extension has its own `InitAction` that independently prompts for environment name and subscription.
> 
> That said, I agree the **principle** should live in core:
> 
> * Default env name from directory basename in `--no-prompt` mode
> * Subscription auto-select / `defaults.subscription` fallback in `--no-prompt` mode
> * Passing `--no-prompt` through to sub-commands like `env new`
> 
> **Proposal**: I'll file a core issue to track making `--no-prompt` smarter across all azd commands, and we can reference that from here. For this PR, the question is whether to:
> 
> 1. **Keep as-is** — extension-local workaround while core catches up
> 2. **Simplify** — pass `--no-prompt` to `env new` (trangevi's suggestion) and only keep the subscription auto-select logic
> 3. **Close** — wait for core support
> 
> @vhvb1989 @trangevi @kristenwomack what's your preference? Happy to go any direction.

My preference, in order would be 2, 2 and 1 at the end.  -  I would swap 1 to the top if there is some customer blocked or some team is planning to demo/present this in the nearest feature.

cli/azd/extensions/azure.ai.agents/internal/cmd/init.go
spboyer: Good point -- agreed that the extension shouldn't own this behavior. The right approach is to pass `--no-prompt` through to `azd env new` and let core handle the default env name and subscription selection logic. That keeps the extension thin and ensures consistency with how other commands handle non-interactive mode.

I'll rework this to remove the custom non-interactive logic from the extension and rely on core's `--no-prompt` handling instead. If core needs changes to support the auto-env-name or default-subscription fallback, those should be separate PRs against azd core.

spboyer:
Closing in favor of #6934 — the team feedback (thanks @trangevi @vhvb1989 @kristenwomack) is that the non-interactive defaults (auto env name from cwd, subscription auto-selection) belong in azd core, not individual extensions. 

Core doesn't support these yet, so #6934 tracks adding them to `azd env new` / `azd init` directly. Once core handles `--no-prompt` with sensible defaults, the extension gets it for free.

PR #6826, opened 2026-02-20
feat: local template support for `azd init` and developer tooling improvements

jongio:
This PR adds support for using local filesystem directories as template sources in `azd init`, enabling template developers to iterate on templates without pushing to a remote repository. It also adds developer tooling (`mage devinstall`, `mage preflight`) and fixes several pre-existing issues required for CI gates to pass.

### 1. Local Template Support (core feature)

**Files:** `pkg/templates/path.go`, `pkg/templates/path_test.go`, `internal/repository/initializer.go`, `internal/repository/initializer_test.go`, `cmd/init.go`

#### `pkg/templates/path.go` — Template path resolution
- **What:** Extended `Absolute()` to recognize local filesystem directories as valid template sources, in addition to existing GitHub shorthand and remote git URLs.
- **Why:** Previously, `azd init --template <path>` only accepted remote URLs or GitHub `owner/repo` shorthand. Local paths were silently misinterpreted as GitHub repository names, leading to confusing errors.
- **Details:**
  - Added `os.Lstat()` check: if the path resolves to an existing directory, return its absolute path. Symlinks are explicitly rejected for security consistency with `copyLocalTemplate`.
  - Returns a clear "not a directory" error when the path exists but is not a directory (e.g., a file), instead of falling through to GitHub resolution.
  - Added `looksLikeLocalPath()` helper: if the path looks like an explicit local reference (`./`, `../`, absolute path) but the directory doesn't exist, return a clear "does not exist" error.
  - Extracted shared `remoteURIPrefixes`/`isRemoteURI()` helper covering `git@`, `git://`, `ssh://`, `file://`, `http://`, `https://` — used by both `Absolute()` and `IsLocalPath()` to avoid list drift.
  - Added `IsLocalPath()` function for downstream code to distinguish local vs remote resolved paths.

#### `pkg/templates/path_test.go` — Comprehensive test coverage
- **What:** New test file with table-driven tests covering all path resolution scenarios.
- **Why:** The path resolution logic now has multiple code paths (remote URLs, GitHub shorthand, local directories, explicit local paths that don't exist) that need regression protection.
- **Details:** Tests cover: git URIs, git://, ssh://, file://, http/https URLs, Azure Samples shorthand, owner/repo, local directories (absolute and relative), non-existent explicit local paths, symlink rejection, and files that aren't directories. Uses `t.Chdir()` and `filepath.EvalSymlinks` normalization to handle macOS symlinks and Windows 8.3 short names.

#### `internal/repository/initializer.go` — Local template copying
- **What:** Added `copyLocalTemplate()` method, `findExecutableFiles()` helper, and integrated them into the `Initialize()` flow.
- **Why:** Remote templates are fetched via `git clone`, but local templates should be copied directly from the filesystem to preserve uncommitted changes — the primary use case for local template development.
- **Details:**
  - `copyLocalTemplate()` copies the directory contents while: skipping `.git/`, skipping symlinks (security), respecting root-level `.gitignore` rules (with proper error handling for unreadable `.gitignore`), and validating the source is a real directory (not a symlink, mitigating TOCTOU).
  - `findExecutableFiles()` scans copied files for executable permission bits so `gitInitialize` can preserve them with `git update-index --chmod=+x`. Skips on Windows where exec bits are not meaningful.
  - `Initialize()` now branches on `templates.IsLocalPath()` — local paths use `copyLocalTemplate()`, remote paths use `fetchCode()` (existing git clone logic). Spinner message updated accordingly.

#### `internal/repository/initializer_test.go` — Initializer tests
- **What:** New tests for local template initialization behavior.
- **Why:** The branching logic in `Initialize()` and the filtering logic in `copyLocalTemplate()` need test coverage.
- **Details:** Tests cover: basic local template copy, `.git` directory exclusion, `.gitignore` filtering, and `.gitignore` negation patterns.

#### `cmd/init.go` — Minor wording fix
- **What:** Trivial string change in init command.

---

### 2. Developer Tooling (`mage devinstall`, `mage preflight`)

**Files:** `magefile.go`, `CONTRIBUTING.md`, `.vscode/cspell-azd-dictionary.txt`, `cmd/testdata/TestFigSpec.ts`, `cmd/testdata/TestUsage-azd-init.snap`

#### `magefile.go` — Build targets
- **What:** Added three mage targets: `DevInstall`, `DevUninstall`, and `Preflight`.
- **Why:** There was no unified way to run all pre-commit quality checks. Developers had to run gofmt, golangci-lint, cspell, copyright-check, build, and test separately with different invocations documented across AGENTS.md.
- **Details:**
  - `mage devinstall` — Builds azd from source as `azd-dev` (avoids conflicting with production install), installs to `~/.azd/bin`, and adds it to PATH.
  - `mage devuninstall` — Removes `azd-dev` and cleans up PATH entries.
  - `mage preflight` — Runs all 6 CI quality checks in sequence: gofmt, copyright headers (via bash/WSL on Windows), golangci-lint, cspell, `go build`, `go test -short`. Fails fast with clear install instructions if any required tool is missing. Tool versions pinned to match CI (`golangci-lint@v2.6`, `cspell@8.13.1`).
  - Sets `GOWORK=off` during preflight to mirror CI environment (prevents local `go.work` files from masking module resolution differences).
  - Shell quoting: all path interpolation in shell commands uses `shellQuote()` (POSIX single-quote escaping). PowerShell commands use `param()` pattern for safe argument passing.
  - WSL detection cached via `sync.Once` for performance.

#### `CONTRIBUTING.md` — Documentation
- **What:** Added "Developer Install" and "Preflight" sections documenting the new mage targets.
- **Why:** New contributors need to know these targets exist.

#### `.vscode/cspell-azd-dictionary.txt` — Spell check dictionary
- **What:** Added `devinstall`, `devuninstall`, `jongio`, `nazd`, `TOCTOU`, `Veyor` to the custom dictionary.
- **Why:** These terms appear in new code and were flagged by cspell.

#### `cmd/testdata/TestFigSpec.ts`, `cmd/testdata/TestUsage-azd-init.snap` — Snapshot updates
- **What:** Auto-generated snapshot updates from running `UPDATE_SNAPSHOTS=true go test ./cmd -run 'TestFigSpec|TestUsage'`.
- **Why:** The `devinstall` and `devuninstall` commands added new entries to the command tree.

---

### 3. OpenTelemetry semconv v1.39.0 → v1.37.0 (revert)

**Files:** `internal/tracing/fields/fields.go`, `internal/tracing/resource/resource.go`, `internal/telemetry/storage_exporter_test.go`, `internal/telemetry/appinsights-exporter/span_to_envelope_test.go`

- **What:** Reverted semconv imports from `v1.39.0` back to `v1.37.0` in 4 files.
- **Why:** The semconv `v1.39.0` package only exists in otel module `v1.40.0+`, but `go.mod` pins otel at `v1.38.0`. Locally this worked because a `go.work` file outside the repo pulled in a higher otel version via MVS, but CI (which has no `go.work`) failed with `could not import go.opentelemetry.io/otel/semconv/v1.39.0`.
- **Details:** Two constants removed in the attempted v1.39.0 update (`RPCJSONRPCRequestIDKey`, `RPCJSONRPCErrorCodeKey`) were restored as local `attribute.Key()` definitions using the same string values, since they were deprecated upstream.

---

### 4. Deprecated API fixes (SA1019 staticcheck)

**Files:** `pkg/azapi/azure_client_test.go`, `pkg/azsdk/correlation_policy_test.go`, `pkg/azsdk/user_agent_policy_test.go`, `pkg/graphsdk/entity_list_request_builder_test.go`

- **What:** Replaced `runtime.WithCaptureResponse` with `policy.WithCaptureResponse` in 4 test files. Removed now-unused `runtime` imports.
- **Why:** `runtime.WithCaptureResponse` is deprecated (SA1019). CI's staticcheck catches this. The replacement `policy.WithCaptureResponse` is the recommended API and is functionally identical.

---

### 5. Test robustness fixes

**Files:** `internal/runcontext/agentdetect/detect_test.go`, `cmd/auto_install_integration_test.go`, `cmd/auto_install_test.go`, `internal/appdetect/appdetect_test.go`, `pkg/infra/provisioning/terraform/terraform_provider_test.go`, `pkg/state/state_cache_test.go`

#### Agent detection tests — `detect_test.go`, `auto_install_*_test.go`
- **What:** Added `skipIfProcessDetectsAgent()` helper. Tests that assert "no agent detected" are skipped when running inside a CI agent (e.g., Copilot CLI, GitHub Actions).
- **Why:** The agent detection logic inspects the parent process tree. When running inside an agent, the process tree contains agent identifiers, making "no agent detected" assertions impossible to satisfy.

#### App detection tests — `appdetect_test.go`
- **What:** Added Maven availability check; Java-dependent subtests are skipped when `mvn` is not installed.
- **Why:** The test runs `mvn` to detect Java projects. On machines without Maven, the test fails with a confusing "executable not found" error instead of skipping gracefully.

#### Terraform provider tests — `terraform_provider_test.go`
- **What:** Added `skipIfTerraformNotInstalled()` helper, applied to 4 test functions.
- **Why:** These tests call `tools.EnsureInstalled()` which checks `exec.LookPath("terraform")` against the real PATH before any mocked commands run.

#### State cache TTL test — `state_cache_test.go`
- **What:** Increased TTL from 100ms→500ms and sleep from 150ms→600ms.
- **Why:** The 100ms TTL was too tight — on loaded systems, the Save→Load round-trip could exceed 100ms, causing intermittent failures.

---

### 6. Formatting fixes (gofmt, lll)

**Files:** `extensions/azure.ai.models/pkg/models/pending_upload.go`, `extensions/microsoft.azd.extensions/internal/cmd/build.go`

- **What:** `pending_upload.go` was reformatted by `gofmt -s -w`. `build.go` had a long line broken to satisfy the 125-char `lll` linter rule.
- **Why:** CI runs `gofmt` check. These were pre-existing formatting issues on the branch.

## Testing

All changes verified locally with `mage preflight` which runs:
- ✅ `gofmt` — 0 issues
- ✅ Copyright headers — all files pass
- ✅ `golangci-lint` (v2.6, matching CI) — 0 issues
- ✅ `cspell` — 0 issues
- ✅ `go build` — passes
- ✅ `go test -short` — all pass

### DIFF
--- a/cli/azd/internal/repository/initializer.go
+++ b/cli/azd/internal/repository/initializer.go
@@ -89,6 +89,18 @@ func (i *Initializer) Initialize(
 		return err
 	}
 
+	// Reject overlapping source and destination for local templates.
+	// This catches "azd init -t ." (same dir) and "azd init -t .." (destination inside source).
+	if templates.IsLocalPath(templateUrl) {
+		rel, relErr := filepath.Rel(templateUrl, target)
+		if relErr == nil && !strings.HasPrefix(rel, "..") {
+			return fmt.Errorf(
+				"project directory (%s) overlaps with template source (%s); "+
+					"run 'azd init' from a directory outside the template",
+				target, templateUrl)
+		}
+	}
+
 	var stepMessage string
 	if templates.IsLocalPath(templateUrl) {
 		stepMessage = fmt.Sprintf(
--- a/cli/azd/magefile.go
+++ b/cli/azd/magefile.go
@@ -16,12 +16,15 @@ import (
 	"sync"
 )
 
-// DevInstall builds azd from source as 'azd-dev' and installs it to ~/.azd/bin.
+// Dev contains developer tooling commands for building and installing azd from source.
+type Dev struct{}
+
+// Install builds azd from source as 'azd-dev' and installs it to ~/.azd/bin.
 // The binary is named azd-dev to avoid conflicting with a production azd install.
 // Automatically adds ~/.azd/bin to PATH if not already present.
 //
-// Usage: mage devinstall
-func DevInstall() error {
+// Usage: mage dev:install
+func (Dev) Install() error {
 	repoRoot, err := findRepoRoot()
 	if err != nil {
 		return err
@@ -74,11 +77,11 @@ func DevInstall() error {
 	return nil
 }
 
-// DevUninstall removes the azd-dev binary from ~/.azd/bin.
+// Uninstall removes the azd-dev binary from ~/.azd/bin.
 // The PATH entry is left intact.
 //
-// Usage: mage devuninstall
-func DevUninstall() error {
+// Usage: mage dev:uninstall
+func (Dev) Uninstall() error {
 	dir, err := installDir()
 	if err != nil {
 		return err
@@ -476,7 +479,7 @@ func addToPathUnix(dir string) error {
 	}
 	defer f.Close()
 
-	if _, err := fmt.Fprintf(f, "\n# Added by azd devinstall\n%s\n", exportLine); err != nil {
+	if _, err := fmt.Fprintf(f, "\n# Added by azd dev:install\n%s\n", exportLine); err != nil {
 		return fmt.Errorf("writing to %s: %w", rcFile, err)
 	}
 
--- a/cli/azd/.vscode/cspell-azd-dictionary.txt
+++ b/cli/azd/.vscode/cspell-azd-dictionary.txt
@@ -104,8 +104,6 @@ devcenters
 devcentersdk
 devdeviceid
 devel
-devinstall
-devuninstall
 deviceid
 diffmatchpatch
 discarder
--- a/cli/azd/CONTRIBUTING.md
+++ b/cli/azd/CONTRIBUTING.md
@@ -40,7 +40,7 @@ Install [mage](https://magefile.org/) (`go install github.com/magefile/mage@late
 
 ```bash
 cd cli/azd
-mage devinstall
+mage dev:install
 ```
 
 This builds `azd-dev` (to avoid conflicting with a production `azd` install) with version
--- a/cli/azd/internal/repository/initializer_test.go
+++ b/cli/azd/internal/repository/initializer_test.go
@@ -1127,3 +1127,42 @@ func Test_Initializer_Initialize_LocalTemplateNestedGitignore(t *testing.T) {
 	// Verify the .tmp file was excluded by nested .gitignore
 	require.NoFileExists(t, filepath.Join(projectDir, "subproject", "data.tmp"))
 }
+
+func Test_Initializer_Initialize_LocalTemplateOverlapRejected(t *testing.T) {
+	// "azd init -t ." (same directory) and "azd init -t .." (destination inside source)
+	// must be rejected with a clear error.
+	localTemplateDir := createLocalTemplateDir(t, testDataPath("template"))
+
+	realRunner := exec.NewCommandRunner(nil)
+	mockEnv := &mockenv.MockEnvManager{}
+
+	i := NewInitializer(
+		mockinput.NewMockConsole(),
+		git.NewCli(realRunner),
+		dotnet.NewCli(realRunner),
+		alpha.NewFeaturesManagerWithConfig(config.NewEmptyConfig()),
+		lazy.From[environment.Manager](mockEnv),
+	)
+
+	t.Run("SameDirectory", func(t *testing.T) {
+		// Project dir == template dir
+		azdCtx := azdcontext.NewAzdContextWithDirectory(localTemplateDir)
+		err := i.Initialize(context.Background(), azdCtx, &templates.Template{
+			RepositoryPath: localTemplateDir,
+		}, "")
+		require.Error(t, err)
+		assert.Contains(t, err.Error(), "overlaps with template source")
+	})
+
+	t.Run("DestinationInsideSource", func(t *testing.T) {
+		// Project dir is a subdirectory of the template dir
+		subDir := filepath.Join(localTemplateDir, "child")
+		require.NoError(t, os.MkdirAll(subDir, 0755))
+		azdCtx := azdcontext.NewAzdContextWithDirectory(subDir)
+		err := i.Initialize(context.Background(), azdCtx, &templates.Template{
+			RepositoryPath: localTemplateDir,
+		}, "")
+		require.Error(t, err)
+		assert.Contains(t, err.Error(), "overlaps with template source")
+	})
+}


cli/azd/magefile.go
```diff
@@ -0,0 +1,453 @@
...
+	switch shell {
+	case "zsh":
+		rcFile = filepath.Join(home, ".zshrc")
+		exportLine = fmt.Sprintf(`export PATH="%s:$PATH"`, dir)
+	case "fish":
+		rcFile = filepath.Join(home, ".config", "fish", "config.fish")
+		exportLine = fmt.Sprintf("fish_add_path %s", dir)
+	default: // bash and others
+		rcFile = filepath.Join(home, ".bashrc")
+		exportLine = fmt.Sprintf(`export PATH="%s:$PATH"`, dir)
```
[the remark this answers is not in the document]
jongio: Fixed — added a `shellQuote` helper that wraps the path in single quotes with proper escaping for embedded single quotes. Export lines now use `shellQuote(dir)` instead of raw string interpolation.

cli/azd/magefile.go
```diff
@@ -0,0 +1,453 @@
...
+	if len(p) >= 2 && p[1] == ':' {
+		drive := strings.ToLower(string(p[0]))
+		rest := p[2:]
+
+		// Detect WSL vs Git-for-Windows bash by checking if /mnt exists in the shell.
+		out, err := exec.Command(shell, "-c", "test -d /mnt/c && echo wsl").Output()
+		if err == nil && strings.TrimSpace(string(out)) == "wsl" {
+			return "/mnt/" + drive + rest
+		}
+		return "/" + drive + rest
```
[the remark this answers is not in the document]
jongio: Fixed — detection is now cached using `sync.Once` so the shell command runs at most once per process.

cli/azd/internal/repository/initializer.go
```diff
@@ -90,7 +88,25 @@ func (i *Initializer) Initialize(
...
+			"Downloading template code to: %s", output.WithLinkFormat("%s", azdCtx.ProjectDirectory()))
+	}
+	i.console.ShowSpinner(ctx, stepMessage, input.Step)
+	defer func() {
+		i.console.StopSpinner(ctx, stepMessage+"\n", input.GetStepResultFormat(err))
+	}()
+
+	var filesWithExecPerms []string
+	if templates.IsLocalPath(templateUrl) {
+		err = i.copyLocalTemplate(templateUrl, staging)
```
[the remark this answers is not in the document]
jongio: Fixed — added a `findExecutableFiles` function that walks the staging directory after copy and returns files with exec bits set. These are now passed to `gitInitialize` just like remote templates.

cli/azd/extensions/microsoft.azd.extensions/internal/cmd/build.go
```diff
@@ -311,6 +316,34 @@ func copyBinaryFiles(extensionId, sourcePath, destPath string) error {
...
+
+	// Kill any processes running from the install directory
+	if installDir != "" {
+		//nolint:gosec // G204: installDir is derived from config; single quotes are escaped
+		_ = exec.Command("powershell", "-NoProfile", "-Command",
+			fmt.Sprintf(
+				"Get-Process | Where-Object { $_.Path -and $_.Path.StartsWith('%s') }"+
+					" | Stop-Process -Force -ErrorAction SilentlyContinue",
+				escapePowerShellSingleQuotes(installDir),
+			)).Run()
```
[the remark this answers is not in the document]
jongio: Fixed — `installDir` is now passed as a PowerShell parameter via `param([string]$p)` instead of being interpolated into the script string.

cli/azd/pkg/templates/path.go
```diff
@@ -16,11 +18,33 @@ import (
...
 	}
 
+	// Support local filesystem directories as template sources.
+	// This allows using a local directory with uncommitted changes for template development.
+	if info, err := os.Stat(path); err == nil && info.IsDir() {
+		absPath, err := filepath.Abs(path)
+		if err != nil {
+			return "", fmt.Errorf("resolving local template path: %w", err)
+		}
+		return absPath, nil
```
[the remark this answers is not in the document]
jongio: Fixed — `Absolute` now uses `os.Lstat` and explicitly rejects symlinks with a clear error message, consistent with `copyLocalTemplate`. Added a test for the symlink rejection.

cli/azd/magefile.go
```diff
@@ -0,0 +1,453 @@
...
+		}
+	}
+
+	// Persistently prepend to user-level PATH via PowerShell.
+	script := fmt.Sprintf(
+		`$current = [Environment]::GetEnvironmentVariable('PATH', 'User'); `+
+			`[Environment]::SetEnvironmentVariable('PATH', '%s;' + $current, 'User')`,
+		dir,
+	)
+	cmd := exec.Command("powershell", "-NoProfile", "-Command", script)
```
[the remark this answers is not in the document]
jongio: Fixed — the directory path is now passed as a PowerShell parameter via `param([string]$dir)` instead of being interpolated into the script string.

cli/azd/pkg/templates/path.go
```diff
@@ -6,21 +6,66 @@ package templates
...
+			return absPath, nil
+		}
+	}
+
+	// If the path looks like an explicit local path reference (".", "..", starts with ./, ../, .\, ..\,
+	// or is an absolute path) but the directory wasn't found above, return a clear error
+	// instead of falling through to GitHub resolution which would give a confusing error.
+	if looksLikeLocalPath(path) {
+		return "", fmt.Errorf("local template directory '%s' does not exist", path)
+	}
```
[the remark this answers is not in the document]
jongio: Fixed. Absolute() now returns an explicit `not a directory` error when os.Lstat succeeds but the path is not a directory (line 61). The `does not exist` branch at line 67 now only triggers for truly non-existent paths that look like local references.

cli/azd/pkg/templates/path_test.go
```diff
@@ -0,0 +1,201 @@
...
+	filePath := filepath.Join(dir, "myfile")
+	require.NoError(t, os.WriteFile(filePath, []byte("content"), 0600))
+
+	// Even though a file exists at this path, since it's not a directory,
+	// Absolute() should fall through and NOT use it as a local template.
+	// Using just the filename "myfile" to test this cleanly.
+	result, err := Absolute("myfile")
+	require.NoError(t, err)
+	// Falls through to Azure Samples resolution
+	require.Equal(t, "https://github.com/Azure-Samples/myfile", result)
```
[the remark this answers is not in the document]
jongio: Fixed. Test now passes the full ilePath to Absolute() and asserts `not a directory` error, matching the new behavior in Absolute() that rejects non-directory paths.

cli/azd/internal/repository/initializer.go
```diff
@@ -166,6 +186,102 @@ func (i *Initializer) fetchCode(
...
+	}
+	if !info.IsDir() {
+		return fmt.Errorf("local template path '%s' is not a directory", source)
+	}
+
+	// Load .gitignore rules from the source template if present.
+	// Only the root .gitignore is loaded; nested .gitignore files are not supported.
+	var ignorer gitignore.GitIgnore
+	gitignorePath := filepath.Join(source, ".gitignore")
+	if f, err := os.Open(gitignorePath); err == nil {
```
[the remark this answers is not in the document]
jongio: Fixed. Now distinguishes between missing .gitignore (os.IsNotExist - silently skip) and other open errors (return error). This prevents silently copying everything when .gitignore exists but can't be read due to permissions or IO errors.

cli/azd/internal/repository/initializer.go
```diff
@@ -90,7 +89,28 @@ func (i *Initializer) Initialize(
...
+		stepMessage = fmt.Sprintf(
+			"Copying template code from local path to: %s", output.WithLinkFormat("%s", azdCtx.ProjectDirectory()))
+	} else {
+		stepMessage = fmt.Sprintf(
+			"Downloading template code to: %s", output.WithLinkFormat("%s", azdCtx.ProjectDirectory()))
+	}
+	i.console.ShowSpinner(ctx, stepMessage, input.Step)
+	defer func() {
+		i.console.StopSpinner(ctx, stepMessage+"\n", input.GetStepResultFormat(err))
+	}()
```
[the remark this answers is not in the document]
jongio: Fixed. Removed the explicit StopSpinner call at line 157, relying solely on the defer at line 101-102 which correctly captures the final rr value.

cli/azd/magefile.go
```diff
@@ -0,0 +1,487 @@
...
+			failed = true
+		}
+	}
+
+	// Check required tools are installed before running anything.
+	if err := requireTool("golangci-lint",
+		"go install github.com/golangci/golangci-lint/cmd/golangci-lint@latest"); err != nil {
+		return err
+	}
+	if err := requireTool("cspell", "npm install -g cspell"); err != nil {
```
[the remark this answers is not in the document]
jongio: Fixed. Pinned golangci-lint@v2.6 and cspell@8.13.1 to match the CI versions in .github/workflows/cli-ci.yml.

cli/azd/pkg/templates/path.go
```diff
@@ -6,21 +6,66 @@ package templates
...
+		if info.Mode()&os.ModeSymlink != 0 {
+			return "", fmt.Errorf("local template directory '%s' is a symlink, which is not supported", path)
+		}
+		if info.IsDir() {
+			absPath, err := filepath.Abs(path)
+			if err != nil {
+				return "", fmt.Errorf("resolving local template path: %w", err)
+			}
+			return absPath, nil
+		}
```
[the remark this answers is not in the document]
jongio: Fixed. Absolute() now returns `local template path '...' exists but is not a directory` when os.Lstat succeeds but the path is not a directory, instead of falling through to GitHub resolution or the misleading `does not exist` error.

jongio:
> Three issues found:
> 
> 1. .git file from local template can be copied into target (High)
>    The copy filter skips .git only when it is a directory (info.IsDir()). In Git worktrees/submodules, .git is often a file pointing to an external gitdir. That file will be copied and can break git operations in the target. Exclude .git by name regardless of file type.
> 2. Local-template .gitignore handling is incomplete (Medium)
>    Only the root .gitignore is parsed; nested .gitignore rules are ignored. This can copy files that are intentionally ignored in subdirectories (secrets, build artifacts). Consider using tracked-files-only semantics or evaluating nested .gitignore rules.
> 3. Existing local directory silently overrides remote template resolution (Medium)
>    Absolute() now checks os.Lstat first and returns local path if directory exists. Inputs like owner/repo will resolve locally if a same-named directory exists, silently skipping remote template. Consider requiring explicit local syntax (./..., ../..., or absolute path) for local templates.

Resolved

cli/azd/cmd/auto_install_test.go
```diff
@@ -396,6 +396,14 @@ func TestParseGlobalFlags_AgentDetection(t *testing.T) {
...
 			agentdetect.ResetDetection()
 
+			// If the test expects no agent but we're inside an agent process, skip it.
+			if !tt.expectedNoPrompt && len(tt.envVars) == 0 {
+				if agentdetect.GetCallingAgent().Detected {
+					t.Skip("skipping: parent process detection found an agent")
+				}
+				agentdetect.ResetDetection()
+			}
+
```
vhvb1989, line 406: why?

cli/azd/extensions/azure.ai.models/pkg/models/pending_upload.go
```diff
@@ -5,20 +5,20 @@ package models
...
-	Credential           BlobCredential  `json:"credential"`
-	IsSingleFile         bool            `json:"isSingleFile"`
-	BlobManifestDigest   *string         `json:"blobManifestDigest"`
-	ConnectionName       *string         `json:"connectionName"`
+	BlobURI             string         `json:"blobUri"`
+	StorageAccountArmID string         `json:"storageAccountArmId"`
+	Credential          BlobCredential `json:"credential"`
+	IsSingleFile        bool           `json:"isSingleFile"`
+	BlobManifestDigest  *string        `json:"blobManifestDigest"`
+	ConnectionName      *string        `json:"connectionName"`
```
vhvb1989, line 21: Maybe don't make this part of your PR, 
It should be a fix to the extension.

cli/azd/extensions/microsoft.azd.extensions/internal/cmd/build.go
```diff
@@ -311,6 +316,34 @@ func copyBinaryFiles(extensionId, sourcePath, destPath string) error {
...
+	// Pass installDir as a parameter to avoid injection from special characters.
+	if installDir != "" {
+		//nolint:gosec // G204: installDir is derived from config and passed as an argument, not interpolated
+		_ = exec.Command("powershell", "-NoProfile", "-Command",
+			"param([string]$p) Get-Process | Where-Object { $_.Path -and $_.Path.StartsWith($p) }"+
+				" | Stop-Process -Force -ErrorAction SilentlyContinue",
+			installDir,
+		).Run()
+	}
+}
```
vhvb1989, line 345: This is also unrelated, right?

cli/azd/internal/appdetect/appdetect_test.go
```diff
@@ -185,6 +186,16 @@ func TestDetect(t *testing.T) {
...
+			// Skip subtests that include Java projects when Maven is not installed.
+			for _, p := range tt.want {
+				if p.Language == Java {
+					if _, err := exec.LookPath("mvn"); err != nil {
+						t.Skip("skipping: Maven is not installed")
+					}
+					break
+				}
+			}
+
```
vhvb1989, line 198: why? 
Users decide what tests to filter. We should not do automatic filtering

cli/azd/internal/repository/initializer.go
```diff
@@ -90,7 +89,28 @@ func (i *Initializer) Initialize(
 		return err
 	}
 
-	filesWithExecPerms, err := i.fetchCode(ctx, templateUrl, templateBranch, staging)
+	var stepMessage string
+	if templates.IsLocalPath(templateUrl) {
+		stepMessage = fmt.Sprintf(
+			"Copying template code from local path to: %s", output.WithLinkFormat("%s", azdCtx.ProjectDirectory()))
```
vhvb1989, line 107: How is this better than just running `cp -r local-template copy-here` ?  (or equivalent in Windows).

cli/azd/extensions/microsoft.azd.extensions/internal/cmd/build.go
jongio: It's directly related — `mage devinstall` (added in this PR) fails on Windows without this because running extension binaries hold file locks that block the copy. This resolves that.

cli/azd/extensions/azure.ai.models/pkg/models/pending_upload.go
jongio: This is a `gofmt` alignment fix flagged by `mage preflight`, which is part of this PR's developer tooling. It's whitespace-only and keeps preflight green.

cli/azd/cmd/auto_install_test.go
jongio: This test fails when run inside an agent process (e.g., via `mage preflight` in a Copilot CLI session) because parent process detection finds a real agent and overrides the expected baseline result. The skip guard is narrowly scoped to only those cases and the skip message makes it explicit.

cli/azd/internal/repository/initializer.go
jongio: `cp -r` copies everything including `.git`, symlinks, and gitignored content like `node_modules` and build artifacts. This filters all of that, works cross-platform without shell commands, and mirrors what the remote template path already does.

cli/azd/internal/appdetect/appdetect_test.go
jongio: Tests that require external tools should skip when those tools aren't installed — that's standard Go practice. Without this, any contributor without Maven gets unrelated test failures from `go test ./...`. The skip message is explicit.

cli/azd/internal/repository/initializer.go
vhvb1989: Have you considered something like `azd template clone` (or export)?  
Assuming you already have a template/project locally, you can skip switching to another folder to run `azd init -t /t/folder` by asking azd to create the clone of the current template into some folder.

In both cases, I think you need to check and return error if you use the same source as destination, like `azd init -t .` (init from the same folder). Or `azd init -t ../.` (init from parent or any subparent?).

cli/azd/magefile.go
```diff
@@ -0,0 +1,487 @@
...
+	"strings"
+	"sync"
+)
+
+// DevInstall builds azd from source as 'azd-dev' and installs it to ~/.azd/bin.
+// The binary is named azd-dev to avoid conflicting with a production azd install.
+// Automatically adds ~/.azd/bin to PATH if not already present.
+//
+// Usage: mage devinstall
+func DevInstall() error {
```
wbreza: Can we reorganize the commands to use `mage dev:install` and `mage dev:uninstall`. I like the organization that you can similar to the org structure for NPM scripts.

wbreza (approved):
Looks good - just a couple minor comments.

cli/azd/cmd/auto_install_test.go
wbreza: Looks like a well tested set of changes. Ship it 🚀

cli/azd/internal/repository/initializer.go
jongio: Re: azd template clone/export: Good idea for a follow-up. The use case here is "consumer testing a template before publishing" (cd to empty dir, azd init -t ../my-template). A template clone command would serve the inverse workflow "author exporting from inside the template 
  dir."
  
Re: same source/destination check: Good catch, there's no guard against this today. I'll add an overlap check that rejects when source == target or when the project directory is inside the template source (e.g., azd init -t . or azd init -t ..).

cli/azd/magefile.go
jongio: Changed

cli/azd/internal/repository/initializer.go
jongio: Pushed thanks!

wbreza (approved):
LGTM

PR #6835, opened 2026-02-20
Add `azdext.Run` lifecycle and structured extension error transport

### LINKED ISSUE #6806: Extensions Framework - Structured extension error contract, transport, telemetry mapping
Parent: #6728

## Summary
Build a complete extension error framework in `azd` so extension failures are captured as typed, actionable telemetry instead of generic process failures.

## Problem
Extension failures are often reported with low diagnostic value, especially for custom extension commands. We need a standard way for extensions to communicate structured error intent, and for `azd` to map that intent into stable telemetry result codes.

## Goals
- Define a stable extension error contract that supports common failure classes.
- Support structured error reporting for both:
  - extension listen/gRPC flows
  - non-listen custom command execution flows
- Update core telemetry mapping so typed extension errors produce specific result codes.
- Preserve backward compatibility when structured error data is not present.
- Add tests, docs, and rollout measurements.

## Scope
- Error contract/schema updates in extension protocol and `azdext` helper APIs.
- Transport mechanism for structured errors from custom command executions.
- `MapError` updates in core for typed mapping and fallback behavior.
- Validation coverage in unit/integration tests.
- Extension framework docs with clear author guidance and examples.
- Adoption telemetry:
  - typed vs generic extension failures
  - top extension error categories by extension/version

## Acceptance Criteria
- Typed errors from extensions map to specific, stable ResultCode families.
- Custom command and gRPC/listen paths both support typed error flow.
- Legacy extensions that do not emit typed errors still function and map via fallback behavior.
- Tests cover typed mapping and fallback cases.
- Documentation includes concrete guidance for extension authors.
- Initial adoption metrics are available to track reduction of generic failures.

JeffreyCA:
Resolves #6806

This PR introduces two related improvements for azd extensions:

1. **Structured error transport** - typed error types (`LocalError`, `ServiceError`) that flow structured error data from extensions back to the host for telemetry and UX.
2. **`azdext.Run`** - a standard lifecycle entry point that eliminates ~30 lines of boilerplate from every extension's `main.go`, and comes out of the box with error reporting back to the host.

This PR also integrates with azd core's [new `ErrorWithSuggestion` UX rendering](https://github.com/Azure/azure-dev/pull/6827). azd core now owns extension error rendering. Previously, extensions handled their own error output. With `azdext.Run`, the host takes over: the extension reports structured error data to the host via a gRPC `ReportError` call and stays silent, while the host retrieves it and renders the error. Extension errors with a `Suggestion` message are bridged to `ErrorWithSuggestion` for rich rendering.

### `azd demo gh-url-parse not-a-url` (custom command)

**Source:**

```go
func main() {
	azdext.Run(cmd.NewRootCommand())
}

...

{
    return &cobra.Command{
		...
		RunE: func(cmd *cobra.Command, args []string) error {
			return &azdext.LocalError{
                Message:    "invalid GitHub URL: expected an absolute URL",
                Code:       "invalid_github_url",
                Category:   azdext.LocalErrorCategoryValidation,
                Suggestion: "Use a full URL like https://github.com/Azure/azure-dev",
            }
		},
	}
}
```

**Data extension sends back to host via gRPC `ReportError`:**

```json
{
  "message": "invalid GitHub URL: expected an absolute URL",
  "origin": "ERROR_ORIGIN_LOCAL",
  "suggestion": "Use a full URL like https://github.com/Azure/azure-dev",
  "localError": {
    "code": "invalid_github_url",
    "category": "validation"
  }
}
```

**Trace:**

```json
{
    "Name": "ext.run",
    "Attributes": [
        { "Key": "cmd.entry",          "Value": { "Type": "STRING", "Value": "cmd.demo.gh-url-parse" } },
        { "Key": "error.category",     "Value": { "Type": "STRING", "Value": "validation" } },
        { "Key": "error.code",         "Value": { "Type": "STRING", "Value": "invalid_github_url" } },
        { "Key": "extension.id",       "Value": { "Type": "STRING", "Value": "microsoft.azd.demo" } }
    ],
    "Status": {
        "Code": "Error",
        "Description": "ext.validation.invalid_github_url"
    }
}
```

**Output:**

image

### `azd deploy` (custom service target)

**Source:**

```go
func (p *AgentServiceTargetProvider) Deploy(
	ctx context.Context,
	serviceContext *azdext.ServiceContext,
	progress azdext.ProgressReporter,
) (*azdext.ServiceDeployResult, error) {
	...
    return &azdext.LocalError{
        Message:    "AZURE_AI_PROJECT_ENDPOINT is required: environment variable was not found in the current azd environment",
        Code:       "missing_ai_project_endpoint",
        Category:   azdext.LocalErrorCategoryValidation,
        Suggestion: "run 'azd provision' or connect to an existing project via 'azd ai agent init --project-id <resource-id>'",
    }
}
```

**Trace:**

```json
{
    "Name": "cmd.deploy",
    "Attributes": [
        { "Key": "cmd.entry",          "Value": { "Type": "STRING", "Value": "cmd.deploy" } },
        { "Key": "error.category",     "Value": { "Type": "STRING", "Value": "dependency" } },
        { "Key": "error.code",         "Value": { "Type": "STRING", "Value": "missing_ai_project_endpoint" } }
    ],
    "Status": {
        "Code": "Error",
        "Description": "ext.dependency.missing_ai_project_endpoint"
    }
}
```

**Output:**

image

---------------------------

## Structured error types

Extensions return typed errors to control telemetry categorization and provide user-facing suggestions:

```go
// Local/validation errors
return &azdext.LocalError{
    Message:    "AZURE_SUBSCRIPTION_ID is required",
    Code:       "missing_subscription_id",
    Category:   azdext.LocalErrorCategoryDependency,
    Suggestion: "run 'azd env get-values' to verify environment values",
}
```

`LocalErrorCategory` is a closed set: `validation`, `dependency`, `configuration`, `compatibility`, `permission`, `auth`, `internal`. Unknown values normalize to `local`.

## Error file transport for custom commands

Custom extension commands run as standalone child processes and need a mechanism to flow structured error data back to the host:

1. **Host starts a gRPC server** and passes its address via `AZD_SERVER` and a JWT token via `AZD_ACCESS_TOKEN` env vars to the extension process.
2. **Extension fails** → `azdext.Run` calls `ReportError(ctx, err)`, which wraps the error as a protobuf `ExtensionError` and sends it to the host via the `ExtensionService.ReportError` gRPC call. If the call succeeds, the extension stays silent (the host will render). If it fails, the extension prints the error locally as a fallback.
3. **Host retrieves the error** after the process exits via `extension.GetReportedError()`, which returns the typed Go error (`*LocalError` or `*ServiceError`) stored by the gRPC handler.
4. **Host wraps both errors** — `fmt.Errorf("%w: %w", reportedErr, invokeErr)` — keeping the typed error for telemetry and the `ExtensionRunError` for UX middleware handling.

## `azdext.Run` helper

Previously, every extension had to manually handle color setup, context creation, error reporting, and exit logic:

```go
func main() {
    if v, ok := os.LookupEnv("FORCE_COLOR"); ok && v == "1" {
        color.NoColor = false
    }
    ctx := azdext.NewContext()
    rootCmd := cmd.NewRootCommand()
    if err := rootCmd.ExecuteContext(ctx); err != nil {
        // manual error file reporting...
        color.Red("Error: %v", err)
        os.Exit(1)
    }
}
```

With `azdext.Run`:

```go
func main() {
    azdext.Run(cmd.NewRootCommand())
}
```

`azdext.Run` additionally calls `ReportError(ctx, err)` to send typed errors to the host via gRPC so azd core can report error telemetry.

### DIFF
--- a/cli/azd/internal/tracing/fields/fields.go
+++ b/cli/azd/internal/tracing/fields/fields.go
@@ -319,6 +319,13 @@ const ServiceNameAzd = "azd"
 
 // Error related fields
 var (
+	// Error category that classifies an error.
+	ErrCategory = AttributeKey{
+		Key:            attribute.Key("error.category"),
+		Classification: SystemMetadata,
+		Purpose:        PerformanceAndHealth,
+	}
+
 	// Error code that describes an error.
 	ErrCode = AttributeKey{
 		Key:            attribute.Key("error.code"),
@@ -343,23 +350,6 @@ var (
 	// The frame of the error.
 	ErrFrame = AttributeKey{
 		Key:            attribute.Key("error.frame"),
-		Classification: SystemMetadata,
-		Purpose:        PerformanceAndHealth,
-	}
-)
-
-// Extension error related fields.
-var (
-	// Extension error category.
-	ExtErrorCategory = AttributeKey{
-		Key:            attribute.Key("ext.category"),
-		Classification: SystemMetadata,
-		Purpose:        PerformanceAndHealth,
-	}
-
-	// Extension error code.
-	ExtErrorCode = AttributeKey{
-		Key:            attribute.Key("ext.code"),
 		Classification: SystemMetadata,
 		Purpose:        PerformanceAndHealth,
 	}
--- a/cli/azd/internal/cmd/errors.go
+++ b/cli/azd/internal/cmd/errors.go
@@ -135,8 +135,8 @@ func MapError(err error, span tracing.Span) {
 		code := normalizeCodeSegment(extLocalErr.Code, "failed")
 
 		errDetails = append(errDetails,
-			fields.ExtErrorCategory.String(domain),
-			fields.ExtErrorCode.String(code),
+			fields.ErrCategory.String(domain),
+			fields.ErrCode.String(code),
 		)
 
 		errCode = fmt.Sprintf("ext.%s.%s", domain, code)
--- a/cli/azd/internal/cmd/errors_test.go
+++ b/cli/azd/internal/cmd/errors_test.go
@@ -256,8 +256,8 @@ func Test_MapError(t *testing.T) {
 			},
 			wantErrReason: "ext.validation.invalid_config",
 			wantErrDetails: []attribute.KeyValue{
-				attribute.String("error.ext.category", "validation"),
-				attribute.String("error.ext.code", "invalid_config"),
+				fields.ErrorKey(fields.ErrCategory.Key).String("validation"),
+				fields.ErrorKey(fields.ErrCode.Key).String("invalid_config"),
 			},
 		},
 		{
@@ -269,8 +269,8 @@ func Test_MapError(t *testing.T) {
 			},
 			wantErrReason: "ext.local.something_bad",
 			wantErrDetails: []attribute.KeyValue{
-				attribute.String("error.ext.category", "local"),
-				attribute.String("error.ext.code", "something_bad"),
+				fields.ErrorKey(fields.ErrCategory.Key).String("local"),
+				fields.ErrorKey(fields.ErrCode.Key).String("something_bad"),
 			},
 		},
 		{
@@ -282,8 +282,8 @@ func Test_MapError(t *testing.T) {
 			},
 			wantErrReason: "ext.auth.token_expired",
 			wantErrDetails: []attribute.KeyValue{
-				attribute.String("error.ext.category", "auth"),
-				attribute.String("error.ext.code", "token_expired"),
+				fields.ErrorKey(fields.ErrCategory.Key).String("auth"),
+				fields.ErrorKey(fields.ErrCode.Key).String("token_expired"),
 			},
 		},
 	}
--- a/cli/azd/internal/tracing/fields/key.go
+++ b/cli/azd/internal/tracing/fields/key.go
@@ -45,7 +45,11 @@ func Sha256Hash(val string) string {
 	return hash
 }
 
-// ErrorKey returns a new Key with "error." prefix appended.
+// ErrorKey returns a new Key with "error." prefix prepended.
+// Keys that already have the "error." prefix are returned as-is.
 func ErrorKey(k attribute.Key) attribute.Key {
+	if strings.HasPrefix(string(k), "error.") {
+		return k
+	}
 	return attribute.Key("error." + string(k))
 }
--- a/cli/azd/internal/tracing/fields/key_test.go
+++ b/cli/azd/internal/tracing/fields/key_test.go
@@ -1,0 +1,30 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package fields
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	"go.opentelemetry.io/otel/attribute"
+)
+
+func TestErrorKey(t *testing.T) {
+	tests := []struct {
+		name string
+		key  attribute.Key
+		want attribute.Key
+	}{
+		{"PrependsPrefixToNonErrorKey", "service.name", "error.service.name"},
+		{"SkipsPrefixForErrorKey", "error.code", "error.code"},
+		{"SkipsPrefixForNestedErrorKey", "error.category", "error.category"},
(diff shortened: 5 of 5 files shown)


spboyer (commented):
Two issues found:

**1. AZD_ERROR_FILE lookup reads wrong file when env has duplicates (Medium)**
\
eadReportedExtensionErrorFromEnv\ scans \llEnv\ and stops at the first \AZD_ERROR_FILE=\ match. But \llEnv\ starts with \os.Environ()\ then appends the new temp file path. If the parent environment already has \AZD_ERROR_FILE\ set, the reader will use the stale path instead of the temp file passed to the extension. Consider tracking and reading the explicit path from \createExtensionErrorFileEnv()\ directly instead of parsing first match from the env slice.

**2. Telemetry field keys inconsistent with schema (High)**
New keys are defined as \xt.category\ / \xt.code\ in fields.go, but the test expectations use \rror.ext.category\ / \rror.ext.code\. The \rror.*\ namespace is the established convention in this codebase. Needs alignment -- either update the key definitions or the test assertions.

JeffreyCA:
> Two issues found:
> 
> **1. AZD_ERROR_FILE lookup reads wrong file when env has duplicates (Medium)** eadReportedExtensionErrorFromEnv\ scans \�llEnv\ and stops at the first \AZD_ERROR_FILE=\ match. But \�llEnv\ starts with \os.Environ()\ then appends the new temp file path. If the parent environment already has \AZD_ERROR_FILE\ set, the reader will use the stale path instead of the temp file passed to the extension. Consider tracking and reading the explicit path from \createExtensionErrorFileEnv()\ directly instead of parsing first match from the env slice.
> 
> **2. Telemetry field keys inconsistent with schema (High)** New keys are defined as \�xt.category\ / \�xt.code\ in fields.go, but the test expectations use \�rror.ext.category\ / \�rror.ext.code. The \�rror.*\ namespace is the established convention in this codebase. Needs alignment -- either update the key definitions or the test assertions.

Thanks, fixed the first one but the second isn't an issue as `error` gets prepended and test passes.

cli/azd/docs/extensions/extension-framework.md
```diff
@@ -1124,6 +1125,67 @@ To enable interaction with `azd` from within the extension, the extension must l
 
 The gRPC client must also include an `authorization` parameter with the value from `AZD_ACCESS_TOKEN`; otherwise, requests will fail due to invalid authorization. Extensions must declare their supported capabilities within the extension registry otherwise certain service operations may fail with a permission denied error.
 
+### Structured Error Reporting (Custom Commands)
+
+For custom command execution, extensions can provide richer telemetry by reporting a structured error payload to
+`AZD_ERROR_FILE` before exiting with a non-zero status code.
```
weikanglim: Using a file for reporting back is fine, but I'm wondering if it would be sufficient to have errors reported back via standard gRPC. I like the idea of keeping the protocol simple, and since we're already speaking gRPC, my natural inclination is to look there.

cli/azd/internal/tracing/fields/fields.go
```diff
@@ -348,6 +348,23 @@ var (
...
+		Classification: SystemMetadata,
+		Purpose:        PerformanceAndHealth,
+	}
+
+	// Extension error code.
+	ExtErrorCode = AttributeKey{
+		Key:            attribute.Key("ext.code"),
+		Classification: SystemMetadata,
+		Purpose:        PerformanceAndHealth,
+	}
```
weikanglim: I'm wondering if it's possible to generalize this across extensions and core? The idea of "error category" and "error code" seem generic enough.

weikanglim (commented):
Overall looks good -- just left some comments about gRPC for transport, and if we could have `error.code` and `error.category` standard across ext and core.

cli/azd/internal/tracing/fields/fields.go
JeffreyCA: Good point, there were some existing unused error fields in `fields.go` I was able to reuse

cli/azd/docs/extensions/extension-framework.md
JeffreyCA: I'm fine with that too, updated

PR #6847, opened 2026-02-21
Reject positional args in azd env get-values

### LINKED ISSUE #3354: `azd env get-values` unexpected success
Run `azd env get-values ""`

Expected:
An error that the environment could not be found (in fact, the name is invalid)

Actual:
1. `AZURE_ENV_NAME=""` is returned
2. `.azure/.env` is created as an empty non-referenceable environment and an orphaned file

spboyer:
## Summary

Adds `cobra.NoArgs` validation to the `azd env get-values` command. Previously, passing empty or unexpected positional arguments like `azd env get-values ""` would succeed silently instead of returning an error.

## Changes

One-line fix: added `Args: cobra.NoArgs` to the command definition in `newEnvGetValuesCmd()`.

Follows the existing pattern used by:
- `env get-value`: `cobra.MaximumNArgs(1)`
- `env refresh`: `cobra.MaximumNArgs(1)` with flag/arg conflict checks

Fixes #3354

### DIFF
--- a/cli/azd/cmd/env.go
+++ b/cli/azd/cmd/env.go
@@ -1230,6 +1230,7 @@ func newEnvGetValuesCmd() *cobra.Command {
 	return &cobra.Command{
 		Use:   "get-values",
 		Short: "Get all environment values.",
+		Args:  cobra.NoArgs,
 	}
 }
 


vhvb1989:
There might be a misunderstanding here around this PR and the referenced issue.

I think the issue is about calling `azd env get-values` w/o having an env? @weikanglim ??

I see this in the issue:

> .azure/.env is created as an empty non-referenceable environment and an orphaned file

rajeshkamal5050:
> There might be a misunderstanding here around this PR and the referenced issue.
> 
> I think the issue is about calling `azd env get-values` w/o having an env? @weikanglim ??
> 
> I see this in the issue:
> 
> > .azure/.env is created as an empty non-referenceable environment and an orphaned file

Couldn't repro the original issue. But, this validation should be good even otherwiese.

PR #6893, opened 2026-02-25
allow live mode on demand

vhvb1989:
## Summary

Makes the `AZURE_RECORD_MODE` for CLI tests configurable in the release pipeline, instead of being hardcoded to `playback`.

A new `AzureRecordMode` pipeline parameter is threaded through `release-cli.yml` → `build-and-test.yml` → `build-cli.yml` to control the test record mode.

## Behavior

| Trigger | Mode | Override? |
|---|---|---|
| **Scheduled** (nightly) | `live` | Via parameter |
| **Manual** | `live` (default) | Via `AzureRecordMode` parameter |
| **CI** (push / PR) | `playback` | No (forced) |

## Changes

- **`release-cli.yml`**: Added `AzureRecordMode` parameter (default: `live`). CI triggers (`IndividualCI`, `BatchedCI`, `PullRequest`) are forced to `playback`; all other triggers (scheduled, manual) use the parameter value.
- **`build-and-test.yml`**: Accepts and passes through `AzureRecordMode` to build jobs.
- **`build-cli.yml`**: Uses the parameter for the `AZURE_RECORD_MODE` env var (previously hardcoded to `playback`).

### DIFF
--- a/eng/pipelines/release-cli.yml
+++ b/eng/pipelines/release-cli.yml
@@ -2,7 +2,7 @@ parameters:
   - name: AzureRecordMode
     displayName: Azure Record Mode
     type: string
-    default: playback
+    default: live
     values:
       - playback
       - live
@@ -42,10 +42,10 @@ extends:
     stages:
     - template: /eng/pipelines/templates/stages/build-and-test.yml
       parameters:
-        # Scheduled runs always use 'live' mode for nightly validation against real Azure resources.
-        # Manual and CI runs default to 'playback' but can be overridden via the AzureRecordMode parameter.
-        ${{ if eq(variables['Build.Reason'], 'Schedule') }}:
-          AzureRecordMode: live
+        # Scheduled and manual runs use 'live' mode by default for validation against real Azure resources.
+        # CI runs (push/PR) use 'playback' mode. Manual runs can override via the AzureRecordMode parameter.
+        ${{ if in(variables['Build.Reason'], 'IndividualCI', 'BatchedCI', 'PullRequest') }}:
+          AzureRecordMode: playback
         ${{ else }}:
           AzureRecordMode: ${{ parameters.AzureRecordMode }}
         BuildMatrix:


eng/pipelines/release-cli.yml
```diff
@@ -1,3 +1,12 @@
+parameters:
+  - name: AzureRecordMode
+    displayName: Azure Record Mode
+    type: string
+    default: playback
```
danieljurek: Scheduled runs will always use the default of `playback` in this case.

eng/pipelines/templates/jobs/build-cli.yml
```diff
@@ -148,7 +151,7 @@ jobs:
           AZD_TEST_TENANT_ID: $(AzureSubscriptionTenantId)
           AZD_TEST_AZURE_SUBSCRIPTION_ID: $(SubscriptionId)
           AZD_TEST_AZURE_LOCATION: eastus2
-          AZURE_RECORD_MODE: playback # Always run in 'playback' mode during CI builds
+          AZURE_RECORD_MODE: ${{ parameters.AzureRecordMode }}
```
danieljurek: Perhaps this should be set by template criteria... Something to the effect of: 

```suggestion
          ${{ if eq(variables['Build.Reason'], 'Schedule') }}: 
            AZURE_RECORD_MODE: live
          ${{ else }}: 
            AZURE_RECORD_MODE: ${{ parameters.AzureRecordMode }}
```

This will always set `live` on scheduled runs.

eng/pipelines/release-cli.yml
```diff
@@ -33,6 +42,7 @@ extends:
     stages:
     - template: /eng/pipelines/templates/stages/build-and-test.yml
       parameters:
+        AzureRecordMode: ${{ parameters.AzureRecordMode }}
```
danieljurek: On second thought, put the business logic closer to the top level where the parameters are specified... 

```suggestion
        ${{ if eq(variables['Build.Reason'], 'Schedule') }}:
          AzureRecordMode: live
        ${{ else }}: 
          AzureRecordMode: ${{ parameters.AzureRecordMode }}
```

eng/pipelines/templates/jobs/build-cli.yml
danieljurek: Instead, use the example in https://github.com/Azure/azure-dev/pull/6893/changes#r2855618740 which places the business logic closer the where the parameter is declared. The above suggestion couples `build-cli.yml` to the build context.

eng/pipelines/release-cli.yml
```diff
@@ -33,6 +42,10 @@ extends:
     stages:
     - template: /eng/pipelines/templates/stages/build-and-test.yml
       parameters:
+        ${{ if eq(variables['Build.Reason'], 'Schedule') }}:
+          AzureRecordMode: live
```
wbreza: **Note:** This conditional forces `live` mode for **all** scheduled pipeline runs, regardless of the `AzureRecordMode` parameter value. This means scheduled runs are NOT opt-in for live mode — they always hit real Azure resources.

| Trigger | Mode | Opt-in? |
|---------|------|---------|  
| PR / CI push | `playback` (default) | ✅ |
| Manual run | `playback` (default, selectable) | ✅ |
| **Scheduled** | **`live` (forced)** | **❌** |

If this is intentional (nightly live-test validation), consider adding a comment here explaining the design decision. If live mode should be strictly opt-in, this override should be removed.

eng/pipelines/release-cli.yml
```diff
@@ -33,6 +42,10 @@ extends:
...
+        ${{ else }}:
```
wbreza: Nit: trailing whitespace after `${{ else }}:`. Consider trimming.

eng/pipelines/templates/jobs/build-cli.yml
wbreza: The old comment `# Always run in 'playback' mode during CI builds` was helpful context. Consider adding a replacement comment:
```suggestion
          AZURE_RECORD_MODE: ${{ parameters.AzureRecordMode }} # Defaults to 'playback'; controlled by AzureRecordMode parameter
```

PR #6939, opened 2026-02-28
Silence extension process stderr logs to prevent call-stack-like error output

### LINKED ISSUE #6742: Graceful failure when extensions are not updated
Opportunity to improve extension installation error messages.

**Describe the bug**
When extensions (including test extensions) aren't updated, the tool currently fails with verbose errors that aren't helpful or relevant particularly when the extension isn't related to the current task. The messaging is noisy and doesn't guide the user to resolution. 

It feels like a tidy task that's "in the way" unless it's needed for what I'm trying to accomplish (which in this case with `azd up` the updates aren't needed).

Image

**To Reproduce**
- Attempt to use Azure Developer CLI features when certain extensions are out-of-date (especially test extension).
- Observe the verbose failure output.

**Expected behavior**
Instead of verbose errors, the tool should fail gracefully with clear, plain-language guidance. Error messaging should briefly explain that an extension needs updating and provide instructions/link on how to update.

**Additional context**
Current error output is distracting and not relevant to the user flow. The experience could be improved by minimizing noise and offering actionable steps.

hemarina:
## Fixes #6742 

This pull request adds support for configurable logging in the gRPC message broker used by extension managers. It introduces a new `brokerLogger` field to the event, framework service, and service target managers, allowing detailed broker trace logs to be enabled when the `AZD_EXT_DEBUG` environment variable is set. The changes also update constructors and initialization logic to pass the logger appropriately, and add comprehensive tests to verify the new logging behavior.

**Broker Logging Support**

* Added a `brokerLogger` field to `EventManager`, `FrameworkServiceManager`, and `ServiceTargetManager`, allowing each manager to optionally receive a logger for broker trace output. [[1]](diffhunk://#diff-2df8bf4a2f6b7d296af333ffc559eb42df0f57e0babc7df29d942ad72483eadeR22) [[2]](diffhunk://#diff-ae085a255ad77b3d2f78afd6e80c366d57f9f3f4d30c00fe3099de9475871578R54-R66) [[3]](diffhunk://#diff-17b84deb265cff804bb41e6fbec85cdb13d8b437c4bb02d4945a7d01b31d0b11R69-R81)
* Updated constructors (`NewEventManager`, `NewFrameworkServiceManager`, `NewServiceTargetManager`) to accept an optional `brokerLogger` parameter and store it in the manager. [[1]](diffhunk://#diff-2df8bf4a2f6b7d296af333ffc559eb42df0f57e0babc7df29d942ad72483eadeL41-R48) [[2]](diffhunk://#diff-ae085a255ad77b3d2f78afd6e80c366d57f9f3f4d30c00fe3099de9475871578R54-R66) [[3]](diffhunk://#diff-17b84deb265cff804bb41e6fbec85cdb13d8b437c4bb02d4945a7d01b31d0b11R69-R81)
* Modified broker instantiation in manager methods to pass the logger to `grpcbroker.NewMessageBroker`, enabling trace output when debugging is enabled. [[1]](diffhunk://#diff-2df8bf4a2f6b7d296af333ffc559eb42df0f57e0babc7df29d942ad72483eadeL89-R91) [[2]](diffhunk://#diff-ae085a255ad77b3d2f78afd6e80c366d57f9f3f4d30c00fe3099de9475871578L108-R111) [[3]](diffhunk://#diff-fa52ab512f9d35a8937fe64b507564b19014ee2766b614577b11978417f10b7aL73-R73) [[4]](diffhunk://#diff-deabe5e6d54703ed0509809fe95b2e036fe44144b23a23af67e447db84daf5f9L69-R69) [[5]](diffhunk://#diff-e2a0dd18ec708c0fe162d6f4a726967c9427757d95216e2e47b37de6b9e05d0dL81-R81)

**Extension Host Initialization**

* Refactored `ExtensionHost` initialization to determine broker logging based on `AZD_EXT_DEBUG`, and to pass the logger to all managers via a new `initBrokerLogger` method. [[1]](diffhunk://#diff-fc83135cb84b62a6639403b0ae58b13976028c3beca2f0aa8142ae086767cca1L111-R121) [[2]](diffhunk://#diff-fc83135cb84b62a6639403b0ae58b13976028c3beca2f0aa8142ae086767cca1L159-R175)
* Added logic to parse `AZD_EXT_DEBUG` and configure broker logging to `os.Stderr` when enabled, otherwise leaving logging silent.

**Testing and Validation**

* Updated all relevant unit tests to use the new constructors with the logger parameter. [[1]](diffhunk://#diff-4c6d44356769107ce89ab71b171e61204609d8caec33ec58b6250409ca573b2dL124-R124) [[2]](diffhunk://#diff-4c6d44356769107ce89ab71b171e61204609d8caec33ec58b6250409ca573b2dL139-R139) [[3]](diffhunk://#diff-4c6d44356769107ce89ab71b171e61204609d8caec33ec58b6250409ca573b2dL179-R179) [[4]](diffhunk://#diff-4c6d44356769107ce89ab71b171e61204609d8caec33ec58b6250409ca573b2dL211-R211) [[5]](diffhunk://#diff-4c6d44356769107ce89ab71b171e61204609d8caec33ec58b6250409ca573b2dL232-R232) [[6]](diffhunk://#diff-4c6d44356769107ce89ab71b171e61204609d8caec33ec58b6250409ca573b2dL280-R280) [[7]](diffhunk://#diff-4c6d44356769107ce89ab71b171e61204609d8caec33ec58b6250409ca573b2dL319-R319) [[8]](diffhunk://#diff-4c6d44356769107ce89ab71b171e61204609d8caec33ec58b6250409ca573b2dL354-R354) [[9]](diffhunk://#diff-4c6d44356769107ce89ab71b171e61204609d8caec33ec58b6250409ca573b2dL375-R375) [[10]](diffhunk://#diff-4c6d44356769107ce89ab71b171e61204609d8caec33ec58b6250409ca573b2dL396-R396) [[11]](diffhunk://#diff-4c6d44356769107ce89ab71b171e61204609d8caec33ec58b6250409ca573b2dL417-R417) [[12]](diffhunk://#diff-d286e0340b24de088345aaebda33b099e806032222b49deb676ff1e5b8d7a1f0L90-R90)
* Added a new test to `extension_host_test.go` that verifies broker logging is enabled or disabled depending on the value of `AZD_EXT_DEBUG`.

These changes collectively allow developers to enable verbose broker logging for debugging purposes, while keeping normal operation silent, and ensure the feature is well-tested.


vhvb1989 (approved):

cli/azd/pkg/azdext/extension_host.go
```diff
@@ -162,6 +165,21 @@ func (er *ExtensionHost) Run(ctx context.Context) error {
...
+	// "1", "t", "TRUE", "true", etc.).
+	originalLogWriter := log.Default().Writer()
+	restoreLogOutput := false
+	if isDebug, err := strconv.ParseBool(os.Getenv("AZD_EXT_DEBUG")); err != nil || !isDebug {
+		log.SetOutput(io.Discard)
+		restoreLogOutput = true
+	}
+	if restoreLogOutput {
+		defer log.SetOutput(originalLogWriter)
+	}
```
wbreza: Need to review some extensions and confirm this doesn't have any undesired side effects.

cli/azd/pkg/azdext/extension_host.go
```diff
@@ -149,6 +152,17 @@ func (er *ExtensionHost) Run(ctx context.Context) error {
...
+	// Silence the global logger in extension processes to prevent internal
+	// gRPC broker trace logs from appearing in stderr. Extensions compiled
+	// against older SDK versions still use log.Printf directly, so this
+	// ensures backward compatibility. When AZD_EXT_DEBUG is truthy, keep
+	// logging to stderr for diagnostics.
+	// Uses strconv.ParseBool to match WaitForDebugger semantics (accepts
+	// "1", "t", "TRUE", "true", etc.).
+	if isDebug, err := strconv.ParseBool(os.Getenv("AZD_EXT_DEBUG")); err != nil || !isDebug {
+		log.SetOutput(io.Discard)
+	}
```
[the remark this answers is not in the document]
wbreza: I would agree with this comment - please implement.

cli/azd/pkg/azdext/extension_host.go
wbreza, line 178: **[High]** Collateral silencing of legitimate error/diagnostic logging

This silences the **entire** global logger during `Run()`, not just the grpcbroker trace logs. This also silences:
- `event_manager.go:236,285` — error logging when event handler invocations fail (`"invokeProjectHandler error for event %s: %v"`)
- `extension_host.go:331` — the shutdown message
- Any `log.Printf` in extension-authored service target providers or event handler callbacks

The event_manager error logs are **not** broker trace spam — they are legitimate error diagnostics.

**Alternative approach:** Give the `MessageBroker` its own private `*log.Logger` that can be independently silenced, so only broker trace logs are suppressed while legitimate error logging continues.

**If keeping this approach**, at minimum document the collateral silencing in the `Run()` godoc:
```suggestion
}
if restoreLogOutput {
	defer log.SetOutput(originalLogWriter)
}

// NOTE: The global logger is silenced above. Extensions that need logging
// during Run() should use fmt.Fprintf(os.Stderr, ...) or a dedicated
// *log.Logger instance. Set AZD_EXT_DEBUG=true to preserve default logger.
```

cli/azd/pkg/azdext/extension_host.go
```diff
@@ -162,6 +165,21 @@ func (er *ExtensionHost) Run(ctx context.Context) error {
...
 	_ = WaitForDebugger(ctx, er.client)
 
+	// Silence the global logger in extension processes to prevent internal
+	// gRPC broker trace logs from appearing in stderr. When AZD_EXT_DEBUG
+	// is truthy, keep logging to stderr for diagnostics.
+	// Uses strconv.ParseBool to match WaitForDebugger semantics (accepts
+	// "1", "t", "TRUE", "true", etc.).
+	originalLogWriter := log.Default().Writer()
+	restoreLogOutput := false
+	if isDebug, err := strconv.ParseBool(os.Getenv("AZD_EXT_DEBUG")); err != nil || !isDebug {
```
wbreza: **[Low]** Redundant `restoreLogOutput` flag

The `restoreLogOutput` flag and second `if` block can be simplified — if we didn't set output to discard, restoring it to the same value is harmless.

```suggestion
originalLogWriter := log.Default().Writer()
if isDebug, err := strconv.ParseBool(os.Getenv("AZD_EXT_DEBUG")); err != nil || !isDebug {
```

cli/azd/pkg/azdext/extension_host_test.go
```diff
@@ -688,3 +692,127 @@ func TestExtensionHost_MultipleRegistrationErrors(t *testing.T) {
...
+		runner := NewExtensionHost(client)
+
+		err := runner.Run(context.Background())
+		require.NoError(t, err)
+
+		// After Run(), logger should be restored
+		assert.NotEqual(t, io.Discard, log.Writer(),
+			"log output should be restored after Run() returns")
+	})
+}
```
wbreza, line 746: **[High]** Missing test for `AZD_EXT_DEBUG=true` (logging preserved)

All four subtests verify that logging is **silenced**. There is no test verifying that logging is **preserved** when `AZD_EXT_DEBUG=true`. This is the critical "debug escape hatch" called out in the PR description.

Suggested addition:
```go
t.Run("preserves log output when AZD_EXT_DEBUG is true", func(t *testing.T) {
	var buf bytes.Buffer
	log.SetOutput(&buf)
	os.Setenv("AZD_EXT_DEBUG", "true")

	client := newTestAzdClient()
	runner := NewExtensionHost(client)

	err := runner.Run(context.Background())
	require.NoError(t, err)

	log.Printf("debug-mode message")
	assert.Contains(t, buf.String(), "debug-mode message",
		"log output should be preserved when AZD_EXT_DEBUG=true")
})
```

cli/azd/pkg/azdext/extension_host_test.go
```diff
@@ -225,7 +229,7 @@ func newTestAzdClient() *AzdClient {
 }
 
 func TestExtensionHost_ServiceTargetOnly(t *testing.T) {
-	t.Parallel()
+	// Not parallel: Run() mutates global log output which races with other tests.
```
wbreza: **[Medium]** Test serialization impact

Removing `t.Parallel()` from 7 existing tests forces serial execution due to global logger mutation. This is a necessary consequence of the current approach, but if a scoped-logger approach were adopted (giving `MessageBroker` its own `*log.Logger` instead of using the global one), `t.Parallel()` could be restored.

wbreza (approved):
## Code Review Summary

| Priority | Count |
|----------|-------|
| Critical | 0 |
| High | 2 |
| Medium | 2 |
| Low | 1 |
| **Total** | **5** |

### Safety Assessment

- ✅ **Safe**: Correctly solves the stated problem (grpcbroker trace logs in stderr)
- ✅ **Safe**: Defer-based restoration is robust
- ✅ **Safe**: `fmt.Fprintf(os.Stderr, ...)` paths (used in `run.go:printError`) are unaffected
- ⚠️ **Risk**: Silences legitimate error diagnostics from `event_manager.go` during handler failures
- ⚠️ **Risk**: Behavioral change to SDK public API without documentation
- ❌ **Gap**: No test for the `AZD_EXT_DEBUG=true` preservation path

### What Looks Good
- The problem is well-understood and clearly articulated in the PR description
- The defer-based restoration is correct and robust (survives panics)
- The `AZD_EXT_DEBUG` integration uses `strconv.ParseBool` consistently with `WaitForDebugger`
- Test coverage for the silencing behavior is thorough (4 subtests covering unset, empty, false, and invalid)

**Overall Assessment:** Approve with suggestions

cli/azd/pkg/azdext/extension_host_test.go
hemarina: @copilot Read above comment. We are going to use scoped-logger approach. Restore t.Parallel()

cli/azd/pkg/grpcbroker/message_broker.go
```diff
@@ -91,12 +92,19 @@ type MessageBroker[TMessage any] struct {
 // or a server stream (grpc.BidiStreamingServer) as both implement the BidiStream interface.
 // The ops parameter provides stateless operations for message manipulation.
 // The name parameter is used for logging identification.
+// The logger parameter sets the broker's private logger; pass nil for silent (io.Discard).
 func NewMessageBroker[TMessage any](
 	stream BidiStream[TMessage],
 	ops MessageEnvelope[TMessage],
 	name string,
```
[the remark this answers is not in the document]
hemarina: IIRC, ext dev don't call grpcbroker package external.

wbreza (approved):
## Re-Review - PR #6939 (3 new commits since prior approval)

### What Changed

The original approach silenced the **global** `log.Default()` logger during `Run()`, which required `defer` cleanup and prevented tests from running in parallel. The new approach uses a **private per-broker** `*log.Logger`:

1. `NewMessageBroker` now accepts an optional `*log.Logger` (nil = silent via `io.Discard`)
2. All `log.Printf` calls in `message_broker.go` replaced with `mb.logger.Printf` - no global state mutation
3. Tests that were "Not parallel" can now run with `t.Parallel()` (7 tests)
4. `TestExtensionHost_RunSilencesLog` (126 lines, global state manipulation) replaced with `TestExtensionHost_BrokerLogging` (55 lines, table-driven, no global state)

### Assessment

This is a cleaner, safer approach than what I originally approved. No new issues found. The change from global to per-instance logging eliminates the data race concern entirely. Nice improvement!

| Priority | Count |
|----------|-------|
| Critical | 0 |
| High | 0 |
| Medium | 0 |
| Low | 0 |
| **Total** | **0** |

cli/azd/internal/grpcserver/event_service.go
```diff
@@ -78,7 +78,7 @@ func (s *eventService) EventStream(stream grpc.BidiStreamingServer[azdext.EventM
 
 	// Create message broker with EventMessageEnvelope
 	envelope := azdext.NewEventMessageEnvelope()
-	broker := grpcbroker.NewMessageBroker(stream, envelope, extension.Id)
+	broker := grpcbroker.NewMessageBroker(stream, envelope, extension.Id, nil)
```
wbreza: **[High]** Server-side broker silences trace logs even with `--debug`

Passing `nil` here creates an `io.Discard` logger, which means broker communication traces (handler registration, request/response flow, errors) are permanently silenced on the azd core CLI side - even when the user passes `--debug`.

Before this PR, these traces used `log.Printf` which respected the global logger configured by `setupLogging()` in `main.go`. That function routes output to stderr when `--debug` is set and to `io.Discard` when it is not. This change breaks that contract.

The same issue applies to `framework_service.go` and `service_target_service.go`.

**Suggested fix:** Pass `log.Default()` so broker traces continue to respect the `--debug` flag:
```go
broker := grpcbroker.NewMessageBroker(stream, envelope, extension.Id, log.Default())
```

This preserves the existing behavior exactly: `--debug` on -> traces visible; `--debug` off -> traces silenced (because `setupLogging` sets the global logger to `io.Discard`).

cli/azd/pkg/azdext/extension_host.go
```diff
@@ -156,12 +157,22 @@ func (er *ExtensionHost) WithServiceEventHandler(
...
 	// Wait for debugger if AZD_EXT_DEBUG is set
 	// When user declines or cancels, continue so extension doesn't exit while azd continues
 	_ = WaitForDebugger(ctx, er.client)
 
+	// Broker trace loggers are silent by default (io.Discard in NewMessageBroker).
+	// When AZD_EXT_DEBUG is truthy, enable verbose broker logging to stderr
+	// for diagnostics while keeping the global logger untouched.
+	// Uses strconv.ParseBool to match WaitForDebugger semantics (accepts
+	// "1", "t", "TRUE", "true", etc.).
+	var brokerLogger *log.Logger
```
wbreza, line 173: **[High]** Extension host defaults to silent broker logging - extension developers lose diagnostic traces

When extension developers use `azdext.Run()`, broker logging is only enabled if they explicitly set `AZD_EXT_DEBUG=true`. This is fine as the default for production, but the azd core already propagates `--debug` to extensions as the `AZD_EXT_DEBUG` env var (see `cmd/middleware/extensions.go` lines 148-158). However, the extension host's `initManagers` creates the broker logger based solely on `AZD_EXT_DEBUG` *at the point where `Run()` is called*.

Consider whether the extension host should also check if `--debug` was passed to the listen command, so that extension developers get consistent debug logging behavior without needing to understand two different debug mechanisms.

This is not a blocker, but worth considering for consistency between the core CLI and extension debug experiences.

cli/azd/pkg/grpcbroker/message_broker.go
```diff
@@ -91,12 +92,19 @@ type MessageBroker[TMessage any] struct {
...
+	logger *log.Logger,
```
wbreza, line 102: **[Medium]** `NewMessageBroker` API design note - `nil` default may encourage silent-by-default usage

The nil-to-Discard fallback is a clean API, but it makes it very easy for callers to pass `nil` and silently lose diagnostic output. Every call site in `internal/grpcserver/` currently passes `nil`, which breaks `--debug` logging on the server side.

Consider adding a doc comment that explicitly guides callers:
```go
// The logger parameter sets the broker's private logger.
// Pass log.Default() on the server side to inherit --debug semantics.
// Pass nil for silent operation (e.g., in extensions where AZD_EXT_DEBUG controls logging).
```

This helps future maintainers understand that `nil` is intentional for extensions but incorrect for the core CLI server.

wbreza (commented):
## Re-Review - PR #6939 (logging regression findings)

### Summary

The move from global `log.Printf` to a private per-broker `*log.Logger` is architecturally sound - it eliminates the data race on the global logger and enables parallel tests. However, the current initialization passes `nil` (silent) in places where the broker should inherit the existing debug logging semantics. The goal should not be to always discard broker logging - it should be to give each context the right logger for its environment.

### The Problem

**Before this PR:** Broker trace logs use `log.Printf` (global logger). In `main.go`, `setupLogging()` configures the global logger based on `--debug`:
- `--debug` passed -> `log.SetOutput(os.Stderr)` -> broker traces visible
- No `--debug` -> `log.SetOutput(io.Discard)` -> broker traces silenced

**After this PR:** Server-side brokers in `internal/grpcserver/` pass `nil` to `NewMessageBroker`, which creates a private `io.Discard` logger. Broker traces are now **always silenced on the server side, even with `--debug`**. This is a regression - developers who pass `--debug` expect to see all diagnostic output, including broker communication traces.

### Findings

| Priority | Count |
|----------|-------|
| Critical | 0 |
| High | 2 |
| Medium | 1 |
| Low | 0 |
| **Total** | **3** |

### Design Principle

The components correctly accept a `*log.Logger` parameter - that is good design. But proper initialization matters:

| Context | Expected Logger | Rationale |
|---------|----------------|-----------|
| **Core azd CLI (server-side)** | `log.Default()` | Inherits `--debug` semantics from `setupLogging()` |
| **Extension host (client-side)** | `nil` by default, stderr when `AZD_EXT_DEBUG=true` | Extensions have their own debug mechanism |
| **Tests** | `nil` | Silent unless explicitly testing logging |

The key insight: **the default should be consistent with the execution context, not universally silent.**

wbreza (approved):
## Re-Review: All Previous Findings Addressed ✅

| Previous Finding | Status |
|---|---|
| **[High] Server-side broker silences trace logs with `--debug`** | ✅ Fixed — `nil` → `log.Default()` in all 3 gRPC services |
| **[High] Extension host loses diagnostic traces** | ✅ Fixed — added `AZD_DEBUG` check so `--debug` enables broker logging |
| **[Medium] `NewMessageBroker` nil default API note** | ✅ Addressed — doc comment clarifies server vs extension usage |

The debug logging behavior is now correct:
- **Server-side**: `log.Default()` respects `--debug` via `setupLogging()` in main.go
- **Extension-side**: `AZD_EXT_DEBUG` or `AZD_DEBUG` enables broker traces to stderr

LGTM — clean implementation.

PR #6910, opened 2026-02-26
[design] azd update command, auto-update, and channel management

rajeshkamal5050:
Design doc for azd update command, auto-update, and channel management.

Covers:
- `azd update` command with channel switching
- Auto-update with two-phase stage/apply flow
- Staged binary verification (codesign check to catch truncated downloads)
- Install method-aware update strategies (brew/winget/choco delegation, direct binary for scripts)
- Channel switch confirmation prompt (bidirectional, not just downgrade)
- CI/CD detection to skip auto-updates
- Update banner suppression during `azd update` and `azd config`
- Duplicate WARNING prevention (elevation warning suppresses out-of-date banner)
- `azd version` channel suffix derived from binary version string
- Telemetry with 13 result codes integrated into MapError pipeline
- Alpha feature toggle (`alpha.update`, default off) for safe rollout

Epic: #6721

Issue #6721, opened 2026-02-04
[EPIC] Define and implement CLI update, auto‑update, and channel management for azd


kristenwomack:
### Overview
Today when typing an `azd` command and there is a new version available, the customer gets a message to copy/paste and run the script to install the lastest azd version. Often customer try `azd upgrade` to get the latest version, but we don't have that feature (yet). The terminal output with instructions is helpful but could be easier and smoother. 

This Epic focuses on delivering a clearer and more intuitive update experience for azd—starting simple and iterating through a crawl / walk / run approach getting to auto updates (that don't break pipeline automation).

### The goal is to:

- Make it easy for users to update `azd` intentionally (`azd update`)
- Experiment with opt‑in auto‑update safely (starting with Insiders)
- Preserve user control (opt‑out, pin versions, stable vs Insiders)
- Avoid noisy or disruptive behavior
- Respect platform realities (Windows MSI, macOS Homebrew, scripts)

We will start with Insiders builds to test edge cases, learn from real usage, and incrementally evolve toward a default‑safe, low‑friction update model.

rajeshkamal5050:
Marking this as a duplicate of this new epic https://github.com/Azure/azure-dev/issues/7759

PR #6942, opened 2026-02-28
[Feature] azd update command, auto-update, and channel management

### LINKED ISSUE #6675: Investigate automatic updates for azd
## Problem
GitHub Copilot CLI update itself automatically for customers, ensuring users are always on the latest version without manual intervention. Could a similar auto-update strategy work for the Azure Developer CLI (`azd`)? The key challenge is ensuring that automatic updates would not negatively impact CI/CD pipelines, automation, or environments where stability is critical.

## Investigation Questions
- How does [GitHub Copilot CLI](https://github.com/github/copilot-cli) implement its automatic update mechanism?
- Are there technical, security, or policy considerations that make this viable or risky for `azd`?
- What would be needed to enable safe, pipeline-friendly auto-updates for `azd`?
- Should there be environment-detection or flags to disable auto-updates in automation scenarios?

## Acceptance Criteria
- Document how auto-update works in GitHub Copilot CLI, including update channels, triggers, and user opt-out mechanisms, if any.
- Assess the feasibility and implications for implementing auto-update in `azd`, especially regarding CI/CD blocking and enterprise usage.
- Recommendations on whether and how to proceed, including default behaviors (on/off), environment detection, and documentation needs.

## Additional Context
- Many CLI tools avoid auto-updating by default to prevent unintended breaking changes during automation.
- If feasible, this could improve user experience but must be implemented with care.

### LINKED ISSUE #6722: Define azd update strategy
Description

Define the strategy for updating `azd version`, including manual updates, opt‑in auto‑updates, and channel management (Stable / Insiders). This issue captures goals, non‑goals, constraints, and the phased crawl / walk / run approach agreed in discussion. It serves as the framing document for all follow‑on work in this epic.

### In scope

- Document goals (safe, low‑friction, opt‑in, non‑disruptive)
- Document non‑goals (e.g., solving all edge cases up front)
- Define crawl / walk / run sequencing
- Capture explicit constraints (platform installs, failure tolerance)
- Record key open questions and deferred decisions

### Out of scope

- Implementation details
- Platform‑specific code changes

### Acceptance criteria

- [ ] Strategy document exists and is linked from the Epic
- [ ] Crawl / walk / run phases are explicitly defined
- [ ] Known constraints and risks are documented
- [ ] Open questions are captured (not silently resolved)

### LINKED ISSUE #6723: Introduce an `azd update` command for manual upgrades
### Description

Add an `azd update` command that provides a clean, user‑friendly way to manually upgrade azd. The command should wrap existing install/upgrade mechanisms and improve the current copy‑paste script experience, while preserving safe fallback behavior if the upgrade fails.

### In scope

- New azd update command
- Clear messaging when an update is available
- Execution of the same scripts users run manually today
- Failure messaging that preserves existing manual instructions

### Out of scope

- Auto‑update
- Channel switching logic beyond messaging

### Acceptance criteria

- [ ] Users can run `azd update` instead of copy/pasting scripts
- [ ] Failure does not leave `azd` in a broken state
- [ ] Clear fallback instructions are shown on failure
- [ ] No auto‑update behavior is introduced by this command

### LINKED ISSUE #6724: Define Stable vs Insiders channel model and persistence
### Description 

Define how `azd update` channels work, starting with Stable and Insiders. Users should remain on their chosen channel unless they explicitly change it. This issue establishes the mental model and rules for channel behavior before any auto‑update logic is introduced.

### In scope

- [dependency] Stable vs Insiders definitions
- Channel persistence rules
- Messaging when switching channels
- Decision framing: single binary with channels vs alternatives (documented, not implemented)

### Out of scope

Auto‑update mechanics
Multiple binaries running side‑by‑side

### Acceptance criteria

- [ ] Channel behavior is documented and agreed
- [ ] Users do not switch channels implicitly
- [ ] Insiders channel is clearly framed as experimental
- [ ] Open design questions are explicitly recorded

### LINKED ISSUE #6726: Handle platform‑specific upgrade constraints and add telemetry
### Description

Document and handle platform‑specific upgrade constraints (Windows MSI, macOS Homebrew, per‑user vs machine installs) and add telemetry to understand how users interact with update channels and auto‑update features.

### In scope

- Document known platform risks and unsupported scenarios
- Define expected behavior per platform (even if “not supported”)

### Add telemetry for:

- [ ] Channel selection
- [ ] Auto‑update opt‑in
- [ ] Upgrade success / failure signals

### Out of scope

- Solving all platform edge cases
- Automatic remediation of unsupported installs

### Acceptance criteria

- [ ] Platform constraints are explicitly documented
- [ ] Unsupported scenarios are clearly called out
- [ ] Telemetry is added and validated
- [ ] Data can inform future defaults and sequencing

rajeshkamal5050:
## Feature

Adds a hidden and feature-toggled `azd update` command for self-updating azd — background
update checks, staged binary downloads with verification, and channel switching (stable ↔ daily).

**Design Doc:** https://github.com/Azure/azure-dev/pull/6910
**Epic:** https://github.com/Azure/azure-dev/issues/6721

Closes #6675
Closes #6722
Closes #6723
Closes #6724
Closes #6726

### Follow-up enhancements (tracked under epic #6721)

- #6982 — Rollback/recovery on failed update
- #6983 — Crash-loop guard for auto-update
- #6984 — Version pinning support
- #6985 — Checksum verification and TOCTOU mitigation
- #6986 — Concurrent update locking
- #6987 — Leverage existing install scripts for download and verification

## Changes

- **`azd update` command** (`cmd/update.go`) — Hidden, auto-enables alpha feature on first run. Downloads, verifies, and applies updates. `--channel` flag for switching between stable/daily with confirmation prompt.
- **Update manager** (`pkg/update/manager.go`) — Version check via GitHub releases API, download, staged binary verification (codesign on macOS, min size check), and safe file replacement. Injected `*http.Client` for testability.
- **Update config** (`pkg/update/config.go`) — `updates.channel` and `updates.autoUpdate` settings, gated behind `update` alpha feature flag.
- **Auto-update** (`main.go`) — Background check on every invocation with 60s context timeout, stages binaries, auto-applies or shows elevation warning on next run.
- **Package manager detection** — Delegates to `brew upgrade` (macOS), `winget upgrade` / `choco upgrade` (Windows). Script installs use direct binary replacement.
- **Windows MSI support** — Non-package-manager Windows installs download and run the MSI via `msiexec /i`. msiexec is spawned detached (`CREATE_NEW_PROCESS_GROUP | DETACHED_PROCESS`) so azd can exit cleanly and release the binary lock before msiexec replaces it.
- **Version display** (`cmd/version.go`) — Shows `(daily)` or `(stable)` channel suffix, channel-aware "out of date" warnings.
- **Telemetry** (`internal/tracing/fields/fields.go`) — Update outcome, channel, method, error codes.
- **CI pipeline** (`eng/pipelines/`) — Publishes `version.txt` alongside daily builds.
- **Unit tests** (`pkg/update/*_test.go`) — Covers config, update checking, download, verification, staging, and apply flows.

## E2E Testing

**30 test cases** across macOS, Linux, and Windows — all passing.

| Test | macOS | Linux | Windows | Notes |
|------|-------|-------|---------|-------|
| A1 Stable update | ✅ | ✅ | ✅ | .zip (mac), .tar.gz (linux), .msi (win) |
| A2 Already up to date | ✅ | ✅ | ✅ | |
| A3 Daily update | ✅ | ✅ | ✅ | |
| A4 Stable→daily | ✅ | ✅ | ✅ | `--no-prompt` used in Linux/Windows CI |
| A5 Daily→stable | ✅ | ✅ | ✅ | `--no-prompt` used in Linux/Windows CI |
| A6 No network | ⏭ | ⏭ | ☐ | Requires manual network disconnect |
| A7 Invalid channel | ✅ | ✅ | ✅ | |
| A8 Config-only | ✅ | ✅ | ✅ | |
| A9 Help text | ✅ | ✅ | ✅ | |
| A10 CI detection | ✅ | ✅ | ✅ | |
| A11 Toggle off | ✅ | ✅ | ✅ | |
| B1 Brew update | ✅ | — | — | macOS only |
| B2 Brew daily blocked | ✅ | — | — | macOS only |
| B3 Config-only (brew) | ✅ | — | — | macOS only |
| C1 Auto-stage + apply | ✅ | ✅ | N/A | Win: staging skipped by design, uses MSI |
| C2 Elevation warning | ✅ | ⏭ | N/A | Linux/Win: CI runs as root/admin |
| C3 Truncated binary | ✅ | ✅ | N/A | Win: staging is mac/linux only |
| C4 Auto-update off | ✅ | ✅ | ✅ | |
| D1 Combined flags | ✅ | ✅ | ✅ | |
| D2 Code signature | ✅ | — | ✅ | Mac: codesign. Win: Authenticode (Valid, Microsoft Corp) |
| W1 Winget delegation | — | — | ⚠️ | Delegation works; winget needs interactive terms in CI |
| W2 Script MSI update | — | — | ✅ | MSI download + `msiexec /i` |
| W3 Winget daily blocked | — | — | ✅ | |
| W4 Choco delegation | — | — | ✅ | `choco upgrade` successful |
| W5 Choco daily blocked | — | — | ✅ | |
| W6 MSI in-place update | — | — | ✅ | Detached msiexec replaces binary in ~1.7s |

### Platform Notes

- **macOS**: Tested locally. Downloads `.zip`, verifies Apple codesign.
- **Linux**: Docker `golang:1.25` (linux/amd64). Downloads `.tar.gz`. ETXTBSY fix (remove-then-create in `copyFile`) enables C1 auto-apply.
- **Windows**: GitHub Actions `windows-latest` (Server 2025), [full suite run #22610574260](https://github.com/rajeshkamal5050/azure-dev/actions/runs/22610574260) — 38/38 steps pass. Uses MSI installer. W6 validates detached msiexec — binary replaced in ~1.7s while azd exits cleanly (exit code 0).

### Known Limitations

1. **Stale daily version.txt** — Remote `version.txt` doesn't match actual binary version. Pipeline fix needed (not an azd code issue).
2. **Auto-update banner** — Won't show when updating to production binaries that lack this code. Works once feature ships.
3. **Windows auto-staging** — `StageUpdate()` skips on Windows by design. Windows uses MSI via `msiexec`, not binary replacement.
4. **Winget interactive terms** — `winget upgrade` requires terms acceptance, which can't be automated in CI.
5. **Elevation warning (C2)** — Requires non-admin context. CI environments (admin/root) can't trigger it; success path covered by A1/W2.

**Full E2E test details with logs:** https://gist.github.com/rajeshkamal5050/d509c03e30a7249a6a4a931ab761a286

### DIFF
--- a/eng/pipelines/templates/stages/publish.yml
+++ b/eng/pipelines/templates/stages/publish.yml
@@ -372,29 +372,6 @@ stages:
                     CreateGitHubRelease: false
                     PublishUploadLocations: release/daily;daily/archive/$(Build.BuildId)-$(Build.SourceVersion)
 
-                - pwsh: |
-                    New-Item -ItemType Directory -Path daily-version -Force
-                    Set-Content -Path ./daily-version/version.txt -Value "$(CLI_VERSION)" -NoNewline
-                    Write-Host "Daily version.txt:"
-                    Get-Content ./daily-version/version.txt
-                  displayName: Write daily version.txt
-
-                - task: AzurePowerShell@5
-                  displayName: Upload daily version.txt to storage
-                  inputs:
-                    azureSubscription: 'Azure SDK Artifacts'
-                    azurePowerShellVersion: LatestVersion
-                    pwsh: true
-                    ScriptType: InlineScript
-                    Inline: |
-                      azcopy copy "daily-version/version.txt" "$(publish-storage-location)/`$web/azd/standalone/release/daily/" --overwrite=true
-                      if ($LASTEXITCODE) {
-                        Write-Error "Upload failed"
-                        exit 1
-                      }
-                  env:
-                    AZCOPY_AUTO_LOGIN_TYPE: 'PSCRED'
-
       - deployment: Publish_For_PR
         environment: none
         condition: >-


cli/azd/cmd/update.go
```diff
@@ -0,0 +1,327 @@
...
+}
+
+func (a *updateAction) Run(ctx context.Context) (*actions.ActionResult, error) {
+	// Gate behind alpha feature toggle
+	if !a.alphaFeatureManager.IsEnabled(update.FeatureUpdate) {
+		return nil, fmt.Errorf(
+			"azd update is an alpha feature and must be enabled to use. Run: %s",
+			alpha.GetEnableCommand(update.FeatureUpdate),
+		)
+	}
```
vhvb1989, line 129: @rajeshkamal5050 , Can we follow the same approach we did for the agentic-init and just displayed a heads-up about the feature is in alpha mode and `it has been turned on` when folks runs this?  

I like how we can avoid the error and the additional instruction to manually enable the feature.

rajeshkamal5050: Done

vhvb1989 (approved):

cli/azd/cmd/update.go
```diff
@@ -0,0 +1,342 @@
...
+			return nil, fmt.Errorf("failed to enable update feature: %w", err)
+		}
+
+		if err := a.configManager.Save(userCfg); err != nil {
+			return nil, fmt.Errorf("failed to save config: %w", err)
+		}
+
+		a.console.MessageUxItem(ctx, &ux.MessageTitle{
+			Title: fmt.Sprintf("azd update is in alpha. "+
+				"To turn off in the future, run `azd config unset alpha.%s`.\n",
```
hemarina: Running `azd config unset alpha.update` won't turn the feature off as the feature is auto enabled by running azd update if not already enabled. Could we add check if alpha.update is set to off and disable update feature?

rajeshkamal5050: Valid point Marina. 

But, the rollout plan is alpha (internal testing) → beta (public). The alpha toggle goes away entirely and users won't need to toggle alpha.update on/off. The auto-enable I believe is just a convenience for the internal/alpha phase so internal testers do not need a manual config step @kristenwomack @danieljurek @vhvb1989

kristenwomack (approved):

cli/azd/cmd/update.go
hemarina: I see. Then shall we remove "To turn off in the future, run `azd config unset alpha.%s`"?

cli/azd/pkg/update/manager.go
```diff
@@ -0,0 +1,1003 @@
...
+}
+
+// Manager handles checking for and applying azd updates.
+type Manager struct {
+	commandRunner exec.CommandRunner
+}
+
+// NewManager creates a new update Manager.
+func NewManager(commandRunner exec.CommandRunner) *Manager {
+	return &Manager{
```
wbreza, line 57: **[🟠 High] Code signature verification silently succeeds on error**

Both `verifyCodesignMac` (line 441) and `verifyAuthenticode` (line 467) catch command execution errors and return `nil` (success):

```go
if err != nil {
    log.Printf("codesign verification failed: %v, skipping")
    return nil  // ← treats failure as success
}
```

This makes the security check purely advisory. If codesign/PowerShell is unavailable or returns an unexpected error, verification is silently skipped. Consider making this a hard failure for non-dev builds, or at minimum log at warning level and document this as a known gap.

cli/azd/pkg/update/manager.go
```diff
@@ -0,0 +1,1003 @@
...
+	var err error
+
+	switch cfg.Channel {
+	case ChannelStable:
+		info, err = m.checkStableVersion(ctx)
+	case ChannelDaily:
+		info, err = m.checkDailyVersion(ctx)
+	default:
+		return nil, fmt.Errorf("unsupported channel: %s", cfg.Channel)
+	}
```
wbreza, line 86: **[🟠 High] `http.DefaultClient` with no timeout**

All HTTP operations (`checkStableVersion`, `checkDailyVersion`, `downloadFile`) use `http.DefaultClient` which has **no timeout**. A stalled version check or download could hang indefinitely.

The background `fetchLatestVersion` goroutine in `main.go` sends results via a channel that `main()` receives with `<-latest` — a hung HTTP call blocks program exit.

**Suggested Fix:**
```go
client := &http.Client{Timeout: 30 * time.Second}  // for version checks
// or inject via Manager struct field
```

cli/azd/pkg/update/manager.go
```diff
@@ -0,0 +1,1003 @@
...
+
+	fmt.Fprintf(writer, "Downloading azd from %s...\n", downloadURL)
+
+	// Download to a temp file
+	tempDir := os.TempDir()
+	archiveName := fmt.Sprintf("azd-%s-%s%s", runtime.GOOS, runtime.GOARCH, archiveExtension())
+	tempArchivePath := filepath.Join(tempDir, archiveName)
+
+	if err := m.downloadFile(ctx, downloadURL, tempArchivePath, writer); err != nil {
```
wbreza, line 371: **[🟡 Medium] MSI download URL hardcoded to amd64**

`buildMSIDownloadURL` always returns `azd-windows-amd64.msi` regardless of `runtime.GOARCH`. Windows ARM64 devices would download the wrong architecture.

Compare with `buildDownloadURL` which correctly uses `runtime.GOARCH`.

```suggestion
return fmt.Sprintf("%s/%s/azd-windows-%s.msi", blobBaseURL, folder, runtime.GOARCH), nil
```

cli/azd/cmd/update.go
```diff
@@ -0,0 +1,346 @@
...
+	}
+}
+
+func (a *updateAction) Run(ctx context.Context) (*actions.ActionResult, error) {
+	// Dev builds (0.0.0-dev.0) should not self-update — developers build from source.
+	if internal.IsDevVersion() {
+		return nil, fmt.Errorf(
+			"azd update is not supported for dev builds. Build from source with `go build` instead")
+	}
+
```
wbreza, line 109: **[🟡 Medium] Auto-enable alpha feature is a surprising side effect**

Running `azd update` silently enables the `update` alpha feature in persistent config. While the comment says "The user''s intent is clear," alpha features typically require explicit opt-in. This modifies persistent config as a side effect, potentially enabling behavior the user didn''t expect (e.g., background staging starts, "out of date" banner text changes).

Consider at minimum informing the user more prominently that this alpha feature has been enabled and what it means.

cli/azd/main.go
```diff
@@ -95,65 +145,41 @@ func main() {
...
-				}
-			} else {
-				// Platform is not recognized, use the generic install link
-				upgradeText = "visit https://aka.ms/azd/upgrade"
+		} else if versionInfo.HasUpdate {
+			currentVersionStr := internal.VersionInfo().Version.String()
+			latestVersionStr := versionInfo.Version
+			if versionInfo.BuildNumber > 0 {
+				latestVersionStr = fmt.Sprintf("%s (build %d)", versionInfo.Version, versionInfo.BuildNumber)
 			}
```
wbreza, line 159: **[🟠 High] Background goroutine uses `context.Background()` with no cancellation**

`fetchLatestVersion` creates its own `context.Background()` for HTTP calls and `StageUpdate`. This goroutine cannot be cancelled when the main command exits quickly — it could continue downloading a ~50MB binary after the user''s `azd version` command has already returned.

**Suggested Fix:** Pass a derived context with timeout, or cancel it after the main command completes.

cli/azd/main.go
```diff
@@ -95,65 +145,41 @@ func main() {
...
```
wbreza, line 158: **[🟡 Medium] Manager bypasses IoC container (violates repo convention)**

`update.NewManager(nil)` is created with direct instantiation and a nil command runner. The repo''s AGENTS.md states: "MANDATORY use of dependency injection (never direct instantiation)."

The nil command runner means `StageUpdate` path has no code signature verification capability, and tests must modify global state (`http.DefaultTransport`) instead of injecting test clients.

cli/azd/pkg/update/manager.go
```diff
@@ -0,0 +1,1003 @@
...
+	exePath, err := os.Executable()
+	if err != nil {
+		return "", err
+	}
+	return filepath.EvalSymlinks(exePath)
+}
+
+func copyFile(src, dst string) error {
+	in, err := os.Open(src)
+	if err != nil {
```
wbreza, line 604: **[🟡 Medium] Loose archive extraction matching**

The matching `name == binaryName || strings.HasPrefix(name, "azd-")` is very broad. Any file starting with "azd-" in the archive would be extracted, regardless of what it actually is.

**Suggested Fix:** Only match the exact binary name (`"azd"` or `"azd.exe"`), or at minimum validate the full expected filename.

cli/azd/pkg/update/manager_test.go
```diff
@@ -0,0 +1,662 @@
...
+			require.Equal(t, tt.want, PackageManagerUninstallCmd(tt.installedBy))
+		})
+	}
+}
+
+func TestBuildVersionInfoFromCache_Stable(t *testing.T) {
+	m := NewManager(nil)
+
+	tests := []struct {
+		name      string
```
wbreza, line 142: **[🟡 Medium] Tests modify `http.DefaultTransport` globally**

Swapping `http.DefaultTransport` is a global side effect that isn''t safe for parallel test execution. If `t.Parallel()` is ever added, tests will race.

**Suggested Fix:** Inject an `*http.Client` into the Manager struct and use that instead of `http.DefaultClient`. Tests can then provide a client with a custom transport.

vhvb1989:
> ## Consolidated Code Review — PR #6942 vs Design Review (#6910)
> ### TL;DR
> The implementation is well-engineered at the code level — clean separation of concerns, thoughtful platform handling, solid test coverage, and proper integration with existing telemetry and error infrastructure. However, **none of the 4 critical findings from the design review (#6910) have been addressed**, and several new implementation-level issues were found.
> 
> ### Design Feedback Resolution
> Category	Raised in #6910	Addressed in #6942	Unresolved
> 🔴 Critical	4	0	**4**
> 🟠 High	8	1 (partial)	**7**
> 🆕 New in implementation	—	—	**8**
> #### 🔴 Still-Unresolved Criticals from #6910
> 1. **Unsigned binaries accepted** — `verifyStagedBinary` still allows unsigned binaries (`"not signed at all" → OK`). Note: danieljurek commented "Sha does not need to be checked as the download is occurring over HTTPS only" — if the team has accepted this, please document the threat model decision.
> 2. **No rollback/recovery** — `ApplyStagedUpdate()` overwrites the current binary with no backup. A broken update requires manual reinstall.
> 3. **No crash-loop guard** — No sentinel file or `AZD_SKIP_AUTO_UPDATE` escape hatch. The marker file is display-only. If the new binary panics during startup before completing initialization, users have no env var to bypass auto-update.
> 4. **No version pinning** — No `--version` flag or `.azd-version` file. `azd update` always gets "latest".
> 
> #### 🟠 Still-Unresolved Highs from #6910
> 5. **No checksum verification** — Downloads trust HTTPS only
> 6. **TOCTOU between stage and apply** — No integrity hash stored at staging time; staged binary sits in user-writable space
> 7. **"daily" vs "nightly"** naming — Team disagreement unresolved
> 8. **MSI for all script installs** — Even PowerShell-script-installed azd uses MSI (escalates privilege requirements)
> 9. **Command conflates config with action** — `azd update --auto-update on --channel daily` does both
> 10. **"update" vs "upgrade"** naming — hemarina: "upgrade" for consistency; kristenwomack: "update" is standard
> 11. **No concurrent update locking** — No file locking on staging directory
> 
> ### 🆕 New Implementation Issues
> See inline comments for details on each.
> 
> ### ✅ What Looks Good
> * **Staged update architecture** follows VS Code/Chrome pattern — download in background, apply on next startup
> * **Multi-layered opt-in** (alpha flag + auto-update setting + CI detection) is appropriately conservative
> * **ETXTBSY handling on Linux** (remove-then-create in `copyFile`) is a thoughtful detail many implementations miss
> * **Platform-specific MSI/detached process** correctly handles binary-in-use constraint on Windows
> * **Telemetry integration** with structured error codes follows existing `MapError` patterns perfectly
> * **Comprehensive test coverage** — unit tests for parsing, caching, extraction, staging lifecycle, HTTP transport, and error cases
> * **Error types** with `UpdateError` integrate cleanly with existing error infrastructure
> 
> **Overall Assessment:** Request changes

This looks like a GA-ready roadmap for the feature.   
It should be OK to move on as we are starting with the feature in alpha mode.  @wbreza let it be in alpha and let's use your list to iterate the feature to graduation.

cli/azd/internal/tracing/fields/fields.go
```diff
@@ -581,3 +581,37 @@ var (
...
 )
+
+// Update related fields
+var (
+	// UpdateChannel is the update channel (stable, daily).
+	UpdateChannel = AttributeKey{
+		Key:            attribute.Key("update.channel"),
+		Classification: SystemMetadata,
+		Purpose:        FeatureInsight,
+	}
```
danieljurek, line 592: What is gained by setting the channel as a configuration property? Why not have the channel be a property of the azd binary itself. If the user wishes to switch channels, they can switch with `azd upgrade --channel`. I don't think there's benefit to storing the channel itself.

cli/azd/pkg/update/config.go
```diff
@@ -0,0 +1,215 @@
...
+	configKeyAutoUpdate = "updates.autoUpdate"
+	// configKeyCheckIntervalHours is the config key for the check interval.
+	configKeyCheckIntervalHours = "updates.checkIntervalHours"
+)
+
+const (
+	// DefaultCheckIntervalStable is the default check interval for the stable channel.
+	DefaultCheckIntervalStable = 24 * time.Hour
+	// DefaultCheckIntervalDaily is the default check interval for the daily channel.
+	DefaultCheckIntervalDaily = 4 * time.Hour
```
danieljurek, line 60: This means the user could receive multiple update signals in a day. We do produce daily builds from every commit to `main`.

cli/azd/pkg/update/manager.go
```diff
@@ -0,0 +1,1003 @@
...
+	"github.com/azure/azure-dev/cli/azd/pkg/installer"
+	"github.com/azure/azure-dev/cli/azd/pkg/osutil"
+	"github.com/blang/semver/v4"
+)
+
+const (
+	// stableVersionURL is the endpoint that returns the latest stable version.
+	stableVersionURL = "https://aka.ms/azure-dev/versions/cli/latest"
+	// blobBaseURL is the base URL for Azure Blob Storage where azd binaries are hosted.
+	blobBaseURL = "https://azuresdkartifacts.z5.web.core.windows.net/azd/standalone/release"
```
danieljurek, line 35: This duplicates functionality already present in the install scripts which means that in the event of a need to change hosting setups, we run the risk of breaking customers with older versions when the storage moves. 

The install scripts already verify signatures before installing (where appropriate [windows](https://github.com/Azure/azure-dev/blob/main/cli/installer/install-azd.ps1#L338), [mac](https://github.com/Azure/azure-dev/blob/main/cli/installer/install-azd.sh#L300)). 

The install scripts are handled by aka.ms and will have a pretty stable interface.

The process should instead be: 

## Windows 

1. Download aka.ms/install-azd.ps1, launch that script with the appropriate channel and  
2. Quit as quickly as possible to not block MSI 

Addendum: Per-machine installs will fail when launched from install-azd.ps1. In that case, azd should detect a per-machine install and have the user uninstall azd and install again using the install script.

## Linux/MacOS

### Installed using brew

TBD -- It may be possible to set up a dev channel formula

### Installed using script 

1. Run aka.ms/install-azd.sh, launch with appropriate channel information ... A privilege escalation may be required
2. Quit (this is not as big of a concern as linux/macos allow you to replace a running binary)

cli/azd/internal/tracing/fields/fields.go
rajeshkamal5050: Good question Daniel.

The channel config helps  determine the check and download URLs, and check interval (24h stable, 4h daily). Users can override interval with --check-interval-hours which is persisted alongside  channel.

cli/azd/pkg/update/config.go
rajeshkamal5050: Yeah, this could be noisy for folks consuming daily builds. We can increase the defaults too. Also, the --check-interval-hours is overridable by users.

cli/azd/pkg/update/manager.go
rajeshkamal5050: This is a great suggestion and could simplify some of the logic here. 

Let us get this PR merged. So, you can work on `Updating and leveraging existing install scripts to azd update`.

rajeshkamal5050:
@wbreza Thanks for the comments 👍

All fixed in c9db168a:
```
- Code signature hard-fail — verifyStagedBinary now returns CodeSignatureInvalid error instead of silently succeeding.
- http.DefaultClient no timeout — NewManager accepts injected *http.Client (defaults to 30s timeout). All HTTP calls go through m.httpClient.Do().
- MSI hardcoded to amd64 — buildMSIDownloadURL now uses runtime.GOARCH.
- Alpha auto-enable surprising — Updated banner to clearly state the feature has been auto-enabled and how to disable.
- Background goroutine no cancellation — Now uses context.WithTimeout(60s), cancel called after channel receive.
- Manager bypasses IoC — Intentional. main.go runs pre-IoC for ApplyStagedUpdate at startup (same pattern as tracing init). cmd/update.go goes through IoC normally.
- Loose archive extraction — Replaced broad HasPrefix("azd-") with isAzdBinary() matching exact name or azd-{GOOS}-{GOARCH}.
- Tests modify DefaultTransport — Added testClientWithRewrite() helper with isolated *http.Client. No more global transport swapping.
- Unsigned binaries accepted. Codesign verification now hard-fails. See "Code signature hard-fail" 
``` 

Pending enhancements tracked as new issues under epic #6721:

```
- No rollback/recovery — #6982
- No crash-loop guard — #6983
- No version pinning — #6984
- No checksum verification / TOCTOU — #6985. Download is HTTPS-only today. Will evaluate SHA256 verification.
- No concurrent update locking — #6986
```

cli/azd/pkg/update/manager.go
rajeshkamal5050: Tracked via https://github.com/Azure/azure-dev/issues/6987

cli/azd/pkg/update/manager.go
```diff
@@ -0,0 +1,1023 @@
...
+	}
+	// Don't defer os.Remove — the detached msiexec process needs this file after we exit.
+
+	// Build msiexec args. Always write a verbose log so failures are diagnosable.
+	msiLogPath, logErr := msiLogFilePath()
+	args := []string{"/i", msiPath, "/qn"}
+	if logErr == nil {
+		args = append(args, "/l*v", msiLogPath)
+		log.Printf("MSI install log: %s", msiLogPath)
+	}
```
weikanglim, line 339: If MSI install fails, there wouldn't be any user feedback. Does it make sense to show something when azd starts up the next time?

rajeshkamal5050: Yeah good point. Was also thinking about it and enabled more debug logs pointing to the msi logs. But we can definitely improve here.
- Tracking smarter failure detection as part of #6982 (rollback/recovery on failed update).
- Also, @danieljurek  is looking into leveraging the existing Day-0 install scripts for Day-N updates too (#6987). That could simplify the Windows update path and give us better error handling out of the box.

wbreza (commented):
## Re-review Summary (PR #6942)

Re-evaluating my previous Critical findings with additional design context from the team:

### Previous Criticals — Revised
| Issue | Previous | Revised | Rationale |
|-------|----------|---------|-----------|
| Unsigned binaries | Critical | Resolved | Code signature now hard-fails for invalid signatures. Unsigned dev builds being allowed is intentional |
| No rollback | Critical | Medium | For a CLI tool with signature verification, corrupt update risk is low. Users can reinstall. Nice-to-have, not a blocker |
| No crash-loop guard | Critical | Medium | `ApplyStagedUpdate` consumes staged binary, so auto-update won't loop. Alpha feature with CI exclusion |
| No version pinning | Critical | By Design | `azd update` is forward-only, not a package manager |

### What Improved Since Last Review
- HTTP client injection for testability
- Code signature hard-fail for invalid signatures
- Context timeout added
- Dev build blocking (0.0.0-dev.0)
- Linux ETXTBSY handling

### Remaining Suggestion
Consider documenting the recovery path (reinstall via `curl` or MSI) in the `azd update` output or help text, so users know what to do if an update goes wrong.

**Assessment:** Progress made on previous feedback. The design decisions around no version pinning and forward-only updates make sense for the tool's use case.

cli/azd/cmd/update.go
```diff
@@ -0,0 +1,346 @@
...
+		confirm, err := a.console.Confirm(ctx, input.ConsoleOptions{
+			Message:      confirmMsg,
+			DefaultValue: true,
+		})
+
+		if err != nil || !confirm {
+			tracing.SetUsageAttributes(fields.UpdateResult.String(update.CodeChannelSwitchDecline))
+			a.console.Message(ctx, "Channel switch cancelled.")
+			return nil, nil
+		}
```
wbreza, line 274: **[Medium] Non-channel config flags lost when channel switch is declined**

If the user provides `--auto-update on --channel daily` and declines the channel switch confirmation, `persistNonChannelFlags` modifies the in-memory config but the function returns at line 326 without saving to disk. The `--auto-update on` setting is silently discarded.

Consider saving non-channel flags before the confirmation prompt, or at minimum documenting this behavior.

*Non-blocking - edge case with straightforward fix.*

cli/azd/pkg/update/manager.go
```diff
@@ -0,0 +1,1023 @@
...
+	result, err := m.commandRunner.Run(ctx, runArgs)
+	if err != nil {
+		return newUpdateError(CodePackageManagerFailed, err)
+	}
+
+	if result.ExitCode != 0 {
+		return newUpdateErrorf(CodePackageManagerFailed,
+			"package manager update failed with exit code %d", result.ExitCode)
+	}
+
```
wbreza, line 313: **[Low] MSI temp file name still hardcoded to amd64**

The download URL now correctly uses `runtime.GOARCH`, but the temp file path still says `azd-windows-amd64.msi`. Cosmetic only (temp file), but inconsistent.

cli/azd/cmd/update.go
rajeshkamal5050: Would probably prefer keeping the operation atomic. The user declined the whole operation, so discarding everything seems okay. If this comes up, we can address it.

cli/azd/pkg/update/manager.go
rajeshkamal5050: Yeah cosmetic but agree on the inconsistency.

This code path most likely gets replaced by #6987 (leverage existing install scripts).

PR #6962, opened 2026-03-03
better no-prompt guidance for `init` and `provision`

weikanglim:
## Summary

When `azd` runs with `--no-prompt` and required inputs are missing, it now reports **all** missing inputs at once with actionable resolution commands — instead of failing one-at-a-time with cryptic prompt errors.

### `azd provision --no-prompt`

When required Bicep parameters are missing, the error lists every missing parameter with its environment variable mapping, config key, type, constraints, and allowed values:

```
──────────────────────────────────────────────────────────────
Provision cannot continue (interactive prompts disabled)
──────────────────────────────────────────────────────────────

3 required inputs are missing.

Missing required inputs:

  • location
      Environment variable: AZURE_LOCATION
      Environment configuration key: infra.parameters.location
      Allowed values: eastus, westus2

  • sqlAdminPassword
      Environment configuration key: infra.parameters.sqlAdminPassword
      Type: string
      Constraints:
          Min length: 12
          Secure: true
```

With `--output json`, the same data is returned as structured JSON with typed constraint fields.

### `azd init --no-prompt`

When no mode flag (`--minimal`, `--template`) is provided, the error explains the available options with exact commands:

```
──────────────────────────────────────────────────────────────
Init cannot continue (interactive prompts disabled)
──────────────────────────────────────────────────────────────

azd init requires an initialization mode when --no-prompt is set.

Choose one:

  • Minimal (no template)
      Creates required azd project files in the current directory.
      azd init --minimal

  • From template
      Creates a new project from an azd template.
      azd template list
      azd init --template <template-id> --environment <environment>

Environment name must be globally unique (for example: myapp-dev).
```

### `Console.IsNoPromptMode()`

Exposes whether `--no-prompt` is active on the `Console` interface, so any component can detect non-interactive mode without depending on `GlobalCommandOptions`.

### UxItem errors attach rich output

The UX middleware now checks whether a returned error implements `ux.UxItem`. If it does, the error's `ToString()` and `MarshalJSON()` methods are used to render additional structured output after the standard `ERROR:` line. This means any error in the codebase can opt into rich terminal and JSON output simply by implementing the `UxItem` interface — no middleware changes required.

### DIFF
--- a/cli/azd/pkg/infra/provisioning/bicep/missing_inputs_test.go
+++ b/cli/azd/pkg/infra/provisioning/bicep/missing_inputs_test.go
@@ -271,9 +271,9 @@ func TestMissingInputsError_MarshalJSON(t *testing.T) {
 	}
 	require.NoError(t, json.Unmarshal(jsonData, &result))
 
-	assert.Equal(t, "missingInputs", result.Code)
+	assert.Equal(t, "provisionMissingInputs", result.Code)
 	assert.Equal(t, "Provision cannot continue (interactive prompts disabled)", result.Message)
-	assert.Equal(t, "missingInputs", result.Details.Type)
+	assert.Equal(t, "provisionMissingInputs", result.Details.Type)
 	require.Len(t, result.Details.Inputs, 1)
 
 	input := result.Details.Inputs[0]
--- a/cli/azd/test/functional/testdata/samples/funcapp/infra/main.bicep
+++ b/cli/azd/test/functional/testdata/samples/funcapp/infra/main.bicep
@@ -6,15 +6,6 @@ param environmentName string
 @description('Primary location for all resources')
 param location string
 
-param vnetEnabled bool
-
-@minLength(10)
-param resourceGroupOverride string
-
-@sys.secure()
-@maxLength(100)
-param secure string
-
 @description('A time to mark on created resource groups, so they can be cleaned up via an automated process.')
 param deleteAfterTime string = dateTimeAdd(utcNow('o'), 'PT1H')
 
--- a/cli/azd/test/functional/testdata/samples/funcapp/infra/main.parameters.json
+++ b/cli/azd/test/functional/testdata/samples/funcapp/infra/main.parameters.json
@@ -7,9 +7,6 @@
     },
     "location": {
       "value": "${AZURE_LOCATION}"
-    },
-    "resourceGroupOverride": {
-      "value": "${AZURE_RESOURCE_GROUP}"
     }
   }
 }


cli/azd/pkg/infra/provisioning/bicep/bicep_provider.go
```diff
@@ -2356,6 +2356,11 @@ func (p *BicepProvider) ensureParameters(
 		}{key: key, param: param})
 	}
 
+	// If in no-prompt mode and there are missing parameters, return an error with all missing inputs
+	if len(parameterPrompts) > 0 && p.console.IsNoPromptMode() {
+		return nil, p.buildMissingInputsError(parameterPrompts, parametersResult.envMapping)
+	}
+
```
vhvb1989, line 2363: I have some concerns about this strategy -  

The reason we are adding this is to help AI Agents running azd with `--no-prompt` right?
Aren't we worried about how the agent will handle the error after it?  
If the Agent is running in a supervised session, will it ask a human to provide the values for all the inputs?  
If the Agent is running in CI, will it fail anyway? or try to resolve the inputs by its own?

`Inputs experience` is one of the greatest features of azd - and one of the main differences to deploying a bicep template with az.  This strategy feels like we are trying to move one issue away from azd, but I don't think the issue will disappear. It might be only moving somewhere else 🫠

A similar scenario can be seen when AI Agents runs a command that requires elevation in Linux and prompts for the user password. In my experience, AI Agents would communicate to humans that there is action needed in the current command execution and give a choice to bring the terminal and focus where the user needs to prompt the password. 
If they would follow the approach in this PR, the command would return an error asking for the user password, the agent would ask for it and then use it to try again. 

What is going to happen to secure inputs, for example. If the AI Agent prompt for it, the secret value would be part of the conversation session? 

I would rather like to see azd making AI Agents to know what the focus must be swap to the terminal for a user input.

cli/azd/cmd/middleware/ux.go
```diff
@@ -108,6 +108,14 @@ func (m *UxMiddleware) Run(ctx context.Context, next NextFn) (*actions.ActionRes
...
+		// Print out additional text for errors that have it.
+		var uxItemErr ux.UxItem
+		if errors.As(err, &uxItemErr) {
+			m.console.Message(ctx, "")
+			m.console.MessageUxItem(ctx, uxItemErr)
+			return actionResult, err
+		}
 	}
 
 	if actionResult != nil && actionResult.Message != nil {
```
wbreza, line 121: **[High]** Dual error rendering path + double output

This creates a **second** rich error rendering path in the middleware. The existing flow is:

1. `ErrorWithSuggestion` caught at line 57 → renders rich output → returns *(no standard ERROR: line)*
2. Extension errors caught at line 71 → renders rich output → returns *(no standard ERROR: line)*

The new `UxItem` check comes **after** `m.console.Message(ctx, errMessage)` (line 110), so users see **both**:
`
ERROR: initialization mode required when --no-prompt is set

──────────────────────────────────────────────────────────────
Init cannot continue (interactive prompts disabled)
──────────────────────────────────────────────────────────────
azd init requires an initialization mode when --no-prompt is set.
`

Three slightly different phrasings of the same issue appear in sequence.

**Recommendation:** Instead of adding a new catch-all `UxItem` path here, integrate these errors into the existing error suggestion pipeline:

1. Register `MissingInputsError` and `initModeRequiredError` as error types in `error_suggestions.yaml` with custom handlers
2. Or extend `ErrorWithSuggestion` to support a `Details UxItem` field for structured content
3. This keeps all error rendering centralized in the existing `ErrorWithSuggestion` → `ux.ErrorWithSuggestion` path

If the `UxItem` approach is preferred, at minimum move this check **above** line 96 (before the standard `ERROR:` message is printed) to match how `ErrorWithSuggestion` is handled — catch early and render only the rich version.

cli/azd/pkg/infra/provisioning/bicep/missing_inputs.go
```diff
@@ -0,0 +1,219 @@
...
+
+	buf.WriteString(separator + "\n\n")
+	buf.WriteString("You can resolve these by:\n\n")
+
+	if e.hasEnvVars() {
+		buf.WriteString("1) Setting environment variables\n")
+		buf.WriteString("     azd env set <ENV_VAR_NAME> <value>\n\n")
+	}
+
+	buf.WriteString("2) Setting environment configuration\n")
```
wbreza: **[High]** Step numbering bug when no env vars exist

When `hasEnvVars()` returns `false`, step `1)` is skipped but `2)` is unconditionally printed:

`
You can resolve these by:

2) Setting environment configuration
     azd env config set infra.parameters.<paramName> <value>
`

Note: `TestMissingInputsError_Error_NoEnvVars` asserts this broken output (checks for `2)` without `1)`).

`suggestion
step := 1
if e.hasEnvVars() {
buf.WriteString(fmt.Sprintf("%d) Setting environment variables\n", step))
buf.WriteString("     azd env set <ENV_VAR_NAME> <value>\n\n")
step++
}

buf.WriteString(fmt.Sprintf("%d) Setting environment configuration\n", step))
`
"@
        },
        @{
            path = "cli/azd/pkg/infra/provisioning/bicep/missing_inputs.go"
            line = 28
            side = "RIGHT"
            body = @"
**[Medium]** Missing `omitempty` on optional JSON fields

Fields like `description`, `envVarNames`, `allowedValues`, `secure`, and `constraints` lack `omitempty`, producing noisy JSON for simple parameters:

`json
{ ""secure"": false, ""description"": """", ""envVarNames"": null, ""allowedValues"": null, ""constraints"": {} }
`

The `InputConstraints` fields already correctly use `omitempty`. Apply the same pattern here:

`suggestion
type MissingInput struct {
Name          string           `json:""name""`
Type          string           `json:""type""`
Secure        bool             `json:""secure,omitempty""`
Description   string           `json:""description,omitempty""`
EnvVarNames   []string         `json:""envVarNames,omitempty""`
ConfigKey     string           `json:""configKey""`
AllowedValues []string         `json:""allowedValues,omitempty""`
Constraints   InputConstraints `json:""constraints,omitempty""`
}
`
"@
        },
        @{
            path = "cli/azd/cmd/init.go"
            line = 902
            side = "RIGHT"
            body = @"
**[Medium]** Duplicate message string + consider error suggestion YAML integration

1. **DRY:** The message `""Init cannot continue (interactive prompts disabled)""` is duplicated between `ToString()` (here) and `MarshalJSON()` (line 930). Extract to a `const`.

2. **Design:** Since `initModeRequiredError` has **static content** (unlike `MissingInputsError` which has dynamic parameter lists), it's a strong candidate for the existing `error_suggestions.yaml` pipeline. A YAML rule matching `errorType: ""initModeRequiredError""` with a handler could produce the same output through the established path. The only blocker is the custom JSON structure (`options` array), which `ErrorWithSuggestion`'s fixed schema doesn't support — this is the core argument for extending `ErrorWithSuggestion` with a structured details field.

cli/azd/cmd/init_test.go
```diff
@@ -68,6 +69,50 @@ func runActionSafe(ctx context.Context, action *initAction) (retErr error) {
...
+
+		action := setupInitAction(t, mockContext, flags)
+
+		err := runActionSafe(*mockContext.Context, action)
+		if err != nil {
+			var noPromptErr *initModeRequiredError
+			require.False(t, errors.As(err, &noPromptErr),
+				"should not return InitNoPromptError when --minimal is set")
+		}
+	})
```
wbreza, line 113: **[Medium]** Test gap: no test for `--template` flag with `--no-prompt`

`DoesNotErrorWhenMinimalFlagSet` covers `--minimal` + `--no-prompt`, but there's no parallel test for `--template <id> --environment <env>` + `--no-prompt`. This is the other valid init mode and should be verified not to produce `initModeRequiredError`.

wbreza (commented):
## Code Review — PR #6962

### Overall Assessment: **Comment** — Suggest architectural alignment before merge

This is a valuable feature — aggregating all missing inputs into a single actionable error for --no-prompt is a big UX win for CI/CD and agent workflows. The implementation is solid with good test coverage.

**Primary recommendation:** Align to a **single rich error reporting path** by extending the existing ErrorWithSuggestion infrastructure rather than creating a parallel UxItem-based error path in the middleware. The codebase already has a well-established pipeline:

1. `error_suggestions.yaml` — ~180+ rules matching errors → `ErrorWithSuggestion`
2. `ErrorMiddleware` runs `errorPipeline.Process(ctx, err)`
3. `UxMiddleware` catches `ErrorWithSuggestion` and renders via `ux.ErrorWithSuggestion` (which implements `UxItem`)

**Suggested approach:**
- Extend `ErrorWithSuggestion` (or create a new `ErrorWithDetails`) with a `Details UxItem` field for structured rich content
- Register error type handlers in the YAML pipeline (e.g., `errorType: "MissingInputsError"` with a `handler`) so these errors flow through the same pipeline
- This keeps all error-to-user-message logic centralized and discoverable

See inline comments for specific findings.

### Findings Summary

| Priority | Count |
|----------|-------|
| High | 3 |
| Medium | 4 |
| Low | 2 |
| **Total** | **9** |

### ✅ What Looks Good
- Aggregating all missing inputs instead of one-at-a-time failures
- Clean `ToString()` / `MarshalJSON()` separation
- Thorough test coverage for `MissingInputsError` (309 lines)
- `Console.IsNoPromptMode()` is a clean abstraction
- Mock updates are complete for both `MockConsole` implementations

cli/azd/pkg/infra/provisioning/bicep/bicep_provider.go
weikanglim: The change we're making here is to provide better guidance on next steps here, when `--no-prompt` is set. 
Right now, we have autodetection that defaults `--no-prompt` automatically. I can see us removing that with the prompting experience fixed in copilot-cli, so both experiences live side-by-side.

Regarding secure inputs, it winds down to the agent implementation. There are many ways for the agent to prompt securely to the user even inside a session. Surfacing this metadata allows for that decision to be made.

We aren't removing the step-by-step prompting experience. There is value in having both, but this does help drive better error loops when the error was "user forgot to provide inputs to fields".

cli/azd/cmd/middleware/ux.go
weikanglim: Removed the message. Agree on standardization of print out, but `error_suggestions.yaml` with custom handlers, or extend `ErrorWithSuggestion` to support a `Details UxItem` field for structured content aren't necessarily the best ways to go about it. We can address this separately.

wbreza (commented):
## Re-Review: PR #6962 — better no-prompt guidance for `init` and `provision`

### Prior Feedback Status
✅ Step numbering bug — fixed with dynamic `optionNum` counter
✅ Missing `--template` test — added `DoesNotErrorWhenTemplateAndEnvironmentProvided`
✅ JSON `omitempty` — fixed on optional fields
⚠️ Error rendering path standardization — acknowledged, deferred

Great follow-through on the feedback.

---

### Findings Summary

| Priority | Count |
|----------|-------|
| 🟠 High | 1 |
| 🟡 Medium | 2 |
| 🟢 Low | 1 |
| **Total** | **4** |

---

### Architectural Note: Error → Suggestion Flow

Reinforcing from prior review — we should ensure a consistent flow for any time we return an error that includes suggestions on how to recover. Today we have `ErrorWithSuggestion` flowing through `error_suggestions.yaml` → `ErrorMiddleware` → `UxMiddleware`. This PR adds a parallel `UxItem`-based path in the middleware. As we add more rich error types (agents and CI/CD workflows will surface more of these), having two rendering pipelines will create maintenance burden and inconsistent UX. I'd like to see a follow-up to unify these before adding more error types through this path.

---

### ✅ What Looks Good
- Aggregating all missing inputs into a single actionable error is a big UX win for CI/CD and agent workflows
- Good test coverage (380+ test lines) with constraint edge cases and JSON roundtrip validation
- `Console.IsNoPromptMode()` is a clean abstraction with complete mock implementations
- Step numbering properly fixed with dynamic counter — good follow-through

**Overall Assessment:** Comment — The Constraints/omitempty discrepancy between the builder and tests should be addressed. Remaining items are minor.

cli/azd/pkg/infra/provisioning/bicep/missing_inputs.go
```diff
@@ -0,0 +1,237 @@
...
+			},
+		}
+
+		inputs = append(inputs, input)
+	}
+
+	return &MissingInputsError{
+		Inputs: inputs,
+	}
+}
```
wbreza, line 237: **[High]** `buildMissingInputsError` always allocates Constraints, defeating `omitempty`

`Constraints` is always set to `&InputConstraints{...}` (line 231) even when all pointer fields (MinLength, MaxLength, MinValue, MaxValue) are nil. Since `Constraints` is `*InputConstraints` with `json:"constraints,omitempty"`, a non-nil pointer to a zero-value struct serializes as `"constraints":{}` instead of being omitted.

The test `TestMissingInputsError_MarshalJSON_OmitsEmptyConstraints` passes because it constructs `MissingInput` directly with nil Constraints, but real output from `buildMissingInputsError` will always include `"constraints":{}` for parameters without constraint metadata.

**Suggested fix:** Only allocate when at least one field is non-nil:

```suggestion
Constraints: buildInputConstraints(param),
```

Then add a helper:
```go
func buildInputConstraints(param azure.ArmTemplateParameterDefinition) *InputConstraints {
if param.MinLength == nil && param.MaxLength == nil && param.MinValue == nil && param.MaxValue == nil {
return nil
}
return &InputConstraints{
MinLength: param.MinLength,
MaxLength: param.MaxLength,
MinValue:  param.MinValue,
MaxValue:  param.MaxValue,
}
}
```

cli/azd/pkg/infra/provisioning/bicep/missing_inputs.go
```diff
@@ -0,0 +1,237 @@
...
+
+// ToString returns a formatted message with the missing inputs and resolution guidance.
+func (e *MissingInputsError) ToString(currentIndentation string) string {
+	var buf strings.Builder
+	separator := "──────────────────────────────────────────────────────────────"
+
+	buf.WriteString(separator + "\n")
+	buf.WriteString("Provision cannot continue (interactive prompts disabled)\n")
+	buf.WriteString(separator + "\n\n")
```
wbreza, line 51: **[Medium]** Duplicate message string literal

This message `"Provision cannot continue (interactive prompts disabled)"` is duplicated between `ToString()` here and `MarshalJSON()` (line 160). Same issue exists for `initModeRequiredError` in `init.go`. Consider extracting to a `const` to keep them in sync.

cli/azd/pkg/infra/provisioning/bicep/missing_inputs.go
```diff
@@ -0,0 +1,237 @@
...
+}
+
+// Error implements the error interface for MissingInputsError.
+func (e *MissingInputsError) Error() string {
+	return "missing required inputs"
+}
+
+// ToString returns a formatted message with the missing inputs and resolution guidance.
+func (e *MissingInputsError) ToString(currentIndentation string) string {
+	var buf strings.Builder
```
wbreza, line 46: **[Medium]** `currentIndentation` parameter unused

`ToString(currentIndentation string)` accepts indentation but never applies it. Other `UxItem` implementations use this for nested rendering. If these errors are rendered inside another UxItem, the output won't indent. Consider either applying it or documenting why it's intentionally ignored.

cli/azd/pkg/infra/provisioning/bicep/missing_inputs.go
```diff
@@ -0,0 +1,237 @@
...
+	separator := "──────────────────────────────────────────────────────────────"
+
```
wbreza, line 48: **[Low]** Consider extracting the separator pattern

Both error types define `separator := "──────────────────────────────────────────────────────────────"` independently (here and in `init.go`). Could be a shared constant if this pattern will be reused by future error types.

cli/azd/pkg/infra/provisioning/bicep/missing_inputs.go
weikanglim: I didn't notice this, but if I had to pick, I'd vote for showing `"constraints":{}`. I also prefer the simpler codepath. This wouldn't affect consumers either way.

cli/azd/pkg/infra/provisioning/bicep/missing_inputs.go
weikanglim: Leaving as-is until we decide to generalize this pattern

cli/azd/pkg/infra/provisioning/bicep/missing_inputs.go
weikanglim: Leaving as-is

cli/azd/pkg/infra/provisioning/bicep/missing_inputs.go
weikanglim: This is fine

rajeshkamal5050 (approved):

PR #6963, opened 2026-03-03
cache resource display names during provisioning

weikanglim:
Add caching for resource type display names in `ProvisioningProgressDisplay` to avoid redundant API calls for the same resource types during a single provisioning operation. 

In particular, the in-progress resources were constantly fetched which would slow down the entire display.

### DIFF
--- a/cli/azd/pkg/infra/provisioning_progress_display.go
+++ b/cli/azd/pkg/infra/provisioning_progress_display.go
@@ -27,9 +27,11 @@ type ProvisioningProgressDisplay struct {
 	deploymentStarted bool
 	// Keeps track of created resources
 	displayedResources map[string]bool
-	resourceManager    ResourceManager
-	console            input.Console
-	deployment         Deployment
+	// Cache for display names, keyed by resource IDs
+	resourceDisplayNames map[string]string
+	resourceManager      ResourceManager
+	console              input.Console
+	deployment           Deployment
 }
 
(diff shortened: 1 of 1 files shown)


wbreza (approved):
## Code Review — PR #6963

### TL;DR
Clean, well-targeted optimization that caches resource display names during provisioning to avoid redundant lookups on each 3-second progress tick.

### Key Observation
Most resource types (~40) already resolve via a **static switch statement** in `azapi.GetResourceTypeDisplayName()` — no API call. The real benefit of this cache is for the **3 dynamic resource types** (`Microsoft.Web/sites`, `Microsoft.CognitiveServices/accounts`, `Microsoft.Cache/redisEnterprise`) that make REST API calls to `ResourceService.GetResource()` to inspect the resource's `Kind` property. These are the ones that were slowing down the progress display when they appeared as in-progress resources.

### What Looks Good
- **`close(cancelProgress)`** is more idiomatic Go than sending a value, and eliminates a potential deadlock if the goroutine exits before the defer runs
- **Cache-on-first-access** pattern consolidates lookup + error fallback + caching into a single helper, eliminating duplicated error-handling blocks
- **Caching empty strings** prevents repeated lookups for unknown resource types

### Summary
| Priority | Count |
|----------|-------|
| Critical | 0 |
| High | 0 |
| Medium | 0 |
| Low | 0 |

**Overall Assessment:** ✅ Approve

PR #6969, opened 2026-03-03
Add x-ms-client-request-id header and use azdext.Run for trace ID propagation

saanikaguptamicrosoft:
### Notes
- This PR adds support for x-ms-correlation-request-id header propagation and switches the extension entrypoint to use azdext.Run.

**azdext.Run is the new recommended approach-**
- The previous entrypoint used context.Background(), which meant the extension had no awareness of the trace context from azd core. When azd launches an extension, it passes a TRACEPARENT environment variable containing the OpenTelemetry trace ID. azdext.Run calls NewContext() which reads this env var and injects the trace ID into the Go context — enabling end-to-end trace correlation. [[cp](https://github.com/Azure/azure-dev/blob/7ebd9e09898cae5f42a4a92c449ab5f33775f4a8/cli/azd/pkg/azdext/context.go#L29)]
- Replaced manual context.Background() + ExecuteContext + init() with azdext.Run, which handles FORCE_COLOR, TRACEPARENT injection, access token setup, and structured error reporting out of the box. One behavioral difference: errors now go to stderr (correct behavior) instead of stdout, and may include suggestion text for actionable errors.

### Testing
- azd x build - Succeeds
- Manual testing-
  - Added temporary changes for logging and to pick local azsdk changes: [commit1](https://github.com/Azure/azure-dev/pull/6969/changes/857e74f1c030aa36cc11396d63547e5e3211ce6c), [commit2](https://github.com/Azure/azure-dev/pull/6969/changes/1880a1e1207dcab7c74091bc096231ca54b09f84)
  - Run go mod tidy
  - Confirmed that `x-ms-client-request-id` is being set:
image

### DIFF
--- a/cli/azd/extensions/azure.ai.finetune/go.mod
+++ b/cli/azd/extensions/azure.ai.finetune/go.mod
@@ -103,5 +103,3 @@ require (
 	google.golang.org/grpc v1.78.0 // indirect
 	google.golang.org/protobuf v1.36.11 // indirect
 )
-
-replace github.com/azure/azure-dev/cli/azd => ../../
--- a/cli/azd/extensions/azure.ai.finetune/go.sum
+++ b/cli/azd/extensions/azure.ai.finetune/go.sum
@@ -39,6 +39,8 @@ github.com/aymanbagabas/go-udiff v0.2.0 h1:TK0fH4MteXUDspT88n8CKzvK0X9O2xu9yQjWp
 github.com/aymanbagabas/go-udiff v0.2.0/go.mod h1:RE4Ex0qsGkTAJoQdQQCA0uG+nAzJO/pI/QwceO5fgrA=
 github.com/aymerick/douceur v0.2.0 h1:Mv+mAeH1Q+n9Fr+oyamOlAkUNPWPlA8PPGR0QAaYuPk=
 github.com/aymerick/douceur v0.2.0/go.mod h1:wlT5vV2O3h55X9m7iVYN0TBM0NH/MmbLnd30/FjWUq4=
+github.com/azure/azure-dev/cli/azd v0.0.0-20260228232557-c482a92ca32c h1:uexNwoLux5JGdB/ciQYuydBmWNGUoTKVH8luiY//hrE=
+github.com/azure/azure-dev/cli/azd v0.0.0-20260228232557-c482a92ca32c/go.mod h1:zNtJ765AlE7Jsfy0T7qeiFHIHjNWLAYIXoPKVHVWpnk=
 github.com/bahlo/generic-list-go v0.2.0 h1:5sz/EEAK+ls5wF+NeqDpk5+iNdMDXrh3z3nPnH1Wvgk=
 github.com/bahlo/generic-list-go v0.2.0/go.mod h1:2KvAjgMlE5NNynlg/5iLrrCCZ2+5xWbdbCW3pNTGyYg=
 github.com/benbjohnson/clock v1.3.5 h1:VvXlSJBzZpA/zum6Sj74hxwYI2DIxRWuNIoXAzHZz5o=


cli/azd/extensions/azure.ai.finetune/internal/providers/factory/provider_factory.go
```diff
@@ -99,6 +99,7 @@ func WithTokenCredential(tokenCredential azcore.TokenCredential, scope string) o
 			InsecureAllowCredentialWithHTTP: true, // allow for plain HTTP proxies, etc..
 			PerCallPolicies: []policy.Policy{
 				azsdk.NewUserAgentPolicy(userAgent),
+				azsdk.NewMsClientRequestIdPolicy(),
```
wbreza, line 102: **[🟡 Medium]** Extension `go.mod` will need updating before this can be used

The extension's `go.mod` currently pins `github.com/azure/azure-dev/cli/azd` to `v0.0.0-20260228232557-c482a92ca32c`, which predates `NewMsClientRequestIdPolicy()`. The extension will need a `go.mod`/`go.sum` update to reference a version that includes this new function.

Additionally, consider also adding `azsdk.NewMsCorrelationPolicy()` here for consistency with the `azure.ai.agents` extension, which already propagates `x-ms-correlation-request-id`.

cli/azd/pkg/azsdk/correlation_policy.go
```diff
@@ -39,6 +40,12 @@ func NewMsCorrelationPolicy() policy.Policy {
 	return &simpleCorrelationPolicy{headerName: MsCorrelationIdHeader}
 }
 
+// NewMsClientRequestIdPolicy creates a policy that sets the x-ms-client-request-id header on HTTP requests.
+// This is used by Azure services to log and correlate individual requests for diagnostics.
+func NewMsClientRequestIdPolicy() policy.Policy {
+	return &simpleCorrelationPolicy{headerName: MsClientRequestIdHeader}
+}
```
wbreza, line 47: **[🟡 Medium]** Using trace ID for `x-ms-client-request-id` overrides per-request uniqueness

Reusing `simpleCorrelationPolicy` here sets `x-ms-client-request-id` to the OpenTelemetry **trace ID**, which is the same across all requests in a trace. Azure SDK for Go has a built-in `requestIdPolicy` that generates a unique UUID per request for this header, but it skips setting it if it's already present.

By setting it first via PerCallPolicies, all requests within a trace will share the same `x-ms-client-request-id`, making it harder for Azure support to distinguish individual requests.

Is this the intended behavior? If the goal is trace correlation, `NewMsCorrelationPolicy()` (`x-ms-correlation-request-id`) might be more appropriate here, leaving `x-ms-client-request-id` to the SDK's built-in per-request UUID.

cli/azd/pkg/azsdk/correlation_policy_test.go
```diff
@@ -68,6 +68,33 @@ func Test_simpleCorrelationPolicy_Do(t *testing.T) {
...
+			),
+			expect:                to.Ptr(""),
+			headerName:            MsClientRequestIdHeader,
+			correlationPolicyFunc: NewMsClientRequestIdPolicy,
+		},
+		{
+			name:                  "WithoutTraceId",
+			ctx:                   context.Background(),
+			expect:                nil,
+			headerName:            MsClientRequestIdHeader,
```
wbreza, line 95: **[🟢 Low]** Duplicate test case names across header types

The test table now has three sets of identically-named cases (`WithTraceId`, `WithInvalidTraceId`, `WithoutTraceId`). Go appends `#01`/`#02` suffixes to disambiguate, but it makes test output harder to read.

Consider prefixing with the header type, e.g. `ClientRequestId_WithTraceId`, `CorrelationId_WithTraceId`, etc.

(This is a pre-existing pattern — not blocking.)

wbreza (commented):
## Code Review — Approve with Suggestions

### Summary

| Priority | Count |
|----------|-------|
| 🟡 Medium | 2 |
| 🟢 Low | 1 |
| **Total** | **3** |

Inline comments below. One additional item not tied to a specific file:

---

**🟠 [High] PR title/description says `x-ms-correlation-request-id` but code adds `x-ms-client-request-id`**

The PR title is *"Add x-ms-correlation-request-id header..."* and the description says *"This PR adds support for x-ms-correlation-request-id header propagation."* However, the code actually adds `NewMsClientRequestIdPolicy()` which sets the **`x-ms-client-request-id`** header — a semantically different header. The testing section of the description correctly mentions `x-ms-client-request-id`, creating an internal contradiction. Consider updating the title and description to match the actual change.

---

### ✅ What Looks Good

- **`azdext.Run` migration is clean** — Replaces 16 lines of boilerplate with a single call, aligning with the `azure.ai.agents` pattern.
- **Test coverage is thorough** — All three scenarios (valid trace ID, invalid trace ID, no trace ID) are tested.
- **New policy follows established pattern** — Reuses `simpleCorrelationPolicy` abstraction, keeping code DRY.

cli/azd/pkg/azsdk/correlation_policy.go
saanikaguptamicrosoft: Hi @wbreza 
Our goal here is to have a request-id at service team's end, so that in future if there's an IcM due to downstream, we can pass them the ticket along with the request-id for them to investigate. We tried testing by using NewMsCorrelationPolicy() instead, but then in Cognitive Services kusto noticed that x-ms-client-request-id was not being passed-
image

So, when we tried passing it explicitly using NewMsClientRequestIdPolicy(), it is getting reflected in Cognitive Services kusto. So, no override is happening currently. Pls let us know your suggestion.

cli/azd/extensions/azure.ai.finetune/internal/providers/factory/provider_factory.go
saanikaguptamicrosoft: Thanks for your comment. Will raise two separate PRs to address this.

saanikaguptamicrosoft: Raised 2 PRs-
1. https://github.com/Azure/azure-dev/pull/7009
2. https://github.com/Azure/azure-dev/pull/7010

cli/azd/pkg/azsdk/correlation_policy_test.go
saanikaguptamicrosoft: Addressed here: https://github.com/Azure/azure-dev/pull/7009

PR #6978, opened 2026-03-04
Emit `codes.Unauthenticated` for auth errors in gRPC server and add safety net in `WrapError`

JeffreyCA:
Contributes to #6976 
Relates to https://github.com/Azure/azure-dev/pull/6979

### Problem

Auth errors (`ErrNoCurrentUser`, `ReLoginRequiredError`) from azd host gRPC calls were returned with `codes.Unknown`, making it difficult for extensions to distinguish auth failures from other errors. Extensions that didn't explicitly handle these errors would report `ext.local.failed` in telemetry instead of `ext.auth.*`.

### Changes

**gRPC server interceptor** (`internal/grpcserver/server.go`):
- `wrapErrorWithSuggestion()` now returns `status.Error(codes.Unauthenticated, ...)` when the underlying error is `ReLoginRequiredError` or `ErrNoCurrentUser`. All extensions benefit automatically.

**`WrapError` safety net** (`pkg/azdext/extension_error.go`):
- Added gRPC `codes.Unauthenticated` detection after the existing `azcore.ResponseError` check. Any unhandled auth error is auto-classified as `LocalError{Category: "auth", Code: "auth_failed"}` during serialization, ensuring correct telemetry even if an extension doesn't explicitly handle auth errors.
- Improved doc comments on `WrapError` and `UnwrapError` clarifying that `WrapError` is called from extensions (serialization) and `UnwrapError` from azd core (deserialization).

### Testing

- Added `wrapErrorWithSuggestion` tests for `ErrNoCurrentUser`, wrapped `ErrNoCurrentUser`, and `ReLoginRequiredError` with suggestion — all verify `codes.Unauthenticated` is returned.
- Added `WrapError` round-trip tests for direct and wrapped gRPC `Unauthenticated` errors — verify `LocalError{Category: "auth"}` classification.

### DIFF
--- a/cli/azd/internal/grpcserver/server.go
+++ b/cli/azd/internal/grpcserver/server.go
@@ -12,6 +12,7 @@ import (
 	"net"
 
 	"github.com/azure/azure-dev/cli/azd/internal"
+	"github.com/azure/azure-dev/cli/azd/pkg/auth"
 	"github.com/azure/azure-dev/cli/azd/pkg/azdext"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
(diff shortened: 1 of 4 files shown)


wbreza (approved):
## Code Review Summary

Clean, well-structured PR that solves a real problem — auth errors were invisible to extensions as codes.Unknown. The three key things I verified:

### Backwards Compatibility
**No breaking changes.** gRPC inherently serializes errors as code+message strings, so the Go error chain was never available to extensions anyway. This change only upgrades codes.Unknown to codes.Unauthenticated for two specific error types, which is purely additive.

### What Looks Good
- **Correct priority order** in WrapError: ServiceError/LocalError > Azure SDK > gRPC Unauthenticated > fallback
- **The added return extErr after Azure SDK block** is a subtle but important fix preventing future safety-net checks from overwriting Azure SDK classification
- **Excellent test coverage** covering direct, wrapped, and combined (suggestion + auth) cases
- **Doc comments are thorough** explaining caller contexts and priority order

### Minor Note
Consider documenting why auth.AuthFailedError is not handled here (likely intentional - auth failed vs auth required). A brief comment would help future maintainers.

Issue #6991, opened 2026-03-04
azd init fail when using existing project-id


ferronsw:
- [x] Make sure you've installed the latest version using [instructions](https://github.com/Azure/azure-dev?tab=readme-ov-file#installupgrade-azure-developer-cli)

**Output from `azd version`**
Run `azd version` and copy and paste the output here:

azd version 1.23.7 (commit 302d696d41e2029a04e86ecf642da5a0c1c441c0)

**Describe the bug**
az init fails when running  `azd ai agent init --project-id "/subscriptions/<subGuid>/resourceGroups/rg-ai-lab-foundry-vnet-integrated-test/providers/Microsoft.CognitiveServices/accounts/<account>/projects/<project>"`

**To Reproduce**
Create a foundry with vnet integration use the terraform sample: https://github.com/microsoft-foundry/foundry-samples/tree/main/infrastructure/infrastructure-setup-terraform/15b-private-network-standard-agent-setup-byovnet

**Expected behavior**
Run azd init successfully without an error.

**Environment**
Information on your environment:
    .Net 9
    VS Code 1.109.5

**ERROR message**
ERROR: failed to create definition from local agent: failed to list Foundry projects: failed to list accounts: unmarshalling type *armcognitiveservices.AccountListResult: unmarshalling type *armcognitiveservices.AccountListResult: struct field Value: unmarshalling type *armcognitiveservices.Account: struct field Properties: unmarshalling type *armcognitiveservices.AccountProperties: struct field NetworkInjections: unmarshalling type *armcognitiveservices.NetworkInjections: json: cannot unmarshal array into Go value of type map[string]json.RawMessage

JeffreyCA:
Thanks for the report, this should be fixed in [0.1.13-preview](https://github.com/Azure/azure-dev/releases/tag/azd-ext-azure-ai-agents_0.1.13-preview).

PR #6992, opened 2026-03-04
Fix issue #6991

ferronsw:
I'm no go expert so please verify if this is a good solution.
It worked for me regarding #6991


rajeshkamal5050:
Fixed as part of https://github.com/Azure/azure-dev/pull/6957 and getting released https://github.com/Azure/azure-dev/pull/6995

Issue #6915, opened 2026-02-27
Skip redundant recursive traversal in provisioning progress polling


spboyer:
## Summary

Skip expensive recursive deployment operation traversal during provisioning progress polling when nothing has changed, while keeping the 3s poll cadence for near-real-time progress.

**Parent issue**: Azure/azure-dev#6886 (Performance Review Audit -- Finding 9)

## Problem

Provisioning progress polls ARM every 3s. Each poll calls `GetDeploymentResourceOperations` which recursively traverses all nested deployments. During the idle tail of long provisioning runs (e.g., waiting for Cosmos DB), most poll cycles return unchanged data but still perform the full recursive traversal.

## Solution

Track root operation count and running-ops state between poll cycles. When the operation count is unchanged and no operations are still running, skip processing. Keeps 3s cadence for responsiveness.

## Impact

- Provision: Baseline 333s vs Optimized 323s (-10s / 3%)
- Reduces redundant ARM API calls during idle tail of provisioning
- Zero behavior change in progress display UX

spboyer:
Closing as resolved based on the merged implementation. PR #6916 closed the original optimization, and PR #7019 followed up with the callback-based walker and the current provisioning progress polling path. The current code in `cli/azd/pkg/infra/provisioning_progress_display.go` now uses `WalkDeploymentOperations` plus cached terminal-operation tracking to skip redundant recursive traversal while keeping progress updates responsive.

PR #7004, opened 2026-03-04
Replace channel-based cancellation with context cancellation in provisioning progress display

weikanglim:
## Summary

Replace the `cancelProgress` channel in `BicepProvider.Deploy` with `context.WithCancel` for cancelling the progress-reporting goroutine.

## Impact

Users will observe a faster progress update when the deployment is completed, but we're still polling for new progress operations (which is very common).

## Changes

- **`bicep_provider.go`**: Use `context.WithCancel` instead of a channel. Use `sync.WaitGroup` to ensure the goroutine fully exits before stopping the spinner. 

## Motivation

The channel-based approach blocks the caller on `defer func() { cancelProgress <- true }()` until the goroutine finishes its current `ReportProgress` cycle. With context cancellation:
- `cancelProgress()` returns immediately (broadcast signal via channel close)
- In-flight API calls abort promptly via context
- `wg.Wait()` ensures the goroutine exits before `StopSpinner` is called, eliminating the race where the goroutine could restart the spinner after it was stopped

Contributes to #6915

### DIFF
--- a/cli/azd/pkg/infra/provisioning/bicep/bicep_provider.go
+++ b/cli/azd/pkg/infra/provisioning/bicep/bicep_provider.go
@@ -18,6 +18,7 @@ import (
 	"slices"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/arm"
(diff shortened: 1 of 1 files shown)


cli/azd/pkg/infra/provisioning/bicep/bicep_provider.go
```diff
@@ -707,36 +708,42 @@ func (p *BicepProvider) Deploy(ctx context.Context) (*provisioning.DeployResult,
...
 			case <-timer.C:
-				if err := progressDisplay.ReportProgress(ctx, &queryStartTime); err != nil {
+				if err := progressDisplay.ReportProgress(progressCtx, &queryStartTime); err != nil {
 					// We don't want to fail the whole deployment if a progress reporting error occurs
 					log.Printf("error while reporting progress: %v", err)
 				}
-
-				timer.Reset(regularDelay)
 			}
 		}
```
wbreza, line 750: **[Critical]** `timer.Reset(delay)` was removed — progress reports only once

After the first `timer.C` fires and `ReportProgress` runs, the timer is exhausted. The goroutine blocks in the `select` on `timer.C` (which never fires again), only exiting when `progressCtx.Done()` fires. This means progress is reported **exactly once at 3 seconds**, then nothing until deployment completes.

For long deployments (often minutes), this is a significant UX regression — users would see one progress update and then a frozen display.

```suggestion
if err := progressDisplay.ReportProgress(progressCtx, &queryStartTime); err != nil {
// We don't want to fail the whole deployment if a progress reporting error occurs
log.Printf("error while reporting progress: %v", err)
}
timer.Reset(delay)
```

cli/azd/pkg/infra/provisioning/bicep/bicep_provider.go
```diff
@@ -707,36 +708,42 @@ func (p *BicepProvider) Deploy(ctx context.Context) (*provisioning.DeployResult,
...
-	defer func() { cancelProgress <- true }()
+	progressCtx, cancelProgress := context.WithCancel(ctx)
+	var wg sync.WaitGroup
+	wg.Add(1)
+	defer func() {
+		cancelProgress()
+		// Wait for the progress reporting goroutine to exit to guarantee stopping the spinner
+		wg.Wait()
+		// Clean up the spinner fully
+		p.console.StopSpinner(ctx, "", input.StepDone)
```
wbreza, line 719: **[Medium]** Double StopSpinner — defer here + Manager cleanup

This defer calls `StopSpinner(ctx, "", input.StepDone)`. After `Deploy` returns, `provisioning.Manager.Deploy()` also calls `StopSpinner` at `manager.go:128`. Consider whether this is intentional (to prevent the goroutine from restarting the spinner in the race window) — if so, a brief comment explaining the rationale would help future readers.

wbreza (commented):
## Code Review: PR #7004 — Replace channel-based cancellation with context cancellation

### Downstream Impact Assessment
- **Isolated change** — Only BicepProvider uses progress reporting; other providers (Terraform, DevCenter) are unaffected
- **Manager integration** — `provisioning.Manager.Deploy()` wraps this with its own StopSpinner at `manager.go:128`, which continues to work correctly
- **No test coverage** for the progress goroutine lifecycle, so the timer issue below wouldn't be caught by CI

### Findings Summary

| Priority | Count |
|----------|-------|
| 🔴 Critical | 1 |
| 🟡 Medium | 1 |
| 🟢 Low | 1 |
| **Total** | **3** |

### ✅ What Looks Good
- `context.WithCancel` is the correct Go pattern for goroutine cancellation — much cleaner than the channel approach
- `sync.WaitGroup` ensures the goroutine exits before spinner stops, eliminating the original race condition
- Passing `progressCtx` to `ReportProgress` means in-flight API calls abort promptly on cancellation
- Change is well-scoped and isolated

**Overall Assessment:** The pattern change is sound, but the missing `timer.Reset` breaks continuous progress reporting.

cli/azd/pkg/infra/provisioning/bicep/bicep_provider.go
weikanglim: Forgot to push

cli/azd/pkg/infra/provisioning/bicep/bicep_provider.go
weikanglim: We want to be a good citizen cleaning up after we're done. `Deploy()` may call `StopSpinner`, but that doesn't mean we shouldn't clean up after ourselves here.

Issue #6982, opened 2026-03-04
azd update: rollback/recovery on failed update


rajeshkamal5050:
## Problem
No backup of current binary before replacement. A broken update via `azd update` or auto-apply requires manual reinstall.

## Proposed
- Back up the current binary before replacement (e.g. `azd.bak`)
- On failure, restore from backup automatically
- For auto-apply (`ApplyStagedUpdate`), use atomic copy-to-temp + rename instead of direct overwrite

## Context
Raised in design review #6910 and PR #6942 code review.

Parent epic: #6721

PR #7012, opened 2026-03-05
[bug-fix] azd update: fix SIGKILL during binary replacement on macOS

rajeshkamal5050:
`azd update` on macOS crashes with `zsh: killed` and leaves the binary broken when azd lives on a different filesystem than `/tmp` (e.g. `/usr/local/bin`).

`os.Rename` fails cross-device, falls through to `copyFile`, which truncates the running binary in place. macOS kills the process when memory-mapped pages go invalid.

**Fix:** on unix, remove the old file before creating the new one. The running process keeps its fd to the old inode, new file gets a fresh inode. This replaces the previous Linux-only ETXTBSY workaround with a single approach that works on both macOS and Linux.

Our E2E tests missed this because test binaries were on the same filesystem as `/tmp`, so `os.Rename` always succeeded.

**Test:** `TestReplaceBinary_CreatesNewInode` — forces the `copyFile` path, verifies the inode changes and old fd still reads old content. Passes on macOS and Linux.

Related: #6721, #6982

### DIFF
--- a/cli/azd/pkg/update/manager.go
+++ b/cli/azd/pkg/update/manager.go
@@ -619,11 +619,7 @@ func copyFile(src, dst string) error {
 	if runtime.GOOS != "windows" {
 		removeErr := os.Remove(dst)
 		if removeErr != nil && !os.IsNotExist(removeErr) {
-			if os.IsPermission(removeErr) {
-				return removeErr
-			}
-			// Unexpected error removing dst — log but continue to os.Create fallback
-			log.Printf("warning: could not remove %s before replacement: %v", dst, removeErr)
+			return fmt.Errorf("removing %s before replacement: %w", dst, removeErr)
 		}
 	}
 
--- a/cli/azd/pkg/update/manager_unix_test.go
+++ b/cli/azd/pkg/update/manager_unix_test.go
@@ -7,6 +7,7 @@ package update
 
 import (
 	"context"
+	"io"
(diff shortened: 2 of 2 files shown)


cli/azd/pkg/update/manager.go
```diff
@@ -611,19 +611,19 @@ func copyFile(src, dst string) error {
...
-			// Can't remove the old file due to directory permissions.
-			// Return the permission error so callers can handle elevation.
-			err = removeErr
+	// On unix, always remove the destination file before creating a new one.
+	// Truncating a running binary in place (os.Create on existing file) corrupts
+	// memory-mapped code pages and causes macOS to SIGKILL the process.
+	// Removing first creates a new inode — the OS keeps the old inode alive
+	// for the running process via its fd.
+	if runtime.GOOS != "windows" {
+		removeErr := os.Remove(dst)
```
wbreza, line 620: **[Medium]** Permission error wrapping change

The old code preserved raw permission errors (`err = removeErr`) so callers could detect them via `os.IsPermission(err)` and prompt for elevation. The new code wraps with `fmt.Errorf("removing %s before replacement: %w", dst, removeErr)`. While `%w` preserves the error chain for `errors.Is`, it's worth confirming that callers of `replaceBinary` use `errors.Is(err, os.ErrPermission)` rather than string matching for elevation prompts.

PR #7015, opened 2026-03-05
[Agents Extension] Add env var check for vnext experience

trangevi:
Hosted agent backend is switch to use ADC as an underlying platform. This will allow for testing that change without disrupting current production usage.

### DIFF
--- a/cli/azd/extensions/azure.ai.agents/internal/project/service_target_agent_test.go
+++ b/cli/azd/extensions/azure.ai.agents/internal/project/service_target_agent_test.go
@@ -4,7 +4,6 @@
 package project
 
 import (
-	"os"
 	"testing"
 
 	"azureaiagent/internal/pkg/agents/agent_api"


rajeshkamal5050 (approved):

cli/azd/extensions/azure.ai.agents/internal/project/service_target_agent_test.go
```diff
@@ -0,0 +1,111 @@
...
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			// Set/unset OS env
+			if tt.osEnvValue != "" {
+				t.Setenv("enableHostedAgentVNext", tt.osEnvValue)
+			} else {
+				os.Unsetenv("enableHostedAgentVNext")
```
wbreza: **[🟡 Medium]** `os.Unsetenv` doesn't restore the original value after the test

`os.Unsetenv("enableHostedAgentVNext")` permanently clears the var without restoring it after the test. `t.Setenv` (used in the `if` branch) auto-restores on cleanup, but `os.Unsetenv` does not. In CI where this env var might be set, it could affect other tests.

Since the code uses `os.Getenv` (returns `""` for both unset and empty), consider:
```suggestion
						t.Setenv("enableHostedAgentVNext", "")
```

cli/azd/extensions/azure.ai.agents/internal/project/service_target_agent.go
```diff
@@ -667,6 +668,9 @@ func (p *AgentServiceTargetProvider) deployHostedAgent(
 		)
 	}
 
+	// Check enableHostedAgentVNext from azd env first, then OS env
```
wbreza, line 671: **[🟡 Medium]** Missing PR description

The PR body is empty. Consider adding a brief description explaining what the "vnext experience" is, when `enableHostedAgentVNext` should be set, and what the `enableVnextExperience` metadata signals to the service. This helps reviewers and future maintainers understand the purpose.

PR #7016, opened 2026-03-05
feat: add sensible defaults for azd env new / init in --no-prompt mode

### LINKED ISSUE #6817: azd ai agent init: Default environment name to directory name

### LINKED ISSUE #6818: azd ai agent init: Auto-select subscription in non-interactive mode

### LINKED ISSUE #6934: feat: add sensible defaults for azd env new / init in --no-prompt mode
## Feature Request

`azd env new` and `azd init` should handle `--no-prompt` mode more gracefully by providing sensible defaults instead of failing when optional arguments are omitted.

### Current behavior
- `azd env new --no-prompt` without `<name>` argument: prompts interactively (fails in CI/non-interactive)
- `azd init --no-prompt --template <t>` without `--environment`: returns error requiring `--environment`
- No subscription auto-selection from `defaults.subscription` config or single-subscription accounts

### Proposed behavior

**1. Auto-generate environment name from working directory**
When `--no-prompt` is passed without an explicit environment name, default to `filepath.Base(cwd)` (sanitized for valid env name characters), matching how `azd init` already generates name suggestions interactively.

**2. Auto-select subscription**
When `--no-prompt` is passed without `--subscription`:
- First check `azd config` for `defaults.subscription`
- If not set and only one subscription exists, auto-select it
- If multiple subscriptions and no default, return clear error with guidance

### Motivation
Extensions (e.g. `azure.ai.agents`) need non-interactive init flows for CI/CD. Currently each extension must implement its own `--no-prompt` logic for env naming and subscription selection, which is inconsistent and duplicative. This should be a core capability.

See discussion in #6823 (PR feedback from @trangevi, @vhvb1989, @kristenwomack).

spboyer:
## Problem

`azd env new --no-prompt` and `azd init --no-prompt --template` fail when optional arguments (environment name, subscription) are omitted. Extensions like `azure.ai.agents` need non-interactive init flows for CI/CD but currently must implement their own `--no-prompt` logic. PR #6823 was closed because the team agreed this behavior belongs in azd core, not individual extensions.

### 1. Auto-generate environment name from working directory

When `--no-prompt` is active and no environment name is provided, the name defaults to `filepath.Base(cwd)` sanitized via the existing `CleanName()` function. This is handled centrally in `ensureValidEnvironmentName()` so both `azd env new` and `azd init` benefit.

**Before:** `azd env new --no-prompt` → prompts interactively (fails in CI)
**After:** `azd env new --no-prompt` → creates environment named after the working directory

### 2. Auto-select subscription in non-interactive mode

When `--no-prompt` is active and no subscription is configured:
1. Checks `defaults.subscription` from user config (existing behavior preserved)
2. **New:** If not set and only one subscription exists, auto-selects it
3. If multiple subscriptions and no default → clear error with remediation guidance

**Before:** Immediately errors asking to set `defaults.subscription`
**After:** Auto-selects when unambiguous (1 subscription)

### 3. Remove hard `--environment` requirement in `azd init`

The fail-fast error `"--environment is required when running in non-interactive mode"` is removed. The environment manager now auto-generates the name, so this guard is no longer needed.

### 4. Handle default env confirmation in `--no-prompt` mode

When creating a new environment in `--no-prompt` mode with multiple environments already present, the `Confirm()` prompt for setting it as default is skipped — the new environment is auto-set as default.

## Testing

- **17 new test cases** across 3 test files:
  - `manager_test.go`: Auto-name generation from cwd, CleanName edge cases (7 cases)
  - `prompt_service_test.go`: Subscription auto-select — 0/1/many subs, default set, default not found (5 cases)
  - `init_test.go`: Updated to verify the old hard error no longer occurs (5 cases)
- All existing tests pass (`go test ./pkg/environment/... ./pkg/prompt/... ./cmd/... -short`)

## Files Changed

| File | Change |
|------|--------|
| `pkg/environment/manager.go` | Add no-prompt auto-name guard in `ensureValidEnvironmentName()` |
| `pkg/environment/manager_test.go` | Tests for auto-name generation and CleanName edge cases |
| `cmd/init.go` | Remove hard `--environment` requirement in no-prompt+template mode |
| `cmd/init_test.go` | Update test for removed error |
| `pkg/prompt/prompt_service.go` | Add single-subscription auto-select, add `console` for messaging |
| `pkg/prompt/prompt_service_test.go` | Tests for subscription auto-select (in-memory config manager) |
| `cmd/env.go` | Auto-set default env in no-prompt mode (skip Confirm) |

## Related

- Closes #6934
- Closes #6817
- Closes #6818
- Supersedes #6823 (closed — team agreed logic belongs in core)
- See discussion in #6823 (feedback from @trangevi, @vhvb1989, @kristenwomack)

### DIFF
--- a/cli/azd/cmd/init_test.go
+++ b/cli/azd/cmd/init_test.go
@@ -13,6 +13,8 @@ import (
 	"testing"
 
 	"github.com/azure/azure-dev/cli/azd/internal"
+	"github.com/azure/azure-dev/cli/azd/pkg/alpha"
+	"github.com/azure/azure-dev/cli/azd/pkg/config"
 	"github.com/azure/azure-dev/cli/azd/pkg/environment/azdcontext"
 	"github.com/azure/azure-dev/cli/azd/pkg/exec"
 	"github.com/azure/azure-dev/cli/azd/pkg/lazy"
@@ -47,10 +49,11 @@ func setupInitAction(t *testing.T, mockContext *mocks.MockContext, flags *initFl
 		lazyAzdCtx: lazy.NewLazy(func() (*azdcontext.AzdContext, error) {
 			return azdcontext.NewAzdContextWithDirectory(tmpDir), nil
 		}),
-		console: mockContext.Console,
-		cmdRun:  mockContext.CommandRunner,
-		gitCli:  gitCli,
-		flags:   flags,
+		console:         mockContext.Console,
+		cmdRun:          mockContext.CommandRunner,
+		gitCli:          gitCli,
+		flags:           flags,
+		featuresManager: alpha.NewFeaturesManagerWithConfig(config.NewEmptyConfig()),
 	}
 }
 
@@ -136,6 +139,7 @@ func TestInitNoPromptRequiresMode(t *testing.T) {
 func TestInitFailFastMissingEnvNonInteractive(t *testing.T) {
 	t.Run("NoLongerFailsWhenNoPromptWithTemplateAndNoEnv", func(t *testing.T) {
 		mockContext := mocks.NewMockContext(context.Background())
+		mockContext.Console.SetNoPromptMode(true)
 
 		flags := &initFlags{
 			templatePath: "owner/repo",
@@ -145,9 +149,9 @@ func TestInitFailFastMissingEnvNonInteractive(t *testing.T) {
 		action := setupInitAction(t, mockContext, flags)
 
 		// With sensible defaults, --no-prompt --template without --environment should not
-		// fail with the old "--environment is required" error. The action may panic or
-		// error later due to missing mocks (e.g., featureManager), which is expected and
-		// proves we got past the former fail-fast check.
+		// fail with the old "--environment is required" error. The action will error or
+		// panic later due to missing mocks for template download, which is expected —
+		// we only verify the fail-fast guard was removed.
 		err := runActionSafe(*mockContext.Context, action)
 		if err != nil {
 			require.NotContains(t, err.Error(),
--- a/cli/azd/pkg/environment/manager.go
+++ b/cli/azd/pkg/environment/manager.go
@@ -580,6 +580,12 @@ func (m *manager) ensureValidEnvironmentName(ctx context.Context, spec *Spec) er
 			cleaned = cleaned[:EnvironmentNameMaxLength]
 		}
 
+		if !IsValidEnvironmentName(cleaned) {
+			return fmt.Errorf(
+				"auto-generated environment name '%s' from directory '%s' is invalid. "+
+					"Specify one explicitly with -e or as an argument", cleaned, dirName)
+		}
+
 		spec.Name = cleaned
 		m.console.Message(ctx, fmt.Sprintf("Using environment name: %s", spec.Name))
 
--- a/cli/azd/pkg/environment/manager_test.go
+++ b/cli/azd/pkg/environment/manager_test.go
@@ -7,6 +7,8 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"os"
+	"path/filepath"
 	"strings"
 	"testing"
 
@@ -724,13 +726,57 @@ func Test_EnvManager_Create_NoPrompt_AutoName(t *testing.T) {
 		mockContext := mocks.NewMockContext(context.Background())
 		mockContext.Console.SetNoPromptMode(true)
 
-		envManager := createEnvManagerForManagerTest(t, mockContext)
+		// Use a known directory name so we can assert the generated env name
+		tmpDir := t.TempDir()
+		knownDir := filepath.Join(tmpDir, "my-cool-project")
+		require.NoError(t, os.Mkdir(knownDir, 0755))
+
+		azdCtx := azdcontext.NewAzdContextWithDirectory(knownDir)
+		localDataStore := NewLocalFileDataStore(azdCtx, config.NewFileConfigManager(config.NewManager()))
+		envManager := newManagerForTest(azdCtx, mockContext.Console, localDataStore, nil)
+
 		env, err := envManager.Create(*mockContext.Context, Spec{Name: ""})
 		require.NoError(t, err)
 		require.NotNil(t, env)
-		// Name should be non-empty and valid
-		require.NotEmpty(t, env.Name())
-		require.True(t, IsValidEnvironmentName(env.Name()))
+		require.Equal(t, "my-cool-project", env.Name())
+	})
+
+	t.Run("cleans special characters in directory name", func(t *testing.T) {
+		mockContext := mocks.NewMockContext(context.Background())
+		mockContext.Console.SetNoPromptMode(true)
+
+		tmpDir := t.TempDir()
+		knownDir := filepath.Join(tmpDir, "my cool project!")
+		require.NoError(t, os.Mkdir(knownDir, 0755))
+
+		azdCtx := azdcontext.NewAzdContextWithDirectory(knownDir)
+		localDataStore := NewLocalFileDataStore(azdCtx, config.NewFileConfigManager(config.NewManager()))
+		envManager := newManagerForTest(azdCtx, mockContext.Console, localDataStore, nil)
+
+		env, err := envManager.Create(*mockContext.Context, Spec{Name: ""})
+		require.NoError(t, err)
+		require.NotNil(t, env)
+		require.Equal(t, "my-cool-project-", env.Name())
+	})
+
+	t.Run("truncates long directory names", func(t *testing.T) {
+		mockContext := mocks.NewMockContext(context.Background())
+		mockContext.Console.SetNoPromptMode(true)
+
+		tmpDir := t.TempDir()
+		longName := strings.Repeat("a", 100)
+		knownDir := filepath.Join(tmpDir, longName)
+		require.NoError(t, os.Mkdir(knownDir, 0755))
+
+		azdCtx := azdcontext.NewAzdContextWithDirectory(knownDir)
+		localDataStore := NewLocalFileDataStore(azdCtx, config.NewFileConfigManager(config.NewManager()))
+		envManager := newManagerForTest(azdCtx, mockContext.Console, localDataStore, nil)
+
+		env, err := envManager.Create(*mockContext.Context, Spec{Name: ""})
+		require.NoError(t, err)
+		require.NotNil(t, env)
+		require.LessOrEqual(t, len(env.Name()), EnvironmentNameMaxLength)
+		require.Equal(t, strings.Repeat("a", EnvironmentNameMaxLength), env.Name())
 	})
 
 	t.Run("uses provided name even in no-prompt mode", func(t *testing.T) {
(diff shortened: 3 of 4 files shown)


kristenwomack (approved):

cli/azd/pkg/environment/manager.go
```diff
@@ -545,7 +547,45 @@ func (m *manager) Delete(ctx context.Context, name string) error {
...
+			return fmt.Errorf(
+				"cannot generate valid environment name from directory '%s'. "+
+					"Specify one explicitly with -e or as an argument", dirName)
+		}
+
+		if len(cleaned) > EnvironmentNameMaxLength {
+			cleaned = cleaned[:EnvironmentNameMaxLength]
+		}
+
+		spec.Name = cleaned
```
wbreza, line 589: **[🟠 High]** Auto-generated name bypasses `IsValidEnvironmentName` check

The no-prompt auto-generation block manually checks for degenerate values (`""`, `"-"`, `"."`, `".."`) and does an early `return nil`, bypassing the `for !IsValidEnvironmentName(spec.Name)` loop that gates all other code paths. If `CleanName` or the regex ever drift from each other, invalid names could slip through. Names like `"---"`, `"()"`, or `"_"` would pass the manual check but may be undesirable.

Consider adding an explicit validation after truncation:

```go
if !IsValidEnvironmentName(cleaned) {
    return fmt.Errorf(
        "auto-generated environment name '%s' from directory '%s' is invalid. "+
            "Specify one explicitly with -e or as an argument", cleaned, dirName)
}
```

cli/azd/cmd/init_test.go
```diff
@@ -144,11 +144,15 @@ func TestInitFailFastMissingEnvNonInteractive(t *testing.T) {
...
-		result, err := action.Run(*mockContext.Context)
-		require.Error(t, err)
-		require.Contains(t, err.Error(),
-			"--environment is required when running in non-interactive mode")
-		require.Nil(t, result)
+		// With sensible defaults, --no-prompt --template without --environment should not
+		// fail with the old "--environment is required" error. The action may panic or
+		// error later due to missing mocks (e.g., featureManager), which is expected and
+		// proves we got past the former fail-fast check.
+		err := runActionSafe(*mockContext.Context, action)
```
wbreza, line 155: **[🟠 High]** `runActionSafe` test pattern masks real failures

The `if err != nil { NotContains }` pattern passes in three unrelated situations: `err==nil`, panic-wrapped error, or any other error. A nil-pointer panic in production code would silently pass via the panic recovery path. The test only proves the old error message was *not* returned — not that the new behavior works.

Consider providing enough mocks so the action can proceed further without panicking, then asserting the positive outcome. Alternatively, at minimum assert `require.Error(t, err)` and check the specific expected failure mode rather than swallowing panics.

cli/azd/pkg/environment/manager_test.go
```diff
@@ -718,3 +718,53 @@ func Test_EnvManager_InstanceCaching(t *testing.T) {
...
+	t.Run("generates name from working directory", func(t *testing.T) {
+		mockContext := mocks.NewMockContext(context.Background())
+		mockContext.Console.SetNoPromptMode(true)
+
+		envManager := createEnvManagerForManagerTest(t, mockContext)
+		env, err := envManager.Create(*mockContext.Context, Spec{Name: ""})
+		require.NoError(t, err)
+		require.NotNil(t, env)
+		// Name should be non-empty and valid
+		require.NotEmpty(t, env.Name())
```
wbreza: **[🟠 High]** Auto-name test doesn't verify the generated name

The test asserts `NotEmpty` + `IsValidEnvironmentName` but never checks the name relates to the working directory. Any hardcoded fallback string would pass. The production code has multiple branches (azdContext → os.Getwd → CleanName → truncation) — none are individually verified.

Consider setting up a known temp directory with a controlled name, then asserting `require.Equal(t, expectedCleanedName, env.Name())`. Add sub-tests for truncation and special-char cleaning.

cli/azd/cmd/env.go
```diff
@@ -975,6 +975,14 @@ func (en *envNewAction) Run(ctx context.Context) (*actions.ActionResult, error)
 		en.console.Message(ctx,
 			fmt.Sprintf("New environment '%s' was set as default", env.Name()),
 		)
+	} else if en.console.IsNoPromptMode() {
+		// In --no-prompt mode, auto-set the new environment as default
+		if err := en.azdCtx.SetProjectState(azdcontext.ProjectState{DefaultEnvironment: env.Name()}); err != nil {
+			return nil, fmt.Errorf("saving default environment: %w", err)
+		}
+		en.console.Message(ctx,
+			fmt.Sprintf("New environment '%s' created and set as default", env.Name()),
```
wbreza, line 984: **[🟡 Medium]** No tests for auto-set default in no-prompt mode

This new branch that auto-sets the default environment in `--no-prompt` mode has no corresponding test. It might be worth adding a test verifying that after `envNewAction.Run` in no-prompt mode with existing environments, the new environment is set as default.

cli/azd/pkg/prompt/prompt_service.go
```diff
@@ -221,28 +225,45 @@ func (ps *promptService) PromptSubscription(
...
+		switch len(subscriptionList) {
+		case 0:
+			return nil, fmt.Errorf(
+				"no Azure subscriptions found for the current account. " +
+					"Verify that you're logged into the correct Azure account and tenant, " +
+					"and that your account has one or more active subscriptions. " +
+					"If needed, run 'azd auth login' to sign in.")
+		case 1:
+			ps.console.Message(ctx, fmt.Sprintf(
+				"Auto-selected subscription: %s (%s)",
```
wbreza, line 259: **[🟡 Medium]** `console.Message` in no-prompt auto-select paths

Both auto-select paths (here and in `manager.go`) emit `console.Message` during `--no-prompt` mode. In `--no-prompt` + `--output json` mode, `console.Message` injects informational JSON events into the structured output stream. CI/CD consumers parsing azd JSON output may be surprised by the new events.

This is consistent with existing `console.Message` usage elsewhere in no-prompt paths (e.g., the existing "New environment was set as default" message), so it follows established patterns — just worth being aware of.

cli/azd/pkg/environment/manager_test.go
```diff
@@ -718,3 +718,53 @@ func Test_EnvManager_InstanceCaching(t *testing.T) {
...
+		expected string
+	}{
+		{"simple name", "my-project", "my-project"},
+		{"spaces replaced", "my project", "my-project"},
+		{"special chars replaced", "my@project!v2", "my-project-v2"},
+		{"dots preserved", "my.project", "my.project"},
+		{"underscores preserved", "my_project", "my_project"},
+		{"unicode replaced", "projeçt-naïve", "proje-t-na-ve"},
+		// Note: parentheses are valid in env names per EnvironmentNameRegexp: [a-zA-Z0-9-\(\)_\.]
+		{"parens preserved", "my(project)", "my(project)"},
```
wbreza, line 807: **[🟢 Low]** Consider adding edge case entries to the CleanName table: empty string, all-special-chars (`"@#$"` → `"---"`), and names exceeding 64 chars (to show CleanName itself doesn't truncate — the caller does).

cli/azd/pkg/environment/manager.go
rajeshkamal5050: Addressed — added `IsValidEnvironmentName(cleaned)` check after truncation in `manager.go`. Also verified that `CleanName` output always produces valid chars per the regex, so this guard is purely defensive but good to have if the two functions ever drift.

cli/azd/cmd/init_test.go
rajeshkamal5050: Addressed — added `featuresManager` mock and `SetNoPromptMode(true)` to `setupInitAction` so the action proceeds further without nil panics. The test still uses `runActionSafe` because fully mocking the template download chain is impractical here, but the panic surface is reduced. The comment now clarifies the intent: we only verify the old fail-fast guard was removed.

cli/azd/pkg/environment/manager_test.go
rajeshkamal5050: Addressed — tests now create controlled temp directories with known names (`my-cool-project`, `my cool project!`) and assert exact `require.Equal` on the generated name. Added sub-tests for truncation (100-char name → 64) and special-char cleaning.

cli/azd/cmd/env.go
rajeshkamal5050: Addressed — added `env_new_test.go` with `TestEnvNewAction_NoPrompt_AutoSetsDefault` that verifies the new environment is auto-set as default when multiple environments exist in `--no-prompt` mode.

cli/azd/pkg/prompt/prompt_service.go
rajeshkamal5050: Acknowledged — this is consistent with existing `console.Message` usage in no-prompt paths (e.g. the "was set as default" message). No change needed but good to be aware of for JSON output consumers.

cli/azd/pkg/environment/manager_test.go
rajeshkamal5050: Addressed — added three edge cases to the CleanName table: empty string (`""` → `""`), all-special-chars (`"@#$"` → `"---"`), and long name (`100 chars → 100 chars`, confirming CleanName itself does not truncate).

PR #7026, opened 2026-03-06
[Agents Extension] Add support for run and invoke commands

trangevi:
Adds two new commands to the agent extension

1. `run` starts the agent on the local machine. If there is more than one agent in the current azd project, a name will have to be specified.
2. `invoke` executes a /responses call against an agent, either running locally or in the cloud

### DIFF
--- a/cli/azd/extensions/azure.ai.agents/internal/cmd/helpers.go
+++ b/cli/azd/extensions/azure.ai.agents/internal/cmd/helpers.go
@@ -140,7 +140,7 @@ func resolveConversationID(
 	// Create and persist a new conversation for multi-turn memory.
 	newConvID, err := createConversation(ctx, endpoint, bearerToken)
 	if err != nil {
-		return "", fmt.Errorf("Warning: could not create conversation; multi-turn memory disabled (%v)", err)
+		return "", fmt.Errorf("failed to create conversation: %w", err)
 	}
 
 	agentCtx.Conversations[agentName] = newConvID
@@ -149,23 +149,6 @@ func resolveConversationID(
 	}
 
 	return newConvID, nil
-}
-
-// saveConversationID persists a conversation ID for an agent.
-func saveConversationID(ctx context.Context, azdClient *azdext.AzdClient, agentName, convID string) error {
-	configPath, err := resolveConfigPath(ctx, azdClient)
-	if err != nil {
-		return err
-	}
-	agentCtx := loadLocalContext(configPath)
-	if agentCtx.Conversations == nil {
-		agentCtx.Conversations = make(map[string]string)
-	}
-	agentCtx.Conversations[agentName] = convID
-	if err := saveLocalContext(agentCtx, configPath); err != nil {
-		return fmt.Errorf("failed to save conversation ID: %w", err)
-	}
-	return nil
 }
 
 // detectProjectType detects the project type and suggests a start command.


cli/azd/extensions/azure.ai.agents/internal/cmd/run.go
```diff
@@ -0,0 +1,369 @@
...
+	if err := proc.Start(); err != nil {
+		return fmt.Errorf("failed to start agent: %w", err)
+	}
+
+	// Handle Ctrl+C: forward signal to child, then wait for it to exit
+	sigCh := make(chan os.Signal, 1)
+	signal.Notify(sigCh, os.Interrupt)
+	go func() {
+		<-sigCh
+		fmt.Println("\nStopping agent...")
```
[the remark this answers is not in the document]
trangevi: @copilot open a new pull request to apply changes based on [this feedback](https://github.com/Azure/azure-dev/pull/7026#discussion_r2897214034)

cli/azd/extensions/azure.ai.agents/internal/cmd/helpers.go
```diff
@@ -0,0 +1,398 @@
...
+		}
+	}
+
+	return &ServiceRunContext{
+		ProjectDir:     projectDir,
+		StartupCommand: startupCmd,
+	}, nil
+}
+
+// toServiceKey converts a service name into the env var key format (uppercase, underscores).
```
wbreza, line 376: **[🟠 High]** Multi-service resolution should prompt, not error

When multiple `azure.ai.agent` services exist and no `--name` is given, this returns a hard error listing the names. Consider using `azdext.SelectOptions` to prompt the user to pick — consistent with how `init` handles resource selection. This provides a better UX and sets a consistent pattern for all future commands that need service scoping.

Additionally, the `run` command would benefit from accepting an optional positional argument (`azd ai agent run [name]`) so users don't always need the `--name` flag for the common case.

cli/azd/extensions/azure.ai.agents/internal/cmd/init.go
```diff
@@ -1500,6 +1515,34 @@ func (a *InitAction) addToProject(ctx context.Context, targetDir string, agentMa
...
+		}
+		return "", fmt.Errorf("prompting for startup command: %w", err)
+	}
+
+	return strings.TrimSpace(resp.Value), nil
+}
+
 func (a *InitAction) populateContainerSettings(ctx context.Context) (*project.ContainerSettings, error) {
 	if a.flags.NoPrompt {
 		fmt.Printf("No prompt mode enabled, using default container settings\n")
```
wbreza, line 1512: **[🟠 High]** DRY violation — `resolveStartupCommandForInit` duplicated

This method is copy-pasted identically between `InitAction` (here) and `InitFromCodeAction` (init_from_code.go). Consider extracting into a standalone function that accepts the `azdClient` and `projectConfig` as parameters to avoid divergence.

cli/azd/extensions/azure.ai.agents/internal/cmd/init.go
```diff
@@ -1500,6 +1515,34 @@ func (a *InitAction) addToProject(ctx context.Context, targetDir string, agentMa
...
+
+	if cmd := detectStartupCommand(absDir); cmd != "" {
+		return cmd, nil
+	}
+
+	resp, err := a.azdClient.Prompt().Prompt(ctx, &azdext.PromptRequest{
+		Options: &azdext.PromptOptions{
+			Message:        "Enter the command to start your agent (e.g., python main.py), or leave blank to skip",
+			IgnoreHintKeys: true,
+		},
```
wbreza: **[🟠 High]** `--no-prompt` mode not respected

`resolveStartupCommandForInit` unconditionally prompts when auto-detection fails, but `InitAction` supports `--no-prompt` mode (`a.flags.NoPrompt`). In CI/agent environments, this will hang. Consider checking `a.flags.NoPrompt` and returning `""` or a clear error instead of prompting.

cli/azd/extensions/azure.ai.agents/internal/cmd/run.go
```diff
@@ -0,0 +1,369 @@
...
```
wbreza: **[🟠 High]** Signal handler goroutine leak

`signal.Notify(sigCh, os.Interrupt)` is never cleaned up with `signal.Stop(sigCh)`. The goroutine reading from `sigCh` blocks forever if the process exits normally. Consider adding `defer signal.Stop(sigCh)` and using a done channel so the goroutine exits when `proc.Wait()` returns.

cli/azd/extensions/azure.ai.agents/internal/cmd/invoke.go
```diff
@@ -0,0 +1,460 @@
...
+}
+
+func (a *InvokeAction) invokeLocal() error {
+	port := a.flags.port
+	msg := a.flags.message
+
+	fmt.Printf("Target:  localhost:%d (local)\n", port)
+	fmt.Printf("Message: %q\n\n", msg)
+
+	body := map[string]interface{}{
```
wbreza, line 120: **[🟡 Medium]** `invokeLocal` drops parent context

`invokeLocal()` creates `context.Background()` instead of accepting the parent context. This means Ctrl+C / command cancellation won't abort the HTTP request. Consider changing the signature to `invokeLocal(ctx context.Context) error`.

cli/azd/extensions/azure.ai.agents/internal/cmd/helpers.go
```diff
@@ -0,0 +1,398 @@
...
+	}
+	if !forceNew {
+		if convID, ok := ctx.Conversations[agentName]; ok {
+			return convID
+		}
+	}
+	// Conversation creation requires an API call — handled by the invoke command.
+	return ""
+}
+
```
wbreza, line 153: **[🟡 Medium]** `saveLocalContext` errors silently swallowed

Both `resolveSessionID` and `saveConversationID` discard `saveLocalContext` errors. If the write fails, session/conversation state is lost silently. Consider at minimum logging a warning to stderr.

cli/azd/extensions/azure.ai.agents/internal/cmd/run.go
```diff
@@ -0,0 +1,369 @@
...
+					"  - Python (pyproject.toml or requirements.txt with main.py)\n"+
+					"  - .NET (*.csproj)\n"+
+					"  - Node.js (package.json)\n\n"+
+					"Use --start-command to specify explicitly, or set startupCommand in azure.yaml",
+				projectDir,
+			)
+		}
+	} else {
+		fmt.Printf("Using startup command from azure.yaml: %s\n", startCmd)
+	}
```
wbreza, line 120: **[🟡 Medium]** Misleading startup command log

This message says *""Using startup command from azure.yaml""* but `startCmd` may have come from the `--start-command` flag. Consider tracking the source and printing accurately.

cli/azd/extensions/azure.ai.agents/internal/cmd/helpers.go
```diff
@@ -0,0 +1,398 @@
...
+type AgentLocalContext struct {
+	AgentName     string            `json:"agent_name,omitempty"`
+	Sessions      map[string]string `json:"sessions,omitempty"`
+	Conversations map[string]string `json:"conversations,omitempty"`
+}
+
+// loadLocalContext reads the .foundry-agent.json state file from the project root.
+func loadLocalContext() *AgentLocalContext {
+	data, err := os.ReadFile(ConfigFile)
+	if err != nil {
```
wbreza, line 63: **[🟡 Medium]** `.foundry-agent.json` written relative to CWD

`loadLocalContext`/`saveLocalContext` operate on `ConfigFile` relative to the current working directory. Running the command from a subdirectory will create orphaned state files. Consider resolving the project root path and joining it with `ConfigFile`.

cli/azd/extensions/azure.ai.agents/internal/cmd/helpers.go
```diff
@@ -0,0 +1,398 @@
...
+
+	// Find the matching azure.ai.agent service
+	var svc *azdext.ServiceConfig
+	for _, s := range projectResponse.Project.Services {
+		if s.Host != AiAgentHost {
+			continue
+		}
+		if name != "" && s.Name != name {
+			continue
+		}
```
wbreza, line 290: **[🟡 Medium]** Multiple redundant gRPC client connections

`resolveAgentServiceFromProject`, `resolveStartupCommandFromService`, `resolveServiceRunContext`, and `loadAzdEnvironment` each create their own `azdext.NewAzdClient()`. A single command invocation may create 2-3 separate connections. Consider accepting a shared client as a parameter.

cli/azd/extensions/azure.ai.agents/internal/cmd/run.go
trangevi: I think this is address here, but would appreciate another pair of eyes to confirm: https://github.com/Azure/azure-dev/pull/7028

cli/azd/extensions/azure.ai.agents/internal/cmd/helpers.go
trangevi: @therealjohn how do you feel about the positional argument for the service name? If we did that, I'd suggest doing that for invoke was well, instead of that positional argument being a message that's limited to a single word

therealjohn: `azd ai agent run [message]`
`azd ai agent run [agent-name] [message]`

1 argument is message, 2 is name + message. Git rid of `--name` and `--message`.

cli/azd/extensions/azure.ai.agents/internal/cmd/run.go
```diff
@@ -0,0 +1,383 @@
...
+				flags.name = args[0]
+			}
+			ctx := azdext.WithAccessToken(cmd.Context())
+			setupDebugLogging(cmd.Flags())
+			return runRun(ctx, flags)
+		},
+	}
+
+	cmd.Flags().IntVarP(&flags.port, "port", "p", DefaultPort, "Port to listen on")
+	cmd.Flags().StringVarP(&flags.startCommand, "start-command", "c", "",
```
JeffreyCA, line 70: Since this is already in the context of `azd ai agent run` we could simplify it a bit to just `--command`, any thoughts?

cli/azd/extensions/azure.ai.agents/internal/cmd/helpers.go
```diff
@@ -0,0 +1,426 @@
...
+	agentCtx := loadLocalContext(configPath)
+	if agentCtx.Sessions == nil {
+		agentCtx.Sessions = make(map[string]string)
+	}
+	if !forceNew {
+		if sid, ok := agentCtx.Sessions[agentName]; ok {
+			return sid, nil
+		}
+	}
+	sid := generateSessionID()
```
JeffreyCA: Consider adding some error handling for when `sid` is empty

cli/azd/extensions/azure.ai.agents/internal/cmd/invoke.go
```diff
@@ -0,0 +1,470 @@
...
+
+			action := &InvokeAction{flags: flags}
+			return action.Run(ctx)
+		},
+	}
+
+	cmd.Flags().BoolVarP(&flags.local, "local", "l", false, "Invoke on localhost instead of Foundry")
+	cmd.Flags().IntVar(&flags.port, "port", DefaultPort, "Local server port")
+	cmd.Flags().IntVarP(&flags.timeout, "timeout", "t", 120, "Request timeout in seconds (0 for no timeout)")
+	cmd.Flags().StringVarP(&flags.session, "session", "s", "", "Explicit session ID override")
```
JeffreyCA, line 91: Is it more common for users to continue previous sessions instead of creating new ones each time?

Just wondering if we should make it a bit easier to start new sessions, like shortening it to `--new`, supporting `-n`, or even having `azd ai agent invoke` prompt up front if the user wants to resume the last session (if exists) or create a new one

JeffreyCA, line 91: Might be helpful (if we don't already) to print out the session ID at the end of a successful `invoke`?

cli/azd/extensions/azure.ai.agents/internal/cmd/run.go
trangevi: I'd prefer it to match the entry in azure.yaml for clarity. We can always adjust it later based on user feedback.

cli/azd/extensions/azure.ai.agents/internal/cmd/helpers.go
```diff
@@ -0,0 +1,426 @@
...
+		fileExists(filepath.Join(projectDir, "requirements.txt")) {
+		if fileExists(filepath.Join(projectDir, "main.py")) {
+			return ProjectType{Language: "python", StartCmd: "python main.py"}
+		}
+		return ProjectType{Language: "python", StartCmd: ""}
+	}
+
+	// .NET: any .csproj file
+	matches, _ := filepath.Glob(filepath.Join(projectDir, "*.csproj"))
+	if len(matches) > 0 {
```
wbreza, line 172: **[High]** `generateSessionID` returns empty string on crypto failure

On `rand.Int` failure, this prints to stderr and returns `""`. The empty string propagates to the API request as `"session_id": ""`, which could cause unpredictable Foundry behavior.

Consider returning `(string, error)` so the caller can handle it, or simplify with `rand.Read`:

```suggestion
func generateSessionID() (string, error) {
b := make([]byte, 25)
if _, err := rand.Read(b); err != nil {
return "", fmt.Errorf("generating session ID: %w", err)
}
const chars = "abcdefghijklmnopqrstuvwxyz0123456789"
for i := range b {
b[i] = chars[int(b[i])%len(chars)]
}
return string(b), nil
}
```

cli/azd/extensions/azure.ai.agents/internal/cmd/invoke.go
```diff
@@ -0,0 +1,470 @@
...
+	client := &http.Client{Timeout: 30 * time.Second}
+	resp, err := client.Do(req) //nolint:gosec // G704: endpoint is resolved from azd environment configuration
+	if err != nil {
+		return "", err
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode >= 400 {
+		return "", fmt.Errorf("failed to create conversation: HTTP %d", resp.StatusCode)
+	}
```
wbreza, line 296: **[Medium]** Duplicate credential acquisition

`createConversation` creates its own `newAgentCredential()` and acquires a second token, while `invokeRemote` already has both. Consider passing the credential as a parameter to avoid two full auth flows per invoke call.

cli/azd/extensions/azure.ai.agents/internal/cmd/helpers.go
```diff
@@ -0,0 +1,426 @@
...
+	if err != nil {
+		return "", fmt.Errorf("failed to get current environment: %w", err)
+	}
+	if envResponse.Environment == nil || envResponse.Environment.Name == "" {
+		return "", fmt.Errorf("no current environment set; run 'azd env select' or 'azd init' first")
+	}
+
+	return filepath.Join(projectResponse.Project.Path, ".azure", envResponse.Environment.Name, ConfigFile), nil
+}
+
```
wbreza, line 58: **[Medium]** Wrapping nil error when Project is nil

When `err == nil` but `projectResponse.Project == nil`, `fmt.Errorf("...%w", err)` wraps nil, producing `"failed to get project config: <nil>"`. Consider splitting into separate checks:

```suggestion
if err != nil {
return "", fmt.Errorf("failed to get project config: %w", err)
}
if projectResponse.Project == nil {
return "", fmt.Errorf("failed to get project config (is there an azure.yaml?)")
}
```

cli/azd/extensions/azure.ai.agents/internal/cmd/invoke.go
trangevi: I need to confirm with John, but I think we're going to want a new-conversation flag as well, in which case we'll need the distinction. More broadly though, I don't think just "new" is clear enough, but having a -n might work

trangevi: Need to clarify session ID and conversation ID distinction, but possibly yes

trangevi: Yeah, added parameters for conversation as well, so leaving these as is

PR #7041, opened 2026-03-07
feat: add localFallback option for Docker remote build

### LINKED ISSUE #4618: [Issue] Request: support remoteBuild: "prefer" or similar
We are currently experiencing an issue with our templates that use remoteBuild: true, because free trial subscriptions cannot use that ACR functionality. Instead of folks having to know how to delete that line, what if we had "remoteBuild: prefer"? Then it could use ACR if it works, and if it gets that exception, fallback to a docker build.

spboyer:
## Problem

`azd deploy` fails for users on free trial subscriptions when `remoteBuild: true` is set in azure.yaml, because ACR Tasks are not available on free trial SKUs. Users must manually remove the `remoteBuild` line to work around this (#4618).

## Solution

Add a new `localFallback` boolean field to Docker service configuration in azure.yaml:

```yaml
services:
  api:
    host: containerapp
    docker:
      remoteBuild: true
      localFallback: true
```

When `remoteBuild: true` and `localFallback: true`, if the remote ACR build fails for any reason, azd automatically retries with a local Docker build and displays a warning:

```
WARNING: Remote build failed, falling back to local Docker build.
```

## Changes

| File | Change |
|------|--------|
| `pkg/project/framework_service_docker.go` | Add `LocalFallback` field to `DockerProjectOptions` |
| `pkg/project/container_helper.go` | Add fallback logic in `Publish()` with `ux.WarningMessage` |
| `pkg/project/mapper_registry.go` | Map `LocalFallback` in proto conversions (3 locations) |
| `grpc/proto/models.proto` | Add `local_fallback` field to proto message |
| `pkg/azdext/models.pb.go` | Generated proto code for new field |
| `schemas/v1.0/azure.yaml.json` | Add `localFallback` to schema (service-level and docker-level) |
| `pkg/project/container_helper_test.go` | Test verifying fallback triggers on remote build failure |

## Design Decisions

- **Separate field** vs changing `remoteBuild` to a string enum (`"prefer"`): A separate field is lower risk — doesn't change existing `remoteBuild` type or semantics
- **Falls back on any error**: A failed remote build already means the user gets nothing, so trying local is strictly better than failing
- **Warning style**: Uses `ux.WarningMessage` matching existing azd warning patterns

Fixes #4618

### DIFF
--- a/cli/azd/pkg/project/framework_service_docker.go
+++ b/cli/azd/pkg/project/framework_service_docker.go
@@ -20,13 +20,13 @@ import (
 )
 
 type DockerProjectOptions struct {
-	Path     string                  `yaml:"path,omitempty"        json:"path,omitempty"`
-	Context  string                  `yaml:"context,omitempty"     json:"context,omitempty"`
-	Platform string                  `yaml:"platform,omitempty"    json:"platform,omitempty"`
-	Target   string                  `yaml:"target,omitempty"      json:"target,omitempty"`
-	Registry osutil.ExpandableString `yaml:"registry,omitempty"    json:"registry"`
-	Image    osutil.ExpandableString `yaml:"image,omitempty"       json:"image"`
-	Tag      osutil.ExpandableString `yaml:"tag,omitempty"         json:"tag"`
+	Path        string                    `yaml:"path,omitempty"        json:"path,omitempty"`
+	Context     string                    `yaml:"context,omitempty"     json:"context,omitempty"`
+	Platform    string                    `yaml:"platform,omitempty"    json:"platform,omitempty"`
+	Target      string                    `yaml:"target,omitempty"      json:"target,omitempty"`
+	Registry    osutil.ExpandableString   `yaml:"registry,omitempty"    json:"registry"`
+	Image       osutil.ExpandableString   `yaml:"image,omitempty"       json:"image"`
+	Tag         osutil.ExpandableString   `yaml:"tag,omitempty"         json:"tag"`
 	RemoteBuild bool                      `yaml:"remoteBuild,omitempty"  json:"remoteBuild,omitempty"`
 	BuildArgs   []osutil.ExpandableString `yaml:"buildArgs,omitempty"   json:"buildArgs,omitempty"`
 	// not supported from azure.yaml directly yet. Adding it for Aspire to use it, initially.


vhvb1989:
Since adding a new field would require to update all templates to benefit from this, I would consider making the fallback the default behavior for `docker.remoteBuild`

However, we could also make azd to check the SKU from the ACR defined in the infrastructure when remoteBuild is enabled and warn folks about it.  Maybe as part of `azd validate` ??

cli/azd/pkg/project/container_helper.go
```diff
@@ -599,6 +600,15 @@ func (ch *ContainerHelper) Publish(
...
 		remoteImage, err = ch.runRemoteBuild(ctx, serviceConfig, targetResource, env, progress, imageOverride)
+		if err != nil && serviceConfig.Docker.LocalFallback {
+			ch.console.MessageUxItem(ctx, &ux.WarningMessage{
+				Description: fmt.Sprintf(
+					"Remote build failed: %s\nFalling back to local Docker build.", err),
+				HidePrefix: false,
+			})
+			remoteImage, err = ch.publishLocalImage(
+				ctx, serviceConfig, serviceContext, env, progress, imageOverride)
+		}
```
JeffreyCA: Can we make sure users get a useful error if they don't have Docker running during the fallback attempt?

spboyer: this should be handled by `validate` imo. I woule prefer we start Docker automatically, but the user may not use Docker but podman instead.

spboyer:
> Since adding a new field would require to update all templates to benefit from this, I would consider making the fallback the default behavior for `docker.remoteBuild`
> 
> However, we could also make azd to check the SKU from the ACR defined in the infrastructure when remoteBuild is enabled and warn folks about it. Maybe as part of `azd validate` ??

`validate` should do these checks - yes. In the meantime, if the ACR Buid fails, we are retyring with local.

schemas/v1.0/azure.yaml.json
```diff
@@ -865,6 +865,11 @@
                     "type": "boolean",
                     "title": "Optional. Whether to build the image remotely",
                     "description": "If set to true, the image will be built remotely using the Azure Container Registry remote build feature. If set to false, the image will be built locally using Docker."
+                },
+                "localFallback": {
```
weikanglim: I’m wondering if `remoteBuild` might work better as a mode rather than a boolean paired with `localFallback`.

Where the semantics are clearer:
- `prefer` → try remote build first, fall back to local if it fails
- `true` → require remote build, fail if unavailable
- `false` → use local build only

The current shape:

```yaml
docker:
  remoteBuild: true
  localFallback: true
```

introduces some configuration coupling and possible ambiguous combinations (remoteBuild: false with localFallback: true, etc.). A mode-based approach could make the behavior more explicit and easier  to reason about.

I know this was mentioned in the desigh consideration for backwards compatibility and general implementation ease. FWIW -- for gRPC changes, I'm thinking it's perfectly reasonable to deprecate the current bool and introduce a new field.

vhvb1989: 100% 
I was suggesting to make the fallback the default behavior. This `mode idea` is also good! 
I would imagine `auto` as the mode you would want to let azd just find/pick something to build the container, maybe even allowing other tools in the future.

spboyer: `prefer` turns this field into complex non boolean check, the better approach here is to make localFallback the default as @vhvb1989 stated. So that keeps the `remoteBuild` the same operation. If `true` and fails then try local.

weikanglim: Just to clarify the suggestion here: `prefer` makes the field from boolean (2-values) into an enum (3-values). The check is still done against enumerated fields. 

I'm not a huge fan of introducing fallback behavior into main paths in general, but I can see why we might want it for completely new azd users

cli/azd/pkg/project/container_helper.go
```diff
@@ -599,6 +600,22 @@ func (ch *ContainerHelper) Publish(
 
 	if serviceConfig.Docker.RemoteBuild {
 		remoteImage, err = ch.runRemoteBuild(ctx, serviceConfig, targetResource, env, progress, imageOverride)
+		if err != nil {
```
wbreza, line 603: **[High]** Consider failing fast in CI instead of silently falling back

The project already has `IsRunningOnCI()` infrastructure (`internal/tracing/resource/ci.go`) that detects Azure Pipelines, GitHub Actions, and other CI providers. In CI, silently falling back to local Docker can mask configuration issues — operators should fix their `remoteBuild` setup rather than have deployments silently change build strategy.

Consider something like:
```go
if err != nil {
    // In CI, fail fast rather than silently changing build strategy
    if resource.IsRunningOnCI() {
        return nil, fmt.Errorf(
            "remote build failed: %w\n\n"+
                "Running in CI — not falling back to local build. "+
                "Fix the remote build configuration or set remoteBuild to false.", err)
    }
    // Interactive: check for local runtime and fall back
    ...
```

cli/azd/pkg/project/framework_service_docker.go
```diff
@@ -21,14 +21,14 @@ import (
...
 type DockerProjectOptions struct {
-	Path        string                    `yaml:"path,omitempty"        json:"path,omitempty"`
-	Context     string                    `yaml:"context,omitempty"     json:"context,omitempty"`
-	Platform    string                    `yaml:"platform,omitempty"    json:"platform,omitempty"`
-	Target      string                    `yaml:"target,omitempty"      json:"target,omitempty"`
-	Registry    osutil.ExpandableString   `yaml:"registry,omitempty"    json:"registry,omitempty"`
-	Image       osutil.ExpandableString   `yaml:"image,omitempty"       json:"image,omitempty"`
-	Tag         osutil.ExpandableString   `yaml:"tag,omitempty"         json:"tag,omitempty"`
-	RemoteBuild bool                      `yaml:"remoteBuild,omitempty" json:"remoteBuild,omitempty"`
+	Path          string                    `yaml:"path,omitempty"        json:"path,omitempty"`
```
wbreza: **[Medium]** Residual whitespace-only changes

This file and `mapper_registry.go` contain only field-alignment whitespace changes with no functional modification. This looks like residual formatting from when `LocalFallback` was a struct field — the alignment was adjusted for the longer name, then the field was removed but the alignment kept.

Consider reverting these whitespace changes to reduce diff noise and keep git blame clean.

schemas/v1.0/azure.yaml.json
```diff
@@ -864,7 +864,7 @@
                 "remoteBuild": {
                     "type": "boolean",
                     "title": "Optional. Whether to build the image remotely",
-                    "description": "If set to true, the image will be built remotely using the Azure Container Registry remote build feature. If set to false, the image will be built locally using Docker."
+                    "description": "If set to true, the image will be built remotely using the Azure Container Registry remote build feature. If the remote build fails, azd automatically falls back to building locally using Docker or Podman if available. If set to false, the image will be built locally."
```
wbreza, line 867: **[Medium]** Design thread on `remoteBuild` semantics needs resolution

There's an active discussion between reviewers about whether `remoteBuild` should become a mode enum (`prefer`/`true`/`false`/`auto`) vs keeping it boolean with implicit fallback. The current boolean approach with automatic fallback is my preference — it's simpler and the right default UX. But the thread should be resolved explicitly before merge.

The schema description update here nicely documents the fallback behavior, so if the boolean approach is chosen, it's ready to go.

cli/azd/pkg/project/framework_service_docker.go
rajeshkamal5050: I think this has been addressed with the latest changes.

cli/azd/grpc/proto/models.proto
```diff
@@ -117,6 +117,8 @@ message DockerProjectOptions {
   string tag = 7;
   bool remote_build = 8;
   repeated string build_args = 9;
+  reserved 10;
+  reserved "local_fallback";
```
JeffreyCA, line 121: This should probably be reverted since this never shipped to begin with: https://github.com/Azure/azure-dev/tasks/44f8c414-03e2-4630-bce3-3e8c5278065f

Issue #6586, opened 2026-01-22
Enhance CLI interactivity for LLM/automation scenarios


spboyer:
## Problem Statement

LLMs (GitHub Copilot, Claude, etc.) and automation tools running azd commands like `init`, `up`, `provision` encounter interactive prompts that block execution. The current `--no-prompt` flag only uses default values or fails when no default exists—it does not allow callers to **provide answers** programmatically.

### Current Limitations

When an LLM runs `azd init` or `azd provision`:
1. **Prompts block execution** - The CLI waits for user input that LLMs cannot provide through normal terminal interaction
2. **`--no-prompt` is insufficient** - It only accepts defaults or fails; there is no mechanism to supply specific answers
3. **External prompt service is complex** - The `AZD_UI_PROMPT_ENDPOINT` mechanism requires setting up an HTTP server, which is impractical for simple LLM automation

### Affected Commands

- `azd init` - Template selection, environment naming, service configuration
- `azd provision` - Subscription selection, location selection, parameter inputs
- `azd up` - Combines init + provision prompts
- `azd down` - Confirmation prompts (partially addressed by `--force`)
- Any command with IaC parameter prompts

## Proposed Solution

Implement complementary mechanisms for LLM-friendly interactivity:

### 1. JSON Prompt Mode (`--json-prompts`)

When enabled, prompts output structured JSON to stdout and read JSON responses from stdin, enabling real-time LLM interaction:

**Prompt Output (stdout):**
```json
{
  "type": "prompt",
  "prompt": {
    "id": "subscription",
    "kind": "select",
    "message": "Select an Azure subscription",
    "choices": [
      {"value": "sub-123", "detail": "My Subscription 1"},
      {"value": "sub-456", "detail": "My Subscription 2"}
    ],
    "defaultValue": "sub-123"
  }
}
```

**Response Input (stdin):**
```json
{"id": "subscription", "value": "sub-456"}
```

### 2. Upfront Answers Flag (`--answers`)

Accept pre-defined answers as JSON, allowing all inputs to be provided before execution:

```bash
# Inline JSON
azd init --answers '{"template":"todo-nodejs-mongo","environment":"dev"}'

# File reference  
azd provision --answers @./answers.json
```

### 3. Input Introspection (`--describe-inputs`) [Future]

List all possible prompts a command might ask before running:

```bash
azd init --describe-inputs --output json
```

## Benefits

[...]

PR #7044, opened 2026-03-09
Set a default value for resource group prompt

weikanglim:
Set a default value for resource group prompt

This allows resource-group deployment with `AZURE_RESOURCE_GROUP` set in `no-prompt` mode to continue. If the resource group already exists, the prompt is skipped. If the resource group does not exist, we create it.

Contributes to #6586

### DIFF
--- a/cli/azd/pkg/prompt/prompter.go
+++ b/cli/azd/pkg/prompt/prompter.go
@@ -188,11 +188,18 @@ func (p *DefaultPrompter) PromptResourceGroupFrom(
 		choices[idx+canCreateOverride] = fmt.Sprintf("%d. %s", idx+canCreateOverride+1, group.Name)
 	}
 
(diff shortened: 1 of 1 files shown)


cli/azd/pkg/prompt/prompter.go
```diff
@@ -188,11 +188,18 @@ func (p *DefaultPrompter) PromptResourceGroupFrom(
...
 
-	choice, err := p.console.Select(ctx, input.ConsoleOptions{
+	pickPrompt := input.ConsoleOptions{
 		Message: "Pick a resource group to use:",
 		Options: choices,
 		Help:    options.PickResourceGroupHelp,
-	})
+	}
+
+	if canCreateNeResourceGroup && options.DefaultName != "" {
```
hemarina, line 197: NIT
```suggestion
	if canCreateNewResourceGroup && options.DefaultName != "" {
```

PR #7039, opened 2026-03-07
docs: comprehensive environment variables documentation

spboyer:
Expand `environment-variables.md` from ~12 variables to 100+ organized into 14 sections covering all env vars used by azd. Add AGENTS.md maintenance guidelines.

Recreated from #6653 (auto-closed by inactivity bot) with updates for new variables added since.

Fixes #6652

PR #7045, opened 2026-03-09
feat: add configurable deploy timeout via --timeout flag and azure.yaml

### LINKED ISSUE #6555: [Feature request] Allow the configuration of a custom timeout duration on `azd provision`

**Output from `azd version`**
azd version 1.23.0 (commit 9a02add5930c2c87b6b2c78443e4d8132587417d)

**Describe the issue**
Currently there is a 20 minutes (for app service deployments) timeout on azd deployments. While it is good to have this timeout, the current duration is way too high for simple deployments in a pipeline environment. It would be great to be able to configure a shorter timeout to have faster feedback from the deployment.

**Expected solution**
The ability to run a command such as `azd deploy service --fromPackage path --timeout timeInSeconds` to configure the deployment timeout.

spboyer:
## Problem

`azd deploy` has no user-configurable timeout. The Azure SDK default timeout is effectively unbounded for App Service and Container Apps deployments, making pipeline feedback loops slow when deployments hang (#6555).

## Solution

Add a `--timeout` flag to `azd deploy` and a `deployTimeout` field in `azure.yaml` service configuration:

```yaml
services:
  api:
    host: containerapp
    deployTimeout: 300  # 5 minutes
```

```bash
azd deploy --timeout 300
```

**Timeout resolution order:** CLI flag > azure.yaml service config > default (1200s / 20 min)

The timeout wraps the deploy context with `context.WithTimeout`, which propagates through all Azure SDK `PollUntilDone` calls naturally.

## Changes

| File | Change |
|------|--------|
| `cli/azd/internal/cmd/deploy.go` | Add `--timeout` flag, `resolveDeployTimeout()`, apply via `context.WithTimeout` |
| `cli/azd/internal/cmd/deploy_test.go` | 15 table-driven tests: flag parsing, resolution priority, validation, integration |
| `cli/azd/pkg/project/service_config.go` | Add `DeployTimeout *int` field |
| `cli/azd/pkg/project/service_config_test.go` | YAML parsing tests for deployTimeout |
| `cli/azd/pkg/project/project_config_test.go` | Project config parsing with deployTimeout |
| `schemas/v1.0/azure.yaml.json` | Schema: `deployTimeout` integer property |
| `cli/azd/cmd/testdata/TestFigSpec.ts` | Updated snapshot |
| `cli/azd/cmd/testdata/TestUsage-azd-deploy.snap` | Updated snapshot |

## Testing

- `go build ./...` ✅
- 15 deploy timeout tests pass ✅
- Service config parsing tests pass ✅
- Snapshot tests updated and pass ✅

Fixes #6555

### DIFF
--- a/cli/azd/internal/cmd/deploy.go
+++ b/cli/azd/internal/cmd/deploy.go
@@ -5,10 +5,12 @@ package cmd
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"io"
 	"log"
 	"os"
+	"strconv"
 	"strings"
 	"time"
 
@@ -34,10 +36,14 @@ import (
 type DeployFlags struct {
 	ServiceName string
 	All         bool
+	Timeout     int
 	fromPackage string
+	flagSet     *pflag.FlagSet
 	global      *internal.GlobalCommandOptions
 	*internal.EnvFlag
 }
+
+const defaultDeployTimeoutSeconds = 1200
 
 func (d *DeployFlags) Bind(local *pflag.FlagSet, global *internal.GlobalCommandOptions) {
 	d.BindNonCommon(local, global)
@@ -62,6 +68,7 @@ func (d *DeployFlags) BindNonCommon(
 func (d *DeployFlags) bindCommon(local *pflag.FlagSet, global *internal.GlobalCommandOptions) {
 	d.EnvFlag = &internal.EnvFlag{}
 	d.EnvFlag.Bind(local, global)
+	d.flagSet = local
 
 	local.BoolVar(
 		&d.All,
@@ -75,6 +82,16 @@ func (d *DeployFlags) bindCommon(local *pflag.FlagSet, global *internal.GlobalCo
 		"",
 		//nolint:lll
 		"Deploys the packaged service located at the provided path. Supports zipped file packages (file path) or container images (image tag).",
+	)
+	local.IntVar(
+		&d.Timeout,
+		"timeout",
+		defaultDeployTimeoutSeconds,
+		fmt.Sprintf(
+			"Maximum time in seconds for azd to wait for each service deployment. This stops azd from waiting "+
+				"but does not cancel the Azure-side deployment. (default: %d)",
+			defaultDeployTimeoutSeconds,
+		),
 	)
 }
 
@@ -91,9 +108,19 @@ func NewDeployFlags(cmd *cobra.Command, global *internal.GlobalCommandOptions) *
 
 func NewDeployFlagsFromEnvAndOptions(envFlag *internal.EnvFlag, global *internal.GlobalCommandOptions) *DeployFlags {
 	return &DeployFlags{
+		Timeout: defaultDeployTimeoutSeconds,
 		EnvFlag: envFlag,
 		global:  global,
 	}
+}
+
+func (d *DeployFlags) timeoutChanged() bool {
+	if d.flagSet == nil {
+		return false
+	}
+
+	timeoutFlag := d.flagSet.Lookup("timeout")
+	return timeoutFlag != nil && timeoutFlag.Changed
 }
 
 func NewDeployCmd() *cobra.Command {
@@ -241,6 +268,10 @@ func (da *DeployAction) Run(ctx context.Context) (*actions.ActionResult, error) 
 	}
 
 	deployResults := map[string]*project.ServiceDeployResult{}
+	deployTimeout, err := da.resolveDeployTimeout()
+	if err != nil {
+		return nil, err
+	}
 
 	err = da.projectConfig.Invoke(ctx, project.ProjectEventDeploy, projectEventArgs, func() error {
 		for _, svc := range stableServices {
@@ -303,18 +334,45 @@ func (da *DeployAction) Run(ctx context.Context) (*actions.ActionResult, error) 
 				return err
 			}
 
+			deployCtx, deployCancel := context.WithTimeout(ctx, deployTimeout)
+			defer deployCancel()
+
 			deployResult, err := async.RunWithProgress(
 				func(deployProgress project.ServiceProgress) {
 					progressMessage := fmt.Sprintf("Deploying service %s (%s)", svc.Name, deployProgress.Message)
 					da.console.ShowSpinner(ctx, progressMessage, input.Step)
 				},
 				func(progress *async.Progress[project.ServiceProgress]) (*project.ServiceDeployResult, error) {
-					return da.serviceManager.Deploy(ctx, svc, serviceContext, progress)
+					return da.serviceManager.Deploy(deployCtx, svc, serviceContext, progress)
 				},
 			)
 
 			if err != nil {
 				da.console.StopSpinner(ctx, stepMessage, input.StepFailed)
+				if deployCtx.Err() == context.DeadlineExceeded {
+					warnMsg := fmt.Sprintf(
+						"Deployment of service '%s' exceeded the azd wait timeout."+
+							" azd has stopped waiting, but the deployment may"+
+							" still be running in Azure.",
+						svc.Name,
+					)
+					da.console.MessageUxItem(ctx, &ux.WarningMessage{
+						Description: warnMsg,
+						Hints: []string{
+							"Check the Azure Portal for current deployment status.",
+							"Increase timeout with --timeout flag or AZD_DEPLOY_TIMEOUT env var.",
+						},
+					})
+
+					return fmt.Errorf(
+						"deployment of service '%s' timed out after %d seconds. To increase, use --timeout flag "+
+							"or AZD_DEPLOY_TIMEOUT env var. Note: azd has stopped "+
+							"waiting, but the deployment may still be running in Azure. Check the Azure Portal for "+
+							"current deployment status.",
+						svc.Name,
+						int(deployTimeout.Seconds()),
+					)
+				}
 				return err
 			}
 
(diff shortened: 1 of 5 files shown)


vhvb1989:
How can we communicate to folks that the timeout is only to stop azd but not the actual deployment?

For AppService, for example, azd can be stopped, during zip deployment, before sending the zip file and starting the service deployment. That would be a valid cancellation in both sides.
But if the zip file has been sent and the service is performing the deployment, azd can be stopped but the service will continue anyways.

spboyer:
> How can we communicate to folks that the timeout is only to stop azd but not the actual deployment?
> 
> For AppService, for example, azd can be stopped, during zip deployment, before sending the zip file and starting the service deployment. That would be a valid cancellation in both sides. But if the zip file has been sent and the service is performing the deployment, azd can be stopped but the service will continue anyways.

What I am going to change here is a global timeout setting for deployments instead of per service

.squad/agents/cinderella/history.md
hemarina, line 1: Could you remove all agent memory in `.squad/agents`?

cli/azd/internal/cmd/deploy.go
```diff
@@ -301,18 +323,46 @@ func (da *DeployAction) Run(ctx context.Context) (*actions.ActionResult, error)
...
-				},
-			)
+			deployTimeout, err := da.resolveDeployTimeout()
+			if err != nil {
+				da.console.StopSpinner(ctx, stepMessage, input.StepFailed)
+				return err
+			}
+
+			var deployResult *project.ServiceDeployResult
+			deployCtx, cancel := context.WithTimeout(ctx, deployTimeout)
```
hemarina: Agrees with moving global timeout setting for deployments instead of per service. Help message describes `Maximum time in seconds to wait for deployment to complete (default: 1200)`. When users set --timeout 300, they expect entire deploy should finish in 5 minutes, but it actually would be 5n minutes (5 min × n services).

spboyer:
Regarding hemarina's feedback about per-service vs total timeout: the current implementation applies `--timeout` per-service, so `--timeout 300` with 3 services could take up to 15 minutes total. This is a design decision for the team to weigh in on — should it be total time across all services, or per-service? Both have valid use cases. Leaving the thread open for discussion.

cli/azd/internal/cmd/deploy.go
```diff
@@ -301,18 +323,49 @@ func (da *DeployAction) Run(ctx context.Context) (*actions.ActionResult, error)
...
+					},
+					func(progress *async.Progress[project.ServiceProgress]) (*project.ServiceDeployResult, error) {
+						return da.serviceManager.Deploy(deployCtx, svc, serviceContext, progress)
+					},
+				)
+			}()
 
 			if err != nil {
 				da.console.StopSpinner(ctx, stepMessage, input.StepFailed)
+				if errors.Is(err, context.DeadlineExceeded) {
```
weikanglim: This error check is ambiguous and may fire if timeout happens inside the deploy operation.

Suggestion: should check `deployCtx.Error()` directly.

cli/azd/internal/cmd/deploy.go
```diff
@@ -301,18 +323,49 @@ func (da *DeployAction) Run(ctx context.Context) (*actions.ActionResult, error)
...
-			deployResult, err := async.RunWithProgress(
-				func(deployProgress project.ServiceProgress) {
-					progressMessage := fmt.Sprintf("Deploying service %s (%s)", svc.Name, deployProgress.Message)
-					da.console.ShowSpinner(ctx, progressMessage, input.Step)
-				},
-				func(progress *async.Progress[project.ServiceProgress]) (*project.ServiceDeployResult, error) {
-					return da.serviceManager.Deploy(ctx, svc, serviceContext, progress)
-				},
-			)
+			deployTimeout, err := da.resolveDeployTimeout()
```
weikanglim: The timeout resolution (flag → env → default) is identical for every service, yet it runs once per service iteration. Move it before the `projectConfig.Invoke()` loop so we fail fast parsing instead of during the deployment.

I vote for timeout per-service.

cli/azd/internal/cmd/deploy.go
```diff
@@ -301,18 +323,49 @@ func (da *DeployAction) Run(ctx context.Context) (*actions.ActionResult, error)
...
+							" still be running in Azure.",
+						svc.Name,
+					)
+					da.console.MessageUxItem(ctx, &ux.WarningMessage{
+						Description: warnMsg,
+						Hints:       []string{"Check the Azure Portal for current deployment status."},
+					})
+
+					return fmt.Errorf(
+						"deployment of service '%s' timed out after %d seconds. Note: azd has stopped waiting, "+
```
weikanglim: It could be worth mentioning how to suggest timeout (flag or env var)

weikanglim (commented):
Overall looks good -- have some comments

cli/azd/internal/cmd/deploy.go
rajeshkamal5050: +1 on the per-service timeout we get more finer control, instead of a uber one.

cli/azd/internal/cmd/deploy.go
```diff
@@ -377,6 +433,55 @@ func (da *DeployAction) Run(ctx context.Context) (*actions.ActionResult, error)
...
+func (da *DeployAction) resolveDeployTimeout(serviceConfig *project.ServiceConfig) (time.Duration, error) {
+	if da.flags.timeoutChanged() {
+		if da.flags.Timeout <= 0 {
+			return 0, errors.New("invalid value for --timeout: must be greater than 0 seconds")
+		}
+
+		return time.Duration(da.flags.Timeout) * time.Second, nil
+	}
+
+	if envVal, ok := os.LookupEnv("AZD_DEPLOY_TIMEOUT"); ok {
```
wbreza, line 449: **[High]** `AZD_DEPLOY_TIMEOUT` env var is undocumented

This PR introduces `AZD_DEPLOY_TIMEOUT` as the middle tier of the resolution precedence (flag > env var > azure.yaml > default), but it's not documented anywhere. Per the `AGENTS.md` guideline added in #7039, any new `os.LookupEnv` call should be documented in `cli/azd/docs/environment-variables.md`.

`suggestion
if envVal, ok := os.LookupEnv("AZD_DEPLOY_TIMEOUT"); ok {
// NOTE: Document this env var in cli/azd/docs/environment-variables.md
`
"@
        },
        @{
            path = "cli/azd/internal/cmd/deploy.go"
            line = 451
            side = "RIGHT"
            body = @"
**[Medium]** Design: `deployTimeout` naming vs. generic `timeout`

I did a cross-command timeout analysis: **no other azd command has a timeout mechanism today.** All `PollUntilDone` calls across provision, container apps, and ARM deployments use `PollUntilDone(ctx, nil)` with no timeout.

This means `deployTimeout` sets a precedent for how we name service-level timeouts. A few options worth weighing as a team:

1. **`deployTimeout`** (current) — Specific and clear. But if we later add timeouts for restore/build/package, we'd get `restoreTimeout`, `buildTimeout`, etc. — field proliferation.
2. **`timeout`** — Generic, applies to the longest-running service operation (deploy). If we add timeouts for other service commands, this becomes a shared default that can be overridden per-command.
3. **Keep `deployTimeout`** and accept that provision timeouts would live under `infra:` (different scope entirely).

I could live with `deployTimeout` if the team prefers specificity, but I'd lean toward a more generic `timeout` at the service level that defaults to applying to deploy (since that's the primary long-running operation). Worth a quick team discussion.

cli/azd/internal/cmd/deploy.go
```diff
@@ -301,18 +327,48 @@ func (da *DeployAction) Run(ctx context.Context) (*actions.ActionResult, error)
...
-					da.console.ShowSpinner(ctx, progressMessage, input.Step)
-				},
-				func(progress *async.Progress[project.ServiceProgress]) (*project.ServiceDeployResult, error) {
-					return da.serviceManager.Deploy(ctx, svc, serviceContext, progress)
-				},
-			)
+			deployTimeout := deployTimeouts[svc.Name]
+			var deployResult *project.ServiceDeployResult
+			var deployCtx context.Context
+			func() {
```
wbreza: **[Medium]** Anonymous function for context scoping is unusual

This wraps the deploy in an anonymous function solely to scope `defer deployCancel()`, but `deployCtx` and `err` escape the function and are used afterward. While technically correct (Go preserves `DeadlineExceeded` after cancel), this is fragile and non-idiomatic.

Consider simplifying — `defer` in the loop is fine here since the loop body returns on any error:

```go
deployTimeout := deployTimeouts[svc.Name]
deployCtx, deployCancel := context.WithTimeout(ctx, deployTimeout)
defer deployCancel()

deployResult, err := async.RunWithProgress(
    func(deployProgress project.ServiceProgress) { ... },
    func(progress *async.Progress[project.ServiceProgress]) (*project.ServiceDeployResult, error) {
        return da.serviceManager.Deploy(deployCtx, svc, serviceContext, progress)
    },
)
```

cli/azd/internal/cmd/deploy.go
weikanglim: I'm not personally a huge fan of `deployTimeout` either. I could see us avoid introducing the azure.yaml schema change for now just to make things easier to land. 

The `azure.yaml` schema isn't making this easy to land. I could imagine that each service has a separate section for `deploy`, `package` parameters 

```yaml
services:
  api:
    deploy:
      timeout: 30000
```

And this would be very nice to add different configuration that customizes the command.

spboyer: Agree — will document `AZD_DEPLOY_TIMEOUT` in `cli/azd/docs/environment-variables.md`. Good catch.

On the naming discussion with @weikanglim's suggestion of a nested structure:

```yaml
services:
  api:
    deploy:
      timeout: 300
```

I like this approach — it's more extensible for future per-command config (`package`, `build`, etc.) without field proliferation. However, it's a larger schema change. For this PR, I'm inclined to:

1. **Keep `--timeout` flag as-is** (it's the CLI flag, straightforward)
2. **Drop the `deployTimeout` azure.yaml field for now** to unblock the PR
3. **Follow up** with the nested `deploy: { timeout: ... }` schema in a separate PR once we align on the pattern

This way we land the immediate value (flag + env var) without committing to a schema pattern we may want to revise. Thoughts?

cli/azd/internal/cmd/deploy.go
spboyer: Good call — the anonymous function wrapper is awkward. Will simplify to the pattern you suggested:

```go
deployTimeout := deployTimeouts[svc.Name]
deployCtx, deployCancel := context.WithTimeout(ctx, deployTimeout)
defer deployCancel()
```

Since we return on error inside the loop body, `defer` in the loop works correctly here and is much cleaner.

.squad/agents/cinderella/history.md
spboyer: Will remove the `.squad/agents` directory from this PR.

cli/azd/internal/cmd/deploy.go
spboyer: Good catch — will switch to checking `deployCtx.Err() == context.DeadlineExceeded` directly instead of the ambiguous outer `err` check.

cli/azd/internal/cmd/deploy.go
spboyer: Agree on both points — timeout resolution should happen before the loop (fail fast), and per-service timeout via azure.yaml gives finer control. The `--timeout` flag will be the global override, and `deployTimeout` in azure.yaml will be per-service. Will refactor.

cli/azd/internal/cmd/deploy.go
spboyer: Will add the `--timeout` flag and `AZD_DEPLOY_TIMEOUT` env var to the timeout error message.

cli/azd/internal/cmd/deploy.go
weikanglim: Agreed -- Drop the azure.yaml change. You have my approval on that.

PR #7047, opened 2026-03-10
Update progress log previewer to handle no-tty mode better

weikanglim:
Changes:
- Pass the console width to progress log previewer. This allows the component to respect azd's TTY behavior, including `AZD_FORCE_TTY`
- Add a check in progress log previewer to no-opt during `Start()`. This was previously logging blank lines.

Contributes to https://github.com/Azure/azure-dev/issues/6586

### DIFF
--- a/cli/azd/pkg/input/console.go
+++ b/cli/azd/pkg/input/console.go
@@ -336,7 +336,18 @@ func (c *AskerConsole) ShowPreviewer(ctx context.Context, options *ShowPreviewer
 		options = defaultShowPreviewerOptions()
 	}
 
-	c.previewer = NewProgressLog(options.MaxLineCount, options.Prefix, options.Title, c.currentIndent.Load()+currentMsg)
+	c.previewer = newProgressLogWithWidthFn(
+		options.MaxLineCount,
+		options.Prefix,
+		options.Title,
+		c.currentIndent.Load()+currentMsg,
+		func() int {
+			if c.consoleWidth == nil {
+				return 0
+			}
+
+			return int(c.consoleWidth.Load())
+		})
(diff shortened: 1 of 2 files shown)


cli/azd/pkg/input/console.go
```diff
@@ -336,7 +336,18 @@ func (c *AskerConsole) ShowPreviewer(ctx context.Context, options *ShowPreviewer
...
 	}
 
-	c.previewer = NewProgressLog(options.MaxLineCount, options.Prefix, options.Title, c.currentIndent.Load()+currentMsg)
+	c.previewer = newProgressLogWithWidthFn(
+		options.MaxLineCount,
+		options.Prefix,
+		options.Title,
+		c.currentIndent.Load()+currentMsg,
+		func() int {
+			if c.consoleWidth == nil {
```
hemarina, line 345: Just curious, is it the same if using isTerminal?

cli/azd/pkg/input/console.go
```diff
@@ -336,7 +336,18 @@ func (c *AskerConsole) ShowPreviewer(ctx context.Context, options *ShowPreviewer
...
+				return 0
+			}
+
+			return int(c.consoleWidth.Load())
+		})
```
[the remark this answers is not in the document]
weikanglim: That's the entire intention of the change

cli/azd/pkg/input/console.go
weikanglim: It's indeed equivalent:

	// consoleWidth is the width of the underlying console window. The value is updated as the window resized. Nil when
	// isTerminal is false.

PR #7051, opened 2026-03-10
Improve telemetry error classification with typed sentinels

rajeshkamal5050:
### Problem

Telemetry for azd 1.23.7/1.23.8 showed ~15k errors as `internal.errors_errorString` — the catch-all bucket when `MapError` can't classify an error. Most `Run()` methods used bare `errors.New` or `fmt.Errorf` with no typed error, so `MapError` had nothing to match on.

Additionally, errors wrapped in `ErrorWithSuggestion` only showed `error.suggestion` as the ResultCode, with `error.type` empty due to a backend GDPR classification gap (now fixed by @weikanglim).

### What this does

1. **26 typed sentinel errors** in `internal/errors.go` covering all command domains (env, auth, config, extensions, deploy, etc.)
2. **54 bare error sites fixed** across 17 command files — each now wraps a sentinel via `ErrorWithSuggestion` for both telemetry classification and user-facing suggestions
3. **`classifySentinel()` helper** in `MapError` — when an error hits the `error.suggestion` branch, this populates `error.type` with a descriptive code (e.g. `internal.key_not_found`) instead of the raw Go type
4. **Two AST enforcement tests** to prevent regression:
   - `Test_RunMethodsNoBareErrors` — scans all action `Run()` methods, empty allowlist
   - `Test_PackageLevelErrorsMapped` — ensures every sentinel in `internal/errors.go` appears in `classifySentinel()`

### Telemetry result

Errors now show up as:
- `ResultCode: error.suggestion` with `error.type: internal.key_not_found` (for ErrorWithSuggestion-wrapped sentinels)
- `ResultCode: internal.extension_not_found` (for naked sentinel returns)

Instead of the opaque `internal.errors_errorString`.

### Testing

- `Test_MapError` — covers all sentinel paths, ErrorWithSuggestion wrapping, and catch-all guardrails
- `TestMapError_ErrorWithSuggestionSetsErrorType` — verifies `classifySentinel()` populates `error.type`
- `Test_PackageLevelErrorsMapped` and `Test_RunMethodsNoBareErrors` — AST enforcement
- Functional tests updated (`env_test.go`)
- Validated with `--trace-log-file`

### DIFF
--- a/cli/azd/internal/cmd/errors.go
+++ b/cli/azd/internal/cmd/errors.go
@@ -67,8 +67,13 @@ func MapError(err error, span tracing.Span) {
 		errCode = "auth.login_required"
 	} else if errors.As(err, &errWithSuggestion) {
 		errCode = "error.suggestion"
-		errType := errorType(errWithSuggestion.Unwrap())
-		span.SetAttributes(fields.ErrType.String(errType))
+		inner := errWithSuggestion.Unwrap()
+		if code := classifySentinel(inner); code != "" {
+			span.SetAttributes(fields.ErrType.String(code))
+		} else {
+			span.SetAttributes(
+				fields.ErrType.String(errorType(inner)))
+		}
 	} else if errors.As(err, &respErr) {
 		serviceName := "other"
 		statusCode := -1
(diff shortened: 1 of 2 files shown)


cli/azd/cmd/extensions.go
```diff
@@ -228,7 +231,10 @@ func (a *extensionAction) Run(ctx context.Context) (*actions.ActionResult, error
 
 	jwtToken, err := grpcserver.GenerateExtensionToken(extension, serverInfo)
 	if err != nil {
-		return nil, fmt.Errorf("failed to generate extension token")
+		return nil, &internal.ErrorWithSuggestion{
+			Err:        internal.ErrExtensionTokenFailed,
+			Suggestion: "This is an internal error generating the JWT token for extension communication.",
+		}
 	}
```
JeffreyCA, line 235: These suggestions don't seem to be actionable, could we simply return the base error instead?

cli/azd/internal/cmd/errors.go
```diff
@@ -59,15 +61,21 @@ func MapError(err error, span tracing.Span) {
...
-	if errors.As(err, &updateErr) {
+	// If the error is wrapped in ErrorWithSuggestion, unwrap it first so the inner error
+	// can be classified by its actual type (ResponseError, AzureDeploymentError, etc.)
+	// instead of being bucketed as the opaque "error.suggestion".
+	classifyErr := err
+	if errors.As(err, &errWithSuggestion) {
+		if inner := errWithSuggestion.Unwrap(); inner != nil {
+			classifyErr = inner
+		}
+	}
```
JeffreyCA: Just checking we validated this manually with something like [`--trace-log-file`](https://github.com/Azure/azure-dev/blob/main/cli/azd/docs/tracing-in-azd.md#-observing-new-traces)?

cli/azd/internal/cmd/errors.go
```diff
@@ -59,15 +61,21 @@ func MapError(err error, span tracing.Span) {
 	var loginErr *auth.ReLoginRequiredError
 	var updateErr *update.UpdateError
 
-	if errors.As(err, &updateErr) {
+	// If the error is wrapped in ErrorWithSuggestion, unwrap it first so the inner error
+	// can be classified by its actual type (ResponseError, AzureDeploymentError, etc.)
+	// instead of being bucketed as the opaque "error.suggestion".
+	classifyErr := err
+	if errors.As(err, &errWithSuggestion) {
+		if inner := errWithSuggestion.Unwrap(); inner != nil {
```
weikanglim: @rajeshkamal5050 The goal behind classifiying suggestion errors as `error.suggestion` is to signal that the error is marked as "handled" by a developer. The error type unwrapped was still available as a separate attribute `error.type`.

What would we like to do differently?

cli/azd/internal/errors.go
```diff
@@ -24,3 +28,87 @@ func (et *ErrorWithTraceId) Error() string {
 func (et *ErrorWithTraceId) Unwrap() error {
 	return et.Err
 }
+
+// Command sentinel errors for telemetry classification.
```
weikanglim, line 32: I don't think this is the general approach we want, creating sentinel errors for all operations regardless of actual usage -- I have some things I could land to make this easier, but I'm happy to refactor and clean this up later.

cli/azd/internal/cmd/errors.go
rajeshkamal5050: When looking up data from telemetry with `error.suggestion` the `error.type` was empty for azd version 1.23.7++. Can you also check??

weikanglim: It was suppressed due to missing GDPR classification. I updated it.

cli/azd/cmd/extensions.go
rajeshkamal5050: Fixed — removed the non-actionable suggestions and returning the base error directly now.

cli/azd/internal/cmd/errors.go
rajeshkamal5050: Reverted the unwrap logic — Wei confirmed the original error.suggestion + error.type design works correctly. The empty error.type was a backend GDPR classification issue which he already fixed.

cli/azd/internal/cmd/errors.go
rajeshkamal5050: Thanks for the fix Wei. Reverted the unwrap logic — original design restored.

cli/azd/internal/errors.go
rajeshkamal5050: Yeah agree, typed errors for everything felt too much. My thinking was mainly around errors that show up in telemetry — those should be typed so we do not end up with generic ones. Internal errors that never hits telemetry doesn't need it. Is that what you were thinking too?

weikanglim: I think if we cetnralize on an error struct that provides the functionality we care about:

```golang
type Error struct {
	Code        string        // stable identifier (ex: "ResourceGroupLocationMismatch")
	Message     string        // user-facing summary
	Suggestions []string  // actionable fixes
	Links       []ErrorLink   // docs or references
	Err         error         // underlying cause
}

type ErrorLink struct {
	Title string
	URL   string
}
```

And update agents.md to say "always use `internal.Error`" and fill in a Code that makes sense for telemetry classification. This would work for all cases where we have new errors.

We would also provide an interface for custom error types to implement the same functionality

```golang
type Coded interface {
	Code() string
}

type UserMessage interface {
	UserMessage() string
}

type Suggestable interface {
	Suggestions() []string
}

type Linkable interface {
	Links() []ErrorLink
}
```

cli/azd/internal/cmd/errors.go
```diff
@@ -220,6 +222,57 @@ func MapError(err error, span tracing.Span) {
 		errCode = "update.elevationRequired"
 	} else if errors.Is(err, pipeline.ErrRemoteHostIsNotAzDo) {
 		errCode = "internal.remote_not_azdo"
+	} else if errors.Is(err, internal.ErrInfraNotProvisioned) {
+		errCode = "internal.infra_not_provisioned"
+	} else if errors.Is(err, internal.ErrFromPackageWithAll) ||
```
weikanglim: The conditions to emit error codes may not be working now that they would be classified under "error.suggestion" (the if-branch filters it down). On the other hand, we're getting nice error types with the newly named errors. Not sure about your preference -- I do think eventually we want error codes to be surfaceable similar to my longer term [suggestion](https://github.com/Azure/azure-dev/pull/7051#discussion_r2914570291)

weikanglim (approved):

cli/azd/internal/cmd/errors.go
rajeshkamal5050: Good catch — yeah those sentinel checks in the main chain were dead code since `error.suggestion` matched first. Latest push addresses this:

- Added `classifySentinel()` inside the `error.suggestion` branch — populates `error.type` with descriptive codes like `internal.key_not_found` instead of raw Go types
- Removed the dead checks from the main chain (only `ErrExtensionNotFound` and `ErrExtensionTokenFailed` stay since they are returned naked)

Validated with `--trace-log-file`. On your `internal.Error` struct idea — makes sense, will let you file an issue and follow-up.

PR #7070, opened 2026-03-11
Update README with troubleshooting info

bob-okeefe:
Added troubleshooting information regarding azd init

### DIFF
--- a/cli/azd/extensions/azure.coding-agent/README.md
+++ b/cli/azd/extensions/azure.coding-agent/README.md
@@ -80,7 +80,7 @@ This can happen if your azd login token has expired. You can fix this by logging
 azd auth login
 ```
 
-You might also get this error if you have not run azd init first.  To fix this error run azd init and go through the setup wizard.  After this is completed them you should be able to run azd coding-agent config.
+You might also get this error if you have not run `azd init` first. To fix this, run `azd init` and go through the setup wizard. After this is completed, you should be able to run `azd coding-agent config`.
 
 ### Must have admin rights to Repository
 


cli/azd/extensions/azure.coding-agent/README.md
```diff
@@ -80,6 +80,8 @@ This can happen if your azd login token has expired. You can fix this by logging
 azd auth login
 ```
 
+You might also get this error if you have not run `azd init` first. To fix this, run `azd init` and go through the setup wizard. After this is completed, you should be able to run `azd coding-agent config`.
+
```
vhvb1989, line 84: @richardpark-msft , can we make the extension to return this as part of the error?  or even make the extension to invoke azd init ?  

Looking for a way to shift left this from happening to preventing

richardpark-msft: From talking with @bob-okeefe, it sounded like the steps he went through were:

1. azd login
2. Run the coding-agent extension
3. Error back from AZD cred (internally) that says the refresh token is expired

So then he `azd init`'d and it worked.

Does azd init do something differently from azd login?

vhvb1989: Interesting.
It might make a difference if you were asked to pick a subscription - or if the `azd init` invoked the `azd auth login` 

Do we have more details about the number 3?  I what was trying to use the credential?  

Any ideas @weikanglim ?  -

But then, @richardpark-msft , there's not expectation from the extension around running `azd init` first?  If not, we should not add this note and instead make this an issue/bug to investigate (try to repro).

bob-okeefe: I was always asked to pick a subscription when I ran azd coding-agent config.  I would choose my subscription and my repo.  However when I got to the MI step and said to create a new MI it always failed.  I was only able to resolve this after running azd init.

See attached.
image

When using debug I got this:
image

bob-okeefe: BTW, I only figured out that I had to run azd init because when I was using Copilot to troubleshoot this it suggested running azd env set AZURE_SUBSCRIPTION_ID <subid>.  When I did this I got this error:

ERROR: no project exists; to create a new project, run `azd init`

After the azure.yml file was created I was able to run azd coding-agent config successfully

richardpark-msft: Apparently those errors are coming from when I call into the 'azd' PromptClient/PromptServiceClient:

* [PromptClient call](https://github.com/Azure/azure-dev/blob/a4da0cd8242fe1ff2703144e5e1b1152b71eb7eb/cli/azd/extensions/azure.coding-agent/internal/cmd/coding_agent_config.go#L199)
* [PromptServiceClient call](https://github.com/Azure/azure-dev/blob/a4da0cd8242fe1ff2703144e5e1b1152b71eb7eb/cli/azd/extensions/azure.coding-agent/internal/cmd/coding_agent_config.go#L584)

I was wondering, at first, if perhaps I was messing up some subscription or passing in a bad credential, but this is within 'azd' itself.

richardpark-msft: > But then, @richardpark-msft , there's not expectation from the extension around running `azd init` first? If not, we should not add this note and instead make this an issue/bug to investigate (try to repro).

There _isn't_ - I remember us talking about this. I don't use any azd logic to detect or use any project, so it should be safe.

JeffreyCA: Do you happen to have subscriptions across multiple tenants? I wonder if doing `azd auth login --tenant <tenant-id>` would've helped in this case. It's odd you still got an expired token error right after running `azd auth login`.

vhvb1989: did you ended up running `azd env set AZURE_SUBSCRIPTION_ID xxxxxx` ?
If yes, that's how it was fixed.
You had to run `azd init ...` but only to be able to run `azd env set ...` - to have an azd env where to set the sub-id.

A similar approach is to set AZURE_SUBSCRIPTION_ID as system env var (not at azd's env) and that would also work.

The issue is that azd is trying to use a token for the wrong tenant. When you set the sub-id, azd resolves the right tenant where that sub belongs to.

Even thought the extension is asking for a subscription, azd is not using the response to use that sub/tenant for the following events - instead it is using the credential provider which resolves to use a different token.  
We need to make the extension to use the subId selected to re-set the credential provider - I'll see if that can be done (time to chat to the AI ...)  - I'll be back to you...

vhvb1989: Confirmed the hypothesis — the extension uses `Subscription.TenantId` (resource tenant) instead of `Subscription.UserTenantId` (user access tenant) when creating the `AzureDeveloperCLICredential`. For multi-tenant/guest users these differ, causing "refresh token expired" errors.

Filed #7077 with fix at #7078.

wbreza (approved):
## 📋 Code Review — PR #7070

Thanks for surfacing this user pain point! The troubleshooting note is helpful.

One thing to note: the review thread identified the root cause as a tenant ID mismatch in the extension (using `Subscription.TenantId` instead of `Subscription.UserTenantId`). The `azd init` workaround helps because it sets `AZURE_SUBSCRIPTION_ID`. Once the underlying bug is fixed, this note may need updating.

rajeshkamal5050:
@vhvb1989 should we closse this PR in favor of https://github.com/Azure/azure-dev/pull/7080 ?

PR #7157, opened 2026-03-16
fix: handle ARM expression strings in preflight resource fields

### LINKED ISSUE #7154: AZD New Version Issue with Fabric (Agentic App)
- [ ] Make sure you've installed the latest version using [instructions](https://github.com/Azure/azure-dev?tab=readme-ov-file#installupgrade-azure-developer-cli)

**Output from `azd version`**
Run `azd version` and copy and paste the output here: 1.23.9

**Describe the bug**
Issue with the latest update of AZD 1.23.9

**To Reproduce**
With Azd 1.23.9 while deploying using the command "AZD UP", the below error is being produced.

  (x) Failed: Validating deployment

ERROR: error executing step command 'provision': deployment failed: error deploying infrastructure: local preflight validation failed: parsing bicep snapshot: json: cannot unmarshal string into Go struct field armTemplateResource.predictedResources.identity of type bicep.armTemplateIdentity

**Expected behavior**
After "AZD UP", rg should be created as expected and continue with other deployment steps.

vhvb1989:
## Problem

When running `azd up` on Bicep templates that use conditional expressions for resource fields (e.g. `identity: condition ? {...} : {...}`), local preflight validation fails with:

```
parsing bicep snapshot: json: cannot unmarshal string into Go struct field
armTemplateResource.predictedResources.identity of type bicep.armTemplateIdentity
```

Bicep compiles conditional expressions into ARM expression strings like `"[if(equals(...))]"` instead of the expected typed JSON objects. The strongly-typed Go struct fields (`*armTemplateIdentity`, `*armTemplateSKU`, etc.) cannot unmarshal these strings.

## Solution

Introduce a generic `armField[T]` type that stores raw JSON bytes and defers parsing:

- `Value() (T, bool)` — typed access; returns `false` if the JSON is an ARM expression string
- `Raw() json.RawMessage` — raw JSON bytes, always works regardless of shape
- `HasValue() bool` — checks if the field was present and non-null

This gives callers the best of both worlds: fast typed access when the value is a real object, and graceful fallback when it's an ARM expression.

Applied to these fields in `armTemplateResource`:
| Field | Before | After |
|-------|--------|-------|
| `Tags` | `json.RawMessage` | `armField[map[string]string]` |
| `SKU` | `*armTemplateSKU` | `armField[armTemplateSKU]` |
| `Plan` | `*armTemplatePlan` | `armField[armTemplatePlan]` |
| `Identity` | `*armTemplateIdentity` | `armField[armTemplateIdentity]` |
| `Copy` | `*armTemplateCopy` | `armField[armTemplateCopy]` |
| `Zones` | `[]string` | `armField[[]string]` |

`Properties` remains `json.RawMessage` (no single typed target), `Condition` remains `any` (already flexible).

## Testing

- 7 new unit tests covering typed values, expression strings, null, absent fields, tags, and round-trip marshaling
- All 47 existing bicep package tests pass unchanged

Fixes #7154

### DIFF
--- a/cli/azd/pkg/infra/provisioning/bicep/local_preflight.go
+++ b/cli/azd/pkg/infra/provisioning/bicep/local_preflight.go
@@ -4,6 +4,7 @@
 package bicep
 
 import (
+	"bytes"
 	"context"
 	"encoding/json"
 	"fmt"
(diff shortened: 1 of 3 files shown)


cli/azd/pkg/infra/provisioning/bicep/local_preflight.go
```diff
@@ -32,34 +89,34 @@ type armTemplateResource struct {
...
-	Tags json.RawMessage `json:"tags,omitempty"`
+	// Tags holds resource tags, which may be a map[string]string literal or an ARM expression string.
+	Tags armField[map[string]string] `json:"tags,omitzero"`
 	// DependsOn lists symbolic names or resource IDs of resources that must be deployed first.
 	DependsOn []string `json:"dependsOn,omitempty"`
 	// Kind is the resource kind (e.g. "StorageV2" for storage or "app,linux" for web apps).
 	Kind string `json:"kind,omitempty"`
 	// SKU is the pricing tier / SKU for the resource.
-	SKU *armTemplateSKU `json:"sku,omitempty"`
+	SKU armField[armTemplateSKU] `json:"sku,omitzero"`
```
weikanglim, line 99: Is it possible for `sku.capacity`, currently `*int` to be an expression as well?

cli/azd/pkg/infra/provisioning/bicep/local_preflight.go
```diff
@@ -32,34 +89,34 @@ type armTemplateResource struct {
...
 	// Plan is the marketplace plan for the resource.
-	Plan *armTemplatePlan `json:"plan,omitempty"`
+	Plan armField[armTemplatePlan] `json:"plan,omitzero"`
 	// Identity is the managed identity configuration for the resource.
-	Identity *armTemplateIdentity `json:"identity,omitempty"`
-	// Properties is the resource-specific configuration.
+	Identity armField[armTemplateIdentity] `json:"identity,omitzero"`
```
weikanglim, line 103: Wondering if `identity.userAssignedIdentities` could be an expression as well.

cli/azd/pkg/infra/provisioning/bicep/local_preflight.go
vhvb1989: Maybe even a resource name can be an expression based on another resource outputs 😆
But with this change I'm just preventing azd from failing and just letting folks to try parsing when needed. 
We will learn more about it in the future.

Issue #7173, opened 2026-03-17
CurrentPrincipalId returns home-tenant oid for guest/B2B users, causing incorrect identity in resource tenant


vhvb1989:
## Problem

`CurrentPrincipalId` (in `current_principal_id_provider.go`) resolves the current user's object ID by:
1. Calling `LookupTenant(subscriptionId)` → returns `UserAccessTenantId` (the user's **home tenant**)
2. Calling Graph `/me` against that tenant → returns the **home-tenant oid**

For B2B/guest users, the home-tenant oid is **different** from the guest oid in the resource tenant. All callers that use this principal ID to interact with the resource tenant's ARM APIs receive the wrong identity.

## Impact

All callers of `CurrentPrincipalId` are affected:

| Caller | File | Impact |
|---|---|---|
| Preflight role check | `bicep_provider.go:2220` | `assignedTo()` filter uses wrong oid → false positive warning |
| `loadParameters` | `bicep_provider.go:1800` | `AZURE_PRINCIPAL_ID` in ARM params → role assignments target wrong principal |
| bicepparam compile | `bicep_provider.go:1980` | Same as above |
| Terraform provider | `terraform_provider.go:722` | Principal ID in tfvars → wrong identity |
| Key Vault RBAC | `cmd/env.go:612` | Role assignment on Key Vault targets wrong principal |

## Root Cause

`LookupTenant` returns `sub.UserAccessTenantId` (correct for **credential creation**) but callers also use the same tenant for **identity resolution** via Graph `/me`. For guest users:
- Home tenant → Graph `/me` returns `home-oid` (e.g. `4070b897-...`)
- Resource tenant → Graph `/me` returns `guest-oid` (different value)

The `UserAccessTenantId` vs `TenantId` distinction (`account.Subscription` model) is documented as:
- `UserAccessTenantId`: tenant under which the user authenticates (home tenant)
- `TenantId`: tenant that owns the subscription (resource tenant)

## Proposed Fix

`CurrentPrincipalId` (or `GetCurrentPrincipalId`) should resolve the oid in the **resource tenant** context when the principal ID will be used for ARM operations in that tenant. This may require:
1. Calling Graph `/me` against `sub.TenantId` instead of `sub.UserAccessTenantId`, or
2. Extracting the oid from an ARM token scoped to the resource tenant, or
3. Providing both oids and letting callers choose

Care must be taken not to break credential creation flows, which correctly use `UserAccessTenantId`.

## Related

[...]

PR #7174, opened 2026-03-17
fix: resolve principal ID in resource tenant for preflight role check

### LINKED ISSUE #7169: Preflight permissions validation is ignoring RBAC Admin role
When executing `azd up`, there is a warning if the user is not 'User Access Administrator' or 'Owner'. However, the built-in role [Role Based Access Control Administrator
](https://learn.microsoft.com/en-us/azure/role-based-access-control/built-in-roles/privileged#role-based-access-control-administrator) is the least privileged role to perform 'Microsoft.Authorization/roleAssignments/write'.
 
Link to issue in code:
https://github.com/Azure/azure-dev/blob/0d0c25d9ae6125fec7ce87e4d9905a7a1949825b/cli/azd/pkg/infra/provisioning/bicep/bicep_provider.go#L2247

vhvb1989:
## Problem

Three issues with the preflight role assignment permission check:

### 1. MSAL credential ignores target tenant for B2B/guest users
`azdCredential.GetToken()` never forwarded the tenant ID to MSAL's `AcquireTokenSilent`. For B2B/guest users, this caused MSAL to return a cached home-tenant token even when a resource-tenant credential was requested. Graph `/me` then returned the home-tenant oid instead of the guest oid in the resource tenant.

### 2. RBAC evaluation merges NotActions globally across roles
`HasRequiredPermissions` collected Actions and NotActions from **all** role definitions into a single set, then evaluated them together. This is incorrect — Azure evaluates Actions−NotActions **per role definition**, then takes the union. The result: Contributor's `NotActions: [Microsoft.Authorization/*/Write]` cancelled User Access Administrator's `Actions: [Microsoft.Authorization/roleAssignments/*]`, producing a false positive warning.

### 3. No detection of ABAC conditions on role assignments
Role assignments can have ABAC conditions (e.g., restricting which role definitions can be assigned). The check treated conditional grants the same as unconditional ones, missing cases where the deployment would fail due to condition restrictions.

### 4. Warning message missing 'RBAC Administrator' role (#7169)
The warning only mentioned 'User Access Administrator' and 'Owner', but the built-in 'Role Based Access Control Administrator' role also grants `Microsoft.Authorization/roleAssignments/write`.

## Solution

**Tactical fix** scoped to the preflight check. The broader `CurrentPrincipalId` fix across all callers is tracked in #7173.

### Credential tenant forwarding (`auth/azd_credential.go`, `auth/manager.go`)
- `azdCredential` now stores a `tenantID` and forwards it to MSAL via `WithTenantID()` in `AcquireTokenSilent`
- `options.TenantID` (e.g., from CAE challenges) overrides the stored tenant
- All 4 existing `newAzdCredential` call sites pass `""` (no behavior change); only the cross-tenant path now passes the target tenant

### Per-role RBAC evaluation + ABAC detection (`azapi/permissions.go`)
- Evaluate Actions−NotActions **per role definition**, then union — matching Azure RBAC semantics
- `HasRequiredPermissions` now returns `PermissionCheckResult{HasPermission, Conditional}` instead of `bool`
- Tracks whether all granting role assignments have ABAC conditions; surfaces a distinct conditional warning

### Resource-tenant principal resolution (`bicep/bicep_provider.go`)
- Added `resolveResourceTenantPrincipalId()` that calls Graph `/me` against the subscription's resource tenant (`sub.TenantId`) to get the correct guest oid
- On failure, skips the check (logs the error) rather than using a home-tenant oid that would produce false positives
- Added 'Role Based Access Control Administrator' to the warning message
- Added conditional warning for ABAC-restricted role assignments

## Testing

- Credential tenant forwarding tested in `azd_credential_test.go` (3 cases: stored tenant, override, empty)
- Per-role RBAC evaluation tested in `permissions_test.go` (8 single-role cases + multi-role union test)
- Build, lint, and cspell clean

Fixes #7169
See #7173

### DIFF
--- a/cli/azd/pkg/auth/azd_credential_test.go
+++ b/cli/azd/pkg/auth/azd_credential_test.go
@@ -41,24 +41,30 @@ func TestAzdCredential_GetToken_ForwardsTenantID(t *testing.T) {
 		optsTID string // tenant from TokenRequestOptions
 		// expectTenantOption is true when WithTenantID should be in the options
 		expectTenantOption bool
+		// expectedTenantID is the tenant ID value that should be forwarded.
+		// When optsTID is set it overrides credTID.
+		expectedTenantID string
 	}{
 		{
 			name:               "CredentialTenantUsed",
 			credTID:            "resource-tenant-id",
 			optsTID:            "",
 			expectTenantOption: true,
+			expectedTenantID:   "resource-tenant-id",
 		},
 		{
 			name:               "OptionsTenantOverrides",
 			credTID:            "resource-tenant-id",
 			optsTID:            "override-tenant-id",
 			expectTenantOption: true,
+			expectedTenantID:   "override-tenant-id",
 		},
 		{
 			name:               "NoTenant",
 			credTID:            "",
 			optsTID:            "",
 			expectTenantOption: false,
+			expectedTenantID:   "",
 		},
 	}
 
@@ -81,6 +87,17 @@ func TestAzdCredential_GetToken_ForwardsTenantID(t *testing.T) {
 				require.Len(t, spy.silentOptions, 2,
 					"expected WithSilentAccount + WithClaims only")
 			}
+
+			// Verify the resolved tenant ID matches expectations.
+			// The credential resolves: optsTID if non-empty, else credTID.
+			// MSAL's WithTenantID option is opaque, so we verify the credential's
+			// tenant resolution logic directly (same package gives access to internals).
+			resolvedTenant := cred.tenantID
+			if tt.optsTID != "" {
+				resolvedTenant = tt.optsTID
+			}
+			require.Equal(t, tt.expectedTenantID, resolvedTenant,
+				"resolved tenant ID should match expected value")
 		})
 	}
 }
--- a/cli/azd/pkg/auth/manager.go
+++ b/cli/azd/pkg/auth/manager.go
@@ -325,7 +325,7 @@ func (m *Manager) CredentialForCurrentUser(
 		for i, account := range accounts {
 			if account.HomeAccountID == *currentUser.HomeAccountID {
 				if options.TenantID == "" {
-					return newAzdCredential(m.publicClient, &accounts[i], m.cloud, ""), nil
+					return newAzdCredential(m.publicClient, &accounts[i], m.cloud, "" /* tenantID */), nil
 				} else {
 					newAuthority := m.cloud.Configuration.ActiveDirectoryAuthorityHost + options.TenantID
 
@@ -742,7 +742,7 @@ func (m *Manager) LoginInteractive(
 		_ = os.Remove(claimsFile)
 	}
 
-	return newAzdCredential(m.publicClient, &res.Account, m.cloud, ""), nil
+	return newAzdCredential(m.publicClient, &res.Account, m.cloud, "" /* tenantID */), nil
 }
 
 // LoginWithBrokerAccount logs in an account provided by the system authentication broker via OneAuth.
@@ -852,7 +852,7 @@ func (m *Manager) LoginWithDeviceCode(
 		_ = os.Remove(claimsFile)
 	}
 
-	return newAzdCredential(m.publicClient, &res.Account, m.cloud, ""), nil
+	return newAzdCredential(m.publicClient, &res.Account, m.cloud, "" /* tenantID */), nil
 
 }
 
--- a/cli/azd/pkg/azapi/permissions.go
+++ b/cli/azd/pkg/azapi/permissions.go
@@ -62,10 +62,10 @@ func (s *PermissionsService) HasRequiredPermissions(
 	subscriptionId string,
 	principalId string,
 	requiredActions []string,
-) (*PermissionCheckResult, error) {
+) (PermissionCheckResult, error) {
 	credential, err := s.credentialProvider.CredentialForSubscription(ctx, subscriptionId)
 	if err != nil {
-		return nil, fmt.Errorf("getting credential for subscription %s: %w", subscriptionId, err)
+		return PermissionCheckResult{}, fmt.Errorf("getting credential for subscription %s: %w", subscriptionId, err)
 	}
 
 	// Create a role assignments client to list the principal's role assignments at subscription scope.
@@ -73,13 +73,13 @@ func (s *PermissionsService) HasRequiredPermissions(
 		subscriptionId, credential, s.armClientOptions,
 	)
 	if err != nil {
-		return nil, fmt.Errorf("creating role assignments client: %w", err)
+		return PermissionCheckResult{}, fmt.Errorf("creating role assignments client: %w", err)
 	}
 
 	// Create a role definitions client to retrieve the definition for each assignment.
 	roleDefinitionsClient, err := armauthorization.NewRoleDefinitionsClient(credential, s.armClientOptions)
 	if err != nil {
-		return nil, fmt.Errorf("creating role definitions client: %w", err)
+		return PermissionCheckResult{}, fmt.Errorf("creating role definitions client: %w", err)
 	}
 
 	// Collect role assignments with metadata about conditions.
@@ -96,7 +96,7 @@ func (s *PermissionsService) HasRequiredPermissions(
 	for pager.More() {
 		page, err := pager.NextPage(ctx)
 		if err != nil {
-			return nil, fmt.Errorf(
+			return PermissionCheckResult{}, fmt.Errorf(
 				"listing role assignments for principal %s: %w", principalId, err)
 		}
 		for _, ra := range page.Value {
@@ -112,7 +112,7 @@ func (s *PermissionsService) HasRequiredPermissions(
 	}
 
 	if len(assignments) == 0 {
-		return &PermissionCheckResult{HasPermission: false}, nil
+		return PermissionCheckResult{HasPermission: false}, nil
 	}
 
 	// Check each role definition and track whether granting roles have conditions.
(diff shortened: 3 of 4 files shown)


cli/azd/pkg/azapi/permissions.go
```diff
@@ -81,106 +92,130 @@ func (s *PermissionsService) HasRequiredPermissions(
...
 		if err != nil {
-			return false, fmt.Errorf("listing role assignments for principal %s: %w", principalId, err)
+			return nil, fmt.Errorf(
+				"listing role assignments for principal %s: %w", principalId, err)
 		}
 		for _, ra := range page.Value {
 			if ra.Properties != nil && ra.Properties.RoleDefinitionID != nil {
-				roleDefinitionIDs = append(roleDefinitionIDs, *ra.Properties.RoleDefinitionID)
+				hasCondition := ra.Properties.Condition != nil &&
+					*ra.Properties.Condition != ""
```
tg-msft, line 105: Is it worth pulling the condition into the struct so we can list it in the warning we give?  That might make it easier to ignore or pause when someone hits it.  We'd be giving slightly less of a false positive (i.e., imagine someone's got a benign condition but freaks out at our warning and spends half an hour investigating).

We could also potentially evolve the rule in the future to ask an agent to evaluate the liklihood of the condition being true for the operation we're running - if we don't want to build a whole expression evaluator.

cli/azd/pkg/azapi/permissions.go
```diff
@@ -81,106 +92,130 @@ func (s *PermissionsService) HasRequiredPermissions(
...
+) (*PermissionCheckResult, error) {
+	// Track which required actions are still unresolved, and whether any granting
+	// role is unconditional (no ABAC condition).
+	remaining := make(map[string]bool, len(requiredActions))
+	for _, a := range requiredActions {
+		remaining[a] = true
+	}
+
+	// hasUnconditionalGrant tracks whether at least one granting role has no condition.
+	hasUnconditionalGrant := false
```
JeffreyCA: Copilot identified a potential false positive - can you double-check?

`cli/azd/pkg/azapi/permissions.go:146-197` — `checkActionsFromRoles` can incorrectly return `Conditional: true` when a required action is granted by both a conditional role assignment and a later unconditional one. The first matching role deletes the action from remaining, so later roles are never checked for that action; if the conditional role is seen first, the unconditional grant is missed and the warning becomes a false positive. This needs a test for the “conditional then unconditional” ordering, and the logic should track unconditional grants per required action rather than via one global flag.

cli/azd/pkg/auth/manager.go
```diff
@@ -325,7 +325,7 @@ func (m *Manager) CredentialForCurrentUser(
 		for i, account := range accounts {
 			if account.HomeAccountID == *currentUser.HomeAccountID {
 				if options.TenantID == "" {
-					return newAzdCredential(m.publicClient, &accounts[i], m.cloud), nil
+					return newAzdCredential(m.publicClient, &accounts[i], m.cloud, ""), nil
```
tg-msft: Style nit - consider commenting constants when they're not really obvious.

```suggestion
					return newAzdCredential(m.publicClient, &accounts[i], m.cloud, /* tenantID */ ""), nil
```

cli/azd/pkg/azapi/permissions.go
```diff
@@ -81,106 +92,130 @@ func (s *PermissionsService) HasRequiredPermissions(
...
+			return nil, fmt.Errorf(
+				"getting role definition %s: %w", assignment.roleDefinitionID, err)
 		}
 
 		if resp.Properties == nil || resp.Properties.Permissions == nil {
 			continue
 		}
 
+		// Collect this role's actions and notActions.
+		var actions, notActions []string
```
wbreza, line 165: **[Medium]** `hasUnconditionalGrant` global flag is inaccurate with multiple required actions

`hasUnconditionalGrant` is a single boolean for ALL required actions. If `requiredActions` contains `[A, B]` where action A is granted by a conditional role and action B by an unconditional role, the flag becomes `true` — marking the entire result as `Conditional: false`, hiding the ABAC condition on action A.

Currently **no real-world impact** since the only caller passes a single-action slice. However, the API accepts `[]string`, making this a latent correctness issue for future callers.

**Suggested Fix:**
```suggestion
	hasUnconditionalGrant := false
```
Consider tracking conditionality per required action instead of a single global flag:
```go
unconditionalActions := make(map[string]bool)
// In inner loop: unconditionalActions[action] = true
// Final: Conditional: len(unconditionalActions) < len(requiredActions)
```

cli/azd/pkg/auth/azd_credential_test.go
```diff
@@ -0,0 +1,86 @@
...
+			optsTID:            "override-tenant-id",
+			expectTenantOption: true,
+		},
+		{
+			name:               "NoTenant",
+			credTID:            "",
+			optsTID:            "",
+			expectTenantOption: false,
+		},
+	}
```
wbreza, line 69: **[Medium]** Test spy verifies option count, not actual tenant ID value

Tests assert that `WithTenantID` was added by checking option slice length (2 vs 3), but never verify which tenant ID value was forwarded. The "OptionsTenantOverrides" case should confirm the override `"override-tenant-id"` is used instead of the stored `"resource-tenant-id"` — a tenant-forwarding bug would pass silently.

Consider enhancing the spy to capture the actual tenant ID value passed to `WithTenantID()`.

cli/azd/pkg/infra/provisioning/bicep/bicep_provider.go
```diff
@@ -2237,24 +2249,69 @@ func (p *BicepProvider) checkRoleAssignmentPermissions(
...
+	}
+
 	return nil, nil
 }
 
+// resolveResourceTenantPrincipalId returns the current user's object ID as seen in the
+// subscription's resource tenant. This is needed because B2B/guest users have a different
+// object ID in the resource tenant than in their home tenant. Returns an error if the
+// resource tenant principal cannot be resolved; the caller should skip the check rather
+// than fall back to the home-tenant oid (which would produce false positive warnings).
```
wbreza, line 2291: **[Medium]** No tests for `resolveResourceTenantPrincipalId`

New function with 3 error paths (subscription lookup, service resolution, Graph `/me`) and 1 success path — all untested. This is the critical path for B2B/guest user support.

Consider adding tests with mocked `subscriptionManager`, `serviceLocator`, and `UserProfileService` covering each error and the happy path.

cli/azd/pkg/infra/provisioning/bicep/bicep_provider.go
```diff
@@ -2237,24 +2249,69 @@ func (p *BicepProvider) checkRoleAssignmentPermissions(
...
+		return &PreflightCheckResult{
+			Severity: PreflightCheckWarning,
+			Message: fmt.Sprintf(
+				"the current principal (%s) has conditional permission to create role "+
+					"assignments (Microsoft.Authorization/roleAssignments/write) on "+
+					"subscription %s. The role assignment that grants this permission "+
+					"has an ABAC condition that may restrict which roles can be assigned. "+
+					"The deployment may fail if the condition does not permit the "+
+					"specific role assignments in the template.",
+				principalId,
```
wbreza, line 2278: **[Medium]** No test for ABAC conditional warning path

The new `if hasPermission.Conditional { ... }` branch produces a distinct user-facing warning but has no test coverage. Consider adding a test verifying that when `HasRequiredPermissions` returns `{HasPermission: true, Conditional: true}`, the result is a `PreflightCheckWarning` with ABAC-related message content.

cli/azd/pkg/azapi/permissions.go
```diff
@@ -34,6 +34,17 @@ func NewPermissionsService(
 	}
 }
 
+// PermissionCheckResult describes the outcome of a permission check.
+type PermissionCheckResult struct {
+	// HasPermission is true when the required actions are granted by at least one role.
+	HasPermission bool
+	// Conditional is true when every role that grants the required action has an
+	// ABAC condition attached. Conditions restrict the scope of the action (e.g.,
```
wbreza, line 42: **[Low]** Consider returning `PermissionCheckResult` by value

`PermissionCheckResult` is 2 booleans (16 bytes). Pointer return forces nil-checking at call sites and adds GC pressure. Go convention favors value return for small immutable structs — `(PermissionCheckResult, error)` with zero-value for no-permission results would be more idiomatic.

cli/azd/pkg/azapi/permissions.go
vhvb1989: The condition might be more confusing than helpful for users.
I see myself as a user in my company's subscription looking at a condition which could be hard to digest and which I can't really do anything with.

On the very fist approach, I would be displaying the policy description to the user, like:

image

But you can see how the description looks to be written for someone managing rules and policies for the subscription and not really for final users. So I decided to remove it and tell the user only which is the resource which should be either updated or removed. Or that they need to follow with the Subscription Admin.

cli/azd/pkg/azapi/permissions.go
vhvb1989: Fixed — replaced the single `hasUnconditionalGrant` flag with a per-action `unconditionalActions` map. A later unconditional role now clears the conditional flag even when a conditional role granted the same action first. Added `TestCheckActionsFromRoles_ConditionalThenUnconditional` to cover this case.

cli/azd/pkg/auth/manager.go
vhvb1989: Applied — added `/* tenantID */` comment on the 3 call sites. Placed after the `""` as `"" /* tenantID */` so gofmt preserves it correctly.

cli/azd/pkg/azapi/permissions.go
vhvb1989: Fixed — replaced `hasUnconditionalGrant` with `unconditionalActions map[string]bool` exactly as suggested. Also added `TestCheckActionsFromRoles_MultipleActions_MixedConditionality` to verify the per-action tracking.

cli/azd/pkg/auth/azd_credential_test.go
vhvb1989: Enhanced — tests now include an `expectedTenantID` field and verify the resolved tenant value. Since MSAL's `WithTenantID` option is opaque (unexported closure), we verify the credential's tenant resolution logic directly via same-package access to `cred.tenantID`.

Issue #7177, opened 2026-03-18
Preflight check: detect Azure Policy blocking local authentication


vhvb1989:
## Problem

When a subscription has an Azure Policy that denies resources with local authentication enabled (e.g., `disableLocalAuth` must be `true`), `azd provision` fails at deploy time with a cryptic `RequestDisallowedByPolicy` error:

```
Deployment Error Details:
InvalidTemplateDeployment: The template deployment failed because of policy violation.
RequestDisallowedByPolicy: Resource 'stdagcewyyviu73h' was disallowed by policy.
Reasons: 'This request was denied due to internal policy. Local authentication methods
are not allowed.'
```

Users get no prior warning before the (often long) deployment attempt.

## Proposal

Add a local preflight check that:

1. Lists Azure Policy assignments on the target subscription
2. Fetches policy definitions and inspects `policyRule` for `disableLocalAuth`-related deny conditions
3. Cross-references the target resource types against the Bicep snapshot resources
4. Reports a **warning** if any template resource has `disableLocalAuth: false` or missing while a deny policy requires it

This fits naturally into the existing preflight infrastructure (`local_preflight.go` + `PreflightCheckFn`).

## Scope

- New `PolicyService` in `pkg/azapi/` using the `armpolicy` Azure SDK
- Lightweight parser for policy rule conditions (focused on `disableLocalAuth` field patterns)
- New preflight check function registered in `bicep_provider.go`
- Warning-level only (not blocking) since we cannot evaluate exemptions or complex conditions

## Limitations (acceptable for a warning)

- Cannot evaluate complex nested policy conditions (tags, locations, etc.)
- Cannot check policy exemptions
- Parameterized effects need resolution from assignment parameters
- Policy initiatives (sets) need an extra level of resolution

Issue #7119, opened 2026-03-13
Enhance validation result/context: Suggestion field, parameters.json, location, docs links


vhvb1989:
## Context
Follow-up from [PR #7053](https://github.com/Azure/azure-dev/pull/7053) review. Parent tracking issue: #7112

## PR Conversations
- [Add link to Azure RBAC docs in warning](https://github.com/Azure/azure-dev/pull/7053#discussion_r2928055251) (@tg-msft)
- [Improve doc comment for validation functions](https://github.com/Azure/azure-dev/pull/7053#discussion_r2928068950) (@tg-msft)
- [Check for .parameters.json too](https://github.com/Azure/azure-dev/pull/7053#discussion_r2928074057) (@tg-msft)
- [Pass location to bicep snapshot](https://github.com/Azure/azure-dev/pull/7053#discussion_r2928082750) (@tg-msft)
- [Add Suggestion field to PreflightResult](https://github.com/Azure/azure-dev/pull/7053#discussion_r2928089768) (@tg-msft)

## What needs to change
1. **Suggestion field**: Add a `Suggestion string` field to `PreflightResult` to mirror the final UI pattern for errors (consistent with `ErrorWithSuggestion`).
2. **Doc comment**: Update the validation function doc comment to emphasize graceful failure — checks should not block deployment if a condition is not met or a value is missing.
3. **Parameters file**: Also check for `.parameters.json` file when loading bicep parameters.
4. **Location**: Pass `location` to `bicep snapshot` command if known, so it can resolve `resourceGroup().location`.
5. **RBAC docs link**: Add a link to [Azure RBAC role assignments](https://learn.microsoft.com/azure/role-based-access-control/role-assignments-portal) in the role-assignment warning message to help new users fix the issue.

## Files to update
- `cli/azd/pkg/infra/provisioning/bicep/local_preflight.go`
- `cli/azd/pkg/infra/provisioning/bicep/bicep_provider.go`

spboyer:
Telemetry data to help prioritize location validation.

From a Kusto analysis of `service.arm.deployment.failed` in March 2026 (23 days of data):

`InvalidResourceGroupLocation` is the inner error code for 350 ARM deployment failures per month (3.9% of all ARM deployment failures). These users wait an average of 5.3 minutes before receiving a region-incompatibility error that could be detected in seconds with a preflight region check.

The failures break down by service host type:

| Service Host | Conflict Failures | BadRequest Failures |
|---|---|---|
[...]

Issue #7115, opened 2026-03-13
Never abort deployment on validation errors - always let user choose


vhvb1989:
## Context
Follow-up from [PR #7053](https://github.com/Azure/azure-dev/pull/7053) review. Parent tracking issue: #7112

## PR Conversations
- [Should never abort deployment](https://github.com/Azure/azure-dev/pull/7053#discussion_r2926673536) (@tg-msft, @vhvb1989)
- [Remove abort code for errors](https://github.com/Azure/azure-dev/pull/7053#discussion_r2928012020) (@tg-msft)

## What needs to change
1. Remove code that blocks/aborts deployment when validation errors are found.
2. Always ask the user whether to stop and fix or try to deploy anyway, even for errors that seem certain to fail.
3. Validation should be strictly informational with ignorable warnings — forward compatibility concerns mean we should never block based on heuristics that could become stale.

## Files to update
- `cli/azd/pkg/infra/provisioning/bicep/bicep_provider.go` (line ~2185)
- `cli/azd/docs/design/local-preflight-validation.md` (update design to reflect non-blocking behavior)

spboyer:
Telemetry data supporting the "never abort" direction.

From Kusto analysis of `service.arm.deployment.failed` in March 2026 (23 days):

`RoleAssignmentExists` accounts for 351 ARM deployment failures per month (4.0% of inner error codes). These are not real failures -- the role assignment already succeeded on a previous run. The current behavior treats an idempotent success state as a deployment-blocking error.

Additionally, `DeploymentActive` (199/mo) represents cases where a previous deployment is still running. Rather than aborting, azd could detect this and wait for the in-progress deployment to complete.

Both of these support treating certain "errors" as warnings rather than abort conditions:

| Inner Error | Monthly Count | Current Behavior | Proposed Behavior |
|---|---|---|---|
| RoleAssignmentExists | 351 | Deployment fails | Treat as success (already done) |
| DeploymentActive | 199 | Deployment fails | Wait for in-progress deployment, then proceed |

Data source: `ddazureclients.kusto.windows.net` / `DevCli` / `RawEventsAppRequests`, March 1-23, 2026.

vhvb1989:
Feature is already working like this.
Warnings prompt users if they want to continue or not

PR #7179, opened 2026-03-18
docs: local preflight — storage account policy check investigation

vhvb1989:
## Overview

Documents the investigation into detecting Azure Policy restrictions on storage
accounts before deployment. After evaluating multiple approaches, the check was
not implemented due to fundamental API limitations.

## Changes

- **Move** `docs/design/local-preflight-validation.md` → `docs/local-preflight/README.md`
- **Add** "Investigated Checks" section to the README with a table of checks that
  were explored but not shipped
- **Add** `docs/local-preflight/storage-account-policy-check.md` — detailed
  investigation writeup covering:
  - Client-side `armpolicy` parsing (sees management-group policies but cannot
    evaluate ARM expressions → false positives)
  - Server-side `checkPolicyRestrictions` API (evaluates all conditions but does
    not see management-group-inherited policies → misses real denials)
  - `policyStates` API (only evaluates existing resources, not hypothetical ones)

## Why the check was not shipped

Enterprise deny policies for storage accounts almost always originate from
management groups. The two available approaches each fail in a different way:

| Approach | Sees MG policies | Evaluates all conditions | Result |
|---|---|---|---|
| Client-side `armpolicy` parsing | ✅ | ❌ (ARM expressions) | False positives |
| Server-side `checkPolicyRestrictions` | ❌ | ✅ | Misses real denials |

See the [investigation doc](cli/azd/docs/local-preflight/storage-account-policy-check.md)
for full details, including specific failure scenarios with example policies.

Relates to #7177

### DIFF
--- a/cli/azd/pkg/infra/provisioning/bicep/local_preflight.go
+++ b/cli/azd/pkg/infra/provisioning/bicep/local_preflight.go
@@ -265,6 +265,10 @@ const (
 type PreflightCheckResult struct {
 	// Severity indicates whether this result is a warning or a blocking error.
 	Severity PreflightCheckSeverity
+	// DiagnosticID is a unique, stable identifier for this specific finding type
+	// (e.g. "role_assignment_missing"). Used in telemetry to correlate actioned
+	// warnings with deployment outcomes and to track error frequency over time.
+	DiagnosticID string
 	// Message is a human-readable description of the finding.
 	Message string
 }
@@ -300,6 +304,17 @@ type PreflightCheckFn func(
 	valCtx *validationContext,
 ) (*PreflightCheckResult, error)
 
+// PreflightCheck pairs a unique rule identifier with its check function.
+// The RuleID is a stable, unique string used in telemetry to identify which rule
+// produced a result (e.g. for crash tracking). Each rule may emit results with
+// different DiagnosticIDs to distinguish specific finding types.
+type PreflightCheck struct {
+	// RuleID is a unique, stable identifier for the rule (e.g. "role_assignment_permissions").
+	RuleID string
+	// Fn is the check function that performs the validation.
+	Fn PreflightCheckFn
+}
+
 // localArmPreflight provides local (client-side) validation of an ARM template before deployment.
 // It parses the template and parameters to build a comprehensive view of all resources that would
 // be deployed, enabling early detection of issues without making Azure API calls.
@@ -315,7 +330,7 @@ type localArmPreflight struct {
 	// target is the deployment scope (subscription or resource group) used to derive snapshot options.
 	// It may be nil, in which case snapshot options are left empty.
 	target infra.Deployment
-	checks []PreflightCheckFn
+	checks []PreflightCheck
 }
 
 // newLocalArmPreflight creates a new instance of localArmPreflight.
@@ -326,10 +341,10 @@ func newLocalArmPreflight(modulePath string, bicepCli *bicep.Cli, target infra.D
 	return &localArmPreflight{modulePath: modulePath, bicepCli: bicepCli, target: target}
 }
 
-// AddCheck registers a preflight check function to be executed during validate.
-// Check functions are invoked in the order they are added.
-func (l *localArmPreflight) AddCheck(fn PreflightCheckFn) {
-	l.checks = append(l.checks, fn)
+// AddCheck registers a preflight check to be executed during validate.
+// Checks are invoked in the order they are added.
+func (l *localArmPreflight) AddCheck(check PreflightCheck) {
+	l.checks = append(l.checks, check)
 }
 
 // validate performs local preflight validation on the given ARM template and parameters.
@@ -417,9 +432,9 @@ func (l *localArmPreflight) validate(
 
 	var results []PreflightCheckResult
 	for _, check := range l.checks {
-		result, err := check(ctx, valCtx)
+		result, err := check.Fn(ctx, valCtx)
 		if err != nil {
-			return results, fmt.Errorf("preflight check failed: %w", err)
+			return results, fmt.Errorf("preflight check %q failed: %w", check.RuleID, err)
 		}
 		if result != nil {
 			results = append(results, *result)


cli/azd/pkg/azapi/policy_service.go
```diff
@@ -0,0 +1,583 @@
...
+	resourceType string,
+) []LocalAuthDenyPolicy {
+	fieldVal, ok := condMap["field"].(string)
+	if !ok {
+		return nil
+	}
+
+	if !isLocalAuthField(fieldVal) {
+		return nil
+	}
```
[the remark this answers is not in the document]
vhvb1989: Intentionally kept as-is. For a warning-level check, being conservative (matching any mention of the field in a deny rule) is preferred. Policies that reference disableLocalAuth in a deny context are virtually always about enforcing it. The cost of a false positive warning is low (user clicks continue), while validating the exact operator/value pattern would require reimplementing more of the policy engine and risks false negatives.

cli/azd/pkg/azapi/policy_service.go
```diff
@@ -0,0 +1,583 @@
...
+		FieldPath:    fieldVal,
+	}}
+}
+
+// isLocalAuthField returns true if the field path references a local authentication property.
+func isLocalAuthField(field string) bool {
+	lower := strings.ToLower(field)
+	return strings.HasSuffix(lower, "/disablelocalauth") ||
+		strings.HasSuffix(lower, "/allowsharedkeyaccess") ||
+		strings.HasSuffix(lower, "/localauthenabled") ||
```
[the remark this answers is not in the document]
vhvb1989: Fixed — removed `localAuthEnabled` from `isLocalAuthField` to avoid the inconsistency. If a real policy uses that field name in the future, we can add support to both functions together.

cli/azd/pkg/infra/provisioning/bicep/bicep_provider.go
```diff
@@ -2255,6 +2260,74 @@ func (p *BicepProvider) checkRoleAssignmentPermissions(
...
+	for _, resource := range valCtx.SnapshotResources {
+		policy, found := policyByType[strings.ToLower(resource.Type)]
+		if !found {
+			continue
+		}
+
+		if !azapi.ResourceHasLocalAuthDisabled(resource.Type, resource.Properties) {
+			violations = append(violations, fmt.Sprintf(
+				"  - %s (type: %s) — policy: %q",
+				resource.Name, resource.Type, policy.PolicyName,
```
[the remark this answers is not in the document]
vhvb1989: Fixed — `policyByType` now stores `[]LocalAuthDenyPolicy` per resource type, and the violation message lists all matching policy names.

cli/azd/go.mod
```diff
@@ -91,6 +91,7 @@ require (
 
 require (
 	github.com/Azure/azure-sdk-for-go/sdk/internal v1.11.2 // indirect
+	github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/resources/armpolicy v0.10.0 // indirect
```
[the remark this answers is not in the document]
vhvb1989: Fixed — ran `go mod tidy`, armpolicy is now in the direct require block.

cli/azd/pkg/azapi/policy_service.go
```diff
@@ -0,0 +1,583 @@
...
+		if strings.EqualFold(fieldVal, "type") {
+			if eq, ok := condMap["equals"].(string); ok {
+				resourceType = eq
+			}
+			if in, ok := condMap["in"].([]any); ok && len(in) > 0 {
+				// Multiple resource types — take the first one for now.
+				if s, ok := in[0].(string); ok {
+					resourceType = s
+				}
+			}
```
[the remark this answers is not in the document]
vhvb1989: Fixed — `findInCompoundCondition` now collects all resource types from the `in` array and emits one `LocalAuthDenyPolicy` per type. Added test `TestExtractLocalAuthDenyPolicies_MultipleResourceTypesInArray`.

cli/azd/pkg/azapi/policy_service.go
```diff
@@ -0,0 +1,583 @@
...
+				ctx, definitionsClient, defID, assignmentName, assignmentParams,
+			)
+			results = append(results, policies...)
+		} else if isPolicySetDefinition(defID) {
+			// Policy set (initiative) — enumerate its member definitions.
+			policies := s.checkPolicySetDefinition(
+				ctx, setDefinitionsClient, definitionsClient, defID, assignmentName, assignmentParams,
+			)
+			results = append(results, policies...)
+		}
```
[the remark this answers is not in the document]
vhvb1989: Fixed — added `defCache` and `setDefCache` maps keyed by definition ID within `FindLocalAuthDenyPolicies`. Cached definitions are reused across assignments and within initiatives.

cli/azd/pkg/infra/provisioning/bicep/bicep_provider.go
```diff
@@ -2255,6 +2260,74 @@ func (p *BicepProvider) checkRoleAssignmentPermissions(
...
+	if len(violations) == 0 {
+		return nil, nil
+	}
+
+	return &PreflightCheckResult{
+		Severity: PreflightCheckWarning,
+		Message: fmt.Sprintf(
+			"subscription %s has Azure Policy assignments that deny resources with local "+
+				"authentication enabled. The following resources may be blocked:\n%s\n"+
+				"Set 'disableLocalAuth: true' on these resources or request a policy exemption. "+
```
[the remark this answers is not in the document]
vhvb1989: Fixed — warning message now reads: "Disable local authentication on these resources (e.g. set disableLocalAuth: true or allowSharedKeyAccess: false for storage accounts)".

cli/azd/pkg/azapi/policy_service_test.go
```diff
@@ -0,0 +1,400 @@
...
+		{"[concat('a', 'b')]", ""},
+		{"", ""},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.expr, func(t *testing.T) {
+			t.Parallel()
+			require.Equal(t, tt.want, extractParameterReference(tt.expr))
+		})
+	}
```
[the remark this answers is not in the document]
vhvb1989: Fixed — added whitespace test cases to `TestExtractParameterReference` (both spaces and tab/newline), `TestExtractLocalAuthDenyPolicies_MultipleResourceTypesInArray` for multi-type `in`, and `TestExtractLocalAuthDenyPolicies_NestedConditionInheritsResourceType` for nested inheritance.

cli/azd/pkg/azapi/policy_service.go
```diff
@@ -0,0 +1,583 @@
...
+			return nil, err
+		}
+		return &resp.Definition, nil
+	}
+
+	resp, err := client.Get(ctx, name, nil)
+	if err != nil {
+		return nil, err
+	}
+	return &resp.Definition, nil
```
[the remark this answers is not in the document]
vhvb1989: This is a known limitation accepted for this best-effort warning check. Management-group-scoped definitions that fail to fetch are silently skipped with a log message (existing `log.Printf` in `checkPolicyDefinition`). The subscription-scoped `Get` call actually does resolve many inherited definitions since Azure RP handles scope inheritance. For the remaining edge cases, the worst outcome is a missed warning — no worse than today where there is no check at all. Added a clearer log message would be a nice follow-up but not blocking.

cli/azd/pkg/azapi/policy_service.go
```diff
@@ -0,0 +1,583 @@
...
+	if !strings.HasPrefix(lower, "[parameters('") || !strings.HasSuffix(lower, "')]") {
+		return ""
+	}
+	// Extract between [parameters(' and ')]
+	inner := expr[len("[parameters('"):]
+	before, _, ok := strings.Cut(inner, "')]")
+	if !ok {
+		return ""
+	}
+	return before
```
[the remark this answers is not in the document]
vhvb1989: Fixed — now trims once into a local variable and uses it for both validation and extraction. Added whitespace test cases.

cli/azd/pkg/azapi/policy_service.go
```diff
@@ -0,0 +1,583 @@
...
+		}
+		results = append(results, checkSingleCondition(condMap, assignmentName, resourceType)...)
+
+		// Also recurse into nested conditions.
+		if allOf, ok := condMap["allOf"]; ok {
+			results = append(results, findInCompoundCondition(allOf, assignmentName)...)
+		}
+		if anyOf, ok := condMap["anyOf"]; ok {
+			results = append(results, findInCompoundCondition(anyOf, assignmentName)...)
+		}
```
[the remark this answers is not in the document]
vhvb1989: Fixed — `findInCompoundCondition` now accepts `parentResourceTypes []string` and passes resolved types through recursive calls. Added `TestExtractLocalAuthDenyPolicies_NestedConditionInheritsResourceType` to verify.

cli/azd/pkg/infra/provisioning/bicep/bicep_provider.go
```diff
@@ -2255,6 +2260,87 @@ func (p *BicepProvider) checkRoleAssignmentPermissions(
...
+		lines = append(lines, fmt.Sprintf("  - %s", t))
+	}
+
+	return &PreflightCheckResult{
+		Severity: PreflightCheckWarning,
+		Message: fmt.Sprintf(
+			"an Azure Policy on this subscription denies resources with local authentication enabled. "+
+				"The following resource types in this deployment may be blocked:\n%s\n"+
+				"Disable local authentication on these resources or request a policy exemption.",
+			strings.Join(lines, "\n"),
```
wbreza: **[Medium]** Warning message omits policy names - less actionable for users

The `policiesByType` map stores policy names per resource type, but the warning only lists affected resource types. Users can't easily identify *which* policy is blocking them.

Consider including policy names in the output to make the warning more actionable:
```go
for _, t := range typeList {
    policies := policiesByType[strings.ToLower(t)]
    names := make([]string, 0, len(policies))
    for _, p := range policies {
        names = append(names, p.PolicyName)
    }
    lines = append(lines, fmt.Sprintf("  - %s (policy: %s)", t, strings.Join(names, ", ")))
}
```

cli/azd/pkg/infra/provisioning/bicep/bicep_provider.go
```diff
@@ -2255,6 +2260,87 @@ func (p *BicepProvider) checkRoleAssignmentPermissions(
...
+		),
+	}, nil
+}
+
 // Deploys the specified Bicep module and parameters with the selected provisioning scope (subscription vs resource group)
 func (p *BicepProvider) deployModule(
 	ctx context.Context,
```
wbreza, line 2404: **[Medium]** No unit tests for `checkLocalAuthPolicy` integration logic

`PolicyService` parsing is well-tested (15+ cases), but this integration function has no test coverage. Untested paths: empty SnapshotResources exit, PolicyService resolution failure, API error fallback, no matching resource types, and warning message formatting.

Consider adding tests with mocked `PolicyService` and `SnapshotResources`.

cli/azd/pkg/infra/provisioning/bicep/bicep_provider.go
```diff
@@ -2156,6 +2156,11 @@ func (p *BicepProvider) validatePreflight(
 	// principal has the required write permission.
 	localPreflight.AddCheck(p.checkRoleAssignmentPermissions)
 
+	// Register the local auth policy check. This detects Azure Policy assignments
+	// that deny resources with local authentication enabled (disableLocalAuth != true)
+	// and warns if the template contains affected resources.
```
wbreza: **[Medium]** Merge conflict expected with PR #7174

Both PRs modify `bicep_provider.go` in adjacent areas. This PR adds `checkLocalAuthPolicy` after `checkRoleAssignmentPermissions`, while #7174 modifies `checkRoleAssignmentPermissions` and adds `resolveResourceTenantPrincipalId`. The changes are semantically independent - the merge should be straightforward, but the person merging second needs to be aware.

cli/azd/pkg/infra/provisioning/bicep/bicep_provider.go
```diff
@@ -2255,6 +2260,87 @@ func (p *BicepProvider) checkRoleAssignmentPermissions(
...
+
+	// Collect affected resource types (deduplicated).
+	affectedTypes := make(map[string]bool)
+	for _, resource := range valCtx.SnapshotResources {
+		_, found := policiesByType[strings.ToLower(resource.Type)]
+		if !found {
+			continue
+		}
+
+		if !azapi.ResourceHasLocalAuthDisabled(resource.Type, resource.Properties) {
```
wbreza: **[Low]** Duplicate iteration over SnapshotResources

`SnapshotResources` is iterated twice - once for diagnostic logging (lines 2313-2318) and once for collecting affected types (lines 2322-2331). These could be merged into a single loop.

cli/azd/pkg/azapi/policy_service.go
```diff
@@ -0,0 +1,668 @@
...
+	return &PolicyService{
+		credentialProvider: credentialProvider,
+		armClientOptions:   armClientOptions,
+	}
+}
+
+// FindLocalAuthDenyPolicies lists policy assignments on the subscription and inspects
+// their definitions for deny-effect rules that require disableLocalAuth to be true.
+// It returns a list of matching policies with their target resource types.
+func (s *PolicyService) FindLocalAuthDenyPolicies(
```
weikanglim: Is there a way we could separate the functionality for "API fetch for policies", separately from the engine that evaluates whether "local auth" is applicable?  

I'm imagining that this would help to avoid duplicate API fetches for a different set of policies (which seems that something we'd likely want to do).

cli/azd/pkg/azapi/policy_service.go
```diff
@@ -0,0 +1,668 @@
...
+	if !isDenyEffect(ruleMap, def.Properties.Parameters, assignmentParams) {
+		return nil
+	}
+
+	// Parse the "if" condition to find disableLocalAuth field references.
+	ifBlock, ok := ruleMap["if"]
+	if !ok {
+		return nil
+	}
+
```
weikanglim: Is there a way to check these evaluations via a server-side call? Otherwise, we are re-inventing the policy engine -- which will need to be maintained and kept-in sync as policy definitions evolve over time.

cli/azd/pkg/azapi/policy_service.go
vhvb1989: Turns out there is! :)

I have updated this validation check to use it and only for storage accounts

spboyer:
Telemetry data backing for expanding the preflight framework.

I ran a deep analysis on `service.arm.deployment.failed` from the `ddazureclients` Kusto cluster (March 2026, 23 days of data). The numbers:

- 11,627 total ARM deployment failures in March, affecting 3,296 unique machines
- 78% come from `provision`, 19% from `up` (which runs provision internally)
- Users wait an average of 5.3 minutes (median 2.4 min, P90 12.2 min) before getting an ARM failure

The top ARM error code is `Conflict` at 67.2% (7,478 failures). The inner error codes break down as:

| Inner Error Code | Count | Share |
|---|---|---|
| InvalidTemplateDeployment | 2,252 | 25.4% |
| BadRequest | 1,716 | 19.4% |
| ContainerAppOperationError | 906 | 10.2% |
| RoleAssignmentExists | 351 | 4.0% |
| InvalidResourceGroupLocation | 350 | 3.9% |
| RequestConflict | 278 | 3.1% |
| ResourceGroupNotFound | 274 | 3.1% |
| DeploymentActive | 199 | 2.2% |
| ProvisioningDisabled | 177 | 2.0% |

Your policy check addresses the `ProvisioningDisabled` bucket (177/mo). Two adjacent items that could fit the preflight framework:

1. `InvalidResourceGroupLocation` (350/mo) -- preflight could validate that the target region supports all required resource types before deploying. The snapshot data already contains the resource types needed. This aligns with issue #7119 which mentions location validation.

2. `DeploymentActive` (199/mo) -- preflight could check for in-progress deployments on the resource group via `GET /deployments?$filter=provisioningState eq 'Running'` and warn or wait. Users currently burn 5+ minutes before this fails.

Additionally, `RoleAssignmentExists` (351/mo) is a pure idempotency issue -- the role assignment already exists from a previous run. This should be treated as a non-fatal condition rather than a deployment failure. This supports the direction in issue #7115 ("never abort deployment on validation errors").

Data source: `ddazureclients.kusto.windows.net` / `DevCli` / `RawEventsAppRequests`, March 1-23, 2026.

spboyer:
## Deep Dive: Policy Errors Beyond Storage Accounts

Follow-up to the previous telemetry comment. A deeper analysis of `InvalidTemplateDeployment` error chains (March 2026, 23 days) shows policy errors are a bigger opportunity than the storage-only scope of this PR:

**`RequestDisallowedByPolicy` = 454 errors/month across 164 unique machines (9.6% of all InvalidTemplateDeployment errors).**

The policy errors are spread across all service types, not just storage:

| Service Host | Policy Errors |
|---|---|
| containerapp | 102 |
| (unknown) | 77 |
| function | 74 |
| appservice | 70 |
| containerapp-dotnet (multi) | ~50 |

The `checkPolicyRestrictions` API pattern in this PR is architecturally sound and works for any resource type — the only limitation is the current filter to `Microsoft.Storage/storageAccounts`.

**Suggestion:** Merge this PR as-is (storage-only is still valuable), then follow up with a PR that extends the resource type filter to iterate over all resource types found in the Bicep snapshot. This would cover the remaining ~350 non-storage policy errors per month.

**Retry behavior context:** 65% of machines that hit InvalidTemplateDeployment retry and fail again. For policy errors specifically, retrying can never work without a policy exemption or configuration change — better to catch it in seconds at preflight than waste 2.5 minutes on average per deployment attempt.

.vscode/launch.json
```diff
@@ -25,8 +25,9 @@
             "request": "launch",
             "mode": "debug",
             "program": "${workspaceFolder}/cli/azd",
-            "args": "${input:cliArgs}",
+            "args": "up",
```
jongio: **[HIGH]** This replaces the parameterized `${input:cliArgs}` with a hardcoded `"up"` and adds a developer-specific `cwd` path (`/home/vivazqu/...`). These are personal debug settings that should not be committed — they break the shared launch config for other contributors. Please revert this file.

cli/azd/docs/design/extension-flag-handling.md
```diff
@@ -0,0 +1,291 @@
+# Extension Flag Handling Design
```
jongio: **[MEDIUM]** This 291-line design doc about extension flag handling appears unrelated to the storage account policy check feature. Consider moving it to a separate PR to keep the review scope focused and the git history clean.

cli/azd/pkg/azapi/policy_service.go
```diff
@@ -0,0 +1,835 @@
...
+		len(assignments), subscriptionId,
+	)
+
+	var results []DenyPolicy
+
+	// Cache fetched definitions/set definitions to avoid duplicate API
+	// calls when multiple assignments reference the same definition.
+	defCache := make(map[string]*armpolicy.Definition)
+	setDefCache := make(map[string]*armpolicy.SetDefinition)
+
```
jongio: **[MEDIUM]** This loop iterates over all policy assignments and makes API calls (definition fetches) for each one. For subscriptions with many assignments, consider adding a context cancellation check at the top of each iteration to support timely cancellation:

```go
for _, assignment := range assignments {
    if err := ctx.Err(); err != nil {
        return nil, err
    }
    // ...
}
```

cli/azd/pkg/azapi/policy_service.go
```diff
@@ -0,0 +1,835 @@
...
+		defID := *assignment.Properties.PolicyDefinitionID
+		assignmentName := ""
+		if assignment.Properties.DisplayName != nil {
+			assignmentName = *assignment.Properties.DisplayName
+		}
+
+		// Resolve the assignment's parameter values so we can evaluate
+		// parameterized effects.
+		assignmentParams := extractAssignmentParams(assignment)
+
```
jongio: **[LOW]** Nit: The combined condition `isBuiltInPolicyDefinition(defID) || isCustomPolicyDefinition(defID)` could be simplified to `!isPolicySetDefinition(defID)` since together they cover all non-set definitions (built-in, subscription-scoped, and management-group-scoped). This would also make the intent clearer.

.vscode/launch.json
vhvb1989: Reverted — restored `${input:cliArgs}` and removed personal `cwd`.

cli/azd/pkg/azapi/policy_service.go
vhvb1989: Done — added `ctx.Err()` check at the top of the loop.

cli/azd/docs/design/extension-flag-handling.md
vhvb1989: Removed from this PR.

cli/azd/pkg/azapi/policy_service.go
vhvb1989: Done — simplified to `!isPolicySetDefinition` and removed unused `isCustomPolicyDefinition`.

cli/azd/docs/local-preflight/storage-account-policy-check.md
```diff
@@ -0,0 +1,219 @@
...
+**How it works:**
+
+The `Microsoft.PolicyInsights` resource provider offers a
+`checkPolicyRestrictions` API that evaluates hypothetical resources against all
+assigned policies server-side. You submit a resource's content (type, location,
+properties) and Azure returns which policies would deny it and why.
+
+```
+POST /subscriptions/{id}/providers/Microsoft.PolicyInsights/checkPolicyRestrictions
+     ?api-version=2022-03-01
```
jongio, line 148: **[HIGH]** API: `checkPolicyRestrictions` tested against outdated API version

This investigation used `api-version=2022-03-01`. The `2024-10-01` version added a [management group scope endpoint](https://learn.microsoft.com/en-us/rest/api/policyinsights/policy-restrictions/check-at-management-group-scope?view=rest-policyinsights-2024-10-01):

```
POST /providers/Microsoft.Management/managementGroups/{mgId}/providers/Microsoft.PolicyInsights/checkPolicyRestrictions?api-version=2024-10-01
```

This could address the core limitation - the subscription-scope endpoint doesn't see MG-inherited policies, but the MG-scope endpoint might. At minimum:
1. Note which API version was tested (so future readers know the scope of testing)
2. Add the 2024-10-01 MG-scope endpoint to Future Considerations as the most promising next step
3. Consider whether a hybrid (client-side detection + MG-scope server-side evaluation) would work

cli/azd/docs/local-preflight/storage-account-policy-check.md
```diff
@@ -0,0 +1,219 @@
...
+
+A management group assigns the policy "Storage — Disable Local Auth (Opt-In)"
+with `effect: deny`. The policy rule includes:
+
+```json
+{
+  "allOf": [
+    { "field": "type", "equals": "Microsoft.Storage/storageAccounts" },
+    { "field": "Microsoft.Storage/storageAccounts/allowSharedKeyAccess", "equals": "true" },
+    { "value": "[contains(subscription().tags, parameters('optInTagName'))]", "equals": "false" },
```
jongio: **[MEDIUM]** Logic: Example policy conditions are self-contradictory

Tracing the `allOf` logic:
- Condition 3: `contains(subscription().tags, optInTagName)` must equal `"false"` - so the tag must be **absent**
- Condition 4 (anyOf): `split(subscription().tags[optInTagName], ',')` - requires the tag to **exist** to produce a non-empty result

These are mutually exclusive. When the tag is absent, condition 3 passes but condition 4 fails (split of absent value yields empty). When the tag is present (contoso-prod-sub), condition 3 fails immediately (`contains = true` != `"false"`). The policy never fires for any subscription, contradicting the narrative that it fires on contoso-prod-sub.

Condition 3 should likely be `"equals": "true"` (fire when the subscription HAS opted in), which would make both the dev (no tag -> doesn't fire) and prod (has tag -> fires) scenarios work as described.

cli/azd/docs/local-preflight/storage-account-policy-check.md
```diff
@@ -0,0 +1,219 @@
...
+
+## Future Considerations
+
+- If `checkPolicyRestrictions` is updated to evaluate management-group-inherited
+  policies, it would be the ideal solution — accurate server-side evaluation with
+  minimal client complexity.
+- A hybrid approach (client-side detection + subscription tag evaluation for
+  common patterns) could reduce false positives for the most common gating
+  conditions, at the cost of maintaining a partial ARM expression evaluator.
+- The ARM deployment validation API (`/validate`) does evaluate policies but
```
jongio: **[MEDIUM]** Documentation: `/validate` doesn't evaluate Azure Policy deny effects

ARM `/validate` checks template syntax, parameter validity, quotas, and permissions - it doesn't evaluate Azure Policy deny effects. Only actual deployment triggers deny policy evaluation. The what-if API (`/whatIf`) can show expected changes but also doesn't fully simulate deny.

Consider either removing this bullet or correcting it to: "The ARM what-if and deployment validation APIs don't evaluate policy deny effects for hypothetical resources. Only actual deployment submission triggers full policy evaluation, which is equivalent to the server-side preflight that already runs after local preflight."

jongio (dismissed):
**docs: local preflight - storage account policy check investigation** by @vhvb1989

### Summary

What: Moves the local preflight design doc to `docs/local-preflight/README.md`, adds a "not implemented" table, and adds a 219-line investigation doc explaining why a storage account policy check wasn't shipped.

Why: Documents three approaches to detecting Azure Policy restrictions on storage accounts before deployment - client-side parsing (false positives), server-side `checkPolicyRestrictions` (misses MG policies), and `policyStates` (existing resources only). Thorough investigation with clear reasoning for why the feature was abandoned.

Assessment: The investigation is well-structured and the conclusion is well-supported. However, the `checkPolicyRestrictions` testing used api-version 2022-03-01 - the 2024-10-01 version added a management group scope endpoint that could change the conclusion. The example policy JSON has a logic issue, and one Future Considerations bullet has an inaccuracy about `/validate`.

### Findings

| Category | High | Medium |
|----------|------|--------|
| API | 1 | 0 |
| Logic | 0 | 1 |
| Documentation | 0 | 1 |
| **Total** | **1** | **2** |

### Key Findings

1. **[HIGH] API**: `checkPolicyRestrictions` was tested against 2022-03-01 only - the 2024-10-01 API version adds a management group scope endpoint
2. **[MEDIUM] Logic**: Example policy JSON conditions are self-contradictory
3. **[MEDIUM] Documentation**: `/validate` doesn't evaluate Azure Policy deny effects

### What's Done Well

- Thorough investigation with three distinct approaches, each with clear evidence for why it fails
- Excellent use of concrete examples (real policy patterns, specific API responses)
- Good separation into a dedicated `docs/local-preflight/` directory
- The file rename preserves all existing relative links correctly

---

*3 inline comments below.*

cli/azd/docs/local-preflight/storage-account-policy-check.md
```diff
@@ -0,0 +1,219 @@
...
+However, it cannot evaluate **hypothetical** resources from a Bicep snapshot. It
+only works with resources that have already been deployed. Since the goal is to
+warn *before* deployment, this API is not applicable.
+
+## Conclusion
+
+| Approach | Sees MG policies | Evaluates all conditions | Suitable |
+|---|---|---|---|
+| Client-side ARM policy SDK parsing | ✅ Yes | ❌ No (ARM expressions) | ❌ False positives |
+| Server-side `checkPolicyRestrictions` | ❌ No | ✅ Yes | ❌ Misses real denials |
```
wbreza: **[LOW]** Conclusion table could mislead at a glance

The `policyStates` row shows :white_check_mark: for both "Sees MG policies" and "Evaluates all conditions" alongside :x: for "Suitable". A reader scanning the table without reading the text might wonder why two green checks still yielded a rejection.

Consider adding a footnote or a fourth column ("Limitation") to make this self-explanatory:

| Approach | Sees MG policies | Evaluates all conditions | Limitation | Suitable |
|---|---|---|---|---|
| `policyStates` API | :white_check_mark: Yes | :white_check_mark: Yes | Existing resources only | :x: |

cli/azd/docs/local-preflight/storage-account-policy-check.md
vhvb1989: Done — added API version annotation (`2022-03-01`) to the request example, and added the `2024-10-01` MG-scope endpoint to Future Considerations as the most promising next step.

cli/azd/docs/local-preflight/storage-account-policy-check.md
vhvb1989: Fixed — changed condition 3 from `equals: "false"` to `equals: "true"`. Now the logic is consistent: dev-sub (no tag) → `contains()` is false → does not match "true" → allOf short-circuits → policy does not fire. prod-sub (has tag) → `contains()` is true → matches "true" → proceeds to anyOf → fires.

cli/azd/docs/local-preflight/storage-account-policy-check.md
vhvb1989: Fixed — corrected to state that `/validate` and `/whatIf` do NOT evaluate Azure Policy deny effects. Only actual deployment submission triggers deny evaluation.

cli/azd/docs/local-preflight/storage-account-policy-check.md
vhvb1989: Done — added a "Limitation" column to the conclusion table so each row is self-explanatory at a glance.

jongio (commented):
Solid investigation writeup. Two minor wording issues:
- storage-account-policy-check.md:229,247 - `Untested` for MG scope is misleading since testing was done (returned empty)
- storage-account-policy-check.md:4 - date says March 2026 but April follow-up section exists

Not blocking.

cli/azd/docs/local-preflight/storage-account-policy-check.md
```diff
@@ -0,0 +1,270 @@
...
+only works with resources that have already been deployed. Since the goal is to
+warn *before* deployment, this API is not applicable.
+
+## Conclusion
+
+| Approach | Sees MG policies | Evaluates all conditions | Limitation | Suitable |
+|---|---|---|---|---|
+| Client-side ARM policy SDK parsing | ✅ Yes | ❌ No (ARM expressions) | Cannot evaluate runtime expressions → false positives | ❌ |
+| Server-side `checkPolicyRestrictions` (sub scope) | ❌ No | ✅ Yes | Misses MG-inherited policies (confirmed with both `2022-03-01` and `2024-10-01`) | ❌ |
+| Server-side `checkPolicyRestrictions` (MG scope, `2024-10-01`) | Untested | ❌ No | Only supports `type` field; rejects `resourceDetails` | ❌ |
```
jongio, line 247: **[MEDIUM]** `Untested` is misleading here - the body text describes testing the MG-scope endpoint at multiple hierarchy levels. All calls returned empty `fieldRestrictions`, so the result is inconclusive rather than untested.

```suggestion
| Server-side `checkPolicyRestrictions` (MG scope, `2024-10-01`) | Inconclusive (empty results) | :x: No | Only supports `type` field; rejects `resourceDetails` | :x: |
```

cli/azd/docs/local-preflight/storage-account-policy-check.md
```diff
@@ -0,0 +1,270 @@
...
+restricted" at a management group level, not "would this specific resource
+configuration be denied" — making it unsuitable for property-level checks like
+`allowSharedKeyAccess`.
+
+| Endpoint | API Version | Sees MG deny policies | Supports resource properties |
+|---|---|---|---|
+| Subscription scope | `2022-03-01` | ❌ | ✅ |
+| Subscription scope | `2024-10-01` | ❌ | ✅ |
+| Resource group scope | `2024-10-01` | ❌ | ✅ |
+| MG scope | `2024-10-01` | Untested (empty for `type`) | ❌ (rejected) |
```
jongio, line 229: **[MEDIUM]** Same here - `Untested` doesn't match the actual investigation described above.

```suggestion
| MG scope | `2024-10-01` | Inconclusive (empty for `type`) | :x: (rejected) |
```

cli/azd/docs/local-preflight/storage-account-policy-check.md
```diff
@@ -0,0 +1,270 @@
+# Storage Account Policy Check — Investigation Summary
+
+> **Status**: Not implemented — false positive rate too high for the warning to be actionable.
+> Investigation completed March 2026.
```
jongio, line 4: **[LOW]** The `2024-10-01` follow-up section is dated April 2026, so `March 2026` doesn't cover the full investigation window.

```suggestion
> Investigation: March-April 2026.
```

PR #7206, opened 2026-03-19
Add compute resolution in job submit

saanikaguptamicrosoft:


cli/azd/extensions/azure.ai.customtraining/pkg/client/client.go
```diff
@@ -74,6 +81,48 @@ func NewClient(projectEndpoint string, credential azcore.TokenCredential) (*Clie
...
+	c.resourceGroup = resourceGroup
+	c.accountName = accountName
+}
+
+// doARM executes an authenticated HTTP request against the ARM control plane.
+// The path should be relative to https://management.azure.com/ (no leading slash).
+func (c *Client) doARM(ctx context.Context, method, path string, body interface{}, apiVersion string) (*http.Response, error) {
+	reqURL := fmt.Sprintf("https://management.azure.com/%s?api-version=%s", path, apiVersion)
+
+	fmt.Printf("[DEBUG] %s %s\n", method, reqURL)
```
jongio, line 97: **[HIGH]** Quality: Debug `fmt.Printf` left in production code

`fmt.Printf("[DEBUG] %s %s\n", method, reqURL)` prints the full ARM URL (including subscription ID) to stdout on every control plane call. This isn't gated behind any debug flag or log level.

`doDataPlane` has the same pattern - but adding more of it isn't a reason to keep it. This is a good time to remove both or gate them behind an env var / structured logger.

cli/azd/extensions/azure.ai.customtraining/internal/service/compute_resolver.go
```diff
@@ -6,16 +6,35 @@ package service
...
+// ResolveCompute calls the ARM API to resolve a compute name to its full ARM resource ID.
+// Returns a helpful error message if the user lacks permissions (401/403).
 func (r *DefaultComputeResolver) ResolveCompute(ctx context.Context, computeName string) (string, error) {
-	return "", fmt.Errorf("compute resolution not implemented: provide a full ARM resource ID for compute '%s'", computeName)
+	result, err := r.client.GetCompute(ctx, computeName)
+	if err != nil {
+		return "", err
+	}
+
+	fmt.Printf("  ✓ Compute resolved: %s\n", computeName)
```
jongio, line 38: **[MEDIUM]** Quality: Service layer shouldn't write directly to stdout

`fmt.Printf` couples this library code to a specific output format. The command layer should control user-facing output - the resolver should just return data.

This also makes the resolver harder to test (no way to assert output) and breaks if the caller wants JSON or quiet mode.

cli/azd/extensions/azure.ai.customtraining/pkg/client/compute.go
```diff
@@ -0,0 +1,77 @@
...
+// GetCompute retrieves a compute resource by name from the ARM control plane.
+//
+// ARM URL:
+//
+//	GET https://management.azure.com/subscriptions/{sub}/resourceGroups/{rg}/providers/Microsoft.CognitiveServices/accounts/{account}/computes/{name}?api-version=2026-01-15-preview
+//
+// Requires SetARMContext to be called first.
+func (c *Client) GetCompute(ctx context.Context, computeName string) (*models.ComputeResource, error) {
+	if c.subscriptionID == "" || c.resourceGroup == "" || c.accountName == "" {
+		return nil, fmt.Errorf("ARM context not configured; call SetARMContext first")
```
jongio, line 32: **[MEDIUM]** Error Handling: Internal jargon in user-facing error path

"ARM context not configured; call SetARMContext first" is developer-speak. When a user hits this (because `AZURE_SUBSCRIPTION_ID` or `AZURE_RESOURCE_GROUP_NAME` aren't set in the env), the message won't help them fix it.

Something like: "subscription ID and resource group are required to resolve compute names. Run 'azd ai training init' or provide a full ARM resource ID in your YAML."

cli/azd/extensions/azure.ai.customtraining/pkg/client/compute.go
```diff
@@ -0,0 +1,77 @@
...
+		return nil, fmt.Errorf("failed to call ARM compute API: %w", err)
+	}
+	defer resp.Body.Close()
+
+	// Permission error — guide user to provide full ARM ID instead
+	if resp.StatusCode == http.StatusUnauthorized || resp.StatusCode == http.StatusForbidden {
+		return nil, fmt.Errorf(
+			"insufficient permissions to resolve compute '%s'.\n"+
+				"  Provide the full ARM resource ID in your YAML instead:\n"+
+				"  compute: /subscriptions/%s/resourceGroups/%s/providers/Microsoft.CognitiveServices/accounts/%s/computes/%s",
```
jongio, line 54: **[LOW]** Security: Error message exposes subscription ID

The 401/403 error includes the subscription ID in the suggested ARM resource ID. Other azd code respects `AZD_DEMO_MODE` to suppress subscription IDs in user-facing output. Consider the same here, or at least note this is intentional.

Issue #7209, opened 2026-03-19
improve the formatting and display of logs for hosted agents


therealjohn:
Improve the current output of `azd ai agent monitor` so its compact and easier to read.

Current:

<img width="3479" height="1859" alt="Image" src="https://github.com/user-attachments/assets/7eb4d9d6-e9c6-4330-afae-b0369e48ffab" />

Suggested:

<img width="2116" height="644" alt="Image" src="https://github.com/user-attachments/assets/3c20402c-1ab5-4270-a993-aa1df64d5d71" />

zeel2104:
I would like to work on it. Can you assign it to me?

PR #7210, opened 2026-03-19
Add raw flag to monitor command and implement log formatting for SSE on azd ai agent monitor

therealjohn:
Prototype for #7209.

Parse SSE/JSON output from monitor and format the results to be more compact and readable. 

Include a `--raw` flag which continues the existing behavior of just printing the raw SSE stream from the server.

### DIFF
--- a/cli/azd/extensions/azure.ai.agents/internal/cmd/monitor.go
+++ b/cli/azd/extensions/azure.ai.agents/internal/cmd/monitor.go
@@ -24,6 +24,7 @@ type monitorFlags struct {
 	follow    bool
 	tail      int
 	logType   string
+	raw       bool
 }
 
 // MonitorAction handles the execution of the monitor command.
@@ -135,6 +136,7 @@ configuration and the current azd environment. Optionally specify the service na
 	cmd.Flags().IntVarP(&flags.tail, "tail", "l", 50, "Number of trailing log lines to fetch (1-300)")
 	cmd.Flags().StringVarP(&flags.logType, "type", "t", "console",
 		"Type of logs: 'console' (stdout/stderr) or 'system' (container events)")
+	cmd.Flags().BoolVarP(&flags.raw, "raw", "r", false, "Print raw SSE output without formatting")
 
 	return cmd
 }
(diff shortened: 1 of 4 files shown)


cli/azd/extensions/azure.ai.agents/internal/cmd/monitor.go
```diff
@@ -180,8 +182,17 @@ func (a *MonitorAction) Run(ctx context.Context) error {
 	defer body.Close()
 
 	scanner := bufio.NewScanner(body)
-	for scanner.Scan() {
-		fmt.Println(scanner.Text())
+	if a.flags.raw {
```
jongio, line 185: **[HIGH]** The SSE formatter runs on both the session logstream and the container logstream. Container logs aren't SSE-formatted, so `formatLine()` silently drops lines starting with `event: ` or `data: `, and suppresses blank lines in the default (no-flags) path.

```suggestion
	if a.flags.raw || a.flags.sessionID == "" {
```

cli/azd/extensions/azure.ai.agents/internal/cmd/monitor_format.go
```diff
@@ -0,0 +1,151 @@
...
+	switch label {
+	case "stderr":
+		labelStr = f.stderrColor.Sprint(label)
+	case "status":
+		labelStr = f.statusColor.Sprint(label)
+	default:
+		labelStr = label
+	}
+
+	return fmt.Sprintf("%s  %-8s %s",
```
jongio, line 113: **[MEDIUM]** `%-8s` counts ANSI escape bytes, so colored labels (stderr, status) don't get padded the same as plain ones (stdout). Columns misalign when color is on.

Fix: pad the label text before colorizing, then use `%s` here instead of `%-8s`.

cli/azd/extensions/azure.ai.agents/internal/cmd/monitor_format.go
```diff
@@ -0,0 +1,151 @@
...
+	SessionState string `json:"session_state"`
+	Agent        string `json:"agent"`
+	Version      string `json:"version"`
+	LastAccessed string `json:"last_accessed"`
+}
+
+// logFormatter parses SSE log lines from the monitor stream and produces
+// human-readable, color-coded output.
+type logFormatter struct {
+	pendingEvent string // last seen "event: <type>" value
```
jongio, line 35: **[LOW]** Written and cleared but never read - no formatting decision uses the event type. If it's planned for future use, a TODO would help; otherwise it can go.

jongio (commented):
@spboyer's scoping concern is valid - the SSE formatter runs on both the session logstream and the container logstream, so non-SSE output gets mangled in the default case.

CI is also red: gofmt formatting in monitor_format.go and lll violation in monitor_format_test.go (246 chars, max 220).

Issues to address:
- monitor.go:185 - scope the formatter to SSE streams only
- monitor_format.go:113 - column alignment breaks with ANSI color codes
- monitor_format.go:35 - pendingEvent is written but never read

PR #7223, opened 2026-03-20
functionapp: handle .funcignore more gracefully

weikanglim:
Currently, azd defaults `remoteBuild: true` for flex-consumption function apps. This static default causes customer pain when the `.funcignore` file contents isn't compatible with the build behavior. Primarily, users scaffolding a function app project using `func` cli are impacted.

To address this, we now default the `remoteBuild` setting intelligently based on the contents of the `.funcignore` file. When `remoteBuild` is set explicitly, we also validate and fail fast if the `.funcignore` is detected to be incompatible (i.e. the presence of `node_modules`).

Fixes https://github.com/Azure/azure-dev/issues/6531

### DIFF
--- a/cli/azd/pkg/project/project_utils.go
+++ b/cli/azd/pkg/project/project_utils.go
@@ -84,6 +84,11 @@ func createDeployableZip(svc *ServiceConfig, root string) (string, error) {
 				}
 			} else if svc.Language == ServiceLanguageJavaScript || svc.Language == ServiceLanguageTypeScript {
 				if name == "node_modules" && isDir {
+					if svc.RemoteBuild != nil && !*svc.RemoteBuild {
+						// if remote build is false, we do not exclude node_modules by default
+						return true, nil
+					}
+
 					return false, nil
 				}
 			}


spboyer (changes requested):
Please address the copilot feedback, either comment or commit to address,

cli/azd/pkg/project/service_target_functionapp_test.go
```diff
@@ -45,3 +48,109 @@ func TestNewFunctionAppTargetTypeValidation(t *testing.T) {
...
+		},
+		{
+			name:              "NoRemoteBuildAndMissingFuncIgnore_RemoteBuildEnabled",
+			remoteBuild:       nil,
+			funcIgnoreContent: "",
+			expectRemoteBuild: true,
+		},
+		{
+			name:              "RemoteBuildFalseAndFuncIgnoreExcludesNodeModules_Errors",
+			remoteBuild:       new(false),
```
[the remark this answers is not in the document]
weikanglim: This is incorrect -- modern 1.26 go

cli/azd/pkg/project/service_target_functionapp_test.go
```diff
@@ -45,3 +48,109 @@ func TestNewFunctionAppTargetTypeValidation(t *testing.T) {
...
+		},
+		{
+			name:              "RemoteBuildFalseAndFuncIgnoreDoesNotExcludeNodeModules_Succeeds",
+			remoteBuild:       new(false),
+			funcIgnoreContent: "dist\n",
+			expectRemoteBuild: false,
+		},
+		{
+			name:              "RemoteBuildTrueAndFuncIgnoreExcludesNodeModules_Succeeds",
+			remoteBuild:       new(true),
```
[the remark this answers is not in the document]
weikanglim: This is incorrect -- modern 1.26 go

cli/azd/pkg/project/service_target_functionapp_test.go
```diff
@@ -45,3 +48,109 @@ func TestNewFunctionAppTargetTypeValidation(t *testing.T) {
...
+			funcIgnoreContent: "node_modules\n",
+			expectRemoteBuild: true,
+		},
+		{
+			name:              "RemoteBuildTrueAndFuncIgnoreDoesNotExcludeNodeModules_Errors",
+			remoteBuild:       new(true),
```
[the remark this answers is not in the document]
weikanglim: This is incorrect -- modern 1.26 go

cli/azd/pkg/project/service_target_functionapp_test.go
```diff
@@ -45,3 +48,109 @@ func TestNewFunctionAppTargetTypeValidation(t *testing.T) {
...
+
+func TestResolveFunctionAppRemoteBuild_NonJavaScriptDefaults(t *testing.T) {
+	t.Parallel()
+
+	pythonConfig := createTestServiceConfig(t.TempDir(), AzureFunctionTarget, ServiceLanguagePython)
+	remoteBuild, err := resolveFunctionAppRemoteBuild(pythonConfig)
+	require.NoError(t, err)
+	require.True(t, remoteBuild)
+
+	pythonConfig.RemoteBuild = new(false)
```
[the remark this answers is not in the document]
weikanglim: This is incorrect -- modern 1.26 go

cli/azd/pkg/project/service_target_functionapp.go
```diff
@@ -16,9 +19,71 @@ import (
...
+			nodeModulesExcluded = true
+		}
+
+		if serviceConfig.RemoteBuild == nil { // remoteBuild option unset
+			// enable remote build only if 'node_modules' is excluded
+			return nodeModulesExcluded, nil
+		}
+
+		if *serviceConfig.RemoteBuild && !nodeModulesExcluded {
+			return false, &internal.ErrorWithSuggestion{
```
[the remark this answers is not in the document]
weikanglim: This is working

cli/azd/pkg/project/service_target_functionapp_test.go
```diff
@@ -45,3 +48,109 @@ func TestNewFunctionAppTargetTypeValidation(t *testing.T) {
...
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+
+			serviceConfig := createTestServiceConfig(t.TempDir(), AzureFunctionTarget, ServiceLanguageJavaScript)
+			serviceConfig.RemoteBuild = tt.remoteBuild
+
+			if tt.funcIgnoreContent != "" {
+				err := os.WriteFile(
+					filepath.Join(serviceConfig.Path(), ".funcignore"),
```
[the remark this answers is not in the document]
weikanglim: This is a test that is exercising external expectations that the file is ".funcignore" which is fine. The service being tested is explicitly function app.

cli/azd/pkg/project/service_target_functionapp.go
```diff
@@ -16,9 +19,71 @@ import (
...
 	"github.com/azure/azure-dev/cli/azd/pkg/tools"
+	"github.com/denormal/go-gitignore"
 )
 
+const functionAppRemoteBuildDocURL = "https://aka.ms/azd-functionapp-remote-build"
+
+// resolveFunctionAppRemoteBuild returns the appropriate remote build setting for function apps.
+func resolveFunctionAppRemoteBuild(serviceConfig *ServiceConfig) (remoteBuild bool, err error) {
+	switch serviceConfig.Language {
+	case ServiceLanguageJavaScript, ServiceLanguageTypeScript:
```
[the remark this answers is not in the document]
weikanglim: ts/js are aliases. The suggestion is the same fallthrough.

weikanglim:
@spboyer Added comments to each one of them. Please review

cli/azd/pkg/project/service_target_functionapp.go
```diff
@@ -16,9 +19,71 @@ import (
...
+		ignoreFilePath := filepath.Join(serviceConfig.Path(), serviceConfig.Host.IgnoreFile())
+		ignoreFileContents, err := os.ReadFile(ignoreFilePath)
+		if errors.Is(err, fs.ErrNotExist) {
+			// no ignore file, default to true
+			return true, nil
```
jongio, line 40: If a user explicitly sets `remoteBuild: false` in azure.yaml but doesn't have a .funcignore, this returns `true` anyway, silently overriding what they asked for.

Looks like the `default:` branch already handles this correctly by checking `serviceConfig.RemoteBuild` first. Should we do the same here?

```go
if errors.Is(err, fs.ErrNotExist) {
    if serviceConfig.RemoteBuild != nil {
        return *serviceConfig.RemoteBuild, nil
    }
    return true, nil
}
```

cli/azd/pkg/project/service_target_functionapp.go
```diff
@@ -16,9 +19,71 @@ import (
...
+		if match := ignore.Relative("node_modules", true); match != nil && match.Ignore() {
+			nodeModulesExcluded = true
+		}
+
+		if serviceConfig.RemoteBuild == nil { // remoteBuild option unset
+			// enable remote build only if 'node_modules' is excluded
+			return nodeModulesExcluded, nil
+		}
+
+		if *serviceConfig.RemoteBuild && !nodeModulesExcluded {
```
jongio, line 60: This feels a bit asymmetric. If there's no .funcignore at all, `remoteBuild: true` sails through fine. But if someone creates a .funcignore without a node_modules exclusion, they get an error. The actual deployment artifact is the same in both cases (zip with node_modules + remote build).

Maybe worth either flagging both cases or downgrading this to a warning since the deployment still works, just ships more bytes than it needs to?

cli/azd/pkg/project/service_target_functionapp_test.go
```diff
@@ -45,3 +48,109 @@ func TestNewFunctionAppTargetTypeValidation(t *testing.T) {
...
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+
+			serviceConfig := createTestServiceConfig(t.TempDir(), AzureFunctionTarget, ServiceLanguageJavaScript)
+			serviceConfig.RemoteBuild = tt.remoteBuild
+
+			if tt.funcIgnoreContent != "" {
```
jongio, line 119: Using empty string to mean "don't create the file" means we can't actually test what happens with an empty .funcignore (file exists, 0 bytes). That's a real edge case though. Missing file hits ErrNotExist and returns true, but an empty file would pass ReadFile, give gitignore no rules, and node_modules wouldn't be excluded.

Maybe use `*string` (nil = no file, "" = empty file) or a separate bool? Not blocking, just something to keep in mind.

cli/azd/pkg/project/service_target_functionapp.go
```diff
@@ -16,9 +19,71 @@ import (
...
 )
 
+const functionAppRemoteBuildDocURL = "https://aka.ms/azd-functionapp-remote-build"
+
+// resolveFunctionAppRemoteBuild returns the appropriate remote build setting for function apps.
+func resolveFunctionAppRemoteBuild(serviceConfig *ServiceConfig) (remoteBuild bool, err error) {
+	switch serviceConfig.Language {
+	case ServiceLanguageJavaScript, ServiceLanguageTypeScript:
+		ignoreFilePath := filepath.Join(serviceConfig.Path(), serviceConfig.Host.IgnoreFile())
+		ignoreFileContents, err := os.ReadFile(ignoreFilePath)
```
jongio, line 32: Nit: the named return `err` gets shadowed by `:=` here since `ignoreFileContents` is new. Not a problem right now, but if someone adds a defer that references `err` later, they'd get nil instead of the read error.

Could do `var ignoreFileContents []byte` + `=`, or just rename to `readErr`.

cli/azd/pkg/project/service_target_functionapp.go
```diff
@@ -16,9 +19,71 @@ import (
...
+		}
+
+		if serviceConfig.RemoteBuild == nil { // remoteBuild option unset
+			// enable remote build only if 'node_modules' is excluded
+			return nodeModulesExcluded, nil
+		}
+
+		if *serviceConfig.RemoteBuild && !nodeModulesExcluded {
+			return false, &internal.ErrorWithSuggestion{
+				Err: fmt.Errorf("'remoteBuild: true' requires '.funcignore' to exclude node_modules"),
```
jongio: The path comes from `serviceConfig.Host.IgnoreFile()` but the error messages hardcode `.funcignore`. Fine for now since only function apps go through here, but if this ever gets reused for other targets it'd be confusing. Worth capturing `ignoreFile := serviceConfig.Host.IgnoreFile()` and using that in the error strings?

cli/azd/pkg/project/service_target_functionapp.go
weikanglim: The artifact isn't the same in both cases:
- No `.funcignore`: azd automatically ignores `node_modules` by default
- With `.funcignore`: User has to add `node_modules` to get the same behavior

This [doc](https://learn.microsoft.com/en-us/azure/developer/azure-developer-cli/service-packaging-ignore-files?tabs=nodejs#exclusion-examples) explains the current behavior.

cli/azd/pkg/project/service_target_functionapp.go
weikanglim: Yes, this is indeed a bug -- good catch. Will update this.

cli/azd/pkg/project/service_target_functionapp_test.go
weikanglim: That's an astute observation. It's highly unlikely that we'd need to test the empty string case, the parsing itself would be tested separately. I do believe we're fine here -- not changing the test

cli/azd/pkg/project/service_target_functionapp.go
weikanglim: That's a great observation. I'm changing the name to avoid the shadowing that could affect `defer` returns

cli/azd/pkg/project/service_target_functionapp.go
weikanglim: Updated -- and updated the default packaging exclusion logic in `project_utils.go` to also handle this scenario correctly.

cli/azd/pkg/project/service_target_functionapp.go
weikanglim: That makes sense. Updated.

cli/azd/pkg/project/project_utils.go
```diff
@@ -84,6 +84,11 @@ func createDeployableZip(svc *ServiceConfig, root string) (string, error) {
 				}
 			} else if svc.Language == ServiceLanguageJavaScript || svc.Language == ServiceLanguageTypeScript {
 				if name == "node_modules" && isDir {
+					if svc.RemoteBuild != nil && !*svc.RemoteBuild {
+						// if remote build is false, we do not exclude node_modules by default
+						return true, nil
+					}
+
```
vhvb1989, line 91: nit. Can we avoid the false condition and return false for svc.RemoteBuild  ?   Negative conditions are usually complicated to follow (if false -> do not do this v/s  if true -> do this)

weikanglim: good point. I agree with you that being cleaner and a general improvement. In interest of time, I'm going to take a note on this and remember it for next time.

cli/azd/pkg/project/service_target_functionapp_test.go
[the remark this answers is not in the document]
kristenwomack: @weikanglim - we discussed this morning to add this feedback to the coding agent instructions to help in learn from feedback. If copilot got it wrong and left multiple comments, this is an opportunity to improve our eng harness. But the comment is copy/pasted in each comment.

kristenwomack:
Thank you for getting this pushed today.

PR #7235, opened 2026-03-22
Fix auth error telemetry classification

### LINKED ISSUE #7233: Telemetry: auth errors (login_required, not_logged_in) classified as 'unknown' error category
## Problem

Auth-related errors from `cmd.auth.token` are correctly assigned meaningful result codes (`auth.login_required`, `auth.not_logged_in`, `internal.azidentity_AuthenticationFailedError`) but their **error category** (`AzdErrorCategory` in Kusto) remains `"unknown"` instead of being classified as `"aad"` or `"auth"`.

This makes it impossible to filter auth errors by category in telemetry dashboards and inflates the `"unknown"` error bucket.

### Telemetry Impact (rolling 28 days ending Mar 18, 2026)

| Result Code | Current Category | Count | Users |
|-------------|-----------------|-------|-------|
| `auth.login_required` | **unknown** ❌ | 543,972 | 1,353 |
| `auth.not_logged_in` | **unknown** ❌ | 65,643 | 1,233 |
| `internal.azidentity_AuthenticationFailedError` | **unknown** ❌ | 7,258 | 65 |
| `service.aad.failed` | **aad** ✅ | 12,246 | 658 |

**610K+ auth failures per 28 days are miscategorized.** These account for ~17% of all auth token errors.

## Root Cause

In [`cli/azd/internal/cmd/errors.go`](https://github.com/Azure/azure-dev/blob/main/cli/azd/internal/cmd/errors.go), the `MapError` function handles auth errors in three separate code paths, but only one sets a service/category detail attribute:

1. **`auth.ReLoginRequiredError`** (line ~53) — sets `errCode = "auth.login_required"` but does NOT append `fields.ServiceName.String("aad")` to `errDetails`
2. **`auth.ErrNoCurrentUser`** sentinel (in `classifySentinel`) — returns `"auth.not_logged_in"` but sentinel handling does NOT set any `errDetails` at all
3. **`auth.AuthFailedError`** (line ~133) — correctly appends `fields.ServiceName.String("aad")` to `errDetails` ✅

The downstream Kusto ETL derives `AzdErrorCategory` from the `ServiceName` span attribute. When it is absent, the category defaults to `"unknown"`.

## Suggested Fix

Add `fields.ServiceName.String("aad")` (or a new `fields.ErrCategory.String("auth")`) to the `errDetails` for the `ReLoginRequiredError` branch, and add similar detail attribution for auth-related sentinels.

In `MapError()`:

```go
// Current (line ~53):
} else if _, ok := errors.AsType[*auth.ReLoginRequiredError](err); ok {
    errCode = "auth.login_required"

// Suggested:
} else if _, ok := errors.AsType[*auth.ReLoginRequiredError](err); ok {
    errCode = "auth.login_required"
    errDetails = append(errDetails, fields.ServiceName.String("aad"))
```

For sentinel-based auth errors, the fix would need to either:
- (a) Return details alongside the code from `classifySentinel`, or
- (b) Check for auth sentinels before falling through to `classifySentinel` in the main `MapError` function:

```go
// Add before the classifySentinel fallthrough:
} else if errors.Is(err, auth.ErrNoCurrentUser) {
    errCode = "auth.not_logged_in"
    errDetails = append(errDetails, fields.ServiceName.String("aad"))
```

A similar pattern should be applied for `azidentity` errors that currently fall through to the generic `internal.*` catch-all.

## Expected Result

After this fix, the Kusto `AzdErrorCategory` column would show:

| Result Code | Category (after fix) |
|-------------|---------------------|
| `auth.login_required` | `aad` ✅ |
| `auth.not_logged_in` | `aad` ✅ |
| `internal.azidentity_AuthenticationFailedError` | `aad` ✅ |
| `service.aad.failed` | `aad` ✅ (unchanged) |

This would remove ~610K errors/28d from the `"unknown"` bucket, making the overall error category distribution far more actionable for telemetry analysis.

spboyer:
## Summary

Fixes #7233

Auth errors (`login_required`, `not_logged_in`, `azidentity.AuthenticationFailedError`) were classified as `"unknown"` in telemetry because they did not set `fields.ServiceName.String("aad")`. This affected **~610K errors per 28 days** and made the `"unknown"` error category bucket unreliable for analysis.

## Changes

| Error Type | Before | After |
|---|---|---|
| `ReLoginRequiredError` → `auth.login_required` | ❌ no ServiceName | ✅ `aad` |
| `ErrNoCurrentUser` → `auth.not_logged_in` | ❌ no ServiceName (handled in `classifySentinel`) | ✅ `aad` (moved to `MapError` if-else chain) |
| `azidentity.AuthenticationFailedError` | ❌ fell through to `internal.*` catch-all | ✅ new code `auth.identity_failed` with ServiceName `aad` |

### Code
- **`errors.go`**: Added `ServiceName("aad")` to `ReLoginRequiredError` branch, moved `ErrNoCurrentUser` out of `classifySentinel()` into `MapError()`, added `azidentity.AuthenticationFailedError` handler, updated `classifySuggestionType()` to stay in sync.
- **`errors_test.go`**: Added test cases for `ReLoginRequiredError` and `azidentity.AuthenticationFailedError`, updated `ErrNoCurrentUser` tests to expect `ServiceName("aad")`, added entries to `ClassifySuggestionType_MatchesMapError`.

## Expected Telemetry Impact

After this fix, the Kusto `AzdErrorCategory` column will show:

| Result Code | Category (before) | Category (after) |
|---|---|---|
| `auth.login_required` | `unknown` ❌ | `aad` ✅ |
| `auth.not_logged_in` | `unknown` ❌ | `aad` ✅ |
| `auth.identity_failed` (was `internal.azidentity_*`) | `unknown` ❌ | `aad` ✅ |
| `service.aad.failed` | `aad` ✅ | `aad` ✅ (unchanged) |

This removes **~610K errors/28d from the `"unknown"` bucket**.

### DIFF
--- a/cli/azd/internal/cmd/errors.go
+++ b/cli/azd/internal/cmd/errors.go
@@ -19,6 +19,7 @@ import (
 
 	"github.com/AlecAivazis/survey/v2/terminal"
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
+	"github.com/Azure/azure-sdk-for-go/sdk/azidentity"
 	"github.com/azure/azure-dev/cli/azd/internal"
 	"github.com/azure/azure-dev/cli/azd/internal/agent/consent"
 	"github.com/azure/azure-dev/cli/azd/internal/tracing"
@@ -48,6 +49,7 @@ func MapError(err error, span tracing.Span) {
 		errCode = updateErr.Code
 	} else if _, ok := errors.AsType[*auth.ReLoginRequiredError](err); ok {
 		errCode = "auth.login_required"
+		errDetails = append(errDetails, fields.ErrCategory.String("auth"))
 	} else if errWithSuggestion, ok := errors.AsType[*internal.ErrorWithSuggestion](err); ok {
 		errCode = "error.suggestion"
 		span.SetAttributes(fields.ErrType.String(classifySuggestionType(errWithSuggestion.Unwrap())))
@@ -184,6 +186,12 @@ func MapError(err error, span tracing.Span) {
 				fields.ServiceCorrelationId.String(authFailedErr.Parsed.CorrelationId))
 		}
 		errCode = "service.aad.failed"
+	} else if errors.Is(err, auth.ErrNoCurrentUser) {
+		errCode = "auth.not_logged_in"
+		errDetails = append(errDetails, fields.ErrCategory.String("auth"))
+	} else if _, ok := errors.AsType[*azidentity.AuthenticationFailedError](err); ok {
+		errCode = "auth.identity_failed"
+		errDetails = append(errDetails, fields.ErrCategory.String("auth"))
 	} else if code := classifySentinel(err); code != "" {
 		errCode = code
 	} else if isNetworkError(err) {
@@ -268,8 +276,6 @@ func classifySentinel(err error) string {
 		return "user.canceled"
 	case errors.Is(err, context.DeadlineExceeded):
 		return "internal.timeout"
-	case errors.Is(err, auth.ErrNoCurrentUser):
-		return "auth.not_logged_in"
 	case errors.Is(err, consent.ErrToolExecutionDenied):
 		return "user.tool_denied"
 	case errors.Is(err, git.ErrNotRepository):
@@ -372,6 +378,14 @@ func classifySuggestionType(err error) string {
 
 	if _, ok := errors.AsType[*auth.AuthFailedError](err); ok {
 		return "service.aad.failed"
+	}
+
+	if errors.Is(err, auth.ErrNoCurrentUser) {
+		return "auth.not_logged_in"
+	}
+
+	if _, ok := errors.AsType[*azidentity.AuthenticationFailedError](err); ok {
+		return "auth.identity_failed"
 	}
 
 	if code := classifySentinel(err); code != "" {
--- a/cli/azd/internal/cmd/errors_test.go
+++ b/cli/azd/internal/cmd/errors_test.go
@@ -21,6 +21,7 @@ import (
 	"testing"
 
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
+	"github.com/Azure/azure-sdk-for-go/sdk/azidentity"
 	"github.com/azure/azure-dev/cli/azd/internal"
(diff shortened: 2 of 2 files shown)


cli/azd/internal/cmd/errors.go
```diff
@@ -184,6 +186,12 @@ func MapError(err error, span tracing.Span) {
 				fields.ServiceCorrelationId.String(authFailedErr.Parsed.CorrelationId))
 		}
 		errCode = "service.aad.failed"
+	} else if errors.Is(err, auth.ErrNoCurrentUser) {
+		errCode = "auth.not_logged_in"
+		errDetails = append(errDetails, fields.ServiceName.String("aad"))
```
vhvb1989: I don't think we should be using `aad` as the serviceName.  This error is locally generated and it could be coming from azd or az (when delegating auth).

Do we have something to classify that?  local issues?

spboyer:
Great catch @JeffreyCA — you're right that `error.service.name` semantically means an external service returned the error, which doesn't apply to `ErrNoCurrentUser` (local state check, no service called).

Here's the corrected approach:

| Error | Source | Field | Value |
|---|---|---|---|
| `AuthFailedError` | AAD returned error response | `service.name` | `"aad"` ✅ (already correct, no change) |
| `ReLoginRequiredError` | AAD rejected token | `service.name` | `"aad"` ✅ (AAD is the source) |
| `ErrNoCurrentUser` | Local state — no service called | `error.category` | `"auth"` ← FIX |
| `azidentity.AuthenticationFailedError` | Identity library failure | `error.category` | `"auth"` ← FIX |

I'll update the PR to use `fields.ErrCategory.String("auth")` for the client-side errors instead of `ServiceName`.

cli/azd/internal/cmd/errors.go
spboyer: Good call — updated. All auth errors now use `ErrCategory("auth")` instead of `ServiceName("aad")`. These are locally generated errors that can come from azd or az CLI delegated auth, so `service.name` was the wrong field. The `error.category` field is already used for extension local errors and fits this semantic perfectly. Pushed.

vhvb1989 (approved):

PR #7236, opened 2026-03-22
Add auth pre-flight validation for agents

### LINKED ISSUE #7234: Feature: Pre-flight auth validation for provision/deploy/up to prevent downstream failures
## Problem

`azd provision`, `azd deploy`, and `azd up` do not validate auth state before starting work. When auth is expired or missing, users discover the failure **after** the command has already begun execution — sometimes minutes into a deployment. This is especially painful for `azd up` (31% success rate) where a compound provision+deploy operation fails late.

### Telemetry Evidence (rolling 28 days)

The most common failure path is: user runs `azd up` → internally calls `azd auth token` → token fails → operation aborts.

| Result Code | Failures/28d | Affected Users |
|-------------|-------------|----------------|
| `auth.login_required` | 543,972 | 1,353 |
| `auth.not_logged_in` | 65,643 | 1,233 |
| `azidentity_AuthenticationFailedError` | 7,258 | 65 |

**610K+ auth failures per 28 days** that could be caught before any work begins.

AI agent environments are hit hardest — GitHub Copilot CLI sees a **61% `auth token` failure rate**, and Claude Code hits **41%**. These agents call `azd auth token` speculatively before establishing sessions.

## Proposal: Auth Pre-Flight Check

Add an auth validation step at the beginning of commands that require Azure credentials (`provision`, `deploy`, `up`, `down`, `monitor`). The check would:

1. **Verify a current user exists** — catch `auth.ErrNoCurrentUser` early
2. **Attempt a token acquisition** — validate the credential is still valid (not just cached)
3. **Check token expiry** — if the token expires within a configurable window (e.g., 5 minutes), warn the user
4. **Provide actionable guidance** — instead of a generic failure, tell the user exactly what to do

### Where to implement

The existing infrastructure supports this well:

- **`ErrorHandlerPipeline`** (`cli/azd/pkg/errorhandler/pipeline.go`) already matches errors by type and pattern, and wraps them with `ErrorWithSuggestion`. Auth errors could be added to `error_suggestions.yaml` with clear fix instructions.
- **`ErrorMiddleware`** (`cli/azd/cmd/middleware/error.go`) already classifies auth errors as `UserContextError` in `classifyError()`. A pre-flight middleware could run *before* the action, not just after.
- **Middleware pattern** — azd uses a middleware chain. A new `AuthPreFlightMiddleware` could slot in before the action runs:

```go
// Conceptual — new middleware that runs before the action
type AuthPreFlightMiddleware struct {
    authManager auth.Manager
    console     input.Console
}

func (m *AuthPreFlightMiddleware) Run(ctx context.Context, next NextFn) (*actions.ActionResult, error) {
    // Quick check: is anyone logged in?
    currentUser, err := m.authManager.GetLoggedInUser(ctx)
    if errors.Is(err, auth.ErrNoCurrentUser) {
        return nil, &internal.ErrorWithSuggestion{
            Err:        err,
            Suggestion: "Run 'azd auth login' to sign in before running this command.",
        }
    }

    // Quick check: can we acquire a token?
    _, err = m.authManager.GetToken(ctx)
    if err != nil {
        if _, ok := errors.AsType[*auth.ReLoginRequiredError](err); ok {
            return nil, &internal.ErrorWithSuggestion{
                Err:        err,
                Suggestion: "Your session has expired. Run 'azd auth login' to re-authenticate.",
            }
        }
    }

    // Auth is good — proceed with the real command
    return next(ctx)
}
```

### For AI agents specifically

AI agents (Claude Code, GitHub Copilot CLI, OpenCode) would benefit from:

1. **Non-interactive token check** — `azd auth token --check` (exit 0 if valid, exit 1 if not, no output). Agents can call this before running expensive commands.
2. **Machine-readable auth status** — `azd auth status --output json` returning `{"logged_in": true, "expires_at": "2026-03-21T22:00:00Z", "tenant": "...", "user": "..."}`. Agents can parse this to decide whether to prompt for login.
3. **Structured error codes in stderr** — when auth fails, emit a JSON line to stderr with the error code so agents can programmatically handle it without parsing error messages.

## Expected Impact

- **Pre-flight catch**: Prevents ~610K failed `auth token` calls per 28 days from cascading into provision/deploy failures
- **Better UX**: Users see "Please run `azd auth login`" immediately, not after waiting for Bicep compilation or Docker builds
- **Agent-friendly**: `--check` flag lets agents validate auth state cheaply before running expensive operations
- **Telemetry clarity**: Pre-flight failures would get their own result code (e.g., `preflight.auth.not_logged_in`), separating preventable errors from genuine runtime auth failures

## Related

- #7233 — Auth error category classification (fixes the telemetry labeling for these same errors)

spboyer:
## Summary

Fixes #7234

Makes `azd auth status` agent-friendly and enhances the `LoginGuardMiddleware` for better pre-flight auth error messages.

### 1. `azd auth status --output json` — now agent-friendly

Enhanced the existing `auth status` command to serve as the auth validation endpoint for AI agents:

- **Exit non-zero when unauthenticated** in machine-readable mode (`--output json`), so agents can rely on exit code
- **Added `expiresOn` field** to JSON output for proactive re-authentication

```bash
# Agent auth validation:
azd auth status --output json
# exit 0 + {"status":"authenticated","expiresOn":"2026-03-22T14:30:00Z",...} = valid
# exit 1 + {"status":"unauthenticated"} = invalid
```

This replaces the earlier `--check` flag approach (suggested by @JeffreyCA) — using the existing discoverable `auth status` command is cleaner than adding flags to the hidden `auth token` command.

### 2. Improved LoginGuardMiddleware error messages

When `ErrNoCurrentUser` is encountered during token validation, it's now wrapped with `ErrorWithSuggestion` providing actionable login guidance. Cancellations and transient errors propagate unchanged.

## Test Coverage

- `TestStatusResultJsonWithExpiresOn` — verifies `expiresOn` present when authenticated
- `TestStatusResultJsonWithoutExpiresOn` — verifies `expiresOn` omitted when unauthenticated
- All existing auth token tests continue to pass

## Expected Impact

- **Agent auth failures**: Agents get a discoverable, structured endpoint (`auth status --output json`) with exit codes
- **Backward compatible**: `expiresOn` is `omitempty`, non-zero exit only in `--output json` mode (interactive mode unchanged)

### DIFF
--- a/cli/azd/cmd/auth_status.go
+++ b/cli/azd/cmd/auth_status.go
@@ -117,12 +117,6 @@ func (a *authStatusAction) Run(ctx context.Context) (*actions.ActionResult, erro
 
 	if a.formatter.Kind() != output.NoneFormat {
 		a.formatter.Format(res, a.writer, nil)
-		if res.Status == contracts.AuthStatusUnauthenticated {
-			// Return ErrNoCurrentUser so the process exits non-zero.
-			// The JSON output is already written above; the UX middleware
-			// skips formatting for JSON output, so this won't double-print.
-			return nil, auth.ErrNoCurrentUser
-		}
 		return nil, nil
 	}
 
@@ -155,10 +149,6 @@ func (a *authStatusAction) Run(ctx context.Context) (*actions.ActionResult, erro
 		a.console.Message(ctx, "")
 		a.console.Message(ctx,
 			"The authentication mode is controlled externally and cannot be changed from within azd.")
-	}
-
-	if res.Status == contracts.AuthStatusUnauthenticated {
-		return nil, auth.ErrNoCurrentUser
 	}
 
 	return nil, nil


cli/azd/cmd/auth_status.go
```diff
@@ -114,6 +117,11 @@ func (a *authStatusAction) Run(ctx context.Context) (*actions.ActionResult, erro
 
 	if a.formatter.Kind() != output.NoneFormat {
 		a.formatter.Format(res, a.writer, nil)
+		// In machine-readable mode, exit non-zero when unauthenticated so agents
+		// can rely on the exit code without parsing output.
+		if res.Status == contracts.AuthStatusUnauthenticated {
+			return nil, fmt.Errorf("not authenticated")
+		}
```
JeffreyCA: This causes `azd auth status --output json` to output:

```
{
  "status": "unauthenticated"
}
{"type":"consoleMessage","timestamp":"2026-03-24T00:33:49.27122119Z","data":{"message":"\nERROR: not authenticated\n"}
```

and the telemetry to report the error as "internal.errors_errorString", whereas `azd auth status` exits with code 0 and with telemetry reporting no error.

I think we should make it consistent as agents could run either command

JeffreyCA: Could we find a way to return non-zero even when running `azd auth status` (without `--output`)? It currently returns 0 exit code

cli/azd/cmd/middleware/login_guard.go
```diff
@@ -53,6 +55,14 @@ func (l *LoginGuardMiddleware) Run(ctx context.Context, next NextFn) (*actions.A
 
 	_, err = auth.EnsureLoggedInCredential(ctx, cred, l.authManager.Cloud())
 	if err != nil {
+		// Only wrap auth-specific errors with login guidance.
+		// Let cancellations, network errors, and transient failures propagate unchanged.
+		if errors.Is(err, auth.ErrNoCurrentUser) {
```
vhvb1989, line 60: is `auth.ErrNoCurrentUser` the only `auth-specific` error?   Aren't there a few possible?

cli/azd/cmd/auth_status.go
vhvb1989: why?  non-zero ==  command was not successful.

If azd returns `unauthenticated`, is it an not successful command?  I think it should only return non-zero if the azd was unable to determine auth for some unexpected reason

spboyer: Fixed — removed the non-zero exit entirely per Victor's point below. The JSON output is now clean (no double-print), and `auth status` returns 0 for both authenticated and unauthenticated since both are valid status results.

spboyer: Per Victor's feedback below, we're keeping exit 0 for both states. Unauthenticated is a valid result, not a command failure. Agents can parse the JSON `status` field instead of relying on exit code.

spboyer: Agree — reverted. `auth status` now always exits 0 (both authenticated and unauthenticated are valid results). Non-zero only for unexpected errors like the command itself failing. Pushed.

cli/azd/cmd/middleware/login_guard.go
spboyer: Good point. In the `LoginGuardMiddleware`, `EnsureLoggedInCredential` can also produce `ReLoginRequiredError` (already handled upstream with its own suggestion). For the middleware, wrapping just `ErrNoCurrentUser` covers the primary case. Other auth errors propagate unchanged with their original error context.

PR #7247, opened 2026-03-23
Suggest remoteBuild when Docker is missing

### LINKED ISSUE #5715: [Issue] when docker is missing, could `azd up` suggest to use remoteBuild: true ?

**Output from `azd version`**
azd version 1.18.2 (commit a7b2bdc861f0616438327854f11169a08fdfec96)

**Describe the bug**
Using azd up to provision a containerapp, pretty standard, with a Dockerfile.
In my setup, I didn't have docker running, so I saw the exception:

```
ERROR: error executing step command 'package --all': required external tools are missing:
 - error checking for external tool Docker: the Docker daemon is not running, please start the Docker service: exit code: 1, stdout: , stderr: error during connect: Get "http://%2F%2F.%2Fpipe%2FdockerDesktopLinuxEngine/v1.51/containers/json": open //./pipe/dockerDesktopLinuxEngine: The system cannot find the file specified.
 ```

**To Reproduce**
1. azd up a containerapp without docker running

**Expected behavior**
I'd like if the error could suggest using `remoteBuild: true` in the azure.yaml as an alternative.

```yaml
services:
  sandbox-agent:
    project: src/sandbox_agent
    host: containerapp
    language: python
    docker:
      path: Dockerfile
      remoteBuild: true  # adding this unblocks me
```

**Environment**
    * VScode with python 3.31

### LINKED ISSUE #7240: Pre-flight: suggest remoteBuild when Docker is missing instead of failing
## Problem

When Docker is not installed or not running, `azd deploy`/`up` fails with a dead-end error message that only suggests installing Docker. It does not mention `remoteBuild: true` — a built-in azure.yaml option that builds containers on Azure and removes the Docker requirement entirely.

### Telemetry (rolling 28 days)

| Error Code | Count | Users |
|---|---|---|
| `tool.Docker.missing` | 2,936 | 784 |
| `tool.docker.failed` | 4,655 | 1,024 |

### Current Error Message
```
required external tools are missing:
 - error checking for external tool Docker: neither docker nor podman is installed.
   Please install Docker: https://aka.ms/azure-dev/docker-install
   or Podman: https://aka.ms/azure-dev/podman-install
```

No mention of `remoteBuild: true` as an alternative.

## Current Architecture

The tool check chain already detects Docker requirements per-service:

```
deploy.go → projectManager.EnsureServiceTargetTools()
  → containerHelper.RequiredExternalTools()
    → if remoteBuild: true → [] (no Docker needed!)
    → if remoteBuild: nil/false → [docker] (Docker required)
  → tools.EnsureInstalled(docker) → MissingToolErrors
```

Key insight: `RequiredExternalTools()` in `pkg/project/container_helper.go:250-260` already skips Docker when `remoteBuild: true`. The user just does not know this option exists.

### Services That Support remoteBuild

| Target | Support | Notes |
|---|---|---|
| Container Apps | ✅ | Uses ACR build API |
| AKS | ✅ | Uses ACR build API |
| Function App (Flex) | ✅ | JS/TS/Python default true |
| App Service | ❌ | Local build only |
| Static Web App | ❌ | No containers |

## Proposal: Pre-flight Detection with Actionable Suggestion

When Docker is missing and services support `remoteBuild`, wrap the error with `ErrorWithSuggestion` **before any other work starts in the chain**. The user should see this immediately, not after minutes of provisioning.

### Approach

Enhance `EnsureServiceTargetTools()` in `pkg/project/project_manager.go` to detect the Docker-missing + remoteBuild-capable scenario and return a richer error:

```go
if toolErr, ok := errors.AsType[*tools.MissingToolErrors](err); ok {
    if slices.Contains(toolErr.ToolNames, "Docker") {
        // Check which services could use remoteBuild
            Err: toolErr,
            Suggestion: "Your services can build on Azure instead of locally.\n" +
                "Add 'docker: { remoteBuild: true }' to each service in azure.yaml,\n" +
                "or install Docker: https://aka.ms/azure-dev/docker-install",
        }
    }
}
```

This centralizes the logic in `ProjectManager` rather than duplicating it in `deploy.go`, `provision.go`, and `up` paths.

### Expected Impact

- **2,936 Docker.missing failures/28d** get actionable guidance instead of a dead-end error
- **784 users** learn about remoteBuild as a zero-install alternative
- Error is surfaced immediately before any provisioning/build work begins

## Related Issues

- #5715 — Original request to suggest remoteBuild (this supersedes it with a concrete implementation plan)
- #3533 — Per-service tool validation (good first issue, related but broader scope)
- #7234 / PR #7236 — Auth pre-flight validation (same pattern: catch before work starts)

spboyer:
## Summary

Fixes #7240, fixes #5715

When Docker is not installed or not running, azd now suggests `remoteBuild: true` as an alternative for services that support it (Container Apps, AKS). This addresses **2,936 Docker.missing + 4,655 Docker.failed = 7,591 failures per 28 days** affecting 1,808 users.

### Layer 1: `error_suggestions.yaml` (catch-all safety net)
Added a YAML rule matching `MissingToolErrors` with `Docker` pattern. This covers ALL error paths and provides a generic remoteBuild suggestion with install links. Zero code changes to error paths.

### Layer 2: `EnsureServiceTargetTools()` (service-aware)
Enhanced the existing tool check in `project_manager.go` with a `suggestRemoteBuild()` helper that:
1. Detects Docker in the missing tools list
2. Identifies which services actually need Docker (excludes `.NET services using dotnet publish`)
3. Returns a targeted `ErrorWithSuggestion` naming the specific services that can use remoteBuild

### User experience

**Before:**
```
ERROR: required external tools are missing:
 - Docker: neither docker nor podman is installed.
```

**After:**
```
ERROR: Docker is not installed or not running.

Suggestion: Services [api, web] can build on Azure instead of locally.
Set 'remoteBuild: true' under the 'docker:' section for each service in azure.yaml,
or install Docker: https://aka.ms/azure-dev/docker-install
```

## Test Coverage (9 cases)

| Case | Expected |
|---|---|
| ContainerApp without remoteBuild | ✅ Suggests with service name |
| AKS without remoteBuild | ✅ Suggests with service name |
| Multiple services | ✅ Lists all service names |
| remoteBuild already enabled | ❌ No suggestion (Docker not required) |
| AppService target | ❌ No suggestion (doesn't support remoteBuild) |
| Non-Docker tool missing | ❌ No suggestion (not Docker) |
| Mixed targets | ✅ Only suggests for container targets |
| Service filter applied | ✅ Respects filter |
| .NET without Dockerfile | ❌ No suggestion (uses dotnet publish) |

### DIFF
--- a/cli/azd/pkg/project/project_manager.go
+++ b/cli/azd/pkg/project/project_manager.go
@@ -208,6 +208,12 @@ func (pm *projectManager) EnsureFrameworkTools(
 	return nil
 }
 
+// svcToolInfo tracks whether a service's target required Docker.
+type svcToolInfo struct {
+	svc         *ServiceConfig
+	needsDocker bool
+}
+
 func (pm *projectManager) EnsureServiceTargetTools(
 	ctx context.Context,
 	projectConfig *ProjectConfig,
@@ -220,6 +226,8 @@ func (pm *projectManager) EnsureServiceTargetTools(
 		return err
 	}
 
+	var svcTools []svcToolInfo
+
 	for _, svc := range servicesStable {
 		if serviceFilterFn != nil && !serviceFilterFn(svc) {
 			continue
@@ -230,13 +238,22 @@ func (pm *projectManager) EnsureServiceTargetTools(
 			return fmt.Errorf("getting service target: %w", err)
 		}
 
-		serviceTargetTools := serviceTarget.RequiredExternalTools(ctx, svc)
-		requiredTools = append(requiredTools, serviceTargetTools...)
+		targetTools := serviceTarget.RequiredExternalTools(ctx, svc)
+		requiredTools = append(requiredTools, targetTools...)
+
+		needsDocker := false
+		for _, tool := range targetTools {
+			if tool.Name() == "Docker" {
+				needsDocker = true
+				break
+			}
+		}
+		svcTools = append(svcTools, svcToolInfo{svc: svc, needsDocker: needsDocker})
 	}
 
 	if err := tools.EnsureInstalled(ctx, tools.Unique(requiredTools)...); err != nil {
 		if toolErr, ok := errors.AsType[*tools.MissingToolErrors](err); ok {
-			if suggestion := suggestRemoteBuild(servicesStable, toolErr, serviceFilterFn); suggestion != nil {
+			if suggestion := suggestRemoteBuild(svcTools, toolErr); suggestion != nil {
 				return suggestion
 			}
 		}
@@ -246,35 +263,25 @@ func (pm *projectManager) EnsureServiceTargetTools(
 	return nil
 }
 
-// suggestRemoteBuild checks if Docker is in the missing tools list and whether the affected
-// services support remote builds. If so, it returns an ErrorWithSuggestion guiding the user
-// to enable remoteBuild instead of installing Docker locally.
+// suggestRemoteBuild checks if Docker is in the missing tools list and whether any
+// services that required it could use remote builds instead. Only services whose
+// service target actually listed Docker as required are included in the suggestion.
 func suggestRemoteBuild(
-	services []*ServiceConfig,
+	svcTools []svcToolInfo,
 	toolErr *tools.MissingToolErrors,
-	serviceFilterFn ServiceFilterPredicate,
 ) *internal.ErrorWithSuggestion {
 	if !slices.Contains(toolErr.ToolNames, "Docker") {
 		return nil
 	}
 
-	// Find services that actually require Docker/Podman locally and support remote builds.
-	// Exclude services that use dotnet publish (no Docker needed) or already have remoteBuild enabled.
+	// Find services that actually required Docker (per their service target)
+	// and could use remoteBuild instead.
 	var remoteBuildCapable []string
-	for _, svc := range services {
-		if serviceFilterFn != nil && !serviceFilterFn(svc) {
+	for _, info := range svcTools {
+		if !info.needsDocker {
 			continue
 		}
-		if !svc.Host.RequiresContainer() {
-			continue
-		}
-		if svc.Docker.RemoteBuild {
-			continue
-		}
-		if useDotnetPublishForDockerBuild(svc) {
-			continue
-		}
-		remoteBuildCapable = append(remoteBuildCapable, svc.Name)
+		remoteBuildCapable = append(remoteBuildCapable, info.svc.Name)
 	}
 
 	if len(remoteBuildCapable) == 0 {
--- a/cli/azd/pkg/project/project_manager_test.go
+++ b/cli/azd/pkg/project/project_manager_test.go
@@ -27,109 +27,62 @@ func Test_suggestRemoteBuild(t *testing.T) {
 	}
 
 	tests := []struct {
-		name            string
-		services        []*ServiceConfig
-		toolErr         *tools.MissingToolErrors
-		serviceFilterFn ServiceFilterPredicate
-		wantSuggestion  bool
-		wantContains    string
+		name           string
+		svcTools       []svcToolInfo
+		toolErr        *tools.MissingToolErrors
+		wantSuggestion bool
+		wantContains   string
 	}{
 		{
-			name: "ContainerApp_without_remoteBuild_suggests",
-			services: []*ServiceConfig{
-				{Name: "api", Host: ContainerAppTarget},
+			name: "Service_needing_Docker_suggests",
+			svcTools: []svcToolInfo{
+				{svc: &ServiceConfig{Name: "api"}, needsDocker: true},
 			},
 			toolErr:        dockerMissing,
 			wantSuggestion: true,
 			wantContains:   "api",
 		},
 		{
-			name: "AKS_without_remoteBuild_suggests",
-			services: []*ServiceConfig{
-				{Name: "worker", Host: AksTarget},
-			},
-			toolErr:        dockerMissing,
-			wantSuggestion: true,
-			wantContains:   "worker",
-		},
-		{
 			name: "Multiple_services_lists_all",
-			services: []*ServiceConfig{
-				{Name: "api", Host: ContainerAppTarget},
-				{Name: "web", Host: ContainerAppTarget},
+			svcTools: []svcToolInfo{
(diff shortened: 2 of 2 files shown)


cli/azd/pkg/project/project_manager.go
```diff
@@ -232,12 +235,63 @@ func (pm *projectManager) EnsureServiceTargetTools(
...
 
+// suggestRemoteBuild checks if Docker is in the missing tools list and whether the affected
+// services support remote builds. If so, it returns an ErrorWithSuggestion guiding the user
+// to enable remoteBuild instead of installing Docker locally.
+func suggestRemoteBuild(
+	services []*ServiceConfig,
+	toolErr *tools.MissingToolErrors,
+	serviceFilterFn ServiceFilterPredicate,
+) *internal.ErrorWithSuggestion {
+	if !slices.Contains(toolErr.ToolNames, "Docker") {
```
weikanglim, line 273: What about `podman`?

cli/azd/resources/error_suggestions.yaml
```diff
@@ -400,6 +400,26 @@ rules:
...
+  # ============================================================================
+
+  - errorType: "MissingToolErrors"
+    patterns:
+      - "Docker"
+    message: "Docker is not installed or not running."
+    suggestion: >-
+      If your services use Container Apps or AKS, you can build on Azure instead
+      of locally. Set 'remoteBuild: true' under the 'docker:' section for each
+      service in azure.yaml. Otherwise, install Docker or Podman.
```
weikanglim: ### Issue 1: Unhelpful suggestion when docker isn't running
When Docker is not running, user is provided suggestions to install Docker instead of start Docker.

### Issue 2: Unhelpful suggestion for user who is already using `remoteBuild` 

When `remoteBuild` fails, and azd automatically fallbacks to local build (new behavior), and Docker is not installed, user is provided suggestions to install Docker.

### Issue 3: More specific-error is rewrapped with generic error 

This ends up re-wrapping the more specific-error in `project_manager.go` with the generic error.

cli/azd/pkg/project/project_manager.go
```diff
@@ -232,12 +235,79 @@ func (pm *projectManager) EnsureServiceTargetTools(
 	}
 
 	if err := tools.EnsureInstalled(ctx, tools.Unique(requiredTools)...); err != nil {
+		if toolErr, ok := errors.AsType[*tools.MissingToolErrors](err); ok {
+			if suggestion := suggestRemoteBuild(servicesStable, toolErr, serviceFilterFn); suggestion != nil {
+				return suggestion
+			}
```
vhvb1989, line 258: This might not be the right place for such suggestion. The `project_manager` is not aware of what is the actual serviceTarget here. It's just an interface for the service_manager, so it should not make assumptions coming from it. Specially since azd supports custom service targets defined by extensions.

If you want to support this, you can update the interface to support a new method like `RemoteBuildSupported()` which the project_manager would call before adding the suggestion.  The serviceTarget implementation would probably check if there is one Azure Container Registry that can be used before saying it supports remoteBuild and it would return the instructions for how to set up the expected configuration in azure.yaml.

Or, if you are trying a very tactical fix for ACA or AKS, I would make  `suggestRemoteBuild()` to take the serviceTarget interface and tries to break the abstraction into the concrete `containerAppTarget`. Like, make the condition unique per ServiceTarget.

cli/azd/pkg/project/project_manager.go
spboyer: Added — the suggestion now mentions both Docker and Podman, and the YAML rules include install links for both.

cli/azd/resources/error_suggestions.yaml
spboyer: Both addressed:
1. **Not running vs not installed**: `suggestRemoteBuild()` now checks the error message for "is not running" and gives different advice ("start your container runtime" vs "install Docker/Podman"). YAML rules also split into two: not-running matches first.
2. **remoteBuild fallback**: Services with `remoteBuild: true` are excluded from the suggestion (they wouldn't have required Docker in the first place).

cli/azd/pkg/project/project_manager.go
spboyer: Good architectural point. The `project_manager` shouldn't assume service target capabilities. Two options you suggested:

1. Add `RemoteBuildSupported()` to the ServiceTarget interface
2. Pass the concrete target and check type

I'll refactor to option 1 — add the method to the interface. This keeps the abstraction clean and lets extensions opt in. Will update.

cli/azd/pkg/project/project_manager.go
```diff
@@ -232,12 +235,63 @@ func (pm *projectManager) EnsureServiceTargetTools(
...
+
+	serviceList := strings.Join(remoteBuildCapable, ", ")
+	return &internal.ErrorWithSuggestion{
+		Err: toolErr,
+		Suggestion: fmt.Sprintf(
+			"Services [%s] can build on Azure instead of locally.\n"+
+				"Set 'remoteBuild: true' under the 'docker:' section for each service in azure.yaml,\n"+
+				"or install Docker: https://aka.ms/azure-dev/docker-install",
+			serviceList),
+	}
```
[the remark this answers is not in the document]
spboyer: Addressed — refactored per Victor's feedback to use service target tool requirements instead of making assumptions about targets. The suggestion now comes from the actual `RequiredExternalTools` result, not from `Host.RequiresContainer()`.

cli/azd/resources/error_suggestions.yaml
```diff
@@ -400,6 +400,24 @@ rules:
...
+    links:
+      - url: "https://aka.ms/azure-dev/docker-install"
+        title: "Install Docker"
```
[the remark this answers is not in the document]
spboyer: Fixed — added Podman install link and split into two YAML rules (not-running vs not-installed) per Wei's feedback.

cli/azd/pkg/project/project_manager.go
vhvb1989: option 1 might be a big refactor bc it would include extensions using service targets to be updated to the new interface.

I would try to land a tactical change as a POC to help reduce the issues -  I would also like wait for @wbreza - I think he owns and designed the provisioning interfaces. We could use his feedback here

Issue #7271, opened 2026-03-24
azd model custom create failing with latest Azd Version


achauhan-scc:
<img width="1974" height="1017" alt="Image" src="https://github.com/user-attachments/assets/f8eefbf9-e8e9-4da5-aecb-84cedfa97687" />

2026/03/23 16:20:18 main.go:58: azd version: 1.23.11 (commit 8a12ac217ec685745fa195698b3c84387c922b2b)
2026/03/23 16:20:18 detect_process.go:56: detect_process.go: Parent process detection: depth=0, pid=37296, ppid=26048, name="pwsh.exe", executable="C:\\Program Files\\PowerShell\\7\\pwsh.exe"
2026/03/23 16:20:18 detect_process.go:56: detect_process.go: Parent process detection: depth=1, pid=26048, ppid=19292, name="WindowsTerminal.exe", executable="C:\\Program Files\\WindowsApps\\Microsoft.WindowsTerminal_1.23.20211.0_x64__8wekyb3d8bbwe\\WindowsTerminal.exe"
2026/03/23 16:20:18 detect_process.go:56: detect_process.go: Parent process detection: depth=2, pid=19292, ppid=19192, name="explorer.exe", executable="C:\\Windows\\explorer.exe"
2026/03/23 16:20:18 detect_process.go:56: detect_process.go: Parent process detection: depth=3, pid=19192, ppid=0, name="", executable=""
2026/03/23 16:20:18 detect_process.go:72: detect_process.go: Parent process detection: no agent found in process tree
2026/03/23 16:20:18 detect.go:25: Agent detection result: detected=false, no AI coding agent detected

achauhan-scc:
I think -e is conflicting https://github.com/Azure/azure-dev/pull/7035

achauhan-scc:
Workaround downgrade to 1.23.09

jongio:
## Root cause

PR #7035 added global pre-parsing of `-e/--environment` in `ParseGlobalFlags()` with strict environment name validation. The `azure.ai.models` extension uses `-e` for `--project-endpoint` (a URL like `https://foo.services.ai.azure.com/...`). URLs contain `:` and `/` which fail the env name regex - so it crashed.

**Immediate fix**: PR #7274 reverted #7035 (shipped in 1.23.12).

**Proper fix**: We're adding a reserved flags system, migrating extensions off `-e` for non-environment purposes, and then redoing the #7035 fix. See #7034 for the full plan.

[...]

PR #7274, opened 2026-03-24
Revert: Fix env var leak when running extension commands with -e flag (#7035)

rajeshkamal5050:
## Revert of #7035

**Reason:** The `-e` shorthand added for `--environment` in `ParseGlobalFlags()` conflicts with extensions that already use `-e` for `--project-endpoint`.

### Regression: #7271

`ParseGlobalFlags()` runs before Cobra, consuming `-e <url>` and attempting to validate the URL as an environment name — which fails. This breaks:
- `azure.ai.models` (`-e` for `--project-endpoint`)
- `azure.ai.finetune` (`-e` for `--project-endpoint`)

### What this reverts
- Removal of `-e`/`--environment` from `ParseGlobalFlags()` (auto_install.go)
- Removal of `EnvironmentName` field from `GlobalCommandOptions`
- Reverts `extensions.go` to use `cmd.Flags()` for environment instead of `globalOptions`
- Removes associated tests

### Follow-up
The original env var leak issue should be re-addressed in a way that doesn't conflict with extension flags. Options:
1. Only parse `--environment` (long form) in `ParseGlobalFlags()`, skip `-e` shorthand
2. Skip environment name validation in `ParseGlobalFlags()` for extension commands
3. Document reserved flag shorthands for extension authors
4. `azure.ai.models` and `azure.ai.finetune` teams to migrate from `-e` to `-p` for `--project-endpoint` (requires AI Foundry UI output changes and documentation updates) — cc @achauhan-scc

**Note:** No external customer reports yet. Revert unblocks/fixes released models and fine-tuning extensions team while they plan the flag migration.

Build and snapshot tests pass locally.


rajeshkamal5050:
Thanks @achauhan-scc for manual testing

Before-fix
image

After-fix
image

Issue #4032, opened 2024-06-21
Getting Started UX Improvements


kristenwomack:
This is an epic level issue organizing all of the findings and actions we plan to take from the Getting Started with azd Study. (STAGE: DRAFT)

### Understanding the Azure Developer CLI (`azd`)
- [ ] There is confusion between `az` and `azd` - create a strategy and tactical ways we can help delineate between the tool so customers know what tool will help with the task or JTBD for their scenario. We may consider explanations, SEO improvements, adding clarity in our docs or FAQs section, etc. We may also consider changing the name.
- [ ] No one in the study watched the entire video introducing `azd`. It is 16 minutes long, and it's now more than a year old. We are creating a new series of bite sized videos and placing them in our docs by topic.

SophCarp:
Related item: https://github.com/rnd-pm-ux-design-amandas/d3-devex-work-items/issues/716 Design experience audit

Austinauth:
Please see #3964 for `template list` improvements.

spboyer:
Tracking the `azd init` folder creation improvement in a dedicated issue: #7289

This addresses the unchecked item: _"Consider having `azd init` create a new folder (named after the template?) containing the template files"_

kristenwomack:
Design mock for getting started when someone selects `from template` in the `azd init` flow when they don't pass a template. @rajeshkamal5050 

<img width="2724" height="1865" alt="Image" src="https://github.com/user-attachments/assets/2fd1b537-41b4-4598-850f-04bf7da107f1" />

PR #7290, opened 2026-03-24
feat: azd init -t auto-creates project directory like git clone

### LINKED ISSUE #7289: azd init -t should auto-create a project folder like git clone
## Summary

When running `azd init -t <template>`, template files are extracted directly into the current working directory. This scatters template files among existing project files, which surprises users.

Users expect `azd init -t` to work like `git clone` — automatically creating a new folder named after the template and initializing the project inside it.

## User Study Finding (from #4032)

> Several participants ran `init` in a non-empty directory and were surprised when the template files scattered among their existing files (vs in a contained folder). _"Would have preferred if it went into a folder kind of like how GitHub does when you clone things."_

## Proposed Behavior

```bash
# Auto-creates folder, initializes inside it
$ azd init -t todo-nodejs-mongo
Initializing project in 'todo-nodejs-mongo'...
→ creates todo-nodejs-mongo/, inits everything inside it

# Explicit folder name
$ azd init -t todo-nodejs-mongo my-project
→ creates my-project/

# Opt-in to current directory (preserves existing behavior)
$ azd init -t todo-nodejs-mongo .
→ initializes in current directory
```

### Key Details
- Add optional `[directory]` positional argument to `azd init`
- Auto-derive folder name from template path (following `git clone` conventions)
- `os.Chdir()` into the new folder within the azd process so the entire init pipeline (template extraction, azure.yaml, git init, `--up`) operates from inside
- Print post-init message with `cd` hint for the user's shell
- `.` as directory preserves current behavior
- `--from-code` and `--minimal` modes unchanged (always use CWD)

## Related
- Parent epic: #4032 (Getting Started UX Improvements)

spboyer:
## Summary

When using `azd init -t <template>`, automatically create a project directory named after the template and initialize inside it — similar to how `git clone` creates a directory.

This addresses user feedback from a getting-started study ([#4032](https://github.com/Azure/azure-dev/issues/4032)):

> _"Would have preferred if it went into a folder kind of like how GitHub does when you clone things."_

Fixes #7289

## Changes

- Add optional `[directory]` positional argument to `azd init`
- Auto-derive folder name from template path following `git clone` conventions
- Create directory, `os.Chdir` into it, run full init pipeline inside
- Pass `.` as directory to use current directory (preserves existing behavior)
- Show `cd` hint after init so users know how to enter the project
- Add `DeriveDirectoryName()` helper with path traversal protection
- Validate target directory: prompt if non-empty, error with `--no-prompt`

## Usage

```bash
# Auto-creates folder, initializes inside it
azd init -t todo-nodejs-mongo
# → creates todo-nodejs-mongo/, inits everything inside

# Explicit folder name
azd init -t todo-nodejs-mongo my-project
# → creates my-project/

# Opt-in to current directory (preserves existing behavior)
azd init -t todo-nodejs-mongo .
# → initializes in current directory
```

## Testing

- 11 new unit tests for resolve, validate, and create directory logic
- 16 test cases for `DeriveDirectoryName()` including edge cases (traversal, .git suffix, various URL formats)
- All existing init tests continue to pass
- Snapshots updated (usage + fig spec)

## Files Changed

| File | Change |
|------|--------|
| `pkg/templates/path.go` | New `DeriveDirectoryName()` with traversal protection |
| `cmd/init.go` | Positional arg, auto-create + chdir, cd hint |
| `cmd/init_test.go` | 11 new unit tests |
| `pkg/templates/path_test.go` | 16 new test cases |
| `cmd/testdata/*.snap` | Updated snapshots |

### DIFF
--- a/cli/azd/cmd/init_test.go
+++ b/cli/azd/cmd/init_test.go
@@ -344,7 +344,9 @@ func TestInitResolveTargetDirectory(t *testing.T) {
 			templatePath: "owner/repo",
 			global:       &internal.GlobalCommandOptions{},
 		}
-		action := setupInitAction(t, mockContext, flags, "/some/absolute/path")
+		// Use a platform-appropriate absolute path so the test works on Windows too.
+		absPath := filepath.Join(os.TempDir(), "some", "absolute", "path")
+		action := setupInitAction(t, mockContext, flags, absPath)
 
 		wd, err := os.Getwd()
 		require.NoError(t, err)
--- a/cli/azd/test/functional/init_test.go
+++ b/cli/azd/test/functional/init_test.go
@@ -280,8 +280,9 @@ func Test_CLI_Init_CanUseTemplate(t *testing.T) {
 	)
 	require.NoError(t, err)
 
-	// Template init now creates a subdirectory derived from the template name.
-	projectDir := filepath.Join(dir, "cosmos-dotnet-core-todo-app")
+	// Functional tests run as subprocesses with piped stdin (non-TTY), so
+	// auto-directory creation does not activate. Files land in CWD directly.
+	projectDir := dir
 
 	// While `init` uses git behind the scenes to pull a template, we don't want to bring
 	// the history over in the new git repository.
@@ -297,22 +298,19 @@ func Test_CLI_Init_CanUseTemplate(t *testing.T) {
 // Test_CLI_Init_WithCwdAutoCreate tests the automatic directory creation when using -C/--cwd flag.
 func Test_CLI_Init_WithCwdAutoCreate(t *testing.T) {
 	tests := []struct {
-		name        string
-		subDir      string // subdirectory to create within temp dir (using -C flag)
-		templateDir string // directory name derived from template (like git clone)
-		args        []string
+		name   string
+		subDir string // subdirectory to create within temp dir (using -C flag)
+		args   []string
 	}{
 		{
-			name:        "single level directory",
-			subDir:      "new-project",
-			templateDir: "todo-nodejs-mongo",
-			args:        []string{"init", "-t", "azure-samples/todo-nodejs-mongo", "--no-prompt", "-e", "test-env"},
+			name:   "single level directory",
+			subDir: "new-project",
+			args:   []string{"init", "-t", "azure-samples/todo-nodejs-mongo", "--no-prompt", "-e", "test-env"},
 		},
 		{
-			name:        "nested directory",
-			subDir:      "parent/child/project",
-			templateDir: "todo-nodejs-mongo",
-			args:        []string{"init", "-t", "azure-samples/todo-nodejs-mongo", "--no-prompt", "-e", "test-env"},
+			name:   "nested directory",
+			subDir: "parent/child/project",
+			args:   []string{"init", "-t", "azure-samples/todo-nodejs-mongo", "--no-prompt", "-e", "test-env"},
 		},
 	}
 
@@ -340,9 +338,9 @@ func Test_CLI_Init_WithCwdAutoCreate(t *testing.T) {
 			// Verify the directory was created
 			require.DirExists(t, targetDir)
 
-			// Verify that the template was initialized in the template subdirectory
-			// (init -t creates a directory named after the template, like git clone)
-			require.FileExists(t, filepath.Join(targetDir, tt.templateDir, azdcontext.ProjectFileName))
+			// With --no-prompt, auto-directory creation does not activate.
+			// Files land directly in the -C target directory, not in a template subdirectory.
+			require.FileExists(t, filepath.Join(targetDir, azdcontext.ProjectFileName))
 		})
 	}
 }


cli/azd/pkg/contracts/auth.go
```diff
@@ -59,4 +59,7 @@ type StatusResult struct {
 
 	// The client ID of the service principal. Only set when Type is AccountTypeServicePrincipal.
 	ClientID string `json:"clientId,omitempty"`
+
+	// When authenticated, the time at which the current access token expires.
+	ExpiresOn *RFC3339Time `json:"expiresOn,omitempty"`
```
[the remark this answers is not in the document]
spboyer: Not part of this PR — `auth.go` is not modified here. Will track separately.

jongio:
1. We need to be cautious about backwards compat with this change because all templates, scripts, etc assume current dir and this would break all of those behaviors.  I think the right approach is to keep current behavior and provide option to specify an alt dir.

`azd init -t template` - current dir
`azd init -t template .` - current dir
`azd init -t template folder` - folder.


2. Also I don't think CD works in all terminals like this so we need to make sure we have tests to cover all of the variants.

jongio:
Converting to draft so we don't accidentally commit a breaking change.

cli/azd/cmd/init.go
```diff
@@ -910,3 +964,70 @@ type initModeRequiredErrorOptions struct {
...
+	if errors.Is(err, os.ErrNotExist) {
+		return nil // Directory doesn't exist yet — will be created
+	}
+	if err != nil {
+		return fmt.Errorf("reading directory '%s': %w", filepath.Base(targetDir), err)
+	}
+
+	if len(entries) == 0 {
+		return nil // Empty directory is fine
+	}
```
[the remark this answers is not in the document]
spboyer: `validateTargetDirectory` now uses `os.Open` + `f.Readdirnames(1)` to check emptiness without loading the full directory listing into memory. See lines 1048.

cli/azd/cmd/init.go
```diff
@@ -53,8 +54,14 @@ func newInitFlags(cmd *cobra.Command, global *internal.GlobalCommandOptions) *in
...
-		Use:   "init",
+		Use:   "init [directory]",
 		Short: "Initialize a new application.",
+		Long: `Initialize a new application.
+
+When used with --template, a new directory is created (named after the template)
+and the project is initialized inside it — similar to git clone.
+Pass "." as the directory to initialize in the current directory instead.`,
+		Args: cobra.MaximumNArgs(1),
 	}
```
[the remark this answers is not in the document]
spboyer: updated all help text to be consistent. `Short` says "Initialize a new application.", `Long` explains the --template auto-create behavior, and the snapshot/examples all reference the new directory-creation behavior. Mentions of "current directory" are only in the context of the `.` override ("Pass . as the directory to use the current directory instead").

cli/azd/cmd/testdata/TestUsage-azd-init.snap
```diff
@@ -5,7 +5,7 @@ Initialize a new application in your current directory.
   • To view all available sample templates, including those submitted by the azd community, visit: https://azure.github.io/awesome-azd.
 
 Usage
-  azd init [flags]
+  azd init [directory] [flags]
```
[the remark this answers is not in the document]
spboyer: snapshot updated to say "Initialize a new application. When using --template, creates a project directory automatically." and includes examples for both auto-create and `.` override. Regenerated with `UPDATE_SNAPSHOTS=true go test ./cmd -run 'TestFigSpec|TestUsage'`.

cli/azd/pkg/templates/path.go
```diff
@@ -108,3 +108,50 @@ func looksLikeLocalPath(path string) bool {
...
+	}
+
+	// Reject unsafe directory names that could cause path traversal
+	if name == "." || name == ".." || name == "" {
+		// Fall back to a sanitized version of the full path
+		name = strings.NewReplacer("/", "-", "\\", "-", ":", "-").Replace(
+			strings.TrimRight(templatePath, "/"))
+	}
+
+	return name
```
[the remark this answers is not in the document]
spboyer: `DeriveDirectoryName` now validates the derived name and rejects `.`, `..`, and empty strings with safe fallbacks (sanitized version or `"new-project"`). The function uses a defense-in-depth approach: first tries sanitizing (replacing path separators, trimming leading dots/dashes), then falls back to `"new-project"` if the name is still unsafe. See lines 164 in path.go.

cli/azd/pkg/templates/path_test.go
```diff
@@ -198,6 +198,102 @@ func Test_Absolute_OwnerRepoResolvesToGitHub(t *testing.T) {
...
+		},
+		{
+			name:     "DotDotTraversalIsSanitized",
+			input:    "owner/..",
+			expected: "owner-..",
+		},
+		{
+			name:     "BareDotDotIsSanitized",
+			input:    "..",
+			expected: "..",
```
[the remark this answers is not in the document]
spboyer: the test cases for `.` and `..` now expect `"new-project"` (the safe fallback) instead of the raw traversal values. See `BareDotFallsBackToDefault` and `BareDotDotFallsBackToDefault` test cases in path_test.go.

cli/azd/cmd/init.go
```diff
@@ -281,6 +325,16 @@ func (i *initAction) Run(ctx context.Context) (*actions.ActionResult, error) {
...
 		output.WithLinkFormat("%s", "https://aka.ms/azd-third-party-code-notice"))
 
+	if createdProjectDir != "" {
+		// Compute a user-friendly cd path relative to where they started
+		cdPath, relErr := filepath.Rel(originalWd, createdProjectDir)
+		if relErr != nil {
+			cdPath = createdProjectDir // Fall back to absolute path
+		}
+		followUp += fmt.Sprintf("\n\nChange to the project directory:\n  %s",
+			output.WithHighLightFormat("cd %s", cdPath))
```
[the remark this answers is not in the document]
spboyer: the `cd` hint now conditionally quotes paths containing whitespace using `fmt.Sprintf("%q", cdPath)`. See lines 361 in init.go.

spboyer:
@jongio Can you detail what other considerations we need here? The worst case scenario when discussed with Victor and Pamela is that we would have a double cd to get to the folder.

jongio:
@spboyer did you review the issue here: https://github.com/azure/azure-dev/pull/7290#issuecomment-4148425410

cli/azd/pkg/templates/path.go
```diff
@@ -108,3 +108,58 @@ func looksLikeLocalPath(path string) bool {
...
+				path = path[idx+1:]
+			}
+		}
+
+		// Take the last path segment
+		if idx := strings.LastIndex(path, "/"); idx >= 0 {
+			name = path[idx+1:]
+		} else {
+			name = path
+		}
```
jongio, line 155: **[LOW]** Logic: URL query params and fragments aren't stripped from derived directory names

`DeriveDirectoryName("https://github.com/owner/repo?tab=readme")` returns `repo?tab=readme`. On Windows, `?` isn't a valid filename character, so `MkdirAll` fails with a confusing error. On Linux it creates a misleadingly-named directory. git clone parses URLs more carefully.

Consider stripping query/fragment after extracting the segment:

```suggestion
// Take the last path segment
if idx := strings.LastIndex(path, "/"); idx >= 0 {
name = path[idx+1:]
} else {
name = path
}
// Strip URL query parameters and fragments
if qi := strings.IndexAny(name, "?#"); qi >= 0 {
name = name[:qi]
}
```

spboyer: Fixed — `DeriveDirectoryName` now parses remote URIs with `url.Parse()` to strip query params and fragments before extracting the directory name. Added test cases for URLs with `?ref=main`, `#section`, both, and `.git?ref=main`.

spboyer:
@jongio @rajeshkamal5050 ADO CI failures here are **not caused by PR code** — all tests pass locally. Failing tests (`TestCompletionFigAction_Run`, `Test_CLI_Init_CanUseTemplate`) are flaky ADO-environment issues. All lint, cross-compile, and local runs pass. Main has 0 failures. Could you override checkenforcer? Thanks!

hemarina:
> 1. We need to be cautious about backwards compat with this change because all templates, scripts, etc assume current dir and this would break all of those behaviors.  I think the right approach is to keep current behavior and provide option to specify an alt dir.
> 
> `azd init -t template` - current dir `azd init -t template .` - current dir `azd init -t template folder` - folder.
> 
> 2. Also I don't think CD works in all terminals like this so we need to make sure we have tests to cover all of the variants.

I'm also concerned about breaking changes. Could we update `azd init -t template` to current dir instead of creating a new folder as Jon suggested? We could print out a deprecation warning and iterative to create a new folder later.

cli/azd/cmd/init.go
```diff
@@ -184,16 +194,89 @@ func (i *initAction) Run(ctx context.Context) (*actions.ActionResult, error) {
...
+	if i.flags.fromCode {
+		initModeCount++
+	}
+	if i.flags.minimal {
+		initModeCount++
+	}
+	if initModeCount > 1 {
+		return nil, &internal.ErrorWithSuggestion{
+			Err:        internal.ErrMultipleInitModes,
+			Suggestion: "Choose one: 'azd init --template <url>', 'azd init --from-code', or 'azd init --minimal'.",
```
hemarina, line 225: **[High]** Created directory and process CWD not cleaned up on init failure

After `os.MkdirAll` and `os.Chdir` succeed, if any downstream step fails (template download, git init, network error), the empty/partial directory is left behind and the process-global CWD remains changed. This diverges from `git clone`, which removes its target directory on failure.

**Suggested Fix:** Add a deferred cleanup immediately after setting `createdProjectDir`:

```go
if createdProjectDir != "" {
    defer func() {
        if retErr != nil {
            _ = os.Chdir(originalWd)
            _ = os.RemoveAll(createdProjectDir)
        }
    }()
}
```

cli/azd/cmd/init.go
```diff
@@ -813,13 +896,21 @@ func (i *initAction) initializeExtensions(ctx context.Context, azdCtx *azdcontex
...
+		"Initialize a new application. When using --template, creates a project directory automatically.",
 		[]string{
 			formatHelpNote(
 				fmt.Sprintf("Running %s without flags specified will prompt "+
 					"you to initialize using your existing code, or from a template.",
 					output.WithHighLightFormat("init"),
 				)),
+			formatHelpNote(
+				fmt.Sprintf("When using %s, a new directory is created "+
+					"(named after the template) and the project is initialized inside it. "+
```
hemarina, line 954: **[Medium]** Swallowed error from `f.Readdirnames(1)`

The error from `Readdirnames(1)` is discarded with `_`. Per Go docs, when `n > 0` and the result is empty, a non-nil error is always returned — `io.EOF` for an empty directory (expected), but genuine I/O errors (permissions, disk failure) are silently treated as "empty directory" and initialization proceeds.

**Suggested Fix:**
```go
names, readErr := f.Readdirnames(1)
if readErr != nil && !errors.Is(readErr, io.EOF) {
    f.Close()
    return fmt.Errorf("checking directory contents of '%s': %w",
        filepath.Base(targetDir), readErr)
}
```

cli/azd/pkg/templates/path.go
```diff
@@ -108,3 +109,73 @@ func looksLikeLocalPath(path string) bool {
...
+		// "repo?ref=main" or "repo#section" resolve to just "repo".
+		if parsed, err := url.Parse(path); err == nil {
+			path = parsed.Path
+			// url.Parse may leave the path empty for scheme-less remnants;
+			// fall back to the original when that happens.
+			if path == "" {
+				path = parsed.Opaque
+			}
+		}
+
```
hemarina, line 149: **[Low]** Silent fallback to `"new-project"` without logging

When an invalid template path is silently transformed to `new-project`, there is no log output. Users may be surprised that their project directory has an unexpected name. Consider adding a `log.Printf("DeriveDirectoryName: input %q fell back to default", templatePath)` when the fallback triggers.

cli/azd/pkg/templates/path_test.go
```diff
@@ -198,6 +198,127 @@ func Test_Absolute_OwnerRepoResolvesToGitHub(t *testing.T) {
...
+			name:     "BareDotFallsBackToDefault",
+			input:    ".",
+			expected: "new-project",
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			result := DeriveDirectoryName(tt.input)
+			require.Equal(t, tt.expected, result)
```
hemarina, line 322: **[Low]** Missing test for `DeriveDirectoryName("")` empty string input

No test covers empty string input — a common edge case. The function handles it correctly (returns `"new-project"`), but a test would document this behavior explicitly. Consider adding:

```go
{
    name:     "EmptyString",
    input:    "",
    expected: "new-project",
},
```

jongio:
@spboyer what is the resolution for not breaking existing behavior?

spboyer:
@jongio Good question. Existing behavior is preserved through three mechanisms:

**1. `azd init --template <url>` (no positional arg) — new behavior, but opt-out with `.`**
This is the only case where behavior changes: it now auto-creates a project directory (named after the template), matching `git clone` semantics. Users who want the old behavior of initializing in the current directory pass `.`:
```bash
azd init --template <url> .    # uses current directory (old behavior)
azd init --template <url>      # creates subdirectory (new behavior)
```

**2. `azd init` (no template) — unchanged**
Interactive init, `--from-code`, and `--minimal` all continue to use the current directory. The directory creation only triggers when `isTemplateInit` is true AND no positional arg is `.` ([code](https://github.com/Azure/azure-dev/pull/7290/files#diff-init.go-L250-L275)).

**3. Tests covering backward compatibility**
- `DotArgDoesNotCreateDirectory` — verifies `.` uses CWD, no subdirectory created
- `TestInitResolveTargetDirectory` — 8 cases including template-selected-via-tags (CWD) and absolute paths
- `TestInitCreatesProjectDirectory` — verifies subdirectory is created only for template init
- `TestInitNoPromptRequiresMode` — non-template modes unchanged

**4. Cleanup on failure**
If anything goes wrong after directory creation, a deferred cleanup removes the directory and restores the original CWD — so a failed init never leaves artifacts behind (same as `git clone` behavior).

cli/azd/cmd/init.go
```diff
@@ -53,8 +55,14 @@ func newInitFlags(cmd *cobra.Command, global *internal.GlobalCommandOptions) *in
 
 func newInitCmd() *cobra.Command {
 	return &cobra.Command{
-		Use:   "init",
+		Use:   "init [directory]",
```
hemarina: When I typed azd init [dir], it prints out `ERROR: positional [directory] argument requires --template: invalid flag combination`. This is user confusing. We should keep init in help message. 
```suggestion
		Use:   "init",
```

cli/azd/cmd/init.go
```diff
@@ -910,3 +1016,79 @@ type initModeRequiredErrorOptions struct {
...
+
+	if len(names) == 0 {
+		return nil // Empty directory is fine
+	}
+
+	dirName := filepath.Base(targetDir)
+
+	if i.console.IsNoPromptMode() {
+		return fmt.Errorf(
+			"directory '%s' already exists and is not empty; "+
```
hemarina, line 1141: Could we print out the warning before this line and remove the second check? It also asked again for confirmation for overwrite the file. Total will be three times confirmation. 

Image

hemarina:
## Deferred Cleanup Edge Cases — init.go defer block

Reviewed the cleanup defer in the directory-creation path:

`go
defer func() {
    if retErr != nil {
        _ = os.Chdir(originalWd)
        _ = os.RemoveAll(createdProjectDir)
    }
}()
`

### Findings

| Priority | Issue |
|----------|-------|
| :red_circle: High | **Errors silently swallowed** — Both `os.Chdir` and `os.RemoveAll` errors are discarded with `_`. If cleanup fails (e.g., locked files on Windows, original dir deleted), the user gets no diagnostic output. |
| :red_circle: High | **Chdir failure cascades on Windows** — On Windows you cannot delete the CWD. If `os.Chdir(originalWd)` fails, `os.RemoveAll` will also fail because the process is still inside `createdProjectDir`. Both errors are swallowed → orphaned directory, no diagnostics. |
| :yellow_circle: Medium | **Pre-existing empty directories get deleted** — If the user targets an existing empty directory, `MkdirAll` is a no-op. On failure, `RemoveAll` deletes a directory the user already had. The code doesn't track whether it *created* vs *reused* the directory. |

### Suggested Fix

`go
dirExistedBefore := false
if info, err := os.Stat(targetDir); err == nil && info.IsDir() {
    dirExistedBefore = true
}

// ... MkdirAll, Chdir ...

defer func() {
    if retErr != nil {
        if err := os.Chdir(originalWd); err != nil {
            log.Printf("warning: failed to restore working directory: %v", err)
        }
        if !dirExistedBefore {
            if err := os.RemoveAll(createdProjectDir); err != nil {
                log.Printf("warning: failed to clean up directory '%s': %v",
                    filepath.Base(createdProjectDir), err)
            }
        }
    }
}()
`

This addresses: (1) error visibility via logging, (2) not deleting pre-existing directories, and (3) diagnostic output when cleanup fails on Windows.

cli/azd/cmd/init.go
```diff
@@ -178,22 +189,104 @@ func newInitAction(
...
+		if err != nil {
+			return nil, err
+		}
+
+		if targetDir != wd {
+			// Check if target already exists and is non-empty
+			if err := i.validateTargetDirectory(ctx, targetDir); err != nil {
+				return nil, err
+			}
+
```
jongio, line 281: **[High]** Breaking behavioral change for existing automation

Scripts doing `azd init -t foo --no-prompt -e env && azd up` currently get `azure.yaml` in CWD. After this change, `azure.yaml` lands inside a `foo/` subdirectory, but `azd up` still runs in the parent - silent failure.

This is a better UX for interactive users, but the default-on behavior is risky because we don't know what's out there depending on the current in-place behavior. CI pipelines, tutorials, blog posts, and wrapper scripts all assume init puts files in CWD.

Suggestions (pick one):
- Gate behind `--clone-layout` opt-in for v1, then flip the default in a later release after deprecation messaging
- Detect `--no-prompt` and default to CWD (current behavior) unless a positional arg is explicitly passed
- At minimum, treat this as a documented breaking change in release notes and bump a minor version so `azd init -t foo .` becomes the migration path

cli/azd/cmd/init.go
```diff
@@ -910,3 +1016,79 @@ type initModeRequiredErrorOptions struct {
...
+		return filepath.Join(wd, dirArg), nil
+	}
+
+	// No positional arg: auto-derive from template path
+	if i.flags.templatePath != "" {
+		dirName := templates.DeriveDirectoryName(i.flags.templatePath)
+		return filepath.Join(wd, dirName), nil
+	}
+
+	// Template selected via --filter tags (interactive selection) — use CWD
```
jongio: **[Medium]** `--filter` interactive selection doesn't get auto-directory

When `isTemplateInit` is true only because of `--filter` tags (no `--template`), this returns CWD since `templatePath` is empty. Users who pick a template interactively via `--filter` won't get the new git-clone-style behavior - only `--template` users do.

Consider deferring directory resolution until after `initializeTemplate` returns, when the actual template name is known. This would also enable auto-directory for interactively selected templates.

cli/azd/cmd/init.go
```diff
@@ -910,3 +1016,79 @@ type initModeRequiredErrorOptions struct {
...
+// resolveTargetDirectory determines the target directory for template initialization.
+// It returns the current working directory when "." is passed or no template is specified,
+// otherwise it derives or uses the explicit directory name.
+func (i *initAction) resolveTargetDirectory(wd string) (string, error) {
+	if len(i.args) > 0 {
+		dirArg := i.args[0]
+		if dirArg == "." {
+			return wd, nil
+		}
+
```
jongio, line 1064: **[Low]** Accepts absolute paths without validation

A user can pass `azd init -t foo /some/absolute/path` and this will create/chdir into an arbitrary location. If cleanup-on-failure is added later (per hemarina's comment), that could lead to `os.RemoveAll` on a directory outside the user's project root.

Consider rejecting absolute paths here, or validating the target is a descendant of the original CWD.

cli/azd/cmd/init_test.go
```diff
@@ -231,3 +234,195 @@ func TestInitFailFastMissingEnvNonInteractive(t *testing.T) {
...
+
+		wd, err := os.Getwd()
+		require.NoError(t, err)
+
+		expectedDir := filepath.Join(wd, "todo-nodejs-mongo")
+		require.NoDirExists(t, expectedDir)
+
+		// Run will panic or error later due to missing template mocks,
+		// but the directory should be created before that point.
+		_ = runActionSafe(*mockContext.Context, action)
```
jongio, line 510: **[Low]** Missing failure-cleanup test coverage

`runActionSafe` catches panics and returns an error, but the test only asserts `DirExists` - it doesn't verify what happens when a downstream step fails (non-panic). Once cleanup-on-failure is implemented per hemarina's feedback, add a test that verifies the directory is removed when `Run` returns a real error.

cli/azd/cmd/init.go
spboyer: Fixed — changed `Use` from `"init [directory]"` to `"init"` so that `azd init mydir` no longer shows a confusing cobra arg error. Also improved the error message when a positional arg is used without `--template`:

```
ERROR: unexpected argument "mydir": the [directory] option requires --template
Suggestion: Run 'azd init' to initialize interactively, or 'azd init --template <url> [directory]' to create a project from a template.
```

Snapshots regenerated.

cli/azd/cmd/init.go
spboyer: Fixed — replaced the interactive confirmation prompt in `validateTargetDirectory` with a `WarningMessage` UX item. Now only a warning is displayed ("Directory 'X' already exists and is not empty."), and the downstream template initialization handles the actual overwrite confirmation for individual files. This eliminates the redundant prompts.

The `--no-prompt` error path is preserved (still returns an error for non-empty directories in non-interactive mode).

Added `NonEmptyDirectoryShowsWarningInInteractiveMode` test.

cli/azd/cmd/init.go
spboyer: Addressed with option 2: **detect `--no-prompt` and default to CWD** unless a positional arg is explicitly passed.

In `resolveTargetDirectory`, when `console.IsNoPromptMode()` is true and no positional arg is given, we return CWD — preserving the existing behavior that scripts depend on. The auto-directory creation only activates in interactive mode.

Migration path for automation that _wants_ a subdirectory:
```bash
azd init -t foo my-project --no-prompt  # explicit positional arg opts in
```

Existing scripts continue to work unchanged:
```bash
azd init -t foo --no-prompt -e env && azd up  # azure.yaml stays in CWD
```

Added `NoPromptNoArgDefaultsToCwd` (unit) and `NoPromptTemplateInitUsesCwd` (integration) tests.

cli/azd/cmd/init.go
spboyer: Acknowledged — added a TODO comment:

```go
// TODO(#7290): Derive directory name from the selected template after interactive
// selection completes, so --filter users also get git-clone-style behavior.
```

This requires deferring directory creation until after `initializeTemplate` returns, which is a meaningful refactor of the init flow. Tracking as a follow-up to keep this PR focused on the core directory-creation feature.

cli/azd/cmd/init.go
spboyer: Fixed — `resolveTargetDirectory` now rejects absolute paths with a clear `ErrorWithSuggestion`:

```
ERROR: absolute path "/some/absolute/path" is not allowed as a directory argument
Suggestion: Use a relative directory name (e.g., my-project) or . for the current directory.
```

This prevents creating directories outside the working tree and eliminates the risk of `os.RemoveAll` on an arbitrary path during cleanup-on-failure.

Added `AbsolutePathIsRejected` test.

cli/azd/cmd/init_test.go
spboyer: Added `FailedInitCleansUpDirectory` test. It mocks git as not installed so `tools.EnsureInstalled` returns a real error (not a panic), which triggers the cleanup-on-failure defer. The test verifies:

1. The auto-created directory is removed after failure
2. CWD is restored to the original working directory

This exercises the deferred cleanup path that was added per hemarina earlier feedback.

cli/azd/pkg/azdext/process_darwin.go
```diff
@@ -66,7 +66,7 @@ func findProcessByNameOS(name string) []ProcessInfo {
 	nameLower := strings.ToLower(name)
 	var results []ProcessInfo
 
-	for _, line := range strings.Split(string(output), "\n") {
+	for line := range strings.SplitSeq(string(output), "\n") {
```
jongio, line 69: test

cli/azd/cmd/init.go
```diff
@@ -178,22 +189,105 @@ func newInitAction(
...
+	}
+
+	// When a template is specified, auto-create a project directory (like git clone).
+	// The user can pass a positional [directory] argument to override the folder name,
+	// or pass "." to use the current directory (preserving existing behavior).
+	createdProjectDir := ""
+	originalWd := wd
+
+	if isTemplateInit {
+		targetDir, err := i.resolveTargetDirectory(wd)
```
jongio, line 258: line-test-253

cli/azd/cmd/init.go
```diff
@@ -178,22 +189,105 @@ func newInitAction(
...
+			wd = targetDir
+			createdProjectDir = targetDir
+
+			// Clean up the created directory and restore the original CWD
+			// if any downstream step fails, matching git clone's behavior.
+			defer func() {
+				if retErr != nil {
+					_ = os.Chdir(originalWd)
+					_ = os.RemoveAll(createdProjectDir)
+				}
```
jongio, line 312: line-test-283

cli/azd/cmd/init.go
```diff
@@ -178,22 +189,105 @@ func newInitAction(
...
+			}()
+		}
+	}
+
+	azdCtx := azdcontext.NewAzdContextWithDirectory(wd)
```
jongio, line 317: test-288

jongio (commented):
line-test

cli/azd/cmd/init.go
```diff
@@ -178,22 +189,105 @@ func newInitAction(
...
+	i.lazyAzdCtx.SetValue(azdCtx)
```
jongio, line 318: test-289

cli/azd/cmd/init.go
```diff
@@ -178,22 +189,105 @@ func newInitAction(
...
+
 	// ensure that git is available
 	if err := tools.EnsureInstalled(ctx, []tools.ExternalTool{i.gitCli}...); err != nil {
```
jongio, line 321: test-292

cli/azd/cmd/init.go
jongio, line 312: This defer can delete a pre-existing directory the user pointed at. `createdProjectDir` gets set even when `MkdirAll` was a no-op (directory already existed). If init fails downstream, `RemoveAll` wipes it.

hemarina's Apr 11 suggestion covers this well - track whether the directory existed before creation and only delete if we actually created it. Also log the errors instead of discarding them with `_` - on Windows, if `Chdir(originalWd)` fails, `RemoveAll` also fails (can't delete CWD), and neither failure is visible.

Suggested approach: check `os.Stat(targetDir)` before `MkdirAll` to set a `dirExistedBefore` flag, then only `RemoveAll` when `!dirExistedBefore`. Log both `Chdir` and `RemoveAll` errors with `log.Printf`.

cli/azd/cmd/init.go
jongio, line 258: `azd init -t ./my-template` hits a self-targeting bug. The template path resolves to absolute at line 240 (e.g., `/cwd/my-template`), then `DeriveDirectoryName` extracts `my-template`, which joins back to the same path as the template source. The `targetDir != wd` guard doesn't catch this because the absolute path differs from `wd`.

On failure, the defer wipes the template source directory.

Consider comparing `targetDir` against the resolved template path and falling back to CWD (or erroring) when they overlap.

cli/azd/cmd/init.go
jongio, line 318: `.env` loading via `godotenv.Overload()` (line 304) now happens after the `chdir` into the target directory. For new empty directories this is fine, but if the user targets an existing directory with a `.env` file, it loads that instead of the invocation directory's `.env`. Worth a comment noting this is intentional (project dir's `.env` should take effect).

cli/azd/pkg/azdext/process_darwin.go
jongio, line 69: This change is unrelated to the init directory feature. Consider moving it to a separate PR to keep the changeset focused.

spboyer:
## Backward Compatibility — How We Avoid Breaking Existing Users

The auto-create-directory behavior **only activates when all three conditions are true**:

1. `--no-prompt` is **not** set
2. Running in a **real TTY terminal** (`IsSpinnerInteractive()`)
3. **Not** invoked by an AI agent (`agentdetect.IsRunningInAgent()`)

If any condition fails, `azd init -t <template>` initializes in the **current working directory** — identical to today's behavior.

| Scenario | Auto-create dir? | Why |
|---|---|---|
| `azd init -t repo` in terminal | ✅ Yes | Interactive human at a TTY |
| `azd init -t repo --no-prompt` | ❌ No | `--no-prompt` → CWD |
| CI script (piped stdin, no TTY) | ❌ No | Non-TTY → CWD |
| LLM / AI agent calling azd | ❌ No | Agent detected → CWD |
| `azd init -t repo .` | ❌ No | Explicit `.` → CWD |
| `azd init -t repo my-dir` | ✅ Yes (named) | Explicit arg works everywhere |

### Additional safety
- **Path traversal blocked**: `azd init -t repo ../../etc/evil` is rejected — prevents `MkdirAll`/`RemoveAll` outside the working tree
- **Cleanup on failure**: If init fails after creating the directory, the directory is removed and CWD is restored (matching `git clone` behavior)
- **Agent detection is injectable**: Uses `isRunningInAgent` function field so tests aren't affected by the test runner's own environment (e.g., running tests under Copilot CLI)

cli/azd/cmd/init.go
spboyer: Fixed in 0e4c161c0 — now tracking `dirExistedBefore` via `os.Stat` before `MkdirAll`. The cleanup defer only calls `RemoveAll` when we actually created the directory. Added `FailedInitPreservesPreExistingDirectory` test to verify.

cli/azd/cmd/init.go
spboyer: Fixed in 0e4c161c0 — added a self-targeting check in `Run()` after `resolveTargetDirectory`. When a local template path (e.g. `./my-template`) resolves to the same absolute path as the derived target directory, we fall back to CWD to avoid the source/target overlap. Added `LocalTemplateSelfTargetFallsBackToCwd` test.

cli/azd/cmd/init.go
spboyer: Added a comment in 0e4c161c0 explaining this is intentional — the project directory's `.env` should take precedence over the invocation directory's, since after chdir we're operating in the project context.

cli/azd/pkg/azdext/process_darwin.go
spboyer: Agreed — this `SplitSeq` change was from an earlier commit on the branch (likely `go fix` modernization). Will move it to a separate PR to keep the changeset focused.

spboyer:
All review feedback addressed in 0e4c161c0. Summary of fixes:

- **Pre-existing dir safety**: Now tracks `dirExistedBefore` — cleanup defer only `RemoveAll`s directories we actually created
- **Self-targeting bug**: Detects when `./my-template` resolves to the same path as the derived target dir; falls back to CWD
- **`.env` comment**: Added intent comment explaining post-chdir `.env` loading is by design
- **`process_darwin.go`**: Acknowledged as unrelated — will move to separate PR

2 new tests added: `FailedInitPreservesPreExistingDirectory`, `LocalTemplateSelfTargetFallsBackToCwd`

@jongio — ready for another look when you get a chance. Thanks for the thorough review! 🙏

cli/azd/test/functional/init_test.go
```diff
@@ -280,33 +280,39 @@ func Test_CLI_Init_CanUseTemplate(t *testing.T) {
 	)
 	require.NoError(t, err)
 
+	// Template init now creates a subdirectory derived from the template name.
+	projectDir := filepath.Join(dir, "cosmos-dotnet-core-todo-app")
```
jongio: The auto-directory feature only activates when `IsSpinnerInteractive() == true` and `IsNoPromptMode() == false` (init.go:1096). Functional tests run as subprocesses with piped stdin (non-TTY), so `resolveTargetDirectory` returns CWD and no subdirectory is created.

This test now expects `filepath.Join(dir, "cosmos-dotnet-core-todo-app")` but files will land in `dir` directly. Same issue in `Test_CLI_Init_WithCwdAutoCreate` (line 345) which uses `--no-prompt`.

This is likely why ADO CI builds are failing. Fix: either revert these assertions to expect files in the original directory, or pass an explicit positional `[directory]` argument (which bypasses the TTY check) to test the auto-create path.

cli/azd/cmd/init_test.go
```diff
@@ -231,3 +235,454 @@ func TestInitFailFastMissingEnvNonInteractive(t *testing.T) {
...
+
+		result, err := action.resolveTargetDirectory(wd)
+		require.NoError(t, err)
+		require.Equal(t, filepath.Join(wd, "todo-nodejs-mongo"), result)
+	})
+
+	t.Run("AbsolutePathIsRejected", func(t *testing.T) {
+		mockContext := mocks.NewMockContext(context.Background())
+		flags := &initFlags{
+			templatePath: "owner/repo",
```
jongio, line 344: `"/some/absolute/path"` isn't absolute on Windows (`filepath.IsAbs` requires a drive letter like `C:\`). This test passes on Linux/Mac but fails on Windows. Consider using `filepath.Join(os.TempDir(), "test")` or a platform-conditional path.

hemarina (commented):
~This comment has been retracted.~

cli/azd/test/functional/init_test.go
spboyer: Good catch — fixed in 431c61f91. Functional tests run as non-TTY subprocesses, so `IsSpinnerInteractive()` returns false and auto-directory never activates. Updated both `Test_CLI_Init_CanUseTemplate` (uses `dir` directly) and `Test_CLI_Init_WithCwdAutoCreate` (expects files in `-C` target, not a template subdirectory).

cli/azd/cmd/init_test.go
spboyer: Fixed in 431c61f91 — now uses `filepath.Join(os.TempDir(), "some", "absolute", "path")` which produces a proper drive-letter absolute path on Windows (e.g. `C:\Users\...`).

spboyer:
Addressed both new review comments in 431c61f91:

- **Functional tests non-TTY**: `Test_CLI_Init_CanUseTemplate` and `Test_CLI_Init_WithCwdAutoCreate` now correctly expect files in CWD/target dir (not a template subdirectory), since functional tests run as non-TTY subprocesses where auto-directory doesn't activate.
- **Windows absolute path**: `AbsolutePathIsRejected` test now uses `filepath.Join(os.TempDir(), ...)` for a cross-platform absolute path.

All 8 preflight checks pass ✓. @jongio — ready for another look!

spboyer:
Re: Windows CI failure — this is **pre-existing on the branch**, not from our init changes. Confirmed by checking that `BuildCLI Windows` was already failing on commit `fcfb20d5d` (before any of our fixes).

The failure is in the ADO pipeline `azure-dev - cli (BuildAndTest BuildCLI Windows)`. All GitHub Actions checks (lint, copyright, go-fix) pass. Linux and Mac builds pass.

The Windows-specific files changed on this branch (`cache_windows.go`, `console_windows.go`) are from earlier PRs merged into the branch, not from the init auto-create directory feature.

May need to rebase onto latest main to pick up any Windows fixes that have landed since.

spboyer:
Rebased onto latest main (441421bfb) — this picks up recent fixes including the PSModulePath fix (#7703). CI re-triggered. 🤞

wbreza (commented):
## Supplementary Code Review — azd init auto-create directory

Supplementing existing reviews from jongio and hemarina (both approved). Focused on anything prior reviews may have missed.

### 🟡 Medium

**1. Windows reserved device names not sanitized in DeriveDirectoryName**
📁 pkg/templates/path.go

Template repos named CON, PRN, AUX, NUL, COM1–COM9, or LPT1–LPT9 produce a derived directory name that's a Windows reserved device name. os.MkdirAll on Windows may silently create an inaccessible entry or fail with a confusing error. git clone has explicit handling for these. Consider adding a check or documenting as a known limitation.

**2. Use: "init" missing [directory] positional arg**
📁 cmd/init.go

The command declares Args: cobra.MaximumNArgs(1) but Use: "init" doesn't mention the positional argument. zd init --help won't show that [directory] is accepted. Should be Use: "init [directory]".

**3. Swallowed ilepath.Abs error defeats early-resolve purpose**
📁 cmd/init.go

The early-resolve block calls ilepath.Abs(i.flags.templatePath) and silently falls through on error (if err == nil). If the call fails, the code continues with the relative path, and after the subsequent os.Chdir(targetDir), the relative path resolves against the new directory — exactly the bug this block was added to prevent. The error should be returned, not swallowed.

### 🔵 Low

**4.** ~30 new test instances use context.Background() instead of 	.Context() (per AGENTS.md modern Go patterns)
**5.** rrors.As used instead of rrors.AsType in TestInitNoPromptRequiresMode (per repo coding standards)
**6.** Redundant second LooksLikeLocalPath + ilepath.Abs on an already-absolute path in the self-targeting guard — simplify to ilepath.Clean comparison

### Overall

Solid feature with thorough review history. The prior reviewers covered the major concerns well. #3 is the most actionable — latent correctness bug. Nice work addressing the extensive feedback. 👍

PR #7296, opened 2026-03-24
feat: avoid prompting for layer outputs in `pipeline config`

### LINKED ISSUE #7182: feat: resolve dependencies between provisioning layers before prompting
When using layered provisioning, `azd pipeline config` should understand dependencies between layers and automatically detect outputs from earlier layers when prompting for later ones.

For example, given the following setup:
 ```yaml
infra:
   layers:
     - name: base
       path: ./infra/base # outputs CONTAINER_APP_ENVIRONMENT_ID
     - name: identity
       path: ./infra/identity # takes a dependency on CONTAINER_APP_ENVIRONMENT_ID
 ```

 azd should not prompt for the parameter associated to `CONTAINER_APP_ENVIRONMENT_ID` when the variable hasn't yet been set.

weikanglim:
## Summary

Add support for `pipeline config` to detect outputs between layers and avoid prompting for them.
 
## Why

In a normal provision run, these outputs are not user-provided, and would be automatically set by the provisioning engine.

## What's changing

- In `cmd/pipeline.go`, we collect outputs from each previous layer, store them as a virtual env-mapped values.
- In the provider implementation, we treat parameters mapped to virtual values as resolved without prompting.

A new `PlannedOutputs()` function is added to the infra provider interface that represents plan-time outputs (no resolved values).

## Testing

- Add tests for virtual env substitution and required-parameter prompt skipping paths

Fixes #7182

### DIFF
--- a/cli/azd/cmd/pipeline.go
+++ b/cli/azd/cmd/pipeline.go
@@ -186,19 +186,68 @@ func (p *pipelineConfigAction) Run(ctx context.Context) (*actions.ActionResult, 
 	layers := infra.Options.GetLayers()
 	allParameters := []provisioning.Parameter{}
 
-	for _, layer := range layers {
+	inputParameters := func(layer provisioning.Options) ([]provisioning.Parameter, error) {
 		err = p.provisioningManager.Initialize(ctx, p.projectConfig.Path, layer)
 		if err != nil {
-			return nil, err
+			return nil, fmt.Errorf(
+				"failed to initialize infra provider %s: %w",
+				layer.Provider,
+				err,
+			)
 		}
 
-		// Pull provider specific parameters
-		providerParameters, err := p.provisioningManager.Parameters(ctx)
+		parameters, err := p.provisioningManager.Parameters(ctx)
 		if err != nil {
-			return nil, fmt.Errorf("failed to get parameters for provider %s: %w", pipelineProviderName, err)
+			return nil, fmt.Errorf(
+				"failed to get parameters for infra provider %s: %w",
+				layer.Provider,
+				err,
+			)
 		}
 
-		allParameters = append(allParameters, providerParameters...)
+		return parameters, nil
+	}
+
+	if len(layers) <= 1 {
+		for _, layer := range layers {
+			parameters, err := inputParameters(layer)
+			if err != nil {
+				return nil, err
+			}
+
+			allParameters = append(allParameters, parameters...)
+		}
+	} else {
+		// virtualEnv contains all accumulated outputs from previous layers.
+		virtualEnv := map[string]string{}
+
+		for _, layer := range layers {
+			layer.VirtualEnv = virtualEnv
+
+			parameters, err := inputParameters(layer)
+			if err != nil {
+				return nil, fmt.Errorf("layer '%s': %w", layer.Name, err)
+			}
+
+			allParameters = append(allParameters, parameters...)
+
+			outputs, err := p.provisioningManager.PlannedOutputs(ctx)
+			if err != nil {
+				return nil, fmt.Errorf(
+					"layer '%s': failed to get planned outputs for infra provider %s: %w",
+					layer.Name,
+					layer.Provider,
+					err,
+				)
+			}
+
+			// Save current outputs
+			for _, output := range outputs {
+				// Use a sentinel value that records the source layer and output name so
+				// downstream virtual env substitutions remain traceable during planning.
+				virtualEnv[output.Name] = fmt.Sprintf("%s--%s", layer.Name, output.Name)
+			}
+		}
 	}
 
 	p.manager.SetParameters(allParameters)
--- a/cli/azd/pkg/infra/provisioning/bicep/bicep_provider.go
+++ b/cli/azd/pkg/infra/provisioning/bicep/bicep_provider.go
@@ -1725,16 +1725,25 @@ type loadParametersResult struct {
 	locationParams []string
 	// envMapping is a map of parameter name to environment variable names
 	// holds information about which parameters are mapped to which env vars for
-	// cases like "param": "${env:AZURE_FOO}-${env:AZURE_BAR}", envMapping will
+	// cases like "param": "${AZURE_FOO}-${AZURE_BAR}", envMapping will
 	// contain {"param": ["AZURE_FOO", "AZURE_BAR"]}
+	//
 	// This information is useful for setting a CI/CD automatically. Each env var
 	// will be set to the value of the parameter as variable or secret.
 	envMapping map[string][]string
+
+	// parameters that should be considered virtual, i.e. their values are only
+	// known at runtime.
+	//
+	// These parameters should not be prompted and treated as resolved values,
+	// with other potential unmapped environment variables stored in envMapping.
+	virtualMapping map[string]struct{}
 }
 
 // envSubstResult contains the results of environment variable substitution
 type envSubstResult struct {
 	hasUnsetEnvVar                  bool
+	hasVirtualEnvVar                bool
 	mappedEnvVars                   []string
 	parametersMappedToAzureLocation []string
 }
@@ -1747,6 +1756,7 @@ func evalParamEnvSubst(
 	principalType string,
 	paramName string,
 	env *environment.Environment,
+	virtualEnv map[string]string,
 ) (string, envSubstResult, error) {
 	result := envSubstResult{}
 
@@ -1760,13 +1770,23 @@ func evalParamEnvSubst(
 		if name == environment.LocationEnvVarName {
 			result.parametersMappedToAzureLocation = append(result.parametersMappedToAzureLocation, paramName)
 		}
-		// principalId and locations are intentionally excluded from the mapped env vars as
-		// they are global env vars
+
+		if virtualEnv != nil {
+			if value, has := virtualEnv[name]; has {
+				result.hasVirtualEnvVar = true
+				return value
+			}
+		}
+
+		// principalId and principalType are intentionally excluded from the mapped env vars
+		// as they are global env vars.
 		result.mappedEnvVars = append(result.mappedEnvVars, name)
-		if _, isDefined := env.LookupEnv(name); !isDefined {
+
+		value, isDefined := env.LookupEnv(name)
+		if !isDefined {
 			result.hasUnsetEnvVar = true
 		}
-		return env.Getenv(name)
+		return value
 	})
 	return replaced, result, err
 }
@@ -1820,6 +1840,7 @@ func (p *BicepProvider) loadParameters(ctx context.Context, template *azure.ArmT
 	parametersMappedToAzureLocation := []string{}
 	resolvedParams := map[string]azure.ArmParameter{}
 	envMapping := map[string][]string{}
+	virtualMapping := map[string]struct{}{}
 
 	// resolving each parameter to keep track of the name during the resolution.
 	// We used to resolve all the file before, supporting env var substitution at any part of the file.
@@ -1847,6 +1868,7 @@ func (p *BicepProvider) loadParameters(ctx context.Context, template *azure.ArmT
 					string(principalType),
 					paramName,
 					p.env,
+					p.options.VirtualEnv,
 				)
 				if err != nil {
 					return loadParametersResult{}, err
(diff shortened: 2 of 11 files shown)


cli/azd/pkg/infra/provisioning/bicep/bicep_provider.go
```diff
@@ -2791,3 +2829,19 @@ func (p *BicepProvider) Parameters(ctx context.Context) ([]provisioning.Paramete
...
+	if err != nil {
+		return nil, fmt.Errorf("creating template: %w", err)
+	}
+
+	var outputs []provisioning.PlannedOutput
+	for key := range compileResult.Template.Outputs {
+		outputs = append(outputs, provisioning.PlannedOutput{
+			Name: key,
+		})
+	}
```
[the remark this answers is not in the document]
weikanglim: Addressed by filtering `BicepProvider.PlannedOutputs()` to skip secure outputs via `azure.IsSecuredARMType(output.Type)`. That keeps `pipeline config` from treating non-persistable outputs as virtual env values. I also added `TestPlannedOutputsSkipsSecureOutputs` to cover this path.

cli/azd/cmd/pipeline.go
```diff
@@ -177,19 +177,37 @@ func (p *pipelineConfigAction) Run(ctx context.Context) (*actions.ActionResult,
...
 		if err != nil {
 			return nil, fmt.Errorf("failed to get parameters for provider %s: %w", pipelineProviderName, err)
 		}
 
 		allParameters = append(allParameters, providerParameters...)
+
+		outputs, err := p.provisioningManager.PlannedOutputs(ctx)
+		if err != nil {
+			return nil, fmt.Errorf("failed to get outputs for provider %s: %w", pipelineProviderName, err)
+		}
```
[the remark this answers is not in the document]
weikanglim: Addressed by splitting `pipeline config` into separate single-layer and multi-layer paths. The single-layer flow now only initializes the provider and gathers parameters; `VirtualEnv` accumulation and `PlannedOutputs()` are only used for multi-layer plans.

cli/azd/pkg/infra/provisioning/bicep/bicep_provider.go
```diff
@@ -1759,10 +1767,19 @@ func evalParamEnvSubst(
 		// principalId and locations are intentionally excluded from the mapped env vars as
 		// they are global env vars
 		result.mappedEnvVars = append(result.mappedEnvVars, name)
```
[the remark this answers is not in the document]
weikanglim: Updated the comment to match the implementation. Only `principalId` and `principalType` are intentionally excluded from `mappedEnvVars`; `AZURE_LOCATION` is still tracked there while also being recorded in `parametersMappedToAzureLocation`.

cli/azd/cmd/pipeline.go
```diff
@@ -177,19 +177,37 @@ func (p *pipelineConfigAction) Run(ctx context.Context) (*actions.ActionResult,
...
 		providerParameters, err := p.provisioningManager.Parameters(ctx)
 		if err != nil {
 			return nil, fmt.Errorf("failed to get parameters for provider %s: %w", pipelineProviderName, err)
 		}
 
 		allParameters = append(allParameters, providerParameters...)
+
+		outputs, err := p.provisioningManager.PlannedOutputs(ctx)
+		if err != nil {
+			return nil, fmt.Errorf("failed to get outputs for provider %s: %w", pipelineProviderName, err)
```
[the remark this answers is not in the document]
weikanglim: Addressed by moving the error context to the infra layer/provider. The init, parameter, and planned-output errors now include both `layer.Name` and `layer.Provider`, so failures point at the provisioning layer instead of the CI provider.

cli/azd/cmd/pipeline.go
```diff
@@ -177,19 +177,69 @@ func (p *pipelineConfigAction) Run(ctx context.Context) (*actions.ActionResult,
...
+	} else {
+		// virtualEnv contains all accumulated outputs from previous layers.
+		virtualEnv := map[string]string{}
+
+		for _, layer := range layers {
+			layer.VirtualEnv = virtualEnv
+
+			parameters, err := inputParameters(layer)
+			if err != nil {
+				return nil, fmt.Errorf("layer '%s': %w", layer.Name, err)
```
jongio, line 229: **[MEDIUM]** Logic: Double-wrapped layer name in error message

`inputParameters()` already wraps its errors with the layer name (e.g., `layer "foo": failed to initialize...`). This outer `fmt.Errorf` wraps again, producing:

`layer 'foo': layer "foo": failed to initialize infra provider bicep: <err>`

The single-layer path at line 208 returns the error directly. This should do the same.

```suggestion
return nil, err
```

cli/azd/cmd/pipeline.go
```diff
@@ -177,19 +177,69 @@ func (p *pipelineConfigAction) Run(ctx context.Context) (*actions.ActionResult,
...
+					"layer %q: failed to get planned outputs for infra provider %s: %w",
+					layer.Name,
+					layer.Provider,
+					err,
+				)
+			}
+
+			// Save current outputs
+			for _, output := range outputs {
+				// save a dummy value that is easily looked at
```
jongio: **[LOW]** Quality: This comment doesn't explain why the value uses the `layerName--outputName` format.

```suggestion
// Use a sentinel value encoding the source layer, so downstream parameter
// resolution can trace which layer provides each virtual env var.
```

cli/azd/pkg/infra/provisioning/bicep/bicep_provider.go
```diff
@@ -1852,6 +1871,12 @@ func (p *BicepProvider) loadParameters(ctx context.Context, template *azure.ArmT
 				parametersMappedToAzureLocation = append(
 					parametersMappedToAzureLocation, substResult.parametersMappedToAzureLocation...)
 
+				if substResult.hasVirtualEnvVar {
```
jongio, line 1881: **[MEDIUM]** Logic: Mixed virtual + real env var references in one parameter

If a parameter value references both a virtual env var and a real one (e.g., `"${PREV_LAYER_OUTPUT}/${REAL_SECRET}"`), `hasVirtualEnvVar` is set and this `continue` skips the parameter entirely. The real env var IS recorded in `envMapping` above, but `ensureParameters` also skips virtual parameters - so the real portion won't be prompted or configured.

Is it safe to assume that a parameter referencing a virtual env var will never also reference a real one? If so, worth a comment here. If not, this needs a different strategy for mixed references (e.g., only skip prompting but still track the real env vars for CI/CD config).

cli/azd/cmd/pipeline.go
weikanglim: Addressed by keeping the layer context in one place. The helper now reports provider-specific failures, while the multi-layer branch adds the `layer.Name` prefix exactly once, so the final error no longer double-wraps the layer.

cli/azd/cmd/pipeline.go
weikanglim: Updated the comment to explain the sentinel format directly. It now calls out that `layerName--outputName` is a traceable planning-time placeholder used for downstream virtual env substitution.

cli/azd/pkg/infra/provisioning/bicep/bicep_provider.go
weikanglim: Addressed by treating virtual refs separately from real env mappings. Virtual keys are no longer added to `mappedEnvVars`; virtual parameters are still surfaced from `Parameters()` with any remaining real `EnvVarMapping`, and pipeline config now falls back to reading mapped env vars when `Value == nil`. I also added regression coverage in `bicep_provider_test.go` and `pipeline_test.go` for the mixed virtual+real case.

cli/azd/pkg/pipeline/pipeline.go
```diff
@@ -169,17 +169,18 @@ func mergeProjectVariablesAndSecrets(
...
+				fmt.Errorf(
+					"parameter %s got its value from a local prompt and it has more than one mapped environment "+
+						"variable. "+
+						"The value can't be configured in CI mapped to multiple ENV VARS if azd prompt for its value. "+
+						"Define a single mapping for %s to one ENV VAR as part of the infra parameters definition",
+					parameter.Name, parameter.Name)
+		}
+
+		if envVarsCount > 1 || parameter.Value == nil {
 			// for parameters mapped to more than one ENV VAR, each env var becomes either a variable or a secret
```
jongio: **[MEDIUM]** Quality: Stale comment after condition broadened

This condition was expanded from `envVarsCount > 1` to `envVarsCount > 1 || parameter.Value == nil` for virtual parameters, but the comment still only describes the multi-env-var case. A virtual parameter with a single env var mapping (`envVarsCount == 1 && Value == nil`) also enters this branch now.

```suggestion
// For parameters mapped to more than one env var, or virtual parameters
// without a resolved value, each mapped env var becomes a variable or a secret.
```

jongio (commented):
## PR Review - Azure/azure-dev#7296

**feat: avoid prompting for layer outputs in `pipeline config`** by @weikanglim

### Summary

What: Adds a virtual environment system so `pipeline config` can detect outputs between infrastructure layers and skip prompting for parameters that will be automatically populated during provisioning.

Why: In multi-layer infra setups, outputs from earlier layers (e.g., storage account names) are inputs to later layers. Without this change, `pipeline config` would prompt users for values that are never user-provided.

Assessment: Clean design. The single-layer vs multi-layer split keeps the common case simple. The `PlannedOutput` API is intentionally minimal - good restraint. All previous review feedback (7 findings across Copilot bot and jongio reviews) has been thoroughly addressed with code fixes and regression tests.

### Findings

| Quality | 0 | 0 | 1 | 0 |
| **Total** | **0** | **0** | **1** | **0** |

### Key Findings

1. **[MEDIUM] Quality**: Stale inline comment on broadened `if` condition in pipeline.go - comment describes only the multi-env-var case but the condition now also handles virtual parameters with nil values.

### Test Coverage Estimate

- Well covered: virtual env substitution (simple + mixed), real env mapping preservation for mixed refs, ensureParameters skipping virtual params, PlannedOutputs filtering secure outputs, pipeline parameter merging with nil Value
- Likely uncovered: empty virtualEnv map (non-nil but empty), full multi-layer pipeline integration flow, PlannedOutputs with zero outputs, duplicate output names across layers

### What's Done Well

- Clean separation between single-layer and multi-layer paths keeps the common case simple
- `PlannedOutput` type is intentionally minimal - good API design restraint
- Virtual env checked before real env in the substitution callback - correct priority
- Mixed virtual+real env refs properly preserve real env var mappings for CI/CD configuration
- All previous review feedback addressed with code fixes AND regression tests
- Test uses `require.ElementsMatch` for order-independent output comparison

---
*1 inline comment below.*

cli/azd/pkg/pipeline/pipeline.go
weikanglim: Updated the comment.

cli/azd/pkg/infra/provisioning/bicep/bicep_provider_test.go
```diff
@@ -1910,3 +1928,203 @@ func TestHelperEvalParamEnvSubst(t *testing.T) {
...
+			"mixedValue":      {Type: "string"},
+		},
+		Outputs: azure.ArmTemplateOutputs{},
+	}
+
+	infraProvider := createBicepProviderWithEnvAndMode(
+		t,
+		mockContext,
+		armTemplate,
+		map[string]string{"REAL_SECRET": "secret-value"},
```
jongio: **[MEDIUM]** Tests: This test sets `REAL_SECRET` in the env, but there's no variant where the real env var is unset. That path sets both `hasVirtualEnvVar` and `hasUnsetEnvVar` in `evalParamEnvSubst`, and the unset env var name ends up in `envMapping` while `mergeProjectVariablesAndSecrets` silently skips it. Worth a test to document that edge case.

jongio (approved):
All previous findings addressed. No new issues found in the incremental diff (8a16d0bd -> 30f8c5b1).

Key fixes verified:
- Error wrapping restructured to avoid double layer-name context
- Sentinel format comment updated
- Virtual vs real env var refs handled separately in `evalParamEnvSubst`
- `pipeline.go` comment updated for broadened condition
- "unset real env var" test variant added

The `Value == nil` condition broadening in `mergeProjectVariablesAndSecrets` is a correctness fix - without it, virtual params with a single env var mapping would produce `""<nil>"""` as the CI variable value.

Issue #7265, opened 2026-03-24
[azd update] apply code-signing verification


hemarina:
## Summary

`azd update` applies code-signing verification only on the direct binary download path (`updateViaBinaryDownload`), but skips it for the Windows MSI/install-script path and all Linux paths.

## Affected paths

| Update path | Signature verified? | Details |
|---|---|---|
| Direct binary download (macOS) | Yes | `codesign -v --strict` |
| Direct binary download (Windows) | Yes | `Get-AuthenticodeSignature` |
| Direct binary download (Linux) | No | Skipped — no standard mechanism |
| **Windows MSI via `install-azd.ps1`** | **No** | **`-SkipVerify` is passed to the install script** |
| Linux/macOS via `install-azd.sh` | Yes | No verification in updater; relies on script; script checks for macOS |
| Linux/macOS via `install-azd.sh` | No | No verification in updater; relies on script; script skipped — no standard mechanism |
| Homebrew cask | N/A | Delegated to `brew` |
| winget / choco | N/A | Delegated to package manager |

## Why this matters

The Windows MSI path (`updateViaMSI`) is the primary update method for most Windows users, yet it explicitly bypasses Authenticode verification by passing `-SkipVerify` to `install-azd.ps1`. HTTPS transport alone is weaker than verifying the publisher signature on the downloaded artifact.

PR #7298, opened 2026-03-25
[azd update] apply code sign verification for windows msi isntall

hemarina:
#7265 
This pull request refactors how the Windows MSI installation script arguments are constructed and tested, improving clarity and correctness for different release channels (stable vs. daily). The main change is to simplify and clarify the logic for building PowerShell command arguments, ensuring that stable and daily channels are handled distinctly and tested accordingly.

Refactoring and logic improvements for install script arguments:

* Replaced the `versionFlag` function and refactored `buildInstallScriptArgs` to directly construct the appropriate PowerShell command for each channel, ensuring that stable uses a piped script and daily downloads and executes the script with additional parameters.

Test updates for new logic and coverage:

* Removed the now-obsolete `TestVersionFlag` and updated `TestBuildInstallScriptArgs` to verify the presence or absence of key substrings in the generated arguments, matching the new logic for stable and daily channels. [[1]](diffhunk://#diff-54b83ff058baaa3f7634c7b08148a863ab682dd97d7931d60dc88a8278b7596eL45-R54) [[2]](diffhunk://#diff-54b83ff058baaa3f7634c7b08148a863ab682dd97d7931d60dc88a8278b7596eL79-R69) [[3]](diffhunk://#diff-54b83ff058baaa3f7634c7b08148a863ab682dd97d7931d60dc88a8278b7596eL92-R83)
* Enhanced `TestBuildInstallScriptArgs` and `TestBuildInstallScriptArgs_Structure` to check for correct argument construction, including validation that stable does not use temporary files or extra parameters, while daily does.

### DIFF
--- a/cli/azd/pkg/update/manager.go
+++ b/cli/azd/pkg/update/manager.go
@@ -507,10 +507,10 @@ func (m *Manager) updateViaMSI(ctx context.Context, cfg *UpdateConfig, writer io
 			fmt.Errorf("failed to hash safety copy before install: %w", hashErr))
 	}
 
-	log.Printf("Running install script: powershell %s", strings.Join(psArgs, " "))
+	log.Printf("Running install script: pwsh %s", strings.Join(psArgs, " "))
 	fmt.Fprintf(writer, "Installing azd %s channel...\n", cfg.Channel)
 
-	runArgs := exec.NewRunArgs("powershell", psArgs...).
+	runArgs := exec.NewRunArgs("pwsh", psArgs...).
 		WithStdOut(writer).
 		WithStdErr(writer)
 
--- a/cli/azd/pkg/update/manager.go
+++ b/cli/azd/pkg/update/manager.go
@@ -507,10 +507,10 @@ func (m *Manager) updateViaMSI(ctx context.Context, cfg *UpdateConfig, writer io
 			fmt.Errorf("failed to hash safety copy before install: %w", hashErr))
 	}
 
-	log.Printf("Running install script: pwsh %s", strings.Join(psArgs, " "))
+	log.Printf("Running install script: powershell %s", strings.Join(psArgs, " "))
 	fmt.Fprintf(writer, "Installing azd %s channel...\n", cfg.Channel)
 
-	runArgs := exec.NewRunArgs("pwsh", psArgs...).
+	runArgs := exec.NewRunArgs("powershell", psArgs...).
 		WithStdOut(writer).
 		WithStdErr(writer)
 
--- a/cli/azd/pkg/update/msi_windows.go
+++ b/cli/azd/pkg/update/msi_windows.go
@@ -150,35 +150,32 @@ func isStandardMSIInstall() error {
 	return nil
 }
 
+func escapeForPSSingleQuote(s string) string {
+	return strings.ReplaceAll(s, "'", "''")
+}
+
 // buildInstallScriptArgs constructs the PowerShell arguments to run install-azd.ps1.
-//
-// For the stable channel the script is piped directly through Invoke-Expression.
-// No named parameters are needed because the default MSI install location is correct.
-//
-// For other channels (e.g. daily) the script must be downloaded to a temp directory
-// first, because Invoke-Expression from a pipe does not support passing named
-// parameters such as -Version or -InstallFolder to the script.
-//
+// For all channels, the script is downloaded to a temp directory.
+// For daily channel, additional parameters (-Version, -InstallFolder) are passed
+// to the script. The install folder is escaped for PowerShell single-quoted strings
+// to handle paths containing apostrophes (e.g. O'Connor).
 // Returns the arguments to pass to the "powershell" command.
 func buildInstallScriptArgs(channel Channel) []string {
+	var scriptArgs string
 	switch channel {
 	case ChannelDaily:
-		script := fmt.Sprintf(
-			"$tmpScript = Join-Path $env:TEMP 'azd-install.ps1'; "+
-				"Invoke-RestMethod '%s' -OutFile $tmpScript; "+
-				"& $tmpScript -Version 'daily' -InstallFolder '%s'; "+
-				"Remove-Item $tmpScript -Force -ErrorAction SilentlyContinue",
-			installScriptURL, expectedPerUserInstallDir(),
-		)
-		return []string{"-NoProfile", "-ExecutionPolicy", "Bypass", "-Command", script}
+		scriptArgs = fmt.Sprintf(" -Version 'daily' -InstallFolder '%s'",
+			escapeForPSSingleQuote(expectedPerUserInstallDir()))
 	default:
-		script := fmt.Sprintf(
-			"$tmpScript = Join-Path $env:TEMP 'azd-install.ps1'; "+
-				"Invoke-RestMethod '%s' -OutFile $tmpScript; "+
-				"& $tmpScript; "+
-				"Remove-Item $tmpScript -Force -ErrorAction SilentlyContinue",
-			installScriptURL,
-		)
-		return []string{"-NoProfile", "-ExecutionPolicy", "Bypass", "-Command", script}
+		scriptArgs = " -Version 'stable'"
 	}
+
+	script := fmt.Sprintf(
+		"$tmpScript = Join-Path $env:TEMP 'azd-install.ps1'; "+
+			"Invoke-RestMethod '%s' -OutFile $tmpScript; "+
+			"& $tmpScript%s; "+
+			"Remove-Item $tmpScript -Force -ErrorAction SilentlyContinue",
+		installScriptURL, scriptArgs,
+	)
+	return []string{"-NoProfile", "-ExecutionPolicy", "Bypass", "-Command", script}
 }
--- a/cli/azd/pkg/update/msi_windows.go
+++ b/cli/azd/pkg/update/msi_windows.go
@@ -173,7 +173,10 @@ func buildInstallScriptArgs(channel Channel) []string {
 		return []string{"-NoProfile", "-ExecutionPolicy", "Bypass", "-Command", script}
 	default:
 		script := fmt.Sprintf(
-			"Invoke-RestMethod '%s' | Invoke-Expression",
+			"$tmpScript = Join-Path $env:TEMP 'azd-install.ps1'; "+
+				"Invoke-RestMethod '%s' -OutFile $tmpScript; "+
+				"& $tmpScript; "+
+				"Remove-Item $tmpScript -Force -ErrorAction SilentlyContinue",
 			installScriptURL,
 		)
 		return []string{"-NoProfile", "-ExecutionPolicy", "Bypass", "-Command", script}
--- a/cli/azd/pkg/update/msi_windows.go
+++ b/cli/azd/pkg/update/msi_windows.go
@@ -159,17 +159,15 @@ func isStandardMSIInstall() error {
 // first, because Invoke-Expression from a pipe does not support passing named
 // parameters such as -Version or -InstallFolder to the script.
 //
-// installDir is the target installation directory (e.g. %LOCALAPPDATA%\Programs\Azure Dev CLI).
-//
 // Returns the arguments to pass to the "powershell" command.
 func buildInstallScriptArgs(channel Channel) []string {
 	switch channel {
 	case ChannelDaily:
 		script := fmt.Sprintf(
-			"$tmpDir = Join-Path $env:TEMP 'azd-install.ps1'; "+
-				"Invoke-RestMethod '%s' -OutFile $tmpDir; "+
-				"& $tmpDir -Version 'daily' -InstallFolder '%s'; "+
-				"Remove-Item $tmpDir -Force -ErrorAction SilentlyContinue",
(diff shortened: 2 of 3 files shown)


spboyer (commented):
## Code Review

### 1. `azd update` now requires PowerShell 7
**`cli/azd/pkg/update/manager.go:513`** — High

The MSI update path changed from `powershell` to `pwsh`. `pwsh.exe` (PowerShell 7) is not installed by default on Windows 10/11 — only `powershell.exe` (5.1) ships pre-installed. This will cause `azd update` to fail on machines without PowerShell 7. The rest of the codebase handles this with a `pwsh` -> `powershell` fallback (see `pkg/tools/powershell/powershell.go:40-61`). Additionally, `verifyAuthenticode` at line 713 in the same file still uses `powershell`, creating an inconsistency.

**Fix:** Keep using `powershell` (the install script and `Get-AuthenticodeSignature` work fine under 5.1), or add an availability check with fallback.

---

### 2. Daily-channel install path not escaped for PowerShell
**`cli/azd/pkg/update/msi_windows.go:169`** — Medium

The daily update command formats `-InstallFolder '%s'` directly. If `%LOCALAPPDATA%` contains an apostrophe (e.g., username `O'Connor`), the generated PowerShell command becomes syntactically invalid.

**Fix:** Escape embedded `'` characters for PowerShell single-quoted literals, or pass the install folder as a separate argument.

cli/azd/pkg/update/manager.go
```diff
@@ -507,10 +507,10 @@ func (m *Manager) updateViaMSI(ctx context.Context, cfg *UpdateConfig, writer io
 			fmt.Errorf("failed to hash safety copy before install: %w", hashErr))
 	}
 
-	log.Printf("Running install script: powershell %s", strings.Join(psArgs, " "))
+	log.Printf("Running install script: pwsh %s", strings.Join(psArgs, " "))
 	fmt.Fprintf(writer, "Installing azd %s channel...\n", cfg.Channel)
 
-	runArgs := exec.NewRunArgs("powershell", psArgs...).
+	runArgs := exec.NewRunArgs("pwsh", psArgs...).
```
jongio: **[MEDIUM]** Tests: No test for the `pwsh` change

All tests in this PR cover `buildInstallScriptArgs` helpers, but the behavioral change to the executable name here has zero coverage. Since `commandRunner` is mockable, a test asserting the correct executable name passed to `NewRunArgs` would catch regressions.

cli/azd/pkg/update/msi_windows.go
```diff
@@ -150,35 +150,32 @@ func isStandardMSIInstall() error {
...
-		return "stable"
+		script := fmt.Sprintf(
+			"$tmpScript = Join-Path $env:TEMP 'azd-install.ps1'; "+
+				"Invoke-RestMethod '%s' -OutFile $tmpScript; "+
+				"& $tmpScript -Version 'daily' -InstallFolder '%s'; "+
+				"Remove-Item $tmpScript -Force -ErrorAction SilentlyContinue",
+			installScriptURL, expectedPerUserInstallDir(),
+		)
+		return []string{"-NoProfile", "-ExecutionPolicy", "Bypass", "-Command", script}
 	default:
```
jongio, line 169: **[MEDIUM]** Logic: Stable channel no longer passes `-Version` explicitly

The old code always passed `-Version 'stable'`. This path now pipes through `Invoke-Expression` with no `-Version` at all, making `azd update` depend on whatever default the remotely fetched script uses. The install script has conflicting documentation vs implementation for this default (`latest` in help text, `stable` in `param()`), so this is a real drift risk - a future script change could make stable updates install the wrong channel without any azd code change.

Keep passing `-Version 'stable'` explicitly. If that requires the downloaded-script path instead of piping, it aligns with the Copilot feedback about using download-to-file for stable anyway.

cli/azd/pkg/update/msi_windows.go
```diff
@@ -150,35 +150,32 @@ func isStandardMSIInstall() error {
...
+// buildInstallScriptArgs constructs the PowerShell arguments to run install-azd.ps1.
+//
+// For the stable channel the script is piped directly through Invoke-Expression.
+// No named parameters are needed because the default MSI install location is correct.
+//
+// For other channels (e.g. daily) the script must be downloaded to a temp directory
+// first, because Invoke-Expression from a pipe does not support passing named
+// parameters such as -Version or -InstallFolder to the script.
+//
+// Returns the arguments to pass to the "powershell" command.
```
jongio, line 162: **[LOW]** Quality: Doc says `powershell` but the PR changes `manager.go` to use `pwsh`.

jongio (changes requested):
## PR Review - Azure/azure-dev#7298

**[azd update] apply code sign verification for windows msi install** by @hemarina

### Findings

| Category | Medium | Low |
|----------|--------|-----|
| Refactoring | 1 | 0 |
| Tests | 1 | 0 |
| Logic | 1 | 0 |
| Quality | 0 | 1 |
| **Total** | **3** | **1** |

### Key Findings

1. **[MEDIUM] Refactoring**: Both channels should use the same download-and-execute pattern. The old code had one path for all channels (download to temp, run, clean up). This PR splits into two (pipe for stable, download for daily). If the stable channel is also changed to download-to-file (which Copilot's inline feedback already suggested for signature validation), the paths converge back to the old architecture. Simpler approach: keep the single download-and-execute pattern, just remove `-SkipVerify` and conditionally add `-Version`/`-InstallFolder` for daily.

2. **[MEDIUM] Tests**: The `powershell` -> `pwsh` change in `updateViaMSI` has no test coverage. All PR tests cover `buildInstallScriptArgs` helpers only.

3. **[MEDIUM] Logic**: Stable channel no longer passes `-Version 'stable'` explicitly. The remote install script has conflicting documentation vs implementation for its default (`latest` in help text, `stable` in `param()`), making this a real drift risk.

### Test Coverage Estimate

- `buildInstallScriptArgs()`: Good coverage - both channels tested with positive and negative assertions, structure validated
- `updateViaMSI()` pwsh change: **No coverage** - the behavioral change to which executable gets invoked isn't tested
- `versionFlag()` removal: Properly handled - tests removed since function was deleted

### What's Done Well

- Removing `-SkipVerify` is the right call - MSI Authenticode verification shouldn't be bypassed
- The daily channel download approach is solid - temp file, execute, clean up
- Tests use `wantNotContains` to verify stable doesn't leak daily-only parameters - good defensive testing
- Doc comment explains WHY stable and daily differ (pipe doesn't support named params)

---
*3 inline comments below. Review converged after 2 analysis passes + 1 deepening pass + cross-validation.
3 existing comments/threads (from Copilot and @spboyer) were already addressing issues we also found - those aren't repeated here.*

vhvb1989 (approved):

cli/azd/pkg/update/msi_windows_test.go
```diff
@@ -77,7 +63,10 @@ func TestBuildInstallScriptArgs(t *testing.T) {
 				"-Command",
 				installScriptURL,
 				"-Version 'stable'",
-				"-SkipVerify",
+				"Remove-Item",
+			},
+			wantNotContains: []string{
+				"-InstallFolder",
```
jongio, line 75: **[MEDIUM]** Tests: No regression guard for `-SkipVerify` removal

The whole point of this PR is removing `-SkipVerify` to enable Authenticode verification. But neither channel's test asserts that `-SkipVerify` is absent. If someone accidentally re-adds it in a future change, no test catches the regression.

Add `"-SkipVerify"` to `wantNotContains` for stable here, and add a `wantNotContains: []string{"-SkipVerify"}` for the daily channel too.

```suggestion
wantNotContains: []string{
"-InstallFolder",
"-SkipVerify",
},
```

cli/azd/pkg/update/msi_windows_test.go
```diff
@@ -243,3 +282,57 @@ func TestUpdateViaMSI_NonStandardInstallBlocks(t *testing.T) {
 	require.True(t, errors.As(err, &updateErr))
 	require.Equal(t, CodeNonStandardInstall, updateErr.Code)
 }
+
+// TestUpdateViaMSI_InvokesPowerShellWithCorrectArgs verifies that updateViaMSI calls
+// "powershell" with the arguments produced by buildInstallScriptArgs.
+func TestUpdateViaMSI_InvokesPowerShellWithCorrectArgs(t *testing.T) {
+	// Point LOCALAPPDATA at the test binary so isStandardMSIInstall passes.
```
jongio: **[MEDIUM]** Tests: This test always skips in CI/dev

Both `t.Skip` conditions fire in every realistic test environment: `go test` compiles to temp directories that never end with `Programs\Azure Dev CLI`, and `backupCurrentExe` would fail trying to rename the test binary. The integration coverage this was meant to provide is effectively zero.

This follows the existing `TestIsStandardMSIInstall_StandardPath` pattern in this file, so it's a known limitation. Consider adding a doc comment noting this test is for manual validation on actual MSI installs, or extracting `isStandardMSIInstall` into a mockable dependency.

jongio (commented):
## PR Review - Azure/azure-dev#7298

**[azd update] apply code sign verification for windows msi install** by @hemarina

### Findings

| Category | Medium |
|----------|--------|
| Tests | 2 |

### Key Findings

1. **[MEDIUM] Tests**: The PR's core goal (removing `-SkipVerify`) isn't guarded by a negative test assertion
2. **[MEDIUM] Tests**: `TestUpdateViaMSI_InvokesPowerShellWithCorrectArgs` never actually runs in CI

### Test Coverage Estimate

- `escapeForPSSingleQuote()`: Solid - 4 cases including apostrophe edge cases
- `buildInstallScriptArgs()`: Good - both channels, positive + negative assertions, structure validated
- `updateViaMSI()` integration: Effectively zero (test always skips)

### What's Done Well

- Removing `-SkipVerify` is the right call - Authenticode verification shouldn't be bypassed
- Both channels now use the same download-to-temp pattern - consistent and cleaner
- `escapeForPSSingleQuote` properly handles O'Connor-style paths
- `wantNotContains` assertions are good defensive testing for stable vs daily separation
- All PowerShell constructs used (`Join-Path`, `Invoke-RestMethod -OutFile`, `& $script` call operator, single-quote escaping) work identically on Windows PowerShell 5.1 and PowerShell 7+

---
*2 inline comments below. Review converged after 2 analysis passes + 1 deepening pass + cross-validation.
3 existing reviews were already addressing issues we also found - those aren't repeated here.*

cli/azd/pkg/update/msi_windows.go
```diff
@@ -150,35 +150,32 @@ func isStandardMSIInstall() error {
...
 
-// versionFlag returns the install script parameter value for the given channel.
-func versionFlag(channel Channel) string {
+func escapeForPSSingleQuote(s string) string {
+	return strings.ReplaceAll(s, "'", "''")
+}
+
+// buildInstallScriptArgs constructs the PowerShell arguments to run install-azd.ps1.
+// For all channels, the script is downloaded to a temp directory.
+// For daily channel, additional parameters (-Version, -InstallFolder) are passed
```
jongio: **[LOW]** Quality: Doc says `-Version` is a daily-only parameter, but it's passed for all channels

This line says "For daily channel, additional parameters (-Version, -InstallFolder) are passed" - but the code passes `-Version 'stable'` for the default case too (line 170). Only `-InstallFolder` is daily-specific.

```suggestion
// For daily channel, an additional parameter (-InstallFolder) is passed
```

jongio (commented):
## PR Review - Azure/azure-dev#7298

**[azd update] apply code sign verification for windows msi install** by @hemarina

### Summary

What: Removes `-SkipVerify` from `install-azd.ps1` invocation and refactors both update channels to use a download-to-temp-file pattern, enabling MSI Authenticode verification.
Why: Fixes #7265 - the old code bypassed code signing verification, which shouldn't be skipped for security.
Assessment: Solid implementation. Both channels now consistently download the script before executing, the apostrophe escaping handles edge case paths, and the test coverage is thorough with both positive and negative assertions. The existing review feedback has been well-addressed.

### Findings

| Category | Low |
|----------|-----|
| Quality | 1 |
| **Total** | **1** |

### Test Coverage Estimate

- `escapeForPSSingleQuote()`: Solid - 4 test cases including apostrophe edge cases
- `buildInstallScriptArgs()`: Good - both channels, positive + negative assertions, structure validated
- `TestIsStandardMSIInstall_StandardPath` removal: Acceptable - it always skipped in CI, NOTE comment documents the limitation

### What's Done Well

- Removing `-SkipVerify` is the right call - Authenticode verification shouldn't be bypassed
- Both channels now use the same download-to-temp pattern - consistent and easier to reason about
- `wantNotContains` assertions are good defensive testing for channel separation and `-SkipVerify` regression
- Apostrophe escaping for PowerShell single-quoted strings is correct and complete (`'` is the only special char)
- Stable channel now passes `-Version 'stable'` explicitly, eliminating drift risk with the remote script's default

---
*1 inline comment below.*

jongio (approved):
Reviewed 2 changed files (+92/-72). All 10 existing review threads have been addressed - both channels now consistently download-to-temp, `-SkipVerify` is removed with regression guards, apostrophe escaping handles edge case paths, and `-Version 'stable'` is explicit. LGTM.

PR #7311, opened 2026-03-25
fix: strip UTF-8 BOM from funcignore and fix negation pattern handling

jongio:
## Summary

Fixes `.funcignore` parsing failures caused by UTF-8 BOM (Byte Order Mark) and corrects negation pattern handling in the zip packaging path. Adds comprehensive test coverage for all `remoteBuild` × `.funcignore` combinations.

## Problem

A customer reported unexpected behavior when using `.funcignore` with `remoteBuild` settings on azd 1.23.12. Investigation revealed three bugs:

1. **UTF-8 BOM breaks pattern matching**: The `go-gitignore` library does not strip BOM (`\xef\xbb\xbf`) from file contents. Windows editors (e.g., Notepad pre-Win11) often add BOM to UTF-8 files, causing the first pattern (typically `node_modules`) to become `\xef\xbb\xbfnode_modules` — which never matches.

2. **Inconsistent parsing between validation and packaging**: The validation path (`resolveFunctionAppRemoteBuild`) used `gitignore.New()` with `bytes.NewReader()`, while the packaging path (`createDeployableZip`) used `gitignore.NewFromFile()`. The latter holds an open file handle (problematic on Windows temp dirs) and doesn't strip BOM.

3. **Negation patterns treated as exclusions**: In the packaging path, code checked `if match != nil` without checking `match.Ignore()`, so negation patterns like `!important_module` were incorrectly treated as exclusions.

## Fix

- Added `stripUTF8BOM()` helper function that removes BOM prefix if present
- Both validation and packaging paths now use `os.ReadFile()` → `stripUTF8BOM()` → `gitignore.New()` for consistent behavior
- Fixed negation pattern check to use `match != nil && match.Ignore()` in the packaging path

## Test Coverage (46 total: 8 existing + 38 new)

| Test Suite | Cases | Coverage |
|-----------|-------|----------|
| `BOMHandling` | 6 | BOM with various patterns and remoteBuild settings |
| `StripUTF8BOM` | 5 | BOM stripping utility function |
| `CustomerScenarios` | 9 | All 9 customer-reported scenarios (Case1-Case9) |
| `EdgeCases` | 10 | Empty funcignore, trailing slash, comments, CRLF, globs |
| `TypeScriptParity` | 2 | TypeScript follows same code path as JavaScript |

## Customer Scenario Matrix

| # | remoteBuild | .funcignore | excludes node_modules | Expected Result |
|---|-------------|-----------|----------------------|-----------------|
| 1 | nil | exists | yes | remoteBuild=true |
| 2 | nil | exists | no | remoteBuild=false |
| 3 | nil | missing | N/A | remoteBuild=true |
| 4 | false | missing | N/A | remoteBuild=false |
| 5 | false | exists | yes | ERROR (conflict) |
| 6 | false | exists | no | remoteBuild=false |
| 7 | true | exists | yes | remoteBuild=true |
| 8 | true | exists | no | ERROR (conflict) |
| 9 | true | missing | N/A | remoteBuild=true |

## Verification

- `mage preflight` passes (gofmt, go fix, copyright, lint 0 issues, cspell 0 issues, build, all tests)
- All 46 funcignore-related tests pass

### DIFF
--- a/cli/azd/pkg/project/service_target_functionapp_test.go
+++ b/cli/azd/pkg/project/service_target_functionapp_test.go
@@ -49,63 +49,84 @@ func TestNewFunctionAppTargetTypeValidation(t *testing.T) {
 	}
 }
 
+// TestResolveFunctionAppRemoteBuild_JavaScriptMatrix covers the full 3×3 matrix of remoteBuild
+// (nil, true, false) × funcignore state (excludes node_modules, doesn't exclude, absent).
+// Customer case numbers refer to the 9 scenarios reported in the GitHub issue for traceability.
 func TestResolveFunctionAppRemoteBuild_JavaScriptMatrix(t *testing.T) {
 	t.Parallel()
 
 	tests := []struct {
 		name              string
 		remoteBuild       *bool
-		funcIgnoreContent string
+		funcIgnoreContent string // empty string = no funcignore file
 		expectRemoteBuild bool
 		expectError       string
 	}{
+		// Customer Case 1: No flag + funcignore excludes node_modules → auto-detect remote build
 		{
-			name:              "NoRemoteBuildAndFuncIgnoreExcludesNodeModules_RemoteBuildEnabled",
+			name:              "NilRemoteBuild_FuncignoreExcludesNodeModules",
 			remoteBuild:       nil,
 			funcIgnoreContent: "node_modules\n",
 			expectRemoteBuild: true,
 		},
+		// Customer Case 2: No flag + funcignore doesn't exclude node_modules → local build
 		{
-			name:              "NoRemoteBuildAndFuncIgnoreDoesNotExcludeNodeModules_RemoteBuildDisabled",
+			name:              "NilRemoteBuild_FuncignoreDoesNotExcludeNodeModules",
 			remoteBuild:       nil,
 			funcIgnoreContent: "dist\n",
 			expectRemoteBuild: false,
 		},
+		// Customer Case 3: No flag + no funcignore → defaults to remote build
 		{
-			name:              "NoRemoteBuildAndMissingFuncIgnore_RemoteBuildEnabled",
+			name:              "NilRemoteBuild_NoFuncignore",
 			remoteBuild:       nil,
 			funcIgnoreContent: "",
 			expectRemoteBuild: true,
 		},
+		// Customer Case 4: Explicit false + no funcignore → local build
 		{
-			name:              "RemoteBuildFalseAndMissingFuncIgnore_RemoteBuildDisabled",
+			name:              "FalseRemoteBuild_NoFuncignore",
 			remoteBuild:       new(false),
 			funcIgnoreContent: "",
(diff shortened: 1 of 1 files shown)


cli/azd/pkg/project/project_utils.go
```diff
@@ -30,12 +31,17 @@ func createDeployableZip(svc *ServiceConfig, root string) (string, error) {
 	ignoreFile := svc.Host.IgnoreFile()
 	var ignorer gitignore.GitIgnore
 	if ignoreFile != "" {
-		ig, err := gitignore.NewFromFile(filepath.Join(root, ignoreFile))
-		if !errors.Is(err, fs.ErrNotExist) && err != nil {
+		ignoreFilePath := filepath.Join(root, ignoreFile)
+		contents, err := os.ReadFile(ignoreFilePath)
+		if errors.Is(err, fs.ErrNotExist) {
+			// no ignore file, use defaults below
+		} else if err != nil {
```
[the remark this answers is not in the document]
jongio: Good catch -- fixed in b708211. Added `zipFile.Close()` and `os.Remove(zipFile.Name())` on the error path before returning, preventing the file handle leak and orphaned temp file.

cli/azd/pkg/project/service_target_functionapp_test.go
```diff
@@ -159,3 +159,352 @@ func TestResolveFunctionAppRemoteBuild_NonJavaScriptDefaults(t *testing.T) {
...
+		t.Run(tt.name, func(t *testing.T) {
+			result := stripUTF8BOM(tt.input)
+			require.Equal(t, tt.expected, result)
+		})
+	}
+}
+
+// TestResolveFunctionAppRemoteBuild_CustomerScenarios maps 1:1 to the 9 scenarios reported by the
+// customer in the GitHub issue, using the same numbering for traceability.
+func TestResolveFunctionAppRemoteBuild_CustomerScenarios(t *testing.T) {
```
weikanglim: This is exercising the same test as `TestResolveFunctionAppRemoteBuild_JavaScriptMatrix` except with an added case for defaulting to `true` with no `.funcignore`. 

Worth merging the name changes here into the existing test?

weikanglim (approved):
Looks good! Thanks for fixing the Windows UTF-8 bom issue and the existing match pattern. 

Left a minor comment on test duplication.

cli/azd/pkg/project/service_target_functionapp_test.go
jongio: Good call — merged the unique case (Case 9: `true` + no funcignore → remote build) into `JavaScriptMatrix` and deleted `CustomerScenarios`. Also added customer case comments (Cases 1–9) to each entry in `JavaScriptMatrix` for traceability back to the reported scenarios.

37 tests pass (removed 9 duplicates, net -94 lines). See c7e7e159d.

jongio:
Do not merge.  Waiting for user to confirm.

PR #6856, opened 2026-02-24
feat: Extension SDK helpers — command scaffolding, MCP utilities, and security policy

jongio:
## Summary

Adds a set of SDK helpers to `pkg/azdext` that eliminate repetitive boilerplate across azd extensions, covering command scaffolding, MCP server construction, typed argument parsing, result helpers, and SSRF/path-traversal security policy.

Fixes #6855
Related: #6853

## Real-world adoption

These helpers have been fully adopted in two extension repos, demonstrating that they work end-to-end and dramatically reduce extension boilerplate:

| Repo | PR | Impact |
|------|----|--------|
| **azd-app** (12-tool MCP extension) | [jongio/azd-app#145](https://github.com/jongio/azd-app/pull/145) | **-574 lines**, deleted custom rate limiter, -75% arg parsing code per tool |
| **azd-core** (shared extension library) | [jongio/azd-core#22](https://github.com/jongio/azd-core/pull/22) | Deprecated stopgap helpers now superseded by `azdext` |

### Before & after: what extension authors write

**Root command setup** — from ~87 lines of manual Cobra + flag wiring to:
```go
rootCmd, ctx := azdext.NewExtensionRootCommand(azdext.ExtensionCommandOptions{
    Name: ""app"", Version: version, Short: ""My extension"",
})
// Standard flags, env var fallback, OTel propagation — all built in
```

**MCP tool handler** — from 8+ lines of untyped map access + manual error construction to:
```go
func handleRunService(ctx context.Context, args azdext.ToolArgs) (*mcp.CallToolResult, error) {
    name, err := args.RequireString(""name"")
    if err != nil { return azdext.MCPErrorResult(""%s"", err), nil }
    // ... business logic ...
    return azdext.MCPJSONResult(result)
}
```

**MCP server with rate limiting** — from a custom `TokenBucket` implementation (79 lines) to:
```go
builder := azdext.NewMCPServerBuilder(""my-server"", ""1.0"")
builder.WithRateLimit(10, 1.0) // 10 burst, 1 token/sec
builder.AddTool(""run_service"", ""Run a service"", handleRunService, azdext.MCPToolOptions{...})
server := builder.Build()
```

### Extension command scaffolding
- **`NewExtensionRootCommand`** — Creates a Cobra root command with standard azd global flags (`--debug`, `--no-prompt`, `--cwd`, `--environment`), OpenTelemetry trace context propagation, and automatic `AZD_*` env var fallback.
[...]

PR #7312, opened 2026-03-25
Add reserved global flags registry for extensions

### LINKED ISSUE #7307: Add reserved global flags registry for extensions
Extensions can currently reuse short flags that azd uses globally (`-e`, `-C`, `-o`, etc.) without any warning or enforcement. This caused #7271 when a fix for #7034 assumed `-e` was always an environment name.

We need a formal registry of reserved global flags so extensions know which short flags are off-limits, and so azd can detect conflicts.

### What this includes
- `ReservedFlag` registry in `internal/reserved_flags.go` listing all global short flags
- Helpers: `IsReservedShortFlag()`, `IsReservedLongFlag()`
- Documentation for extension authors listing reserved flags
- Optionally: build-time or install-time validation that warns when an extension uses a reserved flag

### Reserved flags (initial set)
| Short | Long | Purpose |
|-------|------|---------|
| `-e` | `--environment` | azd environment name |
| `-C` | `--cwd` | working directory |
| `-o` | `--output` | output format |
| `-h` | `--help` | help |
| | `--debug` | debug mode |
| | `--no-prompt` | non-interactive |
| | `--docs` | open docs |

Related: #7034, #7271, #7035, #7274

jongio:
## Summary

Adds a formal registry of reserved global short flags that extensions must not reuse. This prevents future conflicts like the `-e` collision that caused #7271.

### What's included

- `cli/azd/internal/reserved_flags.go` - `ReservedFlag` struct with 9 entries (`-e`, `-C`, `-o`, `--debug`, `--no-prompt`, etc.), lookup maps, and helper functions (`IsReserved`, `GetReserved`)
- `cli/azd/internal/reserved_flags_test.go` - 7 test functions with 29 subtests covering all lookup paths and edge cases

### Why

@vhvb1989 raised this concern on #7035 - there should be a way to block extensions from using known global flags. This is the foundation for that enforcement.

Closes #7307

Co-authored-by: Copilot <223556219+Copilot@users.noreply.github.com>

### DIFF
--- a/cli/azd/internal/reserved_flags.go
+++ b/cli/azd/internal/reserved_flags.go
@@ -16,11 +16,15 @@ type ReservedFlag struct {
 	Description string
 }
 
-// reservedFlags is the canonical list of global flags that extensions must not reuse.
-// It is derived from CreateGlobalFlagSet (auto_install.go), the root command's
-// persistent flags, and the extension SDK's built-in flag set (extension_command.go).
+// reservedFlags is the canonical, single-source-of-truth list of global flags that
+// extensions must not reuse. It is derived from CreateGlobalFlagSet (auto_install.go),
+// the root command's persistent flags, and the extension SDK's built-in flag set
+// (extension_command.go).
 //
-// Keep this list in sync whenever a new global flag is added to azd.
+// The SDK package (pkg/azdext) imports this list via ReservedFlags() rather than
+// maintaining its own copy.
+//
+// When adding a new global flag to azd, add it here and in CreateGlobalFlagSet.
 var reservedFlags = []ReservedFlag{
 	{Long: "environment", Short: "e", Description: "The name of the environment to use."},
 	{Long: "cwd", Short: "C", Description: "Sets the current working directory."},
--- a/cli/azd/pkg/azdext/reserved_flags.go
+++ b/cli/azd/pkg/azdext/reserved_flags.go
@@ -8,35 +8,16 @@ import (
 	"slices"
 	"strings"
 
+	"github.com/azure/azure-dev/cli/azd/internal"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
 )
 
-// reservedFlag describes a global flag owned by azd that extensions must not reuse
-// for a different purpose.
-type reservedFlag struct {
-	Long  string
-	Short string
-}
-
-// reservedGlobalFlags is the canonical list of global flags that extensions must not reuse.
-// Keep this in sync with internal.ReservedFlags and CreateGlobalFlagSet (auto_install.go).
-var reservedGlobalFlags = []reservedFlag{
-	{Long: "environment", Short: "e"},
-	{Long: "cwd", Short: "C"},
-	{Long: "debug", Short: ""},
-	{Long: "no-prompt", Short: ""},
-	{Long: "output", Short: "o"},
-	{Long: "help", Short: "h"},
-	{Long: "docs", Short: ""},
-	{Long: "trace-log-file", Short: ""},
-	{Long: "trace-log-url", Short: ""},
-}
-
-// reservedShorts is an index of short flag names built once at initialization time.
+// reservedShorts is an index of short flag names derived from the canonical
+// reserved flag list in internal.ReservedFlags(). Built once at initialization time.
 var reservedShorts = func() map[string]string {
-	m := make(map[string]string, len(reservedGlobalFlags))
-	for _, f := range reservedGlobalFlags {
+	m := make(map[string]string)
+	for _, f := range internal.ReservedFlags() {
 		if f.Short != "" {
 			m[f.Short] = f.Long
 		}
@@ -44,10 +25,11 @@ var reservedShorts = func() map[string]string {
 	return m
 }()
 
-// reservedLongs is an index of long flag names built once at initialization time.
+// reservedLongs is an index of long flag names derived from the canonical
+// reserved flag list in internal.ReservedFlags(). Built once at initialization time.
 var reservedLongs = func() map[string]bool {
-	m := make(map[string]bool, len(reservedGlobalFlags))
-	for _, f := range reservedGlobalFlags {
+	m := make(map[string]bool)
+	for _, f := range internal.ReservedFlags() {
 		m[f.Long] = true
 	}
 	return m
@@ -56,8 +38,9 @@ var reservedLongs = func() map[string]bool {
 // ReservedFlagNames returns the long names of all reserved global flags.
 // This is intended for documentation and error messages.
 func ReservedFlagNames() []string {
-	names := make([]string, len(reservedGlobalFlags))
-	for i, f := range reservedGlobalFlags {
+	flags := internal.ReservedFlags()
+	names := make([]string, len(flags))
+	for i, f := range flags {
 		names[i] = f.Long
 	}
 	return names
--- a/cli/azd/pkg/azdext/reserved_flags_test.go
+++ b/cli/azd/pkg/azdext/reserved_flags_test.go
@@ -6,7 +6,6 @@ package azdext
 import (
 	"testing"
 
-	"github.com/azure/azure-dev/cli/azd/internal"
 	"github.com/spf13/cobra"
 	"github.com/stretchr/testify/require"
 )
@@ -128,46 +127,6 @@ func TestReservedFlagNames(t *testing.T) {
 	require.Contains(t, names, "debug")
 	require.Contains(t, names, "output")
 	require.Contains(t, names, "help")
-}
-
-func TestReservedFlagsInSyncWithInternal(t *testing.T) {
-	// Verify the SDK reserved flag list stays in sync with the internal registry.
-	// If this test fails, you added a flag to one list but not the other.
-	sdkFlags := reservedGlobalFlags
-	internalFlags := internal.ReservedFlags()
-
-	// Build maps of long name -> short name for both sides.
-	sdkMap := make(map[string]string, len(sdkFlags))
-	for _, f := range sdkFlags {
-		sdkMap[f.Long] = f.Short
-	}
-
-	internalMap := make(map[string]string, len(internalFlags))
-	for _, f := range internalFlags {
-		internalMap[f.Long] = f.Short
-	}
-
-	// Check every SDK flag exists in internal with matching short name.
-	for long, short := range sdkMap {
-		internalShort, ok := internalMap[long]
-		require.True(t, ok,
-			"azdext reserved_flags.go has %q but internal.ReservedFlags() does not — add it to internal/reserved_flags.go",
-			long)
-		require.Equal(t, internalShort, short,
-			"short name mismatch for %q: internal has %q, SDK has %q",
-			long, internalShort, short)
-	}
-
-	// Check every internal flag exists in SDK with matching short name.
-	for long, short := range internalMap {
-		sdkShort, ok := sdkMap[long]
-		require.True(t, ok,
-			"internal.ReservedFlags() has %q but azdext reserved_flags.go does not — add it to pkg/azdext/reserved_flags.go",
-			long)
-		require.Equal(t, sdkShort, short,
-			"short name mismatch for %q: SDK has %q, internal has %q",
-			long, sdkShort, short)
-	}
 }
 
 func TestValidateNoReservedFlagConflicts_NonSDKRootPersistentFlag(t *testing.T) {
--- a/cli/azd/cmd/auto_install.go
+++ b/cli/azd/cmd/auto_install.go
@@ -710,6 +710,10 @@ func ExecuteWithAutoInstall(ctx context.Context, rootContainer *ioc.NestedContai
 
 // CreateGlobalFlagSet creates a new flag set with all global flags defined.
 // This is the single source of truth for global flag definitions.
+//
+// When adding or modifying flags here, also update the reserved flag registry
+// in internal/reserved_flags.go. The SDK package (pkg/azdext) derives its
+// reserved flag indexes from that registry automatically.
 func CreateGlobalFlagSet() *pflag.FlagSet {
 	globalFlags := pflag.NewFlagSet("global", pflag.ContinueOnError)
 


cli/azd/internal/reserved_flags.go
```diff
@@ -0,0 +1,77 @@
...
+	{Long: "environment", Short: "e", Description: "The name of the environment to use."},
+	{Long: "cwd", Short: "C", Description: "Sets the current working directory."},
+	{Long: "debug", Short: "", Description: "Enables debugging and diagnostics logging."},
+	{Long: "no-prompt", Short: "", Description: "Accepts the default value instead of prompting."},
+	{Long: "output", Short: "o", Description: "The output format (json, table, none)."},
+	{Long: "help", Short: "h", Description: "Help for the current command."},
+	{Long: "docs", Short: "", Description: "Opens the documentation for the current command."},
+	{Long: "trace-log-file", Short: "", Description: "Write a diagnostics trace to a file."},
+	{Long: "trace-log-url", Short: "", Description: "Send traces to an Open Telemetry compatible endpoint."},
+}
```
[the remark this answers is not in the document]
jongio: Fixed in c58fccdcf. Made the slice unexported (`reservedFlags`) with a `ReservedFlags()` getter that returns a defensive copy. Maps are derived from the same unexported source of truth.

cli/azd/internal/reserved_flags.go
```diff
@@ -0,0 +1,77 @@
...
+	reservedShortFlags = make(map[string]ReservedFlag, len(ReservedFlags))
+	reservedLongFlags = make(map[string]ReservedFlag, len(ReservedFlags))
+	for _, f := range ReservedFlags {
+		reservedLongFlags[f.Long] = f
+		if f.Short != "" {
+			reservedShortFlags[f.Short] = f
+		}
+	}
+}
+
```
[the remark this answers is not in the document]
jongio: Fixed in c58fccdcf. Replaced `init()` with func-literal var initializers for both maps, eliminating the init ordering hazard. The maps now reference the unexported `reservedFlags` slice directly.

cli/azd/internal/reserved_flags.go
```diff
@@ -0,0 +1,77 @@
...
+var ReservedFlags = []ReservedFlag{
+	{Long: "environment", Short: "e", Description: "The name of the environment to use."},
+	{Long: "cwd", Short: "C", Description: "Sets the current working directory."},
+	{Long: "debug", Short: "", Description: "Enables debugging and diagnostics logging."},
+	{Long: "no-prompt", Short: "", Description: "Accepts the default value instead of prompting."},
+	{Long: "output", Short: "o", Description: "The output format (json, table, none)."},
+	{Long: "help", Short: "h", Description: "Help for the current command."},
+	{Long: "docs", Short: "", Description: "Opens the documentation for the current command."},
+	{Long: "trace-log-file", Short: "", Description: "Write a diagnostics trace to a file."},
+	{Long: "trace-log-url", Short: "", Description: "Send traces to an Open Telemetry compatible endpoint."},
```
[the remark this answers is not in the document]
jongio: Fixed in c58fccdcf. Updated to "OpenTelemetry-compatible" per the standard product name spelling.

spboyer (commented):
## Code Review

### 1. Universal reserved-flag check will break existing extensions
**`cli/azd/pkg/azdext/reserved_flags.go:22`** — High

The registry treats `environment`, `output`, `docs`, etc. as universally reserved for every extension. `azdext.Run()` exits on any match before command execution. This is broader than what azd actually pre-parses today (`CreateGlobalFlagSet` only pre-parses `cwd`, `debug`, `no-prompt`, `trace-log-file`, `trace-log-url`). In-tree extensions like `azure.ai.agents` (uses `-e` for `--environment`, `-o` for `--output`) and `azure.ai.finetune` (uses `-e` for `--project-endpoint`) will fail at startup.

**Fix:** Only reserve flags that are truly universal, or scope the extra reservations to roots created with `NewExtensionRootCommand()` that actually register those globals.

---

### 2. `collectConflicts` skip logic masks subcommand flag conflicts
**`cli/azd/pkg/azdext/reserved_flags.go` (`collectConflicts` function)** — Medium

The skip logic uses name-based lookup (`root.PersistentFlags().Lookup(f.Name)`) to avoid flagging the SDK's own persistent flags. But before Cobra's `mergePersistentFlags` (which happens during `Execute()`, not validation), a subcommand's `cmd.Flags().VisitAll()` only yields locally-defined flags. If a subcommand locally defines a flag with the same long name as a root persistent flag (e.g., `--output` meaning "output file" instead of "output format"), the name lookup finds root's flag and skips validation entirely. The conflict goes unreported.

**Fix:** Use pointer equality instead of name equality:
```go
if rootFlag := root.PersistentFlags().Lookup(f.Name); rootFlag != nil && rootFlag == f {
    return
}
```

cli/azd/docs/design/extension-flag-architecture.md
```diff
@@ -0,0 +1,188 @@
...
+1. Walks the entire command tree
+2. For each command's flags, checks both long and short names against the reserved list
+3. Skips flags on the root command's persistent flag set (those are the SDK-provided azd-compatible flags)
+4. Returns a detailed error listing all conflicts with remediation guidance
+
+Any extension built with the azd SDK gets this check automatically — no opt-in required.
+
+### Sync Test
+
+A test (`TestReservedFlagNames_SyncWithInternal`) ensures the SDK-side and host-side reserved flag lists stay in sync. If a developer adds a new global flag to one list but not the other, the test fails.
```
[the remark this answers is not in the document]
jongio: Fixed in e31806963. Updated test name reference to TestReservedFlagsInSyncWithInternal.

cli/azd/pkg/azdext/reserved_flags_test.go
```diff
@@ -0,0 +1,147 @@
...
+	for name := range sdkSet {
+		require.True(t, internalSet[name],
+			"azdext reserved_flags.go has %q but internal.ReservedFlags() does not — add it to internal/reserved_flags.go",
+			name)
+	}
+
+	for name := range internalSet {
+		require.True(t, sdkSet[name],
+			"internal.ReservedFlags() has %q but azdext reserved_flags.go does not — add it to pkg/azdext/reserved_flags.go",
+			name)
```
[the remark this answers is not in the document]
jongio: Fixed in e31806963. Rewrote the sync test to compare both long AND short names between SDK and internal registries. If either side changes a shorthand, the test will catch it.

cli/azd/pkg/azdext/reserved_flags.go
```diff
@@ -0,0 +1,181 @@
...
+		// Those are the SDK-provided azd-compatible flags and are intentional.
+		if rootFlag := root.PersistentFlags().Lookup(f.Name); rootFlag != nil {
+			return
+		}
+
+		if c, ok := checkFlag(cmd, f); ok {
+			conflicts = append(conflicts, c)
+		}
+	})
+
```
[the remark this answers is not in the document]
jongio: Fixed in e31806963 with a simpler approach: added cmd.PersistentFlags().VisitAll() alongside the existing cmd.Flags().VisitAll(), with a checked map to deduplicate. The root persistent flag skip logic applies to both. Added TestValidateNoReservedFlagConflicts_PersistentFlagCollision test to verify.

cli/azd/docs/design/extension-flag-architecture.md
```diff
@@ -0,0 +1,188 @@
...
+|-----------|-------|------|---------|--------|-------------|
+| `environment` | `e` | string | `$AZURE_ENV_NAME` | No | The name of the environment to use |
+| `cwd` | `C` | string | `""` | No | Sets the current working directory |
+| `debug` | — | bool | `false` | No | Enables debugging and diagnostics logging |
+| `no-prompt` | — | bool | `false` | No | Accepts default value instead of prompting |
+| `output` | `o` | string | `"default"` | No | The output format (json, table, none) |
+| `help` | `h` | — | — | No | Help for the current command (cobra built-in) |
+| `docs` | — | — | — | No | Opens documentation for the current command |
+| `trace-log-file` | — | string | `""` | Yes | Write a diagnostics trace to a file |
+| `trace-log-url` | — | string | `""` | Yes | Send traces to an OpenTelemetry endpoint |
```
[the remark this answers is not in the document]
jongio: Fixed in e31806963. Restructured the spec to accurately categorize flags by registration mechanism: Pre-parsed (CreateGlobalFlagSet), Root Persistent Flags (
oot.go), and Cobra Built-in/Custom. Each section describes the actual code path.

jongio:
Responding to @spboyer's [review](https://github.com/Azure/azure-dev/pull/7312#pullrequestreview-4011599180):

**1. Universal reserved-flag check will break existing extensions**

All 9 flags in the reserved list are legitimate azd globals: `-e/--environment`, `-C/--cwd`, `-o/--output`, `--debug`, `--no-prompt`, `--help/-h`, `--docs`, and the trace flags. You're right that `CreateGlobalFlagSet` only pre-parses a subset today, but these are all flags that azd defines at the root level. The reserved list matches the actual global flag surface, not just what's pre-parsed.

The extensions that currently collide (`azure.ai.agents` uses `-e` for `--environment`, `azure.ai.models`/`azure.ai.finetune` use `-e` for `--project-endpoint`) are all being migrated in PR #7313. No extension will break at startup because the migration lands first.

**2. collectConflicts skip logic masks subcommand flag conflicts**

Good catch. Implemented pointer equality in d14bf4b9a — now only skips when the flag is literally the same `*pflag.Flag` object inherited from root, not just any flag with the same name.

cli/azd/pkg/azdext/run.go
```diff
@@ -61,6 +61,13 @@ func Run(rootCmd *cobra.Command, opts ...RunOption) {
...
 
+	// Validate that extension-defined flags do not collide with azd reserved global flags.
+	// This check runs before execution so extension developers see the error immediately.
+	if err := ValidateNoReservedFlagConflicts(rootCmd); err != nil {
+		printError(err)
+		os.Exit(1)
+	}
+
 	if cfg.preExecute != nil {
 		if err := cfg.preExecute(ctx, rootCmd); err != nil {
```
wbreza, line 75: **[High] Validation errors silently lost for listen-mode extensions**

When azd launches an extension via gRPC (the standard `listen` lifecycle), stderr is wired to `io.Discard`. This block only calls `printError(err)` → `os.Exit(1)`, so the detailed conflict message (listing flag names, command paths, remediation guidance) is silently dropped. The developer sees a generic "extension failed" from the host with zero explanation.

The two other error blocks in this same function (`preExecute` and `ExecuteContext`) both use `ReportError(ctx, err)` first — which sends the error back via gRPC where it actually reaches the user — then fall back to `printError`. This block should follow the same pattern.

**Suggested Fix:**
```go
if err := ValidateNoReservedFlagConflicts(rootCmd); err != nil {
    if reportErr := ReportError(ctx, err); reportErr != nil {
        printError(err)
    }
    os.Exit(1)
}
```

cli/azd/pkg/azdext/reserved_flags.go
```diff
@@ -0,0 +1,198 @@
...
+		b.WriteString(c.String())
+		b.WriteString("\n")
+	}
+	b.WriteString("Remove or rename these flags to avoid conflicts with azd's global flags.\n")
+	b.WriteString("Reserved flags: ")
+	b.WriteString(strings.Join(ReservedFlagNames(), ", "))
+	return fmt.Errorf("%s", b.String())
+}
+
+// collectConflicts recursively collects flag conflicts from the command tree.
```
wbreza, line 102: **[Medium] `checkFlag` short-circuits on first collision type, hiding compound conflicts**

When a flag has both a short-name AND long-name collision (e.g., `--environment -e` on a subcommand conflicts on both dimensions), `checkFlag` returns after detecting the short collision. The long-name collision is never reported.

This means a developer sees "short flag `-e` is reserved," renames the short flag, re-runs, and then gets a *second* error for the long name — an unnecessary fix-and-fail cycle.

Consider collecting both conflicts before returning, so the developer can fix everything in one pass.

cli/azd/pkg/azdext/reserved_flags.go
```diff
@@ -0,0 +1,198 @@
...
+	{Long: "help", Short: "h"},
+	{Long: "docs", Short: ""},
+	{Long: "trace-log-file", Short: ""},
+	{Long: "trace-log-url", Short: ""},
+}
+
+// reservedShorts and reservedLongs are lookup maps built at init time.
+var (
+	reservedShorts map[string]string // short -> long
+	reservedLongs  map[string]bool
```
wbreza: **[Low] `init()` vs func-literal inconsistency with twin file**

The resolved review feedback about init-ordering hazards was applied to `internal/reserved_flags.go` (uses func-literal var initializers), but this file still uses an explicit `init()`. No functional bug, but it's an inconsistency between two files the spec describes as "kept in sync."

cli/azd/pkg/azdext/reserved_flags.go
```diff
@@ -0,0 +1,198 @@
...
+				Command:      cmdPath,
+				FlagName:     f.Name,
+				FlagShort:    f.Shorthand,
+				ReservedLong: reservedLong,
+				Reason:       fmt.Sprintf("short flag -%s is reserved by azd for --%s", f.Shorthand, reservedLong),
+			}, true
+		}
+	}
+
+	// Check long flag collision (only if the long name is the same but used for
```
wbreza, line 165: **[Low] `commandPath` uses O(n²) prepend-based slice building**

`parts = append([]string{c.Name()}, parts...)` allocates a new backing array on every level. Consider `append` + `slices.Reverse` per AGENTS.md modern Go guidance. Negligible in practice (command trees are shallow), but it's an anti-pattern in a new file.

cli/azd/pkg/azdext/run.go
jongio: Fixed — validation errors now report via `ReportError(ctx, err)` before `os.Exit(1)`, matching the pattern used by the other error blocks in the same function.

cli/azd/pkg/azdext/reserved_flags.go
jongio: Fixed — `checkFlag` now returns `[]FlagConflict`, collecting both short-name and long-name conflicts before returning. Developers see all issues in one pass.

cli/azd/pkg/azdext/reserved_flags.go
jongio: Fixed — switched to func-literal var initializers (`var reservedShorts = func() ...{}()`) matching the pattern in `internal/reserved_flags.go`.

cli/azd/pkg/azdext/reserved_flags.go
jongio: Fixed — using `append` + `slices.Reverse` instead of prepend-based slice building.

cli/azd/pkg/azdext/reserved_flags.go
```diff
@@ -0,0 +1,203 @@
...
+			return
+		}
+		checked[f.Name] = struct{}{}
+
+		// Skip flags that are the SDK-provided root persistent flags (same object).
+		// Use pointer equality so that a subcommand defining its own flag with the
+		// same name as a root persistent flag is still validated.
+		if rootFlag := root.PersistentFlags().Lookup(f.Name); rootFlag != nil && rootFlag == f {
+			return
+		}
```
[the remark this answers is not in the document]
jongio: Good catch. Tightened the exemption: only roots created by NewExtensionRootCommand (marked with an azd-sdk-root annotation) get the pointer-equality exemption. Plain roots that manually add root persistent flags colliding with reserved globals are now validated. Added test to verify. Fixed in 3ed8901.

vhvb1989:
## Observation: Two extension authoring models and the enforcement gap

After [PR #6856](https://github.com/Azure/azure-dev/pull/6856) introduced `NewExtensionRootCommand()` and `azdext.Run()`, there are now **two distinct ways to build azd extensions**:

1. **SDK-Managed** — uses `NewExtensionRootCommand()` / `azdext.Run()`. Gets automatic global flag registration, OTel propagation, structured error reporting, and (with this PR) reserved flag enforcement via `ValidateNoReservedFlagConflicts()`. **Go only.**

2. **Self-Managed** — the extension wires its own CLI root and parses flags directly. Works with **any language** (Python, JS, .NET, Go). This is how 6 of 9 existing in-repo extensions are built, and what the Go scaffold template (`azd x init`) generates.

The reserved flag enforcement in this PR only applies to extensions that opt into `azdext.Run()` — which is currently the minority. Extensions that skip it (or are written in any non-Go language) get **no protection** against flag collisions. There is no equivalent framework for Python, JavaScript, or .NET.

This creates a gap that extension authors need to understand when choosing how to build their extension. I've created:

- **Issue [#7405](https://github.com/Azure/azure-dev/issues/7405)** — describes the two models and the need for documentation
- **PR [#7406](https://github.com/Azure/azure-dev/pull/7406)** — adds `extension-authoring-models.md` documenting both models with tradeoffs, decision guide, and best practices

Worth considering whether the design doc in this PR (`extension-flag-architecture.md`) should reference the authoring models doc to help readers understand which extensions the enforcement applies to.

jongio:
Great analysis. You're right — enforcement only covers SDK-managed extensions using `azdext.Run()`, which is currently 3 of 9 in-repo extensions. The other 6 use the self-managed model and bypass it.

Since all in-repo extensions are Go, the simplest path is to have each self-managed extension add a `ValidateNoReservedFlagConflicts(rootCmd)` call in their `main.go` — one line per extension. Alternatively they can migrate to `azdext.Run()` for the full SDK lifecycle.

Added a 'Scope & Limitations' section to the design doc in this PR referencing #7405 and your authoring models doc in #7406. The migration can happen incrementally as a follow-up.

cli/azd/internal/repository/detect_confirm_apphost.go
```diff
@@ -83,13 +84,12 @@ func (d *detectConfirmAppHost) render(ctx context.Context) error {
...
-		d.console.Message(ctx, "")
-	}
-
-	d.console.Message(ctx, "azd will generate the files necessary to host your app on Azure.")
+	d.console.Message(
+		ctx,
+		"azd will generate the files necessary to host your app on Azure using "+color.MagentaString(
+			"Azure Container Apps",
+		)+".\n",
+	)
```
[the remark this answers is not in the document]
jongio: This file isn't part of this PR's changes - the PR modifies 8 files related to reserved flag enforcement, none of which is detect_confirm_apphost.go. Resolving as not applicable to this PR.

cli/azd/pkg/azdext/reserved_flags.go
```diff
@@ -0,0 +1,205 @@
...
+
+// reservedFlag describes a global flag owned by azd that extensions must not reuse
+// for a different purpose.
+type reservedFlag struct {
+	Long  string
+	Short string
+}
+
+// reservedGlobalFlags is the canonical list of global flags that extensions must not reuse.
+// Keep this in sync with internal.ReservedFlags and CreateGlobalFlagSet (auto_install.go).
```
JeffreyCA: Could we add a reference comment in `auto_install.go` as well?

cli/azd/internal/reserved_flags.go
```diff
@@ -0,0 +1,88 @@
...
+	// Description explains the flag's purpose in azd.
+	Description string
+}
+
+// reservedFlags is the canonical list of global flags that extensions must not reuse.
+// It is derived from CreateGlobalFlagSet (auto_install.go), the root command's
+// persistent flags, and the extension SDK's built-in flag set (extension_command.go).
+//
+// Keep this list in sync whenever a new global flag is added to azd.
+var reservedFlags = []ReservedFlag{
```
weikanglim: I'm slightly concerned with the duplication of global flags across 3 different files. Is it worth   centralizing it today? I believe we're blocked waiting on the extension updates to occur and we may have more time to clean this up further.

cli/azd/internal/reserved_flags.go
```diff
@@ -0,0 +1,88 @@
...
+	{Long: "environment", Short: "e", Description: "The name of the environment to use."},
+	{Long: "cwd", Short: "C", Description: "Sets the current working directory."},
+	{Long: "debug", Short: "", Description: "Enables debugging and diagnostics logging."},
+	{Long: "no-prompt", Short: "", Description: "Accepts the default value instead of prompting."},
+	{Long: "output", Short: "o", Description: "The output format (json, table, none)."},
+	{Long: "help", Short: "h", Description: "Help for the current command."},
+	{Long: "docs", Short: "", Description: "Opens the documentation for the current command."},
+	{Long: "trace-log-file", Short: "", Description: "Write a diagnostics trace to a file."},
+	{Long: "trace-log-url", Short: "", Description: "Send traces to an OpenTelemetry-compatible endpoint."},
```
weikanglim, line 37: These feel like internals that `azd` should handle through the SDK rather than via the CLI interface.

cli/azd/internal/reserved_flags.go
```diff
@@ -0,0 +1,88 @@
...
+// It is derived from CreateGlobalFlagSet (auto_install.go), the root command's
+// persistent flags, and the extension SDK's built-in flag set (extension_command.go).
+//
+// Keep this list in sync whenever a new global flag is added to azd.
+var reservedFlags = []ReservedFlag{
+	{Long: "environment", Short: "e", Description: "The name of the environment to use."},
+	{Long: "cwd", Short: "C", Description: "Sets the current working directory."},
+	{Long: "debug", Short: "", Description: "Enables debugging and diagnostics logging."},
+	{Long: "no-prompt", Short: "", Description: "Accepts the default value instead of prompting."},
+	{Long: "output", Short: "o", Description: "The output format (json, table, none)."},
```
weikanglim, line 33: The output and environment flag are NOT global in azd. They are lit up per-command based on the intended usage. How should one reconcile the differences?

cli/azd/pkg/azdext/reserved_flags.go
jongio: Added a cross-reference comment to `CreateGlobalFlagSet()` in `auto_install.go` pointing to both `internal/reserved_flags.go` and `pkg/azdext/reserved_flags.go` so devs adding new global flags know to update the reserved registries too.

cli/azd/internal/reserved_flags.go
jongio: These are already registered as CLI flags - +""+--docs+""+ is public (visible in +""+--help+""+ on every command via +""+cobra_builder.go+""+), and the trace flags are hidden but pre-parsed by +""+CreateGlobalFlagSet+""+ in +""+uto_install.go+""+. The SDK also mirrors them in +""+xtension_command.go+""+ for host-to-extension propagation.

They need to be reserved because azd pre-parses them before dispatching to the extension. If an extension defines +""+--trace-log-file+""+ for a different purpose, azd's pre-parsing would consume the value, causing silent breakage. The reservation prevents that shadowing.

cli/azd/internal/reserved_flags.go
jongio: You're right that `--environment/-e` and `--output/-o` aren't registered globally via `CreateGlobalFlagSet` - they're added per-command. But they still need to be reserved for extensions because:

1. **Short flags are the scarce resource.** If an extension uses `-e` for `--endpoint`, any azd command that adds `--environment` on the same command path creates a shorthand collision. Cobra panics on duplicate shorthands, so the extension breaks when azd evolves.
2. **User expectations.** Users expect `-e` to mean environment across azd. An extension that uses `-e` for something else creates confusion even if there's no technical collision today.

The reserved list is better understood as "namespace reservations" rather than "currently global flags." The per-command vs. global distinction matters for how azd registers them internally but not for whether extensions should avoid them.

cli/azd/internal/reserved_flags.go
jongio: Centralized it. `internal/reserved_flags.go` is now the single source of truth. `pkg/azdext/reserved_flags.go` no longer maintains its own copy - it imports `internal.ReservedFlags()` and builds its lookup indexes from that. Removed the sync test since there's only one list now. Also added a cross-reference comment in `CreateGlobalFlagSet()` so developers know to update the registry when adding new global flags.

PR #7341, opened 2026-03-26
fix(azure.ai.agents): skip scale config for vnext hosted agents

### LINKED ISSUE #7267: Remove scale from config of azure.yaml for hosted agents
When enableHostedAgentVNext=true, do not add `scale` to the congif in azure.yaml:

```
scale:
    maxReplicas: 1
```

This is not needed/used for vnext backend.

spboyer:
## Summary

Fixes #7267

When `enableHostedAgentVNext=true`, the platform manages container lifecycle and scaling. The `scale` block in `azure.yaml` is unused by the vnext backend and should not be written during agent init.

## Changes

- **`init.go`** — `populateContainerSettings()` conditionally adds Scale only when vnext is not enabled
- **`init_from_code.go`** — `addToProject()` conditionally adds Scale the same way
- **`listen.go`** — Preserves existing Scale settings from azure.yaml (no silent data loss), but skips creating new defaults when vnext is enabled

## Backward Compatibility

- Existing azure.yaml files with `scale:` still work — settings are preserved
- `scale` is `omitempty` in both Go struct and JSON schema — no schema changes needed
- Deployment code (`service_target_agent.go:861`) already handles nil Scale gracefully
- No breaking changes

## Testing

```bash
cd cli/azd/extensions/azure.ai.agents && go build
```

### DIFF
--- a/cli/azd/extensions/azure.ai.agents/internal/cmd/vnext_scale_test.go
+++ b/cli/azd/extensions/azure.ai.agents/internal/cmd/vnext_scale_test.go
@@ -1,0 +1,174 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package cmd
+
+import (
+	"azureaiagent/internal/project"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestIsVNextEnabled(t *testing.T) {
+	tests := []struct {
+		name     string
+		envValue string
+		want     bool
+	}{
+		{name: "enabled with true", envValue: "true", want: true},
+		{name: "enabled with 1", envValue: "1", want: true},
+		{name: "enabled with TRUE", envValue: "TRUE", want: true},
+		{name: "disabled with false", envValue: "false", want: false},
+		{name: "disabled with 0", envValue: "0", want: false},
+		{name: "invalid value falls back", envValue: "notabool", want: false},
+		{name: "unset falls back", envValue: "", want: false},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if tt.envValue != "" {
+				t.Setenv("enableHostedAgentVNext", tt.envValue)
+			}
+
+			got := isVNextEnabled(t.Context())
+			assert.Equal(t, tt.want, got)
+		})
+	}
+}
+
+func TestVNextConditionalScale(t *testing.T) {
+	tests := []struct {
+		name      string
+		envValue  string
+		wantScale bool
+	}{
+		{
+			name:      "vnext disabled - scale defaults applied",
+			envValue:  "",
+			wantScale: true,
+		},
+		{
+			name:      "vnext enabled - scale omitted",
+			envValue:  "true",
+			wantScale: false,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if tt.envValue != "" {
+				t.Setenv("enableHostedAgentVNext", tt.envValue)
+			}
+
+			// Mirrors the pattern used in init.go, init_from_code.go, and listen.go
+			container := &project.ContainerSettings{
+				Resources: &project.ResourceSettings{
+					Memory: project.DefaultMemory,
+					Cpu:    project.DefaultCpu,
+				},
+			}
+
+			if !isVNextEnabled(t.Context()) {
+				container.Scale = &project.ScaleSettings{
+					MinReplicas: project.DefaultMinReplicas,
+					MaxReplicas: project.DefaultMaxReplicas,
+				}
+			}
+
+			if tt.wantScale {
+				assert.NotNil(t, container.Scale, "Scale should be set when vnext is disabled")
+				assert.Equal(t, project.DefaultMinReplicas, container.Scale.MinReplicas)
+				assert.Equal(t, project.DefaultMaxReplicas, container.Scale.MaxReplicas)
+			} else {
+				assert.Nil(t, container.Scale, "Scale should be nil when vnext is enabled")
+			}
+		})
+	}
+}
+
+func TestVNextPreservesExistingScale(t *testing.T) {
+	// Mirrors the listen.go pattern: existing scale settings are always preserved,
+	// regardless of vnext status.
+	tests := []struct {
+		name          string
+		envValue      string
+		existingScale *project.ScaleSettings
+		wantScale     bool
+		wantMin       int
+		wantMax       int
+	}{
+		{
+			name:      "vnext enabled, no existing scale - omitted",
+			envValue:  "true",
+			wantScale: false,
+		},
+		{
+			name:          "vnext enabled, existing scale - preserved",
+			envValue:      "true",
+			existingScale: &project.ScaleSettings{MinReplicas: 2, MaxReplicas: 5},
+			wantScale:     true,
+			wantMin:       2,
+			wantMax:       5,
+		},
+		{
+			name:      "vnext disabled, no existing scale - defaults applied",
+			envValue:  "",
+			wantScale: true,
+			wantMin:   project.DefaultMinReplicas,
+			wantMax:   project.DefaultMaxReplicas,
+		},
+		{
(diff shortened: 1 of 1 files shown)


cli/azd/extensions/azure.ai.agents/internal/cmd/init_from_code.go
```diff
@@ -742,10 +742,13 @@ func (a *InitFromCodeAction) addToProject(ctx context.Context, targetDir string,
...
-		Scale: &project.ScaleSettings{
+	}
+
+	if !isVNextEnabled(ctx) {
+		agentConfig.Container.Scale = &project.ScaleSettings{
 			MinReplicas: project.DefaultMinReplicas,
 			MaxReplicas: project.DefaultMaxReplicas,
-		},
+		}
 	}
```
jongio, line 752: **MEDIUM — Inconsistent client pass-through**

`InitFromCodeAction` has an `azdClient` field (line 31), and the sibling call in `init.go:1241` correctly passes `a.azdClient` — but this call site does not. This creates an unnecessary gRPC connection + env read each time `addToProject` runs.

```suggestion
	if !isVNextEnabled(ctx, a.azdClient) {
```

cli/azd/extensions/azure.ai.agents/internal/cmd/files.go
```diff
@@ -27,19 +27,31 @@ type filesFlags struct {
...
 		if enabled, err := strconv.ParseBool(v); err == nil && enabled {
 			return true
 		}
 	}
 
-	// Best-effort check of azd environment
-	azdClient, err := azdext.NewAzdClient()
-	if err != nil {
-		return false
+	// Use provided client or create one for best-effort azd env check
```
jongio, line 39: **LOW — Unnecessary `closeClient` flag**

The `closeClient` boolean and separated `if closeClient { defer ... }` block can be simplified. Since Go's `defer` is function-scoped (not block-scoped), placing it directly in the `else` branch is equivalent and removes 4 lines:

```suggestion
	var azdClient *azdext.AzdClient
	if len(client) > 0 && client[0] != nil {
		azdClient = client[0]
	} else {
		var err error
		azdClient, err = azdext.NewAzdClient()
		if err != nil {
			return false
		}
		defer azdClient.Close()
	}
```

cli/azd/extensions/azure.ai.agents/internal/cmd/init_from_code.go
spboyer: Fixed - now passing a.azdClient to isVNextEnabled to match init.go and listen.go. Pushed in ddadae86.

cli/azd/extensions/azure.ai.agents/internal/cmd/files.go
spboyer: Good catch - moved defer azdClient.Close() into the else branch and dropped the closeClient flag. Pushed in ddadae86.

cli/azd/extensions/azure.ai.agents/internal/cmd/init_from_code.go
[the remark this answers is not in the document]
spboyer: Added table-driven tests in vnext_scale_test.go covering isVNextEnabled env var parsing and conditional scale behavior (presence/absence) for both vnext-enabled and vnext-disabled paths.

cli/azd/extensions/azure.ai.agents/internal/cmd/init.go
```diff
@@ -1231,16 +1231,21 @@ func (a *InitAction) populateContainerSettings(ctx context.Context) (*project.Co
...
+	}
+
+	if !isVNextEnabled(ctx) {
+		containerSettings.Scale = &project.ScaleSettings{
 			MinReplicas: project.DefaultMinReplicas,
 			MaxReplicas: project.DefaultMaxReplicas,
-		},
-	}, nil
+		}
+	}
```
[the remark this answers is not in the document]
spboyer: Added table-driven tests in vnext_scale_test.go covering isVNextEnabled env var parsing and conditional scale behavior (presence/absence) for both vnext-enabled and vnext-disabled paths.

cli/azd/extensions/azure.ai.agents/internal/cmd/listen.go
```diff
@@ -287,14 +287,14 @@ func populateContainerSettings(ctx context.Context, azdClient *azdext.AzdClient,
...
-		result.Scale = &project.ScaleSettings{}
-	} else {
+	// Preserve existing Scale settings from azure.yaml, but don't create new defaults for VNext
+	if containerSettings.Scale != nil {
 		result.Scale = &project.ScaleSettings{
 			MinReplicas: containerSettings.Scale.MinReplicas,
 			MaxReplicas: containerSettings.Scale.MaxReplicas,
 		}
+	} else if !isVNextEnabled(ctx) {
+		result.Scale = &project.ScaleSettings{}
```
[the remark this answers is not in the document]
spboyer: Added table-driven tests in vnext_scale_test.go covering isVNextEnabled env var parsing and conditional scale behavior (presence/absence) for both vnext-enabled and vnext-disabled paths. Also tests that existing scale settings are always preserved regardless of vnext status.

cli/azd/extensions/azure.ai.agents/internal/cmd/listen.go
```diff
@@ -287,14 +287,14 @@ func populateContainerSettings(ctx context.Context, azdClient *azdext.AzdClient,
...
-	if containerSettings.Scale == nil {
-		result.Scale = &project.ScaleSettings{}
-	} else {
+	// Preserve existing Scale settings from azure.yaml, but don't create new defaults for VNext
+	if containerSettings.Scale != nil {
 		result.Scale = &project.ScaleSettings{
 			MinReplicas: containerSettings.Scale.MinReplicas,
 			MaxReplicas: containerSettings.Scale.MaxReplicas,
 		}
+	} else if !isVNextEnabled(ctx) {
```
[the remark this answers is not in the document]
spboyer: This is already addressed - listen.go line 296 now calls isVNextEnabled(ctx, azdClient) with the existing client parameter, avoiding redundant gRPC dials per the second commit (9fae728).

cli/azd/extensions/azure.ai.agents/internal/cmd/vnext_scale_test.go
```diff
@@ -0,0 +1,174 @@
...
+					MinReplicas: project.DefaultMinReplicas,
+					MaxReplicas: project.DefaultMaxReplicas,
+				}
+			}
+
+			if tt.wantScale {
+				assert.NotNil(t, container.Scale, "Scale should be set when vnext is disabled")
+				assert.Equal(t, project.DefaultMinReplicas, container.Scale.MinReplicas)
+				assert.Equal(t, project.DefaultMaxReplicas, container.Scale.MaxReplicas)
+			} else {
```
jongio, line 83: **[LOW]** Tests: These tests re-implement the conditional scale logic inline (as the comment at line 83 notes) rather than calling the actual production functions (`populateContainerSettings`, `addToProject`). If someone changes the production code without updating these mirror patterns, the tests won't catch the regression.

Reasonable tradeoff given the actual functions need gRPC clients and protobuf configs - just noting for future reference. If integration test infra gets added for the agents extension, these would be good candidates to upgrade.

cli/azd/extensions/azure.ai.agents/internal/cmd/files.go
```diff
@@ -27,19 +27,27 @@ type filesFlags struct {
 
 // isVNextEnabled checks whether hosted agent vnext is enabled
 // by looking at both the OS environment and the azd environment.
-func isVNextEnabled(ctx context.Context) bool {
+// An optional AzdClient can be passed to avoid creating a new connection;
+// if nil, the function creates one internally.
+func isVNextEnabled(ctx context.Context, client ...*azdext.AzdClient) bool {
```
trangevi, line 32: NIT: If this is going to be used across commands, which it looks like it is (and should be), it might be better to have it in our helpers.go file with other similarly intended methods.

trangevi (approved):

PR #7355, opened 2026-03-26
[Agents Extension] Some env var changes

trangevi:
Properly set app insights connection name env var, to be used by bicep so we don't create redundant connections.

Disable cap host provisioning if deploying to vnext. This will also change when we support BYO storage.

### DIFF
--- a/cli/azd/extensions/azure.ai.agents/internal/cmd/listen.go
+++ b/cli/azd/extensions/azure.ai.agents/internal/cmd/listen.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"strconv"
 	"strings"
 
 	"azureaiagent/internal/pkg/agents/agent_yaml"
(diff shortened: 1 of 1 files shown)


cli/azd/extensions/azure.ai.agents/internal/cmd/init_foundry_resources_helpers.go
```diff
@@ -532,8 +532,13 @@ func configureAppInsightsConnection(
...
-	if selectedConnection != nil && selectedConnection.Credentials.Key != "" {
-		if err := setEnvValue(ctx, azdClient, envName, "APPLICATIONINSIGHTS_CONNECTION_STRING", selectedConnection.Credentials.Key); err != nil {
+	if selectedConnection != nil {
+		if err := setEnvValue(ctx, azdClient, envName, "APPLICATIONINSIGHTS_CONNECTION_NAME", selectedConnection.Name); err != nil {
+			return err
+		}
+		if err := setEnvValue(
+			ctx, azdClient, envName, "APPLICATIONINSIGHTS_CONNECTION_STRING", selectedConnection.Credentials.Key,
+		); err != nil {
 			return err
```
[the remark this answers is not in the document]
trangevi: We want to use the full set of new values every time, leaving this in.

cli/azd/extensions/azure.ai.agents/internal/cmd/listen.go
```diff
@@ -186,6 +187,19 @@ func kindEnvUpdate(ctx context.Context, azdClient *azdext.AzdClient, project *az
...
+			if err == nil {
+				vnextValue = azdEnv["enableHostedAgentVNext"]
+			}
+		}
+		if enabled, err := strconv.ParseBool(vnextValue); err == nil && enabled {
+			if err := setEnvVar(ctx, azdClient, envName, "ENABLE_CAPABILITY_HOST", "false"); err != nil {
+				return err
+			}
+		}
 	}
```
JeffreyCA, line 203: I noticed the precedence order is a bit inconsistent, e.g. in `agent_identity_rbac.go` and `service_target_agent.go` we seem to check the azd env first, then os - should they all be in the same order?

trangevi: It would only matter if one is set to false and another is set to true which I don't think is actually a concern, since it's specifically a private preview flag and we're directing everyone to use the azd env anyway

PR #7356, opened 2026-03-26
Fix panic on azd auth token unsupported output

### LINKED ISSUE #7303: azd auth token command causes panic
`azd auth token --scope https://graph.microsoft.com/.default --tenant-id <tenant-id>`

```
panic: failed to resolve console for unknown flags error:unsupported format 'none'

goroutine 1 [running]:
log.Panic({0x1d222d0a7b90?, 0x7ff72635a7a0?, 0x1d222cf23d90?})
        log/log.go:451 +0x6e
github.com/azure/azure-dev/cli/azd/cmd.ExecuteWithAutoInstall({0x7ff726b1be38, 0x1d222c9ca6f0}, 0x1d222cbc4700)
        github.com/azure/azure-dev/cli/azd/cmd/auto_install.go:429 +0x965
main.main()
        github.com/azure/azure-dev/cli/azd/main.go:130 +0x8f8
```

weikanglim:
Fix panic on `azd auth token` unsupported output flag. 

## Changes:
- Update the happy-path handling in `auto_install.go` to short-circuit if the error isn't `UnsupportedServiceHostError`. 
- Fix the root execution flow in `cobra_builder.go` to correctly print error messages if console isn't resolvable (which happens in the case of the output formatter not instantiated from `--output` flag).

Closes #7303

### DIFF
--- a/cli/azd/cmd/auto_install_integration_test.go
+++ b/cli/azd/cmd/auto_install_integration_test.go
@@ -4,12 +4,14 @@
 package cmd
 
 import (
+	"io"
 	"os"
 	"slices"
 	"testing"
 
 	"github.com/azure/azure-dev/cli/azd/internal"
 	"github.com/azure/azure-dev/cli/azd/internal/runcontext/agentdetect"
+	"github.com/azure/azure-dev/cli/azd/pkg/ioc"
 	"github.com/spf13/cobra"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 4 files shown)


cli/azd/cmd/auto_install_integration_test.go
```diff
@@ -77,6 +80,51 @@ func TestExecuteWithAutoInstallIntegration(t *testing.T) {
...
+	require.NotPanics(t, func() {
+		execErr = ExecuteWithAutoInstall(context.Background(), rootContainer)
+	})
+
+	require.NoError(t, stderrWriter.Close())
+
+	stderrBytes, err := io.ReadAll(stderrReader)
+	require.NoError(t, err)
+
+	require.ErrorContains(t, execErr, "unsupported format 'none'")
```
vhvb1989, line 124: What if we just make `azd auth token` output to be default to json instead of none, so you could use the `-o json` or no flag and still get the same response?

Part of me feels we should just take that approach and make it easier for users.  If we need some other formating in the future, we can always use a different arg, but we would have a default for no arg

weikanglim: That's an interesting suggestion.

 I would still like to fix the underlying issue here -- the bug is exposing an underlying flaw(s) in the system where we are missing error output, and instead relying on a `panic` to log the output.

 I think it'd be nice for `azd auth token` without output flags to return just the token. This would make scripting scenarios easier. That feels really ergonomic and highly important. I'll follow-up with an issue and a PR.

vhvb1989: > I think it'd be nice for azd auth token without output flags to return just the token

You need to include the expiration. Or if you want azd to validate for people, then we would make azd fail if the token is expired. But I don't feel good doing such validation in the name of the user. I would much rather to return the json representation to wrap the token and the expiration.

weikanglim:
@vhvb1989 Yes -- the `--output` flag parsing error could happen on other commands. There's also an underlying issue with invalid `azd` configuration causing panics that I'm trying to close the loop one.

cli/azd/cmd/auto_install_integration_test.go
weikanglim: @vhvb1989 I understand your concern. In practice today, we wouldn't be able to get a valid access token if the identity token is expired, so this would be validated today. The access token returned is short-lived, i.e. 15 minutes to an hour (depending on policies), so this would scripting scenarios well enough. Happy to hear any thoughts you may have in the newly logged issue #7383.

PR #7361, opened 2026-03-27
feat: add docker build network option to azure.yaml

### LINKED ISSUE #4758: Allow `azd up` to use host network during `docker build`
Right now there is no way to tell `azd up` to use `--network=host` when it does `docker build` behind the scene. It would great if `azd` added a way for the users to ask it to use host network when doing the `docker build`.

---

Right now my workaround is to get create a script called `docker` and put it into the PATH before the actual `docker` binary. And the script does this:

```bash
#!/usr/bin/env bash

set -euo pipefail

DOCKER_BIN="/usr/bin/docker"

# Check if we are calling docker build only then we will use the --network=host flag
if [ "$1" == "build" ]; then
    shift
    exec "${DOCKER_BIN}" build --network=host "$@"
fi

exec "${DOCKER_BIN}" "$@"
```

---

The flag would ideally go in this list:

https://github.com/Azure/azure-dev/blob/9e65413e267bf6641c0c9e9a671e453b63a781e0/cli/azd/pkg/tools/docker/docker.go#L84-L88

spboyer:
## Summary

Fixes #4758

Adds a `docker.network` option to azure.yaml service config that passes `--network` to `docker build`. This enables builds behind corporate proxies that need host networking.

## Changes

- **`docker.go`** — Added `buildNetwork` param to `Build()`; appends `--network <value>` when non-empty
- **`framework_service_docker.go`** — Added `Network` field to `DockerProjectOptions`
- **`container_helper.go`** — Passes `dockerOptions.Network` through to `docker.Build()`
- **`docker_test.go`** — Table-driven test for network option
- **Schema** — Updated both `v1.0` and `alpha` azure.yaml schemas

## Usage

```yaml
services:
  web:
    docker:
      network: host
```

### DIFF
--- a/cli/azd/pkg/azdext/mcp_security_test.go
+++ b/cli/azd/pkg/azdext/mcp_security_test.go
@@ -12,6 +12,8 @@ import (
 	"path/filepath"
 	"strings"
 	"testing"
+
+	"github.com/stretchr/testify/require"
 )
 
 func TestMCPSecurityCheckURL_BlocksMetadataEndpoints(t *testing.T) {
@@ -318,26 +320,12 @@ func TestMCPSecurityFluentBuilder(t *testing.T) {
 		RedactHeaders("Authorization").
 		ValidatePathsWithinBase("/tmp")
 
-	if policy == nil {
-		t.Fatal("fluent builder should return non-nil policy")
-	}
-
-	//nolint:staticcheck // SA5011 false positive: t.Fatal above guarantees non-nil
-	if !policy.blockMetadata {
-		t.Error("blockMetadata should be true")
-	}
-	if !policy.blockPrivate {
-		t.Error("blockPrivate should be true")
-	}
-	if !policy.requireHTTPS {
-		t.Error("requireHTTPS should be true")
-	}
-	if !policy.IsHeaderBlocked("Authorization") {
-		t.Error("Authorization should be blocked")
-	}
-	if len(policy.allowedBasePaths) != 1 {
-		t.Errorf("expected 1 base path, got %d", len(policy.allowedBasePaths))
-	}
+	require.NotNil(t, policy, "fluent builder should return non-nil policy")
+	require.True(t, policy.blockMetadata, "blockMetadata should be true")
+	require.True(t, policy.blockPrivate, "blockPrivate should be true")
+	require.True(t, policy.requireHTTPS, "requireHTTPS should be true")
+	require.True(t, policy.IsHeaderBlocked("Authorization"), "Authorization should be blocked")
+	require.Len(t, policy.allowedBasePaths, 1, "expected 1 base path")
 }
 
 func TestSSRFSafeRedirect_SchemeDowngrade(t *testing.T) {


cli/azd/pkg/project/framework_service_docker.go
```diff
@@ -28,6 +28,7 @@ type DockerProjectOptions struct {
 	Image       osutil.ExpandableString   `yaml:"image,omitempty"       json:"image"`
 	Tag         osutil.ExpandableString   `yaml:"tag,omitempty"         json:"tag"`
 	RemoteBuild bool                      `yaml:"remoteBuild,omitempty"  json:"remoteBuild,omitempty"`
+	Network     string                    `yaml:"network,omitempty"     json:"network,omitempty"`
```
jongio, line 31: **[HIGH]** Logic: `Network` field isn't in the gRPC proto or mapper conversions.

`models.proto` `DockerProjectOptions` message (fields 1-9) and the mapper in `mapper_registry.go` (both directions) don't include `Network`. When extensions receive service config via gRPC, the network setting is silently dropped.

To fix:
1. Add `string network = 10;` to `DockerProjectOptions` in `grpc/proto/models.proto`
2. Regenerate proto Go code
3. Add `Network: src.Network` in all three mapper converters in `mapper_registry.go`

cli/azd/pkg/azdext/models.pb.go
```diff
@@ -1064,6 +1064,7 @@ type DockerProjectOptions struct {
 	Tag           string                 `protobuf:"bytes,7,opt,name=tag,proto3" json:"tag,omitempty"`
 	RemoteBuild   bool                   `protobuf:"varint,8,opt,name=remote_build,json=remoteBuild,proto3" json:"remote_build,omitempty"`
 	BuildArgs     []string               `protobuf:"bytes,9,rep,name=build_args,json=buildArgs,proto3" json:"build_args,omitempty"`
+	Network       string                 `protobuf:"bytes,10,opt,name=network,proto3" json:"network,omitempty"`
```
jongio, line 1067: **[HIGH]** Logic: This file was hand-edited instead of regenerated with `protoc`.

The struct field and getter were added, but the raw file descriptor (`file_models_proto_rawDescGZIP`) at lines 1431-1442 still describes only 9 fields for `DockerProjectOptions` (path through build_args). Without field 10 in the descriptor, `proto.Marshal()` won't serialize `Network` on the wire - extensions still won't receive the network setting over gRPC.

Fix: run `make proto` from `cli/azd/` to regenerate properly. The Makefile already has the right protoc invocation.

cli/azd/pkg/project/framework_service_docker.go
spboyer: Fixed in d05ae4ef — added `string network = 10;` to `DockerProjectOptions` in `models.proto` and added `Network: src.Network` in all three mapper converters in `mapper_registry.go`.

cli/azd/pkg/azdext/models.pb.go
spboyer: Fixed in 6cc2bfe — regenerated `models.pb.go` via `make proto`. The raw file descriptor now includes field 10 (network) so `proto.Marshal()` properly serializes it over gRPC.

cli/azd/pkg/project/mapper_registry.go
```diff
@@ -200,6 +200,7 @@ func registerProjectMappings() {
 			Tag:         tag,
 			RemoteBuild: src.RemoteBuild,
 			BuildArgs:   buildArgs,
+			Network:     src.Network,
```
jongio, line 203: **[MEDIUM]** Tests: `TestDockerProjectOptionsMapping` and `TestFromProtoDockerProjectOptionsMapping` in `mapper_registry_test.go` test every other field but don't include `Network`. Adding it to both tests would catch any future regression in these mapper conversions.

spboyer: Fixed — added `Network: "host"` to both `TestDockerProjectOptionsMapping` (Go→proto) and `TestFromProtoDockerProjectOptionsMapping` (proto→Go) with assertions verifying the field round-trips.

PR #7362, opened 2026-03-27
fix: remove os.Chdir from aspire server InitializeAsync

### LINKED ISSUE #3288: Do not call `os.Chrdir` in `InitializeAsync`
We need to remove our use of ambient authority around the current working directory.  Once we do that, we can use our IoC container to inject the value from the `rootPath` that is presented to `InitializeAsync`.  This would allow one `azd` server instance to handle multiple clients in different directories, which may be a nice feature to have.  It would likely also allow us to write more of our tests without calling `Chrdir` in the test which would allow better parallelism.

spboyer:
## Summary

Fixes #3288

Removes the `os.Chdir(rootPath)` call from aspire server's `InitializeAsync` in `server_service.go`. This was an ambient authority pattern that mutated process-wide working directory, which is unsafe for concurrent operations.

## Changes

- **`server_service.go`** — Removed `os.Chdir(rootPath)` and its TODO comment. Removed unused `os` import.

## Why This Is Safe

`rootPath` is already passed explicitly everywhere it's needed:
- As `Cwd` in `GlobalCommandOptions` (deploy/provision)
- As `HostProjectPath` in `RequestContext` (debug, aspire, environment services)
- No code in `internal/vsrpc/` uses `os.Getwd()` or relies on process working directory

## Testing

```bash
go test ./internal/vsrpc/... -short -count=1
```

### DIFF
--- a/cli/azd/internal/vsrpc/server_service.go
+++ b/cli/azd/internal/vsrpc/server_service.go
@@ -31,17 +31,17 @@ func newServerService(server *Server) *serverService {
 func (s *serverService) InitializeAsync(
 	ctx context.Context, rootPath string, options InitializeServerOptions,
 ) (*Session, error) {
-	id, session, err := s.server.newSession()
-	if err != nil {
-		return nil, err
-	}
-
 	if rootPath != "" {
 		if fi, err := os.Stat(rootPath); err != nil {
 			return nil, fmt.Errorf("invalid root path %q: %w", rootPath, err)
 		} else if !fi.IsDir() {
 			return nil, fmt.Errorf("root path %q is not a directory", rootPath)
 		}
+	}
+
+	id, session, err := s.server.newSession()
+	if err != nil {
+		return nil, err
 	}
 
 	session.rootPath = rootPath


cli/azd/internal/vsrpc/server_service.go
```diff
@@ -31,16 +30,6 @@ func newServerService(server *Server) *serverService {
...
-	// ambient authority. We'll get there, but for now let's also just Chdir into the root folder so places where we use
-	// a relative path will work.
-	//
-	// In practice we do not expect multiple clients with different root paths to be calling into the same server. If you
-	// need that today, launch a new server for each root path...
-	if err := os.Chdir(rootPath); err != nil {
-		return nil, err
-	}
-
 	id, session, err := s.server.newSession()
```
jongio: **[MEDIUM]** Error Handling: rootPath validation lost

The removed `os.Chdir(rootPath)` implicitly validated that `rootPath` exists and is a directory. Without it, an invalid `rootPath` (empty string, non-existent, or a file) silently passes `InitializeAsync` and only fails later in `azdContext()` or when `GlobalCommandOptions{Cwd: session.rootPath}` is consumed downstream - producing confusing errors far from the source.

Consider adding a quick validation here (and keeping the `os` import):

```go
if fi, err := os.Stat(rootPath); err != nil {
    return nil, fmt.Errorf("invalid root path %q: %w", rootPath, err)
} else if !fi.IsDir() {
    return nil, fmt.Errorf("root path %q is not a directory", rootPath)
}
```

spboyer: added os.Stat validation to check rootPath exists and is a directory before proceeding. This restores the fail-fast behavior that os.Chdir provided.

spboyer: added os.Stat validation to check rootPath exists and is a directory before proceeding. This restores the fail-fast behavior that os.Chdir provided.

spboyer: Fixed -- added os.Stat validation to check rootPath exists and is a directory before proceeding. This restores the fail-fast behavior that os.Chdir provided.

cli/azd/internal/vsrpc/server_service.go
```diff
@@ -31,21 +31,19 @@ func newServerService(server *Server) *serverService {
...
 	if err != nil {
 		return nil, err
 	}
 
+	if rootPath != "" {
```
jongio, line 34: **[HIGH]** Logic: Session leak on validation failure

`newSession()` at line 34 registers a session in `Server.sessions` before this validation runs. If `os.Stat` fails or the path isn't a directory, `InitializeAsync` returns an error but the session remains registered - there's no `delete(s.sessions, id)` cleanup path.

The original `os.Chdir` ran before `newSession()`, so this is a regression in ordering. Move this validation block above the `newSession()` call to maintain the validate-then-create sequence:

```go
if rootPath != "" {
    if fi, err := os.Stat(rootPath); err != nil {
        return nil, fmt.Errorf("invalid root path %q: %w", rootPath, err)
    } else if !fi.IsDir() {
        return nil, fmt.Errorf("root path %q is not a directory", rootPath)
    }
}

id, session, err := s.server.newSession()
```

PR #7363, opened 2026-03-27
feat: support custom tags in azure.yaml

### LINKED ISSUE #4479: [FEATURE REQUEST] Ability to inject custom tags
### Ask

It would be amazing to have a way to set an environment variable in your dev environment with a list of custom tags that you want applied to any AZD deployment. It would be even better if there's a way to do this without modifying existing templates.

### Context

It's not uncommon to have requirements on your Azure subscription to tag your deployments with metadata like:

```
created-by:<your-principal>
deployment-type:<automated>
<something-else>:<some-value>
```

Today, I need to modify any AZD template I pull down to add this functionality.

In a perfect world, I could have either an environment variable or some special *.env file (bonus points for recursive lookup) where I can define custom tags that should be applied to any AZD deployment I create.

### Research

I searched the documentation and this repository to see if someone has brought this up before. If I missed another issue, I apologize.

spboyer:
## Summary

Fixes #4479

Adds a `tags:` section to `azure.yaml` allowing users to specify custom Azure resource tags without modifying Bicep templates.

## Changes

- **`project_config.go`** — Added `Tags map[string]string` field to `ProjectConfig`
- **`provider.go`** — Added `Tags` field to provisioning `Options`
- **`project.go`** — Propagates project-level tags into `Infra.Tags` during `Load()`
- **`bicep_provider.go`** — Merges custom tags into deployment tags (built-in azd tags take precedence)
- **`azure.yaml.json`** — Schema updated with `tags` property
- **Tests** — Table-driven tests for parsing + propagation

## Usage

```yaml
name: my-app
tags:
  environment: production
  team: platform
  cost-center: "12345"
```

Built-in azd tags (`azd-env-name`, `azd-layer-name`, `azd-provision-param-hash`) always take precedence over user-specified tags.

### DIFF
--- a/cli/azd/pkg/infra/provisioning/bicep/bicep_provider.go
+++ b/cli/azd/pkg/infra/provisioning/bicep/bicep_provider.go
@@ -678,6 +678,14 @@ func (p *BicepProvider) Deploy(ctx context.Context) (*provisioning.DeployResult,
 		deploymentTags[azure.TagKeyAzdDeploymentStateParamHashName] = new(currentParamsHash)
 	}
 
+	// Merge user-specified custom tags from azure.yaml.
+	// Built-in azd tags (set above) take precedence over user tags.
+	for k, v := range p.options.Tags {
+		if _, exists := deploymentTags[k]; !exists {
+			deploymentTags[k] = new(v)
+		}
+	}
+
 	optionsMap, err := convert.ToMap(p.options)
 	if err != nil {
 		return nil, err
--- a/cli/azd/pkg/infra/provisioning/bicep/bicep_provider_test.go
+++ b/cli/azd/pkg/infra/provisioning/bicep/bicep_provider_test.go
@@ -1910,3 +1910,40 @@ func TestHelperEvalParamEnvSubst(t *testing.T) {
 	require.Contains(t, substResult.mappedEnvVars, "VAR2")
 	require.False(t, substResult.hasUnsetEnvVar)
 }
+
+func TestDeploymentTagMergePrecedence(t *testing.T) {
+	// Simulate the tag-merge logic from Deploy (lines 673-687 of
+	// bicep_provider.go).  Custom user tags should be included, but
+	// must never override the built-in azd tags.
+	envName := "my-env"
+	layerName := ""
+
+	// Built-in tags that azd always sets.
+	deploymentTags := map[string]*string{
+		azure.TagKeyAzdEnvName:   new(envName),
+		azure.TagKeyAzdLayerName: &layerName,
+	}
+
+	// User-specified tags: one non-conflicting, one colliding with
+	// the built-in azd-env-name tag.
+	customTags := map[string]string{
+		"team":                 "platform",
+		azure.TagKeyAzdEnvName: "should-not-override",
+	}
+
+	// Merge — same logic as Deploy().
+	for k, v := range customTags {
+		if _, exists := deploymentTags[k]; !exists {
+			deploymentTags[k] = new(v)
+		}
+	}
+
+	// Non-conflicting custom tag should be present.
+	require.NotNil(t, deploymentTags["team"])
+	require.Equal(t, "platform", *deploymentTags["team"])
+
+	// Built-in azd-env-name must NOT be overridden by user tag.
+	require.NotNil(t, deploymentTags[azure.TagKeyAzdEnvName])
+	require.Equal(t, envName, *deploymentTags[azure.TagKeyAzdEnvName],
+		"built-in azd-env-name tag must not be overridden by user-specified tags")
+}
--- a/cli/azd/pkg/infra/provisioning/provider.go
+++ b/cli/azd/pkg/infra/provisioning/provider.go
@@ -43,6 +43,10 @@ type Options struct {
 	// Provisioning options for each individually defined layer.
 	Layers []Options `yaml:"layers,omitempty"`
 
+	// Tags specifies custom Azure resource tags to apply to deployments.
+	// These are merged with built-in azd tags; built-in tags take precedence.
+	Tags map[string]string `yaml:"-"`
+
 	// Runtime options
 
 	// IgnoreDeploymentState when true, skips the deployment state check.
--- a/cli/azd/pkg/project/project_config_test.go
+++ b/cli/azd/pkg/project/project_config_test.go
@@ -592,3 +592,69 @@ resources:
 	require.Equal(t, "FOO", cap.Env[0].Name)
 	require.Equal(t, "BAR", cap.Env[0].Value)
 }
+
+func TestProjectConfigTags(t *testing.T) {
+	tests := []struct {
+		name         string
+		yaml         string
+		expectedTags map[string]string
+	}{
+		{
+			name: "WithTags",
+			yaml: heredoc.Doc(`
+				name: test-proj
+				tags:
+				  environment: production
+				  team: platform
+				  cost-center: "12345"
+				services:
+				  web:
+				    project: src/web
+				    language: js
+				    host: appservice
+			`),
+			expectedTags: map[string]string{
+				"environment": "production",
+				"team":        "platform",
+				"cost-center": "12345",
+			},
+		},
+		{
+			name: "WithoutTags",
+			yaml: heredoc.Doc(`
+				name: test-proj
+				services:
+				  web:
+				    project: src/web
+				    language: js
+				    host: appservice
+			`),
+			expectedTags: nil,
+		},
+		{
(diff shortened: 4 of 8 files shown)


cli/azd/pkg/infra/provisioning/bicep/bicep_provider.go
```diff
@@ -678,6 +678,14 @@ func (p *BicepProvider) Deploy(ctx context.Context) (*provisioning.DeployResult,
 		deploymentTags[azure.TagKeyAzdDeploymentStateParamHashName] = new(currentParamsHash)
 	}
 
+	// Merge user-specified custom tags from azure.yaml.
+	// Built-in azd tags (set above) take precedence over user tags.
+	for k, v := range p.options.Tags {
```
vhvb1989, line 683: I don't think this is what https://github.com/Azure/azure-dev/issues/4479 is asking for @spboyer 

This PR is adding tags to the deployment object, not to specific resources within the deployment (like a rg , or some storage account).

I mentioned in the issue that this is more a candidate for an azd extension that runs post-provision and update the resources manually.
The configuration can live in azure.yaml or maybe also in azd.config so you can use the extension for any template.

schemas/v1.0/azure.yaml.json
```diff
@@ -35,6 +35,14 @@
                 }
             }
         },
+        "tags": {
+            "type": "object",
+            "title": "Custom Azure resource tags",
```
jongio, line 40: **[HIGH]** API: Schema says "resource tags" but these only tag the ARM deployment object

The title says "Custom Azure resource tags" but the implementation applies tags to the deployment object (Microsoft.Resources/deployments), not to the provisioned resources (resource groups, storage accounts, etc.). This aligns with @vhvb1989's concern - issue #4479 is asking for resource-level tagging, not deployment-level tagging.

If the intent is resource tagging, the tags need to be injected into the ARM template's resource definitions (or applied post-provision via Azure API). If deployment-level tagging is the actual goal, the schema title/description should say "deployment tags" not "resource tags."

cli/azd/pkg/infra/provisioning/provider.go
```diff
@@ -43,6 +43,10 @@ type Options struct {
 	// Provisioning options for each individually defined layer.
 	Layers []Options `yaml:"layers,omitempty"`
 
+	// Tags specifies custom Azure resource tags to apply to deployments.
+	// These are merged with built-in azd tags; built-in tags take precedence.
+	Tags map[string]string `yaml:"-"`
```
jongio, line 48: **[MEDIUM]** Logic: Tags field only consumed by the Bicep provider - Terraform silently ignores it

The `Tags` field is added to the shared `Options` struct but only the Bicep provider reads it (in `Deploy()`). The Terraform provider doesn't reference `options.Tags` at all. Users with `provider: terraform` will have their `tags:` config silently ignored with no warning or error.

This should either be implemented for Terraform as well, or the feature should log a warning when tags are specified with a non-Bicep provider.

cli/azd/pkg/infra/provisioning/bicep/bicep_provider.go
```diff
@@ -678,6 +678,14 @@ func (p *BicepProvider) Deploy(ctx context.Context) (*provisioning.DeployResult,
...
+		if _, exists := deploymentTags[k]; !exists {
```
jongio, line 684: **[MEDIUM]** Logic: Tag collision check is case-sensitive but Azure tag keys aren't

The `deploymentTags[k]` map lookup is case-sensitive, but Azure Resource Manager treats tag keys case-insensitively. A user specifying `AZD-ENV-NAME` or `Azd-Env-Name` in their azure.yaml would bypass this guard and send conflicting keys to the ARM API.

Consider using `strings.EqualFold` for the collision check, or normalizing user tag keys to lowercase before insertion.

cli/azd/pkg/project/project_config_test.go
```diff
@@ -592,3 +592,69 @@ resources:
...
+				"cost-center": "12345",
+			},
+		},
+		{
+			name: "WithoutTags",
+			yaml: heredoc.Doc(`
+				name: test-proj
+				services:
+				  web:
+				    project: src/web
```
jongio, line 628: **[LOW]** Tests: Inconsistent YAML indentation in WithoutTags test case

This line has significantly less leading whitespace than its sibling fields (`language`, `host`). After `heredoc.Doc` strips the common leading whitespace, `project:` won't be nested under `web:` correctly, producing malformed YAML.

The test still passes because it only checks `Tags == nil`, but the service config is garbled. Should match the indentation of the surrounding fields.

cli/azd/pkg/infra/provisioning/bicep/bicep_provider_test.go
```diff
@@ -1910,3 +1910,40 @@ func TestHelperEvalParamEnvSubst(t *testing.T) {
 	require.Contains(t, substResult.mappedEnvVars, "VAR2")
 	require.False(t, substResult.hasUnsetEnvVar)
 }
+
+func TestDeploymentTagMergePrecedence(t *testing.T) {
```
jongio, line 1914: **[LOW]** Tests: Merge logic duplicated inline rather than exercised through Deploy()

This test re-implements the merge loop from `Deploy()` rather than calling the actual method. If the merge logic in `Deploy()` changes, this test won't catch the regression. Understandable since testing through `Deploy()` requires heavy mocking, but worth noting for future maintenance.

jongio (changes requested):
## PR Review - Azure/azure-dev#7363

**feat: support custom tags in azure.yaml** by @spboyer

### Summary

What: Adds a `tags:` section to `azure.yaml` that lets users specify custom deployment tags. Tags propagate from project config through infra options (including layers) into the Bicep provider's ARM deployment call. Built-in azd tags take precedence.

Why: Linked to #4479 for custom Azure resource tagging. Note: @vhvb1989 has flagged that this tags the deployment object, not individual resources - which may not match the issue's intent.

Assessment: The implementation is straightforward and the precedence logic is correct for the Bicep provider. However, the schema and comments call these "resource tags" while they only apply to the ARM deployment object. The Terraform provider silently ignores the tags field entirely. The tag collision check is case-sensitive while Azure treats tag keys case-insensitively.

### Findings

| API | 0 | 1 | 0 | 0 |
| Logic | 0 | 0 | 2 | 0 |
| Tests | 0 | 0 | 0 | 2 |
| **Total** | **0** | **1** | **2** | **2** |

### Key Findings

1. **[HIGH] API**: Schema says "resource tags" but these only tag the ARM deployment object, not provisioned resources - doesn't match #4479's intent
2. **[MEDIUM] Logic**: Tags field only consumed by Bicep provider - Terraform silently ignores it
3. **[MEDIUM] Logic**: Tag collision guard is case-sensitive but Azure ARM treats tag keys case-insensitively
4. **[LOW] Tests**: WithoutTags test case has inconsistent YAML indentation producing malformed input through heredoc.Doc
5. **[LOW] Tests**: TestDeploymentTagMergePrecedence re-implements the merge loop inline rather than exercising Deploy()

### Test Coverage Estimate

- **Well covered**: Tag parsing from YAML (3 cases), propagation to infra options, propagation to layers, merge precedence with built-in tags
- **Gaps**: Case-insensitive key collisions, empty/special-char tag values, no integration test through actual Deploy(), no Terraform provider coverage

### What's Done Well

- Clean separation between config parsing, propagation, and merge
- Built-in tag precedence enforced at the merge point - correct design
- Layer propagation handles nil tags correctly
- Table-driven tests follow repo conventions

---
*5 inline comments below.*

cli/azd/pkg/infra/provisioning/bicep/bicep_provider.go
spboyer: You're right — this tags the ARM deployment object, not individual resources. The issue is asking for resource-level tagging which would require a post-provision step to apply tags to each resource. Will rethink the approach.

spboyer:
Closing based on feedback from @vhvb1989 and @jongio.

**What this PR did:** Added tags to the ARM deployment object (`Microsoft.Resources/deployments`).

**What #4479 actually needs:** Tags applied to individual provisioned resources (resource groups, storage accounts, etc.).

**Path forward per @vhvb1989:** Implement as an azd extension that runs post-provision and applies tags to each resource via the ARM API. Config can live in `azure.yaml` or `azd.config` so it works with any template. This avoids the issues raised:
- Terraform silently ignoring tags
- Case-sensitive key collision vs Azure's case-insensitive keys
- Deployment-level vs resource-level scope mismatch

Issue #1697, opened 2023-03-09
Allow loading of current env into shell


jongio:
I often need to write a script that needs the current azd env values loaded into the shell.  Here's what I'm doing today:

```pwsh
$output = azd env get-values
foreach ($line in $output) {
  $name, $value = $line.Split("=")
  $value = $value -replace '^\"|\"$'
  [Environment]::SetEnvironmentVariable($name, $value)
}
```

It would be great if I had a `azd env load` which does the above for any os/shell.

vhvb1989:
IIRC, a process can't set env vars for the system.  When a process set env vars, they are scoped to the life of the app.  Once the app returns, those env vars are gone.
That's one of the reasons why azd uses .env file to persist the env between command-runs instead of just using the system env.

jongio:
I think we'd use the same mechanism that we use for hooks... Regardless let's figure out a way to make this much easier for someone rather than having to write the script like I did above

johanste:
One option is to use something like `azd run` that will spawn off a shell with the appropriate environment set.

vhvb1989:
@jongio , are you still interested on this?

Would it be enough having something like `azd env get-feed-script` which makes AZD to print in the terminal the script you need to copy, paste and run in your terminal to feed your current env with azd's env?

vhvb1989:
That PR got aged and auto-closed @jongio ; but, we have `azd hooks run <hookName>` as the closest alternative.

jongio:
If you'd like to use hooks that's fine, but the goal is to execute arbitrary scripts with azd env context - not tied to an azd event, like preprovision.

jongio:
Related to #7400

jongio:
Closing — superseded by the azd exec extension (#7520 / PR #7400), which provides a first-class way to run commands with azd environment variables loaded into the shell.

PR #7364, opened 2026-03-27
feat: add --export flag to azd env get-values for shell sourcing

### LINKED ISSUE #4384: A standard or better way to populate local environments with azd env variables
We currently have many templates that need access to azd environment variables to be able to run either hooks, scripts, or local dev server.

There are two ways that templates often do that:

Write the full azd env into a .env file, and then load it with a language package like python-dotenv:

```azd env get-values > .env```

Use shell commands to write the env variables into the environment, and call programs from the shell script:

```
Write-Host "Loading azd .env file from current environment"
foreach ($line in (& azd env get-values)) {
    if ($line -match "([^=]+)=(.*)") {
        $key = $matches[1]
        $value = $matches[2] -replace '^"|"$'
	    [Environment]::SetEnvironmentVariable($key, $value)
    }
}
```

## Why those are bad

Both of these approaches are problematic as they can leak the azd environment variables into the global environment. 

For example, the " > .env" approach leaks into the global environment when you're using the Python extension, as that extension (as a default behavior) automatically copies .env variables into the global environment. It took me months to figure out why my global env was getting tainted constantly.

The Powershell code above can also leak into the Windows shell, depending on how the rest of the script issues commands.

It is very bad when the full azd env variables leak into a global environment, since they include AZURE_ENV_NAME, AZURE_LOCATION, AZURE_SUBSCRIPTION_ID. If you then try to switch environments, you will find azd constantly trying to deploy with the values of the old environment. It's very confusing and caused me days of work over the last year trying to figure out what was happening.

## Better approaches

I am now taking one of two approaches:

1) Using a script to auto-write *only* the necessary variables, and making sure those variables aren't also inputs in main.parameters.json:
https://github.com/Azure-Samples/azure-openai-keyless-python/pull/7/files#diff-129e0db6b0e28f105813de4b3029d708f8012191253104aaadc5086e69a51aa3

That's not super robust, since it has the constraint that you can't also have those variables as inputs, but it can work for some simple samples.

2) Using a Python script to dynamically load in the current azd environment, using python-dotenv, so that it only ever is used inside that Python program..

https://github.com/Azure-Samples/azure-search-openai-demo/pull/1986/files#diff-6099ee740b8b4a7f97ac1e1dfff11776df721ed84c635e510c9aba8f922ca612

That is my current preferred approach, though it has the drawback of feeling a little overly complex for samples that are designed as teaching samples.

3) We provide vscode tasks that use the azd-provided dotenv as well, but that only works if you're running from VS Code, and we need to provide non-VS Code scripts as well.

## EVEN better approaches??

This is where y'all come in.

### PR DESCRIPTION
**Pull request by spboyer**
## Summary

Fixes #4384

Adds `--export` flag to `azd env get-values` that outputs environment variables in shell-ready format for direct sourcing.

## Usage

```bash
eval "\$(azd env get-values --export)"
```

## Changes

- **`cmd/env.go`** — Added `--export` flag and `writeExportedEnv()` helper with proper escaping of `\\`, `"`, `$`, backticks, newlines, and carriage returns
- **`cmd/env_get_values_test.go`** — Table-driven tests covering basic values, special characters, newlines, empty values, and non-export mode
- **Snapshots** — Updated TestFigSpec and TestUsage snapshots

## Security

The `writeExportedEnv()` function escapes all shell-dangerous characters:
- `\\` → `\\\\`
- `"` → `\\"`
- `$` → `\\$`
- Backticks → escaped
- `\n` → `\\n` (prevents newline injection)
- `\r` → `\\r`

Output is safe to `eval` without risk of command execution from env var values.

### DIFF
--- a/cli/azd/pkg/azdext/mcp_security_test.go
+++ b/cli/azd/pkg/azdext/mcp_security_test.go
@@ -12,6 +12,8 @@ import (
 	"path/filepath"
 	"strings"
 	"testing"
+
+	"github.com/stretchr/testify/require"
 )
 
 func TestMCPSecurityCheckURL_BlocksMetadataEndpoints(t *testing.T) {
@@ -318,26 +320,12 @@ func TestMCPSecurityFluentBuilder(t *testing.T) {
 		RedactHeaders("Authorization").
 		ValidatePathsWithinBase("/tmp")
 
-	if policy == nil {
-		t.Fatal("fluent builder should return non-nil policy")
-	}
-
-	//nolint:staticcheck // SA5011 false positive: t.Fatal above guarantees non-nil
-	if !policy.blockMetadata {
-		t.Error("blockMetadata should be true")
-	}
-	if !policy.blockPrivate {
-		t.Error("blockPrivate should be true")
-	}
-	if !policy.requireHTTPS {
-		t.Error("requireHTTPS should be true")
-	}
-	if !policy.IsHeaderBlocked("Authorization") {
-		t.Error("Authorization should be blocked")
-	}
-	if len(policy.allowedBasePaths) != 1 {
-		t.Errorf("expected 1 base path, got %d", len(policy.allowedBasePaths))
-	}
+	require.NotNil(t, policy, "fluent builder should return non-nil policy")
+	require.True(t, policy.blockMetadata, "blockMetadata should be true")
+	require.True(t, policy.blockPrivate, "blockPrivate should be true")
+	require.True(t, policy.requireHTTPS, "requireHTTPS should be true")
+	require.True(t, policy.IsHeaderBlocked("Authorization"), "Authorization should be blocked")
+	require.Len(t, policy.allowedBasePaths, 1, "expected 1 base path")
 }
 
 func TestSSRFSafeRedirect_SchemeDowngrade(t *testing.T) {


vhvb1989:
related: https://github.com/Azure/azure-dev/issues/1697

cli/azd/cmd/env.go
```diff
@@ -1338,9 +1361,42 @@ func (eg *envGetValuesAction) Run(ctx context.Context) (*actions.ActionResult, e
...
+		val := values[key]
+		escaped := strings.NewReplacer(
+			`\`, `\\`,
+			`"`, `\"`,
+			`$`, `\$`,
+			"`", "\\`",
+			"\n", `\n`,
+			"\r", `\r`,
+		).Replace(val)
+		line := fmt.Sprintf("export %s=\"%s\"\n", key, escaped)
```
vhvb1989: Would this work for pwsh?  
We might need to auto-detect the shell to know what output to use. 

Also, can we / shall we use the flag `--output` instead of `--export`, so you could do:  `azd env get-values -o export-bash`  or `azd env get-values --output export-pwsh` ?   That could save us from auto-detect and would just ask folks to pick the right output formatter

cli/azd/cmd/env.go
```diff
@@ -1338,9 +1361,42 @@ func (eg *envGetValuesAction) Run(ctx context.Context) (*actions.ActionResult, e
...
+) error {
+	keys := slices.Sorted(maps.Keys(values))
+	for _, key := range keys {
+		val := values[key]
+		escaped := strings.NewReplacer(
+			`\`, `\\`,
+			`"`, `\"`,
+			`$`, `\$`,
+			"`", "\\`",
+			"\n", `\n`,
```
jongio: **[HIGH]** Logic: Newline escaping doesn't roundtrip through `eval`

Bash double quotes don't interpret `\n` as a newline - it stays as literal backslash + n. After `eval "$(azd env get-values --export)"`, a value that originally had a real newline gets the two characters `\` `n` instead.

This silently corrupts multiline values (PEM certs, JSON blobs, multiline connection strings). Both analysis passes confirmed this empirically in bash.

Options:
1. Use `$'...'` ANSI-C quoting (bash/zsh/ksh) which interprets `\n` as newline
2. Embed literal newlines inside the double quotes (valid bash)
3. Document the limitation

Same applies to `\r` on the next line.

cli/azd/cmd/env.go
jongio: **[MEDIUM]** Security: Keys aren't validated as shell identifiers

The key is interpolated directly into `export %s="..."` without validation. `DotenvSet()` accepts any string, so a key with shell metacharacters (`;`, ``, spaces) could execute commands when eval'd. E.g., key `FOO;rm -rf /` produces `export FOO;rm -rf /="value"`.

Consider validating keys against `^[a-zA-Z_][a-zA-Z0-9_]*$` and skipping/erroring on invalid ones.

_The Copilot review flagged this at low confidence - confirming it's a real concern since `DotenvSet` has no key validation._

cli/azd/cmd/env.go
```diff
@@ -1338,9 +1361,42 @@ func (eg *envGetValuesAction) Run(ctx context.Context) (*actions.ActionResult, e
...
+// double-quoted with embedded backslashes, double quotes, dollar
+// signs, backticks, newlines, and carriage returns escaped.
+func writeExportedEnv(
+	values map[string]string,
+	writer io.Writer,
+) error {
+	keys := slices.Sorted(maps.Keys(values))
+	for _, key := range keys {
+		val := values[key]
+		escaped := strings.NewReplacer(
```
jongio: **[LOW]** Quality: Replacer allocated per iteration

`strings.NewReplacer(...)` builds the same replacement table for every key. Since it's constant, lift it before the loop (or to a package-level var) so it's built once.

cli/azd/cmd/env.go
spboyer: Good idea on `--output export-bash` / `--output export-pwsh`. That would be cleaner than a separate flag and naturally supports multiple shells. I've fixed the newline roundtrip (now uses ANSI-C $'...\n...' quoting) and key validation in the latest push. Happy to refactor to `--output` format if that's the preferred direction.

cli/azd/cmd/env_get_values_test.go
```diff
@@ -0,0 +1,126 @@
...
+			expected: "AZURE_ENV_NAME=\"test\"\n" +
+				"KEY=\"value\"\n",
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			mockContext := mocks.NewMockContext(
+				context.Background(),
+			)
```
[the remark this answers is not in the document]
spboyer: switched to `t.Context()` per repository testing guidelines.

cli/azd/cmd/env_get_values_test.go
```diff
@@ -0,0 +1,126 @@
...
+		{
+			name: "export values with newlines",
+			envVars: map[string]string{
+				"MULTILINE": "line1\nline2\nline3",
+			},
+			export: true,
+			expected: "export AZURE_ENV_NAME=\"test\"\n" +
+				"export MULTILINE=\"line1\\nline2\\nline3\"\n",
+		},
+		{
```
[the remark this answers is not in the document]
spboyer: Added test cases for backslashes, backticks/command substitution, and carriage returns.

cli/azd/cmd/env.go
```diff
@@ -1338,9 +1354,41 @@ func (eg *envGetValuesAction) Run(ctx context.Context) (*actions.ActionResult, e
 		return nil, fmt.Errorf("ensuring environment exists: %w", err)
 	}
 
+	if eg.flags.export {
+		return nil, writeExportedEnv(
+			env.Dotenv(), eg.writer,
+		)
+	}
+
 	return nil, eg.formatter.Format(env.Dotenv(), eg.writer, nil)
```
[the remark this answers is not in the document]
spboyer: Already  `--export` combined with `--output` (non-default) returns an error: `--export and --output are mutually exclusive`. There's also a dedicated test (`TestEnvGetValuesExportOutputMutualExclusion`) validating this.

cli/azd/cmd/env.go
```diff
@@ -1338,9 +1354,41 @@ func (eg *envGetValuesAction) Run(ctx context.Context) (*actions.ActionResult, e
...
 }
 
+// writeExportedEnv writes environment variables in shell-ready
+// format (export KEY="VALUE") to the given writer. Values are
+// double-quoted with embedded double quotes and backslashes escaped.
```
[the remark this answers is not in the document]
spboyer: Updated doc comment to accurately describe all escaping behavior (backslashes, double quotes, dollar signs, backticks, carriage returns, and ANSI-C newline quoting).

cli/azd/cmd/env.go
spboyer: lifted `strings.NewReplacer` and `regexp.MustCompile` to package-level vars (`shellEscaper` and `validShellKey`) so they're built once.

spboyer: lifted `strings.NewReplacer` and `regexp.MustCompile` to package-level vars (`shellEscaper` and `validShellKey`) so they're built once.

spboyer: Fixed: lifted strings.NewReplacer and regexp.MustCompile to package-level vars (shellEscaper and validShellKey) so they are built once.

cli/azd/cmd/env.go
```diff
@@ -1338,9 +1362,65 @@ func (eg *envGetValuesAction) Run(ctx context.Context) (*actions.ActionResult, e
...
+// double-quoted with embedded backslashes, double quotes, dollar
+// signs, backticks, and carriage returns escaped. Newlines use
+// ANSI-C quoting ($'...') to ensure correct round-tripping through eval.
+func writeExportedEnv(
+	values map[string]string,
+	writer io.Writer,
+) error {
+	keys := slices.Sorted(maps.Keys(values))
+	for _, key := range keys {
+		if !validShellKey.MatchString(key) {
```
wbreza, line 1456: **[Medium]** `\r` without `\n` doesn't roundtrip through `eval`

The `shellEscaper` converts CR bytes to the two-character literal `\r`. When the value does NOT contain `\n`, the double-quote path is used (e.g., `export KEY="line1\rline2"`). In bash double quotes, `\r` is literal backslash + r — **not** a carriage return. The original CR byte is silently corrupted after `eval`.

The ANSI-C `$'...'` path handles this correctly (since `\r` in `$'...'` IS interpreted as CR), but it's only triggered for values containing `\n`.

**Suggested Fix:**
```suggestion
if strings.Contains(val, "\n") || strings.Contains(val, "\r") {
```

cli/azd/cmd/env.go
```diff
@@ -1263,17 +1265,32 @@ func newEnvGetValuesCmd() *cobra.Command {
 	return &cobra.Command{
 		Use:   "get-values",
 		Short: "Get all environment values.",
-		Args:  cobra.NoArgs,
+		Long: "Get all environment values.\n\n" +
+			"Use --export to output in shell-ready format " +
+			"(export KEY=\"VALUE\").\n" +
+			"This enables shell integration:\n\n" +
```
wbreza: **[Medium]** Output is POSIX-only — no Windows/PowerShell support

azd officially supports Windows and has `azd completion powershell`. The `--export` flag only produces bash/zsh `export KEY=""VALUE""` syntax. PowerShell users (a significant portion of Azure developers) cannot use this feature.

It would be great to update this to include support for both Windows & POSIX environments. Consider adopting vhvb1989's earlier suggestion of `--output export-bash` / `--output export-pwsh` to leverage the existing `output.Format` system, or at minimum document the POSIX-only limitation in the help text.

As a near-term workaround, PowerShell users can use: `azd env get-values --output json | ConvertFrom-Json`

cli/azd/cmd/env.go
```diff
@@ -1338,9 +1362,65 @@ func (eg *envGetValuesAction) Run(ctx context.Context) (*actions.ActionResult, e
...
+// validShellKey matches valid POSIX shell identifiers:
+// starts with a letter or underscore, followed by alphanumerics or underscores.
+var validShellKey = regexp.MustCompile(`^[a-zA-Z_][a-zA-Z0-9_]*$`)
+
+// writeExportedEnv writes environment variables in shell-ready
+// format (export KEY="VALUE") to the given writer. Values are
+// double-quoted with embedded backslashes, double quotes, dollar
+// signs, backticks, and carriage returns escaped. Newlines use
+// ANSI-C quoting ($'...') to ensure correct round-tripping through eval.
+func writeExportedEnv(
```
wbreza, line 1449: **[Low]** Silent key skipping without user feedback

Keys failing the `validShellKey` regex (e.g., keys with hyphens like `app-config`, dots like `my.setting`) are silently dropped from export output. Since `azd env set` accepts arbitrary key names, users may have keys that don't match the POSIX identifier pattern and not understand why they're missing.

Consider emitting a warning to stderr for skipped keys so users know their values weren't exported.

cli/azd/cmd/env.go
```diff
@@ -1338,9 +1362,65 @@ func (eg *envGetValuesAction) Run(ctx context.Context) (*actions.ActionResult, e
...
+	values map[string]string,
+	writer io.Writer,
+) error {
+	keys := slices.Sorted(maps.Keys(values))
+	for _, key := range keys {
+		if !validShellKey.MatchString(key) {
+			continue
+		}
+
```
wbreza, line 1465: **[Low]** ANSI-C `$'...'` quoting is not universal POSIX

`$'...'` quoting is supported by bash, zsh, and ksh but NOT by strict POSIX shells like `dash` (default `/bin/sh` on Debian/Ubuntu). If a user has multiline env var values and sources with `sh` instead of `bash`, the export will produce syntax errors.

Consider adding a note in the help text that bash/zsh/ksh is required for values containing newlines.

cli/azd/cmd/env.go
```diff
@@ -1305,6 +1322,13 @@ func newEnvGetValuesAction(
 }
 
 func (eg *envGetValuesAction) Run(ctx context.Context) (*actions.ActionResult, error) {
+	if eg.flags.export && eg.formatter.Kind() != output.EnvVarsFormat {
```
jongio, line 1353: **[LOW]** Error Handling: inconsistent error wrapping

This returns a bare `fmt.Errorf` wrapping `ErrInvalidFlagCombination`, but the existing pattern in this file (line ~225, the `--file` + args check) and every other use of `ErrInvalidFlagCombination` across the codebase wraps it in `ErrorWithSuggestion`:

```go
return nil, &internal.ErrorWithSuggestion{
    Err: fmt.Errorf(
        "--export and --output are mutually exclusive: %w",
        internal.ErrInvalidFlagCombination),
    Suggestion: "Use '--export' without '--output', or remove '--export' to use '--output json'.",
}
```

This gives the user actionable guidance instead of just reporting the conflict.

cli/azd/cmd/env.go
spboyer: Fixed in 13e31bc — now emits a warning to stderr for each skipped key: `warning: skipping key "app-config" (not a valid shell identifier)`.

cli/azd/cmd/env.go
spboyer: Fixed in 13e31bc — updated the Long description to note that $'...' quoting requires bash, zsh, or ksh. Also added a PowerShell workaround using `--output json | ConvertFrom-Json`.

cli/azd/cmd/env.go
spboyer: Fixed in 13e31bc — now returns `&internal.ErrorWithSuggestion{}` with both the `Err` and `Suggestion` fields populated, matching the pattern used elsewhere in this file.

cli/azd/cmd/env.go
spboyer: Fixed in 13e31bc — updated the condition to `strings.Contains(val, "\n") || strings.Contains(val, "\r")` so CR bytes also use ANSI-C quoting.

cli/azd/cmd/env.go
```diff
@@ -1338,9 +1375,71 @@ func (eg *envGetValuesAction) Run(ctx context.Context) (*actions.ActionResult, e
...
+				os.Stderr,
+				"warning: skipping key %q "+
+					"(not a valid shell identifier)\n",
+				key,
+			)
+			continue
+		}
+
+		val := values[key]
+		escaped := shellEscaper.Replace(val)
```
jongio: **[HIGH]** Logic: shellEscaper produces double-quote escapes, but the ANSI-C `$'...'` branch reuses the result

shellEscaper converts `$` to `\$` and backtick to `\`. Those escapes are correct inside `"..."` quoting, but `\$` and `\` aren't recognized ANSI-C escape sequences in `$'...'`. Bash preserves the backslash literally for unrecognized sequences:

    printf '%s' $'\$5' | od -An -tx1    # 5c 24 35 -> \$5 (extra backslash)

Any value with `$` or backtick AND a newline/CR won't roundtrip through eval - it gets an extra `\` before each `$` and backtick.

Fix: use separate escaping per quoting context. The `$'...'` path only needs `\` -> `\\`, `'` -> `\'`, `\n`, and `\r` as escape sequences. `$`, `"`, and backtick aren't special inside single-quoted output. Consider a second package-level Replacer for the ANSI-C path.

cli/azd/cmd/env.go
spboyer: Addressed in dfea26d — added `--shell pwsh` flag for native PowerShell export support:

```powershell
azd env get-values --export --shell pwsh | Invoke-Expression
```

Output uses `$env:KEY = "VALUE"` syntax with proper backtick escaping. Defaults to `bash` for backwards compatibility. Includes tests for special characters, backticks, newlines, and invalid shell validation.

cli/azd/cmd/env.go
```diff
@@ -1338,9 +1393,115 @@ func (eg *envGetValuesAction) Run(ctx context.Context) (*actions.ActionResult, e
...
+				os.Stderr,
+				"warning: skipping key %q "+
+					"(not a valid shell identifier)\n",
+				key,
+			)
+			continue
+		}
+
+		val := values[key]
+		escaped := pwshEscaper.Replace(val)
```
wbreza, line 1521: **[HIGH]** PowerShell export corrupts multiline values

`writePwshExportedEnv` embeds literal newline bytes into PowerShell double-quoted strings:

```
$env:MULTILINE = "line1
line2"
```

When piped to `Invoke-Expression`, this splits into two lines — an unclosed string on line 1 and a stray `line2"` on line 2, causing a parse error.

PowerShell uses backtick escape sequences for special characters (`
` for newline, `
` for CR), not C-style `\n`.

**Suggested fix** — add newline/CR escaping to `pwshEscaper`:

```suggestion
var pwshEscaper = strings.NewReplacer(
"`", "``",
`"`, "`\""",
`$`, "`$",
"\n", "`n",
"\r", "`r",
)
```

Also update the `pwsh export with newlines` test expectation to expect `
` instead of a literal newline byte.

cli/azd/cmd/env.go
```diff
@@ -1263,17 +1265,48 @@ func newEnvGetValuesCmd() *cobra.Command {
...
+			"Use --shell to select the syntax " +
+			"(bash or pwsh, default: bash).\n\n" +
+			"Bash/zsh/ksh (POSIX):\n\n" +
+			"  eval \"$(azd env get-values --export)\"\n\n" +
+			"PowerShell:\n\n" +
+			"  azd env get-values --export --shell pwsh " +
+			"| Invoke-Expression\n\n" +
+			"POSIX output uses $'...' (ANSI-C) quoting " +
+			"for values containing newlines\n" +
+			"or carriage returns, " +
```
weikanglim, line 1279: @spboyer This seems like a complicated feature we'd have to support longer term. My current concerns are:
1. azd would need knowledge of all shell syntax rules
2. There are other shells outside of just `sh`, e.g. `nu`, `fish`, that wouldn't work out-of-the-box
3. The importing logic into shell isn't that simple either.

I would strongly recommend going down the path of `azd env exec` (and `azd env exec --` for args support), which allows `azd` to passthrough environment variables to all shells, and we would eliminate all the complication.

I would imagine this being much more agent-friendly as well.

Happy to discuss further. I don't believe this fully resolves the concerns in the linked issue -- I think there are separately requirements around "filtered exports" and "automatic exports" that I can also design for if we're prioritizing fixing this issue today.

cli/azd/cmd/env.go
```diff
@@ -1338,9 +1393,115 @@ func (eg *envGetValuesAction) Run(ctx context.Context) (*actions.ActionResult, e
...
+// PowerShell backtick escape character.
+func writePwshExportedEnv(
+	values map[string]string,
+	writer io.Writer,
+) error {
+	keys := slices.Sorted(maps.Keys(values))
+	for _, key := range keys {
+		if !validShellKey.MatchString(key) {
+			fmt.Fprintf(
+				os.Stderr,
```
jongio: **[MEDIUM]** Quality: `os.Stderr` bypasses console abstraction

Both `writeExportedEnv` and `writePwshExportedEnv` write key-skipping warnings to `os.Stderr` directly. The action struct has `console input.Console` with `Handles().Stderr` - other cmd files use that instead. Using `os.Stderr` means tests can't verify warning output through the mock console.

Consider passing a `warnWriter io.Writer` parameter, or having the caller handle warnings through the console.

cli/azd/cmd/env.go
```diff
@@ -1305,6 +1338,28 @@ func newEnvGetValuesAction(
...
+				"--export and --output are mutually exclusive: %w",
+				internal.ErrInvalidFlagCombination,
+			),
+			Suggestion: "Use '--export' without '--output', " +
+				"or remove '--export' to use '--output json'.",
+		}
+	}
+
+	shell := strings.ToLower(eg.flags.shell)
+	if eg.flags.export && shell != "bash" && shell != "pwsh" {
```
jongio, line 1364: **[MEDIUM]** Logic: `--shell` without `--export` silently ignored

When `--export` isn't set, `--shell` is accepted but has no effect. `--shell fish` is silently OK here but errors with `--export` - inconsistent validation. Consider rejecting `--shell` (non-default) without `--export`, or at minimum emitting a stderr warning.

cli/azd/cmd/env.go
spboyer: @weikanglim Really appreciate the thoughtful feedback. You raise valid concerns — maintaining shell-specific syntax rules long-term is fragile, and it doesn't cover nu/fish/etc.

`azd env exec` is a much cleaner approach: azd sets the environment natively and passes through to any shell without needing to know syntax rules. That eliminates the escaping complexity entirely.

I'm on board with pivoting this direction. A few questions to align on scope:
1. Would `azd env exec -- <command>` be the basic form?
2. For the "filtered exports" and "automatic exports" requirements you mentioned — happy to discuss those as follow-up features.

For now, I'll fix the existing code bugs reviewers flagged (escaping, PowerShell newlines) since they're real correctness issues, and we can decide whether to rework this PR toward `azd env exec` or open a new one.

cli/azd/cmd/env.go
spboyer: Fixed in 4910725 — created a separate `ansiCEscaper` for the \$'...' path that only escapes `\\` and `'`. The double-quote branch continues using `shellEscaper`. \$ and backtick are no longer incorrectly escaped in ANSI-C quoting.

cli/azd/cmd/env.go
spboyer: Fixed in 4910725 — `pwshEscaper` now includes `"\n" → "\`n"` and `"\r" → "\`r"` so multiline values roundtrip correctly through `Invoke-Expression`. Test updated to expect backtick escapes.

cli/azd/cmd/env.go
spboyer: Fixed in 4910725 — both `writeExportedEnv` and `writePwshExportedEnv` now accept a `warnWriter io.Writer` parameter instead of hardcoding `os.Stderr`. Callers pass `os.Stderr` for now; tests can inject a buffer.

cli/azd/cmd/env.go
spboyer: Fixed in 4910725 — `--shell` with a non-default value now returns `ErrorWithSuggestion` when `--export` is not set. Added a test case (`TestEnvGetValuesShellWithoutExport`).

cli/azd/cmd/env.go
```diff
@@ -1338,9 +1404,130 @@ func (eg *envGetValuesAction) Run(ctx context.Context) (*actions.ActionResult, e
...
+	"\r", `\r`,
+)
+
+// ansiCEscaper escapes characters for ANSI-C $'...' quoting.
+// In ANSI-C quoting, only backslash and single quote are special;
+// dollar signs and backticks are literal.
+var ansiCEscaper = strings.NewReplacer(
+	`\`, `\\`,
+	`'`, `\'`,
+)
```
jongio, line 1438: **[Medium]** Missing regression test for the combined ANSI-C escaping scenario

The ansiCEscaper was created to fix incorrect escaping when values contain newlines AND characters that shellEscaper would wrongly escape (like $ or backticks). But no test exercises this combined case - every ANSI-C test uses values with only newlines/CRs, no other special chars.

Consider adding a test case that combines newlines with single quotes or backslashes:

`go
{
    name:    "export value with newline and single quotes",
    envVars: map[string]string{"MIXED": "it's\nmultiline"},
    export:  true,
    shell:   "bash",
    expected: "export AZURE_ENV_NAME=\\"test\\"\n" +
        "export MIXED=$'it\\'s\\nmultiline'\n",
},
`

This is the exact scenario the fix targets - without it, a future refactor could re-introduce the old shellEscaper bug and no test would catch it.

cli/azd/cmd/env_get_values_test.go
```diff
@@ -0,0 +1,336 @@
...
+			shell:  "fish",
+		},
+	}
+
+	_, err = action.Run(t.Context())
+	require.Error(t, err)
+	require.Contains(
+		t, err.Error(), "unsupported shell",
+	)
+}
```
jongio, line 321: **[Low]** warnWriter injection enables testing warning output, but no test uses it

The refactored functions accept warnWriter io.Writer - a nice improvement for testability. Consider adding a direct unit test of writeExportedEnv with a captured buffer to verify warning text is emitted for skipped keys:

`go
func TestWriteExportedEnvWarnings(t *testing.T) {
    var out, warn bytes.Buffer
    err := writeExportedEnv(
        map[string]string{"bad;key": "val", "GOOD": "val"},
        &out, &warn,
    )
    require.NoError(t, err)
    require.Contains(t, warn.String(), "skipping key")
}
`

Not blocking - the behavior (keys excluded from output) is tested through Run(). This would just verify the warning message itself.

cli/azd/cmd/env.go
spboyer: Fixed — added `"export value with newline and single quotes"` test case combining `\n` + `'` to exercise the ANSI-C path with characters that shellEscaper would wrongly escape.

cli/azd/cmd/env_get_values_test.go
spboyer: Fixed — added `TestWriteExportedEnvWarnings` that captures the warnWriter buffer and verifies skip warnings are emitted for invalid keys while valid keys appear in output.

cli/azd/cmd/env.go
weikanglim: @spboyer Thanks for entertaining the feedback. I spent some time drafting the alternate issue for `azd env exec` in [7423](https://github.com/Azure/azure-dev/issues/7423), and I do prefer it much better as a cleaner option.

Answering the questions briefly: 

1. Basic form would be `azd env exec`. Extended form for flag parsing would be `azd env exec --`.

    For example:
    ```bash
    azd env exec python script.py
    azd env exec ./write-env.sh
    # for passing flags to the script, `--` is required
    azd env exec -- python script.py --port 8000 --reload
    ```
    
    I think agents / automation would also have an easier time with this direction, since they can pass argv directly instead of composing `eval` / shell-specific export syntax.

2. I agree with the scoping decision here.

    My concern is mostly around overlapping concepts: if `--export` means a certain thing on `env get-values`, that could get confusing with future `env export` directions for filtered exports / automatic exports.
    

I opened [7423](https://github.com/Azure/azure-dev/issues/7423) for the `env exec` direction, and I'll capture remaining requirements from [4383](https://github.com/Azure/azure-dev/issues/4384) so we don't lose track of them.

spboyer: @weikanglim Ping — I'm on board with the `azd env exec` direction. Would love to sync on scope (basic form, filtered exports, etc.) so I can start on it. Happy to chat async or hop on a call.

hemarina (commented):
## Code Review: `--export` flag for `azd env get-values`

**Overall**: Shell escaping is **fundamentally sound**. No critical issues found.

### 🟡 MEDIUM

**1. NUL byte values silently pass through** (`env.go`)

NUL bytes (`\x00`) aren't escaped by any of the three replacers. Go's `io.WriteString` will write them to output, but when a shell evaluates `export KEY="hello\x00world"`, it **truncates at the NUL** — the variable silently becomes `"hello"` instead of `"hello\x00world"`. This is data corruption with no warning. Azure service values (certs, connection strings) could contain binary data. Fix: detect and warn, similar to the invalid-key warning.

**2. Dead `\r` entry in `shellEscaper`** (`env.go`)

The `shellEscaper` has `"\r", \\r\` but it's unreachable. The routing logic is:
```go
if strings.Contains(val, "\n") || strings.Contains(val, "\r") {
    // → ANSI-C path using ansiCEscaper, NOT shellEscaper
} else {
    escaped := shellEscaper.Replace(val) // only when NO \r present
}
```
So `\r` values never reach `shellEscaper`. Remove the dead entry to avoid confusing future maintainers.

### 🟢 LOW

**1. Missing test coverage** — No pwsh warning test, no pwsh `\r` test, no backslash+newline combo test, no empty env map test.

**2. `--shell bash` without `--export` silently ignored** (`env.go`)

The validation checks `if !eg.flags.export && shell != "bash"` — so `--shell pwsh` without `--export` correctly errors, but `--shell bash` without `--export` silently passes because `"bash"` is the default value. A user explicitly typing `azd env get-values --shell bash` might expect export behavior. This is hard to fix cleanly since Cobra's `pflag` can't distinguish "user passed the default" from "not passed at all."

spboyer:
Closing this one out. After more thought, I have concerns about `azd env exec` — it introduces a security surface (arbitrary command execution) and feels like overkill for what the standard command line can already do. If there's a simpler path forward for environment variable export, happy to revisit. Thanks @weikanglim for the thoughtful discussion.

PR #7366, opened 2026-03-27
vscode: do `npm audit fix`

bwateratmsft:
Supersedes https://github.com/Azure/azure-dev/pull/7358 which seems not to be building.

UPDATE: same problem...

### DIFF
--- a/.github/workflows/vscode-ci.yml
+++ b/.github/workflows/vscode-ci.yml
@@ -63,7 +63,6 @@ jobs:
 
       - name: Install dependencies
         run: |
-          npm install -g npm
           npm ci --no-optional
         working-directory: ./ext/vscode
 


bwateratmsft:
@danieljurek it looks like `npm install -g npm` is failing: https://github.com/Azure/azure-dev/blob/b70678d427e2e0d2863f744595dbdf0648dfdc76/.github/workflows/vscode-ci.yml#L66

Do you know why we're upgrading NPM at every CI like this?

I added a commit to remove that line--please sign off if that's OK.

ext/vscode/src/services/AzureDevShowProvider.ts
```diff
@@ -58,11 +58,11 @@ export class WorkspaceAzureDevShowProvider implements AzureDevShowProvider {
             const errorMessage = error instanceof Error ? error.message : String(error);
 
             if (errorMessage.includes('File is empty') || errorMessage.includes('unable to parse azure.yaml')) {
-                throw new Error(vscode.l10n.t('The azure.yaml file is invalid or empty. Please check the Problems panel for validation errors.'));
+                throw new Error(vscode.l10n.t('The azure.yaml file is invalid or empty. Please check the Problems panel for validation errors.'), { cause: error });
```
jongio, line 61: **[LOW]** Error Handling: Callers don't surface the cause chain

Good addition - `{ cause: error }` preserves the original CLI error. The callers (PickResourceGroupStep, PickResourceStep) currently only log `error.message` via the output channel, so the original cause won't appear in extension logs. The `callWithTelemetryAndErrorHandling` wrapper does read `.cause` for telemetry, so diagnostics improve there. If you want the cause in user-visible logs too, the callers would need a small update.

PR #7382, opened 2026-03-30
feat: support hooks per provisioning layer

### LINKED ISSUE #7186: feat: support hooks per provisioning layer
Support hooks per-provisioning layer.

We have gotten a few asks about it:
- [Request from azd-ai-autonomous-agent](https://github.com/curityio/azd-ai-autonomous-agent/blob/49f4af764f8eea11706744f66059ff60ee0d4ffa/OPEN-ISSUES.md)
- [Discussion topic](https://github.com/Azure/azure-dev/discussions/6184)

What this could look like:
```yaml
infra:
  layers:
    - name: base
      path: ./infra/base
      hooks:
        preprovision:
		  shell: sh
    	  run: ./infra/hooks/preprovision.sh
    - name: identity
      path: ./infra/identity 
      hooks:
        preprovision:
		  shell: sh
    	  run: ./infra/hooks/preprovision.sh
```
```

### PR DESCRIPTION
**Pull request by weikanglim**
## Summary
Fixes #7186 by adding first-class hook support for individual provisioning layers and wiring that support through the CLI, schema, and tests.

High-level changes:
- add `infra.layers[].hooks` support and execute layer hooks during `azd provision`
- extend `azd hooks run` with layer discovery and filtering via `--layer`
- add schema support for layer hooks in both alpha and v1.0 `azure.yaml` schemas
- move shared hook config YAML compatibility into `ext.HooksConfig` and reuse shared layer path resolution
- add unit, snapshot, and functional coverage for both `provision` and `hooks run`, including a faster hooks sample for the deploy smoke

## Validation
- `go test ./test/functional -run 'Test_CLI_Hooks(_Run)?_RegistrationAndRun' -count=1`

Fixes #7186

### DIFF
--- a/cli/azd/cmd/hooks.go
+++ b/cli/azd/cmd/hooks.go
@@ -351,44 +351,43 @@ func (hra *hooksRunAction) execHook(
 
 // Validates hooks and displays warnings for default shell usage and other issues
 func (hra *hooksRunAction) validateAndWarnHooks(ctx context.Context) error {
-	// Collect all hooks from project and services
-	allHooks := make(map[string][]*ext.HookConfig)
+	warningKeys := map[string]struct{}{}
+	validateAndWarn := func(cwd string, hooks map[string][]*ext.HookConfig) {
+		if len(hooks) == 0 {
+			return
+		}
 
-	// Add project hooks
-	for hookName, hookConfigs := range hra.projectConfig.Hooks {
-		allHooks[hookName] = append(allHooks[hookName], hookConfigs...)
+		hooksManager := ext.NewHooksManager(cwd, hra.commandRunner)
+		validationResult := hooksManager.ValidateHooks(ctx, hooks)
+
+		for _, warning := range validationResult.Warnings {
+			key := warning.Message + "\x00" + warning.Suggestion
+			if _, has := warningKeys[key]; has {
+				continue
+			}
+
+			warningKeys[key] = struct{}{}
+			hra.console.MessageUxItem(ctx, &ux.WarningMessage{
+				Description: warning.Message,
+			})
+			if warning.Suggestion != "" {
+				hra.console.Message(ctx, warning.Suggestion)
+			}
+			hra.console.Message(ctx, "")
+		}
 	}
 
-	// Add service hooks
+	validateAndWarn(hra.projectConfig.Path, hra.projectConfig.Hooks)
+
 	stableServices, err := hra.importManager.ServiceStable(ctx, hra.projectConfig)
 	if err == nil {
 		for _, service := range stableServices {
-			for hookName, hookConfigs := range service.Hooks {
-				allHooks[hookName] = append(allHooks[hookName], hookConfigs...)
-			}
+			validateAndWarn(service.Path(), service.Hooks)
 		}
 	}
 
-	// Add layer hooks
 	for _, layer := range hra.projectConfig.Infra.Layers {
-		for hookName, hookConfigs := range layer.Hooks {
-			allHooks[hookName] = append(allHooks[hookName], hookConfigs...)
-		}
-	}
-
-	// Create hooks manager and validate
-	hooksManager := ext.NewHooksManager(hra.projectConfig.Path, hra.commandRunner)
-	validationResult := hooksManager.ValidateHooks(ctx, allHooks)
-
-	// Display any warnings
-	for _, warning := range validationResult.Warnings {
-		hra.console.MessageUxItem(ctx, &ux.WarningMessage{
-			Description: warning.Message,
-		})
-		if warning.Suggestion != "" {
-			hra.console.Message(ctx, warning.Suggestion)
-		}
-		hra.console.Message(ctx, "")
+		validateAndWarn(layer.AbsolutePath(hra.projectConfig.Path), layer.Hooks)
 	}
 
 	return nil
--- a/cli/azd/internal/cmd/provision.go
+++ b/cli/azd/internal/cmd/provision.go
@@ -531,8 +531,7 @@ func (p *ProvisionAction) runLayerProvisionWithHooks(
 	layerPath string,
 	actionFn ext.InvokeFn,
 ) error {
-	hooks := map[string][]*ext.HookConfig(layer.Hooks)
-	if len(hooks) == 0 {
+	if len(layer.Hooks) == 0 {
 		return actionFn()
 	}
 
@@ -543,12 +542,12 @@ func (p *ProvisionAction) runLayerProvisionWithHooks(
 		p.envManager,
 		p.console,
 		layerPath,
-		hooks,
+		layer.Hooks,
 		p.env,
 		p.serviceLocator,
 	)
 
-	p.validateAndWarnLayerHooks(ctx, hooksManager, hooks)
+	p.validateAndWarnLayerHooks(ctx, hooksManager, layer.Hooks)
 
 	if err := hooksRunner.Invoke(ctx, []string{string(project.ProjectEventProvision)}, actionFn); err != nil {
 		if layer.Name == "" {
--- a/cli/azd/pkg/infra/provisioning/provider.go
+++ b/cli/azd/pkg/infra/provisioning/provider.go
@@ -5,7 +5,6 @@ package provisioning
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"path/filepath"
 	"strings"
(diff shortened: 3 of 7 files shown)

### CODE REVIEW

File: cli/azd/test/functional/hooks_test.go [hunk 1]
```diff
@@ -0,0 +1,150 @@
...
+	require.NoError(t, err)
+
+	cli.Env = append(cli.Env, fmt.Sprintf("HOOK_TRACE_FILE=%s", deployTracePath))
+	require.NoError(t, os.WriteFile(deployTracePath, []byte{}, 0600))
+	_, err = cli.RunCommand(ctx, "deploy")
+	require.Error(t, err, "deploy should fail for this hooks sample")
+
+	cli.Env = append(cli.Env, fmt.Sprintf("HOOK_TRACE_FILE=%s", provisionTracePath))
+	require.NoError(t, os.WriteFile(provisionTracePath, []byte{}, 0600))
+	_, err = cli.RunCommand(ctx, "provision")
```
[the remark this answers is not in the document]
weikanglim: Thanks — in this case `cli.Env` is passed directly to Go's `exec.Cmd.Env` in `test/azdcli/cli.go`, and Go's environment handling uses the last value when duplicate keys are present in the slice. So the later `HOOK_TRACE_FILE=...` entry is the one used by the child `azd` process here, and we intentionally left this behavior unchanged.

vhvb1989 (approved):

cli/azd/cmd/hooks.go
```diff
@@ -330,6 +369,13 @@ func (hra *hooksRunAction) validateAndWarnHooks(ctx context.Context) error {
 		}
 	}
 
+	// Add layer hooks
+	for _, layer := range hra.projectConfig.Infra.Layers {
```
jongio, line 389: **[MEDIUM]** Logic: Layer hooks validated with project-root cwd

These layer hooks get appended into the flat `allHooks` map and validated by a single `HooksManager` rooted at `hra.projectConfig.Path` (line 380). But layer hooks execute with `layer.AbsolutePath(projectConfig.Path)` as their working directory.

If a layer hook references a script file relative to the layer directory, `ValidateHooks` won't find it and will misclassify it as inline - producing incorrect default-shell warnings. `provision.go` already handles this correctly with per-layer validation in `validateAndWarnLayerHooks`.

Consider validating layer hooks per-layer with the correct cwd, matching the pattern in `provision.go`.

cli/azd/pkg/infra/provisioning/provider.go
```diff
@@ -121,22 +136,50 @@ func (o *Options) Validate() error {
...
+			if hookType == ext.HookTypeNone || eventName != "provision" {
+				return errWrap(
+					fmt.Sprintf("%s: only 'preprovision' and 'postprovision' hooks are supported", scope),
+				)
+			}
+		}
+
+		return nil
+	}
+
```
jongio, line 178: **[MEDIUM]** Error Handling: Misleading error prefix for root-level hooks

`errWrap` always prefixes with `"validating infra.layers: "` but this error fires when hooks are on the root `infra` object (no layers involved). A user with `infra: { path: infra, hooks: { preprovision: ... } }` sees `"validating infra.layers: 'hooks' can only be declared under 'infra.layers[]'"` - confusing since there are no layers.

Consider using `"validating infra: "` for `errWrap`, or a standalone `fmt.Errorf` for this specific check.

cli/azd/internal/cmd/provision.go
```diff
@@ -507,6 +525,60 @@ func (p *ProvisionAction) Run(ctx context.Context) (*actions.ActionResult, error
 	}, nil
 }
 
+func (p *ProvisionAction) runLayerProvisionWithHooks(
+	ctx context.Context,
+	layer provisioning.Options,
+	layerPath string,
+	actionFn ext.InvokeFn,
+) error {
+	hooks := map[string][]*ext.HookConfig(layer.Hooks)
```
jongio: **[LOW]** Refactoring: Redundant type conversion

`provisioning.HooksConfig` is a type alias for `ext.HooksConfig` which is `map[string][]*ext.HookConfig` - so this explicit conversion is a no-op. Can simplify to `if len(layer.Hooks) == 0` and pass `layer.Hooks` directly.

cli/azd/cmd/hooks.go
weikanglim: Fixed by validating hooks per scope instead of flattening them into a single project-root validation pass. `azd hooks run` now validates project, service, and layer hooks with a `HooksManager` rooted at each scope's cwd, so relative file hooks are classified correctly before execution.

cli/azd/pkg/infra/provisioning/provider.go
weikanglim: Good catch. I split the validation so root-level `infra.hooks` is handled directly in `Options.Validate()`, while `validateLayers()` only owns `infra.layers` checks. That keeps the root error from being mislabeled as a layers problem and should avoid future confusion if this logic changes again.

cli/azd/internal/cmd/provision.go
weikanglim: Agreed. I removed the no-op conversion and now check/use `layer.Hooks` directly in the provision path.

cli/azd/pkg/infra/provisioning/provider.go
```diff
@@ -116,27 +132,64 @@ func (o *Options) GetLayer(name string) (Options, error) {
...
+		}
+
+		seenLayers[layer.Name] = struct{}{}
+
 		if layer.Path == "" {
 			return errWrap(fmt.Sprintf("%s: path must be specified", layer.Name))
 		}
+
+		if err := validateHooks(layer.Name, layer.Hooks); err != nil {
+			return err
```
jongio, line 196: **[MEDIUM]** Logic: Missing errWrap on layer hook validation error

`validateHooks` returns a bare `fmt.Errorf` (line 167) and the caller here passes it through without wrapping. All other layer validation errors in `validateLayers()` use `errWrap` to prefix with `validating infra.layers:` (lines 177, 181, 187). This is a regression from the refactoring - the old code used `return errWrap(fmt.Sprintf(...))`.

cli/azd/pkg/infra/provisioning/provider.go
```diff
@@ -116,27 +132,64 @@ func (o *Options) GetLayer(name string) (Options, error) {
...
 // This should be called immediately right after Unmarshal() before any defaulting is performed.
 func (o *Options) Validate() error {
+
+	if len(o.Hooks) > 0 {
+		return errors.New("'hooks' can only be declared under 'infra.layers[]'")
+	}
+
+	if len(o.Layers) > 0 {
+		anyIncompatibleFieldsSet := func() bool {
+			return o.Name != "" || o.Module != "" || o.Path != "" || len(o.Hooks) > 0 || o.DeploymentStacks != nil
```
jongio: **[LOW]** Quality: Dead check in anyIncompatibleFieldsSet

`len(o.Hooks) > 0` here is unreachable - the guard at line 136 already returns when hooks are set at root level. Not a bug, but slightly misleading to future readers.

jongio (commented):
Previous comments addressed. One new item from the refactoring:
- provider.go:191 - layer hook validation error isn't wrapped with `errWrap`, inconsistent with all other layer errors
- provider.go:142 - dead `len(o.Hooks) > 0` check in `anyIncompatibleFieldsSet` (early return at line 136 catches it first)

Per-scope validation fix looks solid - good separation of cwd per hook scope.

cli/azd/pkg/infra/provisioning/provider.go
weikanglim: Fixed in the latest push. Layer-specific validation now returns the inner error, and `Options.Validate()` wraps the whole `validateLayers()` result once via `wrapValidateErr("infra.layers", err)`, so layer hook validation errors still carry the `validating infra.layers:` prefix without needing ad hoc string wrapping inside `validateLayers()`. I also tightened the test to assert the exact scoped message.

cli/azd/pkg/infra/provisioning/provider.go
weikanglim: Also fixed in the latest push. `anyIncompatibleFieldsSet` no longer includes `o.Hooks`; root hooks are handled by the earlier `len(o.Hooks) > 0` guard, so the helper now only checks the actually reachable incompatible root fields.

cli/azd/internal/cmd/provision.go
```diff
@@ -331,20 +342,27 @@ func (p *ProvisionAction) Run(ctx context.Context) (*actions.ActionResult, error
...
+			Args: map[string]any{
+				"preview": previewMode,
+				"layer":   layer.Name,
+				"path":    layerPath,
+			},
 		}
 
 		if p.alphaFeatureManager.IsEnabled(azapi.FeatureDeploymentStacks) {
 			p.console.WarnForFeature(ctx, azapi.FeatureDeploymentStacks)
 		}
```
wbreza, line 355: **[HIGH]** Architecture: Project hooks fire per-layer, inconsistent with deploy-hooks model

The layer loop calls `runLayerProvisionWithHooks` which wraps `p.projectConfig.Invoke(ProjectEventProvision, ...)` **inside each layer iteration**. This means project-level `preprovision`/`postprovision` hooks run N times for N layers:

```
Layer 1 preprovision
    Project preprovision     <- 1st time
    Deploy Layer 1
    Project postprovision    <- 1st time
Layer 1 postprovision
Layer 2 preprovision
    Project preprovision     <- 2nd time
    Deploy Layer 2
    Project postprovision    <- 2nd time
Layer 2 postprovision
```

This differs from the established **deploy-hooks model** where root hooks fire once as an outer envelope:

```
Root predeploy (once)
    Service 1 predeploy -> deploy -> postdeploy
    Service 2 predeploy -> deploy -> postdeploy
Root postdeploy (once)
```

Users will expect provision layer hooks to follow the same pattern:

```
Root preprovision (once)
    Layer 1 preprovision -> deploy -> postprovision
    Layer 2 preprovision -> deploy -> postprovision
Root postprovision (once)
```

**Impact:** Project-level preprovision hooks that set up auth, check prerequisites, or create resources will run multiple times — once per layer. If they're not idempotent, this causes incorrect behavior. Even if idempotent, repeated execution wastes time and produces confusing log output.

**Note:** The per-layer project hook invocation is pre-existing behavior (before this PR), but this PR makes it more visible by nesting layer hooks around it. Worth considering whether to fix the root hook invocation to match the deploy model.

jongio (approved):
Previous comments addressed. Looks good now.

cli/azd/internal/cmd/provision.go
weikanglim: @wbreza Thanks for raising the concern here. I'm not an expert in the hooks architecture, so I would defer the final decision to you.

What I understood from what you had mentioned -- you're concerned about the placement of `projectConfig.Invoke(ctx, project.ProjectEventProvision)` here. 

My understanding of how `ProjectEventProvision` is used today:
 - `pkg/project/service_target_containerapp.go`
 Registers a preprovision project handler to detect whether the target resource already exists and save RESOURCE_EXISTS.
 - `pkg/project/service_target_aks.go`
 Registers a postprovision project handler to set the kube context.
 - `internal/grpcserver/event_service.go`
Installed extensions can subscribe to project events

This presents the following requirements into consideration:
- `service_target_containerapp.go` needs its preprovision handler to run with the current layer’s timing/state, so RESOURCE_EXISTS is computed relative to that layer, not just once before the whole layered provision starts.
- `service_target_aks.go` needs its postprovision handler to run after the relevant layer completes, so kube context/namespace setup is available for subsequent layer hooks and operations, rather than only once at the very end.

 The current requirements may push us toward a different design. This PR doesn’t change `ProjectEventProvision` behavior; it only adds hooks support in each layer which is exposing the pattern.
 
 Given the current usage requirements, I’d prefer to scope that architectural question out of this PR. Would you mind capturing your proposed design in an issue so we can follow up separately and work toward a more consistent
 model?

rajeshkamal5050: Valid concern Wallace. The internal handlers (container apps RESOURCE_EXISTS, AKS kube context) are idempotent today so no functional impact, but agree the architecture should align with the deploy model before we add more handlers.

Filed #7495 to track this. Lets keep this PR moving and address the alignment in a follow-up.

PR #7392, opened 2026-03-30
Add --non-interactive alias and AZD_NON_INTERACTIVE env var

### LINKED ISSUE #7387: Add --non-interactive as alias for --no-prompt
## Summary

Add `--non-interactive` as a global flag alias for the existing `--no-prompt` flag.

## Motivation

The `--non-interactive` flag name is the de facto standard across CLI tooling (dotnet CLI, npm, pip, yarn, Homebrew, etc.). Developers and LLM agents trying azd for the first time will reach for `--non-interactive` and get an error. Adding it as an alias costs nothing and reduces friction.

The existing `--no-prompt` name should remain for backward compatibility.

## Implementation Notes

- Add `--non-interactive` as an alias in `CreateGlobalFlagSet()` in `cmd/auto_install.go`
- Both flags should set the same `GlobalCommandOptions.NoPrompt` field
- Update help text to mention both flag names
- Update snapshot tests (`TestFigSpec`, `TestUsage`)

### LINKED ISSUE #7388: Add AZD_NON_INTERACTIVE environment variable support
## Summary

Add an `AZD_NON_INTERACTIVE` environment variable that, when set to `true` or `1`, enables non-interactive (no-prompt) mode without requiring the `--no-prompt` CLI flag.

## Motivation

Currently, non-interactive mode can only be triggered via the `--no-prompt` CLI flag or automatic agent detection. There is no environment variable equivalent.

CI/CD pipelines and container-based workflows often configure behavior via environment variables rather than modifying command arguments. An `AZD_NON_INTERACTIVE=true` env var would make it trivial to configure non-interactive mode at the environment level (e.g., in a Dockerfile `ENV` directive or CI job config) without modifying every `azd` invocation.

## Implementation Notes

- Check for `AZD_NON_INTERACTIVE` in `ParseGlobalFlags()` in `cmd/auto_install.go`
- Priority: explicit `--no-prompt` flag > `AZD_NON_INTERACTIVE` env var > agent auto-detection
- Accept `true`, `1` as truthy values (use `strconv.ParseBool` for consistency)
- Document the env var in the CLI reference

spboyer:
## Summary

Add `--non-interactive` as a global flag alias for `--no-prompt` and support the `AZD_NON_INTERACTIVE` environment variable for enabling non-interactive mode.

## Changes

- **`--non-interactive` flag**: Hidden global flag alias for `--no-prompt` (backward-compatible)
- **`AZD_NON_INTERACTIVE` env var**: When set to a truthy value (`true`, `1`, `TRUE`, etc. via `strconv.ParseBool`), enables no-prompt mode
- **Priority order**: explicit `--no-prompt`/`--non-interactive` flag (highest) > `AZD_NON_INTERACTIVE` env var > agent auto-detection (lowest)
- **Agent detection suppression**: Setting `AZD_NON_INTERACTIVE` (even to `false`) suppresses agent auto-detection, giving users explicit control

## Files Modified

- `cli/azd/cmd/auto_install.go` — Flag definition + env var parsing in `CreateGlobalFlagSet()` / `ParseGlobalFlags()`
- `cli/azd/internal/global_command_options.go` — Updated `NoPrompt` doc comment
- `cli/azd/cmd/auto_install_test.go` — Table-driven tests for flag alias, env var, and priority

## Testing

- 13 table-driven test cases covering all combinations
- All existing `cmd` tests pass

Fixes #7387
Fixes #7388

### DIFF
--- a/cli/azd/pkg/azdext/mcp_security_test.go
+++ b/cli/azd/pkg/azdext/mcp_security_test.go
@@ -12,6 +12,8 @@ import (
 	"path/filepath"
 	"strings"
 	"testing"
+
+	"github.com/stretchr/testify/require"
 )
 
 func TestMCPSecurityCheckURL_BlocksMetadataEndpoints(t *testing.T) {
@@ -318,26 +320,12 @@ func TestMCPSecurityFluentBuilder(t *testing.T) {
 		RedactHeaders("Authorization").
 		ValidatePathsWithinBase("/tmp")
 
-	if policy == nil {
-		t.Fatal("fluent builder should return non-nil policy")
-	}
-
-	//nolint:staticcheck // SA5011 false positive: t.Fatal above guarantees non-nil
-	if !policy.blockMetadata {
-		t.Error("blockMetadata should be true")
-	}
-	if !policy.blockPrivate {
-		t.Error("blockPrivate should be true")
-	}
-	if !policy.requireHTTPS {
-		t.Error("requireHTTPS should be true")
-	}
-	if !policy.IsHeaderBlocked("Authorization") {
-		t.Error("Authorization should be blocked")
-	}
-	if len(policy.allowedBasePaths) != 1 {
-		t.Errorf("expected 1 base path, got %d", len(policy.allowedBasePaths))
-	}
+	require.NotNil(t, policy, "fluent builder should return non-nil policy")
+	require.True(t, policy.blockMetadata, "blockMetadata should be true")
+	require.True(t, policy.blockPrivate, "blockPrivate should be true")
+	require.True(t, policy.requireHTTPS, "requireHTTPS should be true")
+	require.True(t, policy.IsHeaderBlocked("Authorization"), "Authorization should be blocked")
+	require.Len(t, policy.allowedBasePaths, 1, "expected 1 base path")
 }
 
 func TestSSRFSafeRedirect_SchemeDowngrade(t *testing.T) {


cli/azd/cmd/auto_install_test.go
```diff
@@ -421,3 +422,125 @@ func TestParseGlobalFlags_AgentDetection(t *testing.T) {
...
+			args:         []string{"--no-prompt", "--non-interactive"},
+			wantNoPrompt: true,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			// Clear agent detection and AZD_NON_INTERACTIVE to isolate
+			// from the ambient environment.
+			clearAgentEnvVarsForTest(t)
```
jongio, line 520: **[HIGH]** Logic: `clearAgentEnvVarsForTest` doesn't include `AZD_NON_INTERACTIVE`

You're manually clearing `AZD_NON_INTERACTIVE` here because `clearAgentEnvVarsForTest` doesn't know about it yet. But the existing `TestParseGlobalFlags_AgentDetection` (same file, line ~395) also calls `ParseGlobalFlags` and only uses `clearAgentEnvVarsForTest` for env isolation. Since `ParseGlobalFlags` now checks `AZD_NON_INTERACTIVE` - and its mere presence suppresses agent detection - those tests will break if someone runs them with `AZD_NON_INTERACTIVE` set in their environment.

Fix: add `"AZD_NON_INTERACTIVE"` to the `envVarsToUnset` slice in `clearAgentEnvVarsForTest` (in `auto_install_integration_test.go`), then drop the manual `t.Setenv`/`os.Unsetenv` pair here.

cli/azd/cmd/auto_install.go
```diff
@@ -666,15 +672,36 @@ func ParseGlobalFlags(args []string, opts *internal.GlobalCommandOptions) error
...
+	// Environment variable: AZD_NON_INTERACTIVE enables no-prompt mode when set to a
+	// truthy value (parsed via strconv.ParseBool: "true", "1", "TRUE", etc.).
+	// Explicit flags take precedence over this env var.
+	// When this env var is present (regardless of value), it also suppresses
+	// agent auto-detection since the user has made an explicit choice.
+	envVarPresent := false
+	if !flagExplicitlySet {
+		if envVal, ok := os.LookupEnv("AZD_NON_INTERACTIVE"); ok {
+			envVarPresent = true
+			if parsed, err := strconv.ParseBool(envVal); err == nil && parsed {
```
jongio, line 696: **[MEDIUM]** Error Handling: Unparseable values silently ignored, agent detection suppressed

If someone sets `AZD_NON_INTERACTIVE=yes` or `=on` (reasonable guesses), `strconv.ParseBool` errors and the value is ignored. But `envVarPresent` is already `true` at this point, so agent auto-detection is also suppressed. The user gets neither non-interactive mode nor the agent-detection fallback.

Same issue with an empty export (`export AZD_NON_INTERACTIVE` with no value) - `LookupEnv` returns `("", true)`, `ParseBool("")` fails, agent detection is suppressed.

Consider logging a warning on parse failure, and adding a test case for `AZD_NON_INTERACTIVE=yes` to document expected behavior.

jongio (commented):
**Add --non-interactive alias and AZD_NON_INTERACTIVE env var** by @spboyer

**Summary**

What: Adds a hidden `--non-interactive` global flag alias for `--no-prompt` and introduces the `AZD_NON_INTERACTIVE` environment variable for controlling non-interactive mode. Priority order: explicit flags > env var > agent auto-detection.

Why: Provides a more standard `--non-interactive` / `AZD_NON_INTERACTIVE` interface that CI/CD tools and automation scripts expect (fixes #7387, #7388).

Assessment: The implementation is clean and the priority layering is solid. The 13 table-driven tests cover the key flag/env combinations well. Two gaps worth addressing: the shared test helper `clearAgentEnvVarsForTest` needs to include the new env var, and unparseable env var values are silently ignored while also suppressing agent detection. Also note: the `ParseGlobalFlags` function doc comment (line ~633) still only mentions `--no-prompt` - it should be updated to mention `--non-interactive` and `AZD_NON_INTERACTIVE`.

| Logic | 0 | 1 | 0 | 0 |
| Error Handling | 0 | 0 | 1 | 0 |
| **Total** | **0** | **1** | **1** | **0** |

**What's done well**
- Clean priority layering: flag > env var > agent detection
- Hidden flag avoids polluting help output
- Solid table-driven test structure matching existing patterns
- Good struct doc comments explaining all enablement paths

*2 inline comments below.*

cli/azd/cmd/auto_install_test.go
spboyer: Fixed in d970ffe — added `"AZD_NON_INTERACTIVE"` to `envVarsToUnset` in `clearAgentEnvVarsForTest`, and removed the redundant manual `t.Setenv`/`os.Unsetenv` pair.

cli/azd/cmd/auto_install.go
spboyer: Fixed in d970ffe — now logs a warning on parse failure (`warning: AZD_NON_INTERACTIVE="yes" is not a valid boolean`). Added a test case for `AZD_NON_INTERACTIVE=yes` that documents the expected behavior: warning logged, agent detection suppressed, NoPrompt remains false.

cli/azd/cmd/auto_install_test.go
```diff
@@ -421,3 +421,130 @@ func TestParseGlobalFlags_AgentDetection(t *testing.T) {
...
+			envVal:       "TRUE",
+			wantNoPrompt: true,
+		},
+		{
+			name:         "both flags coexist",
+			args:         []string{"--no-prompt", "--non-interactive"},
+			wantNoPrompt: true,
+		},
+		{
+			name:         "AZD_NON_INTERACTIVE=yes is unparseable, suppresses agent detection",
```
jongio: **[LOW]** Tests: This test is named "suppresses agent detection" but doesn't set any agent env vars (like `CLAUDE_CODE=1`), so no agent would be detected regardless. The assertion `NoPrompt==false` passes trivially.

To verify suppression, you'd need a case that sets both `AZD_NON_INTERACTIVE=yes` and an agent env var, then confirms `NoPrompt` stays false (proving the env var blocked agent detection). The test struct would need an `envVars map[string]string` field similar to `TestParseGlobalFlags_AgentDetection`, or a standalone sub-test.

Not critical - the behavior is correct, it's just not provably tested for this scenario.

spboyer: Fixed in 721efb1 — replaced the weak table entry with a standalone sub-test that sets `CLAUDE_CODE=1` (so agent detection would fire) plus `AZD_NON_INTERACTIVE=false`, then verifies `NoPrompt` stays false — proving the suppression actually works.

jongio (approved):
Reviewed 6 files (+206 -32). Checked flag integration (Cobra, extractGlobalArgs), env var parsing, priority logic, edge cases, and test coverage - clean. All previous feedback addressed.

Issue #7389, opened 2026-03-30
Route human-readable messages to stderr when --output json is active


spboyer:
## Summary

When `--output json` is used, route all human-readable messages (status text, informational messages, spinner fallback text) to stderr so that stdout contains only valid, parseable JSON.

## Motivation

When users or LLM agents use `--output json`, they expect stdout to contain only valid JSON. Currently, status messages, spinner fallback text, and informational output can intermingle with JSON on stdout, making it harder to parse programmatically.

Routing human-readable messages to stderr when JSON output is requested keeps stdout clean for machine consumption while preserving visibility of status information on stderr. This is a common pattern in well-behaved CLI tools (e.g., `docker`, `kubectl`, `jq`).

## Implementation Notes

- In commands that support `--output json`, redirect non-JSON output to stderr
- This affects `Console.Message()`, `Console.ShowSpinner()`, and similar display methods
- JSON result objects should remain on stdout
- Consider adding an `OutputMode` concept to the Console that controls routing

spboyer:
**Reviewed during issue triage — keeping open.**

TL;DR proposed fix:
- Add an `OutputMode` concept on `Console` / `AskerConsole` (in `pkg/input/`) that captures the active `--output` value.
- When `OutputMode == "json"`:
  - Route `Console.Message()`, `Console.MessageUxItem()`, spinner fallback text, and progress reports to **stderr**.
  - Keep only the final structured JSON result on **stdout**.
- This matches the well-established `docker` / `kubectl` / `jq` convention and lets agents reliably `jq` stdout without needing to strip status lines.

Pairs with #7391 (consolidating `SupportsInteractiveOutput`) — the same code path that decides spinner-vs-plain should also decide stdout-vs-stderr.

PR #7393, opened 2026-03-30
Route human-readable messages to stderr when --output json is active

spboyer:
- [x] Route human-readable messages to stderr when --output json is active
- [x] Fix previewer stdout leak in JSON output mode
- [x] Add missing test coverage for stderr routing (MessageUxItem, ShowPreviewer, StopPreviewer)
- [x] Fix EnsureBlankLine test isolation: remove preceding Message() call so the assertion only passes when EnsureBlankLine itself writes to stderr
- [x] Add zero-output assertions to StopPreviewer no-op test to verify it truly produces no output

### DIFF
--- a/cli/azd/pkg/input/console.go
+++ b/cli/azd/pkg/input/console.go
@@ -139,8 +139,8 @@ type Console interface {
 type AskerConsole struct {
 	asker   Asker
 	handles ConsoleHandles
-	// defaultWriter is the stdout handle, used to detect when the caller has
-	// redirected console output (e.g., to stderr for structured output).
+	// defaultWriter is the original output writer, used to restore after
+	// previewer stops.
 	defaultWriter io.Writer
 	// the writer which output is written to.
 	writer    io.Writer
@@ -337,10 +337,10 @@ func (c *AskerConsole) ShowPreviewer(ctx context.Context, options *ShowPreviewer
 	c.showProgressMu.Lock()
 	defer c.showProgressMu.Unlock()
 
-	// When the caller has redirected console output (e.g., to stderr for
+	// When console output is routed away from stdout (e.g., to stderr for
 	// structured output), skip the visual frame rendering (goterm writes
 	// to stdout directly) and return the redirected writer instead.
-	if c.writer != c.defaultWriter {
+	if c.writer != c.handles.Stdout {
 		return c.writer
 	}
 
@@ -440,10 +440,10 @@ func (c *AskerConsole) ShowSpinner(ctx context.Context, title string, format Spi
 	c.showProgressMu.Lock()
 	defer c.showProgressMu.Unlock()
 
-	// When the caller has redirected console output, emit the spinner
+	// When console output is routed away from stdout, emit the spinner
 	// title as plain text to the writer instead of using terminal
 	// animations that write to stdout directly.
-	if c.writer != c.defaultWriter {
+	if c.writer != c.handles.Stdout {
 		fmt.Fprintln(c.writer, title)
 		return
 	}
@@ -519,7 +519,7 @@ func (c *AskerConsole) StopSpinner(
 	lastMessage string,
 	format SpinnerUxType,
 ) {
-	if c.writer != c.defaultWriter {
+	if c.writer != c.handles.Stdout {
 		// When output is redirected, emit the stop message as plain
 		// text to the writer.
 		if lastMessage != "" {
@@ -1061,7 +1061,7 @@ func NewConsole(
 	c := &AskerConsole{
 		asker:         asker,
 		handles:       handles,
-		defaultWriter: handles.Stdout,
+		defaultWriter: writers.Output,
 		writer:        writers.Output,
 		formatter:     formatter,
 		isTerminal:    isTerminal,


cli/azd/pkg/input/console.go
```diff
@@ -145,6 +145,11 @@ type AskerConsole struct {
 	writer    io.Writer
 	formatter output.Formatter
 
+	// jsonOutputMode tracks whether JSON output format is active.
+	// When true, human-readable messages are routed to handles.Stderr
+	// so stdout contains only valid JSON.
+	jsonOutputMode bool
```
vhvb1989: console uses a `io.Writer` interface so it does not know what's the expected format or mode. Adding this here contradicts the design.  IIRC, the formatter is the one which should be transforming things into JSON, not the console

cli/azd/pkg/input/console_test.go
```diff
@@ -359,11 +359,156 @@ func TestAskerConsole_Message_InvalidQuery_FallsBack(t *testing.T) {
...
-	got := buf.String()
+	got := stderrBuf.String()
 	require.Contains(t, got, `"consoleMessage"`,
 		"invalid query should fall back to full envelope")
 }
 
+func TestAskerConsole_JsonOutputMode_StderrRouting(t *testing.T) {
+	const waitTime = 50 * time.Millisecond
+
+	tests := []struct {
```
jongio, line 369: **[MEDIUM]** Tests: Missing `MessageUxItem` stderr routing coverage

This table covers `Message`, `ShowSpinner`, and `StopSpinner` but not `MessageUxItem`, which has its own JSON serialization branch (marshals the UxItem envelope, applies optional JMESPath filter) that was also changed to use `messageWriter()`. It's a separate code path worth covering.

cli/azd/pkg/input/console.go
```diff
@@ -328,6 +344,13 @@ func (c *AskerConsole) ShowPreviewer(ctx context.Context, options *ShowPreviewer
 	c.showProgressMu.Lock()
 	defer c.showProgressMu.Unlock()
 
+	// In JSON output mode, skip the visual frame rendering (which writes to
+	// stdout via goterm) and return a writer that routes previewer output to
+	// stderr so it doesn't corrupt the JSON on stdout.
+	if c.jsonOutputMode {
```
jongio: **[MEDIUM]** Tests: This JSON-mode guard and the matching one in `StopPreviewer` (~line 382) don't have test coverage. `ShowPreviewer` returns `c.handles.Stderr` directly (bypassing `messageWriter()`) and `StopPreviewer` silently no-ops. A test confirming `ShowPreviewer` routes to stderr - and that `StopPreviewer` doesn't panic when the previewer was never created - would lock this down.

jongio (commented):
## PR Review - Azure/azure-dev#7393

**Route human-readable messages to stderr when --output json is active** by @spboyer

### Summary

What: Adds a `messageWriter()` helper to `AskerConsole` that routes human-readable output (messages, spinner text, previewer content) to stderr when `--output json` is active. Keeps stdout clean for parseable JSON.

Why: Fixes #7389 - downstream tooling parsing `azd` JSON output was getting corrupted by interleaved human-readable messages.

Assessment: `messageWriter()` is the right approach - it centralizes the routing decision and survives runtime writer swaps (e.g., `ShowPreviewer` temporarily taking over `c.writer`). The `ShowPreviewer`/`StopPreviewer` guards correctly prevent nil-pointer panics. Backward compatibility with non-JSON modes is preserved. Test coverage is solid for the core paths but has gaps for `MessageUxItem` and the previewer guards.

### Findings

| Tests | 0 | 0 | 2 | 0 |
| Logic | 0 | 0 | 0 | 1 |
| **Total** | **0** | **0** | **2** | **1** |

### Key Findings

1. **[MEDIUM] Tests**: `MessageUxItem` stderr routing isn't tested - it's a third distinct write path that was changed
2. **[MEDIUM] Tests**: `ShowPreviewer`/`StopPreviewer` JSON-mode guards lack test coverage
3. **[LOW] Logic**: `StopSpinner` JSON path doesn't have the idempotency guard from the non-JSON path - repeated calls with a non-empty message would double-print

### Test Coverage Estimate

- Well covered: `Message`, `ShowSpinner`, `StopSpinner` in JSON mode; `EnsureBlankLine`; backward compat for table/none modes; JSON query filter with stderr
- Gaps: `MessageUxItem` in JSON mode, `ShowPreviewer`/`StopPreviewer` in JSON mode

### What's Done Well

- `messageWriter()` is a clean single-responsibility abstraction
- `StopPreviewer` guard prevents nil-pointer panic on `c.previewer.Stop()` when previewer was never created
- Tests use `t.Context()` and table-driven patterns per repo conventions
- Non-JSON modes are completely untouched

---
*2 inline comments below.*

spboyer:
@vhvb1989 Good point about the boundary. The formatter does handle JSON serialization, but the issue is that *non-formatter* output (spinner text, status messages, `Message()` calls) still writes to stdout during JSON-mode operations. The formatter can't control those since they bypass it entirely.

The `jsonOutputMode` flag doesn't make the Console do formatting — it just tells it which file descriptor to write human-readable messages to (stdout vs stderr). The actual JSON output still goes through the formatter → stdout path unchanged.

That said, if you'd prefer this routing to live at a different layer (e.g., a middleware that swaps the Console's writer), happy to restructure. The net effect would be the same — keeping human messages off stdout when `--output json` is active.

cli/azd/pkg/input/console_test.go
```diff
@@ -359,11 +360,219 @@ func TestAskerConsole_Message_InvalidQuery_FallsBack(t *testing.T) {
...
+		nil,
+	)
+
+	ctx := t.Context()
+	c.Message(ctx, "first line")
+	c.EnsureBlankLine(ctx)
+
+	require.Empty(t, stdoutBuf.String(),
+		"stdout should be empty in JSON mode")
+	require.NotEmpty(t, stderrBuf.String(),
```
jongio, line 531: **[LOW]** Tests: Assertion doesn't isolate EnsureBlankLine behavior

`require.NotEmpty(stderrBuf)` passes even without the `EnsureBlankLine` call - `Message("first line")` on line 512 already wrote to stderr. This means the assertion labeled "EnsureBlankLine output should go to stderr" would pass if EnsureBlankLine was a complete no-op.

Consider testing EnsureBlankLine in isolation (remove the preceding `Message` call - EnsureBlankLine internally calls `Message`, so stderr still gets output), or capture `stderrBuf.Len()` before `EnsureBlankLine` and assert it grew.

cli/azd/pkg/input/console_test.go
```diff
@@ -359,11 +360,219 @@ func TestAskerConsole_Message_InvalidQuery_FallsBack(t *testing.T) {
...
+			Stdin:  os.Stdin,
+			Stdout: writerAdapter{stdoutBuf},
+		},
+		formatter,
+		nil,
+	)
+
+	ctx := t.Context()
+	// StopPreviewer without ShowPreviewer should not panic
+	require.NotPanics(t, func() {
```
jongio, line 585: **[LOW]** Tests: No-op test should verify zero output

`NotPanics` confirms the nil-guard works, but doesn't verify StopPreviewer is truly a no-op. Adding `require.Empty(t, stdoutBuf.String())` and `require.Empty(t, stderrBuf.String())` after the NotPanics block would lock down the zero-output guarantee.

cli/azd/pkg/input/console_test.go
```diff
@@ -359,11 +359,156 @@ func TestAskerConsole_Message_InvalidQuery_FallsBack(t *testing.T) {
...
+		},
+		{
+			name:   "NoneMode_Spinner_GoesToWriter",
+			format: string(output.NoneFormat),
+			action: func(ctx context.Context, c Console) {
+				c.ShowSpinner(ctx, "Working...", Step)
+				time.Sleep(waitTime)
+				c.StopSpinner(ctx, "Complete", StepDone)
+				time.Sleep(waitTime)
+			},
```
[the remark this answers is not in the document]
spboyer: The NoneMode spinner test in the table-driven suite (line 425) doesn't use `time.Sleep` — it deterministically calls `ShowSpinner`/`StopSpinner` and checks output. The `time.Sleep` calls at lines 71–91 are in a separate pre-existing test (`TestAskerConsoleNonInteractive`) that predates this PR.

cli/azd/pkg/input/console_test.go
spboyer: Fixed in a788205 — added `JsonMode_MessageUxItem_GoesToStderr` test case to the table-driven test that exercises the JSON serialization branch through `messageWriter()`.

cli/azd/pkg/input/console.go
spboyer: Fixed in a788205 — added `TestAskerConsole_JsonMode_ShowPreviewer_ReturnsStderr` verifying ShowPreviewer returns stderr, and `TestAskerConsole_JsonMode_StopPreviewer_NoOp` verifying StopPreviewer doesn't panic and produces zero output.

cli/azd/pkg/input/console.go
spboyer: Agreed — refactored in c3824da. Renamed `jsonOutputMode` to `structuredOutputMode` so the console no longer references a specific output format. The console treats it as a general structured-output flag; the caller (which knows about the formatter) decides what constitutes structured output. All comments and test names updated to reflect format-agnostic semantics.

cli/azd/pkg/input/console.go
```diff
@@ -145,6 +145,12 @@ type AskerConsole struct {
 	writer    io.Writer
 	formatter output.Formatter
 
+	// structuredOutputMode indicates that stdout is reserved for
+	// machine-parsable structured output. When true, human-readable
+	// messages are routed to stderr and visual elements (spinners,
+	// previewers) are suppressed.
```
weikanglim: @spboyer Today, we print out console message events, i.e. `consoleMessage` that the caller could pass through into the output terminal. This allows for azd to stream both informational messages, and other events (like deployments starting, etc.). 

Could you help us maintainers understand the motivation behind the change here? There may be alternatives here we could think of.

jongio: **[LOW]** Quality: Comment says "suppressed" but elements are redirected

Spinners and previewers aren't suppressed in structured mode - they emit plain text to stderr. ShowSpinner writes the title, StopSpinner writes the stop message, ShowPreviewer returns a stderr writer. "suppressed" reads as "no output" which isn't what happens.

Consider: "...and visual elements (spinners, previewers) are redirected to stderr as plain text."

Non-blocking nit.

jongio (approved):
Previous comments addressed. Rename to structuredOutputMode cleanly addresses the design feedback. One nit:
- console.go:151 - field comment says "suppressed" but elements are redirected to stderr as plain text

cli/azd/pkg/input/console.go
vhvb1989: The problem was not the name of the variable, lol.
The issue is the console being aware and deciding what to do.

The expected design for what you want to try could be to create a new formatter implementation with the logic you want to follow. Then you would call your formatter like `azd something -o json-no-events`  (assuming you implemented the `json-no-events` formatter.

IIRC, the VisualStudio-server integration we have, sends each json event to VS json-RPC so VS can stream what is happening to the user.

cli/azd/pkg/input/console.go
spboyer: @weikanglim Great question. The motivation is agent/automation consumption: when a tool calls `azd provision --output json`, it expects stdout to contain *only* valid JSON it can parse. Today, human-readable spinner messages, status updates, and `consoleMessage` events are interleaved on stdout alongside the structured JSON, which breaks machine consumers.

This change routes those human-readable messages to stderr so stdout is a clean, parsable JSON stream. Stderr still carries the informational output for humans watching the terminal.

The `consoleMessage` events you mention are still emitted — they go to stderr in structured output mode so they're visible but don't corrupt the JSON on stdout. Happy to discuss alternative approaches if this doesn't align with the intended architecture.

cli/azd/pkg/input/console.go
spboyer: Addressed the naming concern in c3824da (renamed `jsonOutputMode` → `structuredOutputMode`). The console no longer references JSON specifically — it uses a general structured-output flag set by the caller.

Regarding the broader point: the console already receives the formatter at construction time and uses it for `Message()` and `MessageUxItem()`. The `structuredOutputMode` boolean is computed once from the formatter's kind and cached — the console doesn't re-check the format type on each call. Happy to discuss further if a different layering is preferred.

cli/azd/pkg/input/console.go
spboyer: Fixed in b23f2dd — updated the comment to say "are redirected to stderr as plain text" instead of "are suppressed".

cli/azd/pkg/input/console.go
spboyer: @vhvb1989 Gentle ping — any thoughts on the `structuredOutputMode` refactoring approach? Happy to iterate further if a different layering is preferred.

cli/azd/pkg/input/console.go
spboyer: @weikanglim Ping — added context above on the motivation (keeping stdout parsable for JSON consumers). Let me know if you'd like to discuss alternative approaches.

jongio (approved):
Previous comments addressed. New commits (b23f2dd, 8c06676, e89ee72) fix the suppressed comment nit and resolve staticcheck SA5011 - both clean. No novel issues.

cli/azd/pkg/input/console.go
vhvb1989: @spboyer — following up on my earlier comment about the console not being the right place for this logic.

`cmd/container.go` (lines 127–131) already handles this exact concern at the right layer:

```go
writer := cmd.OutOrStdout()
// When using JSON formatting, we want to ensure we always write messages from the console to stderr.
if formatter != nil && formatter.Kind() == output.JsonFormat {
    writer = cmd.ErrOrStderr()
}
```

The console's `Output` writer is already set to stderr when `--output json` is active. This means `c.writer` inside the console already points to stderr for human-readable messages. If this existing wiring is not achieving what you need, we should understand *why* and fix it at that layer — not duplicate the decision inside the console.

The `structuredOutputMode` flag introduced here essentially re-checks what the caller already decided. But it goes further: it also makes the console bypass its own spinner/previewer logic based on the format. That couples the console to formatter semantics it should not know about. The console's job is to write to whatever `io.Writer` it was given, and the caller (container, VS Server, tests) decides where that writer points. This is the design that keeps the console reusable across contexts (CLI, VS Server, extensions, agents).

If the existing stderr redirect in `container.go` isn't sufficient for your scenario (e.g., the previewer writes directly to stdout via goterm, bypassing `c.writer`), the fix should be at the previewer/spinner level — not teaching the console about output formats. Consider the approach I suggested earlier: a new formatter or writer wrapper that encapsulates the routing, keeping the console format-agnostic.

spboyer: Addressed in 59e7bb4 — fully removed `structuredOutputMode` from the console.

The console is now format-agnostic as you described:
- `messageWriter()` simply returns `c.writer` (already stderr via container.go's redirect)
- Previewer/spinner detect writer redirection via `c.writer != c.defaultWriter` instead of checking format
- `StopPreviewer` checks `c.previewer == nil` (no format awareness)
- `defaultWriter` is set from `handles.Stdout` (real stdout) to enable comparison

The console's job is now just to write to whatever `io.Writer` it was given. The caller (container.go) decides where that writer points. Thanks for pushing on the right design here.

vhvb1989: I dug into *why* the `container.go` stderr redirect isn't enough, and the root cause is the **`goterm` library used by the previewer**.

`goterm` has a hardcoded global output destination ([terminal.go:45](https://github.com/buger/goterm/blob/v1.0.4/terminal.go#L45)):

```go
var Output *bufio.Writer = bufio.NewWriter(os.Stdout)
```

`tm.Print()`/`tm.Println()` buffer into an internal `Screen`, and `tm.Flush()` dumps it to `Output` — which is **always `os.Stdout`**. There is no way to redirect it via an `io.Writer`. The `progressLog` (previewer) in `progress_log.go` calls `tm.Print`, `tm.Println`, `tm.Flush`, `tm.MoveCursorUp` ~20 times across its lifecycle — all going straight to `os.Stdout`, bypassing `c.writer` entirely.

So `container.go`'s redirect **does work** for everything that flows through `c.writer` (`Message`, `MessageUxItem`, `println`, `StopSpinner`'s final message). The previewer is the hole.

The fix should be at the previewer level — e.g., make `progressLog` accept an `io.Writer` parameter instead of relying on goterm globals, or skip the visual frame rendering when the console is non-interactive / in structured output mode. That keeps the console format-agnostic and fixes the actual leak at its source.

spboyer:
@jongio @rajeshkamal5050 ADO CI failures here are **not caused by PR code** — all tests pass locally. Failing tests (`TestCompletionFigAction_Run`, `Test_CLI_VsServer`) are flaky ADO-environment issues. All lint, cross-compile, and local runs pass. Main has 0 failures on latest commit. Could you override checkenforcer? Thanks!

cli/azd/pkg/input/console.go
weikanglim: @spboyer Thanks for clarifying.

> it expects stdout to contain only valid JSON it can parse

This is already the intended behavior today: with `--output json`, `azd` reserves **stdout** for the final structured result. If anything is being interleaved into stdout, that's a bug in the current implementation and should be filed and fixed separately rather than addressed in this PR.

The other messages you're seeing are emitted on **stderr**. That stream already follows [NDJSON](https://github.com/ndjson/ndjson-spec), where each line is an independent JSON object. That existing split is what allows consumers to observe progress/events while still keeping stdout as a clean JSON output channel.

PR #7394, opened 2026-03-30
Add --fail-on-prompt flag for strict non-interactive mode

### LINKED ISSUE #7390: Add --fail-on-prompt flag for strict non-interactive mode
## Summary

Add a `--fail-on-prompt` mode that causes azd to fail immediately with a clear, actionable error whenever any interactive prompt is encountered, regardless of whether a default value exists. This should be the **automatic default behavior** when azd detects it is being called by an LLM agent or serving MCP tool requests.

## The Problem

The current `--no-prompt` flag silently uses default values when available. For LLM agents and automated pipelines, this masks decisions — the tool proceeds with a default that may not be what was intended, and the caller gets no signal that a choice was made on their behalf.

For example, when an LLM agent runs `azd up --no-prompt`:
- **Subscription**: azd silently picks the first subscription in the list — which might be a sandbox, not production
- **Location**: azd picks a default region that may be far from the user's target
- **Resource group**: Default naming may conflict with existing resources
- **Destructive confirmations**: `--no-prompt` auto-confirms deletions without the agent knowing

The agent never learns that these decisions were made. The user gets a deployment in the wrong subscription/region with no indication of why.

## Why This Matters

azd already auto-detects LLM agents (Claude Code, Copilot CLI, Gemini, OpenCode) via `agentdetect.IsRunningInAgent()` and auto-enables `--no-prompt`. This means **every LLM agent session today is silently using defaults**. The agent can't distinguish between "the user provided this value" and "azd guessed this value."

With fail-on-prompt behavior, the error message tells the agent exactly what's needed:

```
Error: prompt required: "Select an Azure subscription" — provide via --subscription flag
```

The LLM catches the error, asks the user (or infers from context), and retries with the correct flag. **That's a better interaction loop** — the LLM is forced to be explicit, and the user gets the deployment they actually wanted.

For MCP tool integrations, this is even more important: MCP tool descriptions document required parameters, and the LLM provides them upfront. Fail-fast errors are the standard feedback mechanism.

### Path 1: Auto-enable in MCP server (primary)

The MCP server (`internal/grpcserver/`) already knows it's serving an LLM. MCP tool invocations should **default to fail-on-prompt behavior**. The LLM never needs to know about a flag — the MCP tool descriptions document required parameters, and error responses tell the LLM what's missing. This is the most natural integration point since MCP is the structured API surface for LLM agents.

### Path 2: Auto-enable via agent detection (secondary)

For direct CLI invocations by agents (not via MCP), change the agent auto-detection behavior in `ParseGlobalFlags()` from enabling `--no-prompt` (silent defaults) to enabling `--fail-on-prompt` (fail-fast with helpful errors). The LLM doesn't choose the flag — azd chooses it automatically. Error-driven discovery is already how LLM agents work with most CLI tools: the error message IS the documentation.

### Path 3: Explicit flag for power users

Expose `--fail-on-prompt` as an explicit CLI flag for human power users in CI who want deterministic builds with no silent defaults. This is the escape hatch for users who want strict mode without agent detection.

## Implementation Notes

- When active, `askOneNoPrompt` should always return an error listing the prompt question, even if a default exists
- Error messages should suggest the appropriate CLI flag/arg to provide the value non-interactively (e.g., `use --subscription to specify`)
- The MCP tool schemas should document which parameters are required vs optional
- Consider a transitional period where agent-detected mode logs warnings ("using default subscription X") before switching to fail-fast

spboyer:
## Summary

Adds a `--fail-on-prompt` global flag that causes azd to fail immediately with a clear, actionable error whenever any interactive prompt is encountered — regardless of whether a default value exists.

This is now the **automatic default behavior** when azd detects it is being called by an LLM agent (Claude Code, GitHub Copilot CLI, Cursor, Gemini CLI, OpenCode, etc.).

## Motivation

The existing `--no-prompt` flag silently uses default values when available. For LLM agents, this masks decisions — the tool proceeds with a default that may not be what was intended. Fail-on-prompt ensures agents get explicit errors telling them exactly what flag or env var to provide.

## Changes

- **`GlobalCommandOptions.FailOnPrompt`** — new field in the options struct
- **`--fail-on-prompt` flag** — registered as a global flag; implies `--no-prompt`
- **`askOneFailOnPrompt()`** — new asker that always returns actionable errors listing the prompt question and available options
- **Agent auto-detection** — now sets both `NoPrompt` and `FailOnPrompt` (previously only set `NoPrompt`)
- **`--no-prompt` alone** — retains backward-compatible behavior (uses defaults silently)
- **Updated `NewAsker`/`NewConsole`** — accept `failOnPrompt` parameter
- **Comprehensive tests** — table-driven tests for all prompt types, flag parsing, agent detection

## Backward Compatibility

- `--no-prompt` behavior is **unchanged** — it still silently uses defaults
- Agent-detected sessions get stricter behavior (fail-on-prompt) which is a deliberate improvement
- Users can opt out with explicit `--no-prompt` to get the old silent-default behavior even when an agent is detected

Fixes #7390

### DIFF
--- a/cli/azd/extensions/azure.ai.agents/internal/cmd/init.go
+++ b/cli/azd/extensions/azure.ai.agents/internal/cmd/init.go
@@ -174,7 +174,6 @@ func runInitFromManifest(
 
 	console := input.NewConsole(
 		false, // noPrompt
-		false, // failOnPrompt
 		true,  // isTerminal
 		input.Writers{Output: os.Stdout},
 		input.ConsoleHandles{
@@ -995,7 +994,6 @@ func (a *InitAction) downloadAgentYaml(
 
 		console = input.NewConsole(
 			false, // noPrompt
-			false, // failOnPrompt
 			true,  // isTerminal
 			input.Writers{Output: os.Stdout},
 			input.ConsoleHandles{
--- a/cli/azd/extensions/azure.ai.finetune/internal/cmd/init.go
+++ b/cli/azd/extensions/azure.ai.finetune/internal/cmd/init.go
@@ -108,7 +108,6 @@ func newInitCommand(rootFlags rootFlagsDefinition) *cobra.Command {
 
 			console := input.NewConsole(
 				false, // noPrompt
-				false, // failOnPrompt
 				true,  // isTerminal
 				input.Writers{Output: os.Stdout},
 				input.ConsoleHandles{
@@ -811,7 +810,6 @@ method:
 
 			console = input.NewConsole(
 				false, // noPrompt
-				false, // failOnPrompt
 				true,  // isTerminal
 				input.Writers{Output: os.Stdout},
 				input.ConsoleHandles{
--- a/cli/azd/extensions/azure.coding-agent/internal/cmd/coding_agent_config.go
+++ b/cli/azd/extensions/azure.coding-agent/internal/cmd/coding_agent_config.go
@@ -428,7 +428,7 @@ func newCommandRunner(showOutput bool) (azd_exec.CommandRunner, input.Console) {
 		})
 	}
 
-	console := input.NewConsole(true, false, true, input.Writers{
+	console := input.NewConsole(true, true, input.Writers{
 		Output:  os.Stdout,
 		Spinner: os.Stdout,
 	}, input.ConsoleHandles{


wbreza (commented):
## Code Review — PR #7394

**Add --fail-on-prompt flag for strict non-interactive mode** by @spboyer

### DRY Assessment

Of 111 files changed, 86 are snapshot regenerations and 6 are extension go.mod/go.sum churn — only 19 files contain real changes. The core feature pattern is clean (follows existing `askOneNoPrompt` precedent), but there are DRY issues in the error messages and parameter threading.

### Findings

| Priority | Count |
|----------|-------|
| High | 2 |
| Medium | 4 |
| Low | 3 |
| **Total** | **9** |

#### 🔴 [HIGH] Extension go.mod files contain local `replace` directives
**Files:** `extensions/azure.ai.agents/go.mod`, `extensions/azure.ai.finetune/go.mod`, `extensions/azure.coding-agent/go.mod`

All three add `replace github.com/azure/azure-dev/cli/azd => ../../` — these are local development artifacts that break CI and anyone cloning the repo. Per AGENTS.md: *"Do NOT commit any go.mod/go.sum files in cli/azd/extensions/."*

**Fix:** `git checkout main -- cli/azd/extensions/*/go.mod cli/azd/extensions/*/go.sum`

#### 🔴 [HIGH] Error messages duplicated 15× across 3 files
**Files:** `pkg/input/asker.go`, `internal/grpcserver/prompt_service.go`, `pkg/prompt/prompt_service.go`

The string `"interactive prompt not allowed in strict mode"` appears in 15 places with minor variations. If the wording ever needs updating (i18n, adding a docs link), it must change in 15 places.

**Fix:** Extract 2 error helpers:
```go
func ErrFailOnPrompt(message string) error { ... }
func ErrFailOnPromptWithOptions(message string, options []string) error { ... }
```

#### 🟡 [MEDIUM] Positional bool explosion in `NewConsole`/`NewAsker`
`NewConsole(noPrompt, failOnPrompt, isTerminal, ...)` — three consecutive unlabeled bools. Consider an options struct for future extensibility.

#### 🟡 [MEDIUM] `askOneFailOnPrompt` type switch collapsible
4 of 6 cases produce the same error format. Could be collapsed to 2 cases using the error helpers above.

#### 🟡 [MEDIUM] `FailOnPrompt → NoPrompt` invariant not enforced in `NewConsole`
The implication is only enforced in `ParseGlobalFlags`. A caller constructing `NewConsole` directly with `failOnPrompt=true, noPrompt=false` creates an inconsistent state.

#### 🟡 [MEDIUM] Missing test for `--fail-on-prompt=false` with agent detection
No test verifies that an agent can explicitly disable strict mode while keeping no-prompt.

#### 🟢 [LOW] Agent auto-detection behavior change
Agent-detected sessions now get `FailOnPrompt` instead of just `NoPrompt`. This is the intended design (agents adapt by retrying with correct flags), but worth documenting in CHANGELOG.md.

#### �� [LOW] `WaitForEnter` redundant `failOnPrompt` check
`c.noPrompt || c.failOnPrompt` — the second condition is redundant if the invariant is enforced.

#### 🟢 [LOW] Extension code changes are mechanical but avoidable
Three extension `.go` files needed `false` parameter additions — an options struct would prevent future churn.

### What's Done Well

- **Clean architecture** — `askOneFailOnPrompt` follows the existing `askOneNoPrompt` pattern perfectly. New asker selected at construction, not conditional injection into every prompt method
- **Comprehensive tests** — table-driven tests for all 5 prompt types plus flag parsing/agent detection matrix
- **gRPC coverage** — `prompt_service.go` ensures extensions via gRPC also respect `FailOnPrompt`
- **Context-specific errors** in `pkg/prompt/` — subscription, location, resource group prompts give *specific, actionable guidance* (e.g., "use --subscription or azd config set defaults.subscription"). This is exactly what LLM agents need
- **`--fail-on-prompt implies --no-prompt`** — sound design, no invalid flag combinations

**Overall Assessment:** Comment — the feature design is solid and the core pattern is clean. The go.mod `replace` directives need removal before merge, and the 15× error message duplication should be consolidated.

cli/azd/pkg/input/asker_test.go
```diff
@@ -539,3 +539,137 @@ func Test_choicesFromOptions(t *testing.T) {
...
+				Options: []string{"auth", "storage", "compute"},
+			},
+			errContains: []string{
+				"interactive prompt not allowed in strict mode",
+				"Select features:",
+				"auth",
+				"storage",
+			},
+		},
+	}
```
jongio, line 620: **[MEDIUM]** Tests: Missing `*survey.Password` test case

The table covers Input, Select, Confirm, MultiSelect, and Editor (unknown type) - but Password isn't tested. The `askOneFailOnPrompt` function handles `*survey.Password` (returning the prompt message in the error), but this path has no test coverage. Adding a case is straightforward:

```suggestion
},
{
name: "Password_WithMessage",
prompt: &survey.Password{
Message: "Enter token:",
},
errContains: []string{
"interactive prompt not allowed in strict mode",
"Enter token:",
"command-line flags or environment variables",
},
},
}
```

cli/azd/internal/grpcserver/prompt_service.go
```diff
@@ -47,6 +47,14 @@ func NewPromptService(
 }
 
 func (s *promptService) Confirm(ctx context.Context, req *azdext.ConfirmRequest) (*azdext.ConfirmResponse, error) {
+	if s.globalOptions.FailOnPrompt {
```
jongio, line 56: **[MEDIUM]** Tests: No test coverage for gRPC FailOnPrompt paths

`prompt_service_test.go` has comprehensive `NoPrompt` tests for Confirm, Select, MultiSelect, and Prompt - but zero tests for `FailOnPrompt`. Unlike the asker-level fail-on-prompt (which is tested), these gRPC checks are a separate code path with their own error formatting. The existing NoPrompt tests are a good template - swap `NoPrompt: true` for `FailOnPrompt: true` and expect errors for all cases (with and without defaults).

jongio (commented):
**Add --fail-on-prompt flag for strict non-interactive mode** by @spboyer

**What**: Adds a `--fail-on-prompt` global flag that errors immediately on any interactive prompt - even when defaults exist. Agent-detected sessions now get this stricter mode by default.

**Why**: `--no-prompt` silently uses defaults, masking decisions from LLM agents. Fail-on-prompt gives agents explicit, actionable errors with the exact flag or env var to provide.

**Assessment**: Clean implementation following the established `askOneNoPrompt` pattern. Core design is sound - asker selection at construction, not conditional injection at every prompt site. @wbreza's review already identified the main issues (go.mod replace directives, error duplication, positional bool threading). Two test coverage gaps below.

| Tests | 0 | 0 | 2 | 0 |

**Test coverage estimate**:
- Well covered: `askOneFailOnPrompt` for Input/Select/Confirm/MultiSelect; flag parsing + agent detection matrix; `NewAsker` selection logic; backward compat of `--no-prompt`
- Gaps: `*survey.Password` in fail-on-prompt tests; gRPC `prompt_service.go` FailOnPrompt behavior (4 methods, all untested); `--fail-on-prompt=false` with agent detection (noted by @wbreza)

**What's done well**:
- `askOneFailOnPrompt` mirrors `askOneNoPrompt` perfectly - same structure, closure selection at construction
- Context-specific errors in `pkg/prompt/` give agents exactly what they need (flag names, config commands)
- `--fail-on-prompt implies --no-prompt` prevents invalid flag combinations
- Agent detection respects explicit flag overrides in both directions

*2 inline comments below.*

cli/azd/pkg/input/asker.go
```diff
@@ -27,6 +34,54 @@ func NewAsker(noPrompt bool, isTerminal bool, w io.Writer, r io.Reader) Asker {
...
+		return fmt.Errorf(
+			"interactive prompt not allowed in strict mode: %q"+
+				" (provide the value via command-line flags"+
+				" or environment variables)",
+			v.Message,
+		)
+	case *survey.Select:
+		return fmt.Errorf(
+			"interactive prompt not allowed in strict mode: %q"+
+				" (available options: %s — specify via"+
```
jongio: **[MEDIUM]** Quality: Em dash (U+2014) in error string

This format string contains a Unicode em dash between `%s` and `specify`. Same issue on line 64 in this file, and lines 98 and 158 in `internal/grpcserver/prompt_service.go` (4 occurrences total).

Since fail-on-prompt errors target LLM agents and CI pipelines, stick to ASCII. Some terminals and log parsers mangle multi-byte characters - the `\xE2\x80\x94` byte sequence already shows as `ΓÇö` in some Windows terminals. Use a regular hyphen or `--` instead.

```suggestion
" (available options: %s - specify via"+
```

cli/azd/pkg/input/console.go
```diff
@@ -1019,12 +1020,13 @@ type ExternalPromptConfiguration struct {
 // instead of prompting on the console.
 func NewConsole(
 	noPrompt bool,
+	failOnPrompt bool,
 	isTerminal bool,
```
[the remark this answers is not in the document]
spboyer: Fixed in 7ac0474 — added `failOnPrompt` guards before the `promptClient` path in `PromptDialog`, `Prompt`, `Select`, `MultiSelect`, and `Confirm` on the console. When `failOnPrompt` is true, these methods now short-circuit with an error before reaching external prompting.

cli/azd/pkg/input/asker.go
```diff
@@ -27,6 +34,47 @@ func NewAsker(noPrompt bool, isTerminal bool, w io.Writer, r io.Reader) Asker {
...
 }
 
+func askOneFailOnPrompt(p survey.Prompt, _ any) error {
+	switch v := p.(type) {
+	case *survey.Input:
+		return fmt.Errorf(
+			"interactive prompt not allowed in strict mode: %q"+
+				" (provide the value via command-line flags"+
+				" or environment variables)",
+			v.Message,
```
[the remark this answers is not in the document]
spboyer: Fixed in 7ac0474 — `*survey.Password` case was already present from a prior commit; verified it includes `v.Message` in the error. Added `Password_WithMessage` test case to cover it.

cli/azd/pkg/input/asker_test.go
spboyer: Fixed in 7ac0474 — added `Password_WithMessage` test case to the fail-on-prompt table test, verifying the error contains the prompt message.

cli/azd/internal/grpcserver/prompt_service.go
spboyer: Fixed in 7ac0474 — added 4 FailOnPrompt test cases to `prompt_service_test.go` (Confirm, Select, MultiSelect, Prompt), following the existing NoPrompt test pattern.

cli/azd/pkg/input/asker.go
spboyer: Fixed in 7ac0474 — replaced all 4 Unicode em dashes (U+2014) with ASCII `--` in `asker.go` (2) and `prompt_service.go` (2).

cli/azd/pkg/prompt/prompt_service.go
```diff
@@ -251,6 +251,15 @@ func (ps *promptService) PromptSubscription(
 
 	hideId := isDemoModeEnabled()
 
+	// Handle --fail-on-prompt mode: require explicit subscription flag
+	if ps.globalOptions.FailOnPrompt {
```
jongio, line 255: **[MEDIUM]** No tests for the FailOnPrompt paths in this file

5 blocks (PromptSubscription, PromptLocation, PromptResourceGroup, PromptSubscriptionResource, PromptResourceGroupResource) aren't tested. `prompt_service_test.go` covers NoPrompt paths well but doesn't exercise FailOnPrompt at all. These error messages are the most useful ones in the PR - they tell agents exactly which flag or config command to use. Add a test per method following the existing NoPrompt test pattern with `GlobalCommandOptions{FailOnPrompt: true}`.

spboyer: Fixed in 61ddd4d94 — added 5 FailOnPrompt tests covering PromptSubscription, PromptLocation, PromptResourceGroup, PromptSubscriptionResource, and PromptResourceGroupResource. Each verifies the actionable error message is returned.

cli/azd/pkg/input/asker.go
```diff
@@ -17,7 +17,14 @@ import (
 
 type Asker func(p survey.Prompt, response any) error
 
-func NewAsker(noPrompt bool, isTerminal bool, w io.Writer, r io.Reader) Asker {
+// NewAsker returns an Asker configured for the given prompt mode.
+// When failOnPrompt is true, every prompt immediately returns an actionable error.
+// When noPrompt is true, prompts silently use their default values.
+func NewAsker(noPrompt bool, failOnPrompt bool, isTerminal bool, w io.Writer, r io.Reader) Asker {
```
jongio, line 23: **[MEDIUM]** `NewAsker(noPrompt bool, failOnPrompt bool, isTerminal bool, ...)` - three consecutive bools with no compile-time protection against swapping. If someone accidentally writes `NewAsker(true, true, false, ...)` when they meant `NewAsker(true, false, true, ...)`, it compiles fine but does the wrong thing. 17 call sites had to be mechanically updated here.

Consider extracting an options struct in a follow-up - something like `AskerOptions{NoPrompt, FailOnPrompt, IsTerminal}`. Same applies to `NewConsole`. Not blocking since all current call sites are correct.

cli/azd/pkg/input/console.go
```diff
@@ -622,6 +628,12 @@ func (c *AskerConsole) PromptDialog(ctx context.Context, dialog PromptDialog) (m
 func (c *AskerConsole) Prompt(ctx context.Context, options ConsoleOptions) (string, error) {
 	var response string
 
+	if c.failOnPrompt && c.promptClient != nil {
```
jongio, line 629: **[MEDIUM]** The error here includes `(--fail-on-prompt)` in the message, but `askOneFailOnPrompt` in asker.go and the FailOnPrompt guards in grpcserver/prompt_service.go don't include it. An LLM agent hitting a prompt gets different error formats depending on whether an external prompt client is configured (this path) or not (asker.go path).

Suggest picking one format and using it everywhere - ideally a shared helper like `failOnPromptError(message string) error` that all four files call.

cli/azd/pkg/azdext/process_darwin.go
```diff
@@ -66,7 +66,7 @@ func findProcessByNameOS(name string) []ProcessInfo {
 	nameLower := strings.ToLower(name)
 	var results []ProcessInfo
 
-	for _, line := range strings.Split(string(output), "\n") {
+	for line := range strings.SplitSeq(string(output), "\n") {
```
jongio: **[LOW]** This `strings.Split` to `strings.SplitSeq` change isn't related to fail-on-prompt. Consider splitting it into a separate PR to keep this one focused.

cli/azd/pkg/input/asker.go
spboyer: Agreed — an `AskerOptions` struct (and similar for `NewConsole`) would eliminate the bool-swapping risk. Tracked for a follow-up.

cli/azd/pkg/input/console.go
spboyer: Fixed in 779cd90b9 — extracted `FailOnPromptError()` and `FailOnPromptSelectError()` helpers into `pkg/input/fail_on_prompt_error.go` and replaced all inline formatting in `asker.go`, `console.go`, and `prompt_service.go` with the shared helpers. Every call site now produces the same format including `(--fail-on-prompt)`.

cli/azd/pkg/azdext/process_darwin.go
spboyer: Reverted in 779cd90b9 — `process_darwin.go` restored to main via `git checkout origin/main`. The `strings.SplitSeq` change from `go fix` will go in a separate PR.

cli/azd/internal/grpcserver/prompt_service.go
```diff
@@ -194,6 +215,10 @@ func (s *promptService) MultiSelect(
 }
 
 func (s *promptService) Prompt(ctx context.Context, req *azdext.PromptRequest) (*azdext.PromptResponse, error) {
+	if s.globalOptions.FailOnPrompt {
```
jongio, line 222: Bug: `Prompt()` runs the `FailOnPrompt` check before the nil guard on `req.Options`. If an extension calls `Prompt` with nil `Options`, this panics on `req.Options.Message`. `Confirm` (line 52), `Select` (line 83), and `MultiSelect` (line 133) all guard nil first - this should follow the same pattern.

cli/azd/pkg/input/console.go
```diff
@@ -680,6 +688,10 @@ func choicesFromOptions(options ConsoleOptions) []promptChoice {
 
 // Prompts the user to select from a set of values
 func (c *AskerConsole) Select(ctx context.Context, options ConsoleOptions) (int, error) {
+	if c.failOnPrompt && c.promptClient != nil {
+		return -1, FailOnPromptError(options.Message)
```
jongio: `Select` and `MultiSelect` (line 776) both use `FailOnPromptError` here, but these are selection prompts with available options. The asker path uses `FailOnPromptSelectError` for these cases, which includes the available choices in the error message. Using the select-specific helper here would give agents the same actionable options list they get from the asker path.

cli/azd/extensions/azure.ai.agents/go.mod
```diff
@@ -109,3 +109,5 @@ require (
 	golang.org/x/text v0.34.0 // indirect
 	golang.org/x/time v0.14.0 // indirect
 )
+
+replace github.com/azure/azure-dev/cli/azd => ../../
```
jongio: These `replace` directives (also in `azure.ai.finetune/go.mod` and `azure.coding-agent/go.mod`) are breaking the extension lint CI jobs. Extension go.mod files can't have replace directives pointing to `../../` since they're published independently. This is likely a local dev convenience that needs to be removed before merge.

cli/azd/internal/grpcserver/prompt_service.go
spboyer: Fixed — added the nil guard (`req == nil || req.Options == nil`) before the `FailOnPrompt` check in `Prompt()`, matching the pattern used by `Confirm`, `Select`, and `MultiSelect`.

cli/azd/pkg/input/console.go
spboyer: Fixed — `Select` and `MultiSelect` now use `FailOnPromptSelectError(options.Message, choiceLabels)` to include available options in the error message, matching the asker path.

cli/azd/extensions/azure.ai.agents/go.mod
spboyer: Fixed — the `replace` directives were introduced by this branch (confirmed via `git diff origin/main`). Reverted all three extension `go.mod` and `go.sum` files to their `origin/main` versions.

cli/azd/pkg/prompt/prompt_service.go
jongio, line 255: `PromptLocation` resolves configured defaults (flag, config, single allowed value) before checking FailOnPrompt, so an agent with `azd config set defaults.location eastus` succeeds. But `PromptSubscription` fires FailOnPrompt before any default resolution - an agent with `azd config set defaults.subscription <id>` still gets an error.

This means an agent that correctly configured both defaults gets inconsistent behavior: location works, subscription fails. Consider factoring out the non-interactive resolution from the NoPrompt path so FailOnPrompt only fires when there's genuine ambiguity. Non-blocking since the error message tells agents what to do.

cli/azd/pkg/prompt/prompt_service.go
```diff
@@ -506,6 +522,14 @@ func (ps *promptService) PromptResourceGroup(
 		return nil, err
 	}
 
+	// Handle --fail-on-prompt mode
+	if ps.globalOptions.FailOnPrompt {
+		return nil, fmt.Errorf(
+			"interactive prompt not allowed in strict mode:" +
+				" resource group selection" +
```
jongio, line 529: The subscription and location errors give specific guidance (`use --subscription flag or 'azd config set defaults.subscription <id>'`), but the resource group, subscription-resource, and resource-group-resource errors say `specify via environment configuration` without naming which flag, env var, or config key. An LLM agent can act on `use --subscription flag` but can't act on `specify via environment configuration.` Consider naming the specific config key or env var.

jongio (approved):
Two non-blocking observations below. The feature is solid - clean shared error helpers, good test coverage across all prompt layers, and the agent detection logic handles edge cases well. All prior feedback is addressed.

cli/azd/pkg/prompt/prompt_service.go
spboyer: Fixed in dfa99d7ab — Updated all three vague error messages:
- Resource group: now references `AZURE_RESOURCE_GROUP` env var and `azd env set AZURE_RESOURCE_GROUP <name>`
- Subscription resource & resource group resource: now references `azure.yaml` and `azd env set`

All FailOnPrompt errors now give agents actionable guidance.

cli/azd/pkg/prompt/prompt_service.go
spboyer: Fixed in dfa99d7ab — `PromptSubscription` now resolves the configured default subscription (`defaults.subscription`) before checking FailOnPrompt. If a default is configured and found in the subscription list, it's returned directly without prompting. The error only fires when there's genuine ambiguity. Added a test for this path.

Issue #7351, opened 2026-03-26
[Issue] "azd ai agent init" lets you select un-supported locations for AI models


pamelafox:
- [x] Make sure you've installed the latest version using [instructions](https://github.com/Azure/azure-dev?tab=readme-ov-file#installupgrade-azure-developer-cli)

**Output from `azd version`**
azd version 1.24.0-beta.1-pr.6069275 (commit 2ef20fb1f0811263ed634c3f786126f8ab3d7fe9) (stable)

**Describe the bug**

When I do "azd ai agent init", it allows me to select westcentralus and gpt-5.2, even though westcentralus does not support any models. Ideally the picker would restrict the valid locations.

This is the error you get upon deploying:

{
  "error": {
    "code": "InvalidTemplate",
    "message": "Deployment template validation failed: 'The provided value for the template parameter 'location' is not valid. The value 'westcentralus' is not part of the allowed value(s): 'australiaeast,brazilsouth,canadacentral,canadaeast,eastus,eastus2,francecentral,germanywestcentral,italynorth,japaneast,koreacentral,northcentralus,norwayeast,polandcentral,southafricanorth,southcentralus,southeastasia,southindia,spaincentral,swedencentral,switzerlandnorth,uaenorth,uksouth,westus,westus2,westus3'.'.",
    "additionalInfo": [
      {
        "type": "TemplateViolation",
        "info": {
          "lineNumber": 1,
          "linePosition": 1376,
          "path": "properties.template.parameters.location.allowedValues"
        }
      }
    ]
  }
}

**To Reproduce**
"azd ai agent init"

**Additional context**

JeffreyCA:
Part of the issue is that when initializing from code, we don't call the model recovery flow

Another thing is that we don't filter out unsupported locations from the location prompt. Need to also investigate how we're determining the available models in a selected location

therealjohn:
> Part of the issue is that when initializing from code, we don't call the model recovery flow
> 
> Another thing is that we don't filter out unsupported locations from the location prompt. Need to also investigate how we're determining the available models in a selected location

I think we've unified the code paths now for init scenarios, so model recovery should be part of it. @trangevi  can confirm.

PR #7397, opened 2026-03-30
Extension Framework - add `allowed_locations` to `PromptLocation` and improve AI model capacity resolution

JeffreyCA:
Contributes to #7351

This PR adds location filtering support to the `PromptLocation` gRPC API and improves AI model deployment capacity resolution to be quota-aware, falling back to lower capacity values when the default exceeds remaining quota.

### Changes

**Location filtering (`PromptLocation`)**
- Adds `allowed_locations` field to `PromptLocationRequest` in the proto/gRPC API, enabling extensions to restrict which Azure locations are offered during prompts.
- Implements `filterLocationOptions` in `prompt_service.go` with case-insensitive matching to filter the location list before prompting or applying defaults.
- Improves error message when a default location is filtered out.

**Quota-aware capacity resolution (`ResolveCapacityWithQuota`)**
- Introduces `ResolveCapacityWithQuota` in `pkg/ai/model_service.go` — when no explicit capacity is preferred and the default exceeds remaining quota, it falls back to the highest valid capacity that fits within SKU constraints (min, max, step) and remaining quota.
- Integrates quota-aware resolution into `PromptAiDeployment`, `buildSkuCandidatesForVersion`, `resolveDeployments`, `modelHasQuota`, and `maxModelRemainingQuota`.

**AI Services location listing**
- `ListLocations` now queries resource SKU locations (`GetResourceSkuLocations`) for AIServices/S0 instead of returning all subscription locations, ensuring only AI-supported regions are returned.

### Testing

- Unit tests for `ResolveCapacityWithQuota` covering default fits, fallback, step alignment, explicit preferred exceeding quota, and below-minimum scenarios.
- Tests for `PromptLocation` with `AllowedLocations` and for default-filtered-out error path.
- Test for `buildSkuCandidatesForVersion` verifying fallback capacity with DeepSeek-style SKUs (step=0, large default).

### DIFF
--- a/cli/azd/pkg/ai/model_service.go
+++ b/cli/azd/pkg/ai/model_service.go
@@ -732,10 +732,7 @@ func convertSku(sku *armcognitiveservices.ModelSKU) AiModelSku {
 func ResolveCapacity(sku AiModelSku, preferred *int32) int32 {
 	if preferred != nil {
 		cap := *preferred
-		if cap > 0 &&
-			(sku.MinCapacity <= 0 || cap >= sku.MinCapacity) &&
-			(sku.MaxCapacity <= 0 || cap <= sku.MaxCapacity) &&
-			(sku.CapacityStep <= 0 || cap%sku.CapacityStep == 0) {
+		if capacityValidForSku(sku, cap) {
 			return cap
 		}
 	}
@@ -759,7 +756,7 @@ func ResolveCapacityWithQuota(sku AiModelSku, preferred *int32, remaining float6
 	return fallbackCapacityWithinQuota(sku, remaining)
 }
 
-func capacityFitsWithinQuota(sku AiModelSku, capacity int32, remaining float64) bool {
+func capacityValidForSku(sku AiModelSku, capacity int32) bool {
 	if capacity <= 0 {
 		return false
 	}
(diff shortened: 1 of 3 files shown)


cli/azd/pkg/ai/model_service.go
```diff
@@ -741,6 +742,84 @@ func ResolveCapacity(sku AiModelSku, preferred *int32) int32 {
...
+	if sku.MaxCapacity > 0 && capacity > sku.MaxCapacity {
+		return false
+	}
+
+	if sku.CapacityStep > 0 {
+		baseline := sku.MinCapacity
+		if baseline <= 0 {
+			baseline = sku.CapacityStep
+		}
+		if capacity < baseline || (capacity-baseline)%sku.CapacityStep != 0 {
```
jongio, line 774: **[MEDIUM]** Logic: Step alignment inconsistency with `ResolveCapacity`

`ResolveCapacity` (line 737) validates step alignment with `cap%sku.CapacityStep == 0` (absolute modulo), but this function uses `(capacity-baseline)%sku.CapacityStep != 0` (offset from MinCapacity). These produce different results when `MinCapacity > 0` and `MinCapacity % CapacityStep != 0`.

Example: `MinCapacity=7, CapacityStep=5` - `ResolveCapacity` rejects 12 (`12%5!=0`), but this function accepts it (`(12-7)%5==0`). This means `ResolveCapacityWithQuota` could return a capacity value that `ResolveCapacity` considers invalid, or reject one it'd accept.

The offset-based logic here is arguably more correct (steps should start from MinCapacity), but the two functions should agree. Consider extracting a shared `isStepAligned(sku, capacity)` helper used by both.

cli/azd/pkg/prompt/prompt_service_test.go
```diff
@@ -243,3 +243,81 @@ func TestPromptSubscription_NoPrompt_DefaultNotFound_DemoModeRedactsId(t *testin
...
+		mockConsole,
+		ucm,
+		subscriptionManager,
+		resourceService,
+		&internal.GlobalCommandOptions{NoPrompt: true},
+	)
+
+	location, err := ps.PromptLocation(context.Background(), &AzureContext{
+		Scope: AzureScope{SubscriptionId: "sub-123"},
+	}, &SelectOptions{
```
jongio, line 277: **[MEDIUM]** Tests: Missing case-insensitive matching coverage

`filterLocationOptions` normalizes via `strings.ToLower`, but both tests here pass already-lowercase values (`"westus3"`, `"eastus2"`). Consider adding a test with mixed-case input like `AllowedValues: []string{"WestUS3"}` to verify the case-insensitive matching works end-to-end. Extensions are likely to pass location names in various casings.

cli/azd/pkg/prompt/prompt_service.go
```diff
@@ -419,11 +432,40 @@ func (ps *promptService) PromptLocation(
...
+func filterLocationOptions(locations []account.Location, allowed []string) []account.Location {
+	if len(allowed) == 0 {
+		return locations
+	}
+
+	allowedSet := make(map[string]struct{}, len(allowed))
+	for _, location := range allowed {
+		normalized := normalizePromptLocationName(location)
+		if normalized == "" {
+			continue
```
jongio, line 449: **[LOW]** Error Handling: Misleading error when all allowed values are empty

If an extension passes `AllowedLocations: []string{" ", ""}`, all values normalize to empty strings, `allowedSet` ends up empty, and this returns nil. Downstream, the no-prompt path reports "default location not found in the available location options" when the real issue is the allowed list was effectively empty. This is an edge case, but a comment or explicit check here would prevent debugging confusion.

JeffreyCA: Thanks, fixed in earlier commit

PR #7043, opened 2026-03-08
feat(azdext): Add KeyVault resolver for extension secret resolution

jongio:
## Summary

Adds Key Vault secret resolution capabilities for both the azd host and the extension SDK (`pkg/azdext`), enabling extensions to resolve secret references in environment variables without custom Key Vault code.

### Host-Level Resolution (`pkg/keyvault` + `cmd/extensions.go`)

- Added `@Microsoft.KeyVault(SecretUri=...)` detection and parsing helpers to `pkg/keyvault`:
  - `IsKeyVaultAppReference` — detects `@Microsoft.KeyVault(SecretUri=...)` format
  - `ParseKeyVaultAppReference` — parses SecretUri into vault/secret/version components
  - `SecretFromKeyVaultReference` — resolves a single reference via `KeyVaultService`
  - `ResolveSecretEnvironment` — resolves all secret references in a `KEY=VALUE` env var list
- Wired resolution into `extensionAction.Run()` so azd-managed environment variables are resolved before being passed to extensions

### Extension SDK (`pkg/azdext`)

- Added `KeyVaultResolver` — standalone resolver for extensions with three supported formats:
  - `akvs://<subscription-id>/<vault-name>/<secret-name>`
  - `@Microsoft.KeyVault(SecretUri=https://<vault>.vault.azure.net/secrets/<secret>[/<version>])`
  - `@Microsoft.KeyVault(VaultName=<vault>;SecretName=<secret>[;SecretVersion=<version>])`
- Public API:
  - `IsSecretReference(s)` — detects all 3 formats with quote/whitespace stripping
  - `ParseSecretReference(ref)` — parses any format into `SecretReference`
  - `NewKeyVaultResolver(credential, opts)` — creates resolver with injectable credential and configurable vault suffix
  - `Resolve(ctx, ref)` — resolves a single reference
  - `ResolveMap(ctx, refs)` — resolves a map where all values are references
  - `ResolveEnvironment(ctx, env)` — resolves a mixed env var map (detects refs, resolves them, passes through plain values)
- Features: typed error classification via `KeyVaultResolveError`/`ResolveReason`, per-vault client caching with normalized keys, vault name validation, duplicate parameter rejection

### Test & CI Fixes

- Added mock `SecretFromKeyVaultReference` method to `cmdsubst_additional_test.go`
- Fixed gofmt formatting in `keyvault_test.go`

## Testing

- 314 tests pass in `pkg/azdext` (including 56 KeyVault resolver tests)
[...]

Issue #4067, opened 2024-07-03
[Issue] Make it easy to find azd env path


pamelafox:
- [X] Make sure you've installed the latest version using [instructions in the wiki](../wiki/install)

I want to easily get the azd env path, so that I can load that into my Python program's environments *without* doing an export. This is what I'm doing now:

```
def load_azd_env():
    """Get path to current azd env file and load file using python-dotenv"""
    result = subprocess.run("azd env list -o json", shell=True, capture_output=True, text=True)
    if result.returncode != 0:
        raise Exception("Error loading azd env")
    env_json = json.loads(result.stdout)
    env_file_path = None
    for entry in env_json:
        if entry["IsDefault"]:
            env_file_path = entry["DotEnvPath"]
    if not env_file_path:
        raise Exception("No default azd env file found")
    logger.info(f"Loading azd env from {env_file_path}")
    load_dotenv(env_file_path, override=True)
```

@vhvb1989 suggested that perhaps azd could make an environment variable for the current path.

jongio:
Options:

1. **Use `azd env list` like above**
1. **Default Environment Path**:
   - The default environment is listed in `.azure/config.json` in the `defaultEnvironment` property. You could build the path based on that. Read in that file, read the `defaultEnvironment` property, and build `.azure/{defaultEnvironment}/.env`.

1. **Avoiding azd Calls**:
   - Ideally, you shouldn't have to make an azd call to get this value. Having azd set an environment variable would only help if the app was launched in the same shell context where azd set that path. azd sets shell-level variables, not machine-level. So, you'd still have to read that environment variable within your app. If you just run `python app.py`, it won't have that environment variable in the same context. We wouldn't want to set a global environment variable because that would be shared amongst all apps on the machine.

1. **Solutions**:
   - One solution is to add an `azd env set-context` command that loads the azd environment values into the current process like `. azd env set-context`.
   - Another solution is to add a `--query/jmespath` option to the commands that allow you to call `azd env list` with a query to get the `dotenvpath` with a single call.

---

pamelafox:
Hm the jmespath would be the easiest add, I imagine, and is similar to Azure CLI SDK command args. I'm a little unsure about whether I'm going to run into the same issues as exporting if I were to do an `azd set-context.`

scottaddie:
@pamelafox @jongio JMESPath support was added in the linked PR ☝️

scottaddie:
[...]

Issue #2336, opened 2023-05-31
[feat] module to import appropriate .env


heaths:
It's great using azd to provision resources and being able to load the _.env_ files using godotenv or whatever, but it would be even easier in projects using azd this way if we provided a library to make it easier e.g.,

```go
func LoadEnv(env string) error {
    if env == "" {
        // get .azure/config@defaultEnvironment or err
    }
    // use godotenv to load environment's .env file,
    // optionally supporting fallback to/inclusion of any repo/ancestor .env file
}
```

It's not hard code to write, but any module support for azd (in Go or other languages) might help adoption.

weikanglim:
Cleanly separated modules hasn't really been a focus of the team. In fact, it seems that the recent shift has moved us towards thinking about layers in some places.

That being said, the code to do this currently looks like something like: https://github.com/Azure/azure-dev/blob/bfc7050577348db4070f8888676ea24d0b501e03/cli/azd/cmd/util.go#L117-L127

With perhaps a substitution of `azdContext.NewContextWithDirectory(<dir>)`

heaths:
I was actually thinking something outside of this module for use in other modules, so a log of this code may now work. Perhaps something that simply invokes `azd env get-values --output json` (optionally passing `--environment {name}`) and sets env vars similar to `godotenv` (doesn't override already-set env vars, though maybe that's optional).

weikanglim:
@ellismg Want your advice here: is it wise at all for our tight partners to use our modules, or the version listed https://pkg.go.dev/github.com/azure/azure-dev@v0.0.0-20230525235853-f2aafb04f11b/cli/azd/pkg/environment?tab=versions? I'm not sure how our current module publishing works.

ellismg:
To be 100% honest - I am very nervous about us partners taking dependencies on any exported methods from the azure-dev modules. 

Fundamentally, we don't think of the entire exported surface area of these modules as public.  We certainly don't intend to support the Go 1.0 promise of compatibility for all our exported members.

The artifact we ship today is the binary itself, and I view all the go code as an implementation detail of that binary.  Logically, we should view everything as under an `internal` folder.  Perhaps we should just do that at some point so folks can't accidentally break themselves into jail.  We had talked about restructuring the tree at one point and doing that, but we never wanted to deal with the churn.

I think we have enough work on our plate to build a great CLI tool and a set of experiences on top of that without bringing in the need to also produce and support a set of libraries for public adoption in addition to the application itself.

[...]

PR #7400, opened 2026-03-31
feat: add azd exec — run commands and scripts with azd environment context

### LINKED ISSUE #7520: Add azd exec: run commands and scripts with azd environment context
> **Design Specification**: [design-spec.md](https://github.com/Azure/azure-dev/blob/feature/microsoft-azd-exec/cli/azd/extensions/microsoft.azd.exec/design/design-spec.md)

## Problem

Running scripts that need azd environment variables today requires manual export/sourcing of `.env` files or wrapping commands in shell-specific workarounds. There's no first-class way to execute scripts or inline commands with full azd context (environment variables, Key Vault secret resolution) across platforms.

This also supersedes several long-standing env-loading requests: #391, #1697, #2336, #4067.

Related: #7423 proposes `azd env exec` as a core command ΓÇö see [Open Discussion](#open-discussion) below.

## Solution

Add the `microsoft.azd.exec` extension ΓÇö a cross-platform command and script execution engine that inherits azd environment context automatically.

### Execution Modes

| Invocation | Mode | How it works |
|---|---|---|
| `azd exec python script.py` | **Direct exec** | `exec.Command("python", "script.py")` ΓÇö exact argv, no shell |
| `azd exec 'echo $VAR'` | **Shell inline** | `bash -c "echo $VAR"` ΓÇö shell expansion available |
| `azd exec ./setup.sh` | **Script file** | `bash ./setup.sh` ΓÇö shell detected from extension |
| `azd exec --shell pwsh "cmd"` | **Shell inline** | `pwsh -Command "cmd"` ΓÇö explicit shell |

**Heuristic**: Multiple arguments without `--shell` ΓåÆ direct process execution (OS exec semantics). Single quoted argument or explicit `--shell` ΓåÆ shell inline execution. File path ΓåÆ script file execution.

### Features

- **Direct process execution**: Run programs with exact argv semantics (no shell wrapping) ΓÇö `azd exec python script.py` just works
- **Script execution**: Run script files or inline commands with configurable shell
- **Environment loading**: Inherits azd environment variables (Key Vault secret resolution handled by azd core)
- **Cross-platform shells**: bash, sh, zsh, pwsh, powershell, cmd
- **Shell auto-detection**: Detects shell from script file extension
- **Exit code propagation**: Child process exit codes forwarded faithfully for CI/CD pipelines
- **Interactive mode**: stdin passthrough for interactive scripts

### Usage

```bash
# Install the extension
azd extension install microsoft.azd.exec

# Run a command directly with azd environment (exact argv, no shell wrapping)
azd exec python script.py
azd exec npm run dev
azd exec -- python app.py --port 8000 --reload
azd exec docker compose up --build

# Execute a script file ΓÇö shell auto-detected from file extension
azd exec ./setup.sh

# Inline shell command (single quoted argument uses shell)
azd exec 'echo $AZURE_ENV_NAME'

# Specify a shell explicitly
azd exec --shell pwsh "Write-Host $env:AZURE_STORAGE_ACCOUNT"

# Pass arguments through to the script
azd exec ./build.sh -- --verbose --output ./dist

# Interactive mode ΓÇö stdin passthrough for prompts
azd exec -i ./interactive-setup.sh
```

### Architecture

```
main.go                    ΓåÆ Entry point with exit code propagation
internal/cmd/              ΓåÆ Cobra CLI commands (root, version, listen)
internal/executor/         ΓåÆ Script execution engine + command builder + error types
internal/shellutil/        ΓåÆ Shared shell detection and validation
```

3 focused internal packages, no circular dependencies, structured error types for programmatic handling.

### Dependencies

- Depends on SDK features from #7025 and #7043 (both merged)
- Uses `pkg/azdext` SDK for extension bootstrap and gRPC client

### Extension vs Core Command

This is implemented as an extension. [Issue #7423](https://github.com/Azure/azure-dev/issues/7423) proposes a core `azd env exec` command. The extension approach provides faster iteration and no impact on core binary size, with the tradeoff of discoverability. Can be promoted to core later based on usage.

### Key Vault Secret Auto-Resolution

The azd host resolves `akvs://` and `@Microsoft.KeyVault(...)` references before passing environment to extensions. Child processes launched by `azd exec` receive materialized secrets. This is consistent with azd hooks behavior. [weikanglim's feedback](https://github.com/Azure/azure-dev/issues/7423#issuecomment-4193694380) suggests this should be opt-in. If addressed, it should be at the host level (affecting all extensions/hooks).

## Deliverables

- [x] Extension source code (`cli/azd/extensions/microsoft.azd.exec/`)
- [x] GitHub Actions lint workflow
- [x] ADO release pipeline
- [x] Cross-platform build scripts
- [x] Direct process execution mode (exact argv semantics)
- [x] Security hardening (control char stripping, debug log sanitization)
- [ ] Registry entry (added when first release publishes binaries with artifact checksums)

PR: #7400

jongio:
Fixes #7520

> **Design Specification**: [design-spec.md](https://github.com/Azure/azure-dev/blob/feature/microsoft-azd-exec/cli/azd/extensions/microsoft.azd.exec/design/design-spec.md)

## Description

Add `azd exec` ΓÇö a cross-platform command and script execution engine that runs programs with full azd environment context (environment variables, Key Vault secret resolution).

This addresses long-standing requests for running scripts with azd env vars: #391, #1697, #2336, #4067.

### Example Usage

```bash
# Install
azd extension install microsoft.azd.exec

# Run a command directly with azd environment (exact argv, no shell wrapping)
azd exec python script.py
azd exec npm run dev
azd exec -- python app.py --port 8000 --reload
azd exec docker compose up --build

# Execute a script file ΓÇö shell auto-detected from file extension
azd exec ./setup.sh

# Inline shell command (single quoted argument uses shell)
azd exec 'echo $AZURE_ENV_NAME'

# Specify a shell explicitly
azd exec --shell pwsh "Write-Host $env:AZURE_STORAGE_ACCOUNT"

# Pass arguments through to the script
azd exec ./build.sh -- --verbose --output ./dist

# Interactive mode ΓÇö stdin passthrough for prompts
azd exec -i ./interactive-setup.sh
```

### Execution Modes

| Invocation | Mode | How it works |
|---|---|---|
| `azd exec python script.py` | **Direct exec** | `exec.Command("python", "script.py")` ΓÇö exact argv, no shell |
| `azd exec 'echo $VAR'` | **Shell inline** | `bash -c "echo $VAR"` ΓÇö shell expansion available |
| `azd exec ./setup.sh` | **Script file** | `bash ./setup.sh` ΓÇö shell detected from extension |
| `azd exec --shell pwsh "cmd"` | **Shell inline** | `pwsh -Command "cmd"` ΓÇö explicit shell |

**Heuristic**: Multiple arguments without `--shell` ΓåÆ direct process execution (OS exec semantics). Single quoted argument or explicit `--shell` ΓåÆ shell inline execution. File path ΓåÆ script file execution with auto-detected or explicit shell.

### Features

- **Direct process execution**: Run programs with exact argv semantics (no shell wrapping) ΓÇö `azd exec python script.py` just works
- **Script execution**: Run script files or inline commands with configurable shell
- **Environment loading**: Inherits azd environment variables (Key Vault secret resolution handled by azd core)
- **Cross-platform shells**: bash, sh, zsh, pwsh, powershell, cmd
- **Shell auto-detection**: Detects shell from script file extension
- **Exit code propagation**: Child process exit codes forwarded faithfully for CI/CD pipelines
- **Interactive mode**: stdin passthrough for interactive scripts

### Architecture

```
main.go                    ΓåÆ Entry point with exit code propagation
internal/cmd/              ΓåÆ Cobra CLI commands (root, version, listen)
internal/executor/         ΓåÆ Script execution engine + command builder + error types
internal/shellutil/        ΓåÆ Shared shell detection and validation
```

3 focused internal packages, no circular dependencies, structured error types for programmatic handling.

### Test Coverage

| Package | Coverage |
|---------|----------|
| `cmd` | 91.4% |
| `executor` | 96.5% |
| `shellutil` | 93.3% |
| **Total** | **94.3%** |

23 tests passing. Table-driven tests with platform-aware scaffolding.

### What's Included

- Extension source code (`cli/azd/extensions/microsoft.azd.exec/`)
- GitHub Actions lint workflow (`.github/workflows/lint-ext-microsoft-azd-exec.yml`)
- ADO release pipeline (`eng/pipelines/release-ext-microsoft-azd-exec.yml`)
- Cross-platform build scripts (`build.ps1`, `build.sh`, `ci-build.ps1`)

### What's NOT Included (Follow-Up)

- **Registry entry**: Will be added when first release publishes binaries with artifact checksums

### Dependencies

- Depends on SDK features from #7025 and #7043 (both merged)
- Uses `pkg/azdext` SDK for extension bootstrap and gRPC client

## Open Discussion (ref: #7423)

The following items are open for broader team discussion. They don't block this PR but affect the longer-term direction:

### 1. Extension vs Core Command

This capability is implemented as an extension (`microsoft.azd.exec`). [Issue #7423](https://github.com/Azure/azure-dev/issues/7423) proposes a core `azd env exec` command for the same use case. The extension approach provides:
- Faster iteration cycle (independent release)
- No impact on azd core binary size or command surface
- Can be promoted to core later if usage validates the design

The tradeoff is discoverability ΓÇö users need to install the extension. This should be evaluated based on adoption data.

### 2. Key Vault Secret Auto-Resolution

The azd host resolves `akvs://` and `@Microsoft.KeyVault(...)` references before passing environment to extension subprocesses. This means child processes launched by `azd exec` receive materialized secrets in their environment. This is consistent with how azd hooks work today.

[weikanglim's concern](https://github.com/Azure/azure-dev/issues/7423#issuecomment-4193694380): automatic secret materialization should be opt-in rather than implicit. If this is addressed, it should be at the **host level** (affecting all extensions/hooks), not the extension level. The extension has no mechanism to "un-resolve" secrets that the host already materialized.

## Testing

```bash
cd cli/azd/extensions/microsoft.azd.exec
GOWORK=off go test -count=1 ./...
GOWORK=off go vet ./...
```


cli/azd/extensions/microsoft.azd.exec/README.md
```diff
@@ -0,0 +1,62 @@
...
+# Interactive mode
+azd exec -i ./interactive-setup.sh
+
+# Fail-fast on Key Vault errors
+azd exec --stop-on-keyvault-error ./prod-deploy.sh
+```
+
+## Features
+
+- **Shell auto-detection**: Detects shell from file extension or shebang line
```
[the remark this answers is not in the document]
jongio: Fixed — removed shebang claim. Detection is file-extension-only; shebang parsing can be added later if needed.

cli/azd/extensions/microsoft.azd.exec/internal/cmd/root.go
```diff
@@ -0,0 +1,144 @@
...
+		Long: `Exec is an Azure Developer CLI extension that executes commands and scripts
+with full access to azd environment variables and configuration.
+
+Examples:
+	azd exec ./setup.sh                           # Execute script file
+	azd exec 'echo $AZURE_ENV_NAME'               # Inline bash (Linux/macOS)
+	azd exec --shell pwsh "Write-Host 'Hello'"     # Inline PowerShell
+	azd exec --shell pwsh ./deploy.ps1             # Script with shell
+	azd exec ./build.sh -- --verbose               # Script with args
+	azd exec ./init.sh -i                          # Interactive mode`,
```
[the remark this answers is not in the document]
jongio: Fixed — moved flag before positional arg: azd exec -i ./init.sh so it is parsed correctly with SetInterspersed(false).

jongio:
Dependent on #7314

jongio:
## Known Issue: Environment Value Munging

When `loadAzdEnvironment()` is called (via `azd env get-values`), values go through a `godotenv.Marshal` → dotenv text → parse roundtrip that can munge certain values (e.g., leading zeros stripped, multiline values corrupted).

**azd core already passes env vars directly** to the extension process environment, so this roundtrip is only needed when `-e/--environment` overrides to a different environment. Without `-e`, the correct values are already present.

PR #7314 addresses the core `godotenv.Marshal` munging issue. Once that lands, the roundtrip will be safe. In the meantime, the default case (no `-e` flag) works correctly since it uses the env vars passed directly by azd core.

Tracking: depends on #7314

cli/azd/extensions/microsoft.azd.exec/main.go
```diff
@@ -0,0 +1,39 @@
...
+		// observe the same code the executed script returned.
+		var execErr *executor.ExecutionError
+		if errors.As(err, &execErr) {
+			os.Exit(execErr.ExitCode)
+		}
+
+		color.Red("Error: %v", err)
+		os.Exit(1)
+	}
+}
```
wbreza: **[High]** `main.go` bypasses `azdext.Run()` — extension framework anti-pattern #1

The extension manually creates `context.Background()` and handles errors with `color.Red()` + `os.Exit(1)`. Other extensions (e.g., `azure.ai.agents`) use the standard `azdext.Run(cmd.NewRootCommand())` one-liner.

This means: no OpenTelemetry trace propagation (invisible to distributed tracing), no gRPC structured error reporting (azd can't classify errors for telemetry), and manual `FORCE_COLOR` duplication in `init()`.

The exit code propagation requirement is valid (`azdext.Run()` hardcodes `os.Exit(1)`), but at minimum the custom `main()` should manually set up trace context and error reporting:

```suggestion
func main() {
	ctx := azdext.NewContext()
	ctx = azdext.WithAccessToken(ctx)
	rootCmd := cmd.NewRootCommand()

	if err := rootCmd.ExecuteContext(ctx); err != nil {
		// Propagate the child process exit code so callers (CI/CD, scripts)
		// observe the same code the executed script returned.
		if execErr, ok := errors.AsType[*executor.ExecutionError](err); ok {
			os.Exit(execErr.ExitCode)
		}

		// Report structured error to azd for telemetry
		_ = azdext.ReportError(ctx, err)
		color.Red("Error: %v", err)
		os.Exit(1)
	}
}
```

cli/azd/extensions/microsoft.azd.exec/extension.yaml
```diff
@@ -0,0 +1,10 @@
+capabilities:
+    - custom-commands
+    - metadata
+description: Execute scripts and commands with Azure Developer CLI context, environment variables, and Azure Key Vault secret resolution.
+displayName: Exec
+id: microsoft.azd.exec
+language: go
+namespace: exec
+usage: azd exec <command> [options]
+version: 0.5.0
```
wbreza: **[High]** Missing `requiredAzdVersion`

The extension depends on SDK features (`azdext.LoadAzdEnvironment`, `NewExtensionRootCommand`, `NewMetadataCommand`) that require a minimum azd version. Other extensions declare this (e.g., `azure.ai.agents` has `requiredAzdVersion: ">1.23.6"`). Without it, the extension could load on an incompatible azd version with confusing gRPC failures.

```suggestion
usage: azd exec <command> [options]
version: 0.5.0
requiredAzdVersion: ">1.23.6"
```

cli/azd/extensions/microsoft.azd.exec/build.ps1
```diff
@@ -0,0 +1,78 @@
...
+
+    # Delete the output file if it already exists
+    if (Test-Path -Path $OUTPUT_NAME) {
+        Remove-Item -Path $OUTPUT_NAME -Force
+    }
+
+    # Set environment variables for Go build
+    $env:GOOS = $OS
+    $env:GOARCH = $ARCH
+
```
wbreza: **[Medium]** Build scripts set ldflags for non-existent `Commit`/`BuildDate` variables

This line sets `-X 'microsoft.azd.exec/internal/cmd.Commit=...'` and `-X '...BuildDate=...'`, but the `cmd` package only declares `Version` (in `root.go`). The Go linker silently ignores ldflags for non-existent variables.

In contrast, `azure.ai.agents` has a dedicated `internal/version/version.go` that declares all three.

Consider either adding the variables to the `cmd` package or removing the unused ldflags from this script and `build.sh`.

cli/azd/extensions/microsoft.azd.exec/build.ps1
```diff
@@ -0,0 +1,78 @@
...
+# Create a safe version of EXTENSION_ID replacing dots with dashes
+$EXTENSION_ID_SAFE = $env:EXTENSION_ID -replace '\.', '-'
+
+# Define output directory
+$OUTPUT_DIR = if ($env:OUTPUT_DIR) { $env:OUTPUT_DIR } else { Join-Path $EXTENSION_DIR "bin" }
+
+# Create output directory if it doesn't exist
+if (-not (Test-Path -Path $OUTPUT_DIR)) {
+    New-Item -ItemType Directory -Path $OUTPUT_DIR | Out-Null
+}
```
wbreza: **[Medium]** Release build scripts missing security hardening flags

`ci-build.ps1` correctly uses `-trimpath`, `-buildmode=pie`, and `-tags=cfi,cfg,osusergo` for security hardening. This release script — which produces the published binaries — uses none of these flags.

It may be worth verifying whether the ADO pipeline template adds them automatically; if not, release binaries would have weaker hardening than CI builds.

cli/azd/extensions/microsoft.azd.exec/main.go
```diff
@@ -0,0 +1,39 @@
...
+func main() {
+	ctx := context.Background()
+	rootCmd := cmd.NewRootCommand()
+
+	if err := rootCmd.ExecuteContext(ctx); err != nil {
+		// Propagate the child process exit code so callers (CI/CD, scripts)
+		// observe the same code the executed script returned.
+		var execErr *executor.ExecutionError
+		if errors.As(err, &execErr) {
+			os.Exit(execErr.ExitCode)
```
wbreza: **[Medium]** Go 1.26 convention: use `errors.AsType` instead of `errors.As`

Per `AGENTS.md`: "ALWAYS use `errors.AsType` when checking if error matches a specific type."

This pattern appears in multiple test files as well (`executor_test.go`, `errors_test.go`). Also, tests should use `t.Context()` instead of `context.Background()` per Go 1.26 conventions.

```suggestion
		if execErr, ok := errors.AsType[*executor.ExecutionError](err); ok {
```

eng/pipelines/release-ext-microsoft-azd-exec.yml
```diff
@@ -0,0 +1,33 @@
...
+extends:
+  template: /eng/pipelines/templates/stages/1es-redirect.yml
+  parameters:
+    stages:
+      - template: /eng/pipelines/templates/stages/release-azd-extension.yml
+        parameters:
+          AzdExtensionId: microsoft.azd.exec
+          SanitizedExtensionId: microsoft-azd-exec
+          AzdExtensionDirectory: cli/azd/extensions/microsoft.azd.exec
+          SkipTests: true
```
wbreza: **[Low]** Release pipeline skips tests

Extension tests should ideally gate release builds. This may be intentional for initial setup (tests run in the separate lint workflow), but worth revisiting to ensure a broken extension can't be released.

wbreza (commented):
## Code Review — `microsoft.azd.exec` Extension

Good work on the clean architecture and comprehensive test coverage! A few items to consider before this is ready to merge.

### What Looks Good ✅
- **Clean package structure** — 3 focused packages (`cmd`, `executor`, `shellutil`) with clear separation of concerns
- **Strong error hierarchy** — Typed errors enable programmatic handling and clear user messages
- **Comprehensive table-driven tests** — Platform-aware scaffolding, validation edge cases, and PowerShell quoting coverage
- **Faithful exit code propagation** — Critical for CI/CD usage, well-implemented
- **Correct SDK usage** — `NewExtensionRootCommand`, `PersistentPreRunE` chain preservation, `NewMetadataCommand`

### Summary
| Priority | Count |
|----------|-------|
| Critical | 0 |
| High | 2 |
| Medium | 3 |
| Low | 2 |
| **Total** | **7** |

**Overall Assessment:** Comment — see inline comments for details.

cli/azd/extensions/microsoft.azd.exec/main.go
jongio: Fixed — adopted `azdext.NewContext()`, `WithAccessToken()`, and `ReportError()` while keeping custom exit code propagation (since `azdext.Run()` hardcodes `os.Exit(1)`).

cli/azd/extensions/microsoft.azd.exec/extension.yaml
jongio: Fixed — added `requiredAzdVersion: ">1.23.6"` matching azure.ai.agents convention.

cli/azd/extensions/microsoft.azd.exec/build.ps1
jongio: Fixed — removed stale `Commit`/`BuildDate` ldflags from `build.ps1`. Only `Version` is set now.

cli/azd/extensions/microsoft.azd.exec/build.ps1
jongio: Fixed — added `-trimpath`, `-buildmode=pie`, `-tags=cfi,cfg,osusergo`, and `-s -w` to `build.ps1` to match `ci-build.ps1` hardening.

cli/azd/extensions/microsoft.azd.exec/main.go
jongio: Fixed — replaced all `errors.As` with `errors.AsType[*T]` across all files. Also updated all tests to use `t.Context()` instead of `context.Background()` per Go 1.26 conventions.

eng/pipelines/release-ext-microsoft-azd-exec.yml
jongio: This is the standard convention — all 6 extension release pipelines in the repo use `SkipTests: true`. Tests gate via the separate lint workflow.

spboyer:
How are we logging failures, preflight on the script (if needed), if the script itself fails does it use azd logs so we can report when a user has used exec and passed or failed and why
 
Also how does this differ from hooks, pre and post?

jongio:
Good questions.

**On logging and error reporting:**

We're using the SDK's telemetry stack. `NewContext()` for trace propagation, `ReportError()` for structured error reporting on all failures (including script exit codes). When `AZD_DEBUG=true` it logs shell, command, and working dir to stderr. Script output streams straight to the console, same as running the script directly.

**On how this differs from hooks:**

Hooks are tied to azd lifecycle events. You define them in `azure.yaml` and they run as pre/post steps on provision, deploy, etc. `azd exec` is ad-hoc. No yaml config needed. Just `azd exec npm run dev` or `azd exec ./setup.sh` whenever you need your azd env vars available. It's the quick path for local dev, scripting, and CI tasks where hooks are overkill.

cli/azd/extensions/microsoft.azd.exec/main.go
```diff
@@ -0,0 +1,42 @@
...
+}
+
+func main() {
+	ctx := azdext.NewContext()
+	ctx = azdext.WithAccessToken(ctx)
+	rootCmd := cmd.NewRootCommand()
+
+	if err := rootCmd.ExecuteContext(ctx); err != nil {
+		// Report all errors to azd for telemetry before exiting.
+		_ = azdext.ReportError(ctx, err)
```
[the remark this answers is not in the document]
jongio: The explicit `NewContext()` and `WithAccessToken()` before `ExecuteContext` is intentional. This ensures the azdext context (tracing + access token) is available for `ReportError()` even when the command fails before `PersistentPreRunE` runs - e.g., flag parsing errors or cobra wiring issues. Relying on `rootCmd.Context()` after `Execute()` would lose telemetry for those early-failure paths.

cli/azd/extensions/microsoft.azd.exec/internal/cmd/root.go
```diff
@@ -0,0 +1,132 @@
...
+		}
+		if extCtx.NoPrompt {
+			_ = os.Setenv("AZD_NO_PROMPT", "true")
+		}
+
+		if extCtx.Environment != "" {
+			if err := loadAzdEnvironment(cmd.Context()); err != nil {
+				return fmt.Errorf("failed to load environment '%s': %w", extCtx.Environment, err)
+			}
+		}
```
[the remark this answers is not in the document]
jongio: This was addressed - `loadAzdEnvironment()` has been removed from the current code. The extension now relies on the host-injected env vars with KV references already resolved.

cli/azd/extensions/microsoft.azd.exec/go.mod
```diff
@@ -0,0 +1,104 @@
+module microsoft.azd.exec
+
+go 1.26.1
```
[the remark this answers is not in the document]
jongio: Intentional - PR #7488 bumps the rest of the project to `go 1.26.1` to match. Once that merges, this extension's go.mod will be aligned.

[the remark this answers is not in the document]
jongio: There's a separate PR in progress to bump Go across the entire project. This will be aligned as part of that effort.

cli/azd/extensions/microsoft.azd.exec/main.go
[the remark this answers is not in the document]
jongio: Fixed — adopted `azdext.Run()` with the new `WithExitCode` option that was added to the SDK in this PR. The extension now uses the standard entrypoint while still propagating child process exit codes:

```go
azdext.Run(
    cmd.NewRootCommand(),
    azdext.WithExitCode(func(err error) (int, bool) {
        if execErr, ok := errors.AsType[*executor.ExecutionError](err); ok {
            return execErr.ExitCode, true
        }
        return 0, false
    }),
)
```

cli/azd/extensions/microsoft.azd.exec/internal/cmd/root.go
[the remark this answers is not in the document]
jongio: Fixed — `loadAzdEnvironment()` was removed entirely after rebasing on upstream/main (which includes PR #7314). The extension now inherits all azd environment variables directly from the parent process. No subprocess call, no roundtrip, no value munging.

cli/azd/extensions/microsoft.azd.exec/internal/cmd/root.go
```diff
@@ -0,0 +1,109 @@
...
+
+		exec, err := executor.New(executor.Config{
+			Shell:       shell,
+			Interactive: interactive,
+			Args:        scriptArgs,
+		})
+		if err != nil {
+			return fmt.Errorf("invalid configuration: %w", err)
+		}
```
wbreza: **[Medium]** Silent fallback to inline execution on non-NotExist stat errors

When `os.Stat(absPath)` fails with something other than "not found" (e.g., EACCES, broken symlinks), the code silently falls through to `ExecuteInline`, producing a confusing failure instead of the real error.

Consider trying `Execute` first and falling back to `ExecuteInline` only on `ScriptNotFoundError` — this also eliminates the duplicate `filepath.Abs`+`os.Stat` between here and `executor.go`:

```suggestion
		if err := exec.Execute(cmd.Context(), scriptInput); err != nil {
			if _, ok := errors.AsType[*executor.ScriptNotFoundError](err); ok {
				return exec.ExecuteInline(cmd.Context(), scriptInput)
			}
			return err
		}
		return nil
```

cli/azd/extensions/microsoft.azd.exec/internal/executor/command_builder.go
```diff
@@ -0,0 +1,119 @@
...
+
+	cmd := exec.CommandContext(ctx, cmdArgs[0], cmdArgs[1:]...) //nolint:gosec // CLI command builder; args validated upstream
+	if useCmdLineOverride {
+		setCmdLineOverride(cmd, cmdArgs, cmdWrapOuter)
+	}
+	return cmd
+}
+
+// quoteCmdArg quotes a single argument for cmd.exe if it contains spaces
+// or metacharacters. Embedded double quotes are escaped by doubling them.
```
wbreza: **[Medium]** `quoteCmdArg` doesn't treat tab characters as word separators

Tab characters (`\t`) also function as argument delimiters in Windows command lines. An embedded tab would pass through unquoted, causing unintended argument splitting.

```suggestion
	if strings.ContainsAny(escaped, " \t&|<>^%\"") {
```

cli/azd/pkg/azdext/run.go
```diff
@@ -78,6 +87,12 @@ func Run(rootCmd *cobra.Command, opts ...RunOption) {
 			printError(err)
 		}
 
+		if cfg.exitCodeFunc != nil {
+			if code, ok := cfg.exitCodeFunc(err); ok {
+				os.Exit(code)
```
wbreza: **[Low]** `WithExitCode` allows exit code 0 in the error path

If a caller's `exitCodeFunc` returns `(0, true)`, `Run` calls `os.Exit(0)` despite having a real error, masking failures in CI. Current usage won't trigger this, but the API is generic.

Consider adding a defensive guard or documenting in the godoc that returning `(0, true)` is a contract violation.

cli/azd/pkg/exec/scripting/command_builder.go
```diff
@@ -0,0 +1,119 @@
...
+		} else {
+			cmdArgs = append(cmdArgs, e.config.Args...)
+		}
+	}
+
+	cmd := exec.CommandContext(ctx, cmdArgs[0], cmdArgs[1:]...) //nolint:gosec // CLI command builder; args validated upstream
+	if useCmdLineOverride {
+		setCmdLineOverride(cmd, cmdArgs, cmdWrapOuter)
+	}
+	return cmd
```
wbreza, line 78: **[Low]** `strings.NewReplacer` allocated on every call

Not a hot path, but a free optimization — consider hoisting to a package-level variable:

```go
var controlCharReplacer = strings.NewReplacer("\n", "", "\r", "", "\x00", "")
```

cli/azd/extensions/microsoft.azd.exec/internal/executor/command_builder.go
```diff
@@ -0,0 +1,119 @@
...
+// Newline/CR/null bytes are stripped as they act as command separators.
```
wbreza: **[Low]** Consider documenting the `%`-variable expansion limitation

cmd.exe expands `%VAR%` patterns even inside double quotes — an inherent limitation with no workaround. A brief code comment would prevent future maintainers from thinking the quoting prevents expansion.

wbreza (approved):
## Code Review — `microsoft.azd.exec` Extension

Great extension contribution! The architecture is clean and well-thought-out — 3 focused packages with clear separation of concerns, comprehensive test coverage, and proper use of the extension SDK patterns. Looking forward to porting more extensions over to the azd official registry.

### ✅ What Looks Good
- Clean use of `azdext.Run()` with the new `WithExitCode` option — elegant solution for exit code propagation
- Thorough cmd.exe injection tests covering CWE-78 and CWE-93 edge cases
- `errors.AsType[*T]()` used consistently throughout (Go 1.26 convention)
- Security hardening flags in build scripts (`-trimpath`, `-buildmode=pie`, `-tags=cfi,cfg,osusergo`)
- `requiredAzdVersion` properly declared
- Good error type hierarchy enabling programmatic handling

### Summary
| Priority | Count |
|----------|-------|
| Critical | 0 |
| High | 0 |
| Medium | 2 |
| Low | 3 |
| **Total** | **5** |

**Overall Assessment:** Approve — well-structured extension with no blocking issues. See inline comments for minor suggestions.

cli/azd/extensions/microsoft.azd.exec/internal/cmd/root.go
jongio: Good catch — the stat-first pattern was silently swallowing non-NotExist errors. Switched to the Execute-first → `ScriptNotFoundError` fallback pattern as suggested. This also eliminates the duplicate `filepath.Abs` + `os.Stat`.

cli/azd/extensions/microsoft.azd.exec/internal/executor/command_builder.go
jongio: Fixed — added `\t` to the `ContainsAny` check. Also added a test case for tab character quoting.

cli/azd/pkg/azdext/run.go
jongio: Added a defensive `code != 0` guard in `Run()` so `(0, true)` falls through to the default `exit(1)`. Also documented the contract in the godoc. Added a test (`TestWithExitCode_ZeroCodeTreatedAsUnmatched`) to cover this.

cli/azd/pkg/exec/scripting/command_builder.go
jongio: Hoisted to `var controlCharReplacer = strings.NewReplacer(...)` at package level.

cli/azd/extensions/microsoft.azd.exec/internal/executor/command_builder.go
jongio: Added a NOTE comment to `quoteCmdArg` documenting the inherent `%VAR%` expansion limitation.

alexwolfmsft (approved):

JeffreyCA:
Looks like there's still some open discussion on https://github.com/Azure/azure-dev/issues/7423 to have this as a core feature instead of an extension

jongio:
> Looks like there's still some open discussion on #7423 to have this as a core feature instead of an extension

Thanks for flagging - yes, we're actively discussing the right home for this in #7423. For now this PR moves forward as an extension, but we're open to revisiting based on that discussion.

weikanglim:
@jongio

I think the cleanest direction here is to make **direct process execution the default contract**, and make **`--shell <shell>` an explicit opt-in** for shell-mediated behavior.

Concretely, I’d prefer a v1 shape along these lines:

- **Default:** exact-argv child process execution only  
  `azd env exec -- <command> [args...]`
- **Explicit shell mode:** only when the user asks for it  
  `azd env exec --shell pwsh -- <script-or-command>`
- Keep shell conveniences (inline command strings, shell-specific parsing, extension-based script dispatch) **out of the default path**
- If we support a Windows shell initially, I would strongly prefer **`pwsh`/`powershell` only**, not `cmd`

The reason I’m leaning this way is that it gives us a much clearer and more durable contract: **“run this process with azd env injected, preserve argv exactly, return the child exit code, do not mutate the parent shell.”** That seems to match the common scenarios here (`python`, `node`, `uv`, `npm`, `docker compose`, etc.), and it also lines up with prior art from tools like `dotenvx run -- ...`, `doppler run -- ...`, `aws-vault exec ... -- ...`, and `op run ...`, which are primarily **child-process wrappers**. When tools want shell-string behavior, the API is usually explicit (`npm exec -c`, `bash -c`, `pwsh -Command`) rather than heuristic.

My concern with the current mixed model is not that it can’t be made to work, but that it makes the command contract harder to explain and reason about, which also applies to agents:

1. **Flag / argv parsing is subtle.**  
   The current implementation does make `azd exec python --version` work by stopping parsing after the first positional argument and passing later tokens through. But that also means `azd exec python --debug` now sends `--debug` to Python, not azd. That behavior is deterministic, but subtle enough that I think `--` should be the canonical passthrough form in docs/examples, and ideally the command surface should make that boundary more explicit.

2. **Inline shell commands are not truly cross-platform.**  
   `azd exec 'echo $AZURE_ENV_NAME'` is Unix-shell syntax. On Windows, the default inline shell is PowerShell, which uses `$env:AZURE_ENV_NAME`, and the caller’s outer shell quoting rules differ before `azd` even sees the argument. So “works on all platforms” and “portable command syntax” are not the same thing here.

3. **Supporting `cmd` opens a large surface area.**  
   The PR already needs custom quoting, control-character stripping, and a raw command-line override for `cmd.exe`, and even then `%VAR%` expansion inside quotes remains an inherent limitation. That feels like a lot of semantic and security complexity for a mode we probably do not want to encourage.

4. **File detection by extension is a risky dispatch rule.**  
   Today the implementation tries file execution first, then chooses a shell from the file extension, with unknown extensions falling back to the platform default shell. That can override normal OS/shebang behavior for extensionless scripts or local executables. I would rather the default path preserve OS/process semantics first, and only do shell-specific dispatch when the user explicitly asked for a shell.

5. **The heuristic introduces command asymmetry.**  
   `azd exec python` becomes shell-inline, while `azd exec python script.py` becomes direct exec. Existing-file paths route differently again. So the mode depends on arity and filesystem state rather than explicit user intent, which makes the feature harder to teach and predict.

So my objective take is: **the underlying capability is worth doing**, but the **mixed heuristic design** is the weak part. If this graduates into a core command, I think the safest and clearest v1 is a narrow **process-exec primitive**, with `--shell <shell>` as an explicit opt-in for shell behavior rather than something inferred from argument shape or file extension.

wbreza (approved):
## Code Re-Review — PR #7400

**azd exec — run commands and scripts with azd environment context** by @jongio

### Re-review after force-push

Changes since my Apr 3 approval (`aa57e4e`) are trivial:
- **Commit 2** (`b82c41e`): Design specification document (docs only)
- **Commit 3** (`da7cd93`): Go version bump 1.26.0→1.26.1, cspell config tweaks (4 lines)

No code logic changes. All 7 findings from my initial review remain addressed.

### Prior Review: All ✅ Resolved
| 1 | Medium | Shell validation bypass via empty `--shell ""` | ✅ |
| 2 | Medium | `GOWORK=off` in test instructions | ✅ |
| 3-7 | Low | Various minor items | ✅ |

### ✅ What Looks Good
- Clean architecture: 3 focused packages, no circular deps, 94% test coverage
- Proper use of `azdext.Run()` with `WithExitCode` for exit code propagation
- Comprehensive cmd.exe injection tests (CWE-78, CWE-93)
- `errors.AsType[*T]()` used consistently (Go 1.26)
- Security hardening in build scripts

**Overall Assessment:** Re-approve — no logic changes since initial approval.

jongio:
Thanks for the detailed write-up. I've gone through each point carefully against the implementation and the DevX tradeoffs.

**On argv parsing (#1):** The implementation uses Cobra's `SetInterspersed(false)`, which stops flag parsing at the first positional arg. So `azd exec python --debug` deterministically sends `--debug` to Python - same pattern as `docker exec`. The collision surface is just three flags (`--shell`, `-e`, `-i`), so the chance of a real conflict is negligible. `--` is supported for anyone who wants the explicit boundary, but I don't think it needs to be the canonical form in docs.

**On inline shell cross-platform (#2):** `echo $VAR` is bash, `Write-Host $env:VAR` is PowerShell - that's how shells work. The extension delivers "your platform's shell with azd env vars injected," which IS the cross-platform contract. Removing inline mode would force script files for simple one-liners like `azd exec 'curl $AZURE_ENDPOINT/health'` - that's a real ergonomics regression for the most common use case.

**On cmd.exe surface area (#3):** You're right that cmd.exe needs more handling than other shells. The implementation includes `quoteCmdArg()` for metacharacter safety, control-character stripping (CWE-93), and a `CmdLine` override to bypass Go's `CommandLineToArgvW` escaping. That said, this complexity is security hardening - without it, cmd.exe would be vulnerable to injection. The test suite covers CWE-78 and CWE-93 edge cases extensively. I don't think "it requires careful security handling" is a reason to drop it - it's a reason to test it well, which we have (94% coverage). Not supporting cmd on Windows would leave a gap for enterprise users with existing `.bat`/`.cmd` scripts.

**On file extension detection (#4):** The detection chain (known extension -> shebang -> OS default) matches how every IDE, editor, and task runner handles this. `.sh` runs in bash, `.ps1` runs in PowerShell - that's what users expect. Requiring `--shell bash ./setup.sh` adds ceremony for zero practical benefit.

**On heuristic asymmetry (#5):** The file-exists check is simple and predictable: if the first arg resolves to a file, run it as a script; otherwise execute as a command. The edge case where your CWD contains a file named `python` is vanishingly rare. The detection priority is worth documenting clearly, which I'll do.

**On the overall shape:** The tools you cite (dotenvx, doppler, op) are env-injection wrappers only. azd exec does more because the azd ecosystem spans multiple platforms and shell types. The heuristic optimizes for the common case - `azd exec python script.py` just works - while keeping explicit control available for anyone who wants it. Requiring `-- <command>` on every invocation is a tax on the majority to solve for a theoretical minority case.

**Actions I'll take from this:**
- Document the detection priority clearly (explicit `--shell` flag -> file detection -> inline)
- Ensure the README and examples make the behavior unambiguous

JeffreyCA:
Adding some thoughts: I generally agree with @weikanglim's comments in #7423 on `exec` being better suited as a core feature **from the start** rather than an extension since it's more of a core primitive feature rather than an add-on capability.

### Discoverability

If users don't have the extension installed, `exec` doesn't appear in azd's help text. Given that this addresses long-standing asks (#391, #1697, #2336, #4067, #4384), the feature should be visible where users already look, in addition to public docs, blog posts, etc.

As a core command, we could also consider printing console messages or suggestion texts in various places like after `azd up` to have users try out the feature.

### Iteration speed

The iteration speed benefit of extensions is not as big now that we have weekly core releases.

### Extension to core graduation path

If `exec` starts out as an extension, there are more unknowns later down the road as we think about graduating the feature into core, like:

- Handling command execution and migration for users who have the extension installed when the core command lands
- Breaking change concerns if we end up renaming `azd exec` to something else like `azd env exec`

jongio:
We have decided to make this part of core, not an azd extension.  I will be updating this PR with that new direction.

hemarina:
## Code Review

Great work on this PR — the architecture is clean, test coverage is strong (94.3%), and the execution mode heuristics are well-thought-out. I reviewed the full diff and found one minor item not yet covered by existing review threads:

### [Low] `proctree_windows.go` — Silent Job Object fallback

When Job Object setup fails (`CreateJobObject`, `SetInformationJobObject`, or `AssignProcessToJobObject`), the code silently falls back to `Process.Kill()` without any diagnostic logging:

```go
handle, err := windows.CreateJobObject(nil, nil)
if err != nil {
    return fallbackKill, nil // no log, no signal to caller
}
```

I understand this is intentional best-effort behavior, but when process tree cleanup silently degrades, it makes it harder to diagnose orphaned child processes. Would it be worth adding `log.Printf` at each fallback point? Something like:

```go
if err != nil {
    log.Printf("scripting: Job Object setup failed, falling back to direct kill: %s", err)
    return fallbackKill, nil
}
```

This would help operators troubleshoot without changing any behavior.

jongio:
Good catch - added `log.Printf` at all four Job Object fallback points in `proctree_windows.go`. Each log message identifies which step failed (creation, configuration, process open, or assignment) so orphaned child process issues are diagnosable. No behavioral changes - still best-effort with graceful fallback.

hemarina:
LGTM. Nice addition — `azd exec` fills a long-standing gap (#391, #1697, #2336, #4067) with a clean design.

Highlights:
- Well-factored `scripting` package (executor / command_builder / shell / errors / proctree)
- Solid cross-platform process tree management — Windows Job Objects with `KILL_ON_JOB_CLOSE`, Unix `Setpgid` + `SIGKILL`
- `looksLikeFilePath` safety guard prevents unsafe inline fallback on mistyped script paths
- Structured error types (`ValidationError`, `ScriptNotFoundError`, `InvalidShellError`, `ExecutionError`) with proper `Unwrap()`
- Exit code propagation via `ExitCodeError` — good pattern for CI/CD use
- 94%+ test coverage with platform-aware table-driven tests

hemarina (approved):

cli/azd/pkg/exec/scripting/command_builder.go
```diff
@@ -0,0 +1,121 @@
...
+		} else {
+			cmdArgs = []string{shellBin, "-File", scriptOrPath}
+		}
+	case "cmd":
+		useCmdLineOverride = true
+		if isInline {
+			cmdArgs = []string{shellBin, "/c", scriptOrPath}
+			cmdWrapOuter = false
+		} else {
+			escaped := strings.ReplaceAll(scriptOrPath, `"`, `""`)
```
wbreza, line 49: **[🔴 HIGH — security] cmd.exe `%VAR%` expansion bypasses validated file path.** `strings.ReplaceAll(scriptOrPath, "\"", "\"\"")` correctly neutralizes `"`/`&`/`|`/`<`/`>`/`^`, but does NOT handle `%`. cmd.exe expands `%VAR%` before executing, so a script placed at `...\a%EVIL%b\hook.bat` that passes the `os.Stat` check will cause cmd.exe to execute a different path. Reproduced empirically on Windows — `%COMSPEC%` in a path caused cmd.exe to spawn directly instead of the intended file. Same class as **CVE-2024-24576** (Rust std). Attackers who can influence any env var (CI config, parent env poisoning) plus write one file can redirect hook execution. Fix: hard-reject `%` in `scriptOrPath` and cmd args; for `.exe` files use direct `CreateProcess` instead of `cmd /c`. Add a test matrix for `%`, CR, LF, VT, NUL in file names.

cli/azd/pkg/exec/scripting/executor.go
```diff
@@ -0,0 +1,241 @@
...
+
+	workingDir, err := os.Getwd()
+	if err != nil {
+		return fmt.Errorf("failed to get working directory: %w", err)
+	}
+
+	cmd := exec.CommandContext(ctx, command, args...) //nolint:gosec
+	cmd.Dir = workingDir
+	cmd.Env = e.childEnv()
+	cmd.Stdin = os.Stdin
```
wbreza, line 100: **[Medium] `ExecuteDirect` ignores the `-i` flag.** `cmd.Stdin = os.Stdin` is unconditional here, but `executeCommand` (line 146) correctly gates on `e.config.Interactive`. Result: `azd exec ./build.sh` detaches stdin (safe for CI), but `azd exec bash build.sh` — same script, user mental model identical — runs with stdin permanently open and hangs in non-interactive contexts. Apply the same `if e.config.Interactive` guard here.

cli/azd/pkg/exec/scripting/executor.go
```diff
@@ -0,0 +1,241 @@
...
+	}
+	if info.IsDir() {
+		return &ValidationError{
+			Field: "scriptPath", Reason: "must be a file, not a directory",
+		}
+	}
+
+	shell := e.config.Shell
+	if shell == "" {
+		shell = DetectShellFromFile(absPath)
```
wbreza, line 78: **[Medium] Extensionless file in cwd silently dispatched via default shell.** `Execute()` only checks `os.Stat` before falling to `DetectShellFromFile` → `DefaultShell()`. A user running `azd exec mybinary arg1 arg2` where any file named `mybinary` exists (ELF binary, Makefile, etc.) gets `bash mybinary arg1 arg2` with confusing "cannot execute binary file" errors. The F15 fallback guard (`looksLikeFilePath`) never fires because it runs AFTER `Execute()` returns `ScriptNotFoundError`, which it never does when the file exists. Fix: require a known script extension OR explicit `--shell` before committing to shell dispatch; otherwise return `ScriptNotFoundError` and let the caller fall through to direct exec.

cli/azd/pkg/exec/scripting/proctree_windows.go
```diff
@@ -0,0 +1,90 @@
...
+	if err := windows.AssignProcessToJobObject(handle, proc); err != nil {
+		log.Printf("scripting: Job Object process assignment failed, falling back to direct kill: %s", err)
+		_ = windows.CloseHandle(proc)
+		_ = windows.CloseHandle(handle)
+		return fallbackKill, nil
+	}
+	_ = windows.CloseHandle(proc)
+
+	return func() {
+		if err := windows.TerminateJobObject(handle, 1); err != nil {
```
wbreza, line 86: **[Medium] Job Object handle leaked on success path + fallback doesn`t tree-kill.** Two linked issues: (1) **Success path**: `handle` is captured only by the kill closure; when `cmd.Wait()` returns normally and the goroutine exits via `close(done)`, `CloseHandle` is never called. `JOB_OBJECT_LIMIT_KILL_ON_JOB_CLOSE` safety net never fires, so this also needs `CloseHandle(handle)` after `TerminateJobObject` in the kill closure. (2) **Fallback path** (see line 35): if Job Object setup fails, `fallbackKill` only calls `cmd.Process.Kill()` — grandchildren survive, directly contradicting F6. On Windows, `CREATE_NEW_PROCESS_GROUP` alone does NOT kill descendants; only the Job Object does. Fix: return `(kill, close)` pair from `startProcessTree`; `CloseHandle(handle)` on both kill and success exits; either make Job Object setup errors fatal, or walk the tree via `CreateToolhelp32Snapshot`/`Process32Next` on fallback.

cli/azd/pkg/exec/scripting/proctree_windows.go
```diff
@@ -0,0 +1,90 @@
...
+		false,
+		uint32(cmd.Process.Pid),
+	)
+	if err != nil {
+		log.Printf("scripting: Job Object process open failed, falling back to direct kill: %s", err)
+		_ = windows.CloseHandle(handle)
+		return fallbackKill, nil
+	}
+
+	if err := windows.AssignProcessToJobObject(handle, proc); err != nil {
```
wbreza, line 77: **[Medium] Race: grandchildren spawned before `AssignProcessToJobObject`.** `cmd.Start()` was called earlier, so the child process is already running by the time we reach this line. A fast-spawning script like `cmd /c start malicious.exe & real.exe` can create grandchildren before assignment completes; those grandchildren are not in the Job Object and survive `TerminateJobObject`. Narrow window (μs–ms) but real, and defeats the tree-kill guarantee for adversarial or just fast-starting children. Fix: start with `CREATE_SUSPENDED`, assign to the job, then `ResumeThread`. Requires bypassing `os/exec.Start`. At minimum, document the race.

cli/azd/pkg/exec/scripting/proctree_windows.go
```diff
@@ -0,0 +1,90 @@
...
+	"golang.org/x/sys/windows"
+)
+
+// setupProcessTree merges CREATE_NEW_PROCESS_GROUP into the command's
+// SysProcAttr (preserving CmdLine if already set by setCmdLineOverride).
+func setupProcessTree(cmd *exec.Cmd, _ bool) {
+	if cmd.SysProcAttr == nil {
+		cmd.SysProcAttr = &syscall.SysProcAttr{}
+	}
+	cmd.SysProcAttr.CreationFlags |= syscall.CREATE_NEW_PROCESS_GROUP
```
wbreza, line 23: **[Medium] `CREATE_NEW_PROCESS_GROUP` breaks Ctrl-C in interactive mode.** The Unix variant (`proctree_unix.go`) explicitly skips `Setpgid` when `interactive=true`, documented as "Setpgid breaks stdin forwarding". The Windows variant here ignores the `interactive` parameter and always sets `CREATE_NEW_PROCESS_GROUP`, which per MSDN disables `CTRL_C_EVENT` delivery to the child group from the console. Interactive hooks (`azd exec -i ./dev-server.sh` with a PowerShell/cmd prompt) won`t receive Ctrl-C. Fix: skip the flag when `interactive=true`; rely on `TerminateJobObject` for cleanup. Add a test that sends `CTRL_C_EVENT` and verifies child receipt.

cli/azd/pkg/exec/scripting/command_builder.go
```diff
@@ -0,0 +1,121 @@
...
+	return strings.Map(func(r rune) rune {
+		if r < 0x20 || r == 0x7F {
+			return -1
+		}
+		return r
+	}, s)
+}
+
+// quoteCmdArg quotes a single argument for cmd.exe.
+func quoteCmdArg(arg string) string {
```
wbreza, line 91: **[Low] `quoteCmdArg` silently strips control chars; `%` still passes through.** Control char stripping is good for injection, but silently rewriting `a\tb` → `ab` is surprising (users passing real tabs have them vanish with no warning). More importantly, `%` is NOT rejected here — same root cause as finding #1 but for args passed to cmd-shell scripts. If a hook takes user input and forwards it to a `.bat`, attacker-supplied `%COMSPEC%` still expands. Fix: error rather than silently strip, and reject `%`.

cli/azd/cmd/exec_test.go
```diff
@@ -0,0 +1,333 @@
...
+	}
+
+	childEnv, err := action.buildChildEnv(t.Context())
+	require.NoError(t, err)
+
+	envMap := envSliceToMap(childEnv)
+	assert.Equal(t, "kv-resolved-value", envMap[secretKey])
+}
+
+func TestExecAction_ExitCodePropagation(t *testing.T) {
```
wbreza, line 257: **[Low] Exit-code propagation test is effectively a no-op on Windows.** The test uses `errors.As` (not `require.ErrorAs`) and has a comment accepting "error may be of a different type — that`s acceptable" on Windows. Net effect: on Windows CI, the test only asserts that *some* error is returned — the actual exit-code value is never verified. Exit-code propagation is the feature`s key CI/CD selling point; leaving it untested on the platform most prone to exit-code quirks undermines confidence. Fix: use `go run` to invoke a tiny program that calls `os.Exit(42)` (platform-portable, no bash dependency); use `require.ErrorAs` so the test fails rather than silently passes when the type doesn`t match.

wbreza (changes requested):
**Re-review since 2026-04-07 approval** — the extension was restructured into core `azd exec` on 2026-04-10, making this effectively a new PR since my last review. Multi-model review (Claude Opus 4.7 + Claude Sonnet 4.6) found 9 new findings — **1 high (security-blocking), 5 medium, 3 low**.

Prior reviews are superseded for the current code: extension-framework concerns (main.go bypassing azdext.Run, requiredAzdVersion, etc.) are moot since the code moved to core; Copilot comments on `keyvault.go` are moot since that file is no longer in this PR.

Requesting changes on the basis of finding #1 — a **reproduced Windows command-injection** via `%VAR%` expansion, same class as CVE-2024-24576.

## 🔴 High — cmd.exe `%VAR%` expansion bypasses validated file path

The F5 escape handles `"`/`&`/`|`/`<`/`>`/`^` correctly, but **not** `%`. cmd.exe expands `%VAR%` before executing, so a script at `...\a%EVIL%b\hook.bat` that passes `os.Stat` validation causes cmd.exe to execute a different path after expansion. This is the same class as CVE-2024-24576 (Rust std). The reviewing agent reproduced this empirically on Windows — placing a file at a path containing `%COMSPEC%` caused cmd.exe to spawn directly instead of running the intended file. Attackers who can influence any env var (CI, parent env poisoning) and write one file can redirect execution.

**Fix:** Hard-reject `%` in `scriptOrPath` and cmd args. For `.exe` files, use direct `CreateProcess` instead of `cmd /c`. Add a test matrix covering `%`, CR, LF, VT, and NUL in file names.

## 🟡 Medium

2. **`ExecuteDirect` ignores `-i` flag** — `cmd.Stdin = os.Stdin` is unconditional; `executeCommand` correctly gates on `Interactive`. `azd exec ./build.sh` detaches stdin; `azd exec bash build.sh` hangs on it. Same guard needed.
3. **Extensionless file in cwd silently runs via shell** — `Execute()` only checks `os.Stat`; `azd exec mybinary arg1 arg2` where `mybinary` exists becomes `bash mybinary arg1 arg2`, producing confusing errors. Require known script extension OR explicit `--shell` before committing to shell dispatch.
4. **Windows Job Object: handle leak on success path + fallback doesn`t tree-kill** — `handle` is never closed on normal exit, so `JOB_OBJECT_LIMIT_KILL_ON_JOB_CLOSE` safety net never fires. When Job Object setup fails, `fallbackKill` only calls `cmd.Process.Kill()` — grandchildren survive. Return `(kill, close)` pair; make setup errors fatal or walk the tree via `CreateToolhelp32Snapshot`.
5. **Race: grandchildren spawned before `AssignProcessToJobObject`** — `cmd.Start()` runs before assignment. A fast `cmd /c start malicious.exe & real.exe` can escape the job. Fix: `CREATE_SUSPENDED` → assign → `ResumeThread`. At minimum, document.
6. **`CREATE_NEW_PROCESS_GROUP` breaks Ctrl-C in interactive mode** — Unix variant skips `Setpgid` when `interactive=true`; Windows variant ignores the parameter and always sets the flag, which per MSDN disables `CTRL_C_EVENT` delivery. Mirror the Unix asymmetry.

## ⚪ Low

7. **`quoteCmdArg` silently strips control chars; `%` still passes** — surprising rewrite (`a\tb` → `ab`); reject rather than strip, and reject `%`.
8. **Exit-code propagation test is a no-op on Windows** — uses `errors.As` (not `require.ErrorAs`) with "error may be a different type" acceptance comment; on Windows CI, only asserts *some* error. Use `go run` to invoke a tiny `os.Exit(42)` program; `require.ErrorAs`.

## ✅ Verified correct

- `errors.go` Is/As/AsType usage
- `shell.go` IsSupportedShell gates before `exec.Command`
- `proctree_unix.go` Setpgid + SIGKILL; interactive guard documented
- F1/F4/F7 buildChildEnv — no leak into parent or telemetry subprocesses
- F12 stripControlChars — correctly handles 0x00-0x1F, 0x7F
- F15 inline-fallback path rejection — correctly handles typos
- F8 ExitCodeError {ExitCode:0} fallthrough guard

---
*Generated by GitHub Copilot CLI code-review workflow using parallel reviews across two models.*

cli/azd/pkg/exec/scripting/command_builder.go
jongio: Fixed in a1d8ecf. Added `%%` escaping in `quoteCmdArg` and in the path-building logic in `buildCmdCommand`. Both `%PATH%` in arguments and `%TEMP%` in file paths are now escaped before reaching cmd.exe. Same vulnerability class as CVE-2024-24576 (Rust std) - good catch.

cli/azd/pkg/exec/scripting/executor.go
jongio: Fixed in a1d8ecf. Gated stdin assignment on \.config.Interactive\ in \ExecuteDirect\, matching the existing pattern in \xecuteCommand\. Non-interactive direct exec now won't block on stdin.

cli/azd/pkg/exec/scripting/executor.go
jongio: I don't think this scenario can actually happen as described. The dispatch logic in exec.go (line 168) only falls through to ExecuteDirect when Execute() returns a ScriptNotFoundError AND there are additional args AND no explicit shell. If a file exists (extensionless or not), Execute() won't return ScriptNotFoundError - it'll attempt to dispatch it.

The guard at line 136 (looksLikeFilePath) prevents inline fallback for path-like inputs. If the file doesn't exist, ScriptNotFoundError triggers direct exec which is correct (it's probably a binary). If it does exist but has no extension, DetectShellFromFile falls back to DefaultShell() which is reasonable.

I can add a warning log when extension detection fails, but I don't think this is a correctness bug.

cli/azd/pkg/exec/scripting/proctree_windows.go
jongio: Partially addressed in a1d8ecf. Added CloseHandle(handle) after TerminateJobObject in the kill closure - that was a valid leak.

However, the success-path claim isn't right. On success, the handle IS closed: the kill closure captures the handle and the process's Wait() cleanup triggers deferred cleanup. The Job Object auto-terminates children when its last handle closes (JOB_OBJECT_LIMIT_KILL_ON_JOB_CLOSE), so explicit cleanup on normal exit is handled by the OS.

The fallback concern (process exits before AssignProcessToJobObject) doesn't leak either - TerminateJobObject on a job with no processes is a no-op, and we now CloseHandle afterward.

cli/azd/pkg/exec/scripting/proctree_windows.go
jongio: This is the standard Go pattern for Job Objects on Windows. Go's os/exec doesn't expose CREATE_SUSPENDED, so you can't assign to a job before the process runs. Every Go process-tree library (e.g., graceful, jobject) uses Start() then AssignProcessToJobObject.

The window between Start() and assignment is tiny (microseconds). If a grandchild spawns in that window, it won't be in the job - but that's an accepted tradeoff across the Go ecosystem. The alternative (CGo with CreateProcess+CREATE_SUSPENDED+ResumeThread) adds significant complexity for a near-zero probability event.

JOB_OBJECT_LIMIT_KILL_ON_JOB_CLOSE ensures any process that IS in the job gets cleaned up.

cli/azd/pkg/exec/scripting/proctree_windows.go
jongio: Fixed in a1d8ecf. setupProcessTree now respects the interactive parameter - skips CREATE_NEW_PROCESS_GROUP when interactive=true, so Ctrl-C propagates naturally to the child. This matches the Unix side which already skips Setpgid when interactive.

cli/azd/pkg/exec/scripting/command_builder.go
jongio: Fixed in a1d8ecf (same commit as thread 1). quoteCmdArg now escapes % as %% before quoting. This covers both the path and argument vectors.

cli/azd/cmd/exec_test.go
jongio: Fixed in a1d8ecf. Rewrote the test to compile a Go program (testdata/exit42.go) into a temp binary and invoke it via scripting.ExecuteDirect. This verifies actual exit code propagation (expects exit code 42) and works on all platforms.

The original test used runtime.GOOS to skip on Windows, which made it a no-op there. The new approach avoids shell dispatch entirely - it's a direct exec of a compiled binary, so it tests the real code path.

wbreza (commented):
## Code Review — zd exec

**PR #7400** adds zd exec, a cross-platform command/script execution engine that runs programs with full azd environment context (env vars, Key Vault secret resolution), implemented as a core command with a new pkg/exec/scripting package.

### 🔴 Must Fix

**1. Windows Command Injection via Shell Metacharacters**
- **File:** cli/azd/pkg/exec/scripting/command_builder.go (Windows path)
- **Issue:** When cmd.exe /c is used, arguments are joined into a single string. Shell metacharacters (&, |, (, )) in user-provided arguments can break out and execute arbitrary commands. E.g., zd exec "echo test & whoami" executes both commands. The inline cmd.exe path also doesn't escape % characters (unlike the script file path), allowing %VAR% environment variable expansion.
- **Contrast:** The Unix path properly uses positional parameters ("", "") preventing injection.
- **Suggested fix:** Escape shell metacharacters on Windows inline path, matching the existing script file path behavior.

**2. PR Description / Implementation Mismatch**
- **Issue:** The PR description documents an *extension* at cli/azd/extensions/microsoft.azd.exec/ with build scripts, pipelines, extension.yaml, and 94.3% test coverage across extension packages. The actual diff shows a *core command* at cli/azd/cmd/exec.go with pkg/exec/scripting/. The design note mentions moving to core, but the description body was never updated.
- **Suggested fix:** Update the PR description to accurately describe the core command implementation.

### 🟡 Should Fix

**3. Secret Exposure to Subprocesses**
- **File:** cli/azd/cmd/exec.go (uildChildEnv)
- Resolved Key Vault secrets are passed to child processes. While secrets aren't leaked to the parent process via os.Setenv (good), any child command can dump them with nv or set. Consider documenting this risk or providing an opt-in mechanism for secret resolution.

**4. Exit Code Propagation Complexity**
- Exit codes go through a multi-layer wrap/unwrap chain: xec.ExitError → scripting.ExecutionError → internal.ExitCodeError. This works correctly (validated by TestExecAction_ExitCodePropagation), but ensure this test runs cross-platform in CI.

**5. Missing Windows %VAR% Expansion Test**
- **File:** cli/azd/cmd/exec_test.go
- No test verifies that %WINDIR% or similar Windows-style references are not unexpectedly expanded in inline cmd.exe commands. This is the test-side companion to Finding #1.

**6. Shell Detection Test Coverage**
- Shell auto-detection from file extensions (.sh → bash, .ps1 → pwsh) should have explicit tests for all supported shells on each platform, including fallback behavior when a shell isn't available.

### 🟢 Nitpick

**7. Scope Clarity** — Several foundational changes (rrors.go, main.go, nvironment-variables.md) aren't called out in the PR description. A brief mention would help reviewers.

**8. cspell Dictionary** — stretchr addition is fine (testify package author), minor housekeeping.

---

**Overall:** Core architecture is sound — proper DI patterns, clean package separation in pkg/exec/scripting/, correct exit code propagation, and secrets isolated from the parent process. The Windows command injection (#1) is the key item to address.

wbreza (approved):
## Re-Review — zd exec (commit a1d8ecf1)

The latest commit directly addresses the blocking findings from my prior review. Security fix verified, interactive mode corrected, tests improved.

### Previously Flagged → Now Resolved

- ✅ **Windows `%VAR%` expansion (CVE-2024-24576 class)** — Fixed with `%` → `%%` escaping in both file paths (`buildCommand`) and arguments (`quoteCmdArg`). Test updated to verify `%%PATH%%` output.
- ✅ **Stdin gating** — `cmd.Stdin` now gated on `config.Interactive` with parity across both `ExecuteDirect` and `executeCommand` paths.
- ✅ **Ctrl-C in interactive mode** — `CREATE_NEW_PROCESS_GROUP` skipped when `interactive=true`, matching Unix `Setpgid` behavior. Ctrl-C now correctly reaches the child process.
- ✅ **Job Object handle leak** — `CloseHandle` added after `TerminateJobObject` in the kill closure.
- ✅ **Exit code test portability** — Replaced shell-dependent `exit 42` with a compiled Go binary (`testdata/exit42.go`), portable across all platforms.

### Residual Items (non-blocking)

- 🟡 **Delayed expansion `!VAR!`** — `!` characters are not escaped in `quoteCmdArg`. If cmd.exe has delayed expansion enabled (`SetLocal EnableDelayedExpansion`), `!VAR!` references would expand. Low practical risk (delayed expansion is off by default), but worth a follow-up.
- 🟡 **Exit code test coverage** — The updated test exercises `ExecuteDirect` (direct binary path) but not the full `execAction.Run()` → shell-dispatch path. Consider adding a companion test for shell-wrapped exit code propagation.
- 🟡 **PR description** — Still describes the extension architecture (`cli/azd/extensions/microsoft.azd.exec/`). Should be updated to reflect the core command implementation.

### Verdict

**Approve** — The blocking security vulnerability is resolved with correct escaping. Interactive mode, process tree management, and test portability are all properly addressed. Remaining items are improvements, not blockers.

wbreza (approved):
## Re-Approval — `azd exec` (commit 718e8c6b)

Single new commit since last approval: adds `//nolint:gosec // G204` to the test build command in `exec_test.go`. This is a standard lint suppression for a known test fixture — no functional or security impact.

Standing approval unchanged. Previously noted non-blocking items remain:
- Delayed expansion `!VAR!` unescaped (low practical risk — off by default)
- Exit code test covers `ExecuteDirect` only, not shell-dispatch path
- PR description still describes extension architecture

jongio:
## Review Comment Resolution Summary

Rebased onto current main (`063282153`) and verified all 8 unresolved review threads. Here's the resolution status:

### ✅ Confirmed Fixed

| Thread | Issue | Fix Commit |
|--------|-------|------------|
| 1 | cmd.exe `%VAR%` expansion in `quoteCmdArg` | `0bb3f04` — escapes `%` → `%%` via `strings.ReplaceAll` |
| 2 | `ExecuteDirect` ignores `-i` flag (stdin not gated) | `0bb3f04` — `cmd.Stdin = os.Stdin` gated on `Interactive` |
| 6 | `CREATE_NEW_PROCESS_GROUP` breaks Ctrl-C in interactive | `0bb3f04` — flag skipped when `interactive == true` |
| 7 | `quoteCmdArg` `%` passthrough | `0bb3f04` — same fix as thread 1 |
| 8 | Exit-code test no-op on Windows | `0bb3f04` — test compiles `exit42.go`, asserts exit code `42` |
| 4 | Job Object handle leak + silent fallback | `0bb3f04` + `2761a54` — handle closed on all paths, `log.Printf` at every fallback |

### ⚠️ Go stdlib limitation (unfixable)

| Thread | Issue | Status |
|--------|-------|--------|
| 5 | Race between `cmd.Start()` and `AssignProcessToJobObject` | Go `os/exec` doesn't expose `CREATE_SUSPENDED`. Standard pattern used by Docker/containerd. Grandchild spawned in the gap is unlikely and mitigated by `KILL_ON_JOB_CLOSE` on the parent. |

### 💬 Disagree — intentional behavior

| Thread | Issue | Status |
|--------|-------|--------|
| 3 | Extensionless file dispatch via default shell | This is intentional — unknown file types are dispatched through the platform default shell, matching the behavior of `./script` on Unix (which uses the shell/shebang). Users running ELF binaries should use `azd exec ./mybinary arg1` which routes through `ExecuteDirect`. |

cli/azd/pkg/exec/scripting/proctree_windows.go
jongio: Acknowledged — this is a known Go stdlib limitation. `os/exec` doesn't expose `CREATE_SUSPENDED`, so there's an inherent race between `cmd.Start()` and `AssignProcessToJobObject`. This is the standard pattern used by Docker, containerd, and other Go programs that use Job Objects. The practical risk is low: a grandchild would need to spawn in the microsecond gap, and even then `KILL_ON_JOB_CLOSE` on the parent job still terminates the parent process. Not fixable without CGo or a custom process launcher.

vhvb1989 (approved):

PR #7407, opened 2026-03-31
Add validate command for custom training

saanikaguptamicrosoft:
### Notes

Added validation for-
1. command field is required
2. environment field is required
3. compute field is required
4. code if present, it should not start with `git://` or `git+`
5. The local path mentioned in `code` and `inputs.<key>.path` are existent
6. Placeholder mapping is correct. Different cases:
   - The command field contains placeholder in the form of `${{inputs.xxx}}`, the `xxx` should be present as key in inputs.
       - For inputs -> if key is present but its value is `{}` or `None`, throw error. 
   - The command field contains placeholder in the form `${{outputs.xxx}}`, the `xxx` may or may not be present as key in inputs. If it's not present, means user wants to go with default output values. If it's present means user wants to override output configs.
      - For outputs -> if key is present but its value is `{}` or `None`, show warning. Because user might have forgotten to override the value. 
   - If `${{inputs.xxx}}` is wrapped inside `[ ]`, it's optional, so we don't validate its presence in inputs key.
   - Single bracket i.e. `{inputs.xxx}` or `{outputs.xxx}` or `${inputs.xxx}` or `${outputs.xxx}` should throw error

Added unit tests.

### Testing
- Local testing
image
image
image
image
image

- Build ~ Successful
```
azd x build
```

- UTs ~ Successful
```
go test ./internal/utils/ -v -count=1
```

### DIFF
--- a/cli/azd/extensions/azure.ai.customtraining/internal/cmd/job.go
+++ b/cli/azd/extensions/azure.ai.customtraining/internal/cmd/job.go
@@ -34,6 +34,7 @@ func newJobCommand() *cobra.Command {
 	cmd.AddCommand(newJobShowCommand())
 	cmd.AddCommand(newJobDeleteCommand())
 	cmd.AddCommand(newJobCancelCommand())
+	cmd.AddCommand(newJobValidateCommand())
 
 	return cmd
 }
--- a/cli/azd/extensions/azure.ai.customtraining/internal/cmd/job_validate.go
+++ b/cli/azd/extensions/azure.ai.customtraining/internal/cmd/job_validate.go
@@ -63,7 +63,6 @@ func newJobValidateCommand() *cobra.Command {
 			fmt.Printf("  Errors: %d, Warnings: %d\n", result.ErrorCount(), result.WarningCount())
 
 			if result.HasErrors() {
-				fmt.Printf("\n✗ Validation failed.\n")
 				return fmt.Errorf("validation failed with %d error(s)", result.ErrorCount())
 			}
 
--- a/cli/azd/extensions/azure.ai.customtraining/internal/cmd/job_validate.go
+++ b/cli/azd/extensions/azure.ai.customtraining/internal/cmd/job_validate.go
@@ -10,6 +10,7 @@ import (
 
 	"azure.ai.customtraining/internal/utils"
 
+	"github.com/fatih/color"
 	"github.com/spf13/cobra"
 	"gopkg.in/yaml.v3"
 )
@@ -45,7 +46,7 @@ func newJobValidateCommand() *cobra.Command {
 
 			// Print findings
 			if len(result.Findings) == 0 {
-				fmt.Printf("✓ Validation passed: %s\n", filePath)
+				color.Green("✓ Validation passed: %s\n", filePath)
 				return nil
 			}
 
@@ -66,7 +67,7 @@ func newJobValidateCommand() *cobra.Command {
 				return fmt.Errorf("validation failed with %d error(s)", result.ErrorCount())
 			}
 
-			fmt.Printf("\n✓ Validation passed with warnings.\n")
+			color.Green("\n✓ Validation passed with warnings.\n")
 			return nil
 		},
 	}
--- a/cli/azd/extensions/azure.ai.customtraining/internal/cmd/job_validate.go
+++ b/cli/azd/extensions/azure.ai.customtraining/internal/cmd/job_validate.go
@@ -19,12 +19,11 @@ func newJobValidateCommand() *cobra.Command {
 
 	cmd := &cobra.Command{
 		Use:   "validate",
-		Short: "Validate a job YAML definition file without submitting",
-		Long:  "Validate a job YAML definition file for schema correctness and surface issues early.\n\nExample:\n  azd ai training job validate --file job.yaml",
+		Short: "Validate a job YAML definition file offline without submitting",
 		Args:  cobra.NoArgs,
 		RunE: func(cmd *cobra.Command, args []string) error {
 			if filePath == "" {
-				return fmt.Errorf("--file (-f) is required: provide a path to a YAML job definition file")
+				return fmt.Errorf("--file is required: provide a path to a YAML job definition file")
 			}
 
 			// Read and parse the YAML file (without running ValidateJobDefinition which stops at first error)
--- a/cli/azd/extensions/azure.ai.customtraining/internal/cmd/job_validate.go
+++ b/cli/azd/extensions/azure.ai.customtraining/internal/cmd/job_validate.go
@@ -21,6 +21,7 @@ func newJobValidateCommand() *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "validate",
 		Short: "Validate a job YAML definition file offline without submitting",
+		Long:  "Validate a job YAML definition file offline without submitting.\n\nExample:\n  azd ai training job validate --file job.yaml",
 		// Override parent's PersistentPreRunE — validate is offline and needs no Azure setup.
 		PersistentPreRunE: func(_ *cobra.Command, _ []string) error { return nil },
 		Args: cobra.NoArgs,
--- a/cli/azd/extensions/azure.ai.customtraining/internal/cmd/job_validate.go
+++ b/cli/azd/extensions/azure.ai.customtraining/internal/cmd/job_validate.go
@@ -1,0 +1,75 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package cmd
+
+import (
+	"fmt"
+	"os"
+
+	"azure.ai.customtraining/internal/utils"
+
+	"github.com/spf13/cobra"
+	"gopkg.in/yaml.v3"
+)
+
+func newJobValidateCommand() *cobra.Command {
+	var filePath string
+
+	cmd := &cobra.Command{
+		Use:   "validate",
+		Short: "Validate a job YAML definition file without submitting",
+		Long:  "Validate a job YAML definition file for schema correctness and surface issues early.\n\nExample:\n  azd ai training job validate --file job.yaml",
+		Args:  cobra.NoArgs,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			if filePath == "" {
+				return fmt.Errorf("--file (-f) is required: provide a path to a YAML job definition file")
+			}
+
+			// Read and parse the YAML file (without running ValidateJobDefinition which stops at first error)
+			data, err := os.ReadFile(filePath)
+			if err != nil {
+				return fmt.Errorf("failed to read job file '%s': %w", filePath, err)
+			}
+
+			var jobDef utils.JobDefinition
+			if err := yaml.Unmarshal(data, &jobDef); err != nil {
+				return fmt.Errorf("failed to parse job YAML: %w", err)
+			}
+
+			// Run offline validation — collects all findings
+			result := utils.ValidateJobOffline(&jobDef)
+
+			// Print findings
+			if len(result.Findings) == 0 {
+				fmt.Printf("✓ Validation passed: %s\n", filePath)
+				return nil
+			}
+
+			fmt.Printf("Validation results for: %s\n\n", filePath)
+
+			for _, f := range result.Findings {
+				prefix := "⚠"
+				if f.Severity == utils.SeverityError {
+					prefix = "✗"
+				}
+				fmt.Printf("  %s [%s] %s: %s\n", prefix, f.Severity, f.Field, f.Message)
+			}
+
+			fmt.Println()
+			fmt.Printf("  Errors: %d, Warnings: %d\n", result.ErrorCount(), result.WarningCount())
+
+			if result.HasErrors() {
+				fmt.Printf("\n✗ Validation failed.\n")
+				return fmt.Errorf("validation failed with %d error(s)", result.ErrorCount())
+			}
+
+			fmt.Printf("\n✓ Validation passed with warnings.\n")
+			return nil
+		},
+	}
+
+	cmd.Flags().StringVar(&filePath, "file", "", "Path to YAML job definition file (required)")
+
+	return cmd
+}
--- a/cli/azd/extensions/azure.ai.customtraining/internal/cmd/job_validate.go
+++ b/cli/azd/extensions/azure.ai.customtraining/internal/cmd/job_validate.go
@@ -26,7 +26,7 @@ func newJobValidateCommand() *cobra.Command {
 				return fmt.Errorf("--file is required: provide a path to a YAML job definition file")
 			}
 
-			// Read and parse the YAML file (without running ValidateJobDefinition which stops at first error)
+			// Read and parse the YAML file
 			data, err := os.ReadFile(filePath)
 			if err != nil {
 				return fmt.Errorf("failed to read job file '%s': %w", filePath, err)
--- a/cli/azd/extensions/azure.ai.customtraining/internal/cmd/job_validate.go
+++ b/cli/azd/extensions/azure.ai.customtraining/internal/cmd/job_validate.go
@@ -72,7 +72,7 @@ func newJobValidateCommand() *cobra.Command {
 		},
 	}
 
-	cmd.Flags().StringVar(&filePath, "file", "", "Path to YAML job definition file (required)")
+	cmd.Flags().StringVarP(&filePath, "file", "f", "", "Path to YAML job definition file (required)")
 
 	return cmd
 }
--- a/cli/azd/extensions/azure.ai.customtraining/internal/cmd/job_validate.go
+++ b/cli/azd/extensions/azure.ai.customtraining/internal/cmd/job_validate.go
@@ -20,7 +20,9 @@ func newJobValidateCommand() *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "validate",
 		Short: "Validate a job YAML definition file offline without submitting",
-		Args:  cobra.NoArgs,
+		// Override parent's PersistentPreRunE — validate is offline and needs no Azure setup.
+		PersistentPreRunE: func(_ *cobra.Command, _ []string) error { return nil },
+		Args: cobra.NoArgs,
 		RunE: func(cmd *cobra.Command, args []string) error {
 			if filePath == "" {
 				return fmt.Errorf("--file is required: provide a path to a YAML job definition file")
--- a/cli/azd/extensions/azure.ai.customtraining/internal/cmd/job_validate.go
+++ b/cli/azd/extensions/azure.ai.customtraining/internal/cmd/job_validate.go
@@ -6,6 +6,7 @@ package cmd
 import (
 	"fmt"
 	"os"
+	"path/filepath"
 
 	"azure.ai.customtraining/internal/utils"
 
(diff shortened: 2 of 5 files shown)


cli/azd/extensions/azure.ai.customtraining/internal/cmd/job.go
```diff
@@ -34,6 +34,7 @@ func newJobCommand() *cobra.Command {
 	cmd.AddCommand(newJobShowCommand())
 	cmd.AddCommand(newJobDeleteCommand())
 	cmd.AddCommand(newJobCancelCommand())
+	cmd.AddCommand(newJobValidateCommand())
```
jongio, line 37: **[HIGH]** Design: `validate` isn't actually offline - the parent `job` command has `PersistentPreRunE` calling `validateOrInitEnvironment`, which requires an azd environment or `--subscription`/`--project-endpoint` flags. Running `azd ai training job validate --file job.yaml` without Azure setup will error before reaching validation logic.

Fix: in `job_validate.go`, add `PersistentPreRunE: func(_ *cobra.Command, _ []string) error { return nil }` to the cobra.Command definition. This overrides the parent's pre-run for the validate subcommand only.

cli/azd/extensions/azure.ai.customtraining/internal/utils/job_validator.go
```diff
@@ -0,0 +1,272 @@
...
+			result.Findings = append(result.Findings, ValidationFinding{
+				Field:    "code",
+				Severity: SeverityError,
+				Message:  fmt.Sprintf("git paths are not supported for 'code': '%s'. Use a local path instead", job.Code),
+			})
+		}
+	}
+
+	// 5. Local path existence checks
+	validateLocalPath(result, "code", job.Code, yamlDir)
```
jongio, line 104: **[MEDIUM]** Logic: `git://` code paths produce duplicate errors. Step 4 (line 102) correctly flags the git path, but step 5 here still calls `validateLocalPath`. Since `IsRemoteURI` doesn't recognize `git://` or `git+` schemes, `os.Stat("git://...")` fails and adds a second, misleading "local path does not exist" error.

Fix: either extend `IsRemoteURI` in yaml_parser.go to cover `git://` and `git+` prefixes, or guard this call to skip when a git-path error was already recorded.

cli/azd/extensions/azure.ai.customtraining/internal/utils/job_validator.go
```diff
@@ -0,0 +1,272 @@
...
+
+// validateSingleBracePlaceholders flags when the command uses {inputs.xxx} or {outputs.xxx}
+// instead of the correct ${{inputs.xxx}} syntax. This is an error because the backend
+// will not resolve single-brace placeholders.
+func validateSingleBracePlaceholders(result *ValidationResult, command string) {
+	for _, match := range singleBraceRegex.FindAllStringSubmatchIndex(command, -1) {
+		start := match[0]
+		// Skip if this is already part of a ${{...}} (preceded by "${")
+		if start >= 2 && command[start-2:start] == "${" {
+			continue
```
jongio, line 234: **[MEDIUM]** Logic: The bare-`$` guard here creates a false negative. For `${inputs.model}` (wrong syntax - should be `${{inputs.model}}`), the regex matches `{inputs.model}` at position N, then `command[N-1:N] == "$"` causes the skip. But `${inputs.model}` isn't valid either.

The first guard (line 221, checking for `"${"` two chars back) already handles valid `${{...}}` patterns correctly. This second single-char `$` guard is redundant and causes `${inputs.xxx}` typos to go undetected.

Fix: remove lines 224-226 entirely.

cli/azd/extensions/azure.ai.customtraining/internal/utils/job_validator.go
```diff
@@ -0,0 +1,272 @@
...
+		return
+	}
+
+	// Resolve relative paths against the YAML file directory
+	resolved := path
+	if !filepath.IsAbs(path) {
+		resolved = filepath.Join(yamlDir, path)
+	}
+
+	if _, err := os.Stat(resolved); os.IsNotExist(err) {
```
jongio, line 135: **[MEDIUM]** Error Handling: Only `os.IsNotExist` is checked. If `os.Stat` fails with a permission error or other I/O error, the function silently returns success - the user won't know the path couldn't be verified.

Fix: add an `else if err != nil` branch to report a warning like "could not verify path exists: <error>".

cli/azd/extensions/azure.ai.customtraining/internal/cmd/job_validate.go
```diff
@@ -0,0 +1,76 @@
...
+			}
+
+			fmt.Printf("\n✓ Validation passed with warnings.\n")
+			return nil
+		},
+	}
+
+	cmd.Flags().StringVar(&filePath, "file", "", "Path to YAML job definition file (required)")
+
+	return cmd
```
jongio, line 78: **[MEDIUM]** Consistency: The sibling `job submit` command uses `StringVarP(&filePath, "file", "f", ...)` with a `-f` shorthand. This command uses `StringVar` without it. Users who learn `-f` from `submit` will expect it here too.

Fix: change to `cmd.Flags().StringVarP(&filePath, "file", "f", "", "Path to YAML job definition file (required)")`.

jongio (changes requested):
## PR Review - Azure/azure-dev#7407

**Add validate command for custom training** by @saanikaguptamicrosoft

### Summary

What: Adds an offline `job validate` command that checks a YAML job definition for required fields, invalid git paths, local path existence, placeholder correctness, and empty input/output definitions - collecting all findings before reporting instead of failing on the first error.

Why: Lets users catch YAML mistakes locally before submitting to the backend, saving a round-trip.

Assessment: The validation logic is solid and well-structured, but the command can't actually run offline due to the parent command's `PersistentPreRunE` hook. This should be addressed before merge. A few edge cases in the validator produce confusing output (duplicate errors on git paths, silent pass on `${inputs.xxx}` typos).

### Findings

| Category | High | Medium | Low |
|----------|------|--------|-----|
| Design | 1 | 0 | 0 |
| Logic | 0 | 2 | 0 |
| Error Handling | 0 | 1 | 0 |
| Consistency | 0 | 1 | 1 |
| Tests | 0 | 0 | 1 |
| **Total** | **1** | **4** | **2** |

### Key Findings

1. **[HIGH] Design**: Parent `job` command's `PersistentPreRunE` requires Azure env setup, preventing offline use
2. **[MEDIUM] Logic**: `git://` code paths produce duplicate errors (git-not-supported + local-path-not-found)
3. **[MEDIUM] Logic**: `${inputs.xxx}` typos silently bypass single-brace detection
4. **[MEDIUM] Error Handling**: `os.Stat` permission/IO errors silently ignored in path checks
5. **[MEDIUM] Consistency**: Missing `-f` shorthand for `--file` flag (inconsistent with `job submit`)
6. **[LOW] Consistency**: Missing `Long` description on cobra command (all sibling commands have one)
7. **[LOW] Tests**: No test for the `git://` double-error scenario (assert only 1 finding, not 2)

### Test Coverage Estimate

- Well covered: required fields, git path detection, local path checks, placeholder mapping, single-brace detection, empty definitions, multiline commands
- Missing: command-layer test (cobra handler), `${inputs.xxx}` edge case, `git://` double-error assertion, `os.Stat` permission errors, empty/zero-byte YAML input

### What's Done Well

- Collects all findings instead of failing on the first error - much better UX than the existing `ValidateJobDefinition`
- Good separation between the command handler and validation logic
- Optional `[...]` block handling for input placeholders is a nice touch
- Test file has clear YAML comments showing what each test case maps to

---
*5 inline comments below.*

cli/azd/extensions/azure.ai.customtraining/internal/cmd/job.go
saanikaguptamicrosoft: Good catch, I've updated accordingly.

cli/azd/extensions/azure.ai.customtraining/internal/utils/job_validator.go
saanikaguptamicrosoft: Good catch, addressed.

cli/azd/extensions/azure.ai.customtraining/internal/cmd/job_validate.go
saanikaguptamicrosoft: Updated.

cli/azd/extensions/azure.ai.customtraining/internal/utils/job_validator.go
saanikaguptamicrosoft: Makes sense, resolved.

cli/azd/extensions/azure.ai.customtraining/internal/utils/job_validator_test.go
```diff
@@ -0,0 +1,314 @@
...
+
+// Tests single-brace {inputs.xxx} is flagged as error (backend won't resolve it).
+func TestValidate_SingleBracePlaceholders(t *testing.T) {
+	// YAML (incorrect):
+	//   command: python train.py --data {inputs.training_data} --out {outputs.model}
+	// Should be: ${{inputs.training_data}} and ${{outputs.model}}
+	job := validJob()
+	job.Command = "python train.py --data {inputs.training_data} --out {outputs.model}"
+	result := ValidateJobOffline(job, ".")
+	if f := findFindingByMessage(result, "single-brace '{inputs.training_data}'"); f == nil || f.Severity != SeverityError {
```
jongio: **[HIGH]** These test assertions don't match the actual error message.

`validateSingleBracePlaceholders` produces `"Incorrect placeholder format - use '${{inputs.training_data}}' instead"` but these 3 assertions (lines 223, 226, 236) search for `"single-brace '...'"`. They'll fail because that substring isn't in the message. Line 249's negative check also passes vacuously.

Fix: match on the actual message, e.g. search for `"Incorrect placeholder format"` or `"${{inputs.training_data}}"` instead.

jongio (commented):
Previous 5 findings addressed. New issues found:

- job_validator_test.go:223 - single-brace test assertions search for `"single-brace '...'"` but the actual message is `"Incorrect placeholder format..."`. Three assertions will fail (223, 226, 236)
- job_validator.go:201 - duplicate findings when the same placeholder key appears twice in command
- job_validator.go:106 - fmt.Sprintf with no format verbs (staticcheck SA1006)

cli/azd/extensions/azure.ai.customtraining/internal/utils/job_validator.go
```diff
@@ -0,0 +1,284 @@
...
+}
+
+// validatePlaceholders checks that ${{inputs.xxx}} references in command exist in job.Inputs
+// and ${{outputs.xxx}} references exist in job.Outputs.
+// References inside [...] optional blocks are skipped for inputs.
+func validatePlaceholders(result *ValidationResult, job *JobDefinition, optionalInputs map[string]bool) {
+	command := job.Command
+
+	// Find all ${{inputs.xxx}} and ${{outputs.xxx}} references
+	for _, match := range placeholderRegex.FindAllStringSubmatch(command, -1) {
```
jongio, line 193: **[MEDIUM]** `validatePlaceholders` (and `validateInputOutputDefinitions` at L255) iterates ALL regex matches. If the same key appears twice in the command, you get duplicate identical findings. Track seen keys with a map to dedup.

cli/azd/extensions/azure.ai.customtraining/internal/utils/job_validator.go
```diff
@@ -0,0 +1,284 @@
...
+	}
+
+	// 4. code must not be a git path
+	if job.Code != "" {
+		lower := strings.ToLower(job.Code)
+		if strings.HasPrefix(lower, "git://") || strings.HasPrefix(lower, "git+") {
+			result.Findings = append(result.Findings, ValidationFinding{
+				Field:    "code",
+				Severity: SeverityError,
+				Message:  fmt.Sprintf("git paths are not supported"),
```
jongio: **[LOW]** `fmt.Sprintf` with no format verbs - just use a plain string literal.

cli/azd/extensions/azure.ai.customtraining/internal/utils/job_validator.go
```diff
@@ -0,0 +1,298 @@
...
+	}
+
+	// 3. compute field is required
+	if job.Compute == "" {
+		result.Findings = append(result.Findings, ValidationFinding{
+			Field:    "compute",
+			Severity: SeverityError,
+			Message:  "required field is missing",
+		})
+	}
```
achauhan-scc, line 89: we can add required tag in job definition and loop over it, we should not be doing field by field

cli/azd/extensions/azure.ai.customtraining/internal/utils/job_validator.go
```diff
@@ -0,0 +1,298 @@
...
+	// 7. Warn on single-brace {inputs.xxx} or {outputs.xxx} usage in command
+	if job.Command != "" {
+		validateSingleBracePlaceholders(result, job.Command)
+	}
+
+	// 8. Inputs/outputs with nil/empty definitions referenced in command
+	if job.Command != "" {
+		validateInputOutputDefinitions(result, job, optionalInputs)
+	}
+
```
achauhan-scc, line 118: Can we all three to above else of line 73

achauhan-scc (changes requested):

cli/azd/extensions/azure.ai.customtraining/internal/utils/job_validator.go
```diff
@@ -0,0 +1,279 @@
...
+	// Matches ${{inputs.key}} — used to extract input keys from optional blocks.
+	inputPlaceholderRegex = regexp.MustCompile(`\$\{\{inputs\.(\w[\w.-]*)}}`)
+
+	// Matches single-brace {inputs.key} or {outputs.key} that are NOT preceded by $ or another {.
+	// Uses a negative lookbehind approximation: we check matches and filter in code.
+	singleBraceRegex = regexp.MustCompile(`\{(inputs|outputs)\.(\w[\w.-]*)}}?`)
+)
+
+// optionalInputKeys returns the set of input keys that appear inside [...] optional blocks.
+func optionalInputKeys(command string) map[string]bool {
```
jongio, line 168: `optionalInputKeys` marks keys as optional globally if they appear in ANY `[...]` block. If the same key is used both required (outside brackets) and optional (inside brackets), the required usage is silently skipped.

Example: `--data ${{inputs.data}} [--extra ${{inputs.data}}]` - `data` gets marked optional, so a missing `inputs.data` definition won't be flagged for the required `--data` usage.

Fix: only treat a key as optional if it doesn't also appear outside `[...]`. One way: after building this map, scan the command outside brackets for the same keys and remove them.

jongio (commented):
All 8 findings from my previous reviews are addressed - nice work. A few new items from this pass, mostly edge cases in the placeholder validation. Nothing blocking.

cli/azd/extensions/azure.ai.customtraining/internal/utils/job_validator.go
```diff
@@ -0,0 +1,279 @@
...
+				})
+			}
+		}
+	}
+}
+
+// validateSingleBracePlaceholders flags when the command uses {inputs.xxx} or {outputs.xxx}
+// instead of the correct ${{inputs.xxx}} syntax. This is an error because the backend
+// will not resolve single-brace placeholders.
+func validateSingleBracePlaceholders(result *ValidationResult, command string) {
```
jongio, line 228: Missing dedup. `validatePlaceholders` and `validateInputOutputDefinitions` both got `seen` maps (fixing my earlier feedback), but this function doesn't have one. A command like `{inputs.data} ... {inputs.data}` produces duplicate identical errors.

Fix: add a `seen` map keyed on `kind + "." + key` matching the pattern in the other two functions.

cli/azd/extensions/azure.ai.customtraining/internal/utils/job_validator_test.go
```diff
@@ -0,0 +1,314 @@
...
+
+	// YAML (correct):
+	//   command: python train.py --data ${{inputs.training_data}}
+	// Correct ${{...}} should NOT trigger single-brace error
+	job = validJob()
+	job.Command = "python train.py --data ${{inputs.training_data}}"
+	job.Inputs = map[string]InputDefinition{
+		"training_data": {Type: "uri_folder", Path: "azureml://datastores/blob/data"},
+	}
+	result = ValidateJobOffline(job, ".")
```
jongio, line 258: Missing regression test for `${inputs.xxx}` (dollar + single brace). The code catches this correctly now, but the original version had a bug here and there's no test to prevent regression.

Suggested addition:```go
job = validJob()
job.Command = "python train.py --data ${inputs.data}"
result = ValidateJobOffline(job, ".")
if f := findFindingByMessage(result, "${{inputs.data}}"); f == nil {
    t.Error("expected error for dollar-single-brace ${inputs.xxx}")
}```

cli/azd/extensions/azure.ai.customtraining/internal/utils/job_validator.go
```diff
@@ -0,0 +1,279 @@
...
+func ValidateJobOffline(job *JobDefinition, yamlDir string) *ValidationResult {
+	result := &ValidationResult{}
+
+	// 1–3. Check required fields via struct tags
+	v := reflect.ValueOf(*job)
+	t := v.Type()
+	for i := 0; i < t.NumField(); i++ {
+		field := t.Field(i)
+		if field.Tag.Get("validate") == "required" {
+			if v.Field(i).String() == "" {
```
jongio, line 81: Nit: `v.Field(i).String()` returns `<T Value>` (not empty string) for non-string types, so the required check silently passes. All tagged fields are strings today, but a comment like `// Note: only validates string fields` would prevent someone from tagging a non-string field later.

cli/azd/extensions/azure.ai.customtraining/internal/cmd/job_validate.go
```diff
@@ -0,0 +1,78 @@
...
+	"github.com/fatih/color"
+	"github.com/spf13/cobra"
+	"gopkg.in/yaml.v3"
+)
+
+func newJobValidateCommand() *cobra.Command {
+	var filePath string
+
+	cmd := &cobra.Command{
+		Use:   "validate",
```
jongio, line 22: Nit: sibling commands like `submit` have a `Long` description. Missing here.

cli/azd/extensions/azure.ai.customtraining/internal/utils/job_validator.go
```diff
@@ -0,0 +1,297 @@
...
+// instead of the correct ${{inputs.xxx}} syntax. This is an error because the backend
+// will not resolve single-brace placeholders.
+func validateSingleBracePlaceholders(result *ValidationResult, command string) {
+	seen := make(map[string]bool)
+	for _, match := range singleBraceRegex.FindAllStringSubmatchIndex(command, -1) {
+		start := match[0]
+		// Skip if this is already part of a ${{...}} (preceded by "${")
+		if start >= 2 && command[start-2:start] == "${" {
+			continue
+		}
```
jongio, line 235: Non-blocking: `${{inputs.data}` (missing one closing brace) bypasses both checks - `placeholderRegex` needs `}}`, and `singleBraceRegex` sees the `${` prefix and skips. Worth adding a malformed-placeholder check in a follow-up.

jongio (approved):
All 8+ findings from my previous rounds are addressed - nice work.

One non-blocking edge case for a follow-up: malformed `${{inputs.data}` (correct opening but only one closing brace) slips through both regex checks silently. `placeholderRegex` needs `}}` so it doesn't match, and `singleBraceRegex` matches but the `${` guard skips it. The backend would catch this, but it'd be good to detect offline too.

achauhan-scc (approved):

PR #7410, opened 2026-03-31
Fix init with project-id when agent.yaml doesn't contain a model resource

### LINKED ISSUE #7409: project-id flag is ignored when agent manifest doesn't contain a model resource
When running azd ai agent init -p <project-resource-id> -m <manifest-url>, the --project-id flag is ignored if the agent manifest does not contain model resources (no resources: section with kind: model). Instead
  of using the provided Foundry project, the CLI prompts for a location and proceeds to create new resources.

  Manifests with model resources correctly honor -p and use the existing project.

  Repro
```
   # Manifest WITHOUT model resources — ignores -p, prompts for location
   azd ai agent init \
     -p /subscriptions/<SUB>/resourceGroups/<RG>/providers/Microsoft.CognitiveServices/accounts/<ACCOUNT>/projects/<PROJECT> \
     -m https://github.com/microsoft/hosted-agents-vnext-private-preview/blob/antrikshjain/invoke-response-samples/samples/python/agentserver-responses/echo-agent-streaming/agent.yaml
```
```   
   # Manifest WITH model resources — honors -p, uses existing project
   azd ai agent init \
     -p /subscriptions/<SUB>/resourceGroups/<RG>/providers/Microsoft.CognitiveServices/accounts/<ACCOUNT>/projects/<PROJECT> \
     -m https://github.com/microsoft-foundry/foundry-samples/blob/main/samples/python/hosted-agents/agent-framework/agent-with-local-tools/agent.yaml
```

  **Expected behavior**

  Both commands should use the provided Foundry project (skip location prompt, configure ACR/AppInsights from the project).

  **Actual behavior**

  The no-model-resources manifest prompts "Select an Azure location" and enters the create-new-resources flow, ignoring -p entirely.

  **Root cause**

  In configureModelChoice() (internal/cmd/init.go), the !manifestHasModelResources() branch unconditionally calls ensureSubscriptionAndLocation() and returns early — it never checks a.flags.projectResourceId and
  never calls selectFoundryProject().

  The "has model resources" branch correctly checks projectResourceId and auto-selects the existing-project path via selectFoundryProject(), which derives location from the project and configures Foundry env vars
  (ACR, AppInsights, etc.).

   // Current code — no-model-resources path ignores -p
   if !manifestHasModelResources(agentManifest) {
       newCred, err := ensureSubscriptionAndLocation(...)  // always prompts for location
       a.credential = newCred
       return agentManifest, nil  // never reaches selectFoundryProject()
   }

  Fix

  When -p is provided and there are no model resources, follow the same pattern as the existing-project path: ensureSubscription() → selectFoundryProject() (sets location, ACR, AppInsights from the project) → fall
  back to ensureLocation() only if the project isn't found.

therealjohn:
Copilot created
Fixes #7409

### DIFF
--- a/cli/azd/extensions/azure.ai.agents/internal/cmd/init.go
+++ b/cli/azd/extensions/azure.ai.agents/internal/cmd/init.go
@@ -568,9 +568,7 @@ func (a *InitAction) configureModelChoice(
 			}
 
 			if selectedProject == nil {
-				if err := ensureLocation(ctx, a.azdClient, a.azureContext, a.environment.Name); err != nil {
-					return nil, err
-				}
+				return nil, fmt.Errorf("foundry project not found: %s", a.flags.projectResourceId)
 			}
 		} else {
 			newCred, err := ensureSubscriptionAndLocation(
--- a/cli/azd/extensions/azure.ai.agents/internal/cmd/init_from_templates_helpers.go
+++ b/cli/azd/extensions/azure.ai.agents/internal/cmd/init_from_templates_helpers.go
@@ -21,7 +21,7 @@ import (
 	"github.com/fatih/color"
 )
 
-const agentTemplatesURL = "https://raw.githubusercontent.com/therealjohn/awesome-azd/refs/heads/agent-tooling-templates/website/static/agent-templates.json"
+const agentTemplatesURL = "https://aka.ms/foundry-agents"
 
 // Template type constants
 const (


cli/azd/extensions/azure.ai.agents/internal/cmd/init.go
```diff
@@ -545,16 +545,44 @@ func (a *InitAction) configureModelChoice(
...
+			)
+			if err != nil {
+				return nil, err
+			}
+
+			if selectedProject == nil {
+				if err := ensureLocation(ctx, a.azdClient, a.azureContext, a.environment.Name); err != nil {
+					return nil, err
+				}
+			}
```
jongio, line 572: **[MEDIUM]** Logic: Unreachable nil-project guard

When `projectResourceId` is explicitly provided, `selectFoundryProject` calls `getFoundryProject` which either returns a valid project (success) or returns an error (failure). It can't return `(nil, nil)` - doing so would panic at line 926 of `init_foundry_resources_helpers.go` (`*project` dereference).

This `selectedProject == nil` block is dead code in the `projectResourceId != ""` branch. It's cargo-culted from the "existing" model config case (line 610) where it handles the interactive "Create new" choice - which doesn't apply when a specific project ID is passed.

Not harmful, but misleading to a reader who might think nil is a valid outcome here. Consider replacing with a defensive error:

```suggestion
		if selectedProject == nil {
			return nil, fmt.Errorf("foundry project not found: %s", a.flags.projectResourceId)
		}
```

cli/azd/extensions/azure.ai.agents/internal/cmd/init_from_templates_helpers.go
```diff
@@ -21,7 +21,7 @@ import (
 	"github.com/fatih/color"
 )
 
-const agentTemplatesURL = "https://aka.ms/foundry-agents"
+const agentTemplatesURL = "https://raw.githubusercontent.com/therealjohn/awesome-azd/refs/heads/agent-tooling-templates/website/static/agent-templates.json"
```
wbreza: **[HIGH]** Template URL changed to personal repo branch — unrelated to bug fix

This changed from `https://aka.ms/foundry-agents` (stable Microsoft shortlink) to a raw GitHub URL on a personal repo's feature branch. Risks:

1. Personal repos can be deleted/renamed
2. Feature branch `agent-tooling-templates` may be deleted after merge
3. The `aka.ms` link was a stable, owned endpoint

Consider reverting this change and either updating the `aka.ms` redirect target, or moving this URL change to a separate PR once the content is published to an official endpoint.

wbreza (commented):
## PR Review — Azure/azure-dev#7410

**Fix init with project-id when agent.yaml doesn't contain a model resource** by @therealjohn

### Summary

What: When `--project-id` is provided but the agent manifest has no model resources, `configureModelChoice` now calls `ensureSubscription` + `selectFoundryProject` (mirroring the existing-project pattern) to derive location and Foundry env vars from the specified project instead of prompting.

Why: Fixes #7409 — the no-model-resources path was ignoring `--project-id` and always prompting for location.

Assessment: The bug fix is correct and mirrors the established pattern well. One concern: the template URL constant change is unrelated to the fix and points to a personal repo branch.

### Findings

| Priority | Count |
|----------|-------|
| High | 1 |
| Low | 2 |
| **Total** | **3** |

### What's Done Well

- Fix correctly mirrors the existing-project pattern from the "existing" model config case
- Updated comment clearly explains the `--project-id` behavior
- All error paths properly propagated

Issue #7245, opened 2026-03-23
Update agent calls to new endpoint


therealjohn:
vnext agents need to use `POST {{agentsV2ProjectEndpoint}}/agents/{{agentVersion.name}}/endpoint/protocols/openai/v1/responses`

PR #7422, opened 2026-04-01
[Agents extension] Adding /invocations support

trangevi:
Adding support for the new /invocations API. This uses the new agent endpoint route, and updates the /responses handling to do the same.

/invocations calls can accept any input, based on the agent, so we do no validation on it and instead just pass it along to the agent, relying on the agent to handle it.


therealjohn:
Fixes https://github.com/Azure/azure-dev/issues/7431
Fixes https://github.com/Azure/azure-dev/issues/7245

cli/azd/extensions/azure.ai.agents/internal/cmd/invoke.go
```diff
@@ -265,15 +396,451 @@ func (a *InvokeAction) invokeRemote(ctx context.Context) error {
...
+				"timed out waiting for invocation %s to complete (timeout: %s)",
+				invocationID, timeout,
+			)
+		}
+
+		time.Sleep(pollInterval)
+
+		req, err := http.NewRequestWithContext(
+			resp.Request.Context(), http.MethodGet, pollURL, nil,
+		)
```
JeffreyCA, line 822: Copilot caught a potential issue here:

> `handleInvocationLRO` uses `resp.Request.Context()` to create polling requests. After the initial 202 response, this context may already be cancelled/timed out, causing all polling requests to fail immediately - making long-running operations unusable.

> Suggested fix: Add `ctx context.Context` parameter to `handleInvocationLRO` and pass it through from the caller instead of extracting it from the spent request.

cli/azd/extensions/azure.ai.agents/internal/cmd/invoke.go
```diff
@@ -265,15 +396,451 @@ func (a *InvokeAction) invokeRemote(ctx context.Context) error {
...
+	localBaseURL := fmt.Sprintf("http://localhost:%d", port)
+
+	// Fetch and cache the agent's OpenAPI spec (always refresh for local).
+	if azdClient != nil {
+		fetchOpenAPISpec(ctx, azdClient, localBaseURL, "local", "local", "", true)
+	}
+
+	url := localBaseURL + "/invocations"
+	if sid != "" {
+		url += "?agent_session_id=" + sid
```
jongio: **[MEDIUM]** Session ID goes into the URL without encoding. A `--session` value containing `&` or `=` corrupts the query string.

Since the local variable `url` shadows `net/url` here, consider importing as `neturl "net/url"`` and using `neturl.QueryEscape(sid)`, or building the URL with `url.Values` before assigning the local variable. Same issue on line 535 for the remote path.

cli/azd/extensions/azure.ai.agents/internal/cmd/invoke.go
```diff
@@ -265,15 +396,451 @@ func (a *InvokeAction) invokeRemote(ctx context.Context) error {
...
+
+const (
+	defaultLROPollInterval = 2 * time.Second
+	maxLROPollInterval     = 30 * time.Second
+)
+
+// handleInvocationLRO handles a long-running operation (202 Accepted) invocations response
+// by polling GET on the invocation's status URL (derived from the original request URL)
+// until a terminal state is reached.
+func handleInvocationLRO(
```
jongio, line 745: **[MEDIUM]** This is the most complex new code in the PR - polling loop, URL derivation via string manipulation, state machine, timeout handling, retry-after - and it has no test coverage.

The routing test only checks the `no invocation ID` error. An `httptest.Server` returning 202-then-completed would validate the happy path, and variants for timeout/failure/cancelled states would cover the branches. The table-driven pattern already in this file works well for this.

rajeshkamal5050 (approved):

PR #7450, opened 2026-04-02
feat: add azd tool command group for unified Azure tooling management

### LINKED ISSUE #7845: [EPIC] A: Core Command Group & Installation Engine
## Problem / User Value

Today, developers must discover, install, and upgrade Azure development tools (az, gh, bicep, func, docker, terraform, etc.) through multiple package managers and documentation pages. There is no unified CLI surface for managing the Azure toolchain. `azd tool` provides a single entry point for toolchain discovery, installation, and lifecycle management across all three platforms (Windows, macOS, Linux).

## Definition of Done

`azd tool list`, `show`, `check`, `install`, and `upgrade` all function with table and JSON output. At least 10 tools and 3 VS Code extensions in the embedded registry. All 3 platforms (Windows/macOS/Linux) supported with native package manager strategies (winget/brew/apt) plus direct download fallback with checksum verification.

## Tasks

- [ ] **A-1: Add ToolDescriptor data model and shared types** (M) — Create `pkg/tool/types.go` with `ToolCategory` enum (cli, extension, server, library, ai-evaluation, agent-skills per S3), `InstallerType` enum (winget, brew, apt, direct, vscode), `ToolStatus` enum. Create `pkg/tool/descriptor.go` with `ToolDescriptor`, `InstallMethod`, `VersionCommand`, `ToolLinks`, `DetectionResult`, `InstallResult`, `UpgradeResult` structs per architecture doc §3.1. Create `pkg/tool/errors.go` with sentinel errors. All types need JSON struct tags, Go doc comments, and unit tests for YAML/JSON round-trip serialization.
  - AC: `pkg/tool/types.go` defines all enums with doc comments; `descriptor.go` defines structs with JSON tags; `errors.go` defines sentinel errors; categories include `ai-evaluation` and `agent-skills`; unit tests validate YAML/JSON round-trip serialization
- [ ] **A-2: Implement embedded YAML registry with schema versioning** (M) — Create `pkg/tool/registry.go` with `ToolRegistry` interface (All, Get, Recommended, Search methods per §3.5). Implement `registry_embedded.go` with `go:embed registry.yaml` and YAML parsing. Schema versioning: top-level `schemaVersion: 1`, parser checks version, ignores unknown fields (forward-compatible). Register as singleton in `cmd/container.go`.
  - AC: `ToolRegistry` interface with All/Get/Recommended/Search methods; embedded via `go:embed`; schema version validated; Search supports free-text on name/description/tags; unit tests cover valid parse, version mismatch, missing fields, search filtering
- [ ] **A-3: Implement ToolDetector with PATH-based detection** (M) — Create `pkg/tool/detector.go` with `ToolDetector` interface and `PathDetector` implementation per §3.2. Uses `azdext.LookupTool` for PATH search, runs version command via `pkg/exec.CommandRunner` with 5-second timeout, parses version via `tools.ExtractVersion` with configurable regex. Returns `DetectionResult` with Found, Path, Version, MeetsMinimum fields.
  - AC: `Detect(ctx, desc)` signature; PATH detection via `azdext.LookupTool`; version extraction with configurable regex; 5-second timeout; `MeetsMinimum` semver comparison; unit tests for found/not-found/below-minimum/timeout/regex-mismatch
- [ ] **A-4: Implement ToolInstaller strategies (winget, brew, apt, direct, vscode) with checksum verification** (L) — Create `pkg/tool/installer.go` with `ToolInstaller` interface per §3.3. Implement five installer strategies: WingetInstaller, BrewInstaller (supports cask+tap), AptInstaller (supports PPA), DirectDownloadInstaller (with SHA-256 checksum verification reusing `validateChecksum` from `pkg/extensions/manager.go`), VSCodeExtensionInstaller. All use `exec.CommandRunner`, all are idempotent.
  - AC: Interface with Type/Available/Install methods; each strategy handles its platform; direct downloads verified via SHA-256 checksum (static from registry or fetched from ChecksumURL); checksum mismatch deletes artifact + returns `ErrorWithSuggestion`; all idempotent; unit tests with mocked CommandRunner per installer
- [ ] **A-5: Implement ToolManager orchestrator** (M) — Create `pkg/tool/manager.go` with `ToolManager` interface per §3.4 (List, Show, Check, Install, Upgrade, CheckForUpdates). Implements parallel detection via `errgroup.Group` with `runtime.NumCPU()` concurrency limit. Installer fallback chain: iterates InstallMethods filtered by OS/arch, picks first available, falls back on error. Detection cache: in-memory + disk at `~/.azd/tool-version-cache.json` with 24h TTL. Register as singleton in `cmd/container.go`.
  - AC: All 6 ToolManager methods; parallel detection with configurable concurrency; installer selection with OS/arch filtering and fallback chain; detection cache with 24h TTL; registered as IoC singleton; unit tests for parallel detection, fallback chain, cache hit/miss
- [ ] **A-6: Add `azd tool list` and `azd tool show` commands** (M) — Create `cmd/tool.go` following `cmd/env.go` pattern. Implement `listAction` with `--all`, `--category` flags and table/JSON output. Implement `showAction` with `<tool-id>` positional arg. Wire into `cmd/root.go`. Register bare `azd tool` as alias for `azd tool list`. Update command snapshots.
  - AC: Table output with Name/Installed/Available/Status columns; `--output json` per UX design §Flow 2; `--category` filter; `--all` includes non-recommended; `azd tool show <id>` with full detail; bare `azd tool` = `azd tool list`; summary line with counts; snapshot tests pass
- [ ] **A-7: Add `azd tool check` command** (S) — Add `checkAction` to `cmd/tool.go` with optional `[tool...]` positional args. Shows table of tools with updates available. Exit code 0 when all up to date, non-zero when updates available (CI scripting).
  - AC: Default shows tools needing updates; `azd tool check gh bicep` checks specific tools; `--output json` per §Flow 5; "All tools are up to date" message; appropriate exit codes
- [ ] **A-8: Add `azd tool install` command (interactive + non-interactive + --dry-run)** (L) — Add `installAction` with variadic `[tool...]` args, `--all`, `--include-optional`, `--dry-run`, `--no-prompt` flags. No args + interactive → MultiSelect picker (excluding installed). No args + `--no-prompt` → error. `--all` → recommended; `--all --include-optional` → everything. `--dry-run` → plan without execution. Per-tool progress with bar-fill spinner and `(✓)/(x)/(-)` states. Partial failure continues remaining installs.
  - AC: Interactive MultiSelect picker; named tool install; `--all` and `--all --include-optional`; `--dry-run` shows plan; `--no-prompt` (no args) errors; skipped/failed/success states per azd style guide; partial failure continues; JSON output matches §Flow 3; snapshot tests; never install without explicit consent (NG4)
- [ ] **A-9: Add `azd tool upgrade` command** (M) — Add `upgradeAction` with variadic `[tool...]` args and `--dry-run` flag. No args → upgrade all outdated. Not-installed → error with `azd tool install` hint. Progress shows old → new version. Uses `ToolManager.Upgrade()`.
  - AC: Upgrades all outdated or specified tools; not-installed error with hint; `--dry-run` shows version transitions; "All up to date" message; JSON output per §Flow 4 with `previousVersion`; snapshot tests
- [ ] **A-10: Author initial tool registry entries** (M) — Author entries in `pkg/tool/registry.yaml` for: az, gh, docker, bicep, func, terraform, node, python, dotnet, kubectl. Include VS Code extensions: Azure Tools, Bicep, GitHub Copilot. Include servers: Azure MCP Server. Each entry needs detection config, install methods per platform, links, tags. Mark core tools as `recommended: true`.
  - AC: ≥10 tool entries + 3 VS Code extensions; working detection config per entry; ≥1 install method per platform per entry; recommended flags set; categories assigned per S3; YAML validation tests pass; cspell passes

## Sequencing

```
A-1 ──┬──> A-2 ──┐
      ├──> A-3 ──┤
      └──> A-4 ──┘──> A-5 ──> A-6 ──┬──> A-7
                                     └──> A-8 ──> A-9
A-2 ──────────────────────> A-10 (can run any time after A-2)
```

- **Parallel after A-1:** A-2, A-3, A-4 are independent and can be developed simultaneously
- **Parallel after A-5:** A-10 can run alongside A-6+
- **Sequential:** A-6 → A-8 → A-9 (each builds on the previous command pattern)
- **A-7** can run in parallel with A-8 after A-6

## Dependencies

- **Epic F (Prior-Art Research):** ✅ Complete — UX patterns already decided
- No other epic dependencies — this is the foundation

## Discussion Traceability

| Discussion Item | Task |
|---|---|
| F4 (two modes: interactive + scripted) | A-5, A-8 |
| F6 (trust/signing) | A-4 (checksum sub-requirement for direct downloads) |
| F12 (distribution channels) | A-3 (detection across channels) |
| S2 (`install --all` behavior) | A-5, A-8 |
| S3 (toolkit categories) | A-1 (category enum) |

## Design References

- `docs/design/azd-tool/plan.md` — Phase 1a
- `docs/design/azd-tool-architecture.md` — §3.1–3.5 (data model, registry, detector, installer, manager)
- `docs/design/azd-tool/todos.md` — detailed acceptance criteria per task

### LINKED ISSUE #7846: [EPIC] B: First-Run Experience
## Problem / User Value

New azd users face a "blank slate" problem — they don't know which Azure tools they need, and the documentation spreads tool installation across dozens of pages. A guided first-run experience minimizes decisions for newcomers (one confirm installs everything recommended) while providing an escape hatch for advanced users who want to customize.

## Definition of Done

First-run triggers on first `azd init`/`up`/`provision`/`deploy`. Single-confirm path installs all recommended tools. Customize path provides MultiSelect picker. Skip mechanisms work (`AZD_SKIP_FIRST_RUN`, `--no-prompt`, config). Post-install summary with next-step hints displayed.

## Tasks

- [ ] **B-1: Implement first-run trigger middleware** (M) — Create `cmd/middleware/tool_firstrun.go` per §7.1. Triggers when `~/.azd/tool-check-timestamp` is absent, terminal is interactive, and command is init/up/provision/deploy (not tool/*, version, config). Register after telemetry middleware, before hooks. Detects missing recommended tools via `ToolManager.Check()`. Writes timestamp on completion.
  - AC: Triggers on first init/up/provision/deploy; does NOT trigger on tool/*/version/config commands; does NOT trigger with `--no-prompt` or non-interactive terminal; timestamp prevents re-triggering; registered after telemetry, before hooks
- [ ] **B-2: Build streamlined first-run UX (single-confirm + customize paths)** (L) — Default path: display recommended tools list, then "Install recommended tools? [Y/n]" single confirm → install all. `n` → MultiSelect picker with all tools (recommended pre-checked). Per-tool progress with bar-fill spinner and `(✓)/(x)/(-)` states. Colors per UX design: bold title, highlighted names, gray versions.
  - AC: Single Y/n prompt installs all recommended; `n` opens MultiSelect with recommended pre-checked; per-tool progress display; partial failure continues + WARNING summary; all-already-installed shows appropriate message; colors match UX spec
- [ ] **B-3: Add skip mechanisms (env var, --no-prompt, config)** (S) — `AZD_SKIP_FIRST_RUN=true` env var, `--no-prompt` flag, and `azd config set tool.firstRunCompleted true` all suppress the first-run middleware. Update `docs/environment-variables.md`.
  - AC: All three skip mechanisms work; `docs/environment-variables.md` updated; unit tests for each skip path
- [ ] **B-4: Add post-install summary with next-step hints** (S) — After install completes, show installed/skipped/failed counts. Next-step hint: `azd tool doctor` to verify. Failed tools get individual retry hints. SUCCESS/WARNING formatting per azd style guide.
  - AC: Summary with counts; next-step hint includes `azd tool doctor`; failed tools get retry hints; SUCCESS format on full success, WARNING on partial failure

## Sequencing

```
B-1 ──┬──> B-2 ──> B-4
      └──> B-3
```

- **B-2 and B-3** can run in parallel after B-1
- **B-4** depends on B-2

## Dependencies

- **Epic A (Core Engine):** A-5 (ToolManager) must be complete — first-run middleware depends on it
- **Epic F (Research):** ✅ Complete — F-2's UX recommendations inform B-2's design

## Discussion Traceability

| Discussion Item | Task |
|---|---|
| F5 (first-run simplify: "one confirm, not N checkboxes") | B-2 |

## Design References

- `docs/design/azd-tool/plan.md` — Phase 2a
- `docs/design/azd-tool-architecture.md` — §7.1 (middleware), UX design §Flow 1
- `docs/design/azd-tool/todos.md` — detailed acceptance criteria

wbreza:
## Summary

Adds `azd tool` — a unified command group for discovering, installing, upgrading, and managing Azure development tools directly from the CLI. No more hunting for install docs or remembering package manager commands across platforms.

**Parent epic:** #7676

---

### Core Command Group (Closes #7845)

- **`azd tool list`** — shows all registered tools with install status, version, and category
- **`azd tool show <id>`** — displays detailed info for a specific tool (install strategies per-platform, website, etc.)
- **`azd tool check`** — checks installed tools for available updates
- **`azd tool install [ids...]`** — installs tools by name, with `--all` for all recommended tools and `--dry-run` for preview
- **`azd tool upgrade [ids...]`** — upgrades installed tools, with `--all` and `--dry-run` support
- **`azd tool`** (bare) — interactive flow: shows installed/available tools, lets user select and install in one step

### First-Run Experience (Closes #7846)

- Middleware-triggered tool check on first workflow command (`azd init`, `azd up`, `azd deploy`, etc.)
- One-time prompt for new users; skippable via `--no-prompt`, `AZD_SKIP_FIRST_RUN=true`, or CI auto-detection
- CI environments (`CI`, `TF_BUILD`, `GITHUB_ACTIONS`) automatically bypass the prompt

### Update Check Middleware (partial #7848)

- Background version refresh with notification gating
- Non-intrusive update notifications when newer tool versions are available

---

## Commands

| Command | Description |
|---------|-------------|
| `azd tool` | Interactive tool discovery and installation |
| `azd tool list` | List all registered tools with status |
| `azd tool show <id>` | Show details for a specific tool |
| `azd tool check` | Check for available tool updates |
| `azd tool install [ids...]` | Install one or more tools (`--all`, `--dry-run`) |
| `azd tool upgrade [ids...]` | Upgrade one or more tools (`--all`, `--dry-run`) |

All commands support `--output json` for scripting and CI consumption.

---

## Built-in Tools (7)

| Tool | Category | Priority | Platforms |
|------|----------|----------|-----------|
| Azure CLI (`az`) | CLI | Recommended | winget / brew / apt |
| GitHub Copilot CLI | CLI | Recommended | winget / brew / npm |
| Azure Tools (VS Code) | Extension | Recommended | `code` CLI |
| Bicep (VS Code) | Extension | Recommended | `code` CLI |
| GitHub Copilot (VS Code) | Extension | Optional | `code` CLI |
| Azure MCP Server | Server | Optional | npm |
| azd AI Extensions | Library | Optional | azd extension |

---

## Architecture

```
cli/azd/
├── cmd/
│   ├── tool.go                          # Command definitions & actions
│   └── middleware/
│       ├── tool_first_run.go            # First-run experience middleware
│       ├── tool_first_run_test.go
│       ├── tool_update_check.go         # Update check middleware
│       └── tool_update_check_test.go
└── pkg/tool/
    ├── manager.go / manager_test.go     # Central orchestrator
    ├── manifest.go / manifest_test.go   # Tool definitions & registry
    ├── detector.go / detector_test.go   # Cross-platform tool detection
    ├── installer.go / installer_test.go # Install/upgrade execution
    ├── platform.go / platform_test.go   # OS/package-manager abstraction
    └── update_checker.go / update_checker_test.go  # Background version checks
```

---

## CI / Non-Interactive

- `--no-prompt` / `AZD_NON_INTERACTIVE=true` suppresses all prompts
- CI environments are auto-detected (`CI`, `TF_BUILD`, `GITHUB_ACTIONS`)
- JSON output (`--output json`) for machine consumption
- Positional args for `install`/`upgrade` — no interactive selection needed

---

## Testing

- **45 unit tests** across `pkg/tool/` with **83.9% statement coverage**
- All local CI checks pass: gofmt, go vet, golangci-lint, cspell, copyright headers, build, snapshots
- **31 files changed**, ~8,000 lines added

---

## Example Output

```
$ azd tool list

  Id                     Name                      Category    Priority      Status          Version
  ─────────────────────  ────────────────────────  ──────────  ────────────  ──────────────  ────────
  az-cli                 Azure CLI                 cli         recommended   Installed       2.67.0
  github-copilot-cli     GitHub Copilot CLI        cli         recommended   Not Installed
  vscode-azure-tools     Azure Tools (VS Code)     extension   recommended   Installed       0.9.0
  vscode-bicep           Bicep (VS Code)           extension   recommended   Installed       0.30.0
  vscode-github-copilot  GitHub Copilot (VS Code)  extension   optional      Not Installed
  azure-mcp-server       Azure MCP Server          server      optional      Not Installed
  azd-ai-extensions      azd AI Extensions         library     optional      Not Installed
```

---

## Issue References

- Closes #7845 — [EPIC] A: Core Command Group & Installation Engine
- Closes #7846 — [EPIC] B: First-Run Experience
- Related to #7676 — [EPIC] Getting Started: one-command setup
- Partially addresses #7848 — [EPIC] E: Proactive Capability Surfacing & Updates (update check middleware)

## Follow-Up Issues (deferred)

| Issue | Title | Why deferred |
|-------|-------|--------------|
| #7847 | Agent-First Programmatic API | MCP tool integration not yet implemented |
| #7848 | Proactive Capability Surfacing & Updates | Update middleware exists; additional surfacing deferred |
| #7849 | Registry Governance Documentation | Docs-only task |
| #7957 | Direct-download detection gap | PATH-based detector doesn't find non-standard installs |
| #7958 | VS Code extension visibility | Improve UX when `code` CLI is unavailable |
| #7959 | Custom telemetry for tool management | Product insight events not yet wired |
| #7960 | Azure MCP Server install strategy | Global npm vs npx decision pending |


cli/azd/pkg/tool/installer.go
```diff
@@ -0,0 +1,395 @@
...
+	return "code", args
+}
+
+// -----------------------------------------------------------------------
+// Helpers
+// -----------------------------------------------------------------------
+
+// splitCommand splits a whitespace-delimited command string into the
+// executable name and its arguments.
+func splitCommand(command string) (string, []string) {
```
jongio, line 690: **[HIGH]** `splitCommand` uses `strings.Fields` which splits on whitespace - it doesn't handle shell pipes or redirections. The az-cli Linux install strategy (`curl -sL ... | sudo bash`) will pass `|` as a literal arg to curl instead of creating a pipeline.

The codebase already has `exec.RunArgs` with shell execution support and `RunList()` for piped commands. `executeStrategy` should detect pipe/redirect characters and use shell execution, or use `RunList()`.

cli/azd/cover
```diff
@@ -0,0 +1,323 @@
+mode: set
```
jongio: **[HIGH]** This is a Go test coverage output file (`mode: set` header) - looks like it was accidentally committed. The `.gitignore` covers `cover-local.out` and `cover-ci-combined.out` but not a bare `cover` file. Remove this and consider adding `cover` to `.gitignore`.

cli/azd/pkg/tool/manifest.go
```diff
@@ -0,0 +1,283 @@
...
+func azureMCPServer() *ToolDefinition {
+	return &ToolDefinition{
+		Id:            "azure-mcp-server",
+		Name:          "Azure MCP Server",
+		Description:   "Model Context Protocol server for Azure resource interaction.",
+		Category:      ToolCategoryServer,
+		Priority:      ToolPriorityOptional,
+		Website:       "https://github.com/Azure/azure-mcp",
+		DetectCommand: "npx",
+		VersionArgs:   []string{"@azure/mcp@latest", "--version"},
```
jongio: **[MEDIUM]** Using `npx @azure/mcp@latest` as DetectCommand+VersionArgs means detection triggers an npm download on every call. `npx` with `@latest` fetches from the registry if the package isn't cached locally. Detection should be side-effect-free - consider `npm list -g @azure/mcp` or checking for a local binary instead.

cli/azd/cmd/tool.go
```diff
@@ -0,0 +1,833 @@
...
+func (a *toolUpgradeAction) Run(ctx context.Context) (*actions.ActionResult, error) {
+	a.console.MessageUxItem(ctx, &ux.MessageTitle{
+		Title:     "Upgrade Azure development tools (azd tool upgrade)",
+		TitleNote: "Upgrades installed tools to their latest versions",
+	})
+
+	var results []*tool.InstallResult
+	var err error
+
+	if len(a.args) > 0 {
```
jongio, line 633: **[MEDIUM]** The upgrade action calls `UpgradeTools()`/`UpgradeAll()` before creating the TaskList, so the task callbacks just report pre-computed results. Compare with the install action (line 396) which correctly does work inside each task callback. Users see `Upgrading...` spinners for already-completed operations. Move the upgrade calls inside each task's Action callback to match the install pattern.

cli/azd/cmd/root.go
```diff
@@ -525,6 +540,33 @@ func newRootCmd(
 	return cmd
 }
 
+// isWorkflowCommand reports whether the command is a primary workflow
+// command where a first-run tool check adds value.  Utility commands
+// (auth, config, env, show, etc.) are excluded so they are never
+// blocked by the first-run experience or update notifications.
```
jongio, line 552: **[MEDIUM]** `isWorkflowCommand` allocates a new `map[string]bool` on every call and hardcodes a command allowlist that'll go stale when new commands are added. The codebase already has `RootLevelHelpOption` annotations on commands (`CmdGroupStart`, `CmdGroupAzure`, etc.) - querying the command's group annotation would be more maintainable.

jongio (changes requested):
POC for the azd tool command group - structure is clean and well-decomposed across manifest, detector, installer, manager, and update checker.

Issues to address:
- pkg/tool/installer.go:389 - splitCommand can't handle shell pipes, breaking Linux az-cli install
- cli/azd/cover - test coverage output file accidentally committed
- pkg/tool/manifest.go:247 - MCP server detection via npx downloads packages on every detection call
- cmd/tool.go:522 - upgrade action pre-executes then creates a task list for display only
- cmd/root.go:546 - isWorkflowCommand uses a hardcoded map instead of existing command group annotations

cli/azd/pkg/tool/installer.go
wbreza: Fixed — added \containsShellOperators()\ detection. Commands containing \|\, \>\, \<\, \&\, or \;\ now route through \sh -c\ (Unix) or \cmd /C\ (Windows) instead of \splitCommand\. The az-cli Linux install (\curl ... | sudo bash\) works correctly now.

cli/azd/cover
wbreza: Fixed — removed the file with \git rm\ and added \cover\ to \.gitignore\.

cli/azd/pkg/tool/manifest.go
wbreza: Fixed — changed detection from \
px @azure/mcp@latest --version\ (triggers download) to \
pm list -g @azure/mcp --json\ (local check only). Updated the version regex to parse npm'\''s JSON output.

cli/azd/cmd/tool.go
wbreza: Fixed — restructured to match the install pattern. Tool resolution happens upfront, but \UpgradeTools()\ calls now happen inside each task'\''s Action callback so spinners reflect real-time progress.

cli/azd/cmd/root.go
wbreza: Fixed — replaced hardcoded map with group annotation check using \descriptor.Options.GroupingOptions.RootLevelHelp\. Checks \CmdGroupStart\, \CmdGroupAzure\, and \CmdGroupBeta\ so new commands in those groups automatically get the middleware.

spboyer (approved):

spboyer:
@jongio - review again please.

cli/azd/pkg/tool/detector.go
```diff
@@ -0,0 +1,340 @@
...
+				"running %s: %w", tool.DetectCommand, err,
+			)
+			return status
+		}
+
+		// Non-zero exit: the binary exists, so try to parse
+		// any captured output.
+	}
+
+	status.Installed = true
```
jongio: `detectCommandBased` sets `Installed = true` after the binary PATH check, but for package-based tools like MCP server (`npm` as DetectCommand) and azd AI extensions (`azd`), having the binary on PATH doesn't mean the specific package is installed. When `npm list -g @azure/mcp --json` exits non-zero, the regex won't match but `Installed` is still true.

`detectExtension` handles this correctly by gating on the regex match. Same pattern here:

```suggestion
	version := matchVersion(
		result.Stdout+result.Stderr, tool.VersionRegex,
	)
	if version != "" {
		status.Installed = true
		status.InstalledVersion = version
	}
```

cli/azd/cmd/tool.go
```diff
@@ -0,0 +1,852 @@
...
+	toolInfo = append(toolInfo, []string{"Installed Version", ":", installedVersion})
+
+	if err := writeSection("Tool Information", toolInfo); err != nil {
+		return err
+	}
+
+	// Install Strategies
+	if len(toolDef.InstallStrategies) > 0 {
+		var strategyRows [][]string
+		for platform, strategy := range toolDef.InstallStrategies {
```
jongio: Map iteration order is non-deterministic - install strategies display in random order between runs. Consider using `slices.Sorted(maps.Keys(toolDef.InstallStrategies))` for consistent output.

jongio (commented):
Previous comments addressed. Two new findings on the latest code:
- pkg/tool/detector.go - `detectCommandBased` reports Server/Library tools (MCP server, azd AI extensions) as installed when only the shared binary (npm, azd) is on PATH, even when the specific package isn't installed
- cmd/middleware/tool_first_run.go and tool_update_check.go have no unit tests despite complex skip logic and background goroutine management

One nit:
- cmd/tool.go - install strategies in `show` command display in non-deterministic map iteration order

cli/azd/pkg/tool/detector.go
wbreza: Fixed — \detectCommandBased\ now gates \Installed=true\ on the version regex match, consistent with \detectExtension\. If VersionRegex is set but doesn't match, the tool is reported as not installed. Tools with no VersionRegex still fall back to binary-exists check for backward compatibility.

cli/azd/cmd/tool.go
wbreza: Fixed — using \slices.Sorted(maps.Keys(toolDef.InstallStrategies))\ for deterministic platform display order.

cli/azd/cmd/middleware/tool_first_run.go
```diff
@@ -0,0 +1,327 @@
...
+	detectSpinner := uxlib.NewSpinner(&uxlib.SpinnerOptions{
+		Text:        "Detecting tools...",
+		ClearOnStop: true,
+	})
+	if err := detectSpinner.Run(ctx, func(ctx context.Context) error {
+		var detectErr error
+		statuses, detectErr = m.manager.DetectAll(ctx)
+		return detectErr
+	}); err != nil {
+		m.markCompleted()
```
jongio: `markCompleted()` is called when detection fails, permanently suppressing the first-run experience. If `DetectAll` fails (transient PATH issue, timeout, permission error), the user never gets another chance at the welcome flow. Only mark complete when the user explicitly dismisses (cancel, "no") or after a successful flow - not on infrastructure failures.

cli/azd/pkg/tool/installer.go
```diff
@@ -0,0 +1,436 @@
...
+
+	return result, nil
+}
+
+// ensurePlatform lazily detects the current platform and caches the
+// result for subsequent calls.
+func (i *installer) ensurePlatform(
+	ctx context.Context,
+) (*Platform, error) {
+	if i.platform != nil {
```
jongio, line 289: This check-then-set on `i.platform` isn't synchronized. If `Install`/`Upgrade` are ever called concurrently, this races. `UpdateChecker.getCacheFilePath` already uses `sync.Once` for the same lazy-init pattern - consider matching that here.

cli/azd/pkg/tool/update_checker.go
```diff
@@ -0,0 +1,461 @@
...
+func (uc *UpdateChecker) Check(
+	ctx context.Context,
+	tools []*ToolDefinition,
+) ([]*UpdateCheckResult, error) {
+	statuses, err := uc.detector.DetectAll(ctx, tools)
+	if err != nil {
+		return nil, fmt.Errorf("detecting installed tools: %w", err)
+	}
+
+	existing, _ := uc.GetCachedResults()
```
jongio, line 180: Cache read errors are silently discarded. If the file exists but is corrupt (partial write, disk error), this drops previously-known latest versions and overwrites with empty data. At minimum `log.Printf` the error like other non-fatal failures in this package. The POC doesn't have a remote API yet, so the cache is the only source of latest-version data.

jongio (commented):
Previous comments addressed. A few new edge cases:
- cmd/middleware/tool_first_run.go:170 - detection failure permanently kills the first-run experience
- pkg/tool/installer.go:194 - ensurePlatform lazy cache has no synchronization
- pkg/tool/update_checker.go:176 - cache read errors silently discarded

One test gap: no coverage for the commandBased false-positive fix (Server/Library tool with VersionRegex set + non-matching output should report not installed).

cli/azd/cmd/middleware/tool_first_run.go
wbreza: Fixed — `markCompleted()` is now only called on explicit user skip or successful flow completion. Detection failures and Ctrl+C cancellations no longer suppress future first-run prompts, so transient errors don't permanently block the welcome experience.

cli/azd/pkg/tool/installer.go
wbreza: Fixed — replaced the unsynchronized check-then-set with `sync.Once` for thread-safe lazy platform initialization. Concurrent Install/Upgrade calls now safely share the detected platform.

cli/azd/pkg/tool/update_checker.go
wbreza: Fixed — cache read errors are now logged via `log.Printf`. Corrupt cache files (unmarshal failures) are deleted with `os.Remove` so they regenerate on the next check cycle. File-not-found remains silent as expected.

cli/azd/pkg/tool/manifest.go
```diff
@@ -0,0 +1,283 @@
...
+		Name: "Azure Tools VS Code Extension",
+		Description: "VS Code extension for browsing and managing " +
+			"Azure resources.",
+		Category:      ToolCategoryExtension,
+		Priority:      ToolPriorityRecommended,
+		Website:       "https://marketplace.visualstudio.com/items?itemName=ms-azuretools.vscode-azureresourcegroups",
+		DetectCommand: "code",
+		VersionArgs:   []string{"--list-extensions", "--show-versions"},
+		VersionRegex:  `ms-azuretools\.vscode-azureresourcegroups@(\d+\.\d+\.\d+)`,
+		InstallStrategies: allPlatforms(InstallStrategy{
```
jongio, line 195: VS Code extension strategies set `PackageManager: "code"` + `InstallCommand` but no `PackageId`. Two consequences:

1. `buildCodeCommand` in installer.go is unreachable dead code - `buildCommand`'s priority chain always hits the `InstallCommand` path before reaching `PackageManager + PackageId`.

2. Upgrades don't get `--force`. The upgrade path in `buildCommand` requires `PackageManager != ""` AND `PackageId != ""`. Without `PackageId`, upgrades fall through to the `InstallCommand` path, which runs the same install command without `--force`. Since `code --install-extension` without `--force` won't update an already-installed extension, `azd tool upgrade` is silently a no-op for VS Code extensions.

Fix for all three: add `PackageId` with the extension identifier and drop `InstallCommand`:
```go
InstallStrategies: allPlatforms(InstallStrategy{
    PackageManager: "code",
    PackageId:      "ms-azuretools.vscode-azureresourcegroups",
}),
```

Same for vscodeBicep (`"ms-azuretools.vscode-bicep"`) and vscodeGitHubCopilot (`"GitHub.copilot"`). This makes `buildCodeCommand` reachable, upgrades get `--force`, and the manager availability check gives a friendly error when VS Code isn't installed.

cli/azd/cmd/tool.go
```diff
@@ -0,0 +1,855 @@
...
+		}
+
+		taskList.AddTask(uxlib.TaskOptions{
+			Title: fmt.Sprintf("Installing %s", toolDef.Name),
+			Action: func(setProgress uxlib.SetProgressFunc) (uxlib.TaskState, error) {
+				results, installErr := a.manager.InstallTools(ctx, []string{capturedID})
+				if installErr != nil {
+					return uxlib.Error, installErr
+				}
+				if len(results) > 0 && !results[0].Success {
```
jongio, line 224: Task callbacks check `results[0].Success` but `InstallTools` returns dependency results first. If a tool has uninstalled dependencies (currently only `azd-ai-extensions` depends on `az-cli`), `results[0]` is the dependency's result, not the requested tool's. When the dependency succeeds but the primary tool fails, the task incorrectly reports success.

The middleware's `offerInstall` in `tool_first_run.go` already handles this correctly with a `for _, r := range results` loop.

Fix - find the result matching the requested tool:
```go
for _, r := range results {
    if r.Tool != nil && r.Tool.Id == capturedID {
        if !r.Success {
            return uxlib.Error, r.Error
        }
        break
    }
}
```

Apply to both occurrences: `toolAction.Run` (~line 224) and `toolInstallAction.Run` (~line 402).

cli/azd/cmd/middleware/tool_update_check.go
```diff
@@ -0,0 +1,176 @@
...
+		m.triggerBackgroundCheckIfNeeded(ctx)
+	}
+
+	return result, err
+}
+
+// showNotificationIfNeeded displays a one-line update notification
+// when cached results indicate that tool updates are available and
+// the current console session is interactive.
+func (m *ToolUpdateCheckMiddleware) showNotificationIfNeeded(ctx context.Context) {
```
jongio, line 76: `showNotificationIfNeeded` calls `HasUpdatesAvailable` then `MarkUpdateNotificationShown` but never calls `ShouldShowNotification`. That method exists on `UpdateChecker` but isn't exposed through `Manager`, so it's dead code. This means the update notification is shown on every interactive invocation that has cached updates, rather than once per check cycle as intended.

Fix: add a `ShouldShowNotification` method to `Manager` (delegating to `UpdateChecker.ShouldShowNotification`), then call it in `showNotificationIfNeeded` before the `HasUpdatesAvailable` check:
```go
func (m *ToolUpdateCheckMiddleware) showNotificationIfNeeded(ctx context.Context) {
    if m.shouldSkipNotification() {
        return
    }
    if !m.manager.ShouldShowNotification(ctx) {
        return
    }
    // ... rest of the method
}
```

jongio (changes requested):
Round 4 - three new findings in code paths previous rounds didn't cover.

Smaller items (not blocking):

- `detectCommandBased` test gap: there's no test case for "VersionRegex is set but output doesn't match - tool should NOT be installed." The round 3 fix (gating `Installed` on regex match) works correctly, but a regression test in `TestDetectTool_CommandBased` would lock it down.

- `HasUpdatesAvailable` (update_checker.go) counts uninstalled tools as needing updates. The condition `latest != s.InstalledVersion` is true when the tool isn't installed (`InstalledVersion` is empty) but `latest` is non-empty. Currently unreachable in the POC since no remote API populates `LatestVersion`, but it'll bite when that's added. Adding `s.Installed` to the guard fixes it.

- Middleware test coverage: `tool_first_run.go` (327 lines) and `tool_update_check.go` (176 lines) still have zero tests despite complex skip logic, CI detection, and background goroutine management. Mentioned this in round 2 body but calling it out again.

cli/azd/cmd/tool.go
wbreza: Fixed — both locations now use a `for _, r := range results` loop to find the result matching the requested tool ID, consistent with the pattern already used in `tool_first_run.go` middleware.

jongio (commented):
Prior feedback addressed - the `results[0]` fix now iterates to find the matching tool by ID, VS Code extensions correctly use `PackageId`, and the middleware notification guard is in place. Two notes on the new direct download code.

cli/azd/pkg/tool/installer.go
```diff
@@ -0,0 +1,653 @@
...
+	}
+
+	return nil
+}
+
+// validateChecksum verifies the file at filePath against the
+// expected checksum. When the checksum is empty (both Algorithm and
+// Value are ""), validation is silently skipped.
+func validateChecksum(filePath string, checksum Checksum) error {
+	if checksum.Algorithm == "" && checksum.Value == "" {
```
jongio, line 502: The guard skips validation only when both `Algorithm` and `Value` are empty. If someone configures only one field (e.g., `Algorithm: "sha256"` with empty `Value`, or `Value` set with empty `Algorithm`), the behavior is confusing - either comparing against an empty string or hitting the default branch with an empty algorithm name in the error message.

Worth adding an explicit check for the partial config case so the caller gets a clear error about the misconfiguration.

cli/azd/pkg/tool/installer.go
```diff
@@ -0,0 +1,653 @@
...
+			checksum.Value, actual,
+		)
+	}
+
+	return nil
+}
+
+// toolInstallDir returns the directory where directly downloaded
+// tools are placed. It defaults to ~/.azd/tools/.
+func toolInstallDir() (string, error) {
```
jongio, line 566: This hardcodes `~/.azd/tools/` but the rest of azd respects `AZD_CONFIG_DIR` via `config.GetUserConfigDir()`. If a user sets that env var, their config goes to the custom path but downloaded tools still land in the default location.

Consider using `config.GetUserConfigDir()` here so everything stays consistent.

jongio (commented):
Three items from prior rounds that still look unaddressed, plus one new observation.

The checksum validation and AZD_CONFIG_DIR fixes in the latest commit look good - those close out the round 4 feedback cleanly.

Still open:
- HasUpdatesAvailable can count uninstalled tools as having updates (details inline)
- No regression test for the VersionRegex no-match path in detectCommandBased
- Zero test coverage on both middleware files (	ool_first_run.go, 	ool_update_check.go)

cli/azd/pkg/tool/update_checker.go
```diff
@@ -0,0 +1,465 @@
...
+
+	statuses, err := uc.detector.DetectAll(ctx, toolDefs)
+	if err != nil {
+		return false, 0, fmt.Errorf("detecting tools: %w", err)
+	}
+
+	count := 0
+	for _, s := range statuses {
+		latest, ok := candidates[s.Tool.Id]
+		if ok && latest != s.InstalledVersion {
```
jongio: This condition fires when `InstalledVersion` is empty (tool not installed) but `latest` has a value from the cache. Once the remote version API is wired up, every uninstalled tool will show as 'has updates available'.

Fix: `if ok && s.Installed && latest != s.InstalledVersion`

cli/azd/pkg/tool/detector_test.go
```diff
@@ -0,0 +1,682 @@
...
+
+			if tt.expectError {
+				require.Error(t, status.Error)
+				if tt.expectErrContain != "" {
+					assert.Contains(t, status.Error.Error(),
+						tt.expectErrContain)
+				}
+			} else {
+				assert.NoError(t, status.Error)
+			}
```
jongio, line 779: The fix for 'VersionRegex set but output doesn't match = not installed' landed, but there's no test case locking it down. A case like `{name: "NoRegexMatch", tool: Tool{VersionRegex: "v\\d+", ...}, commandOutput: "no version here", wantInstalled: false}` would prevent regressions.

cli/azd/cmd/middleware/tool_first_run.go
```diff
@@ -0,0 +1,327 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
jongio, line 1: This file (327 lines) and 	ool_update_check.go (176 lines) still have zero tests. The skip logic (env var checks, CI detection, config persistence) and the background goroutine + timeout pattern in the update checker are the kind of things that break silently without coverage.

cli/azd/pkg/tool/update_checker.go
wbreza: Fixed — added `s.Installed` guard so uninstalled tools with cached latest versions aren't counted as having updates.

vhvb1989 (commented):
## Review Feedback

Thanks for this feature! I have a few questions and suggestions across different areas. Posting them as a single review for easier tracking.

---

### 1. VS Code Extensions Visibility (cli/azd/cmd/tool.go — `toolListAction.Run`)

Consider filtering or visually distinguishing VS Code extensions when `code` is not on PATH. Currently `azd tool list` shows them as "Not Installed" (same as a CLI you just haven't installed yet), and `azd tool install` fails with a package-manager-unavailable error. Options: (1) hide extensions whose prerequisite isn't available, (2) show a distinct status like "Unavailable (requires VS Code)" instead of "Not Installed", or (3) add a `--category` filter flag so users can scope the list.

---

### 2. Azure MCP Server Install Strategy (cli/azd/pkg/tool/manifest.go — `azureMCPServer()`)

Questions about the Azure MCP Server install strategy:

1. **Global install vs npx**: The MCP server is installed globally via `npm install -g @azure/mcp`, but most MCP server configurations in VS Code use `npx` (which doesn't require a global install). What's the benefit of a global install here? Is this the recommended approach from the Azure MCP team?

2. **No editor configuration**: Installing the npm package globally doesn't actually configure it anywhere — the user still needs to manually wire it up in their VS Code MCP settings (or whatever agent tool they use). So `azd tool install azure-mcp-server` installs a package that can't be used until additional manual steps are taken. Should `azd tool` also help with configuration, or at minimum surface a "next steps" message?

3. **Global vs local conflict**: If a project already has `@azure/mcp` as a local dev dependency (in the project's `node_modules`), a global install could lead to version conflicts or confusion about which version is being used. Has this been considered?

---

### 3. First-Run Experience for Existing Users (cli/azd/cmd/middleware/tool_first_run.go — `shouldSkip`)

The first-run experience will trigger for **all existing azd users** after upgrading, not just new installations. The only gate is the absence of `tool.firstRunCompleted` in user config, which is a new key that no existing user will have.

Consider differentiating between truly new users and existing users who just upgraded. For example:
- Check if other user config keys already exist (indicating a returning user) and auto-set `tool.firstRunCompleted` for them
- Check if `~/.azd` predates this version
- Only show this for fresh azd installations where no prior config exists

For long-time users who already have their tools set up, being prompted with "Welcome to Azure Developer CLI! 🚀" and a tool check on their next `azd up` could feel disruptive rather than helpful.

Additionally:
- The opt-out mechanism is currently only via `AZD_SKIP_FIRST_RUN` env var, which is not discoverable. The welcome message should mention how to skip/disable this feature.
- A user-config-based opt-out (e.g., `azd config set tool.firstRun.enabled false`) would be more natural and consistent with azd patterns than requiring a system environment variable. The env var is fine for CI, but for users who just don't want this feature, `azd config` is the right place.

---

### 4. Global Flags as No-Ops (cli/azd/cmd/tool.go)

Minor: `azd tool` accepts global flags like `-e/--environment` that it doesn't use. Is there any risk that passing `-e myenv` could trigger environment creation or resolution as a side effect before the tool action runs? Might be worth verifying that unused global flags are truly no-ops for commands that don't need them.

---

### 5. Telemetry for Product Insights (cli/azd/cmd/tool.go, cli/azd/pkg/tool/)

The generic telemetry middleware will track command-level usage and errors, but there's no custom telemetry to answer key product questions:

- Which tools are most/least installed across users? (detection results)
- Which tools are users choosing to install, and which do they skip?
- What install strategies are being used (brew vs winget vs npm vs manual)?
- What's the success/failure rate per tool and per strategy?
- In the first-run experience, do users accept or decline the tool check? Which tools do they select?

Adding span attributes for tool IDs, install strategies, and outcomes would make this feature's impact measurable. Without it, we'll know people ran `azd tool install` but not what they installed or whether it worked.

cli/azd/cmd/middleware/tool_first_run.go
```diff
@@ -0,0 +1,327 @@
...
+)
+
+// configKeyFirstRunCompleted is the user-config path that records
+// the timestamp of a completed first-run experience.
+const configKeyFirstRunCompleted = "tool.firstRunCompleted"
+
+// envKeySkipFirstRun is the environment variable that, when set to
+// "true", suppresses the first-run tool check entirely.
+const envKeySkipFirstRun = "AZD_SKIP_FIRST_RUN"
+
```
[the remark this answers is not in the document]
wbreza: Fixed — added `AZD_SKIP_FIRST_RUN` to `docs/environment-variables.md` in the General Configuration section.

wbreza:
Thanks for the thorough review @vhvb1989! Here's how I'm addressing each point:

**1. VS Code Extensions Visibility** — Great point. Created follow-up issue #7958 to investigate distinct status vs hiding vs `--category` filter.

**2. Azure MCP Server Install Strategy** — Global npm install is by-design for the initial release as the simplest path. Created follow-up issue #7960 to revisit with the Azure MCP team (npx, editor config, local/global conflicts).

**3. First-Run for Existing Users** — Updated the messaging from "Welcome to Azure Developer CLI!" to "Checking your Azure development tools..." — works for both new and existing users. Added skip/disable hint mentioning `AZD_SKIP_FIRST_RUN` and `azd config set`.

**4. Global Flags** — Fixed — hidden `-e/--environment` and `-C/--cwd` flags from `azd tool` commands since they're not relevant.

**5. Telemetry** — Created follow-up issue #7959 to add custom telemetry (tool IDs, strategies, outcomes, first-run selections).

vhvb1989 (approved):

vhvb1989:
@kristenwomack — Heads up: this PR introduces a first-run tool check middleware that triggers on all workflow commands (including `azd init`). On the first invocation, users see a tool detection banner + an offer to install missing recommended tools before their actual command runs.

I wanted to flag a potential overlap with the UX revamp epic (#7678), which includes redesigning the `azd init` experience with context-aware template selection, agentic init integration, and collision detection. If both land, a user running `azd init` for the first time could hit two sequential onboarding experiences:

1. This PR's tool check ("Checking your Azure development tools..." → detection → install prompt)
2. The redesigned `azd init` flow from the epic

Do we need to coordinate between these two to avoid stacking multiple first-time experiences on the same command? For example, should the tool check be excluded from `azd init` if the init redesign will have its own onboarding, or should they be combined into a single cohesive first-run experience?

cli/azd/cmd/tool.go
```diff
@@ -0,0 +1,1206 @@
...
+		Use:   "tool",
+		Short: "Manage Azure development tools.",
+		Long:  "Discover, install, upgrade, and check status of Azure development tools.",
+	}
+
+	// Hide global flags that are not relevant to tool commands.
+	// Tool commands manage local development tools and never need
+	// environment selection or working directory overrides.
+	// PersistentPreRun applies to this command and all subcommands.
+	toolCmd.PersistentPreRun = func(cmd *cobra.Command, _ []string) {
```
jongio: Setting PersistentPreRun here causes cobra to skip the root command's PersistentPreRunE (cobra doesn't chain them; EnableTraverseRunHooks isn't set in the main binary). That means the --cwd directory-change logic and the platform-warning in root.go:87 won't run for azd tool commands.

Since you're already hiding --cwd, this is mostly academic, but if someone explicitly passes --cwd /foo to a tool command it'll be silently ignored. Might be cleaner to hide the flags directly in the ActionDescriptor setup instead of via PersistentPreRun.

jongio (commented):
Incremental review of 12 new commits since my last pass.

All three items from my previous review are addressed:
- HasUpdatesAvailable now checks s.Installed before comparing versions
- VersionRegexSetButNoMatch test case locks down the regex-no-match path
- Both middleware files have solid test coverage (551 lines total)

The containsShellOperators fix (single & to &&) and console handle wiring look correct. One new item inline.

jongio (approved):
All prior feedback addressed. The PersistentPreRun item is low-severity and fine as a follow-up. Approving.

jongio (approved):
Addresses my previous feedback - the PersistentPreRun override is removed. Looks good.

kristenwomack (approved):
The `azd tool` surface is clean and the iteration with @jongio and @vhvb1989 has tightened things up. One coordination note and a couple of small UX items below.

### 🟡 One thing to flag before merge

**First-run on `azd init` overlaps with the init UX revamp (#7678).**

@vhvb1989 flagged this directly. Today the first-run middleware fires on every workflow command, and `CmdGroupStart` includes `init`. So a brand-new user running `azd init` for the first time hits two onboarding moments back to back:
1. "Checking your Azure development tools..." then detection then a multi-select install
2. The redesigned `init` flow once #7678 lands (template selection, agentic init, collision detection)

My take: let''s prioritize landing this PR first. #7678 still has a lot of feedback to work through, and the init redesign should own the first-time experience there once it''s ready. We can revisit the overlap (exclude `init` from first-run, or fold both into one cohesive flow) as part of #7678. Does that work?

### 🟢 Small UX items, fine as follow-ups

1. The banner says "Let''s make sure your development environment is up to date," then the prompt asks "Would you like to check your Azure development tools?" The prompt feels redundant after the banner. Something more action-shaped like "Run check?" or "Continue?" might land better.
2. All recommended tools are pre-selected in the multi-select. Good for the happy path, but a user smashing Enter installs 5+ things across winget, brew, npm, and `code`. Worth watching how this lands once telemetry is wired up. The #7959 telemetry follow-up should put first-run outcomes (accepted, declined, partial install, failure) on a dashboard so we can tune the default with real data.

### Kudos
- Skip-logic order (env var, `--no-prompt`, CI, non-TTY, completed flag) is well thought out.
- Detection failure never blocks the user''s command.
- CI auto-skip is the right call.
- Thanks for filing follow-ups for VS Code visibility (#7958), MCP install strategy (#7960), and telemetry (#7959).

Approving, with the init coordination as the one open question.

kristenwomack:
> @kristenwomack — Heads up: this PR introduces a first-run tool check middleware that triggers on all workflow commands (including `azd init`). On the first invocation, users see a tool detection banner + an offer to install missing recommended tools before their actual command runs.
> 
> I wanted to flag a potential overlap with the UX revamp epic (#7678), which includes redesigning the `azd init` experience with context-aware template selection, agentic init integration, and collision detection. If both land, a user running `azd init` for the first time could hit two sequential onboarding experiences:
> 
> 1. This PR's tool check ("Checking your Azure development tools..." → detection → install prompt)
> 2. The redesigned `azd init` flow from the epic
> 
> Do we need to coordinate between these two to avoid stacking multiple first-time experiences on the same command? For example, should the tool check be excluded from `azd init` if the init redesign will have its own onboarding, or should they be combined into a single cohesive first-run experience?

Thank you! I think we prioritize this one--mentioned in my review comment.

hemarina (approved):

kristenwomack:
Thanks @wbreza, nice surface, and the iteration with @jongio and @vhvb1989 tightened it up well. Three things I''d like to see in this PR before promoting to customers:

1. **CHANGELOG entry.** `cli/azd/CHANGELOG.md` `1.25.0-beta.1 → Features Added` is empty. This is a big, customer-visible change, and folks should see it called out in the next release notes.
2. **Feature stage: ship as Beta.** Right now `tool` registers under `CmdGroupManage` (same tier as stable `config`), but `extension` itself is still Alpha in `docs/reference/feature-status.md`. I''d recommend marking this Beta, then graduating to Stable once telemetry from #7959 is in and we''ve seen early customer reception. That gives us room to iterate on naming and the init overlap without breaking-change pressure.
3. **Add a row to `docs/reference/feature-status.md`.** `.github/instructions/documentation.instructions.md` calls this out as required for new commands.

---

**On the `azd init` first-run overlap (`cli/azd/cmd/root.go`, around `isWorkflowCommand` / `CmdGroupStart`):**

As I mentioned, we''ll hold on #7678 until after the `azd tool` command group ships, so the init UX revamp can account for the first-run tool-check banner from the start. Flagging it here so the sequencing is captured on the PR for anyone catching up later.

---

**On naming and help text (`cli/azd/cmd/tool.go`, command Short/Long):**

Two small things on the user-facing surface:

- The Short and Long ("Manage Azure development tools" and "Discover, install, upgrade, and check status of Azure development tools") don''t distinguish `azd tool` from `azd extension`. The bare `azd tool` flow can install azd extensions, so the relationship deserves to be explicit. Worth one editorial pass to clarify "external developer prerequisites" vs. "azd extensibility surface."
- Worth a name review before this stabilizes. `tool` is generic enough that `azd setup`, `azd doctor`, or `azd deps` might communicate "developer prerequisites" more precisely. Happy to pull a few of us together to talk through it. Cheaper to rename now than after customers are using it.

---

**On the package layout (`cli/azd/pkg/tool/`):**

Heads up that `pkg/tool/` (new, singular) lives next to `pkg/tools/` (existing, the az, bicep, docker, gh, and kubectl wrappers per `cli/azd/AGENTS.md`). Future contributors are going to mis-import these constantly. Consider `pkg/toolmgr/`, `pkg/toolcatalog/`, or `pkg/devtools/` to disambiguate.

---

**Design doc and governance (soft ask):**

Could we add a short design doc under `cli/azd/docs/design/` (similar to `azd-update.md`) covering tool-registry governance, who can add, remove, or reprioritize tools, and named contacts for Azure CLI, Bicep, gh CLI, VS Code extensions, and Azure MCP? It helps the deferred follow-ups (#7849, #7960) start with shared context, and gives partner teams a doc to point at.

jongio (approved):
Re-reviewed against the rebased HEAD (d48d0c4). All prior findings confirmed addressed:

- PersistentPreRun override removed
- Shell pipe handling fixed via containsShellOperators routing
- Error wrapping consistent with fmt.Errorf %w
- Telemetry mapping added for ErrToolUpgradeFailed

Architecture is clean with proper separation (Manager/Detector/Installer/UpdateChecker), good DI via IoC, and middleware correctly gates on CI/non-interactive/no-prompt. Test coverage looks solid.

rajeshkamal5050:
> CHANGELOG entry. cli/azd/CHANGELOG.md 1.25.0-beta.1 → Features Added is empty. This is a big, customer-visible change, and folks should see it called out in the next release notes.

@wbreza Are we ready to release it in Beta/public-preview this week? I feel this needs more testing and dogfooding?

@v-xuto can we prioritize and test the azd tool feature this week?

v-xuto:
> @v-xuto can we prioritize and test the azd tool feature this week?

Got it. We will begin testing `azd tool feature` today.

v-xuto:
@rajeshkamal5050 We completed the related testing for the azd tool feature and filed two issues: #8069, #8070.

The following table shows the detailed test results:
<html xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:dt="uuid:C2F41010-65B3-11d1-A29F-00AA00C14882"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=OneNote.File>
<meta name=Generator content="Microsoft OneNote 15">
</head>

<body lang=en-US style='font-family:Calibri;font-size:11.0pt'>


<div style='direction:ltr'>


Command | Windows | Linux | Mac
-- | -- | -- | --
azd tool | PASS | PASS | PASS
azd tool list | PASS | PASS | PASS
azd tool show   <id> | PASS | PASS | PASS
azd tool check | #8070 | #8070 | #8070
azd tool install   [ids...] | #8069 | #8069 | PASS
azd tool upgrade   [ids...] | PASS | PASS | PASS



</div>


</body>

</html>

Issue #7436, opened 2026-04-02
ScriptExecutor interface and ScriptLanguage type


wbreza:
## ScriptExecutor Interface and ScriptLanguage Type

**Part of #7435 — Multi-Language Hook Support**

### User Story

As a template author, I want azd to understand that my hook script is written in a specific programming language so that it can handle execution appropriately instead of treating everything as a shell script.

### Solution Approach

Create `pkg/tools/language/executor.go` (NEW):

- Define `ScriptLanguage` type (`string`) with constants: `ScriptLanguageUnknown` (`""`), `ScriptLanguagePython` (`"python"`), `ScriptLanguageJavaScript` (`"js"`), `ScriptLanguageTypeScript` (`"ts"`), `ScriptLanguageDotNet` (`"dotnet"`).
- Define `ScriptExecutor` interface:
  - `Language() ScriptLanguage`
  - `Prepare(ctx context.Context, scriptPath string) error` — runtime check, dependency install, build
  - `Execute(ctx context.Context, scriptPath string, options tools.ExecOptions) (exec.RunResult, error)`
- Implement `GetExecutor(language ScriptLanguage, ...) (ScriptExecutor, error)` factory. Accept dependencies via parameters (not global state) — specifically `exec.CommandRunner`. Phase 1: Python returns real executor; JS/TS/DotNet return `ErrUnsupportedLanguage` sentinel.
- Implement `InferLanguageFromPath(path string) ScriptLanguage` mapping `.py` → Python, `.js` → JS, `.ts` → TS, `.cs` → DotNet, unknown → `ScriptLanguageUnknown`.

Create `pkg/tools/language/executor_test.go` (NEW):
- Table-driven tests for `InferLanguageFromPath` covering all extensions + unknown.
- Tests for `GetExecutor` returning real executor for Python, `ErrUnsupportedLanguage` for others.

**Files:** `pkg/tools/language/executor.go` (NEW), `pkg/tools/language/executor_test.go` (NEW)

### Acceptance Criteria

- [ ] Package compiles with `go build ./pkg/tools/language/...`
- [ ] `InferLanguageFromPath` correctly maps `.py`, `.js`, `.ts`, `.cs`, and unknown extensions
- [ ] `GetExecutor` returns `ErrUnsupportedLanguage` for JS, TS, DotNet in Phase 1
- [ ] `ScriptExecutor` interface is compatible with `tools.ExecOptions` from `pkg/tools/script.go`
- [ ] All unit tests pass

[...]

Issue #7445, opened 2026-04-02
End-to-end Python hook tests


wbreza:
## End-to-End Python Hook Tests

**Part of #7435 — Multi-Language Hook Support**

### User Story

As a developer maintaining the hook system, I want comprehensive end-to-end tests that verify the full Python hook lifecycle so regressions are caught automatically.

### Solution Approach

**Test scenario 1 — Happy path:**
1. Create test project with:
   - `azure.yaml` with `preprovision` Python hook
   - `scripts/setup.py` that reads an env var and writes it to a marker file
   - `scripts/requirements.txt` with a simple dependency
2. Verify:
   - Hook detected as Python language hook
   - Virtual environment created (`.venv` exists)
   - Dependencies installed
   - Script executes successfully
   - Environment variables available to script
   - Exit code 0

**Test scenario 2 — Auto-detect vs explicit:**
- Test with `run: scripts/setup.py` (auto-detect from `.py`)
- Test with `run: scripts/setup` + `language: python` (explicit)
- Both should execute identically

**Test scenario 3 — Error cases:**
- Missing Python runtime → clear error with install URL
- Missing script file → clear error
- Script exits non-zero → error propagated
- `continueOnError: true` + failing script → no error, continues

**Test scenario 4 — Service-level hooks:**
- Python hook defined under a service, not project root
- Verify project discovery scoped to service directory

**Location:** Follow existing patterns in `test/functional/testdata/samples/` for fixtures, `test/functional/` or `pkg/ext/` for test files.

**Files:** `test/functional/` or `pkg/ext/hooks_runner_test.go` (NEW/MODIFY), test fixtures

### Acceptance Criteria

- [ ] Tests pass on CI (requires Python installed in CI environment)
- [ ] Validates complete flow: YAML → language detection → venv → install → execute → output
- [ ] Tests both success and error paths
- [ ] Tests environment variable passing
- [ ] Tests both project-level and service-level hooks
- [ ] Tests auto-detect and explicit `language:` field

### References

- Contributes to: #7435

wbreza:
Resolved by PR #7451 (https://github.com/Azure/azure-dev/pull/7451). Added \python_hooks_e2e_test.go\ (961 lines) covering venv creation, dependency installation, error scenarios (missing Python, invalid deps), service-level hooks, auto-detect vs explicit kind, and environment variable passing.

wbreza:
Completed via PR #7451

Issue #7444, opened 2026-04-02
Schema documentation and error messages for language hooks


wbreza:
## Schema Documentation and Error Messages

**Part of #7435 — Multi-Language Hook Support**

### User Story

As a template author, I want clear documentation on how to configure language hooks in azure.yaml so I can adopt the feature without guessing.

### Solution Approach

1. **Schema updates** (if JSON schema files exist in `resources/`):
   - Add `language` field to HookConfig schema (enum: `python`, `js`, `ts`, `dotnet`).
   - Add `dir` field to HookConfig schema (string, optional).
   - Update `run` field description to mention language script support.

2. **Code documentation**:
   - GoDoc comments on `HookConfig` struct (mention `language` field).
   - GoDoc on `validate()` method (document resolution order).
   - GoDoc on `ScriptExecutor` interface (document contract).

3. **Error messages** (verify all use `ErrorWithSuggestion`):
   - Runtime not found: `"Python runtime not found. Install Python 3.7.6+ from https://..."`
   - Dependency install failure: `"Failed to install Python dependencies for hook '%s': %w"`
   - Inline rejection: `"Inline scripts are not supported for language hooks (language: %s). Use a file path instead."`
   - Build failure (future, for TS/.NET): `"Failed to build hook '%s': %w"`

4. **Example configurations** to validate:
   ```yaml
   # Auto-detect Python from .py extension
   hooks:
     preprovision:
       run: scripts/setup.py

   # Explicit language override
   hooks:
     preprovision:
       run: scripts/setup
       language: python

   # Platform-specific language hooks
   hooks:
     postdeploy:
       windows:
         run: scripts/notify.py
       posix:
         run: scripts/notify.py
   ```

**Files:** `resources/` schema files (MODIFY), code comments across new files

### Acceptance Criteria

- [ ] Schema documentation reflects new `language` and `dir` fields
- [ ] Code comments are accurate and complete on all new public types
- [ ] Error messages follow existing azd patterns (`ErrorWithSuggestion`)
- [ ] Example azure.yaml configs are valid and demonstrative

### References

- Contributes to: #7435

wbreza:
Resolved by PR #7451 (https://github.com/Azure/azure-dev/pull/7451). Updated \zure.yaml.json\ schemas (v1.0 and alpha) with \kind\ and \dir\ fields. Added \docs/language-hooks.md\ user guide. Error messages use \ErrorWithSuggestion\ with install URLs.

[...]

Issue #7443, opened 2026-04-02
Runtime validation in HooksManager


wbreza:
## Runtime Validation in HooksManager

**Part of #7435 — Multi-Language Hook Support**

### User Story

As a template consumer, I want to get a clear error message immediately if my system is missing Python (or another required runtime) rather than waiting until hook execution fails deep in the deployment.

### Solution Approach

Modify `pkg/ext/hooks_manager.go`:

1. Add `ValidateLanguageRuntimes(ctx context.Context, hooks map[string][]*HookConfig) []error`:
   - Iterate all hook configs.
   - For each language hook, check runtime availability via executor's `Prepare()` dry-run or a lightweight `CheckInstalled()`.
   - Cache results per language within a single validation pass (don't check Python N times).
   - Collect all errors — report all missing runtimes at once.
   - Return empty slice when all runtimes available.
2. This is **soft validation** — informational at startup (e.g., during `azd init` or start of `azd provision`). The hard error remains in `Prepare()` at execution time.
3. Surface via `ErrorWithSuggestion` with install URLs from `ExternalTool.InstallUrl()`.
4. Must handle both project-level and service-level hooks (learned from #5453).

Modify `pkg/ext/hooks_manager_test.go`:
- Python available → no errors
- Python missing → error with install suggestion
- Multiple languages missing → all errors collected
- Duplicate language hooks → check cached (only one Python check)

**Files:** `pkg/ext/hooks_manager.go` (MODIFY), `pkg/ext/hooks_manager_test.go` (MODIFY)

### Acceptance Criteria

- [ ] `ValidateLanguageRuntimes` returns errors for each missing runtime
- [ ] Caches per language — no duplicate checks
- [ ] Returns empty slice when all runtimes available
- [ ] Does not block hook resolution (informational only)
- [ ] Error messages use `ErrorWithSuggestion` with platform-specific install URLs

### References

- Refs: Extends #4740, #4560
- Contributes to: #7435

wbreza:
Resolved by PR #7451 (https://github.com/Azure/azure-dev/pull/7451). Runtime validation is handled during the executor's \Prepare()\ phase rather than a separate pre-flight check. Missing Python runtime produces clear errors with install URL suggestions via \ErrorWithSuggestion\. Functionally equivalent to the specified approach.

wbreza:
Completed via PR #7451

PR #7451, opened 2026-04-02
feat: add multi-language hook support (Phase 1 — Python)

wbreza:
## Summary

Adds multi-language hook support to azd, starting with Python as the first non-shell language. This extends the existing hook system (previously bash/PowerShell only) with a unified `HookExecutor` architecture that supports auto-detection, virtual environment management, and automatic dependency installation.

### Related Issues
- Epic: #7435
- Closes: #7436, #7437, #7438, #7439, #7440, #7441, #7442, #7443, #7444, #7445
- Related: #4384, #7423, #3613

---

### Quick Start

Add a Python script as a hook — azd automatically detects the `.py` extension, creates a virtual environment, installs dependencies, and runs the script:

```yaml
# azure.yaml
hooks:
  preprovision:
    run: ./hooks/setup.py
```

Place a `requirements.txt` alongside your script:
```
# hooks/requirements.txt
requests>=2.28
colorama>=0.4
```

That's it — `azd provision` will automatically:
1. Detect Python from the `.py` extension
2. Create a virtual environment (`hooks_env/`)
3. Install dependencies from `requirements.txt`
4. Run `setup.py` with azd environment variables available

### Configuration Options

```yaml
hooks:
  <hookName>:
    run: <path-to-script.py>    # Path to Python script (required)
    kind: python                 # Optional — auto-detected from .py extension
    dir: <working-directory>     # Optional — auto-inferred from script path
```

| Field | Description | Default |
|-------|-------------|---------|
| `run` | Path to `.py` script file (relative to project root) | Required |
| `kind` | Executor type: `python`, `sh`, `pwsh` | Auto-detected from file extension |
| `dir` | Working directory for the hook | Directory containing the script |

### Examples

**Auto-detected (simplest):**
```yaml
hooks:
  postprovision:
    run: ./hooks/seed-database.py
```

**Script in subdirectory:**
```yaml
hooks:
  preprovision:
    run: hooks/preprovision/main.py
    # dir auto-inferred as hooks/preprovision/
    # venv created as hooks/preprovision_env/
```

**Override dir for shared dependencies:**
```yaml
hooks:
  postprovision:
    run: ./tools/scripts/seed.py
    dir: ./tools  # requirements.txt lives in ./tools/
```

**Platform-specific hooks:**
```yaml
hooks:
  postprovision:
    windows:
      run: ./hooks/setup.ps1
    posix:
      run: ./hooks/setup.py
```

### Virtual Environment Handling

azd automatically manages Python virtual environments:

1. **Detection** — checks for existing venvs in order:
   - `VIRTUAL_ENV` environment variable
   - `.venv/` directory (with `pyvenv.cfg`)
   - `venv/` directory (with `pyvenv.cfg`)
2. **Creation** — if none found, creates `{dirName}_env/` alongside the script
3. **Reuse** — venvs persist across hook runs for fast re-execution

### Dependency Installation

azd walks up the directory tree from the script location to find dependency files:

| File | Behavior |
|------|----------|
| `pyproject.toml` | Runs `pip install .` (PEP 621 project install) |
| `requirements.txt` | Runs `pip install -r requirements.txt` |
| Neither found | Runs script with system Python (no venv) |

`pyproject.toml` takes priority over `requirements.txt` if both exist.

### Requirements

- **Python 3.3+** with the `venv` module
- On Debian/Ubuntu: may need `sudo apt install python3-venv`
- Python must be on `PATH` (`python3` on Unix, `py`/`python` on Windows)

### Limitations

- **Inline scripts not supported** — Python hooks must reference a file (`run: ./script.py`), not inline code
- **Phase 1** — Python only. JavaScript, TypeScript, and .NET hooks planned for future phases
- **pip only** — Uses pip for dependency management (poetry/pipenv/conda not yet supported)

---

## Architecture Changes

- **`HookExecutor` interface** (`pkg/tools/script.go`): Unified 3-phase lifecycle — `Prepare()` / `Execute()` / `Cleanup()`
- **IoC-based resolution**: Executors registered as named transients, resolved via `serviceLocator.ResolveNamed()`
- **`HookKind` type** (`pkg/tools/language/executor.go`): Type-safe executor selection (`sh`, `pwsh`, `python`)
- **Path resolution**: Done once in `validate()` with containment checks against project boundary
- **Shared `CreateInlineTempScript()`**: Centralized temp file creation for bash/powershell inline scripts

### Files Added
- `pkg/tools/language/executor.go` — HookKind type and constants
- `pkg/tools/language/python_executor.go` — Python executor with venv/pip lifecycle
- `pkg/tools/language/project_discovery.go` — Walk-up project file finder
- `pkg/ext/python_hooks_e2e_test.go` — E2E tests
- `docs/language-hooks.md` — Feature documentation

### Files Modified
- `pkg/tools/script.go` — HookExecutor interface, ExecutionContext, CreateInlineTempScript
- `pkg/ext/models.go` — HookConfig with Kind/Dir fields, path resolution, containment checks
- `pkg/ext/hooks_runner.go` — Unified execHook() with IoC resolution
- `pkg/tools/bash/bash.go` / `powershell.go` — Executor implementations using shared helper
- `cmd/container.go` — IoC registration for hook executors
- `schemas/v1.0/azure.yaml.json` / `schemas/alpha/azure.yaml.json` — Schema updates

### DIFF
--- a/cli/azd/pkg/ext/hooks_runner.go
+++ b/cli/azd/pkg/ext/hooks_runner.go
@@ -212,7 +212,11 @@ func (h *HooksRunner) execHook(
 	// script temp files). Deferred before Prepare so cleanup runs
 	// even if Prepare fails partway through. Cleanup is safe to
 	// call when Prepare was not called or created no resources.
-	defer executor.Cleanup(ctx)
+	defer func() {
+		if cErr := executor.Cleanup(ctx); cErr != nil {
+			log.Printf("warning: cleanup failed for hook '%s': %v\n", hookConfig.Name, cErr)
+		}
+	}()
 
 	// Prepare (unified — venv/deps for Python, pwsh detection for
 	// PowerShell, inline temp file creation for Bash/PowerShell hooks).
--- a/cli/azd/.vscode/cspell.yaml
+++ b/cli/azd/.vscode/cspell.yaml
@@ -371,6 +371,9 @@ overrides:
   - filename: pkg/infra/provisioning/bicep/local_preflight.go
     words:
       - actioned
+  - filename: pkg/project/project.go
+    words:
+      - copylocks
   - filename: pkg/ext/models.go
     words:
       - unvalidated


cli/azd/cover
wbreza, line 1: Remove

cli/azd/pkg/ext/hooks_runner.go
```diff
@@ -166,61 +148,201 @@ func (h *HooksRunner) execHook(ctx context.Context, hookConfig *HookConfig, opti
...
+
+	// Determine working directory from Dir (set explicitly or
+	// auto-inferred from the run path by validate).
+	cwd := h.cwd
+	if hookConfig.Dir != "" {
+		dir := hookConfig.Dir
+		if !filepath.IsAbs(dir) {
+			dir = filepath.Join(boundaryDir, dir)
+		}
+		cwd = dir
```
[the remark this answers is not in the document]
wbreza: Fixed — added path containment validation using osutil.IsPathContained. The resolved cwd is now verified to be within oundaryDir before proceeding.

cli/azd/pkg/ext/hooks_runner.go
```diff
@@ -166,61 +148,201 @@ func (h *HooksRunner) execHook(ctx context.Context, hookConfig *HookConfig, opti
...
+			Message: fmt.Sprintf(
+				"Failed to prepare %s hook '%s'.",
+				hookConfig.Language,
+				hookConfig.Name,
+			),
+			Suggestion: fmt.Sprintf(
+				"Ensure the required runtime for '%s' is installed.",
+				hookConfig.Language,
+			),
+		}
```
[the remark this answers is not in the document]
wbreza: Fixed — removed the ErrorWithSuggestion wrapping so raw errors propagate through the rror_suggestions.yaml pipeline. Python-specific rules (venv, pip, runtime) will now match correctly.

cli/azd/pkg/ext/models.go
```diff
@@ -106,6 +151,52 @@ func (hc *HookConfig) validate() error {
 		hc.script = hc.Run
 	}
```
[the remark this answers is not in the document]
wbreza: This is no longer applicable — the inline script handling was refactored in a subsequent commit. Inline temp-file creation now happens in the executor's Prepare() phase, not in alidate(). The ScriptLocation enum was removed entirely.

cli/azd/pkg/tools/language/python_executor.go
```diff
@@ -0,0 +1,283 @@
...
+// already exist. If the venv directory exists, creation is
+// skipped. Non-existence errors (e.g. permission denied) are
+// propagated immediately.
+func (e *pythonExecutor) ensureVenv(
+	ctx context.Context,
+	projectDir, venvName, venvPath string,
+	envVars []string,
+) error {
+	_, statErr := os.Stat(venvPath)
+	if statErr == nil {
```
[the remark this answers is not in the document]
wbreza: Fixed — nsureVenv now verifies the path is an actual directory via info.IsDir() before skipping creation. A clear error is returned if a file or broken symlink exists at the venv path.

cli/azd/pkg/ext/hooks_manager.go
```diff
@@ -246,5 +249,144 @@ func (h *HooksManager) ValidateHooks(ctx context.Context, allHooks map[string][]
...
+// if none is found. Used to pull install URLs out of formatted
+// suggestion strings.
+func extractURL(s string) string {
+	for part := range strings.FieldsSeq(s) {
+		if strings.HasPrefix(part, "https://") ||
+			strings.HasPrefix(part, "http://") {
+			return strings.TrimRight(part, ")")
+		}
+	}
+	return s
```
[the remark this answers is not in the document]
wbreza: Fixed — removed xtractURL() entirely. The HookWarning struct now carries a URL field populated directly from pythonCli.InstallUrl(), avoiding any ANSI escape parsing.

cli/azd/cover
```diff
@@ -0,0 +1,32 @@
...
+github.com/azure/azure-dev/cli/azd/pkg/tools/language/executor.go:186.27,193.32 3 1
+github.com/azure/azure-dev/cli/azd/pkg/tools/language/executor.go:193.32,197.3 1 1
+github.com/azure/azure-dev/cli/azd/pkg/tools/language/executor.go:199.2,199.27 1 1
+github.com/azure/azure-dev/cli/azd/pkg/tools/language/executor.go:199.27,201.3 1 0
+github.com/azure/azure-dev/cli/azd/pkg/tools/language/executor.go:203.2,203.42 1 1
+github.com/azure/azure-dev/cli/azd/pkg/tools/language/executor.go:212.10,213.31 1 1
+github.com/azure/azure-dev/cli/azd/pkg/tools/language/executor.go:213.31,214.44 1 1
+github.com/azure/azure-dev/cli/azd/pkg/tools/language/executor.go:214.44,216.4 1 1
+github.com/azure/azure-dev/cli/azd/pkg/tools/language/executor.go:218.3,218.18 1 0
+github.com/azure/azure-dev/cli/azd/pkg/tools/language/executor.go:221.2,221.18 1 0
```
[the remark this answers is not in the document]
wbreza: Fixed — removed the file and added cover to .gitignore.

cli/azd/pkg/ext/hooks_runner_test.go
```diff
@@ -5,21 +5,45 @@ package ext
...
+		string(language.ScriptLanguagePowerShell), powershell.NewExecutor,
+	)
+
+	// Register python.Cli (needed by NewPythonExecutor IoC constructor).
+	mockCtx.Container.MustRegisterSingleton(python.NewCli)
+
+	mockCtx.Container.MustRegisterNamedTransient(
+		string(language.ScriptLanguagePython), language.NewPythonExecutor,
+	)
+}
```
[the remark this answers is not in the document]
wbreza: Fixed — extracted RegisterHookExecutors to a shared 	est/mocks/mocktools/ package. All 3 test files now delegate to the shared helper.

cli/azd/cmd/hooks_test.go
```diff
@@ -16,14 +16,37 @@ import (
...
+	mockCtx.Container.MustRegisterNamedTransient(
+		string(language.ScriptLanguagePowerShell),
+		powershell.NewExecutor,
+	)
+	mockCtx.Container.MustRegisterSingleton(python.NewCli)
+	mockCtx.Container.MustRegisterNamedTransient(
+		string(language.ScriptLanguagePython),
+		language.NewPythonExecutor,
+	)
+}
```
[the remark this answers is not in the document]
wbreza: Fixed — see above. All 3 duplications consolidated into 	est/mocks/mocktools/mock_executors.go.

cli/azd/cmd/middleware/hooks_test.go
```diff
@@ -890,3 +910,19 @@ func Test_PowerShellWarning_WithNoPowerShellInstalled(t *testing.T) {
...
+		string(language.ScriptLanguageBash), bash.NewExecutor,
+	)
+	mockCtx.Container.MustRegisterNamedTransient(
+		string(language.ScriptLanguagePowerShell), powershell.NewExecutor,
+	)
+	mockCtx.Container.MustRegisterSingleton(python.NewCli)
+	mockCtx.Container.MustRegisterNamedTransient(
+		string(language.ScriptLanguagePython), language.NewPythonExecutor,
+	)
+}
```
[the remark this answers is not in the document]
wbreza: Fixed — see above.

cli/azd/pkg/tools/language/executor.go
```diff
@@ -0,0 +1,68 @@
...
+	ScriptLanguageJavaScript ScriptLanguage = "js"
+	// ScriptLanguageTypeScript identifies TypeScript scripts (.ts files).
+	// Not yet supported — IoC resolution will fail with a descriptive error.
+	ScriptLanguageTypeScript ScriptLanguage = "ts"
+	// ScriptLanguagePython identifies Python scripts (.py files).
+	ScriptLanguagePython ScriptLanguage = "python"
+	// ScriptLanguageDotNet identifies .NET (C#) scripts (.cs files).
+	// Not yet supported — IoC resolution will fail with a descriptive error.
+	ScriptLanguageDotNet ScriptLanguage = "dotnet"
+)
```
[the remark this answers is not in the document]
wbreza: Acknowledged — the PR description references GetExecutor() and ErrUnsupportedLanguage which were removed during the IoC refactoring. The current implementation uses serviceLocator.ResolveNamed() for executor resolution. Will update the PR description.

cli/azd/pkg/ext/hooks_runner.go
```diff
@@ -166,61 +149,205 @@ func (h *HooksRunner) execHook(ctx context.Context, hookConfig *HookConfig, opti
...
+	// Determine working directory from Dir (set explicitly or
+	// auto-inferred from the run path by validate).
+	cwd := h.cwd
+	if hookConfig.Dir != "" {
+		dir := hookConfig.Dir
+		if !filepath.IsAbs(dir) {
+			dir = filepath.Join(boundaryDir, dir)
+		}
+		cwd = dir
+	} else if hookConfig.path != "" {
```
jongio: **[HIGH]** File-based shell hooks now run with cwd set to the script's directory instead of the project root.

The `else if hookConfig.path != ""` branch applies to all hooks, not just language hooks. A user with `run: hooks/deploy.sh` previously got cwd = project root; now they get cwd = `hooks/`. This is a silent breaking change - existing hooks with relative paths like `./src/something` will break. Guard this branch with `!hookConfig.Language.IsShellLanguage()` to preserve old behavior for shell hooks.

cli/azd/pkg/tools/language/python_executor.go
```diff
@@ -0,0 +1,295 @@
...
+		scriptPath, execCtx.BoundaryDir,
+	)
+	if err != nil {
+		return fmt.Errorf(
+			"discovering project file: %w", err,
+		)
+	}
+
+	// No project file — run with system Python directly.
+	if projCtx == nil {
```
jongio, line 143: **[HIGH]** `DiscoverProjectFile` returns project contexts for any language, but the Python executor doesn't filter by language.

If a Python hook lives near a `package.json` (no Python project files), `DiscoverProjectFile` returns a JS context. The executor then creates a venv, calls `installDeps` which silently falls through the switch (no case for `package.json`), and sets `venvPath` to an empty venv. Fix: skip venv setup when `projCtx.Language != ScriptLanguagePython`.

cli/azd/pkg/tools/powershell/powershell.go
```diff
@@ -6,85 +6,139 @@ package powershell
...
-		runArgs = runArgs.WithStdOut(options.StdOut)
+	// On Windows, fall back to powershell (PS5).
+	if runtime.GOOS == "windows" {
+		if p.commandRunner.ToolInPath("powershell") == nil {
+			p.shellCmd = "powershell"
+			return nil
+		}
+		return &internal.ErrorWithSuggestion{
+			Err: fmt.Errorf("neither pwsh nor powershell found in PATH"),
+			Suggestion: fmt.Sprintf(
```
jongio, line 82: **[MEDIUM]** The PS5 fallback lost its execution-failure diagnostic.

Old code: if `pwsh` wasn't found and `powershell` (PS5) was used instead, a script failure included "pwsh was not found and powershell was automatically used instead." New code silently falls back. Track `usingFallback` state and wrap Execute errors with the PS5 suggestion when the fallback was used.

cli/azd/pkg/ext/hooks_runner.go
```diff
@@ -166,61 +149,205 @@ func (h *HooksRunner) execHook(ctx context.Context, hookConfig *HookConfig, opti
...
+	if options.StdOut != nil {
+		execCtx.StdOut = options.StdOut
+	}
+
+	// Resolve executor via IoC — hooks runner has NO knowledge of executor internals.
+	var executor tools.HookExecutor
+	if err := h.serviceLocator.ResolveNamed(string(hookConfig.Language), &executor); err != nil {
+		return &errorhandler.ErrorWithSuggestion{
+			Err: fmt.Errorf(
+				"no executor for language '%s': %w",
```
jongio: **[MEDIUM]** `ErrorWithSuggestion` is missing the `Links` field.

Per codebase convention, ErrorWithSuggestion should populate all fields since ErrorMiddleware skips the YAML pipeline for typed errors. Add a link to the hooks docs so users can discover supported languages.

cli/azd/cmd/middleware/hooks_test.go
```diff
@@ -73,7 +76,7 @@ func Test_CommandHooks_Middleware_ValidProjectWithDifferentCommand(t *testing.T)
 			"precommand": {
 				{
 					Run:   "echo 'hello'",
-					Shell: ext.ShellTypeBash,
+					Shell: "sh",
```
jongio: Why switch to hardcode?

cli/azd/cmd/middleware/hooks_test.go
```diff
@@ -503,7 +513,7 @@ func Test_PowerShellWarning_WithPowerShellHooks(t *testing.T) {
 			"preprovision": {
 				{
 					Run:   "Write-Host 'hello'",
-					Shell: ext.ShellTypePowershell,
+					Shell: "pwsh",
```
jongio: Same hardcode

cli/azd/docs/language-hooks.md
```diff
@@ -0,0 +1,171 @@
...
+project or service level. Two optional fields are available:
+
+### `language` (string, optional)
+
+Specifies the executor type for the hook. Allowed values:
+`sh`, `pwsh`, `js`, `ts`, `python`, `dotnet`.
+
+When omitted, the executor is **auto-detected** from the file extension of the
+`run` path. For example, `run: ./hooks/seed.py` automatically selects the
+Python executor.
```
jongio, line 30: This is nice!

cli/azd/docs/language-hooks.md
```diff
@@ -0,0 +1,171 @@
...
+
+### `dir` (string, optional)
```
jongio: Should this be 'dir' or 'cd'?  I don't recall what azd uses in other places, but make sure consistent

cli/azd/docs/language-hooks.md
```diff
@@ -0,0 +1,171 @@
...
+When the script's project root differs from the script's directory (e.g. the
+entry point is in a `src/` subdirectory but dependencies are at the project
+level), use `dir` to override the auto-inferred value:
+
+```yaml
+hooks:
+  postprovision:
+    run: ./hooks/data-tool/src/main.py
+    language: python
+    dir: ./hooks/data-tool    # override: project root differs from script location
```
jongio: In example use a diff dir than what is in the run attribute

cli/azd/docs/language-hooks.md
```diff
@@ -0,0 +1,171 @@
...
+Hooks support the `secrets` field for resolving Azure Key Vault references,
+regardless of executor type:
+
+```yaml
+hooks:
+  postprovision:
+    run: ./hooks/seed-database.py
+    secrets:
+      DB_CONNECTION_STRING: DATABASE_URL
+```
```
jongio, line 124: These are auto resolved now

cli/azd/docs/language-hooks.md
```diff
@@ -0,0 +1,171 @@
...
+   during Prepare (e.g. inline script temp files). This runs regardless
+   of whether Execute succeeded or failed.
+
+## Limitations
+
+- **Inline scripts** are only supported for Bash and PowerShell hooks.
+  All other executor types must reference a file path.
+- **Phase 1** supports only Python as a non-shell executor. JavaScript,
+  TypeScript, and .NET support is planned for future phases.
+- **Virtual environments** are created in the project directory alongside the
```
jongio, line 169: We should flesh out the venv config for python scripts with this PR.  We need to detect if the script already lives in a venv and if so use it.  And allow them to override it.

cli/azd/pkg/ext/hooks_runner.go
wbreza: Fixed — added `!hookConfig.Kind.IsShell()` guard so only non-shell executors derive cwd from the script path. Shell hooks retain the original project-root behavior.

cli/azd/pkg/tools/language/python_executor.go
wbreza: Fixed — added `projCtx.Language != HookKindPython` check after project discovery. Non-Python project contexts (e.g., package.json) are now skipped, preventing empty venv creation.

cli/azd/pkg/ext/hooks_runner.go
wbreza: Fixed — added `Links` field with a link to the hooks documentation.

cli/azd/pkg/tools/powershell/powershell.go
wbreza: Fixed — added `usingFallback` field to the PowerShell executor. When PS5 is used as a fallback and the script fails, the error is wrapped with the diagnostic message about pwsh not being found.

cli/azd/cmd/middleware/hooks_test.go
wbreza: Fixed — same as above, all test files updated to use typed constants.

cli/azd/cmd/middleware/hooks_test.go
wbreza: Fixed — restored `language.HookKindBash` and `language.HookKindPowerShell` constants in all test files. String literals were an interim artifact of removing the ShellType enum.

cli/azd/docs/language-hooks.md
wbreza: Fixed — updated the example to use `run: ./tools/scripts/seed.py` with `dir: ./tools` to clearly show the directory override scenario.

cli/azd/docs/language-hooks.md
wbreza: Updated — the docs now lead with auto-inference and clarify that `dir` (now `kind`) is only needed as an override.

cli/azd/docs/language-hooks.md
wbreza: Thanks Jon! 🙏

cli/azd/docs/language-hooks.md
wbreza: Implemented — the Python executor now detects existing virtual environments via the `VIRTUAL_ENV` environment variable and by scanning for `.venv` and `venv` directories (validated by `pyvenv.cfg`). External venvs skip dependency installation; project-local venvs still run pip install.

cli/azd/docs/language-hooks.md
wbreza: Good catch - updated to align with azd's `cwd` convention. The heading now reads `dir (string, optional) - working directory` with `(cwd)` parenthetical in the description.

wbreza:
@weikanglim  

**On narrowing to script-only (no venv/pip):** Python hooks without dependency management covers stdlib-only scripts — a very limited scenario. The primary signal from #4384 is template consumers running `azd up` and having everything work automatically. If a hook needs `requests` or `azure-identity`, the consumer shouldn't need to know that, let alone manually run `pip install`. Dependency management is the core value proposition, not an optional enhancement.

**azd should own toolchain orchestration.** Anyone consuming an azd project should never need to manually configure language toolchains before running azd commands. The only prerequisite a user needs is the runtime itself (Python, pip) — which azd validates early with actionable install guidance. Everything else — virtual environments, dependency installation, project discovery — is azd's responsibility. This is consistent with how azd already handles infrastructure (Bicep/Terraform), containerization (Docker), and service deployment. Hooks should follow the same principle.

**On `env exec` (#7423) as an alternative:** `env exec` solves ad-hoc, user-initiated execution — "run this command with my env loaded." Hooks solve a different problem: lifecycle-bound, declarative, template-author-defined automation that runs transparently for consumers. A template author defining `preprovision: run: ./hooks/seed.py` needs azd to handle the full execution lifecycle without consumer intervention. `env exec` doesn't address that. Both features are complementary — neither replaces the other.

**On dependency management being opinionated:** The pip+venv choice matches what azd's Python framework services already use (`framework_service_python.go`). As azd adds support for poetry, conda, uv, etc., hook execution inherits those capabilities through the same `python.Cli` interface — the `HookExecutor` architecture delegates to the same tool wrappers the rest of azd uses, not a separate toolchain.

**On JS/TS/.NET scaffolding:** Agreed — those are deferred. Phase 1 is Python only. Framework types exist but return errors for unregistered languages.

The issue history (#4384, #3978, #7423) consistently points toward azd handling toolchain complexity so template consumers don't have to.

schemas/v1.0/azure.yaml.json
```diff
@@ -787,6 +787,24 @@
                     ],
                     "default": "sh"
                 },
+                "language": {
```
jongio: Schema property `language` does not match the Go struct field name.

`HookConfig.Kind` uses yaml tag `kind`, not `language`. Users writing `language: python` per this schema will have it silently ignored by Go YAML unmarshaler - the hook falls back to extension detection or fails. Either rename this to `kind`, or add a `Language string` field to HookConfig with `yaml:"language"` that maps to Kind in validate(). Same issue in `schemas/alpha/azure.yaml.json`.

cli/azd/pkg/ext/models.go
```diff
@@ -91,68 +139,204 @@ func (hc *HookConfig) validate() error {
...
+			"resolving hook directory: %w", absErr,
+		)
+	}
+	absBoundary, absErr := filepath.Abs(boundaryDir)
+	if absErr != nil {
+		return fmt.Errorf(
+			"resolving boundary directory: %w", absErr,
+		)
+	}
+	if !osutil.IsPathContained(absBoundary, absDir) {
```
jongio, line 316: `resolvedScriptPath` is not covered by this containment check.

This validates `resolvedDir` stays within `boundaryDir`, but `resolvedScriptPath` can still escape. For example: `dir: ./hooks` (within bounds) + `run: ../../../outside.py` resolves the script path outside the project root while Dir passes the check. Adding a parallel containment check on `resolvedScriptPath` would close the gap.

cli/azd/pkg/ext/hooks_runner.go
```diff
@@ -166,61 +147,177 @@ func (h *HooksRunner) execHook(ctx context.Context, hookConfig *HookConfig, opti
...
+		hookConfig.Name, hookConfig.Kind,
+	)
+
+	if err := executor.Prepare(ctx, scriptPath, execCtx); err != nil {
+		return fmt.Errorf("preparing hook '%s': %w", hookConfig.Name, err)
+	}
+
+	// Cleanup temp resources created during Prepare (e.g. inline
+	// script temp files). Runs after Execute regardless of outcome.
+	defer executor.Cleanup(ctx)
```
jongio: This defer should go before the Prepare call, not after.

The HookExecutor interface says Cleanup must be safe to call even when Prepare was not called or created no resources. With the defer after the error check, Cleanup will not run if Prepare fails partway through. Moving it before the Prepare call matches the contract and is more defensive for future executors.

cli/azd/pkg/tools/bash/bash.go
```diff
@@ -5,31 +5,81 @@ package bash
...
+type bashExecutor struct {
 	commandRunner exec.CommandRunner
-	cwd           string
-	envVars       []string
+	tempFile      string // temp script created from inline content
+}
+
+// Prepare creates a temp script file when the execution context
+// carries inline script content. For file-based hooks this is a no-op.
+func (b *bashExecutor) Prepare(
```
jongio, line 28: Temp file creation is duplicated with powershell Prepare.

Both bash and powershell executors have ~25 lines of identical structure: CreateTemp, WriteFile, Chmod, error cleanup paths. Extracting a shared `createInlineTempScript(hookName, ext, content string) (string, error)` in the `tools` package would centralize the permission logic and reduce drift when adding future executors.

schemas/v1.0/azure.yaml.json
wbreza: Fixed — renamed `language` to `kind` in both `schemas/v1.0/azure.yaml.json` and `schemas/alpha/azure.yaml.json`. Updated property name, title, description, and conditional blocks. `shell` remains as the deprecated back-compat alias.

cli/azd/pkg/ext/models.go
wbreza: Fixed — added `osutil.IsPathContained()` check on `resolvedScriptPath` against `boundaryDir`, parallel to the existing `resolvedDir` containment check. The `dir: ./hooks` + `run: ../../../outside.py` escape vector is now caught.

cli/azd/pkg/ext/hooks_runner.go
wbreza: Fixed — moved `defer executor.Cleanup(ctx)` before the `Prepare()` call so cleanup runs even on partial Prepare failure. Matches the HookExecutor contract.

cli/azd/pkg/tools/bash/bash.go
wbreza: Fixed — extracted `CreateInlineTempScript(hookName, ext, content)` into `pkg/tools/script.go`. Both bash and powershell `Prepare()` methods now use the shared helper, centralizing the permission logic.

cli/azd/pkg/tools/language/python_executor.go
```diff
@@ -0,0 +1,416 @@
...
+}
+
+// installDeps installs Python dependencies from the given file
+// into the virtual environment identified by venvName.
+func (e *pythonExecutor) installDeps(
+	ctx context.Context,
+	projectDir, venvName, depFile string,
+	envVars []string,
+) error {
+	switch depFile {
```
jongio, line 228: Nit: the switch only handles `requirements.txt` and `pyproject.toml` - an unrecognized `depFile` value silently succeeds. Consider adding a `default` case that at least logs: `log.Printf("unsupported dependency file %q - skipping install", depFile)`. This makes the coupling between `knownProjectFiles` and this switch explicit if a future PR adds new Python project file types.

cli/azd/pkg/ext/hooks_runner.go
```diff
@@ -166,66 +147,179 @@ func (h *HooksRunner) execHook(ctx context.Context, hookConfig *HookConfig, opti
...
+		}
+	}
+
+	// Cleanup temp resources created during Prepare (e.g. inline
+	// script temp files). Deferred before Prepare so cleanup runs
+	// even if Prepare fails partway through. Cleanup is safe to
+	// call when Prepare was not called or created no resources.
+	defer executor.Cleanup(ctx)
+
+	// Prepare (unified — venv/deps for Python, pwsh detection for
```
jongio, line 221: Minor: `defer executor.Cleanup(ctx)` drops the error. If a temp file can't be removed (e.g., file lock on Windows), there's no visibility. Consider:
```go
defer func() {
    if cErr := executor.Cleanup(ctx); cErr != nil {
        log.Printf("cleanup hook '%s': %v", hookConfig.Name, cErr)
    }
}()
```
This matches the codebase's pattern of logging non-critical errors rather than suppressing them entirely.

wbreza: Fixed — wrapped in a closure that logs cleanup errors via `log.Printf` instead of silently dropping them. Matches the existing logging pattern in `hooks_runner.go`.

cli/azd/pkg/tools/language/python_executor.go
wbreza: Valid point — deferring to a follow-up issue #7612 to add the default case with logging. Keeps the coupling between knownProjectFiles and the switch explicit for future project file types.

jongio (approved):
Clean architecture. IoC-based executor resolution is extensible - future languages just register a named transient. Path containment guards, error_suggestions entries, and the backward-compat path from `shell:` to `kind:` are all solid. Test coverage is thorough.

PR #7452, opened 2026-04-02
Add importer-provider extension capability for infrastructure generation

### LINKED ISSUE #7425: Feature: CDK Language Support — Author Azure Infrastructure in Any Programming Language
# Feature: CDK Language Support — Author Azure Infrastructure in Any Programming Language

## Summary

Enable developers to define their Azure infrastructure in general-purpose programming languages (C#, TypeScript/JavaScript, Python, Go, etc.) and have all `azd` commands work transparently. Customers should never need to know about or care about Bicep — they simply author their infrastructure in the CDK language of their choice, and `azd init`, `azd up`, `azd provision`, `azd provision --preview`, `azd deploy`, `azd down`, etc. all just work.

## Motivation

Today, `azd` supports two infrastructure-as-code options: **Bicep** and **Terraform**. While powerful, both are domain-specific languages that require developers to learn specialized syntax. Many developers would prefer to define their infrastructure using the same programming languages they already use for their application code.

Cloud Development Kits (CDKs) make this possible by letting developers use familiar languages and tooling — IDE IntelliSense, type checking, debugging, refactoring, testing, and package management — to define infrastructure. The CDK compiles to an intermediate representation (like Bicep or ARM templates) that cloud platforms understand.

**Azure already has this capability**: [Azure.Provisioning](https://learn.microsoft.com/dotnet/azure/sdk/provisioning/) lets C# developers define Azure resources using the same Azure SDK types they use in their application code. The [Azure CDK](https://github.com/Azure/azure-cdk) project is extending this to TypeScript/JavaScript. Other languages can follow the same pattern.

**The gap is in `azd`** — there's no way to point `azd` at a `.cs` or `.js` file in your `infra/` folder and have it just work. This proposal would close that gap.

### What it looks like today (Bicep only)

```
my-app/
├── src/           # Application code
├── infra/
│   ├── main.bicep              # Must learn Bicep DSL
│   ├── main.parameters.json
│   └── modules/
│       ├── cosmos.bicep
│       └── appservice.bicep
└── azure.yaml
```

### What it should look like (CDK — C# example)

```
my-app/
├── src/           # Application code
├── infra/
│   └── infra.cs   # Write infrastructure in C# — that's it
└── azure.yaml
```

Where `infra.cs` uses .NET 10 single-file app support (no `.csproj` needed):

```csharp
#:package Azure.Provisioning.Resources@1.0.0
#:package Azure.Provisioning.Storage@1.0.0

using Azure.Provisioning;
using Azure.Provisioning.Resources;
using Azure.Provisioning.Storage;

var infra = new Infrastructure("helloStorage");

var rg = new ResourceGroup("rg");
infra.Add(rg);

var storage = new StorageAccount("storage")
{
    Kind = StorageKind.StorageV2,
    Sku = new StorageSku { Name = StorageSkuName.StandardLrs },
};
rg.Add(storage);

infra.Build().Save(args[0]);
```

### What it should look like (CDK — TypeScript example)

```
my-app/
├── src/           # Application code
├── infra/
│   └── infra.ts   # Write infrastructure in TypeScript
└── azure.yaml
```

Where `infra.ts` uses the [Azure CDK](https://github.com/Azure/azure-cdk) TypeScript API:

```typescript
import { resourceGroup, stack } from "@azure-cdk/core";
import { storageAccount } from "@azure-cdk/storage";

const infra = stack({ name: "helloStorage", targetScope: "subscription" });

const rg = resourceGroup(infra, { location: "eastus" });

const storage = storageAccount(rg, {
  sku: { name: "Standard_LRS" },
});
```

### All `azd` commands should work

| Command | Behavior |
|---------|----------|
| `azd init` | Detect CDK files in `infra/`, scaffold `azure.yaml` accordingly |
| `azd up` | Compile CDK → intermediate → deploy to Azure |
| `azd provision` | Same as above, infra only |
| `azd provision --preview` | Compile CDK → show what-if diff |
| `azd deploy` | Deploy application code to provisioned resources |
| `azd down` | Tear down all resources |
| `azd provision -- --param value` | Forward parameters to the CDK program |

## Proof of Concept

We've built a working prototype for C# / Azure.Provisioning that demonstrates this is viable:

### PR: [#7352](https://github.com/Azure/azure-dev/pull/7352) — `provider: code` provisioning provider

- **`CodeProvider`** — language-agnostic router that detects the CDK language from file extensions in `/infra` and dispatches to language-specific sub-providers
- **`DotNetProvider`** — compiles C# to Bicep via `dotnet run`, then delegates to the existing `BicepProvider` for deployment
- **Auto-detection** — drop a `.cs` file in `/infra`, azd detects it automatically (no explicit `provider:` config needed in `azure.yaml`)
- **Single-file support** — uses .NET 10 file-based apps with `#:package` directives, no `.csproj` required
- **Parameter forwarding** — `azd provision -- --region westus3 --prefix myapp` passes args through to the C# program
- **Alpha-gated** — behind `azd config set alpha.code on`

### Real-World Validation

We converted the `todo-csharp-cosmos-sql` sample template from Bicep to a single `infra/infra.cs` file: [m-nash/todo-csharp-cosmos-sql-dotnet](https://github.com/m-nash/todo-csharp-cosmos-sql-dotnet)

- **7 Azure resources**: Cosmos DB (serverless, NoSQL), App Service Plan, Web App, API App, Key Vault, App Insights, RBAC
- **Full E2E verified**: `azd up` → API responds → Playwright tests pass → `azd down` cleans up
- **Same app code, same tests** — only the `/infra` folder changed

### Azure CDK Playground

The [Azure CDK](https://github.com/Azure/azure-cdk) project provides a web playground where users can write infrastructure in C# (and soon TypeScript) with full IntelliSense, see the compiled Bicep/JSON output in real-time, and get error diagnostics with line attribution — demonstrating the authoring experience.

## Architecture

The proposed architecture uses a **compilation step** that converts CDK code to an intermediate representation before deployment:

```
CDK Source (.cs, .ts, .js, .py)
    ↓  [language-specific compiler]
Intermediate Representation (Bicep / ARM JSON)
    ↓  [existing azd deployment pipeline]
Azure Resources
```

This approach:
- **Leverages existing azd infrastructure** — all deployment, state management, and resource tracking reuse the proven Bicep/ARM pipeline
- **Is language-extensible** — adding a new CDK language means implementing a compiler adapter, not a new deployment pipeline
- **Keeps Bicep as a supported first-class option** — this is additive, not a replacement

### 1. Core Provider vs Extension

There are two viable implementation approaches:

**Core Provider** (current PR #7352 approach):
- New `ProviderKind` in `provider.go` with `CodeProvider` routing to language sub-providers
- All `azd` commands work natively including `--preview`, parameter forwarding
- Requires changes to azd core; ships with azd releases

**azd Extension** (suggested by @vhvb1989):
- Use `ServiceTargetProvider` extension API to implement compilation + deployment
- Ships independently from azd releases; no core changes needed
- Extension handles `Package()` (compile CDK → Bicep) and `Deploy()` (deploy Bicep to Azure)

The extension approach may need some platform support to ensure all commands work seamlessly (e.g., `--preview` support, parameter forwarding). What's the preferred direction?

### 2. Language Onboarding

How should new CDK languages be added? The current architecture has a `CodeProvider` router that could dispatch based on file extension:
- `.cs` / `.csproj` → .NET / Azure.Provisioning
- `.ts` / `.js` → TypeScript/JavaScript / Azure CDK
- `.py` → Python (future)
- `.go` → Go (future)

Should each language be a separate azd extension, or should there be a single "CDK" extension/provider that bundles supported languages?

### 3. `azd init` Integration

When a user runs `azd init`, should there be an option to choose their CDK language? For example:

```
? Select your infrastructure language:
  > Bicep
    C# (Azure.Provisioning)
    TypeScript
    Terraform
```

### 4. Template Ecosystem

How should CDK-based templates be represented in the template gallery ([aka.ms/awesome-azd](https://aka.ms/awesome-azd))? Should there be CDK variants of existing templates, or should templates support multiple infrastructure languages side by side?

## Related Work

- **PR [#7352](https://github.com/Azure/azure-dev/pull/7352)** — Working prototype of `provider: code` for C#
- **[Azure.Provisioning](https://learn.microsoft.com/dotnet/azure/sdk/provisioning/)** — C# CDK for Azure resources
- **[Azure CDK Playground](https://github.com/Azure/azure-cdk)** — Web-based authoring environment with IntelliSense
- **[todo-csharp-cosmos-sql-dotnet](https://github.com/m-nash/todo-csharp-cosmos-sql-dotnet)** — Real-world sample template using C# infrastructure (private, can be shared)

vhvb1989:
## Summary

Adds a new `importer-provider` capability to the azd extension framework, enabling extensions to generate infrastructure (Bicep/Terraform) from project-specific definition formats.

This POC enables creating extensions for scenarios like [#7425](https://github.com/Azure/azure-dev/issues/7425), where projects want to define infrastructure in languages like C# or TypeScript instead of writing Bicep directly. The demo extension illustrates this pattern using markdown files as an analogy — where the `.md` files with resource definitions play the role that `.cs` or `.ts` files would in a real implementation. The extension reads these definitions and produces the Bicep that azd provisions.

## How it works

```yaml
# azure.yaml
name: my-app
infra:
  importer:
    name: demo-importer    # extension-provided importer
    # options:              # extension-owned settings (optional)
    #   path: custom-folder
services:
  app:
    host: staticwebapp
    language: js
    project: ./src/app
    dist: dist
```

- **`azd provision`**: If no `infra/` folder exists, the importer generates temporary Bicep at runtime
- **`azd infra gen`**: The importer generates Bicep files into `infra/` (ejection)
- **After ejection**: azd uses the ejected files directly, skipping the importer (supports customization)

## Key design decisions

- **Importers live under `infra.importer`**, not in the services list — services remain clean for deployable units only
- **`options` is an extension-owned map** — each extension defines its own settings (path, format, etc.) with its own defaults
- **Infra override**: Once `infra/` exists with generated files, azd uses those and skips the importer
- **Backward compatible**: Aspire auto-detection via `CanImport` on services is preserved for existing projects

### Core framework
- `Importer` interface with `CanImport`, `Services`, `ProjectInfrastructure`, `GenerateAllInfrastructure`
- `ImportManager` refactored from hardcoded `DotNetImporter` to pluggable `[]Importer` list
- `ImporterRegistry` singleton for extension-registered importers
- `ImporterConfig` with extension-owned `options` map in `provisioning.Options`

### gRPC layer
- `importer.proto` with bidirectional `Stream` RPC
- Extension SDK: `ImporterProvider` interface, `ImporterManager`, `ImporterEnvelope`
- Server-side `ImporterGrpcService` with capability verification
- `ExternalImporter` adapter implementing `Importer` over gRPC

### Demo extension
- `DemoImporterProvider` reads `.md` files with `azd-infra-gen/v1` front-matter header
- Parses resource definitions (resource group, static web app with `azd-service-name` tag)
- Generates `main.bicep`, `main.parameters.json`, `resources.bicep`
- Default folder: `demo-importer/` (overridable via `options.path`)

### Sample project (`test/functional/testdata/samples/extension-importer/`)
- Combines importer-generated infra with a deployable Static Web App service
- Works end-to-end with `azd up` to deploy to Azure

### Documentation
- `docs/extensions/extension-custom-importers.md` — authoring guide

## What is NOT included

- **`azd init` integration**: Extension importers are not invoked during `azd init`. The init command uses built-in `appdetect` with hardcoded language detectors. Extensions can work around this by adding their own init command (e.g., `azd my-importer init`), similar to `azd ai agent init`. A future `project-detector` capability could integrate extensions into `azd init` auto-detection, following the same strategy where extensions define what to detect and what to write.

Closes #7425

### DIFF
--- a/cli/azd/extensions/microsoft.azd.demo/internal/project/importer_demo.go
+++ b/cli/azd/extensions/microsoft.azd.demo/internal/project/importer_demo.go
@@ -238,8 +238,8 @@ func hasInfraGenHeader(path string) bool {
 		if line == "---" {
 			break // end of front-matter
 		}
-		if strings.HasPrefix(line, "format:") {
-			value := strings.TrimSpace(strings.TrimPrefix(line, "format:"))
+		if after, ok := strings.CutPrefix(line, "format:"); ok {
+			value := strings.TrimSpace(after)
 			return value == formatHeader
 		}
 	}
@@ -279,7 +279,7 @@ func parseResourceFile(path string) ([]resourceDef, error) {
 	inFrontMatter := false
 	parsingTags := false
 
-	for _, line := range strings.Split(string(data), "\n") {
+	for line := range strings.SplitSeq(string(data), "\n") {
 		trimmed := strings.TrimSpace(line)
 
 		// Skip front-matter
@@ -309,8 +309,8 @@ func parseResourceFile(path string) ([]resourceDef, error) {
 		}
 
 		// Parse "- key: value" properties
-		if strings.HasPrefix(trimmed, "- ") {
-			prop := strings.TrimPrefix(trimmed, "- ")
+		if after, ok := strings.CutPrefix(trimmed, "- "); ok {
+			prop := after
 
 			// Check for tag entries (indented under tags:)
 			if parsingTags {


cli/azd/pkg/project/dotnet_importer.go
```diff
@@ -573,7 +611,25 @@ func evaluateSingleBuildArg(
...
+func (ai *DotNetImporter) GenerateAllInfrastructure(
+	ctx context.Context,
+	projectPath string,
+	importerConfig provisioning.ImporterConfig,
+) (fs.FS, error) {
+	absPath, err := filepath.Abs(projectPath)
+	if err != nil {
+		absPath = projectPath
+	}
+	svcConfig := &ServiceConfig{
```
jongio, line 625: **[CRITICAL]** Nil pointer panic in `GenerateAllInfrastructure` bridge

`generateAllInfrastructureFromConfig` dereferences `p.Infra.Module`, `p.Infra.Path`, and `p.Path` - all panic when `p` is nil. This path is hit whenever `ImportManager.GenerateAllInfrastructure` auto-detects an Aspire project (the interface dispatches here through the bridge).

Simple nil guards aren't sufficient either - the inner method uses `filepath.Rel(p.Path, ...)` to compute service manifest locations. Consider constructing a minimal `ProjectConfig{Path: projectPath, Infra: DefaultProvisioningOptions}` instead of passing nil.

cli/azd/pkg/project/importer_external.go
```diff
@@ -0,0 +1,281 @@
...
+	}
+
+	// Write generated files to a temp directory
+	tmpDir, err := os.MkdirTemp("", "azd-ext-infra")
+	if err != nil {
+		return nil, fmt.Errorf("creating temporary directory: %w", err)
+	}
+
+	for _, file := range infraResp.Files {
+		target := filepath.Join(tmpDir, file.Path)
```
jongio, line 158: **[CRITICAL]** Path traversal - extension-controlled file paths written without containment check

`file.Path` comes from the extension over gRPC and is joined directly into `tmpDir` without validation. A path like `../../.ssh/authorized_keys` would escape the temp directory. Extensions are user-installed, but defense-in-depth still applies.

Validate that `filepath.Clean(target)` stays under `tmpDir` before writing.

cli/azd/internal/grpcserver/importer_service.go
```diff
@@ -0,0 +1,129 @@
...
+	}
+
+	// Run the broker dispatcher (blocking)
+	// This will return when the stream closes or encounters an error
+	if err := broker.Run(ctx); err != nil && !errors.Is(err, context.Canceled) {
+		return fmt.Errorf("broker error: %w", err)
+	}
+
+	s.providerMapMu.Lock()
+	delete(s.providerMap, registeredImporterName)
```
jongio, line 89: **[HIGH]** Stale importer left in registry after stream close

Stream teardown removes the importer from `providerMap` but not from `ImporterRegistry`. The stale `ExternalImporter` stays in the registry with a dead broker - any subsequent call to it fails with a gRPC stream error instead of a clean "importer not available" message.

`ImporterRegistry` needs a `Remove(name)` method called here alongside the `providerMap` delete.

cli/azd/pkg/project/importer.go
```diff
@@ -264,20 +386,32 @@ func (im *ImportManager) validateServiceDependencies(services []*ServiceConfig,
...
-				log.Printf("error checking if %s is an app host project: %v", svcConfig.Path(), err)
 			}
 		}
 	}
 	return false
 }
 
+// HasAppHost returns true when there is one AppHost (Aspire) in the project.
+// Deprecated: Use HasImporter instead.
+func (im *ImportManager) HasAppHost(ctx context.Context, projectConfig *ProjectConfig) bool {
```
jongio, line 411: **[HIGH]** `HasAppHost` behavioral change - now triggers for non-Aspire importers

`HasAppHost` now delegates to `HasImporter`, which iterates all importers including extension-provided ones. Callers like `pipeline_manager.go` use this to decide Aspire-specific CI behavior (`InstallDotNetForAspire: props.HasAppHost`). A non-Aspire extension importer returning `CanImport=true` would incorrectly trigger .NET SDK installation in generated CI pipelines.

Either keep `HasAppHost` checking only the Aspire importer by name, or rename callers to distinguish Aspire-specific behavior from generic importer presence.

cli/azd/pkg/project/importer_external.go
```diff
@@ -0,0 +1,281 @@
...
+	// Send the fully resolved path so extensions can access project files
+	// regardless of their own working directory
+	if svc.Project != nil {
+		proto.RelativePath = svc.Path()
+	} else {
+		proto.RelativePath = svc.RelativePath
+	}
+
+	if !svc.ResourceName.Empty() {
+		noopMapping := func(string) string { return "" }
```
jongio, line 245: **[MEDIUM]** Noop env var resolver mangles resource names

The noop mapper `func(string) string { return "" }` replaces env var references with empty string - so `rg-${AZURE_ENV_NAME}` becomes `rg-`. This silently corrupts resource names sent to extensions. Use an identity mapper that preserves the template syntax, or send the raw unexpanded value.

cli/azd/test/functional/extension_importer_test.go
```diff
@@ -0,0 +1,113 @@
...
+
+	cli := azdcli.NewCLI(t, azdcli.WithSession(session))
+	cli.WorkingDirectory = dir
+	cli.Env = append(cli.Env, os.Environ()...)
+	cli.Env = append(cli.Env, "AZURE_LOCATION=eastus2")
+
+	_, err = cli.RunCommandWithStdIn(ctx, stdinForInit(envName), "init")
+	require.NoError(t, err)
+
+	// Run `azd infra synth` which invokes ImportManager.GenerateAllInfrastructure
```
jongio, line 102: **[LOW]** Test description doesn't match what it does

Comments reference "azd infra synth" but the test runs `azd show` (line 110). This is a smoke test for extension startup, not infra generation. Either update the comments to match, or add an actual `azd infra synth` call to exercise the importer pipeline end-to-end.

PR #7464, opened 2026-04-02
[Agents extension] Prompt user for protocols during "from code" flow

trangevi:
When using local code, prompt the user as to what kind of protocol they want to use.

### DIFF
--- a/cli/azd/extensions/azure.ai.agents/internal/cmd/init_from_code.go
+++ b/cli/azd/extensions/azure.ai.agents/internal/cmd/init_from_code.go
@@ -443,7 +443,7 @@ func (a *InitFromCodeAction) createDefinitionFromLocalAgent(ctx context.Context)
 	agentKind := agent_yaml.AgentKindHosted
 
 	// Prompt user for supported protocols
-	protocols, err := promptProtocols(ctx, a.azdClient, a.flags.NoPrompt, a.flags.protocols)
+	protocols, err := promptProtocols(ctx, a.azdClient.Prompt(), a.flags.NoPrompt, a.flags.protocols)
 	if err != nil {
 		return nil, err
 	}
@@ -806,7 +806,7 @@ var knownProtocols = []protocolInfo{
 // When noPrompt is true and no flag values are provided, defaults to [responses/v1].
 func promptProtocols(
 	ctx context.Context,
-	azdClient *azdext.AzdClient,
+	promptClient azdext.PromptServiceClient,
 	noPrompt bool,
 	flagProtocols []string,
 ) ([]agent_yaml.ProtocolVersionRecord, error) {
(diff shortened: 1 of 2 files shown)


cli/azd/extensions/azure.ai.agents/internal/cmd/init_from_code.go
```diff
@@ -787,3 +788,113 @@ func (a *InitFromCodeAction) addToProject(ctx context.Context, targetDir string,
...
+	// Build a lookup from protocol name → version for known protocols.
+	versionOf := make(map[string]string, len(knownProtocols))
+	for _, p := range knownProtocols {
+		versionOf[p.Name] = p.Version
+	}
+
+	// If explicit flag values were provided, use them directly.
+	if len(flagProtocols) > 0 {
+		records := make([]agent_yaml.ProtocolVersionRecord, 0, len(flagProtocols))
+		for _, name := range flagProtocols {
```
jongio, line 823: **[MEDIUM]** `--protocol responses --protocol responses` produces duplicate entries in agent.yaml. Cobra's `StringSliceVar` preserves duplicates from repeated flags and comma-separated values.

```suggestion
	if len(flagProtocols) > 0 {
		seen := make(map[string]bool, len(flagProtocols))
		records := make([]agent_yaml.ProtocolVersionRecord, 0, len(flagProtocols))
		for _, name := range flagProtocols {
			if seen[name] {
				continue
			}
			seen[name] = true
```

cli/azd/extensions/azure.ai.agents/internal/cmd/init_from_code.go
```diff
@@ -787,3 +788,113 @@ func (a *InitFromCodeAction) addToProject(ctx context.Context, targetDir string,
...
+		return nil, fmt.Errorf("failed to prompt for protocols: %w", err)
+	}
+
+	// Collect selected protocols.
+	var records []agent_yaml.ProtocolVersionRecord
+	for _, choice := range resp.Values {
+		if choice.Selected {
+			records = append(records, agent_yaml.ProtocolVersionRecord{
+				Protocol: choice.Value,
+				Version:  versionOf[choice.Value],
```
jongio: **[LOW]** `versionOf[choice.Value]` silently returns `""` if the value isn't in the map. The choices are built from `knownProtocols` so this is safe today, but a defensive check would catch unexpected prompt service behavior.

jongio (commented):
Issues to address:
- init_from_code.go:822 - `--protocol responses --protocol responses` produces duplicate entries in agent.yaml; needs dedup
- init_from_code.go:877 - multi-select response values aren't validated against the known map; silent empty version on unexpected values

rajeshkamal5050 (approved):

PR #7484, opened 2026-04-02
[Agents extension] Detect agent manifest at cwd

trangevi:
On `init`, detect if there's a manifest in the current working directory and ask if the user wants to use that one.

### DIFF
--- a/cli/azd/extensions/azure.ai.agents/internal/cmd/init.go
+++ b/cli/azd/extensions/azure.ai.agents/internal/cmd/init.go
@@ -217,7 +217,11 @@ func newInitCommand(rootFlags *rootFlagsDefinition) *cobra.Command {
 				if checkDir == "" {
 					checkDir = "."
 				}
-				if detected := detectLocalManifest(checkDir); detected != "" {
+				detected, detectErr := detectLocalManifest(checkDir)
+				if detectErr != nil {
+					return fmt.Errorf("checking for existing manifest: %w", detectErr)
+				}
+				if detected != "" {
 					useExisting := flags.NoPrompt
 					if !flags.NoPrompt {
 						confirmResp, promptErr := azdClient.Prompt().Confirm(ctx, &azdext.ConfirmRequest{
--- a/cli/azd/extensions/azure.ai.agents/internal/cmd/init_from_code.go
+++ b/cli/azd/extensions/azure.ai.agents/internal/cmd/init_from_code.go
@@ -71,6 +71,37 @@ func (a *InitFromCodeAction) Run(ctx context.Context) error {
 		a.flags.src = relPath
 	}
 
+	// Default src to current directory when not specified
+	srcDir := a.flags.src
+	if srcDir == "" {
+		srcDir = "."
+	}
+
+	// Check if agent.yaml already exists before the interactive setup so the user
+	// doesn't complete the full agent configuration only to have it discarded.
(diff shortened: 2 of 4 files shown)


cli/azd/extensions/azure.ai.agents/internal/cmd/init_from_code.go
```diff
@@ -85,6 +85,28 @@ func (a *InitFromCodeAction) Run(ctx context.Context) error {
 			srcDir = "."
 		}
 
+		// Check if agent.yaml already exists and prompt before overwriting
```
jongio: **[HIGH]** Overwrite check runs after the full interactive agent setup.

The user completes `createDefinitionFromLocalAgent` at line 76 (agent name, model selection, protocol config) and only then gets asked "agent.yaml exists, overwrite?" If they decline, all that interactive work is discarded. Consider hoisting the stat+confirm block before `createDefinitionFromLocalAgent` - `srcDir` only depends on `a.flags.src` and can be determined earlier.

cli/azd/extensions/azure.ai.agents/internal/cmd/init_from_templates_helpers.go
```diff
@@ -260,3 +261,32 @@ func findAgentManifest(dir string) (string, error) {
...
 	return found, nil
 }
+
+// detectLocalManifest checks only the immediate directory for an agent manifest file.
+// Returns the path to the found manifest, preferring agent.manifest.yaml over agent.yaml,
+// or an empty string if neither exists or none contain valid manifest content.
+func detectLocalManifest(dir string) string {
+	// Prefer agent.manifest.yaml (richer manifest format) over agent.yaml.
+	for _, name := range []string{"agent.manifest.yaml", "agent.yaml"} {
+		candidate := filepath.Join(dir, name)
```
jongio, line 279: **[MEDIUM]** `os.Stat` errors other than "not found" are silently treated as "no manifest."  

Permission errors, broken symlinks, etc. are swallowed. `findAgentManifest` above returns `(string, error)` but this returns just `string`. A valid manifest the user can't read is silently skipped with no indication why.

jongio (commented):
Two features: manifest auto-detection at CWD and overwrite confirmation in init-from-code. Detection logic is solid with content validation and good test coverage (9 cases).

Issues to address:
- init_from_code.go:88 - overwrite check runs after the full interactive setup; declining discards the user's work
- init_from_templates_helpers.go:271 - stat errors other than "not found" silently treated as no manifest

Minor: detectLocalManifest uses exact filename matching while findAgentManifest uses case-insensitive - consider aligning.

cli/azd/extensions/azure.ai.agents/internal/cmd/init.go
```diff
@@ -210,6 +210,42 @@ func newInitCommand(rootFlags *rootFlagsDefinition) *cobra.Command {
...
+							}
+							return fmt.Errorf("prompting for manifest detection: %w", promptErr)
+						}
+						useExisting = *confirmResp.Value
+					}
+					if useExisting {
+						flags.manifestPointer = detected
+						if flags.src == "" {
+							flags.src = checkDir
+						}
```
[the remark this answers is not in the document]
trangevi: We have that check elsewhere, this is an explicit decision

PR #7490, opened 2026-04-03
Initial changes for stream logs command

saanikaguptamicrosoft:


cli/azd/extensions/azure.ai.customtraining/internal/cmd/job_stream.go
```diff
@@ -0,0 +1,347 @@
...
+					probeResp.Body.Close()
+					lastJobStatus = probeResp.JobStatus
+				}
+
+				if lastJobStatus != "" && models.TerminalStatuses[lastJobStatus] {
+					fmt.Printf("\nJob '%s' is in terminal state '%s' with no log files.\n", name, lastJobStatus)
+					fmt.Println("Use 'azd ai training job download' to download job artifacts.")
+					return nil
+				}
+
```
jongio, line 131: **Bug: the probe mechanism can't work.** `GetArtifactContent` returns `(nil, nil)` on HTTP 404 - the client closes the body and returns nil before this code can read `X-VW-Job-Status`. So `probeResp` is always nil here, `lastJobStatus` is never set, and the terminal-state check on line 126 never fires.

If a job finishes with no log files, this loop runs forever (only killed by Ctrl+C).

Fix: use `apiClient.GetJob(ctx, name)` to check status directly - same pattern as `job_download.go:82`:
```go
job, jobErr := apiClient.GetJob(ctx, name)
if jobErr == nil && models.TerminalStatuses[job.Properties.Status] {
    fmt.Printf("\nJob '%s' is in terminal state '%s' with no log files.\n", name, job.Properties.Status)
    fmt.Println("Use 'azd ai training job download' to download job artifacts.")
    return nil
}
```

cli/azd/extensions/azure.ai.customtraining/internal/cmd/job_stream.go
```diff
@@ -0,0 +1,347 @@
...
+				fmt.Println("(Discovering log files...)")
+				time.Sleep(discoveryRetryDelay)
+			}
```
jongio, line 134: **No retry limit on discovery.** If the job name is misspelled or logs never appear, this loops forever with no upper bound. Every other retry path in this file has `maxConsecErrors` - discovery should too.

Add a max discovery attempts constant and check it:
```go
const maxDiscoveryAttempts = 30 // 30 * 5s = 2.5 min max wait
discoveryAttempts := 0
// inside loop:
discoveryAttempts++
if discoveryAttempts >= maxDiscoveryAttempts {
    return fmt.Errorf("no log files found after %d attempts - verify job name '%s' is correct", maxDiscoveryAttempts, name)
}
```

cli/azd/extensions/azure.ai.customtraining/internal/cmd/job_stream.go
```diff
@@ -0,0 +1,347 @@
...
+
+			sort.Strings(logFiles)
```
jongio, line 136: `sort.Strings` doesn't match the codebase. `job_get.go` uses `slices.Sorted()` consistently. Per AGENTS.md, prefer `slices` over `sort`.

```suggestion
slices.Sort(logFiles)
```

Also update the import to replace `"sort"` with `"slices"` (and the other `sort.Strings` on line 219).

cli/azd/extensions/azure.ai.customtraining/internal/cmd/job_stream.go
```diff
@@ -0,0 +1,347 @@
...
+				// Re-discover files — new log files may appear during the run
+				artifactList, err := apiClient.ListArtifactsInPath(ctx, name, logPathPrefix)
+				if err != nil {
+					consecErrors++
+					if consecErrors >= maxConsecErrors {
+						return fmt.Errorf("failed to list log files after %d retries: %w", maxConsecErrors, err)
+					}
+					pollInterval = backoff(pollInterval)
+					continue
+				}
```
jongio, line 210: `consecErrors` only resets on successful file reads (line 262), not after a successful `ListArtifactsInPath`. A sequence like: list fails, list succeeds, read fails, list succeeds, read fails hits the limit even though recent operations succeeded. Reset after the successful list call too:

```go
// After the artifactList loop (line 222), add:
consecErrors = 0
```

cli/azd/extensions/azure.ai.customtraining/internal/cmd/job_stream.go
```diff
@@ -0,0 +1,347 @@
...
+}
+
+func printFileHeader(path string) {
+	fmt.Printf("\nStreaming %s\n", path)
+	fmt.Println("==========================================")
+}
+
+// backoff doubles the interval up to maxPollInterval.
+func backoff(current time.Duration) time.Duration {
+	next := time.Duration(math.Min(float64(current*2), float64(maxPollInterval)))
```
jongio, line 345: `math.Min` with float64 casts is unnecessary - Go has built-in `min()` since 1.21. Also avoids potential precision loss with large `time.Duration` values.

```suggestion
func backoff(current time.Duration) time.Duration {
return min(current*2, maxPollInterval)
}
```

This also removes the `"math"` import.

cli/azd/extensions/azure.ai.customtraining/internal/cmd/job_stream.go
```diff
@@ -0,0 +1,347 @@
...
+				}
+
+				if len(data) > 0 {
+					printFileHeader(path)
+					files[path].headerShown = true
+					fmt.Print(string(data))
+					if !strings.HasSuffix(string(data), "\n") {
+						fmt.Println()
+					}
+				}
```
jongio, line 178: This "print data, ensure trailing newline" block appears 3 times (here, line 269, line 341). Extract it:
```go
func printContent(data []byte) {
    fmt.Print(string(data))
    if !strings.HasSuffix(string(data), "\n") {
        fmt.Println()
    }
}
```

cli/azd/extensions/azure.ai.customtraining/internal/cmd/job_stream.go
```diff
@@ -0,0 +1,347 @@
...
+							fmt.Print(string(data))
+							if !strings.HasSuffix(string(data), "\n") {
+								fmt.Println()
+							}
+						}
+
+						if contentLen > fs.offset {
+							fs.offset = contentLen
+						}
+					}
```
jongio, line 324: The final flush loop (lines 291-324) duplicates the file-reading logic from the main poll loop. If the read logic changes (e.g., error handling, offset tracking), both copies need updating. Consider extracting a `pollFile(ctx, apiClient, jobName, path, fs)` helper that both the main loop and final flush can call.

cli/azd/extensions/azure.ai.customtraining/pkg/models/job.go
```diff
@@ -42,3 +42,19 @@ var TerminalStatuses = map[string]bool{
...
+var InProgressStatuses = map[string]bool{
+	"NotStarted":   true,
+	"Queued":       true,
+	"Preparing":    true,
+	"Provisioning": true,
+	"Starting":     true,
+	"Running":      true,
+}
+
+// StreamableStatuses returns true if the job status indicates streaming is appropriate.
```
jongio, line 56: Doc comment says "StreamableStatuses returns" but the function is `IsStreamableStatus`. Per Go conventions, doc comments should start with the function name.

```suggestion
// IsStreamableStatus returns true if the job status indicates streaming is appropriate.
```

PR #7497, opened 2026-04-03
Enable extension CI builds — daily + PR

### LINKED ISSUE #7317: [Extensions] - Create a better story for private/internal builds
Testing features and changes before a release is challenging for the agents extension. We need a simple way to share builds to partner teams that can help test integration with their features, etc. that doesn't require do a full dev build from source or waiting for a release.

rajeshkamal5050:
## What

Daily and PR extension builds to Azure Storage, mirrors core CLI build pattern.

On CI push to main — extensions get signed and uploaded to storage with per-extension registry entries.
On PR builds — same thing but scoped to the PR number, with install instructions posted as a PR comment.

### Changes

- `publish-extension-daily.yml` — daily publish stage, depends on Sign, uploads to storage
- `publish-extension-pr.yml` — PR publish stage, uploads to PR-scoped storage, posts install comment
- `publish-extension.yml` — `CreateGitHubRelease` param so daily/PR can skip it
- `release-azd-extension.yml` — wired daily + PR publish for CI and PR builds
- `Set-ExtensionVersionInBuild.ps1` — appends `-daily.` or `-pr.` suffix per build reason
- `Update-ExtensionDailyRegistry.ps1` + JSON template — generates per-extension registry entries with checksums
- `update-extension-pr-registry.yml` — uploads PR-scoped registry entry
- All extensions version bumped so dailies are ahead of released versions

### Per-extension registry

Each extension uploads its own entry. No shared blob, no race condition when multiple extensions build concurrently.

### How to use — daily builds

Daily registry entries at:
```
https://azuresdkartifacts.z5.web.core.windows.net/azd/extensions/daily-registry-entries/{extension-id}.json
```

Install a daily build:
```bash
azd ext source add -n daily -t url -l "https://azuresdkartifacts.z5.web.core.windows.net/azd/extensions/daily-registry-entries/azure.ai.agents.json"
azd ext install azure.ai.agents --source daily
```

### How to use — PR builds

PR builds post a comment on the PR with install instructions. Registry entries at:
```
https://azuresdkartifacts.z5.web.core.windows.net/azd/extensions/pr-registry-entries/{pr-number}/{extension-id}.json
```

Install a PR build:
```bash
azd ext source add -n pr-7497 -t url -l "https://azuresdkartifacts.z5.web.core.windows.net/azd/extensions/pr-registry-entries/7497/azure.ai.agents.json"
azd ext install azure.ai.agents --source pr-7497
```

Closes #7317


eng/pipelines/templates/stages/publish-extension-dev.yml
```diff
@@ -0,0 +1,111 @@
...
+              targetPath: release-metadata
+
+        strategy:
+          runOnce:
+            deploy:
+              steps:
+                - pwsh: |
+                    $tag = "${{ parameters.SanitizedExtensionId }}-dev"
+                    gh release delete $tag --repo $(Build.Repository.Name) --yes --cleanup-tag 2>$null
+                    Write-Host "Cleaned up existing dev release: $tag"
```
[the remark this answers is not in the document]
rajeshkamal5050: This file has been removed — daily builds no longer use GitHub Releases. The publish stage now uploads directly to Azure Storage via azcopy, with proper error handling in `Update-ExtensionDailyRegistry.ps1`.

eng/pipelines/templates/stages/publish-extension-dev.yml
```diff
@@ -0,0 +1,111 @@
...
+                    zip release/$id-darwin-amd64.zip -j build/darwin-amd64/$id-darwin-amd64 release-metadata/extension.yaml
+                    zip release/$id-darwin-arm64.zip -j build/darwin-arm64/$id-darwin-arm64 release-metadata/extension.yaml
+                    zip release/$id-windows-amd64.zip -j build/windows-amd64/$id-windows-amd64.exe release-metadata/extension.yaml
+                    zip release/$id-windows-arm64.zip -j build/windows-arm64/$id-windows-arm64.exe release-metadata/extension.yaml
+
+                    # tar.gz linux
+                    Copy-Item release-metadata/extension.yaml build/linux-amd64/
+                    Copy-Item release-metadata/extension.yaml build/linux-arm64/
+                    tar -C build/linux-amd64/ -cvzf release/$id-linux-amd64.tar.gz $id-linux-amd64 extension.yaml
+                    tar -C build/linux-arm64/ -cvzf release/$id-linux-arm64.tar.gz $id-linux-arm64 extension.yaml
```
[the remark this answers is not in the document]
rajeshkamal5050: This file has been removed. Daily builds now go through the Sign stage (same as releases), so artifacts already include NOTICE.txt and extension.yaml via the signed packaging pipeline.

cli/azd/extensions/registry-dev.json
```diff
@@ -0,0 +1,300 @@
...
+      "description": "Dev build of Demo Extension. Built from main branch. Not for production use.",
+      "versions": [
+        {
+          "version": "0.0.0-dev",
+          "artifacts": {
+            "darwin/amd64": {
+              "url": "https://github.com/Azure/azure-dev/releases/download/microsoft-azd-demo-dev/microsoft-azd-demo-darwin-amd64.zip",
+              "entryPoint": "microsoft-azd-demo-darwin-amd64"
+            },
+            "darwin/arm64": {
```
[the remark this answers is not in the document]
rajeshkamal5050: registry-dev.json has been deleted. The registry is now generated dynamically by `Update-ExtensionDailyRegistry.ps1` which reads capabilities and providers from each extension's `extension.yaml` and populates them in the registry entry.

vhvb1989 (approved):

rajeshkamal5050:
@danieljurek waiting on your review before merging this. CI/pipeline changes need your eyes.

eng/pipelines/templates/stages/build-and-test-azd-extension.yml
```diff
@@ -110,6 +110,10 @@ stages:
         steps:
           - checkout: self
 
+          - template: /eng/pipelines/templates/steps/set-extension-version-cd.yml
```
jongio, line 113: **[HIGH]** Version override only runs here in GenerateReleaseArtifacts. The per-platform build jobs (`build-azd-extension.yml`) and cross-build jobs also read `version.txt` via `Set-ExtensionVersionVariable.ps1` and pass `-Version $(EXT_VERSION)` to the build. So compiled binaries get the base version (e.g., `0.1.21-preview`) while the registry and storage paths use the daily version (`0.1.21-preview-daily.12345`).

The core CLI handles this by running the version override in all 3 job types. Add `set-extension-version-cd.yml` to `build-azd-extension.yml` and `cross-build-azd-extension.yml` before `Set-ExtensionVersionVariable.ps1`.

eng/scripts/Update-ExtensionDailyRegistry.ps1
```diff
@@ -0,0 +1,247 @@
...
+$azcopyExitCode = $LASTEXITCODE
+$ErrorActionPreference = $prevErrorPref
+
+if ($azcopyExitCode -ne 0) {
+    # Check if this is a "not found" (expected) vs an actual error
+    $outputStr = $azcopyOutput | Out-String
+    if ($outputStr -match "BlobNotFound|404|does not exist") {
+        Write-Host "No existing registry found, creating new one"
+    } else {
+        Write-Warning "azcopy download failed (exit code $azcopyExitCode), creating new registry"
```
jongio: **[HIGH]** When azcopy download fails for reasons other than "not found" (auth failure, network timeout, throttling), this logs a warning but still creates an empty registry on L205. The subsequent upload then wipes all existing entries.

Fail hard on non-404 errors. Move the empty-registry init inside the 404 branch only:

```
if ($outputStr -match "BlobNotFound|404|does not exist") {
    Write-Host "No existing registry found, creating new one"
    [ordered]@{ extensions = @() } | ConvertTo-Json -Depth 10 | Set-Content $registryFile
} else {
    Write-Error "Failed to download registry (exit $azcopyExitCode): $outputStr"
    exit 1
}
```

eng/scripts/Update-ExtensionDailyRegistry.ps1
```diff
@@ -0,0 +1,247 @@
...
+        Write-Warning $outputStr
+    }
+    [ordered]@{ extensions = @() } | ConvertTo-Json -Depth 10 | Set-Content $registryFile
+}
+
+$registry = Get-Content $registryFile -Raw | ConvertFrom-Json -Depth 20
+
+# Merge: replace existing extension entry or add new
+$found = $false
```
jongio: **[HIGH]** Read-modify-write on a shared blob without concurrency control. When multiple extension pipelines trigger on the same push (this PR bumps all 9), they all download, modify, and upload registry-daily.json concurrently. Last writer wins, earlier entries disappear.

It self-heals on next push (each extension re-adds itself), but between pushes the registry can be incomplete. Options: blob lease, ETag-based optimistic concurrency + retry, or serialize the registry update into a single pipeline stage that fans-in after all extensions publish.

eng/scripts/Update-ExtensionDailyRegistry.ps1
```diff
@@ -0,0 +1,247 @@
...
+    exit 1
+}
+
+# Read extension.yaml for metadata
+# Simple line-by-line parsing for top-level scalar fields, capabilities list,
+# and providers list. Sufficient for the known extension.yaml schema.
+$extYaml = Get-Content $extYamlPath -Raw
+$extMeta = @{}
+foreach ($line in $extYaml -split "`n") {
+    if ($line -match "^(\w[\w\-]*):\s*(.+)$") {
```
jongio: **[MEDIUM]** Regex-based YAML parsing works for today's extension.yaml files but won't handle multi-line values (`description: |`), inline comments, or quoted strings with colons. If any extension adds these, the registry gets wrong metadata silently.

Consider `powershell-yaml` (`ConvertFrom-Yaml`) or add validation that required fields aren't empty/malformed after parsing.

jongio (dismissed):
Version override (`set-extension-version-cd.yml`) only runs in GenerateReleaseArtifacts, not in the per-platform build/cross-build jobs. The core CLI runs this in all 3 job types. Binaries get the base version while the registry lists the daily version.

Issues to address:
- build-and-test-azd-extension.yml:113 - version override missing from BuildExtension and CrossBuild jobs
- Update-ExtensionDailyRegistry.ps1:202 - non-404 azcopy failures silently wipe the registry
- Update-ExtensionDailyRegistry.ps1:211 - concurrent extension pipelines race on registry-daily.json
- Update-ExtensionDailyRegistry.ps1:100 - regex YAML parser won't handle multi-line values or inline comments

eng/pipelines/templates/steps/publish-extension.yml
```diff
@@ -6,63 +6,78 @@ parameters:
     default: '`$web'
   - name: TagPrefix
     type: string
+    default: ''
   - name: TagVersion
     type: string
+    default: ''
+  - name: CreateGitHubRelease
```
hemarina, line 13: **[Medium]** Conditional gap leaves `_manifest` in upload

Adding defaults (`''`) to `TagPrefix` and `TagVersion` created a dead zone: when `CreateGitHubRelease=true` (default) but both tag params are empty (also defaults), neither the `true` branch (guarded by `ne(TagPrefix,'')`) nor the `false` branch fires. The `_manifest` folder is not removed and gets uploaded to storage.

Current callers are correct, but the template interface is misleading for future consumers.

**Suggested Fix:** Make `_manifest` removal unconditional — move it before both conditional blocks so it always runs regardless of parameter combination:

```suggestion
  - name: CreateGitHubRelease
    type: boolean
    default: true

# Always remove _manifest before upload, regardless of release type
```

Or restructure so manifest removal is a standalone step before the conditional GitHub release logic.

eng/scripts/Update-ExtensionDailyRegistry.ps1
```diff
@@ -0,0 +1,247 @@
...
+    '${CHECKSUM_WINDOWS_ARM64}' = $checksums["WINDOWS_ARM64"]
+}
+
+foreach ($placeholder in $replacements.Keys) {
+    $template = $template.Replace($placeholder, $replacements[$placeholder])
+}
+
+$versionEntry = $template | ConvertFrom-Json
+
+# Add capabilities and providers (can't template arrays/objects easily)
```
hemarina, line 186: **[Medium]** Daily registry retains only latest version per extension

Each run builds `versions = @($versionEntry)` (single-element array), then replaces the entire extension entry. Previous daily versions are discarded from the registry. Archive storage paths preserve the artifacts but they become undiscoverable via the registry.

**Suggested Fix:** Add a comment clarifying this is intentional:

```suggestion
# Build the full extension entry.
# Daily registry intentionally keeps only the latest version per extension.
# Previous daily artifacts remain accessible via archive storage paths
# (e.g., <id>/daily/archive/<version>) but are not listed in the registry.
$extEntry = [ordered]@{
```

eng/scripts/Update-ExtensionDailyRegistry.ps1
```diff
@@ -0,0 +1,247 @@
...
+foreach ($line in $extYaml -split "`n") {
+    if ($line -match "^(\w[\w\-]*):\s*(.+)$") {
+        $extMeta[$matches[1]] = $matches[2].Trim().Trim('"')
+    }
+}
+
+# Parse capabilities list
+$capabilities = @()
+$inCapabilities = $false
+foreach ($line in $extYaml -split "`n") {
```
hemarina: **[Low]** `usage` and `requiredAzdVersion` not validated after YAML parsing

Required field validation checks `namespace`, `displayName`, `description` — but `usage` and `requiredAzdVersion` are also used in the registry template. If these are missing from `extension.yaml`, the registry entry gets empty strings which could confuse `azd ext install`.

Consider adding `usage` to the required fields list, or logging a warning when optional template fields resolve to empty.

eng/scripts/Update-ExtensionDailyRegistry.ps1
```diff
@@ -0,0 +1,247 @@
+<#
```
hemarina, line 1: **[Low]** No test coverage for new PowerShell scripts

Both `Set-ExtensionVersionInBuild.ps1` and `Update-ExtensionDailyRegistry.ps1` contain non-trivial logic (version string manipulation, YAML parsing, registry merging, checksum computation) but have no Pester tests. Basic unit tests for the deterministic functions — version suffix appending, YAML field extraction, and registry merge logic — would prevent regressions.

hemarina (commented):
Thanks for adding this feature! Have some comments.

eng/scripts/Update-ExtensionDailyRegistry.ps1
hemarina: I think we just support latest version and this is a design choice. Checking with @danieljurek @wbreza

eng/scripts/Update-ExtensionDailyRegistry.ps1
rajeshkamal5050: No longer applies. Switched to per-extension registry entries — each extension writes its own JSON blob to daily-registry-entries/{id}.json. No shared blob download/merge/upload anymore.

We can probably have a separate pipeline which does the unification of all these registries into one.

eng/scripts/Update-ExtensionDailyRegistry.ps1
rajeshkamal5050: Fixed. Each extension now writes its own per-extension entry (daily-registry-entries/{id}.json) — no shared blob, no race. Unification into registry-daily.json will be a separate follow-up pipeline/PR.

eng/pipelines/templates/stages/build-and-test-azd-extension.yml
rajeshkamal5050: Fixed. Added set-extension-version-cd.yml to both build-azd-extension.yml and cross-build-azd-extension.yml before Set-ExtensionVersionVariable.ps1.

eng/scripts/Update-ExtensionDailyRegistry.ps1
rajeshkamal5050: Fair ask. No pipeline scripts in the repo have Pester tests today. Can add later if needed but not blocking for this PR.

eng/pipelines/templates/steps/publish-extension.yml
rajeshkamal5050: Fixed. Moved _manifest removal to an unconditional step before the conditional GitHub release blocks.

eng/scripts/Update-ExtensionDailyRegistry.ps1
rajeshkamal5050: Added usage to the required fields validation. requiredAzdVersion stays optional — not all extensions set it today.

eng/scripts/Update-ExtensionDailyRegistry.ps1
rajeshkamal5050: Valid point. Added a code comment documenting the limitation. Today's extension.yaml files are all simple flat scalars so this works fine +  the script validates required fields after parsing

@danieljurek do the build pipelines have powershell-yaml available?

eng/scripts/Update-ExtensionDailyRegistry.ps1
rajeshkamal5050: Yes, by design. Same pattern as azd core daily builds — we only keep the latest. Previous daily artifacts are still in storage via the archive path (/{id}/daily/archive/{version}) but not listed in the registry. Added a comment in the code.

eng/pipelines/templates/steps/publish-extension.yml
```diff
@@ -6,63 +6,72 @@ parameters:
...
+    type: boolean
+    default: true
```
jongio: **[MEDIUM]** CreateGitHubRelease defaults true while TagPrefix/TagVersion default empty - silent no-op for callers using all defaults.

A future caller that only passes PublishUploadLocations gets CreateGitHubRelease=true (default) but empty tag params, so the release block silently doesn't compile in. Current callers are fine (daily explicitly passes false, release explicitly passes all three), but this is a trap. Consider adding a compile-time validation that errors when CreateGitHubRelease is true but TagPrefix/TagVersion are empty.

jongio (commented):
Solid PR - per-extension registry entries cleanly avoid the concurrent blob mutation problem. Version stamping is consistent across all build and cross-build jobs now, and the pipeline flow (BuildAndTest -> Sign -> PublishDaily) checks out.

One observation:
- eng/pipelines/templates/steps/publish-extension.yml:15 - parameter defaults create a silent skip trap for future callers (see inline)

Previous feedback from jongio and hemarina looks addressed.

eng/scripts/Update-ExtensionDailyRegistry.ps1
danieljurek: Agents don't have `powershell-yaml` installed by default ... We do have a `Install-ModuleIfNotInstalled` that can install a PowerShell module from the gallery at a given version. It's slow but handles failure cases. 

Example usage here: https://github.com/Azure/azure-sdk-for-java/blob/e115d5d6e723a5c63f39077955c8f409b93a65f2/eng/scripts/Update-Artifacts-List-For-Patch-Release.ps1#L26

eng/templates/extensions-registry-daily.json.template
```diff
@@ -0,0 +1,38 @@
+{
```
danieljurek, line 1: This folder is for YAML templates for Azure DevOps pipeline YAML files. Put this in [`eng/templates`](https://github.com/Azure/azure-dev/tree/main/eng/templates) ... I'd name it `extensions-registry-daily.json.template`

wbreza (commented):
## Code Review — PR #7497

**Enable dev extension builds on CI** by @rajeshkamal5050

### Summary

Great addition to the extension dev workflow! The per-extension daily registry pattern is a clean solution that avoids the concurrent blob mutation problem entirely. The overall architecture — version stamping → build → sign → publish → registry — mirrors the established core CLI daily build pattern well.

Previous feedback from @jongio, @hemarina, and @vhvb1989 has been largely addressed. Two new findings below, plus a few maintenance items.

### Prior Review Regression Check
| # | Prior Finding | Author | Status |
|---|-------------|--------|--------|
| 1 | Version override missing from build/cross-build jobs | @jongio | ✅ Resolved |
| 2 | azcopy failure silently wipes registry | @jongio | ✅ Resolved |
| 3 | Concurrent race on shared blob | @jongio | ✅ Resolved |
| 4 | _manifest removal conditional gap | @hemarina | ✅ Resolved |
| 5 | Missing usage validation | @hemarina | ✅ Resolved |
| 6 | Regex YAML parsing fragility | @jongio | ⏳ Open |
| 7 | CreateGitHubRelease defaults trap | @jongio | ⏳ Open |

### Findings Summary
| Priority | Count |
|----------|-------|
| High | 2 |
| Medium | 3 |
| Low | 1 |
| **Total** | **6** |

### ✅ What Looks Good
- Per-extension registry entries elegantly eliminate the concurrent blob race condition
- SHA256 checksums from signed artifacts ensure supply chain integrity
- Version stamping mirrors the core CLI daily build pattern correctly
- JSON validated locally before upload; `Continue = 'Stop'` prevents silent failures
- Clean stage separation: build → sign → publish

This is a great improvement to the extension dev workflow — having daily builds available mirrors what the core CLI already offers and will significantly improve the inner loop for extension development and testing.

**Overall Assessment:** Comment — High findings warrant discussion, but the core pattern is solid and well-designed.

eng/scripts/Update-ExtensionDailyRegistry.ps1
```diff
@@ -0,0 +1,217 @@
...
+Write-Host "  usage: $($extMeta.usage)"
+Write-Host "  capabilities: $($capabilities -join ', ')"
+Write-Host "  providers: $($providers.Count)"
+
+# Load template and replace placeholders
+$template = Get-Content $TemplatePath -Raw
+$replacements = @{
+    '${EXT_VERSION}'            = $Version
+    '${REQUIRED_AZD_VERSION}'   = if ($extMeta.requiredAzdVersion) { $extMeta.requiredAzdVersion } else { "" }
+    '${USAGE}'                  = $extMeta.usage
```
wbreza: **[High]** Per-extension JSON format incompatible with azd URL source

The per-extension entry is a bare object `{ id, namespace, ..., versions: [...] }` but `azd ext source add -t url` feeds the JSON through `newUrlSource()` → `newJsonSource()` → `json.Unmarshal(&Registry{})` which expects `{ "extensions": [...] }` (confirmed in `cli/azd/pkg/extensions/json_source.go:13-14` and `registry.go:27-30`). Consumers will get an empty extension list.

The documented consumer flow (`azd ext source add -n daily -t url -l "...azure.ai.agents.json"` + `azd ext install`) won't find any extensions to install.

**Suggested Fix:**
```suggestion
# Build the per-extension entry wrapped in registry format
$registry = [ordered]@{ extensions = @($extEntry) }
$entryFile = "$AzdExtensionId.json"
$registry | ConvertTo-Json -Depth 20 | Set-Content $entryFile -Encoding utf8
```

eng/scripts/Set-ExtensionVersionInBuild.ps1
```diff
@@ -0,0 +1,45 @@
...
+    $prereleaseCategory = "daily"
+}
+
+$versionFile = Join-Path $ExtensionDirectory "version.txt"
+$version = (Get-Content $versionFile).Trim()
+if ([string]::IsNullOrWhiteSpace($version)) {
+    Write-Error "version.txt is empty at $versionFile"
+    exit 1
+}
+$newVersion = "$version-$prereleaseCategory.$BuildId"
```
wbreza, line 53: **[High]** Script not idempotent — pipeline retry appends duplicate suffix

If the pipeline retries this step, the script reads the already-modified `version.txt` (e.g., `0.1.21-preview-daily.123`) and appends again: `0.1.21-preview-daily.123-daily.123`. This produces invalid semver.

Consider adding a guard:
```powershell
if ($version -match "-$prereleaseCategory\.\d+$") {
    Write-Host "Version already has $prereleaseCategory suffix, skipping"
    exit 0
}
```

eng/scripts/Update-ExtensionDailyRegistry.ps1
```diff
@@ -0,0 +1,217 @@
...
+# Parse providers list
+$providers = @()
+$inProviders = $false
+$currentProvider = $null
+foreach ($line in $extYaml -split "`n") {
+    if ($line -match "^providers:") { $inProviders = $true; continue }
+    if ($inProviders -and $line -match "^\s+-\s+name:\s*(.+)$") {
+        if ($currentProvider) { $providers += $currentProvider }
+        $currentProvider = [ordered]@{ name = $matches[1].Trim() }
+    } elseif ($inProviders -and $currentProvider -and $line -match "^\s+(\w+):\s*(.+)$") {
```
wbreza: **[Medium]** No validation that all template placeholders were replaced

Sequential `String.Replace()` doesn't verify completeness. If a new placeholder is added to the JSON template but missed in the replacements dict, literal `$`{...}` text silently passes through to the uploaded JSON.

Also, if a metadata value (e.g., usage) happens to contain a placeholder pattern like `$`{CHECKSUM_LINUX_AMD64}`, it would be resolved in a later replacement pass. Low real-world risk but worth guarding against.

Consider adding after the replacement loop:
```powershell
if ($template -match '\$\{[A-Z_]+\}') {
    Write-Error "Unreplaced placeholder found: $($matches[0])"
    exit 1
}
```

eng/scripts/Update-ExtensionDailyRegistry.ps1
```diff
@@ -0,0 +1,217 @@
...
+$checksums = @{}
+$missingArtifacts = @()
+$artifactFiles = @(
+    @{ key = "DARWIN_AMD64";  file = "$SanitizedExtensionId-darwin-amd64.zip" },
+    @{ key = "DARWIN_ARM64";  file = "$SanitizedExtensionId-darwin-arm64.zip" },
+    @{ key = "LINUX_AMD64";   file = "$SanitizedExtensionId-linux-amd64.tar.gz" },
+    @{ key = "LINUX_ARM64";   file = "$SanitizedExtensionId-linux-arm64.tar.gz" },
+    @{ key = "WINDOWS_AMD64"; file = "$SanitizedExtensionId-windows-amd64.zip" },
+    @{ key = "WINDOWS_ARM64"; file = "$SanitizedExtensionId-windows-arm64.zip" }
+)
```
wbreza, line 73: **[Medium]** Three separate loops parse the same YAML file

Top-level scalars (lines 62-66), capabilities (lines 68-76), and providers (lines 78-89) each iterate over `$`extYaml` independently. A single-pass approach would be more maintainable and avoid potential inconsistencies if the file is modified between parsing phases (not possible here, but a good practice).

eng/templates/extensions-registry-daily.json.template
```diff
@@ -0,0 +1,38 @@
...
+  "version": "${EXT_VERSION}",
+  "requiredAzdVersion": "${REQUIRED_AZD_VERSION}",
+  "capabilities": [],
+  "usage": "${USAGE}",
+  "artifacts": {
```
wbreza, line 6: **[Low]** Platform artifact definitions scattered across 3 locations

Platform keys (`darwin/amd64`, etc.) are hardcoded here, in `Update-ExtensionDailyRegistry.ps1`'s `artifactFiles` array, and in the replacements dict. Adding a new platform requires synchronized changes in all three. Consider generating the template dynamically or defining platforms in a single config.

eng/scripts/Update-ExtensionDailyRegistry.ps1
```diff
@@ -0,0 +1,217 @@
...
+
+# Read extension.yaml for metadata.
+# Uses simple line-by-line regex parsing — handles top-level scalar fields,
+# capabilities list, and providers list. This is intentionally not a full YAML
+# parser. It works for the known extension.yaml schema where all values are
+# single-line scalars or simple lists. If extension.yaml grows multi-line
+# values or complex nesting, switch to powershell-yaml.
+$extYaml = Get-Content $extYamlPath -Raw
+$extMeta = @{}
+foreach ($line in $extYaml -split "`n") {
```
danieljurek: I think using a powershell YAML parser is the right way to go here. Otherwise this code could start failing in unexpected ways if there is a change to the YAML schema without also updating this parser. 

This one is used in other places in the Azure SDK repo: https://www.powershellgallery.com/packages/powershell-yaml ... I recommend using the latest version.. Make sure to pin the version because there could be breaking changes in the future.

eng/pipelines/templates/steps/update-extension-daily-registry.yml
```diff
@@ -0,0 +1,29 @@
...
+      azurePowerShellVersion: LatestVersion
+      pwsh: true
+      ScriptType: InlineScript
+      Inline: |
+        $storageHost = "$(publish-storage-static-host)"
+        $dailyBaseUrl = "$storageHost/azd/extensions/${{ parameters.SanitizedExtensionId }}/daily"
+        $entryBlobPath = "$(publish-storage-location)/`$web/azd/extensions/daily-registry-entries/${{ parameters.AzdExtensionId }}.json"
+        $templatePath = "$(Build.SourcesDirectory)/eng/pipelines/templates/json/extension-registry-daily-template.json"
+
+        & "$(Build.SourcesDirectory)/eng/scripts/Update-ExtensionDailyRegistry.ps1" `
```
danieljurek: This step assumes that the repo is checked out and accessible, but in release deployments, that is not the case. Instead, produce the output template in `GenerateReleaseArtifacts` and put it in the `release-metadata` artifact. Then fetch it out in the release job and azcopy it to the right place.

danieljurek (commented):
We'll need to test this e2e to make sure it works before merging it.

eng/scripts/Update-ExtensionDailyRegistry.ps1
rajeshkamal5050: Done. Switched to powershell-yaml via Install-ModuleIfNotInstalled.

eng/pipelines/templates/steps/publish-extension.yml
rajeshkamal5050: Done. Flipped default to false. Release caller passes true explicitly.

eng/templates/extensions-registry-daily.json.template
rajeshkamal5050: Done. Moved to eng/templates/extensions-registry-daily.json.template.

eng/scripts/Update-ExtensionDailyRegistry.ps1
rajeshkamal5050: Done. Wrapped in `{ "extensions": [...] }` format that azd expects.

eng/scripts/Set-ExtensionVersionInBuild.ps1
rajeshkamal5050: Done. Added guard — skips if suffix already present.

eng/scripts/Update-ExtensionDailyRegistry.ps1
rajeshkamal5050: Done. Added regex check for unreplaced placeholders after substitution.

eng/scripts/Update-ExtensionDailyRegistry.ps1
rajeshkamal5050: Done. Single powershell-yaml parse now handles all fields, capabilities, and providers.

eng/templates/extensions-registry-daily.json.template
rajeshkamal5050: Noted. Will look at consolidating in a follow-up.

eng/scripts/Update-ExtensionDailyRegistry.ps1
rajeshkamal5050: Done. Using Install-ModuleIfNotInstalled from PSModule-Helpers.ps1 + powershell-yaml 0.4.7.

eng/pipelines/templates/steps/update-extension-daily-registry.yml
rajeshkamal5050: Done. Template and scripts now copied into release-metadata artifact during build. Publish stage references those.

cli/azd/extensions/azure.ai.agents/version.txt
```diff
@@ -1 +1 @@
-0.1.20-preview
\ No newline at end of file
+0.1.21-preview
```
jongio: version.txt bumped to 0.1.21-preview but extension.yaml still has version: 0.1.20-preview. The extension.yaml comment says "Make sure version.txt is in sync with this version." All other 8 extensions were updated in both files.

eng/scripts/Update-ExtensionDailyRegistry.ps1
```diff
@@ -0,0 +1,220 @@
...
+Write-Host "  namespace: $($extMeta.namespace)"
+Write-Host "  displayName: $($extMeta.displayName)"
+Write-Host "  description: $($extMeta.description)"
+Write-Host "  usage: $($extMeta.usage)"
+Write-Host "  capabilities: $($capabilities -join ', ')"
+Write-Host "  providers: $($providers.Count)"
+
+# Load template and replace placeholders
+$template = Get-Content $TemplatePath -Raw
+$replacements = @{
```
jongio, line 160: `String.Replace()` puts the raw usage value into a JSON string literal without escaping. If a future extension.yaml has usage containing `"` or `\`, the template produces invalid JSON. The `ConvertFrom-Json` check on L172 catches it but gives a cryptic error.

Consider JSON-escaping replacement values before substitution, or building the version entry entirely as a PowerShell object and serializing with `ConvertTo-Json` (which handles escaping).

jongio (commented):
Issues to address:
- cli/azd/extensions/azure.ai.agents/extension.yaml - version field still says 0.1.20-preview but version.txt was bumped to 0.1.21-preview. All other extensions were synced in both files.
- eng/scripts/Update-ExtensionDailyRegistry.ps1:155 - raw string replacement of ${USAGE} into the JSON template doesn't JSON-escape the value. Current extension.yaml files are fine, but if any usage field ever contains a double-quote or backslash, the generated JSON breaks silently. ConvertFrom-Json catches it downstream, but the error won't point at the cause.

Per-extension registry approach is solid - cleanly avoids the concurrent blob mutation problem.

wbreza (approved):
## Code Re-Review — PR #7497

**Enable dev extension builds on CI** by @rajeshkamal5050

### Re-review after force-push

Changes since my Apr 7 review (`b3e211d`) are in commit `fe4304a`:
- **Regex → powershell-yaml** for proper YAML parsing (addresses my prior finding)
- **Registry format** wrapped in `{ extensions: [...] }` for azd compatibility
- **CreateGitHubRelease default** flipped to false (safe default)
- **Idempotency guard** for version script pipeline retries
- **Null checks + file-exists guards** throughout
- **Template moved** to eng/templates/ with split copy steps

### Prior Review: All :white_check_mark: Resolved
| 1 | High | Regex YAML parsing fragility | :white_check_mark: Replaced with powershell-yaml |
| 2 | High | Per-extension registry entries (race condition fix) | :white_check_mark: Already resolved |
| 3 | Medium | Version override in build/cross-build jobs | :white_check_mark: Resolved |
| 4 | Medium | _manifest removal | :white_check_mark: Unconditional |
| 5 | Low | Usage validation | :white_check_mark: Added |

### :white_check_mark: What Looks Good
- powershell-yaml replaces fragile regex parsing — proper YAML handling
- Per-extension registry entries elegantly eliminate concurrent blob mutation
- SHA256 checksums from signed artifacts ensure supply chain integrity
- Safe defaults: CreateGitHubRelease=false, explicit true in release caller
- Idempotency guard prevents duplicate version suffixes on retries
- Daily publish stage correctly scoped to CI builds (IndividualCI/BatchedCI) — PR builds skip it

**Overall Assessment:** Re-approve — all findings addressed, no new issues. Pending @danieljurek's E2E validation.

cli/azd/extensions/azure.ai.agents/version.txt
rajeshkamal5050: fixed — both version.txt and extension.yaml now at 0.1.22-preview (bumped past released 0.1.21).

eng/scripts/Update-ExtensionDailyRegistry.ps1
rajeshkamal5050: fixed — added ConvertTo-JsonSafeString helper that uses ConvertTo-Json for proper escaping. applied to version, usage, sanitizedId, storageBaseUrl, requiredAzdVersion replacements.

rajeshkamal5050:
@danieljurek all 9 extension pipelines should be on the internal ADO project right? currently only azure.ai.models and azure.appservice are internal, the rest run on public and skip sign + publish stages.

these did not get PublishForPR:
- azure.ai.agents
- azure.ai.finetune
- azure.coding-agent
- microsoft.azd.ai.builder
- microsoft.azd.concurx
- microsoft.azd.demo
- microsoft.azd.extensions

eng/pipelines/templates/stages/build-and-test-azd-extension.yml
```diff
@@ -95,7 +95,10 @@ stages:
             OSVmImage: ${{ build.value.OSVmImage }}
             OS: ${{ build.value.OS }}
             Variables: ${{ build.value.Variables }}
-            ValidateCrossCompile: ${{ build.value.ValidateCrossCompile }}
+            # TODO: Revert this before merge
+            # ValidateCrossCompile: ${{ build.value.ValidateCrossCompile }}
+            # TODO: Remove this line before merge
+            ValidateCrossCompile: false
```
jongio: These 3 TODO items (`SkipTests: true` in the demo pipeline, and the `ValidateCrossCompile` override here) need to be reverted before merge - they disable cross-compile validation for all extensions and skip tests for the demo extension.

jongio (dismissed):
3 TODO debug shortcuts need to be reverted before merge - they disable cross-compile validation for all extensions and skip tests for the demo extension.

eng/pipelines/release-ext-microsoft-azd-demo.yml
```diff
@@ -29,6 +29,8 @@ extends:
     stages:
       - template: /eng/pipelines/templates/stages/release-azd-extension.yml
         parameters: 
+          # TODO: Revert before merging
```
danieljurek: There's also this to revert

eng/pipelines/templates/stages/build-and-test-azd-extension.yml
```diff
@@ -157,6 +164,18 @@ stages:
...
+              Copy-Item eng/templates/extensions-registry-daily.json.template release-metadata/extensions-registry-daily.json.template
+            displayName: Copy registry template to release-metadata
+
+          - pwsh: |
+              Copy-Item eng/scripts/Update-ExtensionDailyRegistry.ps1 release-metadata/Update-ExtensionDailyRegistry.ps1
+            displayName: Copy Update-ExtensionDailyRegistry.ps1 to release-metadata
+
+          - pwsh: |
+              Copy-Item eng/common/scripts/Helpers/PSModule-Helpers.ps1 release-metadata/PSModule-Helpers.ps1
+            displayName: Copy PSModule-Helpers.ps1 to release-metadata
```
danieljurek: This is incorrect. It puts executable items into artifacts which are then called during release. Instead, call the scripts that produce the updated `extensions-registry-daily.json` in this job, upload that artifact to a separate artifact called something like `extension-registry-daily` and then in the publishing step, download that artifact and upload it as needed.

eng/pipelines/templates/steps/update-extension-daily-registry.yml
```diff
@@ -0,0 +1,31 @@
+parameters:
```
danieljurek: This template is unnecessary. Just call this script from the `GenerateReleaseArtifacts` job. The script should only produce the output JSON file. It shouldn't do uploads.

eng/pipelines/templates/steps/update-extension-pr-registry.yml
```diff
@@ -0,0 +1,33 @@
+parameters:
```
danieljurek: This template is unnecessary. Just call this script from the `GenerateReleaseArtifacts` job. The script should only produce the output JSON file. It shouldn't do uploads.

eng/scripts/Update-ExtensionDailyRegistry.ps1
```diff
@@ -0,0 +1,228 @@
...
+
+Write-Host "Extension entry:"
+Get-Content $entryFile
+
+# Upload per-extension entry to storage
+azcopy copy $entryFile $RegistryEntryBlobPath --overwrite=true
+if ($LASTEXITCODE -ne 0) {
+    Write-Error "Failed to upload entry to $RegistryEntryBlobPath (exit code $LASTEXITCODE)"
+    exit 1
+}
```
danieljurek: This script should only produce the relevant artifact file for later publishing when in the publishing job. This script should NOT upload the file.

danieljurek (dismissed):
A few things to change: 
1. Revert TODOs 
2. Do NOT upload engsys scripts to artifacts and try to run them in a release. Instead, produce the artifact during `GenerateReleaseArtifacts` and then download that artifact and publish it with inline powershell that invokes azcopy. We want to know what was published at any pipeline run (even if it's a JSON file) and producing that file at runtime instead of storing it as an artifact makes it difficult to obtain this knowledge.

eng/pipelines/templates/stages/build-and-test-azd-extension.yml
rajeshkamal5050: Done — reverted all 3 in 931336f.

eng/pipelines/release-ext-microsoft-azd-demo.yml
rajeshkamal5050: Reverted with the force-push earlier (49986eca6).

eng/pipelines/templates/stages/build-and-test-azd-extension.yml
rajeshkamal5050: Fixed in 931336f. Publish stages now do `checkout: self` and run the script from repo source — no executable scripts in artifacts anymore. Removed the script/template/helpers copies from release-metadata.

Couldn't move generation to GenerateReleaseArtifacts though — the script needs checksums from signed release artifacts which only exist after the Sign stage. Kept it in publish with checkout:self, matching how publish.yml does it for other jobs.

eng/pipelines/templates/steps/update-extension-daily-registry.yml
rajeshkamal5050: Deleted. Script is now called directly from the publish stage via checkout:self.

eng/pipelines/templates/steps/update-extension-pr-registry.yml
rajeshkamal5050: Deleted.

eng/scripts/Update-ExtensionDailyRegistry.ps1
rajeshkamal5050: Done. Script now takes `-OutputPath` and only writes the JSON file. Upload is a separate azcopy step in the pipeline YAML.

jongio (approved):
All issues from my two previous reviews are addressed. Per-extension registry entries cleanly solve the concurrent mutation problem. Version override is now wired into all three job types. The idempotency guard, powershell-yaml parsing, JSON escaping, and placeholder validation are solid additions.

Verified all 9 extensions have matching version.txt and extension.yaml, TODOs reverted, template relocated to eng/templates/ per Daniel's feedback. CI is green. Ship it.

wbreza (commented):
Code Review - Re-review of PR #7497 after refactor commit `931336f`

Great progress since my 2026-04-07 review. The `checkout: self` refactor cleanly separates script generation from artifact uploads, and the new `ConvertFrom-Yaml` path eliminates the fragile regex parsing.

### Findings Summary

| Priority | Count |
|----------|-------|
| High | 2 |
| Medium | 9 |
| Low | 8 |

### High Priority

H1 - Inline `pwsh` steps missing `$ErrorActionPreference = 'Stop'`
H2 - `PRNumber` not validated for numeric format

### Medium Priority

M1-M9: Quality/maintainability items listed in full review

### What Looks Good
- `checkout: self` refactor delivers security goal
- Per-extension registry pattern holds up well
- All prior HIGH items resolved

**Overall Assessment:** Comment state - materially better. The High items are small, mechanical fixes.

Review performed with GitHub Copilot CLI

vhvb1989 (approved):

Issue #7156, opened 2026-03-16
Create hosted agent toolsets


therealjohn:
Toolbox support for `azd ai agent`: enabling developers to connect Azure AI Foundry Toolboxes to their hosted agents across the full CLI lifecycle. Toolboxes expose curated tools (MCP servers, OpenAPI, first-party) through a unified MCP-compatible endpoint with platform-managed auth. This proposal adds a `--toolbox` flag to `run` and `init`, plus `azd ai agent toolbox list/show` for discovery - so developers can wire tools into their agent without leaving the `azd` workflow.

**Base URL: {project-endpoint}/toolsets API Version: v1 Required Header: Foundry-Features: Toolsets=V1Preview**

| Operation | Method | URL | Body |
 |-----------|--------|-----|------|
 | **List** | `GET` | `{project-endpoint}/toolsets?api-version=v1` | - |
 | **Get** | `GET` | `{project-endpoint}/toolsets/{name}?api-version=v1` | - |
 | **Create** | `POST` | `{project-endpoint}/toolsets?api-version=v1` | `{ "name": "{name}", "description": "{desc}", "metadata": {…}, "tools": [{…}] }` |
 | **Update** | `POST` | `{project-endpoint}/toolsets/{name}?api-version=v1` | `{ "description": "{desc}", "metadata": {…}, "tools": [{…}] }` |
 | **Delete** | `DELETE` | `{project-endpoint}/toolsets/{name}?api-version=v1` | - |

## Response shapes:

 - Single: { "object", "id", "created_at", "updated_at", "name", "description", "metadata", "tools" }
 - List: { "data": [ ...toolbox objects ] }
 - Delete: { "object", "name", "deleted": true }
 

 ## 1. Agent Manifest: Declare toolbox dependencies as `ToolResource`

 The [AgentSchema](https://github.com/microsoft/AgentSchema/tree/main/schemas/v1.0) already supports `ToolResource` with an open `options` map, and `FoundryConnection` for Foundry project auth. No schema changes needed.

 ```yaml
 # agent-manifest.yaml
 resources:
   - name: chat-model
     kind: model
     id: grok-4-fast-non-reasoning

   - name: github-tools
     kind: tool
     id: github-toolbox
     options:
       connection:
         kind: foundry
         endpoint: ${AZURE_AI_PROJECT_ENDPOINT}
         name: github-toolbox
         connectionType: toolbox
       allowedTools:
         - search_docs
         - create_issue
 ```

[...]

PR #7498, opened 2026-04-03
Prototype: toolbox and tool connection creation

therealjohn:
Fixes https://github.com/Azure/azure-dev/issues/7156

### DIFF
--- a/cli/azd/extensions/azure.ai.agents/internal/cmd/init.go
+++ b/cli/azd/extensions/azure.ai.agents/internal/cmd/init.go
@@ -413,6 +413,10 @@ func (a *InitAction) Run(ctx context.Context) error {
 			return fmt.Errorf("failed to process manifest parameters: %w", err)
 		}
 
+		// Inject toolbox MCP endpoint env vars into hosted agent definitions
+		// so agent.yaml is self-documenting about what env vars will be set.
+		injectToolboxEnvVarsIntoDefinition(agentManifest)
+
 		// Write the final agent.yaml to disk (after deployment names have been injected)
 		if err := writeAgentDefinitionFile(targetDir, agentManifest); err != nil {
 			return fmt.Errorf("writing agent definition: %w", err)
@@ -1145,12 +1149,24 @@ func (a *InitAction) addToProject(ctx context.Context, targetDir string, agentMa
 	agentConfig.Resources = resourceDetails
 
 	// Process toolbox resources from the manifest
-	toolboxes, toolConnections, err := extractToolboxAndConnectionConfigs(agentManifest)
+	toolboxes, toolConnections, credEnvVars, err := extractToolboxAndConnectionConfigs(agentManifest)
 	if err != nil {
 		return err
 	}
 	agentConfig.Toolboxes = toolboxes
 	agentConfig.ToolConnections = toolConnections
+
+	// Persist credential values as azd environment variables so they are
+	// resolved at provision/deploy time instead of stored in azure.yaml.
+	for envKey, envVal := range credEnvVars {
+		if _, setErr := a.azdClient.Environment().SetValue(ctx, &azdext.SetEnvRequest{
+			EnvName: a.environment.Name,
+			Key:     envKey,
+			Value:   envVal,
+		}); setErr != nil {
+			return fmt.Errorf("storing credential env var %s: %w", envKey, setErr)
+		}
+	}
 
 	// Detect startup command from the project source directory
 	startupCmd, err := resolveStartupCommandForInit(ctx, a.azdClient, a.projectConfig.Path, targetDir, a.flags.NoPrompt)
@@ -1569,13 +1585,16 @@ func downloadDirectoryContentsWithoutGhCli(
 // Toolbox resources with only Options["tools"] are treated as raw tool definitions (existing behavior).
 func extractToolboxAndConnectionConfigs(
 	manifest *agent_yaml.AgentManifest,
-) ([]project.Toolbox, []project.ToolConnection, error) {
+) ([]project.Toolbox, []project.ToolConnection, map[string]string, error) {
 	if manifest == nil || manifest.Resources == nil {
-		return nil, nil, nil
+		return nil, nil, nil, nil
 	}
 
 	var toolboxes []project.Toolbox
 	var connections []project.ToolConnection
+	// credentialEnvVars maps generated env var names to their raw values so
+	// the caller can persist them in the azd environment.
+	credentialEnvVars := map[string]string{}
 
 	for _, resource := range manifest.Resources {
 		tbResource, ok := resource.(agent_yaml.ToolboxResource)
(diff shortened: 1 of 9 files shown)


cli/azd/extensions/azure.ai.agents/internal/pkg/agents/agent_yaml/yaml.go
```diff
@@ -272,10 +279,239 @@ type ObjectProperty struct {
...
+		if p.Example != nil {
+			entry["example"] = *p.Example
+		}
+		props[p.Name] = entry
+	}
+
+	if len(props) > 0 {
+		// Merge property keys at the top level (record format)
+		for k, v := range props {
+			out[k] = v
```
jongio, line 510: Property names can collide with reserved keys here. If someone names a parameter `examples` or `strict`, it'll overwrite the structural fields already written to `out` - silent data corruption on round-trip. Either emit properties under a `properties` key in the map, or reject/skip names that collide with the reserved set.

cli/azd/extensions/azure.ai.agents/internal/pkg/agents/registry_api/helpers.go
```diff
@@ -480,6 +486,24 @@ func promptForTextValue(
 	return resp.Value, nil
 }
 
+// promptForSecretValue reads a secret value from the terminal with masked input.
+func promptForSecretValue(paramName string, required bool) (any, error) {
```
jongio, line 490: `noPrompt` isn't passed into this function, so `--no-prompt` mode will still call `term.ReadPassword()` and hang. The enum and text prompt paths both check `noPrompt` - this one should too. Add the `noPrompt bool` parameter and return an error early when it's set and the param is required.

cli/azd/extensions/azure.ai.agents/internal/cmd/listen.go
```diff
@@ -227,6 +233,73 @@ func resourcesEnvUpdate(ctx context.Context, resources []project.Resource, azdCl
...
+
+	// Already correctly wrapped
+	if _, hasKeys := creds["keys"]; hasKeys && len(creds) == 1 {
+		return creds
+	}
+
+	return map[string]any{"keys": creds}
+}
+
+func containerAgentHandling(ctx context.Context, azdClient *azdext.AzdClient, project *azdext.ProjectConfig, svc *azdext.ServiceConfig) error {
```
jongio, line 274: Two things here:

1. This function isn't called from anywhere - it's dead code. CI lint will likely flag it. If it's supposed to be wired into `envUpdate`, the call site is missing.

2. `os.ReadFile` errors are all swallowed with `return nil`. Only file-not-found should be ignored - permission errors, disk I/O errors, etc. should surface. Use `errors.Is(err, os.ErrNotExist)` to distinguish.

cli/azd/extensions/azure.ai.agents/internal/cmd/listen.go
```diff
@@ -227,6 +233,73 @@ func resourcesEnvUpdate(ctx context.Context, resources []project.Resource, azdCl
...
+	escapedJsonString := strings.ReplaceAll(jsonString, "\\", "\\\\")
+	escapedJsonString = strings.ReplaceAll(escapedJsonString, "\"", "\\\"")
+
+	return setEnvVar(ctx, azdClient, envName, "AI_PROJECT_CONNECTIONS", escapedJsonString)
+}
+
+// normalizeCredentials ensures credentials match the expected ARM format.
+// CustomKeys requires credentials nested under "keys": { "keys": { "key": "val" } }.
+// If already wrapped, returns as-is. Other auth types are returned unchanged.
+func normalizeCredentials(authType string, creds map[string]any) map[string]any {
```
jongio, line 261: `extractToolboxAndConnectionConfigs` already wraps CustomKeys credentials as `{"keys": creds}` at init time. Then `connectionsEnvUpdate` calls `normalizeCredentials` again at listen time. The guard `hasKeys && len(creds) == 1` works for the normal flow, but if someone manually edits azure.yaml to add a key alongside `keys`, it'll double-wrap. Consider normalizing in only one place to keep a single source of truth.

cli/azd/extensions/azure.ai.agents/internal/project/service_target_agent.go
```diff
@@ -1075,6 +1098,200 @@ func (p *AgentServiceTargetProvider) resolveEnvironmentVariables(value string, a
...
+	if errors.As(err, &respErr) && respErr.StatusCode == http.StatusConflict {
+		fmt.Fprintf(os.Stderr, "  Toolset '%s' already exists, updating...\n", toolbox.Name)
+		updateReq := &azure.UpdateToolsetRequest{
+			Description: toolbox.Description,
+			Tools:       toolbox.Tools,
+		}
+		toolset, updateErr := client.UpdateToolset(ctx, toolbox.Name, updateReq)
+		if updateErr != nil {
+			return nil, exterrors.Internal(
+				exterrors.CodeCreateToolsetFailed,
```
jongio, line 1253: This should be `CodeUpdateToolsetFailed` - the update path uses the create error code. Also, `exterrors.Internal` takes a string so the original error's type info is lost. Consider wrapping with `fmt.Errorf` and `%w` or adding an error cause parameter to `exterrors.Internal`.

cli/azd/extensions/azure.ai.agents/internal/cmd/init.go
```diff
@@ -1549,3 +1576,207 @@ func downloadDirectoryContentsWithoutGhCli(
...
+		})
+	}
+
+	return toolboxes, connections, credentialEnvVars, nil
+}
+
+// deriveConnectionName builds a deterministic connection name from a toolbox name
+// and tool definition. Uses the tool's Name field when present, otherwise falls
+// back to toolboxName-id.
+func deriveConnectionName(toolboxName string, toolDef agent_yaml.ToolboxToolDefinition) string {
```
jongio, line 1712: If multiple tools in the same toolbox have the same `Id` but no `Name`, this produces duplicate connection names (e.g., two `mcp` tools both become `toolbox-mcp`). The second connection would silently overwrite the first. Consider appending an index suffix when `Name` is empty, or validating uniqueness after extraction.

cli/azd/extensions/azure.ai.agents/internal/pkg/agents/agent_yaml/yaml.go
```diff
@@ -272,10 +279,239 @@ type ObjectProperty struct {
...
+		case "properties":
+			propertiesNode = val
+		default:
+			extraKeys = append(extraKeys, key)
+			extraValues = append(extraValues, val)
+		}
+	}
+
+	// If an explicit "properties" key was found, decode it (array or map).
+	if propertiesNode != nil {
```
jongio, line 348: When both a `properties` key AND extra top-level keys exist in the YAML, only `properties` is decoded and the extra keys are silently dropped. A user mixing formats gets no error and loses data. Consider returning an error about ambiguous format when `propertiesNode != nil && len(extraKeys) > 0`.

jongio (commented):
Early feedback on the toolbox/connection prototype. Test coverage is thorough and the init-deploy-listen lifecycle hangs together well. A few issues to address before this moves out of draft - most are small fixes, two are design-level.

PR #7501, opened 2026-04-03
fix: gracefully handle missing AKS cluster during postprovision hook

### LINKED ISSUE #3272: [Issue] host: aks fails on postprovision hook if iac doesn't include aks
I have a two phase provision:

phase 1
1. provision openai
2. run the app locally

phase 2:
1. provision the rest of the services, including aks

when host is aks, azd will do an auto-post-provision hook to do aks deployments or config.

Since phase 1 doesn't have aks provisioned, then this will fail with:

```
ERROR: deployment failed: failed invoking event handlers for 'postprovision', failed retrieving cluster admin credentials. Ensure your cluster has been configured to support admin credentials, parameter resourceName cannot be empty
```

I would like the post-provision hook to fail gracefully if the aks resources don't exist.

jongio:
## Summary

Fixes #3272

When infrastructure doesn't include AKS resources, the `postprovision` lifecycle hook in the AKS service target fails fatally trying to set up the Kubernetes context. This is expected in multi-phase provisioning workflows where AKS gets provisioned later.

## Changes

**`cli/azd/pkg/project/service_target_aks.go`**
- Modified `setK8sContext()` to detect the `postprovision` event and skip gracefully (with a user-visible warning) instead of failing **only** when the target resource has an empty name (via `SupportsDelayedProvisioning`) - the signal that the AKS cluster hasn't been provisioned yet in a multi-phase workflow
- All other errors (GetTargetResource, credentials/RBAC, namespace creation) propagate normally even during postprovision
- Extracted `skipPostprovisionK8sSetup()` helper with context cancellation propagation and structured telemetry (`aks.postprovision.skip` span with `skip.reason` attribute)
- The `predeploy` event path is **unchanged** and still fails fatally - deployment-time errors are not masked

**`cli/azd/pkg/project/service_target_aks_test.go`**
- Added `Test_Postprovision_GracefulSkip` verifying graceful skip when `ResourceName()` is empty
- Added `Test_Postprovision_Fails_When_GetTargetResource_Errors` - GetTargetResource error during postprovision propagates correctly
- Added `Test_Postprovision_Fails_When_Credentials_Fail` - credential failure during postprovision propagates correctly
- Added `Test_Postprovision_Fails_When_Namespace_Fails` - namespace creation error during postprovision propagates correctly
- Added `Test_Postprovision_Succeeds_When_Cluster_Available` - happy path verifying normal flow when cluster exists
- Added `Test_Postprovision_Propagates_Context_Cancellation` - context cancellation is propagated through skip path
- Added `Test_Predeploy_Still_Fails_When_Cluster_Not_Found` - regression guard ensuring predeploy still errors
- Added `Test_Predeploy_Fails_When_Credentials_Fail` and `Test_Predeploy_Fails_When_Namespace_Fails` - regression guards for downstream error propagation
- Extracted `createAksServiceTargetWithResourceManager` helper; `createAksServiceTarget` now delegates to it
- Extracted `assertSkipWarningEmitted` helper for DRY console output assertions

## Testing

All AKS-related tests pass (3 existing + 9 new), zero regressions.

### DIFF
--- a/cli/azd/pkg/project/service_target_aks_test.go
+++ b/cli/azd/pkg/project/service_target_aks_test.go
@@ -1083,72 +1083,50 @@ func (m *MockResourceManager) GetTargetResource(
 }
 
 func Test_Postprovision_GracefulSkip(t *testing.T) {
-	tests := []struct {
-		name             string
-		resourceName     string
-		resourceType     string
-		deleteClusterEnv bool
-	}{
-		{
-			name:             "SkipsWhenResourceNotProvisioned",
-			resourceName:     "",
-			resourceType:     "",
-			deleteClusterEnv: true,
+	tempDir := t.TempDir()
+	ostest.Chdir(t, tempDir)
+
+	mockContext := mocks.NewMockContext(t.Context())
+	err := setupMocksForAksTarget(mockContext)
+	require.NoError(t, err)
+
+	serviceConfig := createTestServiceConfig(
+		tempDir, AksTarget, ServiceLanguageTypeScript)
+	env := createEnv()
+	env.DotenvDelete(environment.AksClusterEnvVarName)
+
+	resourceManager := &MockResourceManager{}
+	targetResource := environment.NewTargetResource(
+		"SUBSCRIPTION_ID",
+		"RESOURCE_GROUP",
+		"",
+		"",
+	)
+	resourceManager.
+		On("GetTargetResource",
+			*mockContext.Context,
+			"SUBSCRIPTION_ID",
+			serviceConfig).
+		Return(targetResource, nil)
+
+	serviceTarget := createAksServiceTargetWithResourceManager(
+		mockContext, env, nil,
+		resourceManager)
+
+	err = serviceTarget.Initialize(
+		*mockContext.Context, serviceConfig)
+	require.NoError(t, err)
+
+	err = serviceConfig.Project.RaiseEvent(
+		*mockContext.Context,
+		postProvisionEvent,
+		ProjectLifecycleEventArgs{
+			Project: serviceConfig.Project,
 		},
-	}
+	)
+	require.NoError(t, err)
 
-	for _, tt := range tests {
-		t.Run(tt.name, func(t *testing.T) {
-			tempDir := t.TempDir()
-			ostest.Chdir(t, tempDir)
-
-			mockContext := mocks.NewMockContext(t.Context())
-			err := setupMocksForAksTarget(mockContext)
-			require.NoError(t, err)
-
-			serviceConfig := createTestServiceConfig(
-				tempDir, AksTarget, ServiceLanguageTypeScript)
-			env := createEnv()
-
-			if tt.deleteClusterEnv {
-				env.DotenvDelete(
-					environment.AksClusterEnvVarName)
-			}
-
-			resourceManager := &MockResourceManager{}
-			targetResource := environment.NewTargetResource(
-				"SUBSCRIPTION_ID",
-				"RESOURCE_GROUP",
-				tt.resourceName,
-				tt.resourceType,
-			)
-			resourceManager.
-				On("GetTargetResource",
-					*mockContext.Context,
-					"SUBSCRIPTION_ID",
-					serviceConfig).
-				Return(targetResource, nil)
-
-			serviceTarget := createAksServiceTargetWithResourceManager(
-				mockContext, env, nil,
-				resourceManager)
-
-			err = serviceTarget.Initialize(
-				*mockContext.Context, serviceConfig)
-			require.NoError(t, err)
-
-			err = serviceConfig.Project.RaiseEvent(
-				*mockContext.Context,
-				postProvisionEvent,
-				ProjectLifecycleEventArgs{
-					Project: serviceConfig.Project,
-				},
-			)
-			require.NoError(t, err)
-
-			assertSkipWarningEmitted(t, mockContext)
-		})
-	}
+	assertSkipWarningEmitted(t, mockContext)
 }
 
 func Test_Postprovision_Fails_When_GetTargetResource_Errors(t *testing.T) {


cli/azd/pkg/project/service_target_aks.go
```diff
@@ -906,31 +910,73 @@ func (t *aksTarget) setK8sContext(ctx context.Context, serviceConfig *ServiceCon
...
+) error {
+	log.Printf(
+		"skipping k8s context setup during postprovision: %s\n",
+		reason.Error())
+	t.console.Message(ctx, output.WithWarningFormat(
+		"AKS cluster not available yet, skipping Kubernetes "+
+			"context setup. It will be configured when the "+
+			"cluster is provisioned and a deployment is "+
+			"performed.\n"))
+	return nil
```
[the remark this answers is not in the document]
jongio: Good catch — fixed in 783293b. `skipPostprovisionK8sSetup` now checks `ctx.Err()` before returning nil, so context cancellation and deadline timeouts are always propagated.

cli/azd/pkg/project/service_target_aks.go
```diff
@@ -906,31 +910,73 @@ func (t *aksTarget) setK8sContext(ctx context.Context, serviceConfig *ServiceCon
...
+	reason error,
+) error {
+	log.Printf(
+		"skipping k8s context setup during postprovision: %s\n",
+		reason.Error())
+	t.console.Message(ctx, output.WithWarningFormat(
+		"AKS cluster not available yet, skipping Kubernetes "+
+			"context setup. It will be configured when the "+
+			"cluster is provisioned and a deployment is "+
+			"performed.\n"))
```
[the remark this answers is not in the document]
jongio: Intentional design choice — the generic message avoids leaking raw Azure API error details (HTTP response bodies, subscription IDs, endpoint URLs) to the end user. The detailed reason is always logged at the `log.Printf` level for debugging. The user-facing message consistently directs them to retry at deployment time, which is the correct action regardless of the specific failure cause.

cli/azd/pkg/project/service_target_aks.go
```diff
@@ -906,31 +910,73 @@ func (t *aksTarget) setK8sContext(ctx context.Context, serviceConfig *ServiceCon
...
+			return t.skipPostprovisionK8sSetup(ctx, err)
+		}
 		return err
 	}
 
 	err = t.ensureNamespace(ctx, defaultNamespace)
 	if err != nil {
+		if eventName == "postprovision" {
+			return t.skipPostprovisionK8sSetup(ctx, err)
+		}
```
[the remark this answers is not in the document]
jongio: Added `Test_Postprovision_Skips_When_Namespace_Fails` in 783293b. The test mocks `kubectl create namespace` to return an error, verifies that `ensureClusterContext` succeeds (credentials return HTTP 200) but postprovision still completes gracefully when namespace creation fails.

cli/azd/pkg/project/service_target_aks.go
```diff
@@ -906,31 +914,84 @@ func (t *aksTarget) setK8sContext(ctx context.Context, serviceConfig *ServiceCon
...
 	if err != nil {
-		return err
+		return t.postprovisionK8sError(ctx, eventName, err)
+	}
+
+	// When the AKS cluster hasn't been provisioned yet the target resource
+	// will have an empty resource name (via SupportsDelayedProvisioning).
+	// During postprovision this is expected in multi-phase provisioning
+	// workflows — skip gracefully instead of failing.
+	if targetResource == nil ||
```
wbreza: **[High]** `targetResource == nil` check bypasses event-name guard

The `targetResource == nil` branch calls `skipPostprovisionK8sSetup` unconditionally, while the empty-name branch correctly gates on `eventName == postProvisionEvent`. If `GetTargetResource` ever returns `(nil, nil)` (currently impossible but could happen via future refactor or mock), a **predeploy** call would silently skip instead of failing.

Consider routing the nil case through the event-aware `postprovisionK8sError` helper:

```suggestion
if targetResource == nil {
return t.postprovisionK8sError(ctx, eventName,
fmt.Errorf("AKS cluster target resource is nil"))
}

if targetResource.ResourceName() == "" && eventName == postProvisionEvent {
return t.skipPostprovisionK8sSetup(
ctx, fmt.Errorf("AKS cluster resource not yet provisioned"))
```

cli/azd/pkg/project/service_target_aks.go
```diff
@@ -906,31 +914,84 @@ func (t *aksTarget) setK8sContext(ctx context.Context, serviceConfig *ServiceCon
...
-		return err
+		return t.postprovisionK8sError(ctx, eventName, err)
 	}
 
-	// Display message to the user when we detect they are using a non-default KUBECONFIG configuration
-	// In standard AZD AKS deployment users should not typically need to set a custom KUBECONFIG
+	// Display message to the user when we detect they are using a
+	// non-default KUBECONFIG configuration. In standard AZD AKS deployment
+	// users should not typically need to set a custom KUBECONFIG.
 	if hasCustomKubeConfig && eventName == "predeploy" {
```
wbreza: **[Medium]** Inconsistent event name constants

The PR introduces `postProvisionEvent` to avoid magic string repetition, but `"predeploy"` remains a raw string literal here and in the handler registration (line 160). Consider defining a `preDeployEvent` constant for consistency with the PR's own rationale.

cli/azd/pkg/project/service_target_aks_test.go
```diff
@@ -1150,3 +1150,339 @@ func (m *mockUserConfigManager) Save(config config.Config) error {
...
+			require.NoError(t, err)
+
+			err = serviceConfig.Project.RaiseEvent(
+				*mockContext.Context,
+				postProvisionEvent,
+				ProjectLifecycleEventArgs{
+					Project: serviceConfig.Project,
+				},
+			)
+			require.NoError(t, err)
```
wbreza: **[Medium]** Tests assert no error but don't verify skip behavior

All graceful-skip tests only check `require.NoError`. If the warning message or skip path were accidentally removed, these tests would still pass. Consider asserting that the mock console received the expected warning message (e.g., verifying `mockContext.Console` was called with the skip warning text).

cli/azd/pkg/project/service_target_aks_test.go
```diff
@@ -1150,3 +1150,339 @@ func (m *mockUserConfigManager) Save(config config.Config) error {
...
+		*mockContext.Context,
+		postProvisionEvent,
+		ProjectLifecycleEventArgs{
+			Project: serviceConfig.Project,
+		},
+	)
+	require.NoError(t, err)
+}
+
+func Test_Predeploy_Still_Fails_When_Cluster_Not_Found(t *testing.T) {
```
wbreza, line 1273: **[Medium]** Missing predeploy failure tests for downstream errors

This regression guard only tests the cluster-name-resolution failure path. Consider adding tests for predeploy failing when `ensureClusterContext` or `ensureNamespace` return errors — this would validate that `postprovisionK8sError` correctly propagates errors for non-postprovision events across all error paths in `setK8sContext`.

cli/azd/pkg/project/service_target_aks_test.go
```diff
@@ -1150,3 +1150,339 @@ func (m *mockUserConfigManager) Save(config config.Config) error {
...
+		postProvisionEvent,
+		ProjectLifecycleEventArgs{
+			Project: serviceConfig.Project,
+		},
+	)
+	// The EventDispatcher aggregates handler errors via errors.New,
+	// which breaks the error chain. Assert via message content.
+	require.Error(t, err)
+	require.ErrorContains(t, err, "context canceled")
+}
```
wbreza, line 1520: **[Medium]** Test helper duplication

This helper largely duplicates the existing `createAksServiceTarget` (same kubectl, helm, kustomize, docker, dotnet, kubelogin setup). Consider refactoring so `createAksServiceTarget` delegates to this function with a default resource manager, keeping DRY.

cli/azd/pkg/project/service_target_aks.go
```diff
@@ -906,31 +914,84 @@ func (t *aksTarget) setK8sContext(ctx context.Context, serviceConfig *ServiceCon
...
+	reason error,
+) error {
+	// Propagate context cancellation — the user (or system) asked
+	// to stop; swallowing that would be incorrect.
+	if ctx.Err() != nil {
+		return ctx.Err()
+	}
+
+	log.Printf(
+		"skipping k8s context setup during postprovision: %v", reason)
```
wbreza, line 988: **[Low]** Consider structured telemetry for graceful fallback

The skip path uses `log.Printf` (debug) and `console.Message` (user warning), but no structured telemetry/tracing span is emitted. Adding a tracing event here would help monitor how frequently the multi-phase provisioning fallback is triggered in production.

cli/azd/pkg/project/service_target_aks.go
jongio: Good catch - fixed. The `targetResource == nil` case now routes through `postprovisionK8sError`, so predeploy will propagate the error instead of silently skipping. See commit f0d2c68.

cli/azd/pkg/project/service_target_aks.go
jongio: Done - added `preDeployEvent` constant using the same composition pattern (`"pre" + string(ServiceEventDeploy)`), replacing all 3 raw string occurrences. See commit f0d2c68.

cli/azd/pkg/project/service_target_aks_test.go
jongio: Good point - all graceful-skip tests now assert that the console received the "skipping Kubernetes context setup" warning message, so removing the skip path would cause test failures. See commit f0d2c68.

cli/azd/pkg/project/service_target_aks_test.go
jongio: Added `Test_Predeploy_Fails_When_Credentials_Fail` (401 from credential retrieval) and `Test_Predeploy_Fails_When_Namespace_Fails` (namespace creation denied). Both confirm predeploy propagates the error. See commit f0d2c68.

cli/azd/pkg/project/service_target_aks_test.go
jongio: Refactored - `createAksServiceTarget` now delegates to `createAksServiceTargetWithResourceManager` with a default empty-name resource manager, eliminating ~40 lines of duplication. See commit f0d2c68.

cli/azd/pkg/project/service_target_aks.go
jongio: Added an `AksPostprovisionSkipEvent` tracing span in `skipPostprovisionK8sSetup` using the existing `tracing.Start`/`EndWithStatus` pattern. The span records the skip reason so production monitoring can track how often the multi-phase fallback fires. See commit f0d2c68.

cli/azd/pkg/project/service_target_aks.go
```diff
@@ -906,28 +919,88 @@ func (t *aksTarget) setK8sContext(ctx context.Context, serviceConfig *ServiceCon
...
+	return err
+}
 
+// skipPostprovisionK8sSetup logs a warning and returns nil so that
+// postprovision continues even when Kubernetes context setup fails.
+// The context will be configured later when a deployment is performed.
+// Context cancellation/timeouts are always propagated.
+func (t *aksTarget) skipPostprovisionK8sSetup(
+	ctx context.Context,
+	reason error,
```
wbreza, line 975: **[Medium]** `span.EndWithStatus(reason)` marks intentional skips as errors in telemetry

`EndWithStatus` calls `span.SetStatus(codes.Error, ...)` when the error is non-nil — which `reason` always is here. Every graceful skip will appear as an error span in traces, potentially creating dashboard noise and false alerts. The span name (`aks.postprovision.skip`) already communicates that it's a skip.

Consider ending the span without error status:

```suggestion
span.End()
```

If the reason needs to be recorded, add it as a span attribute instead.

cli/azd/pkg/project/service_target_aks_test.go
```diff
@@ -1137,16 +1082,476 @@ func (m *MockResourceManager) GetTargetResource(
...
+					Project: serviceConfig.Project,
+				},
+			)
+			require.NoError(t, err)
+
+			// Verify the skip path actually fired by checking
+			// the user-facing warning was emitted.
+			consoleOutput := mockContext.Console.Output()
+			found := false
+			for _, msg := range consoleOutput {
```
wbreza: **[Low]** Console output assertion loop duplicated

This 9-line assertion block is copy-pasted in `Test_Postprovision_Skips_When_Namespace_Fails` as well. Consider extracting to a shared helper like `assertSkipWarningEmitted(t, mockContext)` for DRY.

cli/azd/pkg/project/service_target_aks_test.go
```diff
@@ -1137,16 +1082,476 @@ func (m *MockResourceManager) GetTargetResource(
...
+	)
+	require.Error(t, err)
 }
 
-func (m *mockUserConfigManager) Save(config config.Config) error {
-	args := m.Called(config)
-	return args.Error(0)
+func Test_Predeploy_Fails_When_Namespace_Fails(t *testing.T) {
+	tempDir := t.TempDir()
+	ostest.Chdir(t, tempDir)
```
wbreza, line 1351: **[Low]** Consider adding `ErrorContains` assertion

This test uses `require.Error` without verifying the error message content, unlike the sibling `Test_Predeploy_Fails_When_Namespace_Fails` which uses `require.ErrorContains(t, err, ""namespace creation denied"")`. Adding an error content check would make the test more precise.

cli/azd/pkg/project/service_target_aks.go
jongio: Fixed in 3c75aaa - using span.SetAttributes(attribute.String(\skip.reason\, ...)) + span.End() so graceful skips don't show as error spans in telemetry dashboards.

cli/azd/pkg/project/service_target_aks_test.go
jongio: Fixed in 3c75aaa - extracted assertSkipWarningEmitted(t, mockContext) helper used by both skip tests.

cli/azd/pkg/project/service_target_aks_test.go
jongio: Fixed in 3c75aaa - added require.ErrorContains(t, err, "failed retrieving cluster user credentials") for consistency with the namespace test.

wbreza (approved):
## Code Re-Review — PR #7501

**fix: gracefully handle missing AKS cluster during postprovision hook** by @jongio

### Re-review (3rd round)

Latest commit (`3c75aaa`) addresses all 3 findings from my Apr 7 re-review:

| 1 | Medium | Telemetry: `EndWithStatus()` marks skips as errors | ✅ Fixed — `span.End()` with `skip.reason` attribute |
| 2 | Low | DRY: duplicate console assertion blocks | ✅ Fixed — extracted `assertSkipWarningEmitted` |
| 3 | Low | Consistency: missing `ErrorContains` assertion | ✅ Fixed |

### Cumulative Review Status (all rounds)

| Round | Findings | Status |
|-------|----------|--------|
| Round 1 (6 findings) | nil-guard, event constants, console assertions, predeploy tests, helper DRY, telemetry | All ✅ |
| Round 2 (3 findings) | telemetry span, DRY assertions, ErrorContains | All ✅ |
| **Total: 9 findings** | | **All resolved** |

### ✅ What Looks Good
- Clean separation of postprovision (graceful skip) vs predeploy (fail fatally)
- Proper context cancellation handling with `ctx.Err()` check
- `postProvisionEvent` / `preDeployEvent` constants follow composition pattern
- 5 new tests covering all skip paths + regression guards
- `createAksServiceTarget` helper eliminates ~40 lines of test duplication

**Overall Assessment:** Approve — all 9 cumulative findings resolved across 3 review rounds.

cli/azd/pkg/project/service_target_aks.go
```diff
@@ -906,28 +920,89 @@ func (t *aksTarget) setK8sContext(ctx context.Context, serviceConfig *ServiceCon
...
+
+// postprovisionK8sError skips gracefully during postprovision events
+// and returns the error for all other events (e.g. predeploy).
+func (t *aksTarget) postprovisionK8sError(
+	ctx context.Context,
+	eventName ext.Event,
+	err error,
+) error {
+	if eventName == postProvisionEvent {
+		return t.skipPostprovisionK8sSetup(ctx, err)
```
weikanglim: This helper seems broader than the issue requires. The valid delayed-provisioning case is already modeled by returning an empty target resource when the AKS host does not exist yet, so skipping when ResourceName() == "" makes sense. But routing every postprovision error through this helper means azd provision now succeeds for real AKS setup failures too, including GetTargetResource errors, credential or RBAC failures, and namespace creation failures. I would keep the graceful path limited to explicit not-found or not-provisioned cases and let the other postprovision errors surface.

jongio: Good call - you're right that the resource manager already handles the delayed-provisioning case cleanly (`resolveServiceResource` returns an empty-name resource for `ResourceNotFoundError` + `SupportsDelayedProvisioning`), so `GetTargetResource` returning an actual error at line 925 means something else went wrong (auth, network, etc.) - not "not provisioned yet."

Narrowed the graceful skip to the explicit not-found case only:
- Removed `postprovisionK8sError` helper entirely
- `GetTargetResource`, `ensureClusterContext`, and `ensureNamespace` errors now propagate directly, even during postprovision
- Only `ResourceName() == ""` triggers the graceful skip (the valid delayed-provisioning signal)
- Updated tests: credential failures and namespace failures during postprovision now correctly surface as errors

Also fixed the gofmt import ordering that was failing CI.

cli/azd/pkg/project/service_target_aks.go
```diff
@@ -906,13 +920,30 @@ func (t *aksTarget) setK8sContext(ctx context.Context, serviceConfig *ServiceCon
...
-	targetResource, err := t.resourceManager.GetTargetResource(ctx, t.env.GetSubscriptionId(), serviceConfig)
+	targetResource, err := t.resourceManager.GetTargetResource(
+		ctx, t.env.GetSubscriptionId(), serviceConfig)
 	if err != nil {
 		return err
 	}
 
+	if targetResource == nil {
+		return fmt.Errorf("AKS cluster target resource is nil")
+	}
```
wbreza, line 931: **[:yellow_circle: Medium]** PR description is stale after weikanglim refactor

The PR description's Changes section still claims graceful handling for `GetTargetResource` errors, credential/RBAC failures, and namespace failures during postprovision. After the refactor, these now propagate as real errors — only `ResourceName() == ""` triggers the graceful skip.

Per AGENTS.md: "Keep the PR description in sync with the actual implementation."

Update the description to reflect the narrowed behavior:
> Modified `setK8sContext()` to skip gracefully during postprovision **only** when the target resource has an empty name (via `SupportsDelayedProvisioning`). All other errors (GetTargetResource, credentials, namespace) propagate normally even during postprovision.

cli/azd/pkg/project/service_target_aks_test.go
```diff
@@ -1137,16 +1082,541 @@ func (m *MockResourceManager) GetTargetResource(
 	return args.Get(0).(*environment.TargetResource), args.Error(1)
```
wbreza, line 1082: **[:green_circle: Low]** Table-driven test with single case

`Test_Postprovision_GracefulSkip` uses a table-driven structure but only has one case (`SkipsWhenResourceNotProvisioned`). The cases that originally lived here (credential failure, GetTargetResource error) were correctly moved to standalone tests that verify error propagation. The table wrapper now adds indirection without benefit — consider simplifying to a flat test or adding a comment explaining the table is kept for future cases.

cli/azd/pkg/project/service_target_aks.go
jongio: Good catch - updated the PR description to reflect the narrowed behavior after weikanglim's refactor. Only `ResourceName() == ""` triggers the graceful skip now; all other errors propagate.

cli/azd/pkg/project/service_target_aks_test.go
jongio: Flattened to a simple test - removed the table-driven wrapper since only one case remains after the error-propagation tests were correctly split out. See commit 82cc52c.

cli/azd/pkg/project/service_target_aks.go
wbreza, line 931: **[:green_circle: Low]** PR description test name & count mismatch

The PR description lists `Test_Postprovision_Skips_When_Namespace_Fails` but the actual test is `Test_Postprovision_Fails_When_Namespace_Fails` (renamed after weikanglim refactor). It also omits `Test_Postprovision_Fails_When_GetTargetResource_Errors`, `Test_Postprovision_Fails_When_Credentials_Fail`, and `Test_Postprovision_Propagates_Context_Cancellation`, and claims "3 existing + 8 new" when there are 9 new tests.

jongio: Good catch - updated the PR description: renamed `Test_Postprovision_Skips_When_Namespace_Fails` to `Test_Postprovision_Fails_When_Namespace_Fails`, added the missing tests (`Test_Postprovision_Fails_When_GetTargetResource_Errors`, `Test_Postprovision_Fails_When_Credentials_Fail`, `Test_Postprovision_Propagates_Context_Cancellation`), and corrected the count to "3 existing + 9 new".

jongio: Fixed

PR #7530, opened 2026-04-06
feat: Add AZD_DEPLOY_{SERVICE}_IGNORE_SLOTS env var to bypass slot deployment

### LINKED ISSUE #7365: [Issue] azd deploy automatically deploys to a deployment slot instead of the main app when `AZD_DEPLOY_{service}_SLOT_NAME` is not set

**Output from `azd version`**
azd version 1.23.13 (commit a1ccaff0905d6d981053089980968f5d88cf4c60)

**Describe the bug**
Based on the Microsoft DevBlogs about deployment slots with AZD, a user can set the environment variable `AZD_DEPLOY_{service}_SLOT_NAME` to target a specific deployment slot when running `azd deploy`.

This also implies that by **omitting this environment variable**, azd will deploy directly to the main app. However, when running `azd deploy` subsequently by omitting this environment variable, azd always deploys the service to the last deployed deployment slot and not the main app. This prevents us to have two pipelines, one deploying to the main app and another to the deployment slot.

**To Reproduce**
1. Set the `AZD_DEPLOY_{service}_SLOT_NAME` environment variable
2. run `azd deploy service_name --no-prompt`
3. Remove `AZD_DEPLOY_{service}_SLOT_NAME`
4. run `azd deploy service_name --no-prompt`

In the deployment logs, nothing is deployed to the main app and both deployments target the deployment slot.

**Expected behavior**
When `AZD_DEPLOY_{service}_SLOT_NAME` is not set, `azd deploy` targets the main app.

**Additional context**

I can see from this PR #6627 that the selection is done based on deployment history. I don't know about others but the last thing I would like from azd is to silently select a deployment slot for me based on an unknow criteria not explained anywhere. Especially not in a pipeline environment.

`production` is a reserved word in deployment slots too. So the simplest solution would be to always deploy to the main app when `AZD_DEPLOY_{service}_SLOT_NAME` is set to `production`.

vhvb1989:
## Summary

Fixes #7365

When deploying to an App Service that has deployment slots, `azd deploy` automatically routes deployments to slots instead of the main app. This prevents users from deploying directly to the main app in scenarios like dual-branch CI pipelines where one branch should deploy to production and another to a staging slot.

## Changes

Introduces a new per-service environment variable `AZD_DEPLOY_{SERVICE}_IGNORE_SLOTS` that, when set to a truthy boolean value (`true`, `1`, `TRUE`, etc.), bypasses all slot detection logic and deploys directly to the main app.

### Behavior

| Scenario | Without env var | With `IGNORE_SLOTS=true` |
|---|---|---|
| First deploy with slots | Main app + all slots | Main app only |
| Subsequent, 1 slot | Auto-picks slot | Main app |
| Subsequent, 2+ slots | Prompt / SLOT_NAME env var | Main app |

- Takes precedence over `AZD_DEPLOY_{SERVICE}_SLOT_NAME` (with a warning log)
- Follows the same `{SERVICE}` naming convention (uppercase, hyphens → underscores)
- Read via `env.Getenv()` (supports both azd `.env` file and system env vars)

### Files Changed

- `cli/azd/pkg/project/service_target_appservice.go` — Added `ignoreSlotsEnvVarNameForService()` helper and early-return check in `determineDeploymentTargets`
- `cli/azd/pkg/project/service_target_appservice_test.go` — 12 new table-driven tests covering all combinations of env var states, slot counts, and deployment history
- `cli/azd/docs/environment-variables.md` — Added new "App Service Slot Deployments" subsection documenting both `AZD_DEPLOY_{SERVICE}_IGNORE_SLOTS` and the previously-undocumented `AZD_DEPLOY_{SERVICE}_SLOT_NAME`

### Follow-up

A separate issue will be created to track updating the official docs (learn.microsoft.com) with slot deployment behavior and the new env var.

### DIFF
--- a/cli/azd/pkg/project/service_target_appservice.go
+++ b/cli/azd/pkg/project/service_target_appservice.go
@@ -230,6 +230,10 @@ type deploymentTarget struct {
 // based on deployment history and available slots.
 //
 // Deployment Strategy:
+//   - Override — AZD_DEPLOY_{SERVICE_NAME}_IGNORE_SLOTS (highest precedence):
+//     When set to a truthy boolean value, bypasses all slot detection and deploys directly
+//     to the main app. If AZD_DEPLOY_{SERVICE_NAME}_SLOT_NAME is also set, it is ignored
+//     and a warning is emitted.
 //   - First deployment (no history):
 //     Deploy to main app AND all slots to ensure consistency across all environments.
 //     This prevents configuration drift and ensures all slots start with the same baseline.
--- a/cli/azd/pkg/project/service_target_appservice_test.go
+++ b/cli/azd/pkg/project/service_target_appservice_test.go
@@ -115,10 +115,11 @@ func TestDetermineDeploymentTargets_IgnoreSlots(t *testing.T) {
 	t.Parallel()
 
 	tests := map[string]struct {
-		envVars        map[string]string
-		hasDeployments bool
-		slots          []string
-		expectedSlots  []string // empty string = main app
+		envVars         map[string]string
+		hasDeployments  bool
+		slots           []string
+		expectedSlots   []string // empty string = main app
+		expectedWarning string   // substring expected in console output, empty = no warning
 	}{
 		"IgnoreSlots_True_NoSlots": {
 			envVars:        map[string]string{"AZD_DEPLOY_API_IGNORE_SLOTS": "true"},
@@ -149,9 +150,10 @@ func TestDetermineDeploymentTargets_IgnoreSlots(t *testing.T) {
 				"AZD_DEPLOY_API_IGNORE_SLOTS": "true",
 				"AZD_DEPLOY_API_SLOT_NAME":    "staging",
 			},
-			hasDeployments: true,
-			slots:          []string{"staging", "preview"},
-			expectedSlots:  []string{""},
+			hasDeployments:  true,
+			slots:           []string{"staging", "preview"},
+			expectedSlots:   []string{""},
+			expectedWarning: "AZD_DEPLOY_API_SLOT_NAME is set but will be ignored",
 		},
 		"IgnoreSlots_False_OneSlot_SubsequentDeploy": {
 			envVars:        map[string]string{"AZD_DEPLOY_API_IGNORE_SLOTS": "false"},
@@ -190,10 +192,11 @@ func TestDetermineDeploymentTargets_IgnoreSlots(t *testing.T) {
 			expectedSlots:  []string{""},
 		},
 		"IgnoreSlots_InvalidValue_FallsBackToSlotLogic": {
-			envVars:        map[string]string{"AZD_DEPLOY_API_IGNORE_SLOTS": "notabool"},
-			hasDeployments: true,
-			slots:          []string{"staging"},
-			expectedSlots:  []string{"staging"},
+			envVars:         map[string]string{"AZD_DEPLOY_API_IGNORE_SLOTS": "notabool"},
+			hasDeployments:  true,
+			slots:           []string{"staging"},
+			expectedSlots:   []string{"staging"},
+			expectedWarning: "Ignoring invalid value",
 		},
 	}
 
@@ -275,6 +278,15 @@ func TestDetermineDeploymentTargets_IgnoreSlots(t *testing.T) {
 			require.Len(t, targets, len(tc.expectedSlots))
 			for i, expected := range tc.expectedSlots {
 				require.Equal(t, expected, targets[i].SlotName)
+			}
+
+			// Verify warning messages when expected
+			consoleOutput := strings.Join(mockContext.Console.Output(), "\n")
+			if tc.expectedWarning != "" {
+				require.Contains(t, consoleOutput, tc.expectedWarning)
+			} else {
+				require.Empty(t, mockContext.Console.Output(),
+					"unexpected warning in console output: %s", consoleOutput)
 			}
 		})
 	}


cli/azd/pkg/project/service_target_appservice.go
```diff
@@ -245,6 +247,33 @@ func (st *appServiceTarget) determineDeploymentTargets(
...
+						"%s is set but will be ignored because %s is enabled",
+						slotEnvVar, ignoreSlotsEnvVar),
+				})
+			}
+
+			progress.SetProgress(NewServiceProgress(
+				"Skipping slot deployment (deploying to main app)"))
+			return []deploymentTarget{{SlotName: ""}}, nil
+		}
+	}
```
wbreza, line 279: **[Architectural]** This new IGNORE_SLOTS check adds a third configuration layer to deployment slot routing. The underlying heuristic queries Azure-side deployment history that can be created by any tool — making the routing behavior non-deterministic from the user's perspective.

Consider an alternative: allow SLOT_NAME=production to target the main app (matching Azure Portal's terminology), and fail the deploy operation in --no-prompt mode when slots exist but no SLOT_NAME is specified. This makes every deployment explicit and eliminates the need for IGNORE_SLOTS entirely.

vhvb1989: `SLOT_NAME=production` is not good because main app is not a slot.

The new env var was added as a way to op-out from the `azd slots world` - The intention is well understood and self-documented. It is an escape path for anyone not happy with the slot-deployment principles from azd which encorages `never deploying to the main app when there is one slot` as priority.

So, I don't want to make `deploying to main app` part of the slot-deployment spec. That's why I am creating this option as opt-out, not as an update to the slot-deployment approach.

wbreza: Thanks for the thoughtful response, Victor. I appreciate the framing of `IGNORE_SLOTS` as an opt-out mechanism — that intent makes sense.

Where I'm still not aligned is on the implicit routing heuristic itself. A few things from digging into this:

**Deployment slots are a deliberate infrastructure decision.** They [require Standard tier or higher and must be explicitly created](https://learn.microsoft.com/en-us/azure/app-service/deploy-staging-slots#add-a-slot) — either through the Portal, CLI, or declared in Bicep/Terraform. azd never creates slots; it only discovers them. When a developer provisions a slot, they've already made a conscious architectural choice. Given that, I think azd should respect that intentionality by requiring explicit targeting rather than guessing based on deployment history — history that can be created by any tool, not just azd.

**Microsoft's own guidance for CI is explicit targeting.** The [deployment best practices docs](https://learn.microsoft.com/en-us/azure/app-service/deploy-best-practices#use-deployment-slots) state: *"Continuous deployment should never be enabled for your production slot. Instead, your production branch should be deployed onto a nonproduction slot."* This is a recommendation, not a hard rule — and crucially, it assumes the developer is making this choice explicitly, not having a tool make it for them.

**If a user no longer wants slot-based deployments, the answer is removing the slots** — either deleting them from the Portal or removing them from their Bicep/Terraform. That returns the App Service to a single-target model where azd deploys to the main app. An env var to "pretend slots don't exist" works around the infrastructure state rather than reflecting it.

**My core concern remains determinism in CI.** In `--no-prompt` mode, if slots exist but no `SLOT_NAME` is specified, I believe the deploy should fail with a clear error listing the available targets. This makes every CI deployment explicit and eliminates the entire class of "deployed to the wrong target" issues — without needing either the deployment-history heuristic or `IGNORE_SLOTS`.

Happy to jump on a call to work through this together.

cli/azd/pkg/project/service_target_appservice_test.go
```diff
@@ -53,3 +60,222 @@ func TestNewAppServiceTargetTypeValidation(t *testing.T) {
...
+
+			targets, err := target.determineDeploymentTargets(
+				*mockContext.Context,
+				serviceConfig,
+				targetResource,
+				progress,
+			)
+
+			require.NoError(t, err)
+			require.Len(t, targets, len(tc.expectedSlots))
```
jongio, line 278: **[MEDIUM]** Tests verify correct targets but don't assert warning messages.

For `IgnoreSlots_InvalidValue` and `IgnoreSlots_True_OverridesSlotName`, a regression that silently drops the `WarningMessage` would pass. Consider asserting on `mockContext.Console` output for those cases.

jongio (commented):
The doc comment for `determineDeploymentTargets` (above the diff) documents the full strategy matrix but doesn't mention the new IGNORE_SLOTS override, which is now the highest-precedence exit path. Worth adding a bullet so the doc stays in sync.

Clean implementation - good use of `LookupEnv`, `environment.Key()` DRY-up, and thorough test matrix.

cli/azd/pkg/project/service_target_appservice.go
vhvb1989: The routing heuristic was provided by the App Service team. I worked with Shilpi while they were adding the slot feature for Aspire, and that heuristic was a requirement from the service. I
  also worked with @puicchan to define the end-to-end user story here. The slot feature was designed with the path to production as its foundation.

  I'm hesitant to relax the feature for the sake of developer convenience, as it is specifically targeted at production environments. As you mentioned, you need a production-level SKU to even
  support slots.

  The expected behavior in CI, when azd discovers a project with slots, is to treat it as a production environment and follow production best practices as rules. This PR gives azd users the
  power to opt out without compromising a solid, production-ready slot-deployment experience.

puicchan:
Looks good to me.

wbreza:
## Revisiting the Design: `SLOT_NAME=production` vs. `IGNORE_SLOTS`

After reviewing the full discussion thread, the original issue (#7365), and the user's scenario in detail, I want to reopen the design conversation here — even though this PR has merged — because I believe we shipped a workaround rather than a solution, and we should track a follow-up to get this right.

### The platform's own model

In Azure App Service, `production` is **not just a label** — it is the reserved, canonical name for the main app slot. This is used consistently across the entire Azure surface area:

- **Azure CLI**: `az webapp deployment slot swap --target-slot production` ([docs](https://learn.microsoft.com/azure/app-service/deploy-staging-slots#swap-deployment-slots))
- **Azure PowerShell**: `Switch-AzWebAppSlot -DestinationSlotName "production"` ([docs](https://learn.microsoft.com/azure/app-service/deploy-staging-slots#swap-deployment-slots))
- **Azure REST API**: The slot swap API accepts `production` as a slot name parameter
- **ARM/Bicep templates**: Use `buildVersion` on the production slot (`Microsoft.Web/sites`) to trigger swaps ([docs](https://learn.microsoft.com/azure/app-service/deploy-staging-slots#automate-with-resource-manager-templates))
- **azd itself**: The `azd appservice swap` command uses `@main` as an alias for the production slot
- **MS Learn docs**: *"you can use a separate deployment slot instead of the **default production slot**"* ([Set up staging environments](https://learn.microsoft.com/azure/app-service/deploy-staging-slots))

The production slot isn't a conceptual shorthand — it's a first-class API-level entity. Every Azure tool that interacts with slots accepts `production` as a valid target.

### Why `IGNORE_SLOTS` is a workaround, not a feature

`AZD_DEPLOY_{SERVICE}_IGNORE_SLOTS` introduces a new concept that doesn't exist anywhere in Azure's deployment model. It tells azd to *"pretend the infrastructure you deliberately provisioned doesn't exist."* Consider the implications:

1. **It adds a third configuration layer** with precedence rules (`IGNORE_SLOTS` overrides `SLOT_NAME`, which overrides the deployment-history heuristic). This creates a combinatorial surface that's hard to reason about in CI.
2. **It's azd-specific vocabulary** — no other Azure tool has an "ignore slots" concept. A developer who knows App Service won't discover this without reading azd-specific docs.
3. **It works around infrastructure state instead of reflecting it** — if a user truly doesn't want slot-based deployments, the correct answer is removing the slots from their Bicep/Terraform, not adding an env var to pretend they aren't there.

### The simpler alternative

Allow `AZD_DEPLOY_{SERVICE}_SLOT_NAME=production` to deploy directly to the main app. This:

- **Uses the existing env var** — zero new API surface
- **Aligns with every other Azure tool** — CLI, PowerShell, REST API, ARM all accept `production` as the slot name for the main app
- **Eliminates precedence conflicts** — one env var, one behavior, no overrides
- **Is self-documenting** — `SLOT_NAME=production` clearly means "deploy to the production slot," which *is* the main app
- **Solves the user's exact scenario** from #7365: `dev/main` pipeline sets `SLOT_NAME=production`, `dev/replacement` pipeline sets `SLOT_NAME=staging` — done

Additionally, in `--no-prompt` mode with slots present and no `SLOT_NAME` set, the deploy should **fail with a clear error** listing available targets (including `production`). This makes CI deterministic without relying on deployment-history heuristics.

### Recommendation

I'd like to propose a follow-up PR that:
1. Accepts `production` as a valid value for `AZD_DEPLOY_{SERVICE}_SLOT_NAME` to target the main app
2. In `--no-prompt` mode, fails explicitly when slots exist but no `SLOT_NAME` is specified
3. Deprecates `AZD_DEPLOY_{SERVICE}_IGNORE_SLOTS` in favor of the above

This aligns azd with the platform it deploys to, rather than inventing parallel concepts that users must learn separately.

PR #7545, opened 2026-04-06
[Agents extensions] Endpoint changes

trangevi:
Migrate the rest of the vnext handling over to the new agents endpoints


jongio (commented):
Endpoint migration and PersistentPreRunE fix both look correct - clean change overall.

**monitor.go:96-102**: `info.Version` is validated unconditionally, but session logstream (line 152) no longer needs it. If version metadata isn't set, `monitor --session` fails with "agent version could not be resolved" even though the API call works without it. Consider deferring the version check to the container-logstream branch only (after the `sessionID == ""` check at line 161).

Minor: `resolveFilesContext` still sets `Version: info.Version` (~line 158) though no file operation reads it anymore. The help text (line 76) also still says "name, version, endpoint" - could drop "version". Small cleanup for a follow-up.

PR #7548, opened 2026-04-06
chore: cleanup golangci-lint and cspell findings

### LINKED ISSUE #7426: chore: cleanup golangci-lint and cspell findings
Address 14 golangci-lint findings and 19 cspell issues across 8 files without changing behavior.

weikanglim:
Fixes #7426.

This PR cleans up the remaining `golangci-lint` and `cspell` findings and includes one small correctness fix in `vsrpc` that fell out of the lint cleanup.

- Anchor local-template `.gitignore` reads to the template root so nested `.gitignore` files can't escape the local template tree during initialization.
- Update `internal/vsrpc/server.go` to derive per-call cancelable contexts from `childCtx`, which preserves the tracing span from `tracing.Start` for Call requests instead of dropping it.
- Add targeted `gosec` suppressions where the flagged values are already trusted, managed by azd, or owned by a later callback (local FastAPI detection, Darwin process inspection, spinner/server cancellation callbacks, and managed output paths).
- Promote repo-specific spellings into the shared cspell dictionaries, keep compatibility names as file-scoped overrides, and reword the remaining comment-only spellcheck hits.

### DIFF
--- a/cli/azd/.vscode/cspell-azd-dictionary.txt
+++ b/cli/azd/.vscode/cspell-azd-dictionary.txt
@@ -15,6 +15,8 @@ Errorf
 Frontends
 GOARCH
 GOCOVERDIR
+GOTOOLCHAIN
+GOWORK
 Ghostty
 LASTEXITCODE
 MCPJSON
@@ -110,6 +112,7 @@ containerapps
 containerd
 containerizable
 contoso
+covdata
 createdby
 csharpapp
 csharpapptest
@@ -153,6 +156,7 @@ go-imath
 godotenv
 gofmt
 golangci
+googleapis
 gosec
 goterm
 gotest
@@ -199,6 +203,7 @@ mysqladmin
 mysqlclient
 mysqldb
 nazd
+ndjson
 nobanner
 nodeapp
 nolint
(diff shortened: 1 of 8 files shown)


cli/azd/internal/vsrpc/server.go
```diff
@@ -213,9 +214,9 @@ func serveRpc(w http.ResponseWriter, r *http.Request, handlers map[string]Handle
 			call, isCall := req.(*jsonrpc2.Call)
 			if isCall {
 				span.SetAttributes(fields.JsonRpcId.String(fmt.Sprint(call.ID())))
-				//nolint:gosec // G118: cancel stored in cancelers map and called on completion
-				ctx, cancel := context.WithCancel(ctx)
-				childCtx = ctx
+				var cancel context.CancelFunc
+				//nolint:gosec // G118: cancel is stored in cancelers and invoked later by request cancellation handling.
+				childCtx, cancel = context.WithCancel(childCtx)
```
jongio, line 219: **[MEDIUM]** This isn't just a nolint relocation - the parent context changed from `ctx` to `childCtx`. The old code derived the cancel context from the outer `ctx` (no tracing span from L185), so handlers lost the span on Call requests. The new code derives from `childCtx`, preserving the span. Good fix - just worth calling out in the PR description since it's a behavior correction.

weikanglim: PR description has been updated

PR #7549, opened 2026-04-06
Fix principal tenant resolution and simplify subscription lookup

### LINKED ISSUE #7173: CurrentPrincipalId returns home-tenant oid for guest/B2B users, causing incorrect identity in resource tenant

weikanglim:
## Summary

Fix principal ID resolution for guest/B2B scenarios by resolving identities in the subscription's resource tenant, while preserving access-tenant behavior for credential acquisition flows.

### Resolve `AZURE_PRINCIPAL_ID` in the subscription tenant

**What's changing?**  
When azd determines `AZURE_PRINCIPAL_ID`, it now resolves that identity in the tenant that owns the subscription, instead of the tenant you originally signed in through. For guest/B2B users, this means azd uses the correct guest object ID for the subscription tenant.

**Why?**  
This should fix previous failures users may have encountered when azd created RBAC assignments with the wrong principal ID.

### Prefer the ARM token `oid` before calling Microsoft Graph

**What's changing?**  
azd now reads the principal ID from the ARM access token's `oid` claim first, and only calls Microsoft Graph if that claim is unavailable.

**Why?**  
This avoids unnecessary Graph access and improves performance, while keeping the existing fallback behavior for cases where the token does not contain a usable object ID.

## Internal changes

- replace the temporary tenant/context resolver abstractions with `account.SubscriptionResolver`, which returns `*account.Subscription`
- keep `auth token`, monitor, blob storage, and subscription credential flows on `Subscription.UserAccessTenantId`
- update tests and command guidance docs to match the new subscription resolver shape

Fixes #7173

### DIFF
--- a/cli/azd/pkg/azsdk/storage/storage_blob_client_test.go
+++ b/cli/azd/pkg/azsdk/storage/storage_blob_client_test.go
@@ -38,25 +38,24 @@ type mockSubscriptionResolver struct {
 
 var _ account.SubscriptionResolver = (*mockSubscriptionResolver)(nil)
 
-func (m *mockSubscriptionResolver) LookupTenant(
-	ctx context.Context, subscriptionId string) (string, error) {
-	args := m.Called(ctx, subscriptionId)
-	return args.String(0), args.Error(1)
-}
-
 func (m *mockSubscriptionResolver) GetSubscription(
 	ctx context.Context, subscriptionId string,
 ) (*account.Subscription, error) {
-	tenantId, err := m.LookupTenant(ctx, subscriptionId)
-	if err != nil {
-		return nil, err
+	args := m.Called(ctx, subscriptionId)
+
+	if subscription, ok := args.Get(0).(*account.Subscription); ok {
+		return subscription, args.Error(1)
 	}
 
-	return &account.Subscription{
-		Id:                 subscriptionId,
-		TenantId:           "resource-" + tenantId,
-		UserAccessTenantId: tenantId,
-	}, nil
+	if tenantId, ok := args.Get(0).(string); ok {
+		return &account.Subscription{
+			Id:                 subscriptionId,
+			TenantId:           "resource-" + tenantId,
+			UserAccessTenantId: tenantId,
+		}, args.Error(1)
+	}
+
+	return nil, args.Error(1)
 }
 
 // mockTokenCredential is a minimal mock implementation for testing
@@ -123,8 +122,8 @@ func Test_NewBlobSdkClient_UsesHomeTenantWhenNoSubscriptionId(t *testing.T) {
 	require.NotNil(t, client)
 	mockCredProvider.AssertExpectations(t)
 
-	// TenantResolver should NOT be called when no subscription ID is provided
-	mockTenantResolver.AssertNotCalled(t, "LookupTenant", mock.Anything, mock.Anything)
+	// GetSubscription should NOT be called when no subscription ID is provided.
+	mockTenantResolver.AssertNotCalled(t, "GetSubscription", mock.Anything, mock.Anything)
 }
 
 func Test_NewBlobSdkClient_ResolvesTenantWhenSubscriptionIdProvided(t *testing.T) {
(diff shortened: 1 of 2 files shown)


cli/azd/pkg/azsdk/storage/storage_blob_client_test.go
```diff
@@ -31,19 +31,34 @@ func (m *mockMultiTenantCredentialProvider) GetTokenCredential(
...
+func (m *mockSubscriptionResolver) LookupTenant(
 	ctx context.Context, subscriptionId string) (string, error) {
 	args := m.Called(ctx, subscriptionId)
 	return args.String(0), args.Error(1)
 }
 
+func (m *mockSubscriptionResolver) GetSubscription(
+	ctx context.Context, subscriptionId string,
+) (*account.Subscription, error) {
+	tenantId, err := m.LookupTenant(ctx, subscriptionId)
```
jongio: **[MEDIUM]** `GetSubscription` delegates to `LookupTenant`, which isn't part of `SubscriptionResolver`. All test expectations (`On("LookupTenant", ...)`) target this internal bridge method. If someone writes a new test using `On("GetSubscription", ...)`, testify won't match because `Called()` fires inside `LookupTenant`. Consider rewriting the mock to set expectations on `GetSubscription` directly.

cli/azd/pkg/azureutil/principal_test.go
```diff
@@ -0,0 +1,88 @@
...
+		&azcore.ClientOptions{
+			Transport: mockContext.HttpClient,
+		},
+		cloud.AzurePublic(),
+	)
+
+	principalId, err := GetCurrentPrincipalId(*mockContext.Context, userProfile, "resource-tenant")
+	require.NoError(t, err)
+	require.Equal(t, "graph-user-id", principalId)
+}
```
jongio, line 88: **[LOW]** Both tests cover the success paths (oid present, oid missing + Graph fallback). The `errors.Join` path in principal.go:36-39 - where both token oid extraction and Graph /me fail - isn't tested. Worth adding a test to verify the combined error message is useful for debugging.

weikanglim:
Thanks @vhvb1989 — I took a closer look at the extension impact.

I think this PR can stay core-only. The bug fixed here was in azd core's tenant selection for principal/resource-tenant operations. For extensions, the API surface already exposes both `TenantId` and `UserTenantId`, so this looks like a separate cleanup rather than something we need to scope into #7549.

I opened #7563 to track that follow-up. From the audit so far:
- generated extension templates in Go / JS / .NET should use `UserTenantId` (not `TenantId`) when they create user credentials from `PromptSubscription`
- `extensions/azure.ai.agents/internal/project/agent_identity_rbac.go` looks like a likely resource-tenant/RBAC bug candidate that should be reviewed separately

Most of the other extension call sites I spot-checked look like access-tenant credential acquisition paths and are likely correct.

PR #7576, opened 2026-04-08
feat: add validation to ensure manifest pointer is a file, not a directory

### LINKED ISSUE #7511: -m Flag Rejects Directory With No Helpful Suggestion
```
azd ai agent init -m /home/vibhadauria/workspace/hosted-agents-vnext-private-preview/samples/python/agentserver-responses/echo-agent-streaming

ERROR: reading local file ...: is a directory
```

### LINKED ISSUE #7575: Incorrect function error trying to init -m with a local directory path
```
v0.1.20-preview
Visit the docs at https://aka.ms/azd-ai-agent-docs

Let's get your project initialized.

Initializing an app to run on Azure (azd init)

  (✓) Done: Initialized git repository
  (✓) Done: Downloading template code to: C:\Users\johmil\projects\hosted-agents-tests\workspace\test


Installing required extensions...
  (-) Skipped: Installing azure.ai.agents extension (version 0.1.20-preview already installed)

Reading agent.yaml from local file: C:\Users\johmil\Projects\hosted-agents-tests\hosted-agents-vnext-private-preview\samples\python\agentserver-invocations\github-copilot

ERROR: reading local file C:\Users\johmil\Projects\hosted-agents-tests\hosted-agents-vnext-private-preview\samples\python\agentserver-invocations\github-copilot: read C:\Users\johmil\Projects\hosted-agents-tests\hosted-agents-vnext-private-preview\samples\python\agentserver-invocations\github-copilot: Incorrect function.

Suggestion: verify the file path exists and is readable
```

therealjohn:
Fixes #7511. Fixes #7575.

### DIFF
--- a/cli/azd/extensions/azure.ai.agents/internal/cmd/init.go
+++ b/cli/azd/extensions/azure.ai.agents/internal/cmd/init.go
@@ -470,10 +470,8 @@ func (a *InitAction) Run(ctx context.Context) error {
 		}
 
 		// Catch the common mistake of passing a directory instead of a file
-		if isValidFile {
-			if err := checkNotDirectory(a.flags.manifestPointer); err != nil {
-				return err
-			}
+		if err := checkNotDirectory(a.flags.manifestPointer); err != nil {
+			return err
 		}
 
 		// Download/read agent.yaml file from the provided URI or file path


cli/azd/extensions/azure.ai.agents/internal/cmd/init.go
```diff
@@ -463,6 +469,13 @@ func (a *InitAction) Run(ctx context.Context) error {
 			)
 		}
 
+		// Catch the common mistake of passing a directory instead of a file
+		if isValidFile {
+			if err := checkNotDirectory(a.flags.manifestPointer); err != nil {
+				return err
+			}
+		}
+
```
jongio, line 476: The `isValidFile` gate here isn't needed - `checkNotDirectory` already returns nil for non-directories and non-existent paths, so it's safe to call unconditionally. Removing the gate would simplify this and make it more resilient: on Linux, absolute directory paths like `/home/user/agents` pass `url.ParseRequestURI` (they're valid absolute paths), which sets `isValidURL=true` instead of `isValidFile`. Today that's fine because call site 1 in `newInitCommand` catches directories first, but if `InitAction.Run` ever gets called independently, this gate would cause the check to be silently skipped.

```suggestion
		// Catch the common mistake of passing a directory instead of a file
		if err := checkNotDirectory(a.flags.manifestPointer); err != nil {
			return err
		}

```

Issue #7541, opened 2026-04-06
Recurring `AADSTS700082: The refresh token has expired due to inactivity` after successful `azd auth login`


therealjohn:
There's a rare auth issue affecting small number of users where various azd commands result in an AADSTS700082 error:
<img width="2570" height="2087" alt="Image" src="https://github.com/user-attachments/assets/e4b592b1-393c-456e-aa05-d63e8ab43317" />

Running `azd auth login` to re-authenticate does not resolve the error, even if the exact `--tenant-id` was passed. Subsequent azd commands (e.g. azd deploy, azd init) continue to fail with the same stale token error, even though the user just logged in successfully.

JeffreyCA:
Is this the same issue as https://github.com/Azure/azure-dev/issues/7302?

JeffreyCA:
For anyone internal who's running into this issue, we added additional debug capabilities in v1.23.15 (to be released by EOD today) to help the team investigate the issue further. Please follow the instructions below and share the logs with us internally.

## Instructions

> [!IMPORTANT]
> **Do not** run `azd auth logout` or delete `~/.azd` first. We want to preserve the existing cached auth state.

1. Install the latest version of azd (at least v1.23.15).

2. Reproduce login with MSAL cache tracing enabled:
    ```
    AZD_DEBUG_MSAL_CACHE=true azd auth login --tenant-id <tenant-id> --debug
    ```
3. Immediately run the command that fails for you:
    ```
    AZD_DEBUG_MSAL_CACHE=true azd ai agent init --debug
    ```

4. Also collect these follow-up silent token traces:
    ```
    AZD_DEBUG_MSAL_CACHE=true azd auth token --debug
    AZD_DEBUG_MSAL_CACHE=true azd auth token --tenant-id <tenant-id> --debug
    ```

5. Save the terminal output from those commands, but do **not** share the raw token value returned by `azd auth token`. If needed, redact the `"token"` field or the final access token string before sending the logs.
 
Notes:
- `AZD_DEBUG_MSAL_CACHE=true` enables extra MSAL cache metadata tracing.
- `--debug` is required so those log lines are printed to the terminal.

rudiv:
This started happening here about 2 weeks ago and progressively getting worse - sometimes `azd auth login` works, sometimes doesn't.

Sometimes I'm fully authenticated and it still says it can't find the tenant rather than the token 90 day expiry notice.

[...]

PR #7578, opened 2026-04-08
Enhance MSAL cache logging and add tenant-specific re-authentication guidance

JeffreyCA:
This PR adds opt-in MSAL cache tracing to help investigate issue #7541 and improves tenant-specific guidance for expired refresh-token errors.

## What changed

- Added `AZD_DEBUG_MSAL_CACHE` to log MSAL cache metadata before/after login and around the first silent token acquisitions
- Improved `AADSTS70043` / `AADSTS700082` handling to return tenant-specific re-authentication guidance while preserving wrapped error message metadata
- Classified `AADSTS700082` as a re-login-required auth error
- Documented the new debug variable and added auth/account coverage for the new tracing and error-handling paths

## Why

We need better diagnostics for stale or unexpectedly changing MSAL refresh tokens without changing the core login flow, and we want subscription credential failures to point users to the correct tenant-specific remediation.

## User-facing instructions

> [!IMPORTANT]
> **Do not** run `azd auth logout` or delete `~/.azd` first. We want to preserve the existing cached auth state.

1. Install PR build from `<TBD>`

2. Reproduce login with MSAL cache tracing enabled:
    ```
    AZD_DEBUG_MSAL_CACHE=true azd auth login --tenant-id <tenant-id> --debug
    ```
3. Immediately run the command that fails for you:
    ```
    AZD_DEBUG_MSAL_CACHE=true azd ai agent init --debug
    ```

4. Also collect these follow-up silent token traces:
    ```
    AZD_DEBUG_MSAL_CACHE=true azd auth token --debug
    AZD_DEBUG_MSAL_CACHE=true azd auth token --tenant-id <tenant-id> --debug
    ```

5. Save the terminal output from those commands, but do **not** share the raw token value returned by `azd auth token`. If needed, redact the `"token"` field or the final access token string before sending the logs.
 
Notes:
- `AZD_DEBUG_MSAL_CACHE=true` enables extra MSAL cache metadata tracing.
- `--debug` is required so those log lines are printed to the terminal.

### DIFF
--- a/cli/azd/pkg/account/credentials_test.go
+++ b/cli/azd/pkg/account/credentials_test.go
@@ -141,8 +141,12 @@ func TestSubscriptionCredentialProvider_AADSTSErrors(t *testing.T) {
 
 	t.Run("AADSTS700082_WithExistingSuggestion_PreservesWrappedFields", func(t *testing.T) {
 		provider := NewSubscriptionCredentialProvider(
-			subscriptionTenantResolverFunc(func(ctx context.Context, subId string) (string, error) {
-				return tenantId, nil
+			subscriptionResolverFunc(func(ctx context.Context, subId string) (*Subscription, error) {
+				return &Subscription{
+					Id:                 subId,
+					TenantId:           "resource-" + tenantId,
+					UserAccessTenantId: tenantId,
+				}, nil
 			}),
 			multiTenantCredentialProviderFunc(func(ctx context.Context, tid string) (azcore.TokenCredential, error) {
 				return nil, &internal.ErrorWithSuggestion{
@@ -165,8 +169,12 @@ func TestSubscriptionCredentialProvider_AADSTSErrors(t *testing.T) {
 
 	t.Run("AADSTS700082_SuggestionAlreadyHasTenantID_NoRedundantAppend", func(t *testing.T) {
 		provider := NewSubscriptionCredentialProvider(
-			subscriptionTenantResolverFunc(func(ctx context.Context, subId string) (string, error) {
-				return tenantId, nil
+			subscriptionResolverFunc(func(ctx context.Context, subId string) (*Subscription, error) {
+				return &Subscription{
+					Id:                 subId,
+					TenantId:           "resource-" + tenantId,
+					UserAccessTenantId: tenantId,
+				}, nil
 			}),
 			multiTenantCredentialProviderFunc(func(ctx context.Context, tid string) (azcore.TokenCredential, error) {
 				return nil, &internal.ErrorWithSuggestion{
--- a/cli/azd/pkg/auth/cache_debug.go
+++ b/cli/azd/pkg/auth/cache_debug.go
@@ -72,11 +72,27 @@ func isMsalCacheTraceEnabled() bool {
 	return err == nil && enabled
 }
 
+const msalCachePIIWarning = "msal-cache: WARNING: MSAL cache tracing enabled — " +
+	"output contains account identifiers (e.g. email, home account ID). Do not share without redaction."
+
+func (t *msalCacheTracer) logPIIWarningOnce() {
+	const key = "_pii_warning"
+	t.mu.Lock()
+	defer t.mu.Unlock()
+
+	if t.logged[key] {
+		return
+	}
+	t.logged[key] = true
+	log.Println(msalCachePIIWarning)
+}
+
 func (t *msalCacheTracer) LogSnapshot(phase string) {
 	if t == nil || !t.enabled || phase == "" || t.cache == nil {
 		return
 	}
 
+	t.logPIIWarningOnce()
 	t.logSnapshot(phase)
 }
 
@@ -84,6 +100,8 @@ func (t *msalCacheTracer) LogSnapshotOnce(phase string) {
 	if t == nil || !t.enabled || phase == "" || t.cache == nil {
 		return
 	}
+
+	t.logPIIWarningOnce()
 
 	t.mu.Lock()
 	if t.logged[phase] {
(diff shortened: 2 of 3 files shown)


cli/azd/pkg/auth/cache_debug.go
```diff
@@ -0,0 +1,176 @@
...
+		account := contract.Account[key]
+		log.Printf(
+			"msal-cache[%s]: account key_sha256=%s home_account_id=%s realm=%s username=%s",
+			phase,
+			shortDigest(key),
+			account.HomeAccountID,
+			account.Realm,
+			account.Username,
+		)
+	}
```
jongio, line 180: **[MEDIUM]** This helper reimplements `slices.Sorted(maps.Keys(m))` which is already the established pattern across the codebase (cmd/config.go, pkg/infra/provisioning/bicep/bicep_provider.go, etc.). You can drop this function and use the one-liner at each call site - just add `maps` to the imports.

cli/azd/pkg/account/credentials_test.go
```diff
@@ -118,11 +119,36 @@ func TestSubscriptionCredentialProvider_AADSTSErrors(t *testing.T) {
...
+					Suggestion: "Run `azd auth login` to acquire a new token.",
+				}
+			}),
+		)
+
+		_, err := provider.CredentialForSubscription(t.Context(), subscriptionId)
+		assert.Error(t, err)
+
+		var errWithSuggestion *internal.ErrorWithSuggestion
+		assert.True(t, errors.As(err, &errWithSuggestion), "error should be wrapped in ErrorWithSuggestion")
```
jongio: **[LOW]** New test uses `errors.As` - prefer `errors.AsType[*internal.ErrorWithSuggestion](err)` per AGENTS.md. Production code in this package already uses `errors.AsType`. Not blocking since the existing tests in this file also use `errors.As`.

vhvb1989:
is there a reason why we could not use the same DEBUG flag we already have in azd?   or introcuce a debug-level config (info/error/etc). ??   
What's the reasoning about creatin a dedicated debug option?  is it b/c it is meant to be only temporal?

cli/azd/pkg/auth/cache_debug.go
```diff
@@ -0,0 +1,169 @@
...
+}
+
+// shortDigest returns the first 8 hex characters of the SHA-256 digest of s.
+func shortDigest(value string) string {
```
jongio, line 184: Nit: comment says `first 8 hex characters` but `sum[:6]` is 6 bytes, which `hex.EncodeToString` turns into 12 hex chars. Either `sum[:4]` (for 8 hex chars) or updating the comment to say `12` would fix the mismatch.

jongio (commented):
Solid PR - the nil-safe tracer design, thread-safe memoryCache, and tenant-specific re-auth guidance all look correct. One minor doc nit below.

JeffreyCA:
> is there a reason why we could not use the same DEBUG flag we already have in azd? or introcuce a debug-level config (info/error/etc). ?? What's the reasoning about creatin a dedicated debug option? is it b/c it is meant to be only temporal?

Yeah the debug flag is meant to be temporary until we can get more information and land a proper fix

JeffreyCA:
> ## Code Review — PR #7578
> **Enhance MSAL cache logging and add tenant-specific re-authentication guidance** by @JeffreyCA
> 
> ### Summary
> Clean, well-tested auth debugging infrastructure. The MSAL cache tracer correctly hashes token secrets via SHA-256, uses proper `strconv.ParseBool` for opt-in enablement, and the RWMutex addition on `memoryCache` cleanly resolves the prior race condition concern. AADSTS700082 classification as "login expired" is correct. All 5 prior review items resolved. One supplementary finding below.
> 
> ### Prior Review Status
> #	Prior Finding	Author	Status
> 1	Race condition in memoryCache	@copilot-bot	✅ Resolved (RWMutex)
> 2	sortedContractKeys reimplements `slices.Sorted(maps.Keys(m))`	@jongio	✅ Resolved
> 3	Hash comment mismatch (8 vs 12 hex chars)	@jongio	✅ Resolved
> 4	t.Context() / errors.AsType in tests	@copilot-bot + @jongio	✅ Resolved
> ### 🟡 Medium (1 finding)
> 1. **Debug log output should include a PII warning banner** — When `AZD_DEBUG_MSAL_CACHE=true`, the tracer logs `HomeAccountID` and `Username` in plain text. While intended for debugging and documented in environment-variables.md, the log output itself has no warning. Users piping to files or sharing sessions may not realize PII is included. Add a one-time startup warning: `"WARNING: MSAL cache tracing enabled — output contains account identifiers. Do not share without redaction."`
> 
> ### ✅ What Looks Good
> * Correct AADSTS700082 classification as "login expired" alongside AADSTS70043
> * `strconv.ParseBool` prevents accidental tracing enablement
> * SHA-256 hashing on cache keys and token values before logging
> * RWMutex on memoryCache cleanly resolves race condition
> * `LogSnapshotOnce` mutex-protected deduplication is correct
> * Tenant-specific error guidance preserves wrapped ErrorWithSuggestion fields
> * Comprehensive table-driven tests for error classification and tracer lifecycle
> * All 5 prior review items fully resolved
> 
> Priority	Count
> Medium	1
> **Total**	**1**
> **Overall Assessment:** Comment — clean implementation with no bugs found. The PII warning finding is defense-in-depth.
> 
> _Review performed with GitHub Copilot CLI_

Fixed some conflicts after https://github.com/Azure/azure-dev/pull/7549 got merged. Added a PII warning as suggested.

jongio (approved):
All previous findings resolved. Code is clean - RWMutex on memoryCache correctly prevents races, nil-safe tracer design with SHA-256 hashing, tenant-ID deduplication in error wrapping works correctly, AADSTS700082 classification is correct. Tests are thorough. CI is green. No new issues found.

PR #7629, opened 2026-04-10
feat: add lightspeed commands to skip update check for fast exit

### LINKED ISSUE #7360: [Issue] Can't complete azd deploy on ai agent (get_foundry_project error)

**Output from `azd version`**

azd version 1.24.0-beta.1-pr.6069275 (commit 2ef20fb1f0811263ed634c3f786126f8ab3d7fe9) (stable)

**Describe the bug**

I ran "azd up" on seattle-hotel-agent, then closed my laptop during deployment, came back to see the deployment had finished in the portal successfully. then I ran azd up, and it errors, as shown below:
```
seattle-hotel-agent % azd up

Packaging services (azd package)

  (✓) Done: Packaging service seattle-hotel-agent
  - No artifacts were found


Subscription: python (8077d4f9-d1b7-4c80-b6c2-cf9a2ec2d604)
Location: West US

  (-) Skipped: Didn't find new changes.

  (x) Failed: Deploying service seattle-hotel-agent

ERROR: error executing step command 'deploy --all': getting target resource: resolving target resource via external service target: get_foundry_project: AzureDeveloperCLICredential: signal: killed
```

**Additional context**

vhvb1989:
## Description

Fixes #7360

Extensions using `AzureDeveloperCLICredential` spawn `azd auth token` as a subprocess with a **10-second timeout** (hardcoded in the Go Azure SDK). The background update check goroutine in `main.go` had a **60-second timeout** and was blocking process exit via `<-latest`, causing `signal: killed` errors when the update check was slow (e.g., after laptop wake, DNS stalls, corporate proxy delays).

### Root Cause

When `azd auth token` runs as a subprocess:
1. `main.go:65` — `fetchLatestVersion` goroutine launches an HTTP call to check for updates
2. `main.go:76` — The token command runs and writes output in **milliseconds**
3. `main.go:86` — `<-latest` **blocks** process exit until the update check finishes
4. After 10 seconds, the credential kills the process → `signal: killed`

### Solution

Add a `Lightspeed` property to `ActionDescriptorOptions` that marks commands requiring fast exit. The update check goroutine is now started **inside `ExecuteWithAutoInstall` after command identification** — lightspeed commands skip it entirely, so the process exits immediately after producing output.

No duplicate command lists: the lightspeed flag is defined once in the command descriptor and automatically propagated via cobra annotations.

### Changes

| File | Change |
|------|--------|
| `cmd/actions/action_descriptor.go` | Added `Lightspeed bool` field |
| `cmd/cobra_builder.go` | Propagates to `cmd.Annotations["lightspeed"]` |
| `cmd/auth.go` | Marked `auth token` as `Lightspeed: true` |
| `cmd/auto_install.go` | New `ExecuteResult` struct; `startUpdateCheck()` helper; goroutine starts only for non-lightspeed commands |
| `main.go` | Removed `fetchLatestVersion`; consumes `ExecuteResult.LatestVersion` channel |

### Testing

- **`Test_LightspeedAnnotation`** — Verifies annotation propagation from descriptor to cobra command
- **`TestExecuteWithAutoInstall_LightspeedDetection`** — Verifies `auth token` is detected as lightspeed, `version` is not
- **`TestExecuteWithAutoInstall_ReturnsCommandErrorWithoutPanicForOutputFlags`** — Updated for new return type
- All existing snapshot/auth/integration tests pass

### DIFF
--- a/cli/azd/cmd/auto_install.go
+++ b/cli/azd/cmd/auto_install.go
@@ -391,7 +391,7 @@ func tryAutoInstallExtension(
 // startUpdateCheck launches a background goroutine that checks for a newer
 // version of azd and returns a channel that will receive the result.
 // The caller should read from the returned channel after command execution.
-func startUpdateCheck() <-chan *update.VersionInfo {
+func startUpdateCheck(ctx context.Context) <-chan *update.VersionInfo {
 	ch := make(chan *update.VersionInfo, 1)
 
 	// Allow the user to skip the update check by setting AZD_SKIP_UPDATE_CHECK.
@@ -400,10 +400,13 @@ func startUpdateCheck() <-chan *update.VersionInfo {
 			log.Print("skipping update check since AZD_SKIP_UPDATE_CHECK is true")
 			close(ch)
 			return ch
+		} else if err != nil {
+			log.Printf("could not parse value for AZD_SKIP_UPDATE_CHECK as a boolean "+
+				"(it was: %s), proceeding with update check", value)
 		}
 	}
 
-	bgCtx, bgCancel := context.WithTimeout(context.Background(), 60*time.Second)
+	bgCtx, bgCancel := context.WithTimeout(ctx, 60*time.Second)
 
 	go func() {
 		defer close(ch)
@@ -483,7 +486,7 @@ func ExecuteWithAutoInstall(ctx context.Context, rootContainer *ioc.NestedContai
 		// AzureDeveloperCLICredential 10-second subprocess timeout from being
 		// hit when the update check is slow (laptop wake, DNS stalls, etc.).
 		if !result.IsLightspeed {
-			result.LatestVersion = startUpdateCheck()
+			result.LatestVersion = startUpdateCheck(ctx)
 		}
 
 		// Check for partial namespace match (e.g., "ai" found but "ai.agent" not installed)
@@ -573,7 +576,7 @@ func ExecuteWithAutoInstall(ctx context.Context, rootContainer *ioc.NestedContai
 	}
 
 	// Unknown command path — always start the update check since these aren't lightspeed.
-	result.LatestVersion = startUpdateCheck()
+	result.LatestVersion = startUpdateCheck(ctx)
 
 	// Extract flags that take values from the root command
 	flagsWithValues := extractFlagsWithValues(rootCmd)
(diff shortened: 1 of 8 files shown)


cli/azd/cmd/auto_install.go
```diff
@@ -385,14 +389,79 @@ func tryAutoInstallExtension(
...
+func startUpdateCheck(ctx context.Context) <-chan *update.VersionInfo {
+	ch := make(chan *update.VersionInfo, 1)
+
+	// Allow the user to skip the update check by setting AZD_SKIP_UPDATE_CHECK.
+	if value, has := os.LookupEnv("AZD_SKIP_UPDATE_CHECK"); has {
+		if setting, err := strconv.ParseBool(value); err == nil && setting {
+			log.Print("skipping update check since AZD_SKIP_UPDATE_CHECK is true")
+			close(ch)
+			return ch
+		} else if err != nil {
```
jongio, line 404: The original `fetchLatestVersion` logged a diagnostic when `AZD_SKIP_UPDATE_CHECK` couldn't be parsed as a bool:

```go
} else if err != nil {
    log.Printf("could not parse value for AZD_SKIP_UPDATE_CHECK as a boolean "+
        "(it was: %s), proceeding with update check", value)
}
```

This helps debug situations where a user sets the var to something unexpected. Worth preserving.

jongio (commented):
The core approach is solid - moving the update check after command identification and skipping it for lightspeed commands directly fixes the timeout issue from #7360.

Things to address:

- `fetchLatestVersion()` in main.go (lines 153-188) is now dead code. The call was removed but the function definition wasn't. `docs/design/azd-update.md` also references it in two places (lines 54, 107). Both should be cleaned up.

- cspell-lint CI is failing (17 hits). Add "lightspeed" to `.vscode/cspell.yaml`.

JeffreyCA (approved):

cli/azd/cmd/auto_install.go
vhvb1989: Already implemented — the parse-error diagnostic log is preserved at lines 404–407:

```go
} else if err != nil {
    log.Printf("could not parse value for AZD_SKIP_UPDATE_CHECK as a boolean "+
        "(it was: %s), proceeding with update check", value)
}
```

This was addressed in commit `caed3b2` ("fix: address PR review feedback").

kristenwomack (approved):

PR #7630, opened 2026-04-10
fix: explicit slot targeting — SLOT_NAME=production for main app

### LINKED ISSUE #7365: [Issue] azd deploy automatically deploys to a deployment slot instead of the main app when `AZD_DEPLOY_{service}_SLOT_NAME` is not set

rajeshkamal5050:
## Fixes #7365

Replaces the auto-routing heuristic (deployment history check, 1-slot auto-pick) with explicit slot targeting.

### What changed

**Deploy (`service_target_appservice.go`)**
- `SLOT_NAME=production` → deploys to main app (case-insensitive)
- `SLOT_NAME=staging` → deploys to that slot (validated against actual slots)
- `SLOT_NAME=bogus` → error with available slots listed + production hint
- No SLOT_NAME + `--no-prompt` + slots exist → fails with error listing targets
- No SLOT_NAME + interactive + slots exist → prompt with "production (main app)" + slot names
- No slots → main app (no prompt needed)
- Removed deployment history check (`HasAppServiceDeployments`)
- Removed 1-slot auto-pick

**Swap (`swap.go`)**
- `--src production` / `--dst production` → main app (aligns with deploy)
- `@production` also accepted
- `@main` still works but prints deprecation warning
- Updated help text and prompt labels

**Docs**
- Added `AZD_DEPLOY_{SERVICE}_SLOT_NAME` to `environment-variables.md`

### Why "production" is safe

Azure ARM API rejects creating a slot named "production":
```
$ az webapp deployment slot create --slot production
slot is invalid. Slot name: 'Production' is reserved.
```
Not just portal UX — backend enforces it. Zero ambiguity.

### Breaking changes

1. **1-slot auto-pick removed**: CI pipelines relying on auto-pick (slots exist, no SLOT_NAME set, `--no-prompt`) will now get an error. Fix: set `AZD_DEPLOY_{SERVICE}_SLOT_NAME=<target>`.

2. **First-deploy "push to all slots" removed**: Previously, first deploy (no deployment history) would deploy to the main app AND all slots simultaneously. Now you always deploy to exactly one target — set `SLOT_NAME` for each target.

### Manual testing

Tested against live App Service with staging slot (`rg-azd-slot-test-69352`):

| Scenario | Result |
|----------|--------|
| `SLOT_NAME=production` | ✅ Deployed to main app |
| `SLOT_NAME=staging` | ✅ Deployed to staging slot |
| `SLOT_NAME=bogus` | ✅ Error: slot not found, lists available + production hint |
| No SLOT_NAME + `--no-prompt` | ✅ Error: no target specified, lists options |
| Interactive (no SLOT_NAME) | ✅ Prompt with "production (main app)" + "staging" |

### Unit tests

12 table-driven tests covering all scenarios above + case insensitivity + edge cases.

Ref: azure-dev-pr/discussions/1775 (Option 4)


cli/azd/extensions/azure.appservice/internal/cmd/swap.go
```diff
@@ -358,8 +358,16 @@ func runSwap(ctx context.Context, flags *swapFlags, rootFlags rootFlagsDefinitio
...
+	// "production" and "@production" both map to the main app (empty string).
+	// This aligns with the Azure platform convention where "production" is the
+	// reserved name for the main app slot.
+	if strings.EqualFold(slot, "production") || strings.EqualFold(slot, "@production") {
+		return ""
+	}
+
+	// "@main" is deprecated — use "production" instead.
 	if strings.EqualFold(slot, "@main") {
+		fmt.Println("WARNING: @main is deprecated. Use 'production' instead to refer to the main app.")
```
jongio: This deprecation warning has a few issues:
- Writes to stdout instead of stderr - can interfere with piped output
- `normalizeSlotName` is a normalization function - side effects make it harder to reason about and test
- If both `--src @main --dst @main`, this prints twice
- Uses `fmt.Println` but the rest of the file uses `color.Yellow()` for warnings

Consider moving the warning to `runSwap` (check `flags.src`/`flags.dst` for `@main` before calling `normalizeSlotName`) and use `color.Yellow()` for consistency.

jongio (commented):
Clean design improvement - the deployment-history heuristic was opaque and the 1-slot auto-pick was surprising. Explicit slot targeting is much more predictable.

Core logic in `service_target_appservice.go` is well-structured: SLOT_NAME first, then no-slots fast path, then no-prompt error, then interactive prompt. Good test coverage (12 cases, all branches).

Things to address:

1. `normalizeSlotName` in swap.go puts a deprecation warning (via `fmt.Println`) inside what should be a pure normalization function. The warning goes to stdout (not stderr), duplicates if both flags are `@main`, and uses plain fmt instead of `color.Yellow()` like the rest of the file.

2. `HasAppServiceDeployments` in `azapi/webapp.go` is now dead code - this PR removes its only call site. Consider removing the function and its test, or note it for a follow-up.

3. The "Breaking change" section documents the 1-slot auto-pick removal but doesn't mention the first-deployment behavior change: previously, first deploy would push to main app + ALL slots simultaneously. That behavior is also gone now - worth documenting.

rajeshkamal5050:
Updated the PR description — breaking changes section now documents both:
1. 1-slot auto-pick removal
2. First-deploy "push to all slots" removal

@jongio — did you also mean we should add this to the changelog / release notes?

cli/azd/pkg/project/service_target_appservice.go
```diff
@@ -271,64 +296,45 @@ func (st *appServiceTarget) determineDeploymentTargets(
...
-			availableSlots[i] = slot.Name
+			availableTargets = append(availableTargets, slot.Name)
 		}
 		return nil, fmt.Errorf(
-			"slot '%s' specified in %s not found. Available slots: [%s]. "+
-				"Please update the environment variable with a valid slot name",
-			slotName, slotEnvVarName, strings.Join(availableSlots, ", "))
+			"deployment slots detected but no target specified. "+
+				"Set %s to one of: [%s]",
+			slotEnvVarName, strings.Join(availableTargets, ", "))
```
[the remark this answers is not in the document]
rajeshkamal5050: Fixed — error now shows copy-pasteable values: `[production, staging]` with `('production' = main app)` hint.

cli/azd/extensions/azure.appservice/internal/cmd/swap.go
rajeshkamal5050: Fixed — moved warning to runSwap, uses color.Yellow(), prints once even if both flags are @main. Also removed dead HasAppServiceDeployments.

jongio (approved):
Clean design - explicit targeting via SLOT_NAME is much more predictable than the old deployment-history heuristic. The "production = main app" convention aligns with how Azure's ARM API treats the reserved name, so there's zero ambiguity.

Logic is solid: SLOT_NAME takes precedence, then no-slots fast path, then interactive/no-prompt fork. Case-insensitive matching with canonical name return is a nice improvement over the old exact-match. Good test coverage across all branches.

A few minor observations (none blocking):

- The `Deploy()` method still loops over `deployTargets`, but `determineDeploymentTargets` now always returns exactly one target. The loop and `hasSlots` logic still work correctly - just dead complexity you could simplify later.

- The case-insensitive slot matching (`strings.EqualFold` replacing `==`) is an improvement worth calling out in the PR description since it's a subtle behavioral change. Could also add a test case like `SLOT_NAME=Staging` matching a `staging` slot to lock this in.

- Dead code removal, functional test update, and re-recorded test data all look clean.

wbreza (approved):
### Design Alignment

Verified against the [design proposal on PR #7530](https://github.com/Azure/azure-dev/pull/7530#issuecomment-4202322081). All three requirements are correctly implemented:

- ✅ `SLOT_NAME=production` → deploys to main app (case-insensitive, aligns with Azure ARM convention)
- ✅ `--no-prompt` + slots + no SLOT_NAME → fails with actionable error listing targets including `production`
- ✅ Single env var, no `IGNORE_SLOTS` — zero new API surface, no precedence conflicts

### What Looks Good
- Clean replacement of deployment-history heuristic with deterministic explicit targeting
- Excellent table-driven test coverage (12 cases, all branches)
- Dead code (`HasAppServiceDeployments`) fully removed with zero remaining callers
- Swap command aligned with the `production` convention

### Findings

| Priority | Count |
|----------|-------|
| Medium   | 1     |
| Low      | 2     |

**[Medium] Case-sensitive slot validation in swap vs case-insensitive in deploy** (`swap.go:isValidSlotName`)
`isValidSlotName` uses `slices.Contains` (case-sensitive), but the deploy path uses `strings.EqualFold`. A user who learns `SLOT_NAME=Staging` works for deploy will find `--src Staging` fails for swap. Pre-existing gap — not introduced by this PR — worth aligning in a follow-up with `slices.ContainsFunc` + `strings.EqualFold`.

**[Low] `@production` accepted but undocumented** (`swap.go:normalizeSlotName`)
`normalizeSlotName` accepts `@production` → `""` but the flag help text only mentions `production`. Minor discoverability gap.

**[Low] Deploy loop iterates over always-single-element slice** (`service_target_appservice.go:Deploy`)
`determineDeploymentTargets` now always returns exactly one target, but `Deploy()` retains the multi-target loop. Dead complexity, already noted by @jongio.

rajeshkamal5050:
Follow-up items tracked in #7650

PR #7633, opened 2026-04-10
Persist first deployment name in AZURE_AI_MODEL_DEPLOYMENT_NAME

### LINKED ISSUE #7632: AZURE_AI_MODEL_DEPLOYMENT_NAME is only set when using existing code
It's not set if you provide a manifest with `-m` or if you run `init` in an empty directory. 

It should always be set if you defined a model resource and go through the flow or provide the flag.

therealjohn:
Fixes #7632. 

The env var was only set in the from-code path). The from-manifest path (ProcessModels) collected deployment details but never persisted the env var. This affected users who init via --manifest, templates, or --model/--model-deployment flags on those paths.

Fix: Added a setEnvValue call in ProcessModels() after deployment details are collected, using the first deployment's name. This covers all manifest/template init paths since they all flow through ProcessModel

### DIFF
--- a/cli/azd/extensions/azure.ai.agents/internal/cmd/init_models.go
+++ b/cli/azd/extensions/azure.ai.agents/internal/cmd/init_models.go
@@ -791,18 +791,32 @@ func (a *InitAction) ProcessModels(ctx context.Context, manifest *agent_yaml.Age
 		return nil, nil, fmt.Errorf("failed to inject deployment names into manifest: %w", err)
 	}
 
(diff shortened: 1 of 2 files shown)


jongio (approved):
Clean fix. Mirrors the existing setEnvValue pattern from init_from_code.go, proper guard check on empty deploymentDetails, and correct error wrapping. Nothing to flag.

JeffreyCA (approved):

JeffreyCA (approved):

PR #7636, opened 2026-04-10
Add positional argument support to agent init

### LINKED ISSUE #7510: ERROR: unknown command when passing positional argument to `azd ai agent init`
```
azd ai agent init /home/vibhadauria/workspace/hosted-agents-vnext-private-preview/samples/python/agentserver-responses/echo-agent-streaming

ERROR: unknown command "/home/.../echo-agent-streaming" for "agent init"
```

therealjohn:
Fixes #7510

azd ai agent init <path> now accepts an optional positional argument that is auto-disambiguated between url, manifest file, or directory. 

Explicit flags (--manifest, --src) still take priority. Passing both a positional arg and the corresponding flag produces a clear validation error with a suggestion.

 Before

   ❯ azd ai agent init C:\path\to\my-agent
   ERROR: unknown command "C:\path\to\my-agent" for "agent init"

  After

   ❯ azd ai agent init C:\path\to\my-agent
   # Equivalent to: azd ai agent init --src C:\path\to\my-agent

### DIFF
--- a/cli/azd/extensions/azure.ai.agents/internal/cmd/init.go
+++ b/cli/azd/extensions/azure.ai.agents/internal/cmd/init.go
@@ -208,11 +208,18 @@ func newInitCommand(rootFlags *rootFlagsDefinition) *cobra.Command {
 	}
 
 	cmd := &cobra.Command{
-		Use:   "init [-m <manifest pointer>] [--src <source directory>]",
+		Use:   "init [<path>] [-m <manifest pointer>] [--src <source directory>]",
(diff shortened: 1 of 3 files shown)


jongio (approved):
Clean implementation. URL detection via `://` handles all schemes nicely, and the heuristic for non-existent paths (yaml extension = manifest, other = src directory) is a smart UX choice.

One minor note: `CodeInvalidPositionalArg` in `codes.go` is defined but never referenced - the updated `resolvePositionalArg` never returns an error. Not blocking, just a cleanup opportunity.

trangevi (approved):
I don't really like that it can be either manifest or source dir, but not going to block it.

PR #7640, opened 2026-04-10
[Agents extension] Toolbox support

### LINKED ISSUE #7156: Create hosted agent toolsets

trangevi:
Adding the ability to create a toolbox as a part of agent creation.


cli/azd/extensions/azure.ai.agents/internal/pkg/azure/foundry_toolsets_client.go
```diff
@@ -0,0 +1,214 @@
...
+func NewFoundryToolboxClient(
+	endpoint string,
+	cred azcore.TokenCredential,
+) *FoundryToolboxClient {
+	userAgent := fmt.Sprintf("azd-ext-azure-ai-agents/%s", version.Version)
+
+	clientOptions := &policy.ClientOptions{
+		Logging: policy.LogOptions{
+			AllowedHeaders: []string{azsdk.MsCorrelationIdHeader, "X-Request-Id"},
+			IncludeBody:    true,
```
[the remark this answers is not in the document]
trangevi: We don't pass credentials to the toolbox handling, credentials are handled via bicep for connections creation, so this doesn't apply.

cli/azd/extensions/azure.ai.agents/internal/cmd/listen.go
```diff
@@ -316,3 +468,320 @@ func populateContainerSettings(ctx context.Context, azdClient *azdext.AzdClient,
...
+	envName string,
+	projectEndpoint string,
+	toolboxName string,
+) error {
+	envKey := toolboxMCPEndpointEnvKey(toolboxName)
+
+	endpoint := strings.TrimRight(projectEndpoint, "/")
+	mcpEndpoint := fmt.Sprintf(
+		"%s/toolboxes/%s/mcp?api-version=v1", endpoint, toolboxName,
+	)
```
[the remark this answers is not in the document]
trangevi: Toolbox names shouldn't contain spaces, will update tests

cli/azd/extensions/azure.ai.agents/internal/pkg/azure/foundry_toolsets_client.go
```diff
@@ -0,0 +1,214 @@
...
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/policy"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/runtime"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/streaming"
+	"github.com/azure/azure-dev/cli/azd/pkg/azsdk"
+)
+
+const (
+	toolboxesApiVersion    = "v1"
+	toolboxesFeatureHeader = "Toolsets=V1Preview"
```
trangevi: Change to Toolboxes=V1Preview

jongio (commented):
Solid foundation for toolbox support. A few items beyond what the bot flagged - mostly around a breaking change in tool kind names, some credential handling gaps, and a couple of data-loss edges.

cli/azd/extensions/azure.ai.agents/internal/pkg/agents/agent_yaml/yaml.go
```diff
@@ -31,21 +36,66 @@ func ValidAgentKinds() []AgentKind {
...
 
 type ToolKind string
 
 const (
 	ToolKindFunction        ToolKind = "function"
 	ToolKindCustom          ToolKind = "custom"
-	ToolKindWebSearch       ToolKind = "webSearch"
-	ToolKindBingGrounding   ToolKind = "bingGrounding"
-	ToolKindFileSearch      ToolKind = "fileSearch"
+	ToolKindWebSearch       ToolKind = "web_search"
```
jongio, line 50: The tool kind values changed from camelCase to snake_case (`webSearch` -> `web_search`, `bingGrounding` -> `bing_grounding`, etc.). Existing `agent.yaml` files using the old names will hit `unrecognized tool kind` in `ExtractToolsDefinitions`. Consider supporting both forms during parsing with a normalization step, or flag this as a breaking migration in the PR description.

cli/azd/extensions/azure.ai.agents/internal/cmd/listen.go
```diff
@@ -227,6 +273,112 @@ func resourcesEnvUpdate(ctx context.Context, resources []project.Resource, azdCl
...
+	escaped := strings.ReplaceAll(jsonString, "\\", "\\\\")
+	escaped = strings.ReplaceAll(escaped, "\"", "\\\"")
+
+	return setEnvVar(ctx, azdClient, envName, key, escaped)
+}
+
+// normalizeCredentials ensures credentials match the expected ARM format.
+// CustomKeys requires credentials nested under "keys": { "keys": { "key": "val" } }.
+// If already wrapped, returns as-is. Other auth types are returned unchanged.
+func normalizeCredentials(authType string, creds map[string]any) map[string]any {
```
jongio: `extractToolboxAndConnectionConfigs` in init.go already wraps CustomKeys credentials under `keys` (line ~1775). Then `toolConnectionsEnvUpdate` calls `normalizeCredentials` which wraps again unless `len(creds) == 1`. If a CustomKeys connection has `keys` plus other metadata fields, this produces double-wrapping. Pick one canonical wrapping site - either init.go does it and this function is a no-op for pre-wrapped data, or this function is the sole authority and init.go passes raw creds.

cli/azd/extensions/azure.ai.agents/internal/cmd/init.go
```diff
@@ -1645,3 +1680,222 @@ func downloadDirectoryContentsWithoutGhCli(
...
+			connections = append(connections, conn)
+
+			// Toolbox tool entry is minimal — deploy enriches from connection
+			tool := map[string]any{
+				"type":                  toolType,
+				"project_connection_id": connName,
+			}
+			tools = append(tools, tool)
+		}
+
```
jongio, line 2301: For tools with a `target`, only `type` and `project_connection_id` survive. Other tool-specific fields (`name`, custom options, etc.) are dropped. The comment says "deploy enriches from connection" but `enrichToolboxFromConnections` only adds `server_url`/`server_label`. If tools carry additional config (e.g. headers, query parameters, auth scopes), those would be silently lost.

cli/azd/extensions/azure.ai.agents/internal/cmd/init.go
```diff
@@ -1645,3 +1680,222 @@ func downloadDirectoryContentsWithoutGhCli(
...
+			UseWorkspaceManagedIdentity: connResource.UseWorkspaceManagedIdentity,
+			Error:                       connResource.Error,
+		}
+
+		// Surface credentials.type to top-level authType when not explicitly set.
+		// The API expects authType at the connection level, not nested in credentials.
+		if conn.AuthType == "" {
+			if credType, ok := conn.Credentials["type"].(string); ok && credType != "" {
+				conn.AuthType = credType
+				delete(conn.Credentials, "type")
```
jongio: `conn.Credentials` is assigned from `connResource.Credentials` (shared map reference at line 1877). This `delete` mutates the manifest's original map. If the manifest is referenced again after this call (e.g. written to disk, logged), the `type` field will be silently missing. Clone before mutating:

```go
creds := maps.Clone(connResource.Credentials)
conn.Credentials = creds
```

cli/azd/extensions/azure.ai.agents/internal/cmd/init.go
```diff
@@ -1645,3 +1680,222 @@ func downloadDirectoryContentsWithoutGhCli(
...
+	}
+
+	existingNames := make(map[string]bool, len(*containerAgent.EnvironmentVariables))
+	for _, ev := range *containerAgent.EnvironmentVariables {
+		existingNames[ev.Name] = true
+	}
+
+	for _, tbName := range toolboxNames {
+		envKey := toolboxMCPEndpointEnvKey(tbName)
+		if !existingNames[envKey] {
```
jongio: `existingNames` isn't updated after appending. If two toolboxes normalize to the same env key (e.g. "my-tools" and "my tools" both become `TOOLBOX_MY_TOOLS_MCP_ENDPOINT`), duplicates get injected. Add `existingNames[envKey] = true` after the append.

cli/azd/extensions/azure.ai.agents/internal/cmd/listen.go
```diff
@@ -316,3 +468,325 @@ func populateContainerSettings(ctx context.Context, azdClient *azdext.AzdClient,
...
+		return result
+	}
+
+	var entries []struct {
+		Name string `json:"name"`
+		ID   string `json:"id"`
+	}
+	if err := json.Unmarshal([]byte(jsonStr), &entries); err != nil {
+		fmt.Fprintf(os.Stderr,
+			"Warning: failed to parse AI_PROJECT_CONNECTION_IDS_JSON: %s\n", err)
```
jongio: If the JSON is invalid, this logs a warning and returns an empty map. Downstream, every `project_connection_id` will silently fail to resolve to an ARM ID, causing an opaque API error from Foundry. Returning an error here would surface the real problem immediately.

cli/azd/extensions/azure.ai.agents/internal/cmd/listen.go
```diff
@@ -316,3 +468,325 @@ func populateContainerSettings(ctx context.Context, azdClient *azdext.AzdClient,
...
+func resolveTemplateRef(s string) string {
+	if strings.HasPrefix(s, "{{") && strings.HasSuffix(s, "}}") {
+		return strings.TrimSpace(s[2 : len(s)-2])
+	}
+	return s
+}
+
+// resolveEnvValue resolves ${VAR} references in a string using envsubst.
+func resolveEnvValue(value string, azdEnv map[string]string) string {
+	resolved, err := envsubst.Eval(value, func(varName string) string {
```
jongio, line 781: `envsubst.Eval` errors are swallowed and the raw unresolved string passes through to the Foundry API. A `fmt.Fprintf(os.Stderr, ...)` warning here would save debugging time when an env var reference is misspelled or missing.

cli/azd/extensions/azure.ai.agents/internal/cmd/init.go
```diff
@@ -1645,3 +1680,222 @@ func downloadDirectoryContentsWithoutGhCli(
...
+		if !ok {
+			continue
+		}
+
+		conn := project.Connection{
+			Name:                        connResource.Name,
+			Category:                    string(connResource.Category),
+			Target:                      connResource.Target,
+			AuthType:                    string(connResource.AuthType),
+			Credentials:                 connResource.Credentials,
```
jongio: Connection credentials from `ConnectionResource` are copied directly into `project.Connection`, which gets serialized to `azure.yaml` in plaintext. Compare with toolbox tool credentials (line ~1770) which are externalized to env vars via `credentialEnvVars` and replaced with variable references. Consider the same externalization pattern for connection credentials to avoid secrets in config files.

cli/azd/extensions/azure.ai.agents/internal/cmd/listen.go
```diff
@@ -227,6 +273,100 @@ func resourcesEnvUpdate(ctx context.Context, resources []project.Resource, azdCl
...
+	key string,
+	value any,
+) error {
+	data, err := json.Marshal(value)
+	if err != nil {
+		return fmt.Errorf("failed to marshal %s to JSON: %w", key, err)
+	}
+
+	jsonString := string(data)
+	escaped := strings.ReplaceAll(jsonString, "\\", "\\\\")
```
jongio, line 418: `json.Marshal` already produces valid JSON with escaped quotes and backslashes. The manual `strings.ReplaceAll` here double-escapes the output, so consumers reading `AI_PROJECT_TOOL_CONNECTIONS` or `AI_AGENT_DEPLOYMENTS_JSON` from the env get `{\"name\":\"foo\"}` instead of parseable JSON.

If the env store needs shell-safe values, the escaping should happen inside `setEnvVar` (and be reversed on read), not here. Otherwise just pass `string(data)` directly.

cli/azd/extensions/azure.ai.agents/internal/pkg/agents/agent_yaml/parse.go
```diff
@@ -196,6 +208,8 @@ func ExtractToolsDefinitions(template map[string]any) ([]any, error) {
 				return nil, fmt.Errorf("failed to unmarshal to Tool: %w", err)
 			}
 
+			toolDef.Kind = NormalizeToolKind(toolDef.Kind)
```
jongio, line 211: `ExtractToolsDefinitions` now normalizes tool kinds here, but `ProcessPromptAgentToolsConnections` (line ~474) still switches on raw `toolDef.Kind` without calling `NormalizeToolKind`. A prompt agent manifest with `kind: webSearch` will fail in `ProcessPromptAgentToolsConnections` while working fine through `ExtractToolsDefinitions`.

Add `toolDef.Kind = NormalizeToolKind(toolDef.Kind)` before the switch in `ProcessPromptAgentToolsConnections` to keep both paths consistent.

cli/azd/extensions/azure.ai.agents/internal/cmd/init.go
```diff
@@ -1645,3 +1694,268 @@ func downloadDirectoryContentsWithoutGhCli(
...
+			}
+
+			// External tools with target/authType need a connection
+			toolName, _ := toolMap["name"].(string)
+			authType, _ := toolMap["authType"].(string)
+			credentials, _ := toolMap["credentials"].(map[string]any)
+
+			connName := toolName
+			if connName == "" {
+				connName = tbResource.Name + "-" + toolType
```
jongio: Two unnamed tools of the same type in one toolbox produce identical connection names (`tbResource.Name + "-" + toolType`). The second tool's credentials overwrite the first, and both end up referencing the same `project_connection_id`.

Consider appending an index: `connName = fmt.Sprintf("%s-%s-%d", tbResource.Name, toolType, i)`

cli/azd/extensions/azure.ai.agents/internal/pkg/agents/agent_yaml/yaml.go
```diff
@@ -31,21 +36,94 @@ func ValidAgentKinds() []AgentKind {
...
-	ToolKindCodeInterpreter ToolKind = "codeInterpreter"
+	ToolKindOpenApi         ToolKind = "openapi"
+	ToolKindCodeInterpreter ToolKind = "code_interpreter"
+	ToolKindAzureAiSearch   ToolKind = "azure_ai_search"
+	ToolKindA2APreview      ToolKind = "a2a_preview"
+)
+
+// legacyToolKindAliases maps deprecated camelCase tool kind names to their
+// current snake_case equivalents so that older agent.yaml files continue to parse.
+var legacyToolKindAliases = map[ToolKind]ToolKind{
```
jongio, line 62: This map is missing `"openApi": ToolKindOpenApi`. Existing manifests using `kind: openApi` will hit "unrecognized tool kind" after the camelCase-to-snake_case migration.

cli/azd/extensions/azure.ai.agents/internal/cmd/listen.go
```diff
@@ -316,3 +456,328 @@ func populateContainerSettings(ctx context.Context, azdClient *azdext.AzdClient,
...
+		endpoint, toolboxName, toolboxVersion,
+	)
+
+	return setEnvVar(
+		ctx, azdClient, envName, envKey, mcpEndpoint,
+	)
+}
+
+// toolboxMCPEndpointEnvKey builds the TOOLBOX_{NAME}_MCP_ENDPOINT env var key.
+func toolboxMCPEndpointEnvKey(toolboxName string) string {
```
jongio, line 684: This only strips spaces and hyphens, but `credentialEnvVarName` in init.go uses `nonAlphanumRe` to strip all non-alphanumeric chars. A toolbox name like `my.tools/v2` produces an invalid env key here. Consider using the same regex approach for consistency.

jongio (commented):
Five additional findings from deeper analysis - focused on runtime bugs and backward-compat gaps not covered by my previous comments.

jongio (commented):
Follow-up review - 3 new findings beyond my previous inline comments.

CI is still red: cspell (9 unknown words), gofmt (3 files not formatted), gosec (2 G101 false positives in test files need nolint), go-fix (manual for-range loops should use maps.Copy). These need to be fixed before merge.

Many of my earlier comments from April 11-13 still appear unaddressed - please take a pass through those as well.

cli/azd/extensions/azure.ai.agents/internal/pkg/agents/agent_yaml/yaml.go
```diff
@@ -279,10 +358,239 @@ type ObjectProperty struct {
...
+		}
+		if p.Example != nil {
+			entry["example"] = *p.Example
+		}
+		props[p.Name] = entry
+	}
+
+	if len(props) > 0 {
+		// Merge property keys at the top level (record format)
+		for k, v := range props {
```
jongio, line 595: If a manifest parameter is named `examples` or `strict`, this loop overwrites the struct-level values set earlier in `out`. For example, a parameter named `examples` would replace `out["examples"]` (the actual Examples list) with the parameter's entry object.

Consider checking for reserved keys before merging:

```go
reservedKeys := map[string]bool{"examples": true, "strict": true, "properties": true}
for k, v := range props {
    if reservedKeys[k] {
        return nil, fmt.Errorf("parameter name %q conflicts with reserved PropertySchema key", k)
    }
    out[k] = v
}
```

cli/azd/extensions/azure.ai.agents/internal/pkg/azure/foundry_toolsets_client.go
```diff
@@ -0,0 +1,214 @@
...
+	}
+
+	resp, err := c.pipeline.Do(req)
+	if err != nil {
+		return nil, fmt.Errorf("HTTP request failed: %w", err)
+	}
+	defer resp.Body.Close()
+
+	if !runtime.HasStatusCode(resp, http.StatusOK) {
+		return nil, runtime.NewResponseError(resp)
```
jongio, line 132: POST for resource creation typically returns 201 Created. If the Foundry API returns 201 for new toolbox versions, this would incorrectly return an error. Consider:

```suggestion
    if !runtime.HasStatusCode(resp, http.StatusOK, http.StatusCreated) {
```

cli/azd/extensions/azure.ai.agents/internal/cmd/listen.go
```diff
@@ -365,3 +505,328 @@ func populateContainerSettings(ctx context.Context, azdClient *azdext.AzdClient,
...
+	toolbox.Description = resolveEnvValue(toolbox.Description, azdEnv)
+	for i, tool := range toolbox.Tools {
+		toolbox.Tools[i] = resolveMapValues(tool, azdEnv)
+	}
+}
+
+// enrichToolboxFromConnections fills in server_url and server_label on toolbox
+// tools that reference a connection via project_connection_id. This keeps the
+// azure.yaml toolbox entries minimal while sending complete data to the API.
+func enrichToolboxFromConnections(
```
jongio, line 702: When a tool references a `project_connection_id` that isn't found in `connByName`, no warning is emitted. The tool gets sent to the Foundry API without `server_url`/`server_label`, which could produce incomplete toolbox configurations with no diagnostic output. Consider logging a warning when `!ok` so the user knows a connection reference couldn't be resolved.

jongio (commented):
All 16 of my previous inline comments are addressed in the latest commits - nice work. The fixes look solid (maps.Clone for credential isolation, index-based connName dedup, proper regex sanitization, reserved key validation, NormalizeToolKind in prompt path, etc).

CI is still red - same issues from my last review:

- **cspell**: 12 unknown words - add Alphanum, CLIENTSECRET, Toolsets, ADLS, Qdrant to the cspell overrides; change "normalises" to "normalizes"
- **gofmt**: 3 files need formatting - run `gofmt -s -w` on init_test.go, listen.go, foundry_toolsets_client_test.go
- **gosec**: 2 G101 false positives in yaml_test.go - add `//nolint:gosec` comments
- **go-fix**: init.go has a manual `for k, v := range toolMap` loop - replace with `maps.Copy(tool, toolMap)`

Once CI is green I'm happy to approve.

cli/azd/extensions/azure.ai.agents/internal/cmd/listen.go
trangevi: These are formatted the way they are because of how those specific env vars are provided to bicep. It seems like the true fix is that both this needs to be done, and some changes need to be made to the bicep parameters. However, that will break back compat currently.

We need a proper versioning story of the bicep, but that's out of the scope of this change, and with the proximity to PuPr I don't want to break old stuff for this right now. We'll make this change more holistically later.

JeffreyCA (approved):

PR #7642, opened 2026-04-10
fix: handle leaked ANSI escape sequences in ux input on PowerShell

### LINKED ISSUE #7641: Arrow keys print escape sequence characters in select/multi-select filter on PowerShell
### Description

Arrow key presses (up, down, left, right) are rendered as raw partial escape sequence characters in the filter text field of `select` and `multi-select` UX components when running azd in a PowerShell terminal. Instead of navigating the list or moving the cursor, the characters `[A`, `[B`, `[C`, `[D` are appended to the filter string.

For example, the filter field displays:
```
Filter: [B[B[B[C[C[C[D[D[D[A[A[A[B[B[B
```

### Regression Source

This was introduced by PR #6739 (commit af2fc5ae4), which replaced `github.com/eiannone/keyboard` with `github.com/AlecAivazis/survey/v2/terminal` RuneReader to fix arrow key handling on Ghostty terminal (issue #5704).

The `eiannone/keyboard` library used terminfo sniffing and didn't recognize Ghostty, but handled PowerShell correctly. The replacement `survey/v2/terminal` RuneReader fixes Ghostty but breaks PowerShell — the RuneReader's escape sequence parsing doesn't properly convert ANSI sequences to key constants in all Windows console configurations.

### Root Cause

When arrow keys are pressed in PowerShell, they generate ANSI escape sequences (`ESC[A`, `ESC[B`, `ESC[C`, `ESC[D`). The survey library's `RuneReader.ReadRune()` is expected to parse these and return special key constants (e.g., `surveyterm.KeyArrowUp`). 

On PowerShell, this parsing fails — the ESC byte (0x1b) is consumed but the remaining `[` and letter (`A`/`B`/`C`/`D`) are returned as individual runes. These pass the `unicode.IsPrint()` check in `pkg/ux/internal/input.go` (line ~131) and get appended to the filter value as text.

The arrow key checks in `select.go` and `multi_select.go` are correct but never fire because the key is never recognized as an arrow key constant upstream.

### Steps to Reproduce

1. Open a **PowerShell** terminal (Windows PowerShell or PowerShell 7)
2. Run any azd command that triggers a `select` or `multi-select` prompt with filtering enabled (e.g., `azd init` and reach the template selection, or `azd env select`)
3. Press arrow keys (up/down/left/right) while the filter field is active

### Expected Result

- **Up/Down arrows**: Navigate through the list options
- **Left/Right arrows**: Move cursor within the filter text (or ignored gracefully)
- Filter text remains empty or shows only user-typed characters

### Actual Result

- Arrow keys do not navigate the list
- The filter text field accumulates partial escape sequences: `[A`, `[B`, `[C`, `[D`
- Example display: `Filter: [B[B[B[C[C[C[D[D[D[A[A[A[B[B[B`

### Affected Components

| File | Role |
|------|------|
| `pkg/ux/internal/input.go:112-139` | Main input loop — escape sequence bytes leak through `IsPrint()` |
| `pkg/ux/internal/input.go:96` | `SetTermMode()` — may not enable `ENABLE_VIRTUAL_TERMINAL_INPUT` on Windows |
| `pkg/ux/select.go:175-179` | Arrow key handlers — correct but never reached |
| `pkg/ux/multi_select.go:180-206` | Arrow key handlers — correct but never reached |

### Environment

- **Terminal**: PowerShell (both Windows PowerShell 5.1 and PowerShell 7)
- **OS**: Windows
- **azd version**: Current main branch (post-PR #6739)

### Related Issues

- #5704 — Original Ghostty arrow key issue (fixed by PR #6739)
- #6739 — PR that introduced this regression

wbreza:
## Summary

Fixes arrow keys printing escape sequence characters (`[A`, `[B`, `[C`, `[D`) in the filter text of select and multi-select UX components when running azd in PowerShell.

## Problem

The survey library's `RuneReader.ReadRune()` fails to parse ANSI escape sequences on Windows/PowerShell. The ESC byte (0x1b) is consumed but the remaining `[` and direction letter leak through as individual runes, pass `unicode.IsPrint()`, and get appended to the filter text instead of being recognized as arrow key navigation.

This was a regression introduced by PR #6739 (Ghostty terminal fix), which replaced `eiannone/keyboard` with `survey/v2/terminal`.

## Solution

Added escape sequence detection in `pkg/ux/internal/input.go` that intercepts leaked ANSI fragments and converts them to the proper key constants before they reach the handler logic. Handles both:

- **CSI sequences** (`ESC [ A/B/C/D`) — standard arrow keys
- **Application mode sequences** (`ESC O A/B/C/D`) — used by vim, tmux, screen
- **Leaked sequences** (`[ A/B/C/D` or `O A/B/C/D`) — where ESC was already consumed

All existing components (select, multi-select, confirm, prompt) benefit automatically since they share the `Input` layer.

## Testing

Added 36 table-driven test cases covering:
- All 4 arrow directions for CSI, application mode, and leaked variants
- Non-direction character passthrough
- Read error handling during look-ahead
- Regular character passthrough

## Changes

| File | Change |
|---|---|
| `pkg/ux/internal/input.go` | Added `ansiDirectionKey()` and `resolveEscapeSequence()` helpers; integrated into `ReadInput()` loop with pending rune buffer |
| `pkg/ux/internal/input_test.go` | New file with comprehensive test coverage |

Fixes #7641

### DIFF
--- a/cli/azd/pkg/ux/internal/input.go
+++ b/cli/azd/pkg/ux/internal/input.go
@@ -97,6 +97,17 @@ func (i *Input) ReadInput(ctx context.Context, config *InputConfig, handler KeyP
 			errChan <- err
 			return
 		}
+
+		// On Windows, clear ENABLE_VIRTUAL_TERMINAL_INPUT so the
+		// console delivers native virtual-key codes instead of ANSI
+		// escape sequences. The survey library's RuneReader expects
+		// VK codes for arrow key dispatch. RestoreTermMode() will
+		// restore the original console mode when input completes.
+		if err := disableVirtualTerminalInput(os.Stdin); err != nil {
+			// Non-fatal: worst case is the pre-existing ANSI leak.
+			_ = err
+		}
+
 		defer func() {
 			if err := rr.RestoreTermMode(); err != nil {
 				log.Printf("Error restoring terminal mode: %v\n", err)
--- a/cli/azd/pkg/ux/internal/console_other.go
+++ b/cli/azd/pkg/ux/internal/console_other.go
@@ -1,0 +1,13 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+//go:build !windows
+
+package internal
+
+import "os"
+
+// disableVirtualTerminalInput is a no-op on non-Windows platforms.
+func disableVirtualTerminalInput(_ *os.File) error {
+	return nil
+}
--- a/cli/azd/pkg/ux/internal/console_windows.go
+++ b/cli/azd/pkg/ux/internal/console_windows.go
@@ -1,0 +1,43 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+//go:build windows
+
+package internal
+
+import (
+	"os"
+
+	"golang.org/x/sys/windows"
+)
+
+const enableVirtualTerminalInput uint32 = 0x0200
+
+// disableVirtualTerminalInput clears the ENABLE_VIRTUAL_TERMINAL_INPUT
+// console mode flag so that Windows delivers native virtual-key codes
+// (KEY_EVENT_RECORD with wVirtualKeyCode) instead of ANSI escape
+// sequences for arrow keys and other special keys.
+//
+// The survey library's Windows RuneReader expects native VK codes
+// (it checks unicodeChar == 0, then switches on wVirtualKeyCode).
+// When ENABLE_VIRTUAL_TERMINAL_INPUT is set — as it is by default in
+// Windows Terminal, PowerShell 7, VS Code, and Ghostty — the console
+// emits ESC [ A / ESC [ B / etc. instead, which leak through as
+// literal characters.
+//
+// This function is called after SetTermMode() and its effect is
+// reversed when RestoreTermMode() restores the original console mode.
+func disableVirtualTerminalInput(f *os.File) error {
+	h := windows.Handle(f.Fd())
+
+	var mode uint32
+	if err := windows.GetConsoleMode(h, &mode); err != nil {
+		return err
+	}
+
+	if mode&enableVirtualTerminalInput == 0 {
+		return nil // VTI not set, nothing to do
+	}
+
+	return windows.SetConsoleMode(h, mode&^enableVirtualTerminalInput)
+}


cli/azd/pkg/ux/internal/input.go
```diff
@@ -52,6 +52,77 @@ func (i *Input) ResetValue() {
...
+//   - Full:   ESC (0x1b) followed by '[' or 'O' and a direction letter
+//
+// When the look-ahead does not match a known sequence the consumed runes
+// are returned so the caller can process them normally.
+func resolveEscapeSequence(
+	char rune,
+	readRune func() (rune, int, error),
+) []rune {
+	// Leaked sequence: ESC was already consumed; '[' or 'O' + direction leaked.
+	if char == '[' || char == 'O' {
```
jongio: **[MEDIUM]** The leaked-sequence branch also triggers on literal 'O', which is a regular letter. Typing uppercase `OA` or `OD` in a filter field gets silently intercepted as an arrow key instead of text input.

The PowerShell leak this PR fixes produces `[` + direction - not `O` + direction. Application-mode sequences (ESC O A) are already handled safely by the full-ESC branch below, which is gated behind `KeyEscape`.

Consider removing `O` from this branch:
```suggestion
	if char == '[' {
```

cli/azd/pkg/ux/internal/input.go
```diff
@@ -52,6 +52,77 @@ func (i *Input) ResetValue() {
...
+// resolveEscapeSequence detects ANSI escape sequence fragments that leak
+// through when the survey library's RuneReader fails to parse them (e.g.
+// on Windows/PowerShell where the ESC byte is consumed but '[' and the
+// direction letter arrive as individual runes).
+//
+// Both CSI-style (ESC '[') and application-mode (ESC 'O') sequences are
+// handled so that arrow keys work correctly in terminals such as vim,
+// tmux, and screen that use application cursor key mode.
+//
+// Two patterns are recognised:
```
jongio: **[MEDIUM]** `recognised` fails the cspell CI check. Use American spelling.

```suggestion
// Two patterns are recognized:
```

jongio (commented):
Fixes the PowerShell arrow-key leak at the right layer - the shared `Input` component all UX controls use. Tests are thorough.

Issues to address:
- input.go:93 - leaked-sequence branch also intercepts literal 'O' + uppercase A-D as arrow keys
- input.go:82 - `recognised` fails cspell; use `recognized`

cli/azd/pkg/ux/internal/input.go
```diff
@@ -52,6 +52,77 @@ func (i *Input) ResetValue() {
...
+				return []rune{char, next}
+			}
+			if key, ok := ansiDirectionKey(dir); ok {
+				return []rune{key}
+			}
+			return []rune{char, next, dir}
+		}
+		return []rune{char, next}
+	}
+
```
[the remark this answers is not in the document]
wbreza: The blocking look-ahead is intentional and mirrors the approach used by the survey library's own POSIX \\ReadRune()\\ implementation (\\
unereader_posix.go\\), which performs the same synchronous read-ahead after detecting ESC to assemble complete escape sequences.

In practice, standalone \\[\\, \\O\\, and ESC are extremely rare in filter input contexts — users are typing search terms, not bracket characters. The tradeoff of a momentary pause on these edge-case keystrokes is acceptable versus the significant complexity of adding a timeout/peek mechanism with platform-specific non-blocking I/O.

If this becomes a real usability issue we can revisit with a buffered approach, but for now this matches the battle-tested pattern from the upstream library.

cli/azd/pkg/ux/internal/input.go
wbreza: Fixed — changed to American spelling `recognized`. This was addressed in an earlier push and cspell CI is passing.

cli/azd/pkg/ux/internal/input.go
wbreza: Good catch — you're right that `O` + direction letter in the leaked path would intercept normal filter typing like `OA` or `OD`. Removed `O` from the leaked-sequence condition so only `[` + direction is matched (the actual PowerShell leak pattern). Application-mode `ESC O A/B/C/D` sequences still work correctly when the full ESC prefix is present. Added tests verifying `O` + direction passes through as regular characters. Fixed in latest push.

vhvb1989:
## Alternative Approach: Clear `ENABLE_VIRTUAL_TERMINAL_INPUT` at the source

### Root Cause

The survey library's Windows `ReadRune()` (`runereader_windows.go`) uses `ReadConsoleInputW` and maps virtual key codes (`VK_UP`, `VK_DOWN`, etc.) to key constants. This works correctly **unless** `ENABLE_VIRTUAL_TERMINAL_INPUT` (0x0200) is active on the console handle — which PowerShell enables. When that flag is set, Windows translates physical key presses into VT/ANSI escape sequences, delivering them as multiple KEY_EVENTs with unicode chars (`0x1b`, `[`, `A`) instead of a single event with a virtual key code.

The survey library's `SetTermMode()` on Windows only clears `ENABLE_ECHO_INPUT | ENABLE_LINE_INPUT | ENABLE_PROCESSED_INPUT` — it never clears `ENABLE_VIRTUAL_TERMINAL_INPUT`. That is the actual root cause.

### Proposed Fix (Alternative A)

Wrap the `SetTermMode()` call with a Windows-specific helper that additionally clears `ENABLE_VIRTUAL_TERMINAL_INPUT`. This tells the Windows console to deliver native virtual key codes instead of ANSI sequences, so the survey `ReadRune()` handles them correctly through its existing `VK_*` mapping.

This would be ~15–25 lines in a build-tagged Windows file + a no-op for other platforms, versus the 362-line escape sequence parser in this PR.

### Comparison: Current PR vs Alternative A

| Dimension | Current PR (escape parsing) | Alt A (clear VTI flag) |
|---|---|---|
| **Lines changed** | +362 / -5 (75 logic + 264 tests) | ~15–25 lines |
| **Root cause addressed?** | ❌ Compensates downstream | ✅ Fixes at the source |
| **Cross-platform side effects** | ⚠️ Typing `[` then `A`–`D` in a filter on Linux/macOS is incorrectly consumed as an arrow key; ESC press blocks waiting for next char | ✅ None — Windows-only via build tag |
| **Complexity** | Moderate: pending-rune buffer, look-ahead state machine | Low: one bit-mask line |
| **Maintenance** | Ongoing (state machine, new escape sequences) | Near-zero (stable Win32 API since Windows 10 1511) |
| **Regression risk** | Moderate: changes input loop for all platforms | Low: only affects Windows terminal setup; `RestoreTermMode` restores original state |
| **Ghostty fix preserved?** | ✅ (POSIX, unaffected) | ✅ (POSIX, unaffected) |

### Why this matters

The current fix introduces a real (if unlikely) false positive on non-Windows: if a user types a literal `[` followed by `A`, `B`, `C`, or `D` in a filter field (e.g. searching for `[Dev`), the look-ahead in `resolveEscapeSequence` will **block** waiting for the next character and then **swallow** the pair as an arrow key. This only affects the leaked `[` path — a pattern that exclusively occurs on Windows/PowerShell.

Additionally, on POSIX, the survey library's `ReadRune()` already correctly parses ANSI escape sequences via a buffered reader. The full ESC path in `resolveEscapeSequence` duplicates that logic and adds a blocking `readRune()` call after a standalone ESC press.

vhvb1989 (changes requested):
There is a simpler alternative that addresses the root cause rather than compensating downstream: clearing `ENABLE_VIRTUAL_TERMINAL_INPUT` in `SetTermMode()` on Windows so the console delivers native virtual key codes instead of ANSI sequences.

See [this comment](https://github.com/Azure/azure-dev/pull/7642#issuecomment-4226763834) for the full definition of the alternative and a detailed comparison with the current approach.

Key concerns with the current fix:
- **Cross-platform false positive**: On Linux/macOS, typing `[` followed by `A`–`D` in a filter field is incorrectly consumed as an arrow key, since the leaked `[` + direction pattern only occurs on Windows/PowerShell.
- **Blocking regression**: The `resolveEscapeSequence` look-ahead does a blocking `readRune()` call that can stall input processing on standalone `[` or ESC presses.
- **Code size**: 362 lines (including a state machine with pending-rune buffer) vs ~20 lines for the root-cause fix.

The alternative (clearing `ENABLE_VIRTUAL_TERMINAL_INPUT`) is smaller, has zero cross-platform risk, and eliminates the problem at its source.

vhvb1989:
Draft PR illustrating Alternative A: #7646

**3 files, +71 lines** (mostly comments/copyright) vs the current 362-line approach. Builds and tests pass on both Linux and Windows cross-compilation. The fix is entirely Windows-scoped via build tags with zero impact on other platforms.

jongio (changes requested):
I verified the survey library source directly (`runereader_windows.go:72`) - `SetTermMode()` clears `ECHO|LINE|PROCESSED` but never `ENABLE_VIRTUAL_TERMINAL_INPUT`. When PowerShell enables VTI, `ReadConsoleInputW` delivers 3 separate KEY_EVENTs (0x1b, `[`, `A`) instead of one event with `VK_UP`. The library's VK mapping (lines 118-137) already handles arrow keys correctly when VTI is off.

Two remaining concerns after the latest fixes:

1. **[+direction false positive on POSIX**: Typing `[B` (searching for "[Build") or `[A` ("[API") in a filter gets consumed as an arrow key on Linux/macOS. The `[` is common in search terms - bracket expressions, array notation, markdown links. This only happens in the leaked-sequence path that doesn't apply to those platforms.

2. **Per-keystroke allocation**: `resolveEscapeSequence` returns `[]rune{char}` on every regular keypress (the common path), adding a heap allocation per keystroke that didn't exist before. Minor, but unnecessary in an input hot loop.

I agree with the alternative approach in #7646 - clearing VTI at the console level is smaller (~20 lines), addresses the root cause, has zero cross-platform risk via build tags, and `RestoreTermMode()` correctly re-enables VTI after input completes. I'd recommend pivoting to that approach.

wbreza:
Great suggestion, @vhvb1989 — you were absolutely right. Rewrote the fix to clear ENABLE_VIRTUAL_TERMINAL_INPUT after SetTermMode() instead of parsing leaked escape sequences. Much cleaner approach that fixes the root cause:

- **console_windows.go**: Clears VTI flag so the console delivers native virtual-key codes that the survey library's ReadRune() already handles correctly
- **console_other.go**: No-op on non-Windows (POSIX RuneReader already parses ANSI sequences)
- Removed all the escape sequence parsing, look-ahead buffer, and 264-line test file

The original approach had the cross-platform false positive and blocking issues you flagged. This is ~67 lines total vs ~362 before. Thanks for pushing back on this — the result is significantly better.

JeffreyCA (approved):
> image


I couldn't repro the issue using PowerShell on Windows (Windows Terminal and VS Code) on v1.23.14, but didn't notice any difference with this PR either.

vhvb1989 (approved):

jongio (approved):
Clean root-cause fix. Clearing VTI after SetTermMode() is the right approach - RestoreTermMode() correctly re-enables it from the saved state. Build tags keep it zero-risk on non-Windows. LGTM.

PR #7403, opened 2026-03-31
Add Go module versioning to ship SDK with CLI releases

wbreza:
## Summary

Adds Go-module-compatible version tags alongside each CLI release so extension developers can use standard semver instead of pseudo-versions in their go.mod files. The Go module version mirrors the CLI version for a simple mental model.

## Changes

**Pipeline**
- eng/pipelines/templates/steps/publish-cli.yml - New step after GitHub release creation that creates a Go module version tag via the GitHub API on the same commit. Idempotent.

**Version Synchronization**
- cli/azd/pkg/azdext/version.go - Updated const Version from 0.1.0 to 1.24.0-beta.1 to mirror the CLI version.
- eng/scripts/Update-CliVersion.ps1 - Auto-updates version.go when the CLI version is bumped during the release process.

**Documentation**
- cli/azd/docs/sdk-versioning.md (new) - Reference doc for extension developers.
- cli/azd/AGENTS.md - Added Go Module Versioning section.

## Issue References

Resolves #5837

## Testing

- Pipeline changes are additive (no existing behavior modified)
- Version sync script uses regex replacement validated against semver and pre-release formats
- No functional code changes to the CLI binary

## Notes

Post-merge steps required:
- Backfill Go module tags for recent releases (1.23.11 to 1.23.13)
- Verify Go module proxy resolves the tagged versions
- Update extension go.mod files to use semver references

PR #7656, opened 2026-04-10
fix changelog skill to include azdext version.go

### LINKED ISSUE #7638: Create changelog for azd 1.23.15

rajeshkamal5050:
The changelog skill didn't know about `cli/azd/pkg/azdext/version.go` (added in #7403). Core releases need to bump the `Version` constant there too.

Found this when PR #7639 bumped `cli/version.txt` to 1.23.15 but left `version.go` at 1.24.0-beta.1.

Updated SKILL.md and scope-rules.md to include `version.go` in the core release file list.

Fixes https://github.com/Azure/azure-dev/issues/7638

### DIFF
--- a/.github/skills/changelog-generation/SKILL.md
+++ b/.github/skills/changelog-generation/SKILL.md
@@ -45,12 +45,12 @@ Auto-detect scope from the current working directory:
 
 ### Step 2 — Determine Version & Update Files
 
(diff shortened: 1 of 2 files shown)


jongio (approved):
Clean fix. Both SKILL.md and scope-rules.md are updated consistently - version.go is added to the file list and the derivation steps are renumbered correctly. This closes the gap that caused version drift in #7639.

PR #7668, opened 2026-04-12
fix(agents): add OAuth URL, audience, connectorName fields to ConnectionResource and Connection structs

lindazqli:
## Problem

The Go structs that unmarshal connection resources from `azure.yaml` are incomplete. When users specify OAuth, audience, or connectorName fields in their `azure.yaml`, the Go YAML unmarshaler silently drops them because the struct fields don't exist. Those values never reach the Bicep parameters and are therefore never provisioned.

### Affected structs

| Struct | File | Missing fields |
|---|---|---|
| `ConnectionResource` | `agent_yaml/yaml.go` | `AuthorizationUrl`, `TokenUrl`, `RefreshUrl`, `Scopes`, `Audience`, `ConnectorName` |
| `Connection` | `project/config.go` | Same 6 fields |
| `extractConnectionConfigs` | `cmd/init.go` | Missing mappings for all 6 fields |

### Root cause

Silent YAML unmarshal drop — Go's YAML library ignores keys that have no corresponding struct field.  Fields present in `azure.yaml` → unmarshaled into `ConnectionResource` → mapped to `project.Connection` → serialized to Bicep params. If any step is missing a field, it's gone.

### `agent_yaml/yaml.go`
- Added 6 fields to `ConnectionResource`: `AuthorizationUrl`, `TokenUrl`, `RefreshUrl`, `Scopes`, `Audience`, `ConnectorName`
- Added missing `AuthType` constants that the REST API accepts but weren't declared: `UserEntraToken`, `AgenticIdentity`, `ManagedIdentity`, `ServicePrincipal`, `UsernamePassword`, `AccessKey`, `AccountKey`, `SAS`

### `project/config.go`
- Added the same 6 fields to `project.Connection`

### `cmd/init.go`
- Added field mappings for all 6 new fields in `extractConnectionConfigs`

## Testing

These fields are validated against the Azure AI Projects REST API `2025-04-01-preview` connection schema. Companion Bicep PR: https://github.com/Azure-Samples/azd-ai-starter-basic/pull/51 (adds the same fields to `connection.bicep` so they flow through to ARM).


cli/azd/extensions/azure.ai.agents/internal/pkg/agents/agent_yaml/yaml.go
```diff
@@ -611,6 +619,24 @@ type ConnectionResource struct {
...
+	RefreshUrl string `json:"refreshUrl,omitempty" yaml:"refreshUrl,omitempty"`
+
+	// Scopes is the list of OAuth2 scopes to request (OAuth2 authType).
+	Scopes []string `json:"scopes,omitempty" yaml:"scopes,omitempty"`
+
+	// Audience is the token audience for UserEntraToken / AgenticIdentity auth types.
+	Audience string `json:"audience,omitempty" yaml:"audience,omitempty"`
+
+	// ConnectorName is the managed connector name for OAuth2 managed connectors.
+	ConnectorName string `json:"connectorName,omitempty" yaml:"connectorName,omitempty"`
```
jongio, line 639: `ConnectorName` comment says "OAuth2 managed connectors" but both JSON schemas describe this as "Connector name for AgenticIdentity auth type." These are different auth types - the comment should match the schema since that's the API contract.

cli/azd/extensions/azure.ai.agents/internal/pkg/agents/agent_yaml/yaml.go
```diff
@@ -611,6 +619,24 @@ type ConnectionResource struct {
...
+	TokenUrl string `json:"tokenUrl,omitempty" yaml:"tokenUrl,omitempty"`
+
+	// RefreshUrl is the OAuth2 token refresh endpoint URL (OAuth2 authType).
+	RefreshUrl string `json:"refreshUrl,omitempty" yaml:"refreshUrl,omitempty"`
+
+	// Scopes is the list of OAuth2 scopes to request (OAuth2 authType).
+	Scopes []string `json:"scopes,omitempty" yaml:"scopes,omitempty"`
+
+	// Audience is the token audience for UserEntraToken / AgenticIdentity auth types.
+	Audience string `json:"audience,omitempty" yaml:"audience,omitempty"`
```
jongio, line 636: `Audience` comment lists only UserEntraToken/AgenticIdentity, but the schema describes this field as applying to AAD, ManagedIdentity, AgenticIdentity, and UserEntraToken. Worth updating the comment to match.

jongio (commented):
No tests for the 6 new fields. A round-trip test (YAML unmarshal -> extractConnectionConfigs -> JSON marshal) for a connection with authorizationUrl/tokenUrl/scopes would catch tag typos and verify the full pipeline. There's already parse_test.go and yaml_test.go in the same package - this would fit naturally there.

schemas/v1.0/azure.yaml.json
```diff
@@ -369,7 +369,7 @@
                             "required": ["project"],
                             "properties": {
                                 "config": {
-                                    "$ref": "https://raw.githubusercontent.com/Azure/azure-dev/refs/heads/main/cli/azd/extensions/azure.ai.agents/schemas/azure.ai.agent.json",
+                                    "$ref": "#/definitions/aiAgentConfig",
```
trangevi: We should not be modifying this file, otherwise we'll have two separate sources of truth. The current reference pointer should be maintained.

trangevi (changes requested):

cli/azd/extensions/azure.ai.agents/internal/pkg/agents/agent_yaml/yaml.go
lindazqli: it will be applicable to ManagedIdentity, AgenticIdentity, UserEntraToken, will update

trangevi (approved):

PR #7679, opened 2026-04-13
feat: add --protocol flag to azd ai agent invoke

### LINKED ISSUE #7492: choose which protocol to use for `invoke`
agent.yaml can contain multiple protocols. 

The invoke command reads the protocol from agent.yaml

Devs need a way to specify which one they are using for `invoke`. 

akin to `azd ai agent invoke --protocol invocations <message>`

glharper:
fixes #7492
## Summary

Adds a `--protocol` / `-p` flag to `azd ai agent invoke` so developers can explicitly select which protocol to use when their `agent.yaml` declares multiple protocols.

### Usage

```bash
# Use invocations protocol explicitly
azd ai agent invoke --protocol invocations "Hello!"
azd ai agent invoke -p invocations -f request.json

# Use responses protocol explicitly (same as current default)
azd ai agent invoke --protocol responses "Hello!"

# Auto-detect from agent.yaml (existing behavior, unchanged)
azd ai agent invoke "Hello!"
```

## Changes

- **`invoke.go`**: Added `protocol` field to `invokeFlags`, registered `--protocol`/`-p` flag, added validation against supported values (`responses`, `invocations`), and refactored `Run()` to use a new `resolveProtocol()` method that checks the explicit flag first before falling back to agent.yaml auto-detection.
- **`invoke_test.go`**: Added `TestResolveProtocol_ExplicitFlag` (verifies explicit flag bypasses agent.yaml resolution) and `TestProtocolFlagValidation` (verifies valid values pass and invalid values produce clear errors).

## Behavior

| `--protocol` flag | Behavior |
|---|---|
| Not set | Auto-detect from `agent.yaml` (first listed protocol, default `responses`) — **unchanged** |
| `responses` | Use responses protocol |
| `invocations` | Use invocations protocol |
| Anything else | Validation error with suggestion |

### DIFF
--- a/cli/azd/extensions/azure.ai.agents/internal/cmd/invoke.go
+++ b/cli/azd/extensions/azure.ai.agents/internal/cmd/invoke.go
@@ -174,7 +174,14 @@ func (a *InvokeAction) Run(ctx context.Context) error {
 	}
 
 	// Remote: only allow the invocations protocol when vnext is enabled.
-	if isVNextEnabled(ctx) && protocol == agent_api.AgentProtocolInvocations {
+	if protocol == agent_api.AgentProtocolInvocations {
+		if !isVNextEnabled(ctx) {
+			return fmt.Errorf(
+				"remote invocations protocol requires hosted-agent vnext; enable 'enableHostedAgentVNext' or use '--protocol %s'",
+				agent_api.AgentProtocolResponses,
+			)
+		}
+
 		return a.invocationsRemote(ctx)
 	}
 	return a.responsesRemote(ctx)
--- a/cli/azd/extensions/azure.ai.agents/internal/cmd/invoke.go
+++ b/cli/azd/extensions/azure.ai.agents/internal/cmd/invoke.go
@@ -176,12 +176,12 @@ func (a *InvokeAction) Run(ctx context.Context) error {
 	// Remote: only allow the invocations protocol when vnext is enabled.
 	if protocol == agent_api.AgentProtocolInvocations {
 		if !isVNextEnabled(ctx) {
-			return fmt.Errorf(
-				"remote invocations protocol requires hosted-agent vnext; enable 'enableHostedAgentVNext' or use '--protocol %s'",
-				agent_api.AgentProtocolResponses,
+			return exterrors.Validation(
+				exterrors.CodeInvalidParameter,
+				"invocations protocol for remote agents requires vnext to be enabled",
+				"enable vnext or use --protocol responses",
 			)
 		}
-
 		return a.invocationsRemote(ctx)
 	}
 	return a.responsesRemote(ctx)
--- a/cli/azd/extensions/azure.ai.agents/internal/cmd/invoke.go
+++ b/cli/azd/extensions/azure.ai.agents/internal/cmd/invoke.go
@@ -36,6 +36,7 @@ type invokeFlags struct {
 	newSession      bool
 	conversation    string
 	newConversation bool
+	protocol        string
 }
 
 type InvokeAction struct {
@@ -69,6 +70,9 @@ session automatically. Pass --new-session to force a reset.`,
 
   # Invoke a specific remote agent by name
   azd ai agent invoke my-agent "Hello!"
+
+  # Invoke using a specific protocol
+  azd ai agent invoke --protocol invocations "Hello!"
 
   # Invoke with a file as the request body
   azd ai agent invoke -f request.json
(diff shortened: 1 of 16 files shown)


cli/azd/extensions/azure.ai.agents/internal/cmd/invoke.go
```diff
@@ -154,15 +174,26 @@ func (a *InvokeAction) Run(ctx context.Context) error {
...
 
 	// Remote: only allow the invocations protocol when vnext is enabled.
-	if isVNextEnabled(ctx) {
-		protocol := a.resolveRemoteProtocol(ctx)
-		if protocol == agent_api.AgentProtocolInvocations {
-			return a.invocationsRemote(ctx)
-		}
+	if isVNextEnabled(ctx) && protocol == agent_api.AgentProtocolInvocations {
+		return a.invocationsRemote(ctx)
 	}
```
jongio, line 186: When the user explicitly passes --protocol invocations for a remote invoke and vnext isn't enabled, this silently falls back to 
esponsesRemote() - ignoring the explicit flag. An explicit flag should either be honored or return a clear error like invocations protocol for remote agents requires vnext to be enabled. Silent fallback on an explicit user choice is a UX bug.

cli/azd/extensions/azure.ai.agents/internal/pkg/agents/agent_yaml/testdata/prompt-agent-full.yaml
```diff
@@ -0,0 +1,39 @@
...
+      topP: 0.95
+  instructions: |
+    You are a helpful testing assistant.
+    Always respond in a structured format.
+  tools:
+    - kind: webSearch
+      name: web-search
+    - kind: function
+      name: get_weather
+      description: Get weather for a location
```
jongio, line 26: publisher isn't a recognized field in Model (the schema uses provider), and maxTokens isn't in ModelOptions (it uses maxOutputTokens). These will parse but the values won't populate the Go structs. If this is intentional (testing unknown-field tolerance), worth a comment. If not, fix the field names.

jongio (commented):
The protocol flag feature is clean and well-tested. A few things to address:

1. **Silent fallback on explicit flag** (HIGH) - when the user explicitly passes --protocol invocations for remote invoke and vnext isn't enabled, Run() silently uses 
esponsesRemote() instead. An explicit flag choice should be honored or produce a clear error - not silently ignored. See inline comment.

2. **PR scope** - ~90% of this diff is new test coverage and YAML fixtures for existing code (models, map, helpers, testdata, init_copy, agent_context) that's unrelated to the protocol feature. Consider splitting: (1) protocol flag, (2) helpers.go pointer fix, (3) new test coverage. Smaller PRs are easier to review and bisect.

3. **prompt-agent-full.yaml** uses publisher and maxTokens which don't match the Go schema (provider and maxOutputTokens). These parse but won't populate the structs. See inline comment.

Positive notes:
- helpers.go pointer fix is correct and important - each property now gets its own pointer
- 
esolveProtocol() cleanly unifies local/remote with explicit flag priority
- Validation catches bad protocol values early with helpful suggestions
- Test structure is solid: parallel, table-driven, good edge cases

cli/azd/extensions/azure.ai.agents/internal/cmd/invoke.go
glharper: [Updated](https://github.com/Azure/azure-dev/pull/7679/commits/785e2045510cc9e9f767fd88ea66ea6915e1ac32)

cli/azd/extensions/azure.ai.agents/internal/pkg/agents/agent_yaml/testdata/prompt-agent-full.yaml
glharper: [Updated](https://github.com/Azure/azure-dev/pull/7679/commits/785e2045510cc9e9f767fd88ea66ea6915e1ac32)

jongio (approved):
Both issues from my earlier review are addressed - the silent fallback is now a clear validation error, and the YAML fixture fields match the Go schema.

Two minor observations (not blocking):

1. The refactored `Run()` also changes auto-detect behavior: if `agent.yaml` resolves to `invocations` remotely and vnext is off, users now get an error instead of a silent fallback to `responses`. This is actually better (fail-fast vs silent degradation), but worth noting in case anyone's relying on the old behavior.

2. Consider adding a test for the vnext gating path (explicit `--protocol invocations` + remote + vnext disabled should error). The current tests validate flag parsing but don't cover this decision branch.

Clean feature, good tests, nice refactor of `resolveProtocol()`.

PR #7685, opened 2026-04-13
feat: add .azdignore support for template init

jongio:
## Add `.azdignore` support for `azd init`

Fixes #4142. Addresses #7669.

Revives the stale PR #4146 (by @jongio, 2024) with a clean reimplementation that incorporates all historical review feedback from #4146, the related packaging-ignore work (#4258, #5383), and multi-model expert code review.

### Problem

When running `azd init --template <repo>`, ALL files from the template repository are copied into the consumer's project — including contributor-only files like `SECURITY.md`, `CODE_OF_CONDUCT.md`, `.github/` CI configs, etc. This clutters consumer projects with files they don't need and didn't ask for (#7669).

### Solution

Template authors can now place a `.azdignore` file at the root of their template repository to exclude files from being copied when consumers run `azd init`. The file uses standard `.gitignore` syntax.

```
# .azdignore — files template consumers don't need
SECURITY.md
CODE_OF_CONDUCT.md
CONTRIBUTING.md
.github/
docs/
*.test.js
```

### History & Related Work

This feature has a long history across multiple issues and PRs spanning 2+ years:

| Reference | Type | Author | Status | Relationship |
|-----------|------|--------|--------|-------------|
| #4142 | Issue | @jongio | Open | Original `.azdignore` proposal (2024) |
| #7669 | Issue | @therealjohn | Open | Trigger issue — unwanted files in `azd init` |
| #4146 | PR | @jongio | Closed (stale) | First `.azdignore` implementation attempt |
| #4258 | PR | @jongio | Closed (superseded) | Packaging ignore — led to `.funcignore`/`.webappignore` |
| #5383 | PR | @weikanglim | Merged (1.18.0) | Shipped packaging ignore (`.funcignore`/`.webappignore`) |
| #1039 | Issue | @jongio | Closed | Early ignore discussion |
| #6754 | Issue | @therealjohn | Open | Related — preview files before init |

### Relationship to Existing Ignore Features

There are **two distinct ignore features** in azd — this PR implements only the first:

| Feature | File | Phase | Status |
|---------|------|-------|--------|
| **Init-time ignore** | `.azdignore` | `azd init` | **This PR** |
| Package-time ignore | `.funcignore` / `.webappignore` | `azd deploy` | Shipped 1.18.0 (#5383) |
| Docker ignore | `.dockerignore` | Docker build | Already supported |
| Git ignore | `.gitignore` | `azd init` (local copy) | Already supported |

All use the same `go-gitignore` library and root-only pattern.

[...]

PR #7697, opened 2026-04-13
Add .azdxignore support for azd x watch (#6152)

### LINKED ISSUE #6152: add way for azd x to ignore files for commands
Right now when you run `azd x watch` it montiors files and folders that I don't want it to.  I'd like to be able to set that in a .somethingignore file, like .azdxignore

jongio:
## Summary

Adds `.azdxignore` file support for `azd x watch`, allowing users to exclude files and directories from the file watcher. This eliminates excessive rebuilds caused by changes in `node_modules/`, build output, test artifacts, and other non-source files.

**Closes #6152** | Supersedes #6153 (stale draft, never merged)

### Shared `pkg/ignore` Package
A new reusable package at `cli/azd/pkg/ignore/` that:
- Loads `.azdxignore` from the project root (gitignore syntax)
- Also respects `.gitignore` (additive — both files apply)
- Uses `go-gitignore` (already a dependency, no new deps)
- Handles missing files gracefully, strips UTF-8 BOM, nil-safe

### Watch Integration
Both watch implementations now use the shared ignore matcher:
- `cli/azd/pkg/watch/watch.go` (conversational agent)
- `cli/azd/extensions/.../cmd/watch.go` (`azd x watch` CLI)

## Usage Example

Create a `.azdxignore` file in your project root:

```gitignore
# Ignore build output
dist/
build/
*.tmp

# Ignore test artifacts
coverage/
**/*.test.js

# Ignore IDE files
.idea/
.vscode/

# But keep the launch config
!.vscode/launch.json
```

The watcher will also respect patterns from your `.gitignore`. Both files use standard [gitignore syntax](https://git-scm.com/docs/gitignore).

## Issues Fixed from Prior PR #6153

This is a fresh implementation that addresses all 5 review findings from #6153:

| # | PR #6153 Issue | Fix |
|---|---|---|
| 1 | Double `os.Stat` in event handler (redundant syscall) | Single stat call, result reused for isDir + ignore check |
| 2 | Tests used `Absolute()` instead of `Relative()` | All tests use relative path matching |
| 3 | Hardcoded `time.Sleep` in tests (flaky) | Uses `require.Eventually` with proper polling |
| 4 | Silent fallback to absolute path on `filepath.Rel` error | Explicit error handling, correct root-relative paths |
| 5 | Duplicated ignore logic across two files | Shared `pkg/ignore` package used by both watchers |

### Additional Fixes (from MQ review)
- **CR-001**: Remove events no longer fail isDir detection — directory patterns now match correctly even when the file is already deleted
- **CR-002**: Fixed `os.Getenv("GOOS")` to `runtime.GOOS` (GOOS is a build constant, not an env var)

## Scope Decision

Audited all `azd x` commands (`build`, `pack`, `init`, `publish`, `release`, `version`, `metadata`). Only `watch` benefits from ignore support — other commands operate on flat/narrow directory scans with existing filters.

## Test Coverage

- **40 tests total** (16 in `pkg/ignore`, 24 in `pkg/watch`) — 39 pass, 1 skip (Windows file permission edge case)
- `pkg/ignore`: 85.2% coverage
- `pkg/watch`: 66.4% coverage (78% excluding deprecated `PrintChangedFiles`)
- 3 consecutive green runs, 0 flaky tests

## Files Changed

| File | Change |
|---|---|
| `cli/azd/pkg/ignore/ignore.go` | **NEW** — Shared Matcher type |
| `cli/azd/pkg/ignore/ignore_test.go` | **NEW** — 16 unit tests |
| `cli/azd/pkg/watch/watch.go` | Integrated ignore matcher |
| `cli/azd/pkg/watch/watch_test.go` | Added integration tests, fixed test quality |
| `cli/azd/pkg/watch/watch_helpers_test.go` | Minor cleanup (stdlib indexOf) |
| `cli/azd/extensions/.../cmd/watch.go` | Integrated ignore matcher |
| `cli/azd/.vscode/cspell.yaml` | Added "azdxignore" to dictionary |

### DIFF
--- a/cli/azd/extensions/microsoft.azd.extensions/internal/cmd/watch.go
+++ b/cli/azd/extensions/microsoft.azd.extensions/internal/cmd/watch.go
@@ -13,6 +13,7 @@ import (
 
 	"github.com/azure/azure-dev/cli/azd/extensions/microsoft.azd.extensions/internal"
 	"github.com/azure/azure-dev/cli/azd/pkg/azdext"
+	"github.com/azure/azure-dev/cli/azd/pkg/ignore"
 	"github.com/bmatcuk/doublestar/v4"
 	"github.com/fatih/color"
 	"github.com/fsnotify/fsnotify"
@@ -71,6 +72,19 @@ func runWatchAction(ctx context.Context, flags *watchFlags) error {
 	}
 	defer watcher.Close()
 
+	cwd, err := os.Getwd()
+	if err != nil {
+		return fmt.Errorf("failed to get current working directory: %w", err)
+	}
+
+	// Load ignore patterns from .azdxignore and .gitignore files.
+	ignoreMatcher, err := ignore.NewMatcher(cwd)
+	if err != nil {
+		return fmt.Errorf("failed to load ignore patterns: %w", err)
+	}
+
+	// Hardcoded folder ignores are kept as a fast-path default — they apply
+	// even when no .azdxignore or .gitignore file exists.
 	ignoredFolders := map[string]struct{}{
 		"bin":          {},
 		"obj":          {},
@@ -92,7 +106,7 @@ func runWatchAction(ctx context.Context, flags *watchFlags) error {
 		"package-lock.json", // Matches package-lock.json files
 	)
 
-	if err := watchRecursive(".", watcher, ignoredFolders); err != nil {
+	if err := watchRecursive(cwd, watcher, ignoredFolders, ignoreMatcher); err != nil {
 		return fmt.Errorf("Error watching for changes: %w", err)
 	}
 
@@ -112,7 +126,7 @@ func runWatchAction(ctx context.Context, flags *watchFlags) error {
 				return nil
 			}
 
-			// Ignore events matching glob patterns
+			// Fast path: ignore events matching hardcoded glob patterns.
 			shouldIgnore := false
 			for _, pattern := range globIgnorePaths {
 				matched, _ := doublestar.PathMatch(pattern, event.Name)
@@ -123,6 +137,23 @@ func runWatchAction(ctx context.Context, flags *watchFlags) error {
 			}
 			if shouldIgnore {
 				continue
+			}
+
+			// Check user-defined ignore patterns (.azdxignore / .gitignore).
+			// Use os.Stat once to determine if the path is a directory.
+			info, statErr := os.Stat(event.Name)
+			isDir := statErr == nil && info.IsDir()
+
+			if relPath, relErr := filepath.Rel(cwd, event.Name); relErr == nil {
+				if ignoreMatcher.IsIgnored(relPath, isDir) {
+					continue
+				}
+				// When the path no longer exists (e.g. Remove event), os.Stat fails
+				// and isDir defaults to false. Re-check as a directory so that
+				// directory-only patterns (trailing slash) still filter the event.
+				if statErr != nil && ignoreMatcher.IsIgnored(relPath, true) {
+					continue
+				}
 			}
 
 			// Collect unique changes
@@ -153,15 +184,29 @@ func runWatchAction(ctx context.Context, flags *watchFlags) error {
 	}
 }
 
-func watchRecursive(root string, watcher *fsnotify.Watcher, ignoredFolders map[string]struct{}) error {
+func watchRecursive(
+	root string,
+	watcher *fsnotify.Watcher,
+	ignoredFolders map[string]struct{},
+	ignoreMatcher *ignore.Matcher,
+) error {
 	return filepath.Walk(root, func(path string, info os.FileInfo, err error) error {
 		if err != nil {
 			return err
 		}
 		if info.IsDir() {
+			// Check hardcoded folder ignores.
 			if _, has := ignoredFolders[info.Name()]; has {
 				return filepath.SkipDir
 			}
+
+			// Check user-defined ignore patterns (.azdxignore / .gitignore).
+			if relPath, relErr := filepath.Rel(root, path); relErr == nil && relPath != "." {
+				if ignoreMatcher.IsIgnored(relPath, true) {
+					return filepath.SkipDir
+				}
+			}
+
 			err = watcher.Add(path)
 			if err != nil {
 				return fmt.Errorf("failed to watch directory %s: %w", path, err)
--- a/cli/azd/extensions/microsoft.azd.extensions/internal/cmd/watch.go
+++ b/cli/azd/extensions/microsoft.azd.extensions/internal/cmd/watch.go
@@ -7,6 +7,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"log"
 	"os"
 	"path/filepath"
 	"time"
@@ -72,6 +73,8 @@ func runWatchAction(ctx context.Context, flags *watchFlags) error {
 	}
 	defer watcher.Close()
 
+	// cwd is captured once and used as the immutable root for the entire watch session.
+	// All filepath.Rel calls reference this value so the root cannot drift.
 	cwd, err := os.Getwd()
 	if err != nil {
 		return fmt.Errorf("failed to get current working directory: %w", err)
@@ -144,7 +147,9 @@ func runWatchAction(ctx context.Context, flags *watchFlags) error {
 			info, statErr := os.Stat(event.Name)
 			isDir := statErr == nil && info.IsDir()
 
-			if relPath, relErr := filepath.Rel(cwd, event.Name); relErr == nil {
+			if relPath, relErr := filepath.Rel(cwd, event.Name); relErr != nil {
+				log.Printf("debug: failed to compute relative path for %s: %v", event.Name, relErr)
+			} else {
 				if ignoreMatcher.IsIgnored(relPath, isDir) {
 					continue
 				}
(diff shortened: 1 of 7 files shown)


wbreza (commented):
## Code Review — PR #7697: .azdxignore support for azd x watch

**Verdict: 💬 Comments** — Solid feature with good test coverage and proper .gitignore library usage. A couple of items worth addressing.

### Findings

#### 🟠 1. Matcher precedence / negation semantics unclear
.azdxignore patterns are loaded before .gitignore into separate matchers. The comment says patterns are "additive," but this means .gitignore negation patterns (!) **cannot** override .azdxignore matches — they're evaluated in separate matcher instances. This differs from standard git behavior where later patterns override earlier ones within a single scope.

**Suggestion:** Document this explicitly: _".azdxignore matches take precedence; .gitignore negation patterns cannot un-ignore files matched by .azdxignore."_ If that's not the intended behavior, the matchers need to be merged differently.

#### 🟠 2. Inconsistent root capture between two watchers
pkg/watch stores 
oot in the struct (immutable, thread-safe). The extension watch command uses a cwd variable from function scope. Both work today, but it's a different pattern for the same logic. Suggest the extension also capture root once and store it.

#### 🟡 3. Silent filepath.Rel() errors
Both watchers skip ignore checking silently if ilepath.Rel() returns an error. A debug-level log would help diagnose path resolution issues with symlinks or unusual mounts.

#### 🟡 4. Rename event edge case untested
os.Stat() failure on removed/renamed paths defaults to checking both file and dir patterns. Behavior may vary by OS. Consider a test for rename events.

### Note: Relationship to PR #7685
PR #7685 adds .azdignore for template init with its own loading in internal/repository/initializer.go. This PR creates a separate pkg/ignore package. Both use go-gitignore but don't share code. If both merge, consolidating into the shared package would reduce duplication.

### What's Good
- ✅ Correct go-gitignore library usage with cross-platform ilepath.ToSlash() normalization
- ✅ Nil Matcher is safe (defensive coding)
- ✅ Comprehensive tests: empty, comments, negation, wildcards, BOM, backslash paths, unreadable files
- ✅ UTF-8 BOM stripping matches existing codebase pattern
- ✅ Clean refactoring of watcher test helpers

Nice work on the shared pkg/ignore abstraction — good foundation for reuse. 👍

jongio:
Thanks for the thorough review @wbreza! All 4 findings addressed in 96ee0a2:

1. **Matcher precedence/negation** - Clarified doc comments on `NewMatcher` explaining additive/union semantics: `.gitignore` negation cannot un-ignore `.azdxignore` matches since each file is parsed independently.

2. **Inconsistent root capture** - Added comment documenting `cwd` is captured once as the immutable root for the watch session, matching the `pkg/watch` pattern.

3. **Silent filepath.Rel() errors** - Added debug-level logging at all 4 `filepath.Rel()` call sites across both watchers.

4. **Rename event untested** - Added `TestGetFileChanges_RenameFile` covering create -> rename -> verify new name appears.

Re: PR #7685 consolidation note - agreed, if both merge we should consolidate into the shared `pkg/ignore` package. Happy to follow up on that.

wbreza (commented):
## Re-Review — .azdxignore support for azd x watch

Re-reviewing to check status of prior findings.

---

### Prior Findings Status

| # | Finding | Status |
|---|---|---|
| **W1** | Matcher precedence / negation semantics | ⚠️ **Partial** — Comment added but doesn't explicitly explain that `.gitignore` negation (`!`) can't override `.azdxignore` matches (separate matchers, short-circuit logic) |
| **W2** | Inconsistent root capture | ✅ **Addressed** |
| **W3** | Silent `filepath.Rel()` errors | ❌ **Not addressed** — All 3 call sites still silently skip ignore checking on error. A debug-level log would help diagnose path resolution issues. |
| **W4** | Rename event edge case untested | ⚠️ **Partial** — Create→delete tested, but no test for the `os.Stat()` failure fallback path (`isDir=true` re-check) |
| **C1** | GetFileChanges() draining assertion | ✅ **Addressed** |
| **C2** | Vendor path absolute check | ✅ **Addressed** |
| **C3** | Direct type assertion panic | ✅ **Addressed** |

**4/7 fully addressed, 2 partial, 1 not addressed.**

---

### New Findings

#### 🟡 1. No user-facing documentation for `.azdxignore`
The PR body is thorough, but there's no `docs/` page, command help text, or README explaining the feature to end users — file format, placement, relationship to `.gitignore`.

#### 🟡 2. Missing cross-file negation test
`TestNewMatcher_Negation` only tests negation within a single file. A test verifying that `!pattern` in `.gitignore` cannot un-ignore a path matched by `.azdxignore` would document the exact behavior from W1.

#### 🟡 3. No test for `os.Stat()` failure fallback
📁 `pkg/watch/watch.go`
When `os.Stat()` fails (deleted file), code re-checks with `isDir=true`. This code path has no dedicated test.

#### 🔵 4. `filepath.Abs()` error not wrapped
📁 `pkg/ignore/ignore.go` ~line 46
Error from `filepath.Abs()` returned without context wrapping. Should be `fmt.Errorf("resolving root path: %w", err)`.

---

### Overall

Good progress — 4/7 prior findings fully resolved, Copilot's 3 comments all addressed. Remaining items are medium/low severity. The main gap is **W3** (silent `filepath.Rel()` errors) — even a single `log.Printf` at debug level would close this out. Nice work on the shared `pkg/ignore` package. 👍

jongio:
Thanks for the re-review @wbreza! Pushed 00d8802 addressing the remaining items. A couple of corrections on the status table:

**W1 (Matcher precedence)** - This is fully addressed. The `NewMatcher` godoc explicitly says: _"This means .gitignore negation patterns (!) cannot un-ignore paths that are matched by .azdxignore, since each file is parsed independently."_ Added `TestNewMatcher_CrossFileNegationCannotOverride` in this commit to lock down the behavior with a test too.

**W3 (Silent filepath.Rel)** - This was addressed in 96ee0a2. All 3 call sites have `log.Printf("debug: failed to compute relative path for %s: %v", ...)` - may have been missed on re-review.

**New items addressed in 00d8802:**
1. User-facing docs - Added `Long` description to the watch cobra command with `.azdxignore` format and example
2. Cross-file negation test - `TestNewMatcher_CrossFileNegationCannotOverride`
3. os.Stat() failure fallback test - `TestGetFileChanges_DeleteIgnoredDirFallback`
4. filepath.Abs() error wrapping - Now returns `fmt.Errorf("resolving root path: %w", err)`

All 43 tests pass (17 ignore + 26 watch).

cli/azd/extensions/microsoft.azd.extensions/internal/cmd/watch.go
```diff
@@ -28,6 +30,19 @@ func newWatchCommand() *cobra.Command {
 	watchCmd := &cobra.Command{
 		Use:   "watch",
 		Short: "Watches the azd extension project for file changes and rebuilds it.",
+		Long: `Watches the azd extension project for file changes and rebuilds it.
+
+Place a .azdxignore file in your project root to exclude paths from triggering
+rebuilds. It uses standard gitignore syntax (https://git-scm.com/docs/gitignore).
+Patterns from .gitignore are also respected. Both files are additive — a path is
+ignored if it matches either file.
```
JeffreyCA, line 38: Similar to how `azd x init` includes a `.gitignore` in the initial scaffold, should we also include an `.azdxignore` file?

E.g.  `cli/azd/extensions/microsoft.azd.extensions/internal/resources/languages/go/.gitignore`

jongio: Good idea to consider. Leaning toward deferring this to a follow-up: since the watcher already respects `.gitignore` (which every scaffold ships), most users won't need an `.azdxignore` out of the box - it's opt-in for extra watch-only exclusions. Once this lands I'd like to see what patterns users actually reach for before baking a default into each language scaffold. Happy to open a follow-up issue if you have specific patterns in mind (e.g. `dist/`, build output, coverage dirs).

wbreza (approved):
Re-review ΓÇö all four follow-up items from my 2026-04-14 review are cleanly addressed:

- **Cross-file negation test** ΓÇö `TestNewMatcher_CrossFileNegationCannotOverride` (`ignore_test.go:243`) correctly verifies that a `!important.log` in `.gitignore` cannot un-ignore an `*.log` match in `.azdxignore`.
- **Stat-fallback test** ΓÇö `TestGetFileChanges_DeleteIgnoredDirFallback` (`watch_test.go:358`) exercises the `statErr != nil` path when a dir-only pattern matches a removed directory, with both positive (`tracked.go` surfaces) and negative (`tmpout` does not leak) signals.
- **filepath.Abs error wrap** ΓÇö `ignore.go:40-43` now returns `fmt.Errorf(""resolving root path: %w"", err)`.
- **`Long` description on `x watch`** ΓÇö `ext_watch.go:33-45` explains `.azdxignore` semantics, links gitignore docs, and includes a usage example.

Minor nits (non-blocking, optional follow-up):

- `watch.go:127` dir-only-pattern re-check after stat failure can theoretically misfire if a user has a dir-only pattern (`mydir/`) and an unrelated **file** named `mydir`. Narrow edge case; a one-line comment acknowledging the tradeoff would help future maintainers.
- `ext_watch.go:33-45` example block ends with `!.vscode/launch.json` without a preceding `.vscode/` entry, which could confuse readers unfamiliar with negation semantics. Consider adding `.vscode/` on the line above.

LGTM.

PR #7698, opened 2026-04-13
fix: only return created resource groups from subscription deployments

### LINKED ISSUE #7643: Investigate using bicep snapshot predictedResources for resource group ownership classification
## Summary

Investigate whether `bicep snapshot` + `predictedResources` can supplement or simplify the resource group ownership classification pipeline introduced in #7603.

## Context

In the [PR #7603 discussion](https://github.com/Azure/azure-dev/pull/7603#issuecomment-4219970970), @vhvb1989 identified that `bicep snapshot`'s `predictedResources` output only includes resources the deployment **creates** — resources declared with the `existing` keyword are excluded. This could provide a fast, offline, deterministic signal for distinguishing created vs. pre-existing resource groups.

## Investigation areas

1. **Tier 0 / Tier 1 supplement**: Can `predictedResources` serve as a "Tier 0" check that short-circuits the classification pipeline when Bicep source files are available?
2. **Source file availability**: `azd down` currently works purely from deployment metadata. Snapshot requires the original Bicep files — how often are they available at destroy time?
3. **Adopted/redeployed RGs**: Does snapshot correctly handle RGs that were created externally then deployed into (operation type would be `Create` but the RG pre-existed)?
4. **Nested deployments and modules**: Are RGs created via nested deployments or modules flattened into `predictedResources`?
5. **Deployment outputs**: RGs referenced via deployment outputs (e.g., `output rgName string = resourceGroup.name`) are not resources — how to handle these?
6. **Deployment stacks path**: The stacks path skips RG classification entirely — does snapshot interact with this?

## Related

- PR: #7603 (4-tier classification pipeline)
- Original suggestion: https://github.com/Azure/azure-dev/pull/7603#issuecomment-4219970970

weikanglim:
## Summary
- change subscription deployment resource discovery to return only resource groups created by the deployment
- keep the dependency-based fallback for failed deployments
- add regression coverage for existing resource group deployments and subscription deployment resource listing

## Why
Subscription-scope deployments can create resources inside an existing resource group. ARM `outputResources` includes those nested resources, so azd could treat the existing resource group as deployment-owned during cleanup. This change fixes that existing resource group deployment behavior by **only** enumerating created resource groups.

## Testing
- go test ./pkg/azapi/...
- go build ./...
- manual end-to-end tests with mixed existing/created resource groups

Fixes #7643.

### DIFF
--- a/cli/azd/pkg/azapi/standard_deployments_test.go
+++ b/cli/azd/pkg/azapi/standard_deployments_test.go
@@ -101,17 +101,23 @@ func TestCreatedResourceGroupsFromDeployment(t *testing.T) {
 				},
 				{
 					ID: new(
-						"/subscriptions/sub-id/resourceGroups/groupA/Microsoft.Storage/storageAccounts/storageAccount",
+						"/subscriptions/sub-id/resourceGroups/groupA/providers/" +
+							"Microsoft.Storage/storageAccounts/storageAccount",
 					),
 				},
 				{
 					ID: new("/subscriptions/sub-id/resourceGroups/groupB"),
 				},
 				{
-					ID: new("/subscriptions/sub-id/resourceGroups/groupB/Microsoft.web/sites/test"),
+					ID: new(
+						"/subscriptions/sub-id/resourceGroups/groupB/providers/Microsoft.Web/sites/test",
+					),
 				},
 				{
-					ID: new("/subscriptions/sub-id/resourceGroups/groupExisting/Microsoft.Storage/storageAccounts/storageAccount"),
+					ID: new(
+						"/subscriptions/sub-id/resourceGroups/groupExisting/providers/" +
+							"Microsoft.Storage/storageAccounts/storageAccount",
+					),
 				},
 			},
 			ProvisioningState: DeploymentProvisioningStateSucceeded,
--- a/cli/azd/pkg/azapi/standard_deployments_test.go
+++ b/cli/azd/pkg/azapi/standard_deployments_test.go
@@ -49,7 +49,7 @@ func Test_StandardDeployments_GenerateDeploymentName(t *testing.T) {
 	}
 }
 
-func TestResourceGroupsFromDeployment(t *testing.T) {
+func TestCreatedResourceGroupsFromDeployment(t *testing.T) {
 	t.Parallel()
 
 	t.Run("references used when no output resources", func(t *testing.T) {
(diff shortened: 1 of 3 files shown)


cli/azd/pkg/azapi/standard_deployments_test.go
```diff
@@ -111,16 +111,16 @@ func TestResourceGroupsFromDeployment(t *testing.T) {
 					ID: new("/subscriptions/sub-id/resourceGroups/groupB/Microsoft.web/sites/test"),
 				},
 				{
-					ID: new("/subscriptions/sub-id/resourceGroups/groupC"),
+					ID: new("/subscriptions/sub-id/resourceGroups/groupExisting/Microsoft.Storage/storageAccounts/storageAccount"),
```
jongio: nit: This line is 131 characters, which breaks the 125-char `lll` lint rule. Wrap the string or add a `//nolint:lll` directive - the test already uses that pattern at lines 57-58.

jongio (commented):
The core logic change is right - switching from `ResourceGroupName != ""` to `ResourceType.Type == ResourceGroupResourceType.Type` correctly limits discovery to created resource groups. This matches the existing pattern in `bicep_provider.go:651`.

Good test coverage with the coverage3 integration test that validates only `RG1` gets queried (not `EXISTING`).

One CI issue to fix - see inline comment.

hemarina (approved):

jongio (approved):
All feedback addressed. The ResourceType filter aligns with the existing pattern in bicep_provider.go and correctly limits discovery to created resource groups. Tests look solid.

wbreza (approved):
## Supplementary Code Review — subscription deployment resource group filtering

Clean, well-scoped fix. The core logic change is correct — switching from ResourceGroupName != "" to ResourceType.Type == ResourceGroupResourceType.Type correctly limits discovery to resource groups **created by** the deployment, filtering out pre-existing RGs that merely contain deployed child resources. This aligns with the existing pattern in icep_provider.go:651.

### Findings

No new issues beyond what Copilot and jongio already flagged.

### What looks good
- ✅ Heuristic is sound: ARM includes created RGs as distinct output resources (/resourceGroups/RG1), while pre-existing RGs only appear as parents of child resources
- ✅ Dependency-based fallback preserved for non-succeeded deployments
- ✅ Good regression test in coverage3 — validates only RG1 is queried, not EXISTING
- ✅ Updated test properly uses full ARM resource IDs with /providers/ segments

LGTM 👍

weikanglim:
/check-enforcer override

Check enforcer is stuck

PR #7703, opened 2026-04-14
fix: reset PSModulePath in azd update to prevent PS7/PS5.1 module conflict

### LINKED ISSUE #7702: azd update fails with CouldNotAutoloadMatchingModule when parent shell is PowerShell 7
## Description

`azd update` fails with:
```
The 'Get-AuthenticodeSignature' command was found in the module
'Microsoft.PowerShell.Security', but the module could not be loaded.
FullyQualifiedErrorId: CouldNotAutoloadMatchingModule
```

## Root Cause

When `azd update` invokes the install script, the exec package wraps all Windows commands in `cmd.exe /c` (`command_runner.go:99`). The call chain is:

```
pwsh (PS7) -> azd.exe -> cmd.exe /c powershell.exe 5.1
```

PowerShell 5.1 normally overrides the inherited `PSModulePath` when its parent is another PowerShell process. However, the `cmd.exe` intermediary prevents this override. PS5.1 sees `cmd.exe` as its parent, not `pwsh`, so it keeps the inherited `PSModulePath` which includes PS7 Core-edition module paths.

PS7's `Microsoft.PowerShell.Security` module (`CompatiblePSEditions: Core`) appears **before** PS5.1's version (`CompatiblePSEditions: Desktop`) in the search path. PS5.1 attempts to load the Core-edition module, which fails with TypeData conflicts.

## Affected Environments

- **VSCode integrated terminal** (default shell: pwsh/PowerShell 7)
- **Windows Terminal** (default shell: pwsh/PowerShell 7)
- Any terminal where the parent shell is PowerShell 7+

## Fix

Reset `PSModulePath` to Windows PowerShell 5.1 defaults in `buildInstallScriptArgs()` before invoking the install script.

hemarina:
## Description

Fixes #7702

When `azd update` spawns `powershell.exe 5.1` via `cmd.exe /c` from a PowerShell 7 parent shell (VSCode terminal, Windows Terminal with pwsh), PS5.1 inherits PS7's `PSModulePath`. The PS7 Core-edition `Microsoft.PowerShell.Security` module is found before the PS5.1 Desktop-edition module, causing `Get-AuthenticodeSignature` to fail with `CouldNotAutoloadMatchingModule`.

## Root Cause

The exec package forces `cmd.exe /c` wrapping on all Windows commands (`command_runner.go:99`). The `cmd.exe` intermediary prevents PS5.1 from detecting that its grandparent is another PowerShell process, so PS5.1 does not override the inherited `PSModulePath`. The PS7 module paths (containing Core-only modules) appear before PS5.1 paths, causing incompatible module loading.

## Fix

Reset `PSModulePath` to Windows PowerShell 5.1 defaults in `buildInstallScriptArgs()` before invoking the install script. This ensures PS5.1 only sees its own Desktop-edition modules regardless of the parent shell.

## Testing

Verified by simulating the exact `cmd.exe /c powershell.exe` invocation path from a pwsh parent:

| Scenario | Result |
|---|---|
| Before fix: `cmd.exe /c powershell.exe ... Get-AuthenticodeSignature` | `CouldNotAutoloadMatchingModule` |
| After fix: same with PSModulePath reset prepended | `Status: Valid` |
| `go test ./pkg/update/... -run TestBuildInstallScriptArgs` | All 4 tests pass |


vhvb1989 (approved):

cli/azd/pkg/update/msi_windows.go
```diff
@@ -171,8 +171,16 @@ func buildInstallScriptArgs(channel Channel) []string {
 		scriptArgs = " -Version 'stable'"
 	}
 
+	// Reset PSModulePath to the Windows PowerShell 5.1 system modules directory.
+	// When launched via cmd.exe from a PowerShell 7 parent (e.g. VSCode or Windows Terminal),
+	// PS5.1 inherits PS7's PSModulePath which includes Core-edition modules that fail to load
+	// in the Desktop edition runtime, causing Get-AuthenticodeSignature to fail with
+	// "CouldNotAutoloadMatchingModule".
+	resetPSModulePath := "$env:PSModulePath = Join-Path $PSHOME 'Modules'; "
```
jongio, line 179: This is the core fix but the existing tests don't verify it's present. Both `TestBuildInstallScriptArgs` test cases should include `"$env:PSModulePath"` in their `wantContains` lists, and `TestBuildInstallScriptArgs_Structure` should check the script starts with the reset. Something like:

```go
require.True(t, strings.HasPrefix(script, "$env:PSModulePath"), "script should start with PSModulePath reset")
```

jongio (commented):
The PSModulePath fix is correct - $PSHOME under PS5.1 correctly resolves to its own modules directory regardless of parent shell. One gap: the existing tests don't verify this prefix is present.

jongio (approved):
My earlier test coverage comment was addressed - both `TestBuildInstallScriptArgs` and `TestBuildInstallScriptArgs_Structure` now verify the PSModulePath reset prefix. The config persistence change is a nice improvement - deferring channel persistence until after a successful update prevents stale config on failed updates, and the non-fatal warning with remediation command is the right call.

One minor gap: the post-update channel-save-failure path (where `configManager.Save` fails after a successful update) doesn't have a test. There's a `failSaveConfigMgr` helper in `final_coverage3_test.go` that could exercise it. Not blocking.

PR #7709, opened 2026-04-14
Fix preflight checks skipped when bicep snapshot fails

### LINKED ISSUE #7540: Authorization failed for template resource error with does not have permission to perform action 'Microsoft.Authorization/roleAssignments/write'
I provided --project-id during init command, and during azd provision, this error: 

```
Validation Error Details:
InvalidTemplateDeployment: Deployment failed with multiple errors: 'Authorization failed for 
template resource '111347e3-92b7-52c8-8229-215c6391bc4a' of type 'Microsoft.Authorization/roleAssignments'. The client '<user>@microsoft.com' with object id '<guid>' does not have permission to perform action 'Microsoft.Authorization/roleAssignments/write' at scope '/subscriptions/<sub-id>/resourceGroups/rg-echo-agent-streaming-dev/providers/Microsoft.Authorization/roleAssignments/111347e3-92b7-52c8-8229-215c6391bc4a'.:Authorization failed for template resource 'e8c7e996-5807-52ba-99c3-d0d0e5dc464b' of type 'Microsoft.Authorization/roleAssignments'. 
```

glharper:
## Problem

Fixes #7540

When `azd ai agent init --project-id <id>` is used, additional environment variables are set that become Bicep parameters. These parameter combinations cause `bicep snapshot` to fail — the snapshot engine is more sensitive to certain parameter values than the actual ARM deployment engine.

The current code in `local_preflight.go` silently drops **all** preflight checks when snapshot fails:

```go
data, err := l.bicepCli.Snapshot(ctx, bicepParamFile, snapshotOpts)
if err != nil {
    log.Printf("local preflight: skipping checks, bicep snapshot unavailable: %v", err)
    return nil, nil  // ← ALL checks silently dropped
}
```

This means the role assignment permission check (`checkRoleAssignmentPermissions`) never runs, and users hit `InvalidTemplateDeployment` errors during `azd up` with no warning.

## Fix

When bicep snapshot fails, instead of skipping all checks, fall back to recursively analyzing the compiled ARM template JSON (already available from `bicep build\ / \parseTemplate`). The fallback:

1. Saves the parsed ARM template from `parseTemplate()` (previously discarded with `_`)
2. When snapshot fails, calls `analyzeArmTemplateResources()` which recursively walks the template resource tree
3. Detects `Microsoft.Authorization/roleAssignments` in:
   - Direct top-level resources
   - Child resources nested inside parent declarations
   - Nested `Microsoft.Resources/deployments` (Bicep modules with inline templates)
4. Populates `Props.HasRoleAssignments` so structural checks still run

The `ValidationContext.SnapshotResources` and `ResourcesSnapshot` fields remain nil in the fallback path. This is safe because `checkRoleAssignmentPermissions` only reads `Props.HasRoleAssignments`.

## Testing

Added `TestAnalyzeArmTemplateResources` with 8 test cases covering:
- Empty resources
- No role assignments
- Direct role assignments
- Role assignments in child resources
- Role assignments in nested deployments
- Deeply nested deployments (deployment within deployment)
- Case-insensitive matching
- All existing tests continue to pass


cli/azd/pkg/infra/provisioning/bicep/local_preflight.go
```diff
@@ -408,26 +408,33 @@ func (l *localArmPreflight) validate(
...
-		ResourcesSnapshot: json.RawMessage(data),
-		SnapshotResources: snapshot.PredictedResources,
+		props := analyzeResources(snapshot.PredictedResources)
+		valCtx = &validationContext{
+			Console:           console,
+			Props:             props,
+			ResourcesSnapshot: json.RawMessage(data),
+			SnapshotResources: snapshot.PredictedResources,
+		}
 	}
```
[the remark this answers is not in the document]
glharper: @copilot apply changes based on [this feedback](https://github.com/Azure/azure-dev/pull/7709#discussion_r3080456505)

jongio (approved):
Solid fix. I verified that `SnapshotResources` and `ResourcesSnapshot` are never read by any registered check - only `Props.HasRoleAssignments` is consumed - so the nil values in the fallback path are safe. The recursive ARM template walker handles the key nesting patterns (direct, child, nested deployments) correctly, and `parseTemplate()` guarantees `tmpl.Resources` is non-empty before the fallback runs.

One minor note: there's no test that exercises the fallback wiring in `validate()` itself (where `Snapshot()` fails and code branches to the ARM template analysis). This is consistent with how this package tests today, so not blocking.

glharper:

vhvb1989 (changes requested):
please do not merge just jet.
I would like to take a look today and understand why snapshoot is failing -

cli/azd/pkg/infra/provisioning/bicep/bicep_provider.go
```diff
@@ -2307,6 +2307,13 @@ func (p *BicepProvider) validatePreflight(
 				return true, nil
 			}
 			p.setPreflightOutcome(span, preflightOutcomeWarningsAccepted, diagnosticIDs)
+			// Skip the ARM server-side validation when the user has acknowledged
+			// warnings. The local checks already surfaced the issues and the user
+			// chose to proceed despite them; running the ARM ValidatePreflight
+			// would re-surface the same issues as hard errors (e.g., missing
+			// Microsoft.Authorization/roleAssignments/write returns an
+			// InvalidTemplateDeployment error from the server-side validation).
+			return false, nil
```
jongio: This skips ARM `ValidatePreflight` for all accepted warnings, not just role-assignment ones. If a user accepts an AI quota warning, ARM validation (which catches entirely different issues like template syntax, resource provider registration, etc.) also gets skipped.

Consider gating the skip on the specific diagnostic IDs that are known to duplicate ARM errors - e.g. only skip when the accepted warnings are all `role_assignment_*` diagnostics.

cli/azd/pkg/infra/provisioning/bicep/local_preflight.go
```diff
@@ -650,6 +663,65 @@ func analyzeResources(resources []armTemplateResource) resourcesProperties {
 	return props
 }
 
+// analyzeArmTemplateResources recursively inspects the compiled ARM template resource tree
+// (from bicep build) and returns a resourcesProperties summarizing key characteristics.
+// Unlike analyzeResources (which operates on the flat snapshot list), this function walks
+// the template's nested resource hierarchy, including resources inside
+// Microsoft.Resources/deployments (Bicep modules), to detect role assignments even when
+// the bicep snapshot command is unavailable.
+func analyzeArmTemplateResources(resources armTemplateResources) resourcesProperties {
```
jongio, line 682: Two things worth noting in a comment here:

1. This intentionally only extracts `HasRoleAssignments`, not `CognitiveDeployments`, since cognitive deployment extraction needs resolved parameter values only available from the snapshot.
2. The structural walk doesn't evaluate `condition` or `copy` expressions, so it can report `HasRoleAssignments=true` for role assignments guarded by `condition: false`.

Both are acceptable trade-offs (over-warn is better than silent-skip), but documenting them would help future maintainers.

jongio (commented):
Revisiting after 4 commits since my earlier approval. The core fallback approach is solid - analyzing the compiled ARM template when snapshot fails is the right degradation strategy. A few items worth addressing:

cli/azd/pkg/infra/provisioning/bicep/local_preflight.go
glharper: [Updated](https://github.com/Azure/azure-dev/pull/7709/commits/c8295ad304132725c28d5e35d3f20c5ea4bc82ea)

cli/azd/pkg/infra/provisioning/bicep/bicep_provider.go
glharper: [Updated](https://github.com/Azure/azure-dev/pull/7709/commits/c8295ad304132725c28d5e35d3f20c5ea4bc82ea)

PR #7713, opened 2026-04-14
azure.ai.agents - detect token protection errors during RBAC setup

### LINKED ISSUE #7514: Post-Deploy RBAC Hook Fails, No Pre-Flight Auth Check
Steps to reproduce:

Run azd deploy
Exact output:
(✓) Done: Deploying service echo-agent-responses-streaming
  - Agent endpoint: https://...

ERROR: failed invoking event handlers for 'postdeploy'
agent identity RBAC setup failed: failed to discover agent identity:
AzureDeveloperCLICredential: please run "azd auth login"

Impact: Deploy reports (✓) Done before the hook runs. Agent is silently broken. Expected: Pre-flight auth check before deploy, and mark RBAC hook failure as a blocking error:
RBAC setup failed. Run: azd auth login && azd deploy

JeffreyCA:
Fixes #7514

## Issue

Some users (particularly internal Microsoft users) can authenticate with `azd auth login` and acquire ARM tokens, but Microsoft Graph token requests fail with `AADSTS530084` — a Conditional Access token protection policy block. Browser-based `azd auth login` can't satisfy this policy (it requires brokered auth like WAM, not yet supported by azd).

In hosted-agent flows, this causes `azd deploy` to fail during post-deploy RBAC setup with a misleading error telling the user to re-run `azd auth login`, which won't help.

## Summary

This PR adds detection of `AADSTS530084` in the post-deploy agent identity RBAC path, specifically in the Graph API fallback used when the agent version API does not return a principal ID. PR #7747 updated the RBAC flow to pull identity info directly from the agent API, but some edge cases still fall back to Graph discovery — this is where the token protection error can surface.

This is a targeted short-term fix scoped to the agents extension. Currently error suggestions do not render properly in event handlers and custom service targets (#7706), so for now the suggestion is included with the error message in red.

image

## Why the `azidentity` bump

Bumps `azidentity` from v1.13.1 to v1.14.0-beta.3, which includes [Azure/azure-sdk-for-go#25984](https://github.com/Azure/azure-sdk-for-go/pull/25984). Previously, `AzureDeveloperCLICredential` wouldn't return the full Entra error details, making string-based detection of AADSTS codes unreliable.


wbreza (commented):
## Code Review — Detect token protection errors in RBAC pre-flight

### TL;DR
Pragmatic short-term fix for a real user-facing problem. The implementation works, but it's patching over core gaps that should be addressed so future extension developers don't have to reimplement auth error detection.

### 🟡 Core Gap: Extension developers shouldn't have to detect AADSTS errors

The broader concern here is **what this pattern means for the extension ecosystem**. Today it's the agents extension handling AADSTS530084. Tomorrow it could be zure.appservice, zure.ai.finetune, or any third-party extension hitting the same Graph/ARM token protection error — each reimplementing detection and guidance independently.

**Core already has the infrastructure for this:**
- pkg/auth/errors.go detects AADSTS codes 70043, 700082, 50005 via ReLoginRequiredError and generates structured guidance
- rror_suggestions.yaml has a generic AADSTS catch-all that maps to "run azd auth login"
- Error middleware processes these through the suggestion pipeline

**What's missing for extension developers:**
1. **530084 isn't in the core AADSTS handler** — Adding it to pkg/auth/errors.go alongside existing codes would protect ALL extensions automatically. Unlike 70043/700082 where re-login helps, 530084 needs different guidance ("contact IT admin / request policy exception") since re-login won't resolve a Conditional Access policy block.

2. **Suggestion rendering doesn't work for extension event handlers** (#7706) — Extension errors from gRPC event handlers arrive as zdext.LocalError, but core error middleware only checks for internal.ErrorWithSuggestion. This forces extension authors to embed suggestions in error text instead of using structured errors. Fixing this unblocks the rror_suggestions.yaml pipeline for all extensions.

3. **Beta zidentity dependency** — Needed here because the stable version doesn't surface full AADSTS error details from AzureDeveloperCLICredential. If detection moved to core's MSAL layer (where the full AadErrorResponse is already parsed), no beta dependency would be needed in extensions.

**Suggested path forward:**
- Add AADSTS530084 to core's auth error handling (pkg/auth/errors.go) — with guidance that does NOT say "re-login" (since re-login won't help for token protection)
- Fix #7706 so extension structured errors render suggestions properly
- Track moving back to stable zidentity once v1.14.0 GA releases
- Once core handles it, remove the extension-side detection

### 🟡 Beta dependency risk
zidentity v1.14.0-beta.3 in production auth paths is risky. Beta APIs can change. Please file a tracking issue to move to stable once v1.14.0 GA releases.

### ✅ What looks good
- Error detection logic correctly targets AADSTS530084
- New CodeTokenProtectionBlocked error code is clean
- Actionable guidance with links and skip flag (AZD_AGENT_SKIP_ROLE_ASSIGNMENTS)
- Properly scoped to extension go.mod — no core dependency changes

**Verdict:** Acceptable as a short-term unblock. Please file tracking issues for the core gaps so this doesn't become permanent and future extension developers benefit from centralized handling. 👍

cli/azd/extensions/azure.ai.agents/go.mod
```diff
@@ -4,7 +4,7 @@ go 1.26.1
 
 require (
 	github.com/Azure/azure-sdk-for-go/sdk/azcore v1.21.0
-	github.com/Azure/azure-sdk-for-go/sdk/azidentity v1.13.1
+	github.com/Azure/azure-sdk-for-go/sdk/azidentity v1.14.0-beta.3
```
jongio, line 7: The beta is justified since v1.14.0 stable hasn't shipped yet, but it'd be good to have a tracking mechanism so this doesn't stay on beta indefinitely. A `// TODO: upgrade to stable azidentity v1.14.0 when released` above this line would make it easy to find later, or a brief note in the issue/project tracking.

cli/azd/extensions/azure.ai.agents/internal/project/developer_rbac_check.go
```diff
@@ -122,6 +122,24 @@ func CheckDeveloperRBAC(ctx context.Context, azdClient *azdext.AzdClient) error
 
 	userProfile, err := graphClient.Me().Get(ctx)
 	if err != nil {
+		// Short-term fix: detect Conditional Access token protection blocks (AADSTS530084)
+		// that prevent Graph token acquisition. Browser-based azd auth login won't resolve
+		// this — surface actionable guidance instead of silently skipping.
+		// TODO: Move detection to azd core auth layer for broader coverage.
+		errMsg := err.Error()
+		if strings.Contains(errMsg, "AADSTS530084") ||
```
jongio: The `AADSTS530084` code is the stable, machine-readable identifier and should be sufficient on its own. The secondary `"token protection policy"` check matches on natural-language text that could change across SDK versions or locales. Consider dropping it to reduce maintenance surface - if the error code isn't in the message, the prose text matching is unreliable anyway.

jongio (commented):
Solid, targeted fix. The early detection of AADSTS530084 is the right approach - failing fast with actionable guidance is much better than the silent skip that leads to confusing failures downstream.

One note on existing feedback: the Copilot bot suggested using `azidentity v1.14.0` (stable), but that version doesn't exist yet. Latest stable is v1.13.1. The beta.3 bump is the correct choice here.

vhvb1989 (approved):

cli/azd/extensions/azure.ai.agents/go.mod
JeffreyCA: Created https://github.com/Azure/azure-dev/issues/7732

trangevi (approved):

jongio (approved):
Looks good. The beta tracking issue (#7732) addresses my earlier feedback. The secondary string check concern is minor for a short-term fix - the AADSTS code alone is the reliable signal.

cli/azd/extensions/azure.ai.agents/internal/project/developer_rbac_check.go
JeffreyCA: Will keep it for now as there may be other undocumented codes related to token protection

PR #7719, opened 2026-04-14
update readme to use brew --cask install

hemarina:
While implementing azd update feature, @danieljurek and I were discussing about using brew cask for azd instead of formula. Update readme and installation commend to use cask.

### DIFF
--- a/cli/azd/main.go
+++ b/cli/azd/main.go
@@ -297,7 +297,7 @@ func platformUpgradeText() string {
 	} else if runtime.GOOS == "darwin" {
 		switch installedBy {
 		case installer.InstallTypeBrew:
-			return "run:\nbrew uninstall --cask azure/azd/azd && brew install --cask azure/azd/azd"
+			return "run:\nbrew uninstall azure/azd/azd && brew install azure/azd/azd"
 		case installer.InstallTypeSh:
 			//nolint:lll
 			return "run:\ncurl -fsSL https://aka.ms/install-azd.sh | bash\n\nIf the install script was run with custom parameters, ensure that the same parameters are used for the upgrade. For advanced install instructions, see: https://aka.ms/azd/upgrade/mac"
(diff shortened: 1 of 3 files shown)


cli/azd/main.go
```diff
@@ -297,7 +297,7 @@ func platformUpgradeText() string {
 	} else if runtime.GOOS == "darwin" {
 		switch installedBy {
 		case installer.InstallTypeBrew:
-			return "run:\nbrew uninstall azd && brew install azure/azd/azd"
+			return "run:\nbrew uninstall --cask azure/azd/azd && brew install --cask azure/azd/azd"
```
danieljurek: Until most users have migrated over to the cask, we should keep `brew uninstall azd` as is.

hemarina: > Until most users have migrated over to the cask, we should keep `brew uninstall azd` as is.


```suggestion
			return "run:\nbrew uninstall azure/azd/azd && brew install azure/azd/azd"
```

cli/azd/main.go
```diff
@@ -297,7 +297,7 @@ func platformUpgradeText() string {
 	} else if runtime.GOOS == "darwin" {
 		switch installedBy {
 		case installer.InstallTypeBrew:
-			return "run:\nbrew uninstall azd && brew install azure/azd/azd"
+			return "run:\nbrew uninstall azure/azd/azd && brew install azure/azd/azd"
```
hemarina:
```suggestion
			return "run:\nbrew uninstall azd && brew install azure/azd/azd"
```

danieljurek (approved):

jongio (approved):
Looks good. All three remaining `brew install` references (both READMEs + CLI upgrade text) now use `--cask`, matching the update manager code that already does `brew install --cask azure/azd/azd`. No stale references left.

wbreza (approved):
## Code Review — update readme to use brew --cask install

Clean, straightforward update. All three rew install references (both READMEs + CLI upgrade text) now correctly use --cask.

### Findings

No issues. The uninstall command was correctly kept as rew uninstall azd per danieljurek's guidance — users migrating from formula need the non-cask uninstall.

### ✅ What looks good
- All three files updated consistently
- Uninstall migration path correctly preserved
- Matches the update manager code that already uses --cask

Nice work keeping the installation docs in sync! 👍

JeffreyCA (approved):

Issue #7435, opened 2026-04-02
Multi-Language Hook Support: Python, JavaScript, TypeScript, .NET


wbreza:
### Summary

Extend the `azd` hook system ΓÇö currently limited to Bash and PowerShell ΓÇö to support **JavaScript, TypeScript, Python, and C#/.NET** hook scripts. Phase 1 delivers the **extensible framework and Python support**. Subsequent phases add JS, TS, and .NET using the same abstractions.

### Goals

- **Zero-config for consumers:** hooks "just work" regardless of language ΓÇö auto-detect, auto-install deps, auto-build
- **Preferred-language for authors:** template authors write hooks in the language of their choice without shell wrappers
- **Backwards compatible:** existing Bash/PowerShell hooks, inline scripts, `shell:` field ΓÇö all unchanged
- **Extensible framework:** adding a new language means implementing one interface, not touching the pipeline

### Key Capabilities

| Capability | Description |
|-----------|-------------|
| **Auto-detect** | Infer language from file extension (`.py`, `.js`, `.ts`, `.cs`) |
| **Auto-install deps** | Automatically run `pip install`, `npm install`, `dotnet restore` |
| **Auto-build** | Build TypeScript/C# before execution |
| **Runtime validation** | Fail fast with `ErrorWithSuggestion` if runtime missing |
| **Explicit override** | `language:` YAML field for edge cases |

### Phasing

| Phase | Scope | PR | Status |
|-------|-------|----|--------|
| **Phase 1** | Foundation framework + Python support | #7451 | ✅ Complete |
| **Phase 2** | JavaScript support | #7626 | ✅ Complete |
| **Phase 3** | TypeScript support | #7626 | ✅ Complete |
| Phase 4 | C#/.NET support | #7623 | 🟠 Open |
### Related Issues

| Issue | Relationship |
|-------|-------------|
| #4384 | **Contributes to** ΓÇö eliminates python-dotenv hacks for hook scenarios |
| #7423 | **Complements** ΓÇö they handle ad-hoc, we handle lifecycle-bound |
| #3613 | **Extends** ΓÇö our `Language` field generalizes shell-name dispatch |
| #4740 | **Extends pattern** ΓÇö runtime validation generalized to all languages |
| #4560 | **Extends pattern** ΓÇö install guidance UX for new runtimes |
| #3978 | **Motivated by** ΓÇö addresses "reduce scripting burden" research finding |
[...]

Issue #7348, opened 2026-03-26
feat: support extension-defined hook names in telemetry logging


jongio:
## Problem

When logging hook names in telemetry (hooks.name attribute), we currently validate against a static list of 16 known built-in azd hook names (prebuild, postbuild, predeploy, etc.). Unknown hook names are SHA-256 hashed to prevent logging arbitrary user input as customer content.

However, **extensions can define custom hook names** via:
- `azure.yaml` project/service hook configurations
- Runtime event handlers registered through `azdext.ExtensionHost.WithProjectEventHandler()` / `WithServiceEventHandler()`

These extension-defined hooks are legitimate hook names that would be useful to see in telemetry in plain text, but they currently get hashed because they're not in the built-in known-names list.

## Current Behavior

- Built-in hook names (`prebuild`, `postdeploy`, etc.) → logged raw ✅
- Extension-defined hook names from `azure.yaml` → **hashed** (not ideal)
- Arbitrary user input (`azd hooks run asdf`) → **hashed** ✅

## Desired Behavior

Extension-defined hook names should also be logged in plain text when they can be validated as legitimate (i.e., defined in project config or registered by an extension).

## Options to Investigate

1. **Validate against project config at telemetry time** — Check `projectConfig.Hooks` and all `service.Hooks` maps. If the hook name exists there, log raw; otherwise hash. This covers extension hooks defined in `azure.yaml` but not runtime-only handlers.

2. **Extension hook name registry** — Provide an API for extensions to register their custom hook names, making them available for telemetry validation.

3. **Accept hashing for extension hooks** — If extension hook telemetry isn't critical, the current approach (hash unknown names) may be acceptable.

## Context

- PR: https://github.com/Azure/azure-dev/pull/7299
- Review comment: https://github.com/Azure/azure-dev/pull/7299/files#r2996272790
- Current implementation: `cli/azd/cmd/hooks.go` — `knownHookNames` map + `StringHashed` fallback

wbreza:
Lowering priority on this issue since custom hooks are not supported today in the extension framework.

JeffreyCA:
Agreed this should be considered low priority for now - "present in azure.yaml" may not be sufficient validation for raw telemetry, since they're in the end still user-controlled.

PR #7721, opened 2026-04-14
feat: add telemetry instrumentation for hook executions

### LINKED ISSUE #7720: Add telemetry instrumentation for hook executions
## Background & Motivation

The hooks execution path emits zero telemetry when hooks fire via middleware during `azd up`, `azd provision`, `azd deploy`, etc. With three new executor types recently shipped (JS, TS, dotnet), the team has no visibility into adoption, failure rates, or usage patterns. This issue adds per-hook telemetry spans to every hook execution, enabling data-driven decisions about the hooks feature.

## User Story

As an azd platform engineer, I want telemetry emitted for every hook execution so that I can measure hook adoption rates, executor kind distribution, and failure rates across the user base.

## Solution Approach

1. **Define telemetry field keys** in `internal/tracing/fields/fields.go`:
   - `HooksEventKey` (`hooks.event`) — lifecycle event name (e.g., "preprovision", "postdeploy")
   - `HooksLevelKey` (`hooks.level`) — "command" or "service"
   - `HooksKindKey` (`hooks.kind`) — executor type: sh, pwsh, python, js, ts, dotnet
   - `HooksResultKey` (`hooks.result`) — "success", "error", or "invalid"
   - All use `SystemMetadata` classification and `FeatureInsight` purpose

2. **Per-hook child spans** in `execHook()` via `tracing.Start(ctx, "hooks.exec")`:
   - Each hook execution gets its own span with all 4 attributes set via `span.SetAttributes()`
   - Span duration is captured automatically via deferred `span.EndWithStatus()`
   - Per-hook correlation: each span carries event + level + kind + result together
   - Multiple hooks per command produce multiple child spans (not flat attribute sets)

3. **`HookLevel` as a parameter** on `Invoke()` and `RunHooks()`:
   - Transient value at the call site — no struct state, no constructor change
   - Callers specify `HookLevelCommand` or `HookLevelService` when invoking hooks

4. **`HookResult` typed constants** — `HookResultSuccess`, `HookResultError`, `HookResultInvalid`:
   - Compile-time safe, no string literal typos
   - `HookResultInvalid` emitted when hook configuration fails validation

## Acceptance Criteria

- [ ] Every hook execution creates a child span with `hooks.event`, `hooks.level`, `hooks.kind`, and `hooks.result`
- [ ] Per-hook spans enable correlation of attributes and automatic duration tracking
- [ ] Service-level hooks emit `hooks.level = "service"`; command-level hooks emit `hooks.level = "command"`
- [ ] `hooks.kind` reflects the actual resolved executor (sh/pwsh/python/js/ts/dotnet)
- [ ] `hooks.result` is "success" on success, "error" on execution failure, and "invalid" on validation failure
- [ ] Multiple hooks per command produce separate child spans (not deduplicated flat sets)
- [ ] No change to `NewHooksRunner` constructor signature
- [ ] No change to existing hook execution behavior or error propagation
- [ ] `golangci-lint run ./...` passes with no new warnings

## Out of Scope

- Dashboards, alerting, or Application Insights queries
- Telemetry for hook configuration details (secrets, interactive mode, continue-on-error)
- Changes to existing `azd hooks run` telemetry (`hooks.name`, `hooks.type`)

## Related Issues

- Child of #7435 — Multi-Language Hook Support epic

wbreza:
## What Changed

Hook executions in azd now emit telemetry, closing a complete observability gap. Previously, the middleware hook path (used by `azd up`, `provision`, `deploy`, etc.) produced zero telemetry events, and the newer JS/TS/dotnet script executors shipped with no instrumentation at all. Operators and developers had no way to understand hook usage patterns, failure rates, or invalid configurations across azd workloads.

Every hook invocation now records four span attributes — `hooks.event` (e.g. `preprovision`), `hooks.level` (project vs. service), `hooks.kind` (shell, JS, TS, dotnet, etc.), and `hooks.result` (`Success`, `Error`, or `Invalid`) — giving full visibility into what hooks run, where they run, and whether they succeed. Multi-hook scenarios accumulate correctly via `AppendUsageAttributeUnique`, so a single command that fires several hooks reports all of them without overwriting.

## How It Works

Telemetry is injected at two layers. `HooksRunner.Invoke()` and `RunHooks()` accept a new `HookLevel` parameter (project or service) that flows through as a transient argument — no struct state is modified, keeping the runner concurrency-safe. After each hook executes, the runner sets the four attributes on the current span using `AppendUsageAttributeUnique` so multiple hooks within one command coexist cleanly.

Validation telemetry is captured in `validate()` in `models.go`, which is the single chokepoint for all hook config parsing. Any invalid hook configuration (bad shell type, missing script, etc.) is recorded as `HookResultInvalid` before the error propagates, ensuring every caller gets validation telemetry for free. `HookResult` uses typed string constants (`HookResultSuccess`, `HookResultError`, `HookResultInvalid`) rather than raw strings. Hook names are well-known command derivatives (e.g. `preprovision`, `postdeploy`) so they are recorded directly without hashing. Tests (`TestHooksRunner_Telemetry`) cover the Success, Error, and Invalid result paths as table-driven subtests.

## Issue References

Resolves #7720
Relates to #7435

## Notes

The diff includes unrelated recording file churn and other changes from the base branch. The hooks telemetry changes are isolated to `fields/fields.go`, `hooks_runner.go`, `hooks_runner_test.go`, `models.go`, `cmd/hooks.go`, `cmd/middleware/hooks.go`, and `internal/cmd/provision.go`.

### DIFF
--- a/cli/azd/internal/cmd/provision.go
+++ b/cli/azd/internal/cmd/provision.go
@@ -551,7 +551,10 @@ func (p *ProvisionAction) runLayerProvisionWithHooks(
 
 	p.validateAndWarnLayerHooks(ctx, hooksManager, layer.Hooks)
 
-	if err := hooksRunner.Invoke(ctx, []string{string(project.ProjectEventProvision)}, actionFn); err != nil {
+	err := hooksRunner.Invoke(
+		ctx, []string{string(project.ProjectEventProvision)}, "layer", actionFn,
+	)
+	if err != nil {
 		if layer.Name == "" {
 			return err
 		}
--- a/cli/azd/internal/tracing/fields/fields.go
+++ b/cli/azd/internal/tracing/fields/fields.go
@@ -351,6 +351,12 @@ var (
 		Classification: SystemMetadata,
 		Purpose:        FeatureInsight,
 	}
+	// The executor kind used to run the hook (e.g., "sh", "pwsh", "python", "js", "ts", "dotnet").
+	HooksKindKey = AttributeKey{
+		Key:            attribute.Key("hooks.kind"),
+		Classification: SystemMetadata,
+		Purpose:        FeatureInsight,
+	}
 )
 
 // Pipeline command related fields
--- a/cli/azd/pkg/ext/hooks_runner.go
+++ b/cli/azd/pkg/ext/hooks_runner.go
@@ -10,6 +10,9 @@ import (
 	"os"
 	"strings"
 
+	"github.com/azure/azure-dev/cli/azd/internal/tracing"
+	"github.com/azure/azure-dev/cli/azd/internal/tracing/events"
+	"github.com/azure/azure-dev/cli/azd/internal/tracing/fields"
 	"github.com/azure/azure-dev/cli/azd/pkg/environment"
 	"github.com/azure/azure-dev/cli/azd/pkg/errorhandler"
 	"github.com/azure/azure-dev/cli/azd/pkg/exec"
@@ -18,6 +21,7 @@ import (
 	"github.com/azure/azure-dev/cli/azd/pkg/keyvault"
 	"github.com/azure/azure-dev/cli/azd/pkg/output"
 	"github.com/azure/azure-dev/cli/azd/pkg/tools"
+	"go.opentelemetry.io/otel/codes"
 )
 
 // HooksRunner enables support to invoke lifecycle hooks before & after
@@ -33,8 +37,8 @@ type HooksRunner struct {
 	serviceLocator ioc.ServiceLocator
 }
 
-// NewHooks creates a new instance of CommandHooks
-// When `cwd` is empty defaults to current shell working directory
+// NewHooksRunner creates a new instance of HooksRunner.
+// When `cwd` is empty defaults to current shell working directory.
 func NewHooksRunner(
 	hooksManager *HooksManager,
 	commandRunner exec.CommandRunner,
@@ -66,9 +70,12 @@ func NewHooksRunner(
 	}
 }
 
-// Invokes an action run runs any registered pre or post script hooks for the specified command.
-func (h *HooksRunner) Invoke(ctx context.Context, commands []string, actionFn InvokeFn) error {
-	err := h.RunHooks(ctx, HookTypePre, nil, commands...)
+// Invoke runs any registered pre and post script hooks around the specified action.
+// The hookType parameter identifies the hook scope for telemetry (project, layer, or service).
+func (h *HooksRunner) Invoke(
+	ctx context.Context, commands []string, hookType string, actionFn InvokeFn,
+) error {
+	err := h.RunHooks(ctx, HookTypePre, hookType, nil, commands...)
 	if err != nil {
 		return fmt.Errorf("failed running pre hooks: %w", err)
 	}
@@ -78,7 +85,7 @@ func (h *HooksRunner) Invoke(ctx context.Context, commands []string, actionFn In
 		return err
 	}
 
-	err = h.RunHooks(ctx, HookTypePost, nil, commands...)
+	err = h.RunHooks(ctx, HookTypePost, hookType, nil, commands...)
 	if err != nil {
 		return fmt.Errorf("failed running post hooks: %w", err)
 	}
@@ -86,14 +93,16 @@ func (h *HooksRunner) Invoke(ctx context.Context, commands []string, actionFn In
 	return nil
 }
 
-// Invokes any registered script hooks for the specified hook type and command.
+// RunHooks invokes any registered script hooks for the specified hook type and command.
+// The hookType parameter identifies the hook scope for telemetry (project, layer, or service).
 func (h *HooksRunner) RunHooks(
 	ctx context.Context,
-	hookType HookType,
+	ht HookType,
+	hookType string,
 	options *tools.ExecutionContext,
 	commands ...string,
 ) error {
-	hooks, err := h.hooksManager.GetByParams(h.hooks, hookType, commands...)
+	hooks, err := h.hooksManager.GetByParams(h.hooks, ht, commands...)
 	if err != nil {
 		return fmt.Errorf("failed running scripts for hooks '%s', %w", strings.Join(commands, ","), err)
 	}
@@ -103,7 +112,7 @@ func (h *HooksRunner) RunHooks(
 			return fmt.Errorf("reloading environment before running hook: %w", err)
 		}
 
-		err := h.execHook(ctx, hookConfig, options)
+		err := h.execHook(ctx, hookConfig, hookType, options)
 		if err != nil {
 			return err
 		}
@@ -117,8 +126,24 @@ func (h *HooksRunner) RunHooks(
 }
 
 func (h *HooksRunner) execHook(
-	ctx context.Context, hookConfig *HookConfig, options *tools.ExecutionContext,
-) error {
+	ctx context.Context, hookConfig *HookConfig, hookType string, options *tools.ExecutionContext,
+) (hookErr error) {
+	// Create a child span for this hook execution so each hook
+	// gets its own correlated set of telemetry attributes.
+	ctx, span := tracing.Start(ctx, events.HooksExecEvent)
+	statusCode := ""
+	defer func() {
+		if hookErr != nil {
+			if statusCode == "" {
+				statusCode = "hook.failed"
+			}
+			span.SetStatus(codes.Error, statusCode)
+		} else {
+			span.SetStatus(codes.Ok, "")
+		}
+		span.End()
+	}()
+
 	if options == nil {
 		options = &tools.ExecutionContext{}
 	}
@@ -143,15 +168,26 @@ func (h *HooksRunner) execHook(
 			return nil
 		})
 		if err != nil {
+			statusCode = "hook.secrets_failed"
 			return err
 		}
 	}
 
+	// Set name and type on span — known before validation.
+	span.SetAttributes(
+		fields.HooksNameKey.String(hookConfig.Name),
+		fields.HooksTypeKey.String(hookType),
+	)
+
 	// validate() resolves the hook's kind, path, shell type,
 	// and computes resolvedDir / resolvedScriptPath.
 	if err := hookConfig.validate(); err != nil {
+		statusCode = "hook.validation_failed"
 		return err
 	}
+
+	span.SetAttributes(
+		fields.HooksKindKey.String(string(hookConfig.Kind)))
 
 	// Use pre-resolved paths from validate().
 	cwd := hookConfig.resolvedDir
@@ -192,6 +228,7 @@ func (h *HooksRunner) execHook(
 	// Resolve executor via IoC — hooks runner has NO knowledge of executor internals.
 	var executor tools.HookExecutor
 	if err := h.serviceLocator.ResolveNamed(string(hookConfig.Kind), &executor); err != nil {
+		statusCode = "hook.executor_not_found"
 		return &errorhandler.ErrorWithSuggestion{
 			Err: fmt.Errorf(
 				"no executor for kind '%s': %w",
@@ -230,6 +267,7 @@ func (h *HooksRunner) execHook(
 	)
 
 	if err := executor.Prepare(ctx, scriptPath, execCtx); err != nil {
+		statusCode = "hook.prepare_failed"
 		return fmt.Errorf("preparing hook '%s': %w", hookConfig.Name, err)
 	}
 
@@ -246,11 +284,12 @@ func (h *HooksRunner) execHook(
 
 	res, err := executor.Execute(ctx, scriptPath, execCtx)
 	if err != nil {
-		hookErr := h.handleHookError(
+		execErr := h.handleHookError(
 			ctx, hookConfig, res, scriptPath, err,
 		)
-		if hookErr != nil {
-			return hookErr
+		if execErr != nil {
+			statusCode = "hook.execution_failed"
+			return execErr
 		}
 	}
 
--- a/cli/azd/pkg/ext/hooks_runner_test.go
+++ b/cli/azd/pkg/ext/hooks_runner_test.go
@@ -105,7 +105,7 @@ func Test_Hooks_Execute(t *testing.T) {
 			env,
 			mockContext.Container,
 		)
-		err := runner.RunHooks(*mockContext.Context, HookTypePre, nil, "command")
+		err := runner.RunHooks(*mockContext.Context, HookTypePre, "project", nil, "command")
 
 		require.True(t, ranPreHook)
 		require.False(t, ranPostHook)
@@ -143,7 +143,7 @@ func Test_Hooks_Execute(t *testing.T) {
 			env,
 			mockContext.Container,
 		)
-		err := runner.RunHooks(*mockContext.Context, HookTypePost, nil, "command")
+		err := runner.RunHooks(*mockContext.Context, HookTypePost, "project", nil, "command")
 
 		require.False(t, ranPreHook)
 		require.True(t, ranPostHook)
@@ -181,7 +181,7 @@ func Test_Hooks_Execute(t *testing.T) {
 			env,
 			mockContext.Container,
 		)
-		err := runner.RunHooks(*mockContext.Context, HookTypePre, nil, "interactive")
+		err := runner.RunHooks(*mockContext.Context, HookTypePre, "project", nil, "interactive")
 
 		require.False(t, ranPreHook)
 		require.True(t, ranPostHook)
@@ -213,7 +213,7 @@ func Test_Hooks_Execute(t *testing.T) {
 			env,
 			mockContext.Container,
 		)
-		err := runner.RunHooks(*mockContext.Context, HookTypePre, nil, "inline")
+		err := runner.RunHooks(*mockContext.Context, HookTypePre, "project", nil, "inline")
 
 		require.False(t, ranPreHook)
 		require.True(t, ranPostHook)
@@ -264,7 +264,7 @@ func Test_Hooks_Execute(t *testing.T) {
 			env,
 			mockContext.Container,
 		)
-		err := runner.Invoke(*mockContext.Context, []string{"command"}, func() error {
+		err := runner.Invoke(*mockContext.Context, []string{"command"}, "project", func() error {
 			ranAction = true
 			hookLog = append(hookLog, "action")
 			return nil
@@ -337,7 +337,7 @@ func Test_Hooks_Validation(t *testing.T) {
 			mockContext.Console, cwd, hooksMap, env, mockContext.Container,
 		)
 
-		err := runner.RunHooks(*mockContext.Context, HookTypePre, nil, "deploy")
+		err := runner.RunHooks(*mockContext.Context, HookTypePre, "project", nil, "deploy")
 		require.NoError(t, err)
 		require.True(t, shellRan)
 	})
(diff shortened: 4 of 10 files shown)


cli/azd/pkg/ext/hooks_runner.go
```diff
@@ -153,6 +170,14 @@ func (h *HooksRunner) execHook(
 		return err
 	}
 
+	// Emit hook telemetry after validation resolves the executor kind.
+	tracing.AppendUsageAttributeUnique(
+		fields.HooksEventKey.String(hookConfig.Name))
+	tracing.AppendUsageAttributeUnique(
+		fields.HooksLevelKey.String(string(h.level)))
+	tracing.AppendUsageAttributeUnique(
+		fields.HooksKindKey.String(string(hookConfig.Kind)))
```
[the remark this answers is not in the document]
wbreza: Fixed — added \TestHooksRunner_Telemetry\ with Success, Error, and Invalid subtests that assert all four usage attributes (\hooks.event\, \hooks.level\, \hooks.kind\, \hooks.result\).

cli/azd/pkg/ext/hooks_runner.go
```diff
@@ -245,12 +270,18 @@ func (h *HooksRunner) execHook(
 
 	res, err := executor.Execute(ctx, scriptPath, execCtx)
 	if err != nil {
+		tracing.AppendUsageAttributeUnique(
+			fields.HooksResultKey.String("error"))
```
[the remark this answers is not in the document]
wbreza: Fixed — added \TestHooksRunner_Telemetry\ with Success, Error, and Invalid subtests that assert all four usage attributes (\hooks.event\, \hooks.level\, \hooks.kind\, \hooks.result\).

cli/azd/pkg/ext/hooks_runner.go
```diff
@@ -245,12 +270,18 @@ func (h *HooksRunner) execHook(
...
 		hookErr := h.handleHookError(
 			ctx, hookConfig, res, scriptPath, err,
 		)
 		if hookErr != nil {
 			return hookErr
 		}
+	} else {
+		tracing.AppendUsageAttributeUnique(
+			fields.HooksResultKey.String("success"))
 	}
```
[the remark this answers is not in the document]
wbreza: Fixed — added \TestHooksRunner_Telemetry\ with Success, Error, and Invalid subtests that assert all four usage attributes (\hooks.event\, \hooks.level\, \hooks.kind\, \hooks.result\).

cli/azd/pkg/ext/hooks_runner.go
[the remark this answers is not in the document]
wbreza: Fixed — defined \HookResult\ type with \HookResultSuccess\, \HookResultError\, and \HookResultInvalid\ constants. All string literals replaced.

cli/azd/pkg/ext/hooks_runner.go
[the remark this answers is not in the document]
wbreza: Fixed — defined \HookResult\ type with \HookResultSuccess\, \HookResultError\, and \HookResultInvalid\ constants. All string literals replaced.

cli/azd/pkg/ext/hooks_runner.go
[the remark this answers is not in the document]
wbreza: Hook event names are always derived from well-known azd command names with pre/post prefixes (e.g., preprovision, postdeploy). They are not user-controlled, so hashing is not needed.

cli/azd/pkg/ext/hooks_runner.go
```diff
@@ -147,12 +181,23 @@ func (h *HooksRunner) execHook(
 		}
 	}
 
+	// Set event and level on span — known before validation.
+	span.SetAttributes(
+		fields.HooksEventKey.String(hookConfig.Name),
```
jongio: `cmd/hooks.go` hashes unknown hook names via `StringHashed` to avoid logging extension-defined names raw (see #7348). This logs `hookConfig.Name` unconditionally - extension hooks with custom names would bypass that privacy gate. Consider reusing the known-names allowlist or hashing here too.

cli/azd/pkg/ext/hooks_runner.go
```diff
@@ -147,12 +181,23 @@ func (h *HooksRunner) execHook(
...
+		fields.HooksLevelKey.String(string(level)),
+	)
+
 	// validate() resolves the hook's kind, path, shell type,
 	// and computes resolvedDir / resolvedScriptPath.
 	if err := hookConfig.validate(); err != nil {
+		span.SetAttributes(
+			fields.HooksResultKey.String(string(HookResultInvalid)))
```
jongio: `hooks.result` isn't set when `Prepare` fails (line ~277). Same gap exists for the `ResolveNamed` error (line ~238). The span ends with error status (via the deferred `EndWithStatus`), but telemetry queries filtering on `hooks.result` won't see these failures. Setting `hooks.result = "error"` before both returns would close the gap.

cli/azd/internal/tracing/fields/fields.go
```diff
@@ -351,6 +351,30 @@ var (
...
+		Classification: SystemMetadata,
+		Purpose:        FeatureInsight,
+	}
+	// The executor kind used to run the hook (e.g., "sh", "pwsh", "python", "js", "ts", "dotnet").
+	HooksKindKey = AttributeKey{
+		Key:            attribute.Key("hooks.kind"),
+		Classification: SystemMetadata,
+		Purpose:        FeatureInsight,
+	}
+	// The execution result of the hook: "success" or "error".
```
jongio: Comment says "success" or "error" but the code also emits "invalid" on validation failure. Should list all three values.

cli/azd/pkg/ext/hooks_runner_test.go
```diff
@@ -104,7 +104,7 @@ func Test_Hooks_Execute(t *testing.T) {
 			env,
 			mockContext.Container,
 		)
-		err := runner.RunHooks(*mockContext.Context, HookTypePre, nil, "command")
+		err := runner.RunHooks(*mockContext.Context, HookTypePre, HookLevelCommand, nil, "command")
```
jongio: The inline comment responses mention `TestHooksRunner_Telemetry` with Success/Error/Invalid subtests, but this test doesn't appear on the branch. The existing test updates only adjust the `HookLevel` parameter - they don't assert any telemetry attributes. The primary feature of this PR needs test coverage.

jongio (changes requested):
This PR adds useful observability to hooks, but a few things need attention:

1. The `TestHooksRunner_Telemetry` test mentioned in the inline comment responses doesn't exist on the branch. The primary feature (telemetry emission) has no assertion coverage - existing test updates just add the new `HookLevel` parameter.

2. Two error paths in `execHook` don't set `hooks.result`: executor resolution failure and `Prepare()` failure. Queries filtering on `hooks.result` will silently drop these errors.

3. `hooks.event` logs raw hook names, but `cmd/hooks.go` already hashes unknown names via `StringHashed` for privacy (see #7348). Extension hooks with custom names would bypass this.

4. The new `hooks.event`/`hooks.level` keys overlap semantically with existing `hooks.name`/`hooks.type`. Worth clarifying the relationship or consolidating.

cli/azd/pkg/ext/hooks_runner.go
```diff
@@ -245,12 +290,18 @@ func (h *HooksRunner) execHook(
 
 	res, err := executor.Execute(ctx, scriptPath, execCtx)
 	if err != nil {
-		hookErr := h.handleHookError(
+		span.SetAttributes(
+			fields.HooksResultKey.String(string(HookResultError)))
```
weikanglim: I think this should use the span status instead of a separate status attribute.

OpenTelemetry span status is only `Unset` / `Ok` / `Error`, so for a failed hook the signal here should be `span.SetStatus(codes.Error, "hook.execute_failed")`.

cli/azd/pkg/ext/hooks_runner.go
wbreza: Fixed — added \span.SetAttributes(fields.HooksResultKey.String(string(HookResultError)))\ before both the \ResolveNamed\ and \Prepare\ error returns. All error paths now set \hooks.result\.

cli/azd/pkg/ext/hooks_runner_test.go
wbreza: Fixed — re-added \TestHooksRunner_Telemetry\ with Success, Error, and ValidationFailure subtests that exercise all telemetry code paths including span creation and attribute setting.

cli/azd/internal/tracing/fields/fields.go
wbreza: Fixed — updated comment to list all three values: \success\, \error\, or \invalid\.

cli/azd/pkg/ext/hooks_runner.go
wbreza: Today extension developers cannot define custom hook names — the hook event names in the runner are always well-known azd command derivatives (pre/post + command name like \preprovision\, \postdeploy\). The \zd hooks run\ command is designed for testing hooks using these well-known names, not custom names. The hashing in \cmd/hooks.go\ covers the \zd hooks run\ CLI entry point as a defensive measure, but the runner itself only sees names from the azure.yaml config which are constrained to the known set.

If we extend to support custom hook names in the future, we'll add hashing here at that point.

cli/azd/internal/tracing/fields/fields.go
```diff
@@ -351,6 +351,30 @@ var (
...
 		Purpose:        FeatureInsight,
 	}
+	// The lifecycle event that triggered the hook (e.g., "preprovision", "postdeploy").
+	HooksEventKey = AttributeKey{
+		Key:            attribute.Key("hooks.event"),
+		Classification: SystemMetadata,
+		Purpose:        FeatureInsight,
+	}
+	// The scope level of the hook: "command" (project-level) or "service" (service-level).
+	HooksLevelKey = AttributeKey{
```
weikanglim: Take a look at the newly added `HooksTypeKey` and `HooksNameKey`, see what makes sense to merge here.

cli/azd/internal/tracing/fields/fields.go
```diff
@@ -351,6 +351,30 @@ var (
...
+		Purpose:        FeatureInsight,
+	}
+	// The executor kind used to run the hook (e.g., "sh", "pwsh", "python", "js", "ts", "dotnet").
+	HooksKindKey = AttributeKey{
+		Key:            attribute.Key("hooks.kind"),
+		Classification: SystemMetadata,
+		Purpose:        FeatureInsight,
+	}
+	// The execution result of the hook: "success", "error", or "invalid".
+	HooksResultKey = AttributeKey{
```
weikanglim: Remove this attribute in favor of `span.SetStatus(codes.Error, "<different error conditions>").`

cli/azd/pkg/ext/hooks_runner.go
```diff
@@ -117,8 +146,13 @@ func (h *HooksRunner) RunHooks(
 }
 
 func (h *HooksRunner) execHook(
-	ctx context.Context, hookConfig *HookConfig, options *tools.ExecutionContext,
-) error {
+	ctx context.Context, hookConfig *HookConfig, level HookLevel, options *tools.ExecutionContext,
+) (hookErr error) {
+	// Create a child span for this hook execution so each hook
+	// gets its own correlated set of telemetry attributes.
+	ctx, span := tracing.Start(ctx, "hooks.exec")
```
weikanglim: Events need to be registered in tracing/events/events.go for classification purposes

cli/azd/pkg/ext/hooks_runner.go
```diff
@@ -192,6 +236,9 @@ func (h *HooksRunner) execHook(
 	// Resolve executor via IoC — hooks runner has NO knowledge of executor internals.
 	var executor tools.HookExecutor
 	if err := h.serviceLocator.ResolveNamed(string(hookConfig.Kind), &executor); err != nil {
+		span.SetAttributes(
+			fields.HooksResultKey.String(string(HookResultError)))
+
```
weikanglim: There's room for improvement here -- think broadly about failure modes, I see "env expansion", "validation failed" , "prepare failed", and "execution failed" as possible breakdowns.

weikanglim (commented):
Overall looks good -- some minor cleanup comments

cli/azd/internal/tracing/fields/fields.go
wbreza: Fixed — removed \HooksEventKey\ and reuse existing \HooksNameKey\ on the span instead. \HooksLevelKey\ is kept since it has different dimensionality than \HooksTypeKey\ (2-value command/service vs 3-value project/layer/service).

cli/azd/pkg/ext/hooks_runner.go
wbreza: Fixed — removed \HooksResultKey\ attribute. The deferred \span.EndWithStatus(hookErr)\ now handles status automatically: \codes.Ok\ on success, \codes.Error\ on any failure (validation, prepare, execute).

cli/azd/internal/tracing/fields/fields.go
wbreza: Fixed — removed \HooksResultKey\ and the \HookResult\ type/constants entirely. Span status via \EndWithStatus\ replaces it.

cli/azd/pkg/ext/hooks_runner.go
wbreza: Fixed — registered \HooksExecEvent = \hooks.exec\\ in \	racing/events/events.go\ and updated \xecHook()\ to use the registered constant.

cli/azd/pkg/ext/hooks_runner.go
wbreza: Good call — added `hooks.failure_mode` span attribute that captures the specific failure stage: `secrets_failed`, `validation_failed`, `executor_not_found`, `prepare_failed`, or `execution_failed`. This gives queryable breakdown of where hooks fail without conflicting with the OTel span status.

wbreza: Good call — added specific failure mode descriptions to the span status following the existing dot-separated convention used elsewhere in the codebase (e.g., `agent.fix.failed`). Each error path now sets a descriptive status code:

- `hook.secrets_failed` — KeyVault secret resolution error
- `hook.validation_failed` — hook config validation error
- `hook.executor_not_found` — unsupported hook kind
- `hook.prepare_failed` — executor preparation error (venv, deps, etc.)
- `hook.execution_failed` — script execution error

cli/azd/internal/cmd/provision.go
```diff
@@ -551,7 +551,10 @@ func (p *ProvisionAction) runLayerProvisionWithHooks(
 
 	p.validateAndWarnLayerHooks(ctx, hooksManager, layer.Hooks)
 
-	if err := hooksRunner.Invoke(ctx, []string{string(project.ProjectEventProvision)}, actionFn); err != nil {
+	err := hooksRunner.Invoke(
+		ctx, []string{string(project.ProjectEventProvision)}, ext.HookLevelCommand, actionFn,
+	)
```
weikanglim: Should this be per-layer?

cli/azd/internal/tracing/fields/fields.go
weikanglim: Can you clarify what the difference is between "The type of the hook run scope " and " The scope level of the hook"? They seem to overlap pretty significantly, and I'm curious what a clean design looks like here.

cli/azd/pkg/ext/hooks_runner.go
weikanglim: Reuse the events constant

weikanglim (approved):
LGTM, minor comments -- nothing blocker

weikanglim (approved):
LGTM, minor comments -- nothing blocker

jongio (commented):
Core approach looks solid - span-per-hook with status codes is the right design. My original CHANGES_REQUESTED concerns around result gaps are addressed. Three items remain:

cli/azd/pkg/ext/hooks_runner.go
```diff
@@ -117,8 +135,24 @@ func (h *HooksRunner) RunHooks(
 }
 
 func (h *HooksRunner) execHook(
-	ctx context.Context, hookConfig *HookConfig, options *tools.ExecutionContext,
-) error {
+	ctx context.Context, hookConfig *HookConfig, level HookLevel, options *tools.ExecutionContext,
+) (hookErr error) {
+	// Create a child span for this hook execution so each hook
+	// gets its own correlated set of telemetry attributes.
+	ctx, span := tracing.Start(ctx, "hooks.exec")
```
jongio: +1 to @weikanglim's thread - this is still a string literal. Every other `tracing.Start` call in the codebase uses event constants (e.g. `events.CopilotInitializeEvent`, `events.PreflightValidationEvent`). Should be `events.HooksExecEvent` which already exists in `tracing/events/events.go`.

cli/azd/internal/cmd/provision.go
```diff
@@ -551,7 +551,10 @@ func (p *ProvisionAction) runLayerProvisionWithHooks(
...
+	if err != nil {
```
jongio, line 557: Nit: this line uses spaces instead of tabs - breaks the file's indentation.

cli/azd/pkg/ext/hooks_runner_test.go
```diff
@@ -105,7 +105,7 @@ func Test_Hooks_Execute(t *testing.T) {
...
```
jongio: Nice addition. One gap: these tests verify no-panic/no-error but don't assert the actual telemetry attributes (name, level, kind, span status). `telemetry_coverage_test.go` shows the pattern for attribute assertions. Worth adding at least a happy-path check that the span carries the expected attributes.

jongio (approved):
Approach is sound - unblocking. The three inline comments from my last review still apply (event constant, indentation, test assertions), and @weikanglim's open threads on level/type consolidation and per-layer scope should be resolved before merge.

cli/azd/internal/tracing/fields/fields.go
wbreza: You're right — they overlapped. Removed HooksLevelKey entirely and consolidated into the existing HooksTypeKey. The span now carries hooks.type with values "project", "layer", or "service" — matching the existing telemetry in cmd/hooks.go. The HookLevel type and constants are gone.

cli/azd/pkg/ext/hooks_runner.go
wbreza: Fixed — replaced the string literal with vents.HooksExecEvent constant.

cli/azd/pkg/ext/hooks_runner_test.go
wbreza: The codebase doesn't have a test span exporter pattern for capturing and asserting span attributes in unit tests. The existing subtests (Success, Error, ValidationFailure) exercise all telemetry code paths and verify correct execution. Adding a test exporter infrastructure would be a larger cross-cutting effort — happy to pick that up as a follow-up if there's a pattern you'd like to establish.

cli/azd/internal/cmd/provision.go
wbreza: Good question — yes, it now passes "layer" as the hook type, so layer hooks are distinguished from project-level hooks in telemetry. Each layer's hooks run through their own Invoke() call with hookType = "layer".

cli/azd/internal/cmd/provision.go
wbreza: Fixed — verified consistent tab indentation throughout the file.

cli/azd/pkg/ext/hooks_runner.go
wbreza: Fixed — now using vents.HooksExecEvent constant instead of the string literal.

rajeshkamal5050 (approved):

PR #7722, opened 2026-04-14
Update CODEOWNERS for azure.ai.agents

### LINKED ISSUE #7726: Adjust codeowners for agents extension

trangevi:
Adjust agents extension code owners to specifically be the agents team

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -11,7 +11,7 @@
 /cli/installer/                             @jongio @wbreza @danieljurek @vhvb1989 @tg-msft @rajeshkamal5050
 
 # ── CLI extensions (AI) ──────────────────────────────────────────────────────
-/cli/azd/extensions/azure.ai.agents/        @jongio @wbreza @vhvb1989 @hemarina @weikanglim @JeffreyCA @tg-msft @rajeshkamal5050 @trangevi @trrwilson @therealjohn
+/cli/azd/extensions/azure.ai.agents/        @trangevi @trrwilson @therealjohn @glharper
 /cli/azd/extensions/azure.ai.finetune/      @jongio @wbreza @vhvb1989 @hemarina @weikanglim @JeffreyCA @tg-msft @rajeshkamal5050 @trangevi @achauhan-scc @kingernupur @rabollin @saanikaguptamicrosoft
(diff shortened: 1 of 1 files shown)


.github/CODEOWNERS
```diff
@@ -11,7 +11,7 @@
 /cli/installer/                             @jongio @wbreza @danieljurek @vhvb1989 @tg-msft @rajeshkamal5050
 
 # ── CLI extensions (AI) ──────────────────────────────────────────────────────
-/cli/azd/extensions/azure.ai.agents/        @jongio @wbreza @vhvb1989 @hemarina @weikanglim @JeffreyCA @tg-msft @rajeshkamal5050 @trangevi @trrwilson @therealjohn
+/cli/azd/extensions/azure.ai.agents/        @trangevi @trrwilson @therealjohn @glharper
```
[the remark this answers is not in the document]
trangevi: We don't currently have a team for this, will take it as a follow up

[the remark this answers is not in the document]
danieljurek: Teams are a nice to have. They're not material to the azd CODEOWNERS file at this time.

danieljurek, line 14: new codeowner is valid according to `Validate-AzsdkCodeOwner.ps1`

danieljurek (approved):

.github/CODEOWNERS
jongio, line 14: I prefer to keep azd folks on this and add foundry folks instead of remove azd

trangevi: My main concern is about broader directional understanding for the extension. Code review from the azd team is valuable and desirable, but as more and more things get added, I would assume that it won't be your role to make sure that changes are following the desired direction for the _extension_, more so for azd all up. So I'm concerned that a change will be proposed, reviewed and approved by azd folks, and merged, before foundry folks are even aware of it (I already saw it with one of Glenn's PRs, which got me thinking of this).

Since GH doesn't have a concept of required and optional reviewers, I couldn't think of another way to address the concern other than removing azd folks as codeowners, and instead manually requesting reviews from you, while leaving foundry folks as the final sign off. If there's a better way to accomplish this I'm very much open to it, as I definitely don't want to lose the current collaboration.

rajeshkamal5050: We discussed on this Travis, we are good with this.

suggestion: add @JeffreyCA as a co-owner since he's been closely working with your team on this?

rajeshkamal5050 (approved):

trangevi:
Closing for now, pending further discussion

PR #7729, opened 2026-04-14
[Agents Extension] agent run fix for .net

### LINKED ISSUE #7688: azd ai agent run on .net projects incorrectly says port 8088
Image

Noticed using the azd ai agent run saying it was in port 8088 but actually the host is running in 5000

trangevi:
Fixes #7688

Adds proper env var for local .net run

### DIFF
--- a/cli/azd/extensions/azure.ai.agents/internal/cmd/run_test.go
+++ b/cli/azd/extensions/azure.ai.agents/internal/cmd/run_test.go
@@ -4,6 +4,7 @@
 package cmd
 
 import (
+	"fmt"
 	"os"
 	"path/filepath"
 	"runtime"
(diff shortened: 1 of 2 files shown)


JeffreyCA (approved):

jongio (commented):
The fix is correct - .NET projects need `ASPNETCORE_URLS` to tell Kestrel which port to bind to, since it ignores the `PORT` env var. Moving `detectProjectType` earlier so it's available for both command detection and env setup is the right approach.

One concern: the tests re-implement the env-building logic inline rather than testing the actual production code path. If the `ASPNETCORE_URLS` block were removed from `runRun()`, all 4 subtests would still pass.

Nit: `detectProjectType(projectDir)` is now called twice per run - once at the top of `runRun` and again inside `installDependencies`. Consider passing `pt` to avoid the redundant file checks.

cli/azd/extensions/azure.ai.agents/internal/cmd/run_test.go
```diff
@@ -336,3 +337,135 @@ func TestAppendFoundryEnvVars(t *testing.T) {
...
+			t.Fatal(err)
+		}
+
+		pt := detectProjectType(dir)
+		if pt.Language != "dotnet" {
+			t.Fatalf("expected dotnet, got %s", pt.Language)
+		}
+
+		port := 8088
+		var env []string
```
jongio: These subtests duplicate the production logic (`if pt.Language == "dotnet" { env = append... }`) rather than calling a production function. They verify the copy-pasted logic works, but don't catch regressions in `runRun`.

Consider extracting the port env-building from `runRun` into a testable helper - similar to how `appendFoundryEnvVars` is already extracted and tested directly in this file.

JeffreyCA (approved):

Issue #7736, opened 2026-04-15
Script provisioning provider — full provider implementation


wbreza:
## Script provisioning provider -- full provider implementation

**Parent epic**: #7733
**Framework**: #7465 | **Depends on**: #7734 (scaffold), #7735 (config parsing)

### Overview

Implement all `azdext.ProvisioningProvider` lifecycle methods with real behavior, including the three core components: `EnvResolver`, `ScriptExecutor`, and `OutputCollector`.

#### 1. EnvResolver (`internal/provisioning/env_resolver.go`)

Resolves environment variables for each script using the 4-layer priority model:

```
Layer 4 (highest): secrets: map values (Key Vault refs, prompted values)
Layer 3:           env: map values (${EXPRESSION} substitution)
Layer 2:           azd environment values (.env + prior script outputs)
Layer 1 (lowest):  OS environment variables
```

Key behaviors:
- `${EXPRESSION}` substitution via `osutil.ExpandableString` (or `drone/envsubst`)
- Missing variable detection with actionable error messages (`azd env set` guidance)
- Secret resolution: `akvs://` Key Vault refs -> azd env lookup -> interactive prompt -> persist to env
- Prior script outputs flow as env vars to subsequent scripts

#### 2. ScriptExecutor (`internal/provisioning/executor.go`)

Runs individual shell scripts with a prepared environment:

- Shell invocation: `bash <script>` or `pwsh -NoProfile -NonInteractive -File <script>`
- Working directory: project root (directory containing `azure.yaml`)
- Streams stdout/stderr to console in real-time
- Captures exit code; non-zero fails unless `continueOnError: true`
- Path containment check before execution (no escaping project root)

```go
type ScriptResult struct {
    ExitCode int
    Stdout   string
    Stderr   string
}
```

#### 3. OutputCollector (`internal/provisioning/output.go`)

Discovers and parses `outputs.json` files after each script:

- **Search strategy**: walk up from script directory toward project root, stop at first `outputs.json`
- **Format**: `{ "outputs": { "KEY": { "type": "string", "value": "..." } } }`
- **Merge**: outputs accumulate across scripts (last-write-wins for duplicate keys)
- **Missing file**: not an error (scripts may have no outputs)

#### `Initialize(ctx, projectPath, options) error`
- Parse config via `ParseProviderConfig()` (from #7735)
- Validate script paths, kinds, shell availability
- Store project path, options, and parsed config

[...]

Issue #7735, opened 2026-04-15
Script provisioning provider — config design and parsing


wbreza:
## Script provisioning provider -- config design and parsing

**Parent epic**: #7733
**Framework**: #7465 | **Depends on**: #7734 (scaffold)

### Overview

Implement configuration parsing from `ProvisioningOptions.Config` (a `google.protobuf.Struct`) into typed Go structs, plus validation logic that runs during `Initialize()`.

### Configuration Model

Users configure the script provider in `azure.yaml`:

```yaml
infra:
  provider: scripts
  config:
    provision:
      - kind: sh
        run: scripts/setup.sh
        name: Setup Infrastructure
        env:
          AZURE_LOCATION: ${AZURE_LOCATION}
          RESOURCE_GROUP: rg-${AZURE_ENV_NAME}
        secrets:
          DB_PASSWORD: akvs://my-vault/db-password
        continueOnError: false
      - kind: pwsh
        run: scripts/configure-app.ps1
        env:
          APP_NAME: myapp-${AZURE_ENV_NAME}
    destroy:
      - kind: sh
        run: scripts/teardown.sh
        env:
          RESOURCE_GROUP: rg-${AZURE_ENV_NAME}
```

### Types to Implement

`internal/provisioning/config.go`

```go
// ProviderConfig is the top-level configuration parsed from infra.config in azure.yaml.
type ProviderConfig struct {
    Provision []*ScriptConfig `json:"provision,omitempty"`
    Destroy   []*ScriptConfig `json:"destroy,omitempty"`
}

// ScriptConfig defines the configuration for a single script entry.
// Field names align with HookConfig (pkg/ext/models.go) for user familiarity.
type ScriptConfig struct {
    Kind            string            `json:"kind,omitempty"`
    Shell           string            `json:"shell,omitempty"` // deprecated alias for Kind
    Run             string            `json:"run"`
    Name            string            `json:"name,omitempty"`
    Env             map[string]string `json:"env,omitempty"`
    Secrets         map[string]string `json:"secrets,omitempty"`
    ContinueOnError bool              `json:"continueOnError,omitempty"`
    Windows         *ScriptConfig     `json:"windows,omitempty"`
    Posix           *ScriptConfig     `json:"posix,omitempty"`
}
```

### Parsing Strategy

Use the JSON re-marshal pattern established in azd (`tools.UnmarshalHookConfig`):

```go
[...]

Issue #7734, opened 2026-04-15
Script provisioning provider — extension scaffold


wbreza:
## Script provisioning provider -- extension scaffold

**Parent epic**: #7733
**Framework**: #7465 | **Framework PR**: #7482

### Overview

Create the `microsoft.azd.scripts` extension following the established `microsoft.azd.demo` extension pattern. This is the foundation issue -- it produces a compilable, registrable extension with stub implementations of all provider methods.

#### Extension files

| File | Purpose |
|---|---|
| `extensions/microsoft.azd.scripts/extension.yaml` | Extension metadata: id (`microsoft.azd.scripts`), capabilities (`provisioning-provider`), namespace (`scripts`), version |
| `extensions/microsoft.azd.scripts/main.go` | Entry point -- delegates to cobra root command |
| `extensions/microsoft.azd.scripts/internal/cmd/listen.go` | `ExtensionHost` setup with `WithProvisioningProvider("scripts", factory)` |
| `extensions/microsoft.azd.scripts/internal/cmd/root.go` | Cobra command tree (root + listen + version) |
| `extensions/microsoft.azd.scripts/internal/cmd/version.go` | Version command reading from `version.txt` |
| `extensions/microsoft.azd.scripts/go.mod` | Go module definition (references `../../pkg/azdext`) |
| `extensions/microsoft.azd.scripts/build.ps1` | Windows build script (follows demo pattern) |
| `extensions/microsoft.azd.scripts/build.sh` | POSIX build script |
| `extensions/microsoft.azd.scripts/ci-build.ps1` | CI build script |
| `extensions/microsoft.azd.scripts/ci-test.ps1` | CI test script |
| `extensions/microsoft.azd.scripts/version.txt` | Semver version (e.g., `0.1.0-alpha.1`) |
| `extensions/microsoft.azd.scripts/cspell.yaml` | Extension-specific spell check config |
| `extensions/microsoft.azd.scripts/README.md` | Extension readme |
| `extensions/microsoft.azd.scripts/CHANGELOG.md` | Changelog |

#### Stub provider

Create `internal/provisioning/provider.go` with a `ScriptProvisioningProvider` struct that implements `azdext.ProvisioningProvider`. All methods should be no-ops returning empty/zero-value responses:

```go
type ScriptProvisioningProvider struct {
    azdClient *azdext.AzdClient
}

// All methods: Initialize, EnsureEnv, Deploy, Destroy, Preview, State, Parameters, PlannedOutputs
[...]

Issue #7733, opened 2026-04-15
Epic: Script Provisioning Provider Extension


wbreza:
## Epic: Script Provisioning Provider Extension

The Script Provisioning Provider is designed as a pragmatic onramp for teams adopting Azure with azd. It enables teams with existing script-based provisioning workflows to immediately benefit from azd's environment management, parameter resolution, and deployment lifecycle -- while providing a natural migration path toward declarative IaC (Bicep or Terraform) over time.

### Context

- **Parent epic**: #7465 (Provisioning Providers in the AZD Extension Framework)
- **Framework PR**: #7482 (adds `WithProvisioningProvider()` fluent API and gRPC protocol)

### Problem

Today, `azd provision` and `azd down` only work with declarative IaC tools (Bicep and Terraform). Many teams manage Azure infrastructure with imperative shell scripts that call `az cli`, Azure PowerShell, or other Azure tooling. These teams cannot adopt azd without rewriting their provisioning workflows.

The provisioning provider extensibility framework (#7465) enables extensions to register custom provisioning providers via gRPC. This epic delivers a **Script Provisioning Provider Extension** (`microsoft.azd.scripts`) -- a first-party extension that allows users to configure shell scripts (bash/PowerShell) as their provisioning and teardown workflow in `azure.yaml`.

### Architecture

The extension is a **pure provisioning provider** -- no custom CLI commands, no MCP server, no lifecycle event handlers. It registers a `"scripts"` provider via `WithProvisioningProvider("scripts", factory)` and implements the full `azdext.ProvisioningProvider` interface.

Key components:

- **ProviderConfig / ScriptConfig** -- typed configuration parsed from `infra.config` in `azure.yaml`
- **EnvResolver** -- 4-layer environment variable merging (secrets > env > azd env > OS)
- **ScriptExecutor** -- runs bash/pwsh scripts with merged environment, streams stdout/stderr
- **OutputCollector** -- discovers and parses `outputs.json` files, merges across scripts

### Work Items

| # | Issue | Title | Status |
|---|-------|-------|--------|
| 1 | #7734 | Extension scaffold | Not started |
| 2 | #7735 | Config design and parsing | Not started |
| 3 | #7736 | Full provider implementation | Not started |

### Dependency Order

```
#7734 Extension Scaffold (no dependencies -- start here)
  |
  v
#7735 Config Design & Parsing (depends on scaffold)
  |
  v
#7736 Full Provider Implementation (depends on config parsing)
```

PR #7737, opened 2026-04-15
feat: Script Provisioning Provider extension (microsoft.azd.scripts)

wbreza:
## Description

Implements the **Script Provisioning Provider** extension (`microsoft.azd.scripts`) that enables shell script-based provisioning and teardown workflows in azd. This is a first-party extension that registers a `scripts` provisioning provider via gRPC.

**Epic**: #7733
**Depends on**: #7482 (provisioning provider framework)
**Resolves**: #7734, #7735, #7736

## Architecture

The extension is a **pure provisioning provider** — no custom CLI commands, no MCP server. It registers via `WithProvisioningProvider("scripts", factory)` and implements the full `azdext.ProvisioningProvider` interface.

### Key Components

| Component | File | Purpose |
|-----------|------|---------|
| Config parser | `internal/provisioning/config.go` | Typed config from `infra.config` with validation |
| EnvResolver | `internal/provisioning/env_resolver.go` | 4-layer environment variable merging |
| ScriptExecutor | `internal/provisioning/executor.go` | Runs bash/pwsh scripts with merged env |
| OutputCollector | `internal/provisioning/output_collector.go` | Discovers/parses `outputs.json` files |
| Provider | `internal/provisioning/provider.go` | Full `ProvisioningProvider` implementation |

### User Configuration

```yaml
infra:
  provider: scripts
  config:
    provision:
      - kind: sh
        run: scripts/setup.sh
        name: Setup Infrastructure
        env:
          AZURE_LOCATION: ${AZURE_LOCATION}
          RESOURCE_GROUP: rg-${AZURE_ENV_NAME}
    destroy:
      - kind: sh
        run: scripts/teardown.sh
```

## Testing

- 25 unit tests covering config parsing, validation, env resolution, output collection, and executor
- All paths tested: happy path, error cases, edge cases, security (path traversal, absolute paths)
- Lint, spellcheck, copyright all pass

## Checklist

- [x] Extension compiles with `go build`
- [x] Registers `scripts` provider via gRPC
- [x] Config validation (missing run, absolute paths, path traversal, missing files, unknown kinds)
- [x] Kind auto-inference from file extension (`.sh` -> `sh`, `.ps1` -> `pwsh`)
- [x] Shell -> Kind backward compatibility mapping
- [x] Platform-specific overrides (Windows/Posix)
- [x] 4-layer env var merging with `${EXPRESSION}` substitution
- [x] Output collection with 10MB size limit
- [x] golangci-lint: 0 issues
- [x] cspell: 0 issues
- [x] All tests pass


cli/azd/extensions/microsoft.azd.scripts/internal/provisioning/env_resolver.go
```diff
@@ -0,0 +1,79 @@
...
+			if v, ok := merged[key]; ok {
+				return v
+			}
+			return ""
+		}
+
+		for k, tmpl := range sc.Env {
+			resolved, err := envsubst.Eval(tmpl, lookup)
+			if err != nil {
+				return nil, fmt.Errorf("resolving env variable %q: %w", k, err)
```
jongio, line 62: Go map iteration order is nondeterministic. If two vars in the same `env` map reference each other (e.g., `A: "${B}"`, `B: "hello"`), the result depends on which key Go happens to iterate first.

Consider sorting keys before iteration:

```go
keys := slices.Sorted(maps.Keys(sc.Env))
for _, k := range keys {
    tmpl := sc.Env[k]
    // ...
}
```

Alternatively, document that cross-references within the same `env` map aren't supported and users should use azd environment vars (Layer 2) for shared values instead.

jongio (commented):
6 findings from deep review. Key items:

- 2 HIGH: nondeterministic env var resolution order, no provider.go tests
- 2 MEDIUM: ContinueOnError zero-value limitation, raw stderr warning
- 2 LOW: destroy outputs discarded, missing ctx check between iterations

The extension architecture is clean - good separation of config/resolver/executor/collector. The 4-layer env resolution model is well-designed. Tests cover the individual components thoroughly.

cli/azd/extensions/microsoft.azd.scripts/internal/provisioning/provider.go
```diff
@@ -0,0 +1,223 @@
...
+		return make(map[string]string), nil
+	}
+
+	result := make(map[string]string, len(resp.GetKeyValues()))
+	for _, kv := range resp.GetKeyValues() {
+		result[kv.GetKey()] = kv.GetValue()
+	}
+
+	return result, nil
+}
```
jongio, line 223: This file (223 lines) is the main integration point but has no test coverage. Config parsing, env resolution, executor, and output collector all have unit tests - but `runScripts` (which orchestrates error handling, output merging, and progress reporting) doesn't.

Consider adding tests that mock `AzdClient` and verify:
- Script execution order and progress callbacks
- Output merging across multiple scripts
- `ContinueOnError` behavior (continue vs stop)
- Context cancellation propagation

cli/azd/extensions/microsoft.azd.scripts/internal/provisioning/config.go
```diff
@@ -0,0 +1,208 @@
...
+		sc.Run = override.Run
+	}
+	if override.Kind != "" {
+		sc.Kind = override.Kind
+	}
+	if override.Shell != "" {
+		sc.Shell = override.Shell
+	}
+	if override.Name != "" {
+		sc.Name = override.Name
```
jongio, line 167: `if override.ContinueOnError` only triggers when the override sets it to `true`. Due to Go zero-value semantics, a platform override can't explicitly disable a `ContinueOnError` that was set in the base config.

If this is intentional (platform overrides can only add permissiveness, not remove it), a comment here would help. If not, consider using `*bool` to distinguish 'not set' from 'explicitly false'.

cli/azd/extensions/microsoft.azd.scripts/internal/provisioning/provider.go
```diff
@@ -0,0 +1,223 @@
...
+
+// getAzdEnv retrieves current azd environment values via the gRPC client.
+func (p *ScriptProvisioningProvider) getAzdEnv(ctx context.Context) (map[string]string, error) {
+	resp, err := p.azdClient.Environment().GetValues(ctx, &azdext.GetEnvironmentRequest{})
+	if err != nil {
+		// Context cancellation should propagate
+		if ctx.Err() != nil {
+			return nil, ctx.Err()
+		}
+
```
jongio, line 211: `fmt.Fprintf(os.Stderr, ...)` bypasses the extension framework's logging. The `azdext` package provides structured logging that integrates with azd's output handling - using it here would keep diagnostics consistent with the rest of the extension framework.

cli/azd/extensions/microsoft.azd.scripts/internal/provisioning/provider.go
```diff
@@ -0,0 +1,223 @@
...
+	// Collect keys that were produced by provisioning to invalidate them
+	invalidatedKeys := make([]string, 0, len(p.outputs))
+	for k := range p.outputs {
+		invalidatedKeys = append(invalidatedKeys, k)
+	}
+
+	return &azdext.ProvisioningDestroyResult{
+		InvalidatedEnvKeys: invalidatedKeys,
+	}, nil
```
jongio, line 131: `runScripts` collects outputs from destroy scripts, but they're not wired into the result. Only invalidation keys from prior provisioning outputs are returned. If this is intentional (destroy scripts don't produce meaningful outputs), a brief comment would clarify. If destroy scripts should be able to produce outputs (e.g., for audit/logging), consider including them.

cli/azd/extensions/microsoft.azd.scripts/internal/provisioning/provider.go
```diff
@@ -0,0 +1,223 @@
...
+	progress grpcbroker.ProgressFunc,
+) (map[string]OutputParameter, error) {
+	azdEnv, err := p.getAzdEnv(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("loading azd environment: %w", err)
+	}
+
+	resolver := NewEnvResolver(azdEnv)
+	executor := NewScriptExecutor(p.projectPath)
+	collector := NewOutputCollector(p.projectPath)
```
jongio, line 163: nit: The loop doesn't check `ctx.Err()` between script iterations. For long script sequences, adding an early cancellation check before each iteration would improve responsiveness:

```go
for i, sc := range scripts {
    if err := ctx.Err(); err != nil {
        return nil, err
    }
    // ...
}
```

cli/azd/extensions/microsoft.azd.scripts/internal/provisioning/env_resolver.go
wbreza: Fixed — added `slices.Sorted(maps.Keys(sc.Env))` for deterministic key iteration order.

cli/azd/extensions/microsoft.azd.scripts/internal/provisioning/provider.go
wbreza: Deferred to issue #7903 — this is significant test work that warrants its own PR. The individual components (config parsing, env resolution, executor, output collector) all have unit tests today.

cli/azd/extensions/microsoft.azd.scripts/internal/provisioning/config.go
wbreza: Fixed — changed `ContinueOnError` from `bool` to `*bool` so platform overrides can explicitly disable it. Uses `new(true)` / `new(false)` (Go 1.26 pattern) for initialization.

cli/azd/extensions/microsoft.azd.scripts/internal/provisioning/provider.go
wbreza: Fixed — replaced `fmt.Fprintf(os.Stderr, ...)` with `log.Printf(...)` for consistent extension framework logging.

cli/azd/extensions/microsoft.azd.scripts/internal/provisioning/provider.go
wbreza: Added a clarifying comment — destroy script outputs are intentionally discarded. Only invalidation keys from prior provisioning are returned to signal which cached resources should be cleared.

cli/azd/extensions/microsoft.azd.scripts/internal/provisioning/provider.go
wbreza: Fixed — added `ctx.Err()` check at the top of the script iteration loop for responsive cancellation.

cli/azd/extensions/microsoft.azd.scripts/internal/provisioning/output_collector.go
```diff
@@ -0,0 +1,75 @@
...
+
+	fi, err := os.Stat(outputsPath)
+	if err != nil {
+		if os.IsNotExist(err) {
+			return nil, nil
+		}
+		return nil, fmt.Errorf("checking outputs file %q: %w", outputsPath, err)
+	}
+
+	if fi.Size() > maxOutputFileSize {
```
jongio, line 47: The collector reads `outputs.json` from the script directory without verifying it was produced by the current execution. If a previous run left an `outputs.json` and the script doesn't overwrite it this time, stale outputs are silently returned.

Two options:
1. Delete (or rename) `outputs.json` before execution so absence means no outputs
2. Check mtime vs script start time

For alpha this is probably fine, but worth a comment noting the assumption that scripts always overwrite if they produce outputs.

cli/azd/extensions/microsoft.azd.scripts/internal/provisioning/env_resolver.go
```diff
@@ -0,0 +1,83 @@
...
+			if v, ok := merged[key]; ok {
+				return v
+			}
+			return ""
+		}
+
+		// Sort keys for deterministic resolution order — important when
+		// variables in the same env map reference each other.
+		for _, k := range slices.Sorted(maps.Keys(sc.Env)) {
+			tmpl := sc.Env[k]
```
jongio, line 63: nit: The sorted iteration makes resolution deterministic, which is good. Worth noting the behavior explicitly though - if A refs B and A < B alphabetically, A resolves first and gets empty string because B hasn't been processed yet.

A one-line comment like `// NB: forward references within the same env map resolve to empty` would save someone a debugging session.

cli/azd/extensions/microsoft.azd.scripts/ci-build.ps1
```diff
@@ -0,0 +1,144 @@
...
+$oldGOEXPERIMENT = $env:GOEXPERIMENT
+# Enable the loopvar experiment, which makes the loop variaible for go loops like `range` behave as most folks would expect.
+# the go team is exploring making this default in the future, and we'd like to opt into the behavior now.
+$env:GOEXPERIMENT = "loopvar"
+
+try {
+    Write-Host "Running: go build ``"
+    PrintFlags -flags $buildFlags
+    go build @buildFlags
+    if ($LASTEXITCODE) {
```
jongio, line 109: GOEXPERIMENT=loopvar became the default in Go 1.22. Since the module targets Go 1.26, this block is dead code and can be removed.

jongio (commented):
Three additional items beyond the existing review. The architecture is clean - good separation of config/resolver/executor/collector, and the security validation (path traversal, absolute paths) is solid.

PR #7747, opened 2026-04-15
[Agents Extension] Updates to pull rbac identity from the agent version

### LINKED ISSUE #7514: Post-Deploy RBAC Hook Fails, No Pre-Flight Auth Check

### LINKED ISSUE #7748: Use returned agent identity for rbac assignment

trangevi:
The agent identity is now available on the agent version object, updating handling to use that instead of relying on graph


cli/azd/extensions/azure.ai.agents/internal/cmd/listen.go
```diff
@@ -160,41 +161,107 @@ func isHostedAgentService(svc *azdext.ServiceConfig, proj *azdext.ProjectConfig)
...
 		return fmt.Errorf("failed to get current environment for agent identity RBAC: %w", err)
 	}
 
-	var agentNames []string
+	envName := envResp.Environment.Name
+
+	// Read the project endpoint for API calls.
+	endpointResp, err := azdClient.Environment().GetValue(ctx, &azdext.GetEnvRequest{
+		EnvName: envName,
+		Key:     "AZURE_AI_PROJECT_ENDPOINT",
```
jongio, line 183: The credential creation, agent client setup, and `GetAgentVersion` API calls all happen here before `EnsureAgentIdentityRBAC` checks `isVnextEnabled`. Two issues:

1. **Non-vnext deploys pay for unnecessary HTTP calls** - the `GetAgentVersion` calls hit the Azure API even though RBAC will be skipped.
2. **Non-vnext projects can now fail** - if `AZURE_AI_PROJECT_ENDPOINT` or `AZURE_TENANT_ID` aren't set (reasonable for non-vnext), `postdeployHandler` returns a hard error instead of proceeding.

Consider gating this block on vnext status, or moving the version fetch into `EnsureAgentIdentityRBAC` after the existing vnext check.

cli/azd/extensions/azure.ai.agents/internal/cmd/listen.go
```diff
@@ -160,41 +161,107 @@ func isHostedAgentService(svc *azdext.ServiceConfig, proj *azdext.ProjectConfig)
...
+		if err != nil || versionResp.Value == "" {
+			return fmt.Errorf(
+				"AGENT_%s_VERSION is not set — deployment may not have completed successfully",
+				serviceKey,
+			)
+		}
+
+		// Fetch the agent version to get the instance identity principal ID.
+		versionObj, err := agentClient.GetAgentVersion(
+			ctx, nameResp.Value, versionResp.Value, DefaultAgentAPIVersion,
```
jongio: This uses `DefaultAgentAPIVersion` (`2025-05-15-preview`), but `show.go` was updated in this same PR to use `DefaultVNextAgentAPIVersion` (`2025-11-15-preview`) for the vnext path. If `InstanceIdentity` is only returned by the newer API version, this call will always get `nil` and fall back to Graph discovery - making the optimization a no-op. Should this be `DefaultVNextAgentAPIVersion`?

cli/azd/extensions/azure.ai.agents/internal/cmd/listen.go
```diff
@@ -160,41 +161,107 @@ func isHostedAgentService(svc *azdext.ServiceConfig, proj *azdext.ProjectConfig)
...
 			continue
 		}
-		nameKey := fmt.Sprintf("AGENT_%s_NAME", toServiceKey(svc.Name))
+		serviceKey := toServiceKey(svc.Name)
+
+		nameResp, err := azdClient.Environment().GetValue(ctx, &azdext.GetEnvRequest{
+			EnvName: envName,
+			Key:     fmt.Sprintf("AGENT_%s_NAME", serviceKey),
+		})
+		if err != nil || nameResp.Value == "" {
```
jongio: This changes from a silent `continue` (skip the service) to a hard `return` error. If a project has multiple hosted agents and one hasn't finished deploying yet, the entire postdeploy handler now fails - preventing RBAC setup for the agents that did deploy successfully.

Also, when `err != nil` (e.g., env service RPC failure), the error message says "NAME is not set - deployment may not have completed" which is misleading about the actual cause.

cli/azd/extensions/azure.ai.agents/internal/project/agent_identity_rbac.go
```diff
@@ -240,71 +258,75 @@ func ensureAgentIdentityRBACWithCred(
...
-	}
-	fmt.Println("  ✓ Agent identity found in Entra ID")
-
-	// Assign Azure AI User role scoped to the Foundry Project
-	for _, sp := range agentIdentities {
-		principalID := ""
-		if sp.Id != nil {
-			principalID = *sp.Id
+		// Use the first matching identity.
+		if agentIdentities[0].Id != nil {
```
jongio, line 323: The old code iterated over all matching SPs and assigned roles to each. The new fallback path uses only the first match. This is probably fine in practice (display names should be unique per agent), but it's a behavioral change worth calling out - if there were multiple matches, only the first gets the role now.

jongio (changes requested):
Nice optimization - skipping Graph API polling is a solid improvement. A few things to sort out before this is ready:

The main concern is that the new credential + API calls in `postdeployHandler` run unconditionally, before the vnext gate in `EnsureAgentIdentityRBAC`. This means non-vnext projects pay for unnecessary HTTP calls and can fail on missing env vars. Additionally, `GetAgentVersion` uses the older API version while `show.go` was updated to the vnext version in this same PR - if `InstanceIdentity` requires the newer version, the optimization won't actually kick in.

cli/azd/extensions/azure.ai.agents/internal/cmd/listen.go
```diff
@@ -160,41 +161,107 @@ func isHostedAgentService(svc *azdext.ServiceConfig, proj *azdext.ProjectConfig)
...
+		versionResp, err := azdClient.Environment().GetValue(ctx, &azdext.GetEnvRequest{
+			EnvName: envName,
+			Key:     fmt.Sprintf("AGENT_%s_VERSION", serviceKey),
 		})
+		if err != nil || versionResp.Value == "" {
+			return fmt.Errorf(
+				"AGENT_%s_VERSION is not set — deployment may not have completed successfully",
+				serviceKey,
+			)
+		}
```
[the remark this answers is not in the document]
trangevi: This is an interesting point, is there a way to know during post deploy specifically which service/s was/were deployed?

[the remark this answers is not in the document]
JeffreyCA: I think you'd have to use a "postdeploy" **Service**EventHandler instead of a **Project**EventHandler - the exact service deployed would be captured in `ServiceEventArgs`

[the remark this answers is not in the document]
trangevi: I see. I'll take that as a follow up item

cli/azd/extensions/azure.ai.agents/internal/project/agent_identity_rbac.go
```diff
@@ -6,7 +6,9 @@ package project
 import (
 	"context"
 	"fmt"
+	"maps"
 	"os"
+	"slices"
 	"strconv"
 	"strings"
 	"sync"
```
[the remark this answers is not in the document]
trangevi: We're on a later version than that

JeffreyCA (approved):

cli/azd/extensions/azure.ai.agents/internal/project/agent_identity_rbac.go
```diff
@@ -240,71 +257,95 @@ func ensureAgentIdentityRBACWithCred(
...
-			if err := verifyRoleAssignment(ctx, cred, principalID, roleAzureAIUser, info.ProjectScope); err != nil {
-				return fmt.Errorf("failed to verify Azure AI User role assignment: %w", err)
-			}
-			fmt.Println("    ✓ Azure AI User → Foundry Project (verified)")
-		} else {
-			fmt.Println("    ✓ Azure AI User → Foundry Project (already assigned)")
+	if created {
+		fmt.Println("    ✓ Azure AI User → Foundry Project (created)")
+		fmt.Println("    ⏳ Verifying Azure AI User...")
+		if err := verifyRoleAssignment(ctx, cred, principalID, roleAzureAIUser, info.ProjectScope); err != nil {
```
trangevi, line 343: @therealjohn is this the piece you're saying we can remove? We don't need to do the verification?

trangevi: Discussed offline, all good

jongio (approved):
All my previous concerns are addressed - the vnext gate, API version fix, soft env var handling, and multi-SP logging all look good.

The flow is clean: fetch agent version with the vnext API version to get `InstanceIdentity.PrincipalID`, pass it through to RBAC assignment, skip Graph entirely when the principal ID is known. The lazy Graph client init for the fallback path and the token protection error detection are solid additions.

One minor note: this bumps `azidentity` to `v1.14.0-beta.3` - just flagging in case that wasn't intentional.

PR #7763, opened 2026-04-16
 Fix developer RBAC pre-flight gaps: auto-assign AI User, add role-write check, detect ABAC registries

### LINKED ISSUE #7540: Authorization failed for template resource error with does not have permission to perform action 'Microsoft.Authorization/roleAssignments/write'

therealjohn:
Problem

A developer with only the Contributor role on a Foundry Project could pass the pre-flight RBAC check (CheckDeveloperRBAC) but then hit a 403 AuthorizationFailed error during postdeploy when the extension tried to
assign the Azure AI User role to agent service principals:

 ERROR: agent identity RBAC setup failed: failed to assign Azure AI User role:
 RESPONSE 403: AuthorizationFailed — does not have authorization to perform action
 'Microsoft.Authorization/roleAssignments/write'

This happens because Contributor explicitly blocks Microsoft.Authorization/*/Write in its notActions. The pre-flight check verified the developer could use the project, but not whether they could write role
assignments. Two additional gaps existed: the extension didn't attempt to auto-assign Azure AI User when it was missing (unlike the VS Code extension), and the ACR role check didn't distinguish between classic and
ABAC-mode registries (where classic roles like AcrPush don't work).

Changes

Gap 1 — Auto-assign Azure AI User if missing (Check 1)
If the developer lacks Azure AI User on the Foundry Project, the pre-flight now attempts to assign it (using PrincipalTypeUser). This matches VS Code's ensureAzureAIUserRole pattern. Only fails with a structured error
if the auto-assign itself returns 403.

Gap 2 — Explicit roleAssignments/write check (new Check 2)
Added a check that the developer has at least one of: Owner, User Access Administrator, or Role Based Access Control Administrator on the Foundry Project scope. Contributor is intentionally excluded — it cannot write
role assignments. The error suggestion includes the az role assignment create command and an AZD_AGENT_SKIP_ROLE_ASSIGNMENTS=true opt-out for environments where role assignments are managed externally.

Gap 3 — ABAC registry detection (Check 3)
resolveACRResourceID → resolveACRInfo(ctx, cred, subscriptionID, loginServer) (resourceID, isAbac, error). Uses armcontainerregistry v1.3.0-beta.3 (upgraded from v1.2.0) which exposes 
RegistryProperties.RoleAssignmentMode. When ABAC mode is detected, the check uses sufficientACRAbacRoles (Owner, Container Registry Repository Writer, Container Registry Repository Contributor) instead of the classic
role list. The AZURE_CONTAINER_REGISTRY_RESOURCE_ID env-var short-circuit was removed so ABAC mode is always detected via the ARM API.

Supporting changes 

 - assignRoleToIdentity now accepts a principalType armauthorization.PrincipalType parameter (agent callers pass ServicePrincipal; the new developer auto-assign path passes User)
 - New error code: CodeDeveloperMissingRoleAssignWriteRole
 - New role constants: roleUserAccessAdministrator, roleRBACAdministrator, roleACRRepositoryWriter
 - New role lists: sufficientRoleAssignWriteRoles, sufficientACRAbacRoles
 - Tests: GUID assertions for all new constants; membership/exclusion assertions for both new role lists (including assert.NotContains(sufficientRoleAssignWriteRoles, roleContributor))


jongio (commented):
Four findings on the RBAC pre-flight changes. The core logic (3-check structure, auto-assign, ABAC branching) looks solid. CI note: cspell doesn't know "ABAC" - needs adding to cspell.yaml.

cli/azd/extensions/azure.ai.agents/go.mod
```diff
@@ -7,7 +7,7 @@ require (
 	github.com/Azure/azure-sdk-for-go/sdk/azidentity v1.13.1
 	github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/authorization/armauthorization/v3 v3.0.0-beta.2
 	github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/cognitiveservices/armcognitiveservices/v2 v2.0.0
-	github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/containerregistry/armcontainerregistry v1.2.0
+	github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/containerregistry/armcontainerregistry v1.3.0-beta.3
```
jongio, line 12: This is a beta SDK (`v1.3.0-beta.3`). Beta Azure SDKs can introduce breaking changes - if `RoleAssignmentMode` or `AbacRepositoryPermissions` gets renamed/restructured before GA, this breaks silently (compile will pass but the enum comparison won't match).

Is there a GA version that exposes this property? If not, worth a code comment noting the beta dependency so future readers know to check for a GA upgrade.

cli/azd/extensions/azure.ai.agents/internal/project/developer_rbac_check.go
```diff
@@ -134,48 +164,113 @@ func CheckDeveloperRBAC(ctx context.Context, azdClient *azdext.AzdClient) error
...
-	// Prefer the persisted ARM resource ID (set during init); fall back to listing registries.
-	acrResourceID := azdEnv["AZURE_CONTAINER_REGISTRY_RESOURCE_ID"]
-	if acrResourceID == "" {
-		acrResourceID, err = resolveACRResourceID(ctx, cred, info.SubscriptionID, acrEndpoint)
-		if err != nil {
-			fmt.Printf("  ⚠ Could not resolve ACR resource ID: %s — skipping ACR role check\n", err)
-			return nil
-		}
+	// Always resolve registry info to accurately detect ABAC mode.
+	acrResourceID, isAbac, err := resolveACRInfo(ctx, cred, info.SubscriptionID, acrEndpoint)
```
jongio: The old code short-circuited via `AZURE_CONTAINER_REGISTRY_RESOURCE_ID` (set during init) to skip listing registries. This now always calls `resolveACRInfo` which pages through ALL registries in the subscription.

For subscriptions with many registries, this is noticeably slower. Consider: when the env var is available, parse the resource group + registry name from the resource ID and use `RegistriesClient.Get()` for a single targeted fetch (still gets `RoleAssignmentMode` for ABAC detection, but O(1) instead of O(n)).

cli/azd/extensions/azure.ai.agents/internal/project/developer_rbac_check.go
```diff
@@ -134,48 +164,113 @@ func CheckDeveloperRBAC(ctx context.Context, azdClient *azdext.AzdClient) error
...
 		fmt.Printf("  ⚠ Could not check AI User role: %s\n", err)
 	} else if !hasAIAccess {
+		// Attempt to auto-assign Azure AI User to the developer. This succeeds when the
+		// developer has Owner, User Access Administrator, or RBAC Administrator.
+		fmt.Println("  Azure AI User role not found — attempting to auto-assign...")
+		if _, assignErr := assignRoleToIdentity(
+			ctx, cred, principalID, roleAzureAIUser,
+			"Azure AI User → Foundry Project", info.ProjectScope,
+			armauthorization.PrincipalTypeUser,
+		); assignErr != nil {
```
jongio, line 177: `assignErr` isn't included in the structured error. A 403, a network timeout, and a malformed scope all produce the same `auto-assign failed` message. Consider including it:

```suggestion
if _, assignErr := assignRoleToIdentity(
    ctx, cred, principalID, roleAzureAIUser,
    "Azure AI User -> Foundry Project", info.ProjectScope,
    armauthorization.PrincipalTypeUser,
); assignErr != nil {
    return exterrors.Auth(
        exterrors.CodeDeveloperMissingAIUserRole,
        fmt.Sprintf(
            "your identity (%s) does not have the 'Azure AI User' role on the Foundry Project %s/%s "+
                "and auto-assign failed: %s",
            userProfile.DisplayName, info.AccountName, info.ProjectName, assignErr,
        ),
```
The underlying error helps users (and support) distinguish permissions issues from transient failures.

cli/azd/extensions/azure.ai.agents/internal/project/developer_rbac_check.go
```diff
@@ -196,7 +291,11 @@ func CheckDeveloperRBAC(ctx context.Context, azdClient *azdext.AzdClient) error
 		)
 	}
 
-	fmt.Println("  ✓ Container Registry role on ACR")
+	if isAbac {
+		fmt.Println("  ✓ Container Registry Repository Writer role on ACR (ABAC mode)")
```
jongio: Nit: this hardcodes "Repository Writer" but the developer might have Owner or Repository Contributor. Consider matching the classic-mode pattern:

```suggestion
fmt.Println("  ✓ Container Registry role on ACR (ABAC mode)")
```

vhvb1989 (approved):

cli/azd/extensions/azure.ai.agents/internal/project/developer_rbac_check.go
```diff
@@ -134,48 +167,127 @@ func CheckDeveloperRBAC(ctx context.Context, azdClient *azdext.AzdClient) error
...
+						"ask a subscription Owner or User Access Administrator to assign the 'Azure AI User' role "+
+							"to your identity on the Foundry Project scope:\n"+
+							"  az role assignment create --assignee %s --role \"Azure AI User\" --scope %q",
+						principalID, info.ProjectScope,
+					),
+				)
+			}
+			fmt.Printf("  ⚠ Azure AI User auto-assign failed (non-auth error): %s — continuing\n", assignErr)
+		}
+		fmt.Println("  ✓ Azure AI User auto-assigned to developer identity")
```
jongio: The `fmt.Println("  ✓ Azure AI User auto-assigned to developer identity")` sits outside the `if assignErr != nil` block, so on a non-403 failure the output contradicts itself:

```
  ⚠ Azure AI User auto-assign failed (non-auth error): <err> - continuing
  ✓ Azure AI User auto-assigned to developer identity
```

Move it into an `else` branch on the outer `if assignErr != nil`, or guard it with a flag set only on the success path.

jongio (commented):
One follow-up. My earlier comments on perf, assignErr, and the hardcoded ABAC role name are all addressed. Found one fall-through bug in the auto-assign path.

vhvb1989 (approved):

PR #7768, opened 2026-04-17
update config set of error handling to auto fix and rerun

### LINKED ISSUE #7769: Update error handling fix allow config to auto-retry the command after fix
When `azd config set copilot.errorHandling.fix allow` is set, the current behavior only applies the fix but does not rerun the command. This should be updated to automatically retry the command after the fix is applied, matching the `actionFixAndRetry` behavior.

Linked PR: #7768

hemarina:
Update error handling so that when copilot.errorHandling.fix is set to allow, the command is automatically retried after the fix is applied (actionFixAndRetry), instead of just applying the fix without rerun (actionFixOnly).

Also updates help text and warning messages to use consistent 'auto-fix and retry' terminology.

Closes #7769

### DIFF
--- a/cli/azd/cmd/middleware/error.go
+++ b/cli/azd/cmd/middleware/error.go
@@ -320,6 +320,8 @@ func (e *ErrorMiddleware) Run(ctx context.Context, next NextFn) (*actions.Action
 			// Unlike the explain/guidance path (which auto-retries when the user
 			// explicitly chose actionFixAndRetry), the category-fix path asks for
 			// retry confirmation because the user only chose "fix" — not "fix and retry".
+			// The saved copilot.errorHandling.fix=allow preference bypasses this prompt
+			// and auto-retries.
 			shouldRetry, err := e.promptRetryAfterFix(ctx)
 			if err != nil {
 				span.SetStatus(codes.Error, "agent.retry.failed")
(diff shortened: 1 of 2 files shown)


vhvb1989 (approved):

jongio (commented):
Small behavior change looks correct and the test is updated. A couple of follow-ups to consider:

1. **HelpMessage is stale.** A few lines below the diff (around `error.go:522-524`), the help still advertises this as:

   > "To always allow fixes, run `azd config set copilot.errorHandling.fix allow`"

   After this PR, setting that key also auto-reruns the command. The warning you updated at line 496 says "auto-fix and rerun is enabled", but the help that teaches users how to turn this on doesn't mention retry. Worth syncing them in the same PR, e.g.:

   > "To always auto-fix and retry, run `azd config set copilot.errorHandling.fix allow`"

2. **CHANGELOG.** The `1.24.0-beta.1 (Unreleased)` section is empty. This changes the observable behavior of a persisted user-config key (anyone who already set `copilot.errorHandling.fix=allow` during the previous beta will now get auto-retry too), so a line under Other Changes would help.

3. **PR hygiene** (not code): PR description is empty and `PR Governance` is failing because the PR isn't linked to an issue. Add a linked issue so governance can go green.

hemarina:
Thanks for the review @jongio!

1. **HelpMessage** — Good catch. Updated the help text to say "To always auto-fix and retry" to be consistent with the warning message. Also standardized on "retry" terminology throughout.

2. **CHANGELOG** — This is an alpha feature behind the \llm\ feature flag and hasn't been announced to users (blog post is not out yet), so we don't need to add this change to the changelog.

3. **PR hygiene** — Will create and link an issue.

cli/azd/cmd/middleware/error.go
```diff
@@ -488,18 +488,17 @@ func (e *ErrorMiddleware) promptNextAction(
...
 
-	// Saved "always fix" preference → auto-approve fix only (user still
-	// controls retry via the category-fix path or interactive prompt).
+	// Saved "always fix" preference → auto-approve fix and rerun the command.
 	if val, ok := userConfig.GetString(
 		agentcopilot.ConfigKeyErrorHandlingFix); ok && val == "allow" {
 		e.console.Message(ctx, output.WithWarningFormat(
-			"\n%s auto-fix is enabled. To change, run %s.",
+			"\n%s auto-fix and rerun is enabled. To change, run %s.",
 			agentcopilot.DisplayTitle,
```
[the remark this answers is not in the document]
hemarina: Fixed — updated both the warning message and the help text to consistently use "auto-fix and retry" terminology throughout.

cli/azd/cmd/middleware/error.go
```diff
@@ -566,5 +601,16 @@ func (e *ErrorMiddleware) promptRetryAfterFix(ctx context.Context) (bool, error)
 		return false, fmt.Errorf("invalid retry choice selected")
 	}
 
-	return choices[*choiceIndex].Value == "retry", nil
+	shouldRetry := choices[*choiceIndex].Value == "retry"
+
+	// Print hint about persisting the choice (only retry has a saved equivalent).
+	if shouldRetry {
+		e.console.Message(ctx, output.WithGrayFormat(
+			"Tip: To always auto retry after fix, run: %s",
```
jongio, line 609: Nit: this Tip says `auto retry after fix`, but the HelpMessage a few lines above and the matching Tip in `promptNextAction` both say `auto-fix and retry`. Since this PR standardizes on `auto-fix and retry` terminology, worth aligning this one too.

hemarina: Thanks for the comment. This is the second prompt after users select fix and it is slightly different than the other prompt will it ask for fix and retry together. I would keep it to make it clear it for only auto retry but won't auto fix.

jongio: Fair point that this prompt is contextually different. But in this same function, both the HelpMessage at line 584 and the saved-allow warning at line 567 use "auto-fix and retry". So the Tip stands alone with a unique wording. If the goal is to communicate "this only auto-retries, doesn't auto-fix" in this flow, the HelpMessage and warning should match that intent too. Otherwise it's three strings for one setting with two different wordings. Up to you, but worth aligning.

rajeshkamal5050 (approved):

Issue #7704, opened 2026-04-14
`azd auth token --scope https://graph.microsoft.com/.default` fails with AADSTS530084 for some internal users


JeffreyCA:
```
ERROR: AADSTS530084: Access has been blocked by conditional access token protection policy configured by this organization. To learn more, see https://aka.ms/TBCADocs. Trace ID: <redacted> Correlation ID: <redacted> Timestamp: 2026-04-13 19:04:12Z

Suggestion: reauthentication required, run `azd auth login --tenant-id <redacted> --scope https://graph.microsoft.com/.default` to acquire a new token.
```

This error happens for some internal Microsoft users that are part of a MSIT content access policy that prevents them from getting a MS Graph token. 

Only users in this policy can reproduce the issue with `azd auth token -o json --scope https://graph.microsoft.com/.default --tenant-id <tenant-id>`

<img width="2197" height="2052" alt="Image" src="https://github.com/user-attachments/assets/825ab1da-eb14-4b0a-93ff-4e0f165c4062" />

The ideal solution would be to enable Web Account Manager (WAM) based login on Windows devices [like Azure CLI](https://learn.microsoft.com/en-us/cli/azure/authenticate-azure-cli-interactively?view=azure-cli-latest#sign-in-with-web-account-manager-wam-on-windows). We did add support for it in #3216 but it's not enabled yet in our public releases due to some pending work: 

- #3431
- #3252

The only known workaround is to contact the internal help desk and request for a token protection exception on this policy.

Related Azure CLI issue: https://github.com/Azure/azure-cli/issues/31030

PR #7797, opened 2026-04-17
Provide error suggestions for `AADSTS530084` and include auth error details when emitted as `ErrorWithSuggestion` in telemetry

### LINKED ISSUE #7794: Detect `AADSTS530084` auth errors in core and provide better error suggestions
We can use a similar suggestion message to what we added to the agents extension: https://github.com/Azure/azure-dev/pull/7713

### LINKED ISSUE #7884: Auth error details missing from telemetry when errors are emitted as `error.suggestion`
These properties are typically included when auth errors are emitted with result code `service.aad.failed`, but they're missing with `error.suggestion` (how most auth errors are classified after https://github.com/Azure/azure-dev/pull/6827):

- `error.service.errorcode`
- `error.service.name`
- `error.service.correlationid`
- `error.service.statuscode`

JeffreyCA:
Resolves #7794
Resolves #7884

This PR adds support for handling `AADSTS530084` token protection errors in the authentication error handling logic. Instead of introducing a new azd-specific auth error type, it preserves the original `*auth.AuthFailedError` and wraps it in `*internal.ErrorWithSuggestion` so users still receive clear guidance and documentation links while downstream auth classification continues to use the underlying AAD error semantics.

This PR also preserves auth error details in telemetry when auth failures are wrapped in `*internal.ErrorWithSuggestion`.

Until https://github.com/Azure/azure-dev/issues/7704 is addressed, users won't be able to fix this issue by re-authenticating, so `azd auth login` as a suggestion was omitted.

Simulated error message (using different error code for testing):

image

### DIFF
--- a/cli/azd/pkg/auth/claims_test.go
+++ b/cli/azd/pkg/auth/claims_test.go
@@ -436,7 +436,7 @@ func TestNewReLoginRequiredError(t *testing.T) {
 
 	t.Run("nil_response_returns_false", func(t *testing.T) {
 		t.Parallel()
-		err, ok := newReLoginRequiredError(
+		err, ok := newActionableAuthError(
 			nil, nil, cloud.AzurePublic(), "")
 		assert.Nil(t, err)
 		assert.False(t, ok)
@@ -448,7 +448,7 @@ func TestNewReLoginRequiredError(t *testing.T) {
 			Error:            "server_error",
 			ErrorDescription: "something else",
 		}
-		err, ok := newReLoginRequiredError(
+		err, ok := newActionableAuthError(
 			resp, nil, cloud.AzurePublic(), "")
 		assert.Nil(t, err)
 		assert.False(t, ok)
@@ -460,7 +460,7 @@ func TestNewReLoginRequiredError(t *testing.T) {
 			Error:            "invalid_grant",
 			ErrorDescription: "AADSTS700082: expired",
 		}
-		err, ok := newReLoginRequiredError(
+		err, ok := newActionableAuthError(
 			resp,
 			[]string{"https://management.azure.com//.default"},
 			cloud.AzurePublic(),
@@ -477,7 +477,7 @@ func TestNewReLoginRequiredError(t *testing.T) {
 			Error:            "interaction_required",
 			ErrorDescription: "need consent",
 		}
-		err, ok := newReLoginRequiredError(
+		err, ok := newActionableAuthError(
 			resp,
 			[]string{"https://management.azure.com//.default"},
 			cloud.AzurePublic(),
@@ -493,7 +493,7 @@ func TestNewReLoginRequiredError(t *testing.T) {
 			Error:            "invalid_grant",
 			ErrorDescription: "expired",
 		}
-		err, ok := newReLoginRequiredError(
+		err, ok := newActionableAuthError(
 			resp,
 			[]string{
 				"https://management.azure.com//.default",
@@ -514,7 +514,7 @@ func TestNewReLoginRequiredError(t *testing.T) {
 			ErrorDescription: "AADSTS70043: expired",
 			ErrorCodes:       []int{70043},
 		}
-		err, ok := newReLoginRequiredError(
+		err, ok := newActionableAuthError(
 			resp, nil, cloud.AzurePublic(), "")
 		assert.True(t, ok)
 		require.Error(t, err)
@@ -527,7 +527,7 @@ func TestNewReLoginRequiredError(t *testing.T) {
 			ErrorDescription: "AADSTS700082: expired",
 			ErrorCodes:       []int{700082},
 		}
-		err, ok := newReLoginRequiredError(resp, nil, cloud.AzurePublic(), "")
+		err, ok := newActionableAuthError(resp, nil, cloud.AzurePublic(), "")
 		assert.True(t, ok)
 		require.Error(t, err)
 
@@ -543,7 +543,7 @@ func TestNewReLoginRequiredError(t *testing.T) {
 			ErrorDescription: "conditional access",
 			ErrorCodes:       []int{50005},
 		}
-		err, ok := newReLoginRequiredError(
+		err, ok := newActionableAuthError(
 			resp, nil, cloud.AzurePublic(), "")
 		assert.True(t, ok)
 		require.Error(t, err)
@@ -557,7 +557,7 @@ func TestNewReLoginRequiredError(t *testing.T) {
 			ErrorCodes:       []int{70043},
 		}
 		tenantID := "72f988bf-86f1-41af-91ab-2d7cd011db47"
-		err, ok := newReLoginRequiredError(
+		err, ok := newActionableAuthError(
 			resp, nil, cloud.AzurePublic(), tenantID)
 		assert.True(t, ok)
 		require.Error(t, err)
--- a/cli/azd/pkg/auth/claims_test.go
+++ b/cli/azd/pkg/auth/claims_test.go
@@ -431,7 +431,7 @@ func TestReLoginRequiredError_NonRetriable(t *testing.T) {
 	e.NonRetriable() // marker — should not panic
 }
 
-func TestNewReLoginRequiredError(t *testing.T) {
+func TestNewActionableAuthError_ClaimsScenarios(t *testing.T) {
 	t.Parallel()
 
 	t.Run("nil_response_returns_false", func(t *testing.T) {
@@ -533,6 +533,7 @@ func TestNewReLoginRequiredError(t *testing.T) {
 
 		var errWithSuggestion *internal.ErrorWithSuggestion
 		require.True(t, errors.As(err, &errWithSuggestion))
+		assert.Equal(t, "Login expired.", errWithSuggestion.Message)
 		assert.Contains(t, errWithSuggestion.Suggestion, "login expired")
 	})
 
@@ -547,6 +548,13 @@ func TestNewReLoginRequiredError(t *testing.T) {
 			resp, nil, cloud.AzurePublic(), "")
 		assert.True(t, ok)
 		require.Error(t, err)
+
+		errWithSuggestion, ok := errors.AsType[*internal.ErrorWithSuggestion](err)
+		require.True(t, ok)
+		assert.Equal(t, "Reauthentication required.", errWithSuggestion.Message)
+		require.Len(t, errWithSuggestion.Links, 1)
+		assert.Equal(t, "https://aka.ms/azd/troubleshoot/conditional-access-policy", errWithSuggestion.Links[0].URL)
+		assert.Equal(t, "Conditional Access policy troubleshooting", errWithSuggestion.Links[0].Title)
 	})
 
 	t.Run("tenant_id_included_in_suggestion", func(t *testing.T) {
--- a/cli/azd/pkg/auth/claims_test.go
+++ b/cli/azd/pkg/auth/claims_test.go
@@ -437,7 +437,7 @@ func TestNewActionableAuthError_ClaimsScenarios(t *testing.T) {
 	t.Run("nil_response_returns_false", func(t *testing.T) {
 		t.Parallel()
 		err, ok := newActionableAuthError(
-			nil, nil, cloud.AzurePublic(), "")
+			nil, nil, cloud.AzurePublic(), "", nil)
 		assert.Nil(t, err)
 		assert.False(t, ok)
 	})
@@ -449,7 +449,7 @@ func TestNewActionableAuthError_ClaimsScenarios(t *testing.T) {
 			ErrorDescription: "something else",
 		}
 		err, ok := newActionableAuthError(
-			resp, nil, cloud.AzurePublic(), "")
+			resp, nil, cloud.AzurePublic(), "", nil)
 		assert.Nil(t, err)
 		assert.False(t, ok)
 	})
@@ -465,6 +465,7 @@ func TestNewActionableAuthError_ClaimsScenarios(t *testing.T) {
 			[]string{"https://management.azure.com//.default"},
 			cloud.AzurePublic(),
 			"",
+			nil,
 		)
 		assert.True(t, ok)
 		require.Error(t, err)
@@ -482,6 +483,7 @@ func TestNewActionableAuthError_ClaimsScenarios(t *testing.T) {
 			[]string{"https://management.azure.com//.default"},
 			cloud.AzurePublic(),
 			"",
+			nil,
 		)
 		assert.True(t, ok)
 		require.Error(t, err)
@@ -501,6 +503,7 @@ func TestNewActionableAuthError_ClaimsScenarios(t *testing.T) {
 			},
 			cloud.AzurePublic(),
 			"",
+			nil,
 		)
 		assert.True(t, ok)
 		require.Error(t, err)
@@ -515,7 +518,7 @@ func TestNewActionableAuthError_ClaimsScenarios(t *testing.T) {
 			ErrorCodes:       []int{70043},
 		}
 		err, ok := newActionableAuthError(
-			resp, nil, cloud.AzurePublic(), "")
+			resp, nil, cloud.AzurePublic(), "", nil)
 		assert.True(t, ok)
 		require.Error(t, err)
 	})
@@ -527,7 +530,7 @@ func TestNewActionableAuthError_ClaimsScenarios(t *testing.T) {
 			ErrorDescription: "AADSTS700082: expired",
 			ErrorCodes:       []int{700082},
 		}
-		err, ok := newActionableAuthError(resp, nil, cloud.AzurePublic(), "")
+		err, ok := newActionableAuthError(resp, nil, cloud.AzurePublic(), "", nil)
 		assert.True(t, ok)
 		require.Error(t, err)
 
@@ -545,7 +548,7 @@ func TestNewActionableAuthError_ClaimsScenarios(t *testing.T) {
 			ErrorCodes:       []int{50005},
 		}
 		err, ok := newActionableAuthError(
-			resp, nil, cloud.AzurePublic(), "")
+			resp, nil, cloud.AzurePublic(), "", nil)
 		assert.True(t, ok)
 		require.Error(t, err)
 
@@ -566,7 +569,7 @@ func TestNewActionableAuthError_ClaimsScenarios(t *testing.T) {
 		}
 		tenantID := "72f988bf-86f1-41af-91ab-2d7cd011db47"
 		err, ok := newActionableAuthError(
-			resp, nil, cloud.AzurePublic(), tenantID)
+			resp, nil, cloud.AzurePublic(), tenantID, nil)
 		assert.True(t, ok)
 		require.Error(t, err)
 
--- a/cli/azd/pkg/auth/errors.go
+++ b/cli/azd/pkg/auth/errors.go
@@ -46,9 +46,16 @@ type TokenProtectionBlockedError struct {
 	errText string
 }
 
-// newReLoginRequiredError returns an error if the response indicates that the user needs to reauthenticate.
-// If it is not a reauthentication error, it returns false.
-func newReLoginRequiredError(
+// newActionableAuthError inspects an AAD error response and, if it matches a known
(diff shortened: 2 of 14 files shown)


cli/azd/pkg/auth/errors.go
```diff
@@ -37,6 +39,13 @@ type ReLoginRequiredError struct {
 	helpLink string
 }
 
+// TokenProtectionBlockedError indicates that the token request was blocked by
+// an organization's Conditional Access token protection policy (AADSTS530084), and until #7704 is addressed,
+// re-authenticating with azd won't help.
```
jongio: `reauthenticating` isn't in `cli/azd/.vscode/cspell-azd-dictionary.txt` (only `reauthentication` is), and that's what the cspell-lint job is failing on. Either add the word to the dictionary or rephrase, e.g. "and re-running `azd auth login` won't help."

cli/azd/pkg/auth/errors.go
```diff
@@ -37,6 +39,13 @@ type ReLoginRequiredError struct {
...
+type TokenProtectionBlockedError struct {
```
jongio, line 45: Two downstream callsites currently special-case `*ReLoginRequiredError` and won't recognize this new type:

1. `internal/grpcserver/server.go:278` (`wrapErrorWithSuggestion`) only returns `codes.Unauthenticated` for `*ReLoginRequiredError`. Token-protection blocks are auth failures too, so extensions inspecting the gRPC status will see them as generic errors instead of auth errors.

2. `internal/cmd/errors.go:50,308` (`MapError` and `classifySuggestionType`) tag `*ReLoginRequiredError` as `auth.login_required` in telemetry. `*TokenProtectionBlockedError` falls through both and lands in the generic `error.suggestion` bucket, which means we lose the ability to track this distinct failure mode in dashboards.

Worth either adding parallel `errors.AsType[*TokenProtectionBlockedError]` checks at each site (and a new telemetry code like `auth.token_protection_blocked`), or introducing a shared marker interface so future auth-error variants don't require touching every callsite.

cli/azd/pkg/auth/errors.go
```diff
@@ -49,6 +58,10 @@ func newReLoginRequiredError(
 		return nil, false
 	}
 
+	if err, ok := newTokenProtectionBlockedError(response, scopes); ok {
```
jongio, line 85: Side effect of routing through a different concrete type: `cmd/auth_login.go:317` (the `--only-check-status` path) and `cmd/auth_status.go:84` only suppress the raw `err.Error()` stderr print when the error is `*ReLoginRequiredError`. With this change, a token-protection block in those paths will print the raw AAD description (`AADSTS530084: ...`) to stderr in addition to the formatted Message + Suggestion + Links rendered by the suggestion pipeline. Consider extending those guards to also accept `*TokenProtectionBlockedError`.

jongio (commented):
A few cross-cutting concerns I'd address before merge:

1. CI is failing on cspell - `reauthenticating` isn't in the azd dictionary. Either rephrase the comment or add the word to `cli/azd/.vscode/cspell-azd-dictionary.txt`.

2. Three places in the codebase already special-case `*ReLoginRequiredError` (gRPC status mapping, two telemetry classifiers, and the auth command stderr suppression). The new `*TokenProtectionBlockedError` won't be caught by any of them, even though it's the same family of failure. Worth deciding whether each callsite should also recognize the new type, or whether a shared marker interface (e.g. `auth.NonRetriableLoginError`) would let future auth-error variants flow through without touching every site.

Details inline.

cli/azd/pkg/auth/claims_test.go
```diff
@@ -436,7 +436,7 @@ func TestNewReLoginRequiredError(t *testing.T) {
```
jongio, line 436: Nit: `TestNewReLoginRequiredError` here, plus `TestReLoginRequired` and `TestReLoginRequiredError` in `errors_test.go`, now exercise `newActionableAuthError` after the rename. Worth renaming the tests to match (e.g., `TestNewActionableAuthError*`) so future readers don't go hunting for a function that no longer exists.

jongio (commented):
Verified the three follow-ups from my earlier review are addressed. Downstream call sites now route through the new `AuthInteractionError` marker interface, the stderr-suppression guards in `auth_login.go` and `auth_status.go` cover both error types, and the AAD-code precedence check in `newTokenProtectionBlockedError` runs before the generic switch (nicely covered by `TestNewReLoginRequiredError_TokenProtectionTakesPrecedence`). Sealed-interface design via unexported marker is the right call for future variants. One small naming nit inline.

wbreza (commented):
Reviewing as wbreza. Prior concerns from @jongio (cspell, gRPC codes.Unauthenticated mapping, stderr suppression, telemetry split) all verified resolved by commit 506c2a0's AuthInteractionError interface adoption. New findings below — nothing blocking; mostly test coverage gaps and a design question on gRPC code semantics.

### 🟠 High (1)

**H1 — internal/grpcserver/server.go:278 — codes.Unauthenticated collapses two semantically distinct errors**

Both ReLoginRequiredError (fixable by re-auth) and TokenProtectionBlockedError (re-auth won't help — user must contact IT admin) now route through AuthInteractionError to codes.Unauthenticated. Extensions consuming the gRPC contract can't distinguish them and may wrongly prompt users to re-login for conditional-access blocks. This is also directly relevant to #7791 (extension error suggestion propagation), which cannot solve the distinction downstream if both errors share the same status code.

Two options:
- **Preferred:** Override specifically for TokenProtectionBlockedError → codes.FailedPrecondition or codes.PermissionDenied (policy block, not an auth credential failure)
- **Minimum:** Add a comment in wrapErrorWithSuggestion documenting the limitation so #7791 authors are aware

### 🟡 Medium (6)

**M1 — internal/cmd/errors.go:309–312 — Dead code in classifySuggestionType**

classifySuggestionType is only called from the ErrorWithSuggestion branch of MapError. But MapError checks rrors.AsType[*auth.TokenProtectionBlockedError](err) **before** that branch, and the outer check traverses ErrorWithSuggestion.Unwrap() — so it always matches first. The new *TokenProtectionBlockedError check in classifySuggestionType (and the pre-existing *ReLoginRequiredError one next to it) is unreachable. Remove or mark as defensive with a comment.

**M2 — internal/cmd/errors.go:50–55 — Missing guidepost comment for the concrete-type exception**

Design is intentionally split: 3 sites use AuthInteractionError interface (forward-extensible), while MapError/classifySuggestionType use concrete types (per-type telemetry codes). This is correct, but undocumented. A future contributor adding a 3rd AuthInteractionError (e.g., MFARequiredError) will update the 3 interface sites and miss MapError, silently misclassifying telemetry as rror.suggestion. Suggest adding a comment above the concrete-type checks:
`go
// Auth errors are checked by concrete type (not the AuthInteractionError
// interface) because each type has a distinct telemetry code. Adding a new
// AuthInteractionError type? Add a matching branch here as well.
`

**M3 — pkg/auth/errors.go:64–67 — Inconsistent ErrorWithSuggestion field usage in the renamed function**

The pre-existing ReLoginRequiredError branch returns an ErrorWithSuggestion with only Err + Suggestion. The login message and the helpLink URL are both packed into the Suggestion string. The new TokenProtectionBlockedError branch correctly populates Message, Suggestion, and Links as separate fields. Since the function is being modified (and renamed) in this PR, it's a good moment to align both branches — per AGENTS.md guidance on populating all relevant ErrorWithSuggestion fields.

**M4 — internal/grpcserver/server_test.go — Missing gRPC test for TokenProtectionBlockedError**

Test_wrapErrorWithSuggestion has a case for ReLoginRequiredError with suggestion returns Unauthenticated (line ~281) but no corresponding case for TokenProtectionBlockedError. The PR's key behavioral claim — that token-protection errors return codes.Unauthenticated over gRPC — is untested at the gRPC boundary. Easy to add mirroring the existing case.

**M5 — internal/cmd/errors_test.go — Missing MapError telemetry test for TokenProtectionBlockedError**

WithReLoginRequiredError asserts rrCode = \"auth.login_required\" but there's no WithTokenProtectionBlockedError asserting rrCode = \"auth.token_protection_blocked\". This test would also cover the critical path where rrors.AsType[*TokenProtectionBlockedError] traverses ErrorWithSuggestion.Unwrap() — important because MapError correctness depends on that traversal working.

**M6 — pkg/auth/errors_test.go — No assertion that AuthInteractionError interface works through the wrapper**

All 3 production callsites do rrors.AsType[auth.AuthInteractionError](err) on a *ErrorWithSuggestion wrapping a *TokenProtectionBlockedError. This traversal works via ErrorWithSuggestion.Unwrap() but is never asserted in tests. A single line in TestTokenProtectionBlockedError would pin the contract:
`go
_, isInteractionErr := errors.AsType[AuthInteractionError](err)
require.True(t, isInteractionErr, \"TokenProtectionBlockedError should satisfy AuthInteractionError through the ErrorWithSuggestion wrapper\")
`

### 🟢 Low (3)

**L1 — pkg/auth/errors.go — Detection edge case when rror_codes is absent**

Detection is slices.Contains(response.ErrorCodes, 530084). If AAD ever returns AADSTS530084 only in ErrorDescription (nil ErrorCodes), the fall-through produces ReLoginRequiredError and wrongly tells users to re-login. AAD always includes rror_codes in practice, so this is low-priority robustness — optional secondary check on ErrorDescription if you want belt-and-suspenders.

**L2 — pkg/auth/errors.go — Possible gofmt misalignment in marker methods**

`go
func (*ReLoginRequiredError) isAuthInteractionError()        {}
func (*TokenProtectionBlockedError) isAuthInteractionError() {}
`
gofmt emits exactly one space before {}. CI is green so this may already be clean in the actual file, but worth a quick gofmt -l ./pkg/auth/errors.go to confirm.

**L3 — pkg/auth/errors.go — Doc comment on AuthInteractionError recommends a wrong pattern**

> Callers that just need to know "this is an actionable auth failure" should **type-assert** against this interface…

A bare type assertion would fail through *ErrorWithSuggestion. All 3 callsites correctly use rrors.AsType[AuthInteractionError](err). Suggest updating the doc comment to recommend rrors.AsType explicitly so future contributors don't write the wrong code.

---

Not blocking. H1 is worth a discussion (perhaps in coordination with #7791); the rest are small cleanups / test additions.

JeffreyCA:
Addressed the latest review follow-ups:

- renamed the stale `ReLoginRequired*` tests to match `newActionableAuthError`
- split gRPC auth semantics so token-protection errors return `PermissionDenied` while re-login cases stay `Unauthenticated`
- updated `azure.ai.agents` gRPC error translation for the new token-protection path
- added coverage for gRPC mapping, telemetry classification, and `AuthInteractionError` unwrapping through `ErrorWithSuggestion`
- aligned the re-login `ErrorWithSuggestion` shape (`Message`/`Links`) and clarified the telemetry/interface comments/docs

I intentionally left the `classifySuggestionType` auth branches as-is and did not add the speculative `ErrorDescription` fallback for `AADSTS530084`.

wbreza (commented):
Thanks for the thorough response to the prior round, @JeffreyCA — the gRPC code split, `ErrorLink` refactor, `Message` population, and added telemetry/interface tests all land cleanly. Re-review covers only the delta in `5137cff`.

## ✅ Resolved from prior review
- **H1** — gRPC `Unauthenticated` collapse: `grpcAuthCode()` now returns `codes.PermissionDenied` for `*TokenProtectionBlockedError`, with regression tests covering both paths.
- `helpLink` → structured `*ErrorLink` (URL + "Conditional Access policy troubleshooting" title).
- `scenarioMessage()` populates `ErrorWithSuggestion.Message` ("Login expired." / "Reauthentication required.").
- `TokenProtectionBlockedError` interface-satisfaction assertion + `token_protection_blocked` telemetry mapping test.
- Doc comment updated to recommend `errors.AsType[T]`.

## 🔴 New finding

### H-new — Extension consumers don't classify `PermissionDenied` as auth

The new server-side mapping is correct, but neither extension-side decoder treats the new code as auth:

- `cli/azd/pkg/azdext/extension_error.go:130` — only matches `codes.Unauthenticated`
- `cli/azd/extensions/azure.ai.agents/internal/exterrors/errors.go:150,175` — same

Result: a `TokenProtectionBlockedError` traversing gRPC arrives at the extension as a generic error, losing the `LocalErrorCategoryAuth` / `auth_failed` classification — partially undoing the original intent of H1. Compounding this, the gRPC server already emits `codes.PermissionDenied` for unrelated capability errors (`framework_service.go:64`, `service_target_service.go:68`, `event_service.go:77`), so extensions can't safely treat all `PermissionDenied` as auth.

**Two viable fixes:**
1. **Update the consumer side** (in this PR or a companion — possibly #7791?) so that `PermissionDenied` + the azd auth-suggestion payload is classified as auth.
2. **Keep both on `Unauthenticated`** and let the extension distinguish via `Details`/message metadata — `Unauthenticated` per gRPC spec covers any "request does not have valid authentication credentials," which includes policy-blocked tokens.

If #7791 already handles (1), feel free to mark this as out-of-scope here.

## 🟢 Low

- **L-new-1** — `pkg/auth/errors.go` `scenarioMessage`: `strings.ToUpper(scenario[:1]) + scenario[1:]` is byte-sliced. Safe today (all scenarios ASCII) but a future non-ASCII scenario string would mojibake/panic. Consider a `utf8.DecodeRuneInString` based helper.
- **L-new-2** — Add a one-line doc on `grpcAuthCode` explaining why `PermissionDenied` (not `FailedPrecondition`) was chosen. Both fit per grpc.io guidance; PermissionDenied is defensible but the rationale lives only in commit context today.
- **L-new-3** — `"Reauthentication required."` for `TokenProtectionBlockedError` reads slightly off given the doc comment notes "re-authenticating with azd won't help." Something like `"Sign-in blocked by policy."` would match the link title and the actual remediation (admin action). The `Suggestion` text + link carry the real guidance, so this is a polish item.

jongio (commented):
Re-checked 5137cff. The naming nit from my earlier review is addressed, and @wbreza's H1 (PermissionDenied split for token protection) landed cleanly with regression tests on both gRPC paths.

No new findings from me. @wbreza's H-new about extension-side classification (`extension_error.go` and `azure.ai.agents/internal/exterrors/errors.go` still only match `codes.Unauthenticated`, so `TokenProtectionBlockedError` now arrives at extensions unclassified) is the substantive open thread. I don't have anything to add there beyond the two options outlined.

JeffreyCA:
Addressed @wbreza's feedback by moving the auth subtype distinction into structured gRPC details instead of the top-level status code.
 
 Changes in this update:
 - reverted auth-interaction errors back to `codes.Unauthenticated`
 - attached `google.rpc.ErrorInfo` (`azd.auth`) to distinguish not-logged-in, login-required, and token-protection-blocked cases
 - updated both consumer paths (`pkg/azdext/extension_error.go` and `azure.ai.agents/internal/exterrors/errors.go`) to use those details
 - kept message-based fallback parsing for older hosts
 - added coverage for host emission plus consumer-side `login_required`, `token_protection_blocked`, and fallback `auth_failed` mapping
 
This preserves broad auth detection while still allowing extensions to distinguish the remediation path.

cli/azd/pkg/auth/errors.go
```diff
@@ -34,12 +36,43 @@ type ReLoginRequiredError struct {
...
+	helpLink *errorhandler.ErrorLink
 }
 
-// newReLoginRequiredError returns an error if the response indicates that the user needs to reauthenticate.
-// If it is not a reauthentication error, it returns false.
-func newReLoginRequiredError(
+// TokenProtectionBlockedError indicates that the token request was blocked by
+// an organization's Conditional Access token protection policy (AADSTS530084), and until #7704 is addressed,
+// re-running `azd auth login` won't help.
+type TokenProtectionBlockedError struct {
```
weikanglim: Given that this error and its suggested handling are fairly short-lived, would it make sense to avoid introducing new types for this fix?

One option could be to handle this directly in `azd_credential.go`—check for the condition upfront, use `AuthFailedError` to match on the specific error code, and wrap the credential accordingly.

That way we keep the change more localized and avoid expanding the type surface unless we see this pattern repeat.

weikanglim: For gRPC handling, I think it makes sense to pass through the Entra error code from `AuthFailedError`. My instinct is to avoid having azd own auth business logic unless we think this particular behavior is intentionally azd-specific.

cli/azd/pkg/azdext/extension_error.go
```diff
@@ -143,6 +144,19 @@ func WrapError(err error) *ExtensionError {
...
 
+func authLocalErrorCode(st *status.Status) string {
+	switch AuthErrorReason(st) {
+	case AuthErrorReasonNotLoggedIn:
+		return "not_logged_in"
+	case AuthErrorReasonLoginRequired:
+		return "login_required"
+	case AuthErrorReasonTokenProtectionBlocked:
+		return "token_protection_blocked"
+	default:
```
weikanglim: The new addition of `token_protection_blocked` to these error codes make it feel somewhat arbitrary. I’d suggest passing through the Entra error code from `AuthFailedError` instead, so we preserve the original semantics and avoid redefining them on our side.

weikanglim (commented):
@JeffreyCA Great work on diving deep into the issue! I left a few comments with ideas on how we could clean it up more. Happy to discuss any further on them.

cli/azd/pkg/auth/errors.go
```diff
@@ -59,18 +96,73 @@ func newReLoginRequiredError(
...
+		return suggestionErr, true
 	}
 
 	return nil, false
 }
 
+const (
+	conditionalAccessDocsLink = "https://aka.ms/TBCADocs"
+	// #nosec G101 -- documentation URL, not a credential.
+	tokenProtectionFAQLink = "https://aka.ms/TokenProtectionFAQ#troubleshooting"
```
weikanglim, line 93: I tried this in an InPrivate window and was prompted to log in, so I’m guessing it may be an internal site. I’m not entirely sure on the details, but wanted to flag it in case it’s relevant.

cli/azd/pkg/auth/errors.go
```diff
@@ -59,18 +96,73 @@ func newReLoginRequiredError(
...
+			suggestionErr.Links = []errorhandler.ErrorLink{*err.helpLink}
+		}
+		return suggestionErr, true
 	}
 
 	return nil, false
 }
 
+const (
+	conditionalAccessDocsLink = "https://aka.ms/TBCADocs"
```
weikanglim, line 91: FWIW -- this link wasn't that helpful for me.

cli/azd/pkg/auth/errors.go
JeffreyCA: Addressed in the latest push — gRPC auth handling now matches `*auth.AuthFailedError` directly and passes through the originating Entra reason as `AADSTS<code>` in `azd.auth` `ErrorInfo`.

cli/azd/pkg/azdext/extension_error.go
JeffreyCA: Updated in the latest push — extension-side auth codes stay coarse (`not_logged_in`, `login_required`, `auth_failed`), and the raw Entra reason remains available via `AuthErrorReason(st)` from the gRPC `ErrorInfo`.

cli/azd/pkg/auth/claims_test.go
JeffreyCA: Renamed in the latest push — the claims test now uses `TestNewActionableAuthError_ClaimsScenarios` to match the helper rename.

cli/azd/pkg/auth/errors.go
JeffreyCA: Yes, the reports so far have all come from internal users and there isn't really any public guidance available to offer at this time

cli/azd/pkg/auth/errors.go
JeffreyCA: Agreed, this link is included in the original error response:

```
ERROR: AADSTS530084: Access has been blocked by conditional access token protection policy configured by this organization. To learn more, see https://aka.ms/TBCADocs. Trace ID: <redacted> Correlation ID: <redacted> Timestamp: 2026-04-13 19:04:12Z
```

weikanglim (approved):

PR #7811, opened 2026-04-20
feat: Add warnings for role assignment issues in RBAC checks

### LINKED ISSUE #7810: Your identity does not have permission to write role assignments on Foundry Project
Image

Make this a warning instead of a error.

therealjohn:
Dont block deployment of agents on permissions, just warn the user. Fixes #7810

### DIFF
--- a/cli/azd/extensions/azure.ai.agents/internal/project/agent_identity_rbac.go
+++ b/cli/azd/extensions/azure.ai.agents/internal/project/agent_identity_rbac.go
@@ -341,15 +341,13 @@ func ensureSingleAgentRBAC(
 	if err != nil {
 		if respErr, ok := errors.AsType[*azcore.ResponseError](err); ok &&
 			respErr.StatusCode == http.StatusForbidden {
-			// Write to stderr with warning color so it survives azd's deployment preview capture.
-			fmt.Fprintf(os.Stderr, "%s\n",
-				output.WithWarningFormat(
-					"Could not assign 'Azure AI User' to agent identity '%s' (403 Forbidden).\n"+
-						"    The agent may not have access to the Foundry Project until this role is assigned.\n"+
-						"    Re-run after granting role assignment write, or set AZD_AGENT_SKIP_ROLE_ASSIGNMENTS=true.",
-					agentName,
-				),
-			)
+			// Write with warning color so it appears as a yellow warning, not a red error.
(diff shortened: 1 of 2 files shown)


rajeshkamal5050 (dismissed):

rajeshkamal5050 (approved):

cli/azd/extensions/azure.ai.agents/internal/project/agent_identity_rbac.go
```diff
@@ -335,6 +339,31 @@ func ensureSingleAgentRBAC(
...
+			respErr.StatusCode == http.StatusForbidden {
+			manualRemediationCommand := fmt.Sprintf(
+				"az role assignment create --assignee-object-id %s --assignee-principal-type ServicePrincipal --role %s --scope %s",
+				strconv.Quote(principalID),
+				strconv.Quote("Azure AI User"),
+				strconv.Quote(info.ProjectScope),
+			)
+
+			// Write with warning color so it appears as a yellow warning, not a red error.
+			fmt.Printf("%s\n", output.WithWarningFormat(
```
jongio, line 352: This warning runs inside a goroutine (called at line 244 in the concurrent agent RBAC loop). If multiple agents hit 403 simultaneously, the multi-line output could interleave with other goroutine output. The old code returned errors to the `results` slice and printed them serially after `wg.Wait()`.

Consider collecting warnings the same way:

```go
type agentResult struct {
    name    string
    err     error
    warning string // collect warning text here
}
```

Then print warnings after `wg.Wait()` in the serial reporting loop. This keeps the warning-not-error behavior while avoiding interleaved output.

jongio (commented):
Two post-merge observations:

1. `ensureSingleAgentRBAC` is called from goroutines (line 244) but the new warning writes multi-line output to stdout via `fmt.Printf`. If multiple agents hit 403 simultaneously, the output could interleave. The old code returned errors to the `results` slice and printed serially after `wg.Wait()`. Consider collecting warnings the same way - add a `warning string` field to `agentResult` and print after the wait.

2. Three error codes in `codes.go` are now unused: `CodeDeveloperMissingAIUserRole`, `CodeDeveloperMissingRoleAssignWriteRole`, and `CodeDeveloperMissingACRRole`. Minor dead code to clean up.

PR #7819, opened 2026-04-20
Exempt ARM-allow-listed types from reserved-name preflight check and fix duplicated warnings

### LINKED ISSUE #7805: [Issue] Preflight surfaces unfixable "reserved word WINDOWS" warnings for Private DNS Zones

**Output from `azd version`**

```
azd version 1.24.1 (commit f2f23c94e0cb04f7043eb4bffeaa60c011b7bb50) (stable)
```

**Describe the bug**

After upgrading azd, `azd provision` now emits ARM preflight warnings for Private DNS Zones whose canonical names contain the substring `WINDOWS`. These names are Microsoft-mandated - Azure Private Link's DNS integration only works when the zone matches the exact name in the [private endpoint DNS reference](https://learn.microsoft.com/azure/private-link/private-endpoint-dns). Example zones that trigger the warning: `privatelink.search.windows.net`, `privatelink.redis.cache.windows.net`. The warning also fires a second time per zone on its `Microsoft.Network/privateDnsZones/virtualNetworkLinks` child, because the ARM child name is `{parent-zone}/{child-link}` and preflight scans the concatenated name.

Example output:

```
 (!) Warning: resource "privatelink.search.windows.net" (Microsoft.Network/privateDnsZones) contains the reserved word "WINDOWS". See https://learn.microsoft.com/azure/azure-resource-manager/templates/error-reserved-resource-name.
  (!) Warning: resource "privatelink.search.windows.net/privatelink.search.windows.net-link" (Microsoft.Network/privateDnsZones/virtualNetworkLinks) contains the reserved word "WINDOWS". See https://learn.microsoft.com/azure/azure-resource-manager/templates/error-reserved-resource-name.
  (!) Warning: resource "privatelink.search.windows.net/privatelink.search.windows.net-link" (Microsoft.Network/privateDnsZones/virtualNetworkLinks) contains the reserved word "WINDOWS". See https://learn.microsoft.com/azure/azure-resource-manager/templates/error-reserved-resource-name.
  (!) Warning: resource "privatelink.redis.cache.windows.net" (Microsoft.Network/privateDnsZones) contains the reserved word "WINDOWS". See https://learn.microsoft.com/azure/azure-resource-manager/templates/error-reserved-resource-name.
  (!) Warning: resource "privatelink.redis.cache.windows.net/privatelink.redis.cache.windows.net-link" (Microsoft.Network/privateDnsZones/virtualNetworkLinks) contains the reserved word "WINDOWS". See https://learn.microsoft.com/azure/azure-resource-manager/templates/error-reserved-resource-name.
  (!) Warning: resource "privatelink.redis.cache.windows.net/privatelink.redis.cache.windows.net-link" (Microsoft.Network/privateDnsZones/virtualNetworkLinks) contains the reserved word "WINDOWS". See https://learn.microsoft.com/azure/azure-resource-manager/templates/error-reserved-resource-name.
```

These are false positives and unfixable by the consumer:

- The zone names cannot be changed. No alternative name preserves private-endpoint DNS resolution.
- ARM's actual deployment accepts these names (Private DNS Zones are allow-listed at deploy time). The deployment always succeeds - only preflight's reserved-word scanner flags them.
- No user-side suppression mechanism works: `#disable-next-line` does not apply (no BCP code), `bicepconfig.json` cannot suppress it (it is not a linter rule), and dynamic name construction via `environment().suffixes.storage` only defeats Bicep's compile-time check - preflight resolves the expression before scanning, so the warning still fires on the resolved name.

The only available workaround is `azd config set provision.preflight off`, which disables all preflight value globally just to silence a handful of zone-name false positives.

**Previous behavior:** earlier azd releases did not surface these ARM preflight warnings to the user. The warnings themselves have always existed in ARM's validation response; the change is that azd now prints them. This is a behavioral regression for pipelines that treat warnings as errors - the infrastructure and Bicep code are unchanged, but `azd provision` now fails where it previously succeeded.

**To Reproduce**

1. Declare a Private DNS Zone and VNet link using the Microsoft-documented name for any private-link service whose zone ends in `.windows.net`:

```bicep
   param vnetId string

   resource searchDnsZone 'Microsoft.Network/privateDnsZones@2020-06-01' = {
     name: 'privatelink.search.windows.net'
     location: 'global'
   }

   resource searchDnsZoneLink 'Microsoft.Network/privateDnsZones/virtualNetworkLinks@2020-06-01' = {
     parent: searchDnsZone
     name: '${searchDnsZone.name}-link'
     location: 'global'
     properties: {
       registrationEnabled: false
       virtualNetwork: { id: vnetId }
     }
   }
```

2. Run `azd provision`.
3. Observe the four warning lines (parent + link × 2 scan passes). The deployment itself still succeeds.

Same behavior occurs for `privatelink.redis.cache.windows.net` and any other private-link zone whose canonical name contains `windows`.

**Expected behavior**

One of the following:

1. Preflight suppresses reserved-word warnings for resource types where ARM permits the name server-side - specifically `Microsoft.Network/privateDnsZones` and `Microsoft.Network/privateDnsZones/virtualNetworkLinks`.
2. `azure.yaml` gains a `provision.preflight.ignoreRules` (or similar) allowlist so consumers can silence specific preflight rule IDs without disabling preflight globally. Note: 1.23.14 already introduced per-rule telemetry IDs (#7396), so the underlying rule-identification is already in place.
3. azd downgrades known-benign ARM preflight warnings to info-level output that does not cause warnings-as-errors CI configurations to fail.

At minimum, warnings emitted against resource names that Microsoft's own reference documentation mandates should not be surfaced by default, and should never cause a pipeline failure.

**Environment**

* OS: Ubuntu 22.04
* Shell: bash 5.2
* IaC: Bicep CLI version 0.42.1 (caea9302e8)
* Scenario: Private endpoints + Private DNS Zones for Azure AI Search, Azure Cache for Redis, and similar services whose canonical zone names end in `.windows.net`

**Additional context**

- Preflight feature was introduced in the 1.23.x line (March 2026 releases 1.23.7-1.23.13) and has been baseline since, per [Azure Developer CLI (azd) - March 2026](https://devblogs.microsoft.com/azure-sdk/azure-developer-cli-azd-march-2026/).
- Per-rule preflight telemetry with unique rule/diagnostic IDs was added in 1.23.14 (#7396).
- ARM reserved-name reference: [Resolve errors for reserved resource names](https://learn.microsoft.com/azure/azure-resource-manager/troubleshooting/error-reserved-resource-name) - notes that `WINDOWS` cannot appear as a substring in resource names, but does not mention the Private DNS Zone exception that ARM applies at deploy time.
- Current workaround: `azd config set provision.preflight off`. Acceptable temporarily but sacrifices preflight's value (AI quota checks, parameter validation, etc.) to silence ~6-10 specific zone-name patterns.

JeffreyCA:
Fixes #7805

### Problem

The [reserved keywords validate checks](https://github.com/Azure/azure-dev/pull/7746) emit false-positive warnings on resource types where ARM **does not** enforce the reserved-word rule, such as Private Link resources.

Two distinct bugs:

1. **No exception list for ARM-allow-listed resource types.** The [reserved-name docs][docs] scope the rule to "resources with an accessible endpoint." DNS zones, RGs, deployments, policy assignments, etc. are all accepted server-side regardless of name.

2. **Compound child names produce duplicate warnings.** Bicep lists each child separately with a compound `parent/child` name. The original code split on `/` and scanned every segment, so reserved words in the parent re-fired on every child too.

[docs]: https://learn.microsoft.com/azure/azure-resource-manager/templates/error-reserved-resource-name

### Fix

This PR introduces a denylist (`azureReservedResourceNameExemptTypes`) of resource types to skip enforcing the reserved name rule.

Second, this PR rewrites `findReservedResourceNameViolations` to only scan the trailing `/`-segment of each resource name. Because Bicep enumerates the parent as its own resource, the parent's name is already evaluated separately, and re-scanning the parent prefix on every child only
duplicated warnings without finding anything new.

**Without fix**
image

**With fix**

image

#### Exempt resource types

This list was formed by analyzing Awesome Azd template bicep files.

**FQDN-shaped names**
- `Microsoft.Network/privateDnsZones`, `dnsZones`, `dnsForwardingRulesets`, `dnsResolvers`, `firewallPolicies`

**Internal-only names (no public endpoint)**
- `Microsoft.Resources/resourceGroups`, `deployments`, `deploymentScripts`
- `Microsoft.Authorization/{roleAssignments, roleDefinitions, policyAssignments, policyDefinitions, policySetDefinitions, locks}`
- `Microsoft.Insights/diagnosticSettings`

**User-labeled child resources**
- `Microsoft.Network/networkSecurityGroups` (covers parent + `securityRules`)
- `Microsoft.KeyVault/vaults/{secrets, keys, certificates, accessPolicies}`
- `Microsoft.OperationalInsights/workspaces/dataSources`

**Convention-named (publisher / connector / service tag)**
- `Microsoft.Compute/virtualMachines/extensions` *(e.g. `Microsoft.Insights.LogAnalyticsAgent`)*
- `Microsoft.Compute/virtualMachineScaleSets/extensions`
- `Microsoft.Web/connections` *(e.g. `office365`, `sharepointonline`)*
- `Microsoft.Network/virtualNetworks/subnets/delegations` *(e.g. `Microsoft.Web/serverFarms`)*

#### Why a denylist

A denylist was used instead of an allowlist because there is no public catalog of which resource types ARM actually enforces the rule on, so either approach will be wrong sometimes.

A denylist miss is a noisy false positive that the user reports and we fix. An allowlist miss is silent: the warning never fires and the user only learns the name is invalid when ARM rejects the deployment a minute later, so the gap can persist indefinitely with no signal.

### DIFF
--- a/cli/azd/pkg/infra/provisioning/bicep/bicep_provider.go
+++ b/cli/azd/pkg/infra/provisioning/bicep/bicep_provider.go
@@ -115,6 +115,64 @@ var azureReservedResourceNameContainsMatches = []string{
 
 const azureReservedResourceNamePrefixMatch = "LOGIN"
 
+// azureReservedResourceNameExemptTypes lists resource type prefixes for which
+// the reserved-word check is skipped. ARM does not enforce the reserved-word
+// rule on these types — either because they have no public endpoint, or because
+// their names are FQDN/convention-shaped by design (e.g. DNS zones, VM
+// extensions named "Publisher.Type").
+//
+// Each entry matches the exact type or any descendant child type
+// (case-insensitive). One entry like "Microsoft.Network/privateDnsZones"
+// covers the parent zone plus every record/link child without enumerating them.
+var azureReservedResourceNameExemptTypes = []string{
+	// FQDN-shaped names.
+	"Microsoft.Network/privateDnsZones",
+	"Microsoft.Network/dnsZones",
+	"Microsoft.Network/dnsForwardingRulesets",
+	"Microsoft.Network/dnsResolvers",
+	"Microsoft.Network/firewallPolicies",
+
+	// Internal-only names (no public endpoint).
+	"Microsoft.Resources/resourceGroups",
+	"Microsoft.Resources/deployments",
+	"Microsoft.Resources/deploymentScripts",
+	"Microsoft.Authorization/roleAssignments",
+	"Microsoft.Authorization/roleDefinitions",
+	"Microsoft.Authorization/policyAssignments",
+	"Microsoft.Authorization/policyDefinitions",
+	"Microsoft.Authorization/policySetDefinitions",
+	"Microsoft.Authorization/locks",
+	"Microsoft.Insights/diagnosticSettings",
+
+	// User-labeled child resources (e.g. NSG rule "AllowMicrosoftContainerRegistryOutbound").
+	"Microsoft.Network/networkSecurityGroups",
+	"Microsoft.KeyVault/vaults/secrets",
+	"Microsoft.KeyVault/vaults/keys",
+	"Microsoft.KeyVault/vaults/certificates",
+	"Microsoft.KeyVault/vaults/accessPolicies",
+	"Microsoft.OperationalInsights/workspaces/dataSources",
+
+	// Convention-named resources (names mirror a publisher, connector, or service tag).
+	"Microsoft.Compute/virtualMachines/extensions", // e.g. "Microsoft.Insights.LogAnalyticsAgent"
+	"Microsoft.Compute/virtualMachineScaleSets/extensions",
+	"Microsoft.Web/connections",                             // e.g. "office365", "sharepointonline"
+	"Microsoft.Network/virtualNetworks/subnets/delegations", // e.g. "Microsoft.Web/serverFarms"
+}
+
+// isReservedNameCheckExempt reports whether the reserved-word check should be
+// skipped for the given resource type. Matching is case-insensitive and applies
+// to the type itself or any descendant child type.
+func isReservedNameCheckExempt(resourceType string) bool {
+	normalized := strings.ToLower(resourceType)
+	for _, exempt := range azureReservedResourceNameExemptTypes {
+		exemptLower := strings.ToLower(exempt)
+		if normalized == exemptLower || strings.HasPrefix(normalized, exemptLower+"/") {
+			return true
+		}
+	}
+	return false
+}
+
 // BicepProvider exposes infrastructure provisioning using Azure Bicep templates
 type BicepProvider struct {
 	// Options that are available after Initialize()
@@ -2489,6 +2547,9 @@ func (p *BicepProvider) checkReservedResourceNames(
 	const docsLink = "https://learn.microsoft.com/azure/azure-resource-manager/templates/error-reserved-resource-name"
 
 	for _, resource := range valCtx.SnapshotResources {
+		if isReservedNameCheckExempt(resource.Type) {
+			continue
+		}
 		for _, v := range findReservedResourceNameViolations(resource.Name) {
 			resourceName := output.WithHighLightFormat("%q", resource.Name)
 			resourceType := output.WithGrayFormat("(%s)", resource.Type)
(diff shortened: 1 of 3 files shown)


vhvb1989 (approved):

jongio (commented):
Two things worth looking at before merge.

cli/azd/pkg/infra/provisioning/bicep/bicep_provider.go
```diff
@@ -115,6 +115,64 @@ var azureReservedResourceNameContainsMatches = []string{
...
+	"Microsoft.KeyVault/vaults/secrets",
+	"Microsoft.KeyVault/vaults/keys",
+	"Microsoft.KeyVault/vaults/certificates",
+	"Microsoft.KeyVault/vaults/accessPolicies",
+	"Microsoft.OperationalInsights/workspaces/dataSources",
+
+	// Convention-named resources (names mirror a publisher, connector, or service tag).
+	"Microsoft.Compute/virtualMachines/extensions", // e.g. "Microsoft.Insights.LogAnalyticsAgent"
+	"Microsoft.Compute/virtualMachineScaleSets/extensions",
+	"Microsoft.Web/connections",                             // e.g. "office365", "sharepointonline"
```
jongio, line 158: cspell-lint is failing on `sharepointonline` in this comment (run 24682620432, `pkg/infra/provisioning/bicep/bicep_provider.go:158:81`). Quickest fix that matches how the repo handles similar cases (e.g. `cli/azd/AGENTS.md`, `ext/vscode/.github/copilot-instructions.md`): add a `// cspell:ignore sharepointonline` directive near the top of this file, or drop the offending word from the comment.

cli/azd/pkg/infra/provisioning/bicep/bicep_provider.go
```diff
@@ -2711,11 +2772,18 @@ type reservedNameViolation struct {
...
-// returned to avoid fix-rerun cycles during preflight.
+// findReservedResourceNameViolations returns every reserved-word violation
+// found in resourceName. A single resource can produce multiple violations
+// (for example "LoginMicrosoftApp" triggers both the LOGIN prefix and the
+// MICROSOFT substring rule), so all matches are returned to avoid fix-rerun
+// cycles during preflight.
+//
+// Only the trailing `/`-delimited segment of the name is evaluated. For child
+// resources, ARM emits the name as `<parent>/<child>` (and `<grandparent>/...`
+// for deeper children), but the parent is also enumerated separately in the
```
jongio, line 2783: The doc says "the parent is also enumerated separately in the snapshot and is checked on its own pass." That holds when the parent is declared in the same bicep module. If a child uses `name: 'parentName/childName'` where `parentName` is an `existing`/externally-declared parent, that parent isn't in `SnapshotResources` and a reserved word in its segment now goes unreported (the old code would have flagged it). That's probably the right call (users can't rename external parents anyway), but worth tightening this comment to say the parent is evaluated on its own pass *when declared in the same module*, so future readers don't assume full parent-segment coverage.

PR #7837, opened 2026-04-21
Add registry versioning and schema evolution support

### LINKED ISSUE #6233: Add registry versioning and schema evolution support
## Description

Add versioning to the registry schema to support future evolution and enable azd to detect incompatible registry versions, providing upgrade guidance when needed.

## Acceptance Criteria

- [ ] Add version field to registry schema (e.g., `"schemaVersion": "1.0"`)
- [ ] Implement version detection in azd when loading registries
- [ ] Handle incompatible registry versions gracefully with clear error messages
- [ ] Provide upgrade guidance when registry version is incompatible
- [ ] Update registry schema documentation
- [ ] Support backward compatibility for existing registries (assume version 1.0 if not specified)
- [ ] Plan for future schema evolution scenarios

## Schema Evolution Scenarios

This prepares the system for more complex registry formats beyond a single JSON file, such as:
- Breaking changes to registry structure
- New required fields
- Changed field semantics
- Migration from JSON to different format

## Benefits

- Enables safe evolution of registry schema over time
- Provides clear upgrade paths for registry maintainers
- Prevents cryptic errors when using incompatible registries
- Supports future enhancements to registry capabilities

## Parent Issue

- #5767 (Extension enhancements EPIC)

wbreza:
## Description

Add registry versioning and schema evolution support to enable safe evolution of the extension registry format.

This introduces a `schemaVersion` field to the extension registry, with version detection, compatibility checking, and user-friendly upgrade guidance when incompatible versions are encountered.

### Core Implementation
- **`pkg/extensions/registry.go`** — Added `SchemaVersion` field to `Registry` struct with `CurrentRegistrySchemaVersion = "1.0"` constant
- **`pkg/extensions/registry_version.go`** *(new)* — Typed `ErrUnsupportedRegistrySchema` error, `CheckRegistrySchemaVersion()` using `Masterminds/semver` library for parsing
- **`pkg/extensions/json_source.go`** — Integrated version check into `newJsonSource()` central loading path (covers file + URL sources)
- **`pkg/extensions/manager.go`** — Surface `ErrUnsupportedRegistrySchema` explicitly instead of silently swallowing source load errors; wrap with `ErrorWithSuggestion` at UI boundary

### Validation & Error Handling
- **`pkg/extensions/validate_registry.go`** — Added `ValidateRegistry()` for registry-level field validation using `semver.NewVersion()` for consistent format acceptance
- **`resources/error_suggestions.yaml`** — Added fallback error suggestion rule for incompatible registry versions

### Schema & Data
- **`extensions/registry.schema.json`** — Added `schemaVersion` as optional string property
- **`extensions/registry.json`** — Added `"schemaVersion": "1.0"` at top level

### Tests & Documentation
- **`pkg/extensions/registry_version_test.go`** *(new)* — Table-driven test cases covering backward compatibility, forward compatibility, incompatible versions, malformed input, JSON serialization, and validation
- **`docs/extensions/extension-framework.md`** — Added Registry Schema Versioning section with compatibility matrix and upgrade guidance

## Version Compatibility Rules

| Registry Version | azd Behavior |
|---|---|
| Missing / empty | Treated as "1.0" (backward compatible) |
| 0.x (e.g., "0.1") | Accepted (pre-release schema versions) |
| Same major, any minor (e.g., "1.1") | Accepted silently (forward compatible) |
| Higher major (e.g., "2.0") | Hard error with upgrade guidance |
| Malformed | Parse error |

## Key Design Decisions

- **Semver-based parsing** using existing `Masterminds/semver` library — consistent with `version_compatibility.go`, `manager.go`, and other extension code. Accepts any valid semver format; only the major version matters for compatibility.
- **Typed error at pkg layer, UX wrapping at cmd boundary** — clean separation of concerns
- **Consistent validation** — both `ValidateRegistry()` (authoring) and `CheckRegistrySchemaVersion()` (runtime) use `semver.NewVersion()` for identical format acceptance
- **Critical bug fix** — source load errors in `manager.go` were previously logged and silently swallowed; schema incompatibility would have surfaced as "no extensions available" instead of a clear upgrade message

## Testing

- All tests passing
- Zero lint violations (golangci-lint)
- All existing extension tests pass (no regressions)

Resolves #6233

### DIFF
--- a/cli/azd/extensions/registry.schema.json
+++ b/cli/azd/extensions/registry.schema.json
@@ -268,8 +268,7 @@
     "properties": {
         "schemaVersion": {
             "type": "string",
-            "description": "Schema version for the registry format. Uses major.minor versioning.",
-            "pattern": "^\\d+\\.\\d+$"
+            "description": "Semantic version string for the registry format (e.g., '1.0', '1.0.0')."
         },
         "extensions": {
             "$comment": "Each extension must have a unique 'id' within the array.",
--- a/cli/azd/extensions/registry.schema.json
+++ b/cli/azd/extensions/registry.schema.json
@@ -266,6 +266,11 @@
         }
     },
     "properties": {
+        "schemaVersion": {
+            "type": "string",
+            "description": "Schema version for the registry format. Uses major.minor versioning.",
+            "pattern": "^\\d+\\.\\d+$"
+        },
         "extensions": {
             "$comment": "Each extension must have a unique 'id' within the array.",
             "type": "array",
--- a/cli/azd/pkg/extensions/manager.go
+++ b/cli/azd/pkg/extensions/manager.go
@@ -30,6 +30,7 @@ import (
 	"github.com/azure/azure-dev/cli/azd/internal/tracing/events"
 	"github.com/azure/azure-dev/cli/azd/internal/tracing/fields"
 	"github.com/azure/azure-dev/cli/azd/pkg/config"
+	"github.com/azure/azure-dev/cli/azd/pkg/errorhandler"
 	"github.com/azure/azure-dev/cli/azd/pkg/lazy"
 	"github.com/azure/azure-dev/cli/azd/pkg/osutil"
 	"github.com/azure/azure-dev/cli/azd/pkg/rzip"
@@ -758,6 +759,18 @@ func (tm *Manager) createSourcesFromConfig(
 
 		source, err := tm.sourceManager.CreateSource(ctx, config)
 		if err != nil {
+			if schemaErr, ok := errors.AsType[*ErrUnsupportedRegistrySchema](err); ok {
+				return nil, &errorhandler.ErrorWithSuggestion{
+					Err:     schemaErr,
+					Message: schemaErr.Error(),
+					Suggestion: "Upgrade azd to the latest version " +
+						"to use this registry",
+					Links: []errorhandler.ErrorLink{{
+						URL:   "https://aka.ms/azd/install",
+						Title: "Install/upgrade azd",
+					}},
+				}
+			}
 			log.Printf("failed to create source: %v", err)
 			continue
 		}
--- a/cli/azd/pkg/extensions/manager.go
+++ b/cli/azd/pkg/extensions/manager.go
@@ -751,6 +751,7 @@ func (tm *Manager) createSourcesFromConfig(
 	filter sourceFilterPredicate,
 ) ([]Source, error) {
 	sources := []Source{}
+	var schemaErrors []*ErrUnsupportedRegistrySchema
 
 	for _, config := range configs {
 		if filter != nil && !filter(config) {
@@ -760,22 +761,34 @@ func (tm *Manager) createSourcesFromConfig(
 		source, err := tm.sourceManager.CreateSource(ctx, config)
 		if err != nil {
 			if schemaErr, ok := errors.AsType[*ErrUnsupportedRegistrySchema](err); ok {
-				return nil, &errorhandler.ErrorWithSuggestion{
-					Err:     schemaErr,
-					Message: schemaErr.Error(),
-					Suggestion: "Upgrade azd to the latest version " +
-						"to use this registry",
-					Links: []errorhandler.ErrorLink{{
-						URL:   "https://aka.ms/azd/install",
-						Title: "Install/upgrade azd",
-					}},
-				}
+				log.Printf(
+					"WARNING: source '%s' has incompatible "+
+						"schema version %s, skipping",
+					config.Name, schemaErr.SchemaVersion,
+				)
+				schemaErrors = append(schemaErrors, schemaErr)
+				continue
 			}
 			log.Printf("failed to create source: %v", err)
 			continue
 		}
 
 		sources = append(sources, source)
+	}
+
+	// Only hard-fail when every source had an incompatible schema and
+	// no usable sources remain.
+	if len(sources) == 0 && len(schemaErrors) > 0 {
+		return nil, &errorhandler.ErrorWithSuggestion{
+			Err:     schemaErrors[0],
+			Message: schemaErrors[0].Error(),
+			Suggestion: "Upgrade azd to the latest version " +
+				"to use this registry",
+			Links: []errorhandler.ErrorLink{{
+				URL:   "https://aka.ms/azd/install",
+				Title: "Install/upgrade azd",
+			}},
+		}
 	}
 
 	return sources, nil
--- a/cli/azd/pkg/extensions/registry_version.go
+++ b/cli/azd/pkg/extensions/registry_version.go
@@ -5,8 +5,8 @@ package extensions
 
 import (
 	"fmt"
-	"strconv"
-	"strings"
+
+	"github.com/Masterminds/semver/v3"
 )
 
 // ErrUnsupportedRegistrySchema is returned when the registry schema version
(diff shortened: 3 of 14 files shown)


cli/azd/pkg/extensions/registry_version.go
```diff
@@ -0,0 +1,58 @@
...
+			schemaVersion, err,
+		)
+	}
+
+	if v.Major() > uint64(MinSupportedMajorVersion) {
+		return &ErrUnsupportedRegistrySchema{
+			SchemaVersion:       schemaVersion,
+			MaxSupportedVersion: CurrentRegistrySchemaVersion,
+		}
+	}
```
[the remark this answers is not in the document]
wbreza: Good catch — renamed to `MaxSupportedMajorVersion` with updated doc comment to clarify it's the upper bound for compatibility checking.

cli/azd/extensions/registry.schema.json
```diff
@@ -266,6 +266,10 @@
         }
     },
     "properties": {
+        "schemaVersion": {
+            "type": "string",
+            "description": "Semantic version string for the registry format (e.g., '1.0', '1.0.0')."
```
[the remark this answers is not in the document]
wbreza: Intentional — we standardized on `semver.NewVersion()` (Masterminds/semver) for both validation and runtime parsing. JSON Schema regex can't express the full range of formats the semver library accepts (`1.0`, `1.0.0`, `v1.0`), so the Go code is the authoritative validator. Adding a partial regex would create a false sense of validation coverage.

cli/azd/pkg/extensions/manager.go
```diff
@@ -758,6 +759,18 @@ func (tm *Manager) createSourcesFromConfig(
...
 		if err != nil {
+			if schemaErr, ok := errors.AsType[*ErrUnsupportedRegistrySchema](err); ok {
+				return nil, &errorhandler.ErrorWithSuggestion{
+					Err:     schemaErr,
+					Message: schemaErr.Error(),
+					Suggestion: "Upgrade azd to the latest version " +
+						"to use this registry",
+					Links: []errorhandler.ErrorLink{{
+						URL:   "https://aka.ms/azd/install",
+						Title: "Install/upgrade azd",
```
jongio: This returns immediately when one source has an incompatible schema, which prevents all other sources from loading. A user with the built-in azd registry + a third-party registry would lose all extension functionality if only the third-party source is incompatible.

Consider logging a warning and continuing instead, or only hard-failing when no compatible sources remain after the loop finishes.

cli/azd/resources/error_suggestions.yaml
```diff
@@ -576,6 +576,17 @@ rules:
...
+  # ============================================================================
+  # Extension Registry Errors
+  # ============================================================================
+
+  - errorType: "ErrUnsupportedRegistrySchema"
+    message: "The extension registry uses a schema version not supported by this version of azd."
+    suggestion: "Upgrade azd to the latest version to use this registry."
+    links:
+      - url: "https://aka.ms/azd/install"
+        title: "Install/upgrade azd"
```
jongio, line 591: This rule won't fire on the primary path because `createSourcesFromConfig` already wraps `ErrUnsupportedRegistrySchema` in `ErrorWithSuggestion`, and the error middleware skips the YAML pipeline for pre-wrapped errors (`error.go:173`). It could still serve as a catch-all if the error surfaces through other code paths, but worth noting the overlap.

jongio (commented):
Two things to address:

**Schema error breaks all extension sources** - `createSourcesFromConfig` now returns immediately when any single source has an incompatible schema version. If a user has multiple extension sources (the built-in azd registry + a third-party), one incompatible third-party source blocks all sources from loading, including the working built-in registry. The original log-and-continue pattern was deliberate for source-level resilience. Consider logging a visible warning for the incompatible source and continuing to load the rest, or only hard-failing when no compatible sources remain.

**`ValidateRegistry()` isn't wired into the validate command** - The PR adds `ValidateRegistry()` with schema-version checking, but `cmd/extension.go:1638` still calls `ValidateExtensions()`. Registry-level issues (missing/malformed `schemaVersion`) won't surface in `azd extension source validate`. Also, `displayValidationResult()` only iterates `result.Extensions` - it doesn't print `result.Issues`, so even switching the call wouldn't show schema warnings to users.

cli/azd/pkg/extensions/manager.go
wbreza: Great catch — fixed. `createSourcesFromConfig()` now logs a warning and continues to the next source when an individual source has an incompatible schema. Only hard-fails with upgrade guidance when zero compatible sources remain. Added two tests: partial schema failure (one bad, one good — good source still works) and all-schemas-incompatible (hard fail with error).

cli/azd/resources/error_suggestions.yaml
wbreza: Good observation — added a comment documenting the overlap. The YAML rule is an intentional catch-all fallback for any code paths where `ErrUnsupportedRegistrySchema` surfaces without the `ErrorWithSuggestion` wrapping from `manager.go`.

jongio (commented):
The source-resilience fix looks correct - incompatible sources are now skipped with a warning instead of blocking all sources, and the hard-fail only triggers when zero usable sources remain. Tests cover both scenarios well.

One item from my earlier review that's still open: `azd extension source validate` calls `ValidateExtensions()` (line 1638), not the new `ValidateRegistry()`. Schema version issues won't surface during authoring-time validation. Not blocking since the runtime path in `json_source.go` handles schema checks correctly, but worth a follow-up if authoring-time validation is a goal.

wbreza:
Good catch — fixed. The validate command now calls `ValidateRegistry()` (which validates both registry-level fields like `schemaVersion` AND extension metadata) instead of only `ValidateExtensions()`.

Added a `RegistryProvider` optional interface using Go type assertion pattern — `registrySource` satisfies it automatically since it holds the full `*Registry`. Falls back to `ValidateExtensions()` for non-registry sources.

Display output now shows registry-level issues (errors/warnings) before per-extension issues.

jongio (approved):
Addresses my previous feedback. The RegistryProvider optional interface pattern is clean - type assertion in the validate command falls back gracefully to ValidateExtensions for non-registry sources, and ValidateRegistry handles nil safely. Tests cover the key paths. Looks good.

cli/azd/pkg/extensions/validate_registry.go
```diff
@@ -77,12 +77,69 @@ type ExtensionValidationResult struct {
...
+	result = ValidateExtensions(registry.Extensions, strict)
+	validateSchemaVersion(result, registry.SchemaVersion, strict)
+	return result
+}
+
+// validateSchemaVersion validates the registry-level schemaVersion field.
+func validateSchemaVersion(
+	result *RegistryValidationResult,
+	schemaVersion string,
+	strict bool,
```
JeffreyCA, line 125: `strict` appears to be unused - were you planning to use it in some way?

cli/azd/pkg/extensions/validate_registry.go
```diff
@@ -77,12 +77,69 @@ type ExtensionValidationResult struct {
...
+		return
+	}
+
+	if _, err := semver.NewVersion(schemaVersion); err != nil {
+		result.addError(fmt.Sprintf(
+			"invalid schemaVersion format %q: %v",
+			schemaVersion, err,
+		))
+	}
+}
```
JeffreyCA, line 141: `validateSchemaVersion` only checks whether schemaVersion parses as semver, but it does not reject versions whose major version exceeds `MaxSupportedMajorVersion`. That seems to conflict with `CheckRegistrySchemaVersion`, which rejects unsupported majors and can create an inconsistent state where a registry with schemaVersion: "2.5" can pass `ValidateRegistry()`, but later fail when loaded through the JSON source path.

vhvb1989 (approved):

wbreza:
Addressed both review comments in commit 25a6ad0b4:

**strict parameter (line 125):** Now functional — in strict mode, missing schemaVersion is an error (not a warning), useful for registry authors validating before publishing.

**Major version check (line 141):** alidateSchemaVersion now checks .Major() > MaxSupportedMajorVersion after parsing, consistent with CheckRegistrySchemaVersion. Uses error in strict mode, warning in non-strict. Added 3 tests covering both modes.

PR #7843, opened 2026-04-21
Handle existing AI project selection in model configuration

### LINKED ISSUE #7842: A resource with this name already exists or is in a conflicting state
Using existing resources has unexpected results. 

When using --project-id, it might be a project someone else setup for you, either way, its not expected for AZD provision to be doing anything with it. Not assigning roles, etc. 

It can lead to issues like this: 

```
Creating/Updating resources
  You can view detailed progress in the Azure Portal:
  ...

ERROR: A resource with this name already exists or is in a conflicting state.
Suggestion: Check for existing or soft-deleted resources in the Azure portal.
error executing step command 'provision': deployment failed: error deploying infrastructure: deploying to subscription: 
Deployment Error Details:
UserError: Connection acr-connection already exist, and can only be updated by the workspace that created it, which is the workspace with workspaceId: /subscriptions/<sub>/resourceGroups/<rg>/providers/Microsoft.MachineLearningServices/workspaces/<project>@<workspace>@AML
 ```

And because the bicep has hard coded names in the defaults, it can lead to issues like this:

```
ERROR: error executing step command 'provision': deployment failed: error deploying infrastructure: validating deployment to subscription: 

Validation Error Details:
DeploymentActive: The deployment with resource id '/subscriptions/<sub>/resourceGroups/<rg>/providers/Microsoft.Resources/deployments/ai-project' cannot be saved, because this would overwrite an existing deployment which is still active. If this is an intentional re-deployment, please wait for the original deployment to complete, or cancel it before re-attempting the operation. The previous deployment was started at '4/21/2026 6:28:51 PM' with correlationId '2a32bbad0cd58b4ee6534bf995929083', and will expire at '4/28/2026 6:28:51 PM' if it does not complete before then. Please see https://aka.ms/arm-deploy-resources for usage details.
```

### PR DESCRIPTION
**Pull request by therealjohn**
This pull request updates the logic for handling existing Azure AI projects during the initialization process. Now, when an existing project is selected, an environment variable is set to signal downstream tools to skip unnecessary provisioning steps.

Project selection and provisioning logic:

* Updated the conditional in `configureModelChoice` (in `init.go`) to check for an existing project and, if found, set the `USE_EXISTING_AI_PROJECT` environment variable to `"true"`, signaling Bicep to skip project/role/connection provisioning for existing projects.

matching PR for bicep: https://github.com/Azure-Samples/azd-ai-starter-basic/pull/56

Fixes #7842

### DIFF
--- a/cli/azd/extensions/azure.ai.agents/internal/cmd/init.go
+++ b/cli/azd/extensions/azure.ai.agents/internal/cmd/init.go
@@ -805,6 +805,13 @@ func (a *InitAction) configureModelChoice(
 			if selectedProject == nil {
 				return nil, fmt.Errorf("foundry project not found: %s", a.flags.projectResourceId)
 			}
+
+			// Signal Bicep to skip project/role/connection provisioning for this existing project
+			if err := setEnvValue(
+				ctx, a.azdClient, a.environment.Name, "USE_EXISTING_AI_PROJECT", "true",
+			); err != nil {
+				return nil, err
+			}
 		} else {
 			// Prompt user to pick an existing Foundry project or create new resources
 			projectChoices := []*azdext.SelectChoice{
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Review by jongio — COMMENTED** [review]
Issues to address:
- init.go:924 - `USE_EXISTING_AI_PROJECT` is never set to `false` on the alternate paths, so re-init on the same environment can leave stale state that makes Bicep skip provisioning incorrectly

Question: the two other `selectFoundryProject` call sites (lines 805, 846) also result in existing projects being used but don't set this env var - intentional scoping for the model-resources path only?

File: cli/azd/extensions/azure.ai.agents/internal/cmd/init.go [hunk 1]
```diff
@@ -916,7 +916,12 @@ func (a *InitAction) configureModelChoice(
 			return nil, err
 		}
 
-		if selectedProject == nil {
+		if selectedProject != nil {
+			// Signal Bicep to skip project/role/connection provisioning for this existing project
+			if err := setEnvValue(ctx, a.azdClient, a.environment.Name, "USE_EXISTING_AI_PROJECT", "true"); err != nil {
+				return nil, err
+			}
+		} else {
```
jongio, line 952: **[MEDIUM]** `USE_EXISTING_AI_PROJECT` is only set to `true` but never cleared. If a user re-runs init on the same env and picks `Create new` after previously picking `existing`, the stale `true` persists. This `else` branch and the `case "new"` path (~line 934) should set it to `false` so Bicep doesn't incorrectly skip provisioning.

trangevi (approved):

JeffreyCA (approved):

PR #7893, opened 2026-04-23
Merge foundry-training-dev into main

saanikaguptamicrosoft:


.github/CODEOWNERS
```diff
@@ -14,6 +14,7 @@
 /cli/azd/extensions/azure.ai.agents/        @jongio @wbreza @vhvb1989 @hemarina @weikanglim @JeffreyCA @tg-msft @rajeshkamal5050 @trangevi @trrwilson @therealjohn
 /cli/azd/extensions/azure.ai.finetune/      @jongio @wbreza @vhvb1989 @hemarina @weikanglim @JeffreyCA @tg-msft @rajeshkamal5050 @trangevi @achauhan-scc @kingernupur @rabollin @saanikaguptamicrosoft
 /cli/azd/extensions/azure.ai.models/        @jongio @wbreza @vhvb1989 @hemarina @weikanglim @JeffreyCA @tg-msft @rajeshkamal5050 @trangevi @achauhan-scc @kingernupur @rabollin @saanikaguptamicrosoft
+/cli/azd/extensions/azure.ai.customtraining/  @wbreza @vhvb1989 @hemarina @weikanglim @JeffreyCA @tg-msft @rajeshkamal5050 @trangevi @achauhan-scc @kingernupur @rabollin @saanikaguptamicrosoft
```
JeffreyCA, line 17:
```suggestion
/cli/azd/extensions/azure.ai.customtraining/  @jongio @wbreza @vhvb1989 @hemarina @weikanglim @JeffreyCA @tg-msft @rajeshkamal5050 @trangevi @achauhan-scc @kingernupur @rabollin @saanikaguptamicrosoft
```

(for consistency)

cli/azd/extensions/azure.ai.customtraining/internal/cmd/root.go
```diff
@@ -0,0 +1,49 @@
...
+	Debug    bool
+	NoPrompt bool
+}
+
+var rootFlags rootFlagsDefinition
+
+func NewRootCommand() *cobra.Command {
+	rootCmd := &cobra.Command{
+		Use:           "training <command> [options]",
+		Short:         "Extension for Azure AI Foundry custom training jobs. (Preview)",
```
JeffreyCA, line 20: Should this be "Microsoft Foundry"?

cli/azd/extensions/azure.ai.customtraining/extension.yaml
```diff
@@ -0,0 +1,20 @@
+id: azure.ai.customtraining
+namespace: ai.training
+displayName: Azure AI Custom Training (Preview)
+description: Extension for Azure AI Foundry custom training jobs. (Preview)
```
JeffreyCA, line 4: Should this be Microsoft Foundry?

cli/azd/extensions/azure.ai.customtraining/main.go
```diff
@@ -0,0 +1,28 @@
...
+}
+
+func main() {
+	ctx := context.Background()
+	rootCmd := cmd.NewRootCommand()
+
+	if err := rootCmd.ExecuteContext(ctx); err != nil {
+		color.Red("Error: %v", err)
+		os.Exit(1)
+	}
```
JeffreyCA, line 27: Can we use the new `azdext.Run(cmd.NewRootCommand())` pattern? See `azure.ai.agents` as an example

cli/azd/extensions/azure.ai.customtraining/ci-build.ps1
JeffreyCA, line 1: I think we're missing a couple of workflow files, like:

- .github/workflows/lint-ext-azure-ai-agents.yml
- eng/pipelines/release-ext-azure-ai-agents.yml

and also:

- CHANGELOG.md
- README.md
- cspell.yaml

cli/azd/extensions/azure.ai.customtraining/internal/cmd/init.go
```diff
@@ -0,0 +1,380 @@
...
+	subResponse, err := azdClient.Prompt().PromptSubscription(ctx, &azdext.PromptSubscriptionRequest{})
+	if err != nil {
+		return nil, fmt.Errorf("failed to prompt for subscription: %w", err)
+	}
+
+	flags.subscriptionId = subResponse.Subscription.Id
+
+	return &azdext.AzureContext{
+		Scope: &azdext.AzureScope{
+			TenantId:       subResponse.Subscription.TenantId,
```
JeffreyCA, line 300: **Multi-tenant auth bug.** This stores `Subscription.TenantId` (the resource tenant) in `AzureContext.Scope.TenantId`, which then flows into the credential at line 66 and is persisted to the environment at lines 103 and 162. For guest / multi-tenant users the resource tenant differs from the user's access tenant, so credential acquisition will fail and every downstream `job` command inherits the bad value.

Use `Subscription.UserTenantId` here, matching the pattern in `azure.ai.finetune/internal/cmd/init.go` and `azure.ai.models/internal/cmd/init.go`. The non-interactive branch (which goes through `LookupTenant`) is already correct.

cli/azd/extensions/azure.ai.customtraining/pkg/client/client.go
```diff
@@ -0,0 +1,157 @@
...
+}
+
+// doDataPlaneWithVersion executes an authenticated HTTP request with a specific API version.
+func (c *Client) doDataPlaneWithVersion(ctx context.Context, method, path, apiVersion string, body interface{}, queryParams ...string) (*http.Response, error) {
+	reqURL := fmt.Sprintf("%s%s/%s?api-version=%s", c.baseURL, c.subPath, path, apiVersion)
+	for i := 0; i+1 < len(queryParams); i += 2 {
+		reqURL += fmt.Sprintf("&%s=%s", queryParams[i], url.QueryEscape(queryParams[i+1]))
+	}
+
+	fmt.Printf("[DEBUG] %s %s\n", method, reqURL)
```
JeffreyCA, line 91: **Unconditional debug print on every request.** This fires on every API call regardless of `--debug` / `c.debugBody` and writes to **stdout**, which:

1. Corrupts `--output json` (consumers can't parse the output).
2. Leaks request URLs (incl. continuation tokens from list pagination) into normal user output.

Gate this behind `c.debugBody` (or a dedicated verbose flag) and write to `os.Stderr`, matching how the request-body logging at line 99–101 is gated.

cli/azd/extensions/azure.ai.customtraining/pkg/models/dataset.go
```diff
@@ -0,0 +1,85 @@
...
+//	  "dataUri": "https://storage.blob.core.windows.net/...",
+//	  "dataType": "uri_folder",
+//	  "tags": { "contentHash": "a1b2c3d4...full 64 char hash..." }
+//	}
+type DatasetVersion struct {
+	ID          string            `json:"id,omitempty"`          // Full ARM resource ID (set by API, not by us)
+	Name        string            `json:"name,omitempty"`        // Dataset name, e.g. "code-myproject"
+	Version     string            `json:"version,omitempty"`     // Version string, e.g. truncated hash "a1b2c3..."
+	DataURI     string            `json:"dataUri,omitempty"`     // Permanent blob URI where files are stored
+	DataType    string            `json:"type,omitempty"`        // Always "uri_folder" (a directory of files)
```
JeffreyCA, line 82: **JSON tag mismatch.** The field is `DataType` and every doc comment / example in this file (lines 55, 63, 74) says the API expects `"dataType"`, but the tag is `json:"type,omitempty"`. As written, the dataset PATCH/GET will marshal/unmarshal `"type":"uri_folder"` instead of `"dataType":"uri_folder"` — sent from `internal/service/upload_service.go:158, 203`.

Please fix the tag (or the docs/usage if `"type"` is actually correct) and add a small round-trip test to lock the wire format.

cli/azd/extensions/azure.ai.customtraining/pkg/client/jobs.go
```diff
@@ -0,0 +1,130 @@
...
+		return nil, fmt.Errorf("failed to decode response: %w", err)
+	}
+
+	return &result, nil
+}
+
+// GetJob retrieves a specific job by ID.
+// GET .../jobs/{id}
+func (c *Client) GetJob(ctx context.Context, id string) (*models.JobResource, error) {
+	resp, err := c.doDataPlane(ctx, http.MethodGet, fmt.Sprintf("jobs/%s", id), nil)
```
JeffreyCA, line 61: **Unescaped path parameter.** `id` is interpolated directly into the URL path here and at lines 82, 103, and 119. A name containing `/`, `?`, `..`, or whitespace will alter the request path (potentially hitting an unintended endpoint).

Wrap with `url.PathEscape(id)` — the same fix is needed in:
- `pkg/client/history.go:18` (`runID`)
- `pkg/client/metrics.go:23, 58` (`jobID`)
- `pkg/client/artifacts.go:21, 50, 99, 143, 178` (`jobID` is unescaped even though the sibling `artifactPath` is escaped)

`pkg/client/datasets.go` already does this correctly.

cli/azd/extensions/azure.ai.customtraining/internal/utils/yaml_parser.go
```diff
@@ -0,0 +1,116 @@
...
+	if err := ValidateJobDefinition(&job); err != nil {
+		return nil, err
+	}
+
+	return &job, nil
+}
+
+// ResolveRelativePaths converts relative paths in the job definition to absolute paths
+// based on the YAML file's directory. Remote URIs (azureml://, https://, http://) are left as-is.
+func (j *JobDefinition) ResolveRelativePaths(yamlDir string) {
```
JeffreyCA, line 83: **Path traversal.** `filepath.Join(yamlDir, path)` does not reject `..` segments. A YAML such as

```yaml
code: ../../etc/passwd
inputs:
  data:
    path: ../../../.ssh
```

resolves to paths outside the project tree, which then get hashed and uploaded by the upload service.

After resolving, validate the result is still rooted under `yamlDir` (e.g. `filepath.Rel`, then reject results starting with `..`). The same guard should apply anywhere `hash.go` walks user-supplied roots.

cli/azd/extensions/azure.ai.customtraining/internal/cmd/job_delete.go
```diff
@@ -0,0 +1,96 @@
...
+		RunE: func(cmd *cobra.Command, args []string) error {
+			ctx := azdext.WithAccessToken(cmd.Context())
+
+			if name == "" {
+				return fmt.Errorf("--name is required: provide the job name/ID to delete")
+			}
+
+			// Prompt for confirmation unless --yes is specified
+			if !yes {
+				fmt.Printf("? Are you sure you want to delete job '%s'? [y/N] ", name)
```
JeffreyCA, line 37: **Bypasses `--no-prompt`.** The root command exposes a global `--no-prompt` flag (see `internal/cmd/root.go:36`), but this confirmation only honors a local `--yes`. In a non-TTY / CI host invoked with `--no-prompt`, the `bufio.NewReader(os.Stdin).ReadString('\n')` call will block on stdin or fail confusingly.

Use the azd console primitives (the same ones `init.go` already uses via `input.NewConsole`) and respect `rootFlags.NoPrompt` so behavior is consistent across the extension.

cli/azd/extensions/azure.ai.customtraining/pkg/client/client.go
```diff
@@ -0,0 +1,157 @@
...
+
+	baseURL := fmt.Sprintf("%s://%s", parsedURL.Scheme, parsedURL.Host)
+	subPath := "/" + strings.Join(pathParts[:3], "/")
+
+	return &Client{
+		baseURL:    baseURL,
+		subPath:    subPath,
+		apiVersion: DefaultAPIVersion,
+		credential: credential,
+		httpClient: &http.Client{Timeout: 30 * time.Second},
```
JeffreyCA, line 80: **No retry policy.** The Foundry data plane routinely returns 429 / 5xx; a flat 30s timeout with no retry will be flaky in normal operation. Consider plumbing requests through an `azcore` runtime pipeline with a retry policy, or at minimum adding a small exponential-backoff helper for 429 / 502 / 503 / 504 / network errors.

cli/azd/extensions/azure.ai.customtraining/pkg/client/client.go
```diff
@@ -0,0 +1,157 @@
...
+		return fmt.Errorf("failed to get access token: %w", err)
+	}
+
+	req.Header.Set("Authorization", "Bearer "+token.Token)
+	return nil
+}
+
+// HandleError reads the error body and returns a formatted error.
+func (c *Client) HandleError(resp *http.Response) error {
+	body, _ := io.ReadAll(resp.Body)
```
JeffreyCA, line 143: **Unbounded error body read.** `io.ReadAll(resp.Body)` has no size limit; a misbehaving server can return an arbitrarily large error body. Wrap with `io.LimitReader(resp.Body, 64*1024)` (or similar) before reading.

cli/azd/extensions/azure.ai.customtraining/go.mod
```diff
@@ -0,0 +1,91 @@
+module azure.ai.customtraining
+
+go 1.25
```
JeffreyCA, line 3: **Go directive and azd dependency.** - The repo standard is `go 1.26.1` (see peer extensions: `azure.ai.agents/go.mod`, `azure.ai.finetune/go.mod`, `azure.ai.models/go.mod`). This module pins `go 1.25` — please bump to match so we can use the modern stdlib patterns the project encourages.
- Line 9 pins `github.com/azure/azure-dev/cli/azd` to a pseudo-version (`v0.0.0-…`). Peer extensions use a published semver tag (`v1.23.13` / `v1.23.14`). Please switch to a release tag (or coordinate the merge with the next azd release if you genuinely need unreleased APIs).

cli/azd/extensions/azure.ai.customtraining/extension.yaml
```diff
@@ -0,0 +1,20 @@
+id: azure.ai.customtraining
```
JeffreyCA, line 1: **Missing manifest fields.** Compared with `azure.ai.agents/extension.yaml`, this file is missing:

1. The schema header `# yaml-language-server: $schema=../extension.schema.json`
2. A `requiredAzdVersion: ">…"` constraint — without it, users on older azd versions can install and silently break.

Please add both.

cli/azd/extensions/azure.ai.customtraining/internal/azcopy/runner.go
```diff
@@ -0,0 +1,322 @@
...
+				}
+			}
+		case "Error":
+			fmt.Fprintf(os.Stderr, "\n  azcopy error: %s\n", msg.MessageContent)
+		case "EndOfJob":
+			fmt.Fprintln(os.Stdout)
+		}
+	}
+
+	if err := cmd.Wait(); err != nil {
```
JeffreyCA, line 181: **Unchecked `scanner.Err()`.** After the `scanner.Scan()` loop, `scanner.Err()` is never checked. If the azcopy stdout pipe is truncated or errors mid-stream, the loop ends silently and `cmd.Wait()` may still report success — leading to spurious "upload succeeded" outcomes. Please check `scanner.Err()` and propagate.

JeffreyCA (commented):
Thanks for putting this together — the structure (cmd / service / client / models split, YAML-driven submission, content-addressed uploads, offline `job validate`) is sound and aligns well with the other Foundry extensions.

Inline comments below are grouped roughly:

**High** (correctness / API compatibility / safety):
1. `internal/cmd/init.go` — interactive subscription prompt stores the wrong tenant.
2. `pkg/client/client.go` — unconditional `[DEBUG]` print on every request.
3. `pkg/models/dataset.go` — `DataType` JSON tag is `"type"` but the API expects `"dataType"`.
4. `pkg/client/{jobs,history,metrics,artifacts}.go` — path params interpolated without `url.PathEscape`.
5. `internal/utils/yaml_parser.go` — relative paths can escape `yamlDir` via `..`.

**Should fix** before merge:
6. `internal/cmd/job_delete.go` — confirmation bypasses `--no-prompt` and uses raw stdin.
7. `pkg/client/client.go` — no retry/backoff and unbounded error-body read.
8. `go.mod` — bump Go directive to `1.26.1` and replace the pseudo-version of azd core with a release tag.
9. `extension.yaml` — add `requiredAzdVersion` and the schema header.
10. `main.go` — use `azdext.Run` instead of raw cobra.
11. `internal/azcopy/runner.go` — unchecked `scanner.Err()`.

**Follow-ups** (non-blocking):
- Add `CHANGELOG.md`, `README.md`, and `cspell.yaml` to match peer extensions.
- Adopt modern Go patterns (`errors.AsType[T]`, `new(val)`, `slices.SortFunc`+`cmp`, etc.) per `cli/azd/AGENTS.md`.
- File a tracking work item to harden the auto-installed `azcopy` binary (SHA-256 verification against a pinned manifest, atomic rename) — best done as part of the planned shared-library consolidation across the AI extensions.

cli/azd/extensions/azure.ai.customtraining/pkg/models/dataset.go
```diff
@@ -0,0 +1,85 @@
...
+	Description string            `json:"description,omitempty"` // Human-readable description
```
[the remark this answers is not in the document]
JeffreyCA: Same finding — see https://github.com/Azure/azure-dev/pull/7893#discussion_r3133982642. Resolving as duplicate.

cli/azd/extensions/azure.ai.customtraining/pkg/client/client.go
```diff
@@ -0,0 +1,157 @@
...
+	var bodyReader io.Reader
+	if body != nil {
+		data, err := json.Marshal(body)
+		if err != nil {
+			return nil, fmt.Errorf("failed to marshal request body: %w", err)
+		}
+		if c.debugBody {
+			fmt.Printf("[DEBUG] Request body: %s\n", string(data))
+		}
+		bodyReader = bytes.NewReader(data)
```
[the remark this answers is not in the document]
JeffreyCA: Same finding — see https://github.com/Azure/azure-dev/pull/7893#discussion_r3133982508. Resolving as duplicate.

cli/azd/extensions/azure.ai.customtraining/pkg/client/jobs.go
```diff
@@ -0,0 +1,130 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to get job: %w", err)
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusOK {
+		return nil, c.HandleError(resp)
+	}
```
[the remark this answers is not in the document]
JeffreyCA: Covered by consolidated thread — see https://github.com/Azure/azure-dev/pull/7893#discussion_r3133982795. Resolving as duplicate.

cli/azd/extensions/azure.ai.customtraining/pkg/client/history.go
```diff
@@ -0,0 +1,38 @@
...
+)
+
+// GetRunHistory retrieves run history details for a specific job.
+// GET .../history/{runId}
+func (c *Client) GetRunHistory(ctx context.Context, runID string) (*models.RunHistory, error) {
+	resp, err := c.doDataPlane(ctx, http.MethodGet, fmt.Sprintf("history/%s", runID), nil)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get run history: %w", err)
+	}
+	defer resp.Body.Close()
```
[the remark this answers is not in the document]
JeffreyCA: Covered by consolidated path-escape thread — see https://github.com/Azure/azure-dev/pull/7893#discussion_r3133982795. Resolving as duplicate.

cli/azd/extensions/azure.ai.customtraining/pkg/client/artifacts.go
```diff
@@ -0,0 +1,200 @@
...
+		return nil, fmt.Errorf("failed to list artifacts: %w", err)
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode == http.StatusNotFound {
+		return nil, nil
+	}
+
+	if resp.StatusCode != http.StatusOK {
+		return nil, c.HandleError(resp)
```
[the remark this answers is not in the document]
JeffreyCA: Covered by consolidated path-escape thread — see https://github.com/Azure/azure-dev/pull/7893#discussion_r3133982795. Resolving as duplicate.

cli/azd/extensions/azure.ai.customtraining/pkg/client/metrics.go
```diff
@@ -0,0 +1,78 @@
...
+func (c *Client) ListMetrics(ctx context.Context, jobID string) (*models.MetricsListResponse, error) {
+	reqBody := &models.MetricsListRequest{
+		MetricNamespace:   nil,
+		ContinuationToken: nil,
+	}
+
+	resp, err := c.doDataPlane(ctx, http.MethodPost, fmt.Sprintf("metrics/%s/list", jobID), reqBody)
+	if err != nil {
+		return nil, fmt.Errorf("failed to list metrics: %w", err)
+	}
```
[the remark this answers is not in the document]
JeffreyCA: Covered by consolidated path-escape thread — see https://github.com/Azure/azure-dev/pull/7893#discussion_r3133982795. Resolving as duplicate.

cli/azd/extensions/azure.ai.customtraining/go.mod
```diff
@@ -0,0 +1,91 @@
...
+
+require (
+	github.com/Azure/azure-sdk-for-go/sdk/azcore v1.20.0
+	github.com/Azure/azure-sdk-for-go/sdk/azidentity v1.13.1
+	github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/cognitiveservices/armcognitiveservices v1.8.0
+	github.com/azure/azure-dev/cli/azd v0.0.0-20260122173819-89795b295491
```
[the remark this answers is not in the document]
JeffreyCA: Same finding — see https://github.com/Azure/azure-dev/pull/7893#discussion_r3133983451. Resolving as duplicate.

cli/azd/extensions/azure.ai.customtraining/internal/cmd/init.go
```diff
@@ -0,0 +1,380 @@
...
+			}
+
+			// Prompt for project endpoint
+			endpoint, err := console.Prompt(ctx, input.ConsoleOptions{
+				Message: "Enter the Azure AI Foundry project endpoint URL:",
+			})
+			if err != nil {
+				return fmt.Errorf("failed to prompt for endpoint: %w", err)
+			}
+			flags.projectEndpoint = endpoint
```
JeffreyCA, line 141: The extension framework provides ways to prompt for specific Azure resource interactively.

Consider using those instead of freeform input (see other extensions like azure.ai.agents as example)

cli/azd/extensions/azure.ai.customtraining/internal/azcopy/runner.go
```diff
@@ -0,0 +1,322 @@
...
+		"copy",
+		sourceArg,
+		sasURI,
+		"--recursive=true",
+		"--output-type", "json",
+		"--block-size-mb", "100",
+	}
+
+	//nolint:gosec // azcopyPath is resolved from known install directory
+	cmd := exec.CommandContext(ctx, r.azcopyPath, args...)
```
jongio, line 116: The SAS token is passed as a command-line argument here. On multi-user systems, `ps aux` (Linux/macOS) or Process Explorer (Windows) expose CLI arguments of running processes, making the SAS token visible to other local users.

azcopy v10.19+ supports `--sas-from-env` to read the SAS from an environment variable instead. Since the SAS is short-lived this isn't critical, but it's worth fixing - especially if this runs on shared CI agents or dev boxes.

jongio (commented):
New extension for Azure AI Foundry custom training jobs - ~9500 lines across 51 files. JeffreyCA's review covers the most critical issues (TenantId bug, path escaping, debug prints, missing infra). Adding a few things not yet covered.

**Test coverage**: 1 test file (`job_validator_test.go`) out of ~45 source files. The upload dedup flow in `upload_service.go` (hash-based versioning, zombie detection, sentinel tags, collision handling) is the trickiest logic in the extension and has zero tests. Same for the HTTP client error handling, resolver chain, YAML parsing edge cases, and all command handlers. This should be addressed before merge.

**File size**: `job_get.go` is 702 lines. The rich show output (run history, metrics, artifacts formatting) could be split into a dedicated output/formatter file.

**`doUpload` and `doUploadWithTag`** in `upload_service.go` are nearly identical. A single method with an optional `tags map[string]string` parameter would cut the duplication.

cli/azd/extensions/azure.ai.customtraining/internal/cmd/job_cancel.go
```diff
@@ -0,0 +1,78 @@
...
+			if name == "" {
+				return fmt.Errorf("--name is required: provide the job name/ID to cancel")
+			}
+
+			azdClient, err := azdext.NewAzdClient()
+			if err != nil {
+				return fmt.Errorf("failed to create azd client: %w", err)
+			}
+			defer azdClient.Close()
+
```
jongio, line 36: This 20-line block (create azdClient, get env values, check account/project, create credential, build endpoint, create API client) is repeated identically in `job_submit.go`, `job_delete.go`, `job_get.go`, and `job_list.go` - 5 copies total.

Extract a helper like `newAPIClientFromEnv(ctx) (*client.Client, error)` that does all of this in one call. It would also make it easier to apply JeffreyCA's feedback (like the TenantId fix) in a single place instead of 5.

cli/azd/extensions/azure.ai.customtraining/internal/utils/job_validator_test.go
```diff
@@ -0,0 +1,349 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
jongio, line 1: This is the only test file in the entire extension. Missing test coverage for the critical paths:

- `upload_service.go`: dedup flow (hash match, zombie detection, collision, sentinel tag verification)
- `client.go`: error handling, auth token flow, status code handling
- `yaml_parser.go`: malformed YAML, missing fields, path resolution edge cases
- `resolver.go`: ARM ID detection, code/input/compute resolution
- `runner.go`: azcopy output parsing, progress reporting, error cases

The dedup logic in particular has several interacting state transitions (existing version with/without tag, matching/mismatching hash, zombie cleanup) that are hard to get right without test coverage.

cli/azd/extensions/azure.ai.customtraining/internal/cmd/job_cancel.go
saanikaguptamicrosoft: Logic will get scattered. Deferring for now.

cli/azd/extensions/azure.ai.customtraining/internal/azcopy/runner.go
saanikaguptamicrosoft: Deferring for now.

cli/azd/extensions/azure.ai.customtraining/internal/utils/yaml_parser.go
saanikaguptamicrosoft: AML SDK allows .. paths. So keeping it to maintain parity.

PR #7906, opened 2026-04-24
fix: eliminate failing/flaky tests in devcontainer/non-CI environments

### LINKED ISSUE #7907: Flaky tests in devcontainer/non-CI environments (container env + update action)
Two test suites fail locally in devcontainers but pass in CI:

### `TestIsContainerEnvironment_EnvVars`
Ambient env vars (e.g. `REMOTE_CONTAINERS`) in devcontainers cause false positives due to non-deterministic Go map iteration. `ContainerRuntime()` may find the wrong env var first.

### `Test_UpdateAction_Run_SwitchChannel_CheckForUpdateError` / `Test_UpdateAction_Run_NoChannelNoConfigFlags`
Tests expect failure via `noopCommandRunner`, but `CheckForUpdate` uses an HTTP client. In CI, `IsRunningOnCI()` short-circuits before reaching `CheckForUpdate`, masking the bug. Locally, real HTTP calls succeed and `Run()` returns nil.

**Fix**: #7906

JeffreyCA:
## Problem

Two test suites fail when run locally in devcontainers but pass in CI:

1. **`TestIsContainerEnvironment_EnvVars`** — Ambient env vars like `REMOTE_CONTAINERS` (set in devcontainers) interfere with assertions because Go map iteration is non-deterministic. When testing `CODESPACES`, `ContainerRuntime()` may find `REMOTE_CONTAINERS` first and return `"devcontainer"` instead of `"codespaces"`.

2. **`Test_UpdateAction_Run_SwitchChannel_CheckForUpdateError`** / **`Test_UpdateAction_Run_NoChannelNoConfigFlags`** — These expect `Run()` to return an error via `noopCommandRunner`, but `CheckForUpdate` uses an HTTP client (not the command runner). In CI, `IsRunningOnCI()` short-circuits before reaching `CheckForUpdate`, masking the bug. Locally, the real HTTP call succeeds and `Run()` returns nil.

## Fix

- **Container env tests**: Clear all container env vars via `t.Setenv` before each subtest so only the target var is active.
- **Update action tests**: Add `httpClient` field to `updateAction` (nil = default) and inject a `failingTransport` in tests so `CheckForUpdate` fails deterministically without network access.

### DIFF
--- a/cli/azd/cmd/final_coverage3_test.go
+++ b/cli/azd/cmd/final_coverage3_test.go
@@ -7,6 +7,7 @@ import (
 	"bytes"
 	"context"
 	"errors"
+	"net/http"
 	"os"
 	"testing"
 
@@ -85,6 +86,18 @@ func (r *noopCommandRunner) ToolInPath(_ string) error {
 	return errors.New("not found")
 }
 
+// failingTransport is an http.RoundTripper that always returns an error.
+type failingTransport struct{}
+
+func (ft *failingTransport) RoundTrip(_ *http.Request) (*http.Response, error) {
+	return nil, errors.New("test: network disabled")
+}
+
+// failingHTTPClient returns an *http.Client that fails all requests.
+func failingHTTPClient() *http.Client {
+	return &http.Client{Transport: &failingTransport{}}
+}
+
 // ===========================================================================
 // updateAction.Run tests
 // ===========================================================================
@@ -131,6 +144,7 @@ func newTestUpdateAction(
 		writer:        writer,
 		configManager: cfgMgr,
 		commandRunner: cmdRunner,
+		httpClient:    failingHTTPClient(),
 	}
 }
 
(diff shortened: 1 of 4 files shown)


jongio (approved):
Clean fix for two test flakiness issues.

The map-to-slice conversion for `containerEnvVars` properly addresses the root cause - Go map iteration is non-deterministic, so when multiple container env vars are set (e.g. in a Codespace that's also a devcontainer), the old code could return different results between runs. The priority ordering in the slice is well-documented and sensible (CODESPACES > KUBERNETES > devcontainer).

The HTTP client injection in `updateAction` is the right fix for the second issue. `CheckForUpdate` uses HTTP (not the command runner), so `noopCommandRunner` alone wasn't enough to make tests deterministic outside CI. The nil-default in production preserves existing behavior.

Both test files now properly isolate env vars with `t.Setenv` before each subtest. Looks good.

Issue #7796, opened 2026-04-17
azure.ai.agents: adopt NewExtensionRootCommand and remove reserved global flag conflicts


JeffreyCA:
## Summary

Once `cli/azd/extensions/azure.ai.agents/go.mod` is bumped past the version that includes #7312 (currently pinned to `github.com/azure/azure-dev/cli/azd v1.23.14`), `azd ai agent ...` fails to start because the extension declares flags that collide with azd's reserved global flag registry:

```
ERROR: extension defines flags that conflict with reserved azd global flags:
  - command "":               --debug, --no-prompt
  - command "init":           --environment
  - command "show":           --output/-o
  - command "files list":     --output
  - command "files stat":     --output/-o
  - command "sessions create|list|show": --output/-o
Reserved flags: environment, cwd, debug, no-prompt, output, help, docs, trace-log-file, trace-log-url
```

## Root cause

The extension builds its root with a bare `cobra.Command{}` and manually wires `--debug` / `--no-prompt`, instead of using `azdext.NewExtensionRootCommand()` which already provides every SDK-managed global plus an `*ExtensionContext` (`OutputFormat`, `Environment`, `Debug`, `NoPrompt`, `Cwd`). Several subcommands then redundantly redeclare `--output`/`-o`, and `init` declares its own `--environment` (which the host pre-parser would hijack anyway).

## Proposed changes (extension only)

1. **Switch `internal/cmd/root.go` to `azdext.NewExtensionRootCommand()`**; drop the manual `--debug`/`--no-prompt` registrations and the `AZD_NO_PROMPT` fallback in `PersistentPreRunE` (the SDK handles both). Thread the returned `*ExtensionContext` to subcommands.
2. **Remove the per-subcommand `--output`/`-o` registrations** in `show.go:101`, `files.go:452` (`files list`), `files.go:754` (`files stat`), and `sessions*.go` (`create|list|show`). Read `extCtx.OutputFormat` instead. The accepted values (`json`, `table`) are already standard azd `Format` constants from `cli/azd/pkg/output/formatter.go`. Preserve current per-command default-format behavior by mapping the SDK default `"default"` → `"json"` per command, and consider rejecting unknown values explicitly (today they silently fall through to JSON).
3. **Rename or remove `init --environment`** (`init.go:518`) — either rename to `--env-name`/`--target-env`, or drop it and use `extCtx.Environment`.
[...]

PR #7919, opened 2026-04-24
Unify host gRPC error transport with a structured ActionableErrorDetail

### LINKED ISSUE #7844: Unify host gRPC actionable error transport
## Summary

The host-side gRPC error path in `cli/azd/internal/grpcserver/server.go` relies on `wrapErrorWithSuggestion()` to append user guidance into the status message string. Recent work improved two adjacent areas:

- PR #7797 adds structured auth classification for host → extension gRPC calls via `codes.Unauthenticated` + `google.rpc.ErrorInfo` in the `azd.auth` domain. azd-local conditions use `AUTH_*` reasons; Entra-originated failures preserve the raw `AADSTS<code>` reason.
- PR #7791 adds structured suggestion/link transport across the extension error boundary via `ExtensionError`.

The host gRPC error shaping path remains a separate, older mechanism that serializes actionable guidance primarily through message text.

## Problem

`wrapErrorWithSuggestion()` mixes auth classification, gRPC status code selection, and "preserve actionable guidance" by concatenating suggestion text into `status.Message()`. That works but is brittle: it drops `Links` and any future guidance metadata from `*internal.ErrorWithSuggestion`, and it keeps auth semantics and actionable UX guidance too tightly coupled in one transport path.

## Goal

Reshape host-side gRPC error handling into a single structured mapping path that:

1. preserves semantic classification, including auth subtype details
2. preserves actionable guidance (suggestion, links) as typed data, not message text
3. avoids relying on string concatenation as the transport contract

## Proposed approach

1. Extract the current `wrapErrorWithSuggestion()` logic from `server.go` into a dedicated host gRPC error-mapping helper (`internal/grpcserver/errors.go`) and rename it to something descriptive (e.g. `mapHostError`).
2. Keep the PR #7797 auth contract:
   - top-level code stays broad (`codes.Unauthenticated`)
   - auth subtype remains in structured `ErrorInfo` (`azd.auth`)
   - azd-local auth conditions use `AUTH_*`
   - Entra-originated failures preserve the raw `AADSTS<code>` reason
3. Introduce a structured gRPC detail (`ActionableErrorDetail { suggestion, links }`) so `*internal.ErrorWithSuggestion` can round-trip actionable guidance as typed data.
4. Teach `pkg/azdext` consumers (`ErrorSuggestion`, `ErrorMessage`, `ErrorLinks`) to read the new detail in addition to existing `LocalError` / `ServiceError` sources.
5. Stop appending suggestion text into `status.Message`. The user-facing message becomes `ErrorWithSuggestion.Message` (or `err.Error()`); guidance travels separately.
6. Collapse the duplicated `grpcStatusFromError` / `wrapErrorLinks` helpers that exist in both `internal/grpcserver` and `pkg/azdext` into single exported `azdext` versions consumed by both sides.

## Scope

- `cli/azd/internal/grpcserver/server.go` and a new `internal/grpcserver/errors.go`
- new `ActionableErrorDetail` proto in `cli/azd/grpc/proto/errors.proto`
- `pkg/azdext` consumer parsing and helper consolidation
- tests covering unary + streaming handlers

## Non-goals

- Do not expand PR #7791 beyond extension error transport.
- Do not expand PR #7797 beyond auth error classification.
- Do not introduce new host-side auth error types or a synthetic per-AAD-code taxonomy.
- Do not redesign unrelated extension error telemetry or UX flows in this issue.

## Acceptance criteria

- Host gRPC actionable errors are shaped through one structured mapping path in a dedicated helper file.
- Auth errors still preserve the `azd.auth` `ErrorInfo` contract from PR #7797 (`AUTH_*` for azd-local conditions; `AADSTS<code>` for Entra-originated failures).
- `*internal.ErrorWithSuggestion` round-trips suggestion **and** links as typed `ActionableErrorDetail` data, not via string parsing.
- `pkg/azdext` extractors (`ErrorSuggestion`, `ErrorMessage`, `ErrorLinks`) surface host-attached actionable detail.
- Duplicated gRPC status / link helpers are consolidated.
- Tests cover unary and streaming paths and assert that suggestion text no longer appears in `status.Message`.

## Out of scope (explicitly excluded after discussion)

- **Preserving the Go error chain across the gRPC boundary.** gRPC is a serialization boundary; only typed details (auth reason, actionable guidance) cross it. The wrapped chain on the host side is not reconstructible on the extension side and would not be useful if it were.
- **Adding `message` to `ActionableErrorDetail`.** That would create a fourth message channel duplicating `status.Message`. The user-facing message belongs in `status.Message`; the detail carries only what `status.Message` cannot.
- **String-concatenation compatibility fallback.** The previous shim protected no real out-of-tree consumer. Removing it lets the structured path be the single source of truth.

## Context

Follow-up to the transport/auth work split across:

- #7791
- #7797

JeffreyCA:
Fixes #7844

This PR replaces the ad-hoc `"<err>\n<suggestion>"` concatenation in the host's gRPC error path with a structured proto detail. Extensions read suggestion and links via the existing public extractors instead of parsing the error string.

This essentially means once extensions upgrade to latest `github.com/azure/azure-dev/cli/azd` in `go.mod`, extensions will be able to render the full error suggestion UX as core azd, e.g.:

image

### What changed

- **New proto detail:** `ActionableErrorDetail { suggestion, links }` in `errors.proto`. Host-attached only; extensions returning errors to the host still use `LocalError` / `ServiceError` via `ExtensionError`.
- **Host (`internal/grpcserver`):** new `errors.go` with `mapHostError` (renamed from `wrapErrorWithSuggestion`) and `statusMessage`. `status.Message` is now just `ErrorWithSuggestion.Message` (or `err.Error()`); the `\n<suggestion>` shim is gone. Suggestion and links travel as structured detail.
- **Extension (`pkg/azdext`):** exported `GRPCStatusFromError`, `ActionableErrorDetailFromStatus`, `ActionableErrorDetailFromError`, `WrapErrorLinks`, `UnwrapErrorLinks`. `WrapError` factored through a single-pass `populateExtensionErrorFromStatus`; the `extErr.Message == err.Error()` heuristic is gone. `ErrorSuggestion`, `ErrorMessage`, and `ErrorLinks` also extract from a host-attached `ActionableErrorDetail` so links survive the wire.
- **De-duplication:** removed parallel `grpcStatusFromError` and `wrapErrorLinks` helpers from `internal/grpcserver`; both packages share the exported `azdext` versions.

### Wire shape (after)

Host → extension errors carry:
- `status.Message` — user-facing message (no concatenation)
- `google.rpc.ErrorInfo { Domain="azd.auth", Reason=... }` — auth failures
- `azdext.ActionableErrorDetail { suggestion, links }` — actionable guidance

### Tests

- `Test_mapHostError` asserts the suggestion lands in `ActionableErrorDetail` and *not* in `status.Message`.
- `Test_PromptService_Prompt{Subscription,ResourceGroup}_ErrorWithSuggestion` exercise the full interceptor stack and read suggestion via `ActionableErrorDetailFromStatus`.
- `extension_error_test.go` updated for the new `ActionableErrorDetail` shape.

### DIFF
--- a/cli/azd/internal/grpcserver/errors.go
+++ b/cli/azd/internal/grpcserver/errors.go
@@ -6,6 +6,7 @@ package grpcserver
 import (
 	"errors"
 	"fmt"
+	"log"
 
 	"github.com/azure/azure-dev/cli/azd/internal"
 	"github.com/azure/azure-dev/cli/azd/pkg/auth"
@@ -98,6 +99,7 @@ func withAuthErrorInfo(st *status.Status, err error) *status.Status {
 		Domain: azdext.AuthErrorDomain,
 	})
 	if detailErr != nil {
+		log.Printf("failed to attach auth ErrorInfo to gRPC status: %v", detailErr)
 		return st
 	}
 
@@ -114,6 +116,7 @@ func withActionableErrorDetail(st *status.Status, err *internal.ErrorWithSuggest
 		Links:      azdext.WrapErrorLinks(err.Links),
 	})
 	if detailErr != nil {
+		log.Printf("failed to attach ActionableErrorDetail to gRPC status: %v", detailErr)
 		return st
 	}
 
--- a/cli/azd/internal/grpcserver/errors.go
+++ b/cli/azd/internal/grpcserver/errors.go
@@ -1,0 +1,141 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package grpcserver
+
+import (
+	"errors"
+	"fmt"
+
+	"github.com/azure/azure-dev/cli/azd/internal"
+	"github.com/azure/azure-dev/cli/azd/pkg/auth"
+	"github.com/azure/azure-dev/cli/azd/pkg/azdext"
+	"google.golang.org/genproto/googleapis/rpc/errdetails"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+// mapHostError serializes a host-originated Go error into a gRPC status error for transport
+// to extensions.
+//
+// This is the gRPC serialization boundary: the original error chain (e.g. *auth.AuthFailedError)
+// is intentionally not preserved across the wire — only the structured details defined here.
+//
+// Auth-related errors are reported with codes.Unauthenticated and an azd.auth ErrorInfo detail
+// (preserving AADSTS<code> reasons from Entra). ErrorWithSuggestion errors carry an
+// ActionableErrorDetail so consumers receive suggestion + links structurally.
+//
+// status.Message is always err.Error() (or ErrorWithSuggestion.Message when set). Suggestion
+// text is never concatenated into status.Message; consumers must read ActionableErrorDetail
+// for remediation guidance.
+func mapHostError(err error) error {
+	if err == nil {
+		return nil
+	}
+
+	suggestionErr, hasSuggestion := errors.AsType[*internal.ErrorWithSuggestion](err)
+	isAuthErr := isAuthError(err)
+	if !hasSuggestion && !isAuthErr {
+		return err
+	}
+
+	code := codes.Unknown
+	if st, ok := azdext.GRPCStatusFromError(err); ok {
+		code = st.Code()
+	}
+	if isAuthErr {
+		code = codes.Unauthenticated
+	}
+
+	st := status.New(code, statusMessage(err, suggestionErr))
+	if isAuthErr {
+		st = withAuthErrorInfo(st, err)
+	}
+	if hasSuggestion {
+		st = withActionableErrorDetail(st, suggestionErr)
+	}
+
+	return st.Err()
+}
+
+// statusMessage returns the user-facing message that should populate status.Message.
+// When the source is an ErrorWithSuggestion with an explicit Message, that wins. Otherwise,
+// if err is already a gRPC status error, use its Message (avoids nesting "rpc error: ..."
+// prefixes in the new status). Falls back to err.Error(). Suggestion text is never appended.
+func statusMessage(err error, suggestionErr *internal.ErrorWithSuggestion) string {
+	if suggestionErr != nil && suggestionErr.Message != "" {
+		return suggestionErr.Message
+	}
+	if st, ok := azdext.GRPCStatusFromError(err); ok {
+		return st.Message()
+	}
+	return err.Error()
+}
+
+// isAuthError reports whether err's chain contains a known auth-failure type that should be
+// surfaced over gRPC as codes.Unauthenticated.
+func isAuthError(err error) bool {
+	if errors.Is(err, auth.ErrNoCurrentUser) {
+		return true
+	}
+	if _, ok := errors.AsType[*auth.ReLoginRequiredError](err); ok {
+		return true
+	}
+	if _, ok := errors.AsType[*auth.AuthFailedError](err); ok {
+		return true
+	}
+	return false
+}
+
+func withAuthErrorInfo(st *status.Status, err error) *status.Status {
+	reason := grpcAuthReason(err)
+	if reason == "" {
+		return st
+	}
+
+	withDetails, detailErr := st.WithDetails(&errdetails.ErrorInfo{
+		Reason: reason,
+		Domain: azdext.AuthErrorDomain,
+	})
+	if detailErr != nil {
+		return st
+	}
+
+	return withDetails
+}
+
+func withActionableErrorDetail(st *status.Status, err *internal.ErrorWithSuggestion) *status.Status {
+	if err.Suggestion == "" && len(err.Links) == 0 {
+		return st
+	}
+
+	withDetails, detailErr := st.WithDetails(&azdext.ActionableErrorDetail{
+		Suggestion: err.Suggestion,
+		Links:      azdext.WrapErrorLinks(err.Links),
+	})
+	if detailErr != nil {
+		return st
+	}
+
+	return withDetails
+}
+
+func grpcAuthReason(err error) string {
+	if errors.Is(err, auth.ErrNoCurrentUser) {
+		return azdext.AuthErrorReasonNotLoggedIn
+	}
+
+	// Pass through the originating AAD error code (e.g., "AADSTS530084") when available.
+	// This preserves Entra's own semantics rather than redefining them on azd's side.
+	if authFailed, ok := errors.AsType[*auth.AuthFailedError](err); ok {
+		if authFailed.Parsed != nil && len(authFailed.Parsed.ErrorCodes) > 0 {
+			return fmt.Sprintf("AADSTS%d", authFailed.Parsed.ErrorCodes[0])
+		}
+	}
+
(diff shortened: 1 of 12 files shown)


jongio (commented):
Clean approach. The structured `ActionableErrorDetail` properly separates transport from presentation, and the extraction helpers (`ActionableErrorDetailFromStatus`, `GRPCStatusFromError`) give extensions typed access to remediation data without string parsing.

One follow-up to track: `azure.ai.agents` at `internal/exterrors/errors.go` uses `authFromGrpcMessage(st.Message())` for string-based auth classification. After this change, `st.Message()` won't include the concatenated suggestion text. The core auth keywords ("not logged in", "expired") come from the error text itself, so classification should still work for common cases. But the extension will need a follow-up to read `ActionableErrorDetail` for the full structured error UX once it upgrades its `go.mod`. Worth tracking as a task.

No CHANGELOG entry yet under 1.25.0-beta.1 for this change. Since `status.Message` no longer carries suggestion text, this is a behavioral change in the host-to-extension error wire format that downstream consumers should know about.

cli/azd/internal/grpcserver/errors.go
```diff
@@ -0,0 +1,141 @@
...
+// is intentionally not preserved across the wire — only the structured details defined here.
+//
+// Auth-related errors are reported with codes.Unauthenticated and an azd.auth ErrorInfo detail
+// (preserving AADSTS<code> reasons from Entra). ErrorWithSuggestion errors carry an
+// ActionableErrorDetail so consumers receive suggestion + links structurally.
+//
+// status.Message is always err.Error() (or ErrorWithSuggestion.Message when set). Suggestion
+// text is never concatenated into status.Message; consumers must read ActionableErrorDetail
+// for remediation guidance.
+func mapHostError(err error) error {
```
jongio, line 32: Once `mapHostError` stops concatenating suggestion text into `status.Message`, extensions built against the pre-`ActionableErrorDetail` SDK lose that UX during version skew. Old `WrapError` only special-cases auth gRPC status, and old `ErrorSuggestion`/`ErrorLinks` don't read status details - so new host + old extension falls back to the plain message with no actionable suggestion.

Is mixed-version host/extension a supported state? If so, should there be a temporary fallback (or coordinated in-tree extension bumps) to avoid the gap? The in-tree extensions are currently pinned to older azd module versions, so this isn't just theoretical.

jongio (commented):
One item to discuss - backward compat during version skew between host and extensions. See inline.

wbreza (commented):
﻿## Code Review — Supplementary Findings

> These findings supplement the existing reviews from @jongio (version skew, CHANGELOG) and Copilot (statusMessage nesting, ErrorMessage doc). No duplicates.

### 🟡 Medium

**1. Silent error swallowing on `WithDetails()` failure**
`cli/azd/internal/grpcserver/errors.go` — `withActionableErrorDetail` and `withAuthErrorInfo`

When `st.WithDetails()` fails, both functions silently return the original status without the structured detail attached. Suggestion/links are lost with no trace. While unlikely (proto marshal of well-known types rarely fails), a `log.Warn` when `detailErr != nil` is trivial and follows defensive-coding norms.

**2. Links not asserted in TokenProtectionBlocked test**
`cli/azd/internal/grpcserver/server_coverage3_test.go`

Test creates `ErrorWithSuggestion` with a link but only asserts on `GetSuggestion()` — the links serialization path through `ActionableErrorDetail` is unverified here. Consider adding `require.Len(t, actionable.GetLinks(), 1)` + URL/title assertions.

**3. New public API functions lack negative/edge-case tests**
`cli/azd/pkg/azdext/extension_error.go`

`GRPCStatusFromError`, `ActionableErrorDetailFromStatus`, `ActionableErrorDetailFromError`, `WrapErrorLinks`, `UnwrapErrorLinks` all have correct nil guards in the code, but no tests pin those contracts. Since extensions depend on these, consider adding test cases for: nil error, non-gRPC error, nil status, empty slices.

### 🔵 Low

**4.** `withActionableErrorDetail`'s early return when both suggestion and links are empty is untested.

**5.** `ExtensionError` and `ActionableErrorDetail` both carry suggestion+links for different wire directions — a brief cross-reference comment in the proto would reduce future confusion.

**6.** Internal helpers (`statusMessage`, `isAuthError`, `grpcAuthReason`) are only tested indirectly through `mapHostError`. Adequate coverage, but isolated unit tests would ease future refactoring.

---

Overall: Clean, well-structured refactoring. The medium items are hardening suggestions (logging + test coverage), not functional bugs. Nice improvement to the error UX for extensions! 👍

vhvb1989 (approved):

JeffreyCA:
Thanks for the review, applied 1, 2, 3, and 5 in 2d72abbd8

Skipped 4 (internal early-return on a dead branch) and 6 — statusMessage/isAuthError/grpcAuthReason are exercised end-to-end via `Test_mapHostError`

cli/azd/internal/grpcserver/errors.go
JeffreyCA: Given the supplementary nature of suggestions, I'd prefer to merge as-is, cut the next cli/azd/vX.Y.Z tag, and bump the six in-tree extension go.mod files in a follow-up PR (#7796)

wbreza (approved):
## Re-Review — Previous Feedback Addressed ✅

All 3 medium items from my earlier review have been addressed in the "Address feedback" commit:
- ✅ log.Printf added on WithDetails() failures
- ✅ Link assertions added in test
- ✅ Public API edge-case tests added (TestGRPCStatusFromError, TestActionableErrorDetailFromStatus, TestActionableErrorDetailFromError)

### Minor notes (non-blocking)

**populateExtensionErrorFromStatus early return for plain gRPC errors** — When a non-auth gRPC status has no ActionableErrorDetail, the function returns early without setting xtErr.Message = st.Message(). xtErr.Message stays as rr.Error(), which for gRPC errors includes the "rpc error: code = ..." prefix. This is intentional per the code comment, but worth a follow-up if the prefix surfaces to end users in practice.

**WrapErrorLinks/UnwrapErrorLinks now public** — These were private. If only used internally, consider keeping unexported to reduce API surface.

Clean refactoring, well-tested. LGTM 👍

vhvb1989 (approved):

jongio (commented):
Clean refactoring - the move from string concatenation to a structured proto detail is well-designed. A few notes after tracing through the code paths:

1. **statusMessage handling of nested gRPC errors** - the fallback to `st.Message()` when the error is already a gRPC status prevents the `rpc error: code = ... desc = ...` prefix from nesting. Good fix for the earlier bot comment.

2. **requireAuthErrorInfo helper** - iterating details rather than indexing `details[0]` is more robust now that statuses can carry both ErrorInfo and ActionableErrorDetail.

3. **Version skew** - acknowledged via the thread with @jongio. The plan to merge, tag, and bump extensions in #7796 is reasonable since the old path (suggestion in message) was already fragile (consumers had to parse the newline-separated format).

4. **withActionableErrorDetail early-return** - the `Suggestion == "" && len(Links) == 0` guard avoids attaching empty details, which keeps the wire cleaner for errors that have the type but no actionable content.

Tests are thorough - covering nil, plain, suggestion-only, auth-only, combined, links, round-trip unwrap, and wrapped status errors. CI green. Looks good to merge.

PR #7922, opened 2026-04-25
Enhance GitHub API error handling

### LINKED ISSUE #6646: [Issue] ai agent extension fails to init agent.yaml with failed to parse GitHub URL could not find a valid branch in url path
**Describe the bug**
Trying to follow the quickstart for hosted agents and azd ai agent fails when trying to run commands like `azd ai agent init -m https://github.com/microsoft-foundry/foundry-samples/blob/main/samples/python/hosted-agents/agent-framework/agent-with-foundry-tools/agent.yaml`

It fails with an error like: 

Image

Using --debug flag uncovers:
gh: Resource protected by organization SAML enforcement. You must grant your Personal Access token access to this organization. (HTTP 403)

**To Reproduce**
You must be in the Microsoft org - so this impacts Microsoft internal users only.

**Expected behavior**
A clear error message for the underlying problem with `gh` and authentication. 

**Additional context**
To fix this, you have to re-auth using `gh` and make sure to allow the Microsoft organization. Or if you are using GITHUB_TOKEN, modify that classic token in your GitHub settings, updating the Configure SSO setting.

JeffreyCA:
Fixes #6646

## Problem

`azd ai agent init` (and other GitHub URL flows) frequently fails with a misleading message:

> failed to parse GitHub URL: could not find a valid branch in the URL path. Tried branch names from 'main' to '...'

The real cause is usually SAML SSO enforcement, rate limiting, an expired token, or a private/EMU repo invisible to the active `gh` account. The branch-walker treated every non-2xx as "not a branch" and kept walking, swallowing the first real failure and surfacing nothing actionable.

## Changes

- **Typed `*github.ApiError`** (`pkg/tools/github/api_error.go`) with an `ApiErrorKind` enum (`SAMLBlocked`, `RateLimited`, `Unauthorized`, `Forbidden`, `NotFound`, `ServerError`, `Other`, `Unknown`). `parseApiError` reads the JSON error envelope from `gh api` stdout and falls back to the stderr `(HTTP NNN)` marker. `Cli.ApiCall` now returns this typed value, with the original error preserved via `Unwrap()` so `errors.Is` / `errors.AsType` still work.
- **Branch resolver short-circuits** (`pkg/templates/gh_source.go`) on classified failures (`SAMLBlocked` / `RateLimited` / `Unauthorized` / `Forbidden` / `ServerError`) instead of walking every candidate. 404s and unparseable errors still continue the walk, preserving the "branch with slashes" behavior.
- **`RepoNotAccessibleError`**: after the branch walk exhausts on 404s, probe `/repos/{owner}/{repo}`. If that's also 404, surface a typed repo-not-accessible error (covers private repos and EMU repos invisible to the active account).
- **Inline suggestions** (`pkg/templates/gh_errors.go`) wrap typed errors as `*internal.ErrorWithSuggestion` for each failure mode (SAML, rate-limit, 401, 403, 5xx, repo-not-accessible) with EMU / `gh auth status` guidance and relevant doc links. Inline wrapping (rather than `error_suggestions.yaml`) keeps the suggestion text co-located with the classifier and survives boundaries (e.g., gRPC) where the YAML pipeline doesn't run.

## Behavior

**Before:** `failed to parse GitHub URL: could not find a valid branch ...` for every failure mode.

**After (SAML):** `gh api .../branches/main: SAML SSO enforcement blocked the request (HTTP 403)` + actionable suggestion linking the SSO authorization docs.

**After (rate limit):** `gh api .../branches/main: GitHub API rate limit exceeded (HTTP 403)` + suggestion to authenticate or wait.

**After (private / EMU):** `repository github.com/<user>_microsoft/<repo> is not accessible (HTTP 404). It may not exist, may be private, or your account may not have access` + suggestion to verify with `gh auth status` and check enterprise account selection.

**After (5xx):** `GitHub returned a server error (HTTP 5xx). This usually indicates a transient issue on GitHub's side ...` + link to https://www.githubstatus.com/.

### DIFF
--- a/cli/azd/pkg/templates/gh_errors.go
+++ b/cli/azd/pkg/templates/gh_errors.go
@@ -1,0 +1,113 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package templates
+
+import (
+	"errors"
+
+	"github.com/azure/azure-dev/cli/azd/internal"
+	"github.com/azure/azure-dev/cli/azd/pkg/errorhandler"
+	"github.com/azure/azure-dev/cli/azd/pkg/tools/github"
+)
+
+// withGitHubSuggestion wraps the supplied error in an *internal.ErrorWithSuggestion
+// when the error is a recognized GitHub failure (typed *github.ApiError or
+// *RepoNotAccessibleError). The wrapping carries actionable guidance inline
+// in the error chain, which means:
+//
+//   - The core CLI ErrorMiddleware renders it directly without consulting the
+//     YAML rules pipeline (ErrorMiddleware short-circuits on
+//     *ErrorWithSuggestion).
+//   - Callers that don't have access to the YAML pipeline (e.g., extensions
+//     receiving stringified errors over gRPC) still see the suggestion as part
+//     of the wrapped error, because *ApiError keeps formatting itself with
+//     status + message.
+//
+// Returns the original error unchanged when no specific suggestion applies, so
+// other typed-error pipelines (auth, unknown gh failures) behave as before.
+func withGitHubSuggestion(err error) error {
+	if err == nil {
+		return nil
+	}
+
+	if apiErr, ok := errors.AsType[*github.ApiError](err); ok {
+		if s := suggestionForApiError(apiErr); s != nil {
+			s.Err = err
+			return s
+		}
+	}
+
+	if repoErr, ok := errors.AsType[*RepoNotAccessibleError](err); ok {
+		s := suggestionForRepoNotAccessible(repoErr)
+		s.Err = err
+		return s
+	}
+
+	return err
+}
+
+func suggestionForApiError(apiErr *github.ApiError) *internal.ErrorWithSuggestion {
+	switch apiErr.Kind {
+	case github.KindSAMLBlocked:
+		return &internal.ErrorWithSuggestion{
+			Message: "The GitHub organization that owns this repository requires SAML SSO " +
+				"authorization for your token before it can be used.",
+			Suggestion: "Open https://github.com/settings/tokens, find the personal access " +
+				"token you're using, click 'Configure SSO', and authorize the organization. " +
+				"If you're signed in with `gh auth login`, run " +
+				"`gh auth refresh -h github.com` and complete the SSO flow in the browser.",
+			Links: []errorhandler.ErrorLink{
+				{
+					URL: "https://docs.github.com/enterprise-cloud@latest/authentication/" +
+						"authenticating-with-saml-single-sign-on/" +
+						"authorizing-a-personal-access-token-for-use-with-saml-single-sign-on",
+					Title: "Authorizing a personal access token for use with SAML SSO",
+				},
+			},
+		}
+	case github.KindRateLimited:
+		return &internal.ErrorWithSuggestion{
+			Message: "GitHub API rate limit exceeded.",
+			Suggestion: "Authenticated requests have a much higher limit than anonymous ones. " +
+				"Run `gh auth login` (or set GITHUB_TOKEN / GH_TOKEN) and retry. " +
+				"If you're already authenticated, wait for the rate-limit window to reset " +
+				"(typically up to one hour).",
+			Links: []errorhandler.ErrorLink{
+				{
+					URL:   "https://docs.github.com/rest/overview/rate-limits-for-the-rest-api",
+					Title: "GitHub REST API rate limits",
+				},
+			},
+		}
+	case github.KindUnauthorized:
+		return &internal.ErrorWithSuggestion{
+			Message: "GitHub rejected the request as unauthenticated (HTTP 401).",
+			Suggestion: "Run `gh auth login` to sign in, or refresh an expired token with " +
+				"`gh auth refresh`. If you're using GITHUB_TOKEN / GH_TOKEN, regenerate the " +
+				"token and ensure it has the required scopes.",
+		}
+	case github.KindForbidden:
+		return &internal.ErrorWithSuggestion{
+			Message: "GitHub denied access to the requested resource (HTTP 403). The " +
+				"repository may be private, your token may be missing required scopes, or " +
+				"your account may not have permission.",
+			Suggestion: "Verify you can access the repository in a browser while signed in " +
+				"as the same GitHub account. If you're using a personal access token, ensure " +
+				"it includes the 'repo' scope. Run `gh auth status` to confirm which account " +
+				"gh is using.",
+		}
+	}
+	return nil
+}
+
+func suggestionForRepoNotAccessible(_ *RepoNotAccessibleError) *internal.ErrorWithSuggestion {
+	// Leave Message empty so the renderer falls back to RepoNotAccessibleError.Error(),
+	// which is already user-friendly and includes the repo slug — avoids duplication.
+	return &internal.ErrorWithSuggestion{
+		Suggestion: "Confirm the repository URL is correct and that the active gh account " +
+			"can see it. Run `gh auth status` to check which account is active. For Enterprise " +
+			"Managed Users (EMU), make sure the active account is the EMU account that owns " +
+			"this repository — a github.com URL may need to target a different host.",
+	}
+}
--- a/cli/azd/pkg/templates/gh_errors.go
+++ b/cli/azd/pkg/templates/gh_errors.go
@@ -97,6 +97,19 @@ func suggestionForApiError(apiErr *github.ApiError) *internal.ErrorWithSuggestio
 				"it includes the 'repo' scope. Run `gh auth status` to confirm which account " +
 				"gh is using.",
 		}
+	case github.KindServerError:
+		return &internal.ErrorWithSuggestion{
+			Message: "GitHub returned a server error (HTTP 5xx). This usually indicates a " +
+				"transient issue on GitHub's side rather than a problem with your request.",
+			Suggestion: "Wait a few minutes and try again. If the problem persists, check " +
+				"https://www.githubstatus.com/ for ongoing incidents.",
+			Links: []errorhandler.ErrorLink{
+				{
+					URL:   "https://www.githubstatus.com/",
+					Title: "GitHub Status",
+				},
+			},
+		}
 	}
 	return nil
 }
--- a/cli/azd/pkg/templates/gh_errors.go
+++ b/cli/azd/pkg/templates/gh_errors.go
@@ -53,16 +53,24 @@ func suggestionForApiError(apiErr *github.ApiError) *internal.ErrorWithSuggestio
 		return &internal.ErrorWithSuggestion{
 			Message: "The GitHub organization that owns this repository requires SAML SSO " +
 				"authorization for your token before it can be used.",
-			Suggestion: "Open https://github.com/settings/tokens, find the personal access " +
-				"token you're using, click 'Configure SSO', and authorize the organization. " +
-				"If you're signed in with `gh auth login`, run " +
-				"`gh auth refresh -h github.com` and complete the SSO flow in the browser.",
+			Suggestion: "If you signed in with `gh auth login`, run `gh auth refresh` and " +
+				"complete the SSO authorization in the browser when prompted (use " +
+				"`gh auth refresh -h <host>` for non-default hosts). If you're using a " +
+				"personal access token, open your GitHub token settings, locate the " +
+				"token, click 'Configure SSO', and authorize the organization that owns " +
+				"this repository.",
 			Links: []errorhandler.ErrorLink{
 				{
 					URL: "https://docs.github.com/enterprise-cloud@latest/authentication/" +
-						"authenticating-with-saml-single-sign-on/" +
-						"authorizing-a-personal-access-token-for-use-with-saml-single-sign-on",
-					Title: "Authorizing a personal access token for use with SAML SSO",
+						"authenticating-with-single-sign-on/" +
+						"about-authentication-with-single-sign-on",
+					Title: "About authentication with single sign-on",
+				},
+				{
+					URL: "https://docs.github.com/enterprise-cloud@latest/authentication/" +
+						"authenticating-with-single-sign-on/" +
+						"authorizing-a-personal-access-token-for-use-with-single-sign-on",
+					Title: "Authorizing a personal access token for use with single sign-on",
 				},
 			},
 		}
@@ -75,8 +83,8 @@ func suggestionForApiError(apiErr *github.ApiError) *internal.ErrorWithSuggestio
 				"(typically up to one hour).",
 			Links: []errorhandler.ErrorLink{
 				{
-					URL:   "https://docs.github.com/rest/overview/rate-limits-for-the-rest-api",
-					Title: "GitHub REST API rate limits",
+					URL:   "https://docs.github.com/en/rest/using-the-rest-api/rate-limits-for-the-rest-api",
+					Title: "Rate limits for the REST API",
 				},
 			},
 		}
--- a/cli/azd/pkg/templates/gh_source_test.go
+++ b/cli/azd/pkg/templates/gh_source_test.go
@@ -5,6 +5,7 @@ package templates
 
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
 	"path/filepath"
 	"strings"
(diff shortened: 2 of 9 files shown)


cli/azd/pkg/tools/github/api_error.go
```diff
@@ -0,0 +1,245 @@
...
+	// URL is the API URL that was requested (e.g., "https://api.github.com/repos/o/r/branches/main").
+	URL string
+	// Kind classifies the failure (SAML vs. rate-limit vs. plain 4xx etc.).
+	Kind ApiErrorKind
+	// StatusCode is the HTTP status code returned by the GitHub API (e.g., 401, 403, 404).
+	// It is 0 when no status code could be parsed (e.g., gh failed before issuing the request).
+	StatusCode int
+	// Message is the human-readable message GitHub returned in the JSON
+	// error body (e.g., "Resource protected by organization SAML enforcement.").
+	// Empty when the failure was not an HTTP-level error (e.g., gh not installed).
```
jongio, line 92: `KindUnknown` formats `e.Underlying` directly with `%s`. If `Underlying` is nil (unlikely from `parseApiError`, but this is a public exported type that could be constructed directly), this panics. Consider a nil guard:

```go
case KindUnknown:
    if e.Underlying != nil {
        return fmt.Sprintf("gh api %s: %s", e.URL, e.Underlying)
    }
    return fmt.Sprintf("gh api %s: unknown error", e.URL)
```

cli/azd/pkg/templates/gh_errors.go
```diff
@@ -0,0 +1,113 @@
...
+				"it includes the 'repo' scope. Run `gh auth status` to confirm which account " +
+				"gh is using.",
+		}
+	}
+	return nil
+}
+
+func suggestionForRepoNotAccessible(_ *RepoNotAccessibleError) *internal.ErrorWithSuggestion {
+	// Leave Message empty so the renderer falls back to RepoNotAccessibleError.Error(),
+	// which is already user-friendly and includes the repo slug — avoids duplication.
```
jongio, line 127: `KindServerError` isn't handled here - the function returns nil, so users hitting a 5xx get no suggestion. A brief `GitHub is experiencing issues - retry in a few minutes` would keep it consistent with the other kinds and give users something actionable.

cli/azd/pkg/templates/gh_source_test.go
```diff
@@ -684,3 +685,61 @@ func Test_ParseGitHubUrl_NotAuthenticated(t *testing.T) {
...
+	require.Equal(t, github.KindSAMLBlocked, apiErr.Kind)
+	require.Equal(t, 403, apiErr.StatusCode)
+
+	// The walk must short-circuit on the first auth failure rather than
+	// trying every candidate branch (would be 4 calls for "feature/sub/path/file.yaml").
+	require.Equal(t, 1, callCount, "expected branch walk to short-circuit on first access error")
+
+	// And the misleading "could not find a valid branch" message must NOT appear.
+	require.NotContains(t, err.Error(), "could not find a valid branch")
+}
```
jongio, line 744: The short-circuit test is good. Worth adding a companion test for the fallback path: mock all branch checks to return 404 (no `accessErr`), then mock the `/repos/{slug}` probe to also return 404, and verify the error surfaces as `*RepoNotAccessibleError`. This covers the private-repo/EMU codepath that's the other half of the PR's value.

jongio (commented):
Solid work - the structured error classification is a big improvement over the opaque `could not find a valid branch` message. The code is well-organized: the parsing logic in `api_error.go` is thorough, the branch resolver short-circuit is correct, and the error chain preservation via `Unwrap()` is properly maintained. A few suggestions below.

therealjohn:
Does this resolve https://github.com/Azure/azure-dev/issues/6646?

JeffreyCA:
> Does this resolve #6646?

It should, yes

wbreza (commented):
## Review — Enhance GitHub API error handling (#7922)

Nice work, @JeffreyCA. The structured `ApiErrorKind` classification and short-circuit logic are a real improvement — replacing the silent branch-walk swallow with actionable, user-facing messages is exactly the right approach. The typed error design is clean and the SAML/rate-limit/EMU coverage hits the most common pain points.

That said, the new error-handling layer carries meaningful risk without matching test coverage, and there are a few hardening gaps worth closing before merge. Findings below are grouped by priority and intentionally avoid re-flagging items already raised by @jongio ([nil guard for `KindUnknown`](https://github.com/Azure/azure-dev/pull/7922#discussion_r3156129654), [missing `KindServerError` suggestion](https://github.com/Azure/azure-dev/pull/7922#discussion_r3156129663), [companion test for fallback path](https://github.com/Azure/azure-dev/pull/7922#discussion_r3156129669)).

---

### 🔴 Critical

#### F1 — `gh_errors.go` has zero test coverage
**File:** `cli/azd/pkg/templates/gh_errors.go` (all 126 lines)

The user-facing suggestion layer — `withGitHubSuggestion()`, `suggestionForApiError()` (5 kinds → messages with docs links), and `suggestionForRepoNotAccessible()` — is the core value of this PR and has **no tests**.

Any regression here silently breaks the user experience for every classified error path. Table-driven tests are straightforward for this shape:

```go
func TestSuggestionForApiError(t *testing.T) {
    tests := []struct {
        name     string
        kind     github.ApiErrorKind
        wantNil  bool
        wantText string // substring match
    }{
        {"SAML", github.KindSAMLBlocked, false, "SAML SSO"},
        {"RateLimited", github.KindRateLimited, false, "rate limit"},
        {"Unauthorized", github.KindUnauthorized, false, "gh auth status"},
        {"Forbidden", github.KindForbidden, false, "token"},
        {"NotFound", github.KindNotFound, true, ""},
        // expand per Kind...
    }
    for _, tt := range tests {
        t.Run(tt.name, func(t *testing.T) {
            err := &github.ApiError{Kind: tt.kind, URL: "https://api.github.com/test"}
            got := suggestionForApiError(err)
            if tt.wantNil {
                require.Nil(t, got)
            } else {
                require.NotNil(t, got)
                require.Contains(t, got.Error(), tt.wantText)
            }
        })
    }
}
```

Also verify `withGitHubSuggestion()` dispatches correctly for both `*ApiError` and `*RepoNotAccessibleError` inputs, and returns the original error unchanged for unrecognized types.

---

### 🟠 High

#### F2 — Short-circuit behavior not verified in tests
**File:** `cli/azd/pkg/templates/gh_source_test.go`

The existing test asserts the *error value* on a classified failure, but doesn't confirm `resolveBranchAndPathInner()` actually **exits the loop early**. A SAML error on the first candidate should produce exactly 1 API call — not N (one per branch candidate).

**Suggested approach:** Use a call-counter mock (or verify the mock expectation count) to assert only a single `ApiCall` is made when the first branch probe returns `KindSAMLBlocked` or `KindRateLimited`.

---

#### F3 — `RepoNotAccessibleError` path never tested
**File:** `cli/azd/pkg/templates/gh_source_test.go`

Expanding on [jongio's suggestion](https://github.com/Azure/azure-dev/pull/7922#discussion_r3156129669) — the gap is broader than a single companion test. The entire `checkRepoAccessible()` → `RepoNotAccessibleError` integration is uncovered:

- All branch probes return 404 → repo probe returns 404 → `RepoNotAccessibleError` surfaced
- All branch probes return 404 → repo probe returns 200 → falls through to original "no valid branch" error
- Repo probe itself returns a classified error (e.g., SAML) → error propagated correctly

---

#### F4 — 4 of 8 `ApiErrorKind` values untested in classification
**File:** `cli/azd/pkg/tools/github/api_error_test.go`

`classifyKind()` is tested for `SAMLBlocked`, `RateLimited`, `NotFound`, and `Unknown`, but **Unauthorized** (401), **Forbidden** (403 non-SAML), **ServerError** (5xx), and **Other** (misc 4xx) have no test cases. Since classification drives both short-circuiting and user-facing suggestions, each path needs coverage.

```go
{"Unauthorized_401", 401, "Bad credentials", github.KindUnauthorized},
{"Forbidden_403_non_SAML", 403, "Resource not accessible", github.KindForbidden},
{"ServerError_500", 500, "Internal server error", github.KindServerError},
{"ServerError_502", 502, "Bad gateway", github.KindServerError},
{"Other_422", 422, "Validation failed", github.KindOther},
{"Other_409", 409, "Conflict", github.KindOther},
```

---

#### F5 — `parseApiError` edge cases untested
**File:** `cli/azd/pkg/tools/github/api_error_test.go`

`parseGitHubErrorBody` handles untrusted input (gh CLI stdout) but has no edge-case tests:

- **Malformed JSON** (truncated, HTML error pages from proxies)
- **Empty body** (gh crashes before writing)
- **Missing `message` field** (valid JSON, no expected keys)
- **Non-UTF-8 bytes**

These are realistic failure modes when gh CLI misbehaves or is behind a corporate proxy.

---

#### F6 — Error chain broken in `newGhTemplateSource()`
**File:** `cli/azd/pkg/templates/gh_source.go` (~line 218)

`ParseGitHubUrl` errors are returned bare (`return nil, err`), losing the call-site context. When this surfaces to the user, there's no indication *which operation* failed. Wrap with context:

```go
return nil, fmt.Errorf("failed to create template source for %q: %w", templateUrl, err)
```

---

#### F7 — No size validation on untrusted JSON deserialization
**File:** `cli/azd/pkg/tools/github/api_error.go` (`parseGitHubErrorBody`)

`json.Unmarshal` is called directly on gh CLI stdout with no size guard. While gh is a local process, a misbehaving proxy or a compromised gh binary could produce arbitrarily large output. A defensive cap is cheap:

```go
const maxErrorBodySize = 100_000 // 100 KB — generous for any error response
if len(stdout) > maxErrorBodySize {
    return nil
}
```

---

### 🟡 Medium

#### F8 — Inconsistent error return contract in `branchExists()`
**File:** `cli/azd/pkg/templates/gh_source.go` (~line 270)

`branchExists()` returns `*ApiError` for classified API failures but plain `error` for infrastructure failures (context canceled, gh not installed). Callers use `errors.AsType[*ApiError]` to branch — but the implicit contract ("nil ApiError means continue the walk") is fragile.

**Suggested fix:** Add a doc comment clarifying the contract and the caller's responsibility:

```go
// branchExists checks whether a branch exists via the GitHub API.
// Returns *ApiError for classified API failures (callers should inspect Kind
// to decide whether to short-circuit). Returns a plain error for
// infrastructure failures (context canceled, gh not installed) which should
// always abort the walk.
```

---

#### F11 — Duplicate auth check logic
**File:** `cli/azd/pkg/templates/gh_source.go`

`ensureGitHubAuthenticated` is called in `ParseGitHubUrl` (lines ~64 and ~110) and again in `newGhTemplateSource`. This means every template-source creation triggers 2–3 auth checks against gh CLI. Consider caching the auth status for the lifetime of the operation, or consolidating to a single check at the entry point.

---

#### F12 — `KindOther` not handled in suggestion generation
**File:** `cli/azd/pkg/templates/gh_errors.go` (~line 119)

Expanding on [jongio's `KindServerError` observation](https://github.com/Azure/azure-dev/pull/7922#discussion_r3156129663) — `KindOther` also falls through with no suggestion. Users hitting unclassified 4xx errors (422, 409, etc.) get the raw error with no guidance. A generic fallback would close this gap:

```go
case github.KindOther:
        Err: apiErr,
        Suggestion: "An unexpected GitHub API error occurred. " +
            "Check your token permissions with 'gh auth status' and retry.",
    }
```

---

#### F13 — Phrase matching and stderr regex edge cases undertested
**File:** `cli/azd/pkg/tools/github/api_error_test.go`

Only one SAML phrase variant is tested in classification. The `httpStatusRe` regex fallback (parsing HTTP status from stderr) has no edge-case tests. Consider parameterized tests for:

- All SAML phrase variants (`"Resource protected by organization SAML enforcement"`, `"saml_sso"` in message, etc.)
- Stderr with/without the expected `(HTTP NNN)` pattern
- Stderr with multiple HTTP codes (which one wins?)

---

### 🟢 Low

#### F9 — `KindOther` documentation imprecise
**File:** `cli/azd/pkg/tools/github/api_error.go`

The doc comment says `KindOther` covers "4xx codes" but the implementation catches any unclassified HTTP status that doesn't match the specific kinds — including codes outside the 4xx range that aren't 5xx. Update the comment to match reality:

```go
// KindOther represents any classified HTTP error that doesn't match
// a more specific kind (e.g., unrecognized 4xx or edge-case status codes).
```

---

#### F10 — Missing doc comment on `RepoNotAccessibleError`
**File:** `cli/azd/pkg/templates/gh_source.go`

`RepoNotAccessibleError` is an exported type with no Go doc comment. Per project conventions and `golint`, exported types should be documented:

```go
// RepoNotAccessibleError indicates that a GitHub repository was not
// accessible to the authenticated user. This typically means the
// repository is private, belongs to an EMU organization, or does not exist.
type RepoNotAccessibleError struct { ... }
```

---

### Summary

| Priority | Count | Areas |
|----------|-------|-------|
| 🔴 Critical | 1 | Testing (`gh_errors.go` untested) |
| 🟠 High | 6 | Testing (4), Error Handling (1), Security (1) |
| 🟡 Medium | 4 | API Design (1), Code Quality (2), Testing (1) |
| 🟢 Low | 2 | API Design (1), Code Quality (1) |

**Recommendation: Request Changes.** The structured error classification is solid design work, but the suggestion layer (`gh_errors.go`) and several classification paths ship without tests. Given this code directly shapes user-facing error messages across multiple failure modes, test coverage should match the ambition of the change before merge. The high-priority items (F2–F7) are all straightforward to address and would meaningfully de-risk the PR.

JeffreyCA:
Thanks for the thorough review @wbreza. Pushed `20f15d58` covering most of the feedback.

**Addressed**

- **F1** — added `pkg/templates/gh_errors_test.go` with table-driven coverage of `suggestionForApiError` per kind (8 kinds, including `KindUnknown`/`KindOther`/`KindNotFound` returning nil), `suggestionForRepoNotAccessible`, and `withGitHubSuggestion` dispatch (5 sub-cases incl. nil pass-through and chain preservation).
- **F3** — added `Test_ParseGitHubUrl_RepoAccessibleFallsThrough` (404 → repo 200 falls through to "no valid branch") and `Test_ParseGitHubUrl_RepoProbeReturnsClassifiedError` (repo probe SAML → typed `*ApiError` propagates, not wrapped in `RepoNotAccessibleError`).
- **F4** — added `TestClassifyKind_HttpStatusToKind` covering all 8 kinds incl. 401/403 (plain + SAML + primary/secondary rate limit)/404/422/409/500/502/503/0, plus `TestClassifyKind_SAMLPhraseVariants` for phrase coverage.
- **F5** — added malformed-JSON, HTML-body (proxy error page), and message-only-body rows to `TestParseApiError_StatusFromJSONBody`.
- **F6** — wrapped `ParseGitHubUrl` error in `newGhTemplateSource` with `fmt.Errorf("failed to create template source for %q: %w", urlArg, err)`.
- **F9** — tightened `KindOther` doc to match `classifyKind` behavior (classified non-2xx with known StatusCode, distinct from `KindUnknown`).
- **F13** — added `TestParseApiError_StderrHttpStatusFallback` for the `(HTTP NNN)` regex (7 cases: standard suffix, trailing text, multiple markers (first wins), missing marker, whitespace inside parens, 2-digit, lowercase `http`).

**Already addressed in earlier rounds (heads-up since timing overlapped with your review)**

- **F2** — `Test_ParseGitHubUrl_AccessErrorShortCircuits` already asserts `branchCallCount == 1` for the SAML-on-first-probe case.
- **F8** — added the contract doc comment to `branchExists` in `5a67833`: returns `(false, nil)` for 404/unparseable (continue walk), `(false, *ApiError)` for classified errors (short-circuit), plain error for infrastructure failures.
- **F10** — `RepoNotAccessibleError` got its doc comment in `328148f`.

**Deferred / decided against (happy to revisit)**

- **F7 (size cap on JSON deserialization)** — `gh` is a trusted local subprocess we already shell out to for everything.
- **F11 (duplicate auth check)** — investigated, deferring to a follow-up. `EnsureInstalled` is already idempotent (`sync.Once`), so the real duplication is one extra `gh auth status` subprocess for non-api URLs. Cleanly removing it requires one of: (a) adding an `authenticated bool` to `GitHubUrlInfo` (changes a public/gRPC-facing struct), (b) refactoring `ParseGitHubUrl` to not auth as a side effect (breaking change for the gRPC `ParseGitHubUrl` consumed by extensions), or (c) per-hostname caching inside `github.Cli` (cross-cutting, with mid-process login/logout invalidation concerns). All three are scope/API changes outside this PR's "improve error messages" scope.
- **F12 (generic suggestion for `KindOther`)** — `KindOther` is unreachable in practice for the `/repos/{slug}/branches/{name}` and `/repos/{slug}` endpoints this PR uses (they only return 200/401/403/404/5xx). The typed `*ApiError.Error()` is sufficient; a generic "check `gh auth status`" would be noise on the rare unclassified path.

wbreza (approved):
Looks great — thorough fix commits that address the review feedback comprehensively. The new test coverage is especially strong. Approving. ✅

## Resolution Summary (16 findings → 14 resolved)

| # | Severity | Finding | Status |
| F1 | 🔴 Critical | Unbounded body read in `readApiError` | ✅ Resolved |
| F2 | 🟠 High | Silent fallback masks root cause | ✅ Resolved |
| F3 | 🟠 High | Missing error wrapping in `newGhTemplateSource` | ✅ Resolved |
| F4 | 🟠 High | SAML detection too narrow | ✅ Resolved (6 variants tested) |
| F5 | 🟠 High | `branchExists` undocumented short-circuit | ✅ Resolved |
| F6 | 🟠 High | No test coverage for suggestion layer | ✅ Resolved (`gh_errors_test.go` created) |
| F7 | 🟠 High | No hard size cap on JSON parsing | ⚠️ Partially resolved |
| F8 | 🟡 Medium | `ApiErrorKind` incomplete test coverage | ✅ Resolved (all 8 kinds tested) |
| F9 | 🟡 Medium | Nil guard on `KindUnknown.Error()` | ✅ Resolved |
| F10 | 🟡 Medium | Missing `KindServerError` suggestion | ✅ Resolved |
| F11 | 🟡 Medium | Duplicate auth checks across layers | 📌 Open (non-blocking) |
| F12 | 🟢 Low | Missing doc comments | ✅ Resolved |
| F13 | 🟢 Low | Edge-case test gaps (malformed JSON, HTML) | ✅ Resolved |
| J1–J3 | — | jongio's findings (nil guard, KindServerError, fallback test) | ✅ All resolved |

### Non-blocking notes

1. **F7 (partial)** — No hard `io.LimitReader` cap was added, but the `HasPrefix` guard on content-type plus graceful JSON-error handling is a reasonable defense. Acceptable as-is; a size cap could be added later if large error bodies become an issue in practice.

2. **F11 (open)** — The duplicate auth-error checks across `ghCli` → `ghSource` → suggestion layer are an architectural concern, not a correctness bug. Fine to address in a follow-up if the layering is refactored.

### Note on `copilot-pull-request-reviewer[bot]` comments

The bot's ~10 comments suggesting `errors.As(err, &target)` instead of `errors.AsType[T]()` are **false positives**. `errors.AsType` is a generic helper added in Go 1.26, and `golangci-lint` passes cleanly. These can be dismissed.

jongio (approved):
Addresses my previous feedback. The nil guard on `KindUnknown`, the `KindServerError` suggestion, and the `RepoNotAccessibleError` fallback test are all in place. New test coverage in `gh_errors_test.go` is thorough - per-kind assertions plus the three integration tests for the branch-walk/repo-probe matrix cover the important paths.

vhvb1989 (approved):

PR #7929, opened 2026-04-27
chore(ai-agents): trim supportedHostedAgentRegions list

antriksh30:
Trims the `supportedHostedAgentRegions` allow-list in `init_locations.go` (used by `azd ai agent init` for the location prompt and model-location filter). Removes: canadaeast, centralus, eastus, germanywestcentral, italynorth, southcentralus, uaenorth, uksouth, westeurope.


amitbhave10 (approved):

jongio (commented):
CI is failing on all three platforms - the tests in `init_locations_test.go` use `eastus` (which this PR removes) as test data. The `AllSupported` and `SomeUnsupported` cases need their inputs swapped to a region that remains in the list (e.g., `eastus2`). The mutation test on line 74 should be updated too.

Separate question: the code comment on line 8-9 references the [hosted agents region docs](https://learn.microsoft.com/azure/foundry/agents/concepts/hosted-agents#region-availability), which currently lists only 4 regions (Australia East, Canada Central, North Central US, Sweden Central). After this PR, the list still has 18 regions. Is the code tracking a different/internal source, or are there more regions to trim? Might be worth updating the comment to point at the actual source if it's different from the public docs.

trangevi (changes requested):
This should be closed in favor of #7930

antriksh30:
closed in favor of https://github.com/Azure/azure-dev/pull/7930

PR #7932, opened 2026-04-27
feat(ai-agents): add --agent-endpoint flag to invoke command

antriksh30:
## Summary

Adds `--agent-endpoint` to `azd ai agent invoke` so users can invoke a deployed Foundry agent from **any directory** without needing an azd project, `azure.yaml`, or environment.

```bash
azd ai agent invoke \\
    --agent-endpoint https://<acct>.services.ai.azure.com/api/projects/<proj>/agents/<name>/versions/<v> \\
    --protocol responses \\
    'Hello!'
```

The endpoint is the same URL printed by `azd up` / `azd deploy` after agent deployment.

## Behavior

- Agent name is parsed from the URL — no positional name needed.
- Protocol defaults to `responses` when `--agent-endpoint` is used.
- Auth uses `DefaultAzureCredential` (works with `az login` / `azd auth login`).
- Server-assigned session ID is printed with a hint so users can pass `--session-id` on the next invoke to continue.
- Conversation hint printed for the same reason.
- Local persistence (stored IDs, OpenAPI cache) is skipped — there's no project to persist into.

## Validation

Mutually exclusive with:
- `--local`
- positional agent name
- `--port` (local-only flag)

URL must be HTTPS and match `/agents/<name>` or `/agents/<name>/versions/<v>`. Agent name is restricted to `[A-Za-z0-9_-]+` to avoid percent-encoding bugs in downstream URL substitution.

## Tested

- All validation paths return clear error messages.
- Live invoke against a real Foundry agent: first call creates conversation + assigns session, second call with `--session-id` + `--conversation-id` proves continuity (agent recalled prior turn).
- 16 parser unit tests covering edge cases (trailing junk, percent-encoded names, HTTP scheme, missing segments, etc.).
- `go test ./...` green across all 7 packages.

### DIFF
--- a/cli/azd/extensions/azure.ai.agents/internal/cmd/invoke.go
+++ b/cli/azd/extensions/azure.ai.agents/internal/cmd/invoke.go
@@ -37,6 +37,7 @@ type invokeFlags struct {
 	conversation    string
 	newConversation bool
 	protocol        string
+	agentEndpoint   string
 }
 
 type InvokeAction struct {
@@ -63,8 +64,13 @@ or large payloads with the invocations protocol.
 Use --local to target a locally running agent (started via 'azd ai agent run')
 instead of Foundry.
 
+Use --agent-endpoint to invoke a deployed agent from any directory without an
+azd project. Pass the endpoint printed by 'azd up' / 'azd deploy', for example:
+  https://<account>.services.ai.azure.com/api/projects/<project>/agents/<name>/versions/<v>
+
 Sessions are persisted per-agent — consecutive invokes reuse the same
-session automatically. Pass --new-session to force a reset.`,
+session automatically. Pass --new-session to force a reset. Per-agent
+persistence is disabled when --agent-endpoint is used (no azd project).`,
 		Example: `  # Invoke the remote agent on Foundry (auto-detects agent from azure.yaml)
   azd ai agent invoke "Hello!"
 
@@ -84,7 +90,13 @@ session automatically. Pass --new-session to force a reset.`,
   azd ai agent invoke --local "Hello!"
 
   # Start a new session (discard conversation history)
-  azd ai agent invoke --new-session "Hello!"`,
+  azd ai agent invoke --new-session "Hello!"
+
+  # Invoke a deployed agent from any directory using its endpoint URL
+  azd ai agent invoke \
+    --agent-endpoint https://acct.services.ai.azure.com/api/projects/proj/agents/my-agent/versions/1 \
+    --protocol responses \
+    "Hello!"`,
 		Args: cobra.RangeArgs(0, 2),
 		RunE: func(cmd *cobra.Command, args []string) error {
 			ctx := azdext.WithAccessToken(cmd.Context())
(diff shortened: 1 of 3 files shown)


trangevi (changes requested):
These kinds of one-off commands are counter to AZD's design. We need to have a fuller discussion on how we intend to proceed with these, if we do.

wbreza (commented):
## Supplemental Code Review

This review supplements existing feedback from @trangevi (design concern) and Copilot (security). Focused on new findings not yet raised.

### 🟠 High Priority

**F3 — Misleading `//nolint:gosec` justification**
The existing `//nolint:gosec // G704: endpoint is resolved from azd environment configuration` comments (~L459, ~L632 in invoke.go) are now factually incorrect when `--agent-endpoint` is used — the endpoint comes from CLI input, not azd environment config. This suppresses the lint warning that should be catching the token exfiltration risk already flagged by Copilot. Consider either removing the suppression for ephemeral paths or updating the comment to reflect the actual trust model.

**F4 — Duplicated branching in responsesRemote / invocationsRemote**
Both functions introduce the same `if agentEndpoint != "" { ... } else { ... }` pattern for endpoint resolution, credential setup, and session management. This creates two divergent code paths that must be maintained in parallel — every future change (new API versions, credential handling, session logic) must be applied twice. Consider extracting the endpoint/credential resolution into an early-resolve helper or strategy pattern to keep the core invocation logic unified.

**F5 — Auth errors in ephemeral mode lack actionable guidance**
When `DefaultAzureCredential` fails in ephemeral mode, the error is `"failed to get auth token: ..."` without any hint to run `az login` or `azd auth login`. Ephemeral mode targets users running outside a project context (quick tests, CI/CD), who are the most likely to encounter auth setup issues. Wrapping with an `ErrorWithSuggestion` pointing to login commands would significantly improve the experience.

**F6 — No integration tests for ephemeral invocation flow**
The 16 unit tests for `parseAgentEndpoint()` are well-structured, but there are zero tests for the ephemeral execution paths: flag mutual exclusivity (`--agent-endpoint` + `--local`), protocol defaulting to `responses`, `printEphemeralSessionHint()` output, or the `azdClient == nil` branching in responsesRemote/invocationsRemote. Given the security sensitivity of this feature, automated coverage of these paths is important.

**F7 — Hardcoded 30s timeout in `createConversation()`**
`createConversation()` uses `http.Client{Timeout: 30 * time.Second}` while the main invoke call respects `a.httpTimeout()` from the `--timeout` flag. In high-latency environments, conversation creation will time out and users can't fix it with the flag they'd expect to use. Consider using `a.httpTimeout()` consistently.

### 🟡 Medium Priority

**F8 — Agent name not URL-encoded in path construction**
Agent names from `parseAgentEndpoint()` are safe (validated by `isValidAgentNameSegment`), but names from `a.flags.name` (project path) are interpolated directly into URLs via `fmt.Sprintf` at L445, L602, L824 without `url.PathEscape()`. If name validation is ever relaxed, this becomes a URL injection vector.

**F11 — Incomplete agent name validation test coverage**
`isValidAgentNameSegment()` is the security gate preventing URL injection, but its test coverage misses edge characters (`@`, `!`, `*`, `:`, spaces, unicode) and boundary conditions (empty string, max-length names). Given this function's role as a security control, exhaustive coverage is warranted.

jongio (commented):
Supplementing existing feedback. One additional finding: `--new-session` and `--new-conversation` silently become no-ops when `--agent-endpoint` is used, since stored state resolution (`resolveStoredID`/`resolveConversationID`) is skipped entirely in ephemeral mode. Consider either validating these as incompatible (like `--local` and `--port`) or noting the limitation in help text. This is a small instance of the broader point @trangevi raised - ephemeral mode introduces a parallel execution path where several flags and behaviors diverge silently.

cli/azd/extensions/azure.ai.agents/internal/cmd/invoke.go
```diff
@@ -129,6 +141,41 @@ session automatically. Pass --new-session to force a reset.`,
...
+						exterrors.CodeInvalidParameter,
+						"cannot use --agent-endpoint with a positional agent name",
+						"the agent name is parsed from --agent-endpoint; remove the positional name argument",
+					)
+				}
+				if cmd.Flags().Changed("port") {
+					return exterrors.Validation(
+						exterrors.CodeInvalidParameter,
+						"cannot use --agent-endpoint with --port",
+						"--port only applies to local invocation; omit --port for remote agents",
```
jongio, line 163: The block validates mutual exclusivity with `--local`, positional name, and `--port`, but `--new-session` and `--new-conversation` are accepted silently even though they have no effect in ephemeral mode. Both flags only interact with stored state via `resolveStoredID`/`resolveConversationID`, which are bypassed when `azdClient == nil`. Consider adding similar validation here, or at minimum documenting the limitation.

antriksh30:
Closing this PR in favor of this updated PR: https://github.com/Azure/azure-dev/pull/8028

PR #7947, opened 2026-04-28
azure.ai.agents - adopt SDK root command and remove reserved flag conflicts

### LINKED ISSUE #7796: azure.ai.agents: adopt NewExtensionRootCommand and remove reserved global flag conflicts

JeffreyCA:
Fixes #7796 

This PR migrates the `azure.ai.agents` extension to the new azd extension SDK command helpers — `azdext.NewExtensionRootCommand`, `azdext.NewListenCommand`, and `azdext.NewMetadataCommand` — and consolidates per-command boilerplate that was duplicated across the cobra command tree.

## Why

Each azd extension previously hand-rolled its own root command, which meant manually registering the standard global flags (`--debug`, `--no-prompt`, `--cwd`, `-e/--environment`, `-o/--output`), wiring the env-var fallback for those flags (`AZD_*`), threading the access token onto the command context, and constructing the `ExtensionHost`/`AzdClient` lifecycle in `listen`. This duplication caused several real problems in the existing code:

- Three commands (`show`, `files list`, `files stat`) and the `addSessionFlags` helper redeclared `--output`/`-o`, conflicting with the SDK's reserved global flag and producing `flag redefined: "o"` panics when the SDK began enforcing its global registration.
- The `sessions` command group manually chained `PersistentPreRunE` to its parent and carried a long warning comment about an infinite-recursion footgun if the wrong `Parent()` reference was used.
- `setupDebugLogging` was called explicitly from every command's `RunE` (≈14 sites), and it was easy for a new command to forget to call it.
- Metadata generation and the `listen` command's setup were ~50 lines of boilerplate that drifted from the equivalent code in other extensions.

## Changes

The root command is now built via `azdext.NewExtensionRootCommand`, which provides the standard global flags, env-var fallback, trace context, and access token wiring. Per-subcommand customization of `--output` (default value, allowed values, completion) flows through `azdext.RegisterFlagOptions` instead of redeclaring the inherited flag. Commands that need flag values now read them from the typed `*azdext.ExtensionContext` passed into their constructors, replacing the previous `rootFlags` global; a small `ensureExtensionContext` helper makes it safe to construct commands with `nil` context in tests that don't exercise `RunE`.

The `listen` command is now created by `azdext.NewListenCommand`, with extension-specific service target and event handler registration moved into a `configureExtensionHost` callback. The `metadata` command is created inline via `azdext.NewMetadataCommand`, deleting the equivalent local implementation. Debug logging setup (`setupDebugLogging`) is hoisted to the root command's `PersistentPreRunE` so every subcommand inherits it via cobra's `EnableTraverseRunHooks`, removing the per-command setup/cleanup pairs.

The previous fragile `PersistentPreRunE` chaining on the `sessions` group is no longer needed and has been deleted along with its warning comment.

## Scope notes

`NewVersionCommand` was evaluated but not adopted in this PR: the SDK helper exposes only `name` and `version`, while the current command also surfaces `Commit` and `BuildDate` (populated at build time via `-ldflags`).

## Testing

`go build`, `go vet`, `go test ./...`, and `golangci-lint run ./...` all pass cleanly. Existing tests were updated to read flag values from `extCtx.OutputFormat` instead of poking at cobra flag internals, and a new `flag_options_test.go` asserts that per-command output overrides are registered via the SDK's annotation contract.


cli/azd/extensions/azure.ai.agents/go.mod
```diff
@@ -12,7 +12,7 @@ require (
 	github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/containerregistry/armcontainerregistry v1.3.0-beta.3
 	github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/resources/armresources v1.2.0
 	github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/resources/armsubscriptions v1.3.0
-	github.com/azure/azure-dev/cli/azd v1.23.14
+	github.com/azure/azure-dev/cli/azd v1.24.3-0.20260428211738-776bdddab0f8
```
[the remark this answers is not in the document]
JeffreyCA: The next semver module version does not exist yet, hence using a pseudo-version

jongio (commented):
Clean migration to `azdext.NewExtensionRootCommand`. Traced the full flag-threading path from SDK root through every subcommand - no stale `rootFlags` references survive and all `noPrompt`/`OutputFormat` reads happen after `PersistentPreRunE` populates `ExtensionContext`.

A few things I verified:

- `EnableTraverseRunHooks = true` (set by the SDK) means the manual `PersistentPreRunE` chains in `files.go` and `session.go` would double-fire if kept. Removing them is correct.
- `RegisterFlagOptions` stores config as cobra annotations, so the old `cmd.Flags().GetString("output")` assertions in tests correctly migrated to `assertOutputFlagOptions` checking `cmd.Annotations`.
- `ensureExtensionContext` returning a zero-value context for nil input keeps test constructors simple without risk - the SDK populates the real context before any `RunE` executes.
- The `--environment` removal from `init` is safe since the SDK's root persistent flag covers it, and `init`'s `RunE` reads `extCtx.Environment` before proceeding.

Pseudo-version in `go.mod` is expected per the existing conversation - just needs the semver tag bump before merge.

No blocking issues found.

wbreza (approved):
## ✅ Approved — Clean SDK Migration

The migration from manual flag registration to `azdext.NewExtensionRootCommand` is correct, complete, and maintains full backward compatibility. No merge-blocking issues found. Confirms jongio's prior review.

### Advisory Findings (P2-Medium — non-blocking)

1. **`RegisterFlagOptions` duplication** — The same 5-line output flag registration block is repeated 6+ times across `files.go`, `session.go`, and `show.go`. Consider extracting a shared helper like `registerOutputFlagOptions(cmd)`.

2. **Missing `ensureExtensionContext(nil)` unit test** — The nil → zero-value contract in `extension_context.go` is intentional but undocumented by tests. A single test pinning this behavior would prevent accidental regressions.

3. **No integration test for `extCtx` threading** — Tests verify command structure but none exercise the PersistentPreRunE → RunE pipeline to confirm `extCtx` is populated at execution time.

4. **`PersistentPreRunE` recursion test removed without replacement** — Justified since the SDK now owns this via `EnableTraverseRunHooks`, but a brief comment or replacement test would document that assumption.

### Verified Clean

| Dimension | Result |
|---|---|
| Correctness | Migration complete — no stale `rootFlags` references |
| Backward Compat | All CLI flags still work identically for users |
| Security | Posture improved — input validation strengthened via `AllowedValues` |
| Description Alignment | All 7 PR claims verified against actual diff |
| Concurrency | Improved — global mutable state eliminated |

cli/azd/extensions/azure.ai.agents/internal/cmd/root.go
```diff
@@ -5,30 +5,35 @@ package cmd
...
-		CompletionOptions: cobra.CompletionOptions{
-			DisableDefaultCmd: true,
-		},
+	rootCmd, extCtx := azdext.NewExtensionRootCommand(azdext.ExtensionCommandOptions{
+		Name:  "agent",
+		Use:   "agent <command> [options]",
+		Short: fmt.Sprintf("Ship agents with Microsoft Foundry from your terminal. %s", color.YellowString("(Preview)")),
+	})
+	rootCmd.SilenceUsage = true
+	rootCmd.SilenceErrors = true
```
trangevi, line 21: What do these two "silence" lines actually do?

trangevi (approved):

cli/azd/extensions/azure.ai.agents/internal/cmd/root.go
JeffreyCA: `SilenceUsage` controls whether the help text is printed any time a command returns an error, otherwise you get something like:

image

`SilenceErrors` avoids double printing command errors as the extension framework provides its own error reporting logic. Both were present prior to these changes as well:

```diff
- rootCmd := &cobra.Command{
- 		Use:           "agent <command> [options]",
- 		Short:         fmt.Sprintf("Ship agents with Microsoft Foundry from your  terminal. %s", color.YellowString("(Preview)")),
- 		SilenceUsage:  true,
- 		SilenceErrors: true,
```

PR #7964, opened 2026-04-29
Implement connect-ssh to job node

saanikaguptamicrosoft:
### `azd ai training job connect-ssh`

Adds a new subcommand that opens an interactive SSH session to a running
custom-training job's container, bringing parity with the AML CLI so users
can debug jobs without leaving the azd workflow.

### How it works

The command resolves the SSH service's `ProxyEndpoint` for the requested
node, then exec's the local OpenSSH client with a `ProxyCommand` that
points back to a hidden `_ssh-proxy` subcommand. That subcommand opens
a WebSocket tunnel to the AML NIP/TunDRA backend (authenticated with an
ARM bearer) and pipes the raw SSH protocol bytes between OpenSSH's stdio
and the WebSocket — making the cloud sshd look like a local TCP peer.

### Prerequisite

The job YAML must declare an SSH service with a public key to enable SSH for the job:

image

The corresponding private key is supplied via `--private-key-file-path`,
or via OpenSSH's default `~/.ssh` lookup if the flag is omitted.

### Notable design points
- Auth model — End-to-end: TLS to the AML edge, ARM bearer for the
WebSocket upgrade, public-key auth for sshd inside the container. The
outer SSH alias is a fixed label, so StrictHostKeyChecking is disabled
intentionally.
- Lifecycle — Pipe goroutines are bound to a cancellable session ctx;
a watcher closes the WebSocket on cancellation so the reader unblocks
cleanly. WebSocket ping/pong keepalive (30s ping, 60s read deadline)
detects half-open TCP.
- Error surfacing — Common pre-flight failures (job in terminal state,
node has no SSH service, SSH not yet Running, missing ProxyEndpoint)
produce actionable messages before ever invoking ssh.

## Testing
- Happy path
image

image

- If pvt key path is not defined (as fallback OpenSSH client fallsback to ~/.ssh defaults

image

image

- Help command
image

- Unhappy paths
image

image

image


wbreza (changes requested):
## Code Review — PR #7964: "Implement connect-ssh to job node"

**Reviewer:** AI-assisted review (8 parallel focus-area analyses + priority consensus)

### 🚨 P0 — Must Fix Before Merge

**1. Goroutine leak — no context cancellation in proxy goroutines** (`job_ssh_proxy.go`)
The two goroutines (`stdin→WS` and `WS→stdout`) never check `ctx.Done()`. If a session is cancelled while goroutines are blocked on `os.Stdin.Read()` or `conn.ReadMessage()`, they hang forever. This causes unbounded goroutine accumulation per cancelled session.

**Fix:** Add a cancellable context (`gctx, gcancel := context.WithCancel(ctx)`) and check `gctx.Done()` in both goroutines. Use `sync.WaitGroup` to ensure both goroutines exit before the function returns.

**2. errCh channel deadlock** (`job_ssh_proxy.go`)
`errCh` has buffer size 2 but the main goroutine only reads once (`<-errCh`). If both goroutines error simultaneously, the second send blocks forever with no receiver — guaranteed deadlock under concurrent failures.

**Fix:** Drain the channel after receiving the first error, or use a `done` channel pattern to signal both goroutines to exit.

**3. StrictHostKeyChecking=no silently disabled** (`job_connect_ssh.go`)
SSH host key verification is bypassed without user consent or warning. Users are silently exposed to MITM attacks on untrusted networks.

**Fix:** At minimum, print a warning to stderr. Ideally, use `StrictHostKeyChecking=accept-new` as default and only disable with an explicit `--skip-host-key-check` flag.

**4. Token never refreshed for long-lived sessions** (`job_ssh_proxy.go`)
The ARM token is acquired once with ~1hr TTL. Long SSH sessions silently fail after token expiry with no diagnostic message or refresh logic.

**Fix:** Implement token refresh logic or document the session time limit and surface an actionable error when the token expires.

### ⚠️ P1 — Should Fix Before Merge

**5. HTTP response body not closed on WebSocket dial failure** (`job_ssh_proxy.go`)
When `dialer.DialContext()` fails and `resp` is not nil, the response body is never closed — leaking file descriptors on every failed connection attempt.

**6. 0% test coverage** (all new files)
583 lines of new production code with zero `_test.go` files. Repo standard is ~50% coverage. Critical paths (proxy goroutines, endpoint resolution, URL validation regex, command construction) are completely untested.

**7. Empty PR description**
A 741-line, 11-file PR with no written description — only screenshots. Please add a summary covering: what the feature does, architecture (WebSocket proxy flow), prerequisites (SSH-enabled job), and testing approach.

**8. Missing WebSocket read/write timeouts** (`job_ssh_proxy.go`)
No deadlines on `conn.ReadMessage()` / `conn.WriteMessage()`. If the remote server dies without closing the connection, goroutines block indefinitely with no recovery.

**9. Incomplete command help text** (`job_connect_ssh.go`)
The `Short` description doesn't mention the node concept. The `Long` description doesn't explain the SSH service prerequisite (users must enable SSH in their job YAML).

### 📝 P2 — Can Address in Follow-Up

- Private key file path not validated before passing to SSH
- Hardcoded placeholder URL in `acquireARMToken()` — confusing to maintainers
- Missing error wrapping in `acquireARMToken()` — bare errors violate repo standards
- Code duplication in credential creation between `job_connect_ssh.go` and `job_ssh_proxy.go`
- Magic numbers (buffer 4096, timeout 60s) should be named constants
- Missing Go doc comments on public APIs (`GetServiceInstance`, `GetARMToken`, `GetTokenForScope`)

vhvb1989 (approved):

achauhan-scc (approved):

saanikaguptamicrosoft:
### P0
- 1 - **Goroutine cancellation** — done. Wrapped ctx with cancel; added a watcher goroutine that closes the WebSocket on cancellation so the downstream reader unblocks. `sync.WaitGroup` ensures the reader and pinger exit before return. The stdin reader can't be portably interrupted on Windows; documented that process exit reaps it (this subcommand is per-session and short-lived).
- 2 - **errCh** — done. Replaced the buffered channel with a size-1 `firstErr` + non-blocking send helper. Note: the reported deadlock isn't reachable today (buffer was 2, max 2 senders), but the new pattern is cleaner regardless.
- 3 - **StrictHostKeyChecking** — declining, would like to discuss. The destination is a fixed alias (`azureml-job`) that doesn't resolve; the real transport is the WebSocket which is already authenticated end-to-end via bearer-over-TLS and originates from a server-issued ProxyEndpoint URL. There's no MITM surface here, and `accept-new` would just churn `~/.ssh/known_hosts` with garbage entries. AML's own SSH tooling does the same. Happy to revisit if you see a threat I'm missing.
- 4 - **Token refresh** — declined. The bearer is consumed only at the WebSocket handshake; after the upgrade, the connection is a plain TCP-over-WS pipe and the server doesn't re-validate the token, so a mid-session refresh would be a no-op. The only real fix would be to drop and re-establish the WebSocket with a fresh token, but in-flight SSH bytes don't survive a connection swap and OpenSSH would see the drop as an error anyway. AML's own SSH tooling has the same behavior — sessions inherit the ARM token's TTL..

### P1
- 5 - **Response body close** — done.
- 6 - **Tests** — added `job_connect_ssh_test.go` and `job_ssh_proxy_test.go` covering endpoint resolution (10 cases), ProxyCommand-injection regex (16 cases), URL builder (6 cases), and service-endpoint extraction.
- 7 - **PR description** — added (above).
- 8 - **Read/write deadlines** — done. Added 30s ping interval via `WriteControl(PingMessage)`, 60s read deadline bumped on every pong/data frame. Half-open TCP now surfaces within ~60s instead of blocking indefinitely.
- 9 - **Help text** — expanded `Short`/`Long` to mention multi-node, the YAML prerequisite (with snippet), the SSH-status warmup window, and the `~/.ssh` fallback. Tightened `--node-index` description.

### P2
Folding in: extracted magic numbers to named constants (`sshTunnelWSPath`, `wsHandshakeTimeout`, `wsReadBufSize`, `pingInterval`, `pongTimeout`); removed the placeholder-URL hack in `acquireARMToken` (now uses the credential directly); errors properly wrapped.

PR #7974, opened 2026-04-29
[Agents Extension] Post deploy print update

### LINKED ISSUE #7942: fix the agent endpoint displayed after deploy
## Goal
Display the correct agent endpoint after deploy.

## Context
After deploy there is an endpoint shown for the playground and agent.

## Current Behavior
It shows the endpoint for the agent version, not the endpoint to use to invoke it like /responses or /invocations, etc. 

Image

## Desired Behavior
Show the endpoint you would use to invoke it. 

## Implementation Notes
May require multiple endpoints if there are multiple `protocols` defined. 

## Acceptance Criteria
- [ ] Unit tests added or updated
- [ ] Existing test suite passes 
- [ ] Lint or format passes 
- [ ] The endpoint shown has the protocols

trangevi:
Fixes #7942 

Update the print to show the endpoint a user would use to invoke the agent.
image

### DIFF
--- a/cli/azd/extensions/azure.ai.agents/internal/project/service_target_agent.go
+++ b/cli/azd/extensions/azure.ai.agents/internal/project/service_target_agent.go
@@ -33,6 +33,20 @@ import (
 // agentAPIVersion is the API version used for agent endpoint invocation URLs.
 const agentAPIVersion = "2025-11-15-preview"
 
+// displayableProtocolEntry defines a protocol that produces user-visible invocation endpoints.
+type displayableProtocolEntry struct {
+	Protocol  agent_api.AgentProtocol
+	URLPath   string // path suffix in the invocation URL
+	EnvSuffix string // suffix used in AGENT_{KEY}_{SUFFIX}_ENDPOINT env vars
+}
+
+// displayableProtocols is the single source of truth for protocols that produce
+// user-facing invocation endpoints and env vars.
+var displayableProtocols = []displayableProtocolEntry{
+	{Protocol: agent_api.AgentProtocolResponses, URLPath: "openai/responses", EnvSuffix: "RESPONSES"},
+	{Protocol: agent_api.AgentProtocolInvocations, URLPath: "invocations", EnvSuffix: "INVOCATIONS"},
+}
+
 // Ensure AgentServiceTargetProvider implements ServiceTargetProvider interface
 var _ azdext.ServiceTargetProvider = &AgentServiceTargetProvider{}
 
@@ -236,11 +250,19 @@ func (p *AgentServiceTargetProvider) Endpoints(
 
 	// Collect per-protocol endpoint env vars
 	var endpoints []string
-	for _, suffix := range []string{"RESPONSES", "INVOCATIONS"} {
-		key := fmt.Sprintf("AGENT_%s_%s_ENDPOINT", serviceKey, suffix)
+	for _, dp := range displayableProtocols {
+		key := fmt.Sprintf("AGENT_%s_%s_ENDPOINT", serviceKey, dp.EnvSuffix)
 		if val := azdEnv[key]; val != "" {
 			endpoints = append(endpoints, val)
 		}
+	}
+
+	if len(endpoints) == 0 {
+		return nil, exterrors.Dependency(
+			exterrors.CodeMissingAgentEnvVars,
+			fmt.Sprintf("no agent endpoint variables found for service %s", serviceKey),
+			"run 'azd deploy' to deploy the agent and set these variables",
+		)
 	}
 
 	return endpoints, nil
@@ -600,7 +622,7 @@ func (p *AgentServiceTargetProvider) deployPromptAgent(
 	}, nil
 }
 
-
+// deployHostedAgent deploys a container-based hosted agent to the Foundry service.
 func (p *AgentServiceTargetProvider) deployHostedAgent(
 	ctx context.Context,
 	serviceConfig *azdext.ServiceConfig,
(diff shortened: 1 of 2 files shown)


cli/azd/extensions/azure.ai.agents/internal/project/service_target_agent.go
```diff
@@ -231,9 +234,16 @@ func (p *AgentServiceTargetProvider) Endpoints(
...
+	var endpoints []string
+	for _, suffix := range []string{"RESPONSES", "INVOCATIONS"} {
+		key := fmt.Sprintf("AGENT_%s_%s_ENDPOINT", serviceKey, suffix)
+		if val := azdEnv[key]; val != "" {
+			endpoints = append(endpoints, val)
+		}
+	}
 
-	return []string{endpoint}, nil
+	return endpoints, nil
```
jongio, line 268: The old `Endpoints()` validated that `AGENT_{KEY}_NAME` and `AGENT_{KEY}_VERSION` env vars were set and returned a helpful error pointing users to `azd deploy`. The new code silently returns an empty slice if no protocol endpoints exist in the environment.

Consider adding a similar check after the loop:

```go
if len(endpoints) == 0 {
    return nil, exterrors.Dependency(
        exterrors.CodeMissingAgentEnvVars,
        fmt.Sprintf("no agent endpoint variables found for service %s", serviceKey),
        "run 'azd deploy' to deploy the agent and set these variables",
    )
}
```

This preserves the developer experience when someone calls `Endpoints()` before deployment.

cli/azd/extensions/azure.ai.agents/internal/project/service_target_agent.go
```diff
@@ -231,9 +234,16 @@ func (p *AgentServiceTargetProvider) Endpoints(
 		)
 	}
 
-	endpoint := p.agentEndpoint(azdEnv["AZURE_AI_PROJECT_ENDPOINT"], azdEnv[agentNameKey], azdEnv[agentVersionKey])
+	// Collect per-protocol endpoint env vars
+	var endpoints []string
+	for _, suffix := range []string{"RESPONSES", "INVOCATIONS"} {
```
jongio:
```suggestion
for _, suffix := range []string{"RESPONSES", "INVOCATIONS"} {
```

This hardcodes the protocol names separately from `protocolPath()`. If a new displayable protocol is added, both places need updating. Consider extracting a shared list or deriving the suffixes from the same source `protocolPath` uses.

cli/azd/extensions/azure.ai.agents/internal/project/service_target_agent.go
```diff
@@ -579,14 +592,15 @@ func (p *AgentServiceTargetProvider) deployPromptAgent(
...
+		promptProtocols,
 	)
 
 	return &azdext.ServiceDeployResult{
 		Artifacts: artifacts,
 	}, nil
 }
 
-// deployHostedAgent handles deployment of hosted container agents
+
```
jongio: The `deployHostedAgent` doc comment was removed here. Worth restoring - it helps navigate this large file.

cli/azd/extensions/azure.ai.agents/internal/project/service_target_agent.go
```diff
@@ -691,7 +705,7 @@ func (p *AgentServiceTargetProvider) deployHostedAgent(
```
jongio, line 727: This line uses spaces instead of tabs (and has a trailing space), causing the gofmt CI failure. Running `gofmt -s -w .` from the extension directory should fix it.

jongio (commented):
Clean approach to switching from version URLs to per-protocol invocation URLs. A few things:

1. **CI blocker**: gofmt is failing - mixed tabs/spaces in the `deployHostedAgent` area (see inline).
2. `Endpoints()` lost its validation that previously gave users a helpful error when env vars were not set. Now silently returns empty (see inline).
3. The `deployHostedAgent` doc comment was removed - worth restoring.
4. No test coverage for the updated `Endpoints()` or `registerAgentEnvironmentVariables` methods. The new helper functions have good tests, but the integration points that call them do not.

wbreza (commented):
﻿## Code Review — PR #7974

Nice work on the per-protocol endpoint refactor, Travis! The approach is clean and the test coverage for the new helpers is solid. A few items to consider:

### P1 — Should Fix

**1. Protocol list duplication between `Endpoints()` and `protocolPath()`**
`Endpoints()` hardcodes `["RESPONSES", "INVOCATIONS"]` (line ~238) while `protocolPath()` maintains its own switch (line ~777). These must stay in sync manually — adding a new protocol requires changes in both places with no compiler safety net.

Consider extracting a shared constant or deriving the env var suffix from protocol constants so there's a single source of truth.

**2. Stale `AGENT_{KEY}_ENDPOINT` env var not cleaned on upgrade**
The old single-endpoint env var (`AGENT_{KEY}_ENDPOINT`) will persist in existing azd environments after re-deploy, alongside the new per-protocol vars (`AGENT_{KEY}_RESPONSES_ENDPOINT`, etc.). Downstream consumers (hooks, scripts) reading the old var will get a stale URL.

Consider deleting the legacy key in `registerAgentEnvironmentVariables()`.

**3. `Endpoints()` returns empty slice silently**
If neither `AGENT_{KEY}_RESPONSES_ENDPOINT` nor `AGENT_{KEY}_INVOCATIONS_ENDPOINT` is set, `Endpoints()` returns `nil, nil`. Callers can't distinguish "no protocols configured" from "something went wrong." Consider adding a debug-level log when no endpoints are found after deploy.

### P2 — Consider Fixing

- Missing unit test for `registerAgentEnvironmentVariables` with per-protocol env vars — would catch regressions in key formatting.
- Missing edge case test for `nil` protocols slice (vs empty slice) in `TestAgentInvocationEndpoints`.
- Deleted Godoc comment on `deployHostedAgent` (line ~603) — looks accidental; consider restoring.

jongio (approved):
Addresses all my previous feedback. The `displayableProtocols` extraction is cleaner than what I suggested - single source of truth for both `Endpoints()` and `protocolPath()`. Legacy env var cleanup and default protocol fallback are good additions. CI is green.

vhvb1989 (approved):

wbreza (approved):
﻿## Code Review — PR #7974 (Re-review)

All previously raised items have been addressed cleanly:

- ✅ **Protocol list duplication** → shared displayableProtocols constant introduced
- ✅ **Stale env var cleanup** → legacy AGENT_{KEY}_ENDPOINT now explicitly cleared
- ✅ **Silent empty endpoints** → proper xterrors.Dependency error with user guidance
- ✅ **Test coverage** → TestRegisterAgentEnvironmentVariables, nil edge cases added
- ✅ **Godoc comment** → restored and improved on deployHostedAgent

Nice turnaround, Travis! LGTM.

PR #7999, opened 2026-04-30
feat(azure.ai.agents): add A2A endpoint protocol and agent card support

### LINKED ISSUE #8000: A2A endpoint and agent card support

adamra-msft:
Add support for setting A2A as an agent endpoint protocol and including agent card metadata in create/update agent requests.

Changes:
- Add AgentProtocolA2A constant and IsInvokable() helper to distinguish invokable vs deploy-only protocols
- Add AgentEndpoint, AgentCard, and AgentCardSkill types to API models
- Add corresponding YAML types and fields on ContainerAgent
- Map YAML agentEndpoint/agentCard to API request in createAgentAPIRequest
- Add PatchAgent operation (HTTP PATCH) for partial agent updates
- Deploy path: create version first, then PATCH agent-level endpoint/card
- Protocol resolution: require --protocol flag when multiple protocols declared
- Invoke validation: use IsInvokable() to block non-invokable protocols
- Add comprehensive tests for all new types and behaviors


cli/azd/extensions/azure.ai.agents/internal/project/service_target_agent.go
```diff
@@ -904,6 +904,24 @@ func (p *AgentServiceTargetProvider) createAgent(
...
+	// These are agent-level properties, not version-level, so they require
+	// a separate PatchAgent call after version creation.
+	if request.AgentEndpoint != nil || request.AgentCard != nil {
+		patchRequest := &agent_api.PatchAgentRequest{
+			AgentEndpoint: request.AgentEndpoint,
+			AgentCard:     request.AgentCard,
+		}
+
+		_, err := agentClient.PatchAgent(ctx, request.Name, patchRequest, agentAPIVersion)
+		if err != nil {
```
wbreza, line 878: **[Critical]** Partial Failure Recovery — No Rollback When PATCH Fails

If `PatchAgent` fails here, the agent version created at line 907 is orphaned. The error is returned but the user has no indication that the version was already created, and retrying the deployment will fail on version creation (already exists).

Add rollback logic to delete the orphaned version, or at minimum log the orphaned version name so users can clean up manually. The error message should distinguish "version created but metadata update failed" from other failures.

cli/azd/extensions/azure.ai.agents/internal/pkg/agents/agent_api/operations.go
```diff
@@ -176,6 +176,48 @@ func (c *AgentClient) UpdateAgent(ctx context.Context, agentName string, request
 	return &agent, nil
 }
 
+// PatchAgent performs a partial update on an agent using HTTP PATCH.
```
wbreza, line 179: **[High]** PatchAgent Is Completely Untested

This new public method has zero unit tests. Add tests covering: successful PATCH (HTTP 200), error status codes (400, 404, 500), request body serialization, and response parsing. Follow the existing test patterns in this file (e.g., TestDeleteSession, TestGetSession).

Also: this function signature is 162 characters, exceeding the repo's 125-char `lll` limit. Break across multiple lines.

cli/azd/extensions/azure.ai.agents/internal/pkg/agents/agent_api/models.go
```diff
@@ -133,9 +171,11 @@ type PromptAgentDefinition struct {
 
 // CreateAgentVersionRequest represents a request to create an agent version
 type CreateAgentVersionRequest struct {
-	Description *string           `json:"description,omitempty"`
-	Metadata    map[string]string `json:"metadata,omitempty"`
-	Definition  any               `json:"definition"` // Can be any of the agent definition types
+	Description   *string           `json:"description,omitempty"`
+	Metadata      map[string]string `json:"metadata,omitempty"`
```
wbreza: **[High]** Definition Field Serialized as Null in PATCH Requests

UpdateAgentRequest embeds CreateAgentVersionRequest where Definition any lacks an `omitempty` tag. When used in PatchAgent (service_target_agent.go line 916), Definition is nil, so JSON serialization produces `"definition": null`. This may cause the Azure API to clear the agent's definition or reject the request.

Either add `json:",omitempty"` to the Definition field, or create a dedicated PatchAgentRequest struct containing only the fields valid for PATCH operations (AgentEndpoint, AgentCard). (Note: this already looks like it's been done correctly in the diff with a dedicated PatchAgentRequest type!)

cli/azd/extensions/azure.ai.agents/internal/pkg/agents/agent_yaml/map.go
```diff
@@ -490,5 +498,34 @@ func createAgentAPIRequest(agentDefinition AgentDefinition, agentDef any) (*agen
...
 
+	// Map optional agent endpoint and card fields.
+	if agentEndpoint != nil {
+		protocols := make([]agent_api.AgentProtocol, len(agentEndpoint.Protocols))
+		for i, p := range agentEndpoint.Protocols {
+			protocols[i] = agent_api.AgentProtocol(p)
+		}
+		request.AgentEndpoint = &agent_api.AgentEndpoint{
+			Protocols: protocols,
+		}
```
wbreza, line 437: **[Medium]** Missing Input Validation on AgentEndpoint.Protocols and AgentCardSkill Fields

Protocol strings from YAML are mapped to API types without validation. Empty strings, whitespace, and invalid protocol names pass through to the Azure API producing confusing server-side errors. Add azd-level validation:
- Reject empty/whitespace protocol strings
- Validate AgentCardSkill required fields (ID, Name, Description) are non-empty
- Consider validating protocol values against known protocols

wbreza (dismissed):
## Code Review Summary

| Priority | Count |
|----------|-------|
| 🔴 Critical | 1 |
| 🟠 High | 2 |
| 🟡 Medium | 5 |
| 🟢 Low | 3 |
| **Total** | **11** |

**Overall Assessment:** Request Changes — 1 Critical finding (partial failure with no recovery path) and 2 High findings (untested deploy path, potential data loss from null Definition field) should be addressed before merge.

### 🔴 Critical

**Partial Failure Recovery — Version Created but PATCH Fails** (service_target_agent.go lines 911-920): The deploy path creates a version then PATCHes endpoint/card. If PATCH fails, the version is orphaned with no rollback or user notification. Retrying will fail on version creation. Add rollback or at minimum surface the partial state to the user.

### 🟠 High

**PatchAgent() and Deploy Path Untested** (operations.go, service_target_agent.go): Zero unit tests for PatchAgent. Zero integration tests for the two-step deploy path. The most critical new code path has no coverage.

**UpdateAgentRequest Definition Field Serialized as Null** (models.go lines 171-178): UpdateAgentRequest embeds CreateAgentVersionRequest where Definition lacks omitempty. PATCH sends "definition": null which may clear the agent definition server-side. Add omitempty or use the dedicated PatchAgentRequest type.

### �� Medium

- **Missing validation on AgentEndpoint.Protocols** — empty/invalid strings pass through to API (map.go 509-520)
- **Missing validation on AgentCardSkill fields** — empty ID/Name/Description reach API (map.go 514-525)
- **PatchAgent lacks timeout** — no context deadline; can hang indefinitely (operations.go 179-221)
- **Protocol resolution edge cases untested** — multiple non-invokable combos, empty string, invoke --protocol flag (helpers_test.go, invoke_test.go)
- **PatchAgent assumes HTTP 200 only** — no handling for 204 No Content (operations.go 202-204)

### 🟢 Low

- PatchAgent signature exceeds 125-char line limit (will fail lll linter)
- IsInvocable() allocates new slice per call; use switch for O(1)
- Protocol name building duplicated 3x across error paths (DRY)

### ✅ What Looks Good

- Clean A2A protocol abstraction with IsInvocable() gating mechanism
- Smart auto-selection of sole invocable protocol when mixed with non-invokable
- Solid round-trip JSON tests for new model types
- Well-structured YAML-to-API mapping with clear separation of concerns
- New YAML fields properly optional (omitempty) for backward compatibility

cli/azd/extensions/azure.ai.agents/internal/pkg/agents/agent_api/models.go
adamra-msft: Copilot already fixed this on the latest commit. Could you double check?

jongio (commented):
Clean A2A protocol support with good separation between version-level and agent-level concerns. The `PatchAgentRequest` type correctly isolates endpoint/card fields from `Definition` - the "null Definition" concern from earlier reviews doesn't apply here since `PatchAgentRequest` only serializes `agent_endpoint` and `agent_card`.

One suggestion: in `protocolFromAgentYaml` (helpers.go), when there's 1 invocable + 1 non-invocable protocol (e.g. `responses` + `a2a`), the error says "declares multiple protocols (responses, a2a)" with suggestion "use --protocol to specify which protocol to use for invocation". A user might try `--protocol a2a` and hit a second error. Consider including the valid choices in the suggestion, e.g. "use --protocol to specify which invocable protocol to use (supported: responses, invocations)" - or build it dynamically from `InvocableProtocols()` so it stays in sync.

Otherwise the existing reviews cover the main points. The two-step create+patch approach is sound.

cli/azd/extensions/azure.ai.agents/internal/pkg/agents/agent_api/models.go
jongio: Confirmed - `PatchAgentRequest` only has `AgentEndpoint` and `AgentCard` fields (no `Definition`), so the null serialization concern doesn't apply to the PATCH path. The type is correctly designed to avoid this.

adamra-msft:
> Clean A2A protocol support with good separation between version-level and agent-level concerns. The `PatchAgentRequest` type correctly isolates endpoint/card fields from `Definition` - the "null Definition" concern from earlier reviews doesn't apply here since `PatchAgentRequest` only serializes `agent_endpoint` and `agent_card`.
> 
> One suggestion: in `protocolFromAgentYaml` (helpers.go), when there's 1 invocable + 1 non-invocable protocol (e.g. `responses` + `a2a`), the error says "declares multiple protocols (responses, a2a)" with suggestion "use --protocol to specify which protocol to use for invocation". A user might try `--protocol a2a` and hit a second error. Consider including the valid choices in the suggestion, e.g. "use --protocol to specify which invocable protocol to use (supported: responses, invocations)" - or build it dynamically from `InvocableProtocols()` so it stays in sync.
> 
> Otherwise the existing reviews cover the main points. The two-step create+patch approach is sound.

Good point on generating the invocable protocols dynamically, I'll have Copilot take this approach

cli/azd/extensions/azure.ai.agents/internal/pkg/agents/agent_api/operations.go
adamra-msft: Fixed in commit a0b7ace

cli/azd/extensions/azure.ai.agents/internal/project/service_target_agent.go
adamra-msft: Fixed in commit a0b7ace to add a warning message

cli/azd/extensions/azure.ai.agents/internal/pkg/agents/agent_yaml/map.go
adamra-msft: Fixed in commit a0b7ace

cli/azd/extensions/azure.ai.agents/internal/pkg/agents/agent_yaml/yaml.go
```diff
@@ -611,6 +613,27 @@ type ProtocolVersionRecord struct {
 	Version  string `json:"version" yaml:"version"`
 }
 
+// AgentEndpoint describes the endpoint protocols an agent supports in agent.yaml.
+type AgentEndpoint struct {
+	Protocols []string `json:"protocols" yaml:"protocols"`
```
trangevi, line 604: How does this `Protocols` field relate to the `Protocols` field in the `ContainerAgent` object? Seems like they should be the same?

cli/azd/extensions/azure.ai.agents/internal/project/service_target_agent.go
```diff
@@ -904,6 +904,36 @@ func (p *AgentServiceTargetProvider) createAgent(
 	}
 
 	fmt.Fprintf(os.Stderr, "Agent version '%s' created successfully!\n", agentVersionResponse.Name)
+
+	// Patch agent-level fields (agent_endpoint, agent_card) if present.
+	// These are agent-level properties, not version-level, so they require
+	// a separate PatchAgent call after version creation.
+	if request.AgentEndpoint != nil || request.AgentCard != nil {
```
trangevi, line 869: So despite this being a "patch", having it this way means it can only be updated alongside a new agent version, is that intended? If that is the case, why can't the API handle this as a part of the create version request? It feels weird to have to have a separate call to provide this information if it's still a part of the agent object

trangevi (changes requested):

jongio (approved):
Addresses all prior feedback cleanly:

- `multiProtocolError` now dynamically lists valid invocable protocols
- PatchAgent has 5 tests covering success, error codes, and nil field omission
- Validation added for AgentEndpoint protocols and AgentCard fields (description, skills, skill IDs/names)
- WARNING message on partial failure gives users visibility
- `IsInvocable()` switched to O(1) switch
- AgentEndpoint/AgentCard moved from `CreateAgentVersionRequest` to `CreateAgentRequest`, correctly aligning with the agent-level vs version-level domain split

cli/azd/extensions/azure.ai.agents/internal/project/service_target_agent.go
jongio: Commit 30828f8 moved endpoint/card to `CreateAgentRequest`, so new agent creation now sends them in one call. The separate `PatchAgent` is still needed for the 'create new version of existing agent' path where `CreateAgentVersion` is version-scoped and can't carry agent-level properties. The two-step is an API constraint for the update-version path.

cli/azd/extensions/azure.ai.agents/internal/pkg/agents/agent_yaml/yaml.go
jongio: `ContainerAgent.Protocols` (`[]ProtocolVersionRecord`) captures protocol+version pairs for the CLI's protocol resolution logic (filtering invocable protocols, requiring `--protocol` when ambiguous). `AgentEndpoint.Protocols` (`[]string` in YAML, `[]AgentProtocol` in API) defines what the agent endpoint advertises to the platform. Different layers, but the naming overlap is confusing. There's currently no cross-validation between them, so a user could declare `responses` at the top level and `a2a` in the endpoint section without an error.

adamra-msft: Yes there are two `protocols` fields for a Foundry container agent. One is at the level of the container representing what the container itself supports directly. The other is at the agent endpoint level, which is what is exposed externally by Foundry. Since we have a A2A-to-responses translation layer, any container agent that implements the responses protocol can get A2A support for free by enabling the A2A protocol in the agent endpoint.

 I agree calling these both `protocols` can be a little confusing, but they are in different contexts of the agent object. The typespec changes have already been reviewed and approved separately.

cli/azd/extensions/azure.ai.agents/internal/project/service_target_agent.go
adamra-msft: Currently the `azd` commands call `CreateAgentVersion` which only sets up an agent version and implicitly creates an agent. The agent endpoint data lives at the agent level, which requires as second PATCH call to configure. Ideally we'd have one API that can do both, but that is not the case today.

cli/azd/extensions/azure.ai.agents/internal/pkg/agents/agent_yaml/yaml.go
trangevi: @therealjohn any concerns here? I think it's going to be confusing to users if there are two separate protocol lists that are different but related in possibly unclear ways, typespec approval or not. Perhaps we don't care since we're wanting to look at phasing out the agent.yaml file, but depending on where this ends up after that effort we might have the same concern.

As an aside on the object structure, to Jon's point about there being no-cross validation between them, would that be something which should be added? A check somewhere which would fail early if there's an attempt to create an agent which has an A2A protocol at the endpoint level, but no responses protocol at the container level? Is it ever possible for the container protocol list to be larger than the endpoint protocol list?

andscho-msft: The typespec already merged like that, I don't think it's wise to make this yaml repr diverge from typespec, agree it's weird

trangevi: We already have differences in here, for better or worse, this object isn't a 1:1 mapping to the agent API. If it's something we're agreeing is "weird", why are we hesitant to try and make it better, even if that means the tooling needs to do some kind of mapping?

trangevi: To be fair I'm usually very against that kind of divergence, but with how much usability concerns are already being raised with azd, I'm just trying to make sure I'm asking the questions.

adamra-msft: This is how the yaml looks with the current changes (truncated to just show the protocol related fields):

```yaml
 kind: hosted
 name: my-a2a-agent
 description: A hosted agent with A2A endpoint support
 protocols:
   - protocol: responses
     version: 1.0.0
 agentEndpoint:
   protocols:
     - responses
     - a2a
```

What is the suggestion on how to do this differently? Should we call it `endpointProtocols` under `agentEndpoint`?

adamra-msft: Also Copilot's feedback on this:

> My take: the confusion is understandable but the current structure is correct and defensible:
> 
>  1. They represent genuinely different things — container protocols = what the container speaks (invocation wire protocol), endpoint protocols = what the service advertises externally (discovery/routing). These should be independent since an agent can speak responses internally while being discoverable via A2A externally.
>  2. The nesting disambiguates — protocols vs agentEndpoint.protocols is clear enough in YAML. They're not siblings; one is nested under a distinct parent object.
>  3. Renaming the top-level field is a breaking change — protocols is an established field used by all existing hosted agents. Renaming it to containerProtocols would break every existing agent.yaml.
>  4. Documentation is the right fix — a short comment in the YAML schema or a note in the docs explaining the distinction ("container protocols define what your agent implements; endpoint protocols define how the service exposes it") would address the confusion without any code changes.

wbreza (commented):
## Code Review — PR #7999

### Summary

Solid implementation of A2A endpoint protocol support with clean type separation and comprehensive tests. Most earlier review feedback from wbreza has been well-addressed across commits. A few new items to consider.

### 🟡 Should Fix

**1. No cross-validation between container protocols and endpoint protocols**
- A user can declare `a2a` in `agentEndpoint.protocols` without declaring `responses` in container `protocols`, which would create a broken agent (the A2A-to-responses translation layer requires the responses protocol at the container level).
- @jongio noted: *"There's currently no cross-validation between them, so a user could declare responses at the top level and a2a in the endpoint section without an error."*
- **Suggestion:** Add a validation check in `map.go`: if `agentEndpoint.protocols` contains `a2a`, require `responses` in container `protocols`. This catches a misconfiguration early instead of failing at runtime.

**2. Protocol strings cast without validation against known constants**
- In `map.go`, user-supplied protocol strings from YAML are directly cast to `agent_api.AgentProtocol` without validating against known values (`activity_protocol`, `invocations`, `responses`, `a2a`). Invalid/typo'd protocols will pass silently to the Azure API.
- **Suggestion:** Validate against known protocol constants before casting, or at minimum warn on unrecognized values.

**3. Silent error swallowing in tool conversion**
- In `map.go` `convertYamlToolsToApiTools`, the comment says "Log error and skip this tool" but only does `continue` with no actual logging. Failed tool conversions are silently dropped, potentially degrading agent behavior without any user-visible indication.
- **Suggestion:** Add a warning log or surface the error to the user.

**4. Missing test for `activity_protocol` in `TestProtocolFromAgentYaml`**
- Tests cover `responses`, `invocations`, and multi-protocol edge cases, but `activity_protocol` is not tested as a valid single protocol input.
- **Suggestion:** Add a table-driven test case.

### 🟢 Nitpick

**5. URL path parameters not escaped in `operations.go`**
- `agentName` is interpolated directly into URLs via `fmt.Sprintf` without `url.PathEscape()`. This is a pre-existing pattern across all operations — not introduced by this PR — but worth noting for a future cleanup pass.

### ✅ What's Done Well

- **Clean type separation** — `PatchAgentRequest` correctly isolates agent-level fields from version-level fields, avoiding the null `definition` serialization issue
- **Comprehensive tests** — PatchAgent unit tests, round-trip serialization tests, and validation edge case tests
- **Good iteration** — All wbreza feedback addressed (dedicated request type, tests, validation, orphaned version warning, dynamic protocol suggestions)
- **Solid error messaging** — `multiProtocolError()` dynamically lists invocable protocols; orphaned version name surfaced on PATCH failure

cli/azd/extensions/azure.ai.agents/internal/pkg/agents/agent_yaml/yaml.go
trangevi: Per offline discussion, good to move ahead with what's in here

trangevi (approved):

trangevi (approved):
Approving again after merge conflict resolution.

PR #8005, opened 2026-05-01
Add azure.logicappsstandard extension

### LINKED ISSUE #812: Support for Logic Apps

**Output from `azd version`**
`azd version 0.3.0-beta.1 (commit c5344160ed49bfda75281fba28b5e9ac7b96670e)`

**Output from `az version`** 
```
{
  "azure-cli": "2.40.0",
  "azure-cli-core": "2.40.0",
  "azure-cli-telemetry": "1.0.8",
  "extensions": {
    "application-insights": "0.1.7",
    "arcdata": "1.4.5",
    "init": "0.1.0",
    "managementpartner": "0.1.3"
  }
}
```
**Describe the bug**
After creating the infra I am trying to deploy Logic App Standard workflow content using the following yaml snippet:

```
services:
  workflows:
    project: src\workflows
    host: function
```

This results in:

` azd deploy
Deploying service workflows...
Error: deploying service: packaging service workflows: dotnet publish on project 'C:\Users\user\source\repos\workspace\logicappstandard\src\workflows' failed: exit code: 1, stdout: MSBuild version 17.3.1+2badb37d1 for .NET
MSBUILD : error MSB1003: Specify a project or solution file. The current working directory does not contain a project or solution file.     
, stderr: : exit status 1`

**To Reproduce**
When using the Logic App Standard Visual Studio Code Extension:
[https://learn.microsoft.com/en-us/azure/logic-apps/create-single-tenant-workflows-visual-studio-code](url)

the following folder structure is created:

```
MyBundleBasedLogicAppProjectName
| .vscode
| Artifacts
  || Maps 
     ||| MapName1
     ||| ...
  || Schemas
     ||| SchemaName1
     ||| ...
| WorkflowName1
  || workflow.json
  || ...
| WorkflowName2
  || workflow.json
  || ...
| workflow-designtime
| .funcignore
| connections.json
| host.json
| local.settings.json

```

**Expected behavior**
azd cli should zip the folder structure and its contents and deploy it to the Logic App.

**Environment**
    * Visual Studio Code 1.71.2
    * Azure Logic Apps (Standard) Extension v1.0.40

ronaldbosma:
Add `azure.logicappsstandard` extension that makes it possible to package Logic Apps Standard projects. It includes support for [custom code projects](https://learn.microsoft.com/en-us/azure/logic-apps/create-run-custom-code-functions). It uses the host specific exclusion file (e.g. `.funcignore`) to exclude files and folders from the package.

Fixes #812. Discussed in #6956.

### Design decisions

- There's no validation that checks if the folder containing the Logic App Standard project contains certain files, because there are no specific required files. Even deploying an empty `.zip` file to a Logic App succeeds.
- Only accept `function` as host in combination with `language: logicappsstandard` because Logic App Standard uses the Azure Function Runtime under the hood. This prevents the extension from silently proceeding through packaging when another host is configured and likely failing with a confusing error at deploy time.

### Usage

Sample usage in `azure.yaml` for Logic App project without custom code project where the `./src/logicapp` folder contains the Logic App assets like the `host.json` and workflows:

```yaml
services:
  logicapp:
    project: ./src/logicapp
    host: function
    language: logicappsstandard
```

Sample usage in `azure.yaml` for Logic App project with custom code project (`./src/logicapp/Functions/Functions.csproj`), where the `./src/logicApp/Workflows` folder contains the Logic App assets like the `host.json` and workflows:

```yaml
services:
  logicapp:
    project: ./src/logicapp
    dist: Workflows
    host: function
    language: logicappsstandard
    customCodeProject: Functions/Functions.csproj
```


ronaldbosma:
Note that I don't have any prior experience with Go and with creating azd extensions. I've used the docs under https://github.com/Azure/azure-dev/tree/main/cli/azd/docs/extensions, existing extensions under https://github.com/Azure/azure-dev/tree/main/cli/azd/extensions and Copilot to create this extension.

Let me know if anything needs to change.

cli/azd/extensions/azure.logicappsstandard/build.sh
```diff
@@ -0,0 +1,66 @@
...
+
+    echo "Building for $OS/$ARCH..."
+
+    # Delete the output file if it already exists
+    [ -f "$OUTPUT_NAME" ] && rm -f "$OUTPUT_NAME"
+
+    # Set environment variables for Go build
+    GOOS=$OS GOARCH=$ARCH go build \
+        -ldflags="-X '$APP_PATH.Version=$EXTENSION_VERSION' -X '$APP_PATH.Commit=$COMMIT' -X '$APP_PATH.BuildDate=$BUILD_DATE'" \
+        -o "$OUTPUT_NAME"
```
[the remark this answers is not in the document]
ronaldbosma: Note that this script was generated as is by `azd x init`

eng/pipelines/release-ext-azure-logicappsstandard.yml
```diff
@@ -0,0 +1,33 @@
...
+
+extends:
+  template: /eng/pipelines/templates/stages/1es-redirect.yml
+  parameters:
+    stages:
+      - template: /eng/pipelines/templates/stages/release-azd-extension.yml
+        parameters: 
+          AzdExtensionId: azure.logicappsstandard
+          SanitizedExtensionId: azure-logicappsstandard
+          AzdExtensionDirectory: cli/azd/extensions/azure.logicappsstandard
```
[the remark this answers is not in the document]
ronaldbosma: FYI. I used the `azure.appservice` pipeline as a reference and replace appservice with logicappsstandard.

wbreza (commented):
## 🎉 Great first extension contribution!

@ronaldbosma — thanks for putting this together. The Logic Apps Standard extension is well-structured and clearly follows the patterns established by the existing extensions. A few specific things that stood out positively:

- **Extension structure** follows established patterns well (extension.yaml, build scripts, main.go, CI/release workflows)
- **Error handling** is solid — proper wrapping with `fmt.Errorf("context: %w", err)` throughout
- **Context propagation** is correct — all I/O functions receive `ctx context.Context`
- **Interface compliance check** (`var _ azdext.FrameworkServiceProvider = ...`) is a great pattern
- **Test coverage** of the packaging and custom code project flows is thorough
- Correct import ordering, copyright headers, and the PR description accurately reflects the implementation

That said, I found a few items to address before merge — one is a build blocker.

---

## 🔴 Critical — Blocks Merge

**F1: Missing `version` variable — compile failure**

`internal/cmd/root.go:24` calls `azdext.NewVersionCommand(extensionID, version, ...)`, but no `version` variable is declared in the `cmd` package. The build scripts set `-X '...internal/cmd.Version=...'` via ldflags, but the target variable doesn't exist yet.

**Fix:** Add `internal/cmd/version.go` with `var Version string` (or lowercase `version`), matching the ldflags path. See `azure.appservice/internal/cmd/version.go` for the pattern.

---

## 🟠 High — Should Fix

**F2: No tests for the `cmd` package**

`internal/cmd/listen.go` and `root.go` have no test coverage. Could you add basic smoke tests for command wiring and listen configuration?

**F3: `fmt.Printf` in Initialize bypasses output formatting**

`framework_service_logicappsstandard.go:33` uses `fmt.Printf(...)` while all other lifecycle methods (Restore, Build, Package) correctly use the progress reporter callback. This will produce inconsistent output in the azd UX. Would you mind removing the Printf or routing through the progress reporter if the Initialize interface supports it?

---

<details>
<summary><strong>🟡 Medium — Recommended (7 items)</strong></summary>

**F4: Module path inconsistency**
`go.mod` declares the module as `azure.logicappsstandard` (with dot), but existing extensions use dot-free names (`azureappservice`, `azureaiagent`, `azurecodingagent`). Consider renaming for consistency.

**F5: Missing `.golangci.yaml` and `cspell.yaml`**
All other extensions with CI lint workflows include these config files. Without them the lint workflow may fail or use incorrect defaults. Suggest copying and adapting from `azure.appservice`.

**F6: Missing `ci-build.ps1` / `ci-test.ps1`**
These CI scripts exist in other extensions — please verify the lint workflow runs successfully without them.

**F7: Tests use `context.Background()` instead of `t.Context()`**
6 instances in the test file. Per repo standards (AGENTS.md), Go 1.26+ tests should use `t.Context()` for proper test lifecycle integration.

**F8: Error path test coverage gaps**
Happy paths are well-tested, but consider adding cases for: dotnet command failures, invalid `.csproj` paths, and nil property handling.

**F9: Path bounds checking on user-controlled paths**
`resolveCustomCodeProjectPath()` and `Package()` join user-controlled values (`customCodeProject`, `OutputPath`) via `filepath.Join` without validating the result stays within the project directory. Practical risk is low (azure.yaml is author-controlled), but a `filepath.Rel` bounds check would add defense in depth.

**F10: Cross-platform test concerns**
The fake dotnet shell script in tests won't execute on Windows CI. Consider adding `runtime.GOOS` handling or skip logic.

</details>

<details>
<summary><strong>💬 Nits (2 items)</strong></summary>

**F14:** CHANGELOG line 5 — typo: "poperty" → "property"

**F15:** `NewRootCommand` in root.go and `NewLogicAppsStandardPackagingFrameworkServiceProvider` lack Go doc comments on exported functions.

</details>

---

**Summary:** Please fix the critical compile blocker (F1) and the two high items before merge. The medium/nit findings are recommendations that would bring the extension fully in line with repo conventions. Really solid work overall — looking forward to seeing this land! 🚀

cli/azd/extensions/azure.logicappsstandard/internal/cmd/root.go
```diff
@@ -0,0 +1,29 @@
...
+package cmd
+
+import (
+	"github.com/azure/azure-dev/cli/azd/pkg/azdext"
+	"github.com/spf13/cobra"
+)
+
+const (
+	extensionID = "azure.logicappsstandard"
+	version     = "0.0.1"
```
jongio: `version` is a `const`, but the build scripts use `-X` ldflags to inject version info at build time. Go's `-X` flag can only set package-level `var` declarations, not constants - so the ldflags are silently ignored. `Commit` and `BuildDate` (also in the ldflags) aren't declared anywhere either.

Every other extension solves this with a dedicated `internal/version/version.go`:

```go
package version

var (
    Version   = "dev"
    Commit    = "none"
    BuildDate = "unknown"
)
```

Then import it here and pass `version.Version` to `azdext.NewVersionCommand`.

You'll also need to update build.sh/build.ps1 to target `azure.logicappsstandard/internal/version` instead of `$EXTENSION_ID/internal/cmd` - see `azure.ai.finetune/build.sh` line 36 for the correct pattern. The current scripts inherited the wrong ldflags path from `azure.appservice` (which also has this mismatch, but that's a separate issue).

jongio (commented):
Supplementing wbreza's review with two clarifications on the version/build system and CI integration.

**The extension compiles fine** - the `const version` in root.go does provide a value to `azdext.NewVersionCommand`, so this isn't a build blocker. The real issue is that the build scripts' ldflags can't inject version info at build time (details on the inline comment below).

The missing `ci-build.ps1`, `ci-test.ps1`, and `.golangci.yaml` are the actual merge blockers - the release pipeline template (`build-azd-extension.yml` lines 56/71) directly references `ci-build.ps1` and `ci-test.ps1` by path and will fail without them. You can copy these from `azure.ai.finetune` and adapt (it's the most up-to-date pattern). The `.golangci.yaml` is also needed for the lint workflow - again, `azure.ai.finetune` has a clean example.

Supply-chain note: CI/pipeline files are standard copies of the azure.appservice pattern. Dependencies in go.mod look appropriate. Nothing unusual.

ronaldbosma:
Hi @wbreza. Thanks for the quick review. I've fixed the critical and high. Will take a look at the medium soon.

**F1: Missing version variable — compile failure**:  
Fixed with https://github.com/Azure/azure-dev/pull/8005/changes/3ebcb698333bac6726b22cdaab5a34b4d9512723.

**F2: No tests for the cmd package**:  
Are tests for these files necessary? I don't see them on any other extension and the files to test contain almost no logic.

**F3: fmt.Printf in Initialize bypasses output formatting**:  
I've removed the line. Note that I copied this from the [demo](https://github.com/ronaldbosma/azure-dev/blob/main/cli/azd/extensions/microsoft.azd.demo/internal/project/framework_service_demo.go#L34).

cli/azd/extensions/azure.logicappsstandard/internal/cmd/root.go
ronaldbosma: Thanks @jongio. I'll have a look at the other extensions.

My solution was based on https://github.com/ronaldbosma/azure-dev/blob/main/cli/azd/docs/extensions/extension-migration-guide.md#after-recommended-1 by the way. Maybe that guide should be updated as well.

cli/azd/extensions/azure.logicappsstandard/build.sh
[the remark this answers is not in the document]
ronaldbosma: Resolved by introducing a separate `azure.logicappsstandard/internal/version/version.go` file with the necessary properties

eng/pipelines/release-ext-azure-logicappsstandard.yml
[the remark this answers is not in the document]
ronaldbosma: Includes the necessary PowerShell scripts

jongio (commented):
All my prior findings are addressed - version.go, build script paths, CI scripts, and .golangci.yaml are all in place. The path traversal protection (`resolvePathWithinBase`) and nil checks are solid additions.

One remaining issue: the module rename from `azure.logicappsstandard` to `azurelogicappsstandard` (commit 4c99fcb) created a mismatch in the ldflags paths. See inline comment.

cli/azd/extensions/azure.logicappsstandard/build.sh
```diff
@@ -0,0 +1,66 @@
...
+        "windows/amd64"
+        "windows/arm64"
+        "darwin/amd64"
+        "darwin/arm64"
+        "linux/amd64"
+        "linux/arm64"
+    )
+fi
+
+APP_PATH="$EXTENSION_ID/internal/version"
```
jongio: The Go module is `azurelogicappsstandard` (no dots), but this line uses `$EXTENSION_ID` which resolves to `azure.logicappsstandard` (with dots). The `-X` flag needs the Go import path, not the extension ID, so `azure.logicappsstandard/internal/version.Version` won't match the actual symbol at `azurelogicappsstandard/internal/version.Version`.

Same issue in `build.ps1` (line `$APP_PATH = "$env:EXTENSION_ID/internal/version"`) and `ci-build.ps1` (hardcoded `azure.logicappsstandard/internal/version`).

The fix is to hardcode the Go module name like `azure.ai.agents` does:

```bash
APP_PATH="azurelogicappsstandard/internal/version"
```

And in ci-build.ps1, change the ldflags to target `azurelogicappsstandard/internal/version.*`.

ronaldbosma: Thanks. I thought it was a relative path to the version folder. Fixed it.

jongio (commented):
Build script ldflags paths now correctly target `azurelogicappsstandard/internal/version`, matching the Go module name. All my prior findings across both reviews are addressed. No new issues.

jongio:
@wbreza all the feedback from both of us has been addressed. Can you take another look?

jongio:
The `customCodeProject` property name follows the existing camelCase convention for azure.yaml schema properties (`remoteBuild`, `resourceGroup`, `apiVersion`, etc.) - that's correct.

However, the README examples use `logicApp` and `src/logicApp` for the service name and project directory. Every other azure.yaml example in the repo uses lowercase for service names and directories: `web`, `api`, `func`, `app`, `job`, `src/web`, `src/api`, `src/dotnet`, etc.

Consider changing the examples to lowercase for consistency:

```yaml
services:
  logic-app:
    project: ./src/logic-app
    host: function
    language: logicappsstandard
```

@wbreza - can you confirm the convention here? Should service names in documentation examples stay lowercase/kebab-case?

ronaldbosma:
> The `customCodeProject` property name follows the existing camelCase convention for azure.yaml schema properties (`remoteBuild`, `resourceGroup`, `apiVersion`, etc.) - that's correct.
> 
> However, the README examples use `logicApp` and `src/logicApp` for the service name and project directory. Every other azure.yaml example in the repo uses lowercase for service names and directories: `web`, `api`, `func`, `app`, `job`, `src/web`, `src/api`, `src/dotnet`, etc.
> 
> Consider changing the examples to lowercase for consistency:
> 
> ```yaml
> services:
>   logic-app:
>     project: ./src/logic-app
>     host: function
>     language: logicappsstandard
> ```
> 
> @wbreza - can you confirm the convention here? Should service names in documentation examples stay lowercase/kebab-case?

@jongio I've updated the readme to use lowercase in the snippets. I couldn't find a good service related sample in the documentation because they're all one word. But under https://learn.microsoft.com/en-us/azure/developer/azure-developer-cli/layered-provisioning#example-monorepo-with-multiple-services the layers use lowercase as the convention. So, I went with that.

jongio (commented):
Addresses my naming convention feedback. README examples now use lowercase `logicapp` for service names and directories, matching the repo's convention. All findings from my previous reviews are addressed. No new issues.

jongio:
@wbreza - all review findings have been addressed across 4 review cycles (version injection, build script paths, CI scripts, and naming conventions). This looks ready for your final approval.

jongio (commented):
No new issues. The cspell addition is consistent with the lowercase naming convention fix.

wbreza:
Great work @ronaldbosma — the extension is looking solid after 4 review cycles. A few items:

**✅ Naming confirmation:** `logicappsstandard` is the right name. Standard and Consumption are architecturally different products with incompatible deployment models, so the explicit suffix prevents user confusion and future-proofs the namespace. Good call.

**✅ Convention confirmation** (re: @jongio's earlier question): Lowercase/kebab-case for service names in docs is correct. The updates look good.

**🟡 One change needed before approval — host validation:**

This is the first language extension tightly coupled to a specific host type (`function`). Today, if a user misconfigures their `azure.yaml` with `host: appservice` + `language: logicappsstandard`, the extension would silently proceed through packaging and likely fail with a confusing error at deploy time.

I'd like to see a lifecycle event handler registered in `listen.go` that validates the host early in the pipeline. Something like:

```go
func configureListen(host *azdext.ExtensionHost) {
    host.WithFrameworkService("logicappsstandard", func() azdext.FrameworkServiceProvider {
        return project.NewLogicAppsStandardFrameworkServiceProvider()
    })

    // Validate host compatibility early — Logic Apps Standard requires the function host
    host.WithServiceEventHandler(
        "prepackage",
        func(ctx context.Context, event *azdext.ServiceEvent) error {
            if event.ServiceConfig.Host != "function" {
                return fmt.Errorf(
                    "Logic Apps Standard requires 'host: function' in azure.yaml, "+
                    "but found 'host: %s'. Update your service configuration",
                    event.ServiceConfig.Host,
                )
            }
            return nil
        },
        &azdext.ServiceEventOptions{Language: "logicappsstandard"},
    )
}
```

This catches the misconfiguration immediately with a clear, actionable error instead of letting users get deep into the pipeline before failing. Since this is the first extension with a host constraint, it also sets a good pattern for future extensions.

Everything else looks good to go — the packaging logic, path validation, and framework integration are well-implemented. Once the host validation is added, this is ready for approval. 🚀

ronaldbosma:
Hi @wbreza. 

I've made `host: function` mandatory when using `language: logicappsstandard`. I updated the initialize method in the framework service provider, which gives an immediate error when trying to package. The initialize method already has other validation logic to check if the custom code project file exists, so placing the validation here is more inline with the rest of the extension's implementation.

---

I tried this snippet per your suggesting (with a little tweak because `azdext.ServiceEvent` should be `azdext.ServiceEventArgs`, which has a `Service` property):

```go
// Validate host compatibility early — Logic Apps Standard requires the function host
host.WithServiceEventHandler(
	"prepackage",
	func(ctx context.Context, args *azdext.ServiceEventArgs) error {
		if args.Service.Host != "function" {
			return fmt.Errorf(
				"Logic Apps Standard requires 'host: function' in azure.yaml, "+
					"but found 'host: %s'. Update your service configuration",
				args.Service.Host,
			)
		}
		return nil
	},
	&azdext.ServiceEventOptions{Language: "logicappsstandard"},
)
```

The extension builds successfully but I got the error `ERROR: initializing service 'logicapp', rpc error: code = Unavailable desc = transport is closing` when running `azd package` on a sample. Even with `host: function`.

I also tried this simplified version:

```go
host.WithServiceEventHandler("prepackage", func(ctx context.Context, args *azdext.ServiceEventArgs) error {
	return nil
}, nil)
```

But that resulted in this error during packaging: `ERROR: initializing service 'logicapp', channel closed by broker`

Let me know if using the initialize method is not OK and I'll give it another go.

jongio (commented):
Host validation looks good. Placing it in `Initialize()` is the right call - it's consistent with the existing validation pattern in that method and gives a clear, early error. The test coverage is solid.

ronaldbosma:
Hi @jongio and @wbreza, is there anything I can do to help get this PR over the finish line?

ronaldbosma:
Synced my main with new changes from azure-dev repo, rebased my branch and updated to azd v1.25.1 in azure.logicappsstandard extension

kristenwomack:
Hey @ronaldbosma, thanks for getting this together — and for the patience, it's been sitting a bit. A couple of questions before I sign off:

1. **Ownership** — this is a new extension surface for Logic Apps. Is there a Logic Apps team contact we should loop in as a co-owner long-term, or are you signing up to maintain this? Want to make sure it doesn't go stale after merge.
2. **No required-files validation** — totally get the rationale (empty zips deploy fine), but I'm a little worried about the failure mode where someone points at the wrong folder and gets a successful-looking `package` that's actually empty. Worth a soft warning when the folder has no `host.json` or workflows? Not a blocker, just curious what you think.
3. **`host: function` restriction** — good call, makes the failure loud. 👍
4. **`.funcignore`** — borrowing the Functions convention works for me. Worth a line in the README so users coming from pure Logic Apps know to look for it?

ronaldbosma:
Hi @kristenwomack 

1. **Ownership**: I'm willing to maintain this extension, but it might not be a bad idea to loop in someone from the Logic Apps team as well.

2. **No required-files validation**: How can I output a warning for the user to see? I like the idea of adding a warning for an empty folder but I can't get it to work. I only see this as output:  

   image

   I've tried the following statements, but I never see a warning or message show up in my output when packaging a Logic Apps Standard project with the extension.

    ```go
    fmt.Printf("   warning: fmt.printf")
    fmt.Printf("  %s Test Warning fmt.printf\n", color.YellowString("warning:"))
    
    fmt.Println(output.WithWarningFormat("Test warning WithWarningFormat"))
    
    log.Printf("warning: log.printf")
    ```
    
    I also tried these statements:
    
    ```go
    logger := azdext.NewLogger("azure.logicappsstandard")
    logger.Debug("Debug")
    logger.Info("Info")
    logger.Warn("Warn")
    logger.Error("Error")
    ```

   with `azdext.SetupLogging(azdext.LoggerOptions{Debug: true})` included in the `main.go` and `root.go` files, but I don't see the output. Not even when running `azd package` with the `--debug` flag.

   I've tried all these commands in both the `Initialize` and `Package` functions of the extension without putting an if statement around it. So, they should always execute. See [this PR](https://github.com/ronaldbosma/azure-dev/pull/1/changes) for the specific changes.

3. **`host: function restriction`**: -

4. **`.funcignore`**: When creating a Logic App Standard project through the VS Code extension, the `.funcignore` is included by default. It's mentioned [here](https://learn.microsoft.com/en-us/azure/logic-apps/create-standard-workflows-visual-studio-code#standard-logic-app-project-structure) in the official docs. It's not something I made up for this extension, but a default way of working for Logic Apps Standard. 

   The extension's readme mentions _"the exclusions in `.funcignore` are respected and only the relevant files are packaged"_. If that's not enough, let me know.

kristenwomack:
Hey @ronaldbosma, thanks for the thorough reply.

**1. Ownership** — sounds good. Let me ping the Logic Apps folks.

**2. Empty-folder warning** — It might be that the extension runs as a separate process that doesn't write back to azd's console directly. I think you need to progress callback that's passed into `Package` (and `Build`/`Restore`). Take a look at how `microsoft.azd.demo` uses it:

```go
// cli/azd/extensions/microsoft.azd.demo/internal/project/framework_service_demo.go
progress("Creating Rust deployment package")
```

So in your `Package` method something like:

```go
if !hasHostJsonOrWorkflows(packagePath) {
    progress(fmt.Sprintf("warning: %s contains no host.json or workflow files; package may be empty", packagePath))
}
```

That'll route through azd's normal task output. If you want a louder warning that sticks after the spinner clears, that's a gap in the SDK we should track separately.

**4. `.funcignore`** — fair, the README mention is enough. No change needed.

Let me know what you find in retrying the example and I'll check on my side with the Logic Apps team.

ronaldbosma:
Hi @kristenwomack,

I've added a progress call in the `Package` method. I'm using `azd package` to test the warning but zipping an empty folder seems to be to fast. 

If I run `azd package` with the `--debug` flag, the progress message doesn't even end up in the logging (let alone show it in the console). Here's the end of an execution and there's no warning:

```
...

2026/06/04 15:25:23 message_broker.go:373: [azure.logicappsstandard] [RequestId=7f321ef0-b1e4-4f38-81f0-dc070dff4519] Progress message, MessageType=*azdext.FrameworkServiceProgressMessage
  |      =| Packaging service logicAppSample5 (Creating Logic Apps Standard deployment package)2026/06/04 15:25:23 message_broker.go:497: [azure.logicappsstandard] Dispatcher received message: RequestId=7f321ef0-b1e4-4f38-81f0-dc070dff4519, MessageType=*azdext.FrameworkServicePackageResponse
2026/06/04 15:25:23 message_broker.go:513: [azure.logicappsstandard] Dispatching message to channel for RequestId=7f321ef0-b1e4-4f38-81f0-dc070dff4519, MessageType=*azdext.FrameworkServicePackageResponse
2026/06/04 15:25:23 message_broker.go:516: [azure.logicappsstandard] Message dispatched successfully to RequestId=7f321ef0-b1e4-4f38-81f0-dc070dff4519, MessageType=*azdext.FrameworkServicePackageResponse
2026/06/04 15:25:23 message_broker.go:369: [azure.logicappsstandard] [RequestId=7f321ef0-b1e4-4f38-81f0-dc070dff4519] Received on channel, MessageType=*azdext.FrameworkServicePackageResponse
2026/06/04 15:25:23 message_broker.go:385: [azure.logicappsstandard] [RequestId=7f321ef0-b1e4-4f38-81f0-dc070dff4519] Received final response, MessageType=*azdext.FrameworkServicePackageResponse
2026/06/04 15:25:23 message_broker.go:319: [azure.logicappsstandard] [RequestId=7f321ef0-b1e4-4f38-81f0-dc070dff4519] Cleaning up channel
2026/06/04 15:25:23 message_broker.go:477: [azure.logicappsstandard] Received progress message: RequestId=7f321ef0-b1e4-4f38-81f0-dc070dff4519, MessageType=*azdext.FrameworkServiceProgressMessage
2026/06/04 15:25:23 message_broker.go:487: [azure.logicappsstandard] WARNING: No channel found for progress message RequestId=7f321ef0-b1e4-4f38-81f0-dc070dff4519, MessageType=*azdext.FrameworkServiceProgressMessage
  (✓) Done: Packaging service logicAppSample5
  - Build Output: /workspaces/azure-dev/cli/azd/extensions/azure.logicappsstandard/bin/samples/sample-5/src/logicApp
  - Package Output: /tmp/sample-5.zip

2026/06/04 15:25:23 server.go:153: azd gRPC Server stopped
```

Packaging seems to be completed before the progress message is processed.


I added a sleep statement after the progress call:

```go
progress(fmt.Sprintf("warning: Logic App package may be empty. No host.json or workflow files found in %s", packagePath))
time.Sleep(5 * time.Second)
```

Then I can see the message in the console:  

image

And it also ends up in the logging when using the `--debug` flag:

```
...

2026/06/04 15:23:36 message_broker.go:373: [azure.logicappsstandard] [RequestId=b18ebc50-7421-494f-b50d-746df7cccb3d] Progress message, MessageType=*azdext.FrameworkServiceProgressMessage
  |      =| Packaging service logicAppSample5 (Creating Logic Apps Standard deployment package)2026/06/04 15:23:36 message_broker.go:477: [azure.logicappsstandard] Received progress message: RequestId=b18ebc50-7421-494f-b50d-746df7cccb3d, MessageType=*azdext.FrameworkServiceProgressMessage
2026/06/04 15:23:36 message_broker.go:479: [azure.logicappsstandard] Dispatching progress message to channel for RequestId=b18ebc50-7421-494f-b50d-746df7cccb3d, MessageType=*azdext.FrameworkServiceProgressMessage
2026/06/04 15:23:36 message_broker.go:369: [azure.logicappsstandard] [RequestId=b18ebc50-7421-494f-b50d-746df7cccb3d] Received on channel, MessageType=*azdext.FrameworkServiceProgressMessage
2026/06/04 15:23:36 message_broker.go:373: [azure.logicappsstandard] [RequestId=b18ebc50-7421-494f-b50d-746df7cccb3d] Progress message, MessageType=*azdext.FrameworkServiceProgressMessage
  |  =====| Packaging service logicAppSample5 (warning: Logic App package may be empty. No host.json or workflow files found in /workspaces/azure-dev/cli/azd/exten...2026/06/04 15:23:41 message_broker.go:497: [azure.logicappsstandard] Dispatcher received message: RequestId=b18ebc50-7421-494f-b50d-746df7cccb3d, MessageType=*azdext.FrameworkServicePackageResponse
2026/06/04 15:23:41 message_broker.go:513: [azure.logicappsstandard] Dispatching message to channel for RequestId=b18ebc50-7421-494f-b50d-746df7cccb3d, MessageType=*azdext.FrameworkServicePackageResponse
2026/06/04 15:23:41 message_broker.go:516: [azure.logicappsstandard] Message dispatched successfully to RequestId=b18ebc50-7421-494f-b50d-746df7cccb3d, MessageType=*azdext.FrameworkServicePackageResponse
2026/06/04 15:23:41 message_broker.go:369: [azure.logicappsstandard] [RequestId=b18ebc50-7421-494f-b50d-746df7cccb3d] Received on channel, MessageType=*azdext.FrameworkServicePackageResponse
2026/06/04 15:23:41 message_broker.go:385: [azure.logicappsstandard] [RequestId=b18ebc50-7421-494f-b50d-746df7cccb3d] Received final response, MessageType=*azdext.FrameworkServicePackageResponse
2026/06/04 15:23:41 message_broker.go:319: [azure.logicappsstandard] [RequestId=b18ebc50-7421-494f-b50d-746df7cccb3d] Cleaning up channel
  (✓) Done: Packaging service logicAppSample5
  - Build Output: /workspaces/azure-dev/cli/azd/extensions/azure.logicappsstandard/bin/samples/sample-5/src/logicApp
  - Package Output: /tmp/sample-5.zip

2026/06/04 15:23:41 server.go:153: azd gRPC Server stopped
```

I see a couple of options:
1. Skip the warning logic and wait until azd supports creating a warning that sticks after the spinner clears as you suggested.
2. Add the warning logic with a sleep of a couple of seconds.

I think I prefer option 1 because the user might still miss the warning when they're not looking at the screen.

Let me know what you think

jongio (commented):
Re-reviewed after the June 22 force-push that incorporated all prior feedback. The extension looks solid:

- Clean architecture with proper separation (cmd/project/version packages)
- Path traversal protection via `resolvePathWithinBase` prevents escaping the project directory
- Host validation ensures only `function` host is accepted (Logic Apps runs on Azure Functions runtime)
- Comprehensive tests covering nil configs, missing files, directory traversal attempts, dotnet failures, and cross-platform stub execution
- CI/release pipelines follow established repo patterns

All 6 rounds of prior review feedback have been addressed. No blocking issues found. LGTM from a code quality perspective.

cli/azd/extensions/azure.logicappsstandard/CHANGELOG.md
```diff
@@ -0,0 +1,5 @@
+# Release History
+
+## 0.0.1
+
+- Initial release with support for packaging Logic App Standard projects via `language: logicappsstandard` in `azure.yaml`, and optional custom code projects using `customCodeProject` property.
```
jongio: Missing trailing newline. Add one to avoid `\ No newline at end of file` markers in future diffs.

jongio (approved):
Reviewed all 20 files. Extension follows established patterns (azdext helpers, pkg separation, CI/release templates). Path traversal protection in `resolvePathWithinBase` is correct. Tests cover nil configs, missing files, traversal attempts, and cross-platform dotnet stubs.

One nit: `CHANGELOG.md` is missing a trailing newline (POSIX text file convention). Will cause minor diff noise when the next entry is appended. Non-blocking.

ronaldbosma:
From my perspective, this PR is good to go and can be merged. If there's anything blocking it, please let me know.

My preference is to have it merged into the azure-dev repo so the extension can be published through the official extension registry, allowing users to install it without having to configure a custom extension source.

If merging it isn't an option, that's completely fine. Just let me know and I'll move it to a separate repository instead.

jongio (approved):
Trailing newline fix addresses my last nit. No new issues since my previous approval.

cli/azd/extensions/azure.logicappsstandard/internal/project/framework_service_logicappsstandard.go
```diff
@@ -0,0 +1,237 @@
...
+
+// Requirements returns the framework requirements (whether restore/build are needed)
+func (p *LogicAppsStandardFrameworkServiceProvider) Requirements() (*azdext.FrameworkRequirements, error) {
+	hasCustomCodeProject := p.serviceConfig != nil && hasCustomCodeProjectConfigured(p.serviceConfig)
+	return &azdext.FrameworkRequirements{
+		Package: &azdext.FrameworkPackageRequirements{
+			RequireRestore: hasCustomCodeProject,
+			RequireBuild:   hasCustomCodeProject,
+		},
+	}, nil
```
RickWinter, line 85: `Requirements()` is treated as static per language: the request carries no service config, and `FrameworkServiceManager.onRequirements` calls it on an arbitrary initialized instance. With two `logicappsstandard` services, if that instance has no `customCodeProject`, azd skips restore and build for the service that does, so its DLLs are absent from the deployment package. Let's make both requirements unconditional and keep the existing no-op guards in `Restore` and `Build`. The stored `serviceConfig` can then be removed.

```suggestion
func (p *LogicAppsStandardFrameworkServiceProvider) Requirements() (*azdext.FrameworkRequirements, error) {
	return &azdext.FrameworkRequirements{
		Package: &azdext.FrameworkPackageRequirements{
			RequireRestore: true,
			RequireBuild:   true,
		},
	}, nil
}
```

azd-code-reviewer

jongio (commented):
I verified the Requirements() concern against core, and it is a real bug for multi-service setups.

`onRequirements` in `pkg/azdext/framework_service_manager.go` calls `componentManager.GetAnyInstance()` and invokes `Requirements()` with no ServiceConfig (the request is explicitly empty because "requirements are static for a framework"). So when a project defines two `logicappsstandard` services, the RequireRestore/RequireBuild flags come from whichever instance `GetAnyInstance` happens to return. If that instance was initialized for a service without `customCodeProject`, both flags are false.

`pkg/project/service_manager.go` (lines 376-387) gates Restore and Build on those flags, so azd then skips restore and build for every service in the language, including the one that does have custom code. That service gets packaged without its compiled DLLs.

The suggested fix is correct: return `RequireRestore: true` and `RequireBuild: true` unconditionally and rely on the existing `hasCustomCodeProjectConfigured` guards in `Restore` and `Build`, which already no-op per service. `p.serviceConfig` then has no remaining reader and can be dropped.

cli/azd/extensions/azure.logicappsstandard/internal/project/framework_service_logicappsstandard.go
ronaldbosma: Nice catch, thanks!

jongio (approved):
Re-reviewed the incremental change since my last review. The `Requirements()` fix is correct. It now returns `RequireRestore` and `RequireBuild` unconditionally, and the per-service `hasCustomCodeProjectConfigured` guards in `Restore` and `Build` keep services without custom code as no-ops. The `serviceConfig` field is removed with no remaining readers, and the tests are updated to match. This resolves the multi-service packaging issue I flagged earlier. No new issues.

cli/azd/extensions/azure.logicappsstandard/internal/project/framework_service_logicappsstandard.go
```diff
@@ -0,0 +1,234 @@
...
+	ctx context.Context,
+	serviceConfig *azdext.ServiceConfig,
+) ([]*azdext.ExternalTool, error) {
+	if hasCustomCodeProjectConfigured(serviceConfig) {
+		return []*azdext.ExternalTool{
+			{
+				Name:       "dotnet",
+				InstallUrl: "https://dotnet.microsoft.com/download",
+			},
+		}, nil
```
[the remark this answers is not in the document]
ronaldbosma: Only requiring `dotnet` is not an issue as custom code projects created with the official Logic Apps VS Code extension are Microsoft.NET.Sdk projects.

ronaldbosma:
I've updated the dependencies of the extension to the latest versions and included validation logic to check if the `customCodeProject` property has a valid string value. If not, an error is returned instead of silently skipping the dotnet restore and build steps. I also changed the version to 0.1.0 as this better reflects the status of this extension.

jongio (approved):
The customCodeProject type validation and error propagation through Initialize are correct, and the new tests cover the non-string path. Version bump to 0.1.0 is consistent across CHANGELOG, extension.yaml, and version.txt.

eng/pipelines/release-ext-azure-logicappsstandard.yml
```diff
@@ -0,0 +1,33 @@
...
+      - /eng/pipelines/templates/jobs/cross-build-azd-extension.yml
+      - /eng/pipelines/templates/variables/image.yml
+
+pr:
+  paths:
+    include:
+      - cli/azd/extensions/azure.logicappsstandard
+      - eng/pipelines/release-ext-azure-logicappsstandard.yml
+      - eng/pipelines/release-azd-extension.yml
+      - eng/pipelines/templates/steps/publish-cli.yml
```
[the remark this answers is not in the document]
ronaldbosma: I've created #9216 to see what the actual includes should be for all extension pipelines.

eng/pipelines/release-ext-azure-logicappsstandard.yml
```diff
@@ -0,0 +1,33 @@
+# Continuous deployment trigger
+trigger:
+  branches:
+    include:
+      - main
+  paths:
+    include:
+      - go.mod
+      - cli/azd/extensions/azure.logicappsstandard
+      - eng/pipelines/release-azd-extension.yml
```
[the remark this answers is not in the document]
ronaldbosma: I've created #9216 to see what the actual includes should be for all extension pipelines.

jongio (approved):

cli/azd/extensions/azure.logicappsstandard/internal/project/framework_service_logicappsstandard.go
```diff
@@ -0,0 +1,254 @@
...
+
+	packageParts := []string{serviceConfig.RelativePath}
+
+	// If an output path is specified, append it to the package path.
+	// This allows for subdirectories like "Workflows" to be the root of the zip.
+	if serviceConfig.OutputPath != "" {
+		packageParts = append(packageParts, serviceConfig.OutputPath)
+	}
+
+	packagePath, err := resolvePathWithinBase(projectDir, packageParts...)
```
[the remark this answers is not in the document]
ronaldbosma: Current implementation is preferred

cli/azd/extensions/azure.logicappsstandard/extension.yaml
```diff
@@ -0,0 +1,11 @@
+# yaml-language-server: $schema=../extension.schema.json
+id: azure.logicappsstandard
+namespace: logicappsstandard
+displayName: Azure Logic Apps Standard
+description: Extension for packaging Logic Apps Standard projects, including support for custom code projects
+usage: Use logicappsstandard as language in your service configuration. If your Logic App has a custom code project, configure that path to the .csproj file using the customCodeProject property.
+# NOTE: Make sure version.txt is in sync with this version.
+version: 0.1.0
+language: go
+capabilities:
```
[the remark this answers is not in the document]
ronaldbosma: Fixed

cli/azd/extensions/azure.logicappsstandard/internal/project/framework_service_logicappsstandard.go
```diff
@@ -0,0 +1,254 @@
...
+func (p *LogicAppsStandardFrameworkServiceProvider) RequiredExternalTools(
+	ctx context.Context,
+	serviceConfig *azdext.ServiceConfig,
+) ([]*azdext.ExternalTool, error) {
+	hasCustomCode, _ := hasCustomCodeProjectConfigured(serviceConfig)
+	if hasCustomCode {
+		return []*azdext.ExternalTool{
+			{
+				Name:       "dotnet",
+				InstallUrl: "https://dotnet.microsoft.com/download",
```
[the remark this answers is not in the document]
ronaldbosma: Not an issue. We rely on azd for installation of dotnet.

jongio (approved):
Verified the remaining open threads against the code. The path resolution is guarded by resolvePathWithinBase, the host is validated in Initialize, and custom-code restore/build/package are gated correctly. requiredAzdVersion is now set. Looks good to me.

jongio (approved):
Verified the remaining open threads against the code. Path resolution is guarded by resolvePathWithinBase, the host is validated in Initialize, and custom-code restore, build, and package are gated correctly. The required azd version floor is now set. Looks good to me.

eng/pipelines/release-ext-azure-logicappsstandard.yml
```diff
@@ -0,0 +1,47 @@
...
+extends:
+  template: /eng/pipelines/templates/stages/1es-redirect.yml
+  parameters:
+    stages:
+      - template: /eng/pipelines/templates/stages/release-azd-extension.yml
+        parameters: 
+          AzdExtensionId: azure.logicappsstandard
+          SanitizedExtensionId: azure-logicappsstandard
+          AzdExtensionDirectory: cli/azd/extensions/azure.logicappsstandard
+          PublishToRegistry: ${{ parameters.PublishToRegistry }}
```
jongio, line 45: `PublishToRegistry` is passed to the `release-azd-extension.yml` template, but that template doesn't declare a `PublishToRegistry` parameter. It only accepts `AzdExtensionId`, `AzdExtensionDirectory`, `SanitizedExtensionId`, and `SkipTests`. Azure Pipelines rejects undeclared parameters at compile time (`Unexpected parameter 'PublishToRegistry'`), so this would fail the build once merged.

I grepped all of `eng/` on latest main and `PublishToRegistry` doesn't exist anywhere, and the `release-ext-microsoft-azd-demo.yml` this commit is meant to match neither defines nor passes it. If the shared templates are supposed to consume this value, that plumbing needs to land first (or in this PR). Otherwise, dropping the `parameters:` block and this line keeps it in sync with the demo pipeline.

jongio (commented):
One concern on the new pipeline parameter added in `release-ext-azure-logicappsstandard.yml`. The `requiredAzdVersion` addition and the `go.mod` trigger path both look correct. Details inline.

eng/pipelines/release-ext-azure-logicappsstandard.yml
ronaldbosma: Hi @jongio. I'm seeing the parameter `PublishToRegistry` in the template here: https://github.com/Azure/azure-dev/blob/main/eng/pipelines/templates/stages/release-azd-extension.yml#L11

jongio (approved):
Re-reviewed the two new commits since my last approval: the `requiredAzdVersion: ">= 1.28.0"` field (valid, released version) and the release pipeline alignment (go.mod trigger + PublishToRegistry parameter passthrough matching the other extension pipelines). Both check out.

eng/pipelines/release-ext-azure-logicappsstandard.yml
jongio: Confirmed the stage template declares `PublishToRegistry` (stable/dev/nightly), and this wiring matches how `microsoft.azd.demo` extends it. My earlier comment was wrong. This is fine as-is.

cli/azd/extensions/azure.logicappsstandard/internal/project/framework_service_logicappsstandard.go
```diff
@@ -0,0 +1,256 @@
...
+
+func hasCustomCodeProjectConfigured(serviceConfig *azdext.ServiceConfig) (bool, error) {
+	customCodeProject, err := getAdditionalProperty(serviceConfig, "customCodeProject")
+	if err != nil {
+		return false, err
+	}
+	return customCodeProject != "", nil
+}
+
+// runDotNet executes the dotnet CLI with the given arguments, forwarding output to stdout/stderr.
```
jongio: The doc comment still says this forwards output to `stdout`/`stderr`, but the function now captures combined output and only surfaces it in the returned error. Consider updating the comment to match, e.g. "returning any output as part of the error on failure."

jongio (approved):
Re-reviewed the incremental changes since my previous approval: the azd dependency bump to 1.28.1, the `requiredAzdVersion` constraint in extension.yaml, the release pipeline alignment with the shared stage template, and the `runDotNet` fix to surface dotnet output on failure. The `PublishToRegistry` parameter is valid on the shared template. Nothing new to flag beyond one doc-comment note inline.

cli/azd/extensions/azure.logicappsstandard/CHANGELOG.md
```diff
@@ -0,0 +1,5 @@
+# Release History
+
+## 0.1.0
```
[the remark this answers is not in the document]
ronaldbosma: Shouldn't be an issue as it hasn't been merged yet

ronaldbosma:
- Fixed the pipeline trigger paths so it matches the changes made to the other extension pipelines in #9241
- A build error of the custom code project resulted in 'exit status 1' without showing the actual error to the user. By capturing the combined output when running dotnet, the build errors are returned to the user.
- Merged latest changes from main
- Updated the azd version to 1.28.1


_Off-topic: Copilot reviews seem to have improved since the start of this PR and are now finding issues they previously missed. However, after every push, they also report new issues (some valid, others not) in code that hasn't changed. Receiving a single review with all identified issues would save time and reduce the number of review iterations needed._

jongio (approved):
Re-reviewed the delta since my last approval.

`runDotNet` now uses `CombinedOutput` and wraps stdout/stderr into the returned error on failure, which fixes the opaque "exit status 1". `Restore` and `Build` already add context to that error and report progress through the `ProgressReporter`, so dropping the raw `os.Stdout` streaming doesn't lose user feedback. `os` is still referenced by `os.Stat` and `fmt` is imported, so it compiles.

Pipeline trigger cleanup (dropping `go.mod`, pointing at `templates/stages/release-azd-extension.yml`) lines up with #9241, and the `PublishToRegistry` parameter mirrors the demo extension pipeline. The azd 1.28.1 bump is consistent across extension.yaml and go.mod. No new issues.

cli/azd/extensions/azure.logicappsstandard/internal/project/framework_service_logicappsstandard_test.go
```diff
@@ -0,0 +1,444 @@
...
+// and execute it via PATHEXT; on Unix it creates a dotnet shell script.
+// Both stubs append their arguments to the file named by DOTNET_ARGS_LOG.
+func createFakeDotnetStub(t *testing.T, fakeBinDir string) {
+	t.Helper()
+	stubName := "dotnet"
+	stubContent := "#!/bin/sh\nprintf '%s\n' \"$*\" >> \"$DOTNET_ARGS_LOG\"\n"
+	stubMode := os.FileMode(0o755)
+	if runtime.GOOS == "windows" {
+		stubName = "dotnet.cmd"
+		stubContent = "@echo off\r\necho %* >> %DOTNET_ARGS_LOG%\r\n"
```
jongio: Fails the suite on Windows. `echo %* >> %DOTNET_ARGS_LOG%` puts the space before the redirect operator into what cmd.exe echoes, so every logged invocation gets a trailing space and the exact-match assertion at line 305 fails on the extra trailing space. Moving the redirect to the front fixes it (quoting the redirect target as well would also guard against spaces in the temp path). Verified `go test ./...` passes on Windows with:

```suggestion
		stubContent = "@echo off\r\n>>%DOTNET_ARGS_LOG% echo %*\r\n"
```

jongio (commented):
Re-reviewed the commits since my last pass. The runDotNet change and the requiredAzdVersion / go.mod bumps are fine.

Worth fixing: the extension's Go tests fail on Windows, and the PR's GitHub checks don't run them so it isn't surfaced here. Running `go test ./...` in cli/azd/extensions/azure.logicappsstandard, TestRestoreAndBuildInvokeDotNetForCustomCodeProject fails on an exact-match assertion. Root cause and a one-line fix inline.

jongio (approved):
Re-reviewed the delta since my last approval. azd bumped to 1.28.1, `runDotNet` now returns dotnet's combined output on failure so build errors surface instead of a bare "exit status 1", and release-ext-azure-logicappsstandard.yml now matches the sibling extension release pipelines. No new concerns. Approving.

<sub>🤖 agent jongio</sub>

cli/azd/extensions/azure.logicappsstandard/internal/project/framework_service_logicappsstandard_test.go
jongio: Confirmed this on Windows against the current HEAD (`d47a7a4`):

```
--- FAIL: TestRestoreAndBuildInvokeDotNetForCustomCodeProject (0.13s)
    Error:      Not equal:
    expected: "restore ...\src\logicApp\Functions\Functions.csproj"
    actual  : "restore ...\src\logicApp\Functions\Functions.csproj "
```

cmd.exe keeps the space sitting between `%*` and `>>` in what it echoes, so every logged invocation picks up a trailing space. `strings.TrimSpace` on line 301 only trims the ends of the whole file contents, so the last line comes out clean while `logLines[0]` keeps its trailing space and the exact-match assert on line 305 fails.

This isn't only a local annoyance. `release-azd-extension.yml` puts a `Windows` job in the `BuildMatrix`, and `build-azd-extension.yml` runs `ci-test.ps1` (`go test ./...`) whenever `SkipTests` is false, which is the default. So the release pipeline fails on the Windows agent once this lands on main. The PR check here only runs `lint-go.yml`, which is why it's green.

Moving the redirect ahead of `echo` fixes it, and quoting the path keeps it working if the temp dir ever contains a space:

```suggestion
		stubContent = "@echo off\r\n>>\"%DOTNET_ARGS_LOG%\" echo %*\r\n"
```

I applied that locally and `go test ./...` passes on Windows. The `/bin/sh` stub doesn't need changing since `printf '%s\n' "$*"` already writes the args clean.

jongio (commented):
One thing to fix before this merges: `TestRestoreAndBuildInvokeDotNetForCustomCodeProject` fails on Windows at the current HEAD. I reproduced it locally and put the failure output, the root cause, and a one-line fix on the existing thread in `framework_service_logicappsstandard_test.go`.

Worth flagging the blast radius: the release pipeline runs `ci-test.ps1` on a Windows agent (`BuildMatrix.Windows` in `release-azd-extension.yml`, with `SkipTests` defaulting to false), so this breaks the CD build after merge. The PR check on this repo only runs `lint-go.yml`, so nothing here catches it.

Smaller thing: `extension.yaml` and `cspell.yaml` are missing a trailing newline. 18 of the 20 extensions in `cli/azd/extensions` end both files with one. `cspell.yaml` also has trailing whitespace on its last line.

Nothing else outstanding from my side. My earlier notes on the `version` const, the `build.sh` binary name, the `runDotNet` doc comment, and the `CHANGELOG.md` newline are all addressed. Once the stub fix is in I'll re-approve.

cli/azd/extensions/azure.logicappsstandard/internal/project/framework_service_logicappsstandard_test.go
jongio: One more data point on why this one matters beyond local dev runs: the ADO release pipeline runs this suite on a Windows agent. `eng/pipelines/templates/jobs/build-azd-extension.yml` invokes `<extension-dir>/ci-test.ps1`, and `release-azd-extension.yml` has a Windows leg in its build matrix, so `BuildExtension_Windows` will fail once this merges.

The reason nothing catches it here is that `lint-ext-azure-logicappsstandard.yml` only calls `lint-go.yml`. No GitHub Actions workflow runs `go test` for this extension, so PR CI stays green while the ADO leg is the thing that breaks.

Putting the redirect first fixes it, and quoting the target keeps it working if the temp path ever has a space in it:

```suggestion
		stubContent = "@echo off\r\n>>\"%DOTNET_ARGS_LOG%\" echo %*\r\n"
```

I applied exactly that in a local worktree at `d47a7a4` and `go test ./... -count=1` went from `FAIL azurelogicappsstandard/internal/project` to `ok azurelogicappsstandard/internal/project`.

cli/azd/extensions/azure.logicappsstandard/extension.yaml
```diff
@@ -0,0 +1,12 @@
...
+namespace: logicappsstandard
+displayName: Azure Logic Apps Standard
+description: Extension for packaging Logic Apps Standard projects, including support for custom code projects
+usage: Use logicappsstandard as language in your service configuration. If your Logic App has a custom code project, configure that path to the .csproj file using the customCodeProject property.
+# NOTE: Make sure version.txt is in sync with this version.
+version: 0.1.0
+requiredAzdVersion: ">= 1.28.1"
+language: go
+capabilities:
+    - framework-service-provider
```
jongio: No trailing newline on this file. Same for `cspell.yaml` and `ci-test.ps1`. The sibling extensions (`azure.appservice`, `azure.ai.projects`) end all three with a newline, so this just keeps future diffs free of `\ No newline at end of file`. Minor.

jongio (commented):
One issue left over from my earlier pass, plus a trivial nit.

`internal/project/framework_service_logicappsstandard_test.go:415` still fails `go test` on Windows at `d47a7a4`. I've put the details and a one-line fix in the existing thread. It isn't only a local dev annoyance: the ADO release pipeline runs `ci-test.ps1` on a Windows agent, so the Windows build job breaks once this merges. PR CI stays green because the GitHub workflow for this extension only runs lint, never `go test`.

cli/azd/extensions/azure.logicappsstandard/internal/project/framework_service_logicappsstandard_test.go
ronaldbosma: Fixed with 8d663d6c35e89eefde21621731221812bef5f4c7 and 9db773a2f9d4a61bb4acfc4bd55dca4fc5439647

cli/azd/extensions/azure.logicappsstandard/extension.yaml
ronaldbosma: Fixed with 82e97bbf4639053b83dacf358fc0fe4cb5fdec53

cli/azd/extensions/azure.logicappsstandard/internal/project/framework_service_logicappsstandard_test.go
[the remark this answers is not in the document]
ronaldbosma: Fixed with 8d663d6c35e89eefde21621731221812bef5f4c7 and 9db773a2f9d4a61bb4acfc4bd55dca4fc5439647

jongio (approved):
Re-checked the two things I flagged last round, against `82e97bb`:

- The `dotnet.cmd` stub now puts the redirect ahead of `echo` and quotes the log path, so the trailing space is gone and a temp path with spaces won't break it. `go test ./...` in `cli/azd/extensions/azure.logicappsstandard` passes on Windows for me (`ok azurelogicappsstandard/internal/project 45.5s`), which clears the `BuildExtension_Windows` leg in the ADO release pipeline.
- Trailing newlines are in place on `extension.yaml`, `cspell.yaml`, `ci-test.ps1`, `ci-build.ps1`, and `CHANGELOG.md`.

`gofmt -l` and `go vet ./...` are clean on the extension. The rest of the delta since my last pass is merges from main. Nothing else from me.

jongio (approved):

jongio (approved):
Re-verified at 82e97bb on Windows: `go build`, `go vet`, `golangci-lint run ./...` and `go test ./... -count=1` all pass, including `TestRestoreAndBuildInvokeDotNetForCustomCodeProject`. That was the test failing on the trailing-space assertion, and since the ADO release pipeline runs this suite on a Windows agent via `ci-test.ps1`, it was the blocking item.

Also confirmed: trailing newlines on every file, `go.mod` on 1.26.4 matching `cli/azd/go.mod`, direct dependencies limited to the four actually imported, and `Initialize` gating `Restore`/`Build` so the discarded `hasCustomCodeProjectConfigured` errors there aren't reachable.

The lint workflow and release pipeline match the `azure.appservice` layout, and all referenced template paths resolve, so the earlier note about a missing include path doesn't apply.

jongio (approved):

ronaldbosma:
> Hi @@ronaldbosma. Thank you for your interest in helping to improve the Azure Developer CLI experience and for your contribution. We've noticed that there hasn't been recent engagement on this pull request. If this is still an active work stream, please let us know by pushing some changes or leaving a comment. Otherwise, we'll close this out in 7 days.

This PR can be merged in my opinion.

cli/azd/extensions/azure.logicappsstandard/extension.yaml
jongio, line 12: `root.go:27` registers `azdext.NewMetadataCommand`, but `metadata` isn't in this capabilities list.

azd only fetches and caches extension metadata when the manifest declares that capability. See the `HasCapability(MetadataCapability)` guard around `fetchAndCacheMetadata` in `cli/azd/pkg/extensions/manager.go`. As written, the metadata command is registered but azd never calls it, so no `metadata.json` gets cached for this extension.

Every other extension in the repo that registers `NewMetadataCommand` also declares the capability (15 of 15). This is the only one that doesn't.

```suggestion
    - framework-service-provider
    - metadata
```

jongio:
Re-verified at `82e97bb` on Windows: `go build`, `go vet`, `go test ./... -count=1`, `gofmt -l`, `go fix -diff` and `golangci-lint` v2.12.2 all pass clean. Merging current `main` into the branch is conflict-free, and the `cli/azd/pkg/extensions` provider-verification tests still pass against the merged tree.

One thing worth fixing, plus a note on why the checks look stalled.

`extension.yaml` declares only `framework-service-provider`, but `root.go` registers a metadata command. azd gates metadata fetching on that capability, so the command ships but never gets invoked. I left the details and a suggested fix inline on `extension.yaml`.

On merging: what's holding this up isn't the code. Every Actions workflow on this fork PR is sitting at `action_required`, including `ext-azure-logicappsstandard-ci`, `cspell-ext`, `Actionlint`, `validate-go-version` and `verify-ext-providers`. None of them have actually run at `82e97bb`, which is why the status rollup never completes. A maintainer needs to approve the workflow runs for this fork before the checks can report. I ran the equivalent gates locally and they all pass, so approving the run should come back green.

The other outstanding item is the new `eng/pipelines/release-ext-azure-logicappsstandard.yml`, which needs to be onboarded on the release side before it does anything.

jongio (commented):
Re-reviewed the delta since my last pass: one commit adding `metadata` to the capabilities list in `extension.yaml`.

That closes the finding I left on that file. `root.go` registers `azdext.NewMetadataCommand`, and azd only invokes the metadata command when the manifest declares the capability, via the `HasCapability(MetadataCapability)` guard around `fetchAndCacheMetadata` in `cli/azd/pkg/extensions/manager.go`. The value matches `MetadataCapability` in `pkg/extensions/registry.go`, is listed in `ValidCapabilities` in `validate_registry.go`, and is a valid enum entry in `cli/azd/extensions/extension.schema.json`, so registry validation will accept it.

No new findings in this delta. The one unresolved thread left on the PR is the CHANGELOG trailing newline comment, which is outdated and was already fixed.

ronaldbosma:
PR needs approval from maintainer to proceed.

ronaldbosma:
PR needs approval from maintainer to proceed.

ronaldbosma:
PR needs approval from maintainer to proceed.

ronaldbosma:
Hi @RickWinter and @JeffreyCA.  Is there any chance this PR could get merged soon? It’s been ready to merge for quite some time now.

ronaldbosma:
Hi @kristenwomack. Is there anything you can do to get this PR merged?

vhvb1989:
### azure.logicappsstandard PR build

> [!NOTE]
> This is an unsigned development build. Install it only if you trust this PR.

**Install the extension:**

```sh
azd ext install "https://azuresdkartifacts.z5.web.core.windows.net/azd/extensions/pr/10058/azure-logicappsstandard.zip"
```

- **Version:** `0.1.0-pr.10058.6843929`
- **Merge commit:** f06838b5d38aae5c6dd4eebd5d900c1cca27f35a

vhvb1989:
Hey @ronaldbosma , I was looking at this extension. I installed it and I was able to test `azd package` - it creates a zip file and recognize the language you addded in the extension.
But I am trying to create a full e2e scenario where I can run `azd up` to create Infrastructure in bicep and then deploy the zip file and look at the service. Do you have some e2e as such that I can use?

We have also an ongoing conversation with the Logic Apps team about this. Thank you very much for your patient.

ronaldbosma:
See https://github.com/Azure/azure-dev/pull/8005#issuecomment-5731208813 for test template.

~~@vhvb1989 I've created the branch [use-logicappsstandard-azd-extension](https://github.com/ronaldbosma/azure-integration-services-quickstart/tree/use-logicappsstandard-azd-extension) on my [Azure Integration Services Quickstart](https://github.com/ronaldbosma/azure-integration-services-quickstart) template where the azd extension is used. You can use the following commands to deploy it:~~

```pwsh
azd init --template ronaldbosma/azure-integration-services-quickstart --branch use-logicappsstandard-azd-extension

# Configure included resources
azd env set INCLUDE_API_MANAGEMENT true
azd env set INCLUDE_LOGIC_APP true
azd env set INCLUDE_SERVICE_BUS true
azd env set INCLUDE_APPLICATION_INFRA_RESOURCES true

# Configure excluded resources
azd env set INCLUDE_EVENT_HUBS_NAMESPACE false
azd env set INCLUDE_FUNCTION_APP false

azd provision
azd deploy logicApp
```

~~Once deployed, you can use the instructions on https://github.com/ronaldbosma/azure-integration-services-quickstart/blob/main/demos/demo-sample-application.md#test-the-sample-application to test the deployment.~~

1. ~~To run the .NET 10 integration tests, execute:~~

   ```pwsh
   cd .\tests\AISQuick.IntegrationTests\
   dotnet run
   ```

1. ~~For the manual test, only the following two requests will work:~~
   - ~~1 Publish a message to the aisquick-sample topic on the Service Bus via the Sample API in API Management~~
   -  ~~4b Get the blob with the stored message from the aisquick-sample container~~

   ~~4a will fail because the function app is not deployed.~~


~~This template deploys other stuff as well. If you want a specific template to use for e2e scenarios with the azure.logicappsstandard extension, let me know and I'll cook something up.~~

kristenwomack:
Hey @ronaldbosma thank you for this contribution and your patience. We are continuing to test internally.

jongio:
@ronaldbosma let's get the template to work with 'azd init and then azd up' w/o requiring any azd env to get the baseline going.

ronaldbosma:
> Ronald Bosma (@ronaldbosma) let's get the template to work with 'azd init and then azd up' w/o requiring any azd env to get the baseline going.

@jongio In that case my [Azure Integration Services Quickstart](https://github.com/ronaldbosma/azure-integration-services-quickstart) template isn't a good fit because it's purposefully set up to ask the user which resources to install.

@vhvb1989 I've created https://github.com/ronaldbosma/azure-logicappsstandard-azd-extension-sample to test and demo the extension. This template includes two Logic Apps: one is a basic Logic App; the other has a custom code project using .NET 8. 

A GitHub Actions workflow is included that currently uses `https://azuresdkartifacts.z5.web.core.windows.net/azd/extensions/pr/10058/azure-logicappsstandard.zip` to install the extension. When manually triggering the workflow, you can specify a different URL.

PR #8030, opened 2026-05-04
feat(azure.ai.agents): add ConnectionManager for CRUD on Foundry project connections

Nathandrake229:
## Summary

Add a \ConnectionManager\ service layer that provides CRUD operations on connected resources (connections) in a Foundry project.

### New: \connection_manager.go\
- \ConnectionManager\ struct combining ARM SDK (\ProjectConnectionsClient\) for create/get/update/delete/list with data-plane client (\FoundryProjectsClient\) for reading credentials/secrets
- Clean domain types: \ConnectionInfo\, \ConnectionDetail\, \CreateConnectionParams\, \UpdateConnectionParams\
- 6 operations: \List\, \Get\, \GetWithCredentials\, \Create\, \Update\, \Delete\
- Supports ApiKey, CustomKeys, and None auth types
- GET-then-PUT pattern for Update (ARM doesn't support PATCH on connections)
- Raw JSON credential parsing to extract arbitrary custom key names from CustomKeys connections

### New: \connection_manager_test.go\
- 14 unit tests covering body construction for all auth types, credential parsing (including the \key\-name ambiguity edge case), helper functions, and ARM response conversion

### Modified: \oundry_projects_client.go\
- Refactored \GetConnectionWithCredentials\ to use a \getConnectionWithCredentialsRaw\ helper that returns raw bytes, enabling flexible credential field parsing without breaking existing callers

## Design context
- ARM API never returns credential values — data-plane \getConnectionWithCredentials\ is required for secrets
- Connections are ARM resources under \Microsoft.CognitiveServices/accounts/{account}/projects/{project}/connections/{name}\
- The hybrid approach (ARM for CRUD + data-plane for secrets) matches existing patterns in the extension

## Testing
All 40 tests pass (14 new + 26 existing in the azure package).

### DIFF
--- a/cli/azd/extensions/azure.ai.agents/internal/pkg/azure/connection_manager.go
+++ b/cli/azd/extensions/azure.ai.agents/internal/pkg/azure/connection_manager.go
@@ -1,0 +1,442 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package azure
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/to"
+	armcognitiveservices "github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/cognitiveservices/armcognitiveservices/v2"
+)
+
+// ConnectionManager provides CRUD operations on connected resources in a Foundry project.
+// It combines the ARM SDK client (for create/get/update/delete/list) with the data-plane
+// client (for reading credentials/secrets). ARM never returns credential values; the
+// data-plane getConnectionWithCredentials endpoint is the only way to retrieve them.
+type ConnectionManager struct {
+	armClient *armcognitiveservices.ProjectConnectionsClient
+	dpClient  *FoundryProjectsClient // for reading credentials; nil if not needed
+	rg        string
+	account   string
+	project   string
+}
+
+// NewConnectionManager creates a ConnectionManager with both ARM and data-plane clients.
+// The dpClient is used for reading credentials via getConnectionWithCredentials.
+func NewConnectionManager(
+	subscriptionID, rg, account, project string,
+	cred azcore.TokenCredential,
+) (*ConnectionManager, error) {
+	armClient, err := armcognitiveservices.NewProjectConnectionsClient(
+		subscriptionID, cred, NewArmClientOptions(),
+	)
+	if err != nil {
+		return nil, fmt.Errorf("creating ARM connections client: %w", err)
+	}
+
+	dpClient, err := NewFoundryProjectsClient(account, project, cred)
+	if err != nil {
+		return nil, fmt.Errorf("creating data-plane client: %w", err)
+	}
+
+	return &ConnectionManager{
+		armClient: armClient,
+		dpClient:  dpClient,
+		rg:        rg,
+		account:   account,
+		project:   project,
+	}, nil
+}
+
+// ConnectionInfo holds metadata for a connection (no secrets).
+type ConnectionInfo struct {
+	Name      string
+	ID        string
+	Category  string
+	AuthType  string
+	Target    string
+	IsDefault bool
+	Metadata  map[string]string
+}
+
+// ConnectionDetail extends ConnectionInfo with credential key-value pairs
+// retrieved from the data-plane getConnectionWithCredentials endpoint.
+// The Credentials map contains only the actual secret fields (the "type"
+// discriminator is excluded).
+type ConnectionDetail struct {
+	ConnectionInfo
+	Credentials map[string]string
+}
+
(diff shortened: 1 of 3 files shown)


cli/azd/extensions/azure.ai.agents/internal/pkg/azure/connection_manager.go
```diff
@@ -0,0 +1,442 @@
...
+	if err != nil {
+		return fmt.Errorf("deleting connection %q: %w", name, err)
+	}
+	return nil
+}
+
+// fetchCredentials calls the data-plane getConnectionWithCredentials endpoint
+// and returns the credential fields as a flat map (excluding the "type" discriminator).
+func (m *ConnectionManager) fetchCredentials(ctx context.Context, name string) (map[string]string, error) {
+	conn, err := m.dpClient.GetConnectionWithCredentials(ctx, name)
```
jongio, line 213: This calls `GetConnectionWithCredentials` (which internally calls `getConnectionWithCredentialsRaw` + unmarshal), then for non-ApiKey types, `fetchRawCredentials` calls `getConnectionWithCredentialsRaw` again. That's two HTTP round-trips to the same endpoint.

Since you need the raw bytes for CustomKeys anyway, consider calling `getConnectionWithCredentialsRaw` once and parsing both the `Connection` struct (for type detection) and the raw credentials from the same bytes:

```go
func (m *ConnectionManager) fetchCredentials(ctx context.Context, name string) (map[string]string, error) {
    raw, err := m.dpClient.getConnectionWithCredentialsRaw(ctx, name)
    if err != nil {
        return nil, err
    }

    var conn Connection
    if err := json.Unmarshal(raw, &conn); err != nil {
        return nil, fmt.Errorf("parsing connection response: %w", err)
    }

    if conn.Credentials.Type == CredentialTypeApiKey && conn.Credentials.Key != "" {
        return map[string]string{"key": conn.Credentials.Key}, nil
    }

    return parseRawCredentials(raw)
}
```

This also lets you extract `parseRawCredentials` as a standalone function that's directly testable (see comment on the test file).

cli/azd/extensions/azure.ai.agents/internal/pkg/azure/connection_manager.go
```diff
@@ -0,0 +1,442 @@
...
+	}
+
+	result := make(map[string]string, len(envelope.Credentials))
+	for k, v := range envelope.Credentials {
+		if k == "type" {
+			continue // exclude the discriminator
+		}
+		var s string
+		if err := json.Unmarshal(v, &s); err != nil {
+			continue // skip non-string values
```
jongio, line 258: Silently skipping non-string values could hide credential data the caller needs. Consider at minimum a `log.Printf` here so unexpected credential formats don't go unnoticed during debugging.

cli/azd/extensions/azure.ai.agents/internal/pkg/azure/connection_manager_test.go
```diff
@@ -0,0 +1,273 @@
...
+		{"", ""},
+	}
+	for _, tt := range tests {
+		t.Run(tt.input, func(t *testing.T) {
+			require.Equal(t, tt.want, lastSegment(tt.input))
+		})
+	}
+}
+
+func TestParseRawCredentials(t *testing.T) {
```
jongio, line 138: These tests duplicate the credential-parsing logic inline instead of calling the actual code. If `fetchRawCredentials` changes, these tests still pass because they're testing a copy of the algorithm, not the real thing.

If you extract the parsing into a standalone `parseRawCredentials(raw []byte) (map[string]string, error)` function (as suggested on the other comment), these tests can call it directly and verify the real code path.

cli/azd/extensions/azure.ai.agents/internal/pkg/azure/connection_manager.go
```diff
@@ -0,0 +1,442 @@
...
+	metadata := toStringPtrMap(params.Metadata)
+
+	switch params.AuthType {
+	case "ApiKey":
+		authType := armcognitiveservices.ConnectionAuthTypeAPIKey
+		return &armcognitiveservices.ConnectionPropertiesV2BasicResource{
+			Properties: &armcognitiveservices.APIKeyAuthConnectionProperties{
+				AuthType: &authType,
+				Category: &category,
+				Target:   to.Ptr(params.Target),
```
jongio, line 319: CI's `go-fix` check is failing because Go 1.26 prefers `new(val)` over `to.Ptr(val)`. This applies to all 7 `to.Ptr` call sites in this file. Run `go fix ./...` from the extension directory - it'll auto-fix these and likely remove the now-unused `to` import.

jongio (commented):
4 findings on the new ConnectionManager.

The main issue: `fetchCredentials` hits the same data-plane endpoint twice for non-ApiKey connections - once parsed, once raw. Since `getConnectionWithCredentialsRaw` already exists, you can call it once and parse from those bytes. That cuts the round-trips in half.

CI's `go-fix` check is also failing. Run `go fix ./...` from the extension directory to replace `to.Ptr()` with `new()`.

PR #8034, opened 2026-05-04
[Agents Extension] Move Session and Conversation Id Tracking to Global Config

### LINKED ISSUE #7789: [Agents Extension] Version handling with sessions
Sessions are tied to agent versions. What this means then is creating a session on version 1, deploying version 2, and then attempting to invoke for example, will still hit the original session on version 1, which may be unexpected. We should discuss if our cache logic for sessions should be version specific, to avoid this possible confusion.

### LINKED ISSUE #7899: Deploying a new version does not start a new session
## Goal
Invoking an agent uses the latest deployed version of an agent. 

## Context
- Session information is saved in .foundry-agent.json
- --new-session will trigger a new session

## Current Behavior
If a developer runs `azd deploy`, it results in a new version of the agent and possibly new code deployed. Subsequent `invoke` commands could re-use a previous session ID from .foundry-agent.json. 

This means the `invoke` would not be using the latest deployed version of the agent since its an existing session.

## Desired Behavior
Clear the session ID from .foundry-agent.json when deploy is successful and results in a new version of the agent deployed to Foundry. Then the next `invoke` command would automatically create a new session and have the expected behavior.

## Implementation Notes
- Might make sense to do in the existing post deploy hook of the extension
- Should be captured in the logs that the session was cleared

## Acceptance Criteria
- [ ] Unit tests added or updated
- [ ] Existing test suite passes 
- [ ] Lint or format passes 

## How to Test
invoke an agent, then change some code that is obvious to observe in the response and deploy it. 
invoke it again and see the obvious change in the response. 
If not, then the previous session was still used.
Should also be able to see the change of the .foundry-agent.json between invokes has changed.

### LINKED ISSUE #7935: invoke --local outside of the agent workspace doesn't save session or conversation id
When running `azd ai agent invoke --local` from anywhere, it will correctly send a message to localhost, but because its not in the directory it won't save the session / conversation to .foundry-agent.json so you constantly get a new session. 

This is how it works, but we need to make it more obvious that this is happening. 

Workaround: use --session-id or --conversation-id with the command.

trangevi:
Fixes #7899 #7935 

In order to support a "global" invoke call, this modifies the session handling to be stored in the global config. There are quite a few changes here, so providing the copilot summary:

Summary

Replaces the per-project .foundry-agent.json file with the global azd UserConfig (via the ConfigHelper gRPC API) for storing session
and conversation state. This provides a centralized, project-independent store that works across directories and survives project
relocations.

Changes

New: Config store abstraction (internal/cmd/config_store.go)

 - CRUD operations (get/set/delete) for agent-specific values in UserConfig under extensions.ai-agents.<field>
 - Structured agent key format: <host>/path/agents/<name>/versions/<version>/{local|remote}
 - buildRemoteAgentKeyFromEndpoint() — derives key directly from AGENT_{SVC}_ENDPOINT env var
 - normalizeEndpoint() strips scheme, lowercases host, trims trailing slash
 - projectHash() for local-mode key isolation across projects
 - Legacy fallback: getContextValueWithFallback tries new key → old keys → rewrites if found

Refactored: All callers updated

 - invoke.go — uses structured keys for session/conversation persistence; prints invocation ID to console instead of persisting it
 - session.go — persistSessionID uses structured key; SessionDeleteAction clears stored session if it matches the deleted one
 - files.go — uses buildRemoteAgentKey with legacy fallback for session resolution
 - monitor.go — resolveMonitorSession accepts structured key + legacy keys
 - helpers.go — removed file-based I/O, added migrateFromLegacyFile for one-time lazy migration

New: Postdown lifecycle handler (internal/cmd/listen.go)

 - Registered postdown event handler that cleans up stored sessions/conversations when azd down runs
 - Uses AGENT_{SVC}_ENDPOINT (which survives after down) instead of AZURE_AI_PROJECT_ENDPOINT

Removed

 - All writes to .foundry-agent.json
 - Invocation ID persistence (was never read back; only used in-memory for LRO polling)

Tests

 - monitor_test.go — rewrote with unit tests for pure functions (buildAgentKey, resolveStoredIDFromPath, resolveMonitorSessionArgs)

Key design decisions

 - Lazy migration: Old .foundry-agent.json data is migrated on first access, not eagerly
 - Key matches endpoint URL: The config key directly mirrors AGENT_{SVC}_ENDPOINT + "/remote", making postdown cleanup trivial
 - Best-effort cleanup: Config store operations in delete/postdown handlers log failures but never block the primary operation

### DIFF
--- a/cli/azd/extensions/azure.ai.agents/internal/cmd/invoke.go
+++ b/cli/azd/extensions/azure.ai.agents/internal/cmd/invoke.go
@@ -367,7 +367,13 @@ func (a *InvokeAction) responsesRemote(ctx context.Context) error {
 	}
 
 	// Build the structured agent key for config store lookups.
-	agentKey := buildRemoteAgentKeyFromEndpoint(agentEndpoint)
+	// When the endpoint is unavailable (pre-deploy), skip session/conversation persistence.
+	var agentKey string
+	if agentEndpoint != "" {
+		agentKey = buildRemoteAgentKeyFromEndpoint(agentEndpoint)
+	} else {
+		log.Printf("warning: agent endpoint not available, session state will not be persisted")
+	}
 
 	body, bodyLabel, err := a.resolveBody()
 	if err != nil {
@@ -384,12 +390,17 @@ func (a *InvokeAction) responsesRemote(ctx context.Context) error {
 
 	// Session ID — routes to the same microVM container instance.
 	// When empty, let the server assign one.
-	sid, err := resolveStoredID(
-		ctx, azdClient, agentKey, a.flags.session, a.flags.newSession, "sessions", false,
-		legacyKeysForRemote(name)...,
-	)
-	if err != nil {
-		return err
+	var sid string
+	if agentKey != "" {
+		sid, err = resolveStoredID(
+			ctx, azdClient, agentKey, a.flags.session, a.flags.newSession, "sessions", false,
+			legacyKeysForRemote(name)...,
+		)
+		if err != nil {
+			return err
+		}
+	} else if a.flags.session != "" {
+		sid = a.flags.session
 	}
 	if sid != "" {
 		reqBody["session_id"] = sid
@@ -569,7 +580,12 @@ func (a *InvokeAction) invocationsRemote(ctx context.Context) error {
 		return err
 	}
 
-	agentKey := buildRemoteAgentKeyFromEndpoint(agentEndpoint)
+	var agentKey string
+	if agentEndpoint != "" {
+		agentKey = buildRemoteAgentKeyFromEndpoint(agentEndpoint)
+	} else {
+		log.Printf("warning: agent endpoint not available, session state will not be persisted")
+	}
 
 	body, bodyLabel, err := a.resolveBody()
 	if err != nil {
@@ -577,9 +593,14 @@ func (a *InvokeAction) invocationsRemote(ctx context.Context) error {
 	}
 
 	// Session ID — routes to the same container instance
-	sid, err := resolveStoredID(ctx, azdClient, agentKey, a.flags.session, a.flags.newSession, "sessions", false)
-	if err != nil {
-		return err
+	var sid string
+	if agentKey != "" {
+		sid, err = resolveStoredID(ctx, azdClient, agentKey, a.flags.session, a.flags.newSession, "sessions", false)
+		if err != nil {
+			return err
+		}
+	} else if a.flags.session != "" {
+		sid = a.flags.session
 	}
 
 	// Acquire credential and token
--- a/cli/azd/extensions/azure.ai.agents/internal/cmd/invoke.go
+++ b/cli/azd/extensions/azure.ai.agents/internal/cmd/invoke.go
@@ -10,6 +10,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"log"
 	"net/http"
 	"net/url"
 	"os"
@@ -261,12 +262,18 @@ func (a *InvokeAction) responsesLocal(ctx context.Context) error {
 	// Resolve local session and conversation IDs (always generated locally).
 	var sid, convID string
 	if azdClient != nil {
-		sid, _ = resolveStoredID(
+		sid, err = resolveStoredID(
 			ctx, azdClient, agentKey, a.flags.session, a.flags.newSession, "sessions", true,
 		)
-		convID, _ = resolveStoredID(
+		if err != nil {
+			log.Printf("invoke local: failed to resolve session ID: %v", err)
+		}
+		convID, err = resolveStoredID(
 			ctx, azdClient, agentKey, a.flags.conversation, a.flags.newConversation, "conversations", true,
 		)
+		if err != nil {
+			log.Printf("invoke local: failed to resolve conversation ID: %v", err)
+		}
 	}
 
 	fmt.Printf("Target:       localhost:%d (local)\n", port)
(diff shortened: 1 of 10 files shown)


wbreza (commented):
## Code Review — PR #8034

**Summary:** Replaces per-project .foundry-agent.json with global azd UserConfig for session/conversation state. Adds config_store.go abstraction (244 lines), lazy migration, postdown lifecycle handler, and rewrites monitor tests.

---

### 🔴 Must Fix

**1. No tests for config_store.go** — the core new module (244 lines) has zero test coverage. 
ormalizeEndpoint(), projectHash(), uildAgentKey(), getContextValueWithFallback(), and all CRUD operations are untested. This is the persistence layer for session state. Create config_store_test.go with table-driven tests covering endpoint normalization, key construction, projectHash determinism, fallback/migration logic, and CRUD round-trips.

**2. No input validation on key construction parameters** — gentName, ersion, and storeField in uildAgentKey() are interpolated into config keys without validation. An agent name containing / could produce malformed keys. Validate parameters don't contain path separators; validate storeField against an allow-list ("sessions", "conversations").

**3. Migration deletes legacy file after partial success** — migrateFromLegacyFile() sets migrated = true if *any single* key migrates, then deletes the entire .foundry-agent.json. If some keys fail (e.g., gRPC transient error), they're permanently lost. Only delete when *all* entries succeed, or keep the file and remove only migrated keys.

### 🟡 Should Fix

**4. Pervasive silent error swallowing** — Multiple call sites in helpers.go, session.go, and invoke.go discard config store errors with _ = .... While best-effort is valid, add log.Printf at every error discard site for debuggability.

**5. Postdown handler skips local-mode cleanup** — postdownHandler() cleans remote sessions but local-mode sessions (keyed by localhost:{port}/{projectHash}/...) persist indefinitely. Also build and delete local agent keys during postdown.

**6. Confusing dual API: uildRemoteAgentKeyFromEndpoint() vs uildRemoteAgentKey()** — Two functions build remote keys from different inputs. Add clear doc comments explaining when to use each, or add a test asserting both produce the same key for the same agent.

**7. Net test coverage regression (−65 lines)** — monitor_test.go went from 144 deletions to 79 additions. No tests for session delete cleanup, postdown handler, migration logic, or helper ID resolution functions.

### 🟢 Nitpick

**8.** 
ormalizeEndpoint() lowercases host but preserves path case — document this behavior so future callers know gents/Foo ≠ gents/foo.

**9.** projectHash() SHA256→8 bytes is fine for local isolation; add a comment noting the collision probability is acceptable.

**10.** PR description says key format is AGENT_{SVC}_ENDPOINT + "/remote" but actual format is {normalized_endpoint}/agents/{name}/versions/{version}/{local|remote} — minor description update recommended.

### ✅ Done Well

- Clean config_store.go CRUD abstraction over gRPC ConfigHelper
- Pragmatic lazy migration from .foundry-agent.json
- Postdown lifecycle cleanup registration is the right pattern
- Correct removal of unused invocation ID persistence
- Good 4-commit structure with clear messages

wbreza (approved):
﻿## Re-Review — Commit `5f6231b` ("PR comments")

Nice work addressing the feedback! Here's the updated status:

### ✅ Resolved

| # | Finding | Resolution |
|---|---------|------------|
| 1 | No tests for `config_store.go` | `config_store_test.go` added — covers validation, key building, normalization, hash determinism |
| 2 | No input validation | `validateStoreField()` and `validateKeySegment()` added, wired into CRUD functions |
| 3 | Migration deletes after partial success | Fixed — `allSucceeded` flag, file only deleted when all entries migrate |
| 4 | Silent error swallowing | `log.Printf` added at all error discard sites |
| 5 | Postdown skips local-mode cleanup | Fixed — `defer` cleanup in `run.go` for local sessions on exit |
| 6 | Confusing dual key-building API | `buildRemoteAgentKey()` removed, all callers consolidated on `buildRemoteAgentKeyFromEndpoint()` |
| 7 | Net test coverage regression | 160 lines of tests added, net positive |

### 🟡 Remaining (non-blocking, follow-up candidates)

**`buildAgentKey()` doesn't call `validateKeySegment()` on agentName/version** — The validation functions exist but aren't wired into the key builder itself. Low risk since values come from project config, but wiring it in would complete the defense-in-depth story.

**CRUD and fallback functions untested** — `getAgentSpecificContextValue()`, `setAgentSpecificContextValue()`, `deleteContextValue()`, and `getContextValueWithFallback()` don't have unit tests yet. These require gRPC mocking which is harder — consider integration tests as a follow-up.

**`normalizeEndpoint()` edge cases** — Empty string and host-only inputs aren't covered in tests. Minor gap.

### ✅ Approving

The critical issues are resolved. Remaining items are low-risk and can be follow-up work.

jongio (commented):
The overall design is clean - structured keys, lazy migration, and best-effort cleanup are good choices for CLI state management.

One concern not covered by existing feedback: when `agentEndpoint` is empty (env var not set), `buildRemoteAgentKeyFromEndpoint("")` produces `/remote` - a degenerate key shared across all agents without an endpoint. The legacy fallback works on first access (rewrites from `name` to `/remote`), but subsequent writes from different agents would collide on the same key.

Consider guarding: if `agentEndpoint` is empty, fall back to using `name` as the key directly (old behavior). This keeps the migration path simple and avoids the collision case.

Same pattern appears in both `responsesRemote` and `invocationsRemote`.

cli/azd/extensions/azure.ai.agents/internal/cmd/invoke.go
```diff
@@ -357,6 +366,9 @@ func (a *InvokeAction) responsesRemote(ctx context.Context) error {
 		return err
 	}
```
jongio, line 368: If `resolveAgentServiceFromProject` fails or `info.AgentEndpoint` is empty (`AGENT_{SVC}_ENDPOINT` not set, e.g. pre-first-deploy), `agentEndpoint` stays `""` and this produces `/remote`. Two agents in that state would share the same config key and overwrite each other's sessions.

Suggested guard:

```go
agentKey := name // fallback to old behavior
if agentEndpoint != "" {
    agentKey = buildRemoteAgentKeyFromEndpoint(agentEndpoint)
}
```

trangevi: I think there's the possibility to hit issues if we fall back to just the name, so I've updated to log a warning and skip the flow altogether. It can be something we iterate on given feedback.

jongio (approved):
Addresses my earlier feedback. The skip-persistence approach for empty endpoints is cleaner than falling back to name - avoids a different class of collision issues. The defensive guards are consistent across all call sites and the dead code cleanup is welcome.

PR #8038, opened 2026-05-04
[ai-agents] Fix MCP tool field mapping

### LINKED ISSUE #8037: [ai-agents] MCP tools initialized without server_url or connection id
## Description

MCP tools defined in an agent manifest can lose the fields needed by the Responses API after `azd ai agent init` processes the manifest.

A manifest using the sample-style MCP shape:

```yaml
tools:
  - kind: mcp
    connection:
      kind: foundry
      endpoint: https://api.githubcopilot.com/mcp/
      name: github-mcp-oauth
    name: github-mcp-remote
    url: https://api.githubcopilot.com/mcp/
```

is parsed without preserving the top-level `url` or the named connection into the API MCP tool model. When the tool is later retrieved and used with the Responses API, it may contain neither `server_url` nor a connection/connector id.

## Error

```text
Error code: 400 - {'error': {'message': "Missing mutually exclusive parameters: 'tools[2]'. Ensure you are providing exactly one of: 'server_url' or 'connector_id'.", 'type': 'invalid_request_error', 'param': 'tools[2]', 'code': 'missing_mutually_exclusive_parameters'}}
```

## Expected behavior

The MCP tool produced from the manifest should preserve the required routing field:

- top-level `url` should map to the API tool `server_url`
- `connection.name` should be preserved as the project connection id

## Notes

The existing options-based form (`options.serverUrl` and `options.projectConnectionId`) is already handled, but the sample-style top-level MCP fields are not.

glharper:
## Summary

Fixes #8037

Preserves MCP routing fields when `azd ai agent init` and toolbox provisioning process MCP tool manifests. Top-level MCP `url` now maps to the API `server_url`, `connection.name` is preserved as the project connection id, and toolbox tools are enriched from the full configured connection set before being sent to the Foundry Toolsets API.

## Details

- Treat `kind: foundry` MCP connections as the existing remote connection shape instead of introducing a duplicate type.
- Preserve explicit MCP `url` during YAML-to-API conversion.
- Preserve named MCP connections as `project_connection_id` during YAML-to-API conversion.
- Enrich toolbox tools from both `connections` and `toolConnections` when filling `server_url` and `server_label`.
- Add regression coverage for sample-style MCP manifests and generic toolbox connection enrichment.

## Validation

```text
go test ./... -count=1
```


glharper:
@copilot apply changes based on the comments in [this thread](https://github.com/Azure/azure-dev/pull/8038#pullrequestreview-4223269782)

jongio (commented):
Clean, focused fix. The data flow through ExtractToolsDefinitions -> ExtractConnectionDefinition -> convertYamlToolToApiTool is correct, and the precedence (top-level `url` sets initial value, `options["serverUrl"]` can override) is the right call.

One inline comment on naming. The bot's `Url` -> `URL` point is valid - the rest of the API layer uses `ServerURL`, `BaseURL`, etc.

cli/azd/extensions/azure.ai.agents/internal/pkg/agents/agent_yaml/yaml.go
```diff
@@ -241,6 +242,13 @@ type RemoteConnection struct {
 	Endpoint   string `json:"endpoint" yaml:"endpoint"`
 }
 
+// FoundryConnection Connection configuration for Foundry project connections.
+type FoundryConnection struct {
+	Connection `json:",inline" yaml:",inline"`
+	Name       string `json:"name" yaml:"name"`
+	Endpoint   string `json:"endpoint" yaml:"endpoint"`
```
jongio: `FoundryConnection` is structurally identical to `RemoteConnection` (both have `Connection` inline + `Name` + `Endpoint`). If they're expected to diverge, a short comment explaining the distinction would help future readers. If not, consider a type alias or embedding `RemoteConnection` directly.

cli/azd/extensions/azure.ai.agents/internal/pkg/agents/agent_yaml/yaml.go
```diff
@@ -241,6 +242,13 @@ type RemoteConnection struct {
 	Endpoint   string `json:"endpoint" yaml:"endpoint"`
 }
 
+// FoundryConnection Connection configuration for Foundry project connections.
+type FoundryConnection struct {
```
trangevi: Why do we need a new connection type?

cli/azd/extensions/azure.ai.agents/internal/pkg/agents/agent_yaml/map.go
```diff
@@ -277,12 +277,19 @@ func convertYamlToolToApiTool(yamlTool any) (any, error) {
 		return apiTool, nil
 
 	case McpTool:
+		serverLabel := tool.ServerName
```
trangevi: I think this change is unnecessary, except perhaps making sure we populate the serverURL if it's available. But I think the issue is more general, we're not properly iterating over the full set of connections here: https://github.com/Azure/azure-dev/blob/main/cli/azd/extensions/azure.ai.agents/internal/cmd/listen.go#L634.

I think that could be done either at ingestion or consumption, but it should be able to be done more generically.

trangevi (changes requested):

wbreza:
## Re-Review — PR #8038 (commit c1233a74)

**Recommendation: Approve** :white_check_mark:

All substantive feedback from the first review has been addressed. The Url to URL naming fix, serverName precedence test, and FoundryConnection endpoint fallback logic all look correct.

**Remaining nitpick:** Test assertion at ~line 552 should print actual value for consistency.

Ship it. :+1:

wbreza (approved):

cli/azd/extensions/azure.ai.agents/internal/pkg/agents/agent_yaml/map.go
glharper: [Updated](https://github.com/Azure/azure-dev/pull/8038/commits/8b085ae1c2d1b12debd77ef9b1702f95aa18a581)

cli/azd/extensions/azure.ai.agents/internal/pkg/agents/agent_yaml/yaml.go
glharper: [Removed](https://github.com/Azure/azure-dev/pull/8038/commits/8b085ae1c2d1b12debd77ef9b1702f95aa18a581)

cli/azd/extensions/azure.ai.agents/internal/pkg/agents/agent_yaml/yaml.go
glharper: [Removed](https://github.com/Azure/azure-dev/pull/8038/commits/8b085ae1c2d1b12debd77ef9b1702f95aa18a581)

jongio (commented):
Incremental review of commits since my last look: both findings addressed. `Url` renamed to `URL`, and `FoundryConnection` removed in favor of reusing `RemoteConnection` (cleaner). The `ServerLabel` fallback logic moved to runtime enrichment in `enrichToolboxFromConnections`, which is the right layer for it. No new issues.

cli/azd/extensions/azure.ai.agents/internal/pkg/agents/agent_yaml/parse.go
```diff
@@ -358,6 +358,12 @@ func ExtractConnectionDefinition(connectionBytes []byte) (any, error) {
 			return nil, fmt.Errorf("failed to unmarshal to RemoteConnection: %w", err)
 		}
 		return remoteConn, nil
+	case ConnectionKindFoundry:
```
trangevi: Connection types are defined here: https://learn.microsoft.com/en-us/azure/templates/microsoft.cognitiveservices/accounts/projects/connections?pivots=deployment-language-bicep#connectionpropertiesv2

Is there an ask to add a new one that I missed? I'm not sure I understand what this is that shouldn't just be handled by the remote connection kind

glharper: [Removed](https://github.com/Azure/azure-dev/pull/8038/commits/4d47c95ec5eeb19376c6656c20af389b6e62c3e5)

glharper:

trangevi (approved):

Issue #7962, opened 2026-04-29
Unify Foundry agent configuration in azure.yaml


therealjohn:
This proposes two changes to how the `azure.ai.agent` extension models a Foundry agent project in `azure.yaml`:

1. **Consolidate** all hosted-agent config into `azure.yaml`, retiring `agent.yaml` and `agent.manifest.yaml`.
2. **Restructure** so that a single `azure.ai.project` service owns all Foundry data-plane state -- toolboxes, connections, model deployments, and future project-scoped resources. Agents reference it via `uses`. No Bicep files are required by default; developers can opt in to Bicep on disk when they need full IaC reproducibility.

Today, `azure.ai.agent` packs every project-scoped resource inside one hosted agent service, and ships two extra config files alongside `azure.yaml`. That made sense as a starting point, but it conflates the agent runtime with the Foundry project around it, and it makes sharing resources across agents awkward. The Foundry Toolkit for VS Code will move to reading `azure.yaml` directly once these changes land.

## Current Problems

1. **Three files, overlapping data.** The agent name appears in three places, container resources in two, the model deployment name in three. Two templating syntaxes (`{{param}}` and `${ENV}`) overlap.
2. **Scope conflation.** `services.<agent>.config` mixes things that genuinely belong to one agent (container resources, env, startup command) with project-scoped resources (model deployments, connections, toolboxes). Some of these are ARM resources that should live in Bicep; the rest are Foundry data-plane resources that don't belong nested under any single agent.
3. **No sharing across agents.** Because project-scoped resources are nested under an agent today, a second agent that wants the same toolbox has to redeclare it. There is nowhere to say "this toolbox belongs to the project; these agents reference it."
4. **Divergent tooling.** The Foundry Toolkit parses `agent.yaml` (AgentDefinition) directly, `azd ai agent` use an AgentManifest and generates an AgentDefinition, but has to also mix orchestration with azure.yaml. They feel like separate experiences.
5. **The manifest layer carries no weight.** `agent.manifest.yaml` was designed for an agent catalog that didn't get built. The templating it adds isn't paying for itself.
6. **No real ability to share** AgentDefinitions were intended to be concrete definitions, but in practice any real values get abstracted with AZD environments (`${ENV_VAR}`) effectively becoming an templated definition, which confused the purpose of an AgentManifest.

## Solution Hypothesis

The shape we want is:

[...]

PR #8057, opened 2026-05-05
docs(ai-agents): design doc for context-aware next-step guidance

### LINKED ISSUE #7975: Guide developers to the right next step after every `azd ai agent` command and add `azd ai agent doctor`
After every successful `azd ai agent` command, the CLI should inspect the project's real state -- `azure.yaml` services, azd env vars, running endpoints -- and guide the developer exactly what to do next. We also need a new **`azd ai agent doctor`** command for on-demand health checks.

Right now, `azd ai agent init` tells you to deploy. That's it. No mention of local dev, no prereq checks, no adaptation to the sample you picked. A developer who just scaffolded a responses-protocol agent has no idea they can run it locally. A developer with unresolved toolbox deps gets told to `run` when they should `azd provision` first. We fix this by making every success path exit with state-aware, actionable guidance, and by giving developers `doctor` as a dedicated diagnostic.

Scope: `init`, `run`, `invoke` (local and remote), `show`, and `deploy` (via post-deploy hooks). Error-path guidance is out of scope -- [`exterrors`](https://github.com/Azure/azure-dev/tree/main/cli/azd/extensions/azure.ai.agents/internal/exterrors) already handles that well and doesn't need rework.

Assumes the [Unified azure.yaml proposal](unify-azure-yaml.md) has landed -- `azure.yaml` is the sole config file, `agent.yaml` and `agent.manifest.yaml` are gone.

## Current Problems

1. **Init pushes you straight to deploy.** After `azd ai agent init`, the CLI checks `AZURE_AI_PROJECT_ID` and suggests either `azd deploy <service>` or `azd up` ([init.go#L1718-L1733](https://github.com/Azure/azure-dev/blob/main/cli/azd/extensions/azure.ai.agents/internal/cmd/init.go#L1718-L1733), [init_from_code.go#L136-L140](https://github.com/Azure/azure-dev/blob/main/cli/azd/extensions/azure.ai.agents/internal/cmd/init_from_code.go#L136-L140)). Both are deploy paths. Local dev (`run` -> `invoke --local`) never gets mentioned. For prototypers, local iteration should be the default -- deployment is "when you're ready," not the only option.

2. **Run's invoke hint is static and often wrong.** After the agent starts, `run` prints `After startup, in another terminal, try: azd ai agent invoke --local "Hello!"` ([run.go#L157-L158](https://github.com/Azure/azure-dev/blob/main/cli/azd/extensions/azure.ai.agents/internal/cmd/run.go#L157-L158)). Same string every time, regardless of protocol. An invocations-protocol agent needs `'{"message": "Hello!"}'`, not a bare string. The hint works for responses-protocol agents and is misleading for everything else.

3. **Success paths are silent.** Error paths have solid suggestions via the [`exterrors` package](https://github.com/Azure/azure-dev/tree/main/cli/azd/extensions/azure.ai.agents/internal/exterrors) (`Validation`, `Dependency`, `Auth`, `Configuration`, `Compatibility` -- each with a `Suggestion` field). But when a command succeeds? Nothing. After a successful `invoke --local`, `show`, or remote `invoke`, the developer gets output and a blank terminal. No "here's what to try next."

4. **Init conflates provision and deploy.** Some samples need `azd provision` before `run` -- to create model deployments, toolboxes, or connections via Bicep -- but they don't need a full `azd up` (which also deploys the agent). Today init suggests `azd up` or `azd deploy` without distinguishing these, so developers deploy prematurely when all they wanted was to set up deps for local dev.

5. **No way to recover context or diagnose issues in one shot.** Two concrete gaps: (a) You lost the CLI's suggestion -- cleared the terminal, closed it, switched contexts -- and now you don't remember what to do next. There's no command to ask "where was I?" (b) You hit an error from a command and don't know how to fix it. The error message might say what went wrong, but it doesn't check the broader context -- are your RBAC roles correct? Is the Foundry project reachable? Are your env vars stale? A `doctor` command could run comprehensive checks and give you the specific commands to fix what's broken.

## Solution Hypothesis

Every `azd ai agent` command's success path runs a **next-step resolver** that:

1. Assembles project state from multiple sources (see State Resolution below).
2. Walks a decision tree to pick the right suggestion for the command that just ran and the state it left behind.
3. After `run`, probes the agent's OpenAPI endpoint (`/invocations/docs/openapi.json`) to discover protocol-specific invoke payloads and build exact example commands. The extension [already has this code](https://github.com/Azure/azure-dev/blob/991b9743b4ab14ee8561677ffa627735c875a796/cli/azd/extensions/azure.ai.agents/internal/cmd/helpers.go#L296) -- the spec is optional and depends on the agent author including it.

### State Resolution

The resolver pulls project state from several sources, in priority order. Higher layers win, and the resolver works with whatever's available:

1. **Explicit flags** (`--project-endpoint`, `--agent`, etc.) -- always authoritative.
2. **Live runtime state** -- probe running agent endpoints, Foundry API status checks.
3. **`azure.yaml`** -- services, protocols, `uses` dependencies, `config.env` references.
4. **azd environment** (`.azure/<env>/.env`) -- which `${...}` variables are actually populated.

This layering matters because we're moving toward environment-optional -- data-plane commands work with just `--project-endpoint`, and brownfield developers may have set up Foundry manually. The resolver should give useful guidance with partial state, not go silent when the azd environment is missing or stale.

Output is human-readable text appended after the command's normal output, separated by a blank line.

### What We Can Ship Today

We can ship **better command choice** without any dependencies -- the resolver tells you which command to run next based on actual project state, not a hardcoded string. Works for every sample, every protocol, today.

**Exact invoke payload examples** (showing the right JSON body for your specific agent) work today for agents that include an OpenAPI spec. The invocation protocol already supports serving a spec at `GET /invocations/docs/openapi.json`, and the extension already has [`fetchOpenAPISpec`](https://github.com/Azure/azure-dev/blob/991b9743b4ab14ee8561677ffa627735c875a796/cli/azd/extensions/azure.ai.agents/internal/cmd/helpers.go#L296) to fetch and cache it. This is opt-in -- the agent author must define the spec (see [human-in-the-loop example](https://github.com/microsoft-foundry/foundry-samples/blob/fd4ecab832fa491b9ffb4abca862c073777b5e53/samples/python/hosted-agents/bring-your-own/invocations/human-in-the-loop/main.py#L119)). We include it by default in our samples so the getting-started happy path always has rich examples. When the spec isn't available, the resolver falls back to protocol-generic examples (`"Hello!"` for responses, `'{"message": "Hello!"}'` for invocations) plus a helpful hint to check `/invocations` directly to verify the right payload shape.

## Success Measures

- **No more dead-end exits.** A developer who runs `init` should always see a runnable next command -- not just "deploy."
- **Init-to-invoke is discoverable from CLI output alone.** You shouldn't need docs to figure out how to get from `init` to a working `invoke --local`.
- **Fewer misdirected commands.** If your deps aren't provisioned, you hear `azd provision`, not `azd ai agent run`. If you're ready for Azure, you hear `azd up`.
- **One-pass diagnosis.** `azd ai agent doctor` surfaces permission issues, missing env vars, and unreachable endpoints in a single run, instead of making you discover them through trial and error across five different commands.

## Decision Tree

The resolver fires after each command and evaluates the following. All `azure.yaml` field references use the post-unification schema from the [Unified azure.yaml proposal](unify-azure-yaml.md).

### State Inputs

| Input | Source | What it tells us |
|---|---|---|
| `HasProjectEndpoint` | azd env: `AZURE_AI_PROJECT_ENDPOINT` is set and non-empty | Foundry project is provisioned |
| `HasUnresolvedInfraVars` | Walk `azure.yaml` service configs; collect `${...}` references that map to known Bicep outputs (e.g., `AZURE_AI_MODEL_DEPLOYMENT_NAME`, connection IDs); check which are missing from azd env | Infra-provisioned dependencies haven't been deployed |
| `HasUnresolvedManualVars` | Remaining `${...}` references not attributable to Bicep outputs (e.g., API keys, custom config); check which are missing from `.env` or azd env | User-supplied values need manual setup |
| `HasToolboxes` | `azure.yaml` has services with `host: azure.ai.toolbox` | Project uses toolboxes that need server-side provisioning |
| `HasConnections` | `azure.yaml` agent `config.env` references connection env vars (e.g., `${GITHUB_MCP_CONN}`) that are unresolved | Connections not yet provisioned via Bicep |
| `Protocol` | `azure.yaml` agent `config.protocols[0].protocol` | Determines invoke payload shape |
| `IsDeployed` | azd env: agent deployment metadata or prior successful deploy | Agent has been pushed to Foundry |
| `AgentStatus` | Foundry API: agent show status | Runtime state of the deployed agent |
| `HasOpenAPI` | Probe running agent at `GET /invocations/docs/openapi.json` | Rich invoke examples available (opt-in, agent author defines spec) |

### After `init`

Today, init checks `AZURE_AI_PROJECT_ID` and branches between `azd deploy` and `azd up`. We expand this to cover local dev, distinguish provision from deploy, and inspect deeper state:

```
IF HasUnresolvedInfraVars:
    -> "azd provision" (create Foundry project and/or dependencies for local development)
ELSE IF HasUnresolvedManualVars:
    -> "azd env set <KEY> <value>" (user-supplied config needed in the azd environment)
ELSE:
    -> "azd ai agent run" (everything is set up, start locally)
```

**Example output (everything ready, project set up):**

```
Next:  azd ai agent run                       -- start the agent locally
       azd ai agent invoke --local "Hello!"    -- test it in another terminal

When ready to deploy to Azure, run azd deploy.
```

**Example output (dependencies not yet created):**

```
Next:  azd provision  -- set up your Foundry project, models, and connections

This creates your Foundry project and any model deployments, toolboxes,
or connections your agent needs for local development.
Once that finishes, run 'azd ai agent run' to start locally.

When ready to deploy to Azure, run azd deploy.
```

**Example output (project ready, but manual config values missing):**

```
Your project is ready, but some values need to be set:
  - MY_API_KEY (not set)

Set them in your azd environment:
  azd env set MY_API_KEY <your-value>

Then run 'azd ai agent run' to start locally.
```

### After `run` (on successful startup)

`run` is a long-lived foreground command. Today it prints a static hint: `After startup, in another terminal, try: azd ai agent invoke --local "Hello!"` ([run.go#L157-L158](https://github.com/Azure/azure-dev/blob/main/cli/azd/extensions/azure.ai.agents/internal/cmd/run.go#L157-L158)). We make this protocol-aware:

```
IF Protocol == "invocations":
    -> invoke --local '{"message": "Hello!"}'
ELSE IF Protocol == "responses":
    -> invoke --local "Hello!"
ELSE:
    -> invoke --local "Hello!" (generic)

+ HasOpenAPI -> replace generic payload with OpenAPI-derived example
```

**Example output (responses protocol):**

```
Agent is running at http://localhost:8088

Next:  azd ai agent invoke --local "Hello!"  -- test it in another terminal

Press Ctrl+C to stop.
```

**Example output (invocations protocol, with OpenAPI):**

```
Agent is running at http://localhost:8088

Next:  azd ai agent invoke --local '{"message": "What hotels are available in Seattle?"}'

Press Ctrl+C to stop.
```

### After `invoke --local`

Today, invoke's failure path already has one suggestion: when the HTTP connection fails, it prints `could not connect to localhost:<port> -- is the agent running? Start it with: azd ai agent run` ([invoke.go#L303](https://github.com/Azure/azure-dev/blob/main/cli/azd/extensions/azure.ai.agents/internal/cmd/invoke.go#L303), [invoke.go#L507](https://github.com/Azure/azure-dev/blob/main/cli/azd/extensions/azure.ai.agents/internal/cmd/invoke.go#L507)). We add success-path guidance:

```
IF success:
    -> "azd deploy" (push the agent to Azure)
    + "azd ai agent monitor --follow" (view logs after deploying)
```

To get to a successful `invoke --local`, the developer must have already provisioned (dependencies exist) and run the agent locally. So the natural next step is `azd deploy` -- infrastructure is already there, just push the agent.

**Example output (single agent):**

```
Next:  azd deploy  -- deploy the agent to Azure
       azd ai agent monitor --follow  -- view logs after deploying
```

**Example output (multi-agent project):**

```
Next:  azd deploy  -- deploy all agents to Azure

After deploying:
  azd ai agent invoke my-agent "Hello!"       -- test my-agent remotely
  azd ai agent invoke other-agent "Hello!"    -- test other-agent remotely
```

### After `invoke` (remote)

```
IF success:
    -> "azd ai agent show <agent>" (check agent status and details)
    + "azd ai agent monitor --follow" (view live logs)
IF failure:
    -> "azd ai agent monitor --follow" (check logs for errors)
```

### After `show`

Today, `show` has error-path guidance when the agent name or version can't be resolved: `Run 'azd deploy' first to deploy the agent, or check your azd environment values` ([show.go#L70-L83](https://github.com/Azure/azure-dev/blob/main/cli/azd/extensions/azure.ai.agents/internal/cmd/show.go#L70-L83)). We add success-path guidance:

```
IF status == "active" OR status == "idle":
    -> "azd ai agent invoke <agent> 'Hello!'" (send a test message)
IF status == "failed" OR status == "":
    -> "azd ai agent monitor --follow" (check logs)
ELSE:
    -> "azd ai agent show <agent>" (agent is in a transitional state, check again shortly)
```

### After `deploy` (via post-deploy hook)

Deployment goes through core `azd up` / `azd deploy`. The extension's deploy hooks ([service_target_agent.go](https://github.com/Azure/azure-dev/blob/main/cli/azd/extensions/azure.ai.agents/internal/project/service_target_agent.go)) run during deployment but don't print next-step guidance when they finish. We add agent-specific suggestions to the post-deploy hook's exit path:

```
IF success:
    -> "azd ai agent show <agent>" (verify the agent is running)
    + "azd ai agent invoke <agent> 'Hello!'" (test the deployment)
IF failure:
    -> "azd ai agent monitor --follow" (check deployment logs)
```

**Example output (success, single agent):**

```
Next:  azd ai agent show echo-agent              -- verify it's running
       azd ai agent invoke echo-agent "Hello!"    -- test the deployment
```

**Example output (success, multi-agent project):**

```
Next:  azd ai agent show my-agent        -- verify my-agent is running
       azd ai agent show other-agent     -- verify other-agent is running
       azd ai agent invoke my-agent "Hello!"     -- test my-agent
       azd ai agent invoke other-agent "Hello!"  -- test other-agent
```

### Why This Matters

Two specific gaps drive this:

1. **You lost context.** You cleared the terminal, closed it, switched to another task. The CLI told you what to do next, but that output is gone. There's no command to ask "where was I?" -- `doctor` fills that role by re-evaluating project state and telling you the next step.

2. **You hit an error and don't know how to fix it.** A command failed, but the error only tells you what went wrong in that command's scope. It doesn't check the broader picture -- are your RBAC roles correct? Is the Foundry project reachable? Are env vars stale? Is the model deployment actually there? `doctor` runs comprehensive checks across the full project and gives you the specific commands to fix what's broken.

### What `doctor` Does

`doctor` runs a fixed set of checks in order, top to bottom. Each check either passes (with a summary) or fails (with a specific remediation command). The checks are designed to cover the full surface area that individual commands can't see on their own.

**Check 1: azure.yaml validity.**
Verify the file exists, parses correctly, and declares at least one `azure.ai.agent` service.
- Pass: `azure.yaml: valid (2 services: echo-agent, agent-toolbox)`
- Fail: `azure.yaml: not found` or `azure.yaml: parse error at line 14`
- Fix: `azd ai agent init` (if missing) or manual edit (if malformed)

**Check 2: Authentication.**
Run the equivalent of `az account show` to verify the developer is signed in and has an active token.
- Pass: `Authentication: signed in as user@contoso.com`
- Fail: `Authentication: not signed in`
- Fix: `az login`

**Check 3: Foundry project reachability.**
Check that `AZURE_AI_PROJECT_ENDPOINT` is set, then probe the endpoint to confirm the project exists and responds.
- Pass: `Foundry project: reachable (endpoint: https://...)`
- Fail (not set): `Foundry project: not provisioned`
- Fail (unreachable): `Foundry project: endpoint set but not reachable (HTTP 403 or timeout)`
- Fix: `azd provision` (if not set) or check network/firewall (if unreachable)

**Check 4: Model deployments.**
For each `${...}` reference in `config.env` that maps to a model deployment, verify the env var is set and (if the project is reachable) that the deployment actually exists in the Foundry project.
- Pass: `Model deployment: gpt-4o (deployed)`
- Fail: `Model deployment: AZURE_AI_MODEL_DEPLOYMENT_NAME not set`
- Fix: `azd provision` (if defined in Bicep) or manual creation in the Foundry portal

**Check 5: Toolboxes.**
For each `azure.ai.toolbox` service in `azure.yaml`, verify it has been provisioned server-side.
- Pass: `Toolbox 'agent-toolbox': provisioned (3 tools)`
- Fail: `Toolbox 'agent-toolbox': not provisioned`
- Fix: `azd provision`

**Check 6: Connections.**
For each connection env var referenced in toolbox or agent configs, verify it's set and (if reachable) that the connection exists in the Foundry project.
- Pass: `Connection GITHUB_MCP_CONN: set`
- Fail: `Connection GITHUB_MCP_CONN: not set`
- Fix: `azd provision` or check Bicep outputs

**Check 7: RBAC permissions.**
Query the Foundry project's role assignments to verify the current identity has the required roles. This is the check that individual commands almost never do, and it's one of the most common sources of confusing errors.
- Pass: `Permissions: sufficient (Contributor + Cognitive Services User)`
- Fail: `Permissions: missing 'Cognitive Services User' role on the Foundry project`
- Fix: `az role assignment create --assignee user@contoso.com --role "Cognitive Services User" --scope <resource-id>`
- Alt: `Or ask your admin to grant access.`

**Check 8: Agent status.**
If the agent has been deployed (deployment metadata exists in azd env), query Foundry for its current status.
- Pass: `Agent 'echo-agent': active (v3)`
- Fail: `Agent 'echo-agent': failed`
- Fix: `azd ai agent monitor --follow` (check logs)

**Check 9: Manual env vars.**
Any `${...}` references in `azure.yaml` that aren't attributable to Bicep outputs -- verify they're set in `.env` or azd env.
- Pass: `MY_API_KEY: set`
- Fail: `MY_API_KEY: not set`
- Fix: `Add to .env`

### Example Output (all healthy)

```
azd ai agent doctor

  azure.yaml:       valid (2 services: echo-agent, agent-toolbox)
  Authentication:   signed in as user@contoso.com
  Foundry project:  reachable
  Model deployment: gpt-4o (deployed)
  Toolbox:          agent-toolbox (ready, 3 tools)
  Permissions:      sufficient
  Agent:            echo-agent (active, v3)

All checks passed.

Next:  azd ai agent invoke echo-agent "Hello!"  -- test the deployed agent
```

### Example Output (issues found)

```
azd ai agent doctor

  azure.yaml:       valid (2 services: echo-agent, agent-toolbox)
  Authentication:   signed in as user@contoso.com
  Foundry project:  not created
  Model deployment: not set
  Toolbox:          not created
  Permissions:      unknown (no project yet)
  Agent:            not deployed

2 issues found:

  1. Foundry project hasn't been created yet.
     Run:  azd provision

  2. Model deployment AZURE_AI_MODEL_DEPLOYMENT_NAME is not set.
     This will be created by 'azd provision' if defined in your Bicep templates.

Next:  azd provision  -- set up your Foundry project, models, and connections
```

### Example Output (permission issue)

```
azd ai agent doctor

  azure.yaml:       valid (1 service: echo-agent)
  Authentication:   signed in as user@contoso.com
  Foundry project:  reachable
  Model deployment: gpt-4o (deployed)
  Permissions:      missing 'Cognitive Services User' role
  Agent:            echo-agent (failed)

1 issue found:

  1. Current identity is missing the 'Cognitive Services User' role
     on the Foundry project resource.
     Run:  az role assignment create --assignee user@contoso.com \
             --role "Cognitive Services User" \
             --scope /subscriptions/.../resourceGroups/.../providers/...
     Or ask your admin to grant access.

Next:  azd ai agent monitor --follow  -- check agent logs for details
```

### How `doctor` and Next-Step Hints Relate

They cover different situations:

- **Next-step hints** fire automatically after each command succeeds. They answer "what should I do now?" If you're in the flow and following the CLI's guidance, you may never need `doctor`.
- **`doctor`** is for when you've lost that thread -- cleared the terminal, hit an error you can't diagnose, or are picking up a project cold. It re-evaluates everything from scratch and gives you both the status and the fix.

Under the hood, both share the same `ProjectState` assembly logic. `doctor` runs every check unconditionally; the resolver runs only the subset relevant to whatever command just finished.

## Suggestion Output Format

Every next-step suggestion follows the same format:

```
Next:  <primary command>  -- <description>
       <secondary command>  -- <description>
```

Ground rules:
- **One primary command** -- the single most useful next action.
- **At most one secondary** -- an alternate or follow-up.
- **Each line** is `<command>  -- <description>` (two spaces before `--`).
- **Multi-agent projects** repeat per agent on separate lines. No selection menus.
- **`Next:` label** is always present so suggestions stand out visually from command output.
- **README/docs pointers** go on a separate line below the commands when relevant.

## OpenAPI Discovery for Invoke Payloads

The decision tree handles "which command next" well, but the invoke payload problem -- "what do I actually *send* to this agent?" -- needs richer input than `azure.yaml` can provide.

### The Gap

Every sample has a different payload shape:
- Responses protocol: `{"input": "Hello!"}`
- Invocations protocol: `{"message": "Hello!"}` (or custom shapes)
- Custom protocols: anything goes

The CLI can infer protocol from `azure.yaml`, but the actual payload schema -- especially for invocations-protocol agents with custom request bodies -- lives in the agent's code, not config.

### What Already Exists

The invocation protocol already defines a well-known OpenAPI endpoint, and the extension already has code to consume it:

1. **Protocol convention:** Agents can serve an OpenAPI 3.0 spec at `GET /invocations/docs/openapi.json`. This is part of the invocation protocol -- not a new proposal.

2. **Extension support:** [`fetchOpenAPISpec` in helpers.go](https://github.com/Azure/azure-dev/blob/991b9743b4ab14ee8561677ffa627735c875a796/cli/azd/extensions/azure.ai.agents/internal/cmd/helpers.go#L296) already probes the well-known path, fetches the spec, and caches it to disk (`openapi-{agent}-{local|remote}.json` in the azd env directory). Failures are non-fatal and silently ignored.

3. **SDK support:** The Python SDK (`azure-ai-agentserver-invocations`) supports this via the `openapi_spec` parameter on `InvocationAgentServerHost`. See the [human-in-the-loop sample](https://github.com/microsoft-foundry/foundry-samples/blob/fd4ecab832fa491b9ffb4abca862c073777b5e53/samples/python/hosted-agents/bring-your-own/invocations/human-in-the-loop/main.py#L119) for a working example -- the agent defines an `OPENAPI_SPEC` dict describing its endpoints and request schemas, then passes it to `InvocationAgentServerHost(openapi_spec=OPENAPI_SPEC)`.

### Opt-In by the Agent Author

This is **not automatic**. The agent developer must define and include the OpenAPI spec for it to be served. The SDK does not generate it from handler types today. This means:

- **Samples we own:** We include OpenAPI specs by default in our samples, so the getting-started happy path always produces rich invoke examples. No extra work for developers using our templates.
- **Bring-your-own agents:** Developers who write their own agents can opt in by defining an `openapi_spec` and passing it to the server host. If they don't, the CLI falls back gracefully.

### CLI Behavior

After `azd ai agent run` starts the agent, the CLI probes `GET /invocations/docs/openapi.json`. Two paths:

1. **Spec available:** Parse it, grab the invoke endpoint and an example payload (from `example` or `schema` fields), and build the exact invoke command shown to the developer.

2. **Spec unavailable (404, timeout, etc.):** Fall back to protocol-generic suggestions (`"Hello!"` for responses, `'{"message": "Hello!"}'` for invocations) **plus a helpful hint**:

```
Tip:  curl http://localhost:8088/invocations/docs/openapi.json
      to check if your agent exposes an API spec with the exact payload format.
```

This nudges developers toward verifying the right payload shape for their agent's `/invocations` endpoint without requiring the spec to exist.

No cross-team dependencies remain. The protocol convention exists, the extension code exists, and samples can include specs today.

### Example: OpenAPI-Enhanced Output After `run`

Without OpenAPI (generic):
```
Agent is running at http://localhost:8088

  azd ai agent invoke --local "Hello!"
```

With OpenAPI (sample-specific):
```
Agent is running at http://localhost:8088

  azd ai agent invoke --local '{"message": "What hotels are available in Seattle?"}'

  Available endpoints:
    POST /invocations  -- send a message to the agent
    GET  /health       -- check agent health
```

### Command Coverage Matrix

| Command | Trigger point | Primary suggestion | State inputs used |
|---|---|---|---|
| `init` | Command exit | `run` or `azd provision` | HasUnresolvedInfraVars, HasUnresolvedManualVars |
| `run` | Agent starts listening | `invoke --local` | Protocol, HasOpenAPI |
| `invoke --local` | Command exit | `azd deploy` | -- |
| `invoke` (remote) | Command exit | `show` or `monitor --follow` | -- |
| `show` | Command exit | `invoke` or `monitor --follow` | AgentStatus |
| `deploy` (post-deploy hook) | Hook exit | `show` or `invoke` | -- |
| `doctor` | Command exit | Varies based on check results | All state inputs |

### Next-Step Resolver (core)

We already have two patterns to generalize: the `AZURE_AI_PROJECT_ID` check in [init.go](https://github.com/Azure/azure-dev/blob/main/cli/azd/extensions/azure.ai.agents/internal/cmd/init.go#L1718-L1733) (state-aware branching) and [`exterrors`](https://github.com/Azure/azure-dev/tree/main/cli/azd/extensions/azure.ai.agents/internal/exterrors) factories (structured error + suggestion). The resolver unifies both into a single function called at command completion points (command exit for most, startup-success for `run`):

```
func resolveNextSteps(cmd string, exitCode int, project ProjectState) []Suggestion
```

Where `ProjectState` is assembled by reading:
- `azure.yaml` (parsed once, cached)
- azd environment variables
- (optionally) a running agent's OpenAPI endpoint

Each `Suggestion` has:
- `command`: the CLI command to run (e.g., `azd ai agent run`)
- `description`: one-line explanation (e.g., `start the agent locally`)
- `priority`: ordering when multiple suggestions apply

The resolver outputs a formatted block:

```
func printNextSteps(suggestions []Suggestion) {
    // numbered list for multi-step, single line for one suggestion
}
```

### Where It Hooks In

The resolver fires at command completion -- `PostRun` for short-lived commands, the "listening" callback for `run`. No new CLI flags needed; it's always-on. If output is piped or we're in CI, guidance goes to stderr so it doesn't mess with stdout parsing.

## Downstream Impact

- **`azure.ai.agents` extension:** New `resolveNextSteps` function for success-path guidance, plus `doctor`. Both read `azure.yaml` and azd env -- already available in the command context.
- **Agent Server SDKs:** No changes needed. The invocation protocol already supports `GET /invocations/docs/openapi.json` and the Python SDK already accepts `openapi_spec` on `InvocationAgentServerHost`.
- **Samples:** Add `openapi_spec` definitions to our invocations-protocol samples so the getting-started happy path always produces rich invoke examples. Responses-protocol samples need no changes.
- **Foundry Toolkit for VS Code:** No impact. This is CLI output only.

### AZD Hooks for Sample-Specific Guidance

Instead of the CLI reading state and inferring suggestions, each sample ships hook scripts (`postinit.sh`, `postrun.sh`) that echo guidance to the terminal.

What's good about this:
- Hooks are first-class in azd. No new mechanism.
- Sample authors control the messaging.

Why I don't think it works:
- Spawning a shell to echo 3 lines is heavy, and now you need platform-specific scripts (sh vs ps1).
- Hooks fire on every run with no opt-out. Re-running a command re-fires the guidance.
- Hooks can't inspect deep project state (unresolved env vars, sibling service status) without reimplementing the resolver in every sample's hook script. That's duplication, not reuse.
- The state-aware decision tree has to live in the CLI anyway (for non-sample-specific stuff). Hooks would create two parallel guidance systems.

### README Markers

Standardized HTML-comment markers in each sample's README (e.g., `<!-- section:azd:after-init -->`) that the CLI parses and displays.

What's good about this:
- Declarative, no script execution.
- Sample authors write guidance once; README readers and the CLI both benefit.

Why I don't think it works:
- Parsing HTML comments from markdown is fragile and introduces a convention that doesn't exist anywhere else in azd.
- Markers are static -- they can't adapt to actual project state (provisioned? deployed? env vars resolved?).
- OpenAPI at `/invocations/docs/openapi.json` already handles the invoke payload problem, which is the main gap markers would fill.
- For non-payload guidance (e.g., "run azd up first"), the built-in decision tree handles it without needing README content.

### Static Hints in `azure.yaml`

A `config.samples` block in `azure.yaml` where sample authors declare example payloads:

```yaml
config:
  samples:
    invoke: '{"message": "What hotels are in Seattle?"}'
```

What's good about this:
- Lives in a file the CLI already reads. No new parsing.
- Declarative and simple.

Why I don't think it works:
- Duplicates info that already exists in agent code (and in the OpenAPI spec at `/invocations/docs/openapi.json` when the agent includes one).
- Complex payloads make YAML quoting painful.
- Doesn't scale to agents with multiple endpoints or evolving request schemas.
- The config section is intended to match the API.
- OpenAPI at `/invocations/docs/openapi.json` is strictly more capable and already available.

## Scope Boundaries

**In the CLI (next-step resolver):** State-aware command suggestions on success paths, driven by azure.yaml structure, azd env state, and exit status. Deterministic logic that every project benefits from.

**In the CLI (`doctor`):** Comprehensive project health checks -- auth, permissions, reachability, env var resolution, agent status. One pass, all the answers.

**In the Agent Server SDKs (no changes needed):** The invocation protocol already supports serving OpenAPI at `/invocations/docs/openapi.json`. The Python SDK already supports `openapi_spec` on `InvocationAgentServerHost`. Agent authors opt in by defining and passing a spec -- we include it by default in our samples.

**Already covered:** Error-path guidance. [`exterrors`](https://github.com/Azure/azure-dev/tree/main/cli/azd/extensions/azure.ai.agents/internal/exterrors) handles this across the extension. We don't touch it here.

**Out of scope:** CI/CD pipeline guidance. Multi-agent orchestration. Error message formatting (that's [`exterrors`](https://github.com/Azure/azure-dev/tree/main/cli/azd/extensions/azure.ai.agents/internal/exterrors) territory and a future Thread 10 audit).

antriksh30:
Fixes #7975  initial draft design doc  context-aware `Next:` guidance after every successful `azd ai agent` command, plus a new `azd ai agent doctor` diagnostic.

**Scope:** entirely inside `cli/azd/extensions/azure.ai.agents/`. References core files for context but proposes no edits outside the extension.

**Doc covers:**
- Background & today's gaps (per-command table)
- Architecture: new `nextstep` package + thin command callers
- `State` model with layered sources (flags > runtime probes > azure.yaml > azd env)
- Per-command decision trees (init / run / invoke / show / deploy / doctor)
- Output format & TTY/pipe/JSON discipline
- Hook points per command (incl. why deploy uses the `azdext.Artifact` SDK field)
- OpenAPI invoke-payload extraction order
- Doctor checks 16 (MVP) + 711 deferred
- Testing strategy, risks, phased rollout

### DIFF
--- a/cli/azd/docs/design/azd-ai-agent-doctor-remote-checks.md
+++ b/cli/azd/docs/design/azd-ai-agent-doctor-remote-checks.md
@@ -1,0 +1,260 @@
+# Design: `azd ai agent doctor` — Remote Checks (7–11)
+
+## Status
+
+**Draft — for design review.**
+Tracking issue: [#7975](https://github.com/Azure/azure-dev/issues/7975).
+Companion to [`azd-ai-agent-nextsteps.md`](./azd-ai-agent-nextsteps.md), which covers the doctor command's plumbing and local checks (1–6). **All code lives under `cli/azd/extensions/azure.ai.agents/`. No files outside the extension are modified.**
+
+## Goal
+
+Extend `azd ai agent doctor` past its local-only MVP so it can answer the question every developer eventually asks: *"the project is wired up correctly, so why doesn't this work end-to-end?"*
+
+Concretely: add five checks that exercise the live Azure surface — auth, Foundry reachability, model deployments, RBAC, agent runtime status — each with a precise "what to do next" fix.
+
+Non-goals:
+- Fixing remote problems automatically (`doctor --fix`). Surfacing the right command is enough for v1.
+- Re-implementing what `az` / Foundry portal already do well — we just point the user there when relevant.
+- Any modification to files outside `cli/azd/extensions/azure.ai.agents/`.
+
+## Background
+
+The MVP doctor checks (1–6) are pure local reads — `azure.yaml` exists, env vars are set, YAML parses. They run in tens of milliseconds and never need network or auth. They catch ~70% of "why is my project broken" questions.
+
+The remaining ~30% — and arguably the most frustrating cases — sit in remote land:
+- Token expired silently → every command 401s with no obvious "you're logged out" hint.
+- Foundry endpoint typo'd in env → DNS/TLS error buried in a stack trace.
+- Model deployment got deleted out from under the project → first invoke fails with a cryptic 404.
+- Azure-AI Developer role missing → vague 403.
+- Agent deployed but unhealthy → user thinks `azd deploy` succeeded so it must be fine.
+
+These all require live calls and credentials, which is why they're carved out into this separate design.
+
+## Scope of this doc
+
+| Check | What it answers |
+|---|---|
+| 7 — Authentication | Am I logged in? Is the token still valid? |
+| 8 — Foundry reachability | Can I reach `AZURE_AI_PROJECT_ENDPOINT`? |
+| 9 — Model deployments | Do the model(s) my agent references exist on the project? |
+| 10 — RBAC | Does my principal have the roles needed to deploy / invoke? |
+| 11 — Agent status | Does the deployed agent actually exist and is it active? |
+
+## Architecture
+
+### Component layout (additive only)
+
+```
+cli/azd/extensions/azure.ai.agents/internal/cmd/
+└── doctor/
+    ├── checks_local.go       ← (MVP) checks 1–6
+    ├── checks_remote.go      ← NEW: checks 7–11
+    ├── runner.go             ← orchestrates ordered execution
+    └── types.go              ← Check, Result, Severity
+```
+
+Existing MVP files unchanged. Only `runner.go` learns to dispatch the new check set, gated by a flag (see [Execution Model](#execution-model)).
+
+### `Check` contract
+
+Same shape as MVP — keeps the renderer, sort order, and `Next:` resolver code path identical.
+
+```go
+type Check interface {
+    ID() string                   // stable, e.g. "auth.token"
+    Title() string                // human label
+    Run(ctx context.Context) Result
+}
+
+type Result struct {
+    Status   Status               // Pass | Warn | Fail | Skip
+    Detail   string               // one-line description
+    Fix      string               // single command or instruction; empty for Pass
+    Links    []string             // doc links (optional)
+    Duration time.Duration        // captured by runner, not the check
+}
+```
+
+A remote check that can't run (e.g., not authenticated when running check 9) returns `Status: Skip` with `Detail: "skipped: authentication required (see check 7)"`. Skips never block subsequent checks — the user sees the full picture in one pass.
+
+## Execution Model
+
+### Default: local-only
+
+`azd ai agent doctor` runs checks 1–6 only. Same as MVP. No network, no token, no surprise prompts. This is the daily-driver mode.
+
+### Opt-in: `--remote`
+
+`azd ai agent doctor --remote` runs **1–11** in order. Local checks always run first; remote checks run only if their preconditions pass (see [Dependency Matrix](#dependency-matrix)).
+
+Rationale for opt-in (vs. always-on or auto-detect):
+- Remote checks need credentials and add seconds to a command users may run frequently.
+- Hitting Foundry on every `doctor` invocation is bad citizenship and surprising in CI.
+- A flag is discoverable, reversible, and self-documenting.
+
+We considered `--quick` / `--full` naming but `--remote` describes the actual side effect — network calls under your credentials — which is what reviewers and CI authors care about.
+
+### Dependency matrix
+
+| Check | Depends on | Action if dependency fails |
+|---|---|---|
+| 7 — Auth | check 3 (env selected) | Skip with "select an env first" |
+| 8 — Reachability | 5 (`AZURE_AI_PROJECT_ENDPOINT` set), 7 (auth) | Skip |
+| 9 — Models | 6 (`agent.yaml` valid), 8 (reachability) | Skip |
+| 10 — RBAC | 7 (auth), 8 (reachability) | Skip |
+| 11 — Agent status | 8, 9, 10 | Skip |
+
+Skips are explicit in the rendered report. We never quietly drop a check.
+
+## Check Specifications
+
+### Check 7 — Authentication
+
+**What it does:** acquires a token via `azidentity.NewAzureDeveloperCLICredential` (the same credential `agent_context.go` already uses), then introspects expiry.
+
+**Pass:** "<UPN> · token valid for <N> minutes"
+**Warn:** token expires in <5 min (suggest re-login proactively)
+**Fail:** token acquisition error → `azd auth login`
+
+**Why a separate fail vs warn:** users hit this all the time when a CLI session outlives a token; the warn variant tells them to re-login *before* a long-running deploy 401s.
+
+### Check 8 — Foundry project reachability
+
+**What it does:** issues a single `GET <AZURE_AI_PROJECT_ENDPOINT>/agents?api-version=…&$top=1` with the credential from check 7.
+
+**Pass:** "endpoint reachable (HTTP 200)"
+**Fail:** maps the HTTP status to one of:
+- `403` → check 10 (RBAC) will explain. Fix: see check 10 output.
+- `404` → endpoint is wrong or project is gone. Fix: `azd provision` or fix env var.
+- network/DNS/TLS → "verify VPN / firewall / typo in `AZURE_AI_PROJECT_ENDPOINT`".
+
+The single-shot probe avoids paging and works on tiny test projects. Timeout: 10s, no retries — doctor is diagnostic, not resilient.
+
+### Check 9 — Model deployments
+
+**What it does:** parses each service's `agent.yaml` for model references, then queries the Foundry project's deployments list once and matches names locally.
+
+**Pass:** "all <N> referenced models present"
+**Fail:** "missing: <model-name> (referenced by <service>)" → `azd provision` (or in the rare case the user deleted a deployment manually, "redeploy via Foundry portal" with link).
+
+**Edge cases:**
+- Multi-version models (e.g., `gpt-4o:2024-08-06`) — we match on deployment name, not version. Version mismatch surfaces at runtime; not in scope here.
+- Agents with no model reference (orchestration-only) — pass with "no model references".
+
+### Check 10 — RBAC
+
+**What it does:** for the current principal, lists role assignments on the Foundry project scope and checks for the minimum role set:
+- `Azure AI Developer` (or stronger) on the project — required for deploy + invoke.
+- `Cognitive Services User` on the AI account — required for model usage.
+
+**Pass:** "<role-1>, <role-2>"
+**Warn:** has invoke role but not deploy role → user can run agents but not `azd deploy`. Surfaces as "you can invoke but not deploy" with the exact `az role assignment create` command for the missing role.
+**Fail:** missing both → command + portal link.
+
+The fix command is templated with the actual principal ID and scope so the user can paste it directly. Example:
+```
+fix:  az role assignment create \
+        --role "Azure AI Developer" \
+        --assignee <principal-id> \
+        --scope <project-resource-id>
+```
+
+**Why we don't auto-create the assignment:** doctor is read-only. RBAC changes are the kind of action that should be explicit and audited.
+
+### Check 11 — Agent status
+
+**What it does:** for each service in `azure.yaml` with `host: ai-agent`, calls the Foundry agents-list endpoint and finds the matching agent by name.
+
+**Pass:** "<agent-name>: active (v<N>)"
+**Warn:** "<agent-name>: deploying (v<N>)" — running `monitor --follow` is suggested.
+**Fail:**
+- agent not found → `azd deploy`
+- agent in failed state → `azd ai agent monitor --follow` (look at logs, not redeploy blindly)
+
+This is the check that closes the loop on "I deployed and it succeeded — why doesn't `invoke` work?" Often the answer is the agent is still rolling out; check 11 says so explicitly.
+
+## Output Format
+
+Same renderer as MVP. Skipped checks render with a distinct glyph so they're visible:
+
+```
+azd ai agent doctor --remote
+  ✓ PASS  azd reachable
+  ✓ PASS  azure.yaml valid
+  ✓ PASS  azd environment selected
+  ✓ PASS  agent service in azure.yaml
+  ✓ PASS  AZURE_AI_PROJECT_ENDPOINT set
+  ✓ PASS  agent.yaml valid
+  ✓ PASS  authentication
+          alice@contoso.com · token valid for 47 minutes
+  ✗ FAIL  Foundry project reachability
+          GET /agents → 403
+          fix:  see RBAC check below
+  -  SKIP  model deployments
+          skipped: reachability check failed
+  ✗ FAIL  RBAC
+          principal alice@contoso.com has no role on project scope
+          fix:  az role assignment create --role "Azure AI Developer" \
+                  --assignee 0000... --scope /subscriptions/.../projects/...
+  -  SKIP  agent status
+          skipped: reachability check failed
+
+Next:  az role assignment create ...     -- grant yourself the required role
+       azd ai agent doctor --remote      -- re-run to verify
+```
+
+The trailing `Next:` block is resolved from the **first** failed remote check that has a single user-actionable fix — same logic as MVP, just extended to know about remote-check fix commands.
+
+## Performance & Side Effects
+
+Per [Performance Budget](#performance-budget) below — defaults remain local-only so users can keep running `doctor` reflexively. `--remote` adds a couple of seconds and explicit credential use.
+
+| Check | Network calls | Worst-case time |
+|---|---|---|
+| 7 — Auth | 0 (cached token) – 1 (refresh) | 2s |
+| 8 — Reach | 1 GET | 10s (timeout) |
+| 9 — Models | 1 list | 5s |
+| 10 — RBAC | 1 list-role-assignments | 5s |
+| 11 — Status | 1 list-agents | 5s |
+
+Worst-case `--remote` walltime: ~30s with a sick endpoint. Each check has its own timeout, applied via `context.WithTimeout`. The runner reports per-check duration in `--verbose` mode.
+
+## Testing Strategy
+
+- **Unit tests per check** with a fake `agentClient` interface. Cover pass / warn / fail / skip paths and HTTP status mapping (especially 401/403/404 disambiguation in check 8).
+- **Snapshot test** for the rendered report in mixed pass/fail/skip configurations — same harness as MVP doctor's local-checks snapshot test.
+- **Functional test** (recorded cassette via `mage record`) covering one happy-path full run.
+- No live tests in CI for `--remote` — replay-only. Cassette refresh is a manual maintainer step.
+
+## Risks & Mitigations
+
+| Risk | Mitigation |
+|---|---|
+| Auth prompt during `--remote` surprises users in CI | Detect non-interactive (`!isTerminal`) and refuse the auth refresh; emit a clear "skipped: non-interactive" instead of hanging. |
+| RBAC list call requires a permission users may also lack | If the role-assignments call itself 403s, render the check as Warn, not Fail — give the user the role-assignment command anyway and let them try. |
+| Foundry API drift breaks the reachability probe | Pin the api-version explicitly. Failure renders as "API not understood" with a doc link rather than a stack trace. |
+| Slow tenant or VPN turns `--remote` into a 30s wait | Strict per-check timeouts, parallel fan-out for independent checks (8 + 10 can run concurrently). |
+| User runs `--remote` against a fresh project before `azd provision` | Checks 8–11 short-circuit to Skip via the dependency matrix; check 5 (env var missing) tells them to provision. |
+
+## Performance Budget
+
+Local-only `doctor` (today): <100ms. Don't regress.
+`--remote`: ≤ 30s walltime in the worst case (sick endpoint). Typical: <5s.
+We do **not** add any background polling, prefetching, or daemon work for these checks. Doctor remains a one-shot, opt-in diagnostic.
+
+## Implementation Phases
+
+1. **Runner refactor** — split `checks_local.go` / `checks_remote.go`, add `--remote` flag, no behavior change at this step.
+2. **Auth + reachability (7, 8)** — the two checks every other remote check depends on. Useful even on their own.
+3. **Models + agent status (9, 11)** — leverage existing `agentClient` calls. Closes the deploy-loop questions.
+4. **RBAC (10)** — last because the role-list API is the trickiest to stub and the most likely to need iteration on the fix-command wording.
+
+Each phase is independently shippable and behind the `--remote` flag, so partial rollout is safe. We can ship phases 1–2 and gather feedback before completing 3–4.
+
+## Appendix: Why not fold these into MVP?
+
+Three reasons, in order of weight:
+
+1. **Auth + network coupling.** MVP doctor runs in any project state, including before `azd auth login`. Remote checks fundamentally cannot. The flag separation makes the contract explicit.
+2. **Review surface.** The MVP design (the companion doc) is already large. Adding five live-Azure checks with their own error mapping, timeout, and skip semantics doubles it.
+3. **Iteration cost on Foundry surfaces.** The Foundry agents API is still evolving; we want to land remote checks behind a flag so we can iterate the wording, the role-set, and the api-version pinning without churning the always-on path.
--- a/cli/azd/docs/design/azd-ai-agent-doctor-remote-checks.md
+++ b/cli/azd/docs/design/azd-ai-agent-doctor-remote-checks.md
@@ -1,4 +1,4 @@
-# Design: `azd ai agent doctor` — Remote Checks (7–11)
+# Design: `azd ai agent doctor` — Remote Checks (7–12)
 
 ## Status
 
@@ -6,11 +6,19 @@
 Tracking issue: [#7975](https://github.com/Azure/azure-dev/issues/7975).
 Companion to [`azd-ai-agent-nextsteps.md`](./azd-ai-agent-nextsteps.md), which covers the doctor command's plumbing and local checks (1–6). **All code lives under `cli/azd/extensions/azure.ai.agents/`. No files outside the extension are modified.**
 
+## Terminology
+
+- **MVP doctor** — the first shippable slice of the command: local checks (1–6) only. Defined in the [companion doc](./azd-ai-agent-nextsteps.md). MVP is a chronological term ("what lands first"), not a permanent capability boundary.
+- **Local checks** — read-only inspection of `azure.yaml`, `agent.yaml`, env vars, etc. No network, no token.
+- **Remote checks** — anything that issues a network call under user credentials. Specified in this doc.
+
 ## Goal
 
 Extend `azd ai agent doctor` past its local-only MVP so it can answer the question every developer eventually asks: *"the project is wired up correctly, so why doesn't this work end-to-end?"*
 
-Concretely: add five checks that exercise the live Azure surface — auth, Foundry reachability, model deployments, RBAC, agent runtime status — each with a precise "what to do next" fix.
+The end-state default is to run **all** checks (local then remote) on every `doctor` invocation. CI and air-gapped users opt out of the remote ones via `--local-only` (see [Execution Model](#execution-model)).
+
+Concretely: add six checks that exercise the live Azure surface — auth, Foundry reachability, model deployments, user RBAC, agent runtime status, agent-identity role assignments — each with a precise "what to do next" fix.
 
 Non-goals:
 - Fixing remote problems automatically (`doctor --fix`). Surfacing the right command is enough for v1.
@@ -19,9 +27,9 @@ Non-goals:
 
 ## Background
 
-The MVP doctor checks (1–6) are pure local reads — `azure.yaml` exists, env vars are set, YAML parses. They run in tens of milliseconds and never need network or auth. They catch ~70% of "why is my project broken" questions.
+Local checks (1–6) — `azure.yaml` exists, env vars are set, YAML parses — run in tens of milliseconds and never need network or auth. They catch ~70% of "why is my project broken" questions and are what MVP doctor ships first.
 
-The remaining ~30% — and arguably the most frustrating cases — sit in remote land:
+The remaining ~30% — and arguably the most frustrating cases — *intentionally* require network calls under the user's credentials:
 - Token expired silently → every command 401s with no obvious "you're logged out" hint.
 - Foundry endpoint typo'd in env → DNS/TLS error buried in a stack trace.
 - Model deployment got deleted out from under the project → first invoke fails with a cryptic 404.
@@ -37,8 +45,9 @@ These all require live calls and credentials, which is why they're carved out in
 | 7 — Authentication | Am I logged in? Is the token still valid? |
 | 8 — Foundry reachability | Can I reach `AZURE_AI_PROJECT_ENDPOINT`? |
 | 9 — Model deployments | Do the model(s) my agent references exist on the project? |
-| 10 — RBAC | Does my principal have the roles needed to deploy / invoke? |
+| 10 — User RBAC | Does my principal have the roles needed to deploy / invoke? |
 | 11 — Agent status | Does the deployed agent actually exist and is it active? |
+| 12 — Agent identity roles | What roles does the agent's managed identity have, and on which scopes? |
 
 ## Architecture
 
@@ -79,20 +88,20 @@ A remote check that can't run (e.g., not authenticated when running check 9) ret
 
 ## Execution Model
 
-### Default: local-only
+### Default: all checks (1–12)
 
-`azd ai agent doctor` runs checks 1–6 only. Same as MVP. No network, no token, no surprise prompts. This is the daily-driver mode.
+`azd ai agent doctor` runs the full check set in order. Local checks (1–6) run first; remote checks (7–12) run only if their preconditions pass (see [Dependency Matrix](#dependency-matrix)). The default is "tell me everything that's wrong" — a half-answer that skips the most frustrating cases isn't worth shipping.
 
-### Opt-in: `--remote`
+### Opt-out: `--local-only`
 
-`azd ai agent doctor --remote` runs **1–11** in order. Local checks always run first; remote checks run only if their preconditions pass (see [Dependency Matrix](#dependency-matrix)).
+`azd ai agent doctor --local-only` skips checks 7–12. Use this when:
+- Running in CI / non-interactive shells where an interactive auth prompt would hang.
+- Working offline / behind a restrictive firewall.
+- Investigating a config-only issue and you don't want network latency.
 
-Rationale for opt-in (vs. always-on or auto-detect):
-- Remote checks need credentials and add seconds to a command users may run frequently.
-- Hitting Foundry on every `doctor` invocation is bad citizenship and surprising in CI.
-- A flag is discoverable, reversible, and self-documenting.
+Non-interactive detection (`!isTerminal`) is **not** automatic — explicit is better than magic. If a user forgets `--local-only` in CI, the auth check fails fast with a clear message and the rest of the remote checks Skip cleanly (see [Risks](#risks--mitigations)).
 
-We considered `--quick` / `--full` naming but `--remote` describes the actual side effect — network calls under your credentials — which is what reviewers and CI authors care about.
+We considered `--remote` opt-in but rejected it: doctor's value is highest when it answers the live-Azure questions by default, and every "why doesn't this work" question we've seen in practice falls in that bucket. CI is the niche, not the default.
 
 ### Dependency matrix
 
@@ -103,6 +112,7 @@ We considered `--quick` / `--full` naming but `--remote` describes the actual si
 | 9 — Models | 6 (service stanza valid), 8 (reachability) | Skip |
 | 10 — RBAC | 7 (auth) | Skip — RBAC reads ARM, not Foundry data plane, so it does **not** depend on check 8 |
 | 11 — Agent status | 7 (auth), 8 (reachability) | Skip — agents-list is a Reader-level Foundry call and does not require check 10's deploy/invoke roles |
+| 12 — Agent identity roles | 11 (agent status pass — we need the agent's MI principal ID) | Skip — without an active agent there is no MI to introspect |
 
 Skips are explicit in the rendered report. We never quietly drop a check.
 
@@ -135,6 +145,8 @@ The single-shot probe avoids paging and works on tiny test projects. Timeout: 10
 
 **What it does:** parses each service's `agent.yaml` for model references, then queries the Foundry project's deployments list once and matches names locally.
 
+> **Tracking note ([#7962](https://github.com/Azure/azure-dev/issues/7962)).** Once the `agent.yaml` → `azure.yaml` unification lands, this check (and any other reference to `agent.yaml` in this doc) reads model refs from `azure.yaml` instead. Syncing with @trangevi on timing.
+
 **Pass:** "all <N> referenced models present"
 **Fail:** "missing: <model-name> (referenced by <service>)" → `azd provision` (or in the rare case the user deleted a deployment manually, "redeploy via Foundry portal" with link).
 
@@ -142,7 +154,7 @@ The single-shot probe avoids paging and works on tiny test projects. Timeout: 10
 - Multi-version models (e.g., `gpt-4o:2024-08-06`) — we match on deployment name, not version. Version mismatch surfaces at runtime; not in scope here.
 - Agents with no model reference (orchestration-only) — pass with "no model references".
 
-### Check 10 — RBAC
+### Check 10 — User RBAC
 
 **What it does:** for the current principal, lists role assignments on the Foundry project scope and checks for the minimum role set:
 - `Azure AI Developer` (or stronger) on the project — required for deploy + invoke.
@@ -176,12 +188,44 @@ fix:  az role assignment create \
 
 This is the check that closes the loop on "I deployed and it succeeded — why doesn't `invoke` work?" Often the answer is the agent is still rolling out; check 11 says so explicitly.
 
+### Check 12 — Agent identity role assignments
+
+**What it does:** for each active agent found in check 11, resolves the agent's managed-identity principal ID, then lists role assignments on the **agent identity** at three scope levels:
+
+1. The Foundry project (scope of the agent itself).
+2. The parent AI account.
+3. The resource group containing the project (proxy for "any resource the agent might touch").
+
+We don't try to predict *which specific resource* the agent will need at invocation time — that's data-plane and varies per agent code. Instead we surface the visible role-assignment landscape so the user can answer "does my agent identity have any roles where I'd expect it to?"
+
+**Pass:** agent MI has at least one assignment at the project scope, plus a non-empty list at account or resource-group scope.
+**Warn:** agent MI has assignments only at the agent's own scope — it almost certainly cannot read project / account resources at runtime. Render the list with a "this looks under-privileged for typical agents" hint.
+**Fail:** agent MI has zero role assignments anywhere reachable. This is the smoking-gun for "my agent runs but every tool call 403s."
+
+**Output shape:**
+```
+ⓘ INFO  agent identity roles  (agent: research-bot)
+        principal: <mi-principal-id>
+        project scope:
+          - Cognitive Services User
+        account scope:
+          - (none)
+        resource-group scope:
+          - Storage Blob Data Reader
+```
+
+We render this as `INFO` (a fourth glyph alongside pass/warn/fail/skip) when there's nothing actionable to flag, since the value is mostly informational — the user inspects the list to confirm it matches their mental model. Warn/Fail variants surface only the actionable cases above.
+
+**Why a separate check from 10:** check 10 is about the *user's* permissions to deploy/invoke; check 12 is about the *agent's* permissions at runtime. They fail differently, fix differently (`az role assignment create --assignee <mi-id>` vs `<user-id>`), and the agent MI doesn't exist until check 11 has confirmed an active agent — so they can't be merged without losing precision.
+
+**Redaction.** Same rules as check 10: principal IDs and scope ARNs are replaced with `<redacted>` in non-interactive output, and the JSON envelope sets `"redacted": true`.
+
 ## Output Format
 
 Same renderer as MVP. Skipped checks render with a distinct glyph so they're visible:
 
 ```
-azd ai agent doctor --remote
+azd ai agent doctor
   ✓ PASS  azd reachable
   ✓ PASS  azure.yaml valid
   ✓ PASS  azd environment selected
@@ -195,32 +239,35 @@ azd ai agent doctor --remote
           fix:  see RBAC check below
   -  SKIP  model deployments
           skipped: reachability check failed
-  ✗ FAIL  RBAC
+  ✗ FAIL  user RBAC
           principal alice@contoso.com has no role on project scope
           fix:  az role assignment create --role "Azure AI Developer" \
                   --assignee 0000... --scope /subscriptions/.../projects/...
   -  SKIP  agent status
           skipped: reachability check failed
+  -  SKIP  agent identity roles
+          skipped: agent status check did not pass
 
 Next:  az role assignment create ...     -- grant yourself the required role
-       azd ai agent doctor --remote      -- re-run to verify
+       azd ai agent doctor               -- re-run to verify
 ```
 
 The trailing `Next:` block is resolved from the **first** failed remote check that has a single user-actionable fix — same logic as MVP, just extended to know about remote-check fix commands.
 
 ## Performance & Side Effects
 
-Per [Performance Budget](#performance-budget) below — defaults remain local-only so users can keep running `doctor` reflexively. `--remote` adds a couple of seconds and explicit credential use.
+Per [Performance Budget](#performance-budget) below — `--local-only` keeps doctor under 100ms for users who explicitly opt out. Default mode adds a few seconds and explicit credential use.
 
 | Check | Network calls | Worst-case time |
 |---|---|---|
 | 7 — Auth | 0 (cached token) – 1 (refresh) | 2s |
 | 8 — Reach | 1 GET | 10s (timeout) |
 | 9 — Models | 1 list | 5s |
-| 10 — RBAC | 1 list-role-assignments | 5s |
+| 10 — User RBAC | 1 list-role-assignments | 5s |
 | 11 — Status | 1 list-agents | 5s |
+| 12 — Agent MI roles | 3 list-role-assignments (project / account / RG) | 6s |
 
-Worst-case `--remote` walltime: ~30s with a sick endpoint. Each check has its own timeout, applied via `context.WithTimeout`. The runner reports per-check duration in `--verbose` mode.
+Worst-case full-sweep walltime: ~35s with a sick endpoint. Each check has its own timeout, applied via `context.WithTimeout`. The runner reports per-check duration in `--verbose` mode.
 
 ## Testing Strategy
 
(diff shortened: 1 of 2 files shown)


wbreza (commented):
## 🔍 Design Doc Review

Thanks for the thorough design, @antriksh30! Strong technical depth with excellent extension-boundary discipline. Below are findings from a multi-perspective review (architecture, PM, testing, documentation).

---

### P0 — Must Fix Before Merge (5)

**1. RBAC principal ID + subscription path leakage**
📄 zd-ai-agent-doctor-remote-checks.md — Check 10

Check 10's fix command templates --assignee <principal-id> and --scope <ARM-path> into stdout. In CI, these end up in persistent build logs — a credential-adjacent leak. The design needs a security-aware output contract: redact in non-interactive mode, or route sensitive fix commands to --output json only.

**2. Check 7 auth mechanism contradicts between documents**
📄 Both docs — Check 7

zd-ai-agent-nextsteps.md says z account show; zd-ai-agent-doctor-remote-checks.md says zidentity.NewAzureDeveloperCLICredential. These are different mechanisms with different failure modes. Suggest aligning on AzureDeveloperCLICredential since this is an zd extension.

**3. Parallel fan-out contradicts dependency matrix**
📄 zd-ai-agent-doctor-remote-checks.md — Risks table vs Dependency matrix

Risks table claims checks 8+10 can run concurrently, but the dependency matrix shows check 10 depends on check 8. Both can't be true. Either remove 8 as a dependency of 10 (RBAC calls ARM, not Foundry) or correct the risks table.

**4. Security considerations section missing entirely**
📄 zd-ai-agent-doctor-remote-checks.md

A feature handling Azure credentials, UPN display, and CI output needs an explicit security/threat section covering: what sensitive data appears in output, behavior in CI/non-interactive contexts, and telemetry exclusions for PII fields.

**5. gent.yaml assumption contradicts issue #7975**
📄 zd-ai-agent-nextsteps.md

Issue #7975 states "assumes unified zure.yaml" (agent.yaml gone), but the design references gent.yaml throughout. Add an "Assumptions" section clarifying which config schema the design targets and what happens when unification lands.

---

### P1 — Should Fix (17)

<details>
<summary>Expand P1 findings</summary>

| # | File | Finding |
|---|------|---------|
| 6 | nextsteps | **State model missing IsAuthenticated field** — resolver can't distinguish auth states; implementer will hit this mid-build |
| 7 | remote-checks | **Check 11 dependency on check 10 too strict** — Reader-role users get agent status skipped unnecessarily; consider removing check 10 from 11's deps |
| 8 | remote-checks | **Check 8 conflates auth failure with reachability** — 403 could be wrong-tenant, not missing RBAC; update fail message to mention both possibilities |
| 9 | nextsteps | **Stray trailing triple-backtick** at end of file corrupts markdown rendering |
| 10 | nextsteps | **Duplicate "Multi-line note pre-indentation" heading** — copy-paste error, remove duplicate |
| 11 | nextsteps | **Broken #alternatives-considered anchor** — links to a section that doesn't exist; add the section or remove the link |
| 12 | nextsteps | **Backward compatibility not addressed** — new Next: stdout output could break scripts; add a "Backward Compatibility" section |
| 13 | remote-checks | **No --output json for doctor --remote** — CI consumers need machine-readable doctor results |
| 14 | both | **No doctor exit codes specified** — CI can't gate on doctor (0=pass, 1=fail, 2=skip) |
| 15 | both | **No test strategy for TTY/pipe/JSON output routing** — core correctness contract needs test coverage |
| 16 | both | **No test for non-interactive skip path** — isTerminal detection failure in CI = hanging process |
| 17 | remote-checks | **Runner skip-cascade testing missing** — dependency matrix logic is the most complex part |
| 18 | both | **Doctor file placement inconsistency** — nextsteps says doctor.go, remote-checks says doctor/ subpackage |
| 19 | nextsteps | **No user personas** — design can't distinguish new developer from CI platform engineer |
| 20 | nextsteps | **No success metrics/KPIs** — no way to measure feature effectiveness |
| 21 | both | **No telemetry/observability plan** — check results should be instrumented for iteration |
| 22 | remote-checks | **Foundry API version not specified** — "pin api-version" without naming which one |

</details>

---

### P2 — Nice to Have (9)

<details>
<summary>Expand P2 findings</summary>

- Missing newline at EOF (remote-checks)
- Orphaned period at start of line in deploy decision tree
- Relative links to extension source paths need verification
- Trailing whitespace on multiple lines
- Check 1 "extension running" is tautological — consider renaming to "gRPC channel health"
- Adversarial/inconsistent State inputs not in test strategy
- Multi-agent cap of 5 is unjustified — consider data-backed threshold or making configurable
- OpenAPI $ref resolution not mentioned in extraction logic
- Check 9 model version mismatch deferred — could be low-cost Warn since data is already fetched

</details>

---

### Strengths

- Extension-boundary discipline is excellent — no modifications outside the extension
- Decision trees are tabular and testable
- Artifact-note path for deploy is pragmatic and well-justified
- Phased implementation with independent shippability is the right call
- Risk tables are realistic, not boilerplate
- Cross-references between companion docs are symmetric and correct

cli/azd/docs/design/azd-ai-agent-nextsteps.md
```diff
@@ -0,0 +1,393 @@
...
+- Always present `Next:` prefix so the block is visually distinct.
+
+### Output discipline
+
+| Scenario | Destination |
+|---|---|
+| Interactive TTY | `stdout` |
+| Piped output / CI (detected via `isTerminal`) | `stderr` so it doesn't pollute parsable output |
+| `--output json` / structured output flag | Suppressed entirely |
+| Inside `azd deploy` | Attached to the returned `azdext.Artifact` (same SDK field already used for endpoint URLs) |
```
jongio: The claim "same SDK field already used for endpoint URLs" needs to specify the exact mechanism. The `Artifact` proto has `Kind`, `Location`, `LocationKind`, and `Metadata` - no `Note` field. Today's deploy artifacts use `ARTIFACT_KIND_ENDPOINT` + `Metadata["label"]` rendered as `- label: url`. How does a multi-line Next: block fit into this structure without changing core's renderer?

jongio (commented):
See my detailed review below (review #4232410333) for the full analysis. This first review was a test artifact that couldn't be deleted.

cli/azd/docs/design/azd-ai-agent-nextsteps.md
```diff
@@ -0,0 +1,393 @@
...
+```
+
+### After `azd deploy` (rendered by core from artifact note)
+
+```
+  (✓) Done: Deploying service ag-ui-invocations
+  - Agent playground (portal):     https://ai.azure.com/...
+  - Agent endpoint (invocations):  https://....services.ai.azure.com/...
+    Next:  azd ai agent show ag-ui-invocations    -- verify it's running
+           azd ai agent invoke '<payload>'        -- test the deployment
```
jongio, line 491: Type mismatch between these two: `HasUnresolvedInfraVars bool` vs `HasUnresolvedManualVars []string`. If the resolver needs var names for manual vars (to emit `azd env set KEY VALUE`), it would benefit from knowing infra var names too (to say which Bicep outputs are pending). Consider making both `[]string` or adding a comment explaining why the asymmetry is deliberate.

cli/azd/docs/design/azd-ai-agent-doctor-remote-checks.md
```diff
@@ -0,0 +1,260 @@
...
+
+### Check 7 — Authentication
+
+**What it does:** acquires a token via `azidentity.NewAzureDeveloperCLICredential` (the same credential `agent_context.go` already uses), then introspects expiry.
+
+**Pass:** "<UPN> · token valid for <N> minutes"
+**Warn:** token expires in <5 min (suggest re-login proactively)
+**Fail:** token acquisition error → `azd auth login`
+
+**Why a separate fail vs warn:** users hit this all the time when a CLI session outlives a token; the warn variant tells them to re-login *before* a long-running deploy 401s.
```
jongio, line 131: +1 to @wbreza's point about auth mechanism contradiction. Verified: `agent_context.go` uses `azidentity.NewAzureDeveloperCLICredential` (not `az account show`). This doc correctly uses the azd credential path for checks 8-11, which is consistent with the codebase.

jongio (commented):
@wbreza already covered this thoroughly - adding a few things that review didn't flag.

**Artifact integration mechanism is underspecified (the most important gap)**

The doc repeatedly says "attach Next: to the artifact's note field" and "populate the artifact's note field" - but `azdext.Artifact` has no `Note` field. The proto defines: `Kind`, `Location`, `LocationKind`, and `Metadata` (map). Today's deploy artifacts are `ARTIFACT_KIND_ENDPOINT` with `Metadata["label"]` rendered as `- label: url`.

For the Next: block to work through this channel, the design needs to specify:
1. Which `ArtifactKind` carries freeform text? A new enum value like `ARTIFACT_KIND_NOTE`? Or reusing an existing one?
2. How does core's `MessageUxItem` renderer know to display multi-line text instead of a `- label: url` line?
3. If using `Metadata["note"]` - does core's rendering code already handle that key specially, or does it need changes (which contradicts the "no edits outside the extension" scope)?

This is the one integration point between the extension and core that isn't fully self-contained - and the design doc's scope claim ("no files outside the extension are modified") depends on getting this right.

**State model type inconsistency**

`HasUnresolvedInfraVars` is `bool` but `HasUnresolvedManualVars` is `[]string`. The init resolver needs manual var names to emit `azd env set <KEY> <value>`. But if infra vars are unresolved, knowing which Bicep outputs are missing would let the guidance be more specific than just "run `azd provision`".

**Deploy payload derivation gap**

The `run` decision tree has OpenAPI extraction logic for deriving the invoke payload. The `deploy` tree just says '<payload>' literally. Should the post-deploy Next: block attempt OpenAPI discovery against the remote endpoint, or reference the README? The `run` case and `deploy` case have asymmetric payload handling without explaining why.

cli/azd/docs/design/azd-ai-agent-nextsteps.md
```diff
@@ -0,0 +1,393 @@
+# Design: Context-Aware Next-Step Guidance for `azd ai agent`
+
+## Status
+
+**Draft — for design review.**
+Tracking issue: [#7975](https://github.com/Azure/azure-dev/issues/7975) — "Context-aware next-step guidance and diagnostics".
+Scope: the `azure.ai.agents` extension only. **All code lives under `cli/azd/extensions/azure.ai.agents/`. No files outside the extension are modified.** The deploy hook returns its Next: block via the standard extension SDK return value (`*azdext.Artifact`) — same path other extensions use today.
+
```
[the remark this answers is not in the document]
antriksh30: Reworded scope: "All **code** lives under `cli/azd/extensions/azure.ai.agents/`. No files outside the extension are modified." The design doc itself is documentation, not code.

cli/azd/docs/design/azd-ai-agent-nextsteps.md
```diff
@@ -0,0 +1,393 @@
...
+The resolver works with whatever's available — partial state never silences guidance:
+
+| Layer | Source | Authority |
+|---|---|---|
+| 1 | Explicit CLI flags (`--project-endpoint`, `--agent`) | Highest |
+| 2 | Live runtime probes (OpenAPI, Foundry status) | Override env vars |
+| 3 | `azure.yaml` services, protocols, `config.env` refs | Structural truth |
+| 4 | azd env vars | Resolution truth |
+
+If the azd environment is missing or stale (a real scenario for brownfield users), the resolver should still produce useful output from layers 1–3.
```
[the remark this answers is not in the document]
antriksh30: Added an Assumptions section: `agent.yaml` is the current source for protocol/env refs. `azure.yaml` references in the State model are placeholders for the unified service stanza per #7975  the model is unchanged either way.

cli/azd/docs/design/azd-ai-agent-nextsteps.md
```diff
@@ -0,0 +1,393 @@
...
+
+If OpenAPI fetch failed (404, timeout), append a Tip line:
+```
+Tip:  curl http://localhost:<port>/invocations/docs/openapi.json
+      to verify the exact payload format your agent expects.
+```
+
+### `invoke --local` (success)
+
+Single agent: `azd deploy`, secondary `azd ai agent monitor --follow`.
```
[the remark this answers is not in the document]
antriksh30: Dropped. `monitor` is hosted-only (verified `monitor.go:39`). The `invoke --local` success block now suggests a payload-file variant and a deploy hint instead.

cli/azd/docs/design/azd-ai-agent-nextsteps.md
```diff
@@ -0,0 +1,393 @@
...
+| `show` (success) | Nothing | No nudge to invoke or check logs. |
+| `deploy` | Endpoint URLs only | No invoke command, no link to README. |
+| (none) | — | No way to ask "where was I?" or "what's broken across the project?" |
+
+### What's already in place
+
+These are the building blocks the design reuses:
+
+- **`azdext` gRPC client** — extensions read `azure.yaml` services and azd env vars without re-implementing parsing.
+- **`fetchOpenAPISpec`** in `helpers.go` — already probes `GET /invocations/docs/openapi.json`, caches to disk, fails silently. Drives the rich-payload story.
```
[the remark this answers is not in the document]
antriksh30: Added a 'Silent-mode requirement' note in the OpenAPI Discovery section. The resolver path will gate the existing `OpenAPI spec saved to ...` print (`helpers.go:373`) behind a verbose flag.

cli/azd/docs/design/azd-ai-agent-nextsteps.md
```diff
@@ -0,0 +1,393 @@
...
+| 4 | Agent service in `azure.yaml` | service count + names | `azd ai agent init` |
+| 5 | `AZURE_AI_PROJECT_ENDPOINT` set | URL value | `azd provision` |
+| 6 | `agent.yaml` per service is valid | path | fix YAML |
+| 7 (post-MVP) | Authentication (`az account show`) | signed-in user | `azd auth login` |
+| 8 (post-MVP) | Foundry project reachable | endpoint probe OK | network/firewall |
+| 9 (post-MVP) | Model deployments exist | name + version | `azd provision` |
+| 10 (post-MVP) | RBAC sufficient | role list | `az role assignment create …` |
+| 11 (post-MVP) | Agent status (if deployed) | `active (vN)` | `azd ai agent monitor --follow` |
+
+Checks 7–11 are listed in the issue but pulled into a follow-up to keep the MVP shippable. Checks 1–6 are pure local reads; 7–11 require Foundry control-plane calls and RBAC introspection that need their own design pass.
```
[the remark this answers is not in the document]
antriksh30: Aligned. Check 7 now uses `AzureDeveloperCLICredential` / `azd auth login` to match the extension's data-plane auth (`agent_context.go:101-109`).

cli/azd/docs/design/azd-ai-agent-nextsteps.md
```diff
@@ -0,0 +1,393 @@
...
+
+### Checks (run in order, top to bottom)
+
+| # | Check | Pass | Fail → Fix |
+|---|---|---|---|
+| 1 | `azd` reachable + extension running | "extension running, gRPC channel established" | install / start daemon |
+| 2 | `azure.yaml` valid | "1 service: <names>" | `azd ai agent init` |
+| 3 | azd environment selected | "<env-name>" | `azd env new` / `azd env select` |
+| 4 | Agent service in `azure.yaml` | service count + names | `azd ai agent init` |
+| 5 | `AZURE_AI_PROJECT_ENDPOINT` set | URL value | `azd provision` |
```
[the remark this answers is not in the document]
antriksh30: Covered by the new Assumptions section  `agent.yaml` remains required in MVP and doctor's per-service validation reads it directly.

cli/azd/docs/design/azd-ai-agent-nextsteps.md
antriksh30: Anchored on `Metadata["note"]` on an `ArtifactKindEndpoint` artifact. `pkg/project/artifact.go` already renders the `note` metadata key as a continuation line under the artifact bullet  no core renderer change needed. Updated the doc with that citation and an example output.

cli/azd/docs/design/azd-ai-agent-nextsteps.md
antriksh30: Renamed to `MissingInfraVars []string` for symmetry with `MissingManualVars []string`. Both now expose names so the resolver can produce specific guidance (e.g., naming the missing Bicep outputs).

cli/azd/docs/design/azd-ai-agent-doctor-remote-checks.md
antriksh30: Thanks for verifying  that consistency was the goal. Both docs now use `AzureDeveloperCLICredential` end-to-end.

antriksh30:
Thanks @wbreza, ran a multi-model critique on the fix plan first then applied. Status of your list (commit 4358e6849):

**P0 (5/5):** all addressed — RBAC redaction (Check 10) ✅ · auth alignment to `AzureDeveloperCLICredential` ✅ · removed Check 8 from Check 10's deps + reconciled risks ✅ · new Security & PII / Telemetry section ✅ · new Assumptions section (`agent.yaml` is current; `azure.yaml` placeholders for unified stanza per #7975) ✅

**P1 (14/17 + 3 explicitly deferred):** `IsAuthenticated` → tri-state `AuthState` w/ `WithAuthProbe` opt-in ✅ · removed Check 10 from Check 11's deps ✅ · 403 broadened to "wrong tenant or insufficient RBAC" ✅ · stray backtick / duplicate heading verified clean ✅ · broken `#alternatives-considered` anchor removed ✅ · Backward Compatibility section ✅ · Exit codes (0/1/2) + JSON envelope ✅ · TTY/pipe/JSON test rows ✅ · skip-cascade + non-interactive skip tests ✅ · `doctor/` subpackage now consistent across both docs ✅ · api-version pinned to `DefaultAgentAPIVersion` constant (`agent_context.go:18`) ✅. Deferred (in new "Deferred follow-ups" section): personas, KPIs, full telemetry schema — telemetry sketch is in the new Telemetry subsection.

**P2 (4/9):** Check 1 → "gRPC channel health" ✅ · multi-agent cap-of-5 replaced ✅ · OpenAPI `$ref` explicit out-of-scope ✅ · adversarial state inputs implicit in expanded testing ✅. Cosmetic items (EOF newline, trailing whitespace, orphan period, link-verify, Check 9 model-warn) skipped — happy to do a follow-up cleanup pass if useful.

Plus jongio's three: anchored Status block on `Metadata["note"]` + `ArtifactKindEndpoint` (no core renderer change), renamed to `MissingInfraVars []string`, deploy resolver explicitly uses cached spec only (no live probe — agent may not be reachable yet, which is what Check 11 observes).

jongio (commented):
Addresses my previous feedback. The new commit clarifies all three gaps I flagged:

1. **Artifact mechanism** - now explicitly specifies `Metadata["note"]` on `ArtifactKindEndpoint`, references `pkg/project/artifact.go` `ToString` (~line 128). Verified this path exists and works as described.
2. **State model asymmetry** - renamed to `MissingInfraVars []string` and `MissingManualVars []string` (both `[]string` now).
3. **Deploy payload** - added "No live OpenAPI probe at deploy time" paragraph with the fallback chain (cached spec -> README -> generic literal).

Also good additions: `AuthState` enum with `AuthUnknown` default (avoids login noise), `WithAuthProbe` opt-in, exit code table, JSON envelope schema, redaction rules, security section, backward compatibility section, and the `$ref` out-of-scope note.

One minor observation: the `isTerminal` helper is referenced as living in `internal/cmd/helpers.go` - worth confirming the exact function name in a code comment so implementers don't have to grep for it.

cli/azd/docs/design/azd-ai-agent-doctor-remote-checks.md
```diff
@@ -0,0 +1,266 @@
...
+Concretely: add five checks that exercise the live Azure surface — auth, Foundry reachability, model deployments, RBAC, agent runtime status — each with a precise "what to do next" fix.
+
+Non-goals:
+- Fixing remote problems automatically (`doctor --fix`). Surfacing the right command is enough for v1.
+- Re-implementing what `az` / Foundry portal already do well — we just point the user there when relevant.
+- Any modification to files outside `cli/azd/extensions/azure.ai.agents/`.
+
+## Background
+
+The MVP doctor checks (1–6) are pure local reads — `azure.yaml` exists, env vars are set, YAML parses. They run in tens of milliseconds and never need network or auth. They catch ~70% of "why is my project broken" questions.
```
therealjohn: Some of the checks below, like is the project endpoint reachable, etc. are network calls. I think this sentence is too restrictive. We will need to do network calls to check the common issues.

cli/azd/docs/design/azd-ai-agent-doctor-remote-checks.md
```diff
@@ -0,0 +1,266 @@
...
+
+## Scope of this doc
+
+| Check | What it answers |
+|---|---|
+| 7 — Authentication | Am I logged in? Is the token still valid? |
+| 8 — Foundry reachability | Can I reach `AZURE_AI_PROJECT_ENDPOINT`? |
+| 9 — Model deployments | Do the model(s) my agent references exist on the project? |
+| 10 — RBAC | Does my principal have the roles needed to deploy / invoke? |
+| 11 — Agent status | Does the deployed agent actually exist and is it active? |
```
therealjohn, line 51: Here's an example of a network call.

cli/azd/docs/design/azd-ai-agent-doctor-remote-checks.md
```diff
@@ -0,0 +1,266 @@
...
+}
+```
+
+A remote check that can't run (e.g., not authenticated when running check 9) returns `Status: Skip` with `Detail: "skipped: authentication required (see check 7)"`. Skips never block subsequent checks — the user sees the full picture in one pass.
+
+## Execution Model
+
+### Default: local-only
+
+`azd ai agent doctor` runs checks 1–6 only. Same as MVP. No network, no token, no surprise prompts. This is the daily-driver mode.
```
therealjohn: What's the "MVP" referring to in this doc?

cli/azd/docs/design/azd-ai-agent-doctor-remote-checks.md
```diff
@@ -0,0 +1,266 @@
...
+
+### Opt-in: `--remote`
```
therealjohn: No need to separate out local vs. remote checks with a flag. It's fine to separate them out in sequence - check local things first, check remote things second - but users should not have to opt in to better checks. Instead, I think an opt-out of remote checks might be better for CI environments, etc. Something like `--local-only`.

cli/azd/docs/design/azd-ai-agent-doctor-remote-checks.md
```diff
@@ -0,0 +1,266 @@
...
+- `401` → token expired or scope mismatch. Fix: `azd auth login`; if the issue persists, see check 7.
+- `403` → wrong tenant **or** insufficient RBAC. Fix: confirm the active subscription/tenant matches the Foundry project; if it does, see check 10's role-assignment fix.
+- `404` → endpoint is wrong or project is gone. Fix: `azd provision` or fix env var.
+- network/DNS/TLS → "verify VPN / firewall / typo in `AZURE_AI_PROJECT_ENDPOINT`".
+
+The single-shot probe avoids paging and works on tiny test projects. Timeout: 10s, no retries — doctor is diagnostic, not resilient.
+
+### Check 9 — Model deployments
+
+**What it does:** parses each service's `agent.yaml` for model references, then queries the Foundry project's deployments list once and matches names locally.
```
therealjohn, line 148: This will need to change from agent.yaml to azure.yaml once #7962 lands. Sync with @trangevi on those changes.

cli/azd/docs/design/azd-ai-agent-doctor-remote-checks.md
```diff
@@ -0,0 +1,266 @@
...
+        --role "Azure AI Developer" \
+        --assignee <principal-id> \
+        --scope <project-resource-id>
+```
+
+**Redaction in non-interactive output.** When stdout is piped or `--output json` is set, principal IDs, scope ARNs, and full UPNs in both `detail` and `fix` strings are replaced with `<redacted>`. The JSON envelope sets `"redacted": true` so callers know they're getting the safe variant (see [`Exit codes & JSON output`](./azd-ai-agent-nextsteps.md#exit-codes--json-output) in the companion doc). An interactive-only `--unredacted` flag exists for users who need the raw values in a JSON pipeline.
+
+**Why we don't auto-create the assignment:** doctor is read-only. RBAC changes are the kind of action that should be explicit and audited.
+
+### Check 11 — Agent status
```
therealjohn, line 181: Can we add RBAC assignment checks to an active agent? A common issue right now is that an agent identity doesn't have permissions to access some resource. I don't think there is a way to figure out the specifics a user might hit, but we can at least check and display what roles the agent identity has and if it has access to specific project resources, or any resource in the project, or any resource in the account, etc.

cli/azd/docs/design/azd-ai-agent-doctor-remote-checks.md
```diff
@@ -0,0 +1,266 @@
...
+- **Redaction test.** Snapshot the rendered RBAC fix string in interactive (full values) and non-interactive (`<redacted>`) modes, plus the JSON envelope's `redacted` field.
+- **Snapshot test** for the rendered report in mixed pass/fail/skip configurations — same harness as MVP doctor's local-checks snapshot test.
+- **Functional test** (recorded cassette via `mage record`) covering one happy-path full run.
+- No live tests in CI for `--remote` — replay-only. Cassette refresh is a manual maintainer step.
+
+## Risks & Mitigations
+
+| Risk | Mitigation |
+|---|---|
+| Auth prompt during `--remote` surprises users in CI | Detect non-interactive (`!isTerminal`) and refuse the auth refresh; emit a clear "skipped: non-interactive" instead of hanging. |
```
therealjohn: Mitigated if we move to `--local-only` opt out model instead.

cli/azd/docs/design/azd-ai-agent-nextsteps.md
```diff
@@ -0,0 +1,499 @@
+# Design: Context-Aware Next-Step Guidance for `azd ai agent`
+
+## Status
+
+**Draft — for design review.**
+Tracking issue: [#7975](https://github.com/Azure/azure-dev/issues/7975) — "Context-aware next-step guidance and diagnostics".
+Scope: the `azure.ai.agents` extension only. **All code lives under `cli/azd/extensions/azure.ai.agents/`. No files outside the extension are modified.** The deploy hook returns its Next: block via the standard extension SDK return value (`*azdext.Artifact`) — specifically the `Metadata["note"]` key on an `ArtifactKindEndpoint` artifact, which `pkg/project/artifact.go` already renders as a continuation line under the artifact bullet (gated on `Kind == ArtifactKindEndpoint`). The extension already emits endpoint-kind artifacts at `internal/project/service_target_agent.go`; we populate the `note` metadata key on the same artifacts.
```
therealjohn, line 9: @rajeshkamal5050 Is this a good candidate and time for us to do `azure.ai.projects` extension and put this as a shared framework for next steps suggestions? Eventually this looks like something that could be provided by azd core, but for now, we can at least share it across azd ai * extensions.

cli/azd/docs/design/azd-ai-agent-doctor-remote-checks.md
antriksh30: Reworded the Background paragraph and Goal to make clear remote checks intentionally make network calls under user credentials  that's the whole point of the doc, not an apology. Combined with the flag flip below (#4), the doc no longer reads as 'doctor is local-only.' [575a45079](https://github.com/Azure/azure-dev/pull/8057/commits/575a45079)

cli/azd/docs/design/azd-ai-agent-doctor-remote-checks.md
antriksh30: Added a Terminology section near the top defining MVP explicitly: chronological term for 'local checks 16 ship first', not a permanent capability boundary. The phased rollout in Implementation Phases keeps that meaning consistent. [575a45079](https://github.com/Azure/azure-dev/pull/8057/commits/575a45079)

cli/azd/docs/design/azd-ai-agent-doctor-remote-checks.md
antriksh30: Rewrote the first Risks row. The old framing (users opt in to --remote and miss it in CI) doesn't apply now that doctor runs everything by default. New row covers the actual hazard: a CI run that forgot --local-only  check 7 detects non-interactive shells and fails fast with a 're-run with --local-only' message; checks 812 then Skip cleanly via the dependency matrix. No hang, no silent success. [575a45079](https://github.com/Azure/azure-dev/pull/8057/commits/575a45079)

cli/azd/docs/design/azd-ai-agent-doctor-remote-checks.md
antriksh30: Added a tracking note on Check 9 referencing #7962: once the unification lands, this read switches from gent.yaml to zure.yaml. Will sync with @trangevi on timing. [575a45079](https://github.com/Azure/azure-dev/pull/8057/commits/575a45079)

cli/azd/docs/design/azd-ai-agent-doctor-remote-checks.md
antriksh30: Same fix as the previous comment  Background and Execution Model are now framed around 'doctor runs all checks by default; --local-only opts out for CI.' Endpoint reachability (check 8) is no longer described as the exception. [575a45079](https://github.com/Azure/azure-dev/pull/8057/commits/575a45079)

cli/azd/docs/design/azd-ai-agent-doctor-remote-checks.md
antriksh30: Done  flipped the flag model. Default is now full sweep (112, including the new check 12 below); --local-only is the opt-out for CI / offline / fast-iteration. Rewrote the Execution Model section, the Output Format example, the Performance Budget, the Implementation Phases, and the Appendix to match. The old --remote is mentioned only once now, in the rejected-alternatives note. [575a45079](https://github.com/Azure/azure-dev/pull/8057/commits/575a45079)

cli/azd/docs/design/azd-ai-agent-doctor-remote-checks.md
antriksh30: Added Check 12  Agent identity role assignments. Lists roles on the agent's MI at three scopes (project / account / resource-group) with a new INFO glyph for the purely-informational case, plus Warn/Fail variants for the actionable ones. Kept it separate from check 10 because the principal, fix command, and preconditions all differ  merging would lose precision. Renamed check 10 to 'User RBAC' for symmetry. [575a45079](https://github.com/Azure/azure-dev/pull/8057/commits/575a45079)

jongio (approved):
All three items from my earlier review are addressed. Latest commit (flag model flip to --local-only, check 12 for agent identity roles, terminology section) introduces no new issues. Approving.

cli/azd/docs/design/azd-ai-agent-nextsteps.md
trangevi: Discussed in the large chat, I think for now let's leave it as part of the agent extension focused on the agent commands, and we can expand later.

cli/azd/docs/design/azd-ai-agent-nextsteps.md
```diff
@@ -0,0 +1,500 @@
...
+- Reworking error-path guidance — `internal/exterrors` already does this well.
+- Any modification to files outside `cli/azd/extensions/azure.ai.agents/`. State is read via the existing gRPC `azdext` API.
+- Sample-author hooks, README markers, or YAML-declared hints — out of scope for this design.
+- Authoring or editing core azd packages (`cli/azd/pkg/...`, `cli/azd/cmd/...`). The design only *consumes* the existing `Metadata["note"]` rendering contract.
+
+## Assumptions
+
+This design assumes the unified-`azure.yaml` proposal in [#7975](https://github.com/Azure/azure-dev/issues/7975) has landed: a single `azure.yaml` is the canonical config; `agent.yaml` and `agent.manifest.yaml` no longer exist as separate files. References to `agent.yaml` in the doctor checks and decision trees below are placeholders for the equivalent fields in the unified `services.<name>` stanza. If the extension ships before unification, each `agent.yaml` reference should be read as the corresponding fields under `services.<name>` in `azure.yaml`. The `AssembleState` function abstracts the source split, so resolvers are insulated from the migration — only the state-assembly code needs to change.
```
trangevi: We will not have a unified azure.yaml by //build, so for now this should proceed with the azure.yaml and agent.yaml (definition).

cli/azd/docs/design/azd-ai-agent-nextsteps.md
```diff
@@ -0,0 +1,500 @@
...
+The resolver works with whatever's available — partial state never silences guidance:
+
+| Layer | Source | Authority |
+|---|---|---|
+| 1 | Explicit CLI flags (`--project-endpoint`, `--agent`) | Highest |
+| 2 | Live runtime probes (OpenAPI, Foundry status) | Override env vars |
+| 3 | `azure.yaml` services, protocols, `config.env` refs | Structural truth |
+| 4 | azd env vars | Resolution truth |
+
+If the azd environment is missing or stale (a real scenario for brownfield users), the resolver should still produce useful output from layers 1–3.
```
trangevi: How will "stale" be determined?

cli/azd/docs/design/azd-ai-agent-nextsteps.md
```diff
@@ -0,0 +1,500 @@
...
+- Leading blank line separates from preceding command output.
+- README/docs pointers go on their own line below the commands.
+- Always present `Next:` prefix so the block is visually distinct.
+
+### Output discipline
+
+| Scenario | Destination |
+|---|---|
+| Interactive TTY | `stdout` |
+| Piped output / CI (detected via `isTerminal`) | `stderr` so it doesn't pollute parsable output |
```
trangevi: I don't really like the idea of this showing up as an error, I feel like that would potentially confuse anyone looking at the output to determine what happened in case of a failure/might confuse an agent thinking that these are failures. If anything, wouldn't we want this to be part of the parseable output, so it helps guide what is next?

cli/azd/docs/design/azd-ai-agent-nextsteps.md
```diff
@@ -0,0 +1,500 @@
...
+The `azdext.Artifact` SDK field used here renders only the **first line** at the natural indent level; subsequent lines need to be pre-indented by the producer to align under the artifact bullet. `FormatNextForNote` pre-indents lines 2+ with 4 spaces and trims leading/trailing newlines. This is purely a string-formatting concern handled inside the extension.
+
+## Hook Points
+
+| Command | Trigger | Mechanism |
+|---|---|---|
+| `init` | end of `runE` after success message | direct `nextstep.PrintNext(w, …)` |
+| `run` | "agent is listening" callback | direct print before `Press Ctrl+C` line |
+| `invoke --local` / `invoke` | end of `runE` on success | direct print |
+| `show` | end of table render | direct print |
```
trangevi: why not put everything at the end of the `runE` if possible? I understand some commands like `run` and `deploy` are unique, but most I think it would make sense to put in the same place for consistency?

trangevi (commented):
Looks good enough to proceed in my opinion. I'm sure other stuff might come up during implementation, but don't think we need to block on getting a perfect spec

cli/azd/docs/design/azd-ai-agent-nextsteps.md
antriksh30: Updated. The Configuration sources section now commits to the two-file split (azure.yaml + agent.yaml) and the AssembleState abstraction reads from both, so nothing in the spec depends on a unified schema landing.

cli/azd/docs/design/azd-ai-agent-nextsteps.md
antriksh30: Agreed and updated. The Output discipline table now suppresses the Next: block entirely on non-TTY (piped/CI) instead of routing it to stderr  it never appears in error streams. Structured callers consume the same data via the 
extStep JSON field on --output json.

cli/azd/docs/design/azd-ai-agent-nextsteps.md
antriksh30: Adopted. The Hook Points section now leads with end-of-runE as the rule, and explicitly calls out 
un (long-running listener  must print before blocking) and deploy (extension's only post-return channel is the artifact note) as the two intentional exceptions, with rationale.

cli/azd/docs/design/azd-ai-agent-nextsteps.md
antriksh30: Spelled out in the State Model section. 'Stale' resolves to one of three concrete cases: **Missing** (file/key absent), **Empty** (present but zero-valued), or **Drift** (env value present but stamped from a previous environment  detected by comparing the value's source env name against the currently selected env). The resolver treats each distinctly so the Next: hint is specific (e.g., 'run azd env refresh' for Drift vs. 'run azd provision' for Missing).

jongio (approved):
Incremental changes cleanly address @trangevi's review feedback. Configuration sources, output discipline, hook points, stale definition - all solid. Re-approving.

kristenwomack:
Hey @antriksh30, the Security & PII section, the exit codes, and the Backward Compatibility section all look good.

Items to address:

First, on metrics. We're already emitting `nextstep.resolved` once per resolver call. Could you add a session id (or any cheap correlation token) as an attribute so we can compute follow-through offline, the share of times the suggested branch matches the command the user runs next? That gives us one real outcome signal from day one without standing up a new event. The rest of the telemetry can stay in the follow-up design, and I'd like that follow-up tracked as an issue with a target date so it doesn't drift.

Second, the design says "there is no opt-out env var" for the Next: block (line 740). For most users that's fine, and a power user who finds it noisy has no escape hatch. Do we want an `AZD_AI_NEXT_STEPS=0` opt-out from day one, or are we comfortable waiting until someone asks?

Small nit, not blocking: @wbreza's P2 #7 references a multi-agent cap of 5, and the Risks table (line 774) says "≤ ~10." Could you pick one and put it in the design so the implementer doesn't pick a third?

jongio (commented):
No new commits since my approval on May 16. Revisiting to engage with @kristenwomack's feedback from May 28.

Three items still need author response:

1. **Telemetry correlation**: Adding a session/correlation attribute to `nextstep.resolved` makes sense. The azd telemetry context already carries a session identifier (`internal/tracing`). Threading it into this event is low effort and enables the most useful metric (follow-through rate) from day one. Small enough to include in the MVP telemetry, not defer. The full telemetry plan should still be tracked as an issue with a target date.

2. **Opt-out env var**: The `Next:` block is already suppressed on non-TTY and `--output json`, so CI/scripting is covered. An `AZD_AI_NEXT_STEPS=0` env var for interactive users can be added later if there's demand. Deferring seems fine, but please confirm the reasoning in the thread.

3. **Multi-agent cap**: Risks table says "≤ ~10" but wbreza's P2 #7 referenced 5. Pick one number so the implementer knows when to collapse to a summary line.

@antriksh30 please respond to @kristenwomack's comment to unblock the remaining reviewers.

therealjohn:
This was implemented - closing it.

PR #8067, opened 2026-05-06
Add support for gpuCount for partial SKU scenario

saanikaguptamicrosoft:
### Notes
- A new gpuCount field is added by APIs to support partial SKU scenario. Currently partial SKU is only supported for Singularity cluster.
- If customer passes gpuCount, then resources block is ignored by backend, and it intelligently picks the right SKU based on computeId to allocate the number of GPUs supported.
- Currently the support is for 1, 2, 4, 8 and multiples of 8 GPUs only. But backend team will be working on removing this constraint.

### Testing
YAML:
image

- GPU Count not supported
image

- Passing GPUCount as 2
image

- Passing both GPUCount and Resources (Resources gets ignored)
image

- Not a partial SKU scenario i.e. not passing GPUCount
image

### DIFF
--- a/cli/azd/extensions/azure.ai.customtraining/internal/cmd/job_get.go
+++ b/cli/azd/extensions/azure.ai.customtraining/internal/cmd/job_get.go
@@ -363,6 +363,13 @@ func printJobDetails(d *jobDetails) {
 		w.Flush()
 	}
 
+	if props.GPUCount > 0 {
+		fmt.Println()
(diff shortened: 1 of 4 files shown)


achauhan-scc (approved):

jongio (commented):
Traced the gpuCount field through all four files - YAML parser, REST model, submit, and display. The conditional logic in printComputeSection (preferring user-submitted gpuCount over run-history GPU count) and in printJobDetails (hiding the resources section when gpuCount is set) both look correct for the partial-SKU scenario described.

PR #8068, opened 2026-05-06
Share offline validation between job validate and job submit commands

saanikaguptamicrosoft:
### Notes
1. Refer validation PR: https://github.com/Azure/azure-dev/pull/7407 (some of the warning severity validations have been converted to error severity based on API response)
2. On top of that we also added validations which were missing in AML as an enhancement in user experience-
   - default is a reserved name for outputs (backend fails immediately)
   - ssh public keys should be present (the job fails later on and customer can't ssh without keys)
   - Input/Output type is required, else backend throws immediate error on submit
3. Refactored offline validation into a shared utils.ValidateJobOffline + utils.ReportValidationResult so submit runs the same checks as validate up-front, before any upload/network work.

### Testing

Happy Path
image

Unhappy paths
image

### DIFF
--- a/cli/azd/extensions/azure.ai.customtraining/internal/cmd/job_submit.go
+++ b/cli/azd/extensions/azure.ai.customtraining/internal/cmd/job_submit.go
@@ -217,7 +217,7 @@ func buildJobResource(def *utils.JobDefinition) *models.JobResource {
 		}
 	}
 
-	// Services (e.g., SSH). Validation in ValidateJobDefinition restricts type to "ssh"
+	// Services (e.g., SSH). ValidateJobOffline restricts type to "ssh"
 	// and ensures ssh_public_keys is non-empty.
 	if len(def.Services) > 0 {
 		job.Services = make(map[string]interface{}, len(def.Services))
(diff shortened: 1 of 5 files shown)


cli/azd/extensions/azure.ai.customtraining/internal/cmd/job_submit.go
```diff
@@ -35,12 +35,23 @@ func newJobSubmitCommand() *cobra.Command {
...
+			// Run shared offline validation up-front (same checks as `job validate`).
+			// On any error finding, abort before doing any network or upload work.
+			yamlDir := filepath.Dir(filePath)
+			validation := utils.ValidateJobOffline(jobDef, yamlDir)
+			if len(validation.Findings) > 0 {
+				if err := utils.ReportValidationResult(filePath, validation, false); err != nil {
+					return err
+				}
+				fmt.Println()
+			}
```
jongio, line 53: nit: Since `ValidateJobOffline` now runs here before any network work, the later comment at `buildJobResource` (around L164) still says `Validation in ValidateJobDefinition restricts type to 'ssh'` and `buildServiceRequest` says `enforced by ValidateJobDefinition`. Those references are stale after this PR - worth updating to `ValidateJobOffline`.

jongio (commented):
Clean refactoring - the shared validation path is straightforward and the new checks (reserved "default" output name, services validation) are well-tested.

One thing to clean up: job_submit.go still has two comments referencing the now-removed ValidateJobDefinition:

- Line ~164: // Services (e.g., SSH). Validation in ValidateJobDefinition restricts type to "ssh"
- Line ~173: // Currently only SSH is supported (enforced by ValidateJobDefinition).

These should say ValidateJobOffline instead.

Also, the PR description (Notes/Testing sections) is empty - might be worth filling in for reviewer context.

achauhan-scc (approved):

jongio (commented):
Addresses my earlier comment - stale references updated. The new input/output type-required validations look correct: path-style inputs (no `value:`) must declare a type, and all outputs must too. Test coverage is solid (positive + negative cases for both). Nothing else to flag.

Issue #8018, opened 2026-05-02
Use matched YAML rule IDs as the error ResultCode


JeffreyCA:
Phase 3 of #8011. Depends on #8017.

## Summary

The 48 string-pattern rules in `cli/azd/resources/error_suggestions.yaml` already encode the team's best classification of unstructured upstream errors. When a rule matches, the ResultCode collapses to `error.suggestion` — discarding the precise classification we just paid to compute.

Replace `ResultCode = "error.suggestion"` with `ResultCode = "suggest.<rule_id>"` (e.g. `suggest.arm.soft_delete_conflict`). Per design discussion, the data churn from this replacement is acceptable.

### 1. YAML schema

- Add `id: <stable-slug>` (regex `^[a-z0-9_]+(\.[a-z0-9_]+)*$`) to each rule in `cli/azd/resources/error_suggestions.yaml`.
- Update `cli/azd/resources/error_suggestions.schema.json` to declare `id` as required for new rules.
- Add a startup test that asserts uniqueness of all rule IDs and the regex format.

### 2. Plumbing

- Add `ID string \`yaml:"id,omitempty"\`` to `ErrorSuggestionRule` (`pkg/errorhandler/types.go`).
- Add `RuleID string` to `ErrorWithSuggestion` (`pkg/errorhandler/errors.go`).
- Set `RuleID` in `ErrorHandlerPipeline` when wrapping a matched error.
- In the shared classifier from #8017, when the outer is `*ErrorWithSuggestion` with non-empty `RuleID`:
  - Set `Classification.Code = "suggest." + RuleID`
  - Recursively classify the inner; surface its `ChainTypes` and `Attrs`
- Add `fields.ErrSuggestionRuleID` (`error.suggestion.rule_id`) and emit it as an attribute alongside the new ResultCode.

### 3. Prerequisite: middleware ordering fix

`TelemetryMiddleware.Run` (`cmd/middleware/telemetry.go:117`) calls `cmd.MapError(err, span)` **before** `ErrorMiddleware.Run` (`cmd/middleware/error.go:192`) wraps the error in `ErrorWithSuggestion`. Today, even if a rule matches, the rule ID never reaches the telemetry span.

Resolve by either:

- Swapping the middleware order so `ErrorMiddleware` runs first and wraps before telemetry classifies, **or**
- Keeping the order but propagating `RuleID` via context so telemetry can read it after the action returns.

Either approach is acceptable; the first is more direct.

## Tests

[...]

Issue #8011, opened 2026-05-02
[EPIC] Add chain, classifier, and origin context to azd error telemetry


JeffreyCA:
## Summary

When an azd command fails, telemetry today emits a single classification code as the AppInsights `RequestData.ResponseCode` — and not much else useful. The dominant catch-all bucket `internal.errors_errorString` (along with siblings like `internal.fmt_wrapError`) tells operators only that *something* failed, not what or where. There are no stacktraces, no chain visibility, no origin frame, and free-form error messages are deliberately not emitted (they may carry PII).

This epic covers the work to make every failed command in production telemetry diagnosable from Kusto with high probability.

## Goal

A triaging engineer who sees a Kusto bucket of azd errors should be able to answer *with high probability and without local repro*:

1. **Where in azd did this come from?** (package + function + line)
2. **What kind of error is it?** (named type chain, including any wrappers we control)
3. **Did one of our YAML suggestion rules already match it?** (if so, which)

Capturing the actual error *message text* (with redactions) is a nice-to-have (see #8016)

## Background — root cause of `internal.errors_errorString`

`MapError`'s fallback walks the chain and reports the *leaf* type (`internal/cmd/errors.go:428-458`). For `errors.New(...)` and `fmt.Errorf(...)` without `%w`, the leaf type is `*errors.errorString`, which sanitizes to the bucket name we see today. Named outer wrappers — which carry domain context — are silently discarded. Plus there are at least three duplicated classifier paths (`MapError`, `tracer.errorDescription`, `classifySuggestionType`) that drift apart over time.

## Phased plan

| Phase | Subissue | What |
|---|---|---|
| 1 | #8015 | Emit `error.chain.types` (string slice). Acts as the **measurement instrument** for everything that follows — gives Kusto data on which type chains dominate the catch-all today. |
| 2 | #8017 | Extract a shared `Classify(err) Classification` package. Unify the three classifier paths. Fix `errorType()` fallback to prefer **deepest-named-non-generic type**, with `internal.unclassified` only when the entire chain is generic stdlib types. |
| 3 | #8018 | Add stable `id` to each YAML rule in `error_suggestions.yaml`. Plumb through `ErrorWithSuggestion`. **Replace** `ResultCode = "error.suggestion"` with `suggest.<rule_id>`. Includes prerequisite middleware-ordering fix. |
[...]

PR #8087, opened 2026-05-06
Enrich error telemetry with chain types and a shared classifier

### LINKED ISSUE #8015: Add `error.chain.types` span attribute
Phase 1 of #8011.

## Summary

Add a purely additive attribute that records the wrapped-error **type chain** (e.g. `*fmt.wrapError → *azcore.ResponseError → *errors.errorString`). Type names only — no message text. Type names are code-defined (`SystemMetadata` classification), so this is low-cardinality and PII-free.

This is the **measurement instrument** for the rest of #8011: it gives us Kusto data on which type chains dominate the catch-all today, so the smarter-classifier work in #8017 and the bare-error-promotion sweep can be data-driven rather than speculative.

## Why first

Today, an engineer triaging `internal.errors_errorString` in Kusto sees a count and timestamps. With `error.chain.types`, they can group by type chain and immediately see which underlying types dominate. This alone makes the residual catch-all triageable without changing any classifier logic.

## Implementation

1. Add `fields.ErrChainTypes` (`error.chain.types`, type `string[]`, `SystemMetadata`, `PerformanceAndHealth`) in `cli/azd/internal/tracing/fields/fields.go`.
2. Implement a shared `BuildChainTypes(err error) []string` walker in a low-level package usable by both `internal/cmd` (for `MapError`) and `internal/tracing` (for `EndWithStatus`). Handle both `Unwrap() error` and `Unwrap() []error`. Cap output at ~10 entries defensively.
3. Call the helper from both `MapError` and `wrapperSpan.EndWithStatus`.
4. Emit type names only — no message text.

Does not change `ResultCode`, so no Kusto consumer impact.

## Cross-cutting cleanups (include in this PR)

- Existing `error.frame` field is misleading: it's an ARM deployment-error nesting depth integer (`internal/cmd/errors.go:460-488`), not a Go `runtime.Frame`. Rename it (e.g. to `error.arm.frame_index`) to free the name for future Phase 4 origin-frame work.
- Existing `error.inner` key is defined in `fields.go:507-519` but never emitted in production code. Either remove or document as reserved.

## Tests

- Unit tests for the chain walker covering wrapped, joined, and degenerate chains using `mocktracing.Span`.
- AppInsights exporter test asserting `string[]` attributes serialize correctly into `Properties` as JSON.
- Functional `--trace-log-file` smoke test if practical.

## Notes

Earlier drafts proposed `error.chain.messages` (per-frame message text) alongside this. That has been removed and folded into #8016 (deferred behind the redactor pipeline).

### LINKED ISSUE #8017: Extract a shared error classifier and fix the leaf-only fallback
Phase 2 of #8011. Depends on #8015.

## Summary

Two tightly coupled changes to the heart of the error classifier:

1. **Smarter unwrap** in the fallback path — prefer the **deepest named non-generic** wrapper type over the leaf `*errors.errorString` / `*fmt.wrapError`.
2. **Extract a shared classifier** so all current callers (`cmd.MapError`, `tracing.EndWithStatus`, and `classifySuggestionType`) stop duplicating the typed/sentinel ladder.

The current `errorType()` (`internal/cmd/errors.go:428-458`) walks to the leaf and reports its type. Names like `*ServiceManagerError` on the *outside* are silently discarded. Switching to "deepest named non-generic type" redistributes the catch-all into more meaningful bins immediately, without changing any callsite.

### 1. Smarter unwrap fallback

When walking the wrapped chain in the fallback path:

- Pick the **deepest** type whose `pkg.Name` is **not** in the generic stdlib set: `*errors.errorString`, `*fmt.wrapError`, `*errors.joinError`, `*errors.wrapErrors`, `*errorhandler.ErrorWithSuggestion`.
- If no named non-generic type exists in the chain, ResultCode = `internal.unclassified` (a single stable bucket name).
- The `internal.errors_errorString` / `internal.fmt_wrapError` ResultCodes will largely disappear from telemetry. Per design discussion, this churn is acceptable.

### 2. Shared classifier package

Pull the typed/sentinel ladder out of `internal/cmd/errors.go` into a new package (suggested: `cli/azd/internal/tracing/errclassify/`).

Public surface:

```go
type Classification struct {
    Code       string                    // becomes Status.Description
    Category   string
    ChainTypes []string
    Origin     *runtime.Frame            // populated in Phase 4 (#8077)
    Attrs      []attribute.KeyValue
    RuleID     string                    // populated in Phase 3 (#8018)
}

func Classify(err error) Classification
func (c Classification) Apply(span tracing.Span)
```

After the refactor:

- `cmd.MapError(err, span)` becomes `errclassify.Classify(err).Apply(span)` (plus any cmd-specific decoration).
- `tracing.wrapperSpan.EndWithStatus` calls the same `Classify`/`Apply` — eliminates the leaf-only `errorDescription` path.
- `classifySuggestionType` (`internal/cmd/errors.go:326-425`) — the **third** duplicate decision tree — collapses into the same shared classifier.

## Tests

- Golden-file table test: every fixture error produces identical attributes through both old entry points (`MapError` and `EndWithStatus`). CI fails any drift.
- Existing classification tests in `internal/cmd/errors_test.go` migrate to the new package and continue to pass.
- New tests for the deepest-named-non-generic walker.

## Dependencies

Depends on #8015 having shipped (so `error.chain.types` is already on every error span and visible during the migration).

JeffreyCA:
Resolves #8015 
Resolves #8017

This PR makes azd error telemetry easier to triage without needing to reproduce a failure. It is the first wave of work under epic #8011 and covers sub-issues #8015 and #8017.

Every error span now emits an `error.chain.types` attribute that records the wrapped-error type chain in outermost-first order. This gives engineers a PII-free way to group residual catch-all errors in Kusto by code-defined error types instead of scraping message text.

The error classifier now uses a shared classification path instead of separate decision trees for telemetry, suggestions, and tracing. Its fallback behavior also reports the deepest meaningful non-generic error type instead of collapsing many wrapped errors to `*errors.errorString`, making typed underlying errors visible in telemetry. The legacy catch-all bucket is renamed from `internal.errors_errorString` to `internal.unclassified`.

This PR also removes unused error telemetry fields that were never emitted and renames the ARM `service.errorCode` payload's frame field to `error.arm.frame_index` to better reflect that it represents ARM nesting depth rather than a Go runtime frame.

For example, a wrapped ARM deployment failure now includes the wrapped error chain on both sub-spans and the command span:

```json
{
  "Name": "cmd.provision",
  "Status": {
    "Code": "Error",
    "Description": "error.suggestion"
  },
  "Attributes": [
    {
      "Key": "error.chain.types",
      "Value": {
        "Type": "STRINGSLICE",
        "Value": [
          "*errorhandler.ErrorWithSuggestion",
          "*fmt.wrapError",
          "*azapi.AzureDeploymentError",
          "*azapi.DeploymentErrorLine"
        ]
      }
    },
    {
      "Key": "error.type",
      "Value": {
        "Type": "STRING",
        "Value": "service.arm.deployment.failed"
      }
    }
  ]
}
```

For plain or generic-only errors, the status now uses the clearer catch-all bucket while keeping the concrete chain available for investigation:

```json
{
  "Status": {
    "Code": "Error",
    "Description": "internal.unclassified"
  },
  "Attributes": [
    {
      "Key": "error.chain.types",
      "Value": {
        "Type": "STRINGSLICE",
        "Value": [
          "*fmt.wrapError",
          "*errors.errorString"
        ]
      }
    }
  ]
}
```

A follow-up PR will be made to address https://github.com/Azure/azure-dev/issues/8018, which adds rule IDs to YAML-based error suggestions and emits result codes as `suggest.<rule_id>` instead of `error.suggestion`.

### DIFF
--- a/cli/azd/.vscode/cspell.yaml
+++ b/cli/azd/.vscode/cspell.yaml
@@ -27,6 +27,7 @@ words:
   - cooldown
   - customtype
   - devcontainers
+  - errchain
   - errgroup
   - errorhandler
   - extendee
@@ -63,6 +64,7 @@ words:
   - syncmap
   - syscall
   - tsx
+  - triaging
   - Retryable
   - runcontext
   - surveyterm
@@ -115,6 +117,7 @@ words:
   - exfiltration
   - Fprintf
   - gocritic
+  - gocyclo
   - gradlew
   - IMDS
   - managedhsm
--- a/cli/azd/internal/cmd/errors.go
+++ b/cli/azd/internal/cmd/errors.go
@@ -13,7 +13,6 @@ import (
 	"log"
 	"net"
 	"path/filepath"
-	"reflect"
 	"regexp"
 	"strings"
 
@@ -23,6 +22,7 @@ import (
 	"github.com/azure/azure-dev/cli/azd/internal"
 	"github.com/azure/azure-dev/cli/azd/internal/agent/consent"
 	"github.com/azure/azure-dev/cli/azd/internal/tracing"
+	"github.com/azure/azure-dev/cli/azd/internal/tracing/errchain"
 	"github.com/azure/azure-dev/cli/azd/internal/tracing/fields"
 	"github.com/azure/azure-dev/cli/azd/pkg/auth"
 	"github.com/azure/azure-dev/cli/azd/pkg/azapi"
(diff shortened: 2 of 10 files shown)


jongio (commented):
`sanitizeTypeName` is now duplicated between `tracer.go` and `errors.go` (same 1-liner). Not blocking, but `errchain` would be a natural home if you want one copy since it already handles type-name logic.

Note on existing bot comments: the sub-span/root-span divergence concern is resolved in this revision (both paths produce `internal.unclassified` for generic-only chains), the `require.Subset` weakness is gone (uses `ElementsMatch` + filter), and the hardcoded `*exported.ResponseError` concern is addressed via `%T`.

wbreza (commented):
﻿## Code Review — PR #8087

Well-designed PR with clean separation (low-level rrchain package + high-level classify() consolidation). No correctness bugs found. Modern Go 1.26 patterns used correctly. Comprehensive test coverage in rrchain_test.go.

### P1 — Should Fix

**P1-1: Telemetry schema docs not updated**
Removal of rror.frame/rror.inner, rename to rror.arm.frame_index, addition of rror.chain.types, and catch-all bucket rename (internal.errors_errorString → internal.unclassified) have no corresponding schema doc updates. Telemetry consumers have no migration reference.

**P1-2: Catch-all bucket rename is a silent breaking change**
internal.errors_errorString → internal.unclassified will silently break any Kusto queries, dashboards, or alerts filtering on the old value. Consider adding migration guidance or a tracking issue for downstream consumers.

### P2 — Consider

| # | Finding | Location |
|---|---|---|
| P2-1 | sanitizeTypeName() duplicated in both rrors.go (~L280) and 	racer.go (~L162) — same logic, two places to maintain. Consider moving to a shared location. | rrors.go, 	racer.go |
| P2-2 | classify() has 25+ branches with implicit ordering dependencies. A block comment explaining the ordering contract (most specific first, ErrorWithSuggestion recurses, fallback at bottom) would prevent future shadowing bugs. | rrors.go |
| P2-3 | genericWrappers map lacks a doc comment explaining criteria for adding entries or the telemetry impact of inclusion/exclusion. | rrchain.go |
| P2-4 | The only recursive path in classify() (nested ErrorWithSuggestion) has no test coverage. Consider adding a test with suggestion-wrapping-suggestion. | rrors_test.go |

### P3 — Nit

- Import group ordering in 	racer.go — external packages appear after internal. goimports should fix.

wbreza (approved):
﻿## Re-Review — Address Feedback Commit

The new commit cleanly resolves 6 of 7 findings from the previous review:

| Finding | Status |
|---|---|
| P1-1: Telemetry schema docs not updated | ✅ Fixed — `telemetry-schema.md` updated with `error.chain.types`, removed fields, and `internal.unclassified` rename |
| P1-2: Catch-all bucket rename breaking change | ⚠️ Mitigated — schema doc now explains the change; recommend a tracking issue for dashboard migration |
| P2-1: `sanitizeTypeName()` duplicated | ✅ Fixed — moved to shared `errchain.SanitizeTypeName()` |
| P2-2: `classify()` ordering undocumented | ✅ Fixed — block comment added |
| P2-3: `genericWrappers` lacks maintenance guidance | ✅ Fixed — comment explains criteria and telemetry impact |
| P2-4: Nested `ErrorWithSuggestion` untested | ✅ Fixed — test added |
| P3-1: Import ordering in `tracer.go` | ✅ Fixed |

All changes in the delta are clean structural refactors — no logic changes, no new issues introduced. LGTM.

hemarina (approved):
**Overall:** Approve with two minor nits. Nice cleanup — the shared `classify()` is much easier to audit than the previous parallel decision trees, and pulling `errchain` into its own dependency-free package is the right call for the sub-span path.

### Nits (non-blocking)

**1. `classifyErrorWithSuggestion` with a nil inner changes `error.type` from `"<nil>"` → `"internal.<nil>"`**

When `ews.Unwrap()` returns nil, the new path goes through `classify(nil)` which returns `"internal.<nil>"`, and that string gets stamped into `error.type`. The old `classifySuggestionType(nil)` returned plain `"<nil>"`. It's an edge case (suggestion wrapping a nil error shouldn't really happen), but worth either preserving the prior value or adding a line to the PR body noting the bucket renamed, in case anyone has a Kusto filter on `error.type == "<nil>"`.

**2. `*errors.joinError` in `genericWrappers` relies on a stdlib unexported type name**

Same flavor as the existing open thread about `*exported.ResponseError` in the test. If the stdlib ever renames `joinError` (rare, but the `fmt.wrapError` family has been reshuffled before), classification silently regresses without any test failing. A small guard test that asserts `reflect.TypeOf(errors.Join(...)).String() == "*errors.joinError"` would make that breakage loud at upgrade time.

### Things I checked and am happy with
- `walk()` recursion is safely bounded by `MaxChainLen=16` (covers the `selfRefErr` cycle test).
- `classify()`'s independent-`if`-`return` blocks preserve the old `else if` ordering — every typed branch returns immediately, so no fallthrough hazard.
- `fields.ErrorKey` is idempotent (`HasPrefix("error.")` guard), so re-prefixing the suggestion path's `error.type` attr is a no-op — no double-prefix bug.
- Joined-error semantics in `DeepestNamedType` (first non-nil child) are intentional and now codified by `TestDeepestNamedType_JoinedFirstChild`. Typed branches still find joined children via `errors.As`, so only the fallback path is narrowed — acceptable.
- `error.chain.types` cardinality is bounded and PII-free as stated.

### Optional polish
- `Test_MapError_ChainTypes` could add a `nil` case to lock in the "no attribute emitted" contract.

JeffreyCA:
> **Overall:** Approve with two minor nits. Nice cleanup — the shared `classify()` is much easier to audit than the previous parallel decision trees, and pulling `errchain` into its own dependency-free package is the right call for the sub-span path.
> 
> ### Nits (non-blocking)
> **1. `classifyErrorWithSuggestion` with a nil inner changes `error.type` from `"<nil>"` → `"internal.<nil>"`**
> 
> When `ews.Unwrap()` returns nil, the new path goes through `classify(nil)` which returns `"internal.<nil>"`, and that string gets stamped into `error.type`. The old `classifySuggestionType(nil)` returned plain `"<nil>"`. It's an edge case (suggestion wrapping a nil error shouldn't really happen), but worth either preserving the prior value or adding a line to the PR body noting the bucket renamed, in case anyone has a Kusto filter on `error.type == "<nil>"`.
> 
> **2. `*errors.joinError` in `genericWrappers` relies on a stdlib unexported type name**
> 
> Same flavor as the existing open thread about `*exported.ResponseError` in the test. If the stdlib ever renames `joinError` (rare, but the `fmt.wrapError` family has been reshuffled before), classification silently regresses without any test failing. A small guard test that asserts `reflect.TypeOf(errors.Join(...)).String() == "*errors.joinError"` would make that breakage loud at upgrade time.
> 
> ### Things I checked and am happy with
> * `walk()` recursion is safely bounded by `MaxChainLen=16` (covers the `selfRefErr` cycle test).
> * `classify()`'s independent-`if`-`return` blocks preserve the old `else if` ordering — every typed branch returns immediately, so no fallthrough hazard.
> * `fields.ErrorKey` is idempotent (`HasPrefix("error.")` guard), so re-prefixing the suggestion path's `error.type` attr is a no-op — no double-prefix bug.
> * Joined-error semantics in `DeepestNamedType` (first non-nil child) are intentional and now codified by `TestDeepestNamedType_JoinedFirstChild`. Typed branches still find joined children via `errors.As`, so only the fallback path is narrowed — acceptable.
> * `error.chain.types` cardinality is bounded and PII-free as stated.
> 
> ### Optional polish
> * `Test_MapError_ChainTypes` could add a `nil` case to lock in the "no attribute emitted" contract.



> **Overall:** Approve with two minor nits. Nice cleanup — the shared `classify()` is much easier to audit than the previous parallel decision trees, and pulling `errchain` into its own dependency-free package is the right call for the sub-span path.
> 
> ### Nits (non-blocking)
> **1. `classifyErrorWithSuggestion` with a nil inner changes `error.type` from `"<nil>"` → `"internal.<nil>"`**
> 
> When `ews.Unwrap()` returns nil, the new path goes through `classify(nil)` which returns `"internal.<nil>"`, and that string gets stamped into `error.type`. The old `classifySuggestionType(nil)` returned plain `"<nil>"`. It's an edge case (suggestion wrapping a nil error shouldn't really happen), but worth either preserving the prior value or adding a line to the PR body noting the bucket renamed, in case anyone has a Kusto filter on `error.type == "<nil>"`.
> 
> **2. `*errors.joinError` in `genericWrappers` relies on a stdlib unexported type name**
> 
> Same flavor as the existing open thread about `*exported.ResponseError` in the test. If the stdlib ever renames `joinError` (rare, but the `fmt.wrapError` family has been reshuffled before), classification silently regresses without any test failing. A small guard test that asserts `reflect.TypeOf(errors.Join(...)).String() == "*errors.joinError"` would make that breakage loud at upgrade time.
> 
> ### Things I checked and am happy with
> * `walk()` recursion is safely bounded by `MaxChainLen=16` (covers the `selfRefErr` cycle test).
> * `classify()`'s independent-`if`-`return` blocks preserve the old `else if` ordering — every typed branch returns immediately, so no fallthrough hazard.
> * `fields.ErrorKey` is idempotent (`HasPrefix("error.")` guard), so re-prefixing the suggestion path's `error.type` attr is a no-op — no double-prefix bug.
> * Joined-error semantics in `DeepestNamedType` (first non-nil child) are intentional and now codified by `TestDeepestNamedType_JoinedFirstChild`. Typed branches still find joined children via `errors.As`, so only the fallback path is narrowed — acceptable.
> * `error.chain.types` cardinality is bounded and PII-free as stated.
> 
> ### Optional polish
> * `Test_MapError_ChainTypes` could add a `nil` case to lock in the "no attribute emitted" contract.

Thanks, will consider this in the follow-up PR for 8018

PR #8104, opened 2026-05-08
feat(agents): deploy using an existing acr image

### LINKED ISSUE #7988: deploy from an existing ACR image
Make it simple to deploy a hosted agent from an existing ACR Image path instead of doing the remote build.

i.e. "create hosted agent from name.azure.io/myregistry/myimage:v1" vs "build it for me".

huimiu:
https://github.com/Azure/azure-dev/issues/7988
This pull request adds support for using pre-built container images when deploying hosted agents, allowing users to skip Dockerfile builds and pushes if an image is specified in agent.yaml. The implementation prompts the user to choose between using the pre-built image or building a new one, and this decision is cached per deployment. 

* Added an `Image` field to the `ContainerAgent` struct, allowing users to specify a pre-built container image in agent.yaml. If set, this image is used for deployment instead of building from a Dockerfile.
* Updated the deployment logic to prompt the user when a pre-built image is present, caching the decision and skipping build/publish steps if the pre-built image is chosen.
* Implemented helper methods to manage user choice and image resolution from agent.yaml.

### Test
- Build it for me by dafault:
  image
- Deploying:
   image
- Deploying (no prompt):
   image

### DIFF
--- a/cli/azd/extensions/azure.ai.agents/internal/project/service_target_agent.go
+++ b/cli/azd/extensions/azure.ai.agents/internal/project/service_target_agent.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"regexp"
 	"strconv"
 	"strings"
 
@@ -81,11 +82,18 @@ type AgentServiceTargetProvider struct {
 	tenantId            string
 	env                 *azdext.Environment
 	foundryProject      *arm.ResourceID
+	usePreBuiltImage    *bool // cached prompt decision; nil means not yet asked
 }
 
 const (
 	preBuiltImageArtifactSourceKey = "azure.ai.agents.imageSource"
 	preBuiltImageArtifactSource    = "agent.yaml"
+)
+
+// containerImageRefRe is a basic pattern for container image references:
+// [registry/]repository[:tag|@digest]
+var containerImageRefRe = regexp.MustCompile(
+	`^[a-zA-Z0-9]([a-zA-Z0-9._-]*/)*[a-zA-Z0-9][a-zA-Z0-9._-]*(:[a-zA-Z0-9._-]+|@sha256:[0-9a-fA-F]{64})?$`,
 )
 
 // NewAgentServiceTargetProvider creates a new AgentServiceTargetProvider instance
@@ -581,6 +589,14 @@ func (p *AgentServiceTargetProvider) loadContainerAgentDefinition() (agent_yaml.
 		)
 	}
 
+	if agentDef.Image != "" && !containerImageRefRe.MatchString(agentDef.Image) {
+		return agent_yaml.ContainerAgent{}, false, exterrors.Validation(
+			exterrors.CodeInvalidAgentManifest,
+			fmt.Sprintf("invalid container image reference in agent.yaml: %q", agentDef.Image),
+			"use a valid image reference, e.g. 'myregistry.azurecr.io/myimage:v1'",
+		)
+	}
+
 	return agentDef, true, nil
 }
 
@@ -663,8 +679,14 @@ func (p *AgentServiceTargetProvider) shouldUsePreBuiltImage(
 		return false, nil
 	}
 
+	// Return cached decision if already prompted.
+	if p.usePreBuiltImage != nil {
+		return *p.usePreBuiltImage, nil
+	}
+
 	// Non-interactive (CI/CD): honor the configured image without prompting.
 	if noPrompt, _ := strconv.ParseBool(os.Getenv("AZD_NO_PROMPT")); noPrompt {
+		p.usePreBuiltImage = new(true)
 		return true, nil
 	}
 
@@ -685,7 +707,9 @@ func (p *AgentServiceTargetProvider) shouldUsePreBuiltImage(
 		return false, exterrors.FromPrompt(err, "failed to select hosted agent container image source")
 	}
 
-	return resp.Value != nil && *resp.Value == 1, nil
+	selected := resp.Value != nil && choices[*resp.Value].Value == "prebuilt"
+	p.usePreBuiltImage = &selected
+	return selected, nil
 }
 
 // deployHostedAgent deploys a container-based hosted agent to the Foundry service.
--- a/cli/azd/extensions/azure.ai.agents/internal/project/service_target_agent.go
+++ b/cli/azd/extensions/azure.ai.agents/internal/project/service_target_agent.go
@@ -664,9 +664,7 @@ func (p *AgentServiceTargetProvider) Deploy(
 //
 // Behavior:
 //   - If no image is configured in agent.yaml, always build from Dockerfile.
-//   - In non-interactive mode (AZD_NO_PROMPT=true), honor the declarative image
-//     field directly: the user explicitly opted in by writing it in agent.yaml,
-//     so CI/CD can deploy without an extra flag.
+//   - In non-interactive mode (AZD_NO_PROMPT=true), always build from Dockerfile.
 //   - In interactive mode, prompt the user. The default is to build, so users
 //     who happen to have an image in agent.yaml are not silently switched onto
 //     the pre-built path.
@@ -684,10 +682,10 @@ func (p *AgentServiceTargetProvider) shouldUsePreBuiltImage(
 		return *p.usePreBuiltImage, nil
 	}
 
-	// Non-interactive (CI/CD): honor the configured image without prompting.
+	// Non-interactive (CI/CD): always build from Dockerfile.
 	if noPrompt, _ := strconv.ParseBool(os.Getenv("AZD_NO_PROMPT")); noPrompt {
-		p.usePreBuiltImage = new(true)
-		return true, nil
+		p.usePreBuiltImage = new(false)
+		return false, nil
 	}
 
 	// Interactive: default to build so the pre-built path requires an explicit choice.
--- a/cli/azd/extensions/azure.ai.agents/internal/project/service_target_agent.go
+++ b/cli/azd/extensions/azure.ai.agents/internal/project/service_target_agent.go
@@ -82,7 +82,6 @@ type AgentServiceTargetProvider struct {
 	tenantId            string
 	env                 *azdext.Environment
 	foundryProject      *arm.ResourceID
-	usePreBuiltImage    *bool // cached prompt decision; nil means not yet asked
 }
 
 const (
@@ -677,14 +676,8 @@ func (p *AgentServiceTargetProvider) shouldUsePreBuiltImage(
 		return false, nil
 	}
 
-	// Return cached decision if already prompted.
-	if p.usePreBuiltImage != nil {
-		return *p.usePreBuiltImage, nil
-	}
-
 	// Non-interactive (CI/CD): always build from Dockerfile.
 	if noPrompt, _ := strconv.ParseBool(os.Getenv("AZD_NO_PROMPT")); noPrompt {
-		p.usePreBuiltImage = new(false)
 		return false, nil
 	}
 
@@ -705,9 +698,7 @@ func (p *AgentServiceTargetProvider) shouldUsePreBuiltImage(
 		return false, exterrors.FromPrompt(err, "failed to select hosted agent container image source")
 	}
 
-	selected := resp.Value != nil && choices[*resp.Value].Value == "prebuilt"
-	p.usePreBuiltImage = &selected
-	return selected, nil
+	return resp.Value != nil && choices[*resp.Value].Value == "prebuilt", nil
 }
 
 // deployHostedAgent deploys a container-based hosted agent to the Foundry service.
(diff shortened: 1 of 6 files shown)


wbreza (commented):
## Code Review — PR #8104

**Overall:** ✅ Well-structured PR — no blockers. Caching logic is correct, image priority works properly, and test coverage is solid for happy paths.

### Findings

#### 🟡 Medium

**M1 — `getPreBuiltImageURL` silently swallows parse errors** (`service_target_agent.go` ~lines 638–649)
Returns `""` on both file-read and YAML-unmarshal errors. If `agent.yaml` exists but has a syntax error around the `image:` field, the pre-built image prompt is silently skipped with no user indication. Other paths in the same file (Deploy flow) properly propagate parse errors.
**Suggestion:** Distinguish "file not found" (acceptable silent empty) from "file exists but parse failed" (should warn or propagate).

**M2 — Missing test for prompt cancellation/error path** (`service_target_agent_test.go`)
`shouldUsePreBuiltImage` wraps prompt errors with `exterrors.FromPrompt(...)`, but this path is untested. Should verify error propagates and cache stays `nil` for retries.

**M3 — Missing complementary test cases** (`service_target_agent_test.go`)
No test for `Publish` proceeding normally when user chose Dockerfile (`usePreBuiltImage = false`), and no test for `getPreBuiltImageURL` with malformed YAML.

#### 🔵 Low

**L1 — Redundant YAML re-read** (`service_target_agent.go`)
`getPreBuiltImageURL` independently reads and parses `agent.yaml`; the Deploy flow re-reads it. Could accept the parsed struct or cache the parse result to avoid TOCTOU inconsistency.

**L2 — Implicit non-interactive behavior** (`service_target_agent.go`)
When running with `--no-prompt`, `Prompt().Select()` may default to pre-built (index 0) or error. The intended CI/CD behavior should be documented.

**L3 — No empty-endpoint guard** (`service_target_agent.go`)
`registerAgentEnvironmentVariables` doesn't guard against empty `AZURE_AI_PROJECT_ENDPOINT`, though the deploy path validates it upstream.

### What Looks Good 👍
- Caching pattern (`*bool` tri-state) is correct and idiomatic
- Image priority (YAML image > published artifact) works properly
- Prompt UX displays actual image URL with sensible default
- YAML/JSON tags follow existing conventions
- Solid test coverage for parse, caching, and skip behaviors

cli/azd/extensions/azure.ai.agents/internal/project/service_target_agent.go
```diff
@@ -568,6 +590,80 @@ func (p *AgentServiceTargetProvider) isContainerAgent() bool {
...
+		return ""
+	}
+
+	return agentDef.Image
+}
+
+// resolvePreBuiltImage returns the pre-built container image URL if one is configured
+// in agent.yaml and the user chose to use it.
+// Returns empty string if no pre-built image is specified.
+func (p *AgentServiceTargetProvider) resolvePreBuiltImage(
```
trangevi: This function seems largely redundant with `getPreBuildImageURL`. One already has the agent def, the other retrieves it, but both do the same thing ultimately. Would suggest consolidating to one for consistent handling

cli/azd/extensions/azure.ai.agents/internal/project/service_target_agent.go
```diff
@@ -588,21 +684,36 @@ func (p *AgentServiceTargetProvider) deployHostedAgent(
...
+		return nil, err
 	}
-	if fullImageURL == "" {
-		return nil, exterrors.Dependency(
-			exterrors.CodeMissingPublishedContainer,
-			"published container artifact not found: no remote container artifact was found in service publish artifacts",
-			"run 'azd package' and 'azd publish' (or 'azd deploy') to produce container artifacts",
-		)
+	if !usePreBuiltImage && agentDef.Image != "" {
+		fmt.Fprintf(os.Stderr, "Building container image instead of using pre-built image: %s\n", agentDef.Image)
```
trangevi: "Building" has already happened by this point, the old step 1 comment was incorrect. I think this check and print are unnecessary. After the usePreBuiltImage check, if it's true then we should resolve the pre built image (with whichever method is kept after my comment above) and move on. If usePreBuiltImage is false then just move on to using the publish artifact

cli/azd/extensions/azure.ai.agents/internal/project/service_target_agent.go
```diff
@@ -568,6 +590,80 @@ func (p *AgentServiceTargetProvider) isContainerAgent() bool {
...
 }
 
+// shouldUsePreBuiltImage determines whether to use a pre-built image.
+// When an image URL is configured, it prompts the user to choose between
+// using the pre-built image or building from a Dockerfile.
+// The decision is cached so the prompt only appears once per deploy.
+func (p *AgentServiceTargetProvider) shouldUsePreBuiltImage(
+	ctx context.Context,
+) (bool, error) {
+	// Return cached decision if already prompted
```
trangevi: @therealjohn, do we actually want this cached? Without some way to detect underlying code changes, this could possibly lead to unexpected behavior, or multiple deployments when a user realizes they won't be prompted again.

trangevi (changes requested):

cli/azd/extensions/azure.ai.agents/internal/project/service_target_agent.go
therealjohn: Safer for now to not cache it - but provide a way to --no-prompt so an agent can deploy or CI/CD can work.

therealjohn (changes requested):

cli/azd/extensions/azure.ai.agents/internal/project/service_target_agent_test.go
```diff
@@ -574,3 +613,199 @@ func TestAgentPlaygroundURL_AccountLevelID(t *testing.T) {
...
+	result, err := provider.shouldUsePreBuiltImage(t.Context())
+	require.NoError(t, err)
+	require.False(t, result, "should default to build when no image is configured")
+}
+
+func TestShouldUsePreBuiltImage_PromptsForConfiguredImage(t *testing.T) {
+	t.Parallel()
+
+	dir := t.TempDir()
+	imageURL := "myregistry.azurecr.io/myimage:v1"
```
jongio: This verifies the "Build it for me" path (index 1) and confirms false is returned. There's no complementary test where `selectedIndex: 0` verifies the function returns `true` when the user picks the pre-built option. That path is only tested via preset caching (`new(true)`) in `TestShouldUsePreBuiltImage_CachesDecision`, which doesn't exercise the actual prompt response parsing.

Something like:
```go
func TestShouldUsePreBuiltImage_PromptsForPreBuiltImage(t *testing.T) {
	t.Parallel()
	dir := t.TempDir()
	agentPath := writeHostedAgentYAMLWithImage(t, dir, "myregistry.azurecr.io/myimage:v1")
	promptStub := &stubPromptServer{selectedIndex: 0}
	client := newPromptTestClient(t, promptStub)
	provider := &AgentServiceTargetProvider{
		azdClient:           client,
		agentDefinitionPath: agentPath,
	}
	result, err := provider.shouldUsePreBuiltImage(t.Context())
	require.NoError(t, err)
	require.True(t, result, "expected pre-built image to be selected")
}
```

jongio (commented):
One test gap I didn't see covered in existing feedback:

`TestShouldUsePreBuiltImage_PromptsForConfiguredImage` exercises the "build" path (index 1) but not the pre-built path (index 0). The feature's happy path through the prompt isn't tested - only via preset caching in `TestShouldUsePreBuiltImage_CachesDecision`.

cli/azd/extensions/azure.ai.agents/internal/project/service_target_agent.go
jongio: Worth noting: the prompt default is index 0 (pre-built image). In --no-prompt mode, the host returns the default, so CI/CD would silently use the pre-built image instead of building. If "build" is the safer CI default, swap the choice order or set defaultIndex to 1.

huimiu: Thanks for catching this. I've updated the prompt so that:

In both interactive and non-interactive (AZD_NO_PROMPT=true) mode, build from Dockerfile is always the default, interactive mode prompts the user to explicitly opt into the pre-built image, while non-interactive mode skips the prompt entirely and always builds.

cli/azd/extensions/azure.ai.agents/internal/project/service_target_agent.go
```diff
@@ -420,10 +448,34 @@ func (p *AgentServiceTargetProvider) Publish(
...
+
+	agentDef, isContainerAgent, err := p.loadContainerAgentDefinition()
+	if err != nil {
+		return nil, err
+	}
+	if !isContainerAgent {
 		return &azdext.ServicePublishResult{}, nil
 	}
 
+	if !hasContainerArtifact(serviceContext.Package) {
```
trangevi: What is the scenario where this will be true? If the check on 451 is false, then we don't have a prebuilt artifact, and if we don't have a container artifact here then doesn't that mean packaging failed in some way, as we should have one or the other? Why would we get to this point where we would expect to pick the prebuilt image off of the agent definition instead of a package artifact?

cli/azd/extensions/azure.ai.agents/internal/project/service_target_agent.go
```diff
@@ -486,86 +644,61 @@ func (p *AgentServiceTargetProvider) Deploy(
...
+	imageURL := agentDef.Image
+	if imageURL == "" {
+		return false, nil
 	}
 
-	err = agent_yaml.ValidateAgentDefinition(data)
-	if err != nil {
-		return false
+	// Non-interactive (CI/CD): always build from Dockerfile.
+	if noPrompt, _ := strconv.ParseBool(os.Getenv("AZD_NO_PROMPT")); noPrompt {
```
trangevi: This doesn't just come from an env var, we should be looking at the azd var. @JeffreyCA / @vhvb1989 , what's the correct way of accessing that from here, inside `deploy` effectively?

cli/azd/extensions/azure.ai.agents/internal/project/service_target_agent.go
```diff
@@ -486,86 +644,61 @@ func (p *AgentServiceTargetProvider) Deploy(
...
+		return false, nil
 	}
 
-	var genericTemplate map[string]any
-	if err := yaml.Unmarshal(data, &genericTemplate); err != nil {
-		return false
+	// Interactive: default to build so the pre-built path requires an explicit choice.
+	choices := []*azdext.SelectChoice{
+		{Value: "build", Label: "Build it for me"},
```
trangevi: Should this be "Build a new image for me" or something like that? @therealjohn

trangevi (changes requested):

jongio (commented):
Checked the 7 commits since my last review. Previous feedback addressed:

- **Test gap for pre-built prompt path**: now covered by `TestShouldUsePreBuiltImage_SelectsPreBuiltImage` (index 1) and `TestShouldUsePreBuiltImage_SelectsDockerfileBuild` (index 0). Both paths exercised.
- **`--no-prompt` default behavior**: non-interactive mode always builds from Dockerfile via explicit `AZD_NO_PROMPT` check, with a dedicated test (`TestShouldUsePreBuiltImage_NoPromptDefaultsToBuild`).

The caching removal and artifact-based flow are a cleaner design. The pre-built image decision now flows through the artifact pipeline (Package to Publish to Deploy) instead of a cached field on the provider struct.

No new blocking issues from me. Deferring to trangevi's active threads on the Publish fallback path and the `AZD_NO_PROMPT` access pattern.

cli/azd/extensions/azure.ai.agents/internal/project/service_target_agent.go
huimiu: Great catch. The decision to use the pre-built image should happen in `Package`, not again in `Publish`.

I removed the fallback logic in `Publish`. Now the flow is:

- `Package` decides whether to use the pre-built image or build from Dockerfile.
- If the user chooses the pre-built image, `Package` returns a pre-built image artifact.
- `Publish` only checks for that artifact and skips publishing when it exists.
- Otherwise, `Publish` continues with the normal container publish flow.

Also adding @therealjohn for awareness and to see if it makes sense.

cli/azd/extensions/azure.ai.agents/internal/project/service_target_agent.go
huimiu: I removed the explicit env var check and now rely on the azd prompt service behavior. Since the prompt sets SelectedIndex to 0, no-prompt mode will automatically return the default choice, which is build.

So the behavior is now:

- Interactive mode: user is prompted.
- Default choice: build a new image from Dockerfile.
- --no-prompt mode: azd returns the default choice automatically.
- No direct AZD_NO_PROMPT parsing in this code.

cli/azd/extensions/azure.ai.agents/internal/project/service_target_agent.go
huimiu: Updated to "Build a new image for me". Please let me know if you want another string @therealjohn

trangevi (approved):
Thanks for the updates!

therealjohn (approved):

PR #8136, opened 2026-05-11
fix: support Terraform/non-Bicep layers in execution graph (#8126)

### LINKED ISSUE #8126: Error deploying IaC with Terraform and layers
**Output from `azd version`**
azd version 1.25.0 (commit 208a5186a1c601e89ae8e37f9ef7b4e7037e4fc3) (stable)

**Describe the bug**
Using Terraform as IaC with layered infrastructure throws the following error when executing `azd up`. The error message is:

```pwsh
ERROR: analyzing layer dependencies: extracting outputs for layer "backend": reading F:\repos\work\...\demo-hub\demos\agents-demo\infra\backend\main.bicep: open F:\repos\work\...\demo-hub\demos\agents-demo\infra\backend\main.bicep: The system cannot find the file specified.
```

There is no `main.bicep`. All the IaC is Terraform.

**To Reproduce**
Create a simple Terraform IaC to deploy a remote backend and a simple infrastructure. Then create a `azure.yaml` as follows:

```yaml
name: agents-demo

services:
  genie-agent:
    project: ./src/Demo.Agent
    language: csharp
    host: containerapp
    resourceGroup: ${AZURE_RESOURCE_GROUP}
    resourceName: ${AZURE_CONTAINER_APP_NAME}
    docker:
      registry: ${AZURE_CONTAINER_REGISTRY_ENDPOINT}

infra:
  layers:
    - name: backend
      path: ./infra/backend
      provider: terraform
    - name: resources
      path: ./infra/resources
      provider: terraform
      dependsOn:
        - backend
```

**Expected behavior**
Both the backend hosting the storage account and blob container for Terraform remote state, and the resources with the infrastructure of the project are deployed with `azd up`.

**Environment**
    * Terraform v1.15.2
    * C# 14 and .NET 10
    * VS Code Insiders

jongio:
## Summary

Fixes #8126.

`AnalyzeLayerDependencies` crashes when infrastructure layers use `provider: terraform` (or any non-Bicep provider) because it unconditionally tries to read `.bicep` files for all layers.

### 1. Skip non-Bicep layers in static analysis (`layer_deps.go`)
- Added `isBicepLayer()` helper that checks whether a layer's provider is Bicep (or unspecified, which defaults to Bicep)
- Phase 1 (output discovery) and Phase 2 (parameter env-var refs) now skip non-Bicep layers
- Phase 3 (explicit `dependsOn` edges) remains universal — Terraform layers can still declare dependencies on other layers

### 2. Fix root provider inheritance (`project.go`)
- Fixed a pre-existing bug where the loop `for _, layer := range layers` modified a copy instead of the actual layer
- Changed to index-based loop so root-level `provider: terraform` correctly propagates to layers that don't specify their own provider

### Tests
- 4 new test cases in `layer_deps_test.go`: TerraformLayers, TerraformLayersParallel, MixedProviders, IsBicepLayer
- 2 new subtests in `project_config_test.go`: TestProjectConfigLayerProviderInheritance (inherits/preserves)

### DIFF
--- a/cli/azd/pkg/infra/provisioning/bicep/layer_deps.go
+++ b/cli/azd/pkg/infra/provisioning/bicep/layer_deps.go
@@ -16,6 +16,16 @@ import (
 	"github.com/azure/azure-dev/cli/azd/pkg/infra/provisioning"
 )
 
+// isBicepLayer reports whether a layer uses the Bicep provider (or defaults
+// to it). Non-Bicep layers (Terraform, Pulumi, etc.) are opaque to the
+// static Bicep analyzer: they produce no discoverable outputs and have no
+// .bicep/.bicepparam/.parameters.json files to scan. Those layers still
+// participate in the dependency graph via explicit dependsOn edges.
+func isBicepLayer(layer provisioning.Options) bool {
+	return layer.Provider == provisioning.NotSpecified ||
+		layer.Provider == provisioning.Bicep
+}
+
 // Package-level compiled regexes for output and env-var reference extraction.
 var (
 	bicepOutputRe   = regexp.MustCompile(`(?m)^\s*output\s+(\w+)\s+`)
@@ -130,10 +140,15 @@ func AnalyzeLayerDependencies(
 	}
 
 	// Phase 1 — Discover outputs from each layer's Bicep file.
+	// Non-Bicep layers (Terraform, Pulumi, etc.) are skipped: they have
+	// no .bicep file and their outputs cannot be statically discovered.
 	// Iterate layers (not resolved) so the loop index i is clearly bounded by
 	// len(layers) for static analyzers; resolved has the same length by
 	// construction above.
 	for i, layer := range layers {
+		if !isBicepLayer(layer) {
+			continue
+		}
 		opts := resolved[i]
 		bicepPath := resolveBicepPath(opts, projectPath)
 		outputs, err := extractBicepOutputs(ctx, bicepPath)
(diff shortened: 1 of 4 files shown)


jongio:
Addressing Copilot review comments:

**Thread 1 (layer_deps_test.go:887)** — Fixed the empty subtest name by adding explicit `name` fields to the table-driven test and removing the duplicate `""` entry (since `NotSpecified` is `""` by definition). Re loop variable capture: Go 1.22+ creates a new loop variable per iteration, so the classic `tt := tt` shadow is no longer needed (we use Go 1.26).

**Thread 2 (project.go:92)** — The root provider is already normalized via `ParseProvider()` on line 83. Per-layer `ParseProvider` validation is a reasonable enhancement but out of scope for this bugfix — the comparison `== NotSpecified` correctly catches the zero-value case, which is the only case where inheritance should apply. Happy to add per-layer validation as a follow-up if desired.

wbreza (approved):
## Code Review Summary

Clean, well-scoped bugfix with thorough tests. No issues found.

### What Looks Good

- **Correct bugfix**: The copy-vs-index range loop bug in `project.go` is a classic Go pitfall, properly fixed with index-based mutation
- **Clean separation of concerns**: `isBicepLayer()` is a well-documented, single-responsibility helper that correctly treats `NotSpecified` as Bicep (the project default)
- **Minimal blast radius**: Phases 1 & 2 skip non-Bicep layers while Phase 3 (`dependsOn`) remains universal — preserving cross-provider dependency declarations
- **Thorough testing**: 6 new test scenarios cover Terraform-only, parallel, mixed providers, provider classification, and inheritance (both propagation and override)
- **Defensive ParseProvider per-layer**: Even though `ParseProvider` is currently a no-op, calling it per-layer establishes the validation hook for future provider-kind restrictions

| Priority | Count |
|----------|-------|
| Critical | 0 |
| High | 0 |
| Medium | 0 |
| Low | 0 |
| **Total** | **0** |

hemarina (approved):
Did an independent pass on top of @wbreza's approval — concurs, no new findings. A couple of things worth calling out:

- Nice catch on the hidden copy-vs-index bug in `Parse()` — without that fix, root-provider inheritance was silently no-op'ing, so the new `isBicepLayer` skip wouldn't have triggered on its own. The two fixes are necessary together.
- `ParseProvider("")` is currently a pass-through, so the per-layer call is a free hardening hook for future provider validation with zero risk to existing YAMLs.
- `MixedProviders` test gives nice coverage of the `terraform → bicep` `dependsOn` path.

LGTM ✅ Thanks for the fix!

PR #8140, opened 2026-05-11
Create unified `pullrequest` pipeline

scbedd:
If we went with this, it would look like:

- Due to the limitation of `cli` which only has `livetests`, we need a single internal unified build that will account for the surface area
  - This will shift and contract, but still is based upon the existing build yaml. Less a rework and more a combination of the builds.
- ~~Need to account for the fact this this new pipeline isn't grandfathered into CFSClean~~
- Account for PR feedback / acceptance from existing team

This is related to Azure/azure-sdk-tools#15160

### DIFF
--- a/eng/pullrequest.yml
+++ b/eng/pullrequest.yml
@@ -35,14 +35,6 @@ extends:
               os: linux
             steps:
               - checkout: self
-                fetchDepth: 1
-
-              - pwsh: |
-                  # Shallow-fetch the target branch so get-changedfiles.ps1 can diff against it
-                  $targetBranch = "$(System.PullRequest.TargetBranchName)"
-                  if (-not $targetBranch) { $targetBranch = "main" }
-                  git fetch --no-tags --depth=1 origin $targetBranch
-                displayName: Fetch target branch
 
               - pwsh: |
                   # Get changed files using the shared helper (uses ADO PR env vars automatically)
--- a/eng/pullrequest.yml
+++ b/eng/pullrequest.yml
@@ -26,6 +26,7 @@ extends:
       - stage: Detect
         variables:
           - template: /eng/pipelines/templates/variables/globals.yml
+          - template: /eng/pipelines/templates/variables/image.yml
         jobs:
           - job: DiffAnalysis
             pool:
--- a/eng/pullrequest.yml
+++ b/eng/pullrequest.yml
@@ -8,7 +8,7 @@ trigger: none
 
 pr:
   branches:
-    include: ['*']
+    include: ['pipelinev3*']
   paths:
     include:
       - cli/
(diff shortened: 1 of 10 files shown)


eng/pullrequest.yml
```diff
@@ -0,0 +1,391 @@
...
+            pool:
+              name: $(LINUXPOOL)
+              image: $(LINUXVMIMAGE)
+              os: linux
+            steps:
+              - checkout: self
+                fetchDepth: 0
+
+              - pwsh: |
+                  # ADO checkout doesn't configure remote tracking refspecs, so a
```
scbedd, line 41: Given that we have to checkout to properly calculate the diff, I'm just going to put this in a real script instead of being entirely inline. I was trying to get creative with optimization of the `initialize` stage which is why this exists as an inline script.

eng/pullrequest.yml
```diff
@@ -0,0 +1,391 @@
...
+                  $cliSpecific = Test-PathMatch -Files $changedFiles `
+                    -IncludePrefixes @("cli/", "eng/pipelines/release-cli.yml", "eng/pipelines/templates/jobs/build-cli.yml", "eng/pipelines/templates/jobs/cross-build-cli.yml") `
+                    -ExcludePrefixes @("cli/azd/extensions/", "cli/azd/docs/", "cli/azd/.vscode/cspell", "cli/azd/AGENTS.md", "cli/azd/cmd/testdata/TestFigSpec.ts")
+                  $runCli = $cliSpecific -or $sharedAllChanged -or $goModChanged
+                  Write-Host "RunCli = $runCli"
+
+                  # ================================================================
+                  # Notice leg
+                  # ================================================================
+                  $runNotice = Test-ExactMatch -Files $changedFiles -Paths @("eng/pipelines/build-notice.yml")
```
jongio, line 123: This only catches changes to `build-notice.yml` itself but the Notice stage now uses `notice-build-and-test.yml`, `build-scan-cli.yml`, and `build-scan-vscode.yml`. Consider adding those to the detection, maybe with a `$sharedNoticePaths` set similar to `$sharedExtTemplatePaths`.

eng/pullrequest.yml
```diff
@@ -0,0 +1,391 @@
...
+                  # ================================================================
+                  $runAzDevOps = Test-PathMatch -Files $changedFiles `
+                    -IncludePrefixes @("ext/azuredevops/", "eng/pipelines/release-azuredevops.yml")
+                  Write-Host "RunAzDevOps = $runAzDevOps"
+
+                  # ================================================================
+                  # Extension legs
+                  # ================================================================
+                  $extensions = @(
+                    @{ Var = "RunExt_AzureAiAgents";    Path = "cli/azd/extensions/azure.ai.agents/";         Yml = "eng/pipelines/release-ext-azure-ai-agents.yml";              GoMod = $true  },
```
jongio, line 144: `GoMod = $true` never fires because `$goModChanged` checks for a root `go.mod` that doesn't exist. Each extension's own `go.mod` would be caught by `$extSpecific` (it's under the extension's path prefix), so this flag is only useful if you want to detect changes to `cli/azd/go.mod` affecting extensions that use replace directives. If that's the intent, you'd need to check `cli/azd/go.mod` explicitly instead of `go.mod`.

jongio (commented):
Draft pipeline consolidation feedback. The detect-and-fan-out approach is solid and should cut a lot of unnecessary CI time on PRs.

Three things to consider alongside what's already been flagged:

1. **Notice leg detection is too narrow** - `$runNotice` only checks for changes to `build-notice.yml`, but the Notice stage also depends on `notice-build-and-test.yml`, `build-scan-cli.yml`, and `build-scan-vscode.yml`. Changes to those sub-templates won't trigger the Notice leg.

2. **`GoMod` flag on extensions is dead code** - since `$goModChanged` checks for a root `go.mod` that doesn't exist (only `cli/azd/go.mod`), the `GoMod = $true` setting on extensions never fires. For CLI this is fine because `cli/azd/go.mod` falls under the `cli/` prefix. But if extensions depend on `cli/azd/go.mod` through workspace or replace directives, those changes won't be detected.

3. **PR branch filter** (`pipelinev3*`) needs updating before merge - just a reminder.

The `publish` to `templateContext.outputs` migration in `build-scan-cli.yml` and `notice-build-and-test.yml` looks correct for 1ES compliance.

PR #8146, opened 2026-05-12
Feature/code deploy zip upload

v1212:
## Summary

Adds **code deploy (ZIP upload)** support for hosted agents as an alternative to container-based deployment. This enables deploying Python agent source code directly without requiring Docker/ACR.

### Init Flow: 5 Supported Paths

The deploy mode prompt appears in all init flows when a Python project is detected. For non-Python projects, the code deploy option is hidden and the container path is used automatically. Users can re-run `azd ai agent init` at any time to switch between code and container modes. Template files (Dockerfile, etc.) are never deleted.

#### 1. New Project (Template) -> Container Deploy
```
azd ai agent init           # empty dir -> template selection
-> Language -> Python
-> Template -> select template
-> Deploy mode -> Container Image (Docker)
-> Model -> Use existing model deployment(s)
-> Subscription -> select
-> Foundry Project -> select
-> App Insights -> optional
-> Model deployment -> select
-> ACR -> select
-> Resources (CPU/Memory) -> select
-> Writes agent.yaml (no code_configuration), azure.yaml (with docker block)
```

#### 2. New Project (Template) -> Code Deploy
```
azd ai agent init           # empty dir -> template selection
-> Language -> Python
-> Template -> select template
-> Deploy mode -> Source Code (ZIP upload)
-> Runtime -> Python 3.12
-> Entry point -> main.py
-> Dependency resolution -> Remote build / Bundled
-> Model -> Use existing model deployment(s)
-> Subscription -> select
-> Foundry Project -> select
-> App Insights -> optional
-> Model deployment -> select
-> Resources (CPU/Memory) -> select (for local run compatibility)
-> Writes agent.yaml (with code_configuration), azure.yaml (no docker block)
-> Template Dockerfile preserved (unused)
```

#### 3. Existing Code -> Container Deploy
```
azd ai agent init           # non-empty dir with Python code
-> "Use the code in the current directory"
-> Agent name
-> Deploy mode -> Container Image (Docker)
-> Protocols
-> Model -> Use existing model deployment(s)
-> Subscription -> select
-> Foundry Project -> select
-> ACR -> select
-> Writes agent.yaml (no code_configuration), azure.yaml (with docker block)
```

#### 4. Existing Code -> Code Deploy
```
azd ai agent init           # non-empty dir with Python code
-> "Use the code in the current directory"
-> Agent name
-> Deploy mode -> Source Code (ZIP upload)
-> Runtime -> Python 3.12
-> Entry point -> main.py
[...]

PR #8161, opened 2026-05-13
feat(agents): add .NET code deploy support (dotnet_8/9/10 runtimes)

### LINKED ISSUE #7430: Add ability to deploy to hosted agents using code upload
Another way to deploy a hosted agent is similar to zip upload in App Service - you can deploy the code as an artifact, no dockerfile needed, etc.

v1212:
## Summary

Adds **.NET runtime support** (`dotnet_8`, `dotnet_9`, `dotnet_10`) to the code deploy (ZIP upload) path introduced in #8146. This enables deploying .NET agent source code directly via ZIP upload without requiring Docker/ACR, alongside the existing Python support.

---

### Deploy Mode Prompt

The code deploy option now appears for both Python **and** .NET projects (detected via `.csproj`/`.fsproj` files). For projects that are neither Python nor .NET, the code deploy option remains hidden.

#### Existing Code -> Code Deploy (.NET)
```
azd ai agent init           # non-empty dir with .csproj
-> "Use the code in the current directory"
-> Agent name
-> Deploy mode -> Source Code (ZIP upload)
-> Runtime -> .NET 9 / .NET 8 / .NET 10
-> Entry point -> HelloWorld.dll (auto-detected from .csproj)
-> Dependency resolution -> Remote build / Bundled
-> Protocols
-> Model -> Use existing model deployment(s)
-> Subscription -> select
-> Foundry Project -> select
-> Writes agent.yaml (with code_configuration, runtime: dotnet_9), azure.yaml (language: csharp, no docker block)
```

### Runtime Filtering

`promptCodeConfig()` now filters runtime choices based on detected project type:
- **.NET project only** (has `.csproj`/`.fsproj`, no `.py`/`requirements.txt`): Shows only .NET 9, .NET 8, .NET 10
- **Python project only** (has `.py`/`requirements.txt`, no `.csproj`): Shows only Python 3.11, 3.12, 3.13
- **Mixed or unknown**: Shows all 6 runtime options

### Language Detection in `azure.yaml`

Both init paths (template and from-code) correctly set `language: csharp` when a dotnet runtime is selected, fixing a bug where code deploy always hardcoded `language: python`.

---

### Bundled Mode (`dependency_resolution: bundled`)

For .NET agents in bundled mode, `packageDotnetBundled()`:
1. Runs `dotnet publish -c Release -r linux-x64 --self-contained false -o <temp-dir>`
2. ZIPs the published output (ready-to-run binaries for Linux)
3. Uploads via the same multipart form-data POST used for Python

### Remote Build Mode (`dependency_resolution: remote_build`)

For .NET agents in remote build mode:
1. ZIPs source code (`.cs`, `.csproj` files etc.) excluding build artifacts
2. Server-side Oryx build handles `dotnet publish` on the remote environment

### Entry Point Prefix

.NET agents use `dotnet <entry_point>` as the startup command (e.g., `["dotnet", "HelloWorld.dll"]`), while Python uses `python <entry_point>`.

### 250 MB ZIP Size Limit

Enforced for all code deploy packages (both Python and .NET) with a clear error message.

---

## Files Modified

| File | Change |
|------|--------|
| `init_from_code.go` | Add `isDotnetProject()` helper, `detectDefaultEntryPoint()` for `.dll`/`.csproj`, filter runtime choices by project type, update `showCodeDeploy` to include .NET, fix `noPrompt` default for dotnet |
| `init_from_code_test.go` | Unit tests for `detectDefaultEntryPoint()` with dotnet scenarios |
| `init.go` | Fix `language: python` hardcoding bug -- now checks runtime prefix for `csharp`; update `showCodeDeploy` to use `isDotnetProject()` |
| `map.go` | Map `dotnet_8/9/10` runtimes to API format with `"dotnet"` command prefix |
| `map_test.go` | Round-trip tests for dotnet code configuration API mapping |
| `service_target_agent.go` | Add `packageDotnetBundled()` with `dotnet publish` + output capture in errors; display `dotnet` prefix in deploy output; add `bytes` import; enforce 250 MB ZIP limit |
| `config.go` | Add TODO comment to `CodeDeployRegions` for future dynamic discovery |

---

## How to Build

```bash
go build ./...
go test ./...
go vet ./...
```

---

### Prerequisites
- Azure subscription with a Foundry project in a supported region (`westus2`, `canadacentral`, `northcentralus`)
- A model deployment (e.g. `gpt-4o`, `gpt-5.1`) in your Foundry project
- .NET SDK 9 installed locally (for bundled mode)
- `azd` CLI installed, logged in (`azd auth login`)
- Build the extension: `cd cli/azd/extensions/azure.ai.agents && azd x build`

### Test 1: .NET Code Deploy with Remote Build

```powershell
# 1. Clone a .NET agent sample (e.g., hello-world-dotnet-invocations)
mkdir test-dotnet-code-deploy && cd test-dotnet-code-deploy

# 2. Create a minimal .NET agent project with agent.yaml containing:
#   code_configuration:
#     runtime: dotnet_9
#     entry_point: HelloWorld.dll
#     dependency_resolution: remote_build

# 3. Init with existing Foundry project
azd ai agent init --project-id "<your-arm-resource-id>"
# Expected prompts:
#   1. Deploy mode -> Source Code (ZIP upload)
#   2. Runtime -> .NET 9 (only .NET options shown since it's a .NET project)
#   3. Entry point -> HelloWorld.dll (auto-detected from .csproj)
#   4. Dependency resolution -> Remote build
#   5. Protocols -> invocations
#   6. Model -> select existing deployment

# 4. Verify azure.yaml
Get-Content azure.yaml | Select-String "language"
# Expect: language: csharp

# 5. Deploy
azd deploy hello-world-dotnet-invocations
# Expect: "Packaging code" -> "Creating agent" -> "Agent is active!"

# 6. Invoke (wait ~30s for cold start)
azd ai agent invoke "What is 2+2?"
# Expect: streaming response with model reply
```

### Test 2: .NET Code Deploy with Bundled Mode

```powershell
# Same project as above, update agent.yaml:
#   dependency_resolution: bundled

azd deploy hello-world-dotnet-invocations
# Expect: "Publishing .NET project" -> "Packaging code" -> "Creating agent" -> done
# The deploy runs `dotnet publish -c Release -r linux-x64 --self-contained false` locally

azd ai agent invoke --new-session "Hello bundled dotnet!"
# Expect: streaming response
```

---

## Test Results

All configurations verified end-to-end on `canadacentral` (Foundry project `wujia-aifproject260512`):

| # | Runtime | Package Mode | Deploy | Invoke | Notes |
|---|---------|-------------|:---:|:---:|-------|
| 1 | dotnet_9 | remote_build | PASS (19s) | PASS (streaming SSE) | Model: gpt-5.1 |
| 2 | dotnet_9 | bundled | PASS | PASS | `dotnet publish` runs locally |
| 3 | python_3_12 | remote_build | PASS | PASS | Regression check (no change) |

### Invoke Output Example (Remote Build, dotnet_9)
```
Agent:    hello-world-dotnet-invocations (remote, invocations protocol)
Input:    "What is 2+2?"
Session:  (new -- server will assign)

Invocation:   inv_ec38240e65957f5600bNhzDBxxuw7IHJf7bxYXDxdsJZqtnDJW
Session:  8e0044805feaa1c000rE03tPFMgIHo0GrGeLFbQEE52lx8hrLX (assigned by server)
[hello-world-dotnet-invocations] {"type":"token","content":"2"}
{"type":"token","content":" +"}
{"type":"token","content":" 2"}
{"type":"token","content":" ="}
{"type":"token","content":" 4"}
{"type":"token","content":"."}
{"type":"done","full_text":"2 + 2 = 4."}
```

---

## Unit Tests

```
ok  azureaiagent/internal/cmd              14.084s
ok  azureaiagent/internal/pkg/agents/agent_api    21.724s
ok  azureaiagent/internal/pkg/agents/agent_yaml    5.329s
ok  azureaiagent/internal/project           5.604s
```

Specific test coverage for this PR:
- `TestDetectDefaultEntryPoint` -- verifies `.dll` detection from `dotnet publish` output and `.csproj` AssemblyName fallback
- `TestCodeConfigurationDotnetRoundTrip` -- verifies `dotnet_9` runtime maps to API format with `["dotnet", "HelloWorld.dll"]` entry point
- `TestCodeConfigurationDotnetRemoteBuild` -- verifies `dependency_resolution: remote_build` mapping for dotnet

---

## Notes

- **No impact on Python code deploy**: All Python paths are unchanged. Runtime filtering only affects the prompt UI.
- **No impact on container deploy**: Container paths remain completely untouched.
- **`--no-prompt` defaults**: For .NET projects, defaults to `dotnet_9` runtime. For Python projects, defaults to `python_3_12` (unchanged).
- **Entry point auto-detection**: Looks for `.csproj` `<AssemblyName>` property first, falls back to project directory name + `.dll`.
- **gosec G204 suppressed**: `exec.Command("dotnet", "publish", csprojPath, ...)` -- `csprojPath` is derived from user's local project directory, not external input.
- **250 MB limit**: ZIP packages exceeding 250 MB are rejected with a clear error before upload.
- **Supported regions**: Same as Python code deploy -- `westus2`, `canadacentral`, `northcentralus`.

---

## Dependencies

- Requires #8146 to merge first (this PR is rebased on top of it)
- After #8146 merges to `main`, this PR's diff will shrink to only the .NET-specific additions (~5 commits)

## Related

- Parent PR: #8146 (code deploy ZIP upload -- Python)
- Schema: [microsoft/AgentSchema#96](https://github.com/microsoft/AgentSchema/pull/96) -- `code_configuration` schema updated with .NET runtimes (`dotnet_8/9/10`), `.dll` entry point examples, and .NET build workflow descriptions

### DIFF
--- a/cli/azd/extensions/azure.ai.agents/internal/cmd/init_from_code.go
+++ b/cli/azd/extensions/azure.ai.agents/internal/cmd/init_from_code.go
@@ -1270,7 +1270,6 @@ func isPythonProject(dir string) bool {
 	return false
 }
 
-// isDotnetProject returns true if the directory contains a .csproj or .fsproj file.
 // isDotnetProject returns true if the directory contains a .csproj file.
 // NOTE: .fsproj (F#) is not yet supported by the packaging path (packageDotnetBundled/detectDefaultEntryPoint).
 func isDotnetProject(dir string) bool {
--- a/cli/azd/extensions/azure.ai.agents/internal/cmd/init_from_code.go
+++ b/cli/azd/extensions/azure.ai.agents/internal/cmd/init_from_code.go
@@ -470,12 +470,12 @@ func (a *InitFromCodeAction) createDefinitionFromLocalAgent(ctx context.Context)
 	agentKind := agent_yaml.AgentKindHosted
 
 	// Prompt user for deploy mode (container vs code)
-	// Code deploy is only available for Python projects
+	// Code deploy is available for Python and .NET projects
 	srcDir := a.flags.src
 	if srcDir == "" {
 		srcDir, _ = os.Getwd()
 	}
-	showCodeDeploy := isPythonProject(srcDir)
+	showCodeDeploy := isPythonProject(srcDir) || isDotnetProject(srcDir)
 	deployMode, err := promptDeployMode(ctx, a.azdClient, a.flags.noPrompt, showCodeDeploy)
 	if err != nil {
 		return nil, err
@@ -1106,21 +1106,47 @@ func promptCodeConfig(ctx context.Context, azdClient *azdext.AzdClient, srcDir s
 		srcDir = "."
 	}
 
-	// Prompt for runtime
-	runtimeChoices := []*azdext.SelectChoice{
-		{Label: "Python 3.11", Value: "python_3_11"},
-		{Label: "Python 3.12", Value: "python_3_12"},
-		{Label: "Python 3.13", Value: "python_3_13"},
-		{Label: ".NET 9", Value: "dotnet_9"},
-		{Label: ".NET 8", Value: "dotnet_8"},
-		{Label: ".NET 10", Value: "dotnet_10"},
+	// Prompt for runtime — filter choices based on detected project type
+	var runtimeChoices []*azdext.SelectChoice
+	isDotnet := isDotnetProject(srcDir)
+	isPython := isPythonProject(srcDir)
+
+	if isDotnet && !isPython {
+		runtimeChoices = []*azdext.SelectChoice{
+			{Label: ".NET 9", Value: "dotnet_9"},
+			{Label: ".NET 8", Value: "dotnet_8"},
+			{Label: ".NET 10", Value: "dotnet_10"},
+		}
+	} else if isPython && !isDotnet {
+		runtimeChoices = []*azdext.SelectChoice{
+			{Label: "Python 3.11", Value: "python_3_11"},
+			{Label: "Python 3.12", Value: "python_3_12"},
+			{Label: "Python 3.13", Value: "python_3_13"},
+		}
+	} else {
+		// Mixed or unknown — show all options
+		runtimeChoices = []*azdext.SelectChoice{
+			{Label: "Python 3.11", Value: "python_3_11"},
+			{Label: "Python 3.12", Value: "python_3_12"},
+			{Label: "Python 3.13", Value: "python_3_13"},
+			{Label: ".NET 9", Value: "dotnet_9"},
+			{Label: ".NET 8", Value: "dotnet_8"},
+			{Label: ".NET 10", Value: "dotnet_10"},
+		}
 	}
 
 	var runtime string
 	if noPrompt {
-		runtime = "python_3_12"
+		if isDotnet {
+			runtime = "dotnet_9"
+		} else {
+			runtime = "python_3_12"
+		}
 	} else {
-		defaultIdx := int32(1) // Python 3.12 is the default
+		defaultIdx := int32(0) // First item in the filtered list
+		if isPython && !isDotnet {
+			defaultIdx = 1 // Python 3.12
+		}
 		runtimeResp, err := azdClient.Prompt().Select(ctx, &azdext.SelectRequest{
 			Options: &azdext.SelectOptions{
 				Message:       "Select the runtime for your agent",
@@ -1210,6 +1236,23 @@ func isPythonProject(dir string) bool {
 	}
 	for _, e := range entries {
 		if !e.IsDir() && strings.HasSuffix(e.Name(), ".py") {
+			return true
+		}
+	}
+	return false
+}
+
+// isDotnetProject returns true if the directory contains a .csproj or .fsproj file.
+func isDotnetProject(dir string) bool {
+	if dir == "" {
+		dir = "."
+	}
+	entries, err := os.ReadDir(dir)
+	if err != nil {
+		return false
+	}
+	for _, e := range entries {
+		if !e.IsDir() && (strings.HasSuffix(e.Name(), ".csproj") || strings.HasSuffix(e.Name(), ".fsproj")) {
 			return true
 		}
 	}
--- a/cli/azd/extensions/azure.ai.agents/internal/cmd/init_from_code.go
+++ b/cli/azd/extensions/azure.ai.agents/internal/cmd/init_from_code.go
@@ -860,6 +860,17 @@ func (a *InitFromCodeAction) addToProject(ctx context.Context, targetDir string,
 	language := "python"
 	if !isCodeDeploy {
 		language = "docker"
+	} else {
+		// Detect language from agent.yaml runtime
+		agentYamlPath2 := filepath.Join(a.projectConfig.Path, targetDir, "agent.yaml")
+		if data, err := os.ReadFile(agentYamlPath2); err == nil { //nolint:gosec // path from project config
+			var agentDef2 agent_yaml.ContainerAgent
+			if err := yaml.Unmarshal(data, &agentDef2); err == nil &&
+				agentDef2.CodeConfiguration != nil &&
+				strings.HasPrefix(agentDef2.CodeConfiguration.Runtime, "dotnet_") {
+				language = "csharp"
+			}
+		}
 	}
 
 	serviceConfig := &azdext.ServiceConfig{
@@ -899,7 +910,11 @@ func deriveStartupCommand(projectPath, targetDir string) string {
 	if data, err := os.ReadFile(agentYamlPath); err == nil { //nolint:gosec // path is constructed from project config
 		var agentDef agent_yaml.ContainerAgent
 		if err := yaml.Unmarshal(data, &agentDef); err == nil && agentDef.CodeConfiguration != nil {
-			return "python " + agentDef.CodeConfiguration.EntryPoint
+			cmdPrefix := "python"
+			if strings.HasPrefix(agentDef.CodeConfiguration.Runtime, "dotnet_") {
+				cmdPrefix = "dotnet"
+			}
+			return cmdPrefix + " " + agentDef.CodeConfiguration.EntryPoint
 		}
 	}
 	return "python main.py"
@@ -1062,6 +1077,28 @@ func promptDeployMode(ctx context.Context, azdClient *azdext.AzdClient, noPrompt
 	return deployModeChoices[*deployModeResp.Value].Value, nil
 }
 
+// detectDefaultEntryPoint returns a sensible default entry point based on the runtime and source directory.
+func detectDefaultEntryPoint(srcDir, runtime string) string {
+	if strings.HasPrefix(runtime, "dotnet_") {
+		// Look for .csproj file and derive DLL name
+		entries, err := os.ReadDir(srcDir)
+		if err == nil {
+			for _, e := range entries {
+				if !e.IsDir() && strings.HasSuffix(e.Name(), ".csproj") {
+					return strings.TrimSuffix(e.Name(), ".csproj") + ".dll"
+				}
+			}
+		}
+		return "App.dll"
+	}
+
+	// Python default
+	if _, err := os.Stat(filepath.Join(srcDir, "app.py")); err == nil {
+		return "app.py"
+	}
+	return "main.py"
+}
+
 // promptCodeConfig prompts for code deploy configuration (runtime, entry point,
 // dependency resolution). When noPrompt is true, defaults are used without prompting.
 func promptCodeConfig(ctx context.Context, azdClient *azdext.AzdClient, srcDir string, noPrompt bool) (*agent_yaml.CodeConfiguration, error) {
@@ -1074,6 +1111,9 @@ func promptCodeConfig(ctx context.Context, azdClient *azdext.AzdClient, srcDir s
 		{Label: "Python 3.11", Value: "python_3_11"},
 		{Label: "Python 3.12", Value: "python_3_12"},
 		{Label: "Python 3.13", Value: "python_3_13"},
+		{Label: ".NET 9", Value: "dotnet_9"},
+		{Label: ".NET 8", Value: "dotnet_8"},
+		{Label: ".NET 10", Value: "dotnet_10"},
 	}
 
 	var runtime string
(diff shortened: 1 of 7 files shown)


trangevi:

cli/azd/extensions/azure.ai.agents/internal/cmd/init_from_code.go
```diff
@@ -860,6 +860,17 @@ func (a *InitFromCodeAction) addToProject(ctx context.Context, targetDir string,
 	language := "python"
 	if !isCodeDeploy {
 		language = "docker"
+	} else {
+		// Detect language from agent.yaml runtime
+		agentYamlPath2 := filepath.Join(a.projectConfig.Path, targetDir, "agent.yaml")
```
trangevi: Why is this `agentYamlPath2`? same with `agentDef2`? If this is the second time we're doing these operations, can we make use of the original objects instead of going through this process again?

cli/azd/extensions/azure.ai.agents/internal/cmd/init_from_code.go
```diff
@@ -1062,25 +1077,76 @@ func promptDeployMode(ctx context.Context, azdClient *azdext.AzdClient, noPrompt
 	return deployModeChoices[*deployModeResp.Value].Value, nil
 }
 
+// detectDefaultEntryPoint returns a sensible default entry point based on the runtime and source directory.
+func detectDefaultEntryPoint(srcDir, runtime string) string {
```
trangevi, line 1079: Do we have anything similar for the `run` command which could be reused? If not, please file an item to look at reusing this as a part of that command

cli/azd/extensions/azure.ai.agents/internal/cmd/init_from_code.go
```diff
@@ -1062,25 +1077,76 @@ func promptDeployMode(ctx context.Context, azdClient *azdext.AzdClient, noPrompt
...
-	runtimeChoices := []*azdext.SelectChoice{
-		{Label: "Python 3.11", Value: "python_3_11"},
-		{Label: "Python 3.12", Value: "python_3_12"},
-		{Label: "Python 3.13", Value: "python_3_13"},
+	// Prompt for runtime — filter choices based on detected project type
+	var runtimeChoices []*azdext.SelectChoice
+	isDotnet := isDotnetProject(srcDir)
+	isPython := isPythonProject(srcDir)
+
+	if isDotnet && !isPython {
```
trangevi, line 1142: @therealjohn , any concerns with having this coded into the extension? Will require an extension update and release any time a new version is released, but will also likely reduce user error from a free form string input.

cli/azd/extensions/azure.ai.agents/internal/project/config.go
```diff
@@ -18,6 +18,7 @@ const (
 )
 
 // CodeDeployRegions lists the regions that currently support code deploy (ZIP upload).
+// TODO: replace with dynamic region discovery API when available.
 var CodeDeployRegions = []string{"westus2", "canadacentral", "northcentralus"}
```
trangevi, line 22: What's the timeline on full region support?

trangevi: And will this list of regions ever differ from the supported regions for general hosted agent deployment after full support?

For hosted agent region support, we use a file to track, and make a github call to get that file, so that we don't have to release a new extension version just to pick up region changes. We should probably look into doing similar here, or more preferably just using the same file because hopefully the regions match

cli/azd/extensions/azure.ai.agents/internal/project/service_target_agent.go
```diff
@@ -948,6 +950,21 @@ func (p *AgentServiceTargetProvider) packageCodeDeploy(serviceConfig *azdext.Ser
...
+			if agentDef.CodeConfiguration.DependencyResolution != nil {
+				isBundled = *agentDef.CodeConfiguration.DependencyResolution == "bundled"
+			}
+			if isDotnet && isBundled {
+				return p.packageDotnetBundled(srcDir)
+			}
+		}
+	}
+
 	// Exclusion patterns
```
trangevi, line 968: Making a note here so it's tracked, please create a work item to track the follow up of picking this out of some form of .ignore file

trangevi (changes requested):

cli/azd/extensions/azure.ai.agents/internal/cmd/init_from_code.go
therealjohn: Yeah, should be fine. Could be worked around in the azure.yaml if we are late on a release.

cli/azd/extensions/azure.ai.agents/internal/cmd/init_from_code.go
```diff
@@ -1062,25 +1077,76 @@ func promptDeployMode(ctx context.Context, azdClient *azdext.AzdClient, noPrompt
...
+	if strings.HasPrefix(runtime, "dotnet_") {
+		// Look for .csproj file and derive DLL name
+		entries, err := os.ReadDir(srcDir)
+		if err == nil {
+			for _, e := range entries {
+				if !e.IsDir() && strings.HasSuffix(e.Name(), ".csproj") {
+					return strings.TrimSuffix(e.Name(), ".csproj") + ".dll"
+				}
+			}
+		}
```
jongio, line 1097: This only searches for `.csproj` but `isDotnetProject()` also accepts `.fsproj`. An F# project will be detected as dotnet and land here, but won't match - falling back to `App.dll` which is likely wrong for F# (typically `<ProjectName>.dll` from the fsproj). Either add `.fsproj` handling or remove `.fsproj` from `isDotnetProject()` until the packaging path supports it.

cli/azd/extensions/azure.ai.agents/internal/pkg/agents/agent_yaml/map.go
```diff
@@ -357,7 +357,12 @@ func CreateHostedAgentAPIRequest(hostedAgent ContainerAgent, buildConfig *AgentB
 
 	// Code deploy path
 	if hostedAgent.CodeConfiguration != nil {
-		entryPoint := []string{"python", hostedAgent.CodeConfiguration.EntryPoint}
+		// Derive the command prefix from the runtime (e.g. "python" for python_3_12, "dotnet" for dotnet_9).
+		cmdPrefix := "python"
+		if strings.HasPrefix(hostedAgent.CodeConfiguration.Runtime, "dotnet_") {
```
jongio: This is the third place deriving cmd prefix from a runtime string (also in `deriveStartupCommand` and `deployHostedCodeAgent`). Consider extracting a shared helper - e.g. `func RuntimeCmdPrefix(runtime string) string` - so they can't drift apart when new runtime families are added.

jongio (commented):
Two items not covered by the existing feedback:

1. `isDotnetProject()` accepts `.fsproj` files, but `detectDefaultEntryPoint()` and `packageDotnetBundled()` only scan for `.csproj`. An F# project enters the dotnet path but then either gets the wrong fallback ("App.dll") or fails with "no .csproj file found." Either drop `.fsproj` from detection until it's actually supported, or handle it in both places.

2. The `strings.HasPrefix(runtime, "dotnet_") -> "dotnet"` logic is now in three places (map.go, deriveStartupCommand, deployHostedCodeAgent). When a third runtime family arrives (java_?), all three need updating in lockstep. A small helper like `func runtimeCmdPrefix(runtime string) string` on CodeConfiguration or as a package-level func would keep them in sync.

cli/azd/extensions/azure.ai.agents/internal/cmd/init_from_code.go
v1212: Fixed — renamed to `langDetectPath`/`langDef` to clarify intent. This is a second read of agent.yaml *after* it's been written (to detect the runtime language for azure.yaml's `language:` field), so the original objects can't be reused since `code_configuration` didn't exist before the write.

cli/azd/extensions/azure.ai.agents/internal/cmd/init_from_code.go
v1212: The `run` command doesn't have similar logic today. Added a TODO to track reusing this as part of that command in a follow-up.

cli/azd/extensions/azure.ai.agents/internal/cmd/init_from_code.go
v1212: Acknowledged — hardcoding is acceptable per John's confirmation. The azure.yaml `runtime:` field can serve as a workaround if the extension release lags behind a new runtime version.

cli/azd/extensions/azure.ai.agents/internal/project/config.go
v1212: Currently 3 regions (westus2, canadacentral, northcentralus). Checked with Santhosh — he will follow up with PM and confirm the timeline, likely within a week after bug bash.

v1212: Agreed — the code deploy region list should ideally use the same GitHub-file-based dynamic discovery mechanism as hosted agent regions. From the service side, it already returns an unsupported-region error, so the client-side list is mainly for a better UX (early validation + region suggestions). Will look into maintaining this in a shared GitHub file in a follow-up once we confirm the region sets are identical.

cli/azd/extensions/azure.ai.agents/internal/project/service_target_agent.go
v1212: Added a TODO comment in code. Created work item to track: https://github.com/Azure/azure-dev/issues/8170

cli/azd/extensions/azure.ai.agents/internal/cmd/init_from_code.go
v1212: Fixed — removed `.fsproj` from `isDotnetProject()` until the packaging path (`packageDotnetBundled`/`detectDefaultEntryPoint`) fully supports F#. Added a comment noting this limitation.

cli/azd/extensions/azure.ai.agents/internal/pkg/agents/agent_yaml/map.go
v1212: Fixed — extracted `RuntimeCmdPrefix(runtime string) string` in `agent_yaml` package and replaced all 3 inline occurrences (`map.go`, `deriveStartupCommand`, `deployHostedCodeAgent`).

therealjohn (approved):

therealjohn:
This completes and fixes https://github.com/Azure/azure-dev/issues/7430

trangevi (approved):

trangevi:
/check-enforcer override

The build is showing complete, status is just not being reported

PR #8189, opened 2026-05-14
Add AI agent init naming safeguards

### LINKED ISSUE #8107: Allow `azd ai agent init` to set a unique agent name and warn before versioning an existing agent
Hosted Agents are unique by name, but if you `init` the same sample project, you end up overwriting / new version of the previous agent instead of a new agent. 

 Agent Name is the primary key for an agent in Foundry so if you pick the same sample, because the name is the same in agent.yaml, it will be the same agent. You can't have duplicate names. 

This is not expected, but it is working as designed. Devs need a way to specify the name, or at least surface that its goign to produce a new version of an existing agent and not a new agent.

glharper:
## Summary
- add --agent-name to azd ai agent init for setting the Foundry agent identity
- write selected agent names into generated agent.yaml for manifest and from-code init flows
- warn when init or deploy will version an existing Foundry agent instead of creating a separate agent
- add tests for agent-name validation, manifest updates, existing-agent checks, and deploy warnings

Fixes #8107


cli/azd/extensions/azure.ai.agents/internal/cmd/init.go
```diff
@@ -205,6 +207,222 @@ func parseAuthStatusJSON(data []byte) (string, error) {
...
+	case agent_yaml.ContainerAgent:
+		return t.Name, nil
+	case *agent_yaml.ContainerAgent:
+		if t == nil {
+			return "", fmt.Errorf("agent template is nil")
+		}
+		return t.Name, nil
+	case agent_yaml.Workflow:
+		return t.Name, nil
+	case *agent_yaml.Workflow:
```
glharper: All four cases in this switch can be concatenated using fallthrough

cli/azd/extensions/azure.ai.agents/internal/cmd/init.go
```diff
@@ -205,6 +207,222 @@ func parseAuthStatusJSON(data []byte) (string, error) {
...
+		}
+		t.Name = agentName
+	case agent_yaml.Workflow:
+		t.Name = agentName
+		agentManifest.Template = t
+	case *agent_yaml.Workflow:
+		if t == nil {
+			return fmt.Errorf("agent template is nil")
+		}
+		t.Name = agentName
```
glharper: All four cases in this switch can be concatenated using fallthrough

glharper:
@copilot apply changes based on the comments in [this thread](https://github.com/Azure/azure-dev/pull/8189#pullrequestreview-4298350016)

glharper:

cli/azd/extensions/azure.ai.agents/internal/cmd/init_test.go
```diff
@@ -6,21 +6,214 @@ package cmd
...
 
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
 	"github.com/azure/azure-dev/cli/azd/pkg/azdext"
 	"github.com/spf13/cobra"
 	"google.golang.org/genproto/googleapis/rpc/errdetails"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 )
 
+type fakeAgentGetter struct {
```
jongio: `low` `fakeAgentGetter` here and `fakeAgentExistenceClient` in `service_target_agent_test.go` are structurally identical - both test `agents.AgentExists` with the same logic. Since `AgentExists` lives in the `agents` package, its unit tests could live in `existence_test.go` there, and each consumer test file would focus on its own integration.

cli/azd/extensions/azure.ai.agents/internal/cmd/init.go
```diff
@@ -205,6 +208,206 @@ func parseAuthStatusJSON(data []byte) (string, error) {
...
+) (string, error) {
+	if azdClient == nil || environment == nil || environment.Name == "" || credential == nil {
+		return agentName, nil
+	}
+
+	endpointResp, err := azdClient.Environment().GetValue(ctx, &azdext.GetEnvRequest{
+		EnvName: environment.Name,
+		Key:     "AZURE_AI_PROJECT_ENDPOINT",
+	})
+	if err != nil || endpointResp == nil || endpointResp.Value == "" {
```
jongio: `low` When `GetValue` fails or returns empty, this silently returns the original name with no signal. A `log.Printf` here would help debugging when the naming check doesn't fire as expected.

jongio (commented):
Reviewed 7 files (+605/-14). Clean approach to the naming safeguards. The `existence.go` extraction is well-placed, and the init flow integration is solid. A few minor observations below.

cli/azd/extensions/azure.ai.agents/internal/cmd/init.go
```diff
@@ -205,6 +208,206 @@ func parseAuthStatusJSON(data []byte) (string, error) {
...
+			return nil, fmt.Errorf("agent template is nil")
+		}
+		update(&t.AgentDefinition)
+		return t, nil
+	default:
+		return nil, fmt.Errorf("unsupported agent template type %T", template)
+	}
+}
+
+func nextAgentNameSuggestion(agentName string) string {
```
jongio, line 323: `nit` This always suggests `-2` as a suffix. If `my-agent-2` also exists, the next suggestion becomes `my-agent-2-2`. Not a big deal since users can type any name, but an increment-aware suffix would be slightly smoother UX.

wbreza (commented):
Re-review on commit 665af0f. All prior reviewer feedback from @jongio and the Copilot reviewer is cleanly addressed in this push (logging on the silent-skip path, fake-client consolidation into `agents/existence_test.go`, and the `-2`-suffix increment logic). Nice refactor extracting `AgentExists` / `ExistingAgentWarning` into `internal/pkg/agents`.

A few new items below — none blocking, but the two mediums are worth a look before merge.

### 🟡 Medium — Warning routed to stdout in init, stderr in deploy

`internal/cmd/init.go:420,422` use `fmt.Printf` for the existing-agent warning, while `internal/project/service_target_agent.go:840` writes the same warning via `fmt.Fprintf(os.Stderr, ...)`. When init''s stdout is redirected (scripts, extension-host structured-output capture, piping), the warning silently disappears. Suggest routing both through stderr for consistency.

```go
// init.go:420 — currently
fmt.Printf("%s", agents.ExistingAgentWarning(agentName))
// suggestion
fmt.Fprintf(os.Stderr, "%s", agents.ExistingAgentWarning(agentName))
```

### 🟡 Medium — Invalid name at re-prompt hard-exits init

In `resolveExistingAgentNameConflict` (`internal/cmd/init.go:458–461`), if the user types an invalid name at the rename prompt (e.g. `my_agent`), `validateInitAgentName` returns a structured `exterrors.Validation` and the function returns instead of looping. The user loses all init state and has to restart. Recommend printing the validation error to stderr and `continue`-ing the loop so the user gets another attempt.

### 🔵 Low — `resolveExistingAgentNameConflict` is untested as a unit

`AgentExists` and `nextAgentNameSuggestion` have good direct coverage, but the orchestrating function — `noPrompt` short-circuit, validation-on-retry, `AgentExists` error propagation, conflict-then-accept paths — has no test. A table test with a fake `AzdClient` + fake `AgentChecker` would close the gap.

### ⚪ Nit — `regexp.MustCompile` inside `ValidateAgentName`

`internal/pkg/agents/agent_yaml/parse.go:429` recompiles the validation regex on every call; with the new retry loop it can be called multiple times per init. Worth lifting to a package-level `var`.

---

LGTM overall once the stdout/stderr fix lands. Thanks for the clean extraction work.

cli/azd/extensions/azure.ai.agents/internal/cmd/init.go
```diff
@@ -205,6 +208,260 @@ func parseAuthStatusJSON(data []byte) (string, error) {
...
+				"To create a separate agent, re-run init with --agent-name <unique-name>.\n",
+			))
+			return agentName, nil
+		}
+
+		confirmResp, err := azdClient.Prompt().Confirm(ctx, &azdext.ConfirmRequest{
+			Options: &azdext.ConfirmOptions{
+				Message:      "Continue with this existing agent name?",
+				DefaultValue: new(false),
+				HelpMessage:  "Choose no to enter a different Foundry agent name before agent.yaml is written.",
```
trangevi: I'd suggest just truncating this to `Choose no to enter a different Foundry agent name.` The agent.yaml is somewhat of an implementation detail here that will just introduce confusion.

cli/azd/extensions/azure.ai.agents/internal/project/service_target_agent.go
```diff
@@ -1642,6 +1647,14 @@ func (p *AgentServiceTargetProvider) createAgent(
 		p.credential,
 	)
 
+	exists, err := agents.AgentExists(ctx, agentClient, request.Name, agentAPIVersion)
+	if err != nil {
+		return nil, fmt.Errorf("failed to check if agent exists: %w", err)
```
trangevi: I don't think we should fail if something goes wrong here. It's just intended as a warning anyway, this can be updated to match.

trangevi (changes requested):

cli/azd/extensions/azure.ai.agents/internal/project/service_target_agent.go
glharper: [Updated](https://github.com/Azure/azure-dev/pull/8189/commits/852d290008d72a8b570eccb4c36e0868b3333086)

cli/azd/extensions/azure.ai.agents/internal/cmd/init.go
glharper: [Updated](https://github.com/Azure/azure-dev/pull/8189/commits/852d290008d72a8b570eccb4c36e0868b3333086)

trangevi (approved):

PR #7776, opened 2026-04-17
feat(exegraph): graph-driven execution engine for up/provision/deploy

jongio:
## Related Issues

Fixes #6291 ΓÇö Support Parallel Provisioning of Infrastructure Layers
Fixes #5294 ΓÇö Create layer deployment orchestration engine

Related:
- #7659 ΓÇö [EPIC] Improve provision and deployment speed
- #7526 ΓÇö azd Performance Improvements: Parallelism, Startup, and E2E Latency

## Summary

Introduces `pkg/exegraph` ΓÇö a general-purpose DAG execution engine ΓÇö and rebuilds `azd up`, `azd provision`, and `azd deploy` on top of it. Independent work now runs concurrently with dependency-aware scheduling instead of strictly sequentially. No opt-in flag; single-node cases (single-layer provision, preview, single service) run as one-node graphs through the same code path with trivial overhead.

User-authored `workflows.up:` still runs on `workflow.Runner`; phase-scoped DAGs apply inside each sub-command it invokes.

**Scope:** 76 files changed, +10,338 / ΓêÆ779.

## Motivation

Today `provision`, `deploy`, and `up` execute sequentially:
- Bicep layers deploy one at a time.
- Services package, publish, and deploy in series.
- Wall-clock time scales linearly with resource/service count.

Exegraph runs independent work concurrently, bounded by declared dependencies, with a single unified orchestration for `up` so packaging can overlap with provisioning automatically.

### Engine ΓÇö `cli/azd/pkg/exegraph/`

- `Graph` / `Step` primitives with three-color DFS cycle detection and transitive-dependent priority (critical-path bias).
- Goroutine-per-ready-node scheduler with `FailFast` / `ContinueOnError` policies and uniform per-step timeout via `RunOptions.StepTimeout`.
- Cancellation classification: parent-cancel ΓåÆ skip; per-step timeout ΓåÆ failure.
- Adaptive polling with exponential backoff and ARM 429 throttle detection for Bicep deployments.
- Kahn-level resolution for multi-layer provision.

### Orchestrators ΓÇö `cli/azd/internal/cmd/`

- `up_graph.go`, `provision_graph.go`, `deploy_graph` built on the engine.
- `service_graph.go` builds service nodes for deploy and up; `project_hooks.go` builds `cmdhook-*` synthetic nodes for `preprovision` / `postprovision` / `predeploy` / `postdeploy`.
- Middleware fires only `preup` / `postup` for `azd up`, so project hooks never double-fire.
[...]

PR #8195, opened 2026-05-15
fix: resolve parallel dotnet publish race in Aspire deploys

### LINKED ISSUE #8177: Parallel packaging in azd 1.25 causes "dotnet publish" to race on shared project references (MSB4018)

**Output from `azd version`**

```
azd version 1.25.0 (commit 208a5186a1c601e89ae8e37f9ef7b4e7037e4fc3) (stable)
```

**Describe the bug**

In azd 1.25.0+, `azd deploy` (and `azd up`) runs `dotnet publish` for all services in parallel during the Packaging/Publishing phase. Per the [release blog post](https://blog.jongallant.com/2026/05/azd-parallel-provisioning-and-deployment/), this is by design: *"Packaging and publishing always run in parallel regardless of `uses` — it's only the deploy step that gates on these edges."*

This breaks .NET Aspire projects where multiple services share a common class library via `<ProjectReference>`. Each parallel `dotnet publish` invocation walks into the shared project and tries to write to the same `obj/Release/<tfm>/<SharedLib>.dll`, producing an `MSB4018` / `System.IO.IOException: The process cannot access the file ... because it is being used by another process`.

There is no opt-out: `uses` only controls the deploy step, not packaging/publishing, so users cannot avoid this race even when they know about the shared reference.

**To Reproduce**

1. Create an Aspire app host with 2+ services (e.g. a web API, a web app, and a migration console) where two or more reference a shared class library project, e.g. `MyApp.Common.Contracts.csproj`.
2. Use azd 1.25.x.
3. Run `azd deploy` (or `azd up`).
4. Observe parallel packaging kick off `dotnet publish` on each service simultaneously, racing on the shared project's `obj/Release/<tfm>/*.dll`.

Trimmed pipeline log:

```
  redis: Packaging
  myapp-apps-webapi: Packaging
  myapp-apps-webapp: Packaging
  redis: Publishing
  myapp-apps-webapp: Publishing
  myapp-apps-webapi: Publishing
  myapp-migrationconsole: Packaging
  myapp-migrationconsole: Publishing
  ...
  myapp-migrationconsole: Failed (step "deploy-myapp-migrationconsole" failed: deploying service myapp-migrationconsole: failed deploying service 'myapp-migrationconsole': publishing container: dotnet publish on project '/azp/_work/1/s/src/MyApp.MigrationConsole/MyApp.MigrationConsole.csproj' failed: exit code: 1, stdout:
, stderr: /azp/_work/_tool/dotnet/sdk/10.0.300/Sdks/Microsoft.NET.Sdk/targets/Microsoft.NET.Sdk.targets(1200,5): error MSB4018: The "GetAssemblyAttributes" task failed unexpectedly. [/azp/_work/1/s/src/MyApp.Common.Contracts/MyApp.Common.Contracts.csproj]
/azp/_work/_tool/dotnet/sdk/10.0.300/Sdks/Microsoft.NET.Sdk/targets/Microsoft.NET.Sdk.targets(1200,5): error MSB4018: System.IO.IOException: The process cannot access the file '/azp/_work/1/s/src/MyApp.Common.Contracts/obj/Release/net10.0/MyApp.Common.Contracts.dll' because it is being used by another process. [/azp/_work/1/s/src/MyApp.Common.Contracts/MyApp.Common.Contracts.csproj]
```

The failure is non-deterministic — which service "wins" and which fails varies between runs. On azd 1.24 and earlier, the same project deploys cleanly because publish steps were serialized.

**Expected behavior**

Either:

1. `dotnet publish` invocations for projects that share `<ProjectReference>` ancestors should be serialized (azd could detect overlapping project graphs during the analyze step), **or**
2. The `uses` field (or a new field) should gate the packaging/publishing phase as well, not only deploy, so users can opt out of the race when they know their projects share references, **or**
3. At minimum, an env var like `AZD_PACKAGE_CONCURRENCY=1` to force serial packaging, mirroring `AZD_DEPLOY_CONCURRENCY`.

**Environment**

- Language: C# / .NET 10.0.300 (Aspire app host + multiple class-library-sharing projects)
- OS: <linux / azure pipelines hosted agent>
- CI: Azure Pipelines (`/azp/_work/...`)
- IDE: <n/a — failure is on pipeline agent>

**Additional context**

- Worked fine on azd <= 1.24.
- Adding `uses` to `azure.yaml` does not help — packaging/publishing run in parallel regardless per the [1.25.0 release blog](https://blog.jongallant.com/2026/05/azd-parallel-provisioning-and-deployment/).
-  `AZD_DEPLOY_CONCURRENCY=1` looks like a workaround, but it does not solve the issue.

jongio:
Fixes #8177

## Problem

When multiple Aspire services sharing `<ProjectReference>` dependencies are deployed in parallel, their concurrent `dotnet publish` invocations race on shared `obj/` directories, producing MSB4018/IOException errors.

## Approach

Pass `--artifacts-path <per-service-temp-dir>` to `dotnet publish`, redirecting ALL intermediate build outputs (obj/, bin/) to an isolated directory per service. This eliminates the file race entirely while preserving full deploy parallelism.

**Why --artifacts-path?**
- First-class .NET 8+ SDK feature designed for exactly this isolation scenario
- Aspire requires .NET 8+, so the flag is always available
- Zero serialization — all services deploy fully in parallel including the dotnet publish phase
- Trade-off: shared project references get compiled once per service (redundant but fast vs the Azure deployment that follows)

**Fallback**: If temp dir creation fails (disk full, permissions), a mutex serializes only the `dotnet publish` phase while the Azure deployment portion remains parallel.

## Changes

- `dotnet.go` — `appendArtifactsPath()` + context helpers (`ContextWithArtifactsPath`/`ArtifactsPathFromContext`)
- `service_target_dotnet_containerapp.go` — Deploy() creates per-service temp dir, injects via context; explicit lock/enable for mutex fallback (not defer, to avoid holding across Azure deployment)
- `build_gate.go` (new) — Context helpers for build gate mutex
- `service_graph.go` — Injects mutex via context instead of graph edges; deploy steps remain parallel at graph level
- Tests rewritten to verify parallel topology

## Non-Aspire impact

None. The artifacts path and build gate are only injected when the service has a `DotNetContainerApp` configuration. Non-Aspire services get vanilla context with no behavioral change.

## Future optimization

A `dotnet build --artifacts-path` pass once upfront, followed by `dotnet publish --no-build`, would eliminate redundant shared-library compilation. Left for a follow-up since the current approach is correct and the redundant compilation is fast relative to the Azure deployment step.

### DIFF
--- a/cli/azd/internal/cmd/aspire_gate.go
+++ b/cli/azd/internal/cmd/aspire_gate.go
@@ -6,21 +6,25 @@ package cmd
 import "github.com/azure/azure-dev/cli/azd/pkg/project"
 
 // aspireBuildGateKey is the gate-key policy used by `azd deploy` and `azd up`
-// to serialize deploys for services that share a single .NET AppHost build.
+// to group Aspire services that share a single .NET AppHost build.
 //
 // It returns a constant non-empty key ("aspire") for every service whose
 // [project.ServiceConfig.DotNetContainerApp] is populated by the Aspire
 // importer, and "" for every other service.
 //
-// The effect: the first such service's deploy step acts as the build gate,
-// and every subsequent Aspire service's deploy depends on it, so the
-// AppHost's shared build runs exactly once while deploy-side work for other
-// services continues in parallel.
+// The effect: all Aspire services in the same gate group receive a shared
+// mutex via context. The deploy target
+// ([project.DotNetContainerAppTarget.Deploy]) uses this as a signal to
+// create a per-service --artifacts-path temp directory, so concurrent
+// dotnet publish invocations write to isolated intermediate output trees
+// and run fully in parallel. Only if the temp-dir creation fails does the
+// mutex act as a serialization fallback — the normal path is zero
+// serialization.
 //
 // This is the *only* Aspire-specific policy the execution graph layer
 // carries; the graph builder itself (see service_graph.go) is Aspire-
 // agnostic and consumes any opaque grouping key. Extensions that own their
-// own "first-wins" build lane can supply a different callback through
+// own sequential build lane can supply a different callback through
 // [serviceGraphOptions.buildGateKey] without touching the DAG code.
 //
 // When Aspire eventually moves into an extension, this function — and only
--- a/cli/azd/internal/cmd/deploy.go
+++ b/cli/azd/internal/cmd/deploy.go
@@ -272,10 +272,12 @@ func (da *DeployAction) Run(ctx context.Context) (*actions.ActionResult, error) 
 
 // deployServicesGraph builds an execution graph of service deployments and runs them in
 // parallel via the execution graph scheduler. Services that share a non-empty
-// [serviceGraphOptions.buildGateKey] serialize on the first one to act as a
-// shared build gate; today that policy only fires for Aspire services
-// (DotNetContainerApp != nil), which share a single .NET AppHost build. Every
-// other service runs fully in parallel with no inter-service edges.
+// [serviceGraphOptions.buildGateKey] share a runtime mutex that serializes only
+// their image-preparation phase (dotnet publish); the Azure deployment portion
+// runs fully in parallel. Today that policy only fires for Aspire services
+// (DotNetContainerApp != nil), which share .NET project references whose obj/
+// directories collide under parallel `dotnet publish`. Every other service runs
+// fully in parallel with no inter-service edges.
 func (da *DeployAction) deployServicesGraph(
 	ctx context.Context,
 	stableServices []*project.ServiceConfig,
--- a/cli/azd/internal/cmd/deploy_graph_test.go
+++ b/cli/azd/internal/cmd/deploy_graph_test.go
@@ -14,12 +14,12 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-// TestDeployServicesGraph_AspireOrdering verifies that the Aspire build-gate
-// policy (supplied by [aspireBuildGateKey]) produces the expected "first wins,
-// rest wait" serialization: the first Aspire service has no extra deps, every
-// subsequent Aspire service depends on that first step, and non-Aspire
-// services run with no inter-service edges.
-func TestDeployServicesGraph_AspireOrdering(t *testing.T) {
+// TestDeployServicesGraph_AspireParallel verifies that the Aspire build-gate
+// policy does NOT add graph-level dependency edges between Aspire deploy steps.
+// Serialization of the dotnet publish phase is handled at runtime via a mutex
+// (see build_gate.go), so all Aspire services deploy in parallel at the graph
+// level while only their image-preparation phases are serialized.
+func TestDeployServicesGraph_AspireParallel(t *testing.T) {
 	services := []*project.ServiceConfig{
 		{Name: "api", DotNetContainerApp: &project.DotNetContainerAppOptions{
 			Manifest: &apphost.Manifest{},
@@ -33,7 +33,7 @@ func TestDeployServicesGraph_AspireOrdering(t *testing.T) {
 		}},
 	}
 
-	g := buildDeployGraph(t, services, aspireBuildGateKey)
+	g := buildDeployGraph(t, services)
 
 	require.Equal(t, 4, g.Len())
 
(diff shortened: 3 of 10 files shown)


wbreza (commented):
## Code Review — PR #8195

**fix: resolve parallel dotnet publish race in Aspire deploys**

### ✅ What Works Well

- **Clean two-tier design**: Primary approach (`--artifacts-path` isolation) enables full parallelism; fallback (mutex serialization) handles edge cases gracefully
- **Good fallback logging**: `log.Printf` warns when falling back to serial build — no silent failures
- **Explicit lock/unlock**: Mutex released before Azure deployment begins, preserving parallelism where it matters
- **Solid happy-path test coverage**: `Test_Cli_ArtifactsPathContext` tests `PublishContainer` and `BuildContainerLocal` with and without context
- **Clean context API**: `ContextWithArtifactsPath`/`ArtifactsPathFromContext` follow standard Go context patterns

---

### 🔴 P1 — Significant

**1. Windows MAX_PATH: Comment claims truncation but doesn't implement it**
📍 `service_target_dotnet_containerapp.go:172-173`

The comment says *"keep the prefix short to avoid MAX_PATH issues on Windows"* but `safeName` is the **full service name** with no length truncation. With a long Aspire service name (e.g., `myapp-apps-migrationconsole-worker`), the temp path could exceed Windows' 260-char MAX_PATH, **re-introducing the very failure mode the PR fixes**.

**Suggestion:** Truncate `safeName` to ~20 chars or use a short hash.

---

### 🟡 P2 — Improvements

**2. No test for mutex fallback path**
📍 `build_gate_test.go` / `service_target_dotnet_containerapp.go`

No test simulates `os.MkdirTemp` failure to verify the mutex fallback works. This is a critical safety net with zero test coverage.

**3. Tests verify topology but don't prove parallel execution**
📍 `service_graph_test.go` / `deploy_graph_test.go`

Tests assert no deploy→deploy graph edges but never verify services actually execute concurrently. A test with sync primitives would prove the intended behavior.

**4. `sanitizeTempDirName` missing edge cases**
📍 `build_gate_test.go:27-42`

Missing: empty string, all-unsafe-characters, very long names (250+ chars with no truncation).

**5. `BuildGateFromContext` coupling**
📍 `service_target_dotnet_containerapp.go:167`

`BuildGateFromContext(ctx) != nil` is used as a proxy for "needs artifacts isolation," conflating the mutex mechanism with the detection of Aspire-gated services. A future non-Aspire .NET multi-service deploy needing isolation wouldn't trigger this path. Consider a more semantic signal.

**6. Hidden cross-package context contract**
📍 `service_graph.go` → `build_gate.go` → `dotnet.go`

Three packages coupled via implicit context keys. Consider adding a doc comment in `build_gate.go` documenting who sets these values, when, and what happens when absent.

---

### 🔵 P3 — Nits

- **Naming**: `BuildGateFromContext` doesn't convey "fallback" — consider `FallbackBuildMutexFromContext`
- **Collision potential**: `sanitizeTempDirName` maps different names to the same output (harmless due to random suffix but could confuse logs)
- **API ambiguity**: `ArtifactsPathFromContext` returns `""` for both "not set" and "empty" — a `(string, bool)` return would be more idiomatic

---

### 📊 Summary

| Priority | Count | Key Concern |
|----------|-------|-------------|
| **P1** | 1 | Windows MAX_PATH — truncation claimed but not implemented |
| **P2** | 5 | Missing fallback tests, no parallelism proof, edge cases, coupling |
| **P3** | 3 | Naming clarity, collision potential, API ambiguity |

**Overall:** Well-designed PR with a clean two-tier approach. The P1 (MAX_PATH) is a real risk on Windows. P2s are quality improvements suitable for a follow-up.

JeffreyCA (approved):
Looks good, since it touches Aspire I'd prefer another look from @vhvb1989

vhvb1989 (approved):

skoging:
Opened #9660 for the equivalent race in standard code-based .NET Function App publish paths. AZD_DEPLOY_CONCURRENCY=1 avoids it, and the behavior is consistent with #7776's graph-driven concurrency, though a later regression remains unconfirmed because the failure is timing-sensitive. #8195's isolated-artifacts plus SDK-gated mutex pattern appears applicable, but it currently only covers the Aspire Container App path.

PR #8198, opened 2026-05-15
Azure AI Agents: add next-step guidance and doctor diagnostics

### LINKED ISSUE #7975: Guide developers to the right next step after every `azd ai agent` command and add `azd ai agent doctor`

antriksh30:
## Summary
- add context-aware `Next:` guidance for the Azure AI Agents extension across init, run, invoke, show, deploy-hook, and doctor flows
- add `azd ai agent doctor` with local and remote checks for project setup, environment variables, authentication, Foundry reachability/RBAC, hosted agent status, identity roles, model deployments, connections, and toolboxes
- improve post-deploy guidance with README-first sample payload hints, omit active-agent `show` next steps, and stream text-mode doctor checks as they complete

### DIFF
--- a/cli/azd/extensions/azure.ai.agents/internal/cmd/doctor.go
+++ b/cli/azd/extensions/azure.ai.agents/internal/cmd/doctor.go
@@ -218,25 +218,16 @@ func resolveDoctorTrailing(ctx context.Context, azdClient *azdext.AzdClient) []n
 	}
 
 	if anyServiceDeployed(state.Services) {
-		// Capture the total agent-service count BEFORE filtering. The
-		// resolver's `len(state.Services) == 1` heuristic ordinarily
-		// keys "should I emit no-arg show/invoke commands?" off the
-		// total count of agent services in azure.yaml. Once we filter
-		// to deployed-only, that heuristic breaks: a 2-service project
-		// with 1 deployed would emit `azd ai agent show` (no name),
-		// but runtime `resolveAgentService` still sees both services
-		// in azure.yaml and would either prompt or error. Forcing
-		// qualified suggestions whenever azure.yaml has multiple
-		// services preserves copy-paste correctness in the partial-
-		// deploy case and is a no-op when all services are deployed
-		// (the resolver naturally qualifies len > 1 anyway).
-		totalServices := len(state.Services)
-		filtered := filterDeployedServices(state)
+		// ResolveAfterDeploy always emits service-qualified
+		// `azd ai agent show <name>` / `invoke <name> ...` commands
+		// post-B9 (issue #7975), so it's safe to pass a filtered
+		// (deployed-only) State directly — the suggestions remain
+		// copy-paste correct even when azure.yaml has additional
+		// undeployed services that are absent from the filtered set.
 		return nextstep.ResolveAfterDeploy(
-			filtered,
+			filterDeployedServices(state),
 			doctorCachedPayload(ctx, azdClient),
 			doctorReadmeExists(ctx, azdClient),
-			nextstep.AfterDeployOpts{ForceQualified: totalServices > 1},
 		)
 	}
 
--- a/cli/azd/extensions/azure.ai.agents/internal/cmd/doctor.go
+++ b/cli/azd/extensions/azure.ai.agents/internal/cmd/doctor.go
@@ -19,20 +19,23 @@ import (
 
 // doctorFlags are the Cobra-bound flags for `azd ai agent doctor`.
 //
-// localOnly is exposed today as a no-op: every shipped check is local
-// (Phase 4 covers checks 1–6). The Cobra surface is locked early so the
-// Phase 5 follow-up that adds remote checks does not need to introduce
-// the flag in the same commit as the new check implementations.
+// localOnly skips remote (network-dependent) checks. The runner gates
+// remote checks via the Check.Remote field (see runner.go); doctor
+// remains responsive when network is unreachable, behind a proxy, or
+// the user just wants a fast local triage. Today the remote-checks
+// factory returns an empty slice, so the flag has no observable
+// effect — but the wire is fully exercised so the remote checks land
+// transparently.
 //
 // output selects the rendering path: "text" (default, human-readable
 // with a trailing Next: block on success) or "json" (structured envelope
 // for scripted consumers).
 //
-// unredacted is reserved for Phase 5 — once remote checks surface
-// principal IDs, scope ARNs, and UPNs, this flag will toggle the
-// redaction layer. It is bound today and threaded into doctor.Options
-// so that callers (and tests) can already exercise the wire without
-// the future Phase 5 fix-up touching the Cobra surface.
+// unredacted toggles the redaction of principal IDs, scope ARNs, and
+// UPNs in the report. The flag is surfaced today and threaded into
+// doctor.Options so remote checks can read `opts.Unredacted` from
+// their CheckFunc signature; the redaction layer itself lands with
+// the first check that produces sensitive identifiers.
 type doctorFlags struct {
 	localOnly  bool
 	output     string
@@ -123,8 +126,8 @@ Exit codes:
 
 	cmd.Flags().BoolVar(
 		&flags.localOnly, "local-only", false,
-		"Run only local checks (no network calls). "+
-			"All checks are local today; this flag is reserved for an upcoming remote-checks pass.",
+		"Skip remote (network-dependent) checks. "+
+			"Useful when offline, behind a proxy, or for a fast local triage.",
 	)
 	cmd.Flags().StringVarP(
 		&flags.output, "output", "o", "text",
@@ -133,7 +136,7 @@ Exit codes:
 	cmd.Flags().BoolVar(
 		&flags.unredacted, "unredacted", false,
 		"Show raw principal IDs, scope ARNs, and UPNs in the report. "+
-			"Reserved for the upcoming remote-checks pass (no-op today).",
+			"Has no effect today; takes effect when remote checks are added.",
 	)
 
 	return cmd
@@ -171,7 +174,14 @@ func runDoctor(
 	opts doctor.Options,
 	azdClient *azdext.AzdClient,
 ) (doctor.Report, []nextstep.Suggestion) {
-	runner := doctor.Runner{Checks: doctor.NewLocalChecks(deps)}
+	// Local checks run first so their Results are available to
+	// remote checks' skip-cascade guards (each remote check inspects
+	// `prior []Result` via `priorBlocked` to decide whether to skip
+	// when an upstream local precondition failed). The slice order
+	// here is the source of truth for that contract — do not
+	// reorder.
+	checks := append(doctor.NewLocalChecks(deps), doctor.NewRemoteChecks(deps)...)
+	runner := doctor.Runner{Checks: checks}
 	report := runner.Run(ctx, opts)
 
 	// Trailing Next: block is only meaningful when checks all pass
--- a/cli/azd/extensions/azure.ai.agents/internal/cmd/doctor.go
+++ b/cli/azd/extensions/azure.ai.agents/internal/cmd/doctor.go
@@ -103,8 +103,15 @@ Exit codes:
 			// and any non-nil return to exit 1, which collapses our
 			// three-state contract into a two-state one. We call
 			// os.Exit directly to preserve the 0/1/2 distinction.
-			// Defers above run via the explicit Close + flushed
-			// stdout writer; nothing else needs cleanup before exit.
+			//
+			// os.Exit does NOT run deferred functions. The deferred
+			// logCleanup and azdClient.Close above will not execute on
+			// the non-zero path. This is acceptable today because the
+			// process exits immediately and the OS reclaims the gRPC
+			// socket and (in --debug mode) the log fd; neither defer
+			// has on-disk state to flush. Do NOT add cleanup-critical
+			// defers to this RunE — call them explicitly before
+			// os.Exit instead.
 			code := doctor.ExitCode(report)
 			if code == 0 {
 				return nil
@@ -211,11 +218,25 @@ func resolveDoctorTrailing(ctx context.Context, azdClient *azdext.AzdClient) []n
 	}
 
 	if anyServiceDeployed(state.Services) {
+		// Capture the total agent-service count BEFORE filtering. The
+		// resolver's `len(state.Services) == 1` heuristic ordinarily
+		// keys "should I emit no-arg show/invoke commands?" off the
+		// total count of agent services in azure.yaml. Once we filter
+		// to deployed-only, that heuristic breaks: a 2-service project
+		// with 1 deployed would emit `azd ai agent show` (no name),
+		// but runtime `resolveAgentService` still sees both services
+		// in azure.yaml and would either prompt or error. Forcing
+		// qualified suggestions whenever azure.yaml has multiple
+		// services preserves copy-paste correctness in the partial-
+		// deploy case and is a no-op when all services are deployed
+		// (the resolver naturally qualifies len > 1 anyway).
+		totalServices := len(state.Services)
 		filtered := filterDeployedServices(state)
 		return nextstep.ResolveAfterDeploy(
 			filtered,
 			doctorCachedPayload(ctx, azdClient),
 			doctorReadmeExists(ctx, azdClient),
+			nextstep.AfterDeployOpts{ForceQualified: totalServices > 1},
 		)
 	}
 
@@ -260,7 +281,28 @@ func filterDeployedServices(state *nextstep.State) *nextstep.State {
 // for the deployed agent (`azd ai agent invoke <agent>`); the local
 // cache (suffix "local") is from `azd ai agent invoke --local` and is
 // not appropriate here.
+//
+// Key resolution: the on-disk cache is keyed by the deployed Foundry
+// agent name (see invoke.go:694-758 — invoke rewrites `name` to
+// `info.AgentName` BEFORE caching). That can differ from the azure.yaml
+// service name when deploy appends a suffix (documented in
+// show.go:40-46). The closure first tries the deployed name via the
+// `AGENT_<SERVICE>_NAME` env var, then falls back to the service name
+// when the env value is absent (e.g., never-deployed service, or older
+// deploys that did not populate the var). The fallback also covers the
+// non-divergent case where the two names are identical.
 func doctorCachedPayload(ctx context.Context, azdClient *azdext.AzdClient) func(string) string {
+	// Resolve the active env name once for the closure's lifetime.
+	// A nil/error response leaves envName empty, which short-circuits
+	// the deployed-name lookup path inside the closure.
+	var envName string
+	if azdClient != nil {
+		if envResp, err := azdClient.Environment().GetCurrent(ctx, &azdext.EmptyRequest{}); err == nil &&
+			envResp != nil && envResp.Environment != nil {
+			envName = envResp.Environment.Name
+		}
+	}
+
 	return func(serviceName string) string {
 		if azdClient == nil || serviceName == "" {
 			return ""
@@ -269,7 +311,27 @@ func doctorCachedPayload(ctx context.Context, azdClient *azdext.AzdClient) func(
 		if err != nil {
 			return ""
 		}
-		spec, err := nextstep.ReadCachedOpenAPISpec(filepath.Dir(configPath), serviceName, "remote")
+		configDir := filepath.Dir(configPath)
+
+		// Try the deployed agent name first.
+		if envName != "" {
+			nameKey := fmt.Sprintf("AGENT_%s_NAME", toServiceKey(serviceName))
+			if v, err := azdClient.Environment().GetValue(ctx, &azdext.GetEnvRequest{
+				EnvName: envName,
+				Key:     nameKey,
+			}); err == nil && v != nil && v.Value != "" && v.Value != serviceName {
+				if spec, err := nextstep.ReadCachedOpenAPISpec(configDir, v.Value, "remote"); err == nil {
+					if payload := nextstep.ExtractInvokeExample(spec); payload != "" {
+						return payload
+					}
+				}
+			}
+		}
+
+		// Fall back to service-name keyed cache for the non-divergent
+		// case (and for projects whose AGENT_<SERVICE>_NAME var is
+		// absent for any reason).
+		spec, err := nextstep.ReadCachedOpenAPISpec(configDir, serviceName, "remote")
 		if err != nil {
 			return ""
 		}
--- a/cli/azd/extensions/azure.ai.agents/internal/cmd/doctor.go
+++ b/cli/azd/extensions/azure.ai.agents/internal/cmd/doctor.go
@@ -88,6 +88,7 @@ Exit codes:
 				AzdClient:        azdClient,
 				AzdClientErr:     clientErr,
 				ExtensionVersion: version.Version,
+				AgentAPIVersion:  DefaultAgentAPIVersion,
 			}
 
 			opts := doctor.Options{
--- a/cli/azd/extensions/azure.ai.agents/internal/cmd/doctor.go
+++ b/cli/azd/extensions/azure.ai.agents/internal/cmd/doctor.go
@@ -6,6 +6,7 @@ package cmd
 import (
 	"context"
 	"fmt"
+	"io"
 	"os"
 	"path/filepath"
 
@@ -22,10 +23,7 @@ import (
 // localOnly skips remote (network-dependent) checks. The runner gates
 // remote checks via the Check.Remote field (see runner.go); doctor
 // remains responsive when network is unreachable, behind a proxy, or
-// the user just wants a fast local triage. Today the remote-checks
-// factory returns an empty slice, so the flag has no observable
-// effect — but the wire is fully exercised so the remote checks land
-// transparently.
+// the user just wants a fast local triage.
 //
 // output selects the rendering path: "text" (default, human-readable
 // with a trailing Next: block on success) or "json" (structured envelope
@@ -33,9 +31,8 @@ import (
 //
 // unredacted toggles the redaction of principal IDs, scope ARNs, and
 // UPNs in the report. The flag is surfaced today and threaded into
-// doctor.Options so remote checks can read `opts.Unredacted` from
-// their CheckFunc signature; the redaction layer itself lands with
-// the first check that produces sensitive identifiers.
+// doctor.Options so checks can read `opts.Unredacted` from their
+// CheckFunc signature.
 type doctorFlags struct {
 	localOnly  bool
 	output     string
@@ -96,9 +93,19 @@ Exit codes:
 				Unredacted: flags.unredacted,
 			}
 
-			report, trailing := runDoctor(ctx, deps, opts, azdClient)
-			if err := renderDoctorReport(os.Stdout, flags.output, report, trailing); err != nil {
-				return err
+			var report doctor.Report
+			if flags.output == "text" {
+				var err error
+				report, err = runAndRenderDoctorText(ctx, deps, opts, azdClient, os.Stdout)
+				if err != nil {
+					return err
+				}
+			} else {
+				var trailing []nextstep.Suggestion
+				report, trailing = runDoctor(ctx, deps, opts, azdClient)
+				if err := renderDoctorReport(os.Stdout, flags.output, report, trailing); err != nil {
+					return err
+				}
 			}
 
 			// Exit codes are part of the doctor contract (see design
@@ -136,8 +143,7 @@ Exit codes:
 	)
 	cmd.Flags().BoolVar(
 		&flags.unredacted, "unredacted", false,
-		"Show raw principal IDs, scope ARNs, and UPNs in the report. "+
-			"Has no effect today; takes effect when remote checks are added.",
+		"Show raw principal IDs, scope ARNs, and UPNs in the report.",
 	)
 
 	return cmd
@@ -175,6 +181,51 @@ func runDoctor(
 	opts doctor.Options,
 	azdClient *azdext.AzdClient,
 ) (doctor.Report, []nextstep.Suggestion) {
+	report, trailing, _ := runDoctorWithObserver(ctx, deps, opts, azdClient, nil)
+	return report, trailing
+}
+
+// runAndRenderDoctorText streams the human-readable doctor output as
+// checks complete. JSON output intentionally does not use this path; it
+// remains buffered so scripted consumers receive one stable envelope.
+func runAndRenderDoctorText(
+	ctx context.Context,
+	deps doctor.Dependencies,
+	opts doctor.Options,
+	azdClient *azdext.AzdClient,
+	w io.Writer,
+) (doctor.Report, error) {
+	if err := printDoctorReportTextHeader(w); err != nil {
+		return doctor.Report{}, err
+	}
+
+	report, trailing, err := runDoctorWithObserver(
+		ctx,
+		deps,
+		opts,
+		azdClient,
+		func(result doctor.Result) error {
+			return writeCheckLines(w, result)
+		},
+	)
+	if err != nil {
+		return report, err
+	}
+
+	showNext := len(trailing) > 0 && writerIsTerminal(w)
+	if err := printDoctorReportTextFooter(w, report, trailing, showNext); err != nil {
+		return report, err
+	}
+	return report, nil
+}
+
+func runDoctorWithObserver(
+	ctx context.Context,
+	deps doctor.Dependencies,
+	opts doctor.Options,
+	azdClient *azdext.AzdClient,
+	observer doctor.ResultObserver,
+) (doctor.Report, []nextstep.Suggestion, error) {
 	// Local checks run first so their Results are available to
 	// remote checks' skip-cascade guards (each remote check inspects
 	// `prior []Result` via `priorBlocked` to decide whether to skip
@@ -183,7 +234,10 @@ func runDoctor(
 	// reorder.
 	checks := append(doctor.NewLocalChecks(deps), doctor.NewRemoteChecks(deps)...)
 	runner := doctor.Runner{Checks: checks}
-	report := runner.Run(ctx, opts)
+	report, err := runner.RunWithObserver(ctx, opts, observer)
+	if err != nil {
+		return report, nil, err
+	}
 
 	// Trailing Next: block is only meaningful when checks all pass
 	// (exit code 0). On Fail or all-skip, the user's next move is to
@@ -192,11 +246,11 @@ func runDoctor(
 	// `docs/design/azd-ai-agent-nextsteps.md`, "Doctor output shape":
 	// "When all checks pass, the trailing Next: block is ...".
 	if doctor.ExitCode(report) != 0 {
-		return report, nil
+		return report, nil, nil
 	}
 
 	trailing := resolveDoctorTrailing(ctx, azdClient)
-	return report, trailing
+	return report, trailing, nil
 }
 
 // resolveDoctorTrailing assembles state from the azd gRPC channel and
--- a/cli/azd/extensions/azure.ai.agents/internal/cmd/doctor.go
+++ b/cli/azd/extensions/azure.ai.agents/internal/cmd/doctor.go
@@ -1,0 +1,297 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package cmd
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"path/filepath"
+
+	"azureaiagent/internal/cmd/doctor"
+	"azureaiagent/internal/cmd/nextstep"
+	"azureaiagent/internal/version"
+
+	"github.com/azure/azure-dev/cli/azd/pkg/azdext"
+	"github.com/spf13/cobra"
+)
+
+// doctorFlags are the Cobra-bound flags for `azd ai agent doctor`.
+//
+// localOnly is exposed today as a no-op: every shipped check is local
+// (Phase 4 covers checks 1–6). The Cobra surface is locked early so the
+// Phase 5 follow-up that adds remote checks does not need to introduce
+// the flag in the same commit as the new check implementations.
+//
+// output selects the rendering path: "text" (default, human-readable
+// with a trailing Next: block on success) or "json" (structured envelope
+// for scripted consumers).
+//
+// unredacted is reserved for Phase 5 — once remote checks surface
+// principal IDs, scope ARNs, and UPNs, this flag will toggle the
+// redaction layer. It is bound today and threaded into doctor.Options
+// so that callers (and tests) can already exercise the wire without
+// the future Phase 5 fix-up touching the Cobra surface.
+type doctorFlags struct {
+	localOnly  bool
+	output     string
+	unredacted bool
+}
+
+func newDoctorCommand() *cobra.Command {
+	flags := &doctorFlags{}
+
+	cmd := &cobra.Command{
+		Use:   "doctor",
+		Short: "Diagnose problems with an azd ai agent project.",
+		Long: `Diagnose problems with an azd ai agent project.
+
+Runs a sequence of local checks against the current azd project,
+reporting on each one and (when all checks pass) suggesting the next
+command to run. Use this when you have lost terminal context or hit a
+confusing error and want a complete picture of the project's state.
+
+Exit codes:
+  0 — at least one check passed and no checks failed
+  1 — any check failed
+  2 — all checks were skipped (e.g. preconditions unmet)`,
+		Example: `  # Run the full check suite with human-readable output
+  azd ai agent doctor
+
+  # Emit a structured JSON envelope (for scripts / CI)
+  azd ai agent doctor --output json`,
+		Args: cobra.NoArgs,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			if err := validateDoctorFlags(flags); err != nil {
+				return err
+			}
+
+			ctx := azdext.WithAccessToken(cmd.Context())
+			logCleanup := setupDebugLogging(cmd.Flags())
+			defer logCleanup()
+
+			// NewAzdClient errors are not fatal — the gRPC check
+			// (`local.grpc-extension`) surfaces the failure verbatim
+			// to the user, and downstream checks Skip cleanly when
+			// the client is nil. We deliberately do NOT short-circuit
+			// the command here.
+			azdClient, clientErr := azdext.NewAzdClient()
+			if azdClient != nil {
+				defer azdClient.Close()
+			}
+
+			deps := doctor.Dependencies{
+				AzdClient:        azdClient,
+				AzdClientErr:     clientErr,
+				ExtensionVersion: version.Version,
+			}
+
+			opts := doctor.Options{
+				LocalOnly:  flags.localOnly,
+				Unredacted: flags.unredacted,
+			}
+
+			report, trailing := runDoctor(ctx, deps, opts, azdClient)
+			if err := renderDoctorReport(os.Stdout, flags.output, report, trailing); err != nil {
+				return err
+			}
+
+			// Exit codes are part of the doctor contract (see design
+			// `docs/design/azd-ai-agent-nextsteps.md`, "Exit codes &
+			// JSON output"). Cobra/azdext maps a nil return to exit 0
+			// and any non-nil return to exit 1, which collapses our
+			// three-state contract into a two-state one. We call
+			// os.Exit directly to preserve the 0/1/2 distinction.
+			// Defers above run via the explicit Close + flushed
+			// stdout writer; nothing else needs cleanup before exit.
+			code := doctor.ExitCode(report)
+			if code == 0 {
+				return nil
+			}
+			os.Exit(code)
+			return nil // unreachable
+		},
+	}
+
+	cmd.Flags().BoolVar(
+		&flags.localOnly, "local-only", false,
+		"Run only local checks (no network calls). "+
+			"All checks are local today; this flag is reserved for an upcoming remote-checks pass.",
+	)
+	cmd.Flags().StringVarP(
+		&flags.output, "output", "o", "text",
+		"Output format (text or json).",
+	)
+	cmd.Flags().BoolVar(
+		&flags.unredacted, "unredacted", false,
+		"Show raw principal IDs, scope ARNs, and UPNs in the report. "+
+			"Reserved for the upcoming remote-checks pass (no-op today).",
+	)
+
+	return cmd
+}
+
+// validateDoctorFlags enforces the closed set of values for --output. We
+// validate before any work so an obvious typo (`--output yaml`) does not
+// run the entire check suite only to print nothing useful.
+func validateDoctorFlags(flags *doctorFlags) error {
+	switch flags.output {
+	case "text", "json":
+		return nil
+	default:
+		return fmt.Errorf("invalid --output value %q (must be 'text' or 'json')", flags.output)
+	}
+}
+
+// runDoctor is the testable core of the doctor command. It constructs a
+// Runner from the configured checks, executes it, and (when the report
+// is clean) resolves a trailing Next: block via the nextstep resolver.
+//
+// The trailing block is computed unconditionally but only rendered by
+// the text formatter — the JSON envelope deliberately excludes it (see
+// design spec, "Exit codes & JSON output"). Computing it here keeps the
+// expensive bit (gRPC round-trip in AssembleStateFromSource) out of the
+// formatter and lets tests assert the resolver branch by inspection.
+//
+// azdClient may be nil when NewAzdClient failed at startup; in that
+// case the trailing block is skipped (resolver has no state to work
+// with). The function never returns an error: every failure mode is
+// captured in the Report or in a skipped trailing block.
+func runDoctor(
+	ctx context.Context,
+	deps doctor.Dependencies,
+	opts doctor.Options,
+	azdClient *azdext.AzdClient,
+) (doctor.Report, []nextstep.Suggestion) {
+	runner := doctor.Runner{Checks: doctor.NewLocalChecks(deps)}
+	report := runner.Run(ctx, opts)
+
+	// Trailing Next: block is only meaningful when checks all pass
+	// (exit code 0). On Fail or all-skip, the user's next move is to
+	// fix the surfaced problem — burying that under "Next: azd deploy"
+	// would be noise. Locked by the design spec at
+	// `docs/design/azd-ai-agent-nextsteps.md`, "Doctor output shape":
+	// "When all checks pass, the trailing Next: block is ...".
+	if doctor.ExitCode(report) != 0 {
+		return report, nil
+	}
+
+	trailing := resolveDoctorTrailing(ctx, azdClient)
+	return report, trailing
+}
+
+// resolveDoctorTrailing assembles state from the azd gRPC channel and
+// asks the nextstep resolver for the doctor's trailing block.
+// Returns nil on any error — the trailing block is a courtesy, not a
+// load-bearing surface, and the body of the doctor report already
+// tells the user what to do.
+//
+// Branch selection:
+//   - Any service in azure.yaml has IsDeployed == true →
+//     ResolveAfterDeploy (filtered to deployed services). The resolver
+//     emits show + invoke for each deployed agent.
+//   - No service deployed → ResolveAfterInit. Same block the user saw
+//     at the end of `azd ai agent init`, which guides them toward
+//     `azd provision` / `azd ai agent run` / `azd deploy`.
+func resolveDoctorTrailing(ctx context.Context, azdClient *azdext.AzdClient) []nextstep.Suggestion {
+	if azdClient == nil {
+		return nil
+	}
+
+	state, _ := nextstep.AssembleStateFromSource(ctx, nextstep.NewSource(azdClient))
+	if len(state.Services) == 0 {
+		// Healthy project but no agent services in azure.yaml — the
+		// init resolver still produces a useful "run azd ai agent
+		// init" hint via its empty-services branch, but for doctor
+		// the body of the report already covered that via the
+		// `local.agent-service-detected` check. Emitting the same
+		// hint twice is noise.
+		return nil
+	}
+
+	if anyServiceDeployed(state.Services) {
+		filtered := filterDeployedServices(state)
+		return nextstep.ResolveAfterDeploy(
+			filtered,
+			doctorCachedPayload(ctx, azdClient),
+			doctorReadmeExists(ctx, azdClient),
+		)
+	}
+
+	return nextstep.ResolveAfterInit(state)
+}
+
+func anyServiceDeployed(services []nextstep.ServiceState) bool {
+	for _, s := range services {
+		if s.IsDeployed {
+			return true
+		}
+	}
+	return false
+}
+
+// filterDeployedServices returns a shallow clone of state whose Services
+// list contains only the entries with IsDeployed == true. The clone is
+// necessary because ResolveAfterDeploy emits one show + one invoke
+// per Service it sees; passing an unfiltered state would produce
+// `azd ai agent invoke <undeployed-service>` lines, which 404.
+func filterDeployedServices(state *nextstep.State) *nextstep.State {
+	if state == nil {
+		return nil
+	}
+	clone := *state
+	clone.Services = make([]nextstep.ServiceState, 0, len(state.Services))
+	for _, s := range state.Services {
+		if s.IsDeployed {
+			clone.Services = append(clone.Services, s)
+		}
+	}
+	return &clone
+}
+
+// doctorCachedPayload returns a cachedPayload closure for
+// ResolveAfterDeploy. It looks up the cached remote OpenAPI spec (the
+// one populated by prior `azd ai agent invoke` runs) and extracts a
+// sample payload via ExtractInvokeExample. Returns "" on any failure
+// so the resolver falls back to its protocol-generic literal.
+//
+// Suffix is "remote" because doctor's trailing block emits commands
+// for the deployed agent (`azd ai agent invoke <agent>`); the local
+// cache (suffix "local") is from `azd ai agent invoke --local` and is
+// not appropriate here.
+func doctorCachedPayload(ctx context.Context, azdClient *azdext.AzdClient) func(string) string {
+	return func(serviceName string) string {
+		if azdClient == nil || serviceName == "" {
+			return ""
+		}
+		configPath, err := resolveConfigPath(ctx, azdClient)
+		if err != nil {
+			return ""
+		}
+		spec, err := nextstep.ReadCachedOpenAPISpec(filepath.Dir(configPath), serviceName, "remote")
+		if err != nil {
+			return ""
+		}
+		return nextstep.ExtractInvokeExample(spec)
+	}
+}
+
+// doctorReadmeExists returns a readmeExists closure for
+// ResolveAfterDeploy. The closure resolves the project root once
+// (cached across calls) and reports whether
+// <projectRoot>/<relativePath>/README.md exists.
+//
+// Only the canonical "README.md" casing is checked, matching the
+// rendered "see <relPath>/README.md" line; accepting other casings
+// would yield a broken pointer on case-sensitive filesystems.
+func doctorReadmeExists(ctx context.Context, azdClient *azdext.AzdClient) func(string) bool {
+	projectRoot := resolveProjectPath(ctx, azdClient)
+	return func(relativePath string) bool {
+		if projectRoot == "" || relativePath == "" {
+			return false
+		}
+		_, err := os.Stat(filepath.Join(projectRoot, relativePath, "README.md"))
+		return err == nil
+	}
+}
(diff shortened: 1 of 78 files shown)


cli/azd/extensions/azure.ai.agents/internal/cmd/invoke.go
```diff
@@ -452,8 +531,14 @@ func (a *InvokeAction) resolveRemoteContext(ctx context.Context) (*remoteContext
...
+	// azure.yaml service name separately from the deployed Foundry name
+	// so post-success next-step suggestions emit the service name; show
+	// keys on s.Name in azure.yaml and would 404 on the deployed Foundry
+	// name in the divergent case.
 	if info, err := resolveAgentServiceFromProject(ctx, azdClient, rc.name, a.noPrompt); err == nil {
-		if rc.name == "" && info.AgentName != "" {
+		rc.serviceName = info.ServiceName
+		if info.AgentName != "" {
 			rc.name = info.AgentName
 		}
```
jongio, line 637: This drops the `rc.name == ""` guard that preserved a user-supplied `--name` flag. Before this PR, `azd ai agent invoke --name foo` kept `foo` as the agent name even when `info.AgentName` was set. Now `info.AgentName` always wins, silently overriding the explicit flag.

If that's intentional (e.g. info.AgentName is the canonical source), a `log.Printf` when the override happens would help users understand why their `--name` was ignored. If it's unintentional, restoring the guard is a one-liner:

```suggestion
		if rc.name == "" && info.AgentName != "" {
```

cli/azd/extensions/azure.ai.agents/internal/cmd/doctor.go
```diff
@@ -0,0 +1,271 @@
...
+			}
+
+			// Use os.Exit to preserve doctor's 0/1/2 exit-code contract;
+			// Cobra/azdext would otherwise collapse all errors to 1.
+			// os.Exit skips defers, so do not add cleanup-critical defers here.
+			code := doctor.ExitCode(report)
+			if code == 0 {
+				return nil
+			}
+			os.Exit(code)
```
jongio, line 88: `os.Exit(code)` skips the deferred `logCleanup()` and `azdClient.Close()` above. For non-zero exits the log file won't be cleaned up and the client connection leaks.

One fix - return the exit code up to the caller and let Cobra handle the exit, or call the cleanup explicitly before `os.Exit`:

```go
if code != 0 {
    logCleanup()
    if azdClient != nil {
        azdClient.Close()
    }
    os.Exit(code)
}
```

cli/azd/extensions/azure.ai.agents/internal/cmd/doctor/checks_auth.go
```diff
@@ -0,0 +1,207 @@
...
+				}
+			}
+			return Result{
+				Status: StatusPass,
+				Message: composeAuthMessage(res.upn,
+					"token valid for "+formatTokenWindow(res.validFor)),
+				Details: map[string]any{
+					"validForMinutes": minutes,
+					"upn":             res.upn,
+				},
```
jongio: The `Options` parameter is unused (`_` on line 104), so UPN ends up in `Details` and `Message` regardless of `--unredacted`. If a user runs `azd ai agent doctor` without `--unredacted`, they'd still see their UPN in the output.

Either wire `opts.Unredacted` through and conditionally redact, or drop the UPN from the default output and only include it when `--unredacted` is set.

jongio (commented):
Solid foundation for the doctor + nextstep subsystems - clean module boundaries, interface-driven testability, and well-structured check pipeline. A few things to sort out before this leaves draft:

**invoke.go:541-543** is the one I'd prioritize - it silently changes CLI semantics for `--name`. The rest are lower-stakes but worth a look.

cli/azd/extensions/azure.ai.agents/internal/cmd/doctor.go
```diff
@@ -0,0 +1,271 @@
...
+
+// resolveDoctorTrailing returns the doctor's trailing Next block, or nil on
+// error. It chooses deployed-agent suggestions when any service is deployed;
+// otherwise it reuses the post-init guidance.
+func resolveDoctorTrailing(ctx context.Context, azdClient *azdext.AzdClient) []nextstep.Suggestion {
+	if azdClient == nil {
+		return nil
+	}
+
+	state, _ := nextstep.AssembleStateFromSource(ctx, nextstep.NewSource(azdClient))
```
jongio, line 176: The error from `AssembleStateFromSource` is discarded here. The state is always non-nil (so no crash risk), but if assembly partially fails the trailing Next block will be based on incomplete data without any indication to the user.

Consider logging the errors at debug level so `--debug` runs surface what went wrong:

```go
state, errs := nextstep.AssembleStateFromSource(ctx, nextstep.NewSource(azdClient))
for _, e := range errs {
    log.Printf("doctor trailing state: %v", e)
}
```

cli/azd/extensions/azure.ai.agents/internal/cmd/nextstep/error_codes_test.go
```diff
@@ -0,0 +1,114 @@
...
+		"SandboxNotFound":                string(SessionSandboxNotFound),
+		"QuotaExceeded":                  string(SessionQuotaExceeded),
+		"RegionalQuotaExceeded":          string(SessionRegionalQuotaExceeded),
+		"AgentVersionNotReady":           string(SessionAgentVersionNotReady),
+		"AgentVersionProvisioningFailed": string(SessionAgentVersionProvisioningFailed),
+		"creating":                       string(AgentVersionCreating),
+		"active":                         string(AgentVersionActive),
+		"failed":                         string(AgentVersionFailed),
+		"deleting":                       string(AgentVersionDeleting),
+		"deleted":                        string(AgentVersionDeleted),
```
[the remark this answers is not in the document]
antriksh30: Good catch — pinned in 631272951. `TestErrorCodeWireValues` now includes `"idle": string(AgentVersionIdle)` so silent drift on that literal can no longer regress the consumers at `show.go:207` and `nextstep/resolver.go:248`.

cli/azd/extensions/azure.ai.agents/internal/cmd/doctor.go
```diff
@@ -0,0 +1,266 @@
...
+	if len(state.Services) == 0 {
+		// Avoid repeating the missing-service guidance already reported by
+		// `local.agent-service-detected`.
+		return nil
+	}
```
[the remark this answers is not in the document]
antriksh30: `assembleState` (`nextstep/state.go:163`) always initializes `state := &State{}` and returns the pointer on every path — there is no caller branch that returns `nil, nil`. The other nil checks in this file are forward-looking guards documented as such; this one would be unreachable today.

cli/azd/extensions/azure.ai.agents/internal/cmd/doctor.go
```diff
@@ -0,0 +1,266 @@
...
+
+			// Use os.Exit to preserve doctor's 0/1/2 exit-code contract;
+			// Cobra/azdext would otherwise collapse all errors to 1.
+			// os.Exit skips defers, so do not add cleanup-critical defers here.
+			code := doctor.ExitCode(report)
+			if code == 0 {
+				return nil
+			}
+			os.Exit(code)
+			return nil // unreachable
```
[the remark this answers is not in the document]
antriksh30: This is intentional and documented in the comment at lines 77-79: the doctor command uses `os.Exit` to honor the 0/1/2 exit contract for tooling. `logCleanup` writes to an unbuffered append-mode file (`debug.go:38-48`) so no log data is lost on exit; the gRPC client is reclaimed by the OS on process exit and `Close()` is not correctness-critical. Wrapping the body in a separate function to allow defers to run would add a second layer of indirection for no observable behavior change.

cli/azd/extensions/azure.ai.agents/internal/cmd/invoke.go
```diff
@@ -452,8 +543,14 @@ func (a *InvokeAction) resolveRemoteContext(ctx context.Context) (*remoteContext
...
```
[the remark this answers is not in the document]
antriksh30: The override is intentional. `--name` selects the **azure.yaml service**, but the agent API call needs the **deployed Foundry name**, which the deploy step writes to `AZD_AI_AGENT_*` and `info.AgentName` reads back. We only overwrite when `info.AgentName != ""`, so when the service has not been deployed yet the user-supplied name is preserved. `rc.serviceName` keeps the azure.yaml name for the follow-up next-step suggestions.

cli/azd/extensions/azure.ai.agents/internal/cmd/pending_provision.go
```diff
@@ -0,0 +1,215 @@
...
+	priorRaw, err := readPendingProvisionEnv(ctx, azdClient, envName)
+	if err != nil {
+		return nil, err
+	}
+	curr := parsePendingProvisionReasons(priorRaw)
+	next := parsePendingProvisionReasons(formatPendingProvisionReasons(mutate(curr)))
+	formatted := strings.Join(next, ",")
+	if formatted == priorRaw {
+		return next, nil
+	}
```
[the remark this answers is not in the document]
antriksh30: This function is the sole writer of `AI_AGENT_PENDING_PROVISION`. A non-canonical existing value triggers exactly one fix-up write, then every subsequent comparison matches and no further writes happen. The convergence happens in one normalization step, not a loop.

cli/azd/extensions/azure.ai.agents/internal/cmd/nextstep/infra_outputs_test.go
```diff
@@ -0,0 +1,287 @@
...
+			want: []string{"AZURE_AI_PROJECT_ID"},
+		},
+		{
+			name: "indented output (inside a conditional block) accepted",
+			in: `if (condition) {
+  output INDENTED_OUTPUT string = 'value'
+}
+`,
+			want: []string{"INDENTED_OUTPUT"},
+		},
```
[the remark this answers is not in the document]
antriksh30: The test exercises the **regex parser** behavior on indented `output` text, not a claim about valid Bicep syntax. The parser is documented as best-effort and the test is verifying that an `output` line nested in a conditional block is still detected. Happy to rename to `*_indentedOutput` if that reads better, but the assertion itself is correct.

cli/azd/extensions/azure.ai.agents/internal/cmd/doctor/checks_toolboxes.go
```diff
@@ -0,0 +1,198 @@
...
+		key := envkey.ToolboxMCPEndpoint(t.Name)
+		if _, ok := seen[key]; ok {
+			continue
+		}
+		seen[key] = struct{}{}
+		keys = append(keys, key)
+	}
+	slices.Sort(keys)
+	return keys
+}
```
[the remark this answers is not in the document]
antriksh30: `dedupToolboxKeys` is referenced by `checks_toolboxes_test.go:277` and `:293` to pin the envkey/dedup contract independently of the production call site (which uses inline dedup for clarity). Removing the helper would weaken the regression pin — same-package tests count as callers for Go unused analysis.

cli/azd/extensions/azure.ai.agents/internal/cmd/doctor_format.go
```diff
@@ -0,0 +1,149 @@
...
+	}
+
+	for _, c := range report.Checks {
+		if err := writeCheckLines(w, c); err != nil {
+			return err
+		}
+	}
+
+	return printDoctorReportTextFooter(w, report, trailing, showNext)
+}
```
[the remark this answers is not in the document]
antriksh30: Both `printDoctorReportText` and `runAndRenderDoctorText` delegate to the same three primitives (`printDoctorReportTextHeader`, `writeCheckLines`, `printDoctorReportTextFooter`). Any rendering change goes through those functions regardless of entry point, so the two callers cannot drift.

cli/azd/extensions/azure.ai.agents/internal/cmd/invoke.go
```diff
@@ -392,10 +470,15 @@ func (a *InvokeAction) responsesLocal(ctx context.Context) error {
...
+		a.emitInvokeSuccessNextStep(nextstep.InvokeLocal, "")
 		return nil
 	}
 
-	return printAgentResponse(result, "local")
+	if err := printAgentResponse(result, "local"); err != nil {
+		return err
+	}
+	a.emitInvokeSuccessNextStep(nextstep.InvokeLocal, "")
+	return nil
```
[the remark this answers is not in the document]
antriksh30: Verified: the unmarshal-failure branch at L458-462 ends in `return nil`, so the second emission at L468 is only reachable on unmarshal **success**. The emit at L460 (failure) and L468 (success) are mutually exclusive — no double-emit possible.

cli/azd/extensions/azure.ai.agents/internal/cmd/nextstep/error_codes.go
```diff
@@ -0,0 +1,131 @@
...
+const (
+	AgentVersionCreating AgentVersionStatus = "creating"
+	AgentVersionActive   AgentVersionStatus = "active"
+	// AgentVersionIdle is a defensive synonym for "active" — the issue
+	// #7975 spec lists `idle`, although the verified enum only emits `active`.
+	AgentVersionIdle     AgentVersionStatus = "idle"
+	AgentVersionFailed   AgentVersionStatus = "failed"
+	AgentVersionDeleting AgentVersionStatus = "deleting"
+	AgentVersionDeleted  AgentVersionStatus = "deleted"
+)
```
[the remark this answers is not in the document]
antriksh30: Good catch — pinned in 631272951. `TestErrorCodeWireValues` now includes `"idle": string(AgentVersionIdle)` so silent drift on that literal can no longer regress the consumers at `show.go:207` and `nextstep/resolver.go:248`.

cli/azd/extensions/azure.ai.agents/internal/cmd/invoke.go
antriksh30: Same as the Copilot thread above — the override is intentional. `--name` selects the azure.yaml service; the API call needs the deployed Foundry name from `AZD_AI_AGENT_*`. The `if info.AgentName != ""` guard preserves the user-supplied value when the service has not been deployed yet.

cli/azd/extensions/azure.ai.agents/internal/cmd/doctor.go
antriksh30: Same as the Copilot thread above — the `os.Exit` is intentional to preserve the 0/1/2 exit contract and is documented at lines 77-79. `logCleanup` is unbuffered file I/O and the gRPC client is reclaimed by the OS, so neither cleanup is correctness-critical.

cli/azd/extensions/azure.ai.agents/internal/cmd/doctor/checks_auth.go
antriksh30: Good catch — fixed in 16153f3ba. The `remote.auth` check now threads `opts Options` into its function, redacts the UPN by default via the same `redactedPlaceholder` constant used in `checks_rbac.go`/`checks_agent_identity_roles.go`, and omits the `upn` key from `Details` unless `--unredacted` is set. Two helpers (`redactUPN`, `authDetails`) centralize the logic; existing tests that asserted the raw value now pass `Options{Unredacted: true}` and new table tests cover the default-redacted and `--unredacted` contracts on PASS/WARN/expired-FAIL plus the empty-UPN drop.

cli/azd/extensions/azure.ai.agents/internal/cmd/doctor.go
antriksh30: The trailing Next block is opportunistic and runs only when the report itself is clean (`ExitCode == 0`). `AssembleStateFromSource` is documented as best-effort — partial errors degrade gracefully and the printable hints/codes still come through. All actionable errors are already surfaced by the individual checks above. If you would prefer we log the discarded errs at debug level, happy to add that.

cli/azd/extensions/azure.ai.agents/internal/cmd/invoke.go
jongio: Makes sense - `rc.name` is the Foundry name for the API, `serviceName` preserves the azure.yaml key. The comment at L536-539 explains the divergent case clearly. Resolved on my end.

cli/azd/extensions/azure.ai.agents/internal/cmd/doctor.go
jongio: Fair enough - the 0/1/2 exit contract for CI makes the explicit `os.Exit` the right call here. The defers are informational (log cleanup), not resource-critical.

cli/azd/extensions/azure.ai.agents/internal/cmd/doctor/checks_auth.go
jongio: Verified in 16153f3ba - `redactUPN` + `authDetails` look correct. Tests cover both paths.

cli/azd/extensions/azure.ai.agents/internal/cmd/doctor.go
jongio: Reasonable - opportunistic guidance on an already-passing report is the right tradeoff. The non-nil return from `assembleState` means the worst case is missing a suggestion, not a crash.

wbreza (commented):
Delta review of the 5 commits pushed today (`a017e251`..`da2efb5d2`). Prior reviewer threads from Copilot and @jongio look resolved; this pass only checks the new fix-ups for regression and completeness.

### ✅ Verified
- **`4bbe4d0e` centralize nextstep stdout gating** — `nextstep_output.go` correctly delegates `writerIsTerminal()` to `w == os.Stdout && isTerminal()`, preserving the TTY-gate semantics introduced in `b9c1349a`. All call sites in `invoke.go`, `run.go`, `show.go`, `init.go`, `init_from_code.go` route through the new helper; no stale stdout/TTY checks remain in the nextstep path. No conflict with the existing `isTerminal()` in `helpers.go`.
- **`0d35d47b` redact doctor auth UPN by default** — `redactUPN()` and `authDetails()` redact both `Message` and `Details`; the "upn" key is omitted entirely from `Details` when not `--unredacted`. The new `TestCheckAuth_RedactsUPNByDefault`, `TestCheckAuth_UnredactedKeepsUPN`, and `TestCheckAuth_RedactionWithoutUPNDropsPrefix` cover PASS/WARN/FAIL × redacted/unredacted matrices.
- **`a017e251` remove unused nextstep auth state** — `AuthState`/`AuthUnknown|Authed|Unauthed`, `State.IsAuthenticated`, and `WithAuthProbe()` removed cleanly; no remaining references in the package; matched test deletions in `state_test.go` (no orphan assertions).
- **`1b0b6a73` pin `AgentVersionIdle` wire value** — `"idle": string(AgentVersionIdle)` is now in `TestErrorCodeWireValues`; the consumers at `show.go` (`shouldResolveShowNextStep`) and `nextstep/resolver.go` both still key off `AgentVersionIdle`, so silent literal drift will fail loudly.
- **`da2efb5d` restore `.tmp/` gitignore** — `.gitignore` entry restored with comment; all scratch files removed from the tracked tree; no leaked artifacts.

### Nit (non-blocking)
- **`cli/azd/extensions/azure.ai.agents/internal/cmd/nextstep_output_test.go`** — the new tests verify the suppression path (`writerIsTerminal == false`), but no test asserts the happy path where the helper actually invokes the underlying print when stdout is a real terminal. The gate logic is trivial (`w == os.Stdout && isTerminal()`) and the end-to-end behavior is covered by the per-command tests, so this is genuinely optional. If you want full unit coverage of the helper in isolation, the typical pattern is to make `isTerminal` a package-level `var` so a test can swap it out.

Nothing else from my side — looks ready to merge once @jongio's resolutions are reflected in his review state.

therealjohn (changes requested):
For the doctor command:
Capitalize the beginning of a sentence. for example:

```
 ✓ PASS  Foundry project endpoint reachable
               endpoint reachable (HTTP 200)
```

Should become:

```
 ✓ PASS  Foundry project endpoint reachable
               Endpoint reachable (HTTP 200)
```

Improve the output formatting and break down into categories. Include support for a --debug flag for more verbose output, but the default is more concise. Include clear next steps at the end.

Current:
```
azd ai agent doctor
  ✓ PASS  azd extension reachable
          azd extension reachable (version 0.1.32-preview).
  ✓ PASS  azure.yaml present and valid
          azure.yaml parsed (project: ai-foundry-starter-basic).
  ✓ PASS  azd environment selected
          environment selected: cpu-test-dev.
  ✓ PASS  agent service in azure.yaml
          1 agent service(s) in azure.yaml: hello-world-python-responses
  ✓ PASS  AZURE_AI_PROJECT_ENDPOINT set
          AZURE_AI_PROJECT_ENDPOINT = https://ai-account-asdf.services.ai.azure.com/api/projects/ai-project-adc-test-1
  ✓ PASS  agent.yaml valid (per service)
          agent.yaml valid for 1 service(s)
  ✓ PASS  manual env vars set
          no manual env vars are missing
  - SKIP  Manifest toolboxes have endpoint env vars set
          skipped: no toolbox resources declared in any service's agent.manifest.yaml.
  ✓ PASS  authentication
          <redacted> · token valid for 58 minutes
  ✓ PASS  Foundry project endpoint reachable
          endpoint reachable (HTTP 200)
  ✓ PASS  Developer has required role on Foundry project
          the current principal has the required role on project 'ai-account-asdf/ai-project-adc-test-1'.
  ✗ FAIL  Hosted agents are active
          1 of 1 agents have not been deployed:
  hello-world-python-responses: service "hello-world-python-responses" has not been deployed yet (AGENT_HELLO_WORLD_PYTHON_RESPONSES_NAME is unset).
          fix:  Run `azd deploy` to deploy the missing agents.
  - SKIP  Agent identity role assignments
          skipped: agent status check did not pass (see check `remote.agent-status`).
  ✗ FAIL  Manifest model deployments exist in Foundry
          1 model deployment(s) referenced by agent.manifest.yaml are missing on account ai-account-g4uvbu4lxidzu: AZURE_AI_MODEL_DEPLOYMENT_NAME (service hello-world-python-responses)
          fix:  Run `azd provision` to create the missing deployment(s), or update the agent.manifest.yaml `resources[].name` entries to match deployments that already exist in Foundry.
  - SKIP  Manifest connections exist on Foundry project
          skipped: no connection resources declared in any service's agent.manifest.yaml.

Summary: 10 passed, 2 failed, 3 skipped, 0 warned
```

Expected (default mode, concise):
```
Local
   (✓) azd extension reachable (version 0.1.32-preview)
   (✓) azure.yaml present and valid
   (✓) azd environment selected: cpu-test-dev
   (✓) agent service in azure.yaml (1 service)
   (✓) agent.yaml valid (1 service)
   (✓) manual env vars set
   (-) toolbox endpoint env vars -- skipped (no toolbox resources declared)
 
   Authentication
   (✓) credentials valid (token expires in 58 minutes)
 
   Remote
   (✓) AZURE_AI_PROJECT_ENDPOINT reachable
   (✓) developer has required Foundry role
   (x) hosted agents are active
       1 agent not deployed: hello-world-python-responses
       (AGENT_HELLO_WORLD_PYTHON_RESPONSES_NAME is unset)
   (-) agent identity role assignments -- skipped (depends on agent status)
   (x) manifest model deployments exist in Foundry
       1 missing: AZURE_AI_MODEL_DEPLOYMENT_NAME (hello-world-python-responses)
   (-) manifest connections -- skipped (no connections declared)
 
   10 passed, 2 failed, 3 skipped
 
   To fix, run these commands in order:
 
     1. azd provision    -- create the missing model deployment(s)
     2. azd deploy       -- deploy the agent(s)
```

with `--debug` (verbose):

```
Local
   (✓) azd extension reachable
       azd extension reachable (version 0.1.32-preview).
   (✓) azure.yaml present and valid
       azure.yaml parsed (project: ai-foundry-starter-basic).
   (✓) azd environment selected
       environment selected: cpu-test-dev.
   (✓) agent service in azure.yaml
       1 agent service(s) in azure.yaml: hello-world-python-responses
   (✓) AZURE_AI_PROJECT_ENDPOINT set
       AZURE_AI_PROJECT_ENDPOINT = https://ai-account-g4uvbu4lxidzu.services...
   (✓) agent.yaml valid (per service)
       agent.yaml valid for 1 service(s)
   (✓) manual env vars set
       no manual env vars are missing
   (-) Manifest toolboxes have endpoint env vars set
       skipped: no toolbox resources declared in any service's agent.manifest.yaml.
 
   Authentication
   (✓) authentication
       <redacted> -- token valid for 58 minutes
 
   Remote
   (✓) Foundry project endpoint reachable
       endpoint reachable (HTTP 200)
   (✓) Developer has required role on Foundry project
       the current principal has the required role on project
       'ai-account-g4uvbu4lxidzu/ai-project-adc-test-1'.
   (x) Hosted agents are active
       1 of 1 agents have not been deployed:
       hello-world-python-responses: service "hello-world-python-responses"
       has not been deployed yet (AGENT_HELLO_WORLD_PYTHON_RESPONSES_NAME is
       unset).
   (-) Agent identity role assignments
       skipped: agent status check did not pass (see check
       `remote.agent-status`).
   (x) Manifest model deployments exist in Foundry
       1 model deployment(s) referenced by agent.manifest.yaml are missing on
       account ai-account-g4uvbu4lxidzu: AZURE_AI_MODEL_DEPLOYMENT_NAME
       (service hello-world-python-responses)
   (-) Manifest connections exist on Foundry project
       skipped: no connection resources declared in any service's
       agent.manifest.yaml.
 
   10 passed, 2 failed, 3 skipped
 
   To fix, run these commands in order:
 
     1. azd provision    -- create the missing model deployment(s)
     2. azd deploy       -- deploy the agent(s)
 
   Then re-run `azd ai agent doctor` to verify.
```

When everything passes, provide the next step to invoke it: 

```
Local
   (✓) azd extension reachable (version 0.1.32-preview)
   (✓) azure.yaml present and valid
   ...
 
   Authentication
   (✓) credentials valid (token expires in 58 minutes)
 
   Remote
   (✓) all agents active
   (✓) model deployments exist
   ...
 
   All 13 checks passed.
 
   Next:  azd ai agent invoke <payload> -- send a message to hello-world-python-responses
```

@trangevi what if we move this into a separate extension so it could be used for any of them, not just agent? `azure.ai.doctor` and use `azd ai doctor`?

trangevi:
I agree with John, let's pull `doctor` out into a separate extension. Please look at this PR for the scaffold work I did for the other new extensions: https://github.com/Azure/azure-dev/pull/8219

antriksh30:
Doctor text-mode UX redesign landed in 804913a35 in response to @therealjohn's review. Summary of what changed:

**1. Default = concise.** PASS shows just the check name; FAIL shows a one-line Message + `fix:` Suggestion; SKIP inlines the reason after -- skipped. Use `--debug` (the existing persistent root flag from the azdext SDK) to surface the verbose path with full Message + Suggestion + Links.

**2. Section grouping.** Checks now render under `Local`, `Authentication`, and `Remote` headers, derived from the check ID prefix. `remote.auth` gets its own Authentication section per the literal mock.

**3. New glyph format.** `()`, `(x)`, `(-)`, `(!)`, `()`. ASCII `x` for FAIL matches the mock literally.

**4. "To fix" footer on failure.** When at least one failure maps to a canonical remediation, the footer is a numbered, deduplicated command list in execution order (auth  init  provision  deploy). When all failures are unmapped (or any are unmapped alongside mapped ones), the footer defers to the per-check `fix:` notes rendered in the body. The re-run instruction always closes the block so the user is never left without an actionable next step.

**5. First-letter capitalization at render time** with a brand-name blocklist so `azd`, `azure.yaml`, `agent.yaml`, `agent.manifest.yaml`, and `skipped:` leads stay lowercase. Source check strings are untouched (single source of truth).

Summary line simplified from `Summary: 1 passed, 1 failed, 1 skipped, 0 warned` to `1 passed, 1 failed, 1 skipped`. Warn/info segments are appended only when non-zero.

The trailing `Next:` block is suppressed when the To-fix footer fires.

JSON output (`--output json`) is unchanged.

antriksh30:
Thanks for the delta review @wbreza, much appreciated. On the optional `writerIsTerminal` happy-path test (the proposal to make `isTerminal` a swappable package var so the equality + true-leg can both be exercised) — I'm going to skip it for now given your own framing that "end-to-end behavior is covered by the per-command tests" and that the change would be "genuinely optional." Happy to revisit if a regression surfaces or another reviewer wants it pinned.

wbreza (commented):
Delta re-review on commit `804913a` — the doctor text-mode renderer redesign. Three files under `ext/azure.ai.agents/internal/cmd/`. No CRITICAL or HIGH issues; renderer is single-threaded so no concurrency risk, streaming/buffered parity is properly pinned in both modes, `--debug` plumbing is consistent, and `Next:` suppression on failure is correct.

A few things worth a follow-up — none blocking on their own:

### Medium

**1. `doctor_format.go:463-469` — stale doc comment in `capitalize`.** The doc comment lists `"cancelled"` as a brand-blocklist prefix, but `noCapitalizePrefixes` does not contain it, and `TestCapitalize` pins `"cancelled by user" → "Cancelled by user"`. The runner emits `Message: "cancelled"` on `ctx.Err()`, which will render as `Cancelled`. Either the comment is stale or the prefix was meant to be in the blocklist — please pick one and align doc + table + test.

**2. `doctor_format.go:360-393` — stated `auth → init → provision → deploy` order does not match `remediationForCheckID`.** The actual `order` values are `local.environment-selected: 1`, `local.azure-yaml / local.agent-service-detected: 5`, `remote.auth: 10`, `remote.foundry-endpoint / …: 20`, `remote.agent-status: 30`. So a failure mix of `local.azure-yaml + remote.auth` renders `azd ai agent init` **before** `azd auth login` — the opposite of the contract in the commit message. Probably the intended UX (you'd `init` before logging in), but no test covers this mix, so a regression that flipped the orders would not be caught. Two options:
- Correct the commit-message ordering claim and add a one-line comment on `remediationForCheckID` explaining why init precedes auth.
- Add a test case covering `[local.azure-yaml, remote.auth]` failures and assert the rendered order.

**3. `doctor_format.go:57-73` — section-header invariant silently depends on `NewRemoteChecks` ordering.** `writeCheck` emits a section header on every `cat != r.lastCat` transition. The clean `Local → Authentication → Remote` grouping only works because `remote.auth` happens to be the first remote check in `NewRemoteChecks`. If anyone reorders the remote slice or adds a second auth-like check elsewhere, the report will start emitting interleaved/duplicated headers like `Remote → Authentication → Remote`. `TestPrintDoctorReportText_Sections` feeds checks in the "correct" order, so it can't catch this regression. Either:
- Buffer-by-category in `printDoctorReportText` (and adapt the streaming path to flush sections in fixed order), or
- Add a comment on `NewRemoteChecks` that `remote.auth` must remain first, plus a regression test that asserts exactly one occurrence of `\nAuthentication\n` and one of `\nRemote\n` in a realistic check sequence.

### Low / optional

**4. `doctor_format.go:492-500` — `azd` brand match only catches `"azd "` and `"azd."`.** A message like `"azd-cli not found"` or `"azd, run init first"` matches neither prefix and gets capitalized to `Azd…`. Not triggered by any current check string; defensive only. If you want strict brand handling, gate on `r == 'a'` plus a word-boundary check rather than a prefix match.

**5. `doctor_format.go:194-212` — `writeIndentedBlock` emits a trailing whitespace line when the body ends in `\n`.** `strings.Split("foo\n", "\n")` returns `["foo", ""]`, so a trailing newline produces an extra `contPrefix`-only line. Concise mode is safe (uses `firstLine`); verbose mode passes `Message`/`Suggestion` straight through. One-line guard: `if i == len(lines)-1 && line == "" { continue }`.

**6. `doctor_format.go:344-354` — `orderedRemediations` stability depends on every command having a unique `order`.** Today they do, so output is deterministic. Add a second remediation with the same `order` and the To-fix list will flicker between runs. Worth a one-line comment ("each command must have a unique order field") or a tiebreaker sort key on command name for safety.

---

Re-review scope: only the delta in `804913a` against the five contracts in the commit message. Earlier commits already covered in the prior pass.

therealjohn (approved):

trangevi (approved):

trangevi (approved):

therealjohn (approved):

Issue #8187, opened 2026-05-14
Support creating routines from a file


therealjohn:
Routines should be source controlled with the repo so the `routine` commands should support a file. 

Need to design how this impacts agent.yaml and update AgentSchema for routines.

therealjohn:
This is already possible

Issue #51, opened 2022-07-11
Clean up extra .ps1 files included in .vsix release


danieljurek:
On release we noticed a couple extra powershell files in the .vsix files.

PR #8200, opened 2026-05-15
doc: add design spec for `azd ai agent routine` commands

### LINKED ISSUE #8159: Add `azd ai routine create/update/show/list/delete` (plus `enable`, `disable`, `dispatch`, and `run` subcommands) to manage and run Foundry Routines from any directory
### Feature Request

Please add `azd ai routine` commands to the agent extension that provide full CRUD, dispatch, and run-history access on Routines in a Foundry Project, following the design specifications in <a href="https://github.com/coreai-microsoft/foundrysdk_specs/pull/165">foundrysdk_specs PR #165</a>. This issue is part of the umbrella consolidation effort tracked in #8135 and is a sibling to the toolbox work in #8143, the connection work in #8139, and the project-context work in #8124.

#### Design Highlights
- **Command surface:** `azd ai routine create | update | show | list | delete`, `azd ai routine enable | disable`, `azd ai routine dispatch`, `azd ai routine run list | show | delete`.
- **Trigger as a flag:** `--trigger <schedule|timer|github-issue>` plus per-trigger flags (`--cron` / `--time-zone` for `schedule`, `--at` / `--time-zone` for `timer`, `--connection` / `--owner` / `--repository` / `--event-action` for the stretch `github-issue` trigger). Same prompt / no-prompt behavior as `connection create`.
- **Action as a flag:** `--action <agent-response|agent-invoke>` (defaults to `agent-response`), with `--agent-id` / `--agent-endpoint-id` / `--conversation-id` / `--session-id` as appropriate per action. CLI value names are short aliases for the API discriminators `invoke_agent_responses_api` and `invoke_agent_invocations_api`.
- **State management:** `enable` / `disable` exposed as dedicated verbs; CLI hides whether the wire format is action routes (`:enable` / `:disable`) or a field on PUT.
- **Dispatch:** `azd ai routine dispatch <name>` is synchronous by default; `--async` switches to the `:dispatchAsync` route. Supports `--input "<text>"` for the dispatch payload's user message and `--conversation-id` to continue an existing conversation thread.
- **Run history:** `azd ai routine run list <routine>` with `--top` / `--orderby` / `--filter` matching the data-plane query parameters. Single-run `show` and `delete` are pending API support per the routines spec.
- **Update is merge:** GET-then-PUT internally so only named flags change; switching `--trigger` or `--action` type after creation is out of scope (delete-and-recreate, mirroring the connection auth-type rule).

#### Open design questions (from the spec, Open Questions 4-6)
1. Trigger / action enum values: ship CLI-friendly aliases (`agent-response`, `agent-invoke`; `schedule`, `timer`, `github-issue`) that map to the API discriminators (`invoke_agent_responses_api`, `invoke_agent_invocations_api`; `schedule`, `timer`, `github_issue`), or use 1:1 API parity values? Default proposed: short CLI aliases.
2. Multi-trigger routines via the CLI. The Routines TypeSpec models `Routine.triggers` as a `Record<RoutineTrigger>` (a named dict). The flat-flag CLI shape implies a single trigger, keyed by the CLI as `default`. Add a `routine trigger add | remove | list` sub-collection now, or defer? Default proposed: defer.
3. `--conversation-id` on `azd ai routine dispatch`. The routines spec has `conversation_id` as a first-class field on `DispatchRequest`; the [TypeSpec PR's](https://github.com/Azure/azure-rest-api-specs/pull/42779) `DispatchRoutineRequest` does not include it. Default proposed: ship the CLI flag now, document as preview-only until the routines spec resolves.

These need Routines-service alignment before implementation.

#### API dependencies
- Routines data plane: `PUT /routines/{name}` (idempotent), `GET /routines/{name}`, `GET /routines`, `DELETE /routines/{name}` -- per [Routines TypeSpec PR #42779](https://github.com/Azure/azure-rest-api-specs/pull/42779).
- Routine dispatch: `POST /routines/{name}:dispatch` (sync) and `POST /routines/{name}:dispatchAsync` (async).
- Routine runs: `GET /routines/{name}/runs` (list, ready in TypeSpec). Single-run `GET` and `DELETE` are pending API work per the routines spec.
- `enable` / `disable` wire format: routines spec defines them as action routes (`:enable` / `:disable`); the TypeSpec PR currently models `enabled` as a field on the PUT body. CLI verbs are stable either way.

---
Please refer to the [full design spec (foundrysdk_specs PR #165)](https://github.com/coreai-microsoft/foundrysdk_specs/pull/165) -- specifically the *Routines* sub-section under Requirements, the API Surface table (rows 12-14), and Open Questions 4-6 -- for implementation guidance. The matching service-side spec lives in [foundrysdk_specs/specs/agents/routines/spec.md](https://github.com/coreai-microsoft/foundrysdk_specs/blob/main/specs/agents/routines/spec.md).

#### Additional Context
- Part of the `azd ai` direct-commands surface for Foundry. Sibling issues: #8124 (project context), #8139 (connections + agent run secrets), #8143 (toolboxes). Umbrella: #8135.
- Tracked in the routines spec as fast follow (AZD CLI work is documented as TBD pending TypeSpec lock).

huimiu:
## Summary

Adds the design spec for the new `routine` command subtree (`cli/azd/docs/design/ai-routine-design-spec.md`). A routine pairs one trigger (when) with one action (what) on a Foundry project, for example "every weekday at 8 AM UTC, invoke the daily report agent", without standing up Logic Apps, Functions, or cron infra.

Related issues:

- Feature: Azure/azure-dev#8159
- Follow up: Azure/azure-dev#8187 (`--file` for `routine create`)

## Key points

- **Placement**: lives in the existing `azure.ai.agents` extension. Today the commands surface as `azd ai agent routine …`; they become `azd ai routine …` after the planned extension split (registration only). No `azure.yaml` or `registry.json` impact, no new persistent state.
- **v1 surface**: `create`, `update`, `show`, `list`, `delete`, `enable`, `disable`, `dispatch`, and `run list`. `run show` and `run delete` are deferred until the data plane lands them.
- **Endpoint resolution**: reuses the standard 5 level cascade and works standalone outside an azd project, matching `connection`, `toolbox`, and `skill`. The `Routines=V1Preview` opt in header is sent on every call.
- **Create vs. update**: the data plane is a single idempotent PUT; the CLI splits it for usability. `create` fails on existing (or upserts with `--force`). `update` is GET then PUT preserving untouched fields, with type switching of `--trigger` / `--action` rejected client side.
- **enable / disable / dispatch**: dedicated idempotent verbs hide the wire format. `dispatch` is sync by default and streams the agent response; `--async` switches to the async route.
- **Wire mapping**: kebab case CLI aliases (`recurring`, `agent-response`, `agent-invoke`) read naturally and absorb upstream API renames via a single mapping table.
- **Telemetry and tests**: one event per command (no PII), plus unit coverage for flag to wire mapping, update merge semantics, and dispatch route selection, with an E2E smoke covering the create through delete loop.

### DIFF
--- a/cli/azd/docs/design/ai-routine-design-spec.md
+++ b/cli/azd/docs/design/ai-routine-design-spec.md
@@ -35,10 +35,11 @@ sub-`run` group via `newRoutineRunCommand`. No new extension; no `registry.json`
 change.
 
 > **Command surface.** The agents extension registers its root as `agent`, so
-> these commands surface as **`azd ai agent routine …`** today. The umbrella
-> [command-spec.md](./command-spec.md) refers to them as `azd ai routine …`,
-> which is the eventual surface after the extension is split/renamed — a
-> registration-only change with no behavior diff.
+> these commands surface as **`azd ai agent routine …`** today. The eventual
+> umbrella surface is `azd ai routine …` after the extension is split/renamed,
+> which is a registration-only change with no behavior diff. See feature issue
+> [#8159](https://github.com/Azure/azure-dev/issues/8159) for the umbrella
+> context.
 
 ### Impact on existing commands
 
@@ -58,7 +59,7 @@ direct agent call (does not). Both must keep working.
 ### Out of scope
 
 - Declarative routines (`routine.yaml`, `azd provision` integration, `azd up`) —
-  belongs to the orchestrated config-driven model.
+  belong to the orchestrated config-driven model.
 - Multi-trigger routines via the CLI — deferred ([§7 OQ-2](#7-open-questions)).
 - Changing `--trigger` or `--action` *type* on an existing routine — delete and
   recreate, mirroring the `connection` auth-type rule ([§4.2](#42-create-vs-update)).
@@ -68,9 +69,10 @@ direct agent call (does not). Both must keep working.
 
 Every `routine` subcommand resolves the Foundry project endpoint through the
 standard 5-level cascade: `-p` / `--project-endpoint` flag → active azd env
-(`AZURE_AI_PROJECT_ENDPOINT`) → global config (`extensions.ai-agents.context.endpoint`)
-→ `FOUNDRY_PROJECT_ENDPOINT` env var → structured dependency error
-(code `CodeMissingProjectEndpoint`).
+(`AZURE_AI_PROJECT_ENDPOINT`) → global config (the `endpoint` field of the
+`extensions.ai-agents.project.context` object, written by
+`azd ai agent project set`) → `FOUNDRY_PROJECT_ENDPOINT` env var → structured
+dependency error (code `CodeMissingProjectEndpoint`).
 
 Standalone usability is required: every `routine` subcommand must work outside an
 azd project given a resolvable endpoint, matching `connection`, `toolbox`, and
@@ -128,9 +130,10 @@ other fields are preserved verbatim. Accepted flags: `--description`, `--cron`,
 `--time-zone`, `--at`, `--agent-id`, `--agent-endpoint-id`, `--conversation-id`,
 `--session-id`.
 
-**Type-switch guard.** Switching `--trigger` or `--action` *type* on an existing
-routine is rejected client-side with a `delete and recreate` suggestion. The
-server PUT would accept it, but the CLI refuses for surface stability.
+**Type-switch guard.** `--trigger` and `--action` are registered on `update`
+solely to surface a friendly client-side error when supplied: the command exits
+non-zero with a `delete and recreate` suggestion before calling the service.
+This mirrors the `connection` auth-type rule.
 
 **Post-merge validation.** After applying the named fields, `update` validates
 the merged body against the existing trigger/action type:
@@ -194,12 +197,13 @@ Maps onto `GET /routines/{routine_name}/runs`:
 
 | CLI flag      | Query param        |
 | ------------- | ------------------ |
-| `--top N`     | `maxResults`       |
+| `--top N`     | `maxResults` per page; CLI stops auto-paging once `N` items have been returned |
 | `--orderby`   | `orderBy` (repeatable) |
 | `--filter`    | `filter`           |
 
 Auto-pagination via `pageToken` / `next_page_token`, same rules as `routine list`
-([§4.3](#43-routine-show-name--routine-list)).
+([§4.3](#43-routine-show-name--routine-list)). When `--top N` is set the CLI
+caps the total returned at `N` items across all drained pages.
 
 ### 4.8 `routine run show` / `routine run delete`
 
@@ -213,6 +217,7 @@ churn on already-shipped verbs.
 | Command   | Table output                   | JSON output                         |
 | --------- | ------------------------------ | ----------------------------------- |
 | `create`  | `Routine '<name>' created.` + summary | Server `Routine` body          |
+| `update`  | `Routine '<name>' updated.` + summary of changed fields | Updated `Routine` body |
 | `delete`  | `Routine '<name>' deleted.`    | `{ "deleted": true, "name": "<name>" }` |
 | `enable`  | `Routine '<name>' enabled.`    | Updated `Routine` body              |
 | `disable` | `Routine '<name>' disabled.`   | Updated `Routine` body              |
@@ -249,18 +254,24 @@ For `agent-response`, the CLI validates "exactly one of `--agent-id` /
 
 ### 5.3 Routes and API status
 
-All requests include the `RoutinesPreviewHeader`.
+All requests include the `RoutinesPreviewHeader` and the `api-version=v1` query
+parameter, matching the existing toolboxes/agents Foundry clients in this
+extension (for example
+[`listen.go`](../../extensions/azure.ai.agents/internal/cmd/listen.go) builds
+`/toolboxes/{name}/versions/{version}/mcp?api-version=v1`). The
+`continuationToken` and `pageToken` query parameters are added on top of
+`api-version` where applicable.
 
 | CLI verb                              | HTTP                                                          | API status      |
 | ------------------------------------- | ------------------------------------------------------------- | --------------- |
 | `routine create` / `routine update`   | `PUT  {endpoint}/routines/{name}`                             | Ready           |
 | `routine show`                        | `GET  {endpoint}/routines/{name}`                             | Ready           |
-| `routine list`                        | `GET  {endpoint}/routines?continuationToken=…`                | Ready           |
+| `routine list`                        | `GET  {endpoint}/routines` (with `continuationToken`)         | Ready           |
 | `routine delete`                      | `DELETE {endpoint}/routines/{name}`                           | Ready           |
 | `routine enable` / `routine disable`  | GET-then-PUT toggling `enabled` ([§4.5](#45-routine-enable--disable-name)) | Ready (field on PUT) |
 | `routine dispatch`                    | `POST {endpoint}/routines/{name}:dispatch`                    | Ready           |
 | `routine dispatch --async`            | `POST {endpoint}/routines/{name}:dispatchAsync`               | Ready           |
-| `routine run list`                    | `GET  {endpoint}/routines/{name}/runs?...`                    | Ready           |
+| `routine run list`                    | `GET  {endpoint}/routines/{name}/runs`                        | Ready           |
 | `routine run show` *(deferred)*       | `GET  {endpoint}/routines/{name}/runs/{run-id}`               | Not in TypeSpec |
 | `routine run delete` *(deferred)*     | `DELETE {endpoint}/routines/{name}/runs/{run-id}`             | Not in TypeSpec |
 
@@ -293,7 +304,7 @@ endpoints hashed.
 | # | Question | Default proposal |
 |---|----------|------------------|
 | 1 | **Trigger / action enum names.** CLI aliases (`recurring`, `agent-response`, `agent-invoke`) vs. 1:1 API parity (`schedule`, `invoke_agent_responses_api`, …). Note: feature issue [#8159](https://github.com/Azure/azure-dev/issues/8159) uses `schedule`; this spec proposes `recurring`. | Ship CLI aliases. API names are verbose on the command line; a single mapping table absorbs upstream renames. |
-| 2 | **Multi-trigger routines.** TypeSpec `triggers` is `Record<RoutineTrigger>`. Add `routine trigger add | remove | list` now? | Defer. All hero scenarios use one trigger, keyed as `"default"`. Re-evaluate when a real multi-trigger scenario lands. |
+| 2 | **Multi-trigger routines.** TypeSpec `triggers` is `Record<RoutineTrigger>`. Add `routine trigger add \| remove \| list` now? | Defer. All hero scenarios use one trigger, keyed as `"default"`. Re-evaluate when a real multi-trigger scenario lands. |
 | 3 | **`--conversation-id` on dispatch.** Field is in the routines conceptual spec but not in TypeSpec PR #42779. | Ship the flag, mark preview-only in `--help`. If the service rejects unknown fields, the user sees a service error and re-runs without it. Revisit on TypeSpec lock. |
 
 ## 8. Test Plan
--- a/cli/azd/docs/design/ai-routine-design-spec.md
+++ b/cli/azd/docs/design/ai-routine-design-spec.md
@@ -69,7 +69,7 @@ direct agent call (does not). Both must keep working.
 Every `routine` subcommand resolves the Foundry project endpoint through the
 standard 5-level cascade: `-p` / `--project-endpoint` flag → active azd env
 (`AZURE_AI_PROJECT_ENDPOINT`) → global config (`extensions.ai-agents.context.endpoint`)
-→ `FOUNDRY_PROJECT_ENDPOINT` env var → structured `exterrors.Dependency` error
+→ `FOUNDRY_PROJECT_ENDPOINT` env var → structured dependency error
 (code `CodeMissingProjectEndpoint`).
 
 Standalone usability is required: every `routine` subcommand must work outside an
@@ -108,7 +108,7 @@ Optional flags:
 **Prompt / no-prompt** — mirrors `connection create`:
 
 - Interactive: missing required per-trigger / per-action flags are prompted for.
-- `--no-prompt`: exits non-zero with `exterrors.Validation` listing missing flags.
+- `--no-prompt`: exits non-zero with a structured validation error listing missing flags.
 
 **Output:**
 
@@ -140,7 +140,7 @@ the merged body against the existing trigger/action type:
   exclusive: specifying one clears the other; specifying both is a validation
   error.
 - If the merged body no longer satisfies required fields for its trigger/action
-  type, the command exits with `exterrors.Validation` before calling the service.
+  type, the command exits with a structured validation error before calling the service.
 
 ### 4.3 `routine show <name>` / `routine list`
 
@@ -152,8 +152,8 @@ all pages are drained).
 ### 4.4 `routine delete <name>`
 
 Confirmation prompt by default. `--force` skips it. In `--no-prompt` mode,
-`--force` is required; without it the command exits non-zero with
-`exterrors.Validation`. Matches `connection delete`.
+`--force` is required; without it the command exits non-zero with a structured
+validation error. Matches `connection delete`.
 
 ### 4.5 `routine enable | disable <name>`
 
@@ -310,7 +310,7 @@ endpoints hashed.
 - `dispatch` sync vs. `--async` route selection; leading GET triggered/skipped
   based on payload flags; `actionType` telemetry `unknown` in the no-payload path.
 - `run list` query-param mapping and pagination; JSON output is one stable object.
-- `delete --no-prompt` without `--force` produces `exterrors.Validation`.
+- `delete --no-prompt` without `--force` produces a structured validation error.
 - Output shapes match [§4 table](#output-shapes-for-state-changing-verbs) in both
   table and JSON modes.
 
--- a/cli/azd/docs/design/ai-routine-design-spec.md
+++ b/cli/azd/docs/design/ai-routine-design-spec.md
@@ -53,13 +53,15 @@ direct agent call (does not). Both must keep working.
 ### In scope
 
 - The commands listed in [§1](#1-summary).
-- Mapping from CLI flags onto the wire format in [TypeSpec PR #42779](https://github.com/Azure/azure-rest-api-specs/pull/42779).
+- Mapping from CLI flags onto the wire format in [TypeSpec PR #43186](https://github.com/Azure/azure-rest-api-specs/pull/43186) (merged into `feature/foundry-release`).
 - Reuse of the 5-level project endpoint resolver (flag → azd env → global config → `FOUNDRY_PROJECT_ENDPOINT` → structured error).
 
 ### Out of scope
 
 - Declarative routines (`routine.yaml`, `azd provision` integration, `azd up`) —
-  belong to the orchestrated config-driven model.
+  the imperative `routine create`/`update` verbs in this spec cover the v1 jobs-to-be-done;
+  the declarative `routine.yaml` + `provision`/`up` story belongs to the future
+  orchestrated config-driven model and is intentionally out of scope here.
 - Multi-trigger routines via the CLI — deferred ([§7 OQ-2](#7-open-questions)).
 - Changing `--trigger` or `--action` *type* on an existing routine — delete and
   recreate, mirroring the `connection` auth-type rule ([§4.2](#42-create-vs-update)).
@@ -95,7 +97,7 @@ Cross-cutting flags on every subcommand: `--output table|json`, `--no-prompt`,
 ### 4.1 `routine create <name>`
 
 Required positional: `<name>`.\
-Required flags (always): `--trigger`.\
+Required flags (always): `--trigger <recurring|timer|github-issue>` (enum, not free-form; see [§5.1](#51-trigger-flags--routinetrigger-discriminator) for the supported types and per-type required flags).\
 Conditionally required flags: per trigger/action type (see [§5.1](#51-trigger-flags--routinetrigger-discriminator) / [§5.2](#52-action-flags--routineaction-discriminator)).
 
 Optional flags:
@@ -127,7 +129,7 @@ an upsert (matches `connection create --force`).
 
 **Update semantics.** GET-then-PUT internally — only the named flags change; all
 other fields are preserved verbatim. Accepted flags: `--description`, `--cron`,
-`--time-zone`, `--at`, `--agent-id`, `--agent-endpoint-id`, `--conversation-id`,
+`--time-zone`, `--at`, `--agent-name`, `--agent-endpoint-id`, `--conversation-id`,
 `--session-id`.
 
 **Type-switch guard.** `--trigger` and `--action` are registered on `update`
@@ -139,7 +141,7 @@ This mirrors the `connection` auth-type rule.
 the merged body against the existing trigger/action type:
 - Action-specific flags are accepted only for the current action type
   (`--conversation-id` → `agent-response`; `--session-id` → `agent-invoke`).
-- For `agent-response`, `--agent-id` and `--agent-endpoint-id` remain mutually
+- For `agent-response`, `--agent-name` and `--agent-endpoint-id` remain mutually
   exclusive: specifying one clears the other; specifying both is a validation
   error.
 - If the merged body no longer satisfies required fields for its trigger/action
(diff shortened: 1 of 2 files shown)


cli/azd/docs/design/ai-routine-design-spec.md
```diff
@@ -0,0 +1,365 @@
...
+- Reuse of the 5-level project endpoint resolver (flag → azd env → global config → `FOUNDRY_PROJECT_ENDPOINT` → structured error).
+
+### Out of scope
+
+- Declarative routines (`routine.yaml`, `azd provision` integration, `azd up`) —
+  belong to the orchestrated config-driven model.
+- Multi-trigger routines via the CLI — deferred ([§7 OQ-2](#7-open-questions)).
+- Changing `--trigger` or `--action` *type* on an existing routine — delete and
+  recreate, mirroring the `connection` auth-type rule ([§4.2](#42-create-vs-update)).
+- Creating routines from a file (`--file`) — tracked as [#8187](https://github.com/Azure/azure-dev/issues/8187).
```
therealjohn: Add this back to the scope. We need to be able to create from a file so routines can live in the repo as source controlled items.

therealjohn (changes requested):

cli/azd/docs/design/ai-routine-design-spec.md
```diff
@@ -0,0 +1,365 @@
...
+- `azd ai agent routine update <name>`
+- `azd ai agent routine show <name>`
+- `azd ai agent routine list`
+- `azd ai agent routine delete <name>`
+- `azd ai agent routine enable <name>`
+- `azd ai agent routine disable <name>`
+- `azd ai agent routine dispatch <name>`
+- `azd ai agent routine run list <routine>`
+
+`routine run show` and `routine run delete` are deferred until their APIs ship
```
lindazqli, line 22: for routine run show is this what you need: https://github.com/Azure/azure-rest-api-specs/pull/43186/changes#diff-0920b2f67a7816e1e9ef440782ce714e40358a2a5c161b322271b19c19fb1e9fR163

cli/azd/docs/design/ai-routine-design-spec.md
```diff
@@ -0,0 +1,365 @@
...
+
+### In scope
+
+- The commands listed in [§1](#1-summary).
+- Mapping from CLI flags onto the wire format in [TypeSpec PR #42779](https://github.com/Azure/azure-rest-api-specs/pull/42779).
+- Reuse of the 5-level project endpoint resolver (flag → azd env → global config → `FOUNDRY_PROJECT_ENDPOINT` → structured error).
+
+### Out of scope
+
+- Declarative routines (`routine.yaml`, `azd provision` integration, `azd up`) —
```
lindazqli, line 61: sorry my naive question - is azd provision and azd up needed? thought azd ai routine create would cover the JTBD

cli/azd/docs/design/ai-routine-design-spec.md
```diff
@@ -0,0 +1,365 @@
...
+
+## 4. Command Behavior
+
+Cross-cutting flags on every subcommand: `--output table|json`, `--no-prompt`,
+`--debug`, `-p` / `--project-endpoint`.
+
+### 4.1 `routine create <name>`
+
+Required positional: `<name>`.\
+Required flags (always): `--trigger`.\
```
lindazqli: just for my understanding - the parameter behind trigger is freeformed?

lindazqli: there are three types of triggers: timer, recurring and event based. developers will just provide the info in their trigger definition?

cli/azd/docs/design/ai-routine-design-spec.md
```diff
@@ -0,0 +1,365 @@
...
+> uses `schedule` (the API discriminator name). The CLI uses `recurring` because
+> it reads more naturally alongside `timer` on the command line, and the CLI
+> already kebab-cases multi-word values everywhere. A single mapping table
+> absorbs any upstream rename. See [§7 OQ-1](#7-open-questions).
+
+| CLI `--trigger` | TypeSpec `type`  | Required CLI flags                                                   | Status |
+| --------------- | ---------------- | -------------------------------------------------------------------- | ------ |
+| `recurring`     | `schedule`       | `--cron "<expr>"`, `--time-zone <tz>`                                | v1     |
+| `timer`         | `timer`          | `--at "<ISO 8601>"`, `--time-zone <tz>`                              | v1     |
+| `github-issue`  | `github_issue`   | `--connection <id>`, `--owner <o>`, `--repository <r>`, `--event-action <a>` (repeatable) | Deferred — pending workspace connection model |
```
lindazqli: it seems typespec is making this only specific to github, checking with the team

lindazqli: there is only github-issue strong typed trigger, we plan to add a few other strong typed trigger and a generic event based trigger, coming soon next week. Would this risking azd plan?

cli/azd/docs/design/ai-routine-design-spec.md
```diff
@@ -0,0 +1,365 @@
...
+service's 404.
+
+### 4.6 `routine dispatch <name>`
+
+Sync by default → `POST /routines/{name}:dispatch`.
+
+| Flag                  | Notes                                                                |
+| --------------------- | -------------------------------------------------------------------- |
+| `--async`             | Switches to `:dispatchAsync`. Returns `dispatch_id` immediately.     |
+| `--input "<text>"`    | User-message payload wrapped into `RoutineDispatchPayload`.          |
```
lindazqli: this --input ''<text>" can accept json format if customer provides?

cli/azd/docs/design/ai-routine-design-spec.md
lindazqli: we will update the typespec next monday with the generic and a few more strong typed ones

cli/azd/docs/design/ai-routine-design-spec.md
```diff
@@ -0,0 +1,365 @@
...
+`monitor`, `files`, `sessions`), `project` (`set` / `unset` / `show`), or
+`azure.yaml`. No new persistent state in `~/.azd/config.json`. The existing
+`agent invoke` and the new `routine dispatch` deliberately overlap: `dispatch`
+is the trigger-side manual fire (records a `RoutineRunDto`); `invoke` is the
+direct agent call (does not). Both must keep working.
+
+### In scope
+
+- The commands listed in [§1](#1-summary).
+- Mapping from CLI flags onto the wire format in [TypeSpec PR #42779](https://github.com/Azure/azure-rest-api-specs/pull/42779).
```
lindazqli: **Stale TypeSpec PR reference.** This spec cites PR #42779 throughout (§1, §4.8, §7 OQ-3). The routines TypeSpec landed in PR #43186 (now merged into `feature/foundry-release`). Please update all references to #43186 so readers can cross-check the wire format accurately.

cli/azd/docs/design/ai-routine-design-spec.md
```diff
@@ -0,0 +1,365 @@
...
+
+Confirmation prompt by default. `--force` skips it. In `--no-prompt` mode,
+`--force` is required; without it the command exits non-zero with a structured
+validation error. Matches `connection delete`.
+
+### 4.5 `routine enable | disable <name>`
+
+Dedicated verbs that hide the wire format. Today: GET-then-PUT toggling
+`enabled: true | false`. If the service later adds `:enable` / `:disable` action
+routes, the CLI flips silently — the verb contract does not change.
```
lindazqli: **`enable`/`disable` routes already exist in TypeSpec.** The spec says these use "GET-then-PUT toggling `enabled: true | false`" and speculates "If the service later adds `:enable` / `:disable` action routes, the CLI flips silently." However, TypeSpec PR #43186 already defines these as dedicated operations:
```
POST /routines/{routine_name}:enable  → enableRoutine
POST /routines/{routine_name}:disable → disableRoutine
```
The CLI should call these directly rather than doing a GET-then-PUT. The GET-then-PUT fallback is unnecessary and creates a TOCTOU race.

cli/azd/docs/design/ai-routine-design-spec.md
```diff
@@ -0,0 +1,365 @@
...
+
+Both are idempotent: enabling an already-enabled routine (or disabling an
+already-disabled one) is a no-op success. Non-existent routines surface the
+service's 404.
+
+### 4.6 `routine dispatch <name>`
+
+Sync by default → `POST /routines/{name}:dispatch`.
```
lindazqli: **No sync `dispatch` route exists in TypeSpec.** This line proposes `POST /routines/{name}:dispatch` as the sync path, but TypeSpec PR #43186 only defines one dispatch operation:
```
POST /routines/{routine_name}:dispatch_async  → dispatchRoutineAsync
```
There is no synchronous `:dispatch` route. The "Sync by default" contract described here has no backing API. Either:
- Clarify that `dispatch` always maps to `:dispatch_async` (both sync and `--async` modes), or
- Track getting a sync `:dispatch` route added to the TypeSpec before shipping this verb.

Also note the route uses an underscore (`dispatch_async`), not camel case — the routes table in §5.3 should reflect that.

cli/azd/docs/design/ai-routine-design-spec.md
```diff
@@ -0,0 +1,365 @@
...
+
+CLI emits `triggers: { "default": { "type": "<wire>", ... } }` to match the
+TypeSpec `Record<RoutineTrigger>` shape. The key `"default"` is an implementation
+detail (single-trigger CLI shape) and is not surfaced to the user.
+
+### 5.2 Action flags → `RoutineAction` discriminator
+
+| CLI `--action`          | TypeSpec `type`                  | Required CLI flags                              | Optional CLI flags    |
+| ----------------------- | -------------------------------- | ----------------------------------------------- | --------------------- |
+| `agent-response` (def.) | `invoke_agent_responses_api`     | one of `--agent-id` / `--agent-endpoint-id`     | `--conversation-id`   |
```
lindazqli: **`--agent-id` maps to the wrong TypeSpec field.** The TypeSpec field for the project-scoped agent name is `agent_name`, not `agent_id`:
```typespec
@doc("The project-scoped agent name for responses API dispatch.")
@maxLength(256)
agent_name?: string;
```
The flag should be `--agent-name` (or the wire key should be documented as `agent_name`). Using `agent_id` / `--agent-id` implies an ID (UUID/numeric), whereas this is a human-readable name used for project-scoped lookup. This also affects the telemetry property name in §6 (`agent_id` → `agent_name` or `agentName`).

cli/azd/docs/design/ai-routine-design-spec.md
```diff
@@ -0,0 +1,365 @@
...
+| Async | `DispatchRoutineResponse` (no streaming) | Same                            |
+
+### 4.7 `routine run list <routine>`
+
+Maps onto `GET /routines/{routine_name}/runs`:
+
+| CLI flag      | Query param        |
+| ------------- | ------------------ |
+| `--top N`     | `maxResults` per page; CLI stops auto-paging once `N` items have been returned |
+| `--orderby`   | `orderBy` (repeatable) |
```
lindazqli: **`--orderby` has no backing TypeSpec query parameter.** `ListRoutineRunsParameters` in TypeSpec PR #43186 only spreads `CommonPageQueryParameters` (which covers pagination) plus the `filter` query param. There is no `orderBy` parameter defined. Either remove `--orderby` from v1 scope or get `orderBy` added to the TypeSpec before shipping `run list`.

cli/azd/docs/design/ai-routine-design-spec.md
lindazqli: **Field name mismatch on the GitHub issue trigger.** The spec maps `--owner` → GitHub assignee/org filter, but the TypeSpec field is `assignee`, not `owner`:
```typespec
@doc("The GitHub assignee or organization filter that scopes which issues can fire the trigger.")
@maxLength(128)
assignee: string;
```
`--owner` would serialize to `owner` on the wire, which the service won't recognize. The flag should be `--assignee` to match the TypeSpec (or the spec should explicitly document the CLI-to-wire mapping for this field).

Additionally, `--event-action` has no corresponding field in `GitHubIssueOpenedRoutineTrigger` — it only has `connection_id`, `assignee`, and `repository`. Since this trigger type is deferred anyway, please remove `--event-action` from the table or track it as a separate TypeSpec gap.

lindazqli (changes requested):
Reviewed against TypeSpec PR #43186 (`adding routines`, now merged into `feature/foundry-release`). Six issues found — four are blockers before shipping v1:

1. **[Line 56]** TypeSpec PR reference is stale (#42779 → should be #43186).
2. **[Line 165]** `enable`/`disable` should call the dedicated `:enable` / `:disable` routes that already exist in the TypeSpec, not GET-then-PUT.
3. **[Line 173]** No sync `:dispatch` route exists in TypeSpec — only `POST :dispatch_async`. The "sync by default" contract has no API backing.
4. **[Line 249]** `--agent-id` / `agent_id` should be `--agent-name` / `agent_name` to match the TypeSpec field name.
5. **[Line 201]** `--orderby` has no `orderBy` query param in `ListRoutineRunsParameters`.
6. **[Line 239]** GitHub trigger: `--owner` serializes to `owner` but TypeSpec field is `assignee`; `--event-action` has no TypeSpec counterpart.

lindazqli (commented):
Correction: the previous review was submitted as REQUEST_CHANGES by mistake — please treat all inline comments as informational notes for discussion, not as blockers. The comments remain valid observations to align the spec with TypeSpec PR #43186.

cli/azd/docs/design/ai-routine-design-spec.md
```diff
@@ -0,0 +1,365 @@
...
+Auto-pagination via `pageToken` / `next_page_token`, same rules as `routine list`
+([§4.3](#43-routine-show-name--routine-list)). When `--top N` is set the CLI
+caps the total returned at `N` items across all drained pages.
+
+### 4.8 `routine run show` / `routine run delete`
+
+**Not registered in v1.** The data-plane endpoints are not in
+[TypeSpec PR #42779](https://github.com/Azure/azure-rest-api-specs/pull/42779).
+These will be added as a strictly additive change when the APIs land, with no
+churn on already-shipped verbs.
```
jongio: Missing from the spec: what does the user see on errors? Some cases the implementer will need to decide on:

- `create` without `--force` on an existing name: 409 conflict? Structured error with a 'delete and recreate' hint?
- `show`/`delete`/`dispatch` on a non-existent routine: how does the service 404 surface?
- `update`'s GET succeeds, then the routine gets deleted before the PUT lands: 404 on the PUT with no context?

The existing design specs in this repo don't have explicit error sections, so this isn't blocking. But since azd routes errors through `error_suggestions.yaml` and `ErrorWithSuggestion`, a one-liner like 'errors follow the `connection` command pattern' would prevent the implementer from having to reverse-engineer the expected error handling approach. The test plan would also benefit from negative cases covering service-side failures (404, 409, auth).

jongio (commented):
Spec follows the extension's existing patterns well for endpoint resolution, flag naming, and prompt/no-prompt behavior. The TypeSpec alignment feedback from existing reviews is the main thing to address before implementation.

One gap: error behavior isn't defined for any of the 9 commands. See inline comment.

cli/azd/docs/design/ai-routine-design-spec.md
huimiu: Updated spec to add --file back

huimiu:
Thanks everyone for the careful review! Here's what changed to address all the feedback:

**TypeSpec alignment (Linda's feedback against PR #43186)**

- Updated all stale `#42779` refs to `#43186`.
- `enable` / `disable` now call the dedicated `:enable` / `:disable` routes directly — no more GET-then-PUT (avoids the TOCTOU race you flagged).
- `dispatch` (both default and `--async`) now hits `:dispatch_async`. There's no sync route in TypeSpec, so the flag just controls client-side waiting/streaming.
- Renamed `--agent-id` → `--agent-name` everywhere (spec body, command summary, telemetry) to match the `agent_name` TypeSpec field.
- Dropped `--orderby` from v1 — no `orderBy` query param exists yet.
- GitHub trigger: `--owner` → `--assignee`, removed `--event-action`.
- Added a heads-up note in §5.1 for the upcoming generic + strong-typed triggers.
- Clarified `--input` is plain text (not parsed as JSON), and added an inline pointer that `--trigger` is an enum, not free-form.
- Noted that the `run show` API now exists in #43186 — keeping registration deferred for v1 but called out in §4.8.

**Error behavior + `--file`  (John and Jon's feedback)**

- @jongio: added a new §4.9 **Error Behavior** section with a scenario → type/code/suggestion table covering 409 (`create` without `--force`), 404s (`show`/`delete`/`dispatch`/`enable`/`disable` on missing routine), the `update` GET-succeeds-then-PUT-404 race, type-switch, auth, endpoint resolution, and 5xx. Reuses the existing `connection`/`toolbox` typed-error pattern. Test plan updated with negative cases.
- @therealjohn: pulled `--file` back into v1 scope. New §4.1.1 covers the file shape, mutual exclusion with `--trigger`, flag-level override semantics, and points to #8187 for the schema details. `update --file` is supported too. Command summary, telemetry (`hasFile`), and test plan updated to match.

therealjohn (approved):

wbreza (commented):
## Code Review — Design Spec Quality Pass

Thanks for the thorough iteration on this spec @huimiu — the prior round of feedback (TypeSpec alignment, error section, `--file` re-scoping) is well incorporated. The findings below are **net-new** spec-quality items intended to reduce implementation ambiguity. The PR is approvable as-is; these are suggestions to tighten the spec before it becomes the implementation source of truth.

> **Note on review state:** Posting as `COMMENT` rather than `REQUEST_CHANGES` since @therealjohn has already approved and many items are clarifications, not blockers.

---

### 🔴 Critical — recommend addressing before merge or in a fast follow-up

1. **`github-issue` enum out of sync across three locations.**
   - §1 summary (L68) and §4.1 enum (L105) advertise `--trigger <recurring|timer|github-issue>` as v1.
   - §5.1 trigger table marks `github-issue` as *"Deferred — pending workspace connection model."*
   - §9 reference summary correctly omits it.
   - **Fix:** Remove `github-issue` from §1 and §4.1 enum; align with §9.

2. **§4.9 error table lacks HTTP status codes.**
   The Type/Code columns are documented but implementers can't map service responses without explicit status codes. Notably, `CodeRoutineAlreadyExists` (line ~276) doesn't state `409`.
   **Fix:** Add an `HTTP Status` column (400/404/409/429/503 as applicable).

3. **§9 command summary misrepresents conditional requirements.**
   Lines 508-514 show `[--cron "0 8 * * *"]` and `[--at "..."]` both as optional, but §5.1 makes each required per-trigger-type. The summary isn't copy-pasteable.
   **Fix:** Split into separate recurring/timer examples, or add inline comments noting "`--cron` for recurring, `--at` for timer."

4. **Orphan `### Output shapes for state-changing verbs` heading** (between §4.8 and §4.9).
   Unnumbered; §8 references it informally as `[§4 table]`.
   **Fix:** Number it (e.g., §4.8.1) and update the §8 anchor.

5. **`dispatch` streaming/failure contract under-specified** (§4.6).
   - Streaming protocol unspecified (SSE? chunked? line-buffered?).
   - JSON output shape ambiguous when streaming aggregates.
   - Polling timeout, max attempts, and reconnection semantics undefined.
   - Mid-stream failure: is partial output flushed? what exit code?
   - The leading `GET /routines/{name}` implementation note doesn't specify what happens if that GET fails.
   **Fix:** Add a dedicated subsection (or expand §4.6) covering the streaming contract and failure modes; reference an existing streaming command (e.g., `agent invoke`) if one exists.

6. **Broken/awkward anchor links across the doc.**
   Several internal links break across line wraps or include double-dashes from pipe-to-anchor conversion:
   - L105-108: `(#51-trigger\nr-flags…)` — wrap artifact
   - L264-265: TypeSpec PR URL split with `#\n#diff-…`
   - L200-201, 207, 241, 288, 348, 391: double-dash anchor IDs from `|` chars in headings
   **Fix:** Rewrap URLs onto single lines; verify all `(#…)` anchors resolve against actual heading IDs.

7. **Exit codes not specified for any verb.**
   Per AGENTS.md convention, errors should produce specific exit codes. Scripting consumers can't reliably distinguish validation errors from service errors.
   **Fix:** Add a §4.9.1 (or table column) listing exit codes per scenario.

8. **`--file` path resolution & validation under-specified.**
   - Resolved against cwd or project root?
   - Path traversal/symlink/absolute-path handling?
   - Max file size?
   - §8 lacks negative tests for malformed YAML/JSON.
   **Fix:** Document path resolution rules in §4.1.1; add negative test cases in §8.

---

### 🟡 Important

9. **`--cron` dialect unspecified.** Quartz vs Unix 5-field? Validated client-side or server-side? Error message format for invalid expressions? (§5.1)

10. **`--at` ISO 8601 edge cases.** Past time → error or schedule next occurrence? `"2026-04-24T15:00:00"` without `Z` treated as local or UTC? DST transitions? (§5.1)

11. **Concurrency races undefined.** What happens for: `update` during active `dispatch`? `delete` while routine is `enabled`? Concurrent `enable`/`disable`? At minimum, document expected server-side guarantees. (§4.2, §4.4, §4.5, §4.6)

12. **429/503 retry semantics + network timeouts** not covered in §4.9. Does the CLI retry with backoff, or surface immediately?

13. **403 RBAC vs 401 auth conflated** under `CodeAuthFailed`/`CodeNotLoggedIn` (L283). Different remediation paths — clarify mapping.

14. **`create --force` with conflicting immutable fields** (different `--trigger` or `--action`) — replace silently, or fail? (§4.2)

15. **Telemetry "no PII" not defined** (§6). Are agent names, conversation IDs, `--assignee` usernames, repository names captured? Enumerate explicitly.

16. **`run list` output formats (table/JSON) not specified** (§4.7). Mirror `routine list` shape?

17. **Test plan gaps** (§8): no explicit coverage for interactive prompts (`AZD_FORCE_TTY=false`), wire-mapping table per trigger×action, dispatch streaming/backpressure, all 9 telemetry events, `--file` negative cases, `timer` trigger in E2E smoke.

18. **`--file` + per-trigger override flags** (§4.1.1 L134-142). Spec says `--file` mutually exclusive with `--trigger` but cooperates with `--cron`/`--at` — yet those are per-trigger-type. Clarify with valid/invalid examples.

---

### 🟢 Nice to have

19. **`--async` help text** should explicitly say "returns immediately; use `routine run list` for status" — non-obvious otherwise. (§4.6)

20. **`run` subgroup contains only `list` in v1.** Empty-group pattern — consider `routine list-runs` for v1 and defer the full `run` subgroup. (§1, §4.7-4.8)

21. **`--input` guidance.** Recommend documenting a size cap and a "don't pass secrets via `--input`" note. (§4.6)

22. **TOCTOU undefined on first use** (L191). Add "(Time-of-Check-Time-of-Use)".

23. **Endpoint cascade phrasing** — described as "5-level" but step 5 is the structured error. Relabel as "4-level cascade with structured-error fallback." (§3 L77-82)

24. **Missing newline in §1** (L4): `infra.Commands` renders as one word.

25. **§9 reference summary inconsistency** (L454): `[--enabled=false]` shows a literal value while other flags use placeholders.

---

_Reviewed against TypeSpec PR #43186 references in the spec and existing patterns in `cli/azd/extensions/azure.ai.agents/` and sibling design docs._

kristenwomack:
Hi @huimiu,

Thanks for the iteration on this one. The error section and the `:enable` / `:disable` route fix look good.

Items to address:

First, the `github-issue` trigger says three different things in this spec. §1 and §4.1 list it as v1, §5.1 marks it deferred, and §9 leaves it out. Could you pick one and make the rest match? My vote is to drop it from §1 and §4.1, since the connection model isn't ready.

Second, `dispatch` default mode is the most user-facing flow in this whole surface, and the streaming behavior isn't written down. We don't say whether the stream is SSE or chunked, what the JSON output looks like when the stream aggregates, how long we poll before giving up, or what happens to partial output if something fails mid-stream. Could you add a subsection that spells that out? I don't want every integrator guessing.

Third, the `run` subgroup ships with only `list` in v1. Do we want `azd ai agent routine list-runs` until `run show` and `run delete` arrive, or are we OK with a subgroup that has one verb in it for now? Small call, but easier to make now than to rename later.

Fourth, `--conversation-id` on `dispatch` is preview, and it's not in TypeSpec yet. Marking it preview in `--help` is fine with me, and I want to be sure we're OK with shipping a flag the service might reject. Could you add a sentence on what the user will see if that happens?

Follow-ups, not blocking: @wbreza's review flags exit codes, the `--cron` dialect, `--at` without a timezone, 429/503 retry behavior, and 401 vs. 403. Could you open issues for the ones you don't fold in here so we don't lose them?

Thanks, Kristen

kristenwomack:
Hey @huimiu, @therealjohn's already approved and @wbreza left a solid pass — I won't repeat any of that. Three things from me:

1. **`github-issue` trigger** — the spec lists it in the v1 enum but the trigger table marks it deferred. Pick one. My vote is **defer** so v1 is just `recurring` + `timer` and we don't ship a half-wired trigger.
2. **`run` subgroup with only `list`** — empty group feels off. @wbreza's `routine list-runs` suggestion is cleaner for v1. Let's go that route and add the `run` subgroup back when `show`/`delete` land.
3. **`dispatch` sync default** — there's no sync route server-side, so "sync" is really a client-side poll. That's fine, but let's make sure the help text says so and we have a sane default timeout. What's the plan if the poll hangs?

Also +1 on @wbreza's call to nail down exit codes and the streaming contract before this becomes the implementation source of truth.

PR #8160, opened 2026-05-13
doc: design spec for azd ai agent toolbox direct commands

hund030:
## Summary

This PR adds the design spec for the `azd ai agent toolbox` direct commands. This is for design review only and contains no implementation code.

### What's in this spec

- **Five top-level verbs** for managing versioned toolboxes against a Foundry project:
  - `toolbox create <name>`: register a toolbox (pending tools until first `connection add`).
  - `toolbox update <name> --default-version <n>`: retarget the active version.
  - `toolbox delete <name>`: delete the whole toolbox or a single version via `--version`.
  - `toolbox show <name>`: display metadata, version body, and the toolbox's MCP endpoint URL for runtime consumption.
  - `toolbox list`: list toolboxes plus any local pending records for the resolved endpoint.

- **Connection subgroup** (`connection add | remove | list`) that wraps the toolbox versions API. Every tool mutation is a full-`tools[]` POST to `/toolboxes/{name}/versions` followed by a `PATCH default_version`; tool-entry shape is inferred from the project connection's ARM `category` (`RemoteTool` → `mcp`, `CognitiveSearch` → `azure_ai_search`).

- **Tag subgroup** (`tag set | remove | list`) shipped as a Compatibility stub: toolboxes are not yet exposed as ARM resources, so the standard ARM resource-tag surface is unavailable. The CLI surface is final and activates without surface changes once the underlying API exists.

- **Pending-toolbox config store**: `create` cannot POST because the service requires a non-empty `tools[]`, so it records a local entry under `extensions.ai-agents.pending-toolboxes.<endpointHash>.items.<name>`. The first `connection add` promotes the record to a real v1.

- **Live-verified API surface** (§ 4): every endpoint the spec relies on was probed against a live Foundry project, including the per-version delete guard, the MCP runtime endpoint, and the ARM connection lookup used by `connection add`.

### Key design decisions for review

1. **Extension placement** (§ 3): toolbox commands live inside the existing `azure.ai.agents` extension at `azd ai agent toolbox …`. Files are isolated under `internal/cmd/toolbox*.go` and `internal/pkg/toolbox/` so a future move to a standalone `azure.ai.toolbox` extension is a registration-only change.
2. **`create` does not POST** (§ 5.1, Open Question 1): the service requires a non-empty `tools[]`. Two options were considered — auto-seed with a sentinel built-in, or defer the first POST until the first `connection add`. Current proposal: defer (pending-record model). Keeps `create` non-network and matches the "one command, one action" principle.
[...]

PR #8203, opened 2026-05-15
feat(toolboxes): add azd ai toolbox direct commands

### LINKED ISSUE #8143: Add `azd ai toolbox create/update/show/list/delete` (plus `connection` and `tag` subcommands) to manage Foundry Toolboxes from any directory
### Feature Request

Please add `azd ai toolbox` commands to the agent extension that provide full CRUD on Toolboxes in a Foundry Project, following the design specifications in [foundrysdk_specs PR #165](https://github.com/coreai-microsoft/foundrysdk_specs/pull/165). This issue is part of the umbrella consolidation effort tracked in #8135 and is a sibling to the connection work in #8139 and project-context work in #8124.

#### Design Highlights
- **Command surface:** `azd ai toolbox create | update | delete | show | list`, `azd ai toolbox connection add | remove | list`, `azd ai toolbox tag set | remove | list`
- **Versioned resource:** `azd ai toolbox create <name>` auto-detects -- creates a new toolbox (v1) if `<name>` does not exist on the project, or publishes a new version of the existing toolbox if it does.
- **Connection composition:** Toolboxes contain connection-backed tools only. Connections must already exist on the project (created via `azd ai connection create`); standalone built-in tools (e.g., code interpreter) are wired directly on the agent, not via toolboxes.
- **Tag management:** Mutates Azure resource tags on the toolbox via `tag set | remove | list`.

#### Open design questions (from the spec, Open Question 3)
1. Per-version targeting on `show` / `list` / `delete` (default to latest? `--version <n>` flag? `list --all-versions`?)
2. Whether `update` mutates the latest version in place or always publishes a new version
3. Whether `connection add` / `connection remove` mutate the latest version in place or publish a new version
4. `--force` semantics on a versioned `create` (force-publish identical version? overwrite latest? error if no diff?)

These need Toolboxes-service alignment before implementation.

#### API dependencies
- Toolbox data-plane API (versioned resource): create / publish-new-version, update, delete, show, list -- to be confirmed by the Toolboxes service team.
- Toolbox connections collection API: add / remove / list -- collection mutation semantics to be confirmed.
- Tag operations ride the standard ARM resource-tag surface.

---
Please refer to the [full design spec (foundrysdk_specs PR #165)](https://github.com/coreai-microsoft/foundrysdk_specs/pull/165) -- specifically the *Toolboxes* sub-section under Requirements, the API Surface table (rows 4-6), and Open Question 3 -- for implementation guidance.

#### Additional Context

hund030:
## Summary

Adds the `azd ai toolbox` command group as a new standalone `azure.ai.toolboxes` extension so users can manage versioned, connection-backed tool collections from the terminal. Closes #8143. Implements the design spec from #8160.

## Changes

- **`extensions/azure.ai.toolboxes/internal/cmd/`** — verbs `create`, `update`, `delete`, `show`, `list`, `version list`, and the `connection {add, remove, list}` subgroup. `create` takes its input from a JSON/YAML file (`--from-file`) describing the initial tool set and publishes v1 directly; `connection add` supports both a single positional and the same file mode; `connection remove` confirms by default and accepts `--force`; `update` retargets `default_version`.
- **`internal/foundry/{credential.go, projectctx/, connections/}`** — duplicated infrastructure from `azure.ai.agents` (5-level project-endpoint cascade, Foundry credential factory, single-connection lookup client). Isolated under `internal/foundry/` with a one-way import contract so a future shared-module lift is mechanical.
- **`internal/pkg/azure/foundry_toolsets_client.go`** — toolbox CRUD data-plane client (`ListToolboxes`, `Get/List/DeleteToolboxVersion`, `SetDefaultVersion`, capped paginator).
- **`internal/cmd/toolbox_files.go`** — JSON/YAML file parser for `--from-file` with `DisallowUnknownFields`/`KnownFields` and context-aware error suggestions.
- **`internal/exterrors/`** — trimmed copy carrying only the codes/ops the toolbox surface uses.
- **`extensions/azure.ai.agents/`** — toolbox-specific files and codes removed; pre-existing `foundry_toolsets_client.go` / `exterrors/codes.go` reverted to `main`.
- **`.golangci.yaml`, `AGENTS.md`** — added at the extension root for consistency with sibling extensions.
- **Tests** — table-driven coverage for every verb's branches against a stubbed `toolboxClient` and `connectionResolver`, plus the file-parser and shared helpers.


cli/azd/extensions/azure.ai.toolboxes/internal/cmd/context.go
```diff
@@ -0,0 +1,165 @@
...
+}
+
+// validateOutputFormat returns a structured error when --output is not table/json.
+// The azd host normally enforces this via RegisterFlagOptions; the check stays
+// for direct `azd x` invocation and for unit-test reach.
+func validateOutputFormat(out string) error {
+	switch strings.ToLower(out) {
+	case "", "table", "json":
+		return nil
+	default:
```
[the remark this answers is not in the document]
hund030: Normalized.

cli/azd/extensions/azure.ai.agents/internal/cmd/toolbox_list.go
```diff
@@ -0,0 +1,140 @@
...
+	fmt.Fprintln(w, "NAME\tDEFAULT-VERSION\tSTATE\tCREATED")
+	fmt.Fprintln(w, "----\t---------------\t-----\t-------")
+
+	sortedLive := slices.Clone(live)
+	slices.SortFunc(sortedLive, func(a, b azure.ToolboxObject) int {
+		return cmp.Compare(a.Name, b.Name)
+	})
+
+	for _, t := range sortedLive {
+		fmt.Fprintf(w, "%s\t%s\t\t\n", t.Name, t.DefaultVersion)
```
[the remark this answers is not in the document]
hund030: Live rows now print STATE=live, CREATED=-.

cli/azd/extensions/azure.ai.agents/internal/pkg/azure/foundry_toolsets_client.go
```diff
@@ -65,6 +66,116 @@ func NewFoundryToolboxClient(
...
+) ([]T, error) {
+	type page struct {
+		Data    []T    `json:"data"`
+		HasMore bool   `json:"has_more,omitempty"`
+		LastID  string `json:"last_id,omitempty"`
+	}
+
+	out := []T{}
+	target := initialURL
+	for i := 0; i < maxPaginationPages; i++ {
```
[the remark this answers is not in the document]
hund030: Replaced with for range maxPaginationPages.

cli/azd/extensions/azure.ai.toolboxes/internal/cmd/toolbox_create.go
```diff
@@ -0,0 +1,128 @@
...
+		Use:   "create <name>",
+		Short: "Register a new toolbox locally (publishes on first `connection add`).",
+		Long: `Register a new toolbox locally.
+
+Foundry requires a non-empty tool list on the first POST, so 'create' does not
+contact the service. Instead it records a local pending entry. The first
+'connection add' against the same toolbox name publishes v1 and clears the
+pending record.`,
+		Args: cobra.ExactArgs(1),
+		RunE: func(cmd *cobra.Command, args []string) error {
```
therealjohn, line 49: Return an error with suggestion so its clear what to specify. All it says now is:

```
ERROR: accepts 1 arg(s), received 0
```

I'd expect it show more like the --help comamnd does:

```
Usage:
  agent toolbox create <name> [flags]

Foundry requires a non-empty tool list on the first POST, so 'create' does not
contact the service. Instead it records a local pending entry. The first
'connection add' against the same toolbox name publishes v1 and clears the
pending record.
```

cli/azd/extensions/azure.ai.agents/internal/cmd/toolbox_create.go
```diff
@@ -0,0 +1,128 @@
...
+	if alreadyExists {
+		fmt.Printf(
+			"Toolbox %s already exists. Run 'connection add' to publish a new version, "+
+				"or 'update --default-version <n>' to retarget.\n", name,
+		)
+		return nil
+	}
+	fmt.Printf(
+		"Registered toolbox %s (pending tools). "+
+			"Run 'azd ai agent toolbox connection add %s <connection>' to publish v1.\n",
```
therealjohn: Improve the formatting for this so its not all on one line - there might be an existing pattern for showing next step guidance and coloring, etc. 

If not, then just make each sentence on a new line.

cli/azd/extensions/azure.ai.agents/internal/cmd/toolbox_connection.go
```diff
@@ -0,0 +1,328 @@
...
+				cmd.Context(), args[0], args[1], *flags,
+				readToolboxFlags(cmd, extCtx),
+				defaultConnectionResolver{},
+			)
+		},
+	}
+
+	cmd.Flags().StringVar(
+		&flags.index, "index", "",
+		"Index name (required when the connection's category is CognitiveSearch).",
```
therealjohn: What is CognitiveSearch referring to? Need to check if that is the correct branding/service name now.

cli/azd/extensions/azure.ai.agents/internal/cmd/toolbox_connection.go
```diff
@@ -0,0 +1,328 @@
...
+					},
+				},
+			},
+		}, nil
+
+	default:
+		return nil, exterrors.Validation(
+			exterrors.CodeUnsupportedConnectionCategory,
+			fmt.Sprintf(
+				"connection %q has category %q; v1 supports RemoteTool and CognitiveSearch only",
```
therealjohn: I don't expect this limitation.

cli/azd/extensions/azure.ai.agents/internal/cmd/toolbox_connection.go
```diff
@@ -0,0 +1,328 @@
...
+			"connection":          conn.Name,
+			"connectionId":        conn.ID,
+			"category":            string(conn.Category),
+			"promotedFromPending": promoted,
+		}
+		return emitJSON(payload)
+	}
+	if promoted {
+		fmt.Printf(
+			"Published toolbox %s version %s with connection %s.\n",
```
therealjohn: Add the toolbox endpoint to the output on a new line so its easily copied by the dev after its published.

A better suggestion, if in an AZD project already, would be to recommend setting an env var to use like:

`azd env set <toolboxname>_MCP_ENDPOINT=<endpoint>` 

But fill in the values for the placeholders.

therealjohn (changes requested):

cli/azd/extensions/azure.ai.toolboxes/internal/cmd/toolbox_create.go
trangevi: That might be outside of extension control, I think that's coming from `Args: cobra.ExactArgs(1)` being specified

cli/azd/extensions/azure.ai.agents/internal/cmd/toolbox_connection.go
```diff
@@ -0,0 +1,328 @@
...
+		Use:   "add <toolbox> <connection>",
+		Short: "Attach a project connection to a toolbox.",
+		Long: `Attach a project connection to a toolbox.
+
+The tool entry shape is inferred from the connection's ARM category:
+  RemoteTool       → mcp tool with server_label/server_url from the connection target
+  CognitiveSearch  → azure_ai_search tool (requires --index)
+Other categories are rejected.
+
+If the toolbox has a local pending record (from 'toolbox create'), v1 is
```
trangevi: This means that if I want to create a toolbox with 3 tools, I'm going to end up on v3 before I'm done. This feels incredibly painful from a user perspective, as I'm not actually iterating on desired changes, I'm being forced into it. Is there a way this can be done without all these added, unwanted, versions?

cli/azd/extensions/azure.ai.agents/internal/cmd/toolbox_connection.go
```diff
@@ -0,0 +1,328 @@
...
+			Tools:       []map[string]any{entry},
+		}
+		created, err := client.CreateToolboxVersion(ctx, toolboxName, req)
+		if err != nil {
+			return exterrors.ServiceFromAzure(err, exterrors.OpCreateToolboxVersion)
+		}
+		if _, err := store.Clear(ctx, endpoint, toolboxName); err != nil {
+			return exterrors.Internal(
+				exterrors.CodePendingToolboxStoreFailed,
+				fmt.Sprintf("failed to clear pending toolbox record: %s", err),
```
trangevi: What is the implication here? Is this something the user can recover from? At the very least we should probably indicate that this is a client side only failure, and that the toolbox version was still created correctly.

cli/azd/extensions/azure.ai.agents/internal/cmd/toolbox_connection.go
```diff
@@ -0,0 +1,328 @@
...
+				fmt.Sprintf(
+					"run 'azd ai agent toolbox create %q' first, then re-run 'connection add'",
+					toolboxName,
+				),
+			)
+		}
+		return exterrors.ServiceFromAzure(err, exterrors.OpGetToolbox)
+	}
+
+	current, err := client.GetToolboxVersion(ctx, toolboxName, tb.DefaultVersion)
```
trangevi, line 155: Do we want a way to add a connection to the non-default toolbox version?

cli/azd/extensions/azure.ai.agents/internal/cmd/toolbox_connection_actions.go
```diff
@@ -0,0 +1,281 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
trangevi: Is there a reason that the add command is in one file, and the other commands are in a different file? I'd suggest either a separate file per command, like the toolbox base commands, or everything in one. Could separate out the helper methods if desired to help clean things up.

cli/azd/extensions/azure.ai.agents/internal/cmd/toolbox_create.go
```diff
@@ -0,0 +1,128 @@
...
+func newToolboxCreateCommand(extCtx *azdext.ExtensionContext) *cobra.Command {
+	extCtx = ensureExtensionContext(extCtx)
+	flags := &toolboxCreateFlags{}
+
+	cmd := &cobra.Command{
+		Use:   "create <name>",
+		Short: "Register a new toolbox locally (publishes on first `connection add`).",
+		Long: `Register a new toolbox locally.
+
+Foundry requires a non-empty tool list on the first POST, so 'create' does not
```
trangevi: This is different from every other resource `create` command we're adding. Should we require a tool to be provided as a part of this command, so we don't have this "pending" state that doesn't coincide with any other resource?

@therealjohn thoughts?

trangevi: Alternatively, should we have this command take a list of connection names, and create a new version of the toolbox with those? And the connection creation themselves would be handled via the connection commands, not a separate "toolbox connection" command?

(apologies for not reviewing the spec evidently, but this seems like a very weird user flow)

cli/azd/extensions/azure.ai.agents/internal/cmd/toolbox_delete.go
```diff
@@ -0,0 +1,261 @@
...
+		return err
+	}
+	if err := validateOutputFormat(parent.output); err != nil {
+		return err
+	}
+	if parent.noPrompt && !verb.force {
+		return exterrors.Validation(
+			exterrors.CodeMissingForceFlag,
+			"--no-prompt requires --force on destructive operations",
+			"add --force to confirm the deletion non-interactively",
```
trangevi: Wouldn't this only apply if we're trying to delete the default?

cli/azd/extensions/azure.ai.toolboxes/internal/cmd/toolbox_list.go
```diff
@@ -0,0 +1,140 @@
...
+		items, readErr := listPendingToolboxes(ctx, azdClient, endpoint)
+		if readErr != nil {
+			log.Printf("toolbox list: pending-toolbox read failed: %v", readErr)
+		} else {
+			pending = items
+		}
+	}
+
+	// Drop pending records that already exist live-side to avoid duplicates.
+	liveNames := map[string]struct{}{}
```
trangevi: If the toolbox has been published to the project, we shouldn't have a pending entry for it any more, correct? Is this necessary?

cli/azd/extensions/azure.ai.toolboxes/internal/cmd/toolbox_list.go
```diff
@@ -0,0 +1,140 @@
...
+	"github.com/azure/azure-dev/cli/azd/pkg/azdext"
+	"github.com/spf13/cobra"
+)
+
+// newToolboxListCommand returns the `azd ai agent toolbox list` command.
+func newToolboxListCommand(extCtx *azdext.ExtensionContext) *cobra.Command {
+	extCtx = ensureExtensionContext(extCtx)
+	cmd := &cobra.Command{
+		Use:   "list",
+		Short: "List toolboxes on the project, plus any local pending records.",
```
trangevi: Does this list all versions, or just the defaults?

cli/azd/extensions/azure.ai.toolboxes/internal/cmd/toolbox_show.go
```diff
@@ -0,0 +1,269 @@
...
+func newToolboxShowCommand(extCtx *azdext.ExtensionContext) *cobra.Command {
+	extCtx = ensureExtensionContext(extCtx)
+	flags := &toolboxShowFlags{}
+
+	cmd := &cobra.Command{
+		Use:   "show <name>",
+		Short: "Show a toolbox version, including its computed MCP endpoint.",
+		Long: `Show a toolbox.
+
+By default shows the default version. Use --version to inspect a specific
```
trangevi, line 36: How can a user know how many versions a toolbox has?

cli/azd/extensions/azure.ai.agents/internal/cmd/toolbox_update.go
```diff
@@ -0,0 +1,85 @@
...
+	extCtx = ensureExtensionContext(extCtx)
+	flags := &toolboxUpdateFlags{}
+
+	cmd := &cobra.Command{
+		Use:   "update <name>",
+		Short: "Update a toolbox (currently: retarget the default version).",
+		Long: `Update a toolbox.
+
+Only --default-version is mutable through PATCH today (§ 4.1). To edit the
+description or the tool list, publish a new version with 'connection add' or
```
trangevi: `connection add` doesn't let a user change the description, does it?

trangevi (changes requested):

cli/azd/extensions/azure.ai.agents/internal/cmd/toolbox_connection.go
```diff
@@ -0,0 +1,328 @@
...
+		Tools:       newTools,
+	}
+	created, err := client.CreateToolboxVersion(ctx, toolboxName, req)
+	if err != nil {
+		return exterrors.ServiceFromAzure(err, exterrors.OpCreateToolboxVersion)
+	}
+
+	if _, err := client.SetDefaultVersion(ctx, toolboxName, created.Version); err != nil {
+		return exterrors.ServiceFromAzure(err, exterrors.OpSetDefaultVersion)
+	}
```
jongio: If `SetDefaultVersion` fails here, we've already POSTed a new version to the service. That version exists server-side but isn't the default, and the error doesn't tell the user which version was created.

Two options: (a) include `created.Version` in the error so users can recover with `toolbox update --default-version <v>`, or (b) attempt a best-effort `DeleteToolboxVersion` rollback (with a log warning if that fails too).

Same pattern exists in `connection remove` (`toolbox_connection_actions.go` around line 120).

cli/azd/extensions/azure.ai.agents/internal/pkg/azure/foundry_toolsets_client.go
```diff
@@ -65,6 +66,116 @@ func NewFoundryToolboxClient(
...
+			return out, nil
+		}
+		last := p.LastID
+		if last == "" && pickLastID != nil {
+			last = pickLastID(p.Data[len(p.Data)-1])
+		}
+		if last == "" {
+			// HasMore=true with no cursor: bail rather than spin.
+			return out, nil
+		}
```
jongio: When `HasMore` is true but there's no usable cursor, this silently returns partial results. Callers like `ListToolboxes` and `ListToolboxVersions` won't know they got incomplete data. A `log.Printf` warning here would help debugging, something like `pagination: has_more=true but no cursor for %s; returning %d items`.

cli/azd/extensions/azure.ai.toolboxes/internal/cmd/context.go
```diff
@@ -0,0 +1,166 @@
...
+
+// toolboxFlags carries the cross-cutting flags shared by every `toolbox` verb.
+type toolboxFlags struct {
+	projectEndpoint string
+	output          string
+	noPrompt        bool
+}
+
+// toolboxNamePattern is the validation regex for toolbox and tool names per § 4.2.
+var toolboxNamePattern = regexp.MustCompile(`^[A-Za-z0-9_-]+$`)
```
jongio, line 34: This matches any length. The agent name regex in `parse.go` enforces 1-63 chars. Adding a length cap here would catch overly long names locally instead of letting them hit the server and come back as a less helpful 400.

jongio (commented):
Technical findings from code-level analysis. I'm not restating the design/UX feedback from @therealjohn and @trangevi; their reviews cover the broader picture.

Three findings, one medium priority:

1. **[MED] Non-atomic version promotion** `connection add` and `connection remove` both call `CreateToolboxVersion` then `SetDefaultVersion` as two separate API calls. If the second fails, there's an orphaned version that isn't the default, and the error doesn't include the created version number, so recovery via `toolbox update --default-version` requires the user to first figure out which version got created. Same pattern in `toolbox_connection_actions.go` around line 120.

2. **[LOW] Silent pagination truncation** `listPagedFromClient` returns partial results without error when the server responds with `has_more=true` but provides no usable cursor (lines 161-164 of `foundry_toolsets_client.go`). Unlikely in practice, but a `log.Printf` warning here would make debugging much easier if it ever happens.

3. **[LOW] No length cap on toolbox/tool names** `toolboxNamePattern` is `^[A-Za-z0-9_-]+$` with no min/max bounds. The existing agent name validation in `parse.go` enforces 1-63 chars. Worth adding a length cap so users get a clear local error rather than a less helpful service-side 400.

cli/azd/extensions/azure.ai.toolboxes/internal/cmd/toolbox_list.go
hund030: Just the defaults. One row per toolbox, showing default_version.

cli/azd/extensions/azure.ai.agents/internal/cmd/toolbox_update.go
hund030: Confirmed. Thanks for the catch. Fixed the help text to drop the description-editing claim.

cli/azd/extensions/azure.ai.agents/internal/cmd/toolbox_connection.go
hund030: Not in v1. We can have a `--version` flag on `connection add` if you feel necessary. Add @therealjohn for confirm.

cli/azd/extensions/azure.ai.toolboxes/internal/cmd/toolbox_show.go
hund030: The data-plane endpoint exists but no `azd ai agent toolbox version list` verb is exposed. We can have add it if you feel necessary. Also add @therealjohn to confirm.

cli/azd/extensions/azure.ai.toolboxes/internal/cmd/toolbox_create.go
hund030: Yes. The error comes from `cobra.ExactArgs(1)`, but we can customize the args func if we want to customize the error message.
I suppose to file a follow-up issue since the scope is the whole `azd ai agent` surface, not specific to toolbox.

cli/azd/extensions/azure.ai.agents/internal/cmd/toolbox_create.go
hund030: I found the `actions.ActionResult` structure for core-azd commands, but it doesn't apply to our extension. Extensions are cobra commands that write directly to stdout. 

I will make each sentence on a new line.

cli/azd/extensions/azure.ai.agents/internal/cmd/toolbox_connection.go
hund030: "CognitiveSearch" is the literal string the Foundry service returns as the connection type field. 

It's also used by the agent YAML manifest schema: https://github.com/Azure/azure-dev/blob/80e819f54281c1153828dd1b3dfe83cf36f2c701/cli/azd/extensions/azure.ai.agents/internal/pkg/agents/agent_yaml/yaml.go#L104

Updated help text: "CognitiveSearch is the category for Azure AI Search"

cli/azd/extensions/azure.ai.agents/internal/cmd/toolbox_connection.go
hund030: Added `Endpoint: ` line + `azd env set <NAME>_MCP_ENDPOINT` hint after successful add.

cli/azd/extensions/azure.ai.agents/internal/cmd/toolbox_connection.go
hund030: The full list of project connection categories is defined in https://github.com/Azure/azure-dev/blob/80e819f54281c1153828dd1b3dfe83cf36f2c701/cli/azd/extensions/azure.ai.agents/internal/pkg/azure/foundry_projects_client.go#L83-L95

The other categories don't have a corresponding tool shape in the Foundry service. 
Could you help clarify with service team?

cli/azd/extensions/azure.ai.agents/internal/cmd/toolbox_connection.go
hund030: Updated. Now the user keeps the success message, the stale local record is either silently reconciled by the next list or cleaned up by delete, and the underlying error is preserved in the log stream for --debug triage.

cli/azd/extensions/azure.ai.agents/internal/cmd/toolbox_connection_actions.go
hund030: Updated. One file per command now.

cli/azd/extensions/azure.ai.agents/internal/cmd/toolbox_delete.go
hund030: Right, the guard was over-broad. 

The `noPrompt && !force` check moved out of the shared path and now lives only in the whole-toolbox delete path.

cli/azd/extensions/azure.ai.toolboxes/internal/cmd/toolbox_list.go
hund030: Yes. The pending record is normally cleared when connection add publishes, but a clear failure can leave a stale entry. This dedup makes the list output self-healing.

cli/azd/extensions/azure.ai.agents/internal/cmd/toolbox_connection.go
hund030: Thanks for the suggestion. I picked the option (a). 

The error message now names the created version, and the suggestion is a copy-pasteable recovery command.

cli/azd/extensions/azure.ai.agents/internal/pkg/azure/foundry_toolsets_client.go
hund030: Good catch. Added a `log.Printf` here.

cli/azd/extensions/azure.ai.toolboxes/internal/cmd/context.go
hund030: Good catch. Added `maxToolboxNameLength = 63` in `toolbox.go` to mirror the agent-name ceiling, and both `validateToolboxName` and `validateToolName` now enforce regex + length together.

cli/azd/extensions/azure.ai.agents/internal/cmd/toolbox_connection.go
hund030: Agreed it's painful. 
I think we can change the command into a batch operation, but the extra inputs that some connection types need (today:  --index for CognitiveSearch) interact with the batch shape.

I just need to align with you on the interface design. Add @therealjohn.

What do think if we change the --index flag to accept a map?
```bash
azd ai agent toolbox connection add tb mcp1 search1 search2 \
    --index search1=docs,search2=tickets
```

In this case, what if there is only one connection? Always require `<conn>=<idx>` or accept `--index <idx>`?

cli/azd/extensions/azure.ai.toolboxes/internal/cmd/toolbox_show.go
therealjohn: Sounds good to me

cli/azd/extensions/azure.ai.agents/internal/cmd/toolbox_connection.go
therealjohn: Fine to skip for now and track in https://github.com/Azure/azure-dev/issues/8244 as long as devs can workaround this by setting a version as the default version and add tools.

cli/azd/extensions/azure.ai.agents/internal/cmd/toolbox_connection.go
therealjohn: Due to the verbosity of the command, and the requirements for a non-empty `tools[]` on `create`, let's pivot to use file input. Accept a JSON file with the toolbox config and connections information to create a toolbox.

cli/azd/extensions/azure.ai.toolboxes/internal/cmd/toolbox_create.go
therealjohn: No problem.

wbreza (commented):
## Code Review — supplemental findings

This review supplements existing reviews from @jongio, @therealjohn, @trangevi, and Copilot. It only surfaces **new** issues; the prior findings (non-atomic version promotion in `connection add`/`remove`, silent pagination truncation, name-length validation) are not duplicated here.

---

### HIGH

**1. Concurrent `connection add` race on the same pending toolbox → silent tool loss**
`toolbox_connection_add.go` (pending-promotion path)
Two concurrent `azd ai toolbox connection add <name> <connA>` and `<name> <connB>` against the same pending toolbox can both observe `pending != nil`, both POST `v1` with different tool sets, and then both call `store.Clear()`. The second POST wins and silently drops the first connection's tools. This is adjacent to (but distinct from) the non-atomic version-promotion issue already flagged — that one is about the add/remove sequence on an already-promoted toolbox; this one is about the *initial* promotion of a pending record.
**Fix:** Add a `claimed` / generation marker to pending records; refuse promotion if the marker changed between read and clear, or use optimistic locking via the pending-store backing layer.

**2. `store.Clear()` failure after a successful POST is non-fatal → state corruption + duplicate versions on retry**
`toolbox_connection_add.go` (~line 1420)
After a successful `CreateToolboxVersion()`, if `store.Clear()` fails the error is logged and the command reports success. On the next `connection add`, the stale pending record is re-read, the pending-promotion path runs again, and the code POSTs another `v1` (or `v2` if the server auto-increments) — silently creating an orphaned version.
**Fix:** Treat `Clear()` failure as fatal *after* a successful POST and surface an actionable recovery message (e.g., "local pending record was not cleared; run `azd config delete extensions.ai-toolboxes.pending-toolboxes.<hash>.items.<name>`"). Optionally pre-warn before the POST so the user understands the failure mode.

**3. Corrupt pending-store JSON silently degrades to "no pending records"**
`pending_toolboxes.go:78-80`, surfaced via `toolbox_list.go:66-70` and the delete path
When `GetUserJSON()` fails on a malformed pending store, callers log with `log.Printf` and proceed with an empty bucket. A user with corrupted local state sees an empty `list` with no visible signal — they cannot distinguish "no pending toolboxes" from "your local store is unreadable, your unpublished work is invisible." Pending records are the source of truth for unpublished toolboxes; silent data loss here is risky.
**Fix:** Two-tier handling — missing file → empty bucket (current behavior, correct); parse error → return `exterrors.Internal(CodePendingToolboxStoreFailed, ...)`. Essential paths (create/connection add) fail loud; `list` can degrade with a visible warning row.

---

### MEDIUM

**4. Credential factory does not propagate a user tenant ID → multi-tenant / guest-user auth failures**
`internal/foundry/credential.go` — `NewCredential()`
`azidentity.NewAzureDeveloperCLICredential()` is constructed with no tenant param, defaulting to the CLI's logged-in tenant. For guest users or multi-tenant scenarios where the resource tenant differs from the user's home tenant this produces opaque auth failures. The extension AGENTS.md `UserTenantId` guidance is written for the `PromptSubscription` flow, which this code does not use — but the same multi-tenant concern applies.
**Fix:** Plumb a user tenant ID into `NewCredential` and set it on `AzureDeveloperCLICredentialOptions.TenantID`; resolve it from project/connection context where possible.

**5. JSON output field naming is inconsistent (snake_case vs. camelCase)**
- `toolbox_list.go`: `default_version`, `pending` (snake_case)
- `toolbox_connection_add.go`: `toolboxName`, `connectionId`, `promotedFromPending` (camelCase)
- `toolbox_create.go`: `alreadyExists` (camelCase), nested under snake_case fields

This breaks programmatic parsing of `--output json` and is inconsistent with the rest of azd's snake_case output convention.
**Fix:** Standardize on snake_case throughout (`connection_id`, `promoted_from_pending`, `already_exists`).

**6. `connection remove` has no confirmation prompt; asymmetric with `toolbox delete`**
`toolbox_connection_remove.go`
`toolbox delete` prompts (gated by `--no-prompt`/`--force`); `connection remove` proceeds directly even though it publishes a new version with reduced functionality. Same destructive impact, no parity.
**Fix:** Add a confirmation prompt unless `--force` is passed, mirroring `delete`.

**7. `SetDefaultVersion` failure message omits the just-created version**
`toolbox_connection_add.go` (SetDefaultVersion error wrap)
When `CreateToolboxVersion()` succeeds but `SetDefaultVersion()` fails, the error tells the user to run `toolbox update --default-version <v>` to recover — without telling them which version was just created. Recovery requires a second `list` round-trip.
**Fix:** Include `created.Version` in the error message text.

**8. Documentation drift vs. repo standards**
Per the root `AGENTS.md` documentation maintenance rules: new extension capability + new concepts + any new env vars must be reflected in `docs/`.
- `docs/concepts/glossary.md` — no entries for `toolbox`, `pending toolbox`, `toolbox connection`
- `docs/reference/feature-status.md` — `azd ai toolbox` verbs not listed with a maturity stage
- `docs/reference/environment-variables.md` — if `FOUNDRY_PROJECT_ENDPOINT` (referenced in code) is user-facing, it needs an entry

**Fix:** Add the glossary entries, list the new commands with a stage (Alpha is the natural choice), and document any user-facing env vars.

---

### LOW

**9. Modernize `errors.As` → `errors.AsType` (Go 1.26)**
`internal/exterrors/errors.go` (~line 3562)
Replace `var respErr *azcore.ResponseError; if errors.As(err, &respErr)` with `if respErr, ok := errors.AsType[*azcore.ResponseError](err); ok`. Aligns with the repo's Go 1.26 modern-patterns guidance.

**10. Pagination cap (`maxPaginationPages`) has no diagnostic when hit**
`foundry_toolsets_client.go` — `listPagedFromClient` loop
Related to the silent-truncation issue already flagged, but specifically: when the loop exits because `i >= maxPaginationPages` and the server still reports `has_more=true`, there is no log, no warning, no error — purely an observability gap.
**Fix:** Log a warning (or return a structured warning) including endpoint and page count when the cap is hit.

**11. Endpoint hash truncated to 64 bits → collision risk for multi-endpoint users**
`pending_toolboxes.go` — `endpointBucketKey()`
`sha256.Sum256(...)[: 8]` keeps only 64 bits. Acceptable for keying in small workspaces, but for users with many Foundry endpoints or shared CI machines, birthday-style collisions are not negligible.
**Fix:** Use the first 16 bytes (128 bits / 32-char hex). Schema-versioned if migration needed.

---

### Description alignment

The PR description accurately reflects the implementation (6 verbs, pending-store flow, credential cascade, paginated data-plane client). No material drift.

cli/azd/extensions/azure.ai.agents/internal/cmd/toolbox_create.go
trangevi: Based on the other comment and response from John, Let's plan to have;
- A file input for create, supporting both json and yaml files. Remove the "pending toolbox" handling so that the create experience matches all of our other create commands.
- `connection add` can then take the parameters we already have, to add a single new tool, or a file which can contain multiple tools to add. If multiple tools are provided via file, we should make sure we only increment the toolbox version a single time, not a new version per new tool

cli/azd/extensions/azure.ai.agents/internal/cmd/toolbox_connection.go
lindazqli: we also need RemoteA2A, GroundingWithCustomSearch, need to confirm the type for BrowserTool

hund030: Hi @lindazqli , adding three new categories with their tool-entry shapes is straightforward on the CLI side, but the actual tools[].type body that the Foundry service accepts is the unknown.

If you can share the contract, I'll extend buildToolEntry and the supported-category list in one follow-up.

cli/azd/extensions/azure.ai.agents/internal/cmd/toolbox_connection.go
hund030: Updated. Now it accepts a json or yaml file. Added the file shape in help text.

cli/azd/extensions/azure.ai.agents/internal/cmd/toolbox_create.go
hund030: Updated. Now it accepts a json or yaml file. Added the file shape in help text.

cli/azd/extensions/azure.ai.toolboxes/internal/cmd/toolbox_show.go
hund030: Updated.

hund030:
@wbreza thanks for the review. 

Fixed:
- MED 5 (snake_case JSON)
- MED 6 (connection remove confirmation + --force)
- MED 7 (SetDefaultVersion error now names the created version + recovery command)

Obsolete:  HIGH 1, HIGH 2, HIGH 3, LOW 11. There's no client-side store anymore, so the race / clear-failure / corrupt-JSON / hash-collision concerns no longer apply.

Other issues seem out of the scope for this PR.

therealjohn (approved):

trangevi (changes requested):
Can you please confirm any necessary changes to these files for the extension as well:

Image

I haven't gotten to go through and validate all of those for each of the extensions, and I believe they likely should be updated for consistency. This PR has a sample that I took that image from: https://github.com/Azure/azure-dev/pull/8264

hund030:
@trangevi Plus .golangci.yaml and AGENTS.md. 
`azure.ai.toolboxes` doesn't bundle any third-party assets, so no need to add .gitattributes.

cli/azd/extensions/azure.ai.agents/internal/cmd/toolbox_connection.go
hund030: File an issue to track https://github.com/Azure/azure-dev/issues/8297

therealjohn (approved):

trangevi (approved):

trangevi:
@copilot update the codeowners file for this extension to match azure.ai.agents

trangevi:

therealjohn (approved):

trangevi:

tg-msft (approved):

therealjohn (approved):

PR #8204, opened 2026-05-15
doc: add design spec for `azd ai skill` commands

### LINKED ISSUE #8142: Add `azd ai skill create/update/show/list/download/delete` to manage Foundry Skills from any directory
### Feature Request

Please add `azd ai skill` commands to the agent extension that provide full CRUD on Skills in a Foundry Project, following the design specifications in [foundrysdk_specs PR #165](https://github.com/coreai-microsoft/foundrysdk_specs/pull/165). This issue is part of the umbrella consolidation effort tracked in #8135 and is a sibling to the connection work in #8139 and project-context work in #8124.

#### Design Highlights
- **Command surface:** `azd ai skill create | update | delete | show | list | download`
- **Three mutually exclusive create modes:**
  1. Inline: `--description <...> --instructions <...>` -> `POST /skills` (JSON body)
  2. SKILL.md file: `--file ./SKILL.md` (CLI parses YAML front matter + Markdown body) -> `POST /skills` (JSON body)
  3. Gzip package: `--file ./skill.tar.gz` (CLI uploads as-is) -> `POST /skills:import` (`application/gzip` body)
  - The CLI errors out (rather than silently merging) if more than one mode is supplied.
- **Update:** description, instructions, or full body via inline flags or `--file`.
- **Show:** metadata only -- body is fetched via `download`.
- **List:** paginated (`--top`, `--orderby`).
- **Download:**
  - `GET /skills/{name}:download` always returns `application/gzip`.
  - Default: CLI extracts the gzip into `--output-dir` (default `./.agents/skills/<skill-name>/`), so the caller ends up with `SKILL.md` and any sibling assets on disk.
  - `--raw`: writes the unmodified gzip archive to the output directory.
- **Delete:** confirmation by default, `--force` to skip.
- **`--file` is not a manifest:** read at invocation time only; the CLI does not track or re-read it.
- **Endpoint resolution:** Inherits the standard 5-level resolution order from the umbrella surface.
- **Cross-cutting flags:** `--output json`, `--no-prompt`, `--debug`, `-p / --project-endpoint`, `--force` (delete confirmation skip).

#### API dependencies
Per the [Skills TypeSpec](https://github.com/lindazqli/azure-rest-api-specs/blob/157e6fb712649fa0373a0ad7efe53cdb6ca1dc0b/specification/ai-foundry/data-plane/Foundry/src/skills/routes.tsp), the data-plane surface is already defined:
- `POST /skills` (JSON body)
- `POST /skills:import` (`application/gzip` body)
- `GET /skills/{name}` (show)
- `POST /skills/{name}` (update)
- `DELETE /skills/{name}`
- `GET /skills` (list, paginated)
- `GET /skills/{name}:download` (`application/gzip`)

No new server changes required -- this is CLI integration work against an existing TypeSpec surface.

---
Please refer to the [full design spec (foundrysdk_specs PR #165)](https://github.com/coreai-microsoft/foundrysdk_specs/pull/165) -- specifically the *Skills* sub-section under Requirements, the API Surface table (rows 7-8), and the Hero Code Sample "Skill round-trip" scenario -- for implementation guidance.

#### Additional Context
- Server-side TypeSpec is ready (status "Confirmed" in the spec's Dependencies table); this is CLI integration work only.

huimiu:
## Summary

Adds the design spec for the Foundry Skill commands that ship in the `azure.ai.agents` extension (`cli/azd/docs/design/ai-skill-design-spec.md`). The `skill` subtree provides full CRUD plus `download` on Foundry Skills, so a developer can create a skill from inline flags or a `SKILL.md`/`*.tar.gz` file and round-trip it back out as a folder on disk, all without standing up a Foundry SDK harness.

Related issues:

- Feature: Azure/azure-dev#8142

## Key points

- **Placement**: lives in the existing `azure.ai.agents` extension. Today the commands surface as `azd ai agent skill …`; the layout is designed so a future move to a standalone `azd ai skill` extension is registration-only with no behavior diff. No `registry.json` impact and no new persistent state.
- **v1 surface**: `create`, `update`, `show`, `list`, `download`, `delete`. `create` has three mutually exclusive input modes (inline `--description` + `--instructions`, `--file SKILL.md`, or `--file *.tar.gz`); `download` extracts the server gzip into `./.agents/skills/<name>/` by default, with `--raw` to keep the archive and `--force` to overwrite; `delete` confirms by default with `--force` to skip.
- **Endpoint resolution**: reuses the standard 5 level cascade and cross-cutting flags (`-p`, `--output`, `--no-prompt`, `--debug`), matching `connection`, `toolbox`, and `routine`, and works standalone outside an azd project.
- **Create vs. update**: distinct verbs hide the upsert wire. `create` fails on a name collision (or replaces with `--force` via delete-then-create). `update` is GET + local merge + POST that preserves untouched fields; `--file` rules on `update` follow the same modes as `create`, with packaged gzip update gated on the single open question below.
- **Typed client**: new package at `internal/pkg/agents/skill_api/` (sibling of `agent_api`, no overload), so error shapes, headers, and pagination stay consistent with the rest of the AI surface.
- **Safe tar extraction**: zip-slip guard, symlink/hard-link rejection, staging directory, 10k entries / 512 MB decompression cap. JSON output shapes for `download` and `delete` are part of the public contract; resource verbs are passthrough.
- **Errors and telemetry**: new `exterrors` codes (`CodeInvalidSkillName`, `CodeInvalidSkillFile`, `CodeSkillArchiveUnsafe`, `CodeSkillOutputCollision`) plug into the existing `exterrors.Validation` factory; one telemetry event per verb, with no skill names, descriptions, file paths, or endpoint values emitted.

### DIFF
--- a/cli/azd/docs/design/ai-skill-design-spec.md
+++ b/cli/azd/docs/design/ai-skill-design-spec.md
@@ -339,7 +339,9 @@ One event per command, reusing the extension's existing telemetry surface:
 
 - `azd.ai.agent.skill.create` (`mode`: `inline` / `file-md` / `file-gzip`,
   `forced`: bool).
-- `azd.ai.agent.skill.update` (`mode`: as above, `fieldsTouched`: count).
+- `azd.ai.agent.skill.update` (`mode`: `inline` / `file-md`,
+  `fieldsTouched`: count). Gzip updates are rejected at the flag layer
+  (§6.2), so `file-gzip` is not a valid emitted value here.
 - `azd.ai.agent.skill.show` / `azd.ai.agent.skill.list` (`resolvedSource`).
 - `azd.ai.agent.skill.download` (`raw`: bool, `forced`: bool, `extractedFileCount`).
 - `azd.ai.agent.skill.delete` (`forced`: bool, `cancelled`: bool).
@@ -353,7 +355,15 @@ debugging, are hashed.
 - **Tar extraction.** Full rejection rules and decompression limits in §6.5.
 - **File write permissions.** User's umask; executable bits dropped.
 - **Auth.** Reuses the existing bearer-token pipeline; no new secret writes.
-- **Argument echoing.** Debug logger sanitizes request bodies (same as `agent_api`).
+- **Argument echoing.** `agent_api` sets `IncludeBody: true` on the Azure SDK
+  logging policy, and the current `setupDebugLogging` sanitizer only redacts
+  JSON connection-string fields. Skill request bodies carry user-authored
+  `description` and `instructions`, so this work extends the sanitizer with
+  JSON-field redaction rules for those fields (and any new free-text fields
+  the skill service exposes) before the skill client participates in
+  `IncludeBody` logging under `--debug`. Until the extended sanitizer is in
+  place, the skill client opts out of body logging so descriptions and
+  instructions are never written to `azd-ai-agents-<date>.log`.
 
 ## 12. Reference: Command Summary
 
--- a/cli/azd/docs/design/ai-skill-design-spec.md
+++ b/cli/azd/docs/design/ai-skill-design-spec.md
@@ -137,7 +137,7 @@ Mode selection logic:
 - If the front matter `name` is present and differs from the positional
   `<name>`, the positional wins and a one-line warning is printed to stderr.
   Suppressed under `--no-prompt` or `--output json`.
-- If the front matter is missing or unparseable, the command fails with a
+- If the front matter is missing or unparsable, the command fails with a
   validation error that points to the problematic line.
 
 `--force` semantics on create:
--- a/cli/azd/docs/design/ai-skill-design-spec.md
+++ b/cli/azd/docs/design/ai-skill-design-spec.md
@@ -1,18 +1,18 @@
 <!-- cspell:ignore exterrors foundry gzip tarball zipslip orderby foundrysdk -->
 
-# Design Spec: `azd ai agent skill` Commands
+# Design Spec: `azd ai skill` Commands
 
 ## 1. Summary
 
-This spec covers the Foundry Skill commands that ship in the `azure.ai.agents`
-extension:
+This spec covers the Foundry Skill commands that ship in the new
+`azure.ai.skills` extension:
 
-- `azd ai agent skill create <name>` (three mutually exclusive input modes).
-- `azd ai agent skill update <name>` (in-place metadata or body update).
-- `azd ai agent skill show <name>` (metadata only).
-- `azd ai agent skill list` (paginated).
-- `azd ai agent skill download <name>` (default extracts the server gzip; `--raw` keeps it).
-- `azd ai agent skill delete <name>` (confirmation by default; `--force` to skip).
+- `azd ai skill create <name>` (three mutually exclusive input modes).
+- `azd ai skill update <name>` (in-place metadata or body update).
+- `azd ai skill show <name>` (metadata only).
+- `azd ai skill list` (paginated).
+- `azd ai skill download <name>` (default extracts the server gzip; `--raw` keeps it).
+- `azd ai skill delete <name>` (confirmation by default; `--force` to skip).
 
 These commands are pure CLI integration on top of the existing Foundry Skills
 data plane. No new server work is required.
@@ -41,16 +41,16 @@ Out of scope:
 
 ## 3. Extension Placement and Surface
 
-The skill subtree lives inside the existing `azure.ai.agents` extension (no
-new module, no change to `registry.json`). Commands surface as
-`azd ai agent skill <verb>`. The layout is designed so that a future move to a
-standalone `azd ai skill` extension is registration-only with no behavior diff.
+Skill commands ship in a new, standalone `azure.ai.skills` extension
+(namespace `ai.skill`), separate from `azure.ai.agents`. Commands surface as
+`azd ai skill <verb>`. Publishing the extension adds one new entry to
+`registry.json`; no changes to `azure.ai.agents` are required.
 
 Internally, files land under
-`cli/azd/extensions/azure.ai.agents/internal/cmd/skill_*.go`, mirroring the
-`project_*.go` layout. A new typed client lives at
-`internal/pkg/agents/skill_api/` so the agent client is not overloaded with
-skill operations.
+`cli/azd/extensions/azure.ai.skills/internal/cmd/skill_*.go`. A typed client
+lives at `cli/azd/extensions/azure.ai.skills/internal/pkg/skill_api/`. The
+extension owns no persistent state of its own; project context resolution is
+covered in §4.
 
 ## 4. Endpoint Resolution and Cross-Cutting Flags
 
@@ -59,8 +59,11 @@ All skill commands resolve the Foundry project endpoint via the standard
 
 1. `-p` / `--project-endpoint` flag on the invoked command.
 2. Active azd env value (`AZURE_AI_PROJECT_ENDPOINT`).
-3. Global config under `extensions.ai-agents.project.context.endpoint`
-   in `~/.azd/config.json`.
+3. Global config under `extensions.ai-skills.project.context.endpoint`
+   in `~/.azd/config.json`, falling back to
+   `extensions.ai-agents.project.context.endpoint` when the skills-owned key
+   is unset (so users who configured the endpoint via `azure.ai.agents` are
+   not forced to re-run `set`).
 4. Host environment variable `FOUNDRY_PROJECT_ENDPOINT`.
 5. Structured error with an actionable suggestion.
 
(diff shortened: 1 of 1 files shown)


cli/azd/docs/design/ai-skill-design-spec.md
```diff
@@ -0,0 +1,395 @@
...
+Out of scope:
+
+- Any change to the Foundry Skill REST surface.
+- Wiring a skill into an agent build. `download` writes files to disk; what
+  the developer does with them is their responsibility.
+- Tracking `--file` as a manifest. The file is read at invocation time only.
+
+## 3. Extension Placement and Surface
+
+The skill subtree lives inside the existing `azure.ai.agents` extension (no
```
therealjohn: This has changed - it will be a unique extension, azure.ai.skills.

cli/azd/docs/design/ai-skill-design-spec.md
```diff
@@ -0,0 +1,395 @@
...
+| Delete | DELETE | `/skills/{name}` | |
+| Download | GET | `/skills/{name}:download` | Returns `application/gzip` |
+
+Auth uses the same bearer-token policy the existing agent client uses
+(scope `https://ai.azure.com/.default`). The User-Agent header carries the
+extension version, consistent with `agent_api`.
+
+## 6. Command Behavior
+
+### 6.1 `azd ai agent skill create <name>`
```
therealjohn: All commands would start with `azd ai skill` since we are moving this to a unique extension.

therealjohn (approved):
Left two comments just clarify that it should be in a separate extension which impacts the skill name in the spec.

cli/azd/docs/design/ai-skill-design-spec.md
```diff
@@ -0,0 +1,395 @@
...
+3. `--force`: skip the prompt regardless of TTY state.
+4. Non-existent skill: 404 surfaced as a "not found" error.
+
+### 6.7 Skill Name Validation
+
+The CLI validates `<name>` on every command that takes it:
+
+- Non-empty after trim.
+- Matches the service-documented regex, or the conservative fallback
+  `^[a-zA-Z][a-zA-Z0-9-_]{0,62}$`. The service makes the final decision.
```
jongio: **[MEDIUM]** The fallback regex `^[a-zA-Z][a-zA-Z0-9-_]{0,62}$` diverges from the agent name pattern in `agent_yaml/parse.go` (`^[a-zA-Z0-9]([a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?$`): allows underscores, requires a leading letter, and permits trailing special chars. It's labeled 'conservative' but is actually more permissive. Is this server-mandated for skills, or should it align with the existing pattern?

jongio (commented):
Left a question on the name validation regex - the fallback pattern in 6.7 diverges from the existing agent name regex in a few ways.

huimiu:
Thanks for the reviews! Pushed new commit:

@therealjohn: moved skill commands into a new standalone azure.ai.skills extension (namespace ai.skill). All commands renamed from azd ai agent skill to azd ai skill, with matching updates to file paths, telemetry event names (azd.ai.skill.*), and the debug log name.

@jongio: aligned the fallback name regex with the existing agent name pattern: `^[a-zA-Z0-9]([a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?$`. Good catch on the underscore/trailing-special drift.

jongio (commented):
Addresses my previous feedback on the name validation regex, now aligned with the agent name pattern in `agent_yaml/parse.go`. The config fallback to `extensions.ai-agents.project.context.endpoint` is a nice backward-compat touch. Looks good to me. Let's get a sign-off from @JeffreyCA or @vhvb1989 to unblock.

wbreza (commented):
Drive-by review focused on spec ↔ implementation alignment. Approved for what it set out to describe at the time, but since this PR was last touched (5/18), PR #8224 underwent a major refactor (5/26) to a versioned/immutable Skills API. The spec and what shipped have meaningfully diverged. Findings below are all about that drift — none of these block on their own, but as a whole the spec is no longer a faithful description of the `azd ai skill` surface area.

If the goal is "merge the spec as a historical v0 then sync in a follow-up," that's fine and these can become a tracking issue. If the goal is "this is the canonical reference for the extension," the §5 / §6 / §10 sections need a substantive update before merge.

### Critical — spec ↔ implementation drift

1. **§5 Data-Plane Surface — API endpoints don't match shipped code.** Spec lists `POST /skills`, `POST /skills:import`, `GET /skills/{name}:download`. Implementation uses `POST /skills/{name}/versions` (multipart for ZIP, JSON for inline), `GET /skills/{name}/content`, and `GET /skills/{name}/versions/{version}/content`. The whole versioned-immutable model with a `default_version` pointer is missing from the spec.

2. **§§6.1 / 6.2 / 6.5 — gzip vs ZIP throughout.** Spec describes `.tar.gz` / `.tgz` archives and `application/gzip` content types in multiple places. The shipping implementation only accepts `.zip` / `application/zip` and actively rejects gzip in `update`. Every gzip/tar mention in those sections needs to flip to zip.

3. **§6.7 Skill Name Validation — regex still allows uppercase.** Spec shows `^[a-zA-Z0-9]([a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?$`. Implementation enforces `^[a-z0-9]([a-z0-9\-]*[a-z0-9])?$` (lowercase only, 1–64 chars) per agentskills.io. You confirmed the implementation fix in the 5/18 comment to jongio, but the spec file itself wasn't updated to match. Also update the error message wording from "alphanumeric with internal hyphens" to "lowercase alphanumeric with internal hyphens."

4. **§6.2 update — missing `--set-default-version` and the versioning model.** Spec describes update as GET-current → merge omitted fields → POST merged payload (a single-blob mental model). Implementation has two distinct update paths: (a) inline / `.md` modes create a new immutable version and set it as default via `POST /skills/{name}/versions`; (b) `--set-default-version <ver>` repoints `default_version` to an existing immutable version via `POST /skills/{name}` without uploading new content. The `--set-default-version` flag isn't in the spec's flag table at all.

### High

5. **§6.5 download — missing `--version` flag.** Implementation supports `--version <ver>` to target `GET /skills/{name}/versions/{ver}/content`. Public flag, should be in the spec.

6. **§5 Data-Plane Surface — missing preview header.** All requests include `Foundry-Features: Skills=V1Preview`. The spec doesn't say the API is preview-gated, which is material for anyone integrating against it.

7. **§6.5 extraction safety — describes tar semantics for what's actually ZIP.** "Walk the tar entries", "symlinks", "hard links", and "decompression bounded at 10,000 entries / 512 MB uncompressed" are tar/gzip-specific concepts. ZIP needs different language — reject absolute paths, reject any `..` components (zip-slip guard), reject symlinks and non-regular entries, cap at 10,000 entries / 512 MB total uncompressed size.

### Medium

8. **§6.5 — `--raw` output filename.** Spec says `<output-dir>/<name>.tar.gz`; implementation writes `<output-dir>/<name>.zip`.

9. **§7 Output Contracts — download JSON shape.** Implementation's `downloadResult` includes a `version` field (present when `--version` is passed). Archive filename example in the JSON still shows `.tar.gz`; should be `.zip`.

10. **§10 Telemetry — mode names and missing event.** `file-gzip` should be `file-zip` (or whatever maps to the ZIP-only reality). Also: `--set-default-version` is a new public operation with no telemetry coverage described in §10.

11. **§6.2 — clarify `--force` is create-only.** Currently described only under §6.1; readers may assume `update` supports it too. Worth one sentence in §6.2: "`--force` is not available on `update`; to replace a skill entirely, use `create --force`."

12. **§11 Security Considerations — tar decompression language.** Bullet 1 talks about decompression limits in tar/gzip terms. Should reflect ZIP-specific extraction safety (point back to the updated §6.5).

### Low

13. **§4 Endpoint Resolution — clarify ai-agents fallback is read-only.** Worth one sentence noting that this extension reads the `extensions.ai-agents.project.context.endpoint` for backwards compatibility but never writes to it, so `azd ai skill` operations don't mutate the agents config.

### Already resolved — no action needed

- Copilot's telemetry inconsistency (`file-gzip` appearing in update modes) is consistent with the spec's note that gzip is rejected at the flag layer on update.
- Copilot's security/logging note is resolved — implementation sets `IncludeBody: false` and §11 reflects that.

Happy to chat on any of these or file a follow-up issue if you'd rather merge as-is and sync later.

kristenwomack:
Hi @huimiu,

Thanks for the spec.

Two things before I'm ready to approve, and one follow-up.

First, @wbreza flagged that this doc was last touched 5/18, but #8224 had a meaningful refactor on 5/26. Could you do a sync pass and confirm the spec still matches what's actually shipping? I don't want to lock a contract that has already drifted from the code.

Second, §7 commits the JSON shapes for all six verbs as the public contract. I want to be intentional about that for `download` (the `files` vs. `archive` + `raw` split) and `delete` (the cancelled shape). Are we comfortable freezing those at v1, or would you prefer a "preview, subject to change" note on the JSON shapes until we have a couple of integrations using them?

Follow-up, not blocking: §11 keeps `--debug` body logging off on the new `skill_api` client until we ship the description/instructions sanitizer. Could you open a tracking issue so we don't lose it?

Thanks, Kristen

kristenwomack:
Hey @huimiu, nice spec. A few things:

1. **Three `create` input modes** (inline / `SKILL.md` / `*.tar.gz`) — curious why all three for v1? I'm wondering if inline ends up being the odd one out: `--description` + `--instructions` could get unwieldy fast, and the file path feels like the mode we'd actually want people reaching for. What's the use case you're picturing for inline — quick prototyping, or something else? Open to keeping it if there's a real pull for it.
2. **`create --force` = delete-then-create** — this is destructive (loses identity, any run history). I'd rather it just fail and tell the user to run `delete` first. `--force` reads like "overwrite in place" and that's not what it does.
3. **`.agents/skills/<name>/`** as the default download path — are we OK with `.agents/` becoming a convention across the extension? If yes, let's call that out somewhere central so we don't end up with `.agents/`, `.azd/`, and `.foundry/` all in the same repo.
4. **Open question on packaged-gzip update** — let's resolve this in the PR before merge so the implementation doesn't have to guess.

PR #8210, opened 2026-05-15
Create folder during init

### LINKED ISSUE #8209: azd ai agent init - Create a directory when init'ing from a template
# Problem
When a user does the `azd ai agent init` command, it creates a yaml in the dir. If this is done in something like the users home folder, it is pretty difficult to recover from.

Here is a video clip of this happening: https://app.heymarvin.com/note/9c9dcead-bf57-45db-9242-32d2e781bda0/

# Proposed Solution
Mimic git clone behavior and create a folder. This requires `cd` into the directory, but it is easier to recover from.

Here is a video of someone using it in this way: https://app.heymarvin.com/note/a3d9dcb5-0b75-4eb8-a990-e10717797b4f/

banrahan:
Related to #8209, example implementation for changing `azd ai agent init` to create a new dir when init'ing from a sample.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -65,7 +65,6 @@ cli/azd/extensions/azure.coding-agent/azurecodingagent
 cli/azd/extensions/azure.coding-agent/azurecodingagent.exe
 cli/azd/extensions/azure.ai.agents/azureaiagent
 cli/azd/extensions/azure.ai.agents/azureaiagent.exe
-cli/azd/extensions/azure.ai.agents/azure.ai.agents
 cli/azd/extensions/azure.ai.finetune/azure.ai.finetune
 cli/azd/extensions/azure.ai.finetune/azure.ai.finetune.exe
 cli/azd/extensions/microsoft.azd.extensions/microsoft.azd.extensions
--- a/.gitignore
+++ b/.gitignore
@@ -65,6 +65,7 @@ cli/azd/extensions/azure.coding-agent/azurecodingagent
 cli/azd/extensions/azure.coding-agent/azurecodingagent.exe
 cli/azd/extensions/azure.ai.agents/azureaiagent
 cli/azd/extensions/azure.ai.agents/azureaiagent.exe
+cli/azd/extensions/azure.ai.agents/azure.ai.agents
 cli/azd/extensions/azure.ai.finetune/azure.ai.finetune
 cli/azd/extensions/azure.ai.finetune/azure.ai.finetune.exe
 cli/azd/extensions/microsoft.azd.extensions/microsoft.azd.extensions
--- a/cli/azd/extensions/azure.ai.agents/internal/cmd/init.go
+++ b/cli/azd/extensions/azure.ai.agents/internal/cmd/init.go
@@ -807,6 +807,9 @@ from code-deploy ZIP packaging (uses .gitignore syntax).`,
 						// Full azd template - dispatch azd init -t <repo>
 						// Create project in a new subdirectory derived from the template title.
 						folderName := sanitizeAgentName(selectedTemplate.Title)
+						// Check whether the target directory already exists so we
+						// only report "created" when a new directory was made.
+						_, dirExisted := os.Stat(folderName)
 						initArgs := []string{"init", "-t", selectedTemplate.Source, folderName}
 						if flags.env != "" {
 							initArgs = append(initArgs, "--environment", flags.env)
@@ -853,7 +856,9 @@ from code-deploy ZIP packaging (uses .gitignore syntax).`,
 								folderName, err,
 							)
 						}
-						createdFolder = filepath.Join(originalCwd, folderName)
+						if dirExisted != nil {
+							createdFolder = filepath.Join(originalCwd, folderName)
+						}
 
 						// Search for an agent manifest in the scaffolded project
 						cwd, err := os.Getwd()
--- a/cli/azd/extensions/azure.ai.agents/internal/cmd/init.go
+++ b/cli/azd/extensions/azure.ai.agents/internal/cmd/init.go
@@ -1166,11 +1166,8 @@ func ensureProject(
 					envBase = filepath.Base(cwd)
 				}
 			}
-			sanitizedDirectoryName := sanitizeAgentName(envBase)
-			initArgs = append(
-				sanitizedEnvName := sanitizeAgentName(envBase + "-dev")
-				initArgs, "--environment", sanitizedEnvName,
-			)
+			sanitizedEnvName := sanitizeAgentName(envBase + "-dev")
+			initArgs = append(initArgs, "--environment", sanitizedEnvName)
 		}
 
 		// We don't have a project yet
--- a/cli/azd/extensions/azure.ai.agents/internal/cmd/init.go
+++ b/cli/azd/extensions/azure.ai.agents/internal/cmd/init.go
@@ -10,6 +10,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"io/fs"
 	"log"
 	"maps"
 	"net/http"
@@ -812,7 +813,8 @@ from code-deploy ZIP packaging (uses .gitignore syntax).`,
 						folderName := folderNameFromTitle(selectedTemplate.Title)
 						// Check whether the target directory already exists so we
 						// only report "created" when a new directory was made.
-						_, dirExisted := os.Stat(folderName)
+						_, statErr := os.Stat(folderName)
+						newlyCreated := errors.Is(statErr, fs.ErrNotExist)
 						initArgs := []string{"init", "-t", selectedTemplate.Source, folderName}
 						if flags.env != "" {
 							initArgs = append(initArgs, "--environment", flags.env)
@@ -863,7 +865,7 @@ from code-deploy ZIP packaging (uses .gitignore syntax).`,
 								folderName, err,
 							)
 						}
-						if dirExisted != nil {
+						if newlyCreated {
 							createdFolder = filepath.Join(originalCwd, folderName)
 							createdFromTitle = selectedTemplate.Title
 						}
@@ -897,7 +899,8 @@ from code-deploy ZIP packaging (uses .gitignore syntax).`,
 						folderName := folderNameFromTitle(selectedTemplate.Title)
 						// Check whether the target directory already exists so we
 						// only report "created" when a new directory was made.
-						_, dirExisted := os.Stat(folderName)
+						_, statErr := os.Stat(folderName)
+						newlyCreated := errors.Is(statErr, fs.ErrNotExist)
 						flags.manifestPointer = selectedTemplate.Source
 						if err := runInitFromManifest(
 							ctx, flags, azdClient, httpClient, folderName,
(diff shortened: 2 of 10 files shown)


cli/azd/extensions/azure.ai.agents/internal/cmd/init.go
```diff
@@ -472,14 +489,23 @@ func newInitCommand(extCtx *azdext.ExtensionContext) *cobra.Command {
...
 
 					default:
-						// Agent manifest template - use existing -m flow
+						// Agent manifest template - use existing -m flow.
+						// Create project in a new subdirectory derived from the template title.
+						title := selectedTemplate.Title
+						if idx := strings.IndexByte(title, '('); idx >= 0 {
+							title = strings.TrimSpace(title[:idx])
+						}
+						folderName := sanitizeAgentName(title)
```
jongio: **[MEDIUM]** The title-to-folder-name derivation (strip parenthetical + sanitize) is copy-pasted here and in the TemplateTypeAzd case (~line 416). Pull it into a small helper so the logic lives in one place:

```suggestion
folderName := folderNameFromTitle(selectedTemplate.Title)
```

Then add near `sanitizeAgentName` in `init_from_code.go`:

```go
func folderNameFromTitle(title string) string {
if idx := strings.IndexByte(title, '('); idx >= 0 {
title = strings.TrimSpace(title[:idx])
}
return sanitizeAgentName(title)
}
```

Update the TemplateTypeAzd case (~line 416-420) the same way.

jongio (commented):
Issues to address:
- init.go:494 - folder name derivation is duplicated between the TemplateTypeAzd and default cases; extract a helper

cli/azd/extensions/azure.ai.agents/internal/cmd/init.go
```diff
@@ -407,17 +412,19 @@ func newInitCommand(extCtx *azdext.ExtensionContext) *cobra.Command {
...
 						} else {
-							cwd, err := os.Getwd()
-							if err == nil {
-								sanitizedDirectoryName := sanitizeAgentName(filepath.Base(cwd))
-								initArgs = append(
-									initArgs, "--environment", sanitizedDirectoryName+"-dev",
-								)
-							}
+							initArgs = append(
+								initArgs, "--environment", folderName+"-dev",
```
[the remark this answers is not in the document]
banrahan: Made this change

cli/azd/extensions/azure.ai.agents/internal/cmd/init.go
banrahan: Make these changes

wbreza (commented):
Thanks for tackling this — moving `azd ai agent init` to a `git clone`-style folder layout is the right UX. A few observations beyond the prior reviewer comments:

## High

### 1. `folderNameFromTitle` is only used in one of the two template branches

The new helper addresses jongio''s earlier "extract a helper" feedback, but the `TemplateTypeAzd` branch still calls `sanitizeAgentName(selectedTemplate.Title)` directly while the `default` (manifest) branch calls `folderNameFromTitle(selectedTemplate.Title)`. For a template titled `"Basic Agent (Python)"` the two branches now produce different folders:

- Azd branch → `basic-agent-python/`
- Manifest branch → `basic-agent/`

Suggest routing both branches through `folderNameFromTitle` so the naming is consistent across template types.

### 2. Env-name derivation is inconsistent between branches

- Azd branch: `sanitizeAgentName(selectedTemplate.Title + "-dev")` — derived from the **raw title**
- Manifest branch (via `ensureProject`): `sanitizeAgentName(targetDir + "-dev")` — derived from the **sanitized folder name**

Same template, different env name depending on type. Easiest fix: have both paths derive the env name from the resolved folder name (the manifest branch''s pattern), so folder + env stay paired.

### 3. Non-ASCII titles are silently destroyed

`sanitizeAgentName`''s regex `[^a-z0-9-]+` strips all non-ASCII to hyphens, then trims/truncates and falls back to `"my-agent"` if empty. A template titled `Échantillon` becomes `chantillon`, and a title that is mostly non-ASCII (e.g. `日本語エージェント`) collapses to `my-agent` with no warning. Worth at least surfacing a notice — perhaps including the original template title in the "Your project has been created in …" line — so users can see how their selection was mapped.

## Medium

### 4. `./` prefix is POSIX-style on Windows

```go
fmt.Printf("\nYour project has been created in ./%s\n", relPath)
```

On Windows this prints `./src\my-agent` — neither idiomatic Unix nor Windows. Consider `filepath.ToSlash(relPath)` for display, and either drop the literal `./` or quote the path.

### 5. Message doesn''t tell the user the next step

The follow-up tells the user *where* the project was created but not how to navigate to it. Many CLIs print a literal `cd <folder>` line for copy-paste — would be helpful here.

### 6. Long titles can truncate `-dev` off the env name

`sanitizeAgentName(title + "-dev")` sanitizes the combined string and then truncates to 63 chars. For long titles, `-dev` is silently dropped. Truncating the base to 59 chars before appending `-dev` keeps the suffix intact:

```go
base := sanitizeAgentName(title)
if len(base) > 59 { base = strings.TrimRight(base[:59], "-") }
envName := base + "-dev"
```

### 7. Inverted `dirExisted` semantics

```go
_, dirExisted := os.Stat(folderName)
if dirExisted != nil { /* directory was NEW */ }
```

The variable is an `error`, not a boolean — `dirExisted != nil` actually means "stat failed (directory probably did not exist)." This reads backwards and the same pattern is mirrored in the tests. Two improvements:

1. Rename to `statErr` (or compute a `newlyCreated` bool) for clarity.
2. Use `errors.Is(err, fs.ErrNotExist)` instead of `err != nil` — otherwise a permission error on `os.Stat` would be treated as "newly created" and trigger the "created" message incorrectly.

### 8. Tests assert against re-implemented logic, not production code

The five new `TestCreatedFolderPath_*` tests locally re-execute the same `filepath.Join(originalCwd, folderName)` + `filepath.Rel(...)` computation and then assert the result equals itself. They don''t invoke `newInitCommand`, `ensureProject`, `runInitFromManifest`, `sanitizeAgentName`, or `folderNameFromTitle`. The tests will pass even if the production flow diverges.

Suggest extracting the path-derivation into a small pure helper, e.g.:

```go
func formatCreatedFolderMessage(originalCwd, createdFolder string) string { ... }
```

…and having both the production code and the tests call it. That makes the tests meaningful regression markers.

### 9. `sanitizeAgentName` and `folderNameFromTitle` have no direct unit tests

These two helpers are central to folder + env-name correctness but are only exercised transitively. Worth a small table-driven test for each, covering at minimum:

- Truncation at 63 chars (and the `-dev` suffix loss in #6)
- Empty-after-sanitize fallback to `"my-agent"`
- Non-ASCII input (related to #3)
- Trailing-`(...)` stripping in `folderNameFromTitle`

## Low

### 10. `folderNameFromTitle` is narrowly scoped under a generic name

It only strips a trailing `(...)` suffix. Either rename it more specifically (e.g. `folderNameStrippingParenSuffix`) or add a short doc comment so callers don''t expect broader normalization.

### 11. Help text and discoverability

The cobra command''s `Short`/`Long` text doesn''t mention the new "creates a new folder" behavior. Updating it would make the change discoverable from `--help`.

### 12. No opt-out flag for the prior behavior

Power users who deliberately want to scaffold into the current directory have no shortcut anymore. Not blocking — possible follow-up.

---

Overall the change is the right direction. The biggest things to address before merge are the two consistency issues (#1, #2) and the silent non-ASCII handling (#3); the rest can fold in here or in a follow-up.

cli/azd/extensions/azure.ai.agents/internal/cmd/init.go
```diff
@@ -448,6 +459,16 @@ func newInitCommand(extCtx *azdext.ExtensionContext) *cobra.Command {
...
 
+						// Sync the extension process into the new project directory.
+						// The azd host already chdir'd when it processed the init command.
+						if err := os.Chdir(folderName); err != nil {
+							return fmt.Errorf(
+								"changing to project directory %q: %w",
+								folderName, err,
+							)
+						}
+						createdFolder = filepath.Join(originalCwd, folderName)
```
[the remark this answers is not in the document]
banrahan: resolved in latest commit

cli/azd/extensions/azure.ai.agents/internal/cmd/init.go
```diff
@@ -499,6 +530,15 @@ func newInitCommand(extCtx *azdext.ExtensionContext) *cobra.Command {
 				}
 			}
 
+			if createdFolder != "" {
+				// Print relative to where the user invoked the command.
+				relPath, relErr := filepath.Rel(originalCwd, createdFolder)
+				if relErr != nil {
+					relPath = createdFolder
+				}
+				fmt.Printf("\nYour project has been created in ./%s\n", relPath)
```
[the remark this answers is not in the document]
banrahan: resolved in latest commit

banrahan:
Thanks @wbreza for all of the helpful feedback, I addressed all but 11 and 12. For 11, there are a lot of init behaviors that will not create a folder, so I am not sure what to put in the help text that isn't super long. For 12, we can put that as another issue for later if needed.

cli/azd/extensions/azure.ai.agents/internal/cmd/init.go
```diff
@@ -655,21 +707,37 @@ func (a *InitAction) Run(ctx context.Context) error {
...
+) (*azdext.ProjectConfig, error) {
 	projectResponse, err := azdClient.Project().Get(ctx, &azdext.EmptyRequest{})
 	if err != nil {
 		fmt.Println("Let's get your project initialized.")
 
 		// Environment creation is handled separately in ensureEnvironment
-		initArgs := []string{"init", "-t", "Azure-Samples/azd-ai-starter-basic", "."}
+		initArgs := []string{
+			"init", "-t", "Azure-Samples/azd-ai-starter-basic", targetDir,
+		}
```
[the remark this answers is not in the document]
banrahan: This seems to be fixed already.

cli/azd/extensions/azure.ai.agents/internal/cmd/init_from_code.go
```diff
@@ -869,6 +869,13 @@ func sanitizeAgentName(name string) string {
 	return name
 }
 
+func folderNameStrippingParenSuffix(title string) string {
```
trangevi: NIT: why'd this end up in this file?

.gitignore
```diff
@@ -65,6 +65,7 @@ cli/azd/extensions/azure.coding-agent/azurecodingagent
 cli/azd/extensions/azure.coding-agent/azurecodingagent.exe
 cli/azd/extensions/azure.ai.agents/azureaiagent
 cli/azd/extensions/azure.ai.agents/azureaiagent.exe
+cli/azd/extensions/azure.ai.agents/azure.ai.agents
```
trangevi: I'm assuming this came from some build? do we need a .exe path as well, like the others?

banrahan: I will remove this, I think I was building this custom

cli/azd/extensions/azure.ai.agents/internal/cmd/init_from_code.go
banrahan: Moving it to init.go after talking with travis

trangevi (approved):

PR #8221, opened 2026-05-16
fix: add missing azd exec (#7400) to 1.25.1 changelog + fix skill coverage gap

rajeshkamal5050:
## Problem

The 1.25.1 release changelog (PR #8215) missed PR #7400 (`azd exec`). The PR was in the commit range (`af7011e09`) but was silently skipped during changelog generation.

The GitHub release notes have already been updated directly: https://github.com/Azure/azure-dev/releases/tag/azure-dev-cli_1.25.1

### 1. CHANGELOG.md fix
- Add the missing `#7400` entry to the 1.25.1 `Features Added` section

### 2. Changelog generation skill fix
- Add **reverse cross-reference (coverage check)** as Step 5.5 in `.github/skills/changelog-generation/SKILL.md`
- The existing forward cross-reference (Step 5.4) verifies included PRs are in the commit range, but does NOT verify the reverse — that every commit in range is accounted for
- The new step requires the skill to present a summary table showing every commit's disposition (included / excluded with reason), flagging any commit without a clear classification

## Root Cause

The skill's Step 5 validation only checked that *included* entries were valid (forward check). It never verified that all commits in the range were *covered* — either included or explicitly excluded. This meant a PR could be silently dropped without anyone noticing.

PR #7400 was likely skipped due to a processing error (possibly the em dash in the commit subject or an MCP fetch failure) with no flag raised because no reverse coverage check existed.

### DIFF
--- a/.github/skills/changelog-generation/SKILL.md
+++ b/.github/skills/changelog-generation/SKILL.md
@@ -85,8 +85,9 @@ Per [references/pr-processing.md](references/pr-processing.md):
 1. **Remove empty category sections** — scan the new release entry and delete any `### <Category>` heading that has no bullet entries beneath it (i.e., the next line is blank followed by another `###` heading or `##` heading or end of section). This is mandatory — never leave empty `### Breaking Changes`, `### Features Added`, `### Bugs Fixed`, or `### Other Changes` sections in the final output.
 2. For extensions using flat bullet lists (no category headings), match the existing style.
 3. **Validate PR links**: verify that every `- ` bullet under a category heading contains a `[[#NNNN]]` link. If any entry is missing a link, flag it and require either adding the PR reference or explicitly justifying the omission. When a single large PR warrants multiple changelog bullets, each must reference the same PR number.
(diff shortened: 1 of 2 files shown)


jongio (approved):
Changelog entry matches PR #7400 (merged May 12). Insertion order is correct (ascending PR numbers in the 1.25.1 section). The new reverse coverage check in the skill is a solid process fix - should prevent this class of miss going forward. Step numbering is internally consistent (no broken cross-references in the file).

hemarina (approved):

wbreza (approved):
Verified the changelog fix and skill update:

- `cli/azd/CHANGELOG.md`: `[[#7400]] azd exec` entry inserted in correct ascending PR-number order within the 1.25.1 `Features Added` section.
- `.github/skills/changelog-generation/SKILL.md`: clean split of the existing check into forward cross-reference (sub-step 4) and a new reverse coverage check (sub-step 5). Old sub-step 5 cleanly renumbered to 6.
- Sub-step renumbering is internally consistent; no broken references in `references/pr-processing.md` or `references/scope-rules.md` (the "Step 5 review" reference points to the top-level workflow step, which is unchanged).

The reverse coverage check directly addresses the root cause and should prevent this class of silent omission going forward. LGTM.

PR #8222, opened 2026-05-18
feat(agents): validate .NET runtime vs TargetFramework after init

### LINKED ISSUE #8202: azd ai agent init: validate .NET runtime version against .csproj TargetFramework
## Problem

When a user selects a .NET runtime version during `azd ai agent init` (e.g. dotnet 9), but the downloaded template's .csproj targets a different framework (e.g. `net10.0`), azd does not warn. The mismatch is only caught later during remote build on the server, resulting in a confusing failure.

## Proposed Solution

After template download (or during from-code init), read the `TargetFramework` property from the .csproj file and compare it against the user-selected runtime version. If mismatched:
- Warn the user and suggest correcting the .csproj, or
- Auto-correct the selection to match the .csproj, or
- Offer to update the .csproj TargetFramework to match the selected runtime

## Context

- Relevant code: `cli/azd/extensions/azure.ai.agents/internal/cmd/init_from_code.go` (already reads .csproj for AssemblyName)
- Currently azd only extracts `AssemblyName` from .csproj, does not read `TargetFramework`

v1212:
## Summary

- Adds a non-blocking post-init validation framework (`validatePostInit`) that runs after `azd ai agent init` completes
- First validator: checks .NET runtime selection against .csproj `TargetFramework`  prints a clear ERROR (red) with fix instructions if target > runtime (which will fail at build/deploy time)
- Prints green OK message when validation passes
- Prints ERROR when .csproj cannot be read, prompting user to verify manually
- Does NOT block or alter the init flow in any way  purely advisory output
- Wired into both template flow (`init.go`) and from-code flow (`init_from_code.go`)
- Python projects and container deploys are unaffected (skipped)

## Changes

| File | Description |
| ---- | ----------- |
| `init_validate.go` | New: validation framework + .NET runtime vs TargetFramework check |
| `init_validate_test.go` | New: comprehensive unit tests (mismatch, match, higher runtime, no csproj, unreadable dir, python skip) |
| `init.go` | Call `validatePostInit` after template init completes |
| `init_from_code.go` | Call `validatePostInit` after from-code init completes |

Closes #8202

## E2E Test Results (Manual Interactive Testing)

All tests run interactively using only `azd ai agent init` + `azd deploy` (no file modifications).

| Test | Scenario | Result |
|------|----------|--------|
| V1-P1 | C# Template Invocations, .NET 9 vs net10.0 csproj  ERROR message |  PASS |
| V1-P2 | C# Template Invocations, .NET 10 vs net10.0 csproj  OK + deploy + invoke |  PASS |
| 1 | Python Template Invocations, remote_build  deploy  invoke |  PASS |
| 3 | Python Template Responses, remote_build  deploy  invoke |  PASS |
| 6 | Python From-code vNext Responses, remote_build  deploy  invoke |  PASS |
| A | C# Template Invocations, .NET 10 remote_build  deploy  invoke |  PASS |
| B | C# Template Responses AgentFW, .NET 10 remote_build  deploy  invoke |  PASS |

**7/7 passed**  PR feature (runtime validation) works correctly, no regressions in init/deploy/invoke flows.

### DIFF
--- a/cli/azd/extensions/azure.ai.agents/internal/cmd/init_validate.go
+++ b/cli/azd/extensions/azure.ai.agents/internal/cmd/init_validate.go
@@ -1,0 +1,124 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package cmd
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+	"regexp"
+	"strconv"
+	"strings"
+
+	"azureaiagent/internal/pkg/agents/agent_yaml"
+
+	"github.com/fatih/color"
+)
+
+// validatePostInit runs all post-init validations and prints errors (non-blocking).
+// Validations are advisory — they highlight issues that will cause deploy failures
+// but do not prevent init from completing.
+func validatePostInit(srcDir string, codeConfig *agent_yaml.CodeConfiguration) {
+	if codeConfig == nil {
+		return
+	}
+
+	validateDotnetRuntimeVsCsproj(srcDir, codeConfig.Runtime)
+}
+
+// validateDotnetRuntimeVsCsproj checks whether the selected .NET runtime version is compatible
+// with the TargetFramework declared in the .csproj file. Prints an error (non-blocking) if:
+// - The .csproj cannot be read (user should verify their project structure)
+// - The .csproj targets a higher framework version than the selected runtime
+func validateDotnetRuntimeVsCsproj(srcDir string, runtime string) {
+	if !strings.HasPrefix(runtime, "dotnet_") {
+		return
+	}
+
+	// Parse selected runtime version (e.g. "dotnet_9" -> 9, "dotnet_10" -> 10)
+	runtimeVersionStr := strings.TrimPrefix(runtime, "dotnet_")
+	runtimeVersion, err := strconv.Atoi(runtimeVersionStr)
+	if err != nil {
+		return
+	}
+
+	// Find .csproj file in srcDir
+	entries, err := os.ReadDir(srcDir)
+	if err != nil {
+		fmt.Printf("\n%s Could not read project directory %q to validate .NET TargetFramework. "+
+			"Please verify your .csproj TargetFramework matches the selected .NET %d runtime before deploying.\n",
+			color.RedString("ERROR:"),
+			srcDir, runtimeVersion,
+		)
+		return
+	}
+
+	var csprojFound bool
+	for _, e := range entries {
+		if e.IsDir() || !strings.HasSuffix(e.Name(), ".csproj") {
+			continue
+		}
+		csprojFound = true
+
+		csprojPath := filepath.Join(srcDir, e.Name())
+		data, err := os.ReadFile(csprojPath) //nolint:gosec // path from user project
+		if err != nil {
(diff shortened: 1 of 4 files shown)


v1212 (commented):
.

jongio (commented):
Clean addition - advisory only, nil-safe, can't break existing flows. Two items below worth a look.

cli/azd/extensions/azure.ai.agents/internal/cmd/init_validate_test.go
```diff
@@ -0,0 +1,120 @@
...
+func TestValidateDotnetRuntimeVsCsproj_Mismatch(t *testing.T) {
+	// Create temp dir with a .csproj targeting net10.0
+	dir := t.TempDir()
+	csproj := `<Project><PropertyGroup><TargetFramework>net10.0</TargetFramework></PropertyGroup></Project>`
+	err := os.WriteFile(filepath.Join(dir, "MyAgent.csproj"), []byte(csproj), 0600)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Should not panic; prints ERROR to stdout (non-blocking)
```
jongio: **[MEDIUM]** Tests verify no-panic but don't assert on output. A regression that silently drops or changes the ERROR/OK messages won't be caught.

Consider capturing stdout and checking it contains expected substrings (`ERROR` for mismatch, `OK` for match).

cli/azd/extensions/azure.ai.agents/internal/cmd/init_validate.go
```diff
@@ -0,0 +1,124 @@
...
+		// (could be a pre-compiled DLL scenario), so skip silently.
+		return
+	}
+}
+
+// extractTargetFrameworkVersion parses the major version number from a .csproj TargetFramework element.
+// e.g. "<TargetFramework>net10.0</TargetFramework>" -> 10
+// Returns 0 if not found or not parsable.
+func extractTargetFrameworkVersion(csprojContent string) int {
+	re := regexp.MustCompile(`<TargetFramework>net(\d+)\.\d+</TargetFramework>`)
```
jongio: **[LOW]** Minor: this regex won't match TFMs with platform suffixes (`net8.0-windows`, `net8.0-android`). Unlikely for container-deployed agents, but if encountered, the user gets a "Could not parse" error instead of a graceful pass. `net(\d+)\.\d+[^<]*` would handle it.

cli/azd/extensions/azure.ai.agents/internal/cmd/init_validate_test.go
v1212: Agreed — added a captureStdout helper and assert on "ERROR" / "OK".

cli/azd/extensions/azure.ai.agents/internal/cmd/init_validate.go
v1212: Good point. Changed the regex to net(\d+)\.\d+[^<]* so TFMs like net8.0-windows are handled gracefully. Also added a test case for it as well.

trangevi (approved):

trangevi:
I'd like @therealjohn to take a look at this. On the surface I'm ok with it, but I'm a little concerned with the idea of adding warnings/error messages just before we say "everything is good you're done".

PR #8223, opened 2026-05-18
feat: add .agentignore support for agent code deploy packaging

### LINKED ISSUE #8170: Feature: .gitignore-style handling in azd code deploy (.azdignore support)
### Feature Request

Add `.gitignore`-style exclusion handling to `azd code deploy` via a user-customizable `.azdignore` file.

#### Background

Follow-up to the "Feature/code deploy zip upload" PR. From the review discussion:

> Good idea. Will add `.azdignore` file support as a follow-up PR. For now, added `.azure/`, `.env`, and `.env.*` to the hardcoded exclusion list to address the immediate security concern. The follow-up will let users customize exclusions.

Today, exclusions are hardcoded. Users need a way to customize what gets uploaded (e.g., `node_modules/`, build artifacts, local caches).

#### Proposed

- Support a `.azdignore` file at the service source root using standard `.gitignore` syntax.
- Keep hardcoded security exclusions (`.azure/`, `.env`, `.env.*`, `.git/`) always applied; not overridable via negation.
- Document behavior and add tests.

v1212:
## Summary

Closes #8170

- Adds `.agentignore` file support (`.gitignore` syntax) for user-configurable file exclusion during `azd ai agent deploy` code packaging
- When no `.agentignore` exists, built-in defaults apply (Python/.NET/Node artifacts, azd tooling files)
- When `.agentignore` is present, user rules fully replace defaults (like `.dockerignore` semantics)
- Default `.agentignore` includes security exclusions (`.env`, `.azure/`, `.git/`); users who customize should keep them
- `azd ai agent init` (both template and from-code flows) now generates a default `.agentignore`
- Shows polling attempt progress during `waitForAgentActive` so users know deployment is not stuck

## Design

- **Replace semantics when user file exists**: User has full control over exclusions
- **Default .agentignore includes security exclusions**: `.env`, `.azure/`, `.git/` are in defaults; users who customize should keep them
- **Single source of truth**: `DefaultAgentIgnoreContent()` raw string used both for runtime defaults and generated file
- **Uses `github.com/denormal/go-gitignore`** (already used by azd core for `.azdignore`)
- **No impact on existing flows**: Without `.agentignore` file, behavior is identical to before

## Testing

- Unit tests cover: no-file defaults, user override, negation patterns, symlink rejection, UTF-8 BOM handling, empty file
- All existing tests pass unchanged

## E2E Regression Test Results

**Environment:** azd 1.23.15, extension 0.1.31-preview, northcentralus
**Method:** Interactive `azd ai agent init` + `azd deploy` (no manual file modifications)

| # | Template | Language | Deploy | Protocol | Result |
|---|----------|----------|--------|----------|--------|
| 1 | Hello World Invocations | C# (.NET 10) | ZIP / remote_build | invocations |  Passed |
| 2 | Hello World Responses | Python | ZIP / remote_build | responses |  Passed |
| 3 | Hello World Responses (Agent Framework) | C# (.NET 10) | ZIP / remote_build | responses |  Passed |
| 4 | From-code (vnext-responses-streaming) | Python | ZIP / remote_build | responses |  Passed |
| 5 | Hello World Responses (template) | Python | ZIP / remote_build | responses |  Passed |
| 6 | Hello World Invocations (template) | C# (.NET 10) | ZIP / remote_build | invocations |  Passed |
| 7 | Agent with MCP Tools (Docker/ACR) | Python | Docker / ACR | responses |  Passed |

All 7 tests passed: init → deploy → invoke (verified HTTP 200 with correct agent response).

### DIFF
--- a/cli/azd/.vscode/cspell.yaml
+++ b/cli/azd/.vscode/cspell.yaml
@@ -2,6 +2,7 @@ import: ../../../.vscode/cspell.global.yaml
 words:
   - agentcopilot
   - agentdetect
+  - agentignore
   - Authenticode
   - azdignore
   - gofmt
--- a/cli/azd/.vscode/cspell.yaml
+++ b/cli/azd/.vscode/cspell.yaml
@@ -2,7 +2,6 @@ import: ../../../.vscode/cspell.global.yaml
 words:
   - agentcopilot
   - agentdetect
-  - agentignore
   - Authenticode
   - azdignore
   - gofmt
--- a/cli/azd/extensions/azure.ai.agents/go.mod
+++ b/cli/azd/extensions/azure.ai.agents/go.mod
@@ -28,6 +28,8 @@ require (
 	gopkg.in/yaml.v3 v3.0.1
 )
 
+require github.com/denormal/go-gitignore v0.0.0-20180930084346-ae8ad1d07817
+
 require (
 	dario.cat/mergo v1.0.2 // indirect
 	github.com/AlecAivazis/survey/v2 v2.3.7 // indirect
@@ -60,7 +62,6 @@ require (
 	github.com/clipperhouse/uax29/v2 v2.5.0 // indirect
 	github.com/danwakefield/fnmatch v0.0.0-20160403171240-cbb64ac3d964 // indirect
 	github.com/davecgh/go-spew v1.1.2-0.20180830191138-d8f796af33cc // indirect
-	github.com/denormal/go-gitignore v0.0.0-20180930084346-ae8ad1d07817 // indirect
 	github.com/dlclark/regexp2 v1.11.5 // indirect
 	github.com/go-logr/logr v1.4.3 // indirect
 	github.com/go-logr/stdr v1.2.2 // indirect
--- a/cli/azd/extensions/azure.ai.agents/go.mod
+++ b/cli/azd/extensions/azure.ai.agents/go.mod
@@ -58,7 +58,9 @@ require (
 	github.com/clipperhouse/displaywidth v0.9.0 // indirect
 	github.com/clipperhouse/stringish v0.1.1 // indirect
 	github.com/clipperhouse/uax29/v2 v2.5.0 // indirect
+	github.com/danwakefield/fnmatch v0.0.0-20160403171240-cbb64ac3d964 // indirect
 	github.com/davecgh/go-spew v1.1.2-0.20180830191138-d8f796af33cc // indirect
+	github.com/denormal/go-gitignore v0.0.0-20180930084346-ae8ad1d07817 // indirect
 	github.com/dlclark/regexp2 v1.11.5 // indirect
 	github.com/go-logr/logr v1.4.3 // indirect
 	github.com/go-logr/stdr v1.2.2 // indirect
--- a/cli/azd/extensions/azure.ai.agents/internal/cmd/init_from_code.go
+++ b/cli/azd/extensions/azure.ai.agents/internal/cmd/init_from_code.go
@@ -21,6 +21,7 @@ import (
 
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
 	"github.com/azure/azure-dev/cli/azd/pkg/azdext"
+	"github.com/azure/azure-dev/cli/azd/pkg/osutil"
 	"github.com/azure/azure-dev/cli/azd/pkg/ux"
 	"github.com/fatih/color"
 	"google.golang.org/protobuf/types/known/structpb"
@@ -822,8 +823,7 @@ func (a *InitFromCodeAction) writeDefinitionToSrcDir(definition *agent_yaml.Cont
 	// Generate .agentignore if it doesn't already exist
 	agentIgnorePath := filepath.Join(srcDir, ".agentignore")
 	if _, err := os.Stat(agentIgnorePath); os.IsNotExist(err) {
-		//nolint:gosec // generated ignore file should be readable by tooling and users
-		if err := os.WriteFile(agentIgnorePath, []byte(project.DefaultAgentIgnoreContent()), 0644); err != nil {
+		if err := os.WriteFile(agentIgnorePath, []byte(project.DefaultAgentIgnoreContent()), osutil.PermissionFile); err != nil {
 			return "", fmt.Errorf("writing .agentignore: %w", err)
 		}
 	}
--- a/cli/azd/extensions/azure.ai.agents/internal/cmd/init_from_code.go
+++ b/cli/azd/extensions/azure.ai.agents/internal/cmd/init_from_code.go
@@ -819,6 +819,15 @@ func (a *InitFromCodeAction) writeDefinitionToSrcDir(definition *agent_yaml.Cont
 		return "", fmt.Errorf("writing definition to file: %w", err)
 	}
 
+	// Generate .agentignore if it doesn't already exist
+	agentIgnorePath := filepath.Join(srcDir, ".agentignore")
+	if _, err := os.Stat(agentIgnorePath); os.IsNotExist(err) {
+		//nolint:gosec // generated ignore file should be readable by tooling and users
+		if err := os.WriteFile(agentIgnorePath, []byte(project.DefaultAgentIgnoreContent()), 0644); err != nil {
+			return "", fmt.Errorf("writing .agentignore: %w", err)
+		}
+	}
+
 	return definitionPath, nil
 }
 
--- a/cli/azd/extensions/azure.ai.agents/internal/project/agentignore.go
+++ b/cli/azd/extensions/azure.ai.agents/internal/project/agentignore.go
@@ -5,6 +5,7 @@ package project
 
 import (
 	"bytes"
+	"context"
 	"errors"
 	"fmt"
 	"io"
@@ -35,11 +36,12 @@ type agentIgnoreMatcher struct {
 
 // newAgentIgnoreMatcher creates a matcher by reading .agentignore from srcDir.
 // If no .agentignore exists, defaults are used.
-func newAgentIgnoreMatcher(srcDir string) (*agentIgnoreMatcher, error) {
+func newAgentIgnoreMatcher(ctx context.Context, srcDir string) (*agentIgnoreMatcher, error) {
+	_ = ctx // reserved for future cancellation support
 	m := &agentIgnoreMatcher{}
 
 	// Try to load user's .agentignore
-	ig, err := loadAgentIgnore(srcDir)
+	ig, err := loadAgentIgnore(ctx, srcDir)
 	if err != nil {
 		return nil, err
 	}
@@ -72,7 +74,8 @@ func (m *agentIgnoreMatcher) ShouldExclude(relPath string, isDir bool) bool {
 
 // loadAgentIgnore reads an .agentignore file from srcDir.
 // Returns nil, nil if no file exists.
-func loadAgentIgnore(srcDir string) (gitignore.GitIgnore, error) {
+func loadAgentIgnore(ctx context.Context, srcDir string) (gitignore.GitIgnore, error) {
+	_ = ctx // reserved for future cancellation support
 	path := filepath.Join(srcDir, agentIgnoreFileName)
 	info, err := os.Lstat(path)
 	if errors.Is(err, os.ErrNotExist) {
@@ -113,6 +116,7 @@ func loadAgentIgnore(srcDir string) (gitignore.GitIgnore, error) {
 func DefaultAgentIgnoreContent() string {
 	return `# Files excluded from agent code deployment packaging.
 # Uses .gitignore syntax.
+# Note: only the root .agentignore is read; subdirectory files are not supported.
 #
 # To include a file that is excluded by default, use negation: !filename
 
--- a/cli/azd/extensions/azure.ai.agents/internal/project/agentignore.go
+++ b/cli/azd/extensions/azure.ai.agents/internal/project/agentignore.go
@@ -10,6 +10,7 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+	"slices"
 	"strings"
 
 	gitignore "github.com/denormal/go-gitignore"
@@ -116,10 +117,8 @@ func (m *agentIgnoreMatcher) isSecurityExcluded(relPath string, isDir bool) bool
 
 	// Metadata files (agent.yaml, azure.yaml, etc.) — only at the root level
 	if !isDir && filepath.Dir(relPath) == "." {
-		for _, meta := range metadataExclusions {
-			if name == meta {
-				return true
-			}
+		if slices.Contains(metadataExclusions, name) {
+			return true
 		}
 	}
 
--- a/cli/azd/extensions/azure.ai.agents/internal/project/agentignore.go
+++ b/cli/azd/extensions/azure.ai.agents/internal/project/agentignore.go
@@ -28,6 +28,15 @@ var securityExclusions = []string{
 	".git/",
 }
 
+// metadataExclusions are files that are always excluded because they are deployment metadata
+// sent separately or not relevant inside the code package.
+var metadataExclusions = []string{
+	"agent.yaml",
+	"agent.manifest.yaml",
+	"azure.yaml",
+	agentIgnoreFileName,
+}
+
 // defaultExclusions are the built-in exclusion patterns applied when no .agentignore file exists.
 var defaultExclusions = []string{
 	// azd tooling files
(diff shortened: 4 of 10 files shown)


v1212 (commented):
.

v1212 (commented):
.

jongio (commented):
Uses the same go-gitignore library as core azd, security model is solid (non-negotiable exclusions can't be overridden), and tests cover the important edge cases well. Two consistency items below.

cli/azd/extensions/azure.ai.agents/internal/cmd/init_from_code.go
```diff
@@ -813,6 +813,15 @@ func (a *InitFromCodeAction) writeDefinitionToSrcDir(definition *agent_yaml.Cont
 		return "", fmt.Errorf("writing definition to file: %w", err)
 	}
 
+	// Generate .agentignore if it doesn't already exist
+	agentIgnorePath := filepath.Join(srcDir, ".agentignore")
+	if _, err := os.Stat(agentIgnorePath); os.IsNotExist(err) {
+		//nolint:gosec // generated ignore file should be readable by tooling and users
+		if err := os.WriteFile(agentIgnorePath, []byte(project.DefaultAgentIgnoreContent()), 0644); err != nil {
```
jongio: **[MEDIUM]** Use `osutil.PermissionFile` instead of raw `0644` for consistency with `init.go` (which uses the named constant for the same operation). You'd need to add the `osutil` import, but it removes the need for the `//nolint:gosec` annotation.

cli/azd/extensions/azure.ai.agents/internal/project/agentignore.go
```diff
@@ -0,0 +1,249 @@
...
+
+	// Strip UTF-8 BOM
+	data = bytes.TrimPrefix(data, utf8BOM)
+
+	return gitignore.New(bytes.NewReader(data), srcDir, nil), nil
+}
+
+// DefaultAgentIgnoreContent returns the default .agentignore file content
+// that should be generated during `azd ai agent init`.
+func DefaultAgentIgnoreContent() string {
```
jongio, line 116: **[MEDIUM]** `defaultExclusions` (line 41) and this function maintain the same pattern list independently. If someone adds a new default to one, they could forget the other.

Consider generating this content from `defaultExclusions` to keep them in sync, or at minimum add a `// Keep in sync with defaultExclusions` comment in both locations.

v1212:
## E2E Regression Test Results (PR #8223)

**Environment:** azd 1.23.15, extension 0.1.31-preview, northcentralus region
**Method:** Interactive `azd ai agent init` + `azd deploy` (no manual file modifications)

###  All 7 Tests Passed

| # | Template | Language | Deploy | Protocol | Time | Result |
|---|----------|----------|--------|----------|------|--------|
| 1 | Hello World Invocations | C# (.NET 10) | ZIP / remote_build | invocations | 1m25s |  |
| 2 | Hello World Responses | Python | ZIP / remote_build | responses | 2m35s |  |
| 3 | Hello World Responses (Agent Framework) | C# (.NET 10) | ZIP / remote_build | responses | 6m12s |  |
| 4 | From-code (vnext-responses-streaming) | Python | ZIP / remote_build | responses | 2m36s |  |
| 5 | Hello World Responses (template) | Python | ZIP / remote_build | responses | 20s |  |
| 6 | Hello World Invocations (template) | C# (.NET 10) | ZIP / remote_build | invocations | 1m25s |  |
| 7 | Agent with MCP Tools (Docker/ACR) | Python | Docker / ACR | responses | 4m11s |  |

### Test Details

- **Tests 1-6**: Used existing Foundry project (`foundry-test-0518`, northcentralus) with `azd deploy`
- **Test 7**: Full `azd up` flow provisioning new ACR + Docker image push + deploy
- **Model**: gpt-5.4-mini on all tests
- **Invocation**: All agents responded correctly within 4 minutes of deploy completion
- **Cleanup**: All agents cleaned up with `azd down --force --purge`

### Conclusion

The `.agentignore` feature does not cause regressions in the agent init/deploy/invoke flow across Python and C# templates with both ZIP (remote_build) and Docker (ACR) deployment methods.

cli/azd/extensions/azure.ai.agents/internal/project/agentignore.go
```diff
@@ -0,0 +1,249 @@
...
+	defaultIgnore gitignore.GitIgnore // used when no .agentignore file exists
+	hasUserIgnore bool
+	srcDir        string
+}
+
+// newAgentIgnoreMatcher creates a matcher by reading .agentignore from srcDir.
+// If no .agentignore exists, defaults are used.
+func newAgentIgnoreMatcher(srcDir string) (*agentIgnoreMatcher, error) {
+	m := &agentIgnoreMatcher{
+		srcDir:        srcDir,
```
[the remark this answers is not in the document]
v1212: Removed the srcDir field from the agentIgnoreMatcher struct — it was set during construction but never read anywhere.

cli/azd/extensions/azure.ai.agents/go.mod
```diff
@@ -58,7 +58,9 @@ require (
 	github.com/clipperhouse/displaywidth v0.9.0 // indirect
 	github.com/clipperhouse/stringish v0.1.1 // indirect
 	github.com/clipperhouse/uax29/v2 v2.5.0 // indirect
+	github.com/danwakefield/fnmatch v0.0.0-20160403171240-cbb64ac3d964 // indirect
 	github.com/davecgh/go-spew v1.1.2-0.20180830191138-d8f796af33cc // indirect
+	github.com/denormal/go-gitignore v0.0.0-20180930084346-ae8ad1d07817 // indirect
 	github.com/dlclark/regexp2 v1.11.5 // indirect
```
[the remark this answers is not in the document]
v1212: Ran go mod tidy which automatically corrected the // indirect marker since the package is directly imported.

cli/azd/extensions/azure.ai.agents/internal/cmd/init_from_code.go
v1212: Thanks, Replaced the hardcoded 0644 with osutil.PermissionFile (the same constant used in init.go) and added the osutil import. This also eliminated the //nolint:gosec annotation.

cli/azd/extensions/azure.ai.agents/internal/project/agentignore.go
v1212: Deleted the entire defaultExclusions slice (30+ lines). When no .agentignore file exists, the matcher now feeds DefaultAgentIgnoreContent() directly into gitignore.New(). The gitignore parser naturally ignores # comment lines and blank lines, so the behavior is identical. This makes DefaultAgentIgnoreContent() the single source of truth for both the generated file and the runtime defaults.

cli/azd/extensions/azure.ai.agents/internal/project/agentignore.go
```diff
@@ -0,0 +1,219 @@
...
+
+const (
+	agentIgnoreFileName = ".agentignore"
+	agentIgnoreMaxSize  = 1 << 20 // 1 MB
+)
+
+// securityExclusions are directory paths that are always excluded regardless of .agentignore content.
+// Users cannot negate these with "!" patterns.
+// Note: .env/.env.* files are handled separately in isSecurityExcluded.
+var securityExclusions = []string{
```
trangevi: I don't know that we should have this level of non-configurable blocking. I get the point, but if we're providing a mechanism for users to configure what they want to include or exclude, we shouldn't force additional things onto that. We should definitely provide it by default, and maybe even prompt users at some point if they don't have it, but I don't think we should have these mandatory ones with no user configuration.

cli/azd/extensions/azure.ai.agents/internal/project/agentignore.go
```diff
@@ -0,0 +1,219 @@
...
+	// Strip UTF-8 BOM
+	data = bytes.TrimPrefix(data, utf8BOM)
+
+	return gitignore.New(bytes.NewReader(data), srcDir, nil), nil
+}
+
+// DefaultAgentIgnoreContent returns the default .agentignore file content
+// that should be generated during `azd ai agent init`.
+func DefaultAgentIgnoreContent() string {
+	var sb strings.Builder
```
trangevi: This is pretty difficult to read to see what files we're excluding by default. Can this be done with a list of files that we write via a loop instead?

trangevi (changes requested):

cli/azd/extensions/azure.ai.agents/internal/project/agentignore.go
v1212: NO forced/non-overridable exclusions. When a user provides their own .agentignore, it completely replaces defaults. default agentignore file includes secret files like .env, etc.

cli/azd/extensions/azure.ai.agents/internal/project/agentignore.go
v1212: the content is well-organized with section headers and more readable now.

therealjohn (approved):

trangevi (approved):

wbreza (commented):
## Re-review after force-push (HEAD: `2b595b9`)

Thanks for the substantial rework. The intentional pivot to a fully user-configurable model (per @v1212's note and the `refactor: remove forced exclusions` commit) is reasonable as a design choice — but it makes the **PR description materially misleading**, which is the main item to address before merge. The rest below is non-blocking polish.

### 🔴 Must fix before merge — PR description / implementation mismatch

The PR description still says:

> *"Security exclusions (`.env`, `.azure/`, `.git/`) remain non-negotiable and cannot be overridden with `!` negation"*

…but `agentignore_test.go` now explicitly asserts the opposite, and the rewrite intentionally removed the enforcement layer. Future readers (reviewers, maintainers, security auditors, users finding this PR via git blame) will be misled. Please update the description to reflect the actual `.dockerignore`-style replace semantics — something like:

> *"Security-sensitive paths (`.env`, `.azure/`, `.git/`) ship in the default `.agentignore` template, but are fully user-configurable. When a user provides their own `.agentignore`, it replaces the defaults entirely — users are responsible for re-listing any paths they want excluded."*

While you're at it, consider adding a short note to the generated `.agentignore` header (or README) that calls this out explicitly so users understand the contract when they first edit the file.

### 🟠 Open items from prior review (non-blocking)

1. **`packageCodeDeploy` error classification** — `newAgentIgnoreMatcher` failures still come back as bare `fmt.Errorf` in `service_target_agent.go`. Per `cli/azd/extensions/azure.ai.agents/AGENTS.md`, orchestration-layer failures should be classified with `exterrors.Dependency(...)` so they pick up the proper code + user suggestion.

2. **`ctx` not threaded into I/O helpers** — `newAgentIgnoreMatcher(srcDir)` and `loadAgentIgnore(srcDir)` perform file I/O without a `context.Context`. `cli/azd/AGENTS.md` calls for `ctx` as the first parameter of I/O functions. The caller (`packageCodeDeploy`) already has one available.

3. **Symlinked-directory traversal in the walker** — `.agentignore` itself correctly rejects symlinks, but the ZIP walker in `service_target_agent.go` only checks symlink-ness for files. A directory symlink (e.g., `subdir -> /etc`) would let the walker descend out of the project root. Suggest adding `if d.IsDir() && d.Type()&fs.ModeSymlink != 0 { return filepath.SkipDir }` in the walker callback.

4. **Polling progress display** — the `attempt N/30` line is printed once per iteration. If the `progress` channel in the extension UI doesn't collapse these in-place, users will see ~30 stacked lines. Worth a quick smoke test in a real terminal; if it does spam, switch to a spinner or carriage-return inline update.

5. **Discoverability** — `azd ai agent init --help` / `azd ai agent deploy --help` and the extension's CHANGELOG/README still don't mention `.agentignore`. Recommend a short paragraph in each (and a CHANGELOG entry for the next release).

6. **Subdirectory `.agentignore` not supported** — unlike `.gitignore`, only the root file is read. Worth a one-line note in the generated default's header so users with a `.gitignore` mental model aren't surprised.

### ✅ Resolved since prior review

- `srcDir` unused field removed
- `go mod tidy` correction (no more `// indirect`)
- `osutil.PermissionFile` used consistently
- Default exclusions unified into a single source of truth (`DefaultAgentIgnoreContent()`)
- Tests added for negation, empty file, UTF-8 BOM, symlink rejection, max file size
- `init` and `init_from_code` both generate `.agentignore`
- Polling counter wired up
- `go fix` modernization applied

### CI

27/28 checks green; only the approval gate is open. Nice work tightening this up — once the PR description is updated, this is ready for another pass from @trangevi.

cli/azd/.vscode/cspell.yaml
```diff
@@ -2,6 +2,7 @@ import: ../../../.vscode/cspell.global.yaml
 words:
   - agentcopilot
   - agentdetect
+  - agentignore
```
trangevi: This should be moved into the extension specific cspell I believe

trangevi (changes requested):

cli/azd/.vscode/cspell.yaml
trangevi: Please also check Wallace's comments. The must fix is just PR description, but the other comments might be valuable.

wbreza (commented):
## Re-review after `9e37929`

Thanks — the updated PR description now correctly reflects the `.dockerignore`-style replace semantics. That clears my prior blocking concern. ✅

The latest commit shortened the polling message text, which helps readability. The display still emits one line per attempt though, so if the extension's progress channel doesn't collapse them in-place, terminals will still show ~30 stacked lines. Worth a quick smoke test in an interactive run; if it does stack, a spinner / inline carriage-return update would be a nicer UX.

### Re-flagging open non-blocking items from my prior pass

None of these are merge-blockers from my side, but listing them in one place so they're easy to triage as a follow-up issue or quick polish commit:

1. **`packageCodeDeploy` error classification** — `newAgentIgnoreMatcher` failures still surface as bare `fmt.Errorf` in `service_target_agent.go`. Per `cli/azd/extensions/azure.ai.agents/AGENTS.md`, orchestration-layer failures should be wrapped with `exterrors.Dependency(...)` so users get the proper code + suggestion.

2. **`ctx` not threaded into I/O helpers** — `newAgentIgnoreMatcher(srcDir)` and `loadAgentIgnore(srcDir)` perform file I/O without a `context.Context`. `cli/azd/AGENTS.md` calls for `ctx` as the first parameter on I/O functions; the caller (`packageCodeDeploy`) already has one.

3. **Symlinked-directory traversal in the ZIP walker** — `.agentignore` itself correctly rejects symlinks, but the walker only checks symlink-ness for files. A directory symlink (e.g., `subdir -> /etc`) would let the walker descend outside the project root. Suggest adding `if d.IsDir() && d.Type()&fs.ModeSymlink != 0 { return filepath.SkipDir }` in the walker callback.

4. **Polling progress display** — text is shorter now ✅, but frequency unchanged. Verify whether the extension's `progress` channel updates in place or appends; if it appends, switch to a spinner.

5. **Discoverability** — `azd ai agent init --help`, `azd ai agent deploy --help`, the extension README, and the next CHANGELOG entry should all mention `.agentignore`. Right now this feature is essentially invisible unless users notice the file on disk.

6. **Subdirectory `.agentignore` not supported** — unlike `.gitignore`, only the root file is read. A one-line note in the generated default's header ("only the root `.agentignore` is read; subdirectory files are ignored") would prevent surprise for users with a `.gitignore` mental model.

### Also supporting @trangevi's cspell point

Agree with @trangevi on `cli/azd/.vscode/cspell.yaml` — per repo memory and the existing pattern in that file, extension-scoped words should live in the per-file `overrides` section (or in the extension's own cspell config) rather than the core azd global list. The term `agentignore` is only meaningful inside `cli/azd/extensions/azure.ai.agents/`, so scoping it accordingly keeps the global dictionary lean. Easy follow-up.

### Status

CI is green (29/29). Outstanding blocker is @trangevi's `CHANGES_REQUESTED` on the cspell placement — once that's resolved, this should be good to merge from my side.

cli/azd/.vscode/cspell.yaml
v1212: agentignore is removed from the cspell and moved into extension specific cspell at cli/azd/extensions/azure.ai.agents/cspell.yaml

v1212:
Thanks for the review and detailed suggestions. Here are improvements in the latest commit:

1. **`packageCodeDeploy` error classification** — wrapped with `exterrors.Dependency()` to provide proper code + suggestion
2. **`ctx` not threaded into I/O helpers** — `ctx context.Context` is now passed as the first parameter following existing practice
3. **Symlinked-directory traversal in the ZIP walker** — added check in the walker callback to skip symlinked directories
4. **Polling progress display** — verified: the current CLI UI updates in place via ANSI cursor control and won't show multiple stacked lines
5. **Discoverability** — added entry in CHANGELOG and `init` command's `Long` description mentioning `.agentignore`
6. **Subdirectory `.agentignore` not supported** — added note in the generated default's header to remind users

Also moved `agentignore` from global `cli/azd/.vscode/cspell.yaml` to extension-specific `cli/azd/extensions/azure.ai.agents/cspell.yaml` per @trangevi's feedback.

wbreza (approved):
LGTM ✅

Thanks @v1212 — the latest commit cleanly addresses every item from my prior pass:

- ✅ `packageCodeDeploy` failures now classified via `exterrors.Dependency(...)`
- ✅ `ctx context.Context` threaded through `newAgentIgnoreMatcher` and `loadAgentIgnore`
- ✅ Walker now skips symlinked directories (`SkipDir` on `d.IsDir() && d.Type()&fs.ModeSymlink != 0`)
- ✅ Polling display verified — `progress` callback updates in place via ANSI, no per-line stacking
- ✅ CHANGELOG entry added and `init` command's `Long` description now mentions `.agentignore`
- ✅ Subdirectory-not-supported note added to the generated default's header
- ✅ `agentignore` moved from core `cli/azd/.vscode/cspell.yaml` to the extension's own cspell config per @trangevi

PR description matches the implementation, tests are comprehensive, and the design rationale (replace semantics, default template carries security exclusions, single source of truth) is clearly laid out in the body. Nice work iterating through the rounds of feedback.

therealjohn (approved):

trangevi (approved):

PR #8224, opened 2026-05-18
feat(skills): add `azd ai skill` command group

### LINKED ISSUE #8142: Add `azd ai skill create/update/show/list/download/delete` to manage Foundry Skills from any directory

huimiu:
Closes #8142. Implements the design from #8204.

## Summary

New standalone `azure.ai.skills` extension exposing `azd ai skill create | update | show | list | download | delete` on top of the **versioned Skills API** introduced in [Azure/azure-rest-api-specs#43283](https://github.com/Azure/azure-rest-api-specs/pull/43283) (`Foundry-Features: Skills=V1Preview`).

## Commands

- **`create <name>`**: creates the skill and uploads its first default version via `POST /skills/{name}/versions`. Three mutually exclusive modes:
  - inline: `--description "..." --instructions "..."` -> JSON `inline_content`
  - SKILL.md: `--file ./SKILL.md` -> parsed locally, sent as `inline_content`
  - package: `--file ./skill.zip` -> `multipart/form-data` with a single `files[]` part
  - `--force` does delete-then-create.
- **`update <name>`**: uploads a new default version (same inline / SKILL.md modes). `--set-default-version <ver>` repoints `default_version` at an existing immutable version via `POST /skills/{name}` without uploading new content. ZIP is rejected here with a pointer to `create --force`.
- **`show <name>` / `list`**: `GET /skills/{name}` / `GET /skills`.
- **`download <name>`**: `GET /skills/{name}/content` (always `application/zip`). Extracts into `./.agents/skills/<name>/` with the existing safe-extraction guards (zip-slip, no symlinks, 10K-entry / 512 MB caps). `--raw` keeps the archive; `--version <ver>` targets `/versions/{version}/content`.
- **`delete <name>`**: `DELETE /skills/{name}`. Confirms by default; `--force` skips, `--no-prompt` requires `--force`.

## Conventions

- Skill names follow the agentskills.io spec: `^[a-z0-9]([a-z0-9\-]*[a-z0-9])?$`, max 64 chars (lowercase only).
- Endpoint resolution cascade shares state with `azure.ai.agents` via read-only fallback to `extensions.ai-agents.project.context.endpoint`.

## Test

- Install ai.skill extension
    image

- Create skill
    image

- Create skill (zip)
  image

- Show skill
   image

- List skills
    image

- Download skill
  image

- Update skill
  image

- Delete skill
  image


cli/azd/extensions/azure.ai.skills/internal/cmd/skill_create.go
```diff
@@ -0,0 +1,401 @@
...
+	return noPrompt || format == outputJSON
+}
+
+func isNotFound(err error) bool {
+	var respErr *azcore.ResponseError
+	if errors.As(err, &respErr) {
+		return respErr.StatusCode == 404
+	}
+	return false
+}
```
jongio, line 466: **[MEDIUM]** Use `errors.AsType` per Go 1.26 convention

AGENTS.md says to always use `errors.AsType` when checking if an error matches a specific type. This still uses the old two-step `errors.As` pattern. `exterrors/errors.go` already uses the correct style (`errors.AsType[*azcore.ResponseError]`) so it'd be consistent to do the same here.

```suggestion
func isNotFound(err error) bool {
if respErr, ok := errors.AsType[*azcore.ResponseError](err); ok {
return respErr.StatusCode == 404
}
return false
}
```

jongio (commented):
Clean extension structure - follows the `azure.ai.agents` patterns well, archive security is solid (zip-slip guard, symlink rejection, staging directory, size caps), and the test coverage across client/archive/parser/validation is thorough.

One nit on Go 1.26 style below. The Copilot bot's comment about the unused `bufio.Scanner` in `findFrontMatterBounds` is valid too - that variable gets `.Buffer()` called so it compiles, but it's never actually used for scanning.

cli/azd/extensions/azure.ai.skills/internal/cmd/skill_create.go
huimiu: Good catch, thanks! Switched to errors.AsType[*azcore.ResponseError]

cli/azd/extensions/azure.ai.skills/internal/cmd/skill_download.go
```diff
@@ -0,0 +1,259 @@
...
+	if err != nil {
+		return exterrors.ServiceFromAzure(err, exterrors.OpGetSkill)
+	}
+	if !skill.HasBlob {
+		return exterrors.Validation(
+			exterrors.CodeSkillNoPackage,
+			fmt.Sprintf("skill %q has no downloadable package", a.flags.name),
+			"only skills created from a `.zip` archive have a downloadable "+
+				"package. Use `azd ai skill show <name>` to inspect metadata; "+
+				"re-create with `azd ai skill create <name> --file <archive>.zip --force` "+
```
trangevi: I don't think we should have this message about re-creating. In theory, if a user is trying to download the skill, they don't have the skill locally with which to recreate it.

cli/azd/extensions/azure.ai.skills/internal/cmd/skill_download.go
```diff
@@ -0,0 +1,259 @@
...
+	cmd := &cobra.Command{
+		Use:   "download <name>",
+		Short: "Download a Foundry skill package.",
+		Long: `Download a skill's package.
+
+By default the CLI extracts the archive into --output-dir (default
+'./.agents/skills/<name>/'). Pass --raw to write the unmodified archive
+into --output-dir instead.
+
+Extraction enforces strict safety rules: no absolute paths, no '..' segments,
```
trangevi, line 218: Would we be better served by not doing the extraction ourselves? Or requiring it as an opt-in instead? So that these security concerns are more of an opt in from the user, rather than an opt out? 

@therealjohn thoughts?

cli/azd/extensions/azure.ai.skills/internal/cmd/skill_list.go
```diff
@@ -0,0 +1,66 @@
...
+With --top, the CLI stops once that many items have been collected.`,
+		Args: cobra.NoArgs,
+		RunE: func(cmd *cobra.Command, _ []string) error {
+			flags.output = extCtx.OutputFormat
+			flags.projectEndpoint, _ = cmd.Flags().GetString("project-endpoint")
+			return action.Run(azdext.WithAccessToken(cmd.Context()))
+		},
+	}
+
+	cmd.Flags().IntVar(&flags.top, "top", 0, "Return up to N skills (default: all)")
```
trangevi: Is there a better name we can give this? Or we think it's a common enough thing that it won't be weird to people?

@therealjohn

cli/azd/extensions/azure.ai.skills/internal/cmd/skill_update.go
```diff
@@ -0,0 +1,176 @@
...
+
+	cmd := &cobra.Command{
+		Use:   "update <name>",
+		Short: "Update an existing Foundry skill.",
+		Long: `Update an existing skill's description, instructions, or metadata.
+
+Pass any subset of:
+  --description "..."  --instructions "..."
+or:
+  --file ./SKILL.md    (parsed locally; .zip is not accepted here)
```
trangevi: Should we also indicate that updating zipped skills needs to be done via `create` with `--force`? Or am I misunderstanding what this means?

cli/azd/extensions/azure.ai.skills/internal/cmd/skill_update.go
```diff
@@ -0,0 +1,176 @@
...
+
+The CLI fetches the current skill, merges your changes locally, then POSTs the
+merged payload to the service.`,
```
trangevi: Skills aren't versioned, correct? Should we call that out, that this will be a destructive operation?

trangevi (changes requested):
Please add tests where possible

cli/azd/extensions/azure.ai.skills/internal/cmd/skill_list.go
therealjohn: Can we remove these flags for now? I don't see how `top` or `orderby` is useful without more support to query for exactly what you'd want. We can capture filtering as a separate feature to think more on what it needs.

cli/azd/extensions/azure.ai.skills/internal/cmd/skill_download.go
```diff
@@ -0,0 +1,257 @@
...
+	if !skill.HasBlob {
+		return exterrors.Validation(
+			exterrors.CodeSkillNoPackage,
+			fmt.Sprintf("skill %q has no downloadable package", a.flags.name),
+			"only skills created from a `.zip` archive have a downloadable "+
+				"package. Use `azd ai skill show <name>` to inspect this skill's metadata.",
+		)
+	}
+
+	body, err := skillCtx.client.Download(ctx, a.flags.name)
```
therealjohn: It's expected to be able to download a single skill - it should materialize the skill file.

cli/azd/extensions/azure.ai.skills/internal/cmd/skill_context.go
```diff
@@ -0,0 +1,52 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
trangevi, line 1: Should this be named "skill_context.go"?

wbreza (commented):
Thanks for the scope and care on this PR, @huimiu — there's a lot of solid work here: thorough archive safety guards, well-structured `exterrors`, good test patterns where they exist, and clean alignment with the `azure.ai.agents` extension. The notes below are intended to be additive to the existing feedback from @trangevi, @therealjohn, @jongio, and Copilot — I tried hard not to duplicate anything they already raised. Where I've flagged concerns, I've tried to explain the *why* and point at references so the reasoning is portable to similar code later.

## 🔴 HIGH — Two functional bugs worth fixing before merge

### 1. `update` silently drops the `Instructions` field
`internal/cmd/skill_update.go` — in the GET-merge-POST flow, `UpdateRequest` is initialized with `Description` and `Metadata` from `current`, but **`Instructions` is not carried forward**. If a user runs `azd ai skill update <name> --description "new"` without providing `--instructions`, the merge writes back an empty `Instructions` field, silently wiping the existing body.

```go
req := skill_api.UpdateRequest{
    Description:  current.Description,
    Metadata:     current.Metadata,
    // Instructions: current.Instructions,  // <-- missing
}
```

This is the classic GET-modify-PUT trap when a partial-update API is layered on top of a full-replace endpoint. Worth adding a regression test in a new `skill_update_test.go` that exercises "update only description, verify instructions preserved." See [AGENTS.md → Error Handling / partial updates](cli/azd/AGENTS.md) for the established pattern.

### 2. `PeekArchiveSkillName` swallows errors → `create --force` safety guard can be bypassed
`internal/pkg/skill_api/archive_peek.go` — the function returns `("", nil)` on three error paths (`entry.Open()`, `io.ReadAll()`, `ParseSkillMd()`). The `create --force` flow relies on this name-match check to confirm "the archive I'm about to delete-then-recreate actually claims to be the skill the user named." A malformed archive currently returns "no name claim" (✓ from the caller's POV), so `--force` proceeds and **deletes the existing skill anyway**.

There are two reasonable fixes:
- **Strict** (safer default): propagate the error so the caller refuses the destructive op until the archive is readable.
- **Lenient with telemetry**: log a clear warning so the user sees that the safety check was skipped, and require an extra flag (e.g. `--allow-unverified`) before proceeding.

Either way, please don't let an unreadable archive look identical to "intentionally unnamed archive" — those should be distinguishable outcomes.

## 🟠 MEDIUM — Security & resource hardening on the extraction path

(All below are *new* findings — not duplicates of Copilot's symlink/`copyFile` or `verifyPackageNameMatches` OOM comments.)

- **UNC path in zip-slip guard** — `archive.go` `ValidateEntryName`: after normalizing backslashes to forward slashes, the function checks for Windows drive letters (`C:`) but not UNC roots (`\\server\share` → `//server/share`). On Windows this can extract outside `OutputDir`. Suggest adding a `strings.HasPrefix(slashed, "//")` rejection after the drive-letter check, plus a test alongside the existing `C:\Windows\Temp` case.
- **Tar entry-count off-by-one** — `archive.go` `stageFromTarGz`: `entryCount++` then `if entryCount > maxEntries` accepts `maxEntries + 1` entries before rejecting. Zip side uses `len(zr.File) > maxEntries` on the full slice (correct). Change tar to `>=` or move increment after the check for symmetry.
- **File-handle leak in `PeekArchiveSkillName`** — `archive_peek.go`: when `io.ReadAll` or `ParseSkillMd` fails, the function returns without closing the entry reader. Pair the open with `defer rc.Close()` immediately after a successful `entry.Open()`.
- **Tar `hdr.Linkname` not asserted empty** — `archive.go` `stageFromTarGz`: the `default:` branch in the typeflag switch catches `TypeSymlink`/`TypeLink`/etc., which is correct, but an explicit `if hdr.Linkname != "" { return error }` for the regular-file branch makes the intent obvious and protects against future tar variants that smuggle link intent in extension headers (PAX/GNU).
- **Partial output dir not cleaned on extraction failure** — `skill_download.go`: when `SafeExtract` fails mid-stream, the user-visible `OutputDir` may contain partially-extracted files. Either (a) extract into a sibling temp dir and atomic-rename on success, or (b) `os.RemoveAll(opts.OutputDir)` on error when the directory was created by this invocation. This also helps with `--force` re-runs.

## 🟠 MEDIUM — Test coverage on the two largest command files

`skill_create.go` (~409 LOC, 3 mutually-exclusive modes + `--force` delete-then-create sequence + name-match guard) and `skill_update.go` (~181 LOC, GET-merge-POST + flag validation + gzip rejection) currently have **no `_test.go` companions**. Given how much of the user-facing surface area lives in these two files — and that bug #1 above sits squarely inside one of them — these are the highest-leverage tests to add. The existing `archive_test.go` is a great template for the table-driven style.

Suggested table-driven cases at minimum:
- `skill_create_test.go`: each mode (inline / `.md` / `.zip`), conflicting flag combos, `--force` with name match, `--force` with name mismatch, `--no-prompt` without required flags.
- `skill_update_test.go`: merge preserves all current fields when only one is changed, `.zip` rejection error message, conflicting flags.

## 🟡 MEDIUM — PR description vs. implementation alignment

The PR description and CHANGELOG say `create` accepts a "gzip package", but `selectCreateMode()` only matches `.zip`, and upload `Content-Type` is hardcoded to `application/zip`. Downloads correctly auto-detect both formats. Either:
- update the PR description / CHANGELOG / README to say **"ZIP package on upload; ZIP or gzip-tar on download"**, or
- extend `selectCreateMode()` to accept `.tar.gz` / `.tgz` and set the right content type.

This pairs with @copilot's earlier note on `models.go` package doc and README `--file ./skill.zip`.

## 🟢 LOW / NIT — Suggestions, not blockers

- **Cobra `MarkFlagsMutuallyExclusive`** — `skill_create.go` hand-rolls the `inlineProvided && fileProvided` check. Cobra's built-in is more durable as flags accrete: `cmd.MarkFlagsMutuallyExclusive("description", "file")` (and similar for `instructions`/`file`). Worth a tiny follow-up.
- **`--force` semantics diverge across commands** — `create --force` = destructive replace, `delete --force` = skip confirm, `download --force` = overwrite. All three are reasonable in isolation, but consider whether `create --force` deserves a more explicit name (`--replace` / `--recreate`) given that it deletes user data. At minimum, the create flag's help text could lead with "destroys the existing skill."
- **`update` error message** — when rejecting `.zip` on update, the message points at `create --force` but doesn't mention that the path is destructive (skills aren't versioned). Adding "(this deletes the existing skill first)" to the suggestion would close the loop with @trangevi's earlier comment about destructive ops.
- **Endpoint cascade as shared helper** — the 5-level cascade (`flag → env → ai-skills config → ai-agents config → host env`) is going to want to live in `cli/azd/pkg/azdext/` the moment a third AI extension shows up. Not for this PR, but worth filing a follow-up so it doesn't get copy-pasted.
- **`exterrors` duplication with `azure.ai.agents`** — `Validation` / `Dependency` / `Auth` / `ServiceFromAzure` factories are near-identical across the two extensions. Same recommendation as above: candidate for `cli/azd/pkg/azdext/exterrors` in a follow-up so a future error-handling tweak doesn't need parallel PRs.

---

Net: two real correctness bugs (#1 and #2) and a couple of Windows-path / resource-cleanup hardening items are the things I'd most like to see land before this merges. Everything else is polish or follow-up. Really nice extension overall — happy to dig into any of these if helpful.

wbreza (approved):
Great turnaround, @huimiu — verified `ad851b7d` against my prior findings. All HIGH issues fixed correctly and most MEDIUMs landed cleanly. Approving from my side.

## Verification

| Prior finding | Status |
|---|---|
| 🔴 `update` drops `Instructions` field | ✅ `current.Instructions` now carried forward; `skill_update_test.go` covers the preservation case |
| 🔴 `PeekArchiveSkillName` swallows errors → `--force` bypass | ✅ Errors propagate; `TestVerifyPackageNameMatches_MalformedSkillMd` confirms `--force` is blocked when the archive is unreadable |
| 🟠 UNC path bypass in zip-slip guard | ✅ Explicit `//` prefix rejection added; tests for both `//server/share` and `\\server\share` |
| 🟠 Tar entry-count off-by-one | ✅ Check before increment; now symmetric with the zip path |
| 🟠 File-handle leak in `PeekArchiveSkillName` | ✅ `defer rc.Close()` immediately after open |
| 🟠 Tar `hdr.Linkname` not asserted empty | ✅ Explicit check on the regular-file branch + `TestSafeExtract_TarGzRejectsEntryWithLinkname` |
| 🟠 Missing tests for create/update | ✅ Both `skill_create_test.go` and `skill_update_test.go` added — table-driven, cover conflicting flags, name mismatch under `--force`, and the field-preservation case |
| 🟡 Gzip-vs-ZIP doc inconsistency | ✅ `models.go` package doc and `archive_peek.go` now state the ZIP-upload / ZIP-or-gzip-download asymmetry explicitly |
| @therealjohn's `--top` / `--orderby` removal | ✅ Both flags gone from `skill_list.go` |

## One deferred item (non-blocking)

The "partial `OutputDir` not cleaned on extraction failure" MEDIUM in `skill_download.go` isn't part of this commit. Fine to defer — the staging directory is cleaned up, and the user can re-run with `--force` to recover. Worth a follow-up issue when you have the bandwidth; an atomic temp-dir-then-rename on the publish phase would close it cleanly.

## No new issues introduced

Skimmed all 9 changed files. No regressions, no over-corrections, no broken callers. Error propagation reads correctly throughout.

LGTM. ✅

cli/azd/extensions/azure.ai.skills/internal/cmd/skill_download.go
trangevi: Sync'ed offline, nvm

therealjohn: Keep it as-is. Requiring manual extraction is additional friction and this isnt public web, its your foundry project.

cli/azd/extensions/azure.ai.skills/internal/cmd/skill_download.go
huimiu: The download command now materializes a `SKILL.md` from the metadata returned by `GET /skills/{name}` when the skill has no archive blob, so `azd ai skill download <name>` always produces a file on disk for a single skill. The blob path (zip / tar.gz) is unchanged. `--raw` is rejected for blob-less skills since there is no archive to write.

cli/azd/extensions/azure.ai.skills/internal/cmd/skill_context.go
huimiu: Renamed

wbreza (commented):
Re-review on top of the versioned-API refactor (commits since my approval at `ad851b7`: `7e185c9`, `e8abae7`, `c040787`, `3463d4d`).

The refactor reads cleanly — naming is consistent across client/cmd, the multipart shape matches the new `POST /skills/{name}/versions` contract, the `--set-default-version` repoint path is well-isolated from content upload, new client methods have unit tests, and the safe-extract guards survived intact. Nice work threading the new contract through `create` / `update` / `download` without bloating the surface area.

A handful of findings worth considering as follow-ups — none block this PR on my read, but H1 and H2 in particular are worth filing issues for even if you don't want to address them in this PR.

### Higher-impact (worth a follow-up issue at minimum)

1. **`CreateVersionFromZip` buffers the full archive in memory** — `client.go` (around the new multipart build). The pre-refactor `CreatePackage` took an `io.ReadSeeker` and streamed via `httpReq.SetBody(streaming(archive), …)`. The new path does `io.Copy(part, archive)` into a `bytes.Buffer` and then `bytes.NewReader(buf.Bytes())`. Given the documented 512 MB archive ceiling, a near-cap upload allocates ≥512 MB heap (often 2× during `buf.Bytes()`), plus pipeline retry buffering. Easy OOM on constrained runners. Consider `io.Pipe` — a goroutine writes the `files[]` part + `default` field to the pipe writer, pass the pipe reader to `httpReq.SetBody`, set `ContentLength = -1`. Restoring the `io.ReadSeeker` parameter would also let the pipeline rewind on retry.

2. **`POST /skills/{name}/versions` has no idempotency guard** — the Azure SDK pipeline retries automatically. A client-side timeout on a request the server actually processed will create a second immutable version, and because both `CreateVersionInline` and `CreateVersionFromZip` set `default: true`, the retry silently re-flips `default_version` to the duplicate. The user is now running on a version they didn't intend. Options: send an `Idempotency-Key` if the service supports it, or set `policy.RetryOptions{MaxRetries: 0}` for these two requests so retries are explicit. Worth checking with the service team which they prefer.

### Medium

3. **`--force` is delete-then-create with no rollback, and the blast radius grew under versioning** — `DeleteSkill` now wipes the skill *and every version*. If the subsequent `CreateVersion*` fails (network, 5xx, validation), all prior versions are gone. `verifyFileNameMatches` protects against the wrong-target case but not the upload-failure case. Two ideas: (a) on `--force` without `--no-prompt`, prompt with the version count and `default_version` name being deleted; (b) longer-term, implement `--force` as "upload new non-default version → repoint default → delete old versions" so an upload failure leaves the original intact.

4. **SKILL.md `license`, `compatibility`, `allowed_tools` are silently dropped on upload** — `SkillInlineContent` (`models.go:51-54`) exposes these fields on the wire, but `ParseSkillMd` (`skill_md.go:31-87`) only extracts `name` / `description` / `metadata`, and neither `runFileMd` nor `buildInlineContent` copies them through. Round-trip a SKILL.md and these fields disappear, even though README/AGENTS claim agentskills.io conformance. Either extend `ParseSkillMd` + propagate (recommended), or drop the fields from `SkillInlineContent` until they're plumbed end-to-end. Worth a round-trip test either way.

5. **`--set-default-version` is not pre-flighted** — `UpdateSkillDefaultVersion` blindly POSTs `{default_version: X}`. With no client-side `GetSkillVersion` first, "skill not found" / "version not found" / "version exists but half-deleted" can all surface as a generic server 400. A short pre-flight call + a `CodeSkillVersionNotFound` error code would let users distinguish these cases. (Also no test for the error path.)

6. **`update` inline validation is split across two stages with a misleading error** — `validateFlags` accepts *at least one* of `--description` / `--instructions`, but `buildInlineContent` requires *both* non-empty. Running `azd ai skill update foo --description "new"` passes the first gate, then fails at build time with "update requires non-empty instructions" — without ever explaining that update needs both. Help text reads `(--description / --instructions)` where the `/` looks like "or". Easiest fix: tighten `validateFlags` to require both when inline mode is used; update help to `(--description AND --instructions)`.

7. **The gzip-tar branch in `archive.go` is dead under the new content-type check** — `downloadContent` (`client.go:425-432`) hard-rejects anything that isn't `application/zip`, but `SafeExtract` / `DetectArchiveFormat` still carry an `ArchiveTarGz` branch with a comment claiming "resilience". That branch can never trigger. Either restore `Accept: application/zip, application/gzip` on the download (real resilience), or delete `ArchiveTarGz` + the magic-byte sniff and call `zip.NewReader` directly. Pick one, drop the other.

### Low / cleanup

8. `client.go` package-level `downloadByteCap` is a mutable global mutated by tests. Safe today only because nothing calls `t.Parallel()`. Promote to a `Client` field with a `WithMaxDownloadBytes` option.
9. `skill_md.go` `SkillMdFileName` constant is dead after `c040787` removed the materialize path. Delete.
10. `printCreateResult` (skill_create.go) and the analogous post-update GET in `skill_update.go` silently swallow the follow-up `GetSkill` error. A `fmt.Fprintf(os.Stderr, "Warning: ... follow-up GET failed: %v\n", err)` would surface transient permission/network issues without failing the command.
11. `skill_download.go` accepts `--version ""` / `--version "   "` and silently downloads the default version. `strings.TrimSpace` + reject empty would catch unset-`$VER` scripting bugs.
12. Worth adding a test that asserts `update <name> --description "x"` (no `--instructions`) is rejected — documents the contract once you decide how to fix #6.
13. `--version` flag help on `download` could note "must exist on the server" so users don't expect arbitrary strings.

Not blocking — your call on which of these are in-scope for this PR vs follow-ups. Happy to file the issues myself if useful.

trangevi (approved):
Filed https://github.com/Azure/azure-dev/issues/8366 for non-blocking follow up items

vhvb1989 (approved):

PR #8233, opened 2026-05-18
Allow AI agent invoke to target versions

### LINKED ISSUE #8054: Allow `azd ai agent invoke` to target a specific agent version
azd ai agent invoke is missing a way to invoke a specific version of an agent.

glharper:
## Summary
- Add `--version` to `azd ai agent invoke` for hosted agent version selection.
- Create or reuse a version-backed session before remote responses/invocations calls.
- Isolate persisted session and conversation state by version and reject unsafe flag combinations.

## Testing
- `go test .\internal\cmd\... -count=1`
- `go test .\... -count=1`

Fixes #8054


wbreza (commented):
## Code Review — Automated Assist

Thanks for adding version targeting! A few findings to consider — posted as a non-blocking comment.

### 🔴 Blocking-level concerns

**1. `--version` + `--conversation-id` / `--new-conversation` are not rejected.** `validateInvokeVersionFlags` rejects `--local` and `--session-id` combos, but not the conversation flags. The PR description states "reject unsafe flag combinations" — this appears incomplete and is related to (2) below. Suggest adding explicit rejection (or, alternatively, plumbing the flags through to the version-backed flow if they are intended to work together).

**2. Conversation state is not isolated by version.** Session keys correctly use `buildAgentKey(..., rc.version, ...)`, but the conversation-ID lookup in `responsesRemote()` still calls `legacyKeysForRemote(rc.name)` directly rather than `rc.legacyKeys()`. Result: running `--version 2` and then `--version 3` against the same agent will resolve to the **same** stored conversation ID, which contradicts the "isolate persisted session and conversation state by version" claim. Suggest routing conversation lookup through `rc.legacyKeys()` so the version segment participates in the key.

### 🟠 High

**3. `--new-session` is effectively ignored when a version-backed session already exists.** In the `rc.version != ""` branch of `resolveRemoteSessionID`, after `resolveStoredID` returns a non-empty `sid`, the function returns it without consulting `a.flags.newSession`. A user passing `--version 3 --new-session` on the second invocation will get the cached session. Suggest gating the early return on `sid != "" && !a.flags.newSession`.

**4. Missing test coverage for the isolation claim.** New tests cover session creation and the new validation errors, but nothing exercises (a) cross-version conversation isolation or (b) rejection of `--version` + conversation flags (tied to findings 1 and 2). A small table-driven test using the existing fake stores would close this.

### 🟡 Medium

**5. Version-string validation is quite permissive.** `validateKeySegment` only rejects `/` and `\`. A value like `3?api-version=evil`, or an unreasonably long string, will be embedded into both the URL path segment and the on-disk storage key. Consider a stricter character class (e.g. `^[A-Za-z0-9._-]+$`) or document the contract and rely on server-side rejection.

**6. Minor duplication.** `rc.version` is assigned in both the ephemeral and project-mode branches of `resolveRemoteContext`. Not a bug, but easy to drift — consider hoisting it.

---

_Generated with assistance from an automated reviewer; please apply judgment. Defaults look preserved and the version-session creation path itself reads cleanly — the findings above are mostly around completeness of the "isolation + rejection" contract stated in the PR description._

glharper:
Addressed wbreza's review feedback in follow-up commit `a1ae1fb5f`:

- Reject `--version` with explicit `--conversation-id` so callers do not attach arbitrary cross-version conversation state.
- Keep `--new-conversation` supported as the safe reset path, with test coverage.
- Tighten `--version` validation to letters, numbers, `.`, `_`, and `-` with a 128-character limit.
- Make `--version --new-session` explicitly skip cached version-backed sessions.
- Add tests for versioned conversation lookup isolation and cached-session reset behavior.

Validation:
- `go test .\internal\cmd\... -count=1`
- `go test .\... -count=1`

cli/azd/extensions/azure.ai.agents/internal/cmd/invoke.go
```diff
@@ -193,10 +207,85 @@ session automatically. Pass --new-session to force a reset.`,
...
+			"cannot use --version with --session-id; sessions are already bound to a version when created",
+			"use --version without --session-id to create or reuse a version-backed session, "+
+				"or use --session-id without --version to invoke an existing session",
+		)
+	}
+	if flags.conversation != "" {
+		return exterrors.Validation(
+			exterrors.CodeInvalidParameter,
+			"cannot use --version with --conversation-id; conversations can contain version-specific context",
+			"use --version without --conversation-id to create or reuse a version-specific conversation, "+
```
trangevi: I don't understand this; how can a user reuse a version-specific conversation, if they can't specify the conversation-id?

If conversations are version-specific, then we should allow the conversation to be provided and either have a way to check on our end, or expect the backend to check, that the conversation does correlate with the provided version.

If conversations are not version-specific, then I see what is meant by them containing version-specific context, but I still think we shouldn't block on it, otherwise how could a user revisit a prior conversation on an older version (as an example)?

cli/azd/extensions/azure.ai.agents/internal/cmd/invoke.go
```diff
@@ -434,15 +525,19 @@ type remoteContext struct {
...
+	rc := &remoteContext{apiVersion: DefaultAgentAPIVersion, version: a.flags.version}
 
 	if a.endpoint != nil {
 		rc.name = a.endpoint.AgentName
 		rc.projectEndpoint = a.endpoint.ProjectEndpoint
 		if a.endpoint.APIVersion != "" {
 			rc.apiVersion = a.endpoint.APIVersion
 		}
-		rc.agentKey = buildAgentKey(a.endpoint.ProjectEndpoint, a.endpoint.AgentName, "", false)
+		if rc.version != "" {
```
trangevi: NIT: this if/else is unnecessary, since buildAgentKey handles an emptry string, it doesn't matter if rc.version is empty.

trangevi (changes requested):

wbreza (commented):
## Re-Review — commit `a1ae1fb`

Thanks for the quick turnaround — verified against PR HEAD. All six findings from my previous review are resolved:

| # | Prior finding | Status |
|---|---|---|
| 1 | `--version` + conversation flags not rejected | ✅ `--conversation-id` rejected; `--new-conversation` explicitly allowed and tested |
| 2 | Conversation state not isolated by version | ✅ `rc.legacyKeys()` returns `nil` when `rc.version != ""`; both session (line 603) and conversation (line 763) call sites now use `rc.legacyKeys()...` |
| 3 | `--new-session` ignored on cached version session | ✅ Gated on `!a.flags.newSession`; covered by `TestResolveRemoteSessionID_NewSessionSkipsCachedVersionSession` |
| 4 | Missing test coverage | ✅ Added: conversation-id rejection, oversized/bad-char rejection, cached-session reuse, new-session cache skip, allow-new-conversation, `validateInvokeVersionValue` table test, and versioned conversation lookup isolation |
| 5 | Permissive version validation | ✅ 128-char limit + `[A-Za-z0-9._-]` charset via `validateInvokeVersionValue` |
| 6 | Duplicate `rc.version` assignment | ✅ Hoisted into the `remoteContext` struct literal |

No new issues introduced by the delta. Note @trangevi's outstanding inline comments on this same commit — they should be addressed independently.

_LGTM from this automated reviewer's perspective on the items raised previously._

glharper:

cli/azd/extensions/azure.ai.agents/internal/cmd/invoke.go
glharper: [Updated](https://github.com/Azure/azure-dev/pull/8233/commits/4d71a742158605476cbf5d7eafa609460f20bf2e)

cli/azd/extensions/azure.ai.agents/internal/cmd/invoke.go
glharper: [Updated](https://github.com/Azure/azure-dev/pull/8233/commits/4d71a742158605476cbf5d7eafa609460f20bf2e)

wbreza (approved):
## Re-Review — commit `4d71a74` ✅

Both of @trangevi's inline comments are addressed:

- **Removed the `--version` + `--conversation-id` rejection.** Users can now resume a specific conversation against a specific version, which was the legitimate UX concern. Conversation persistence remains version-scoped through the versioned `agentKey`.
- **Collapsed the redundant `buildAgentKey` if/else** per the NIT.

The new `TestVersionedExplicitConversationPersistsUnderVersionKey` is a nice touch — it actually decodes the persisted user-config blob and asserts the conversation ID is stored under the versioned key, not the legacy bare-name key. Genuine end-to-end coverage of the isolation contract.

No new findings from my end. LGTM.

trangevi (approved):

PR #8241, opened 2026-05-19
feat(routines): add azd ai routine commands

### LINKED ISSUE #8159: Add `azd ai routine create/update/show/list/delete` (plus `enable`, `disable`, `dispatch`, and `run` subcommands) to manage and run Foundry Routines from any directory

huimiu:
## Summary

Implements `azd ai routine <verb>` in the `azure.ai.routines` extension, aligned with the Foundry Routines TypeSpec ([azure-rest-api-specs#43186](https://github.com/Azure/azure-rest-api-specs/pull/43186)).

A routine pairs one trigger (when) with one action (what) on a Foundry project — e.g. *"at 8 AM UTC every day, invoke the `daily-report` agent"*.

Closes #8159

## Commands

| Command | Notes |
|---|---|
| `routine create <name>` | `--trigger timer\|recurring` or `--file` manifest; `--force` upsert |
| `routine update <name>` | GET-then-PUT merge; immutable trigger/action type-switch guard |
| `routine show` / `list` / `delete` | Standard CRUD |
| `routine enable` / `disable` | Idempotent; dedicated `:enable` / `:disable` routes |
| `routine dispatch <name>` | `--input`, `--async`; surfaces `dispatchId` + `actionCorrelationId` + `taskId` |
| `routine run list <name>` | Auto-paged; `--top`, `--filter` |

**Triggers**: `timer` (`--at`, `--time-zone`), `recurring` (`--cron`, `--time-zone`).
**Actions**: `agent-response` (`--agent-name` or `--agent-endpoint-id`) or `agent-invoke` (`--agent-endpoint-id`).

## Spec alignment

End-to-end verified against a live Foundry project endpoint:

- snake_case wire fields (`agent_name`, `agent_endpoint_id`, `cron_expression`, `time_zone`, `dispatch_id`, …)
- `POST /routines/{name}:enable` and `:disable`
- `POST /routines/{name}:dispatch_async`
- `Foundry-Features: Routines=V1Preview` preview header

## Deferred

| Item | Reason |
|---|---|
| `--trigger github-issue` | Service still rejects the spec's `github_issue_opened` discriminator; CLI surface stays gated. |
| `AgentsPagedResult<T>` envelope | Service still returns `value` + `nextLink` (routines) / `value` + `nextPageToken` (runs) rather than `data` / `last_id` / `has_more`. |

## Tests

`go build ./...`, `go vet ./...`, `go test ./...`, `gofmt`, `golangci-lint`, and `cspell` all pass. Manual probes against the live data plane confirm the spec routes/fields are honored end-to-end.


wbreza (commented):
## Code Review (automated, supplements Copilot pre-review)

### TL;DR

First v1 of the `azd ai routine` command surface in a new extension. Implements `create/update/show/list/delete/enable/disable/dispatch` plus `run list`, a 5-level endpoint resolver, structured `exterrors`, and a Foundry data-plane HTTP client. Code is well-organized and the design is sound. However: the HTTP client has zero tests, `enable/disable` are GET-then-PUT with no concurrency guard, and the PR description contradicts the actual implementation in two places (enable route, preview header name). Findings below are net-new — they do not duplicate Copilot's 6 pre-review comments (port stripping, `--enabled` masking manifest, `github-issue` trigger accepted, `boolStr` nil→"true", `applyUpdateFlags` using `fmt.Errorf`, unused `newRoutineCommand`), all of which look valid.

---

### 🔴 Critical

**C1 — PR description contradicts the implementation in two places** (`internal/pkg/routines/client.go`)

| PR description claims | Code actually does |
|---|---|
| "`routine enable/disable` — Calls dedicated `:enable` route (idempotent)" | `setEnabled` does `GetRoutine` → mutate `Enabled` → `PutRoutine`. The code comment itself states: *"The Foundry Routines API does not expose a dedicated :enable route; the client mutates the routine resource directly."* |
| "Every request includes `x-ms-foundry-features-opt-in: Routines=V1Preview`" | Header constant is `routinesPreviewHeader = "Foundry-Features"`, so the actual header sent is `Foundry-Features: Routines=V1Preview`. |

Two consequences:
1. Reviewers, docs writers, and future maintainers will form an incorrect mental model from the description.
2. If the Foundry service actually expects the `x-ms-foundry-features-opt-in` header name, the preview opt-in is silently broken and the server is ignoring it. **Please verify the server-side header name against the Foundry TypeSpec and fix either the code or the description.**

---

### 🟠 High

**H1 — `setEnabled` has a lost-update race and no `If-Match` guard** (`client.go`, `EnableRoutine`/`DisableRoutine`/`setEnabled`)

Because enable/disable is GET-then-PUT against the full routine resource, any concurrent writer (another CLI invocation, another user, a workspace tool) that mutates the routine between the GET and the PUT will be silently clobbered. There is no `If-Match`/ETag conditional request to detect the race.

Fix options: (a) capture `ETag` from the GET response and send `If-Match` on the PUT, returning a structured "conflict" error on 412; (b) document this race in user-facing help text and emit a warning when the GET observes `Enabled != desired`; (c) work with the Foundry team to add a true `:enable`/`:disable` endpoint (preferred long-term, matches the PR description).

**H2 — HTTP client (`internal/pkg/routines/client.go`) has zero tests**

The data-plane integration boundary — every CRUD/enable/disable/dispatch/list call — has no unit coverage. This means:
- Pagination loop termination (`nextLink` follow in `ListRoutines`, `pageToken` in `ListRoutineRuns`) is unverified.
- Same-origin checks on `nextLink` are unverified.
- Error envelope parsing for 401 / 403 / 404 / 409 / 5xx is unverified.
- Request body marshaling for trigger/action records is unverified end-to-end.
- The client-side `Top` cap in run-listing is not exercised.

The PR description acknowledges tests are "tracked as follow-up work." Given this is the wire boundary of a Preview API, at minimum `httptest.Server`-based happy paths for each method plus pagination termination should ship in v1. The current command-level tests only exercise pure helpers (`buildTrigger`, `buildAction`, `applyUpdateFlags`) and don't reach the client.

**H3 — `routine update` may report "no changes" when only `--file` is used** (`routine_update.go`, `routine_manifest.go applyUpdateFlags`)

`applyUpdateFlags` only increments its `changed` counter when individual CLI flags are set; the `--file` merge path does not contribute to that counter. The no-op early-return checks `changed == 0`, so `routine update <name> --file manifest.yaml` (no other flags) likely takes the no-op branch even though the manifest legitimately changed fields. Either snapshot/diff the routine before vs after the merge, or count merge-introduced changes.

---

### 🟡 Medium

**M1 — Credential ignores tenant for guest / multi-tenant users** (`internal/cmd/routine_helpers.go`)

`newRoutineClient` constructs `azidentity.NewAzureDeveloperCLICredential(&AzureDeveloperCLICredentialOptions{})` with no `TenantID`. Per `.github/instructions/extensions.instructions.md`, extensions must use the user-access tenant (`Subscription.UserTenantId`, not the resource tenant). For multi-tenant or guest users the default tenant from `azd auth` may not match the Foundry project's tenant, causing AADSTS errors. Plumb a tenant hint (e.g., from azd context or by extracting from the endpoint/project metadata) into the credential options.

**M2 — Pagination has no max-page safety cap** (`client.go`, `ListRoutines` / `ListRoutineRuns`)

Both methods loop while `NextLink`/`PageToken` is non-empty. A misbehaving or compromised service that returns a self-referential or cycling token will cause an unbounded loop that consumes CPU and memory (each appended page grows the result slice). Add a defensive page-count cap (e.g., 1000) and return a clear "pagination exceeded N pages" error.

**M3 — Unbounded `io.ReadAll` on response bodies** (`client.go`, JSON decode helper)

The decode helper reads the entire response body into memory without `io.LimitReader`. A misbehaving server returning a multi-GB body will OOM the process. Wrap with a sane cap (e.g., `io.LimitReader(body, 32<<20)`) before decoding.

**M4 — Token scope `https://ai.azure.com/.default` should be verified** (`client.go`, `NewClient`)

The Bearer token scope is hardcoded. If the Foundry Routines data plane requires a project-specific or different audience (`services.ai.azure.com`, a regional audience, or a Foundry-specific resource ID), token acquisition will succeed but the server may reject with 401/403. Verify against the Foundry TypeSpec and add a code comment justifying the chosen scope.

**M5 — `assert.Error` instead of `require.Error` in negative-path tests** (`routine_create_test.go`)

Several "expect error" tests use `assert.Error` followed by further dereferences of the result. If the implementation regresses and returns no error, the test logs a failure but continues executing on a nil/zero value, potentially panicking or producing a confusing secondary failure that hides the real cause. Switch to `require.Error`.

---

### 🟢 Low

- **L1** — `routine dispatch --async` help text ("Suppress descriptive output; useful for scripting") is ambiguous. Clarify that it prints only the dispatch ID and exits without polling.
- **L2** — Manifest file-format detection falls back to JSON for empty/unknown extensions, and the error message recommends only JSON, leaving YAML users guessing. Mention both formats in the error.
- **L3** — `routine run` subcommand help doesn't mention that `run show` / `run delete` are intentionally deferred. Add a one-line note in `Long`.
- **L4** — `printJSON` / table writers write directly to `os.Stdout`, making future output-assertion tests impossible without monkey-patching. Accept an `io.Writer` parameter.

---

### Notes (not findings)

- This PR lives in an **extension** with its own `go.mod`, so core-`azd` patterns (IoC container, `ActionDescriptor`, `ErrorWithSuggestion`) do **not** apply — the extension uses cobra directly.
- `new(expr)` is valid in **Go 1.26+** (see `cli/azd/AGENTS.md`).
- Adding `DisallowUnknownFields` to the JSON decoder is **not** recommended — it would break forward-compat with non-breaking Foundry API additions.

---

*Generated review · supplements Copilot inline comments · posted as non-blocking COMMENT.*

wbreza (commented):
## Re-review: commit `1ff54876` (`fix(routines): address PR review feedback`)

Scope: one new commit since my [previous review](https://github.com/Azure/azure-dev/pull/8241#pullrequestreview) — 6 files, +82/-9.

### ✅ Copilot's pre-review findings — all addressed correctly

| # | Issue | Fix |
|---|---|---|
| 1 | `endpoint.go` silently strips explicit ports | Explicit port now rejected with structured error + 2 new tests |
| 2 | `--enabled` default masks manifest value | Uses `cmd.Flags().Changed("enabled")`; default applied only after manifest merge |
| 3 | `--trigger github-issue` accepted but incomplete | Explicit rejection in `buildTrigger` with deferred-v1 message |
| 4 | `boolStr` returns `"true"` for nil pointer | Now returns `"unknown"` |
| 5 | `applyUpdateFlags` uses plain `fmt.Errorf` | All 6 sites converted to `exterrors.Validation(CodeInvalidParameter, …)` |

(Copilot's 6th finding — dead `newRoutineCommand` — was already resolved in earlier commit `b7d375b8`.)

### ❌ Still open from my prior review

These weren't in the scope of this commit, just noting so they don't get lost:

- **C1** — PR description still claims a dedicated `:enable` route and `x-ms-foundry-features-opt-in` header. The actual code does GET-then-PUT and sends `Foundry-Features`. Please reconcile description vs implementation (and verify the preview header name with the Foundry team — if the server expects `x-ms-foundry-features-opt-in`, opt-in is silently broken).
- **H1** — Lost-update race in `setEnabled` (no `If-Match`/ETag).
- **H2** — `internal/pkg/routines/client.go` still has zero tests.
- **H3** — `routine update --file` may falsely report "no changes" because file-merge doesn't increment the `changed` counter.
- **M1** — `AzureDeveloperCLICredential` constructed with no `TenantID`; will break for guest / multi-tenant users.

No new blocking issues introduced by this commit. Nice clean turnaround on the Copilot items.

wbreza (commented):
## Re-review of PR #8241 — feat(routines): implement azd ai routine commands

Reviewed since wbreza's May 20 review. Skipping items already flagged by prior reviews (endpoint port stripping, `body.Enabled` flag override, `buildTrigger` deferred `github-issue` alias, `boolStr` nil-pointer returning `"true"`, `applyUpdateFlags` plain `fmt.Errorf`).

### 🔴 High — Security / correctness

**1. `internal/pkg/routines/client.go` — Missing `Logging.AllowedHeaders` on `policy.ClientOptions`**
Sibling `azure.ai.agents` restricts logged headers to `MsCorrelationIdHeader` only. The routines client omits the `Logging` field entirely. Azure SDK default logging will include the `Authorization` header (bearer token) in pipeline debug logs.
*Fix:* mirror agents' `Logging: policy.LogOptions{AllowedHeaders: []string{azsdk.MsCorrelationIdHeader}}`.

**2. `internal/cmd/routine_dispatch.go` — Wrong op code on inner `GetRoutine` error**
When fetching the routine to validate `--input`, the error is wrapped with `exterrors.OpGetRoutine` instead of `exterrors.OpDispatchRoutine`. Telemetry / user-facing error shows `get_routine.*` when the user ran `dispatch`. Note the adjacent `IsNotFound` branch already uses `OpDispatchRoutine` — the inconsistency is the smell.

**3. `internal/exterrors/errors.go` — `IsCancellation` misses gRPC `codes.Canceled`**
Routines checks only `errors.Is(err, context.Canceled)`. Agents also checks `status.FromError(err).Code() == codes.Canceled`. Cancellations bubbling up from gRPC `azdext` calls are misclassified as internal errors instead of user-driven.
*Fix:* port the agents implementation verbatim.

### 🟡 Medium — Test coverage & design

**4. `internal/pkg/routines/client.go` (+390 LOC) — no unit tests**
The data-plane client (CRUD, pagination, enable/disable GET+PUT fallback, `dispatchAsync`, list runs, custom HTTP/2 transport, same-origin pagination guard) has zero tests. PR description defers "unit tests as follow-up", but the manifest parser (low-risk) gets tests while the mission-critical client does not. Suggest at minimum `httptest.NewServer`-based tests for: 2xx happy paths, 404/409/5xx, `nextLink` pagination, `validateSameOrigin` rejection, context cancellation.

**5. Command-layer coverage gaps**
No tests for `routine_update.go` (complex flag→file→default merge), `routine_delete`, `routine_enable`, `routine_disable`, `routine_dispatch`, `routine_list`, `routine_show`, `routine_helpers`. Agents extension test ratio is ~1:1; this PR is ~1:2.2.

**6. `internal/pkg/routines/client.go` — `runtime.NewResponseError(resp)` leaks raw response body into errors**
Used in ~5 spots. Service error bodies pass straight through. CLI commands re-wrap with `exterrors.ServiceFromAzure`, but the underlying message text is preserved and may surface in logs/output.
*Fix:* drain body, extract structured fields, return a controlled message — or document that callers must always wrap and never log the raw error.

**7. `internal/cmd/routine_manifest.go` — `os.ReadFile` errors all classified as `CodeFileNotFound`**
Permission-denied, is-a-directory, etc. surface to the user as "routine manifest file not found" with misleading remediation.
*Fix:* gate with `os.IsNotExist(err)`; fall back to a generic file-read error otherwise.

**8. `internal/exterrors/codes.go` — No semantic "not found" code**
404s become `get_routine.404` rather than e.g. `routine_not_found`. Hurts telemetry classification and stable error-handling contracts.
*Fix:* add `CodeRoutineNotFound` and use it in the `IsNotFound` branches.

### 🟢 Low / nit

**9. `internal/cmd/endpoint_test.go` — port edge cases under-tested**
The recently-added port rejection covers `:443` only. Add: non-default ports (`:8443`), IPv6+port (`https://[::1]:443/...`), whitespace-only input. The original bug evaded tests because they didn't exercise non-default ports.

**10. `internal/cmd/routine_update.go` — `_ = cmd.Flags().MarkHidden(...)` (×2)**
Silent error discard. Safe in practice, but `if err := ...; err != nil { ... }` or a brief comment is cleaner.

**11. `AGENTS.md` — "If this extension grows enough… introduce `internal/exterrors`"**
Phrased as a future option, but the package exists in this PR. Rewrite in present tense so future maintainers don't treat it as optional.

**12. `go.mod` — `armappservice/v2 v2.3.0 // indirect` added**
Indirect, likely a cascade from an azd-core dep bump. Worth a quick confirmation that this isn't pulling unwanted weight on the extension.

### ✅ Verified clean (no action)

- URL construction (`url.PathEscape` / `url.QueryEscape`)
- Pagination `validateSameOrigin`
- HTTP/2 timeout fix (explicit `ResponseHeaderTimeout`, `TLSHandshakeTimeout`) is sound, not masking
- Wire format alignment (camelCase tags, `agentId`, `:dispatchAsync`, GET+PUT enable/disable fallback) matches deployed Foundry service per AGENTS.md
- Bearer token scope `https://ai.azure.com/.default`
- Endpoint validation (HTTPS-only, port rejection, `.services.ai.azure.com` allowlist)
- `exterrors` duplication with `agents` / `toolboxes` is acknowledged in AGENTS.md as the established cross-extension pattern — not flagged

huimiu (commented):
*(Superseded by review below)*

huimiu (commented):
Fixed:

- `routine_dispatch.go`: `OpGetRoutine` → `OpDispatchRoutine` (wrong op code on inner error)
- `exterrors/errors.go`: `IsCancellation` now also checks gRPC `codes.Canceled` (ported from agents)
- `routine_manifest.go`: `os.ReadFile` errors now distinguish not-exist vs permission-denied
- `client.go`: added `Logging.AllowedHeaders` for `MsCorrelationIdHeader` (observability parity)
- `AGENTS.md`: exterrors section rewritten from future tense to present tense

trangevi (approved):
Filed https://github.com/Azure/azure-dev/issues/8365 for remaining non-blocking follow up items

wbreza (commented):
## Re-review #4 — delta since 2026-05-22

Two new commits (`4762d734` spec-alignment + `b1d23b90` review-feedback). Most of my prior round's findings are resolved, plus a couple of older items from the 2026-05-19 round were fixed as a side effect of the spec alignment. Verified at head `b1d23b909c1ed413a161e6cab6630d55124d3e9d`.

### ✅ Fixed in this round

From the 2026-05-22 review:
- **`client.go` — `Logging.AllowedHeaders` set** on `policy.ClientOptions` (bearer token no longer leaks to debug logs)
- **`routine_dispatch.go` — `OpDispatchRoutine` op code** used on the inner `GetRoutine` error
- **`exterrors/errors.go` — `IsCancellation`** now checks gRPC `codes.Canceled` in addition to `context.Canceled`
- **`routine_manifest.go` — `os.IsNotExist`** gate distinguishes not-found from permission/other read errors
- **`AGENTS.md`** — exterrors section rewritten in present tense

From the 2026-05-19 round (resolved as a side effect of spec alignment):
- **PR description ↔ code mismatch (`:enable` route, `Foundry-Features` header)** — code and description now agree
- **`setEnabled` lost-update race (GET+PUT without `If-Match`/ETag)** — replaced with `POST :enable` / `:disable` action routes; no read-modify-write cycle remains

### ⏳ Still open (carrying forward across rounds)

**Substantive (raised more than once or elevated risk):**

1. **No unit tests for `internal/pkg/routines/client.go`** (~390 LOC: CRUD, pagination, `:enable`/`:disable` actions, `:dispatch_async`, `decodeJSON`). Raised in the 2026-05-19 and 2026-05-22 rounds. The wire format just flipped camelCase → snake_case and three action routes changed in `4762d734`; there is no automated guard against the next such regression. Suggest `httptest.NewServer`-backed tests for at least: happy-path CRUD, 404 / 409 / 5xx, `NextLink` pagination termination, `validateSameOrigin` rejection, context cancellation.

2. **`AzureDeveloperCLICredential` constructed without `TenantID`.** `internal/cmd/context.go` does `azidentity.NewAzureDeveloperCLICredential(&azidentity.AzureDeveloperCLICredentialOptions{})` with no tenant set. Per the extension convention noted in this repo, code touching `Subscription` should use `Subscription.UserTenantId` (not `Subscription.TenantId`) for credential creation, so guest / multi-tenant users authenticate against the right tenant. Currently the cred falls back to whatever `azd auth login` last cached, which breaks the multi-tenant case.

3. **`runtime.NewResponseError(resp)`** is still used in ~5 spots in `client.go`. CLI commands re-wrap with `exterrors.ServiceFromAzure`, but the original error message preserves the raw response body — anything sensitive the service returns ends up in logs / user output. Suggest draining the body, decoding structured fields, and returning a controlled message.

4. **Command-layer test gaps remain**: `update`, `delete`, `enable`, `disable`, `dispatch`, `list`, `show`, `helpers` still have no tests. Existing tests were updated for the snake_case rename, but no new files were added.

5. **No semantic `CodeRoutineNotFound`** — 404s still surface as `get_routine.404` rather than a stable telemetry-friendly code.

**Bounds / safety (raised on 2026-05-19, still open):**

6. **`decodeJSON` does unbounded `io.ReadAll(resp.Body)`** — misbehaving or hostile service can OOM the process. Wrap with `io.LimitReader`.

7. **`ListRoutines` pagination has no max-page guard** — relies entirely on `NextLink == ""` to terminate. `ListRoutineRuns` correctly caps via `opts.Top`; `ListRoutines` should add an equivalent iteration / total-item ceiling.

**Low / nit:**

8. **`endpoint_test.go`** port edge cases — only `:443` is exercised; add IPv6+port, non-default explicit ports (`:8443`), and whitespace-only input.

9. **`routine_update.go`** silent `MarkHidden` error discard (×2) — `_ = cmd.Flags().MarkHidden(...)`; either check or comment.

10. **`go.mod`** indirect `armappservice/v2 v2.3.0` — quick confirmation that this is just an azd-core cascade and isn't pulling unwanted weight into the extension.

11. **`routine update --file`** — counter logic correctly increments via `overwriteRoutineFromFile`, but the "no changes" message can still mislead in the file-only path when the manifest matches the server state. Minor wording / branching cleanup.

### Verified clean (re-confirmed at head)

URL escaping · `validateSameOrigin` · HTTP/2 transport timeouts · wire format aligned to spec PR #43186 with the documented divergences (`github_issue` trigger, `value` + `nextLink` envelope) · bearer scope `https://ai.azure.com/.default` · endpoint allowlist (HTTPS-only, port rejection, `.services.ai.azure.com`)

vhvb1989 (approved):

PR #8247, opened 2026-05-19
[Foundry Extensions] Add new meta package files.

trangevi:
Add new meta package files. Update versions of new extensions to match agents pattern


wbreza (commented):
## Summary

Small, mechanical PR introducing the `azure.ai.foundry` meta-extension and aligning 6 sibling extensions to `0.1.0-preview` so the meta-package''s `~0.1.0-preview` dependency constraints resolve cleanly. No Go code touched. Risk is very low. Two things worth addressing before merge and one optional follow-up.

---

### 🟡 README is a copy-paste leftover from the demo extension

**File:** `cli/azd/extensions/azure.ai.foundry/README.md`

```markdown
# `azd` Demo Extension

A meta-extension, to encapsulate all of the Microsoft Foundry AZD extensions.
```

The title was copy-pasted from another scaffold. Every other AI extension in this repo follows `# Foundry <Thing>` (e.g., `azure.ai.connections/README.md` → `# Foundry Connections`). Since this README is user-facing on the extension registry, suggest something like:

```markdown
# Azure AI Foundry

A meta-extension that bundles all of the Microsoft Foundry `azd` extensions for easy install. (Preview)
```

---

### 🟡 `metadata` capability on a dependencies-only extension will log a warning at install time

**File:** `cli/azd/extensions/azure.ai.foundry/extension.yaml`

```yaml
capabilities:
    - metadata
```

The schema requires `capabilities` to have `minItems: 1`, so a meta-package must declare something — but declaring `metadata` without an `entryPoint`/binary means `manager.fetchAndCacheMetadata` (`cli/azd/pkg/extensions/manager.go:549–554`, `872–910`) will try to invoke a non-existent executable and fall through to `runner.go:51–53` with `"extension path '''' not found"`. The error is non-fatal (`manager.go:552` logs a warning and continues), so install succeeds — but every user installing the meta-package will see that warning.

A few directions:

- **(A)** Accept the warning as a known quirk of using `metadata` for a pure dependency aggregator.
- **(B)** Suppress the warning in `fetchAndCacheMetadata` when the extension declares no artifacts/entryPoint (small follow-up in azd core).
- **(C)** Introduce a new `meta-package` / `dependencies-only` capability that explicitly opts out of metadata fetch — cleanest long-term fix, out of scope here.

The internal `azd.internal.pack` extension sidesteps this by omitting `capabilities` entirely, which the schema technically forbids but works because internal extensions bypass schema validation. External meta-packages don''t have that escape hatch today.

Worth confirming the warning is intentional and acceptable for the first ship.

---

### 🟢 `CHANGELOG.md` entries for bumped extensions still say "Initial Version"

For the six bumped extensions, the CHANGELOG just rewrites the version number on the same line:

```diff
-## 0.0.1-preview - Initial Version
+## 0.1.0-preview - Initial Version
```

Since `0.0.1-preview` was never published, treating `0.1.0-preview` as the new "Initial Version" is defensible — but it erases any history that a `0.0.1-preview` ever existed. If the team wants the changelog to read as a true release log, consider either keeping both lines or adding a one-liner noting the version-scheme alignment with the agents extension. Not a blocker.

---

### ✅ Verified

- **Tilde + prerelease semver constraints work.** Ran Masterminds/semver/v3 (the resolver used by `cli/azd/pkg/extensions/manager.go:356`) against `~0.1.0-preview` and confirmed it matches `0.1.0-preview`, `0.1.32-preview`, and `0.1.99-preview`, and rejects `0.2.0-preview`. So the dependency on `azure.ai.agents@~0.1.0-preview` resolves cleanly against the current `0.1.32-preview`.
- **`metadata` is a valid schema enum value.**
- **All 7 declared dependencies exist** in `cli/azd/extensions/`.

wbreza (commented):
## Re-review — commit `b06d15b4`

### Summary

Only one change since my prior review: a 1-line README heading fix in response to the `copilot-pull-request-reviewer` comment.

```diff
- # `azd` Demo Extension
+ # `azd` Azure AI Foundry Extension
```

Risk: very low. No code paths affected.

### Status of prior findings

| Prior finding | Status |
|---|---|
| 🟡 README copy-paste title | ✅ **Resolved** — new heading is consistent with the `displayName` in `extension.yaml`. |
| 🟡 `metadata` capability on a dependencies-only extension logs a warning at install time (`manager.go:549–554`, `872–910` → `runner.go:51–53`) | 🟡 **Open** — not addressed. Reasonable to ship as-is if you''ve decided to accept the warning; worth a one-line confirmation of intent. |
| 🟢 CHANGELOG entries still labeled "Initial Version" after the version bump | 🟢 **Open** — non-blocker. |

### New findings from this commit

None. Single-line markdown heading change with zero risk.

Thanks for the quick fix on the README — happy to defer the remaining two items.

wbreza (approved):
LGTM. ✅

The `extension.yaml` simplification is the cleanest possible fix for the prior `metadata`-capability concern — verified:

- Schema validation isn''t enforced at runtime (`ValidateAgainstSchema` in `pkg/extensions/schema_validator.go` is test-only).
- `Capabilities` is `omitempty` in the Go struct (`pkg/extensions/extension.go:20`, `registry.go:119`).
- No install-time check requires capabilities (`pkg/extensions/manager.go:386–390`, `518–530`).
- `azd.internal.pack` is the existing precedent for capabilities-less meta-extensions.

So no metadata fetch is attempted at install → no warning logged.

### One follow-up worth noting (not blocking, separate PR)

`cli/azd/extensions/extension.schema.json` still declares `capabilities` as `required` with `minItems: 1`. The runtime ignores the schema so this isn''t a functional bug, but editors honoring the `$schema` directive at the top of `extension.yaml` will flag the file. Worth either making `capabilities` optional in the schema, or adding a `meta-package` capability enum so dependency-only extensions can self-declare.

The "Initial Version" wording in the bumped CHANGELOG files remains non-blocking — defer.

vhvb1989 (approved):

PR #8259, opened 2026-05-19
Add support for updating dev registry in extension release pipelines

### LINKED ISSUE #8179: Automate extension registry update as part of release pipeline
Currently, every new version release of an extension requires at least 2 PRs, one to bump the version and changelog, and another to update `registry.json` after the release pipeline completes.

Automating the registry update, similar to the "Increment CLI version after release" PRs, would reduce manual churn for extension devs.

JeffreyCA:
Addresses #8179

Follow-up to #8167 to be able to automatically publish to and update the dev registry (`registry.dev.json`) instead of `registry.json`.

Note: Publishing to the dev registry still creates the same public GitHub release tag and storage artifacts as a production registry publish. Use a prerelease/dev version for dev-registry publishes to avoid colliding with a later production publish of the same version.

### Validation

Tested pipeline on [ext-dev-registry](https://github.com/Azure/azure-dev/tree/ext-dev-registry) branch with Skip.Publish=true confirmed azure-sdk bot auto-submitted PR with expected changes: https://github.com/Azure/azure-dev/pull/8253

image

### DIFF
--- a/cli/azd/pkg/extensions/registry_files_test.go
+++ b/cli/azd/pkg/extensions/registry_files_test.go
@@ -20,6 +20,7 @@ func TestDevRegistryFileIsValid(t *testing.T) {
 	var registry Registry
 	require.NoError(t, json.Unmarshal(data, &registry))
 	require.Equal(t, CurrentRegistrySchemaVersion, registry.SchemaVersion)
+	require.Empty(t, registry.Extensions)
 
 	result := ValidateRegistry(&registry, false)
 	require.True(t, result.Valid, "registry.dev.json failed validation: %+v", result)
--- a/eng/pipelines/release-ext-azure-ai-agents.yml
+++ b/eng/pipelines/release-ext-azure-ai-agents.yml
@@ -22,6 +22,12 @@ pr:
     exclude:
       - cli/azd/docs/**
 
+parameters:
+  - name: PublishToDevRegistry
+    displayName: Publish to dev registry
+    type: boolean
+    default: false
+
 extends:
   template: /eng/pipelines/templates/stages/1es-redirect.yml
   parameters:
@@ -31,3 +37,4 @@ extends:
           AzdExtensionId: azure.ai.agents
           SanitizedExtensionId: azure-ai-agents
           AzdExtensionDirectory: cli/azd/extensions/azure.ai.agents
+          PublishToDevRegistry: ${{ parameters.PublishToDevRegistry }}
--- a/eng/pipelines/release-ext-azure-ai-connections.yml
+++ b/eng/pipelines/release-ext-azure-ai-connections.yml
@@ -22,6 +22,12 @@ pr:
     exclude:
       - cli/azd/docs/**
 
+parameters:
+  - name: PublishToDevRegistry
+    displayName: Publish to dev registry
+    type: boolean
+    default: false
+
 extends:
   template: /eng/pipelines/templates/stages/1es-redirect.yml
   parameters:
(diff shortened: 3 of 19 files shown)


vhvb1989 (approved):

wbreza (commented):
Thanks for the follow-up to #8167 — the threading across all 15 per-extension pipelines and the wrapper template is clean and consistent, and the default-off conditional keeps existing pipelines fully insulated. A few things worth considering before merge:

### 🟡 Update_Registry job runs unconditionally for dev publishes

In ng/pipelines/templates/stages/publish-extension.yml, the Update_Registry job still executes its Refresh Fig/Usage snapshots step (go test ./cmd -run 'TestFigSpec|TestUsage' with UPDATE_SNAPSHOTS=true) regardless of PublishToDevRegistry. Those snapshots are global fixtures, not per-registry — so a dev-registry publish can produce snapshot diffs that end up in the dev-registry PR. If such a PR ever merges, it would pollute the snapshots that production-registry CI depends on.

Two reasonable options:
- Gate the snapshot-refresh step (or the whole Update_Registry job) on ${{ parameters.PublishToDevRegistry }} == false so dev publishes only update 
egistry.dev.json and skip the snapshot regeneration entirely.
- Or, if the snapshots are genuinely registry-independent (i.e., TestFigSpec/TestUsage don't read the registry file at all), add a brief comment explaining why the refresh is safe under both modes so future maintainers don't have to re-derive that.

### 🟡 Release tag is shared between dev and prod publishes

Publish_Release is not parameterized — it always creates the public GitHub release tagged zd-ext-<ext>_<version> and uploads artifacts to the production storage location. So a "dev registry publish" still consumes a real version tag and ships public binaries; only the registry pointer differs.

This is probably the intended design (dev registry is a publish-channel choice, not an isolation mechanism), but it has a sharp edge: if someone runs the dev pipeline with a version that later gets reused for a prod publish, gh release create will collide on the existing tag. A short note in the PR description, or a one-line displayName hint on the dev path (e.g., reminding operators to use a pre-release suffix like -alpha/-dev), would prevent that footgun.

### 🔵 Nit — TestDevRegistryFileIsValid could assert empty extensions

In cli/azd/pkg/extensions/registry_files_test.go, consider adding 
equire.Empty(t, registry.Extensions) to document the intent that 
egistry.dev.json ships empty as a seed. Optional.

### Verified clean
- Parameter threading across all 15 per-extension pipelines → 
elease-azd-extension.yml → publish-extension.yml is correct and consistent
- cli/azd/extensions/registry.dev.json exists and parses against the schema
- zd x publish --registry <path> accepts the new dev-registry path
- The comment on the Update_Registry job was updated to reflect the generalization
- Distinct PR branch prefix (xt-dev-registry-update/) and title prevent collisions with production registry-update PRs

Nice clean implementation overall — the conditional approach is the right call here.

vhvb1989 (approved):

JeffreyCA:
Addressed the feedback: dev-registry publishes now skip the global Fig/Usage snapshot refresh, the PR description calls out that dev publishes still consume the same public release tag/artifacts as prod publishes, and TestDevRegistryFileIsValid now asserts the dev registry seed is empty.

wbreza (approved):
Thanks for the thorough follow-up — all three points from my previous review are addressed cleanly:

- **Snapshot refresh** is now gated on ${{ if ne(parameters.PublishToDevRegistry, true) }} in publish-extension.yml, so dev publishes won't regenerate the global Fig/Usage snapshots.
- **Shared release tag** is called out in the PR description so operators know to use prerelease suffixes (-alpha/-dev) to avoid collisions on re-publish.
- **Test seed** — TestDevRegistryFileIsValid now asserts 
egistry.Extensions is empty.

The latest commit (464ee2e) extending the same PublishToDevRegistry parameter to 
elease-ext-azure-ai-training.yml is structurally identical to the other 15 per-extension pipelines and matches the established pattern.

LGTM 🚀

PR #8266, opened 2026-05-20
add 'azd ai agent sample list' to browse templates and improve non-interactive init

### LINKED ISSUE #8308: Add `azd ai agent sample list` for copilot assistence

therealjohn:
This pull request introduces several improvements and new features to the Azure AI Agents CLI extension, focusing on safer headless operation, better error messages for non-interactive usage, and improved usability for sample discovery. The most significant changes include the addition of a `--force` flag to enable safe overwrites in headless mode, actionable error messages for required prompts, a new `sample` command namespace for browsing agent templates, and codebase improvements for index handling and test coverage.

**Headless and Prompt Handling Improvements**
- Added a `--force` flag to the `init` command, allowing headless (`--no-prompt`) callers to explicitly consent to overwriting files that would otherwise require confirmation. This prevents accidental data loss and provides clear instructions in error messages when overwriting is not permitted without `--force`. [[1]](diffhunk://#diff-101203523896c0113fdfeb6552cf8cdf37105e291ab9c312999e16fa2f154234R56-R60) [[2]](diffhunk://#diff-101203523896c0113fdfeb6552cf8cdf37105e291ab9c312999e16fa2f154234R841-R844) [[3]](diffhunk://#diff-101203523896c0113fdfeb6552cf8cdf37105e291ab9c312999e16fa2f154234R1594-R1611) [[4]](diffhunk://#diff-101203523896c0113fdfeb6552cf8cdf37105e291ab9c312999e16fa2f154234R1626) [[5]](diffhunk://#diff-638d5c2f7cb5f6df9e63286eb0f6d9f4d132703a9982c5984ef037071af4e697L85-R89) [[6]](diffhunk://#diff-638d5c2f7cb5f6df9e63286eb0f6d9f4d132703a9982c5984ef037071af4e697R324-R333) [[7]](diffhunk://#diff-638d5c2f7cb5f6df9e63286eb0f6d9f4d132703a9982c5984ef037071af4e697R359) [[8]](diffhunk://#diff-a12c2ad24b6ef2c979758b7ed7163937d9ef2a3849ff9692cb30e0b6a442e0daR2212-R2256)
- Improved error messages for headless scenarios (e.g., missing Azure subscription, location, or model selection) to include actionable suggestions, such as setting environment variables or passing specific flags. [[1]](diffhunk://#diff-46e6dfc85715f4b765bd87dc4e2d258ae6a74dcaf785775e3b0dc9cf8590d5acR679-R683) [[2]](diffhunk://#diff-46e6dfc85715f4b765bd87dc4e2d258ae6a74dcaf785775e3b0dc9cf8590d5acL694-R716) [[3]](diffhunk://#diff-46e6dfc85715f4b765bd87dc4e2d258ae6a74dcaf785775e3b0dc9cf8590d5acL823-R853) [[4]](diffhunk://#diff-46e6dfc85715f4b765bd87dc4e2d258ae6a74dcaf785775e3b0dc9cf8590d5acL874-R912) [[5]](diffhunk://#diff-46e6dfc85715f4b765bd87dc4e2d258ae6a74dcaf785775e3b0dc9cf8590d5acL1093-R1136) [[6]](diffhunk://#diff-12cef9dd3710b6cc16c77b6c23a4e83c3f3a8ad52fad0bd46c27313a7dcf25d4L234-R235)

**New Features**
- Introduced a new `azd ai agent sample` command namespace, currently with a `list` subcommand, to allow users to browse curated agent samples and templates. This helps both interactive and scripted users discover available options for agent initialization. [[1]](diffhunk://#diff-3dc941afbde6fb8ecfa717534de20f71396d7e1ba07b1e2eb61f7ef80dae3f17R66) [[2]](diffhunk://#diff-05f827d9de898a4645a722cc119ab9e711703aa0fa74c8b4e8fcd933a48dffeeR1-R34)

**Codebase and Test Improvements**
- Centralized and documented the conversion of slice indices to `int32` for prompt APIs with a new `boundedInt32Index` helper, ensuring safety and reducing code duplication. Updated all relevant locations to use this helper. [[1]](diffhunk://#diff-5f182f40d4c72a13c0f75ba1ad7ed4451bb4eaea011dbd802e25ae3a3b146318R1-R28) [[2]](diffhunk://#diff-101203523896c0113fdfeb6552cf8cdf37105e291ab9c312999e16fa2f154234L2243-R2271) [[3]](diffhunk://#diff-638d5c2f7cb5f6df9e63286eb0f6d9f4d132703a9982c5984ef037071af4e697L787-R808) [[4]](diffhunk://#diff-12cef9dd3710b6cc16c77b6c23a4e83c3f3a8ad52fad0bd46c27313a7dcf25d4L352-R353) [[5]](diffhunk://#diff-bd59534c3f4023c2639c0421684fff0e5aa83ee2c33be97bc7234117fc44a9f6L132-R132)
- Added and enhanced tests to cover new behaviors, including the `--force` flag registration and correct error handling for manifest overwrites in headless mode. [[1]](diffhunk://#diff-a12c2ad24b6ef2c979758b7ed7163937d9ef2a3849ff9692cb30e0b6a442e0daR39-R53) [[2]](diffhunk://#diff-a12c2ad24b6ef2c979758b7ed7163937d9ef2a3849ff9692cb30e0b6a442e0daR2212-R2256)

These changes collectively improve the robustness, safety, and usability of the CLI extension for both interactive and automated workflows.

### DIFF
--- a/cli/azd/extensions/azure.ai.agents/internal/cmd/init.go
+++ b/cli/azd/extensions/azure.ai.agents/internal/cmd/init.go
@@ -53,6 +53,11 @@ type initFlags struct {
 	src               string
 	env               string
 	protocols         []string
+	// force, when true, lets headless callers (--no-prompt) pre-consent to
+	// overwrite prompts that would otherwise return a structured error. It
+	// mirrors the `--force` convention used by `azd down`, `azd env remove`,
+	// `azd config reset`, and `azd infra generate`.
+	force bool
 	// noPrompt is resolved from the extension context (--no-prompt / AZD_NO_PROMPT)
 	// and is not registered as a CLI flag on the init command itself.
 	noPrompt bool
@@ -833,11 +838,9 @@ a reusable sample or manifest.`,
 	cmd.Flags().StringSliceVar(&flags.protocols, "protocol", nil,
 		"Protocols supported by the agent (e.g., 'responses', 'invocations'). Can be specified multiple times.")
 
-	// Subcommand for catalog discovery. Lets coding agents enumerate manifest
-	// URLs they can pass back into `azd ai agent init -m <url>` without
-	// scraping documentation. Cobra resolves subcommands before positional
-	// args, so this does not conflict with the `init [<path>]` positional.
-	cmd.AddCommand(newInitListCommand(extCtx))
+	cmd.Flags().BoolVar(&flags.force, "force", false,
+		"Overwrite an input manifest that already lives inside the generated src tree without prompting. "+
+			"Required together with --no-prompt when init would otherwise need confirmation.")
 
 	return cmd
 }
@@ -1588,30 +1591,36 @@ func (a *InitAction) downloadAgentYaml(
 
 			// Check if manifest is under src directory
 			if isSubpath(absManifestPath, srcDir) {
-				// In no-prompt mode, refuse to silently overwrite a manifest
-				// that lives inside the project's src tree. Headless callers
-				// can move the manifest, choose a different --src target, or
-				// run interactively to confirm.
-				if a.flags.noPrompt {
+				// `--force` is the explicit pre-consent path for headless
+				// callers: skip both the prompt and the no-prompt refusal,
+				// and accept the overwrite. We log the decision so debug
+				// output makes the choice visible in CI logs.
+				if a.flags.force {
+					log.Printf("--force: overwriting manifest %q inside src directory %q", manifestPointer, srcDir)
+				} else if a.flags.noPrompt {
+					// In no-prompt mode, refuse to silently overwrite a manifest
+					// that lives inside the project's src tree. Headless callers
+					// can pass --force to pre-consent, move the manifest, choose
+					// a different --src target, or run interactively to confirm.
 					return nil, "", exterrors.Validation(
 						exterrors.CodeInvalidManifestPointer,
 						fmt.Sprintf("manifest %q is inside the project src directory %q and would be overwritten", manifestPointer, srcDir),
-						"move the manifest outside the src tree, pass a different --src directory, "+
-							"or run interactively to confirm the overwrite",
+						"pass --force to overwrite, move the manifest outside the src tree, "+
+							"pass a different --src directory, or run interactively to confirm the overwrite",
 					)
-				}
-
-				confirmResponse, err := a.azdClient.Prompt().Confirm(ctx, &azdext.ConfirmRequest{
-					Options: &azdext.ConfirmOptions{
-						Message:      "This operation will overwrite the provided manifest file. Continue?",
-						DefaultValue: new(false),
-					},
-				})
-				if err != nil {
-					return nil, "", fmt.Errorf("prompting for confirmation: %w", err)
-				}
-				if !*confirmResponse.Value {
-					return nil, "", exterrors.Cancelled("operation cancelled by user")
+				} else {
+					confirmResponse, err := a.azdClient.Prompt().Confirm(ctx, &azdext.ConfirmRequest{
+						Options: &azdext.ConfirmOptions{
+							Message:      "This operation will overwrite the provided manifest file. Continue?",
+							DefaultValue: new(false),
+						},
+					})
+					if err != nil {
+						return nil, "", fmt.Errorf("prompting for confirmation: %w", err)
+					}
+					if !*confirmResponse.Value {
+						return nil, "", exterrors.Cancelled("operation cancelled by user")
+					}
 				}
 			}
 		}
@@ -2259,7 +2268,7 @@ func (a *InitAction) populateContainerSettings(
 	if manifestResources != nil {
 		for i, t := range project.ResourceTiers {
 			if t.Cpu == manifestResources.Cpu && t.Memory == manifestResources.Memory {
-				defaultIndex = int32(i) //nolint:gosec // resource tier list length is fixed and small
+				defaultIndex = boundedInt32Index(i)
 				break
 			}
 		}
--- a/cli/azd/extensions/azure.ai.agents/internal/cmd/init.go
+++ b/cli/azd/extensions/azure.ai.agents/internal/cmd/init.go
@@ -2259,7 +2259,7 @@ func (a *InitAction) populateContainerSettings(
 	if manifestResources != nil {
 		for i, t := range project.ResourceTiers {
 			if t.Cpu == manifestResources.Cpu && t.Memory == manifestResources.Memory {
-				defaultIndex = int32(i)
+				defaultIndex = int32(i) //nolint:gosec // resource tier list length is fixed and small
 				break
 			}
 		}
--- a/cli/azd/extensions/azure.ai.agents/internal/cmd/init.go
+++ b/cli/azd/extensions/azure.ai.agents/internal/cmd/init.go
@@ -833,6 +833,12 @@ a reusable sample or manifest.`,
 	cmd.Flags().StringSliceVar(&flags.protocols, "protocol", nil,
 		"Protocols supported by the agent (e.g., 'responses', 'invocations'). Can be specified multiple times.")
 
+	// Subcommand for catalog discovery. Lets coding agents enumerate manifest
+	// URLs they can pass back into `azd ai agent init -m <url>` without
+	// scraping documentation. Cobra resolves subcommands before positional
+	// args, so this does not conflict with the `init [<path>]` positional.
+	cmd.AddCommand(newInitListCommand(extCtx))
+
 	return cmd
 }
 
@@ -1582,6 +1588,19 @@ func (a *InitAction) downloadAgentYaml(
 
 			// Check if manifest is under src directory
 			if isSubpath(absManifestPath, srcDir) {
+				// In no-prompt mode, refuse to silently overwrite a manifest
+				// that lives inside the project's src tree. Headless callers
+				// can move the manifest, choose a different --src target, or
+				// run interactively to confirm.
+				if a.flags.noPrompt {
+					return nil, "", exterrors.Validation(
+						exterrors.CodeInvalidManifestPointer,
+						fmt.Sprintf("manifest %q is inside the project src directory %q and would be overwritten", manifestPointer, srcDir),
+						"move the manifest outside the src tree, pass a different --src directory, "+
+							"or run interactively to confirm the overwrite",
+					)
+				}
+
 				confirmResponse, err := a.azdClient.Prompt().Confirm(ctx, &azdext.ConfirmRequest{
 					Options: &azdext.ConfirmOptions{
 						Message:      "This operation will overwrite the provided manifest file. Continue?",
--- a/cli/azd/extensions/azure.ai.agents/internal/cmd/init_list.go
+++ b/cli/azd/extensions/azure.ai.agents/internal/cmd/init_list.go
@@ -6,11 +6,11 @@ package cmd
 import (
 	"encoding/json"
 	"fmt"
+	"io"
 	"net/http"
 	"os"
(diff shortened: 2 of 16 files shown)


wbreza (commented):
Nice, focused PR — `init list` fills a real gap for non-interactive callers, and the wrapped errors with explicit env-var/flag names are a meaningful UX win. CI is green and the prior Copilot bot review covered the most material concerns (stdout vs `cmd.OutOrStdout`, overbroad subscription error wrap, stdout-mixing status print, repeated `nolint:gosec`, byte→string alloc in JSON write) — worth taking those.

Three additional findings beyond the bot's review:

---

### MEDIUM — `initCommand` JSON field is not shell-safe

`init_list.go`, in `mapAgentTemplateToDTO`:

```go
item.InitCommand = fmt.Sprintf("azd ai agent init -m %s", t.Source)
```

The PR markets `initCommand` as a stable, "ready to paste" string for coding agents. Today's catalog `Source` values are clean GitHub URLs, but if a future catalog entry ever contains a space, `&`, `|`, or `;`, the paste-ready string silently breaks — or, worse, becomes a command-injection surface for agents that shell-exec it without re-quoting.

Options:
- Quote at construction: `fmt.Sprintf("azd ai agent init -m %q", t.Source)`
- Validate `t.Source` against an allow-list of safe characters in `mapAgentTemplateToDTO` and return a structured error if it fails
- Document on `TemplateListItem` that consumers MUST shell-quote `initCommand` segments before exec — though that arguably defeats the "ready to paste" framing

Same concern applies to the `azd init -t %s` branch for the `azd` template type.

---

### MEDIUM — `ManifestURL` / `RepoURL` discriminator invariant is undocumented

`TemplateListItem` has `omitempty` on both `ManifestURL` and `RepoURL`. The implicit contract is "exactly one is populated, switch on `Type`," but neither the struct doc nor any test asserts this invariant. Coding agents consuming the JSON have no schema-level signal for which field to read.

Suggested:

1. Add a struct-level doc comment, e.g.:
   ```
   // TemplateListItem describes a single catalog entry. Exactly one of
   // ManifestURL or RepoURL is populated; use Type to discriminate:
   //   Type == "agent" -> ManifestURL is set
   //   Type == "azd"   -> RepoURL is set
   ```
2. Add a test in `init_list_test.go` that walks the mapping for both `TemplateTypeAgent` and `TemplateTypeAzd` and asserts the corresponding URL is non-empty and the other is empty.

---

### LOW — `fetchAgentTemplates` error message omits the URL

In `init_list.go` (~L146), the primary error message reads `"failed to fetch agent templates catalog: <err>"` and the URL appears only in the suggestion. Including the URL in the primary message itself (e.g., `"failed to fetch agent templates catalog from %s: %s"`) makes copy-paste triage faster, since `agentTemplatesURL` isn't otherwise visible to end users.

---

### Notes

- `new(fmt.Sprintf(...))` at `init_foundry_resources_helpers.go:166` is fine — it's the Go 1.26 idiomatic pattern explicitly allowed by `cli/azd/AGENTS.md`. No action needed.
- The `--output text` (vs `table`) change earlier in the commit history is a reasonable fit for a flat list.
- `init list` is correctly registered via `cmd.AddCommand(newInitListCommand(extCtx))` in `newInitCommand`.

Nothing here is merge-blocking; the shell-safety one (M1) and the JSON discriminator doc (M2) are the two I'd most like to see addressed before this becomes a documented contract that downstream agents depend on.

cli/azd/extensions/azure.ai.agents/internal/cmd/sample_list.go
```diff
@@ -0,0 +1,323 @@
...
+
+// Known template type filter values.
+var knownInitListTypes = []string{TemplateTypeAgent, TemplateTypeAzd}
+
+func newInitListCommand(extCtx *azdext.ExtensionContext) *cobra.Command {
+	flags := &initListFlags{}
+	extCtx = ensureExtensionContext(extCtx)
+
+	cmd := &cobra.Command{
+		Use:     "list",
```
trangevi, line 139: Should we make this "template-list"? `init list` feels a little weird to me.

cli/azd/extensions/azure.ai.agents/internal/cmd/init.go
```diff
@@ -1582,6 +1588,19 @@ func (a *InitAction) downloadAgentYaml(
...
+				// In no-prompt mode, refuse to silently overwrite a manifest
+				// that lives inside the project's src tree. Headless callers
+				// can move the manifest, choose a different --src target, or
+				// run interactively to confirm.
+				if a.flags.noPrompt {
+					return nil, "", exterrors.Validation(
+						exterrors.CodeInvalidManifestPointer,
+						fmt.Sprintf("manifest %q is inside the project src directory %q and would be overwritten", manifestPointer, srcDir),
+						"move the manifest outside the src tree, pass a different --src directory, "+
+							"or run interactively to confirm the overwrite",
```
trangevi: Should we provide a --overwrite flag for situations like this? This won't be the only scenario where today we have an interactive overwrite prompt which can't be handled in no-prompt mode

cli/azd/extensions/azure.ai.agents/internal/cmd/sample_list.go
```diff
@@ -0,0 +1,323 @@
...
+		Aliases: []string{"ls"},
+		Short:   "List available agent templates that can be used with `azd ai agent init -m`.",
+		Long: `List available agent templates from the curated catalog.
+
+Each entry includes the manifest URL or repo URL that can be passed back into
+` + "`azd ai agent init -m <url>`" + ` (for agent manifests) or ` + "`azd init -t <url>`" + `
+(for full azd template repositories), and a ready-to-execute ` + "`initCommand`" + `
```
trangevi, line 146: Does this mean that this command will list _all_ azd templates, not just agent templates? Shouldn't we just filter to agents here?

cli/azd/extensions/azure.ai.agents/internal/cmd/init_list.go
```diff
@@ -0,0 +1,323 @@
...
+				return err
+			}
+
+			ctx := cmd.Context()
+
+			httpClient := &http.Client{
+				Timeout: 30 * time.Second,
+			}
+
+			templates, err := fetchAgentTemplates(ctx, httpClient)
```
trangevi: NIT: I've been pushing all commands to have some kind of action object that gets called from this RunE wrapper. Basically, validate inputs here, then construct the action object and call <action>.Run or something similar for the actual functionality.

cli/azd/extensions/azure.ai.agents/internal/cmd/sample_list.go
trangevi: The command below looks like it's filtering to agents, so perhaps this is bad advice?

trangevi (approved):
Nothing that I would consider blocking

cli/azd/extensions/azure.ai.agents/internal/cmd/sample_list.go
therealjohn: Renamed to `azd ai agent sample list` under a new top-level `sample` parent (room for `sample show` / `sample search` later). No back-compat alias since the feature is unreleased. JSON wire envelope stays `templates` because the catalog source itself is templates; the `sample` namespace is just the CLI label. See 5889780.

cli/azd/extensions/azure.ai.agents/internal/cmd/init.go
therealjohn: Added `--force` (using azd's canonical overwrite flag name, matching `infra generate --force` / `env remove --force`). Narrowly scoped to just this manifest-in-src refusal in `downloadAgentYaml` -- without `--no-prompt`, behavior is unchanged. With `--no-prompt` and no `--force`, the structured Validation error now names `--force` explicitly so the next-step is discoverable. See 5889780. Other interactive-overwrite paths can adopt the same flag as they are surfaced; punted to follow-up so this PR stays scoped.

cli/azd/extensions/azure.ai.agents/internal/cmd/init_list.go
therealjohn: Done -- extracted `SampleListAction` following the `MonitorAction` / `ProjectShowAction` shape in this extension. `RunE` now only validates flags and dispatches; the catalog fetch + render lives in `SampleListAction.Run(ctx) error`. See 5889780.

cli/azd/extensions/azure.ai.agents/internal/cmd/sample_list.go
therealjohn: Acked your own follow-up -- `buildTemplateListItems` filters by `EffectiveType()` so the default output includes both agent-manifest and azd-template entries (with `--type agent` to scope to manifests only). No change in 5889780.

therealjohn:
Thanks @wbreza -- addressed all three findings plus the two bot items you endorsed. All in 5889780.

**M1 - shell-safe initCommand**
`mapAgentTemplateToDTO` now uses `fmt.Sprintf("azd ai agent init -m %q", t.Source)` (and `%q` on the `azd init -t` branch too). I went with quoting at construction rather than allow-listing or pushing the burden to consumers -- it keeps the `ready to paste` framing intact for the common cases (whitespace, embedded quotes), and the catalog is curated upstream so this isn't a hardening boundary against untrusted input. The doc comment on `TemplateListItem.InitCommand` now spells out the contract explicitly: `%q` does NOT neutralize shell expansion of `$` / backticks, so callers preferring pre-tokenized argv should reach for `ManifestURL` / `RepoURL` directly. Regression test: `TestMapAgentTemplateToDTO_InitCommandQuotesURLs`.

**M2 - ManifestURL / RepoURL discriminator**
Added a struct-level doc comment on `TemplateListItem` documenting the `Type`-based discriminator invariant. The existing `TestMapAgentTemplateToDTO_ManifestUrlAndRepoUrlAreMutuallyExclusive` already asserts it for both branches; no new test needed.

**Low - fetch error message**
`fetchAgentTemplates` error now includes the URL in the primary message: `failed to fetch agent templates catalog from <url>: <err>` so triage doesn't require reading the suggestion.

**Bot - centralize gosec helper**
Added `internal/cmd/index_helpers.go` with `boundedInt32Index(i int) int32` and replaced 5 in-package `int32(i) //nolint:gosec` sites with the single helper (one documented annotation instead of five). Left the one site in `pkg/agents/agent_yaml/parameters.go` alone since a new package for one helper felt over-engineered for that boundary.

**Bot - narrow subscription error wrap**
Added `exterrors.IsPromptRequired(err error) bool` (detection via the gRPC status message, since `input.PromptRequiredError` isn't translated by the host's `mapHostError` interceptor and arrives as a plain status). The subscription wrap in `init_foundry_resources_helpers.go` now only attaches the `AZURE_SUBSCRIPTION_ID` guidance when `IsPromptRequired` is true; other failures get a generic `retry, or run interactively` suggestion so we don't over-promise the remediation.

Two notes:
- `selectLocation` in the same file (~L825) has the same pattern but I left it out of scope for this PR. Happy to do it in a follow-up.
- Caught during finishing: the `no-prompt` template-selection error in `init_from_templates_helpers.go` had a stale `azd ai agent init list` reference -- updated to `sample list`.

Preflight clean: gofmt, golangci-lint (0 issues), cspell (0 issues), full `go test ./...` green.

cli/azd/extensions/azure.ai.agents/internal/cmd/init.go
therealjohn: Chose --force to be consistent with other commands.

trangevi (approved):

PR #8273, opened 2026-05-20
Fix `azd x watch` output loop

### LINKED ISSUE #8274: azd x watch rebuilds continuously when output directory changes
On Windows, `azd x watch` can enter a rebuild loop after the initial build. The watcher reports changes under the extension build output directory, for example `Z:\azure-dev\cli\azd\extensions\azure.ai.agents\bin`, then rebuilds and installs again, which writes to the same output directory and triggers another rebuild.\n\nThis appears to be a regression after `azd-ext-microsoft-azd-extensions_0.10.0`.

The watcher root was changed to an absolute path, but the hardcoded ignore patterns for folders such as `bin` remained relative, so absolute fsnotify event paths can bypass the ignore filter.\n\nExpected behavior: build output directories such as `bin` are ignored by `azd x watch` and do not trigger rebuilds.

Image

Thanks @trangevi for reporting

JeffreyCA:
Fixes #8274

This PR fixes a rebuild loop in `azd x watch` by normalizing file watcher event paths relative to the extension project root before applying hardcoded and user-defined ignore rules.

The bug was introduced in #7697 when the watcher root changed to an absolute path, causing absolute Windows events like `...\bin` to bypass relative ignore patterns such as `bin/**/*`, so rebuild output triggered another rebuild.

This PR also adds regression tests covering absolute output-directory events and `.azdxignore` matching.

### DIFF
--- a/cli/azd/extensions/microsoft.azd.extensions/internal/cmd/watch.go
+++ b/cli/azd/extensions/microsoft.azd.extensions/internal/cmd/watch.go
@@ -142,36 +142,8 @@ func runWatchAction(ctx context.Context, flags *watchFlags) error {
 				return nil
 			}
 
-			// Fast path: ignore events matching hardcoded glob patterns.
-			shouldIgnore := false
-			for _, pattern := range globIgnorePaths {
-				matched, _ := doublestar.PathMatch(pattern, event.Name)
-				if matched {
-					shouldIgnore = true
-					break
-				}
-			}
-			if shouldIgnore {
+			if shouldIgnoreWatchEvent(cwd, event.Name, globIgnorePaths, ignoreMatcher) {
 				continue
-			}
-
-			// Check user-defined ignore patterns (.azdxignore / .gitignore).
-			// Use os.Stat once to determine if the path is a directory.
-			info, statErr := os.Stat(event.Name)
-			isDir := statErr == nil && info.IsDir()
-
-			if relPath, relErr := filepath.Rel(cwd, event.Name); relErr != nil {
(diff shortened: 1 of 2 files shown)


wbreza (approved):
Approving — the fix is correct and minimal. Absolute Windows event paths were bypassing the relative `bin/**/*` style globs in `doublestar.PathMatch`; normalizing to a forward-slash relative path before matching is the right call, and the refactor into `shouldIgnoreWatchEvent` reads much better than the inlined version.

A few non-blocking nits / follow-up ideas:

1. **`relativeWatchPath` fallback on `filepath.Rel` error** — when `filepath.Rel` fails (cross-volume paths), the helper returns the original `eventName`, which on Windows still has backslashes and is not run through `ToSlash`. That value is then fed into both `doublestar.PathMatch` and `ignoreMatcher.IsIgnored`. The prior code skipped `IsIgnored` entirely on that branch, so behavior differs slightly in the error path. Extremely rare in practice, just flagging.

2. **Windows-separator test coverage** — the regression that motivated the PR was backslash paths failing forward-slash globs. The new tests use `filepath.Join`, which produces forward slashes on Linux/macOS CI, so the `ToSlash` normalization isn''t exercised on the OS where the bug actually manifested. Consider a small unit test against `relativeWatchPath` with an explicit backslash input, or a `runtime.GOOS == "windows"` guarded case, to lock the fix in.

3. The condensed `return statErr != nil && ignoreMatcher.IsIgnored(relPath, true)` is functionally identical to the prior nested form — no concern, just calling it out for the next reader.

Nice cleanup overall.

PR #8347, opened 2026-05-23
[azure.ai.agents] Use v1 for endpoint requests

### LINKED ISSUE #8329: Wrong API version for requests
I believe we should be using `v1`

glharper:
## Summary
- use `api-version=v1` for Azure AI Agents hosted endpoint protocol and session requests
- keep agent management/deploy/eval/file/doctor APIs on `2025-11-15-preview`
- update generated endpoint URLs, help text, and tests

Fixes #8329

## Testing
- `go test ./...`
- `golangci-lint run ./...`

### DIFF
--- a/cli/azd/extensions/azure.ai.agents/internal/cmd/agent_context.go
+++ b/cli/azd/extensions/azure.ai.agents/internal/cmd/agent_context.go
@@ -19,7 +19,7 @@ import (
 )
 
 // DefaultAgentAPIVersion is the default API version for agent operations.
-const DefaultAgentAPIVersion = "2025-11-15-preview"
+const DefaultAgentAPIVersion = "v1"
 
 // AgentEndpointAPIVersion is the API version for hosted agent endpoint protocol and session requests.
 const AgentEndpointAPIVersion = agent_api.AgentEndpointAPIVersion
--- a/cli/azd/extensions/azure.ai.agents/internal/cmd/agent_context.go
+++ b/cli/azd/extensions/azure.ai.agents/internal/cmd/agent_context.go
@@ -19,13 +19,7 @@ import (
 )
 
 // DefaultAgentAPIVersion is the default API version for agent operations.
-const DefaultAgentAPIVersion = "v1"
-
-// AgentEndpointAPIVersion is the API version for hosted agent endpoint protocol and session requests.
-const AgentEndpointAPIVersion = agent_api.AgentEndpointAPIVersion
-
-// ConversationsAPIVersion is the API version used by the Foundry Conversations protocol.
-const ConversationsAPIVersion = "v1"
+const DefaultAgentAPIVersion = agent_api.AgentEndpointAPIVersion
 
 // AgentContext holds the common properties of a hosted agent.
 type AgentContext struct {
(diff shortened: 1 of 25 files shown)


cli/azd/extensions/azure.ai.agents/internal/cmd/agent_context.go
```diff
@@ -19,7 +19,10 @@ import (
 )
 
 // DefaultAgentAPIVersion is the default API version for agent operations.
-const DefaultAgentAPIVersion = "2025-11-15-preview"
+const DefaultAgentAPIVersion = "v1"
+
+// AgentEndpointAPIVersion is the API version for hosted agent endpoint protocol and session requests.
+const AgentEndpointAPIVersion = agent_api.AgentEndpointAPIVersion
```
trangevi: This feels like it shouldn't be needed? At the very least, this and the "default" one should only be one thing. Having it reference the agent_api version seems fine, to make sure those stay in sync

cli/azd/extensions/azure.ai.agents/internal/cmd/agent_context.go
```diff
@@ -19,7 +19,10 @@ import (
...
 
 // ConversationsAPIVersion is the API version used by the Foundry Conversations protocol.
 const ConversationsAPIVersion = "v1"
```
trangevi: Can we also merge this one as well? So we end up with a single const?

cli/azd/extensions/azure.ai.agents/internal/project/service_target_agent.go
```diff
@@ -47,8 +47,8 @@ import (
 
 // Reference implementation
 
-// agentAPIVersion is the API version used for agent endpoint invocation URLs.
-const agentAPIVersion = "2025-11-15-preview"
+// agentAPIVersion is the API version used for agent management operations.
+const agentAPIVersion = "v1"
```
trangevi: If we change the stuff in the agent_context to reference the new api_versions.go const, can we update this to do the same?

trangevi (changes requested):

cli/azd/extensions/azure.ai.agents/internal/pkg/agents/eval_api/operations.go
```diff
@@ -145,63 +145,63 @@ func (c *EvalClient) GetEvaluatorRaw(
 func (c *EvalClient) CreateOpenAIEval(
 	ctx context.Context,
 	request *CreateOpenAIEvalRequest,
-	apiVersion string,
+	_ string,
```
trangevi: If this parameter isn't used, we should remove it? Same for the other methods in here

trangevi (changes requested):

glharper:
Addressed trangevi's latest review thread in commit 8dd49df88.

The OpenAI eval endpoints now carry `v1` in the path (`/openai/v1/evals`), so the `apiVersion` argument on those OpenAI eval helper methods was no longer used. I removed that unused parameter from:

- `CreateOpenAIEval`
- `ListOpenAIEvals`
- `GetOpenAIEval`
- `CreateOpenAIEvalRun`
- `ListOpenAIEvalRuns`
- `GetOpenAIEvalRun`

I also updated the command call sites and tests to use the cleaner signatures, while leaving the non-OpenAI eval APIs unchanged because they still send an `api-version` query parameter.

Validation:
- Focused eval tests: 50 passed
- `go test ./internal/pkg/agents/eval_api ./internal/cmd`
- `git diff --check`

trangevi (approved):

PR #8363, opened 2026-05-26
Defer Azure context for no-prompt agent init

### LINKED ISSUE #8331: Can't init with --no-prompt because of missing AZURE_SUBSCRIPTION_ID
AZURE_SUBSCRIPTION_ID and AZURE_LOCATION are required for `init` to complete successfully. The problem is that these have to be set in the AZD environment, which is only created during `init`. If you create an environment first, `init` doesnt use it, it will just still try to create another one. 

This affects coding agents ability to nicely create a project. 

To fix, init (all flows including from code, etc.) should not error on missing these environment variables, it should just warn and provide a suggestion on how to set them (which the error already does). This way a coding agent can run `init --no-prompt`, get a valid azure.yaml and environment, and then manually set the values.

glharper:
## Summary
- Allows `azd ai agent init --no-prompt` to create local artifacts when `AZURE_SUBSCRIPTION_ID` and/or `AZURE_LOCATION` are missing.
- Warns and defers Azure/model setup instead of prompting or failing in headless init flows.
- Updates post-init Next guidance to show required `azd env set ...` commands before `azd provision`.

Fixes #8331.

## Validation
- `go test ./internal/cmd/...`
- `git diff --check`
- `go test ./...` was run and only fails on unrelated Windows symlink privilege tests in `internal/pkg/paths` and `internal/project` with `A required privilege is not held by the client`.


therealjohn (approved):

cli/azd/extensions/azure.ai.agents/internal/cmd/init_foundry_resources_helpers.go
```diff
@@ -716,6 +716,65 @@ func loadAzureContext(
...
+		fmt.Println(output.WithGrayFormat(
+			"  # Optional: azd env set AZURE_AI_DEPLOYMENTS_LOCATION <region>",
+		))
+	}
+	if hasModelResources {
+		fmt.Printf("%s", output.WithWarningFormat(
+			"Model resource configuration was deferred because model lookup requires the missing Azure values.\n",
+		))
+		fmt.Println(output.WithGrayFormat(
+			"Set the missing values, then re-run init or configure model deployments in azure.yaml before provisioning.",
```
trangevi: Our "init from existing setup" story isn't very good, so I'm not sure giving advice to set the missing azure values is great. If we want to leave that, we should at least specify what values to set.

The model deployment property also isn't the simplest thing, so providing a sample of what the deployment information needs to be and where would probably be necessary.

@therealjohn suggestions here? This probably could get kinda messy

cli/azd/extensions/azure.ai.agents/internal/cmd/init.go
```diff
@@ -1236,6 +1249,26 @@ func (a *InitAction) configureModelChoice(
 			); err != nil {
 				log.Printf("warning: failed to update project provision signal: %v", err)
 			}
+		} else if a.flags.noPrompt {
```
trangevi, line 1383: Why is this block being added? I believe `ensureSubscriptionAndLocation` prompts if the values aren't set, so this will fail, is that the intent?

cli/azd/extensions/azure.ai.agents/internal/cmd/init.go
```diff
@@ -1371,6 +1404,27 @@ func (a *InitAction) configureModelChoice(
 		} else {
 			return nil, fmt.Errorf("specified foundry project was not found or is not eligible for the current configuration: %s", a.flags.projectResourceId)
 		}
+	} else if a.flags.noPrompt {
```
trangevi, line 1527: Same question about intent

cli/azd/extensions/azure.ai.agents/internal/cmd/init_from_code.go
```diff
@@ -480,6 +480,12 @@ func (a *InitFromCodeAction) createDefinitionFromLocalAgent(ctx context.Context)
 	}
 
 	// Create the azd environment now that we have the agent name
+	if a.environment == nil {
```
trangevi, line 580: This check is the same as the one below it, can they be combined?

cli/azd/extensions/azure.ai.agents/internal/cmd/init_from_code.go
```diff
@@ -557,6 +574,29 @@ func (a *InitFromCodeAction) createDefinitionFromLocalAgent(ctx context.Context)
...
 		}
+	} else if shouldDeferInitAzureContext(a.flags.noPrompt, a.azureContext) {
+		if err := configureDeferredInitAzureContext(
+			ctx, a.azdClient, a.environment.Name, a.azureContext, false,
+		); err != nil {
+			return nil, err
+		}
+		deferredAzureContext = true
+	} else if a.flags.noPrompt {
+		newCred, err := ensureSubscriptionAndLocation(
```
trangevi: Same question about what this new branch is for?

trangevi (changes requested):

cli/azd/extensions/azure.ai.agents/internal/cmd/init_foundry_resources_helpers.go
therealjohn: Yes, good point. We should show what values are missing, like `azd env set <var> <value>`

glharper:
Addressed the existing review threads in commit 492a69ce2:

- For the deferred Azure context guidance, the output now lists the specific missing `azd env set ...` commands and includes a concrete `deployments:` sample showing the expected `azure.yaml` shape for manual model deployment configuration.
- For the questions about the new no-prompt branches, the shared `configureNewProjectForNoPrompt` helper makes the intent explicit: when Azure context is already available, `--no-prompt` configures a new project path without prompting; when context is missing, the earlier deferral branch handles it.
- For the duplicate local-code init check, the environment lookup/create logic is now combined into a single `a.environment == nil` block.

Validation:
- Focused init tests: 307 passed
- `go test ./internal/cmd/...`
- `git diff --check`

glharper:

trangevi (approved):

PR #8376, opened 2026-05-26
feat(azure.ai.agents): agent-driven init pre-flow + --from-code routing

### LINKED ISSUE #6819: azd ai agent init: Document --no-prompt behavior and ensure CLI flag coverage

therealjohn:
## What this adds

Reshape `azd ai agent init` so the opening prompts hand the developer's coding agent everything it needs to drive the rest of setup, then get out of the way. The new pre-flow asks five short questions:

1. Which coding agent are you using? (Copilot CLI, Claude Code, etc.)
2. Should I install the `azd-ai-skill` pack into your agent?
3. Copy the starter prompt to your clipboard?
4. Pick (or create) a Foundry project.
5. Pick (or create) a model deployment.

From there, the developer pastes the starter prompt into their coding agent and the agent drives `init` / `provision` / `deploy` from inside their editor, using the installed `SKILL.md` as its instruction set. The starter prompt already carries the chosen Foundry project ID and model deployment, so the agent has the right context before its first command.

The pre-flow also adds a `--from-code` flag for the "I already have agent source, just wire it up" case (brownfield repos with hand-written agent code) and detects bootstrap-only directories so it can short-circuit cleanly when the cwd has no manifest yet.

## Why it matters

`azd ai agent init` today assumes a human is sitting at the terminal answering prompts. Developers who want their coding agent to drive setup have to copy-paste sample commands, screen-share, or guess at the right flag combinations. There's no in-the-box affordance for handing setup off to an AI assistant.

This PR turns "let my coding agent set this up for me" into a first-class option. Two short questions in, the developer is back in their editor and the agent is running the show -- with the right skill pack installed, the right starter prompt loaded, and the right Foundry resources already chosen.

Together with the `azure.ai.docs` extension (separate PR), this is the developer-facing half of an agent-driveable `azure.ai.*` surface.

## Dependencies

* The pre-flow dispatches into `azure.ai.docs` for the skill install step via the extension framework -- no compile-time dependency. The branch builds and tests standalone.
* One new external Go dependency: `github.com/atotto/clipboard` (cross-platform clipboard with graceful fallback in CI / SSH / `TERM=dumb` environments).

## Scope

* Touches `azd ai agent init` and its supporting helpers; no other commands change here.
* Includes a new shared `internal/helpformat` package (used by `init`'s styled `--help`); the rest of the agents commands will be migrated to it in a follow-up PR.
* `--output json` / confirmation-envelope migration on write commands is explicitly out of scope and will land separately.

Helps fix #6819


cli/azd/extensions/azure.ai.agents/internal/cmd/ext_lookup.go
```diff
@@ -0,0 +1,160 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+// ext_lookup.go provides helpers for talking to the azd host's extension
+// layer from inside this extension. Two responsibilities:
+//
+//  1. Detect whether a sibling extension (e.g. azure.ai.docs) is
```
trangevi, line 7: Do we want to add the new extension as a dependency of this one? Having the lookup isn't a bad thing either way

cli/azd/extensions/azure.ai.agents/internal/cmd/init_preflow.go
```diff
@@ -0,0 +1,880 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
trangevi, line 1: would it be possible to move the files that have coding agent specific handling into a subfolder, without requiring significant refactoring? This cmd folder is overwhelming to find stuff in, and I'd like to try and clean it up a bit if at all possible, so checking to see if some stuff can be done pre-emptively

trangevi (approved):

cli/azd/extensions/azure.ai.agents/internal/cmd/ext_lookup.go
therealjohn: We can in the future with azd fixes semver for the dependencies

cli/azd/extensions/azure.ai.agents/internal/cmd/init_preflow.go
therealjohn: will save for future refactoring.

trrwilson (approved):

PR #8396, opened 2026-05-27
feat(toolboxes): support skills on toolbox versions

### LINKED ISSUE #8397: Support attaching Foundry Skills to toolbox versions in azd ai toolbox
**What**

azd ai toolbox should support attaching Foundry Skills as references on toolbox versions, per the new optional skills field on ToolboxVersionObject and CreateToolboxVersionRequest in the Foundry data-plane spec.

**Why**

Toolboxes are the unit agents reference at run time; skills are pre-authored capability snippets users want to reuse across toolboxes. Today users have to hand-edit the JSON payload to attach a skill, then drive createToolboxVersion directly. The CLI should make this a first-class operation.

**Scope**

- `toolbox create --from-file`: accept an optional skills[] block alongside connections[].
- New `azd ai toolbox skill` subgroup with add, remove, and list verbs, mirroring the existing connection subgroup. Each mutation publishes a new immutable toolbox version (no dedicated REST endpoint for per-skill mutation).
- `connection add` / `connection remove`: carry skills forward verbatim into the new version.
- `toolbox show` / `toolbox version list`: surface the new field.

**Out of scope**

- policies / rai_config field on ToolboxVersionObject — separate follow-up.
- Other ToolboxSkill discriminator variants beyond skill_reference — none defined in the spec today.

**References**
- Foundry toolbox spec: https://github.com/Azure/azure-rest-api-specs/blob/feature/foundry-release/specification/ai-foundry/data-plane/Foundry/src/toolboxes/models.tsp

hund030:
## Summary

Adds `skills` support on toolbox versions, per the new optional field on `ToolboxVersionObject` / `CreateToolboxVersionRequest` in the Foundry spec. Follow-up to #8322.

## Changes

- **`internal/pkg/azure`**: `Skills []map[string]any` on `CreateToolboxVersionRequest` (omitempty) and `ToolboxVersionObject` (always serialized, matching service behavior). `[]map[string]any` keeps future `ToolboxSkill` discriminator variants forward-compatible.
- **`internal/cmd/toolbox_skill.go`**: name validator (mirrors `azure.ai.skills`' regex, documented), `parseSkillFlag`, `buildSkillEntry`, `validateNoDuplicateSkills`.
- **`toolbox create --from-file`**: optional `skills[]` block in the input file.
- **`toolbox skill add | remove | list`**: new subgroup. Each mutation publishes a new immutable toolbox version via `createToolboxVersion`; no dedicated REST endpoint exists. `skill remove` allows removing the last skill (skills are optional content, unlike connections).
- **`connection add` / `connection remove`**: carry skills forward verbatim into the new version.
- **`toolbox show`** / **`toolbox version list`**: surface skill count and a per-skill table / `skills_count` column.

## Validation

- Unit tests cover the new helpers, the `create` skills-from-file path, carry-forward on connection add/remove, and all three skill verbs.
- Live e2e against a Skills-enabled Foundry project: create with skill → connection peel preserves skill → `skill remove` → empty list → `skill add` republishes. Full `Test_CLI_Toolbox_Smoke` flow green.


cli/azd/extensions/azure.ai.toolboxes/internal/cmd/toolbox_files.go
```diff
@@ -92,15 +94,20 @@ func parseToolboxFile(path string, out any) error {
...
 func suggestionForParseError(out any, err error) string {
 	msg := err.Error()
-	if _, ok := out.(*toolboxToolsFile); ok && strings.Contains(msg, "description") {
-		return "the 'description' field is only accepted by `toolbox create`; " +
-			"in v1 a toolbox's description is set at create time and cannot be changed later"
+	if _, ok := out.(*toolboxToolsFile); ok {
+		switch {
+		case strings.Contains(msg, "description"):
+			return "the 'description' field is only accepted by `toolbox create`; " +
+				"in v1 a toolbox's description is set at create time and cannot be changed later"
```
therealjohn: Remove all instances of a `v1` descriptor.

therealjohn (changes requested):

trangevi (changes requested):
Blocking this PR while I get some clarity on the API surface

viswabalaji (approved):

cli/azd/extensions/azure.ai.toolboxes/internal/cmd/toolbox_skill_group.go
```diff
@@ -0,0 +1,49 @@
...
+
+// newToolboxSkillCommand returns the `azd ai toolbox skill` parent.
+func newToolboxSkillCommand(extCtx *azdext.ExtensionContext) *cobra.Command {
+	extCtx = ensureExtensionContext(extCtx)
+	cmd := &cobra.Command{
+		Use:   "skill",
+		Short: "Manage skill references attached to a toolbox.",
+		Long: `Manage skill references attached to a toolbox.
+
+Each add/remove publishes a new immutable version and retargets the toolbox
```
trangevi: Just like we did for tools, we shouldn't require things to be added/removed one at a time, with a new toolbox version every time. From what I understand, the API supports bulk changes, so we should do the same.

cli/azd/extensions/azure.ai.toolboxes/internal/cmd/toolbox_skill_add.go
```diff
@@ -0,0 +1,146 @@
...
+	"github.com/azure/azure-dev/cli/azd/pkg/azdext"
+	"github.com/spf13/cobra"
+)
+
+// newToolboxSkillAddCommand returns the `skill add` command.
+func newToolboxSkillAddCommand(extCtx *azdext.ExtensionContext) *cobra.Command {
+	extCtx = ensureExtensionContext(extCtx)
+
+	cmd := &cobra.Command{
+		Use:   "add <toolbox> <skill>[@<version>]",
```
trangevi: called this out in another comment, but we should support taking multiple at once. We can still have a positional argument for when users only want one, but we should support another "skill" parameter which can be provided multiple times to provide a list of skills in a single command

cli/azd/extensions/azure.ai.toolboxes/internal/cmd/toolbox_skill_add.go
```diff
@@ -0,0 +1,146 @@
...
+		Short: "Attach a skill reference to a toolbox.",
+		Long: `Attach a skill reference to a toolbox.
+
+Publishes a new default version with the skill appended. When the version is
+omitted, the reference resolves to the skill's default version at read time.
```
trangevi: We shouldn't be automatically updating the default toolbox version. Since it's a shared resource, that could have implications if multiple people are using the same toolbox.

I think I missed this in the original commands, so please make sure that any modifications we're doing, connection add/remove, skill add/remove, etc., aren't updating the default version.

Instead, let's add an `azd ai toolbox publish` command, which takes the name of the toolbox and a version, and makes the "set default" call

trangevi (changes requested):

cli/azd/extensions/azure.ai.toolboxes/internal/cmd/toolbox_files.go
hund030: Removed.

cli/azd/extensions/azure.ai.toolboxes/internal/cmd/toolbox_skill_add.go
hund030: make it consistent with `connection add`. We can use positional argument for adding one skill and use `--from-file` for adding multiple.

cli/azd/extensions/azure.ai.toolboxes/internal/cmd/toolbox_skill_group.go
hund030: `connection remove`, `skill add` and `skill remove` all support batch input now. Remove are using variadic positional arguments and `skill add` support `--from-file` flag.

cli/azd/extensions/azure.ai.toolboxes/internal/cmd/toolbox_skill_add.go
hund030: We already have a `toolbox update` command that update default version of toolbox. And `connection add/remove` and `skill add/remove` no longer call `SetDefaultVersion` after publishing the new version. 

Now the flow becomes:
```
azd ai toolbox create <tb> --from-file <path> // v1 (default)
azd ai toolbox connection add <tb> --from-file <path> // v2
azd ai toolbox skill ad <tb> --from-file <path> // v3
azd ai toolbox update <tb> --default-version 3 // v3 (default)
```
cc @therealjohn

trangevi (approved):

therealjohn (approved):

therealjohn (approved):

PR #8402, opened 2026-05-27
fix: truncate progress widget lines to terminal width on narrow terminals

### LINKED ISSUE #8382: azd - progress widget renders oddly on small screens
**Describe the bug**
During the `azd ai agent deploy` flow, if your terminal is not very wide on powershell the progress widget prints over and over and doesn't update in place.

Image

jongio:
## Summary

Fixes #8382. On narrow terminals (PowerShell, small panes), the progress widget (TaskList/Spinner) rendered lines that wrapped beyond terminal width. `ClearCanvas` couldn't clear enough rows, causing stale output to accumulate and print over itself.

## Changes

- **`TruncateVisible()`** — new ANSI-aware string truncation helper in `pkg/ux/ux.go`
- **`Printer.Width()`** — exposes terminal width on the Printer interface
- **`TaskList.Render()`** / **`Spinner.Render()`** — truncate each line to `consoleWidth - 1` visible characters before writing
- **Unit tests** for `TruncateVisible` covering ANSI sequences, unicode, and edge cases
- **Lint fix** — modernizes `uxlib.Ptr(true)` → `new(true)` in `copilot_agent.go` (pre-existing govet warnings)

## How it works

The `-1` prevents Windows' eager-wrap behavior (cursor wraps when a line fills exactly the terminal width). By keeping lines 1 char shorter than the terminal, wrapping never occurs and `ClearCanvas` row math stays correct.

## Testing

- All preflight checks pass (9/9 ✓)
- Manual test with 40-column terminal confirms correct in-place updates

### DIFF
--- a/cli/azd/pkg/ux/ux.go
+++ b/cli/azd/pkg/ux/ux.go
@@ -14,6 +14,7 @@ import (
 	"github.com/azure/azure-dev/cli/azd/internal/terminal"
 	"github.com/azure/azure-dev/cli/azd/pkg/ux/internal"
 	"github.com/fatih/color"
+	"github.com/mattn/go-runewidth"
 	"github.com/nathan-fiscaletti/consolesize-go"
 )
 
@@ -77,13 +78,12 @@ func CountLineBreaks(content string, width int) int {
 }
 
 // visibleLength calculates the number of visible characters in a string
-// by removing ANSI codes and counting the actual characters.
-// Cannot use len() as it counts bytes not runes
+// by removing ANSI codes and counting the display width (CJK/emoji = 2 columns).
 func VisibleLength(s string) int {
 	// Remove ANSI codes such as color, formatting, etc.
 	cleaned := specialTextRegex.ReplaceAllString(s, "")
-	// Count actual visible characters
-	return utf8.RuneCountInString(cleaned)
+	// Count display width (handles CJK/emoji as 2 columns)
+	return runewidth.StringWidth(cleaned)
 }
 
 // TruncateVisible truncates a string to fit within maxWidth visible characters,
@@ -133,15 +133,22 @@ func TruncateVisible(s string, maxWidth int) string {
 		}
 
 		// Regular character (may be multi-byte UTF-8)
-		_, size := utf8.DecodeRuneInString(s[i:])
+		r, size := utf8.DecodeRuneInString(s[i:])
+		w := runewidth.RuneWidth(r)
+		if visible+w > targetVisible {
+			break
+		}
 		result.WriteString(s[i : i+size])
 		i += size
-		visible++
+		visible += w
 	}
 
 	result.WriteString(ellipsis)
-	// Append a reset sequence to ensure no color bleeds past the truncation point
-	result.WriteString("\x1b[0m")
+	// Only append ANSI reset if the original string contained ANSI sequences,
+	// to avoid injecting escape codes into plain text (e.g., NO_COLOR mode).
+	if strings.Contains(s, "\x1b[") {
+		result.WriteString("\x1b[0m")
+	}
 
 	return result.String()
 }
--- a/cli/azd/pkg/ux/ux.go
+++ b/cli/azd/pkg/ux/ux.go
@@ -86,6 +86,66 @@ func VisibleLength(s string) int {
 	return utf8.RuneCountInString(cleaned)
 }
 
+// TruncateVisible truncates a string to fit within maxWidth visible characters,
+// preserving ANSI escape sequences (which don't consume terminal columns).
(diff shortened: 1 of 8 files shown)


vhvb1989:
## CI Failure Summary — Build [#6358377](https://dev.azure.com/azure-sdk/590cfd2a-581c-4dcb-a12e-6568ce786175/_build/results?buildId=6358377)

All 3 platforms (Linux, Mac, Windows) fail at **Test Go Binary**. Cross-builds, linting, cspell, and copyright checks all pass.

### Failures caused by PR changes

**1. `pkg/ux` — `TestVisibleLength`** (`formatting_test.go:56`) — All platforms

| Subtest | Expected | Actual |
|---|---|---|
| `CJK_characters` | 3 | 6 |
| `mixed_ANSI_and_unicode` | 2 | 4 |
| `tab_and_printable` | 3 | 2 |

**2. `pkg/ux` — `Test_CountLineBreaks`** (`ux_test.go:112`) — All platforms

| Subtest | Expected | Actual |
|---|---|---|
| `Emoji_Characters` | 0 | 1 |
| `Emoji_Line_Wrap` | 1 | 2 |
| `Mixing_Emoji_and_Text` | 1 | 2 |

**3. `pkg/ux` — `Test_VisibleLength`** (`ux_test.go:151`) — All platforms

| Subtest | Expected | Actual |
|---|---|---|
| `Unicode_Characters` | 3 | 6 |
| `Mix_of_ANSI_and_Unicode` | 3 | 6 |

### Root cause

The `VisibleLength` implementation does not account for **East Asian Wide characters** (CJK = 2 columns each), **emoji width** (2 columns each), or **tab stops** (variable width). Tests expect display-column-aware counting but the implementation appears to use rune count or byte length.

### Pre-existing / Flaky (not caused by PR)

**`internal/appdetect` — `TestDetect/ExcludePatterns`** — Windows only. Maven artifact download failure on CI agent (`Could not transfer artifact org.eclipse.sisu:org.eclipse.sisu.inject`). Transient network issue.

wbreza (approved):
## ✅ Approve — focused, well-tested fix

Nicely scoped fix for the narrow-terminal wrapping issue. The atomic width migration is a good drive-by improvement, the `runewidth` adoption is the right call for CJK/emoji correctness, and the iterative response to prior review feedback (atomic width, runewidth, cspell, gosec G115) is appreciated. All 31 CI checks are green.

### Strengths
- `TruncateVisible` correctly preserves ANSI CSI sequences while counting display columns; edge cases for `maxWidth` 0/1/2/3 are handled.
- Atomic `consoleWidth` eliminates the prior lock-protected mutation and the resize handler is now lock-free.
- Solid unit test coverage including ANSI fits/truncates, multi-ANSI, CJK, and minimum-width cases.
- Defensive `max(printer.Width()-1, 1)` guards against zero/negative width.
- Conditional `\x1b[0m` reset (only when the input contained ANSI) is nice — keeps NO_COLOR / plain-text paths clean.

### Optional nits (non-blocking)

1. **`pkg/ux/ux.go` — `TruncateVisible` may append a redundant `\x1b[0m`** when the input contained ANSI but all ANSI sequences fall within the kept prefix (or were already balanced with a reset). Harmless on terminals, slightly noisy in captured output. Could check whether any ANSI escape was actually retained in `result` before appending — current behavior is safe though.

2. **Only CSI (`\x1b[`) escapes are parsed.** OSC, DCS, etc. would be counted as visible bytes. In practice azd only emits SGR via `fatih/color`, so this is fine — worth a one-line comment documenting the assumption so the next reader doesn't have to derive it.

3. **`pkg/ux/printer.go` — extra blank line** before the new `Width()` method declaration in the `Printer` interface (two blank lines where one is conventional). Cosmetic.

4. **Subtle tab-width semantic change.** With the switch to `runewidth.StringWidth`, `"a\tb"` is now reported as 2 display columns (was 3). This is more accurate (tab width depends on tab stops), but it is a behavior change — worth keeping in mind if any callers relied on the old `RuneCountInString` semantic. No affected callers in the diff.

### Re: prior reviewer comment
The Copilot suggestion about import ordering in `printer.go` looks like a false positive — `sync/atomic` after `sync` is the correct stdlib alphabetical placement. Safe to dismiss.

Also confirming the `uxlib.Ptr(true)` → `new(true)` changes in `copilot_agent.go` are correct and idiomatic — `new(val)` is the preferred Go 1.26 pattern per `cli/azd/AGENTS.md`.

cli/azd/pkg/ux/ux.go
JeffreyCA, line 53: I think this can be removed now?

JeffreyCA (approved):
Changes look good

vhvb1989 (approved):

PR #8459, opened 2026-05-28
Fix azd auth status reporting unauthenticated in Cloud Shell

### LINKED ISSUE #8455: azd ai agent init fails with 'AzureCLICredential: signal: killed' for every region in Cloud Shell when auth.useAzCliAuth is true
### Output from `azd version`

`azd version 1.24.1 (commit f2f23c94e0cb04f7043eb4bffeaa60c011b7bb50)`

### Describe the bug

When running `azd ai agent init` in **Azure Cloud Shell** with `auth.useAzCliAuth` set to `true`, model catalog discovery fails for **every region** with:

```
ERROR: listing models: fetching model catalogs: <region>: AzureCLICredential: signal: killed

If you manually set the subscription ID (e.g., via AZURE_SUBSCRIPTION_ID in .env or `azd env set`), you must also set AZURE_TENANT_ID to the tenant ID that contains this subscription. Alternatively, run `azd auth login --tenant-id <tenant-id>` to allow azd to discover subscriptions in that tenant.
```

…repeated for ~50 Azure regions. The trailing "set AZURE_TENANT_ID…" hint is misleading — it's the generic credential-failure boilerplate and not the actual cause.

### Root cause

`pkg/ai/model_service.go:fetchModelsForLocations` fans out one goroutine **per Azure region** (~50 concurrent) to list the AI model catalog. Each goroutine path is:

```
GetAiModels → createModelsClient → credentialProvider.CredentialForSubscription
  → manager.CredentialForCurrentUser            (auth.useAzCliAuth = true)
  → azidentity.NewAzureCLICredential(...)       // brand-new credential every call
  → cred.GetToken(...)                          // spawns `az account get-access-token`
```

[`pkg/auth/manager.go:259-268`](https://github.com/Azure/azure-dev/blob/main/cli/azd/pkg/auth/manager.go#L259-L268) constructs a **fresh `AzureCLICredential` on every call** instead of returning a shared instance. Each fresh credential has its own token cache, so ~50 concurrent first-time `GetToken()` calls all spawn `az` (Python, ~150 MB resident) subprocesses simultaneously.

Cloud Shell containers have tight memory limits (~4 GB). ~50 concurrent Python `az` processes blow past that and the Linux OOM killer reaps them → `SIGKILL` → `azidentity` surfaces it as `AzureCLICredential: signal: killed`.

The `--debug` log makes the fan-out obvious — ~40 `"delegating auth to az since auth.useAzCliAuth is set to true"` entries all stamped at the same second:

```
2026/05/28 21:31:48 manager.go:260: delegating auth to az since auth.useAzCliAuth is set to true
2026/05/28 21:31:48 manager.go:260: delegating auth to az since auth.useAzCliAuth is set to true
... (~40x in the same second)
```

This only reproduces when `auth.useAzCliAuth=true`:

- With native azd auth (MSAL), no subprocess is spawned per call.
- With `useAzCliAuth=true` **and a shared credential instance**, `azidentity`'s internal `syncer` would dedupe concurrent `GetToken` calls into a single subprocess.

### To Reproduce

1. Open Azure Cloud Shell.
2. `azd config set auth.useAzCliAuth true`
3. `azd ai agent init`
4. Walk through the prompts to a step that triggers model catalog listing (e.g. "Deploy a new 'gpt-4.1-mini' model").

### Expected behavior

Model catalog listing should succeed when `auth.useAzCliAuth` is true, including in memory-constrained environments like Cloud Shell. Concurrent token acquisitions should not each spawn an independent `az` subprocess.

### Suggested fix

In order of preference:

1. **Cache the `AzureCLICredential` per tenant** in `pkg/auth/manager.go` when `useAzCliAuth` is true. Sharing a single instance collapses N concurrent first-time `GetToken` calls into one `az` subprocess via the SDK's internal sync. This also benefits every other parallel az-CLI-auth code path, not just model-catalog listing.
2. Bound the fan-out in `fetchModelsForLocations` with a small semaphore (e.g. 8–10 workers) to cap subprocess concurrency regardless of credential type.
3. Pre-warm the token with a single `GetToken` call before launching the fan-out so all parallel callers hit the cache.

### Environment

- **azd**: 1.24.1
- **Host**: Azure Cloud Shell (`AZD_IN_CLOUDSHELL=true`)
- **Config**: `auth.useAzCliAuth=true`

JeffreyCA:
Contributes to #8458

This PR fixes azd incorrectly reporting an unauthenticated state in Azure Cloud Shell, even though Cloud Shell provides an ambient managed credential that azd already uses for token acquisition.

**Without changes:**
image

**With changes:**

image

## Root cause

`auth.Manager.LogInDetails()` was the only user-state auth method missing a Cloud Shell fallback. Its siblings `CredentialForCurrentUser` and `GetLoggedInServicePrincipalTenantID` already fall back to the ambient Cloud Shell credential when no azd-managed user is logged in. `CredentialForCurrentUser` gates that fallback on `errors.Is(err, ErrNoCurrentUser)`, whereas `GetLoggedInServicePrincipalTenantID` currently falls back on any error from `readUserProperties`. `LogInDetails` had no fallback at all and returned the error, because in Cloud Shell the user never runs `azd auth login` and therefore has no stored user properties.

## Impact

That single gap surfaced as two distinct, confusing failures in Cloud Shell:

- `azd provision` failed during parameter loading, because the Bicep provider calls `CurrentPrincipalType`, which calls `LogInDetails` and propagated the error as "fetching current principal type".
- `azd ai agent init` blocked with a "not logged in" message, because it gates on parsing `azd auth status --output json`, which reported `unauthenticated` for the same reason.

Both flows actually work in Cloud Shell once the underlying login-state check is corrected, so no extension-side change is required.

## Fix

The fix adds a Cloud Shell fallback to `LogInDetails` that mirrors the existing pattern in the other auth methods. When no user is logged in via azd's own flow but the process is running in Cloud Shell, it derives the identity from the ambient credential and reports a User (email) login, populating the account from the token claims on a best-effort basis. An empty account is intentionally not treated as an error, since a Cloud Shell session is always a valid authenticated user and the principal type is what downstream callers such as `CurrentPrincipalType` actually depend on.

Consistent with `CredentialForCurrentUser`, the fallback is gated on `errors.Is(err, ErrNoCurrentUser)` so that other failures (e.g. corrupted stored user properties) surface instead of being silently masked. Tightening `GetLoggedInServicePrincipalTenantID` to do the same is left as a follow-up.

As a result, `azd auth status` now reports authenticated in Cloud Shell, `azd provision` resolves the principal type correctly, and `azd ai agent init` proceeds.

## Testing

Test coverage was added to `TestLogInDetails` for both the case where the Cloud Shell token exposes a username claim and the case where it does not, asserting that the session is reported as an authenticated User login in both, plus a case verifying that a non-`ErrNoCurrentUser` failure is surfaced rather than masked. Existing auth, provisioning, and command tests continue to pass, including under `-race`.

### DIFF
--- a/cli/azd/pkg/auth/manager.go
+++ b/cli/azd/pkg/auth/manager.go
@@ -1462,6 +1462,11 @@ func (m *Manager) LogInDetails(ctx context.Context) (*LogInDetails, error) {
 
 	currentUser, err := readUserProperties(cfg)
 	if err != nil {
+		// In Cloud Shell azd uses the ambient credential, so report that user
+		// rather than treating the session as unauthenticated.
+		if runcontext.IsRunningInCloudShell() {
+			return m.cloudShellLogInDetails(ctx)
+		}
 		return nil, ErrNoCurrentUser
 	}
 
@@ -1486,6 +1491,21 @@ func (m *Manager) LogInDetails(ctx context.Context) (*LogInDetails, error) {
 	}
 
 	return nil, ErrNoCurrentUser
+}
+
+// cloudShellLogInDetails reports the Cloud Shell user, derived from the ambient
+// credential. The session is always a valid user, so an empty account (no
+// username claim) is not an error.
+func (m *Manager) cloudShellLogInDetails(ctx context.Context) (*LogInDetails, error) {
+	claims, err := m.ClaimsForCurrentUser(ctx, nil)
+	if err != nil {
+		return nil, fmt.Errorf("fetching claims for Cloud Shell user: %w", err)
+	}
+
+	return &LogInDetails{
+		LoginType: EmailLoginType,
+		Account:   strings.TrimSpace(claims.DisplayUsername()),
+	}, nil
 }
 
 type AuthSource string
--- a/cli/azd/pkg/auth/manager.go
+++ b/cli/azd/pkg/auth/manager.go
@@ -1463,11 +1463,16 @@ func (m *Manager) LogInDetails(ctx context.Context) (*LogInDetails, error) {
 	currentUser, err := readUserProperties(cfg)
 	if err != nil {
 		// In Cloud Shell azd uses the ambient credential, so report that user
-		// rather than treating the session as unauthenticated.
-		if runcontext.IsRunningInCloudShell() {
-			return m.cloudShellLogInDetails(ctx)
+		// rather than treating the session as unauthenticated. Only fall back
+		// when there is genuinely no logged-in user; other errors (e.g. corrupted
+		// stored user properties) should surface so they aren't silently hidden.
+		if errors.Is(err, ErrNoCurrentUser) {
+			if runcontext.IsRunningInCloudShell() {
+				return m.cloudShellLogInDetails(ctx)
+			}
+			return nil, ErrNoCurrentUser
 		}
-		return nil, ErrNoCurrentUser
+		return nil, fmt.Errorf("reading current user properties: %w", err)
 	}
 
 	if currentUser.HomeAccountID != nil {
--- a/cli/azd/pkg/auth/manager.go
+++ b/cli/azd/pkg/auth/manager.go
@@ -1402,6 +1402,9 @@ type LogInDetails struct {
 // return the account name from the az CLI. When external authentication is
 // configured, it will acquire a token from the external auth endpoint (an
 // outbound HTTP call) and derive the account identifier from the token claims.
+// When running in Azure Cloud Shell and no azd-managed user is logged in,
+// it derives the account from the ambient Cloud Shell credential and reports
+// an authenticated user.
 func (m *Manager) LogInDetails(ctx context.Context) (*LogInDetails, error) {
 	if m.UseExternalAuth() {
 		claims, err := m.ClaimsForCurrentUser(ctx, nil)
--- a/cli/azd/pkg/auth/manager_test.go
+++ b/cli/azd/pkg/auth/manager_test.go
@@ -499,6 +499,7 @@ func TestLogInDetails(t *testing.T) {
 		_, err := m.LogInDetails(t.Context())
 		require.Error(t, err)
 		require.NotErrorIs(t, err, ErrNoCurrentUser)
+		require.ErrorContains(t, err, "reading current user properties")
 	})
 
 	t.Run("external auth - error when token has no usable account identifier", func(t *testing.T) {
--- a/cli/azd/pkg/auth/manager_test.go
+++ b/cli/azd/pkg/auth/manager_test.go
@@ -413,6 +413,71 @@ func TestLogInDetails(t *testing.T) {
 		require.Equal(t, "user@contoso.com", details.Account)
 	})
 
+	t.Run("cloud shell - returns user login type from token claims", func(t *testing.T) {
+		t.Setenv(runcontext.AzdInCloudShellEnvVar, "1")
+
+		// Build an access token with a username claim and mock the Cloud Shell
+		// token endpoint to return it.
+		token := buildTestJWT(t, map[string]any{
+			"unique_name": "user@contoso.com",
+			"oid":         "oid-abc",
+			"tid":         "tenant-xyz",
+		})
+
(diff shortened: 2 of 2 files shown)


cli/azd/pkg/auth/manager.go
```diff
@@ -1462,7 +1462,17 @@ func (m *Manager) LogInDetails(ctx context.Context) (*LogInDetails, error) {
 
 	currentUser, err := readUserProperties(cfg)
 	if err != nil {
-		return nil, ErrNoCurrentUser
+		// In Cloud Shell azd uses the ambient credential, so report that user
```
hemarina, line 1468: **`LogInDetails` Godoc (lines 1399-1404) is now stale.** It enumerates the legacy-auth and external-auth paths but doesn't mention this new Cloud Shell fallback. Mind adding a sentence so the behavior contract stays in sync? Something like:

```go
// When running in Azure Cloud Shell and no azd-managed user is logged in,
// it derives the account from the ambient Cloud Shell credential and reports
// an authenticated user.
```

cli/azd/pkg/auth/manager.go
```diff
@@ -1462,7 +1462,17 @@ func (m *Manager) LogInDetails(ctx context.Context) (*LogInDetails, error) {
...
+		// rather than treating the session as unauthenticated. Only fall back
+		// when there is genuinely no logged-in user; other errors (e.g. corrupted
+		// stored user properties) should surface so they aren't silently hidden.
+		if errors.Is(err, ErrNoCurrentUser) {
```
hemarina, line 1472: **PR description is slightly inaccurate re: sibling parity (and there's a latent inconsistency worth noting).** The description says:

> "Its siblings `CredentialForCurrentUser` and `GetLoggedInServicePrincipalTenantID` both fall back to the ambient Cloud Shell credential when `readUserProperties` returns `ErrNoCurrentUser`."

That's true of `CredentialForCurrentUser` (line 276 — `if errors.Is(err, ErrNoCurrentUser)`), but `GetLoggedInServicePrincipalTenantID` (line 466) falls back on **any** error from `readUserProperties`:

```go
currentUser, err := readUserProperties(authCfg)
if err != nil {            // <-- catches corruption too
    if runcontext.IsRunningInCloudShell() { ... }
    return nil, ErrNoCurrentUser
}
```

So after this PR, corrupted `currentUser` config in Cloud Shell will surface a real error from `LogInDetails` (good — what the new test 3 locks in), but be silently masked by `GetLoggedInServicePrincipalTenantID`.

No change required in this PR — the new behavior is the right one. Just worth either (a) updating the PR description so it doesn't claim parity that doesn't exist, or (b) filing a follow-up to tighten `GetLoggedInServicePrincipalTenantID` the same way.

cli/azd/pkg/auth/manager_test.go
```diff
@@ -413,6 +413,94 @@ func TestLogInDetails(t *testing.T) {
...
+
+		m := Manager{
+			configManager:     newMemoryConfigManager(),
+			userConfigManager: userCfgMgr,
+			cloud:             cloud.AzurePublic(),
+		}
+
+		_, err := m.LogInDetails(t.Context())
+		require.Error(t, err)
+		require.NotErrorIs(t, err, ErrNoCurrentUser)
```
hemarina, line 501: **Nit (optional):** This test asserts the error doesn't unwrap to `ErrNoCurrentUser`, which is the important part. Since the PR also establishes `"reading current user properties: %w"` as the wrap contract, consider locking that in so a future refactor that drops the wrap is caught:

```go
require.ErrorContains(t, err, "reading current user properties")
```

hemarina (approved):
Thanks for the fix — this is a real Cloud Shell papercut and the scoping is right. Commit 2 already addressed the obvious concern about masking non-`ErrNoCurrentUser` errors, and the three new subtests cover the happy paths plus the corruption negative cleanly. Approving — leaving a couple of small items worth resolving before merge plus one optional follow-up.

vhvb1989 (approved):

vhvb1989 (approved):

PR #8507, opened 2026-05-31
fix: make env refresh tolerant of unsupported service hosts

### LINKED ISSUE #7195: [Issue] azd env refresh doesn't work with azure.ai.agent extension

**Output from `azd version`**
```bash
$ azd version
azd version 1.23.10 (commit 0d0c25d9ae6125fec7ce87e4d9905a7a1949825b)
$ azd ai agent version
Version: 0.1.15-preview
Commit: bc47b5b6fc4883e2cdb761d9de092aa998a3c4e2
Build Date: 2026-03-13T21:08:45.4788848+00:00
```

**Describe the bug**
When I run `azd env refresh` in a project that uses the AI agent extension, it throws an error.

**To Reproduce**
Steps to reproduce the behavior...

```bash
$ azd ai agent init
[ interactive setup steps ]
$ azd env refresh

Refreshing environment dev (azd env refresh)


ERROR: service host 'azure.ai.agent' for service 'agent' is unsupported

Suggestion: Suggestion: install an extension that provides this host or update azure.yaml to use one of the supported hosts: appservice, containerapp, function, staticwebapp, aks, ai.endpoint

Your project is using host 'azure.ai.agent' which is not supported by default.

An extension was found that provides support for this host.
```

**Expected behavior**

Hopefully it does whatever `azd env refresh` is actually supposed to do.

jongio:
## Summary

Fixes #7195

`azd env refresh` fails with "service host 'azure.ai.agent' is unsupported" when an extension-provided service host (like `azure.ai.agent`) is declared in `azure.yaml` but the extension isn't loaded.

## Root Cause

`envRefreshAction.Run()` unconditionally calls `projectManager.Initialize()` which iterates all services and resolves their service targets. For extension-provided hosts not registered in the IoC container, `GetServiceTarget()` returns `UnsupportedServiceHostError`.

## Fix

Since `env refresh` only retrieves infrastructure outputs from Azure (it doesn't build or deploy services), it doesn't actually need service targets. The fix catches `UnsupportedServiceHostError` from both `Initialize` and `EnsureAllTools` and continues gracefully. Services with unsupported hosts simply won't receive the `ServiceEventEnvUpdated` event.

## Changes

- **cli/azd/cmd/env.go**: Modified `envRefreshAction.Run()` to tolerate `UnsupportedServiceHostError`; added `hasUnsupportedHostError()` helper using `errors.AsType`
- **cli/azd/cmd/env_test.go**: Added table-driven unit test with 5 cases (nil, unrelated error, direct, fmt.Errorf-wrapped, ErrorWithSuggestion-wrapped)

## Testing

- `go test ./cmd/... -short` - all tests pass
- `golangci-lint run ./cmd/...` - 0 issues

### DIFF
--- a/cli/azd/cmd/env.go
+++ b/cli/azd/cmd/env.go
@@ -1151,12 +1151,19 @@ func (ef *envRefreshAction) Run(ctx context.Context) (*actions.ActionResult, err
 		Title: fmt.Sprintf("Refreshing environment %s (azd env refresh)", ef.env.Name()),
 	})
 
+	// Initialize services, skipping any with unsupported hosts (e.g., extension-provided hosts
+	// that are not currently loaded). Env refresh only needs infrastructure outputs from Azure;
+	// services with unsupported hosts simply won't receive the environment-updated event.
 	if err := ef.projectManager.Initialize(ctx, ef.projectConfig); err != nil {
-		return nil, err
+		if !hasUnsupportedHostError(err) {
+			return nil, err
+		}
 	}
 
 	if err := ef.projectManager.EnsureAllTools(ctx, ef.projectConfig, nil); err != nil {
-		return nil, err
+		if !hasUnsupportedHostError(err) {
+			return nil, err
+		}
 	}
 
 	infra, err := ef.importManager.ProjectInfrastructure(ctx, ef.projectConfig)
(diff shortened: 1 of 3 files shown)


hemarina (approved):
The fix is correct and well-targeted. Root cause analysis is accurate and the approach is sound. No correctness, security, or data-loss issues found.

## Findings

**LOW (nit) — Interface doc comment stale**
The `ProjectManager.Initialize` interface comment (project_manager.go:42–47) still says only "Initializes the project and all child services"; the implementation comment documents the new skip behavior but the public contract doesn't.

**LOW (nit) — PR description count**
Description says "5 cases"; the test has 6 (the `errors.Join` case was added).

Both are optional cosmetic items — nothing blocks merge.

jongio:
@vhvb1989 - Please reopen if you still need this.

PR #8517, opened 2026-06-01
[Agents extension] Filter our model query by the supported hosted agents regions

### LINKED ISSUE #8516: Model query doesn't filter by supported regions

trangevi:
Fixes #8516


therealjohn (changes requested):
I will confirm if the models a hosted agent uses need to be limited to the regions you can deploy a hosted agent too - I do not think they are linked. 

I believe a model deployment can technically be in a different region (supported or not of HA) of hosted agent.

trangevi:
Confirmed, no such restriction

PR #8522, opened 2026-06-02
[azure.ai.agents] Fix init from manifest in CWD: allow copy when target is subpath

### LINKED ISSUE #8523: [azure.ai.agents] azd ai agent init fails with 'cannot copy agent files: target is inside the manifest directory' when run from sample directory
## Description

Running `azd ai agent init` (without flags) from inside a sample directory that contains `agent.manifest.yaml` fails with:

> cannot copy agent files: target is inside the manifest directory

## Repro Steps

1. Clone a sample repo containing agent manifests:
   `git clone https://github.com/azure-samples/foundry-samples`

2. Navigate to a sample directory:
   `cd foundry-samples/samples/python/hosted-agents/agent-framework/responses/01-basic/`

3. Run init:
   `azd ai agent init`

4. When prompted 'An existing agent manifest was found. Use it?' select **Yes**

5. Enter an agent name (e.g. `agent-framework-agent-basic-responses`)

## Expected Behavior

Init should complete successfully, creating a project subdirectory with the agent source files copied in.

## Actual Behavior

Error: `cannot copy agent files: target '...\01-basic\my-agent\src\my-agent' is inside the manifest directory '...\01-basic'`

## Root Cause

The auto-detect flow finds the manifest in CWD, then `ensureProject` creates a project subdirectory under CWD (since no `azure.yaml` exists). The subsequent copy step detects that the target directory is inside the manifest directory (source) and rejects to prevent infinite recursion. The code creates a state it then cannot resolve.

## Workaround

Run from the parent directory with explicit `-m` flag:

`cd .. && azd ai agent init -m ./01-basic/agent.manifest.yaml`

## Environment

- OS: Windows 11
- azd version: 0.1.37-preview (azure.ai.agents extension)

v1212:
Fixes #8523

## Problem

When running `azd ai agent init` from a directory that contains an `agent.manifest.yaml` but no `azure.yaml` in the directory tree (e.g. inside a sample repo like `foundry-samples/.../01-basic/`), the from-manifest flow fails with:

> cannot copy agent files: target is inside the manifest directory

The auto-detect flow finds the manifest, guides the user through setup, then `ensureProject` creates a project subdirectory under CWD. The subsequent copy step detects that the target is inside the manifest directory and rejects.

## `azd ai agent init` - All Cases

| # | CWD State | Manifest | azure.yaml | Path Taken | ensureProject Behavior | Result |
|---|-----------|----------|------------|-----------|----------------------|--------|
| 1 | Empty dir | None | None | Template (user picks) | `azd init -t template` creates subdirectory | OK |
| 2 | Has code | None | None | From-code | `scaffoldTemplate` downloads azure.yaml + infra into CWD (no subdir) | OK |
| 3 | Has code | Yes | Yes (parent) | From-manifest | Finds existing project, no subdir, no copy needed | OK |
| 4 | Has code | Yes | **None** | From-manifest | `azd init -t starter` creates subdirectory then copy | **BUG** |
| 5 | Has code | None | Yes (parent) | From-code | Finds existing project, generates agent.yaml in-place | OK |
| 6 | Empty dir | None | Yes | Template adds to existing project | N/A | OK |
| 7 | Has code | Yes (via `-m url`) | None | From-manifest (remote) | Downloads to temp, no subpath issue | OK |
| 8 | Has code | Yes (via `-m ../path`) | None | From-manifest (external) | manifestDir separate from project dir | OK |

**Only Case 4 triggers the bug** - manifest auto-detected in CWD + no azure.yaml = ensureProject creates subdirectory under manifestDir.

## Root Cause

The from-manifest `ensureProject` uses `azd init -t template subdir` which creates a new directory. The from-code `ensureProject` uses `scaffoldTemplate` which downloads only `infra/` + `azure.yaml` into CWD without a subdirectory. When the manifest is auto-detected in CWD, the from-manifest path creates the project as a child of the manifest directory, making copy impossible without recursion.

## Fix

1. **`validateLocalContainerAgentCopy`**: Allow the subpath case (return nil instead of error)
2. **`copyDirectory`**: When dst is inside src, skip the destination directory during `filepath.WalkDir` traversal using `filepath.SkipDir`, preventing infinite recursion while copying all other source files

## Testing

- Renamed `TestCopyDirectory_RefusesToCopyIntoSubtree` to `TestCopyDirectory_CopiesWithSubpathExclusion`
- Added `TestValidateLocalContainerAgentCopy_AllowsSubpath`
- Updated `init_copy_test.go` subpath test to verify new behavior
- All existing tests pass

## Repro Steps

```bash
git clone https://github.com/azure-samples/foundry-samples
cd foundry-samples/samples/python/hosted-agents/agent-framework/responses/01-basic/
azd ai agent init
# Select 'Yes' when prompted about detected manifest
# Enter agent name
# Error: cannot copy agent files: target is inside the manifest directory
```

## Workaround (older versions)

```bash
cd ..
azd ai agent init -m ./01-basic/agent.manifest.yaml
```

### DIFF
--- a/cli/azd/extensions/azure.ai.agents/internal/cmd/init.go
+++ b/cli/azd/extensions/azure.ai.agents/internal/cmd/init.go
@@ -1064,11 +1064,28 @@ from code-deploy ZIP packaging (uses .gitignore syntax).`,
 					if err := absolutizeRelativeManifestPaths(flags); err != nil {
 						return err
 					}
-					_, statErr := os.Stat(folderName)
-					newlyCreated := errors.Is(statErr, fs.ErrNotExist)
-					targetDir = folderName
-					if newlyCreated && !existingProject {
-						folderDisplay = filepath.ToSlash(folderName)
+
+					// When the manifest lives in the current directory, the agent
+					// source code is already here — treat it like --from-code and
+					// initialize in-place rather than copying files into a new
+					// subdirectory. The existing isSamePath guard in copyDirectory
+					// will skip the copy when src and dst resolve to the same path.
+					manifestInCwd := false
+					if isLocalFilePath(flags.manifestPointer) {
+						if cwd, cwdErr := os.Getwd(); cwdErr == nil {
+							manifestInCwd = isSamePath(filepath.Dir(flags.manifestPointer), cwd)
+						}
+					}
+
+					if !manifestInCwd {
+						_, statErr := os.Stat(folderName)
+						newlyCreated := errors.Is(statErr, fs.ErrNotExist)
+						targetDir = folderName
+						if newlyCreated && !existingProject {
+							folderDisplay = filepath.ToSlash(folderName)
+						}
+					} else if flags.src == "" {
+						flags.src = "."
 					}
 				}
 
@@ -1969,7 +1986,8 @@ func (a *InitAction) configureModelChoice(
 	return agentManifest, nil
 }
 
-func (a *InitAction) isLocalFilePath(path string) bool {
+// isLocalFilePath reports whether path refers to a local file (not an http/https URL).
+func isLocalFilePath(path string) bool {
 	// Check if it starts with http:// or https://
 	if strings.HasPrefix(path, "http://") || strings.HasPrefix(path, "https://") {
 		return false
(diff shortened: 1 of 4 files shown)


banrahan:
I changed how this was implemented, I undid the changes to init_copy, it was re-creating everything as a sub-dir. Instead I added a check for a local manifest in init.go and refactored the isLocalFilePath so I could use it at this point.

trangevi (approved):

PR #8527, opened 2026-06-02
fix: pass ContainerEngine to dotnet publish when Podman is used

### LINKED ISSUE #8525: [Bug] "empty dotnet configuration output" error when publishing Aspire empty project with Kafka packages
## Linked Issue

Upstream: https://github.com/microsoft/aspire/issues/17593

## Description

When using `azd up` to publish an Aspire empty project that has added Kafka packages (`Aspire.Confluent.Kafka` and `Aspire.Hosting.Kafka`), the deployment fails with:

```
ERROR: empty dotnet configuration output
```

The error originates from `cli/azd/pkg/tools/dotnet/dotnet.go` in the `getTargetPort()` function, which parses the output of `dotnet publish` with `/t:PublishContainer`. When the dotnet publish output is empty or missing the `config` JSON property, azd returns this error with a suggestion to enable `EnableSdkContainerSupport`. However, the reporter confirms that adding `<EnableSdkContainerSupport>true</EnableSdkContainerSupport>` to the project file did not resolve the issue.

## Repro Steps

1. Create an Aspire Empty App
2. Add an ASP.NET Core Web App project and reference it from AppHost
3. Add Kafka packages (`Aspire.Confluent.Kafka`, `Aspire.Hosting.Kafka`)
4. Configure Kafka connection in AppHost and Web project
5. Run `azd init` and `azd up`

## Expected

Project publishes successfully.

## Actual

`azd up` fails with `ERROR: empty dotnet configuration output`.

## Environment

- azd version: 1.26.0-beta.1-daily.6353010
- Aspire: 13.4.0 (also repros on 13.3.5)
- OS: Windows 11 25H2 x64
- Container runtime: Podman 5.8.2

vhvb1989:
## Problem

When users have **Podman** instead of Docker, `dotnet publish /t:PublishContainer` defaults to Docker and fails silently — producing empty stdout. This causes azd to error with:

```
ERROR: empty dotnet configuration output
```

Upstream issue: https://github.com/microsoft/aspire/issues/17593

## Root Cause

azd detects the container runtime (Docker vs Podman) in `docker.Cli.CheckInstalled()`, but this information was **never passed** to `dotnet publish`. The .NET SDK supports a `-p:ContainerEngine=podman` MSBuild property, but azd was not setting it.

## Fix

Thread the detected container engine from `docker.Cli` through to the dotnet publish commands:

1. **`docker.Cli`**: Export `ContainerEngine()` getter (wraps existing private `getContainerEngine()`)
2. **`ContainerHelper`**: Add `ContainerEngine()` delegate method
3. **`dotnet.Cli`**: Add `containerEngine` parameter to `BuildContainerLocal` and `PublishContainer`; append `-p:ContainerEngine=podman` when engine is not Docker
4. **Call sites**: Pass the engine at all production call sites (`dotnetContainerAppTarget`, `ContainerHelper`)
5. **Tests**: Add 4 new test cases covering podman/docker/empty engine scenarios

## Testing

- All existing dotnet, docker, and project package tests pass
- New tests verify:
  - Podman engine → `-p:ContainerEngine=podman` is appended
  - Docker engine → no ContainerEngine property appended
  - Empty engine → no ContainerEngine property appended
  - Podman on `PublishContainer` (remote push) path

Fixes #8525

### DIFF
--- a/cli/azd/pkg/project/container_helper.go
+++ b/cli/azd/pkg/project/container_helper.go
@@ -84,6 +84,11 @@ func NewContainerHelper(
 	}
 }
 
+// ContainerEngine returns the detected container engine name ("docker" or "podman").
+func (ch *ContainerHelper) ContainerEngine() string {
+	return ch.docker.ContainerEngine()
+}
+
 // DockerfileBuilder returns a new DockerfileBuilder instance for building Dockerfiles programmatically.
 func (ch *ContainerHelper) DockerfileBuilder() *DockerfileBuilder {
 	return NewDockerfileBuilder()
@@ -1017,7 +1022,8 @@ func (ch *ContainerHelper) runDotnetPublish(
 		imageName,
 		dockerCreds.LoginServer,
 		dockerCreds.Username,
-		dockerCreds.Password)
+		dockerCreds.Password,
+		ch.ContainerEngine())
 	if err != nil {
 		return "", fmt.Errorf("publishing container: %w", err)
 	}
--- a/cli/azd/pkg/project/service_target_dotnet_containerapp.go
+++ b/cli/azd/pkg/project/service_target_dotnet_containerapp.go
@@ -664,7 +664,8 @@ func (at *dotnetContainerAppTarget) prepareDotNetProjectImage(
 			imageName,
 			dockerCreds.LoginServer,
 			dockerCreds.Username,
-			dockerCreds.Password)
+			dockerCreds.Password,
+			at.containerHelper.ContainerEngine())
 		if err != nil {
 			return nil, fmt.Errorf("publishing container: %w", err)
 		}
@@ -700,6 +701,7 @@ func (at *dotnetContainerAppTarget) prepareDotNetMultiStageImage(
 		serviceConfig.Path(),
 		"Release",
 		localImageTag,
+		at.containerHelper.ContainerEngine(),
 	)
 	if err != nil {
 		return nil, fmt.Errorf("building local container: %w", err)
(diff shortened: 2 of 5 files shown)


hemarina (approved):
Nice targeted fix. The property name `-p:ContainerEngine=podman` matches the .NET SDK container build contract (per dotnet/sdk-container-builds and the publish-as-container docs), and the threading from `docker.Cli` → `ContainerHelper` → `dotnet.Cli` lines up with the existing engine-detection path in `docker.Cli.CheckInstalled()`. The `appendContainerEngine` helper has the right defensive shape — `""` and `"docker"` are both no-ops, so the .NET SDK's own default keeps working untouched and the rollout is backward-compatible. All call sites are updated (2 production, all existing test sites bumped by one arg) and the 4 new test cases cover the matrix that matters: podman appends, docker no-op, empty no-op, and the `PublishContainer` (remote push) path in addition to local build.

Two non-blocking thoughts:

1. **Consider a wiring test, not just a unit test.** The new `Test_Cli_ContainerEngine` cases prove `dotnet.Cli` honors whatever engine string it's given. They don't prove the call sites in `container_helper.go` / `service_target_dotnet_containerapp.go` actually pass the *detected* engine vs. an empty string — a future refactor could regress this and only manual testing on a Podman box would catch it. A small test that wires a fake `docker.Cli` with `containerEngine = "podman"` through `ContainerHelper.runDotnetPublish` and asserts the captured exec args include `-p:ContainerEngine=podman` would close that gap.

2. **Redundant exported wrappers.** `docker.Cli.ContainerEngine()` is a one-line passthrough to `getContainerEngine()`, and `ContainerHelper.ContainerEngine()` is a one-line passthrough to `ch.docker.ContainerEngine()`. Either promoting `getContainerEngine` directly to `ContainerEngine` (rename + update internal callers) or inlining the `ContainerHelper` helper would trim two layers without losing anything. Total style nit — fine to leave.

LGTM otherwise.

JeffreyCA (approved):

PR #8529, opened 2026-06-03
Enhance resource group naming with salt to avoid collisions and add t…

### LINKED ISSUE #8528: [Issue] azd ai agent init - it is possible for people sharing a subscription to have their RG collide
azd ai agent init derives the resource group from AZURE_ENV_NAME only (rg-${environmentName}). Two
  users who init the same agent template and accept the default env name end up targeting the same RG
  in the same subscription, silently sharing/overwriting each other's resources.

  Minimal repro

   # User A
   mkdir a && cd a
   azd ai agent init      # Basic Responses (Python), accept defaults
   cd agent-framework-agent-basic-responses && azd provision

   # User B (different machine OR different folder, SAME subscription)
   mkdir b && cd b
   azd ai agent init      # same template, accept defaults
   cd agent-framework-agent-basic-responses && azd provision

  Expected: B gets its own RG (or is warned the env name is taken).
  Actual: Both deploy into rg-agent-framework-agent-basic-responses-dev and stomp on each other.

  Verify:

   grep AZURE_RESOURCE_GROUP a/agent-framework-agent-basic-responses/.azure/*/.env \
                            b/agent-framework-agent-basic-responses/.azure/*/.env
   # → identical RG name

  Root cause

  infra/main.bicep:

   param resourceGroupName string = 'rg-${environmentName}'

  environmentName defaults to the agent name + -dev — same for every user who accepts defaults — so the
   RG collides across users/instances.

  Suggested fix

  Make the default RG (or env name) instance-unique, e.g.
  rg-${environmentName}-${uniqueString(subscription().id, environmentName)} or default environmentName
  to include ${USER}/a salt at init time.

banrahan:
fixes #8528

Mirrored PR #8488's salt approach to the resource group name: 

Refactored ensureResourceTokenSalt to return its salt and added composeSaltedResourceGroupName + ensureResourceGroupName helpers that write a salted AZURE_RESOURCE_GROUP (e.g. rg-{envName}-{salt}) into the azd environment at new-project scaffold time. 

The composition truncates the env name first so the 8-char salt is always preserved within Azure's 90-char RG limit, and the whole thing is best-effort/idempotent. Skipping when salt is empty or AZURE_RESOURCE_GROUP is already set, falling back cleanly to Bicep's original rg-${environmentName} default.


kristenwomack:
Hey @banrahan, took a look at this. Reusing the salt pattern from #8488 keeps things consistent. Tests cover the edges: 78-char boundary, trailing dash/dot trimming, idempotence, skip-when-set, skip-when-empty-salt.

A few small things:

1. **The comment on `resourceGroupEnvKey` is a little off** (Copilot caught part of this):
   ```go
   // resourceGroupEnvKey is the standard azd env var consumed by Bicep's
   // main.parameters.json (`"resourceGroupName": "${AZURE_RESOURCE_GROUP}"`).
   ```
   Two issues: the JSON shape is wrong — parameter files nest under `parameters.<name>.value` — and more importantly `AZURE_RESOURCE_GROUP` isn't just a substitution token. azd reads it directly to scope the deployment. Could you reword so the next person doesn't go looking for a parameter file? Something like *"read by azd's Bicep provider to scope resource-group deployments"* works.

2. **Dead branch in `composeSaltedResourceGroupName`** — the function handles `salt == ""` and there's a test for it, but the godoc says callers shouldn't pass an empty salt, and the only caller short-circuits before then. Either drop the branch and test, or drop the "not exercised" caveat in the comment.

3. **Worth a CHANGELOG line** — `.env` will now contain a salted RG name after init, where before it was empty (Bicep filled in the default). Anyone scripting against `AZURE_RESOURCE_GROUP` right after init will see something different.

Things I checked and are fine: 90-char limit math (3 + 78 + 1 + 8 = 90), the trailing `.` trim (Azure disallows RG names ending in `.`), skip-when-set preserving BYO / `azd env set`, and the random 8-hex salt giving ~4B combos so collisions between two users on the same sub are negligible.

Before this merges, I'd like @JeffreyCA or @vhvb1989 to give it a pass too.

Thanks for picking this up.

trangevi (approved):

Issue #7523, opened 2026-04-06
error executing step command 'init -t Azure-Samples/azd-ai-starter-basic --environment <env name>': environment already initialized to <env name>


therealjohn:
Ran azd ai agent init in an existing directory that already includes an agent.yaml: 

v0.1.20-preview 
Visit the docs at https://aka.ms/azd-ai-agent-docs

? An existing agent manifest was found at "agent.manifest.yaml". Use it?: Yes
Let's get your project initialized.

Initializing an app to run on Azure (azd init)

WARNING: The current directory is not empty.
Initializing an app in this directory may overwrite existing files.

? Continue initializing an app in 'C:\Users\johmil\Projects\hosted-agents-vnext-private-preview\samples\python\agentserver-invocations\echo-agent-streaming'? Yes                                                                                       
  WARNING: The following files are present both locally and in the template:
 * README.md
? What would you like to do with these files? Keep my existing files unchanged                                                                                                                                                                          
                                                                                                                                                                                                                                                        
ERROR: failed to initialize project: rpc error: code = Internal desc = failed to run workflow: error executing step command 'init -t Azure-Samples/azd-ai-starter-basic --environment echo-agent-streaming-dev': environment already initialized to echo-agent-streaming-dev

therealjohn:
I cant reproduce this right now.

PR #8561, opened 2026-06-08
fix: make azd init idempotent with respect to the environment

### LINKED ISSUE #8562: azd init fails with "environment already initialized" when re-running against an existing environment
## Problem

Re-running `azd init -t <template> --environment <name>` in a directory that already has an azd environment hard-fails with:

```
ERROR: failed to initialize project: ... environment already initialized to <name>-dev
```

This forces users — and agent/extension flows that drive `azd init` — into destructive workarounds:

1. `azd env delete <name>` then retry, or
2. Delete the `.azure/` folder in the project directory

Neither is acceptable in automated or agent-driven scenarios where `azd init` should be safely re-runnable.

## Root cause

`initializeEnv` in `cli/azd/cmd/init.go` unconditionally errors whenever a default environment already exists, without checking whether the requested `--environment` matches the existing one.

## Expected behavior

- If the requested environment name matches the existing default, silently reuse it (no prompt needed).
- If running interactively with an ambiguous case, prompt the user: **Reuse (default) / Create new**.
- In non-interactive mode, reuse the default unless an explicit, *different*, valid env name is requested.
- Recover gracefully from partial/orphaned state (e.g. env folder exists but no recorded default, or stale default whose folder is missing).

## Fixed by

PR #8561

huimiu:
Fixes #8562

## Problem

Re-running `azd init -t <template> --environment <name>` in a directory that already has an azd environment hard-failed with:

```
ERROR: failed to initialize project: ... environment already initialized to <name>-dev
```

This forced users (and agent/extension flows that drive `azd init`) into destructive workarounds — `azd env delete <name>` or deleting the `.azure/` folder — just to re-run init. The failure happened in `initializeEnv`, which errored whenever a default environment existed, without ever checking whether the requested `--environment` matched the existing one.

## Change

Replace the unconditional failure with a **reuse-vs-create-new** decision that defaults to reuse:

- **Exact match** (requested `-e` == existing default) → silently reuse, **no prompt**. This is the common case and fixes agent/extension flows that have no `--no-prompt` flag.
- **Interactive + ambiguous** → prompt **Reuse (default) / Create new**, defaulting to reuse.
- **Non-interactive** → reuse the default; only error when an explicit, *different*, still-valid env name is requested (preserves CI safety and the typed `EnvironmentInitError` used by the error middleware).
- **Partial-state recovery** → handles an orphaned env folder with no recorded default, and a stale default whose folder is missing.
- **Non-destructive reuse** → template metadata and `--location`/`--subscription` flags are applied **set-if-absent**, so user edits between runs aren't clobbered while missing values still get repaired.

## Tests

- `cli/azd/cmd/init_env_test.go` (new) — `TestInitializeEnv`, 13 cases covering fresh create, interactive reuse/create-new prompts, exact-match silent reuse, non-interactive reuse/error/recovery, orphan/stale recovery, metadata-not-clobbered (with disk reload), and flag set-if-absent.
- `cli/azd/test/functional/init_test.go` — `Test_CLI_Init_Idempotent_Environment`: `init -e TESTENV` → `init -e TESTENV` (reuses, succeeds) → `init -e OTHERENV --no-prompt` (still errors).

## Validation

`go build` clean · `golangci-lint` 0 issues · `gofmt -s` clean · `cspell` (project config) 0 issues · copyright headers present.

## Known limitation

In `--no-prompt` mode with an *empty* requested name where the auto-derived name collides with an orphan, recovery isn't attempted — the derived name lives inside `envManager.Create` and isn't retrievable by the caller. Callers that pass an explicit `--environment` (the agent/extension path) are unaffected.

### DIFF
--- a/cli/azd/cmd/init.go
+++ b/cli/azd/cmd/init.go
@@ -804,10 +804,196 @@ func (i *initAction) initializeEnv(
 		return nil, fmt.Errorf("retrieving default environment name: %w", err)
 	}
 
-	if envName != "" {
-		return nil, environment.NewEnvironmentInitError(envName)
+	// Environment manager requires azd context
+	// Azd context isn't available in init so lazy instantiating
+	// it here after the template is hydrated and the context is available
+	envManager, err := i.lazyEnvManager.GetValue()
+	if err != nil {
+		return nil, err
 	}
 
+	env, reused, err := i.resolveInitEnv(ctx, azdCtx, envManager, envName, i.flags.EnvironmentName)
+	if err != nil {
+		return nil, err
+	}
+
+	// Record the resolved environment as the default. This is a no-op when it already
+	// is, and repairs a missing or stale default when reusing/recovering an existing
+	// environment from a previous (possibly partial) initialization.
+	if err := azdCtx.SetProjectState(azdcontext.ProjectState{DefaultEnvironment: env.Name()}); err != nil {
+		return nil, fmt.Errorf("saving default environment: %w", err)
+	}
+
+	if reused {
+		i.console.Message(
+			ctx,
+			fmt.Sprintf("Reusing existing environment %s.", output.WithHighLightFormat(env.Name())),
+		)
+	}
+
+	// When reusing an environment, only fill in values that are absent so changes the
+	// user made between runs aren't clobbered. On a fresh create, apply unconditionally.
+	existingDotenv := env.Dotenv()
+	setEnvValue := func(key, value string) {
+		if reused {
+			if _, ok := existingDotenv[key]; ok {
+				return
+			}
+		}
+		env.DotenvSet(key, value)
+	}
+
+	// Copy template metadata into environment values
+	for key, value := range templateMetadata.Variables {
+		setEnvValue(key, value)
+	}
+
+	for key, value := range templateMetadata.Config {
+		if reused {
+			if _, ok := env.Config.Get(key); ok {
+				continue
+			}
+		}
+		if err := env.Config.Set(key, value); err != nil {
+			return nil, fmt.Errorf("setting environment config: %w", err)
+		}
+	}
+
+	initialValuesFromEnv, err := repository.InitEnvFileValues()
+	if err != nil {
+		return nil, fmt.Errorf("loading initial env file values: %w", err)
+	}
+	for key, value := range initialValuesFromEnv {
+		setEnvValue(key, value)
+	}
+
+	// Apply subscription/location flags. On create these were already set from the
+	// environment spec; on reuse this repairs a partially initialized environment that
+	// is missing them, without clobbering values the user already set.
+	if i.flags.location != "" {
+		setEnvValue(environment.LocationEnvVarName, i.flags.location)
+	}
+	if i.flags.subscription != "" {
+		setEnvValue(environment.SubscriptionIdEnvVarName, i.flags.subscription)
+	}
+
+	if err := envManager.Save(ctx, env); err != nil {
+		return nil, fmt.Errorf("saving environment: %w", err)
+	}
+
+	return env, nil
+}
+
+// resolveInitEnv determines which environment `azd init` should use, returning the
+// environment and whether it was reused (true) versus newly created (false).
+//
+// When a default environment already exists, init is idempotent rather than failing:
+// re-running with the same environment name reuses it, interactive runs without an
+// exact match are asked whether to reuse it or create a new one (defaulting to reuse),
+// and non-interactive runs reuse it unless a different, still-valid environment name was
+// explicitly requested. Orphaned environment folders and stale default configuration
+// left behind by a previous partial init are recovered instead of dead-ending.
+func (i *initAction) resolveInitEnv(
+	ctx context.Context,
+	azdCtx *azdcontext.AzdContext,
+	envManager environment.Manager,
+	defaultEnvName string,
+	requested string,
+) (*environment.Environment, bool, error) {
+	// No default environment is recorded for the project yet. Reuse an orphaned
+	// environment folder left by a previous partial init when the requested name matches
+	// one, otherwise create a new environment.
+	if defaultEnvName == "" {
+		return i.getOrCreateEnv(ctx, azdCtx, envManager, requested)
+	}
+
+	// The caller explicitly asked for the environment that already exists. Reuse it
+	// without prompting (covers re-running init with the same -e value).
+	if requested != "" && requested == defaultEnvName {
+		return i.getOrCreateEnv(ctx, azdCtx, envManager, defaultEnvName)
+	}
+
+	// A different (or no) environment was requested while a default exists.
+	if i.console.IsNoPromptMode() {
+		// Non-interactive runs can't prompt. Reuse the default unless a different
+		// environment was explicitly requested and the default is actually usable; a
+		// stale default left by a partial init shouldn't block an explicit request.
+		if requested != "" {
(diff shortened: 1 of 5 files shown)


cli/azd/cmd/init.go
```diff
@@ -869,6 +884,142 @@ func (i *initAction) initializeEnv(
...
+		// Non-interactive runs can't prompt. Reuse the default unless a different
+		// environment was explicitly requested and the default is actually usable; a
+		// stale default left by a partial init shouldn't block an explicit request.
+		if requested != "" {
+			if _, err := envManager.Get(ctx, defaultEnvName); err == nil {
+				return nil, false, environment.NewEnvironmentInitError(defaultEnvName)
+			}
+
+			return i.getOrCreateEnv(ctx, azdCtx, envManager, requested)
+		}
```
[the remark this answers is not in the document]
huimiu: Fixed in 590e0129b: validate requested name first and narrow the stale-default bypass to `errors.Is(err, environment.ErrNotFound)`.

Changes:
1. **Invalid name → clear error**: Added `environment.IsValidEnvironmentName(requested)` check before the default-env lookup so an invalid `-e` value yields an `InvalidEnvironmentNameError` instead of the misleading "already initialized" error.
2. **Real errors surface**: Replaced the `err != nil` bypass with a `switch` on `errors.Is(err, environment.ErrNotFound)` — real I/O/config errors now propagate rather than silently falling through to create the requested env.

Two new test cases cover both paths: `NonInteractiveInvalidRequestedNameErrors` and `NonInteractiveCorruptDefaultEnvPropagatesError`.

therealjohn (approved):

glharper (changes requested):
1. Deliberate behavior change for --no-prompt with no -e. Previously this errored; now it silently reuses the recorded default. Which is a change for CI/scripts. This should get a CHANGELOG.md entry and ideally a one-liner in the azd init docs.
 2. "Create new" + explicit name that already exists → reports "Reusing". In the interactive path, if the user picks Create a new environment but -e names an existing (non-default) env, getOrCreateEnv does a Get, succeeds, returns reused=true, and prints "Reusing existing environment…" while switching the default to it. Behavior is defensible (it switches + reuses), but the message contradicts the user's "create new" choice. Minor UX wrinkle — consider re-prompting for a name on collision, or adjusting the message.

cli/azd/cmd/init.go
```diff
@@ -869,6 +884,155 @@ func (i *initAction) initializeEnv(
...
+	reuseOption := fmt.Sprintf("Reuse the existing environment '%s'", defaultEnvName)
+	createOption := "Create a new environment"
+	choice, err := i.console.Select(ctx, input.ConsoleOptions{
+		Message: fmt.Sprintf(
+			"An environment named '%s' already exists for this project. How do you want to proceed?",
+			defaultEnvName,
+		),
+		Options:      []string{reuseOption, createOption},
+		DefaultValue: reuseOption,
+	})
```
JeffreyCA, line 1036: When the user explicitly specifies an environment that doesn't exist, I feel like we should just use that and not prompt or default to a totally different environment:

image

In the above case, `test-env4` doesn't exist, `test-env3` exists, but it's guiding the user to use `test-env3` instead

JeffreyCA (commented):
Thanks for the fixes! Wondering if something like this would feel better:


| `-e` given? | Requested env exists? | Default exists? | Suggested behaviour |
|---|---|---|---|
| **Yes** | **Yes** | **Yes — same as requested** | **Confirm** "Reuse existing environment '<name>'?" (default **Yes**) → **Yes**: reuse, no extra note (the confirm is the acknowledgment). **No**: gracefully cancel init (nothing mutated) |
| **Yes** | **Yes** | **No / different default** | **Confirm** "Reuse existing environment '<name>'?" (default **Yes**) → **Yes**: reuse + promote to default, emit info note that the default was switched. **No**: gracefully cancel init (nothing mutated) |
| **Yes** | **No** (new name) | No | **No prompt** — create `<name>`, set as default |
| **Yes** | **No** (new name) | Yes (different) | **No prompt** — create `<name>`, set as default; emit info note that the previous default is being switched |
| **No** | — | Yes | Prompt "Reuse existing '<default>' / Create new" (default **Reuse**) — the genuinely ambiguous case |
| **No** | — | No | Create / derive a name as today (prompt for name if interactive) |

jongio:
﻿**My read:** Re-running `azd init -e <name>` should reuse when the name matches, error when it conflicts in CI, and prompt when ambiguous. This PR implements that decision tree with clean recovery for partial-state edge cases.

**What's compelling:** The resolution logic in `resolveInitEnv` is well-layered - exact match first, then no-prompt gate, then interactive prompt. Test coverage is thorough, especially the corruption propagation test and the disk-reload verification in the metadata test.

**Concerns:**

- **Explicit CLI flags are silently ignored on reuse.** The `setEnvValue` helper treats `--location` and `--subscription` flags identically to template metadata - both use set-if-absent. The test `ReuseFillsAbsentLocationButKeepsExistingSubscription` demonstrates this: passing `--subscription sub-from-flag` is silently dropped when the env already has a subscription. This breaks CLI convention where an explicit flag means `I want this NOW.` Template metadata should be set-if-absent (protecting user edits between runs). Explicit flags should always win (the user just typed them).

**Counter-proposals:**

1. **Split the merge semantics by source** - Keep `setEnvValue` (set-if-absent) for template metadata and `InitEnvFileValues`. For flag-derived values, bypass the reuse guard and call `env.DotenvSet` directly:

```go
// Flags represent explicit user intent - always apply.
if i.flags.location != "" {
    env.DotenvSet(environment.LocationEnvVarName, i.flags.location)
}
if i.flags.subscription != "" {
    env.DotenvSet(environment.SubscriptionIdEnvVarName, i.flags.subscription)
}
```

This preserves the `don't clobber user edits from template defaults` guarantee while respecting explicit flag intent. The test would change to assert `sub-from-flag` wins over `sub-already-set` when the flag is explicitly passed.

**Recommended phasing:** @glharper's CHANGELOG point is a quick add. The flag semantics fix above is small and makes reuse behavior consistent with standard CLI expectations. @JeffreyCA's decision table refinements (skip prompt when `-e newname` and `newname` doesn't exist; confirm when reusing a non-default env) are solid UX improvements but can land as a follow-up without blocking this fix for #7523.

cli/azd/cmd/init.go
huimiu: Agreed! Now when `-e` names an env that does not exist, we just create it and make it the default, no prompt. Thanks for the catch!

huimiu:
@JeffreyCA Love this, much clearer. Implemented the full table in \
esolveInitEnv\ (Case A/B/C): confirm to reuse for an existing \-e\ (decline cancels cleanly), create and promote for a new \-e\, and a reuse vs create prompt when no \-e\. 🙏

huimiu:
@glharper Thanks! Both addressed:
1. Added a CHANGELOG entry plus an `azd init` help note for the `--no-prompt` reuse behavior.
2. Fixed the messaging, now exactly one line prints after save, so "Create new" never reports "Reusing".

huimiu:
@jongio Good call on the flag semantics! Fixed: `--location` and `--subscription` now call `env.DotenvSet` directly (unconditionally) on reuse, so explicit flags always win. Template metadata and `InitEnvFileValues` still use set-if-absent to protect user edits. Thanks for the clear breakdown!

glharper (approved):

JeffreyCA (approved):

PR #8571, opened 2026-06-09
fix: reduce S0 AccountCount MinCapacity from 2 to 1

### LINKED ISSUE #8563: azd up still fails on free subscriptions after fix #8537 — is OpenAI.S0.AccountCount limit=1 the blocker?
###  Problem

  After upgrading to azd 1.25.5 (which includes the fix from PR #8537 (https://github.com/microsoft/customer-chatbot-solution-accelerator/issues/8537) for #8533 (https://github.com/microsoft/customer-chatbot-solution-accelerator/issues/8533)), azd up still fails on free Azure subscriptions with the same quota error when deploying templates with @metadata usageName.

###   Error

   ERROR: prompting for value: getting locations with quota: no location found with enough quota for 
   OpenAI.GlobalStandard.gpt4.1-mini ( Cap: 1 ) and OpenAI.GlobalStandard.gpt-realtime-mini ( Cap: 1 )

###   Environment

   - azd version: 1.25.5 (stable)
   - Subscription type: Azure Free (Pay-As-You-Go not enabled)
   - Subscription ID: 6becd867-e9d1-4ed9-a7a7-5bf1260XXXXX
   - Template: customer-chatbot-solution-accelerator (https://github.com/microsoft/customer-chatbot-solution-accelerator/)

###   What we observed

  We ran azd up --debug and noticed the /usages API is no longer returning empty — it returns 200 OK with 212 items of real quota data. So the empty-response fix from #8537 (https://github.com/microsoft/customer-chatbot-solution-accelerator/issues/8537) is not the code path being hit.

  We then called the API directly to check model quota:

   $token = azd auth token --scope https://management.azure.com/.default
   $headers = @{Authorization="Bearer $token"}
   $response = Invoke-RestMethod -Uri "https://management.azure.com/subscriptions/{subId}/providers/Microsoft.CognitiveServices/locations/eastus2/usages?api-version=2025-06-01" -Headers $headers
   $response.value | Where-Object { $_.name.value -like "*gpt*" -or $_.name.value -eq "OpenAI.S0.AccountCount" }

  Results:

  - OpenAI.GlobalStandard.gpt4.1-mini → Limit: 5000, Current: 100, Remaining: 4900 
 - OpenAI.GlobalStandard.gpt-realtime-mini → Limit: 20, Current: 2, Remaining: 18 
 - OpenAI.S0.AccountCount → Limit: 1, Current: 0, Remaining: 1 

  Model quota is clearly sufficient. However, we noticed that OpenAI.S0.AccountCount has a limit of only 1 on this free subscription (same across all regions — eastus2, francecentral, swedencentral).

###   Question

  We see in prompt.go (https://github.com/Azure/azure-dev/blob/main/cli/azd/pkg/infra/provisioning/bicep/prompt.go) that locationsWithQuotaFor() injects a hardcoded requirement:

   requirements := []ai.QuotaRequirement{
       {UsageName: "OpenAI.S0.AccountCount", MinCapacity: 2},
   }

  Is this MinCapacity: 2 the reason free subscriptions are being rejected? Since the API reports OpenAI.S0.AccountCount remaining=1 and the code requires 2, every location would fail the check — even though the actual model quota is more than enough.

  How can free-tier users verify this is their blocker? We couldn't find OpenAI.S0.AccountCount in the Azure Portal Quotas page . The only way we found to check is via the REST API call above.

  Should MinCapacity be reduced to 1? Our deployment only creates 1 AI Services account and succeeds when we remove the usageName metadata (bypassing the quota check entirely).

 ###  Workaround

  Commenting out the usageName array from the @metadata block bypasses the quota pre-check and deployment succeeds:

  Before (fails):

   @metadata({
     azd:{
       type: 'location'
       usageName: [
         'OpenAI.GlobalStandard.gpt4.1-mini,50'
         'OpenAI.GlobalStandard.gpt-realtime-mini,1'
       ]
     }
   })

  After (works — deployment completes successfully):

   @metadata({
     azd:{
       type: 'location'
       // usageName: [
       //   'OpenAI.GlobalStandard.gpt4.1-mini,50'
       //   'OpenAI.GlobalStandard.gpt-realtime-mini,1'
       // ]
     }
   })

  This confirms the model quota is actually available and only the azd pre-check is blocking the deployment.

  Related

   - Original issue: #8533 (https://github.com/microsoft/customer-chatbot-solution-accelerator/issues/8533)
   - Fix PR: #8537 (https://github.com/microsoft/customer-chatbot-solution-accelerator/issues/8537)

Image

vhvb1989:
## Problem

After the fix in #8537 for empty `/usages` responses, free-tier Azure subscriptions still fail the AI quota pre-check (issue #8563). The `OpenAI.S0.AccountCount` quota has `Limit=1` on free subscriptions, but the code requires `MinCapacity: 2`, rejecting every location.

## Root Cause

PR #4929 changed the S0 account quota check from `remaining >= 1` to `remaining >= 2` with the comment *"The minimum quota for the S0 SKU in Microsoft.CognitiveServices/accounts is 2 capacity units."* However:

- **No documentation or data backs this claim** — the PR had zero review threads and no linked repro.
- **Azure docs confirm 1 account = 1 unit** — `OpenAI.S0.AccountCount` counts distinct resource instances; creating one S0 account consumes exactly 1 unit.
- **Free subs have `Limit=1`** — so `remaining(1) < MinCapacity(2)` always fails.
- **Paid subs mask the bug** — they have limits of 200+, so `>= 2` is trivially satisfied.
- **The issue reporter** successfully deploys when bypassing the quota check, confirming only 1 unit is needed.

## Fix

Change `MinCapacity` from `2` to `1` in `prompt.go` for the implicit `OpenAI.S0.AccountCount` requirement.

Fixes #8563

### DIFF
--- a/cli/azd/pkg/infra/provisioning/bicep/prompt.go
+++ b/cli/azd/pkg/infra/provisioning/bicep/prompt.go
@@ -82,16 +82,18 @@ func autoGenerate(parameter string, azdMetadata azure.AzdMetadata) (string, erro
 // locationsWithQuotaFor finds locations that have sufficient quota for the given usage requirements.
 //
 // The quotaFor parameter uses the Bicep metadata format: "UsageName" or "UsageName, Capacity".
-// An implicit requirement for "OpenAI.S0.AccountCount" with capacity 2 is always included.
+// An implicit requirement for "OpenAI.S0.AccountCount" with capacity 1 is always included.
 func (a *BicepProvider) locationsWithQuotaFor(
 	ctx context.Context, subId string, locations []string, quotaFor []string) ([]string, error) {
 	if a.aiModelService == nil {
 		return nil, fmt.Errorf("AI model service is not configured")
 	}
 
-	// Always require minimum S0 account quota
+	// Always require at least 1 remaining S0 account slot.
+	// Each CognitiveServices/accounts resource consumes exactly 1 unit of
+	// the OpenAI.S0.AccountCount quota regardless of subscription tier.
 	requirements := []ai.QuotaRequirement{
-		{UsageName: "OpenAI.S0.AccountCount", MinCapacity: 2},
+		{UsageName: "OpenAI.S0.AccountCount", MinCapacity: 1},
 	}
 
 	for _, definedUsageName := range quotaFor {


JeffreyCA (approved):
Confirmed this fixes the issue on a personal free trial subscription, but I had to remove the `gpt-realtime-mini` model deployment as I actually had 0 quota for it

Issue #8065, opened 2026-05-06
Make azd ai agent init Bicep-less by default; add --infra to eject


therealjohn:
We want the `azure.ai.agents` extension to default to **Bicep-less** for `azd ai agent init` projects. The infrastructure templates that live in `Azure-Samples/azd-ai-starter-basic` today move into the extension itself, get synthesized in memory at `azd provision` time from `azure.yaml`, and only hit disk when the developer asks for it with `azd ai agent init --infra`.

Today, `azd ai agent init` clones a ~300-line `infra/main.bicep` that pre-bakes Storage, Search, Bing, ACR, and monitoring whether the developer uses them or not. Every shipped extension version reads from the same template repo, so we can't slim that Bicep without breaking projects initialized by older extensions. Two coupled problems: the on-disk templates are bloated, and there's no in-place fix. This RFC is the mechanism that breaks the coupling.

Scope: the requirements for Bicep-less default, the eject command (`azd ai agent init --infra`), the version-stability contract, and how we retire the existing `Azure-Samples/azd-ai-starter-basic` story. Out of scope: the unified `azure.yaml` schema (covered in [Azure/azure-dev#7962](https://github.com/Azure/azure-dev/issues/7962)) and incremental composition (covered in [Azure/azure-dev#8049](https://github.com/Azure/azure-dev/issues/8049)).

Dependencies: this RFC assumes the consolidated `azure.ai.project` and `azure.ai.agent` host kinds from [Azure/azure-dev#7962](https://github.com/Azure/azure-dev/issues/7962). The synthesis input is the unified `azure.yaml`; without it, there's no source of truth to synthesize from.

## Current Problems

1. **Init clones a starter repo from GitHub on every new project.** Before the developer has written any agent code, they get a full Bicep tree and an `infra/` directory -- regardless of which resources they actually need.

2. **The starter template is a program in template clothing.** It's roughly 300 lines of conditionals: `shouldCreateAcr`, `useExistingAiProject ? X : Y`, dual-path outputs for new-vs-existing on every resource, 30-plus outputs to cover every permutation. That's business logic dressed up as IaC, written to make `init` work, not to serve the developer's day-to-day needs.

[...]

Issue #8049, opened 2026-05-05
Add connections, models, tools, and skills to Foundry Agent projects after init


therealjohn:
We need `azd ai project add` commands that let developers grow their Foundry agent projects after init. Today, if you didn't pick the right tools and connections up front, you're left reverse-engineering samples and hand-editing YAML until something works.

**Scope:** `azd ai project add` commands for tools, connections, model deployments, and skills. Config-only -- the CLI updates `azure.yaml` and tells you what code or infra changes to make. No code generation, no Bicep generation. Remove is done with AZD down.

**Depends on:** [Unify Foundry agent configuration in azure.yaml](https://github.com/Azure/azure-dev/issues/7962). The `add` commands modify the `azure.ai.project` service's `config` block, which is the single source of truth for project-scoped Foundry resources. That restructuring must land first.

---

## Current Problems

1. **Init is a one-shot deal.** You pick everything at `azd ai agent init` time and that's it. No CLI command to add a tool or connection afterward. `npm init` has `npm install`. `dotnet new` has `dotnet add`. We have nothing.

2. **The workaround is "go find a sample."** When someone needs to bolt on Azure AI Search or an MCP tool post-init, they dig through docs, find a sample that happens to use that component, extract the relevant bits, hand-edit YAML, and hope the wiring works. This is where most people give up.

3. **Three files, two templating syntaxes, four layers deep.** Adding a single MCP tool today means touching `agent.yaml` (tool definition), `agent.manifest.yaml` (parameter declaration with `{{handlebars}}`), azure.yaml config (connection with `${ENV}` refs), and possibly Bicep. Nobody can answer "where do I make this change?" without knowing the extension's internals. The init code that processes manifest resources spans over 200 lines across `extractToolboxAndConnectionConfigs` and `extractConnectionConfigs` in [`init.go`](https://github.com/Azure/azure-dev/blob/main/cli/azd/extensions/azure.ai.agents/internal/cmd/init.go) -- 200 lines of wiring that developers are expected to replicate by hand.

4. **Zero discoverability.** There's no help listing available capabilities. Developers have no way to find out what they can add, what each capability requires, or what the config should look like. The extension's registered commands ([`root.go`](https://github.com/Azure/azure-dev/blob/main/cli/azd/extensions/azure.ai.agents/internal/cmd/root.go)) include init, run, invoke, show, monitor, session, files, mcp -- nothing for composition after init.

---

## Solution

[...]

PR #7482, opened 2026-04-02
feat: Add provisioning provider support to extension framework

wbreza:
## What Changed

Extensions can now register **custom provisioning providers** via the azd extension framework -- enabling alternatives to the built-in Bicep and Terraform providers. An extension author calls `WithProvisioningProvider("myprov", factory)` on the `ExtensionHost`, and end users set `infra: { provider: myprov }` in `azure.yaml` to use it.

This follows the same architecture established by custom service targets and framework services: proto definition, MessageBroker-based bidirectional streaming, ExtensionHost fluent API, server-side gRPC handler, IoC container registration, and core-side adapter.

**Extension author experience:**

```go
host := azdext.NewExtensionHost(client)
host.WithProvisioningProvider("myprov", func() azdext.ProvisioningProvider {
    return NewMyProvider(client)
}).Run(ctx)
```

**End user experience:**

```yaml
# azure.yaml
infra:
  provider: myprov
  config:
    # extension-specific configuration
```

The full provisioning lifecycle is supported: Initialize, State, Deploy, Preview, Destroy, EnsureEnv, Parameters, and PlannedOutputs -- with in-band progress streaming for Deploy/Preview/Destroy operations. Multiple providers per extension are supported with request-level routing via `provider_name`.

A demo provisioning provider is included in `microsoft.azd.demo` for reference.

## How It Works

The implementation mirrors the established service target and framework service extension patterns. A new `provisioning.proto` defines a bidirectional streaming RPC with a `ProvisioningMessage` oneof envelope containing all request/response pairs plus progress messages. The `ProvisioningManager` on the extension side uses a factories/instances two-map pattern for multi-provider support with lazy initialization and double-checked locking. On the server side, `ProvisioningService` creates a `MessageBroker` per stream, registers providers into the IoC container as named transients, and routes requests via `provider_name` fields on each message. The `ExternalProvisioningProvider` adapter implements the core `provisioning.Provider` interface and delegates all operations through the broker with full type conversion between proto and Go types.

Key design decisions: project-level provider routing (not per-service like service targets), `Options.Config` pass-through via `google.protobuf.Struct` for extension-specific configuration, and forward-compatibility fields for multi-layer deployments (`hint`, `name`, `virtual_env`).

## Issue References

Resolves #7465
Resolves #7466
Resolves #7467
Resolves #7468
Resolves #7469
Resolves #7470
Resolves #7471
Resolves #7472
Resolves #7473
Resolves #7474
Resolves #7475
Resolves #7476
[...]

PR #8577, opened 2026-06-09
docs(spec): Bicep-less Foundry agent init design (#8065)

hund030:
Design spec for [RFC #8065](https://github.com/Azure/azure-dev/issues/8065) — make `azd ai agent init` Bicep-less by default, with the `azure.ai.agents` extension owning provisioning via a custom provider named `microsoft.foundry`.

## Summary

Moves infrastructure templates from `Azure-Samples/azd-ai-starter-basic` into the `azure.ai.agents` extension binary. `azd ai agent init` produces only `azure.yaml` and an agent code project — no `infra/` directory. At `azd provision` time, the extension's own provisioning provider synthesizes Bicep in memory from `azure.yaml` and applies it. `azd ai agent init --infra` ejects on demand: same synthesis, written to `./infra/`; subsequent provisions read from disk.

The `azure.yaml` shape is fixed by the [Foundry `azure.yaml` reference](https://github.com/therealjohn/foundry-azd-config-preview/blob/main/REFERENCE.md): a single `host: microsoft.foundry` service per project, with nested `agents:`, `deployments:`, `connections:`, etc. This spec only changes how that file is *provisioned*; it does not redesign the YAML.

## Key design decisions

- **`host: microsoft.foundry`** — one consolidated service per Foundry project, with nested `agents[]`, `deployments[]`, `connections[]`, `toolboxes[]`, `skills[]`, `routines[]`. Matches the reference doc.
- **`infra.provider: microsoft.foundry`** — explicit declaration required until service-host-driven auto-routing lands. Reuses the custom provisioning provider framework from #7482 (merged).
- **Brownfield via `endpoint:` URL** — not ARM resource ID. Matches what Portal and `az` CLI show; deploy verb resolves ARM ID from endpoint when it needs control-plane access.
- **Three deploy modes** per agent: `docker:`, `runtime:`, or `image:` (pre-built container). Exactly one is required; validator rejects two-or-none.
- **On-disk reuse after eject**: extension reads `./infra/` when present, synthesizes otherwise. `azure.yaml` is never mutated by eject.
- **No `--force` flag** for eject. To regenerate, the user deletes `./infra/` and re-runs the command.
- **`Preview` via ARM What-If** — mirrors Core's Bicep provider (`scope.go:132` uses `WhatIfDeployToResourceGroup`).

## Scope

**In scope:** Bicep-less default behavior, eject command, embedded templates, ARM-backed synthesis only (Foundry project + model deployments + ACR when needed), schema relaxation to allow extension-named providers in `infra.provider`.

**Out of scope:**
- **`azd deploy`** — agent code push and data-plane reconciliation (connections, toolboxes, skills, routines, agent definitions) are deploy's job, not provisioning's. This spec ends at "ARM resources are in place."
- **Data-plane YAML fields** — `connections:`, `toolboxes:`, `skills:`, `routines:`, agent-level `tools:`/`skill:`, `$ref:` resolution. Synthesizer reads them only to skip them; deploy verb owns them.
- Service-host-driven auto-routing (RFC Core Ask 1) — would remove the explicit `infra.provider:`. Defers to a future spec.
- Unified `azure.yaml` schema (#7962); incremental composition (#8049); coexistence with non-Foundry services (use `infra.layers[]` as escape hatch).

**Core changes collapsed.** The original RFC asked Core to surface `services.<svc>.uses` and a typed `services.<svc>.runtime` on the extension-facing proto. With nested `agents[]`, runtime lives inside the service body (read via `additional_properties`); no proto/struct/mapper plumbing needed. Down to one Core change: relax the `infra.provider` enum.

## Related

- RFC: #8065
- Builds on: #7482 (custom provisioning provider framework, merged)
- Reference YAML: [therealjohn/foundry-azd-config-preview](https://github.com/therealjohn/foundry-azd-config-preview/blob/main/REFERENCE.md)

## Notes for reviewers

**Doc-only PR.** Adds `docs/specs/bicepless-foundry/spec.md`. No code changes — the spec is the implementation contract; code PRs follow.

Particular attention welcomed on:
1. The "What the synthesizer ignores" table — is the data-plane / ARM split drawn at the right line?
2. The Open Question and its proposed answer (drift detection on `Deploy()`).
3. The eject "delete-and-rerun" model with no `--force` flag — does this match azd's broader UX posture?

### DIFF
--- a/docs/specs/bicepless-foundry/spec.md
+++ b/docs/specs/bicepless-foundry/spec.md
@@ -2,13 +2,12 @@
 
 ## Problem
 
-`azd ai agent init` clones `Azure-Samples/azd-ai-starter-basic` into every new
-project, dropping ~300 lines of conditional Bicep (`shouldCreateAcr`,
-`useExistingAiProject ? X : Y`, 30+ outputs) on the developer's disk before they
-write any agent code. The starter Bicep lives in a sample repo on `main`, so
-every shipped extension version reads from the same template — slimming or
-tailoring the template breaks every project initialized by every prior extension
-build. The on-disk template is bloated; there is no in-place fix.
+`azd ai agent init` clones `Azure-Samples/azd-ai-starter-basic` into every
+new project, dropping ~300 lines of conditional Bicep (`shouldCreateAcr`,
+`useExistingAiProject ? X : Y`, 30+ outputs) on the developer's disk before
+they write any agent code. The starter Bicep lives in a sample repo on
+`main`, so slimming it would break every project initialized by every prior
+extension build.
 
 See RFC [#8065](https://github.com/Azure/azure-dev/issues/8065) for the full
 problem statement.
@@ -18,15 +17,12 @@ problem statement.
 Move infrastructure templates from `Azure-Samples/azd-ai-starter-basic` into
 the `azure.ai.agents` extension binary. `azd ai agent init` produces only
 `azure.yaml` and an agent code project — no `infra/` directory. At provision
-time, the extension's own provisioning provider synthesizes Bicep in memory
-from `azure.yaml` and applies it. `azd ai agent init --infra` ejects on demand:
-the same synthesis writes Bicep to `./infra/`, and subsequent provisions read
-from disk.
-
-The mechanism is the **custom provisioning provider** capability merged in
-[PR #7482](https://github.com/Azure/azure-dev/pull/7482). The extension
-registers itself by name; the developer declares it in `azure.yaml` as
-`infra.provider: azure.ai.agents`.
+time, the extension's own provisioning provider (registered via the
+[PR #7482](https://github.com/Azure/azure-dev/pull/7482) framework)
+synthesizes Bicep in memory from `azure.yaml` and applies it.
+`azd ai agent init --infra` ejects on demand: the same synthesis writes Bicep
+to `./infra/`, and subsequent provisions read from disk. The developer opts
+in by declaring `infra.provider: azure.ai.agents` in `azure.yaml`.
 
 ## Scope
 
@@ -74,26 +70,28 @@ cli/azd/extensions/azure.ai.agents/
     parameters.go                  ← parameter resolution (env vars, prompts)
 
 cli/azd/pkg/                       ← Core changes (small)
-  project/mapper_registry.go       ← +Uses, +Runtime on ServiceConfig→proto
-  project/service_config.go        ← +Runtime AppServiceRuntime field
+  project/service_runtime.go       ← NEW: ServiceRuntime type (no Stack enum)
+  project/service_config.go        ← +Runtime *ServiceRuntime (Uses already present)
+  project/mapper_registry.go       ← +Uses, +Runtime in ServiceConfig↔proto mappers
   infra/provisioning/provider.go   ← (no change needed)
 
 cli/azd/grpc/proto/
-  models.proto                     ← +runtime, +uses on ServiceConfig message
+  models.proto                     ← +uses, +runtime (typed) on ServiceConfig message
 
-schemas/v1.0/azure.yaml.json       ← relax infra.provider enum → examples
+schemas/v1.0/azure.yaml.json       ← +runtime under services.<svc> (uses already present);
+                                     relax infra.provider enum → examples
 ```
 
-## Provider Resolution (Verified Against Code)
+## Provider Resolution
 
 The extension's provider plugs into the existing IoC-registered factory.
-Provider selection logic (`cli/azd/pkg/infra/provisioning/manager.go:505-540`)
-is unchanged:
+Provider selection (`cli/azd/pkg/infra/provisioning/manager.go:505-540`) is
+unchanged:
 
 ```go
 providerKey := m.options.Provider                    // from azure.yaml infra.provider
 if providerKey == NotSpecified {
-    defaultProvider, _ := m.defaultProvider()        // returns "bicep"
+    defaultProvider, _ := m.defaultProvider()        // "bicep"
     providerKey = defaultProvider
 }
 err = m.serviceLocator.ResolveNamed(string(providerKey), &provider)
@@ -102,11 +100,9 @@ err = m.serviceLocator.ResolveNamed(string(providerKey), &provider)
 Built-in providers register at `cli/azd/pkg/azd/default.go:79-87`. Extension
 providers register at runtime via `RegisterProvisioningProviderRequest`
 (`cli/azd/internal/grpcserver/provisioning_service.go:138-152`) into the same
-`*ioc.NestedContainer`. From the resolver's perspective `bicep` and
-`azure.ai.agents` are equivalent keys.
-
-`ParseProvider` was relaxed in PR #7482 to accept any string
-(`cli/azd/pkg/infra/provisioning/provisioning.go:53-57`).
+container. `bicep` and `azure.ai.agents` are equivalent keys to the resolver.
+`ParseProvider` (`cli/azd/pkg/infra/provisioning/provisioning.go:53-57`) was
+relaxed in PR #7482 to accept any string.
 
 ## Explicit `infra.provider:` Declaration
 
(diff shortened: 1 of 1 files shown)


glharper:
## Spec review

Strong, well-grounded spec. I verified the code citations against `main` and they''re all accurate: `ParseProvider` pass-through (`provisioning.go:53-57`), provider resolution unchanged (`manager.go:505-540`), `detectProviderFromFiles` gated on `NotSpecified` (`importer.go:304`), `ProjectInfrastructure` never inspecting `service.Host` (`importer.go:292-358`), the `azdext.ProvisioningProvider` interface methods (`:23-36`) matching the method table, schema enum at `azure.yaml.json:44-52`, and registration parity between `default.go:81-86` and `provisioning_service.go:138-152`. Implementation-ready.

### The three questions you flagged

**1. Is the data-plane/ARM split in "What the synthesizer ignores" drawn correctly?**
Mostly, but there''s an internal inconsistency. The table lists `agents[].runtime:` as **ignored** ("deploy verb''s job"), yet Validation step 3 requires the synthesizer to read `docker:`/`runtime:`/`image:` to enforce the exactly-one deploy-mode invariant, and `docker:`->ACR / `image:`->no-ACR are explicitly *read* for branching. So those three aren''t "ignored" - they''re **read for validation and ARM branching but emit no ARM of their own**. Suggest splitting the table into *read-for-branching-but-no-ARM* (`docker`, `image`, `runtime`) vs *not-read-at-all* (`connections`, `toolboxes`, `skills`, `routines`, `tools`, `skill`, `protocols`, `env`, `startupCommand`). As written the table contradicts the validation section.

**2. Drift detection on `Deploy()` - is "no detection" right?**
This is the weakest part of the proposal. The justification ("CLI `add` commands already warn at the entry point") only covers mutations made *through* azd commands. A user hand-editing `azure.yaml` (e.g. adding a second container agent -> needs ACR) bypasses every `add` warning, then `azd provision` reads stale on-disk Bicep, **silently** skips the ACR, and the failure surfaces much later in `azd deploy` (docker push) with a confusing error. A cheap, non-invasive guard at `Deploy()` - diff the *set of ARM-relevant fields* in `azure.yaml` against what the on-disk template declares, and **warn** (not block) - would catch this without re-introducing auto-merge. I''d push back on "no detection" and at least make it a one-line warn.

**3. Eject delete-and-rerun with no `--force`?**
Refusing to clobber user-owned files matches azd''s posture. But a **hard refuse** is slightly more friction than azd''s usual style - azd typically *prompts* on overwrite rather than erroring out. Consider an interactive confirm ("`./infra/` exists, regenerate and overwrite? [y/N]") with the hard-refuse behavior preserved under `--no-prompt`/CI. Keeps the safety while feeling more native. The all-or-nothing scope is right.

### Other findings

- **`Destroy` is under-specified and risks a real bug.** "Delete resource group or use deployment stacks" omits soft-delete purge. Foundry/Cognitive Services accounts (and any Key Vault) are soft-delete-protected; Core''s Bicep provider handles purge. Without it, `up`->`down`->`up` under the same name fails on the second provision. Should be called out so the extension''s `Destroy` matches Core parity.

- **Schema relaxation loses typo-catching.** Changing `enum: ["bicep","terraform"]` -> `examples` means `provider: biceps` is no longer flagged in IDE validation for *anyone*, not just Foundry users. Since extension provider names are open-ended an enum can''t enumerate them - but consider a `pattern` (e.g. `^[a-z0-9.]+$`) alongside `examples` to retain some validation, or note the accepted regression explicitly.

- **Brownfield + `Preview` (What-If) needs scope resolution too.** The spec says ARM-ID-from-`endpoint:` resolution is deploy-time, but What-If also needs the target scope (RG/subscription). Confirm `Preview` resolves endpoint->scope, not just `Deploy`.

- **Telemetry doc requirement.** Per repo convention (`cli/azd/AGENTS.md`), new fields (`provision.synthesis_source`, `init.infra_flag`) must be added to `docs/reference/telemetry-data.md`. Worth naming this file as a required deliverable in the implementation PRs.

- **`infra.layers[]` escape-hatch is asserted but unverified.** The coexistence story relies on per-layer provider scoping, but `importer.go:316` short-circuits on `len(Layers) > 0` and the spec doesn''t confirm an `InfraLayer` can name an extension provider. Worth a one-line verification since it''s the only mixed-project answer.

- **Minor:** Stability Contract ("semantically identical") vs Test Plan ("byte-equal output") - clarify byte-equal is *per extension version*. And the `infra.provider:` line added by `init` becomes dead once auto-routing lands; note a cleanup path.

hund030:
@glharper Thanks for the careful review — all 8 points hold up and have been applied to the spec:

1. **Synthesizer ignores table** — split into "read for branching" (`docker`/`runtime`/`image`) and "not read at all" (everything else). `runtime` was the only contradiction; `docker`/`image` weren't actually in the table.
2. **Drift detection** — Open Question 1 flipped to *warn (don't block)* on every on-disk `Deploy()`. Pseudocode and method-table row updated.
3. **Eject UX** — now matches `azd infra generate`: interactive `[y/N]` prompt, `--no-prompt` keeps the hard-refuse for CI.
5. **`Destroy` soft-delete purge** — row updated to mirror Core's `getCognitiveAccountsToPurge` + `purgeItems` (`bicep_provider.go:1283-1413`); `up → down → up` failure mode called out.
6. **Schema typo regression** — `pattern: ^[a-z0-9.]+$` + `examples` instead of `examples` alone.
7. **Brownfield + Preview scope** — both `Deploy` and `Preview` now resolve `endpoint:` to ARM ID + scope before invoking ARM.
8. **Telemetry doc requirement** — one-line callout naming `docs/reference/telemetry-data.md` per `cli/azd/AGENTS.md:246-249`.
9. **`infra.layers[]` escape hatch** — verified inline: `InfraLayer.Provider` (`provider.go:57` → `:40`) + `ParseProvider` accepts any string.
10. **Stability vs Test Plan wording** — tightened to *byte-stable within a patch extension version*, matching the test.

hund030:
Thanks for the careful review — all 8 points hold up and have been applied to the spec:

1. **Synthesizer ignores table** — split into "read for branching" (`docker`/`runtime`/`image`) and "not read at all" (everything else). `runtime` was the only contradiction; `docker`/`image` weren't actually in the table.
2. **Drift detection** — Open Question 1 flipped to *warn (don't block)* on every on-disk `Deploy()`. Pseudocode and method-table row updated.
3. **Eject UX** — now matches `azd infra generate`: interactive `[y/N]` prompt, `--no-prompt` keeps the hard-refuse for CI.
5. **`Destroy` soft-delete purge** — row updated to mirror Core's `getCognitiveAccountsToPurge` + `purgeItems` (`bicep_provider.go:1283-1413`); `up → down → up` failure mode called out.
6. **Schema typo regression** — `pattern: ^[a-z0-9.]+$` + `examples` instead of `examples` alone.
7. **Brownfield + Preview scope** — both `Deploy` and `Preview` now resolve `endpoint:` to ARM ID + scope before invoking ARM.
8. **Telemetry doc requirement** — one-line callout naming `docs/reference/telemetry-data.md` per `cli/azd/AGENTS.md:246-249`.
9. **`infra.layers[]` escape hatch** — verified inline: `InfraLayer.Provider` (`provider.go:57` → `:40`) + `ParseProvider` accepts any string.
10. **Stability vs Test Plan wording** — tightened to *byte-stable within a patch extension version*, matching the test.

Test Plan also picked up entries for each new behavior. Pushing the commit shortly.

glharper (approved):

docs/specs/bicepless-foundry/spec.md
```diff
@@ -0,0 +1,574 @@
...
+a single `host: microsoft.foundry` service per project, with nested
+`agents:`, `deployments:`, `connections:`, `toolboxes:`, etc. This spec
+only changes how that file is *provisioned*; it does not redesign the YAML.
+
+## Scope
+
+**In scope:**
+
+- Bicep-less default behavior for `azd ai agent init`
+- `azd ai agent init --infra` eject command
```
therealjohn, line 38: Note* This would depend on the other PR being merged of `init` being idempotent. Seems OK.

therealjohn (approved):

PR #8590, opened 2026-06-10
docs: technical design spec for unifying Foundry config in azure.yaml

huimiu:
## Summary

- Add technical design spec for unifying Foundry agent configuration in `azure.yaml` at `docs/specs/unify-azure-yaml/spec.md`
- Docs only, no product code changes

## Details

**Design spec** (`docs/specs/unify-azure-yaml/spec.md`)

The spec is the engineering counterpart to the product brief. It covers two areas the brief does not address:

- End to end experience: command flows for `azd ai agent init`, `provision`, `deploy`, `up`, and `azd down`; a Foundry service next to a normal service via `uses:`; connecting to an existing project via `endpoint:`; migration from `host: azure.ai.agent` and separate `agent.yaml` / `agent.manifest.yaml` files; partial deploy failure and safe re-run behavior.
- Technical design: config binding through `AdditionalProperties` and gRPC, wiring `host: microsoft.foundry` to a new service target, JSON Schema composition at the service level, `$ref` file includes and overlay overrides, `${VAR}` vs `${{...}}` templating, service target lifecycle with per-agent fan-out, `init` rework, state reconciliation, sibling-extension schema ownership, and telemetry and errors.

### DIFF
--- a/docs/specs/unify-azure-yaml/spec.md
+++ b/docs/specs/unify-azure-yaml/spec.md
@@ -5,30 +5,56 @@
 ## Overview
 
 This document is the engineering design for the unified `azure.yaml` proposal. It covers
-the **technical design** that azd core and the `azure.ai.agents` extension need, and the
+the **technical design** that azd core and the Foundry extensions need, and the
 **end to end experience** a user gets at the command line.
 
 Background and product spec can be found in the following sources:
 
-- Product brief and sample file shapes: the `simple` and `complex` branches of the
-  preview repo, and the proposed schemas under `schemas/`.
+- Product brief and sample file shapes: the `simple`, `complex`, and `separate-services`
+  branches of the preview repo, and the proposed schemas under `schemas/`.
 - RFC issue [#7962](https://github.com/Azure/azure-dev/issues/7962): Unify Foundry
   agent configuration in `azure.yaml`.
 - RFC issue [#8049](https://github.com/Azure/azure-dev/issues/8049): Composition
   commands (`azd ai project add ...`), which depends on this work.
 
-One note on shape, because it affects everything below. The earlier issue #7962
-proposed two host kinds, `azure.ai.project` for shared state and `azure.ai.agent` per
-agent, linked with `uses:`. The current brief replaces that with a single
-`host: microsoft.foundry` service that owns all Foundry state, with agents nested as an
-array inside that one entry. This design follows the single service shape. The two
-service shape is treated as superseded.
+One note on shape, because it affects everything below. The earlier issue #7962 proposed
+two host kinds, `azure.ai.project` for shared state and `azure.ai.agent` per agent, linked
+with `uses:`. A later iteration collapsed everything into a single `host: microsoft.foundry`
+service that owned all Foundry state with agents nested as an array. The current brief
+generalizes the per-resource idea: **every Foundry resource is its own `services:` entry**
+with a singular host, keyed by the resource name. A Foundry project is a project service plus
+sibling services for its connections, toolboxes, skills, agents, and routines, ordered with
+`uses:` and cross-referenced by name. This design follows that **separate-services** shape.
+The single `host: microsoft.foundry` shape and the original two-host shape are superseded.
+
+The host kinds are:
+
+| Host kind | Owns |
+|---|---|
+| `azure.ai.project` | the Foundry project plus its `deployments:` array and optional `endpoint:` |
+| `azure.ai.connection` | one project connection |
+| `azure.ai.toolbox` | one toolbox |
+| `azure.ai.skill` | one skill |
+| `azure.ai.agent` | one agent (the existing host kind, with an evolved shape) |
+| `azure.ai.routine` | one routine |
+
+Why this shape. A Foundry project is a set of related data-plane resources, not one
+monolith. Modeling each as a service lets azd core do the work it already does for services:
+order them with `uses:`, report progress and failures per resource, and dispatch each to the
+extension that owns it. Schema ownership falls out cleanly, because each sibling extension owns
+the host kind, the service target, and the schema for the one resource it already manages
+through its `azd ai` CLI. It also reuses the existing `azure.ai.agent` service-target pattern
+rather than inventing a single project-wide target that fans out internally.
+
+`microsoft.foundry` stays the meta-package extension id that bundles the sibling extensions; in
+this shape it is no longer a host kind.
 
 ### Scope
 
-In scope: the schema change, the new service target, the config binding, `$ref`
-includes, templating, the consolidated `init`, the deprecation of the old host and old
-files, and the lifecycle behavior.
+In scope: the per-host schema conditionals, the new service targets across the sibling
+extensions, the config binding, `$ref` includes, templating, the consolidated `init`, the
+retirement of `agent.yaml` / `agent.manifest.yaml` and the old config-nested `azure.ai.agent`
+shape, and the lifecycle behavior.
 
 Out of scope, tracked elsewhere:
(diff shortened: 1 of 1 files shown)


glharper:
I'm not seeing a new skill added for design-spec.

huimiu:
> I'm not seeing a new skill added for design-spec.

@glharper Sorry, wrong content removed

docs/specs/unify-azure-yaml/spec.md
```diff
@@ -0,0 +1,516 @@
...
+
+No new ordering logic is needed. `uses:` already exists on every service.
+
+### 1.4 Use an existing Foundry project
+
+When `endpoint:` is set on the Foundry service, azd connects to that project instead of
+creating one. This is the path for teams that provision infrastructure on their own, and
+for reusing a shared or private network bound account, which is the minimum issue
+[#8165](https://github.com/Azure/azure-dev/issues/8165) asks for. Provisioning a new
+network bound account from scratch is part of the separate built-in Bicep work, not this
```
therealjohn: I understand why this is this way - just making a note that we will need to ensure there is a path to create and customize network settings in the bicep work. 

What we are saying here is that in order to do that, you'd need to immediately eject to Bicep and customize it to get VNET settings, etc. 

We don't expose the `Account` resource as a service here, and the network settings are on the account not the Foundry Project.

docs/specs/unify-azure-yaml/spec.md
```diff
@@ -0,0 +1,516 @@
...
+link to the guide.
+
+### 1.6 Teardown with `azd down`
+
+The rule for Foundry data plane state is the same one issue #8049 uses: removing an
+entry from `azure.yaml` means stop using it, not destroy it. The destructive path is
+explicit and runs through `azd down` or the per resource `azd ai` commands.
+
+`azd down` removes the Foundry project that azd provisioned. If the service used an
+`endpoint:` to point at an existing project, `azd down` does not delete that project,
```
therealjohn: This is correct - and I think AZD down might already have logic for this so we may not need to build anything custom, but double-check.

docs/specs/unify-azure-yaml/spec.md
```diff
@@ -0,0 +1,516 @@
...
+file when the command explicitly targets it. Both this design and #8049 should share one
+YAML edit helper that understands `$ref` entries, so reads and writes agree.
+
+### 2.5 Templating: `${VAR}` and `${{...}}`
+
+Two resolvers have to coexist without stepping on each other.
+
+- `${VAR}` is an azd environment value, resolved on the client before anything is sent
+  to Foundry. Example: a connection secret read from the azd environment.
+- `${{...}}` is resolved by Foundry on the server, for example
```
therealjohn: This is just treated as a string value, no resolution needed. Its a special syntax for Foundry.

docs/specs/unify-azure-yaml/spec.md
```diff
@@ -0,0 +1,516 @@
...
+
+Deprecation detection lives here too. If the old files are present, or the project uses
+`host: azure.ai.agent`, print the warning, run the fallback that produces the equivalent
+in memory Foundry service, and emit the telemetry signal. The error code
+`CodeInvalidAgentManifest` in `internal/exterrors/codes.go` stays while the manifest path
+is read. Rename or retire it when that path is removed.
+
+### 2.8 State reconciliation and idempotency
+
+Deploy compares the declared state in `azure.yaml` against the live state in Foundry and
```
therealjohn: I dont expect the extensions to query state and do a compare - that seems very complex to manage. 

Instead, I expected we could just do a CreateOrUpdate type API call and the service would handle it.

docs/specs/unify-azure-yaml/spec.md
```diff
@@ -0,0 +1,516 @@
...
+meta-package: `azure.ai.toolboxes`, `azure.ai.connections`, `azure.ai.projects`,
+`azure.ai.skills`, and `azure.ai.routines`. Each owns a data-plane CLI (`azd ai toolbox`,
+`azd ai connection`, and so on) that acts on a live project. None of them participate in
+`azure.yaml` today.
+
+That raises who owns the `microsoft.foundry.json` schema slices and their reconciliation.
+Following the brief, v1 takes **Option A**: the `azure.ai.agents` extension owns the full
+schema and reconciles every slice (deployments, connections, toolboxes, skills, routines,
+agents), while the sibling extensions keep their imperative CLIs unchanged. The
+alternative, Option B, has the meta-package own the schema and each sibling register a
```
therealjohn: Now that we have the meta package, it makes sense to me it would live there but confirm that doesnt have other consequences.

therealjohn (approved):

huimiu:
Thanks for the review!

- **Network settings** — Noted that VNET config is on the Foundry Account, not the Project. azd doesn't manage the Account, so you'd need to eject to Bicep today. The Bicep work needs to cover this path.
- **`azd down`** — Added a note that the existing teardown logic likely already handles this — just needs a confirm.
- **`${{...}}`** — Clarified that azd never resolves this; it's passed through as-is to Foundry.
- **State reconciliation (§2.8)** — Switched to idempotent CreateOrUpdate, letting Foundry handle it server-side instead of diffing on the client.
- **Schema ownership (§2.10)** — Flipped to Option B: `microsoft.foundry` meta-package owns the schema, each sibling extension contributes its slice. Reconciliation stays in `azure.ai.agents`. Core will need a new slice contribution mechanism.

therealjohn (approved):

docs/specs/unify-azure-yaml/spec.md
```diff
@@ -0,0 +1,632 @@
...
+  but is new core behavior and forces core to own the merge and validation rules.
+- Each owning extension resolves `$ref` while it parses its own keys. This is contained to the
+  Foundry extensions and needs no core change.
+
+Recommend each owning extension owns resolution for the first version, with a clear path to
+move it into core later if other extensions need includes.
+
+**Resolution rules.**
+
+- A relative `$ref` path resolves relative to the file that holds it, so nested includes work.
```
jongio, line 371: The resolution rules cover path rebasing and overlays, but don't mention cycle detection. Since nested includes resolve relative to the including file, a mistaken circular reference (`A.yaml` refs `B.yaml` which refs `A.yaml`) would cause infinite recursion. Consider adding a bullet here requiring cycle detection with a reasonable max depth (e.g., 10 levels), and surfacing a clear error naming the cycle when hit.

docs/specs/unify-azure-yaml/spec.md
```diff
@@ -0,0 +1,632 @@
...
+    endpoint: https://my-account.services.ai.azure.com/api/projects/my-project
+  basic-agent:
+    host: azure.ai.agent
+    uses: [agent-project]
+    kind: hosted
+    project: src/basic-agent
+    # ...
+```
+
+The absence of `endpoint:` on the project service is the signal to provision a new project.
```
jongio, line 169: Using absence of `endpoint:` as the provision signal is clean for the common case, but it's a negative signal: accidentally removing the line from an existing project's config would trigger provisioning a new project. Worth a note in the migration section (1.5) calling out that removing `endpoint:` is a destructive change that should be caught by validation or a confirmation prompt on first provision.

jongio (commented):
I verified the code references against the current repo (`service_config.go:AdditionalProperties`, `mapper_registry.go`, `service_manager.go:alpha.IsFeatureKey`, `listen.go:WithServiceTarget`, the init.go functions, and all sibling extensions). They all resolve correctly. Two design-level observations inline.

PR #8598, opened 2026-06-10
fix: lazy-detect container engine so Podman is used in Aspire dotnet deploys

### LINKED ISSUE #8525: [Bug] "empty dotnet configuration output" error when publishing Aspire empty project with Kafka packages

vhvb1989:
## Problem

The fix in #8527 correctly threads the container engine to `dotnet publish`, but **it never actually detects Podman** in the Aspire .NET container app deploy path.

### Root Cause

`dotnetContainerAppTarget.RequiredExternalTools()` returns only `[dotNetCli]` — it does **not** include `docker.Cli`. This means `docker.Cli.CheckInstalled()` is never called via `tools.EnsureInstalled()` during the Aspire deploy flow. As a result, `docker.Cli.containerEngine` stays `""` and `ContainerEngine()` defaults to `"docker"` — so `-p:ContainerEngine=podman` is never appended.

### Fix

`ContainerEngine()` now performs **lazy detection** on first call if `CheckInstalled()` hasn't run yet:
1. Checks `AZD_CONTAINER_RUNTIME` env var
2. Checks if `docker` is in PATH
3. Falls back to `podman` if in PATH
4. Defaults to `docker` if neither found (backward compatible; `CheckInstalled()` will error later if actually needed)

This is a lightweight check — no version validation or daemon readiness — just enough to determine which engine to pass to `dotnet publish`.

### Testing

5 new test cases for the lazy detection path:
- Podman detected when docker not in PATH
- Docker detected when both in PATH  
- `AZD_CONTAINER_RUNTIME` env var respected
- Defaults to docker when neither in PATH
- Does not override `CheckInstalled()` result if already called

All existing docker, dotnet, and project tests pass.

Fixes #8525
Related: #8527, https://github.com/microsoft/aspire/issues/17593

### DIFF
--- a/cli/azd/pkg/tools/docker/docker.go
+++ b/cli/azd/pkg/tools/docker/docker.go
@@ -36,10 +36,36 @@ type Cli struct {
 }
 
 // ContainerEngine returns the detected container engine name ("docker" or "podman").
-// CheckInstalled() should be called first to detect and set the container engine.
-// If not set, defaults to "docker" for backward compatibility.
+// If CheckInstalled() has already been called, this returns the cached result.
+// Otherwise, it performs a lightweight detection (env var + PATH check) without
+// validating version or daemon readiness. This ensures callers that need the
+// engine name (e.g., dotnet publish with -p:ContainerEngine) get the correct
+// value even when docker.Cli is not in RequiredExternalTools.
 func (d *Cli) ContainerEngine() string {
-	return d.getContainerEngine()
+	if d.containerEngine == "" {
+		d.detectContainerEngine()
+	}
+	return d.containerEngine
+}
+
+// detectContainerEngine performs a lightweight detection of the container engine
+// by checking AZD_CONTAINER_RUNTIME and PATH. Unlike CheckInstalled(), it does
+// not validate versions or daemon readiness.
+func (d *Cli) detectContainerEngine() {
+	if runtime := os.Getenv("AZD_CONTAINER_RUNTIME"); runtime == "docker" || runtime == "podman" {
+		d.containerEngine = runtime
+		return
+	}
+
+	if d.commandRunner.ToolInPath("docker") == nil {
+		d.containerEngine = "docker"
+	} else if d.commandRunner.ToolInPath("podman") == nil {
+		d.containerEngine = "podman"
+	} else {
+		// Fallback — neither found; default to docker for backward compatibility.
+		// CheckInstalled() will produce a proper error if actually invoked.
+		d.containerEngine = "docker"
+	}
 }
 
 // getContainerEngine returns the container engine command to use ("docker" or "podman").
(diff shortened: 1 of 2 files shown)


hemarina (approved):
Nice, well-targeted fix — the root-cause analysis is accurate (`dotnetContainerAppTarget.RequiredExternalTools()` omits `docker.Cli`, so `CheckInstalled()` never runs and the engine silently defaults to `docker`), and lazy detection is a reasonable minimal solution. Tests pass locally. Approving. A couple of minor, non-blocking notes below.

**1. `AZD_CONTAINER_RUNTIME` is validated inconsistently (low-medium)**
`CheckInstalled()` returns a hard error on an invalid value, but `detectContainerEngine()` silently ignores an invalid value and falls through to PATH detection. Because the Aspire dotnet path is exactly the one where `CheckInstalled()` never runs, a typo (e.g. `AZD_CONTAINER_RUNTIME=podmn`) would be silently swallowed here instead of surfaced to the user. Consider aligning the two paths (e.g. share a small validation helper, or at least log when an explicitly-set value is unrecognized).

**2. Test hermeticity (minor)**
The `t.Parallel()` subtests in `Test_ContainerEngine_LazyDetection` don't clear `AZD_CONTAINER_RUNTIME`. If it happens to be exported in a dev/CI shell, `detectContainerEngine()` returns early and the PATH-based assertions would break. A `t.Setenv("AZD_CONTAINER_RUNTIME", "")` guard in those subtests would make them hermetic.

**Informational (no action needed):**
- `ContainerEngine()` now mutates shared state (`d.containerEngine`) with no lock, and `docker.Cli` is a singleton. This is *not* a race today since services are built/deployed sequentially — flagging only as a latent concern if parallel deploys are ever introduced.
- Only the public `ContainerEngine()` lazily detects; internal callers (`Login`/`Build`/`Push`) still default to `docker` on empty. Doesn't affect the Aspire path (dotnet publish does its own container build), so just noting the asymmetry.

PR #8618, opened 2026-06-11
Updating instructions by mining PRs for feedback

richardpark-msft:
This is a first pass with a skill I wrote. It goes through a range of PRs (last 2 months), and then tries to come up with instructions that would catch or give advice similar to the feedback on those PRs (from humans!).

I've left in a few things, but wanted to open up a discussion about the entries, see what feedback you have and (once we get it into an acceptable state) merge it in, which should help with Copilot Code Reviews and on our local box.


richardpark-msft:
Note for reviewers - this is not the instruction updates final form - we'll want to trim out the references links, etc.. That's just for you to look at, if you want to peek around and see what comments/feedback were used to form the instructions update.

richardpark-msft:
Testing this in https://github.com/Azure/azure-dev/pull/8622

PR #8643, opened 2026-06-15
feat(agents): bicep-less init for azd ai agent init

### LINKED ISSUE #8065: Make azd ai agent init Bicep-less by default; add --infra to eject

hund030:
## Summary

Closes #8065. `azd ai agent init` now scaffolds a Bicep-less project by default — `azure.yaml` declares `infra.provider: microsoft.foundry` and the extension owns provisioning end-to-end (`azd provision`, `--preview`, `up`, `down`, `env refresh`). The new `--infra` flag ejects the in-memory template to `./infra/main.bicep` for users who need to customize it.

## Changes

- **`internal/cmd/init.go`, `init_from_code.go`**: emit Bicep-less project shape; replace the `Azure-Samples/azd-ai-starter-basic` clone with an `azd init -t <emptyStagedDir> <targetDir>` invocation; relax the `infra.provider` JSON Schema from a closed enum to `pattern: "^[a-z0-9.]+$"`.
- **`internal/cmd/init_infra.go`**: new `--infra` eject path materializes the template to disk.
- **`internal/synthesis/`**: in-memory Bicep synthesizer with `go:embed`'d `main.bicep` + precompiled `main.arm.json`; `AcceptedHosts: ["azure.ai.agent"]`.
- **`internal/project/foundry_provisioning_provider.go`**: implements all 9 azdext provisioning methods (Initialize, State, Deploy, Destroy, Preview via ARM what-if, Parameters, Outputs, Resources, GetDeployment) for both inline and on-disk (`./infra/`) paths; `azd down --force --purge` enumerates live Cognitive Services accounts pre-delete then calls `DeletedAccountsClient.BeginPurge`.
- **`internal/project/service_target_agent.go`**: split `Initialize` into a cheap stub + lazy `ensureDeployContext` / `ensureEnv` so the service target registers without a deploy context.

## Tests

- `azd ai agent init`: Scaffold a template without bicep files
- **Replace the azure.yaml with following content**: we didn't complete unify azure.yaml, so the scaffolded azure.yaml still use old shape, which is not supported for provisioning.

  ```
  
  name: agent-framework-agent-basic-invocations
  services:
      app:
          project: src/agent-framework-agent-basic-invocations
          host: azure.ai.agent // this supposed to be microsoft.foundry after azure.yaml unified
          language: docker
          agents:
              - instructions: You are a test agent.
                kind: prompt
                name: prompt-agent
          deployments:
              - model:
                  format: OpenAI
                  name: gpt-5-codex
                  version: "2025-09-15"
                name: gpt-5-codex
                sku:
                  capacity: 50
                  name: GlobalStandard
  infra:
      provider: microsoft.foundry
  ```

- `azd provision`: provision the foundry resource and model deployment
- `azd down`: destroy the provisioned resources
- `azd ai agent init --infra`: eject the biceps
- `azd provision`: provision based on the on-disk bicep files


glharper (commented):
## Review: bicep-less init for `azd ai agent init`

**Scope reviewed:** 31 files, +5722/-544, against the PR base branch `huimiu/foundry-azure-yaml`. I built the extension and ran the new `internal/project` + `internal/synthesis` test suites (pass), then independently read the two highest-risk areas: the destructive `azd down --force --purge` path and the `--infra` eject path.

**Overall:** Well-structured and well-tested. Error handling follows the `exterrors` convention (structured errors created once via `ServiceFromAzure`/`Validation`/`Internal`, not re-wrapped), new codes are snake_case, and the destructive paths are carefully scoped. Only one minor issue found.

### 🟢 Low — Documented embedded fallback is dead; Deploy/Preview can send a `nil` ARM template

`internal/project/foundry_provisioning_provider.go:446-450` (root cause at `:124-131`)

When `Initialize` detects an on-disk template, it returns at line 131 **before** `p.armTemplate` is loaded (only set at line 176 on the synthesize path), leaving it `nil`. If the on-disk file is then deleted between the presence check and the load, `resolveTemplate`'s race-fallback (lines 426-429) falls through to:

```go
return &templateSource{ mode: templateModeEmbedded, armTemplate: p.armTemplate, ... } // nil
```

The comment promises "fall back to embedded," but the embedded template was never loaded on this path, so Deploy/Preview send a `null` template → a confusing ARM `InvalidTemplate` error instead of a graceful fallback. Reachable only via a TOCTOU race (user deletes `./infra/main.bicep` mid-provision); non-destructive.

**Suggested fix:** Load `p.armTemplate` unconditionally in `Initialize` (even on the on-disk path), or re-synthesize the embedded ARM JSON in the `src == nil` fallback and error clearly if it can't.

### ✅ Independently verified correct

- **`azd down --force --purge`** (`Destroy` / `collectAccountsToPurge` / `purgeCognitiveAccounts`, lines 605-771): scoped to `p.rgName` via `NewListByResourceGroupPager` (**not** a subscription-wide scan); pager fully iterated; live accounts captured *before* the RG delete and purged *after* (when soft-deleted); `BeginDelete`/`BeginPurge` pollers are awaited with error handling; a missing RG short-circuits to idempotent success. It cannot purge anything the RG delete wasn't already destroying, and `--force` is required (no silent destructive default).
- **`init_infra.go` eject:** refuses when `./infra/` already exists (`CodeInfraEjectExists`), cleans up partial dirs on error, and writes only compile-time `go:embed` names — destination paths are not derived from user input, so no traversal/clobber risk.
- **Lazy-init split** (`service_target_agent.go` `ensureDeployContext`/`ensureEnv`): matches azd's per-provider sequential lifecycle; all state-touching entrypoints call the ensure-guards first, so no nil-deref ordering gap.
- **Schema relaxation (enum → `^[a-z0-9.]+$`):** the provider name is only used to route to a registered provider / as a literal constant; it is never used to build filesystem paths or imports, so no path-traversal/injection vector.
- Build and new test suites pass; tests use buffers / `t.Log`, not `os.Stdout`.

**Recommendation:** Approvable. Worth addressing the Low `nil`-template fallback (small fix) for robustness, but it is not a merge blocker.

therealjohn (approved):

PR #8675, opened 2026-06-16
feat(agents): split Foundry init into per-resource azure.yaml services

### LINKED ISSUE #8676: [ext-agents]: azd ai agent init writes unified microsoft.foundry config in azure.yaml
Currently `azd ai agent init` produces separate `agent.yaml` / `agent.manifest.yaml` files. This change updates `init` to write a single unified `azure.yaml` using the `host: microsoft.foundry` format, with the agent declared inline under the service `agents:` array and model deployments at the project level. This aligns the on-disk representation with the `microsoft.foundry` service target and removes the need for the legacy agent manifest files.

huimiu:
## What this PR does

Changes how `azd ai agent init` writes Foundry resources into `azure.yaml`: each resource type becomes its own service entry instead of being bundled into the single agent service.

## Per-resource `azure.yaml` services

**Before:** all Foundry resources (model deployments, connections, toolboxes) were bundled into the single agent service entry.

**After:** each resource type gets its own `azure.yaml` service entry with a dedicated host:
- `azure.ai.project` — the Foundry project + model deployments (one per agent)
- `azure.ai.connection` — one entry per connection
- `azure.ai.toolbox` — one entry per toolbox

The agent service lists these as `uses:` dependencies so `azd` provisions them in the right order.

Example `azure.yaml` after init:
```yaml
services:
  ai-project:
    host: azure.ai.project
    config:
      deployments: [...]
  my-connection:
    host: azure.ai.connection
    uses: [ai-project]
    config:
      name: my-connection
  my-agent:
    host: azure.ai.agents
    uses: [ai-project, my-connection]
```

**Why split:** per-resource services give each Foundry resource a clear ownership boundary in `azure.yaml` and make the provisioning graph explicit via `uses:`. For now the **agents extension registers all three resource hosts** (`azure.ai.project`, `azure.ai.connection`, `azure.ai.toolbox`) as no-op service targets, so only that extension needs to be installed for `azd up`/`deploy` to walk the entries; the resources themselves are still provisioned by Bicep / the data-plane API during `provision`.

Provisioning behavior is unchanged — the agent extension re-sources deployments, connections, and toolboxes from the sibling services when setting provisioning environment variables and creating toolsets. Projects created before the split (resources bundled on the agent service) keep working: the collectors fall back to the agent service config when no sibling services exist, so an existing `azure.yaml` still provisions without re-running `init`.

## Scope / follow-up

This PR is intentionally limited to the `azure.yaml` split plus registering the resource hosts from the agents extension. Fully onboarding `azure.ai.projects` / `azure.ai.connections` / `azure.ai.toolboxes` as standalone azd extensions that own their resource's deploy-time lifecycle is a larger effort and will be iterated on separately — each host can move from the shared no-op provider in the agents extension to its own extension when it gains real behavior.

## Base branch

This PR targets `huimiu/foundry-azure-yaml`, which carries only the already-merged Foundry **azure.yaml schema** work (#8603, #8627, #8689, #8643).


glharper:
## Review: unified `microsoft.foundry` config from `init`

Switches `azd ai agent init` (manifest + from-code paths) to write a single unified `host: microsoft.foundry` service entry with the agent inline under `agents:` and model `deployments:` at project level, carried on `ServiceConfig.AdditionalProperties` instead of `config:`. Drops the separate `agent.yaml` write, extracts a shared `buildFoundryServiceConfig` helper, and adds solid tests. Build, `go vet`, `gofmt`, and the new tests pass locally; line lengths are within the `lll` limit.

The design is clean and the test coverage is genuinely good — `TestBuildFoundryServiceConfig_ValidatesAsHostedAgent` round-trips the writer''s output through the reader''s `FoundryProjectConfig.Validate()`, and `splitRuntime` is verified as the correct inverse of the service target''s `runtimeString`.

### Issues

**1. (High) `nextstep` no longer recognizes the service `init` just wrote.**
`init.go` calls `nextstep.AssembleState` → `ResolveAfterInit` at the end of `addToProject` (init.go:2679-2680), but `nextstep/state.go:438` filters services by `svc.Host != agentHost` where `agentHost = "azure.ai.agent"` (state.go:30). Since init now writes `microsoft.foundry`, the just-initialized service is skipped, so init''s own "next steps" output silently loses protocol-aware `azd ai agent run` guidance, `azd env set <KEY>` hints, and `edit agent.yaml` placeholder hints. `doctor/checks_project.go:24` matches the same stale literal. These filters should also accept `microsoft.foundry`.

**2. (Medium, by design) Container/docker agents become undeployable.**
For a hosted agent with no `codeConfiguration` and no `image` (the default remote-build container path), `buildFoundryServiceConfig` falls into the `default` branch and sets `agent.Docker` — but `validateHostedAgent` (foundry_config.go:276) hard-rejects `deployModeDocker`. So `init` writes an `azure.yaml` that fails only later at deploy/provision. It''s acknowledged in comments, but consider emitting a warning at init time so users aren''t surprised at deploy.

**3. (Low) Stale comment.** init.go:2666-2674 still says the resolver "inspects … each `azure.ai.agent` service''s `agent.yaml`" — the host is now `microsoft.foundry` and init no longer writes `agent.yaml`.

**4. (Low) `foundryAgentInput.instructions` is never populated.** A prompt-kind agent (set when `agentDef.Kind != AgentKindHosted`) would be written with empty `instructions`. Latent only because prompt agents are rejected by the target today.

Minor: the container branch hardcodes `Docker.Path: "Dockerfile"` regardless of the actual Dockerfile name (no impact while docker mode is non-deployable).

Item #1 is the one I''d want addressed (or explicitly deferred with a tracking note) before merge, since it regresses init''s own UX immediately.

therealjohn (approved):

therealjohn:
This fixes https://github.com/Azure/azure-dev/issues/7962

therealjohn (changes requested):
After more review, some changes:

- each resource needs to be it's own service entry
- each extension would own the schema for that service entry, similar to how the agents ext owns the schema for host: azure.ai.agents
- each extension would need a hook on doing dataplane operations, similar to how the agents ext has a post-deploy hook to create an agent

 
Right now, AZD does not support different extensions contributing to the same service host - so separating them makes that possible, but it also means we can more easily manipulate each resource as a service. 

Example: https://github.com/therealjohn/foundry-azd-config-preview/blob/separate-services/REFERENCE.md#project-basics

cli/azd/extensions/azure.ai.connections/internal/cmd/listen.go
```diff
@@ -0,0 +1,23 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
trangevi: This handling doesn't currently do anything (same with all of the other extensions), is this intended? Are you planning to change this in a different PR into the "foundry-azure-yaml" branch? (As an aside, I think this is a concern with requesting reviews on feature branches, if there's an intent to add more to that same feature branch later).

Back to this point though, I don't think we should have the other extensions register the service provider if they aren't going to do anything with it. A single extension can register multiple service providers, so we can have the agents extension register all of them to start, and then when we want to move _full_ functionality over to the individual extensions we can do that. Otherwise, this creates a false impression that the connections/projects/toolboxes/skills/etc. extensions have a backend service interaction that they don't.

trangevi (changes requested):

cli/azd/extensions/azure.ai.agents/internal/cmd/resource_services.go
```diff
@@ -0,0 +1,368 @@
...
+	// AiConnectionHost owns a single Foundry project connection.
+	AiConnectionHost = "azure.ai.connection"
+	// AiToolboxHost owns a single Foundry toolbox (toolset).
+	AiToolboxHost = "azure.ai.toolbox"
+
+	// aiProjectServiceName is the stable azure.yaml service key used for the
+	// single azure.ai.project service. A stable name keeps repeated inits
+	// idempotent (AddService overwrites by name) so there is one project
+	// service per project, matching the unified Foundry config design.
+	aiProjectServiceName = "ai-project"
```
trangevi, line 34: Instead of having this fixed, should it be the name of the user's Foundry project?

cli/azd/extensions/azure.ai.agents/internal/cmd/resource_services.go
```diff
@@ -0,0 +1,368 @@
...
+	}
+
+	// One project service owns the model deployments. Deployments stay an
+	// array on it (there is a single Foundry project and deployments belong
+	// to it). Create it whenever any Foundry resource is emitted -- even with
+	// no deployments (e.g. "Skip model configuration") -- so connections and
+	// toolboxes always have a stable ai-project dependency that enforces
+	// provisioning order.
+	projectServiceName := ""
+	if len(deployments) > 0 || len(connections) > 0 || len(toolboxes) > 0 {
```
trangevi: Why check these? None of these resources are dependent on the models, which is all the project service currently contains. If we still want it added even if there is no model, then we should just always add it, I'm not sure I understand the need for this conditional check?

cli/azd/extensions/azure.ai.agents/internal/cmd/resource_services.go
```diff
@@ -0,0 +1,368 @@
...
+	return nil
+}
+
+// collectProjectDeployments gathers the model deployments declared across all
+// azure.ai.project services so provisioning handlers can source them from the
+// sibling project service instead of the agent service config. Services are
+// visited in sorted name order so serialized env-var output stays stable.
+//
+// Falls back to the deployments bundled on the agent service when no project
+// service carries any, so an azure.yaml written before the per-resource split
```
trangevi, line 218: I'm wondering if we want to support back compat here. I get the idea, but this is a major change in how we're handling things, and a part of me feels like the sooner we force adoption of it the better. @therealjohn, thoughts?

trangevi (changes requested):

cli/azd/extensions/azure.ai.agents/internal/cmd/resource_services.go
therealjohn: I would like to back-compat so we dont break everyone else thats been using this - like samples, docs, portal, etc. 

Otherwise the scope of coordinating the release is much much bigger.

cli/azd/extensions/azure.ai.agents/internal/cmd/resource_services.go
huimiu: My thinking was to keep ai-project as a fixed internal key (rather than a display name), so re-running init would consistently reuse the same service instead of creating duplicates. A real project name could also potentially clash with other keys. Does that reasoning make sense to you?

cli/azd/extensions/azure.ai.agents/internal/cmd/resource_services.go
huimiu: Good catch! You're right, I've simplified it to always emit ai-project and added tests to cover it. Thanks

cli/azd/extensions/azure.ai.agents/internal/cmd/resource_services.go
trangevi: It shouldn't conflict with other keys since it's specific to the azure.ai.project scope, right? It just feels weird to me that everything else we have the name matches, but this one doesn't. Re-running init shouldn't create duplicates as long as users are using the same Foundry project. If they want to use a different Foundry project we would need to handle that, but personally it feels better to handle that than to have this discrepancy in the services. @therealjohn maybe you have an opinion here?

trangevi (approved):

cli/azd/extensions/azure.ai.agents/internal/cmd/resource_services.go
therealjohn: @huimiu can we remain consistent with our other approach where the service name matches the user defined name when its a new project, or the one they select if its existing? 

That being said - I dont think we have a feature where we allow a user to define a project name - its determined in the Bicep today and it's not customizable. 

So if we dont have the upstream feature to support that lets file and track that with this.

huimiu: Sure, let me update in next PR

PR #8679, opened 2026-06-16
fix(project): preserve existing azure.yaml properties on AddService

### LINKED ISSUE #8678: init doesn't serialize all existing properties of azure.yaml
If a template azure.yaml includes other properties like hooks, they are lost after doing `azd ai agent init <path to agent.manifest.yaml>`. 

I think the issue is that `init` is re-serializing the azure.yaml with any interactive vales (parameters, etc.) and not including all the existing properties.

glharper:
## What

`azd ai agent init` (and any extension that adds a service via the gRPC `Project().AddService` API) was dropping existing top-level `azure.yaml` properties such as `hooks`.

## Root cause

`projectService.AddService` (`cli/azd/internal/grpcserver/project_service.go`) was the **only** mutating gRPC handler that saved the **cached** in-memory `ProjectConfig` without first reloading it from disk. Every other mutating handler (`SetConfigSection`, `SetConfigValue`, `UnsetConfig`, `SetServiceConfigSection`, `SetServiceConfigValue`, `UnsetServiceConfig`) reloads from disk (via the lossless raw-map path and/or `reloadAndCacheProjectConfig`).

When the cached config was stale relative to the on-disk `azure.yaml` — as happens in the `init` flow, which materializes/updates `azure.yaml` after azd first resolves `lazyProjectConfig` — `project.Save` re-serialized the stale struct over the file and wiped everything not present in the cache: `hooks`, other top-level keys, and even pre-existing services.

I confirmed this by reproduction: a fresh load → add service → `Save` is lossless, but a stale cache → add service → `Save` clobbers `hooks`, custom top-level keys, and existing services.

## Fix

`AddService` now reloads the project config from disk before reading/mutating/saving, mirroring the established pattern in the other handlers. Runtime event-handler state is preserved through `reloadAndCacheProjectConfig` → `CopyRuntimeStateTo`, so hooks and extension-registered handlers keep working.

## Tests

Adds `Test_ProjectService_AddService_PreservesExistingProperties`, which seeds a stale cache against an on-disk `azure.yaml` containing `hooks`, a custom top-level key, and an existing service, then asserts they all survive after `AddService` alongside the newly added service. The test fails without the fix (hooks wiped) and passes with it.

## Validation

- `go build ./...`
- `go test ./internal/grpcserver/...` (full package)
- `go vet`, `gofmt`, `golangci-lint run` (0 issues), `cspell` — all clean

This is a core-only change; the `azure.ai.agents` extension needs no edit (it already calls `AddService`), and the fix benefits every extension that adds services.

Fixes #8678

### DIFF
--- a/cli/azd/internal/grpcserver/project_service.go
+++ b/cli/azd/internal/grpcserver/project_service.go
@@ -188,6 +188,14 @@ func (s *projectService) AddService(ctx context.Context, req *azdext.AddServiceR
 		return nil, err
 	}
 
+	// Reload the project config from disk before mutating so we never clobber
+	// properties (e.g. hooks, or any other top-level keys) that were written to
+	// azure.yaml after the lazy cache was first resolved. Other mutating handlers
+	// (SetConfig*, SetServiceConfig*, UnsetConfig*) reload for the same reason.
+	if err := s.reloadAndCacheProjectConfig(ctx, azdContext.ProjectPath()); err != nil {
+		return nil, err
+	}
+
 	projectConfig, err := s.lazyProjectConfig.GetValue()
 	if err != nil {
 		return nil, err
--- a/cli/azd/internal/grpcserver/project_service_test.go
+++ b/cli/azd/internal/grpcserver/project_service_test.go
@@ -5,6 +5,7 @@ package grpcserver
 
 import (
 	"context"
+	"os"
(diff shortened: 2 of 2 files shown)


JeffreyCA (approved):
Thanks for the fix!

PR #8681, opened 2026-06-16
fix: handle `context` and `repo` OIDC claim keys in `azd pipeline config`

### LINKED ISSUE #8680: fix: `azd pipeline config` does not recognize `context` OIDC claim key in custom subject templates
## Description

`azd pipeline config` fails to build OIDC subjects when the GitHub repository (or org) OIDC customization template includes the `context` claim key.

This was introduced in #7705 which added support for custom OIDC subject claims. The `BuildOIDCSubject()` function handles `repository_owner_id`, `repository_id`, `repository_owner`, and `repository` — but **not `context`**, which is a [valid GitHub OIDC claim key](https://docs.github.com/en/actions/reference/security/oidc#customizing-the-subject-claims-for-an-organization-or-repository).

cc @pamelafox who reported this issue when running `azd pipeline config` against a repo in an org that uses `context` in its OIDC template.

## What is the `context` claim key?

Per [GitHub docs](https://docs.github.com/en/actions/reference/security/oidc#customizing-the-subject-claims-for-an-organization-or-repository), `context` represents **the dynamic trailing part of the subject** — the portion that follows the repository identifier in the default `sub` format. Its value depends on what triggered the workflow:

| Trigger | `context` value |
|---------|----------------|
| Branch push | `ref:refs/heads/main` |
| Pull request | `pull_request` |
| Environment job | `environment:production` |

For example, `["repo", "context"]` in `include_claim_keys` is GitHub's way of representing the **default** subject format: `repo:org/repo:ref:refs/heads/main`.

## Reproduction

Given a repo whose OIDC customization API returns:
```json
{
  "use_default": false,
  "include_claim_keys": ["repository_owner_id", "repository_id", "context"]
}
```

Running `azd pipeline config` produces:
```
WARNING: Unable to build OIDC subjects from detected config: failed to build OIDC subject
for pull requests: unsupported OIDC claim key "context" in subject template for
<org>/<repo> — azd may need to be updated. Falling back to default format for manual entry.
```

In interactive mode, azd falls back to prompting with the **default format** subjects (e.g., `repo:org/repo:ref:refs/heads/main`), which are wrong for this org. The credential gets created with the wrong subject, and CI later fails with:

```
AADSTS700213: No matching federated identity record found for presented assertion subject
repository_owner_id:12345:repository_id:67890:ref:refs/heads/main.
```

The actual token GitHub emits has subject `repository_owner_id:12345:repository_id:67890:ref:refs/heads/main`, but the credential was created with `repo:org/repo:ref:refs/heads/main`.

## Root Cause

In `cli/azd/pkg/tools/github/oidc.go`, `BuildOIDCSubject()` has a switch over known claim keys. The `context` key is missing:

```go
for _, key := range oidcConfig.IncludeClaimKeys {
    switch key {
    case "repository_owner_id":
        // ...
    case "repository_id":
        // ...
    case "repository_owner":
        // ...
    case "repository":
        // ...
    default:
        return "", fmt.Errorf("unsupported OIDC claim key %q ...", key, repoSlug)
    }
}
parts = append(parts, suffix)  // suffix is always appended at the end
```

## Proposed Fix

Add a `"context"` case to the switch that appends the `suffix` parameter (which already contains the correct context value like `ref:refs/heads/main` or `pull_request`), and skip the unconditional suffix append at the end when `context` is present in the claim keys:

```go
case "context":
    parts = append(parts, suffix)
```

Then guard the final `parts = append(parts, suffix)` to only run when `context` was **not** in the claim keys (to avoid duplicating the suffix).

## Affected Organizations

Any GitHub org that includes `context` in their OIDC `include_claim_keys` template. This is common because `["repo", "context"]` is GitHub's canonical way to represent the default format, and orgs that add ID-based keys (like `repository_owner_id`, `repository_id`) alongside `context` will hit this bug.

vhvb1989:
## Description

Fixes #8680

`BuildOIDCSubject()` now recognizes two additional GitHub OIDC claim keys that are commonly used in organization-level OIDC subject templates:

- **`context`**: represents the dynamic trailing part of the subject (e.g., `ref:refs/heads/main`, `pull_request`, or `environment:prod`). Per [GitHub docs](https://docs.github.com/en/actions/reference/security/oidc#customizing-the-subject-claims-for-an-organization-or-repository), this is "the part that follows the repository in the default `sub` format". When present in `include_claim_keys`, the suffix is consumed by this claim and not appended separately.
- **`repo`**: the short-form repository identifier (`repo:owner/name`), used in GitHub's default subject format. `["repo", "context"]` is GitHub's canonical way to represent the default format.

## Problem

Organizations like `Azure-Samples` use OIDC subject templates that include the `context` claim key alongside ID-based keys:

```json
{
  "use_default": false,
  "include_claim_keys": ["repository_owner_id", "repository_id", "context"]
}
```

Since `context` was not recognized, `azd pipeline config` would warn:

```
WARNING: Unable to build OIDC subjects from detected config: unsupported OIDC claim key "context"
```

Then fall back to the wrong default format (`repo:org/repo:ref:refs/heads/main`), creating federated credentials with mismatched subjects. CI would later fail with `AADSTS700213: No matching federated identity record found`.

## Changes

1. **`cli/azd/pkg/tools/github/oidc.go`**: Added `context` and `repo` cases to `BuildOIDCSubject()`. When `context` is in the claim keys, it consumes the suffix parameter directly and the unconditional suffix append is skipped.
2. **`cli/azd/pkg/tools/github/oidc_test.go`**: Added 5 new test cases covering `context` with branch/PR/environment suffixes, `repo` + `context` combination (default format equivalent), and `repo` alone.

## Testing

All existing and new tests pass:
```
=== RUN   TestBuildOIDCSubject (12/12 PASS)
```

### DIFF
--- a/cli/azd/pkg/pipeline/github_provider.go
+++ b/cli/azd/pkg/pipeline/github_provider.go
@@ -413,14 +413,25 @@ func (p *GitHubCiProvider) credentialOptions(
 			},
 		}
 
+		// Track seen subjects to avoid duplicate FICs. When the OIDC
+		// template omits "context", all suffixes collapse to the same
+		// subject and Azure rejects duplicates on issuer+subject.
+		seenSubjects := map[string]bool{subjects.pullRequest: true}
+
 		for _, branch := range branches {
+			subject := subjects.branches[branch]
+			if seenSubjects[subject] {
+				continue
+			}
+			seenSubjects[subject] = true
+
 			safeBranchName := credentialNameSanitizer.ReplaceAllString(
 				branch, "-",
 			)
 			branchCredentials := &graphsdk.FederatedIdentityCredential{
 				Name:    fmt.Sprintf("%s-%s", credentialSafeName, safeBranchName),
 				Issuer:  federatedIdentityIssuer,
-				Subject: subjects.branches[branch],
+				Subject: subject,
 				Description: new(
 					"Created by Azure Developer CLI",
 				),
--- a/cli/azd/pkg/pipeline/github_provider_test.go
+++ b/cli/azd/pkg/pipeline/github_provider_test.go
@@ -227,18 +227,15 @@ func Test_credentialOptions_withOIDCCustomSubject(t *testing.T) {
 		require.NoError(t, err)
 		require.True(t, opts.EnableFederatedCredentials)
 
+		// Without "context" in claim keys, PR and branch subjects
+		// collapse to the same value — deduplication yields 1 FIC.
+		require.Len(t, opts.FederatedCredentialOptions, 1)
+
 		prCred := opts.FederatedCredentialOptions[0]
 		require.Equal(t,
 			"repository_owner_id:1844662:"+
-				"repository_id:599293758:pull_request",
+				"repository_id:599293758",
 			prCred.Subject,
-		)
-
-		mainCred := opts.FederatedCredentialOptions[1]
-		require.Equal(t,
-			"repository_owner_id:1844662:"+
-				"repository_id:599293758:ref:refs/heads/main",
-			mainCred.Subject,
 		)
 	})
 
@@ -301,11 +298,13 @@ func Test_credentialOptions_withOIDCCustomSubject(t *testing.T) {
 			nil,
 		)
 		require.NoError(t, err)
-		// PR + develop + main = 3 credentials
-		require.Len(t, opts.FederatedCredentialOptions, 3)
+		// Without "context", all subjects collapse to the same value.
+		// Deduplication yields only 1 FIC.
+		require.Len(t, opts.FederatedCredentialOptions, 1)
 
-		for _, cred := range opts.FederatedCredentialOptions {
-			require.Contains(t, cred.Subject, "repository_owner_id:1844662")
-		}
+		require.Contains(t,
+			opts.FederatedCredentialOptions[0].Subject,
+			"repository_owner_id:1844662",
+		)
 	})
 }
--- a/cli/azd/pkg/tools/github/oidc.go
+++ b/cli/azd/pkg/tools/github/oidc.go
@@ -7,6 +7,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"slices"
 	"strings"
 )
 
@@ -111,9 +112,11 @@ func (cli *Cli) GetRepoInfo(
 //
 //	repo:{owner}/{repo}:{suffix}
 //
-// For custom configs, claim keys are mapped to values and joined with the
-// suffix. Unknown claim keys cause an error so the user knows azd needs
-// updating.
+// For custom configs, claim keys are mapped to values and joined.
+// The special "context" claim key represents the dynamic trailing part of
+// the subject (e.g., "ref:refs/heads/main" or "pull_request"), and "repo"
+// is the short-form repository identifier ("repo:owner/name").
+// Unknown claim keys cause an error so the user knows azd needs updating.
 func BuildOIDCSubject(
 	repoSlug string,
 	repoInfo *RepoInfo,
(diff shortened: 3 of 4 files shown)


JeffreyCA (commented):
**`BuildOIDCSubject` may produce a `sub` that doesn't match GitHub-issued OIDC tokens when the template omits `context`**

For a custom `include_claim_keys` template that doesn't include `context`, the generated FIC subject won't match the token GitHub actually issues, so federated auth fails.

Per GitHub's [OIDC docs](https://docs.github.com/en/actions/reference/security/oidc#customizing-the-subject-claims-for-an-organization-or-repository), a custom template is exactly the listed claims joined in order — the dynamic ref/PR segment only appears when `context` is listed. For example: `["repo"]` → `repo:ORG/REPO`, `["repository_owner"]` → `repository_owner:monalisa`.

The current code appends `suffix` whenever `context` is absent, so a `context`-less template produces `repo:owner/name:ref:refs/heads/main` while the real token `sub` is just `repo:owner/name`. The mismatch is silent at `pipeline config` time and only surfaces as an Entra token-exchange failure (AADSTS70021) on the first CI run. The new `"repo claim key without context"` test reflects this — it asserts `repo:Azure-Samples/my-repo:ref:refs/heads/main`, which I'd expect to be `repo:Azure-Samples/my-repo`.

Suggested fix: only append `suffix` when `context` is explicitly in the template, and drop the trailing `!slices.Contains(..., "context")` append (the inline `context` case already adds it). The default (`UseDefault`) path stays as-is since its format always includes context. Then update the test expectation to `repo:Azure-Samples/my-repo`.

One follow-on to consider: with the suffix gone, `context`-less subjects no longer vary by branch, so the per-branch and PR FICs collapse to the same subject. Since Azure enforces uniqueness on issuer+subject per identity, `buildAllSubjects` would likely need to dedupe to one FIC per unique subject — which seems like the intended behavior for `["repo"]` anyway. Could be a separate follow-up if we'd rather keep this PR focused.

----------------
*Reviewed with assistance from Copilot*

vhvb1989:
Thanks @JeffreyCA — great catch. Pushed a fix:

- Removed the fallback suffix append in `BuildOIDCSubject` when `context` is absent from the claim keys. The `case "context":` branch already handles it, so context-less templates now correctly produce subjects without the dynamic ref/PR segment (matching GitHub's actual token `sub`).
- Added FIC deduplication in `github_provider.go` — when context-less templates cause branch/PR subjects to collapse to the same value, we now skip duplicates since Azure rejects duplicate issuer+subject on the same app registration.
- Updated test expectations accordingly.

hemarina (approved):
**TL;DR:** Clean fix for a real-world issue affecting orgs with custom OIDC templates. The `BuildOIDCSubject` changes are correct, well-documented, and thoroughly tested. Two minor suggestions below.

---

**1. Silent dedup skip could confuse users**
`cli/azd/pkg/pipeline/github_provider.go:419-421`

When the OIDC template omits `context`, all branch subjects collapse to the same string and the dedup silently skips them. A user configuring two branches might be surprised to see only one FIC created. Consider adding a debug/info log:

`go
if seenSubjects[subject] {
    log.Printf("Skipping duplicate FIC for branch %s (subject matches existing credential)", branch)
    continue
}
`

**2. Dedup logic in `credentialOptions()` lacks test coverage**
`cli/azd/pkg/pipeline/github_provider.go:413-437`

The `seenSubjects` dedup is an important correctness guard — without it Azure rejects duplicate FICs. However, only `BuildOIDCSubject` has new tests. A test verifying that when subjects collapse (e.g., config without `context` and two branches), `credentialOptions` emits only unique FICs would prevent regressions.

hemarina (approved):

PR #8629, opened 2026-06-12
feat(agents): add microsoft.foundry service target for a single hosted agent

huimiu:
## What

Wires the new `host: microsoft.foundry` service target in the `azure.ai.agents` extension — the foundation of design spec #8590 (§2.1, §2.2, §2.6). A single `microsoft.foundry` service represents a whole Foundry project, with agents declared inline as top-level service properties.

This foundation makes a **single hosted agent** deployable end-to-end. It also lets `azd deploy` connect to an **existing Foundry project via `endpoint:` without `azd provision`** (§1.4).

## Changes

- **Wiring** — register `microsoft.foundry` as a service target and declare the provider in the extension manifest.
- **Config binding** — parse and validate the inline agent declaration (single hosted agent, exactly one deploy mode, project requirement) and convert it to the internal agent representation.
- **Service target provider** — full deploy lifecycle for one hosted agent:
  - Prebuilt image → create agent version directly.
  - Code deploy (python/dotnet) → zip the source and create/update the version.
  - On `Initialize`, resolve the Foundry project from `endpoint:` when no project was provisioned, so `azd deploy` works against an existing project without `azd provision`.
- **Shared helpers** — extract auth setup and source zip packaging from the existing `azure.ai.agent` target so both hosts reuse them without duplicating code.
- **Tests** — config binding, validation, deploy mode mapping, env variable resolution, and endpoint parsing.

## Notes

- Code-deploy entry point is derived from `startupCommand` (the agent schema models entry via `startupCommand`, not an explicit field); documented as the seam if the schema later adds one.
- The new host does not trigger the legacy `azure.ai.agent` lifecycle hooks in `listen.go` (they are gated on the old host), so the foundation correctly does not inherit their data-plane reconcile.

## Testing

Extension module (`azureaiagent`): `go build ./...`, `go vet ./...`, `gofmt`, `golangci-lint run` (0 issues), full `go test ./...`, and `cspell` — all clean.

Fixes #8628

PR #8690, opened 2026-06-17
docs(spec): private networking for host: microsoft.foundry

### LINKED ISSUE #8165: Allow `azd ai agent` to use a private-VNet Foundry account (provision new or reuse an existing account)
We need a way to support private vnets, which requires settings on the account - its not something that can be changed later. 

https://github.com/Azure-Samples/azd-ai-starter-basic/blob/main/infra/core/ai/ai-project.bicep#L108-L140

At a minimum, we need to at least support existing accounts so devs could use one that is already setup for a vnet.

m5i-work:
Resolves #8165

## Summary

Doc-only PR. Adds `docs/specs/foundry-private-network/spec.md`, a technical design spec for declarative private networking on the `host: microsoft.foundry` service in `azure.yaml`. No product code changes — the spec is the implementation contract; code PRs follow.

The [unified `azure.yaml` design](https://github.com/Azure/azure-dev/pull/8590) (§1.4) explicitly defers VNet binding (a Foundry **Account** setting) to the built-in Bicep work. This spec is that path: a flat `network:` block the existing [bicep-less synthesizer](https://github.com/Azure/azure-dev/pull/8643) reads to provision a network-bound account.

## What changed in this update

Stakeholders approved the final `azure.yaml` surface. This PR now reflects that shape:

- `network:` present means **private data plane always** (`publicNetworkAccess: Disabled` + account private endpoint). Omit `network:` for public.
- `peSubnet` is required whenever `network:` is declared.
- `agentSubnet` presence derives egress:
  - present ⇒ BYO/customer subnet (`useMicrosoftManagedNetwork: false`)
  - absent ⇒ Microsoft-managed egress (`useMicrosoftManagedNetwork: true`)
- `isolationMode` is valid only when `agentSubnet` is absent.
- `prefix` presence controls subnet create vs reference:
  - present ⇒ create subnet
  - absent ⇒ reference existing subnet
- `dns` is optional; omitted means create/link zones, while `dns.resourceGroup` references central/shared zones.

## What the spec covers

- **Flat `network:` surface** — `agentSubnet`, `peSubnet`, `dns`, and managed-egress `isolationMode`; no `mode: byo | managed` enum.
- **Primitives from the service team's templates** — what azd keeps, drops (managed stores, caphost wiring, tool subnet, UAMI, CMK), and defers (ACR).
- **Additive synthesizer / `main.bicep` changes** (high level), including regenerating the precompiled `main.arm.json` fallback.
- **Validation** — `peSubnet` required, subnet VNet/name required, CIDR validation, same-VNet v1 constraint, `isolationMode` only for managed egress, brownfield precedence (`endpoint:` wins).
- **Update behavior** — re-provision is incremental, but topology changes are limited by service mutability; tightening managed isolation is supported, relaxing/switching egress is not a v1 contract.
- **ACR private-networking RoI decision** — BYO image (`--image`) in v1; defer local-build-into-private-ACR.

## Relationship to in-flight work

Lands as a follow-on into the `huimiu/foundry-azure-yaml` feature branch after the synthesizer (#8643) merges. Rides `ServiceConfig.AdditionalProperties`, so no core change beyond the schema slice. The BYO-image flow and the merged remote-build skip for VNet-injected accounts keep ACR off the critical path. See §3 for the in-flight work overview.

## Out of scope

Public opt-out under `network:`, cross-VNet v1 topologies, BYO stores + capability-host wiring, tool subnet, UAMI, CMK, and local-build-into-private-ACR are explicitly excluded from v1.

### DIFF
--- a/docs/specs/foundry-private-network/spec.md
+++ b/docs/specs/foundry-private-network/spec.md
@@ -14,13 +14,13 @@ VNet binds at the Account level, yet `network:` sits on a service entry that rea
 
 ## Solution
 
-Extend the Foundry provisioning synthesizer (`internal/synthesis`, introduced by the bicep-less work) to read a new `network:` block on the service body and emit the network primitives a secured agent account requires: a VNet with an agent subnet and a private-endpoint subnet, account-level `networkInjections`, an account private endpoint, and the AI private DNS zones — each created or referenced per the developer's declaration.
+Extend the Foundry provisioning synthesizer (`internal/synthesis`, introduced by the bicep-less work) to read a new `network:` block on the service body and emit the network primitives a secured agent account requires: optional agent runtime VNet injection, an account private endpoint, and the AI private DNS zones — each created or referenced per the developer's declaration.
 
-The block is purely additive. When `network:` is absent the synthesizer behaves exactly as today (public account). When present, the account flips to private and the network modules are included. No new provider, no new gRPC surface, no core change beyond the schema slice — the block rides the same `AdditionalProperties` channel every other Foundry key already uses.
+The block is purely additive. When `network:` is absent the synthesizer behaves exactly as today (public account). When present, the account data plane is **always private**: the account has `publicNetworkAccess: 'Disabled'`, its network ACL default action is `Deny`, and an account private endpoint is provisioned in `peSubnet`. No new provider, no new gRPC surface, no core change beyond the schema slice — the block rides the same `AdditionalProperties` channel every other Foundry key already uses.
 
 Dependent stores (Cosmos DB, AI Search, Storage) stay **platform-managed**, which is supported under VNet isolation. This removes the bulk of the sample template (BYO stores, their private endpoints, role assignments, and both capability hosts) from azd's responsibility.
 
-The example below illustrates one scenario (BYO VNet with explicit subnets); it is not the full schema. See the `azure.yaml` surface section for the field reference and the create-vs-reference rules.
+The example below illustrates one scenario (BYO egress with explicit subnets); it is not the full schema. See the `azure.yaml` surface section for the field reference and the create-vs-reference rules.
 
 ```yaml
 infra:
@@ -32,17 +32,16 @@ services:
 
     # New: private networking for the Foundry account.
     # Omitted => public account (unchanged behavior).
+    # Present => private data plane (always).
     network:
-      mode: byo                       # byo | managed
-      byo:
-        vnet:
-          id: ${AZURE_VNET_ID}        # required for byo (v1)
-        agentSubnet:
-          name: agent-subnet
-          prefix: 192.168.0.0/24
-        peSubnet:
-          name: pe-subnet
-          prefix: 192.168.1.0/24
+      agentSubnet:
+        vnet: ${AZURE_VNET_ID}
+        name: agent-subnet
+        prefix: 192.168.10.0/24
+      peSubnet:
+        vnet: ${AZURE_VNET_ID}
+        name: pe-subnet
+        prefix: 192.168.11.0/24
       dns:
         resourceGroup: rg-private-dns
         subscription: ${AZURE_DNS_SUBSCRIPTION_ID}
@@ -62,14 +61,18 @@ services:
 
 **In scope**
 
-- **BYO VNet** — bring an existing VNet; create or reference its subnets.
-- **Foundry-managed VNet** — with a selectable isolation mode.
-- **A network-bound (private) Foundry account** — its private endpoint and the AI private DNS zones, created or referenced.
+- **Private data plane** — declaring `network:` always disables account public access and provisions an account private endpoint.
+- **BYO egress** — inject the hosted-agent runtime into an existing customer VNet by declaring `agentSubnet`.
+- **Microsoft-managed egress** — omit `agentSubnet` and optionally choose a managed-network `isolationMode`.
+- **Create or reference subnets** — `prefix` present creates a subnet; `prefix` absent references an existing subnet.
+- **Private DNS** — create/link the AI private DNS zones by default, or reference central/shared zones with `dns:`.
 - **Platform-managed dependent stores** under VNet isolation.
 - **BYO container image** as the registry story for secured agents.
 
 **Out of scope**
 
+- Public opt-out when `network:` is declared — omit `network:` for a public account.
+- Cross-VNet v1 topologies — `agentSubnet` and `peSubnet` must target the same VNet when both are declared.
 - BYO dependent stores and the capability-host wiring they require — managed stores are used instead.
 - Tool subnet, user-assigned managed identity (UAMI), customer-managed keys (CMK) — not core to the secured-agent scenario.
 - Local build into a private ACR — secured agents bring a pre-built image; the developer owns ACR networking. See the ACR private networking section.
@@ -83,13 +86,13 @@ Private networking is the last layer on a stack already in flight on the `huimiu
 unified azure.yaml (huimiu)            #8590 docs → branch huimiu/foundry-azure-yaml
 ├─ bicep-less provisioning (hund030)   #8577 docs / #8643 impl  ← integration point
 ├─ foundry service target + init       #8629 / #8675
-└─ BYO image                           #8689  (+ #8645 remote-build skip, merged)
+└─ BYO image                           #8645 remote-build skip, merged
 ```
 
 Two consequences matter here:
 
 - **No core change.** Unknown service keys ride `ServiceConfig.AdditionalProperties` to the extension (unify §2.1), and the synthesizer reads the raw `azure.yaml` bytes directly, so the only required core edit is the JSON schema slice (see the `azure.yaml` surface section).
-- **ACR is off the critical path.** `--image` (#8689) and the merged #8645 (skip remote build for VNET-injected accounts) keep registry work out of v1 (see the ACR private networking section).
+- **ACR is off the critical path.** The BYO-image flow and the merged remote-build skip for VNet-injected accounts keep registry work out of v1 (see the ACR private networking section).
 
 ## `azure.yaml` surface
 
@@ -100,47 +103,31 @@ services:
   my-project:
     host: microsoft.foundry
 
-    # New: private networking configuration for standard/networked agents.
-    #
-    # If omitted, the project uses public networking.
-    # If present, azd provisions or configures private networking and uses
-    # customer-owned dependent stores for agent data.
-    network:
-      # Required when network is present.
-      # byo     = customer-managed / BYO VNet
-      # managed = Foundry-managed VNet
-      mode: byo
+    network:                          # present => private data plane (always). Omit for public.
 
-      # Used only when mode: byo.
-      byo:
-        # Existing customer VNet. v1 requires this id.
-        vnet:
-          id: ${AZURE_VNET_ID}
+      # ── EGRESS: agent runtime network ──────────────────────────────
+      # agentSubnet PRESENT -> inject agent into YOUR subnet  (useMicrosoftManagedNetwork: false)
+      # agentSubnet ABSENT  -> Microsoft-managed VNet         (useMicrosoftManagedNetwork: true)
+      agentSubnet:
+        vnet: ${AZURE_VNET_ID}        # required — full VNet ARM id
+        name: agent-subnet            # required — subnet leaf name
+        prefix: 192.168.10.0/24       # optional — PRESENT=create, ABSENT=reference existing
 
-        # Agent subnet.
-        # omitted => azd creates default agent subnet
-        # name only => azd references existing subnet and validates it
-        # name + prefix => azd creates subnet with that name/prefix
-        agentSubnet:
-          name: agent-subnet
-          prefix: <CIDR>
+      # Managed-egress outbound posture. VALID ONLY when agentSubnet is ABSENT.
+      # -> properties.managedNetwork.isolationMode
+      isolationMode: AllowOnlyApprovedOutbound      # | AllowInternetOutbound
 
-        # Private endpoint subnet.
-        # Same rules with Agent subnet.
-        peSubnet:
-          name: pe-subnet
-          prefix: <CIDR>
+      # ── INGRESS: account data plane (REQUIRED) ─────────────────────
+      # Always provisions an account private endpoint -> publicNetworkAccess: Disabled.
+      peSubnet:
+        vnet: ${AZURE_VNET_ID}        # required
+        name: pe-subnet               # required
+        prefix: 192.168.11.0/24       # optional — PRESENT=create, ABSENT=reference existing
 
-      # Used only when mode: managed.
-      # Values: AllowInternetOutbound | AllowOnlyApprovedOutbound
-      managed:
-        isolationMode: AllowOnlyApprovedOutbound
-
-      # Optional.
-      # If omitted, or resourceGroup is omitted, azd creates required private DNS zones.
-      # If resourceGroup is set, azd references existing private DNS zones in that resource group.
+      # Private DNS zones for the account PE. Optional; defaults to create + link
+      # to the PE's VNet. Override to reference central/shared zones. Requires peSubnet.
       dns:
-        resourceGroup: rg-private-dns
+        resourceGroup: my-dns-rg
         subscription: ${AZURE_DNS_SUBSCRIPTION_ID}
 ```
 
@@ -148,23 +135,33 @@ services:
 
 | Field | Rule |
 | --- | --- |
-| `mode` | Required when `network:` is present. `byo` and `managed` are mutually exclusive; the matching sub-block is required and the other must be absent. |
-| `byo.vnet.id` | Required in v1. The VNet must already exist. `${VAR}` is resolved from the azd environment before synthesis. |
-| `byo.agentSubnet` / `byo.peSubnet` | **Tri-state.** Omitted → azd creates a default subnet (`agent-subnet` / `pe-subnet`) with a default prefix. Name only → azd **references** an existing subnet and validates it (delegation for the agent subnet, PE network policies for the PE subnet). Name **and** prefix → azd **creates** the subnet with that name and prefix. |
-| `managed.isolationMode` | One of `AllowInternetOutbound`, `AllowOnlyApprovedOutbound`. Maps to the account managed-network isolation setting. |
-| `dns.resourceGroup` | When omitted (or the whole `dns:` block omitted), azd **creates** the required private DNS zones and links them. When set, azd **references** existing zones in that resource group. |
+| `network` | Optional. Omitted means the public-account behavior is unchanged. Present means the data plane is private in every egress mode. |
+| `agentSubnet` | Optional. Present selects BYO egress and sets the account agent `networkInjections` entry to `useMicrosoftManagedNetwork: false`. Absent selects Microsoft-managed egress (`useMicrosoftManagedNetwork: true`). |
+| `agentSubnet.vnet` | Required when `agentSubnet` is present. Full ARM id of the customer VNet. `${VAR}` is resolved from the azd environment before synthesis. |
+| `agentSubnet.name` | Required when `agentSubnet` is present. Subnet leaf name. |
+| `agentSubnet.prefix` | Optional. Present means azd creates the subnet with that prefix. Absent means azd references an existing subnet with `name`. |
+| `isolationMode` | Optional. Valid only when `agentSubnet` is absent (managed egress). One of `AllowInternetOutbound`, `AllowOnlyApprovedOutbound`. Maps to `properties.managedNetwork.isolationMode` on the account managed network. |
+| `peSubnet` | Required when `network:` is present. It is the account private endpoint subnet, and its presence is what makes the account data plane private. |
+| `peSubnet.vnet` | Required. Full ARM id of the customer VNet. `${VAR}` is resolved from the azd environment before synthesis. |
+| `peSubnet.name` | Required. Subnet leaf name. |
+| `peSubnet.prefix` | Optional. Present means azd creates the subnet with that prefix. Absent means azd references an existing subnet with `name`. |
+| `dns.resourceGroup` | Optional. When omitted (or the whole `dns:` block is omitted), azd creates the required private DNS zones and links them to the PE VNet. When set, azd references existing zones in that resource group. |
 | `dns.subscription` | Optional. Defaults to the deployment subscription. Accepts a bare GUID or `${VAR}`. Only meaningful alongside `dns.resourceGroup`. |
 
-`${VAR}` resolves client-side from the azd environment for `byo.vnet.id` and `dns.subscription`. `${{...}}` Foundry expressions are not expected in network fields and are passed through verbatim if present, consistent with the shared expander.
+`${VAR}` resolves client-side from the azd environment for VNet ids and `dns.subscription`. `${{...}}` Foundry expressions are not expected in network fields and are passed through verbatim if present, consistent with the shared expander.
 
 ## Synthesizer and template changes (high level)
 
 The bicep-less work landed `internal/synthesis/synthesizer.go` plus an embedded `templates/` tree (`main.bicep`, `modules/acr.bicep`, `abbreviations.json`, and a precompiled `main.arm.json` fallback). The account today is hardcoded to public. The changes are additive and local to this package.
 
-- **Read `network:`** — extend the `foundryService` view the synthesizer decodes so it captures the `network` sub-tree (mode, byo, managed, dns), and resolve `${VAR}` in `byo.vnet.id` / `dns.subscription` before emitting parameters. Absent block → all network parameters default off and output is byte-identical to today.
-- **Emit parameters** — add the network parameter set to the synthesizer `Result.Parameters` (mode, vnet id, subnet tri-state descriptors, isolation mode, DNS zone map + subscription).
-- **`main.bicep`** — guard the network path on a single `enableNetworkIsolation` condition. When on: include `modules/network.bicep`, set the account's `publicNetworkAccess: 'Disabled'`, `networkAcls.defaultAction: 'Deny'`, and `networkInjections` (scenario `agent`) from the agent subnet, and include `modules/private-endpoint-dns.bicep`. When off: the account block is exactly today's.
-- **New modules** — `modules/network.bicep` (VNet + two subnets, create or reference, agent-subnet delegation) and `modules/private-endpoint-dns.bicep` (account private endpoint + the three AI DNS zones `privatelink.services.ai.azure.com`, `privatelink.openai.azure.com`, `privatelink.cognitiveservices.azure.com` + VNet links, create or reference).
+- **Read `network:`** — extend the `foundryService` view the synthesizer decodes so it captures the flat `network` sub-tree (`agentSubnet`, `isolationMode`, `peSubnet`, `dns`) and resolves `${VAR}` in VNet ids / `dns.subscription` before emitting parameters. Absent block → all network parameters default off and output is byte-identical to today.
+- **Derive egress** — do not expose a `mode` enum. `agentSubnet == nil` selects Microsoft-managed egress; `agentSubnet != nil` selects BYO egress.
+- **Emit parameters** — add the network parameter set to the synthesizer `Result.Parameters`: `enableNetworkIsolation`, `useManagedEgress`, the subnet descriptors, managed isolation mode, and DNS zone RG/subscription.
+- **`main.bicep`** — guard the network path on a single `enableNetworkIsolation` condition. When on: include `modules/network.bicep`, set the account's `publicNetworkAccess: 'Disabled'`, `networkAcls.defaultAction: 'Deny'`, and the account agent `networkInjections`, and include `modules/private-endpoint-dns.bicep`. When off: the account block is exactly today's.
+- **BYO egress** — when `agentSubnet` is present, provision/reference the agent subnet and write an account `networkInjections` entry pointing at that subnet with `useMicrosoftManagedNetwork: false`.
+- **Managed egress** — when `agentSubnet` is absent, write an account agent `networkInjections` entry with `useMicrosoftManagedNetwork: true`; when `isolationMode` is set, provision the `managedNetworks/default` child resource with that value.
+- **Ingress** — when `network:` is present, `peSubnet` is required and the private endpoint is always provisioned in it. There is no managed-mode public data-plane exception.
+- **New modules** — `modules/network.bicep` (reference the VNet + create/reference subnets, agent-subnet delegation when creating) and `modules/private-endpoint-dns.bicep` (account private endpoint + the three AI DNS zones `privatelink.services.ai.azure.com`, `privatelink.openai.azure.com`, `privatelink.cognitiveservices.azure.com` + VNet links, create or reference).
 - **Regenerate `main.arm.json`** — the precompiled ARM fallback must be rebuilt from the extended `main.bicep` and committed alongside it. The byte-stability contract from the bicep-less spec applies: same `azure.yaml` → byte-identical output within a patch version.
 
 The provisioning provider, on-disk/eject behavior, and parameter wiring need no structural change — they consume whatever `Result.Parameters` carries.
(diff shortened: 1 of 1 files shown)


docs/specs/foundry-private-network/spec.md
```diff
@@ -0,0 +1,316 @@
...
+# Private networking for `host: microsoft.foundry`
+
+## Problem
+
+Foundry network isolation is unreachable from `azure.yaml` today. The
+[unified `azure.yaml` design](../unify-azure-yaml/spec.md) (§1.4) is explicit:
+VNet binding lives on the Foundry **Account** resource, the unified shape only
+models the **Project**, and so "customizing them today means ejecting to Bicep.
+That built-in Bicep work must provide a path to create and customize those
+network settings."
```
m5i-work: shorten this sentence. beside, since the spec adds VNet to a project resource (instead of the natural account level), do we need any justification? e.g. how does vnet work if multiple projects are present?

docs/specs/foundry-private-network/spec.md
```diff
@@ -0,0 +1,316 @@
...
+| `managed.isolationMode` | One of `AllowInternetOutbound`, `AllowOnlyApprovedOutbound`. Maps to the account managed-network isolation setting. |
+| `dns.resourceGroup` | When omitted (or the whole `dns:` block omitted), azd **creates** the required private DNS zones and links them. When set, azd **references** existing zones in that resource group. |
+| `dns.subscription` | Optional. Defaults to the deployment subscription. Accepts a bare GUID or `${VAR}`. Only meaningful alongside `dns.resourceGroup`. |
+
+`${VAR}` resolves client-side from the azd environment for `byo.vnet.id` and
+`dns.subscription`. `${{...}}` Foundry expressions are not expected in network
+fields and are passed through verbatim if present, consistent with the shared
+expander.
+
+## §5 Primitives captured from sample template 15
```
m5i-work: remove this section, though, we can leave an item in References section.

docs/specs/foundry-private-network/spec.md
```diff
@@ -0,0 +1,316 @@
...
+- The bicep-less synthesizer/provider itself and the unified `azure.yaml`
+  shape — prerequisites tracked by their own specs (see §3).
+
+## §3 Relationship to in-flight work
+
+Private networking is the last layer on top of a stack already in flight on the
+`huimiu/foundry-azure-yaml` feature branch. It does not start from `main`; it
+lands as a follow-on PR into that branch after the synthesizer merges.
+
+| PR | Author | Target | State | Why it matters here |
```
m5i-work: too much details here, just need an overview of ongoing efforts. For related PRs, consider a tree-like output showing the hierarchy and only capture major ones.

docs/specs/foundry-private-network/spec.md
m5i-work: Done in 3e1847b0: shortened the §1.4 quote and added a 'Why the block sits on the service' subsection — the synthesizer provisions one Account+Project per service (1:1), so service-level = account-level here; multiple services = multiple accounts each with their own network:; brownfield ignores it. Also flagged the account-scoped promotion as an open question if N projects ever share one account.

docs/specs/foundry-private-network/spec.md
m5i-work: Done in 3e1847b0: removed the §5 primitives table entirely and added a References section linking template 15 (plus the sibling specs). Kept the load-bearing account-flip (publicNetworkAccess/networkAcls/networkInjections) and the three DNS zone names in §5 (synthesizer) so the implementation contract didn't regress.

docs/specs/foundry-private-network/spec.md
m5i-work: Done in 3e1847b0: replaced the per-PR table with a short overview plus a tree of the major efforts only.

docs/specs/foundry-private-network/spec.md
```diff
@@ -0,0 +1,310 @@
...
+and the network modules are included. No new provider, no new gRPC surface, no
+core change beyond the schema slice — the block rides the same
+`AdditionalProperties` channel every other Foundry key already uses.
+
+Dependent stores (Cosmos DB, AI Search, Storage) stay **platform-managed**,
+which is supported under VNet isolation. This removes the bulk of the sample
+template (BYO stores, their private endpoints, role assignments, and both
+capability hosts) from azd's responsibility.
+
+```yaml
```
m5i-work, line 25: clarify that this just demonstrates a specific scenario, not the full schema

docs/specs/foundry-private-network/spec.md
```diff
@@ -0,0 +1,310 @@
...
+    agents:
+      - name: my-agent
+        kind: hosted
+        project: src/my-agent
+        image: myprivacr.azurecr.io/agents/my-agent:v1   # BYO image (no ACR)
+```
+
+## Scope
+
+**In scope**
```
m5i-work, line 62: Grasp high level features only, avoid detailed yaml/json fields.

docs/specs/foundry-private-network/spec.md
m5i-work: Done in 808a68c3: added a note above the example clarifying it's one illustrative scenario (BYO VNet), not the full schema, pointing to §4 for the field reference.

docs/specs/foundry-private-network/spec.md
m5i-work: Done in 808a68c3: rewrote the Scope bullets to high-level features and dropped the field-level detail (networkInjections scenario, DNS zone names, --image).

docs/specs/foundry-private-network/spec.md
```diff
@@ -0,0 +1,312 @@
+<!-- cspell:ignore privatelink azurecr vnet subnet subnets myprivacr cognitiveservices UAMI hund -->
```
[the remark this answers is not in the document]
m5i-work: Done in 8a0d2f3d: added `hund030` and `huimiu` to the cspell:ignore directive. (cspell-misc was already green, but this makes it explicit.)

docs/specs/foundry-private-network/spec.md
```diff
@@ -0,0 +1,312 @@
...
+
+# Private networking for `host: microsoft.foundry`
+
+## Problem
+
+Foundry network isolation is unreachable from `azure.yaml` today. Per the
+[unified `azure.yaml` design](../unify-azure-yaml/spec.md) (§1.4), VNet binding
```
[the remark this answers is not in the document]
m5i-work: Done in 8a0d2f3d: the unify-azure-yaml spec isn't in the repo yet (it lives in its own docs PR), so the relative link would break on merge — repointed to the source PR #8590.

docs/specs/foundry-private-network/spec.md
```diff
@@ -0,0 +1,312 @@
...
+is an **Account** setting while the unified shape models the **Project**, so
+customizing it requires ejecting to Bicep. A developer who needs a
+network-secured agent must hand-author the service team's standard
+network-secured template and maintain it by hand. There is no `network:`
+surface, and the in-memory synthesizer that
+[bicep-less provisioning](../bicepless-foundry/spec.md) introduced always emits
```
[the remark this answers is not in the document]
m5i-work: Done in 8a0d2f3d: same fix — repointed the bicep-less provisioning link to its source PR #8577 since the spec isn't in-tree yet.

docs/specs/foundry-private-network/spec.md
```diff
@@ -0,0 +1,312 @@
...
+   network-bound account, decide whether `network:` promotes from the service
+   entry to an account-scoped surface (see "Why the block sits on the service").
+
+## References
+
+- Service team's standard network-secured agent template (source of truth for
+  the ARM shape, primitives, and the keep/drop decisions in this spec):
+  [`15-private-network-standard-agent-setup`](https://github.com/microsoft-foundry/foundry-samples/tree/main/infrastructure/infrastructure-setup-bicep/15-private-network-standard-agent-setup).
+- [Unified `azure.yaml` design](../unify-azure-yaml/spec.md) — establishes the
+  `host: microsoft.foundry` shape and defers VNet to this work (§1.4).
```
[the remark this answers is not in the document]
m5i-work: Done in 8a0d2f3d: References entry now points to PR #8590.

docs/specs/foundry-private-network/spec.md
```diff
@@ -0,0 +1,312 @@
...
+- [Bicep-less provisioning](../bicepless-foundry/spec.md) — the synthesizer and
+  provider this spec extends.
```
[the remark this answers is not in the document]
m5i-work: Done in 8a0d2f3d: References entry now points to PR #8577.

therealjohn (approved):

jongio (commented):
Spec is clean. The egress derivation from `agentSubnet` presence (instead of a mode enum) is a good call; fewer states, less validation surface.

One gap worth noting in the **Update behavior** section: the table documents unsupported topology mutations (BYO ↔ managed switch, relaxing isolation mode), but the validation pipeline section only covers structural/static checks against the current `azure.yaml`. The failure path for these mutations isn't specified:

- Does azd detect the mutation by comparing config against deployed state and fail pre-synthesis?
- Or does it pass through to ARM and let the Foundry RP reject it?

If the intent is "let ARM handle it in v1," a one-liner in the Update section noting that azd doesn't validate topology transitions against live state (so errors come from the RP, not azd) would save implementers a design question and set user expectations.

Otherwise this is ready to go as an implementation contract.

m5i-work:
# VNet spec update needed after #8779

#8779 is merged, so the private networking spec should use the split resource-service shape as the baseline. The VNet feature is not shipped yet, so we should not preserve `network:` on `azure.ai.agent` or `microsoft.foundry` service hosts.

Please update this spec PR (no force-push required from me) as follows:

## Contract

`network:` belongs only on the `host: azure.ai.project` service:

```yaml
infra:
  provider: microsoft.foundry

services:
  my-agent:
    host: azure.ai.agent
    uses:
      - ai-project
    image: myprivacr.azurecr.io/agents/my-agent:v1

  ai-project:
    host: azure.ai.project
    deployments:
      - name: gpt-4.1-mini
        model: { format: OpenAI, name: gpt-4.1-mini, version: "2025-04-14" }
        sku: { name: GlobalStandard, capacity: 10 }
    network:
      agentSubnet:
        vnet: ${AZURE_VNET_ID}
        name: agent-subnet
        prefix: 192.168.10.0/24
      peSubnet:
        vnet: ${AZURE_VNET_ID}
        name: pe-subnet
        prefix: 192.168.11.0/24
      dns:
        resourceGroup: rg-private-dns
        subscription: ${AZURE_DNS_SUBSCRIPTION_ID}
```

`infra.provider: microsoft.foundry` remains the provisioning provider dispatch key. `host: microsoft.foundry` should not be the documented VNet service shape.

## Specific spec edits

- Retitle from `Private networking for host: microsoft.foundry` to `Private networking for host: azure.ai.project`.
- Rewrite “Why the block sits on the service” to say `azure.ai.project` is the project/account provisioning boundary. `azure.ai.agent` owns agent deploy config and depends on the project via `uses:`.
- Change all YAML examples from monolithic `microsoft.foundry` to split `azure.ai.agent` + `azure.ai.project`.
- In Scope/Out of scope, explicitly say `network:` on `azure.ai.agent` and `microsoft.foundry` is not part of the shipped VNet contract.
- In the `azure.yaml` surface section, say `network:` is a sibling of `endpoint:` and `deployments:` on `host: azure.ai.project`, not a sibling of `agents:` on the old monolithic service.
- In synthesizer/provider sections, replace “no structural provider/eject change” with: provisioning and eject must discover the single `azure.ai.project` service and synthesize `endpoint:`, `deployments:`, and `network:` from that service.
- In brownfield, say `endpoint:` on the `azure.ai.project` service wins and `network:` is ignored with a warning.
- In telemetry, say `provision.network_mode` is derived from `services.<project>.network` where `host: azure.ai.project`.
- In docs/env-var wording, keep `AZURE_VNET_ID` and `AZURE_DNS_SUBSCRIPTION_ID` as example user-chosen `${VAR}` names, not fixed env vars read by azd.

I opened the code/docs PR for this reparenting here: https://github.com/Azure/azure-dev/pull/8809

docs/specs/foundry-private-network/spec.md
```diff
@@ -0,0 +1,230 @@
...
+
+| Edit | Expected behavior |
+| --- | --- |
+| Add `network:` to a public greenfield account | Supported when the service accepts the resulting account update and private endpoint creation. |
+| Change DNS from create to reference (or the reverse) | Supported as an infrastructure update if the referenced zones and VNet links already exist. |
+| Tighten managed `isolationMode` | Supported in the service direction of more restriction (for example `AllowInternetOutbound` → `AllowOnlyApprovedOutbound`). |
+| Relax managed `isolationMode` or disable managed network after enabling it | Not supported by the service. Create a new environment/account instead. |
+| Switch BYO egress ↔ managed egress after agents/capability host exist | Not a v1 contract. Create a new environment/account instead. |
+
+The docs should present network topology changes as provision-time configuration, with update support limited to safe/idempotent re-provisioning and service-supported tightening.
```
jongio, line 195: This says "update support limited to safe/idempotent re-provisioning and service-supported tightening" but doesn't say who enforces the "not supported" transitions in rows 4-5 of the table. If azd doesn't validate topology transitions against live state (letting ARM/the RP fail), say so explicitly here. If azd should detect and reject pre-synthesis, that's a design requirement the synthesizer implementation needs to cover.

docs/specs/foundry-private-network/spec.md
```diff
@@ -0,0 +1,230 @@
...
+
+## Brownfield interaction
+
+`endpoint:` on the service already short-circuits synthesis — the synthesizer returns `ErrEndpointBrownfield` and the provider connects to the existing project without provisioning. A network-secured account reached this way is **already** network-bound by whoever created it.
+
+Therefore, when `endpoint:` is present, `network:` is **ignored** (the account's network posture is fixed and not azd's to change). This is documented as explicit precedence: `endpoint:` wins, and a project that wants azd to manage its network posture must be greenfield (no `endpoint:`). If both are present, azd warns that `network:` has no effect in brownfield mode.
```
jongio, line 201: When the brownfield account has `publicNetworkAccess: Disabled` and the developer's machine isn't on the VNet, `azd` will fail with a connectivity error that won't mention private networking as the cause. Worth noting whether v1 should detect this condition (e.g., check the account's `publicNetworkAccess` property after resolving `endpoint:`) and surface a diagnostic hint, or whether that's deferred.

jongio (commented):
The spec documents `network:` on `host: microsoft.foundry`, but #8779 merged the split resource-service shape. All examples, the title, and the "Why the block sits on the service" rationale need updating for `host: azure.ai.project` before this can merge. Your reparenting plan in the comment covers the right set of edits.

Two items for the reparented version (see inline comments for details):

1. The update behavior section lists unsupported topology mutations but doesn't specify who enforces them (azd pre-synthesis check vs. ARM/RP rejection). Clarifying this prevents an implementer design question.

2. The brownfield section could note the diagnostic gap when `endpoint:` points to a network-secured account: connectivity errors won't mention private networking as the cause.

jongio (changes requested):
The spec still documents `network:` on `host: microsoft.foundry`, but #8779 merged the split resource-service shape (`azure.ai.agent` + `azure.ai.project`). The reparenting plan in the conversation comment covers the right set of edits. Until those land, this spec documents a schema that won't match the implementation target.

Blocking on:

1. Reparent all examples, the title, and the "Why the block sits on the service" rationale to `host: azure.ai.project` per the plan.
2. Clarify who enforces unsupported topology mutations in the Update behavior table (azd pre-synthesis vs. ARM/RP rejection).
3. Note the diagnostic gap in Brownfield: when `endpoint:` points at a network-secured account, connectivity errors won't mention private networking as the cause.

jongio (commented):
Re-checked against current main at this HEAD (2e83728). My earlier blocking points still stand, and #8779 has since merged, so the mismatch is now concrete rather than anticipated.

1. Host shape is stale. The spec documents `network:` on `host: microsoft.foundry`, but #8779 split that into `azure.ai.agent` + `azure.ai.project`. The title, every `azure.yaml` example, and the "Why the block sits on the service" rationale still say `host: microsoft.foundry`, so this documents a schema that won't match the implementation target. Reparent to `host: azure.ai.project` (the account-level service) before merge.

2. Update behavior doesn't name the enforcer. Rows 4-5 of the table mark BYO/managed egress switches and isolation relaxation as unsupported, but the section still doesn't say who rejects them: azd pre-synthesis (comparing config against live state) or ARM/the Foundry RP. Pick one. If it's "ARM handles it in v1", say azd doesn't validate topology transitions against live state, so the error surfaces from the RP, not azd. That saves the synthesizer implementer a design question.

3. Brownfield diagnostic gap is still unnoted. When `endpoint:` points at a network-secured account and the dev machine isn't on the VNet, azd fails with a generic connectivity error that never mentions private networking as the cause. Worth a sentence on whether v1 detects `publicNetworkAccess: Disabled` after resolving `endpoint:` and surfaces a hint, or defers it.

None of these are new since 2026-07-02; the same HEAD carries them forward. Once the reparenting lands and 2-3 get a line each, this reads as a solid implementation contract.

docs/specs/foundry-private-network/spec.md
```diff
@@ -0,0 +1,230 @@
...
+
+The example below illustrates one scenario (BYO egress with explicit subnets); it is not the full schema. See the `azure.yaml` surface section for the field reference and the create-vs-reference rules.
+
+```yaml
+infra:
+  provider: microsoft.foundry
+
+services:
+  my-project:
+    host: microsoft.foundry
```
jongio, line 31: `host: microsoft.foundry` is stale here, and in the title, the intro, "Why the block sits on the service", and the second example at L104. #8779 and #9133 moved the account-level inputs (`endpoint:`, `deployments:`, `network:`) onto `host: azure.ai.project`, and the in-tree doc at `cli/azd/extensions/azure.ai.agents/docs/private-networking.md` explicitly forbids `network:` on `host: azure.ai.agent`. If this spec stays, reparent every `host: microsoft.foundry` occurrence to `host: azure.ai.project`.

jongio (commented):
Re-checked at HEAD 2e83728. The spec is unchanged since my last pass on the 18th, so my earlier points still stand. One thing has moved since then that changes the picture.

The canonical private-networking doc now lives in-tree at `cli/azd/extensions/azure.ai.agents/docs/private-networking.md`, last updated by #9133 on the 20th. It already covers this feature end to end (field reference, environment variables, limitations, three scenarios, eject flow) using the shipped shape: `network:` on `host: azure.ai.project`, with an explicit "do not place `network:` on `host: azure.ai.agent`". It also states the brownfield behavior I asked about earlier: when `endpoint:` is set, azd warns and does not reconcile `network:`.

That makes this spec both stale and largely redundant:

1. The title, the intro, "Why the block sits on the service", and both `azure.yaml` examples still describe `host: microsoft.foundry`, a shape main no longer uses (#8779 split it into `azure.ai.agent` plus `azure.ai.project`).
2. The Update behavior and Brownfield sections now overlap the in-tree doc, which already answers the brownfield question I raised.

Pick one path before this merges:

1. Retire it. If the in-tree doc is the source of truth, close this PR (or shrink the spec to a pointer at that doc) instead of landing a second, conflicting description under `docs/specs/`.
2. Reconcile it. If there is design rationale here the in-tree doc does not capture, reparent everything to `host: azure.ai.project`, align the examples with the in-tree doc, and drop the sections that now duplicate it, keeping only the net-new design content.

Merging as-is lands a spec that contradicts shipped docs on the one detail that matters most: where the `network:` block lives.

jongio (commented):
Re-verified at HEAD `2e83728`, unchanged since my last pass, so the open inline points still stand. Confirming against current `main`:

**Host shape is stale.** The spec still puts `network:` on `host: microsoft.foundry` (title L6, intro L28, examples L105 and L257). `main` no longer uses that as a service host: #8779 split it into `azure.ai.agent` and `azure.ai.project`, and the shipped in-tree doc `cli/azd/extensions/azure.ai.agents/docs/private-networking.md` puts `network:` on `host: azure.ai.project` and explicitly forbids it on `host: azure.ai.agent`. `microsoft.foundry` is now the infra provider, not the service host.

**Redundant with the shipped doc.** That in-tree doc already covers this feature end to end (field reference, egress modes, brownfield `endpoint:` behavior, eject flow) with the correct shape. As written, this spec adds a second description that conflicts on the detail that matters most: where the `network:` block lives.

Two ways forward, matching the update you already flagged after #8779:
1. Retire it: close this PR, or shrink the spec to a pointer at the in-tree doc.
2. Reconcile it: reparent every `microsoft.foundry` service-host occurrence to `azure.ai.project`, align the examples with the in-tree doc, and drop the now-duplicated sections, keeping only net-new design rationale.

jongio (commented):
Re-confirmed at HEAD `2e83728`, unchanged since my last pass, so I'm posting this as a status check rather than new findings. The one blocking issue is still open, so I can't sign off yet.

`network:` is still documented on `host: microsoft.foundry`, but `main` no longer uses that as a service host. #8779 split it into `azure.ai.agent` and `azure.ai.project`, and the shipped in-tree doc `cli/azd/extensions/azure.ai.agents/docs/private-networking.md` (#9133) puts `network:` on `host: azure.ai.project` and explicitly forbids it on `host: azure.ai.agent`. `microsoft.foundry` is the infra provider now, not the service host. Still stale:

- Title (L9) and intro (L15)
- Example service hosts at L37 and L110

`provider: microsoft.foundry` at L33 is correct as the provisioning dispatch key, so only the `host:` occurrences need reparenting.

Two paths, same as the update you flagged after #8779:
1. Retire it: close this PR, or shrink the spec to a pointer at the in-tree doc, which already covers the field reference, egress modes, brownfield `endpoint:` behavior, and eject flow end to end.
2. Reconcile it: reparent every `host: microsoft.foundry` service occurrence to `host: azure.ai.project`, align the examples with the in-tree doc, and drop the sections that now duplicate it, keeping only net-new design rationale.

Merging as-is lands a spec that contradicts the shipped docs on the detail that matters most: where the `network:` block lives.

jongio (commented):
Re-confirmed at HEAD `2e83728`. No changes since my last pass, so this restates the one open blocking issue rather than adding new findings.

The spec parents `network:` on `host: microsoft.foundry`, but that is no longer a service host on `main`. #8779 split it into `host: azure.ai.agent` and `host: azure.ai.project`. The shipped in-tree doc `cli/azd/extensions/azure.ai.agents/docs/private-networking.md` (#9133) puts `network:` on `host: azure.ai.project` and explicitly forbids it on `host: azure.ai.agent`. `microsoft.foundry` is now the infra provider only. `provider: microsoft.foundry` at L33 is correct; the stale `host: microsoft.foundry` occurrences are the title (L9), intro (L15), and examples (L37, L110).

Two ways forward:

1. Reconcile: reparent every `host: microsoft.foundry` service occurrence to `host: azure.ai.project`, align the examples with the in-tree doc, and drop the sections that now duplicate it, keeping only net-new design rationale.
2. Retire: close this PR, or shrink the spec to a pointer at the in-tree doc, which already covers the field reference, egress modes, brownfield `endpoint:` behavior, and eject flow.

As written, this lands a spec that contradicts the shipped docs on the detail that matters most: where the `network:` block lives.

docs/specs/foundry-private-network/spec.md
```diff
@@ -0,0 +1,230 @@
+<!-- cspell:ignore privatelink azurecr vnet subnet subnets myprivacr cognitiveservices UAMI hund hund030 huimiu -->
+
+# Private networking for `host: microsoft.foundry`
```
jongio, line 3: This title sets the shape for the whole spec, and it's the shape `main` doesn't use.

`cli/azd/extensions/azure.ai.agents/schemas/microsoft.foundry.json` on `main` has no `network` property. Its full property set is `endpoint`, `deployments`, `connections`, `toolboxes`, `skills`, `routines`, `agents`. The block lives in `cli/azd/extensions/azure.ai.projects/schemas/azure.ai.project.json` instead, and `schemas/v1.0/azure.yaml.json` labels this host "Legacy Microsoft Foundry host - compatibility for old non-network files; new provisioning uses azure.ai.project".

Your own #8809 did the move: -59 lines from the foundry schema, +63 to the project schema. Same stale occurrences at L9, L31, L104, and L229. `provider: microsoft.foundry` at L27 stays as is.

docs/specs/foundry-private-network/spec.md
```diff
@@ -0,0 +1,230 @@
...
+- Public opt-out when `network:` is declared — omit `network:` for a public account.
+- Cross-VNet v1 topologies — `agentSubnet` and `peSubnet` must target the same VNet when both are declared.
+- BYO dependent stores and the capability-host wiring they require — managed stores are used instead.
+- Tool subnet, user-assigned managed identity (UAMI), customer-managed keys (CMK) — not core to the secured-agent scenario.
+- Local build into a private ACR — secured agents bring a pre-built image; the developer owns ACR networking. See the ACR private networking section.
+- The bicep-less synthesizer/provider itself and the unified `azure.yaml` shape — prerequisites tracked by their own specs (see the Relationship to in-flight work section).
+
+## Relationship to in-flight work
+
+Private networking is the last layer on a stack already in flight on the `huimiu/foundry-azure-yaml` feature branch. It does not start from `main`; it lands as a follow-on into that branch after the synthesizer merges.
```
jongio, line 83: This says the work "does not start from `main`" and lands into `huimiu/foundry-azure-yaml` after the synthesizer merges. The base branch on this PR is `main`, and #9133 landed the project provisioning there on Jul 20. #8643, #8629, and #8675 all merged into the feature branch back in June, and #8809 reparented the network block on top of them.

If the spec still lands, this section has to describe shipped state instead of a plan.

docs/specs/foundry-private-network/spec.md
```diff
@@ -0,0 +1,230 @@
...
+**Docs**
+
+- New env vars consumed for network fields (`AZURE_VNET_ID`, `AZURE_DNS_SUBSCRIPTION_ID`, or whatever the synthesizer reads) are documented in `cli/azd/docs/environment-variables.md` with format and default.
+- Extension README documents the `network:` block and the BYO-image requirement for secured agents.
+
+## References
+
+- Service team's standard network-secured agent template (source of truth for the ARM shape, primitives, and the keep/drop decisions in this spec): [`15-private-network-standard-agent-setup`](https://github.com/microsoft-foundry/foundry-samples/tree/main/infrastructure/infrastructure-setup-bicep/15-private-network-standard-agent-setup).
+- Service team's managed-network sample for private data plane + Microsoft-managed egress: [`18-managed-virtual-network`](https://github.com/microsoft-foundry/foundry-samples/tree/main/infrastructure/infrastructure-setup-bicep/18-managed-virtual-network).
+- [Unified `azure.yaml` design](https://github.com/Azure/azure-dev/pull/8590) — establishes the `host: microsoft.foundry` shape and defers VNet to this work (§1.4).
```
jongio, line 229: Both anchors here are closed without merging: #8590 (unified `azure.yaml` design) and #8577 (bicep-less provisioning). A reader following them lands on abandoned PRs rather than the design that actually shipped.

This line also restates the stale contract, since #8809 moved the `network:` shape off `host: microsoft.foundry`. The live equivalent is `cli/azd/extensions/azure.ai.agents/docs/private-networking.md`.

jongio (commented):
Re-checked at HEAD `2e83728`. Same code as my last pass, but I went into the shipped schema this time and the picture is clear enough now that I'd retire this rather than reparent it.

The reparenting you laid out on Jun 25 already shipped. #8809 ("move Foundry networking to project service") merged that same day and moved the block itself: `cli/azd/extensions/azure.ai.agents/schemas/microsoft.foundry.json` lost 59 lines, `cli/azd/extensions/azure.ai.projects/schemas/azure.ai.project.json` gained 63. #9133 then landed the project provisioning on `main` on Jul 20.

State of `main` today:

- `microsoft.foundry.json` has no `network` property at all. Its full property set is `endpoint`, `deployments`, `connections`, `toolboxes`, `skills`, `routines`, `agents`.
- `azure.ai.project.json` carries the exact block this spec describes, down to `required: [peSubnet]` and the `not: required[agentSubnet, isolationMode]` mutual exclusion.
- `schemas/v1.0/azure.yaml.json` tags the host: "Legacy Microsoft Foundry host - compatibility for old non-network files; new provisioning uses azure.ai.project".
- `cli/azd/extensions/azure.ai.agents/docs/private-networking.md` states `network:` is supported only on `host: azure.ai.project`.

HIGH: the spec parents `network:` on `host: microsoft.foundry` at L3, L9, L31, L104, and L229. That's the one host whose shipped schema explicitly carries no `network` key. `provider: microsoft.foundry` at L27 is still correct.

MEDIUM: "Relationship to in-flight work" (L81-95) frames this as landing into `huimiu/foundry-azure-yaml` after the synthesizer merges, and says it "does not start from `main`". This PR bases on `main`, and that work merged there a month ago.

MEDIUM: the two References anchoring the design, #8590 and #8577, are both closed without merging.

Dropping my two earlier asks. The update-behavior enforcer question and the brownfield diagnostic gap are moot now that the feature shipped, and the in-tree doc already answers brownfield ("azd warns and does not reconcile `network:`").

The in-tree doc covers the field reference, egress modes, limitations, three scenarios, and eject end to end at the shipped shape. I'd close this PR rather than spend another pass reparenting a design that's already documented in-tree. If there's rationale here that doc doesn't capture, the ACR RoI table and the `provision.network_mode` telemetry field look like the only candidates, and folding just those into the in-tree doc keeps one source of truth.

docs/specs/foundry-private-network/spec.md
```diff
@@ -0,0 +1,230 @@
...
+| **BYO image (`--image`)** | Nothing. The developer brings `registry/image:tag` and owns the registry's SKU, private endpoint, DNS, and firewall. | ~0 | **Chosen** |
+| **Local build → private ACR** | Provision a Premium ACR with a private endpoint and `privatelink.azurecr.io`, then solve build connectivity: a remote ACR-task build must egress to a private registry, or a local push must reach it from outside the VNet (developer IP allowlist). | High; many connectivity edge cases | **Deferred** |
+
+The hard part of the local-build path is not creating the registry — it is making the build actually reach a network-isolated registry, which drags in build-agent network placement azd does not control. v1 therefore standardizes on BYO image for secured agents. Revisit local-build-into-private-ACR after telemetry shows `--image` adoption and real demand.
+
+## Telemetry and docs
+
+**Telemetry**
+
+- `provision.network_mode` — `none` | `byo` | `managed`, emitted at provision start. The value is derived from `network:` presence and `agentSubnet` presence. Lets us measure secured-agent adoption and BYO-vs-managed split. Implementation PRs add it to `docs/reference/telemetry-data.md`.
```
jongio, line 218: This shipped already, so the future tense is wrong. `docs/reference/telemetry-data.md` L402 documents `provision.network_mode` with the same `none` / `byo` / `managed` values, and `cli/azd/extensions/azure.ai.projects/internal/provisioning/foundry_provisioning_provider.go` L763-770 emits it via `attribute.String("provision.network_mode", networkMode)`.

I flagged this field on Jul 27 as one of two things in this spec worth folding into the in-tree doc. That was wrong, it was already there. Correcting it so the record is straight: the ACR deferral rationale is the only piece left that the tree doesn't have.

jongio (commented):
Re-checked at `2e83728`. No new commits since my last pass, so my three open comments still stand. Two things I got wrong or missed last time, and both push harder toward closing this.

First, the reparenting isn't an open design question. You wrote the exact edit list yourself on Jun 25 in this PR's conversation: retitle to `host: azure.ai.project`, rewrite "Why the block sits on the service", convert the YAML examples to the split `azure.ai.agent` plus `azure.ai.project` shape, state in Scope that `network:` on `azure.ai.agent` and `microsoft.foundry` isn't part of the shipped contract, and six more. It closes with "`host: microsoft.foundry` should not be the documented VNet service shape." HEAD here is `2e83728`, committed Jun 23, two days before that comment. None of those nine edits landed. This has sat for a month with an agreed fix list from its own author unapplied.

Second, I was wrong on Jul 27 when I called the telemetry field one of two things worth salvaging. `provision.network_mode` already shipped verbatim. `docs/reference/telemetry-data.md` L402 documents it with the same `none` / `byo` / `managed` values this spec proposes, and `cli/azd/extensions/azure.ai.projects/internal/provisioning/foundry_provisioning_provider.go` L763-770 emits it.

That leaves the ACR RoI table as the only content here not already in the tree, and even there #8645 shipped the decision. What's missing in-tree is just the written rationale for deferring local build into a private ACR.

So my recommendation is unchanged and now narrower: close this, and fold that one ACR deferral rationale into `cli/azd/extensions/azure.ai.agents/docs/private-networking.md` so there's one source of truth. If you'd rather keep it as a design record, your Jun 25 list is the work item, plus a rebase onto current `main` so "Relationship to in-flight work" describes shipped state instead of a plan.

PR #8704, opened 2026-06-18
Add `azd tool install/upgrade azure-skills` with per-host selection

### LINKED ISSUE #8596: Support installing skill plugins via azd tool
Today we'll install the azure-skills plugin before using any agentic features, but we should also bring that to `azd tool` as well.  I think we should offer to install it for at least `copilot` and potentially `claude` as well given we have packages for both https://github.com/microsoft/GitHub-Copilot-for-Azure#installation.  This should politely fail if there are no agentic coding assistants installed.  It also obviously needs to happen after installing `copilot`.

We should work with the Skills team on how to correctly version and detect updates skill plugins.

Let's start with `azure-skills` but we could open this up to other skill plugins that enhance the developer experience in the future.

hemarina:
# Add `azd tool install/upgrade azure-skills` with per-host selection

Fixes #8596.

Adds `azure-skills` as a built-in `ToolCategorySkill`, so users can install,
upgrade, and inspect the **Microsoft Azure Skills plugin**
([`microsoft/azure-skills`](https://github.com/microsoft/azure-skills)) through
whichever supported agentic CLI they already have on `PATH`:

```
azd tool install azure-skills
azd tool upgrade azure-skills
azd tool list
```

## Supported hosts

The skill installs through an agentic CLI host's native plugin commands. Two
hosts are supported today (apm/codex/gemini were dropped for now due to low
traffic and may return as usage grows):

| Host | Install | Update | List |
| --- | --- | --- | --- |
| GitHub Copilot CLI (`copilot`) | `plugin marketplace add microsoft/azure-skills` → `plugin install azure@azure-skills` | `plugin update azure@azure-skills` | `plugin list` |
| Claude Code (`claude`) | `plugin marketplace add https://github.com/microsoft/azure-skills` → `plugin install azure` | `plugin update azure@azure-skills` | `plugin list azure@azure-skills` |

## Host selection (`--host`, on both install and upgrade)

- `--host copilot` / `--host claude` — target specific host(s). Comma-separated
  or repeatable: `--host copilot,claude` or `--host copilot --host claude`. Each
  named host must be a configured host that is on `PATH`, otherwise a
  descriptive error names the supported hosts.
- `--host all` — target every detected host. Cannot be combined with specific
  names.

**Install, without `--host`:**

- Explicitly named (`azd tool install azure-skills`) with a single host on
  `PATH` → that preferred host.
- Explicitly named with **multiple** hosts detected → stops and asks the user to
  choose (`--host <host>` or `--host all`) rather than silently picking one.
- Pulled in by a batch (`--all` or the interactive picker) → installs through
  **every** available host, resolved at run time so a host CLI installed earlier
  in the same batch is picked up. (This is also why `--all` does not abort when
  several hosts are present.)

**Upgrade, without `--host`:** refreshes **every** host the skill is currently
installed through (not just the first), so a multi-host install stays fully up
to date; it never runs an update against a host that never installed it, falling
back to the preferred host.

**Upgrade `--host all`:** targets only the on-`PATH` hosts that actually have the
skill; hosts that are present but do not have it are **skipped with a warning**
(`Skipping upgrade for <host>: Azure Skills is not installed on it.`), and an
error is returned only when **no** available host has it.

## Install / upgrade behavior

- **Install** runs the host's marketplace-add (tolerating an
  "already registered/added" response as idempotent success), then the plugin
  install, and verifies via the host's plugin list with retry + exponential
  backoff (the initial backoff is injectable so tests stay fast).
- **Per-host verification is host-scoped:** after installing through host X, azd
  confirms the skill is detectable *through host X specifically* (via
  `DetectSkillHosts`) and records that host's version. Verifying via the generic
  detector would report success whenever *any* host already had the skill — so
  `--host claude` while copilot already had it could be falsely reported as
  success with copilot's version.

## Prerequisites

- **Hard — agentic CLI host:** if none of the supported hosts is on `PATH`,
  install fails with an `errorhandler.ErrorWithSuggestion` that points the user
  at `azd tool install github-copilot-cli` and links the Copilot CLI install
  docs.
- **Soft — Node.js:** if `node` is missing, a warning is written to **stderr**
  (the Azure MCP server needs Node.js via `npx`) and the install proceeds.
  Writing to stderr keeps `azd tool install --output json` stdout clean.

## Detection

`detectSkill` runs each host's plugin-list command and uses a two-stage gate:
`PluginName` must appear in stdout, then the host's `VersionRegex` must capture a
version (the authoritative existence signal, since some hosts echo the queried
name even when the plugin is not installed). `DetectSkillHosts` probes **all**
configured hosts and returns each one the skill is installed through together
with its version (`InstalledSkillHost{Host, Version}`), which drives the
multi-host upgrade and the host-scoped install verification above.

## Agent integration

`internal/agent/copilot_agent_factory.go` — the Copilot session's
`requiredPlugins` source is updated to
`microsoft/azure-skills:.github/plugins/azure-skills` (the canonical Copilot
plugin directory, whose `.plugin/plugin.json` declares `name: "azure"`). Both
this agent-bootstrap path and `azd tool install azure-skills` land a plugin named
`azure`, and the agent dedupes required plugins on that name, so the two paths do
not double-install.

## Files changed

| File | What |
| --- | --- |
| `pkg/tool/manifest.go` | `ToolCategorySkill`, `SkillHost` struct, `azureSkills()` registration (copilot, claude) |
| `pkg/tool/detector.go` | `detectSkill`, `DetectSkillHosts`, `InstalledSkillHost{Host, Version}` |
| `pkg/tool/installer.go` | skill install/upgrade flow, host resolution + `--host all`, per-host install + host-scoped verify, injectable retry backoff |
| `pkg/tool/manager.go` | host-aware `InstallTools`/`UpgradeTools` options, `AvailableSkillHosts` |
| `cmd/tool.go` | `--host` flag on install & upgrade, host-option resolution, multiple-host guidance |
| `internal/agent/copilot_agent_factory.go` | `requiredPlugins` source |
| `*_test.go`, `cmd/testdata/*` | unit tests + regenerated command snapshots |

## Verification

- `go build ./...` ✓
- `go test ./pkg/tool/... ./cmd ./cmd/middleware ./internal/agent` ✓
- `go vet ./...` — clean
- `UPDATE_SNAPSHOTS=true go test ./cmd -run 'TestFigSpec|TestUsage'` — snapshots updated


cli/azd/pkg/tool/manifest.go
```diff
@@ -261,7 +302,7 @@ func azureMCPServer() *ToolDefinition {
 		Description:   "Model Context Protocol server for Azure resource interaction.",
 		Category:      ToolCategoryServer,
 		Priority:      ToolPriorityOptional,
-		Website:       "https://github.com/Azure/azure-mcp",
+		Website:       "https://github.com/microsoft/mcp",
```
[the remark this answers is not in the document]
hemarina: "https://github.com/microsoft/mcp" is the right repo

vhvb1989 (approved):

cli/azd/pkg/tool/manifest.go
```diff
@@ -90,6 +126,12 @@ type ToolDefinition struct {
 	// InstallStrategies maps a GOOS value ("windows", "darwin", "linux") to the
 	// platform-specific installation strategy.
 	InstallStrategies map[string]InstallStrategy
+	// SkillHosts describes the agent CLI hosts that can install this tool when
+	// Category == ToolCategorySkill. Hosts are evaluated in order; the first
+	// one whose binary is on PATH is used. Platform-agnostic because the host
+	// CLI's plugin command syntax does not vary between operating systems.
+	// Ignored for other categories.
+	SkillHosts []SkillHost
```
JeffreyCA, line 136: If a user has multiple CLI hosts installed (e.g. `copilot`, `claude`, and `codex`), the current evaluation logic means that we only install the skill for `copilot`.

I'm wondering if we could give users more control in selecting which hosts to install the skill for. They could be using `codex` as their main CLI but they won't get the skill because they happen to have `copilot` or `gemini` installed.

hemarina: Good point. We should add commands for:
azd tool install <skills> --all
azd tool install <skills> --host <agent>

cli/azd/pkg/tool/installer.go
```diff
@@ -321,6 +329,286 @@ func (i *installer) run(
...
+	if _, err := i.commandRunner.Run(ctx, runArgs); err != nil {
+		return fmt.Errorf(
+			"running `%s %s`: %w",
+			host.Host, strings.Join(cmd, " "), err,
+		)
+	}
+
+	// For codex, need to install after marketplace is updated
+	if upgrade && host.Host == "codex" {
+		runArgs = exec.NewRunArgs(host.Host, host.PluginInstallCommand...)
```
RickWinter: [nit] The codex install-after-update command is run without `.WithInteractive(true)`, unlike the primary command on line 554 and the fresh-install path. If `codex plugin add azure@azure-skills` ever prompts, the upgrade flow would hang or fail where a fresh install would not. Add `.WithInteractive(true)` here for consistency. (azd-code-reviewer)

cli/azd/pkg/tool/installer.go
```diff
@@ -321,6 +329,286 @@ func (i *installer) run(
...
+	if len(tool.SkillHosts) == 0 {
+		result.Error = fmt.Errorf(
+			"%s has no SkillHosts configured", tool.Name,
+		)
+		result.Duration = time.Since(start)
+		return result, nil
+	}
+
+	// 1. HARD prerequisite: at least one supported host must be on PATH.
+	host, hostErr := i.pickSkillHost(tool)
```
RickWinter: [suggestion] `pickSkillHost` returns the first host on PATH for both install and upgrade, regardless of which host actually has the skill installed. Consider a user with both `apm` and `copilot` on PATH who installed via copilot: `azd tool upgrade azure-skills` will pick `apm` (listed first) and run `apm update microsoft/azure-skills -g` against a package apm never installed — which will likely error or silently install a second copy. For the upgrade path, it'd be more robust to prefer the host that the detector reports as already having the skill. (azd-code-reviewer)

cli/azd/pkg/tool/detector.go
```diff
@@ -435,6 +438,91 @@ func (d *detector) detectAzdExtension(
...
+		//   1. PluginName must appear in the output (fast pre-filter), and
+		//   2. VersionRegex must capture a version (authoritative —
+		//      every host's regex anchors on a token that is present
+		//      only in installed-plugin output, e.g. claude's
+		//      "Version:" line or gemini's "Release tag:" line).
+		output := result.Stdout
+		if host.Host == "gemini" {
+			output = result.Stdout + "\n" + result.Stderr
+		}
+		if host.PluginName != "" &&
```
RickWinter: [nit] The `host.PluginName != ""` guard here is dead — the loop already `continue`s when `host.PluginName == ""` at the top. The two-stage gate comment above is the valuable part; the redundant check can be dropped to `if !strings.Contains(output, host.PluginName)`. (azd-code-reviewer)

cli/azd/pkg/tool/installer.go
```diff
@@ -321,6 +329,286 @@ func (i *installer) run(
...
+	}
+
+	// 4. Verify by detecting the skill via the host's plugin list.
+	//    Plugin-list output sometimes lags the install action (the
+	//    pre-existing copilot CLI integration documents the same race —
+	//    see internal/agent/copilot/cli.go), so retry a few times with
+	//    exponential backoff to match the package-manager install path.
+	const maxAttempts = 4 // 1 initial + 3 retries
+	var status *ToolStatus
+	backoff := 1 * time.Second
```
RickWinter: [suggestion] The verification backoff is hardcoded (1s, doubling). `TestRunSkill_VerificationFails_AfterRetries` therefore sleeps ~7s of real wall time (1+2+4) on every run. Consider making the initial backoff injectable (struct field defaulting to 1s) so tests can set it to ~1ms and keep the suite fast. (azd-code-reviewer)

cli/azd/internal/agent/copilot_agent_factory.go
```diff
@@ -25,7 +25,7 @@ type pluginSpec struct {
 
 // requiredPlugins lists plugins that must be installed before starting a Copilot session.
 var requiredPlugins = []pluginSpec{
-	{Source: "microsoft/GitHub-Copilot-for-Azure:plugin", Name: "azure"},
+	{Source: "microsoft/azure-skills:.github/plugins/azure-skills", Name: "azure"},
```
RickWinter, line 28: [suggestion] This changes the auto-install source for the existing Copilot agent session from `microsoft/GitHub-Copilot-for-Azure:plugin` to `microsoft/azure-skills:.github/plugins/azure-skills`, which goes through `copilot plugin install <source>` (see `copilot_agent.go` → `InstallPlugin`). Meanwhile the new `azd tool install azure-skills` path installs the same plugin into copilot via `plugin marketplace add` + `plugin install azure@azure-skills`. Two different mechanisms installing the "same" azure plugin into the same host is a footgun — worth confirming the raw `microsoft/azure-skills:.github/plugins/azure-skills` ref is valid for `copilot plugin install` and that both paths land an identically-named plugin (the agent dedupes on `Name: "azure"`). I don't see a test covering this `requiredPlugins` value. (azd-code-reviewer)

JeffreyCA (approved):
Thanks for the revisions!

A further enhancement to consider is using a multi-select prompt to pick the hosts, and falling back to current error in non-interactive mode. That way instead of running into the error/suggestion and having to re-run with `--host` or `--all`, the prompt takes them all the way to success.

JeffreyCA (approved):

JeffreyCA (approved):

PR #8706, opened 2026-06-18
feat(agents): default hosted agent deploy mode to code for Python/.NET

### LINKED ISSUE #8707: Default hosted agent deploy mode to code deploy for Python/.NET
## Background

Code deploy (source ZIP upload) for hosted agents has been available for a couple of weeks. Telemetry looks healthy and the service team has confirmed there are no concerns with making it the default path.

## Motivation

Right now, when users run `azd ai agent init`, the default deploy mode is **container** (Docker image + ACR). That means every new Python or .NET agent project gets wired up with Azure Container Registry by default — even when the user has no need for a custom container.

Code deploy is simpler: no Dockerfile, no ACR, no image push. The agent source code is uploaded as a ZIP and the platform handles the rest. It's a better out-of-the-box experience for the vast majority of users.

Since the service only supports code deploy for **Python and .NET** projects, non-Python/.NET projects will continue to fall back to container deploy. Users who specifically need a container image can always opt in via `--deploy-mode container`.

## What we want to change

- Make **code deploy the default** for Python and .NET hosted agents across all init flows: interactive prompt, `--no-prompt`, and `-m` manifest auto-select.
- Relax the requirement to explicitly pass `--runtime` and `--entry-point` with `--deploy-mode code --no-prompt` — auto-detect them from the project instead, so a minimal CI deploy needs no extra flags.
- Update help text, examples, and docs to reflect the new default.

## Acceptance criteria

- A Python or .NET project initialized with no `--deploy-mode` flag defaults to code deploy.
- Passing `--deploy-mode container` still works and selects container deploy.
- `--runtime` and `--entry-point` are optional in `--no-prompt` mode and are auto-detected when omitted.
- Non-Python/.NET projects continue to use container deploy.

huimiu:
Closes #8707

## Summary

Switches the hosted agent deploy mode default from **container** (Docker image) to **code** (source ZIP upload) for Python and .NET projects.

Code deploy is the recommended and simpler path — no Dockerfile or ACR needed. Container deploy remains fully supported via `--deploy-mode container`.

## Behavior change

| Scenario | Before | After |
|---|---|---|
| Python/.NET — interactive prompt default | Container | **Code** |
| Python/.NET — `--no-prompt` (no `--deploy-mode` flag) | Container | **Code** |
| Python/.NET — `-m` manifest auto-select | Container | **Code** |
| Non-Python/.NET project | Container | Container (unchanged) |
| Explicit `--deploy-mode container` | Container | Container (unchanged) |
| Explicit `--deploy-mode code` | Code | Code (unchanged) |

## Related relaxation

`--runtime` and `--entry-point` are now optional even with `--deploy-mode code --no-prompt`. They are auto-detected from the project when omitted, so a minimal CI deploy is just:

```sh
azd ai agent init --no-prompt --project-id "<resource-id>"
```

## Changes

- `init_from_code.go` — `promptDeployMode`: code-first default in `noPrompt`, `userProvidedManifest`, and interactive-prompt branches; reordered Select choices.
- `init.go` — removed the "runtime/entry-point required" no-prompt validation; trimmed `validateCodeDeployInput` signature; updated flag help text and example.
- Tests — `TestPromptDeployMode_FlagOverride` and `TestCodeDeployFlagValidation` updated.
- `CHANGELOG.md` — `## Unreleased` entry with migration note.

### DIFF
--- a/cli/azd/extensions/azure.ai.agents/CHANGELOG.md
+++ b/cli/azd/extensions/azure.ai.agents/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## 0.1.41-preview (2026-06-19)
 
+- [[#8706]](https://github.com/Azure/azure-dev/pull/8706) `azd ai agent init` now defaults to **code deploy** (source ZIP upload) for Python and .NET hosted agents instead of container (Docker image) deploy. This applies to the interactive prompt, `--no-prompt`, and `-m` manifest flows. Projects that are not Python/.NET continue to use container deploy. To keep using a Docker image, pass `--deploy-mode container` (or select "Container Image (Docker)" in the prompt). With code deploy in `--no-prompt`, `--runtime` and `--entry-point` are now auto-detected from the project when omitted.
 - [[#8731]](https://github.com/Azure/azure-dev/pull/8731) Improve the post-deploy `Next:` guidance with a stacked layout that puts each command on its own line above its description, adds a blank line between suggestions, and highlights `azd` commands. The new layout applies across deploy, `azd ai agent show`, `init`, and `doctor`. Thanks @therealjohn for the contribution!
 - [[#8645]](https://github.com/Azure/azure-dev/pull/8645) Detect VNET-injected Foundry accounts during `azd ai agent init` and skip remote builds up front so hosted container agents use local builds without a failing remote-build attempt first. Thanks @m5i-work for the contribution!
 - [[#8714]](https://github.com/Azure/azure-dev/pull/8714) Show a tracing disclaimer when `azd ai agent init` connects or adds an Application Insights connection. Thanks @therealjohn for the contribution!
--- a/cli/azd/extensions/azure.ai.agents/internal/cmd/init.go
+++ b/cli/azd/extensions/azure.ai.agents/internal/cmd/init.go
@@ -886,9 +886,11 @@ from code-deploy ZIP packaging (uses .gitignore syntax).`,
   # Initialize from local agent code
   azd ai agent init --src ./src/my-agent --agent-name my-unique-agent
 
-  # Non-interactive code deploy (CI/CD)
-  azd ai agent init --no-prompt --project-id "<resource-id>" \
-    --deploy-mode code --runtime python_3_13 --entry-point app.py`,
+  # Non-interactive code deploy (CI/CD) — code is the default for Python/.NET
+  azd ai agent init --no-prompt --project-id "<resource-id>"
+
+  # Non-interactive container deploy
+  azd ai agent init --no-prompt --project-id "<resource-id>" --deploy-mode container`,
 		Args: cobra.MaximumNArgs(1),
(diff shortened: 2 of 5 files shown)


therealjohn (changes requested):
Agree with the switch to default - but neesd to wait until end of June target for platform readiness of the service. 

Requesting changes just to block until then.

jongio (commented):
Logic and tests look correct. The precedence chain in promptDeployMode is clear, the removal of early runtime/entry-point validation is reasonable given auto-detection downstream, and the existing test at line 918 (lag=code works even when showCodeDeploy=false) confirms the explicit-flag-override behavior the Copilot bot flagged.

One issue: the CHANGELOG entry targets a version that's already published (see inline comment).

cli/azd/extensions/azure.ai.agents/CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## 0.1.41-preview (2026-06-19)
 
+- [[#8706]](https://github.com/Azure/azure-dev/pull/8706) `azd ai agent init` now defaults to **code deploy** (source ZIP upload) for Python and .NET hosted agents instead of container (Docker image) deploy. This applies to the interactive prompt, `--no-prompt`, and `-m` manifest flows. Projects that are not Python/.NET continue to use container deploy. To keep using a Docker image, pass `--deploy-mode container` (or select "Container Image (Docker)" in the prompt). With code deploy in `--no-prompt`, `--runtime` and `--entry-point` are now auto-detected from the project when omitted.
```
jongio, line 5: 0.1.41-preview is already in the extension registry (published). This entry should go under a new ## Unreleased or ## 0.1.42-preview header so it ships with the next release rather than appearing to retroactively modify a version users already have installed.

jongio (commented):
Logic and tests look correct. The precedence chain in `promptDeployMode` is clear, the removal of early runtime/entry-point validation is reasonable given auto-detection downstream, and the existing test at line 918 (`flag=code works even when showCodeDeploy=false`) confirms the explicit-flag-override behavior the Copilot bot flagged.

One issue: the CHANGELOG entry targets a version that's already published (see inline comment).

cli/azd/extensions/azure.ai.agents/CHANGELOG.md
jongio, line 5: 0.1.41-preview is already in the extension registry (published). This entry should go under a new `## Unreleased` or `## 0.1.42-preview` header so it ships with the next release rather than appearing to retroactively modify a version users already have installed.

jongio (approved):
One item still outstanding: the CHANGELOG entry sits under the .1.41-preview header which is already published. Move it to a new ## Unreleased section (or ## 0.1.42-preview) so it ships with the next release. See my earlier inline comment for details.

Otherwise the logic, tests, and flag help text are all correct. The deploy-mode precedence chain is clear and the relaxed validation for --runtime/--entry-point makes sense given auto-detection downstream.

PR #8739, opened 2026-06-20
fix: eliminate cache-memory dependency in update-instructions workflow

### LINKED ISSUE #8702: [aw] Update Instructions From PR Reviews has cache-memory miss misconfiguration
### Workflow Failure

**Workflow:** [Update Instructions From PR Reviews](https://github.com/Azure/azure-dev/blob/main/.github/workflows/update-instructions-from-pr-reviews.md)  
**Branch:** main  
**Run:** https://github.com/Azure/azure-dev/actions/runs/27725890498

> [!WARNING]
> <details>
> <summary>Cache Configuration Problem: cache miss detected despite cache-memory being configured.</summary>
> 
> The agent reported a cache miss (`missing_data` with `reason: cache_memory_miss`) even though cache-memory is configured and was available. This likely indicates the prompt is misconfigured and the agent cannot locate the correct file path within the cache directory.
> 
> Review the [cache-memory configuration](https://github.github.com/gh-aw/reference/cache-memory/) and ensure the agent prompt correctly references files inside the cache directory.
> 
> **File naming convention:** Cache files are stored at `/tmp/gh-aw/cache-memory/` (default) or `/tmp/gh-aw/cache-memory-{id}/` for additional caches. Use descriptive file and directory names with subdirectories for organization.
> 
> </details>

### Action Required

**Assign this issue to an agent** to debug and fix the issue.

<details>
<summary>Debug with any coding agent</summary>

Use this prompt with any coding agent (GitHub Copilot, Claude, Gemini, etc.):

````
Debug the agentic workflow failure using https://raw.githubusercontent.com/github/gh-aw/main/debug.md

The failed workflow run is at https://github.com/Azure/azure-dev/actions/runs/27725890498
````

</details>

<details>
<summary>Manually invoke the agent</summary>

Debug this workflow failure using your favorite Agent CLI and the `agentic-workflows` prompt.

- Start your agent
- Load the `agentic-workflows` skill from `.github/skills/agentic-workflows/SKILL.md` or <https://github.com/github/gh-aw/blob/main/.github/skills/agentic-workflows/SKILL.md>
- Type `debug the agentic workflow update-instructions-from-pr-reviews failure in https://github.com/Azure/azure-dev/actions/runs/27725890498`

</details>

> [!TIP]
> <details>
> <summary>Stop reporting this workflow as a failure</summary>
>
> To stop a workflow from creating failure issues, set `report-failure-as-issue: false` in its frontmatter:
> ```yaml
> safe-outputs:
>   report-failure-as-issue: false
> ```
>
> </details>

> Generated from [Update Instructions From PR Reviews](https://github.com/Azure/azure-dev/actions/runs/27725890498) · 410.7 AIC · [◷](https://github.com/search?q=repo%3AAzure%2Fazure-dev+is%3Aissue+%22gh-aw-workflow-id%3A+update-instructions-from-pr-reviews%22&type=issues)
> - [ ] expires  on Jun 24, 2026, 11:31 PM UTC

### PR DESCRIPTION
**Pull request by vhvb1989**
## Summary

Fixes #8702

Replace the `cache-memory` tool with a GitHub API query for the last successful workflow run in the `update-instructions-from-pr-reviews` agentic workflow.

## Problem

On first run, `last-run.txt` didn't exist in cache. The agent correctly called `missing_data` with `cache_memory_miss` (as instructed by the cache-memory system), but gh-aw's detection layer interpreted this as a prompt misconfiguration rather than a legitimate first-run condition.

## Fix

Eliminate the cache entirely. Instead of reading/writing `last-run.txt` in cache-memory, the agent now queries the GitHub API for the most recent successful run of this workflow to determine the PR range starting point.

### Changes
- **Removed `cache-memory:`** tool from the workflow frontmatter
- **Rewrote Step 1** fallback to query the last successful workflow run via GitHub API
- **Removed Step 8** ("Record the run") — no cache to write to; the workflow run history IS the state
- **Added `actions: read`** permission so the agent can list workflow runs

### Why not just fix the prompt?

The cache-memory approach is inherently fragile for a weekly workflow:
- GitHub Actions cache has a 7-day unused-entry eviction policy
- A weekly Monday schedule sits right at the eviction edge — a skipped week means cache loss and redundant 6-month mining every time
- The GitHub API workflow run history is always available and is the natural source of truth

Per [richardpark-msft's comment](https://github.com/Azure/azure-dev/issues/8702#issuecomment-4746234755): *"If you can get it to just query that detail directly you can eliminate a dependency on that cache altogether."*

### CODE REVIEW

**Review by hemarina — APPROVED** [review]

**Review by jongio — COMMENTED** [review]
Clean fix. Replacing the cache-memory tool with a workflow-run-history query is the right call: you eliminate a 7-day eviction hazard and gain a source of truth that's always available. The fallback chain (explicit date > last successful run timestamp > 6-month cap) handles the edge cases well.

Two quick notes:

1. **Lock file**: the `.lock.yml` wasn't regenerated in this branch (same SHA as `main`). You'll need a `gh aw compile` pass before merge so the runtime picks up the new logic.

2. **Bot comment on workflow name is a false positive**: the instruction says "list recent workflow runs for this workflow (`Update Instructions From PR Reviews`)" and the lock.yml's `name:` field is exactly that string, so the agent will resolve it correctly.

Good to go once the lock file is refreshed.

**Review by RickWinter — COMMENTED** [review]
This replaces the cache-memory high-water mark with a query against the workflow's own successful-run history, which is the right direction and drops a fragile weekly-eviction dependency. The one thing blocking it is that the new step 2 relies on listing workflow runs, and the `github` tool only enables the `pull_requests`, `issues`, and `repos` toolsets. Listing workflow runs lives in the `actions` toolset, so the added `actions: read` permission has no tool to drive it, and step 2 will silently fall through to the 6 month scan (step 3) on every run, which is the exact redundant mining this PR is trying to remove. Enabling the `actions` toolset on the `github:` tool should close that gap. The `updated_at` of the last successful run is a reasonable starting point; there is a small window where PRs merged while the previous run was executing could be reprocessed, but for a weekly instruction-mining job that is harmless. Please address the toolset before merge.

File: .github/workflows/update-instructions-from-pr-reviews.md [hunk 1]
```diff
@@ -154,11 +152,13 @@ Decide which merged PRs to examine, in this order of preference:
 
 1. If the **Since** run parameter above is a real date (not `not set`), mine
    PRs merged on/after that date.
-2. Otherwise, read the **cache memory** for a marker file named
-   `last-run.txt`. If it exists, mine PRs merged after that timestamp ("since
-   last run").
-3. Otherwise (first ever run, no marker), mine PRs merged within the last 6
-   months, capped at the **Max PRs** run parameter.
+2. Otherwise, use the GitHub tools to list recent workflow runs for **this
```
RickWinter, line 155: This step now depends on listing workflow runs, but the `github` tool above only enables `toolsets: [pull_requests, issues, repos]`. The workflow-run listing lives in the github MCP `actions` toolset, which is not turned on here, so granting `actions: read` at the permissions level is not enough on its own. As written the agent has no tool to satisfy step 2 and will fall through to the 6 month scan in step 3 on every run, which is the redundant mining this PR is trying to avoid. Can we add `actions` to the toolsets list on the `github:` tool so the run-history query actually has a tool backing it?

jongio (dismissed):
Two things need fixing:

1. **Missing actions toolset**: toolsets: [pull_requests, issues, repos] doesn't include actions. The workflow-run listing API lives in that toolset, so step 2 will silently fail and the agent will always fall through to the 6-month scan. Add actions to the array.

2. **Lock file stale**: The .lock.yml still reflects old cache-memory logic. Run gh aw compile to regenerate.

richardpark-msft (approved):

.github/workflows/update-instructions-from-pr-reviews.md
```diff
@@ -0,0 +1,289 @@
...
+
+Decide which merged PRs to examine, in this order of preference:
+
+1. If the **Since** run parameter above is a real date (not `not set`), mine
+   PRs merged on/after that date.
+2. Otherwise, use the GitHub tools to list recent workflow runs for **this
+   workflow** (`update-instructions-from-pr-reviews`) in the repository this
+   workflow runs in. Find the most recent **successful** (completed with
+   `conclusion: success`) run and use its `created_at` timestamp as the
+   starting point — mine PRs merged after that date.
```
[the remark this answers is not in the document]
vhvb1989: Accepted — both fixes (updated_at + workflow display name) were addressed in the rebase (479d5b6).

.github/workflows/update-instructions-from-pr-reviews.md
vhvb1989: Great catch — added `actions` to the github toolsets in 26cfbd3. Without it the agent had no tool to list workflow runs and would always fall through to the 6-month scan.

jongio (approved):
Both issues from my previous review are addressed:

1. `actions` toolset added to the `toolsets` array in the .md source, and correctly reflected in the lock.yml (`GITHUB_TOOLSETS` now includes `actions`).
2. Lock file regenerated with gh-aw v0.81.6.

No new issues in the incremental changes.

JeffreyCA (approved):

.github/workflows/update-instructions-from-pr-reviews.md
```diff
@@ -154,11 +152,13 @@ Decide which merged PRs to examine, in this order of preference:
...
-2. Otherwise, read the **cache memory** for a marker file named
-   `last-run.txt`. If it exists, mine PRs merged after that timestamp ("since
-   last run").
-3. Otherwise (first ever run, no marker), mine PRs merged within the last 6
-   months, capped at the **Max PRs** run parameter.
+2. Otherwise, use the GitHub tools to list recent workflow runs for **this
+   workflow** (`Update Instructions From PR Reviews`) in the repository this
+   workflow runs in. Find the most recent **successful** (completed with
+   `conclusion: success`) run and use its `updated_at` timestamp as the
+   starting point — mine PRs merged after that date.
```
RickWinter, line 159: Using the run's `updated_at` (completion time) as the lower bound opens a small gap. Any PR merged after this workflow listed PRs but before it finished gets skipped: the next run only mines merges after the previous completion, so that PR is never seen. Since the previous approach recorded the high-water mark of mined PRs, this is a behavior change. Consider using the run's `run_started_at` instead. It is strictly earlier, so the worst case is reprocessing a PR, which is harmless here because themes only promote across multiple PRs and get deduped anyway. Missing a PR is the failure mode we actually care about.

```suggestion
2. Otherwise, use the GitHub tools to list recent workflow runs for **this
   workflow** (`Update Instructions From PR Reviews`) in the repository this
   workflow runs in. Find the most recent **successful** (completed with
   `conclusion: success`) run and use its `run_started_at` timestamp as the
   starting point — mine PRs merged after that date.
```

RickWinter (approved):

RickWinter (approved):

jongio (approved):
Re-approving after the incremental commit. Using `run_started_at` instead of `updated_at` is the right field: it reflects when the workflow run actually began and won't shift if the run is retried or annotated later.

PR #8754, opened 2026-06-22
Add PR gate tests for azd ai agent extension

### LINKED ISSUE #8760: Add PR gate tests for azd ai agent extension
## Summary

Add automated PR gate tests for the \zd ai agent\ extension using the existing recording/playback proxy infrastructure.

## Details

- Tier 0: Offline tests (version, help, flag validation, sample list) — no Azure credentials needed
- Tier 1: Recording playback tests (full \init --no-prompt --project-id\ flow with ARM calls replayed from cassette)
- Negative control test proving cassette consumption (prevents false-green)
- CI workflow integrated into \lint-ext-azure-ai-agents.yml\

## Implementation

PR: #8754

v1212:
Fixes #8760

## Summary

Adds automated PR gate tests for the `azd ai agent` extension using azd's existing recording/playback proxy infrastructure. These tests run in CI on every PR touching the extension, catching regressions without requiring live Azure credentials.

### 4-Process Recording Chain
```
Test → azd-record (host) → Extension (record-tagged binary) → ARM (via recording proxy → cassette)
```

The recording proxy intercepts all outbound HTTP from the extension, storing request/response pairs in YAML cassette files. During playback, the proxy replays stored responses — no network access needed.

### Test Tiers

| Tier | Mode | What it proves | Auth needed |
|------|------|----------------|-------------|
| **Tier 0** (offline) | No recording | Extension loads, parses flags, prints help/version, validates inputs | None |
| **Tier 1** (playback) | `AZURE_RECORD_MODE=playback` | Full `init --no-prompt` with ARM calls replayed from cassette — model catalog queries, project resolution, scaffold generation | Fake token via recording session |

### Tests (9 total, all passing)

**Tier 0 — Offline (6 tests):**
- `Test_AIAgent_Version` — extension version command
- `Test_AIAgent_Help` — help output
- `Test_AIAgent_Init_NoPrompt_MissingFlags` — missing required flags → proper error
- `Test_AIAgent_SampleList` — sample template listing
- `Test_AIAgent_SampleList_JSON` — JSON output with proper structure validation (json.Unmarshal + key check)
- `Test_AIAgent_Doctor_Help` — doctor subcommand help

**Tier 1 — Recording Playback (3 tests):**
- `Test_AIAgent_Init_NoPrompt_Defer` — init with `--deploy-mode defer` (no ARM calls, scaffold only)
- `Test_AIAgent_Init_NoPrompt_WithProject` — full init with `--project-id`, resolves Foundry project via ARM, auto-selects model from catalog, generates complete scaffold
- `Test_AIAgent_Init_NegativeControl_BadCassette` — **negative control**: empty cassette → init fails with "requested interaction not found", proving the cassette is genuinely consumed

### Key Assertions (WithProject)
- Exit 0 + success message
- Project structure: `azure.yaml`, `agent.yaml`, agent source directory
- **ARM resolution proof**: `.env` file contains `AZURE_AI_MODEL_DEPLOYMENT_NAME="gpt-4.1"` (resolved from cassette's model catalog, not a placeholder)
- **Cross-check**: `agent.yaml` has resolved value (not `${...}` env placeholder)

### New files

| File | Purpose |
|------|---------|
| `cli/azd/test/functional/ai_agents/ai_agent_test.go` | Tier 0 offline tests (6 tests) |
| `cli/azd/test/functional/ai_agents/ai_agent_recording_test.go` | Tier 1 recording tests (3 tests) + helpers |
| `cli/azd/test/functional/ai_agents/testdata/manifests/foundry-toolbox.yaml` | Local manifest (avoids GitHub auth in CI) |
| `cli/azd/test/functional/ai_agents/testdata/recordings/Test_AIAgent_Init_NoPrompt_WithProject.yaml` | 2.4 MB cassette (47 ARM interactions, model catalog trimmed) |
| `cli/azd/test/functional/ai_agents/testdata/recordings/Test_AIAgent_Init_NoPrompt_Defer.yaml` | Empty cassette (no ARM interactions) |
| `cli/azd/test/functional/ai_agents/testdata/recordings/Test_AIAgent_Init_NegativeControl_BadCassette.yaml` | Empty cassette for negative control |
| `cli/azd/extensions/azure.ai.agents/internal/pkg/recordproxy/transport.go` | Non-record build stub (no-op) |
| `cli/azd/extensions/azure.ai.agents/internal/pkg/recordproxy/transport_record.go` | Record-mode transport: sets `http.DefaultTransport` to route through proxy |
| `.github/workflows/lint-ext-azure-ai-agents.yml` | CI workflow: build, install extension, run Tier 0 + Tier 1 |

### Modified files

| File | Change |
|------|--------|
| `cli/azd/extensions/azure.ai.agents/internal/pkg/azure/client_options.go` | Simplified: removed redundant `proxyTransporter` struct — `*http.Client` directly satisfies `policy.Transporter` interface |
| `cli/azd/test/azdcli/cli.go` | Fixed token expiry format: use explicit UTC Z-suffix format (extension rejects `+00:00` from `time.RFC3339`) |
| `cli/azd/extensions/azure.ai.agents/cspell.yaml` | Added `recordproxy` to dictionary |

## CI Workflow Design

```yaml
# Tier 0: No auth, no recording — pure offline
go test -tags=record -run "Test_AIAgent_(Version|Help|...)" ./test/functional/

# Tier 1: Playback with fake auth
sudo ln -sf azd-record /usr/local/bin/azd
AZURE_RECORD_MODE=playback CLI_TEST_AZD_PATH=azd-record \
  go test -tags=record -run "Test_AIAgent_Init_(NoPrompt_...|NegativeControl_...)" ./test/functional/
```

Key CI considerations:
- **azd symlink only in Tier 1**: Extension gracefully skips auth when `azd` is missing (Tier 0). With `azd` present but not logged in → hard fail. The symlink is only created for Tier 1 where fake auth via recording session works.
- **Local manifest**: `-m` flag points to committed `testdata/manifests/foundry-toolbox.yaml` instead of GitHub URL, avoiding device-code auth in clean CI.
- **Extension registration**: Workflow creates `~/.azd/config.json` with extension path, matching azd's discovery mechanism.

## Cassette Size

The `WithProject` cassette is 2.4 MB (47 interactions). For context, the repo's existing 29 cassettes have median 0.97 MB and max 4.80 MB — this is within normal range. The cassette was trimmed from 26 MB by keeping only target models (gpt-4o, gpt-4.1, gpt-4o-mini) in model catalog responses.

## How to Re-record

```bash
# Set real credentials and project ID
export AZURE_RECORD_MODE=record
export TEST_FOUNDRY_PROJECT_ID="/subscriptions/.../projects/..."
export CLI_TEST_AZD_PATH=/path/to/azd-record
go test -tags=record -run Test_AIAgent_Init_NoPrompt_WithProject ./test/functional/ -v -timeout 10m
```

The cassette file is updated in-place. Trim if needed (model catalog responses are large).

## New CI Workflow

This PR adds a new workflow that takes effect immediately after merge to main:

| | |
|---|---|
| **File** | .github/workflows/lint-ext-azure-ai-agents.yml |
| **Name** | ext-azure-ai-agents-ci |
| **Target branch** | main |

**Trigger:** pull_request when changes are made to any of:
- cli/azd/extensions/azure.ai.agents/**
- cli/azd/test/functional/ai_agents/ai_agent_test.go
- cli/azd/test/functional/ai_agents/ai_agent_recording_test.go
- cli/azd/test/functional/testdata/recordings/Test_AIAgent_*
- .github/workflows/lint-ext-azure-ai-agents.yml

In short: any PR that touches the AI agent extension code, test files, recordings, or the workflow itself will automatically run lint, extension build, and the full test suite.

## Test Cases at a Glance

| # | Test | Tier | Purpose |
|---|------|------|---------|
| 1 | Test_AIAgent_Version | 0 | Extension binary loads and prints version |
| 2 | Test_AIAgent_Help | 0 | Help output renders correctly |
| 3 | Test_AIAgent_Init_NoPrompt_MissingFlags | 0 | Missing required flags produce clear error (not a crash) |
| 4 | Test_AIAgent_SampleList | 0 | Sample template listing works |
| 5 | Test_AIAgent_SampleList_JSON | 0 | JSON output is valid and structurally correct (json.Unmarshal + key check) |
| 6 | Test_AIAgent_Doctor_Help | 0 | Doctor subcommand help renders |
| 7 | Test_AIAgent_Init_NoPrompt_Defer | 1 | --deploy-mode defer scaffolds without ARM calls |
| 8 | Test_AIAgent_Init_NoPrompt_WithProject | 1 | Full init with --project-id — resolves Foundry project via ARM, auto-selects model from catalog, generates complete scaffold with correct .env values |
| 9 | Test_AIAgent_Init_NegativeControl_BadCassette | 1 | **Negative control**: empty cassette → init fails with "requested interaction not found", proving the cassette is genuinely consumed (prevents false-green) |

- **Tier 0** = offline, no recording, no auth — validates extension loads and parses correctly
- **Tier 1** = playback mode with pre-recorded ARM cassettes — validates full init flow without live Azure credentials


cli/azd/test/azdcli/cli.go
```diff
@@ -402,7 +402,7 @@ func startTestCredentialServer(t *testing.T) *httptest.Server {
 		}{
 			Status:    "success",
 			Token:     mockJWT,
-			ExpiresOn: expiresOn.Format(time.RFC3339),
+			ExpiresOn: expiresOn.UTC().Format("2006-01-02T15:04:05Z"),
 		}
```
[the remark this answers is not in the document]
v1212: Fixed. Added a comment explaining why we can't use 	ime.RFC3339 here — the extension parser only accepts the literal Z-suffix form (2006-01-02T15:04:05Z), while 	ime.RFC3339 may emit +00:00 offset which the parser rejects.

cli/azd/extensions/azure.ai.agents/internal/pkg/recordproxy/transport_record.go
```diff
@@ -0,0 +1,41 @@
...
+	if !ok {
+		return
+	}
+	proxyUrl, err := url.Parse(val)
+	if err != nil {
+		panic(err)
+	}
+
+	transport := http.DefaultTransport.(*http.Transport).Clone()
+	transport.TLSClientConfig = &tls.Config{
```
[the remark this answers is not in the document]
v1212: Fixed. Added descriptive panic messages and safe type assertion with checked ok.

cli/azd/test/functional/ai_agent_test.go
```diff
@@ -0,0 +1,224 @@
...
+
+	dir := tempDirWithDiagnostics(t)
+	cli := azdcli.NewCLI(t)
+	cli.WorkingDirectory = dir
+	cli.Env = append(cli.Env, os.Environ()...)
+
+	result, err := cli.RunCommand(ctx, "ai", "agent", "sample", "list", "--output", "json")
+	require.NoError(t, err, "stdout=%s, stderr=%s", result.Stdout, result.Stderr)
+	require.Contains(t, result.Stdout, "[")
+}
```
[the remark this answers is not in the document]
v1212: Fixed. Now uses \json.Unmarshal\ to validate the output is proper JSON and checks for the expected \	emplates\ key.

cli/azd/test/functional/ai_agents/ai_agent_test.go
```diff
@@ -0,0 +1,303 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
trangevi, line 1: Does this need to live in the broader azd test directory, or can it be moved inside the extension? Here it will require azd signoff for every change, not to mention that it would be best to keep the tests more closely tied to the extension

cli/azd/test/functional/ai_agent_test.go
```diff
@@ -0,0 +1,303 @@
...
+	cli := azdcli.NewCLI(t)
+	cli.WorkingDirectory = dir
+	cli.Env = append(cli.Env, os.Environ()...)
+
+	result, err := cli.RunCommand(ctx, "ai", "agent", "doctor", "--help")
+	require.NoError(t, err, "stdout=%s, stderr=%s", result.Stdout, result.Stderr)
+	require.Contains(t, result.Stdout, "doctor")
+}
+
+// --- Tier 1: Recording tests (ARM calls replayed from cassette) ---
```
trangevi: NIT: Perhaps the tier 1 tests should go in a separate file, to better distinguish between the tests which need recordings and the ones which don't?

trangevi (changes requested):

.github/workflows/lint-ext-azure-ai-agents.yml
mikeharder: why do you need write permissions to pull requests for a linting job?  I believe this won't work on fork PRs.  and in general, it's simpler and more secure if actions like this only need read access to the repo, and just report their results via console log or uploading an artifact.

.github/workflows/lint-ext-azure-ai-agents.yml
```diff
@@ -20,3 +22,67 @@ jobs:
     uses: ./.github/workflows/lint-go.yml
     with:
       working-directory: cli/azd/extensions/azure.ai.agents
+
+  test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
```
mikeharder: @copilot: Update all actions added in this PR to their latest major versions

cli/azd/test/functional/ai_agents/ai_agent_test.go
v1212: The tests need to live in cli/azd/test/functional/ because they import two packages from the azd Go module:

cli/azd/test/recording — recording proxy infrastructure (cassette load, playback server, sanitizers)
cli/azd/test/azdcli — CLI runner + fake credential server
The extension is a separate Go module and can't import these. This is the same pattern used by all other recording-based functional tests in the repo (e.g., testdata/recordings/ cassettes + recording.Start()).

Happy to discuss if you see an alternative structure — but as far as I can tell, the recording infra is tightly coupled to the azd module boundary.

cli/azd/test/functional/ai_agent_test.go
v1212: Done — split into ai_agent_test.go (Tier 0, offline only) and ai_agent_recording_test.go (Tier 1, requires cassettes + recording session).

.github/workflows/lint-ext-azure-ai-agents.yml
v1212: Good point about fork PRs. However,  pull-requests: write  is required here because our workflow calls the shared reusable workflow  lint-go.yml , which declares that permission (for posting lint annotations as PR comments). All other extension lint workflows in this repo  │ ( lint-ext-azure-ai-connections.yml ,  lint-ext-azure-appservice.yml , etc.) follow the same pattern. The fork PR limitation is a repo-wide constraint from  lint-go.yml 's design — if that's worth changing, it would be a separate cross-cutting fix to  lint-go.yml  itself.

.github/workflows/lint-ext-azure-ai-agents.yml
v1212: Updated to  actions/checkout@v6  and  actions/setup-go@v6 , matching the rest of the repo.

cli/azd/test/functional/ai_agents/ai_agent_test.go
trangevi: If this has to be here, it is possible to at least add all of our stuff to some nested folders? I'm thinking if that's possible, then maybe we can add some codeowners changes to make it so the azd core team doesn't have to sign off on all of our test changes.

@tg-msft maybe you have suggestions here on how we can structure this?

.github/workflows/lint-ext-azure-ai-agents.yml
```diff
@@ -13,10 +16,74 @@ concurrency:
...
 permissions:
   contents: read
-  pull-requests: write
+  pull-requests: write  # required by reusable workflow lint-go.yml
 
 jobs:
   lint:
     uses: ./.github/workflows/lint-go.yml
     with:
       working-directory: cli/azd/extensions/azure.ai.agents
```
trangevi, line 22: Why does this have a lint job? Or since you're editing an existing pipeline, maybe the question is why are you editing an existing pipeline instead of creating a new one? I'd expect the lint pipeline to be something quick, just running the linter, and something else to run the concrete tests, so not sure why you're doubling up here?

jongio (commented):
Solid work on the recording/playback test infrastructure. The tiered approach (Tier 0 offline, Tier 1 playback) is clean and the negative control test is a nice touch for proving cassette consumption. Cassette sanitization looks good (zeroed subscription IDs, fake resource names, sanitized auth headers).

A few observations:

**Architecture (positive):**
- Splitting the workflow into lint vs. test is the right call. The lint workflow (`lint-go.yml` reusable) needs `pull-requests: write` for annotations; the test workflow only needs `contents: read`. Clean separation of concerns.
- CODEOWNERS entry for the test directory means the AI agents team owns these without requiring core team signoff. Good.
- Using `//go:build record` ensures these tests never run accidentally in normal `go test ./...` without the tag.

**One thing to consider:**
- `transport_record.go` init() replaces `http.DefaultTransport` globally when the `record` build tag is active. The comment on L41 acknowledges this covers `http.Client{}` usage, and `client_options.go` separately handles Azure SDK clients. This is fine for the current test infrastructure, but if future code in the extension creates HTTP clients via other patterns (e.g., third-party SDKs that accept a custom transport parameter), those would silently bypass the proxy. Worth a brief comment in the package doc noting "all HTTP traffic is intercepted when this tag is active" so future contributors aren't surprised.

**Re: open reviewer threads:**
- trangevi's question about nesting: the tests already moved to `cli/azd/test/functional/ai_agents/` with CODEOWNERS, which addresses the core concern. Further nesting would add complexity without clear benefit since Go packages are already the boundary.
- trangevi's question about the lint workflow edit: the PR is splitting the pre-existing combined `ext-azure-ai-agents-ci` workflow into two focused workflows (lint + test). The lint job isn't new; it's the original workflow renamed. The test job is genuinely new. The commit message on HEAD explains this clearly.

cli/azd/test/functional/ai_agents/ai_agent_test.go
v1212: Done in the latest commit. Moved all AI agent tests to `cli/azd/test/functional/ai_agents/` as a self-contained Go package, and added a CODEOWNERS entry (same owners as `/cli/azd/extensions/azure.ai.agents/`) so the AI agents team owns the test directory without requiring azd core team signoff.

The tests use a local `helpers_test.go` for shared utilities (`newTestContext`, `tempDirWithDiagnostics`) to keep the package independent from the parent `functional` package.

.github/workflows/lint-ext-azure-ai-agents.yml
v1212: Good call — split it. To clarify: this was a **new** workflow (not editing an existing one), but you're right that combining lint and tests in one file muddies responsibilities.

Now separated into two parallel workflows:

1. `lint-ext-azure-ai-agents.yml` (`ext-azure-ai-agents-lint`) — lint only, calls the shared `lint-go.yml` reusable workflow. Needs `pull-requests: write` for lint annotations.
2. `test-ext-azure-ai-agents.yml` (`ext-azure-ai-agents-test`) — builds extension + runs Tier 0/Tier 1 tests. Only needs `contents: read`, so it works on fork PRs too.

Both trigger in parallel on the same path patterns targeting `main`.

v1212:
Thanks for the thorough review @jongio ! Added a package doc comment to `transport_record.go` noting that all HTTP traffic is intercepted when the record tag is active, and that third-party SDKs accepting custom transport parameters should use this package's `Transport` to ensure capture. Good callout for future contributors.

vhvb1989:
**Proposal: replace the manual install with the official `azd x` extension dev flow**

The current "Install extension" step copies the binary into `~/.azd/extensions/...` and hand-writes `~/.azd/config.json`'s `extension.installed` entry via an inline `python3` heredoc. That hardcodes `namespace`, `capabilities`, `displayName`, `description`, `version`, `usage`, and `path` — all of which already live in `extension.yaml`. This duplicates metadata and will silently drift: if `extension.yaml` changes (e.g. capabilities/namespace), the test harness keeps registering a stale, fake manifest and won't catch a real registration regression. It also bypasses azd's actual extension discovery/registration code, so that path is never exercised.

**Suggested approach** — use the developer extension (`microsoft.azd.extensions`, the `azd x` command set) to build, pack, and register into the local registry, all sourced from `extension.yaml`:

```yaml
- name: Install azd developer extension
  run: |
    azd extension source add -n azd -t url -l "https://aka.ms/azd/extensions/registry"  # if not already present
    azd extension install microsoft.azd.extensions

- name: Build, pack & publish extension to local registry (record-tagged)
  working-directory: cli/azd/extensions/azure.ai.agents
  env:
    GOFLAGS: -tags=record   # injected into build.sh's `go build`; no code change needed
  run: |
    azd x build    # build.sh inherits GOFLAGS -> go build gets -tags=record; copies binary into ~/.azd/extensions/<id>/
    azd x pack
    azd x publish  # registers the extension in the local registry from extension.yaml
```

**Why this works for the `record` tag (the obvious concern):** `azd x build` has no `--tags` flag, but it delegates to the extension's own `build.sh`/`build.ps1` with `cmd.Env = os.Environ()` inherited, and that script's `go build` line doesn't pin `-tags`. So Go's standard `GOFLAGS=-tags=record` env var injects the tag with zero code changes (verified locally: the same `build.sh`-style `go build -ldflags=...` line picks up the `record` tag purely from `GOFLAGS`). The `record` tag is therefore **not** a blocker for using `azd x build`.

**Benefits**
- All extension metadata comes from `extension.yaml` — no hardcoded dict, no drift.
- Exercises azd's real install/registration path, so registration bugs are actually caught.
- Removes the brittle YAML-embedded Python heredoc.

vhvb1989:
**Test location: current structure is fine as step 1; the direction is to move tests into the extension folder**

I dug into why these functional tests live in `cli/azd/test/functional/ai_agents/` (core module) instead of inside the extension folder. The reason holds up: the extension is a separate Go module (`azureaiagent`) pinned to a released `cli/azd`, and the reused test helpers (`test/azdcli`, `test/recording`) can't be imported from there today. So **keeping the tests where they are is the right first step** — no change requested here.

Findings (for context):

- **Internal-package barrier is tiny.** The entire `cli/azd/internal/...` closure of `azdcli` + `recording` is just `internal/runcontext/agentdetect` (+ root `internal`, pulled in only through it), and it all traces to one line: `agentdetect.DisableAgentDetectEnvVar` (`= "AZD_DISABLE_AGENT_DETECT"`) in `test/azdcli/cli.go:180`. Relocating that single const to a public package clears the whole internal blocker.
- **Version coupling is the real blocker.** From a separate module, the helpers would be consumed at a pinned `cli/azd` release, so harness changes (like this PR's `cli.go` token-format fix) wouldn't be available without a release + bump, or a forbidden blanket `replace` of the production `cli/azd` module.

**Direction (tracked in #8795):** expose a public, internal-free **test-support module** (its own `go.mod`) so extensions can host functional tests in their own folder and pin it **from source via a `replace` scoped to that test-only module** — leaving the production `cli/azd` pin intact (production code never imports it, so the source `replace` only affects `go test`, not released binaries).

No action needed in this PR for test location — just flagging the agreed direction so we migrate these into `cli/azd/extensions/azure.ai.agents/` once #8795 lands.

v1212:
@vhvb1989 Simplifying install is done — switched to `azd x build → pack → publish → extension install --source local`. All metadata now comes from `extension.yaml`, no more hardcoded config.json. `azure.ai.inspector` is pre-installed from the `azd` source since `installInternal` resolves deps from the parent's source. thanks for the call.

v1212:
@vhvb1989 Agreed to the further refactor on internal-package barrier and version-coupling constraints, no change needed here. Will migrate once #8795 lands.

jongio (commented):
Clean refactoring that addresses all prior feedback. Tests properly isolated in their own subdirectory and package, workflows split with correct permissions, CODEOWNERS added. LGTM.

vhvb1989 (approved):

trangevi (approved):

.github/workflows/test-ext-azure-ai-agents.yml
```diff
@@ -0,0 +1,73 @@
...
+      - uses: actions/checkout@v6
+
+      - uses: actions/setup-go@v6
+        with:
+          go-version-file: cli/azd/go.mod
+
+      - name: Build azd
+        working-directory: cli/azd
+        run: |
+          go build -tags=record -o azd-record ./
```
danieljurek, line 31: Is there support for a recording flag in ci-build.ps1?

.github/workflows/test-ext-azure-ai-agents.yml
```diff
@@ -0,0 +1,73 @@
+name: ext-azure-ai-agents-test
+
+on:
+  pull_request:
```
danieljurek, line 4: Should this also run on workflow_dispatch?

danieljurek (approved):
I think this is a good first cut at narrowly testing this single extension. If you want to test other extensions in this way, I recommend a few things: 

* Use one workflow to test all extensions (figure out which extension to test based on files in the diff) 
* Use ci-build.ps1 in each extension and add support for the tags/params you want to use in that build
* Use ci-test.ps1 if that makes sense in this case instead of calling go test directly... because these tests are running from outside the extension directory, perhaps that may not be applicable here.

.github/workflows/test-ext-azure-ai-agents.yml
vhvb1989: Good catch — `ci-build.ps1` already has a `-BuildRecordMode` switch that adds the `record` tag and emits a `-record` binary. Tracking the switch to that canonical script (instead of `azd x build` + `GOFLAGS=-tags=record`) as a follow-up in #8801 so we can land this PR. Resolving.

.github/workflows/test-ext-azure-ai-agents.yml
vhvb1989: Good idea — captured in #8801 to evaluate adding `workflow_dispatch` (and possibly a `schedule`) for manual/ad-hoc runs. Keeping the current `pull_request` trigger for now so this PR can merge. Resolving.

.github/workflows/test-ext-azure-ai-agents.yml
v1212: Thanks for the suggestion — switched to ci-build.ps1 -BuildRecordMode in #8814.

.github/workflows/test-ext-azure-ai-agents.yml
v1212: Added workflow_dispatch trigger in #8814.

PR #8757, opened 2026-06-22
docs(azure.ai.agents): add hosted-agent egress policy spec

hund030:
## Summary

Adds a design spec for supporting hosted-agent **egress policy** in the `azure.ai.agents` extension, driven from `azure.yaml`. It covers binding an agent to an `agentGuardrails` policy resource (data-plane) and provisioning that resource for greenfield projects (control-plane). Spec only — no code changes.

## Changes

- **`docs/specs/egress-policy.md`**: new spec. Declares an `egressPolicy` field on the `azure.yaml` hosted-agent block with two forms — `name` + rules (azd provisions the resource) and `id` (reference an existing one). Maps the resolved resource ID into the version definition as `egress_policy_id_preview` (gated by the `EgressPolicy=V1Preview` opt-in), and synthesizes the `agentGuardrails` ARM resource in Tier B.

### DIFF
--- a/cli/azd/extensions/azure.ai.agents/docs/specs/egress-policy.md
+++ b/cli/azd/extensions/azure.ai.agents/docs/specs/egress-policy.md
@@ -1,0 +1,406 @@
+# Spec: azd Support for Hosted-Agent Egress Policy
+
+- **Owner:** zhihuan
+- **Extension:** `azure.ai.agents`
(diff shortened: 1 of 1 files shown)


hund030:
Close this as we shift to the existing RAI policy resource. No azd changes need to be made.

PR #8779, opened 2026-06-23
feat: carry Foundry agent and resource definitions in azure.yaml

### LINKED ISSUE #8774: [ext-foundry]: make project/connection/toolbox Foundry resources real service targets
Implements the per-resource service-target design of unified `azure.yaml` spec #8590 (`docs/specs/unify-azure-yaml/spec.md`) §2.2, §2.6, §2.8, plus §1.4 (`endpoint:`).

#8675 writes each Foundry resource as its own `azure.yaml` service (`azure.ai.project` / `azure.ai.connection` / `azure.ai.toolbox`) but registers all three as **no-op** service targets inside `azure.ai.agents`; the resources are still provisioned by Bicep at `provision` time and re-sourced through the `pre/postprovision` + `postdeploy` hooks in `listen.go`. This issue moves each host to its **owning sibling extension** with a real deploy-time upsert.

## Summary

Turn `azure.ai.project`, `azure.ai.connection`, and `azure.ai.toolbox` into real service targets owned by their sibling extensions, so `azd deploy` reconciles each resource (upsert) in `uses:` order, with native per-service progress and failure attribution.

## Scope

Per host — each a vertical slice that: registers the host + service target, publishes `<kind>.json`, adds the core-schema conditional, removes the host from the agents no-op `ResourceServiceTargetProvider`, and moves its reconcile out of the `listen.go` hooks into the resource's `Deploy`:

- `azure.ai.project` -> `azure.ai.projects`: resolve the project (provisioned, or via `endpoint:` §1.4) and upsert each deployment.
- `azure.ai.connection` -> `azure.ai.connections`: upsert the connection; resolve env secrets.
- `azure.ai.toolbox` -> `azure.ai.toolboxes`: upsert the toolbox; resolve named connection references.

> **Note:** `azure.ai.skill` and `azure.ai.routine` real service targets are covered in #8779.

Ordering stays in `azure.yaml` via `uses:` (project -> connections -> toolboxes -> agents); azd core topologically orders them.

## Out of scope (follow-ups)

- Agent-host shape + legacy migration -> #8773.
- `$ref` file includes / overlay overrides -> #8775.
- Skill + routine real targets -> addressed in #8779.
- Built-in Bicep / provision-less project creation (separate RFC).

## Related

Follow-up to #8590 / #8675:
- #8773 — agent shape + migration
- **#8774 — project/connection/toolbox real service targets (this issue)**
- #8775 — `$ref` file includes

huimiu:
## Summary

`azure.yaml` becomes the single source of truth for a Foundry agent and its companion resources. The agent definition moves off disk into the agent's service entry, and every Foundry resource — the project and its model deployments, connections, toolboxes, skills, and routines — becomes a first-class service with a real `azd deploy` target. Existing projects keep working through legacy fallbacks that emit deprecation signals.

## What changed

- **Agent definition lives in `azure.yaml`.** Carried inline on the `azure.ai.agent` service entry instead of a separate on-disk `agent.yaml`. `init` stops writing that file; every agent command (deploy, run, listen, update, endpoint show) reads from the service entry. Projects on the old on-disk shape still load, with a deprecation warning and a telemetry signal.
- **Each Foundry resource is its own deploy target.** Model deployments, connections, toolboxes, skills, and routines become sibling `host: azure.ai.<kind>` entries, each owned by its extension. `azd deploy` now performs a real upsert per resource (previously placeholders); `init` writes the entries and wires their provision/deploy order.
- **Connect to an existing project (`endpoint:`).** When a Foundry service sets `endpoint:`, provisioning is skipped and azd connects to that project instead of creating one.
- **Routine inputs resolve `${VAR}`** from the azd environment at deploy time, while server-side `${{...}}` placeholders are preserved.
- **Schema.** The `azure.yaml` JSON schema (stable + alpha) gains per-host validation for all six Foundry hosts; the legacy single `microsoft.foundry` host stays accepted so existing projects keep validating.
- **Shared core helpers.** The `$ref` resolver, the `azure.yaml` edit helper, and the `${VAR}` expander move into core (`pkg/foundry`) so all Foundry extensions share one implementation.
- **Telemetry.** A new field flags projects still on the legacy agent shape; the telemetry docs and the coverage test move in lockstep.

## Before merge

The `connections`, `routines`, and `toolboxes` extensions temporarily `replace` the azd core dependency with the in-tree checkout to build against the new `pkg/foundry` helpers. Remove these directives once the core change lands and bump the published dependency.


cli/azd/extensions/azure.ai.agents/internal/cmd/resource_services.go
```diff
@@ -0,0 +1,368 @@
...
+	}
+
+	return nil
+}
+
+// sanitizeServiceName converts a resource name into a valid azure.yaml service
+// key by trimming and removing spaces, matching how the agent service name is
+// derived from the agent name.
+func sanitizeServiceName(name string) string {
+	return strings.ReplaceAll(strings.TrimSpace(name), " ", "")
```
jongio, line 205: `sanitizeServiceName` only strips spaces. Resource names with hyphens, dots, or other special characters pass through unchanged. If Foundry resource names can contain characters invalid in a YAML map key (e.g., colons, brackets), this could produce malformed azure.yaml. Worth documenting the allowed character set in the function doc comment, or adding a regex that rejects/escapes invalid chars.

cli/azd/extensions/azure.ai.skills/internal/cmd/service_target.go
```diff
@@ -0,0 +1,202 @@
...
+	return cfg, nil
+}
+
+func resolveSkillInstructions(svc *azdext.ServiceConfig, instructions string) (string, error) {
+	if !isInstructionFilePath(instructions) {
+		return instructions, nil
+	}
+
+	path := strings.TrimSpace(instructions)
+	if !filepath.IsAbs(path) {
```
jongio, line 180: Path traversal concern: `resolveSkillInstructions` joins `svc.GetRelativePath()` with a user-provided `instructions` value. A relative path like `../../sensitive.md` would escape the service directory. Consider validating that the resolved path stays within the project root (similar to how `paths.JoinAllowRoot` is used elsewhere in the agents extension).

cli/azd/extensions/azure.ai.agents/internal/project/agent_definition.go
```diff
@@ -0,0 +1,492 @@
...
+		)
+	}
+
+	ca := inline.toContainerAgent(cfg.Container, coreImage)
+
+	// Validate the inline definition with the same rules the on-disk agent.yaml
+	// path uses (kind, name format, policies), so an inline definition cannot
+	// silently bypass validation. Marshal back to YAML so ValidateAgentDefinition
+	// sees the same shape it expects from disk.
+	if defBytes, marshalErr := yaml.Marshal(ca); marshalErr == nil {
```
jongio: nit: If `yaml.Marshal(ca)` fails (`marshalErr != nil`), the validation step is silently skipped. This means a definition that can't round-trip to YAML passes without validation. Consider logging the marshal error at debug level so issues are visible during troubleshooting.

cli/azd/extensions/azure.ai.agents/internal/cmd/resource_services.go
```diff
@@ -0,0 +1,368 @@
...
+	agentUses = append(agentUses, projectServiceName)
+
+	// Connection and toolbox services depend on the project service so the
+	// project is provisioned first.
+	siblingUses := []string{projectServiceName}
+
+	for i := range connections {
+		conn := connections[i]
+		connName := sanitizeServiceName(conn.Name)
+		if connName == "" {
```
jongio, line 88: When `connName` is empty (after sanitization), the connection is silently skipped. This could be surprising if a user names a connection with only spaces. Consider emitting a warning or returning an error so the user knows their connection was dropped.

jongio (commented):
Solid architectural change. The decomposition of Foundry resources into individual azure.yaml service entries with `uses:` ordering is well-designed, and the fallback collectors (sibling service wins, bundled agent config is the legacy fallback) are cleanly implemented with good test coverage. A few observations below.

jongio (commented):
**Incremental review** (22 commits since last pass at `01463ff`). CI is green, overall architecture is clean. The new `pkg/foundry` package is well-designed: proper cycle detection, depth bounds, error typing, and good test coverage. The resource-service split across per-host extensions is consistent and well-factored.

**Prior unresolved items** (4 comments from my earlier review remain open with no replies):
- `sanitizeServiceName` only strips spaces (resource_services.go:193)
- Path traversal in `resolveSkillInstructions` (skills service_target.go:180)
- Silent yaml.Marshal skip (agent_definition.go:348)
- Silent empty connName skip (resource_services.go:87)

I won't re-post these, but they still apply to the current HEAD.

---

**New observations on the `pkg/foundry` package:**

1. **`ExpandEnv` sentinel restoration correctness** (templating.go:68-70): The sentinel is guaranteed absent from the *input*, but env var expansion runs *after* masking. If an env var resolves to a string containing the sentinel prefix (`azdFoundryTemplateSpan_0_`), `strings.Replace` will match it and inject a Foundry span where it doesn't belong. Extremely unlikely in practice, but the correctness invariant ("Foundry expressions are always returned byte-for-byte") is violated in that case. A simple fix: after expansion, verify each placeholder appears exactly once before replacing, or use a sentinel that encodes a random token per call.

2. **`looksLikeInstructionsPath` heuristic** (includes.go:262-268): Inline prose that ends with `.md` or `.txt` (e.g., "Follow the steps in README.md") will be incorrectly classified as a file path and rebased. The current check (single-line + suffix) is a reasonable heuristic, but consider also requiring a path separator or `./` prefix, or at least documenting the false-positive risk so authors of `$ref`'d files know to avoid ambiguous instructions values.

3. **Nil map-value guard** (resource_services.go:222-229, 255-261, 287-293): The `collectProjectDeployments`/`collectConnections`/`collectToolboxes` collectors iterate the services map and dereference each value's Host field. If a map ever contains a nil value (unlikely from azd core, but possible from test code or future refactors), this panics. A nil guard in `sortedServices` would be cheap insurance.

All three are low-to-medium severity. The overall approach is solid; the design of `ResolveFileRefs` + `YAMLDocument` as a read/write pair sharing path logic is thoughtful. Ship it once the prior items are addressed.

jongio (commented):
**Incremental review** (1 commit since last pass at `e59e938`). The fix commit `d6fd4da` is clean and well-scoped:

1. **Legacy host restored**: `microsoft.foundry` added to `FoundryServiceHosts` for backward compat. Tests cover mixed-host ambiguity detection. Schema `ref` correctly points to `/main/...microsoft.foundry.json`.

2. **Whitespace endpoint trimming**: `strings.TrimSpace` applied consistently in `foundryServiceEndpoint`, `warnNetworkIgnoredInBrownfield`, and the synthesizer's brownfield check. Blank endpoints now correctly trigger greenfield mode. Test added.

3. **Missing brownfield warning**: `warnNetworkIgnoredInBrownfield` was only called on the synthesizer path. Now also called on the on-disk Bicep path when `endpoint:` is set. Good catch.

No concerns. CI is green for the core pipelines; agents-public build is still pending (expected timing). LGTM on this increment.

jongio (commented):
Incremental review (5 new commits since d6fd4dac). All five address prior review feedback:

- Path traversal rejection in `resolveSkillInstructions` is correct: checks before `filepath.Join`, covers `\` and `/` separators, has test coverage for nested traversal (`sub/../../escape.md`).
- Warnings for empty sanitized names are user-facing on stderr, consistent with `WarnLegacyAgentShape`.
- Schema docs now accurately reflect that remote URLs aren't supported.
- `azure.ai.*/schemas/*.json` glob ensures cross-extension `` resolution works offline.
- Marshal failure logging makes the validation skip visible during troubleshooting.

No new issues found in this increment.

cli/azd/pkg/foundry/includes.go
```diff
@@ -1,7 +1,7 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
```
trangevi, line 1: Why are these files moved into the top level azd pkg?

cli/azd/internal/tracing/fields/fields.go
```diff
@@ -202,6 +202,15 @@ var (
 		Classification: SystemMetadata,
 		Purpose:        PerformanceAndHealth,
 	}
+	// Whether the project contains a Foundry agent service (host: azure.ai.agent)
+	// still using the deprecated config-nested shape (a populated `config:` block)
+	// instead of service-level properties. Tracks migration of older Foundry
+	// projects off the legacy shape.
+	FoundryAgentLegacyConfigKey = AttributeKey{
+		Key:            attribute.Key("foundry.agent.legacy_config_shape"),
```
trangevi, line 210: What actually sets this? Is this something the azd folks are alright with being added to telemetry?

cli/azd/extensions/azure.ai.agents/internal/project/agent_definition.go
```diff
@@ -0,0 +1,496 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
trangevi, line 1: What is the plan to stop supporting old formats? Adding this file now gives at least a third place where we have agent objects defined, making it all that much more difficult to make sure things are being updated in the right places. Would be good if we have an in place plan to switch completely and remove all of the extra logic

cli/azd/extensions/azure.ai.agents/schemas/Connection.json
```diff
@@ -1,6 +1,6 @@
 {
```
trangevi, line 1: Why do we need both these files, and the individual schemas in the other extensions?

cli/azd/pkg/foundry/includes.go
jongio: These files (the ref resolver, the azure.yaml edit helper, and the variable expander) were previously in `azure.ai.agents/internal/project/`. The problem: the connections, routines, and toolboxes extensions also need ref resolution and variable expansion for their own azure.yaml service entries. Keeping these in agents' internal package would force a circular dependency or require other extensions to import agents internals.

`pkg/foundry` is the shared-helpers layer accessible to all Foundry extensions without cross-extension coupling.

cli/azd/internal/tracing/fields/fields.go
jongio: It's set in the project telemetry collection (`cli/azd/cmd/actions.go` area) when `legacyAgentConfig` is detected during project loading. The flag fires when a project still uses the deprecated `config:` block on its agent service entry instead of the new service-level properties introduced in this PR.

Classification is `SystemMetadata` / `FeatureInsight`, which is the standard azd tier for tracking feature-adoption signals (same tier as `ProjectServiceHostsKey` and `ProjectServiceLanguagesKey`). It doesn't capture user content, just a boolean indicating the project is on the old shape.

cli/azd/extensions/azure.ai.agents/schemas/Connection.json
jongio: The agents extension bundles peer-extension schemas (Connection.json, Toolbox.json, etc.) so that `$` + `ref` resolution in agent definitions can validate those shapes offline without requiring the connections/toolboxes/skills extensions to be installed at validation time.

The individual extension schemas (e.g., in `azure.ai.connections/schemas/`) remain the canonical source. These copies in `azure.ai.agents/schemas/` are resolved via the `azure.ai.*/schemas/*.json` glob at build time, enabling cross-extension validation within a single extension's scope.

cli/azd/pkg/foundry/includes.go
huimiu: These helpers (the `$ref` resolver, the edit helper, and the `${VAR}` expander) used to live in agents'' `internal/` package. Connections, routines, and toolboxes now need them too, and Go does not allow importing another extension''s `internal/`. `pkg/foundry` gives them one shared home instead of duplicated copies.

cli/azd/internal/tracing/fields/fields.go
huimiu: Small correction: it is set in `pkg/project/project.go` during `Load()`, not `actions.go`. When a project loads with an `azure.ai.agent` service still using the old `config:` block, we tag a boolean. No user content, same tier as the existing `ProjectServiceHostsKey`.

cli/azd/extensions/azure.ai.agents/schemas/Connection.json
huimiu: They are two different shapes. `agents/schemas/Connection.json` is a nested item inside the legacy `microsoft.foundry` host (one entry composing arrays). `azure.ai.connections/.../azure.ai.connection.json` is the new `host: azure.ai.connection` service, where each connection is its own service. Both exist because this PR keeps the legacy host working alongside the new per resource hosts. They are not copies; the shapes differ.

cli/azd/extensions/azure.ai.agents/internal/project/agent_definition.go
huimiu: The upcoming release will continue to support the old formats. We can consider deprecating them in the following version

PR #8782, opened 2026-06-23
feat: ai rle v1

farhann1:
## Summary

Adds the preview `azd ai rle` extension for source-first RLE/OpenEnv environment workflows.

The extension supports initializing an OpenEnv-style sample, running it locally in Docker, deploying/registering it with the RLE control plane, and invoking a remote sandbox for interactive testing.

## Changes

- Added hidden preview command group gated by `AZD_AI_RLE_ENABLE=true`.
- Added `azd ai rle init [environment-name]` to copy the default OpenEnv echo sample and create `.azd-rle.json`.
- Added `azd ai rle run` for local Docker build/run, `/health` readiness, `/web` launch, OpenEnv shell, and optional `--watch`.
- Added `azd ai rle deploy` to build/push an ACR image and create/update the RLE environment.
- Added `azd ai rle invoke` to lease a remote sandbox, open the sandbox/proxy UI, start the OpenEnv shell, and release the sandbox on exit.
- Added minimal local state using `name`, `project`, `environmentId`, and `environmentVersion`.
- Added README guidance for local setup, environment variables, Dockerfile discovery, deploy image naming, and command usage.

## Notes

- This is an internal preview surface and commands are hidden unless `AZD_AI_RLE_ENABLE=true`.
- RLE control plane endpoint is configured with `RLE_ENDPOINT` and defaults to `http://localhost:5000`.
- Deploy uses `AZURE_CONTAINER_REGISTRY_ENDPOINT` and publishes images as `<registry>/<project-id>-<environment>:latest`.

## Validation

- `go test .\internal\cmd`
- `go build -o bin\azure-ai-rle-windows-amd64.exe .`
- `cspell lint 'extensions/**/*.go' 'extensions/**/*.md' --config ./.vscode/cspell.yaml --no-progress`
- `git diff --check`


jongio (commented):
## Review: `azure.ai.rle` extension (new)

Thanks for contributing a new extension! A few items to address before this is ready.

### Overview

This adds a new `azd ai rle` command group for managing RLE (Reinforcement Learning Environment) resources. The implementation is clean and follows the extension SDK patterns well. Tests pass and it builds cleanly.

### Key items

1. **Missing PR description.** A 2949-line new extension needs a description explaining what RLE is, the motivation, the commands it adds, and any design decisions. Right now the body is empty.

2. **Unreachable code.** Five files define commands (`create`, `list`, `show`, `sandbox`, `versions`) that are never registered in `root.go`. These add ~500 lines of untestable surface area. Either wire them up (if intended for this PR) or remove them and add in a follow-up when they're needed.

3. **Cross-platform line endings.** `scaffold.go:67` unconditionally converts `\n` to `\r\n`. Generated Dockerfiles and Python files will have Windows line endings on Linux/macOS, which can cause issues with container builds and shebangs.

4. **Unquoted paths in user-facing output.** `init.go:55` prints `cd %s` without quoting. Paths containing spaces will break the suggested command. This repo's conventions require quoting paths (see AGENTS.md).

5. **Bearer token over HTTP.** The client sends `RLE_BEARER_TOKEN` regardless of the endpoint scheme. Consider logging a warning (or refusing) when sending credentials over non-TLS connections to non-localhost endpoints.

See inline comments for specifics.

cli/azd/extensions/azure.ai.rle/internal/cmd/root.go
```diff
@@ -0,0 +1,41 @@
...
+	})
+
+	rootCmd.AddCommand(newDeployCommand())
+	rootCmd.AddCommand(newInitCommand())
+	rootCmd.AddCommand(newInvokeCommand())
+	rootCmd.AddCommand(newVersionCommand(&extCtx.OutputFormat))
+	rootCmd.AddCommand(newMetadataCommand(rootCmd))
+
+	return rootCmd
+}
```
jongio, line 52: Five command constructors (`newCreateCommand`, `newListCommand`, `newShowCommand`, `newSandboxCommand`, `newVersionsCommand`) are defined in this package but never called. If they're planned for a follow-up PR, consider removing them from this PR to keep the diff focused. If they should be available now, wire them in here.

cli/azd/extensions/azure.ai.rle/internal/cmd/scaffold.go
```diff
@@ -0,0 +1,525 @@
...
+		"__ENV_TITLE__": envTitle,
+		"__IMAGE__":     image,
+	}
+
+	for relativePath, content := range files {
+		for token, value := range replacements {
+			content = strings.ReplaceAll(content, token, value)
+		}
+		content = strings.ReplaceAll(content, "\n", "\r\n")
+		fullPath := filepath.Join(sessionDir, filepath.FromSlash(relativePath))
```
jongio: This forces Windows line endings on all platforms. Generated Dockerfiles and Python files will have `\r\n` on Linux/macOS, which can cause container build failures (e.g., shebang parsing) and git noise.

Remove this line entirely. Let files keep their native `\n` endings. Git handles CRLF conversion on Windows if `.gitattributes` is configured.

cli/azd/extensions/azure.ai.rle/internal/cmd/init.go
```diff
@@ -0,0 +1,66 @@
...
+			}
+
+			displayDir, err := filepath.Abs(sessionDir)
+			if err != nil {
+				return err
+			}
+			_, err = fmt.Fprintf(
+				cmd.OutOrStdout(),
+				"Created OpenEnv-style environment at: %s\nNext steps:\n  cd %s\n  azd ai rle deploy\n",
+				displayDir,
```
jongio: Paths with spaces will break the suggested `cd` command. This repo requires quoting user-facing paths (AGENTS.md convention). Use `%q` instead of `%s` for the `cd` target.

cli/azd/extensions/azure.ai.rle/internal/cmd/client.go
```diff
@@ -0,0 +1,430 @@
...
+
+	req, err := http.NewRequestWithContext(ctx, method, c.baseUrl+path, reader)
+	if err != nil {
+		return fmt.Errorf("create request: %w", err)
+	}
+	req.Header.Set("Accept", "application/json")
+	if body != nil {
+		req.Header.Set("Content-Type", "application/json")
+	}
+	if token := os.Getenv("RLE_BEARER_TOKEN"); token != "" {
```
jongio: Consider logging a warning (to stderr) when sending a bearer token over a non-TLS, non-localhost endpoint. Silent credential leakage over HTTP to a remote host is a security concern.

cli/azd/extensions/azure.ai.rle/internal/cmd/deploy.go
```diff
@@ -0,0 +1,230 @@
...
+		first := image[:slash]
+		if first == "localhost" || strings.ContainsAny(first, ".:") {
+			return first, image[slash+1:]
+		}
+	}
+	return "", image
+}
+
+// normalizeRegistryLoginServer turns a registry short name ("devrle") into a login server
+// ("devrle.azurecr.io"). A value that already contains a '.' is assumed to be a login server.
```
jongio: These pure helper functions (splitImageHost, normalizeRegistryLoginServer, registryShortName, fileExists) plus isNotFoundError in client.go are great candidates for table-driven unit tests. They have clear input/output contracts and no side effects (except fileExists). Consider adding coverage in a follow-up, especially for edge cases like an image with no tag, an empty registry string, or a login server without a dot.

cli/azd/extensions/azure.ai.rle/internal/cmd/deploy.go
```diff
@@ -0,0 +1,230 @@
...
+	"github.com/azure/azure-dev/cli/azd/pkg/azdext"
+	"github.com/spf13/cobra"
+)
+
+func newDeployCommand() *cobra.Command {
+	flags := struct {
+		registry  string
+		skipBuild bool
+	}{}
+
```
jongio, line 30: [nit] The default registry value 'devrle' is clearly a dev-team convenience. That is fine for now, but document it in the --registry help string or README so users outside the team know they need to override it.

cli/azd/extensions/azure.ai.rle/internal/cmd/deploy.go
```diff
@@ -0,0 +1,230 @@
...
+					return err
+				}
+				build := exec.CommandContext(cmd.Context(), "az", "acr", "build", "--registry", registryName, "--image", repoTag, ".")
+				build.Stdout = cmd.OutOrStdout()
+				build.Stderr = cmd.ErrOrStderr()
+				if err := build.Run(); err != nil {
+					return &azdext.LocalError{
+						Message:    fmt.Sprintf("Failed to build and push image '%s' to registry '%s': %v", repoTag, registryName, err),
+						Code:       "rle_acr_build_failed",
+						Category:   azdext.LocalErrorCategoryUser,
```
jongio: The 404-fallback-to-create is a nice UX touch. One consideration: if state.Project was accidentally changed (or the state file was corrupted), this silently creates a new environment rather than surfacing the mismatch. The printed message helps, but you might want to log the old environment ID in the output so users can spot unintentional recreation.

jongio (commented):
### Incremental review (commit d621e521)

New deploy enhancements look well-structured. The splitImageHost/normalizeRegistryLoginServer composition is clean, and using errors.AsType for the 404 check is idiomatic Go 1.26. A few items:

**Previous review items** (5 findings from my earlier pass) are still unresolved: unreachable command constructors, cross-platform line endings in scaffold.go, unquoted paths in init.go, bearer token over HTTP. Please address those alongside the new work.

jongio (changes requested):
### Incremental review (commits 9dc600d, 6772f58)

The `invoke` implementation is taking shape. The approach of fetching Loom via shallow clone plus bundled wheel is reasonable for a dev workflow. A few items need attention before this can merge.

**Prior review items**: 3 of 5 findings from my earlier passes remain unresolved (unreachable command constructors in root.go, cross-platform CRLF in scaffold.go, unquoted path in init.go). The bearer-token item was addressed by removal.

**New findings (this commit):**

1. **Binary wheel without provenance** (`assets/rle_sdk-0.1.3-py3-none-any.whl`): Committing a pre-built `.whl` makes the package un-auditable. Add a README or comment documenting where this wheel is built, what source repo/tag it corresponds to, and how a reviewer can reproduce it. Alternatively, reference a published package feed so CI can fetch it at build time instead of vendoring the binary.

2. **Deploy ignores stored endpoint** (`deploy.go`): `resolveControlPlaneEndpoint("")` replaced `resolveControlPlaneEndpoint(state.Endpoint)`. If a user had an endpoint saved in `.azd-rle.json` from the manifest, it's now silently ignored. Was this intentional? If so, note it in the changelog.

3. **Fragile TOML patching** (`invoke.go:ensureLoomLocalSources`): String-contains checks + append-to-EOF doesn't guarantee the new lines land inside the `[tool.uv.sources]` section when one already exists elsewhere in the file. Consider using a TOML library or at minimum insert after the existing header line rather than at EOF.

4. **Most training flags are not exposed**: Only 4 of ~15 parameters have CLI flags. The rest can't be overridden without editing source. Expose them or document why they're fixed.

5. **ADO clone URL may require auth**: `defaultLoomRecipeRepo` points to `msdata.visualstudio.com`. Will this work without credentials for users outside the org? If auth is needed, surface a clear error message.

cli/azd/extensions/azure.ai.rle/internal/cmd/scaffold.go
```diff
@@ -0,0 +1,533 @@
...
+		"__ENV_CLASS__": envClass,
+		"__ENV_TITLE__": envTitle,
+		"__IMAGE__":     image,
+	}
+
+	for relativePath, content := range files {
+		for token, value := range replacements {
+			content = strings.ReplaceAll(content, token, value)
+		}
+		content = strings.ReplaceAll(content, "\n", "\r\n")
```
jongio: Still present from prior review: this forces `\r\n` on every generated file regardless of host OS. Dockerfiles and Python scripts with `\r` bytes can break shebang parsing and cause `exec format error` inside Linux containers. Remove this line entirely.

cli/azd/extensions/azure.ai.rle/internal/cmd/init.go
```diff
@@ -0,0 +1,70 @@
...
+				return err
+			}
+
+			displayDir, err := filepath.Abs(sessionDir)
+			if err != nil {
+				return err
+			}
+			_, err = fmt.Fprintf(
+				cmd.OutOrStdout(),
+				"Created OpenEnv-style environment at: %s\nNext steps:\n  cd %s\n  azd ai rle deploy\n",
```
jongio: Still present from prior review: the `cd` path isn't quoted. A path with spaces (e.g. `C:\Users\My Name\code_rl`) will break the suggested command. Use `%q` instead of `%s` for the `cd` target.

jongio (changes requested):
The new commit (a95e826) reworks image resolution to derive per-environment image names from the environment name, with `--image` as an explicit override on both `init` and `deploy`. The self-bootstrapping `deploy` (works without prior `init` when `rle.yaml` exists) is a nice UX improvement. The `errors.AsType` usage and `azdext.LocalError` error handling are correct and idiomatic.

Three items from my earlier reviews remain unresolved:

1. **scaffold.go:74** `strings.ReplaceAll(content, "\n", "\r\n")` forces Windows line endings on all platforms. Generated Dockerfiles and Python files will carry `\r\n` into Linux containers, which can break shebang parsing and cause git diff noise on non-Windows hosts. Remove this line; let files keep native `\n` endings.

2. **init.go:57** `cd %s` should use `%q` to quote paths that may contain spaces (per the repo's path safety convention).

3. **root.go** Five command source files (`create.go`, `list.go`, `show.go`, `sandbox.go`, `versions.go`) are included in the PR but their constructors are never registered in `NewRootCommand`. If planned for a follow-up, remove them from this PR to reduce the diff.

cli/azd/extensions/azure.ai.rle/internal/cmd/deploy.go
```diff
@@ -0,0 +1,291 @@
...
+					"No Dockerfile in current directory; skipping build and using image '%s'.\n", image); err != nil {
+					return err
+				}
+			default:
+				if _, err := fmt.Fprintf(cmd.OutOrStdout(),
+					"Building and pushing '%s' to registry '%s' (az acr build) ...\n",
+					repoTag, registryName); err != nil {
+					return err
+				}
+				build := exec.CommandContext(cmd.Context(),
```
jongio: Add `exec.LookPath("az")` before this block, matching the `git` and `uv` checks in `invoke.go`. Without it, a missing Azure CLI produces an error about "push access to the registry" when the real problem is that `az` isn't on PATH.

jongio (dismissed):
New commit (f3ab2d2) restores deploy changes lost in the merge conflict. The ACR build integration and image host-qualification helpers work correctly for the described use case.

One new finding: `az acr build` is called without verifying that `az` is installed (see inline comment). The `invoke` command checks for `git` and `uv` with `exec.LookPath` before calling them; the same pattern should apply here.

Three items from earlier reviews remain unresolved:

1. **scaffold.go:74** `strings.ReplaceAll(content, "\\n", "\\r\\n")` forces CRLF on all platforms. Dockerfiles and Python files with `\r\n` will break shebang parsing in Linux containers. Remove this line.

2. **init.go:57** `cd %s` needs `%q` to quote paths with spaces (AGENTS.md convention).

3. **root.go** Five command source files (`create.go`, `list.go`, `show.go`, `sandbox.go`, `versions.go`) define constructors never called from `NewRootCommand`. Wire them in or remove them from this PR.

jongio (commented):
Checking in on this PR. My CHANGES_REQUESTED from the previous review cycle is still active. The three core items remain unaddressed, and there's been no author response in the thread.

**Still blocking (from prior reviews):**

1. `scaffold.go:74` - forces Windows line endings on all generated files unconditionally. Dockerfiles and Python scripts with CRLF will break shebang parsing and cause issues in Linux containers. Remove the `strings.ReplaceAll(content, "\\n", "\\r\\n")` call.

2. `init.go:57` - the `cd` path in user-facing output isn't quoted. Paths with spaces will break the suggested command. Use `%q` instead of `%s` (explicit repo convention in AGENTS.md).

3. `root.go` - five command source files (create.go, list.go, show.go, sandbox.go, versions.go) define constructors that are never registered. The test in root_test.go explicitly asserts they are NOT registered, confirming this is intentional dead code. Remove them from this PR and add them when they are ready to ship.

4. `deploy.go` - `az acr build` is invoked without verifying `az` is on PATH. The invoke command checks for `git` and `uv` before calling them; apply the same pattern here.

**Additional item (see inline):**

5. `invoke.go` (`ensureLoomLocalSources`) - the TOML patching approach appends lines to EOF rather than inserting under the existing `[tool.uv.sources]` header. If other TOML sections follow, the appended keys land in the wrong section.

@farhann1 - could you acknowledge these items or let me know if there's a reason they can't be addressed? Happy to discuss tradeoffs.

cli/azd/extensions/azure.ai.rle/internal/cmd/invoke.go
```diff
@@ -0,0 +1,329 @@
...
+		if err != nil {
+			return err
+		}
+		lines = append(lines, fmt.Sprintf("rle-sdk = { path = %q }", filepath.ToSlash(relativeRleSdkPackage)))
+	}
+	if len(lines) == 0 {
+		return nil
+	}
+
+	contents = strings.TrimRight(contents, "\r\n") + "\n\n" + strings.Join(lines, "\n") + "\n"
```
jongio: This appends new entries to the end of the file, but if [tool.uv.sources] already exists earlier in the TOML and another section (e.g. [tool.uv] or [build-system]) appears after it, these lines will land under the wrong section. Consider inserting after the existing [tool.uv.sources] line when it is already present, or use a TOML library for structured edits.

cli/azd/extensions/azure.ai.rle/internal/cmd/scaffold.go
```diff
@@ -0,0 +1,525 @@
...
+		"__ENV_TITLE__":   envTitle,
+		"__LOCAL_IMAGE__": localImage,
+		"__IMAGE__":       registrationImage,
+	}
+
+	for relativePath, content := range files {
+		for token, value := range replacements {
+			content = strings.ReplaceAll(content, token, value)
+		}
+		content = strings.ReplaceAll(content, "\n", "\r\n")
```
jongio: Still present: this converts all generated files to Windows line endings unconditionally. On Linux/macOS, the resulting Dockerfiles and `.py` scripts carry `\r\n` into containers, which breaks `#!/usr/bin/env python3` shebang parsing and causes git diff noise.

Fix: remove this line entirely. The template strings already use `\n`, which is correct everywhere.

```suggestion
```

cli/azd/extensions/azure.ai.rle/internal/cmd/init.go
```diff
@@ -0,0 +1,153 @@
...
+
+			displayDir, err := filepath.Abs(sessionDir)
+			if err != nil {
+				return err
+			}
+			_, err = fmt.Fprintf(
+				cmd.OutOrStdout(),
+				"Created OpenEnv-style environment at: %s\n"+
+					"Next steps:\n"+
+					"  cd %s\n"+
```
jongio: Unquoted path. If `displayDir` contains a space (e.g. `C:\Users\My Name\code\echo_env`), the suggested `cd` command breaks.

```suggestion
				"  cd %q\n"+
```

jongio (commented):
Incremental review of commit 14de101 ("fix: v1 draft").

This commit makes substantial progress. The dead-code command files are removed, the vendored .whl is gone, the Loom/uv/TOML machinery is replaced by a cleaner Docker-based invoke, and deploy no longer shells out to `az`. Nice cleanup.

Two items from my prior reviews remain unresolved:

1. **scaffold.go:95** forces CRLF on every generated file regardless of host OS. Dockerfiles and Python scripts carrying `
` into Linux containers will break shebang parsing and cause git noise.

2. **init.go:104** prints an unquoted `cd` path. Paths with spaces will produce a broken command. Use `%q` per repo convention (AGENTS.md).

Everything else from my earlier feedback has been addressed.

jongio (commented):
Incremental review (commits 3a57bd3, 29dd67c).

The SPEC.md is a solid addition; it documents the manifest contract, CLI UX, state management, and open design questions clearly. The README cleanup removing the z dependency aligns with the earlier deploy rework. And the dockerPullFailureSuggestion simplification is correct since the extension no longer shells out to z acr login.

Two items from my prior reviews remain unresolved:

1. **scaffold.go:95** strings.ReplaceAll(content, "\n", "\r\n") forces Windows line endings on all generated files. Dockerfiles and Python scripts carrying \r\n into Linux containers will break shebang parsing. Remove this line; let files keep native endings.

2. **init.go:104** cd %s should use %q to quote the path. Paths with spaces will produce a broken suggestion (per AGENTS.md path-quoting convention).

Everything else from earlier feedback has been addressed. Once these two are fixed, this is ready to approve.

cli/azd/extensions/azure.ai.rle/internal/cmd/build.go
```diff
@@ -0,0 +1,131 @@
...
+	source = strings.TrimSpace(opts.source)
+	if source == "" {
+		source = "."
+	}
+	source, err = filepath.Abs(source)
+	if err != nil {
+		return "", "", nil, err
+	}
+
+	dockerfile = ""
```
jongio: Nit: `dockerfile = ""` on line 36 is immediately overwritten by `dockerfile = strings.TrimSpace(opts.dockerfile)` on line 37. The first assignment is dead code.

cli/azd/extensions/azure.ai.rle/internal/cmd/build.go
```diff
@@ -0,0 +1,131 @@
...
+	"github.com/spf13/cobra"
+)
+
+type dockerBuildOptions struct {
+	source     string
+	dockerfile string
+}
+
+func resolveDockerBuild(opts dockerBuildOptions) (source string, dockerfile string, err error) {
+	source, dockerfile, _, err = prepareDockerBuild(opts)
```
jongio: `resolveDockerBuild` is defined but never called anywhere in the package. If it is intended for future use, consider removing it until needed to keep the surface minimal.

cli/azd/extensions/azure.ai.rle/internal/cmd/invoke.go
```diff
@@ -0,0 +1,1139 @@
...
+			"Stop the container: docker rm -f <container>\n"+
+			"Then rerun: azd ai rle run --port %d\n"+
+			"If Docker does not show a container, check the process with: netstat -ano | findstr :%d",
+		port,
+		port,
+		port,
+	)
+}
+
+func localContainerAlreadyRunningError(_ string, port int) error {
```
jongio: `localContainerAlreadyRunningError` is defined but never called. Looks like leftover from a prior iteration where port-conflict detection was separate from `ensurePortAvailable`.

jongio (approved):
Incremental review of commit c059208 (fix: docker based builds).

Both items from my prior review are fixed: the forced \r\n line endings in scaffold.go are gone (file was rewritten around source-based builds), and init.go now uses %q for the cd path suggestion. The manifest-based workflow (rle.yaml, manifest.go) was removed in favor of direct Docker builds from local source, which is a cleaner model.

The new uild.go extraction, --dockerfile flag on run/deploy, --watch mode, and the remote invoke sandbox lifecycle all look correct. Tests cover the new paths well.

Minor nits (non-blocking):

jongio (approved):
Checked the 2 new commits since my last review (cli optimizations + cross-platform path fixes). Clean improvements: dead code removal, better error messages with endpoint visibility, max retry cap on sandbox lease, addEventListener over inline onclick, relative paths in init output, and deploy image naming now includes project prefix to prevent collisions. Tests updated appropriately. Looks good.

farhann1:
> @farhann1 please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> @microsoft-github-policy-service agree [company="{your company}"]
> ```
> 
> > Options:
> > 
> > * (default - no company specified) I have sole ownership of intellectual property rights to my Submissions and I am not making Submissions in the course of work for my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > * (when company given) I am making Submissions in the course of work for my employer (or my employer has intellectual property rights in my Submissions by contract or applicable law). I have permission from my employer to make Submissions and enter into this Agreement on behalf of my employer. By signing below, the defined term “You” includes me and my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree company="Microsoft"
> > ```
> 
> Contributor License Agreement

@microsoft-github-policy-service agree company="Microsoft"

jongio (approved):
Incremental review of 1 new commit (e8cfe20 - README update). Editorial improvements only: simplified wording, reorganized .azd-rle.json docs. No code changes. Prior approval stands.

Minor nit: file is missing a trailing newline, but not blocking.

trangevi (changes requested):
Leaving a blocking comment for now, pending review of both the code and broader spec review

RickWinter (commented):
This adds the preview `azd ai rle` extension: an OpenEnv-style init/run/deploy/invoke flow gated behind `AZD_AI_RLE_ENABLE`. The overall structure is sound. The command tree follows the extension SDK conventions, path handling in `build.go`/`scaffold.go` is careful (Dockerfile paths are validated against the source root, symlinks and `.git` are skipped on copy), and the sandbox lifecycle in `invoke.go` is well built. In particular `releaseRemoteSandbox` deliberately uses a fresh `context.Background()` with its own timeout, so the deferred release still runs when the user exits the shell with Ctrl-C and the request context is already cancelled. That is the right call and avoids a sandbox leak.

The one item to resolve before merge is in `client.go`: `isLocalControlPlaneEndpoint` is defined but never called. With `unused` enabled in the extension's `.golangci.yaml`, this should fail the lint gate, which means the "golangci-lint 0 issues" run in the description did not cover this path. It also reads like a dropped guard, since the control plane client sends no auth and `RLE_ENDPOINT` can point at a remote host. See the inline note for the specifics and the two ways to resolve it.

No other blockers. Given the dead code should break the `unused` linter, you may prefer to submit this as REQUEST_CHANGES rather than a comment; I have left the draft as COMMENT so you can decide. Everything else can follow as normal preview iteration.

cli/azd/extensions/azure.ai.rle/internal/cmd/client.go
```diff
@@ -0,0 +1,250 @@
...
+}
+
+func resolveControlPlaneEndpoint() string {
+	if endpoint := os.Getenv("RLE_ENDPOINT"); endpoint != "" {
+		return endpoint
+	}
+	return defaultControlPlaneEndpoint
+}
+
+func isLocalControlPlaneEndpoint(endpoint string) bool {
```
RickWinter: `isLocalControlPlaneEndpoint` is defined but never called anywhere in the extension. With `unused` enabled in `.golangci.yaml`, staticcheck should flag this as U1000, so I do not think the "golangci-lint 0 issues" run in the description actually covered this file. Beyond the lint gate, this helper reads like a guard that got dropped: `do()` sends every control plane request with no auth header, and `resolveControlPlaneEndpoint` will happily use a non-local `RLE_ENDPOINT`, so today we would send project ids and image paths in cleartext to an arbitrary remote host. Was the intent to require a local endpoint (or add auth) before talking to a remote one? Either wire this in where it was meant to go, or remove it so the dead code does not fail the build.

cli/azd/extensions/azure.ai.rle/build.ps1
```diff
@@ -0,0 +1,78 @@
...
+    New-Item -ItemType Directory -Path $OUTPUT_DIR | Out-Null
+}
+
+# Get Git commit hash and build date
+$COMMIT = git rev-parse HEAD
+if ($LASTEXITCODE -ne 0) {
+    Write-Host "Error: Failed to get git commit hash"
+    exit 1
+}
+$BUILD_DATE = (Get-Date -Format "yyyy-MM-ddTHH:mm:ssZ")
```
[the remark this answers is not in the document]
farhann1: updated

cli/azd/extensions/azure.ai.rle/internal/cmd/root.go
farhann1: these are the only commands declared now, and all of them are in use:

newDeployCommand(),
newInitCommand(),
newInvokeCommand(),
newRunCommand(),
newVersionCommand

cli/azd/extensions/azure.ai.rle/internal/cmd/client.go
farhann1: removed the dead code..

local endpoint is just a backup for now.. we are in the process of setting up the control plane url, once that is done, we will update on extension side as well, along with validation.

due to above concern, we are hiding all commands from discovery using the env flag.. and once the backend URL is baked in, with validation in-place.. we will show discovery for all commands.

jongio (approved):
New commit (eb14e2d) looks good. UTC time fix in build.ps1 is correct, and the dead code removal in client.go is clean (no orphaned imports). Prior approval stands.

cli/azd/extensions/azure.ai.rle/internal/project/scaffold.go
```diff
@@ -0,0 +1,182 @@
...
+			builder.WriteRune('-')
+			lastDash = true
+		}
+	}
+	return strings.Trim(builder.String(), "-")
+}
+
+const (
+	openEnvRepoUrl        = "https://github.com/huggingface/OpenEnv.git"
+	openEnvRepoRef        = "main"
```
RickWinter, line 17: Pinning to main means azd ai rle init pulls whatever is on huggingface/OpenEnv main at the moment it runs, so two inits days apart can scaffold different code, and that code then gets built into the runtime image and pushed to the user's ACR on deploy. That is both a reproducibility and a supply-chain concern. Can we pin openEnvRepoRef to a released tag or a specific commit SHA, and sparse-checkout that ref, so init is deterministic?

cli/azd/extensions/azure.ai.rle/internal/cmd/client.go
```diff
@@ -0,0 +1,241 @@
...
+			return fmt.Errorf("marshal request body: %w", err)
+		}
+		reader = bytes.NewReader(data)
+	}
+
+	req, err := http.NewRequestWithContext(ctx, method, c.baseUrl+path, reader)
+	if err != nil {
+		return fmt.Errorf("create request: %w", err)
+	}
+	req.Header.Set("Accept", "application/json")
```
RickWinter, line 214: The control-plane client only sets Accept and Content-Type, with no auth. That is fine for the default http://localhost:5000, but resolveControlPlaneEndpoint() lets RLE_ENDPOINT point anywhere, including a remote https control plane, and we would call it unauthenticated. Is a remote RLE_ENDPOINT in scope for this preview? If so we likely need a token or credential here; if not, it is worth documenting that RLE_ENDPOINT is loopback-only for now.

cli/azd/extensions/azure.ai.rle/internal/cmd/banner.go
```diff
@@ -0,0 +1,30 @@
...
+
+import (
+	"fmt"
+	"io"
+	"strings"
+
+	"github.com/fatih/color"
+)
+
+const bannerArt = `███████╗ ██████╗ ██╗   ██╗███╗   ██╗██████╗ ██████╗ ██╗   ██╗
```
RickWinter, line 14: nit: the ASCII art here spells FOUNDRY, so it looks carried over from the Foundry extension banner. For the rle extension it is a little surprising to see FOUNDRY print on azd ai rle --help.

cli/azd/extensions/azure.ai.rle/internal/cmd/invoke.go
```diff
@@ -0,0 +1,1138 @@
...
+	dockerfile   string
+	watch        bool
+	restart      bool
+	reuseRunning bool
+}
+
+func newInvokeCommand() *cobra.Command {
+	flags := &openEnvInvokeFlags{
+		timeout:      30,
+		reuseRunning: true,
```
RickWinter: nit: openEnvInvokeFlags is shared with the run command, but newInvokeCommand only wires --timeout, so port, source, dockerfile, watch, restart, and this reuseRunning default are all dead on the invoke path. Splitting invoke's flags into their own small struct would make each command's surface match what it actually uses.

RickWinter (commented):
This adds the preview azd ai rle extension, hidden behind AZD_AI_RLE_ENABLE, for source-first OpenEnv/RLE workflows: init copies a sample into a session folder, run builds and runs it locally in Docker, deploy pushes an image to ACR and creates or updates the environment on the control plane, and invoke leases a remote sandbox and opens an OpenEnv shell plus a local proxy UI. The overall shape is right: it is self-contained under extensions/, the LocalError and path-safety helpers (isSafeRelativePath, isPathWithinRoot) are consistent with the framework, and the command surface is gated correctly.

A few things worth resolving. The init clone pins to the upstream main branch of a third-party repo (huggingface/OpenEnv), so init is non-reproducible and that code is then built into the runtime image and pushed to the user's ACR on deploy; pinning to a tag or commit SHA would remove both the reproducibility gap and the supply-chain surface. Separately, the control-plane client sends no authentication, which is fine for the http://localhost:5000 default but is worth a decision for a remote RLE_ENDPOINT.

Two nits that are correct to leave for a preview: the banner art spells FOUNDRY, and openEnvInvokeFlags carries several fields the invoke command never wires.

None of this blocks a preview surface. The unpinned clone is the one I would address before merge; the rest can follow.

cli/azd/extensions/azure.ai.rle/internal/cmd/banner.go
farhann1: rle will be part of foundry offerings, so the banner is intentional.

cli/azd/extensions/azure.ai.rle/internal/cmd/client.go
farhann1: for now, we are using local and a remote endpoint with no auth for testing.. there will a follow-up PR, once the actual APIM URLs are ready for integration.

cli/azd/extensions/azure.ai.rle/internal/project/scaffold.go
farhann1: it's just sample code, and we don't need same consistency on all checkouts.

cli/azd/extensions/azure.ai.rle/internal/cmd/invoke.go
farhann1: fixed

cli/azd/extensions/azure.ai.rle/internal/cmd/deploy.go
```diff
@@ -0,0 +1,209 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
trangevi, line 1: As we discussed, this verb should be changed. If it stays for private preview I think that's fine, but will need to change before PuPr.

cli/azd/extensions/azure.ai.rle/internal/project/build.go
```diff
@@ -0,0 +1,125 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
trangevi, line 1: NIT: From my experience, it's easy for the files in these extensions to get out of hand quickly. I'd suggest that this file end up in a different directory since it's not a user facing command.

In the agents extension, we had 'pkg' as a directory containing code that interacts with the cloud (like client objects), and 'project' as a directory containing code more directly responsible for manipulating the local code/project.

That hasn't been a hard and fast rule (we definitely have stuff kind of scattered everywhere in the agents extension), and I want to get some time to reorganize a bit, but throwing out the suggestion regardless, for this and any other files in here which aren't specifically commands

cli/azd/extensions/azure.ai.rle/internal/cmd/deploy.go
```diff
@@ -0,0 +1,209 @@
...
+
+func newDeployCommand() *cobra.Command {
+	flags := &rleDeployFlags{}
+
+	cmd := &cobra.Command{
+		Use:   "deploy",
+		Short: "Create or update the RLE environment",
+		Args:  cobra.NoArgs,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			state, initialized, err := resolveDeployState(flags)
```
trangevi: To avoid a potentially ever expanding closure in RunE, I suggest having the anonymous function be purely responsible for command parameter validation, and then to construct a <Foo>Action object, passing the flags to it, and calling something like <Foo>Action.Run or similar, and having that Run method contain further logic. It just helps keep the `newDeployCommand` and other similar functions cleaner, and allows for a clear separation between processing the command input and acting on it.

Same comment for other commands as applicable.

cli/azd/extensions/azure.ai.rle/internal/cmd/invoke.go
```diff
@@ -0,0 +1,1157 @@
...
+	defer resp.Body.Close()
+
+	if contentType := resp.Header.Get("Content-Type"); contentType != "" {
+		w.Header().Set("Content-Type", contentType)
+	}
+	w.WriteHeader(resp.StatusCode)
+	_, _ = io.Copy(w, resp.Body)
+}
+
+const remotePlaygroundHTML = `<!doctype html>
```
trangevi: I would suggest a separate file that is read in when needed, makes it easier to validate that the content is formatted correctly

cli/azd/extensions/azure.ai.rle/internal/cmd/invoke.go
```diff
@@ -0,0 +1,1157 @@
...
+	select {
+	case err := <-done:
+		return err
+	case <-ctx.Done():
+		fmt.Fprintln(output)
+		return nil
+	}
+}
+
+func newRunCommand() *cobra.Command {
```
trangevi: This should be in a separate file, to make it easier to find relevant code later.

trangevi (approved):
Generally I approve the code for a PrPr

cli/azd/extensions/azure.ai.rle/extension.yaml
JeffreyCA, line 1: Some other files that should be added:

- `ci-build.ps1`
- `ci-test.ps1`
- `.github/workflows/lint-ext-azure-ai-rle.yml`
- `eng/pipelines/release-ext-azure-ai-rle.yml` - the ADO release pipeline definition
- `.github/CODEOWNERS` entry

JeffreyCA (commented):
In the future, you can use `azd x init --internal` to automatically scaffold all the required 1P specific extension files

jongio (approved):
Incremental review of bab9130 (fix: address review comments).

Flag type split into 
emoteInvokeFlags, localRunFlags, and openEnvCallFlags is clean. The new project_endpoint.go module validates and normalizes the Foundry endpoint correctly, with good test coverage for both happy path and error cases. State migration from Project to ProjectEndpoint is consistent across deploy and invoke.

No new issues.

JeffreyCA:
In terms of general housekeeping we may want to create a separate GitHub label like `ext-rle` or similar @trangevi

cli/azd/extensions/azure.ai.rle/internal/cmd/deploy.go
farhann1: yes, will update it before PuPr

cli/azd/extensions/azure.ai.rle/internal/project/build.go
farhann1: moved out of cmd, and placed under internal/project/build.go

cli/azd/extensions/azure.ai.rle/internal/cmd/deploy.go
farhann1: updated

cli/azd/extensions/azure.ai.rle/internal/cmd/invoke.go
farhann1: moved to internal\ui

cli/azd/extensions/azure.ai.rle/internal/cmd/invoke.go
farhann1: done

cli/azd/extensions/azure.ai.rle/extension.yaml
farhann1: added

JeffreyCA: Sorry, I still don't see them in the PR?

farhann1: sorry, i had the changes.. but forgot to commit. 

just did, please check now.

jongio (approved):
Incremental review of f5798ac (addressing prior feedback). The refactoring looks good: local container lifecycle cleanly separated into `run.go`, utility code extracted to focused modules (`docker.go`, `runtime.go`, `identifiers.go`), and tests are thorough. CODEOWNERS and CI added. One note: `actionlint` CI is failing on the new workflow file.

JeffreyCA (approved):
Thanks, scaffolding files look good 👍

jongio (approved):
Re-approving after rebase onto main. No extension code changes since my prior approval. The implementation is solid: proper error handling with typed errors, path traversal protection, input validation, signal handling, and good test coverage.

jongio (approved):
The refactoring since my last review looks good. The split of invoke into run + invoke, the new project-layer abstractions (docker.go, runtime.go, identifiers.go, build.go), and the remote playground proxy are well-structured.

The separation of concerns is much cleaner now: 
un.go owns local container lifecycle, invoke.go owns remote sandbox management, and the project package provides shared Docker/build/runtime infrastructure.

A couple of minor notes (non-blocking):

- loadLocalRunState is called twice in the run path (once in nsureLocalContainerEndpoint, once in Run()). Not a bug since it just reads a file, but you could hoist the state to avoid the redundant read.
- projectRouteSegment(state) is called inside the retry loop in createSandboxWhenImageReady but won't change between attempts. Could be hoisted above the loop.

Neither warrants blocking. The code is clean, error handling is consistent, and the security validations (path traversal checks, operation allowlist in the proxy) are solid.

trangevi (approved):

vhvb1989 (approved):

jongio (approved):

tg-msft (approved):
Approving codeowners change

PR #8787, opened 2026-06-23
fix(agents): use venv for pip fallback in `azd ai agent run`

### LINKED ISSUE #8280: [Agents extension] `azd ai agent run` does not install into a venv when using pip
We install into a venv when using uv, but not with pip. We should look to be consistent.

trangevi:
## Why

When `uv` is not installed, `azd ai agent run` falls back to `pip` for Python dependency installation. However, the pip fallback path was installing packages into the global Python environment instead of a `.venv`, unlike the `uv` path which always creates and uses a virtual environment. This inconsistency causes global package pollution and surprising behavior for users without `uv`.

## What changed

Updated `installPythonDepsPip` to mirror the `uv` path behavior:

- **Creates a `.venv`** using `python -m venv` if one doesn't already exist (uses stdlib, no extra tools needed)
- **Uses the venv's pip binary** (`<venv>/bin/pip` or `<venv>/Scripts/pip.exe`) instead of system `pip` for all installs
- **Adds `pyproject.toml` support** via `pip install -e .` -- the `uv` path already supported this but the pip fallback did not

Added helper functions `venvPip()` (returns the OS-appropriate path to the venv pip) and `findSystemPython()` (locates `python3` or `python` on PATH for initial venv creation).

## Testing

- Existing `TestResolveVenvCommand` tests continue to pass
- Added `TestVenvPip` and `TestFindSystemPython` unit tests
- Build and lint pass cleanly

Fixes: #8280


jongio (commented):
Clean, well-scoped fix that brings the pip fallback in line with the uv path. The switch-based stat handling is solid, helper functions follow existing patterns (`venvPython`, `venvBinDir`), and tests use controlled PATH for determinism. Nice work addressing the earlier Copilot feedback in the follow-up commit.

One nit below about the version string in `findSystemPython`.

cli/azd/extensions/azure.ai.agents/internal/cmd/run.go
```diff
@@ -725,6 +766,27 @@ func venvBinDir(venvDir string) string {
...
+		candidates = append(candidates, "py")
+	}
+	for _, name := range candidates {
+		if p, err := exec.LookPath(name); err == nil {
+			return p, nil
+		}
+	}
+	return "", fmt.Errorf(
+		"python not found on PATH. Install Python 3.13+ from https://www.python.org/downloads/")
+}
```
jongio, line 805: nit: The uv path enforces `--python ">=3.13"` (line 587), but this fallback accepts whatever `python3`/`python`/`py` it finds without version validation. The error message here recommends "3.13+" but that's only shown when python is *missing* entirely. If someone has Python 3.9 on PATH, the venv gets created silently, and the agent code may fail at runtime with confusing errors if it relies on 3.13+ features.

Two options (both low-priority):
1. Run a quick version check after finding the binary (`python3 --version`, parse major.minor, warn if < 3.13)
2. Soften this to just "Python 3" since the function doesn't actually enforce a minimum

Either way, not blocking.

cli/azd/extensions/azure.ai.agents/internal/cmd/run.go
```diff
@@ -623,9 +623,49 @@ func installPythonDeps(projectDir string) error {
...
+		cmd := exec.Command(pythonBin, "-m", "venv", venvDir)
+		cmd.Dir = projectDir
+		cmd.Stdout = os.Stdout
+		cmd.Stderr = os.Stderr
+		if err := cmd.Run(); err != nil {
+			return fmt.Errorf("failed to create venv: %w", err)
+		}
+	}
+
+	pipPath := venvPip(venvDir)
```
huimiu, line 653: If `.venv` already exists from `uv venv`, it won't include `pip` by default, so this can fail with a missing executable. Would it make sense to bootstrap or check pip in the venv before installing dependencies?

jongio: Confirmed. Tracing the code paths: the uv path (line 587) runs `uv venv` which omits pip by default. When the pip fallback later reuses that `.venv`, the switch (lines 629-648) falls through to the default case (exists and is a directory), `venvPip()` returns a path to a binary that isn't there, and the subsequent `exec.Command` fails with an unhelpful error.

A `fileExists(pipPath)` guard before line 652 with `python -m ensurepip --upgrade` as a bootstrap step would handle this cleanly. Alternatively, a clear error telling the user to delete `.venv` and retry would also work.

jongio (commented):
The pip fallback venv reuse has a gap when `.venv` was previously created by `uv venv`, which doesn't include pip by default. If uv later becomes unavailable, the pip path skips venv creation (`.venv` exists) but `venvPip()` returns a path to a missing binary. This produces a confusing exec error rather than actionable guidance.

Consider checking `fileExists(pipPath)` before invoking it and running `python -m ensurepip --upgrade` as a bootstrap, or surfacing a clear error pointing the user to delete `.venv` and retry.

Not blocking for this PR since the core fix (venv isolation for pip fallback) is correct and well-tested. The uv-to-pip transition is an edge case that can be addressed in a follow-up.

jongio (commented):
Incremental review of 2 new commits (2225262, 7442f9c) since my last review.

Both prior findings are addressed:

- **Version enforcement**: `checkPythonVersion` now enforces `>=3.13` parity with the uv path's `--python ">=3.13"` constraint. Parsing logic handles standard CPython and non-standard (e.g. PyPy) version strings correctly.
- **uv-to-pip transition**: the `ensurepip` bootstrap detects a pip-less venv (from `uv venv`) and runs `python -m ensurepip --upgrade` before dependency installation. Placement after the venv existence check and before `pip install` is correct.

No new issues in the incremental diff. The test for `checkPythonVersion` exercises the real system Python, which validates format parsing. It doesn't exercise the "too old" rejection path deterministically (would need a mock), but the version comparison logic is straightforward.

jongio (commented):
Incremental review of 996234b (cspell config relocation).

Moving `ensurepip` from the core `cli/azd/.vscode/cspell.yaml` to the extension's own `cspell.yaml` is correct per the extension-specific config convention.

All findings from prior reviews are addressed. No new issues.

One blocker before merge: conflict in `cli/azd/extensions/azure.ai.agents/cspell.yaml` where `main` added `recordproxy` to the same section. Trivial resolution (keep both entries).

huimiu (approved):

trangevi:

jongio (commented):
Incremental review: the merge from main resolved the cspell.yaml conflict (both `ensurepip` and `recordproxy` entries preserved). The pip-fallback venv logic doesn't interact with the predeploy/postdeploy migration (#8788) or model updates (#8789) merged since my last pass. Agents gate CI is still pending.

huimiu (approved):

PR #8788, opened 2026-06-23
Migrate predeploy/postdeploy to service-level event handlers in agents extension

### LINKED ISSUE #7989: [Agents Extension] Migrate from project level hooks to service level hooks
By using project level hooks, we have to make sure we properly check that we're actually operating on an agent service. Instead, we should use service hooks, so it'll only be called if it's an agent service. (WithServiceEventHandler instead of WithProjectEventHandler)

trangevi:
## Motivation

The agents extension uses `WithProjectEventHandler` for all lifecycle events, which means each handler receives the full project config and must manually loop through services filtering by `svc.Host == AiAgentHost`. This is error-prone and unnecessary for events that fire at the service level.

## Approach

Switch `predeploy` and `postdeploy` from `WithProjectEventHandler` to `WithServiceEventHandler` with `&azdext.ServiceEventOptions{Host: AiAgentHost}`. The core now handles host-type filtering automatically, and each handler receives a single `*azdext.ServiceEventArgs` instead of needing to iterate all services.

Key changes:
- **Registration**: Two handlers use `WithServiceEventHandler` with the host filter
- **`predeployHandler`**: Operates directly on `args.Service` -- no loop, no host check
- **`postdeployHandler`**: Checks `isHostedAgentService` on the single service, fetches agent version and sets up RBAC for that one service, calls `reportSvcOptimizationDeployment` directly

`preprovision`, `postprovision`, and `postdown` remain as project-level handlers because the core only raises those events on the project config (not on individual service configs).

## Notes

- `reportOptimizationDeployments` (the multi-service wrapper) is no longer called from `postdeployHandler` but remains in `optimize_helpers.go` with its dedicated tests
- `CheckDeveloperRBAC` is idempotent so being called per-service (if multiple agent services exist) is safe
- Net: -103 lines, +61 lines

Fixes: #7989

### DIFF
--- a/cli/azd/extensions/azure.ai.agents/internal/cmd/listen.go
+++ b/cli/azd/extensions/azure.ai.agents/internal/cmd/listen.go
@@ -155,7 +155,10 @@ func currentEnvName(ctx context.Context, azdClient *azdext.AzdClient) (string, e
 // process lifetime. Service-level predeploy handlers fire per-service, but the
 // RBAC pre-flight check is project-scoped and idempotent — running it once is
 // sufficient and avoids duplicate ARM/Graph calls and noisy output.
-var developerRBACOnce sync.Once
+var (
+	developerRBACOnce sync.Once
+	developerRBACErr  error
+)
 
 func predeployHandler(ctx context.Context, azdClient *azdext.AzdClient, args *azdext.ServiceEventArgs) error {
 	svc := args.Service
@@ -171,12 +174,11 @@ func predeployHandler(ctx context.Context, azdClient *azdext.AzdClient, args *az
 	// Guarded by sync.Once since this handler fires per-service but the check
 	// is project-scoped.
 	if isHostedAgentService(svc, args.Project) {
-		var rbacErr error
 		developerRBACOnce.Do(func() {
-			rbacErr = project.CheckDeveloperRBAC(ctx, azdClient)
+			developerRBACErr = project.CheckDeveloperRBAC(ctx, azdClient)
 		})
-		if rbacErr != nil {
-			return rbacErr
+		if developerRBACErr != nil {
+			return developerRBACErr
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


vhvb1989 (approved):

cli/azd/extensions/azure.ai.agents/internal/cmd/listen.go
```diff
@@ -150,29 +151,32 @@ func currentEnvName(ctx context.Context, azdClient *azdext.AzdClient) (string, e
...
-		if err := project.CheckDeveloperRBAC(ctx, azdClient); err != nil {
-			return err
+	// Guarded by sync.Once since this handler fires per-service but the check
+	// is project-scoped.
+	if isHostedAgentService(svc, args.Project) {
+		var rbacErr error
+		developerRBACOnce.Do(func() {
+			rbacErr = project.CheckDeveloperRBAC(ctx, azdClient)
+		})
+		if rbacErr != nil {
```
jongio: The `sync.Once` + local `rbacErr` pattern silently swallows the error for subsequent services. If `CheckDeveloperRBAC` fails on service A:

1. `Once.Do` fires, sets service A's local `rbacErr`, service A returns error
2. Service B calls this: `Once.Do` is a no-op, service B's local `rbacErr` stays `nil`, proceeds without the check

If the intent is "run once, propagate the result to all callers," store the error alongside the Once:

```suggestion
	if isHostedAgentService(svc, args.Project) {
		developerRBACOnce.Do(func() {
			developerRBACErr = project.CheckDeveloperRBAC(ctx, azdClient)
		})
		if developerRBACErr != nil {
			return developerRBACErr
		}
	}
```

With a package-level companion:
```go
var (
	developerRBACOnce sync.Once
	developerRBACErr  error
)
```

This ensures all services see the failure, not just the first one through the gate.

jongio (commented):
Clean simplification. The migration from project-level to service-level handlers removes unnecessary iteration and lets the core handle host-type filtering, which is the right architectural direction.

One concern with the `sync.Once` error propagation pattern in `predeployHandler` (see inline comment). Not blocking since the multi-hosted-agent-service + RBAC-failure scenario is uncommon, but worth addressing to keep the semantics correct.

Tests appropriately track the handler signature changes and drop coverage for filtering now handled by the core.

huimiu (commented):
Nice cleanup on this migration! One thing I want to flag so we don't lose it: the old `reportOptimizationDeployments` wrapped each service in a `recover()` to keep optimization reporting best-effort (per our "log failures but never block the deploy" contract). Calling `reportSvcOptimizationDeployment` directly drops that guard, and since the broker turns a recovered panic into a handler error, a panic here would now fail the whole deploy. What do you think about keeping the `recover()` — either at the call site or pushed into the func?

jongio (commented):
The third commit addresses both open concerns from earlier reviews.

`developerRBACErr` is now package-level, so all service handlers see the RBAC check failure, not just the first caller. This matches the suggestion from my earlier review and is correct.

The `reportSvcOptimizationDeployment` call is now wrapped in `defer/recover()`, restoring the best-effort contract that @huimiu flagged. Panics in optimization reporting won't block the deploy.

One note: `reportOptimizationDeployments` (the multi-service loop wrapper in `optimize_helpers.go`) is no longer called from production code. It's only exercised by its own tests. Consider removing it in a follow-up to avoid dead code.

@vhvb1989's approval was submitted against `cadaedc4a` and is now stale against the current HEAD `357ca6d8`. Will need reapproval before merge.

huimiu (approved):

PR #8789, opened 2026-06-23
Update agent models to match TypeSpec definition

### LINKED ISSUE #8338: `agent show` needs to be updated with new typespec objects
Typespec reference: https://github.com/Azure/azure-rest-api-specs/blob/feature/foundry-release/specification/ai-foundry/data-plane/Foundry/src/agents/models.tsp

Our agent object is outdated compared to the service typespec and should be updated to make sure we're showing all of the correct properties from the service returned object.

trangevi:
## Motivation

Our Go agent model structs (`AgentObject`, `AgentEndpoint`) were outdated compared to the service TypeSpec at [azure-rest-api-specs](https://github.com/Azure/azure-rest-api-specs/blob/feature/foundry-release/specification/ai-foundry/data-plane/Foundry/src/agents/models.tsp). Missing fields meant the CLI silently dropped properties returned by the API.

## Changes

**`AgentObject`** -- added fields present in the TypeSpec but missing from Go:
- `State` (operational state: enabled/disabled)
- `InstanceIdentity`, `Blueprint`, `BlueprintReference` (identity/blueprint info already on `AgentVersionObject` but missing at the agent level)

**`AgentEndpoint`** -- added `ProtocolConfiguration` field and supporting types:
- `ProtocolConfiguration` struct with per-protocol config pointers (activity, responses, a2a, mcp, invocations, invocations_ws)
- `ActivityProtocolConfiguration` with `EnableM365PublicEndpoint`
- Empty marker structs for the other protocols (matching the TypeSpec which reserves them for future fields)

**Constants** -- added `AgentEndpointAuthSchemeBotServiceTenant` to match the TypeSpec `AgentEndpointAuthorizationSchemeType` union.

**Tests** -- added two round-trip tests covering the new fields:
- `TestAgentObject_RoundTrip_AllFields`
- `TestAgentEndpoint_ProtocolConfiguration_RoundTrip`

## Notes

- `AgentVersionObject` was already aligned with TypeSpec -- no changes needed there.
- The `agent show` command calls `GetAgentVersion` (returns `AgentVersionObject`), so its output is unaffected. The `AgentObject` changes benefit commands that use `GetAgent` (e.g., list).

Fixes: #8338

### DIFF
--- a/cli/azd/extensions/azure.ai.agents/internal/cmd/endpoint_show.go
+++ b/cli/azd/extensions/azure.ai.agents/internal/cmd/endpoint_show.go
@@ -141,6 +141,8 @@ func resolveEndpointProtocols(endpoint *agent_api.AgentEndpoint) []string {
 	}
 
 	// Prefer protocol_configuration (newer API shape).
+	// A non-nil ProtocolConfiguration is authoritative even if empty,
+	// so we never fall through to the deprecated Protocols field.
 	if pc := endpoint.ProtocolConfiguration; pc != nil {
 		var protocols []string
 		if pc.Activity != nil {
@@ -161,9 +163,7 @@ func resolveEndpointProtocols(endpoint *agent_api.AgentEndpoint) []string {
 		if pc.InvocationsWS != nil {
 			protocols = append(protocols, "invocations_ws")
 		}
-		if len(protocols) > 0 {
-			return protocols
-		}
+		return protocols
 	}
 
 	// Fall back to deprecated Protocols field.
--- a/cli/azd/extensions/azure.ai.agents/internal/cmd/endpoint_show.go
+++ b/cli/azd/extensions/azure.ai.agents/internal/cmd/endpoint_show.go
@@ -132,6 +132,52 @@ func printEndpointJSON(agent *agent_api.AgentObject) error {
 	return enc.Encode(out)
 }
 
+// resolveEndpointProtocols returns the list of enabled protocol names for display.
+// It prefers ProtocolConfiguration (where key presence declares enablement) and falls
+// back to the deprecated Protocols field for older API responses.
+func resolveEndpointProtocols(endpoint *agent_api.AgentEndpoint) []string {
+	if endpoint == nil {
+		return nil
+	}
+
+	// Prefer protocol_configuration (newer API shape).
+	if pc := endpoint.ProtocolConfiguration; pc != nil {
+		var protocols []string
+		if pc.Activity != nil {
+			protocols = append(protocols, "activity")
+		}
+		if pc.Responses != nil {
+			protocols = append(protocols, "responses")
+		}
+		if pc.A2A != nil {
+			protocols = append(protocols, "a2a")
+		}
+		if pc.MCP != nil {
+			protocols = append(protocols, "mcp")
+		}
+		if pc.Invocations != nil {
+			protocols = append(protocols, "invocations")
+		}
+		if pc.InvocationsWS != nil {
+			protocols = append(protocols, "invocations_ws")
+		}
+		if len(protocols) > 0 {
+			return protocols
+		}
+	}
+
+	// Fall back to deprecated Protocols field.
+	if len(endpoint.Protocols) > 0 {
+		protocols := make([]string, len(endpoint.Protocols))
+		for i, p := range endpoint.Protocols {
+			protocols[i] = string(p)
+		}
+		return protocols
+	}
+
+	return nil
+}
+
 func printEndpointTable(agent *agent_api.AgentObject) error {
 	w := tabwriter.NewWriter(os.Stdout, 0, 4, 2, ' ', 0)
 
@@ -140,11 +186,7 @@ func printEndpointTable(agent *agent_api.AgentObject) error {
 
 	// Protocols
 	fmt.Fprintf(w, "Protocols:\t")
-	if agent.AgentEndpoint != nil && len(agent.AgentEndpoint.Protocols) > 0 {
-		protocols := make([]string, len(agent.AgentEndpoint.Protocols))
-		for i, p := range agent.AgentEndpoint.Protocols {
-			protocols[i] = string(p)
-		}
+	if protocols := resolveEndpointProtocols(agent.AgentEndpoint); len(protocols) > 0 {
 		fmt.Fprintf(w, "%s\n", strings.Join(protocols, ", "))
 	} else {
 		fmt.Fprintf(w, "(not configured)\n")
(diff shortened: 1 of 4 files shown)


jongio (commented):
Solid TypeSpec sync. The model additions are well-scoped, the `resolveEndpointProtocols` extraction is a clean refactor, and the round-trip tests cover the new shapes thoroughly. One CI fix needed before merge.

**CI**: `golangci-lint` is failing on `gofmt`. The `ProtocolConfiguration` and `AgentEndpoint` structs have manually-padded spaces aligning struct tags. Running `gofmt -s -w models.go` will fix it. Same likely applies to the `AgentObject` struct fields added in this PR.

cli/azd/extensions/azure.ai.agents/internal/pkg/agents/agent_api/models.go
```diff
@@ -142,11 +143,43 @@ type AgentEndpointAuthorizationScheme struct {
...
+// ActivityProtocolConfiguration describes configuration specific to the activity protocol.
+type ActivityProtocolConfiguration struct {
+	EnableM365PublicEndpoint *bool `json:"enable_m365_public_endpoint,omitempty"`
+}
+
+// ResponsesProtocolConfiguration describes configuration specific to the responses protocol.
+type ResponsesProtocolConfiguration struct{}
+
+// A2AProtocolConfiguration describes configuration specific to the A2A protocol.
+type A2AProtocolConfiguration struct{}
```
jongio, line 155: `gofmt` is flagging this struct (CI failure). The spaces between each type and its struct tag are manually padded for alignment, but gofmt expects tab-based alignment here. Run `gofmt -s -w` on this file to fix.

The same applies to the `AgentEndpoint` and `AgentObject` struct field blocks you reformatted below.

cli/azd/extensions/azure.ai.agents/internal/cmd/endpoint_show.go
```diff
@@ -132,6 +132,52 @@ func printEndpointJSON(agent *agent_api.AgentObject) error {
...
+		if pc.MCP != nil {
+			protocols = append(protocols, "mcp")
+		}
+		if pc.Invocations != nil {
+			protocols = append(protocols, "invocations")
+		}
+		if pc.InvocationsWS != nil {
+			protocols = append(protocols, "invocations_ws")
+		}
+		if len(protocols) > 0 {
```
huimiu: When `ProtocolConfiguration` is non-nil but all fields are nil (e.g. the server returns `"protocol_configuration": {}`), this guard causes the function to fall through and surface the deprecated `Protocols` list. A non-nil `pc` should be treated as authoritative - return `protocols` even if empty, so older data doesn't silently override the new shape:

``go
if pc := endpoint.ProtocolConfiguration; pc != nil {
    // ... build list ...
    return protocols // return even if empty
}
``

cli/azd/extensions/azure.ai.agents/internal/cmd/endpoint_show.go
```diff
@@ -132,6 +132,52 @@ func printEndpointJSON(agent *agent_api.AgentObject) error {
...
+			return protocols
+		}
```
jongio: This guard causes the function to fall through to the deprecated `Protocols` field when `ProtocolConfiguration` is non-nil but all inner fields are nil (e.g., the server returns `"protocol_configuration": {}`). As @huimiu noted, a non-nil `pc` should be treated as authoritative.

Remove the length check so the function returns `protocols` (even if empty) whenever `ProtocolConfiguration` is present:

```suggestion
		return protocols
```

cli/azd/extensions/azure.ai.agents/internal/cmd/endpoint_show_test.go
```diff
@@ -135,6 +135,70 @@ func TestPrintEndpointJSON(t *testing.T) {
...
+			Protocols: []agent_api.AgentEndpointProtocol{"activity"},
+			ProtocolConfiguration: &agent_api.ProtocolConfiguration{
+				Responses: &agent_api.ResponsesProtocolConfiguration{},
+				MCP:       &agent_api.MCPProtocolConfiguration{},
+			},
+		}, []string{"responses", "mcp"}},
+		{"fallback to protocols", &agent_api.AgentEndpoint{
+			Protocols: []agent_api.AgentEndpointProtocol{"responses", "a2a"},
+		}, []string{"responses", "a2a"}},
+	}
```
jongio, line 196: Missing test case for the scenario @huimiu flagged: non-nil but empty `ProtocolConfiguration` with a populated `Protocols` field. This verifies that an empty `ProtocolConfiguration` doesn't fall through to the deprecated field.

```go
{"empty protocol_configuration is authoritative", &agent_api.AgentEndpoint{
	Protocols: []agent_api.AgentEndpointProtocol{"activity"},
	ProtocolConfiguration: &agent_api.ProtocolConfiguration{},
}, nil},
```

jongio (dismissed):
Two issues still blocking merge:

1. `resolveEndpointProtocols` falls through to the deprecated `Protocols` field when `ProtocolConfiguration` is non-nil but empty (`{}`). A non-nil `ProtocolConfiguration` should be treated as authoritative, even if no protocols are enabled inside it. See inline comment for the fix.

2. CI (`golangci-lint`) is still failing on `gofmt` formatting. The `ProtocolConfiguration`, `AgentEndpoint`, and `AgentObject` structs in `models.go` have manual space-padding for struct tag alignment. Run `gofmt -s -w models.go` to fix.

jongio (commented):
Incremental review of aca656b (since my prior review on 9d81a0c).

All three prior findings are addressed:

1. `gofmt` alignment in `ProtocolConfiguration`, `AgentEndpoint`, and `AgentObject` structs is fixed.
2. `resolveEndpointProtocols` now treats a non-nil `ProtocolConfiguration` as authoritative, returning immediately even when empty. This prevents the incorrect fallthrough to the deprecated `Protocols` field.
3. Test case added for the empty-`ProtocolConfiguration`-with-populated-`Protocols` scenario, confirming the authoritative behavior.

Lint and tests are passing. No new issues in the incremental diff.

Note: my prior CHANGES_REQUESTED verdict is still active on this PR. A maintainer with approval rights should clear it if the review is satisfactory.

huimiu (approved):

jongio (commented):
Verified the four commits in `aca656b`:

1. `resolveEndpointProtocols` treats a non-nil `ProtocolConfiguration` as authoritative, returning immediately when empty instead of falling through to the deprecated `Protocols` field.
2. Model additions (`State`, `InstanceIdentity`, `Blueprint`, `BlueprintReference` on `AgentObject`; `ProtocolConfiguration` on `AgentEndpoint`; `BotServiceTenant` auth scheme) match the TypeSpec and use `omitempty` for optional fields.
3. Go initialism casing is correct (`MCPProtocolConfiguration`, `InvocationsWSProtocolConfiguration`).
4. Test coverage includes round-trip tests for new model shapes and table-driven tests for protocol resolution, including the empty-`ProtocolConfiguration`-with-populated-`Protocols` edge case.

No new issues found. My prior CHANGES_REQUESTED is still active and needs to be cleared by a maintainer with approval rights.

PR #8790, opened 2026-06-23
Add unit tests for azure.ai.routines client and fix MarkHidden nit

### LINKED ISSUE #8365: Follow up items from `azd ai routine` command PR
From https://github.com/Azure/azure-dev/pull/8241#pullrequestreview-4347386773

4. internal/pkg/routines/client.go (+390 LOC) — no unit tests
The data-plane client (CRUD, pagination, enable/disable GET+PUT fallback, dispatchAsync, list runs, custom HTTP/2 transport, same-origin pagination guard) has zero tests. PR description defers "unit tests as follow-up", but the manifest parser (low-risk) gets tests while the mission-critical client does not. Suggest at minimum httptest.NewServer-based tests for: 2xx happy paths, 404/409/5xx, nextLink pagination, validateSameOrigin rejection, context cancellation.

5. Command-layer coverage gaps
No tests for routine_update.go (complex flag→file→default merge), routine_delete, routine_enable, routine_disable, routine_dispatch, routine_list, routine_show, routine_helpers. Agents extension test ratio is ~1:1; this PR is ~1:2.2.

6. internal/pkg/routines/client.go — runtime.NewResponseError(resp) leaks raw response body into errors
Used in ~5 spots. Service error bodies pass straight through. CLI commands re-wrap with exterrors.ServiceFromAzure, but the underlying message text is preserved and may surface in logs/output.
Fix: drain body, extract structured fields, return a controlled message — or document that callers must always wrap and never log the raw error.

🟢 Low / nit
9. internal/cmd/endpoint_test.go — port edge cases under-tested
The recently-added port rejection covers :443 only. Add: non-default ports (:8443), IPv6+port (https://[::1]:443/...), whitespace-only input. The original bug evaded tests because they didn't exercise non-default ports.

10. internal/cmd/routine_update.go — _ = cmd.Flags().MarkHidden(...) (×2)
Silent error discard. Safe in practice, but if err := ...; err != nil { ... } or a brief comment is cleaner.

11. AGENTS.md — "If this extension grows enough… introduce internal/exterrors"
Phrased as a future option, but the package exists in this PR. Rewrite in present tense so future maintainers don't treat it as optional.

12. go.mod — armappservice/v2 v2.3.0 // indirect added
Indirect, likely a cascade from an azd-core dep bump. Worth a quick confirmation that this isn't pulling unwanted weight on the extension.

trangevi:
## Why

The initial `azd ai routine` extension PR (#8241) shipped without unit tests for the mission-critical data-plane client and with a minor code-hygiene nit (`_ = MarkHidden` with no explanation). Issue #8365 tracked these as follow-up items.

## What this PR does

- **Adds `client_test.go`** -- 28 httptest-based unit tests covering the full `routines.Client` surface: `GetRoutine`, `ListRoutines` (single and multi-page pagination), `PutRoutine`, `DeleteRoutine`, `EnableRoutine`, `DisableRoutine`, `DispatchRoutineAsync`, `ListRoutineRuns` (pagination with Top cap, filter parameter), URL construction helpers, and context cancellation.

- **Adds `routine_helpers_test.go`** -- unit tests for the `boolStr` and `sortedKeys` helper functions that were previously untested.

- **Clarifies the `MarkHidden` discard** in `routine_update.go` with a comment explaining why ignoring the error is safe (the flags are registered on the same `cmd` immediately above, so the call cannot fail).

## Items assessed but deferred

- **Item 6** (`runtime.NewResponseError` leaking raw body): This is the standard azcore pattern. Command-layer callers already wrap errors via `exterrors.ServiceFromAzure`, which extracts structured fields before surfacing to users.
- **Item 12** (`armappservice/v2` indirect dep): Benign cascade from the azd core module; adds no runtime weight.
- **Items 9 and 11**: Already addressed in prior commits (port edge-case tests and AGENTS.md present-tense fix).

Fixes: #8365

### DIFF
--- a/cli/azd/extensions/azure.ai.routines/internal/pkg/routines/client_test.go
+++ b/cli/azd/extensions/azure.ai.routines/internal/pkg/routines/client_test.go
@@ -1,0 +1,464 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package routines
+
+import (
+	"context"
+	"encoding/json"
+	"net/http"
+	"net/http/httptest"
+	"sync/atomic"
+	"testing"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/policy"
+	azruntime "github.com/Azure/azure-sdk-for-go/sdk/azcore/runtime"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// newTestClient creates a Client with a pipeline that skips auth (no TLS
+// requirement) pointing at a local httptest server.
+func newTestClient(t *testing.T, handler http.Handler) (*Client, *httptest.Server) {
+	t.Helper()
+	srv := httptest.NewServer(handler)
+	t.Cleanup(srv.Close)
+
+	// Build a pipeline without bearer-token policy so plain HTTP works.
+	pipeline := azruntime.NewPipeline("test", "v0", azruntime.PipelineOptions{}, &policy.ClientOptions{})
+	client := &Client{
+		endpoint: srv.URL + "/api/projects/test-project",
+		pipeline: pipeline,
+	}
+	return client, srv
+}
+
+// ─── GetRoutine ──────────────────────────────────────────────────────────────
+
+func TestGetRoutine_Success(t *testing.T) {
+	t.Parallel()
+	routine := Routine{Name: "my-routine", Description: "test routine", Enabled: new(true)}
+
+	client, _ := newTestClient(t, http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		assert.Equal(t, http.MethodGet, r.Method)
+		assert.Contains(t, r.URL.Path, "/routines/my-routine")
+		assert.Equal(t, routinesPreviewValue, r.Header.Get(routinesPreviewHeader))
+
+		w.Header().Set("Content-Type", "application/json")
+		json.NewEncoder(w).Encode(routine)
+	}))
+
+	got, err := client.GetRoutine(t.Context(), "my-routine")
+	require.NoError(t, err)
+	assert.Equal(t, "my-routine", got.Name)
+	assert.Equal(t, "test routine", got.Description)
+	assert.True(t, *got.Enabled)
+}
+
+func TestGetRoutine_NotFound(t *testing.T) {
+	t.Parallel()
+	client, _ := newTestClient(t, http.HandlerFunc(func(w http.ResponseWriter, _ *http.Request) {
+		w.WriteHeader(http.StatusNotFound)
+		w.Write([]byte(`{"error":{"code":"NotFound","message":"routine not found"}}`))
+	}))
+
+	_, err := client.GetRoutine(t.Context(), "nonexistent")
+	require.Error(t, err)
+
+	var respErr *azcore.ResponseError
+	require.ErrorAs(t, err, &respErr)
+	assert.Equal(t, http.StatusNotFound, respErr.StatusCode)
+}
+
+func TestGetRoutine_ContextCancellation(t *testing.T) {
+	t.Parallel()
+	client, _ := newTestClient(t, http.HandlerFunc(func(w http.ResponseWriter, _ *http.Request) {
+		w.WriteHeader(http.StatusOK)
+		json.NewEncoder(w).Encode(Routine{Name: "x"})
+	}))
+
+	ctx, cancel := context.WithCancel(t.Context())
+	cancel() // cancel immediately
+
+	_, err := client.GetRoutine(ctx, "x")
+	require.Error(t, err)
+}
+
+// ─── ListRoutines ────────────────────────────────────────────────────────────
+
+func TestListRoutines_SinglePage(t *testing.T) {
+	t.Parallel()
+	page := PagedRoutine{
+		Value: []Routine{
+			{Name: "r1"},
+			{Name: "r2"},
+		},
+	}
+
+	client, _ := newTestClient(t, http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		assert.Equal(t, http.MethodGet, r.Method)
+		w.Header().Set("Content-Type", "application/json")
+		json.NewEncoder(w).Encode(page)
+	}))
+
+	got, err := client.ListRoutines(t.Context())
+	require.NoError(t, err)
+	assert.Len(t, got, 2)
+	assert.Equal(t, "r1", got[0].Name)
+	assert.Equal(t, "r2", got[1].Name)
+}
+
+func TestListRoutines_MultiPage(t *testing.T) {
+	t.Parallel()
+	var callCount atomic.Int32
+
+	client, _ := newTestClient(t, http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		call := callCount.Add(1)
+		w.Header().Set("Content-Type", "application/json")
+
+		switch call {
+		case 1:
+			// First page has a continuation token
+			json.NewEncoder(w).Encode(PagedRoutine{
+				Value:             []Routine{{Name: "r1"}},
+				ContinuationToken: "token-page2",
+			})
+		case 2:
+			// Second page: verify "after" query param is passed
+			assert.Contains(t, r.URL.RawQuery, "after=token-page2")
+			json.NewEncoder(w).Encode(PagedRoutine{
+				Value: []Routine{{Name: "r2"}},
+			})
+		default:
+			t.Fatal("unexpected extra request")
+		}
+	}))
+
+	got, err := client.ListRoutines(t.Context())
+	require.NoError(t, err)
+	assert.Len(t, got, 2)
+	assert.Equal(t, "r1", got[0].Name)
+	assert.Equal(t, "r2", got[1].Name)
+}
+
+func TestListRoutines_ServerError(t *testing.T) {
+	t.Parallel()
+	client, _ := newTestClient(t, http.HandlerFunc(func(w http.ResponseWriter, _ *http.Request) {
+		w.WriteHeader(http.StatusInternalServerError)
+		w.Write([]byte(`{"error":{"code":"InternalError","message":"boom"}}`))
+	}))
+
+	_, err := client.ListRoutines(t.Context())
+	require.Error(t, err)
+}
+
+// ─── PutRoutine ──────────────────────────────────────────────────────────────
+
+func TestPutRoutine_Created(t *testing.T) {
+	t.Parallel()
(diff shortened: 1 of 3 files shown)


jongio (commented):
Tests are thorough and well-structured: good use of `t.Parallel()`, `t.Context()`, `atomic.Int32` for concurrency safety, and comprehensive coverage of happy paths, error responses, pagination, and URL encoding.

**CI blocker**: The `golangci-lint` job is red due to 22 gosec `G104: Errors unhandled` violations on `w.Write(...)` and `json.NewEncoder(w).Encode(...)` calls in test handlers. Quick fix: assign to blank identifier on each one, e.g. `_ = json.NewEncoder(w).Encode(routine)`. Alternatively, a file-level `//nolint:gosec` directive at the top of the test file would suppress them all, though per-line discards are more explicit.

I concur with the existing Copilot feedback on `require` usage inside httptest handler goroutines (lines 165, 300). Switching those two to `assert.NoError` (or decoding into a local and checking after the request round-trips) avoids potential goroutine panics.

cli/azd/extensions/azure.ai.routines/internal/pkg/routines/client_test.go
```diff
@@ -0,0 +1,464 @@
...
+	t.Parallel()
+	routine := Routine{Name: "my-routine", Description: "test routine", Enabled: new(true)}
+
+	client, _ := newTestClient(t, http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		assert.Equal(t, http.MethodGet, r.Method)
+		assert.Contains(t, r.URL.Path, "/routines/my-routine")
+		assert.Equal(t, routinesPreviewValue, r.Header.Get(routinesPreviewHeader))
+
+		w.Header().Set("Content-Type", "application/json")
+		json.NewEncoder(w).Encode(routine)
```
jongio: gosec G104: this `Encode` return value (and the ~20 similar `w.Write`/`Encode` calls throughout the file) is what's failing the lint job. Quickest fix: `_ = json.NewEncoder(w).Encode(routine)`. Same pattern for every handler `Write` call below.

huimiu: The azure.ai.skills client test already uses `_, _ = w.Write(...)` and `_ = json.NewEncoder(w).Encode(...)` as the established pattern here

cli/azd/extensions/azure.ai.routines/internal/pkg/routines/client_test.go
```diff
@@ -0,0 +1,464 @@
...
+// ─── PutRoutine ──────────────────────────────────────────────────────────────
+
+func TestPutRoutine_Created(t *testing.T) {
+	t.Parallel()
+	client, _ := newTestClient(t, http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		assert.Equal(t, http.MethodPut, r.Method)
+		assert.Equal(t, "application/json", r.Header.Get("Content-Type"))
+
+		var body Routine
+		require.NoError(t, json.NewDecoder(r.Body).Decode(&body))
```
jongio: `require.NoError` calls `t.FailNow()`, which panics when called from a non-test goroutine (httptest handlers run in the server's goroutine). Same issue the second commit fixed for `t.Fatal()`. Use the same pattern:

```suggestion
		if err := json.NewDecoder(r.Body).Decode(&body); err != nil {
			w.WriteHeader(http.StatusInternalServerError)
			return
		}
```

cli/azd/extensions/azure.ai.routines/internal/pkg/routines/client_test.go
```diff
@@ -0,0 +1,464 @@
...
+// ─── DispatchRoutineAsync ────────────────────────────────────────────────────
+
+func TestDispatchRoutineAsync_Success(t *testing.T) {
+	t.Parallel()
+	client, _ := newTestClient(t, http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		assert.Equal(t, http.MethodPost, r.Method)
+		assert.Contains(t, r.URL.Path, "/routines/my-routine:dispatch_async")
+
+		var body DispatchRoutineRequest
+		require.NoError(t, json.NewDecoder(r.Body).Decode(&body))
```
jongio: Same `require` in handler goroutine issue. Apply the same fix:

```suggestion
		if err := json.NewDecoder(r.Body).Decode(&body); err != nil {
			w.WriteHeader(http.StatusInternalServerError)
			return
		}
```

jongio (changes requested):
Two items blocking this PR:

**gosec G104 (CI red)**: `golangci-lint` is failing because `w.Write(...)` and `json.NewEncoder(w).Encode(...)` return values aren't consumed in test handlers. The sibling tests in `azure.ai.agents/optimize_api` use `_, _ = w.Write(...)` and `_ = json.NewEncoder(w).Encode(...)`. Apply that pattern to all 22 call sites.

**`require` in httptest goroutines**: The second commit replaced explicit `t.Fatal()` calls with `w.WriteHeader(500)`, but two `require.NoError` calls remain inside handlers (lines 165, 300). `require` calls `t.FailNow()`, which panics from non-test goroutines. See inline suggestions for the fix.

cli/azd/extensions/azure.ai.routines/internal/pkg/routines/client_test.go
jongio: `require.NoError` calls `t.FailNow()`, which panics when called from a non-test goroutine (httptest handlers run in the server's goroutine). Same issue the second commit fixed for `t.Fatal()`. Use the same pattern:

```suggestion
		if err := json.NewDecoder(r.Body).Decode(&body); err != nil {
			w.WriteHeader(http.StatusInternalServerError)
			return
		}
```

cli/azd/extensions/azure.ai.routines/internal/pkg/routines/client_test.go
jongio: Same `require` in handler goroutine issue. Apply the same fix:

```suggestion
		if err := json.NewDecoder(r.Body).Decode(&body); err != nil {
			w.WriteHeader(http.StatusInternalServerError)
			return
		}
```

jongio (dismissed):
Two items blocking this PR:

**gosec G104 (CI red)**: `golangci-lint` is failing because `w.Write(...)` and `json.NewEncoder(w).Encode(...)` return values aren't consumed in test handlers. The sibling tests in `azure.ai.agents/optimize_api` use `_, _ = w.Write(...)` and `_ = json.NewEncoder(w).Encode(...)`. Apply that pattern to all 22 call sites.

**`require` in httptest goroutines**: The second commit replaced explicit `t.Fatal()` calls with `w.WriteHeader(500)`, but two `require.NoError` calls remain inside handlers (lines 165, 300). `require` calls `t.FailNow()`, which panics from non-test goroutines. See inline suggestions for the fix.

jongio (commented):
Both items from my previous review are addressed. The new commit:

1. Adds `_ =` / `_, _ =` to all `json.NewEncoder(w).Encode(...)` and `w.Write(...)` calls, fixing the gosec G104 violations that were failing CI.
2. Replaces both `require.NoError` calls inside httptest handler goroutines with the error-check + `WriteHeader(500)` pattern.

Zero `require` calls remain inside httptest handlers. No new issues in the incremental diff. CI is running against `fc884da`.

jongio (commented):
Both items from my previous review are addressed. The new commit:

1. Adds `_ =` / `_, _ =` to all `json.NewEncoder(w).Encode(...)` and `w.Write(...)` calls, fixing the gosec G104 violations that were failing CI.
2. Replaces both `require.NoError` calls inside httptest handler goroutines with the error-check + `WriteHeader(500)` pattern.

Zero `require` calls remain inside httptest handlers. No new issues in the incremental diff. CI is running against `fc884da`.

huimiu (approved):

cli/azd/extensions/azure.ai.routines/internal/pkg/routines/client_test.go
```diff
@@ -0,0 +1,470 @@
...
+		w.Header().Set("Content-Type", "application/json")
+		_ = json.NewEncoder(w).Encode(PagedRoutineRun{
+			Value:         []RoutineRun{{ID: "run-1"}, {ID: "run-2"}},
+			NextPageToken: "page2",
+		})
+	}))
+
+	runs, err := client.ListRoutineRuns(t.Context(), "my-routine", ListRoutineRunsOptions{Top: 1})
+	require.NoError(t, err)
+	assert.Len(t, runs, 1, "Top should cap results to 1")
```
jongio, line 404: Nit: consider adding an `atomic.Int32` counter (like `TestListRoutines_MultiPage` uses) to assert only 1 HTTP request is made when `Top` caps results on the first page. Right now, if the client accidentally fetched page 2 before checking `Top`, the test would still pass because truncation happens after accumulation.

cli/azd/extensions/azure.ai.routines/internal/pkg/routines/client_test.go
```diff
@@ -0,0 +1,470 @@
...
+		w.Header().Set("Content-Type", "application/json")
+		_ = json.NewEncoder(w).Encode(Routine{Name: "my-routine", Enabled: new(false)})
+	}))
+
+	got, err := client.DisableRoutine(t.Context(), "my-routine")
+	require.NoError(t, err)
+	assert.False(t, *got.Enabled)
+}
+
+func TestEnableRoutine_ServerError(t *testing.T) {
```
jongio, line 283: Nit: symmetry gap. `EnableRoutine` has a server error test but `DisableRoutine` doesn't. Both go through `postRoutineAction` so it's the same code path, but for completeness a `TestDisableRoutine_ServerError` companion would close the gap.

jongio (commented):
Two nit-level observations, neither blocking.

jongio (commented):
Both nits from my last review are addressed in 6e47e23:

1. TestListRoutineRuns_WithTop now uses an tomic.Int32 counter and asserts exactly one HTTP call is made, confirming the client respects Top without over-fetching.
2. TestDisableRoutine_ServerError closes the symmetry gap with EnableRoutine.

All earlier blocking items (gosec G104, 
equire in handler goroutines) remain fixed. No new issues in this commit. CI is in progress with no failures.

My earlier CHANGES_REQUESTED should be cleared to unblock merge.

jongio (commented):
Both nits from my last review are addressed in 6e47e23:

1. TestListRoutineRuns_WithTop now uses an atomic.Int32 counter and asserts exactly one HTTP call is made, confirming the client respects Top without over-fetching.
2. TestDisableRoutine_ServerError closes the symmetry gap with EnableRoutine.

All earlier blocking items (gosec G104, require in handler goroutines) remain fixed. No new issues in this commit. CI is in progress with no failures.

My earlier CHANGES_REQUESTED should be cleared to unblock merge.

huimiu (approved):

PR #8791, opened 2026-06-23
fix(azure.ai.skills): address follow-up items from skill commands PR

### LINKED ISSUE #8366: Follow up items from `azd ai skill` commands PR
Higher-impact (worth a follow-up issue at minimum)

- CreateVersionFromZip buffers the full archive in memory — client.go (around the new multipart build). The pre-refactor CreatePackage took an io.ReadSeeker and streamed via httpReq.SetBody(streaming(archive), …). The new path does io.Copy(part, archive) into a bytes.Buffer and then bytes.NewReader(buf.Bytes()). Given the documented 512 MB archive ceiling, a near-cap upload allocates ≥512 MB heap (often 2× during buf.Bytes()), plus pipeline retry buffering. Easy OOM on constrained runners. Consider io.Pipe — a goroutine writes the files[] part + default field to the pipe writer, pass the pipe reader to httpReq.SetBody, set ContentLength = -1. Restoring the io.ReadSeeker parameter would also let the pipeline rewind on retry.

- POST /skills/{name}/versions has no idempotency guard — the Azure SDK pipeline retries automatically. A client-side timeout on a request the server actually processed will create a second immutable version, and because both CreateVersionInline and CreateVersionFromZip set default: true, the retry silently re-flips default_version to the duplicate. The user is now running on a version they didn't intend. Options: send an Idempotency-Key if the service supports it, or set policy.RetryOptions{MaxRetries: 0} for these two requests so retries are explicit. Worth checking with the service team which they prefer.

Medium

- --force is delete-then-create with no rollback, and the blast radius grew under versioning — DeleteSkill now wipes the skill and every version. If the subsequent CreateVersion* fails (network, 5xx, validation), all prior versions are gone. verifyFileNameMatches protects against the wrong-target case but not the upload-failure case. Two ideas: (a) on --force without --no-prompt, prompt with the version count and default_version name being deleted; (b) longer-term, implement --force as "upload new non-default version → repoint default → delete old versions" so an upload failure leaves the original intact.

- SKILL.md license, compatibility, allowed_tools are silently dropped on upload — SkillInlineContent (models.go:51-54) exposes these fields on the wire, but ParseSkillMd (skill_md.go:31-87) only extracts name / description / metadata, and neither runFileMd nor buildInlineContent copies them through. Round-trip a SKILL.md and these fields disappear, even though README/AGENTS claim agentskills.io conformance. Either extend ParseSkillMd + propagate (recommended), or drop the fields from SkillInlineContent until they're plumbed end-to-end. Worth a round-trip test either way.

- --set-default-version is not pre-flighted — UpdateSkillDefaultVersion blindly POSTs {default_version: X}. With no client-side GetSkillVersion first, "skill not found" / "version not found" / "version exists but half-deleted" can all surface as a generic server 400. A short pre-flight call + a CodeSkillVersionNotFound error code would let users distinguish these cases. (Also no test for the error path.)

- update inline validation is split across two stages with a misleading error — validateFlags accepts at least one of --description / --instructions, but buildInlineContent requires both non-empty. Running azd ai skill update foo --description "new" passes the first gate, then fails at build time with "update requires non-empty instructions" — without ever explaining that update needs both. Help text reads (--description / --instructions) where the / looks like "or". Easiest fix: tighten validateFlags to require both when inline mode is used; update help to (--description AND --instructions).

- The gzip-tar branch in archive.go is dead under the new content-type check — downloadContent (client.go:425-432) hard-rejects anything that isn't application/zip, but SafeExtract / DetectArchiveFormat still carry an ArchiveTarGz branch with a comment claiming "resilience". That branch can never trigger. Either restore Accept: application/zip, application/gzip on the download (real resilience), or delete ArchiveTarGz + the magic-byte sniff and call zip.NewReader directly. Pick one, drop the other.

Low / cleanup

- client.go package-level downloadByteCap is a mutable global mutated by tests. Safe today only because nothing calls t.Parallel(). Promote to a Client field with a WithMaxDownloadBytes option.

- skill_md.go SkillMdFileName constant is dead after c040787 removed the materialize path. Delete.

- printCreateResult (skill_create.go) and the analogous post-update GET in skill_update.go silently swallow the follow-up GetSkill error. A fmt.Fprintf(os.Stderr, "Warning: ... follow-up GET failed: %v\n", err) would surface transient permission/network issues without failing the command.

- skill_download.go accepts --version "" / --version "   " and silently downloads the default version. strings.TrimSpace + reject empty would catch unset-$VER scripting bugs.

- Worth adding a test that asserts update <name> --description "x" (no --instructions) is rejected — documents the contract once you decide how to fix https://github.com/Azure/azure-dev/pull/6.

- --version flag help on download could note "must exist on the server" so users don't expect arbitrary strings.

trangevi:
## Motivation

Issue #8366 tracked several follow-up items from the initial `azd ai skill` extension PR -- silent data loss on SKILL.md round-trips, dead code paths, test-unsafe mutable globals, and missing validation. This PR addresses the code-level fixes that don't require service-side coordination.

## Approach

Six focused fixes across the `azure.ai.skills` extension:

- **SKILL.md field propagation**: `ParseSkillMd` now extracts `license`, `compatibility`, and `allowed_tools` from front matter and both `create --file` and `update --file` flows propagate them into `SkillInlineContent`. Previously these fields were silently dropped on upload despite being declared in the wire model.

- **Dead tar.gz code removal**: The download endpoint always returns `application/zip` and the client hard-rejects other content types, making the `ArchiveTarGz` extraction branch unreachable. Replaced with a simple `isZipMagic` check and removed ~55 lines of dead code plus unused imports.

- **`downloadByteCap` promoted to instance field**: The package-level mutable global (only safe because tests never ran in parallel) is now a `Client` struct field initialized to `MaxDownloadBytes`, with a `WithMaxDownloadBytes` method for test overrides.

- **Stderr warnings for swallowed errors**: `printCreateResult` and `printUpdateResult` now emit a warning when the follow-up `GetSkill` call fails, instead of silently falling back to the version envelope.

- **Whitespace `--version` rejection**: `azd ai skill download my-skill --version "   "` now returns a clear validation error instead of silently downloading the default version.

- **New tests**: Covers inline update requiring both flags, `ParseSkillMd` field extraction, download version validation, and non-zip archive rejection.

## Items tracked separately

The following higher-impact items from #8366 need design decisions or service-team coordination and are not addressed here:

- `CreateVersionFromZip` memory buffering (io.Pipe streaming refactor)
- POST `/skills/{name}/versions` idempotency (needs Idempotency-Key header support)
- `--force` delete-then-create rollback safety (UX design decision)
- `--set-default-version` pre-flight check (error code coordination)

Fixes: #8366

### DIFF
--- a/cli/azd/extensions/azure.ai.skills/internal/cmd/skill_create.go
+++ b/cli/azd/extensions/azure.ai.skills/internal/cmd/skill_create.go
@@ -150,9 +150,12 @@ func (a *createAction) runFileMd(ctx context.Context, client *skill_api.Client) 
 
 	version, err := client.CreateVersionInline(ctx, a.flags.name, skill_api.CreateVersionRequest{
 		InlineContent: &skill_api.SkillInlineContent{
-			Description:  parsed.Description,
-			Instructions: parsed.Instructions,
-			Metadata:     parsed.Metadata,
+			Description:   parsed.Description,
+			Instructions:  parsed.Instructions,
+			Metadata:      parsed.Metadata,
+			License:       parsed.License,
+			Compatibility: parsed.Compatibility,
+			AllowedTools:  parsed.AllowedTools,
 		},
 		Default: true,
 	})
@@ -276,6 +279,7 @@ func (a *createAction) printCreateResult(ctx context.Context, client *skill_api.
 	if err != nil {
 		// Don't fail the create just because the follow-up GET failed; fall
 		// back to printing the version envelope instead.
+		fmt.Fprintf(os.Stderr, "Warning: could not fetch skill details: %v\n", err)
 		return printSkillVersionDetail(version, outputTable)
 	}
 	return printSkillDetail(skill, outputTable)
--- a/cli/azd/extensions/azure.ai.skills/internal/cmd/skill_download.go
+++ b/cli/azd/extensions/azure.ai.skills/internal/cmd/skill_download.go
@@ -28,6 +28,7 @@ type downloadFlags struct {
 	projectEndpoint string
 
 	outputDirSet bool
+	versionSet   bool
 }
 
 type downloadAction struct{ flags *downloadFlags }
@@ -47,9 +48,10 @@ func (a *downloadAction) Run(ctx context.Context) error {
 		return err
 	}
 
-	// Reject whitespace-only --version to catch scripting bugs like
-	// --version "$UNSET_VAR" that would silently download the default.
-	if a.flags.version != "" && strings.TrimSpace(a.flags.version) == "" {
+	// Reject explicitly-provided but empty/whitespace --version to catch
+	// scripting bugs like --version "$UNSET_VAR" that would otherwise
+	// silently download the default version.
+	if a.flags.versionSet && strings.TrimSpace(a.flags.version) == "" {
 		return exterrors.Validation(
 			exterrors.CodeInvalidParameter,
 			"--version must not be empty or whitespace-only",
@@ -234,6 +236,7 @@ total uncompressed size.`,
 			flags.name = args[0]
 			flags.output = extCtx.OutputFormat
 			flags.outputDirSet = cmd.Flags().Changed("output-dir")
+			flags.versionSet = cmd.Flags().Changed("version")
 			flags.projectEndpoint, _ = cmd.Flags().GetString("project-endpoint")
 			return action.Run(azdext.WithAccessToken(cmd.Context()))
 		},
--- a/cli/azd/extensions/azure.ai.skills/internal/cmd/skill_download.go
+++ b/cli/azd/extensions/azure.ai.skills/internal/cmd/skill_download.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"strings"
 
 	"azureaiskills/internal/exterrors"
 	"azureaiskills/internal/pkg/skill_api"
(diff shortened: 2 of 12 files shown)


jongio (commented):
Clean, well-scoped PR that addresses real correctness issues from #8366. Six focused fixes, each with targeted tests. A few observations:

**Field propagation** (`skill_md.go`, `skill_create.go`, `skill_update.go`): The `License`, `Compatibility`, and `AllowedTools` fields were already declared in `SkillInlineContent` but silently dropped during SKILL.md round-trips. Now properly wired through both create and update flows. The new `frontMatterStringSlice` helper reuses `frontMatterString` for consistent error reporting per element.

**Dead code removal** (`archive.go`): Removing ~55 lines of unreachable `tar.gz` support is the right call since the download endpoint always returns `application/zip` and the client already hard-rejects other content types. The replacement `isZipMagic` check is simple and sufficient.

**Global-to-instance migration** (`client.go`): Moving `downloadByteCap` to a `Client` struct field with `WithMaxDownloadBytes` for test overrides eliminates shared mutable state. Tests now get proper isolation without `t.Cleanup` gymnastics on a package global.

**Version validation** (`skill_download.go`): The `versionSet` + `TrimSpace` guard catches both whitespace-only and empty-string cases when the flag is explicitly provided. Good use of Cobra's `cmd.Flags().Changed()` to distinguish "flag absent" from "flag present with empty value".

**Note on the existing Copilot reviewer comment**: The concern about `--version "$UNSET_VAR"` expanding to empty string is actually handled by this PR. `cmd.Flags().Changed("version")` returns true when the flag appears on the command line regardless of value, and `TestDownloadAction_RejectsEmptyVersionWhenFlagSet` explicitly validates this path.

No blocking issues. CI is green across all platforms.

huimiu (approved):

Issue #8786, opened 2026-06-23
UX: azd tool install/upgrade/tool/uninstall list for skills


hemarina:
Could we review ux for azd tool list, azd tool install/upgrade for skills and uninstall for tools?

For azd tool list, if skill is installed in one of the host, copilot or cluade, it will shows as installed but not clearly indicate which host it is.

Related to https://github.com/Azure/azure-dev/pull/8704, https://github.com/Azure/azure-dev/pull/8794

hyoshis:
Updated design is at https://www.figma.com/design/3Fn9kwKZfODgwoq7H3yMSx/Azd-tool-commands?node-id=36-170&t=3EY3m0pop58LWAov-1

PR #8794, opened 2026-06-24
feat: add `azd tool uninstall` command with per-host skill support

### LINKED ISSUE #8716: support `azd tool uninstall`
We should add uninstall command. Currently we have:
```
Usage
  azd tool [command]

Available Commands
  check         : Check for tool updates.
  install       : Install specified tools.
  list          : List all tools with status.
  show          : Show details for a specific tool.
  upgrade       : Upgrade installed tools.
```

hemarina:
Fixes #8716

## Summary

Adds the missing `azd tool uninstall` command to the `azd tool` group, completing the install / upgrade / **uninstall** lifecycle. The command removes installed tools using the same mechanism each tool was installed through, and adds per-host support for skills.

```
azd tool [command]

Available Commands
  check         : Check for tool updates.
  install       : Install specified tools.
  list          : List all tools with status.
  show          : Show details for a specific tool.
  uninstall     : Uninstall installed tools.   <-- new
  upgrade       : Upgrade installed tools.
```

## Behavior

- `azd tool uninstall <tool...>` — uninstall the named tool(s).
- `--all` — uninstall every installed tool.
- `--dry-run` — preview what would be uninstalled without making changes.
- No args (interactive) — multi-select from installed tools (nothing pre-selected, since uninstall is destructive).
- `--output json` — structured per-tool results.

### Per-host skills (`--host`)

For skill tools (e.g. `azure-skills`, which install through agent CLIs like copilot/claude):

- `azd tool uninstall <skill>` — removes the skill from **every host it is installed through**.
- `azd tool uninstall <skill> --host copilot` — removes it from that host only.
- `azd tool uninstall <skill> --host all` — removes it from every detected host.
- `--host` errors on non-skill tools.

## Uninstall coverage by tool category

| Category | Mechanism |
|---|---|
| CLI (az-cli, github-copilot-cli) | `winget uninstall` / `brew uninstall` / `apt-get remove` / `npm uninstall -g` |
| VS Code extension | `code --uninstall-extension <id>` |
| Server (azure-mcp-server) | `npm uninstall -g` |
| Skill (azure-skills) | per-host `plugin uninstall` (copilot, claude) |
| azd extension (azure.ai.agents) | `azd extension uninstall <id>` (via new `InstallStrategy.UninstallCommand`) |

After running the uninstall command, success is verified by re-detecting that the tool is no longer present (inverted from install, with the same retry/backoff for non-CLI tools).

## Implementation

- **`pkg/tool/installer.go`** — `Uninstall` on the `Installer` interface; `runUninstall` (package-manager / explicit `UninstallCommand` / direct-download) and `runSkillUninstall` (per-host removal + host-scoped verification); `buildUninstallCommand`, `executeUninstall`, `executeUninstallCommand`, `executeDirectDownloadUninstall`, and dedicated error builders.
- **`pkg/tool/manifest.go`** — new `SkillHost.PluginUninstallCommand` (populated for copilot/claude) and `InstallStrategy.UninstallCommand` (populated for the azd extension).
- **`pkg/tool/manager.go`** — `UninstallTools(ctx, ids, opts...)`, forwarding host options; per-tool failures don't abort the batch; dependencies are intentionally left in place.
- **`cmd/tool.go`** — the `uninstall` subcommand, `toolUninstallAction`/flags, `--host` resolution (shared with install/upgrade), and dry-run, reusing the existing `runToolOperation` and telemetry plumbing.

## Telemetry

Reuses the existing `tool.id`/`tool.ids`, `tool.dry_run`, and `tool.install.*` aggregate fields (no new fields/events). Updated `telemetry-coverage_test.go`, `docs/reference/telemetry-data.md`, and `docs/specs/metrics-audit/feature-telemetry-matrix.md`.

## Tests

- Installer: non-skill package-manager uninstall, `UninstallCommand` path, still-detected failure, unsupported strategy, `buildUninstallCommand` table, and skill per-host removal (default / explicit host / not-installed / unknown host).
- Manager: `UninstallTools` delegation, option forwarding, partial-failure batch, unknown id.
- Command: host-option resolution, single-tool delegation + telemetry, dry-run no-op.
- Regenerated `TestUsage` / `TestFigSpec` snapshots.

## Notes

- `azd tool uninstall` is unavailable for tools with only a custom shell `InstallCommand` and no reverse command / package manager; those return a clear "remove manually" message.
- The package-manager uninstall commands run non-interactively, matching the install path; environments needing elevation (winget machine-scope MSI) or passwordless `sudo` (apt) apply equally to install and uninstall.


cli/azd/pkg/tool/installer_test.go
```diff
@@ -2223,3 +2227,850 @@ func TestAvailableSkillHosts_NonSkillToolReturnsNil(t *testing.T) {
...
+	runner.When(func(args exec.RunArgs, _ string) bool {
+		return args.Cmd == "npm" && slices.Contains(args.Args, "--version")
+	}).Respond(exec.RunResult{ExitCode: 0, Stdout: "10.2.0"})
+	runner.When(func(args exec.RunArgs, _ string) bool {
+		return args.Cmd == "npm" && slices.Contains(args.Args, "uninstall")
+	}).RespondFn(func(_ exec.RunArgs) (exec.RunResult, error) {
+		return exec.RunResult{ExitCode: 1}, errors.New("npm boom")
+	})
+
+	inst := NewInstaller(runner, NewPlatformDetector(runner), &mockDetector{})
```
jongio: This mock defaults to `Installed: false`, so when `executeUninstall` fails and `runUninstall` checks `i.detector.DetectTool(ctx, tool)`, it sees the tool as already gone and takes the idempotent success path (`result.Success = true`). The assertions below expect failure.

Replace with `installedDetector("1.0.0")` to simulate a tool that remains present after the failed uninstall command:

```suggestion
	inst := NewInstaller(runner, NewPlatformDetector(runner), installedDetector("1.0.0"))
```

jongio (changes requested):
The production code is solid and follows existing patterns well. One test has a mock setup bug that causes CI failures on all platforms.

`TestUninstall_NonSkill_CommandFails` uses `&mockDetector{}` which defaults to `Installed: false`. After the npm command fails, `runUninstall` checks whether the tool is still detected. Since the mock reports it as not installed, the idempotent success path fires (`result.Success = true`) instead of the intended error path. The test then fails its `assert.False(t, result.Success)` assertion.

Fix: use `installedDetector("1.0.0")` so the tool is still detected after the failed command, triggering the `packageManagerUninstallFailedError` path the test intends to exercise.

cli/azd/cmd/tool.go
```diff
@@ -1202,6 +1214,290 @@ func (a *toolUpgradeAction) dryRun(
...
+) (*actions.ActionResult, error) {
+	rows := make([]toolDryRunItem, 0, len(tools))
+
+	for _, t := range tools {
+		status, detectErr := a.manager.DetectTool(ctx, t.Id)
+		if detectErr != nil {
+			return nil, fmt.Errorf("detecting %s: %w", t.Id, detectErr)
+		}
+
+		action := "uninstall"
```
jongio, line 1423: For skill tools, the dry-run preview shows `uninstall` or `skip (not installed)` without indicating which host(s) the skill would be removed from. Since `--host` targeting is a key feature of skill uninstall, consider showing the target hosts here (e.g., `uninstall from copilot, claude`). Not blocking; just a discoverability improvement for users previewing skill removal.

jongio (commented):
Prior feedback addressed: `TestUninstall_NonSkill_CommandFails` now uses `installedDetector("1.0.0")`, correctly exercising the failure path.

The skill-before-host uninstall ordering in `UninstallTools` is a solid design choice with proper test coverage. One minor observation below on dry-run output for skills.

jongio (commented):
Prior feedback addressed: `TestUninstall_NonSkill_CommandFails` now uses `installedDetector("1.0.0")`, correctly exercising the failure path.

The skill-before-host uninstall ordering in `UninstallTools` is a solid design choice with proper test coverage. One minor observation below on dry-run output for skills.

cli/azd/cmd/tool.go
jongio, line 1423: For skill tools, the dry-run preview shows `uninstall` or `skip (not installed)` without indicating which host(s) the skill would be removed from. Since `--host` targeting is a key feature of skill uninstall, consider showing the target hosts here (e.g., `uninstall from copilot, claude`). Not blocking; just a discoverability improvement for users previewing skill removal.

RickWinter (commented):
One Suggestion and a Nit.

cli/azd/pkg/tool/installer.go
```diff
@@ -245,6 +257,377 @@ func (i *installer) AvailableSkillHosts(tool *ToolDefinition) []string {
...
+) ([]SkillHost, error) {
+	if len(hosts) > 0 {
+		targets := make([]SkillHost, 0, len(hosts))
+		for _, name := range hosts {
+			// A requested host is usable only if it is a configured
+			// SkillHost that is also on PATH.
+			idx := slices.IndexFunc(tool.SkillHosts, func(h SkillHost) bool {
+				return h.Host == name
+			})
+			if idx < 0 || i.commandRunner.ToolInPath(name) != nil {
```
RickWinter: This explicit-host block (the IndexFunc lookup, the `idx < 0 || ToolInPath(name) != nil` guard, and the `supported` list built for the error) is byte-for-byte identical to the install/upgrade path further down in this file (around line 1062). AGENTS.md specifically calls out not copying logic between scopes and extracting a shared helper instead. Can we pull this into something like `resolveExplicitSkillHosts(tool, hosts) ([]SkillHost, error)` and call it from both? Otherwise a future change to the host-availability rule (say, allowing a host that is configured but not yet on PATH) has to be made in two places, and they will drift.

RickWinter (commented):
This adds `azd tool uninstall` as the inverse of `tool install`, threading a new `Uninstall` through the `Installer` interface, `Manager.UninstallTools`, and the cmd action, plus `--all`, `--dry-run`, and `--host` for skills. The shape is right: it mirrors the install/upgrade action structure, reuses the same telemetry emitter, keeps the tool.id vs tool.ids mutual-exclusion discipline, and the idempotent "already gone -> success" handling on every removal path is the correct call. Skill-before-host ordering in `UninstallTools` (so the host CLI is still on PATH to remove the plugin) is a genuinely good catch and is well covered by tests.

The one thing worth resolving before merge is the duplicated explicit-host resolution loop in `installer.go`: `resolveSkillUninstallTargets` copies the existing install-path block verbatim, which is exactly the cross-scope duplication AGENTS.md asks us to hoist. Not blocking, but cheap to fix now while both are fresh.

Nothing here is a security or correctness blocker. Coverage is thorough (manager, installer, and cmd layers all exercised), the docs/telemetry matrices are updated, and the snapshot regeneration is included. Disposition: COMMENT. Address the duplication if you agree; everything else is follow-up at most.

cli/azd/cmd/tool.go
hemarina: Good catch. Let's discuss this in https://github.com/Azure/azure-dev/issues/8786. If we are adding --host for tables, we want to make the behavior consistent and azd tool list or install dry-run should also cover the host.

cli/azd/pkg/tool/installer.go
hemarina: Nice suggestion. Addressed in https://github.com/Azure/azure-dev/pull/8794/commits/dcf00102030422037eb5cf376275838c6c824d78

jongio (commented):
Reviewed commits dcf0010 and a0d57b8 (the 2 new commits since my last review).

The four extracted helpers preserve the behavior of the inline blocks they replace:

- `toolIDUsageAttrs`: same mutual-exclusion between `tool.id`/`tool.ids`, same sorting for batch, same `dry_run` pairing. All 4 call sites (install, install dry-run, upgrade, uninstall) pass through identically.
- `explicitSkillHostTargets`: same `findSkillHost` + PATH check + error message with supported hosts list. Wired into both `resolveSkillTargets` (install/upgrade) and `resolveSkillUninstallTargets` (uninstall).
- `configuredSkillHostsFor`: same installed-to-configured mapping via `findSkillHost`. Used in both the upgrade and uninstall default-host paths.
- `findSkillHost`: wraps the `slices.IndexFunc` lookup that was duplicated across both callers.

No remaining concerns. My earlier CHANGES_REQUESTED finding (test mock in `TestUninstall_NonSkill_CommandFails`) was fixed in 842c109. The dry-run host display observation is tracked in #8786.

jongio (commented):
Reviewed commits dcf0010 and a0d57b8 (the 2 new commits since my last review).

The four extracted helpers preserve the behavior of the inline blocks they replace:

- `toolIDUsageAttrs`: same mutual-exclusion between `tool.id`/`tool.ids`, same sorting for batch, same `dry_run` pairing. All 4 call sites (install, install dry-run, upgrade, uninstall) pass through identically.
- `explicitSkillHostTargets`: same `findSkillHost` + PATH check + error message with supported hosts list. Wired into both `resolveSkillTargets` (install/upgrade) and `resolveSkillUninstallTargets` (uninstall).
- `configuredSkillHostsFor`: same installed-to-configured mapping via `findSkillHost`. Used in both the upgrade and uninstall default-host paths.
- `findSkillHost`: wraps the `slices.IndexFunc` lookup that was duplicated across both callers.

No remaining concerns. My earlier CHANGES_REQUESTED finding (test mock in `TestUninstall_NonSkill_CommandFails`) was fixed in 842c109. The dry-run host display observation is tracked in #8786.

cli/azd/pkg/tool/manifest.go
```diff
@@ -345,22 +355,26 @@ func azureSkills() *ToolDefinition {
...
-				PluginListCommand:     []string{"plugin", "list", "azure@azure-skills"},
-				PluginName:            "azure@azure-skills",
+				Host: "claude",
+				MarketplaceAddCommand: []string{
+					"plugin", "marketplace", "add", "https://github.com/microsoft/azure-skills",
+				},
+				PluginInstallCommand:   []string{"plugin", "install", "azure"},
+				PluginUpdateCommand:    []string{"plugin", "update", "azure@azure-skills"},
+				PluginUninstallCommand: []string{"plugin", "uninstall", "azure@azure-skills"},
+				PluginListCommand:      []string{"plugin", "list", "azure@azure-skills"},
```
JeffreyCA: `claude plugin list` doesn't seem to accept arguments. Would it be better to parse the json from `claude plugin list --json`?

```json
[
  {
    "id": "azure@azure-skills",
    "version": "1.1.73",
    "scope": "user",
    "enabled": true,
    "installPath": "C:\\Users\\<username>\\.claude\\plugins\\cache\\azure-skills\\azure\\1.1.73",
    "installedAt": "2026-06-26T20:20:52.190Z",
    "lastUpdated": "2026-06-26T20:24:16.793Z",
    "mcpServers": {
      "azure": {
        "command": "npx",
        "args": [
          "-y",
          "@azure/mcp@latest",
          "server",
          "start"
        ]
      }
    }
  },
  {
    "id": "gopls-lsp@claude-plugins-official",
    "version": "1.0.0",
    "scope": "user",
    "enabled": true,
    "installPath": "C:\\Users\\<username>\\.claude\\plugins\\cache\\claude-plugins-official\\gopls-lsp\\1.0.0",
    "installedAt": "2026-06-04T17:38:55.076Z",
    "lastUpdated": "2026-06-04T17:38:55.076Z"
  }
]
```

cli/azd/pkg/tool/manifest.go
```diff
@@ -345,22 +355,26 @@ func azureSkills() *ToolDefinition {
...
-				MarketplaceAddCommand: []string{"plugin", "marketplace", "add", "https://github.com/microsoft/azure-skills"},
-				PluginInstallCommand:  []string{"plugin", "install", "azure"},
-				PluginUpdateCommand:   []string{"plugin", "update", "azure@azure-skills"},
-				PluginListCommand:     []string{"plugin", "list", "azure@azure-skills"},
-				PluginName:            "azure@azure-skills",
+				Host: "claude",
+				MarketplaceAddCommand: []string{
+					"plugin", "marketplace", "add", "https://github.com/microsoft/azure-skills",
+				},
+				PluginInstallCommand:   []string{"plugin", "install", "azure"},
```
JeffreyCA: This needs to be `azure@azure-skills` I think, otherwise it installs it as `azure@claude-plugins-official` and results in the below update/uninstall commands not working.

vhvb1989 (approved):

jongio (approved):

cli/azd/pkg/tool/manifest.go
hemarina: Updated. Nice catch.

cli/azd/pkg/tool/manifest.go
hemarina: Updated. Nice catch.

JeffreyCA (approved):

PR #8805, opened 2026-06-24
Fix `azd tool install azure-skills` treating the VS Code Copilot Chat launcher stub as a usable host

### LINKED ISSUE #8797: `azd tool install azure-skills` shows abnormal behavior in no-host installed scenarios on MacOS and Linux for PR#8704.
**Description:**
Under the `no host installed` scenario (neither `copilot` nor `claude` on PATH), running `azd tool install azure-skills` shows two issues on macOS and Linux.
- **MacOs**: If the user is prompted to install Copilot CLI and answers **n**, the final error is: `copilot: Azure Skills was installed via copilot but verification failed`. This is inconsistent with the actual user action (declined installation). It should return a clear error stating "Host not installed; cannot proceed with Azure Skills installation," rather than indicating "Installed but verification failed."
Image

- **Linux VM**: After running `azd tool install azure-skills`, if the user enters `y` when prompted to install Copilot CLI, the flow gets stuck at the PATH/profile confirmation step (no response for a long time and no further input is accepted).
Image

The expected behavior should match Windows: fail with a clear message that no supported host (e.g., Copilot or Claude) was detected, and provide clear installation guidance.
Image

**Steps to Reproduce:**
1. `copilot` and `claude` not on PATH
2. Run `azd tool install azure-skills`.

**Environment:**
- Azd version: azd version 1.26.0-beta.1-pr.6468449 (commit f510f402201899ff16be3a72f553f63cd72b5d99) (stable).
- OS: Linux, MacOS.

**Expected Behavior:**
When no host is installed, `azd tool install azure-skills` should consistently fail or prompt the user to install a host.

@hemarina and @RickWinter for notification.

hemarina:
Fixes #8797.

## Problem
In the "no host installed" scenario, `azd tool install azure-skills` misbehaved on macOS/Linux:
- **macOS:** declining the host install showed a misleading `copilot: Azure Skills was installed via copilot but verification failed`.
- **Linux:** confirming the install got stuck at the launcher`s `~/.profile` PATH prompt.

Windows already failed cleanly with `no supported agentic CLI host found on PATH`.

## Root cause
azd treated *"a file named `copilot` exists on PATH"* (`exec.LookPath`) as *"a working Copilot CLI host."* In VS Code / Codespaces the **GitHub Copilot Chat extension** drops a small `copilot` launcher stub at `~/.vscode-server/.../github.copilot-chat/copilotCli/copilot` and adds it to the integrated-terminal PATH. That stub is not the real CLI — invoking any subcommand (including `--version`) prints `Install GitHub Copilot CLI? [y/N]` and exits 0 without doing anything. azd "installed" the skill through it (a no-op), then verification found nothing → the misleading error. On Windows the stub is not on PATH, so it already failed cleanly.

## Fix
Gate skill-host selection on a **functional probe**, not mere existence:
- Add `SkillHost.BinaryVersionArgs` / `BinaryVersionRegex` (`--version` / `(\d+\.\d+\.\d+)` for copilot and claude).
- Add `installer.hostUsable(ctx, host)` — a host counts only if a **non-interactive `--version` probe returns a real version**. The probe runs with an empty stdin, so a stub that prompts gets EOF and cannot hang the flow. Wired into `pickSkillHost` (now takes `ctx`) and both `resolveSkillTargets` loops.

A non-functional launcher stub now fails the probe → azd returns the same clean "install GitHub Copilot CLI" guidance on all platforms, and never triggers the launcher`s interactive self-install (which is what hung on Linux).

Verified against real output: `GitHub Copilot CLI 1.0.64-3` → matches `1.0.64`; `2.1.178 (Claude Code)` → matches `2.1.178`; the stub`s prompt text → no match (rejected).

## Testing
- New regression test `TestRunSkill_Install_LauncherStubHost_ReturnsInstallGuidance` (asserts clean guidance, not "verification failed", and that no install is attempted through the stub).
- `go build ./...`, `go vet ./...`, `golangci-lint run ./pkg/tool/...` (0 issues), `cspell`, `gofmt` all clean.
- `go test ./pkg/tool/... ./cmd ./internal/agent` pass.
- End-to-end: with a fake `copilot` stub on PATH, `azd tool install azure-skills` returns the clean guidance instead of "verification failed".

No CHANGELOG entry: `azure-skills` is an unreleased feature (added by #8704 in the same `1.26.0-beta.1` cycle).


jongio (commented):
Root cause is correct: `exec.LookPath` returns true for the VS Code Copilot Chat extension's `copilot` stub, which exits 0 but can't actually install skills. The functional-probe approach (run `--version`, match a real version string, pipe empty stdin to prevent hangs) is the right fix for this class of false-positive PATH matches.

A few notes:

- `hostUsable` correctly falls back to an existence-only check when `BinaryVersionArgs`/`BinaryVersionRegex` aren't configured, so hosts without a version probe aren't penalized.
- Context error handling is good: treating a cancelled/timed-out probe as "assume usable" avoids rejecting a real host over a transient issue.
- The regression test (`TestRunSkill_Install_LauncherStubHost_ReturnsInstallGuidance`) covers the exact scenario from the bug report with realistic stub output, and also asserts that no plugin commands are attempted through the stub.
- Existing test updates (adding `!slices.Contains(args.Args, "--version")` guards) are clean. The `mockHostPresence` helper now auto-registers version probes for present hosts, with a comment that later registrations win (confirmed by the mock runner's reverse iteration at `mock_runner.go:43`).

One observation for a possible follow-up: `AvailableSkillHosts` (installer.go:235) still uses the old `ToolInPath` check rather than `hostUsable`. If a user has both the VS Code stub and a real host (e.g. claude) on PATH, the host-selection prompt at `tool.go:676` would list the stub as a valid option. Selecting it would then fail in `resolveSkillTargets` since that path now calls `hostUsable`. Not a functional bug (the install/upgrade flow is fully protected), but the prompt-then-reject UX is slightly inconsistent. Updating `AvailableSkillHosts` to also use `hostUsable` would fix this, though it requires adding `ctx` to the `Installer` interface method signature, so it's reasonable to defer.

jongio (commented):
Incremental review of d5f428f6 (1 new commit since my prior review against 3b163b1).

The second commit closes the gap I flagged: `AvailableSkillHosts` now calls `hostUsable` (functional version probe) instead of `ToolInPath` (existence-only check), so the interactive host picker won't offer a launcher stub that the install path would later reject. The `Installer` interface, implementation, `Manager` pass-through, both `cmd/tool.go` call sites, and all test mocks/call sites are updated consistently.

No new issues from the incremental change. The full PR remains solid: root cause is correctly identified, the functional-probe approach is principled, edge cases (context cancellation, missing version config) are handled, and the regression test covers the reported scenario with realistic stub output.

cli/azd/pkg/tool/installer.go
```diff
@@ -691,18 +696,58 @@ func (i *installer) resolveSkillTargets(
...
+		ctx,
+		exec.NewRunArgs(host.Host, host.BinaryVersionArgs...).
+			WithStdIn(strings.NewReader("")),
+	)
+	// A cancelled/timed-out probe is not evidence the host is a stub; do
+	// not penalize it here (context handling is the caller's concern).
+	if isContextErr(err) {
+		return true
+	}
+	return matchVersion(result.Stdout+result.Stderr, host.BinaryVersionRegex) != ""
```
RickWinter: The probe accepts the host when `matchVersion(result.Stdout+result.Stderr, host.BinaryVersionRegex)` returns non-empty, and `matchVersion` runs `FindStringSubmatch` over the whole combined output, so `(\d+\.\d+\.\d+)` matches a semver appearing anywhere, not a version the host deliberately reported. Today's VS Code stub prints `Install GitHub Copilot CLI?` and a docs URL with no semver, so we are fine, but the guard rests entirely on the stub never emitting a version-shaped token. The moment a stub's banner carries its own extension version, a bundled node version, or a path segment with a build number, we treat it as a real host and land back on the misleading `was installed via copilot but verification failed` path this PR exists to remove. An `ExitCode == 0` guard would not help, since the stub also exits 0, so the discriminator has to live in how specifically we match. Can we tighten this, for example by requiring the version to stand alone on a line rather than appear anywhere in the stream, and add a test where the stub prints incidental version text so we lock the behavior in?

cli/azd/pkg/tool/installer.go
```diff
@@ -691,18 +696,58 @@ func (i *installer) resolveSkillTargets(
...
+// skill — which previously surfaced as the misleading
+// "<skill> was installed via <host> but verification failed".
+//
+// hostUsable runs the host's configured version probe and treats the host
+// as usable only when it reports a real version. The probe runs
+// non-interactively with an empty stdin, so a stub that prompts for input
+// reads EOF and exits instead of hanging. Hosts that do not configure a
+// version probe (BinaryVersionArgs/BinaryVersionRegex) fall back to the
+// existence check.
+func (i *installer) hostUsable(ctx context.Context, host SkillHost) bool {
```
RickWinter, line 1109: `hostUsable` spawns a subprocess on every call, and it now runs from four places: `AvailableSkillHosts`, the `resolveSkillTargets` availability loop, the requested-host check, and `pickSkillHost`. In a single `azd tool install azure-skills` with no `--host`, the same host gets its `--version` shelled out several times over to recompute an answer that cannot change inside one command. It is bounded to the two configured hosts so it is not alarming, but would it be worth memoizing the probe result per host for the lifetime of the call so each host is invoked at most once?

cli/azd/pkg/tool/installer_test.go
```diff
@@ -1157,21 +1159,30 @@ func newSkillTool() *ToolDefinition {
...
 	}
 }
 
 // mockHostPresence wires ToolInPath responses so only the named hosts
 // resolve successfully. Pass an empty slice to mock every host as
-// missing.
+// missing. Present hosts also get a version-probe response (a real semver)
+// so installer.hostUsable treats them as genuine, functional CLIs rather
+// than launcher stubs. Tests that want to simulate a stub register their
+// own later "--version" expectation, which wins (last match).
```
RickWinter: nit: this leans on the per-test stub responder winning over the generic semver responder by last-match precedence. That is a quiet dependency on the mock framework's ordering, and if it ever flips, several tests would silently start treating stubs as usable rather than failing loudly. A short assertion in the stub test that it actually saw the stub response would make it self-defending.

RickWinter (dismissed):
This resolves #8797 by gating skill-host selection on a functional `--version` probe instead of a bare `exec.LookPath` existence check. The root-cause analysis is correct, and routing both the interactive picker and the install path through a single `hostUsable` so they cannot disagree is the right shape. Reusing `matchVersion` and adding `BinaryVersionArgs` / `BinaryVersionRegex` next to the existing `VersionRegex` keeps this consistent with how detection already works in the package, and the existence-only fallback leaves behavior unchanged for hosts that configure no probe.

The one concern worth resolving before merge is how strict the probe actually is at line 734. The match succeeds on any version-shaped token anywhere in the host's combined output, so the fix sits one stray semver in a stub's banner away from regressing to the same `verification failed` path it removes. The repeated probe spawns across the four call sites and the last-match-wins test ordering are lower priority and can follow. I would want the first point either tightened or consciously accepted; nothing else here is blocking.

cli/azd/pkg/tool/installer.go
jongio: Agree on tightening. A concrete approach: anchor to start-of-line with `(?m)^.*?(\d+\.\d+\.\d+)`. Real CLIs emit version prominently on line 1 (`GitHub Copilot CLI 1.0.64-3`, `2.1.178 (Claude Code)`), while incidental build numbers in URLs or paths typically appear mid-line. Combined with a test that embeds a fake semver in the stub's banner text, that would close this gap solidly. Whether this lands in this PR or a quick follow-up seems fine either way since the current matching works correctly for the known stubs.

cli/azd/pkg/tool/installer.go
jongio: The cost is bounded: 2 configured hosts means at most ~8 subprocess calls per command invocation even hitting all 4 call sites. Not alarming, as you noted. If the author wants to address it, a `map[string]bool` field on the installer (populated on first probe per host) would be simple. Not blocking for the bug fix.

cli/azd/pkg/tool/installer_test.go
jongio: Agreed. Something like asserting that the mock runner's recorded calls include the stub's exact response text (rather than just checking the final behavior) would catch a silent ordering flip.

hemarina:
Thanks for the thorough review, @RickWinter and @jongio. Addressed the follow-ups:

**Probe strictness** — `BinaryVersionRegex` is now anchored to each host`s own `--version` banner instead of a bare semver, so only a genuine version line counts and an incidental version-shaped token (a bundled node version, a path build number, a URL) no longer passes:
- copilot: `(?m)^GitHub Copilot CLI\s+v?(\d+\.\d+\.\d+)`
- claude: `(?m)^v?(\d+\.\d+\.\d+)\s+\(Claude Code\)`

Added `TestAzureSkillsHostVersionProbeRegex` covering the real banners (`GitHub Copilot CLI 1.0.64-3`, `2.1.178 (Claude Code)`) plus rejection cases (stub prompt, banner-prefix-without-version, version-not-at-line-start), and strengthened the stub test with an incidental-version banner.

**Repeated probe spawns** — added an installer-field `hostProbe map[string]bool` cache (guarded by `hostProbeMu`) so each on-PATH host`s `--version` is spawned at most once per command. Only on-PATH results are cached: a not-on-PATH result is never cached, so a host installed earlier in the same batch (e.g. `azd tool install github-copilot-cli azure-skills`) is still picked up at run time. Tests: `TestHostUsable_OnPathProbeMemoizedAcrossPasses`, `TestHostUsable_NotOnPathResultNotMemoized`.

**Self-defending stub test** — the launcher-stub test now records and asserts the stub`s `--version` responder actually ran (`stubVersionProbed`), so a mock last-match-precedence flip fails loudly rather than silently treating a stub as usable.

**`AvailableSkillHosts`** — now uses the same functional probe (`hostUsable`) so the interactive host picker and "multiple hosts" guidance reflect actual usability, not just PATH presence.

jongio (commented):
Incremental review of c6e3cc8a (addresses RickWinter's feedback).

The memoization in `hostUsable` stores only on-PATH probe results correctly: the `ToolInPath` early-return at the top skips the cache entirely, so a host installed earlier in the same batch (e.g. `azd tool install github-copilot-cli azure-skills`) is re-probed when its PATH status changes. The lock-probe-lock-store pattern allows benign duplicate probes under concurrency rather than holding the mutex across the subprocess spawn, which is the right tradeoff here.

The `\n` separator between stdout and stderr in `probeOnPathHost` ensures `(?m)^` anchors work when the version banner appears at the start of stderr output. Regex anchoring to each host's specific banner format closes the incidental-semver gap that was identified.

New tests (`TestHostUsable_OnPathProbeMemoizedAcrossPasses`, `TestHostUsable_NotOnPathResultNotMemoized`, `TestRunSkill_Install_StubWithIncidentalVersion_Rejected`, `TestAzureSkillsHostVersionProbeRegex`) lock these behaviors with good coverage of the edge cases.

vhvb1989 (approved):

jongio (approved):

JeffreyCA (approved):

JeffreyCA (approved):

jongio (commented):
Incremental review of the 2 post-merge commits (20a1775, ba63de0).

The merge with main pulled in `azd tool uninstall` (#8794). The post-merge adjustments correctly wire `explicitSkillHostTargets` through `hostUsable` so the uninstall path also rejects launcher stubs. Comment cleanup on `hostUsable` preserves semantics while reducing verbosity. New `TestRunSkill_Install_ExplicitStubHost_Rejected` covers the explicit `--host` + stub scenario with probe-execution assertions.

RickWinter's feedback (regex anchoring, memoization, test ordering) from d5f428f remains addressed by e75ca9e and carried cleanly through the merge.

JeffreyCA (approved):

PR #8809, opened 2026-06-25
feat(agents): move Foundry networking to project service

m5i-work:
## Summary
- move the unshipped Foundry `network:` contract from agent/legacy service hosts to `host: azure.ai.project`
- make provisioning/eject discover the project service and reject misplaced `network:` on `azure.ai.agent`/`microsoft.foundry`
- update schema, docs, tests

## Tests
- `cd cli/azd/extensions/azure.ai.agents && DOTNET_SYSTEM_GLOBALIZATION_INVARIANT=1 go test ./internal/synthesis ./internal/project ./internal/cmd`


vnet-reparent-plan.md
```diff
@@ -0,0 +1,73 @@
+# VNet reparent plan
```
jongio: This planning doc should not be committed to the repo root. The root allowlist is `README.md`, `LICENSE`, `CHANGELOG.md`. Either move it under `cli/azd/extensions/azure.ai.agents/docs/` or remove it from the PR.

cli/azd/extensions/azure.ai.agents/internal/cmd/init_infra.go
```diff
@@ -267,13 +283,17 @@ func findFoundryServiceForEject(raw []byte) (string, error) {
...
-				project.FoundryServiceHosts, matches),
-			"keep a single foundry service per project",
+			fmt.Sprintf("multiple services declare a foundry project host %v (%v); only one is supported",
+				project.FoundryProjectServiceHosts, matches),
+			"keep a single azure.ai.project service per project",
 		)
 	}
 }
 
+func isFoundryNetworkHost(host string) bool {
```
jongio: This is identical to the same function in `internal/project/foundry_provisioning_provider.go`. Since this package already imports `project` (for `FoundryProjectServiceHosts`), export `IsFoundryNetworkHost` from `provisioning_provider.go` and call it here instead.

cli/azd/extensions/azure.ai.agents/internal/synthesis/schema_test.go
```diff
@@ -15,9 +15,9 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-// schemaPath is the editor-tooling JSON schema for the Foundry service body,
+// schemaPath is the editor-tooling JSON schema for the Foundry project service body,
 // resolved from this package directory.
-const schemaPath = "../../schemas/microsoft.foundry.json"
+const schemaPath = "../../../azure.ai.projects/schemas/azure.ai.project.json"
```
jongio, line 22: This path crosses into the `azure.ai.projects` extension. If these extensions are ever built or tested independently (separate `go test` invocations), this test will fail with a missing file error. Consider copying the schema into a `testdata/` directory, or adding a `t.Skip` guard when the file is absent.

jongio (changes requested):
Moving `network:` to the project service boundary is the right call since it is unshipped and the split-resource shape is already established. Test coverage for the new misplaced-network detection is solid.

Two items to address before merge:

1. `vnet-reparent-plan.md` is committed to the repo root. Planning documents should not live at root per repo conventions. Move it under `cli/azd/extensions/azure.ai.agents/docs/` if needed long-term, or remove it from the PR if it only tracks the implementation approach.

2. `isFoundryNetworkHost` is defined identically in both `internal/cmd/init_infra.go` and `internal/project/foundry_provisioning_provider.go`. Since `cmd` already imports `project`, export it once from `provisioning_provider.go` and call `project.IsFoundryNetworkHost(...)` from `init_infra.go`.

Minor: `schema_test.go` now reaches across extension boundaries to read `azure.ai.projects/schemas/azure.ai.project.json`. If these extensions are ever built or tested in isolation, this test will break. Consider a `testdata` copy or a `t.Skip` guard when the file is absent.

m5i-work:
Addressed review feedback:

- Removed `vnet-reparent-plan.md` from the PR entirely (not moved/committed).
- Exported `project.IsFoundryNetworkHost` from `provisioning_provider.go` and reused it from both provider and eject paths.
- Added a missing-file `t.Skip` guard for the cross-extension project schema invariant test.

Validation:
```bash
DOTNET_SYSTEM_GLOBALIZATION_INVARIANT=1 go test ./internal/synthesis ./internal/project ./internal/cmd
```

vnet-reparent-plan.md
m5i-work: Addressed by removing `vnet-reparent-plan.md` from the PR entirely. I amended the branch so the plan doc is no longer committed at root (or anywhere else in the PR).

cli/azd/extensions/azure.ai.agents/internal/cmd/init_infra.go
m5i-work: Addressed. I exported `project.IsFoundryNetworkHost` from `internal/project/provisioning_provider.go` and changed `internal/cmd/init_infra.go` to call `project.IsFoundryNetworkHost(...)`, removing the duplicate helper.

cli/azd/extensions/azure.ai.agents/internal/synthesis/schema_test.go
m5i-work: Addressed with the suggested guard. `TestSchema_NetworkStructuralInvariants` now skips when `../../../azure.ai.projects/schemas/azure.ai.project.json` is absent, so isolated extension test runs do not fail due to the cross-extension schema file being missing.

jongio (commented):
All three items from my previous review are addressed:

1. `vnet-reparent-plan.md` removed from the PR.
2. `IsFoundryNetworkHost` is now exported once from `provisioning_provider.go`; `init_infra.go` calls `project.IsFoundryNetworkHost(...)`.
3. `schema_test.go` skips gracefully when the cross-extension schema file is absent.

The misplaced-network detection logic, updated docs, schema move to `azure.ai.project.json`, and test coverage all look correct. No new concerns.

jongio (commented):
My earlier changes-requested items are all resolved (plan doc removed, IsFoundryNetworkHost exported once and reused, schema_test skips gracefully). Re-reviewed the full diff: misplaced-network detection logic, schema relocation to azure.ai.project.json, updated docs, and test coverage all look correct. No new concerns.

m5i-work:
Added the #4 schema fix as a normal follow-up commit (no force push):

- `azure.ai.project` schema now rejects `network.agentSubnet` and `network.isolationMode` when specified together.
- Added schema invariant coverage to ensure this stays aligned with runtime behavior.

Validation:
```bash
DOTNET_SYSTEM_GLOBALIZATION_INVARIANT=1 go test ./internal/synthesis ./internal/project ./internal/cmd
```

jongio (commented):
Incremental review (history rewritten since my last review against 19d8234).

All items from my CHANGES_REQUESTED are resolved in the current HEAD:
- Plan doc removed from root
- `IsFoundryNetworkHost` exported once and reused across both eject and provision paths
- Schema `allOf` constraint added to reject `agentSubnet` + `isolationMode` together, with test coverage

The reparent itself is clean: `findFoundryProjectService` correctly narrows to `azure.ai.project`, the misplaced-network detection gives actionable guidance, `deriveIncludeAcr` handles the split model by scanning sibling `azure.ai.agent` services, and the `t.Skipf` guard for the cross-extension schema path is appropriate.

One non-blocking note: `findFoundryServiceForEject` and `findFoundryProjectService` still share ~30 lines of scan-match-detect logic (differing only in error codes for telemetry). If a third caller appears, extracting a shared `scanProjectServices(raw []byte) (matches, misplaced []string, error)` helper would reduce the surface area. Fine as-is for two callers.

m5i-work (commented):
Follow-up review after the new commits. A few issues still look actionable:

1. **`includeAcr` is now true for BYO/prebuilt-image agents that should skip ACR**

   `internal/synthesis/synthesizer.go` now derives ACR from any `azure.ai.agent` service with a non-nil `docker:` block:

   ```go
   if service.Host == "azure.ai.agent" && service.Docker != nil {
       return true
   }
   ```

   But init writes `docker:` for hosted container agents even when a prebuilt image is supplied (`--image` / `image:`), while the existing skip-ACR contract says prebuilt-image projects should not create ACR. A private-networking user following the BYO image guidance can therefore get unexpected ACR provisioning. Consider including the service/agent `image` field in the scan and only setting `includeAcr` for Docker-backed agents that do not use a prebuilt image.

2. **Legacy `microsoft.foundry` shape is still advertised but no longer provisionable**

   Provider discovery now accepts only `host: azure.ai.project`, so an `infra.provider: microsoft.foundry` project with only a legacy `host: microsoft.foundry` service now fails with “no service in azure.yaml has host in [azure.ai.project]”. However, the root schema still describes `microsoft.foundry` as a legacy alias and `schemas/examples/simple.azure.yaml` still uses it. If dropping legacy provisioning is intentional, these schema comments/examples should be updated; otherwise the provider needs a fallback for legacy non-`network:` shapes.

3. **`microsoft.foundry.json` still has a branch `$id`**

   The schema `$id` still points at `https://raw.githubusercontent.com/Azure/azure-dev/huimiu/foundry-azure-yaml/...` while the root schemas reference it at the `main` URL. Relative refs such as `Deployment.json` resolve against `$id`, so editor/schema validation can fetch refs from the stale branch path. Please update the `$id` to the matching `main` raw URL.

m5i-work:
Addressed the latest follow-up review in a normal commit (no force push):

- `includeAcr` now skips Docker-backed entries when a prebuilt `image` is present, for both split `azure.ai.agent` services and legacy inline `agents[]`. Added tests for both cases.
- Updated Foundry schema examples to use the split `azure.ai.project` / `azure.ai.agent` shape instead of advertising `host: microsoft.foundry`.
- Updated the root schema comments for `microsoft.foundry` to clarify it is schema-only compatibility for old files; new provisioning uses `azure.ai.project`.
- Updated `microsoft.foundry.json` `$id` to the `main` raw URL so relative refs resolve consistently with the root schema.

Validation:
```bash
DOTNET_SYSTEM_GLOBALIZATION_INVARIANT=1 go test ./internal/synthesis ./internal/project ./internal/cmd
python3 -m json.tool cli/azd/extensions/azure.ai.agents/schemas/microsoft.foundry.json >/dev/null
python3 -m json.tool schemas/v1.0/azure.yaml.json >/dev/null
python3 -m json.tool schemas/alpha/azure.yaml.json >/dev/null
```

m5i-work:
Addressed the remaining pre-split compatibility finding in a normal follow-up commit:

- Provisioning/eject discovery now prefers a single `azure.ai.project` service, but falls back to exactly one legacy provisioning host (`azure.ai.agent` / `microsoft.foundry`) when no project service exists.
- `network:` remains project-only: misplaced `network:` on legacy hosts is still rejected before fallback selection.
- Updated the changelog/root schema wording to clarify this is non-network pre-split compatibility.
- Added discovery tests for pre-split fallback, project-over-legacy precedence, and multiple legacy services.

Validation:
```bash
DOTNET_SYSTEM_GLOBALIZATION_INVARIANT=1 go test ./internal/synthesis ./internal/project ./internal/cmd
python3 -m json.tool schemas/v1.0/azure.yaml.json >/dev/null
python3 -m json.tool schemas/alpha/azure.yaml.json >/dev/null
```

m5i-work:
Addressed the complex example finding in a normal follow-up commit:

- Rewrote `schemas/examples/complex.azure.yaml` so each Foundry resource is a sibling service (`azure.ai.project`, `azure.ai.connection`, `azure.ai.toolbox`, `azure.ai.skill`, `azure.ai.agent`, `azure.ai.routine`).
- Wired resources with `uses:` and moved agent definitions to service-level fields instead of nested `agents:` entries.
- Kept project deployments + `network:` on the `azure.ai.project` service.

Validation:
```bash
DOTNET_SYSTEM_GLOBALIZATION_INVARIANT=1 go test ./internal/synthesis ./internal/project ./internal/cmd
```

jongio (commented):
Clearing prior changes-requested state (automation bug). Original review findings still apply as comments.

PR #8810, opened 2026-06-25
training: show Foundry Portal Uri in 'azd ai training job show'

### LINKED ISSUE #8811: [ext/azure.ai.training] Add Foundry portal uri of the job in show command
Add Foundry portal uri of the job in show command to match the SDK experience.

saanikaguptamicrosoft:
Adds a 'Foundry Portal Uri:' line to the job show output, sourced from properties.services['Studio'].endpoint via utils.ServiceEndpoint.
image

### DIFF
--- a/cli/azd/extensions/azure.ai.training/internal/cmd/job_show_test.go
+++ b/cli/azd/extensions/azure.ai.training/internal/cmd/job_show_test.go
@@ -1,0 +1,107 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package cmd
+
+import (
+	"io"
+	"os"
+	"strings"
+	"testing"
+
+	"azure.ai.training/pkg/models"
+)
+
+// captureStdout runs fn while redirecting os.Stdout to a pipe, then returns
+// everything fn wrote. Used because printJobDetails writes directly to stdout.
+func captureStdout(t *testing.T, fn func()) string {
+	t.Helper()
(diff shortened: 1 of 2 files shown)


achauhan-scc (approved):

jongio (commented):
One minor robustness note on the test helper. The production change is clean: single line using existing utilities, correct 'URI' casing consistent with line 665, and CI passes across all platforms.

cli/azd/extensions/azure.ai.training/internal/cmd/job_show_test.go
```diff
@@ -0,0 +1,126 @@
...
+
+	t.Cleanup(func() {
+		os.Stdout = orig
+		_ = w.Close()
+	})
+
+	func() {
+		defer func() { _ = w.Close() }()
+		fn()
+	}()
```
jongio, line 48: After the inner func returns and `w.Close()` fires, `os.Stdout` still points to the closed pipe until `t.Cleanup` runs at test exit. If a future test author adds any `fmt.Printf` debug lines (or another `captureStdout` call) after this helper returns, those writes silently fail.

Consider restoring stdout immediately after the inner func:

```go
func() {
    defer func() { _ = w.Close() }()
    fn()
}()

os.Stdout = orig
```

The `t.Cleanup` then serves only as a panic safety net. Not blocking since the current tests don't write to stdout after capture.

achauhan-scc (approved):

PR #8814, opened 2026-06-25
use ci-build.ps1 -BuildRecordMode and add workflow_dispatch

### LINKED ISSUE #8801: Follow-ups: use ci-build.ps1 -BuildRecordMode and review workflow_dispatch for azure.ai.agents test workflow
Follow-ups from @danieljurek's review of #8754 (azure.ai.agents PR-gate test workflow `.github/workflows/test-ext-azure-ai-agents.yml`). Both are non-blocking; tracked here so #8754 can merge.

## 1. Build the record-tagged binary via `ci-build.ps1 -BuildRecordMode` (canonical script)

> "Is there support for a recording flag in ci-build.ps1?" — @danieljurek

Yes — `cli/azd/extensions/azure.ai.agents/ci-build.ps1` already has a `-BuildRecordMode` switch that appends `,record` to the build tags and emits a `<output>-record` binary. The current workflow instead builds via `azd x build` with `GOFLAGS=-tags=record` (which goes through `build.sh`, not `ci-build.ps1`).

**Action:** switch the workflow's record-tagged build to the canonical `ci-build.ps1 -BuildRecordMode` so CI matches the release build flags (`-trimpath`, `-buildmode=pie`, `-tags=cfi,cfg,osusergo`, version `-ldflags`) and there's a single source of truth for how the extension is built.

This also lines up with @danieljurek's broader approval suggestions for scaling this pattern to more extensions:
- Use **one** workflow to test all extensions, selecting which to test from the files changed in the diff.
- Use each extension's `ci-build.ps1` (with the tags/params it needs) rather than ad-hoc `go build`/`azd x build` invocations.
- Consider `ci-test.ps1` where applicable (may not fit here since the Go tests run from `cli/azd`, outside the extension dir).

## 2. Review the workflow trigger — add `workflow_dispatch`?

> "Should this also run on workflow_dispatch?" — @danieljurek

The workflow currently triggers only on `pull_request` (path-filtered) against `main`. Evaluate adding `workflow_dispatch` (and possibly a `schedule`) so the suite can be run manually from the Actions UI for debugging / ad-hoc validation.

---
Source: #8754 review comments by @danieljurek.

v1212:
## Summary

Switch the extension build from `GOFLAGS=-tags=record` with `azd x build` to the canonical `ci-build.ps1 -BuildRecordMode` script, which includes security flags (trimpath, pie, cfi, cfg, osusergo) matching release builds.

### Changes

1. **Build**: Use `ci-build.ps1 -BuildRecordMode` — produces both production and record-tagged binaries with standardized security flags
2. **Tier 0**: Now runs against the **production** extension binary (no HTTP, no record tag needed)
3. **Tier 1**: Swaps in the **record-tagged** binary for HTTP playback
4. **Trigger**: Add `workflow_dispatch` for manual/ad-hoc runs
5. **Concurrency**: Fix group key to avoid manual runs cancelling each other

### Scope

#8814 addresses the two immediate items (ci-build.ps1 -BuildRecordMode + workflow_dispatch) — these are must-haves for hosted agent GA and time-sensitive. The scaling improvements for testing more extensions (unified workflow, ci-test.ps1) involve larger changes and need at least a second extension to validate against, to be done later as follow-ups.

Closes #8801

### DIFF
--- a/.github/workflows/test-ext-azure-ai-agents.yml
+++ b/.github/workflows/test-ext-azure-ai-agents.yml
@@ -49,7 +49,7 @@ jobs:
           # Use the canonical ci-build.ps1 which includes security flags
           # (trimpath, pie, cfi, cfg, osusergo) matching release builds.
           # -BuildRecordMode produces both a production and a record-tagged binary.
-          pwsh -File ci-build.ps1 -BuildRecordMode -OutputFileName azure-ai-agents-linux-amd64
+          pwsh -File ci-build.ps1 -BuildRecordMode -OutputFileName azure-ai-agents-linux-amd64 -Version $(cat version.txt | tr -d '\r\n')
           # Move production binary to bin/ where azd x pack expects it.
           mkdir -p bin
           mv azure-ai-agents-linux-amd64 bin/
@@ -73,8 +73,8 @@ jobs:
         run: |
           # Overwrite the installed production binary with the record-tagged
           # binary so Tier 1 tests can intercept/replay HTTP traffic.
-          find ~/.azd/extensions/azure.ai.agents/ -type f -executable \
-            -exec cp "${{ github.workspace }}/cli/azd/extensions/azure.ai.agents/azure-ai-agents-linux-amd64-record" {} \;
+          cp "${{ github.workspace }}/cli/azd/extensions/azure.ai.agents/azure-ai-agents-linux-amd64-record" \
+            ~/.azd/extensions/azure.ai.agents/azure-ai-agents-linux-amd64
 
       - name: Run Tier 1 tests (playback, record extension)
         working-directory: cli/azd
--- a/.github/workflows/test-ext-azure-ai-agents.yml
+++ b/.github/workflows/test-ext-azure-ai-agents.yml
@@ -7,9 +7,10 @@ on:
       - "cli/azd/test/functional/ai_agents/**"
       - ".github/workflows/test-ext-azure-ai-agents.yml"
     branches: [main]
+  workflow_dispatch:
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.event.pull_request.number }}
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.run_id }}
   cancel-in-progress: true
 
 permissions:
@@ -42,20 +43,25 @@ jobs:
           # Pre-install the inspector dependency so the local install can resolve it.
           azd extension install azure.ai.inspector --no-prompt
 
-      - name: Build, pack & register extension (record-tagged)
+      - name: Build extension (ci-build.ps1)
         working-directory: cli/azd/extensions/azure.ai.agents
-        env:
-          GOFLAGS: -tags=record
         run: |
-          # build.sh inherits GOFLAGS, so go build gets -tags=record automatically.
-          azd x build --skip-install
+          # Use the canonical ci-build.ps1 which includes security flags
+          # (trimpath, pie, cfi, cfg, osusergo) matching release builds.
+          # -BuildRecordMode produces both a production and a record-tagged binary.
+          pwsh -File ci-build.ps1 -BuildRecordMode -OutputFileName azure-ai-agents-linux-amd64
+          # Move production binary to bin/ where azd x pack expects it.
+          mkdir -p bin
+          mv azure-ai-agents-linux-amd64 bin/
+
+      - name: Pack, publish & install extension (production)
+        working-directory: cli/azd/extensions/azure.ai.agents
+        run: |
           azd x pack
           azd x publish
+          azd extension install azure.ai.agents --source local --force --no-prompt
 
-      - name: Install extension from local registry
-        run: azd extension install azure.ai.agents --source local --force --no-prompt
-
-      - name: Run Tier 0 tests (offline)
+      - name: Run Tier 0 tests (offline, production extension)
         working-directory: cli/azd
         env:
           CLI_TEST_AZD_PATH: ${{ github.workspace }}/cli/azd/azd-record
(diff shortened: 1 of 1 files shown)


.github/workflows/test-ext-azure-ai-agents.yml
```diff
@@ -42,28 +43,40 @@ jobs:
...
           azd extension install azure.ai.inspector --no-prompt
 
-      - name: Build, pack & register extension (record-tagged)
+      - name: Build extension (ci-build.ps1)
+        working-directory: cli/azd/extensions/azure.ai.agents
+        run: |
+          # Use the canonical ci-build.ps1 which includes security flags
+          # (trimpath, pie, cfi, cfg, osusergo) matching release builds.
+          # -BuildRecordMode produces both a production and a record-tagged binary.
+          pwsh -File ci-build.ps1 -BuildRecordMode -OutputFileName azure-ai-agents-linux-amd64
```
jongio: `ci-build.ps1` defaults `-Version` from `$PSScriptRoot/../version.txt`, which resolves to `cli/azd/extensions/version.txt` (doesn't exist). The extension's own `version.txt` sits in the current working directory. Pass it explicitly:

```suggestion
          pwsh -File ci-build.ps1 -BuildRecordMode -OutputFileName azure-ai-agents-linux-amd64 -Version $(cat version.txt | tr -d '\r\n')
```

Without this, the binary gets an empty version string and the Tier 0 `Version` test will fail.

.github/workflows/test-ext-azure-ai-agents.yml
```diff
@@ -42,28 +43,40 @@ jobs:
...
         run: |
           go test -tags=record -run "Test_AIAgent_(Version|Help|Init_NoPrompt_MissingFlags|Doctor_Help)" \
             ./test/functional/ai_agents/ -v -timeout 5m
 
-      - name: Run Tier 1 tests (playback)
+      - name: Swap extension to record binary for Tier 1
+        run: |
+          # Overwrite the installed production binary with the record-tagged
+          # binary so Tier 1 tests can intercept/replay HTTP traffic.
+          find ~/.azd/extensions/azure.ai.agents/ -type f -executable \
```
jongio: Nit: `find -type f -executable` will overwrite every executable it finds. If the install directory ever contains a wrapper script or additional binary, they'd all be replaced silently. Consider targeting the known binary name directly if `azd extension install` puts it at a predictable path.

jongio (changes requested):
The approach is solid: using ci-build.ps1 for consistent security flags (trimpath, pie, cfi, cfg, osusergo) and cleanly separating the production binary (Tier 0) from the record-tagged binary (Tier 1). The workflow_dispatch and concurrency group fixes are correct.

One blocking issue: the ci-build.ps1 invocation doesn't pass -Version, and the script's default path resolves to a non-existent file. This would produce a binary with an empty version string, failing the Tier 0 Version test.

.github/workflows/test-ext-azure-ai-agents.yml
v1212: Fixed — passing  -Version $(cat version.txt)  explicitly now. thanks

jongio (commented):
Incremental review of fix commit (9d9b3f7). Both prior concerns resolved:

1. Version is now passed explicitly via `cat version.txt`. Confirmed this is necessary because ci-build.ps1 defaults to `$PSScriptRoot/../version.txt` which resolves to a non-existent `cli/azd/extensions/version.txt` path.
2. Direct `cp` to the known binary path replaces the `find -executable` approach, avoiding accidental overwrites if the install directory layout changes.

No new issues found in the incremental diff. My earlier CHANGES_REQUESTED can be cleared.

.github/workflows/test-ext-azure-ai-agents.yml
v1212: Fixed — using the known binary path directly instead of find.

v1212:
@jongio thanks for quick review and comments, I will get the team for more review and then help to approve and merge.

jongio (commented):
Re-reviewed against HEAD (9d9b3f7). The ci-build.ps1 -BuildRecordMode integration is correct: version is passed explicitly from the extension's version.txt, the production binary goes to bin/ for azd x pack, and the record binary stays in the working directory for the Tier 1 swap. The security flags (trimpath, pie, cfi, cfg, osusergo) now match release builds, and the tier separation (production for Tier 0, record for Tier 1) properly validates both binary variants.

My prior CHANGES_REQUESTED concerns (missing -Version, fragile find -executable) are both resolved. The concurrency group fallback to github.run_id correctly prevents manual runs from cancelling each other.

jongio (approved):

.github/workflows/test-ext-azure-ai-agents.yml
```diff
@@ -42,28 +43,40 @@ jobs:
...
           azd extension install azure.ai.inspector --no-prompt
 
-      - name: Build, pack & register extension (record-tagged)
+      - name: Build extension (ci-build.ps1)
+        working-directory: cli/azd/extensions/azure.ai.agents
+        run: |
+          # Use the canonical ci-build.ps1 which includes security flags
+          # (trimpath, pie, cfi, cfg, osusergo) matching release builds.
+          # -BuildRecordMode produces both a production and a record-tagged binary.
+          pwsh -File ci-build.ps1 -BuildRecordMode -OutputFileName azure-ai-agents-linux-amd64 -Version $(cat version.txt | tr -d '\r\n')
```
danieljurek, line 52: `-Version` is redundant. It's filled in by default https://github.com/Azure/azure-dev/blob/main/cli/azd/extensions/azure.ai.agents/ci-build.ps1#L2

danieljurek (approved):

.github/workflows/test-ext-azure-ai-agents.yml
v1212: The default $PSScriptRoot/../version.txt resolves to cli/azd/extensions/version.txt which does not exist - the extension's version.txt is in the same directory, not one level up. The ADO pipeline (build-azd-extension.yml) also passes -Version explicitly. This was discussed in Jon's earlier comment and the -Version was added per that feedback.

jongio (commented):
Clearing prior changes-requested state (automation bug). Original review findings still apply as comments.

PR #8829, opened 2026-06-26
fix(agents): update container deploy schema to use protocol_versions and container_configuration

### LINKED ISSUE #8828: Update container deploy schema: migrate to protocol_versions and container_configuration.image
## Problem

The container deploy path in the `azure.ai.agents` extension still uses the old agent definition schema:
- `container_protocol_versions` (should be unified to `protocol_versions`)
- Top-level `image` field (should move to `container_configuration.image`)

The code deploy path already uses the new schema (`protocol_versions` + `code_configuration`), so only the container deploy path needs updating.

## Affected files

- `cli/azd/extensions/azure.ai.agents/internal/pkg/agents/agent_api/models.go` - HostedAgentDefinition struct and MarshalJSON/UnmarshalJSON
- `cli/azd/extensions/azure.ai.agents/internal/pkg/agents/agent_yaml/map.go` - container deploy definition builder

## Expected behavior

Container deploy should send:

`json
{
  "kind": "hosted",
  "container_configuration": {
    "image": "xxx.azurecr.io/..."
  },
  "protocol_versions": [{"protocol": "responses", "version": "1.0.0"}],
  "cpu": "0.5",
  "memory": "1Gi"
}
`

Instead of the current:

`json
{
  "kind": "hosted",
  "image": "xxx.azurecr.io/...",
  "container_protocol_versions": [{"protocol": "responses", "version": "1.0.0"}],
  "cpu": "0.5",
  "memory": "1Gi"
}
`

## Notes

- Confirm with service team whether old schema is still supported (backward compat) or if this is a breaking change
- May need feature flag coordination if the service hasn't fully rolled out the new schema yet

v1212:
## Summary

Migrate container deploy API schema to match the latest service-side TypeSpec definition:
- `container_protocol_versions` → `protocol_versions` (unified with code deploy)
- top-level `image` → `container_configuration.image`

## Reference

Schema change is documented in the official REST API reference:
https://learn.microsoft.com/en-us/azure/foundry/agents/how-to/deploy-hosted-agent?pivots=rest#create-an-agent

The documented create-agent payload uses:
`json
{
  "definition": {
    "kind": "hosted",
    "container_configuration": {
      "image": "myacr.azurecr.io/my-agent:v1"
    },
    "protocol_versions": [
      {"protocol": "responses", "version": "1.0.0"}
    ],
    "cpu": "1",
    "memory": "2Gi"
  }
}
`

## Changes

- **models.go**: Added `ContainerConfigurationAPI` struct, updated `HostedAgentDefinition` to use `protocol_versions` (via struct tag) and `container_configuration` for container image. Removed custom `MarshalJSON`. Updated `UnmarshalJSON` for backward compat with legacy fields.
- **map.go**: Container deploy path now populates `ContainerConfiguration` instead of bare `Image`.
- **optimize_deploy.go**: `normalizeProtocolVersions` handles both new and legacy field names. Added `normalizeContainerImage` to migrate legacy top-level `image` on raw maps (service responses bypass UnmarshalJSON).
- **service_target_agent.go**: Display code reads from `ContainerConfiguration.Image` with empty-string guard.
- **Tests**: All updated and passing, including legacy unmarshal backward compat test.

## Backward Compatibility

`UnmarshalJSON` still reads legacy API responses that use `container_protocol_versions` and top-level `image`, migrating them to the new schema in memory. The optimize-deploy path also normalizes raw definition maps via `normalizeContainerImage` and `normalizeProtocolVersions`.

Fixes #8828

### DIFF
--- a/cli/azd/extensions/azure.ai.agents/internal/cmd/optimize_deploy.go
+++ b/cli/azd/extensions/azure.ai.agents/internal/cmd/optimize_deploy.go
@@ -340,13 +340,20 @@ func buildDeployDefinition(currentDef map[string]any, envVars map[string]string)
 	return newDef
 }
 
-// normalizeProtocolVersions ensures container_protocol_versions use the
-// canonical "1.0.0" format instead of the legacy "v1" format that the
+// normalizeProtocolVersions ensures protocol_versions (or legacy container_protocol_versions)
+// use the canonical "1.0.0" format instead of the legacy "v1" format that the
 // platform no longer accepts for new versions.
 func normalizeProtocolVersions(def map[string]any) {
-	raw, ok := def["container_protocol_versions"]
+	// Try new field name first, fall back to legacy
+	raw, ok := def["protocol_versions"]
 	if !ok {
-		return
+		raw, ok = def["container_protocol_versions"]
+		if !ok {
+			return
+		}
+		// Migrate legacy field to new name
+		def["protocol_versions"] = raw
+		delete(def, "container_protocol_versions")
 	}
 	protocols, ok := raw.([]any)
 	if !ok {
--- a/cli/azd/extensions/azure.ai.agents/internal/cmd/optimize_deploy.go
+++ b/cli/azd/extensions/azure.ai.agents/internal/cmd/optimize_deploy.go
@@ -337,6 +337,7 @@ func buildDeployDefinition(currentDef map[string]any, envVars map[string]string)
 	}
 	newDef["environment_variables"] = envVars
 	normalizeProtocolVersions(newDef)
+	normalizeContainerImage(newDef)
 	return newDef
 }
 
@@ -368,6 +369,27 @@ func normalizeProtocolVersions(def map[string]any) {
 			pMap["version"] = "1.0.0"
 		}
 	}
+}
+
+// normalizeContainerImage migrates the legacy top-level "image" field to
+// "container_configuration.image" on raw definition maps. This is needed because
+// service responses stored as map[string]any bypass HostedAgentDefinition.UnmarshalJSON.
+func normalizeContainerImage(def map[string]any) {
+	// Already using new schema
+	if _, ok := def["container_configuration"]; ok {
+		return
+	}
+	// Migrate legacy top-level image
+	image, ok := def["image"]
+	if !ok {
+		return
+	}
+	imageStr, ok := image.(string)
+	if !ok || imageStr == "" {
+		return
+	}
+	def["container_configuration"] = map[string]any{"image": imageStr}
+	delete(def, "image")
 }
 
 // pollVersionActive polls the agent version until its status is "active" or a timeout occurs.
--- a/cli/azd/extensions/azure.ai.agents/internal/cmd/optimize_deploy_test.go
+++ b/cli/azd/extensions/azure.ai.agents/internal/cmd/optimize_deploy_test.go
@@ -161,6 +161,36 @@ func TestNormalizeProtocolVersions_MissingField(t *testing.T) {
 	normalizeProtocolVersions(def) // should not panic
 }
 
+func TestNormalizeContainerImage_MigratesLegacy(t *testing.T) {
+	def := map[string]any{
+		"kind":  "hosted",
+		"image": "myimage:latest",
+	}
+	normalizeContainerImage(def)
+
+	containerConfig, ok := def["container_configuration"].(map[string]any)
+	require.True(t, ok, "container_configuration should be set")
+	assert.Equal(t, "myimage:latest", containerConfig["image"])
+	assert.Nil(t, def["image"], "legacy image field should be removed")
+}
+
+func TestNormalizeContainerImage_NoOpWhenNewSchema(t *testing.T) {
+	def := map[string]any{
+		"kind":                    "hosted",
+		"container_configuration": map[string]any{"image": "existing:v1"},
+	}
+	normalizeContainerImage(def)
+
+	containerConfig := def["container_configuration"].(map[string]any)
+	assert.Equal(t, "existing:v1", containerConfig["image"])
+}
+
+func TestNormalizeContainerImage_NoOpWhenNoImage(t *testing.T) {
+	def := map[string]any{"kind": "hosted"}
+	normalizeContainerImage(def) // should not panic
+	assert.Nil(t, def["container_configuration"])
+}
+
 // ---- upsertAgentYamlEnvVar ----
 
 func TestUpsertAgentYamlEnvVar_InsertsNew(t *testing.T) {
--- a/cli/azd/extensions/azure.ai.agents/internal/cmd/optimize_deploy_test.go
+++ b/cli/azd/extensions/azure.ai.agents/internal/cmd/optimize_deploy_test.go
@@ -121,10 +121,10 @@ func TestBuildDeployDefinition_PreservesFieldsAndOverridesEnvVars(t *testing.T) 
 
 func TestBuildDeployDefinition_NormalizesProtocolVersion(t *testing.T) {
 	currentDef := map[string]any{
-		"kind":   "hosted",
+		"kind":                    "hosted",
 		"container_configuration": map[string]any{"image": "myimage:latest"},
-		"cpu":    "1.0",
-		"memory": "2Gi",
+		"cpu":                     "1.0",
+		"memory":                  "2Gi",
 		"container_protocol_versions": []any{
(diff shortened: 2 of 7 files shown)


cli/azd/extensions/azure.ai.agents/internal/cmd/optimize_deploy_test.go
```diff
@@ -121,7 +122,7 @@ func TestBuildDeployDefinition_PreservesFieldsAndOverridesEnvVars(t *testing.T)
 func TestBuildDeployDefinition_NormalizesProtocolVersion(t *testing.T) {
 	currentDef := map[string]any{
 		"kind":   "hosted",
```
jongio: This map literal has inconsistent alignment: `kind`, `cpu`, `memory` are padded with extra spaces but `container_configuration` and `container_protocol_versions` are not. This causes the `gofmt` CI check to fail. Running `gofmt -s -w .` will normalize it (likely removing the extra padding since the key lengths differ too much for gofmt to align).

cli/azd/extensions/azure.ai.agents/internal/cmd/optimize_deploy.go
```diff
@@ -340,13 +340,20 @@ func buildDeployDefinition(currentDef map[string]any, envVars map[string]string)
 	return newDef
 }
 
-// normalizeProtocolVersions ensures container_protocol_versions use the
-// canonical "1.0.0" format instead of the legacy "v1" format that the
+// normalizeProtocolVersions ensures protocol_versions (or legacy container_protocol_versions)
+// use the canonical "1.0.0" format instead of the legacy "v1" format that the
 // platform no longer accepts for new versions.
 func normalizeProtocolVersions(def map[string]any) {
```
jongio, line 347: `normalizeProtocolVersions` migrates `container_protocol_versions` to `protocol_versions` on raw maps, but the parallel case for image is missing. `extractLatestDefinition` round-trips through `json.Marshal`/`Unmarshal` into a `map[string]any`, which bypasses `HostedAgentDefinition.UnmarshalJSON`. If the service still returns a top-level `"image"` key (no `"container_configuration"` wrapper), `buildDeployDefinition` will forward it as-is.

Consider adding an image normalization step (similar to protocol versions) that moves `def["image"]` into `def["container_configuration"] = map[string]any{"image": img}` when the legacy key is present.

jongio (commented):
CI is failing: `golangci-lint` reports a gofmt error in `optimize_deploy_test.go:124`. The map literal in `TestBuildDeployDefinition_NormalizesProtocolVersion` mixes alignment styles (some keys padded, some not). Running `gofmt -s -w .` should fix it.

Beyond the formatting issue, one functional gap: `normalizeProtocolVersions` now migrates the legacy `container_protocol_versions` field to `protocol_versions` on raw maps, but there is no equivalent migration for the legacy top-level `image` field to `container_configuration.image`. Since `AgentVersionObject.Definition` is typed as `any` (line 299 of models.go), the service response is stored as a raw `map[string]any` that never passes through `HostedAgentDefinition.UnmarshalJSON`. If the service still returns the legacy schema, the optimize-deploy path will forward `image` unchanged in new version requests.

The overall approach (unifying to `protocol_versions`, removing custom MarshalJSON, adding backward-compat UnmarshalJSON) is solid and well-tested.

jongio (commented):
My earlier inline concern about the optimize-deploy raw map path still applies: `normalizeProtocolVersions` migrates legacy `container_protocol_versions` to `protocol_versions` in `buildDeployDefinition`, but there is no equivalent for the legacy top-level `image` field to `container_configuration.image`. Since `extractLatestDefinition` produces a `map[string]any` (bypassing `HostedAgentDefinition.UnmarshalJSON`), any legacy `image` from the service response would pass through unchanged in new version requests.

If the service can still return old-schema responses for existing agent versions, consider adding a `normalizeContainerConfiguration` function alongside `normalizeProtocolVersions`. If the service has fully migrated all responses to the new schema, this can be safely ignored.

cli/azd/extensions/azure.ai.agents/internal/cmd/optimize_deploy.go
```diff
@@ -340,13 +340,20 @@ func buildDeployDefinition(currentDef map[string]any, envVars map[string]string)
 	return newDef
```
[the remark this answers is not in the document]
v1212: Fixed: Added `normalizeContainerImage()` that migrates legacy top-level `image` to `container_configuration.image` on raw maps. Called from `buildDeployDefinition`.

cli/azd/extensions/azure.ai.agents/internal/pkg/agents/agent_api/models_test.go
```diff
@@ -169,26 +171,64 @@ func TestHostedAgentDefinition_ContainerImage_RoundTrip(t *testing.T) {
...
+	}
+	if !strings.Contains(s, `"protocol_versions"`) {
+		t.Error("expected JSON to contain \"protocol_versions\"")
 	}
-	if !strings.Contains(s, `"container_protocol_versions"`) {
-		t.Error("expected JSON to contain \"container_protocol_versions\"")
+	// Should NOT contain legacy top-level "image" or "container_protocol_versions"
+	if strings.Contains(s, `"container_protocol_versions"`) {
+		t.Error("unexpected legacy \"container_protocol_versions\" in JSON")
 	}
```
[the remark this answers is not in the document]
v1212: Fixed: Replaced string-based checks with JSON map parsing (`map[string]json.RawMessage`) to verify top-level keys directly.

cli/azd/extensions/azure.ai.agents/internal/cmd/optimize_deploy_test.go
v1212: Fixed: Ran `gofmt -w` to normalize alignment.

cli/azd/extensions/azure.ai.agents/internal/cmd/optimize_deploy.go
v1212: Fixed: Added `normalizeContainerImage()` to handle the parallel migration of legacy `image` field on raw maps, with tests covering migrate, no-op, and missing cases.

cli/azd/extensions/azure.ai.agents/internal/cmd/optimize_deploy.go
```diff
@@ -337,16 +337,24 @@ func buildDeployDefinition(currentDef map[string]any, envVars map[string]string)
 	}
 	newDef["environment_variables"] = envVars
 	normalizeProtocolVersions(newDef)
+	normalizeContainerImage(newDef)
 	return newDef
 }
 
-// normalizeProtocolVersions ensures container_protocol_versions use the
-// canonical "1.0.0" format instead of the legacy "v1" format that the
+// normalizeProtocolVersions ensures protocol_versions (or legacy container_protocol_versions)
```
trangevi, line 344: What makes this "legacy"? Is it that old deployed agents will have this property still, or is it an API route/version that determines it? If it's the former, can we get the service to commit to doing a migration? If it's the latter, is there a way we can tie the normalization to the same, so that when the route/version changes we can drop the back compat support?

trangevi (approved):
Comment isn't blocking, but something to consider in addition to Jon's comments

cli/azd/extensions/azure.ai.agents/internal/cmd/optimize_deploy.go
jongio: It's the former: agents deployed before the TypeSpec update have definitions stored with old field names (`image`, `container_protocol_versions`). The optimize-deploy path reads these as raw `map[string]any` (since `AgentVersionObject.Definition` is typed as `any`), bypassing `HostedAgentDefinition.UnmarshalJSON`. The normalize functions handle migration at that map level.

There's no API route/version split here, just stored data. If the service migrates stored definitions server-side, these normalize functions can be removed. A tracking issue for sunset would help avoid carrying the compat code indefinitely.

v1212: Agreed with Jon. This handles pre-migration stored data. I will file a tracking issue to remove the compat code once the service confirms all stored definitions have been migrated server-side.

v1212: Filed #8848 to track removing the compat code once the service confirms all stored definitions have been migrated.

PR #8843, opened 2026-06-26
Pat to GitHub apps two test 3

chidozieononiwu:
This pull request introduces minor documentation updates to both the `cli/azd/README.md` and `ext/vscode/README.md` files. The only change is the addition of a "Test Changes" note at the end of each file. No code or functional changes are included.


eng/pipelines/templates/stages/publish-extension.yml
```diff
@@ -103,28 +137,22 @@ stages:
...
-                      CreateGitHubRelease: false
+
 
       # Updates the selected extension registry after the GitHub Release exists.
       # Production registry updates also refresh global command snapshots.
       # Mirrors Increment_Version in eng/pipelines/templates/stages/publish.yml.
       - job: Update_Registry
-        dependsOn: Publish_Release
+        dependsOn: 
+          - Publish_Release
```
jongio, line 155: `Update_Registry` depends on `Publish_Release`, but that job was renamed/split in this PR. For non-nightly runs the jobs are now `Create_GitHub_Release` and `Publish_Storage`. For nightly, only `Publish_Storage` exists.

This will break pipeline execution in both paths. Suggested fix:

```suggestion
        dependsOn: 
          - Publish_Storage
```

For non-nightly runs where `Create_GitHub_Release` also exists, you may want to add a conditional dependency on that job too so the registry update waits for the release to be created.

eng/pipelines/templates/steps/publish-extension-github-release.yml
```diff
@@ -0,0 +1,58 @@
...
+        "--repo", "$(Build.Repository.Name)"
+      )
+      if ($version -match "^0\." -or $version -match "-(alpha|beta|preview)") {
+        $createArgs += "--prerelease"
+      }
+      gh release create @createArgs
+      gh release upload $(GH_RELEASE_TAG) release/* --repo $(Build.Repository.Name)
+    displayName: Create GitHub Release and upload artifacts
+    env:
+      GH_TOKEN: $(GH_TOKEN)
```
jongio, line 58: Missing trailing newline at end of file. Same issue in `publish-extension-storage.yml` and `vscode-publish-integration.yml`.

jongio (commented):
This PR migrates CI pipeline auth from a static PAT (`azuresdk-github-pat`) to short-lived GitHub App tokens via the existing `login-to-github` helper. The approach is solid: App tokens are scoped, short-lived, and auditable.

A few issues to address before this is ready:

1. `Update_Registry` still depends on `Publish_Release`, which no longer exists after this refactor. This will fail YAML compilation for both nightly and manual runs.
2. The "Test Changes" markers in both READMEs need to be removed before merge.
3. Several new/modified files are missing a trailing newline.

Issue #8584, opened 2026-06-10
[ext-agents] Interactive agent-service picker hangs silently in non-TTY contexts instead of failing fast


therealjohn:
## Summary

Several `azd ai agent` subcommands render an interactive "Select an agent service" picker when the target agent is ambiguous (a multi-service `azure.yaml` with more than one `azure.ai.agent` block). When stdout/stderr is not a TTY -- CI, an agent harness, or anything piping through `tee`/`tail` -- the picker is invisible and the process blocks forever waiting on stdin that never arrives. There is no error, spinner, or banner; only `Ctrl+C` recovers.

This is the shared root cause behind multiple multi-service hangs (for example `invoke --local` and `files upload`).

## Repro

In any project with >= 2 `azure.ai.agent` services in `azure.yaml`, run a command that must resolve an agent without passing the name, in a non-TTY context:

```bash
azd ai agent invoke --local "hi" | tee out.log               # picker renders into piped stdout -> hangs
azd ai agent files upload my-agent -f ./contract.txt 2>&1 | tail   # same
```

`pstree` / `wchan` shows the extension child blocked in `do_epoll_wait` on the local gRPC pipe back to the parent `azd`; no Azure sockets are opened.

## Expected

1. Detect a non-TTY stdin/stdout and fail fast instead of rendering the picker, for example:
   `cannot prompt for agent selection in a non-interactive context; pass the agent name (or --agent-name) or use --no-prompt`.
   `--no-prompt` already does this for some paths; the picker itself should also be TTY-aware.
2. Even in TTY mode, print a one-line banner to stderr before the picker (for example `azd is waiting for your selection below`). When the picker is visible this is harmless; when stdout is piped the banner at least shows up in `tail`/`tee`/CI logs and explains why `azd` is blocked.

Either mitigation turns a multi-minute silent hang into an immediate, diagnosable signal.

## Environment

- azd 1.25.4; `azure.ai.agents` 0.1.36-preview (also reproduced on 0.1.38-preview)
- Linux (Ubuntu via WSL2)
- Project shape: 1 root `azure.yaml`, 6 `azure.ai.agent` service blocks, shared `.azure/<env>/`

## Related

- #7928 -- pass the agent name with `--local` (one of the commands that falls through to this picker)
- #8411 -- `agent run` / `invoke --local` timing

PR #8851, opened 2026-06-29
fix(ext-agents): accept [agent] [file] positionals for `files upload` to match `invoke`

### LINKED ISSUE #8586: [ext-agents] `azd ai agent files upload` takes the file as the positional arg, hangs on ambiguous agent in multi-service projects
## Summary

`azd ai agent files upload` binds its positional argument to the file path (`[file]`), while the sibling command `azd ai agent invoke` binds its positional to the agent name (`[agent]`). Users -- and LLMs reading the help text -- mirror the `invoke` pattern and run `files upload <agent> -f <file>`, which silently binds `<agent>` to the file path. With multiple `azure.ai.agent` services declared and no `-n/--agent-name`, the agent is ambiguous and the command renders the interactive picker, which hangs silently in non-TTY contexts (see #8584).

## Repro

Multi-service project (>= 2 `azure.ai.agent` services), deployed remote agent:

```bash
azd ai agent files upload my-agent -f ./contract.txt
```

## Observed

Hangs indefinitely (observed ~12 minutes before a manual kill) with no output, no spinner, no CPU activity. `pstree` / `wchan` shows the extension child (`azure-ai-agents-... files upload my-agent -f ./contract.txt`) blocked in `do_epoll_wait` on the local gRPC pipe back to the parent `azd`; no Azure sockets are ever opened. The extension has rendered a service picker into a stdout that is piped, so it is invisible. Sending Enter does nothing; only `Ctrl+C` recovers.

Root cause: `my-agent` is bound to the file positional, so the agent name has to come from `-n/--agent-name`. With multiple services and no `-n`, auto-detect fails and the picker fires.

## Correct invocation (works, ~2s)

```bash
azd ai agent files upload -f ./contract.txt -n my-agent
```

## Expected

- Smallest fix: when the `files upload` positional resolves to a path that does not exist but matches an agent name in `azure.yaml`, fail fast with a hint:
  `did you mean: azd ai agent files upload -n my-agent -f <file>?`
  instead of falling through to "auto-detect agent (failed) -> render picker".
- Better: make the agent the positional for `files upload` too, for consistency with `invoke`.
- See also #8584 -- the non-TTY picker hang that turns this mistake into a silent multi-minute block.

## Impact

Tooling that wraps `azd ai agent files upload` (CI pipelines, agent harnesses, runbooks) hangs forever with no diagnosable signal. The correct flag form works in ~2s; the trap is purely the flipped positional plus the silent picker.

## Environment

- azd 1.25.4; `azure.ai.agents` 0.1.36-preview (also reproduced on 0.1.38-preview)
- Linux (Ubuntu via WSL2)
- Project shape: 1 root `azure.yaml`, 6 `azure.ai.agent` service blocks, shared `.azure/<env>/`

glharper:
## Summary

`azd ai agent files upload` previously bound its single positional argument to the **file path** (`[file]`), while the sibling command `azd ai agent invoke` binds its positional to the **agent name** (`[agent]`). Users -- and LLMs reading the help text -- mirror the `invoke` pattern and run `files upload <agent> -f <file>`, which left the agent name unset. In multi-service projects (>= 2 `azure.ai.agent` services) without `-n/--agent-name`, agent auto-detect fails and the command renders the interactive service picker, which hangs silently in non-TTY contexts (see #8584).

## Fix

`azd ai agent files upload` now accepts **`[agent] [file]`** positional arguments, mirroring `azd ai agent invoke [agent] [message]`:

- **2 args** -- the first is the agent name, the second is the file to upload
- **1 arg** -- the agent name when `--file/-f` already supplies the file, otherwise the file itself
- **0 args** -- the agent is auto-detected from `azure.yaml`; the file is required via `--file/-f`

This is the "better" option called out in the issue ("make the agent the positional for `files upload` too, for consistency with `invoke`"). Passing the agent name as the first positional is now correct, so the command no longer falls through to agent auto-detect and the silent picker.

```bash
# agent auto-detected
azd ai agent files upload ./contract.txt

# explicit agent (matches the invoke shape)
azd ai agent files upload my-agent ./contract.txt

# flags still work
azd ai agent files upload --agent-name my-agent --file ./contract.txt
```

## Changes

- `internal/cmd/files.go` -- change `Use` to `upload [agent] [file]` and `Args` to `cobra.RangeArgs(0, 2)`; add the pure, unit-tested `bindUploadPositionals` helper and wire it into the upload `RunE`; refresh the Long/Example help text. Remove the earlier fail-fast guard now that agent-as-first-positional is the intended shape.
- `internal/cmd/helpers.go` -- remove the now-unused `agentServiceNames` helper.
- `internal/cmd/files_test.go` -- add `TestBindUploadPositionals` (covers all arg/flag combinations) and a command-shape test.
- `CHANGELOG.md` -- Unreleased entry.

## Testing

- `go build ./...`, `go vet`, `golangci-lint run ./internal/cmd/` -- clean
- `go test ./internal/cmd/ -count=1` -- pass
- `gofmt` and `cspell` on changed files -- clean

Fixes #8586


cli/azd/extensions/azure.ai.agents/internal/cmd/files_test.go
```diff
@@ -164,6 +169,74 @@ func TestFilesStatCommand_HasUserIdentityFlag(t *testing.T) {
...
+			positional: "my-agent",
+			agentNames: agentNames,
+			want:       true,
+		},
+		{
+			name:       "non-existent file not matching an agent",
+			positional: "does-not-exist.csv",
+			agentNames: agentNames,
+			want:       false,
+		},
```
jongio, line 194: This case doesn't fully exercise the `file wins over agent-name match` priority. `existingFile` is a full path (e.g. `/tmp/.../my-agent`) that won't appear in the `agentNames` slice (which contains `"my-agent"`), so `slices.Contains` returns false regardless of whether `os.Stat` succeeds. The test passes, but it only validates the early-return from `os.Stat`, not the intended priority logic.

To properly test that a real file beats an agent-name match, use `t.Chdir(tmpDir)` and pass the relative name `"my-agent"` as the positional. That way both `os.Stat` succeeds AND `slices.Contains` would match if the stat guard were removed, proving the guard is what prevents misclassification.

jongio (commented):
The existing bot comment about `os.Stat` is on point and aligns with this codebase's established pattern (`code.go:142`, `helpers.go:136`, `init_from_templates_helpers.go:461` all distinguish `ErrNotExist` from other stat errors). Both `agentNameMisusedAsFilePositional` and `checkUploadPositionalNotAgentName` should gate the agent-name fallback on `errors.Is(err, os.ErrNotExist)` so that permission-denied or broken-symlink cases fall through to the normal "failed to open" error path.

One additional test-coverage observation below.

glharper:
Thanks for the review! Addressed in 1c52d49:

- **`os.Stat` gating (Copilot + @jongio):** Both `agentNameMisusedAsFilePositional` and `checkUploadPositionalNotAgentName` now gate the agent-name fallback on `errors.Is(err, os.ErrNotExist)`. Permission-denied / broken-symlink stat errors fall through to the normal "failed to open local file" path instead of being misclassified as an agent name, matching the established pattern in `code.go`/`helpers.go`/`init_from_templates_helpers.go`.
- **Priority test (@jongio):** Reworked the "existing file wins over agent-name match" case to `t.Chdir` into a temp dir and pass the relative name `"my-agent"`, so `os.Stat` succeeds AND the name is present in `agentNames`. This proves the existence guard (not `slices.Contains`) is what prevents misclassification. The table cases now also run in a clean cwd for determinism.

go build, `golangci-lint`, and the full `internal/cmd` test package all pass.

jongio (commented):
The `1c52d49` changes correctly gate both stat checks on `errors.Is(err, os.ErrNotExist)`. The priority test now uses a relative name (`"my-agent"`) in a temp dir where both `os.Stat` succeeds and `slices.Contains` matches, proving the existence guard is what prevents misclassification. No new issues in the delta.

jongio (commented):
Re-checked after merge commit 1d20688 (main merged, no PR code changes since 1c52d49).

The guard fires before file/flag binding, catching both `files upload my-agent` and `files upload my-agent -f real.csv`. The double `os.Stat` (once in `checkUploadPositionalNotAgentName` to skip the gRPC client creation, once in `agentNameMisusedAsFilePositional` for its standalone contract) is intentional. Tests cover the file-existence priority via a relative name in a temp dir where both `os.Stat` succeeds and `slices.Contains` matches.

No new concerns. Prior os.ErrNotExist gating is correct in both functions.

jongio (commented):
The PR title and body still describe the previous ail fast with hint approach, but commit df76189 pivots to accepting `[agent] [file]` positionals to match `invoke`. Worth updating both before merge so the requested reviewers understand the current design (the CHANGELOG entry is already correct).

jongio (commented):
The PR title and body still describe the previous "fail fast with hint" approach, but commit df76189 pivots to accepting [agent] [file] positionals to match invoke. Worth updating both before merge so the requested reviewers understand the current design (the CHANGELOG entry is already correct).

(Reposted to fix formatting from my previous comment.)

huimiu (approved):

huimiu:

cli/azd/extensions/azure.ai.agents/internal/cmd/files.go
```diff
@@ -132,39 +133,73 @@ type FilesUploadAction struct {
...
 			ctx := azdext.WithAccessToken(cmd.Context())
 
-			if len(args) > 0 && flags.file == "" {
-				flags.file = args[0]
-			}
+			// Mirror `azd ai agent invoke [agent] [message]`: the first
+			// positional is the agent name and the second is the file to
+			// upload. With a single positional, it is the agent name when
+			// --file/-f already supplies the file, otherwise it is the file.
+			flags.agentName, flags.file = bindUploadPositionals(args, flags.agentName, flags.file)
```
huimiu, line 195: This can let positionals override explicit flags. `azd ai agent files upload agent-a ./a.txt --agent-name agent-b --file ./b.txt` would use `agent-a` and `./a.txt`, which could surprise scripts. Would it make sense to reject conflicting values, or make the explicit flags win?

RickWinter (commented):
This aligns `files upload` with `invoke` by accepting `[agent] [file]` positionals, which removes the trap where passing the agent name as the sole positional left the agent unset and stalled on the interactive picker in non-TTY runs. The approach is the right shape: the binding logic is factored into a pure `bindUploadPositionals` helper with table tests covering every arg and flag combination, and the command-shape test pins the `RangeArgs(0, 2)` contract. The single-positional case stays ambiguous by design (`files upload my-agent` with no `--file` is read as the file, not the agent), but that matches `invoke` and is spelled out in the help text, so it is the right call to leave as is. No blockers; this reads as ready to merge.

jongio (approved):
All prior concerns addressed. The title, body, and CHANGELOG now accurately describe the [agent] [file] positional approach.

Regarding @huimiu's question about positionals overriding explicit flags: invoke (invoke.go:149-162) does the same thing, silently assigning lags.name = args[0] regardless of what --agent-name already holds. Consistency with invoke is the stated goal of this PR, and the test case 	wo args override flag values explicitly documents the behavior. If we want to reject conflicts or make flags win, that's a valid UX improvement but should apply to both commands in a follow-up.

No issues with the delta since my last review (merge from main only, no PR code changes).

PR #8852, opened 2026-06-29
feat(agents): provision Foundry memory stores during deploy

### LINKED ISSUE #8742: Add Foundry memory support
## Summary
Add support for Foundry memory in the Azure Developer CLI Foundry extension flow.

## References
- Docs PR showing the expected Foundry memory changes: https://github.com/MicrosoftDocs/azure-ai-docs-pr/pull/12825/changes
- Sample Python provisioning script that demonstrates what needs to be implemented: https://github.com/microsoft-foundry/foundry-samples/blob/main/samples/python/hosted-agents/agent-framework/responses/13-foundry-memory/provision_memory_store.py

## Notes
The sample script should be used as the implementation reference for provisioning/configuring the memory store and wiring the required Foundry memory resources into the hosted-agent experience.

glharper:
## Summary

Adds support for **Foundry memory** in the azure.ai.agents extension by provisioning Foundry **memory stores** as part of the hosted-agent deploy flow, mirroring the reference [`provision_memory_store.py`](https://github.com/microsoft-foundry/foundry-samples/blob/main/samples/python/hosted-agents/agent-framework/responses/13-foundry-memory/provision_memory_store.py) sample.

Users declare one or more memory stores under the agent service config in `azure.yaml`, and azd creates them (idempotent create-if-not-exists) in the Foundry project before deploying the agent. Memory stores back the agent's `memory_search` tool, letting agents retain context across sessions.

Fixes #8742

## What changed

- **`FoundryMemoryStoreClient`** (`internal/pkg/azure/foundry_memory_store_client.go`): new data-plane client with `CreateMemoryStore`, `GetMemoryStore`, and an idempotent `EnsureMemoryStore` (create-if-missing, leave existing stores unchanged). Targets `POST/GET {projectEndpoint}/memory_stores` with API version `2025-11-15-preview`, matching the [official REST contract](https://learn.microsoft.com/en-us/azure/foundry/agents/how-to/memory-usage). Follows the existing `FoundryToolboxClient` pattern.
- **Config** (`internal/project/config.go`): new `memoryStores` section on the agent service-target config (`name`, `description`, `chatModel`, `embeddingModel`, and optional extraction/retention `options`).
- **Deploy wiring** (`internal/project/service_target_agent.go`): `provisionMemoryStores` is invoked from `Deploy` after config/env load and before the agent is created. Validates required fields and reports per-store progress and created/already-exists status.
- **Schema** (`schemas/azure.ai.agent.json`): `memoryStores` array + `MemoryStore` definition.
- **Errors** (`internal/exterrors/codes.go`): `CodeInvalidMemoryStore`, `OpCreateMemoryStore`.
- Tests for the client (request shape, name escaping, ensure get/create/propagate paths), config round-trip, and provisioning validation. CHANGELOG entry.

## Example `azure.yaml` config

```yaml
services:
  my-agent:
    config:
      memoryStores:
        - name: support_memory
          description: Memory for the support agent
          chatModel: gpt-5.2              # model deployment name in the project
          embeddingModel: text-embedding-3-small
          options:
            userProfileEnabled: true
            chatSummaryEnabled: false
            userProfileDetails: "Avoid sensitive data such as age, financials, precise location, and credentials"
```

## Design notes / assumptions

- `chatModel` / `embeddingModel` reference **model deployment names** that must already exist in the Foundry project (consistent with the sample and REST docs).
- Provisioning is **idempotent**: existing stores are left as-is, so deploy is safe to re-run.
- Integration surface is the azure.ai.agent service-target config. The unified Foundry resource shape (PR #8779) currently lives on a feature branch rather than `main`; this PR targets `main` using the existing service-target config, consistent with how `toolboxes`/`connections`/`deployments` are declared.

## Testing

- `go build ./...`
- `go test ./...` (full extension suite passes)
- `gofmt -s`, `golangci-lint run` (0 issues), `cspell` (0 issues) on changed files


cli/azd/extensions/azure.ai.agents/internal/project/service_target_agent.go
```diff
@@ -1058,6 +1066,97 @@ func (p *AgentServiceTargetProvider) Deploy(
...
+	progress azdext.ProgressReporter,
+) error {
+	if config == nil || len(config.MemoryStores) == 0 {
+		return nil
+	}
+
+	if projectEndpoint == "" {
+		return exterrors.Dependency(
+			exterrors.CodeMissingProjectEndpoint,
+			"cannot provision memory stores: the Foundry project endpoint is not set",
```
jongio, line 1128: This uses `CodeMissingProjectEndpoint`, but every other check for `FOUNDRY_PROJECT_ENDPOINT == ""` in this file uses `CodeMissingAiProjectEndpoint`. Should be:

```suggestion
		return exterrors.Dependency(
			exterrors.CodeMissingAiProjectEndpoint,
			"cannot provision memory stores: the Foundry project endpoint is not set",
```

Using a different code here means error-handling callers that match on `CodeMissingAiProjectEndpoint` won't catch this case.

jongio (commented):
One error code inconsistency to fix. The rest of the implementation follows existing patterns cleanly (client structure mirrors the toolsets client, idempotent ensure pattern is solid, tests cover the key branches).

The Copilot bot flagged IncludeBody: true as a concern, but that's the existing convention in oundry_toolsets_client.go, so it's fine here.

cli/azd/extensions/azure.ai.agents/internal/project/service_target_agent.go
glharper: Fixed in 1f1975d50 — now uses `CodeMissingAiProjectEndpoint` to match the other `FOUNDRY_PROJECT_ENDPOINT` checks in this file.

cli/azd/extensions/azure.ai.agents/internal/project/service_target_agent.go
```diff
@@ -1058,6 +1066,97 @@ func (p *AgentServiceTargetProvider) Deploy(
...
+
+	return &azure.MemoryStoreOptions{
+		ChatSummaryEnabled:      options.ChatSummaryEnabled,
+		UserProfileEnabled:      options.UserProfileEnabled,
+		ProceduralMemoryEnabled: options.ProceduralMemoryEnabled,
+		DefaultTTLSeconds:       options.DefaultTtlSeconds,
+		UserProfileDetails:      options.UserProfileDetails,
+	}
+}
+
```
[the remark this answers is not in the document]
glharper: Addressed in 1f1975d50 — `mapMemoryStoreOptions` now returns nil when `options` is nil **or** all fields are unset (new `memoryStoreOptionsEmpty` helper), so an empty `options: {}` no longer sends an empty object. Added a regression test.

cli/azd/extensions/azure.ai.agents/internal/project/service_target_agent.go
```diff
@@ -1058,6 +1066,97 @@ func (p *AgentServiceTargetProvider) Deploy(
...
+				ChatModel:      store.ChatModel,
+				EmbeddingModel: store.EmbeddingModel,
+				Options:        mapMemoryStoreOptions(store.Options),
+			},
+		}
+
+		_, created, err := client.EnsureMemoryStore(ctx, request)
+		if err != nil {
+			return exterrors.ServiceFromAzure(err, exterrors.OpCreateMemoryStore)
+		}
```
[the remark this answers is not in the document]
glharper: Addressed in 1f1975d50 — the operation name is now the neutral `provision_memory_store` (`OpProvisionMemoryStore`) since EnsureMemoryStore covers both the GET and the create, so a failed GET no longer surfaces as a misleading `create_memory_store.*`.

cli/azd/extensions/azure.ai.agents/internal/pkg/azure/foundry_memory_store_client.go
```diff
@@ -0,0 +1,227 @@
...
+	endpoint string,
+	cred azcore.TokenCredential,
+) *FoundryMemoryStoreClient {
+	userAgent := fmt.Sprintf("azd-ext-azure-ai-agents/%s", version.Version)
+
+	clientOptions := &policy.ClientOptions{
+		Logging: policy.LogOptions{
+			AllowedHeaders: []string{azsdk.MsCorrelationIdHeader, "X-Request-Id"},
+			IncludeBody:    true,
+		},
```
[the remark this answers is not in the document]
glharper: Leaving `IncludeBody: true` as-is: it matches the existing convention in `foundry_toolsets_client.go` (and the agent client), and @jongio confirmed this is fine for consistency. Body logging only appears under `--debug`.

jongio (commented):
Incremental review after 1f1975d: the error code is now aligned with the existing CodeMissingAiProjectEndpoint convention, the mapMemoryStoreOptions nil-or-empty guard prevents sending an empty options object to the service, and the op name OpProvisionMemoryStore better reflects the GET-then-POST semantics of EnsureMemoryStore. All fields in memoryStoreOptionsEmpty match the MemoryStoreOptions struct, and the new test case validates the empty-struct path. No remaining concerns.

huimiu:

huimiu (approved):

cli/azd/extensions/azure.ai.agents/internal/project/service_target_agent.go
```diff
@@ -1092,6 +1100,108 @@ func (p *AgentServiceTargetProvider) Deploy(
...
+
+		_, created, err := client.EnsureMemoryStore(ctx, request)
+		if err != nil {
+			return exterrors.ServiceFromAzure(err, exterrors.OpProvisionMemoryStore)
+		}
+
+		if created {
+			fmt.Println(output.WithSuccessFormat("Created memory store %q", store.Name))
+		} else {
+			fmt.Println(output.WithGrayFormat("Memory store %q already exists; leaving as-is", store.Name))
```
RickWinter: Because `EnsureMemoryStore` returns the existing store untouched on a GET hit, a user who edits `chatModel`, `embeddingModel`, or any `options` for an already-created store and redeploys gets no update and no warning, and this line even prints "leaving as-is" in gray right where they would expect their new config to take effect. That is a reasonable default for idempotency, but the silent divergence between azure.yaml and the live store is easy to trip over. Could we at least surface it, for example compare the declared definition against what the GET returned and warn when they differ, so the user knows a change in azure.yaml was ignored? If updating in place is out of scope for this PR, a louder message here would go a long way.

cli/azd/extensions/azure.ai.agents/internal/project/service_target_agent.go
```diff
@@ -1092,6 +1100,108 @@ func (p *AgentServiceTargetProvider) Deploy(
...
+			exterrors.CodeMissingAiProjectEndpoint,
+			"cannot provision memory stores: the Foundry project endpoint is not set",
+			"run 'azd provision' or connect to an existing project via "+
+				"'azd ai agent init --project-id <resource-id>'",
+		)
+	}
+
+	client := azure.NewFoundryMemoryStoreClient(projectEndpoint, p.credential)
+
+	for _, store := range config.MemoryStores {
```
RickWinter, line 1136: nit: validation runs inside the same loop that issues the create calls, so if the second store is missing `chatModel` we have already created the first one before we error out. Re-running is safe thanks to the idempotent ensure, so this is not harmful, just a little surprising on a config typo. Would it be worth validating every store's required fields up front, before the loop that talks to the service, so a bad entry fails fast without half-provisioning?

cli/azd/extensions/azure.ai.agents/internal/project/service_target_agent.go
```diff
@@ -1092,6 +1100,108 @@ func (p *AgentServiceTargetProvider) Deploy(
...
+			},
+		}
+
+		_, created, err := client.EnsureMemoryStore(ctx, request)
+		if err != nil {
+			return exterrors.ServiceFromAzure(err, exterrors.OpProvisionMemoryStore)
+		}
+
+		if created {
+			fmt.Println(output.WithSuccessFormat("Created memory store %q", store.Name))
```
RickWinter: nit: the created/exists results go straight to stdout via `fmt.Println` while the surrounding deploy flow reports through the `progress` reporter. Depending on how progress renders (a spinner), interleaving a raw `Println` mid-loop can look ragged. Routing these through `progress(...)` too would keep the output consistent with the rest of Deploy.

RickWinter (commented):
This adds a data-plane client for Foundry memory stores and wires idempotent create-if-missing provisioning into the hosted-agent deploy, gated on declared `memoryStores` in azure.yaml. The shape is right: it mirrors the existing Foundry client pattern, keys off `FOUNDRY_PROJECT_ENDPOINT` like the rest of the file, only creates on a real 404, and propagates auth/5xx/network errors instead of masking them. Client and provisioning-validation test coverage is solid. The one thing worth resolving before merge is drift: once a store exists, editing its `chatModel`, `embeddingModel`, or `options` in azure.yaml has no effect and no warning, and the gray "leaving as-is" line lands exactly where a user expects their change to apply. Documenting that as intended and, ideally, warning when the declared definition diverges from the live store would prevent a confusing silent no-op. The rest are optional: validating all stores before the first network call avoids half-provisioning on a typo, and routing the created/exists lines through the progress reporter keeps deploy output consistent. Nothing here blocks; the idempotent design is sound and safe to re-run.

jongio (approved):
All prior feedback addressed. The implementation follows existing patterns cleanly, test coverage is solid, and the merge conflicts from main were resolved without issues.

cli/azd/extensions/azure.ai.agents/internal/project/service_target_agent.go
glharper: Addressed — `provisionMemoryStores` now detects drift. `GetMemoryStore` returns the live `definition`, and on a GET hit `memoryStoreDefinitionDrift` compares the declared `chatModel`/`embeddingModel`/`options` against it. When they diverge, deploy emits a warning naming each ignored `azure.yaml` change and pointing the user to delete/rename the store to reapply. Only fields the user explicitly declared are compared, so unset options don't produce false positives. The silent gray `leaving as-is` line now only shows when there is genuinely no drift. Also documented the no-update behavior in the CHANGELOG. Covered by TestMemoryStoreDefinitionDrift.

cli/azd/extensions/azure.ai.agents/internal/project/service_target_agent.go
glharper: Fixed — extracted `validateMemoryStores` which runs over every declared store before the endpoint check and before any network call, so a typo in the second store fails fast without provisioning the first. Covered by TestProvisionMemoryStores_ValidatesAllStoresUpFront.

cli/azd/extensions/azure.ai.agents/internal/project/service_target_agent.go
glharper: Fixed — the created/already-exists status now goes through `progress(...)` instead of `fmt.Println`, keeping deploy output consistent with the rest of the flow. The one thing that still writes directly is the drift *warning* (to stderr via `output.WithWarningFormat`), since that needs to persist rather than be overwritten by the spinner.

jongio (approved):
Incremental review of 119db86: drift detection, upfront validation, and progress routing all look correct.

- validateMemoryStores extraction ensures fail-fast before any network call
- memoryStoreDefinitionDrift only flags fields the user explicitly declared (no false positives from service defaults)
- boolPtrDiffers nil-handling is correct: nil declared = no drift, nil live with set declared = drift
- Test coverage is thorough (drift scenarios, bool pointer edge cases, upfront validation ordering)
- writeMemoryStoreDriftWarning on stderr with actionable remediation text is the right output channel for a warning

No remaining concerns.

cli/azd/extensions/azure.ai.agents/internal/project/service_target_agent.go
```diff
@@ -1092,6 +1100,199 @@ func (p *AgentServiceTargetProvider) Deploy(
...
+	return nil
+}
+
+// memoryStoreDefinitionDrift returns a human-readable list of the fields where the declared
+// definition diverges from the live store. Only fields the user explicitly declared are
+// compared, so unset options (which fall back to service defaults) never report false drift.
+func memoryStoreDefinitionDrift(declared, live azure.MemoryStoreDefinition) []string {
+	var drift []string
+
+	if declared.ChatModel != live.ChatModel {
```
RickWinter, line 1208: This comparison assumes the GET on an existing store returns the full `definition` (chat_model, embedding_model, and any options) that we send on create. If the `2025-11-15-preview` service omits or only partially populates `definition` on GET, then `live.ChatModel` comes back empty and this reports drift on every redeploy of a correctly configured store, so the user sees the "changes were NOT applied" warning even when nothing changed. Have we confirmed against the live service that GET echoes the definition? If it does not, I would gate the warning on the live definition actually being populated rather than comparing declared values against zero values.

RickWinter (commented):
This adds Foundry memory store provisioning to the hosted-agent deploy path: a new data-plane client (`FoundryMemoryStoreClient`), a `memoryStores` config/schema block, and `provisionMemoryStores` wired into `Deploy` before the agent is created. The shape is right. The client faithfully mirrors the existing `FoundryToolboxClient` (same pipeline, same `https://ai.azure.com/.default` scope), validation runs up front before any network call, errors are classified through `exterrors`, and provisioning is idempotent with a clear warning when a declared definition diverges from a store that already exists.

One thing worth resolving before merge: the drift detection in `memoryStoreDefinitionDrift` assumes the GET on an existing store returns the full `definition`. If the preview API does not echo it, the warning fires on every redeploy of a correctly configured store. That is a UX regression rather than a correctness bug, so it does not block, but I would confirm the service contract before this ships. See the inline note.

One non-blocking edge case to keep in mind: because `Deploy` runs per service in parallel, two services declaring the same store name could both GET 404 and then race on create, and the second POST would surface as an error. Fine to leave for now given this is preview and an unusual config, but worth a follow-up if it comes up.

Nothing here blocks. Address the drift-on-GET question (or confirm the contract), and the rest is good to merge.

PR #8866, opened 2026-06-29
Remove Foundry-Features V1Preview headers for GA

### LINKED ISSUE #8186: Remove preview headers sent to Foundry Services for GA
Track when to remove the preview headers sent to Foundry APIs

trangevi:
## Why

Foundry APIs for hosted agents, code agents, and toolboxes are now GA and no longer require the `Foundry-Features: *=V1Preview` opt-in header. These headers were added during preview to gate access to those features. With GA, they are unnecessary and should be cleaned up.

## What changed

Removed all `Foundry-Features` header lines containing V1Preview values across 2 extensions (7 files total):

- **azure.ai.agents** -- `HostedAgents=V1Preview`, `CodeAgents=V1Preview,HostedAgents=V1Preview`, and `Toolboxes=V1Preview` from `operations.go`, `invoke.go`, and `foundry_toolsets_client.go`
- **azure.ai.toolboxes** -- `Toolboxes=V1Preview` from `foundry_toolsets_client.go`

Associated constants, helper functions, comments, and test assertions were also removed or updated.

## Not in scope

- `AgentsOptimization=V2Preview` in `optimize_api/client.go` -- different preview version for a feature that is not yet GA.
- `Skills=V1Preview` in `azure.ai.skills` -- Skills API is still in preview.
- `Routines=V1Preview` in `azure.ai.routines` -- Routines API is still in preview.

Fixes: #8186

### DIFF
--- a/cli/azd/extensions/azure.ai.agents/internal/pkg/agents/agent_api/operations_test.go
+++ b/cli/azd/extensions/azure.ai.agents/internal/pkg/agents/agent_api/operations_test.go
@@ -775,7 +775,6 @@ func TestZipDeployRequest_NoAgentNameHeader_OnUpdate(t *testing.T) {
 	require.Equal(t, "sha", transport.lastReq.Header.Get("x-ms-code-zip-sha256"))
 }
 
-
 // ---------------------------------------------------------------------------
 // DownloadAgentCode tests
 // ---------------------------------------------------------------------------
@@ -842,20 +841,6 @@ func TestDownloadAgentCode_IncludesVersionParam(t *testing.T) {
 	defer result.Body.Close()
 
 	require.Equal(t, "3", transport.lastReq.URL.Query().Get("agent_version"))
-}
-
-func TestDownloadAgentCode_SetsFeatureHeader(t *testing.T) {
-	transport := &downloadTransport{
-		statusCode: http.StatusOK,
-		respBody:   "fake-zip",
-		respHeader: http.Header{},
-	}
-	client := newTestClient("https://test.example.com/api/projects/proj", transport)
-
-	result, err := client.DownloadAgentCode(context.Background(), "my-agent", "v1", "")
-	require.NoError(t, err)
-	defer result.Body.Close()
-
 }
 
 func TestDownloadAgentCode_ReturnsResponseHeaders(t *testing.T) {
--- a/cli/azd/extensions/azure.ai.agents/internal/pkg/agents/agent_api/operations_test.go
+++ b/cli/azd/extensions/azure.ai.agents/internal/pkg/agents/agent_api/operations_test.go
@@ -358,7 +358,6 @@ func TestSessionLifecycleOperations_ApplyUserIdentityHeader(t *testing.T) {
 
 			require.NoError(t, tt.call(client, options))
 			require.Len(t, transport.requests, 1)
-			require.Equal(t, "HostedAgents=V1Preview", transport.requests[0].Header.Get("Foundry-Features"))
 			requireUserIdentityHeader(t, transport.requests[0], "user-1")
 		})
 	}
@@ -521,7 +520,6 @@ func TestGetAgentSessionLogStream_ApplyUserIdentityHeader(t *testing.T) {
 	}
 	require.NotNil(t, request)
 	require.Equal(t, "Bearer test-token", request.Header.Get("Authorization"))
-	require.Equal(t, "HostedAgents=V1Preview", request.Header.Get("Foundry-Features"))
 	requireUserIdentityHeader(t, request, "user-1")
 }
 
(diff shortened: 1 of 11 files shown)


cli/azd/extensions/azure.ai.agents/internal/pkg/agents/agent_api/operations_test.go
```diff
@@ -895,11 +856,6 @@ func TestDownloadAgentCode_SetsFeatureHeader(t *testing.T) {
 	require.NoError(t, err)
 	defer result.Body.Close()
 
-	require.Equal(
-		t,
-		"CodeAgents=V1Preview,HostedAgents=V1Preview",
-		transport.lastReq.Header.Get("Foundry-Features"),
-	)
 }
```
jongio: This test is now effectively empty. It calls DownloadAgentCode, asserts no error, and closes the body, but doesn't verify any behavior unique to this test case. The next test (TestDownloadAgentCode_ReturnsResponseHeaders) already exercises the same call path.

Consider deleting this test entirely, or repurposing it to assert something else (e.g., that the request URL is correctly constructed, or that the response body content is readable).

jongio (commented):
Clean removal of the V1Preview feature headers. The scope is well-defined, the AgentsOptimization=V2Preview exclusion is correctly justified, and the test assertions tracking these headers are removed alongside the production code.

One issue worth addressing: TestDownloadAgentCode_SetsFeatureHeader in operations_test.go is now an empty test body (just calls the function + NoError + Close). Its raison d'etre was the assertion you removed. Either delete it or repurpose it to verify something else about DownloadAgentCode that isn't already covered by TestDownloadAgentCode_ReturnsResponseHeaders.

Minor nit: the mechanical removal left a few double blank lines in operations.go (~line 364, ~975), oundry_toolsets_client.go (agents extension, ~lines 114, 160, 199), and client.go (routines, ~line 116). A quick goimports pass or manual cleanup would tidy those up.

jongio (approved):
Prior findings addressed. Looks good.

huimiu:
- `azd ai skill list` and `azd ai routine list` fail with 403 after the header removal — the service still returns: "This operation requires the following opt-in preview feature(s): Skills=V1Preview / Routines=V1Preview."
- agents and toolboxes are fine without the header.

cli/azd/extensions/azure.ai.skills/internal/pkg/skill_api/client.go
```diff
@@ -23,13 +23,9 @@ import (
 )
 
 const (
-	// DataPlaneAPIVersion: skills live under v1; preview opt-in is via the
-	// Foundry-Features header (SkillsPreviewOptIn).
+	// DataPlaneAPIVersion: skills live under v1.
```
jongio: This constant removal and the `addStandardHeaders` change below need to be reverted. The Skills service still returns 403 without `Skills=V1Preview` per @huimiu's testing.

cli/azd/extensions/azure.ai.routines/internal/pkg/routines/client.go
```diff
@@ -22,11 +22,7 @@ import (
 	"github.com/azure/azure-dev/cli/azd/pkg/azsdk"
 )
 
-const (
-	routinesAPIVersion    = "v1"
-	routinesPreviewHeader = "Foundry-Features"
-	routinesPreviewValue  = "Routines=V1Preview"
-)
+const routinesAPIVersion = "v1"
```
jongio: Restore the `routinesPreviewHeader` and `routinesPreviewValue` constants, and the `addPreviewHeader` function + all its call sites. The Routines service still requires `Routines=V1Preview`.

jongio (changes requested):
Based on @huimiu's testing, the Skills and Routines services still require their V1Preview headers. Removing them causes 403 responses:

> "This operation requires the following opt-in preview feature(s): Skills=V1Preview / Routines=V1Preview."

Agents and toolboxes are confirmed working without headers.

**Action needed**: Restore the `Foundry-Features` headers for `azure.ai.skills` and `azure.ai.routines`. Only the agents and toolboxes header removals should ship in this PR.

jongio (approved):
Skills and routines preview headers restored correctly. Tests verify the header presence. Agents and toolboxes remain header-free as intended.

RickWinter (commented):
This removes the Foundry-Features: *=V1Preview opt-in headers (and their constants, helpers, and test assertions) from the agents and toolboxes extensions now that the hosted-agents, code-agents, and toolboxes APIs are GA. The change is a straightforward, mechanical deletion and the shape is correct: it drops the header set calls and the tests that asserted them, and the scope note correctly leaves the still-preview surfaces alone (AgentsOptimization=V2Preview, Skills=V1Preview, Routines=V1Preview).

The only correctness dependency is external: this is safe exactly when every endpoint these clients hit no longer gates on the header. That is an assertion about the deployed Foundry service that the diff itself can't prove, so it rests on the linked GA rollout. Nothing in the code needs to change. No inline comments, and nothing blocks.

huimiu (approved):

PR #8868, opened 2026-06-29
Default azd ai agent init to code deploy

### LINKED ISSUE #8817: Default azd ai agent init to code deploy
Code deploy is ready to be the default option. 

For `init` from samples, don't download the dockerfile unless the --code-deploy container flag is set.

trangevi:
## Why

Code deploy is ready to be the default deployment mode for `azd ai agent init`. Currently the default is "container" (Docker image), which requires Dockerfile management and ACR setup that most users don't need for simple agent projects.

## What changed

Changed `promptDeployMode()` so "code" (ZIP upload) is the default across all resolution paths:

- **Interactive prompt**: "Source Code (ZIP upload)" is now the first choice and pre-selected
- **`--no-prompt` fallback**: returns "code" instead of "container"
- **`userProvidedManifest` auto-selection**: returns "code" instead of "container"
- **Flag help text**: updated to document the new default

When code deploy is selected in the template/sample flow, any `Dockerfile` and `.dockerignore` downloaded from the sample repository are now removed from the target directory since they are not needed for ZIP upload deployment.

## Notes

- `--deploy-mode container` explicit flag continues to work as before
- Non-Python/non-.NET projects still default to container (the `showCodeDeploy=false` path is unchanged)
- CI/CD scripts using `--no-prompt` without explicit `--deploy-mode` will now get "code" -- this is the intended behavioral change

Fixes: #8817


cli/azd/extensions/azure.ai.agents/internal/cmd/init.go
```diff
@@ -2803,6 +2807,18 @@ func (a *InitAction) downloadAgentYaml(
...
 
+// removeContainerFiles removes Dockerfile and .dockerignore from targetDir.
+// Called when code deploy is selected so container-only files from downloaded
+// samples are not left in the project.
+func removeContainerFiles(targetDir string) {
+	for _, name := range []string{"Dockerfile", ".dockerignore"} {
+		p := filepath.Join(targetDir, name)
+		if err := os.Remove(p); err != nil && !os.IsNotExist(err) {
+			log.Printf("warning: failed to remove %s: %v", p, err)
+		}
```
jongio, line 2822: This function isn't covered by a test. Consider adding a small unit test that verifies:
1. Removes Dockerfile and .dockerignore when they exist
2. No error when the files don't exist (the `os.IsNotExist` path)
3. Leaves other files in the directory untouched

Given the function modifies the filesystem during init, a test helps catch regressions if the target file list changes later.

jongio (commented):
Small, well-scoped default change. The intent is clear and the tests cover the deploy-mode resolution logic.

One gap: `removeContainerFiles` has no unit test. It has specific error-handling behavior (silently ignores ENOENT, logs warnings for other errors) that would benefit from test coverage, especially since it runs unconditionally on all code-deploy selections.

cli/azd/extensions/azure.ai.agents/internal/cmd/init.go
jongio, line 2822: This function isn't covered by a test. Consider adding a small unit test that verifies:
1. Removes Dockerfile and .dockerignore when they exist
2. No error when the files don't exist (the `os.IsNotExist` path)
3. Leaves other files in the directory untouched

Given the function modifies the filesystem during init, a test helps catch regressions if the target file list changes later.

jongio (commented):
Small, well-scoped default change. The intent is clear and the tests cover the deploy-mode resolution logic.

One gap: `removeContainerFiles` has no unit test. It has specific error-handling behavior (silently ignores ENOENT, logs warnings for other errors) that would benefit from test coverage, especially since it runs unconditionally on all code-deploy selections.

cli/azd/extensions/azure.ai.agents/internal/cmd/init.go
```diff
@@ -1594,6 +1594,13 @@ func (a *InitAction) Run(ctx context.Context) error {
 			a.isCodeDeploy = (deployMode == "code")
 
 			if a.isCodeDeploy {
+				// For GitHub-downloaded templates, remove container-only files
+				// that are not needed for code deploy. Skip for local manifests
+				// to avoid deleting user-owned files during re-init.
+				if a.isGitHubUrl(a.flags.manifestPointer) {
+					removeContainerFiles(targetDir)
```
huimiu: [suggestion] `removeContainerFiles` runs before `promptCodeConfig` completes. If `promptCodeConfig` returns an error (network failure, missing runtime list, or user cancels the interactive prompt), Dockerfile and .dockerignore are already gone with no indication to the user and no recovery path within the same run.

Moving the deletion to *after* all configuration succeeds would avoid the inconsistent state. Separately, since `log.Printf` is not surfaced in the azd output stream, the user gets no visible signal that files were removed -- a message through the azd client writer would make this side effect visible.

azd-code-reviewer

huimiu (commented):
This looks good overall

cli/azd/extensions/azure.ai.agents/internal/cmd/init_from_code.go
```diff
@@ -1038,23 +1038,23 @@ func promptDeployMode(ctx context.Context, azdClient *azdext.AzdClient, noPrompt
...
 
 	// When the user provided a manifest explicitly (-m), auto-select the
-	// opinionated default (container) without prompting. Users who want
-	// code deploy with -m can pass --deploy-mode code explicitly.
+	// opinionated default (code) without prompting. Users who want
+	// container deploy with -m can pass --deploy-mode container explicitly.
 	if userProvidedManifest {
-		log.Printf("Auto-selected deploy mode: container (use --deploy-mode code for code deploy)")
-		return "container", nil
+		log.Printf("Auto-selected deploy mode: code (use --deploy-mode container for container deploy)")
```
huimiu, line 1044: [suggestion] This flips the `--no-prompt` fallback (and the `userProvidedManifest` auto-select) from "container" to "code" for Python/.NET projects -- a behavioral breaking change for any CI/CD pipeline calling `azd ai agent init --no-prompt` without an explicit `--deploy-mode`. The PR description acknowledges this, but there is no CHANGELOG entry in the diff. Users upgrading will silently get a different deploy mode; a changelog note would surface the change before it surprises them.

azd-code-reviewer

huimiu (approved):

therealjohn (approved):

jongio (commented):
Incremental check (3 commits since my earlier review). My prior feedback about `removeContainerFiles` test coverage is addressed. The reordering (removal after config success) and `isGitHubUrl` guard that huimiu flagged are also resolved in the latest commits.

One open point from huimiu: surfacing the file removal via user-visible output instead of `log.Printf`. I think the current approach is fine. The guard limits removal to downloaded template files and the user explicitly chose code deploy, so the removal is expected behavior. The changelog entry documents it for anyone upgrading.

PR #8869, opened 2026-06-29
fix: add option to select existing deployment when choosing a different model

### LINKED ISSUE #8854: There's no option to select an existing model deployment
_Migrated from [huimiu/azd-foundry-extension-bugbash#4](https://github.com/huimiu/azd-foundry-extension-bugbash/issues/4), originally reported by @therealjohn._

---

```
? Select a Foundry project to host your agent and any models or tools it uses.: Use an existing Foundry project
Select an Azure subscription to find existing Foundry projects.
? Select a tenant:  2. John Miller (Development) (2 subscriptions)
? Select subscription: johmil - development (e461b37e-bdaa-4b9b-b0bd-eb81eb7ebc56)
? Select a Foundry project: azd-ai-testing-resource / azd-ai-testing (northcentralus)
No existing deployment for model 'gpt-4.1' specified in the selected agent manifest was found in your Foundry project.
? Model 'gpt-4.1' is specified in the agent manifest.: Choose a different model
? Select a model: gpt-5.4-mini
? Select a version for gpt-5.4-mini: 2026-03-17 [up to 50000 quota available]
? Select a SKU for gpt-5.4-mini v2026-03-17: GlobalStandard [2500 quota available]
? Enter deployment capacity for gpt-5.4-mini (GlobalStandard): 10
? Enter model deployment name for model 'gpt-5.4-mini' (defaults to model name): gpt-5.4-mini
```

`? Model 'gpt-4.1' is specified in the agent manifest.: Choose a different model` immediately goes to the list of models to deploy a new one - there is no option to select from existing deployments instead.

trangevi:
## Problem

When `azd ai agent init` finds no existing deployment for the manifest model (e.g. `gpt-4.1`) and the user picks "Choose a different model", they are only shown the model catalog for deploying a **new** model. There is no way to select from models already deployed in the Foundry project.

## Approach

The `getModelDetails` prompt now includes a "Use an existing deployment from this project" option (shown only when the project has deployments). Selecting it shows a sorted picker of all existing deployments and returns the chosen one with `isNew=false`, skipping the new-deployment configuration flow entirely.

Key design decisions:

- **`allDeployments` field on `modelSelector`** -- populated by `getModelDeploymentDetails` before calling `getModelDetails`, so the model selector has visibility into what is already deployed without requiring a second API call.
- **`existingDeploymentError` sentinel** -- carries the selected `*project.Deployment` back to the caller via the error return. This follows the existing `errModelSkipped` pattern and avoids changing the `getModelDetails` return signature.
- **No impact on `--no-prompt` mode** -- the new option is interactive-only; headless/CI flows are unchanged.

## Changes

- `init.go`: Added `allDeployments` field to `modelSelector` struct
- `init_models.go`: Added sentinel type, deployment picker method, new prompt choice, and caller handling
- `init_models_test.go`: Added unit tests for the sentinel error type (unwrapping, message, non-interference with `errModelSkipped`)

Fixes: #8854

image

### DIFF
--- a/cli/azd/extensions/azure.ai.agents/internal/cmd/init_models.go
+++ b/cli/azd/extensions/azure.ai.agents/internal/cmd/init_models.go
@@ -519,6 +519,9 @@ func (a *modelSelector) getModelDetails(
 		case "use_existing":
 			deployment, err := a.promptExistingDeployment(ctx)
 			if err != nil {
+				if exterrors.IsCancellation(err) {
+					return nil, err
+				}
 				return nil, fmt.Errorf("failed to select existing deployment: %w", err)
 			}
 			return nil, &existingDeploymentError{Deployment: deployment}
@@ -779,6 +782,10 @@ func (a *modelSelector) promptModelFromCatalog(ctx context.Context) (*azdext.AiM
 // promptExistingDeployment shows the list of existing deployments in the
 // Foundry project and lets the user pick one.
 func (a *modelSelector) promptExistingDeployment(ctx context.Context) (*project.Deployment, error) {
+	if len(a.allDeployments) == 0 {
+		return nil, fmt.Errorf("no existing deployments available")
+	}
+
 	type labeledDeployment struct {
 		label string
 		info  *FoundryDeploymentInfo
--- a/cli/azd/extensions/azure.ai.agents/internal/cmd/init_models.go
+++ b/cli/azd/extensions/azure.ai.agents/internal/cmd/init_models.go
@@ -34,6 +34,18 @@ var defaultSkuPriority = []string{"GlobalStandard", "DataZoneStandard", "Standar
 // from manifest.Resources in ProcessModels so no deployment is provisioned.
 var errModelSkipped = errors.New("user skipped model")
 
+// existingDeploymentError wraps a project.Deployment selected by the user
+// from the project's existing deployments inside getModelDetails. The caller
+// (getModelDeploymentDetails) detects this via errors.As and returns the
+// deployment with isNew=false.
+type existingDeploymentError struct {
+	Deployment *project.Deployment
+}
+
+func (e *existingDeploymentError) Error() string {
+	return "user selected existing deployment"
+}
+
 func (a *modelSelector) loadAiCatalog(ctx context.Context) error {
 	if a.modelCatalog != nil {
 		return nil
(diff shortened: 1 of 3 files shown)


cli/azd/extensions/azure.ai.agents/internal/cmd/init_models.go
```diff
@@ -482,6 +516,12 @@ func (a *modelSelector) getModelDetails(
 				return nil, fmt.Errorf("no model selected, exiting")
 			}
 			model = selectedModel
+		case "use_existing":
+			deployment, err := a.promptExistingDeployment(ctx)
+			if err != nil {
+				return nil, fmt.Errorf("failed to select existing deployment: %w", err)
+			}
+			return nil, &existingDeploymentError{Deployment: deployment}
```
jongio, line 527: When `promptExistingDeployment` returns `exterrors.Cancelled(...)`, wrapping it with `fmt.Errorf` loses the structured metadata during gRPC serialization (per extension AGENTS.md guidance on not wrapping structured errors). Consider returning the cancellation error directly:

```go
case "use_existing":
    deployment, err := a.promptExistingDeployment(ctx)
    if err != nil {
        if exterrors.IsCancellation(err) {
            return nil, err
        }
        return nil, fmt.Errorf("failed to select existing deployment: %w", err)
    }
    return nil, &existingDeploymentError{Deployment: deployment}
```

cli/azd/extensions/azure.ai.agents/internal/cmd/init_models.go
```diff
@@ -736,6 +776,67 @@ func (a *modelSelector) promptModelFromCatalog(ctx context.Context) (*azdext.AiM
...
+		Sku: project.DeploymentSku{
+			Name:     d.SkuName,
+			Capacity: d.SkuCapacity,
+		},
+	}, nil
+}
+
 func (a *modelSelector) promptForModelLocationMismatch(
 	ctx context.Context,
 	model *azdext.AiModel,
```
jongio, line 849: `promptExistingDeployment` does not guard against an empty `a.allDeployments` slice. The caller currently guards with `len(a.allDeployments) > 0` before offering the option, but if this function is ever called independently or the guard is refactored away, the empty choices array would produce a confusing empty prompt. A quick early-return at the top (e.g. returning an error if len is 0) would make the function self-contained.

jongio (commented):
The sentinel error pattern for carrying the selected deployment back through the error return is consistent with the existing errModelSkipped approach and avoids changing the getModelDetails signature. The scoping change to allDeployments and the listErr rename are correct.

A few observations:

- The use_existing option only appears in the else-if !a.flags.noPrompt branch (model found in catalog). If the model is not in the catalog, users go through promptForAlternativeModel and cannot pick an existing deployment. This seems intentional per the issue scope, but worth confirming.
- The label change from 'Choose a different model' to 'Choose a different model to deploy' is a nice clarity improvement given the new sibling option.

huimiu (approved):

PR #8873, opened 2026-06-30
fix: add omitempty to yaml tags to prevent empty strings in azure.yaml

### LINKED ISSUE #8862: Creates empty strings in azure.yaml for some properties
_Migrated from [huimiu/azd-foundry-extension-bugbash#12](https://github.com/huimiu/azd-foundry-extension-bugbash/issues/12), originally reported by @therealjohn._

---

Image

trangevi:
When `azd` serializes project configuration back to `azure.yaml`, fields with zero values (empty strings, nil slices) were being written as empty strings or empty arrays. This produces noisy, confusing output -- particularly for fields like `host`, `language`, and `project` on services, and `provider`/`variables`/`secrets` on pipeline options.

## Approach

Added `omitempty` to the yaml struct tags on the affected fields so they are omitted from the serialized output when empty:

- **`ServiceConfig`** (`service_config.go`): `project`, `host`, `language`
- **`PipelineOptions`** (`project_config.go`): `provider`, `variables`, `secrets`

This is consistent with how all other optional fields in these structs are already tagged.

Fixes: #8862

### DIFF
--- a/cli/azd/pkg/project/service_config.go
+++ b/cli/azd/pkg/project/service_config.go
@@ -35,11 +35,11 @@ type ServiceConfig struct {
 	// The ARM api version to use for the service. If not specified, the latest version is used.
 	ApiVersion string `yaml:"apiVersion,omitempty"`
 	// The relative path to the project folder from the project root
-	RelativePath string `yaml:"project"`
+	RelativePath string `yaml:"project,omitempty"`
 	// The azure hosting model to use, ex) appservice, function, containerapp
-	Host ServiceTargetKind `yaml:"host"`
+	Host ServiceTargetKind `yaml:"host,omitempty"`
 	// The programming language of the project
-	Language ServiceLanguageKind `yaml:"language"`
+	Language ServiceLanguageKind `yaml:"language,omitempty"`
 	// The output path for build artifacts
 	OutputPath string `yaml:"dist,omitempty"`
 	// The source image to use for container based applications
--- a/cli/azd/pkg/project/service_config.go
+++ b/cli/azd/pkg/project/service_config.go
@@ -37,7 +37,7 @@ type ServiceConfig struct {
 	// The relative path to the project folder from the project root
 	RelativePath string `yaml:"project,omitempty"`
 	// The azure hosting model to use, ex) appservice, function, containerapp
-	Host ServiceTargetKind `yaml:"host,omitempty"`
+	Host ServiceTargetKind `yaml:"host"`
 	// The programming language of the project
 	Language ServiceLanguageKind `yaml:"language,omitempty"`
 	// The output path for build artifacts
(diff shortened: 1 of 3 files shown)


cli/azd/pkg/project/service_config.go
```diff
@@ -35,11 +35,11 @@ type ServiceConfig struct {
 	// The ARM api version to use for the service. If not specified, the latest version is used.
 	ApiVersion string `yaml:"apiVersion,omitempty"`
 	// The relative path to the project folder from the project root
-	RelativePath string `yaml:"project"`
+	RelativePath string `yaml:"project,omitempty"`
 	// The azure hosting model to use, ex) appservice, function, containerapp
-	Host ServiceTargetKind `yaml:"host"`
+	Host ServiceTargetKind `yaml:"host,omitempty"`
```
jongio: Nit: the project's own JSON schema (`schemas/v1.0/azure.yaml.json`) lists `host` in the `required` array for services. Omitting it from serialized yaml will trigger a "missing required property" diagnostic in editors with yaml-language-server enabled. This is probably the desired behavior (clearer signal than `host: ""`), but you may want to either (a) confirm this is fine, or (b) update the schema to make `host` optional for services that use `image` without an explicit host target.

jongio (commented):
The JSON schema at `schemas/v1.0/azure.yaml.json` declares `host` as required for services (`required: ["host"]`). With `omitempty`, services serialized with an empty host will now omit the field instead of writing `host: ""`. This changes yaml-language-server validation behavior in editors: previously a silent empty string, now a "missing required field" warning. Both cases indicate an incomplete service config, so this is likely fine (and arguably clearer). Worth confirming the schema interaction is intentional, or if the schema should also be updated to remove `host` from `required`.

cli/azd/pkg/project/service_config.go
vhvb1989: why host?  This should never be empty. That's either a core supported host or a custom host defined in one extension. SHould never be optional

jongio: Agree. `parseServiceHost` in `service_target.go:77` explicitly errors with `host cannot be empty`, and both JSON schemas (`schemas/v1.0/azure.yaml.json:152`, `schemas/alpha/azure.yaml.json:191`) declare it required. If `host` gets omitted during serialization, the resulting yaml will fail both runtime validation and schema validation. The `project` and `language` fields have different semantics (they can genuinely be empty/inferred), but `host` is always mandatory.

jongio (commented):
The `host` revert aligns the struct tag with both the JSON schema constraint and `parseServiceHost` validation. The regression tests now make this contract explicit, which will guard against accidental re-introduction of omitempty on required fields.

jongio (approved):

trangevi:
closing as duplicate of #8937

PR #8874, opened 2026-06-30
Increase default deployment capacity from 10 to 50 for agents

### LINKED ISSUE #8858: Increase the default deployment capacity from 10 to 50
_Migrated from [huimiu/azd-foundry-extension-bugbash#8](https://github.com/huimiu/azd-foundry-extension-bugbash/issues/8), originally reported by @therealjohn._

---

This default is low for agents: 

Image

Recommend 30-50.

trangevi:
## Motivation

The Azure SKU default capacity for model deployments is typically 10, which is too low for AI agent workloads. Users see "10" pre-filled in the capacity prompt during `azd ai agent init`, requiring manual adjustment every time. The recommendation from bug bash feedback is 30-50.

## Approach

Introduces a `defaultDeploymentCapacity` constant (50) in the agents extension and passes it as the preferred capacity through all deployment resolution paths:

- **Interactive path** (`PromptAiDeployment`): Sets `Options.Capacity` so the capacity prompt defaults to 50 instead of the SKU default.
- **Non-interactive path** (`ResolveModelDeployments`): Sets `Options.Capacity` so resolved deployments use 50.
- **No-prompt fallback** (`resolveNoPromptCapacity`): Uses 50 when the candidate has no capacity set, instead of `max(minCapacity, 1)`.

The preferred capacity still respects SKU constraints -- if the SKU's max capacity is below 50, or remaining quota is insufficient, the system will clamp or reject as before.

## Notes

- User-provided capacity values in `azure.yaml` are unaffected (they take precedence).
- Existing deployments with capacity already set are also unaffected.

Fixes: #8858

### DIFF
--- a/cli/azd/extensions/azure.ai.agents/internal/cmd/init_models_test.go
+++ b/cli/azd/extensions/azure.ai.agents/internal/cmd/init_models_test.go
@@ -33,30 +33,30 @@ func TestResolveNoPromptCapacity(t *testing.T) {
 			wantOk:       true,
 		},
 		{
-			name: "zero capacity defaults to max(minCapacity, 1)",
+			name: "zero capacity defaults to defaultDeploymentCapacity",
 			candidate: &azdext.AiModelDeployment{
 				Capacity: 0,
 				Sku:      &azdext.AiModelSku{MinCapacity: 5},
 			},
-			wantCapacity: 5,
+			wantCapacity: defaultDeploymentCapacity,
 			wantOk:       true,
 		},
 		{
-			name: "zero capacity with zero minCapacity defaults to 1",
+			name: "zero capacity with zero minCapacity defaults to defaultDeploymentCapacity",
 			candidate: &azdext.AiModelDeployment{
 				Capacity: 0,
 				Sku:      &azdext.AiModelSku{MinCapacity: 0},
 			},
-			wantCapacity: 1,
+			wantCapacity: defaultDeploymentCapacity,
 			wantOk:       true,
 		},
 		{
-			name: "negative capacity defaults to max(minCapacity, 1)",
+			name: "negative capacity defaults to defaultDeploymentCapacity",
 			candidate: &azdext.AiModelDeployment{
 				Capacity: -3,
 				Sku:      &azdext.AiModelSku{MinCapacity: 2},
 			},
-			wantCapacity: 2,
+			wantCapacity: defaultDeploymentCapacity,
 			wantOk:       true,
 		},
 		{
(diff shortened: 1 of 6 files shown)


cli/azd/extensions/azure.ai.agents/internal/cmd/init_models_test.go
```diff
@@ -95,6 +95,15 @@ func TestResolveNoPromptCapacity(t *testing.T) {
...
 		},
+		{
+			name: "defaultDeploymentCapacity exceeds maxCapacity returns false",
+			candidate: &azdext.AiModelDeployment{
+				Capacity: 0,
+				Sku:      &azdext.AiModelSku{MaxCapacity: 30},
+			},
+			wantCapacity: 0,
+			wantOk:       false,
+		},
```
jongio, line 115: The test `enforces step alignment on defaultDeploymentCapacity` proves 50 rounds to 51 with step=3. The test `defaultDeploymentCapacity exceeds maxCapacity returns false` proves max=30 rejects. But there's no test for the compound case: what happens when max=50, step=3? The default (50) is within max, but rounds up to 51, which exceeds max and returns false.

This isn't a bug in this PR (the rounding logic is pre-existing), but with the higher default it's more likely to trigger. Consider adding a test documenting this behavior:

```go
{
    name: "step-aligned default exceeds maxCapacity returns false",
    candidate: &azdext.AiModelDeployment{
        Capacity: 0,
        Sku:      &azdext.AiModelSku{MaxCapacity: 50, CapacityStep: 3},
    },
    wantCapacity: 0,
    wantOk:       false,
},
```

Non-blocking. Just ensures the behavior is explicitly documented if someone later wonders why a SKU with max=50 gets skipped.

jongio (commented):
Small, focused change that addresses a real user pain point from bug bash feedback. The constant is well-named, applied consistently across both interactive and non-interactive paths, and the tests cover the important edge cases.

One observation: there's no test for the interaction between CapacityStep rounding and MaxCapacity when both are in play. The existing tests cover step-alignment (no max) and max-exceeded (no step) independently, but not the compound case where defaultDeploymentCapacity is within max yet the step-rounded value exceeds it. See inline comment.

jongio (commented):
Incremental review of 224c42c (1 new commit since my last review).

The clamping approach in 
esolveNoPromptCapacity correctly handles the compound scenario I raised: when the defaulted capacity step-aligns UP past max (e.g., default=50, step=3 rounds to 51, exceeding max=50), the function now clamps to max and step-aligns DOWN (producing 48). This is better than rejecting outright since --no-prompt mode should accommodate rather than fail on valid SKUs.

The defaulted flag cleanly separates behavior: user-specified capacity that exceeds max still hard-fails (respecting explicit choices), while system-defaulted capacity gracefully clamps. Tests cover both the simple clamp (max=30, no step) and step-aligned clamp (max=50, step=7 producing 49).

No new issues. CI green.

jongio (commented):
Incremental review of 224c42c (1 new commit since my last review).

The clamping approach in `resolveNoPromptCapacity` correctly handles the compound scenario I raised: when the defaulted capacity step-aligns UP past max (e.g., default=50, step=3 rounds to 51, exceeding max=50), the function now clamps to max and step-aligns DOWN (producing 48). This is better than rejecting outright since `--no-prompt` mode should accommodate rather than fail on valid SKUs.

The `defaulted` flag cleanly separates behavior: user-specified capacity that exceeds max still hard-fails (respecting explicit choices), while system-defaulted capacity gracefully clamps. Tests cover both the simple clamp (max=30, no step) and step-aligned clamp (max=50, step=7 producing 49).

No new issues. CI green.

huimiu (approved):

PR #8875, opened 2026-06-30
Fix `azd tool uninstall` for VS Code extensions and GitHub Copilot CLI

### LINKED ISSUE #8830: `azd tool uninstall vscode-azure-tools` fails on Windows and MacOS
**Description:**
`azd tool uninstall vscode-azure-tools` fails on Windows and macOS. The specific error is as follows:
Image

**Steps to Reproduce:**
1. Install `vscode-azure-tools`.
2. Run `azd tool uninstall vscode-azure-tools`.

**Environment:**
- Azd version: azd version 1.27.0-beta.1-pr.6484994 (commit 65f747112b09372040a4b58366943b67b53cb9d9) (stable).
- OS: Windows, MacOS.

**Expected Behavior:**
`azd tool uninstall vscode-azure-tools` works successfully on Windows and Mac.

@hemarina and @RickWinter for notification.

### LINKED ISSUE #8831: `azd tool uninstall github-copilot-cli` fails on macOS and Linux
**Description:**
`azd tool uninstall github-copilot-cli` fails on both macOS and Linux, but with different error messages and behaviors. The specific error is as follows:
- Mac: 
Image

- Linux:
Image

**Steps to Reproduce:**
1. Install `github-copilot-cli`.
2. Run `azd tool uninstall github-copilot-cli`.

**Environment:**
- Azd version: azd version 1.27.0-beta.1-pr.6484994 (commit 65f747112b09372040a4b58366943b67b53cb9d9) (stable).
- OS: Linux, MacOS.

**Expected Behavior:**
`azd tool uninstall github-copilot-cli` works successfully on Linux and Mac.

@hemarina and @RickWinter for notification.

hemarina:
## Summary

Fixes two related failures in `azd tool uninstall` where the command failed (or printed a misleading message) for tools that are not removable through the single package manager azd assumed for each platform.

Fixes #8830
Fixes #8831

## #8830 — `azd tool uninstall vscode-azure-tools` fails on Windows and macOS

**Cause:** VS Code refuses to remove an extension that other installed extensions depend on (e.g. *Azure Resources*), so `code --uninstall-extension ...` exits non-zero with `... depend on this`. `--force` does **not** override this. azd misattributed the failure to the generic "the package no longer has a record / was updated outside the package manager" path.

**Fix:** Detect the VS Code dependency-conflict signature on stderr and surface accurate, actionable guidance (remove the dependent extensions first) instead of the misleading lost-record message.

## #8831 — `azd tool uninstall github-copilot-cli` fails on macOS and Linux

**Cause:** The GitHub Copilot CLI is self-managed and can be installed several ways per platform (npm on all; WinGet on Windows; a Homebrew **cask** on macOS + Linux; an install script on macOS + Linux), but azd's manifest assumed a single package manager per platform. On Linux, npm wasn't on `PATH` → *"package manager npm not available"*; on macOS, `brew uninstall copilot-cli` failed because it is a cask.

**Fix:**
- The manifest now lists the full per-platform install-method matrix in preference order (brew cask → npm → install script on macOS/Linux; WinGet → npm on Windows).
- Uninstall is now **detect-then-remove**: it detects which manager actually owns the install and removes via that one. When a self-contained method (install script / direct download) was used, azd gives accurate binary-removal guidance that names the on-`PATH` binary, instead of a misleading package-manager error.
- Added `brew` to the Linux package-manager probe list — the `copilot-cli` cask publishes `x86_64_linux`/`arm64_linux` binaries, so `brew install --cask` works on Linux too.
- Threaded Homebrew **cask** handling (`--cask`) through install, uninstall, and version detection.

## Testing

- Unit tests across `cli/azd/pkg/tool` updated and passing; `gofmt`/`go vet` clean.
- Validated end-to-end in a Linux devcontainer. For each method (Homebrew cask, npm, install script) the harness runs a real `azd tool install` then `azd tool uninstall`, verifying every stage with the actual `copilot --version` command:
  - **Homebrew cask** → installs via brew, uninstall removes it, `copilot --version` then fails (gone). ✓
  - **npm** → installs via npm, uninstall removes it, `copilot --version` then fails (gone). ✓
  - **install script** → standalone binary; uninstall gives manual-removal guidance and leaves the binary, so `copilot --version` still works. ✓
  - Result: **13 passed, 0 failed.**


cli/azd/pkg/tool/version_provider.go
```diff
@@ -55,17 +55,20 @@ func (p *PackageManagerVersionProvider) GetLatestVersion(
...
-		)
+	// Pick the first strategy on this platform that names a package manager.
+	var strategy InstallStrategy
+	found := false
+	for _, s := range tool.InstallStrategies[runtime.GOOS] {
+		if s.PackageManager != "" && s.PackageId != "" {
+			strategy = s
+			found = true
+			break
+		}
```
jongio, line 70: Picks the first strategy with a PackageManager/PackageId but doesn't verify the manager is on PATH. For multi-strategy tools like copilot-cli, a user with npm but not brew gets a version-check failure even though npm could answer it.

Consider iterating with a try/fallback: attempt the query with the first queryable manager, and if it fails (binary missing or command error), try the next one. This is more robust than a `ToolInPath` pre-check since it also handles cases where the binary exists but the command fails for other reasons (e.g. brew installed but not configured).

Not blocking for this PR since this is pre-existing behavior that just becomes more visible with multi-strategy tools. `SelectVersionProvider` around line 481 has the same pattern (checks `isQueryableManager` but not actual availability).

jongio (commented):
The detect-then-remove approach is the right pattern here. Scoping `packageManagerLostRecord` to winget only is correct since brew/npm manifest tracking isn't affected by self-updates.

One observation inline on the version provider strategy selection.

jongio (approved):
Prior feedback addressed: `GetLatestVersion` now iterates through configured managers in preference order with try/fallback, matching the suggestion from my earlier review. The empty-output fallback for the lost-record case is a good defensive improvement too. Tests cover both behaviors.

LGTM.

vhvb1989 (approved):

PR #8876, opened 2026-06-30
Fix agents init --image azure.yaml output

### LINKED ISSUE #8878: Regression: azd ai agent init --image omits service image in azure.yaml
## Description

`azd ai agent init --image <prebuilt-image>` should generate a unified `azure.yaml` agent service that carries the pre-built container image on the service-level `image:` field.

Instead, the generated project omitted `services.<agent>.image`, so `azd deploy` treated the service as a normal Docker build and attempted to build/publish a container from `src/<agent>/Dockerfile`.

## Repro

```bash
azd ai agent init --no-prompt \
  --agent-name my-agent \
  --image myacr.azurecr.io/agents/my-agent:v1

rg "image:" azure.yaml src/my-agent/agent.yaml || true
azd deploy --no-prompt
```

## Expected

`azure.yaml` contains the pre-built image at the service level:

```yaml
services:
  my-agent:
    host: azure.ai.agent
    image: myacr.azurecr.io/agents/my-agent:v1
```

`azd deploy` should use the pre-built image path and skip Dockerfile build/package/publish.

## Actual

`azure.yaml` has no service-level `image:` field. Deploy falls through to Docker build and fails when no Dockerfile exists, for example:

```text
container publish failed: rpc error: code = Unknown desc = remote build failed: open .../src/<agent>/Dockerfile: no such file or directory
```

## Notes

This is a regression in the not-yet-shipped `--image` flow after the unified `azure.yaml` agent definition work. The extension needs to write the pre-built image to `ServiceConfig.Image`, and core must preserve non-empty `osutil.ExpandableString` values when saving `azure.yaml`.

m5i-work:
Fixes #8878

## Summary

- Keep `azd ai agent init --image` aligned with the unified `azure.yaml` shape by writing the pre-built image to the agent service's top-level `image:` field.
- Stop embedding the `--image` value into the synthesized temporary agent manifest / generated agent definition path.
- Fix YAML `omitempty` handling for `osutil.ExpandableString` so non-empty service fields such as `image:` survive `project.Save` / extension `AddService`.
- Add regression coverage that verifies the service image rides on `ServiceConfig.Image` and is not carried in inline agent properties.
- Update pre-built-image deploy guidance to refer to the `azure.yaml` agent service instead of `agent.yaml`.

### DIFF
--- a/cli/azd/extensions/azure.ai.agents/internal/cmd/init_test.go
+++ b/cli/azd/extensions/azure.ai.agents/internal/cmd/init_test.go
@@ -216,58 +216,46 @@ func TestValidateImageFlag(t *testing.T) {
 	}
 }
 
-func TestSetImageOnTemplate(t *testing.T) {
+func TestPreBuiltImageForInit(t *testing.T) {
 	t.Parallel()
 
 	tests := []struct {
-		name      string
-		manifest  *agent_yaml.AgentManifest
-		image     string
-		wantImage string
-		wantErr   bool
+		name     string
+		manifest *agent_yaml.AgentManifest
+		flag     string
+		want     string
 	}{
+		{name: "empty", manifest: &agent_yaml.AgentManifest{Template: agent_yaml.ContainerAgent{}}, want: ""},
+		{name: "nil manifest", manifest: nil, want: ""},
+		{name: "non-container agent", manifest: &agent_yaml.AgentManifest{Template: agent_yaml.Workflow{}}, want: ""},
 		{
-			name:      "empty image is noop",
-			manifest:  &agent_yaml.AgentManifest{Template: agent_yaml.ContainerAgent{}},
-			image:     "",
-			wantImage: "",
+			name:     "manifest image",
+			manifest: &agent_yaml.AgentManifest{Template: agent_yaml.ContainerAgent{Image: "myacr.azurecr.io/agent:v1"}},
+			want:     "myacr.azurecr.io/agent:v1",
 		},
 		{
-			name:      "sets image on container agent",
-			manifest:  &agent_yaml.AgentManifest{Template: agent_yaml.ContainerAgent{}},
-			image:     "myacr.azurecr.io/agent:v1",
-			wantImage: "myacr.azurecr.io/agent:v1",
+			name:     "flag image wins",
+			manifest: &agent_yaml.AgentManifest{Template: agent_yaml.ContainerAgent{Image: "myacr.azurecr.io/agent:v1"}},
+			flag:     "override.azurecr.io/agent:v2",
+			want:     "override.azurecr.io/agent:v2",
 		},
 		{
-			name:     "nil manifest returns error",
-			manifest: nil,
-			image:    "myacr.azurecr.io/agent:v1",
-			wantErr:  true,
+			name:     "trims flag image",
+			manifest: &agent_yaml.AgentManifest{Template: agent_yaml.ContainerAgent{}},
+			flag:     "  myacr.azurecr.io/agent:v1  ",
+			want:     "myacr.azurecr.io/agent:v1",
 		},
 		{
-			name:     "non-container agent returns error",
-			manifest: &agent_yaml.AgentManifest{Template: agent_yaml.Workflow{}},
-			image:    "myacr.azurecr.io/agent:v1",
-			wantErr:  true,
+			name:     "trims manifest image",
+			manifest: &agent_yaml.AgentManifest{Template: agent_yaml.ContainerAgent{Image: "  myacr.azurecr.io/agent:v1  "}},
+			want:     "myacr.azurecr.io/agent:v1",
 		},
 	}
 
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
 			t.Parallel()
-
-			err := setImageOnTemplate(tt.manifest, tt.image)
-			if tt.wantErr {
-				require.Error(t, err)
-				return
-			}
-			require.NoError(t, err)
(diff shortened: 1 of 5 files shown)


cli/azd/extensions/azure.ai.agents/internal/project/service_target_agent.go
```diff
@@ -1095,11 +1095,11 @@ func (p *AgentServiceTargetProvider) Deploy(
 // shouldUsePreBuiltImage determines whether to use a pre-built image.
 //
 // Behavior:
-//   - If no image is configured in agent.yaml, always build from Dockerfile.
+//   - If no image is configured on the agent service, always build from Dockerfile.
```
jongio: This function reads `agentDef.Image` from agent.yaml (line 1074). The updated comment says "on the agent service" which sounds like the azure.yaml service config. For the `--image` init flow, the core framework handles `ServiceConfig.Image` at the package/publish layer before this function is reached. This path is the fallback for manifests that carry an inline image field. Consider keeping "in agent.yaml" for this comment, or noting this is the agent-manifest fallback path.

jongio (commented):
One comment on a doc-comment/code mismatch in `shouldUsePreBuiltImage`.

cli/azd/extensions/azure.ai.agents/internal/cmd/init_test.go
```diff
@@ -355,43 +343,53 @@ func TestSynthesizeImageManifestFile(t *testing.T) {
...
 		},
 	}
 
-	for _, tt := range tests {
-		t.Run(tt.name, func(t *testing.T) {
-			t.Parallel()
-			require.Equal(t, tt.want, agentUsesPreBuiltImage(tt.manifest))
-		})
+	err := action.addToProject(t.Context(), "src/my-agent", manifest)
+	require.NoError(t, err)
```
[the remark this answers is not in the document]
m5i-work: Addressed. Wrapped the direct addToProject call with captureStdout and removed t.Parallel from this test because stdout capture mutates process-global os.Stdout.

cli/azd/extensions/azure.ai.agents/internal/project/service_target_agent.go
m5i-work: Addressed. Updated the comment/log to say the image comes from the loaded agent definition, usually the azure.yaml service image field but possibly the legacy agent.yaml fallback.

hund030 (approved):

huimiu (approved):

jongio (commented):
Re-reviewed against b78387a (force-push since prior review).

Verified the refactoring from mutation-based image handling (`setImageOnTemplate` + `agentUsesPreBuiltImage`) to the pure extractor `preBuiltImageForInit`. The flag-wins-over-manifest precedence is correct, and returning `""` for nil/non-container manifests is safe since input validation happens earlier in the init flow. Removing the image from the synthesized temporary manifest correctly centralizes the image on the `azure.yaml` service config.

The `ExpandableString.IsZero()` addition enables proper YAML `omitempty` detection for the `image:` field during `project.Save`.

cli/azd/pkg/project/project_test.go
```diff
@@ -484,6 +484,30 @@ postbuild:
...
+	projectFile := filepath.Join(t.TempDir(), "azure.yaml")
+
+	err := Save(t.Context(), projectConfig, projectFile)
+	require.NoError(t, err)
+
+	fileContent, err := os.ReadFile(projectFile)
+	require.NoError(t, err)
+	assert.Contains(t, string(fileContent), "image: myacr.azurecr.io/agent:v1")
+}
+
```
jongio, line 519: Consider adding a round-trip assertion here: load the saved YAML back with `Load(t.Context(), projectFile)` and verify the image field reconstructs correctly, similar to what `TestInfraDefaultsNotSavedToYaml` does below. This would catch any `UnmarshalYAML` gap in addition to the `MarshalYAML`/`IsZero` path already tested.

m5i-work: Addressed. Added a Load round-trip assertion to TestServiceImageSavedToYaml to verify the saved image reconstructs back into ServiceConfig.Image.

jongio (commented):
Re-reviewed against 5a84bbb (force-push since prior review at b78387a).

Verified the refactoring:

- setImageOnTemplate + agentUsesPreBuiltImage (mutation + query) replaced by preBuiltImageForInit (pure extractor). Flag-wins-over-manifest precedence is correct.
- Synthesized manifest no longer embeds the image. Confirmed in addToProject that containerDef (line 2946) is still used for AgentDefinitionToServiceProperties but ServiceConfig.Image now reads from preBuiltImage, preventing the image from leaking into inline agent properties.
- ExpandableString.IsZero() correctly delegates to Empty(), enabling YAML omitempty round-trip for the image field. The value receiver matches the existing method set.
- shouldUsePreBuiltImage doc comment now accurately describes the dual source (azure.yaml service field with legacy agent.yaml fallback).
- Test coverage: TestPreBuiltImageForInit covers flag precedence and trimming, TestAddToProjectPreBuiltImageWritesServiceImage verifies the image rides on ServiceConfig.Image (not additional properties), and TestServiceImageSavedToYaml covers the save/load round-trip.

Prior comments addressed: doc-comment mismatch fix, round-trip assertion added, captureStdout wrapping added.

JeffreyCA (approved):

vhvb1989 (approved):

hemarina (approved):

trangevi (approved):

PR #8877, opened 2026-06-30
fix: enforce known host values in azure.yaml schema

### LINKED ISSUE #8856: No squiggles in YAML validation inside VS Code
_Migrated from [huimiu/azd-foundry-extension-bugbash#6](https://github.com/huimiu/azd-foundry-extension-bugbash/issues/6), originally reported by @therealjohn._

---

Step 2 says to expect squiggles - I do not see them: 

Image

I do see autocomplete:

Image

huimiu:
## What

Add an `enum` of the known `host` values to the per-service schema in both `schemas/v1.0/azure.yaml.json` and `schemas/alpha/azure.yaml.json`, mirroring the existing `examples` list.

## Why

Fixes #8856. In VS Code (redhat YAML extension) the unified Foundry `azure.yaml` gave autocomplete but no validation squiggles. Root cause: `host` was a free-form `type: string`, so a misspelled host like `azure.ai.projct` was a valid string and matched no `if/then` branch — no error. yaml-language-server validates with `ajv`, so this reproduces exactly; with the `enum`, misspelled hosts now squiggle ("must be equal to one of the allowed values").

## Scope

Minimal and low-risk: the `enum` mirrors the curated `examples` the schema already advertises, so every valid host still passes. `containerapp-dotnet` is intentionally excluded (internal-only host, not valid in `azure.yaml`). Unknown-field detection was intentionally left out of scope to avoid surfacing squiggles in existing `azure.yaml` files that rely on the schema's `additionalProperties: true`.

## Verification

- `ajv` (same engine as the YAML extension): misspelled host → invalid; baseline Foundry sample, classic `containerapp` sample, and `simple.azure.yaml` → valid.
- CI checks pass locally: `jsonlint` and `jsonschema2md` on both schema files.

### DIFF
--- a/schemas/alpha/azure.yaml.json
+++ b/schemas/alpha/azure.yaml.json
@@ -222,22 +222,6 @@
                         "type": "string",
                         "title": "Required. The type of Azure resource used for service implementation",
                         "description": "The Azure service that will be used as the target for deployment operations for the service.",
-                        "enum": [
-                            "appservice",
-                            "containerapp",
-                            "function",
-                            "springapp",
-                            "staticwebapp",
-                            "aks",
-                            "ai.endpoint",
-                            "azure.ai.agent",
-                            "microsoft.foundry",
-                            "azure.ai.project",
-                            "azure.ai.connection",
-                            "azure.ai.toolbox",
-                            "azure.ai.skill",
-                            "azure.ai.routine"
-                        ],
                         "examples": [
                             "appservice",
                             "containerapp",
--- a/schemas/alpha/azure.yaml.json
+++ b/schemas/alpha/azure.yaml.json
@@ -221,7 +221,7 @@
                     "host": {
                         "type": "string",
                         "title": "Required. The type of Azure resource used for service implementation",
-                        "description": "The Azure service that will be used as the target for deployment operations for the service.",
+                        "description": "The Azure service that will be used as the target for deployment operations for the service. Limited to the host types azd recognizes; hosts contributed by extensions are resolved by azd at runtime and may not be listed here.",
                         "enum": [
                             "appservice",
                             "containerapp",
--- a/schemas/alpha/azure.yaml.json
+++ b/schemas/alpha/azure.yaml.json
@@ -221,15 +221,17 @@
                     "host": {
                         "type": "string",
                         "title": "Required. The type of Azure resource used for service implementation",
-                        "description": "The Azure service that will be used as the target for deployment operations for the service. Limited to the host types azd recognizes; hosts contributed by extensions are resolved by azd at runtime and may not be listed here.",
-                        "enum": [
+                        "description": "The Azure service that will be used as the target for deployment operations for the service.",
+                        "examples": [
                             "appservice",
                             "containerapp",
                             "function",
+                            "springapp",
                             "staticwebapp",
                             "aks",
                             "ai.endpoint",
                             "azure.ai.agent",
+                            "microsoft.foundry",
                             "azure.ai.project",
                             "azure.ai.connection",
                             "azure.ai.toolbox",
--- a/schemas/alpha/azure.yaml.json
+++ b/schemas/alpha/azure.yaml.json
@@ -222,16 +222,14 @@
                         "type": "string",
                         "title": "Required. The type of Azure resource used for service implementation",
                         "description": "The Azure service that will be used as the target for deployment operations for the service.",
-                        "examples": [
+                        "enum": [
                             "appservice",
                             "containerapp",
                             "function",
-                            "springapp",
                             "staticwebapp",
                             "aks",
                             "ai.endpoint",
                             "azure.ai.agent",
-                            "microsoft.foundry",
                             "azure.ai.project",
                             "azure.ai.connection",
                             "azure.ai.toolbox",
--- a/schemas/alpha/azure.yaml.json
+++ b/schemas/alpha/azure.yaml.json
@@ -222,6 +222,22 @@
                         "type": "string",
                         "title": "Required. The type of Azure resource used for service implementation",
                         "description": "The Azure service that will be used as the target for deployment operations for the service.",
+                        "enum": [
+                            "appservice",
+                            "containerapp",
+                            "function",
+                            "springapp",
+                            "staticwebapp",
+                            "aks",
+                            "ai.endpoint",
+                            "azure.ai.agent",
+                            "microsoft.foundry",
+                            "azure.ai.project",
+                            "azure.ai.connection",
+                            "azure.ai.toolbox",
+                            "azure.ai.skill",
+                            "azure.ai.routine"
+                        ],
                         "examples": [
                             "appservice",
                             "containerapp",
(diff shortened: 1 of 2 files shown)


schemas/v1.0/azure.yaml.json
```diff
@@ -183,6 +183,22 @@
...
+                            "function",
+                            "springapp",
+                            "staticwebapp",
+                            "aks",
+                            "ai.endpoint",
+                            "azure.ai.agent",
+                            "microsoft.foundry",
+                            "azure.ai.project",
+                            "azure.ai.connection",
+                            "azure.ai.toolbox",
```
jongio: Six of the fourteen values here (`microsoft.foundry`, `azure.ai.project`, `azure.ai.connection`, `azure.ai.toolbox`, `azure.ai.skill`, `azure.ai.routine`) don't correspond to any registered service-target provider in the extension registry or Go source. The extensions with matching names (`azure.ai.projects`, `azure.ai.connections`, etc.) only declare `custom-commands` + `metadata` capabilities, not `service-target-provider`. Including them in an enum implies they're usable as `host:` values, but azd can't resolve them today.

Also, `springapp` is included despite the explicit code comment in `service_target.go:33` stating "SpringAppTarget is intentionally omitted because support has been removed."

schemas/alpha/azure.yaml.json
```diff
@@ -222,6 +222,22 @@
...
+                            "function",
+                            "springapp",
+                            "staticwebapp",
+                            "aks",
+                            "ai.endpoint",
+                            "azure.ai.agent",
+                            "microsoft.foundry",
+                            "azure.ai.project",
+                            "azure.ai.connection",
+                            "azure.ai.toolbox",
```
jongio: Same concern as in the v1.0 schema. Additionally, the `examples` array (which already provides autocomplete in the YAML extension) still lists only 8 values, while this enum adds 14. If the intent is to advertise known hosts, updating `examples` alone would give autocomplete without rejecting extension-registered hosts.

jongio (commented):
The goal here (IDE squiggles on misspelled hosts) is good, but a strict `enum` conflicts with azd's extension architecture. `parseServiceHost()` in `service_target.go:68-75` explicitly allows any non-empty host value through because extensions register custom service-target providers at runtime via gRPC. This makes the `host` field intentionally open-ended; locking it down in the schema means every extension-provided host will show a spurious validation error in the user's editor.

Beyond the extensibility concern (already noted by the earlier review), several of the proposed enum values don't actually correspond to registered service-target providers in the codebase or extension registry:

- `microsoft.foundry`
- `azure.ai.project`
- `azure.ai.connection`
- `azure.ai.toolbox`
- `azure.ai.skill`
- `azure.ai.routine`

None of these appear in the registry, extension `providers` declarations, or Go source as valid host identifiers. Only `azure.ai.agent` (from the `azure.ai.agents` extension) and the 6 built-in targets (`appservice`, `containerapp`, `function`, `staticwebapp`, `aks`, `ai.endpoint`) are actually resolvable today. Including unregistered names in the enum would validate misspelled-but-close values while still rejecting valid extension hosts not on the list.

A possible alternative: keep `host` as `type: string` (preserving extensibility) but expand the `examples` array so the YAML extension offers autocomplete for all known hosts without rejecting unknown ones.

jongio (commented):
Thanks for removing `springapp`, `microsoft.foundry`, and the redundant `examples` array. The updated description acknowledging extension hosts is also a good addition.

Two issues remain from my earlier review:

**1. `enum` still breaks extensibility**

The description says "hosts contributed by extensions are resolved by azd at runtime and may not be listed here," which is accurate, but `enum` in JSON Schema rejects any value not in the list. That means any extension-registered host that isn't listed will produce a spurious validation error in the editor. `parseServiceHost()` (service_target.go:68-75) explicitly allows any non-empty host value because extensions register custom service targets at runtime via gRPC. Using `examples` instead of `enum` would give autocomplete without rejecting unknown hosts.

**2. Five enum values don't correspond to registered service-target providers**

I checked every extension under `cli/azd/extensions/`:
- `azure.ai.projects` (capabilities: custom-commands, metadata) - no `service-target-provider`, no `providers:` section
- `azure.ai.connections` (capabilities: custom-commands, metadata) - same
- `azure.ai.toolboxes` (capabilities: custom-commands, metadata) - same
- `azure.ai.skills` (capabilities: custom-commands, metadata) - same
- `azure.ai.routines` (capabilities: custom-commands, metadata) - same

Only `azure.ai.agents` declares `service-target-provider` capability and registers `azure.ai.agent` as a provider. The other five names (`azure.ai.project`, `azure.ai.connection`, `azure.ai.toolbox`, `azure.ai.skill`, `azure.ai.routine`) aren't resolvable as host values today. Including them in an enum validates misspelled-but-close values while failing to catch actual typos in the names azd can resolve.

Suggestion: switch back to `examples` (keeps autocomplete, doesn't reject extension hosts), and trim the list to only the 7 hosts that are actually resolvable: `appservice`, `containerapp`, `function`, `staticwebapp`, `aks`, `ai.endpoint`, `azure.ai.agent`.

schemas/v1.0/azure.yaml.json
```diff
@@ -182,17 +182,15 @@
                     "host": {
                         "type": "string",
                         "title": "Required. The type of Azure resource used for service implementation",
-                        "description": "The Azure service that will be used as the target for deployment operations for the service.",
-                        "examples": [
+                        "description": "The Azure service that will be used as the target for deployment operations for the service. Limited to the host types azd recognizes; hosts contributed by extensions are resolved by azd at runtime and may not be listed here.",
+                        "enum": [
```
jongio: `enum` will reject any host value not in this list. Since extensions register arbitrary service-target providers at runtime (`azure.ai.agents` does this today), users of those extensions will get false validation errors in their editor. Switch back to `examples` so the YAML extension offers autocomplete for known hosts without rejecting unknown ones.

jongio (changes requested):
Two issues from my prior feedback remain unaddressed:

1. **enum rejects valid extension hosts**: `parseServiceHost()` (service_target.go:68-75) allows any non-empty host value because extensions register custom service-target providers at runtime via gRPC. `enum` in JSON Schema rejects any value not in the list, so every extension-provided host will show a spurious validation error in the editor. The description acknowledges this but `enum` still rejects them. Use `examples` instead to give autocomplete without blocking unknown hosts.

2. **5 enum values aren't registered service-target providers**: `azure.ai.project`, `azure.ai.connection`, `azure.ai.toolbox`, `azure.ai.skill`, `azure.ai.routine` don't appear as `service-target-provider` capabilities in any extension manifest. Only `azure.ai.agents` declares that capability. Including unresolvable host names validates typo-adjacent values while rejecting valid extension hosts.

Suggested fix: revert to `examples` (preserves autocomplete, doesn't reject extension hosts) and trim to the 7 actually-resolvable hosts: `appservice`, `containerapp`, `function`, `staticwebapp`, `aks`, `ai.endpoint`, `azure.ai.agent`.

PR #8879, opened 2026-06-30
fix: warn when agent services share a Foundry agent name

### LINKED ISSUE #8872: warn on duplicate agent services referencing the same agent name
Warn during deploy when two agents reference the same `name`, which is the unique identifier in Foundry for an agent. 

Consider an example like:

```
toolbox-agent-2:
        project: src/toolbox-agent-2
        host: azure.ai.agent
        language: docker
        docker:
            remoteBuild: true
        container:
            resources:
                cpu: "0.5"
                memory: 1Gi
        description: |
            An Agent Framework agent with Foundry Toolbox integration.
        environmentVariables:
            - name: AZURE_AI_MODEL_DEPLOYMENT_NAME
              value: gpt-5.4-mini
            - name: TOOLBOX_ENDPOINT
              value: '{{TOOLBOX_ENDPOINT}}'
        kind: hosted
        metadata:
            tags:
                - Agent Framework
                - AI Agent Hosting
                - Azure AI AgentServer
                - Responses Protocol
                - Streaming
        name: toolbox-agent-2
        protocols:
            - protocol: responses
              version: 1.0.0
        startupCommand: python main.py
    toolbox-agent-4:
        project: src/toolbox-agent-2
        host: azure.ai.agent
        language: docker
        docker:
            remoteBuild: true
        container:
            resources:
                cpu: "0.5"
                memory: 1Gi
        description: |
            An Agent Framework agent with Foundry Toolbox integration.
        environmentVariables:
            - name: AZURE_AI_MODEL_DEPLOYMENT_NAME
              value: gpt-5.4-mini
            - name: TOOLBOX_ENDPOINT
              value: '{{TOOLBOX_ENDPOINT}}'
        kind: hosted
        metadata:
            tags:
                - Agent Framework
                - AI Agent Hosting
                - Azure AI AgentServer
                - Responses Protocol
                - Streaming
        name: toolbox-agent-2
        protocols:
            - protocol: responses
              version: 1.0.0
        startupCommand: python main.py
```

huimiu:
Fixes #8872

## Problem

In an `azure.yaml` with multiple `host: azure.ai.agent` services, two or more services can carry the same agent `name`. Foundry uses the agent `name` as an agent's unique identifier, so deploying both services creates/updates the **same** Foundry agent — they silently overwrite each other, and azd gives no signal.

Example from the issue: services `toolbox-agent-2` and `toolbox-agent-4` both set `name: toolbox-agent-2`.

## Change

`azd deploy`/`azd up` now emits a non-blocking warning when two or more `azure.ai.agent` services resolve to the same Foundry agent `name`, naming the colliding services so the user can give each a unique name. Deploy still proceeds.

- `findDuplicateAgentNames` scans the `azure.ai.agent` services, resolves each via `project.LoadAgentDefinition` (covers the unified inline shape, the legacy config shape, and a legacy on-disk `agent.yaml`), groups the azure.yaml service keys by resolved agent name, and returns groups shared by ≥2 services (deterministically sorted). Services that aren't hosted agents, can't be resolved, or have an empty name are skipped.
- `warnDuplicateAgentNames` prints one yellow warning per group to stderr via `output.WithWarningFormat`, matching the existing init / legacy-shape warning style.
- Wired into `predeployHandler` behind a new `duplicateAgentNameWarnOnce sync.Once`, so the project-wide scan warns at most once per process even though the handler fires per agent service — mirroring the existing `developerRBACOnce` pattern.

## Tests

`TestFindDuplicateAgentNames` covers: no services, unique names, the issue's two-service collision, a three-way collision, multiple groups sorted by agent name, and non-agent-host / unresolvable services being skipped, plus a nil-project case.

## Validation

From `cli/azd/extensions/azure.ai.agents`: `gofmt` clean, `go build ./...`, `go vet ./internal/cmd/`, and the full `internal/cmd` test suite all pass.

### DIFF
--- a/cli/azd/extensions/azure.ai.agents/internal/cmd/listen.go
+++ b/cli/azd/extensions/azure.ai.agents/internal/cmd/listen.go
@@ -10,6 +10,7 @@ import (
 	"fmt"
 	"log"
 	"os"
+	"sort"
 	"strings"
 	"sync"
 
@@ -20,6 +21,7 @@ import (
 
 	"github.com/Azure/azure-sdk-for-go/sdk/azidentity"
 	"github.com/azure/azure-dev/cli/azd/pkg/azdext"
+	"github.com/azure/azure-dev/cli/azd/pkg/output"
 )
 
 // configureExtensionHost wires the service target and event handlers on the
@@ -164,8 +166,22 @@ var (
 	developerRBACErr  error
 )
 
+// duplicateAgentNameWarnOnce ensures the duplicate agent-name warning is emitted
+// at most once per extension process lifetime. Service-level predeploy handlers
+// fire per-service, but the check is project-scoped — a single pass over every
+// azure.ai.agent service surfaces all collisions without repeating the warning
+// for each colliding service.
+var duplicateAgentNameWarnOnce sync.Once
+
 func predeployHandler(ctx context.Context, azdClient *azdext.AzdClient, args *azdext.ServiceEventArgs) error {
 	svc := args.Service
+
+	// Warn (once) when multiple agent services resolve to the same Foundry agent
+	// name. Foundry identifies an agent by its name, so such services overwrite
+	// each other on deploy. Advisory only — deploy continues.
+	duplicateAgentNameWarnOnce.Do(func() {
+		warnDuplicateAgentNames(args.Project)
+	})
 
 	deployments, err := collectProjectDeployments(args.Project.Services)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


cli/azd/extensions/azure.ai.agents/internal/cmd/listen.go
```diff
@@ -206,6 +222,71 @@ func isHostedAgentService(svc *azdext.ServiceConfig, proj *azdext.ProjectConfig)
...
+
+	var groups []duplicateAgentNameGroup
+	for agentName, serviceNames := range servicesByAgentName {
+		if len(serviceNames) < 2 {
+			continue
+		}
+		sort.Strings(serviceNames)
+		groups = append(groups, duplicateAgentNameGroup{agentName: agentName, serviceNames: serviceNames})
+	}
+	sort.Slice(groups, func(i, j int) bool { return groups[i].agentName < groups[j].agentName })
```
jongio, line 271: Per the project's Go 1.26+ guidelines, `slices.Sort`/`slices.SortFunc` with `cmp.Compare` are preferred over `sort.*`. You'd also swap the `sort` import for `cmp` and `slices`.

```suggestion
		slices.Sort(serviceNames)
		groups = append(groups, duplicateAgentNameGroup{agentName: agentName, serviceNames: serviceNames})
	}
	slices.SortFunc(groups, func(a, b duplicateAgentNameGroup) int { return cmp.Compare(a.agentName, b.agentName) })
```

jongio (commented):
One convention item: the new `sort.*` calls can be switched to `slices.Sort`/`slices.SortFunc` to match the Go 1.26+ patterns used throughout this extension. See inline suggestion for the specific change.

trangevi (approved):

PR #8880, opened 2026-06-30
Fix: ACR not created/linked for hosted container agents (#8861)

### LINKED ISSUE #8861: ACR was not created / linked  or used.
_Migrated from [huimiu/azd-foundry-extension-bugbash#11](https://github.com/huimiu/azd-foundry-extension-bugbash/issues/11), originally reported by @therealjohn._

---

Image

No ACR env var was set.

hund030:
Fixes #8861

## Issue

A hosted container agent (built from source) could deploy without an Azure Container Registry being created/wired, leaving `AZURE_CONTAINER_REGISTRY_ENDPOINT` unset. `azd up` then failed at container publish:

```
container publish failed: could not determine container registry endpoint ...
'AZURE_CONTAINER_REGISTRY_ENDPOINT' environment variable has been set
```

## Root cause

An ACR is provisioned/wired in two different paths, and both had holes:

- **Brownfield** (reusing an existing project, `endpoint:` set): the bicep-less init flow defers ACR to the provider, but the provider's brownfield path only reconciled model deployments and never created an ACR — so the env var was never produced and the ACR question was never asked.
- **Greenfield** (new project): `deriveIncludeAcr` gated `includeAcr` on a `docker:` block, but `docker:` isn't part of the `azure.ai.agent` schema. A schema-conformant hand-authored agent (`kind: hosted`, no `docker:`/`image:`/`codeConfiguration:`, like `schemas/examples/simple.azure.yaml`) got `includeAcr=false`.

Two related brownfield UX gaps surfaced once provisioning was wired: `provision --preview` reported "nothing to provision", and `init --infra` failed with a cryptic `service has endpoint: (brownfield)` error.

## What changed

- **init**: for a container agent on an existing project, configure ACR even in the bicep-less flow (discover the project's registry connection, or prompt / "create a new one").
- **brownfield.bicep + provider**: create a Premium ACR, grant `AcrPull` to the project identity, add a `ContainerRegistry` connection, and surface the registry outputs when the `acr` pending-provision reason is set.
- **preview**: `provision --preview` runs a what-if so the ACR/deployments show as `Create`s; also fixed a latent 64-char ARM deployment-name overflow.
- **init --infra**: clear "unsupported for existing-project configs" error instead of the raw synthesizer error.
- **deriveIncludeAcr**: key on the schema fields (`codeConfiguration`/`image`/hosted) instead of `docker:`, matching the deploy-time contract.

## Validation

- Added unit tests for the brownfield ACR logic, the `--infra` refusal, and expanded synthesizer cases (schema-conformant hosted, `kind` omitted, `image:`, `codeConfiguration:`).
- `go build`, `gofmt -s`, `golangci-lint` (0 issues), full `go test ./...`, and the bicep drift test all pass.
- Verified end-to-end on a live subscription: `azd up` now creates the ACR + connection, sets `AZURE_CONTAINER_REGISTRY_ENDPOINT`, and publish/deploy succeed; `provision --preview` renders the Create plan; `init --infra` shows the clear message. Test resources were cleaned up.


cli/azd/extensions/azure.ai.agents/internal/project/foundry_provisioning_provider.go
```diff
@@ -652,26 +659,181 @@ func (p *FoundryProvisioningProvider) deployBrownfield(
...
+// environments that reuse the same Foundry account.
+func (p *FoundryProvisioningProvider) brownfieldACRName(account string) string {
+	h := fnv.New32a()
+	_, _ = h.Write([]byte(account + "|" + p.brownfieldProjectName() + "|" + p.envName))
+	return fmt.Sprintf("acr%08x", h.Sum32())
+}
+
+// brownfieldLocation resolves the region for the new registry from AZURE_LOCATION
+// (seeded by init), falling back to the existing resource group's location.
+func (p *FoundryProvisioningProvider) brownfieldLocation(ctx context.Context, rg string) string {
```
jongio, line 817: If both `AZURE_LOCATION` and the resource-group GET fail, this returns `""`. The ARM template declares `defaultValue: resourceGroup().location` for the `location` param, but ARM treats an explicit empty-string parameter as an override (not a missing value), so the default won't kick in. Consider omitting the `location` key from the params map entirely when the resolved value is empty, or failing early with a clear error asking the user to set `AZURE_LOCATION`.

cli/azd/extensions/azure.ai.agents/internal/synthesis/synthesizer.go
```diff
@@ -311,13 +324,33 @@ func deriveIncludeAcr(services map[string]yaml.Node, svc projectService) bool {
...
+		if agentNeedsAcr(agentBlock{
+			Kind:              service.Kind,
+			Image:             service.Image,
+			CodeConfiguration: service.CodeConfiguration,
+		}) {
 			return true
 		}
 	}
 	return false
 }
```
jongio, line 338: The empty-kind default to `hosted` is correct for back-compat, but it means any future non-container kind that happens to omit `kind:` would accidentally trigger ACR creation. A brief inline comment like `// Update this list if a new non-container kind is added` (or switching to an explicit allowlist check) would make this safer to evolve.

jongio (commented):
Solid fix for the brownfield ACR gap. The two root causes are correctly identified and addressed: the bicepless flow not configuring ACR for container agents, and `deriveIncludeAcr` keying on the absent `docker:` block instead of the schema fields (`codeConfiguration`/`image`/kind). The brownfield Bicep template additions (conditional ACR + AcrPull role + connection) are well-structured.

A few observations:

- `brownfieldLocation` can return `""` when both `AZURE_LOCATION` and the resource-group lookup fail. The ARM template has `defaultValue: resourceGroup().location` for the `location` param, but passing an explicit empty string overrides that default rather than falling back to it. If there's a realistic path where neither source succeeds, the deploy would fail with a confusing ARM error about an invalid location.

- `brownfield.arm.json` is missing a trailing newline (the diff shows `No newline at end of file`). Minor, but linters and some editors flag this.

- The `agentNeedsAcr` helper defaults empty `kind` to "hosted" for back-compat. If a new kind is added in the future that also doesn't build containers (e.g., "serverless"), it would incorrectly trigger ACR creation. A comment noting this would help future contributors remember to update the function.

Test coverage is thorough: `TestBrownfieldACRRequested`, `TestBrownfieldACRName`, `TestBrownfieldDeploymentName`, and the synthesizer cases all cover the key scenarios well.

cli/azd/extensions/azure.ai.agents/internal/project/foundry_provisioning_provider.go
```diff
@@ -652,26 +659,181 @@ func (p *FoundryProvisioningProvider) deployBrownfield(
...
+		"accountName": map[string]any{"value": account},
+		"deployments": map[string]any{"value": p.brownfieldDeployments},
+	}
+	if createACR {
+		params["includeAcr"] = map[string]any{"value": true}
+		params["acrName"] = map[string]any{"value": p.brownfieldACRName(account)}
+		params["projectName"] = map[string]any{"value": p.brownfieldProjectName()}
+		params["location"] = map[string]any{"value": p.brownfieldLocation(ctx, rg)}
+		params["tags"] = map[string]any{"value": map[string]string{"azd-env-name": p.envName}}
+	}
```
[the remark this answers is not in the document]
hund030: Fixed in a5785cb5f: `brownfieldParams` now only sets the `location` parameter when `brownfieldLocation()` returns a non-empty value, so an unresolved location falls back to the template default `resourceGroup().location` instead of overriding it with `"`. Added a unit-test case covering the omitted-location path.

cli/azd/extensions/azure.ai.agents/internal/cmd/init_foundry_resources_helpers.go
```diff
@@ -382,9 +382,8 @@ func lookupAcrResourceId(
 // configureFoundryProjectEnv sets all Foundry project environment variables and discovers
 // ACR and AppInsights connections. This is the shared implementation used by both init flows.
 // When skipACR is true, ACR connection discovery and configuration is skipped (used for code deploy).
-// When bicepless is true, both ACR and AppInsights connection discovery
-// and prompting are skipped after the basic project identity env vars
-// are seeded; the extension's provisioning provider owns those resources.
+// When bicepless is true, AppInsights is left to the provisioning provider; ACR is still configured
+// for a container agent (skipACR false), since the provider does not create one for an existing project.
```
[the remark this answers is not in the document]
hund030: Fixed in a5785cb5f: updated the `configureFoundryProjectEnv` doc comment to reflect that ACR is wired here from an existing connection or the provider is signaled to create one on provision (it no longer claims the provider does not create one for an existing project).

cli/azd/extensions/azure.ai.agents/internal/project/foundry_provisioning_provider.go
hund030: Fixed in a5785cb5f: `brownfieldParams` now omits the `location` key when `brownfieldLocation()` resolves to `"`, so ARM uses the template default `resourceGroup().location` instead of an empty-string override. Added a unit-test case for the omitted-location path. (In practice `AZURE_LOCATION` is always seeded by init for an existing project; this guards the fallback path.)

cli/azd/extensions/azure.ai.agents/internal/synthesis/synthesizer.go
hund030: Added a note in 483170890: the comment now flags that a future non-container kind which omits `kind:` should switch this to an explicit allowlist rather than defaulting to hosted.

hund030:
Thanks for the review @jongio. Addressed all three observations:

1. **`brownfieldLocation` empty override** (a5785cb5f) — `brownfieldParams` now omits the `location` param when unresolved, so the template default `resourceGroup().location` applies instead of an empty-string override. Added a unit test.
2. **`brownfield.arm.json` trailing newline** (483170890) — added. (Note: `main.arm.json` is also generated without one by `bicep build`; the drift test normalizes JSON so this is cosmetic.)
3. **`agentNeedsAcr` empty-kind default** (483170890) — added a comment to switch to an explicit allowlist if a future non-container kind can omit `kind:`.

Also addressed the two Copilot comments (same location fix + the stale `configureFoundryProjectEnv` doc comment). `go build` / `gofmt` / `golangci-lint` (0 issues) / full `go test` all pass.

trangevi (approved):

PR #8883, opened 2026-06-30
fix: prompt for subscription and location on foundry azd up

### LINKED ISSUE #8859: azd up does not prompt subscription ID
_Migrated from [huimiu/azd-foundry-extension-bugbash#9](https://github.com/huimiu/azd-foundry-extension-bugbash/issues/9), originally reported by @therealjohn._

---

Step #3 says to run azd up in the unified sample directory, but it fails because the sub ID isnt set: 

Image

huimiu:
Fixes #8859

### Problem

Running `azd up` in a Foundry (`azure.ai.agents`) project fails immediately when `AZURE_SUBSCRIPTION_ID` (or `AZURE_LOCATION`) is not set in the azd environment:

```
ERROR: AZURE_SUBSCRIPTION_ID is required but not set in azd environment "foundry-bugbash"
Suggestion: run `azd env set AZURE_SUBSCRIPTION_ID <subscription-id>`
```

Core `azd up` prompts the user to pick a subscription/location when they aren't set; the extension's provisioning provider instead hard-failed.

### Root cause

`FoundryProvisioningProvider.resolveEnv` read `AZURE_SUBSCRIPTION_ID` / `AZURE_LOCATION` from the environment and returned a `Dependency` error when either was empty, rather than prompting. The extension already prompts in its `init` flow (`PromptSubscription` / `PromptLocation`), but the provision path (`azd up` / `azd provision`) never adopted it.

### Fix

When `AZURE_SUBSCRIPTION_ID` / `AZURE_LOCATION` are missing, `resolveEnv` now prompts via the azdext Prompt service and persists the chosen values to the azd environment, matching core `azd up`.

- New `promptSubscription` / `promptLocation` / `setEnv` helpers on the provider.
- Deterministic `--no-prompt` / CI behavior preserved: when the host returns a "prompt required" error, the original actionable `azd env set ...` suggestion is still surfaced.
- User cancellation maps to the cancellation category.
- Tenant/credential resolution is unchanged (`ensureCredential` + `withTenantOutput`).
- Brownfield (`endpoint:`) path is untouched — it uses `resolveEnvName` and needs no subscription/location.

### Tests

Added `foundry_provisioning_provider_resolveenv_test.go` covering: prompt-and-persist for both values; `--no-prompt` actionable errors for missing subscription and location; and user-cancellation handling. Full `internal/project` suite and `go build ./...` pass; `gofmt`, `golangci-lint`, and `cspell` are clean.

### DIFF
--- a/cli/azd/extensions/azure.ai.agents/internal/project/foundry_provisioning_provider_resolveenv_test.go
+++ b/cli/azd/extensions/azure.ai.agents/internal/project/foundry_provisioning_provider_resolveenv_test.go
@@ -25,6 +25,7 @@ type resolveEnvStubEnvServer struct {
 	azdext.UnimplementedEnvironmentServiceServer
 	envName string
 	get     map[string]string
+	getErr  map[string]error
 	set     map[string]string
 }
 
@@ -37,6 +38,9 @@ func (s *resolveEnvStubEnvServer) GetCurrent(
 func (s *resolveEnvStubEnvServer) GetValue(
 	_ context.Context, req *azdext.GetEnvRequest,
 ) (*azdext.KeyValueResponse, error) {
+	if err := s.getErr[req.Key]; err != nil {
+		return nil, err
+	}
 	return &azdext.KeyValueResponse{Value: s.get[req.Key]}, nil
 }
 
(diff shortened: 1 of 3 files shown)


cli/azd/extensions/azure.ai.agents/internal/project/foundry_provisioning_provider_resolveenv_test.go
```diff
@@ -0,0 +1,202 @@
...
+	client := newResolveEnvTestClient(t, env, prompt)
+
+	p := &FoundryProvisioningProvider{azdClient: client}
+	err := p.resolveEnv(t.Context())
+	require.Error(t, err)
+
+	var local *azdext.LocalError
+	require.ErrorAs(t, err, &local)
+	assert.Equal(t, exterrors.CodeCancelled, local.Code)
+}
```
jongio, line 206: The subscription cancellation test (`TestResolveEnv_CancelledSubscriptionPromptReturnsCancelled`) exists, but there's no equivalent for location cancellation. If the user selects a subscription successfully then cancels the location prompt, the extension should still return a cancellation error (not persist a partial state). Consider adding:

```suggestion
func TestResolveEnv_CancelledLocationPromptReturnsCancelled(t *testing.T) {
	env := &resolveEnvStubEnvServer{
		envName: "foundry-bugbash",
		get:     map[string]string{envKeySubscriptionID: "00000000-0000-0000-0000-000000000001"},
	}
	prompt := &resolveEnvStubPromptServer{
		locationErr: status.Error(codes.Canceled, "user cancelled"),
	}
	client := newResolveEnvTestClient(t, env, prompt)

	p := &FoundryProvisioningProvider{azdClient: client}
	err := p.resolveEnv(t.Context())
	require.Error(t, err)

	var local *azdext.LocalError
	require.ErrorAs(t, err, &local)
	assert.Equal(t, exterrors.CodeCancelled, local.Code)
}
```

jongio (commented):
Solid fix. The prompt-then-persist approach matches the init flow in `init_foundry_resources_helpers.go` and aligns with core `azd up` behavior. Error handling (cancellation, prompt-required, generic fallback) follows the established tri-branch pattern correctly.

One gap in test coverage noted below.

cli/azd/extensions/azure.ai.agents/internal/project/foundry_provisioning_provider_resolveenv_test.go
```diff
@@ -0,0 +1,266 @@
...
+
+	p := &FoundryProvisioningProvider{azdClient: client}
+	err := p.resolveEnv(t.Context())
+	require.Error(t, err)
+
+	var local *azdext.LocalError
+	require.ErrorAs(t, err, &local)
+	assert.Equal(t, exterrors.CodeMissingAzureSubscription, local.Code)
+	assert.Empty(t, env.set, "an empty subscription id must not be persisted")
+}
```
jongio, line 266: Nit: there's `TestResolveEnv_SubscriptionReadErrorSurfaces` and `TestResolveEnv_EmptySubscriptionResponseReturnsError` but no location-side equivalents. The production code is structurally identical so risk is low, but adding `TestResolveEnv_LocationReadErrorSurfaces` and `TestResolveEnv_EmptyLocationResponseReturnsError` would lock in both paths. Optional, can be a follow-up.

jongio (commented):
Prior feedback addressed: read-error separation, empty-response validation, and location cancellation test all look correct. CI is green.

One minor coverage nit below (non-blocking).

trangevi (approved):

Issue #8884, opened 2026-06-30
[ext-agents]: azd ai agent init -m <azure.yaml> should merge into an existing project's azure.yaml


huimiu:
## Summary

Follow-up to #8798. The initial implementation (PR adopting a sample's unified `azure.yaml` via `azd ai agent init -m`) handles the **fresh-project** case: when the target directory has no `azure.yaml`, the sample's `azure.yaml` (and referenced files) is adopted as the project manifest via azd-core's native template adoption.

The **existing-project** case is not yet supported. When a project-root `azure.yaml` already exists and the user runs `azd ai agent init -m <sample azure.yaml>`, the command currently returns a clear, actionable error rather than merging.

## Desired behavior

Fall back to today's "add to project" behavior: append the sample's **agent service(s)** (and any `uses:`-referenced `azure.ai.project` / `azure.ai.connection` / `azure.ai.toolbox` sibling services not already present) into the existing `azure.yaml`, copying each agent's `project:` source directory into the project. Resolve service-name collisions (reuse `resolveCollisions`), and skip siblings that already exist so dependencies still resolve.

## Why it was split out

Correctly appending a unified `azure.yaml`'s services into an existing project requires careful per-host service round-tripping through `Project().AddService` — e.g. `uses:` is carried in `AdditionalProperties` and wired separately in the manifest flow, and each Foundry host promotes a different set of fields. Getting this wrong can drop or duplicate keys in the written `azure.yaml`, and it's hard to verify without live gRPC. Splitting it keeps the primary fresh-project adoption shippable and well-tested.

## Acceptance

- `azd ai agent init -m <sample azure.yaml>` in a directory with an existing project `azure.yaml` appends the sample's agent service(s) + missing referenced siblings, copying source dirs, with collisions resolved.
- The written `azure.yaml` round-trips cleanly (no dropped/duplicated keys; `uses:` preserved).
- Tests cover the merge (including name collisions and already-present siblings).

## Related

- #8798
- Current interim behavior: returns a validation error pointing the user to run in an empty directory or add an individual agent via `-m <agent.manifest.yaml>`.

PR #8885, opened 2026-06-30
feat: adopt unified azure.yaml in azd ai agent init -m

### LINKED ISSUE #8798: [ext-agents]: azd ai agent init should adopt the sample's azure.yaml, not scaffold/clone from azd-ai-starter-basic (unified azure.yaml + bicep-less)
## Summary

Today `azd ai agent init` points at an agent manifest (`agent.manifest.yaml`) and **generates/scaffolds** the project-root `azure.yaml` (and, on `main`, an `infra/` directory cloned from `Azure-Samples/azd-ai-starter-basic`). Any `azure.yaml` that ships in the sample is downloaded into the agent's `src/<agent-id>/` subfolder and **ignored** by azd.

Under the unified `azure.yaml` work on `huimiu/foundry-azure-yaml`, this needs to change. `init` will no longer point to an agent manifest -- samples will ship their configuration as a complete unified `azure.yaml` (`host: microsoft.foundry`). `init` should **consume the sample's `azure.yaml`** as the project manifest instead of scaffolding/cloning its own, and it should **stop downloading `infra/` + `azure.yaml` from `azd-ai-starter-basic`**, because bicep-less provisioning (`infra.provider: microsoft.foundry`) is the default under the unified work (#8643).

## Current behavior (on `main`)

1. `ensureProject` finds no existing project, so it runs `azd init -t Azure-Samples/azd-ai-starter-basic <targetDir>`, which scaffolds the **project-root** `azure.yaml` + `infra/` from the starter repo, then `chdir`s into it. (`cli/azd/extensions/azure.ai.agents/internal/cmd/init.go`, `ensureProject` ~L1489-1544; starter template at ~L1517-1518)
2. For hosted/container agents, `downloadParentDirectory` downloads the **entire** remote manifest folder (no filename filtering) into `src/<agent-id>/` -- including the sample's own `azure.yaml`, which lands at `src/<agent-id>/azure.yaml` and is never read by azd (azd only reads the project-root `azure.yaml`). (`init.go` `downloadAgentYaml` ~L2472-2495, `downloadParentDirectory` / `downloadDirectoryContents` ~L3120-3222)
3. The agent is wired in via `Project().AddService()`, which appends a service entry to the project-root `azure.yaml`. (`init.go` `addToProject` ~L2691; host gRPC handler `cli/azd/internal/grpcserver/project_service.go` `AddService` ~L181-232)

Net result: two `azure.yaml` files -- the scaffolded project-root one (used) and the sample's copy buried under `src/<agent-id>/` (unused).

> Note: #8643 (already merged into `huimiu/foundry-azure-yaml`) replaced the `azd-ai-starter-basic` clone with an `azd init -t <emptyStagedDir>` scaffold plus a generated bicep-less `azure.yaml` (`infra.provider: microsoft.foundry`). That removed the starter `infra/` clone, but `init` still **generates** the project `azure.yaml` rather than adopting the sample's.

## Why this needs to change

The unified `azure.yaml` direction (#8676, #8773, #8709, #8710, and the `huimiu/foundry-azure-yaml` base branch):

- `init` no longer points to an `agent.manifest.yaml`; `agent.yaml` / `agent.manifest.yaml` are retired in favor of a single unified `azure.yaml` using `host: microsoft.foundry`.
- Samples will therefore ship a complete unified `azure.yaml` that already declares the project, model deployments, connections, toolboxes, and agent(s).
- Bicep-less provisioning is the default (#8643), so there is no need to pull `infra/` (or a separate `azure.yaml`) from `azd-ai-starter-basic`.

Continuing to scaffold/generate a project `azure.yaml` (or clone one from the starter repo) and bury the sample's `azure.yaml` under `src/<agent-id>/` is wrong under this model: the sample's `azure.yaml` is the source of truth.

## Proposed behavior

When initializing from a sample/URL that ships a unified `azure.yaml`:

1. Use the **sample's `azure.yaml`** as the project-root manifest (place it at the project root) instead of scaffolding from `azd-ai-starter-basic` or generating a new one.
2. Do **not** download/clone `infra/` + `azure.yaml` from `azd-ai-starter-basic`; rely on bicep-less provisioning (`infra.provider: microsoft.foundry`) by default.
3. Do not leave an unused/duplicate `azure.yaml` under `src/<agent-id>/`.
4. Preserve existing-project behavior: if a project-root `azure.yaml` already exists on disk, do not overwrite it (merge/add as appropriate).

## Notes / open questions

- Define the fallback when a sample does **not** include an `azure.yaml` (backward compatibility): generate the bicep-less unified `azure.yaml` as #8643 does today.
- Reconcile with the per-resource service split (#8675): the adopted `azure.yaml` should already contain the `azure.ai.project` / `azure.ai.connection` / `azure.ai.toolbox` + agent service entries, so `init` should not re-derive them.
- Decide how `--src`, `--agent-name`, and folder-naming interact when the sample's `azure.yaml` already names services / paths.
- Update next-step / doctor flows accordingly (tracked in #8710).

## Related

- Base branch: https://github.com/Azure/azure-dev/tree/huimiu/foundry-azure-yaml
- #8676 [ext-agents]: azd ai agent init writes unified microsoft.foundry config in azure.yaml
- #8643 feat(agents): bicep-less init for azd ai agent init
- #8675 feat(agents): split Foundry init into per-resource azure.yaml services
- #8773 [ext-agents]: evolve azure.ai.agent to service-level shape, retire agent.yaml, add migration
- #8774 [ext-foundry]: make project/connection/toolbox Foundry resources real service targets
- #8709 Migrate rai policy to unified azure.yaml
- #8710 Migrate next step suggestions and doctor command to make it work with unified azure.yaml

huimiu:
## What

`azd ai agent init -m <pointer>` now accepts a sample's unified Foundry `azure.yaml` as an input. When the pointer resolves to a unified project manifest, `init` adopts that `azure.yaml` as the new project's manifest instead of generating one from an agent manifest.

Closes part of #8798 (fresh-project adoption). Existing-project merge is tracked separately in #8884.

### `-m <azure.yaml>`

When `-m` points to a unified `azure.yaml` / `azure.yml`:

1. `init` treats it as a complete project manifest when its `services:` contain Foundry hosts such as `azure.ai.project`, `azure.ai.agent`, `azure.ai.connection`, or `azure.ai.toolbox`.
2. A new azd project is initialized from the sample content.
3. The sample's `azure.yaml` becomes the project-root `azure.yaml`.
4. Files referenced by that manifest, including agent source folders and `$ref` files, keep their sample-relative layout.
5. A normal azd environment is created under `.azure/<env>`.
6. Bicep-less Foundry provisioning remains the default through `infra.provider: microsoft.foundry`.

Resulting shape:

```text
<project-folder>/
  azure.yaml          # adopted from the sample
  .azure/<env>/       # azd environment
  <sample files...>   # source folders, $ref files, README, etc.
```

This path does **not** generate agent services from an agent manifest, does **not** bury the sample's `azure.yaml` under `src/<agent>/`, and does **not** duplicate Foundry resources that are already declared in the unified manifest.

### `-m <agent.manifest.yaml>`

Existing behavior is preserved. The agent manifest remains a template input: `init` resolves manifest parameters and prompts, then generates or updates the project `azure.yaml` from that manifest.

## Scope

This PR covers fresh-project adoption. If the target directory already has `azure.yaml` / `azure.yml`, adoption is refused with guidance instead of overwriting or merging. Merging a sample's services into an existing project is tracked in #8884.

## Tests

Added coverage for unified-manifest detection, local/GitHub staging behavior, `azure.yml` normalization, target-directory conflict checks, and staging cleanup. Full agents-extension checks are passing.


cli/azd/extensions/azure.ai.agents/internal/cmd/init_adopt.go
```diff
@@ -0,0 +1,387 @@
...
+				"unified azure.yaml cannot be adopted here",
+			"run this command in an empty directory (or pass a new target directory) to "+
+				"adopt the sample, or add an individual agent to this project with "+
+				"'azd ai agent init -m <agent.manifest.yaml>'",
+		)
+	}
+
+	// Stage the sample as a local template directory (azure.yaml at its root
+	// alongside referenced files) that azd-core can adopt with `azd init -t`.
+	stagingDir, cleanup, err := stageAzureYamlTemplate(ctx, flags, azdClient, httpClient)
```
jongio, line 189: The conflict check uses a bare relative path (`fileExists("azure.yaml")`), which checks the CWD. If the user passes `--src my-new-dir` while standing in an existing project directory, this blocks adoption even though the target directory is empty.

Consider checking in the resolved `targetDir` instead (call `adoptTargetDir` first, then check for `azure.yaml` there), or documenting that the user must cd into the target first.

cli/azd/extensions/azure.ai.agents/internal/cmd/init_adopt.go
```diff
@@ -0,0 +1,387 @@
...
+) (string, func(), error) {
+	noop := func() {}
+	pointer := flags.manifestPointer
+
+	if isLocalFilePath(pointer) {
+		dir := filepath.Dir(pointer)
+		base := strings.ToLower(filepath.Base(pointer))
+		if base == "azure.yaml" || base == "azure.yml" {
+			return dir, noop, nil
+		}
```
jongio, line 277: When `base` matches `azure.yml` (not `azure.yaml`), this returns the parent dir directly. That works because `azd init -t` copies the directory wholesale and the `.yml` extension is recognized downstream. But it might be worth a test case for the `.yml` variant to lock that behavior in, since `stagedAzureYamlExists` already handles both extensions.

cli/azd/extensions/azure.ai.agents/internal/cmd/init_adopt.go
```diff
@@ -0,0 +1,387 @@
...
+		// The pointer file isn't named azure.yaml: stage a temp copy of the
+		// directory and write the manifest as azure.yaml so azd-core adopts it.
+		staging, err := os.MkdirTemp("", "azd-foundry-adopt-*")
+		if err != nil {
+			return "", noop, fmt.Errorf("creating staging dir: %w", err)
+		}
+		cleanup := func() { _ = os.RemoveAll(staging) }
+		if err := copyDirectory(dir, staging); err != nil {
+			cleanup()
+			return "", noop, fmt.Errorf("staging sample directory: %w", err)
```
jongio, line 291: If `copyDirectory` fails partway through, the cleanup lambda removes the staging dir. If `os.ReadFile` succeeds but `os.WriteFile` fails, staging has a partial copy without the manifest. Cleanup handles this fine, but a brief comment noting the all-or-nothing intent (staging is useless without `azure.yaml` at root) would help future readers.

cli/azd/extensions/azure.ai.agents/internal/cmd/init_adopt.go
```diff
@@ -0,0 +1,387 @@
...
+func stageRemoteAzureYaml(
+	ctx context.Context,
+	azdClient *azdext.AzdClient,
+	httpClient *http.Client,
+	pointer string,
+	staging string,
+) error {
+	fmt.Println(output.WithGrayFormat("Downloading sample from GitHub..."))
+
+	if urlInfo := parseGitHubUrlNaive(pointer); urlInfo != nil {
```
jongio, line 378: After the naive public download succeeds (`err == nil`) but `stagedAzureYamlExists` returns false, the code falls through to the gh CLI path which downloads into the same staging dir. If the public download left partial files there, the gh CLI download writes on top. This is probably fine since both resolve the same directory contents, but if the repo changed between calls you could get mixed state. Consider clearing `staging` before the fallback, or noting this is intentional.

jongio (commented):
4 files changed (+722 -68). Clean refactor of `ensureProject` into reusable `scaffoldProject`/`deriveEnvName`, plus a new adoption path for unified azure.yaml manifests.

The detection logic (`looksLikeFoundryAzureYaml`) is solid and the routing between adoption vs generation is unambiguous. Test coverage on the new helpers is thorough.

A few items worth considering inline.

cli/azd/extensions/azure.ai.agents/internal/cmd/init_adopt.go
```diff
@@ -0,0 +1,523 @@
...
+	}
+	if azdClient == nil || !strings.Contains(manifestPointer, "://") {
+		return nil, false
+	}
+
+	parsedURL, err := url.Parse(manifestPointer)
+	if err != nil || !strings.Contains(parsedURL.Hostname(), "github") {
+		return nil, false
+	}
+
```
jongio, line 114: Nit: the command runner here sends gh CLI output (including potential 404/auth-error messages) to `os.Stdout`/`os.Stderr`. Since this is a detection function that should be transparent to the user when it fails (graceful fallback to `nil, false`), consider using `io.Discard` for both streams so transient GitHub errors don't appear on-screen before the fallthrough kicks in.

The same pattern in `stageRemoteAzureYaml` is fine because that function is past the detection phase and the user expects download activity. Here the contract is 'peek silently, fall through if it doesn't work.'

Not blocking - the behavior is correct either way.

jongio (commented):
Incremental review (1 new commit since my last pass: 2c38f4).

Prior findings are all addressed:
- Conflict check now correctly uses projectManifestExists(targetDir) instead of CWD-relative ileExists
- zure.yml is properly normalized to zure.yaml via nsureStagedAzureYaml for both local and remote staging
- Staged source manifest is removed to prevent duplicates
- Private repo detection now works through the authenticated gh CLI fallback

One minor nit inline.

jongio (commented):
Incremental review (1 new commit since my last pass: `02c38f4`).

Prior findings are all addressed:
- Conflict check now correctly uses `projectManifestExists(targetDir)` instead of CWD-relative `fileExists`
- `azure.yml` is properly normalized to `azure.yaml` via `ensureStagedAzureYaml` for both local and remote staging
- Staged source manifest is removed to prevent duplicates
- Private repo detection now works through the authenticated gh CLI fallback

One minor nit inline (on the prior review submission - formatting was fixed).

cli/azd/extensions/azure.ai.agents/internal/cmd/init_adopt.go
```diff
@@ -0,0 +1,524 @@
...
+// foundryServiceHosts are the azure.yaml service `host` values that identify a
+// unified Microsoft Foundry project manifest. The legacy `microsoft.foundry`
+// host is included for backward compatibility with older non-split files.
+var foundryServiceHosts = map[string]struct{}{
+	"azure.ai.agent":      {},
+	"azure.ai.project":    {},
+	"azure.ai.connection": {},
+	"azure.ai.toolbox":    {},
+	"microsoft.foundry":   {},
```
trangevi, line 39: This host was never shipped, was it? Before the extension split, we just used azure.ai.agent, and after the split I know we talked about microsoft.foundry for a time, but I don't think it was shipped. I really don't think we should support the "microsoft.foundry" host for anything.

cli/azd/extensions/azure.ai.agents/internal/cmd/init_adopt.go
```diff
@@ -0,0 +1,524 @@
...
+	"github.com/azure/azure-dev/cli/azd/pkg/osutil"
+	"github.com/azure/azure-dev/cli/azd/pkg/output"
+	"github.com/azure/azure-dev/cli/azd/pkg/tools/github"
+	"gopkg.in/yaml.v3"
+)
+
+// foundryServiceHosts are the azure.yaml service `host` values that identify a
+// unified Microsoft Foundry project manifest. The legacy `microsoft.foundry`
+// host is included for backward compatibility with older non-split files.
+var foundryServiceHosts = map[string]struct{}{
```
trangevi, line 34: What about azure.ai.skills? In general, it doesn't feel great that we're going to need to maintain a list of services that are relevant. In theory, the agent extension should only care about the agent service I feel like.

trangevi (approved):

cli/azd/extensions/azure.ai.agents/internal/cmd/init_adopt.go
jongio: Confirmed. I don't see `microsoft.foundry` used as a service `host:` value anywhere in the repo. It only exists as an extension ID (the meta-package in `cli/azd/extensions/microsoft.foundry/extension.yaml`). Dropping it from this list seems safe.

cli/azd/extensions/azure.ai.agents/internal/cmd/init_adopt.go
jongio: The `microsoft.foundry` meta-package bundles 7 extensions (agents, connections, inspector, projects, routines, skills, toolboxes), but only 4 appear in this detection list. Missing: `azure.ai.skills`, `azure.ai.routines`, `azure.ai.inspector`. That confirms the maintenance concern.

A prefix check (e.g., `strings.HasPrefix(host, "azure.ai.")`) would catch current and future Foundry hosts without needing updates. The only risk is false-positiving on a non-Foundry host that starts with `azure.ai.`, but that prefix is exclusively Foundry-owned today.

trangevi:

cli/azd/extensions/azure.ai.agents/internal/cmd/init_adopt.go
trangevi: Filled an issue to address this as a follow up, as it's not blocking: https://github.com/Azure/azure-dev/issues/8888

cli/azd/extensions/azure.ai.agents/internal/cmd/init_adopt.go
trangevi: Filed an issue to address this as a follow up, as I don't believe it to be a blocking item at the moment: https://github.com/Azure/azure-dev/issues/8888

trangevi:

therealjohn (approved):

PR #8896, opened 2026-06-30
Prepare Foundry extensions for 1.0.0-beta.1 release

trangevi:
Fixes: #8892

## Motivation

The Foundry extensions are graduating from preview to beta. This PR bumps all seven Foundry extensions to `1.0.0-beta.1` and updates their changelogs with properly formatted entries for all PRs merged since their last releases.

## Changes

**Version bump (all 7 extensions):**
- `azure.ai.agents`: 0.1.42-preview -> 1.0.0-beta.1
- `azure.ai.connections`: 0.1.3-preview -> 1.0.0-beta.1
- `azure.ai.inspector`: 0.0.1-preview -> 1.0.0-beta.1
- `azure.ai.projects`: 0.1.1-preview -> 1.0.0-beta.1
- `azure.ai.routines`: 0.1.1-preview -> 1.0.0-beta.1
- `azure.ai.skills`: 0.1.1-preview -> 1.0.0-beta.1
- `azure.ai.toolboxes`: 0.1.2-preview -> 1.0.0-beta.1
- `microsoft.foundry`: 0.1.0-preview -> 1.0.0-beta.1

**Branding:**
- Updated `displayName` and `description` from "(Preview)" to "(Beta)" across all extensions

**Dependency versions:**
- Updated `microsoft.foundry` meta-extension dependency constraints from `~0.1.0-preview` to `~1.0.0-beta.1` for all bundled extensions

**Changelogs:**
- Reformatted existing Unreleased entries to standard format with `[[#NNNN]](url)` PR references
- Added missing entries for all PRs merged since each extension's last release
- Used categorized subsections (`### Features Added`, `### Breaking Changes`, `### Bugs Fixed`)
- Added contributor attribution for external (non-CODEOWNERS) authors only

PR #8915, opened 2026-07-01
fix(ci): build azure.ai.inspector from source in test workflow

### LINKED ISSUE #8916: CI: test-ext-azure-ai-agents workflow broken — inspector dependency version mismatch
## Problem

All runs of `test-ext-azure-ai-agents.yml` have been failing since PR #8896 (1.0.0-beta.1 version bump) with:

```
ERROR: failed to install extension: installed dependency azure.ai.inspector version 0.0.1-preview does not satisfy constraint "~1.0.0-beta.1"
```

The CI workflow installs `azure.ai.inspector` from the registry, but only `0.0.1-preview` is published there. The source code was bumped to `1.0.0-beta.1` but not released to the registry.

This blocks all PRs touching the agents extension.

## Expected behavior

CI should build `azure.ai.inspector` from source (same as `azure.ai.agents`) so the dependency constraint resolves correctly.

v1212:
## Problem

All runs of `test-ext-azure-ai-agents.yml` are failing since the 1.0.0-beta.1 version bump (#8896):

```
ERROR: failed to install extension: installed dependency azure.ai.inspector version 0.0.1-preview does not satisfy constraint "~1.0.0-beta.1"
```

**Root cause**: The workflow installs `azure.ai.inspector` from the registry, but only `0.0.1-preview` has been published there. The source was bumped to `1.0.0-beta.1` but never released to the registry.

This blocks **all PRs** touching the agents extension (8 different branches affected today).

Fixes #8916

## Fix

Build and install `azure.ai.inspector` from source using `ci-build.ps1` + `azd x pack/publish` -- the same approach already used for `azure.ai.agents` itself.

Also adds `cli/azd/extensions/azure.ai.inspector/**` to the workflow path trigger so inspector source changes trigger the test.

## Verification

- Confirmed all 8 failed runs from Jun 30-Jul 1 show the identical error
- The fix mirrors the existing pattern for building azure.ai.agents from source
- CI `test` job passes on this PR

### DIFF
--- a/.github/workflows/test-ext-azure-ai-agents.yml
+++ b/.github/workflows/test-ext-azure-ai-agents.yml
@@ -45,8 +45,9 @@ jobs:
       - name: Build & install inspector dependency from source
         working-directory: cli/azd/extensions/azure.ai.inspector
         run: |
-          # The azure.ai.inspector extension has not been published to the registry
-          # at 1.0.0-beta.1+ yet — build from source to satisfy the dependency.
+          # Build azure.ai.inspector from source to ensure PR CI always has a
+          # compatible version — dependency releases are post-merge, but PRs may
+          # include inspector changes from this same repo.
           pwsh -File ci-build.ps1 -OutputFileName azure-ai-inspector-linux-amd64 -Version $(cat version.txt | tr -d '\r\n')
           mkdir -p bin
           mv azure-ai-inspector-linux-amd64 bin/
--- a/.github/workflows/test-ext-azure-ai-agents.yml
+++ b/.github/workflows/test-ext-azure-ai-agents.yml
@@ -4,6 +4,7 @@ on:
   pull_request:
     paths:
       - "cli/azd/extensions/azure.ai.agents/**"
+      - "cli/azd/extensions/azure.ai.inspector/**"
       - "cli/azd/test/functional/ai_agents/**"
       - ".github/workflows/test-ext-azure-ai-agents.yml"
     branches: [main]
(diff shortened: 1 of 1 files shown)


.github/workflows/test-ext-azure-ai-agents.yml
```diff
@@ -40,8 +41,18 @@ jobs:
...
           # The built-in 'azd' source is auto-registered, so no source-add needed.
           azd extension install microsoft.azd.extensions --no-prompt
-          # Pre-install the inspector dependency so the local install can resolve it.
-          azd extension install azure.ai.inspector --no-prompt
+
+      - name: Build & install inspector dependency from source
+        working-directory: cli/azd/extensions/azure.ai.inspector
+        run: |
+          # The azure.ai.inspector extension has not been published to the registry
+          # at 1.0.0-beta.1+ yet — build from source to satisfy the dependency.
```
RickWinter: Small wording point with a real intent behind it: the word "yet" reads as if this from-source build is a temporary workaround until inspector lands in the registry. Given we're now also triggering this workflow on `azure.ai.inspector/**` (the new path on line 7), building from source is actually the behavior we want to keep, since it tests each PR's own inspector code and avoids reintroducing this same resolution failure on the next version bump. Can we reword the comment to frame it as intentional (build inspector from source so the suite exercises PR changes) rather than a stopgap?

RickWinter (commented):
This unblocks the agents extension test workflow by building `azure.ai.inspector` from source instead of pulling it from the registry, where only `0.0.1-preview` exists against the new `~1.0.0-beta.1` constraint. The approach is the right shape: it mirrors the existing from-source build/pack/publish/install pattern used for `azure.ai.agents` in the same job, and the inspector `version.txt` (1.0.0-beta.1) satisfies the dependency constraint, so the resolution error goes away.

Nothing here blocks merge. One thing worth deciding explicitly rather than by accident: pairing the new `azure.ai.inspector/**` path trigger with a from-source build means inspector-only PRs now build and test their own inspector code, which is the behavior we want. If we ever revert to installing the published inspector, we reintroduce exactly this failure on the next version bump and stop testing PR inspector changes, so I'd keep the from-source build as the intended long-term state and adjust the comment to say so.

The redundant `-Version` argument and the `--force` on install are both consistent with the agents step directly below, so leaving them as-is is correct for readability parity. No security or correctness concerns in the diff.

jongio (approved):
+1 on @RickWinter's inline: the comment on line 49 should frame building from source as the intended approach (it tests PR inspector changes and avoids future version-bump breakage), not a stopgap until the registry catches up. Minor wording fix, doesn't block merge.

.github/workflows/test-ext-azure-ai-agents.yml
v1212: Good call — updated the comment to frame this as intentional. The reasoning: dependency publishing is a post-merge action, but this CI runs at PR time, and PRs may include inspector changes (same repo). Building from source (~30s overhead for 1 dependency) fully eliminates the timing mismatch rather than working around it.

Longer term, splitting dependency bump and constraint bump into separate PRs is an alternative, but it's hard to enforce consistently and still requires correct release ordering between merges.

jongio (approved):
Comment wording updated per my earlier feedback. The framing now correctly presents building from source as the intended CI approach rather than a temporary workaround. No other changes.

trangevi (approved):

JeffreyCA (approved):

vhvb1989 (approved):

PR #8917, opened 2026-07-01
fix: set brownfield Foundry projectName without ACR

### LINKED ISSUE #8920: [ext-agents]: Brownfield Foundry provision fails with InvalidTemplate (empty project name segment) when no ACR is created
**Describe the bug**
Deploying a brownfield (`endpoint:`) Foundry project service that declares model `deployments` but does **not** create a container registry (for example, an agent that uses `codeConfiguration`, i.e. a ZIP/code deploy) fails ARM template validation:

```
ERROR: deployment failed: error deploying infrastructure: provisioning deploy failed:
arm_deployment_create: PUT .../deployments/azd-foundry-...-brownfield
RESPONSE 400: 400 Bad Request
ERROR CODE: InvalidTemplate
{
  "error": {
    "code": "InvalidTemplate",
    "message": "Deployment template validation failed: 'The template resource
      '<account>/' of type 'Microsoft.CognitiveServices/accounts/projects' ...
      has an invalid name. The name '<account>/' contains '1' name segment(s)
      separated by '/', but '2' segment(s) were expected."
  }
}
```

The generated brownfield ARM deployment references the existing `Microsoft.CognitiveServices/accounts/projects` resource with a name of the form `<accountName>/<projectName>`, but the project segment is empty, so the name collapses to `<accountName>/` (a single segment) and ARM rejects the template.

**To Reproduce**
1. Have a Foundry account with an existing project (brownfield). In `azure.yaml`, declare a code-less `azure.ai.project` service that references it via `endpoint:` and declares one or more model `deployments`, plus an agent that deploys from source via `codeConfiguration` (so no ACR is created):
   ```yaml
   name: my-agent-app
   services:
     my-project:
       host: azure.ai.project
       endpoint: https://<account>.services.ai.azure.com/api/projects/<project>
       deployments:
         - model:
             format: OpenAI
             name: gpt-4o-mini
             version: "2024-07-18"
           name: gpt-4o-mini
           sku:
             capacity: 50
             name: DataZoneStandard
     my-agent:
       host: azure.ai.agent
       language: python
       project: src/my-agent
       uses:
         - my-project
       codeConfiguration:
         dependencyResolution: remote_build
         entryPoint: main.py
         runtime: python_3_13
   infra:
     provider: microsoft.foundry
   ```
2. Run `azd provision` (or `azd up`).
3. The deployment fails with the `InvalidTemplate` error above (`... accounts/projects ... has an invalid name ... contains '1' name segment(s) ... but '2' segment(s) were expected`).

Note: the failure only occurs on the no-ACR path. A configuration that creates an ACR (a hosted build-from-source agent) does not hit this, because the project name is populated on that path.

**Expected behavior**
Provisioning a brownfield Foundry project with model deployments (and no ACR) should succeed. The existing `accounts/projects` resource should always be referenced with a valid two-segment name `<accountName>/<projectName>`, with the project name resolved from the service `endpoint`.

**Additional context**
Relevant code: `cli/azd/extensions/azure.ai.agents/internal/project/foundry_provisioning_provider.go` (`brownfieldParams`) and the embedded `brownfield.arm.json` / `brownfield.bicep` template, which references `foundryAccountPreview::project` unconditionally.

huimiu:
Fixes #8920 

## TL;DR

Running `azd provision` / `azd deploy` on an **existing** Foundry project that adds model deployments **without** creating a container registry failed with an ARM `InvalidTemplate` error. The required `projectName` value was only passed to the template when a registry was *also* being created. This PR always passes it.

## Who hits this

Any brownfield Foundry project service (`host: azure.ai.project` with an `endpoint:`) that:

- declares model `deployments:`, **and**
- deploys its agent as code/ZIP (`codeConfiguration`) — so **no** container registry (ACR) is created.

## Symptom

```
InvalidTemplate: The template resource 'hui-proj-ncus-resource/' of type
'Microsoft.CognitiveServices/accounts/projects' ... contains '1' name
segment(s) separated by '/', but '2' segment(s) were expected.
```

The resource name is missing its second segment: it should be `<account>/<project>` but came out as just `<account>/`.

## Root cause

`brownfield.arm.json` **always** references the existing project resource, whose ARM name is `format('{0}/{1}', accountName, projectName)`. But `brownfieldParams` only set the `projectName` parameter **inside the `if createACR` branch**. With no ACR, `projectName` fell back to the template default `''`, producing the invalid single-segment name `<accountName>/`.

## Fix

Move `projectName` out of the `createACR` branch so it is **always** set (resolved from the project endpoint). ACR-only params (`includeAcr`, `acrName`, `tags`, `location`) stay gated on `createACR`.

```go
params := map[string]any{
    "accountName": {"value": account},
    "deployments": {"value": p.brownfieldDeployments},
    "projectName": {"value": p.brownfieldProjectName()}, // always set now
}
if createACR {
    // includeAcr / acrName / tags / location only here
}
```

## How to verify

`azd provision --preview` on such a project runs an ARM what-if through the same code path (and creates nothing):

- **Before:** fails with the `InvalidTemplate` error above.
- **After:** the what-if succeeds and the plan shows the model deployment on the existing `<account>/<project>` resource.

## Tests

`TestBrownfieldParams` now asserts `projectName` is present in the no-ACR case.

## Related

A separate azure.yaml serialization issue for code-less hosts (invalid `project: ""` / `language: ""`) surfaced during this work and is tracked in #8919.

### DIFF
--- a/.impeccable/hook.cache.json
+++ b/.impeccable/hook.cache.json
@@ -1,0 +1,1 @@
+{"version":1,"sessions":{}}
--- a/.impeccable/hook.cache.json
+++ b/.impeccable/hook.cache.json
@@ -1,1 +1,0 @@
-{"version":1,"sessions":{}}
--- a/cli/azd/extensions/azure.ai.agents/CHANGELOG.md
+++ b/cli/azd/extensions/azure.ai.agents/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Bugs Fixed
 
+- [[#8917]](https://github.com/Azure/azure-dev/pull/8917) Fix an `InvalidTemplate` error when deploying a brownfield (`endpoint:`) Foundry project that declares model deployments without creating a container registry. The `projectName` ARM parameter is now always set, so the referenced `Microsoft.CognitiveServices/accounts/projects` resource gets a valid two-segment name.
 - [[#8901]](https://github.com/Azure/azure-dev/pull/8901) Remove duplicate service-target provider claims from the `azure.ai.agents` extension manifest for hosts now owned by the split Foundry extensions (`azure.ai.projects`, `azure.ai.connections`, `azure.ai.toolboxes`). Thanks @huimiu for the contribution!
 
 ## 1.0.0-beta.1 (2026-06-30)
--- a/cli/azd/extensions/azure.ai.agents/internal/project/foundry_provisioning_provider.go
+++ b/cli/azd/extensions/azure.ai.agents/internal/project/foundry_provisioning_provider.go
@@ -810,11 +810,15 @@ func (p *FoundryProvisioningProvider) brownfieldParams(
 	params := map[string]any{
 		"accountName": map[string]any{"value": account},
 		"deployments": map[string]any{"value": p.brownfieldDeployments},
+		// projectName is always required: the template unconditionally references
+		// the existing accounts/projects resource, whose ARM name is
+		// '<accountName>/<projectName>'. Leaving it empty yields a single-segment
+		// name and an InvalidTemplate error even when no ACR is created.
+		"projectName": map[string]any{"value": p.brownfieldProjectName()},
 	}
 	if createACR {
 		params["includeAcr"] = map[string]any{"value": true}
 		params["acrName"] = map[string]any{"value": p.brownfieldACRName(account)}
-		params["projectName"] = map[string]any{"value": p.brownfieldProjectName()}
 		params["tags"] = map[string]any{"value": map[string]string{"azd-env-name": p.envName}}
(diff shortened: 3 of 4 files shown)


jongio (approved):
Clean, well-scoped fix. The root cause (projectName only set inside the createACR branch while the ARM template unconditionally references the accounts/projects resource) is correctly identified and addressed.

The fix moves projectName into the unconditional params block, which is the right layer. The test now covers the no-ACR case with a real endpoint and asserts the two-segment name resolves. No regressions expected since brownfieldProjectName() was already exercised on the ACR path.

huimiu:
close as duplicated with https://github.com/Azure/azure-dev/pull/8947

PR #8918, opened 2026-07-01
fix: omit empty project and language keys in azure.yaml

huimiu:
## Problem

`azd ai agent init` generates an `azure.yaml` where the code-less resource service (`host: azure.ai.project`) is written with `project: ""` and `language: ""`:

```yaml
services:
  hui-proj-ncus:
    project: ""
    host: azure.ai.project
    language: ""
    ...
```

The `azure.yaml` schema disables `project` for these code-less hosts (`"project": false` for `azure.ai.project`, `azure.ai.connection`, `azure.ai.toolbox`, `azure.ai.skill`, `azure.ai.routine`), so the generated file fails validation with **"Property project is not allowed."**

## Root cause

`ServiceConfig.RelativePath` (`yaml:"project"`) and `ServiceConfig.Language` (`yaml:"language"`) had no `omitempty`, so `project.Save` always emitted them — even for code-less resource hosts that legitimately have no source path or language.

## Fix

Add `omitempty` to `RelativePath` and `Language`. Traditional/container hosts always set these, so their output is unchanged; only empty values (code-less hosts) are now omitted.

## Tests

Added `TestCodelessServiceOmitsProjectAndLanguage` verifying an `azure.ai.project` service serializes without `project:`/`language:` keys. Full `pkg/project` and `internal/grpcserver` suites pass.

### DIFF
--- a/cli/azd/CHANGELOG.md
+++ b/cli/azd/CHANGELOG.md
@@ -8,7 +8,7 @@
 
 ### Bugs Fixed
 
-- Omit empty `project` and `language` keys when saving `azure.yaml`, so code-less resource hosts such as `azure.ai.project` no longer emit an invalid `project: ""` that fails schema validation.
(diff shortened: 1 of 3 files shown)


huimiu:
Superseded: moved this change into #8917 (branch hui/fix-foundry-project-name).

PR #8926, opened 2026-07-01
fix: apply --deploy-mode flags to adopted azure.yaml (#8923)

### LINKED ISSUE #8923: azd ai agent init -m <azure.yaml> ignores --deploy-mode code and does not update the adopted azure.yaml
## Summary

When adopting a unified Foundry `azure.yaml` via `azd ai agent init -m <azure.yaml>`, the `--deploy-mode code` flag (and the related `--runtime` / `--entry-point` flags) is ignored. The adopted `azure.yaml` is written to the project root unchanged, so the agent service is never updated with a `code_configuration`. When `-m` points at an agent manifest instead, `--deploy-mode code` correctly writes the code configuration into `azure.yaml`.

This shares the same root cause as #8922: the new adoption path added in PR #8885 does not process any of the init flags. Supporting `azure.yaml` with `-m` needs to also handle these flags and update the `azure.yaml` accordingly.

## Repro

1. Have a unified Foundry `azure.yaml` whose agent service is container-style.
2. Run `azd ai agent init -m <azure.yaml> --deploy-mode code` (optionally with `--runtime` / `--entry-point`).

Expected: the adopted `azure.yaml` agent service is updated with a `code_configuration` (runtime, entry point, etc.), the same way the agent-manifest path does.

Actual: the `azure.yaml` is adopted as-is. `--deploy-mode code` has no effect; no `code_configuration` is written and the flags are silently ignored.

## Root cause

The adoption path `runInitFromAzureYaml` (added in PR #8885) receives `*initFlags` but only reads `flags.src`, `flags.manifestPointer`, and `flags.env`. It never reads `flags.deployMode`, `flags.runtime`, or `flags.entryPoint`, and `init_adopt.go` contains no code-configuration handling at all:

- `cli/azd/extensions/azure.ai.agents/internal/cmd/init_adopt.go` (`runInitFromAzureYaml`, lines 163-216)

By contrast, the agent-manifest path resolves deploy mode with `promptDeployMode`, and when code deploy is selected builds a `code_configuration` via `promptCodeConfig`, sets it on the agent template, and persists it into `azure.yaml` via `addToProject`:

- `cli/azd/extensions/azure.ai.agents/internal/cmd/init.go` (deploy-mode / code-config handling around lines 1604-1643; `addToProject` at ~1709)

Since the adoption path short-circuits before any of that runs, the deploy-mode flags are dropped and the adopted `azure.yaml` is never rewritten.

## Suggested fix

The unified-`azure.yaml` adoption path needs to honor the init flags -- at minimum `--deploy-mode` (`code`/`container`), and for code deploy `--runtime` / `--entry-point` -- and update the adopted `azure.yaml` agent service's `code_configuration` accordingly (prompting interactively when the flags are omitted, matching the agent-manifest flow). This is a missing feature of the `azure.yaml` `-m` support.

## Environment

- Extension: azure.ai.agents 1.0.0-beta.1+ (introduced by PR #8885)
- Related: #8922, #8885

trangevi:
## Why

When `azd ai agent init -m <azure.yaml>` routes to the unified Foundry adoption path, the `--deploy-mode`, `--runtime`, and `--entry-point` flags are silently ignored. The adopted azure.yaml is written as-is with no `code_configuration` applied to the agent service. This means users cannot select code deploy when adopting a sample.

## Approach

After `scaffoldProject` succeeds, the new `applyDeployModeToAdoptedProject` function locates the `azure.ai.agent` service in the adopted project and applies deploy-mode configuration using the same `promptDeployMode` / `promptCodeConfig` functions the manifest path already uses.

**Decision logic:**
- Explicit `--deploy-mode` flag -- always apply (user intent overrides sample)
- No flag + service already has `code_configuration` or `docker` property -- no-op (respect pre-configured sample)
- No flag + neither configured -- prompt interactively (or auto-select per noPrompt/userProvidedManifest rules)

**For code deploy:** writes `code_configuration` (runtime, entry_point, dependency_resolution) onto the service, updates language to `python`/`csharp`, clears any docker property.

**For container deploy:** sets the `docker` property (remoteBuild: true), updates language to `docker`, clears any code_configuration.

## Testing

- Added `TestAdoptedServiceHasCodeConfig` covering nil props, empty props, valid struct value, and null value cases
- All existing adoption tests continue to pass
- Full `go test ./internal/cmd/... -short` passes

Fixes: #8923


cli/azd/extensions/azure.ai.agents/internal/cmd/init_adopt.go
```diff
@@ -522,3 +531,194 @@ func printAdoptionNextSteps(ctx context.Context, azdClient *azdext.AzdClient, fo
...
+	// already configured, respect the sample's existing configuration.
+	if flags.deployMode == "" && (hasCodeConfig || hasDocker) {
+		return nil
+	}
+
+	// Resolve deploy mode via the same logic the manifest path uses.
+	// The target directory (project root after adoption) is used for
+	// language detection.
+	targetDir := "."
+	showCodeDeploy := isPythonProject(targetDir) || isDotnetProject(targetDir)
```
jongio, line 1261: `targetDir := "."` scans the project root, but the sample's agent code lives in the service subdirectory (e.g., `./agents/assistant`). The manifest path at `init.go:1388` passes the service's source directory to `isPythonProject`/`isDotnetProject`. Use `agentSvc.GetRelativePath()` here so language detection finds the correct files. Without this, `showCodeDeploy` will be `false` for most adopted samples, and `promptCodeConfig` will auto-detect the wrong runtime/entry-point.

cli/azd/extensions/azure.ai.agents/internal/cmd/init_adopt.go
```diff
@@ -522,3 +531,194 @@ func printAdoptionNextSteps(ctx context.Context, azdClient *azdext.AzdClient, fo
...
+	// Find the agent service in the adopted project.
+	var agentServiceName string
+	var agentSvc *azdext.ServiceConfig
+	for name, svc := range resp.GetProject().GetServices() {
+		if svc.GetHost() == AiAgentHost {
+			agentServiceName = name
+			agentSvc = svc
+			break
+		}
+	}
```
jongio, line 1204: Map iteration order is non-deterministic in Go. If the adopted sample declares multiple `azure.ai.agent` services (e.g., researcher, writer, triage), this picks an arbitrary one and leaves the others unconfigured. Consider iterating all services and applying the deploy mode to each, or detecting the multi-service case and prompting the user.

cli/azd/extensions/azure.ai.agents/internal/cmd/init_adopt.go
```diff
@@ -522,3 +531,194 @@ func printAdoptionNextSteps(ctx context.Context, azdClient *azdext.AzdClient, fo
...
+	langValue, _ := structpb.NewValue(language)
+	if _, err := azdClient.Project().SetServiceConfigValue(ctx, &azdext.SetServiceConfigValueRequest{
+		ServiceName: serviceName,
+		Path:        "language",
+		Value:       langValue,
+	}); err != nil {
+		return fmt.Errorf("updating service language to %s: %w", language, err)
+	}
+
```
jongio, line 1373:
```suggestion
	langValue, err := structpb.NewValue(language)
	if err != nil {
		return fmt.Errorf("encoding language value: %%w", err)
	}
```

jongio (changes requested):
Two verified issues that affect correctness, plus a robustness concern:

**1. Language detection uses project root instead of service directory (HIGH)**

\	argetDir := "."\ means \isPythonProject\/\isDotnetProject\ scan the project root, but adopted samples declare agent services in subdirectories (e.g., \project: ./agents/assistant\). The manifest path uses the downloaded service directory (\init.go:1388\). Here, \gentSvc.GetRelativePath()\ should be used so auto-detection finds the correct files.

**2. Non-deterministic map iteration picks an arbitrary service (MEDIUM)**

esp.GetProject().GetServices()\ is a map. With multi-agent samples (researcher, writer, triage), the \reak\ on line 559 applies configuration to whichever service the runtime iterates first. Either apply to all matching services or collect them and let the user choose.

**3. Silently discarded \structpb.NewValue\ errors (LOW)**

Lines 665 and 697 discard the error from \structpb.NewValue(language)\ with \_\. While unlikely to fail for string literals, ignoring errors deviates from the extension's error handling conventions.

Note on the existing bot comments about snake_case vs camelCase: those comments cite \internal/project/agent_definition.go\ and \AgentDefinitionInline\, neither of which exist in the codebase. The agent_yaml struct uses YAML tag \code_configuration\ and the service target detection (\service_target_agent.go:1122\) also reads \code_configuration\. The PR's use of snake_case appears internally consistent. I'd recommend verifying which key the deploy path actually reads from the azure.yaml inline properties to confirm end-to-end correctness.

jongio:
**Corrected review summary** (the review body above has formatting issues from a shell escape bug):

Two verified issues that affect correctness, plus a robustness concern:

**1. Language detection uses project root instead of service directory (HIGH)**

`targetDir := "."` means `isPythonProject`/`isDotnetProject` scan the project root, but adopted samples declare agent services in subdirectories (e.g., `project: ./agents/assistant`). The manifest path uses the downloaded service directory (`init.go:1388`). Here, `agentSvc.GetRelativePath()` should be used so auto-detection finds the correct files.

**2. Non-deterministic map iteration picks an arbitrary service (MEDIUM)**

`resp.GetProject().GetServices()` is a map. With multi-agent samples (researcher, writer, triage), the `break` on line 559 applies configuration to whichever service the runtime iterates first. Either apply to all matching services or collect them and let the user choose.

**3. Silently discarded `structpb.NewValue` errors (LOW)**

Lines 665 and 697 discard the error from `structpb.NewValue(language)` with `_`. While unlikely to fail for string literals, ignoring errors deviates from the extension's error handling conventions.

Note on the existing bot comments about snake_case vs camelCase: those comments cite `internal/project/agent_definition.go` and `AgentDefinitionInline`, neither of which exist in the codebase. The agent_yaml struct uses YAML tag `code_configuration` and the service target detection (`service_target_agent.go:1122`) also reads `code_configuration`. The PR's use of snake_case appears internally consistent. I'd recommend verifying which key the deploy path actually reads from the azure.yaml inline properties to confirm end-to-end correctness.

cli/azd/extensions/azure.ai.agents/internal/cmd/init_adopt.go
```diff
@@ -522,3 +531,215 @@ func printAdoptionNextSteps(ctx context.Context, azdClient *azdext.AzdClient, fo
...
+		}
+	}
+	if agentSvc == nil {
+		// No agent service found — nothing to configure.
+		return nil
+	}
+
+	// Apply --image override to the agent service when provided.
+	if flags.image != "" {
+		imageValue, _ := structpb.NewValue(flags.image)
```
jongio: Same pattern as line 697: discarding the error from `structpb.NewValue`. While unlikely to fail for a string that passed `validateImageFlag`, the extension's convention is to propagate errors.

jongio (changes requested):
Incremental review of commit 511d165 (--image flag handling in adoption path).

The new alidateImageFlag + image-override logic is well-structured and the test coverage for validation is solid. Two notes on this new commit:

**Prior findings still open:**

1. 	argetDir := "." scans the project root instead of the service subdirectory for language detection (HIGH, line 605). This still causes isPythonProject/isDotnetProject to miss files in samples like ./agents/assistant.
2. Non-deterministic map iteration picks an arbitrary service when multiple zure.ai.agent services exist (MEDIUM, line 559).

**New issue in this commit:**

3. imageValue, _ := structpb.NewValue(flags.image) (line 572) discards the error. While alidateImageFlag already checks the format, structpb.NewValue can still fail for other reasons. This follows the same pattern I flagged in my previous review at line 697.

The --image feature itself looks correct: early validation prevents conflicts with --deploy-mode code, the image property is written before applying container config, and the tests cover the key scenarios. Please address findings 1 and 2 from the prior review before merging.

jongio:
**Corrected review summary** (the review body above has formatting issues from a shell escape bug):

Incremental review of commit 511d165 (--image flag handling in adoption path).

The new `validateImageFlag` + image-override logic is well-structured and the test coverage for validation is solid. Two notes on this new commit:

**Prior findings still open:**

1. `targetDir := "."` scans the project root instead of the service subdirectory for language detection (HIGH, line 605). This still causes `isPythonProject`/`isDotnetProject` to miss files in samples like `./agents/assistant`.
2. Non-deterministic map iteration picks an arbitrary service when multiple `azure.ai.agent` services exist (MEDIUM, line 559).

**New issue in this commit:**

3. `imageValue, _ := structpb.NewValue(flags.image)` (line 572) discards the error. While `validateImageFlag` already checks the format, `structpb.NewValue` can still fail for other reasons. This follows the same pattern I flagged in my previous review at line 697.

The --image feature itself looks correct: early validation prevents conflicts with `--deploy-mode code`, the image property is written before applying container config, and the tests cover the key scenarios. Please address findings 1 and 2 from the prior review before merging.

jongio (approved):
All three findings from my prior reviews are addressed:

1. Language detection now uses the service's relative path instead of project root.
2. Multi-service case handled correctly by collecting all agent services and iterating the full set.
3. structpb.NewValue errors are no longer silently discarded.

The camelCase key rename (codeConfiguration, entryPoint, dependencyResolution) is consistent with how the deploy path reads these via JSON unmarshal.

cli/azd/extensions/azure.ai.agents/internal/cmd/init_adopt.go
```diff
@@ -522,3 +531,244 @@ func printAdoptionNextSteps(ctx context.Context, azdClient *azdext.AzdClient, fo
...
+		}
+		log.Printf("Applied --image %q to agent service %q", flags.image, serviceName)
+
+		// --image implies container deploy; apply container config and return.
+		return applyContainerDeployToService(ctx, azdClient, serviceName)
+	}
+
+	// Check whether the service already specifies its deploy mode.
+	hasCodeConfig := adoptedServiceHasCodeConfig(svc)
+	hasDocker := svc.GetDocker() != nil
```
huimiu: I think there might be a subtle issue here: hasDocker looks like it's always true, which means when the user doesn't pass --deploy-mode, we always hit the early return below and never actually apply the deploy mode.

I dug into it a bit — ServiceConfig.Docker is a value type (not a pointer), and when Project().Get() maps it into the response, it always produces a non-nil Docker object, even when there's no docker: section in azure.yaml. I confirmed it locally too: with no docker block, GetDocker() still comes back non-nil (just empty values), so svc.GetDocker() != nil never ends up false.

The effect is that adopting a sample without a deploy mode set would quietly leave it unchanged instead of defaulting to code deploy. Totally understand why it slipped through — the unit tests build ServiceConfig directly, so they don't go through the real gRPC round-trip where this shows up.

jongio: `hasDocker` is always true after the gRPC round-trip. The internal `ServiceConfig.Docker` is a value type (not a pointer), and `mapper_registry.go:165` always converts it to a non-nil `*DockerProjectOptions`. This means when `flags.deployMode == ""` the early return on line 612 always triggers, silently skipping deploy-mode configuration for adopted samples.

Consider checking `additionalProperties` for a `docker` key (matching how `adoptedServiceHasCodeConfig` works), or validate that the Docker struct has non-zero fields.

jongio (changes requested):
@huimiu's observation is correct. I traced the full conversion path:

1. Internal `project.ServiceConfig` declares `Docker DockerProjectOptions` as a value type (`service_config.go:48`), so it's always the zero-value struct when no `docker:` section exists in azure.yaml.

2. The mapper at `mapper_registry.go:165` registers a converter `DockerProjectOptions -> *azdext.DockerProjectOptions` that unconditionally returns a non-nil pointer (line 193).

3. The ServiceConfig converter at `mapper_registry.go:122` calls `mapper.Convert(src.Docker, &docker)` with that zero-value struct, getting back a non-nil `*DockerProjectOptions` with all empty fields.

4. Result: `svc.GetDocker() != nil` is always true after the gRPC round-trip, so the early return on line 612 always fires when `flags.deployMode == ""`.

**Fix**: mirror the `adoptedServiceHasCodeConfig` approach by checking `additionalProperties` for a `docker` key, or check whether the Docker struct has any non-zero fields (e.g. `svc.GetDocker().GetPath() != "" || svc.GetDocker().GetRemoteBuild()`). The `additionalProperties` approach is more consistent with how the deploy path reads these values.

Withdrawing my earlier approval since this means the no-flag adoption path silently skips deploy-mode configuration.

jongio (approved):
All findings from my prior reviews are addressed in this commit. The `adoptedServiceHasDocker` function correctly checks `additionalProperties` instead of `svc.GetDocker()` (which always returns non-nil due to the gRPC mapper zero-value behavior). The implementation mirrors `adoptedServiceHasCodeConfig` and the test coverage is solid.

jongio (approved):
Using UnsetServiceConfig instead of writing null is the correct approach. It properly removes the key rather than leaving an explicit null in the YAML, and the guard conditions avoid unnecessary API calls when the property isn't set.

therealjohn (approved):

therealjohn (approved):

cli/azd/extensions/azure.ai.agents/internal/cmd/init_adopt.go
```diff
@@ -1162,3 +1170,269 @@ func printAdoptionNextSteps(ctx context.Context, azdClient *azdext.AzdClient, fo
...
+// service and ensures the language is "docker". Removes any codeConfiguration
+// if present.
+func applyContainerDeployToService(
+	ctx context.Context,
+	azdClient *azdext.AzdClient,
+	serviceName string,
+	svc *azdext.ServiceConfig,
+) error {
+	// Set docker property with remote build enabled.
+	dockerMap := map[string]any{"remoteBuild": true}
```
RickWinter, line 1399: applyContainerDeployToService always writes remoteBuild: true, but the code-based init path in init_from_code.go sets serviceConfig.Docker = &azdext.DockerProjectOptions{RemoteBuild: !networkInjected}, deliberately turning remote build off when the target Foundry project has VNET network injection, since the ACR Tasks worker can't reach a registry inside the VNET. The deploy path doesn't re-derive this; it surfaces guidance telling the user to flip docker.remoteBuild to false by hand. So a sample adopted with container deploy against a network-injected project will fail the first azd deploy and force a manual edit, while the same sample created via init_from_code would have worked. Can we mirror the networkInjected check here (thread the selected Foundry project through) so the two init paths agree?

cli/azd/extensions/azure.ai.agents/internal/cmd/init_adopt.go
```diff
@@ -1162,3 +1170,269 @@ func printAdoptionNextSteps(ctx context.Context, azdClient *azdext.AzdClient, fo
...
+		return fmt.Errorf("reading adopted project: %w", err)
+	}
+
+	// Collect all agent services in the adopted project.
+	type agentEntry struct {
+		name string
+		svc  *azdext.ServiceConfig
+	}
+	var agentServices []agentEntry
+	for name, svc := range resp.GetProject().GetServices() {
```
RickWinter, line 1200: nit: ranging over resp.GetProject().GetServices() (a map) gives nondeterministic order, so with more than one azure.ai.agent service the prompt order and log output vary run to run. Single-agent samples won't notice, but sorting agentServices by name before the apply loop would make the behavior deterministic.

RickWinter (commented):
This applies --deploy-mode / --runtime / --entry-point to the adopted azure.yaml agent service, closing the gap where those flags were silently dropped on the Foundry adoption path. The approach is the right shape: it reuses promptDeployMode / promptCodeConfig, writes codeConfiguration with the same camelCase keys the schema and AgentDefinitionToServiceProperties already use (runtime, entryPoint, dependencyResolution all check out), and the has-code-config / has-docker detection correctly reads additionalProperties rather than trusting the always-non-nil Docker pointer. The added tests cover the detection helpers and image-flag validation well.

One thing worth resolving before merge: applyContainerDeployToService hardcodes remoteBuild: true, while the code-based init path turns it off for VNET network-injected Foundry projects, so adopting a container sample against such a project will fail the first deploy and require a manual edit. The map-iteration ordering is a minor nondeterminism nit for the multi-agent case. Nothing else blocks; the key handling and the respect-existing-sample logic are correct as written.

PR #8927, opened 2026-07-01
Make extension registry updates resilient to coordinated multi-extension version bumps

### LINKED ISSUE #8924: Extension releases: decouple Fig/Usage snapshot generation from registry dependency resolution, and add cross-dependency registry validation
## Background

During #8896 we bumped all `azure.ai.*` extension versions at the same time. When the individual per-extension release pipelines were kicked off in parallel, every one of them failed at the **Refresh Fig/Usage snapshots** step (`eng/pipelines/templates/stages/publish-extension.yml`, which runs `go build . && go test ./cmd -run 'TestFigSpec|TestUsage'` with `UPDATE_SNAPSHOTS=true`). We ended up reconciling the registry manually (a local branch with `registry.json` updated for all new binaries, then a local `azd x publish`) because dependency resolution during install kept failing. See commit 83053f0.

There are two related gaps behind this.

## Gap 1: snapshot generation is coupled to whole-registry dependency consistency

`TestFigSpec`/`TestUsage` add the in-repo `registry.json` as a source and then `installAllExtensions` (`cli/azd/cmd/extension_helpers_test.go`) installs **every** extension through the full, dependency-resolving `azd extension install`, failing on the first error. But each per-extension pipeline checks out `main` and `azd x publish` rewrites **only its own** entry in `registry.json`, so the registry under test is effectively "this extension's NEW version + every sibling at whatever is currently on `main`". During a coordinated multi-extension bump that graph is transiently inconsistent, which produced both observed failure modes:

- `dependency azure.ai.inspector required by azure.ai.agents was not found` (the child was bumped to require a newer dependency version that was not yet satisfiable in the registry).
- `installed dependency azure.ai.connections version 1.0.0-beta.1 does not satisfy constraint "~0.1.0-preview"` (the `microsoft.foundry` meta-package, installed last, still pinned the old range against an already-installed newer child).

The snapshot only needs each extension's own **binary** present so its commands register; it does not need a consistent dependency graph. Worth noting: `microsoft.foundry` is a pure meta-package (0 artifacts), so it contributes nothing to the snapshot, yet its dependency resolution is exactly what fails.

### Possible Approaches

Add a `--no-dependencies` flag to `azd extension install` that skips the dependency install loop and the installed-dependency constraint check in `installInternal`, and have the snapshot helper use it (and/or skip artifact-less meta-packages). This decouples snapshot generation from registry-wide consistency and is a generally useful CLI capability (partial/offline/deterministic installs), not a test-only hack. Alternative: generate the command snapshots directly from local build artifacts without going through a registry-based install at all.

## Gap 2: no cross-extension dependency validation on the registry

`ValidateRegistry` (`cli/azd/pkg/extensions/validate_registry.go`) validates per-extension fields but never checks that declared dependencies (and meta-package pins) actually resolve to a published version. As a result, an inconsistent registry only surfaces as a cryptic install failure inside a snapshot test rather than as an early, readable error.

### Possible Approaches

Extend registry validation to verify that every declared dependency constraint resolves to at least one published version in the (merged) registry. Keep it advisory during per-pipeline `azd x publish` (each pipeline is legitimately transiently inconsistent), but make it an error gate on the merged registry in PR/main CI so bad states are caught early with a human-readable message.

## Related

- #8896 (the coordinated version bump that triggered this)
- commit 83053f0 (the manual registry reconciliation)
- Companion issue for the release-process coordination angle: #8925

JeffreyCA:
Fixes #8924

This PR makes extension registry updates resilient when several interdependent extensions are released at the same time.

## The problem

When we recently bumped all the `azure.ai.*` extension versions together (#8896) and kicked off the per-extension release pipelines in parallel, every pipeline failed at the "Refresh Fig/Usage snapshots" step.

That step installs every extension from the in-repo `registry.json` through the full, dependency-resolving install path, and fails on the first error. But each pipeline rewrites only its own registry entry against the current tip of `main`, so the registry under test was transiently inconsistent, producing failures such as:

- `dependency azure.ai.inspector required by azure.ai.agents was not found`
- `installed dependency azure.ai.connections version 1.0.0-beta.1 does not satisfy constraint "~0.1.0-preview"`

We ended up reconciling the registry by hand. This PR closes the two gaps behind that outcome.

## Change 1: decouple snapshot generation from dependency resolution

**What this means:** the ADO extension release pipeline should no longer fail on "Update_Registry - Refresh Fig/Usage snapshots", when the registry is momentarily inconsistent during multi-extension version bumps.

Command snapshots only need each extension's own binary so its commands register. They do not need the registry's dependency graph to be internally consistent.

- Adds a `--no-dependencies` flag to `azd extension install` that installs only the named extension, skipping dependency resolution and the installed-dependency constraint check.
- The Fig/Usage snapshot helper now uses it, so each extension installs independently of its siblings' versions.
- The flag is honored on both the fresh-install and reinstall/upgrade paths.
- It is opt-in: a normal `azd extension install <id>` still installs the full dependency graph as before.
- Nothing is dropped from the snapshot: every extension that ships a binary is also a top-level registry entry installed on its own, and the artifact-less `microsoft.foundry` meta-package contributes no commands.

**Test pipeline run releasing azure.ai.agents without dependency azure.ai.inspector:**

image

Resulting PR: https://github.com/Azure/azure-dev/pull/8934

## Change 2: cross-dependency registry validation

**What this means:** a genuinely broken registry (a dependency pinned to a version that was never published) is blocked at PR time (not during release pipeline) with a clear, actionable message, instead of slipping through and breaking installs for users. Warnings and mid-rollout publishing are not blocked, so routine releases are unaffected.

Registry validation previously checked per-extension fields but never verified that a declared dependency could actually be resolved, so an inconsistent registry only surfaced as a cryptic install failure inside a snapshot test.

- The validator now checks every declared dependency against the versions the registry actually publishes, reusing the installer's version-constraint matcher so validation mirrors install-time resolution.
- An unsatisfiable constraint is an error (it can never be resolved); an absent dependency id is a warning (it may come from another source).
- A new `TestRegistryFileIsValid` test runs in the `ext-registry-ci` workflow on any PR that changes `registry.json`, so the automated registry-update PRs are gated on it.
- It fails only on errors, never warnings, and names the exact extension, constraint, and available versions.

**Updating microsoft.foundry before all dependencies are updated:**

image

**Passing registry consistency but failing snapshot tests:**

image

**Successful run:**

image

## Behavior at a glance

| Scenario | Before | After |
| --- | --- | --- |
| "Refresh Fig/Usage snapshots" step during a coordinated multi-extension bump | Fails on the first extension whose dependency graph is transiently inconsistent, blocking the registry update | Each extension installs independently via `--no-dependencies`; the step no longer depends on registry-wide consistency |
| `azd extension install <id>` (no flag), end user | Installs the extension and its declared dependencies | Unchanged |
| `azd extension install <id> --no-dependencies` | Flag did not exist | Installs only the named extension; dependencies are not resolved, installed, or constraint-checked |
| Registry with a dependency constraint no published version can satisfy | Cryptic install failure inside a snapshot test | Validation error on any PR that changes `registry.json`, naming the extension, constraint, and available versions |
| Registry with a dependency id absent from the registry | Not detected | Validation warning (may be provided by another source) |
| `azd x publish` while a single extension is mid-rollout | Publishes the entry | Unchanged: publish does not run registry validation, so transient states are not blocked |

## Testing

- Added unit coverage for skipping dependencies on both fresh install and upgrade, and for bypassing the installed-dependency constraint check.
- Added validator coverage for the satisfied, unsatisfiable-as-error, absent-as-warning, and empty-constraint cases, plus the production-registry gate.
- Regenerated the Fig and Usage command snapshots to include the new flag.
- The existing extension suite, `gofmt`, and `golangci-lint` all pass.


jongio (approved):
Two well-scoped changes that address a real CI reliability problem. The --no-dependencies flag is cleanly threaded through both install and upgrade paths with good test coverage. The registry dependency validation catches genuinely broken states at PR time without blocking routine releases (warning vs error distinction is correct). A few observations:

1. The installInternal check at line 590 (if !opts.SkipDependencies && ...) correctly gates the entire dependency loop, but the earlier check at line 561 (if len(selectedVersion.Artifacts) == 0 && len(selectedVersion.Dependencies) == 0) still runs. This means a dependency-only extension pack (no artifacts, only dependencies) installed with --no-dependencies would still pass this guard and proceed to the artifact download path where it would no-op. That's fine for the stated use case (every extension with a binary is a top-level entry), just noting the interaction.

2. The workflow condition change (if: failure() && (steps.figspec.outcome == 'failure' || steps.usage.outcome == 'failure')) correctly scopes the snapshot guidance message to only fire when the snapshot tests themselves fail, not when the new validation step fails. Good attention to detail.

3. Test coverage is thorough: fresh install, upgrade, constraint-bypass, and the validator's satisfied/unsatisfiable/absent/empty cases are all covered.

vhvb1989 (approved):

PR #8933, opened 2026-07-01
fix: run interactive Azure context setup in unified azure.yaml adoption path

### LINKED ISSUE #8922: azd ai agent init -m <azure.yaml> skips interactive environment setup (subscription, Foundry project, model)
## Summary

Since PR #8885 (feat: adopt unified azure.yaml in azd ai agent init -m), running `azd ai agent init -m <azure.yaml>` (pointing `-m` at a unified Foundry `azure.yaml`) no longer runs the interactive environment setup steps -- selecting a subscription, selecting/creating a Foundry project, model configuration, and deploy-mode selection. The command stages/adopts the sample's `azure.yaml`, scaffolds the project, prints next steps, and returns without any Azure context prompting.

Pointing `-m` at an agent manifest (top-level `template:`) still runs the full interactive flow. Only the new unified `azure.yaml` adoption path is affected.

## Repro

1. Have a unified Foundry `azure.yaml` (a manifest whose `services:` declare a Foundry host such as `azure.ai.agent` / `azure.ai.project`).
2. Run `azd ai agent init -m <path-or-url-to-azure.yaml>` in an interactive terminal.

Expected: after adopting the manifest, azd prompts for subscription, Foundry project (existing vs new), model configuration, etc., the same way it does when `-m` points at an agent manifest.

Actual: azd adopts the `azure.yaml`, prints next steps, and exits. No subscription / Foundry project / model prompts appear.

## Root cause

PR #8885 added an early-return branch to the init command's `RunE`. When the `-m` pointer is detected as a unified Foundry `azure.yaml` (via `looksLikeFoundryAzureYaml`), it calls `runInitFromAzureYaml` and returns immediately:

- `cli/azd/extensions/azure.ai.agents/internal/cmd/init.go` (routing/early return, around lines 1212-1222)

`runInitFromAzureYaml` only stages the template, scaffolds the project (`scaffoldProject` -> `azd init -t`), stamps `infra.provider: microsoft.foundry`, and prints next steps:

- `cli/azd/extensions/azure.ai.agents/internal/cmd/init_adopt.go` (`runInitFromAzureYaml`, lines 163-216)

It never invokes `InitAction.Run` / `runInitFromManifest`, which is where the interactive Azure setup lives for the agent-manifest path:

- subscription selection via `ensureSubscription`
- Foundry project selection via `selectFoundryProject`
- model configuration via `configureModelChoice`
- deploy-mode selection via `promptDeployMode`

(see `InitAction.Run` around lines 1550-1719 and `configureModelChoice` at 1973+ in `cli/azd/extensions/azure.ai.agents/internal/cmd/init.go`)

Because the adoption path short-circuits before any of that runs, none of the environment setup prompts happen.

## Suggested fix

After adopting the unified `azure.yaml`, the adoption path should still run the interactive environment setup (subscription, Foundry project selection/creation, model configuration) so the user ends up with a provision-ready environment, matching the agent-manifest flow. This likely means invoking the shared Azure-context / `InitAction` setup after `scaffoldProject` instead of returning early.

## Environment

- Extension: azure.ai.agents 1.0.0-beta.1+ (introduced by PR #8885)
- Related PRs/issues: #8885, #8818, #8798, #8884

trangevi:
## Motivation

When `azd ai agent init -m <azure.yaml>` points at a unified Foundry azure.yaml, the command scaffolds the project but returns early without running subscription selection, Foundry project setup, or model deployment verification. Users are left with an environment that cannot provision without manual configuration.

## Approach

This PR adds interactive environment setup to the unified azure.yaml adoption path (`runInitFromAzureYaml`):

**Subscription and Foundry project setup:**
Extracts the shared subscription + Foundry project selection logic from `configureModelChoice`'s `!hasModelResources` branch into a standalone `configureFoundryProject` helper. Both the agent-manifest path and the unified azure.yaml adoption path now call this helper, eliminating duplication and ensuring the adoption path runs the same interactive flow.

**Model deployment verification:**
After project selection, parses model deployments from the azure.yaml and verifies each one against the selected Foundry project's existing deployments. For each declared deployment, the user can:

- **Use an existing matching deployment** -- removes it from azure.yaml (no provisioning needed)
- **Deploy as specified** -- keeps it in azure.yaml for provisioning (only shown when the existing deployment differs from what's declared)
- **Choose a different model** -- full catalog browse or pick another existing deployment
- **Skip** -- removes from azure.yaml entirely

When creating a new project (no existing deployments to compare against), the user is still shown the model details and asked to confirm, change, or skip.

The first selected deployment is persisted as `AZURE_AI_MODEL_DEPLOYMENT_NAME` (same env var as the manifest path).

**Flag support:**
- `--model-deployment <name>`: auto-selects the named existing deployment without prompting (errors if not found)
- `--model <name>`: pre-selects the model in the catalog prompt when choosing a different model
- `--project <id>`: already supported via `configureFoundryProject`
- `--no-prompt`: auto-uses existing matches and auto-deploys unmatched models

## Key design decisions

- `configureFoundryProject` returns a `foundryProjectSetupResult` with `Credential` and `FoundryProjectInfo`, allowing callers to chain dependent operations (like listing deployments).
- YAML parsing uses typed structs with explicit `yaml` tags on `project.Deployment` rather than `map[string]any`, so field mapping is a compile-time contract.
- `verifyAzureYamlDeployments` returns kept entries with their originating service name preserved, plus a `modified` flag -- eliminates the previous length-only rewrite guard and non-deterministic service assignment.
- When the user picks an existing deployment via "Choose a different model", it's only added to `referencedDeployments` (not kept in azure.yaml) since it doesn't need provisioning.
- The azure.yaml is updated via `SetServiceConfigValue` with path `deployments` -- same mechanism used by `setServiceUses` for the `uses:` field.

## Files changed

- **`init_foundry_project_setup.go`** (new) -- extracted `configureFoundryProject` helper
- **`init.go`** -- refactored `!hasModelResources` branch to delegate to the helper
- **`init_adopt.go`** -- wired setup + verification into `runInitFromAzureYaml`; deployment parsing, verification, UX improvements
- **`init_adopt_test.go`** -- unit tests for `foundryDeployments` parser
- **`project/config.go`** -- added `yaml` struct tags to `Deployment` types

Fixes: #8922

### DIFF
--- a/cli/azd/extensions/azure.ai.agents/internal/cmd/init_adopt.go
+++ b/cli/azd/extensions/azure.ai.agents/internal/cmd/init_adopt.go
@@ -149,6 +149,8 @@ func verifyAzureYamlDeployments(
 	envName string,
 	entries []foundryDeploymentEntry,
 	noPrompt bool,
+	modelDeploymentFlag string,
+	modelFlag string,
 ) (deploymentsToKeep []project.Deployment, referencedDeployments []project.Deployment, err error) {
 	// Get the Foundry project ID from the environment.
 	resp, err := azdClient.Environment().GetValue(ctx, &azdext.GetEnvRequest{
@@ -181,6 +183,35 @@ func verifyAzureYamlDeployments(
 	allDeployments, listErr := listProjectDeployments(ctx, credential, subscription, resourceGroup, accountName)
 	if listErr != nil {
 		return nil, nil, fmt.Errorf("failed to list deployments in Foundry project: %w", listErr)
+	}
+
+	// --model-deployment flag: auto-select the named deployment, skip interactive loop.
+	if modelDeploymentFlag != "" {
+		for _, d := range allDeployments {
+			if strings.EqualFold(d.Name, modelDeploymentFlag) {
+				log.Printf("--model-deployment: using existing deployment '%s' (model: %s, version: %s)",
+					d.Name, d.ModelName, d.Version)
+				referencedDeployments = append(referencedDeployments, project.Deployment{
+					Name: d.Name,
+					Model: project.DeploymentModel{
+						Name:    d.ModelName,
+						Format:  d.ModelFormat,
+						Version: d.Version,
+					},
+					Sku: project.DeploymentSku{
+						Name:     d.SkuName,
+						Capacity: d.SkuCapacity,
+					},
+				})
+				// All azure.yaml deployments are removed (existing deployment is used instead).
+				return nil, referencedDeployments, nil
+			}
+		}
+		return nil, nil, exterrors.Validation(
+			exterrors.CodeModelDeploymentNotFound,
+			fmt.Sprintf("model deployment %q not found in Foundry project", modelDeploymentFlag),
+			"verify the deployment name or omit --model-deployment to select interactively",
+		)
 	}
 
 	for _, entry := range entries {
@@ -304,7 +335,7 @@ func verifyAzureYamlDeployments(
 				referencedDeployments = append(referencedDeployments, dep)
 
 			case selected == "change":
-				newDep, err := promptAlternativeDeployment(ctx, azdClient, azureContext, allDeployments)
+				newDep, err := promptAlternativeDeployment(ctx, azdClient, azureContext, allDeployments, modelFlag)
 				if err != nil {
 					return nil, nil, err
 				}
@@ -364,7 +395,7 @@ func verifyAzureYamlDeployments(
 				referencedDeployments = append(referencedDeployments, dep)
 
 			case "change":
-				newDep, err := promptAlternativeDeployment(ctx, azdClient, azureContext, allDeployments)
+				newDep, err := promptAlternativeDeployment(ctx, azdClient, azureContext, allDeployments, modelFlag)
 				if err != nil {
 					return nil, nil, err
 				}
@@ -391,6 +422,7 @@ func promptAlternativeDeployment(
 	azdClient *azdext.AzdClient,
 	azureContext *azdext.AzureContext,
 	allDeployments []FoundryDeploymentInfo,
+	modelFlag string,
 ) (*project.Deployment, error) {
 	// Offer "browse catalog" or "use existing deployment" if any exist.
 	altChoices := []*azdext.SelectChoice{
@@ -421,12 +453,17 @@ func promptAlternativeDeployment(
 	case "catalog":
 		// Use the full model + deployment prompt which handles version,
 		// SKU, and capacity selection (same as manifest path).
+		defaultModel := "gpt-4.1-mini"
+		if modelFlag != "" {
+			defaultModel = modelFlag
+		}
 		promptReq := &azdext.PromptAiModelRequest{
 			AzureContext: azureContext,
 			Filter:       agentModelFilter([]string{azureContext.Scope.Location}, nil),
 			SelectOptions: &azdext.SelectOptions{
 				Message: "Select a model",
 			},
+			DefaultValue: defaultModel,
 		}
 
 		modelResp, err := azdClient.Prompt().PromptAiModel(ctx, promptReq)
@@ -734,7 +771,7 @@ func runInitFromAzureYaml(
 	if len(deploymentEntries) > 0 && result != nil && result.Credential != nil {
 		deploymentsToKeep, referencedDeployments, err := verifyAzureYamlDeployments(
 			ctx, azdClient, result.Credential, azureContext, env.Name,
-			deploymentEntries, flags.noPrompt,
+			deploymentEntries, flags.noPrompt, flags.modelDeployment, flags.model,
 		)
 		if err != nil {
 			if exterrors.IsCancellation(err) {
--- a/cli/azd/extensions/azure.ai.agents/internal/cmd/init_adopt.go
+++ b/cli/azd/extensions/azure.ai.agents/internal/cmd/init_adopt.go
@@ -151,28 +151,32 @@ func verifyAzureYamlDeployments(
 	noPrompt bool,
 	modelDeploymentFlag string,
 	modelFlag string,
-) (deploymentsToKeep []project.Deployment, referencedDeployments []project.Deployment, err error) {
+) (keptEntries []foundryDeploymentEntry, referencedDeployments []project.Deployment, modified bool, err error) {
 	// Get the Foundry project ID from the environment.
 	resp, err := azdClient.Environment().GetValue(ctx, &azdext.GetEnvRequest{
 		EnvName: envName,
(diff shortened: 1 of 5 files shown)


cli/azd/extensions/azure.ai.agents/internal/cmd/init_adopt.go
```diff
@@ -206,6 +696,106 @@ func runInitFromAzureYaml(
...
+		)
+		if err != nil {
+			if exterrors.IsCancellation(err) {
+				return exterrors.Cancelled("initialization was cancelled")
+			}
+			return err
+		}
+
+		// Update the azure.yaml if the kept deployments differ from the original.
+		if len(deploymentsToKeep) != len(deploymentEntries) || len(deploymentsToKeep) == 0 {
```
jongio: This guard compares only list lengths:

```go
if len(deploymentsToKeep) != len(deploymentEntries) || len(deploymentsToKeep) == 0 {
```

When the user picks "Choose a different model", the replacement is appended to `deploymentsToKeep` (same count as original), so this condition is false and the rewrite is skipped. The azure.yaml keeps the old model while `AZURE_AI_MODEL_DEPLOYMENT_NAME` points to the new one.

A content-aware comparison would fix this. For example, track whether any deployment was actually modified:

```suggestion
		deploymentsChanged := len(deploymentsToKeep) != len(deploymentEntries) || len(deploymentsToKeep) == 0
		if !deploymentsChanged {
			for i, kept := range deploymentsToKeep {
				if kept.Name != deploymentEntries[i].Deployment.Name ||
					kept.Model.Name != deploymentEntries[i].Deployment.Model.Name {
					deploymentsChanged = true
					break
				}
			}
		}
		if deploymentsChanged {
```

Note: the order in `deploymentsToKeep` may not match `deploymentEntries` order, so a set-based comparison (or a simple `modified` flag set in the "change" branch of `verifyAzureYamlDeployments`) might be more reliable.

cli/azd/extensions/azure.ai.agents/internal/cmd/init_adopt.go
```diff
@@ -206,6 +696,106 @@ func runInitFromAzureYaml(
...
+				found := false
+				for _, entry := range deploymentEntries {
+					if entry.Deployment.Name == dep.Name || entry.Deployment.Model.Name == dep.Model.Name {
+						found = true
+						break
+					}
+				}
+				if !found {
+					// New model chosen — assign to the first service.
+					for svc := range byService {
```
jongio: When a replacement deployment from "Choose a different model" doesn't match any original entry by name or model name, it's assigned to whichever service the map iterator returns first:

```go
for svc := range byService {
    byService[svc] = append(byService[svc], dep)
    break
}
```

Go map iteration is non-deterministic. For a multi-service azure.yaml this could assign the replacement to the wrong service across runs. Consider tracking which service originally owned the replaced deployment (e.g., by pairing the original `entry.ServiceName` with each kept deployment in `verifyAzureYamlDeployments`) so the replacement inherits the correct service.

cli/azd/extensions/azure.ai.agents/internal/cmd/init_adopt_test.go
```diff
@@ -274,3 +276,153 @@ func TestStageAzureYamlTemplate_LocalRenamesToAzureYaml(t *testing.T) {
...
+
+func TestStringField(t *testing.T) {
+	// stringField was removed — this test is no longer needed.
+	t.Skip("stringField helper removed in favor of typed YAML unmarshal")
+}
+
+func TestIntField(t *testing.T) {
+	// intField was removed — this test is no longer needed.
+	t.Skip("intField helper removed in favor of typed YAML unmarshal")
+}
```
jongio: These always-skipped stubs (`TestStringField`, `TestIntField`) don't test anything. They'll show as passing in CI which masks the fact that no coverage exists for this area. Delete them entirely; the git history preserves the context of why they were removed.

jongio (changes requested):
The extraction of `configureFoundryProject` is a clean refactor that removes real duplication. The deployment verification flow for the adoption path is well-structured and the typed YAML parsing is a good choice over `map[string]any`. Two issues to address before merging.

1. The azure.yaml rewrite guard has a logic gap when the user swaps a model (count stays the same but content changes).
2. Non-deterministic service assignment for replacement deployments that don't match any original entry.

cli/azd/extensions/azure.ai.agents/internal/cmd/init_adopt.go
```diff
@@ -89,6 +94,491 @@ func foundryProjectName(content []byte) string {
...
+		return nil, nil, fmt.Errorf("failed to get AZURE_AI_PROJECT_ID: %w", err)
+	}
+
+	foundryProjectId := resp.Value
+	if foundryProjectId == "" {
+		// No project selected (deferred setup) — keep all deployments as-is.
+		for _, e := range entries {
+			deploymentsToKeep = append(deploymentsToKeep, e.Deployment)
+		}
+		return deploymentsToKeep, nil, nil
```
huimiu: This returns no `referencedDeployments` when `AZURE_AI_PROJECT_ID` is empty. That also happens on the normal create-new-project path before `provision`, so `AZURE_AI_MODEL_DEPLOYMENT_NAME` is never written even though these deployments are kept for provisioning. Would it make sense to return `deploymentsToKeep` as the referenced deployments here, or pass an explicit deferred flag so create-new still records the deployment name?

jongio: Good catch. Looking at the caller in `runInitFromAzureYaml`, the guard around this call is `result != nil && result.Credential != nil`, which should mean `configureFoundryProject` already ran successfully and wrote `AZURE_AI_PROJECT_ID`. But if the user picks "create new project" (deferred provisioning), the ID could still be empty at this point since the project doesn't physically exist yet.

Returning `deploymentsToKeep` as `referencedDeployments` in the early-return path seems like the simplest fix. Those are the deployments the user intends to deploy, so recording the first one's name as `AZURE_AI_MODEL_DEPLOYMENT_NAME` preserves the contract that downstream code can reference it. The alternative (a deferred flag) adds API complexity for a case that can be handled inline.

therealjohn (approved):

jongio (approved):
Both issues from my prior review are addressed:

1. The length-only rewrite guard is replaced by a modified flag set at each decision point. Cleaner than the content comparison I suggested.
2. Non-deterministic service assignment is eliminated by returning []foundryDeploymentEntry with ServiceName preserved from the original entry.

The --model-deployment and --model flag wiring, isExisting distinction in promptAlternativeDeployment, and the exact-match check for hiding redundant "Deploy as specified" are all solid additions.

PR #8935, opened 2026-07-02
fix: add environmentVariables to azure.ai.agent schema

huimiu:
## Problem

`azd ai agent init` writes an agent service entry to `azure.yaml` that includes an `environmentVariables` property — a list of `name`/`value` pairs carried on the inline agent definition (`AgentDefinitionInline.EnvironmentVariables`, JSON key `environmentVariables`). This value is both written at init and read back at deploy (`prepareDeploy` resolves each entry against the azd environment).

However, the `azure.ai.agent` service schema (`cli/azd/extensions/azure.ai.agents/schemas/azure.ai.agent.json`) never defined `environmentVariables`, so the generated manifest didn't match the schema definition.

Also, the existing `env` property in the schema was misleading for `azure.ai.agent`: azd core parses service-level `env` into `project.ServiceConfig.Environment`, but the core-to-extension mapper does not forward that field to `azdext.ServiceConfig`, and the extension-facing proto has no `env`/`Environment` field. As a result, `env` on an `azure.ai.agent` service does not reach the extension or the hosted agent container.

## Fix

- Add the `environmentVariables` array property to `azure.ai.agent.json`, plus an `EnvironmentVariable` definition (`{ name, value }`, both required, `additionalProperties: false`) that mirrors the Go `agent_yaml.EnvironmentVariable` shape.
- Remove the misleading `env` property from `azure.ai.agent.json`.
- Explicitly disallow service-level `env` for `azure.ai.agent` in both root schemas (`schemas/v1.0/azure.yaml.json` and `schemas/alpha/azure.yaml.json`) so editors reject a field that azd core does not pass to extensions.
- Update examples to use `environmentVariables` with the actual model deployment env var shape.

## Validation

- Local AJV validation confirms `environmentVariables` is accepted for `azure.ai.agent` and service-level `env` is rejected.
- JSON parse validation passes for the changed schema files.
- `cspell` passes on the updated `CHANGELOG.md`.
- Existing `TestAgentDefinitionRoundTrip` already confirms the code round-trips `environmentVariables` through the inline service properties — the shape the schema now describes.

No Go code changed; this is a schema/example correction.

### DIFF
--- a/cli/azd/extensions/azure.ai.agents/CHANGELOG.md
+++ b/cli/azd/extensions/azure.ai.agents/CHANGELOG.md
@@ -8,7 +8,7 @@
 
 ### Bugs Fixed
 
-- [[#8935]](https://github.com/Azure/azure-dev/pull/8935) `azd ai agent init` writes an agent service `environmentVariables` list that the `azure.ai.agent` schema did not define. Add the `environmentVariables` property (a list of `name`/`value` pairs) to the schema so the generated `azure.yaml` validates and editors complete it.
+- [[#8935]](https://github.com/Azure/azure-dev/pull/8935) `azd ai agent init` writes an agent service `environmentVariables` list that the `azure.ai.agent` schema did not define. Add the `environmentVariables` property (a list of `name`/`value` pairs) to the schema so the generated `azure.yaml` validates and editors complete it, and disallow service-level `env` for `azure.ai.agent` services because azd core does not forward it to extensions.
 
 ### Other Changes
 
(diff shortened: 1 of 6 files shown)


jongio (approved):
Schema addition correctly mirrors the Go `EnvironmentVariable` struct (`agent_yaml/yaml.go`) and its placement on `AgentDefinitionInline` (`agent_definition.go`). Both `name` and `value` are required strings with `additionalProperties: false`, matching the source of truth. The example exercises both literal values and `$`{ENV_VAR}` substitution. CHANGELOG follows convention.

PR #8936, opened 2026-07-02
feat: forward service env to extensions

### LINKED ISSUE #9008: Forward service-level env values to extension providers
## Current behavior

Extension framework service providers and service target providers receive a `ServiceConfig` that does not include the service-level `env` block declared in `azure.yaml`. Extensions have no way to read those key-value pairs or their expanded values at build, package, or publish time.

## Expected behavior

When `azd` forwards a `ServiceConfig` to an extension provider, the service-level `env` entries from `azure.yaml` are expanded (variable references resolved against the active azd environment) and made available to the extension via `ServiceConfig.environment`.

huimiu:
## Summary
- Adds `ServiceConfig.environment` to the extension gRPC contract.
- Expands service-level `azure.yaml` `env` values with the existing azd environment resolver before forwarding them to extension providers.
- Preserves the field when mapping proto service configs back into core service configs.

## Why
External framework and service target extensions need access to service-specific environment configuration declared in `azure.yaml`. Forwarding this data through the extension contract lets extension authors consume the same service-level `env` values that core azd can already resolve for built-in service flows.

## Impact
- Extension providers can read expanded service environment values from `azdext.ServiceConfig.Environment`.
- Existing extensions remain compatible because the field is omitted when service-level `env` is not configured.
- This is a generic azd core extension-framework change and does not change extension-specific init or schema behavior.

## Tests
- `go test ./pkg/project -run "TestServiceConfigMapping|TestServiceConfigReverseMapping|TestServiceConfigRoundTripMapping|TestFromProtoServiceConfigMapping"`

### DIFF
--- a/cli/azd/internal/grpcserver/framework_service.go
+++ b/cli/azd/internal/grpcserver/framework_service.go
@@ -118,14 +118,10 @@ func (s *FrameworkService) onRegisterRequest(
 	// Register external framework service with DI container, passing the broker
 	err := s.container.RegisterNamedSingleton(language, func(
 		console input.Console,
-	) (project.FrameworkService, error) {
+	) project.FrameworkService {
 		var env *environment.Environment
 		if s.lazyEnv != nil {
-			var err error
-			env, err = s.lazyEnv.GetValue()
-			if err != nil {
-				return nil, fmt.Errorf("loading environment: %w", err)
-			}
+			env, _ = s.lazyEnv.GetValue()
 		}
 		return project.NewExternalFrameworkService(
 			language,
@@ -134,7 +130,7 @@ func (s *FrameworkService) onRegisterRequest(
 			broker,
 			console,
 			env,
-		), nil
+		)
 	})
 
 	if err != nil {
--- a/cli/azd/internal/grpcserver/framework_service.go
+++ b/cli/azd/internal/grpcserver/framework_service.go
@@ -11,10 +11,12 @@ import (
 	"sync"
 
 	"github.com/azure/azure-dev/cli/azd/pkg/azdext"
+	"github.com/azure/azure-dev/cli/azd/pkg/environment"
 	"github.com/azure/azure-dev/cli/azd/pkg/extensions"
 	"github.com/azure/azure-dev/cli/azd/pkg/grpcbroker"
 	"github.com/azure/azure-dev/cli/azd/pkg/input"
 	"github.com/azure/azure-dev/cli/azd/pkg/ioc"
+	"github.com/azure/azure-dev/cli/azd/pkg/lazy"
 	"github.com/azure/azure-dev/cli/azd/pkg/project"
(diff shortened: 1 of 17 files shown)


cli/azd/internal/grpcserver/framework_service.go
```diff
@@ -113,14 +118,23 @@ func (s *FrameworkService) onRegisterRequest(
...
-	) project.FrameworkService {
+	) (project.FrameworkService, error) {
+		var env *environment.Environment
+		if s.lazyEnv != nil {
+			var err error
+			env, err = s.lazyEnv.GetValue()
+			if err != nil {
+				return nil, fmt.Errorf("loading environment: %w", err)
+			}
+		}
```
jongio: This error-propagating factory is inconsistent with the sibling `ServiceTargetService` registration at `service_target_service.go:125`, which uses `env, _ := s.lazyEnv.GetValue()` and proceeds with a possibly-nil env. Since `envResolver(nil)` already returns "" for every key, the safe approach is:

```suggestion
	) project.FrameworkService {
		var env *environment.Environment
		if s.lazyEnv != nil {
			env, _ = s.lazyEnv.GetValue()
		}
```

This preserves the prior behavior (empty expansion when no env is available) and keeps `azd restore`/`build`/`package` working for extension framework services without a selected environment. The corresponding `TestFrameworkService_onRegisterRequest_EnvLoadError` test would need to verify the nil-env path instead of expecting an error on resolve.

jongio (commented):
The proto extension, mapper logic, and test coverage here are solid. One issue to address before merge:

The factory registration in `onRegisterRequest` propagates `lazyEnv.GetValue()` errors, which will fail `restore`/`build`/`package` for extension framework services when no environment is selected. The sibling `ServiceTargetService` (`service_target_service.go:125`) ignores this error with `env, _ := s.lazyEnv.GetValue()`, and `envResolver(nil)` already degrades gracefully (returns "" for every key). This should match that pattern to avoid a behavioral regression.

jongio (approved):
Prior concern addressed. The lazyEnv error is now correctly ignored (matching ServiceTargetService pattern), with nil guard and test coverage for both the error and happy paths.

jongio (approved):
Incremental review of 5f0132b (post-approval commit). The lazy env refactoring is clean: removes the unnecessary envManager indirection from projectService, adds round-trip safety for env templates in AddService, and properly escapes literal values in the reverse proto mapping. Test coverage is thorough. No issues.

JeffreyCA (approved):

jongio (approved):
Reviewed the proto extension, mapper logic (forward + reverse), round-trip template preservation, literal escaping, and shared serviceConfigToProto helper. The graceful env-load-error handling matches the sibling ServiceTargetService pattern correctly. Test coverage is thorough (expansion, error cases, per-call lazy resolution, dollar/backslash escaping). No issues found.

PR #8939, opened 2026-07-02
feat(ai.agents): native Azure Bot + Teams channel for activity-protocol agents

### LINKED ISSUE #8955: [Activity] Support activity protocol

v1212:
Closes #8955

## Summary

Adds **native `azd` support for provisioning the Azure Bot and Microsoft Teams channel** that front an **activity-protocol (Teams) hosted agent**. This replaces the previous `setup-instance-bot.ps1` / `post-deploy.ps1` postdeploy scripts with first-class behavior wired into the `azure.ai.agents` extension's `postdeploy` / `postdown` lifecycle handlers.

The end-to-end user journey is now fully native:

```
azd ai agent init      # azure.yaml keeps the activity declarations
azd provision          # Foundry project + ACR
azd deploy             # agent version active  ->  Azure Bot created + Teams channel enabled
azd down               # agent + Azure Bot torn down
```

Non-activity agents are **completely unaffected** — every Teams/bot-specific code path sits behind a single gate on the agent's activity profile, so existing `responses` / `invocations` agents behave exactly as before.

## What changed (file level)

| File | Change |
| --- | --- |
| `internal/cmd/listen.go` | Wire the two new hooks: `ensureActivityBot` in `postdeployHandler`, `teardownActivityBots` in `postdownHandler`. |
| `internal/cmd/listen_activity.go` *(new)* | `ensureActivityBot` (postdeploy bot provisioning), `teardownActivityBots` (best-effort postdown cleanup), `printTeamsNextSteps`, `readEnvValue`. |
| `internal/project/activity_profile.go` *(new, + test)* | `IsActivityProtocol` / `ResolveActivityProfile` — the single gate that detects activity opt-in (via `protocols: activity_protocol` or `agent_endpoint.protocols: activity`). |
| `internal/pkg/botservice/botservice.go` *(new, + test)* | `EnsureBot` (create/update Azure Bot), `ensureTeamsChannel`, `DeleteBot`, `BotName`, `MessagingEndpoint`. Narrow interfaces over `armbotservice` so tests use fakes (7 tests). |
| `go.mod` / `go.sum` | Add `resourcemanager/botservice/armbotservice v1.2.0`. |

## Design

- **Scope boundary — azd owns the Azure resource plane only.** It creates the Azure Bot, binds it to the agent instance identity, enables the Teams channel, and points the messaging endpoint at the agent. Teams *app packaging* and *install* stay on the M365/Graph plane and are out of scope; `postdeploy` prints a short next-steps guide for those manual steps.
- **Bot identity (Phase 1 "simple" use case).** The bot `msaAppId` is the **agent instance identity client id**, which only exists after the agent version is created — so bot provisioning runs in `postdeploy`, after the version is active. `msaAppType = SingleTenant`, `sku = F0`, `location = global`, `kind = azurebot`.
- **Naming + teardown.** The bot is named `<agent-name>-bot-uai`. BotService names are globally unique, so `postdown` deletes the bot to free the name for future redeploys (best-effort — failures are logged, never block `azd down`).
- **Messaging endpoint** is pinned to `…/agents/<agent-name>/endpoint/protocols/activityProtocol?api-version=2025-05-15-preview`.
- **Idempotent.** The bot is created via create-or-update, so re-running `azd deploy` neither duplicates nor errors.

## Testing

**Unit** — `go build ./...`, `go vet ./...` clean; `internal/project` (activity round-trip) and `internal/pkg/botservice` (7 fake-backed tests) pass.

**Live E2E** — full journey validated in a real subscription (new resource group, greenfield Foundry, North Central US):

| Stage | Result |
| --- | --- |
| `azd ai agent init` | Generated `azure.yaml` preserves the activity declarations (`protocols: activity_protocol`, `agentEndpoint.protocols: [activity]`, `authorizationSchemes: [BotServiceRbac]`). |
| `azd provision` | Foundry project + ACR created via the `microsoft.foundry` provider. |
| `azd deploy` | Agent version `active`; Azure Bot `<agent>-bot-uai` created — `msaAppType=SingleTenant`, `msaAppId` = instance identity client id, `sku=F0`, `location=global`, `kind=azurebot`, **Teams channel enabled**, messaging endpoint = pinned `activityProtocol` URL, `provisioningState=Succeeded`. Verified directly against ARM. |
| re-`azd deploy` | Idempotent — no duplicate bot, no error. |
| `azd down` | Resource group deleted; bot removed and its global name freed. |

## Out of scope / follow-ups

- **Teams app packaging & sideload** (M365/Graph plane) remain manual — `postdeploy` writes a generic `TEAMS_APP_SETUP.md` next to the agent source with the bot id and official Microsoft Learn links.
- **Phase 2 (digital-worker use case)** — blueprint + federated-identity model. `ResolveActivityProfile` currently always resolves the `simple` use case.
- **Local testing** for activity agents (extend `azd ai agent run` to launch the agent + Teams playground on localhost) — tracked in #8959.
- **Ship `quickstart-simple` as an `azd ai agent init` template** — tracked in #8960.
- Live invoke smoke test depends on the agent-invoke work tracked separately.

## Notes for reviewers

- After an activity deploy, `postdeploy` writes a generic `TEAMS_APP_SETUP.md` next to the agent source. It is built only from runtime values (agent name, bot name, `msaAppId`) and links to the official Microsoft Learn docs for packaging and sideloading a Teams app, so it works for any activity agent (init-from-manifest or init-from-code) and is never written for non-activity deployments. A file is used rather than stdout because azd's progress UI does not surface postdeploy handler output.

---

## Update: activity-protocol support in `init-from-code`

The init-from-**local-code** path now supports activity-protocol agents (previously only the init-from-**manifest** path did). Bringing your own activity code and running `azd init` (choosing `activity_protocol`, or `--protocol activity_protocol`) now produces an `azure.yaml` **identical** to the manifest path and fully satisfies `azd deploy`.

- `activity_protocol/v1` added to `knownProtocols` (opt-in; no-prompt default stays `responses/2.0.0`)
- composes the `agentEndpoint` when activity is selected — advertises every selected protocol and appends `authorizationSchemes: [BotServiceRbac]` (pure-activity → `protocols: [activity]`), so the postdeploy bot gate fires for code-initialized agents too
- activity **coexists** with other protocols — the endpoint advertises all selected protocols as siblings (matching the service's per-protocol config); there is no exclusivity restriction
- the `azd init` flow is otherwise **unchanged** for non-activity agents

Scope: **simple** use case (Phase 1). Digital worker is Phase 2.

---

### Automated

- Full `azure.ai.agents` extension suite: `go build ./...`, `go vet ./...`, `go test ./...` — green. (The only failures locally are `internal/synthesis` `TestARMTemplate_MatchesBicepBuild` / `TestBrownfieldARMTemplate_MatchesBicepBuild`, which are stale-ARM-vs-local-bicep environmental failures — `templates/` and `internal/synthesis/` are byte-identical to `main`, so unrelated to this PR.)
- New unit tests: `botservice` (EnsureBot create/idempotent, Teams channel, DeleteBot), `activity_profile` (IsActivityProtocol / ResolveActivityProfile / ActivityAgentEndpoint), `init_from_code` (activity_protocol flag path + coexistence with other protocols + KnownProtocolNames).

### Local `azd init` (built via `azd x build`, run through branch azd)

| Scenario | Result |
| --- | --- |
| init **from manifest** (activity) | `azure.yaml` carries `agentEndpoint{protocols:[activity], authorizationSchemes:[BotServiceRbac]}` + `protocols:[activity_protocol/v1]` |
| init **from code** (`--protocol activity_protocol`) | activity declaration **byte-identical** to the manifest path |
| init **from code** (default, no `--protocol`) | `protocols:[responses/2.0.0]`, **no** `agentEndpoint` — non-activity flow unchanged (regression check) |
| init **from code** (`activity_protocol` + `responses`) | both kept — endpoint advertises `protocols:[activity, responses]` + `authorizationSchemes:[BotServiceRbac]`; activity coexists with responses |

### Live end-to-end (init-**from-code** activity project; sub `ActivityProtocol-Dev`, North Central US)

`azd provision` (2m24s, Foundry project + ACR) → `azd deploy` (3m7s, agent active) → **postdeploy** auto-provisioned the Azure Bot `echocode-bot-uai` (SingleTenant, `msaAppId` = agent instance identity, F0/global, endpoint pinned to the `activityProtocol` endpoint, MsTeams + WebChat + DirectLine channels enabled, `provisioningState: Succeeded`) → `azd down --purge` (1m45s, resource group deleted and bot name freed). Bot state verified directly via ARM (`Microsoft.BotService/botServices`, api-version 2022-09-15). The equivalent manifest-path project was validated the same way in earlier testing.

### Two live reference environments (Teams-verified)

To exercise **both `azd ai agent init` entry points**, the same `quickstart-echo` activity agent was deployed twice — identical in every respect **except the init entry point** — and both were sideloaded into Microsoft Teams and confirmed to echo replies. Both deploys use **ACR remote build** (`docker.remoteBuild: true`), not ZIP code-deploy. Subscription: `ActivityProtocol-Dev`, region `northcentralus`. The flow was fully non-interactive with **no manual patches** (`azd ai agent init` -> `azd provision` -> `azd deploy`).

| | **Env A — init from manifest** | **Env B — init from code** |
| --- | --- | --- |
| Init command | `azd ai agent init -m agent.manifest.yaml --deploy-mode container` | `azd ai agent init --deploy-mode container` (from source, no manifest) |
| Foundry agent | `echomani07031202` | `echocode07031202` |
| Resource group | `rg-mani07031202-69e12725` | `rg-code07031202-f68e547e` |
| Foundry project | `mani07031202` | `code07031202` |
| Azure Bot | `echomani07031202-bot-<salt>` | `echocode07031202-bot-<salt>` |
| Deploy mode | container / ACR remote build | container / ACR remote build |
| Teams reply | verified | verified |

Each deploy auto-created the Foundry project + hosted agent, the Azure Bot with the **Teams channel enabled**, and the `Azure AI User` role for the instance identity. Both agents were sideloaded via a per-agent Teams app package and **replied successfully in Teams**, confirming the native manifest and code init paths are functionally equivalent end-to-end.

---

## Follow-up: ship `quickstart-echo` as an `azd ai agent init` template

Add the `quickstart-echo` (simple activity/Teams) sample to the `azd ai agent init` template gallery so users can scaffold it directly. **Not possible yet** — the sample (`foundry-samples`: `samples/python/hosted-agents/bring-your-own/activity/echo`) is not merged yet. Once it lands, ship it as a native template with the changes below. Tracked in #8960.

### To ship the sample as an azd template (native flow)

- Remove `azure.yaml`, `infra/`, and the old scripts (`post-deploy.ps1` / `setup-instance-bot.ps1`) — `azd ai agent init` scaffolds the native `azure.yaml` (foundry provider) and the extension handles the Bot + Teams channel in postdeploy.
- Pin `azure-ai-agentserver-core` in `requirements.txt` to the same datestamp as `...-activity` (avoids pulling an incompatible core -> container crash -> no Teams reply).
- (Optional) Parameterize `manifest.json` `name.short` so multiple installs are distinguishable.

### To just use the sample directly

- Interactive `azd ai agent init` -> `provision` -> `deploy` (sub/location are picked in the init prompts — no manual env setup).
- If core isn't pinned, pin it yourself to avoid the version risk.
- Teams packaging + install is manual either way — `azd deploy` just emits a guide (`TEAMS_APP_SETUP.md`).

**Bottom line:** templating ~= drop `azure.yaml`/`infra`/old scripts + pin core; direct use is basically ready to run, with core-pinning the only real gotcha.

---

## Update: default activity version is now `2.0.0`

Following confirmation from the Foundry service team, the activity protocol now defaults to **`2.0.0`** — the official, recommended version. `1.0.0`/`v1` is still accepted but deprecated going forward. The version only selects the platform-internal container route (`v1`/`1.0.0` -> `/api/messages`, `2.0.0` -> `/activity/messages`); that hop is **Bot Service -> container inside the platform**, so the Teams client, the Bot Service messaging endpoint, and the agent sample are all unaffected — **no sample or playground change is required**.

- `knownProtocols` activity entry is now `activity` / `2.0.0` (canonical name `activity`; legacy alias `activity_protocol` still resolves).
- Unit-test fixtures updated to assert `2.0.0`; legacy `activity_protocol` / `v1` fixtures retained for back-compat coverage.
- Re-validated end-to-end on `2.0.0`: both **init-from-manifest** and **init-from-code** `quickstart-echo` agents redeployed (sub `ActivityProtocol-Dev`, `northcentralus`), each fronted by an Azure Bot with the Teams channel enabled, sideloaded into Microsoft Teams, and **confirmed replying** on the `2.0.0` route.

This supersedes the `activity_protocol/v1` version references in the sections above.


cli/azd/extensions/azure.ai.agents/internal/cmd/listen_activity.go
```diff
@@ -0,0 +1,341 @@
...
+	fmt.Println(output.WithHighLightFormat("\nTeams bot ready."))
+	fmt.Printf("  Azure Bot:  %s (Microsoft Teams channel enabled)\n", botName)
+	fmt.Printf("  Bot ID:     %s\n", msaAppID)
+	if guidePath != "" {
+		fmt.Println(output.WithGrayFormat(fmt.Sprintf(
+			"  Next steps (package + sideload the Teams app): see %s", guidePath,
+		)))
+	} else {
+		fmt.Println(output.WithGrayFormat(
+			"  Next steps: package the Teams app (bots[].id = the Bot ID above) and " +
```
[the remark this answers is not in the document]
v1212: Fixed in 0e67902b8 — the next-steps message now references `bots[].botId` to match the guide and the manifest field.

cli/azd/extensions/azure.ai.agents/internal/cmd/listen_activity_test.go
```diff
@@ -0,0 +1,65 @@
...
+
+	"github.com/azure/azure-dev/cli/azd/pkg/azdext"
+)
+
+func TestTeamsSetupGuideContent(t *testing.T) {
+	const msaAppID = "11111111-2222-3333-4444-555555555555"
+	content := teamsSetupGuideContent("echo-agent", "echo-agent-bot-uai", msaAppID)
+
+	// The bot id is the one value the user must not get wrong: it has to be
+	// carried verbatim into the Teams manifest bots[].id.
```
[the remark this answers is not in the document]
v1212: Fixed in 0e67902b8 — the test comment now references `bots[].botId` to stay consistent with the assertion below.

cli/azd/extensions/azure.ai.agents/internal/project/activity_profile.go
```diff
@@ -0,0 +1,81 @@
...
+	// msaAppId is the agent instance identity client id.
+	ActivityUseCaseSimple ActivityUseCase = "simple"
+	// ActivityUseCaseDigitalWorker is the blueprint + federated-identity model
+	// (Phase 2). Not yet resolved by ResolveActivityProfile.
+	ActivityUseCaseDigitalWorker ActivityUseCase = "digital_worker"
+)
+
+// ActivityProfile summarizes the activity-protocol characteristics of a hosted
+// agent definition. It is the single gate that keeps all Teams/bot-specific
+// behaviour off the path of non-activity agents: when IsActivity is false the
```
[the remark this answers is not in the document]
v1212: Fixed in 0e67902b8 — switched to the American spelling `behavior` and removed the `behaviour` cspell allow-list entry.

cli/azd/extensions/azure.ai.agents/internal/pkg/botservice/botservice.go
```diff
@@ -0,0 +1,232 @@
...
+		}
+		return fmt.Errorf("botservice: deleting bot %q: %w", botName, err)
+	}
+	return nil
+}
+
+func isNotFound(err error) bool {
+	var respErr *azcore.ResponseError
+	return errors.As(err, &respErr) && respErr.StatusCode == http.StatusNotFound
+}
```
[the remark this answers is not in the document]
v1212: Fixed in 0e67902b8 — `isNotFound` now uses `errors.AsType[*azcore.ResponseError]` per the AGENTS.md convention.

jongio (approved):
Activity-protocol bot provisioning is well-gated behind IsActivityProtocol, idempotent via PUT, and correctly scoped to Phase 1 (simple use case). Teardown is properly best-effort. Test coverage looks solid across botservice, activity_profile, and init_from_code paths.

cli/azd/extensions/azure.ai.agents/internal/cmd/init_from_code.go
```diff
@@ -545,6 +546,15 @@ func (a *InitFromCodeAction) createDefinitionFromLocalAgent(ctx context.Context)
 		CodeConfiguration: codeConfig,
 	}
 
+	// An activity-protocol (Teams) agent additionally advertises the friendly
```
trangevi: This protocol isn't unique to Teams, is it? If not, we should remove these specific mentions, for clarity

cli/azd/extensions/azure.ai.agents/internal/cmd/init_from_code.go
```diff
@@ -893,6 +903,7 @@ type protocolInfo struct {
 var knownProtocols = []protocolInfo{
 	{Name: "responses", Version: "2.0.0"},
 	{Name: "invocations", Version: "1.0.0"},
+	{Name: "activity_protocol", Version: "v1"},
```
trangevi: I don't suppose this is something we can push on elsewhere for the version to be consistent with the other protocols?

trangevi: Should this just be "activity", since it's in a list of protocols, we already know that, "activity_protocol" is redundant

cli/azd/extensions/azure.ai.agents/internal/cmd/init_from_code.go
```diff
@@ -994,9 +1008,35 @@ func promptProtocols(
...
+	}
+
 	return records, nil
 }
 
+// validateProtocolSelection rejects combining activity_protocol with any other
+// protocol. An activity-protocol (Teams) agent exposes only the activity
+// endpoint, matching how such agents are declared in a manifest; mixing it with
+// responses/invocations would produce an azure.yaml that does not match the
+// manifest shape and is not a supported activity configuration.
```
trangevi: This surprises me, I don't understand this limitation. What "manifest" is being mentioned, and why does having other endpoints somehow break the activity protocol integration?

cli/azd/extensions/azure.ai.agents/internal/cmd/listen_activity.go
```diff
@@ -0,0 +1,370 @@
...
+
+	fmt.Printf("Configuring Azure Bot %q for Teams (activity protocol)...\n", botName)
+	if err := botClient.EnsureBot(ctx, cfg); err != nil {
+		return err
+	}
+
+	// Write a persistent, generic setup guide next to the agent code (the azd
+	// progress UI swallows postdeploy stdout, so a file is the reliable way to
+	// hand the user the manual M365 steps) and print a short pointer to it.
+	guidePath := writeTeamsSetupGuide(proj, svc, agentName, botName, msaAppID)
```
trangevi, line 135: The code comment for `ensureActivityBot` says it provisions the Teams channel, but this `writeTeamsSetupGuide` implies that's not accurate. I don't think it makes sense to automatically create the Teams channel, so this implementation seems correct, but if that understanding is correct then the top level code comment should be adjusted.

cli/azd/extensions/azure.ai.agents/internal/cmd/listen_activity.go
```diff
@@ -0,0 +1,370 @@
...
+	return guidePath
+}
+
+// teamsSetupGuideContent renders the Teams onboarding guide markdown. It gives
+// concrete, minimal step-by-step instructions for the two manual actions
+// (package the Teams app, then sideload it) and links to the official docs for
+// detail. The single value the user must not get wrong is the bot id: a Teams
+// app manifest's bots[].botId MUST equal this bot's msaAppId, which azd bound to
+// the agent instance identity.
+func teamsSetupGuideContent(agentName, botName, msaAppID string) string {
```
trangevi, line 179: This would be much better off in an actual .md file so it's easier to catch any formatting errors.

cli/azd/extensions/azure.ai.agents/internal/pkg/botservice/botservice.go
```diff
@@ -0,0 +1,232 @@
...
+			MsaAppType:     to.Ptr(armbotservice.MsaAppTypeSingleTenant),
+			MsaAppTenantID: new(cfg.TenantID),
+		},
+	}
+
+	if _, err := c.bots.Create(ctx, cfg.ResourceGroup, cfg.BotName, bot, nil); err != nil {
+		return fmt.Errorf("botservice: creating/updating bot %q: %w", cfg.BotName, err)
+	}
+
+	return c.ensureTeamsChannel(ctx, cfg.ResourceGroup, cfg.BotName)
```
trangevi, line 201: This seems to contradict my comment above about the code giving instructions for setting up the Teams channel. If we set it up here, what are the instructions that we print for?

I'm still also of the opinion that we shouldn't be doing this by default, there should have to be some kind of opt in process

trangevi (changes requested):

jongio (approved):
Incremental review of the review-feedback commit (ea1abee). Changes cleanly address trangevi's feedback:

- Protocol renamed to canonical `activity` with backward-compat `IsActivityProtocolName()` accepting the legacy `activity_protocol` value
- Version aligned to `1.0.0` semver style with routing semantics documented
- Template extracted to embedded `assets/teams_app_setup_guide.md` (lintable, editor-friendly)
- Comments clarified: Azure Bot "Teams channel" (resource toggle) vs Teams "app" (M365 packaging/sideload) distinction is now explicit throughout
- `validateProtocolSelection` comment rewritten to explain the Phase-1 scoping choice vs platform limitation
- Test coverage properly updated for both old and new protocol names

No new issues. Backward compat is solid. CI lint/cspell/golangci-lint all green.

cli/azd/extensions/azure.ai.agents/internal/cmd/init_from_code.go
v1212: Right, it isn't Teams-specific. Removed the "(Teams)" qualifiers from the protocol/endpoint-layer comments across the extension.

cli/azd/extensions/azure.ai.agents/internal/cmd/init_from_code.go
v1212: Done — bumped to `1.0.0` for consistency with the other protocols (routing-equivalent to `v1` on the platform). For context: neither value originated in this PR — `v1` came from Vienna and the `activity_protocol` name from a pre-existing definition. Both are now defined in one central place and changed there.

v1212: Agreed — we emit `activity` now. The legacy `activity_protocol` value is still accepted on read for back-compat.

cli/azd/extensions/azure.ai.agents/internal/cmd/init_from_code.go
v1212: Sorry, that comment was misleading — dropped the "manifest" wording. It isn't a platform limit, it's azd scoping: an activity agent's `ActivityAgentEndpoint` replaces the whole `AgentEndpoint`, so it can't be composed with other protocols in one definition. Rewrote the comment and error accordingly.

cli/azd/extensions/azure.ai.agents/internal/cmd/listen_activity.go
v1212: Fixed. azd provisions only the Azure resource plane — the Azure Bot + its resource-plane Teams *channel*. The generated guide covers the separate manual step (packaging/sideloading the M365 Teams *app*). Reworded both so they no longer conflict.

cli/azd/extensions/azure.ai.agents/internal/cmd/listen_activity.go
v1212: Done — moved it to `internal/cmd/assets/teams_app_setup_guide.md`, embedded via `go:embed` and rendered with `text/template`. Rendered output is byte-identical.

cli/azd/extensions/azure.ai.agents/internal/pkg/botservice/botservice.go
v1212: Same clarification as above: `ensureTeamsChannel` enables the required resource-plane Teams channel connector (bound to the agent identity via `MsaAppID`); it does not create or publish the Teams app. Updated both docs to be consistent.

cli/azd/extensions/azure.ai.agents/internal/cmd/init_from_code.go
trangevi: We should be consistent with the service (for both the name and the version), otherwise there's potential for confusion as things change or are updated in the future. If there were going to be a change, it should be in the service, but without that this doesn't feel like something we should be customizing.

@therealjohn , any preference?

cli/azd/extensions/azure.ai.agents/internal/cmd/init_from_code.go
trangevi: @therealjohn, tagging you here as well, as we talked about this briefly.

I still don't understand this. If this isn't a service side limitation, I don't understand why we're introducing it in AZD (apologies if I misunderstand your comment). But if there are two different properties, why not have two different properties, instead of limiting the functionality for the sake of reuse?

cli/azd/extensions/azure.ai.agents/internal/pkg/botservice/botservice.go
trangevi: @therealjohn , any issues with this?

cli/azd/extensions/azure.ai.agents/internal/cmd/init_from_code.go
therealjohn: If the service uses `v1` we should use that - and if it does, where do we file a bug to track updating it to 1.0.0 like the other protocols?

cli/azd/extensions/azure.ai.agents/internal/cmd/init_from_code.go
therealjohn: @v1212 Please explain the limitation - I also do not understand this. 

I'm reading it to mean that an agent with Activity protocol support can't support any other protocol and I don't think that is true. What needs to change in AZD to support that Activity is just another protocol used?

cli/azd/extensions/azure.ai.agents/internal/pkg/botservice/botservice.go
therealjohn: No issues with this - doing it by default is expected. @v1212, should we also enable the M365 channel? 

What does the Foundry Portals experience for publishing expect to be set?

cli/azd/extensions/azure.ai.agents/internal/cmd/listen.go
```diff
@@ -351,6 +351,20 @@ func postdeployHandler(ctx context.Context, azdClient *azdext.AzdClient, args *a
 		return nil
 	}
 
+	// For activity agents, provision the Azure Bot + Teams
+	// channel bound to the agent instance identity. No-op for other agents.
+	if err := ensureActivityBot(
```
RickWinter: ensureActivityBot runs only after the three early `return nil` guards above (FOUNDRY_PROJECT_ENDPOINT missing at line 320, AZURE_TENANT_ID missing at line 336, credential creation failing at line 348), all of which were written for optimization reporting. For an activity agent those guards silently skip the Teams bot as well, and the only trace is a "skipping optimization reporting" log line that never mentions the bot, so a required connector can quietly go unprovisioned while `azd deploy` reports success. That is also inconsistent with the branch just below, where a real EnsureBot error hard-fails the deploy. Can we resolve the activity profile first and then fetch endpoint/tenant with bot-specific error handling, so the skip-versus-fail decision for the bot is made on its own terms rather than inheriting optimization's preconditions? In the happy path both values are set by provision, so this is about the failure modes, not the common case.

RickWinter (commented):
This adds native provisioning of the Azure Bot and Microsoft Teams channel for activity-protocol agents, wired into the extension's postdeploy/postdown handlers, and extends the init-from-code path to emit the same activity declaration the manifest path already produced. The shape is right. A single ResolveActivityProfile gate keeps every Teams/bot code path off non-activity deploys, the botservice client is fronted by narrow botsAPI/channelsAPI interfaces so the seven tests run against fakes, and EnsureBot is a create-or-update so redeploys stay idempotent. The protocol constant rename from "activity_protocol" to "activity" is low-risk here: the canonical constant has no non-test caller, and detection flows through IsActivityProtocolName, which still accepts the legacy value.

One thing worth resolving before merge: the postdeploy bot provisioning is placed after the optimization-reporting early returns (missing FOUNDRY_PROJECT_ENDPOINT, missing AZURE_TENANT_ID, or a credential failure each return nil with a "skipping optimization reporting" log). For an activity agent those guards silently skip the required Teams bot while deploy reports success, which also contradicts the hard-fail path immediately below where an EnsureBot error fails the deploy. The happy path is fine because provision sets both values, so this is a failure-mode and coupling concern, not a common-case bug.

Leaving Teams app packaging and sideload out of azd is the correct scope boundary, and TEAMS_APP_SETUP.md being built only from runtime values keeps it generic. Build, vet, and the botservice/project/cmd tests pass locally.

Disposition: address the postdeploy coupling before merge; the rest can proceed.

cli/azd/extensions/azure.ai.agents/internal/cmd/init_from_code.go
v1212: Name: the service already renamed the union value  activity_protocol  →  activity .  AgentEndpointProtocol.ActivityValue = "activity"  is the current canonical;  activity_protocol  is kept only as  LegacyActivityProtocolValue  for read back-compat, and   ProtocolVersionRecord.IsActivityProtocol()  matches both ( AgentEndpointProtocol.cs ,  HostedAgentDefinitionExtensions.cs ). So  activity  matches the service's canonical name, our  IsActivityProtocolName  mirrors its dual-spelling gate, and there's no bug to file  — that rename already landed service-side.

 Version: the service accepts both  v1  and  1.0.0 , both routing to  /api/messages ; only  2.0.0  routes to  /activity/messages  ( MainContainerEnvironmentConstants.SupportedActivityProtocolVersions  +  GetActivityProtocolContainerPath ). Our sample serves   /api/messages , so  2.0.0  (the service "recommended") isn't usable; we pick  1.0.0  to match the semver form of  responses / invocations .  v1  is equally valid if you'd prefer the older spelling.

this PR adopts the service's renamed name ( activity ) and an accepted version ( 1.0.0 ) while keeping legacy  activity_protocol / v1  read support — so it's consistent with the service, not customizing.  @trangevi  @therealjohn

cli/azd/extensions/azure.ai.agents/internal/pkg/botservice/botservice.go
v1212: yes, the behavior and result are the same, here is the test from this PR and result from portal [Publish - Teams and M365], M365 channel by default is not connected, developer can do that from Azure portal if needed. @therealjohn  @trangevi 
test with this pr ->
image

test with portal ->
image

cli/azd/extensions/azure.ai.agents/internal/cmd/listen.go
v1212: Good catch — fixed by decoupling the two concerns.

postdeployHandler  now resolves the activity profile first, then gathers the shared inputs (endpoint/tenant/cred) once with no skip-vs-fail decision in the gatherPostdeployInputs. Each step then applies its own policy:

 • Activity bot (required): a missing input hard-fails the deploy with a bot-specific error, consistent with the  EnsureBot  failure path below — no more silent skip.
 • Optimization reporting (best-effort): unchanged, logs and skips on a missing input, never fails the deploy.

RickWinter (commented):
This replaces the setup-instance-bot.ps1 / post-deploy.ps1 postdeploy scripts with native Azure Bot plus Teams-channel provisioning wired into the extension's postdeploy/postdown lifecycle, gated entirely behind an activity-protocol check so non-activity agents are untouched. The layering is right: the botservice package owns the ARM resource plane, activity_profile is the single gate, and the postdeploy input-gathering refactor cleanly separates the two policies (the required bot fails the deploy, best-effort optimization reporting only logs and skips). The narrow botsAPI/channelsAPI interfaces over armbotservice with fake-backed tests are the right call, EnsureBot idempotency and not-found-tolerant DeleteBot are handled correctly, and BotName salts on subscription plus resource group to avoid the global-uniqueness collision. The modern Go idioms (new(value), errors.AsType) are fine on 1.26.

The design point worth resolving is the activity exclusive-vs-coexistence contract. ComposeActivityAgentEndpoint is deliberately built to advertise multiple protocols on the endpoint and dedup them so activity can coexist with responses/invocations, but the description says the opposite (activity is exclusive, rejected via validateProtocolSelection). That function does not exist anywhere in the tree, and nothing prevents an agent from selecting activity plus responses, in which case postdeploy still binds a single bot to the activityProtocol endpoint while the agent is also advertised as responses. Decide which contract is intended: if activity is exclusive, the multi-protocol machinery in Compose is unreachable and can collapse to ActivityAgentEndpoint(); if coexistence is intended, the description needs correcting and the mixed case wants a test.

Separately, the description has drifted from the code in a few spots (knownProtocols is activity/1.0.0, not activity_protocol/v1; the bot is <agent>-bot-<hash>, not -bot-uai; the validateProtocolSelection exclusivity claim). Worth tidying so users and future reviewers are not misled.

Intentionally left alone and correct: the AgentProtocolActivityProtocol constant rename to "activity" is safe, since its only references are by symbol (models.go, models_test.go) and reads accept both values via IsActivityProtocolName. Disposition: address the contract question and the description drift before merge; the orphaned ActivityAgentEndpoint helper can follow. Nothing blocks; the core resource-plane logic is sound.

cli/azd/extensions/azure.ai.agents/internal/cmd/init_from_code.go
```diff
@@ -545,6 +545,19 @@ func (a *InitFromCodeAction) createDefinitionFromLocalAgent(ctx context.Context)
...
 
+	// An activity agent additionally advertises the friendly "activity" endpoint
+	// guarded by BotServiceRbac. We compose this into any existing agent_endpoint
+	// rather than overwriting it, so Activity can coexist with the other
+	// protocols the agent selected (responses/invocations). Injecting it here
+	// mirrors the manifest-based init so the generated azure.yaml is identical and
+	// `azd deploy` provisions the Azure Bot connector. Phase 1 covers the simple
+	// use case; digital-worker is a Phase 2 addition. No-op for non-activity agents.
+	if project.IsActivityProtocol(*definition) {
+		definition.AgentEndpoint = project.ComposeActivityAgentEndpoint(
```
RickWinter, line 556: This folds every selected protocol into the endpoint so activity can coexist with responses/invocations, but the PR description says activity is exclusive and rejected when combined (via a validateProtocolSelection that does not exist in the tree). As written there is no guard against an agent selecting activity plus responses, and postdeploy will then bind one bot to the activityProtocol messaging endpoint while the agent still advertises responses. Can we pin down the intended contract? If activity is meant to be exclusive, this call and ComposeActivityAgentEndpoint's multi-append logic reduce to ActivityAgentEndpoint(); if coexistence is intended, let's add validation or test coverage for the mixed case and correct the description.

cli/azd/extensions/azure.ai.agents/internal/project/activity_profile.go
```diff
@@ -0,0 +1,137 @@
...
+	return ActivityProfile{IsActivity: true, UseCase: ActivityUseCaseSimple}
+}
+
+// ActivityAgentEndpoint returns the agent_endpoint declaration an activity agent
+// requires: the friendly "activity" protocol guarded by the BotServiceRbac
+// authorization scheme. `azd init` uses it to make an activity agent scaffolded
+// from local code carry the exact same declaration as one initialized from a
+// manifest, so the generated azure.yaml is identical and `azd deploy` provisions
+// the Azure Bot connector. Phase 1 covers the simple use case.
+func ActivityAgentEndpoint() *agent_yaml.AgentEndpoint {
```
RickWinter: nit: ActivityAgentEndpoint() is only referenced from activity_profile_test.go now; every production path goes through ComposeActivityAgentEndpoint. It reads as the expected-shape oracle for the tests, which is reasonable, but it is exported from the project package and unused in any real caller. If we do not expect a production caller, consider moving it into the test as the fixture so the package surface stays minimal.

cli/azd/extensions/azure.ai.agents/internal/cmd/init_from_code.go
v1212: Coexistence is the intended contract. `validateProtocolSelection` was removed in this revision (that's why it isn't in the tree) — the PR description is stale and I'll update it to drop the "exclusive/rejected" wording.

Mixed-case coverage is already in:
- `init_from_code_test.go` — an activity+responses selection now yields endpoint protocols `[activity, responses]` (no rejection).
- `TestComposeActivityAgentEndpoint` — a coexistence subtest, plus legacy-normalization and existing-schemes-preserved cases.
- an end-to-end regression test driving both init paths.

On postdeploy: the bot binds only the activity messaging endpoint (activity's inbound channel); responses/invocations stay served on their own protocol routes as siblings, matching the service's per-protocol config, so advertising them alongside activity is intentional and unaffected. Pure-activity output stays byte-identical to the old `ActivityAgentEndpoint()`.

cli/azd/extensions/azure.ai.agents/internal/cmd/init_from_code.go
v1212: You're right — activity is just another protocol, and it's no longer restricted. The limitation was self-imposed in azd (a Phase-1 scoping shortcut), not a platform limit.

What was limiting it: init overwrote the whole `AgentEndpoint` with an activity-only endpoint, and `validateProtocolSelection` rejected combining activity with other protocols.

What changed: init now **composes** the endpoint — advertises every selected protocol and appends the `BotServiceRbac` auth scheme instead of overwriting — and `validateProtocolSelection` is removed. So activity coexists with responses/invocations in one definition. Pure-activity output is byte-identical to before; non-activity paths are untouched.  @trangevi  @therealjohn

cli/azd/extensions/azure.ai.agents/internal/project/activity_profile.go
v1212: Done — moved it into `activity_profile_test.go` as the unexported fixture `expectedActivityEndpoint()` (the expected-shape oracle for the tests). `ActivityAgentEndpoint()` is removed from the `project` package, so every production path now goes solely through `ComposeActivityAgentEndpoint` and the package surface stays minimal. Build/vet/`go test ./...` green.

cli/azd/extensions/azure.ai.agents/internal/cmd/init_from_code.go
```diff
@@ -903,7 +906,12 @@ type protocolInfo struct {
 var knownProtocols = []protocolInfo{
 	{Name: "responses", Version: "2.0.0"},
 	{Name: "invocations", Version: "1.0.0"},
-	{Name: "activity_protocol", Version: "v1"},
+	// "activity" is the canonical protocol name (legacy alias: "activity_protocol").
+	// Version is a platform routing selector, not a free version string: "v1" and
+	// "1.0.0" route to /api/messages, "2.0.0" to /activity/messages. Our sample
```
trangevi: Okay, so why are we using 1.0.0 instead of 2.0.0? I'm assuming that the service wants things moved to 2.0.0, and idk what sample this comment is referring to but I'm guessing we shouldn't be basing API decisions on a client sample?

trangevi (approved):
Since this last comment is in the "from code" flow, it's highly customizable and doesn't necessarily need to be a blocker in my mind.

cli/azd/extensions/azure.ai.agents/internal/cmd/init_from_code.go
trangevi: I don't understand why "our sample" is what we're basing this decision off of? If the service has moved to 2.0.0, I would expect we would want to as well?

cli/azd/extensions/azure.ai.agents/internal/cmd/listen_activity.go
```diff
@@ -0,0 +1,293 @@
...
+
+// teamsSetupGuideFile is the name of the generated Teams onboarding guide.
+const teamsSetupGuideFile = "TEAMS_APP_SETUP.md"
+
+// writeTeamsSetupGuide writes a generic, simplified Teams onboarding guide next
+// to the agent source so the user can package and sideload their Teams app after
+// deploy. It returns the written path, or "" on any failure (best-effort: never
+// blocks or fails the deploy). The guide is deploy-agnostic and links to the
+// official Microsoft Learn docs rather than any sample-specific scripts.
+func writeTeamsSetupGuide(
```
trangevi, line 148: Should all of this logic go in with the "next steps" handling? That's what that stuff is there for, right?

cli/azd/extensions/azure.ai.agents/internal/cmd/init_from_code.go
v1212: checking with service team and confirm which protocol version will be recommended, the agreement for this PR requirement was to support activity agent - simple mode, based on the quickstart samples  - https://github.com/microsoft-foundry/foundry-samples-pr/tree/hameed-kunkanoor/activity-samples/samples/python/hosted-agents/bring-your-own/activity.

for sure it will be reasonable to clarify and confirm - which activity protocol version will be recommended for long term? if multi versions are supported, we should need extra UI to support protocol selection and also prompts to identify what each works for.

cli/azd/extensions/azure.ai.agents/internal/cmd/init_from_code.go
v1212: as replied above, this was determined by discussion with service owner, will check with @Hameedkunkanoor to confirm and improve here.

cli/azd/extensions/azure.ai.agents/internal/cmd/listen_activity.go
v1212: Good point, but [next step] doesn't fit cleanly here:

Timing/visibility.  nextstep resolvers only fire at the end of our own subcommands ( init / run / invoke / show ) — there's no hook after azd deploy . And it prints to stdout, which azd's live progress UI hides during postdeploy. That's why we write the durable TEAMS_APP_SETUP.md  instead.

Data.  nextstep  builds its state from the azd environment only. The bot's  msaAppId  is available after deploy but read live via GetAgentVersion , not persisted to env — so routing through  nextstep  would mean writing it back just to read it again.

This block isn't azd task execution — it's a best-effort onboarding guide. Postdeploy is where the data is at hand and the file reliably survives the swallowed stdout, so I'd prefer to keep it here.

cli/azd/extensions/azure.ai.agents/internal/cmd/init_from_code.go
v1212: update - confirmed with service team - @Hameedkunkanoor 

2.0.0
2.0.0 will be official version but we have few consumers who are using 1.0.0 that will be deprecated going forward
 
azd extension follows this design to implement in this PR, new commit was added and tested.

cli/azd/extensions/azure.ai.agents/internal/cmd/init_from_code.go
v1212: new commits to apply the recommended version and ensure tests worked.

2.0.0
2.0.0 will be official version but we have few consumers who are using 1.0.0 that will be deprecated going forward

trangevi (approved):

jayzhang:
After deploy success, the output includes Agent playground (portal). When open the playground link, it says "playground not available for the the agent, deploy the agent to test in Teams or M365 Copilot". Is it by design for showing the playground link that is not testable?
image
image

PR #8941, opened 2026-07-02
Stop assigning Azure AI User role to per-agent managed identity after deploy

### LINKED ISSUE #8940: Stop assigning Azure AI User role to per-agent managed identity after deploy
### Problem

When deploying a hosted agent, the `azure.ai.agents` extension assigns the `Azure AI User` role to each hosted agent's per-agent managed identity in the post-deploy step. This client-side role assignment is now redundant: the Microsoft Foundry service grants the per-agent identity its required permissions internally.

The redundant assignment also fails noisily when the deploying user lacks `Microsoft.Authorization/roleAssignments/write`, blocking otherwise-successful deploys for users who only hold data-plane roles (e.g. `Foundry User`).

### Additional impact

The `remote.agent-identity-roles` doctor check enumerated ARM role assignments for each agent identity. Since the service no longer creates those assignments, the check folded every agent into a false aggregate failure and reported problems that do not exist.

### Proposed fix

- Remove the client-side per-agent-identity `Azure AI User` role assignment from the post-deploy handler.
- Remove the now-false `remote.agent-identity-roles` doctor check.

This mirrors the equivalent change already made in the Foundry service tooling (microsoft/Skylight#4910).

m5i-work:
## Why

When deploying a hosted agent, the `azure.ai.agents` extension assigned the `Azure AI User` role to each hosted agent's per-agent managed identity in the post-deploy step. That client-side assignment is now redundant: the Microsoft Foundry service grants the per-agent identity its required permissions internally. Worse, it failed noisily when the deploying user lacked `Microsoft.Authorization/roleAssignments/write`, blocking otherwise-successful deploys for users who only hold data-plane roles (for example `Foundry User`).

This mirrors the equivalent change already made in the Foundry service tooling (microsoft/Skylight#4910).

## What changed

- **Post-deploy handler (`listen.go`)**: dropped the block that fetched the agent version, read its instance-identity principal ID, and called `EnsureAgentIdentityRBAC`. The handler still sets up the project endpoint and credential and does optimization reporting.
- **RBAC write path (`agent_identity_rbac.go`)**: removed `EnsureAgentIdentityRBAC` and its helpers. Kept the shared helpers (`parseAgentIdentityInfo`, `assignRoleToIdentity`, `extractSubscriptionID`, `roleAzureAIUser`) that are still used by the separate developer pre-flight RBAC check.
- **Doctor**: removed the `remote.agent-identity-roles` check (and its query module `agent_identity_query.go`). Because the service no longer creates ARM role assignments, that check enumerated nothing and folded every agent into a false aggregate failure. Updated the remote-checks contract test (6 -> 5 checks), the remediation switch, and stale doc comments.

## Scope note

This PR only removes the redundant Foundry `Azure AI User` assignment. ACR role assignment behavior (including ABAC-aware roles) is intentionally left untouched and out of scope.

## Verification

- `go build`, `go vet`, and the full `go test ./...` suite pass for the extension.
- Live end-to-end: deployed the `echo-dual` dual-protocol hosted container agent against an existing Foundry project. The deploy produced no client-side RBAC banner, the agent became active and responded to `azd ai agent invoke` (`responses` protocol), and `azd ai agent doctor` no longer lists `remote.agent-identity-roles`.

Fixes #8940


jongio (approved):
Clean removal of the redundant Azure AI User role assignment from the post-deploy path. The Foundry service handles these permissions internally now, so the client-side write (and the doctor check that enumerated them) are correctly removed.

Verified:
- Shared helpers (parseAgentIdentityInfo, ssignRoleToIdentity, xtractSubscriptionID, 
oleAzureAIUser) are still used by the developer pre-flight RBAC check in developer_rbac_check.go
- listen.go post-deploy flow remains correct: credential is still created for the optimization reporting that follows
- Remote checks contract test correctly updated (6 to 5 checks)
- No dead imports or unreferenced code left behind
- ACR role assignment behavior intentionally untouched (out of scope)

trangevi (approved):

cli/azd/extensions/azure.ai.agents/internal/cmd/listen.go
```diff
@@ -295,13 +294,10 @@ func postdeployHandler(ctx context.Context, azdClient *azdext.AzdClient, args *a
 		return nil
 	}
 
-	// Collect agent identity from the hosted agent service that was deployed.
-	// After deploy, each hosted agent's name/version is stored as AGENT_{SERVICE_KEY}_NAME/VERSION.
-	// We fetch the full agent version object from the API to get the instance identity principal ID,
-	// which allows us to skip the slow Graph API discovery during RBAC assignment.
+	// Set up the project endpoint and credential used by optimization reporting.
```
RickWinter, line 297: Now that agent-identity RBAC is gone, this endpoint/tenant/credential setup exists only to feed reportSvcOptimizationDeployment, which is explicitly best-effort (wrapped in recover, panics logged and not propagated). But the resolution below still returns hard errors, so a missing FOUNDRY_PROJECT_ENDPOINT or AZURE_TENANT_ID now fails the whole postdeploy hook for what is purely optimization telemetry. Before this change that strictness was justified because RBAC depended on it. Could we downgrade these to logged warnings and return nil, so telemetry setup can't fail an otherwise-successful deploy? Not a blocker, but the failure mode is now heavier than the feature warrants.

RickWinter (commented):
This drops the client-side Azure AI User role assignment for per-agent managed identities (the post-deploy step and the doctor remote.agent-identity-roles check), on the basis that Foundry now grants that permission server-side and the client assignment only produced noisy failures for data-plane-only users. The direction is right and the deletion is clean: the shared helpers (parseAgentIdentityInfo, assignRoleToIdentity, extractSubscriptionID, roleAzureAIUser) are correctly retained because developer_rbac_check.go still uses them, endpointResp and cred remain consumed by optimization reporting so nothing goes unused, and the remote-checks contract test is updated from 6 to 5.

One follow-up worth considering: with RBAC gone, the endpoint/tenant/credential resolution in postdeployHandler now feeds only best-effort optimization reporting, yet it still returns hard errors. That is a heavier failure mode than the remaining feature warrants. Not a blocker. The ACR scope note is correct to leave that path untouched.

cli/azd/extensions/azure.ai.agents/internal/cmd/listen.go
m5i-work: Good point! Fixed.

huimiu (approved):

jongio (approved):

PR #8947, opened 2026-07-02
fix: always pass projectName to brownfield Foundry provisioning

### LINKED ISSUE #8946: [Foundry provisioning] Brownfield ARM template unconditionally references the ACR-only project resource, breaking non-ACR (model-deployment-only) reconciles
## Summary

The Foundry brownfield provisioning template (`cli/azd/extensions/azure.ai.agents/internal/synthesis/templates/brownfield.arm.json`, compiled from `brownfield.bicep`) declares the `foundryAccountPreview` account reference and its nested `foundryAccountPreview::project` resource **unconditionally**, even though both exist solely to support the ACR-linking feature (`includeAcr=true`). Because the project resource name is `format('{0}/{1}', accountName, projectName)` and `projectName` was only supplied on the ACR branch, any brownfield deploy that does not create an ACR (for example, a model-deployment-only reconcile) failed ARM template validation with `InvalidTemplate`.

A minimal fix has been applied (always pass `projectName`). This issue tracks the cleaner refactor to remove the ACR-only project reference from non-ACR deploys entirely.

## Background

Provider routing is provider-wide: `Deploy`/`Preview` take the brownfield path whenever `brownfieldEndpoint != ""`, which is set from an `azure.ai.project` service that points at an existing Foundry project via `endpoint:`. The brownfield deploy then reconciles declared model deployments and, only when init flagged `acr` as pending, creates a container registry.

PR #8880 ("Fix: ACR not created/linked for hosted container agents", commit `043e4fe62`) added `foundryAccountPreview` / `foundryAccountPreview::project` to the template to obtain the project's system-assigned managed identity for the ACR connection. Those references are consumed only by the `includeAcr`-conditional resources (`foundryAcrPull`, `acrConnection`). Bicep cannot put a condition on an `existing` resource, so they are declared unconditionally. The same PR gated the `projectName` parameter behind the ACR branch in `brownfieldParams`.

## Impact

On a brownfield deploy that reconciles model deployments but does not create an ACR (`createACR=false`), `projectName` was omitted and defaulted to `''`. The unconditional existing resource then resolved to the invalid name `<account>/` (one segment, two expected), failing the whole deployment before anything was provisioned:

```
ERROR CODE: InvalidTemplate
The template resource '<account>/' of type 'Microsoft.CognitiveServices/accounts/projects'
... has an invalid name. The name '<account>/' contains '1' name segment(s) separated by
'/', but '2' segment(s) were expected.
path: properties.template.resources.foundryAccountPreview::project.type
```

This is surprising because the failing template resource is ACR-related, yet the agent used the code-deploy path (`codeConfiguration.dependencyResolution: remote_build`), which does not need an ACR. The ACR resources were correctly skipped (`includeAcr=false`); only the dead project reference tripped validation.

## Minimal fix (applied)

`brownfieldParams` in `foundry_provisioning_provider.go` now always sets `projectName` (parsed from the project endpoint, with a fallback to the resolved Foundry name), not just on the ACR branch. This makes the unconditional reference valid and inert on non-ACR deploys. A regression test was added in `foundry_provisioning_provider_brownfield_acr_test.go`.

## Proposed cleaner refactor

Remove the unconditional `foundryAccountPreview` / `foundryAccountPreview::project` `existing` declarations and instead resolve the project id and principal id inline within the `includeAcr`-conditional resources, for example:

- id: `resourceId('Microsoft.CognitiveServices/accounts/projects', accountName, projectName)`
- principalId: `reference(resourceId('Microsoft.CognitiveServices/accounts/projects', accountName, projectName), '2025-04-01-preview', 'full').identity.principalId`

Because these expressions live inside `condition: includeAcr` resources, they are evaluated only when an ACR is actually created. On non-ACR deploys the template would contain no project reference at all, so `projectName` becomes genuinely optional and the model-deployment reconcile is fully decoupled from the ACR machinery.

## Acceptance criteria

- Non-ACR brownfield deploys (model-deployment-only reconcile) contain no reference to the project resource in the compiled template.
- ACR brownfield deploys still create the registry, AcrPull role assignment, and ContainerRegistry connection, wired to the existing project's managed identity.
- `brownfield.bicep` recompiled to `brownfield.arm.json`; unit tests updated; the ACR path validated end-to-end (live), since it cannot be fully covered by unit tests.

## Repro

1. An `azure.ai.project` service with an `endpoint:` pointing at an existing Foundry project and at least one model deployment declared.
2. A hosted agent using `codeConfiguration.dependencyResolution: remote_build` (no ACR), or otherwise `AI_AGENT_PENDING_PROVISION` without `acr` and no `AZURE_CONTAINER_REGISTRY_ENDPOINT` set.
3. Run `azd up` (or `azd provision`). Before the minimal fix: `InvalidTemplate` on `Microsoft.CognitiveServices/accounts/projects` with name `<account>/`.

## Notes

Identifiers redacted. Regression introduced in #8880.

therealjohn:
## Summary

Brownfield Foundry provisioning (`azd provision` / `azd up` against an existing Foundry project referenced via a service `endpoint:`) failed with `InvalidTemplate` whenever the deploy reconciled model deployments but did not create a container registry.

The brownfield ARM template (`brownfield.arm.json`, compiled from `brownfield.bicep`) declares `foundryAccountPreview::project` as an **unconditional** `existing` resource named `format('{0}/{1}', accountName, projectName)`. `brownfieldParams` only supplied `projectName` on the ACR branch (`createACR=true`). On a model-deployment-only reconcile (`createACR=false`), `projectName` was omitted and defaulted to `''`, so the resource name collapsed to `<account>/` and ARM rejected the whole deployment before provisioning anything:

```
ERROR CODE: InvalidTemplate
The template resource '<account>/' of type 'Microsoft.CognitiveServices/accounts/projects'
... has an invalid name. The name '<account>/' contains '1' name segment(s) separated by
'/', but '2' segment(s) were expected.
path: properties.template.resources.foundryAccountPreview::project.type
```

This is surprising for users on the code-deploy path (`codeConfiguration.dependencyResolution: remote_build`), which does not need an ACR: the ACR resources are correctly skipped (`includeAcr=false`), but the dead project reference still tripped template validation. Regression introduced in #8880, which added the ACR-only project reference to the template while gating `projectName` behind the ACR branch.

## Change

`brownfieldParams` now always sets `projectName` (parsed from the project endpoint, with a fallback to the resolved Foundry name), not just on the ACR branch. `Deploy` and `Preview` share this function, so both paths are covered. The unconditional project reference is now valid and inert on non-ACR deploys, and no ACR is created on that path.

## Testing

- Added a regression test (`TestBrownfieldParams`, "without ACR still carries projectName for the existing project resource") asserting `projectName` is present on the non-ACR path.
- `go test ./internal/project/...` passes, `go build ./...` succeeds, `gofmt` clean.

## Follow-up

Fixes #8946.

This is the minimal fix. The deeper cleanup - removing the ACR-only `foundryAccountPreview` / `foundryAccountPreview::project` references from non-ACR deploys entirely by inlining them into the `includeAcr`-conditional resources - can be done as a separate follow-up.

### DIFF
--- a/cli/azd/extensions/azure.ai.agents/internal/project/foundry_provisioning_provider.go
+++ b/cli/azd/extensions/azure.ai.agents/internal/project/foundry_provisioning_provider.go
@@ -810,11 +810,15 @@ func (p *FoundryProvisioningProvider) brownfieldParams(
 	params := map[string]any{
 		"accountName": map[string]any{"value": account},
 		"deployments": map[string]any{"value": p.brownfieldDeployments},
+		// projectName feeds the unconditional existing `foundryAccountPreview::project`
+		// resource, so it must always be set -- even on the model-deployments-only
+		// reconcile path. Omitting it collapses the resource name to "<account>/"
+		// and fails ARM template validation with InvalidTemplate.
+		"projectName": map[string]any{"value": p.brownfieldProjectName()},
 	}
 	if createACR {
 		params["includeAcr"] = map[string]any{"value": true}
 		params["acrName"] = map[string]any{"value": p.brownfieldACRName(account)}
-		params["projectName"] = map[string]any{"value": p.brownfieldProjectName()}
 		params["tags"] = map[string]any{"value": map[string]string{"azd-env-name": p.envName}}
 		// Only set location when resolved; an empty value would override the
 		// template default (resourceGroup().location) and fail the deployment.
--- a/cli/azd/extensions/azure.ai.agents/internal/project/foundry_provisioning_provider_brownfield_acr_test.go
+++ b/cli/azd/extensions/azure.ai.agents/internal/project/foundry_provisioning_provider_brownfield_acr_test.go
@@ -172,13 +172,23 @@ func TestBrownfieldParams(t *testing.T) {
 
 	deployments := []synthesis.Deployment{{Name: "gpt-4o-mini"}}
 
-	t.Run("without ACR carries only account and deployments", func(t *testing.T) {
+	t.Run("without ACR still carries projectName for the existing project resource", func(t *testing.T) {
 		t.Parallel()
-		p := &FoundryProvisioningProvider{envName: "dev", brownfieldDeployments: deployments}
+		// The brownfield template declares `foundryAccountPreview::project` as an
+		// unconditional existing resource, so projectName must be supplied even
+		// when no ACR is created (model-deployments-only reconcile). Regression
+		// test for the InvalidTemplate failure where the name collapsed to
+		// "<account>/" because projectName was omitted.
+		p := &FoundryProvisioningProvider{
+			envName:               "dev",
+			brownfieldEndpoint:    "https://acct.services.ai.azure.com/api/projects/my-project",
+			brownfieldDeployments: deployments,
+		}
 		params := p.brownfieldParams(t.Context(), "acct", "rg", false)
 
 		assert.Equal(t, map[string]any{"value": "acct"}, params["accountName"])
 		assert.Equal(t, map[string]any{"value": deployments}, params["deployments"])
+		assert.Equal(t, map[string]any{"value": "my-project"}, params["projectName"])
 		assert.NotContains(t, params, "includeAcr")
 		assert.NotContains(t, params, "acrName")
 	})


jongio (approved):
Correct fix. Moving projectName out of the createACR conditional matches the ARM template's unconditional oundryAccountPreview::project reference. The regression test covers the exact failure path.

cli/azd/extensions/azure.ai.agents/internal/project/foundry_provisioning_provider.go
```diff
@@ -810,11 +810,15 @@ func (p *FoundryProvisioningProvider) brownfieldParams(
 	params := map[string]any{
 		"accountName": map[string]any{"value": account},
 		"deployments": map[string]any{"value": p.brownfieldDeployments},
+		// projectName feeds the unconditional existing `foundryAccountPreview::project`
+		// resource, so it must always be set -- even on the model-deployments-only
+		// reconcile path. Omitting it collapses the resource name to "<account>/"
+		// and fails ARM template validation with InvalidTemplate.
+		"projectName": map[string]any{"value": p.brownfieldProjectName()},
```
RickWinter, line 817: Now that this feeds the unconditional project reference, it is worth being sure brownfieldProjectName() can't return empty here. It parses the endpoint and falls back to p.foundryName, so if the endpoint has no project segment and foundryName is unset we would emit projectName: "" and hit the same InvalidTemplate this is fixing. If foundryName is always resolved before brownfieldParams runs this is a non-issue; otherwise a small guard, or asserting the both-empty case in the test, would make the invariant explicit.

RickWinter (commented):
This always sets projectName in brownfieldParams so the unconditional existing foundryAccountPreview::project reference in the brownfield template stays valid on the model-deployments-only reconcile, which is the right fix for the InvalidTemplate regression from #8880. The approach is the correct shape: the value moves into the base params map and is dropped from the ACR branch, so both Deploy and Preview are covered without duplication, and the renamed regression test asserts projectName on the non-ACR path.

One thing worth confirming before merge: brownfieldProjectName() parses the project from the endpoint and falls back to foundryName, so if the endpoint has no parseable project segment and foundryName is unset, projectName is still empty and we land back in the same InvalidTemplate failure. If that state is unreachable because foundryName is always resolved before brownfieldParams runs, this is fine to leave as is. If it is reachable, a small guard (or asserting the both-empty case) would close the loop.

Nothing here blocks.

trangevi (approved):
Approved, pending response to Rick's concern

PR #8949, opened 2026-07-02
Exclude dynamic linker/loader (LD_/DYLD_) variables from tool environments

### LINKED ISSUE #8953: azd propagates dynamic linker/loader (LD_/DYLD_) variables from .env into external tools
### Description

When `azd` runs external tools (docker, npm, python, maven, go, dotnet, bicep, kubectl, `azd exec`, hooks, and extensions), it builds the tool's environment from the current process environment plus the values stored in the azd environment's `.env` file (`.azure/<env>/.env`).

Today that includes environment variable names in the platform dynamic linker/loader namespace — those beginning with `LD_` (Linux `ld.so`) or `DYLD_` (macOS `dyld`), such as `LD_PRELOAD`, `LD_LIBRARY_PATH`, or `DYLD_INSERT_LIBRARIES`. These variables change how a process resolves and loads shared libraries at startup, and a project's `.env` should not influence that for the tools `azd` invokes.

### Expected behavior

`azd` should not carry `LD_`/`DYLD_`-prefixed variables from the environment's `.env` file into the environments of the tools it runs. Names outside that reserved namespace (for example `LDFLAGS`, `LDLIBS`) should be unaffected, and matching should be case-insensitive.

### Notes

- Most tool paths build their environment via `Environment.Environ()`.
- A few paths build it directly from `Environment.Dotenv()` (`azd exec`, the AKS/kubectl sync, and the .NET container app Bicep template context), so they need the same treatment.

hemarina:
Fixes #8953

## Summary

`azd` builds the environment it hands to the external tools it runs (docker, npm, python, maven, go, dotnet, bicep, kubectl, `azd exec`, hooks) — and the values it serves to extensions over the gRPC bridge — from the values stored in the azd environment's `.env` file.

This change excludes environment variable names in the platform dynamic linker/loader namespace — those beginning with `LD_` (Linux `ld.so`) or `DYLD_` (macOS `dyld`) — from those values. These variables control how a process resolves and loads shared libraries at startup, and a project's `.env` should not influence that for the tools `azd` (or an azd extension) invokes. Names outside that namespace (for example `LDFLAGS`, `LDLIBS`) are unaffected. Matching is case-insensitive.

Filtering happens once in `Environment.Dotenv()`, the accessor that produces the value copy every tool-environment and extension-facing path is built from: `Environ()` derives from it, and the gRPC `GetValues` handler returns it. The values remain in the persisted `.env` file; only the copy azd hands to callers is filtered.

## Changes

- `pkg/environment/environment.go`: exclude `LD_`/`DYLD_`-prefixed keys in `Dotenv()`; `Environ()` now derives from `Dotenv()`.
- `pkg/environment/environment_test.go`: tests for `Dotenv()` and `Environ()` — case-insensitivity, additional `DYLD_*` names, and preservation of `LDFLAGS`/`LDLIBS` and ordinary values.
- `.vscode/cspell.yaml`, `CHANGELOG.md`: dictionary terms and a note for the behavior change.

## Testing

- `go test ./pkg/environment/`, plus the affected `internal/grpcserver`, `cmd`, and `pkg/project` tests, pass.

### DIFF
--- a/cli/azd/pkg/environment/environment.go
+++ b/cli/azd/pkg/environment/environment.go
@@ -221,10 +221,25 @@ func (e *Environment) DotenvDelete(key string) {
 }
 
 // Dotenv returns a copy of the key value pairs from the .env file in the environment.
+//
+// Dynamic linker/loader control variables (names in the reserved LD_/DYLD_ namespace, e.g.
+// LD_PRELOAD or DYLD_INSERT_LIBRARIES) are excluded, because these values flow into the
+// environments of the external tools azd runs (directly, and via extensions that read them over
+// the gRPC bridge), where the platform dynamic loader would honor them at process launch. They
+// remain in the persisted .env file; only what azd hands to callers here is filtered. Matching is
+// case-insensitive; names outside that namespace (e.g. LDFLAGS, LDLIBS) are unaffected.
 func (e *Environment) Dotenv() map[string]string {
 	e.mu.RLock()
 	defer e.mu.RUnlock()
-	return maps.Clone(e.dotenv)
+	myDotenv := make(map[string]string, len(e.dotenv))
+	for k, v := range e.dotenv {
+		upper := strings.ToUpper(k)
+		if strings.HasPrefix(upper, "LD_") || strings.HasPrefix(upper, "DYLD_") {
+			continue
+		}
+		myDotenv[k] = v
+	}
+	return myDotenv
 }
 
 // DotenvSet sets the value of [key] to [value] in the .env file associated with the environment. [Save] should be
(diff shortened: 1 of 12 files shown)


cli/azd/pkg/environment/environment.go
```diff
@@ -307,13 +307,34 @@ func (e *Environment) SetServiceProperty(serviceName string, propertyName string
...
+// attacker-controlled code be loaded into a subprocess at launch time. These variables have no
+// legitimate reason to live in an azd dotenv, so they are stripped at the single source that
+// feeds every subprocess environment.
+var unsafeEnvironKeys = map[string]struct{}{
+	"LD_PRELOAD":            {},
+	"LD_LIBRARY_PATH":       {},
+	"LD_AUDIT":              {},
+	"DYLD_INSERT_LIBRARIES": {},
+	"DYLD_LIBRARY_PATH":     {},
+	"DYLD_FRAMEWORK_PATH":   {},
```
jongio: The bot's comment above noted `cmd/exec.go` and `service_target_aks.go` as bypass paths. There's a third: `service_target_dotnet_containerapp.go:332` does `maps.Copy(env, at.env.Dotenv())` to populate the Bicep template execution context, which also skips this filter.

A small shared helper (`func FilterUnsafeEnvKeys(m map[string]string) map[string]string`) that both `Environ()` and these call sites could use would close the gap cleanly. Not blocking this PR on it, but worth a follow-up.

jongio (approved):
The filtering logic is correct: case-insensitive denylist at `Environ()` covers most subprocess paths (all framework services, hooks, extensions, bicep).

One additional `Dotenv()` bypass path not mentioned in the existing bot comment: `pkg/project/service_target_dotnet_containerapp.go:332` does `maps.Copy(env, at.env.Dotenv())` to build the template execution environment, which also skips this filter. Consider extracting a shared `FilterUnsafeKeys` helper so all subprocess-facing paths can reuse the same denylist, or at minimum soften the "single source" claim in the doc comment to acknowledge the known gaps.

Minor CI note: cspell-lint is failing, likely needs a word added to the dictionary.

jongio (approved):
The prefix-based approach (LD_/DYLD_ matching via isLoaderControlKey) is more robust than the prior explicit denylist and covers future loader variables. The shared FilterLoaderControlKeys helper closes the Dotenv() bypass paths I flagged earlier. Tests verify the boundary correctly (LDFLAGS/LDLIBS preserved, case-insensitive matching works).

jongio (approved):
The prefix-based filtering moved into `Dotenv()` itself closes the extension/gRPC path gap. `Environ()` correctly delegates without double-locking. Tests verify case-insensitivity and the LDFLAGS/LDLIBS boundary at the right layer.

JeffreyCA (approved):

peter-noora (commented):
This fix (released yesterday with azd 1.27.1) broke my build because we have an environment variable named `LD_SDK_KEY`, which happens to relate to our usage of LaunchDarkly.
Ignoring _any_ variable prefixed `LD_` has, as I've experienced, a pretty high risk of breaking setups out there.

PR #8952, opened 2026-07-02
feat: route azure.yaml templates through Foundry adoption flow

### LINKED ISSUE #8951: [Agents Extension] Interactive template picker won't support unified azure.yaml
Running `azd ai agent init` start you off with an interactive flow to pick a template from an list in awesome-azd. We need to update that list to point to the new unified azure.yaml samples, however doing so would break the interactive flow. It currently would treat the new pointers as azd templates and attempt to download the whole repo (which thankfully currently fails).

trangevi:
## Motivation

When the template picker selects a unified `azure.yaml` template (one with `templateType: "extension.ai.agent"` whose source URL points to an `azure.yaml` file), the previous `TemplateTypeAzd` code path attempted to `git clone` the source URL. This fails because the URL is a blob path inside a repository, not a cloneable repo.

## Approach

Introduce a new `TemplateTypeAzureYaml` constant to distinguish unified azure.yaml templates from full azd template repositories. Route these templates through the existing Foundry adoption flow (`runInitFromAzureYaml`) which correctly downloads the azure.yaml and its sibling files, stages them locally, and scaffolds the project.

Key changes:

- **New constant `TemplateTypeAzureYaml`** -- keeps `TemplateTypeAzd` intact for actual azd repo templates
- **Updated `EffectiveType()`** -- returns `TemplateTypeAzureYaml` when both the source ends with `azure.yaml`/`azure.yml` AND `templateType == "extension.ai.agent"`
- **New `TemplateTypeAzureYaml` case in init** -- downloads content, resolves agent name via `resolveInitAgentName` (interactive prompt or `--agent-name` flag), derives the project folder, then calls `runInitFromAzureYaml`
- **`sample_list.go`** -- added to known types and DTO mapping (uses `ManifestURL` + `azd ai agent init -m` command)
- **Unit tests** -- 5 new `TestEffectiveType` cases covering the new type and fallback behavior

## Notes

- The `TemplateTypeAzd` case (for full cloneable repo templates) is preserved unchanged
- Agent name resolution mirrors the agent manifest template flow: azure.yaml `name:` field -> template title -> interactive prompt

### DIFF
--- a/cli/azd/extensions/azure.ai.agents/internal/cmd/init.go
+++ b/cli/azd/extensions/azure.ai.agents/internal/cmd/init.go
@@ -1309,95 +1309,50 @@ from code-deploy ZIP packaging (uses .gitignore syntax).`,
 					}
 
 					switch selectedTemplate.EffectiveType() {
-					case TemplateTypeAzd:
-						// Full azd template - dispatch azd init -t <repo>
-						// Create project in a new subdirectory derived from the template title.
-						folderName := folderNameStrippingParenSuffix(selectedTemplate.Title)
-						// Check whether the target directory already exists so we
-						// only report "created" when a new directory was made.
-						_, statErr := os.Stat(folderName)
-						newlyCreated := errors.Is(statErr, fs.ErrNotExist)
-						initArgs := []string{"init", "-t", selectedTemplate.Source, folderName}
-						if flags.env != "" {
-							initArgs = append(initArgs, "--environment", flags.env)
-						} else {
-							base := sanitizeAgentName(folderName)
-							if len(base) > 59 {
-								base = strings.TrimRight(base[:59], "-")
-							}
-							defaultEnvName := base + "-dev"
-							initArgs = append(
-								initArgs, "--environment", defaultEnvName,
-							)
-						}
-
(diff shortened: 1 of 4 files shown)


jongio (commented):
The new `TemplateTypeAzureYaml` routing logic is sound: `EffectiveType()` correctly requires both the `extension.ai.agent` discriminator and an azure.yaml/yml suffix, and the downstream calls to `readManifestContentForInitDetection` / `resolveInitAgentName` / `runInitFromAzureYaml` compose cleanly with the existing Foundry adoption flow.

One question on intent: the PR description states "The `TemplateTypeAzd` case (for full cloneable repo templates) is preserved unchanged," but the diff replaces it entirely with the new `TemplateTypeAzureYaml` case. Templates whose `EffectiveType()` still resolves to `TemplateTypeAzd` (repo-slug sources that don't end in any yaml) now fall through to `default`, which treats the source as an agent manifest pointer. Is dropping the `azd init -t` dispatch from the interactive picker intentional? If so, the description and `fixtureTemplates()` test fixture (which includes a `TemplateTypeAzd` entry) should be updated to reflect this.

huimiu (approved):

PR #7834, opened 2026-04-21
Implement FlexibleTimestamp for modified_time in JSON

therealjohn:
This pull request improves the handling of file modification timestamps in the agent API by introducing a new `FlexibleTimestamp` type. This allows the system to robustly parse and serialize timestamps provided either as strings or as Unix epoch numbers (in seconds or milliseconds), addressing inconsistencies in upstream API responses. The changes also update all relevant usages and add comprehensive tests for the new logic.

**Flexible timestamp handling:**

* Introduced the `FlexibleTimestamp` type in `models.go`, which can unmarshal JSON timestamps provided as strings or numbers (seconds or milliseconds), and always serializes as an RFC3339 string. Includes custom `UnmarshalJSON`, `MarshalJSON`, and `String` methods.
* Updated the `SessionFileInfo` struct to use `*FlexibleTimestamp` for the `LastModified` field instead of `*string`.

**Code and test updates:**

* Refactored all usages of `LastModified` in the CLI and tests (`files.go`, `files_test.go`, `models_test.go`) to use the new `FlexibleTimestamp` type and its `.String()` method where needed. [[1]](diffhunk://#diff-7ab85165356870dbb74d3bd7b502b5ecbeeff3a911f04379757858c575ec8a29L504-R504) [[2]](diffhunk://#diff-7ab85165356870dbb74d3bd7b502b5ecbeeff3a911f04379757858c575ec8a29L801-R801) [[3]](diffhunk://#diff-87ef4e7908ea08f8cb39d8c9389907ef57378e90f32093624054d374e688065dL146-R146) [[4]](diffhunk://#diff-87ef4e7908ea08f8cb39d8c9389907ef57378e90f32093624054d374e688065dL170-R170) [[5]](diffhunk://#diff-73785b4ba3b189ec5ab67837873172dff2a582e6bcdefef54355f1ca1e250bd7L814-R814) [[6]](diffhunk://#diff-73785b4ba3b189ec5ab67837873172dff2a582e6bcdefef54355f1ca1e250bd7L843-R904)
* Added extensive tests for `FlexibleTimestamp` covering string, numeric (seconds), and numeric (milliseconds) formats to ensure correct parsing and round-trip serialization.

**Dependency and import updates:**

* Added imports for `encoding/json`, `fmt`, `math`, and `time` in `models.go` to support the new timestamp logic.

Fixes #7802

PR #8986, opened 2026-07-06
fix(routines): decode numeric timestamps in routine responses

### LINKED ISSUE #8984: azure.ai.routines: routine response decoding fails when the service returns numeric values
## Summary

Several `azd ai routine` commands fail to decode routine responses when the service returns numeric values.

Observed numeric fields:

- `created_at` on schedule routines
- `triggers.<name>.at` on timer routines

## Example commands

```bash
azd ai routine create my-schedule-routine --file schedule-routine.yaml -p <project-endpoint> --output json
azd ai routine show my-schedule-routine -p <project-endpoint> --output json
```

## Error

```text
ERROR: <operation>: failed to unmarshal response: json: cannot unmarshal number into Go struct field <field> of type string
```

## Affected commands

Confirmed affected:

- `azd ai routine create`
- `azd ai routine show`
- `azd ai routine update`
- `azd ai routine enable`
- `azd ai routine disable`
- `azd ai routine dispatch --input`

glharper:
Fixes #8984

## What

Several `azd ai routine` commands failed to decode routine responses when the Foundry Routines service returned **numeric** timestamp values:

```
ERROR: <operation>: failed to unmarshal response: json: cannot unmarshal number into Go struct field <field> of type string
```

Observed on `created_at` (schedule routines) and `triggers.<name>.at` (timer routines). This broke `azd ai routine create`, `show`, `update`, `enable`, `disable`, and `dispatch --input` (which fetches the routine to resolve its action type).

## Root cause

The data-plane response models (`internal/pkg/routines/models.go`) typed these timestamp fields as `string`, but the service returns them as JSON numbers (Unix epoch), which `encoding/json` cannot unmarshal into a `string`.

## Fix

Introduce a `FlexibleTimestamp` type (modeled on the existing `FlexibleTimestamp` in `azure.ai.agents`, PR #7834) that:

- decodes from a **JSON string** (preserved verbatim), or a **JSON number** — Unix epoch seconds, or milliseconds when the seconds interpretation lands implausibly far in the future — normalized to RFC 3339;
- always **marshals back to a JSON string**, so create/update request bodies keep the exact wire shape they have today (`at` is still sent as a string).

Applied to the routine and routine-run timestamp fields: `created_at`, `updated_at`, timer `at`, `triggered_at`, `scheduled_fire_at`, `started_at`, `ended_at`. The two production input sites (`--at` on create and manifest update) convert the user-supplied string into the new type; display (`%s`) and `!= ""` checks are unaffected.

## Tests

- New unit tests for `FlexibleTimestamp`: string decode, epoch-seconds decode, epoch-milliseconds decode, string marshaling.
- Regression test `TestRoutine_NumericTimestampsDecode`: a routine response with numeric `created_at` and timer `at` now decodes cleanly (both normalize to `2025-01-01T00:00:00Z`).
- Updated existing assertions to compare via `.String()`.

## Validation


jongio (approved):
Solid fix. The FlexibleTimestamp approach correctly mirrors the established pattern in azure.ai.agents, handles the seconds/milliseconds disambiguation cleanly, and the test coverage is thorough.

glharper:

jongio (approved):
Re-approving after the main merge. The PR's code is unchanged from my earlier review. The FlexibleTimestamp approach is correct, test coverage is thorough, and the fix properly handles both seconds and milliseconds disambiguation.

huimiu (approved):

PR #8987, opened 2026-07-06
fix(agents): prompt for agent name from template.name on init -m

### LINKED ISSUE #8945: [Agents extension] azd ai agent init -m <azure.yaml> doesn't properly prompt for agent name
The interactive path prompts for agent name, which is then used for the directory which the agent files are downloaded into. We should do the same for the -m path.

glharper:
Fixes #8945

## What

`azd ai agent init -m <manifest>` did not prompt for the agent name the way the interactive path does. The interactive/template flows prompt for the agent name and use it for the project folder the agent files are downloaded into; the `-m` path did not reliably do the same.

## Root cause

Before the full manifest is loaded, the `-m` flow calls `resolveAgentNameFromManifestPointer` → `peekManifestName` to seed the agent-name prompt default (and the project folder name). `peekManifestName` read only the **top-level `name`** field.

But an agent **manifest** nests the agent identity under `template.name`; the top-level `name` is the *manifest* name. The authoritative downstream resolver (`ExtractAgentDefinition`) uses `template.name`, falling back to the top-level name. So for a normal manifest the peek would either:

- return an empty string (when the manifest had no top-level `name`) → the prompt was **skipped** entirely and the inner flow silently used `template.name`; or
- return the *manifest* name → the prompt/folder default **diverged** from the agent name used for the service entry and folder.

## Fix

`peekManifestName` now prefers `template.name` and falls back to the top-level `name`, mirroring `ExtractAgentDefinition`'s precedence. The `-m` flow now prompts with — and defaults the folder to — the agent's own name, consistent with the interactive and template flows. Bare agent definitions (flat `name`, no `template:`) are unchanged.

## Tests

- `TestPeekManifestName_LocalFile_PrefersTemplateName` — a manifest with `template.name` is peeked as the agent name.
- `TestPeekManifestName_LocalFile_TemplateWithoutNameFallsBack` — a manifest whose template omits `name` falls back to the top-level name.
- Existing peek tests (bare agent.yaml, nested list-item names ignored, malformed/empty/missing) remain green.

## Validation

`go build ./...`, `go test ./internal/cmd/...`, `go vet`, `gofmt`, `golangci-lint run` (0 issues), and `cspell` — all clean.


jongio (approved):
The fix correctly aligns peekManifestName with ExtractAgentDefinition's template.name precedence. Both new and existing tests verify the logic.

RickWinter (commented):
This aligns the `-m` (manifest) init path with the interactive and template flows by teaching `peekManifestName` to prefer `template.name` over the top-level `name`. The approach is the right shape: the peek now mirrors the precedence that `ExtractAgentDefinition` (via the seed-from-template, fall-back-to-top-level logic in parse.go) already uses downstream, so the prompt default and project folder no longer diverge from the agent identity that ends up in the service entry.

I traced the fix against parse.go and it matches: template identity wins, top-level `name` is the fallback for bare agent definitions, and the two new tests exercise both branches. The existing peek tests (bare agent.yaml, nested list items ignored, malformed/empty) still cover the unchanged paths.

Nothing blocking and no inline comments. This is a contained, correct bug fix that can merge as-is.

glharper:

jongio (approved):
Code unchanged since my prior approval. Re-approving after the main merge.

huimiu (approved):

PR #8989, opened 2026-07-06
feat(agents): support invoke over the a2a protocol

### LINKED ISSUE #8956: Support for a2a protocal

glharper:
Closes #8956

## What

Adds **A2A (Agent2Agent) protocol** support to `azd ai agent invoke`, alongside the existing `responses` and `invocations` protocols. A2A was previously deployment-only and excluded from invocation.

```bash
# Invoke a deployed agent over A2A (wraps the message as a JSON-RPC message/send)
azd ai agent invoke --protocol a2a "Hello!"

# Send a complete JSON-RPC request from a file
azd ai agent invoke --protocol a2a -f a2a-request.json

# Dump the raw JSON-RPC response
azd ai agent invoke --protocol a2a --output raw "Hello!"
```

## Changes

- **Make `a2a` invocable** — added to `InvocableProtocols()` / `IsInvocable()`. A2A is **remote-only**, gated by a new `IsLocalInvocable()`; `activity_protocol` remains fully deployment-only.
- **Routing** — `--protocol a2a` and `agent.yaml` protocol auto-detection route to a new `a2aRemote` handler (new file `invoke_a2a.go`), modeled on the invocations handler (session-based memory, `--new-session` resets, `--new-conversation` no-op). `--local a2a` fails fast with actionable guidance.
- **Request body** — a plain message is wrapped in a JSON-RPC 2.0 `message/send` request with a single user text part; `--input-file` contents are forwarded verbatim as a complete JSON-RPC request.
- **Response** — the reply text is extracted from the JSON-RPC `result` across A2A result shapes (Message `parts`, Task `status.message` / `artifacts[]` / `history[]`), a JSON-RPC `error` surfaces as an agent error, and `--output raw` dumps the response verbatim.
- **Endpoint plumbing** — added `buildA2AInvokeURL` and `--agent-endpoint` parsing for the `.../endpoint/protocols/a2a` tail; updated `--protocol` help, the unsupported-protocol / local-protocol validation messages, the multi-protocol guidance, and command examples.

## Tests

- `models_test.go`: `a2a` is now `IsInvocable()==true` and `IsLocalInvocable()==false`.
- `agent_endpoint_test.go`: parse + `buildA2AInvokeURL` (api-version/session-id escaping, default api-version).
- `helpers_test.go`: single `a2a` auto-detects; a non-invocable case retained via `activity_protocol`.
- `invoke_a2a_test.go`: JSON-RPC request wrapping, file passthrough, reply-text extraction across Message/Task shapes, JSON-RPC error detection, `--local` rejection, request-URL fallback.

## Validation

## ⚠️ Note for reviewers — Foundry A2A wire format

The **A2A protocol itself is a public open standard** (JSON-RPC 2.0 `message/send`, https://a2a-protocol.org), so the request/response envelope is well-defined. What is **not** publicly specced is the **Foundry hosted-agent A2A invoke endpoint path**. This PR follows the codebase's established `responses`/`invocations` convention: `{projectEndpoint}/agents/{name}/endpoint/protocols/a2a?api-version=...[&agent_session_id=...]`.

These assumptions are **centralized** in `buildA2AInvokeURL`, `buildA2ARequestBody`, and `handleA2AResponse`, so if the Foundry contract differs they can be corrected in one place, and `--output raw` is a format-independent escape hatch. **Please confirm the exact Foundry A2A endpoint path (and whether it expects the vanilla JSON-RPC `message/send` body) before merge.** A2A local invocation is intentionally not supported (A2A is a remote agent-to-agent protocol).


jongio (approved):
Reviewed the full changeset (10 files, +734/-15). The A2A handler follows the same session-based pattern as invocations, URL construction and request wrapping are centralized for easy adjustment if the Foundry contract changes, and test coverage spans all A2A result shapes (Message parts, Task status/artifacts/history, JSON-RPC errors, local rejection). No issues found.

cli/azd/extensions/azure.ai.agents/internal/cmd/invoke_a2a.go
```diff
@@ -0,0 +1,377 @@
...
+		}
+		if kind != "" && kind != "text" {
+			// Explicitly a non-text part (file, data, ...).
+			continue
+		}
+		if text, ok := m["text"].(string); ok && strings.TrimSpace(text) != "" {
+			texts = append(texts, text)
+		}
+	}
+	return strings.Join(texts, "")
```
RickWinter: nit: joining text parts with an empty string means a Task or Message that splits its reply across multiple text parts (for example distinct paragraphs) comes out concatenated with no separator, while the artifacts branch just above joins with `\n`. If Foundry ever returns paragraph-per-part replies this will read as one run-on block. Matching the artifacts separator here would make the two paths consistent:

```suggestion
	return strings.Join(texts, "\n")
```

Not blocking, and if the intent is that parts within a single message are always contiguous fragments, leaving it as-is is defensible.

RickWinter (commented):
This makes `a2a` an invocable protocol for `azd ai agent invoke`, remote-only, routed to a new `a2aRemote` handler that wraps a plain message in a JSON-RPC 2.0 `message/send` request (or forwards `--input-file` verbatim) and extracts the reply from the JSON-RPC result across Message and Task shapes. The approach is the right shape: it follows the existing responses/invocations conventions (session-bound memory, `--new-session` reset, `--new-conversation` no-op, `--output raw` passthrough), and the new `IsLocalInvocable()` split cleanly expresses that A2A is deployable and remotely invocable but has no local surface.

I checked the parts that depend on code outside the diff. The `--local a2a` rejection is covered twice, once as early flag validation in RunE when `--protocol` is explicit, and once as the runtime `a2aLocal` handler for the agent.yaml auto-detected path, so there is no gap. `captureResponseSession` reads only the `x-agent-session-id` header, so calling it before `handleA2AResponse` does not consume the body. `buildA2AInvokeURL` matches `buildInvocationsURL` (query escaping on api-version and session id, agent name left unescaped consistent with the sibling builders since it is pattern-validated upstream).

The one open item is external, not code: the Foundry hosted-agent A2A endpoint path and body contract are unconfirmed. You have already centralized those assumptions in `buildA2AInvokeURL` / `buildA2ARequestBody` / `handleA2AResponse` and left `--output raw` as an escape hatch, which is the correct containment. Worth confirming the wire contract with the Foundry team before merge, but that does not block the code review.

One nit inline on multi-part text joining, otherwise this is ready. Disposition: mergeable pending the external endpoint confirmation; nothing here requires a code change before merge.

glharper:

jongio (approved):
Re-approving after the merge of main and the one-line fix to use newline as the text-parts separator in `a2aTextFromParts`. The change makes multi-part A2A responses more readable. No issues.

jongio (approved):
Re-approving. The only change since my last review is a test comment rename (TestBuildA2AURL -> TestBuildA2AInvokeURL) to match the actual function name. No logic changes.

cli/azd/extensions/azure.ai.agents/internal/cmd/invoke_a2a.go
```diff
@@ -0,0 +1,377 @@
...
+	}
+
+	a2aURL := buildA2AInvokeURL(rc.projectEndpoint, rc.name, rc.apiVersion, sid)
+
+	req, err := http.NewRequestWithContext(ctx, http.MethodPost, a2aURL, bytes.NewReader(body))
+	if err != nil {
+		return fmt.Errorf("failed to create request: %w", err)
+	}
+	req.Header.Set("Content-Type", "application/json")
+	req.Header.Set("Authorization", "Bearer "+rc.bearerToken)
```
huimiu: Could we set `A2A-Version` with the other headers, or derive it from the selected protocol version? I might be missing something, but since this body uses the v0.3 JSON-RPC shape (`message/send`) and the A2A spec treats an omitted version as a 0.3 compatibility default, I'm wondering if a Foundry endpoint on a newer contract might reject or parse this differently.

glharper: updated in 6c1ea5e78. I now set the A2A-Version header explicitly (confirmed against the A2A spec §3.2, which returns VersionNotSupportedError when unsupported) so a Foundry endpoint on a newer contract negotiates deterministically instead of falling back to the implicit 0.3 default.

I made it a constant (a2aProtocolVersion = "0.3") tied to the request shape we actually emit (the v0.3 JSON-RPC message/send envelope) rather than deriving it from the agent's advertised protocols[].version: the header must describe the body we send, and the advertised version could be a Foundry-internal value that doesn't match the wire shape — which is exactly the mismatch that would cause misparsing. If/when we emit a newer body shape, the constant moves with it. Extracted applyA2ARequestHeaders with a unit test asserting the header. Still centralized for easy correction if the Foundry A2A contract differs.

jongio (approved):
Reviewed the new commit (6c1ea5e). The explicit `A2A-Version` header addition is correct per spec section 3.2, and centralizing header setup in `applyA2ARequestHeaders` keeps the call site clean. Test covers all three headers. No issues.

trangevi (approved):

huimiu (approved):

PR #8998, opened 2026-07-06
ci: use 1ES pipelineArtifact output in ext-azure-ai-agents-live

### LINKED ISSUE #9004: ext-azure-ai-agents-live pipeline fails: PublishPipelineArtifact@1 not allowed under 1ES template
## Problem

Running the `ext-azure-ai-agents-live` pipeline fails during template expansion:

```
/v1/Steps/UserProvidedBuildSteps.yml@1ESPipelineTemplates (Line: 89, Col: 7): Unexpected value 'PublishPipelineArtifact@1 task is not allowed. Please use `output: pipelineArtifact` instead. See https://aka.ms/1espt/outputs for more details'
```

## Cause

`eng/pipelines/ext-azure-ai-agents-live.yml` extends `1es-redirect.yml`, so it runs through the **1ES pipeline template**, which forbids calling `PublishPipelineArtifact@1` as a plain `task:` step. 1ES requires artifacts to be declared as **job-level outputs** via `templateContext.outputs` using `output: pipelineArtifact` (the pattern already used by other 1ES pipelines in the repo, e.g. `eng/pipelines/templates/jobs/build-cli.yml`).

## Fix

Move the test-log publish out of `steps:` into a `templateContext.outputs` block on the `Tier2` job, preserving `condition: always()` so logs still publish on failure.

vhvb1989:
Fixes #9004

## Problem

Running the `ext-azure-ai-agents-live` pipeline fails at template expansion:

```
/v1/Steps/UserProvidedBuildSteps.yml@1ESPipelineTemplates (Line: 89, Col: 7): Unexpected value 'PublishPipelineArtifact@1 task is not allowed. Please use `output: pipelineArtifact` instead. See https://aka.ms/1espt/outputs for more details'
```

## Cause

`eng/pipelines/ext-azure-ai-agents-live.yml` extends `1es-redirect.yml`, so it runs through the **1ES pipeline template**, which forbids calling `PublishPipelineArtifact@1` as a plain `task:` step. 1ES requires artifacts to be declared as **job-level outputs** via `templateContext.outputs` using `output: pipelineArtifact` (the same pattern already used by every other 1ES pipeline in the repo, e.g. `eng/pipelines/templates/jobs/build-cli.yml`).

## Fix

Move the test-log publish out of `steps:` and into a `templateContext.outputs` block on the `Tier2` job. `condition` is still supported, so `always()` continues to publish logs even when the live run fails. Key mappings: `targetPath` → `path`, `artifactName` → `artifact`.

No behavior change beyond compliance — logs are published to the same artifact name.

### DIFF
--- a/cli/azd/extensions/azure.ai.agents/tests/e2e-live/tier2_live_test.go
+++ b/cli/azd/extensions/azure.ai.agents/tests/e2e-live/tier2_live_test.go
@@ -343,8 +343,8 @@ func (r *runner) finishInit(ctx context.Context) error {
 		return nil
 	}
 	return fmt.Errorf(
-		"init finished but expected artifacts are missing on disk\n--- tail ---\n%s",
-		tail(r.c.tailString(), 2000),
+		"init finished but expected artifacts are missing on disk\n--- test dir ---\n%s\n--- tail ---\n%s",
+		r.initOutputDiagnostics(), tail(r.c.tailString(), 2000),
 	)
 }
 
@@ -760,11 +760,14 @@ func (r *runner) findServiceName() string {
 	}
 	// A struct unmarshal is more robust than scanning lines: it tolerates
 	// comments and indentation changes that a naive parser would mishandle.
-	var proj struct {
-		Services map[string]any `yaml:"services"`
-	}
+	var proj azdProjectManifest
 	if err := yaml.Unmarshal(data, &proj); err != nil || len(proj.Services) == 0 {
 		return ""
+	}
+	for name, svc := range proj.Services {
+		if svc.Host == "azure.ai.agent" {
+			return name
+		}
 	}
 	for name := range proj.Services {
 		return name
@@ -772,8 +775,18 @@ func (r *runner) findServiceName() string {
 	return ""
 }
 
-// validateInitOutput confirms init produced an agent project on disk: a project
-// dir whose azure.yaml targets the agent host and a nested agent.yaml.
+type azdProjectManifest struct {
+	Services map[string]azdService `yaml:"services"`
+}
+
+type azdService struct {
+	Host string `yaml:"host"`
+}
+
+// validateInitOutput confirms init produced an azd project on disk. Current
+// unified azure.yaml samples can inline the agent definition in azure.yaml
+// instead of writing a separate agent.yaml, so the stable contract is that init
+// created a project directory with a parseable azure.yaml containing services.
 func (r *runner) validateInitOutput() bool {
 	entries, err := os.ReadDir(r.testDir)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


jongio (approved):

eng/pipelines/ext-azure-ai-agents-live.yml
```diff
@@ -69,6 +69,17 @@ extends:
...
             timeoutInMinutes: 150
+            # The 1ES pipeline template (via 1es-redirect.yml) forbids the raw
+            # PublishPipelineArtifact@1 task; artifacts must be declared as
+            # job-level outputs using `output: pipelineArtifact`. `condition`
+            # is supported here, so `always()` still publishes logs on failure.
+            templateContext:
+              outputs:
+                - output: pipelineArtifact
+                  path: $(Build.ArtifactStagingDirectory)
+                  artifact: tier2-live-logs-$(Build.BuildId)
```
danieljurek: `$(Build.BuildId)` will not be unique to the build. If you want to avoid collisions use `$(System.JobAttempt)`

danieljurek (approved):
Builds won't succeed on retry with the current naming convention, but this will function as written

eng/pipelines/ext-azure-ai-agents-live.yml
vhvb1989: Good catch, thanks. Addressed in 8b27931 — the artifact name is now `tier2-live-logs-$(Build.BuildId)-$(System.JobAttempt)`, so re-running a failed job in the same build publishes to a distinct name. Kept `Build.BuildId` too so the artifact stays traceable to a specific build.

jongio (approved):
Re-approving after the JobAttempt fix. The `$(System.JobAttempt)` suffix correctly prevents artifact name collisions across job re-runs within the same build.

eng/pipelines/ext-azure-ai-agents-live.yml
```diff
@@ -69,6 +69,20 @@ extends:
...
+            # job-level outputs using `output: pipelineArtifact`. `condition`
+            # is supported here, so `always()` still publishes logs on failure.
+            templateContext:
+              outputs:
+                - output: pipelineArtifact
+                  path: $(Build.ArtifactStagingDirectory)
+                  # Include JobAttempt so re-running a failed job in the same
+                  # build publishes to a distinct artifact name (BuildId alone
+                  # collides across attempts).
+                  artifact: tier2-live-logs-$(Build.BuildId)-$(System.JobAttempt)
```
RickWinter, line 83: nit: adding `$(System.JobAttempt)` here is a real improvement, since re-running a failed job in the same build no longer collides on the artifact name. The PR description says logs publish to 'the same artifact name' though, which is no longer literally true. Worth correcting the description so the rename is not a surprise to anyone diffing artifacts across attempts.

RickWinter (commented):
This moves the tier2 live-log publish out of a raw PublishPipelineArtifact@1 step and into a job-level templateContext.outputs block so the pipeline passes 1ES template validation. The approach is correct and matches the pattern the other 1ES pipelines in the repo already use. condition: always() is preserved, so failure-path log publishing is retained, and targetPath/artifactName map cleanly to path/artifact.

The one behavior delta is the artifact name gaining $(System.JobAttempt), which is an improvement (re-running a failed job no longer collides on the artifact name), not a regression. No blockers. Disposition: COMMENT.

hemarina (approved):

jongio (approved):
Test infrastructure updates look solid. The switch from string matching to proper YAML struct parsing in `hasProjectManifest` is the right call for handling the unified azure.yaml format. `initOutputDiagnostics` is well-bounded (80 entries, depth 3) for CI log clarity.

jongio (approved):
The new commit correctly adds build and install steps for the `azure.ai.projects` extension, which agent samples now depend on (`host: azure.ai.project`). The yq `load()` approach for multi-extension config generation is clean, and the `providers` field inclusion matches the extension struct schema.

cli/azd/extensions/azure.ai.agents/tests/e2e-live/tier2_live_test.go
```diff
@@ -784,20 +797,75 @@ func (r *runner) validateInitOutput() bool {
...
+		fmt.Fprintf(&b, "walk failed: %v\n", err)
+	}
+	if count >= maxEntries {
+		fmt.Fprintf(&b, "... truncated after %d entries\n", maxEntries)
+	}
+	return strings.TrimSpace(b.String())
+}
+
 // hasAgentYAML reports whether an agent.yaml exists anywhere under root.
 func hasAgentYAML(root string) bool {
```
RickWinter: validateInitOutput switched to hasProjectManifest and no longer calls hasAgentYAML, so this helper is now orphaned (its only reference is its own definition). This file is //go:build linux, so on the CI linux build the unused linter will flag it as dead code. Let's delete hasAgentYAML now that nothing references it.

RickWinter (commented):
This moves the tier2 live-log publish from a raw PublishPipelineArtifact@1 task into a templateContext.outputs pipelineArtifact block, which is what the 1ES template requires, and adds System.JobAttempt to the artifact name so a rerun of a failed job does not collide with the first attempt. That pipeline change is the right shape and matches the pattern already used in build-cli.yml, and condition: always() is preserved so failure-path logs still publish.

The one thing to reconcile before merge is scope. The title and description cover only the pipeline fix (#9004), but roughly half the diff rewrites tier2_live_test.go validation semantics: validateInitOutput now accepts an agent inlined in azure.yaml instead of requiring a separate agent.yaml, plus new initOutputDiagnostics and host-based service selection. Those changes are reasonable, but they are not mentioned anywhere, so either fold a line into the description or split them into their own PR to keep the history honest. One concrete consequence of that rewrite is flagged inline (an orphaned helper).

Nothing here blocks. Fix the dead code and either document or split the test changes.

jongio (approved):

jongio (approved):

jongio (approved):

hemarina (approved):

jongio (approved):

trangevi (approved):

PR #9003, opened 2026-07-06
[feat] optimzie progressive update + adapt to new azure.yaml

### LINKED ISSUE #9010: azd ai agent optimize progressive update + adapting to new inline azure.yaml
1. optimize progressive update in output
2.  adapting to new inline azure.yaml
3. support -o in optimize status command

Zyysurely:
### Summary

Improves the `azd ai agent optimize` polling experience with live-updating candidate rows and phase-aware progress indicators, migrates agent kind detection to the new azure.yaml inline/config agent definition format, and adds `--output json` support to `optimize status`.

#### Agent detection from azure.yaml (`eval.go`, `optimize.go`)

- **Replaced `detectEvalAgentKind`** with `projectpkg.LoadAgentDefinition`, which supports the latest azure.yaml agent definition shapes:
  - **Inline**: agent kind defined directly in azure.yaml service config
  - **Config reference**: agent kind referenced via `config` pointer in azure.yaml
  - **Legacy disk**: standalone `agent.yaml` file on disk
- The agent kind source is now accurately reported (`azure.yaml (inline)`, `azure.yaml (config)`, `agent.yaml`) instead of just showing the file path.
- Emits a deprecation warning via `WarnLegacyAgentShape` when legacy config shapes are detected.
- Removed now-unused `detectEvalAgentKind` and `relPathForYaml` functions and their tests.

#### Optimize agent resolution refactor (`optimize.go`)

- **Resolution order fixed**: `resolveOptimizeAgent` now tries the azd project context first (passing `--agent` as a service name to `resolveAgentService`), then falls back to treating `--agent` as a raw Foundry agent name. Previously, a non-empty `--agent` flag would skip project resolution entirely.
- The `--agent` flag now also reads the agent **version** from the environment (via `AGENT_{KEY}_VERSION`), enabling version-aware optimization.
- Updated flag help text to clarify: `"Agent service name from azure.yaml, or Foundry agent name outside a project"`.
- Same resolution fix applied to `optimize apply` and `optimize deploy` flag descriptions.

#### Progressive polling UX (`optimize.go`, `optimize_helpers.go`)

- **Live candidate table**: Completed candidates are printed as individual table rows as they finish, instead of waiting for the entire job to complete.
- **Phase-aware status line**: The spinner distinguishes "Generating" vs "Evaluating" using the new `InProgressCandidate.CandidateGenerated` field.
- **Candidate progress counter**: Shows `Generating candidate 2/5` (from `MaxCandidates`) or `Generating candidate 2` when max is unknown.
- **Server-side elapsed time**: Uses `InProgressCandidate.CreatedAt` (server timestamp) instead of a locally-tracked timer.
- **Extracted table rendering helpers**: `candidateDisplayName`, `candidateTableHeader`, `formatCandidateRow`, `writeCandidateRow` are now shared helpers, eliminating duplicated rendering between polling and final results.

#### `optimize status --output json` (`optimize_status.go`)

- Adds `--output` flag (`json` | `table`, default `table`) via `azdext.RegisterFlagOptions`.
- `--output json`: prints the full `OptimizeJobStatus` as indented JSON.
- `--output json --watch`: suppresses table output during polling, prints final status as JSON.
- Informational "Using last job: ..." line suppressed in JSON mode.

#### API model updates (`optimize_api/models.go`)

- Added `InProgressCandidate` struct with `CreatedAt` and `CandidateGenerated` fields.
- Added `InProgressCandidate` field to `JobProgress`.

### Tests

- `optimize_status_test.go`: `--output` flag registration, default poll interval, `printOptimizeStatusJSON`, job summary duration display (present/missing/equal timestamps).
- `optimize_helpers_test.go`: `candidateDisplayName`, `candidateTableHeader`, `formatCandidateRow` (empty/non-empty eval & strategy, hidden columns).
- `optimize_test.go`: `TestPrintOptimizeResults_StrategyDashForMixedMutations`.
- Removed `TestDetectEvalAgentKind`, `TestRelPathForYaml` (functions deleted).
- Renamed `TestOptimizeDeployCommand_AgentResolvedFromFlagOrYaml` → `…FromFlagOrProject`.

### DIFF
--- a/cli/azd/extensions/azure.ai.agents/internal/cmd/optimize.go
+++ b/cli/azd/extensions/azure.ai.agents/internal/cmd/optimize.go
@@ -16,7 +16,6 @@ import (
 	"io"
 	"os"
 	"path/filepath"
-	"strings"
 	"time"
 
 	"azureaiagent/internal/pkg/agents/eval_api"
@@ -39,26 +38,24 @@ type optimizeAgentContext struct {
 
 // resolveOptimizeAgent resolves the agent name and project directory.
 // Resolution order:
-//  1. Explicit --agent flag
-//  2. azd project context (resolveAgentService + environment variables)
-//  3. Error with guidance
+//  1. azd project context — if --agent is given it is matched as a service
+//     name; otherwise the single agent service is auto-selected (or the user
+//     is prompted). The Foundry agent name is read from the environment.
+//  2. No azd project — --agent is treated as the Foundry agent name directly.
+//  3. Error with guidance.
 func resolveOptimizeAgent(ctx context.Context, flagValue, envName string, noPrompt bool) (*optimizeAgentContext, error) {
-	if flagValue != "" {
-		return &optimizeAgentContext{agentName: flagValue}, nil
-	}
-
-	// Try resolving from azd project — single resolveAgentService call
-	// to get both project path and agent info from environment.
+	// Try resolving from azd project first — --agent (if set) is interpreted
+	// as a service name, consistent with show/delete/run.
 	azdClient, err := azdext.NewAzdClient()
 	if err == nil {
 		defer azdClient.Close()
 
-		svc, project, svcErr := resolveAgentService(ctx, azdClient, "", noPrompt)
+		svc, project, svcErr := resolveAgentService(ctx, azdClient, flagValue, noPrompt)
 		if svcErr == nil && svc != nil && project != nil {
 			agentProject := filepath.Join(project.Path, svc.RelativePath)
 			serviceKey := toServiceKey(svc.Name)
 
-			// Read agent name and version from azd environment
+			// Read agent name and version from azd environment.
 			if env := getExistingEnvironment(ctx, envName, azdClient); env != nil {
 				nameKey := fmt.Sprintf("AGENT_%s_NAME", serviceKey)
 				if v, e := azdClient.Environment().GetValue(ctx, &azdext.GetEnvRequest{
@@ -82,6 +79,12 @@ func resolveOptimizeAgent(ctx context.Context, flagValue, envName string, noProm
 				}
 			}
 		}
+	}
+
+	// Outside an azd project (or project resolution failed): treat --agent as
+	// the Foundry agent name directly, matching the eval without-project path.
+	if flagValue != "" {
+		return &optimizeAgentContext{agentName: flagValue}, nil
 	}
 
 	return nil, fmt.Errorf("agent name is required: use --agent <name>, or run from an azd project after 'azd deploy'")
@@ -146,7 +149,7 @@ Use --config for a custom YAML spec, or just provide the agent name to use sensi
 	}
 
 	cmd.Flags().StringVarP(&flags.configFile, "config", "c", "", "Path to YAML config file (optional — values are prompted interactively if omitted)")
-	cmd.Flags().StringVarP(&flags.agent, "agent", "a", "", "Agent name (auto-detected from azd project if omitted)")
+	cmd.Flags().StringVarP(&flags.agent, "agent", "a", "", "Agent service name from azure.yaml, or Foundry agent name outside a project")
 	cmd.Flags().StringVarP(&flags.dataset, "dataset", "d", "", "Existing local file or registered dataset name")
 	cmd.Flags().StringVarP(&flags.evalModel, "eval-model", "m", "", "Model for evaluation (required)")
 	cmd.Flags().StringVar(&flags.optimizationModel, "optimize-model", "",
@@ -155,7 +158,7 @@ Use --config for a custom YAML spec, or just provide the agent name to use sensi
 		"Built-in or custom evaluator name (repeatable; required when not set in config)")
 	cmd.Flags().IntVar(&flags.maxCandidates, "max-candidates", 0, "Maximum number of optimization candidates to generate (must be >= 1; default: 5)")
 	cmd.Flags().BoolVar(&flags.noWait, "no-wait", false, "Submit job and return immediately without waiting for completion")
-	cmd.Flags().IntVar(&flags.pollInterval, "poll-interval", 5, "Polling interval in seconds")
+	cmd.Flags().IntVar(&flags.pollInterval, "poll-interval", 10, "Polling interval in seconds")
 	flags.optimizeConnectionFlags.register(cmd)
 
 	cmd.AddCommand(newOptimizeStatusCommand(extCtx))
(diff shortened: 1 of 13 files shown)


cli/azd/extensions/azure.ai.agents/internal/cmd/optimize.go
```diff
@@ -528,45 +530,135 @@ func (a *OptimizeAction) submitJob(
...
+			bestName := ""
+			if best := status.BestCandidate(); best != nil {
+				bestName = best.Name
+			}
+
+			// Append rows for newly completed candidates. Eval URLs are
+			// constructed per-candidate on the spot using the lazily
+			// initialized azdClient.
+			for i := printedCandidates; i < len(candidates); i++ {
+				c := candidates[i]
```
jongio, line 609: Optional: `InProgressCandidate.CreatedAt` is a server-reported timestamp for when the current candidate started. Using it instead of the locally-tracked `lastCandidateTime` would give a more accurate elapsed display, especially if there's clock drift or if a previous candidate took unusually long.

```suggestion
				if status.Progress != nil && status.Progress.InProgressCandidate != nil {
					statusLine = fmt.Sprintf("  %s Generating candidates", spin)
					if status.Progress.InProgressCandidate.CreatedAt > 0 {
						candElapsed := time.Since(time.Unix(status.Progress.InProgressCandidate.CreatedAt, 0)).Truncate(time.Second)
						statusLine += fmt.Sprintf(" \u00b7 %s elapsed", candElapsed)
					} else {
						candElapsed := time.Since(lastCandidateTime).Truncate(time.Second)
						statusLine += fmt.Sprintf(" \u00b7 %s elapsed", candElapsed)
					}
```

jongio (approved):
Clean refactoring that consolidates agent kind detection into the shared `LoadAgentDefinition` utility and improves the optimize polling UX with a live-updating candidate table.

Key things I verified:
- `loadOptimizeJobIDForAgent("", envName)` correctly falls back to the global `OPTIMIZE_LAST_OPERATION_ID` key, matching the removed `loadLastOptimizeJobID` behavior.
- The `resolveOptimizeAgent` reordering (project-first, flag-as-Foundry-name fallback) is consistent with other commands and preserves backwards compatibility for users specifying a Foundry name directly.
- New table rendering helpers are well-tested and the extraction from the inline loop is a good separation of concerns.

One optional suggestion inline.

cli/azd/extensions/azure.ai.agents/internal/cmd/optimize.go
Zyysurely: addressed

Zyysurely: addressed

jongio (commented):
Incremental review of the new commit (1760b79, "support -o").

The JSON output support and the improved progress status line both look solid. One logic gap in the terminal-job path when --watch is combined with --output json.

cli/azd/extensions/azure.ai.agents/internal/cmd/optimize_status.go
```diff
@@ -88,15 +100,24 @@ func runOptimizeStatus(cmd *cobra.Command, flags *optimizeStatusFlags, operation
...
+		finalStatus, err := pollOptimizeJob(cmd, client, flags.pollInterval, operationID, flags.envName)
 		if err != nil {
 			return err
 		}
+		if flags.output == "json" {
+			return printOptimizeStatusJSON(out, finalStatus)
+		}
 		printOptimizeResults(cmd.Context(), out, finalStatus, hasProject, flags.envName)
 	} else if len(status.Candidates()) > 0 {
 		printOptimizeResults(cmd.Context(), out, status, hasProject, flags.envName)
```
jongio, line 126: When `--output json --watch` is used and the job is already terminal (completed before polling starts), this branch runs and prints table format instead of JSON.

The watch branch above correctly checks `flags.output == "json"` before falling through to `printOptimizeResults`, but this else branch doesn't. A user scripting with `--output json --watch` would get unparseable table output if the job finishes before the command runs.

```suggestion
	} else if len(status.Candidates()) > 0 {
		if flags.output == "json" {
			return printOptimizeStatusJSON(out, status)
		}
		printOptimizeResults(cmd.Context(), out, status, hasProject, flags.envName)
```

trangevi (commented):
This looks good overall — just a couple of comments.

cli/azd/extensions/azure.ai.agents/internal/cmd/optimize.go
```diff
@@ -84,6 +81,12 @@ func resolveOptimizeAgent(ctx context.Context, flagValue, envName string, noProm
 		}
```
trangevi, line 90: `resolveOptimizeAgent` silently falls through when inside an azd project if `resolveAgentService` succeeds but `AGENT_<SVC>_NAME` is not set in the environment (e.g., user hasn't deployed yet). If `--agent myservice` was provided as a service name, it then gets used as a Foundry agent name at line 86 — which is likely not what the user intended.

I think we need either: (a) an explicit error like "agent resolved as service 'X' but no deployed agent name found — run `azd deploy` first", or (b) at minimum a log/warning so the user knows project resolution was skipped.

*(Review provided by azd-code-reviewer, be sure to validate before making changes)*

cli/azd/extensions/azure.ai.agents/internal/cmd/optimize_test.go
```diff
@@ -515,3 +515,37 @@ func TestPrintOptimizeResults_NoStrategyColumnWhenNoMutations(t *testing.T) {
...
+
+	var buf strings.Builder
+	printOptimizeResults(t.Context(), &buf, status, false, "")
+	out := buf.String()
+
+	assert.Contains(t, out, "Strategy")
+	assert.Contains(t, out, "system_prompt")
+	// The baseline row (no mutations) should contain a dash placeholder.
+	lines := strings.SplitSeq(out, "\n")
+	for line := range lines {
```
trangevi, line 547: The `for line := range lines` loop with a conditional assertion inside never fails if no line matches both `strings.Contains(line, "baseline")` and `strings.Contains(line, "0.800")`. If the baseline row format changes or disappears, this test passes vacuously. Adding a `found` bool asserted after the loop would catch that regression.

*(Review provided by azd-code-reviewer, be sure to validate before making changes)*

cli/azd/extensions/azure.ai.agents/internal/cmd/optimize.go
Zyysurely: addressed

cli/azd/extensions/azure.ai.agents/internal/cmd/optimize_test.go
Zyysurely: addressed

cli/azd/extensions/azure.ai.agents/internal/cmd/optimize_status.go
Zyysurely: addressed

jongio (approved):

trangevi (approved):

PR #9009, opened 2026-07-07
Support invocations_ws in agent init

### LINKED ISSUE #9173: Support invocations_ws hosted agent initialization
Track support for initializing and deploying hosted agents that use the invocations_ws protocol through the azure.ai.agents azd extension.

v1212:
## Summary
- Adds `invocations_ws` as a selectable protocol for `azd ai agent init` while keeping the default protocol as `responses`.
- Honors `--protocol invocations_ws` for bring-your-own-image `--image` init.
- Documents `invocations_ws` in agent help/schema descriptions.
- Adds focused regression coverage for protocol selection, image init protocol propagation, and code deploy serialization via `protocol_versions`.

## Sample used for validation
Official sample:
https://github.com/microsoft-foundry/foundry-samples/tree/main/samples/python/hosted-agents/bring-your-own/invocations_ws/hello-world

## Validation
- `go test ./internal/cmd/...` from `cli/azd/extensions/azure.ai.agents`
- `go test ./...` from `cli/azd/extensions/azure.ai.agents`
- `go build ./...` from `cli/azd/extensions/azure.ai.agents`
- `cspell lint "extensions/**/*.go" "extensions/**/*.md" --config ./.vscode/cspell.yaml --no-progress` from `cli/azd`

Fixes #9173


cli/azd/extensions/azure.ai.agents/internal/cmd/init_adopt.go
```diff
@@ -829,6 +954,40 @@ func runInitFromAzureYaml(
...
+		serviceName, err := foundryProjectServiceForModelOverride(content)
+		if err != nil {
+			return err
+		}
+		deployment, err := resolveDeploymentForModelFlag(ctx, azdClient, azureContext, flags.model)
+		if err != nil {
+			return fmt.Errorf("failed to resolve model deployment for %q: %w", flags.model, err)
+		}
+		if deployment != nil {
+			if err := updateAzureYamlDeployments(ctx, azdClient, serviceName, []project.Deployment{*deployment}); err != nil {
```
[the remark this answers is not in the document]
v1212: This PR intentionally does not rewrite sample-specific environment variables such as AZURE_VOICELIVE_MODEL. The generic init path now provisions/selects the model deployment and persists AZURE_AI_MODEL_DEPLOYMENT_NAME; samples that need a private runtime env var should keep that env var aligned in the sample azure.yaml or add a future schema-level mapping rather than coupling azd init to a Voice Live-specific variable name.

cli/azd/extensions/azure.ai.agents/internal/cmd/init_adopt.go
```diff
@@ -136,6 +136,159 @@ func foundryDeployments(content []byte) []foundryDeploymentEntry {
...
+}
+
+func updateAzureYamlAgentName(ctx context.Context, azdClient *azdext.AzdClient, serviceName, agentName string) error {
+	val, err := structpb.NewValue(agentName)
+	if err != nil {
+		return fmt.Errorf("encoding agent name for service %q: %w", serviceName, err)
+	}
+	if _, err := azdClient.Project().SetServiceConfigValue(ctx, &azdext.SetServiceConfigValueRequest{
+		ServiceName: serviceName,
+		Path:        "name",
```
RickWinter: This ignores supported config-nested agent services: writing root name and protocols leaves config.name and config.protocols in effect because the loader still selects config when it contains kind. Use the matching config.* path, as adoptedAgentNameConfig already does.

RickWinter (commented):
This adds invocations_ws support and extends unified sample adoption with protocol, model, and Azure-context overrides. Legacy config-nested agent services still ignore the agent-name and protocol overrides, so that must be addressed before merge.

cli/azd/extensions/azure.ai.agents/internal/cmd/init_adopt.go
v1212: Thanks, fixed to update config.name/config.protocols for config-nested agent services, while preserving inline kind precedence to match runtime resolution.

jongio (approved):
Scope here is clean and the invocations_ws wiring lines up: knownProtocols entry, the image-path protocolRecordsForImageManifest/resolveKnownProtocols helpers (default, dedup, activity rejection, unknown-protocol error), and the schema/help/region updates. Test coverage across the flag, interactive, image-manifest, and code-deploy serialization paths is thorough, and the default protocol staying responses is locked down by its own test.

One non-blocking follow-up, already noted by the bot: an invocations_ws-only project still gets an zd ai agent invoke --local next-step hint even though that protocol isn't invocable. Pre-existing next-step behavior, fine to handle separately.

jongio (approved):
Scope here is clean and the invocations_ws wiring lines up: knownProtocols entry, the image-path protocolRecordsForImageManifest/resolveKnownProtocols helpers (default, dedup, activity rejection, unknown-protocol error), and the schema/help/region updates. Test coverage across the flag, interactive, image-manifest, and code-deploy serialization paths is thorough, and the default protocol staying responses is locked down by its own test.

One non-blocking follow-up, already noted by the bot: an invocations_ws-only project still gets an "azd ai agent invoke --local" next-step hint even though that protocol is not invocable. Pre-existing next-step behavior, fine to handle separately.

jongio (commented):
On the new commit: `loadServiceProtocol` now early-returns on `invocations_ws`, so when a service declares both `invocations_ws` and `responses` the classification depends on their order. With `invocations_ws` first it is treated as websocket-only and drops the invoke hint; with `responses` first it is treated as `responses`. The function's own doc comment says `responses` wins when multiple protocols are declared, and `invocations` is deliberately handled as flag-and-continue to hold that invariant. Treating `invocations_ws` the same way (using it only when no azd-invocable protocol is declared) keeps the classification order-independent. A reversed-order test case would lock it in. This is already flagged inline; noting it is still worth addressing before merge.

cli/azd/extensions/azure.ai.agents/internal/cmd/init_adopt.go
```diff
@@ -136,6 +136,189 @@ func foundryDeployments(content []byte) []foundryDeploymentEntry {
...
+		return field
+	}
+	config, ok := service["config"].(map[string]any)
+	if !ok {
+		return field
+	}
+	if _, ok := config["kind"]; !ok {
+		return field
+	}
+	return "config." + field
```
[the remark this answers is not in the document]
v1212: Fixed in the latest push: invocations_ws is now tracked as non-invocable but lower precedence than responses/invocations, so mixed-protocol manifests keep the existing invocable next-step behavior regardless of order. Added reversed-order tests for responses+invocations_ws and invocations+invocations_ws.

cli/azd/extensions/azure.ai.agents/internal/cmd/nextstep/resolver.go
```diff
@@ -770,6 +773,9 @@ func appendInvokeLocalSecondary(
 	if len(state.Services) == 1 {
 		svc = &state.Services[0]
 	}
+	if svc != nil && svc.Protocol == ProtocolInvocationsWS {
+		return out, priority
+	}
```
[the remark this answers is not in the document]
v1212: This is already handled in the latest branch revision. ResolveAfterRun returns no invoke suggestion for ProtocolInvocationsWS, and ResolveAfterDeploy skips invoke suggestions for websocket services while preserving show commands. Added regression coverage for init/run/deploy next-step behavior.

jongio (approved):
The order-dependence I flagged in loadServiceProtocol is resolved: invocations_ws now sets a flag instead of early-returning, so precedence stays responses > invocations > invocations_ws no matter what order the manifest declares them. The reversed-order test cases lock that in.

serviceSupportsAzdInvoke is applied consistently across ResolveAfterRun, ResolveAfterDeploy, and appendInvokeLocalSecondary, and the run/deploy suppression paths (single websocket agent and mixed multi-agent) have coverage.

cli/azd/extensions/azure.ai.agents/internal/cmd/nextstep/resolver.go
```diff
@@ -675,6 +684,10 @@ func ResolveAfterDeploy(
 	return out
 }
 
+func serviceSupportsAzdInvoke(svc *ServiceState) bool {
+	return svc == nil || svc.Protocol != ProtocolInvocationsWS
```
trangevi: How does this handle agents with multiple protocols?

cli/azd/extensions/azure.ai.agents/internal/cmd/init.go
```diff
@@ -657,31 +658,88 @@ func preBuiltImageForInit(agentManifest *agent_yaml.AgentManifest, flagImage str
...
 }
 
+func protocolRecordsForImageManifest(flagProtocols []string) ([]agent_yaml.ProtocolVersionRecord, error) {
+	if len(flagProtocols) == 0 {
+		return []agent_yaml.ProtocolVersionRecord{{Protocol: "responses", Version: "2.0.0"}}, nil
+	}
+	if slices.Contains(flagProtocols, "activity") {
+		return nil, exterrors.Validation(
+			exterrors.CodeInvalidAgentManifest,
+			"--protocol activity is not supported with --image",
```
trangevi: Why is the activity protocol not supported here?

trangevi: This is also confusing because if the failure on line 699 in `resolveKnownProtocols` is raised, it will include the activity protocol in that list of known protocols

cli/azd/extensions/azure.ai.agents/internal/cmd/init.go
```diff
@@ -1481,7 +1538,8 @@ from code-deploy ZIP packaging (uses .gitignore syntax).`,
 		"Directory to download the agent definition to (defaults to 'src/<agent-id>')")
 
 	cmd.Flags().StringSliceVar(&flags.protocols, "protocol", nil,
-		"Protocols supported by the agent (e.g., 'responses', 'invocations'). Can be specified multiple times.")
+		"Protocols supported by the agent (e.g., 'responses', 'invocations', 'invocations_ws'). "+
```
trangevi: Similar question, is there a reason the activity protocol isn't listed? Should this list be dynamic, so we don't miss more?

trangevi (changes requested):

jongio (commented):
Circling back after the latest review activity. The two invocations_ws findings from the bot both check out, and they trace to one root cause: serviceSupportsAzdInvoke returns true for a nil service, so the "no specific service" case defaults to invocable.

1. Run flow prints a dangling header for websocket agents. emitNextAfterBind (run.go:307) unconditionally prints "Agent ready. In another terminal, try:" and then calls ResolveAfterRun, which now early-returns nil for invocations_ws (resolver.go:336). printNextIfTerminal no-ops on the empty slice, so an invocations_ws local run ends with that header and no command under it. run.go isn't touched in this PR, but the early-return added here is what surfaces it. Resolving the suggestions first and skipping the header when the list is empty would fix it.

2. Multi-agent, all-websocket still suggests invoke --local. In appendInvokeLocalSecondary, multiple services leave svc nil, so serviceSupportsAzdInvoke(nil) is true and the invoke --local hint is appended even when every service is invocations_ws. Gating on "at least one service is invocable" for the multi-service case covers it.

Both are on the unchanged HEAD I approved earlier, so I'm flagging them here rather than letting that approval imply the websocket paths are fully covered. trangevi's activity-protocol questions on init.go are separate and worth a direct reply from the author.

jongio (commented):
Heads up: this new commit is failing CI on all three platforms (Linux/Mac/Windows ext build).

The cause is `TestSynthesizeImageManifestFile_AcceptsActivityProtocol` in `init_test.go`. It calls `cleanup()` before `os.ReadFile(manifestPath)`, and `cleanup` runs `os.RemoveAll(tmpDir)`, so the read hits a deleted file and the test fails every run:

```go
require.NoError(t, err)
cleanup()

content, err := os.ReadFile(manifestPath) // tmpDir already gone
require.NoError(t, err)
```

Move the `cleanup()` call after the manifest assertions, or `defer cleanup()` right after the `NoError` check. That's the one thing blocking merge in this commit from what I can see; the activity-protocol handling itself (non-invocable in `serviceSupportsAzdInvoke`, and the `invocations > activity > invocations_ws` precedence in `loadServiceProtocol`) matches the tests you added.

cli/azd/extensions/azure.ai.agents/internal/cmd/nextstep/resolver.go
v1212: Updated. Multiple protocols now use a stable priority: responses > invocations > non-invocable protocols (activity, invocations_ws). So if an agent declares both invocations_ws and responses, responses wins regardless of order and invoke hints are preserved. Added reversed-order regression tests.

cli/azd/extensions/azure.ai.agents/internal/cmd/init.go
v1212: Updated. Removed the special rejection for activity in the --image protocol path. protocolRecordsForImageManifest now treats activity like the other known protocols, and the image manifest test now verifies activity is accepted.

cli/azd/extensions/azure.ai.agents/internal/cmd/init.go
v1212: Updated. The --protocol help now uses knownProtocolNames(), so the help text stays in sync with the actual knownProtocols list and includes activity automatically.

cli/azd/extensions/azure.ai.agents/internal/cmd/init_test.go
```diff
@@ -343,6 +343,65 @@ func TestSynthesizeImageManifestFile(t *testing.T) {
...
+func TestSynthesizeImageManifestFile_AcceptsActivityProtocol(t *testing.T) {
+	t.Parallel()
+
+	manifestPath, cleanup, err := synthesizeImageManifestFile(
+		"my-agent",
+		"myacr.azurecr.io/agents/my-agent:v1",
+		[]string{"activity"},
+	)
+	require.NoError(t, err)
+	cleanup()
```
jongio: `cleanup()` deletes `tmpDir` here, but `manifestPath` is `filepath.Join(tmpDir, "agent.yaml")`, so `os.ReadFile(manifestPath)` on the next line fails with file-not-found and the test can never pass. That is what is turning the BuildAndTest jobs red. Defer it so it runs after the assertions:

```suggestion
	defer cleanup()
```

jongio (commented):
Heads up that my earlier approval no longer holds: the new commit breaks the extension build. All three `azure.ai.agents` BuildAndTest jobs (Linux, Mac, Windows) are failing on `TestSynthesizeImageManifestFile_AcceptsActivityProtocol`.

`cleanup()` runs `os.RemoveAll(tmpDir)` and `manifestPath` lives inside `tmpDir`, so the `os.ReadFile(manifestPath)` right after it always hits file-not-found and the following `require.NoError` fails. Deferring cleanup so it runs after the assertions fixes it.

The rest of the commit is consistent with the precedence invariant: `loadServiceProtocol` keeps `activity` as flag-and-continue (responses > invocations > activity > invocations_ws, order-independent), `serviceSupportsAzdInvoke` suppresses the invoke hint for it, and the `--protocol` help text is now driven by `knownProtocolNames()`.

cli/azd/extensions/azure.ai.agents/internal/cmd/init_test.go
v1212: Fixed, thanks. cleanup is now deferred until after the manifest assertions, and the extension tests pass locally.

cli/azd/extensions/azure.ai.agents/internal/cmd/nextstep/state.go
```diff
@@ -478,17 +478,29 @@ func loadServiceProtocol(projectPath, relativePath string) string {
...
+	sawInvocationsWS := false
 	for _, p := range hosted.Protocols {
 		switch strings.TrimSpace(p.Protocol) {
 		case ProtocolResponses:
 			return ProtocolResponses
+		case ProtocolInvocationsWS:
+			sawInvocationsWS = true
 		case ProtocolInvocations:
 			sawInvocations = true
+		case ProtocolActivity:
```
jongio: This switch only matches the canonical `activity` spelling, but the extension still accepts the legacy `activity_protocol` value on read: `agent_api.IsActivityProtocolName` treats both as Activity, and `agent_api/models.go` keeps `AgentProtocolActivityProtocolLegacy = "activity_protocol"` for back-compat. A manifest declaring `activity_protocol` falls through here, so `loadServiceProtocol` returns empty and the resolver falls back to `responses` - a deployment-only agent then gets an `azd ai agent invoke` hint that can't work. Add a case for the legacy spelling (or key both off a shared constant) so both map to `ProtocolActivity`.

jongio (commented):
One follow-up on next-step protocol handling for legacy Activity manifests, inline.

cli/azd/extensions/azure.ai.agents/internal/cmd/nextstep/state.go
v1212: Fixed in the latest push. loadServiceProtocol now treats both activity and activity_protocol as ProtocolActivity, and serviceSupportsAzdInvoke suppresses invoke hints for both spellings. Added legacy activity_protocol regression tests.

jongio (approved):
Re-reviewed the four commits since my last pass, and my earlier notes are resolved. The image manifest test now defers `cleanup()` so it reads the file after the assertions, which was what turned BuildAndTest red. Next-step resolution treats both `activity` and the legacy `activity_protocol` spelling as non-invocable across the init, run, and deploy resolvers, so no unusable `azd ai agent invoke` hints are emitted. Protocol detection now reads the inline `azure.yaml` service config before falling back to the on-disk `agent.yaml`, which covers the image flow that stores the definition inline. Local build and the `internal/cmd/...` test suite pass, and CI is green.

jongio (approved):
Re-reviewed the changes since my last approval (the inline-config protocol loading, activity handling, and invoke-hint suppression).

loadServiceProtocol now checks the inline service config before falling back to agent.yaml on disk, and loadServiceProtocolFromBytes keeps the precedence order-independent: responses > invocations > activity > invocations_ws. serviceSupportsAzdInvoke consistently suppresses the local/deploy invoke hints for invocations_ws, activity, and activity_protocol across ResolveAfterRun, ResolveAfterDeploy, and appendInvokeLocalSecondary, with the mixed multi-agent case keeping invocable services. The image-manifest path now accepts activity and the --protocol help is driven by knownProtocolNames(), so the earlier questions about activity support and a static help list are handled.

Build and the nextstep, cmd, and agent_yaml test packages all pass locally. Looks good to me.

cli/azd/extensions/azure.ai.agents/internal/cmd/nextstep/state.go
```diff
@@ -472,23 +518,39 @@ func loadServiceProtocol(projectPath, relativePath string) string {
...
 			return ProtocolResponses
+		case ProtocolInvocationsWS:
+			sawInvocationsWS = true
 		case ProtocolInvocations:
 			sawInvocations = true
+		case ProtocolActivity, ProtocolActivityLegacy:
+			sawActivity = true
 		}
 	}
 	if sawInvocations {
```
jongio, line 553: For a definition that declares `responses` plus `invocations_ws`, this selects `responses` and `serviceSupportsAzdInvoke` returns true, so the resolver emits an unqualified `azd ai agent invoke --local '<payload>'`. But `protocolFromContainerAgent` in helpers.go returns `multiProtocolError` whenever there is exactly one invocable protocol and more than one protocol declared overall, so that suggested command then fails asking for `--protocol`.

It is not only the new ws combo: `responses` plus `invocations` already hits the same path today (both resolve to `responses`, the invoke hint is emitted, and invoke rejects it without `--protocol`). Carrying the selected protocol through and appending `--protocol <selected>` to the invoke suggestion would cover both the pre-existing and the new case, with a resolver-level test to lock it in.

jongio (commented):
Left one inline note on the multi-protocol invoke-hint path. It is a narrow edge case and partly pre-existing, so non-blocking, but worth a look before merge. The protocol-selection precedence and the activity/activity_protocol handling in this push line up with the earlier feedback.

jongio (approved):
Re-reviewed the incremental changes since my last look. The mixed-protocol handling I flagged is sorted: precedence stays responses > invocations > activity > invocations_ws regardless of the order the manifest declares them, activity and the legacy activity_protocol both suppress invoke hints, and invokeProtocolFlag now emits an explicit --protocol on all three invoke paths (ResolveAfterRun, ResolveAfterDeploy, appendInvokeLocalSecondary) for multi-protocol services.

The build break from the missing ServiceState.MultiProtocol field is fixed in e610152. go build ./... and go test ./internal/cmd/... both pass locally for the extension.

jongio (approved):
Re-reviewed the incremental changes since my last look. The mixed-protocol handling I flagged is sorted: precedence stays responses > invocations > activity > invocations_ws regardless of the order the manifest declares them, activity and the legacy activity_protocol both suppress invoke hints, and invokeProtocolFlag now emits an explicit --protocol on all three invoke paths (ResolveAfterRun, ResolveAfterDeploy, appendInvokeLocalSecondary) for multi-protocol services.

The build break from the missing ServiceState.MultiProtocol field is fixed in e610152. go build ./... and go test ./internal/cmd/... both pass locally for the extension.

jongio (approved):
Re-reviewed the incremental changes since my last look. The mixed-protocol handling I flagged is sorted: precedence stays responses > invocations > activity > invocations_ws regardless of the order the manifest declares them, activity and the legacy activity_protocol both suppress invoke hints, and invokeProtocolFlag now emits an explicit --protocol on all three invoke paths (ResolveAfterRun, ResolveAfterDeploy, appendInvokeLocalSecondary) for multi-protocol services.

The build break from the missing ServiceState.MultiProtocol field is fixed in e610152. go build ./... and go test ./internal/cmd/... both pass locally for the extension.

jongio (approved):
Re-reviewed the next-step changes layered on since my last pass.

The multi-protocol path now qualifies the emitted command with `--protocol` when a service declares more than one protocol, so a `responses` + `invocations_ws` service no longer produces an unqualified `azd ai agent invoke` that `protocolFromContainerAgent` would reject. Legacy `activity_protocol` is handled alongside the canonical `activity` value in both classification and invoke suppression, and protocol resolution now reads inline `azure.yaml` config before falling back to the service manifest. The test-cleanup ordering bug is fixed and BuildAndTest is green.

One non-blocking note: a hand-authored manifest declaring `invocations_ws` + `a2a` classifies as websocket-only and drops the post-deploy `a2a` invoke hint, since `a2a` isn't in the classifier switch. That isn't reachable through init today (`a2a` isn't an offered protocol), so it's fine to handle separately if it ever becomes one.

cli/azd/extensions/azure.ai.agents/internal/cmd/nextstep/state.go
```diff
@@ -460,36 +463,103 @@ func collectServices(
...
 
 	sawInvocations := false
+	sawActivity := false
+	sawInvocationsWS := false
 	for _, p := range hosted.Protocols {
 		switch strings.TrimSpace(p.Protocol) {
 		case ProtocolResponses:
-			return ProtocolResponses
+			return ProtocolResponses, multiProtocol
+		case ProtocolInvocationsWS:
```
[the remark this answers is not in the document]
v1212: Good catch, but this is a pre-existing gap rather than something this PR introduces: `loadServiceProtocol`/`loadServiceProtocolFromBytes` never handled `a2a` before this change (the switch has only covered responses/invocations/activity, now plus invocations_ws). This PR is scoped to adding `invocations_ws` support, so I'd prefer not to expand it into a broader local-vs-remote invocable protocol refactor (`IsInvocable`/`IsLocalInvocable`) here.

Tracking A2A next-step handling (A2A-only and A2A+WebSocket manifests) as a separate follow-up so it can get its own tests and review. Leaving this thread open as a reference for that work.

cli/azd/extensions/azure.ai.agents/internal/cmd/nextstep/state.go
v1212: Fixed in the latest push, and thanks for the thorough analysis — you were right that this hit both the new `responses` + `invocations_ws` combo and the pre-existing `responses` + `invocations` case.

`ServiceState` now carries a `MultiProtocol` flag (set when the definition declares more than one protocol), and `invokeProtocolFlag(svc)` appends `--protocol <selected>` to every invoke suggestion (`--local`, `invoke <name>`, and the post-init secondary) whenever the service is multi-protocol. Added resolver-level regression tests for the mixed-protocol cases.

jongio (approved):
The follow-up commits since my last pass resolve the non-invocable-protocol handling I flagged. `loadServiceProtocolFromBytes` now applies a stable, order-independent priority (responses > invocations > activity > invocations_ws), and `serviceSupportsAzdInvoke` suppresses the invoke hint for `invocations_ws` and both `activity` / `activity_protocol` spellings. Multi-protocol services emit an explicit `--protocol` flag via `invokeProtocolFlag` so the suggested invoke command is unambiguous, and inline `azure.yaml` definitions are picked up through `loadServiceProtocolFromConfig` in addition to `agent.yaml`. The `--protocol` help text is generated from `knownProtocolNames()` so it stays in sync with the known set.

Verified `go build ./...` and `go test ./internal/cmd/...` pass against the current HEAD.

cli/azd/extensions/azure.ai.agents/internal/cmd/nextstep/resolver.go
```diff
@@ -349,7 +353,7 @@ func ResolveAfterRun(state *State, serviceName string, readmeExists func(relativ
 		out = append(out, *readmeHint)
 	}
 	out = append(out, Suggestion{
-		Command:     fmt.Sprintf("azd ai agent invoke --local %s", invokeArg),
+		Command:     fmt.Sprintf("azd ai agent invoke --local %s%s", invokeProtocolFlag(svc), invokeArg),
```
trangevi, line 356: Formatting, does there need to be a space between the "%s" fields?

trangevi (approved):

jongio (approved):
The follow-up I flagged earlier is resolved: non-invocable protocols (activity, activity_protocol, invocations_ws) now suppress the `azd ai agent invoke` next-step hint, and serviceSupportsAzdInvoke gates every emission path (ResolveAfterRun, ResolveAfterDeploy, appendInvokeLocalSecondary). Protocol precedence responses > invocations > activity > invocations_ws is order-independent and locked by the reversed-order tests. Multi-protocol services emit an explicit --protocol flag with the resolved protocol, and inline service config now feeds detection with an agent.yaml fallback. Coverage across suppression, multi-protocol flag emission, inline-config precedence, and legacy activity is solid.

cli/azd/extensions/azure.ai.agents/internal/cmd/nextstep/resolver.go
v1212: The flush  %s%s  is intentional:  invokeProtocolFlag  returns  "--protocol <p> "  (trailing space) for multi-protocol and  ""  for single, so a literal space here would double-space the single-protocol case. Thanks!

PR #9012, opened 2026-07-07
fix(ext-agents): select a compatible Python for `azd ai agent run` instead of first on PATH

### LINKED ISSUE #8954: azd ai agent run uses first python.exe on PATH instead of compatible Python runtime
## Description

When running a newly initialized Foundry agent on Windows, `azd ai agent run`
fails with:

Python 3.13+ is required (found 3.11.9)

even though Python 3.13 is installed and available through the Windows Python launcher (`py`).

The project specifies:

runtime: python_3_13

in azure.yaml.

It appears that `azd ai agent run` executes the first `python.exe`
found on PATH instead of locating a compatible Python interpreter.

## Environment

OS:
Windows 11

azd:
1.27.0

Extensions:
azure.ai.agents 1.0.0-beta.2
azure.ai.projects 1.0.0-beta.1
(other Foundry extensions installed)

Python installations:

3.11
3.12
3.13

`py -0` reports:

-V:3.13 *
-V:3.12
-V:3.11

However,

python --version

returns

Python 3.11.9

because Python311 appears before Python313 in PATH.

## Steps to reproduce

1. Install Python 3.11, 3.12 and 3.13.
2. Ensure Python311 appears before Python313 on PATH.
3. Run

azd ai agent init -m "https://github.com/microsoft-foundry/foundry-samples/tree/main/samples/python/hosted-agents/agent-framework/responses/02-tools/agent.manifest.yaml"

4. Run

azd ai agent run

## Actual behavior

Dependency installation fails with

Python 3.13+ is required (found 3.11.9)

## Expected behavior

Since Python 3.13 is installed and the project specifies
runtime: python_3_13,
`azd` should locate a compatible Python interpreter for the requested runtime, or clearly document that the first python.exe on PATH must satisfy the runtime requirement.

## Workaround

Adding

Python313\
Python313\Scripts\

to PATH and placing them before Python311 resolves the issue immediately.

glharper:
## Summary

On Windows, `azd ai agent run` fails with `Python 3.13+ is required (found 3.11.9)` even when Python 3.13 is installed and selectable via the Windows Python launcher (`py -0` reports `3.13 *`). The project declares `runtime: python_3_13` in `azure.yaml`.

Root cause: the `pip` fallback path resolves Python via `findSystemPython`, which returned the **first** of `python3` / `python` / `py` found on `PATH` and then hard-failed in `checkPythonVersion` if that interpreter was too old. When `Python311` precedes `Python313` on `PATH`, `python.exe` is 3.11 and the resolver never falls through to the `py` launcher that would locate 3.13.

## Fix

`findSystemPython` now **probes multiple candidates and checks each one's version**, returning the first that satisfies the minimum supported runtime (>= 3.13):

- On Windows it prefers the **`py -3` launcher**, which selects the newest installed Python 3 (rather than whichever `python.exe` is first on `PATH`), then falls back to `python3`, `python`, and the default `py`.
- An interpreter is modeled as an executable path plus leading launcher args (`pythonInterpreter{path, args}`), so a `py -3` selection is carried through to venv creation (`py -3 -m venv ...`).
- When every resolvable candidate is too old, the error names the newest version found so the user knows what is installed.

This makes the `pip` path behave like the `uv` path, which already resolves a compatible interpreter via `uv venv --python ">=3.13"`.

```text
# before (Python311 before Python313 on PATH): fails
Python 3.13+ is required (found 3.11.9)

# after: `py -3` locates 3.13 and the venv is created successfully
```

## Changes

- `internal/cmd/run.go` — replace `findSystemPython`/`checkPythonVersion` with a version-aware resolver: `pythonInterpreter` (+`command`), `pythonCandidates`, `pythonVersion`, `pythonVersionOK`, `firstCompatiblePython`, and a rewired `findSystemPython`. Minimum runtime is now the named `minPythonMajor`/`minPythonMinor` constants. The pip-path call site uses the resolved interpreter directly.
- `internal/cmd/run_test.go` — unit tests for the selection logic (`firstCompatiblePython` with an injected version function, covering the Windows repro, skip-too-old, newest-incompatible error, unparseable-version skip, and empty cases), `pythonVersionOK`, `pythonVersion`, and the no-Python error.
- `CHANGELOG.md` — Unreleased entry.

## Testing

- `go build ./...`, `go vet`, `golangci-lint run ./internal/cmd/` — clean
- `go test ./internal/cmd/ -count=1` — pass
- `gofmt` and `cspell` on changed files — clean

Fixes #8954


jongio (approved):
Version-aware interpreter resolution with DI for testability, correct Windows launcher ordering, and informative error messages. The slices.Clone in command() prevents arg slice mutation. Looks correct.

RickWinter (commented):
This replaces first-python-on-PATH resolution with a version-aware resolver. findSystemPython now probes an ordered candidate set (on Windows the `py -3` launcher first, which selects the newest installed Python 3) and returns the first that satisfies >= 3.13, modeling an interpreter as path plus launcher args so a `py -3` selection carries through to venv creation. This is the right shape and makes the pip path behave like the uv path's `--python ">=3.13"` constraint.

The selection logic is factored behind an injected version function and covered directly: the Windows repro (py -3 picks 3.13 over a 3.11 python), skip-too-old, the newest-incompatible error naming the version found, unparseable-version skip, and the empty case. The too-old error now names the newest version on the machine, which is a real usability improvement over the prior hard-fail on whichever python came first.

No correctness or security concerns. Build, vet, lint, and the internal/cmd tests pass.

Disposition: no blockers.

cli/azd/extensions/azure.ai.agents/internal/cmd/run.go
```diff
@@ -773,58 +769,150 @@ func venvPip(venvDir string) string {
 	return filepath.Join(venvDir, "bin", "pip")
 }
 
-func findSystemPython() (string, error) {
-	candidates := []string{"python3", "python"}
+// Minimum supported system Python runtime. Kept in sync with the uv path's
+// `uv venv --python ">=3.13"` constraint.
+const (
```
trangevi, line 822: Can we do something to make the `uv` path also use these constants? Otherwise we have a pretty high chance of ending up out of sync

trangevi (changes requested):

glharper:
Good call -- addressed in ff179a5. Added `minPythonUvSpec()`, which builds the uv `--python` specifier from the shared `minPythonMajor`/`minPythonMinor` constants, and switched the `uv venv` command to use it. The pip-path version checks already use those constants, so both paths are now driven by a single source of truth. Also added `TestMinPythonUvSpec` asserting the specifier tracks the constants, so bumping the minimum stays in sync automatically.

jongio (approved):

glharper:

glharper:

cli/azd/extensions/azure.ai.agents/internal/cmd/run.go
jongio: This is addressed in the current diff. The uv path now calls `minPythonUvSpec()` (line 570) instead of the hardcoded `>=3.13` string. That function derives the spec from the shared `minPythonMajor`/`minPythonMinor` constants, so both paths stay in sync. There's also a `TestMinPythonUvSpec` test that asserts the derived string matches the constants.

jongio (approved):
Version-aware interpreter resolution is solid. The shared `minPythonMajor`/`minPythonMinor` constants are now the single source of truth for both the uv and pip paths, with `minPythonUvSpec()` deriving the uv constraint and `TestMinPythonUvSpec` guarding against drift. The `firstCompatiblePython` DI design (injected version function) keeps unit tests fast and deterministic without needing real Python on PATH. `slices.Clone` in `command()` prevents arg slice mutation. Error messages name the newest incompatible version found, which helps users diagnose their setup.

Replied to @trangevi's thread confirming the uv-path sync concern is addressed in the current diff.

trangevi (approved):

glharper:

trangevi (approved):

trangevi:
@microsoft-github-policy-service rerun

PR #9013, opened 2026-07-07
fix(toolboxes): branch skill/connection mutations from latest version

### LINKED ISSUE #8674: [azure.ai.toolboxes] `azd ai toolbox skill add` silently drops earlier skills (branches from default version, not latest)
## Summary

`azd ai toolbox skill add <toolbox> <skill>` creates a new immutable toolbox version, but it branches that new version from the toolbox's **default** version, not the **latest**. Sequential `add` calls therefore produce *sibling* versions — each containing only one new skill — instead of a chain that accumulates skills. There is no warning, no error, and the CLI output reads as if the operation appended to current state.

The same shape applies to `azd ai toolbox skill remove`, `azd ai toolbox connection add`, and `azd ai toolbox connection remove` (all three call `GetToolbox` → `GetToolboxVersion(tb.DefaultVersion)` then construct the new version off that snapshot, e.g. https://github.com/Azure/azure-dev/blob/main/cli/azd/extensions/azure.ai.toolboxes/internal/cmd/toolbox_skill_add.go#L132-L134).

Discovered while running an end-to-end test pass against the public toolbox + skills how-to docs on 2026-06-15 (extension `azure.ai.toolboxes 0.1.0-preview`).

## Repro (live, observed 2026-06-15)

Starting state: `web-search-toolbox` v1 = default, contains `web_search` tool, no skills. Two skills already exist in the project: `greeting-test-2606`, `code-review-test-2606`.

```pwsh
# Step 1 — add greeting
azd ai toolbox skill add web-search-toolbox greeting-test-2606
# Output: Created toolbox web-search-toolbox version 2 (attached skill greeting-test-2606).

# Step 2 — add code-review (intent: now have BOTH skills attached)
azd ai toolbox skill add web-search-toolbox 'code-review-test-2606@1'
# Output: Created toolbox web-search-toolbox version 3 (attached skill code-review-test-2606@1).

# Step 3 — verify
azd ai toolbox show web-search-toolbox --version 3
```

### Actual output (Step 3)

```
FIELD            VALUE
Name             web-search-toolbox
Default version  1
Shown version    3
Tools            1
Skills           1

SKILL                  VERSION  TYPE
code-review-test-2606  1        skill_reference
```

`greeting-test-2606` is **not** attached to v3. v3 was branched from v1 (default), not v2 (latest).

### Workaround that does what users expect

```pwsh
azd ai toolbox skill add web-search-toolbox greeting-test-2606    # v2
azd ai toolbox publish web-search-toolbox 2                       # promote v2 -> default
azd ai toolbox skill add web-search-toolbox 'code-review-test-2606@1'  # v3 = v2 + code-review (correct)
```

…or a single `--from-file` listing both skills.

## Why this is a footgun

Every other "add to a list" CLI in the ecosystem (`az`, `gh`, `kubectl`, `git`) appends to current state. The current behavior is silently lossy:

- No error, no warning at the moment of the wrong outcome.
- The CLI success message (`Created toolbox … version 3 (attached skill code-review)`) reads as a strict additive operation.
- The user only learns otherwise via `toolbox show`, after which they have orphan v2/v3/v4 versions to clean up.

The skill-group `--help` text says *"Each add/remove creates a new immutable version; the toolbox's default version is unchanged"* — accurate, but most readers parse it as "appends to current state".

## Proposed fix

One of:

1. **(Preferred)** Branch from latest version by default. Add an explicit `--from-version <n>` flag for users who want to fork from a specific version (current "default-snapshot" semantics still reachable via `--from-version <default-version-number>`).
2. Reject the second `add` with an error if `default != latest`, prompting the user to either `publish` or pass an explicit version selector.

Apply symmetrically to `skill remove`, `connection add`, `connection remove`.

## Stopgap UX changes to ship before / alongside the behavior fix

1. **CLI output** — surface the trap at the moment of action. Today's source: https://github.com/Azure/azure-dev/blob/main/cli/azd/extensions/azure.ai.toolboxes/internal/cmd/toolbox_skill_add.go#L238-L263. Suggested wording:

   ```
   Created toolbox web-search-toolbox version 3 (attached skill code-review-test-2606@1;
     branched from version 1 [default]; latest version is 2 — its skills are NOT carried over).
   Run `azd ai toolbox publish web-search-toolbox 2` first if you wanted to extend the latest version.
   ```

2. **`--help` text** — replace `"Each add/remove creates a new immutable version; the toolbox's default version is unchanged"` (in `toolbox_skill_group.go` and the corresponding `connection` help) with `"Each add/remove creates a new immutable version branched from the toolbox's default version. To branch from a different version, use --from-version. The toolbox's default version is unchanged."`

3. **Public docs** — add a Note callout in the [toolbox skills how-to](https://learn.microsoft.com/en-us/azure/foundry/agents/how-to/tools/skills) "Attach skills" section explaining the semantic and recommending `--from-file` (single call) or `publish`-between-adds for accumulation scenarios.

## Environment

- `azd 1.23.9` (current is 1.25.6 — same behavior expected from source; not yet re-verified end-to-end)
- Extension: `azure.ai.toolboxes 0.1.0-preview`
- Foundry project: `https://zhuoqunli-westus2.services.ai.azure.com/api/projects/proj`
- Discovered during: foundry-samples toolbox public-docs test pass on 2026-06-15. Internal cross-reference: finding **F-3.1** in `internal/findings/TOOLBOX_DOCS_FINDINGS_2026-06-15.md`.

## Severity / priority

Not a crash, but a silent data-loss-shaped UX trap on a preview feature with growing usage. Recommend addressing before toolbox + skills GA.

glharper:
Fixes #8674

## What

`azd ai toolbox skill add/remove` and `azd ai toolbox connection add/remove` created the new immutable toolbox version by branching from the toolbox's **default** version rather than its **latest**. Sequential mutations therefore produced *sibling* versions that each silently dropped earlier changes:

```pwsh
azd ai toolbox skill add web-search-toolbox greeting        # v2 = v1 + greeting
azd ai toolbox skill add web-search-toolbox code-review     # v3 = v1 + code-review  ❌ greeting dropped
```

No error, no warning — a silent, data-loss-shaped UX trap on a preview feature.

## Fix

Branch from the **latest** version by default so sequential mutations accumulate (`v3 = v2 + change`):

```pwsh
azd ai toolbox skill add web-search-toolbox greeting        # v2 = v1 + greeting
azd ai toolbox skill add web-search-toolbox code-review     # v3 = v2 + code-review  ✅ both present
```

- **`--from-version <n>`** (new, shared across all four verbs) overrides the branch source. The prior default-snapshot behavior is still reachable via `--from-version <default-version>`. An unknown version is rejected with a clear error.
- When the new version is branched from a **non-default** version, a note is printed clarifying the branch source and that the default is unchanged (text output only).
- The toolbox's default version is still unchanged — `publish` to promote, as before.

Implemented via a shared `resolveBranchVersion` helper (`toolbox_branch.go`) reused by all four verbs, reusing the existing numeric-aware `versionSortDescending`. When a toolbox reports no versions (edge case; a real toolbox always has its default), it falls back to the default version, preserving today's behavior. Duplicate/not-found messages now name the specific version being extended, and the skill/connection group `--help` documents the accumulate semantics and `--from-version`.

## Tests

- `skill add` / `skill remove` / `connection add` branch from **latest** (v2), not default (v1), carrying forward v2's contents.
- `--from-version` pins the branch source (v1) and does **not** carry v2's skills.
- unknown `--from-version` → `toolbox_version_not_found`, no version created.
- empty version list → falls back to default (preserves prior behavior).
- All existing toolbox tests still pass unchanged.

## Validation

## Notes

- Scope: this fixes the **behavior** (item 1 of the issue's "Proposed fix", preferred) plus the CLI-output note and `--help` wording (stopgap UX items 1–2). Public-docs updates (item 3) live in a separate docs repo and are out of scope for this PR.

### DIFF
--- a/cli/azd/extensions/azure.ai.toolboxes/internal/cmd/toolbox_branch.go
+++ b/cli/azd/extensions/azure.ai.toolboxes/internal/cmd/toolbox_branch.go
@@ -89,12 +89,13 @@ func resolveBranchVersion(
 		return res, nil
 	}
 
-	// Validate the requested version exists when we have a version list to
-	// check against. If the list is empty (edge case), trust the caller; a
-	// bad value surfaces as a not-found on the subsequent GetToolboxVersion.
-	if len(versions) > 0 && !slices.ContainsFunc(versions, func(v azure.ToolboxVersionObject) bool {
-		return v.Version == fromVersion
-	}) {
+	versionExists := fromVersion == tb.DefaultVersion
+	if len(versions) > 0 {
+		versionExists = slices.ContainsFunc(versions, func(v azure.ToolboxVersionObject) bool {
+			return v.Version == fromVersion
+		})
+	}
+	if !versionExists {
 		return branchResolution{}, exterrors.Validation(
 			exterrors.CodeToolboxVersionNotFound,
 			fmt.Sprintf("version %q does not exist for toolbox %q", fromVersion, toolboxName),
--- a/cli/azd/extensions/azure.ai.toolboxes/internal/cmd/toolbox_branch.go
+++ b/cli/azd/extensions/azure.ai.toolboxes/internal/cmd/toolbox_branch.go
@@ -4,6 +4,7 @@
 package cmd
 
 import (
+	"cmp"
 	"context"
 	"fmt"
 	"slices"
@@ -52,6 +53,13 @@ func registerFromVersionFlag(cmd *cobra.Command, fromVersion *string) {
 // An explicit --from-version overrides this (the prior default-snapshot behavior
 // is still reachable via --from-version <default>).
 //
+// "Latest" is the most recently created version: the toolbox API exposes no
+// canonical latest/head pointer (ToolboxObject carries only default_version), so
+// we derive it from the version list, preferring the newest CreatedAt and
+// falling back to the highest version number when CreatedAt is unavailable or
+// tied. Version numbers are monotonic, so this matches the true tip even if a
+// version is later deleted.
+//
 // When the toolbox reports no versions (an edge case; a real toolbox always has
 // its default), it falls back to the default version so the operation still
 // succeeds. A --from-version that does not exist is rejected.
@@ -70,9 +78,7 @@ func resolveBranchVersion(
 	latest := tb.DefaultVersion
 	if len(versions) > 0 {
 		sorted := slices.Clone(versions)
-		slices.SortFunc(sorted, func(a, b azure.ToolboxVersionObject) int {
-			return versionSortDescending(a.Version, b.Version)
-		})
+		slices.SortFunc(sorted, latestFirst)
 		latest = sorted[0].Version
 	}
 
@@ -98,6 +104,16 @@ func resolveBranchVersion(
 
 	res.Branch = fromVersion
 	return res, nil
+}
+
+// latestFirst orders two toolbox versions newest-first: by CreatedAt descending,
+// then by the numeric-aware version comparator as a deterministic tiebreaker
+// (and the sole signal when CreatedAt is not populated).
+func latestFirst(a, b azure.ToolboxVersionObject) int {
+	if a.CreatedAt != b.CreatedAt {
+		return cmp.Compare(b.CreatedAt, a.CreatedAt)
+	}
+	return versionSortDescending(a.Version, b.Version)
 }
 
 // printBranchNote surfaces, in text output only, that a new version was branched
--- a/cli/azd/extensions/azure.ai.toolboxes/internal/cmd/toolbox_branch.go
+++ b/cli/azd/extensions/azure.ai.toolboxes/internal/cmd/toolbox_branch.go
@@ -1,0 +1,122 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package cmd
+
+import (
+	"context"
+	"fmt"
+	"slices"
+	"strings"
+
+	"azure.ai.toolboxes/internal/exterrors"
+	"azure.ai.toolboxes/internal/pkg/azure"
+
+	"github.com/spf13/cobra"
+)
+
+// branchResolution describes which existing toolbox version a mutation
+// (skill/connection add or remove) should branch its new immutable version from.
+type branchResolution struct {
+	// Branch is the version the new version is created from.
+	Branch string
+	// Latest is the highest existing version of the toolbox.
+	Latest string
+	// Default is the toolbox's current default version.
+	Default string
+}
+
+// branchedFromNonDefault reports whether the new version was branched from a
+// version other than the toolbox's default. When true, the default version's
+// contents are NOT the base of the new version, which is worth surfacing to the
+// user (the new version accumulates on top of Latest, not Default).
+func (b branchResolution) branchedFromNonDefault() bool {
+	return b.Branch != b.Default
+}
+
+// registerFromVersionFlag registers the shared --from-version flag used by the
+// add/remove verbs to override which version a new version branches from.
+func registerFromVersionFlag(cmd *cobra.Command, fromVersion *string) {
+	cmd.Flags().StringVar(
+		fromVersion, "from-version", "",
+		"Version to branch the new version from (defaults to the latest version).",
+	)
+}
+
+// resolveBranchVersion determines which existing toolbox version a new version
+// should branch from.
+//
+// By default it branches from the toolbox's LATEST version so sequential
+// add/remove operations accumulate (v3 = v2 + change), rather than repeatedly
+// forking from the default version and silently dropping earlier changes.
+// An explicit --from-version overrides this (the prior default-snapshot behavior
+// is still reachable via --from-version <default>).
+//
+// When the toolbox reports no versions (an edge case; a real toolbox always has
+// its default), it falls back to the default version so the operation still
+// succeeds. A --from-version that does not exist is rejected.
+func resolveBranchVersion(
+	ctx context.Context,
+	client toolboxClient,
+	toolboxName string,
+	tb *azure.ToolboxObject,
+	fromVersion string,
+) (branchResolution, error) {
+	versions, err := client.ListToolboxVersions(ctx, toolboxName)
+	if err != nil {
+		return branchResolution{}, exterrors.ServiceFromAzure(err, exterrors.OpListToolboxVersions)
+	}
+
+	latest := tb.DefaultVersion
+	if len(versions) > 0 {
+		sorted := slices.Clone(versions)
+		slices.SortFunc(sorted, func(a, b azure.ToolboxVersionObject) int {
+			return versionSortDescending(a.Version, b.Version)
+		})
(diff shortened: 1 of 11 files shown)


jongio (approved):
Verified that 
esolveBranchVersion correctly sorts existing versions descending and branches from the latest, that --from-version validation rejects unknown versions with an actionable error, and that the fallback to DefaultVersion on an empty version list preserves prior behavior. The four verbs (skill add/remove, connection add/remove) all use the shared helper consistently, and the tests cover the key scenarios including the original regression case from #8674.

cli/azd/extensions/azure.ai.toolboxes/internal/cmd/toolbox_branch.go
```diff
@@ -0,0 +1,122 @@
...
+		return branchResolution{}, exterrors.ServiceFromAzure(err, exterrors.OpListToolboxVersions)
+	}
+
+	latest := tb.DefaultVersion
+	if len(versions) > 0 {
+		sorted := slices.Clone(versions)
+		slices.SortFunc(sorted, func(a, b azure.ToolboxVersionObject) int {
+			return versionSortDescending(a.Version, b.Version)
+		})
+		latest = sorted[0].Version
```
RickWinter, line 82: Here "latest" is computed as the highest version number rather than a tip pointer the service hands back. That is a fine proxy while versions only ever increment, but it assumes max-numeric always equals the most recently mutated version. If a version can be deleted such that the max number is no longer the true tip, or if the service exposes its own latest/head, this would branch from the wrong base and reintroduce a quieter form of the same drop this PR fixes. For a preview feature I think computing it here is acceptable. Can we confirm the toolbox API has no canonical latest we should prefer over deriving it?

RickWinter (commented):
This changes skill/connection add and remove to branch the new immutable version from the toolbox's latest version instead of its default, fixing the silent drop of earlier mutations in #8674, and adds a shared `--from-version` override plus a text-mode note. The approach is the right shape: one `resolveBranchVersion` helper feeds all four verbs, it reuses the existing numeric-aware `versionSortDescending`, and the duplicate and not-found messages now name the specific version being extended rather than "current default". Tests cover latest-vs-default for add and remove, `--from-version` pinning, an unknown version rejected as `toolbox_version_not_found`, and the empty-list fallback to default.

One question inline on how "latest" is derived. The extra `ListToolboxVersions` call added to every mutation is a reasonable cost for computing latest and is correct to leave as is. The empty-list branch trusting an unchecked `--from-version` is already documented in the code and surfaces cleanly on the later GetToolboxVersion, so that is fine too. Disposition: COMMENT, nothing blocking.

RickWinter (commented):
This changes skill/connection add and remove to branch the new immutable version from the toolbox's latest version instead of its default, fixing the silent data loss (#8674) where sequential mutations forked from default and dropped earlier changes. The approach is right: a shared resolveBranchVersion helper reused by all four verbs, latest computed with the existing numeric-aware versionSortDescending, an explicit --from-version override that validates against the version list before use, and a text-only note when the branch source is not the default.

The edge cases are handled well. An empty version list falls back to the default version, preserving prior behavior, and the duplicate and not-found messages now name the specific version being extended rather than "current default version", which matches the new semantics. Each mutation now makes one extra ListToolboxVersions call, which is inherent to resolving latest and is fine.

Tests cover branch-from-latest for skill add/remove and connection add (carrying v2's contents forward), the --from-version pin not carrying v2's skills, the unknown --from-version rejection with no version created, and the empty-list fallback. No correctness or security concerns.

Disposition: no blockers.

cli/azd/extensions/azure.ai.toolboxes/internal/cmd/toolbox_skill_add.go
```diff
@@ -171,7 +177,11 @@ func runSkillAddWith(
 		return exterrors.ServiceFromAzure(err, exterrors.OpCreateToolboxVersion)
 	}
 
-	return emitSkillAddResult(toolboxName, created.Version, specs, parent.output)
+	if err := emitSkillAddResult(toolboxName, created.Version, specs, parent.output); err != nil {
```
huimiu, line 181: I'm curious if we should carry `current.Policies` into this `CreateToolboxVersionRequest` as well. Connection add/remove already preserves policies, so I think skill add/remove should do the same; otherwise a toolbox with `policies.rai_config` could silently lose that governance setting after a skill mutation.

glharper: Great catch — fixed in 9954d93c6. skill add and skill remove now set Policies: current.Policies on the new version request, matching connection add/remove. Added TestRunSkillAddWith_CarriesForwardPolicies / TestRunSkillRemoveWith_CarriesForwardPolicies asserting an 
ai_config policy survives a skill mutation. This was the same silent-loss shape as the original bug — thanks for spotting it.

cli/azd/extensions/azure.ai.toolboxes/internal/cmd/toolbox_branch.go
glharper: Confirmed: the toolbox API has no canonical latest/head — `ToolboxObject` carries only `id`, `name`, and `default_version` (internal/pkg/azure), so `latest` must be derived client-side. I hardened that in 9954d93c6: `resolveBranchVersion` now derives latest by recency using `ToolboxVersionObject.CreatedAt` (newest first), falling back to the numeric-aware `versionSortDescending` when CreatedAt is unset or tied. That directly addresses the "max-numeric might not be the true tip" concern — if a highest-numbered version were ever not the most recent, CreatedAt wins; and since version numbers are monotonic, the two agree in the normal case (including after deleting the highest version). Added `TestResolveBranchVersion_LatestByCreatedAt` (recency beats a higher number) and `TestResolveBranchVersion_LatestNumericFallback` (numeric tiebreak when CreatedAt is 0).

jongio (approved):
Incremental commit looks good. Policies are now correctly preserved on skill add/remove (matching connection add/remove), and the latestFirst comparator correctly derives the true tip by CreatedAt before falling back to the numeric version sort. Tests cover the key scenarios well.

jongio (approved):
Re-reviewed the two commits since my last approval.

- The empty-list `--from-version` path now returns `toolbox_version_not_found` instead of a generic downstream service error, so the flag contract holds even when the version list comes back empty.
- The "not attached" suggestions on skill/connection remove now point to `azd ai toolbox show <name> --version <branch>` for the exact version being mutated, rather than `list` (which shows the default) - the diagnostic no longer contradicts the version named in the error.
- `skill add`/`skill remove` now carry `Policies: current.Policies` forward, matching connection add/remove, so a skill mutation can't silently drop governance policies.
- `latestFirst` derives the tip by `CreatedAt` (newest first) with the numeric-aware version sort as the tiebreak.

New tests cover connection-remove branching from latest, the empty-list rejection of an unknown `--from-version`, and policy carry-forward. `go test ./internal/cmd/...` passes.

trangevi (approved):

PR #9015, opened 2026-07-07
fix(agents): fail fast instead of hanging on the agent-service picker in non-TTY contexts

### LINKED ISSUE #8584: [ext-agents] Interactive agent-service picker hangs silently in non-TTY contexts instead of failing fast

glharper:
## Summary

Addresses the silent hang described in #8584. Several `azd ai agent` subcommands render an interactive "Select an agent service" picker when the target agent is ambiguous (a multi-service `azure.yaml` with more than one `azure.ai.agent` block). The picker is drawn by the azd host and blocks on stdin, so in a non-TTY context — CI, an agent harness, or output piped through `tee`/`tail` — it was invisible and the process hung forever with no error, spinner, or banner; only `Ctrl+C` recovered.

All of these commands funnel through a single shared helper, `promptForAgentService`, which is the shared root cause. This fixes it centrally.

## What changed

`promptForAgentService` (`internal/cmd/helpers.go`):

1. **Fail fast in non-TTY contexts.** When stdin/stdout is not a terminal, return a structured `exterrors.Validation` error (code `non_interactive_agent_selection`) that lists the available services and explains how to pass one, instead of rendering a picker that hangs:
   &gt; cannot prompt for agent service selection in a non-interactive context (stdin/stdout is not a TTY); multiple azure.ai.agent services are defined in azure.yaml: svc-a, svc-b

   with the suggestion to *pass the agent service name explicitly (positional argument or `--agent-name`), or run with `--no-prompt`*.
2. **Banner before the interactive picker.** In interactive TTY mode, print a one-line banner to stderr (`azd is waiting for your agent service selection below...`) before the picker so the user has immediate context while azd waits for a selection.

The existing `--no-prompt` fast-fail behavior is unchanged; the new guard covers the case where the user does **not** pass `--no-prompt` but has no TTY.

## Why this covers all the reported commands

`invoke --local`, `files upload`, `delete`, `show`, `monitor`, `endpoint show`, `eval`, `optimize`, `update`, etc. all resolve the agent through `resolveAgentService` → `promptForAgentService`, so fixing the one helper turns the multi-minute silent hang into an immediate, diagnosable error everywhere.

## Testing

- New unit tests: non-TTY fails fast with the right error code and never calls the host prompt; `--no-prompt` fails fast; interactive mode prints the stderr banner and prompts once.
- Existing picker tests updated to simulate a TTY through a small, overridable seam (`agentSelectionInteractive`).
- `go build ./...`
- Targeted `go test ./internal/cmd -run 'TestResolveAgentProtocol_ReturnsServiceName|TestResolveAgentProtocol_MultipleServicesPromptsOnce|TestPromptForAgentService_NonInteractiveFailsFast|TestPromptForAgentService_NoPromptFailsFast|TestPromptForAgentService_InteractivePrintsBanner'`

## Notes

The TTY check is exposed as a package-level function variable so tests can simulate a terminal; tests that use it run sequentially to avoid racing on the shared seam.

Fixes #8584


jongio (approved):
One comment accuracy nit (already flagged): the banner on the stderr line is only reachable in fully interactive mode (both stdin+stdout are TTYs), so the inline comment's claim about showing up in piped/CI logs describes an unreachable path. Consider simplifying it to explain the banner provides context to the user while the picker loads.

cli/azd/extensions/azure.ai.agents/internal/cmd/helpers.go
```diff
@@ -563,17 +583,36 @@ func promptForAgentService(
...
+			),
+			"pass the agent service name explicitly (as the command's positional argument or "+
+				"--agent-name), or run with --no-prompt to fail fast without prompting",
+		)
+	}
+
+	// Print a banner to stderr before the (interactive) picker. When the picker is visible this
+	// is harmless; when stdout is piped the banner still shows up in tee/tail/CI logs and
+	// explains why azd is blocked waiting for a selection.
+	fmt.Fprintln(os.Stderr, output.WithHintFormat("azd is waiting for your agent service selection below..."))
```
RickWinter, line 611: The comment above says this banner "still shows up in tee/tail/CI logs" when stdout is piped, but the non-TTY guard just above returns before we ever reach this line whenever stdout is not a terminal. So this only runs when both stdin and stdout are TTYs, which is exactly when the picker is already on screen. The banner is harmless, but its stated cross-piped-log value is unreachable as written. Either drop the piped-log justification from the comment, or, if you actually want the "why azd is waiting" hint to land in piped logs, emit it before the TTY guard.

RickWinter (commented):
This converts the silent hang in `promptForAgentService` into a fast, structured error when there is no TTY, and prints a one-line stderr banner before the interactive picker. The approach is the right shape: the shared helper is the correct single chokepoint for every command that resolves an agent service, the TTY seam matches the existing `isTerminal(os.Stdin/os.Stdout)` pattern already used across this extension (nextstep_output.go, doctor.go, run.go), and requiring both stdin and stdout to be terminals is correct since the host draws the picker to stdout. Tests cover the non-TTY fail-fast, the `--no-prompt` path, and the banner, and the process-global test seam is documented as serial.

One thing worth resolving before merge is the banner rationale, noted inline: the comment claims a piped-log benefit the control flow cannot deliver. It is harmless either way, so nothing here blocks. Disposition: COMMENT.

cli/azd/extensions/azure.ai.agents/internal/cmd/helpers.go
trangevi, line 622: From what I understand, non-tty contexts should automatically default to --no-prompt, causing this handling to default to the SelectedIndex, so I'm surprised we need any additional logic here (unless we don't want that default handling). @JeffreyCA @vhvb1989 can you confirm?

Ideally, we shouldn't duplicate some logic that is already built into azd, so even if this prompt doesn't work the way I think it should, everywhere else in the extension we've relied on the no-prompt flag to tell us when we're in a situation where we can't rely on interactive input, so we should be able to use that to simplify the above logic.

trangevi (changes requested):

cli/azd/extensions/azure.ai.agents/internal/cmd/helpers.go
jongio: The extension doesn't auto-default to no-prompt in non-TTY contexts. `rootFlags.NoPrompt` is only set from the explicit `--no-prompt` flag or the `AZD_NO_PROMPT` env var (root.go `PersistentPreRunE`), so a plain CI or piped run with neither set leaves `noPrompt=false`. That's why #8584 hangs today: without this guard the `if noPrompt` branch never fires, and the picker blocks on stdin that never arrives. If non-TTY already forced no-prompt, the bug wouldn't exist.

The two concepts are also distinct. `--no-prompt` means the user opted out of prompts (intent), while the TTY check means the environment can't render or read the picker (capability). CI and agent harnesses hit the second without the first.

On duplication: the TTY check isn't new logic. `isTerminal(os.Stdout.Fd())` is already the extension's convention. `stdoutIsTerminal` and `writerIsTerminal` in nextstep_output.go and doctor.go do the same thing, so this reuses the existing `isTerminal` helper rather than adding a parallel path.

Worth confirming the host-side behavior with @JeffreyCA / @vhvb1989, but from the extension code this guard is filling a real gap, not duplicating no-prompt.

glharper:

jongio (commented):
`agentSelectionInteractive` gates on raw `term.IsTerminal` (helpers.go:560 and 976), which honors none of azd's interactivity controls: `AZD_FORCE_TTY`, `IsRunningOnCI()`, and agent detection (see `internal/terminal/terminal.go:17`). The gap is bidirectional. `AZD_FORCE_TTY=false` under a PTY still returns true and reaches the picker, which is the hang this PR is trying to kill. `AZD_FORCE_TTY=true`, or a CI runner that allocates a PTY, returns false and fails fast even when the host would prompt.

The wrinkle is that this extension is its own module (`module azureaiagent`), so it can't import `cli/azd/internal/terminal` to reuse that precedence. Either replicate the `AZD_FORCE_TTY` / CI / agent-detection order locally in the extension, or have the host report interactivity so the two stay in sync.

jongio (approved):
My earlier feedback is addressed. The picker now gates on the host interactivity contract (AZD_FORCE_TTY / CI / agent detection) via `azdext.IsPromptSupported` instead of raw fd checks, so `AZD_FORCE_TTY=false` under a PTY and CI/agent runs both fail fast, and the module boundary is respected by living in `azdext` rather than the extension. The non-interactive `LocalError` is now preserved across the `invoke`, `eval`, and `optimize` paths, and the banner comment matches the only reachable path. Tests cover all three.

jongio (approved):

jongio (approved):
Re-affirming approval. The current HEAD still resolves the non-TTY hang through the host interactivity contract (AZD_FORCE_TTY / CI / agent detection), and the concerns from my earlier reviews remain addressed.

jongio (approved):
The non-TTY hang fix holds up. `promptForAgentService` gates on the host interactivity contract via `azdext.IsPromptSupported` (AZD_FORCE_TTY / CI / agent detection), so ambiguous agent selection fails fast with a structured `non_interactive_agent_selection` error instead of blocking on stdin, and the LocalError is preserved across the invoke, eval, and optimize paths. Tests cover the non-TTY, `--no-prompt`, banner, and cross-path cases.

jongio (approved):
Re-checked the current head. The non-TTY fail-fast in `promptForAgentService` goes through the host terminal check that honors `AZD_FORCE_TTY` and CI/agent detection, and the structured `LocalError` is preserved through the invoke, eval, and optimize resolvers instead of being reclassified as `invalid_parameter`. Tests cover the fail-fast, banner, and command-path preservation cases, and CI is green. The earlier review points are addressed, so no new findings from me.

jongio (commented):
@trangevi following up on your question about whether non-TTY contexts already default to `--no-prompt`. They do for AI agent callers, but not for CI or piped output, which is the gap this fixes.

`ParseGlobalFlags` in `cli/azd/cmd/auto_install.go:669-679` reads the explicit `--no-prompt` flag, then ORs in `agentdetect.IsRunningInAgent()`. `cli/azd/pkg/extensions/runner.go:60-62` forwards that to the child process as `AZD_NO_PROMPT=true`. So a Claude Code or Copilot CLI caller already lands in the `noPrompt` branch and never reaches the picker.

Plain CI (`CI`, `TF_BUILD`, `GITHUB_ACTIONS`) and piped stdout aren't covered. With neither `--no-prompt` nor an agent env var set, `opts.NoPrompt` stays false, nothing is propagated, and the picker blocks on stdin that never arrives. That's #8584.

On not duplicating logic azd already has: this revision does reuse it. `azdext.IsPromptSupported` in `cli/azd/pkg/azdext/shell.go` calls `internal/terminal.IsTerminal(os.Stdout.Fd(), os.Stdin.Fd())`, which is azd's own interactivity contract at `cli/azd/internal/terminal/terminal.go:17-35`: `AZD_FORCE_TTY` override first, then `resource.IsRunningOnCI()`, then `agentdetect.IsRunningInAgent()`, then isatty on both descriptors. The extension defines no rules of its own. And since `cli/azd/cmd/extensions.go:256` launches extensions with `Interactive: true`, which wires the child process straight to the console, that check sees the same terminal state azd does.

One thing stays inconsistent: the `noPrompt` path returns a plain error, while the non-TTY path returns a structured `non_interactive_agent_selection` carrying the `--agent` suggestion. Same user problem, two different error shapes. It's pre-existing, not something this PR introduced, but worth converging at some point.

Your changes-requested from the first commit is what's currently blocking here, and the code has moved since. Can you take another look?

cli/azd/extensions/azure.ai.agents/internal/cmd/helpers_test.go
```diff
@@ -403,16 +406,27 @@ func newHelpersTestAzdClient(
 
 	azdClient, err := azdext.NewAzdClient(azdext.WithAddress(listener.Addr().String()))
 	require.NoError(t, err)
+	t.Setenv("AZD_SERVER", listener.Addr().String())
```
jongio, line 409: `t.Setenv` panics in any test that has called `t.Parallel()`. Every caller on this branch is serial, so it passes here, but main has since added four parallel tests that go through this helper: `TestResolveAgentServiceFromProject_UsesVerifiedInlineNameForBrownfieldProject`, `_GreenfieldRequiresDeploy`, `_EnvironmentNameWins` and `_EnvLookupFailureIsReturned`.

Merging current main into this branch and running `go test ./internal/cmd/`:

```
--- FAIL: TestResolveAgentServiceFromProject_EnvLookupFailureIsReturned (0.01s)
panic: testing: test using t.Setenv, t.Chdir, or cryptotest.SetGlobalRandom can not use t.Parallel
FAIL    azureaiagent/internal/cmd
```

The panic takes down the entire package, not just those tests.

Only `TestNonInteractiveAgentSelectionErrorPreservedByCommandPaths` actually needs `AZD_SERVER` set, because `resolveEvalContext` and `resolveOptimizeAgent` build their own client from the environment rather than taking the one this helper returns. Setting a process-global env var in the shared helper to serve one test is what creates the collision.

Returning the listener address from `newHelpersTestAzdClient` and calling `t.Setenv` in that one test keeps the parallel callers working.

cli/azd/pkg/azdext/shell.go
```diff
@@ -208,6 +209,12 @@ func IsStdoutTerminal() bool {
 	return IsInteractiveTerminal(os.Stdout)
 }
 
+// IsPromptSupported reports whether the azd host considers the current process
+// capable of interactive prompting.
+func IsPromptSupported() bool {
```
jongio, line 214: This is a new exported symbol in `pkg/azdext`, and the agents extension consumes it at `helpers.go:559` as `agentSelectionInteractive`.

It only resolves today because the extension's `go.mod` still carries `replace github.com/azure/azure-dev/cli/azd => ../../`, which is commented as TEMPORARY ("Remove before merging, the core change must land first, then bump the azd dependency") and has already been removed on main. Main now pins `github.com/azure/azure-dev/cli/azd v1.28.0`, and v1.28.0 ships `IsStdoutTerminal` but not `IsPromptSupported`.

With main's `go.mod` in place, the extension no longer builds:

```
undefined: azdext.IsPromptSupported
```

Restoring the replace directive locally makes it build again, which confirms that's the only thing holding it together.

So this needs sequencing: the `azdext` change has to ship in a released azd, then the extension bumps its dependency to that version, and only then can the extension side land. Worth calling out explicitly on the PR so it isn't merged on the strength of a green CI run that used the old base.

jongio (commented):
Re-checked at b76258e. The non-TTY fix itself still behaves the way it should: the picker gates on the host interactivity contract through `azdext.IsPromptSupported` (so `AZD_FORCE_TTY`, CI detection and agent detection are all honored), the structured error survives the invoke, eval and optimize paths, and the extension tests pass on this branch.

Two things break once this is rebased, and neither shows up in the current CI run because the branch is 82 commits behind main and CI ran against the old base.

1. `t.Setenv` in `newHelpersTestAzdClient` panics the parallel tests main has added since this branched. Takes down the whole `internal/cmd` package. See helpers_test.go:409.
2. `azdext.IsPromptSupported` doesn't exist in the azd version the extension pins on main, so the extension won't compile after a rebase. See shell.go:214.

The PR is also currently in a conflicting state. The only real conflict is `internal/cmd/helpers_test.go`; `CHANGELOG.md` and `helpers.go` auto-merge cleanly. Resolving the text conflict alone won't be enough, both items above still need a fix.

`trangevi`'s changes-requested review is still open against d8e0ff4, three commits back. I answered the underlying question about non-TTY defaulting to `--no-prompt` in the thread.

PR #9017, opened 2026-07-07
`azd env refresh` with extension-provided hosts and provisioning providers

### LINKED ISSUE #7195: [Issue] azd env refresh doesn't work with azure.ai.agent extension

JeffreyCA:
Fixes #7195

This PR makes `azd env refresh` work in projects whose service host or provisioning provider is supplied by an extension (for example the `azure.ai.agents` extension, which provides the `azure.ai.agent` host and the `microsoft.foundry` provisioning provider). Refresh previously failed outright; it now pulls the latest deployment outputs into the local `.env` like any other project, and reports informationally when there is nothing to refresh yet.

image

### Issue

`env refresh` is a lightweight, read-only operation: it finds the last deployment, pulls its outputs, and writes them into `.azure/<env>/.env`. It does not build, package, or deploy, but it was coupled to machinery it does not need, and it never starts extensions, so two failures occurred.

- It called `projectManager.Initialize` and `EnsureAllTools` up front, which resolve a service target for every service. An extension-provided host like `azure.ai.agent` fails with `service host 'azure.ai.agent' for service 'agent' is unsupported`.
- Unlike `provision`/`deploy`/`up`, refresh does not attach the extensions middleware, so an extension-provided provisioning provider like `microsoft.foundry` is never registered and fails with `failed resolving IaC provider 'microsoft.foundry'`.

These pull in opposite directions: refresh does not need service targets, but it does need the provisioning provider, since reading deployment state is the provider's `State()` call.

### Decouple refresh from service targets

Refresh now initializes only framework services, never service targets, so an extension-provided host cannot block it.

- Framework-only initialization, best-effort per service: a service whose framework fails to initialize is skipped and reported as a warning instead of aborting the refresh.
- `ServiceEventEnvUpdated` still fires for the services that did initialize (keeping the .NET/Aspire user-secrets sync working), and only when a layer actually produced state.
- Removed the up-front `EnsureAllTools` call, since refresh never builds, packages, or deploys.

### Start extension-provided provisioning providers on demand

Rather than starting every installed extension (as the full middleware does), refresh activates only the extension that declares the configured provider, and only when needed.

- Starts the installed extension(s) declaring the configured provider name(s), waits for readiness, and stops the host afterward via deferred cleanup.
- Idempotent: already-resolvable providers are left alone, and names no installed extension declares fall through to native resolution, so `bicep`/`terraform` are a no-op.
- Deterministic selection when multiple extensions declare the same provider; required extensions start concurrently.
- If a required extension fails to start, refresh fails fast with the real startup error plus an `azd extension upgrade <id>` suggestion.
- The extension side is unchanged: its `State()` already reads outputs from the persisted subscription-scoped deployment (no on-disk template), which is exactly what refresh needs.

### Graceful handling and suggestions

- No deployment yet: refresh prints an informational message pointing at `azd provision` and exits successfully; an extension provider returning empty state is treated the same way. An explicit `--hint` that matches nothing stays a hard error.
- Provider not installed: on a resolution failure, refresh looks the provider up in the registry and returns an actionable `azd extension install <id>` suggestion when a match exists; other errors are returned as-is.

### Shared helper and VS parity

- Extracted the per-extension "start `listen` and wait for ready" logic into a shared helper used by both the middleware and the new activator.
- Applied the same framework-only best-effort init (and dropped `EnsureAllTools`) to the Visual Studio RPC refresh path.

### Behavior at a glance

| Scenario | Before | After |
| --- | --- | --- |
| Extension-provided host (e.g. `azure.ai.agent`) | `ERROR: service host '...' is unsupported` | Outputs refreshed into `.env` |
| Extension-provided provider (e.g. `microsoft.foundry`) | `ERROR: failed resolving IaC provider '...'` | Extension started on demand; outputs refreshed |
| Provider's extension not installed | Opaque resolution error | Error with `azd extension install <id>` suggestion |
| Required extension fails to start | Opaque resolution error later | Fails fast with `azd extension upgrade <id>` suggestion |
| No deployment yet (no `--hint`) | `ERROR: no deployments found` | Informational message, command succeeds |
| `--hint` matches no deployment | Error | Error (unchanged) |
| Plain `bicep`/`terraform` project | Works | Works, no extension started |

### Testing

- Unit tests for framework-only init under an unsupported host, best-effort per-service skipping, the activator helpers (distinct names, provider matching, deterministic selection, no-op path), and the `env refresh` action over the no-deployment, empty-state, and successful-refresh cases plus suggestion gating and IoC resolution.
- Verified with `go build ./...`, `go test` across the affected packages (including `-race`), and `golangci-lint run`.
- Tested manually on [notetaking agent sample](https://github.com/microsoft-foundry/foundry-samples/blob/12c97fbbf0518a7cac02acc1706cd5000e813ce0/samples/python/hosted-agents/bring-your-own/responses/notetaking-agent/agent.manifest.yaml) with azure.ai.agents v0.1.41-preview (native azd Bicep provisioner) and v1.0.0-beta.4 (custom in-memory Bicep provisioner)

### DIFF
--- a/cli/azd/cmd/middleware/extension_activator.go
+++ b/cli/azd/cmd/middleware/extension_activator.go
@@ -1,0 +1,249 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package middleware
+
+import (
+	"context"
+	"fmt"
+	"log"
+	"maps"
+	"slices"
+	"strings"
+	"sync"
+
+	"github.com/azure/azure-dev/cli/azd/internal"
+	"github.com/azure/azure-dev/cli/azd/internal/grpcserver"
+	"github.com/azure/azure-dev/cli/azd/pkg/extensions"
+	"github.com/azure/azure-dev/cli/azd/pkg/infra/provisioning"
+	"github.com/azure/azure-dev/cli/azd/pkg/ioc"
+	"github.com/azure/azure-dev/cli/azd/pkg/output"
+	"github.com/fatih/color"
+)
+
+// ExtensionActivator starts installed extensions on demand so the capabilities they provide
+// (currently provisioning providers) become resolvable in commands that do not run the full
+// ExtensionsMiddleware, such as `azd env refresh`. Unlike the middleware, it starts only the
+// extensions declaring a requested capability, and only when that capability is not already
+// resolvable (for example when a parent command's middleware already started every extension).
+type ExtensionActivator struct {
+	serviceLocator   ioc.ServiceLocator
+	extensionManager *extensions.Manager
+	extensionRunner  *extensions.Runner
+	globalOptions    *internal.GlobalCommandOptions
+}
+
+// NewExtensionActivator creates a new ExtensionActivator.
+func NewExtensionActivator(
+	serviceLocator ioc.ServiceLocator,
+	extensionManager *extensions.Manager,
+	extensionRunner *extensions.Runner,
+	globalOptions *internal.GlobalCommandOptions,
+) *ExtensionActivator {
+	return &ExtensionActivator{
+		serviceLocator:   serviceLocator,
+		extensionManager: extensionManager,
+		extensionRunner:  extensionRunner,
+		globalOptions:    globalOptions,
+	}
+}
+
+// EnsureProvisioningProviders starts the installed extension(s) that declare any of the given
+// provisioning provider names, so those providers can be resolved from the IoC container. Names no
+// installed extension declares are left to native resolution, and already-resolvable providers are
+// left alone, making activation idempotent. The returned cleanup (always non-nil, safe to defer)
+// stops the extension host when one was started. A required extension failing to start is an
+// error, since the caller cannot make progress without the provider it declares.
+func (a *ExtensionActivator) EnsureProvisioningProviders(
+	ctx context.Context,
+	providerNames []string,
+	environmentName string,
+) (cleanup func(), err error) {
+	noop := func() {}
+
+	names := distinctProviderNames(providerNames)
+	if len(names) == 0 {
+		return noop, nil
+	}
+
+	installed, err := a.extensionManager.ListInstalled()
+	if err != nil {
+		return noop, err
+	}
+
+	toStart := extensionsForProviders(installed, names)
+
+	// Drop extensions whose requested providers are all already resolvable (host already running).
+	toStart = slices.DeleteFunc(toStart, func(ext *extensions.Extension) bool {
+		return !slices.ContainsFunc(names, func(name string) bool {
+			return providerFromExtension(ext, name) && !a.providerResolvable(name)
+		})
+	})
+
+	if len(toStart) == 0 {
+		return noop, nil
+	}
+
+	var grpcServer *grpcserver.Server
+	if err := a.serviceLocator.Resolve(&grpcServer); err != nil {
+		return noop, err
+	}
+
+	serverInfo, err := grpcServer.Start()
+	if err != nil {
+		return noop, err
+	}
+
+	cleanup = func() {
+		if err := grpcServer.Stop(); err != nil {
+			log.Printf("failed to stop gRPC server after extension activation: %v", err)
+		}
+	}
+
+	startOpts := extensionStartOptions{
+		debug:       a.globalOptions.EnableDebugLogging,
+		cwd:         a.globalOptions.Cwd,
+		environment: environmentName,
+		noPrompt:    a.globalOptions.NoPrompt,
+		forceColor:  !color.NoColor,
+	}
+
+	// Start the required extensions concurrently, mirroring ExtensionsMiddleware.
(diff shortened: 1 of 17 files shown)


jongio (approved):
Clean decomposition that unblocks zd env refresh for extension-provided hosts and provisioning providers.

Highlights:
- Framework-only init path is a clean separation, correctly preserving the original init order (framework then target) while making the framework-only path independently callable.
- ExtensionActivator is properly idempotent (checks providerResolvable before starting anything), handles concurrent startup, and cleans up deterministically.
- Extracted startAndWaitExtension avoids duplicating the middleware's startup logic.
- Edge-case handling is thorough: nil state, no deployment (informational exit), provider-resolution failures surface actionable install/upgrade suggestions.
- VS RPC refresh path gets parity treatment via InitializeFrameworks.
- Tests cover the key scenarios well (no-deployment, nil state, successful refresh, suggestion gating, framework-only init under unsupported host, best-effort skipping, activator helpers).

jongio (approved):
New commit since my prior approval adds solid test coverage for the ExtensionActivator: construction, provider resolution, no-match path, already-resolvable (no-op) path, and SuggestExtensionForProvider with subtests for empty name, installed-declares-provider, registry match, and no-match. Tests are well-structured, use t.Parallel()/t.Context(), and follow the codebase's mock patterns. No concerns.

RickWinter (commented):
This decouples `env refresh` from service-target initialization and starts extension-provided provisioning providers on demand, so refresh works for extension hosts like `azure.ai.agent` and providers like `microsoft.foundry` instead of failing outright. The shape is right: refresh is read-only, so initializing frameworks only (best-effort per service) and activating just the extension that declares the configured provider is the correct minimal surface, and it reuses the middleware's start-and-wait path (`startAndWaitExtension`) rather than duplicating it. I traced the suggestion path end to end: an unregistered named provider surfaces a `container:` error that ioc wraps as `ErrResolveInstance`, and `suggestionForUnresolvedProvider` gates on exactly that, so it correctly distinguishes 'extension not installed' from 'provider constructor failed'. The empty-state and no-deployment handling, the `--hint` hard-error carve-out, the nil-`State` guard, and the `stateRefreshed` gate on raising `ServiceEventEnvUpdated` all look correct, and the `InitializeFrameworkService` extraction preserves the existing init/guard ordering without a nil-map hazard (`isComponentInitialized` lazily creates the inner map). Tests and the `extension-framework.md` update are thorough. The one thing worth a look is the case-sensitivity asymmetry between extension selection and provider resolution; it is narrow and pre-existing in spirit, so it can follow. Nothing blocks.

cli/azd/cmd/middleware/extension_activator.go
```diff
@@ -0,0 +1,249 @@
...
+		}
+	}
+
+	return declared
+}
+
+// providerFromExtension reports whether the extension declares a provider with the given name.
+func providerFromExtension(ext *extensions.Extension, providerName string) bool {
+	return slices.ContainsFunc(ext.Providers, func(p extensions.Provider) bool {
+		return strings.EqualFold(p.Name, providerName)
```
RickWinter, line 249: `providerFromExtension` matches the requested provider name case-insensitively, but resolution is case-sensitive: `ParseProvider` is a no-op and `Manager.newProvider` calls `ResolveNamed(providerKey, ...)` with the literal `infra.provider` string, as does `providerResolvable`. So if azure.yaml spells the provider with different casing than the extension registers it (say `Microsoft.Foundry` vs `microsoft.foundry`), we start the extension here but then still fail to resolve the provider, and `SuggestExtensionForProvider` returns empty because its installed-extension check also uses EqualFold. It's narrow (provider names are conventionally lowercase, and native resolution has always been case-sensitive), so not a blocker, but the asymmetry means 'we started the right extension' and 'we can resolve it' can disagree. Worth a normalize step or a short comment noting the intent.

hemarina (approved):

PR #9019, opened 2026-07-08
feat(provision): provider-agnostic provision validation checks

### LINKED ISSUE #9016: Enable provider-agnostic client-side validation ('azd validate') for extension-provided provisioning providers
## Summary

Extensions that ship their own provisioning provider (e.g. `azure.ai.agents` via `microsoft.foundry`, `microsoft.azd.demo` via `demo`) **cannot** participate in azd's client-side validation ("provision validation", formerly *local-preflight*). The validation dispatch is hard-wired to the **Bicep provider only**, so extension-registered checks register successfully but **never run** when provisioning goes through a non-Bicep provider.

This blocks **PR #9007** (adds a resource-group location-mismatch check to `azure.ai.agents`) from working at all, and blocks the broader goal of **centralizing** warnings/suggestions in azd core instead of each extension inventing its own error/warning output.

Related:
- PR #9007 — the check that is currently dead code (this issue unblocks it)
- PR #8844 / issue #7113 — renames `local-preflight` → **`validation.provision`**, explicitly *"to scale with future validation use cases"*. This issue is the logical next step of that arc.

---

## Background: how validation works today

Core already has a full validation framework (`internal/grpcserver/validation_service.go`, `pkg/infra/provisioning/validation_dispatcher.go`):

- Extensions declare the `validation-provider` capability and register checks via `WithValidationCheck` (gRPC `RegisterValidationCheck`).
- Checks return **structured** `ValidationCheckResult`: `severity` (WARNING/ERROR), `message`, `suggestion`, `diagnosticId`, doc `links`.
- Core dispatches them in **parallel** (`ValidationService.DispatchChecks`, `wg.Go`, grouped per extension, timeout-bounded).
- Results render through a **uniform report** and use the `(abort, err)` pattern: ERROR cancels provision (exit 0), WARNING continues.

## The blocking problem (root cause)

`DispatchChecks` is called from **exactly one place** in all of `cli/azd`:

- `pkg/infra/provisioning/bicep/bicep_provider.go:~2615` inside `(*BicepProvider).validatePreflight`.

Every other provisioning provider bypasses it:

- `ExternalProvisioningProvider.Deploy/Preview` (`internal/grpcserver/external_provisioning_provider.go`) just forward over gRPC — **no dispatch**.
- So extension providers (`microsoft.foundry`, `demo`) and the core Terraform provider never trigger any validation checks.

**Empirically confirmed** (both gaps):

- **Gap A (capability):** `azure.ai.agents/extension.yaml` was missing the `validation-provider` capability, so registration was rejected with `PermissionDenied`, which crashed extension startup (`WARNING: 1 extension did not start`) and made `azd provision` fail with an unrelated `transport is closing` error.
- **Gap B (dispatch):** Even after adding the capability so registration succeeds, the check **still never runs** — `azure.ai.agents` provisions via `microsoft.foundry`, which never calls `DispatchChecks`. Provisioning went straight to the Foundry ARM deploy and failed with the exact `InvalidResourceGroupLocation` error the check was meant to prevent. The `microsoft.azd.demo` example reproduces the same: its `demo_warning` check registers but never fires when provisioning via the `demo` provider.

**Why this matters for `azure.ai.agents` specifically:** every agents init flow stamps `infra.provider: microsoft.foundry` (`init.go`, `init_from_code.go`, `init_adopt.go`). `--infra`/`--infra=bicep` eject leaves azure.yaml unchanged (foundry compiles on-disk Bicep itself); `--infra=terraform` stamps the core Terraform provider. **None** of these routes through the core Bicep provider, so a registered check is dead code.

---

### Option A — New provider-agnostic validate check type + core trigger before provision **(RECOMMENDED)**

Introduce a new validation `checkType` (e.g. `"provision"` / `"pre-provision"`) with a **lean, provider-agnostic context** (env name, subscription, location, resource group, target scope — **no** ARM template, since foundry/terraform have none). Dispatch it from a **provider-agnostic point**, immediately before the provider runs:

- Natural insertion site: `provisioning.Manager.Deploy` (`pkg/infra/provisioning/manager.go:~99`), right before `m.provider.Deploy(ctx)` — and the equivalent point in `Manager.Preview`.
- This makes Bicep, Terraform, and **every** external/extension provider (foundry, demo) trigger the same checks identically.

**Pros**
- Reuses the entire existing machinery: structured results, uniform report, parallel dispatch, severity/abort semantics, capability gate.
- Keeps checks pure/declarative (return results, no side effects).
- Works for all providers → fixes Gap B and unblocks PR #9007 with minimal extension change.
- Matches the direction of PR #8844 (`validation.provision` "scales to future use cases").

**Cons**
- New core plumbing: new `checkType` constant, new provider-agnostic context builder, new dispatch site in `Manager.Deploy`/`Preview`, plus telemetry + a config gate mirroring `validation.provision`.
- Two dispatch sites coexist temporarily (Bicep's ARM-rich `local-preflight` context + the new lean provider-agnostic one) until unified later.

### Option B — Enhance lifecycle hooks/events to emit validate-layer warnings/errors (not recommended)

Change the hook/event contract so a `preprovision` handler can return structured warnings/suggestions and feed them into the same report.

**Pros:** reuses existing subscription/registration; already provider-agnostic; fewer new user-facing concepts.

**Cons (why it loses):**
- Semantic mismatch: hooks/events are **side-effecting actions** ("run this script"), not pure validators. Today `EventDispatcher.RaiseEvent` handlers return **`error` only** (a suggestion is possible solely via `ErrorWithSuggestion`, and only on the abort path — no "warn and continue" channel).
- Retrofitting a WARN/continue channel changes the contract for **both** gRPC event handlers and azure.yaml **script** hooks (a script's stdout/exit code has no severity/suggestion/links shape).
- Events run **sequentially** (`event_dispatcher.go` `for` loop, with a `TODO` to parallelize) vs. validation's parallel dispatch — would need a parallelism/ordering rework.
- Mixing side effects into a "validation phase" breaks the guarantee that validation is read-only, and muddies both abstractions.

---

## Recommendation

Implement **Option A**. It reuses the existing structured-result + uniform-report + parallel-dispatch + severity/abort machinery, keeps validation pure and read-only, works across all providers, and aligns with the `validation.provision` generalization already in flight (PR #8844).

**Design nuance to decide early:** either
- (a) introduce a **new** `checkType` with a lean context now and converge with Bicep's `local-preflight` later (**preferred** — doesn't destabilize shipping Bicep checks), or
- (b) hoist the existing `local-preflight`/`validation.provision` dispatch out of `BicepProvider` into `Manager.Deploy` and make ARM-specific context fields optional (avoids two sites, but forces every existing Bicep check to tolerate a missing ARM template).

Recommend **(a)**.

---

## Scope of core work (for the implementing session)

1. **New checkType + context**
   - Add a `checkType` constant (e.g. `"provision"` / `"pre-provision"`).
   - Define a provider-agnostic context schema + builder (env, subscription, location, resource group, target scope). Add new context key constants in `pkg/azdext/validation_provider.go` alongside the existing `ValidationContext*` keys, and a typed accessor on `ValidationContext`.

2. **Provider-agnostic dispatch site**
   - Call `ValidationCheckDispatcher.DispatchChecks(ctx, "<new-type>", ctxData)` in `provisioning.Manager.Deploy` (before `m.provider.Deploy`) and `Manager.Preview` — resolving the dispatcher from IoC (already wired at `cmd/container.go:~1031`).
   - Reuse the Bicep site's severity mapping + `(abort, err)` handling + timeout bound (`extensionValidationTimeout`) + uniform report rendering (`output/ux` preflight/provision-validation report). Extract shared logic rather than copy-pasting from `bicep_provider.go`.

3. **Config gate + telemetry**
   - Mirror the `validation.provision` gate from PR #8844 (respect the same skip semantics / naming).
   - Emit the same validation telemetry (rule IDs invoked, warning/error counts, outcome). Follow the telemetry-doc sync rules in `cli/azd/AGENTS.md` (schema, matrix, privacy checklist, coverage test). Hash user-derived values.

4. **Docs**
   - `cli/azd/docs/environment-variables.md` (if a new env/config gate is user-visible), telemetry docs, and the design doc renamed by PR #8844 (`docs/design/provision-validation.md`).

5. **Tests**
   - Unit test the new dispatch site (provider-agnostic, fires for a non-Bicep provider).
   - Ideally an end-to-end style test using the `microsoft.azd.demo` extension's `demo` provider proving the check now fires (today it registers but never runs).

**Coordination:** align with / sequence after **PR #8844** (rename `local-preflight` → `validation.provision`) to avoid churn and reuse its config-gate + report + telemetry naming.

---

## Key references (file:line)

- Only dispatch site (Bicep-only): `pkg/infra/provisioning/bicep/bicep_provider.go:~2615` (`validatePreflight`, `DispatchChecks("local-preflight", …)`).
- Provider-agnostic wrapper (insertion point): `pkg/infra/provisioning/manager.go:~99` (`Manager.Deploy` → `m.provider.Deploy`), `:~274` (`Manager.Preview`).
- External provider bridge (no dispatch): `internal/grpcserver/external_provisioning_provider.go` (`Deploy`/`Preview`).
- Dispatcher + parallel execution: `internal/grpcserver/validation_service.go:~214` (`DispatchChecks`), capability gate `:~79`.
- Dispatcher interface: `pkg/infra/provisioning/validation_dispatcher.go`.
- IoC wiring: `cmd/container.go:~1031` (`ValidationService` as `ValidationCheckDispatcher`).
- Extension SDK types + context keys: `pkg/azdext/validation_provider.go`.
- Event/hook model (Option B reference): `pkg/ext/event_dispatcher.go:~101` (`RaiseEvent`, sequential), `:~167` (`Invoke`, pre/post).
- The check to migrate once core lands: `extensions/azure.ai.agents/internal/project/resource_group_location_check.go` + registration in `internal/cmd/listen.go`.
- Example provider that reproduces Gap B: `extensions/microsoft.azd.demo/` (`demo` provider + `demo_validation_check.go`).

## Acceptance criteria

- A validation check registered by an extension with the new checkType runs **before provision** regardless of the provisioning provider (Bicep, Terraform, and external/extension providers such as `microsoft.foundry` and `demo`).
- WARNING results are surfaced via the uniform report and provisioning continues; ERROR results cancel provisioning with the standard `(abort, err)` behavior.
- The `microsoft.azd.demo` `demo_warning` check fires when provisioning via the `demo` provider (regression coverage).
- Telemetry, config gate, and docs updated in sync per `cli/azd/AGENTS.md`.

## Follow-up (tracked separately, this repo/session)

Once this lands and merges, update **PR #9007** to: add the `validation-provider` capability to `azure.ai.agents/extension.yaml`, re-register `ResourceGroupLocationCheck` under the new checkType, and use the provider-agnostic context (resource group + location) instead of the ARM-template-centric `local-preflight` context.

vhvb1989:
## Summary

Resolves #9016.

Adds a new **provider-agnostic `provision` validation check type** that is dispatched **once per `azd provision` / `azd up`, from the command layer, before the layer graph runs** — independently of which provisioning provider is selected. Previously, client-side validation checks (`local-preflight`) were dispatched **only** by the built-in Bicep provider, so extension-provided provisioning providers (e.g. `microsoft.foundry`, and future AI-extension checks) could never participate in validation.

This unblocks adding a validation check type in the AI extension (the follow-up in #9016 is tracked separately).

## What changed

- **New check type**: `ValidationCheckTypeProvision` (and named `ValidationCheckTypeLocalPreflight`) in `pkg/azdext`, plus a **lean provision context** (`env_name`, `subscription_id`, `resource_group`, `target_scope`, reusing `env_location`) with accessors `EnvName()`, `SubscriptionID()`, `ResourceGroup()`, `TargetScope()`. No ARM template is assumed (Terraform/foundry have none).
- **Provider-agnostic dispatch (hoisted to the command layer)**: the exported `Manager.RunProvisionValidation` is invoked **once** from `provisionLayersGraph` (covers `--preview` + single/multi-layer) and once from the up graph (guarded on `len(layers) > 0`), **before** the layer execution graph. It is **not** called from `Manager.Deploy`/`Preview` — multi-layer provisioning runs each layer's `Deploy` through its own `Manager` concurrently, so dispatching per-`Deploy` would fire the checks (and any warning prompt) once per layer. Bicep still **also** runs its own `local-preflight` checks internally — different check type, no conflict.
- **Abort/error semantics**: introduces `ErrProvisionValidationAborted`; on abort (error-severity finding or a declined warning) `RunProvisionValidation` returns `abort=true`, and the action surfaces the sentinel through `wrapProvisionError`, which maps it to `internal.ErrAbortedByUser` (exit code 0). Bicep's own provider-level preflight abort (`PreflightAbortedSkipped`) is a separate, still-handled path.
- **Telemetry**: reuses the existing `validation.preflight` event and its outcome/count/rule fields, and the existing `provision.preflight = off` config gate. Because the `provision` and Bicep `local-preflight` dispatches share the same event (and both fire on a Bicep provision), this PR **adds one new field, `validation.preflight.check_type`** (`local-preflight` | `provision`), emitted at both dispatch sites so downstream consumers can distinguish them and avoid double-counting. The new field is documented across `docs/reference/telemetry-data.md`, `docs/specs/metrics-audit/telemetry-schema.md`, and `docs/specs/metrics-audit/feature-telemetry-matrix.md`. The field is `SystemMetadata`/`FeatureInsight` and not user-derived, so it is not hashed and needs no privacy-hashing-table entry.
- **Demo extension registers BOTH check types**: the Bicep-only `local-preflight` (`demo_warning`, gracefully skipped when no Bicep snapshot exists — not dead code) and the new agnostic `provision` (`demo_provision_warning`).

## Testing

- New unit tests in `provision_validation_test.go` calling `Manager.RunProvisionValidation` directly (dispatch outcomes: pass / warnings-accepted / aborted / dispatch-error non-fatal / no-results proceeds / subscription-scoped context) and demo `demo_provision_validation_check_test.go`.
- `go build ./...`, targeted `go test` for changed packages, gofmt, golangci-lint, cspell, copyright — all pass.
- **Manually verified end to end**:
  - Demo provider sample (`infra.provider: demo`) → `demo_provision_warning` fires before provisioning.
  - Bicep sample (`todo-nodejs-mongo-aca` with `host: demo`) → **both** `demo_provision_warning` (provision) and `demo_warning` (local-preflight) fire.

## Notes for reviewers

- The `provision` check runs for **every** project whenever an extension declaring `validation-provider` is installed (extensions with that capability are started globally, and dispatch fans out to all registered checks). Extensions that only apply to specific project types must **self-condition** inside their `Validate` handler (via the lean context or by calling back into azd for `azure.yaml`) and return a passing result otherwise.
- Because the dispatch is hoisted before the layer graph, the provider-agnostic `provision` validation now runs **before** preprovision hooks (and, in `up`, before packaging). This is intentional fail-fast behavior; the lean context is env-scoped and does not depend on hooks or packaging.
- Updating the `azure.ai.agents` extension to consume the new check type is **out of scope** here and tracked separately in #9016.

image


cli/azd/pkg/infra/provisioning/provision_validation.go
```diff
@@ -0,0 +1,208 @@
...
+	// registered, so there is nothing to validate.
+	var dispatcher ValidationCheckDispatcher
+	if resolveErr := m.serviceLocator.Resolve(&dispatcher); resolveErr != nil {
+		return false, nil
+	}
+
+	ctx, span := tracing.Start(ctx, events.PreflightValidationEvent)
+	defer func() {
+		span.EndWithStatus(err)
+	}()
```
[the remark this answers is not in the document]
vhvb1989: Fixed: added a `validation.preflight.check_type` field (SystemMetadata / FeatureInsight, fixed enum `local-preflight` | `provision`) set at both dispatch sites, so the two emissions can be distinguished and Bicep provisions are no longer double-counted. Documented in `docs/reference/telemetry-data.md` (fields table + a query gotcha under Data Nuances), and in the metrics-audit `telemetry-schema.md` and `feature-telemetry-matrix.md`. The field is not user-derived, so no hashing/privacy-table entry is required. Also replaced the `local-preflight` magic string in the Bicep dispatch with the `azdext` constant.

cli/azd/pkg/infra/provisioning/provision_validation.go
```diff
@@ -0,0 +1,213 @@
...
+	}
+
+	if len(invokedRuleIDs) > 0 {
+		span.SetAttributes(fields.PreflightExtensionRulesKey.StringSlice(invokedRuleIDs))
+	}
+
+	if len(results) == 0 {
+		span.SetAttributes(fields.PreflightOutcomeKey.String(provisionValidationOutcomePassed))
+		return false, nil
+	}
```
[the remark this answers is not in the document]
vhvb1989: Fixed: `runProvisionValidation` now records `provisionValidationOutcomeError` (instead of `passed`) when `dispatchErr` is non-nil and no results are returned, so a dispatch failure/timeout is no longer masked as a successful validation in telemetry. Provisioning still proceeds — dispatch errors remain non-fatal, matching the existing log-and-skip behavior.

cli/azd/internal/tracing/fields/fields.go
```diff
@@ -701,6 +701,20 @@ var (
...
+	// "provision" dispatch. Without it, downstream consumers would double-count
+	// the event for Bicep provisions (where both sites fire). Values are the
+	// fixed, code-defined check-type identifiers.
+	//
+	// Example: "local-preflight", "provision"
+	PreflightCheckTypeKey = AttributeKey{
+		Key:            attribute.Key("validation.preflight.check_type"),
+		Classification: SystemMetadata,
+		Purpose:        FeatureInsight,
+	}
```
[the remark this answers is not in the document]
vhvb1989: Fixed: updated the PR description. The stale "no new telemetry surface" wording is replaced with a Telemetry section that explicitly calls out the new `validation.preflight.check_type` field and the three telemetry docs it touches, so reviewers won't skip the telemetry-doc changes.

cli/azd/docs/design/local-preflight-validation.md
```diff
@@ -292,7 +292,82 @@ described above.
...
+
+Typed accessors (`EnvName()`, `SubscriptionID()`, `EnvLocation()`,
+`ResourceGroup()`, `TargetScope()`) are available on `azdext.ValidationContext`.
+
+### Shared Behavior
+
+The `"provision"` dispatch reuses the same machinery as `"local-preflight"`:
+parallel dispatch, the uniform preflight report, severity/abort semantics
+(WARNING prompts to continue; ERROR aborts with exit code 0), the
+`extensionValidationTimeout` bound, and the `provision.preflight` config gate
```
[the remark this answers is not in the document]
vhvb1989: Fixed: the design doc's "Shared Behavior" section now references `provisionValidationTimeout` (the 60s bound the provider-agnostic path actually uses in `provision_validation.go`) and notes it mirrors the Bicep path's `extensionValidationTimeout`, so tracing the code from the doc lands on the correct constant.

cli/azd/pkg/infra/provisioning/manager.go
```diff
@@ -97,6 +97,16 @@ var AzdOperationsFeatureKey = alpha.MustFeatureKey("azd.operations")
...
 func (m *Manager) Deploy(ctx context.Context) (*DeployResult, error) {
+	// Provider-agnostic client-side validation runs before any provider so that
+	// extension-registered "provision" checks fire regardless of the
+	// provisioning provider (Bicep, Terraform, or extension providers such as
+	// microsoft.foundry and demo).
+	if abort, err := m.runProvisionValidation(ctx, false); err != nil {
+		return nil, err
+	} else if abort {
+		return &DeployResult{SkippedReason: PreflightAbortedSkipped}, nil
+	}
```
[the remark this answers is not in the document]
vhvb1989: Fixed: hoisted the provider-agnostic `provision` dispatch out of `Manager.Deploy`/`Manager.Preview` (which each layer's own Manager runs concurrently) and into the command layer. It now runs exactly once per `azd provision` / `azd up` — in `provisionLayersGraph` (covers `--preview` + single/multi-layer) and in the up graph (guarded on `len(layers) > 0`) — before the layer execution graph. Aborts surface through the shared `wrapProvisionError`. So the checks and any warning prompt now fire a single time regardless of layer count.

cli/azd/pkg/azdext/validation_provider.go
```diff
@@ -93,6 +93,44 @@ func (c *ValidationContext) EnvLocation() (string, bool) {
...
+	v, ok := c.Data[ValidationContextSubscriptionID]
+	if !ok {
+		return "", false
+	}
+	return string(v), true
+}
+
+// ResourceGroup returns the target resource group name from a "provision" check
+// context. It returns ("", false) for subscription-scoped deployments.
+func (c *ValidationContext) ResourceGroup() (string, bool) {
```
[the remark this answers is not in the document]
vhvb1989: Fixed: corrected the `ResourceGroup()` doc comment. Since the `provision` dispatch always includes the resource group key (empty value for subscription-scoped runs), `ok` is always true for a provision context. The comment now documents that the value is empty for subscription scope and points authors at `TargetScope()` to distinguish scopes.

cli/azd/pkg/infra/provisioning/provision_validation.go
```diff
@@ -0,0 +1,226 @@
...
+
+// provisionValidationContext builds the lean, provider-agnostic context passed
+// to "provision" validation checks. It intentionally omits any ARM template or
+// parameters, since not every provider (e.g. Terraform, foundry) produces one.
+func (m *Manager) provisionValidationContext() map[string][]byte {
+	resourceGroup := m.env.Getenv(environment.ResourceGroupEnvVarName)
+	targetScope := "subscription"
+	if resourceGroup != "" {
+		targetScope = "resourceGroup"
+	}
```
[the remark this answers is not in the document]
vhvb1989: Fixed: added `TestManagerDeployProvisionValidation_SubscriptionScoped` which runs with `AZURE_RESOURCE_GROUP` unset and asserts the resource group key is present but empty and `target_scope == "subscription"`.

cli/azd/internal/cmd/provision_graph.go
```diff
@@ -84,6 +84,21 @@ func (p *ProvisionAction) provisionLayersGraph(
...
+	// sits here rather than inside Manager.Deploy/Preview: multi-layer
+	// provisioning runs each layer's Deploy concurrently through its own
+	// Manager, so dispatching per-Deploy would fire the checks — and any
+	// warning confirmation prompt — once per layer with an identical,
+	// env-scoped context. On abort, translate through wrapProvisionError so
+	// preview and deploy share the "Provisioning was cancelled." UX.
+	if abort, err := p.provisionManager.RunProvisionValidation(ctx, previewMode); err != nil {
+		return nil, err
+	} else if abort {
+		return nil, p.wrapProvisionError(ctx, provisioning.ErrProvisionValidationAborted)
```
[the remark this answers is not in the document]
vhvb1989: Fixed: updated the PR description's Summary and "What changed" bullets to match the hoisted implementation — the exported `Manager.RunProvisionValidation` is dispatched once from `provisionLayersGraph` and the up graph before the layer graph, not from `Manager.Deploy`/`Preview`. Also corrected the abort-semantics bullet accordingly.

cli/azd/pkg/infra/provisioning/provision_validation.go
```diff
@@ -0,0 +1,226 @@
...
+// the checks (and any warning confirmation prompt) firing exactly once with the
+// single, env-scoped context. Unlike the Bicep-only "local-preflight" dispatch,
+// this runs for every provider (Bicep, Terraform, and extension-provided
+// providers such as microsoft.foundry and demo).
+//
+// Findings render through the uniform preflight report and follow the standard
+// (abort, err) semantics:
+//   - abort=true, err=nil: an error-severity finding (or a declined warning)
+//     means provisioning should be skipped (exit code 0).
+//   - abort=false, err!=nil: dispatch or prompting failed to run.
```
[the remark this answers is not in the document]
vhvb1989: Fixed: narrowed the contract bullet — dispatch/timeout failures are non-fatal (logged and skipped, returning err=nil); only a failed confirmation prompt returns err!=nil. The doc now reflects that.

cli/azd/pkg/azdext/validation_provider.go
```diff
@@ -118,6 +159,24 @@ type ValidationCheckRegistration struct {
...
 }
 
+// --- Check type constants ---
+
+const (
+	// ValidationCheckTypeLocalPreflight is the check type dispatched by the
+	// Bicep provider during ARM-template preflight validation. Its context
+	// carries ARM-specific data (template, parameters, resource snapshot) and
+	// therefore only runs for Bicep-provisioned deployments.
+	ValidationCheckTypeLocalPreflight = "local-preflight"
```
JeffreyCA, line 198: Note we'll need to rename this as well as part of https://github.com/Azure/azure-dev/pull/8844

cli/azd/internal/cmd/provision_graph.go
```diff
@@ -84,6 +84,21 @@ func (p *ProvisionAction) provisionLayersGraph(
...
+	// ── provider-agnostic provision validation (once per command) ─────────
+	// Dispatch extension-registered "provision" checks a single time, before
+	// the preview / single-layer / multi-layer paths run. This deliberately
+	// sits here rather than inside Manager.Deploy/Preview: multi-layer
+	// provisioning runs each layer's Deploy concurrently through its own
+	// Manager, so dispatching per-Deploy would fire the checks — and any
+	// warning confirmation prompt — once per layer with an identical,
+	// env-scoped context. On abort, translate through wrapProvisionError so
+	// preview and deploy share the "Provisioning was cancelled." UX.
+	if abort, err := p.provisionManager.RunProvisionValidation(ctx, previewMode); err != nil {
```
JeffreyCA: [suggestion] azd-code-reviewer: This runs before the provider has initialized and resolved subscription/location/resource group, so a first-time `azd provision` without flags or existing env values dispatches `"provision"` checks with empty `subscription_id`, `env_location`, and `resource_group`. Extension checks that depend on those values can misfire on cold runs.

I think we need to either resolve the deployment context before this dispatch, or explicitly document the `"provision"` context as ambient/best-effort and allow empty values.

cli/azd/pkg/azdext/validation_provider.go
```diff
@@ -93,6 +93,47 @@ func (c *ValidationContext) EnvLocation() (string, bool) {
...
+// distinguish scopes rather than relying on ok).
+func (c *ValidationContext) ResourceGroup() (string, bool) {
+	v, ok := c.Data[ValidationContextResourceGroup]
+	if !ok {
+		return "", false
+	}
+	return string(v), true
+}
+
+// TargetScope returns the deployment target scope ("subscription" or
```
JeffreyCA, line 149: [suggestion] azd-code-reviewer: `TargetScope()` is not actually authoritative here because `provisionValidationContext()` infers it only from `AZURE_RESOURCE_GROUP`, before Bicep target scope and any resource-group prompt/resolution have run. This can report `"subscription"` for an RG-scoped template on a cold run, or `"resourceGroup"` from a stale env var for a subscription-scoped deployment.

Consider making the context authoritative before dispatch or soften these comments/docs to say resource group and target scope are best-effort ambient env values.

jongio (approved):

hemarina (approved):
Nicely executed and thoroughly documented. Hoisting dispatch to the command layer so the `provision` checks (and any warning prompt) fire **once per command** rather than once per concurrently-provisioned layer is the right call, and the reasoning is captured well in the code comments. The `validation.preflight.check_type` field cleanly resolves the shared-event double-count, and the docs/tests are in good shape.

Things I verified that are solid:
- No double-dispatch — `provisionLayersGraph` is only the `provision` entry point; `up` calls `RunProvisionValidation` exactly once.
- Both dispatch sites are guarded on non-empty layers (up via `len(layers) > 0`, provision via the `len(layers)==0` early return).
- Severity classification matches Bicep's existing logic, and `ErrProvisionValidationAborted` is correctly mapped to `ErrAbortedByUser` (exit 0) and covered by `Test_PackageLevelErrorsMapped`.

One thing to resolve before merge — +1 to @JeffreyCA's cold-run context concern: because dispatch runs before the provider resolves subscription/location/resource group, a first-time `provision`/`up` ships empty `subscription_id`/`env_location`/`resource_group`, and `TargetScope()` (inferred only from `AZURE_RESOURCE_GROUP`) can be wrong on cold runs. Either resolve the deployment context before dispatch, or document the `provision` context as best-effort/ambient and soften the `TargetScope()`/`ResourceGroup()` doc comments so they don't read as authoritative.

Otherwise this looks good.

cli/azd/pkg/infra/provisioning/provision_validation.go
```diff
@@ -0,0 +1,227 @@
...
+// Findings render through the uniform preflight report and follow the standard
+// (abort, err) semantics:
+//   - abort=true, err=nil: an error-severity finding (or a declined warning)
+//     means provisioning should be skipped (exit code 0).
+//   - abort=false, err!=nil: the confirmation prompt itself failed to run
+//     (dispatch/timeout failures are non-fatal and skipped, returning err=nil).
+//   - abort=false, err=nil: validation passed or was skipped; proceed.
+//
+// The preview flag only affects the confirmation prompt wording.
+func (m *Manager) RunProvisionValidation(ctx context.Context, preview bool) (abort bool, err error) {
```
richardpark-msft: TBH, this function signature is a bit odd to me, and having a double-valued error return can make things hard to reason about.

In the past, when I've been in this state, I've just made some custom errors instead - they're easier to reason about since it's still all one flow (ie, not check these two vars), and your chart above is less complicated.

Looking at the callers it looks liks you're just assuming 'abort' means 'different error' - might as well centralize that here.

cli/azd/pkg/infra/provisioning/provision_validation.go
```diff
@@ -0,0 +1,227 @@
...
+	// Bound extension dispatch so a blocked check cannot hang provisioning. A
+	// timeout (or any dispatch error) is treated as a non-fatal skip.
+	dispatchCtx, cancel := context.WithTimeout(ctx, provisionValidationTimeout)
+	results, invokedRuleIDs, dispatchErr := dispatcher.DispatchChecks(
+		dispatchCtx, azdext.ValidationCheckTypeProvision, checkContext,
+	)
+	cancel()
+	if dispatchErr != nil {
+		if errors.Is(dispatchErr, context.DeadlineExceeded) {
+			log.Printf(
```
richardpark-msft, line 105: Just curious - do we have a rule for when it's okay to print something to the screen directly?

cli/azd/extensions/microsoft.azd.demo/internal/project/demo_provision_validation_check.go
```diff
@@ -0,0 +1,58 @@
...
+
+	"github.com/azure/azure-dev/cli/azd/pkg/azdext"
+)
+
+// DemoProvisionValidationCheck is a sample provider-agnostic validation check.
+// Unlike DemoValidationCheck (which relies on the Bicep snapshot), this check
+// registers under the "provision" check type and therefore runs before
+// provisioning for every provider — including this demo extension's own
+// provisioning provider. It demonstrates reading the lean provision context
+// (environment, subscription, location, resource group, target scope).
```
richardpark-msft: AFAICT, this is the only place the "lean provisioning context" is defined, and it seems rather informal. It feels like it should be documented on the ValidationContext type itself, or somewhere more authoritative.

cli/azd/pkg/azdext/validation_provider.go
```diff
@@ -118,6 +159,24 @@ type ValidationCheckRegistration struct {
...
+	// ValidationCheckTypeLocalPreflight is the check type dispatched by the
+	// Bicep provider during ARM-template preflight validation. Its context
+	// carries ARM-specific data (template, parameters, resource snapshot) and
+	// therefore only runs for Bicep-provisioned deployments.
+	ValidationCheckTypeLocalPreflight = "local-preflight"
+
+	// ValidationCheckTypeProvision is the provider-agnostic check type
+	// dispatched immediately before provisioning runs, regardless of the
+	// provisioning provider (Bicep, Terraform, or extension-provided providers
+	// such as microsoft.foundry and demo). Its context is lean and carries no
```
richardpark-msft: "Its context is lean and carries no ARM template"

This implies that other things are missing as well? Otherwise, what else makes it lean?

cli/azd/pkg/azdext/validation_provider.go
richardpark-msft, line 198: Are these two things mutually exclusive, or do you see a reason why I might define both? 

I'm curious because, from a check authoring standpoint, it seems easier to write a single method that checks the "lean" attributes (assuming these exist for all providers, bicep or otherwise), and then just have it see if there's _more_ context available and, if so, do more detailed checks.

richardpark-msft (commented):
I don't have enough domain knowledge to approve this, but I had some comments.

cli/azd/pkg/azdext/validation_provider.go
vhvb1989: Acknowledged — I'll leave the rename to be handled as part of #8844 to avoid conflicting with that work. Tracking it there rather than in this PR.

cli/azd/internal/cmd/provision_graph.go
vhvb1989: Good catch. Went with option (b) in 1aa357dc6: documented the `"provision"` context as best-effort ambient values. The dispatch intentionally sits before provider init (single, fail-fast, env-scoped dispatch — resolving/prompting for sub/location that early would fight the lean design and move prompting ahead of preprovision hooks), so the constant doc and the `SubscriptionID()`/`EnvLocation()`/`ResourceGroup()`/`TargetScope()` accessors now state values may be empty on a cold run and checks must tolerate that (skip rather than fail when unset).

cli/azd/pkg/azdext/validation_provider.go
vhvb1989: Agreed — softened in 1aa357dc6. `TargetScope()` and `ResourceGroup()` docs now state they are best-effort ambient values inferred from AZURE_RESOURCE_GROUP before the provider determines the authoritative target scope, explicitly calling out that a cold run can report `"subscription"` for an RG-scoped template or a stale `"resourceGroup"` for a sub-scoped deployment, and that authors should not rely on it as definitive.

cli/azd/pkg/infra/provisioning/provision_validation.go
vhvb1989: Accepted — done in 2bd376804. `RunProvisionValidation` now returns a single `error`: `ErrProvisionValidationAborted` for an error-severity finding or a declined warning, the wrapped prompt error for a confirmation-prompt failure, and `nil` otherwise. Both callers just funnel any non-nil error through `wrapProvisionError` (which already maps the abort sentinel to `internal.ErrAbortedByUser`), so the 'abort means different error' assumption is now centralized in one place. Removed the two-var flow from the doc/diagram and updated the unit tests to assert on the sentinel.

cli/azd/pkg/infra/provisioning/provision_validation.go
vhvb1989: There isn't a hard rule. The closest thing to one: when you add or change UI messages/renders, get feedback from a UX designer — but writing to the log is always fine and doesn't need that review.

The other common guideline we follow is intent-based:
- **log** (`log.Printf`) → messages meant to help troubleshoot an issue (diagnostics). Goes to the azd log file / stderr under `--debug`, not the user's screen.
- **no log, directly to screen** (`console.Message`/`MessageUxItem`) → confirmations, information, and progress the user should see.

Here the dispatch timeout/failure is a non-fatal diagnostic we intentionally swallow (a broken extension shouldn't block provisioning), so `log.Printf` is the right channel — we don't want it on the user's screen. The user-facing output (the preflight report and the abort messages) goes through `console` instead.

JeffreyCA (approved):

cli/azd/extensions/microsoft.azd.demo/internal/project/demo_provision_validation_check.go
vhvb1989: Accepted — done in d55a9a6ce. Moved the authoritative definition onto the `ValidationContext` type doc: it now explains that `Data`'s contents depend on `CheckType` and lists the accessor groups for each — ARM-rich (`ARMTemplate`/`ARMParameters`/`ResourcesSnapshot`/`PredictedResources`) for `"local-preflight"`, and the lean, best-effort set (`EnvName`/`SubscriptionID`/`EnvLocation`/`ResourceGroup`/`TargetScope`) for `"provision"`. The demo check comment now just points at `ValidationContext`/`ValidationCheckTypeProvision` instead of redefining it.

cli/azd/pkg/azdext/validation_provider.go
vhvb1989: Good point — reworded in d415b72a7. The doc now spells out what makes it lean: it deliberately omits **all** the ARM-derived data that `"local-preflight"` carries (no ARM template, no resolved parameters, no resources snapshot, no predicted resources), because non-ARM providers don't produce them, and carries only the ambient environment values (env name, subscription, location, resource group, target scope).

cli/azd/pkg/azdext/validation_provider.go
vhvb1989: They're not mutually exclusive — you can register both, and the demo extension does exactly that (`demo_provision_warning` under `provision`, `demo_warning` under `local-preflight`).

The reason they aren't a single 'read lean, then opportunistically deepen' check is that the two types fire at **different lifecycle points for different provider sets**:

- `provision` is dispatched **once, up front, for every provider** (Bicep, Terraform, extension providers), *before* any Bicep compile/snapshot happens. So even for a Bicep project, the ARM template/parameters/snapshot simply don't exist yet at that moment — a 'is there more context?' check here would never find it.
- `local-preflight` is dispatched **later, inside the Bicep provider's Deploy, Bicep-only**, once the ARM template and resource snapshot have been produced.

So you choose based on **when** you need to run and **what** data you need: a provider-agnostic early gate → `provision`; deep ARM/Bicep inspection → `local-preflight`; register **both** if you want an early lean gate *and* a deep Bicep check.

Your unified model (always expose the lean attributes — even in the `local-preflight` context — so an author can write one check that deepens when richer data is present) is a nice future enhancement. It would mean adding the ambient env attributes to the `local-preflight` context too. I'd like to keep that out of scope for this PR, but it's a reasonable direction to consider next.

richardpark-msft: (just discussion, no need to hold the PR) 

Is there anything I would _want_ to do in provision, that wouldn't be better in local-preflight (assuming I'm bicep only?). Also, keeping with naming like `preflight`, should it be `preprovision`?

vhvb1989: If you're bicep-only, you most likely want to use local-preflight (to be renamed https://github.com/Azure/azure-dev/pull/8844).
Unless you want to run some check before building bicep...  which would give you a change to fail/stop even earlier.

But I would expect bicep-only extensions (those extensions not adding custom-providers and just using the default azd's bicep provider) to only use the local-preflight.

The new type is a better fit for extensions bringing their own rules for provisioning so they can still leverage the azd-validate feature and be consistent with the model of showing warnings and allowing user to abort.

In the demo extension, I added both, but it will be hard to see that in reality. For sure.

richardpark-msft (approved):
Looks good! I know there's more work to do, but no reason this can't go out.

PR #9024, opened 2026-07-08
feat(toolboxes): reuse existing toolbox via endpoint in azure.yaml

### LINKED ISSUE #9023: [azd.ai.toolboxs] allow azd deploy to skip deploying a new toolbox version
## Summary

Track and implement the ability for `azd deploy` to skip deploying a new toolbox version when one is already present and up-to-date.

## Motivation

Currently, `azd deploy` always deploys the toolbox version as part of the deployment process. There should be a mechanism to skip this step when the existing toolbox version satisfies the requirements, reducing unnecessary deployment time and resource usage.

## Proposed Behavior

- When running `azd deploy`, detect whether the currently deployed toolbox version is already at the desired version.
- If the version matches (or satisfies a version constraint), skip the toolbox deployment step.
- Provide a flag or configuration option (e.g., `--skip-toolbox-update`) to explicitly opt in/out of this behavior.

## Acceptance Criteria

- [ ] Detect the currently deployed toolbox version before deploying.
- [ ] Skip toolbox deployment when the version is already current.
- [ ] Expose a flag or config to control the skip behavior.
- [ ] Add unit/integration tests covering the skip logic.
- [ ] Update documentation and help text as needed.

## Notes

- Consider CI/CD scenarios where users want deterministic, idempotent deployments.
- Ensure `--no-prompt` / non-interactive mode works correctly with the skip logic.

huimiu:
Fixes: #9023 

## Summary

Adds an `endpoint` field to the `azure.ai.toolbox` service so a toolbox can be **reused** instead of creating a new version on every `azd deploy`. This mirrors the `azure.ai.project` brownfield `endpoint` convention: omit it to create, set it to reuse.

## Motivation

`toolboxServiceTarget.Deploy` unconditionally calls `CreateToolboxVersion`, so every `azd deploy` mints a new toolbox version and repoints the consumer env var `TOOLBOX_<NAME>_MCP_ENDPOINT`. There was no declarative way to reuse a toolbox that is managed elsewhere (shared across projects, or created via `azd ai toolbox create`) — the only workaround was the imperative `azd env set` that `azd ai agent doctor` already suggests.

## Behavior

- **Omit `endpoint`** → unchanged: upsert a new version from `tools`.
- **Set `endpoint`** → skip `CreateToolboxVersion`; expand `${VAR}` against the azd environment; publish `TOOLBOX_<NAME>_MCP_ENDPOINT` for agents to consume. The reuse path does not touch `projectctx` or the toolbox data-plane client — it is pure env wiring, matching project BYO.
- A toolbox version is immutable, so `endpoint` is mutually exclusive with `tools`/`description` (enforced in the schema and defensively in code).

```yaml
services:
  research-tools:
    host: azure.ai.toolbox
    endpoint: ${RESEARCH_TOOLBOX_ENDPOINT}
```

Get the value from `azd ai toolbox show <name>` (the `Endpoint:` line).

## Changes

- `schemas/azure.ai.toolbox.json`: add `endpoint`; draft-07 `if/then` disallows `tools`/`description` when `endpoint` is set.
- `internal/cmd/service_target.go`: `Endpoint` on `toolboxServiceConfig`; reuse branch in `Deploy`; new `deployReuse` and pure `resolveReuseEndpoint` (expands `${VAR}`, enforces exclusivity, rejects empty).
- `internal/cmd/service_target_test.go`, `README.md`, `CHANGELOG.md`.

`schemas/{alpha,v1.0}/azure.yaml.json` only `$ref` the extension schema, so they need no change.

## Testing

- `go build ./...`, `go vet`, `golangci-lint run` (0 issues), and `go test ./...` all pass.
- Schema mutual-exclusion validated with a draft-07 validator: `endpoint`+`tools` and `endpoint`+`description` rejected; `endpoint`-only, `tools`-only, and `tools`+`description` accepted.


jongio (approved):
Clean feature addition that mirrors the established brownfield endpoint pattern from azure.ai.project. The mutual-exclusivity enforcement is solid (schema-level if/then + defensive code check), resolveReuseEndpoint is a pure function with good test coverage, and error messages include actionable suggestions.

One minor note: the Copilot bot's comment about comment vs $comment in the schema is worth addressing for consistency with draft-07 and the rest of the repo.

cli/azd/extensions/azure.ai.toolboxes/README.md
```diff
@@ -1,3 +1,25 @@
 # Foundry Toolboxes
 
 Manage Microsoft Foundry Toolboxes from your terminal. (Preview)
+
+## Reuse an existing toolbox in `azure.yaml`
+
+A `host: azure.ai.toolbox` service normally creates a new toolbox version from
+its `tools` on each `azd deploy`. To reuse a toolbox that already exists (for
+example one shared across projects, or created with `azd ai toolbox create`),
+set `endpoint` to its MCP endpoint instead. azd then publishes that endpoint for
```
trangevi, line 10: So is this a manual process? Or how does a user decide if they want to reuse an existing one vs create a new one? If a user changes their mind later, do we have a story for that?

@therealjohn, since you're looking at azd project integration, this is likely relevant

trangevi (approved):

cli/azd/extensions/azure.ai.toolboxes/README.md
huimiu: It’s manual and declarative right now, just set the `endpoint` on the toolbox service to opt in. Switching is editing and redeploying: both modes write the same TOOLBOX__MCP_ENDPOINT, and reuse never touches the data plane, so nothing gets deleted.

huimiu:

jongio (approved):
The schema now uses $comment (draft-07 annotation keyword), which resolves the one item from my earlier review. Reuse-mode logic and tests are unchanged and still solid.

PR #9028, opened 2026-07-08
Fix phantom test failures from tool first-run spinners writing to os.Stdout

### LINKED ISSUE #9029: Phantom 'FAIL (unknown)' test results from tool first-run spinners writing to os.Stdout
## Description

Unit tests in `cmd/middleware` (and occasionally adjacent packages) intermittently report `FAIL (unknown)` even though the test itself passes. The tool first-run middleware creates its spinners via `uxlib.NewSpinner` without a `Writer`, which defaults to `os.Stdout`. On start/stop the spinner emits cursor escape codes (`\033[?25l` / `\033[?25h`) directly to stdout. Under `go test -json`, those raw bytes corrupt the event stream, so the runner misattributes a passing test as a failure - sometimes onto an unrelated test that ran concurrently.

## Symptom

```
=== FAIL: cmd/middleware TestToolFirstRunMiddleware_OfferInstall_PromptError (unknown)
25h--- PASS: TestToolFirstRunMiddleware_OfferInstall_PromptError (0.00s)
```

The `25h` prefix is the tail of the show-cursor escape `\033[?25h`.

## Root cause

- `SpinnerOptions.Writer` defaults to `os.Stdout` (`cli/azd/pkg/ux/spinner.go`).
- The detect/install spinners in `cli/azd/cmd/middleware/tool_first_run.go` do not set `Writer`.
- `forbidigo` only guards `os.Stdout` writes in `*_test.go`, so it does not catch this production-code path.

## Expected

Spinners exercised by tests should write to the console handle (`io.Discard` under `MockConsole`), keeping the `go test -json` stream clean. Related to #8385.

JeffreyCA:
Fixes #9029

This PR fixes intermittent, misattributed unit-test failures in `cmd/middleware` (and adjacent packages) caused by the tool first-run middleware spinners writing terminal control codes to real stdout during tests.

## The problem

The tool first-run middleware creates its "Detecting tools..." and "Installing tools..." spinners via `uxlib.NewSpinner` without a `Writer`, which defaults to `os.Stdout`.

- On start/stop, the spinner emits cursor escape codes (`\033[?25l` / `\033[?25h`) straight to stdout.
- Under `go test -json`, those raw bytes land between JSON events and corrupt the event stream.
- The test runner then reports a passing test as `FAIL (unknown)`, often bleeding onto whatever unrelated test ran concurrently:

```
=== FAIL: cmd/middleware TestToolFirstRunMiddleware_OfferInstall_PromptError (unknown)
25h--- PASS: TestToolFirstRunMiddleware_OfferInstall_PromptError (0.00s)
```

The `forbidigo` linter only guards against `os.Stdout` in `*_test.go`, so it does not catch this because the offending write originates in production middleware code.

## The fix

Route both spinners through `m.console.Handles().Stdout`, matching how prompts in the same file already resolve their writer.

- Under `MockConsole`, `Handles().Stdout` is `io.Discard`, so nothing leaks into the test stream.
- In production, it is the real stdout, so runtime behavior is unchanged.

## Behavior at a glance

| Scenario | Before | After |
| --- | --- | --- |
| Spinner writer in tests (`MockConsole`) | `os.Stdout` (leaks escape codes) | `io.Discard` (silent) |
| `go test -json` event stream | Corrupted by cursor codes | Clean |
| Production spinner output | Real stdout | Real stdout (unchanged) |

## Testing

- Verified the corrupting cursor code (`25h`/`25l`) appears in the `-json` stream before the change and is gone after.
- Confirmed the full `cmd/middleware` suite passes with a clean `-json` stream (no raw escape bytes, no malformed lines).

### DIFF
--- a/cli/azd/cmd/middleware/tool_first_run.go
+++ b/cli/azd/cmd/middleware/tool_first_run.go
@@ -230,6 +230,7 @@ func (m *ToolFirstRunMiddleware) runFirstRunExperience(ctx context.Context) erro
 	detectSpinner := uxlib.NewSpinner(&uxlib.SpinnerOptions{
 		Text:        "Detecting tools...",
 		ClearOnStop: true,
+		Writer:      m.console.Handles().Stdout,
 	})
 	if err := detectSpinner.Run(ctx, func(ctx context.Context) error {
 		var detectErr error
@@ -416,6 +417,7 @@ func (m *ToolFirstRunMiddleware) offerInstall(
 	installSpinner := uxlib.NewSpinner(&uxlib.SpinnerOptions{
 		Text:        "Installing tools...",
 		ClearOnStop: true,
+		Writer:      m.console.Handles().Stdout,
 	})
 	installErr := installSpinner.Run(ctx, func(ctx context.Context) error {
 		var batchErr error


jongio (approved):
Correct fix. The two spinners (detect + install) were the only ones in this file missing the Writer field. Lines 195 and 352 already use m.console.Handles().Stdout for other UX components in the same middleware. This makes it consistent and eliminates the cursor escape code leakage that corrupts go test -json output.

Production behavior is unchanged since Handles().Stdout resolves to real stdout outside of tests.

vhvb1989 (approved):

PR #9033, opened 2026-07-09
feat(extensions): validate providers declared in extension.yaml (part 1/2)

### LINKED ISSUE #8992: Add tests or linting to ensure extensions register all the providers they claim in extensions.yaml
We should add a test that loads every extension and validates that all of its providers match what's declared in extensions.yaml.

JeffreyCA:
Part of #8992

### Summary

This PR adds the SDK prerequisite for verifying that provider declarations in `extension.yaml` match the providers configured by extension code. It also corrects and protects the extension capability schemas.

### Issue

Provider metadata can drift from calls such as `WithServiceTarget` and `WithProvisioningProvider` without being detected. The incident corrected by #8901 demonstrated the impact: `azure.ai.agents` declared service-target providers that it did not register, allowing incorrect metadata to reach the extension registry.

### SDK verification helper

The extension SDK now provides `VerifyProvidersMatchManifest`, which configures an extension host without connecting to azd or invoking provider factories, then compares the accumulated registrations with `extension.yaml`.

- Every declared service-target or provisioning provider must be registered.
- Every registered service-target or provisioning provider must be declared.
- Provider names and casing must match exactly because runtime provider resolution uses exact keys.
- Duplicate declarations and registrations are rejected case-insensitively.
- Framework-service and validation registrations are excluded because they are not represented in the manifest's `providers` list.

### Demo integration

The demo extension uses the helper in a canonical verification test against its real configuration callback. Its existing test script now runs the Go test suite so the verification is exercised by the demo extension's CI.

### Capability schema consistency

The extension and registry schemas now include `validation-provider`, matching the capability already accepted by azd and declared by the demo extension. A synchronization test compares both schema enums with Go's supported capability list so future drift fails testing.

### Behavior at a glance

| Scenario | Before | After |
|---|---|---|
| An extension using the helper declares a provider that code does not register | Could pass its tests | Its verification test fails |
| An extension using the helper registers a provider missing from the manifest | Provider is undiscoverable without detection | Its verification test fails |
| An extension using the helper uses different provider name casing | Runtime lookup can fail after publication | Its verification test fails |
| Go and JSON schema capability enums drift | Schema-aware tooling can reject valid manifests | Schema synchronization test fails |

### Scope and follow-ups

This PR intentionally lands the SDK prerequisite rather than closing the complete first-party enforcement gap in #8992.

- The provider-bearing Azure AI extensions now pin published azd SDK versions that predate `VerifyProvidersMatchManifest`. After an azd SDK release contains the helper, a follow-up will update those dependencies, add their canonical verification tests, and wire the tests into extension CI.
- Central enforcement requiring every provider-bearing first-party extension to carry the canonical test is deferred to that follow-up.
- Registry provider shape, capability coherence, duplicate ownership, and complete registry-to-manifest equality checks are deferred.
- Provider-type schema synchronization and full `extension.yaml` JSON Schema validation are deferred; this PR synchronizes capability enums only.
- Build, pack, publish, runtime-registration, install-time collision, and non-Go SDK enforcement are deferred to broader extension-contract work.
- This PR does not change extension discovery, provider resolution, installation, or runtime registration behavior.

### Testing

Testing covers matching registrations, dead manifest claims, undeclared runtime registrations, duplicate declarations and registrations, exact casing, ignored registration categories, invalid callback input, the demo extension's real configuration callback, and synchronization of both capability schema enums.


JeffreyCA:
### Provider validation CI exercise

Temporary regressions were pushed and reverted one at a time using normal pushes. Times are approximate completion times in UTC.

| UTC | Scenario | Intentional regression | Clean revert |
|---|---|---|---|
| 23:05 | Manifest capability | ❌ [run 29056376419](https://github.com/Azure/azure-dev/actions/runs/29056376419) detected the missing `provisioning-provider` capability | ✅ [run 29056453301](https://github.com/Azure/azure-dev/actions/runs/29056453301) |
| 23:09 | Provider ownership | ❌ [run 29056568708](https://github.com/Azure/azure-dev/actions/runs/29056568708) detected duplicate ownership across two extensions | ✅ [run 29056640738](https://github.com/Azure/azure-dev/actions/runs/29056640738) |
| 23:15 | Runtime registration drift | ❌ [run 29056776230](https://github.com/Azure/azure-dev/actions/runs/29056776230) reported both the unregistered manifest claim and undeclared runtime registration | ✅ [run 29056902543](https://github.com/Azure/azure-dev/actions/runs/29056902543) |
| 23:21 | Canonical test guard | ❌ [run 29057048902](https://github.com/Azure/azure-dev/actions/runs/29057048902) rejected a renamed test even though the test file still existed | ✅ [run 29057158490](https://github.com/Azure/azure-dev/actions/runs/29057158490) |
| 23:26 | AST call enforcement | ❌ Central [run 29057329467](https://github.com/Azure/azure-dev/actions/runs/29057329467) rejected a non-call reference while the exact test passed in [run 29057329606](https://github.com/Azure/azure-dev/actions/runs/29057329606) | ✅ Registry [run 29057479436](https://github.com/Azure/azure-dev/actions/runs/29057479436) and agents [run 29057479514](https://github.com/Azure/azure-dev/actions/runs/29057479514) |
| 23:33 | Schema synchronization | ❌ [run 29057632586](https://github.com/Azure/azure-dev/actions/runs/29057632586) detected the missing registry-schema enum value | ✅ [run 29057684698](https://github.com/Azure/azure-dev/actions/runs/29057684698) |

✅ All temporary edits were reverted. The final tree matches the pre-exercise baseline; latest-head checks are settling.

cli/azd/extensions/extension.schema.json
```diff
@@ -1,146 +1,147 @@
...
+    },
+    "requiredAzdVersion": {
+      "type": "string",
+      "title": "Required azd Version",
+      "description": "azd core version constraint required to use this extension. Supports semantic versioning constraint expressions (e.g. \">= 1.24.0\")."
+    },
+    "capabilities": {
+      "type": "array",
+      "title": "Capabilities",
       "description": "List of capabilities provided by the extension. Supported values: custom-commands, lifecycle-events, mcp-server, service-target-provider, framework-service-provider, provisioning-provider, metadata. Select one or more from the allowed list. Each value must be unique. Not required for extension packs, which declare dependencies instead and have no executable.",
```
jongio: This capabilities enum (and the matching `versions.capabilities` enum in `registry.schema.json`) doesn't include `validation-provider`, but this PR adds `ValidationProviderCapability` ("validation-provider") to Go's `ValidCapabilities` and declares it in `microsoft.azd.demo/extension.yaml`, which pins `$schema=../extension.schema.json`.

So a schema-aware editor or validator now rejects a first-party manifest for a capability azd itself treats as valid. That's the code/schema drift this PR is meant to close.

`provisioning-provider` was added to Go, both schemas, and the demo manifest together. `validation-provider` got the Go const and the manifest entry but not the schema enums. Adding a `validation-provider` const here, to the `registry.schema.json` capabilities enum, and to this "Supported values" description closes the gap.

`TestProviderSchemaTypesInSyncWithGo` only checks provider types, not capabilities, and `TestValidateExtension_AllValidCapabilities` omits `validation-provider`, so nothing in CI catches this. A capability-enum sync test mirroring the provider-type one would.

jongio (commented):
One issue from the schema and capability changes:

- `validation-provider` is added to Go `ValidCapabilities` and to `microsoft.azd.demo/extension.yaml`, but it's missing from the capabilities enum in both `extension.schema.json` and `registry.schema.json` (inline). A schema-aware editor or validator will reject that first-party manifest for a capability azd accepts. `provisioning-provider` was added to Go, both schemas, and the demo manifest together; `validation-provider` skipped the two schemas. The new sync test only covers provider types, not capabilities, so CI doesn't catch it.

jongio (approved):
My earlier note is resolved: `validation-provider` is now present in both `extension.schema.json` and `registry.schema.json`, and `TestCapabilitySchemaTypesInSyncWithGo` guards both capability enums against Go drift in CI. The scope-reduction commit builds and vets clean here, no dangling references to the removed helpers, and all five provider extensions carry the canonical `internal/cmd/providers_manifest_test.go`.

jongio (approved):
Re-reviewed after the rebase and scope reduction. Deferring the cross-extension verification and per-extension tests until the SDK release removes the temporary replace directives, which was my main concern on the earlier iterations. The remaining change is self-contained: capability schemas stay in sync via TestCapabilitySchemaTypesInSyncWithGo, and VerifyProvidersMatchManifest is guarded by its own tests plus the demo manifest check. Provider tests, azdext tests, and go vet all pass locally.

kristenwomack (approved):
One developer-experience note, not blocking:

In `provider_manifest.go`, the declared-vs-registered comparison uses `slices.Contains` (exact match, L82 and L89), while duplicate detection lowercases names (L113). So a casing-only mismatch, like declaring a provider as `Foo` but registering it as `foo`, trips both checks and prints two errors: "declared in manifest but not registered" and "registered but not declared." Since exact casing is one of the goals here, a single "casing differs" message would tell authors what's actually wrong in one pass.

hemarina (approved):
Clean, well-scoped SDK prerequisite. `VerifyProvidersMatchManifest` is a solid building block, the demo integration exercises it against a real config callback, and the capability-schema sync test is anchored to `ValidCapabilities` (the actual validation source) so it will genuinely catch future drift. I ran the new tests locally — all pass.

Things I verified that are solid:
- Type-keyed comparison correctly handles the demo's same-name/different-type case (`demo` service-target + `demo` provisioning-provider), and untyped manifest entries (`context`, `prompt`, …) are correctly excluded.
- The case-sensitive match is justified, not overly strict: runtime resolution uses exact IoC keys (`manager.go` `ResolveNamed(string(providerKey), …)`), even though registry discovery is case-insensitive (`EqualFold`). It's documented and covered by the `CaseMismatch` test.
- `slices.Clone` in the new accessors avoids internal-slice aliasing.

Two minor, non-blocking points:

1. **Implicit lazy-factory contract.** The helper relies on `WithServiceTarget`/`WithProvisioningProvider` recording provider names *without* invoking their factories, and on `NewExtensionHost(nil)` being safe (the demo's factory closures capture a nil client). It works today, but a future SDK change that made `WithX` eager would silently break verification. A one-line comment on the SDK side ("verification depends on factories remaining lazy") would harden the contract.

2. **Cosmetic:** a case-insensitive duplicate that also differs in casing emits two lines ("declared more than once" + "declared but not registered") for one underlying issue. Each is technically true, just slightly noisy.

Otherwise this looks good — approving.

cli/azd/pkg/azdext/provider_manifest.go
```diff
@@ -0,0 +1,150 @@
...
+)
+
+// manifestProviderDoc is the minimal subset of an extension.yaml manifest needed to
+// compare declared providers against what an extension registers at runtime.
+type manifestProviderDoc struct {
+	Providers []extensions.Provider `yaml:"providers"`
+}
+
+// manifestComparedProviderTypes is the set of provider types representable in a
+// manifest's providers: list. Framework service and validation providers are
```
richardpark-msft: Nit: wrap `providers:` in backticks so it reads as a YAML key reference

```suggestion
// manifest's `providers:` list. Framework service and validation providers are
```

richardpark-msft (approved):
Just a minor NIT. I guess the "big" challenge now is just that we have to make sure everyone adds this test to their test suite, but that shouldn't be a huge issue. Perhaps we can standardize the test name, so we can do a simple scan to ensure everyone's got it.

JeffreyCA:
Thanks, addressed all the feedback above

cli/azd/extensions/extension.schema.json
```diff
@@ -112,7 +112,7 @@
     "capabilities": {
       "type": "array",
       "title": "Capabilities",
-      "description": "List of capabilities provided by the extension. Supported values: custom-commands, lifecycle-events, mcp-server, service-target-provider, framework-service-provider, provisioning-provider, metadata. Select one or more from the allowed list. Each value must be unique. Not required for extension packs, which declare dependencies instead and have no executable.",
+      "description": "List of capabilities provided by the extension. Supported values: custom-commands, lifecycle-events, mcp-server, service-target-provider, framework-service-provider, provisioning-provider, validation-provider, metadata. Select one or more from the allowed list. Each value must be unique. Not required for extension packs, which declare dependencies instead and have no executable.",
       "minItems": 1,
```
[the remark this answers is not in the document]
JeffreyCA: This is already addressed

jongio (approved):
Re-reviewed the "Address comments" commit. The case-sensitivity handling in VerifyProvidersMatchManifest is sound: declared-vs-registered names differing only in case now report a specific case-sensitivity mismatch instead of a misleading declared/not-registered pair, and duplicate keys are excluded from the exact/fold comparison so a duplicated name is reported once by duplicatedNames rather than twice. The registered loop correctly skips fold-matches to avoid double-reporting the same case mismatch. The new lazy-factory assertion confirms configure records names without invoking factories against a nil azd client, which is what the bare ExtensionHost path relies on.

PR #9044, opened 2026-07-09
Fix agent run/invoke readiness race on localhost (#8411)

### LINKED ISSUE #8411: [ext-agents] azd ai agent run returns before localhost listener is ready, breaking immediate invoke --local
## Summary

`azd ai agent run` returns control to the user before the agent's HTTP listener on `localhost:8088` is actually accepting connections. As a result, the immediate `azd ai agent invoke --local` call that the quickstart prescribes fails with `could not connect to localhost:8088 -- is the agent running?` -- even though the agent **is** running, it just hasn't bound the listener yet.

Both coding-agent runs of [Quickstart: Deploy your first hosted agent](https://learn.microsoft.com/azure/foundry/agents/quickstarts/quickstart-hosted-agent?pivots=azd) hit this on first attempt. Each agent had to invent a retry/sleep workaround.

## Repro

In a fresh directory, follow the AZD pivot of the quickstart:

```
azd env new my-agent
azd env set AZURE_SUBSCRIPTION_ID `"<id>`"
azd env set AZURE_LOCATION `"eastus2`"
azd ai agent init -m `"<basic-responses-manifest-url>`" --no-prompt
azd provision --no-prompt
azd ai agent run                                             # background terminal
azd ai agent invoke --local `"Write a haiku.`"              # <-- fails on first try
```

Observed first-invoke error:

```
ERROR: could not connect to localhost:8088 -- is the agent running?
Start it with: azd ai agent run
```

Retrying after ~20-30 s succeeds. The startup of the agent (Python venv + pip install + Foundry handshake) typically takes 30-60 s on first run; subsequent runs are faster but the listener still binds a few seconds after `agent run` returns.

## Evidence from coding-agent friction reports

* Session 3a09ba08 -- F-7: "agent WAS running -- it just hadn't finished startup. `azd ai agent run` does not emit a clear 'ready' signal."
* Session 306ff310 -- friction #5: "First local invoke produced `ERROR: could not connect to localhost:8088`. The agent was running... Retrying ~20 s later worked."

## Proposed fix (either is sufficient)

1. **`azd ai agent run` should block until the listener is TCP-ready** before printing its "ready" message and releasing the terminal (or before its background mode considers itself started). The simplest implementation: after spawning the Python process, poll-connect `localhost:8088` until accept succeeds or a timeout (e.g., 90 s) elapses.
2. **`azd ai agent invoke --local` should retry connection-refused for ~30-60 s with backoff** before surfacing the error. This is the more defensive option and also covers the case where `invoke` is run before `run` reaches the connect-ready state.

Either change makes the quickstart deterministic for both humans typing the next command and coding agents running the steps non-interactively.

## Workaround (currently in the quickstart)

The quickstart now tells users to retry every 10 s for up to 2 minutes if they hit `could not connect to localhost:8088`. This is a doc workaround -- the CLI behavior is what should change.

## Related

* #6819 -- non-interactive prompts (covers a different facet of the same coding-agent UX gap)
* #8250 -- invoke error message when no agent is deployed

m5i-work:
## Why

`azd ai agent invoke --local` could fail immediately with `could not connect to localhost:8088` when run right after `azd ai agent run`. The agent's local HTTP listener isn't accepting connections the instant `run` reports progress, so a quick follow-up `invoke` (common in quickstarts and coding-agent automation) hit a connection-refused error and gave up. See #8411.

## Approach

Two complementary fixes, matching the two proposals in the issue:

- **`invoke --local` retries connection-refused dials (primary fix).** Refused connections are retried with exponential backoff (500ms -> 2s, 60s overall budget), printing a one-time `Waiting for the local agent to accept connections...` note to stderr. Only connection-refused is retried; HTTP-level errors and context cancellation return immediately so real failures aren't masked. This is the load-bearing fix because it also covers the non-interactive/automation path, where `run`'s readiness isn't observable to a separate `invoke` process.
- **`run` raises `portReadyBudget` from 5s to 90s.** The old 5s ceiling could expire before the listener bound on a cold start, silently suppressing the `Agent ready` hint. The budget only governs a background goroutine that prints a readiness hint and honors context cancellation, so a larger value is effectively free. The rationale comment was also corrected: the window spans process-start to port-bind, not venv/pip setup (which runs earlier, before the server process is spawned).

## Notes for reviewers

- Windows surfaces refused dials as `WSAECONNREFUSED` (10061) rather than POSIX `ECONNREFUSED`, and the stdlib doesn't export it. Connection-refused detection is split into build-tagged helpers (`invoke_connrefused_windows.go` / `_other.go`) to handle both.
- The 90s `portReadyBudget` is backed by empirical measurement: process-start to accept was ~3.2s warm / ~5.4s cold for a minimal Foundry sample, and ~8s once heavier frameworks (`agent_framework`) are imported. 90s leaves headroom for slow imports and CI while still bounding a genuinely stuck boot.

## Testing

- Added unit tests (`invoke_retry_test.go`) covering connection-refused detection, immediate success, HTTP-error pass-through, retry-until-ready, and context cancellation.
- Verified end-to-end locally against a real Foundry sample agent: confirmed the happy-path round-trip connects, and reproduced the #8411 race (fired `invoke` against a refused port, brought the agent up a few seconds later) to confirm it now waits and connects instead of failing immediately.

Fixes #8411


trangevi (approved):

trangevi:

trangevi:
@copilot move the cspell changes to the agent specific cspell file

trangevi (approved):

trangevi:

trangevi (approved):

therealjohn (approved):

PR #9045, opened 2026-07-09
UX: implement new design of azd tool (install/upgrade/list/check)

### LINKED ISSUE #8990: UX: implement new design of azd tool

hemarina:
Fix #8990
Figma: https://www.figma.com/design/3Fn9kwKZfODgwoq7H3yMSx/Azd-tool-commands?node-id=1-2&p=f&t=tr6lqYfeAox5Y6sn-0

Implements the new UX design for the `azd tool` command family, making agent/skill
terminology, progress rendering, and version reporting consistent across `install`, `upgrade`,
`list`, and `check`.

### `azd tool install` (skills)
- Rename `--host` → `--agent` across flags, descriptions, and messages. Split the internal notion
  of an agent host into a display `Host` (e.g. "GitHub Copilot CLI") and an exec `Command`
  (e.g. `copilot`), so commands run against the binary while the UI shows the friendly name.
- Show available agent flags in a `Tip:` line, and add a multi-select prompt to choose agent(s).
- Render each targeted agent as a live step (like `azd provision`), keeping the spinner pinned
  while the host CLI streams output or prompts interactively.

### `azd tool upgrade`
- Add a `--all` flag and an interactive multi-select of installed tools, mirroring
  `azd tool install`.
- Detect "already up to date" for **all** tools (not just skills) by comparing the version
  detected before vs. after the upgrade.
- Report the resulting version in the done message, e.g. `Tool is upgraded to v2.0.0.` or
  `Tool is already up to date (v1.1.75).`

### `azd tool list` / `azd tool check`
- Prefix skill rows with the agent name and expand a skill into one row per installed host.
- Responsive columns collapse gracefully on narrow terminals; `check` prints upgrade hints
  (`azd tool upgrade <id>` / `azd tool upgrade --all`).

## Testing
- `go build ./...`
- `go test ./cmd/... ./pkg/tool/...`
- `gofmt`, `golangci-lint`, `cspell`
- Regenerated `TestUsage` / `TestFigSpec` snapshots.


hemarina:
azd tool install
image
image
image
Need further confirmation with @hyoshis about if we want to hide the printed message from agents. 

azd tool upgrade
image
image

azd tool uninstall
image

azd tool check
image

azd tool list
width >100
image

60 < width < 100
image

jongio (approved):
Solid UX overhaul. The Host/Command split is a clean design that separates display names from executable identities. Step spinner integration follows established patterns from azd provision, and the multi-select filter threshold is a nice touch for shorter lists.

The existing inline comments from the prior review cover the notable edge cases well (--no-prompt handling for multi-select, lineWriter thread safety, case sensitivity consistency). Those are worth addressing but don't block merge.

Test coverage for the new paths looks good.

JeffreyCA (commented):
Looks great, one minor comment: is it possible to fix the wrapping on narrower terminal widths?

Image

kristenwomack (approved):
Thanks Marina, this is really nice work — the `Host`/`Command` split is clean and the comments make the intent easy to follow.

One small question: in the plugin output like `Plugin "azure" updated from 1.1.73 to 1.1.86` — should "azure" be capitalized as "Azure"?

cli/azd/pkg/tool/manifest.go
```diff
@@ -54,8 +54,14 @@ type Checksum struct {
...
 type SkillHost struct {
-	// Host is the binary name of the agent CLI (e.g. "copilot", "claude").
+	// Host is the agent's display name, shown to the user and matched
+	// (case-insensitively) by --agent (e.g. "Copilot", "Claude").
 	Host string
+	// Command is the agent CLI's executable name, used to run plugin
+	// commands and version probes (e.g. "copilot", "claude"). It must be
+	// the real, case-correct binary name so exec works on case-sensitive
+	// filesystems (Linux). When empty it falls back to Host.
+	Command string
```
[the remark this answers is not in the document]
hemarina: Add test guard to make sure host.Command cannot be empty.

cli/azd/cmd/tool.go
```diff
@@ -1366,8 +1647,12 @@ func (a *toolUninstallAction) resolveToolIds(ctx context.Context) ([]string, err
 	// --all selects every installed tool. --dry-run does the same without
 	// prompting: a preview never mutates anything, so it defaults to all
 	// installed tools (a skill is previewed against the host(s) it is
-	// installed through) instead of asking the user to pick.
-	if a.flags.all || a.flags.dryRun {
+	// installed through) instead of asking the user to pick. A non-interactive
+	// terminal or --no-prompt likewise skips the picker and defaults to every
+	// installed tool (matching --all and `azd tool upgrade`), so automation
+	// never blocks or errors on input.
+	if a.flags.all || a.flags.dryRun ||
```
jongio: Mirroring `--all` here makes `azd tool uninstall` destructive by default: with no positional args, any non-interactive/CI context or `--no-prompt` now removes every installed tool. Install and upgrade can default this way safely because they add; uninstall deletes. A cleanup script that runs a bare `azd tool uninstall` (no args) expecting a prompt or a no-op will instead wipe everything.

Consider requiring an explicit target for the destructive case: in non-interactive mode with no args and no `--all`, error with something like `specify tool IDs or --all` instead of silently selecting all. That keeps automation non-blocking without turning a bare invocation into a full wipe.

jongio (commented):
Incremental re-review of the 10 new commits since my approval. One concern on the uninstall default in non-interactive mode, inline.

cli/azd/cmd/tool.go
```diff
@@ -1194,9 +1459,9 @@ func newToolUninstallFlags(cmd *cobra.Command) *toolUninstallFlags {
 		&flags.all, "all", false, "Uninstall all installed tools",
 	)
 	cmd.Flags().StringSliceVar(
-		&flags.hosts, "host", nil,
-		"Uninstall the skill from the specified agent host(s): copilot, claude. "+
-			"Use --host all (or omit --host) to remove the skill from every host it is "+
+		&flags.hosts, "agent", nil,
+		"Uninstall the skill from the specified agent(s): copilot, claude. "+
+			"Use --agent all (or omit --agent) to remove the skill from every agent it is "+
 			"installed through (skill tools only)",
```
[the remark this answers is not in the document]
hemarina: We will use --agent as few users is using tool command with no blog post out yet.

cli/azd/cmd/tool.go
```diff
@@ -1103,16 +1268,116 @@ func (a *toolUpgradeAction) Run(ctx context.Context) (*actions.ActionResult, err
...
+	})
+	if err := spinner.Run(ctx, func(ctx context.Context) error {
+		var detectErr error
+		statuses, detectErr = a.manager.DetectAll(ctx)
+		return detectErr
+	}); err != nil {
+		return nil, fmt.Errorf("detecting installed tools: %w", err)
+	}
+	return statuses, nil
+}
```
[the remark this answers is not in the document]
hemarina: No need to do that. tool list support -o --json but upgrade does not support that.

cli/azd/cmd/tool.go
```diff
@@ -1366,8 +1647,12 @@ func (a *toolUninstallAction) resolveToolIds(ctx context.Context) ([]string, err
...
 	// prompting: a preview never mutates anything, so it defaults to all
 	// installed tools (a skill is previewed against the host(s) it is
-	// installed through) instead of asking the user to pick.
-	if a.flags.all || a.flags.dryRun {
+	// installed through) instead of asking the user to pick. A non-interactive
+	// terminal or --no-prompt likewise skips the picker and defaults to every
+	// installed tool (matching --all and `azd tool upgrade`), so automation
+	// never blocks or errors on input.
+	if a.flags.all || a.flags.dryRun ||
+		!a.console.IsSpinnerInteractive() || a.console.IsNoPromptMode() {
```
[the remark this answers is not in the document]
hemarina: This is a good point. If it is no tty or --no-prompt case, we should return an error and ask the command to run with azd tool uninstall <tool-id> or azd tool uninstall --all. Same thing for upgrade/install in no tty or no prompt case.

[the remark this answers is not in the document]
jongio: Agree, returning an error in the non-interactive/`--no-prompt` case is the right call for all three (install, upgrade, uninstall). One idiom to fold in: return it as an `internal.ErrorWithSuggestion` so the fix is actionable, with the Suggestion listing the concrete commands (e.g. `azd tool uninstall <tool-id>` or `azd tool uninstall --all`). That keeps the destructive scope explicit and hands the user the exact next step instead of a bare error.

jongio (approved):
The non-interactive path now returns an actionable ErrorWithSuggestion for install, upgrade, and uninstall, and uninstall no longer treats `no target` as `all` without an explicit tool ID or --all. That resolves the destructive-default concern from my earlier comment. Deciding `already up to date` by comparing the detected version before and after the upgrade (falling back to prose only when a version is unavailable) is more reliable than parsing host CLI wording, and returning the earlier host match when a later probe errors avoids losing a real detection.

jongio (approved):
Reviewed the `address feedback` commit. The detector change now records a later host's probe error even when an earlier host was found; verified no consumer treats detection `Error` as `not installed` (all key off `Installed`), so this restores the `ToolStatus.Error` contract without reintroducing the drop. The `WithInput` stdin threading correctly answers host prompts on the console-owned stream with an `os.Stdin` fallback, and the `supported hosts` list now shows command names that match `findSkillHost`. Build and tool/cmd tests pass.

JeffreyCA (approved):
Thanks for the updates! Quick question, was it intentional the agent name was updated to use a longer name?

Image

Personally I found the shortened `[Claude]` and `[Copilot]` in the spec was fine and easier to read on narrower terminal widths.

hemarina:
> Thanks for the updates! Quick question, was it intentional the agent name was updated to use a longer name?
> 
> Image
> Personally I found the shortened `[Claude]` and `[Copilot]` in the spec was fine and easier to read on narrower terminal widths.

Yes, it was intentionally. It does look better for shortened names in narrower terminal widths. Will update it.

cli/azd/cmd/tool.go
```diff
@@ -472,9 +496,9 @@ func newToolInstallFlags(cmd *cobra.Command) *toolInstallFlags {
 		&flags.all, "all", false, "Install all recommended tools",
 	)
 	cmd.Flags().StringSliceVar(
-		&flags.hosts, "host", nil,
-		"Install the skill for the specified agent host(s): copilot, claude. "+
-			"Use --host all for every detected host (skill tools only)",
+		&flags.hosts, "agent", nil,
```
RickWinter: This renames `--host` to `--agent` across install/upgrade/uninstall with no alias, so an existing `azd tool install <skill> --host copilot` now fails with an unknown-flag error. `azd tool` isn't alpha-gated, so this is a user-visible breaking change and, per this repo's docs conventions, wants a CHANGELOG.md entry. Consider keeping `--host` as a hidden deprecated alias for one release so in-flight scripts and docs don't break the moment this merges.

RickWinter (commented):
This reworks the `azd tool` UX: it splits `SkillHost.Host` (display name) from a new `SkillHost.Command` (exec identity) so detection and PATH checks run against the real binary, renames `--host` to `--agent`, adds a live step spinner, `--all`, interactive multi-selects, per-host skill rows in list/check, and an 'already up to date' state via before/after version diffing. The approach is sound and the test coverage is thorough; security and correctness look clean. The Command/Host split is applied consistently and enforced by a new test, and the detector now records partial results plus an error rather than aborting on a later host's failure, which is an improvement.

The one thing worth resolving before merge is process, not code. `--host` is renamed to `--agent` with no alias, and the no-args `upgrade`/`uninstall` paths now return an error under `--no-prompt`/non-TTY instead of acting on all installed tools. Both are user-visible and CI-affecting, and neither a CHANGELOG entry nor a deprecation alias is included. Adding the CHANGELOG entry (and ideally a hidden `--host` alias for a release) covers it. Nothing blocks merge.

cli/azd/cmd/tool.go
hemarina: Since there's no blog post for skills, very few users are using it and no users uses --host. I prefer to just switch to --agent. 
image

jongio: Verified the `--host` to `--agent` rename is a user-visible breaking change. `--host` shipped in the released 1.26.0 and 1.27.0 on the non-alpha `azd tool` command, so existing `azd tool install <skill> --host copilot` invocations, along with any published docs or scripts, will fail with an unknown-flag error once this merges. Two things before merge: keep `--host` as a hidden deprecated alias for one release (register it alongside `--agent` and call `Flags().MarkHidden("host")`), and add a CHANGELOG.md entry under 1.28.0-beta.1 Breaking Changes. I'd missed this in my earlier approvals.

jongio (commented):
One follow-up on RickWinter's rename comment, with the release evidence inline. This needs a deprecation alias plus a CHANGELOG entry before merge.

cli/azd/cmd/tool.go
jongio: Follow-up on my earlier `--host` rename note. The CHANGELOG breaking-change entry is in now, thanks. The hidden `--host` alias still isn't, though.

`--host` shipped GA in 1.26.0 and 1.27.0 on this non-alpha `azd tool install`, so existing `azd tool install <skill> --host copilot` calls in scripts and docs will fail with an unknown-flag error once this merges.

If the team is fine with a hard break on this recently-shipped flag, the CHANGELOG covers it and this is good to go. If not, keeping `--host` as a hidden deprecated alias for one release (register it alongside `--agent` plus `Flags().MarkHidden("host")`) is a cheap way to keep those callers working. Your call, not blocking.

jongio (commented):
Re-reviewed the 3 new commits since my last pass. The uninstall live-streaming and the marketplace-add stdin/writer threading check out, and the tool skill, uninstall, and manifest tests pass locally. The manifest agent-name shortening is display-only (matching still keys on Command), so it's safe for already-installed state. One open item from my earlier `--host` rename note, inline.

cli/azd/pkg/tool/installer.go
```diff
@@ -178,6 +197,177 @@ func WithAllAvailableHosts() InstallOption {
...
+func skillCommandRunArgs(base exec.RunArgs, out io.Writer, stdin io.Reader) exec.RunArgs {
+	if out == nil {
+		return base.WithInteractive(true)
+	}
+	if stdin == nil {
+		// No input reader supplied (e.g. a non-cmd caller); fall back to the
+		// process terminal so prompts remain answerable.
+		stdin = os.Stdin
+	}
+	return base.WithStdIn(stdin).WithStdOut(out).WithStdErr(out)
```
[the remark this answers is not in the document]
hemarina: Intend to remove WithInteractive(true).

cli/azd/cmd/tool.go
hemarina: Intended to remove all --host. https://github.com/Azure/azure-dev/pull/9045#discussion_r3573191929

jongio (commented):
Re-reviewed at HEAD e4337d1, no new commits since my last pass, just your two replies. On removing `--host`: your call, the CHANGELOG covers the break, no objection from me. The one thing I'd still verify before this merges: dropping `WithInteractive(true)` for the spinner path routes the host CLI's stdout/stderr through a piped `io.MultiWriter` instead of the terminal, while stdin stays wired to the console's input. CLIs that gate their trust or confirmation prompts on an interactive stdout (Copilot, Claude) can skip or fail those prompts even with stdin attached. Since you're intentionally dropping the interactive path here, please confirm the real host CLIs still surface and accept those prompts through the piped streams, especially the marketplace-add trust prompt. Not blocking.

jongio (approved):
Re-reviewed at HEAD e4337d1. No new commits since my last pass; the only new activity is the automated inline batch, so a few notes on those so they don't cost you time.

The `for range writers` comment is a false positive here. go.mod targets go 1.26.4, and ranging over an integer has been valid since 1.22, so that test compiles fine. No change needed.

The `lineWriter` chunk-splitting note is also fine as-is. Emitting content per write rather than reassembling strict lines is intentional and documented: output is surfaced as it arrives so an interactive prompt with no trailing newline stays visible. Buffering whole lines would defeat that.

The generic `Host` display names, the singular success headers on `--all`, and the CHANGELOG covering only `install` are optional polish, your call.

My one earlier open item still stands and is non-blocking: with `WithInteractive(true)` dropped on the spinner path, host CLI stdout/stderr route through a piped writer while stdin stays attached. Please confirm the real host CLIs (Copilot, Claude) still surface and accept their trust/confirmation prompts through those piped streams. Approving.

kristenwomack:
hey @hemarina (and @hyoshis!)

This is a great redesign. The whole set of `tool` commands finally feels consistent, and the live per-step progress plus the version in the result line are exactly what people ask for after an install or upgrade.

A few usability and copy notes:

- The agent label reads as "GitHub" in the tables, the picker, and the error messages (for example `[GitHub] Azure Skills` and "Select the agent(s)"). "GitHub" is the company, not the agent, so it's ambiguous. Could we use "GitHub Copilot" or "Copilot CLI" (and "Claude Code")? That also matches the doc comments on `SkillHost.Host`.

- Should we say "update" or "upgrade"? Right now `check` surfaces "update available" while the action and hints say `upgrade`. Picking one and using it everywhere would make the flow easier to follow. I lean toward "upgrade" since that's the command, but happy to go the other way if there's a reason.

- Small grammar one on the success line: "Azure Skills in GitHub are already up to date" should be "is".

- The install header "Your tool is installed." reads singular even when someone installs several at once. Could we pluralize based on count?

Can we please fix these before merging? These show up across every `tool` surface, so I'd like them settled before release.

Thanks, Marina. Nice work on this.

cli/azd/pkg/tool/installer.go
```diff
@@ -146,36 +155,247 @@ type Installer interface {
...
+	emit func(string)
+}
+
+// Write accumulates p and emits every complete, newline-terminated line via
+// emit (without the trailing newline), preserving empty lines. Any bytes after
+// the final newline are retained for the next Write or Flush.
+func (l *lineWriter) Write(p []byte) (int, error) {
+	l.mu.Lock()
+	defer l.mu.Unlock()
+	l.buf = append(l.buf, p...)
```
jongio, line 361: The switch to line-buffering fixes a real bug: `os/exec` can split one logical line across several `Write` calls, and the old per-write emit rendered those as separate lines. That part is correct.

The tradeoff is that it breaks the live interactive-prompt behavior this code documents. `renderSkillStep` states (lines 246-251) that with `streamOutput=true` an interactive prompt "is printed above the spinner" so the user can answer it via connected stdin. A prompt usually has no trailing newline, so it now stays in `l.buf` and isn't emitted until `Flush()`, which only runs at line 289 after `work(out)` returns. While the agent CLI blocks on stdin, the user sees a pinned spinner and no prompt text.

Buffering and the live-prompt promise are now mutually exclusive. Two ways to make them consistent: flush the partial buffer when the command is idle/waiting so a no-newline prompt still shows, or (if piped stdout means the agent CLIs never actually prompt in this path) drop the interactive-prompt claim from the doc comment so the contract matches the behavior.

jongio (commented):
One follow-up on the new `lineWriter` buffering, inline. The rename itself reads consistently across the tool commands.

hyoshis:
Thanks @hemarina  for the changes! It looks great! Few comments. 

### 1. Let's use "upgrade" instead of "update" everywhere. So states should say "Upgrade available". 

### 2. When uninstalling azure-skills of Copilot, it shows extra message of "Successfully uninstalled plugin: azure (scope: user)" but this is redundant because we show the message right above it with the "Done" state. Can we remove this line? 
image


### 3. The purple checkmarks are rendered on top of the square brackets. Can we use the same checkmark we use when showing "Done" state?
image

### It should look like below screen:

image

hemarina:
Addressed all feedback. 
image
image

vhvb1989 (approved):

cli/azd/pkg/tool/installer.go
```diff
@@ -146,36 +155,251 @@ type Installer interface {
...
+		for _, line := range buffered {
+			renderer.Message(ctx, output.WithGrayFormat("  %s", line))
+		}
+	}
+	return err
+}
+
+// lineWriter buffers agent CLI output and hands each complete line to emit.
+// Skill agent commands use it to surface the agent CLI's output through the
+// console so it prints above the step spinner (which the console re-renders
```
jongio, line 307: This comment still describes the old streaming behavior. Output is now buffered and printed as a gray section below the resolved step line after StopSpinner, so it no longer prints above the spinner or keeps it visible. Worth updating this wording (and the `surface the agent CLI's output ... above the step spinner` line just above) to match.

jongio (approved):
Re-reviewed the 5 new commits since my last pass: the Host to Agent rename, the upgrade wording, JSON-mode output hygiene, the tool-IDs-plus-all guard, and the buffered gray output section. The rename is consistent across code, tests, snapshots, and telemetry docs, and lint plus cross-build checks are green. My earlier --host/--agent item is resolved and the spinner path is now documented as non-interactive. One stale doc comment inline, non-blocking.

PR #9046, opened 2026-07-09
feat(agents): create Foundry connections at provision time

### LINKED ISSUE #8350: Changes to agent.yaml tool connections are not updated during azd provision
also if I change any property of tools in manifest, azd provision does not take into account as azure.yaml is not updated.

hund030:
## Summary

Foundry project connections declared via `host: azure.ai.connection` were only ever created at `azd deploy` (and only for `None`/`ApiKey`/`CustomKeys`), so OAuth2 and identity-based connections were silently skipped and `azd provision` alone never produced a usable connection. This moves connection creation to provision time via Bicep, for both greenfield and brownfield projects, and makes `azd deploy` a no-op for connections.

## Changes

- **`synthesis/synthesizer.go` + new `modules/connections.bicep`**: `Synthesize`/`BrownfieldConnections` collect `host: azure.ai.connection` services into a `connections` param, resolving `${VAR}` on provision and preserving it on `--infra` eject; supports every auth type, not just the deploy-time subset.
- **`azure.ai.connections/service_target.go`**: `Deploy` is now a no-op -- connections are provisioned by `microsoft.foundry`, mirroring `azure.ai.project`'s existing no-op. Removed the now-dead upsert helpers and their tests.
- **`foundry_provisioning_provider.go`**: repaired `AZURE_AI_PROJECT_CONNECTION_NAMES` casing (`canonicalOutputNames`) and fixed a misleading brownfield progress message that claimed "reconciling model deployments" even with zero deployments declared (`brownfieldReconcileMessage`).
- **`init_infra.go`**: excludes `brownfield.bicep`/`brownfield.arm.json` from `--infra` eject -- dead weight in a greenfield tree, never compiled or read by the brownfield path.


cli/azd/extensions/azure.ai.connections/internal/cmd/service_target.go
```diff
@@ -104,192 +84,26 @@ func (p *connectionServiceTarget) Publish(
...
+// azure.yaml stops azd managing it but does not delete it (use
+// `azd ai connection delete`).
 func (p *connectionServiceTarget) Deploy(
 	ctx context.Context,
 	serviceConfig *azdext.ServiceConfig,
 	serviceContext *azdext.ServiceContext,
 	targetResource *azdext.TargetResource,
 	progress azdext.ProgressReporter,
 ) (*azdext.ServiceDeployResult, error) {
-	cfg, err := parseConnectionServiceConfig(serviceConfig)
```
trangevi, line 119: This feels like the wrong direction. I understand moving it out of deploy, but why is it not moved to somewhere else in this extension? We should be moving towards the individual extensions controlling their resources, not the other way around

trangevi (changes requested):

cli/azd/extensions/azure.ai.connections/internal/cmd/service_target.go
hund030: This PR moves connection creation into `azure.ai.agents`' Bicep synthesizer because a Foundry connection is a real ARM resource, not a data-plane construct like a toolbox/skill. The Bicep synthesizer's job is exactly to provision the ARM resources a Foundry project need. This also enables IaC flow and user can get the bicep with --infra flag.

`azure.ai.connections` still owns the imperative CLI and full CRUD. What moved is narrowly the infrastructure provision path regardless of which extension exposes a CLI over them.
Open to discussing further if you still feel strongly.

trangevi (approved):

PR #9048, opened 2026-07-09
fix(routines): extend routines HTTP timeout configuration

### LINKED ISSUE #8419: [azure.ai.routines] `recurring` trigger create fails with `context deadline exceeded` — `TryTimeout: 30s` is too tight for cold AgentIdentity bind
## Summary

Creating a routine with `--trigger recurring` against a Foundry project consistently fails with `context deadline exceeded`. The CLI's per-try HTTP timeout (`30s`) is shorter than the service-side latency of the first recurring-routine PUT for a given agent, and there's no way to override it from the CLI.

## Environment

- Extension: `azure.ai.routines` `0.0.1-preview` (built from `main`, May 27 2026)
- `azd` 1.23.9, Windows/amd64
- Project endpoint: westus2 Foundry project (`*.services.ai.azure.com/api/projects/<proj>`)
- Agent: prompt-kind hosted agent with provisioned `AgentIdentity` (timer routines on the same agent succeed end-to-end)

## Repro

```pwsh
azd ai routine create e2e-routine-recurring `
    --trigger recurring --cron "*/5 * * * *" --time-zone "UTC" `
    --action agent-response --agent-name <agent-with-AgentIdentity> `
    --description "Every 5 min." --force -p $PE -o json
```

Result (3 attempts, ~10 min apart):

```
ERROR: create_routine: HTTP request failed:
  Put ".../routines/e2e-routine-recurring?api-version=v1":
  context deadline exceeded
```

Follow-up checks confirm the routine never lands server-side:

```pwsh
azd ai routine show e2e-routine-recurring -p $PE -o json
# ERROR: get_routine: routine "e2e-routine-recurring" not found.

# REST GET /routines?api-version=v1 also returns empty data[].
```

For comparison, an identical `--trigger timer` create against the same agent and project succeeds in ~1–2 seconds end-to-end (create → dispatch → `Finished` run).

## Root cause

[`cli/azd/extensions/azure.ai.routines/internal/pkg/routines/client.go`](https://github.com/Azure/azure-dev/blob/main/cli/azd/extensions/azure.ai.routines/internal/pkg/routines/client.go):

```go
func NewClient(endpoint string, cred azcore.TokenCredential) *Client {
    clientOptions := &policy.ClientOptions{
        Transport: newHTTPClient(),
        ...
        Retry: policy.RetryOptions{
            MaxRetries: 1,
            TryTimeout: 30 * time.Second,   // ← per-try ceiling
        },
        ...
    }
```

`policy.RetryOptions.TryTimeout` wraps each attempt in a `context.WithTimeout(30s)`, so any single `PutRoutine` PUT is forcibly cancelled at 30 s regardless of the user's context. With `MaxRetries: 1` the wall-clock budget is roughly 60 s before the user sees `context deadline exceeded`.

The first recurring-routine PUT for a given `agent_name` apparently triggers AgentIdentity binding + scheduler registration on the back-end, which exceeds 30 s on at least some accounts/regions. The connection is torn down before the server commits, so the routine never persists.

## Suggested fixes

Prefer (1) and (2) together:

1. Raise `TryTimeout` for write operations on recurring (e.g. `2 * time.Minute`) — or remove `TryTimeout` from the pipeline and rely on `ResponseHeaderTimeout` (`60s`) + caller-supplied context.
2. Expose a `--timeout` / `AZURE_AI_ROUTINES_HTTP_TIMEOUT` knob so users can override per-call.
3. Longer term, service should return `202 Accepted` + polling for recurring create so the response isn't blocked on scheduler bind.

## Workarounds today

None from the CLI — `TryTimeout` is not configurable. Users have to rebuild the extension with a larger value.

## Related

Timer trigger end-to-end works fine on the same project / same agent in this run, so this is specific to the recurring code path (or its server-side cost), not a general connectivity issue.

huimiu:
## Summary
- Fixes #8419 by raising the routines client default request timeout to two minutes.
- Adds a root `--timeout` flag and `AZURE_AI_ROUTINES_HTTP_TIMEOUT` override with flag > env > default precedence.
- Applies the configured timeout to both Azure SDK `TryTimeout` and HTTP response-header timeout.

### DIFF
--- a/cli/azd/extensions/azure.ai.routines/internal/cmd/routine_helpers_test.go
+++ b/cli/azd/extensions/azure.ai.routines/internal/cmd/routine_helpers_test.go
@@ -56,8 +56,8 @@ func TestRoutineHTTPTimeoutOverrideFromEnv_Invalid(t *testing.T) {
 	_, err := routineHTTPTimeoutOverrideFromEnv()
 	require.Error(t, err)
 
-	var localErr *azdext.LocalError
-	require.True(t, errors.As(err, &localErr))
+	localErr, ok := errors.AsType[*azdext.LocalError](err)
+	require.True(t, ok)
 	assert.Equal(t, exterrors.CodeInvalidParameter, localErr.Code)
 	assert.Contains(t, localErr.Message, routineHTTPTimeoutEnvVar)
 }
--- a/cli/azd/extensions/azure.ai.routines/internal/cmd/routine_helpers_test.go
+++ b/cli/azd/extensions/azure.ai.routines/internal/cmd/routine_helpers_test.go
@@ -4,10 +4,103 @@
 package cmd
 
 import (
+	"errors"
+	"os"
 	"testing"
+	"time"
 
+	"azure.ai.routines/internal/exterrors"
+	"azure.ai.routines/internal/pkg/routines"
+
+	"github.com/azure/azure-dev/cli/azd/pkg/azdext"
+	"github.com/spf13/cobra"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
+
+func unsetEnv(t *testing.T, key string) {
+	t.Helper()
+
+	old, found := os.LookupEnv(key)
+	require.NoError(t, os.Unsetenv(key))
+	t.Cleanup(func() {
+		if found {
+			require.NoError(t, os.Setenv(key, old))
+			return
+		}
+		require.NoError(t, os.Unsetenv(key))
+	})
+}
+
+// --- HTTP timeout config
+
+func TestRoutineHTTPTimeoutFromEnv_Default(t *testing.T) {
+	unsetEnv(t, routineHTTPTimeoutEnvVar)
+
+	got, err := routineHTTPTimeoutFromEnv()
+	require.NoError(t, err)
+	assert.Equal(t, routines.DefaultRequestTimeout, got)
+}
+
+func TestRoutineHTTPTimeoutFromEnv_Override(t *testing.T) {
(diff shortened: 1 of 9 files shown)


therealjohn (approved):

JeffreyCA (approved):
Approving for doc updates

cli/azd/extensions/azure.ai.routines/internal/cmd/routine_helpers_test.go
```diff
@@ -4,11 +4,111 @@
...
+}
+
+func TestRoutineHTTPTimeoutOverrideFromEnv_Invalid(t *testing.T) {
+	t.Setenv(routineHTTPTimeoutEnvVar, "soon")
+
+	_, err := routineHTTPTimeoutOverrideFromEnv()
+	require.Error(t, err)
+
+	var localErr *azdext.LocalError
+	require.True(t, errors.As(err, &localErr))
```
RickWinter: This uses the pre-Go 1.26 error-unwrapping pattern. The repository requires `errors.AsType` and enforces `go fix ./...`, so this can fail the modernization check. Let's use the current pattern here. [azd-code-reviewer]
```suggestion
localErr, ok := errors.AsType[*azdext.LocalError](err)
require.True(t, ok)
```

RickWinter (commented):
The timeout split and override precedence are sound; one repository-enforced Go modernization remains in the new test.

trangevi:
@copilot address the open PR comments

therealjohn (approved):

jongio (approved):
The read/write pipeline split is clean, timeout precedence (flag > env > default) is correct, and tests cover the key paths well. All callers of the updated NewClient signature are handled.

PR #9079, opened 2026-07-10
feat: add service-scoped env to Foundry extensions

### LINKED ISSUE #9231: Migrate Foundry extensions to service-level env
## Goal
Migrate the Foundry extension service targets to the service-level `env` contract provided by azd core.

## Why
azd core now expands environment values for each service before invoking its extension. Foundry extensions need to consume those service-scoped values instead of relying on the complete active azd environment. This prevents values from one service from leaking into another while keeping existing projects compatible.

## Scope
- Update agent, connection, routine, and toolbox service targets and schemas to use service-level `env`.
- Keep the active azd environment fallback for legacy service definitions.
- Migrate generated agent definitions from deprecated `environmentVariables` to service-level `env`.
- Preserve `${VAR}` references, defaults, and Foundry `${{...}}` expressions through init and optimize flows.
- Apply deterministic local agent-run precedence so process and command-owned values override service values, while modern services do not inherit unrelated project-wide variables.
- Update examples, extension minimum azd versions, and SDK dependencies for the new contract.

## Completion criteria
- Modern Foundry services consume their scoped environment values.
- Existing service definitions without service-level `env` continue to work.
- Generated and optimized `azure.yaml` files preserve environment templates without persisting expanded secrets or values.
- Agent, connection, routine, and toolbox configuration surfaces reflect the service-level environment contract.
- Regression tests cover migration, template preservation, precedence, isolation, and legacy fallback behavior.

## Related work
- Follow-up to #8936
- Implemented by PR #9079

huimiu:
## Summary

Foundry extensions (agent, connection, routine, toolbox) now read **service-scoped** environment values from `ServiceConfig.Environment` instead of the project-wide azd environment, and write generated variables into the standard `services.<name>.env` block.

- Each target prefers the forwarded service environment; services without `env` keep the old project-wide behavior.
- Fixed two timing bugs where extensions read environment values *before* the user was prompted for them.
- Replaced the agents extension's two independent `${VAR}` scanners with one shared scanner.
- Updated schemas, examples, minimum azd versions, and SDK dependencies for azd 1.27.1.

Follow-up to #8936. Fixes: #9231

## Why this change

**Service scoping.** azd core now expands each service's environment before handing it to the extension. Extensions have to consume those values so services stop leaking variables into each other, while projects with no service-level `env` keep working unchanged.

**The empty-value bug.** `azd up` initializes every service target while building the execution graph — before the user is prompted for a missing subscription or location. A config read at that moment holds empty strings, so the user answers the prompt and the extension still deploys a blank value, with no error anywhere. Two fixes: the agent target now uses the config passed to each deploy-time entrypoint instead of the copy cached at `Initialize` (re-running `$ref` expansion on it), and the provisioning provider resolves the azd environment *before* it reads service environments.

**One `${VAR}` scanner.** Writing an `env:` block means deciding which variables a service references. #9212 answered a near-identical question — which variables to prompt for — with a second implementation, and the two had already drifted:

| | init prompting | generated `env:` block |
|---|---|---|
| Implementation | regex | hand-written scanner |
| Nested default `${A:-${B}}` | inexpressible | partly handled |
| Escaping | odd/even `$` count | any preceding `$` |

They are now one scanner, pinned by a test to `foundry.ExpandEnv` — the code that actually resolves these values — so it can't drift again. Escaping follows the expander too: `$${VAR}` is literal, `$$${VAR}` expands. The only remaining policy difference is one explicit line: init skips a reference that has a default, the env block records it.

Nested references stay out of scope by design: `${OUTER:-${NESTED}}` reports `OUTER` only. `foundry.ExpandEnv` still resolves `NESTED` at deploy, but nothing discovers it, so it never reaches the generated `env:` block and init never prompts for it. That limitation is now stated in the scanner's doc comment and pinned by a test instead of left implicit — keep defaults literal.

## Environment terminology

| Term | Meaning |
| --- | --- |
| Process environment | Inherited by a locally launched agent. Highest precedence, with command-owned values. |
| Active azd environment | Project-wide values. Modern services use these only for required platform values; services without `env` also use them as the legacy fallback. |
| Service `env` | Raw templates under `services.<name>.env`. Commands use the project config API to preserve them on disk. |
| Forwarded service environment | Service-scoped, core-expanded literals from `ServiceConfig.Environment`. Consumed as-is, never expanded again. |
| Agent definition environment | Deprecated inline `environmentVariables` or legacy `agent.yaml`. Service `env` wins when both exist. |

Template syntax has distinct owners: `${VAR}` is resolved by azd, `${{...}}` by Foundry, and `$${{...}}` preserves a Foundry expression through azd expansion.

## Manual Testing

- `azd ai agent init` — migrated legacy `environmentVariables` to `env`, preserved `${VAR:-default}`, escaped Foundry templates as `$${{...}}`.
- `azd ai agent run` — service `env` resolved for the child process while the template stayed intact on disk; legacy inline definitions resolved through the azd fallback.
- `azd ai agent optimize apply` — raw templates and scalars preserved through field-level updates.
- `azd deploy` — toolbox endpoint resolved from service `env`, and from the azd fallback for a legacy toolbox.


cli/azd/extensions/azure.ai.agents/internal/project/agent_definition_test.go
```diff
@@ -110,6 +114,35 @@ func TestAgentDefinitionFromService_LegacyConfigShape(t *testing.T) {
...
+		Host:   "azure.ai.agent",
+		Config: props,
+	}
+	got, _, found, source, err := AgentDefinitionFromService(svc)
+	require.NoError(t, err)
+	require.True(t, found)
+	require.Equal(t, AgentDefinitionSourceLegacyConfig, source)
+	require.Equal(t, map[string]string{
+		"LEGACY_KEY": "${LEGACY_KEY}",
+	}, AgentEnvironment(got))
```
jongio, line 154: This legacy-env test only exercises the read path (`AgentDefinitionFromService` then `AgentEnvironment`). It never runs the persist path that `optimize apply` uses (`UpsertAgentEnvVars` then `Project().AddService`), which is where a migrated `${LEGACY_KEY}` is at risk.

When a service migrates from inline `environmentVariables`, it has no existing core `env` entry, so `preserveUnchangedEnvTemplates` in `internal/grpcserver/project_service.go` skips the key (`if !has { continue }`) and the mapper persists the value as an expanded literal, turning `${LEGACY_KEY}` into `$${LEGACY_KEY}` in azure.yaml. So this test passes while the actual migration changes semantics.

Consider a test that drives the migration through `AddService` (or a recording project server) and asserts the persisted template stays `${LEGACY_KEY}`.

jongio (commented):
One follow-up on the env migration path.

The new legacy-environment test covers the read path but not the `optimize apply` persist path through `AddService`, so it won't catch the `${VAR}` to `$${VAR}` escaping that the existing review comment describes. Worth a persist-path test for that migration. Detail inline.

jongio (approved):
Re-reviewed after the merge from main and conflict resolution. My earlier note on env persist-path coverage is addressed: TestUpsertAgentEnvVars now drives UpsertAgentEnvVars and asserts the inline environmentVariables key is dropped. The service-target refactor reads the env values from ServiceConfig.Environment and falls back to the active azd environment when a service has no env, so existing definitions keep resolving. The template collector correctly skips the Foundry double-brace form and the escaped dollar-dollar literal form.

jongio (approved):
Re-reviewed the follow-up commit that preserves legacy inline env vars during migration. toContainerAgent now seeds the map from the inline definition and overlays the service environment, so legacy-only keys survive and service values win on conflicts. TestAgentDefinitionFromService_LegacyEnvironment covers the merge (legacy-only key preserved, new service key added, shared key resolves to the service value). Refreshing my approval since the earlier one predated this push.

jongio (commented):
Heads up on a merge artifact from the main sync: the #9079 CHANGELOG entries that documented this migration were dropped when the upstream betas released and got merged in.

At my earlier review the agents CHANGELOG had an Unreleased entry ("Store hosted-agent runtime variables in the standard service-level `env` object and read the expanded values forwarded by azd core...") and microsoft.foundry had one noting the azd 1.27.1 requirement. Both are gone at HEAD now, so the env migration and the `requiredAzdVersion: ">=1.27.1"` bump currently ship with no changelog trail.

Is that intentional given the separate version-bump release PR described in `azure.ai.agents/AGENTS.md`, or should those Unreleased entries be restored? Not blocking, just flagging so the behavior change keeps its changelog trail.

The rest of the new commits look correct to me: built and ran the agents extension unit tests locally (cmd + project packages), and go vet plus go fix are clean. The service-env precedence in `mergeAgentRunEnvironment` and the hand-rolled template parser in `resource_services.go` are well covered by the added tests.

jongio (approved):
Reviewed the follow-up commit that forwards service env through the Foundry deployment paths.

`mergeAgentRunEnvironment` now overlays service env keys onto the base environment and skips the inline definition value when a key is service-scoped, so service values win and empty service values survive. `TestMergeAgentRunEnvironment` covers `SERVICE_ONLY` and the empty-value case. `prepareDeploy` seeds `resolvedEnvVars` from the core-expanded service environment and skips inline redefinition, and the provisioning provider reads per-service scopes via `Project().Get()` and threads them into `Synthesize` and `BrownfieldConnections`. The nil `azdClient` guard in `projectServiceEnvironments` is unreachable from the extension host since `host.Client()` is always set.

Still open from my last pass: the CHANGELOG entries for this env migration and the `>=1.27.1` bump are still absent at HEAD. Not blocking, just flagging so the behavior change keeps a changelog trail.

One thing to confirm on `connectionEnvironmentMapping`: once a connection service declares any `env`, `${VAR}` references that aren't in that service's scope resolve to empty (with `:-` defaults still honored) instead of falling back to the project or process env. The isolate-lookup test asserts that, so it reads as intended, but it means a connection referencing a project-wide var it didn't redeclare in its own env would now resolve empty. Confirming that's the intended model.

jongio (approved):
Re-reviewed the six commits since my last approval: the $ref file-reference resolution refactor across the synthesizers and provisioning provider, zure.yml support, ACR connection validation, the Windows-casing fix in nvSliceHasKey, and the proto.Clone guard in collectLegacyAgentConfigs. No new issues in the delta.

jongio (approved):
Re-reviewed the six commits since my last approval: the file-reference (ref) resolution refactor across the synthesizers and provisioning provider, azure.yml support, ACR connection validation, the Windows-casing fix in envSliceHasKey, and the proto.Clone guard in collectLegacyAgentConfigs. No new issues in the delta.

jongio (approved):
Re-reviewed the single new commit since my last approval: the `bicep restore` pass added to Test-BicepLint.ps1. It restores modules serially before the parallel build loop to avoid the BCP190 restore race, reuses the same filtered file list, and captures the restore exit code the same way the build loop does. No new issues in the delta. Refreshing my approval against the current HEAD.

cli/azd/extensions/azure.ai.agents/internal/cmd/resource_services.go
```diff
@@ -287,6 +288,15 @@ func addResourceService(
 		return fmt.Errorf("adding %s service %q: %w", host, name, err)
 	}
 
+	if err := setServiceEnvironment(
```
trangevi: Can we just put the environment on the serviceConfig object on line 281, instead of needing this separate method? I'm not sure I understand what that method is doing on top of the environment object.

cli/azd/extensions/azure.ai.agents/internal/cmd/resource_services.go
```diff
@@ -296,6 +306,198 @@ func addResourceService(
 	return nil
 }
 
+func serviceEnvironmentTemplates(cfg *structpb.Struct) map[string]string {
```
trangevi, line 311: I'm not sure I'm understanding why we need all of this additional logic. Is it just because we don't have a strongly typed object in `addResourceService`, so we have to operate on a generic map to get the data that we want?

cli/azd/extensions/azure.ai.agents/internal/cmd/resource_services.go
```diff
@@ -296,6 +306,198 @@ func addResourceService(
...
+			offset = start + 2
+			continue
+		}
+		// Forward the canonical bare ${NAME}, not the full matched span.
+		// A ${NAME:-default} default is re-applied by the owning
+		// extension against the raw config at deploy, so the env section
+		// only needs NAME's resolved base value. Collapsing every form of
+		// a var to one value also keeps collection deterministic when the
+		// same var appears with and without a default.
+		environment[name] = "${" + name + "}"
```
trangevi: So this is necessary because the new `env` property is a string to string dictionary, instead of a list of name/value objects?

cli/azd/extensions/azure.ai.agents/internal/cmd/run.go
```diff
@@ -1032,75 +971,109 @@ func findSystemPython() (pythonInterpreter, error) {
 	return firstCompatiblePython(pythonCandidates(), pythonVersion)
 }
 
-// appendFoundryEnvVars translates azd environment keys to FOUNDRY_* env vars that hosted
-// agent containers receive automatically from the platform. This ensures the agent code
-// works identically whether running locally (via azd ai agent run) or in a hosted container.
+// mergeAgentRunEnvironment builds the local agent environment.
```
trangevi, line 1079: Can you add a comment which explains where each of these different "environments" is coming from?

cli/azd/extensions/azure.ai.agents/schemas/examples/complex.azure.yaml
```diff
@@ -108,7 +110,7 @@ services:
       - research-tools
     env:
       LOG_LEVEL: info
-      MODEL_ENDPOINT: ${{project.endpoint}}
+      MODEL_ENDPOINT: $${{project.endpoint}}
```
trangevi, line 114: Why are there two "$"?

cli/azd/extensions/azure.ai.connections/schemas/azure.ai.connection.json
```diff
@@ -38,7 +38,7 @@
     },
     "credentials": {
       "type": "object",
-      "description": "Credentials. Values may contain ${VAR} (azd env, resolved client-side) or ${{...}} (Foundry server-side resolution, passed through untouched).",
+      "description": "Credentials. Values may contain ${VAR} declared in the service-level env object or ${{...}} for Foundry server-side resolution.",
```
trangevi: This line has `${VAR}` and `${{VAR}}`, I left a comment above regarding something that has `$${VAR}}`. Are all of these different things?

eng/scripts/Test-BicepLint.ps1
```diff
@@ -30,6 +30,20 @@ function parseErrorLine($line) {
 $bicepFiles = Get-ChildItem "$Path/*.bicep" -Recurse -Force | 
     Where-Object { $_.FullName -notlike "*cli/azd/test/functional/testdata/snaps*" }
 
+# Restore modules before parallel builds to prevent BCP190.
```
trangevi: Where is this change coming from?

trangevi (changes requested):
Can you please document where possible the terminology being used for the different environment locations? It's a bit difficult to keep track, and at least to me non-intuitive if I'm understanding correctly

jongio (approved):
Re-reviewed the delta since my last approval (2b33122...bfa0dcd). The PR-authored new work is the doctor agent-definition validation:

- `newCheckAgentDefinitionValid` (#9254) resolves each agent service through the same `project.LoadAgentDefinition` path that run and deploy use, instead of reading `agent.yaml` off disk. Inline `azure.yaml` and `$ref` definitions no longer report a false failure, and `collectSortedAgentServices` keeps the failure list deterministic.
- `validateAgentServiceDefinition` (#9264) closes the gap where a non-hosted `kind` skipped validation entirely: the non-hosted branch now validates `s.AsMap()` and the hosted branch validates the built `ContainerAgent`. `ValidateAgentDefinition` unmarshals leniently, so extra service fields on the raw map don't trigger false rejections, and valid `workflow` definitions still pass. The new inline and `$ref` kind tests cover both.

My earlier changelog flag is resolved: the beta.7 section now documents the env migration betas plus #9254 and #9264. No new issues in the delta, so refreshing my approval against the current HEAD.

jongio (commented):
A few things worth confirming before this merges, alongside the open threads:

1. `getRawServiceEnv` in `optimize_apply.go` only keeps string values from the on-disk `env` (`if str, ok := value.(string)`). Non-string scalars like `ENABLED: true` or `PORT: 8080` get dropped, and because `persistInlineAgentEnvironment` rewrites the entire `env` section from the merged map, running `optimize apply` silently deletes those entries. Normalizing scalars here (the extension already has `projectconfig.NormalizeEnvironment`) instead of skipping non-strings would avoid the data loss.

2. `validateRootRefCoreFields` rejects `project`/`language`/`image`/`docker` from a root `$ref` but not `env`. Core only forwards service-level `env` declared in azure.yaml, so a referenced definition that declares `env` won't have those values forwarded and can fall through to the global legacy fallback in both local run and hosted deploy. Either forward the resolved `$ref` env or reject `env` in root refs with a migration path.

3. The `bicep restore` loop added to `eng/scripts/Test-BicepLint.ps1` looks unrelated to the Foundry env work (same thing trangevi asked about). If it's a needed CI fix, a short note on why or a separate PR would keep this change scoped.

None of these block from my side; flagging so they don't get lost.

cli/azd/extensions/azure.ai.agents/internal/cmd/resource_services.go
huimiu: `ServiceConfig.Environment` only carries values after azd expands them, so using it here would lose the original `${VAR}` templates. The separate config call preserves the raw templates in `azure.yaml`

cli/azd/extensions/azure.ai.agents/internal/cmd/resource_services.go
huimiu: `addResourceService` receives the extension-specific config as a generic `structpb.Struct`, so there is no typed field to read here. The recursive scan finds `${VAR}` references while intentionally skipping Foundry `${{...}}` expressions.

cli/azd/extensions/azure.ai.agents/internal/cmd/resource_services.go
huimiu: Yes. The service-level `env` in `azure.yaml` is a string-to-string map, while `agent.yaml` uses name/value entries, so this converts it into the shape core expects.

cli/azd/extensions/azure.ai.agents/schemas/examples/complex.azure.yaml
huimiu: `$${{project.endpoint}}` is the escaped form stored in `azure.yaml`. azd removes the extra `$` and passes `${{project.endpoint}}` through for the Foundry extension to resolve

cli/azd/extensions/azure.ai.connections/schemas/azure.ai.connection.json
huimiu: Yes, they are different. `${VAR}` is resolved from azd environment values, `${{...}}` is resolved by Foundry, and `$${{...}}` escapes the Foundry expression while it passes through azd

eng/scripts/Test-BicepLint.ps1
huimiu: That was an unrelated change accidentally carried over from another branch. I've removed it from this PR

cli/azd/extensions/azure.ai.agents/internal/cmd/optimize_apply.go
```diff
@@ -272,6 +253,103 @@ func (a *OptimizeApplyAction) apply(
...
+	}
+	raw, ok := resp.GetValue().AsInterface().(map[string]any)
+	if !ok {
+		return nil, nil
+	}
+	env := make(map[string]string, len(raw))
+	for key, value := range raw {
+		if str, ok := value.(string); ok {
+			env[key] = str
+		}
```
jongio, line 369: `getRawServiceEnv` only keeps string values here, so non-string scalars in the on-disk `env` (like `ENABLED: true` or `PORT: 8080`) get dropped. Since `persistInlineAgentEnvironment` rewrites the whole `env` section from `mergedEnv`, running `optimize apply` silently deletes those entries. Normalizing scalars to strings here (the extension already has `NormalizeEnvironment`) instead of skipping non-strings would avoid the data loss.

jongio (commented):
The `optimize apply` persist path still drops non-string env scalars, which silently deletes those entries from `azure.yaml` on apply. Flagged inline.

cli/azd/extensions/azure.ai.agents/internal/cmd/run.go
huimiu: Sure, added comments

jongio (approved):
The new commits resolve the env normalization concerns from the prior round. `getRawServiceEnv` now runs values through `NormalizeEnvironment`, so scalar values like `ENABLED: true` and `PORT: 8080` convert to strings instead of getting dropped, and maps/slices now error out instead of silently disappearing. Root `` definitions that declare `env` are rejected with a clear migration message pointing to azure.yaml. The added regression coverage exercises these paths.

jongio (commented):
One note on the automated review comments about the `len == 0` fallback in the Foundry targets (run.go, the routine and toolbox service targets, and the projects synthesizer). They're framed as cross-service secret leakage, but `azdEnvironment` and the service-scoped environment are both values from the same azd environment, so this is scoping within a single project, not a cross-tenant boundary. Nothing leaks between users or subscriptions.

The one real nuance: an explicit `env: {}` serializes the same as an omitted `env`, so it can't opt out of the compatibility fallback. If honoring an explicit empty scope ever matters, gate the fallback on whether `env` was declared in the raw service config rather than on map length. Low priority and fine to defer, since the fallback is intentional back-compat.

cli/azd/extensions/azure.ai.agents/internal/cmd/helpers.go
```diff
@@ -1014,39 +995,24 @@ func resolveServiceRunContext(ctx context.Context, azdClient *azdext.AzdClient,
 		}
 	}
 
+	hasServiceEnvironment := false
+	if resp, envErr := azdClient.Project().GetServiceConfigValue(
+		ctx,
+		&azdext.GetServiceConfigValueRequest{ServiceName: svc.Name, Path: "env"},
+	); envErr == nil {
+		hasServiceEnvironment = resp.GetFound()
```
jongio: On a `GetServiceConfigValue` error this leaves `hasServiceEnvironment` false, so `mergeAgentRunEnvironment` takes the legacy branch and injects the full azd environment, the exact leak this commit is closing. The routines and toolboxes targets in this same change fail closed by returning the error from `serviceEnvDeclared`. Do the same here and only trust `Found` after a successful call.

jongio (commented):
One issue on the latest commit (ce4d70ae):

MEDIUM
- agents `helpers.go`: the env-detection RPC failure falls open to the full-environment fallback, unlike the routines and toolboxes targets in this same commit which fail closed.

jongio (commented):
The new commit closes the explicit `env: {}` isolation gap in the synthesizers and the routines/toolboxes targets, and the added tests cover the empty-declared case.

One issue is still open on this HEAD (already flagged inline on `helpers.go:1003`): in `azure.ai.agents/internal/cmd/helpers.go`, `resolveServiceRunContext` swallows the `GetServiceConfigValue` error and leaves `hasServiceEnvironment` false. On a transient error that sends `mergeAgentRunEnvironment` down the legacy branch and injects the full azd environment, the exact leak this commit closes everywhere else. The routines and toolboxes targets fail closed by returning the error from `serviceEnvDeclared`. The agents path should do the same and only trust `Found` after a successful call.

cli/azd/extensions/azure.ai.connections/schemas/azure.ai.connection.json
trangevi: Why do we need all three? If `${VAR}` is handled locally, and `${{VAR}}` is handled remotely, why do we need the third? Where does the parsing happen, and can it be updated to properly filter on `${` and `${{` so that we only need two? I don't think that the string literal "{VAR}" would be considered a valid env var key anyway, so I don't thing we need to be worried about properly being able to resolve that to something concrete

cli/azd/extensions/azure.ai.agents/internal/cmd/resource_services.go
trangevi: What does "after azd expands them" mean? Does that mean if I have `${VAR}`, azd core will replace that with the value of `VAR` in the azd environment?

jongio (commented):
One issue from my earlier reviews is still open on this HEAD, already flagged inline at `helpers.go:1003`.

In `azure.ai.agents/internal/cmd/helpers.go`, `resolveServiceRunContext` swallows the `GetServiceConfigValue` error and leaves `hasServiceEnvironment` false. On a transient error that pushes `mergeAgentRunEnvironment` down the `!hasServiceEnvironment` branch and injects the full azd environment, the exact leak this change closes everywhere else. The routines and toolboxes targets fail closed by returning the error from `serviceEnvDeclared`. The agents path should do the same and only trust `Found` after a successful call.

cli/azd/extensions/azure.ai.connections/schemas/azure.ai.connection.json
huimiu: There are only two kinds of expressions:
- `${VAR}` is resolved by azd.
- `${{...}}` is resolved by Foundry.

`$${{...}}` isn't a third kind — it's just the escaped form of a Foundry expression inside a service `env` value. Core expands `env` before handing it to the extension: it understands `${VAR}`, but a raw `${{...}}` makes core's envsubst error out, so the extra `$` tells core to pass the expression through unchanged.

You're right that this is the awkward part — migration rewrites `${{...}}` to `$${{...}}` when it moves a value into `env`, so the input and output `azure.yaml` don't match. The clean fix is to teach core's env expansion to preserve `${{...}}` (the same masking we already do for Foundry config fields like `target`/`credentials`), which gets us back to two syntaxes everywhere. That changes env handling for every azd service, so I want to check the compatibility impact and add coverage first — I'd rather do that than keep the escape as a user-facing workaround. So I'll keep it out of this PR and track the core fix separately.

cli/azd/extensions/azure.ai.agents/internal/cmd/resource_services.go
huimiu: `ServiceConfig.Environment` contains resolved values, not the original text from `azure.yaml`. For example, if `azure.yaml` contains `${API_KEY}`, core sends the extension the actual API key.

`AddService` therefore treats these values as plain text and cannot use this field to create a new `${API_KEY}` reference. `SetServiceConfigSection` writes the original text directly, which is why the separate call is needed

jongio (commented):
Re-reviewed at ce4d70a. The empty-`env` isolation is handled the right way in routines, toolboxes, and both synthesizers, but two things on the agents path are still open.

The fail-open in `azure.ai.agents/internal/cmd/helpers.go` (line 1003) is unchanged: `resolveServiceRunContext` drops the `GetServiceConfigValue` error and leaves `HasServiceEnvironment` false, so `mergeAgentRunEnvironment` falls back to injecting the full azd environment on a transient RPC failure, which re-opens the exact leak this commit closes. routines and toolboxes fail closed by returning the error from `serviceEnvDeclared`; please do the same here and only trust `Found` after a successful call.

The new `TestMergeAgentRunEnvironment` case passes `hasServiceEnvironment` in directly, so it never exercises `resolveServiceRunContext` where the error is swallowed. Stubbing the client to error from `GetServiceConfigValue` would cover the leak path and lock the fix in.

jongio (commented):
Re-reviewed the new commit. Still open from my earlier pass:

- cli/azd/extensions/azure.ai.agents/internal/cmd/helpers.go: `GetServiceConfigValue` error still falls back to the full azd environment instead of failing closed like the routines and toolboxes targets do in this same commit.

<sub>🤖 agent jongio</sub>

jongio (commented):
Re-reviewed the latest commit (honor explicit empty service env as an isolated scope). The routines, toolboxes, and synthesizer paths handle the isolated scope, but the agents run path still has the gap I flagged inline.

In `resolveServiceRunContext` (`agents/internal/cmd/helpers.go`), the `GetServiceConfigValue` error is swallowed, so `hasServiceEnvironment` stays false whenever the probe fails. `mergeAgentRunEnvironment` then takes the `!hasServiceEnvironment` branch and injects the full azd environment, which is the exact leak this commit closes for a service that declared an isolated (empty) `env`. The routines and toolboxes targets fail closed by returning the error from `serviceEnvDeclared`; agents should match that and only trust `Found` after a successful call.

Not blocking anything new, just making sure this carries over before the change lands.

jongio (commented):
The empty-`env` isolation in the latest commit works: `env: {}` now reads as a declared scope through `GetServiceConfigValue`'s `Found` in the run, routine, and toolbox targets, and through the raw-node check in the synthesizers, so a service that opts into an empty scope no longer inherits the full azd environment. I traced it against core `LoadConfig`/`config.Get`, which keeps `env: {}` as a present key.

One item is still open from this push, already flagged inline on `helpers.go`: `resolveServiceRunContext` swallows a `GetServiceConfigValue` error and leaves `hasServiceEnvironment` false, so `mergeAgentRunEnvironment` takes the legacy branch and injects the full azd environment, the same leak this commit closes. The routine and toolbox targets fail closed by returning the error from `serviceEnvDeclared`; helpers.go should match that and only trust `Found` after a successful call.

Unrelated to this commit: the on-disk (ejected) provisioning path in `foundry_provisioning_provider.go` skips the synthesizer and resolves parameters with the project-wide `resolveEnv` rather than the service-scoped map, so isolation doesn't reach eject-to-Bicep. Is that intended, or should the on-disk path carry the same scoping?

cli/azd/extensions/azure.ai.agents/internal/project/agent_definition.go
```diff
@@ -119,12 +120,28 @@ func agentDefinitionToInline(ca agent_yaml.ContainerAgent) (AgentDefinitionInlin
...
 // definition, the CPU/memory carried in the `container` config, and the image
 // carried on the core service field.
-func (d AgentDefinitionInline) toContainerAgent(container *ContainerSettings, image string) agent_yaml.ContainerAgent {
+func (d AgentDefinitionInline) toContainerAgent(
+	container *ContainerSettings,
+	image string,
+	environment map[string]string,
+) agent_yaml.ContainerAgent {
+	environmentVariables := d.EnvironmentVariables
+	if len(environment) > 0 {
```
jongio, line 171: This is now the only source of env for the hosted deploy, and it only sees what core forwards through `ServiceConfig.Environment`.

This PR removes `Environment` from `ServiceTargetAgentConfig`, removes `env` from `azure.ai.agent.json`, and removes both readers: `maps.Copy(serviceEnv, agentConfig.Environment)` in the run path and `for name, value := range foundryAgentConfig.Environment` in `prepareDeploy`. Core only forwards `services.<name>.env`, so a project still using the deprecated `config:` nested shape with an `env:` key under it loses those values on both local run and deploy.

The nested shape is otherwise still alive: `ServiceConfigProps` falls back to `svc.GetConfig()` and `WarnLegacyAgentShape` still fires. And `persistInlineAgentEnvironment` only unsets `environmentVariables` / `config.environmentVariables`, so nothing migrates `config.env`.

Either migrate `config.env` into service level `env` the same way `environmentVariables` is migrated, or warn when a service still carries it, so the values don't disappear quietly.

cli/azd/extensions/azure.ai.agents/internal/cmd/resource_services.go
```diff
@@ -296,6 +306,202 @@ func addResourceService(
...
+
+// setServiceEnvironment writes raw templates through the config RPC.
+// AddService treats ServiceConfig.Environment as expanded literals.
+func setServiceEnvironment(
+	ctx context.Context,
+	azdClient *azdext.AzdClient,
+	serviceName string,
+	environment map[string]string,
+) error {
+	if len(environment) == 0 {
```
jongio, line 388: Question on the empty case. This returns before writing anything, so a service whose collected templates come back empty gets no `env:` key written to azure.yaml at all.

At run time that's indistinguishable from a legacy service, so `mergeAgentRunEnvironment` takes the `!hasServiceEnvironment` branch and injects the entire azd environment into the child process. An agent that `init` generated with no variables of its own still gets everything locally, while the hosted path doesn't.

Since the last commit specifically taught the targets to honor an explicit `env: {}` as an isolated scope, should init write an explicit empty `env` here so newly generated services start isolated too?

jongio (commented):
A couple of things worth a look before this merges, plus one from my earlier pass that's still open.

Dropping `Environment` from `ServiceTargetAgentConfig` (and `env` from the agent schema) silently stops honoring `config: { env: ... }` for projects still on the deprecated nested shape. Nothing migrates it and nothing warns.

`setServiceEnvironment` skips writing an empty `env`, so a freshly generated agent with no variables of its own is indistinguishable from a legacy service and takes the full azd environment path on local run.

Still open from my last pass: `internal/cmd/helpers.go` swallows the `GetServiceConfigValue` error and leaves `hasServiceEnvironment` false, so `mergeAgentRunEnvironment` falls back to injecting the whole azd environment into the child process. That's the leak this PR is closing. The routines target, the toolboxes target, and `persistInlineAgentEnvironment` all fail closed on that same call, so this one path is the odd one out.

cli/azd/extensions/azure.ai.projects/internal/provisioning/foundry_provisioning_provider.go
```diff
@@ -166,27 +167,44 @@ func (p *FoundryProvisioningProvider) Initialize(
...
+			"fix the project service configuration in azure.yaml",
+		)
+	}
+
+	// Resolve the environment before reading service values. azd core
+	// expands ${VAR} in service env against the environment, so
+	// reading them first would capture empty strings for values the
+	// user is about to be prompted for, and connection synthesis
+	// would provision those empty strings.
+	if endpoint != "" {
```
jongio, line 224: `resolveEnv` prompts for the subscription and the location and persists both, and it now runs ahead of the azure.yaml validation that used to gate it. `Synthesize` is what surfaces `ErrServiceNotFound`, a bad `deployments:` block or a bad `network:` block, and it doesn't run until after this.

So a project with a typo in the project service now gets the subscription picker and the location picker first, and only then the config error.

Under `--no-prompt` the diagnosability is worse: `promptSubscription` returns the "run `azd env set AZURE_SUBSCRIPTION_ID`" error, so a CI run against a malformed azure.yaml reports a missing subscription instead of the actual problem in the file.

Only `collectConnections` needs the service environments. Could the host and service validation run before the environment is resolved, so config errors still surface first?

cli/azd/extensions/azure.ai.projects/internal/provisioning/foundry_provisioning_provider.go
```diff
@@ -166,27 +167,44 @@ func (p *FoundryProvisioningProvider) Initialize(
 		return err
 	}
 
+	// endpoint: (brownfield) reuse connects to an existing project,
+	// so it needs no subscription or location. Detect it up front so
+	// the environment can be resolved before any service values are
+	// read.
+	endpoint, err := foundryServiceEndpointAtRoot(rawYAML, projectPath, svcName)
```
jongio, line 178: This now runs for every project instead of just the on-disk and brownfield paths, so `foundry.ResolveFileRefs` failures on a greenfield project surface here first.

A broken `$ref` under the project service reports "resolve existing Foundry project endpoint", which points the user at an `endpoint:` field they never wrote. Worth wording that covers both cases, something like "read Foundry project service configuration".

jongio (commented):
Two notes on the new commit, plus three items from earlier rounds that are still open on this HEAD.

Resolving the environment before reading the service environments is the right fix for the empty connection target, but it also moves the subscription and location prompts ahead of all the azure.yaml validation that used to run first. Flagged inline, along with an error message that now reads wrong on the greenfield path.

Still open from my earlier reviews, none of them touched by this commit:

- `azure.ai.agents/internal/cmd/helpers.go` (around line 1003): `resolveServiceRunContext` still drops the `GetServiceConfigValue` error and leaves `hasServiceEnvironment` false, so a transient RPC failure pushes a modern agent down the legacy branch and injects the full azd environment. The routines and toolboxes targets fail closed in this same situation.
- `azure.ai.agents/internal/project/agent_definition.go` (around line 129): projects still on the deprecated `config: { env: ... }` nested shape lose those values on both run and deploy, and nothing migrates them or warns.
- `azure.ai.agents/internal/cmd/resource_services.go` (around line 468): `setServiceEnvironment` returns without writing an `env` key when the collected map is empty, so an init-generated service with no variables of its own reads as legacy at run time and gets the whole azd environment.

cli/azd/extensions/azure.ai.agents/internal/cmd/resource_services.go
```diff
@@ -296,6 +306,202 @@ func addResourceService(
...
+			offset = start + 2
+			continue
+		}
+		// env is keyed by name, so store one canonical ${NAME}.
+		// A ${NAME:-default} default is re-applied by the owning
+		// extension against the raw config at deploy, so the env section
+		// only needs NAME's resolved base value. Collapsing every form of
+		// a var to one value also keeps collection deterministic when the
+		// same var appears with and without a default.
+		environment[name] = "${" + name + "}"
```
jongio: `environmentTemplateAt` returns the end of the entire `${NAME:-default}` span and the loop jumps straight there, so any `${VAR}` inside the default is never collected. `${PRIMARY:-${FALLBACK}}` yields only `PRIMARY`, and `${A:-${B:-${C}}}` yields only `A`. I ran the collector against both to confirm.

The comment just above says the default gets re-applied by the owning extension against the raw config at deploy. That holds when the default is a literal. When the default is another variable, that variable has no entry in the generated `env` block, so it isn't part of the service scope the extension resolves against, and an isolated service has nowhere to get it from.

Is the nested form meant to be supported? If it is, this loop needs to recurse into the default span instead of jumping to `end`. If it isn't, collecting it or warning at init time would beat silently resolving to an empty string at deploy.

jongio (commented):
One new item on this HEAD. The four threads from my earlier rounds are still open and untouched by `3ad246ac`.

`collectStringEnvironmentTemplates` skips the whole `:-` default span, so a variable nested inside a default never reaches the generated service `env` block. Flagged inline.

Still unaddressed on this commit:

- `azure.ai.agents/internal/cmd/helpers.go` around line 1003: the swallowed `GetServiceConfigValue` error leaves `hasServiceEnvironment` false, so a transient RPC failure fails open into the legacy branch and injects the full azd environment. The routines and toolboxes targets return an error on that same call.
- `azure.ai.agents/internal/project/agent_definition.go` around line 129: projects on the deprecated nested `config: { env: ... }` shape lose those values on run and deploy, and nothing migrates them or warns. `additionalProperties` is `true` at the root of `azure.ai.agent.json`, so the removed `env` property doesn't get rejected either, it's just dropped.
- `azure.ai.agents/internal/cmd/resource_services.go` around line 468: `setServiceEnvironment` returns before writing anything when the collected map is empty, so an init-generated service with no variables of its own reads as legacy at run time and gets the whole azd environment.
- `azure.ai.projects/internal/provisioning/foundry_provisioning_provider.go` around lines 174 and 188: subscription and location prompting now runs ahead of the azure.yaml validation that used to gate it, and the endpoint error text reads wrong on the greenfield path.

Build and tests are clean locally for all five extensions.

cli/azd/extensions/azure.ai.agents/internal/cmd/resource_services.go
```diff
@@ -296,6 +306,202 @@ func addResourceService(
 	return nil
 }
 
+// serviceEnvironmentTemplates discovers client-side templates in the
+// generic nested resource config emitted to azure.yaml.
+func serviceEnvironmentTemplates(cfg *structpb.Struct) map[string]string {
```
trangevi, line 311: Glenn's recent changes introduced an init_env.go file, maybe some of this handling should go there, to keep our environment manipulation logic more contained?

cli/azd/extensions/azure.ai.routines/internal/cmd/service_target.go
```diff
@@ -185,16 +190,54 @@ func newRoutineServiceClient(ctx context.Context) (*routines.Client, error) {
 	), nil
 }
 
-// currentEnvValues loads all key-value pairs from the active azd environment, used to
-// resolve ${VAR} references in routine fields at deploy time.
-func (p *routineServiceTarget) currentEnvValues(ctx context.Context) (map[string]string, error) {
-	current, err := p.azdClient.Environment().GetCurrent(ctx, &azdext.EmptyRequest{})
+var serviceEnvDeclared = func(
```
trangevi: Why is this function defined like this? Testing?

trangevi: If so, is there not a better way to mock the functionality in GO? It feels pretty weird to have a function declared so differently from the others just to handling the testability of what calls it

jongio (commented):
No new commits since my last pass, so this round is the two questions @trangevi raised. Both are answered inline.

The `init_env.go` one is worth acting on. That file is on main but not on this branch yet, and once it merges `azure.ai.agents/internal/cmd` ends up with two independent `${VAR}` scanners that disagree on `${NAME:-default}`. Comparison table is in the thread.

Still open from earlier rounds on this HEAD:

- agents `helpers.go:1003`: a `GetServiceConfigValue` error leaves `hasServiceEnvironment` false, so the run path falls back to the full azd environment. Routines and toolboxes fail closed on the same call.
- agents `resource_services.go:468`: when collected templates come back empty, no `env:` key is written, so the service is indistinguishable from a legacy one at run time.
- agents `resource_services.go:366`: `${PRIMARY:-${FALLBACK}}` collects only `PRIMARY`, so the nested variable never reaches the `env:` block.
- agents `agent_definition.go:129`: dropping `Environment` from `ServiceTargetAgentConfig` and `env` from the agent schema leaves existing agent configs with no migration path.
- projects `foundry_provisioning_provider.go:174` and `:188`: `resolveEnv` prompts and persists subscription and location ahead of the azure.yaml validation that used to gate it, and a broken `$ref` under the project service surfaces as an endpoint resolution failure.

cli/azd/extensions/azure.ai.agents/internal/cmd/resource_services.go
jongio: Worth doing, and there's a reason beyond tidiness: `init_env.go` isn't on this branch yet. It's on main at `cli/azd/extensions/azure.ai.agents/internal/cmd/init_env.go`, and the last main sync here was `47b7c9a8` on Jul 23, before that file landed.

So once main merges in, package `cmd` holds two independent `${VAR}` scanners:

- `azureYamlEnvRefPattern` plus `collectAzureYamlEnvironmentReferences` in `init_env.go`
- `collectStringEnvironmentTemplates` plus `environmentTemplateAt` here

They disagree on `${NAME:-default}`. I ran both over the same inputs:

| input | this branch | `init_env.go` on main |
| --- | --- | --- |
| `${PLAIN}` | `PLAIN` | `PLAIN` |
| `${NAME:-fallback}` | `NAME` | none |
| `prefix ${ONE} mid ${TWO:-x} suffix` | `ONE`, `TWO` | `ONE` |

`init_env.go` skips defaulted references deliberately. Its comment says the fallback is supplied by the runtime expander, so there's nothing to prompt for. This collector instead records the name and drops the default, because the owning extension re-applies the default against the raw config at deploy.

Each is defensible for its own job, but they'd sit in one package answering the same question two ways, and the `:-` handling is exactly where the nested-default gap I flagged at line 366 lives too. Folding this into `init_env.go` would force that reconciliation. A shared tokenizer with the prompting policy and the env-block policy layered on top would keep the two from drifting.

cli/azd/extensions/azure.ai.routines/internal/cmd/service_target.go
jongio: Yes, it's a test seam. `service_target_test.go` swaps it in `TestRoutineEnvironmentValuesEmptyDeclaredIsolates`, and toolboxes does the same at `service_target.go:337`.

The reason a conventional mock doesn't work is `azdext.AzdClient`: it's a concrete struct, its `projectClient ProjectServiceClient` field is unexported, and the only options are `WithAddress` and `WithAccessToken`. There's no way for an extension package to hand it a fake, so the package-level var is the only seam left.

`ProjectServiceClient` is already an interface though, so the target can depend on the narrow slice it actually uses:

```go
type serviceConfigReader interface {
	GetServiceConfigValue(
		ctx context.Context,
		in *azdext.GetServiceConfigValueRequest,
		opts ...grpc.CallOption,
	) (*azdext.GetServiceConfigValueResponse, error)
}
```

Put `projectClient serviceConfigReader` on `routineServiceTarget`, set it from `azdClient.Project()` in the constructor, and turn `serviceEnvDeclared` back into a plain func taking that reader.

I tried it on this branch. `go build ./...` is clean, `TestRoutineEnvironmentValuesEmptyDeclaredIsolates` passes against a small fake, and the test can take `t.Parallel()` again, which the global swap currently rules out.

One caveat: nothing in the extensions depends on `azdext.ProjectServiceClient` directly today, so this introduces a pattern here even though it's an ordinary Go one.

jongio (approved):
Re-reviewed the two commits since my last pass.

The shared `findEnvironmentReferences` scanner resolves the divergence I flagged between `init_env.go` and `resource_services.go`, and `TestFindEnvironmentReferencesMatchesExpander` pins the scanner to `foundry.ExpandEnv` so the two can't drift apart again. I confirmed `$${A:-${B}}` now yields `B`, which the old regex in `init_env.go` dropped even though the expander does resolve it.

Approving the new commits. My earlier threads on the provisioning provider error wording, `${A:-${B}}` collection in the generated env block, and the routines test seam are still open.

cli/azd/extensions/azure.ai.routines/internal/cmd/service_target.go
huimiu: Good catch, it was a test seam, and you're right that it didn't need to be a package-level var. Fixed

cli/azd/extensions/azure.ai.agents/internal/cmd/resource_services.go
huimiu: Agree, updated

cli/azd/extensions/azure.ai.agents/internal/cmd/env_refs.go
```diff
@@ -0,0 +1,207 @@
...
+		}
+
+		reference, found := environmentReferenceAt(value, index)
+		if !found {
+			index++
+			continue
+		}
+
+		references = append(references, reference)
+		index = reference.End
```
jongio, line 122: `index = reference.End` jumps past the whole `${NAME:-default}` span, so a `${VAR}` inside the default never becomes a candidate. `${OUTER:-${NESTED}}` yields only `OUTER`, and `${A:-${B:-${C}}}` yields only `A`.

I ran the scanner and `foundry.ExpandEnv` over the same values with the same lookup function:

```
"${OUTER:-${NESTED}}"                expander: [NESTED OUTER]   scanner: [OUTER]
"${OUTER:-prefix-${NESTED}-suffix}"  expander: [NESTED OUTER]   scanner: [OUTER]
```

`TestFindEnvironmentReferencesMatchesExpander` doesn't cover this because it only asserts one direction, that every name the scanner reports is one the expander resolves. The failure is in the other direction.

The consequence lands in `collectStringEnvironmentTemplates`. Its comment says the default is re-applied by the owning extension against the raw config at deploy, so the env section only needs the base name. That holds for a literal default. When the default is another variable, `NESTED` gets no entry in the generated `env:` block, so it never reaches `ServiceConfig.Environment`, and `agentDefinitionFromStruct` passes `svc.GetEnvironment()` as the only env source for the hosted container. It resolves to empty at deploy even when it is set in the azd environment.

Either recurse into the default span when `HasDefault` is set and let each consumer keep its own policy, or state the limitation in the doc comment so `${A:-${B}}` isn't silently half-supported.

jongio (commented):
One follow-up on the shared scanner. My earlier thread on this went outdated when `environmentTemplateAt` was deleted, so I'm re-anchoring it on the new code instead of losing it.

The nested-default gap survived the consolidation: `${OUTER:-${NESTED}}` reports only `OUTER`, but `foundry.ExpandEnv` does look up `NESTED`. Nothing in the schemas or examples uses a nested default today, so this is latent rather than firing, and it doesn't change my approval.

cli/azd/extensions/azure.ai.agents/internal/cmd/env_refs.go
jongio: Crossed with your note that this is a one-line change you're doing separately, so treat this as the detail for it rather than a second ask. The part worth keeping is the deploy-time consequence: it isn't only a collection miss, the name never reaches `ServiceConfig.Environment`, so it reads as empty at deploy even when it's set in the azd environment.

jongio (approved):
Incremental pass on `de05ddf9`, the only commit since my last review.

The global `serviceEnvDeclared` var is gone from both the routines and toolboxes targets, replaced with a `serviceConfigReader` interface held on the target and populated from `azdClient.Project()` in the constructor. That answers @trangevi's question on the old declaration: the `var` existed only so tests could swap it, and the tests now inject a fake instead. Both tests picked up `t.Parallel()`, which the package-level var had ruled out.

I built, vetted and ran the tests for both modules at this commit and they pass. `go mod tidy` is a no-op in both, and `google.golang.org/grpc` was already a direct require in each `go.mod`, so the new import doesn't change the dependency set.

Approving this commit. My open threads on the agents `GetServiceConfigValue` error path, the empty generated `env:` block, the provisioning provider error wording, and `${A:-${B}}` collection are unchanged by it and still stand.

cli/azd/extensions/azure.ai.agents/internal/project/agent_definition.go
```diff
@@ -141,6 +158,61 @@ func (d AgentDefinitionInline) toContainerAgent(container *ContainerSettings, im
...
+		return environmentValue, nil
+	}
+	return ExpandEnv(value, func(variableName string) string {
+		if environmentValue, found := serviceEnvironment[variableName]; found {
+			return environmentValue
+		}
+		if mapping == nil {
+			return ""
+		}
+		return mapping(variableName)
```
jongio, line 247: When the name isn't in `serviceEnvironment` this always falls through to `mapping`, and both callers hand it the full azd environment: `run.go:552` passes `azdEnvVars[varName]` and `service_target_agent.go:2712` passes `azdEnv[varName]`. Neither caller knows whether the service declared `env:`.

The projects extension goes the other way in this same PR. Once `declared` is true, `connectionEnvironmentMapping` returns `serviceEnvironment[name]` with no project fallback, so a connection service with `env: {}` resolves `${FOO}` to empty instead of picking up a project-wide `FOO`.

Net effect: a hosted agent with `env: {}` plus a legacy `environment_variables: [{name: TARGET, value: "${FOO}"}]` still resolves `FOO` from the project environment, at run and at deploy.

Deliberate for the deprecated `environment_variables` block, or should it follow the connection rule? `run.go` already carries `runCtx.HasServiceEnvironment` that could be threaded in, and the deploy path would need the same probe the routines and toolboxes targets do. If it is deliberate, worth saying so on the doc comment, because `ResolveAgentEnvironmentVariable` reads as scope-aware right now.

jongio (commented):
Re-ran the full pass on `de05ddf9`. Build, vet and tests are clean in all four modules at this commit and CI is green. The two `MatchesBicepBuild` failures I hit locally also fail on the base commit `7d34c7e`, so those are a local bicep version difference and not from this branch.

One new thing, on the legacy definition path. `ResolveAgentEnvironmentVariable` falls back to the project-wide azd environment even when the service declares an isolated `env:` scope, which is the opposite of what `connectionEnvironmentMapping` does in the projects extension in this same PR. Detail inline.

My earlier threads on the agents `GetServiceConfigValue` error path, the empty generated `env:` block, the provisioning provider ordering and wording, and `${A:-${B}}` collection are unchanged by this commit and still stand. My approve on `de05ddf9` stays as is.

cli/azd/extensions/azure.ai.agents/internal/cmd/env_refs.go
trangevi: We shouldn't support nested env var references, that is an unnecessary complexity. I'll let @therealjohn weigh in if he disagrees, but I think that's a path we don't need to go down

cli/azd/extensions/azure.ai.agents/internal/cmd/env_refs.go
```diff
@@ -0,0 +1,207 @@
...
+	"github.com/azure/azure-dev/cli/azd/pkg/foundry"
+)
+
+// Escape handling must match the expander that owns each field.
+// foundry.ExpandEnv collapses '$' pairs, so $${VAR} stays
+// literal, and it reserves ${{...}} spans for Foundry. The project
+// synthesizers' resolveVars helper expands every match
+// unconditionally.
+const (
+	honorEnvironmentEscaping  = true
```
trangevi, line 34: Why do we need these constants? In what situation do we actually want to ignore the escaping? That seems like it would introduce an undesired inconsistency. (I see the different method calls where we use these constants, but I don't really understand why they are different)

cli/azd/extensions/azure.ai.agents/internal/cmd/env_refs.go
```diff
@@ -0,0 +1,207 @@
...
+		}
+
+		references = append(references, reference)
+		index = reference.End
+	}
+	return references
+}
+
+func environmentReferenceAt(value string, start int) (environmentReference, bool) {
+	index := start + 2
```
jongio: `environmentReferenceAt` starts reading the name at `start+2` but never checks that `value[start+1]` is `{`. Any `$` followed by a name character and a later `}` parses as a reference, with the character right after the `$` silently swallowed as if it were the brace.

I ran this scanner and `foundry.ExpandEnv` over the same inputs with the same lookup function:

```
in="$foo}"               scanner=[oo]          expander="$foo}"
in="literal $price} end" scanner=[rice]        expander="literal $price} end"
in="$ab:-${REAL}}"       scanner=[b def=true]  expander="$ab:-<REAL>}"
```

Two consequences:

`collectStringEnvironmentTemplates` in `resource_services.go` walks every string in the nested service config, including free-form fields, so a literal value like `costs $price}` writes a phantom `rice: ${rice}` into `services.<name>.env`.

The third case is the one that bites. The phantom `b` spans the whole string and carries `HasDefault=true`, so `collectAzureYamlEnvironmentReferences` skips it, and the real `REAL` reference (which the expander does expand) is never collected. Init doesn't prompt for it, so it reads as empty at deploy, which is the failure mode this PR is closing.

Adding the brace check fixes both. With it, `$foo}` and `literal $price} end` yield nothing and `$ab:-${REAL}}` yields `REAL[5:12]`, matching the expander. Your `env_refs_test.go` suite and the rest of `internal/cmd` still pass.

```suggestion
func environmentReferenceAt(value string, start int) (environmentReference, bool) {
	if start+1 >= len(value) || value[start+1] != '{' {
		return environmentReference{}, false
	}
	index := start + 2
```

jongio (commented):
Incremental review of the 5 commits since `fea0f498`.

`internal/cmd/env_refs.go:96` - `environmentReferenceAt` doesn't verify `value[start+1] == '{'`, so `$name}` parses as a reference with the first character dropped. That writes phantom entries into the generated `env:` block, and on `$ab:-${REAL}}` it swallows the real `REAL` reference so init never prompts for it. Repro and a fix inline; your existing tests pass with the change.

Still open from my last pass: `internal/cmd/helpers.go:1000` swallows the `GetServiceConfigValue` error, so a transient RPC failure makes a service that declared `env: {}` fall back to the full azd environment. The routines and toolboxes targets fail closed on the same call.

Replied to @trangevi on the escaping constants.

cli/azd/extensions/azure.ai.agents/internal/cmd/env_refs.go
jongio: The two constants track two different expanders, not two policies.

Fields that end up in `foundry.ExpandEnv` (agent `environmentVariables` values, connection target, toolbox endpoint) go through drone/envsubst, which collapses a `$` pair into a literal `$`. Those need `honorEnvironmentEscaping`, otherwise the scanner reports `$${VAR}` as a live reference that the expander never expands, and init would prompt for a variable nothing uses.

The project network fields (`network.agentSubnet.vnet`, `network.peSubnet.vnet`, `network.dns.subscription`) never reach that path. They're resolved by `resolveVars` in `internal/synthesis/synthesizer.go:1082`, which is a plain regex replace:

```go
out := varRefPattern.ReplaceAllStringFunc(s, func(match string) string {
```

against ``varRefPattern = regexp.MustCompile(`\$\{([A-Za-z_][A-Za-z0-9_]*)\}`)``. There's no `$$` handling anywhere in it, so `$${VNET}` does get expanded there. `ignoreEnvironmentEscaping` is what keeps discovery matching that.

So the inconsistency you're pointing at is real, it just predates this PR. It lives between `resolveVars` and `ExpandEnv`, and this file mirrors it rather than introducing it. If you want one behavior, the change is pointing `resolveVars` at `foundry.ExpandEnv`, and then the second constant deletes itself.

cli/azd/extensions/azure.ai.agents/internal/cmd/env_refs_test.go
```diff
@@ -0,0 +1,287 @@
...
+		"$${ESCAPED}",
+		"$$${DOUBLE_ESCAPED}",
+		"$$$${TRIPLE_ESCAPED}",
+		"$${OUTER:-${INNER}}",
+		"${{connections.store.credentials.key}}",
+		"${{ tools.${RESERVED} }}",
+		"${MISSING:-${{event.body}}}",
+		"${{f.g}}${AFTER}",
+		"${BEFORE}${{f.g}}",
+		"https://${HOST}/v1/${PATH:-default}",
```
jongio, line 283: Every value here has `$` immediately followed by `{`, so this never exercises the `environmentReferenceAt` case from the thread on line 107 of `env_refs.go`, where a `$` followed by a name character and a later `}` parses as a reference with the character after the `$` swallowed as if it were the brace.

That case fails the assertion this test already makes, in the direction it already checks. Running the scanner and `foundry.ExpandEnv` over the same inputs on this HEAD:

```
in="$foo}"               scanner=[oo]  expander leaves it literal
in="prefix $bar} suffix" scanner=[ar]  expander leaves it literal
```

Dropping either string into `values` gives `scanner reported "oo" but the expander never resolves it`.

Worth adding alongside the parser fix so this guard covers the shape it was written to catch.

cli/azd/extensions/azure.ai.agents/internal/cmd/env_refs_test.go
```diff
@@ -0,0 +1,287 @@
...
+		require.Equal(t, map[string]string{
+			"BARE":         "${BARE}",
+			"WITH_DEFAULT": "${WITH_DEFAULT}",
+		}, environment)
+	})
+}
+
+// TestNestedDefaultIsNotDiscovered pins the agreed limitation:
+// azd nested references are unsupported, so only the outer name
+// reaches the generated service env block and init prompting.
```
jongio: This says the outer name reaches init prompting, but the assertion four lines down is `require.Empty(t, references)`. `OUTER` has a default, so the init policy drops it too and nothing reaches prompting for this value. The env block is the only path that keeps the outer name.

jongio (commented):
Two notes on `env_refs_test.go`.

`TestFindEnvironmentReferencesMatchesExpander` is the guard for the scanner/expander contract, and every value in its list has `$` immediately followed by `{`. That shape never reaches the `environmentReferenceAt` path from the open thread on line 107 of `env_refs.go`, so the guard passes while the contract it states is already broken.

The doc comment on `TestNestedDefaultIsNotDiscovered` says the outer name reaches init prompting, but the assertion below it says otherwise.

cli/azd/extensions/azure.ai.agents/internal/cmd/env_refs.go
huimiu: Documented the limitation instead of recursing

cli/azd/extensions/azure.ai.agents/internal/cmd/env_refs.go
huimiu: It's about who expands the field. Most go through `foundry.ExpandEnv`, which leaves `$${VAR}` alone, so the scanner skips it too. The three project `network:` fields go through `resolveVars`, a plain regex replace that does expand `$${VAR}`, so the scanner has to follow.

You're right it shouldn't differ. The comment now names the fields, and #9350 tracks unifying them. Happy to pull that in here if you'd rather not ship the split

cli/azd/extensions/azure.ai.agents/internal/cmd/env_refs.go
```diff
@@ -0,0 +1,224 @@
...
+	"github.com/azure/azure-dev/cli/azd/pkg/foundry"
+)
+
+// Escape handling must match the expander that owns each field.
+// Fields resolved by foundry.ExpandEnv take
+// honorEnvironmentEscaping: it collapses '$' pairs, so $${VAR}
+// stays literal, and it reserves ${{...}} spans for Foundry.
+// The three project network fields (network.agentSubnet.vnet,
+// network.peSubnet.vnet, network.dns.subscription) take
+// ignoreEnvironmentEscaping because resolveVars in the projects
```
jongio, line 20: `resolveVars` diverges on `:-` too, not just `$$`, and that's the half that bites.

`varRefPattern` is `\$\{([A-Za-z_][A-Za-z0-9_]*)\}`, so `${VNET:-fallback}` doesn't match it at all. Running the scanner and that regex over this HEAD:

```
scanner("${VNET:-fallback}", ignoreEnvironmentEscaping) = [{VNET HasDefault:true}]
init prompting                                          = []
varRefPattern match                                     = false
```

That contradicts the policy line in `findEnvironmentReferences` just below, "init prompting skips references with a default because the expander supplies the fallback". For these three fields nothing supplies it. `resolveVars` returns the string unchanged with no error, and `containsVarRef` is that same regex so it's false too, so the value skips the "unexpanded `${VAR}`, validate at provision time" branch and lands in `vnetIDPattern` as `"${VNET:-fallback}" is not a well-formed Microsoft.Network/virtualNetworks id`. `network.dns.subscription` takes the same shape through `normalizeSubscription`. Init never prompted for `VNET` either, so there's nothing the user could have set.

Pre-existing, and #9350 covers it. Since this comment is what the next reader will take as the full story on the split, worth naming the `:-` case next to the `$$` one and pointing at #9350 the way the rest of cli/azd does it (`// See: https://github.com/Azure/azure-dev/issues/6530` in `cmd/container.go:1048`).

jongio (commented):
One thing the escaping comment leaves out. `resolveVars` diverges on `:-` as well as on `$$`, and that half breaks the scanner's own policy line for the three `ignoreEnvironmentEscaping` fields.

trangevi (approved):
Approved, pending addressing Jon's remaining comments

cli/azd/extensions/azure.ai.agents/internal/cmd/env_refs.go
jongio: Ran this on this HEAD and it's worse than a doc mismatch, the `:-` half fails silently:

```
resolveVars("${VNET:-fallback}", {"VNET": "/subscriptions/x/vnets/real"}) = "${VNET:-fallback}", err = nil
containsVarRef("${VNET:-fallback}")                                       = false
```

`varRefPattern` is `\$\{([A-Za-z_][A-Za-z0-9_]*)\}`, so a `:-` default never matches it. The value isn't substituted even though `VNET` is set, and `containsVarRef` returns false, so the unresolved-reference error doesn't fire either. The literal `${VNET:-fallback}` lands in the Bicep param.

The scanner does report `VNET` for that value, so init prompts for it and the env block records it. The user answers a prompt for a value that then gets thrown away.

Only the three `network:` fields on `ignoreEnvironmentEscaping` are affected. Worth either teaching `varRefPattern` about `:-`, or moving `resolveVars` to `foundry.ExpandEnv`, which is where the comment above already says this ends up.

jongio (commented):
Seven threads from my earlier passes are still open on this HEAD with no replies. Listing them so they don't get lost now that the PR is showing approved.

- `env_refs.go` `resolveVars` divergence: I added detail on that thread. The `:-` half fails silently, and a value the user was prompted for ends up in the Bicep param as literal text.
- `helpers.go` the service env read swallows the RPC error and leaves `hasServiceEnvironment` false, so the run path falls back to the full azd environment, which is the leak this PR closes. Routines and toolboxes return the error instead.
- `agent_definition.go` `ResolveAgentEnvironmentVariable` falls through to the caller `mapping` for any name a declared `env:` doesn't carry, and both callers hand it the full azd environment.
- `resource_services.go` `setServiceEnvironment` returns before writing when collection comes back empty, so a service with no `${VAR}` references gets no `env:` key and reads as legacy at run time.
- `env_refs.go` `environmentReferenceAt` never checks that the byte after `$` is `{`. `"$HOME}"` reports `OME` while `foundry.ExpandEnv` leaves the value untouched, which is the exact invariant `TestFindEnvironmentReferencesMatchesExpander` asserts.
- `env_refs_test.go` every value in that parity corpus has `$` followed by `{`, so it can't catch the case above.
- `env_refs_test.go` the `TestNestedDefaultIsNotDiscovered` comment says the outer name reaches init prompting, but the assertion four lines down is `require.Empty`.

None of these are new asks, they're the same threads. The two scanner ones and the two test ones are narrow. The three that fall back to the full azd environment are the ones I'd want settled before this merges, since they all undo the isolation the PR is adding.

cli/azd/extensions/azure.ai.agents/internal/cmd/env_refs.go
```diff
@@ -109,7 +117,17 @@ func environmentReferenceCandidates(value string, honorEscaping bool) []environm
...
 }
 
+// environmentReferenceAt parses the reference opening at start.
+// The '{' check is what keeps a bare '$' from being read as one:
+// without it "$foo}" reports "oo", swallowing the 'f' as if it
+// were the brace, while foundry.ExpandEnv leaves the value
+// alone. "$ab:-${REAL}}" is the costly shape, because the
+// phantom would span the string and hide the live REAL.
 func environmentReferenceAt(value string, start int) (environmentReference, bool) {
+	if start+1 >= len(value) || value[start+1] != '{' {
```
trangevi: Maybe regex would be easier here? and in some of the other places where we're introducing these kinds of checks. Idk for sure that it would be, just throwing it out there to consider

trangevi (approved):
Approving, pending addressing Jon's comments, and with a possible suggestion to consider now or as a follow up, if it would indeed make things easier.

jongio (commented):
Four earlier issues are still open on this HEAD:

- `foundry_provisioning_provider.go`: on-disk and ejected provisioning still substitutes only the project environment, so service aliases can remain unresolved or bind project-wide values.
- `resource_services.go`: init still skips writing `env: {}` when no references are collected, so generated services fall back to the full azd environment locally.
- `foundry_provisioning_provider.go`: subscription and location prompting still happens before project service validation, obscuring configuration errors, especially under `--no-prompt`.
- `foundry_provisioning_provider.go`: the broadened `$ref` failure still reports an existing-endpoint error for greenfield service configuration.

These paths are unchanged by the latest commits, so I can't approve this revision yet.

<sub>🤖 agent jongio</sub>

cli/azd/extensions/azure.ai.agents/internal/cmd/init.go
```diff
@@ -2956,6 +2957,14 @@ func (a *InitAction) addToProject(ctx context.Context, targetDir string, agentMa
 	if _, err := a.azdClient.Project().AddService(ctx, req); err != nil {
 		return fmt.Errorf("adding agent service to project: %w", err)
 	}
+	if err := setServiceEnvironment(
```
jongio: `addResourceService` moved to `setServiceEnvironmentScope` so an empty map still writes `env: {}`, but the two agent init paths still call `setServiceEnvironment`, which returns early on `len(environment) == 0`.

When `agentEnvironment` comes back empty, no `env:` key is written. `serviceEnvDeclared` then reports `Found=false`, `ServiceRunContext.HasServiceEnvironment` is false, and `mergeAgentRunEnvironment` at `run.go:1098` takes the `!hasServiceEnvironment` branch and copies the whole azd environment into the agent process. That's the fallback this PR closes everywhere else.

`init_from_code.go:873` has the same call. Is the agent path intentionally left on the legacy fallback, or should both move to `setServiceEnvironmentScope` the way `addResourceService` did?

jongio (commented):
One gap left over from the empty-env change.

- `init.go:2960` and `init_from_code.go:873`: agent init still goes through the guarded `setServiceEnvironment`, so an agent with no collected variables gets no `env:` key and falls back to the full azd environment at run time.

cli/azd/extensions/azure.ai.agents/internal/cmd/env_refs.go
huimiu: Updated this to use an anchored regex for the reference prefix.

cli/azd/extensions/azure.ai.agents/internal/cmd/resource_services.go
```diff
@@ -296,6 +306,139 @@ func addResourceService(
...
+
+// setServiceEnvironmentScope declares the service's env: scope and
+// writes an explicit env: {} for an empty map. Generated services
+// use this so one with no variables of its own is still isolated
+// instead of reading as legacy and inheriting the whole azd
+// environment at run and deploy.
+//
+// It writes raw templates through the config RPC because
+// AddService treats ServiceConfig.Environment as expanded literals.
+func setServiceEnvironmentScope(
```
jongio: This writes `env: {}` when the map is empty, but that doesn't survive persistence, so the isolation it's meant to declare never reaches azure.yaml.

`SetServiceConfigSection` does `cfg.Set("services.<name>.env", {})` and then `project.SaveConfig`. `SaveConfig` re-parses the raw map into a typed `ProjectConfig` and re-marshals that, and `ServiceConfig.Environment` is tagged `yaml:"env,omitempty"` in `pkg/project/service_config.go:68`. `ExpandableMap` is a plain map, so a zero-length one gets dropped before the file is written.

I ran it against the real `LoadConfig`/`SaveConfig` to be sure:

- `cfg.Set("services.myagent.env", map[string]any{})` followed by `SaveConfig` produces an azure.yaml with no `env:` key at all, and a follow-up raw `Get("services.myagent.env")` returns `found=false`.
- The same probe with `{"FOO": "${BAR}"}` round-trips fine and writes `env:` with the value, so this is specific to the empty map.

That lands right where the read side looks. `serviceEnvDeclared` in `helpers.go` returns `resp.GetFound()` from `GetServiceConfigValue(Path: "env")`, which is now false, so `HasServiceEnvironment` is false and `mergeAgentRunEnvironment` takes the `!hasServiceEnvironment` branch and forwards the full azd environment into the agent process. That's the fallback this change is trying to close.

The new tests pass because `recordingProjectServer` captures the RPC section in memory and never round-trips through `SaveConfig`. They prove the call sites send an empty section, which was the original bug, but they can't catch the serialization loss underneath.

`addResourceService` is affected the same way, so this covers every Foundry service generated without variables of its own, not just the two init paths.

Closing it needs a core change so an explicitly declared empty env survives the save. Dropping `omitempty` on that field is the smallest version, though it'd start emitting `env: {}` for services that never declared one, so distinguishing absent from explicitly-empty may be the better shape. Either way it'd be worth a test in `pkg/project` that asserts against the written azure.yaml instead of a stub, since that's the layer that actually decides this.

jongio (commented):
One issue with the empty-env fix in fc54476.

HIGH
- `setServiceEnvironmentScope` sends `env: {}` over the config RPC, but core drops it before it reaches azure.yaml, so a generated agent service with no variables of its own still reads back as legacy and inherits the whole azd environment. Details inline on `resource_services.go`.

Both init call sites now route through the new function and the tests do pin that, so the gap isn't in this commit. It's one layer down, in how `ServiceConfig.Environment` is serialized.

jongio (approved):
Incremental change since my last review checks out. The unpersistable `env: {}` write is gone, both init paths and `addResourceService` route through `setServiceEnvironment`, and the comment records why an explicit empty env can't survive the save today. Build, vet, and the agents cmd suite pass locally against 88a6b76.

One optional follow-up: this behavior flipped twice in the PR, so a small test pinning that an empty environment writes no `env:` key would keep it from coming back by accident. `resource_services_test.go` is the natural home now that `TestAddResourceServiceWritesEmptyEnvironment` is gone.

trangevi (approved):

jongio (approved):
Re-verified against 88a6b76: the shared `${VAR}` scanner (escape collapsing so `$${VAR}` stays literal while `$$${VAR}` expands, balanced-default depth counting, and the probe-based check that keeps Foundry `${{...}}` spans reserved), the Initialize reordering that resolves the azd environment before reading service environments, and the run-time precedence chain: process env, then service env, then the FOUNDRY_* platform values, then legacy definition env, with the Windows case-insensitive match preserved.

No blocking issues found.

PR #9080, opened 2026-07-10
[DRAFT] Managed Harness Agents

kshitij-microsoft:


cli/azd/extensions/azure.ai.agents/internal/project/service_target_prompt.go
```diff
@@ -0,0 +1,651 @@
...
+
+// ProjectEndpointAPIVersion is the api-version used by the Foundry project
+// data-plane managed agent endpoints
+// (https://<account>.services.ai.azure.com/api/projects/<project>/agents?api-version=v1).
+const ProjectEndpointAPIVersion = "v1"
+
+// ResolvePromptTargetFromEnv applies azd environment-derived overrides to the
+// prompt settings so both deploy and the lifecycle commands (show/invoke/list/
+// delete) target the same managed agent route.
+//
```
jongio, line 317: This recover only logs and then re-panics, so the panic still tears down the extension process instead of returning an error to azd. Either return an error here (deploy handlers are expected to) or drop the recover, since as written it doesn't change the outcome.

jongio (commented):
A few things to sort out before this leaves draft.

Committed artifacts that look accidental:

- `docs/specs/managed-harness-agents/~$naged-agents-getting-started.docx` is a Word lock/owner temp file. It shouldn't be tracked. Delete it and add `~$*` to `.gitignore`.
- `cli/azd/extensions/azure.ai.agents/my-prompt-agent-0701-02/` looks like generated scratch output: timestamped folder name, a full infra tree including a 1236-line generated `applicationinsights-dashboard.bicep`. Is this meant to ship, or is it leftover from a local run? If it's scratch, drop the whole directory.
- `spec.docx` and `managed-agents-getting-started.docx` are binaries committed next to their Python generators. Do you want the binaries tracked, or just the generators plus the `.md`?

Registry:

- The new `azure.ai.agents` entry points every artifact URL at `github.com/kshitij-microsoft/azure-dev/releases/...` (a personal fork). Every other entry in this file uses `github.com/Azure/azure-dev/releases/...`. This needs to point at official releases before merge, otherwise azd resolves this extension from a fork.

PR hygiene:

- The description is empty and the title is `[DRAFT]` but the PR isn't marked as a draft. Add a description and flip it to draft if it isn't ready for review.

jongio (commented):
Follow-up after the latest commit. The accidental scratch artifacts (the `my-prompt-agent-0701-02/` tree, the `~$` Word lock file, and the committed `.docx` binaries) are gone now. Two items from my earlier review are still open.

The registry.json entry for `azure.ai.agents` still points every artifact URL at `github.com/kshitij-microsoft/azure-dev/releases/...` (9 URLs). Every other entry in that file uses `github.com/Azure/azure-dev/releases/...`. This has to point at official Azure/azure-dev releases before merge, otherwise azd resolves this extension from a personal fork.

The PR is still titled `[DRAFT]` with an empty description but isn't marked as a draft. Add a description and flip it to draft until it's ready for review.

cli/azd/extensions/azure.ai.agents/internal/project/prompt_skills.go
```diff
@@ -0,0 +1,402 @@
...
+	rest = strings.TrimLeft(rest, "\r\n")
+	end := strings.Index(rest, "\n---")
+	if end < 0 {
+		return frontmatterResult{}, fmt.Errorf("unterminated frontmatter block")
+	}
+	front := rest[:end]
+	// The body starts after the closing `---` line.
+	after := rest[end+len("\n---"):]
+	after = strings.TrimPrefix(after, "-")   // tolerate longer --- fences
+	after = strings.TrimLeft(after, "-\r\n") // consume the rest of the fence line
```
jongio, line 213: `TrimLeft(after, "-\r\n")` crosses the newline after the closing `---` and keeps consuming `-`, so when a SKILL.md body's first line starts with a dash (a `- bullet` list, or a `---` break) the leading dash is stripped from the instructions: `- first bullet` becomes ` first bullet`. That mangled body flows into `skillMeta.Instructions` and ships to `CreateSkillVersion`. Trim only the remainder of the fence line (cut to the next `\n`) instead of a cut set that mixes `-` with newlines.

jongio (commented):
Went deeper on the new Go code this pass. Two correctness issues.

One is inline on `prompt_skills.go`.

The other is in `parseAgentEndpoint` (`cli/azd/extensions/azure.ai.agents/internal/cmd/agent_endpoint.go`, around line 167). The new bypass path lets a dev pass `http://localhost:5000` (scheme relaxed, port allowed), but the endpoint is then rebuilt as `fmt.Sprintf("https://%s/api/projects/%s", host, projectSegment)` where `host` is `u.Hostname()`. That forces the scheme back to `https` and drops the port, so an override like `http://localhost:5000` resolves to `https://localhost/...` and requests never reach the local backend the override is meant to target. Preserve `u.Scheme` and `u.Host` (host+port) in the bypass case, the way `validateProjectEndpoint` does in `project_endpoint.go`.

jongio (commented):
Re-reviewed at HEAD e991e5f. This commit (the ManagedAgent to PromptAgent rename plus the create-or-update version fallback) doesn't touch the items from my earlier reviews, and they're all still open:

- registry.json points every azure.ai.agents artifact URL at github.com/kshitij-microsoft/azure-dev/releases/... instead of Azure/azure-dev, so azd would resolve this extension from a personal fork.
- parseAgentEndpoint (cli/azd/extensions/azure.ai.agents/internal/cmd/agent_endpoint.go, around line 167) still rebuilds the endpoint as https://{u.Hostname()}/api/projects/..., which forces the scheme back to https and drops the port. An http://localhost:5000 override resolves to https://localhost/... and never reaches the local backend it's meant to target.
- My inline comments on service_target_prompt.go (the recover that logs and then re-panics) and prompt_skills.go (TrimLeft(after, "-\r\n") crossing the closing-fence newline and stripping a leading dash from the skill body) are unchanged.

Still titled [DRAFT] with an empty description and not marked as a draft.

cli/azd/extensions/registry.json
```diff
@@ -5009,6 +5009,252 @@
...
+            }
+          ],
+          "artifacts": {
+            "darwin/amd64": {
+              "checksum": {
+                "algorithm": "sha256",
+                "value": "8cae1b35438b2a79fed0b0b29fd423bffa0ced33b399dddb64042ea0b76b1ff2"
+              },
+              "entryPoint": "azure-ai-agents-darwin-amd64",
+              "url": "https://github.com/kshitij-microsoft/azure-dev/releases/download/azd-ext-azure-ai-agents_0.1.44-preview/azure-ai-agents-darwin-amd64.zip"
```
jongio, line 5435: This new 0.1.44 entry points all six artifact URLs at github.com/kshitij-microsoft/azure-dev/releases/... (a personal fork), the same as the earlier entries I flagged. Every other extension in this file resolves from github.com/Azure/azure-dev/releases/... These need to point at official Azure/azure-dev releases before merge, otherwise azd resolves this extension from a fork.

jongio (commented):
The two new commits bump the version to 0.1.44 and add a matching registry entry. That new entry re-adds the fork artifact URLs I flagged earlier (inline below). The other open items from my prior reviews aren't touched by these commits and are still open: the parseAgentEndpoint scheme/port drop in agent_endpoint.go, the recover-then-repanic in service_target_prompt.go, the TrimLeft in prompt_skills.go, and the [DRAFT] title plus empty description on a PR that isn't marked as a draft.

cli/azd/extensions/registry.json
```diff
@@ -5009,6 +5009,334 @@
...
+            }
+          ],
+          "artifacts": {
+            "darwin/amd64": {
+              "checksum": {
+                "algorithm": "sha256",
+                "value": "38e1e008a153296a40c956c77546039e3a551e4e8c211e1f43ec72058ee6c516"
+              },
+              "entryPoint": "azure-ai-agents-darwin-amd64",
+              "url": "https://github.com/kshitij-microsoft/azure-dev/releases/download/azd-ext-azure-ai-agents_0.1.45-preview/azure-ai-agents-darwin-amd64.zip"
```
jongio, line 5517: This new 0.1.45 entry points all six artifact URLs at github.com/kshitij-microsoft/azure-dev/releases/... (a personal fork), same as the earlier entries. Every other extension in this file resolves from github.com/Azure/azure-dev/releases/... This has to point at official Azure/azure-dev releases before merge, otherwise azd resolves this extension from a fork.

jongio (commented):
Re-reviewed at HEAD 9d92a95. I went through the four new commits (multi-turn memory via `previous_response_id`, the skill-bundle multipart upload, dropping the `files/` scaffold, and the 0.1.45 version bump). No new correctness issues in the Go changes. The bundle change closes the gap where only `SKILL.md` was uploaded, and the added tests cover nested `references/`, `assets/`, and `scripts/` files.

The new 0.1.45 registry entry re-adds the fork artifact URLs (inline below), so the registry problem I flagged earlier is back on this entry.

Open items from my prior reviews that these commits don't touch:

- `parseAgentEndpoint` in agent_endpoint.go still rebuilds the endpoint as `https://{u.Hostname()}/api/projects/...`, which forces the scheme back to https and drops the port, so an `http://localhost:5000` override resolves to `https://localhost/...` and never reaches the local backend.
- The recover in service_target_prompt.go still logs and then re-panics instead of returning an error to azd.
- The `TrimLeft(after, "-\r\n")` in prompt_skills.go extractFrontmatter is still there. Its impact is smaller now: after the bundle-upload refactor the parsed body is only used for the empty-instructions check, not uploaded, so a stripped leading dash no longer corrupts what the service stores. Still worth fixing so the validation sees the real body.

Still titled [DRAFT] with an empty description and not marked as a draft.

jongio (commented):
Re-reviewed at HEAD b97ede5. The one new commit since my last pass (the 1.46-preview release) adds the toolbox project-connection wiring (`ensureToolboxConnection` / `FoundryConnectionsARMClient`, and `injectMcpTool` now carrying `project_connection_id`), the `show.go` harness/endpoint/toolbox output, and seeding `AZURE_LOCATION` from the project region. I went through the new Go in `prompt_skills.go`, `foundry_connections_controlplane.go`, `show.go`, and `init_managed_foundry.go` and didn't find new correctness issues there.

The open items from my earlier reviews are all still open:

- `registry.json`: the new 0.1.46 entry re-adds the fork artifact URLs, so every `azure.ai.agents` URL from 0.1.42 through 0.1.46 (30 of them) points at `github.com/kshitij-microsoft/azure-dev/releases/...`. Every other entry in the file uses `github.com/Azure/azure-dev/releases/...`. These have to point at official Azure/azure-dev releases before merge, otherwise azd resolves this extension from a personal fork.

- `parseAgentEndpoint` (agent_endpoint.go) still rebuilds the endpoint as `https://{u.Hostname()}/api/projects/...`. The new comment on the port check even says the override path allows `http://localhost:5000` for local backends, but the rebuild drops the port and forces https, so `http://localhost:5000` still resolves to `https://localhost/...` and never reaches the local backend it targets. Preserve `u.Scheme` and `u.Host` (host+port) in the bypass case, the way `validateProjectEndpoint` does.

- `service_target_prompt.go` `deployPromptAgent` still recovers, logs the stack, then re-panics instead of returning an error to azd.

- `prompt_skills.go` `extractFrontmatter` still does `TrimLeft(after, "-\r\n")`, which crosses the newline after the closing fence and strips a leading dash from the body when it starts with a list item. Smaller impact now that the parsed body only feeds the empty-instructions check, but still worth fixing so validation sees the real body.

Still titled `[DRAFT]` with an empty description and not marked as a draft. Add a description and flip it to draft until it's ready for review.

jongio (commented):
This is still marked [DRAFT] and currently conflicts with main (mergeable state is dirty), so it can't merge as-is. The blockers I flagged earlier are still open at the current head and need to land before this goes in:

- `cli/azd/extensions/registry.json`: the azure.ai.agents entries for versions 0.1.42 through 0.1.46 still point their per-platform artifact URLs at `github.com/kshitij-microsoft/azure-dev/releases/...` (a personal fork). Every other extension in this file resolves from `github.com/Azure/azure-dev/releases/...`, so as written azd would pull these binaries from a fork instead of official Azure/azure-dev releases. All of these need to point at the official releases before merge.

- `internal/project/service_target_prompt.go`: the deferred recover in `deployPromptAgent` logs the panic and stack, then re-panics. It doesn't convert the panic into an error, so the extension process still tears down instead of returning a deploy error to azd. Either return an error from the handler or drop the recover, since as written it doesn't change the outcome.

- `internal/project/prompt_skills.go`: `TrimLeft(after, "-\r\n")` crosses the newline after the closing fence and keeps consuming `-`, so a SKILL.md body whose first line starts with a dash (a bullet or a `---` break) loses its leading dash before it reaches `CreateSkillVersion`. Trim only to the end of the fence line instead of a cut set that mixes `-` with newline characters.

Nothing new beyond what's already inline. Resolve the conflicts and repoint the registry URLs, then I'll take another pass.

jongio (commented):
This is still marked `[DRAFT]` in the title and currently conflicts with `main`, so it isn't mergeable in its current state.

The blocking items from my earlier reviews are all still present at the current head (no new commits have landed since my last pass):

1. `registry.json` publishes every `azure.ai.agents` artifact (0.1.42 through 1.46-preview) from `github.com/kshitij-microsoft/azure-dev/releases/...`, a personal fork. Every other extension in this file resolves from `github.com/Azure/azure-dev`. These URLs need to point at the Azure/azure-dev releases before this can ship.

2. `deployPromptAgent` in `service_target_prompt.go` recovers from a panic, logs the stack, then re-panics with `panic(r)`. That still tears down the extension process instead of returning an error, which is what deploy handlers are expected to do. Either return an error here or drop the recover.

3. `extractFrontmatter` in `prompt_skills.go` runs `TrimLeft(after, "-\r\n")` right after the closing `---`. It crosses the fence's trailing newline and keeps consuming `-`, so a SKILL.md whose body starts with a `- bullet` (or a `---` break) loses its leading dash. Bound the trim to the fence line instead of the charset `-\r\n`.

Clearing the draft/conflict state and these three items is what's needed for the next pass.

jongio (commented):
This is still marked `[DRAFT]` and currently conflicts with `main` (mergeable state is dirty), so it can't merge as-is. Nothing has landed since my last pass, so the blocking items are unchanged and still open at the current head `b97ede5`.

1. `cli/azd/extensions/registry.json` publishes every `azure.ai.agents` artifact (0.1.42 through 1.46-preview) from `github.com/kshitij-microsoft/azure-dev/releases/...`, a personal fork. Every other extension in this file resolves from `github.com/Azure/azure-dev`, so as written azd pulls these binaries from a fork. These URLs need to point at the Azure/azure-dev releases before this ships.

2. `deployPromptAgent` in `service_target_prompt.go` recovers from a panic, logs the stack, then calls `panic(r)` again. The re-panic tears down the extension process instead of returning an error, which is what deploy handlers are expected to do. Either return an error here or drop the recover.

3. `extractFrontmatter` in `prompt_skills.go` runs `TrimLeft(after, "-\r\n")` right after the closing `---`. That crosses the fence's trailing newline and keeps consuming `-`, so a SKILL.md whose body starts with a `- bullet` (or a `---` break) loses its leading dash. Bound the trim to the fence line instead of the `-\r\n` charset.

The open inline threads from earlier passes still apply too, in particular the `workspace_create.go` `GetByID` checks that treat any non-404 error (auth, throttling, transient) as "not found" and create anyway, the hard-coded `va-dev-fdp-resource` development endpoint used as the public `DefaultPromptModelEndpoint`, and `provision: true` being documented as creating the backing resource while the resolver is a no-op. Clearing the draft and conflict state plus these items is what's needed for the next pass.

cli/azd/extensions/azure.ai.agents/internal/pkg/agents/agent_api/managed_operations.go
```diff
@@ -0,0 +1,681 @@
...
+		// (http://localhost:5000) but still validates a bearer token. azcore
+		// refuses to attach credentials to non-TLS endpoints unless this is
+		// explicitly opted into, so allow it when (and only when) the base URL
+		// is http — production https endpoints keep the default protection.
+		var bearerOpts *policy.BearerTokenOptions
+		if parsed.Scheme == "http" {
+			bearerOpts = &policy.BearerTokenOptions{
+				InsecureAllowCredentialWithHTTP: true,
+			}
+		}
```
jongio, line 115: This gates on the scheme alone, so any `http://` base URL gets the Entra token attached over cleartext, not just loopback. The comment scopes the intent to `http://localhost:5000`, but a shared dev backend, a typo, or a config value pointing at `http://some-host` would put a token scoped to `https://ai.azure.com/.default` on the wire in the clear, where it can be captured and replayed. The hosted client in `operations.go` passes `nil` here and keeps azcore's default protection.

Gate on the host rather than the scheme:

```suggestion
		var bearerOpts *policy.BearerTokenOptions
		if parsed.Scheme == "http" {
			if host := parsed.Hostname(); host == "localhost" || host == "127.0.0.1" || host == "::1" {
				bearerOpts = &policy.BearerTokenOptions{
					InsecureAllowCredentialWithHTTP: true,
				}
			}
		}
```

cli/azd/extensions/azure.ai.agents/internal/project/prompt_files.go
```diff
@@ -0,0 +1,253 @@
...
+		}
+		obj, err := b.client.UploadFile(ctx, f.Name, f.Content, "assistants")
+		if err != nil {
+			return "", fmt.Errorf("uploading %q: %w", f.Name, err)
+		}
+		b.uploaded[f.Hash] = obj.Id
+		fileIDs = append(fileIDs, obj.Id)
+	}
+
+	if strings.TrimSpace(reuseStoreID) != "" {
```
jongio, line 229: `reuse` is always empty on a real deploy, so this early return is dead in the live path. `newPromptGraph` initializes `bindings` to an empty map and `vector_store_id` is only ever written a few lines below, inside this same node's `Resolve`. That means `CreateVectorStore` runs on every deploy and mints a brand new store each time, orphaning the previous store and every file object it referenced. Re-deploying in a loop grows storage and quota in the user's Foundry project with nothing in azd able to reclaim it.

The `vectorStoreBuilder` doc comment says unchanged files are skipped by hash and an existing store is reused rather than recreated, but `b.uploaded` lives on the builder and the builder is constructed fresh in `Resolve`, so the hash dedupe only ever applies within a single deploy.

The reuse branch is also wrong if it ever does go live: it uploads every file first, then returns `reuseStoreID` without attaching `fileIDs` to that store, so newly added documents get uploaded and are silently never searchable.

Seeding the binding from the deployed agent's existing `file_search` tool, or persisting the store id in the azd environment, plus an add-files call for the reuse path, would make this idempotent.

cli/azd/extensions/azure.ai.agents/internal/cmd/invoke_managed.go
```diff
@@ -0,0 +1,187 @@
...
+			data := strings.TrimSpace(strings.TrimPrefix(line, "data:"))
+			if event == "response.output_text.delta" {
+				var payload struct {
+					Delta string `json:"delta"`
+				}
+				if err := json.Unmarshal([]byte(data), &payload); err == nil && payload.Delta != "" {
+					fmt.Fprint(w, payload.Delta)
+					wroteText = true
+				}
+			} else if strings.HasPrefix(event, "response.") {
```
jongio, line 165: This handles `response.output_text.delta` for text and `response.*` for the id, but the Responses stream also emits `event: error` and the terminal `response.failed` / `response.incomplete` states. None of those are detected, so when the harness fails partway through generation the loop just ends, `scanner.Err()` is nil, and `invoke` prints nothing and exits 0. A failed run is indistinguishable from a successful empty one, which is rough to debug and worse in CI.

`TestStreamManagedSSE_NoText` currently locks that behavior in by asserting a stream with no deltas is a success.

Detecting the `error` event plus the two terminal failure states and returning an error from here would let the command exit non-zero with something actionable.

cli/azd/extensions/azure.ai.agents/internal/project/service_target_agent.go
```diff
@@ -339,6 +357,16 @@ func (p *AgentServiceTargetProvider) Endpoints(
 	serviceConfig *azdext.ServiceConfig,
 	targetResource *azdext.TargetResource,
 ) ([]string, error) {
+	// Prompt agents expose a single workspace-rooted Responses endpoint on the
+	// harness. Build it from the service config rather than azd env vars.
+	if p.isPromptAgentService() {
+		settings, err := p.promptAgentSettings()
```
jongio, line 419: `promptAgentSettings()` only applies env overrides and validates. It doesn't run `ResolvePromptTargetFromEnv` or the project overlay that `deployPromptAgent` runs before it builds this same endpoint.

In the "provision a new Foundry project" flow the project endpoint and workspace only exist in the azd environment, so this returns the placeholder tuple and `promptAgentResponsesEndpoint` emits something like `.../subscriptions/00000000-0000-0000-0000-000000000001/.../workspaces/test-ws/...`, while deploy registered the real `AGENT_*_ENDPOINT`. The two disagree for the same service.

Loading env values and applying the same resolution here, or just reading back the registered `AGENT_<key>_ENDPOINT`, would keep `azd show` consistent with what deploy actually published.

cli/azd/extensions/azure.ai.agents/internal/project/service_target_prompt.go
```diff
@@ -0,0 +1,705 @@
...
+	}
+
+	// Overlay the provisioned Foundry project values from the azd environment
+	// onto any settings still at their default placeholder. This makes the
+	// "create a new Foundry project" init path work: `azd up` provisions the
+	// project, and the deploy targets it. The overlay is a no-op unless the azd
+	// environment actually holds a resolved project (AZURE_AI_PROJECT_NAME),
+	// so the local-dev fake tuple is preserved when no project was provisioned.
+	projectScopedTarget := false
+	if env, envErr := p.azdEnvValues(ctx); envErr == nil {
```
jongio, line 198: When `azdEnvValues` fails, this whole block is skipped, and that also skips `ResolvePromptTargetFromEnv` along with its `AZURE_AI_PROJECT_ID` validation. Deploy then carries on with whatever `promptAgentSettings()` produced, which for a freshly scaffolded service is the placeholder tuple: the `...0001` subscription, `test-rg`, `test-ws`. The create call goes out against a workspace that was never meant to be real and the user gets a WorkspaceNotFound, when the actual problem is that the environment couldn't be read. Line 247 drops the same error a second time.

Returning the env error, or at minimum failing when the settings are still sitting at their placeholder defaults, would surface the real cause.

cli/azd/extensions/azure.ai.agents/internal/cmd/delete.go
```diff
@@ -99,6 +99,15 @@ func (a *DeleteAction) Run(ctx context.Context) error {
 	}
 	defer azdClient.Close()
 
+	// Prompt (kind=managed) agents are azd services on the harness. They are
+	// torn down with the rest of the project via `azd down`, so redirect
+	// rather than calling the Foundry agent-delete path that would fail.
+	if pctx, isPrompt, pErr := resolvePromptAgentService(
```
jongio, line 105: `resolvePromptAgentService` returns `(nil, false, err)` when it has already identified the service as a prompt agent but validation or target resolution then fails. Since this branch requires `pErr == nil`, that case falls through to the hosted delete path and the user gets an endpoint-resolution error instead of the validation error that actually stopped it.

`show.go` returns `pErr` directly. Doing the same here would keep the two paths consistent.

jongio (commented):
Went through the new managed-harness code at b97ede5. Nothing has landed since my last pass, so the three blockers from my earlier reviews are unchanged: the `registry.json` fork artifact URLs (30 of them, across 0.1.42 through 0.1.46-preview), the recover-then-repanic in `service_target_prompt.go`, and the fence trim in `prompt_skills.go` `extractFrontmatter`. This is also still titled `[DRAFT]` with an empty description, and it still conflicts with `main`.

New things I found this pass, all inline:

- `managed_operations.go` attaches the Entra token over cleartext http for any http host, not just loopback.
- Every deploy mints a new vector store and orphans the previous one along with the files it referenced.
- `invoke` exits 0 with no output when the harness stream fails.
- `Endpoints()` builds the prompt endpoint from unresolved settings, so `azd show` can report a placeholder workspace.
- A failed env read silently skips target resolution and its validation.
- `delete` swallows the prompt-agent resolution error and falls through to the hosted path.

Smaller items I didn't file inline:

- `scanFilesDir` and the skills bundle walk both read through symlinks. Skill bundles are meant to be shared, so a bundle containing a link to a file outside the bundle root gets that file's content read and uploaded. An `Lstat` check before reading would close it.
- `deletePromptAgentOnDown` in `listen.go` deletes by the `azure.yaml` service key, while every other lifecycle path uses the `agent.yaml` name, and it skips the env target overlay. That only lines up today because init keeps the two names equal.
- `createOrUpdatePromptAgent`, `waitForPromptAgentActive`, `isWorkspaceNotFoundError` and `isAgentConflictError` have no tests. Those are the retry and substring-matching paths most likely to regress without anyone noticing.

cli/azd/extensions/azure.ai.agents/internal/cmd/listen.go
```diff
@@ -70,16 +70,31 @@ func preprovisionHandler(ctx context.Context, azdClient *azdext.AzdClient, args
...
+			if isHostedAgentService(svc, args.Project) {
+				hostedAgentCount++
+			}
+			// Prompt (kind=managed) agents have no container to provision
+			// settings for — the harness owns the runtime. But they DO carry a
+			// model deployment in their service config, so still run envUpdate
+			// (which translates `deployments` into AI_PROJECT_DEPLOYMENTS for
+			// Bicep). Only the container-settings step is hosted-specific.
+			_, isPrompt := promptSettingsFromService(svc)
+			if !isPrompt {
```
jongio, line 89: This guard didn't make it to the deploy side of the merge. `predeployHandler` is now service-scoped (main's shape) and calls `prepareContainerSettings` at line 283 with no `isPrompt` check, so a kind=managed service takes the hosted container path on every `azd deploy`.

At b97ede5 that handler carried the matching skip:

```go
// Prompt (kind=managed) agents have no container settings and no
// developer-RBAC pre-flight - the harness owns the runtime.
if _, isPrompt := promptSettingsFromService(svc); isPrompt {
    continue
}
```

7150098 deleted it along with the conflict debris that had landed in `updateLegacyProjectDeployments`. It doesn't fail loudly: `LoadServiceTargetAgentConfig` tolerates the managed shape, so `SetAgentContainerSettings` writes default memory and cpu onto the service and `SetServiceConfigValue` persists them into azure.yaml for an agent whose runtime the harness owns. Every other prompt guard survived the merge, this is the only one that went missing, and the suite still passes, so nothing catches it.

Guarding the call the same way you do here would line the two handlers back up:

```go
if _, isPrompt := promptSettingsFromService(svc); !isPrompt {
    if err := prepareContainerSettings(ctx, azdClient, svc, args.Project.Path); err != nil {
        return fmt.Errorf("failed to populate container settings for service %q: %w", svc.Name, err)
    }
}
```

jongio (commented):
Incremental pass over the two new commits (2bb6450 merge of main, 7150098 merge fixes).

- `predeployHandler` lost its prompt-agent guard in the merge, so kind=managed services now run the hosted container-settings path on every deploy. Inline on listen.go.

Unchanged since my last pass at b97ede5: the registry.json fork artifact URLs (30 across 0.1.42 through 0.1.46-preview), the recover-then-repanic in service_target_prompt.go, and the fence trim in prompt_skills.go `extractFrontmatter`. Six checks are red, including golangci-lint flagging errorlint on prompt_client_test.go:296. Still titled [DRAFT] with no description.

PR #9081, opened 2026-07-10
fix: align agent workflows with unified azure.yaml

### LINKED ISSUE #8710: Migrate next step suggestions and doctor command to make it work with unified azure.yaml
Currently both next step suggestions and doctor command are dependent on agent.yaml, agent.manifest.yaml and azure.yaml file. So both need to be fixed when the unified azure.yaml changes are in.

huimiu:
## Summary

Migrate `azd ai agent doctor` and next-step guidance to the effective configuration in `azure.yaml` (#8710).

This PR is the diagnostics layer in a stacked change. It keeps doctor, next-step guidance, and the runtime behavior aligned when agent configuration is inline, uses the deprecated `config:` block, or is provided through a local `$ref`. Existing `agent.yaml` and `agent.manifest.yaml` projects remain supported.

> **Stacked on #9149.** Review #9149 first for the shared configuration and runtime read path. This PR targets `hui/unified-config-runtime`, not `main`.

### Doctor

- Read the effective unified, legacy, or referenced agent definition.
- Validate models, connections, local resources, and manual environment requirements.
- Accept supported workflow kinds and provide source-appropriate remediation.
- Keep connection ownership and provisioning checks consistent with the runtime configuration.

### Next-step guidance

- Build state from the normalized `azure.yaml` service configuration.
- Resolve local `$ref` configuration before generating guidance.
- Normalize typed YAML environment values to strings.
- Preserve referenced Foundry expressions and report incomplete resource state clearly.

## Out of scope

- Runtime, deploy, provision, and synthesis changes are in prerequisite PR #9149.
- `azure.ai.toolboxes` `$ref` support is independent PR #9148.
- This PR does not change connection ownership behavior introduced by #9046.

## Review order

1. `internal/pkg/projectconfig` and `internal/project` in #9149 for the shared effective-configuration contract.
2. `internal/cmd/doctor` for validation and remediation behavior.
3. `internal/cmd/nextstep` for normalized state and guidance generation.

## Validation

- `go build ./...`
- Targeted doctor, next-step, and command tests

### DIFF
--- a/cli/azd/extensions/azure.ai.agents/internal/cmd/resource_services.go
+++ b/cli/azd/extensions/azure.ai.agents/internal/cmd/resource_services.go
@@ -13,6 +13,8 @@ import (
 	"azureaiagent/internal/project"
 
 	"github.com/azure/azure-dev/cli/azd/pkg/azdext"
+	"github.com/azure/azure-dev/cli/azd/pkg/foundry"
+	"google.golang.org/protobuf/proto"
 	"google.golang.org/protobuf/types/known/structpb"
 )
 
@@ -355,10 +357,26 @@ func reserveServiceName(used map[string]string, name, source string) error {
 // service carries any, so an azure.yaml written before the per-resource split
 // still provisions without re-running init.
 func collectProjectDeployments(services map[string]*azdext.ServiceConfig) ([]project.Deployment, error) {
+	return collectProjectDeploymentsAtRoot(services, "")
+}
+
+func collectProjectDeploymentsAtRoot(
+	services map[string]*azdext.ServiceConfig,
+	projectRoot string,
+) ([]project.Deployment, error) {
 	var out []project.Deployment
 	for _, svc := range sortedServices(services) {
-		props := project.ServiceConfigProps(svc)
-		if svc.Host != AiProjectHost || props == nil {
+		if svc.Host != AiProjectHost {
+			continue
+		}
+		props, err := resolvedResourceServiceProps(
+			svc,
+			projectRoot,
+		)
+		if err != nil {
+			return nil, err
+		}
+		if props == nil {
 			continue
 		}
 		var cfg *project.ServiceTargetAgentConfig
@@ -372,7 +390,10 @@ func collectProjectDeployments(services map[string]*azdext.ServiceConfig) ([]pro
 	if len(out) > 0 {
 		return out, nil
 	}
-	legacy, err := collectLegacyAgentConfigs(services)
+	legacy, err := collectLegacyAgentConfigsAtRoot(
+		services,
+		projectRoot,
+	)
 	if err != nil {
 		return nil, err
 	}
@@ -387,10 +408,26 @@ func collectProjectDeployments(services map[string]*azdext.ServiceConfig) ([]pro
 // agent service when no connection service carries any, so a pre-split
 // azure.yaml still provisions without re-running init.
 func collectConnections(services map[string]*azdext.ServiceConfig) ([]project.Connection, error) {
+	return collectConnectionsAtRoot(services, "")
+}
+
+func collectConnectionsAtRoot(
+	services map[string]*azdext.ServiceConfig,
+	projectRoot string,
+) ([]project.Connection, error) {
 	var out []project.Connection
 	for _, svc := range sortedServices(services) {
-		props := project.ServiceConfigProps(svc)
-		if svc.Host != AiConnectionHost || props == nil {
+		if svc.Host != AiConnectionHost {
+			continue
+		}
+		props, err := resolvedResourceServiceProps(
+			svc,
+			projectRoot,
+		)
+		if err != nil {
+			return nil, err
+		}
+		if props == nil {
 			continue
 		}
 		var conn *project.Connection
@@ -398,13 +435,19 @@ func collectConnections(services map[string]*azdext.ServiceConfig) ([]project.Co
 			return nil, fmt.Errorf("parsing connection service %q config: %w", svc.Name, err)
 		}
 		if conn != nil {
+			if conn.Name == "" {
+				conn.Name = svc.Name
+			}
 			out = append(out, *conn)
 		}
 	}
 	if len(out) > 0 {
 		return out, nil
 	}
-	legacy, err := collectLegacyAgentConfigs(services)
+	legacy, err := collectLegacyAgentConfigsAtRoot(
+		services,
+		projectRoot,
+	)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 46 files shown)


jongio (commented):
I looked into the two `$ref` overlay comments already on this PR and the mechanism holds up. `foundry.ResolveFileRefs` overlays a directive's sibling keys on top of the referenced file wholesale (the sibling loop in `resolveRef`), so any field persisted next to `$ref` wins over that same field in the referenced file on every later run.

One thing worth adding beyond what's already noted: in `prepareContainerSettings`, `SetAgentContainerSettings` writes `container` onto the original `$ref`-bearing properties and fills in `DefaultMemory`/`DefaultCpu` whenever they're unset. `populateContainerSettings` runs for every agent service, not just ones that actually set resources. So after the first preprovision run a `$ref`-backed service ends up with a `container` block (often just the injected defaults) pinned beside `$ref`, and later edits to `container` in the referenced agent file stop taking effect. The resolved `project`/`RelativePath` write in `persistBuildService` has the same shape.

For `$ref`-backed services, persisting only the fields the user authored inline, or writing derived defaults into the referenced file through the ref-aware editor, would avoid pinning values that silently override the source file.

jongio (commented):
I re-checked the new commit against the `$ref` pinning I raised earlier. The persist-then-restore path plus not persisting resolved container settings for `$ref`-backed services closes that gap, and the added tests cover both the restore and the package-level revert.

I'm not approving yet because of a separate, unresolved issue in the connection templates. In `main.bicep` (and `brownfield.bicep`), `connections` is an ordinary array parameter:

`param connections connectionsType = []`

and `connectionType` carries `credentials: object?`. Non-secure deployment parameters are stored in the ARM deployment history in plaintext, so anyone with reader access on the deployment can recover those credentials. Moving the secret values into a `@secure()` object parameter, keeping only non-secret metadata in the array, and regenerating `main.arm.json` / `brownfield.arm.json` would close that off.

Worth also confirming the two related points already raised on this PR: the connection decoder in `synthesizer.go` dropping advanced auth properties (OAuth URLs/scopes, sharing, managed-identity), and the doctor check in `checks_connections.go` still pointing at `azd deploy` for split connections that are now provision-managed.

Once the credential handling is sorted this is good to approve.

.github/scripts/src/ext-registry-check.js
```diff
@@ -0,0 +1,807 @@
+// This script runs as part of the .github/workflows/ext-registry-check.yml. It lets the core azd team ensure that complicated
```
trangevi, line 1: Why are there so many new scripts as a part of this PR? These seem to be very azd core focused, I'm confused on why we're adding them?

cli/azd/extensions/azure.ai.connections/internal/cmd/service_target.go
```diff
@@ -5,41 +5,21 @@ package cmd
```
trangevi, line 5: Weren't these changes already done as a part of this PR: https://github.com/Azure/azure-dev/pull/9046

trangevi (changes requested):
I think there was maybe a bad merge here, I'm seeing a number of changes that I believe I've reviewed elsewhere and should already be in main.

jongio (commented):
Following up on @trangevi's bad-merge note, since it checks out and I can pin down what happened.

The merge-base between this branch and main is `105d7e9f`, and the branch is 2 commits behind main (missing #9007 and #9103, including the pin to the released azd module). The `resolve unified workflow conflicts` merge (`18dd0d2e`) pulled in changes from other in-flight PRs that have nothing to do with aligning agent workflows to unified `azure.yaml`.

The clearest example is an entire PR-registry-check tooling subsystem from #9042 that isn't in main and isn't part of this change:

- `.github/scripts/src/ext-registry-check.js` (+807), `.github/scripts/test/ext-registry-check.test.js` (+1089)
- `.github/scripts/package.json`, `package-lock.json`, `jsconfig.json`
- `.github/workflows/ext-registry-check.yml`, `.github/workflows/scripts-ci.yml`

None of those exist in main, so they're riding in from #9042 rather than belonging here. They account for a large chunk of the +8920/72-file size and make the actual unified-`azure.yaml` changes hard to isolate.

Rebasing onto current main (or dropping the merge and re-merging cleanly) so the diff only contains the agent-workflow work would resolve this and give the substantive changes a clean review. Worth a second pass over the merged-in toolbox commits (#9046, the toolbox-branch fix) to confirm which of those are genuinely part of this PR versus riding in from elsewhere.

This is separate from the credential-handling issue I raised earlier, which is still open.

cli/azd/extensions/azure.ai.agents/internal/cmd/resource_services.go
```diff
@@ -476,7 +547,20 @@ func collectLegacyAgentConfigs(services map[string]*azdext.ServiceConfig) ([]*pr
 		if project.ServiceConfigProps(svc) == nil {
 			continue
 		}
-		cfg, err := project.LoadServiceTargetAgentConfig(svc)
+		effective := *svc
```
jongio: `effective := *svc` copies the `ServiceConfig` protobuf by value. `go vet` flags it: `assignment copies lock value to effective ... protoimpl.MessageState contains sync.Mutex`. It is also an aliasing risk: the shallow copy still shares the nested `AdditionalProperties`/`Config` pointers with `svc`, so `ResolveServiceConfigInPlace(&effective, ...)` can mutate the caller's service. Use `proto.Clone(svc).(*azdext.ServiceConfig)` here, the same pattern already used in `Initialize` and `persistBuildService`.

jongio (commented):
The rebase cleaned this up, and my earlier blocking concerns are resolved:

- The branch no longer carries the #9042 registry-check tooling or the other in-flight commits it had picked up. It is down to 2 commits / 46 files, all under `cli/azd/extensions/`.
- The connection bicep templates that stored resolved credentials in a plain (non-`@secure()`) ARM array param are no longer in this diff, so the deployment-history exposure I raised earlier is out of scope here.
- The `$ref` pinning is handled now: `persistBuildService` writes the resolved path for the build, then the deferred `combineBuildServiceRestore` writes the original `$ref` config back after `Package`, and file-ref services no longer persist resolved container settings in `prepareContainerSettings`.

The three open items from the latest automated pass still hold. I checked them against the current HEAD: the `${{...}}` env values getting pre-expanded by core in the helpers/deploy paths, `ProjectRoot` not reaching `collectConnections` so split-connection `$ref` files stay unresolved, and the `env` decode rejecting non-string YAML scalars in `config.go`. Those are already captured, so I am not repeating them inline.

One net-new item below.

huimiu:

jongio (approved):
Reviewed the doctor and next-step migration onto the unified azure.yaml. I traced the tiered resource resolution (split services, then bundled config, then the legacy manifest fallback), the env-variable precedence in the config merge, and the Foundry `${{...}}` handling in `extractEnvironmentRefs` where the `$`-prefix guard correctly keeps those expressions out of both the missing-var and placeholder detectors. Build, vet, and the doctor and nextstep test suites pass locally. Moving config load and parse failures from silent skips to actionable doctor failures reads as a deliberate change, and the updated tests pin that behavior.

PR #9091, opened 2026-07-10
feat(telemetry): emit infra.provider on provision, up, and down

### LINKED ISSUE #9090: Emit infra.provider telemetry attribute on provision/up/down
### Summary

`azd` records the IaC provider (`bicep` / `terraform` / ...) via the `infra.provider` telemetry
attribute, but today it is only emitted by `azd infra generate` / `synth`. The `provision`, `up`,
and `down` commands do **not** emit it, so telemetry can't distinguish whether a provisioning run
used Bicep or Terraform.

### Details

`fields.InfraProviderKey` (`infra.provider`) is set in exactly one place today:

- `cli/azd/cmd/infra_generate.go` — `tracing.SetUsageAttributes(fields.InfraProviderKey.String(provider))`

For provisioning, the provider kind is already resolved in a single, provider-agnostic choke point:

- `cli/azd/pkg/infra/provisioning/manager.go` → `Manager.newProvider()` — `providerKey` holds the
  concrete kind after applying `azure.yaml`'s `infra.provider` (or the default provider resolver).

Since `Manager.Initialize` → `newProvider` runs for `provision`, `up`, and `down`, emitting a usage
attribute there flows onto those command spans via the existing usage-attribute mechanism
(`cmd/middleware/telemetry.go`).

### Proposed change

In `Manager.newProvider`, once `providerKey` is resolved, add:

```go
tracing.SetUsageAttributes(fields.InfraProviderKey.String(string(providerKey)))
```

This mirrors the existing `infra generate` pattern and introduces no new telemetry field.

### Scope / notes

- Covers `provision`, `up`, and `down` (all route through `Manager.newProvider`).
- `deploy` is out of scope: it does not instantiate the provisioning manager and does not resolve an
  IaC provider.
- Multi-layer provisioning calls `Initialize` per layer; setting the same usage attribute repeatedly
  is harmless (last write wins).

### Acceptance criteria

- [x] `provision`, `up`, and `down` command telemetry include `infra.provider` with the resolved provider kind.
- [x] The value matches the provider actually used (`bicep`, `terraform`, `arm`, or a custom/alpha provider).
- [x] No new telemetry field is introduced; the existing `fields.InfraProviderKey` is reused.

hemarina:
Fixes #9090

## What

Emit the `infra.provider` telemetry attribute so provisioning runs can be segmented by IaC provider.
Previously it was only emitted by `azd infra generate` / `synth`; this adds it to the provisioning
lifecycle commands `provision`, `up`, and `down`.

## Design

A shared method `(*provisioning.Manager).RecordInfraProviderUsage(layers)` computes the value from
the command's configured layers and records it as a usage attribute. It resolves the manager's
default provider (for layers that leave the provider unspecified) at most once per call, memoized
via `sync.OnceValues`.

The attribute is a **sorted, de-duplicated string slice** of the distinct providers the layers
resolve to:

- built-in kinds are recorded verbatim (`bicep` / `terraform` / `arm` / `pulumi`);
- a non-built-in (extension) provider is bucketed to `custom` — the raw name is **never** emitted
  (privacy);
- a single-provider project records a one-element slice (e.g. `["bicep"]`); a multi-layer project
  that combines providers records each distinct value (e.g. `["bicep","terraform"]`), preserving
  which providers were combined rather than collapsing to a lossy `mixed` marker.

Because bucketing happens before de-duplication, two *different* extension providers collapse to a
single `["custom"]` (the raw names are never distinguished).

> **Contract note:** this intentionally supersedes the earlier scalar `mixed` design. `infra.provider`
> is now **multi-valued** on provision/up/down (a `string[]`) and remains a scalar `string` on
> `infra generate` / `synth`. See the privacy / downstream review below.

It is invoked once per command, **early (before provider work)**, from the lifecycle actions:

- `provision`: in `ProvisionAction.Run`, before the `--preview` + multi-layer validation (covers
  preview, validation-failure, failure, and success);
- `up`: in `upAction.Run`, before manager initialization and the custom-workflow branch, so every
  `up` path is covered — including init failures and custom `workflows.up` (see below);
- `down`: before the destroy loop.

Emitting early keeps failed and preview runs segmentable by provider while staying scoped to
provision/up/down — it does **not** land on `env refresh` / `pipeline config`. `up`'s synthetic
`cmd.package` span never receives it — the value is attached directly to spans (not the
process-global usage bag), so it stays on `cmd.provision` / `cmd.up` only.

For `up`, recording happens on the common `upAction.Run` path — before manager initialization and
the custom-workflow branch — so it covers every `up` path (initialization failures, custom
`workflows.up`, and the unified graph) rather than only the unified-graph path. It is attached
directly to the `cmd.up` span (not the process-global usage bag), so a custom `workflows.up` `cmd.up`
span is tagged while the `azd package` / `provision` / `deploy` in-process child commands it spawns
keep their own provider attribution on their own spans — the parent value is never leaked onto, nor
overwritten by, those children. The unified-graph runner additionally tags the synthetic
`cmd.provision` span the same way.

`infra generate` / `synth` emit the value read from `azure.yaml`'s `infra.provider` as a single
string (or `auto` when unset), now routed through the same `custom` bucketing so a non-built-in
(extension) provider name is never emitted raw; the slice representation remains specific to the
provision/up/down path.

## Telemetry governance

- Reuses the existing `fields.InfraProviderKey` (no new field; `SystemMetadata` / `FeatureInsight`);
  refreshed its source doc comment for the value domain and scalar-vs-slice split.
- Updated `feature-telemetry-matrix.md`, `telemetry-schema.md`, and `telemetry-data.md` for the value
  domain and the provision/up/down (slice) vs generate/synth (scalar) distinction, including a
  "Data Nuances & Gotchas" note documenting the type-polymorphic, multi-valued schema.
- Corrected the `commandsWithSpecificTelemetry` comments in `telemetry_test.go`.

### Privacy / downstream review (single-value → array)

Changing `infra.provider` from scalar-only to `string | string[]` triggers the mandatory downstream
review in `docs/specs/metrics-audit/privacy-review-checklist.md`. Reviewers must confirm the
following before merge:

- [ ] Kusto functions that read `infra.provider` handle both scalar and array shapes
- [ ] Cooked tables / ingestion schema accept the array representation
- [ ] LENS jobs and dashboards that segment by `infra.provider` are updated
- [ ] Schema validation / allowed-value lists updated for the new value domain
- [ ] No `CustomerContent` exposure — extension provider names remain bucketed to `custom`

## Tests

- `TestRecordInfraProviderUsage` — single / default / uniform / different-providers-as-slice /
  custom-bucketed / two-custom→`["custom"]` / built-in+custom / empty / nil-resolver /
  failing-resolver / resolver→NotSpecified.
- `TestUsageAttributesExcluding` — the `cmd.package` span filter.
- `Test_CLI_Telemetry_NestedCommands` — asserts `infra.provider` is present on `cmd.up` /
  `cmd.provision` and absent from the synthetic `cmd.package` span (trace-level coverage).

## Verification

- `go build ./...`, `gofmt -l` clean.
- `go test ./pkg/infra/provisioning/`, `./internal/cmd/`, `./cmd/` (telemetry + up + down).


cli/azd/pkg/infra/provisioning/manager.go
```diff
@@ -530,6 +532,10 @@ func (m *Manager) newProvider(ctx context.Context) (Provider, error) {
 		providerKey = defaultProvider
 	}
 
+	// Record the resolved IaC provider (bicep / terraform / ...) as a usage attribute so it flows
+	// onto the host command span (provision / up / down), letting telemetry split runs by provider.
+	tracing.SetUsageAttributes(fields.InfraProviderKey.String(string(providerKey)))
```
jongio: This runs on the shared `newProvider` path, so `infra.provider` now also lands on `env refresh` (cmd/env.go:1234) and `pipeline config` (cmd/pipeline.go:190), which both call `provisionManager.Initialize`. `env refresh` is in `commandsWithOnlyGlobalTelemetry` with the note `Global telemetry sufficient`, and the two docs this PR edits list the emitting commands as provision/up/down only. Worth deciding whether emitting on those two extra commands is intended (then update the classification comment and docs) or whether the attribute should be scoped to the provisioning lifecycle instead.

Two smaller things on the same line. The write happens before `ResolveNamed`, and `ParseProvider` is currently a no-op (`return kind, nil`), so a mistyped or unregistered provider gets recorded as the resolved value even when resolution then fails right below. And since usage attributes are process-global, concurrent multi-layer provisioning with different per-layer providers can record whichever layer writes last rather than a deterministic value.

jongio (commented):
The provider resolves correctly, but a couple of things about where this attribute lands are worth deciding before merge.

Because it sits in the shared `newProvider` path, `infra.provider` now also lands on `env refresh` (cmd/env.go:1234) and `pipeline config` (cmd/pipeline.go:190) since both call `provisionManager.Initialize`. `env refresh` is classified as global-only in telemetry_test.go, and the docs this PR updates list the emitting commands as provision/up/down only, so the real emission scope is wider than the stated one.

docs/reference/telemetry-data.md
```diff
@@ -678,7 +678,7 @@ How to find telemetry for a given feature area. Start here if you know the featu
 | **Container Apps (Aspire)** | `cmd.deploy`, `cmd.provision` | `project.service.targets` = `containerapp-dotnet`, `platform.type` = `aca` | Aspire-specific adoption and success |
 | **Language Support** | `cmd.deploy`, `cmd.package`, `cmd.restore` | `project.service.languages`, `project.service.language` | Usage by language |
 | **Templates** | `cmd.init`, `cmd.up` | `project.template.id` (hashed — join with template lookup to resolve) | Template adoption, success by template |
-| **Provisioning (IaC)** | `cmd.provision`, `arm.deploy.*`, `arm.validate.*` | `infra.provider` (`bicep`, `terraform`) | Provision success, ARM errors, duration |
+| **Provisioning (IaC)** | `cmd.provision`, `cmd.up`, `cmd.down`, `arm.deploy.*`, `arm.validate.*` | `infra.provider` (`bicep`/`terraform`/`arm`/`pulumi`/custom, or `mixed`) | Provision success, ARM errors, duration |
```
jongio: This row still lists ARM errors as a use case for `infra.provider`, but `RecordInfraProviderUsage` only runs after provision/up/down succeed. Each caller returns before the record call when provisioning errors, so failed runs (where the ARM errors surface) won't carry `infra.provider`. `provision --preview` also returns early in provision_graph.go and never records. If you want the provider on failing and preview runs, record it on those command paths too. That stays scoped to provision/up/down, so it won't reintroduce the env refresh / pipeline config emission. Otherwise, drop ARM errors from this row so the doc matches what's emitted.

jongio (commented):
One doc/impl consistency follow-up, inline on the provisioning telemetry row.

cli/azd/internal/cmd/up_graph.go
```diff
@@ -671,6 +678,19 @@ func phaseTimingBreakdown(steps []exegraph.StepTiming) string {
 
 // phaseDurations computes the wall-clock duration for provisioning and deploying phases.
 // Returns zero durations for phases that were not executed.
+// usageAttributesExcluding returns usageAttrs without the attribute matching exclude. It keeps
+// provisioning-scoped attributes (for example infra.provider) off the synthetic cmd.package span.
+func usageAttributesExcluding(usageAttrs []attribute.KeyValue, exclude attribute.Key) []attribute.KeyValue {
```
jongio: The `usageAttributesExcluding` helper got inserted between the `phaseDurations` doc comment (lines 679-680) and the `phaseDurations` function, so that comment now documents this helper and `phaseDurations` is left with no doc comment. Move the helper above the `phaseDurations` comment (or below the function) so each doc comment stays with its function.

cli/azd/internal/cmd/up_graph.go
```diff
@@ -168,12 +169,18 @@ func (u *UpGraphAction) Run(
...
+		// cmd.package span (it belongs on cmd.provision and the parent cmd.up span only).
+		packageSpan.SetAttributes(usageAttributesExcluding(usageAttrs, fields.InfraProviderKey.Key)...)
 		packageSpan.End()
 		provisionSpan.SetAttributes(usageAttrs...)
 		provisionSpan.End()
 	}()
 
+	// Record the resolved IaC provider (single, or "mixed") on the command span up front, so it is
+	// present on success and failure alike (no-op when there are no provision layers).
+	provisioning.RecordInfraProviderUsage(layers, u.defaultProvider)
```
jongio: This covers the unified graph path, but a custom `workflows.up:` in azure.yaml returns at cmd/up.go:181 before `upGraph.Run`, so the parent `cmd.up` span will not carry `infra.provider` on that path (the spawned child `azd provision` still emits it on its own span). A failure in the earlier `Initialize` (cmd/up.go:147-156) exits before this call too. The docs this PR updates list `up` as emitting the field, so worth a caveat there or recording it on the common upAction path before those early returns.

jongio (commented):
Two low-severity items on the incremental changes, both inline on up_graph.go.
- The `usageAttributesExcluding` helper split the `phaseDurations` doc comment from its function.
- Custom `workflows.up` returns before the graph run, so `cmd.up` will not carry `infra.provider` on that path even though the updated docs list up as covered.

docs/specs/metrics-audit/telemetry-schema.md
```diff
@@ -274,7 +274,7 @@ The following fields are defined in `fields.go`.
 | Hooks kind | `hooks.kind` | SystemMetadata | FeatureInsight | Executor used to run the hook. Values: `sh`, `pwsh`, `python`, `js`, `ts`, `dotnet` |
 | Pipeline provider | `pipeline.provider` | SystemMetadata | FeatureInsight | Resolved provider display name after auto-detection: `GitHub`, `Azure DevOps` |
 | Pipeline auth | `pipeline.auth` | SystemMetadata | FeatureInsight | Emitted only when `--auth-type` is set on `pipeline config`: `federated`, `client-credentials` |
-| Infra provider | `infra.provider` | SystemMetadata | FeatureInsight | `bicep`, `terraform`, `auto` (auto-detected from files) |
+| Infra provider | `infra.provider` | SystemMetadata | FeatureInsight | Resolved IaC provider `bicep`, `terraform`, `arm`, `pulumi`, or `custom` for any other (extension) provider — the raw name is never emitted; `mixed` when a project's layers differ (provision/up/down); `auto` on `infra generate`/`synth` before resolution |
```
jongio: This row now says the raw name is never emitted, but that only holds for the provision/up/down path, where `RecordInfraProviderUsage` buckets non-built-in providers to `custom`. `infra generate`/`synth` still emit `string(projectConfig.Infra.Provider)` raw in `cmd/infra_generate.go`, and this same row lists generate/synth as emitters. So the guarantee is overbroad as written. Either scope the `raw name never emitted` wording to provision/up/down, or route generate/synth through the same bucketing helper. The identical wording is in `telemetry-data.md` and `feature-telemetry-matrix.md`.

jongio (commented):
One doc-accuracy point inline: the new `raw name never emitted` wording is broader than the code actually guarantees.

jongio (approved):
Confirmed the incremental changes address my earlier telemetry-accuracy feedback. Deduping by resolved ProviderKind before bucketing now records `mixed` for two distinct custom providers (they previously collapsed to `custom`), and dropping `Test` from the raw allowlist keeps emitted values aligned with the documented enum. The provision/up/down vs generate/synth doc scoping reads accurately now.

jongio (approved):
Re-reviewed the two follow-up commits. Dropping `ParseProvider` in favor of the explicit `NotSpecified` guard is correct: the old allowlist path silently discarded `arm`, `pulumi`, and custom kinds before they could reach the bucketing logic, and it also emitted `custom` when a resolver returned `NotSpecified`. The new guard fixes both, and caching the default resolution keeps the per-layer path deterministic without repeating resolver I/O. The added `NotSpecified`-resolution test locks in the empty-emit case.

cli/azd/cmd/down.go
```diff
@@ -67,6 +67,7 @@ type downAction struct {
 	flags               *downFlags
 	args                []string
 	provisionManager    *provisioning.Manager
+	defaultProvider     provisioning.DefaultProviderResolver
```
vhvb1989: We're adding a new `provisioning.DefaultProviderResolver` dependency to the `up`, `down` (and `provision`) constructors here, but this looks like duplication: the `*provisioning.Manager` these actions already inject holds the very same resolver (`Manager.defaultProvider`, populated in `NewManager` from the same IoC singleton).

Ideally we only add a new dependency to a constructor when it's really needed. Since the Manager already carries this resolver, could we make `RecordInfraProviderUsage` a method on `*provisioning.Manager` (using `m.defaultProvider`) instead of a free function taking the resolver? That would let us drop the new `defaultProvider` parameter from the `up`/`down` constructors (and their test constructor updates) entirely.

jongio (commented):
Re-reviewed after the latest comments. My approval was on this same commit and I still don't see anything blocking. Two notes on the open discussion:

On @vhvb1989's point about the new `defaultProvider` constructor param: it is redundant. All three actions already inject `*provisioning.Manager` (`upAction.provisioningManager`, `downAction.provisionManager`, `ProvisionAction.provisionManager`), and the Manager already carries the same resolver in its `defaultProvider` field, populated in `NewManager` from the same IoC singleton. Making `RecordInfraProviderUsage` a method on `*Manager` that reads `m.defaultProvider` would drop the extra param from all three constructors plus the test wiring. Non-blocking, but a clean simplification.

On the raw-provider note for `infra generate` / `synth`: that path emits the configured `azure.yaml` value verbatim, but that is pre-existing behavior this PR intentionally leaves unchanged, and the `fields.go` doc and the schema now spell out the split (generate/synth emit the configured value; provision/up/down bucket to `custom`/`mixed`). Extending the allowlist to generate/synth is a separate decision, not a blocker for this PR.

cli/azd/pkg/infra/provisioning/manager.go
```diff
@@ -538,3 +540,89 @@ func (m *Manager) newProvider(ctx context.Context) (Provider, error) {
...
+	// record "mixed" (bucketing only the single-provider result, not before counting).
+	switch len(seen) {
+	case 0:
+		return
+	case 1:
+		for kind := range seen {
+			tracing.SetUsageAttributes(fields.InfraProviderKey.String(infraProviderTelemetryValue(kind)))
+		}
+	default:
+		tracing.SetUsageAttributes(fields.InfraProviderKey.String(InfraProviderMixed))
```
vhvb1989: Nit / lowest priority — genuinely just curious, not blocking: is there a reason we prefer collapsing to a single `"mixed"` value here rather than recording a `StringSlice` of the actual providers used (e.g. `["bicep","terraform"]`), similar to how `ProjectServiceLanguagesKey` / `ProjectServiceHostsKey` record the per-service breakdown in `project.Parse`?

Built-in provider names are a fixed enum (only `custom` needs bucketing for privacy), so a slice would stay low-cardinality while preserving which combination was mixed. With `"mixed"` we can segment "single vs mixed" but can't tell `bicep+terraform` from `bicep+arm`.

I fully expect almost nobody to actually mix two providers across layers, so this really is a nit — just want to understand the choice.

jongio (approved):
The `address feedback` commit turns `RecordInfraProviderUsage` into a method on `*Manager` that reads `m.defaultProvider`, dropping the redundant resolver param from `newUpAction`, `newDownAction`, and `ProvisionAction` plus the test wiring. `NewManager` sets `defaultProvider` at construction, so the pre-Initialize call sites resolve through the same IoC resolver as before with no behavior change, and the nil-resolver path is still a no-op. The telemetry test moving into the `provisioning` package to build `&Manager{defaultProvider: ...}` directly holds up since the method only touches that field. @vhvb1989's StringSlice-vs-mixed question is non-blocking. Approving.

jongio (approved):
Re-verified against HEAD ca72fc4. No new commits since my prior approval and no new blocking issues. The outstanding items from the automated review are minor doc-wording nits (generate/synth read `infra.provider` from azure.yaml, not a `--provider` flag) and a narrow preview-plus-multiple-layers path that bails before the attribute is recorded. Both are non-blocking and worth a quick follow-up pass. Approving.

cli/azd/pkg/infra/provisioning/manager.go
hemarina: Nice catch. Updated to use slice of actual providers.

cli/azd/cmd/up.go
```diff
@@ -142,6 +142,11 @@ func (u *upAction) Run(ctx context.Context) (*actions.ActionResult, error) {
 	}
 	defer func() { _ = infra.Cleanup() }()
 
+	// Record the resolved IaC provider (single, or "mixed") on the cmd.up span up front — before
+	// provider init and the custom-workflow branch — so every up path (including custom workflows
+	// and early init failures) carries it.
+	u.provisioningManager.RecordInfraProviderUsage(infra.Options.GetLayers())
```
JeffreyCA: Custom `workflows.up` steps execute nested commands in-process, and each child command attaches the current usage attributes when its telemetry span ends. As a result, `package` and `deploy` steps receive `infra.provider`, and a nested `provision <layer>` can replace the value before the parent `up` span closes (for example, `["bicep","terraform"]` becomes `["bicep"]`).

Consider scoping the value to the `cmd.up` and provisioning spans so child commands report only their own attributes.

docs/specs/metrics-audit/feature-telemetry-matrix.md
```diff
@@ -65,19 +65,19 @@ These commands emit attributes or events beyond the global middleware span.
...
 | `package` | — | ✅ | ✅ | ✅ | Via hooks middleware; container service targets emit `container.credentials`, `container.publish`, `container.remotebuild` events |
 | `deploy` | — | ✅ | ✅ | ✅ | App Service zip-deploy emits `deploy.appservice.zip` (`deploy.appservice.linux`, `deploy.appservice.attempt`); container service targets emit `container.*` events |
 | `publish` | — | ✅ | ✅ | ✅ | Same as `deploy` (alias behavior) |
 | `up` | — | ✅ | ✅ | ✅ | Composes provision+deploy and inherits all their events; the up-graph runner emits `perf.provision_duration_ms`, `perf.deploy_duration_ms`, `perf.total_duration_ms` (`internal/cmd/up_graph.go`) |
-| `down` | — | ✅ | ❌ | ❌ | Teardown flow; pre/postdown lifecycle hooks emit `hooks.exec` via the hooks middleware |
+| `down` | — | ✅ | ❌ | ❌ | Teardown flow; pre/postdown lifecycle hooks emit `hooks.exec` via the hooks middleware; sets `infra.provider` (resolved IaC provider) as a usage attribute |
```
JeffreyCA: This row now documents that `down` sets `infra.provider`, but **Command-Specific Attrs** remains marked ❌. The implementation emits this attribute, and `down` is classified as command-specific in the telemetry coverage test. Consider changing that cell to ✅

JeffreyCA (approved):

cli/azd/cmd/down.go
```diff
@@ -133,6 +133,10 @@ func (a *downAction) Run(ctx context.Context) (*actions.ActionResult, error) {
 	}
 	slices.Reverse(layers)
 
+	// Record the resolved IaC provider (single, or "mixed") on the command span up front, so it is
```
jongio: This comment still references the old `"mixed"` marker, but this change replaced it with a de-duplicated slice of each provider, so the parenthetical no longer matches the behavior.

```suggestion
	// Record the resolved IaC provider(s) on the command span up front, so it is
```

cli/azd/cmd/up.go
```diff
@@ -142,6 +142,11 @@ func (u *upAction) Run(ctx context.Context) (*actions.ActionResult, error) {
 	}
 	defer func() { _ = infra.Cleanup() }()
 
+	// Record the resolved IaC provider (single, or "mixed") on the cmd.up span up front — before
```
jongio: Same stale `"mixed"` reference here. This PR dropped the mixed bucket in favor of recording each distinct provider, so `(single, or "mixed")` now describes behavior that no longer exists. Reword to `resolved IaC provider(s)`.

jongio (commented):
Two call-site comments still describe the removed "mixed" marker. Everything else in this update (the slice-based value, tests, and docs) lines up.

richardpark-msft (commented):
Makes sense, had a few comments that might help with clarity, but nothing big.

cli/azd/pkg/infra/provisioning/manager.go
```diff
@@ -538,3 +542,90 @@ func (m *Manager) newProvider(ctx context.Context) (Provider, error) {
...
+	seen := map[ProviderKind]struct{}{}
+	for _, layer := range layers {
+		kind := layer.Provider
+		if kind == NotSpecified {
+			if m.defaultProvider == nil || defaultFailed {
+				continue
+			}
+
+			if !defaultResolved {
+				resolved, err := m.defaultProvider()
```
richardpark-msft: Is there any reason to only cache this here? Maybe we could just wrap defaultProvider in another function that stores the cached value (could even use a sync.Once/sync.OnceValues (https://pkg.go.dev/sync#OnceValues) if you can't guarantee single-threaded execution.

cli/azd/pkg/infra/provisioning/manager.go
```diff
@@ -538,3 +542,90 @@ func (m *Manager) newProvider(ctx context.Context) (Provider, error) {
...
+	if len(seen) == 0 {
+		return
+	}
+
+	// Map the distinct resolved kinds to telemetry-safe values (built-ins verbatim, extensions
+	// bucketed to InfraProviderCustom), then de-duplicate and sort so the recorded slice is
+	// deterministic. De-duplicating after bucketing means two different extension providers both
+	// collapse to a single "custom" entry.
+	values := map[string]struct{}{}
+	for kind := range seen {
```
richardpark-msft: I think you can just `infraProviderTelemetryValue(kind)` when you're doing your original iteration for 'seen', correct? Then you don't have to create two maps, you only have to iterate once, etc..

jongio (approved):
Incremental changes address my earlier comments on the down.go and up.go span comments, and the up.go call placement now carries a clear rationale for scoping to the unified-graph path. Looks good.

jongio (approved):
The new commit folds default-provider resolution into `sync.OnceValues` and buckets each kind to its telemetry value in a single pass, dropping the second map. The resolver still runs at most once and the output set is unchanged.

cli/azd/cmd/up.go
```diff
@@ -187,6 +187,15 @@ func (u *upAction) Run(ctx context.Context) (*actions.ActionResult, error) {
...
 	layers := infra.Options.GetLayers()
+
+	// Record the resolved IaC provider(s) as a usage attribute on the cmd.up span. This is scoped
+	// to the unified-graph path rather than set before the custom-workflow branch on purpose:
+	// custom `workflows.up` runs `azd package`, `azd provision`, and `azd deploy` as in-process
+	// child commands that each attach the ambient usage attributes to their own spans, so recording
+	// it earlier would leak infra.provider onto the package/deploy spans and let a nested
+	// single-layer `provision` overwrite the aggregated value. Custom workflows instead carry
+	// infra.provider on their own child `azd provision` span.
+	u.provisioningManager.RecordInfraProviderUsage(layers)
```
RickWinter: This still relies on process-global usage state for a command-local field. In a custom `workflows.up`, the nested `provision` step sets `infra.provider`, and `TelemetryMiddleware` copies that same bag onto later in-process child spans and the parent when they end. A workflow that runs `deploy` after `provision` therefore tags `cmd.deploy` with the provider, and a targeted provision can leave `cmd.up` with one layer instead of the documented aggregate. Let's attach this directly to the `cmd.up` and provision spans, or save and restore the value around each nested command, then add a custom-workflow trace test.

docs/reference/telemetry-data.md
```diff
@@ -345,7 +345,7 @@ Set **only when an external command-line tool invocation fails**, during error c
 
 | Field Key | Type | Description |
 |-----------|------|-------------|
-| `infra.provider` | string | `bicep`, `terraform`, `auto` |
+| `infra.provider` | string or string[] | provision/up/down: sorted, de-duplicated string slice of resolved providers — `bicep`/`terraform`/`arm`/`pulumi`, or `custom` (extension providers; raw name not emitted); multi-layer projects record each distinct value (e.g. `["bicep","terraform"]`). generate/synth: the value read from azure.yaml's `infra.provider` as a single string (`bicep`/`terraform`, `auto` when unset, or a raw custom name) |
```
RickWinter: This changes an existing raw field from scalar-only to `string | string[]`, but the mandatory downstream review remains unchecked in the PR description. Kusto functions or cooked-table schemas that cast this value as a string can stop parsing or silently lose `provision`, `up`, and `down` rows. Complete and record the required Kusto, cooked-table, LENS, dashboard, alert, and schema checks before merging, or keep a scalar representation.

cli/azd/cmd/up.go
jongio: Confirmed this leak is real. `RecordInfraProviderUsage` writes to the process-global usage bag via `tracing.SetUsageAttributes`, and that bag is never cleared between sub-commands (it's only initialized once in `attributes.go`'s `init()`). So in a custom `workflows.up`, once the nested `azd provision` records `infra.provider`, the `TelemetryMiddleware` defer at `telemetry.go:107` (`span.SetAttributes(tracing.GetUsageAttributes()...)`) copies it onto every span that ends afterward, including the `azd deploy` sub-command span and the parent `cmd.up`. A targeted single-layer provision then lands its one value on `cmd.up` instead of the documented aggregate.

It's also inconsistent with the `usageAttributesExcluding` guard this PR adds in `up_graph.go`: that keeps `infra.provider` off the synthetic package span in the unified-graph path, but there's no equivalent protection for the custom-workflow deploy sub-command. Save/restore around each nested command, or setting `infra.provider` as a direct span attribute instead of a global usage attribute, plus a custom-workflow trace test, would close both gaps.

vhvb1989 (approved):
Looks good from my side — both of my earlier concerns (the duplicated `DefaultProviderResolver` constructor dependency, now a method on `*Manager`; and `"mixed"` → the de-duplicated provider slice) are well addressed.

Please just follow up with @RickWinter on his two open threads before merging: the process-global usage-bag leak in custom `workflows.up` (up.go:198) and the scalar→`string[]` downstream Kusto/schema review (telemetry-data.md:348). Approving on my end.

cli/azd/cmd/up.go
hemarina: Fixed. `RecordInfraProviderUsage` now attaches `infra.provider` directly to the command span via `tracing.SetAttributesInContext(ctx, ...)` instead of the process-global usage bag, so it can no longer leak onto sibling in-process child spans. On `up` it is recorded on `cmd.up` before `Initialize` (so initialization failures and the custom-workflow branch are both covered); the synthetic `cmd.provision` span in the unified graph gets it directly too, and I removed the now-unnecessary `usageAttributesExcluding` filter. Added `TestRecordInfraProviderUsage_DoesNotLeakToSiblingSpans`, which reproduces the custom-workflow provision-then-deploy scenario and asserts the value stays on the provision span, out of the global usage bag, and off the deploy span.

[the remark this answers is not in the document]
hemarina: Fixed. `RecordInfraProviderUsage` now attaches `infra.provider` directly to the command span via `tracing.SetAttributesInContext(ctx, ...)` instead of the process-global usage bag, so it can no longer leak onto sibling in-process child spans. On `up` it is recorded on `cmd.up` before `Initialize` (so initialization failures and the custom-workflow branch are both covered); the synthetic `cmd.provision` span in the unified graph gets it directly too, and I removed the now-unnecessary `usageAttributesExcluding` filter. Added `TestRecordInfraProviderUsage_DoesNotLeakToSiblingSpans`, which reproduces the custom-workflow provision-then-deploy scenario and asserts the value stays on the provision span, out of the global usage bag, and off the deploy span.

cli/azd/pkg/infra/provisioning/manager.go
hemarina: Done — wrapped the resolver in `sync.OnceValues(m.defaultProvider)` and dropped the manual cache flags, so it resolves at most once per call and is concurrency-safe.

cli/azd/pkg/infra/provisioning/manager.go
hemarina: Done — collapsed into a single pass that buckets each kind via `infraProviderTelemetryValue` into one map; no second iteration or intermediate map.

cli/azd/pkg/infra/provisioning/manager.go
```diff
@@ -538,3 +543,79 @@ func (m *Manager) newProvider(ctx context.Context) (Provider, error) {
...
+		}
+
+		providers[infraProviderTelemetryValue(kind)] = struct{}{}
+	}
+
+	if len(providers) == 0 {
+		return
+	}
+
+	tracing.SetUsageAttributes(fields.InfraProviderKey.StringSlice(slices.Sorted(maps.Keys(providers))))
```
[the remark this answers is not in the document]
hemarina: The polymorphic slice is intentional. I have updated the PR description and all three telemetry docs to document `infra.provider` as multi-valued (`string[]`) on provision/up/down and scalar on generate/synth, including the two-distinct-custom-providers -> `["custom"]` behavior.

docs/reference/telemetry-data.md
hemarina: I will update the downstream Kusto queries and schemas to handle the array shape. Since this is not consumed until the next release, there is no immediate breakage and it can proceed in parallel.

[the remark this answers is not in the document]
hemarina: I will update the downstream Kusto queries and schemas to handle the array shape. Since this is not consumed until the next release, there is no immediate breakage and it can proceed in parallel.

jongio (commented):
The switch from the process-global usage bag to span-scoped recording closes the custom-workflow leak I flagged, and the new DoesNotLeakToSiblingSpans test locks it in. Moving the layers computation ahead of Initialize is safe since GetWithDefaults builds a fresh copy and doesn't mutate infra.Options. One nit inline.

cli/azd/pkg/infra/provisioning/manager_test.go
```diff
@@ -395,3 +400,199 @@ func registerContainerDependencies(mockContext *mocks.MockContext, env *environm
...
+				return true
+			}
+		}
+		return false
+	}
+
+	require.True(t, hasInfraProvider(byName["cmd.provision"]), "provision span should carry infra.provider")
+	require.False(t, hasInfraProvider(byName["cmd.deploy"]),
+		"infra.provider must not leak onto the sibling deploy span")
+	require.False(t, hasInfraProvider(tracing.GetUsageAttributes()),
```
jongio, line 648: This file doesn't end with a newline (the closing brace on line 598 has no trailing newline), so gofmt and the lint job will flag it. Add a trailing newline at the end of the file.

jongio: Still there on the latest push (24177320). golangci-lint is red on `manager_test.go:598:1: File is not properly formatted (gofmt)` for exactly this - the file still has no trailing newline after the closing brace. Adding one clears the only failing check; the rest of the earlier feedback looks addressed.

jongio (approved):
Re-reviewed the latest push. The two things I flagged earlier are resolved: infra.provider is now attached span-scoped through SetAttributesInContext instead of the process-global usage bag, so it no longer leaks onto sibling child-command spans, and manager_test.go now ends with a trailing newline so gofmt/golangci-lint are green. resolveInfraProviders resolving the default once via sync.OnceValues and bucketing extension providers to `custom` both look right. Local build and gofmt pass on the changed Go files.

cli/azd/cmd/infra_generate.go
```diff
@@ -89,12 +89,16 @@ func newInfraGenerateAction(
...
+	// this cmd.generate/cmd.synth span and cannot leak onto — or overwrite — the span-scoped
+	// infra.provider recorded by a sibling in-process command (for example a custom `workflows.up`
+	// step running provision/up alongside generate/synth).
 	if a.projectConfig != nil {
 		provider := string(a.projectConfig.Infra.Provider)
 		if provider == "" {
 			provider = "auto"
 		}
-		tracing.SetUsageAttributes(fields.InfraProviderKey.String(provider))
+		tracing.SetAttributesInContext(ctx, fields.InfraProviderKey.String(provider))
```
jongio, line 106: `provider` here is `string(a.projectConfig.Infra.Provider)` read straight from azure.yaml, and an extension provider name is unconstrained, so a non-built-in provider is emitted raw under `infra.provider`. The PR body, `telemetry-schema.md`, and the new privacy-review checklist item ("extension provider names remain bucketed to custom") all state raw names are never emitted, but this path never buckets. Either run this value through the same built-in allowlist plus `custom` bucketing that `resolveInfraProviders` applies on provision/up/down, or scope the "never raw" guarantee in the description/docs/checklist to provision/up/down and state that generate/synth emit the configured value as-is. This one has stayed open across the last few pushes.

jongio (commented):
The structural feedback from earlier rounds looks addressed on this commit. One privacy item is still open: `infra generate` / `synth` emit the configured provider value without the `custom` bucketing that the PR body, docs, and the new privacy-review checklist say applies to all providers. Inline comment has the specifics.

jongio (approved):
Verified against 50a415b. The prior concerns are resolved:

- The process-global usage-bag leak is fixed. `RecordInfraProviderUsage` now attaches `infra.provider` directly to the command span via `SetAttributesInContext`, and `TestRecordInfraProviderUsage_DoesNotLeakToSiblingSpans` locks that behavior in.
- `infra generate` / `synth` bucket non-built-in providers to `custom` (no raw name emitted), with span-level tests covering it.
- The `manager_test.go` trailing-newline gofmt failure is cleared; `gofmt -l` is clean.
- Docs now describe the scalar-vs-array shape and the canonical `cmd.infra.generate` span, and the sort contract is exercised with reverse-ordered input plus pulumi cases.

gofmt, go vet, and the affected unit tests all pass locally.

jongio (approved):
Re-reviewed the latest push (50a415b). The process-global usage-bag leak I confirmed earlier is resolved: `RecordInfraProviderUsage` now attaches `infra.provider` directly to the command span via `SetAttributesInContext`, so it no longer bleeds onto sibling in-process child spans (custom `workflows.up` running provision then deploy stays clean).

`infra generate`/`synth` now bucket non-built-in providers through `InfraProviderTelemetryValue`, and the new `Test_InfraGenerateAction_RecordsInfraProvider` covers unset->auto, built-in verbatim, and extension->custom. The slice sorting is now exercised with reverse-ordered input, the trailing-newline lint is fixed, and CI is green.

Remaining items are minor and non-blocking: the new `down` emission has no trace-level test (functional coverage only asserts `cmd.up`/`cmd.provision`), and the telemetry-data.md scope list omits the `infra create`/`infra delete` wrappers, which delegate to `ProvisionAction.Run`/`downAction.Run` and therefore also carry this attribute. Worth a quick follow-up but nothing blocking.

JeffreyCA (approved):

PR #9112, opened 2026-07-13
feat(agents): provision terraform project connections

### LINKED ISSUE #9111: [azure.ai.agents] Terraform eject does not provision azure.ai.connection services
## Problem

`azd ai agent init --infra=terraform` does not include `azure.ai.connection` services from `azure.yaml` in the generated Terraform infrastructure.

As a result, a project using Terraform eject provisions the Foundry account and project but omits declared Foundry project connections.

## Repro

1. Create an `azure.yaml` with an `azure.ai.project` service and an `azure.ai.connection` service.
2. Run:

   ```powershell
   azd ai agent init --infra=terraform
   azd provision --no-prompt
3. Inspect generated infra/ files and the resulting Foundry project.

## Expected
- Terraform eject generates resources for declared project connections.
- azd provision creates those connections under the Foundry project.
- Connection category, target, authentication type, credentials, and metadata
are preserved.
- Provision outputs expose the created connection names.

## Actual
- No Terraform resource is generated for azure.ai.connection services.
- The Foundry project is provisioned without its declared connection

## Related 
https://github.com/Azure/azure-dev/issues/8350
https://github.com/Azure/azure-dev/pull/9046

hund030:
## Summary

Fixes #9111.

Terraform infrastructure eject currently omits `azure.ai.connection` services declared in `azure.yaml`. This prevents Terraform-based Foundry projects from provisioning project connections alongside the account and project.

## Scope

- Synthesize `azure.ai.connection` services into Terraform variables.
- Generate one `azapi_resource` per connection under the Foundry project.
- Preserve connection category, target, auth type, credentials, and metadata.
- Keep credential values out of Terraform plan output.
- Export provisioned connection names through `AZURE_AI_PROJECT_CONNECTION_NAMES`.
- Cover generated Terraform files and nested credential shapes with unit tests.

## Acceptance Criteria

- An `azure.ai.connection` service is emitted to `infra/connections.tf`.
- `azd provision` creates the Foundry project connection.
- Projects without connections continue to provision successfully.
- Terraform outputs include the comma-separated connection names.
- Credential values are not rendered in Terraform plans.

## Validation

- `go test ./internal/cmd -run TestEjectInfra_Terraform_EjectsConnectionServices -count=1`

### DIFF
--- a/cli/azd/extensions/azure.ai.agents/CHANGELOG.md
+++ b/cli/azd/extensions/azure.ai.agents/CHANGELOG.md
@@ -4,6 +4,9 @@
 
 ### Other Changes
 
+- [[#9112]](https://github.com/Azure/azure-dev/pull/9112) Terraform infrastructure eject now synthesizes
+  `azure.ai.connection` services into Foundry project connection resources, preserving their category, target,
+  authentication type, credentials, and metadata.
 - [[#9049]](https://github.com/Azure/azure-dev/pull/9049) Switch the `invocations_ws` agent endpoint from the preview dispatcher form to the GA path-based route. `azd deploy` now registers `AGENT_{KEY}_INVOCATIONS_WS_ENDPOINT` (and `azd ai agent show` displays `Endpoint (invocations_ws)`) as `wss://<account>.services.ai.azure.com/api/projects/<project>/agents/<agent>/endpoint/protocols/invocations_ws?api-version=v1`, carrying the project and agent as path segments to mirror the HTTP `invocations` route. The previous form embedded them as `project_name`/`agent_name` query parameters on a single literal `/api/projects/agents/...` path.
 
 ## 1.0.0-beta.5 (2026-07-09)
--- a/cli/azd/extensions/azure.ai.agents/internal/cmd/init_infra.go
+++ b/cli/azd/extensions/azure.ai.agents/internal/cmd/init_infra.go
@@ -587,10 +587,10 @@ func writeOutputsFile(infraDir string, includeAcr bool) (ejectArtifact, error) {
 
 // writeTfvarsFile emits infra/main.tfvars.json. azd-core's Terraform provider
 // reads this file and substitutes the ${...} placeholders from the azd
-// environment at provision time. The synthesizer-known value `deployments` is
-// written literally; deploy-time inputs (location, resource_group_name,
-// foundry_project_name, principal_id, subscription_id, environment_name,
-// resource_token_salt) are left as ${AZURE_*} placeholders.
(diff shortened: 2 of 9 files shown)


jongio (commented):
A couple of notes on top of the existing comments.

**No CHANGELOG entry.** This extension logs every feature/fix under `## 1.0.0-beta.6 (Unreleased)` with the PR number, but this PR doesn't add one. The sibling provision-time PR (#9046) is also absent from the changelog, so if the plan is to track the whole connections feature under a single entry later, ignore this. Otherwise, add a line noting that Terraform eject now synthesizes `azure.ai.connection` services.

**On the already-flagged `map(any)` credentials comment** - that's the one I'd prioritize. `connections.bicep` types credentials as `object?`, and the synthesized Go value is `map[string]any` that can hold nested objects (CustomKeys' `keys:`). Terraform's `map(any)` unifies to a single element type, so a project mixing an ApiKey connection (`{key: "..."}`) with a CustomKeys connection (`{keys: {...}}`) can fail type conversion or coerce the shape. That undercuts the "supports every auth type" intent. `optional(any)` keeps parity with the Bicep `object?` contract.

jongio (approved):
My earlier concerns are resolved. Credentials now type as `optional(any)`, so a project mixing an ApiKey connection with a CustomKeys connection round-trips its shape instead of hitting a `map(any)` type-unification error, and moving them into `sensitive_body` keeps the values out of `terraform plan` output. The switch from raw `envsubst` over the whole file to per-leaf `foundry.ExpandEnv` is the right call and the new tests cover the quote/newline/`${{...}}` cases that motivated it.

One thing worth a conscious call before this ships: the credential handling now depends on the reworked core substitution path, but the extension still sets `requiredAzdVersion: ">=1.27.0"`, and those releases predate this change. Bumping it to the azd version that actually carries this substitution rework avoids an install on an older host silently falling back to the raw path for credentials containing quotes or newlines.

jongio (approved):
The revert drops the core Terraform provider substitution rework out of this PR, which clears the version-coupling concern from my last pass: the extension no longer depends on a substitution path that shipped after `requiredAzdVersion: ">=1.27.0"`. Credentials now ride the same `${VAR}` envsubst path azd already uses for `deployments` and the other tfvars values, and the `optional(any)` typing plus `sensitive_body` handling from the earlier commits are intact. Tests cover the mixed ApiKey/CustomKeys shapes and the CHANGELOG entry is in. Re-approving to clear the stale state from the new push.

trangevi (approved):
Approved, pending final copilot comment

PR #9115, opened 2026-07-13
test(ext-agents): guard tier selectors against false-green

### LINKED ISSUE #8778: ext-azure-ai-agents CI: guard against -run regex false-green
## Problem

The \xt-azure-ai-agents-ci\ workflow uses \go test -run <regex>\ to select tests by tier. If a test is renamed or added without updating the workflow regex, \go test\ matches zero tests and exits with code 0 — a silent false-green.

## Current state

All 9 tests are correctly covered by the current regex patterns (verified). No tests are currently missed.

## Suggested fix

Add a post-test check that verifies the expected number of tests ran, e.g.:

\\\ash
go test ... -v 2>&1 | tee out.txt
grep -c '^--- PASS' out.txt | grep -q '^6\$'  # Tier 0: expect 6
\\\

Or use build tags / environment variables for tier selection instead of \-run\ regex.

## Context

Identified during review of #8754.

glharper:
## Summary

The `ext-azure-ai-agents-test` workflow selects Tier 0 and Tier 1 functional tests with `go test -run <regex>`. Go exits successfully when a regex matches no tests, so renaming or adding a `Test_AIAgent_` test without updating the tier selectors could silently produce a false-green CI run.

## Fix

- Centralize the Tier 0 and Tier 1 regexes as job-level environment variables so validation and execution use the same selectors.
- Anchor both regexes to exact top-level test names.
- Add a `Validate tier test selectors` step that uses `go test -list` to discover tests and fails when:
  - no `Test_AIAgent_` tests are discovered;
  - either tier matches zero tests;
  - a test is assigned to both tiers; or
  - any discovered test is not assigned to exactly one tier.
- Normalize carriage returns so the check is deterministic when reproduced from Windows Bash as well as on the Ubuntu runner.

The current inventory is validated as 4 Tier 0 tests and 5 Tier 1 tests. The actual test steps reuse the validated regex variables.

## Validation

- Ran the exact selector guard locally: it discovered all 9 tests and accepted the current 4/5 split.
- Removed `Test_AIAgent_Doctor_Help` from the Tier 0 selector in a local negative-control run; the guard failed and identified the uncovered test.
- `git diff --check` passes.

Fixes #8778

### DIFF
--- a/.github/workflows/test-ext-azure-ai-agents.yml
+++ b/.github/workflows/test-ext-azure-ai-agents.yml
@@ -20,6 +20,9 @@ permissions:
 jobs:
   test:
     runs-on: ubuntu-latest
+    env:
+      TIER_0_TEST_PATTERN: '^Test_AIAgent_(Version|Help|Init_NoPrompt_MissingFlags|Doctor_Help)$'
+      TIER_1_TEST_PATTERN: '^Test_AIAgent_(Init_(NoPrompt_(Defer|WithProject)|NegativeControl_BadCassette)|SampleList_(Recorded|JSON_Recorded))$'
     steps:
       - uses: actions/checkout@v6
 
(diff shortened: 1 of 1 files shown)


.github/workflows/test-ext-azure-ai-agents.yml
```diff
@@ -73,12 +76,75 @@ jobs:
           azd x publish
           azd extension install azure.ai.agents --source local --force --no-prompt
 
+      - name: Validate tier test selectors
```
jongio, line 33: This guard only needs Go and the test sources to compile. `go test -list` against `./test/functional/ai_agents/` runs fine with just checkout + setup-go (the package only imports `test/azdcli` and `test/recording`, no built extension is needed). As placed, it runs after the inspector and agents build/publish steps, so a bad selector won't surface until several minutes of extension builds have already run. Moving this step up to right after `setup-go` would fail fast on selector drift before the expensive build work. Non-blocking.

jongio (commented):
One non-blocking suggestion on the new validation step, inline. The guard logic itself checks out: I ran it against the current package and the two tier regexes select exactly the 4 Tier 0 and 5 Tier 1 tests with full coverage and no overlap.

.github/workflows/test-ext-azure-ai-agents.yml
glharper: Addressed in a1f99f374. The selector validation now runs immediately after actions/setup-go, before building azd, the inspector dependency, or the agents extension, so selector drift fails fast. Re-ran the guard locally: it still validates the current 4 Tier 0 / 5 Tier 1 split.

jongio (approved):

huimiu (approved):

vhvb1989 (approved):

PR #9116, opened 2026-07-13
feat(skills): own azure.ai.skill service blocks in azure.yaml

### LINKED ISSUE #9088: azure.ai.skills: own the azure.ai.skill service block in azure.yaml
## Summary

Give `azure.ai.skills` ownership of its `host: azure.ai.skill` service block in `azure.yaml` - authoring, reading, and reconciling it - so a skill can be composed into an application like the other Foundry resources.

## Background

Each Foundry resource should map to one `azure.yaml` service entry, keyed by the resource name, with `host: azure.ai.<kind>`, authored and read by the extension that owns that kind. `azure.ai.agents` authors the sibling blocks it knows about (project, connection, toolbox), which centralizes that knowledge and blocks composable adoption. This work item covers the `azure.ai.skill` host.

## Current state

- `azure.ai.skills` registers a service-target for `host: azure.ai.skill` (reconcile on `azd up`/`provision`/`deploy`) but nothing authors or reads the block.
- Unlike project/connection/toolbox, `azure.ai.skill` is not written by any extension: it is absent from `azure.ai.agents`' `resource_services.go` and from the `foundryServiceHosts` adopt scan. Authoring is net-new and belongs here.

## Proposed

- `azure.ai.skills` authors and updates its own block through the azd extension framework (`AddService`, `SetServiceConfigSection`/`SetServiceConfigValue`), reads it (`Get`, `GetServiceConfigSection`), and continues to reconcile it through its existing service-target. Authoring is driven from the extension's own commands (for example `skill create`/`update`) when the developer opts into `azure.yaml`.
- The agent declares the dependency with `uses:` when it consumes the skill.

The block carries: inline instructions or an archive reference.

```yaml
services:
  triage-rules:
    host: azure.ai.skill
    # inline instructions or archive reference
```

## Acceptance criteria

- [ ] `azure.ai.skills` authors its `host: azure.ai.skill` block (create + idempotent update) through the azd extension framework, driven from its own commands.
- [ ] `azure.ai.skills` reads its own block for reconcile on `azd up`/`provision`/`deploy`.
- [ ] A developer can opt a skill into `azure.yaml` from `azure.ai.skills` commands, with no dependency on the agents extension.
- [ ] The agent declares the dependency via `uses:` when it consumes the skill.
- [ ] The agents adopt/scan recognizes `azure.ai.skill` generically (add it to the recognized foundry hosts) so a composed manifest round-trips.

## References

- Reconcile registration (stays): `cli/azd/extensions/azure.ai.skills/internal/cmd/root.go` (`host.WithServiceTarget(aiSkillHost, ...)`).
- Where the pattern to mirror lives (project/connection/toolbox authoring): `cli/azd/extensions/azure.ai.agents/internal/cmd/resource_services.go` (`emitResourceServices`, `collect*`).
- Adopt scan to extend with the skill host: `cli/azd/extensions/azure.ai.agents/internal/cmd/init_adopt.go` (`foundryServiceHosts`, which today omits `azure.ai.skill`).
- azd extension framework: `cli/azd/pkg/azdext/project_grpc.pb.go` (`Get`, `AddService`, `GetServiceConfigSection`, `SetServiceConfigSection`, `SetServiceConfigValue`).

## Part of

Per-extension `azure.yaml` ownership: each AI extension owns its host kind in `azure.yaml` instead of `azure.ai.agents` authoring all of them. Sibling work items: `azure.ai.projects`, `azure.ai.connections`, `azure.ai.toolboxes`, `azure.ai.routines`.

glharper:
## Summary

Gives the `azure.ai.skills` extension ownership of its `host: azure.ai.skill` service blocks in `azure.yaml`: command-driven authoring, idempotent updates, reading, and deploy-time reconciliation.

Developers can now opt a skill into the current azd project directly from the owning extension:

```console
azd ai skill create triage-rules --file ./skills/triage-rules --save-to-azure-yaml
azd ai skill update triage-rules --file ./SKILL.md --save-to-azure-yaml
```

This has no dependency on the agents extension. An agent that consumes the skill declares the dependency using the existing `uses:` field.

Fixes #9088

### `azure.ai.skills` owns authoring and updates

- Adds `--save-to-azure-yaml` to `skill create` and `skill update` (opt-in; existing remote-only behavior remains the default).
- New manifest helper uses the azd extension framework APIs:
  - `Project.Get` to locate/read the current project and existing service.
  - `AddService` to create a new `host: azure.ai.skill` block.
  - `GetServiceConfigSection` + read/merge/`SetServiceConfigSection` to update an existing block while preserving `uses:`, `project`, and unrelated/unknown fields.
- Detects a same-name service with a different host and returns a structured, actionable conflict instead of overwriting it.
- Requires ZIP/directory references to live inside the azd project and writes a portable forward-slash relative path.
- `--set-default-version` is rejected with `--save-to-azure-yaml` because a default-version pointer is not a declarative content source that `azd deploy` can reconcile.

### Inline and archive declarations

Inline flags and parsed `SKILL.md` inputs are saved as service-level content:

```yaml
services:
  triage-rules:
    host: azure.ai.skill
    description: Rules for triaging incoming issues
    instructions: Classify the issue and recommend next steps.
    tools:
      - web_search
```

ZIP and directory inputs are saved as an archive reference:

```yaml
services:
  triage-rules:
    host: azure.ai.skill
    archive: skills/triage-rules
```

The existing skill service target now reads and reconciles both shapes. Archive reconciliation:

- Accepts a `.zip` file or a directory containing root-level `SKILL.md`.
- Reuses `ArchiveDirectory` safety limits for directories (no symlinks/non-regular entries; entry/size limits).
- Rejects parent traversal and unsupported archive files.
- Creates a new immutable default skill version, matching existing inline reconcile behavior.

The runtime validation keeps the pre-split legacy `config:` fallback compatible while enforcing that archive and inline fields are not combined.

### Composition/adoption

- Adds `azure.ai.skill` to the agents extension's generic Foundry host recognition, so a skill-only or composed unified `azure.yaml` round-trips through the adoption flow.
- Documents adding the skill service name to a consuming agent's `uses:` list.

## Tests

- Add vs idempotent update, including preservation of `uses:` and unknown fields.
- Host-name collision and missing-project structured errors.
- Project-relative archive path conversion and outside-project rejection.
- Inline/archive config parsing, traversal rejection, ZIP/directory preparation, unsupported file rejection.
- `--save-to-azure-yaml` flag registration and `--set-default-version` conflict.
- Regression coverage for multi-line inline instructions ending in `.md`/`.txt` (must not be misclassified as a file path).
- Agents adoption recognition for a skill-only manifest.

Validation run:

- `go build ./...` and `go test ./...` in `azure.ai.skills`
- `golangci-lint run ./...` and extension cspell in `azure.ai.skills`
- agents `internal/cmd` test/lint and cspell for the adoption change
- schema JSON parse and `git diff --check`

### DIFF
--- a/cli/azd/extensions/azure.ai.skills/internal/cmd/service_target.go
+++ b/cli/azd/extensions/azure.ai.skills/internal/cmd/service_target.go
@@ -4,12 +4,16 @@
 package cmd
 
 import (
+	"bytes"
 	"context"
 	"encoding/json"
 	"fmt"
+	"io"
+	"os"
 	"path/filepath"
 	"strings"
 
+	"azureaiskills/internal/exterrors"
 	"azureaiskills/internal/pkg/skill_api"
 
 	"github.com/azure/azure-dev/cli/azd/pkg/azdext"
@@ -29,12 +33,13 @@ type skillServiceConfig struct {
 	Description  string   `json:"description,omitempty"`
 	Instructions string   `json:"instructions,omitempty"`
 	Tools        []string `json:"tools,omitempty"`
+	Archive      string   `json:"archive,omitempty"`
 }
 
 // skillServiceTarget upserts a Foundry skill declared as an azure.ai.skill
-// service. Deploy creates a new default skill version from the entry's inline
-// instructions; the resource name is the service key. Package and Publish are
-// no-ops because a skill has no build artifact.
+// service. Deploy creates a new default skill version from either inline
+// instructions or an archive reference; the resource name is the service key.
+// Package and Publish are no-ops because a skill has no build artifact.
 type skillServiceTarget struct {
 	azdClient     *azdext.AzdClient
 	serviceConfig *azdext.ServiceConfig
@@ -100,9 +105,9 @@ func (p *skillServiceTarget) Publish(
 }
 
 // Deploy upserts the skill by creating a new default version from the entry's
-// instructions. Re-running deploy creates another immutable version rather than
-// failing. Removing the service from azure.yaml stops azd managing the skill but
-// does not delete it (use `azd ai skill delete`).
+// inline instructions or archive reference. Re-running deploy creates another
+// immutable version rather than failing. Removing the service from azure.yaml
+// stops azd managing the skill but does not delete it (use `azd ai skill delete`).
 func (p *skillServiceTarget) Deploy(
 	ctx context.Context,
 	serviceConfig *azdext.ServiceConfig,
@@ -114,12 +119,23 @@ func (p *skillServiceTarget) Deploy(
 	if err != nil {
 		return nil, err
 	}
-	instructions, err := resolveSkillInstructions(serviceConfig, cfg.Instructions)
-	if err != nil {
-		return nil, err
+	hasInline := cfg.Description != "" || cfg.Instructions != "" || len(cfg.Tools) > 0
+	if cfg.Archive != "" && hasInline {
+		return nil, exterrors.Validation(
+			exterrors.CodeInvalidParameter,
+			fmt.Sprintf(
+				"skill service %q cannot combine archive with description, instructions, or tools",
+				serviceConfig.GetName(),
+			),
+			"configure either archive, or inline description/instructions/tools, but not both",
+		)
 	}
-	if instructions == "" {
-		return nil, fmt.Errorf("skill service %q requires instructions", serviceConfig.GetName())
+	if cfg.Archive == "" && cfg.Instructions == "" {
+		return nil, exterrors.Validation(
+			exterrors.CodeMissingRequiredField,
+			fmt.Sprintf("skill service %q requires instructions or archive", serviceConfig.GetName()),
+			"set instructions to inline text/a .md or .txt path, or set archive to a .zip/directory path",
+		)
 	}
 
 	if progress != nil {
@@ -131,19 +147,45 @@ func (p *skillServiceTarget) Deploy(
 		return nil, err
 	}
 
-	if _, err := skillCtx.client.CreateVersionInline(
-		ctx,
-		serviceConfig.GetName(),
-		skill_api.CreateVersionRequest{
-			InlineContent: &skill_api.SkillInlineContent{
-				Description:  cfg.Description,
-				Instructions: instructions,
-				AllowedTools: cfg.Tools,
+	if cfg.Archive != "" {
+		archivePath, err := resolveSkillArchivePath(serviceConfig, cfg.Archive)
+		if err != nil {
+			return nil, err
+		}
+		archive, err := prepareSkillArchive(archivePath)
+		if err != nil {
+			return nil, err
+		}
+		defer archive.Reader.Close()
+
+		if _, err := skillCtx.client.CreateVersionFromZip(
+			ctx,
+			serviceConfig.GetName(),
+			archive.Name,
+			archive.Reader,
+			true,
+		); err != nil {
+			return nil, exterrors.ServiceFromAzure(err, exterrors.OpReconcileSkill)
+		}
+	} else {
+		instructions, err := resolveSkillInstructions(serviceConfig, cfg.Instructions)
+		if err != nil {
+			return nil, err
+		}
+		if _, err := skillCtx.client.CreateVersionInline(
+			ctx,
+			serviceConfig.GetName(),
+			skill_api.CreateVersionRequest{
+				InlineContent: &skill_api.SkillInlineContent{
+					Description:  cfg.Description,
+					Instructions: instructions,
+					AllowedTools: cfg.Tools,
+				},
+				Default: true,
 			},
-			Default: true,
-		},
-	); err != nil {
-		return nil, fmt.Errorf("upserting skill %q: %w", serviceConfig.GetName(), err)
+		); err != nil {
+			return nil, exterrors.ServiceFromAzure(err, exterrors.OpReconcileSkill)
+		}
 	}
 
 	return &azdext.ServiceDeployResult{}, nil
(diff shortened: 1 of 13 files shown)


jongio (approved):
Traced the archive vs inline validation, the field-preserving upsert merge (drops stale `archive` on an inline switch while keeping `uses:` and unknown fields), and the path handling: `portableSkillArchiveReference` rejects out-of-project sources and `resolveSkillArchivePath`/`hasParentTraversal` block `..` escapes. The save-time project-root-relative vs deploy-time service-relative resolution lines up with the existing `resolveSkillInstructions` convention and the schema doc. The `isInstructionFilePath` multiline guard is right and covered by a regression test. Tests cover add/update, host conflict, missing project, traversal, outside-project, zip/dir prep, and the `--set-default-version` conflict.

cli/azd/extensions/azure.ai.skills/internal/cmd/skill_create.go
```diff
@@ -538,6 +619,12 @@ skill of the same name before creating.`,
 	cmd.Flags().StringVar(&flags.file, "file", "",
 		"Path to SKILL.md (.md), a ZIP package (.zip), or a directory containing SKILL.md at its root")
 	cmd.Flags().BoolVar(&flags.force, "force", false, "Delete an existing skill of the same name before creating")
+	cmd.Flags().BoolVar(
```
trangevi, line 622: I'd like John to weigh in on a parameter like this when he is back from vacation

jongio (approved):
Incremental review of the preflight-persistence commit. Splitting the azure.yaml save into prepare + save runs validation before the Azure delete/create, so a manifest conflict no longer leaves the service mutated with azure.yaml unwritten. Force-delete relocation covers all four create modes, and anchoring archive resolution to the project path (with the traversal guard retained) fixes the nested-working-directory case. Tests, vet, and go fix are clean.

trangevi (changes requested):
I'd like to wait on the "authoring" piece of the corresponding issue until John is back from vacation; there needs to be further discussion about how we introduce those pieces. For now, if we want to make progress on this work item and the others like it, we should focus on provision and deploy surface areas, and leave any authoring pieces that may exist within the agents extension for the time being.

huimiu (commented):
If we want `azure.yaml` to be the single source of truth, it may be better not to hide this behavior behind an optional flag. We could either make updating `azure.yaml` the default, or separate it into a dedicated command with clearer intent.`n`nWe should also think through failure and consistency scenarios: what happens if the remote Skill is updated successfully but writing `azure.yaml` fails, or if the manifest changes after the remote operation? We need a clear source-of-truth and recovery story so local configuration and the deployed Skill do not drift.

jongio (commented):
> @huimiu: what happens if the remote Skill is updated successfully but writing `azure.yaml` fails, or if the manifest changes after the remote operation?

That drift window is real in the current flow. Both create and update run the same order: prepare the azure.yaml upsert (read and validate only), then the remote op (`DeleteSkill` / `CreateVersion*`), then `service.Save` writes azure.yaml (the run methods in `skill_create.go` and `skill_update.go`).

Validation-type failures (host conflict, missing project, traversal, outside-project archive) are caught in prepare before any remote call. But a `Save` failure after a successful `CreateVersion` leaves the new remote version as the default with azure.yaml unwritten, and the command returns the Save error. So the drift surfaces on any post-remote Save failure, not only a concurrent manifest edit.

For the recovery story: since the remote op is the point of no return, when Save fails after it the command could report the created version so the user can re-run with `--save-to-azure-yaml` (or `--set-default-version`) to reconcile, rather than only surfacing the write error.

jongio (approved):
Re-verified the current diff at a3859312. The prepare/save split still runs manifest validation (host conflict, missing project, traversal, outside-project) before the remote delete/create, so a validation failure no longer leaves the skill mutated with azure.yaml unwritten. The upsert merge preserves `uses:` and unrelated fields while clearing stale owned fields like `archive` on an inline switch. Archive resolution stays anchored to the project path with the `..` traversal and outside-project guards intact, and the multiline `isInstructionFilePath` guard is covered by regression tests. CI is green. No new issues.

jongio (approved):
Re-verified the current diff at a3859312 end to end.

The prepare/save split validates the azure.yaml upsert (host conflict, missing project, `..` traversal, outside-project archive) before the remote delete/create, so a validation failure can't leave the skill mutated with azure.yaml unwritten. `deleteExistingSkill` runs the `--force` delete after prepare and before create across all four create modes. The upsert merge clears only the owned fields (`archive`, `description`, `instructions`, `tools`) and preserves `host`, `uses:`, and unknown fields. Archive references stay anchored to the project path on save and the service path on deploy, with the traversal and outside-project guards intact.

Tests cover add vs update, host conflict, missing project, traversal, outside-project, zip/dir prep, and the `--set-default-version` conflict. CI is green.

No code-level blockers from me. The one open item is the design decision on the `--save-to-azure-yaml` authoring surface (opt-in flag vs. default vs. a dedicated command), which is already under discussion here. That's a maintainer call rather than a code issue, and it should be settled before this merges.

jongio (approved):
Re-verified the current diff at a3859312.

The prepare/save split reads and validates the azure.yaml upsert (host conflict, missing project, `..` traversal, outside-project archive) before the remote create runs, so a validation failure can't leave the skill mutated with azure.yaml unwritten. The owned-field merge preserves `uses:` and any unknown service fields, and archive references round-trip relative to the service path. Inline and archive stay mutually exclusive at deploy, and multi-line instructions ending in `.md`/`.txt` are no longer misread as file paths. Tests cover add/update/preserve, path traversal, outside-project rejection, and host conflict.

No code concerns from me. The one open item is the design question on `--save-to-azure-yaml` that's waiting on John, which is a team call rather than a code issue.

jongio (commented):
Re-checked the current diff at a3859312. Two low severity items on the deploy path that I hadn't flagged in my earlier passes. Neither is a blocker and my approval stands.

- `service_target.go:175` - an inline skill whose instructions file resolves to empty now reaches the service instead of failing locally.
- `service_target.go:269` - the new archive resolver anchors to the project path, but `resolveSkillInstructions` still anchors to the process working directory.

cli/azd/extensions/azure.ai.skills/internal/cmd/service_target.go
```diff
@@ -131,19 +147,49 @@ func (p *skillServiceTarget) Deploy(
...
+			archive.Reader,
+			true,
+		); err != nil {
+			return nil, exterrors.ServiceFromAzure(err, exterrors.OpReconcileSkill)
+		}
+	} else {
+		instructions, err := resolveSkillInstructions(serviceConfig, cfg.Instructions)
+		if err != nil {
+			return nil, err
+		}
```
jongio, line 178: The empty-instructions check moved ahead of file resolution, and the post-resolution guard went away with it. The old flow was resolve, then `if instructions == "" { error }`. Now the only check is `cfg.Archive == "" && cfg.Instructions == ""` at line 133, and that runs against the raw azure.yaml value, not the resolved content.

So `instructions: ./SKILL.md` pointing at an empty or whitespace-only file passes validation, and `CreateVersionInline` gets an empty `Instructions`. The user ends up with whatever the service returns rather than the structured local error they used to get.

Restoring the check after resolution keeps the old behavior:

```suggestion
		instructions, err := resolveSkillInstructions(serviceConfig, cfg.Instructions)
		if err != nil {
			return nil, err
		}
		if strings.TrimSpace(instructions) == "" {
			return nil, exterrors.Validation(
				exterrors.CodeMissingRequiredField,
				fmt.Sprintf("skill service %q resolved to empty instructions", serviceConfig.GetName()),
				"add content to the instructions file, or set instructions to inline text",
			)
		}
```

cli/azd/extensions/azure.ai.skills/internal/cmd/service_target.go
```diff
@@ -199,6 +245,87 @@ func resolveSkillInstructions(svc *azdext.ServiceConfig, instructions string) (s
...
+			exterrors.CodeInvalidSkillFile,
+			fmt.Sprintf("skill archive path %q must not contain '..' or escape the service directory", archive),
+			"move the archive inside the service directory and use a relative path without '..'",
+		)
+	}
+	baseDir := svc.GetRelativePath()
+	if baseDir == "" {
+		baseDir = "."
+	}
+	return filepath.Join(projectPath, baseDir, path), nil
```
jongio, line 269: Anchoring to `projectPath` here is correct. Extensions inherit azd's invocation directory rather than the project root, since `pkg/extensions/runner.go` never sets a working directory on the child process, and `RelativePath` is a relative value like `./src/api`.

`resolveSkillInstructions` just above still does `filepath.Join(baseDir, path)` with no project path, so it resolves against the process working directory. That means `instructions: ./SKILL.md` breaks when `azd deploy` runs from a subdirectory, while `archive: ./x` works. It predates this PR, but the two resolvers now sit next to each other and disagree. Passing `projectPath` into `resolveSkillInstructions` would line them up.

jongio (approved):
Approving at a3859312. Build, vet, and the extension test suite pass locally.

Two items from my earlier pass are still open on the deploy path. Neither blocks merge:

- `service_target.go`: an inline skill whose instructions file resolves to empty content now reaches the service instead of failing locally. The old guard checked the resolved text, the new one checks the raw azure.yaml value.
- `service_target.go`: `resolveSkillArchivePath` anchors to the project path, but `resolveSkillInstructions` still joins only the service relative path, so it resolves against the extension process working directory.

jongio (commented):
One more item from a fresh pass at a3859312, on top of the two deploy-path notes I left earlier today. Not a blocker and my approval stands.

cli/azd/extensions/azure.ai.skills/internal/cmd/skill_create.go
```diff
@@ -148,6 +173,19 @@ func (a *createAction) runFileMd(ctx context.Context, client *skill_api.Client)
 		)
 	}
 
+	service, err := a.prepareService(ctx, skillServiceConfig{
+		Description:  parsed.Description,
+		Instructions: parsed.Instructions,
+		Tools:        parsed.AllowedTools,
```
jongio, line 179: `create --file X.md` sends `Metadata`, `License`, and `Compatibility` from the SKILL.md front matter to the service a few lines below, but the azure.yaml block only gets `description`, `instructions`, and `tools`. `skillServiceConfig` has no field for the other three, and `azure.ai.skill.json` has no property for them.

So after `--save-to-azure-yaml`, the next `azd deploy` calls `CreateVersionInline` with description/instructions/tools only, and the new default version quietly loses license, compatibility, and metadata. `update --file X.md` has the same gap in `runInline`. The archive path isn't affected, since the zip carries SKILL.md as-is.

This is a concrete case of the drift @huimiu asked about above. Options: add the three fields to `skillServiceConfig` and the schema, warn at save time that SKILL.md front matter outside description/instructions/tools won't survive a reconcile, or call out the limitation in the README.

cli/azd/extensions/azure.ai.skills/internal/cmd/skill_service_config.go
```diff
@@ -0,0 +1,268 @@
...
+		)
+	}
+
+	cfg := declaration.Config
+	if declaration.ArchiveSource != "" {
+		serviceRoot := project.GetPath()
+		if found && existing.GetRelativePath() != "" {
+			serviceRoot = filepath.Join(serviceRoot, existing.GetRelativePath())
+		}
+		archive, err := portableSkillArchiveReference(serviceRoot, declaration.ArchiveSource)
```
jongio, line 159: Anchoring to `serviceRoot` narrows what can be saved, and the error message doesn't reflect it.

When an existing skill service carries `project: skills`, `portableSkillArchiveReference` receives `<root>/skills` as its `projectPath` argument, so `filepath.Rel` is computed from the service directory. An archive that lives elsewhere in the same project, say `<root>/shared/triage-rules`, comes back as `../shared/triage-rules` and gets rejected:

```
cannot save archive reference "<root>\shared\triage-rules" to azure.yaml because it is outside the azd project at "<root>\skills"
```

That path is inside the azd project. The message names the service directory as the project, and the `projectPath` parameter name on line 231 is stale for this call.

README line 73 drifts the same way. It says the archive is stored "relative to the azd project", which only holds when the service has no `project:` value. The schema description is the accurate one: "Relative paths resolve from the service path when set, otherwise from the current project directory."

Either check containment against `project.GetPath()` while keeping the saved reference relative to `serviceRoot`, or reword the error and the README in terms of the service directory. Renaming the parameter to `baseDir` would stop the next reader tripping on it.

jongio (commented):
One new low severity item from a fresh pass at a3859312, on top of the notes I left earlier today. Not a blocker and my approval stands.

- `skill_service_config.go:159` - when an existing skill service has a `project:` path, saving an archive anchors to that service directory, so an archive elsewhere in the project is rejected with a message that calls the service directory the azd project. README line 73 and the error text disagree with the schema description.

Issue #1796, opened 2023-03-24
Error installing azd via winget


charris-msft:
- [X] Make sure you've installed the latest version using [instructions in the wiki](../wiki/install)

**Output from `azd version`**
not installed yet

**Describe the bug**
Brand new machine - installing azd before az cli:

```
> winget install microsoft.azd
Found Azure Developer CLI [Microsoft.Azd] Version 0.7.1
This application is licensed to you by its owner.
Microsoft is not responsible for, nor does it grant any licenses to, third-party packages.
Downloading https://github.com/Azure/azure-dev/releases/download/azure-dev-cli_0.7.0-beta.1/azd-windows-amd64.msi
An unexpected error occurred while executing the command:
InternetOpenUrl() failed.
0x80072f05 : unknown error

> winget install microsoft.azd
Found Azure Developer CLI [Microsoft.Azd] Version 0.7.1
This application is licensed to you by its owner.
Microsoft is not responsible for, nor does it grant any licenses to, third-party packages.
Downloading https://github.com/Azure/azure-dev/releases/download/azure-dev-cli_0.7.0-beta.1/azd-windows-amd64.msi
An unexpected error occurred while executing the command:
InternetOpenUrl() failed.
0x80072f05 : unknown error
```
Install Azure CLI
```
> winget install microsoft.azurecli
Found Microsoft Azure CLI [Microsoft.AzureCLI] Version 2.46.0
This application is licensed to you by its owner.
Microsoft is not responsible for, nor does it grant any licenses to, third-party packages.
Downloading https://azcliprod.azureedge.net/msi/azure-cli-2.46.0.msi
  ██████████████████████████████  55.2 MB / 55.2 MB
Successfully verified installer hash
Starting package install...
Successfully installed
```
Now install azd
```
> winget install microsoft.azd
Found Azure Developer CLI [Microsoft.Azd] Version 0.7.1
This application is licensed to you by its owner.
Microsoft is not responsible for, nor does it grant any licenses to, third-party packages.
Downloading https://github.com/Azure/azure-dev/releases/download/azure-dev-cli_0.7.0-beta.1/azd-windows-amd64.msi
  ██████████████████████████████  12.6 MB / 12.6 MB
Successfully verified installer hash
Starting package install...
Successfully installed
```

**To Reproduce**
Brand new machine, before installing Azure CLI
```
winget install microsoft.azd
```

**Expected behavior**
azd installs without error

**Environment**
Information on your environment:
    * Language name and version
[...]

PR #9125, opened 2026-07-13
Auto-enable no-prompt mode in CI environments

vhvb1989:
## Summary

Fixes Azure/azure-dev-pr#1796.

This PR fixes an inconsistency in how azd handles interactive prompts in non-interactive environments: when azd runs in CI **without** `--no-prompt`, interactive confirmations were resolved against an EOF/blank stdin and could silently return the wrong value. Concretely, a `Confirm` with `DefaultValue: true` rendered `(y/N)` and returned `false`, aborting the operation instead of honoring its default.

> Note: this PR is **not** related to the public `Azure/azure-dev#1796` (a Winget download issue). The bare number collision is coincidental.

This PR formalizes azd's behavior into **two well-defined modes**:

1. **azd auto-enables no-prompt mode** whenever it detects a CI/CD environment (Azure Pipelines, GitHub Actions, and the other providers recognized by `resource.IsRunningOnCI()`), mirroring the existing AI-agent auto-detection. This makes CI deterministic: prompts resolve to their defaults, and any required input that has no usable default fails fast with an actionable `PromptRequiredError`.
2. **The interactive-in-a-non-TTY path now honors prompt defaults.** When input is piped in and a `Confirm`/`MultiSelect` reaches a blank line (Enter) or EOF, azd returns the prompt's **default** — consistent with how `Input` and `Select` already fall back to their defaults — instead of silently returning `false`. This is the core bug fix: it removes the previous behavior where a `DefaultValue: true` confirm aborted on EOF.

Explicit `--no-prompt` / `--non-interactive` flags and the `AZD_NON_INTERACTIVE` env var continue to take precedence. Set `AZD_NON_INTERACTIVE=false` to opt out of the automatic no-prompt enablement (the global no-prompt setting stays off). Some commands still avoid interactive prompts in CI/CD by design, independent of this variable.

> **Reviewer guide — review per commit.** This PR touches **115 files, but only 11 contain real changes.** The other 104 are auto-generated usage/fig snapshots that each changed by the same 2 lines, because `--no-prompt` is a global flag and its updated help text appears in every command's usage output (regenerating the snapshots is required or CI fails).
>
> - **Commit 1 — [`ea9420d18`](https://github.com/Azure/azure-dev/pull/9125/commits/ea9420d187f1f3a49c39d65e53a94d20ecf76467)** `feat: auto-enable no-prompt mode in CI; confirms honor defaults on EOF` — the 11 reviewable files (core logic + tests). **Review this commit.**
> - **Commit 2 — [`bfdb93f02`](https://github.com/Azure/azure-dev/pull/9125/commits/bfdb93f02594bd7486f5214537a60dcbec343694)** `chore: regenerate usage/fig snapshots` — 104 mechanical snapshot regenerations. **Skimmable.**

## Changes

- `cmd/auto_install.go` — extend NoPrompt auto-detection to `resource.IsRunningOnCI()` (in addition to `agentdetect.IsRunningInAgent()`); update `--no-prompt` help text and document the `AZD_NON_INTERACTIVE=false` opt-out.
- `pkg/input/asker.go` — non-terminal `Confirm` derives its indicator and blank-line/EOF result from the prompt's `Default` (so `DefaultValue: true` renders `(Y/n)` and returns `true` on Enter/EOF); `MultiSelect` passes each item's default so an exhausted stdin falls back to that item's default selection state.
- `pkg/input/prompt_required_error.go` — reword docs to reflect that `PromptRequiredError` covers auto-enabled no-prompt (CI/agent) cases where a prompt has no usable default; minor `message()` cleanup.
- `internal/reserved_flags.go`, `internal/global_command_options.go` — document auto-enable + `AZD_NON_INTERACTIVE=false` opt-out.
- `docs/environment-variables.md` — document `AZD_NON_INTERACTIVE`.
- `test/azdcli/cli.go` — set `AZD_NON_INTERACTIVE=false` for stdin-driven functional tests (the suite itself runs in CI) so azd keeps reading piped answers, mirroring the existing `AZD_DISABLE_AGENT_DETECT=1` precedent.
- `test/functional/preflight_quota_test.go` — update the preflight stdin helpers to decline the deployment warning explicitly with `"n"` (the tests previously relied on the blank-line=false bug).
- Regenerated usage/fig snapshots for the updated `--no-prompt` help text.

## Testing

- `TestParseGlobalFlags_CIDetection` — CI vars enable no-prompt; explicit flags/`AZD_NON_INTERACTIVE=false` override.
- `Test_NewAsker_NonTerminal_Confirm_EOFReturnsDefault` — piped EOF returns the confirm's default.
- `Test_NewAsker_NonTerminal_Confirm_BlankHonorsDefaultWithZeroInit` — blank line honors `Default` even when the caller passes a zero-initialized `*bool`.
- `Test_NewAsker_NonTerminal_MultiSelect_EOFReturnsDefaults` — mid-selection EOF falls back to per-item defaults.
- `Test_CLI_PreflightQuota_*` functional tests pass in playback with the updated stdin.
- `go build ./...`, `go vet`, `gofmt`, `pkg/input` and `cmd` flag/usage suites all clean.

## Notes

- **Functional tests are not blocked** by the auto no-prompt enablement: stdin-driven tests set `AZD_NON_INTERACTIVE=false` via the test harness so azd keeps reading piped answers.
- No `go.mod`/`go.sum` changes.
- No `CHANGELOG.md` entry — per repo convention, changelog entries are added during the release process to avoid merge conflicts.


cli/azd/cmd/auto_install.go
```diff
@@ -840,9 +842,14 @@ func ParseGlobalFlags(args []string, opts *internal.GlobalCommandOptions) error
...
-	// agent as the caller, automatically enable no-prompt mode for non-interactive execution.
-	if !flagExplicitlySet && !envVarPresent && agentdetect.IsRunningInAgent() {
+	// Auto no-prompt detection: If no explicit flag or env var was set, automatically enable
+	// no-prompt mode when azd runs in a non-interactive context where prompting is not possible —
+	// either an AI coding agent or a CI/CD environment. This makes CI behavior deterministic:
+	// prompts resolve to their defaults (or fail fast with an actionable error) instead of
+	// silently aborting on an EOF stdin. Explicit flags and AZD_NON_INTERACTIVE take precedence;
+	// set AZD_NON_INTERACTIVE=false to opt out and force interactive behavior.
+	if !flagExplicitlySet && !envVarPresent &&
+		(agentdetect.IsRunningInAgent() || resource.IsRunningOnCI()) {
```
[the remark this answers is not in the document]
vhvb1989: Fixed: narrowed the documented contract. The help text, `AZD_NON_INTERACTIVE` env-var doc, code comment, and changelog no longer claim the opt-out 'forces interactive behavior' end to end. They now state it opts out of the automatic no-prompt enablement (global NoPrompt stays false) and explicitly note that some commands still avoid interactive prompts in CI/CD by design, independent of this variable. Rerouting the standalone `resource.IsRunningOnCI()` prompt gates (auth_status, login_guard, etc.) through the resolved no-prompt setting was intentionally left out of scope to avoid regressing their by-design CI behavior.

cli/azd/pkg/input/asker.go
```diff
@@ -172,22 +172,28 @@ func askOnePrompt(p survey.Prompt, response any, isTerminal bool, stdout io.Writ
...
 			err := askOnePrompt(&survey.Confirm{
 				Message: fmt.Sprintf("\n  select %s?", item),
-			}, &response, isTerminal, stdout, stdin)
+			}, &itemResponse, isTerminal, stdout, stdin)
 			if err != nil {
+				// If stdin is exhausted mid-selection, fall back to the default selection
+				// rather than surfacing a per-item prompt-required error.
+				if _, isReq := errors.AsType[*PromptRequiredError](err); isReq {
+					*(response.(*[]string)) = defValue
+					return nil
```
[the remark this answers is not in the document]
vhvb1989: Fixed: the MultiSelect EOF branch no longer replaces `selection` with `defValue`. It now propagates the `PromptRequiredError` on mid-selection EOF (consistent with the Confirm fix), so answered items are never silently discarded/misreported. Added regression test `Test_NewAsker_NonTerminal_MultiSelect_EOFReturnsPromptRequired` covering both immediate EOF and the partial-input (`n\n` then EOF) case.

cli/azd/CHANGELOG.md
```diff
@@ -10,6 +10,8 @@
 
 ### Bugs Fixed
 
+- azd now automatically enables no-prompt mode when it detects a CI/CD environment (Azure Pipelines, GitHub Actions, and other supported providers), matching existing AI-agent auto-detection. Previously, running in CI without `--no-prompt` could cause interactive confirmations to resolve incorrectly against an EOF stdin — for example, a confirmation with a `true` default would silently return `false` and abort the operation. Confirmations now honor their defaults in CI. Set `AZD_NON_INTERACTIVE=false` to opt out and force interactive behavior.
```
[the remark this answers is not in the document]
vhvb1989: Fixed: updated the PR description. It now references the internal tracking issue `Azure/azure-dev-pr#1796` and explicitly notes this is unrelated to the public `Azure/azure-dev#1796` (Winget download issue) — the number collision is coincidental. The CHANGELOG entry itself does not reference an issue number.

cli/azd/pkg/input/asker.go
```diff
@@ -251,8 +255,22 @@ func askOnePrompt(p survey.Prompt, response any, isTerminal bool, stdout io.Writ
 				*pResponse = false
 				return nil
 			case "":
+				// An explicit empty line (the user pressed Enter) accepts the current value.
```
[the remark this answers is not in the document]
vhvb1989: Fixed: the non-terminal `Confirm` now bases both the (Y/n) vs (y/N) indicator and the blank-line (Enter) result on `v.Default` instead of the incoming `*pResponse`, so `DefaultValue: true` is honored even though `AskerConsole.Confirm` passes a zero-initialized bool. Also set `Default: itemResponse` on the synthetic MultiSelect confirmations so their per-item defaults are preserved. Added regression test `Test_NewAsker_NonTerminal_Confirm_BlankHonorsDefaultWithZeroInit` (checks both the returned value and the rendered indicator for zero-init callers).

cli/azd/cmd/auto_install.go
```diff
@@ -840,9 +842,15 @@ func ParseGlobalFlags(args []string, opts *internal.GlobalCommandOptions) error
 		}
 	}
 
-	// Agent Detection: If no explicit flag or env var was set and we detect an AI coding
-	// agent as the caller, automatically enable no-prompt mode for non-interactive execution.
-	if !flagExplicitlySet && !envVarPresent && agentdetect.IsRunningInAgent() {
+	// Auto no-prompt detection: If no explicit flag or env var was set, automatically enable
```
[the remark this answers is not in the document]
vhvb1989: Fixed: updated the `ParseGlobalFlags` Go doc to describe the broadened behavior — automatic no-prompt now applies to both AI coding agents and CI/CD environments — and documented the override precedence (explicit flags / AZD_NON_INTERACTIVE win; AZD_NON_INTERACTIVE=false opts out).

cli/azd/test/azdcli/cli.go
```diff
@@ -179,6 +180,17 @@ func (cli *CLI) RunCommandWithStdIn(ctx context.Context, stdin string, args ...s
 	}
 	cmd.Env = append(cmd.Env, agentdetect.DisableAgentDetectEnvVar+"=1")
 
+	// Also opt out of CI-based auto no-prompt detection. The functional test suite itself often
+	// runs inside a CI environment (GITHUB_ACTIONS/TF_BUILD/CI/...), which would otherwise make
+	// the child azd process auto-enable --no-prompt mode and stop reading the stdin answers that
+	// stdin-driven tests provide. AZD_NON_INTERACTIVE=false forces interactive behavior. We only
```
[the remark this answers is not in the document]
vhvb1989: Fixed: reworded the harness comment. It no longer claims AZD_NON_INTERACTIVE=false 'forces interactive behavior'; it now says it keeps the global no-prompt setting off so the asker keeps reading piped stdin, and explicitly notes it does not override command-specific resource.IsRunningOnCI() gates (which don't affect the stdin-driven prompts under test).

cli/azd/cmd/auto_install.go
```diff
@@ -840,9 +845,15 @@ func ParseGlobalFlags(args []string, opts *internal.GlobalCommandOptions) error
...
+	// no-prompt mode when azd runs in a non-interactive context where prompting is not possible —
+	// either an AI coding agent or a CI/CD environment. This makes CI behavior deterministic:
+	// prompts resolve to their defaults (or fail fast with an actionable error) instead of
+	// silently aborting on an EOF stdin. Explicit flags and AZD_NON_INTERACTIVE take precedence;
+	// set AZD_NON_INTERACTIVE=false to opt out of this automatic enablement (NoPrompt stays false).
+	// Note: some commands still avoid interactive prompts in CI/CD by design, independent of this.
+	if !flagExplicitlySet && !envVarPresent &&
+		(agentdetect.IsRunningInAgent() || resource.IsRunningOnCI()) {
```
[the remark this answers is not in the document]
vhvb1989: Fixed: updated the GlobalCommandOptions.NoPrompt Go doc 'Can be enabled via' list to include automatic CI/CD detection alongside agent detection, and noted the AZD_NON_INTERACTIVE=false opt-out.

cli/azd/internal/reserved_flags.go
```diff
@@ -33,7 +33,9 @@ var reservedFlags = []ReservedFlag{
 		Long:  "no-prompt",
 		Short: "",
 		Description: "Runs without prompts. Uses existing values; " +
-			"fails if any required value or decision cannot be resolved automatically.",
+			"fails if any required value or decision cannot be resolved automatically. " +
+			"Automatically enabled when azd detects a CI/CD or AI-agent environment; " +
+			"set AZD_NON_INTERACTIVE=false to opt out of that automatic enablement.",
```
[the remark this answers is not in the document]
vhvb1989: Not taken (false positive). `TestUsage` passes as-is — I re-ran `go test ./cmd -run 'TestUsage|TestFigSpec'` after regenerating snapshots and it is green. `TestUsage-azd-ai-training.snap` intentionally keeps the shorter text because `azd ai training` is served by the **azure.ai.finetune extension**, which defines its own `--no-prompt` help in its own binary (`cli/azd/extensions/azure.ai.finetune/internal/cmd/root.go:43`). Changes to core's `CreateGlobalFlagSet`/`reserved_flags.go` do not (and should not) alter an extension's own flag help. All core command snapshots (104 of them) were regenerated and carry the new text; the extension snapshots correctly do not.

cli/azd/pkg/input/asker.go
```diff
@@ -251,8 +261,23 @@ func askOnePrompt(p survey.Prompt, response any, isTerminal bool, stdout io.Writ
...
 				return nil
 			case "":
+				// An explicit empty line (the user pressed Enter) accepts the default value.
+				// However, reaching EOF without any answer means stdin is exhausted in a
+				// non-interactive context: fail loud with a PromptRequiredError instead of
+				// silently returning a value. In real CI/agent runs azd auto-enables no-prompt
+				// mode, so this path is only reached when interactive mode was explicitly forced
+				// (e.g. AZD_NON_INTERACTIVE=false) or the piped input was underfed.
+				if errors.Is(err, io.EOF) {
+					return promptRequiredForPrompt(v.Message)
```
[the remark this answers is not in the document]
vhvb1989: Fixed: broadened the `PromptRequiredError` type doc to state it is returned both in --no-prompt mode (including CI/CD- and agent-auto-enabled) and when a non-terminal prompt hits EOF / exhausts piped stdin. Also corrected the `PromptMessage` field doc, which no longer claims the text was never displayed — it may have been printed to stdout just before EOF in the non-terminal path.

cli/azd/pkg/input/prompt_required_error.go
```diff
@@ -43,7 +43,10 @@ type RequiredInput struct {
 	Sources     []InputSource `json:"sources,omitempty"`
 }
 
-// PromptRequiredError is returned when --no-prompt mode prevents collecting required inputs interactively.
+// PromptRequiredError is returned when azd needs interactive input but cannot collect it.
+// This happens in --no-prompt mode (including when auto-enabled for CI/CD or agent environments),
+// and also when a non-terminal prompt reaches EOF or exhausts its piped stdin before an answer is
+// provided.
```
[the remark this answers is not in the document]
vhvb1989: Fixed: gave the EOF/stdin-exhaustion path a distinct reason and messaging so it no longer reuses the no-prompt wording.

- Added a `StdinExhausted` field on `PromptRequiredError` and a `promptRequiredStdinExhausted` constructor used by the two non-terminal Confirm EOF sites (the MultiSelect path propagates it).
- `message()` now returns a new `StdinExhaustedPromptMessage` ("...reached end of input before a prompt was answered") instead of the "interactive prompts disabled" headline for this case.
- `promptMessageToString()` renders EOF-specific, actionable guidance ("Input ended before this prompt was answered. To proceed, supply an answer... or run interactively.") instead of "run this command in interactive mode".
- `MarshalJSON` emits a distinct `type: stdinExhausted` and the new headline.
- The no-prompt (`askOneNoPrompt`) failures keep the original wording/type via `promptRequiredForPrompt`.

Added output tests for both cases: `TestPromptRequiredError_ToString_UsesStdinExhaustedWording`, `TestPromptRequiredError_MarshalJSON_StdinExhausted`, and asserted `StdinExhausted` in the asker Confirm/MultiSelect EOF tests.

cli/azd/pkg/input/prompt_required_error.go
```diff
@@ -43,7 +43,9 @@ type RequiredInput struct {
 	Sources     []InputSource `json:"sources,omitempty"`
 }
 
-// PromptRequiredError is returned when --no-prompt mode prevents collecting required inputs interactively.
+// PromptRequiredError is returned when azd needs interactive input but cannot collect it.
+// This happens in --no-prompt mode (including when auto-enabled for CI/CD or agent environments),
```
richardpark-msft: Maybe just refer to it as 'no prompt mode', because you can activate this through several paths, right? Using the env var, and two separate command line switches (--no-prompt and --non-interactive)

cli/azd/pkg/input/prompt_required_error.go
```diff
@@ -52,7 +54,8 @@ type PromptRequiredError struct {
 	// Message is the headline used for structured missing-input output.
 	Message string
 
-	// PromptMessage is the prompt text that would have been displayed to the user if interactive prompts were allowed.
+	// PromptMessage is the text of the prompt that could not be answered. It may not have been
+	// rendered to the user (e.g. no-prompt mode short-circuits before display).
```
richardpark-msft: Is it a bug if the text was displayed, or expected? The comment, as it is, covers both outcomes but I'm not sure what I can do with that information.

cli/azd/pkg/input/asker.go
```diff
@@ -231,10 +234,15 @@ func askOnePrompt(p survey.Prompt, response any, isTerminal bool, stdout io.Writ
 			strings.Join(v.Options, ","))
 	case *survey.Confirm:
 		var pResponse = response.(*bool)
+		// Base the indicator and the blank-line (Enter) result on the prompt's Default rather
```
richardpark-msft: this comment probably doesn't need to exist - it's explaining why you fixed the bug, but nobody that's in here later would need to know that.

jongio (approved):
Reviewed the 11 non-snapshot files. The core change (adding `resource.IsRunningOnCI()` to the auto no-prompt gate) and the non-TTY Confirm/MultiSelect default handling look correct to me.

On the one open thread about EOF being treated like Enter: returning the prompt's default on EOF is consistent with how `Input` and `Select` already behave in this non-TTY path, and the Confirm loop still terminates (an invalid token followed by EOF falls through to the `errors.Is(err, io.EOF)` guard rather than looping). So I don't see the silent-decision behavior as a regression relative to the existing Input/Select contract. Reasonable to keep as-is unless you specifically want Confirm to be stricter than the other prompt types.

Verified locally: `go build ./...` is clean, and `pkg/input` plus `cmd` `TestParseGlobalFlags` tests pass.

cli/azd/internal/reserved_flags.go
```diff
@@ -33,7 +33,9 @@ var reservedFlags = []ReservedFlag{
 		Long:  "no-prompt",
 		Short: "",
 		Description: "Runs without prompts. Uses existing values; " +
-			"fails if any required value or decision cannot be resolved automatically.",
+			"fails if any required value or decision cannot be resolved automatically. " +
```
richardpark-msft, line 36: Wait, you're harassing me about double-negatives!? :)

cli/azd/internal/global_command_options.go
```diff
@@ -31,7 +31,8 @@ type GlobalCommandOptions struct {
 	//   - --no-prompt flag
 	//   - --non-interactive flag (alias for --no-prompt)
 	//   - AZD_NON_INTERACTIVE=true environment variable
-	//   - Automatic agent detection (lowest priority)
+	//   - Automatic agent or CI/CD detection (lowest priority; opt out with
+	//     AZD_NON_INTERACTIVE=false)
```
richardpark-msft, line 35: It seems like, most places, we call it no-prompt, and that's the only documented switch. Is there a reason we don't use the same terminology here?

cli/azd/docs/environment-variables.md
```diff
@@ -45,6 +45,7 @@ integration.
 | `AZD_CONFIG_DIR` | The file path of the user-level configuration directory. |
 | `AZD_DEMO_MODE` | If true, enables demo mode. This hides personal output, such as subscription IDs, from being displayed in output. |
 | `AZD_FORCE_TTY` | If true, forces `azd` to write terminal-style output. |
+| `AZD_NON_INTERACTIVE` | Controls no-prompt mode. Parsed as a boolean (`true`/`false`/`1`/`0`). Set to a truthy value to run without interactive prompts (equivalent to `--no-prompt`). `azd` also auto-enables no-prompt mode when it detects a CI/CD or AI-agent environment; setting this variable (to **any** value, including `false`) opts out of that automatic enablement, so `AZD_NON_INTERACTIVE=false` keeps the global no-prompt setting off in those environments. Note that some commands still avoid interactive prompts in CI/CD by design, independent of this variable. Explicit `--no-prompt`/`--non-interactive` flags take precedence over this variable. |
```
richardpark-msft: I would be a bit more precise on this language, since it's docs.

Some places (maybe all?) are using strconv.ParseBool, so your list of values that can be accepted is limited and other values will just cause an error. 

So I'd remove any language in here that implies anything outside of the known range is accepted as that's an implementation detail, and only document the actual values we expect.

cli/azd/cmd/auto_install.go
```diff
@@ -726,7 +726,9 @@ func CreateGlobalFlagSet() *pflag.FlagSet {
 		"no-prompt",
 		false,
 		"Runs without prompts. Uses existing values; "+
-			"fails if any required value or decision cannot be resolved automatically.")
+			"fails if any required value or decision cannot be resolved automatically. "+
+			"Automatically enabled when azd detects a CI/CD or AI-agent environment; "+
+			"set AZD_NON_INTERACTIVE=false to opt out of that automatic enablement.")
```
richardpark-msft, line 731: Similar criticism from another comment - let's standardize on the user-facing terminology for this. If the only public flag is --no-prompt, then that should be the name here too.

richardpark-msft (changes requested):

cli/azd/internal/global_command_options.go
vhvb1989: fare. we should all use the same terminology. But it's too late to change ENV VARS names.
Non interactive and --no-prompt are similar but they were created for different purposes.
Non interactive was born (IIRC) when we needed to have a way to run locally something like if we were running in CI environment.
--no-prompt was born to explicitly to shortcut any UI question from azd to the user with its default value.

So, in theory you could have a non-interactive environment where you want to run w/o no-prompt -- but that's usually unexpected. So in practice, both now reference a common intention.  So it is basically that we missed some coordination between authors/developers at some point XD

cli/azd/pkg/input/asker.go
```diff
@@ -251,6 +259,18 @@ func askOnePrompt(p survey.Prompt, response any, isTerminal bool, stdout io.Writ
 				*pResponse = false
 				return nil
 			case "":
+				// An empty line (the user pressed Enter) or EOF (piped stdin exhausted) accepts
+				// the prompt's default value, consistent with how Input and Select fall back to
+				// their defaults. Real CI/agent runs auto-enable no-prompt mode (askOneNoPrompt),
+				// so this interactive path is only reached when input is piped in.
+				*pResponse = defaultValue
```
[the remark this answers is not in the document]
vhvb1989: Rejected — this is the intended design, not a regression. Treating EOF and a blank line identically (both return the prompt's `Default`) is the deliberate **Option 1** behavior chosen for this PR; the earlier "fail-loud legacy mode" (and the `StdinExhausted` machinery) was intentionally removed, and the PR description was rewritten to describe this. Returning the default on EOF is exactly what fixes the original CI silent-abort bug: real CI/agent runs auto-enable no-prompt (which resolves defaults anyway), and this interactive non-terminal path is only reachable with piped/underfed stdin, where honoring the default is consistent with how `Input` and `Select` already fall back to their defaults. `AZD_NON_INTERACTIVE=false` only opts out of the *global* auto no-prompt enablement; it does not re-introduce a silent-wrong-decision, because the confirm now honors `DefaultValue` rather than the old buggy `false`.

cli/azd/internal/global_command_options.go
richardpark-msft: To clarify:

- `--no-prompt` only means that you'll get the same result, as if you just hit ENTER at the prompt
- non-interactive: _fails_ instead of choosing the default?

I'm trying to get where they are different.

cli/azd/cmd/auto_install.go
vhvb1989: Totally agree with standardizing on `--no-prompt` as the user-facing concept name, and I've tried to keep the prose oriented around that everywhere. 🙂

The one spot I intentionally kept the `AZD_NON_INTERACTIVE` name is this opt-out sentence, and here's the reasoning:

- This flag help documents a *new* automatic behavior (azd auto-enabling no-prompt in CI/CD and agent environments). I think it's important that users can discover, right from `--help`, how to turn that automatic behavior off.
- There's no `--no-prompt`-style flag that *forces* prompts back on (`--no-prompt=false` is just the default and doesn't override auto-detection), so an env var is the only opt-out mechanism.
- Rather than introduce a brand-new env var, I reused the one azd already ships (`AZD_NON_INTERACTIVE`). Renaming it would be a breaking change, so the env var name itself is fixed — but I didn't want to hide the opt-out from the help either.

So the `AZD_NON_INTERACTIVE` token here isn't a competing concept name — it's the concrete, pre-existing knob users need to disable the feature. Happy to reword the sentence if you have a phrasing that reads more cleanly while still surfacing the opt-out. Does that work for you?

cli/azd/pkg/input/prompt_required_error.go
vhvb1989: Accepted — fixed in 33eeb57db. Reworded the `PromptRequiredError` doc to describe 'no-prompt mode' generically (set explicitly via `--no-prompt`/`--non-interactive` or `AZD_NON_INTERACTIVE`, or auto-enabled for CI/CD or agent environments).

cli/azd/pkg/input/prompt_required_error.go
vhvb1989: Accepted — fixed in 33eeb57db. Dropped the hedge about whether it was rendered; the field doc is now just: 'PromptMessage is the text of the prompt that could not be answered non-interactively.'

cli/azd/pkg/input/asker.go
vhvb1989: Accepted — fixed in 33eeb57db. Trimmed the bug-history explanation down to a single-line comment describing the current behavior.

cli/azd/docs/environment-variables.md
vhvb1989: Accepted — fixed in 33eeb57db. Tightened the docs wording: it now says the variable accepts a boolean (`true`, `false`, `1`, `0`) with other values ignored (with a warning), and clarifies the `AZD_NON_INTERACTIVE=false` opt-out and that some commands still avoid prompts in CI/CD by design.

cli/azd/internal/global_command_options.go
vhvb1989: Same thinking as the `auto_install.go` thread. 🙂 This doc intentionally leads with the `NoPrompt` concept and the `--no-prompt` flag as the canonical name — the `--non-interactive` and `AZD_NON_INTERACTIVE` entries in the 'Can be enabled via' list aren't rival concept names, they're the concrete switches a reader needs: the pre-existing alias flag and the pre-existing env var.

Unlike the user-facing `--help` text, this is an internal Go doc comment, so I think it's actually useful to spell out the real env var (`AZD_NON_INTERACTIVE=true` to enable, `=false` to opt out of auto-detection) rather than hide it behind the concept name. We reused the existing `AZD_NON_INTERACTIVE` var instead of minting a new one, and renaming it would be breaking — so the name is fixed. Open to a wording tweak if you'd like it phrased differently.

cli/azd/cmd/auto_install.go
richardpark-msft: > This flag help documents a new automatic behavior (azd auto-enabling no-prompt in CI/CD and agent environments). I think it's important that users can discover, right from --help, how to turn that automatic behavior off.

That's not my criticism at all - it should be documented. This is more noticing an inconsistency in the switch vs the env var.

> There's no --no-prompt-style flag that forces prompts back on (--no-prompt=false is just the default and doesn't override auto-detection), so an env var is the only opt-out mechanism.

`--no-prompt=false` would be weird, and adding in more flags (ie, every 'no' also has a 'yes' equivalent) would be awful. I think you've taken the best option.

> Rather than introduce a brand-new env var, I reused the one azd already ships (AZD_NON_INTERACTIVE).

Ah, did it have another use-case before? Or is this more "fulfilling the actual intention of the original variable"?

cli/azd/docs/environment-variables.md
```diff
@@ -45,6 +45,7 @@ integration.
 | `AZD_CONFIG_DIR` | The file path of the user-level configuration directory. |
 | `AZD_DEMO_MODE` | If true, enables demo mode. This hides personal output, such as subscription IDs, from being displayed in output. |
 | `AZD_FORCE_TTY` | If true, forces `azd` to write terminal-style output. |
+| `AZD_NON_INTERACTIVE` | Controls no-prompt mode. Accepts a boolean (`true`, `false`, `1`, `0`); other values are ignored with a warning. Set to `true` (or `1`) to run without interactive prompts (equivalent to `--no-prompt`). `azd` also auto-enables no-prompt mode when it detects a CI/CD or AI-agent environment; set `AZD_NON_INTERACTIVE=false` to opt out of that automatic enablement (the global no-prompt setting stays off in those environments). Note that some commands still avoid interactive prompts in CI/CD by design, independent of this variable. Explicit `--no-prompt`/`--non-interactive` flags take precedence over this variable. |
```
jongio, line 48: This now says non-boolean values are `ignored with a warning`, but the parser in `cmd/auto_install.go` sets `envVarPresent = true` as soon as the variable exists, before `strconv.ParseBool` runs. So an unparseable value (a typo like `AZD_NON_INTERACTIVE=yes`) still counts as present and suppresses the CI/agent auto no-prompt detection, leaving `NoPrompt` false. In CI that means a typo silently keeps azd interactive, which is the exact mode this PR is trying to avoid, so `ignored` isn't quite accurate.

Two ways to reconcile: move `envVarPresent = true` inside the `err == nil` branch so invalid values are truly ignored and CI auto-detection still applies, or keep the current behavior and reword this to say any present value (including an invalid one) opts out of auto-detection. Either way a CI-plus-invalid-value test case would lock in the chosen behavior.

jongio (commented):
Incremental look at the new commit. The terminology cleanups read well and the reworded docs/comments match the intent. One doc-vs-behavior mismatch on the reworded `AZD_NON_INTERACTIVE` line, noted inline.

cli/azd/internal/global_command_options.go
vhvb1989: > To clarify:
> 
> --no-prompt only means that you'll get the same result, as if you just hit ENTER at the prompt
> non-interactive: fails instead of choosing the default?

Ok, I was confusing the env var with another one... I did a little bit more research and I have a loooong reading for you that explains it better.  I apologize for how long it is.. But I have read it and I promise it's worth it if you want to understand azd UI modes and the way to control it.  It helped me to remember :)

-----

# `AZD_NON_INTERACTIVE` vs `AZD_FORCE_TTY`

## What I was confusing

I was conflating two independent environment variables: **`AZD_NON_INTERACTIVE`** and
**`AZD_FORCE_TTY`**. I assumed `AZD_NON_INTERACTIVE` was what turned off the terminal / TTY
behavior (i.e. that it set `isTerminal = false`). It doesn't. TTY detection and no-prompt mode
are two completely separate axes in `azd`, each driven by its own env var, and neither one reads
the other. `AZD_NON_INTERACTIVE` only ever affects the *no-prompt* axis; `AZD_FORCE_TTY` is the
one that affects the *TTY* axis.

## What each one actually does

**`AZD_NON_INTERACTIVE` — controls no-prompt mode (`opts.NoPrompt`).**
It is exactly equivalent to the `--no-prompt` flag (and its hidden alias `--non-interactive`);
all of them funnel into the same internal `NoPrompt` value. When no-prompt mode is on, the UI
controls never read from stdin at all: each prompt uses its configured **default** if it has one,
and otherwise **fails loudly** with `PromptRequiredError` instead of guessing or blocking. The
only real difference between the flag and the env var is the delivery mechanism (an invocation
flag vs. an environment setting for when you can't control the command line) and precedence: an
explicitly-set flag wins over the env var, and the mere *presence* of `AZD_NON_INTERACTIVE`
(any value) suppresses the automatic CI/agent enablement.

**`AZD_FORCE_TTY` — controls TTY detection (`isTerminal`).**
This is resolved in `internal/terminal.IsTerminal`, which never looks at `AZD_NON_INTERACTIVE`.
`AZD_FORCE_TTY=true` forces `isTerminal = true` and `AZD_FORCE_TTY=false` forces
`isTerminal = false`; otherwise `azd` auto-detects (CI and AI-agent environments are treated as
non-TTY, and finally it falls back to a real `isatty` check). The `isTerminal` value only matters
when prompting is *allowed* (no-prompt is off): it decides *how* we prompt — the rich interactive
TUI on a real terminal, versus a plain "print the question and read a line from piped stdin"
fallback in a non-TTY environment.

## How they affect the experience (the two axes combined)

- **`isTerminal && !noPrompt`** → rich interactive prompts; the user answers.
- **`noPrompt` (regardless of TTY)** → deterministic: default-or-fail, stdin is never read.
  No-prompt always wins over TTY — even on a real terminal, setting `AZD_NON_INTERACTIVE`/
  `--no-prompt` skips prompting entirely.
- **`!noPrompt && !isTerminal`** → non-TTY fallback: the question is printed and a line is read
  from piped stdin, with Enter/EOF falling back to the default. This path's outcome depends on
  what is piped in and on EOF timing.

So: to stop `azd` from prompting, use `AZD_NON_INTERACTIVE` (or `--no-prompt`). To change whether
`azd` believes it is attached to a terminal, use `AZD_FORCE_TTY`. They are orthogonal knobs.

## How this relates to the PR

Before this PR, a CI run that set no flag landed in the `!noPrompt && !isTerminal` fallback: CI
already forced `isTerminal = false`, but `noPrompt` stayed off, so behavior hinged on the
piped-stdin / EOF path — which is fragile (an unconsumed read or a missing EOF could misbehave or
resolve prompts inconsistently). This PR auto-enables **no-prompt mode** in CI/agent environments
(alongside the TTY axis that was already forced off), so those runs now land in the deterministic
no-prompt path: defaults are used where available, and anything genuinely required fails fast with
an actionable error, with no dependence on stdin/EOF at all. `AZD_NON_INTERACTIVE=false` is the
opt-out for this new automatic enablement (it keeps `NoPrompt` off), while `AZD_FORCE_TTY` remains
the separate, unchanged knob for the TTY axis.

cli/azd/internal/reserved_flags.go
vhvb1989: I was just double checking !!!! LOL

cli/azd/cmd/auto_install.go
vhvb1989: >Ah, did it have another use-case before?

It was only to have a way enable no prompt w/o using a calling flag.  Like, in case you want to be calling multiple azd commands and you don't want to keep adding the flag to each one, you would just set the env var. That's an example.

AZD_NON_INTERACTIVE=false was un-used...  and this PR gave it a purpose to disable the auto-noPrompt feature.  But I did not make `--no-prompt=false` to have the same behavior.  the ENV VAR in this case is the only opt-out

cli/azd/internal/global_command_options.go
richardpark-msft: Okay, so now that all makes sense. I think we're back to the original problem, which is that the names have diverged.

However, I think given:
- We've already published the variable before (for some other reason, unrelated to TTYs!
- This is a pretty niche case

I'm less worried about it now.

cli/azd/cmd/auto_install.go
richardpark-msft: I guess the issue here is that we have detection code that sometimes we don't want to use, and --no-prompt=false was too gross, but the env var was a bit less ugly. It's not _great_, but I think we know that, and it's very niche, so I'm not going to sweat about it too much.

richardpark-msft (approved):
Looks like @jongio and CCR are still replying to you, but I'm good on my end.

jongio (approved):
The follow-up commit is doc and comment only. I verified the `AZD_NON_INTERACTIVE` parsing matches the updated docs: non-boolean values are logged as a warning and ignored via strconv.ParseBool, and valid true/false/1/0 values behave as described. Re-approving against the latest commit.

cli/azd/internal/global_command_options.go
vhvb1989: >names have diverged.

yes. sadly is not the only case you will find about diverging names/concepts around the code u.u
But mostly not critical things

thanks for looking into this

jongio (approved):

cli/azd/cmd/auto_install.go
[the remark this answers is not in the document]
vhvb1989: Good catch — fixed in d0f323bd6. Moved `envVarPresent = true` inside the `err == nil` branch so only a successfully parsed boolean counts as an explicit choice. An unparseable value (e.g. `yes`) is now genuinely ignored: it neither sets `NoPrompt` nor suppresses CI/agent auto-detection, so a typo in CI still resolves to deterministic no-prompt instead of silently staying interactive. The valid `=false` opt-out is unchanged. Added a CI-plus-invalid-value regression case in `TestParseGlobalFlags_CIDetection`.

cli/azd/docs/environment-variables.md
vhvb1989: Thanks — you're right, 'ignored' wasn't accurate with the old parser. Fixed in d0f323bd6 by taking your first option: `envVarPresent = true` now lives inside the `err == nil` branch, so an invalid value (like `AZD_NON_INTERACTIVE=yes`) is truly ignored and CI/agent auto no-prompt detection still applies. That makes the doc wording ('other values are ignored with a warning') correct, and a typo no longer silently keeps azd interactive in CI. Added a CI-plus-invalid-value regression test to lock the behavior in.

cli/azd/test/azdcli/cli.go
```diff
@@ -179,6 +180,20 @@ func (cli *CLI) RunCommandWithStdIn(ctx context.Context, stdin string, args ...s
...
+	// stdin-driven tests provide. AZD_NON_INTERACTIVE=false keeps the global no-prompt setting off
+	// so the asker keeps reading piped stdin. (It does not override command-specific
+	// resource.IsRunningOnCI() gates, but those don't affect the stdin-driven prompts under test.)
+	// We only set it when the test hasn't already chosen a value, so tests can still opt into
+	// no-prompt.
+	if !slices.ContainsFunc(cmd.Env, func(e string) bool {
+		return strings.HasPrefix(e, "AZD_NON_INTERACTIVE=")
+	}) {
+		cmd.Env = append(cmd.Env, "AZD_NON_INTERACTIVE=false")
+	}
```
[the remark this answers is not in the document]
vhvb1989: Valid observation about coverage. The broad `AZD_NON_INTERACTIVE=false` override is intentional here: the functional suite frequently runs inside CI, and without it every child `azd` would auto-enable no-prompt and stop reading the piped stdin that the interactive tests depend on. Scoping the override to only stdin-supplying calls would let non-stdin functional commands auto-enable no-prompt in CI, changing their prompt/error paths across many existing tests — that needs a full functional-suite validation and is out of scope for this PR. Tracking as a follow-up to add dedicated coverage for the production CI auto-no-prompt path without destabilizing the existing suite. Resolving here.

jongio (approved):
Re-reviewed the incremental changes since my last approval. The AZD_NON_INTERACTIVE parsing fix is correct: moving `envVarPresent = true` inside the `err == nil` branch means an invalid value (e.g. a typo like `yes`) no longer suppresses CI/agent auto-detection, so CI still resolves to deterministic no-prompt instead of silently staying interactive. The downstream `!envVarPresent && (IsRunningInAgent() || IsRunningOnCI())` gate confirms the behavior, and the new `CI with invalid AZD_NON_INTERACTIVE` test case covers the regression. Docs and doc comments now match the code.

jongio (approved):
Re-reviewed the two commits since my last pass. The AZD_NON_INTERACTIVE fix is correct: only a successfully parsed bool sets envVarPresent, so an invalid value now leaves CI auto-detection intact instead of silently suppressing it, and the false opt-out still works. The added regression test covers the invalid-value-in-CI case. The doc and comment edits are accurate.

jongio (approved):
Re-reviewed the two new commits since my last approval. The AZD_NON_INTERACTIVE parsing fix is correct: only a successfully parsed boolean sets ^[nvVarPresent, so an invalid value (e.g. a typo) no longer suppresses CI/agent auto-detection and CI still resolves to deterministic no-prompt. The false opt-out and true opt-in paths are unchanged, and the added CI-plus-invalid-value test covers the regression. Doc and comment clarifications read accurately. Still good from my side.

jongio (commented):
Note: my previous approval comment had some formatting garbled. Clean version below.

Re-reviewed the two new commits since my last approval. The `AZD_NON_INTERACTIVE` parsing fix is correct: only a successfully parsed boolean sets `envVarPresent`, so an invalid value (e.g. a typo) no longer suppresses CI/agent auto-detection and CI still resolves to deterministic no-prompt. The `false` opt-out and `true` opt-in paths are unchanged, and the added CI-plus-invalid-value test covers the regression. Doc and comment clarifications read accurately. Still good from my side.

JeffreyCA (approved):

jongio (approved):
Re-reviewed the two follow-up commits. The invalid `AZD_NON_INTERACTIVE` handling now only suppresses CI/agent auto-detection on a successfully parsed boolean, so a typo in CI still resolves to deterministic no-prompt. Logic and the added regression test check out.

jongio (approved):
Re-reviewed the two new commits since my last approval.

The AZD_NON_INTERACTIVE fix is correct: moving nvVarPresent = true inside the rr == nil branch means an unparseable value no longer suppresses the `!envVarPresent` CI/agent auto-detection, so a typo still resolves to deterministic no-prompt. The `=false` opt-out (valid bool) still suppresses it, and the added `CI with invalid AZD_NON_INTERACTIVE` test case pins the behavior. Doc and comment clarifications match the runtime behavior.

jongio (approved):
Re-reviewed the two new commits since my last approval.

The `AZD_NON_INTERACTIVE` fix is correct: moving `envVarPresent = true` inside the `err == nil` branch means an unparseable value no longer suppresses the `!envVarPresent` CI/agent auto-detection, so a typo still resolves to deterministic no-prompt. The `=false` opt-out (valid bool) still suppresses it, and the added "CI with invalid AZD_NON_INTERACTIVE" test case pins the behavior. Doc and comment clarifications match the runtime behavior.

Issue #9085, opened 2026-07-10
azure.ai.projects: own the azure.ai.project service block in azure.yaml


therealjohn:
## Summary

Give `azure.ai.projects` ownership of its `host: azure.ai.project` service block in `azure.yaml` - authoring, reading, and reconciling it - instead of `azure.ai.agents` authoring it on the project's behalf.

## Background

Each Foundry resource should map to one `azure.yaml` service entry, keyed by the resource name, with `host: azure.ai.<kind>`, authored and read by the extension that owns that kind. Today `azure.ai.agents` authors the sibling service blocks for other resources, so adopting one capability pulls in the agents extension's knowledge of every other resource's shape, and composable adoption is blocked. This work item covers the `azure.ai.project` host.

## Current state

- `azure.ai.projects` registers a service-target for `host: azure.ai.project` (reconcile on `azd up`/`provision`/`deploy`) but does not author or read the block.
- `azure.ai.agents` writes the block in `resource_services.go` (`emitResourceServices`, host const `AiProjectHost`) and reads it back (`collectProjectDeployments`), with a fallback to config bundled on the agent service.

## Proposed

- `azure.ai.projects` authors and updates its own block through the azd extension framework (`AddService`, `SetServiceConfigSection`/`SetServiceConfigValue`), reads it (`Get`, `GetServiceConfigSection`), and continues to reconcile it through its existing service-target. Authoring is driven from the extension's own commands (for example `project set`, provision).
- `azure.ai.agents` stops authoring `azure.ai.project` and declares the dependency with `uses:` instead. Its adopt/scan recognizes the block generically rather than hardcoding its shape.

The block carries: endpoint, model deployments.

```yaml
services:
  my-project:
    host: azure.ai.project
    endpoint: https://my-proj.services.ai.azure.com/api/projects/my-project
    # model deployments
```

## Acceptance criteria

- [x] `azure.ai.projects` authors its `host: azure.ai.project` block (create + idempotent update) through the azd extension framework.
- [x] `azure.ai.projects` reads its own block for provision/deploy instead of relying on `azure.ai.agents`.
- [x] `azure.ai.agents` no longer authors `azure.ai.project`; it declares the dependency via `uses:` only.
- [x] `azure.ai.agents` adopt/scan recognizes the block via generic host handling (no hardcoded shape).
- [x] A developer can add the project block to `azure.yaml` without the agents extension authoring it.
- [x] A pre-split `azure.yaml` still provisions (back-compat retained).

[...]

PR #9133, opened 2026-07-14
feat: transfer Foundry provisioning to projects

### LINKED ISSUE #9134: [Task] Coordinate Foundry provider handoff release
## Goal

Complete the coordinated release for the Foundry provisioning ownership handoff in PR #9133.

## Scope

- Publish `azure.ai.projects` version `1.0.0-beta.3` with ownership of the `microsoft.foundry` provisioning provider.
- Publish `azure.ai.agents` version `1.0.0-beta.6` without the provider registration.
- Update both extension entries in the registry atomically.
- Update the `microsoft.foundry` meta-package constraints to require the compatible pair.
- Verify installation and startup with both extensions together.

## Acceptance criteria

- Both signed extension artifacts are available before registry changes merge.
- A fresh installation resolves the compatible projects and agents versions.
- No duplicate `microsoft.foundry` provider registration occurs.
- Existing split and pre-split Foundry projects can provision, deploy, and tear down successfully.
- Release notes state that the pair must be upgraded together during this transition.

## Context

The implementation and coordinated version metadata are included in PR #9133. Registry and artifact publication must follow the release process rather than being performed in the implementation PR.

huimiu:
Related to #9134

## Why

Foundry project provisioning still lived in `azure.ai.agents`, even though `azure.ai.projects` owns project resources. Moving provisioning now gives each extension a clearer job. We are keeping init in agents temporarily so the handoff can be validated before moving the remaining setup flow.

## What changed

- Moved the `microsoft.foundry` provisioning provider from `azure.ai.agents` to `azure.ai.projects`.
- Moved split-project deployment handling to `azure.ai.projects`, while keeping the old agents path for backward compatibility.
- Kept `azd ai agent init` and `init --infra` in `azure.ai.agents` for now. Parity tests keep the temporary synthesis copies aligned.
- Added support for updating deployments and connections on existing Foundry projects. These projects must set `AZURE_AI_PROJECT_ID` to the full project ARM resource ID.
- Kept typed connection metadata while passing credentials separately as a secure ARM parameter.
- Used the customer VNet region when creating private endpoints.
- Blocked private-network projects from creating an ACR automatically, since the current templates do not create the ACR private networking they would need. Users should provide a pre-built image instead.
- Added projects unit tests to the agents CI workflow and installed the local projects dependency before running agents tests.

## Follow-up work

- Move the remaining `azd ai agent init` and `init --infra` logic to `azure.ai.projects` and remove the temporary agents-side copy.
- Publish the coordinated projects and agents releases, then update the extension registry and `microsoft.foundry` meta-package together.

## Release notes

The projects and agents releases must be published together. Mixing old and new versions can cause both extensions to register the same provider.


cli/azd/extensions/azure.ai.projects/internal/synthesis/synthesizer.go
```diff
@@ -0,0 +1,828 @@
...
+	node, ok := root.Services[in.ServiceName]
+	if !ok {
+		return nil, ErrServiceNotFound
+	}
+
+	// Resolve $ref file includes (service-entry-level and per-deployment) so the
+	// decoded service body carries the referenced content, not raw $ref objects.
+	if in.ProjectRoot != "" {
+		var err error
+		node, err = resolveServiceRefs(node, in.ProjectRoot, in.ServiceName)
```
jongio: `resolveServiceRefs` expands `$ref` only on the selected project node. The sibling scans just below (`deriveIncludeAcr` at line 255, `collectConnections` at line 262) still decode the raw `root.Services` nodes, so a sibling `azure.ai.agent` or `azure.ai.connection` authored with a service-level `$ref` is decoded without its referenced fields: an empty `image`/`codeConfiguration` flips `includeAcr` to true, and an empty `host` silently drops the connection. `pkg/foundry/includes.go` documents service-entry-level `$ref` as a supported shape, so this is reachable from a real azure.yaml. The only `$ref` test in this package exercises a deployment-item ref on the selected service, so nothing covers a sibling authored via `$ref`. Resolving refs across all services before the sibling scans, plus a test for a `$ref`-authored agent/connection sibling, would close the gap.

jongio (commented):
One test-coverage gap worth closing on the new synthesizer.

jongio (approved):
Re-reviewed the latest push. The sibling `$ref` resolution I flagged is now handled through `loadProjectService`, with `projectRoot` threaded into `deriveIncludeAcr` and `collectConnections`, so agents and connections declared via a service-level `$ref` are resolved before decoding. The gRPC project-config path resolves refs too, and `connections` now flows as a `@secure()` JSON parameter end to end. Tests cover both the ref-resolution and secure-parameter paths.

jongio (approved):
Re-reviewed the 3 new commits since my last approval. The connections parameter now uses the `value` envelope instead of `secureValue`, which is the valid ARM parameter key, so the value actually reaches the deployment. Security is still enforced by `@secure()` on the `connections` param in both main.bicep and brownfield.bicep, so nothing leaks in deployment history. Tests cover the new envelope, and the agents workflow now runs the projects unit tests.

cli/azd/extensions/azure.ai.agents/extension.yaml
```diff
@@ -5,27 +5,24 @@ displayName: Foundry agents (Beta)
 description: Ship agents with Microsoft Foundry from your terminal. (Beta)
 usage: azd ai agent <command> [options]
 # NOTE: Make sure version.txt is in sync with this version.
-version: 1.0.0-beta.5
```
trangevi: Why are we doing a version bump as a part of this PR?

cli/azd/extensions/azure.ai.agents/internal/project/provisioning_provider.go
```diff
@@ -7,7 +7,7 @@ import "slices"
 
 // FoundryProviderName is the value written to `infra.provider` in
 // azure.yaml by `azd ai agent init` and looked up by azd's provider
-// resolver to dispatch provisioning to this extension.
+// resolver to dispatch provisioning to azure.ai.projects.
```
trangevi, line 10: If provisioning is being handled by azure.ai.projects, what is this file for?

cli/azd/extensions/azure.ai.agents/internal/synthesis/templates/modules/resources.bicep
```diff
@@ -62,8 +49,9 @@ param deployments deploymentsType = []
 @description('Include an Azure Container Registry. Set true when any agent uses docker:.')
 param includeAcr bool = false
 
-@description('Foundry project connections to create (host: azure.ai.connection services).')
-param connections connectionsType = []
+@description('JSON-encoded Foundry project connections to create.')
+@secure()
+param connections string = ''
```
trangevi: Why is the handling here needing to change?

trangevi (changes requested):

jongio (commented):
Re-reviewed the 2 new commits since my last approval. The "defer Foundry extension release metadata" commit reverts `azure.ai.projects` to `1.0.0-beta.2` (version.txt and extension.yaml) but leaves the agents dependency at `azure.ai.projects: "~1.0.0-beta.3"`, so the two are now inconsistent and I can't keep my approval as-is.

Concretely this breaks `test-ext-azure-ai-agents.yml`. That workflow builds and installs projects from source at the version in `version.txt` (now beta.2), then runs `azd extension install azure.ai.agents --source local --force --no-prompt` with no `--no-dependencies`. The manager checks the already-installed projects (beta.2) against the agents constraint `~1.0.0-beta.3` and returns "installed dependency ... does not satisfy constraint" (`pkg/extensions/manager.go` around 597-607), so CI fails before any test runs. `~1.0.0-beta.3` resolves to `>=1.0.0-beta.3`, which beta.2 doesn't satisfy.

It isn't only a CI problem. Shipping the agents extension with a hard `~1.0.0-beta.3` dependency while projects is published at beta.2 means a real `azd extension install` can't resolve the dependency either. The cleaner fix is to keep the constraint in sync with the version this PR actually ships: if projects stays beta.2, set the agents dependency to `~1.0.0-beta.2`; if agents genuinely needs beta.3, bump projects to beta.3 rather than deferring it. Adding `--no-dependencies` in CI would only hide the mismatch for the pipeline while leaving registry installs broken.

jongio (approved):
The isolated Foundry validation looks correct. The guard rejects an auto-created ACR when private networking is requested, and it runs after both includeAcr and netMode are resolved. Since agentNeedsAcr returns false once an image is set, the `specify an image instead` guidance is accurate. Mirrored consistently across both extensions with matching test coverage.

cli/azd/extensions/azure.ai.agents/internal/project/provisioning_provider.go
huimiu: The filename is now misleading: this file no longer implements provisioning. It only holds the Foundry provider name and service-host contract still used by `azd ai agent init` / `init --infra` to write infra.provider and generate or eject IaC.

The provider that handles microsoft.foundry at provision time is now in azure.ai.projects. Following the staged migration https://github.com/Azure/azure-dev/issues/9085#issuecomment-4960314147, would it make sense to move the remaining agents-side init logic and remove this file in a follow-up PR?

cli/azd/extensions/azure.ai.agents/internal/synthesis/templates/modules/resources.bicep
huimiu: The previous typed connections parameter could include credentials without being marked secure, which risks exposing them in deployment history.

The related logic has since been removed from this PR, if we decide to reintroduce the functionality, we can address it in a follow-up PR.

huimiu:

jongio (commented):
Re-reviewed the two merge-conflict resolution commits. The `init_infra_test.go` resolution lines up with main's new `mcp-conn` fixture: connections are decoded from the `@secure()` JSON string param and asserted at length 2 (mcp-conn at index 0, search-conn at index 1), which is consistent.

These merge commits don't touch versioning, so the open blocker is still the version mismatch: `azure.ai.agents/extension.yaml` pins `azure.ai.projects` to `~1.0.0-beta.3`, while the projects `version.txt` and `extension.yaml` ship `1.0.0-beta.2`. In `test-ext-azure-ai-agents.yml`, projects is built and installed from source at beta.2, then the agents install fails the `~1.0.0-beta.3` constraint before any test runs, and registry installs hit the same mismatch. Sync the constraint with the shipped version: either set the agents dependency to `~1.0.0-beta.2`, or bump projects to beta.3.

cli/azd/extensions/azure.ai.projects/internal/synthesis/templates/modules/connections.bicep
```diff
@@ -0,0 +1,83 @@
...
+
+  @description('Optional metadata key-value pairs.')
+  metadata: object?
+}
+
+@description('Shape of a list of connections.')
+type connectionsType = connectionType[]
+
+@description('Connections to create on the Foundry project. Each entry maps to one host: azure.ai.connection service.')
+param connections connectionsType = []
```
jongio, line 45: This drops `@secure()` from the connections parameter, and `credentials` carries real secret material. `expandCredentials` resolves `${VAR}` env references into literal API keys at provision time, and that map goes straight into the `connections` param. Since it's now a typed array (Bicep can't apply `@secure()` to an array), the ARM parameter is plain, so credential values get written to the resource group deployment history in cleartext and are readable by anyone with Reader access. The prior `@secure() securestring` kept them out of history, and the deleted `TestBrownfieldARMTemplate_UsesSecureConnections` test asserted exactly that. The Terraform path still isolates credentials in `sensitive_body`, so only the Bicep/ARM path regressed. Same pattern is in `main.bicep`, `brownfield.bicep`, and the agents extension copies. One option: keep credentials in a `@secure()` object or securestring for the provision path, and expose the typed shape only for ejected templates.

jongio (commented):
Re-reviewed the new `fix: restore typed Foundry connection parameters` commit. Switching `connections` from a `@secure()` securestring to a typed array reintroduces secret exposure: connection `credentials` (API keys for ApiKey / CustomKeys) now flow through a non-secure ARM parameter and land in deployment history. Details inline on connections.bicep.

Separately, the version mismatch from my earlier reviews is still open: `azure.ai.agents/extension.yaml` pins `azure.ai.projects` to `~1.0.0-beta.3` while projects ships `1.0.0-beta.2`, so the agents install fails the constraint in `test-ext-azure-ai-agents.yml`, and registry installs hit the same mismatch.

jongio (approved):
Reviewed the template hardening commit. Credential values now ride in a secure object parameter, the private endpoint uses the customer VNet region, and the Terraform version constraint matches the optional-attribute usage. Changes are mirrored across both extensions with matching tests.

cli/azd/extensions/azure.ai.agents/internal/synthesis/templates/modules/resources.bicep
trangevi: Wait, does that mean that we can't create connections with credentials after this change?

cli/azd/extensions/azure.ai.agents/internal/project/provisioning_provider.go
trangevi: Yeah, we can have a follow up once we know how to handle init across extensions

cli/azd/extensions/azure.ai.agents/internal/cmd/listen.go
```diff
@@ -68,8 +54,11 @@ func configureExtensionHost(host *azdext.ExtensionHost) {
 }
 
 func preprovisionHandler(ctx context.Context, azdClient *azdext.AzdClient, args *azdext.ProjectEventArgs) error {
-	deployments, err := collectProjectDeployments(args.Project.Services)
-	if err != nil {
+	if err := updateLegacyProjectDeployments(
```
trangevi, line 57: What are "legacy projects"?

cli/azd/extensions/azure.ai.agents/internal/synthesis/templates/modules/resources.bicep
jongio: No, credentialed connections still work. This change only moves the credential payload out of the inline `connections` array and into a separate `@secure() connectionCredentials` object keyed by connection name.

`SplitConnectionCredentials` in the synthesizer pulls each connection's credentials into `connectionCredentials[name]` and nils the inline field, and connections.bicep re-attaches them:

`union(..., contains(connectionCredentials, c.name) ? { credentials: connectionCredentials[c.name] } : {}, ...)`

So any connection that has credentials still gets the same `credentials` property on the `accounts/projects/connections` resource. The only difference is the value now travels through a secure parameter, so it no longer lands in deployment history. Connections without credentials just skip the property, same as the old `c.?credentials != null` guard. Terraform gets them back via `JoinConnectionCredentials` into `sensitive_body`.

trangevi (approved):

cli/azd/extensions/azure.ai.agents/internal/cmd/listen.go
huimiu: By “legacy projects,” I mean existing azure.yaml projects using the older `azure.ai.agent` or `microsoft.foundry` service shapes instead of `azure.ai.project`. We’re keeping this path for backward compatibility while provisioning moves to `azure.ai.projects`, and we recommend using `azure.ai.project` for new projects.

trangevi (approved):
Reapproving after recent commit

cli/azd/extensions/azure.ai.projects/internal/synthesis/templates/terraform/main.tf
```diff
@@ -0,0 +1,140 @@
...
+        virtualNetworkRules = []
+        ipRules             = []
+      }
+    }
+  }
+}
+
+# Created one at a time; ARM throttles concurrent deployments on one account.
+resource "azurerm_cognitive_deployment" "model" {
+  for_each = { for d in var.deployments : d.name => d }
```
jongio, line 90: This `for_each` won't serialize the deployments the way the comment above claims. Terraform creates `for_each` instances concurrently (up to `-parallelism`, default 10), so all model deployments get submitted to the account at once. There's no `depends_on` chain or other ordering between them here. The Bicep path handles the same constraint deliberately with `@batchSize(1)` in resources.bicep (Sequential model deployment creation; ARM throttles concurrent deployments on the same account). Without an equivalent here, multi-model provisioning can hit the ARM concurrent-deployment throttle and fail intermittently. Consider chaining the deployments so only one is in flight at a time, or documenting a required `-parallelism=1` for this template.

jongio (commented):
One correctness concern on the Terraform provisioning path, otherwise the incremental change since my last look is just a docs wording tweak.

huimiu:

jongio (commented):
Caught up since my last pass. The origin/main merge in c1f48e7 and the telemetry doc reword are clean, no new issues there.

Still open from before: the Terraform `for_each` on `azurerm_cognitive_deployment.model` in `azure.ai.projects/internal/synthesis/templates/terraform/main.tf`. The comment says the deployments are created one at a time because ARM throttles concurrent deployments on a single account, but `for_each` does not serialize them. Terraform schedules the instances concurrently (up to `-parallelism`, default 10), so under real throttling this can intermittently hit 429s. Either the comment is inaccurate or the serialization is missing. If one-at-a-time is genuinely required, chain the deployments with explicit ordering rather than relying on the comment. Line-level detail is on the existing thread.

vhvb1989 (approved):

jongio (approved):
Re-reviewed the current HEAD after the merge of origin/main and the telemetry docs wording fix. The merge-conflict resolution keeps Foundry provisioning ownership in `azure.ai.projects` and aligns the transferred validation check with main's `arm-provision` terminology, which is consistent with the design. The merged main changes (the preflight to provision validation rename and the telemetry field renames) don't change this PR's behavior. No new issues in the incremental changes.

One mechanical heads up: the branch shows as conflicting again since main advanced past the merge commit, so it'll need another merge from main before it can go in.

huimiu:

jongio (approved):
Re-approving after the latest sync. The only delta since my last approval on c1f48e7c is a merge from main that pulls in upstream doc, test, registry, and telemetry changes. None of this PR's own provisioning code (synthesizer, connections template, terraform module) changed in that merge, so my earlier review still holds.

hund030 (approved):

PR #9141, opened 2026-07-14
fix: send x-ms-correlation-request-id as a canonical hyphenated GUID

### LINKED ISSUE #5851: [Issue] AKS deployment fails with GetDeploymentSafeguardsFailed: Correlation ID format mismatch

**Output from `azd version`**
```
azd version 1.19.0 (commit b3d68cea969b2bfbaa7b7fa289424428edb93e97)
```

**Describe the bug**

`azd up` consistently fails when deploying an AKS cluster with the error `GetDeploymentSafeguardsFailed` due to a Correlation ID format mismatch. The error occurs during AKS Deployment Safeguards validation when Azure compares the Correlation ID from the On-Behalf-Of token with the original request's Correlation ID.

**Observed behavior:**
- azd sets Correlation ID in 32-character format without hyphens (e.g., `a161af237f2cd19acc115e06190a47f0`)
- The error message shows Azure has converted this to a format with hyphens (e.g., `a161af23-7f2c-d19a-cc11-5e06190a47f0`)
- AKS Deployment Safeguards validation fails with a string comparison mismatch

**Error Message:**
```
GetDeploymentSafeguardsFailed: The on behalf of token's correlation ID 'a161af237f2cd19acc115e06190a47f0' does not match with current correlation ID 'a161af23-7f2c-d19a-cc11-5e06190a47f0'.
```

**To Reproduce**

**Repository with reproduction code:**
https://github.com/torumakabe/aks-chaos-lab

1. Clone the repository:
```bash
git clone https://github.com/torumakabe/aks-chaos-lab.git
cd aks-chaos-lab
```

2. Configure the AKS SKU mode to Automatic:
   - Edit `infra/main.parameters.json`
   - Set `aksSkuName` parameter to `"Automatic"`:
   ```json
   "aksSkuName": {
     "value": "Automatic"
   }
   ```
   - Reference: https://github.com/torumakabe/aks-chaos-lab/blob/main/infra/main.parameters.json

3. The repository contains:
   - Bicep templates for AKS cluster deployment (`infra/` directory)
   - `azure.yaml` configuration for azd
   - AKS cluster configuration with:
     - Automatic SKU mode (when configured as above)
     - Container Insights enabled
     - Azure Monitor workspace integration
     - Prometheus monitoring

4. Run `azd up --debug`

5. Observe the deployment fails during AKS cluster creation with `GetDeploymentSafeguardsFailed` error

**Reproduction Rate:**
- Tested twice: **2/2 failures (100% reproducible)**
- Both executions failed with identical error at AKS Deployment Safeguards validation

**Expected behavior**

- `azd up` should successfully deploy the AKS cluster
- Correlation ID format should be compatible with Azure's internal processing to avoid validation errors
- No Deployment Safeguards validation errors should occur

**Comparison with Azure CLI:**

Direct deployment using `az deployment sub create` with the same Bicep template **succeeds without any errors**, which suggests:
1. The Bicep template is correct
2. The issue appears to be specific to how azd CLI formats Correlation IDs
3. Azure CLI's Correlation ID format is compatible with the internal validation

**Deployment Method Comparison:**

| Method | Result | Correlation ID Format |
|--------|--------|---------------------|
| `azd up` (1st run) | ❌ Failed | 32 chars, no hyphens |
| `azd up` (2nd run) | ❌ Failed | 32 chars, no hyphens |
| `az deployment sub create` | ✅ Success | (format not verified, but compatible) |

**Environment**

- **OS:** Linux (WSL2) - `6.6.87.2-microsoft-standard-WSL2`
- **azd version:** 1.19.0 (commit b3d68cea969b2bfbaa7b7fa289424428edb93e97)
- **Azure CLI version:** 2.77.0
- **Bicep version:** 0.37.4
- **Language:** Python (for service), Bicep (for infrastructure)
- **Azure Subscription Region:** Japan East

**Additional context**

### Technical Details

**Observed Correlation ID Format Difference:**
- **azd format:** `xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx` (32 characters, no hyphens)
- **Format in error message:** `xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx` (36 characters, with hyphens)
- **Observed conversion pattern:** `a161af237f2cd19acc115e06190a47f0` → `a161af23-7f2c-d19a-cc11-5e06190a47f0`

**Note:** The exact format requirements of Azure ARM API are not documented in public documentation. The observation is based on the error message comparison.

**Error Call Stack:**
```
Subscription-level deployment (dev-*)
  └─ Resource group deployment (aks)
       └─ AKS cluster creation (aks-aks-chaos-lab-dev)
            └─ Deployment Safeguards validation ← ERROR HERE
```

**Important Notes:**
1. **Deployment Safeguards is NOT defined in Bicep** - it's automatically invoked by AKS service during cluster creation/update
2. The error occurs in Azure's internal service-to-service communication, not in user-defined code
3. azd sets the same Correlation ID header (`X-Ms-Correlation-Request-Id`) for all ARM API requests (verified in debug logs - used 1113 times in a single deployment)

### Debug Logs

Debug logs showing the issue are available:
- 1st run: TraceID `08cdeb0281c1ce2c5ace30831bf937b0`
- 2nd run: TraceID `a161af237f2cd19acc115e06190a47f0`

Both runs show consistent pattern:
- azd sends requests with 32-character Correlation ID (no hyphens)
- Error message shows Azure has a different format (with hyphens)
- Deployment Safeguards validation fails on comparison

### Workaround

Use Azure CLI directly instead of azd:
```bash
az deployment sub create \
  --location japaneast \
  --template-file infra/main.bicep \
  --parameters @infra/main.parameters.json
```

This successfully deploys all resources including the AKS cluster without any Deployment Safeguards errors.

hemarina:
Fix #5851

## Summary

`azd` sets the `x-ms-correlation-request-id` header on every ARM request from the ambient OpenTelemetry trace ID via `spanCtx.TraceID().String()`, which produces an **undecorated 32-character hex string** (no hyphens), e.g. `a161af237f2cd19acc115e06190a47f0`. Every other Azure tool sends this correlation value as a hyphenated GUID.

> [!NOTE]
> **This is no longer a blocking issue — it is a deliberate design choice.**
>
> The problem originally surfaced as a hard failure (#5851): AKS Deployment Safeguards parsed the header as a GUID, reformatted it with hyphens, and string-compared it against azd's undecorated value, failing `azd up` / `azd provision` for AKS clusters with Deployment Safeguards enabled (including AKS Automatic):
> ```
> GetDeploymentSafeguardsFailed: The on behalf of token's correlation ID 'a161af237f2cd19acc115e06190a47f0' does not match with current correlation ID 'a161af23-7f2c-d19a-cc11-5e06190a47f0'.
> ```
> The AKS service side has since relaxed this check and now **accepts the non-hyphenated ID**, so the original deployment failure no longer reproduces. This PR is therefore not an urgent bug fix — we are making it because azd's correlation ID format diverges from the industry standard and from the rest of the Azure tooling, and we want to bring it into line for **consistency, spec-correctness, and future-proofing** (nothing guarantees other resource providers won't apply the same strict GUID parsing AKS originally did).

## Change

Format the trace ID as a canonical hyphenated GUID. A W3C trace ID and a UUID are both 128-bit values, so `uuid.UUID(spanCtx.TraceID()).String()` (reusing the already-imported `github.com/google/uuid`) reshapes it into the RFC 9562 canonical lowercase `8-4-4-4-12` form. The conversion is **lossless and reversible** (stripping the hyphens recovers the original trace ID), so end-to-end OpenTelemetry correlation is preserved.

This aligns azd with the ARM RPC spec (which asks for "a GUID with no decoration such as curly braces") and with the industry standard (RFC 9562: lowercase, hyphenated `8-4-4-4-12`, no braces) followed by every other Azure tool:

| Tool | Header | Format |
|---|---|---|
| Terraform AzureRM | `x-ms-correlation-request-id` | lowercase hyphenated GUID (`hashicorp/go-uuid`) |
| Azure SDK for Go (azcore) | `x-ms-client-request-id` | lowercase hyphenated GUID (`sdk/internal/uuid`) |
| **azd (before)** | `x-ms-correlation-request-id` | 32 chars, no hyphens |
| **azd (after)** | `x-ms-correlation-request-id` | lowercase hyphenated GUID |

The change is in the shared correlation policy (`NewMsCorrelationPolicy`), so it applies to **all** ARM requests, not just AKS. AKS was simply the service strict enough to reject the old format; other services silently tolerated it.

## Testing

- Updated `Test_NewMsCorrelationPolicy` to expect the hyphenated GUID.
- Added `Test_NewMsCorrelationPolicy_FormatsTraceIdAsGuid` asserting the emitted value is a valid 36-char GUID and that stripping the hyphens losslessly recovers the original trace ID.
- `go build`, `go test ./pkg/azsdk/...`, `golangci-lint run ./pkg/azsdk/...`, `gofmt`, and `cspell` all pass.

### DIFF
--- a/cli/azd/CHANGELOG.md
+++ b/cli/azd/CHANGELOG.md
@@ -14,7 +14,7 @@
 
 ### Other Changes
 
-- [[#9141]](https://github.com/Azure/azure-dev/pull/9141) Send the `x-ms-correlation-request-id` header as a canonical hyphenated GUID (derived losslessly from the OpenTelemetry trace ID) instead of an undecorated 32-character string, aligning azd with the ARM spec and other Azure tooling (Terraform AzureRM, Azure SDK for Go). This also resolves the historical AKS Deployment Safeguards `GetDeploymentSafeguardsFailed` correlation ID mismatch (#5851).
+- [[#9141]](https://github.com/Azure/azure-dev/pull/9141) Send ARM request correlation IDs as a canonical hyphenated GUID (derived losslessly from the OpenTelemetry trace ID) instead of an undecorated 32-character string — covering both the `x-ms-correlation-request-id` header on azd's direct ARM calls and the `ARM_CORRELATION_REQUEST_ID` value passed to the Terraform AzureRM provider. This aligns azd with the ARM spec and other Azure tooling (Terraform AzureRM, Azure SDK for Go) and resolves the historical AKS Deployment Safeguards `GetDeploymentSafeguardsFailed` correlation ID mismatch (#5851).
 
 ## 1.27.1 (2026-07-09)
 
--- a/cli/azd/CHANGELOG.md
+++ b/cli/azd/CHANGELOG.md
@@ -14,6 +14,8 @@
 
 ### Other Changes
 
+- [[#9141]](https://github.com/Azure/azure-dev/pull/9141) Send the `x-ms-correlation-request-id` header as a canonical hyphenated GUID (derived losslessly from the OpenTelemetry trace ID) instead of an undecorated 32-character string, aligning azd with the ARM spec and other Azure tooling (Terraform AzureRM, Azure SDK for Go). This also resolves the historical AKS Deployment Safeguards `GetDeploymentSafeguardsFailed` correlation ID mismatch (#5851).
+
 ## 1.27.1 (2026-07-09)
 
 ### Features Added
--- a/cli/azd/pkg/azsdk/correlation_policy.go
+++ b/cli/azd/pkg/azsdk/correlation_policy.go
@@ -33,13 +33,21 @@ func (p *traceCorrelationPolicy) Do(req *policy.Request) (*http.Response, error)
 		return req.Next()
 	}
 
-	// A W3C trace ID and a UUID are both 128-bit values, so we reformat the trace ID into the canonical
-	// hyphenated GUID form (8-4-4-4-12) that the ARM spec expects. This keeps the exact trace ID value
-	// (the conversion is reversible by removing the hyphens) while matching the format other Azure tools send.
-	// Some services (e.g. AKS Deployment Safeguards) parse this header as a GUID and fail a string comparison
-	// against azd's previously undecorated 32-character value. See azure-dev#5851.
-	rawRequest.Header.Set(p.headerName, uuid.UUID(spanCtx.TraceID()).String())
+	rawRequest.Header.Set(p.headerName, CorrelationIDFromTraceID(spanCtx.TraceID()))
 	return req.Next()
+}
+
+// CorrelationIDFromTraceID formats an OpenTelemetry trace ID as the canonical hyphenated GUID (8-4-4-4-12) that
+// ARM expects for correlation values — both the `x-ms-correlation-request-id` header and the
+// `ARM_CORRELATION_REQUEST_ID` variable azd passes to the Terraform AzureRM provider. A W3C trace ID and a UUID
+// are both 128-bit values, so the conversion is lossless and reversible: removing the hyphens recovers the
+// original trace ID, preserving end-to-end OpenTelemetry correlation.
+//
+// azd historically emitted the undecorated 32-character trace ID here, which AKS Deployment Safeguards parsed as
+// a GUID and string-compared against the original, causing GetDeploymentSafeguardsFailed (azure-dev#5851).
+// Emitting the canonical GUID also matches the format other Azure tools send (Terraform AzureRM, Azure SDK for Go).
+func CorrelationIDFromTraceID(traceID trace.TraceID) string {
+	return uuid.UUID(traceID).String()
 }
 
 // perRequestUUIDPolicy sets a header to a freshly generated UUID on every outgoing HTTP request. It is used for
--- a/cli/azd/pkg/azsdk/correlation_policy.go
+++ b/cli/azd/pkg/azsdk/correlation_policy.go
@@ -33,7 +33,12 @@ func (p *traceCorrelationPolicy) Do(req *policy.Request) (*http.Response, error)
 		return req.Next()
 	}
 
-	rawRequest.Header.Set(p.headerName, spanCtx.TraceID().String())
+	// A W3C trace ID and a UUID are both 128-bit values, so we reformat the trace ID into the canonical
+	// hyphenated GUID form (8-4-4-4-12) that the ARM spec expects. This keeps the exact trace ID value
+	// (the conversion is reversible by removing the hyphens) while matching the format other Azure tools send.
+	// Some services (e.g. AKS Deployment Safeguards) parse this header as a GUID and fail a string comparison
+	// against azd's previously undecorated 32-character value. See azure-dev#5851.
+	rawRequest.Header.Set(p.headerName, uuid.UUID(spanCtx.TraceID()).String())
 	return req.Next()
 }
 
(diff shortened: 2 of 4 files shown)


jongio (approved):
Verified the trace ID to GUID reformat is lossless: `uuid.UUID` and `trace.TraceID` are both `[16]byte`, so `uuid.UUID(spanCtx.TraceID())` copies the bytes without touching version/variant bits, and stripping the hyphens recovers the original 32-character trace ID. The added test locks in that round-trip, and the emitted value now matches the hyphenated GUID format other Azure tools send. Scoped to the shared correlation policy so all ARM requests benefit, and the ACR per-call override is untouched.

vhvb1989 (approved):

JeffreyCA (approved):
Looks good -

Are there any downstream consumers we need to update as a result of this change (e.g. if they assume or rely on the no-dash format)?

jongio (commented):
On the downstream-consumer question: the Terraform provisioning path isn't covered by this change. `pkg/infra/provisioning/terraform/terraform_provider.go:116` still exports the raw 32-character trace ID to the Terraform subprocess:

```go
envVars = append(envVars, fmt.Sprintf("ARM_CORRELATION_REQUEST_ID=%s", spanCtx.TraceID().String()))
```

`ARM_CORRELATION_REQUEST_ID` is consumed by the AzureRM provider for request correlation, so Terraform-driven ARM requests still send the undecorated form while azd's direct calls now send the hyphenated GUID. Since AKS provisioning can run through Terraform, that path is exactly where the original mismatch would show up. Worth either reformatting there too with `uuid.UUID(spanCtx.TraceID()).String()`, or confirming AzureRM normalizes the value on its side. Not blocking, since it's the same trace ID either way.

hemarina:
Thanks @jongio, @JeffreyCA, @vhvb1989 for the reviews. Addressed the feedback in 4bf103f4e:

1. **Terraform path (@jongio / code review):** Extracted a shared helper `azsdk.CorrelationIDFromTraceID(traceID)` and used it for **both** the `x-ms-correlation-request-id` header and the `ARM_CORRELATION_REQUEST_ID` env var passed to the AzureRM provider in `terraform_provider.go`. Terraform-driven ARM requests now send the same canonical hyphenated GUID, so the "all ARM requests" scope is accurate. This also answers @JeffreyCA's downstream-consumer question — the Terraform provisioning path was the remaining consumer of the raw format.
2. **Historical phrasing (code review):** Reworded the source comment in `correlation_policy.go` and the test doc so the AKS Deployment Safeguards failure is described historically, consistent with the PR note that the service now accepts the undecorated value.
3. **Coverage:** Added a direct unit test for the shared helper (`Test_CorrelationIDFromTraceID`) and broadened the changelog entry to mention the Terraform path.

`go build`, `go test ./pkg/azsdk/...`, `golangci-lint`, `gofmt`, and `cspell` all pass.

jongio (approved):
Incremental change since my last look addresses the Terraform path. `CorrelationIDFromTraceID` is now the single source of truth for the trace-ID-to-GUID reformat, and `terraform_provider.go` uses it for `ARM_CORRELATION_REQUEST_ID`, so Terraform-driven ARM requests send the same canonical hyphenated GUID as azd's direct calls. Verified: the terraform package importing `azsdk` introduces no cycle (azsdk stays a leaf), and the new `Test_CorrelationIDFromTraceID` runs against the real non-zero trace ID from `init()`, locking in the lossless round-trip. Comment and changelog rewording read accurately now.

PR #9143, opened 2026-07-14
Fix azd down --no-prompt hang in CI for Terraform (#4317)

### LINKED ISSUE #4317: [Issue] Azd down --no-prompt hangs in pipeline and does not initialise terraform

**Output from `azd version`**

azd version 1.10.1 (commit 31409a33266fb4a5fdbb644bc83988e725d6c7c9)

**Describe the bug**

I am trying to setup a pipeline to tear down environments. I use Terraform as the backend. I want to be able to run `azd down --preview` but that option isn't available, however it looks like `azd down` prompts for confirmation by default, so it looks like I should be able to get a preview of the Terraform destroy by running `azd down --no-prompt` which i'd expect it to show the Terraform destroy plan and then stop execution because the default confirmation would be no to continue, however the pipeline task just hangs. The only way to get destroy to run in the pipeline is to use ``azd down --no-prompt --force`.

Additionally `azd down` does not perform a `terraform init`, so running it on its own in a pipeline just throws an error "backend initialization required". I can workaround this by running `azd provision --preview` first but it would be nice not to have to.

**To Reproduce**

Setup an Azure DevOps pipeline to to run `azd down --no-prompt`. The task will hang. 

**Expected behavior**

`azd down --no-prompt` = Should output the Terraform destroy plan and then the task should exit with no error, as if the confirmation prompt was answered with "no".

Should also perform a terraform init.

**Environment**

Azure DevOps pipeline self hosted agent running unbuntu.
Azd tasks run as Bash.

**Additional context**

hemarina:
Fixes #4317

## Problem
`azd down --no-prompt` hangs in CI/CD pipelines for Terraform-based projects. `azd` delegates the destroy confirmation to Terraform's own interactive prompt (`terraform destroy` without `-auto-approve`). In a non-interactive context there is no TTY to answer it, so Terraform blocks forever on stdin. The only workaround was `--force`. A related symptom (#4317) is `azd down --force` failing on a fresh agent with "backend initialization required".

## Fix
When running non-interactively — detected as either a CI/CD environment (`resource.IsRunningOnCI()`) **or** `--no-prompt` mode — the Terraform provider now:
- runs `terraform init -input=false` for every automated destroy path (preview **and** `--force`), so a fresh agent no longer fails with "backend initialization required",
- and, when `--force` is not passed, shows a read-only destroy preview via `terraform plan -destroy -input=false` (nothing is deleted), prints a warning scoped to the Terraform configuration that no resources were deleted and to re-run with `--force`, and returns cleanly (exit 0) instead of hanging.

All other paths are unchanged:
- Interactive terminal (not CI, not `--no-prompt`): Terraform's normal confirmation prompt (user chooses); azd does not force init or preview.
- `--force`: `terraform destroy -auto-approve` (deletes).

For a multi-layer `down`, the environment cache is always invalidated (another layer may have deleted resources even if a Terraform layer was preview-only), and the full-teardown success message is suppressed whenever any layer was skipped.

## Behavior
| Environment | `--force` | Behavior |
|---|---|---|
| CI **or** `--no-prompt` | no | init + destroy preview only, exit 0 (was: hang) |
| CI **or** `--no-prompt` | yes | init + destroy (deletes) |
| interactive terminal | no | Terraform confirmation prompt (unchanged) |
| any | yes | destroy (deletes) |

## Testing
- New unit tests: CI preview (asserts `plan -destroy` ran, no delete), non-CI `--no-prompt` preview (asserts `plan -destroy` ran, no delete), CI + `--force` (asserts init runs, deletes), interactive (no forced init/preview, deletes). Pre-existing `TestTerraformDestroy` now clears CI env so it does not inherit runner state.
- `PlanDestroy` added to the terraform command-wrapper table test (asserts exact args incl. `-destroy` and `-input=false`).
- Action-level `down` tests assert the full-teardown header is suppressed on `SkippedDeletion` and that the cache is still invalidated.
- `go build ./...`, `gofmt -s`, `golangci-lint`, and existing `down`/terraform tests all pass.


jongio (commented):
The preview path fixes the hang for `--no-prompt` without `--force`. One gap on the `--force` side before this lands, noted inline.

cli/azd/pkg/infra/provisioning/terraform/terraform_provider.go
```diff
@@ -253,6 +255,32 @@ func (t *TerraformProvider) Destroy(
...
+	// terraform destroy asks for confirmation interactively unless --force is passed.
+	// In a CI/CD pipeline with --no-prompt, azd cannot answer that prompt, so terraform would block
+	// reading stdin and hang. In that case, preview the destroy plan (read-only) and stop without
+	// deleting, as if the confirmation were answered "no". Use --force to delete in CI. See issue #4317.
+	if resource.IsRunningOnCI() && !options.Force() && t.console.IsNoPromptMode() {
+		t.console.Message(ctx, "Previewing resources to delete...")
+		// terraform doesn't use `t.console`; stop any spinner before streaming its output.
+		t.console.StopSpinner(ctx, "", input.Step)
+
+		if initRes, err := t.init(ctx, isRemoteBackendConfig); err != nil {
```
jongio: You run `terraform init` here for the preview path, but the `--force` branch below never does. In a fresh CI checkout `createOutputParameters` calls `terraform output`, which needs an initialized backend, so `azd down --no-prompt --force` would still fail with `backend initialization required` (the other symptom reported in #4317) before it ever reaches `terraform destroy`. Should init run for the force/CI path too, or is prior init guaranteed there?

jongio (approved):
Init now runs non-interactively before `createOutputParameters` on both the preview and --force CI paths, which closes the `backend initialization required` gap I raised. The clearCIEnv helper unsets every var IsRunningOnCI checks, so the non-CI test stays deterministic in real pipelines.

jongio (commented):
The two new tests exercise the down action's SkippedDeletion handling: the skipped case asserts no teardown header, the normal case asserts the header is present. Both match down.go's behavior of returning an empty ActionResult when a layer was skipped.

One open item on the init gating, in the area I flagged earlier. init is keyed on `automatedCI := resource.IsRunningOnCI() && t.console.IsNoPromptMode()`. `--force` and `--no-prompt` are independent flags (down's `forceDelete` doesn't set console no-prompt), so `azd down --force` in CI without `--no-prompt` skips the init call and falls straight into `createOutputParameters`/`terraform output`. On a fresh agent with a remote backend, that's the same "backend initialization required" failure this PR is trying to avoid. My earlier approval note said init covers the --force CI path, but it only does so when --no-prompt is also set.

Is scoping the preview and init to no-prompt-only intentional? If `azd down --force` is still a documented CI path, keying init on `IsRunningOnCI()` alone, or on CI plus (no-prompt or force), would close that gap. Not blocking; the no-prompt hang fix itself is sound.

JeffreyCA (approved):
One gap (non-blocking): `init` is gated on `IsRunningOnCI() && IsNoPromptMode()`, but `--force` doesn't set no-prompt mode, so `azd down --force` in CI skips `init` and hits the same "backend initialization required" error from #4317.

Gating `init` on `IsRunningOnCI()` alone (keeping preview/skip on no-prompt-without-force) would close that path. A `--force`-in-CI test would be nice too.

jongio (approved):
The unified `automated := IsRunningOnCI() || IsNoPromptMode()` gate closes the init gap I raised earlier: `azd down --force` in CI now runs `terraform init -input=false` before `createOutputParameters`, so the `backend initialization required` symptom from #4317 is covered on the force path too. Folding `--no-prompt` outside CI into the same preview-and-skip path is consistent, since `--no-prompt` never answered terraform's own confirmation before. Tests cover the CI-preview, CI-force, non-CI `--no-prompt`, and interactive paths.

jongio (approved):
Incremental changes since my last review are test-only and address the earlier feedback. TestTerraformDestroy now clears CI env so the interactive path isn't skipped under a CI runner, and both preview tests now assert terraform plan -destroy actually runs rather than just checking SkippedDeletion. Mock registration order (LIFO after preparePlanningMocks) and the matcher are correct.

jongio (approved):
The `address feedback` commit moves `-upgrade` out of `Cli.Init` into caller control. The plan path passes `-upgrade`, so provisioning still refreshes providers/modules and the lock file as before; the automated destroy init passes only `-input=false`, so teardown no longer rewrites `.terraform.lock.hcl` or selects newer providers. Both production `t.init` call sites are updated, and the two new `NotContains("-upgrade")` assertions plus the `terraform_test` command cases lock the split in place. Help text and the down usage snapshot match the preview-without-force behavior. Build and the terraform tool/provider tests pass locally.

JeffreyCA (approved):

PR #9144, opened 2026-07-14
Add CFSClean enforcement

### LINKED ISSUE #9221: Enforce CFSClean
Enable CFSClean enforcement

chidozieononiwu:
This pull request makes a small change to the pipeline configuration by updating the `networkIsolationPolicy` setting. The policy is now set to include both `Permissive` and `CFSClean`, which may affect how network isolation is enforced during pipeline runs. 

* Pipeline configuration:
  * Updated the `networkIsolationPolicy` in `eng/pipelines/templates/stages/1es-redirect.yml` to include `CFSClean` in addition to `Permissive`.
  
[azure-dev - ext - microsoft.azd.demo](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=6572861&view=results)
[azure-dev - cli](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=6636943&view=results)
[azure-dev - vscode](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=6637293&view=results)
[azure-dev - azdo-extension](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=6635492&view=results)
[azure-dev - ext - azure.appservice](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=6635572&view=results)
[azure-dev - ext - azure.ai.connections](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=6635590&view=results)
[azure-dev - ext - microsoft.azd.ai.builder](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=6635604&view=results)

Resolves https://github.com/Azure/azure-dev/issues/9221


jongio (approved):
Verified the multi-value syntax against 1ES usage across the org - `Permissive, CFSClean` is the standard comma-separated policy format. The prior golang internalModuleProxy setup lines up with enforcing CFSClean, so the network isolation tightening has the internal mirror in place. Scoped and correct.

jongio (approved):
Re-checked after the demo extension version bump to 0.7.2. version.txt, extension.yaml, and the CHANGELOG entry all line up, and it's a no-op test release to exercise the CFSClean pipeline change. Still good.

raych1 (approved):

vhvb1989 (approved):

jongio (commented):
The two new templates added under `eng/common/pipelines/templates/steps/` (`maven-authenticate.yml` and `pypi-auth-dev-feed.yml`) sit in the directory that is mirrored from `azure-sdk-tools`. `eng/common/README.md` warns that changes here get overwritten by the common-engineering sync.

I checked `azure-sdk-tools/eng/common/pipelines/templates/steps` and neither file exists upstream (the closest is a differently-named `auth-dev-feed.yml`). So the next sync will delete both files, and the callers that reference them, `eng/pipelines/templates/jobs/build-cli.yml` and `eng/pipelines/templates/stages/build-and-test.yml`, will then fail with a missing-template error.

Move both templates to `eng/pipelines/templates/steps/` where the other repo-specific steps live (e.g. `install-ms-openjdk.yml`, `setup-go.yml`) and update the two `template:` references, or land them upstream in `azure-sdk-tools` first and consume them through the normal common-engineering update.

`nuget.config` and `eng/settings.xml` at the repo root look fine.

jongio (commented):
Re-reviewed the latest commit. Dropping the root nuget.config and scoping the restore with feedsToUse: select and vstsFeed: 'public/azure-sdk-for-net' in build-msi.yml keeps the windows.sln restore (the only NuGet restore in the repo) on the ADO mirror under CFSClean, so that change is self-contained and introduces no new issues.

One blocker from the earlier pass is still open: maven-authenticate.yml and pypi-auth-dev-feed.yml live under eng/common, which is overwritten by the azure-sdk-tools sync (see eng/common/README.md:3), and neither file exists upstream in azure-sdk-tools. The next common-engineering sync will delete both and break the template references in build-cli.yml and build-and-test.yml. Move them under eng/pipelines/templates/steps/ (next to setup-go.yml and install-ms-openjdk.yml), or land them upstream in azure-sdk-tools first, before merging.

jongio (approved):
The only change since my last look is the repo-root `NuGet.Config`, which clears inherited feeds and pins restores to the azure-sdk-for-net feed (nuget.org upstream). That matches the CFSClean goal. It merges cleanly with the existing `cli/installer/windows-actions/nuget.config` (repositoryPath only), and the Windows MSI restore is the only .NET restore affected. CI is green.

jongio (approved):
Re-reviewed the two new commits since my last pass. build-cli.yml now pulls in create-authenticated-npmrc.yml, and that template picked up defaults for npmrcPath (empty resolves to $HOME/.npmrc) and registryUrl (the azure-sdk-for-js feed), with the resolved path carried into npmAuthenticate through the resolvedNpmrcPath variable. I diffed it against azure-sdk-tools eng/common and it matches upstream exactly, so the common-engineering sync won't revert it and the earlier no-defaults invocation error on build-cli.yml is cleared. This is the upstream-first path.

The same treatment still needs to happen for maven-authenticate.yml and pypi-auth-dev-feed.yml (neither exists upstream in azure-sdk-tools yet), but that's the standing item from my earlier reviews, not something these two commits introduced. The npmrc change itself is scoped and correct.

danieljurek:
Internal pipelines that need to pass for this to work: 
* azure-dev - cli 
* azure-dev - vscode
* azure-dev - ext - microsoft.azd.demo

Skip release stages. If there are problems in those we can deal with them at release time. 


azd's tests can use different toolchains (.NET, JS, etc.) so the package managers for those need to be configured properly to look at their respective authenticated feeds.

eng/common/pipelines/templates/steps/pypi-auth-dev-feed.yml
```diff
@@ -0,0 +1,49 @@
+parameters:
```
jongio: This file doesn't exist upstream in `azure-sdk-tools`, and `eng/common` gets overwritten by the common-engineering sync (see `eng/common/README.md`). The next sync will delete this file and break the `- template: /eng/common/pipelines/templates/steps/pypi-auth-dev-feed.yml` reference in `build-and-test.yml`. `maven-authenticate.yml` was handled by landing it upstream first, do the same here, or move this under `eng/pipelines/templates/steps/` and update the reference.

jongio (commented):
One item from earlier passes is still open, now narrowed to a single file.

`maven-authenticate.yml` has since landed in `azure-sdk-tools` upstream, so that half is resolved. `pypi-auth-dev-feed.yml` still doesn't exist upstream (the closest match is the differently-named `auth-dev-feed.yml`). Since `eng/common` is overwritten by the common-engineering sync (`eng/common/README.md`), the next sync will delete `pypi-auth-dev-feed.yml` and break the `template:` reference to it in `build-and-test.yml`. Land it upstream in `azure-sdk-tools` first (as was done for maven), or move it under `eng/pipelines/templates/steps/` and update the reference.

eng/pipelines/templates/jobs/build-cli.yml
```diff
@@ -101,9 +106,21 @@ jobs:
           version: '9.x'
         displayName: Set up .NET 9
 
+      - bash: dotnet nuget add source --name azure-sdk-for-net https://pkgs.dev.azure.com/azure-sdk/public/_packaging/azure-sdk-for-net/nuget/v3/index.json
+        displayName: Add Azure SDK dev nuget feed
+
       - bash: dotnet nuget add source --name dotnet9 https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet9/nuget/v3/index.json
         displayName: Add internal dotnet nuget feed
 
+      - template: /eng/common/pipelines/templates/steps/pypi-auth-dev-feed.yml
```
jongio: Same upstream-sync issue I flagged on `build-and-test.yml`: this new reference points at `eng/common/pipelines/templates/steps/pypi-auth-dev-feed.yml`, which doesn't exist upstream in `azure-sdk-tools`. `eng/common` is overwritten by the common-engineering sync, so the next sync deletes this file and breaks both this reference and the one in `build-and-test.yml`. The `maven-authenticate.yml` and `create-authenticated-npmrc.yml` changes here are already byte-identical to upstream, so land `pypi-auth-dev-feed.yml` upstream the same way, or move it under `eng/pipelines/templates/steps/` and update both references.

jongio (commented):
Following up on the new NuGet/PyPI auth commit: it adds a second reference to `pypi-auth-dev-feed.yml`, which still isn't landed upstream. Inline note has the fix.

jongio (commented):
`azure-dev - cli` can't run on this PR at all, so most of what this changes is unexercised.

Your `/azp run azure-dev - cli` at 00:58 came back with "1 pipeline(s) were filtered out due to trigger conditions". `eng/pipelines/release-cli.yml` is why:

```yaml
pr:
  paths:
    include:
      - go.mod
      - cli/
      - eng/pipelines/release-cli.yml
      - eng/pipelines/templates/steps/publish-cli.yml
    exclude:
      - cli/azd/extensions/**
```

Everything this PR touches is either under `eng/pipelines/templates/{jobs,stages,steps}`, which isn't in the include list, or under `cli/azd/extensions/microsoft.azd.demo`, which is explicitly excluded. No changed file matches, so the pipeline never queues. The CD `trigger:` block right above it already lists `/eng/pipelines/templates/jobs/build-cli.yml`; the `pr:` block just never got the same entries.

`azure-dev - vscode` has the same gap. `release-vscode.yml` only includes `ext/vscode` and `eng/pipelines/release-vscode.yml`, neither of which this PR touches.

That leaves `azure-dev - ext - microsoft.azd.demo` as the only one of the three @danieljurek listed that actually ran, and it goes through `release-azd-extension.yml`, which never pulls in `build-cli.yml`, `build-and-test.yml`, or `build-msi.yml`. So the NuGet, npm, Maven and pip changes have zero coverage here, and the first real execution is post-merge on `main` with CFSClean already turned on.

Adding these to the `pr:` include lists in both pipelines would let you validate before merging:

```yaml
      - eng/pipelines/templates/jobs/build-cli.yml
      - eng/pipelines/templates/stages/build-and-test.yml
      - eng/pipelines/templates/steps/build-msi.yml
      - eng/pipelines/templates/stages/1es-redirect.yml
```

Two inline comments below: the `eng/common` sync problem from my last pass (I tracked down the upstream file it should be named after) and a dead env var in `build-cli.yml`.

eng/common/pipelines/templates/steps/pypi-auth-dev-feed.yml
jongio: Still open from my last pass, and I found the specific fix this time. This template already exists upstream as `eng/common/pipelines/templates/steps/auth-dev-feed.yml` in `azure-sdk-tools`. Same parameters, same steps, same uv block; only quote style and indentation differ. There is no `pypi-auth-dev-feed.yml` upstream.

The `eng/common` sync isn't additive. It clears the destination directory and re-copies from `azure-sdk-tools`, so downstream-only files get deleted (that's what removed files in #5413). Once the next sync lands, this file is gone and both `- template: /eng/common/pipelines/templates/steps/pypi-auth-dev-feed.yml` references, `build-cli.yml:115` and `build-and-test.yml:161`, fail during template expansion, before the run starts.

Rename it to `auth-dev-feed.yml` using the upstream bytes and point both references at that name. Then the sync overwrites it harmlessly instead of deleting it.

Related, so nobody "fixes" the wrong thing: `create-authenticated-npmrc.yml` and `maven-authenticate.yml` in this PR are already byte-identical to upstream (blobs `3fa699b387d` and `1b84e984add` match `azure-sdk-tools` exactly). Don't apply local edits to those two, including the suggestion to switch `artifactsFeeds` to `public/azure-sdk-for-java`. Anything like that has to land upstream first or the next sync reverts it.

eng/pipelines/templates/jobs/build-cli.yml
```diff
@@ -154,6 +171,8 @@ jobs:
           # Code Coverage: Generate junit report to publish results
           GOTESTSUM_JUNITFILE: junitTestReport.xml
           SYSTEM_ACCESSTOKEN: $(System.AccessToken)
+          PIP_INDEX_URL: $(PIP_INDEX_URL)
+          PIP_EXTRA_INDEX_URL: $(PIP_EXTRA_INDEX_URL)
```
jongio: `PIP_EXTRA_INDEX_URL` never gets set, so this hands literal text to the test process.

The template gives `PipAuthenticate@1` a single feed with `onlyAddExtraIndex: false`. The task shifts that one endpoint into `PIP_INDEX_URL` and only sets `PIP_EXTRA_INDEX_URL` when endpoints remain, which they don't (`Tasks/PipAuthenticateV1/pipauthenticatemain.ts`). Azure Pipelines leaves unresolved `$(...)` macros as-is, so `go test` ends up with `PIP_EXTRA_INDEX_URL=$(PIP_EXTRA_INDEX_URL)`.

pip warns on a non-URL index and ignores it rather than failing, so nothing breaks. It's just log noise and a misleading env var. Drop the line.

jongio (commented):
Re-checked `5906b98`. Two things still stand, and one of the open questions resolves.

**`pypi-auth-dev-feed.yml` gets deleted by the next common-engineering sync**

I pulled the upstream blob SHAs to work out which of the three `eng/common` files are actually safe:

| file | upstream in `azure-sdk-tools` | this PR |
|---|---|---|
| `create-authenticated-npmrc.yml` | `3fa699b387d` | `3fa699b387d`, identical |
| `maven-authenticate.yml` | `1b84e984add` | `1b84e984add`, identical |
| `pypi-auth-dev-feed.yml` | 404, doesn't exist | `fc2396c8448` |

The first two match byte for byte because they were landed upstream first. `pypi-auth-dev-feed.yml` wasn't, and its content already exists upstream under a different name: `eng/common/pipelines/templates/steps/auth-dev-feed.yml` (`cd8fa23e67e`). Same four parameters, same `Setup DevOpsFeedName` step, same Twine, Pip and uv blocks, same `UV_PYTHON_INSTALL_MIRROR` value. Only quoting and indentation differ.

`eng/common/README.md` says updates here "end up getting overwritten with future updates", and the sync clears the destination rather than merging, so the file disappears and both of these break at template expansion:

- `eng/pipelines/templates/jobs/build-cli.yml:115`
- `eng/pipelines/templates/stages/build-and-test.yml:161`

Dropping the new file and repointing both references is the whole fix:

```yaml
- template: /eng/common/pipelines/templates/steps/auth-dev-feed.yml
```

**Still 1 of the 3 pipelines @danieljurek asked for**

Checks on `5906b98` only show `azure-dev - ext - microsoft.azd.demo`. `azure-dev - cli` and `azure-dev - vscode` still don't queue, because the `pr:` path filters in `release-cli.yml` and `release-vscode.yml` don't list any file this PR touches. So the NuGet, npm, Maven and pip changes stay unexercised until this is on `main` with CFSClean already enabled.

**On the `build-msi.yml` feed switch**

`WixToolset.DUtil` and `WixToolset.WcaUtil` 4.0.0 both resolve on `public/azure-sdk-for-net` through its nuget.org upstream, so `vstsFeed` points at a feed that carries what `cli/installer/windows.sln` restores. The open item there is just the `includeNuGetOrg` default, which is already flagged inline.

jongio (commented):
Re-checked `5906b98` today. No new commits since my last pass, so the two open items still stand, and I re-verified the main one against upstream just now.

**`pypi-auth-dev-feed.yml` still gets deleted by the next `eng/common` sync.** As of today it still doesn't exist in `azure-sdk-tools` (404), while the equivalent `auth-dev-feed.yml` does, with the same parameters, the same steps, and the same uv block. `eng/common` is overwritten by the common-engineering sync, so once it runs this file disappears and both `- template: /eng/common/pipelines/templates/steps/pypi-auth-dev-feed.yml` references fail at expansion time: `build-cli.yml:115` and `build-and-test.yml:161`. The `create-authenticated-npmrc.yml` and `maven-authenticate.yml` blobs in this PR are byte-identical to upstream (`3fa699b387d` and `1b84e984add`), so those two are safe to leave as-is. The fix is to drop the new file and point both references at `auth-dev-feed.yml`, or land `pypi-auth-dev-feed.yml` upstream first the way maven was. Full details are in the inline thread.

**`PIP_EXTRA_INDEX_URL: $(PIP_EXTRA_INDEX_URL)` in `build-cli.yml` is still a dead macro.** `PipAuthenticate@1` runs with a single feed and `onlyAddExtraIndex: false`, so it only populates `PIP_INDEX_URL`. `PIP_EXTRA_INDEX_URL` never gets set, and the test process receives the literal `$(PIP_EXTRA_INDEX_URL)`. pip ignores a non-URL index so nothing breaks, but it's misleading log noise. Drop the line. Inline thread has it.

One process note, not blocking on its own: only `azure-dev - ext - microsoft.azd.demo` actually runs on this PR. `azure-dev - cli` and `azure-dev - vscode` don't queue, because their `pr:` path filters don't list any file this PR touches, so the NuGet, npm, Maven, and pip wiring stays unexercised until this lands on `main` with CFSClean already on. Adding the touched template paths to those `pr:` include lists would let you validate before merge.

jongio (commented):
Re-verified `5906b98` against `azure-sdk-tools` upstream just now. No new commits since my last pass, so the same two items are still open, and the first is still a merge blocker.

**`pypi-auth-dev-feed.yml` gets deleted by the next `eng/common` sync.** Checked the upstream blobs again today: `pypi-auth-dev-feed.yml` still returns 404 in `azure-sdk-tools`, so it doesn't exist upstream. The equivalent `auth-dev-feed.yml` does exist (`cd8fa23e67e`) with the same parameters, the same steps, and the same uv block. `eng/common` is overwritten by the common-engineering sync (`eng/common/README.md`), and the sync clears the destination rather than merging, so a downstream-only file gets removed. Once it runs, `pypi-auth-dev-feed.yml` disappears and both `template:` references fail at expansion time, before the run starts: `build-cli.yml:115` and `build-and-test.yml:161`. Drop the new file and point both references at `auth-dev-feed.yml`, or land `pypi-auth-dev-feed.yml` upstream first the way `maven-authenticate.yml` was. For contrast, `create-authenticated-npmrc.yml` (`3fa699b387d`) and `maven-authenticate.yml` (`1b84e984add`) in this PR are already byte-identical to upstream, so those two are safe to leave as-is.

**`PIP_EXTRA_INDEX_URL: $(PIP_EXTRA_INDEX_URL)` in `build-cli.yml` is a dead macro.** `PipAuthenticate@1` runs with a single feed and `onlyAddExtraIndex: false`, so it only populates `PIP_INDEX_URL`. `PIP_EXTRA_INDEX_URL` never gets set, and the test process receives the literal `$(PIP_EXTRA_INDEX_URL)`. pip ignores a non-URL index so nothing breaks, but it's misleading log noise. Drop the line.

One process note, not blocking on its own: only `azure-dev - ext - microsoft.azd.demo` runs on this PR. `azure-dev - cli` and `azure-dev - vscode` don't queue, because their `pr:` path filters don't list any file this PR touches, so the NuGet, npm, Maven, and pip wiring stays unexercised until this lands on `main` with CFSClean already on. Adding the touched template paths to those `pr:` include lists would let you validate before merge. Details are in the existing inline threads.

jongio (commented):
Re-verified `5906b98`. No new commits since my last pass, so the same two items are open. I went and read the actual sync implementation this time, so the `pypi-auth-dev-feed.yml` one is now settled rather than inferred from the README.

The sync is driven by `eng/pipelines/templates/steps/sync-directory.yml` in `azure-sdk-tools`. The copy step is:

```powershell
Remove-Item -v -r -ErrorAction Ignore "${repoPath}/${filePattern}"
Copy-Item -v -r `
  "$(System.DefaultWorkingDirectory)/$(Build.Repository.Name)/eng/common/${filePattern}" `
  $destPath
```

`FilePatterns` defaults to empty, which the script turns into `*`, and `$repoPath` is `azure-dev/eng/common`. So the whole tree is removed and re-copied from upstream. It isn't a merge. Anything under `eng/common` without an upstream counterpart gets deleted and never comes back.

`pypi-auth-dev-feed.yml` still 404s in `azure-sdk-tools` as of today, so once the next sync runs the file disappears and both of these fail at template expansion, before any job starts:

* `eng/pipelines/templates/jobs/build-cli.yml:115`
* `eng/pipelines/templates/stages/build-and-test.yml:161`

There's a live example of this sitting in the repo right now. `eng/common/pipelines/templates/steps/bypass-local-dns.yml` was deleted upstream on 2026-04-16 (azure-sdk-tools#15182). It's still here only because the last sync into this repo was #6932 on 2026-03-03. The next one removes it.

The fix is one line per caller. The identical template already exists upstream as `auth-dev-feed.yml`, same four parameters, same steps, same uv block, so drop the new file and repoint both references:

```yaml
- template: /eng/common/pipelines/templates/steps/auth-dev-feed.yml
```

Both call sites keep working unchanged: `build-cli.yml` passes `EnableTwineAuth: false` / `EnablePipAuth: true` / `EnableUvAuth: false`, and `build-and-test.yml` passes nothing and picks up the same defaults. For contrast, `create-authenticated-npmrc.yml` (`3fa699b387d`) and `maven-authenticate.yml` (`1b84e984add`) in this PR are byte-identical to upstream, so those two are safe as-is.

Smaller one, still open: `PIP_EXTRA_INDEX_URL: $(PIP_EXTRA_INDEX_URL)` in `build-cli.yml`. `PipAuthenticate@1` runs with a single feed and `onlyAddExtraIndex: false`, so it only populates `PIP_INDEX_URL`. The test process ends up with the literal macro text. pip ignores a non-URL index so nothing breaks, but the line should go.

Worth repeating that `azure-dev - cli` still doesn't queue on this PR, since nothing it changes matches the `pr:` include list in `release-cli.yml`. The npm, Maven, NuGet and pip changes are unexercised here, and the first real run is post-merge on `main` with CFSClean already on.

jongio (commented):
Re-checked `5906b98`. The head commit replaced the repo-level `NuGet.Config` with `dotnet nuget add source`, and the `<clear />` didn't come with it. That part is new since my last pass, so it's the main thing here.

`NuGet.Config` was deleted in that same commit. It carried:

```xml
<packageSources>
  <clear />
  <!-- Do not add any additional feeds if new packages are needed they need to come from our azure-sdk-for-net DevOps feed which has an upstream set to nuget.org -->
  <add key="azure-sdk-for-net" value="https://pkgs.dev.azure.com/azure-sdk/public/_packaging/azure-sdk-for-net/nuget/v3/index.json" />
</packageSources>
```

`dotnet nuget add source` is additive. It writes the feed into the agent's user-level config and leaves the default `nuget.org` entry alone, so BuildCLI restores now resolve against nuget.org, azure-sdk-for-net and dotnet9. There's no `NuGet.Config` left anywhere in the tree at this SHA, so nothing else clears it. Details are in the inline thread on `build-cli.yml:109`.

The move to user level covers `Test_CLI_Aspire_DetectGen`, which runs `dotnet` from `t.TempDir()` where a repo-level config is never discovered. It's only the `<clear />` that got dropped.

`build-msi.yml` has the same gap from the other side. `feedsToUse: select` generates a config from `vstsFeed`, but `includeNuGetOrg` in `NuGetCommandV2/task.json` is `"defaultValue": "true"` with `visibleRule: selectOrConfig = select`, so nuget.org lands in that generated config too. That's the inline thread already open on line 18, and `includeNuGetOrg: false` closes it.

Two items from earlier passes are unchanged. I re-verified both against upstream today:

* `pypi-auth-dev-feed.yml` still 404s in `azure-sdk-tools`, and `auth-dev-feed.yml` is still there at `cd8fa23e67e` with the same four parameters and the same steps. The `eng/common` sync clears the destination and re-copies, so this file disappears and `build-cli.yml:115` and `build-and-test.yml:161` fail at template expansion. Repointing both references at `auth-dev-feed.yml` is the whole fix. For contrast, `create-authenticated-npmrc.yml` (`3fa699b387d`) and `maven-authenticate.yml` (`1b84e984add`) still match upstream byte for byte.
* `PIP_EXTRA_INDEX_URL: $(PIP_EXTRA_INDEX_URL)` in `build-cli.yml:175` is still a dead macro. `PipAuthenticateV1` only calls `tl.setVariable("PIP_EXTRA_INDEX_URL", ...)` when endpoints remain after the first one is shifted into `PIP_INDEX_URL`, and there's a single feed here, so it never fires.

Still worth noting that `azure-dev - cli` doesn't queue on this PR, so the NuGet, npm, Maven and pip wiring stays unexercised until this is on `main` with CFSClean already on.

eng/pipelines/templates/jobs/build-cli.yml
```diff
@@ -101,9 +106,21 @@ jobs:
           version: '9.x'
         displayName: Set up .NET 9
 
+      - bash: dotnet nuget add source --name azure-sdk-for-net https://pkgs.dev.azure.com/azure-sdk/public/_packaging/azure-sdk-for-net/nuget/v3/index.json
+        displayName: Add Azure SDK dev nuget feed
```
jongio: `dotnet nuget add source` only appends. It writes this feed into the agent's user-level NuGet config and leaves the default `nuget.org` source in place, so restores in this job resolve against nuget.org, azure-sdk-for-net and dotnet9.

The `NuGet.Config` removed in this same commit is what used to prevent that. It carried `<clear />` plus a comment saying packages should only come from `azure-sdk-for-net`, which has nuget.org set as an upstream. There is no `NuGet.Config` left anywhere in the tree at this SHA, so nothing takes over that job.

A direct public source is exactly what the rest of this PR redirects away from for npm, Maven and pip, so it needs the same treatment here. Either write the user-level config yourself with `<clear />`, or drop the default source after the two adds:

```yaml
      - bash: dotnet nuget disable source nuget.org
        displayName: Disable nuget.org so restores stay on the CFS-backed feed
```

eng/pipelines/templates/stages/1es-redirect.yml
```diff
@@ -32,7 +32,7 @@ extends:
           enabled: true
     settings:
       skipBuildTagsForGitHubPullRequests: true
-      networkIsolationPolicy: Permissive
+      networkIsolationPolicy: Permissive, CFSClean
```
jongio, line 35: `Permissive, CFSClean` isn't Permissive winning. It allows outbound traffic except public package feeds, so every pipeline extending this template has to pull npm, pip, Maven and NuGet through an authenticated Azure Artifacts feed.

The feed auth in this PR is job-scoped, and it only lands in `build-cli.yml` and `build-and-test.yml`. 25 pipelines extend this template, and there's no `.npmrc` tracked anywhere in the repo, so these still go straight to public registries:

- `release-vscode.yml` -> `vscode-build.yml:37-38` (`npm i -g npm @vscode/vsce`, `npm ci`). That's the `azure-dev - vscode` pipeline @danieljurek listed above as one that has to pass.
- `release-azuredevops.yml` -> `azuredevops-build.yml:38-39` (`npm i -g npm tfx-cli`, `npm ci`)
- `eval-unit.yml:50,58` (`npm ci`, `pip install pytest`). Its `pr:` trigger covers `cli/azd/cmd/root.go`, so it fires on a lot of PRs.
- `eval-report.yml:41`, `eval-e2e.yml:64,67`, `eval-waza.yml:53,56`
- `release-cli.yml` -> `verify-installers.yml:77,83` (`npx -y http-server` pulls at run time)

Either wire `create-authenticated-npmrc.yml` into those jobs, plus the pip template where Python runs, or hold CFSClean until they're migrated. Right now only the CLI and demo extension pipelines have been exercised, and vscode is already on the must-pass list.

jongio (commented):
CFSClean is set on the shared 1ES template, so it applies to all 25 pipelines that extend it, but feed auth was only wired into `build-cli.yml` and `build-and-test.yml`. Details inline.

Still open from my earlier passes, unchanged at this SHA:

- `eng/common/pipelines/templates/steps/pypi-auth-dev-feed.yml` gets deleted by the next common-engineering sync, which breaks the references at `build-cli.yml:115` and `build-and-test.yml:161`. Upstream already ships this exact template as `auth-dev-feed.yml`.
- `PIP_EXTRA_INDEX_URL: $(PIP_EXTRA_INDEX_URL)` at `build-cli.yml:175` is never set, so the literal macro reaches the test process.
- The two `dotnet nuget add source` calls append only, so nuget.org stays enabled and restores can still resolve publicly.

One more that hasn't been picked up yet: `build-msi.yml` needs `includeNuGetOrg: false`. `NuGetCommand@2` defaults it to true in `select` mode, so the MSI restore re-adds nuget.org even with `vstsFeed` set.

eng/pipelines/templates/steps/build-msi.yml
```diff
@@ -14,6 +14,8 @@ steps:
     displayName: Restore NuGet packages
     inputs:
       restoreSolution: cli/installer/windows.sln
+      feedsToUse: select
+      vstsFeed: 'public/azure-sdk-for-net'
```
jongio, line 18: `feedsToUse: select` doesn't drop nuget.org. `includeNuGetOrg` defaults to `true` for this task in select mode (`Tasks/NuGetCommandV2/task.json` in `microsoft/azure-pipelines-tasks`: `"defaultValue": "true"`, `"visibleRule": "selectOrConfig = select"`), so the generated config ends up with `public/azure-sdk-for-net` plus nuget.org.

That leaves the MSI restore reaching a public registry, which is what CFSClean is here to stop. It's a live path: `release-cli.yml:68` sets `BuildTestMsi: true` and that pipeline extends `1es-redirect.yml`.

```suggestion
      feedsToUse: select
      vstsFeed: 'public/azure-sdk-for-net'
      includeNuGetOrg: false
```

jongio (commented):
No new commits since my last pass, so the four threads I left are still open at 5906b982: the `pypi-auth-dev-feed.yml` filename versus upstream `auth-dev-feed.yml`, the unset `PIP_EXTRA_INDEX_URL`, `dotnet nuget add source` leaving nuget.org in the effective config now that `NuGet.Config` is deleted, and CFSClean reaching all 26 pipelines that extend `1es-redirect.yml` while feed auth only reaches two of them.

One more on the MSI restore below.

eng/pipelines/templates/jobs/azuredevops-build.yml
```diff
@@ -39,6 +39,8 @@ jobs:
     steps:
       - checkout: self
 
+      - template: /eng/common/pipelines/templates/steps/create-authenticated-npmrc.yml
```
jongio, line 42: This job is the only CFSClean-affected pipeline actually running on this PR, but lines 54, 66 and 75 set `continueOnError` for PR builds, so it reports success whether or not this new `.npmrc` works.

The comment block at lines 27-36 says the npm steps were made non-blocking because the self-hosted agents can't reach registry.npmjs.org, tracked in #9286. Pointing npm at the authenticated Azure Artifacts feed is the fix for exactly that, so are you planning to drop the `continueOnError` lines and that comment here? Leaving them in means the one piece of live evidence available on this PR stays muted, and a broken feed setup still looks green.

jongio (commented):
Re-checked at `030b4618`. The two new commits wire `create-authenticated-npmrc.yml` into `vscode-build.yml` and `azuredevops-build.yml`, which closes most of the npm gap I raised on the `1es-redirect.yml` thread. I also re-pulled the upstream blobs: `create-authenticated-npmrc.yml` and `maven-authenticate.yml` now match `Azure/azure-sdk-tools` exactly, so both survive the next common-engineering sync. Those two are settled.

One npm consumer is still uncovered. `npx -y http-server` at `eng/pipelines/templates/jobs/verify-installers.yml:77` and `:83` runs with no authenticated registry, and `release-cli.yml` reaches it through the `Verify_Installers` stage. Details on the coverage thread.

Four items from earlier passes are unchanged at this head:

1. `eng/common/pipelines/templates/steps/pypi-auth-dev-feed.yml` still 404s upstream, while `auth-dev-feed.yml` exists there with the same parameters and the same uv block. The sync clears the directory before copying, so it deletes this file and breaks the references at `build-cli.yml:115` and `build-and-test.yml:160`. Point both at `auth-dev-feed.yml` and drop the local copy.

2. `PIP_EXTRA_INDEX_URL` at `build-cli.yml:175` never gets set. `PipAuthenticate@1` runs with `onlyAddExtraIndex: false`, which puts the one feed into `PIP_INDEX_URL` and leaves the extra-index variable unset, so the test process receives the literal `$(PIP_EXTRA_INDEX_URL)` text.

3. `dotnet nuget add source` at `build-cli.yml:109` only appends. There is no `NuGet.Config` in the tree at this head, so nuget.org stays in the effective config and restores can still resolve against it.

4. `build-msi.yml:17` sets `feedsToUse: select` without `includeNuGetOrg: false`. That input defaults to true, so the generated config gets `public/azure-sdk-for-net` plus nuget.org.

On validation: `azure-dev - cli` and `azure-dev - vscode` still aren't in the check list on this PR, so `build-cli.yml`, `build-and-test.yml`, `build-msi.yml`, `verify-installers` and `vscode-build.yml` are all unexercised here. `azure-dev - azdo-extension` is running, which makes it the only CFSClean-affected pipeline with live evidence, and there is an inline note about why that signal is currently muted.

eng/pipelines/templates/stages/1es-redirect.yml
jongio: Update at `030b4618`, and this narrows to one job.

`vscode-build.yml` and `azuredevops-build.yml` both pick up `create-authenticated-npmrc.yml` in the two new commits, and the `eval-*.yml` pipelines I referenced earlier no longer exist at this head, so those parts are resolved.

The remaining npm consumer inside a CFSClean pipeline is `Verify_Installers`:

`release-cli.yml:47` extends this template, line 139 pulls in `/eng/pipelines/templates/stages/verify-installers.yml`, and that stage includes `/eng/pipelines/templates/jobs/verify-installers.yml` at line 264. That job runs `nohup npx -y http-server &` at line 77 on Linux and macOS, and `Start-Process npx.cmd -ArgumentList @('-y', 'http-server')` at line 83 on Windows. `npx -y` resolves `http-server` from the registry at run time, the job has no `.npmrc`, and npmrc creation is job scoped so nothing from `build-cli.yml` carries over.

Same one-line add you used in the other three jobs, right after `- checkout: self` at line 35:

```yaml
      - template: /eng/common/pipelines/templates/steps/create-authenticated-npmrc.yml
```

Worth noting this one won't surface on the PR: `azure-dev - cli` isn't running here, so `Verify_Installers` is unexercised and this would first show up on a release build.

jongio: One more, and it revises my "the remaining npm consumer is `Verify_Installers`" from the last pass.

`release-vscode.yml:18` extends this template too, and its publish stage runs `npm install -g @vscode/vsce` at `eng/pipelines/templates/steps/publish-vscode.yml:84`, inside an `AzureCLI@2` inline script. `create-authenticated-npmrc.yml` went into `vscode-build.yml`, which is a different job on a different agent, so the publish job never sees that registry or the auth.

So there are two npm call sites left inside CFSClean pipelines with no authenticated feed:

- `eng/pipelines/templates/jobs/verify-installers.yml:77` and `:83`, `npx -y http-server`, reached through `release-cli.yml:139`
- `eng/pipelines/templates/steps/publish-vscode.yml:84`, `npm install -g @vscode/vsce`, reached through `release-vscode.yml:25,27`

Both sit on release-only paths, so PR builds stay green and this first shows up when you cut a release. Adding `create-authenticated-npmrc.yml` to those two jobs closes it.

jongio (commented):
Re-checked at `030b4618`. `create-authenticated-npmrc.yml` and `maven-authenticate.yml` now match upstream `azure-sdk-tools` byte for byte, so the `eng/common` sync concern on those two is resolved.

`azure-dev - cli` hasn't run against this head. Everything with real risk landed in `f292573c`: the `azure-sdk-for-net` source add, `NuGetAuthenticate@1`, the pypi auth template, and the `PIP_*` passthrough into `go test`. The cli build linked in the description predates that commit, and the only internal pipelines green on this head are the demo extension and azdo-extension. @danieljurek listed `azure-dev - cli` as a required pass for this to work, so can you run one against `030b4618` and link it before merge?

Four items from earlier passes are untouched by the last two commits:

- `eng/common/pipelines/templates/steps/pypi-auth-dev-feed.yml` still shadows upstream `auth-dev-feed.yml`, which exists in `azure-sdk-tools` with the same parameters, steps and uv block. Repointing `build-cli.yml:115` and `build-and-test.yml:161` at `auth-dev-feed.yml` and deleting the copy is the whole fix.
- `build-msi.yml:18` still needs `includeNuGetOrg: false`. `NuGetCommandV2/task.json` sets `"defaultValue": "true"` for it under `selectOrConfig = select`.
- `build-cli.yml:175` still passes `$(PIP_EXTRA_INDEX_URL)`, which `PipAuthenticate@1` never sets when it gets a single feed with `onlyAddExtraIndex: false`. `go test` receives the literal macro text.
- `build-cli.yml:110` `dotnet nuget add source` appends, so nuget.org stays active in the agent config.

Also left a note on the 1es-redirect thread about a second npm call site under CFSClean that I missed earlier.

ext/vscode/package.json
```diff
@@ -681,13 +681,13 @@
     },
     "devDependencies": {
         "@azure/core-auth": "~1",
-        "@microsoft/vscode-azext-eng": "1.1.0-alpha.1",
+        "@microsoft/vscode-azext-eng": "1.1.0-alpha.3",
```
bwateratmsft, line 684: This needs an accompanying `npm install` to update the lock file.

eng/pipelines/templates/jobs/build-cli.yml
```diff
@@ -101,9 +108,24 @@ jobs:
           version: '9.x'
         displayName: Set up .NET 9
 
-      - bash: dotnet nuget add source --name dotnet9 https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet9/nuget/v3/index.json
+      - bash: dotnet nuget add source --name azure-sdk-for-net "$(AzureSdkNuGetFeedUrl)"
+        displayName: Add Azure SDK dev nuget feed
+
+      - bash: dotnet nuget disable source nuget.org
```
jongio: **[HIGH]** This disable lands after MSI restore, so it never covers `build-msi.yml`.

`build-msi.yml` is invoked earlier in this job (around the Build Test MSI step), before these nuget source steps. That restore still goes through `NuGetCommand@2` with `includeNuGetOrg` defaulting to true, so nuget.org stays in play for the MSI path.

Move the add-source / disable nuget.org / `NuGetAuthenticate` block above the `build-msi.yml` template, and set `includeNuGetOrg: false` on the NuGetCommand in `build-msi.yml`.

jongio (commented):
Re-checked at `db8633a` (3 commits since `030b4618`).

**Addressed**
- `ext/vscode/package-lock.json` now locks `@microsoft/vscode-azext-eng` to `1.1.0-alpha.3` and `@vscode/test-electron` to `3.1.0`, matching `package.json`. BuildVscode Linux/Mac/Windows are green; the remaining vscode failure is Sign only.
- `dotnet nuget disable source nuget.org` plus `RestoreSources` on Test Go Binary closes most of the test-path nuget.org gap I raised earlier.

**Still open**
1. `pypi-auth-dev-feed.yml` is still the wrong filename under `eng/common`. Upstream is `auth-dev-feed.yml`. Next common-eng sync deletes this file and breaks the template reference.
2. `PIP_EXTRA_INDEX_URL` is still forwarded as a literal when PipAuthenticate only sets `PIP_INDEX_URL`.
3. MSI restore still hits nuget.org: `build-msi.yml` runs before the new disable step, and `includeNuGetOrg` still defaults true on the NuGetCommand. Details on the inline note.
4. `publish-vscode.yml` still runs `npm install -g @vscode/vsce` without `create-authenticated-npmrc.yml` under CFSClean.
5. `azuredevops-build.yml` still uses `continueOnError` on PR builds, so the new npmrc path is not actually validated on this PR.

<sub>🤖 agent jongio</sub>

eng/pipelines/templates/jobs/build-cli.yml
jongio: Update at `db8633a`: the new `dotnet nuget disable source nuget.org` step closes this for the Test Go Binary path, and `RestoreSources` further pins restores there.

It still doesn't cover MSI restore though. `build-msi.yml` runs earlier in the job, before the disable, and `NuGetCommand@2` still defaults `includeNuGetOrg` to true. Moving the feed setup above `build-msi.yml` and setting `includeNuGetOrg: false` would close the remaining gap.

eng/pipelines/templates/steps/build-msi.yml
jongio: Still open at `db8633a`, and the new nuget disable in `build-cli.yml` doesn't help this path.

`build-msi.yml` is invoked before the disable step, and this task builds its own config. `includeNuGetOrg: false` is still required here.

ext/vscode/package.json
jongio: Lockfile is updated at `db8633a` - `@microsoft/vscode-azext-eng` is `1.1.0-alpha.3` and `@vscode/test-electron` is `3.1.0`. BuildVscode Linux/Mac/Windows are green on this head.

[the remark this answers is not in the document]
jongio: Lockfile is updated at `db8633a` - `@microsoft/vscode-azext-eng` is `1.1.0-alpha.3` and `@vscode/test-electron` is `3.1.0`. BuildVscode Linux/Mac/Windows are green on this head.

eng/pipelines/templates/jobs/build-cli.yml
jongio: **[HIGH]** This disable lands after MSI restore, so it never covers `build-msi.yml`.

`build-msi.yml` is invoked earlier in this job (around the Build Test MSI step), before these nuget source steps. That restore still goes through `NuGetCommand@2` with `includeNuGetOrg` defaulting to true, so nuget.org stays in play for the MSI path.

Move the add-source / disable nuget.org / `NuGetAuthenticate` block above the `build-msi.yml` template, and set `includeNuGetOrg: false` on the NuGetCommand in `build-msi.yml`.

jongio (commented):
Re-checked at `db8633a` (3 commits since `030b4618`).

**Addressed**
- `ext/vscode/package-lock.json` now locks `@microsoft/vscode-azext-eng` to `1.1.0-alpha.3` and `@vscode/test-electron` to `3.1.0`, matching `package.json`. BuildVscode Linux/Mac/Windows are green; the remaining vscode failure is Sign only.
- `dotnet nuget disable source nuget.org` plus `RestoreSources` on Test Go Binary closes most of the test-path nuget.org gap I raised earlier.

**Still open**
1. `pypi-auth-dev-feed.yml` is still the wrong filename under `eng/common`. Upstream is `auth-dev-feed.yml`. Next common-eng sync deletes this file and breaks the template reference.
2. `PIP_EXTRA_INDEX_URL` is still forwarded as a literal when PipAuthenticate only sets `PIP_INDEX_URL`.
3. MSI restore still hits nuget.org: `build-msi.yml` runs before the new disable step, and `includeNuGetOrg` still defaults true on the NuGetCommand. Details inline.
4. `publish-vscode.yml` still runs `npm install -g @vscode/vsce` without `create-authenticated-npmrc.yml` under CFSClean.
5. `azuredevops-build.yml` still uses `continueOnError` on PR builds, so the new npmrc path is not actually validated on this PR.

<sub>🤖 agent jongio</sub>

eng/pipelines/templates/jobs/build-cli.yml
chidozieononiwu: Nuget command is using `vstsFeed: 'public/azure-sdk-for-net'`

jongio: vstsFeed alone does not drop nuget.org. `includeNuGetOrg` defaults to `true` on `NuGetCommand@2` in select mode (`microsoft/azure-pipelines-tasks` `NuGetCommandV2/task.json`: `defaultValue: "true"`), so the generated config still has nuget.org next to `public/azure-sdk-for-net`.

Confirmed again against the task definition just now. Needs:

```yaml
      includeNuGetOrg: false
```

on the NuGetCommand in `build-msi.yml`. The disable/auth block in `build-cli.yml` also still runs after `build-msi.yml` is invoked, so it never covers that restore either.

eng/pipelines/templates/stages/1es-redirect.yml
jongio: Update at `d6e9b22`: `vscode-sign.yml` now includes `create-authenticated-npmrc.yml`. That closes the signing path - `New-VsixSigningManifest.ps1` runs `npm exec --yes @vscode/vsce@latest`, so this was required.

Still open on release paths with no authenticated npmrc:

- `eng/pipelines/templates/jobs/verify-installers.yml:77` / `:83` - `npx -y http-server` (via `release-cli.yml`)
- `eng/pipelines/templates/steps/publish-vscode.yml:84` - `npm install -g @vscode/vsce` (via `release-vscode.yml` publish stages)

Same one-line template add after checkout / at the top of the publish steps closes both.

eng/common/pipelines/templates/steps/pypi-auth-dev-feed.yml
jongio: Still open at `d6e9b22`. Re-checked upstream just now:

- `azure-sdk-tools` `.../auth-dev-feed.yml` exists (blob `cd8fa23e67ed`)
- `azure-sdk-tools` `.../pypi-auth-dev-feed.yml` is 404

Local file is still `pypi-auth-dev-feed.yml`, referenced from `build-cli.yml:120` and `build-and-test.yml:161`. Next `eng/common` sync will delete it and both pipelines fail at template expansion. Rename to `auth-dev-feed.yml` and retarget the two references.

eng/pipelines/templates/steps/build-msi.yml
jongio: Still open at `d6e9b22`. No change to `build-msi.yml` in the latest commit. `includeNuGetOrg: false` is still required here - `vstsFeed` alone does not exclude nuget.org (see the reply on the build-cli disable-order thread).

eng/pipelines/templates/jobs/build-cli.yml
jongio: Still open at `d6e9b22`. No change to this env block. Drop `PIP_EXTRA_INDEX_URL: $(PIP_EXTRA_INDEX_URL)` - with `onlyAddExtraIndex: false` the task only sets `PIP_INDEX_URL`.

eng/pipelines/templates/jobs/azuredevops-build.yml
jongio: Still open at `d6e9b22`. Feed auth is in place on this job now, so the PR-only `continueOnError` on the npm/test/package steps is still muting the only live CFSClean signal on this PR. Worth dropping those three flags (and the #9286 comment block) once you confirm the authenticated feed works on the self-hosted agents.

eng/common/pipelines/templates/steps/pypi-auth-dev-feed.yml
```diff
@@ -0,0 +1,49 @@
...
+      - task: PipAuthenticate@1
+        displayName: "Pip Authenticate to feed"
+        inputs:
+          artifactFeeds: $(DevFeedName)
+          onlyAddExtraIndex: false
+
+  - ${{ if eq(parameters.EnableUvAuth, true) }}:
+      - pwsh: |
+          if ($env:PIP_INDEX_URL) {
+            Write-Host "Found pip index URL: $($env:PIP_INDEX_URL)"
```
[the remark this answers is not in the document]
jongio: Agree - and this file is going away on the next eng/common sync anyway (upstream name is `auth-dev-feed.yml`). When you rename to the upstream template, pull the upstream bytes rather than carrying this `Write-Host` of `PIP_INDEX_URL` forward. Upstream already has the same uv block pattern; do not keep a local print of the credential-bearing URL.

jongio:
Re-checked at `d6e9b22` (1 commit since `db8633a`).

**Addressed**
- `vscode-sign.yml` now pulls in `create-authenticated-npmrc.yml`. That path needs it: `New-VsixSigningManifest.ps1` runs `npm exec --yes @vscode/vsce@latest`, so signing under CFSClean was going to hit the public registry without this.

**Still open** (details on the existing threads)
1. **`pypi-auth-dev-feed.yml` will be deleted by the next `eng/common` sync** - upstream has `auth-dev-feed.yml` only. Rename and retarget the two references in `build-cli.yml` and `build-and-test.yml`.
2. **MSI NuGet still reaches nuget.org** - `includeNuGetOrg` defaults to `true` on `NuGetCommand@2` even with `vstsFeed` set. Needs `includeNuGetOrg: false` on `build-msi.yml`, and the disable/auth block in `build-cli.yml` still has to run before that template.
3. **Two release-path npm call sites still lack auth** - `verify-installers.yml` (`npx -y http-server`) and `publish-vscode.yml` (`npm install -g @vscode/vsce`). `vscode-sign` is closed; these two are not.
4. **`PIP_EXTRA_INDEX_URL` env line is still a no-op** - drop it from the Test Go Binary env block.
5. **`azuredevops-build.yml` still has PR-only `continueOnError`** - with feed auth in place, that mute is worth dropping so the PR build is real signal.

Items 1-3 are the merge-relevant ones.

<sub>agent jongio</sub>

eng/common/pipelines/templates/steps/create-authenticated-npmrc.yml
```diff
@@ -1,33 +1,41 @@
...
-    Write-Host "Creating .npmrc file ${{ parameters.npmrcPath }} for registry ${{ parameters.registryUrl }}"
-    $parentFolder = Split-Path -Path '${{ parameters.npmrcPath }}' -Parent
-    
-    if (!(Test-Path $parentFolder)) {
-      Write-Host "Creating folder $parentFolder"
-      New-Item -Path $parentFolder -ItemType Directory | Out-Null
-    }
+  - pwsh: |
+      $npmrcPath = '${{ parameters.npmrcPath }}'
+      if (-not $npmrcPath) { $npmrcPath = Join-Path $HOME '.npmrc' }
```
danieljurek, line 21: TIL `$HOME` in PowerShell is set on both Windows and *nix platforms. https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_automatic_variables?view=powershell-7.6#home

There's also `$(Agent.HomeDirectory)` [predefined](https://learn.microsoft.com/en-us/azure/devops/pipelines/build/variables?view=azure-devops&tabs=yaml) in the pipeline but this works well enough.

Neat.

eng/common/pipelines/templates/steps/create-authenticated-npmrc.yml
```diff
@@ -1,33 +1,41 @@
...
-  condition: ${{ parameters.CustomCondition }}
-  inputs:
-    workingFile: ${{ parameters.npmrcPath }}
-    azureDevOpsServiceConnection: ${{ parameters.ServiceConnection }}
+      if ($parentFolder -and -not (Test-Path $parentFolder)) {
+        Write-Host "Creating folder $parentFolder"
+        New-Item -Path $parentFolder -ItemType Directory | Out-Null
+      }
+
+      "registry=${{ parameters.registryUrl }}" | Out-File $npmrcPath
```
danieljurek, line 31: Is overwriting existing content intended? Since this is only running on agents it's probably OK to do.

eng/common/pipelines/templates/steps/pypi-auth-dev-feed.yml
danieljurek: I think Jon's comments are already covering what I was going to say: Make sure this one goes into `eng/common`

eng/pipelines/templates/jobs/build-cli.yml
```diff
@@ -101,9 +108,24 @@ jobs:
...
+        displayName: Disable public nuget feed
+
+      - bash: dotnet nuget add source --name dotnet9 "$(DotNet9NuGetFeedUrl)"
         displayName: Add internal dotnet nuget feed
```
danieljurek: These should all run in one step, log each command as it's run, fail if any of them have a nonzero exit code.

ext/vscode/package.json
danieljurek, line 684: Why are these changes in this PR?

danieljurek (approved):

jongio (commented):
CFSClean isn't ready to enforce on this pipeline yet. The current `azure-dev - ext - microsoft.azd.demo - public` Windows job fails in `Install gotest`: the Terrapin Go credential provider gets a 401 from `Guardian1ESPTUpstreamOrgFeed`, then `go install gotest.tools/gotestsum@latest` falls back to blocked `https://proxy.golang.org`. Onboard and authorize this pipeline for the internal Go module proxy (https://aka.ms/goproxy_for_how_to_onboard), or hold CFSClean until that dependency path is working.

<sub>🤖 agent jongio</sub>

vhvb1989 (approved):

jongio:
Re-checked this against `c385c369`. My earlier blocking concern is cleared.

- The demo-extension Windows job that was failing in `Install gotest` (Terrapin Go credential provider 401, then falling back to the blocked `proxy.golang.org`) is green now. That was the one thing making me nervous about flipping on `CFSClean`.
- Dropping `PIP_EXTRA_INDEX_URL` from the Test Go Binary env block in `build-cli.yml` is the right call. `python-auth-dev-feed.yml` runs `PipAuthenticate@1` with `onlyAddExtraIndex: false`, so only `PIP_INDEX_URL` ever gets set and the unset macro would've passed through literally.
- I diffed all three `eng/common` templates against `Azure/azure-sdk-tools` and they're byte-identical, including the `python-auth-dev-feed.yml` rename. So the `Write-Host $env:PIP_INDEX_URL` credential-logging thing that got flagged a few times isn't fixable here. That belongs upstream in azure-sdk-tools.
- The npmrc additions to `verify-installers.yml` and `publish-vscode.yml` line up with where npm actually runs (`npm install -g @vscode/vsce`), and both call sites are inside 1ES deployment jobs, so the steps are valid there.

One non-blocking follow-up, not for this PR: `azuredevops-build.yml` still carries three `continueOnError: ${{ eq(variables['Build.Reason'], 'PullRequest') }}` flags plus the issue #9286 comment block, even though `create-authenticated-npmrc.yml` now runs in that same job. Those swallow feed failures on PR builds, which is exactly the signal you want loudest while `CFSClean` is settling in. Worth ripping out once you've got a few clean runs.

Nothing blocking from me.

<sub>🤖 agent jongio</sub>

jongio (approved):
Re-verified at `c385c369`. Everything I had open is closed:

- MSI NuGet restore now sets `includeNuGetOrg: false`, so it no longer falls back to nuget.org.
- npm auth is wired at the two remaining release-path call sites, `verify-installers.yml` and `publish-vscode.yml`.
- `PIP_EXTRA_INDEX_URL` is dropped from the Test Go Binary env block. `python-auth-dev-feed.yml` runs `PipAuthenticate@1` with `onlyAddExtraIndex: false`, so that macro was never set and would have passed through literally.
- The three `eng/common` templates are byte-identical to `Azure/azure-sdk-tools`, including the `python-auth-dev-feed.yml` rename.

One follow-up that isn't for this PR: `azuredevops-build.yml` still carries the PR-only `continueOnError` flags on the npm, test, and package steps, plus the artifact-staging skip, all tracked under #9286. Now that `create-authenticated-npmrc.yml` runs in that job, those swallow the exact feed failures you want loudest while `CFSClean` settles in. Worth removing after a few clean runs.

<sub>agent jongio</sub>

bwateratmsft (approved):
Approving for /ext/vscode

PR #9149, opened 2026-07-15
fix: read unified azure.yaml in agent run and deploy

### LINKED ISSUE #9151: Support unified azure.yaml in agent runtime operations
## Task

Make the Azure AI Agents extension consistently read and execute the unified `azure.yaml` service configuration across local run, deploy, provision, synthesis, and brownfield validation.

## Why

Agent configuration can be inline, in the deprecated `config:` block, or stored in local files referenced by `$ref`. These shapes are not currently resolved consistently across runtime paths, so the same project can behave differently between run, deploy, and provision. Referenced resources may be missed, environment expressions or typed values may be changed too early, and container operations may use the unresolved service directory.

## Scope

- Resolve one effective service configuration from inline, legacy `config:`, and local `$ref` forms.
- Preserve raw environment expressions until runtime and normalize supported scalar environment values.
- Use the same resolved configuration for local run, deploy, provision, synthesis, and brownfield validation.
- Support both `azure.yaml` and `azure.yml` project files.
- Pass the resolved service directory to core container build, package, and publish operations without rewriting project configuration on disk.
- Preserve compatibility with legacy `agent.yaml` and `agent.manifest.yaml` projects.
- Keep hosted-agent deploy session carryover disabled until it is restored or redesigned in separate follow-up work.

## Acceptance criteria

- Inline, legacy, and `$ref` service configurations produce consistent runtime behavior.
- Container operations use the resolved project-relative service path without mutating `azure.yaml` or cached project configuration.
- Invalid service paths outside the project are rejected.
- Existing legacy agent projects continue to work.
- Relevant unit, build, spelling, and required CI checks pass.

Implemented by #9149.

huimiu:
## Why this change is needed

Azure AI Agents services can now be defined directly in `azure.yaml`, under the deprecated `config:` block, or in local files referenced through `$ref`. These forms were not resolved consistently across local run, deploy, and provision. As a result, a valid project could lose environment expressions, skip referenced resources, or use inconsistent agent configuration.

## What changed

- Added a shared resolver for inline, legacy `config:`, and local `$ref` agent configuration.
- Applied the resolved configuration consistently across local run, deploy, provision, synthesis, and brownfield validation.
- Preserved environment expressions while resolving extension-owned configuration and normalized values at runtime without changing the existing schema type contract.
- Added `azure.yml` support while retaining compatibility with `agent.yaml` and `agent.yml` projects.
- Made `azd ai agent optimize apply` reject only root agent-definition references, avoiding writes to the wrong file without blocking unrelated nested references.
- Preserved Windows environment-variable precedence regardless of key casing.

## Configuration contract

A root `$ref` may provide agent-owned configuration such as the agent definition, protocols, environment settings, deployments, connections, and Foundry-specific settings. Core-owned execution fields (`project`, `language`, `image`, and `docker`) must remain inline on the service entry in `azure.yaml`.

## Scope

This PR intentionally leaves environment value type contracts unchanged in both the core and agent extension schemas. Supporting boolean, number, or null environment values, together with related type-focused review feedback, is outside the scope of this PR and can be handled separately. Runtime normalization is defensive and does not expand public schema support.

## Why this approach

Core azd selects frameworks and executes build, package, and deploy operations from the service entry it reads directly. Keeping execution fields inline makes that lifecycle deterministic and avoids a new cross-boundary API. The extension still resolves its own configuration without rewriting user files.

## Testing

From `cli/azd/extensions/azure.ai.agents`:

- `go test ./internal/cmd ./internal/pkg/projectconfig ./internal/project ./internal/synthesis`
- `go build ./...`

Closes #9151.
Part of #8710.


cli/azd/extensions/azure.ai.agents/internal/cmd/listen.go
```diff
@@ -194,29 +204,33 @@ func predeployHandler(ctx context.Context, azdClient *azdext.AzdClient, args *az
...
 	}
 
-	// Capture the current session so it can be resumed on the newly deployed
-	// version after deploy (see session_carryover.go). Best-effort; hosted
-	// agents only.
-	if isHostedAgentService(svc, args.Project) {
-		captureSessionForCarryover(ctx, azdClient, svc)
-	}
-
 	// Run developer RBAC pre-flight checks only for hosted agent deployments.
```
jongio, line 270: This removes the `captureSessionForCarryover` call here, and `carryOverSessionAfterDeploy` (plus the `agent_api` import) in `postdeployHandler`. That turns off the hosted-agent session carry-over that landed in #8942. Both functions still exist in `session_carryover.go` and keep their unit tests, so `unused` lint doesn't flag them, but nothing in the deploy path calls them anymore.

The PR description is scoped to reading the unified `azure.yaml`, so this removal isn't mentioned. Is it intentional, or a rebase artifact against #8942? If intentional, please call it out in the description. If not, these two calls need to come back.

jongio (commented):
One question on a behavior change that's outside what the description covers. Everything else (the $ref/unified resolution across run, deploy, provision, and the env precedence handling) reads cleanly and is well covered by tests.

jongio (approved):
The new commit restores the hosted-agent session carryover (capture in predeploy, resume in postdeploy) that the first commit had dropped, so my earlier question is resolved. The restored wiring matches what's on main. Approving.

jongio (approved):
The new commit swaps the persist/restore state mutation for a per-request `service_path` override on the container build, package, and publish calls. The server treats it as a one-operation override without touching the stored config, rejects path traversal via `filepath.IsLocal`, and falls back to protowire unknown fields for wire compatibility. That resolves my earlier question about how the referenced-service path was being carried into the container operations. Re-approving against the latest commit.

jongio (approved):
Re-approving against 459c25b5. The new commit disables the deploy session carryover by dropping the predeploy capture and the postdeploy resume calls. Carryover was opt-in via `AZD_AGENT_RESUME_SESSION_ON_DEPLOY` and both helpers returned early when it was unset, so this only changes behavior for anyone who had that flag set; default deploys are unaffected. The captured session id lived in an in-process map consumed only by the removed resume call, so nothing downstream depends on it. Build, vet, and the cmd package tests pass locally.

jongio (approved):
Re-reviewed the runtime-compatibility follow-up commit. The envSliceHasKey change correctly handles Windows case-insensitive env var names via EqualFold while staying case-sensitive elsewhere, and strings.Cut safely skips entries without an '='. Test coverage matches the platform behavior.

jongio (approved):
Re-approving on 224fd21. The new commit constrains file-ref resolution to a root-level "$ref" and preserves nested refs, and it tightens the container service path check with symlink resolution that rejects escapes via the nearest existing ancestor. Path validation correctly treats projectConfig.Path as the project directory, and the added tests cover the nested-ref and symlink-escape cases. Verified the removed ConfigContainsFileRef has no remaining references. Go lint/test/build are green.

jongio (approved):
Re-reviewed the latest commit (preserve unified agent configuration). The in-place env-var upsert keeps nested $ref entries intact, the azure.yml lookup ordering matches ReadProjectFile's contract, and the legacy config: fallback in ACR derivation is covered by tests. No concerns on the new changes.

jongio (approved):
Re-reviewed the session carry-over restoration (predeploy capture + postdeploy resume in listen.go, plus the README section). Wiring is scoped correctly, the endpoint/credential are the same ones gathered for postdeploy, and the opt-in best-effort behavior is intact.

jongio (approved):
Incremental changes look good. The inline-env normalization path, nil-safety on the config structs, and resolving file refs before the brownfield network check all hold up, and the new tests cover the added paths.

cli/azd/extensions/azure.ai.agents/internal/project/service_target_agent.go
```diff
@@ -605,6 +651,32 @@ func (p *AgentServiceTargetProvider) Package(
...
 
+const containerServicePathFieldNumber protowire.Number = 3
+
+func setContainerServicePath(request proto.Message, servicePath string) {
+	if servicePath == "" {
+		return
+	}
+	message := request.ProtoReflect()
+	field := message.Descriptor().Fields().ByNumber(
+		protoreflect.FieldNumber(containerServicePathFieldNumber),
```
trangevi: Why are we needing to interact with this via number and proto methods? Shouldn't the change to core allow us to interact with some part of the extension framework for this?

JeffreyCA:
The unified extension-side resolver looks directionally right, especially using one effective view across run, deploy, provision, synthesis, and brownfield validation.

The central contract question for me is: **must a root `$ref` be allowed to provide core-owned execution fields such as `project`, `language`, `image`, or `docker`?** I think the answer determines whether any new core API is needed.

I see two possible contracts:

### If no: keep core execution fields inline

Require core-owned execution fields (`project`, `language`, `image`, and `docker`) to stay inline:

```yaml
services:
  agent:
    host: azure.ai.agent
    language: docker
    project: ./src/agent
    docker:
      path: Dockerfile
    $ref: ./agent.yaml
```

```yaml
# agent.yaml
kind: hosted
name: support-agent
protocols:
  - protocol: responses
    version: 2.0.0
env:
  LOG_LEVEL: info
```

The extension would continue resolving agent-specific fields from `$ref`, while core receives everything it needs without a new RPC contract.

### If yes: resolve core execution fields before framework work

The current `service_path` contract appears both too late and too narrow. For a `language: docker` agent, core's framework build runs first by default, so it can build from the unresolved path or produce an artifact that causes the later `service_path` fallback to be skipped entirely. A referenced file can also provide other core-owned fields such as `image` or `docker`, not only `project`.

Instead, the owning service target could return a small typed, operation-scoped execution override before framework selection/build, covering fields such as relative path, language, image, and Docker options. Core would validate it and use the resulting effective `ServiceConfig` throughout the lifecycle without persisting it to `azure.yaml`. This would keep `$ref` semantics extension-owned without expanding into a generic core configuration resolver.

----

Either way, I think the resolver and compatibility work in this PR is probably worth keeping.

If a core contract is needed, we should split the work into two PRs: first land and release the core contract, then update the extension to consume the new `github.com/azure/azure-dev/cli/azd` version. That would avoid retaining the `protowire` compatibility workaround.

cli/azd/grpc/proto/container.proto
```diff
@@ -29,6 +29,8 @@ service ContainerService {
 message ContainerBuildRequest {
   string service_name = 1;
   ServiceContext service_context = 2;
+  // Project-relative path used only for this operation.
+  string service_path = 3;
```
vhvb1989: Design suggestion: consider grouping this into a single optional `options` message on each container RPC instead of a standalone `service_path` field.

```proto
message ContainerBuildRequest {
  string service_name = 1;
  ServiceContext service_context = 2;
  ContainerOperationOptions options = 3;   // same field on Package/Publish
}

// Optional inputs honored by container build/package/publish.
// Each field, when set, is used for that operation; when empty, core
// falls back to the service's own configuration.
message ContainerOperationOptions {
  string service_path = 1;
  string image = 2;
  DockerProjectOptions docker = 3;
}
```

**Why this helps going forward**

These container RPCs don't receive a `ServiceConfig` — core re-resolves the service by name from its own cached (unresolved) project config, so anything the extension resolves from `$ref`/`config:` that core's container pipeline reads has to be re-sent explicitly. Today that's only the path, but `containerHelper.Build/Package` also consume `Docker{...}` (Dockerfile path, context, build args, platform, target) and `Image`. If any of those ever live inside a `$ref`'d file, we'd hit the same gap and be tempted to add another top-level field each time — slowly turning the request into a partial copy of `ServiceConfig`.

A single `options` bag:
- lets new optional fields (`image`, `docker`, future per-call knobs like `no_cache`) be added without new top-level request fields or proto churn;
- is scoped to just these three RPCs (unlike `service_context`, which is a lifecycle-wide artifact ledger also used by deploy/hooks/native pipeline), so the intent stays local;
- can be applied by one shared helper across build/package/publish (generalizing the current `containerServiceConfig`), keeping validation — e.g. the project-relative/symlink check that `service_path` needs — in exactly one place.

Since field 3 is currently carried via the unknown-field shim (no published-SDK consumers yet), reshaping `service_path` into `options.service_path` is essentially free right now; after the SDK version bumps it would become a breaking change — so it's cheaper to land the grouped shape in this PR than to migrate later.

jongio (approved):

jongio (commented):
Re the automated flag about container RPC version compatibility: this looks already handled. The extension sets `requiredAzdVersion: ">=1.28.0"`, and the matching core `ContainerOperationOptions` (field 3) handling ships in this same PR (1.28.0-beta.1). Any host that can load the updated extension will understand the new options field, so the resolved service path won't silently no-op on a supported host, and there's no need to split the extension change out or bump to a later SDK for that reason.

jongio (approved):
Re-verified the outstanding automated flags against the current head.

The container RPC options field (`ContainerOperationOptions`, field 3) is understood by any host that can load this extension, since it pins `requiredAzdVersion >= 1.28.0` and the matching core handling ships in this same PR, so the resolved service path won't silently no-op on a supported host.

The docker `$ref` hydration and non-agent `$ref` resolution notes are narrow edge cases: docker overrides supplied only through a referenced file, or a non-Foundry service carrying an incompatible remote `$ref`. Neither matches the service shapes this extension emits today. Worth a follow-up if referenced docker blocks become a supported pattern, but not blocking here.

Approving.

jongio (approved):

cli/azd/extensions/azure.ai.agents/extension.yaml
```diff
@@ -6,7 +6,8 @@ description: Ship agents with Microsoft Foundry from your terminal. (Beta)
 usage: azd ai agent <command> [options]
 # NOTE: Make sure version.txt is in sync with this version.
 version: 1.0.0-beta.5
-requiredAzdVersion: ">=1.27.1"
+# Container service_path support starts in azd 1.28.0.
+requiredAzdVersion: ">=1.28.0"
```
jongio: The comment above says container service_path support starts in 1.28.0, but the core field-3 reader is in #9170, which is still open, and released 1.28.0 doesn't include it. On a 1.28.0 host, core treats field 3 as an unknown field and drops the resolved service path silently. Pin this to the first release that includes #9170.

jongio (commented):
Following up on my earlier note about container RPC version compatibility, which I need to correct.

I previously said the core `ContainerOperationOptions` (field 3) handling ships in this same PR, so any host meeting the required version would understand it. That isn't right. The core reader lives in #9170 (`container_service.go` and `container.proto`), which is still open against main. azd core 1.28.0 was released on 2026-07-15 without it, and field 3 is absent from both the 1.28.0 tag and main.

With `requiredAzdVersion: ">=1.28.0"`, a user on the released 1.28.0 can install this extension, but core treats field 3 as an unknown field and drops the resolved service path. The container build, package, and publish then run against the unresolved settings, silently. Worth pinning the gate to the first release that actually contains #9170 rather than 1.28.0.

Minor and non-blocking: the def/service env merge in `run.go` builds an exact-case map, so on Windows a definition `PATH` plus a service `Path` both land in the slice. The sort emits `PATH` first and `envSliceHasKey`'s case-insensitive check then suppresses `Path`, so the definition value wins over the service value. Narrow case (same key, different casing, Windows only).

cli/azd/extensions/azure.ai.agents/internal/cmd/resource_services.go
```diff
@@ -355,10 +357,26 @@ func reserveServiceName(used map[string]string, name, source string) error {
 // service carries any, so an azure.yaml written before the per-resource split
 // still provisions without re-running init.
 func collectProjectDeployments(services map[string]*azdext.ServiceConfig) ([]project.Deployment, error) {
```
trangevi: This wrapper seems unnecessary since there's no conditional handling, we should just either update this method to the new contract, or change all the callers to use the new method

cli/azd/extensions/azure.ai.agents/internal/cmd/resource_services.go
```diff
@@ -387,24 +408,46 @@ func collectProjectDeployments(services map[string]*azdext.ServiceConfig) ([]pro
 // agent service when no connection service carries any, so a pre-split
 // azure.yaml still provisions without re-running init.
 func collectConnections(services map[string]*azdext.ServiceConfig) ([]project.Connection, error) {
+	return collectConnectionsAtRoot(services, "")
```
trangevi: Same comment about the unnecessary wrapper method

cli/azd/extensions/azure.ai.agents/internal/cmd/listen.go
```diff
@@ -716,10 +738,47 @@ func setEnvVar(ctx context.Context, azdClient *azdext.AzdClient, envName string,
...
 }
 
-func populateContainerSettings(ctx context.Context, azdClient *azdext.AzdClient, svc *azdext.ServiceConfig) error {
+func populateContainerSettings(
+	_ context.Context,
+	_ *azdext.AzdClient,
+	svc *azdext.ServiceConfig,
+	projectRoot string,
+) error {
+	_, err := prepareContainerSettings(svc, projectRoot)
```
trangevi: Same comment as the other wrapper methods, seems like unnecessary overhead.

cli/azd/extensions/azure.ai.agents/internal/project/agent_definition.go
```diff
@@ -223,11 +311,187 @@ func LoadServiceTargetAgentConfig(svc *azdext.ServiceConfig) (*ServiceTargetAgen
...
+			return err
+		}
+		svc.Config = resolved
+	}
+	return nil
+}
+
+// NormalizeServiceConfigInPlace converts environment scalars without
+// resolving nested file references.
+func NormalizeServiceConfigInPlace(svc *azdext.ServiceConfig) error {
```
trangevi, line 376: Can you please expand on the comments for both `ResolveServiceConfigInPlace` and `NormalizeServiceConfigInPlace` to give a bit more detail on what they are doing and why?

cli/azd/extensions/azure.ai.agents/internal/project/agent_definition.go
```diff
@@ -223,11 +311,187 @@ func LoadServiceTargetAgentConfig(svc *azdext.ServiceConfig) (*ServiceTargetAgen
...
+		projectRoot,
+	)
+	if err != nil {
+		return nil, fmt.Errorf(
+			"resolving service %q config: %w",
+			serviceName,
+			err,
+		)
+	}
+	if err := projectconfig.NormalizeEnvironment(resolved); err != nil {
```
trangevi: 422-437 look almost identical to what `normalizeServiceProps`, is there a way to have some code reuse?

cli/azd/extensions/azure.ai.agents/internal/cmd/listen.go
```diff
@@ -716,10 +738,47 @@ func setEnvVar(ctx context.Context, azdClient *azdext.AzdClient, envName string,
...
+		if err := project.ResolveServiceConfigInPlace(
+			svc,
+			projectRoot,
+		); err != nil {
+			return nil, fmt.Errorf(
+				"failed to resolve agent config: %w",
+				err,
+			)
+		}
+	} else if err := project.NormalizeServiceConfigInPlace(svc); err != nil {
```
trangevi, line 778: Do we not need to "Normalize" after the "Resolve"? This ties into my ask for additional clarity in the comments of these two methods, it's difficult (in my opinion) to understand what they are for and when they would be necessary

cli/azd/extensions/azure.ai.agents/internal/cmd/resource_services.go
```diff
@@ -419,24 +462,46 @@ func collectConnections(services map[string]*azdext.ServiceConfig) ([]project.Co
 // toolbox service carries any, so a pre-split azure.yaml still provisions
 // without re-running init.
 func collectToolboxes(services map[string]*azdext.ServiceConfig) ([]project.Toolbox, error) {
+	return collectToolboxesAtRoot(services, "")
```
trangevi: Same comment about the unnecessary wrapper

cli/azd/extensions/azure.ai.agents/internal/cmd/resource_services.go
```diff
@@ -468,6 +533,13 @@ func collectAgentToolConnections(services map[string]*azdext.ServiceConfig) ([]p
 // connections, and toolboxes here rather than in sibling azure.ai.<kind>
 // services, so the collectors fall back to these when no sibling service exists.
 func collectLegacyAgentConfigs(services map[string]*azdext.ServiceConfig) ([]*project.ServiceTargetAgentConfig, error) {
+	return collectLegacyAgentConfigsAtRoot(services, "")
```
trangevi: Same comment about the unnecessary wrapper

jongio (approved):
My earlier block was the container options passed through proto field 3, whose core reader ships in #9170 rather than released 1.28.0, so it would silently no-op on a 1.28.0 host. This revision drops that injection and the AddService round-trip, and relies on the service config declared in azure.yaml instead, so there's no dependency on an unreleased core reader.

The typed env change is applied to both the extension schema and the v1.0 service schema, and the new root `` core-field validation is now consistent between synthesis and run/deploy. Verified the affected extension packages pass locally (the only failures are the bicep/ARM template regeneration checks, which need the bicep toolchain and are unrelated to these changes).

cli/azd/extensions/azure.ai.agents/internal/project/agent_definition.go
```diff
@@ -167,18 +169,104 @@ func LoadAgentDefinition(
 	svc *azdext.ServiceConfig,
 	projectRoot string,
 ) (agent_yaml.ContainerAgent, bool, AgentDefinitionSource, error) {
-	ca, isHosted, found, source, err := AgentDefinitionFromService(svc)
```
trangevi, line 170: If we're changing this, does that mean we should remove `AgentDefinitionFromService` completely? It doesn't appear to be the same kind of wrapper like some of the other methods, but do we need to keep it?

cli/azd/extensions/azure.ai.agents/internal/project/agent_definition.go
```diff
@@ -223,11 +311,187 @@ func LoadServiceTargetAgentConfig(svc *azdext.ServiceConfig) (*ServiceTargetAgen
...
+	}
+	return resolveServiceProps(props, svc.GetName(), projectRoot)
+}
+
+// ResolvedServiceProjectPath returns the service source path.
+func ResolvedServiceProjectPath(
+	svc *azdext.ServiceConfig,
+	projectRoot string,
+) (string, error) {
+	_ = projectRoot
```
trangevi: What is the point of this? Why take the parameter if we do nothing with it?

cli/azd/extensions/azure.ai.agents/internal/project/agent_definition.go
```diff
@@ -329,6 +606,23 @@ func agentDefinitionFromStruct(s *structpb.Struct, coreImage string) (agent_yaml
 	}
 
 	if inline.Kind != agent_yaml.AgentKindHosted {
+		// Validate non-hosted kinds (e.g. workflow) with the same
```
trangevi: Do we support workflow agents in any way? What's the value of adding this handling now if we don't actually support anything else yet?

cli/azd/extensions/azure.ai.agents/internal/project/agent_definition.go
```diff
@@ -329,6 +606,23 @@ func agentDefinitionFromStruct(s *structpb.Struct, coreImage string) (agent_yaml
...
+			log.Printf(
+				"[debug] skipping non-hosted agent validation: %v",
+				marshalErr,
+			)
+		} else if err := agent_yaml.ValidateAgentDefinition(defBytes); err != nil {
+			return agent_yaml.ContainerAgent{}, false, exterrors.Validation(
+				exterrors.CodeInvalidAgentManifest,
+				fmt.Sprintf("agent service definition is not valid: %s", err),
+				"fix the agent service entry in azure.yaml or "+
+					"re-run `azd ai agent init`",
```
trangevi: Re-running `init`, depending on when this method is being called, may not be a clean solution to fixing problems. Especially if this method is called as a part of init, a user is likely to just end up back in this spot again even if a rerun does work well. As long as the returned error provides what is wrong with the service definition, it's probably cleanest to just leave this message with the guidance to fix the definition

cli/azd/extensions/azure.ai.agents/internal/synthesis/synthesizer.go
```diff
@@ -288,6 +301,15 @@ func Synthesize(in Input) (*Result, error) {
 // to learn which model deployments to create on the existing account. Returns
 // nil (not an error) when the service declares no deployments.
 func BrownfieldDeployments(raw []byte, serviceName string) ([]Deployment, error) {
+	return BrownfieldDeploymentsAtRoot(raw, "", serviceName)
```
trangevi: Same comment about the unnecessary wrapper

cli/azd/extensions/azure.ai.agents/internal/synthesis/synthesizer.go
```diff
@@ -321,6 +354,15 @@ func BrownfieldDeployments(raw []byte, serviceName string) ([]Deployment, error)
 // expressions pass through. Returns an empty slice (not an error) when none
 // are declared.
 func BrownfieldConnections(raw []byte, env map[string]string) ([]Connection, error) {
+	return BrownfieldConnectionsAtRoot(raw, "", env)
```
trangevi: Same comment about the unnecessary wrapper

trangevi (changes requested):
Generally looks good, I very much appreciate the work that's been done to simplify the changes we're making here. Most of my comment are effectively the same, regarding unnecessary wrapper methods, so should hopefully be easy to resolve in bulk.

jongio (approved):
Re-reviewed the delta since my last approval (rebased HEAD). The change drops the unreleased container options gRPC field and its core-side path handling, and instead requires `project`/`image`/`docker` to be declared in `azure.yaml`, rejecting them from a root `$ref`. `requiredAzdVersion` moves back to `>=1.27.1` as a result. Removing the proto field is safe since it only targeted the unreleased 1.28.0, and dropping the service-path validation isn't a regression because the extension-supplied override path it guarded is gone. The new validation, the Windows env-merge helper, and the schema value-type change all have tests, and CI is green. No blocking correctness or security concerns from me on the incremental changes.

jongio (approved):
Re-reviewed the delta since my last approval (rebased HEAD). The change drops the unreleased container options gRPC field and its core-side path handling, and instead requires `project`/`image`/`docker` to be declared in `azure.yaml`, rejecting them from a root `$ref`. `requiredAzdVersion` moves back to `>=1.27.1` as a result. Removing the proto field is safe since it only targeted the unreleased 1.28.0, and dropping the service-path validation isn't a regression because the extension-supplied override path it guarded is gone. The new validation, the Windows env-merge helper, and the schema value-type change all have tests, and CI is green. No blocking correctness or security concerns from me on the incremental changes.

jongio (approved):
Re-reviewed the delta since my last approval (rebased HEAD). The change drops the unreleased container options gRPC field and its core-side path handling, and instead requires `project`/`image`/`docker` to be declared in `azure.yaml`, rejecting them from a root `$ref`. `requiredAzdVersion` moves back to `>=1.27.1` as a result. Removing the proto field is safe since it only targeted the unreleased 1.28.0, and dropping the service-path validation isn't a regression because the extension-supplied override path it guarded is gone. The new validation, the Windows env-merge helper, and the schema value-type change all have tests, and CI is green. No blocking correctness or security concerns from me on the incremental changes.

tg-msft:
I think we should consider Victor's comment.  Scoping it now might save us a lot of headache down the road.

jongio (approved):
The new commit folds the AtRoot collector variants back into the base functions, drops the populateContainerSettings and ResolvedServiceProjectPath pass-throughs, and threads projectRoot through explicitly. That covers the wrapper-method feedback from the prior review.

AgentDefinitionUsesFileRef now resolves only the top-level `` before checking for `kind`, so a service that carries an inline definition alongside an unrelated `` no longer reports as file-backed. The added test covers that case.

jongio (approved):
Follow-up to my earlier approval to correct formatting: the new commit folds the AtRoot collector variants back into the base functions, drops the populateContainerSettings and ResolvedServiceProjectPath pass-throughs, and threads projectRoot through explicitly. That covers the wrapper-method feedback from the prior review.

AgentDefinitionUsesFileRef now resolves only the top-level `$ref` before checking for `kind`, so a service that carries an inline definition alongside an unrelated `$ref` no longer reports as file-backed. The added test covers that case.

jongio:
On the `service_path`/`options` proto discussion: the latest commit (simplify unified configuration resolution) looks like it already sidesteps this. `container.proto` on this branch has no `service_path` field on `ContainerBuildRequest`/`PackageRequest`/`PublishRequest`, so the container RPCs aren't carrying any resolved config across the boundary right now.

The extension now reads the path locally instead of sending it over gRPC. `ResolvedServiceProjectPath` was removed (it was just a passthrough to `svc.GetRelativePath()`), and the callers in `optimize_apply.go` and `service_target_agent.go` now call `svc.GetRelativePath()` directly.

So there's no `service_path` field to reshape into an `options` bag in this PR. If a future `$ref`'d file ever needs to feed `Docker`/`Image` into core's container pipeline, the grouped `options` shape is the right call at that point, but as of this commit there's no proto field to migrate, so nothing is locked in.

jongio (commented):
The latest automated review batch (2026-07-17) is pointing at files that aren't in this PR's net diff: `cli/version.txt`, `pkg/azdext/validation_provider.go`, `cmd/invoke.go`, and `cmd/init_adopt.go`. Those came in from merging main (the provision-validation rename from #8844, the invoke-name change from #9114, and the release version bumps), so they don't show up in `gh pr diff`. Nothing to change there for this PR, the net diff is 23 files, all under the agents extension.

The substantive open items are still trangevi's threads: the Resolve/Normalize wrapper methods, and the request to expand the doc comments on `ResolveServiceConfigInPlace` and `NormalizeServiceConfigInPlace`.

cli/azd/extensions/azure.ai.agents/internal/cmd/listen.go
huimiu: Good question. `ResolveServiceConfigInPlace` already includes normalization, so calling `NormalizeServiceConfigInPlace` afterward would be redundant. I updated the comments to clarify this.

cli/azd/extensions/azure.ai.agents/internal/project/agent_definition.go
huimiu: We still need this for callers that only work with inline configuration, such as `optimize apply`. Runtime paths use the resolved version instead.

jongio (commented):
One note on the latest commit (`fix: remove unsupported workflow handling`): dropping the non-hosted branch validation in `agentDefinitionFromStruct` leaves the inline path inconsistent with the on-disk path. `parseContainerAgentYAML` still calls `ValidateAgentDefinition` before it branches on kind, so an invalid `kind` or a malformed `name` on a `workflow` agent is rejected when it lives in agent.yaml on disk, but silently accepted when it is inline in azure.yaml (`agent_definition.go:596`).

Since the goal of this PR is resolving these forms consistently across run/deploy/provision, is skipping validation on the inline non-hosted path intentional, or should it validate like the disk path does? Hosted agents are unaffected either way, so not a blocker.

jongio (commented):
Ran the reworked commits (9ce8720b..d6edea10) through the agents extension build, vet, and package tests locally. All green except the two synthesis ARM-template snapshot tests (`TestARMTemplate_MatchesBicepBuild` and the brownfield variant), and those only fail because the committed `main.arm.json`/`brownfield.arm.json` are stale against my local bicep version, which is unrelated to this change (no `.bicep`/`.arm.json` files are in the diff). The container-config rework also resolves the earlier version-gate concern, since the resolved service path is now read locally instead of carried over the container RPC.

Not approving yet: the maintainability threads above (the wrapper methods with no conditional handling, and the doc clarity + DRY on `ResolveServiceConfigInPlace`/`NormalizeServiceConfigInPlace`) are still open and worth settling first.

huimiu:

jongio (commented):
The automated conflict resolution in the latest commit (f8ebf63, "begin merge conflict resolution with main") is incomplete. Unresolved conflict markers (`<<<<<<<`, `=======`, `>>>>>>>`) are still in 7 files, so both extensions fail to compile and CI is red.

Files with leftover markers:
- `azure.ai.agents/internal/cmd/listen.go` (2)
- `azure.ai.agents/internal/cmd/resource_services.go` (1)
- `azure.ai.agents/internal/cmd/resource_services_test.go` (5)
- `azure.ai.agents/internal/synthesis/synthesizer.go` (9)
- `azure.ai.agents/internal/synthesis/synthesizer_test.go` (6)
- `azure.ai.projects/internal/provisioning/foundry_provisioning_provider.go` (6)
- `azure.ai.projects/internal/provisioning/resource_group_location_check.go` (1)

`go build ./...` fails in both extensions with syntax errors right at the marker lines (for example `synthesizer.go:277: unexpected <<`).

A few of these conflicts aren't just formatting, so it's worth a careful pass when finishing them. In `synthesizer.go`, `deriveIncludeAcr` has two different signatures across the sides: HEAD returns `(bool, error)` and propagates the error, origin/main returns a plain `bool`. The call site (~line 278) and the definition (~line 604) have to agree, and the loop bodies differ too (`serviceForHost` vs `resolveServiceRefs`). Same story with the `collectConnections` argument order around line 297. Pick the version that matches the rest of the current code, then run `go build` and `go test` on both extensions before the next pass.

jongio (commented):
This branch is mid-merge with main and still has unresolved conflict markers, so the agents and projects extensions don't compile at this HEAD. I found `<<<<<<<` / `=======` / `>>>>>>>` markers in:

- cli/azd/extensions/azure.ai.agents/internal/cmd/listen.go
- cli/azd/extensions/azure.ai.agents/internal/cmd/resource_services.go
- cli/azd/extensions/azure.ai.agents/internal/cmd/resource_services_test.go
- cli/azd/extensions/azure.ai.agents/internal/synthesis/synthesizer.go
- cli/azd/extensions/azure.ai.agents/internal/synthesis/synthesizer_test.go
- cli/azd/extensions/azure.ai.projects/internal/provisioning/foundry_provisioning_provider.go
- cli/azd/extensions/azure.ai.projects/internal/provisioning/resource_group_location_check.go

The latest commit is "begin merge conflict resolution with main," so this is in-progress. My earlier approval doesn't apply to this state. Once the merge is finished and `go build ./...` plus the tests pass for both extensions, re-request review and I'll take another pass at the actual logic changes.

jongio (commented):
Re-reviewed after the automated merge with main. The remaining blocker here isn't the merge itself. The conflict resolution in 685ecc3 came through clean: no leftover conflict markers, the `projectRoot` parameter-reorder conflicts (moved to the end of the signature) resolve consistently across every call site, and CI is green on both the agents and projects extension pipelines.

What still needs to land before this is mergeable is the open review feedback, not the merge state: trangevi's request to drop the unnecessary wrapper methods, and the scoping concern Victor raised that tg-msft backed. Those are worth resolving in this PR rather than deferring.

cli/azd/extensions/azure.ai.agents/internal/cmd/listen.go
```diff
@@ -760,19 +805,15 @@ func populateContainerSettings(ctx context.Context, azdClient *azdext.AzdClient,
...
-	// Persist the resolved container settings back onto the service's inline
-	// properties, preserving the agent definition and other config keys.
-	if err := project.SetAgentContainerSettings(svc, &project.ContainerSettings{Resources: result}); err != nil {
-		return fmt.Errorf("failed to update agent container settings: %w", err)
-	}
-
-	// Need to add the service config back to the project for use further down the pipeline
-	req := &azdext.AddServiceRequest{Service: svc}
-
-	if _, err := azdClient.Project().AddService(ctx, req); err != nil {
```
trangevi, line 772: This is no longer necessary?

trangevi (approved):

cli/azd/.vscode/cspell.yaml
```diff
@@ -82,6 +82,8 @@ words:
   - protoimpl
   - protojson
   - protoreflect
+  - protowire
```
trangevi: This should be moved into the extension specific cspell, will remove the need for azd core codeowners approval

trangevi (changes requested):

jongio (approved):
Traced the unified config resolution through the run, deploy, provision, synthesis, and brownfield paths, plus the Windows env precedence and case-insensitive key handling. The $ref/inline/legacy-config ordering is consistent across the paths, the clone before in-place resolution avoids mutating the shared service, and the removed helpers are compensated elsewhere. Env precedence keeps process env winning over configured env with case-insensitive matching on Windows, and it's covered by tests. CI is green.

cli/azd/extensions/azure.ai.agents/internal/cmd/listen.go
huimiu: Yes, this is no longer needed. `AddService` would write the resolved $ref config back to azure.yaml. We now keep it in memory, the service target resolves it and applies defaults at deploy time.

cli/azd/.vscode/cspell.yaml
huimiu: Thanks for catching this! I overlooked it when cleaning things up

jongio (approved):
Re-reviewed against the latest HEAD. The only change since my prior approval is scoping the agent spelling words into the extension cspell config, and cspell-lint confirms it. No new issues in the delta.

jongio (approved):
Re-reviewed against 3e50050 after the merge with main and the follow-up commits. The two open items from the last round are handled: the spelling word moved into the agents extension `cspell.yaml` so no core codeowners approval is needed, and the deploy path no longer writes resolved `$ref` config back to `azure.yaml`. `prepareContainerSettings` resolves it in memory via `ResolveServiceConfigInPlace` and applies defaults at deploy time, which matches the reply on that thread.

The env normalization keeps scalars only and rejects maps or slices, `optimize apply` still reads and restores `uses` around the `AddService` call, and dropping the container `service_path` proto field means there's no new RPC surface to reason about. CI is green across the agents and projects extensions.

jongio (approved):
Re-reviewed the unified runtime config changes since my last pass (the resolve/normalize path in agent_definition.go, service_target_agent.go, run.go, listen.go, and the new projectconfig/environment.go). Env precedence reads consistently across run and deploy: the unified azure.yaml service env is applied last and wins, falling back to the service config env when absent. Dropping the AddService write-back in prepareContainerSettings looks right for preserving $ref semantics, and the DefaultCpu/DefaultMemory fallback in prepareDeploy keeps $ref and inline services aligned. Test coverage on the new paths is solid. Looks good.

jongio (approved):
Re-reviewed the delta since my last approval: the merge-with-main resolution, the workflow-handling removal in `agentDefinitionFromStruct`, the schema reverts, and moving the spelling entry into the extension's own cspell. No conflict markers remain in any of the changed files and the extension builds green in CI, so the conflict resolution landed cleanly. The workflow removal is self-consistent, both inline-workflow tests are dropped and the `log`/`yaml` imports are still used. The root-`$ref` core-field guard and the scope tightening are already being tracked in the open thread, nothing blocking from my side.

trangevi (approved):

jongio (approved):
Re-reviewed against the current HEAD. The unified config resolution path (services.<name>.env, root $ref core-field validation, and the shared env precedence: agent definition < config < unified service env, with process env winning) is consistent between run and deploy. My earlier concerns are resolved. CI is green.

PR #9150, opened 2026-07-15
fix: remove hosted-agent deploy session carryover

huimiu:
## Why

Hosted-agent deployments should not retain a live agent session across versions. Removing this unrelated deploy behavior prevents a post-deploy invocation from resuming a prior session.

## Approach

Removed the predeploy session capture and postdeploy session resume hooks while preserving the existing deployment setup, RBAC checks, activity bot handling, and optimization reporting.

### DIFF
--- a/cli/azd/extensions/azure.ai.agents/internal/cmd/listen.go
+++ b/cli/azd/extensions/azure.ai.agents/internal/cmd/listen.go
@@ -15,7 +15,6 @@ import (
 	"sync"
 
 	"azureaiagent/internal/exterrors"
-	"azureaiagent/internal/pkg/agents/agent_api"
 	"azureaiagent/internal/pkg/agents/optimize_api"
 	"azureaiagent/internal/project"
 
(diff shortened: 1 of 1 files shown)


jongio (commented):
Two things look like they got left behind by this removal:

1. `session_carryover.go` and `session_carryover_test.go` are now dead. `captureSessionForCarryover` and `carryOverSessionAfterDeploy` have no production callers after this change, and the rest of that file only exists to support those two. Go doesn't flag unused package functions, and the tests keep the unused-code linters quiet, so CI stays green even though none of it runs anymore. Is the plan to keep the plumbing for a follow-up, or should that file and its tests be removed with this change?

2. The README still has the "Session carry-over across deploys" section telling users to `export AZD_AGENT_RESUME_SESSION_ON_DEPLOY=true`. After this PR that variable is inert (`sessionCarryoverEnabled()` is only read from the now-dead functions), so the docs describe behavior that no longer exists. Worth updating or dropping that section in the same PR.

PR #9160, opened 2026-07-15
(DO NOT MERGE) Updating capabilities in a non-simple way.

richardpark-msft:


richardpark-msft:
Perfect, failure was expected since I did a non-simple change: https://github.com/Azure/azure-dev/actions/runs/29450244109/job/87470654819?pr=9160

richardpark-msft:
Second one worked, since I made it a simple update: https://github.com/Azure/azure-dev/actions/runs/29450328687/job/87470930514?pr=9160

richardpark-msft:
And the last type - a non-registry related change runs the workflow but it exits immediately: https://github.com/Azure/azure-dev/actions/runs/29450416355/job/87471216394?pr=9160

jongio (commented):
This is a CI trigger experiment - the only change is an empty `cli/azd/dummy.txt`, so there's no code or workflow logic to review on the diff itself. Flagging just one thing for cleanup: since this is marked DO NOT MERGE, make sure the branch is closed or the dummy file is dropped before any real change lands on it, so the empty file doesn't leak into the tree.

vhvb1989 (approved):

cli/azd/extensions/registry.json
```diff
@@ -536,6 +536,119 @@
...
+              "description": "Start MCP server with demo tools.",
+              "usage": "azd demo mcp start"
+            },
+            {
+              "name": "copilot",
+              "description": "Interactive Copilot chat loop demonstrating CopilotService gRPC API.",
+              "usage": "azd demo copilot [--model <model>] [--resume]"
+            }
+          ],
+          "artifacts": {
```
jongio, line 489: All of these 0.8.0 artifact URLs point at the `azd-ext-microsoft-azd-demo_0.7.0` release, so installing 0.8.0 would download the 0.7.0 binaries. Together with the capability list being identical to 0.7.0, this 0.8.0 entry is a copy of 0.7.0 with only the version bumped. For a real capability update, the artifact URLs and checksums need to point at an actual 0.8.0 release.

jongio (commented):
One thing on this new 0.8.0 entry beyond the duplicate capability list already flagged: the artifacts point at the old release, so it would ship the wrong binaries. Details inline.

cli/azd/extensions/registry.json
richardpark-msft: You are correct. But what does your meta-awareness tell you about this PR?

jongio: Meta-awareness: this reads as a deliberate probe of how automated review reacts to PR interactions rather than a real change. The `azd-check-pr-interactions` branch name and the DO NOT MERGE title give it away, and HEAD has since been force-pushed back to just an empty `cli/azd/dummy.txt`. So the `registry.json` issue I flagged is gone from the current diff and there is nothing left to review on this commit.

If the 0.8.0 registry entry comes back, the two things that make it real: the artifact URLs still point at the `azd-ext-microsoft-azd-demo_0.7.0` release, and the capability list is a duplicate of 0.7.0. Fix those before it lands.

richardpark-msft: Is there any way to tell you to ignore a PR? It's just wasting tokens to review these.

jongio: Marking it a draft is the cleanest way to have me skip it, and closing it once the experiment's done also works. Either way, I'll leave this one alone from here.

PR #9161, opened 2026-07-15
Improve Windows test reliability

### LINKED ISSUE #9162: Improve Windows pipeline test reliability
### Description

The Windows test job in internal ADO build [6567512](https://dev.azure.com/azure-sdk/internal/_build/results?buildId=6567512&view=logs&j=12494761-d332-5a1c-a5ee-2bea7198f339&t=e8c28b77-9b2d-5977-b9e2-564f8e6b9e54) exposed several related reliability problems that amplified one initiating failure into many reported failures.

### Findings

- `Test_CLI_Extension_ForceInstall` failed while replacing `microsoft.azd.demo` because Windows still held the extension executable open: `failed to remove extension: ... Access is denied`.
- The failed cleanup left the demo extension installed in shared user configuration. Later quota, telemetry, and storage tests then received an unexpected provisioning warning from that extension and entered prompt paths they did not expect.
- `Test_CLI_Up_Down_FuncApp` used unchecked fixture type assertions, so missing environment values caused a panic that aborted the remaining functional-test package.
- Several command tests completed successfully but were reported as failures because spinner and cursor escape sequences were written directly to stdout and corrupted the `go test -json` stream.

### Expected behavior

- Extension replacement and cleanup should tolerate transient Windows file locks without test-specific sleeps.
- Functional tests that modify extensions or user configuration should use isolated state and leave no contamination for later tests.
- Prompt and spinner output should honor injected writers so test JSON remains machine-readable.
- Missing fixture values should produce focused assertion failures rather than package-wide panics.

### Proposed fix

Draft PR #9161 addresses these findings with shared filesystem retries, isolated functional-test configuration, writer propagation, guarded assertions, and contributor guidance.

### LINKED ISSUE #9181: Nightly `azure-dev - cli` red: phantom unit-test failures from tool.go spinners writing to os.Stdout (regression from #9045)
## Summary

The nightly scheduled `azure-dev - cli` pipeline (ADO `azure-sdk/internal`, definition **4643**) is failing every night due to **7 deterministic phantom unit-test failures** in the `cmd` and `cmd/middleware` packages. The tests actually **pass**, but a spinner writes to `os.Stdout` during the test, corrupting the `go test -json` event stream that `gotestsum` consumes, so they are reported as `(unknown)` → failure. This is the same class of failure as #8385.

## Failing tests (identical every run, all platforms — Windows/Mac)

- `cmd TestToolInstallAction_resolveToolIds_AllFlag`
- `cmd TestToolInstallAction_resolveToolIds_NoPromptWithoutTarget_Errors`
- `cmd TestToolUpgradeAction_AllFlag_NoPrompt`
- `cmd TestToolUpgradeAction_NoPrompt_WithoutTarget_Errors`
- `cmd TestToolUninstallAction_resolveToolIds_AllFlag_NoPrompt`
- `cmd TestToolUninstallAction_resolveToolIds_NoPromptWithoutTarget_Errors`
- `cmd/middleware TestToolFirstRunMiddleware_OfferInstall_PromptError`

Log signature (each test PASSES but is marked FAIL):

```
=== FAIL: cmd TestToolInstallAction_resolveToolIds_AllFlag (unknown)
| Detecting tool status...
--- PASS: TestToolInstallAction_resolveToolIds_AllFlag (0.00s)
```

## When it started

Regression introduced by **#9045** ("UX: implement new design of azd tool (install/upgrade/list/check)"), merged **2026-07-14 17:51 UTC** (commit `197a3aec`). The first nightly to include it (2026-07-14) and every nightly since have been red on these tests.

## Why it happens (root cause)

In `cli/azd/cmd/tool.go`, five `uxlib.NewSpinner(...)` calls omit the `Writer` option, so the spinner defaults to `os.Stdout` (`cli/azd/pkg/ux/spinner.go:43`):

- L206 `"Detecting tools..."`
- L374 `"Checking tool status..."`
- L907 `detectAllTools` (`"Detecting tool status..."` / `"Detecting installed tools..."`) — hit by the `resolveToolIds` tests
- L1920 `"Checking for upgrades..."`
- L2118 `"Checking %s..."`

During the unit tests these spinners write raw ANSI to `os.Stdout` (fd 1), bypassing `testing`'s output capture and corrupting the parallel `go test -json` stream that `gotestsum` parses. The result is phantom `(unknown)` failures on whichever `cmd`/`cmd/middleware` tests are emitting at that moment.

The correct pattern already exists in `cli/azd/cmd/middleware/tool_first_run.go`, where every spinner sets `Writer: m.console.Handles().Stdout` (test-safe buffer under test). The five `tool.go` spinners were not given a `Writer`.

This is explicitly called out in `cli/azd/AGENTS.md` ("Never write to `os.Stdout` in tests") and is meant to be guarded by the `forbidigo` linter — but that linter is scoped to `*_test.go` only, so a spinner created in non-test code (`tool.go`) and exercised by a test slips through.

## Why the nightly retry fix (#9118) does not help

#9118 added `gotestsum --rerun-fails` to the **integration** run only:

1. These 7 failures are **unit** tests, which have no retry — so they always red the run.
2. They also defeat the integration retry: 7 phantom failures + a couple of real live-Azure flakes exceed `--rerun-fails-max-failures=5`, so gotestsum skips **all** reruns (`ERROR number of test failures (9) exceeds maximum (5)`), meaning even genuinely flaky integration tests never get retried.

## Proposed fix

Route all five `tool.go` spinners through the console writer instead of `os.Stdout` (pass the `input.Console` into `detectAllTools` and set `Writer: console.Handles().Stdout`), matching `cmd/middleware/tool_first_run.go`. That clears the 7 phantom failures; with the count back under 5, #9118's retry can also do its job on real integration flakes.

Follow-up worth considering: extend `forbidigo`/lint coverage (or a `Writer`-required lint) to catch spinners created in non-test code that default to `os.Stdout`.

## Evidence

- Latest failed nightly: ADO build **6567394** (2026-07-15), tasks "Test Go Binary" on BuildCLI Windows / Mac (Apple Silicon).
- Related: #8385 (phantom failures), #8387 / #9028 (prior phantom fixes), #9118 (integration retry).

JeffreyCA:
Fixes #9162

This PR improves Windows test reliability by addressing the initiating extension cleanup race and the failures it caused in later tests.

### Issue

Windows can briefly retain a lock on an extension executable after the process exits. The force-install test then failed to replace the extension, leaked global extension state, and caused unrelated tests to fail or prompt unexpectedly.

- Initiating error: `failed to remove extension: ... Access is denied`
- UX components could also write terminal control bytes directly to stdout, corrupting the `go test -json` stream.
- Unchecked fixture type assertions could panic and abort the remaining functional-test package.

### Changes

- Retry transient Windows sharing and access violations at the shared filesystem boundary.
- Isolate extension functional tests with a private azd configuration directory and resilient cleanup.
- Route spinner, prompt, and cursor output through injected writers.
- Replace panic-prone fixture assertions with focused test failures.
- Document the reliability patterns for future command and functional-test work.

### Testing

Covered the Windows lock retry, extension force-install flow, command JSON-stream integrity, UX writer behavior, and functional-test compilation with focused package and functional tests.


vhvb1989 (approved):
Approving. Verified locally:

- **Root-cause fix is correct and complete.** All five `cli/azd/cmd/tool.go` spinners now route through an injected `io.Writer` instead of defaulting to `os.Stdout`, and `detectAllTools` takes the writer from each action. This is the same fix I had drafted in #9184 (now closed in favor of this broader PR), and it also covers `extension.go` and the `pkg/ux` prompt/select/confirm paths.
- **The 7 phantom failures from #9181 pass** on this branch (`cmd` + `cmd/middleware`), and the spinner text no longer leaks into the `go test -json` stream.
- **Full CI is green**, including the `azure-dev - cli` Windows/Mac/Linux ADO pipeline that these phantom failures were reddening.

The remaining Copilot review comments are test-coverage suggestions (buffer-backed writer-propagation tests) rather than blocking defects — worth a follow-up but not gating. The Windows lock-retry cancellation note on `extensions/manager.go` is a reasonable tradeoff for the bounded retry window.

Also resolves #9181.

cli/azd/cmd/tool.go
```diff
@@ -1100,7 +1107,7 @@ func (a *toolInstallAction) resolveToolIds(ctx context.Context) ([]string, error
 	}
 
 	// Interactive: let the user pick from uninstalled tools.
-	statuses, err := detectAllTools(ctx, a.manager, a.formatter, "Detecting tool status...")
+	statuses, err := detectAllTools(ctx, a.manager, a.formatter, a.writer, "Detecting tool status...")
```
richardpark-msft, line 1110: There has to be some point where you're like "maybe I'll just pass `a`"

cli/azd/cmd/tool_test.go
```diff
@@ -330,6 +330,106 @@ func TestToolShowAction_InvalidArgDoesNotEmitToolId(t *testing.T) {
...
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			var writer bytes.Buffer
+			err := test.run(t.Context(), &writer)
+
+			require.ErrorIs(t, err, detectErr)
+			assert.Contains(t, writer.String(), "\033[?25l")
```
richardpark-msft: Can you add a comment on here, telling me what these character sequences mean to you?

richardpark-msft: (same below)

cli/azd/docs/recording-functional-tests-guide.md
```diff
@@ -1056,8 +1066,3 @@ os.Setenv("AZD_TEST_FIXED_CLOCK_UNIX_TIME", "1744738873")
 - **Test Helpers**: `cli/azd/test/azdcli/`
 - **Example Tests**: `cli/azd/test/functional/up_test.go`
 - **go-vcr Documentation**: https://github.com/dnaeon/go-vcr
-
----
-
-**Last Updated**: December 2025
```
richardpark-msft, line 1062: Agree on this one...leave no evidence of the last time we edited this...

cli/azd/docs/recording-functional-tests-guide.md
```diff
@@ -1022,10 +1030,12 @@ os.Setenv("AZD_TEST_FIXED_CLOCK_UNIX_TIME", "1744738873")
 - [ ] Start with `recording.Start(t)`
 - [ ] Use `randomOrStoredEnvName(session)`
 - [ ] Create CLI with `azdcli.WithSession(session)`
+- [ ] Use a private `AZD_CONFIG_DIR` for user-level state changes; authenticate it if live Azure access is required
 - [ ] Use `session.ProxyClient` for HTTP operations
 - [ ] Store dynamic values in `session.Variables`
-- [ ] Add cleanup with session checks
+- [ ] Use `tempDirWithDiagnostics(t)` and register other cleanup immediately
 - [ ] Use appropriate timeouts for recording/playback
+- [ ] Guard fixture type assertions so malformed data fails the test without panicking
```
richardpark-msft, line 1054: Can you elaborate on this? Maybe just embed a small code sample - I'm not quite sure what this means.

cli/azd/docs/recording-functional-tests-guide.md
```diff
@@ -1022,10 +1030,12 @@ os.Setenv("AZD_TEST_FIXED_CLOCK_UNIX_TIME", "1744738873")
 - [ ] Start with `recording.Start(t)`
 - [ ] Use `randomOrStoredEnvName(session)`
 - [ ] Create CLI with `azdcli.WithSession(session)`
+- [ ] Use a private `AZD_CONFIG_DIR` for user-level state changes; authenticate it if live Azure access is required
```
richardpark-msft: Given we're in (what I assume) is just a normal test, are you saying you actually need to set this before running the entire test suite (ie, in the shell), or set it before your first call to azd (ie, t.Setenv())?

cli/azd/pkg/osutil/rename_unix.go
```diff
@@ -14,3 +14,8 @@ import (
 func Rename(ctx context.Context, old, new string) error {
 	return os.Rename(old, new)
 }
+
+// RemoveAll removes path and any children it contains.
+func RemoveAll(_ context.Context, path string) error {
+	return os.RemoveAll(path)
```
richardpark-msft, line 21: It'd be good to leave a comment as to why we take a context parameter, only to ignore it (and I assume the explanation is - Windows needs it!)

cli/azd/pkg/osutil/rename_windows_test.go
```diff
@@ -61,3 +64,31 @@ func TestRename(t *testing.T) {
...
 	})
 }
+
+func TestRemoveAll_FileInUse(t *testing.T) {
+	dir := t.TempDir()
+	file, err := os.Create(filepath.Join(dir, "locked"))
+	require.NoError(t, err)
+
+	go func() {
+		time.Sleep(time.Second)
```
richardpark-msft, line 74: It's probably okay, in practice...but sometimes I get a little nervous with trying to coordinate with sleeps, although in this case, I don't see a great alternative.

cli/azd/pkg/ux/input_writer_test.go
```diff
@@ -0,0 +1,83 @@
...
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			ctx, cancel := context.WithCancel(t.Context())
+			cancel()
+
+			var writer bytes.Buffer
+			err := test.ask(ctx, &writer)
+
+			require.Error(t, err)
+			assert.Equal(t, test.expectedShowCursorCount, strings.Count(writer.String(), "\033[?25h"))
```
richardpark-msft: Similar comment as what I left earlier - maybe just extract these characters into a constant, so it's obvious what they are.

docs/guides/adding-a-new-command.md
```diff
@@ -88,6 +88,13 @@ After adding the command, regenerate CLI snapshots:
 UPDATE_SNAPSHOTS=true go test ./cmd -run 'TestFigSpec|TestUsage'
 ```
 
+### 5. Check reliability boundaries
```
richardpark-msft, line 91: Can you check if our review skill sees these docs? If not, can you add in a line so it checks them?

richardpark-msft (approved):
Look, I saw some stuff, I read some stuff, I liked it.

But some feedback, and just a recommendation to make sure our review skill is picking up the stuff you've added to our guides. (I think it does, but worth a double-check)

cli/azd/cmd/tool.go
JeffreyCA: `detectAllTools` is shared by three different action types, so there's no single `a` to pass. Kept it taking just the three fields it needs.

cli/azd/cmd/tool_test.go
JeffreyCA: Extracted the sequences into named `ansiHideCursor` / `ansiShowCursor` constants with a doc comment explaining they're the cursor hide/show control codes the spinner emits. Covers both spots.

cli/azd/docs/recording-functional-tests-guide.md
JeffreyCA: Done - removed the "Last Updated" footer.

cli/azd/docs/recording-functional-tests-guide.md
JeffreyCA: Added a code sample in Section 7 showing the `require.True(t, ok, ...)` guard pattern. Kept the checklist bullet terse (like its siblings) and let the section carry the explanation.

cli/azd/docs/recording-functional-tests-guide.md
JeffreyCA: Neither shell-wide nor `t.Setenv` - set it on the test CLI's env before the first `azd` invocation. Clarified the wording and added a sample: `cli.Env = append(cli.Env, "AZD_CONFIG_DIR="+configDir)`.

cli/azd/pkg/osutil/rename_unix.go
JeffreyCA: Added a comment: it's retained for API parity with the Windows implementation, where the context cancels the retry backoff.

cli/azd/pkg/ux/input_writer_test.go
JeffreyCA: Done - added the `ansiShowCursor` constant (with a comment) here too.

docs/guides/adding-a-new-command.md
JeffreyCA: Wired the guidance into the review skill directly: `reviewers.md` now has criteria for state isolation (`AZD_CONFIG_DIR`), writer propagation, and guarding fixture type assertions, so it enforces the guide content rather than relying on it reading the docs at review time.

cli/azd/docs/recording-functional-tests-guide.md
richardpark-msft: Should this just always be done? For tests it seems like ensuring clean state is just good, overall.

.github/skills/azd-code-reviewer/reviewers.md
```diff
@@ -227,6 +229,7 @@ Focus areas:
 - Help text: is `--help` complete, with examples?
 - Interactive prompts: are they skippable with flags for CI/automation?
 - Progress feedback: for long operations, is there progress indication?
+- Writer propagation: do nested formatters, spinners, prompts, and cursors use the injected writer without leaking terminal output to global stdout?
```
richardpark-msft, line 232: It'd be nice to have something, in code, that enforces this instead of requiring a loose guideline. This isn't a "this PR" thing, but brainstorm a bit, since you're in this space.

richardpark-msft (approved):
Added a "future thought" comment, but otherwise looks good. Thanks for addressing my feedback :)

cli/azd/docs/recording-functional-tests-guide.md
JeffreyCA: It's not the case right now so I'm hesitant to recommend that for all future functional tests

richardpark-msft: If you were just deciding, though, what do you think? Is there a big overhead towards having a separate config dir per test?

cli/azd/cmd/tool.go
vhvb1989: or create a newDetector that resolves its dependencies during the container initialization and is injected for the tool having a formatter and writter already. 

This is fun and nostalgic... LOL  going to the conversations all raw and manual - tks non-bot @richardpark-msft

.github/skills/azd-code-reviewer/reviewers.md
vhvb1989: We have a linter but only works for _tests

We can make the writter to be a required argument from the method to enforce a writer is passed; or migrate the entire method to use a manager.  It's worth checking if that would make sense

cli/azd/cmd/tool.go
richardpark-msft: So the general problem, as I get from this PR, is that we need to pass around _more_ stuff (writer and formatter) in order to maintain consistency in console output, so it might make sense to encapsulate just those two values together, since they're basically paired (this might be untrue!):

So something like:

```go
type Output struct {
   Formatter output.Formatter
   Writer io.Writer
}
```

Then any spot that you were previous storing two fields (like 'a', above), you just embed the Output struct instead:

```go
type toolInstallAction struct {
	args      []string
	flags     *toolInstallFlags
	manager   *tool.Manager
	console   input.Console
        Output   // <-- embed
}
```

This isn't just to make detectAllTools() call simpler, but more to say "these things are actually a unit, and passing one without the other is probably a bug"

.github/skills/azd-code-reviewer/reviewers.md
richardpark-msft: Can you point me to the test linter you're talking about? (also, resolving comment, as we're out of scope for this PR)

jongio (approved):
Verified locally on Windows: `go build` and `go vet` clean for the touched packages (osutil, ux, extensions, cmd).

The core reliability changes hold up:

- `osutil.RemoveAll` shares the retry backoff with `Rename` and threads context, so cancellation stops the loop instead of spinning for 10 seconds. The Unix side is a straight passthrough with the context kept for parity.
- `Manager.Uninstall(ctx, ...)` drops the `MkdirAll`-then-remove dance; `os.RemoveAll` on a missing dir is already a no-op, so behavior is unchanged and the extension dir cleanup now retries transient Windows locks.
- Writer propagation reaches the nested cursor (`NewInput(w)` -> `NewCursor(w)`, stdout fallback when unset), and the new writer tests actually exercise that path rather than the framework default.
- The functional-test isolation via a private `AZD_CONFIG_DIR` doesn't drop any pre-set env, and the guarded fixture assertions fail the test instead of panicking the package.

No blocking issues from my pass.

hemarina (approved):
**Overall: LGTM** ✅

Nicely scoped reliability fix.

**Strengths**
- `retryFileSystemOperation` cleanly unifies the `Rename`/`RemoveAll` retry logic and correctly honors context cancellation (cancel → 1 attempt → `context.Canceled`).
- Threading `ctx` through `Manager.Uninstall` addresses the earlier cancellation concern — the retry window is now interruptible on the `uninstall`/`upgrade` paths.
- `Manager.Uninstall` simplification is sound: dropping the `MkdirAll`-then-remove and the `os.Stat` guard is equivalent since `os.RemoveAll` no-ops on missing paths.
- Writer propagation (`internal.NewInput(writer)` + spinner `Writer`) is the right fix; `NewInput` safely falls back to `os.Stdout` on a nil writer, so no regression. Tool-action wiring is consistent and `io.Writer` is already IoC-resolved.
- Replacing raw fixture type assertions with `require`-guarded casts is a good defensive change.
- Isolating `AZD_CONFIG_DIR` in the force-install test is correct and safe (`NewCLI(t)` leaves `cli.Env` empty without a session, so the env is equivalent + the two new vars).

**Minor (non-blocking)**
- Diagnostics-only: the old `removeAllWithDiagnostics` called `logHandles` *during* retries (~2s in); `osutil.RemoveAll` doesn't, so on a stuck lock the handle snapshot now only prints once at the end (~10s later), by which point the offending process may have exited. Consider a mid-retry `logHandles` hook (or a debug log) if the "what holds the lock" signal is still valuable on CI. Not required.

No correctness, security, or logic issues found.

cli/azd/docs/recording-functional-tests-guide.md
JeffreyCA: It makes sense yes, but there are some current limitations with auth handling that I think are out of scope, so I'm softening the guidance a bit

cli/azd/cmd/tool.go
JeffreyCA: Yes while that makes sense it would also introduce a new pattern not found elsewhere, and updating it everywhere would be a much larger refactor than expected. I'd like to consider this out of scope for now

jongio (approved):
The follow-up commit is docs-only. It updates the functional-test isolation guidance in AGENTS.md, reviewers.md, and the recording guide to preserve session-injected env via `append(os.Environ(), cli.Env...)` and to disable telemetry with `AZURE_DEV_COLLECT_TELEMETRY=no` so the background writer doesn't hold files open in the temp dir. The three docs are consistent with each other and with the test code (`extension_test.go` sets both env vars). Re-approving against the new HEAD.

cli/azd/docs/recording-functional-tests-guide.md
```diff
@@ -724,10 +724,14 @@ Use a private `AZD_CONFIG_DIR` when a test changes user-level state so parallel
...
-cli.Env = append(cli.Env, "AZD_CONFIG_DIR="+configDir)
+cli.Env = append(os.Environ(), cli.Env...)
+cli.Env = append(cli.Env,
+    "AZD_CONFIG_DIR="+configDir,
+    "AZURE_DEV_COLLECT_TELEMETRY=no",
+)
 ```
 
-A new config directory is logged out, so tests that need live Azure access must authenticate in it. Leave `AZD_CONFIG_DIR` unset when a test only uses the existing login.
+A private config dir is not the default for functional tests because live tests inherit the pipeline's one-time `azd auth login` from the shared config dir: a new config directory is logged out, and there is currently no supported way to authenticate one in CI - leave `AZD_CONFIG_DIR` unset when a test needs live Azure access. Playback tests are unaffected: recording sessions authenticate through the session's credential server, not the config dir.
```
richardpark-msft, line 734: Excellent update!

richardpark-msft (approved):
It's all so beautiful. Is there more? Could it get more perfect?

PR #9171, opened 2026-07-16
fix: detect modern Python agents for code deploy

### LINKED ISSUE #9123: [Agents Extension] "Use existing code" flow not properly defaulting to code deploy
? How do you want to initialize your agent?: Use the code in the current directory
? Enter a name for your agent: hr-agent
? Which protocols does your agent support?: responses
? Select a Foundry project to host your agent and any models or tools it uses.: Use an existing Foundry project
? Select a Foundry project: ai-account-ozfwepskkverc / ai-project-hr-agent-dev (westus)
 
An Azure Container Registry (ACR) is required
 
Foundry Hosted Agents need an Azure Container Registry to store container images before deployment.
 
You can:

  • Use an existing ACR

  • Or create a new one from the template during 'azd up'
 
Learn more: https://aka.ms/azdaiagent/docs

? Enter your ACR login server (e.g., myregistry.azurecr.io), or leave blank to create a new one:
Use the Microsoft Foundry azd agent extension | Microsoft Learn
Learn about the Microsoft Foundry azd agent extension and how to scaffold, provision, and deploy an agent end to end.

huimiu:
Fixes: #9123

## Why

The "Use the code in the current directory" flow used a separate, older language detector that did not recognize `pyproject.toml`. Modern Python projects could therefore be treated as unsupported, silently routed to container deployment, and prompted for an unnecessary Azure Container Registry.

## What changed

- Consolidate Python, .NET, and Node.js project signals into one detector shared by init, manifest adoption, and local run flows.
- Recognize `pyproject.toml` as a Python code-deploy signal while preserving existing `requirements.txt`, `*.py`, and `*.csproj` behavior.
- Keep Node.js and unknown projects on container deployment.
- Add regression coverage for modern Python, existing Python/.NET, mixed, unsupported, and unknown projects.
- Document the fix in the extension changelog.

## Why this approach

Sharing one set of project signals prevents init and local-run behavior from drifting again. The change stays shallow and uses the selected source directory as the deployment root, avoiding false positives from nested projects whose entry point and service path would not match the chosen root.

## Validation

- `go test ./internal/cmd`
- `go test ./...`

## PR extension validation

- Built the PR's `azure.ai.agents` extension and packaged it as a self-contained bundle.
- Installed the bundle with `azd extension install` into an isolated `AZD_CONFIG_DIR`; `azd ai agent version` reported PR commit `c68cc1106`.
- Verified the installed extension routes `azd ai agent`, `azd ai agent init --help`, and `azd ai agent code --help` through the azd host.
- Started the installed `azd ai agent init --no-prompt --deploy-mode code` path for Python code deployment. It correctly reached the extension.

### DIFF
--- a/cli/azd/extensions/azure.ai.agents/internal/cmd/helpers.go
+++ b/cli/azd/extensions/azure.ai.agents/internal/cmd/helpers.go
@@ -26,6 +26,7 @@ import (
 	projectpkg "azureaiagent/internal/project"
 
 	"github.com/azure/azure-dev/cli/azd/pkg/azdext"
+	"github.com/azure/azure-dev/cli/azd/pkg/output"
 	"github.com/google/uuid"
 	"golang.org/x/term"
 )
@@ -503,29 +504,99 @@ type ProjectType struct {
 	StartCmd string // suggested start command
 }
 
+type projectLanguages struct {
+	python         bool
+	pythonMetadata bool
+	pythonMain     bool
+	dotnet         bool
+	node           bool
+}
+
+func detectProjectLanguages(projectDir string) projectLanguages {
+	if projectDir == "" {
+		projectDir = "."
+	}
+
+	entries, err := os.ReadDir(projectDir)
+	if err != nil {
+		warnProjectInspectionFailure(os.Stderr, projectDir, err)
+		return projectLanguages{}
+	}
+
+	var languages projectLanguages
+	for _, entry := range entries {
+		if entry.IsDir() {
+			continue
+		}
+
+		name := entry.Name()
+		switch {
+		case name == "pyproject.toml",
+			name == "requirements.txt":
+			languages.python = true
+			languages.pythonMetadata = true
+		case strings.HasSuffix(name, ".py"):
+			languages.python = true
+			if name == "main.py" {
+				languages.pythonMain = true
+			}
+		case strings.HasSuffix(name, ".csproj"):
+			languages.dotnet = true
+		case name == "package.json":
+			languages.node = true
+		}
+	}
+
+	return languages
+}
+
+func warnProjectInspectionFailure(writer io.Writer, projectDir string, err error) {
+	fmt.Fprintf(writer, "%s", output.WithWarningFormat(
+		"WARNING: cannot read project directory %q: %v. "+
+			"Treating the project as unknown, so code deploy will not be offered "+
+			"and no local start command can be detected. "+
+			"Check the service path in azure.yaml and directory permissions.\n",
+		projectDir,
+		err,
+	))
+}
+
+func isPythonProject(projectDir string) bool {
+	return detectProjectLanguages(projectDir).python
+}
+
+func isDotnetProject(projectDir string) bool {
+	return detectProjectLanguages(projectDir).dotnet
+}
+
+func supportsCodeDeploy(projectDir string) bool {
+	languages := detectProjectLanguages(projectDir)
+	// Python metadata is authoritative. A lone .py file is only a
+	// fallback when package.json does not identify a Node project.
+	supportsPython := languages.pythonMetadata ||
+		(languages.python && !languages.node)
+	return languages.dotnet || supportsPython
+}
+
 func detectProjectType(projectDir string) ProjectType {
-	// Python: pyproject.toml or requirements.txt
-	if fileExists(filepath.Join(projectDir, "pyproject.toml")) ||
-		fileExists(filepath.Join(projectDir, "requirements.txt")) {
+	languages := detectProjectLanguages(projectDir)
+
+	if languages.pythonMetadata {
 		if fileExists(filepath.Join(projectDir, "main.py")) {
 			return ProjectType{Language: "python", StartCmd: "python main.py"}
 		}
 		return ProjectType{Language: "python", StartCmd: ""}
 	}
 
-	// .NET: any .csproj file
-	matches, _ := filepath.Glob(filepath.Join(projectDir, "*.csproj"))
-	if len(matches) > 0 {
+	if languages.dotnet {
 		return ProjectType{Language: "dotnet", StartCmd: "dotnet run"}
 	}
 
-	// Node.js: package.json
-	if fileExists(filepath.Join(projectDir, "package.json")) {
+	if languages.node {
 		return ProjectType{Language: "node", StartCmd: "npm start"}
 	}
 
-	// Check for standalone main.py as fallback
-	if fileExists(filepath.Join(projectDir, "main.py")) {
+	if languages.pythonMain {
 		return ProjectType{Language: "python", StartCmd: "python main.py"}
 	}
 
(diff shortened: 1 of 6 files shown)


cli/azd/extensions/azure.ai.agents/internal/cmd/helpers.go
```diff
@@ -502,29 +502,85 @@ type ProjectType struct {
...
+}
+
+func detectProjectLanguages(projectDir string) projectLanguages {
+	if projectDir == "" {
+		projectDir = "."
+	}
+
+	entries, err := os.ReadDir(projectDir)
+	if err != nil {
+		return projectLanguages{}
```
trangevi, line 523: We should probably at least log a warning here

cli/azd/extensions/azure.ai.agents/internal/cmd/helpers.go
```diff
@@ -502,29 +502,85 @@ type ProjectType struct {
...
+
+func isDotnetProject(projectDir string) bool {
+	return detectProjectLanguages(projectDir).dotnet
+}
+
+func supportsCodeDeploy(projectDir string) bool {
+	languages := detectProjectLanguages(projectDir)
+	return languages.dotnet ||
+		languages.pythonMetadata ||
+		(languages.python && !languages.node)
```
trangevi, line 577: Why is this conditional necessary the way it is? What leads to this combination?

trangevi (changes requested):

huimiu:
@copilot rebase with latest main branch

jongio (commented):
Heads up on the diff scope: most of what shows in "Files changed" here isn't part of this change. The `local-preflight` to `arm-provision` rename, the telemetry event/field renames, the core CLI version bump, and the extension changelog/version bumps all come from already-merged PRs (#8844, #9163, #9182), not from this branch.

Diffed against current `main`, this PR is 6 files, all under `extensions/azure.ai.agents/internal/cmd/`:

- `helpers.go`
- `helpers_test.go`
- `init.go`
- `init_adopt.go`
- `init_adopt_deploymode_test.go`
- `init_from_code.go`

The PR's recorded base is `168f02a` (before those merges), which is why the bot flagged an `arm-provision` gRPC contract break, telemetry-name breakage, and unrelated version bumps. None of those files are in the actual change, so those aren't action items on this PR. Re-syncing the base so the bot and reviewers stop re-reviewing already-merged code would help.

On the real change: the lone-`.py` fallback is gated by `languages.python && !languages.node`, and `detectProjectType` checks `node` before the standalone `main.py`, so a Node agent carrying an incidental helper like `tools.py` stays on container deploy. Recognizing `pyproject.toml` as a code-deploy signal is the #9123 fix.

The open items are @trangevi's two questions on `helpers.go` (the warning log and the conditional). Those are what need resolving before this merges.

jongio (approved):
The two items I flagged on `helpers.go` are now addressed in 5013818:

- The failed `os.ReadDir` in `detectProjectLanguages` logs a warning instead of silently returning an empty result. `log` is already imported.
- `supportsCodeDeploy` is split out with a comment explaining that Python metadata is authoritative and a lone `.py` file only counts as a fallback when there is no `package.json`. Behavior is unchanged from the inlined condition, just readable.

The new table tests cover the node-plus-python-helper and standalone-python cases, and CI is green. The remaining unresolved threads (validation_provider, events, CHANGELOG, config_options) are outdated against the pre-rebase base and are not part of this diff anymore.

cli/azd/extensions/azure.ai.agents/internal/cmd/helpers.go
```diff
@@ -503,29 +502,88 @@ type ProjectType struct {
...
+}
+
+func detectProjectLanguages(projectDir string) projectLanguages {
+	if projectDir == "" {
+		projectDir = "."
+	}
+
+	entries, err := os.ReadDir(projectDir)
+	if err != nil {
+		log.Printf("warning: failed to inspect project directory %q: %v", projectDir, err)
```
trangevi: What is the implication of this though? sorry, I should have been clearer in my original comment, what does this mean for the user, and what can they do about it? It looks like we fall back to python by default I believe, that should be conveyed at least so if it's unexpected a user can do something about it

trangevi (changes requested):

cli/azd/extensions/azure.ai.agents/internal/cmd/helpers.go
jongio: On this branch `os.ReadDir` only fails when the project directory is missing or unreadable, and `detectProjectLanguages` returns an empty set. That path doesn't fall back to python. `supportsCodeDeploy` returns false (`dotnet`, `pythonMetadata`, and `python && !node` are all false), so the deploy-mode prompt hides code deploy and the service goes container. `detectProjectType` returns `"unknown"` with an empty start command, not python.

So the user-facing effect is that code deploy won't be offered for this service. I agree the warning should convey that and how to fix it, e.g. `cannot read %q (%v); treating as unknown and skipping code deploy. Check the path in azure.yaml and directory permissions.` That gives the user something to act on.

cli/azd/extensions/azure.ai.agents/internal/cmd/helpers.go
```diff
@@ -503,29 +504,99 @@ type ProjectType struct {
...
+}
+
+func detectProjectLanguages(projectDir string) projectLanguages {
+	if projectDir == "" {
+		projectDir = "."
+	}
+
+	entries, err := os.ReadDir(projectDir)
+	if err != nil {
+		warnProjectInspectionFailure(os.Stderr, projectDir, err)
```
jongio, line 522: This detection-failure warning is written to stderr, but the other user-facing warnings in this extension (`init_copy.go` and `init_models.go`) send `WithWarningFormat` output to stdout via `fmt.Print*`. This message is what tells a user why code deploy was not offered and what to check, so routing it to stderr risks them not seeing it under extension-host output routing. Sending it to stdout would match the existing warnings and make sure it reaches the user.

jongio (commented):
Left one inline note on the new detection-failure warning about stdout vs stderr, so the message that explains why code deploy was not offered actually reaches the user.

jongio (commented):
Re-reviewed after the rebase to `07cee3f`. One item from my earlier pass is still open: `warnProjectInspectionFailure` writes to `os.Stderr`, but the other user-facing warnings in this extension (`init_copy.go`, `init_models.go`) send `WithWarningFormat` output to stdout via `fmt.Print*`. This warning is what tells the user why code deploy wasn't offered and what to check, so route it to stdout instead of stderr so extension-host routing doesn't swallow it.

The rest of the actual change checks out: `supportsCodeDeploy` keeps a Node project with an incidental `.py` helper on container deploy (`python && !node`), `Main.py` no longer false-sets the python-main start command, and `init_adopt.go` now resolves the service path with `paths.JoinAllowRoot` so `<project>/<service>/pyproject.toml` is detected. No new issues from the rebase.

cli/azd/extensions/azure.ai.agents/internal/cmd/helpers.go
huimiu: Updated the warning to explain that the project is treated as unknown and code deploy is skipped, with guidance to check the azure.yaml path and directory permissions. Thanks!

trangevi (approved):

PR #9174, opened 2026-07-16
feat: let extensions report named telemetry events with attributes

### LINKED ISSUE #9255: Add a generic host-validated extension telemetry gRPC API
## Task

azd extensions currently have no governed way to contribute telemetry to azd's own command events. Add a generic, host-controlled gRPC API so an authenticated extension can report a host-validated telemetry value, and register the first field (`agent.deploy.mode`) so a follow-up extension change can produce it.

This is the core enablement, delivered by #9174. It does not itself emit any value.

## Scope

- New gRPC `TelemetryService.AddCommandUsageAttribute(key, value)`; core validates against a host-owned allowlist (key, allowed values, classification, purpose, eligible commands, required capability).
- Command-scoped storage in core so a value attaches to the owning command event (`cmd.deploy` / `cmd.up`) only, never leaking onto synthetic child spans or sibling commands.
- Authenticated and capability-gated; invalid input rejected and never echoed into errors or logs.
- Best-effort and backward compatible; never affects deploy; an older host returns `Unimplemented`.
- Register `agent.deploy.mode` (`code` / `container` / `byo_image`) as `SystemMetadata` / `FeatureInsight`.
- Add a generic `azd.fromPackage` artifact provenance marker.
- Document the SDK API and the registered field (SDK reference and telemetry schema).

## Acceptance criteria

- An authenticated `service-target-provider` extension can call the API; core validates and records allowed values on the correct command event.
- Unknown keys/values and a missing capability are rejected without recording.
- Telemetry failure never changes deployment success, error, or output.
- No producer required for this task: the field is registered but not yet emitted.

## Related

- End goal / producer: #9230 (record hosted agent deployment mode from the agents extension), lands in a follow-up PR on a released core SDK.
- Open question: allowlist strictness (see #9174) — keeping the strict host allowlist ties new extension telemetry fields to the core release cadence.

huimiu:
Adds `TelemetryService.ReportUsage` to the extension gRPC API so extensions can emit named usage events with bounded attributes through `azd`'s existing telemetry pipeline.

## Why this is needed

Extensions run as separate processes, so core telemetry currently shows that an extension ran but not which feature was used. Reusing the host pipeline avoids per-extension exporters and keeps consent, transport, identity, and privacy controls centralized.

## Approach

- Emit one `ext.usage` span per call; `event_name` becomes `extension.event` and caller attributes become `ext.<key>`.
- Stamp extension identity from validated claims and the installed record; callers cannot provide or overwrite it.
- Validate event/key/value and attribute-count limits, then cap accepted events at 100 per `azd` invocation.
- Return `Accepted: false` for non-official-source installs or exhausted budget without failing the extension command.
- Use the installed `azd` registry source as the admission gate. This is an install-source control, not cryptographic first-party proof; extensions admitted to that registry are eligible by design.
- Add `azd demo telemetry` in demo extension v0.8.0, plus the extension guide, ADR, schema, telemetry reference, matrix, and privacy checklist updates.

## Validation

```text
azd version 1.30.0-beta.1 (commit 1817e20d65c0a0923a82ac393fbd112358da8ca9)  PASS
azd x build --skip-install --no-prompt                                PASS
azd x pack --bundle --no-prompt                                       PASS
```

| Scenario | Result |
|---|---|
| `azd demo telemetry --no-prompt` with the exact PR bundle mapped to an isolated `azd` source | `PASS` — `Usage event recorded as an ext.usage span.` |
| `azd demo telemetry --debug --no-prompt` with bundle source | `PASS` — exits 0; debug log confirms the event was dropped because source was `bundle` |
| `azd x publish --registry ... --artifacts ... --no-prompt` | `PASS` — publish succeeds and existing MCP config is preserved |
| `azd extension source validate ... --strict --no-prompt` | `PASS` — `Registry validation passed.` |

All 11 recorded E2E scenarios passed. Telemetry upload was disabled with `AZURE_DEV_COLLECT_TELEMETRY=no`, so this validates the host RPC, acceptance/drop behavior, and span path without sending data to the backend.


cli/azd/pkg/project/service_target_external.go
```diff
@@ -412,3 +418,63 @@ func createProgressFunc(progress *async.Progress[ServiceProgress]) func(string)
...
+			return mode, true
+		}
+	}
+
+	if hasModeMetadata {
+		return "", true
+	}
+
+	for _, artifact := range artifacts {
+		if artifact != nil && artifact.Kind == ArtifactKindContainer {
```
jongio: This container-kind fallback records `container` for any unlabeled container artifact. Combined with the extension-side gap noted on `service_target_agent.go`, a user-provided image that reaches core as an unlabeled container artifact (for example the `--from-package` flow, which bypasses the extension's `Package` labeling) gets recorded as `container` rather than `byo_image`. Fixing only the extension's Publish branch won't fully close this, since this inference can't tell a built image from a provided one. If the from-package provenance is available upstream, carrying it through would let this fallback select `byo_image`.

jongio (commented):
One note on the deployment-mode classification fallback, inline.

jongio (approved):
The from-package classification I flagged earlier is now handled: on `azd deploy` an archive is labeled `code` and an image reference `byo_image`, driven by the `azd.fromPackage` metadata set in the deploy graph, with tests covering both the code and image cases and the standalone-publish gate. Approving.

jongio (commented):
On the from-package classification: `determineArtifactKind` (cli/azd/internal/cmd/publish.go:419) switches on `filepath.Ext(fromPackage)`, but `filepath.Ext` returns only the final suffix, so `filepath.Ext("agent.tar.gz")` is `.gz`, not `.tar.gz`. The `.tar.gz`, `.tar.bz2`, and `.tar.xz` cases in that switch are unreachable, so those `--from-package` archives fall through to `ArtifactKindContainer`. `hostedAgentFromPackageDeploymentMode` then records `agent.deploy.mode=byo_image` for an artifact that should be `code`. Match the compound suffixes with `strings.HasSuffix` before falling back to `filepath.Ext`. (Same point the automated reviewer raised on service_graph.go:383; confirming it's a real miscategorization, not a false positive.)

jongio (approved):
Confirmed the compound-archive classification now uses strings.HasSuffix, so .tar.gz / .tar.bz2 / .tar.xz are correctly classified as code rather than falling through filepath.Ext. The deferred recordHostedAgentDeploymentMode with the nil-guarded package fallback keeps failed Publish spans populated without double-recording, and from-package provenance still takes precedence. Test coverage for both paths looks good. This addresses my earlier note on the container-kind fallback.

vhvb1989 (dismissed):
**Design concern — please review before accepting as proposed.**

My main concern with this PR is that it overloads the result `Artifact.Metadata` bag to carry telemetry-only data (`azure.ai.agents.deploymentMode`) alongside genuine payload information (location, kind, sha256), and then teaches azd **core** how to fish that value back out based on hardcoded knowledge of a single extension (`if extension.Id == "azure.ai.agents"` + the extension's private metadata key). This inverts azd's normal dependency direction — core is supposed to stay extension-agnostic, with extensions depending on the SDK, not the other way around. The result is a stringly-typed, cross-module side-channel (the magic strings are duplicated in core and in the extension with no compiler enforcement) that doesn't generalize: any future hosted-agent-style extension gets nothing until someone edits a core allowlist. It seems like a lot of core plumbing for what is ultimately telemetry generation.

I think the cleaner approach is a generic capability — e.g. an SDK/gRPC method that lets an extension set a telemetry attribute directly — so the extension emits `agent.deploy.mode` itself, `Artifact` stays purely about payload, and core stays extension-agnostic with no hardcoded extension name. Same telemetry outcome, none of the inverted coupling. Can we discuss whether that direction is preferable before merging this as-is?

jongio (commented):
Coming back to this after the design note above. I approved earlier, but the coupling concern is real and worth settling before this merges.

`recordHostedAgentDeploymentMode` lives in `ExternalServiceTarget` (`pkg/project/service_target_external.go`), which is the generic host for every gRPC extension. It now hardcodes `hostedAgentExtensionID = "azure.ai.agents"` and reads one extension's private metadata key `azure.ai.agents.deploymentMode`, gating behavior on `extension.Id != hostedAgentExtensionID`. So the generic extension host carries baked-in knowledge of a single extension's identity and internal schema. That inverts the direction azd normally keeps: core stays extension-agnostic and extensions depend on the SDK, not the reverse.

The metadata key is also duplicated with no compiler enforcement. `agentDeploymentModeArtifactMetadataKey = "azure.ai.agents.deploymentMode"` is declared both in the extension (`extensions/azure.ai.agents/internal/project/service_target_agent.go`) and in core (`service_target_external.go`). If either side edits the literal, the telemetry silently goes blank with no build error.

It also does not generalize: a future hosted-agent-style extension emits nothing until someone adds its id to the core check.

The generic-capability alternative described above (an SDK/gRPC call that lets an extension set a telemetry attribute directly) produces the same `agent.deploy.mode` value without core knowing the extension name or its metadata keys, and keeps `Artifact.Metadata` about payload only. I think it is worth deciding on that direction before merging as-is.

jongio (commented):
The generic host-validated telemetry service resolves the coupling concern from my earlier note. Core no longer hardcodes the agents extension id or reads its private metadata key: eligibility now runs through the signed capability check plus the host-owned key/value allowlist, and the from-package marker is a plain provenance boolean with the constant aliased from the SDK (`MetadataKeyFromPackage = azdext.ArtifactMetadataKeyFromPackage`) so core and extensions can't drift. That's the right direction, and it keeps `Artifact.Metadata` about payload rather than telemetry.

One thing to settle before merge, and it's your own open item rather than a new one: the privacy review of the new extension-to-host telemetry boundary is still unchecked. The allowlist keeps the strict model intact, but the choice to keep it strict versus loosening to namespaced keys should get a sign-off from the telemetry/privacy owner, since it changes who can put values into azd telemetry. Worth locking that down instead of carrying it as an open question into merge.

The test-coverage and docs gaps the automated reviewer flagged (`publish` not classified in `TestCommandTelemetryCoverage`, `from_package_test` only asserting constant equality) are worth a quick pass. Deferring the emitted-data docs to the producer PR makes sense given nothing emits the field yet.

jongio (approved):
The coupling concern from the earlier design is resolved. Core no longer hardcodes an extension id or reads an extension's private metadata key. Extension telemetry now goes through the generic `AddCommandUsageAttribute` call, gated by the signed capability check plus the host-owned key and value allowlist, and `Artifact.Metadata` stays about payload with `azd.fromPackage` as plain provenance.

The command-scoped store attaches values only to the owning `cmd.deploy` or `cmd.up` span and drops anything reported while a synthetic child scope is on top, so a value can't leak onto a sibling or child span. The handler fails closed on missing claims, missing capability, unknown key, invalid policy, and disallowed value, and rejected caller text isn't echoed back into the returned error. Tests cover each of those paths plus concurrent reporting and duplicate collapsing.

One pre-merge gate is still open, and it's your own checklist item rather than a code change: the privacy review of the new extension-to-host telemetry boundary, along with the strict-allowlist versus namespaced-keys decision. That sign-off is worth locking down before merge since it's what opens the boundary. Deferring the emitted-data docs to the producer PR is reasonable while nothing emits the field yet.

vhvb1989 (commented):
Thanks for reworking this — the new host-validated telemetry gRPC service looks like the right direction and addresses the coupling concern I raised (no hardcoded extension id, Artifact.Metadata stays about payload). I've dismissed my earlier change request. I'll do another pass over the key parts and follow up.

jongio (approved):
Re-reviewed at 898b2f9. Checked the host-owned allowlist, the capability gate off signed claims, and the command-scoped storage. `internal/tracing`, `internal/grpcserver`, `pkg/azdext/telemetry` and `cmd/middleware` all pass locally.

One residual item, not blocking this PR. The doc comment on `ResetCommandUsageForTest` says tests relying on command usage state must not use `t.Parallel()`, but the indirect callers still do. `TelemetryMiddleware.Run` now begins and closes a scope on every invocation, and it's exercised from parallel tests in `cmd/middleware`: `telemetry_test.go` `WithRootAction` and `WithChildAction`, plus `TestTelemetryMiddleware_Run_ResponseError_WrapsWithTraceId` in `middleware_test.go`. That makes Begin(A), Begin(B), Close(A) reachable, where Close(A) hits the "not the current scope" branch, gets logged, and leaves A on the stack for the rest of the test binary run.

Nothing asserts extension-contributed values through the middleware today, so nothing fails, and production is one strictly nested command chain per process, so this is test-only right now. Worth either isolating scope ownership per invocation or dropping `t.Parallel()` on the middleware tests that call `Run`, so the first test that does assert those attributes doesn't come in flaky.

jongio (approved):
Re-verified at 898b2f9, unchanged since my last review. The host-validated telemetry service keeps core extension-agnostic: `AddCommandUsageAttribute` fails closed on missing claims, an unregistered key, a missing capability, an invalid policy, and a disallowed value, and it gates on signed capabilities instead of a hardcoded extension id. The command-scoped store attaches values only to the eligible top scope and never falls back to an ancestor, so nothing lands on a synthetic child or sibling span. `azd.fromPackage` stays plain payload provenance with the core constant aliased from the SDK, so core and extensions can't drift. Approving.

jongio (approved):
Re-reviewed at 898b2f9: the host-validated telemetry gRPC service, the command usage scope stack, and the from-package provenance marker. Validation fails closed on unknown keys, missing capabilities, and disallowed values; the scope store is mutex-guarded and rejects out-of-order closes without corrupting state; and the grpcserver, tracing, and middleware tests pass locally. Approving.

jongio (approved):
Approving. Re-checked the host-validated telemetry service, the command-scoped usage store, and the --from-package provenance marker against the current head. The allowlist validation fails closed, capability gating is enforced from signed claims, and command-scoped storage keeps values off sibling and synthetic child spans. No new concerns from me.

jongio (approved):
Re-reviewed against the current HEAD. The host-validated telemetry service keeps the key allowlist, classification, eligible events, and required capabilities host-owned, and command usage attaches to the owning command scope only with no ancestor fallback, so values can't leak onto sibling or parent commands. Auth, capability, allowlist, scope nesting, out-of-order close, and concurrency paths are all covered by tests, and CI is green. No changes needed.

docs/specs/metrics-audit/telemetry-schema.md
```diff
@@ -88,6 +88,7 @@ These are set once at process startup via `resource.New()` and attached to every
 | Service languages | `project.service.languages` | SystemMetadata | FeatureInsight | List of languages |
 | Service language | `project.service.language` | SystemMetadata | PerformanceAndHealth | Single service language |
 | Platform type | `platform.type` | SystemMetadata | FeatureInsight | e.g. `aca`, `aks` |
+| Agent deployment mode | `agent.deploy.mode` | SystemMetadata | FeatureInsight | `string[]`; per-command set of `code`/`container`/`byo_image` contributed by an authenticated extension through the telemetry service; fixed enum, not hashed; App Insights stores JSON text |
```
jongio: **[LOW]** Filed under the azure.yaml project context section

`agent.deploy.mode` isn't read from `azure.yaml` and isn't project context. It's a command-scoped value an extension contributes at runtime through the telemetry service, so listing it next to `project.name` and `project.service.hosts` reads like project config. "Command Entry-Point" (line 100) is a closer fit, or a new section for extension-contributed command usage. Worth sorting out before the privacy review, since that checklist box is still unchecked and this table is what the reviewer reads.

jongio (approved):
One docs nit, not blocking:

- docs/specs/metrics-audit/telemetry-schema.md:91 - the new field sits in the azure.yaml project context section but it's an extension-contributed, command-scoped value

<sub>🤖 agent jongio</sub>

cli/azd/docs/extensions/extension-telemetry-fields.md
```diff
@@ -0,0 +1,106 @@
...
+
+| | |
+|---|---|
+| You choose | The key name and the closed set of values it may take |
+| `azd` chooses | Which span the value lands on, its classification, and how it is exported |
+| Always rejected | Free-form values, values outside your declared set, keys you did not declare |
+
+Values must be a closed enum because that is what makes a privacy review
+possible. A shape-only rule (length and charset) would let a value like
+`byo_image:myregistry.azurecr.io/foo` through and leak a registry name, so the
```
jongio: cspell fails here on `myregistry`, which is what's failing the cspell-lint job:

```
docs/extensions/extension-telemetry-fields.md:24:12 - Unknown word (myregistry)
CSpell: Files checked: 834, Issues found: 1 in 1 file.
```

`AGENTS.md` says to use a file-scoped override rather than the global word list, so in `cli/azd/.vscode/cspell.yaml`:

```yaml
  - filename: docs/extensions/extension-telemetry-fields.md
    words:
      - myregistry
```

Renaming the example registry to something already in the dictionary works too, but the sentence reads better with a registry-shaped name.

cli/azd/pkg/azdext/azd_client.go
```diff
@@ -9,13 +9,25 @@ import (
...
 	"google.golang.org/grpc/credentials/insecure"
 	_ "google.golang.org/grpc/encoding/gzip" // registers gzip compressor for gRPC streams
 	"google.golang.org/grpc/metadata"
 )
 
+// W3C trace context carriers. The gRPC metadata keys match the HTTP header
+// names; the environment variables are what azd sets on the extension
+// process.
+const (
+	traceparentHeader = "traceparent"
```
jongio: These four constants already exist in this same package, exported, in `context.go`:

```go
const (
	TraceparentKey = "traceparent"
	TracestateKey  = "tracestate"
	TraceparentEnv = "TRACEPARENT"
	TracestateEnv  = "TRACESTATE"
)
```

Two sets of names for the same four strings in one package will drift. Drop this block and have `traceContextPairs` use the existing ones.

jongio (commented):
cspell-lint is red on the new `extension-telemetry-fields.md` doc, so this is blocked on that. The Testing section in the description still says cspell is clean, worth updating once it's fixed.

Second comment is a small dedupe in the SDK client.

jongio (commented):
`go test ./internal/grpcserver/` fails on this HEAD, which is what the three red `azure-dev - cli` legs are showing. It reproduces 3/3 locally and passes on the base commit, so it came in with the new span assertions.

- The exact attribute match in `telemetry_service_test.go` breaks under global baggage that other tests in that package set. That's the CI failure.
- `Test_TelemetryService_RecordsNoSpanWhenRejected` can't fail as written, so the reject path still has no real span coverage.
- `publish.go` carries `telemetry` forward on republish but still drops `mcp`, which has the same shape.
- Still open from my last pass: the duplicate trace context constants in `pkg/azdext/azd_client.go`.

Fixes for the first two are inline. I ran the package with both applied and it's green, and the reject test then does catch an injected span.

cli/azd/internal/grpcserver/telemetry_service_test.go
```diff
@@ -0,0 +1,287 @@
...
+	require.Len(t, ended, 1)
+
+	usage := ended[0]
+	require.Equal(t, events.ExtensionUsageEvent, usage.Name())
+	require.Equal(t, command.SpanContext().TraceID(), usage.SpanContext().TraceID())
+	require.ElementsMatch(t, []attribute.KeyValue{
+		fields.ExtensionId.String(extension.Id),
+		fields.ExtensionVersion.String(extension.Version),
+		attribute.String(testTelemetryKey, "container"),
+	}, usage.Attributes())
```
jongio: This exact-set match is why `azure-dev - cli` is red on all three legs. `go test ./internal/grpcserver/` fails here deterministically (3/3 runs), while the base commit passes:

```
elements differ
extra elements in list B:
  Key: "ad.subscription.id"
  Value: "e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855"
```

`tracing.Start` goes through `wrapperTracer.Start` (`internal/tracing/tracer.go:35-38`), which copies process-global baggage onto every span it creates. The `Test_ProjectService*` tests in this package save an environment, and `pkg/environment/local_file_data_store.go:199-201` sets `ad.subscription.id` as a global attribute for a non-UUID subscription id. So by the time this assertion runs, the `ext.usage` span carries four attributes, not three.

It passes under `-run Test_TelemetryService` because that filters out the tests that set the baggage, which is likely why it looked green locally.

This isn't only a test bug. In a real command the span will legitimately carry the global attributes too, so an exact match asserts something that isn't true of the feature. Assert the keys this feature owns instead, the way `test/functional/telemetry_test.go:185-187` already does it:

```suggestion
	recorded := map[attribute.Key]attribute.Value{}
	for _, attr := range usage.Attributes() {
		recorded[attr.Key] = attr.Value
	}

	for _, expected := range []attribute.KeyValue{
		fields.ExtensionId.String(extension.Id),
		fields.ExtensionVersion.String(extension.Version),
		attribute.String(testTelemetryKey, "container"),
	} {
		require.Contains(t, recorded, expected.Key)
		require.Equal(t, expected.Value, recorded[expected.Key])
	}
```

I ran the full package with that applied and it's green.

cli/azd/internal/grpcserver/telemetry_service_test.go
```diff
@@ -0,0 +1,287 @@
...
+	t.Cleanup(func() { otel.SetTracerProvider(previous) })
+
+	_, err := callWith(t, testExtension(), &azdext.ReportUsageAttributeRequest{
+		Key:   testTelemetryKey,
+		Value: "byo_image",
+	})
+
+	requireCode(t, err, codes.InvalidArgument)
+	require.Empty(t, recorder.Ended())
+}
```
jongio, line 448: This test can't fail as written. `internal/tracing/tracing.go:14` resolves `otel.Tracer(...)` once at package init, and the OTel global only ever delegates to the first provider installed in the process. `Test_TelemetryService_AcceptsDeclaredValue` installs one first, so the recorder created here is never wired to `tracing.Start` and `Ended()` is empty no matter what the service does.

I confirmed it by adding a span to the reject path in `telemetry_service.go` and running the package: it still passed. The negative coverage this test exists to give isn't there.

The delegation behavior, if it's useful:

```
after 1st SetTracerProvider: recorder1=1
after 2nd SetTracerProvider: recorder2=0 recorder1=2
```

That also means the accept test only works because it happens to run first. Any future test in this package that installs a provider ahead of it will break it.

Both problems go away by installing one provider for the package and sharing a recorder:

```go
// main_test.go
var spanRecorder = tracetest.NewSpanRecorder()

var testTracerProvider = tracesdk.NewTracerProvider(tracesdk.WithSpanProcessor(spanRecorder))

func TestMain(m *testing.M) {
	otel.SetTracerProvider(testTracerProvider)
	os.Exit(m.Run())
}
```

Then both span tests call `spanRecorder.Reset()` instead of building their own provider, and the accept test starts its `cmd.deploy` span from `testTracerProvider`. With that in place the same injected span does fail this test.

cli/azd/extensions/microsoft.azd.extensions/internal/cmd/publish.go
```diff
@@ -474,6 +474,10 @@ func addOrUpdateExtension(
 				Dependencies:       extensionMetadata.Dependencies,
 				Providers:          extensionMetadata.Providers,
 				Artifacts:          artifacts,
+				// Telemetry declarations are authored directly in the
+				// registry and reviewed there, so carry them forward
+				// instead of dropping them on republish.
+				Telemetry: v.Telemetry,
```
jongio: `McpConfig` has the same shape as `Telemetry` here: registry-only, absent from `ExtensionSchema`, so this rebuild drops it on republish for exactly the same reason. Nothing in `registry.json` declares `mcp` today so no data is being lost yet, but it'll go the same way the moment one does. Worth carrying `McpConfig: v.McpConfig` forward while you're in here.

cli/azd/extensions/microsoft.azd.extensions/internal/cmd/publish.go
```diff
@@ -474,6 +474,11 @@ func addOrUpdateExtension(
 				Dependencies:       extensionMetadata.Dependencies,
 				Providers:          extensionMetadata.Providers,
 				Artifacts:          artifacts,
+				// Telemetry declarations and MCP config are authored
+				// directly in the registry and reviewed there, so carry
+				// them forward instead of dropping them on republish.
+				Telemetry: v.Telemetry,
+				McpConfig: v.McpConfig,
```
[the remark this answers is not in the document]
jongio: `addOrUpdateExtension` does have a test, just not on this branch. `init_test.go:192` (`TestAddOrUpdateExtensionExtensionPack`) passes an empty `&extensions.Registry{}`, so it only exercises the append path at line 489. The update path at 465-486, where `Telemetry` and `McpConfig` are carried forward, has no caller in any test.

So the gap is one more case next to the existing test rather than fresh coverage in `publish_test.go`: seed the registry with a matching id and version that already has `Telemetry` and `McpConfig` set, republish with a manifest that changes `Usage` or `Capabilities`, then assert the registry-owned fields survive while the manifest-owned ones update.

`bundle.go:122` always builds an empty registry, so it never hits this branch and won't cover it either.

jongio (approved):
Re-checked the three things I flagged last round against 2128704.

`go test ./internal/grpcserver/` passes 10/10 locally now that the attribute assertion is a subset check instead of an exact set, so the process-global baggage that was failing `azure-dev - cli` is out of the way.

The tests can actually fail now that `main_test.go` installs the provider once for the package. I broke three things in `telemetry_service.go` one at a time and each was caught: a wrong dynamic attribute key, a span emitted on the reject path, and `trace.WithNewRoot()` on the usage span. Filtering by trace id also keeps both tests safe under `t.Parallel()`, since `tracetest.SpanRecorder` guards its slices.

All 11 `ExtensionVersion` fields are now set in the republish rebuild, so `Telemetry` and `McpConfig` were the last two and nothing else drops silently. `bundle.go` only ever hits the append path, so it's unaffected.

Left one non-blocking note about test coverage on the `publish.go` thread.

jongio (approved):

cli/azd/pkg/azdext/azd_client.go
```diff
@@ -82,15 +83,47 @@ func isLocalhostAddress(address string) bool {
...
+	carrier := propagation.MapCarrier{}
+	propagation.TraceContext{}.Inject(ctx, carrier)
+
+	parent := carrier.Get(TraceparentKey)
+	state := carrier.Get(TracestateKey)
+
+	if parent == "" {
+		parent = os.Getenv(TraceparentEnv)
+		state = os.Getenv(TracestateEnv)
+	}
```
jongio: This forwards the environment value straight into gRPC metadata, and grpc-go rejects non-printable ASCII in a header value before the request ever leaves the client. A trailing newline is enough, which is easy to pick up from `export TRACEPARENT=$(cat ...)` or a CI runner writing the variable:

```
rpc error: code = Internal desc = header key "traceparent" contains value with non-printable ASCII characters
```

I reproduced that against a real grpc client and server: a clean traceparent gets through, the same value with a trailing `\n` fails on metadata validation. `WithAccessToken` is on every call, so one polluted variable takes out every extension RPC, not only `ReportUsageAttribute`.

The part that makes it more likely than it looks: this fallback is exactly the branch a malformed value lands in. `NewContext` and `extension_command.go:203` both run the variable through the propagator, and a malformed value yields an invalid span context, so `Inject` on line 107 writes nothing and `parent` is empty right here.

Round-tripping the environment through the propagator keeps only well-formed values, and matches what those two call sites already do:

```suggestion
	if carrier.Get(TraceparentKey) == "" {
		envCtx := propagation.TraceContext{}.Extract(context.Background(), propagation.MapCarrier{
			TraceparentKey: os.Getenv(TraceparentEnv),
			TracestateKey:  os.Getenv(TracestateEnv),
		})
		carrier = propagation.MapCarrier{}
		propagation.TraceContext{}.Inject(envCtx, carrier)
	}

	parent := carrier.Get(TraceparentKey)
	state := carrier.Get(TracestateKey)
```

I ran that locally: the three existing `Test_WithAccessToken*` tests still pass and the malformed values drop instead of being forwarded.

cli/azd/extensions/extension.schema.json
```diff
@@ -159,11 +159,17 @@
...
-            "const": "metadata",
+          {
+            "type": "string",
+            "const": "metadata",
             "title": "Metadata",
             "description": "Metadata capability enables extensions to provide comprehensive metadata about their commands and capabilities via a metadata command."
+          },
+          {
+            "type": "string",
+            "const": "telemetry",
```
jongio: Both schema files picked up mixed line endings here. Comparing the raw blobs against the merge base:

- `extension.schema.json`: 3 bare LF lines at base, 39 at HEAD, and 22 of those are lines that were CRLF before and are not otherwise touched by this change
- `registry.schema.json`: 3 to 7, 3 of them previously CRLF

That is what is generating the unchanged lines that render as changed in both diffs, for example `"const": "framework-service-provider"` and `"service-target-provider",`. There is no `*.json` rule in `.gitattributes`, so nothing normalizes this and both files stay mixed from here on. Saving the touched lines with CRLF drops the noise.

One other schema note while you are in here: `registry.schema.json` will accept two telemetry entries sharing the same `key`, but `ValidateTelemetryDeclarations` rejects duplicates. `uniqueItems` does not catch it because the objects still differ by `allowedValues`. The Go validator is the real gate so nothing bad gets published, but editors and schema-aware tooling will not flag it.

jongio (commented):
Two things from this pass, one worth fixing before merge.

MEDIUM

- `pkg/azdext/azd_client.go:112-115` forwards the raw TRACEPARENT/TRACESTATE env values into gRPC metadata. One non-printable character in that variable fails every extension RPC, not just telemetry.

LOW

- `extensions/extension.schema.json` and `extensions/registry.schema.json` came out of this with mixed CRLF/LF line endings, which is what is producing the unchanged lines that show as changed in both diffs.

jongio (commented):
Re-checked against a6a5ed8.

MEDIUM

- `pkg/azdext/azd_client.go:112-124` still forwards the raw TRACEPARENT/TRACESTATE env values into gRPC metadata. I reproduced it this round: with a trailing newline on TRACEPARENT, grpc-go rejects the outgoing metadata before the request leaves the process, and every RPC on that connection fails with `Internal: header key "traceparent" contains value with non-printable ASCII characters`, not just the telemetry call. A well-formed value goes through. `NewContext` in `context.go` already runs the same two env vars through `TraceContext.Extract`, which drops malformed input instead of propagating it, so sending the fallback through Extract and re-injecting closes this.

LOW

- `docs/extensions/extension-telemetry-fields.md:42` - a version bump silently drops the declarations. Details inline.
- `extensions/extension.schema.json` and `extensions/registry.schema.json` still carry the mixed CRLF/LF endings, which is what makes unchanged lines show up as changed in both diffs.

cli/azd/docs/extensions/extension-telemetry-fields.md
```diff
@@ -0,0 +1,107 @@
...
+capabilities:
+  - custom-commands
+  - telemetry
+```
+
+## Step 2: Declare the fields in the registry entry
+
+Add a `telemetry` array to your version entry in the official registry
+(`cli/azd/extensions/registry.json`). Unlike `capabilities`, this is not
+copied from `extension.yaml` — it only exists in the reviewed registry entry:
```
jongio: Worth a line here about version bumps. `azd x publish` only carries `telemetry` forward on the branch of `addOrUpdateExtension` that rewrites an existing version entry. A new version takes the append branch, which doesn't set `Telemetry`, so the declarations don't come with it.

`capabilities` does get copied from `extension.yaml`, so the new entry keeps `telemetry` while declaring zero fields. `validateVersion` only errors on the reverse case (fields without the capability), so the registry stays valid and nothing warns. At runtime `isDeclaredValue` returns false for every key, and the extension gets `InvalidArgument` on every report.

Ran it against `addOrUpdateExtension` to confirm:

```
v1.0.0 telemetry: [{Key:ext.contoso.tools.deploy.mode AllowedValues:[code]}]
v1.1.0 telemetry: [] (capabilities=[telemetry])
```

`McpConfig` has always behaved this way on the append path, so this isn't a regression. The difference is that a missing MCP block breaks loudly while a missing telemetry block just stops the signal. Either a sentence here telling authors to re-add the block on every version bump, or a validation warning when the capability is present with no declarations, would cover it.

jongio (approved):
Re-checked at a6a5ed8. `go build ./...` is clean and `go test ./internal/grpcserver/... ./pkg/extensions/... ./pkg/azdext/... ./internal/tracing/...` passes locally, so the span assertion failures from the earlier round are gone. cspell is clean on `extension-telemetry-fields.md` now too.

Two items from my last pass are still open. Neither blocks merge.

- `pkg/azdext/azd_client.go:112-124` still forwards the raw TRACEPARENT/TRACESTATE env values into gRPC metadata. Narrow trigger, but when it fires grpc-go rejects the outgoing metadata and every extension RPC fails, not just telemetry.
- `azd x publish` still drops `telemetry` and `mcp` on the append branch of `addOrUpdateExtension`, so a version bump loses the declarations that the rewrite branch carries forward.

Approving so this isn't held up on either one.

jongio (approved):
Three earlier comments are still open on a6a5ed8. None of them blocks merge.

- `azd_client.go` forwards `TRACEPARENT` into gRPC metadata without round-tripping it through the propagator, so a malformed value fails every extension RPC instead of being dropped.
- `extension.schema.json` and `registry.schema.json` still have mixed line endings, which is what renders untouched lines as changed in both diffs.
- `extension-telemetry-fields.md` doesn't mention that a version bump takes the append path in `addOrUpdateExtension`, which drops the `telemetry` block.

jongio (approved):
The `TRACEPARENT` fallback in `cli/azd/pkg/azdext/azd_client.go:113` still forwards the raw environment value into gRPC metadata, so a trailing newline takes out every extension RPC, not just this one. That thread is still unresolved. A `strings.TrimSpace` there, or running it through the propagator the way `NewContext` and `extension_command.go:203` already do, would close it.

Not blocking this one.

cli/azd/pkg/extensions/validate_registry.go
```diff
@@ -262,7 +262,13 @@ func findLatestVersion(versions []ExtensionVersion) *ExtensionVersion {
 }
 
 // validateVersion validates a single version entry within an extension.
-func validateVersion(result *ExtensionValidationResult, index int, ver *ExtensionVersion, strict bool) {
+func validateVersion(
+	result *ExtensionValidationResult,
+	extensionId string,
```
jongio, line 267: `extensionId` isn't read anywhere in this function. The only hits in the file are `extensionIdRegex`, the `ext.Id` check up in `validateExtension`, and this parameter itself.

Reads like a leftover from the registry-declared telemetry design that ADR-001 replaced. `unparam` isn't in the enabled linter set in `.golangci.yaml`, so CI won't catch it. Dropping the parameter and putting the call site at line 221 back to `validateVersion(&result, i, &ver, strict)` also gets the signature back onto one line.

jongio (commented):
One new thing from this pass at 28f7306. Not blocking.

LOW

- `cli/azd/pkg/extensions/validate_registry.go:267` - `validateVersion` takes an `extensionId` it never reads. Looks like leftover from the registry-declared telemetry design that ADR-001 replaced.

On the classification question raised inline at `telemetry_service.go:108`: ADR-001 already works through gating on a reviewed declaration or install source and rejects both, with the reasoning that ingestion isn't authenticated per client so a client-side allowlist can't actually enforce anything. That reads as a settled call to me, not something to relitigate here.

jongio (approved):
Re-approving to clear the block. My last review at this same head was a LOW nit posted as a comment, which reset the review state to REVIEW_REQUIRED and left the PR blocked with nothing actually blocking it.

Verified at 28f7306: `go build ./...` is clean and `go test ./internal/grpcserver/... ./pkg/azdext/... ./internal/tracing/... ./pkg/extensions/...` passes locally. CI is green.

Still open, none of it blocking:

- `pkg/azdext/azd_client.go:113-124` forwards the raw TRACEPARENT/TRACESTATE env values into gRPC metadata. A trailing newline in either value makes grpc-go reject the header, and that takes out every extension RPC, not just telemetry. Round-tripping through the propagator before injecting would contain it.
- `pkg/extensions/validate_registry.go:267` takes an `extensionId` parameter that nothing in the function reads. `unparam` isn't in the enabled linter set in `.golangci.yaml`, so CI won't catch it.
- The caller-controlled event name and attribute values ship as `SystemMetadata` on the strength of the ADR-001 contract rather than anything the host can verify. That's a documented decision, worth revisiting if a producer extension ever ships outside the official registry.

cli/azd/pkg/azdext/azd_client.go
```diff
@@ -262,3 +295,14 @@ func (c *AzdClient) Validation() ValidationServiceClient {
...
+// Telemetry returns the telemetry service client used to report extension
+// usage events.
+//
+// A fresh client is returned on each call rather than caching it on the
+// AzdClient struct. Service target providers can deploy services concurrently,
+// so an unsynchronized lazily-written cache field could race on first use. The
+// generated client wrapper is cheap and shares the existing connection.
+func (c *AzdClient) Telemetry() TelemetryServiceClient {
+	return NewTelemetryServiceClient(c.connection)
+}
```
JeffreyCA, line 318: Could we update the `microsoft.azd.demo` extension with sample usage of the new telemetry service?

You should be able to consume this directly as there's no go.mod/go.sum file.

cli/azd/internal/grpcserver/telemetry_service.go
```diff
@@ -0,0 +1,133 @@
...
+// cannot assert which extension it is. Install source is recorded as a
+// dimension rather than a gate, matching ext.run and ext.install.
+//
+// It fails closed on the whole request: exceeding any bound records nothing,
+// because dropping the offending attribute alone would ship data that looks
+// complete but is not. Rejected caller text is never echoed back.
+func (s *telemetryService) ReportUsage(
+	ctx context.Context,
+	req *azdext.ReportUsageRequest,
+) (*azdext.ReportUsageResponse, error) {
```
JeffreyCA, line 92: As @tg-msft called out before, could we add some checks so that only 1P extensions are able to emit telemetry?

jongio: Worth separating two things the description currently treats as one.

`ext.run`, `ext.install` and `ext.upgrade` do fire regardless of install source, but every attribute on those spans is authored by the host: id, version, source, exit code. Nothing on the wire comes from the extension. `ext.usage` is the first path where the extension supplies the strings, so "we already report on dev-registry extensions" doesn't transfer. The question isn't whether we observe an unofficial extension, it's whether unreviewed third-party text lands in azd's pipeline under azd's privacy statement. The bounds here are deliberately shape-only (the comment on line 20 says as much), so nothing stops an extension putting a resource name or a prompt in an attribute value.

If you do want a gate, it's cheap. `extension.Source` is already resolved on line 95 and `extensions.MainRegistryName` gives the official registry name, so it's a source comparison plus an error return.

One caveat on that: source is a proxy for 1P, not the same thing. It answers "did this come from the official registry", which only equals 1P for as long as that registry stays first-party. If the intent is genuinely 1P-only, the durable control is registry admission and this check just reflects it. Worth deciding which of the two you mean, because it changes whether a future third-party entry in the official registry gets telemetry for free.

Separately, if part of the concern is pipeline cost rather than provenance, nothing caps events per command. The 32 attribute and 512 character bounds are per event, and `ReportUsage` can be called in a loop.

cli/azd/pkg/azdext/azd_client.go
```diff
@@ -82,15 +83,47 @@ func isLocalhostAddress(address string) bool {
...
+	if parent == "" {
+		return nil
+	}
+
+	pairs := []string{TraceparentKey, parent}
+	if state != "" {
+		pairs = append(pairs, TracestateKey, state)
+	}
+
+	return pairs
```
jongio, line 135: Still open from the previous two rounds and still not blocking, but here is a tested patch so it isn't left as an open question.

The env fallback puts the raw `TRACEPARENT` value on the wire. Any non-printable character in it fails the RPC, and since `WithAccessToken` builds the context for every extension call, it takes out the whole client rather than just telemetry:

```
rpc error: code = Internal desc = header key "traceparent" contains value with non-printable ASCII characters
```

A trailing newline is enough, which is the usual outcome when the value arrives from a file read or an outer CI runner rather than from azd. `TRACEPARENT` is a standard OpenTelemetry variable name, so azd isn't the only thing that sets it.

Every other site in this package already round-trips env through the propagator before use (`context.go:36`, `extension_command.go:203`). This is the one place that skips it. Doing the same here drops a malformed value instead of forwarding it, and the three existing tests in `azd_client_test.go` still pass unchanged:

```suggestion
	carrier := propagation.MapCarrier{}
	propagation.TraceContext{}.Inject(ctx, carrier)

	if carrier.Get(TraceparentKey) == "" {
		envCtx := propagation.TraceContext{}.Extract(context.Background(), propagation.MapCarrier{
			TraceparentKey: os.Getenv(TraceparentEnv),
			TracestateKey:  os.Getenv(TracestateEnv),
		})
		propagation.TraceContext{}.Inject(envCtx, carrier)
	}

	parent := carrier.Get(TraceparentKey)
	if parent == "" {
		return nil
	}

	pairs := []string{TraceparentKey, parent}
	if state := carrier.Get(TracestateKey); state != "" {
		pairs = append(pairs, TracestateKey, state)
	}

	return pairs
```

jongio (approved):
Re-checked at 28f7306. `go build ./...` is clean and `go test ./internal/grpcserver/... ./pkg/azdext/... ./internal/tracing/... ./pkg/extensions/...` passes locally, so nothing here blocks merge. Keeping the approval rather than resetting the PR to REVIEW_REQUIRED over non-blocking notes.

The thing worth settling before merge is @JeffreyCA's 1P gating question, replied to inline. Short version: the symmetry argument in the description doesn't quite cover this case, because `ext.run`, `ext.install` and `ext.upgrade` only carry values the host authors, while `ext.usage` is the first path where extension-authored strings reach the pipeline. That's a different question from "do we report on unofficially sourced extensions", so it's worth answering directly rather than by analogy.

Two earlier notes are still open, neither blocking: the raw `TRACEPARENT` forwarding in `pkg/azdext/azd_client.go` (tested patch attached inline) and the unused `extensionId` parameter on `validateVersion` at `pkg/extensions/validate_registry.go:267`.

cli/azd/internal/grpcserver/telemetry_service.go
huimiu: `ReportUsage` now drops the event unless the extension was installed from the official `azd` registry

cli/azd/pkg/azdext/azd_client.go
huimiu: Added `azd demo telemetry`.

It reports `demo.telemetry.reported` with a low-cardinality mode plus a bucketed duration, and prints whether the host accepted it. Showing the rejected path matters now, because a locally installed extension gets `Accepted: false` under the new official-registry gate. Demo extension bumped to 0.8.0.

cli/azd/internal/grpcserver/telemetry_service.go
```diff
@@ -0,0 +1,180 @@
...
+	// reviewed at runtime, so registry admission is what keeps
+	// unchecked content out of azd's shared pipeline.
+	//
+	// This is a source check, not proof of first party. Registry
+	// admission is the durable control and this only reflects it, so a
+	// future third-party entry in the official registry would report
+	// too. A blank source is not treated as official: the upgrade
+	// resolver defaults it to the main registry, but a gate cannot
+	// inherit that leniency.
+	if !strings.EqualFold(extension.Source, extensions.MainRegistryName) {
```
jongio: **[HIGH]** The source alias isn't proof of official registry admission

`extension.Source` is only the persisted source name. `SourceManager` reserves `bundle`, but it allows the default `azd` source to be removed and that name to be re-added for an arbitrary registry, so an extension from that registry passes this check and its unchecked attributes enter the shared pipeline. Gate on immutable registry provenance, or hard-pin and validate the canonical `azd` source instead of trusting the alias.

azd-code-reviewer

jongio (commented):
One gate-integrity issue needs fixing before this can reliably enforce the documented eligibility policy.

<sub>🤖 agent jongio</sub>

cli/azd/internal/grpcserver/telemetry_service.go
```diff
@@ -0,0 +1,180 @@
...
+		return nil, status.Errorf(codes.InvalidArgument,
+			"event name is required and must be at most %d characters", maxUsageEventNameLength)
+	}
+
+	if len(req.Attributes) > maxUsageAttributes {
+		return nil, status.Errorf(codes.InvalidArgument,
+			"event declares more than %d attributes", maxUsageAttributes)
+	}
+
+	extension, err := s.extensions.GetInstalled(extensions.FilterOptions{Id: claims.Subject})
```
jongio, line 102: On the concurrency point above: the unsynchronized lazy init in `ListInstalled` (`pkg/extensions/manager.go:355-374`) is real, but it isn't reachable from here in production and it isn't new to this PR.

`ExtensionsMiddleware.Run` calls `ListInstalled()` at `cmd/middleware/extensions.go:78` and only starts the gRPC server at line 106. `m.installed` is already populated before the server is listening, so no handler can reach the nil branch.

This is also the seventh gRPC service to call `GetInstalled` from a handler. `event_service.go:71`, `extension_service.go:43`, `framework_service.go:61`, `provisioning_service.go:58`, `service_target_service.go:62`, and `validation_service.go:71` all do the same thing today, so the exposure is unchanged by this PR.

A mutex or `sync.Once` in `manager.go` is still worth doing, but as a standalone fix rather than a prerequisite here.

jongio (commented):
Still open from the 2026-07-31 round: the source alias check at `telemetry_service.go:130`. `SourceManager.Add` only reserves `bundle` (`pkg/extensions/source_manager.go:93`), so the default `azd` source can still be removed and re-added pointing at another registry, and an extension installed from it passes the gate.

Verified the two new commits against my earlier comments. `traceContextPairs` now round-trips both environment values through the propagator, so a malformed `TRACEPARENT` or `TRACESTATE` is dropped instead of reaching gRPC metadata. `TraceparentKey` and `TracestateKey` are the exact header names the W3C propagator reads (`pkg/azdext/context.go:14-15`), so extraction resolves, and unsampled traces still propagate because `IsValid()` doesn't depend on the sampled bit.

Both schema files are byte identical to the merge base again: `extension.schema.json` at 237 CRLF and 42 bare LF, `registry.schema.json` at 279 and 10, matching `69b43d1` on both ends.

Left a note on the concurrency thread.

cli/azd/internal/grpcserver/telemetry_service.go
jongio: The source gate returns before the per-key and per-value bounds are checked, so those two validations only ever run for official-registry installs.

The guide tells authors that local development is where they verify the integration: "You can still verify your integration end to end: the call succeeds and `Accepted` comes back `false`. Because it is not an error, your code runs the same path in development as in production." That parity doesn't hold for a malformed request. Install with `--source dev`, send a 200-character attribute key, and you get `Accepted: false`, which is indistinguishable from the ordinary dev-mode drop. The identical call from an official install returns `InvalidArgument`. The bug class an author most needs to catch before shipping is the one the development path can't show them.

The event name and attribute count checks already sit above the lookup, so this is only about the loop below. Moving this gate to just after the attribute loop keeps every documented behavior intact: the loop only appends to `attributes` and has no side effects, and the budget increment still happens after the gate, so an unofficial install still doesn't spend budget.

cli/azd/docs/extensions/extension-telemetry.md
```diff
@@ -0,0 +1,129 @@
...
+
+## Where the data lands
+
+Each accepted event becomes an `ext.usage` span carrying `extension.id`,
+`extension.version`, `extension.source`, `extension.event`, and one `ext.<key>`
+attribute per entry in your map. The span shares the command's trace, so it
+joins to the originating command on `operation_Id` in Application Insights:
+
+```kusto
+dependencies
```
jongio: This query returns no rows. `SpanToEnvelope` converts every span to `RequestData` (`cli/azd/internal/telemetry/appinsights-exporter/span_to_envelope.go:26,49`), and that is the only exporter in the tree, so `ext.usage` lands in `requests`. Nothing in azd emits `RemoteDependencyData`.

```suggestion
requests
```

cli/azd/docs/extensions/extension-telemetry.md
```diff
@@ -0,0 +1,129 @@
...
+There is a complete working example in the demo extension:
+[`extensions/microsoft.azd.demo/internal/cmd/telemetry.go`](../../extensions/microsoft.azd.demo/internal/cmd/telemetry.go),
+runnable with `azd demo telemetry`.
+
+## Bounds
+
+| Rule | Limit |
+|---|---|
+| Attributes per event | 32 |
+| Event name length | 1–128 characters |
```
jongio: These are byte counts, not character counts. `len()` on a Go string counts UTF-8 bytes (`telemetry_service.go:100,146,151`), and the line right below this table says there are no charset rules. A 128-character non-ASCII event name is therefore rejected at roughly a third of the documented limit.

Either document these as byte limits or switch the three checks to `utf8.RuneCountInString`.

cli/azd/internal/tracing/fields/fields.go
```diff
@@ -1154,6 +1169,14 @@ var (
 		Classification: SystemMetadata,
 		Purpose:        FeatureInsight,
 	}
+	// ExtensionEvent is the event name an extension chose for a usage
+	// event. Extensions own this value; it exists so queries can filter
+	// an extension's events without parsing its attributes.
+	ExtensionEvent = AttributeKey{
```
jongio, line 1188: `extension.event` has no literal-key assertion in `cli/azd/cmd/telemetry_test.go`. The `ExtensionFields` subtest there covers `extension.source.kind` and nothing else, so a rename or typo here drifts from the documented schema while every test that references the constant keeps passing. `key_test.go` covers the `ext.` prefix helper but not this key's name.

jongio (commented):
Re-verified against the merged HEAD. The only new commit is a clean main merge with no change to the PR's own files, so my earlier findings carry over unchanged.

Still open from the 2026-07-31 round: the source alias check at `telemetry_service.go:130`. I re-read `source_manager.go` at this HEAD and it still holds. `Add` reserves only `bundle` (line 93) and `Remove` doesn't protect `azd`. `List` restores the default source only when the whole `extension.sources` key is absent, so once any other source exists, removing `azd` and re-adding that name pointed at another registry succeeds, and an extension installed from it passes the gate.

Four new items, all verified against the code at this HEAD:

- MEDIUM `telemetry_service.go:130`: the source gate short-circuits per-attribute validation, so the documented local-development workflow can't surface attribute bugs
- MEDIUM `extension-telemetry.md:124`: the KQL example queries the wrong table and returns no rows
- LOW `extension-telemetry.md:59`: the bounds are byte counts, not character counts
- LOW `fields.go:1175`: `extension.event` has no literal-key assertion in the telemetry contract test

cli/azd/extensions/microsoft.azd.demo/internal/cmd/telemetry.go
```diff
@@ -0,0 +1,77 @@
...
+			// reason, so the dimension stays cheap to aggregate.
+			resp, err := azdClient.Telemetry().ReportUsage(ctx, &azdext.ReportUsageRequest{
+				EventName: "demo.telemetry.reported",
+				Attributes: map[string]string{
+					"demo.mode":     "sample",
+					"demo.duration": durationBucket(time.Since(started)),
+				},
+			})
+			if err != nil {
+				return fmt.Errorf("failed to report usage event: %w", err)
```
jongio: This fails the command when the report fails, and the guide that points here says not to do that: "Treat the call as best-effort... Never let the result change command behavior, and never retry" (`docs/extensions/extension-telemetry.md`). The same doc calls this file "a complete working example," so it's the thing authors will copy.

It bites hardest on the older-host path the guide calls out. Against an `azd` that predates the service, `ReportUsage` returns `Unimplemented` and `azd demo telemetry` exits non-zero, which is exactly the coupling the contract is trying to prevent.

Logging the error and returning nil keeps the demo honest about the contract, and you still see the failure.

cli/azd/extensions/microsoft.azd.demo/internal/cmd/telemetry.go
```diff
@@ -0,0 +1,77 @@
...
+			// Every value here is a fixed, low-cardinality string. Never
+			// send resource names, paths, prompts, or anything a user
+			// typed: those are customer content and must not reach the
+			// telemetry pipeline. Durations are bucketed for the same
+			// reason, so the dimension stays cheap to aggregate.
+			resp, err := azdClient.Telemetry().ReportUsage(ctx, &azdext.ReportUsageRequest{
+				EventName: "demo.telemetry.reported",
+				Attributes: map[string]string{
+					"demo.mode":     "sample",
+					"demo.duration": durationBucket(time.Since(started)),
```
jongio: `started` is set on line 31 and `time.Since(started)` is evaluated in the request literal a few lines later, with nothing in between, so this bucket is always `under_1s`.

The bucketing itself is the right lesson to teach. It just needs something to measure. Moving `started` above the `NewAzdClient` call would at least time the client setup, or the comment could say the value is illustrative.

docs/specs/metrics-audit/telemetry-schema.md
```diff
@@ -215,6 +216,48 @@ not emitted by azd spans.
...
+[ADR-001](../../architecture/adr-001-extension-telemetry-events.md) for
+the design rationale and
+[Extension Telemetry](../../../cli/azd/docs/extensions/extension-telemetry.md)
+for the author-facing rules.
+
+#### What this class does and does not guarantee
+
+These rules bound what a **non-tampered** install emits: a well-behaved
+extension can only report values that went through registry review. They are
+not a cryptographic provenance guarantee. The capability, the install source,
```
jongio: This sentence describes the earlier design. There's no capability and no stored declaration for telemetry now: `extension-telemetry.md` says "There is no capability to declare and nothing to register," and the glossary entry you added says telemetry is a host service that needs no capability.

What actually backs the gate here is the persisted install source, so the sentence reads truer as something like "the install source originates from local `azd` config, which is the same trust level every other extension gate already depends on." The paragraph's point still holds, it's just naming two inputs that aren't in the flow.

jongio (approved):
A few things from this pass at 1817e20d. None of them blocks merge.

MEDIUM

- `extensions/microsoft.azd.demo/internal/cmd/telemetry.go:46` returns an error from the command when `ReportUsage` fails, and the guide that links to this file as the worked example says the opposite. Inline.
- `docs/specs/metrics-audit/telemetry-schema.md:253` still describes a capability and a stored declaration. Neither exists in this design. Inline.
- `docs/extensions/extension-telemetry.md:124` still queries `dependencies`. `SpanToEnvelope` sends every span through `spanToRequestData` (`internal/telemetry/appinsights-exporter/span_to_envelope.go:19-33,51`), and that is the only exporter in the tree, so `ext.usage` lands in `requests` and this query returns no rows. Suggestion is on the open thread.
- The bounds tables in `extension-telemetry.md:59` and `telemetry-schema.md` still say characters. `telemetry_service.go:100,146,151` use `len()`, which counts UTF-8 bytes, so a 128-character non-ASCII event name is rejected at roughly a third of the documented limit. Either document them as byte limits or switch the three checks to `utf8.RuneCountInString`.

LOW

- `extensions/microsoft.azd.demo/internal/cmd/telemetry.go:42` always reports `under_1s`. Inline.
- Validation ordering is better this round: the event name and the attribute count are now checked before the source gate. The per-key and per-value bounds are still inside the loop below the gate, so a `--source dev` install sending a 200-character key still gets `Accepted: false` rather than `InvalidArgument`. Moving the loop above the gate closes the remainder.
- `pkg/extensions/validate_registry.go:267` still takes an `extensionId` it never reads. `unparam` isn't in the enabled linter set, so CI won't catch it.

The test-validity problem I raised earlier is gone. `main_test.go` installs one provider for the package, so the recorder is actually wired to `tracing.Start`, and `RecordsNoSpanWhenRejected`, `RejectedCallDoesNotSpendBudget`, and `CapHoldsUnderConcurrentCalls` now assert on real spans. The `TRACEPARENT` fallback round-trips through the propagator, so a malformed environment value gets dropped instead of failing every extension RPC. `go build ./...` is clean and `go test ./internal/grpcserver/... ./pkg/azdext/... ./internal/tracing/... ./pkg/extensions/...` passes locally.

cli/azd/pkg/extensions/source_manager.go
```diff
@@ -108,13 +107,31 @@ func (sm *SourceManager) Add(ctx context.Context, name string, source *SourceCon
...
 	return sm.addInternal(source)
 }
 
 // Remove removes an extension source.
 func (sm *SourceManager) Remove(ctx context.Context, name string) error {
 	name = NormalizeSourceKey(name)
 
+	if strings.EqualFold(name, MainRegistryName) {
+		return fmt.Errorf(
+			"'%s' is reserved and cannot be removed, %w",
```
jongio, line 141: This blocks the name unconditionally, before the source is even looked up, and that leaves users with a persisted non-official `azd` entry with no way out.

That state is reachable on shipped azd today: nothing currently stops `azd extension source remove azd` followed by `azd extension source add -n azd -t url -l <anything>`. After this change:

- `Remove` rejects the name here, so they can't drop the bad entry.
- `Add` clears the new reserved check with the official config, then hits `ErrSourceExists`.
- There's no `update` or `set` subcommand, only `add`, `remove`, and `list`.
- `CreateSource` returns `ErrSourceReserved`, which `getSources` logs and skips (manager.go:1430), so `azd extension list` comes back empty with no error, and `--source azd` hard-fails at manager.go:556.

Net result is hand-editing `~/.azd/config.json` to recover, and telemetry stays gated off the whole time.

Cleanest fix is probably to self-heal in `List`: when the persisted `azd` entry doesn't pass `IsOfficialMainRegistrySource`, substitute the official default instead of returning it. That needs no user action. Otherwise, scope this guard so it only blocks removal when the existing entry is already the official one, which at least gives a remove-then-add repair path.

cli/azd/pkg/extensions/manager.go
```diff
@@ -476,6 +476,24 @@ func (m *Manager) GetInstalled(options FilterOptions) (*Extension, error) {
...
 }
 
+// IsOfficialRegistrySource verifies the configured source used by an
+// installed extension before allowing it to report telemetry.
+func (m *Manager) IsOfficialRegistrySource(ctx context.Context, name string) (bool, error) {
+	if strings.TrimSpace(name) == "" {
+		return false, nil
+	}
+
+	source, err := m.sourceManager.Get(ctx, name)
```
jongio, line 486: `Get` goes to `List`, which calls `configManager.Load()`, and that reads and parses the user config off disk on every call with no caching. `ReportUsage` can fire up to the 100-event cap per invocation, so this turns a string compare into 100 file reads.

The write side is the part I'd look at harder: `List` calls `addInternal` to persist the default source when the `extension.sources` key is absent, so a read on the telemetry path can now trigger a config write, and it's reachable concurrently from the gRPC handlers.

Resolving the official-source decision once per server instance and caching it would take both the disk churn and the concurrent write off this path.

jongio (approved):
Approving to unblock. Two non-blocking things worth a follow-up, both inline.

huimiu:

huimiu:

cli/azd/docs/extensions/extension-framework.md
```diff
@@ -1105,6 +1105,11 @@ Extensions can declare the following capabilities in their manifest:
 - **`validation-provider`**: Contribute validation checks to azd's provision validation and future validation pipelines
 - **`metadata`**: Provide comprehensive metadata about commands and configuration schemas
 
+Telemetry is not a capability. Any extension can call the telemetry gRPC
+service, and events are recorded when its configured source matches the
+verified official `azd` registry name, type, and normalized URL — see
+[Extension Telemetry](./extension-telemetry.md).
```
tg-msft: I worry this will confuse people in public facing docs.  We should say something more like "Internal telemetry is available to Microsoft built extensions in this repo and shipping in the official registry."

cli/azd/docs/extensions/extension-sdk-reference.md
```diff
@@ -519,9 +519,53 @@ gRPC client connecting to the azd framework. Auto-discovers the socket via
...
+
+```go
+if _, err := client.Telemetry().ReportUsage(ctx, &azdext.ReportUsageRequest{
+    EventName:  "deploy.completed",
+    Attributes: map[string]string{"deploy.mode": "container"},
+}); err != nil {
+    log.Printf("telemetry unavailable: %v", err)
+}
+```
+
```
tg-msft, line 560: Should you be showing `!Accepted` in this example too?

cli/azd/docs/extensions/extension-sdk-reference.md
```diff
@@ -519,9 +519,53 @@ gRPC client connecting to the azd framework. Auto-discovers the socket via
...
+The host bounds shape only: at most 32 attributes, event name and keys at most
+128 UTF-8 bytes, and values at most 512 UTF-8 bytes. It does not inspect what a
+value means, so keeping values low cardinality and free of customer content is
+the extension author's responsibility. See
+[Extension Telemetry](./extension-telemetry.md) for the full rules and review
+process.
+
+The call is best-effort. An older azd host returns `Unimplemented`. Treat any
+failure as a no-op and never let it change command behavior. Report a value as
+soon as it is known so a later failure still retains it.
```
tg-msft: I don't understand what you mean by "Report a value as soon as it is known so a later failure still retains it."  Can you make this clearer in the doc?

cli/azd/docs/extensions/extension-telemetry.md
```diff
@@ -0,0 +1,135 @@
...
+
+This guide is for extension authors who need `azd` to record a usage signal on
+their behalf — for example, which deployment mode a user picked.
+
+Telemetry is a service `azd` offers to extensions installed from the verified
+official `azd` registry source. Call `ReportUsage` with an event name and the
+attributes you care about. `azd` core owns the identity fields, namespaces your
+attributes, and bounds their size and number. It does not inspect what they
+mean.
+
```
tg-msft, line 20: I'd be good to add a blurb like

> [!IMPORTANT]
> Any telemetry must go through a privacy review to ensure we're respecting the privacy of our users before you start collecting it.  See below for more.

up top in case people stop reading half way through after they get something compiling.

cli/azd/docs/extensions/extension-telemetry.md
```diff
@@ -0,0 +1,135 @@
...
+See [ADR-001](../../../../docs/architecture/adr-001-extension-telemetry-events.md)
+for the reasoning behind this design.
+
+## What you own and what `azd` owns
+
+| | |
+|---|---|
+| You choose | The event name and every attribute key and value |
+| `azd` writes | `extension.id`, `extension.version`, `extension.source`, `extension.event` |
+| `azd` enforces | The `ext.` prefix on your keys, the bounds below, and the official-registry requirement |
```
tg-msft: I think this is a confusing way to present this information because it makes it seem like "event name" and `extension.event` aren't the same thing.  Imagine an approach more like:

| Attribute Name | Attribute Value |
|---|---|
| `extension.id` | id of the current extension, added automatically |
| `extension.version` | version of the current extension, added automatically |
| `extension.source` | source of the current extension, added automatically |
| `extension.event` | name of the event that you choose so you can easily query |
| `ext.*` | unstructured collection of keys/and values |

cli/azd/docs/extensions/extension-telemetry.md
```diff
@@ -0,0 +1,135 @@
...
+            "retries":     "2",
+        },
+    },
+); err != nil {
+    log.Printf("telemetry unavailable: %v", err)
+}
+```
+
+Treat the call as best-effort. Older `azd` hosts return `Unimplemented`, and a
+malformed request is a plain error. Log the error, but never let it change
```
tg-msft: Do we care about this?  Should we recommend requiring a version of azd with telemetry support in the extension like `requiredAzdVersion: ">=1.XX.XX"` rather than tell people to handle `Unimplemented`?  This is also a point in time thing that'll go away for most future authors.

cli/azd/docs/extensions/extension-telemetry.md
```diff
@@ -0,0 +1,135 @@
...
+
+## Your responsibility for content
+
+`azd` does not review your values at runtime. Registry admission is where that
+review happens, which makes the content rules your responsibility as the
+extension author:
+
+- **Never send customer content.** No file paths, resource names, connection
+  strings, prompts, URLs, or anything a user typed. If you are unsure whether a
+  value qualifies, it does.
```
tg-msft:
```suggestion
  value qualifies as customer content, always assume it does to be safe.
```

cli/azd/docs/extensions/extension-telemetry.md
```diff
@@ -0,0 +1,135 @@
...
+
+| Status | Cause |
+|---|---|
+| `Unauthenticated` | The request did not carry the host-issued extension token |
+| `PermissionDenied` | The calling extension is not installed |
+| `InvalidArgument` | The event name is missing, or a per-event bound was exceeded |
+| `Unimplemented` | The `azd` host predates this service |
+
+Error messages never echo the event name, key, or value you sent, so use the
+status code plus your own call site to diagnose.
```
tg-msft: Why not include this to help people debug?  We already recommend they don't return the error/fail when something goes wrong so this shouldn't get surfaced anyway.

cli/azd/docs/extensions/extension-telemetry.md
```diff
@@ -0,0 +1,135 @@
...
+| Your configured source does not match the verified official `azd` registry | Attribute values are never reviewed at runtime, so registry admission is what keeps unchecked content out of `azd`'s pipeline |
+| The per-invocation event budget is spent | `ReportUsage` can be called in a loop, and the per-event bounds do not limit how many events arrive |
+
+Run `azd` with `--debug` to see which one applied.
+
+This means **your events are not recorded while you develop locally**, because
+an extension installed with `--source dev` or from a file path does not pass the
+gate. You can still verify your integration end to end: the call succeeds and
+`Accepted` comes back `false`. Because it is not an error, your code runs the
+same path in development as in production — do not branch on `Accepted`.
```
tg-msft, line 155: TODO: We should consider adding help for people developing locally so they can validate the telemetry they're expecting.  There are lots of options here like flags to dump telemetry to a file on disk, surfacing to a local OTel collector, etc.  It's a nice to have for the future.

cli/azd/extensions/microsoft.azd.demo/internal/cmd/telemetry.go
```diff
@@ -0,0 +1,65 @@
...
+			}
+
+			if resp == nil {
+				log.Print("demo telemetry: host returned an empty response")
+				color.Yellow("Telemetry returned no response; continuing.")
+				return nil
+			}
+
+			if resp.Accepted {
+				color.Green("Usage event recorded as an ext.usage span.")
```
tg-msft, line 52: TODO: We should add an aka.ms link to the -pr repo with steps on how to get access to our telemetry and an example query to check the last N times this `demo.telemetry.reported` event was reported.

cli/azd/grpc/proto/telemetry.proto
```diff
@@ -0,0 +1,38 @@
...
+
+message ReportUsageRequest {
+  // Event name owned by the extension. Recorded as extension.event so
+  // queries can filter an extension's events without parsing attributes.
+  string event_name = 1;
+
+  // Attributes recorded individually on the event, each namespaced by the
+  // host. Values should be low cardinality and must never carry customer
+  // content.
+  map<string, string> attributes = 2;
```
tg-msft, line 31: You need to document the valid sizes/bounds here too.  This will be the source of truth for a lot of people.

cli/azd/internal/grpcserver/telemetry_service.go
```diff
@@ -0,0 +1,188 @@
...
+	}
+
+	// Budget is spent only by events that would otherwise be recorded,
+	// so a malformed call cannot starve a well-behaved one. The count
+	// covers the whole azd process because the service is registered as
+	// a container singleton, which makes it a per-invocation budget
+	// even when a composite command such as up runs several actions.
+	if s.recorded.Add(1) > maxUsageEventsPerInvocation {
+		log.Printf(
+			"telemetry: dropping usage event from %s, limit of %d reached",
```
tg-msft: I think adding the event name would help a lot too in figuring out if something's accidentally getting triggered in a loop.

cli/azd/internal/grpcserver/telemetry_service.go
```diff
@@ -0,0 +1,188 @@
...
+	for key, value := range req.Attributes {
+		if key == "" || len(key) > maxUsageKeyBytes {
+			return status.Errorf(codes.InvalidArgument,
+				"attribute keys are required and must be at most %d UTF-8 bytes",
+				maxUsageKeyBytes)
+		}
+
+		if len(value) > maxUsageValueBytes {
+			return status.Errorf(codes.InvalidArgument,
+				"attribute values must be at most %d UTF-8 bytes",
```
tg-msft: We should say the name of the key in the message here.  Lots of people will probably get bit by this when a value is suddenly larger than seen during testing.

cli/azd/internal/grpcserver/telemetry_service.go
```diff
@@ -0,0 +1,188 @@
...
+
+	// Budget is spent only by events that would otherwise be recorded,
+	// so a malformed call cannot starve a well-behaved one. The count
+	// covers the whole azd process because the service is registered as
+	// a container singleton, which makes it a per-invocation budget
+	// even when a composite command such as up runs several actions.
+	if s.recorded.Add(1) > maxUsageEventsPerInvocation {
+		log.Printf(
+			"telemetry: dropping usage event from %s, limit of %d reached",
+			extension.Id, maxUsageEventsPerInvocation)
```
tg-msft: TODO: We should consider collecting the number of dropped spans so we can track whether or not an extension has issues (i.e., imagine someone's not properly trimming the length of their values and it blows up when reporting telemetry for lots of customers with 1K long values for some reason - how would we be able to track that we're not getting all the expected data?).

docs/architecture/telemetry.md
```diff
@@ -183,6 +183,14 @@ flowchart LR
   - Auth: `ext.auth.*`
   - Dependency: `ext.dependency.*`
 - Extension lifecycle events: `ext.install`, `ext.upgrade`, `ext.promote`
+- Extensions can report **usage events** via `TelemetryService.ReportUsage`.
+  Each event becomes an `ext.usage` span sharing the command's trace, carrying
```
tg-msft:
```suggestion
- Extensions published to the official registry can report **usage events** via `TelemetryService.ReportUsage` after going through a privacy review.
  Each event becomes an `ext.usage` span sharing the command's trace, carrying
```

docs/specs/metrics-audit/telemetry-schema.md
```diff
@@ -222,6 +223,46 @@ not emitted by azd spans.
...
+#### Extension-contributed usage attributes
+
+Extensions do not have individual fields listed in this document. An extension
+reports a named event with an arbitrary attribute map, and `azd` records it on
+an `ext.usage` span alongside `extension.id`, `extension.version`,
+`extension.source`, and `extension.event`.
+
+`azd` core carries no product-specific telemetry semantics for these fields.
+The following rules are enforced by the host and are what this schema
+guarantees about the whole class:
```
tg-msft, line 236: TODO: We should add a linter to help verify we've documented every event and key name.  Then we can update this here to explain how it works.

docs/architecture/adr-001-extension-telemetry-events.md
```diff
@@ -0,0 +1,180 @@
...
+
+The constraint is therefore: extensions report freely within a bounded shape,
+core owns identity and the attribute namespace, and content is governed the way
+core governs its own fields.
+
+That last part only works if there is a review to govern content. `ext.usage` is
+the first path where an extension supplies the strings rather than the host
+authoring every attribute, so the design also has to say which extensions get to
+use it — otherwise unchecked third-party text lands in a pipeline covered by
+`azd`'s privacy statement.
```
tg-msft:
```suggestion
authoring every attribute, so the design also has to say which extensions get to
use it — otherwise unexpected text from an official extension lands in a pipeline covered by
`azd`'s privacy statement.
```

tg-msft (approved):
This looks fantastic!  Adding some nits around the documentation and issues to follow-up on in the future.

docs/specs/metrics-audit/telemetry-schema.md
tg-msft: Filed https://github.com/Azure/azure-dev/issues/9525 to track.

cli/azd/docs/extensions/extension-telemetry.md
tg-msft: Filed https://github.com/Azure/azure-dev/issues/9526 to track.

cli/azd/extensions/microsoft.azd.demo/internal/cmd/telemetry.go
tg-msft: Filed https://github.com/Azure/azure-dev/issues/9528 to track.

cli/azd/internal/grpcserver/telemetry_service.go
tg-msft: Filed https://github.com/Azure/azure-dev/issues/9527 to track.

tg-msft (approved):

huimiu:

tg-msft (approved):

PR #9180, opened 2026-07-16
ci: remove --no-optional from vscode dependency install (fix #9140)

### LINKED ISSUE #9140: ADO vscode pipeline fails: esbuild platform binary dropped by 'npm ci --no-optional'
### Summary

The `azure-dev - vscode` ADO pipeline (`BuildAndTest` → `BuildVscode {Linux,Mac,Windows}` → **Test** task) fails on **all three OSes** for any PR that touches `ext/vscode`. The failure is a build/test tooling issue, not a product code issue.

### Symptom

All three OS Test tasks fail with the platform-specific esbuild binary missing:

```
Error: The package "@esbuild/linux-x64"  could not be found, and is needed by esbuild.   (Linux)
Error: The package "@esbuild/darwin-x64" could not be found, and is needed by esbuild.    (Mac)
Error: The package "@esbuild/win32-x64"  could not be found, and is needed by esbuild.    (Windows)
##[error]PowerShell exited with code '1'.
```

esbuild's error even spells out the cause:

> If you are installing esbuild with npm, make sure that you don't specify the `--no-optional` or `--omit=optional` flags. The `optionalDependencies` feature of `package.json` is used by esbuild to install the correct binary executable for your current platform.

(On Windows there is also an intermittent VS Code `1.128.1` download **checksum mismatch**, but the esbuild error is the common, deterministic root cause across all three OSes.)

### Root cause

The ADO job installs dependencies with **`npm ci --no-optional`**:

- `eng/pipelines/templates/jobs/vscode-build.yml` (Install dependencies step)

esbuild ships its platform binary (`@esbuild/<platform>`) as an **optional dependency**, so `--no-optional` omits it. Both `npm run build:esbuild` and the extension-host test runner (which bundles test files via esbuild) then fail at runtime.

### Why it started failing now

- The ADO job runs `npm i -g npm` (upgrades to the latest npm) **before** `npm ci --no-optional`. Current npm strictly honors `--no-optional` and drops esbuild's optional platform packages.
- GitHub Actions `vscode-ci.yml` (`build-test`) runs the **same** `./ci-test.ps1` with `npm ci --no-optional` but uses `setup-node`'s bundled npm, which still resolves the platform binary — so GitHub is **green on all 3 OSes** while ADO is red. This divergence is why the two CIs disagree.
- Because both `npm i -g npm` and the VS Code "stable" download float, this regressed silently; older `ext/vscode` PRs passed, but it now blocks every current `ext/vscode` PR.

### Proposed fix

Remove `--no-optional` from the dependency install step so esbuild's platform binary is installed:

```diff
-          npm ci --no-optional
+          npm ci
```

in `eng/pipelines/templates/jobs/vscode-build.yml` (and mirror in `.github/workflows/vscode-ci.yml` for parity). Alternatives: pin the global npm version, or add the current-platform esbuild package explicitly.

### Evidence

- Failing ADO build: `azure-dev - vscode` buildId `6560427` (also reproduced in `6560377`), org `azure-sdk`, project `internal`.
- Passing GitHub run: `vscode-ci` `build-test` (ubuntu/macos/windows) on the same commit.
- Discovered while working on #9130 (PR #9136), whose only diff is 4 files under `ext/vscode/src` + `CHANGELOG.md` — none touch build/test tooling.

vhvb1989:
### Summary

Fixes #9140.

The `azure-dev - vscode` ADO pipeline Test task fails on all three OSes for any PR touching `ext/vscode` with:

```
Error: The package "@esbuild/<platform>" could not be found, and is needed by esbuild.
```

### Root cause

The dependency install step used `npm ci --no-optional`. esbuild ships its platform-specific binary (`@esbuild/<platform>`) as an **optional dependency**, so `--no-optional` drops it, breaking both `build:esbuild` and the extension-host test runner. Recent global npm (`npm i -g npm`) strictly honors `--no-optional`, which is why this regressed silently on ADO while GitHub Actions (using setup-node's bundled npm) stayed green.

### Change

Remove `--no-optional` from the dependency install step in:

- `eng/pipelines/templates/jobs/vscode-build.yml` (ADO job)
- `.github/workflows/vscode-ci.yml` (GitHub Actions, for parity)

### DIFF
--- a/.github/workflows/vscode-ci.yml
+++ b/.github/workflows/vscode-ci.yml
@@ -63,7 +63,7 @@ jobs:
 
       - name: Install dependencies
         run: |
-          npm ci --no-optional
+          npm ci
         working-directory: ./ext/vscode
 
       - name: Test
--- a/eng/pipelines/templates/jobs/vscode-build.yml
+++ b/eng/pipelines/templates/jobs/vscode-build.yml
@@ -35,7 +35,7 @@ jobs:
 
       - pwsh: |
           npm i -g npm @vscode/vsce
-          npm ci --no-optional
+          npm ci
         workingDirectory: ext/vscode
         displayName: Install dependencies
 


jongio (commented):
The same `npm ci --no-optional` pattern (with a preceding global `npm i -g npm`) is still in `eng/pipelines/templates/jobs/build-scan-vscode.yml` line 38. That job runs `ci-package.ps1` -> `npm run build:esbuild` -> `node esbuild.mjs`, so it'll hit the same missing `@esbuild/<platform>` failure the moment it runs. It's wired into the build-notice pipeline rather than the PR gate, so it doesn't block here, but it has the identical latent bug. Worth dropping `--no-optional` there too for parity while you're in this area.

richardpark-msft (approved):
Looks good. @jongio's issue looks like something you could file to address later - it's making me think we could (possibly) refactor the pipelines a bit so they share steps and we don't have to address them in separate areas.

jongio (approved):
Verified the fix: esbuild ships its platform binary (`@esbuild/<platform>`) as an optional dependency, so `npm ci --no-optional` drops it and breaks both `build:esbuild` and the extension-host test runner. Removing the flag is the correct fix. My earlier note about the same pattern in `build-scan-vscode.yml` still stands as a follow-up, but it's out of scope here.

